PR #98, opened 2024-08-22
feat: ledger UTxO tracking

agaffney:


state/state.go
```diff
@@ -64,85 +72,224 @@ func NewLedgerState(dataDir string, eventBus *event.EventBus, logger *slog.Logge
...
+			return fmt.Errorf("remove rolled-back UTxO: %w", err)
+		}
+	}
+	// Restore spent UTxOs
+	result = ls.db.Metadata().Where("deleted_slot > ?", e.Point.Slot).Find(&tmpUtxos)
+	if result.Error != nil {
+		return fmt.Errorf("restore spent UTxOs after rollback: %w", result.Error)
+	}
+	for _, utxo := range tmpUtxos {
+		utxo.DeletedSlot = 0
```
wolf31o2, line 181: It's best to UPDATE them at once in SQL.

state/state.go
```diff
@@ -64,85 +72,224 @@ func NewLedgerState(dataDir string, eventBus *event.EventBus, logger *slog.Logge
...
+			return fmt.Errorf("remove block: %w", err)
+		}
+	}
+	// Delete rolled-back UTxOs
+	var tmpUtxos []models.Utxo
+	result = ls.db.Metadata().Where("added_slot > ?", e.Point.Slot).Order("id DESC").Find(&tmpUtxos)
+	if result.Error != nil {
+		return fmt.Errorf("remove rolled-backup UTxOs: %w", result.Error)
+	}
+	for _, utxo := range tmpUtxos {
```
wolf31o2, line 170: We should be able to UPDATE the list at once via SQL.

wolf31o2 (changes requested):

state/state.go
agaffney: I switched this one to a single UPDATE query

state/state.go
agaffney: The delete operation is also touching BadgerDB. We can look at batching both sets of operations later as an optimization.

agaffney:
Most of the places where we're looping over DB rows to operate one object at a time is because we're also touching Badger as part of each object. I optimized the one place that I was only touching sqlite. We can look at batching of those sorts of updates as a later optimization

wolf31o2 (approved):

PR #155, opened 2024-09-30
feat: support for loading cardano-node and genesis configs

agaffney:
This commit also includes the upstream cardano-node and genesis config files for the preview network


Dockerfile
```diff
@@ -8,4 +8,6 @@ RUN make build
 
 FROM debian:bookworm-slim AS node
 COPY --from=build /code/node /bin/
+COPY ./configs/cardano /opt/cardano/configs
+ENV CARDANO_CONFIG=/opt/cardano/configs/preview/config.json
```
wolf31o2, line 12: It's `/opt/cardano/config` in the cardano-node images. Any reason not to pull the configs from the container image like we do there?

```
FROM ghcr.io/blinklabs-io/cardano-configs:20240725-1 AS cardano-configs
```

Then, you can copy the entire set into the container.

```
COPY --from=cardano-configs /config/ /opt/cardano/config/
```

wolf31o2 (changes requested):

Dockerfile
agaffney: I've updated the in-container path to match the cardano-node images.

I was adding the files to our repo because it's useful for local testing (without having to rebuild the docker image over and over) and it's useful for the tests in this PR. Both could be solved in different ways, though.

wolf31o2 (approved):

PR #174, opened 2024-10-18
feat: ledger epoch tracking

### LINKED ISSUE #172: Epoch tracking in ledger
This is needed to support tracking protocol parameter updates (#33). We'll need to hard-code (for now) enough per-era protocol params to do the epoch calculations. We'll also need some basic era tracking to know which set of hard-coded protocol params to use.

agaffney:
Fixes #172

### DIFF
--- a/state/models/epoch.go
+++ b/state/models/epoch.go
@@ -1,0 +1,53 @@
+// Copyright 2024 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


state/models/epoch.go
```diff
@@ -0,0 +1,53 @@
...
+
+package models
+
+import "github.com/blinklabs-io/node/database"
+
+type Epoch struct {
+	ID uint `gorm:"primarykey"`
+	// NOTE: we would normally use this as the primary key, but GORM doesn't
+	// like a primary key value of 0
+	EpochId             uint
```
wolf31o2, line 23: We're definitely going to need an index on this one.

state/models/epoch.go
```diff
@@ -0,0 +1,53 @@
...
+		var err error
+		ret, err = EpochLatestTxn(txn)
+		return err
+	})
+	return ret, err
+}
+
+func EpochLatestTxn(txn *database.Txn) (Epoch, error) {
+	var ret Epoch
+	result := txn.Metadata().Order("id DESC").
```
wolf31o2, line 47: Use epoch number instead of database row

wolf31o2 (changes requested):

wolf31o2 (approved):

PR #213, opened 2024-11-05
fix: assign missed logger to mempool

Newt6611:


wolf31o2 (approved):

wolf31o2:
Nice catch! I had noticed that I wasn't getting any debug logs from the mempool, but I had not had time to look into it. Thank you.

PR #234, opened 2024-11-15
refactor: decouple chainsync from blockfetch

### LINKED ISSUE #175: Decouple chain-sync client and block-fetch
We currently call block-fetch directly in the chain-sync client RollForward handler. Instead, we should generate an event, and the ledger can decide when/if to fetch the block content. We'll need to provide a way for the ledger to ask for a range of blocks, such as a callback function that the consumer can define. This may need to include a connection ID(s) to ensure we ask a node that has already sent us the block headers

agaffney:
* generate events for blockfetch
* create interface for ledger to request range of blocks from node
* migrate bulk mode to ledger and operate in smaller batches
* enable pipelining for chainsync RequestNext messages
* include tip information with chainsync events

Fixes #175


state/event.go
```diff
@@ -37,17 +39,28 @@ type ChainRollbackEvent struct {
 }
 
 const (
-	ChainsyncEventType event.EventType = "chainsync.event"
+	BlockfetchEventType event.EventType = "blockfetch.event"
+	ChainsyncEventType  event.EventType = "chainsync.event"
 )
 
+// BlockfetchEvent represents either a Block or BatchDone blockfetch event. We use
+// a single event tyep for both to make synchronization easier.
```
wolf31o2, line 47: type

wolf31o2 (approved):
Only comment is a typo in a comment, feel free to fix it later

wolf31o2 (approved):

PR #308, opened 2024-12-30
refactor: move config testdata and remove extra config files

wolf31o2:
This is a follow up from the previous inclusion of the cardano-configs image data.


internal/config/config.go
```diff
@@ -38,7 +38,7 @@ type Config struct {
 
 var globalConfig = &Config{
 	BindAddr:      "0.0.0.0",
-	CardanoConfig: "./configs/cardano/preview/config.json",
+	CardanoConfig: "./config/cardano/preview/config.json",
```
agaffney, line 41: This path doesn't exist. We could use `./config/cardano/testdata/config.json` here, but that feels kinda wrong

wolf31o2: I really would prefer we embed the default set into the binary, so we function out of the box without additional files. That's the direction I was going with this, but I'll hold off and do it all at once.

wolf31o2: I cheated and made a symlink, for now. It was a symlink before, so I am calling it fair game.

agaffney (approved):

PR #314, opened 2024-12-31
feat: UTxO by address/ref ledger queries

agaffney:


state/queries.go
```diff
@@ -118,3 +125,43 @@ func (ls *LedgerState) queryShelley(query *olocalstatequery.ShelleyQuery) (any,
 		return nil, fmt.Errorf("unsupported query type: %T", q)
 	}
 }
+
+func (ls *LedgerState) queryShelleyUtxoByAddress(addrs []ledger.Address) (any, error) {
```
verbotenj, line 129: Can we have better return type than `any`?

wolf31o2 (approved):

state/queries.go
agaffney: The parent functions must return `any`, so there's not much point

Issue #43, opened 2024-05-20
Implement UTxO RPC API


wolf31o2:
Support the UTxO RPC spec (v0.5 or higher)

- [x] query #318 
- [x] submit #310 
- [x] sync #313 
- [x] watch #315

PR #315, opened 2024-12-31
feat: utxorpc watch module

wolf31o2:
Relates to #43

### DIFF
--- a/utxorpc/watch.go
+++ b/utxorpc/watch.go
@@ -16,9 +16,11 @@ package utxorpc
 
 import (
 	"context"
-	"log"
+	"fmt"
 
 	"connectrpc.com/connect"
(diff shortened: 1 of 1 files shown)


utxorpc/watch.go
```diff
@@ -37,11 +39,107 @@ func (s *watchServiceServer) WatchTx(
...
+		s.utxorpc.config.Logger.Error(
+			"failed to get chain iterator",
+			"error", err,
+		)
+		return err
+	}
+
+	for {
+		// Check for available block
+		next, err := chainIter.Next(false)
```
agaffney, line 90: You probably want `true` here to make the call blocking. Otherwise this is going to enter a fairly tight loop that's constantly hitting the DB while waiting for a new block.

agaffney (approved):

PR #318, opened 2024-12-31
feat: utxorpc query module

wolf31o2:
Relates to #43

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -31,6 +31,7 @@ import (
 	ouroboros "github.com/blinklabs-io/gouroboros"
 	"github.com/blinklabs-io/gouroboros/cbor"
 	"github.com/blinklabs-io/gouroboros/ledger"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
 	ochainsync "github.com/blinklabs-io/gouroboros/protocol/chainsync"
 	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


state/state.go
```diff
@@ -672,6 +672,11 @@ func (ls *LedgerState) Tip() ochainsync.Tip {
 	return ls.currentTip
 }
 
+// GetCurrentPParams returns the currentPParams value
+func (ls *LedgerState) GetCurrentPParams() any {
```
agaffney, line 676: Change the return type from `any` to `lcommon.ProtocolParameters` and then you can use `.Utxorpc()` on the return value directly

utxorpc/query.go
```diff
@@ -36,9 +41,32 @@ func (s *queryServiceServer) ReadParams(
...
+		fmt.Sprintf(
+			"Got a ReadParams request with fieldMask %v",
+			fieldMask,
+		),
+	)
 	resp := &query.ReadParamsResponse{}
 
+	lsParams := s.utxorpc.config.LedgerState.GetCurrentPParams()
+	// TODO: support all eras for protocol parameters
+	protoParams := lsParams.(lconway.ConwayProtocolParameters)
```
agaffney, line 54: You can remove this type cast if you change the return type on `GetCurrentPParams()` above

state/state.go
wolf31o2: I had originally done that but didn't type cast it, so I stuck w/ any and left it to downstream. This is way better and what I originally intended. Thanks.

agaffney (approved):

verbotenj (approved):

PR #469, opened 2025-03-12
feat(database): metadata store interface

### LINKED ISSUE #325: Create relational database interface
The `database` package Database interface uses GORM, but the implementation is specific to SQLite within the package.

This will need to be done along with #327 as a pair. The interface needs to be generalized and implementations need to register themselves. SQLite specifics should be moved into a sub-package and we should perform the initialization and configuration routines on the storage systems based on registration.

wolf31o2:
Create a `MetadataStore` interface with the following methods to decouple the `database` package from `gorm.DB` directly.

The MetadataStore interface matches the `gorm.DB` type for the following functions.
- AutoMigrate
- Begin
- Close (really DB.DB().Close())
- Create
- DB
- First
- Order
- Where

We have also created custom functions.
- GetCommitTimestamp
- SetCommitTimestamp
- Transaction

This fully removes gorm from the `database` package.

Closes #325

### DIFF
--- a/database/plugin/metadata/sqlite/database.go
+++ b/database/plugin/metadata/sqlite/database.go
@@ -15,6 +15,7 @@
 package sqlite
 
 import (
+	"database/sql"
 	"errors"
 	"fmt"
 	"io"
@@ -41,8 +42,8 @@ func init() {
 	)
 }
 
-// Database stores all data in sqlite. Data may not be persisted
-type Database struct {
+// MetadataStoreSqlite stores all data in sqlite. Data may not be persisted
+type MetadataStoreSqlite struct {
 	dataDir string
 	db      *gorm.DB
 	logger  *slog.Logger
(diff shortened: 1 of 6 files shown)


database/plugin/metadata/sqlite/database.go
```diff
@@ -121,6 +122,55 @@ func (d *Database) init() error {
...
+func (d *MetadataStoreSqlite) Close() error {
+	// get DB handle from gorm.DB
+	db, err := d.DB().DB()
+	if err != nil {
+		return err
+	}
+	return db.Close()
+}
+
+// Create creates a table
```
verbotenj, line 145: Creates a record https://gorm.io/docs/create.html#Create-Record

agaffney (approved):

verbotenj (approved):

PR #523, opened 2025-03-24
fix(config): support absolute paths in cardano-node config

### LINKED ISSUE #519: Absolute paths disallowed in `config.json`
(sorry if this is the wrong repo for this issue)

In your `config.json`, if you specify an absolute path

```json
{
    "ByronGenesisFile": "/home/dev/.yaci-cli/local-clusters/default/node/genesis/byron-genesis.json",
}
```

It will fail to recognize this as an absolute path, and instead fail (as it tries to parse it as a relative path)

wolf31o2:
If the given paths to the genesis files in the cardano-node config are absolute, use them. Otherwise, prepend the `config.json` directory to the file names.

Closes #519

### DIFF
--- a/config/cardano/node.go
+++ b/config/cardano/node.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Blink Labs Software
+// Copyright 2025 Blink Labs Software
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/blinklabs-io/gouroboros/ledger/alonzo"
 	"github.com/blinklabs-io/gouroboros/ledger/byron"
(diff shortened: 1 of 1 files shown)


config/cardano/node.go
```diff
@@ -69,9 +70,13 @@ func NewCardanoNodeConfigFromFile(file string) (*CardanoNodeConfig, error) {
 }
 
 func (c *CardanoNodeConfig) loadGenesisConfigs() error {
+	prefix := "/"
 	// Load Byron genesis
 	if c.ByronGenesisFile != "" {
-		byronGenesisPath := path.Join(c.path, c.ByronGenesisFile)
+		byronGenesisPath := c.ByronGenesisFile
+		if !strings.HasPrefix(byronGenesisPath, prefix) {
```
agaffney, line 77: There's the `path/filepath.IsAbs()` function for this

agaffney (dismissed):

config/cardano/node.go
wolf31o2: Neat. Switched to it.

agaffney (approved):

PR #539, opened 2025-03-25
fix(database): defer commit at transaction create

wolf31o2:


database/utxo.go
```diff
@@ -127,6 +129,7 @@ func (d *Database) UtxoDelete(
 ) error {
 	if txn == nil {
 		txn = d.Transaction(false)
```
agaffney, line 131: This should be `true` to make it a read/write txn

database/utxo.go
```diff
@@ -155,6 +158,7 @@ func (d *Database) UtxosDelete(
...
```
agaffney, line 160: This should be `true` to make it a read/write txn

agaffney (changes requested):

agaffney (approved):

PR #550, opened 2025-03-27
fix(database): delete consumed utxos in a single sqlite query

wolf31o2:


database/plugin/metadata/sqlite/utxo.go
```diff
@@ -193,6 +193,27 @@ func (d *MetadataStoreSqlite) DeleteUtxosAfterSlot(
 	return nil
 }
 
+// DeleteUtxosBeforeSlot deletes Utxos marked as deleted before a given slot
+func (d *MetadataStoreSqlite) DeleteUtxosBeforeSlot(
```
agaffney, line 197: Should this be `DeleteUtxosDeletedBeforeSlot`?

agaffney (approved):

database/plugin/metadata/sqlite/utxo.go
wolf31o2: I had that originally, but we already have `DeleteUtxosAfterSlot` from a previous PR, so I made it match that.

PR #615, opened 2025-04-09
feat(config): support dingo.yaml config file loading with env overrides in dingo

### LINKED ISSUE #529: Support a dingo.yaml
This configuration should be Dingo's configuration and not the Cardano (`cardano-node`) network configuration file. This should be loaded along with the environment variables, similarly to how we handle configuration in `tx-submit-api` repository.

arepala-uml:
1. Added `--config` flag in cmd/dingo to make sure of custom configuration file as input.
2. Added proper yaml tags on all fields in the dingo `Config` struct to enable unmarshaling from YAML.
3. Implemented logic to load configuration from YAML first, and then override with environment variables
4. Updated node.Run() to accept loaded config from main instead of reloading internally.
5. Wrote unit tests to validate config loading with and without a YAML file.
6. Executed the cmd/dingo/main.go with and without config file as input and it is working fine.

Closes #529

### DIFF
--- a/cmd/dingo/main.go
+++ b/cmd/dingo/main.go
@@ -19,6 +19,7 @@ import (
 	"log/slog"
 	"os"
 
+	"github.com/blinklabs-io/dingo/internal/config"
 	"github.com/blinklabs-io/dingo/internal/version"
 	"github.com/spf13/cobra"
 	"go.uber.org/automaxprocs/maxprocs"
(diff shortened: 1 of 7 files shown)


cmd/dingo/main.go
```diff
@@ -76,7 +80,8 @@ func main() {
 	rootCmd := &cobra.Command{
 		Use: programName,
 		Run: func(cmd *cobra.Command, args []string) {
-			serveRun(cmd, args)
+			cfg := config.FromContext(cmd.Context())
+			serveRun(cmd, args, cfg)
```
agaffney, line 84: We should add a check for `cfg == nil` like you added in `serveCommand()`

agaffney (commented):
This looks great. I had just one minor comment

cmd/dingo/main.go
arepala-uml: I have added the check for this and please review it once

wolf31o2 (approved):

PR #649, opened 2025-04-24
feat(database): string representation (bech32) of account

wolf31o2:


database/account.go
```diff
@@ -28,6 +32,19 @@ func (a *Account) TableName() string {
...
+	convData, err := bech32.ConvertBits(a.StakingKey[:], 8, 5, true)
+	if err != nil {
+		return ""
+	}
+	encoded, err := bech32.Encode("stake", convData)
+	if err != nil {
+		return ""
+	}
+	return encoded
+}
```
agaffney, line 46: I don't think this is going to do what you think that it is. You probably need to build a `ledger/common.Address` and use its `String()`

wolf31o2:
Fair enough. Dropping this, for now.

PR #708, opened 2025-05-11
feat: Support Vote DelegationCertificate

### LINKED ISSUE #626: Support VoteDelegationCertificate

arepala-uml:
1. Added support for vote delegation certificate part.

Closes #626


database/plugin/metadata/sqlite/account.go
```diff
@@ -287,6 +287,38 @@ func (d *MetadataStoreSqlite) SetStakeVoteRegistrationDelegation(
...
+	stakeKey := cert.StakeCredential.Credential.Bytes()
+	drep := cert.Drep.Credential[:]
+
+	tmpItem := models.VoteDelegation{
+		StakingKey: stakeKey,
+		Drep:       drep,
+		AddedSlot:  slot,
+	}
+
+	if err := d.SetAccount(stakeKey, nil, drep, slot, true, txn); err != nil {
```
agaffney, line 305: This will overwrite any existing pool delegation information. Instead, you should call `GetAccount()` with the stake key and use the return value to help populate the values in the `SetAccount()` call to make sure we're only changing the relevant ones (`drep` in this case)

agaffney (dismissed):

database/plugin/metadata/sqlite/account.go
arepala-uml: I have made the changes for this and please review it once.

database/plugin/metadata/sqlite/account.go
```diff
@@ -287,6 +289,47 @@ func (d *MetadataStoreSqlite) SetStakeVoteRegistrationDelegation(
...
+	cert *lcommon.VoteDelegationCertificate,
+	slot uint64,
+	txn *gorm.DB,
+) error {
+	stakeKey := cert.StakeCredential.Credential.Bytes()
+	drep := cert.Drep.Credential[:]
+
+	// Fetch current account
+	existingAccount, err := d.GetAccount(stakeKey, txn)
+	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
```
wolf31o2, line 303: The account must exist here. If it doesn't, something has gone terribly wrong.

Follow the example set in `SetStakeDelegation` here, but instead of setting `tmpAccount.Pool` you'll be setting `tmpAccount.Drep` before the Save.

wolf31o2 (changes requested):

database/plugin/metadata/sqlite/account.go
arepala-uml: Modified the changes and please review it once

PR #799, opened 2025-07-19
feat(ledger): Create a slot timer

### LINKED ISSUE #761: Create a slot timer
Create a slot timer, which can be used for any slot-based recurring tasks.

arepala-uml:
1.  Implemented a SlotTimer that ticks at a specified interval (initially based on current epoch length from ledger state).
2. Added ChangeInterval method such that it cleanly updates the timer frequency during runtime. Added this inside processEpochRollover to adjust timer when slot length changes between epochs.
3. Timer runs its own goroutine (Start()) and independently tracks ticks, executes due tasks, and stops cleanly with Stop().
4. Added two test functions such as to ensures scheduled tasks run at expected ticks and verification of dynamic interval changes.
5. Added the functionality to register a function to run on custom interval (ticks/slots).

Closes #761

### DIFF
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -402,6 +402,10 @@ func (ls *LedgerState) processEpochRollover(
 	if err := ls.loadEpochs(txn); err != nil {
 		return fmt.Errorf("load epochs: %w", err)
 	}
+	// Update the slot timer interval based on the new epoch's slot length
+	if ls.SlotTimer != nil {
+		ls.SlotTimer.ChangeInterval(time.Duration(ls.currentEpoch.SlotLength) * time.Millisecond)
+	}
 	ls.config.Logger.Debug(
 		"added next epoch to DB",
 		"epoch", fmt.Sprintf("%+v", ls.currentEpoch),
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -75,6 +75,7 @@ type LedgerState struct {
 	config                           LedgerStateConfig
 	db                               *database.Database
 	timerCleanupConsumedUtxos        *time.Timer
+	SlotTimer                        *SlotTimer
 	currentPParams                   lcommon.ProtocolParameters
 	currentEpoch                     database.Epoch
 	epochCache                       []database.Epoch
@@ -729,6 +730,13 @@ func (ls *LedgerState) loadEpochs(txn *database.Txn) error {
 		ls.currentEpoch = epochs[len(epochs)-1]
 		ls.currentEra = eras.Eras[ls.currentEpoch.EraId]
 	}
+
+	// Initialize timer with current slot length
+	if ls.SlotTimer == nil && ls.currentEpoch.SlotLength > 0 {
+		ls.SlotTimer = NewSlotTimer(time.Duration(ls.currentEpoch.SlotLength) * time.Millisecond)
+		ls.SlotTimer.Start()
+	}
+
 	// Update metrics
 	ls.metrics.epochNum.Set(float64(ls.currentEpoch.EpochId))
 	return nil
--- a/ledger/timer.go
+++ b/ledger/timer.go
@@ -1,0 +1,112 @@
+// Copyright 2025 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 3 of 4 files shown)


ledger/state.go
```diff
@@ -729,6 +739,17 @@ func (ls *LedgerState) loadEpochs(txn *database.Txn) error {
...
+
+	// Initialize timer with current slot length
+	if ls.SlotTimer == nil && ls.currentEpoch.SlotLength > 0 {
+		interval, err := safeMiliToDuration(ls.currentEpoch.SlotLength)
+		if err != nil {
+			return fmt.Errorf("slot length: %w", err)
+		}
+		ls.SlotTimer = NewSlotTimer(interval)
+		ls.SlotTimer.Start()
+	}
```
agaffney, line 751: Move this into `Start()` after the call to `createGenesisBlock()`. We only need this to be after we've loaded the initial epoch information, and we only need it once.

ledger/timer.go
```diff
@@ -0,0 +1,115 @@
...
+	st.mutex.Lock()
+	defer st.mutex.Unlock()
+
+	st.tickCount++
+	for _, task := range st.tasks {
+		if st.tickCount-task.lastRun >= task.interval {
+			go task.task()
+			task.lastRun = st.tickCount
+		}
+	}
```
agaffney, line 86: We should avoid tracking an absolute tick count. On a 32-bit platform with a 1ms ticket interval, we would overflow the tick count in ~24 days (based on my quick math).

Instead, let's track the number of ticks since each scheduled task last ran (incrementing it on each task on each loop). When a task's tick count reaches the configured count, the task function is run and the tick count is reset.

ledger/timer.go
```diff
@@ -0,0 +1,115 @@
...
+	"time"
+)
+
+type ScheduledTask struct {
+	interval int
+	lastRun  int
+	task     func()
+}
+
+type SlotTimer struct {
```
agaffney, line 28: Rename this to `Scheduler` instead of `SlotTimer`

agaffney (changes requested):

ledger/timer.go
arepala-uml: Renamed this to Scheduler and made the necessary changes as well. Please review it.

ledger/timer.go
arepala-uml: Made the change for this and please review it once

ledger/state.go
arepala-uml: Moved this logic to inside the Start() and please review it once

wolf31o2:
Since the linter errors are not related to your code, feel free to add `noctx` to the disabled list for golangci-lint and create a separate issue to fix the `noctx` failures and enable the check again.

ledger/state.go
```diff
@@ -136,11 +138,28 @@ func (ls *LedgerState) Start() error {
...
 	return nil
 }
 
+func safeMiliToDuration(ms uint) (time.Duration, error) {
+	maxSafe := uint(math.MaxInt64 / time.Millisecond)
+	if ms > maxSafe {
+		return 0, fmt.Errorf("value %d too large; overflows time.Duration", ms)
+	}
+	return time.Duration(int64(ms)) * time.Millisecond, nil
+}
```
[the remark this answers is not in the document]
arepala-uml: I have move this code inside the start() function and please review it.

ledger/chainsync.go
```diff
@@ -402,6 +403,16 @@ func (ls *LedgerState) processEpochRollover(
...
 		return fmt.Errorf("load epochs: %w", err)
 	}
+	// Update the scheduler interval based on the new epoch's slot length
+	if ls.Scheduler != nil {
+		slotLength := ls.currentEpoch.SlotLength
+		maxSafe := uint(math.MaxInt64 / time.Millisecond)
+		if slotLength > maxSafe {
+			return fmt.Errorf("SlotLength %d too large; overflows time.Duration", slotLength)
+		}
+		interval := time.Duration(int64(slotLength)) * time.Millisecond
```
wolf31o2, line 413: The slot length can be less than a second. In our `devnet` [configuration](https://github.com/blinklabs-io/docker-cardano-configs/blob/5ba968a40716046c189600b482e93582ab7cf798/config/devnet/shelley-genesis.json#L4), it's `0.1` (seconds).

ledger/state.go
```diff
@@ -136,6 +138,17 @@ func (ls *LedgerState) Start() error {
...
 		return fmt.Errorf("failed to create genesis block: %w", err)
 	}
+	// Initialize timer with current slot length
+	if ls.currentEpoch.SlotLength > 0 {
+		maxSafe := uint(math.MaxInt64 / time.Millisecond)
+		ms := ls.currentEpoch.SlotLength
+		if ms > uint(maxSafe) {
+			return fmt.Errorf("slot length %d too large; overflows time.Duration", ms)
+		}
+		interval := time.Duration(int64(ms)) * time.Millisecond
```
wolf31o2, line 148: Same as above, it can be fractions of a second. It must be above 0.

wolf31o2 (changes requested):

PR #814, opened 2025-07-30
feat: Add initial block forge functionality

### LINKED ISSUE #809: Add initial block forge functionality
Create a function called something like `forgeBlock()` as part of `ledger.LedgerState`, and schedule it to run at the configured block interval for the network. This can be calculated by `1 / ls.config.CardanoNodeConfig.ShelleyGenesis().ActiveSlotsCoeff`. This should only be activated if "dev mode" (#808) is enabled.

This function should construct an "empty" block, with the minimum required params populated in the header (prev hash, block number, and slot number, all based on the current chain tip). We should then marshal the block to CBOR and print a log entry at the DEBUG level with the generated block CBOR (as a hex string).

arepala-uml:
1. Added forgeBlock() method to LedgerState for creating empty blocks. 
2. Created an empty block with required header parameters such as (previous hash from current chain tip,incremented block number & slot number) and marshal blocks to CBOR & logged the hex string.
3. Schedule block forging at calculated block interval when dev mode is enabled.
4. Add devMode field to LedgerStateConfig and updated node configuration to pass dev mode to ledger state.

Closes #809

### DIFF
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -61,6 +61,7 @@ type LedgerStateConfig struct {
 	CardanoNodeConfig  *cardano.CardanoNodeConfig
 	PromRegistry       prometheus.Registerer
 	ValidateHistorical bool
+	DevMode            bool
 	// Callback(s)
 	BlockfetchRequestRangeFunc BlockfetchRequestRangeFunc
 }
@@ -152,6 +153,30 @@ func (ls *LedgerState) Start() error {
 		interval := time.Duration(durationNs)
 		ls.Scheduler = NewScheduler(interval)
 		ls.Scheduler.Start()
+
+		// Schedule block forging if dev mode is enabled
+		if ls.config.DevMode {
+			// Calculate block interval from ActiveSlotsCoeff
+			shelleyGenesis := ls.config.CardanoNodeConfig.ShelleyGenesis()
+			if shelleyGenesis != nil {
+				// Calculate block interval (1 / ActiveSlotsCoeff)
+				activeSlotsCoeff := shelleyGenesis.ActiveSlotsCoeff
+				if activeSlotsCoeff.Rat != nil && activeSlotsCoeff.Rat.Num().Int64() > 0 {
+					activeSlotsCoeffFloat := float64(activeSlotsCoeff.Rat.Num().Int64()) / float64(activeSlotsCoeff.Rat.Denom().Int64())
+					blockInterval := int(1.0 / activeSlotsCoeffFloat)
+
+					// Scheduled forgeBlock to run at the calculated block interval
+					ls.Scheduler.Register(blockInterval, ls.forgeBlock)
+
+					ls.config.Logger.Info(
+						"dev mode block forging enabled",
+						"component", "ledger",
+						"block_interval", blockInterval,
+						"active_slots_coeff", activeSlotsCoeffFloat,
+					)
+				}
+			}
+		}
 	}
 	// Start goroutine to process new blocks
 	go ls.ledgerProcessBlocks()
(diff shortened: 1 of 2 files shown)


ledger/state.go
```diff
@@ -152,6 +153,30 @@ func (ls *LedgerState) Start() error {
...
+		// Schedule block forging if dev mode is enabled
+		if ls.config.DevMode {
+			// Calculate block interval from ActiveSlotsCoeff
+			shelleyGenesis := ls.config.CardanoNodeConfig.ShelleyGenesis()
+			if shelleyGenesis != nil {
+				// Calculate block interval (1 / ActiveSlotsCoeff)
+				activeSlotsCoeff := shelleyGenesis.ActiveSlotsCoeff
+				if activeSlotsCoeff.Rat != nil && activeSlotsCoeff.Rat.Num().Int64() > 0 {
+					activeSlotsCoeffFloat := float64(activeSlotsCoeff.Rat.Num().Int64()) / float64(activeSlotsCoeff.Rat.Denom().Int64())
+					blockInterval := int(1.0 / activeSlotsCoeffFloat)
```
agaffney, line 166: While it doesn't *really* matter here because we're multiplying by `1`, you should generally do this math as `(1 *  activeSlotsCoeff.Rat.Num().Int64()) / activeSlotsCoeff.Rat.Denom().Int64()` to avoid loss of precision. It's the same reason that we don't store the ratio as a float to begin with.

ledger/state.go
```diff
@@ -917,3 +942,66 @@ func (ls *LedgerState) ValidateTx(
...
+func (ls *LedgerState) forgeBlock() {
+	// Get current chain tip
+	currentTip := ls.Tip()
+
+	// Calculate next slot and block number
+	nextSlot := currentTip.Point.Slot + 1
+	nextBlockNumber := currentTip.BlockNumber + 1
+
+	// Create a proper Shelley empty block with minimum required parameters
+	emptyBlock := struct {
```
agaffney, line 956: We should start with Conway and use the `conway.ConwayBlock` type here

ledger/state.go
```diff
@@ -61,6 +61,7 @@ type LedgerStateConfig struct {
 	CardanoNodeConfig  *cardano.CardanoNodeConfig
 	PromRegistry       prometheus.Registerer
 	ValidateHistorical bool
+	DevMode            bool
```
agaffney, line 64: Rename this to `ForgeBlocks`

arepala-uml: Renamed to ForgeBlocks and please review it

ledger/state.go
arepala-uml: Makes sense and made the change. Please review it.

ledger/state.go
```diff
@@ -917,3 +942,68 @@ func (ls *LedgerState) ValidateTx(
 	}
 	return nil
 }
+
+// forgeBlock creates an empty conway block and logs its CBOR representation
+func (ls *LedgerState) forgeBlock() {
+	// Get current chain tip
+	currentTip := ls.Tip()
```
agaffney, line 949: Use `ls.chain.Tip()` instead. We want to use the raw chain tip (rather than the ledger tip) because we'll be adding our new block to the raw chain.

ledger/state.go
```diff
@@ -917,3 +942,68 @@ func (ls *LedgerState) ValidateTx(
...
+
+	// Calculate next slot and block number
+	nextSlot := currentTip.Point.Slot + 1
```
agaffney, line 952: The slot number should be based on the current time instead of the previous block. You can use `ls.TimeToSlot(time.Now())` to calculate this.

agaffney (changes requested):

ledger/state.go
arepala-uml: Made the change and please review it.

agaffney (approved):

PR #886, opened 2025-09-01
feat(database): Implement key-value interface over Google Cloud Storage

### LINKED ISSUE #330: Implement key-value interface over Google Cloud Storage

arepala-uml:
1. Added Google cloud storage blob store (BlobStoreGCS) with initialization, client setup, and bucket handling.
2. Added GetCommitTimestamp and SetCommitTimestamp as same as in badger using a Google Cloud Store Object.
3. Added basic prometheus metrics for counting operations and transferred bytes.
4. Added GcsLogger to keep logging consistent with same as badger.

Closes #330


database/plugin/blob/gcs/database.go
```diff
@@ -0,0 +1,120 @@
...
+	logger *slog.Logger,
+	promRegistry prometheus.Registerer,
+) (*BlobStoreGCS, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+
+	if logger == nil {
+		logger = slog.New(slog.NewJSONHandler(io.Discard, nil))
+	}
+
+	bucketName := os.Getenv("BLOB_GCS_BUCKET")
```
wolf31o2, line 64: Rather than do this, let's support it via `dataDir` with a prefix.

```
gcs://bucket
```

wolf31o2 (changes requested):

PR #889, opened 2025-09-01
feat: track assets with utxos

### LINKED ISSUE #339: Track coin amount and assets for UTxOs
Add a field to the `utxo` table for coin amount and a separate database table for assets. This is needed for querying for UTxOs by policy/asset.

jkawan:
1. Create Assets table and related get and set functions
2. Added Assets to Utxo table

For the fields in Assets table, took reference from this : https://github.com/Andamio-Platform/andamio-indexer/blob/master/database/plugin/metadata/sqlite/models/asset.go

Closes #339


database/plugin/metadata/sqlite/utxo.go
```diff
@@ -203,6 +205,7 @@ func (d *MetadataStoreSqlite) SetUtxo(
 	payment []byte, // payment
 	stake []byte, // stake
 	amount uint64, // amount
+	asset *lcommon.MultiAsset[lcommon.MultiAssetTypeOutput], // asset
```
wolf31o2, line 208: This should be assets (it's multi-asset, so multiple). Add them to tmpUtxo.Assets before saving tmpUtxo if it's not nil.

database/utxo.go
```diff
@@ -128,7 +131,16 @@ func (d *Database) UtxoByRef(
 	if err != nil {
 		return tmpUtxo, err
 	}
-	tmpUtxo = Utxo(utxo)
+	tmpUtxo = Utxo{
```
wolf31o2, line 134: We need to add assets to the Utxo struct above so this doesn't need to change.

jkawan: @wolf31o2 what should be the type for assets  in Utxo struct here ?

wolf31o2:
While trying to test this, I ran into an interesting issue.

```
{"time":"2025-10-02T20:36:17.258199-04:00","level":"ERROR","msg":"failed to process block: add produced UTxO: sql: converting argument $36 type: uint64 values with high bit set are not supported"}
```
This only occurs with Asset tracking. Checking the Dingo tip, we're here:
```
Ledger Tip: Slot: 31283557, Hash: d6dba37f653ebfa2f6bd8261673e270f7c03ada3cef86a7f48f34144aee373d5
```
The block after d6b is bf6bdd159d595cac19ebd2b51ff4cb81b6022192372288718e528de26581ac2f which includes Tx 901d297b9c56b2e8efd273c3778c90636d62a304e4f291029dc271e1ce5f00ce which is doing some mad stuff with minting and outputs.

PR #922, opened 2025-09-15
fix(database): get utxo by ref should return it, even if deleted

wolf31o2:


wolf31o2:
Validation rules and other things depends on this. Leaving it alone for now.

PR #935, opened 2025-10-08
WIP: Initial changes to track transaction

### LINKED ISSUE #864: Track initial simple transaction data
- [ ] Transaction Type
- [ ] Transaction Block Hash and Index (Ref)
- [ ] Transaction Inputs
- [ ] Transaction Outputs

jkawan:
Closes #864

### DIFF
--- a/database/plugin/metadata/sqlite/utxo.go
+++ b/database/plugin/metadata/sqlite/utxo.go
@@ -206,14 +206,16 @@ func (d *MetadataStoreSqlite) SetUtxo(
 	stake []byte, // stake
 	amount uint64, // amount
 	assets *lcommon.MultiAsset[lcommon.MultiAssetTypeOutput], // assets (multi-asset)
+	transactionId *uint,
 	txn *gorm.DB,
 ) error {
 	tmpUtxo := models.Utxo{
-		TxId:       txId,
-		OutputIdx:  idx,
-		AddedSlot:  slot,
-		PaymentKey: payment,
-		StakingKey: stake,
+		TxId:          txId,
+		OutputIdx:     idx,
+		AddedSlot:     slot,
+		PaymentKey:    payment,
+		StakingKey:    stake,
+		TransactionID: transactionId,
 	}
 
 	// Handle assets if present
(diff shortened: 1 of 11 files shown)


database/plugin/metadata/sqlite/utxo.go
```diff
@@ -206,14 +206,16 @@ func (d *MetadataStoreSqlite) SetUtxo(
 	stake []byte, // stake
 	amount uint64, // amount
 	assets *lcommon.MultiAsset[lcommon.MultiAssetTypeOutput], // assets (multi-asset)
+	transactionId *uint,
```
wolf31o2, line 209: Get rid of this. You won't know this information anywhere you need it.

database/plugin/metadata/sqlite/utxo.go
```diff
@@ -206,14 +206,16 @@ func (d *MetadataStoreSqlite) SetUtxo(
...
-		OutputIdx:  idx,
-		AddedSlot:  slot,
-		PaymentKey: payment,
-		StakingKey: stake,
+		TxId:          txId,
+		OutputIdx:     idx,
+		AddedSlot:     slot,
+		PaymentKey:    payment,
+		StakingKey:    stake,
+		TransactionID: transactionId,
```
wolf31o2, line 218: Remove the TransactionID here. It's not the given TxId. It's a little confusing because of the overloading of terms. `TransactionID` in `Utxo` refers to the internal database ID field (primary key) on the Transaction table. It is not the "Transaction ID" as we refer to it in Cardano, which is the hash of the transaction body.

This field would normally be populated by GORM when we save it while embedded in another structure. I'll show you how that works in another pass.

database/plugin/metadata/sqlite/utxo.go
```diff
@@ -241,10 +243,9 @@ func (d *MetadataStoreSqlite) AddUtxos(
 ) error {
 	items := make([]models.Utxo, 0, len(utxos))
 	for _, utxo := range utxos {
-		items = append(
-			items,
-			utxoLedgerToModel(utxo.Utxo, utxo.Slot),
-		)
+		item := utxoLedgerToModel(utxo.Utxo, utxo.Slot)
+		item.TransactionID = utxo.TxId
+		items = append(items, item)
```
wolf31o2, line 248: Revert this back to the original code.

ledger/chainsync.go
```diff
@@ -217,6 +218,29 @@ func (ls *LedgerState) createGenesisBlock() error {
...
+			make([]byte, 32), // Dummy block hash
+			0,
+			inputsCbor,
+			outputsCbor,
+			txn,
+		)
+		if err != nil {
+			return fmt.Errorf("create genesis transaction: %w", err)
+		}
+		genesisTxId := genesisTx.ID
```
wolf31o2, line 242: There's no genesis transaction, only UTxOs. Remove this.

wolf31o2 (changes requested):

PR #960, opened 2025-10-21
fix(database): no error on retire non-existent pool

wolf31o2:


wolf31o2:
I don't think we should throw such errors at the database layer (especially not deep within the plugin, if we do it at database/pool.go I could agree)

PR #967, opened 2025-10-22
fix(chain): lock around writes and deletes in manager

wolf31o2:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced concurrent operation reliability by resolving potential race conditions during simultaneous data updates.


agaffney (changes requested):
The only place we call this is from `Chain.AddBlock()`, and we're locking the manager from there before calling `Manager.addBlock()`

PR #980, opened 2025-10-26
feat: changes to track transaction metadata

### LINKED ISSUE #873: Track transaction metadata

jkawan:
1. Added changes to track transaction metadata

Closes #873 

## Summary by CodeRabbit

* **New Features**
  * Transactions now capture and persist arbitrary transaction metadata.
  * Transaction metadata is serialized and stored alongside each transaction.

### DIFF
--- a/database/plugin/metadata/sqlite/transaction.go
+++ b/database/plugin/metadata/sqlite/transaction.go
@@ -15,12 +15,9 @@
 package sqlite
 
 import (
-	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/blinklabs-io/dingo/database/models"
 	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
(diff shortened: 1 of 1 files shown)


database/plugin/metadata/sqlite/transaction.go
```diff
@@ -123,3 +127,24 @@ func (d *MetadataStoreSqlite) SetTransaction(
 	}
 	return nil
 }
+
+// createTransactionMetadata creates metadata for a transaction
+func createTransactionMetadata(tx lcommon.Transaction, point ocommon.Point, idx uint32) []byte {
```
agaffney: I don't think this is quite what we had in mind with the ticket, but that's probably on us 😅 

Each transaction can contain metadata, which is an arbitrary data structure that provides additional context for a transaction. This structure is used for things like notes about the TX or even NFT metadata.

This metadata (if any) for a TX is available via the `.Metadata()` function on the `ledger.Transaction` object, which returns a `*cbor.LazyValue`. You can call `.Cbor()` on that to get the original metadata bytes to store in the databaes.

agaffney (dismissed):

PR #1009, opened 2025-11-09
feat: embed preview network configs in binary

### LINKED ISSUE #652: Embed default/preview Cardano network configs
Currently, we default to using a local file system path based on the root of the repo checkout. Use `embed` to embed the configuration directory into the binary and load it from the embed when another location is not specified by the user via config or environment variables. Use this instead of defaulting to a relative path.

wolf31o2:
Embeds preview network Cardano configs using Go `embed` package. Preview network uses embedded configs by default, other networks use `/opt/cardano` paths.

Closes #652

## Summary by CodeRabbit

* **New Features**
  * Embedded preview network configuration and genesis assets so the app can run without external Cardano config files.
  * Added fallback loading to use embedded preview configs when filesystem configs are missing.

* **Behavior Change**
  * Default Cardano config path is now determined by network (preview uses embedded path; mainnet/devnet use system paths). Environment-based overrides are applied during config loading.

* **Tests**
  * Added tests validating embedded defaults, embedded genesis presence, fallback loading, and error cases.

### DIFF
--- a/config/cardano/embed.go
+++ b/config/cardano/embed.go
@@ -1,0 +1,24 @@
+// Copyright 2025 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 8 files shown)


wolf31o2:
@coderabbitai review

PR #1019, opened 2025-11-13
fix(database): track a closed sqlite and prevent vacuum

wolf31o2:
## Summary by CodeRabbit

* **Bug Fixes**
  * The metadata store now marks itself as closed during shutdown, preventing further background maintenance and scheduled daily work.
  * Automatic maintenance (including database vacuum) is skipped when storage is unavailable or after closure, avoiding unexpected work during shutdown and improving resource safety.


wolf31o2:
@coderabbitai review

PR #1038, opened 2025-11-23
fix: guard against nil cert deposits

wolf31o2:
## Summary by cubic
Guarded transaction processing against nil certificate deposit maps to prevent errors when deposits are absent. Also pre-sized the deposit map to the number of certificates for safer and slightly faster handling.

- **Bug Fixes**
  - Skip certificate processing when certDeposits is nil.
  - Fail fast on unsupported certificate types for clearer errors.

- **Performance**
  - Pre-sized certDeposits map with len(certs) to avoid extra allocations.

<sup>Written for commit aecf39304cf6fbc04ba7584e4d579a6f49476a03. Summary will update automatically on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for unsupported certificate types with explicit failure notifications.
  * Enhanced null-check validation during transaction processing.

* **Performance**
  * Optimized memory allocation for certificate processing operations.

### DIFF
--- a/database/transaction.go
+++ b/database/transaction.go
@@ -59,6 +59,7 @@ func (d *Database) processCertificate(
 	case *lcommon.VoteRegistrationDelegationCertificate:
 		return d.SetVoteRegistrationDelegation(c, slot, deposit, txn)
 	default:
+		// Fail fast on unsupported certificate types to ensure new types are properly implemented
 		return fmt.Errorf("unsupported certificate type %T", cert)
 	}
 }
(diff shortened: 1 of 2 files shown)


database/transaction.go
```diff
@@ -88,34 +89,32 @@ func (d *Database) SetTransaction(
 				return err
 			}
 		}
-	}
-	err := d.metadata.SetTransaction(tx, point, idx, txn.Metadata())
```
agaffney: Should this be removed? It seems kinda important

PR #1051, opened 2025-11-24
fix(connmanager): guard against setting nil metrics

wolf31o2:
This fixes the current crash on main.

---
## Summary by cubic
Prevent crashes in ConnectionManager by guarding against nil metrics when updating counters. We now early-return if the manager or metrics are nil, and check each gauge before calling Set, fixing the current crash on main.

<sup>Written for commit c7f61e7200c5ca17ed5680512ff3efca80b72140. Summary will update automatically on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced robustness by adding defensive nil checks to prevent potential crashes in metric update operations.
  * Improved stability when handling edge cases where metrics may not be initialized.

### DIFF
--- a/connmanager/connection_manager.go
+++ b/connmanager/connection_manager.go
@@ -120,6 +120,9 @@ func (c *ConnectionManager) initMetrics() {
 }
 
 func (c *ConnectionManager) updateConnectionMetrics() {
+	if c == nil {
+		return
+	}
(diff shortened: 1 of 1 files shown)


wolf31o2:
@coderabbitai review

agaffney (approved):

PR #1064, opened 2025-11-27
refactor: make methods public for _test packages

wolf31o2:
---
## Summary by cubic
Refactored tests to use external _test packages and exported the minimal symbols needed to support them. Added safety checks around Cardano config reads and fixed a mempool consumer lock issue; behavior is otherwise unchanged.

- **Refactors**
  - Moved tests to package_name_test across config, database, ledger, and plugin modules; updated imports.
  - Exported required symbols: internal/config.NewDefaultConfig (and reset helpers), gcs.ValidateCredentials, ledger.LedgerState fields (CurrentEra, Config, EpochCache, CurrentEpoch), and BlockfetchBatchSlotThresholdDefault; made AWS CmdlineOptions public.
  - Made store option fields/loggers/registries public in blob stores (AWS/Badger/GCS) and sqlite metadata (Logger, PromRegistry, DataDir).
  - Made plugin registry thread-safe with a mutex for registration and option population.

- **Bug Fixes**
  - Guarded Cardano config access in LedgerView.NetworkId, SystemStart query, and HardFork era history to avoid panics.
  - Ensured mempool.RemoveConsumer unlocks via defer to prevent races.

<sup>Written for commit bc8ee27796cac02aa702ec9d6ca14f14a896df81. Summary will update automatically on new commits.</sup>

## Summary by CodeRabbit

* **Refactor**
  * Expanded public API: many previously internal fields, options and helpers are now exported to improve extensibility and interoperability.
  * Introduced default global configuration helpers and safe topology accessors.
  * Made plugin registration concurrent-safe.

* **Tests**
  * Converted many tests to external-package style to validate the public API.
  * Enhanced tests with runtime validations for config and genesis loading.

* **Chores**
  * Standardized logging and metrics wiring across storage and blob plugins.

### DIFF
--- a/config/cardano/node_test.go
+++ b/config/cardano/node_test.go
@@ -12,52 +12,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package cardano
+package cardano_test
 
 import (
 	"path"
-	"reflect"
 	"testing"
+
+	"github.com/blinklabs-io/dingo/config/cardano"
 )
 
 const (
 	testDataDir = "testdata"
 )
 
-var expectedCardanoNodeConfig = &CardanoNodeConfig{
-	path:               testDataDir,
-	AlonzoGenesisFile:  "alonzo-genesis.json",
-	AlonzoGenesisHash:  "7e94a15f55d1e82d10f09203fa1d40f8eede58fd8066542cf6566008068ed874",
-	ByronGenesisFile:   "byron-genesis.json",
-	ByronGenesisHash:   "83de1d7302569ad56cf9139a41e2e11346d4cb4a31c00142557b6ab3fa550761",
-	ConwayGenesisFile:  "conway-genesis.json",
-	ConwayGenesisHash:  "9cc5084f02e27210eacba47af0872e3dba8946ad9460b6072d793e1d2f3987ef",
-	ShelleyGenesisFile: "shelley-genesis.json",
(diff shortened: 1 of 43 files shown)


config/cardano/node_test.go
```diff
@@ -12,52 +12,29 @@
...
-	tmpCfg.shelleyGenesis = nil
-	tmpCfg.alonzoGenesis = nil
-	tmpCfg.conwayGenesis = nil
-	if !reflect.DeepEqual(&tmpCfg, expectedCardanoNodeConfig) {
-		t.Fatalf(
-			"did not get expected object\n     got: %#v\n  wanted: %#v\n",
-			tmpCfg,
-			expectedCardanoNodeConfig,
-		)
-	}
```
agaffney: Removing this code makes the test effectively useless. I see why you did it, but we need to find an alternative way to do the comparison after loading from file.

database/plugin/blob/aws/plugin.go
```diff
@@ -21,12 +21,12 @@ import (
...
-		bucket string
-		region string
-		prefix string
+	CmdlineOptions struct {
+		Bucket string
+		Region string
+		Prefix string
 	}
-	cmdlineOptionsMutex sync.RWMutex
+	CmdlineOptionsMutex sync.RWMutex
```
agaffney: This mutex is only referenced inside this package and should not be exported.

Separately, is this mutex even useful? There doesn't seem much change of race conditions unless we plan to initialize multiple copies of this plugin concurrently.

wolf31o2:
This grew in complexity too much. Closing.

PR #1143, opened 2025-12-21
feat(node): per-node life-cycle management

wolf31o2:
---
## Summary by cubic
Introduce node-scoped context and shutdown hooks to manage service lifecycles. Core services now start with the node context and stop cleanly on cancel.

- **New Features**
  - Node creates a context.WithCancel and cancels it during shutdown.
  - LedgerState.Start now accepts a context; Load passes context.Background.
  - Added per-component shutdown functions for ledger, connection manager, peer governor, and UTxO RPC.
  - Ouroboros BlockFetch, ChainSync, and TxSubmission stop work on connection errors.
  - Chain iterator is cancelable and is cancelled when a ChainSync client is removed.

- **Bug Fixes**
  - Validate ChainSync availability on a connection before starting the client.
  - Validate TxSubmission availability on a connection before starting the client.

<sup>Written for commit a2932a2a54f9699f128a704e0fb0ff82b85d2870. Summary will update automatically on new commits.</sup>

## Summary by CodeRabbit

* **Refactor**
  * Core services now use context-driven lifecycle for coordinated startup/shutdown and graceful cancellation.
  * HTTP server startup unified with clearer TLS/non‑TLS handling and graceful shutdown on cancel.

* **Bug Fixes**
  * Improved responsiveness to connection errors and cancellations across networking, block fetch, chain sync, and tx submission.
  * Reduced risk of hanging loops and ensured earlier termination of in‑flight work.

### DIFF
--- a/node.go
+++ b/node.go
@@ -46,7 +46,8 @@ type Node struct {
 	ouroboros      *ouroborosPkg.Ouroboros
 	shutdownFuncs  []func(context.Context) error
 	config         Config
-	done           chan struct{}
+	ctx            context.Context
+	cancel         context.CancelFunc
 	shutdownOnce   sync.Once
 }
 
@@ -55,7 +56,6 @@ func New(cfg Config) (*Node, error) {
 	n := &Node{
 		config:   cfg,
 		eventBus: eventBus,
-		done:     make(chan struct{}),
 	}
 	if err := n.configPopulateNetworkMagic(); err != nil {
 		return nil, fmt.Errorf("invalid configuration: %w", err)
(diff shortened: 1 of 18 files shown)


node.go
```diff
@@ -66,13 +66,14 @@ func New(cfg Config) (*Node, error) {
 	return n, nil
 }
 
-func (n *Node) Run() error {
+func (n *Node) Run(ctx context.Context) error {
 	// Configure tracing
 	if n.config.tracing {
-		if err := n.setupTracing(); err != nil {
+		if err := n.setupTracing(); err != nil { //nolint:contextcheck
```
agaffney: Why are we skipping the `contextcheck` linter here?

wolf31o2: It complains otherwise. setupTracing starts a goroutine which can't be cancelled, so there's no reason to wire through a context.

PR #1237, opened 2026-01-22
fix(ledger): db worker pool deadlock

wolf31o2:
---
## Summary by cubic
Fixes a shutdown deadlock in the ledger database worker pool. Workers now drain queued operations and track in-flight tasks so shutdown completes cleanly without losing work.

- **Bug Fixes**
  - Drain remaining tasks on shutdown and wait for in-flight ops via WaitGroup to prevent deadlocks.
  - Add panic-safe execution in workers; log and continue on panic.
  - Use non-blocking ResultChan sends to avoid blocking when result channels are full.
  - Correct queue-full handling (undo WaitGroup add, return error) and add tests covering shutdown, errors, queue pressure, concurrency, and multiple shutdowns.

<sup>Written for commit 70beb215346b4d4bb174d9dd3611a46375e11d53. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Worker pool now drains pending work on shutdown and reports remaining operations.
  * Improved accounting to keep pending counts consistent when queueing fails or during shutdown.
  * Added panic protection and guaranteed completion signaling to improve worker stability.

* **Tests**
  * Expanded test coverage for worker pool behavior: shutdown, queue-full, concurrency, errors, and result handling.

### DIFF
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -82,6 +82,7 @@ func DefaultDatabaseWorkerPoolConfig() DatabaseWorkerPoolConfig {
 type DatabaseWorkerPool struct {
 	db         *database.Database
 	taskQueue  chan DatabaseOperation
+	drainQueue chan DatabaseOperation // Reference for draining after shutdown
 	shutdownCh chan struct{}
 	wg         sync.WaitGroup
 	closed     bool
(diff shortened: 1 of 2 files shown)


wolf31o2:
@coderabbitai review

ledger/state.go
```diff
@@ -123,17 +126,58 @@ func (p *DatabaseWorkerPool) worker() {
 	for {
 		select {
 		case op := <-p.taskQueue:
-			// Execute the database operation
-			result := DatabaseResult{}
-			result.Error = op.OpFunc(p.db)
-			if op.ResultChan != nil {
+			// Execute the database operation with panic protection
+			func() {
+				defer p.wg.Done()
```
verbotenj: @coderabbitai review if wg count is correct it looks like we will go into negative numbers because we call wg.Done twice.

PR #1402, opened 2026-02-13
fix(topology): close after read and limit size

wolf31o2:
## Summary by cubic
Close the topology config file after reading and cap reads at 10 MB to prevent file descriptor leaks and unbounded memory use.

- **Bug Fixes**
  - Defer Close() in NewTopologyConfigFromFile; add tests for file close and not-found paths.
  - Limit reads to 10 MB in NewTopologyConfigFromReader and return a clear error for oversized inputs (tested).

<sup>Written for commit 02ec656c6816e269772c8011692b99478491230b. Summary will update on new commits.</sup>

### DIFF
--- a/topology/topology.go
+++ b/topology/topology.go
@@ -16,6 +16,7 @@ package topology
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"os"
 )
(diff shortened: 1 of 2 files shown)


wolf31o2:
@cubic-dev-ai review

PR #1411, opened 2026-02-13
fix(event): prevent publishWithTimeout goroutine leak

wolf31o2:
## Summary by cubic
Fixes a goroutine leak in async event publishing by making channel delivery non-blocking and time-bounding remote deliveries. Slow or blocked subscribers no longer stall workers; overflow events are dropped safely.

- **Bug Fixes**
  - Removed publishWithTimeout/AsyncPublishTimeout; async workers call Publish directly. Remote subscribers are bounded via deliverWithTimeout (RemoteDeliverTimeout) with timeouts tracked by a new deliveryTimeouts metric.
  - Made channelSubscriber.Deliver non-blocking; drops when the buffer is full; Deliver after Close returns nil.
  - Added regression tests for goroutine leaks (Publish/PublishAsync), non-blocking Deliver with full buffers, and safe dropping on full buffers.

<sup>Written for commit c7cca446d921a8847626af3fbb4d07ae0aafb61f. Summary will update on new commits.</sup>

### DIFF
--- a/event/delivery_test.go
+++ b/event/delivery_test.go
@@ -3,6 +3,7 @@ package event
 import (
 	"fmt"
 	"testing"
+	"time"
 )
 
 // mockSubscriber returns an error on Deliver to simulate a failing remote client.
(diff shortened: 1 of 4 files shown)


wolf31o2:
@cubic-dev-ai review

PR #1493, opened 2026-02-22
fix: genesis hash validation

wolf31o2:
## Summary by cubic
Validate the database’s genesis against the node config to prevent syncing on a different network. Adds a blob check for genesis CBOR and fails fast with a clear error if mismatched.

- **Bug Fixes**
  - Added Database.HasGenesisCbor(slot, hash) to check for genesis CBOR at the expected blob key.
  - In createGenesisBlock, when a chain already exists (slot > 0), verify genesis CBOR for slot 0 and the computed Byron genesis hash; return a clear mismatch error otherwise.

- **Migration**
  - If you see “genesis hash mismatch” on startup, delete the database directory and resync.

<sup>Written for commit 7bd6cb00b4bd7f67e2987e1c11ef3bc59e81c585. Summary will update on new commits.</sup>

### DIFF
--- a/database/block.go
+++ b/database/block.go
@@ -120,6 +120,21 @@ func (d *Database) SetGenesisCbor(slot uint64, hash []byte, cborData []byte, txn
 	return nil
 }
 
+// HasGenesisCbor checks whether genesis CBOR data exists at the expected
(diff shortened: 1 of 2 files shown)


wolf31o2:
@coderabbitai review

PR #1829, opened 2026-04-05
fix(connmanager): skip source port binding and reduce recycling during catch-up

wcatz:
## Summary
Two changes to improve catch-up stability:

1. **Skip source port binding during catch-up** — outbound connections bind to listen port 3001 for peer sharing. During catch-up, connection churn creates TIME_WAIT sockets that block reconnection (60s per peer). When `IsAtTipFunc` returns false, connections use ephemeral ports instead. Peer sharing is useless during catch-up.

2. **5x stall detection thresholds during catch-up** — stall grace period (2min→10min), plateau detection (4min→20min), and recycle cooldown (4min→20min) are extended when not at tip. Connection recycling during bulk sync causes pipeline resets and dropped rollbacks that slow catch-up far more than the stall itself.

## Root cause
`OutboundSourcePort` (set in `internal/node/node.go:269`) binds every outbound dial to `:3001` with `SO_REUSEPORT`. When connections close, the 4-tuple enters TIME_WAIT for 60s. New dials to the same peer fail with `cannot assign requested address`. During catch-up with many peers churning, this exhausts available connections and the node falls further behind.

## Evidence
- All preview nodes hit `cannot assign requested address` during catch-up
- Nodes that started with ephemeral ports (source port fix) caught up in minutes
- Nodes without the fix stall indefinitely at 40-60k gap
- Tested on texas (SATA HDD), az1 (arm64 control-1), mini-1 (arm64), Pi (arm64 RPi4)

## Test plan
- [x] `go build ./...` clean
- [x] All tests pass
- [x] Deployed to preview fleet — nodes with fix catch up, nodes without stall
- [ ] Verify peer sharing works correctly once node reaches tip (source port re-enabled)

---
## Summary by cubic
Improve catch-up stability by using ephemeral source ports while behind tip and relaxing chainsync stall thresholds to reduce recycling. This prevents TIME_WAIT socket exhaustion and cuts connection churn so nodes catch up faster.

- **Bug Fixes**
  - `connmanager`: Added `IsAtTipFunc`; when not at tip, outbound dials skip binding to `:3001` and use ephemeral ports. Peer sharing resumes automatically at tip.
  - `node`: Increased chainsync stall grace, plateau detection, and recycle cooldown by 5x during catch-up, and wired `IsAtTipFunc` into the connection manager.

<sup>Written for commit 21a70b42ec1e439d3dececaba0c1eea5bc2ce924. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Outbound connections now intelligently manage binding based on synchronization status to optimize resource usage and reduce connection churn during blockchain synchronization.

### DIFF
--- a/connmanager/connection_manager.go
+++ b/connmanager/connection_manager.go
@@ -87,7 +87,13 @@ type ConnectionManagerConfig struct {
 	Listeners          []ListenerConfig
 	OutboundConnOpts   []ouroboros.ConnectionOptionFunc
 	OutboundSourcePort uint
-	MaxInboundConns    int // 0 means use DefaultMaxInboundConnections
+	// IsAtTipFunc returns whether the node is synced to tip. When
+	// provided and returning false, outbound connections skip source
+	// port binding to avoid TIME_WAIT socket exhaustion during
+	// catch-up. Source port binding is only needed for peer sharing
+	// which is not useful while the node is behind.
+	IsAtTipFunc     func() bool
+	MaxInboundConns int // 0 means use DefaultMaxInboundConnections
 	// MaxConnectionsPerIP limits the number of concurrent inbound
 	// connections from the same IP address. IPv6 addresses are grouped
 	// by /64 prefix. A value of 0 means use DefaultMaxConnectionsPerIP.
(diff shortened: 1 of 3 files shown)


wcatz:
Closing — skipping source port binding breaks peer sharing. The catch-up recycling multiplier is the correct fix and will be submitted separately.

PR #1913, opened 2026-04-15
fix: connection ID collision avoidance

### LINKED ISSUE #1912: ConnectionId collision when both sides use listen-port reuse orphans forged blocks
When both Dingo and a Haskell relay use listen-port reuse (OutboundSourcePort), both connections produce identical ConnectionId values on Dingo's side. The inbound connection silently overwrites the outbound entry in the connmanager's connections map, corrupting the IsInboundConnection result for that ConnectionId. All chainsync client callbacks are then silently dropped, Dingo goes deaf to the chain, forged blocks cannot propagate, and the stall detector eventually kills the connection.

## Observed behavior

From preview network logs (2026-04-14), 1 Haskell relay + 1 Dingo block producer:

1. Dingo receives blocks normally via outbound chainsync
2. Relay connects inbound to Dingo (both sides use listen port as source port)
3. Inbound connection overwrites outbound in connmanager (same ConnectionId)
4. IsInboundConnection now returns true for both connections
5. All chainsync client RollForward/RollBackward callbacks silently dropped
6. Dingo stops making chain progress (120s stall timeout fires)
7. Block forged during the deaf window never propagates to relay
8. On reconnect, relay chain does not include Dingo's block, rollback occurs

## Root cause

Both connections share the same (LocalAddr, RemoteAddr) because both sides bind outbound connections to their listen port:

- Dingo outbound: local=Dingo:3005 remote=Relay:3003
- Relay outbound (Dingo inbound): local=Dingo:3005 remote=Relay:3003

addConnectionImpl (connmanager/connection_manager.go:558) does a direct map assignment with no collision check, so the second registration silently overwrites the first. The chainsync client callbacks (ouroboros/chainsync.go:484 and :528) check IsInboundConnection, which now returns true, causing all chain data to be discarded.

## Proposed fix

Two-layer defense matching cardano-node connection dedup behavior.

### Layer 1: Connmanager deduplication

In addConnectionImpl, when connections[connId] already exists:

- Existing outbound + new inbound: close the new inbound, return without registering. The outbound is the preferred connection. Log at WARN.
- Existing inbound + new outbound: close the existing inbound, allow the outbound to take its place. Log at INFO.
- Same direction (replacement after disconnect): allow overwrite as today.

### Layer 2: Chainsync callback immunity

Stop using connmanager.IsInboundConnection(connId) in chainsync client callbacks. Instead, record the connection direction when the chainsync client is started and check that recorded value in callbacks.

- Add StartedAsOutbound bool to chainsync.TrackedClient
- HandleOutboundConnEvent sets StartedAsOutbound = true
- HandleInboundConnEvent sets StartedAsOutbound = false
- New helper: Ouroboros.isInboundChainsyncClient(connId) checks the tracked client flag
- Replace the two IsInboundConnection calls in chainsyncClientRollBackward and chainsyncClientRollForward

### Files changed

- connmanager/connection_manager.go -- collision detection + dedup in addConnectionImpl
- chainsync/chainsync.go -- add StartedAsOutbound to TrackedClient, expose query method
- ouroboros/chainsync.go -- replace IsInboundConnection calls with tracked-client check
- ouroboros/ouroboros.go -- pass direction flag through registerTrackedChainsyncClient

### Testing

- Unit test: register outbound then inbound with same connId; verify outbound survives and inbound is closed
- Unit test: register inbound then outbound with same connId; verify outbound wins
- Unit test: chainsync tracked client direction flag preserves original value after connmanager collision
- Existing tests pass unchanged

This is a prerequisite for the broader inbound peer governance work (Phase 2: admission metadata and identity).

wolf31o2:
Closes #1912 

## Release Notes

* **Refactor**
  * Improved internal connection management by adding directional tracking for client connections, enabling better distinction between outbound and inbound initiated connections.
  * Enhanced collision handling for duplicate connection scenarios.

* **Tests**
  * Added comprehensive test coverage for connection direction tracking, collision detection, and event publishing during connection lifecycle.

---
## Summary by cubic
Fixes ConnectionId collisions that could overwrite live connections and misclassify chainsync direction. Outbound now wins collisions, chainsync records connection direction, and stale close events from evicted connections are suppressed.

- **Bug Fixes**
  - `connmanager`: Detect and resolve ConnectionId collisions.
    - Keep existing outbound when a new inbound collides; replace existing inbound when a new outbound collides; safely clean up metrics/IP slots and inbound peer counters; replace on same-direction collision.
    - Make `RemoveConnection` idempotent and return a bool; error watcher exits on false to prevent double cleanup.
    - Make `AddConnection`/`addConnectionWithIPKey`/`addNtCConnectionWithIPKey` return a bool to reject on shutdown/collision; listeners and outbound paths honor this and close or skip as needed.
    - Do not publish close events for evicted (stale) connections.
  - `chainsync`: Track connection direction at registration (`StartedAsOutbound`) and expose getters/setters.
    - Add direction-aware APIs: `TryAddClientConnIdWithDirection`, `TryAddObservedClientConnIdWithDirection`.
  - `ouroboros`: Use tracked direction (`isInboundChainsyncClient`) to ignore inbound chainsync clients and avoid feeding them into ledger; treat untracked clients as inbound. Pass direction when registering clients.

- **Migration**
  - Update callers of `connmanager.AddConnection`, `addConnectionWithIPKey`, and `addNtCConnectionWithIPKey` to check the returned bool and close the connection on false.
  - Update callers of `connmanager.RemoveConnection` to check the returned bool to avoid duplicate cleanup.

<sup>Written for commit da33b556be9e8295bbe1979047adee7ddb565b84. Summary will update on new commits.</sup>

### DIFF
--- a/chainsync/chainsync.go
+++ b/chainsync/chainsync.go
@@ -94,6 +94,12 @@ type TrackedClient struct {
 	// tip/activity metrics but must not consume the eligible
 	// client pool or become active for ledger ingress.
 	ObservabilityOnly bool
+	// StartedAsOutbound records whether the connection was
+	// initiated by us (outbound). This is stamped at
+	// registration time and can be refreshed only when the
+	// same ConnectionId is explicitly re-registered after a
+	// connmanager collision replacement.
+	StartedAsOutbound bool
 	LastActivity      time.Time
 	HeadersRecv       uint64
 	// TODO: BytesRecv needs to be wired to the underlying
@@ -358,11 +364,13 @@ func (s *State) promoteBestClientLocked() {
 func (s *State) addTrackedClientLocked(
 	connId ouroboros.ConnectionId,
 	observabilityOnly bool,
+	startedAsOutbound bool,
 ) {
 	s.trackedClients[connId] = &TrackedClient{
 		ConnId:            connId,
 		Status:            ClientStatusSyncing,
 		ObservabilityOnly: observabilityOnly,
+		StartedAsOutbound: startedAsOutbound,
 		LastActivity:      time.Now(),
 	}
 	// Set as active if there's no active client
@@ -389,10 +397,15 @@ func (s *State) addTrackedClientLocked(
 // Returns true if the client was added, false if rejected
 // (already tracked or at capacity). If no active client exists,
 // this connection is automatically set as the active client.
+// The client is recorded as outbound (StartedAsOutbound=true).
 func (s *State) AddClientConnId(
 	connId ouroboros.ConnectionId,
 ) bool {
-	return s.TryAddClientConnId(connId, s.config.MaxClients)
+	return s.TryAddClientConnIdWithDirection(
+		connId,
+		s.config.MaxClients,
+		true,
+	)
 }
 
 // TryAddObservedClientConnId adds a connection to observability-only tracking.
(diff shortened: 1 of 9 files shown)


wolf31o2:
@coderabbitai review

chrisguiney (approved):

PR #1922, opened 2026-04-16
fix(ledger): preserve in-flight blockfetch batch across chain switch

wcatz:
## Problem

When the active connection changed during a chain switch, `handoffPipelineOnSwitchLocked` unconditionally cancelled any in-flight blockfetch batch. This meant blocks that were mid-flight had to be re-requested from scratch on the new connection.

Under conditions where chain switches occur frequently — equal-tip peers alternating, peer reconnects — the pipeline would repeatedly cancel and restart batches faster than they could complete, causing a stall.

## Fix

Remove the cancel logic from `handoffPipelineOnSwitchLocked`. `selectedBlockfetchConnId` is already updated to the new connection at the point of the switch, so the **next** batch will naturally target the new peer. The in-flight batch is allowed to complete against its original connection.

This is safe because:
- The blocks being fetched are still valid (they were requested against a valid chain point)
- The ledger processes them normally on arrival regardless of which connection delivered them
- `selectedBlockfetchConnId` already ensures all subsequent batches use the new connection

## Impact

Eliminates pipeline stalls caused by repeated batch cancellation during chain switches. Particularly impactful on nodes with few peers (e.g. block producers) where equal-tip switching is frequent.

---
## Summary by cubic
Preserves in-flight blockfetch batches across chain switches so work isn’t canceled and restarted on every peer change. The current batch completes on the old peer; the next batch uses the new best peer to prevent stalls during equal-tip switching.

- **Bug Fixes**
  - Removed cancel logic in `handoffPipelineOnSwitchLocked`.
  - `selectedBlockfetchConnId` targets the next batch; `activeBlockfetchConnId` finishes the current one.
  - Updated `TestDetectConnectionSwitchHandsOffQueuedHeadersToNewActiveConnection` to assert no immediate restart (requestCount=0), old connection stays active, new connection is selected for the next batch, and one pending blockfetch event is queued.

<sup>Written for commit 4a28169eb9c122d5d5d570cc0fd9c8d1cff88f4e. Summary will update on new commits. <a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/1922?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced resilience of block fetching during connection switches by allowing in-flight operations to complete gracefully before transitioning to a new connection.


wolf31o2:
Better question would be why we're switching rapidly between near-identical peers. We shouldn't be doing equal-tip switching, at all.

wolf31o2 (approved):

PR #1924, opened 2026-04-16
fix(chainselection): use stable remote address for equal-tip tiebreaker

wcatz:
## Problem

`comparePeerTips` uses `ConnectionId.String()` as the final tiebreaker when two peers have chains that are otherwise equal (same block number, slot, VRF output). `ConnectionId.String()` formats as `localAddr<->remoteAddr`, e.g.:

```
192.168.9.127:42550<->35.188.35.224:3001
```

The local address includes the **ephemeral TCP port**, which changes on every reconnect. When a peer reconnects, its new `ConnectionId` may sort differently than the old one, flipping the tiebreaker result and triggering a chain switch.

The incumbent preservation in `evaluateBestPeerLocked` can't catch this because it looks up the *previous best* connection ID in `peerTips` — which is already gone after the reconnect. So the spurious switch goes through.

On nodes with only 2 peers (e.g. block producers) where both peers frequently reach the equal-tip tiebreaker, this produces continuous chain switches on every peer reconnect — pipeline churn with no benefit.

## Fix

Replace the full connection ID string with `RemoteAddr.String()` only. The remote address (`35.188.35.224:3001`) is stable across reconnects — it identifies the actual peer, not the transport-level connection. The tiebreaker result is now deterministic regardless of how many times a peer has reconnected.

Nil guards are included for the zero-value `ConnectionId` case, though in practice `RemoteAddr` is always set when a connection is active.

## Test update

`TestChainSelectorVRFTiebreakerWithNilVRF` comment updated to reflect that the fallback is now remote address ordering (result unchanged — test connection IDs have distinct remote ports so ordering is preserved).

---
## Summary by cubic
Use the peer’s stable `RemoteAddr` for the final equal-tip tiebreaker in `comparePeerTips` to stop chain flip-flops on reconnect. This makes ordering deterministic and prevents spurious switches on 2‑peer nodes.

- **Bug Fixes**
  - Compare `RemoteAddr.String()` instead of `ConnectionId.String()` when tips are otherwise equal.
  - Add nil guards for zero-value `ConnectionId`.
  - Clarify test comments for the remote-address tiebreaker (behavior unchanged).

<sup>Written for commit 5daf2765e41c73758927be772560824063b8f198. Summary will update on new commits.</sup>

### DIFF
--- a/chainselection/selector.go
+++ b/chainselection/selector.go
@@ -763,10 +763,22 @@ func (cs *ChainSelector) comparePeerTips(
 			return vrfComparison
 		case ChainEqual, ChainComparisonUnknown:
 		}
-		if connIdA.String() < connIdB.String() {
+		// Use stable remote address for final tiebreak rather than the full
+		// connection ID string. The full string includes the local ephemeral TCP
+		// port, which changes on every reconnect. That causes the sort order to
+		// flip whenever a peer reconnects — triggering spurious chain switches
+		// and pipeline stalls on resource-constrained nodes (e.g. Raspberry Pi).
+		var remoteA, remoteB string
+		if connIdA.RemoteAddr != nil {
+			remoteA = connIdA.RemoteAddr.String()
+		}
+		if connIdB.RemoteAddr != nil {
+			remoteB = connIdB.RemoteAddr.String()
+		}
+		if remoteA < remoteB {
 			return ChainABetter
 		}
-		if connIdB.String() < connIdA.String() {
+		if remoteB < remoteA {
 			return ChainBBetter
 		}
 		return ChainEqual
--- a/chainselection/selector_test.go
+++ b/chainselection/selector_test.go
@@ -1275,7 +1275,9 @@ func TestChainSelectorVRFTiebreakerWithNilVRF(t *testing.T) {
 	cs.UpdatePeerTip(connId2, tip, nil)
 
 	// When VRF comparison returns equal (due to nil), fall back to remote address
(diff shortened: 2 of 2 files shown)


wolf31o2:
This doesn't happen on a proper P2P Cardano network. The local connection port will be the same as the inbound port on the remote. It will be stable in the real world.

chainselection/selector.go
```diff
@@ -763,10 +763,22 @@ func (cs *ChainSelector) comparePeerTips(
 			return vrfComparison
 		case ChainEqual, ChainComparisonUnknown:
 		}
-		if connIdA.String() < connIdB.String() {
+		// Use stable remote address for final tiebreak rather than the full
+		// connection ID string. The full string includes the local ephemeral TCP
+		// port, which changes on every reconnect. That causes the sort order to
+		// flip whenever a peer reconnects — triggering spurious chain switches
```
wolf31o2, line 769: RemoteAddr is not a unique connection key. Two simultaneous connections to the same peer can share RemoteAddr while LocalAddr differs; this branch then returns ChainEqual and removes the previous deterministic tie-break. Preserve a unique stable key or add explicit handling and coverage.

wolf31o2 (changes requested):
The new fallback does not preserve a unique connection tie-break.

wcatz:
Closing this. wolf31o2's objection stands: on a proper P2P Cardano network the local connection port is the remote's inbound port and is stable in the real world, so the instability this works around is not the failure mode I took it for, and the fallback does not preserve a unique per-connection tie-break. If equal-tip selection turns out to need a stable discriminator, it should be a connection identity the peer governor already owns rather than an address heuristic.

PR #1929, opened 2026-04-17
feat(blockfetch): parallel shadow blockfetch with per-peer latency tracking

wcatz:
## Summary

- **Phase 1 — Per-peer latency EWMA**: Track time from `RequestRange` to first block response per connection using an exponential moving average (alpha=0.2). New methods on `chainsync.State`: `RecordBlockfetchLatency` and `BlockfetchLatency`.

- **Phase 2 — Latency tiebreaker in chain selection**: When two peers have equal observed tips (same block number), prefer the peer with lower measured blockfetch latency before falling back to connection ID string comparison.

- **Phase 3 — Shadow blockfetch dispatch**: For near-tip batches (≤4 headers), dispatch the same `RequestRange` to a second eligible peer (`PeersWithBlock` query against `observedHeaders`). Accept blocks from whichever peer responds first. Hash-based deduplication prevents double-apply; `BlockNotFitChainTipError` is already handled gracefully downstream.

- **Phase 4 — Node wiring**: Connect all new callbacks through `LedgerStateConfig` (`PeersWithBlockFunc`, `RecordBlockfetchLatencyFunc`) and `ChainSelectorConfig` (`BlockfetchLatency`) in `node.go`.

## Motivation

Block producers must receive new blocks as fast as possible to minimize time-to-forge. The existing single-peer blockfetch creates a latency bottleneck: if the active peer is slow or stalls, no fallback exists until a connection timeout fires. This change:

1. Races two peers for near-tip blocks (≤4-header batches) — the common case during active chain production
2. Records actual delivery latency so chain selection naturally gravitates toward faster peers
3. Preserves all existing behavior for bulk sync (large batches remain single-peer to avoid bandwidth waste)

## Test plan

- [x] `go build ./...` — clean
- [x] `go test ./chainselection/... ./chainsync/... ./ledger/...` — all pass
- [ ] Deploy to preview network nodes and observe shadow peer selection at tip
- [ ] Verify no duplicate block application errors in logs

---
## Summary by cubic
Add parallel shadow blockfetch near tip and per‑peer EWMA blockfetch latency tracking to cut first‑block latency and prefer faster peers. Also use latency as a tiebreaker when peer tips are equal.

- **New Features**
  - Track per‑peer first‑block EWMA latency (alpha=0.2) in `chainsync` via `RecordBlockfetchLatency`/`BlockfetchLatency`; wired in `node` through `LedgerStateConfig` and `ChainSelectorConfig`.
  - Use latency as a tiebreaker in `chainselection` when peers share the same observed tip and VRF.
  - Near tip (≤4 headers), dispatch a shadow `RequestRange` to a second peer; accept the first response, record the winner’s latency, and drop duplicates by hash. Bulk sync remains single‑peer.

- **Bug Fixes**
  - Accept `BatchDone` from either primary or shadow; if shadow wins, treat it as active for completion. Reset per‑batch shadow state on cleanup and fork restarts to avoid stale IDs.
  - Continue scanning shadow dispatch candidates when a request fails.

<sup>Written for commit a47975ce7df65672e9b50afe6a657fbe1f230071. Summary will update on new commits. <a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/1929?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **New Features**
  * Per-peer blockfetch latency tracking (EWMA) with query and recording hooks.
  * Peer selection can use lower observed latency as a deterministic tiebreaker when tips tie and samples exist.
  * Optional parallel ("shadow") blockfetch from a secondary peer near tip to improve resilience and recovery.
  * First-block latency measured once per batch and recorded.

* **Bug Fixes**
  * Per-batch deduplication of delivered blocks and improved cleanup on connection close.


wolf31o2 (dismissed):

wolf31o2 (changes requested):
Local review result: the minimum fix path before merge looks like this:

1. Let the shadow peer complete the active near-tip batch. Right now shadow blocks are accepted, but BatchDone is only accepted from activeBlockfetchConnId, so the <=4-header shadow path still waits on the slow primary or the timeout. Either treat shadow BatchDone as valid for the current batch, or transfer batch ownership when the first shadow block wins.

2. Clear shadow state on every restart path. restartQueuedBlockfetchAfterForkLocked manually tears down the old batch and re-enters startQueuedBlockfetchLocked without blockfetchRequestRangeCleanup(), so shadowBlockfetchConnId and related per-batch state can leak across a fork restart. Centralize restart through the normal cleanup path or explicitly reset shadowBlockfetchConnId, shadowBlockReceivedHashes, firstBlockReceived, and activeBlockfetchStart there.

3. Add regressions for both cases: shadow peer wins and completes the batch; fork restart does not accept stale shadow blocks from the previous batch.

I validated the concern locally in a temporary worktree. The touched package tests still pass, but these two state-machine holes are enough for me to keep this at no-go until fixed.

ledger/chainsync.go
```diff
@@ -1842,10 +1845,34 @@ func (ls *LedgerState) handleEventBlockfetchBlock(e BlockfetchEvent) error {
 	// chain promptly without paying a full blob transaction cost for
 	// every single block. We still flush well before BatchDone to
 	// avoid downstream ChainSync idle timeouts.
-	if ls.chainsyncBlockfetchReadyChan == nil ||
-		!sameConnectionId(e.ConnectionId, ls.activeBlockfetchConnId) {
+	if ls.chainsyncBlockfetchReadyChan == nil {
+		return nil
+	}
+	fromPrimary := sameConnectionId(e.ConnectionId, ls.activeBlockfetchConnId)
```
wolf31o2, line 1877: This is the new acceptance point for shadow-delivered blocks, but the batch still only completes when `BatchDone` arrives from `activeBlockfetchConnId`. In the intended `<=4` header case, that means the shadow peer can win the race on data and we still wait on the slow primary or the timeout. Minimum fix is to let the current shadow peer complete the batch too, or transfer batch ownership to the winner when the first shadow block arrives.

ledger/chainsync.go
```diff
@@ -2998,6 +3076,10 @@ func (ls *LedgerState) blockfetchRequestRangeCleanup() {
 	}
 	// Increment generation to ensure any pending timer callbacks are ignored
 	ls.chainsyncBlockfetchTimerGeneration++
+	// Clear shadow blockfetch state for the completed batch
```
wolf31o2, line 3184: This cleanup handles the new shadow fields for the normal batch-completion path, but `restartQueuedBlockfetchAfterForkLocked` tears down and restarts the batch without going through this helper. That lets `shadowBlockfetchConnId` and related per-batch state leak across a fork restart. Minimum fix is to funnel the restart path through the same cleanup, or explicitly clear the shadow fields there before reissuing the next batch.

wolf31o2:
@coderabbitai The finding misreads the design intent. The two fields are intentionally batch-scoped, not a per-connection bug:                                                                                                         
                                                                                                                                                                                                                           
  1. Primary and shadow are not "concurrent flows" — they race for the same batch.                                                                                                                                         
  At ledger/chainsync.go:2005-2046, startQueuedBlockfetchLocked dispatches the primary RequestRange, then dispatches the shadow RequestRange for the same headerStart/headerEnd. Both peers are racing to deliver the same 
  range. There is exactly one active batch at a time per LedgerState.                                                                                                                                                      
                                                                                                                                                                                                                           
  2. The latency sample is deliberately the winner's latency, not per-peer.                                                                                                                                                
  At ledger/chainsync.go:1894-1901, the first arriving block from either primary or shadow records the sample and attributes it to e.ConnectionId (the winner). The firstBlockReceived flag intentionally suppresses any
  further sample for this batch — that's the definition of "shadow blockfetch latency" (time-to-first-block among racing peers). Per-connection tracking would record latency for both peers, which is a behavior change,  
  not a bug fix.                                                       
                                                                                                                                                                                                                           
  3. There is no data race.                                            
  All accesses are serialized by chainsyncBlockfetchMutex:
  - Reads/writes in handleEventBlockfetchBlock (line 1895-1897) — caller handleEventBlockfetch holds the mutex (line 198-199).                                                                                             
  - Writes in startQueuedBlockfetchLocked (line 2005-2006) — *Locked suffix indicates the mutex is held by all callers; verified at lines 421, 578, 1348, 1606, etc.                                                       
  - Write in blockfetchRequestRangeCleanup (line 3187) — comment at line 1199 explicitly documents the mutex requirement.                                                                                                  
                                                                                                                                                                                                                           
  4. The existing field comment is accurate.                                                                                                                                                                               
  firstBlockReceived bool // true after latency sample recorded for this batch — "for this batch" is the correct scope. Replacing with map[connID]struct{...} would track per-peer flags that are never read per-peer,     
  since the code only checks "did any peer deliver yet."                                                                                                                                                                   
                                                                                                                                                                                                                           
  The proposed change would also introduce work that doesn't exist today: map allocation/cleanup on each batch, connection-id keying for state already keyed by "current batch." No issue to fix here.

wolf31o2 (approved):

PR #1983, opened 2026-04-20
fix: single-relay block producer connection hang

### LINKED ISSUE #1982: Single-relay block producer wedges when relay dials first after crash
## Summary

When a Dingo block producer has a single upstream Haskell relay, and the relay crashes and restarts while Dingo is in outbound reconnect backoff, Dingo accepts the relay's inbound dial but never catches tip again until it is restarted.

## Timeline observed in the field

- Outbound dial fails, peergov schedules retry with exponential backoff (seen as retry 280, delay ~2m8s)
- Relay comes back up and dials Dingo first, before Dingo's next retry fires
- Dingo accepts inbound. `ouroboros.HandleInboundConnEvent` starts chainsync and txsubmission client protocols on the full-duplex inbound connection
- Outbound retry eventually fires, sees an existing client-capable duplex, and stops trying (`peergov/connections.go:209`, from #1612)
- Txsubmission runs normally and begins rejecting transactions that reference UTxOs Dingo never synced (conway utxo validation rule 22/24), which masks the fact that chainsync is silent

## Root cause

The inbound chainsync client is registered with `ingressEligible=false` in `ouroboros/ouroboros.go:541`. On every header received, `ouroboros/chainsync.go:527-606` hard-codes `ingressEligible = false` for any inbound-started client, which means:

- `PeerTipUpdateEvent` is not published, so `chainSelector.GetBestPeer()` returns nil
- The header is dropped as "not ingress eligible" and `ChainsyncEventType` is never published to the ledger
- `promoteBestClientLocked` in `chainsync/chainsync.go:332` skips observability-only clients, so `activeClientConnId` stays nil
- The plateau recycler in `node_chainsync_recycler.go` keys off `chainSelector.GetBestPeer()` and cannot recover

Restart fixes it because Dingo boots outbound-first, wins the dial race, and chainsync is registered with `startedAsOutbound=true, ingressEligible=true`.

This is the composition of two separately correct fixes. #1612 suppresses outbound when a duplex inbound exists. #1699 filters inbound peers from chain selection so random downstream peers do not pollute it. The failure mode is specific to a configured upstream peer winning the inbound dial race.

## Fix plan

Upgrade an inbound full-duplex chainsync to ingress-eligible when the remote address matches a configured outbound peer (one that peergov is actively managing as an upstream). This preserves the #1699 guarantee for random downstream peers while fixing the single-upstream case.

## Reproduction

1. Run Dingo as a block producer with a single upstream relay in topology
2. Let Dingo connect and sync to tip
3. Kill the relay, wait for Dingo's outbound retry backoff to grow to tens of seconds or more
4. Restart the relay before Dingo's next retry fires so the relay wins the dial race
5. Observe that Dingo accepts inbound, txsubmission stays live, but chain tip does not advance until Dingo is restarted

wolf31o2:
Closes #1982 

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized peer connection eligibility validation for chain synchronization, ensuring consistent handling of inbound peer connections during rollback and roll-forward operations across the system.

* **Tests**
  * Added comprehensive test coverage for chain synchronization events with inbound peer connections, validating proper event publication behavior for both eligible and ineligible connection states.

---
## Summary by cubic
Fixes a stall where a single-relay block producer stops syncing when the relay dials us first and chainsync runs inbound. Eligible inbound upstream peers now feed the ledger and chain selection; random inbound peers remain observability-only. Addresses #1982.

- **Bug Fixes**
  - `ouroboros`: Gate ledger and `PeerTipUpdateEvent` on ingress eligibility for roll-forward and rollback; remove inbound short-circuits. Introduce `shouldPublishChainsyncToLedger` fallback: when no policy is configured, outbound defaults to eligible and inbound defaults to observability-only; apply this during inbound/outbound registration.
  - `peergov`: Update `chainSelectionEligible` to reject `PeerSourceInboundConn` while keeping topology/P2P-discovered peers eligible even if the connection is inbound.
  - Tests: Add coverage for eligible inbound roll-forward/rollback publishing, ineligible inbound peers staying observability-only, and nil-policy inbound failing closed.

<sup>Written for commit 8723e95b4ce28dc3b57a1f034332293f3220ddf0. Summary will update on new commits.</sup>

### DIFF
--- a/ouroboros/chainsync.go
+++ b/ouroboros/chainsync.go
@@ -481,9 +481,6 @@ func (o *Ouroboros) chainsyncClientRollBackward(
 	point ocommon.Point,
 	tip ochainsync.Tip,
 ) error {
-	if o.isInboundChainsyncClient(ctx.ConnectionId) {
-		return nil
-	}
 	if !o.reconcileChainsyncIngressAdmission(
 		ctx.ConnectionId,
 		o.shouldPublishChainsyncToLedger(ctx.ConnectionId),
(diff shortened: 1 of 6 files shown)


wolf31o2:
@coderabbitai review

chrisguiney (approved):

PR #2015, opened 2026-04-23
fix(ledger): per-peer loop detection

wolf31o2:
## Summary by cubic
Make rollback loop detection per-peer and per-point to avoid suppressing valid rollbacks from other peers and to stop repeat loops from the same peer. This improves chain selection stability and prevents oscillation during sync.

- **Bug Fixes**
  - Track rollback history by exact point (slot + hash) and connection key; skip only when the same peer repeats the same point within the window.
  - Preserve history across connection switches without poisoning other peers.
  - Added tests for cross-peer vs same-peer behavior and updated the rollback record structure.

<sup>Written for commit a0a557fad6b61bf0b9b293fd12e19ec2aaadf334. Summary will update on new commits.</sup>

### DIFF
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -482,10 +482,10 @@ func (ls *LedgerState) detectConnectionSwitch() (
 		}
 		ls.lastActiveConnId = activeConnId
 		// Preserve rollbackHistory across connection switches so the
-		// rollback loop detector can fire when multiple peers all send
-		// RollBackward to the same slot during rapid chain selection
-		// changes (e.g., post-Mithril startup). Clearing it here
-		// previously allowed unbounded oscillation.
+		// loop detector can still catch repeated rollbacks from the
+		// same peer/session. The detector keys on exact rollback
+		// point + connection, so peer switches do not poison healthy
+		// fork convergence.
 	}
 	return activeConnId, true
 }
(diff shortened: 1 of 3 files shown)


ledger/chainsync.go
```diff
@@ -1002,10 +1002,16 @@ func (ls *LedgerState) handleEventChainsyncRollback(e ChainsyncEvent) error {
...
-	// the same slot appears too frequently within the detection window.
+	// the same peer repeats the same rollback point too frequently
+	// within the detection window.
 	now := time.Now()
+	connKey := connIdKey(e.ConnectionId)
 	ls.rollbackHistory = append(ls.rollbackHistory, rollbackRecord{
-		slot:      e.Point.Slot,
+		point: ocommon.Point{
+			Slot: e.Point.Slot,
+			Hash: append([]byte(nil), e.Point.Hash...),
```
chrisguiney, line 1012: nit: `slices.Clone(e.Point.Hash)`

ledger/chainsync_rollback_test.go
```diff
@@ -518,7 +582,11 @@ func TestRecoverAfterLocalRollbackResetsStateWithoutTrackedClients(t *testing.T)
 	fixture.ls.headerMismatchCount = 3
 	fixture.ls.rollbackHistory = []rollbackRecord{
 		{
-			slot:      fixture.currentTip.Point.Slot,
+			point: ocommon.Point{
+				Slot: fixture.currentTip.Point.Slot,
+				Hash: append([]byte(nil), fixture.currentTip.Point.Hash...),
```
chrisguiney, line 587: same nit as before -- `slices.Clone`

chrisguiney (approved):

ledger/chainsync.go
wolf31o2: I usually catch this with another `modernize` pass

PR #2017, opened 2026-04-23
feat(peergov): admission metadata and identitiy

### LINKED ISSUE #1932: peergov: inbound peer governance phase 2 (admission metadata and identity)
## Summary

Preserve the metadata needed for good inbound peer decisions. Phase 1 (#1931) gives inbound peers an explicit budget; phase 2 ensures the peer governor has enough context about each inbound peer to govern on quality rather than mere existence.

This phase also introduces safe identity rules for matching inbound connections to configured topology peers — important for the operator pattern where two relays are explicitly configured as known peers and either side may initiate the connection.

## What Needs to Happen

1. Record inbound-specific metadata needed by later phases (promotion, pruning):
   - duplex capability (whether the connection is reusable for outbound mini-protocols)
   - peer identity material carried from connmanager into peergov
   - first-seen timestamp / tenure
   - whether the peer is reaching us from a configured topology address
2. Carry duplex capability and peer identity cleanly from connection manager into peer governor — today these are tracked in disconnected layers.
3. Distinguish repeated arrivals from genuinely new inbound peers where possible (so flapping peers can be detected later).
4. Define safe topology-peer identity rules for inbound reuse:
   - exact peer-address match first
   - then unambiguous topology-host match only when it cannot collapse distinct configured peers
   - guard against host-level matching that is too loose (would merge distinct peers on the same host into one topology identity)
5. Avoid treating every short-lived inbound connection as a meaningful peer candidate.

## Operator Pattern To Support

Two relays are operator-owned and explicitly configured as known peers. Either side should be allowed to initiate the connection. If a configured topology peer reaches us inbound from a different source port than the configured listen port, the inbound connection should still be matchable to the topology peer (subject to the safety rules above). Today the reusable-connection logic is centered on exact remote peer address matching, which is safe for TCP 4-tuple collision avoidance but does not capture this operator policy.

## Success Criteria

- Peer governor has enough inbound context to govern on quality rather than existence.
- Inbound connections from known topology peers can be safely matched to their topology identity, without merging distinct peers that share a host.

## Depends On

- #1931 (phase 1 — explicit budget and state model, milestone Q2 2026)

## Roadmap

Q2 2026 — networking and peer quality.

cryptodj413:
Closes #1932 

## Summary by CodeRabbit

* **New Features**
  * Exported peer-address normalization and host-only identity resolution for consistent inbound identity.
  * Inbound connection events include normalized remote address and provisional duplex metadata.
  * Persisted peer fields for inbound arrivals, last-arrival timestamp, topology match, and duplex; new inbound provisional window config.
  * New metrics for inbound arrivals, topology-matched, and duplex-held counts.

* **Refactor**
  * Centralized duplex detection and consolidated provisional-window census for inbound quota/metrics.

* **Tests**
  * Extensive phase-2 tests for inbound identity, arrivals, provisional-window census, and duplex behavior.

---
## Summary by cubic
Adds phase 2 inbound admission metadata and identity to `peergov`, and aligns `connmanager` so inbound peers use a shared, canonical transport key. Introduces a short provisional window to filter flash connects (per #1932) and expands metrics and event fields for better visibility.

- **New Features**
  - Canonical identity: export `NormalizePeerAddr`; `InboundConnectionEvent` adds `NormalizedRemoteAddr` and `IsDuplex`; listener populates both; `peergov` falls back to `NormalizePeerAddr` if missing.
  - Inbound identity: resolve to an existing peer by exact address, else by a safe, unambiguous topology host match; record the matched topology `GroupID`.
  - Per-peer admission metadata: track `InboundDuplex` (best-effort from event, authoritative from handshake), `InboundArrivals`, and `LastInboundArrival`; increment on re-arrivals.
  - Provisional window: `InboundProvisionalWindow` (default 3s; negative disables) excludes fresh inbound peers from Warm/Hot counts; applied to Prometheus gauges and `QuotaStatusEvent`.
  - Metrics/events: add `InboundArrivalsTotal`, `InboundTopologyMatched`, `InboundDuplexHeld`; `QuotaStatusEvent` now reports `InboundTopologyMatched` and `InboundDuplex` using the same filtered inbound census as Prometheus.

- **Refactors**
  - Central inbound census feeds both Prometheus gauges and `QuotaStatusEvent` (with provisional-window filtering) to keep views in sync.
  - Listener detects duplex via `connectionIsDuplex`; inbound normalization updated across `connmanager`; avoid replacing a client-capable connection with a responder-only inbound.

<sup>Written for commit 51a441e00877ff41fdca8b87db6e79a3af60d699. Summary will update on new commits.</sup>

### DIFF
--- a/peergov/connections.go
+++ b/peergov/connections.go
@@ -525,9 +525,12 @@ func (p *PeerGovernor) handleInboundConnectionEvent(evt event.Event) {
 	}
 	oldSource := tmpPeer.Source
 	oldConn := clonePeerConnection(tmpPeer.Connection)
-	// Accept the event-embedded duplex hint as a provisional value; the
-	// connmanager lookup below is authoritative when present.
-	tmpPeer.InboundDuplex = e.IsDuplex
+	// Accept an event-embedded duplex=true hint as a provisional upgrade;
+	// do not clear a previously known true value on best-effort false.
+	// The connmanager lookup below is authoritative when present.
+	if e.IsDuplex {
+		tmpPeer.InboundDuplex = true
+	}
 	if p.config.ConnManager != nil {
 		conn := p.config.ConnManager.GetConnectionById(e.ConnectionId)
 		if conn != nil {
--- a/peergov/connections.go
+++ b/peergov/connections.go
@@ -464,23 +464,22 @@ func (p *PeerGovernor) handleInboundConnectionEvent(evt event.Event) {
 		return
 	}
 	address := e.RemoteAddr.String()
-	// Resolve address before acquiring lock to avoid blocking DNS
-	normalized := p.resolveAddress(address)
+	// Key on the connmanager's canonical transport identity so inbound
+	// peer entries share a key space with HasInboundPeerAddress and
+	// inboundPeerAddrs. Never DNS-resolve the inbound remote: the
+	// 4-tuple is an IP:port and calling resolveAddress would diverge
+	// from the connmanager's non-DNS normalization. When an event is
+	// synthesized in tests without a populated NormalizedRemoteAddr,
+	// fall back to NormalizePeerAddr to preserve the same contract.
+	normalized := e.NormalizedRemoteAddr
+	if normalized == "" {
+		normalized = connmanager.NormalizePeerAddr(address)
+	}
+	now := time.Now()
 
 	var selectionEvents []pendingEvent
 	p.mu.Lock()
-	peerIdx := -1
-	// Check if peer already exists (possibly as inbound)
-	for i, peer := range p.peers {
-		if peer == nil {
-			continue
-		}
-		// Match by exact address or normalized address
-		if peer.Address == address || peer.NormalizedAddress == normalized {
-			peerIdx = i
-			break
-		}
-	}
+	peerIdx, topologyGroupID := p.resolveInboundIdentity(address, normalized)
 	var tmpPeer *Peer
 	if peerIdx == -1 {
 		// Enforce hard cap on peer list size for inbound peers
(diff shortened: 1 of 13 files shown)


peergov/connections.go
```diff
@@ -464,23 +464,22 @@ func (p *PeerGovernor) handleInboundConnectionEvent(evt event.Event) {
...
-		if peer == nil {
-			continue
-		}
-		// Match by exact address or normalized address
-		if peer.Address == address || peer.NormalizedAddress == normalized {
-			peerIdx = i
-			break
-		}
-	}
+	peerIdx, topologyGroupID := p.resolveInboundIdentity(address, normalized)
```
wolf31o2, line 482: Attaching an inbound ephemeral connection to an existing topology peer record is fine, but this path then rebinds `tmpPeer.Connection` to `e.ConnectionId` unconditionally below. That lets a responder-only inbound replace an already client-capable outbound connection record, even though `createOutboundConnection()` explicitly says only client-capable connections should suppress the outbound dial. I reproduced that locally with a temporary test. This needs a guard so a non-reusable inbound does not displace an authoritative client-capable connection.

peergov/metrics.go
```diff
@@ -181,6 +187,81 @@ func (p *PeerGovernor) initMetrics() {
...
+	now := time.Now()
+	window := p.config.InboundProvisionalWindow
+	for _, peer := range p.peers {
+		if peer == nil {
+			continue
+		}
+		if peer.InboundTopologyMatch != "" {
+			c.TopologyMatched++
+		}
+		if peer.InboundDuplex {
```
wolf31o2: `InboundDuplexHeld` / `QuotaStatusEvent.InboundDuplex` are documented as current counts, but this census increments `Duplex` from the sticky `peer.InboundDuplex` bit and `handleConnectionClosedEvent()` never clears that field. A peer that was duplex once continues to be counted after its inbound connection is gone. I reproduced that locally with a temporary test. Either clear `InboundDuplex` on close or derive this metric/event from live connection state instead of the retained admission flag.

wolf31o2 (changes requested):
Local review result: no-go for merge as-is.

1. `peergov` can replace an already client-capable topology connection with a responder-only inbound when the new host-match path attaches an ephemeral inbound arrival to an existing topology peer record. That conflicts with the existing duplex-reuse rule and can leave peergov tracking the wrong connection.

2. The new duplex observability surfaces overcount disconnected peers because `InboundDuplex` is retained after close but `InboundDuplexHeld` / `QuotaStatusEvent.InboundDuplex` are documented as current counts.

I reviewed and tested this locally in `/tmp/dingo-pr-2017`. `GOWORK=off GOCACHE=/tmp/dingo-pr-2017-gocache go test ./connmanager ./peergov` passes, and I also validated both findings with temporary local repro tests.

wolf31o2 (approved):

PR #2043, opened 2026-04-25
fix(ledger): epoch switch rollover safe nonce calc

wolf31o2:
## Release Notes

* **Bug Fixes**
  * Improved epoch derivation in block forging to use slot-based calculation instead of current ledger epoch.
  * Enhanced error handling for epoch resolution failures during block creation.
  * Refined epoch nonce computation, particularly at epoch boundaries.

* **Tests**
  * Added comprehensive test coverage for slot-based epoch derivation and nonce selection edge cases.

---
## Summary by cubic
Fixes leader election at epoch switch by using the block slot’s epoch nonce and makes next-epoch nonce calculation rollover-safe. Prevents wrong VRF nonce usage and stabilizes forging across epoch boundaries.

- **Bug Fixes**
  - `DefaultBlockBuilder` resolves the block slot’s epoch via `EpochForSlot(slot)` and uses that epoch’s nonce for VRF; if epoch resolution fails, it now returns an error.
  - Ledger `computeNextEpochNonce` computes the next-epoch nonce at the next epoch’s start via `computeEpochNonceForSlot`, honoring the imported tip anchor and removing fragile manual mixing. Tests added for slot-epoch selection, tip-anchor handling, and error propagation.
  - `EpochNonceProvider` and the `node_forging` adapter updated to support `EpochForSlot` (adapter calls `SlotToEpoch`).

- **Migration**
  - Add `EpochForSlot(slot uint64) (uint64, error)` to any custom `EpochNonceProvider` implementations.

<sup>Written for commit 40351c856854523d8299b5f71aab2cf67c17a713. Summary will update on new commits.</sup>

### DIFF
--- a/ledger/forging/builder.go
+++ b/ledger/forging/builder.go
@@ -58,6 +58,8 @@ type ChainTipProvider interface {
 type EpochNonceProvider interface {
 	// CurrentEpoch returns the current epoch number.
 	CurrentEpoch() uint64
+	// EpochForSlot returns the epoch containing the given slot.
+	EpochForSlot(slot uint64) (uint64, error)
 	// EpochNonce returns the nonce for the given epoch.
 	EpochNonce(epoch uint64) []byte
 }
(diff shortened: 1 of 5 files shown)


wolf31o2:
@cubic-dev-ai review

wcatz (approved):

PR #2079, opened 2026-04-28
feat: mid-epoch transition stability

chrisguiney:
contributes to #1857 

## Summary by cubic
Surfaces upcoming era transitions mid-epoch by detecting ratifiable HardForkInitiation proposals after the voting deadline. Sets TransitionKnown for the next epoch so clients see the transition earlier and avoid transient Unknown after rollbacks.

- New Features
  - Added `ledger/governance/stability.go` with StabilityCheckInputs and EvaluateRatifiableHardForkInitiation. Uses live tallies, active DReps, committee quorum, and parent-chain lineage; pre-Conway pparams return nil.
  - Integrated `evaluateHardForkInitiationStability` in `ledger/state.go` (startup, post-epoch checks, and after rollback). Runs only after the voting deadline (epochEnd - 2*stabilityWindow), upgrades Unknown/Impossible to Known, and never overrides an existing Known.
  - Added tests for bootstrap semantics, non-HFI filtering, pre-/post-deadline behavior, pre-Conway short-circuit, idempotency, and priority over Impossible or external Known.

<sup>Written for commit 721936b1f6ab9d7aa64cf83a16cb4d90c50f779d. Summary will update on new commits. <a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2079?utm_source=github">Review in cubic</a></sup>

## Release Notes

* **New Features**
  * Proactive hard fork stability detection that evaluates active hard fork proposals during in-epoch processing and determines their ratification likelihood based on current governance voting state.
  * Early notification of upcoming protocol upgrades before epoch boundaries.

* **Tests**
  * Comprehensive test coverage for hard fork stability detection, including edge cases and protocol version-dependent governance logic.

### DIFF
--- a/ledger/governance/epoch.go
+++ b/ledger/governance/epoch.go
@@ -185,6 +185,16 @@ func ProcessEpoch(
 	}
 
 	// --- RATIFICATION -------------------------------------------------
+	//
+	// The inputs assembled below (TallyContext, activeDRepCount,
+	// rootsByPurpose, committeeState, ccQuorum, conwayPParams,
+	// majorVersion, ccInNoConfidence) feed ShouldRatify. A parallel
+	// build for HardForkInitiation specifically exists in
+	// EvaluateRatifiableHardForkInitiation (governance/stability.go),
+	// which runs the same check mid-epoch to surface upcoming
+	// transitions before the boundary tick fires. Adding a new
+	// ratification input here without updating the mid-epoch path
+	// will silently make the two answers diverge — keep them in sync.
 	tallyCtx := &TallyContext{
 		DB:           in.DB,
 		Txn:          in.Txn,
@@ -385,6 +395,10 @@ func stakeEpochFor(newEpoch uint64) uint64 {
 // countActiveDReps returns the number of DReps eligible to vote in the
 // current tick. AlwaysAbstain / AlwaysNoConfidence virtual DReps are
 // not included.
+//
+// countActiveDRepsAtEpoch in stability.go is the parallel for the
+// mid-epoch ratifiability check — it takes plain inputs instead of
+// EpochInput. Behaviour changes here must mirror there.
 func countActiveDReps(in *EpochInput) (int, error) {
 	dreps, err := in.DB.GetActiveDreps(in.Txn)
 	if err != nil {
--- a/ledger/governance/stability.go
+++ b/ledger/governance/stability.go
@@ -1,0 +1,224 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package governance
+
+import (
+	"errors"
+	"fmt"
+	"log/slog"
+
+	"github.com/blinklabs-io/dingo/database"
+	"github.com/blinklabs-io/dingo/database/models"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/conway"
+)
+
+// StabilityCheckInputs collects the state needed to determine whether
+// any active HardForkInitiation proposal would ratify if the boundary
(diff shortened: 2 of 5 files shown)


chrisguiney:
## Review

### Summary
Adds proactive detection of upcoming era transitions during epoch processing, rather than waiting for the boundary tick. New `ledger/governance/stability.go` runs the same tally + threshold logic as the boundary RATIFY phase to determine whether any in-flight `HardForkInitiation` would ratify *now*; if so, sets `TransitionKnown(currentEpoch+1)` so clients see the transition earlier. Wired into three call sites in `ledger/state.go`: `Start`, post-rollback, and post-epoch tip update.

### What's good
- **Documentation** — every non-obvious decision (why duplicate ProcessEpoch's setup, why post-deadline only, the priority ordering, lock semantics) has a *why* comment. Future maintainers won't have to re-derive this.
- **Correctness rationale is sound**: post-`epochEnd - 2*stabilityWindow` the inputs to ratification are frozen (no new votes/regs/pool changes), so "would ratify now" ≡ "will ratify at boundary tick".
- **Defensive arithmetic**: `checkedSlotAdd` for overflow, `if epochEndSlot > deadlineGap` for synthetic small-epoch cases.
- **Priority order**: Known supersedes Unknown/Impossible; never overrides an existing Known. `TestEvaluateHardForkInitiationStability_PreservesKnownFromOtherSource` pins this so a `TestXHardForkAtEpoch` override can't be silently clobbered.
- **Tests are thorough** — pre/post deadline, pre-Conway short-circuit, action-type filtering, bootstrap positive/negative, idempotency, priority. Each name pins the invariant it protects.

### Concerns

**1. Drift risk from duplicated RATIFY inputs (medium).**
Acknowledged in `stability.go:34-40`, but worth weighing: if `ProcessEpoch` ever gains a new ratification input (new pparam, new committee dimension), this path will silently miss it and mid-epoch will diverge from the boundary answer. At minimum, leave a breadcrumb in the boundary path noting the parallel build that must stay in sync. Better: a shared "tally + decide" helper that both call sites use, even if input gathering stays separate.

**2. `countActiveDRepsAtEpoch` parallels `countActiveDReps` (low-medium).**
Same concern in miniature. Two functions doing the same thing with different input shapes — a behavior change to one could leave this version stale. Either share a core or note the relationship so a grep for "active DRep" lands on both.

**3. `RatifiableHardForkInitiation.NewMajor` / `NewMinor` are computed but unused by the caller (low).**
`evaluateHardForkInitiationStability` only uses `result != nil` and `currentEpoch+1`. The decode step in `stability.go:183-204` populates fields nobody reads. Either drop the decode (and the `NewMajor`/`NewMinor` fields) or use them in the `TransitionKnown` payload / logging. As-is it's wasted work and a future bug magnet.

**4. Per-block invocation cost (low, bounded but worth noting).**
Runs after every epoch tip update. The pre-deadline guard and the already-Known short-circuit bound the work, but in the post-deadline-pre-Known window every block triggers `GetActiveGovernanceProposals` + `GetActiveDreps` + `LoadCommitteeVotingState` + `GetLastEnactedGovernanceProposal` ×2 + `conwayRatifyQuorum`. For a long voting-deadline window with no ratifiable proposal that's a non-trivial repeat read. Consider memoizing the "no ratifiable proposal at slot ≥ X" answer until a new vote arrives or the epoch advances. Not a blocker.

**5. Lock semantics comment vs. actual call sites (verify).**
`state.go:566` says "Call under `ls.Lock()` (runtime tip-update) or without a lock during single-threaded startup." The rollback site (`state.go:525`) needs verification that this holds, and the doc could be more specific about which sites satisfy which precondition.

**6. Multiple ratifiable proposals — selection is implicit (low).**
The loop at `stability.go:158` returns the first in `GetActiveGovernanceProposals`'s iteration order. Tests don't cover the multi-ratifiable case. If two HFIs are simultaneously ratifiable, the boundary path's selection rule should match this. Worth a test that pins "two ratifiable proposals → returns the same one the boundary path would enact."

**7. Decode-failure handling masks data-integrity bugs (low).**
`stability.go:183-195` logs a warning and continues on decode failure. The comment correctly identifies this as a data-integrity issue. Consider a metric/counter so it doesn't get lost in log volume, and include the txhash in the warning so the bad row is findable.

### Nits
- `Logger != nil` checks suggest the field is optional; either require it or document the no-logger path.
- Test stake amounts of `1_000` are fine for unit semantics but don't exercise large-number edge cases. Probably out of scope.

### Decision: Approve with comments
Core change is correct, well-reasoned, well-tested. Concerns 1 and 3 are the two I'd want addressed before merge — #1 with at least a cross-reference comment in the boundary path, #3 by either using or removing the decoded version fields. The rest can land as follow-ups.

wolf31o2 (approved):

PR #2032, opened 2026-04-24
fix: repair mithril snapshot bootstrap correctness and resume handling

wolf31o2:
Closes #2019 
Closes #2027 

---
## Summary by cubic
Repairs Mithril snapshot bootstrap and resume so restarts are safe and fast. Persists per-output UTxO offsets during `ImmutableDB` copy in the same transaction, replays gap blocks with epoch anchoring for Conway governance, and only runs historical backfill in API mode when a checkpoint exists.

- **Bug Fixes**
  - Recover missing consumed UTxOs for normal and gap txs; record spender `spent_at_tx_id`; repair same-slot rows via updated `SetUtxoDeletedAtSlot`; restore using UTxO blob offsets or producer tx slots via `GetTransactionSlotByHash`.
  - Process Conway governance in gap blocks; handle votes-before-proposal; anchor proposals, PParams, constitution, and committee to epoch-start slots via `GetEpochBySlot`.
  - Compute produced-output offsets by structurally walking the CBOR tx body (incl. `collateral_return`) to avoid false byte matches.
  - `TieredCborCache.ResolveTxCbor` now reads via the caller’s blob transaction so uncommitted writes (e.g., during immutable copy) are visible.
  - Skip immutable copy when chain tip is past the `ImmutableDB` tip; decode binary TxIn indices as big-endian.
  - Make block-nonce writes idempotent with upsert on unique `(hash, slot)`, migrate the legacy non-unique index, and preserve `is_checkpoint=true` on upsert.

- **New Features**
  - Add `Chain.AddRawBlocksWithCallback` to run per-block callbacks in the same transaction during raw import; used to persist UTxO offset indexes during `ImmutableDB` copy.
  - Add `GetEpochBySlot`, `GetTransactionSlotByHash`, and `GetTransactionIDByHash`; clearer serve/backfill logs; metadata backfill runs only in API mode when a checkpoint exists.

<sup>Written for commit 54741bd8791935d22202f3768cb3631f92d27284. Summary will update on new commits. <a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2032?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **New Features**
  * Per-block callbacks during atomic block ingestion for immediate UTxO-offset persistence and safer header-only imports
  * Epoch-aware gap-block processing with governance indexing and automatic proposal repair
  * UTxO records now include spender transaction hashes and a slot→epoch lookup

* **Bug Fixes**
[...]

PR #2083, opened 2026-04-28
feat: add SetTransactionBatched to sqlite/mysql/postgres

### LINKED ISSUE #1799: Add SetTransactionBatched to MetadataStoreSqlite
## Summary

Create a variant of `SetTransaction` that accumulates metadata into a `BatchAccumulator` instead of writing immediately. Same model-building logic, but rows go to the accumulator for later bulk flush.

Part of the backfill batching optimization.

## Design

`SetTransactionBatched` does the same work as `SetTransaction` with these differences:

| Operation | SetTransaction | SetTransactionBatched |
|-----------|---------------|----------------------|
| Transaction record (upsert) | Immediate | Immediate (needs auto-increment ID for FKs) |
| UTxO outputs | Per-item `db.Create` | `acc.AddUtxoOutput()` |
| UTxO spends | Per-input UPDATE | `acc.AddUtxoSpend()` |
| Witnesses | Per-item `db.Create` | `acc.AddKeyWitness()` |
| Scripts | Per-item `db.Create` | `acc.AddScript()` + `acc.AddWitnessScript()` |
| Datums | Per-item `db.Create` | `acc.AddPlutusData()` |
| Redeemers | Per-item `db.Create` | `acc.AddRedeemer()` |
| Address TXs | Per-item `db.Create` | `acc.AddAddressTx()` |
| Certificates | Per-TX immediate | Per-TX immediate (17+ type switch, already fast) |
| Governance | Per-TX immediate | Per-TX immediate |

The transaction record must be written immediately because witnesses, scripts, and other records reference the auto-increment `Transaction.ID` as a foreign key.

### `database/plugin/metadata/sqlite/transaction.go`

Add `SetTransactionBatched` method. Extract shared model-building logic from `SetTransaction` where practical to avoid duplication, or inline if the paths diverge too much.

### Tests (`database/plugin/metadata/sqlite/batch_accumulator_test.go`)

- `TestSetTransactionBatched_AccumulatesCorrectly` — call with a mock transaction, verify accumulator has expected witness/output/spend counts, flush and verify DB state

## Deliverables

- [ ] `SetTransactionBatched` method on `MetadataStoreSqlite`
- [ ] Transaction record written immediately for FK dependency
- [ ] All other metadata accumulated into `BatchAccumulator`
- [ ] Test verifying accumulation counts and flush correctness
- [ ] Existing `SetTransaction` behavior unchanged
- [ ] All existing tests pass

jkawan:
closes #1799 

---
## Summary by cubic
Adds `SetTransactionBatched` to `sqlite`, `mysql`, and `postgres` to batch per-item metadata writes while still upserting the transaction row immediately for FK safety. Also adds `MergeFrom` to combine accumulators before a `FlushBatch`.

- **New Features**
  - `SetTransactionBatched` + accumulator/`FlushBatch` API across all backends.
  - Immediate writes: transaction upsert; collateral/reference-input markers; certificates/governance; datum-hash index.
  - Deferred writes: UTxO outputs/collateral return/spends; witnesses/scripts/plutus data/redeemers; address-transaction index.
  - `MergeFrom` to combine accumulators before flushing.

- **Bug Fixes**
  - Set `AddedSlot` consistently for accounts during certificate processing in `sqlite`/`postgres` to avoid stale slot timestamps on updates.

<sup>Written for commit e6fb662e4d5ed94074f7a76740434494f5e6bc9d. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Refactor**
  * Batched transaction persistence added across backends to defer per-item metadata writes, queue UTxO outputs/spends and witness-related rows for bulk flush, and separate collateral-return handling.
  * Added ability to merge deferred batches before flush.

* **Bug Fixes**
  * Improved idempotency and retry handling when a transaction ID is discovered after initial upsert to avoid duplicate/orphaned rows.

* **Tests**
  * Added test verifying immediate transaction row persistence with deferred metadata accumulation and flush.

### DIFF
--- a/database/plugin/metadata/mysql/transaction.go
+++ b/database/plugin/metadata/mysql/transaction.go
@@ -3255,7 +3255,7 @@ func (d *MetadataStoreMysql) SetTransactionBatched(
 				}
 				caseStmt := strings.Join(whenClauses, " ")
 				query := fmt.Sprintf(
-					"UPDATE certs SET certificate_id = CASE %s END WHERE id IN (?)",
+					"UPDATE certs SET certificate_id = CASE %s END WHERE id IN ?",
 					caseStmt,
 				)
 				values = append(values, ids)
--- a/database/plugin/metadata/mysql/transaction.go
+++ b/database/plugin/metadata/mysql/transaction.go
@@ -2075,13 +2075,21 @@ func (d *MetadataStoreMysql) SetTransactionBatched(
 	// ------------------------------------------------------------------ //
 	// 1. Write transaction record immediately (needed for FK IDs)         //
 	// ------------------------------------------------------------------ //
+	var feeUint uint64
+	if txFee := tx.Fee(); txFee != nil {
+		if txFee.BitLen() > 64 {
+			feeUint = math.MaxUint64
+		} else {
+			feeUint = txFee.Uint64()
+		}
+	}
 	tmpTx := &models.Transaction{
 		Hash:       txHash,
 		Type:       tx.Type(),
 		BlockHash:  point.Hash,
 		BlockIndex: idx,
 		Slot:       point.Slot,
-		Fee:        types.Uint64(tx.Fee().Uint64()),
+		Fee:        types.Uint64(feeUint),
 		TTL:        types.Uint64(tx.TTL()),
 		Valid:      tx.IsValid(),
 	}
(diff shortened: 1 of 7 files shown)


database/plugin/metadata/mysql/transaction.go
```diff
@@ -2010,6 +2010,1260 @@ func (d *MetadataStoreMysql) SetTransaction(
...
+		tmpTx.Metadata = tmpMetadata
+		metadataLabels = tmpLabels
+	}
+
+	collateralReturn := tx.CollateralReturn()
+	produced := tx.Produced()
+
+	// For invalid transactions with collateral returns, fix indices via CBOR matching
+	// since Produced() uses enumerated indices rather than real transaction indices
+	var realIndexMap map[lcommon.Blake2b256]uint32
```
wolf31o2: This realIndexMap workaround was deliberately removed from SetTransaction on main in commit 5ce498c4 (#2032). The replacement comment is: tx.Produced() already returns correct indices for both valid transactions (regular outputs at 0, 1, ...) and invalid transactions (collateral return at len(Outputs())). Postgres and sqlite SetTransactionBatched in this PR correctly omit the workaround. The mysql batched path here re-applies the rewrite on top of indices that are now already correct, which corrupts the collateral-return output_idx for invalid transactions. After rebasing onto main the same fix needs to land here too. Drop the realIndexMap init below and the rewrite branch inside the produced loop a few lines down (around line 2107).

database/plugin/metadata/sqlite/transaction.go
```diff
@@ -2158,6 +2159,1256 @@ func (d *MetadataStoreSqlite) SetTransaction(
...
+			if result := db.Model(&models.Certificate{}).
+				Where("transaction_id = ?", tmpTx.ID).
+				Pluck("id", &unifiedIDs); result.Error != nil {
+				return fmt.Errorf(
+					"query existing unified certificates (batched): %w",
+					result.Error,
+				)
+			}
+			if len(unifiedIDs) > 0 {
+				tables := []string{
```
wolf31o2, line 2682: Missing pool_registration_owner and pool_registration_relay before pool_registration. The mysql SetTransactionBatched path in this PR (around line 2487) and the postgres equivalent (around line 2448) both list those children first with the comment: Child tables must be deleted before parent tables (FK constraints). SQLite has FK enforcement enabled (_pragma=foreign_keys(1) in database/database.go) and models.PoolRegistration declares Owners and Relays with constraint:OnDelete:CASCADE, but the parent delete is issued before the cascade resolves so the FK check fires first. On a retry of a pool-registration transaction this will FK-violate. The non-batched SetTransaction at line 1277 has the same omission, so this perpetuates an existing inconsistency rather than creating a fresh one, but the fix landed in the other two backends in this PR and should land here while these lines are being added. Add pool_registration_owner and pool_registration_relay at the top of this slice.

wolf31o2 (changes requested):
Local code review. Two blocking issues flagged inline. The PR base predates main, so the comments target lines in the new code added by this PR.

database/plugin/metadata/sqlite/transaction.go
jkawan: addressed

wolf31o2 (approved):

PR #2140, opened 2026-05-01
fix(ledgerstate): extract nonces correctly from snapshot

wolf31o2:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected nonce extraction to handle multiple snapshot shapes and strengthened validation.
  * Improved epoch-boundary consensus checks to detect convergence/regression around epoch transitions.

* **Refactor**
  * Centralized nonce extraction into a dedicated implementation to clarify array-shape handling.

* **Tests**
  * Added integration test for epoch-boundary consensus behavior.
  * Added unit tests validating nonce extraction across snapshot variants.

### DIFF
--- a/internal/devnet/scenarios/epoch_boundary_test.go
+++ b/internal/devnet/scenarios/epoch_boundary_test.go
@@ -1,0 +1,107 @@
+//go:build devnet
+
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


arepala-uml (approved):

internal/devnet/scenarios/epoch_boundary_test.go
```diff
@@ -0,0 +1,107 @@
...
+package scenarios
+
+import (
+	"testing"
+	"time"
+
+	"github.com/blinklabs-io/dingo/internal/devnet"
+	"github.com/stretchr/testify/require"
+)
+
```
arepala-uml, line 26: I just want to confirm like this integration test verifies if Dingo successfully crosses the epoch boundary and stays in sync with cardano-node, that proves the VRF check did not fail, which proves the nonce was read correctly from PraosState(After 12 mins). Can you please confirm like my understanding of this epoch_boundary_test is correct or not?

wolf31o2: Correct. However, it only runs when you pass the build flag and run tests, so it doesn't run by default.

PR #2145, opened 2026-05-01
fix(bark): get block proxy metadata

### LINKED ISSUE #2144: Bark proxy not invoked from BlockFetch server iterator path despite tombstone fix
Follow-up to #2104 / PR #2136. After the tombstone fix landed, the storage side works as designed (pruner writes tombstone markers, bp/bi/bh keys stay in place), but an external BlockFetch against a pruning node still cannot resolve a tombstoned block.

## Repro (same setup as #2104)

1. Two Dingo nodes peered with a single block producer.
2. Node A (archive): `blobPlugin: s3`, `barkPort: 3003`, S3 bucket reachable.
3. Node B (pruning): `blobPlugin: badger`, `barkBaseUrl: http://nodeA:3003`, `barkPrunerFrequency: 5s`.
4. Wait for tip past the stability window (3k/f = 300 slots on the testnet.yaml shipped with internal/test/devnet/, so e.g. tip ≥ 400).
5. Pick a target block at slot ≪ tip - stability_window (e.g. slot 50).
6. Connect to node B over NtN from a third client and BlockFetch the target.

Expected: node B proxies the request through bark to node A and returns the block CBOR.

Actual: node B's BlockFetch server replies BatchDone with zero blocks, in ~200 microseconds. Far too fast for the bark/S3 round trip to have happened.

## Storage state is correct

Verified directly with a small badger probe against the pruning node's blob dir after the test:

- `bp(slot=61, hash=…)`: 4 bytes, value starts `44425431` ("DBT1") - the tombstone marker
- `bh(hash=…)`: 42 bytes pointing to the bp key
- `bi(id=14)`: 42 bytes pointing to the same bp key
- 128 full-CBOR bp values + 135 tombstoned bp values present

So the pruner correctly tombstoned the block and preserved both indexes.

## What the logs show on node B during the BlockFetch attempt

```
listener: accepted connection from 172.21.0.1:43420
handshake server: propose versions
blockfetch server: request range
blockfetch server: calling StartBatch()
blockfetch server: calling BatchDone()       <-- 200us after StartBatch
[60s pass: client times out]
peergov: connection closed
```

No `blockfetch: failed to start batch`, no `blockfetch: iterator error, aborting batch`, no `blockfetch: failed to send block to peer`.

On node A (archive), bark's gRPC listener is up and reachable from the pruning container, but the archive sees no incoming FetchBlock RPCs during the test window.

## Where the path appears to short-circuit

The BlockFetch server uses `o.LedgerState.GetChainFromPoint(start, true)` then `chainIter.Next(false)` (`ouroboros/blockfetch.go`). The non-blocking iterator reaches `chain.iterNext` (`chain/chain.go`) which calls `c.blockByIndex(iter.nextBlockIndex)` → `cm.db.BlockByIndex(idx, nil)` (`database/block.go`).

`BlockByIndex` does:
1. `blob.Get(blobTxn, biKey)` - bi pointer lookup, passthrough through `BlobStoreBark.Get` to badger. Returns the bp-key bytes (verified present).
2. `blockByKey(txn, val)` - decodes the bp key to (slot, hash), then calls `blob.GetBlock(txn.Blob(), slot, hash)`.

`blob` here is `txn.DB().Blob()`. After `node.go:320` does `n.db.SetBlobStore(bark.NewBarkBlobStore(...))`, `Database.Blob()` returns the bark wrapper, so this should land in `BlobStoreBark.GetBlock` (`bark/blob.go`). For a past-window slot, that function should fall through directly to `fetchBlockFromArchive`, which makes a connect-go gRPC call to `barkBaseUrl`.

The 200 microsecond turnaround and the absence of any FetchBlock RPC at the archive both rule out `fetchBlockFromArchive` actually being entered. So either:

- The chain iterator is hitting a different path than the one above, or
- `blob.GetBlock` is returning a `types.ErrBlobKeyNotFound` from somewhere (which `blockByKey` translates to `models.ErrBlockNotFound` → `ErrIteratorChainTip` → silent BatchDone), but I cannot identify what would return that error given the storage state is intact, or
- `txn.DB().Blob()` is returning something other than the bark wrapper in this particular call site.

## Suggested next steps

- Add a debug-level log at the top of `BlobStoreBark.GetBlock` (and inside both branches of the security-window check) so it is obvious whether the wrapper is being entered.
- If it is being entered: log the fall-through to `fetchBlockFromArchive` and any error returned from the gRPC call.
- If it is not being entered: confirm whether `Database.Blob()` returns the wrapped store from inside `database.BlockByIndex`'s freshly-opened `BlobTxn`. The path `BlobTxn → txn.DB() → Blob()` should be walked.

A small `bark/blob_test.go`-level test that opens a wrapped store, tombstones a block, and asks for it via the iterator-driven path (rather than a direct `GetBlock` call) would cover the gap.

## Why now

Blocks issue #2067 (archive node demo): the demo's headline assertion is "BlockFetch on the pruning node transparently returns a tombstoned block via bark." That assertion still fails today.

wolf31o2:
Closes #2144 

---
## Summary by cubic
Fixes missing block metadata when fetching tombstoned or pruned blocks via the archive proxy. Keeps the local block ID so BlockByIndex works again.

- **Bug Fixes**
  - `bark`: Always query upstream first; on `ErrBlockTombstoned` or `ErrBlobKeyNotFound` fetch CBOR from the archive and merge metadata (keep upstream ID; fill Type/Height/PrevHash from archive if zero).
  - `badger`: `GetBlock` returns metadata even for tombstoned blocks (error raised after reading); `TombstoneBlock` keeps `bp_metadata`.

<sup>Written for commit 1faa9d7b17c6780ab3eadd991157bb41932497ea. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved block retrieval fallback so upstream fetches are tried first and clearer errors are returned for genuine upstream failures.
  * On fallback, archive data is used to backfill missing metadata fields when upstream data is incomplete, ensuring more consistent block results.
  * Preserved metadata for tombstoned (deleted) blocks so historical metadata is retained and returned where available.

### DIFF
--- a/bark/blob.go
+++ b/bark/blob.go
@@ -203,31 +203,46 @@ func (b *BlobStoreBark) GetBlock(
 	slot uint64,
 	hash []byte,
 ) ([]byte, types.BlockMetadata, error) {
-	currentSlot, err := b.config.LedgerState.CurrentSlot()
-	if err != nil {
-		return nil, types.BlockMetadata{},
-			fmt.Errorf("failed to get current slot: %w", err)
+	// Always consult the upstream first so we can pick up the local
+	// BlockMetadata (most importantly the block ID, which the chain
+	// iterator's BlockByIndex path depends on). Upstream reports
+	// ErrBlockTombstoned for archive-only blocks while still returning
+	// the metadata it kept around for exactly this purpose. Fall through
+	// to the archive on ErrBlobKeyNotFound too: blocks pruned before the
+	// tombstone-on-prune fix (or never indexed locally, e.g. snapshot
+	// bootstrap) have no bp entry, but the archive can still serve them.
+	upstreamCbor, upstreamMeta, err := b.upstream.GetBlock(txn, slot, hash)
+	if err == nil {
+		return upstreamCbor, upstreamMeta, nil
 	}
-
-	// If the requested slot is within the security window, the block is
-	// expected to live locally. Defer to upstream — but if upstream reports
-	// the block was tombstoned (rare in-window case, e.g. after a rollback
-	// crosses a previously pruned slot back into the window) fall through
-	// to the archive proxy.
-	securityWindow := b.config.LedgerState.StabilityWindow()
-	if securityWindow > currentSlot ||
-		slot >= currentSlot-securityWindow {
-		cbor, meta, err := b.upstream.GetBlock(txn, slot, hash)
-		if !errors.Is(err, types.ErrBlockTombstoned) {
-			return cbor, meta, err
-		}
+	if !errors.Is(err, types.ErrBlockTombstoned) &&
+		!errors.Is(err, types.ErrBlobKeyNotFound) {
+		return nil, types.BlockMetadata{}, err
 	}
 
 	ctx, cancel := context.WithTimeout(
 		context.Background(), archiveFetchTimeout,
 	)
 	defer cancel()
-	return b.fetchBlockFromArchive(ctx, slot, hash)
+	archiveCbor, archiveMeta, archErr := b.fetchBlockFromArchive(ctx, slot, hash)
+	if archErr != nil {
+		return nil, types.BlockMetadata{}, archErr
+	}
+	// Prefer the local metadata for ID (the archive does not know our
+	// local block IDs), but trust the archive for Type/Height/PrevHash
+	// in case upstream returned a zero metadata struct alongside the
+	// tombstone error.
+	merged := upstreamMeta
+	if merged.Type == 0 {
+		merged.Type = archiveMeta.Type
+	}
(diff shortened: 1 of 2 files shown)


arepala-uml (approved):

arepala-uml:
@wolf31o2 I have understood the change and I just  have one doubt like identifying this bug only if we configure bark with our any remote_archive data storage(s3/gcs) right and can you please confirm only we run into this issue when we look for any historical block which is not present in our local badger data store?

wolf31o2:
Yes. I discovered it while working on an archive mode demo.

agaffney (approved):

PR #2152, opened 2026-05-03
fix(ledger): drain replayBufferedHeadersAsync on Close to avoid DB Closed panic

### LINKED ISSUE #2107: ledger: shutdown closes DB while replayBufferedHeadersAsync is iterating, panics with DB Closed
## Summary

A clean SIGTERM during catch-up sync can panic with `DB Closed` if `ledger.replayBufferedHeadersAsync` is mid fork-resolution probe when shutdown phase 3 closes the database. The panic propagates from a worker goroutine and kills the process before deferred cleanup runs (so for example a `--cpuprofile` capture is left empty).

Observed during a profiling run on bb52f1a7 against preview. Stack:

```
panic: DB Closed [recovered, repanicked]
github.com/dgraph-io/badger/v4.(*Txn).NewIterator
github.com/blinklabs-io/dingo/database/plugin/blob/badger.(*BlobStoreBadger).NewIterator
        database/plugin/blob/badger/database.go:515
github.com/blinklabs-io/dingo/database.BlockByHashTxn
        database/block.go:316
github.com/blinklabs-io/dingo/database.BlockByHash
        database/block.go:208
github.com/blinklabs-io/dingo/ledger.(*LedgerState).findPeerForkPath
        ledger/chainsync.go:670
github.com/blinklabs-io/dingo/ledger.(*LedgerState).tryResolveFork
        ledger/chainsync.go:1687
github.com/blinklabs-io/dingo/ledger.(*LedgerState).handleEventChainsyncBlockHeader
        ledger/chainsync.go:1488
github.com/blinklabs-io/dingo/ledger.(*LedgerState).replayBufferedHeaderEvents
        ledger/chainsync.go:950
github.com/blinklabs-io/dingo/ledger.(*LedgerState).replayBufferedHeadersAsync.func1
        ledger/chainsync.go:926
created by ledger.(*LedgerState).replayBufferedHeadersAsync
        ledger/chainsync.go:919
```

Shutdown ordering observed in the log:

```
shutdown phase 1: stopping new work
shutdown phase 1 complete  (elapsed 0)
shutdown phase 2: draining connections
shutdown phase 2 complete  (elapsed 1ms)
shutdown phase 3: flushing state
panic: DB Closed
```

## Why this matters

- Profiling and ops tooling that relies on the `--cpuprofile` / `--memprofile` flags loses captured data because deferred Stop calls in main do not run when a worker goroutine panics.
- A clean operator-initiated shutdown produces a non-zero exit and a panic stack in logs, which looks like data loss.
- The `findPeerForkPath` -> `BlockByHash` -> Badger `NewIterator` path stays hot during catch-up (see #2105), so this race is reachable in practice, not just in profiling.

## Suspected fix

Phase 1 stops new work but the async header-replay goroutine that began before phase 1 keeps running. Phase 3 closes the DB while that goroutine is still issuing reads. Either:

1. Wait for the async header-replay goroutine to drain in phase 2 (treat it as a connection-equivalent worker that must complete before state flush), or
2. Have `replayBufferedHeadersAsync` honor a shutdown context so it stops issuing new DB reads as soon as phase 1 fires, then phase 3 can safely close.

Option 2 is the smaller change and matches how the rest of the connection workers behave.

## Reproduce

1. Load `database/immutable/testdata` into a data dir.
2. Start dingo in serve mode against preview using that data dir.
3. Wait until `chainsync rollback exceeds security parameter K` or any fork-resolution log line shows up.
4. SIGTERM. Observe the `DB Closed` panic in logs and a non-zero exit.

The race does not always reproduce. A second 60 s profiling run with the same setup shut down cleanly. The race is real; just not 100%.

## Notes

- Caught while running for #1893 (operational profiling).
- Related path is the fork-resolution iterator hot path tracked in #2105.

SAY-5:
Fixes #2107.

`replayBufferedHeadersAsync` spawned a goroutine that took `chainsyncMutex` and then called into `BlockByHash` (Badger `NewIterator`) for fork resolution. A SIGTERM during catch-up could let the DB owner close the database in shutdown phase 3 while that goroutine was mid-iteration, panicking with `DB Closed` and skipping deferred `--cpuprofile` cleanup.

Two small changes pin the lifetime to `LedgerState.Close`:

- `replayBufferedHeadersAsync` checks `ls.closed` before spawning, and again after acquiring `chainsyncMutex`, so it never reaches DB reads once Close has started. In-flight workers track themselves with a new `replayWG`.
- `Close` waits on `replayWG` (with the same 10s timeout pattern as the rollback drain) before returning, so the database owner can safely close the DB after `Close` returns.

Regression tests in `ledger/chainsync_replay_shutdown_test.go` cover both: the no-op-after-closed case and that `Close` blocks until an in-flight worker exits.

`go test -race -timeout 90s ./ledger/...` passes.

---
## Summary by cubic
Prevents a shutdown race that caused a "DB Closed" panic by skipping header replay workers once closing starts, serializing Add/Wait, and draining in-flight replays during `Close` before DB teardown. Fixes #2107 and makes shutdown safe during catch-up.

- **Bug Fixes**
  - `replayBufferedHeadersAsync` holds `replayMu`, checks `ls.closed` before spawn and after `chainsyncMutex`, and tracks workers via `replayWG`.
  - `Close` unconditionally waits for `replayWG` under `replayMu` before DB shutdown, preventing Add-after-Wait and "DB Closed" panics.
  - Added regression tests for no-op after close and that `Close` blocks until a running replay exits.

<sup>Written for commit 4dabe43345f66b951e8f72bc80df666cafaa2c68. Summary will update on new commits.</sup>

## Bug Fixes
* Enhanced shutdown handling to ensure graceful termination of in-flight operations and prevent potential errors during ledger shutdown.

### DIFF
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -935,9 +935,23 @@ func (ls *LedgerState) nextBufferedHeaderConnId() (
 func (ls *LedgerState) replayBufferedHeadersAsync(
 	connId ouroboros.ConnectionId,
 ) {
+	// Bail out if Close has already started so we never spawn a worker
+	// that issues DB reads after phase 3 closes the database (#2107).
+	if ls.closed.Load() {
+		return
+	}
+	ls.replayWG.Add(1)
 	go func() {
+		defer ls.replayWG.Done()
 		ls.chainsyncMutex.Lock()
 		defer ls.chainsyncMutex.Unlock()
+		// Re-check after acquiring the mutex in case Close started
+		// while we were waiting for the lock; the DB reads inside
+		// handleEventChainsyncBlockHeader (BlockByHash etc.) will
+		// panic with "DB Closed" once the owner closes the DB.
+		if ls.closed.Load() {
+			return
+		}
 		if ls.headerPipelineConnId != (ouroboros.ConnectionId{}) ||
 			ls.chain.HeaderCount() > 0 {
 			return
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -935,12 +935,17 @@ func (ls *LedgerState) nextBufferedHeaderConnId() (
 func (ls *LedgerState) replayBufferedHeadersAsync(
 	connId ouroboros.ConnectionId,
 ) {
-	// Bail out if Close has already started so we never spawn a worker
-	// that issues DB reads after phase 3 closes the database (#2107).
+	// Hold replayMu so Close cannot start Wait between our closed check
+	// and Add(1); otherwise Wait could observe a zero counter and the
+	// subsequent Add(1) would panic with "WaitGroup misuse: Add called
+	// concurrently with Wait" (#2107).
+	ls.replayMu.Lock()
 	if ls.closed.Load() {
+		ls.replayMu.Unlock()
 		return
 	}
 	ls.replayWG.Add(1)
+	ls.replayMu.Unlock()
 	go func() {
 		defer ls.replayWG.Done()
 		ls.chainsyncMutex.Lock()
--- a/ledger/chainsync_replay_shutdown_test.go
+++ b/ledger/chainsync_replay_shutdown_test.go
@@ -1,0 +1,86 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


wolf31o2 (changes requested):
There's a couple errors pointed out by the bots. Clear those up, first.

SAY-5:
Pushed 4dabe43 to address the bot comments:

- chainsync.go: added replayMu serializing closed.Load + replayWG.Add(1), mirroring the rollbackMu pattern; closes the TOCTOU window flagged by cubic.
- state.go: drop the 10s timeout on replayWG.Wait and lock replayMu around it, so Close cannot return while a worker is still issuing DB reads (CodeRabbit P2).
- chainsync_replay_shutdown_test.go: replace the 200ms sleep with a poll on ls.closed.Load before asserting Close hasn't returned, so the test actually exercises the abort path.

go test -race ./ledger -count=5 green locally.

wolf31o2 (approved):

PR #1915, opened 2026-04-15
fix(database): block hash index

wolf31o2:
## Summary by cubic
Add a hash→block-key index to make BlockByHash lookups O(1) and avoid full scans by default. Implemented for Badger, S3, and GCS with cleanup on delete and a safe fallback for legacy blocks.

- **Bug Fixes**
  - Introduced `BlockHashIndexKey` with `"bh"` prefix for hash→block-key mapping.
  - On SetBlock/DeleteBlock, write/remove index entries in Badger, S3, and GCS.
  - BlockByHash now checks the index first and falls back to a scan for pre-index data.
  - Updated Badger key sizing to account for the new index key.

<sup>Written for commit 96f195bec30f0a260a1ed8ef05ec1635d47920a5. Summary will update on new commits.</sup>

PR #2167, opened 2026-05-05
fix(database): treat BlockByHash index miss as hard miss

### LINKED ISSUE #2105: database: BlockByHashTxn iterator fallback hot during fork resolution on catch-up
## Summary

During catch-up sync against preview, `database.BlockByHashTxn` in `database/block.go` is the second largest CPU consumer at 11.2% of cumulative time and 7.6% of allocation churn (370 MB/min flat), behind only the SQLite metadata write path. None of this cost shows up in a load-only profile, so it is specific to the live chainsync path.

The fast-path hash-index lookup at line 308 (`blob.Get(hashIndexKey)`) is cheap (around 10 ms in a 60 s sample). The cum cost is the iterator-fallback path that scans the block-blob prefix sequentially, used when the hash index lookup misses. That fallback is exercised heavily during fork resolution (`ledger.findPeerForkPath` -> `BlockByHash` -> Badger `NewIterator`). It is the same code path that hit a shutdown-time panic when the DB was closed mid-iteration during a first profiling attempt.

## Why this matters

- 11% of CPU on catch-up sync is the largest non-SQLite cost we have. Fork resolution is the lever.
- The iterator fallback also opens a Badger iterator per call. Under load that is one iterator per fork-path probe per peer, which scales badly.
- The shutdown panic showed the iterator can outlive its surrounding txn assumptions when the DB is closed concurrently. That is a separate bug but the iterator hot path makes it likelier.

## What to investigate

1. Why is the hash-index lookup missing for blocks reached during fork resolution? Two suspects:
   - Blocks that pre-date the hash index (added in #1915) are not back-filled. Confirm by checking whether older preview blocks have entries in the `block_hash_index` keyspace.
   - The hash index is not written for blocks ingested via certain paths (Mithril restore? backfill?). Confirm by counting blocks vs index entries after a Mithril snapshot import.
2. Either back-fill the missing entries on first read or treat a missing index entry as a hard miss for fork resolution and surface a metric so we can track the false-fallback rate.
3. Once the iterator path is no longer hot, decide whether to keep the fallback at all.

## Reproduce

```
./dingo --network preview --data-dir <db-from-load> --debug-port 6060 ...
# in another shell, after catch-up starts:
go tool pprof -top -cum http://127.0.0.1:6060/debug/pprof/profile?seconds=60
```

Look for `database.BlockByHashTxn` in the cum-time top.

## Numbers

Captured on bb52f1a7 against preview, 60 s sample, post-bootstrap:

- BlockByHashTxn cum CPU: 9.48 s of 84.92 s total (11.16%)
- BlockByHashTxn flat alloc: 370 MB / 60 s (7.6% of churn)
- Fast-path blob.Get: 10 ms / 60 s

SAY-5:
Fixes #2105.

The legacy iterator-fallback in `BlockByHashTxn` ran a full block-blob prefix scan on every hash-index miss, dominating fork-resolution CPU during preview catch-up (11.2% cum, 370 MB/min churn).

Since #1915 every block has a hash-index entry, so a miss on a healthy DB means the hash is genuinely unknown, `findPeerForkPath` can rotate to the next peer without scanning. Real backend errors are still surfaced.

Adds `BlockByHashStats()` (atomic hit/miss counters) so operators can track the false-fallback rate and decide whether a pre-#1915 index back-fill is needed, the metric the issue asks for.

Regression test `TestBlockByHashTxn_UnknownHashIsHardMiss` plus a `-benchmem` benchmark guard the constant-time miss path.

Tests: `go test -race -timeout 90s ./database/` passes locally.

---
## Summary by cubic
Treat `BlockByHashTxn` hash-index misses as hard “not found” with no `bp` prefix scan to cut fork-resolution CPU during catch-up. Adds hit/miss stats and Prometheus metrics (per registry) so operators can judge pre-#1915 backfill needs (refs #2105).

- **Bug Fixes**
  - On index miss, increment miss and return `models.ErrBlockNotFound`; surface non-`types.ErrBlobKeyNotFound` backend errors.
  - Empty hash-index values now return a clear corruption error.
  - Removed the legacy `bp`-prefix fallback scan and its deadline/warning path; dropped the obsolete warning test.
  - Increment the hit counter only after the block is successfully loaded.

- **New Features**
  - `BlockByHashStats()` returns cumulative hit/miss counts; added tests and a `-benchmem` benchmark.
  - Export Prometheus counters `dingo_database_block_hash_index_hits_total` and `dingo_database_block_hash_index_misses_total` via `RegisterBlockByHashMetrics(reg)`; ignores `prometheus.AlreadyRegisteredError`, returns other registration errors; safe to call per registry and on reused registries; wired in `New()` (logs a warn on registration failure).

<sup>Written for commit ad93af8f81ea7ecf7e363f6db6058c1c6c99059a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2167?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added cumulative hit/miss statistics for block-hash lookups and optional Prometheus metrics registration.

* **Behavior**
  * Lookup now prefers a constant-time hash-index path; misses increment the miss counter and return “not found”.
  * Empty index entries are treated as database corruption and surface an explicit error.

* **Tests**
  * Added tests and a benchmark validating hit/miss counting, known/unknown lookup behavior, and corruption handling.
* **Chore**
  * Removed an obsolete test related to a fallback-scan warning.

[Review Change Stack](https://app.coderabbit.ai/change-stack/blinklabs-io/dingo/pull/2167?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


database/block.go
```diff
@@ -302,46 +331,26 @@ func BlockByHashTxn(txn *Txn, hash []byte) (models.Block, error) {
...
-		return blockByKey(txn, key)
-	}
-	if err := it.Err(); err != nil {
+	if err != nil && !errors.Is(err, types.ErrBlobKeyNotFound) {
+		// Surface real backend errors (I/O, closed DB, etc), don't
+		// hide them behind ErrBlockNotFound.
 		return models.Block{}, err
 	}
+	blockByHashIndexMisses.Add(1)
 	return models.Block{}, models.ErrBlockNotFound
```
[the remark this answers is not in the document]
wolf31o2: This is valid.

wolf31o2 (changes requested):
Clean up the merge conflicts and fix the issue from the coderabbit review

database/block.go
```diff
@@ -321,26 +337,43 @@ func BlockByHashTxn(txn *Txn, hash []byte) (models.Block, error) {
...
+		// not a soft miss. Surface a descriptive error so the operator sees
+		// the problem rather than retrying on another peer.
+		return models.Block{}, fmt.Errorf("empty block hash index entry for %s", hex.EncodeToString(hash))
+	}
+	if err != nil && !errors.Is(err, types.ErrBlobKeyNotFound) {
+		// Surface real backend errors (I/O, closed DB) instead of folding
+		// them into ErrBlockNotFound.
+		return models.Block{}, err
+	}
+	blockByHashIndexMisses.Add(1)
```
wolf31o2, line 405: After counting the index miss, this still falls through into the `bp` prefix iterator below. That leaves the fork-resolution miss path doing per-peer scan work instead of the hard miss #2105 needs; please return `models.ErrBlockNotFound` here or move any backfill to an offline/one-time path.

database/block.go
```diff
@@ -40,6 +40,22 @@ var (
...
+// node is the load-only signal that operators can use to justify a
+// hash-index back-fill for pre-#1915 blocks (see #2105).
+var (
+	blockByHashIndexHits   atomic.Uint64
+	blockByHashIndexMisses atomic.Uint64
+)
+
+// BlockByHashStats returns the cumulative hit/miss counts for the
+// hash-index fast path used by BlockByHashTxn.
+func BlockByHashStats() (hits, misses uint64) {
```
wolf31o2, line 91: This accessor is only used by tests, so operators still do not get the miss-rate signal requested in #2105. Please wire these counters into the metrics endpoint or remove the claim that the PR exposes an operator metric.

wolf31o2 (changes requested):
Requesting changes: current head still scans on every BlockByHash index miss, and the stats are not operator-visible. The focused tests pass, but the PR benchmark still shows ~1.7 ms/op, 253 KB/op, and 8242 allocs/op on a 1024-block DB, so this does not resolve #2105 yet.

database/block.go
```diff
@@ -22,24 +22,62 @@ import (
...
 )
 
+// RegisterBlockByHashMetrics wires the block-hash index hit/miss counters
+// into the given Prometheus registry. It is idempotent; subsequent calls
+// after the first successful registration are no-ops.
+func RegisterBlockByHashMetrics(reg prometheus.Registerer) {
+	if reg == nil {
+		return
+	}
+	blockByHashMetricsOnce.Do(func() {
```
wolf31o2: This package-level `sync.Once` makes the first registry win for the whole process. `database.New` can be called with different `PromRegistry` instances (tests already do this pattern elsewhere), but after the first call later registries will never get `dingo_database_block_hash_index_*` registered while lookups still increment the first registry's counters. Please register per provided registry or make these metrics DB-scoped, and handle `prometheus.AlreadyRegisteredError` when reusing a registry.

database/block.go
```diff
@@ -22,24 +22,62 @@ import (
...
+		factory := promauto.With(reg)
+		blockByHashHitsCounter = factory.NewCounter(prometheus.CounterOpts{
+			Name: "dingo_database_block_hash_index_hits_total",
+			Help: "Total block-hash index lookups resolved via the O(1) fast path",
+		})
+		blockByHashMissesCounter = factory.NewCounter(prometheus.CounterOpts{
+			Name: "dingo_database_block_hash_index_misses_total",
+			Help: "Total block-hash index lookups that fell through to the prefix scan",
```
wolf31o2: This help text is now stale: misses no longer fall through to a prefix scan. It should describe a hard index miss / `ErrBlockNotFound`, otherwise operators will read the metric incorrectly.

database/block_byhash_test.go
```diff
@@ -0,0 +1,145 @@
...
+// resetBlockByHashStats zeros the hit/miss counters between tests.
+func resetBlockByHashStats() {
+	blockByHashIndexHits.Store(0)
+	blockByHashIndexMisses.Store(0)
+}
+
+// TestBlockByHashTxn_UnknownHashRecordsMissAndNotFound verifies that an
+// unknown hash increments the miss counter (so operators can track the
+// false-fallback rate from #2105) and surfaces ErrBlockNotFound after
+// the time-bounded fallback scan finds nothing.
```
wolf31o2: The test comment still describes the removed time-bounded fallback scan. This should say the miss returns `ErrBlockNotFound` directly after the index miss, matching the new behavior.

wolf31o2 (changes requested):
Requesting changes. The request-time BlockByHash prefix scan is gone and the focused checks/benchmark look good, but the Prometheus wiring still is not reliable enough to merge: the package-level registration only attaches these counters to the first registry seen in-process, so later DBs/nodes/tests with their own PromRegistry will not expose the operator-visible metrics. Please make the counters register per provided registry or DB-scope them, and handle AlreadyRegisteredError instead of relying on promauto. Also sweep the remaining stale comments/docs that still describe the removed prefix-scan fallback (including the blob-plugin tombstone comments).

Checks I ran locally: git diff --check origin/main...HEAD; go test ./database; go test -race ./database; go test -run '^$' -bench=BenchmarkBlockByHashTxn_UnknownHash -benchmem ./database. PR checks are green. ARCHITECTURE.md checked and not affected; DATABASE.md is not present on this branch.

database/block.go
wolf31o2: The new Prometheus collectors are package-level and protected by a package-level `sync.Once`, but `Database.New` can be called with different registries. The first registry wins and every later registry silently misses `dingo_database_block_hash_index_*`, while lookups keep incrementing the first collector. Since #2105 explicitly asks for an operator metric, please make this registry-safe, for example with per-registry collectors/GaugeFuncs over the atomics or instance-scoped registration, and add coverage that a fresh registry receives these metrics.

database/block.go
```diff
@@ -321,70 +359,42 @@ func BlockByHashTxn(txn *Txn, hash []byte) (models.Block, error) {
...
 		return models.Block{}, err
 	}
+	blockByHashIndexMisses.Add(1)
+	if blockByHashMissesCounter != nil {
+		blockByHashMissesCounter.Inc()
+	}
+	// Index miss is a hard miss (#2105). Per-peer chainsync /
+	// intersect-point callers should fall back to PeerHeaderLookupFunc
+	// rather than scan the bp prefix on every miss; legacy blocks
+	// predating the hash index need an offline backfill.
```
wolf31o2, line 409: This is now a storage contract change: pre-#1915 blocks need `bh<hash>` entries or `BlockByHash` hard-misses. Please document that in `DATABASE.md` and clean up the stale blob-plugin comments that still say `BlockByHash` has a sequential-scan fallback (`badger`, `gcs`, `s3` TombstoneBlock docs). Operators need to know whether and how to backfill before this lands.

wolf31o2 (changes requested):
No-go for merge pending the inline issues below.

Local validation performed on PR head `a611c2c7`:
- `go test -race ./database` passed in 344.250s
- `gofmt -l database/block.go database/block_byhash_test.go database/database.go database/transaction_test.go` was clean
- `git diff --check origin/main...HEAD` was clean

`DATABASE.md`/`ARCHITECTURE.md` were checked; `DATABASE.md` appears affected and needs an update, `ARCHITECTURE.md` does not.

database/block.go
SAY-5: Reworked in 30ce129: the sync.Once and package-level promauto counters are gone. RegisterBlockByHashMetrics now creates CounterFuncs over the shared atomics per provided registry, handles AlreadyRegisteredError on reuse, and there is a test that registers two fresh registries and checks both gather the counters with the shared totals. go test -race ./database passes locally (177s).

database/block.go
SAY-5: Done in 30ce129. DATABASE.md now has a Block Hash Index Contract section spelling out the hard-miss behavior, that pre-#1915 blocks need a one-time bh backfill to stay reachable by hash, and that the hit/miss counters are the signal for whether one is needed. The badger, gcs, and s3 TombstoneBlock comments no longer mention the removed sequential-scan fallback.

wolf31o2 (approved):

PR #2168, opened 2026-05-05
database: include block_index in cert ordering for account restore

### LINKED ISSUE #2036: account restore uses ambiguous cert ordering (missing block_index) in sqlite/mysql/postgres
## Summary

`database/plugin/metadata/{sqlite,mysql,postgres}/account.go` rebuilds account state on rollback by ranking per-stake-key certificate rows with `ORDER BY added_slot DESC, cert_index DESC`. That ordering is ambiguous whenever two relevant certificates live in the same slot but in different transactions of the same block, because `cert_index` resets per transaction — so the winning row is picked by the DB's undefined tiebreaker. The pool-side queries already include `block_index` and avoid the problem; the account-side queries do not, which contradicts the documented invariant in `CLAUDE.md`:

> Cert ordering: multiple certs per slot — tie-break with `Order("added_slot DESC, block_index DESC, cert_index DESC")` (`cert_index` resets per tx, so `block_index` is required to disambiguate across txs in the same block).

## Affected sites

All three metadata backends, in the `batchFetch*` helpers used by `RestoreAccountStateAtSlot`. In `database/plugin/metadata/sqlite/account.go` alone, 10 queries use the ambiguous ordering (lines 189, 228, 273, 328, 371, 409, 447, 486, 530, 576). The MySQL and PostgreSQL copies mirror the same pattern.

The `certs` table already carries a `BlockIndex` column (`database/models/transaction.go:42`), so no schema change is needed.

## Impact

A rollback that targets a slot `R` can pick the wrong pool/DRep state for an account if two relevant certificates were written in the same slot across different transactions. It is silent (no error), non-deterministic across backends, and hard to reproduce without specifically constructed multi-tx blocks. Observable downstream as incorrect account delegation state after a rollback that crosses such a block.

## Context

Caught during review of PR #2035 (Conway pv10 HARDFORK rule). That PR bumps `Account.AddedSlot` to the epoch-rollover boundary so the rewritten row is excluded from rollback restores targeting slots before the boundary, which is correct — but the restore machinery it relies on has this pre-existing ambiguity and so cannot make a strong \"correctly re-derive\" guarantee for the underlying account state. The docstrings in #2035 were softened accordingly; the actual ordering fix belongs here.

## Proposed fix

1. Update all `batchFetch*` helpers in each of the three `account.go` files to order by `added_slot DESC, block_index DESC, cert_index DESC`. The existing `INNER JOIN certs c ON c.id = t.certificate_id` already joins the row that has `block_index`, so the change is purely ordering.
2. Audit any other sites in `account.go` (and adjacent files) that rank cert-derived rows; apply the same ordering.
3. Add a regression test that constructs multiple certs for the same stake key in the same slot across different transactions and asserts the rollback restore picks the block-order-latest cert.
4. Consider factoring the ordering clause into a named constant/helper so future queries cannot reintroduce the drift.

## References

- CLAUDE.md \"Non-obvious invariants\" section (cert ordering).
- Pool queries that already follow the invariant: `database/plugin/metadata/{sqlite,postgres,mysql}/pool.go` (e.g. sqlite/pool.go:696, 701, 744).
- PR #2035 review comment that surfaced this: https://github.com/blinklabs-io/dingo/pull/2035

SAY-5:
Fixes #2036.

The 10 `batchFetch*` helpers in `database/plugin/metadata/{sqlite,mysql,postgres}/account.go` ranked per-stake-key cert rows with `ORDER BY added_slot DESC, cert_index DESC`, which is ambiguous across transactions in the same block (`cert_index` resets per tx). Each query now joins `transaction` and orders by `added_slot DESC, block_index DESC, cert_index DESC`, matching the pool-side queries and the invariant in CLAUDE.md. New regression test `TestBatchFetchCerts_SameSlotTiebreakByBlockIndex` constructs two stake_delegation certs in the same slot but different block_index, asserts the higher-block_index pool wins, and fails on `main` (non-deterministic) before this fix.

---
## Summary by cubic
Make account restore deterministic by ranking certs with added_slot, block_index, then cert_index to break same-slot ties across transactions and tables.

- **Bug Fixes**
  - LEFT JOIN `transaction` and order by added_slot DESC, COALESCE(tx.block_index, 0) DESC, cert_index DESC in all account batch fetch queries for `sqlite`, `mysql`, and `postgres`.
  - Thread `block_index` through `certRecord` and use it in the cache merge comparator to resolve cross-transaction and cross-table ties.
  - Add regression tests: `TestBatchFetchCerts_SameSlotTiebreakByBlockIndex` and `TestBatchFetchCerts_CrossTableTiebreakByBlockIndex`.

<sup>Written for commit e9b37c60947a7bace589a61e610272a3ef2fa26f. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deterministic selection of the “latest” certificate/ stake record when multiple events share the same slot by adding an additional tie-breaker, reducing inconsistent restoration results.

* **Tests**
  * Added tests covering same-slot and cross-table tie-break scenarios to ensure consistent certificate selection and cache behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/blinklabs-io/dingo/pull/2168)

### DIFF
--- a/database/plugin/metadata/mysql/account.go
+++ b/database/plugin/metadata/mysql/account.go
@@ -211,25 +211,30 @@ func (d *MetadataStoreMysql) SetAccount(
 
 // certRecord holds common fields extracted from certificate records for
 // batch processing during account state restoration.
-// Includes certIndex for same-slot disambiguation.
+// Includes blockIndex and certIndex for same-slot disambiguation across
+// transactions and within a transaction respectively.
 type certRecord struct {
-	pool      []byte
-	drep      []byte
-	drepType  uint64
-	addedSlot uint64
-	certIndex uint32
+	pool       []byte
+	drep       []byte
+	drepType   uint64
+	addedSlot  uint64
+	blockIndex uint64
+	certIndex  uint32
 }
 
 // isMoreRecent checks if this certificate is more recent than the other.
-// When slots are equal, certIndex determines order (higher = later in transaction).
+// Order matches the SQL ORDER BY clause used in batchFetch helpers:
+// added_slot DESC, block_index DESC, cert_index DESC. block_index is
+// required to disambiguate same-slot certs across different transactions
+// because cert_index resets per transaction.
 func (r certRecord) isMoreRecent(other certRecord) bool {
-	if r.addedSlot > other.addedSlot {
-		return true
+	if r.addedSlot != other.addedSlot {
+		return r.addedSlot > other.addedSlot
 	}
-	if r.addedSlot == other.addedSlot && r.certIndex > other.certIndex {
-		return true
+	if r.blockIndex != other.blockIndex {
+		return r.blockIndex > other.blockIndex
 	}
-	return false
+	return r.certIndex > other.certIndex
 }
 
 // accountCertCache holds batch-fetched certificate data for all accounts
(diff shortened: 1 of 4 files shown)


wolf31o2 (changes requested):
The bots are right. These need to be fixed.

wolf31o2 (approved):

PR #2175, opened 2026-05-05
fix(peergov): DNS resolution failure silently falls back to hostname

### LINKED ISSUE #1526: peergov: DNS resolution failure silently falls back to hostname
## Summary

In peergov/peers.go, the resolveAddress function silently falls back to using the raw hostname when DNS resolution fails. This means the peer governance system will track the same peer under different keys (hostname vs IP) depending on whether DNS was working at the time, leading to inconsistent scoring, duplicate entries, and possible bypass of deny lists.

## Current State

In peergov/peers.go, lines 235-238:

```
ips, err := net.LookupIP(host)
if err != nil || len(ips) == 0 {
    // Can't resolve, just lowercase the hostname
    return net.JoinHostPort(strings.ToLower(host), port)
}
```

No error is logged. The caller gets back a hostname-based address instead of an IP-based one. If the same peer was previously resolved to an IP, it will now have two entries in the peer map.

## What Needs to Happen

1. Log a warning when DNS resolution fails so operators can detect infrastructure issues.
2. Consider whether the fallback behavior is correct -- if a peer cannot be resolved, should it be added to the peer map at all?
3. Ensure that peer scoring and deny lists work correctly regardless of whether the address is stored as hostname or IP. If they compare by string equality, a DNS failure can cause a denied peer to be re-added under its hostname.

arepala-uml:
1. Added a warning log message in resolveAddress when DNS lookup fails instead of silently falling back to the hostname.
2. Made changes in AddPeer where it checks both the resolved address and the hostname-normalized address before adding a peer.
3. Also, made changes in AddPeer where it dedupes peers across IP and hostname forms, so the same peer is not tracked twice when DNS works once and fails later.
4. Made changes in DenyPeer to store deny entries for both known forms of a peer like resolved/IP form and hostname form.
5. Made changes in IsDenied where it checks both forms, so a peer denied as an IP cannot be allowed back as a hostname and ledger_discovery now applies the same two-form deny check before adding ledger-discovered peers.
6. Added the tests to cover DNS failure logging, duplicate prevention, and deny-list behavior when DNS succeeds first and fails later.

Closes #1526 

---
## Summary by cubic
Add unit tests to verify `peergov` dedupes peers and enforces the deny list when DNS resolution fails, and that DNS failures are logged. Introduces an injectable `lookupIP` and updates `resolveAddress` to use it for reliable DNS stubbing in tests.

<sup>Written for commit 7ac6ee93d00998a04a8819a2a89451a567efd775. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved peer denial checks to work consistently across different address variants
  * Enhanced DNS resolution error handling with better fallback mechanisms  
  * More reliable peer deduplication when address normalization is involved

### DIFF
--- a/peergov/connections.go
+++ b/peergov/connections.go
@@ -727,14 +727,14 @@ func (p *PeerGovernor) DenyPeer(address string, duration time.Duration) {
 	normalized := p.resolveAddress(address)
 	p.mu.Lock()
 	defer p.mu.Unlock()
-	// if idx := p.peerIndexByAddress(address); idx != -1 && p.peers[idx] != nil {
-	// 	p.denyList[p.peers[idx].NormalizedAddress] = time.Now().Add(duration)
-	// 	p.denyList[p.normalizeAddress(p.peers[idx].Address)] = time.Now().
-	// 		Add(duration)
-	//} else {
-	p.denyList[normalized] = time.Now().Add(duration)
-	// 	p.denyList[p.normalizeAddress(address)] = time.Now().Add(duration)
-	// }
+	if idx := p.peerIndexByAddress(address); idx != -1 && p.peers[idx] != nil {
+		p.denyList[p.peers[idx].NormalizedAddress] = time.Now().Add(duration)
+		p.denyList[p.normalizeAddress(p.peers[idx].Address)] = time.Now().
+			Add(duration)
+	} else {
+		p.denyList[normalized] = time.Now().Add(duration)
+		p.denyList[p.normalizeAddress(address)] = time.Now().Add(duration)
+	}
 	p.config.Logger.Debug(
 		"peer added to deny list",
 		"address", address,
@@ -748,18 +748,17 @@ func (p *PeerGovernor) DenyPeer(address string, duration time.Duration) {
 func (p *PeerGovernor) IsDenied(address string) bool {
 	// Resolve address before acquiring lock to avoid blocking DNS
 	normalized := p.resolveAddress(address)
-	// hostnameNormalized := p.normalizeAddress(address)
+	hostnameNormalized := p.normalizeAddress(address)
 	p.mu.Lock()
 	defer p.mu.Unlock()
-	return p.isDeniedLocked(normalized)
-	// if p.isDeniedLocked(normalized) || p.isDeniedLocked(hostnameNormalized) {
-	// 	return true
-	// }
-	// if idx := p.peerIndexByAddress(address); idx != -1 && p.peers[idx] != nil {
-	// 	return p.isDeniedLocked(p.peers[idx].NormalizedAddress) ||
-	// 		p.isDeniedLocked(p.normalizeAddress(p.peers[idx].Address))
-	// }
-	// return false
+	if p.isDeniedLocked(normalized) || p.isDeniedLocked(hostnameNormalized) {
+		return true
+	}
+	if idx := p.peerIndexByAddress(address); idx != -1 && p.peers[idx] != nil {
+		return p.isDeniedLocked(p.peers[idx].NormalizedAddress) ||
+			p.isDeniedLocked(p.normalizeAddress(p.peers[idx].Address))
+	}
+	return false
 }
 
 // isDeniedLocked checks if a peer is on the deny list.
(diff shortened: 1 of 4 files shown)


arepala-uml:
1. Added this test (**TestPeerGovernor_AddPeer_DedupesHostnameAfterDNSFailure**) - This test checks that peergov does not add the same peer twice when DNS resolution changes over time. It starts with a peer already known as relay.example.com:3001, but stored with normalized IP 44.0.0.1:3001, which represents the case where DNS worked earlier. The expected result is that the peer list still contains only one peer.

```
akhil-mac@Akhils-MacBook-Air dingo % 
akhil-mac@Akhils-MacBook-Air dingo % go test ./peergov -run 'TestPeerGovernor_AddPeer_DedupesHostnameAfterDNSFailure'                    
--- FAIL: TestPeerGovernor_AddPeer_DedupesHostnameAfterDNSFailure (0.00s)
    peergov_test.go:1262: 
                Error Trace:    /Users/akhil-mac/repos/dingo/peergov/peergov_test.go:1262
                Error:          "[0x140000c6ea0 0x140000c7040]" should have 1 item(s), but has 2
                Test:           TestPeerGovernor_AddPeer_DedupesHostnameAfterDNSFailure
FAIL
FAIL    github.com/blinklabs-io/dingo/peergov   0.348s
FAIL
akhil-mac@Akhils-MacBook-Air dingo % 
```
Without fix, the test fails before the fix because the current AddPeer only compares normalized address. During DNS failure, the new normalized address becomes relay.example.com:3001, which does not equal the existing 44.0.0.1:3001, so old code incorrectly adds a duplicate peer.

2. Added this test(**TestPeerGovernor_DenyPeer_BlocksHostnameAfterDNSFailure**) - This test checks that a denied peer cannot bypass the deny list when DNS behavior changes. It first makes DNS resolution succeed, so relay.example.com:3001 is denied as the resolved IP 44.0.0.1:3001. Then the test makes DNS resolution fail and tries to add the same hostname again. The expected result is that the peer is not added and IsDenied("relay.example.com:3001") returns true.

```
akhil-mac@Akhils-MacBook-Air dingo % go test ./peergov -run 'TestPeerGovernor_DenyPeer_BlocksHostnameAfterDNSFailure'                    
--- FAIL: TestPeerGovernor_DenyPeer_BlocksHostnameAfterDNSFailure (0.00s)
    peergov_test.go:1285: 
                Error Trace:    /Users/akhil-mac/repos/dingo/peergov/peergov_test.go:1285
                Error:          Should be empty, but was [0x140000cc4e0]
                Test:           TestPeerGovernor_DenyPeer_BlocksHostnameAfterDNSFailure
    peergov_test.go:1286: 
                Error Trace:    /Users/akhil-mac/repos/dingo/peergov/peergov_test.go:1286
                Error:          Should be true
                Test:           TestPeerGovernor_DenyPeer_BlocksHostnameAfterDNSFailure
FAIL
FAIL    github.com/blinklabs-io/dingo/peergov   0.310s
FAIL
akhil-mac@Akhils-MacBook-Air dingo % 
```

Without fix, the test fails before the fix because currently it stores only the resolved IP in the deny list. Later, when DNS fails, AddPeer checks the hostname form relay.example.com:3001, which does not match the denied IP 44.0.0.1:3001. It incorrectly allows the denied peer to be added again.

3. TestPeerGovernor_ResolveAddress_LogsDNSFailure  - This test checks that peergov logs a warning when DNS resolution fails or not.

```
akhil-mac@Akhils-MacBook-Air dingo % go test ./peergov -run 'TestPeerGovernor_ResolveAddress_LogsDNSFailure'         
--- FAIL: TestPeerGovernor_ResolveAddress_LogsDNSFailure (0.00s)
    peergov_test.go:1304: 
                Error Trace:    /Users/akhil-mac/repos/dingo/peergov/peergov_test.go:1304
                Error:          Should be true
                Test:           TestPeerGovernor_ResolveAddress_LogsDNSFailure
                Messages:       expected DNS resolution failure to be logged
FAIL
FAIL    github.com/blinklabs-io/dingo/peergov   0.407s
FAIL
akhil-mac@Akhils-MacBook-Air dingo % 
```

wolf31o2 (approved):

PR #2202, opened 2026-05-07
fix(ledgerstate): extract gov action parents from cert state at import

wolf31o2:
---
## Summary by cubic
Extract and persist parent governance action IDs and raw action CBOR during cert-state import in `ledgerstate` so chained proposals validate and enactments decode correctly after snapshots. Also mark the uniquely matching Hard Fork Initiation as ratified at the snapshot to avoid a 1‑epoch enactment lag.

- **Bug Fixes**
  - Extract parent `GovActionId` for chained types (0, 1, 3, 4, 5); skip unchained types.
  - Persist `ParentTxHash`, `ParentActionIdx`, and `GovActionCbor` on proposals during import.
  - When exactly one active HFI’s parent equals the snapshot HardFork root, set its ratified epoch/slot to the snapshot values to align enactment timing.

<sup>Written for commit 0d40cc5884924b9e5cb5b025f41965262363663b. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Track parent governance-action predecessors for proposals and persist raw action payloads; store richer proposal metadata and apply snapshot-time ratification for matching hard-fork candidates.
* **Chores / Performance**
  * Throttle expensive hard-fork stability evaluations and avoid redundant work after rollbacks.
* **Bug Fixes**
  * Scale blockfetch batching to avoid pipeline starvation when far behind the tip.
* **Tests**
  * Adjust blockfetch batch-size tests to reflect new scaling behavior.

### DIFF
--- a/ledgerstate/import.go
+++ b/ledgerstate/import.go
@@ -15,6 +15,7 @@
 package ledgerstate
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
 	"errors"
@@ -1942,6 +1943,34 @@ func importGovState(
 
 	// Import governance proposals
 	if len(govState.Proposals) > 0 {
+		// Snapshot bootstrap close-the-1-epoch-lag heuristic. Canon
+		// nodes mark a HFI ratified at boundary N-1 -> N and enact at
+		// boundary N -> N+1. The snapshot at epoch N reflects the
+		// ratification (in cgsDRepPulsingState's DRComplete result)
+		// but we don't parse that field, so without intervention our
+		// local node would re-ratify at boundary N -> N+1 and only
+		// enact at N+1 -> N+2 — putting us one epoch behind canon
+		// for the protocol-version bump. As a block producer that is
+		// the difference between forging vN+ blocks alongside canon
+		// and forging vN- blocks that risk rejection on TX-validation
+		// rule mismatches gated by the new version.
+		//
+		// Heuristic: when there is exactly one active HFI whose
+		// parent equals the snapshot's per-purpose HardFork root,
+		// mark it ratified at the snapshot's epoch so ENACT picks
+		// it up at the next boundary tick. Skips silently when the
+		// match is ambiguous (0 or >1 candidates) — the lag returns
+		// in those cases but correctness is preserved.
+		ratifiedHFI := findRatifiedHFICandidate(govState)
+		if ratifiedHFI != nil {
+			cfg.Logger.Info(
+				"marking active HFI as ratified at snapshot epoch (close 1-epoch lag)",
+				"component", "ledgerstate",
+				"tx_hash", hex.EncodeToString(ratifiedHFI.TxHash),
+				"action_index", ratifiedHFI.ActionIndex,
+				"ratified_epoch", cfg.State.Epoch,
+			)
+		}
 		if err := func() error {
 			txn := cfg.Database.MetadataTxn(true)
 			defer txn.Release()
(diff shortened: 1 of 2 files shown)


wolf31o2:
@cubic-dev-ai review

ledgerstate/import.go
```diff
@@ -1942,6 +1943,34 @@ func importGovState(
...
+		// and forging vN- blocks that risk rejection on TX-validation
+		// rule mismatches gated by the new version.
+		//
+		// Heuristic: when there is exactly one active HFI whose
+		// parent equals the snapshot's per-purpose HardFork root,
+		// mark it ratified at the snapshot's epoch so ENACT picks
+		// it up at the next boundary tick. Skips silently when the
+		// match is ambiguous (0 or >1 candidates) — the lag returns
+		// in those cases but correctness is preserved.
+		ratifiedHFI := findRatifiedHFICandidate(govState)
```
chrisguiney, line 1964: This heuristic is only asserted-safe ("correctness is preserved") for the ambiguous (0 or >1) case. The single-match path here marks an HFI ratified based purely on its structural shape — that it chains off `PrevGovActionIds.HardFork` — but the snapshot doesn't tell us whether canon actually ratified it at boundary N-1→N. `cgsDRepPulsingState`'s DRComplete carries the actual ratification result; this PR explicitly chooses not to parse it.

Failure mode: a chained HFI is proposed but no canon majority materialized at the boundary. This single-match path then causes the local node to enact a protocol-version bump that canon didn't enact. For a block producer, that means forging vN+ blocks while canon stays at vN-, with rejection at the first new-version-gated TX rule. That's the same divergence-class bug the comment cites as the *reason* for the heuristic — just moved to a different boundary.

Two ways forward:
- Parse DRComplete and use the actual ratification result. The PR is already paying the cert-state-parsing cost; this is the principled fix.
- Gate the heuristic behind a config flag, default off, with the divergence risk documented at the flag site.

At minimum this needs unit tests that exercise the failure mode (single match, but the candidate is unratifiable) — there is no current test demonstrating the single-match path is actually safe.

ledger/state.go
```diff
@@ -3676,6 +3677,23 @@ func (ls *LedgerState) evaluateHardForkInitiationStability() {
...
+	// total CPU during chain catchup on preview when post-deadline
+	// applies for ~14h per epoch (boundary-window 51840 slots, far
+	// from the "few minutes" the function header guesses). Inputs to
+	// the tally only change with a new vote/registration/stake-change
+	// transaction, so re-running on every block-apply (and especially
+	// on every rollback recovery iteration) is wasted work. 30s is
+	// short enough that any newly-cast vote in the deadline window
+	// still surfaces TransitionKnown well before the boundary tick,
+	// while bounding worst-case CPU to one full tally per 30s.
+	if !ls.lastHFIStabilityEvalAt.IsZero() &&
```
chrisguiney: This throttle is unconditional, but the rollback path at `state.go:1780-1782` explicitly clears `transitionInfo` to Unknown and immediately calls `evaluateHardForkInitiationStability()` to re-derive Known(N+1). With this gate, that re-derivation is silently skipped when the previous evaluation ran <30s ago — leaving `transitionInfo` stuck at Unknown for up to 30s. `CurrentTransitionInfo`, `queryHardForkEraHistory`, and `BuildSummary` all read it, so clients querying era history during the throttle window get safeZone-capped answers instead of confirmed boundaries.

The block comment at `state.go:1768-1775` documents the rollback contract this throttle now violates ("evaluateHardForkInitiationStability restores Known(N+1) if a HardForkInitiation governance action survived the rollback"). It's degraded UX rather than chain divergence, but it's a regression of stated behavior.

Same concern at `state.go:720` (startup) — the contract there is also "first call after restart must run."

Suggested shape: scope the throttle to the block-apply call site (`state.go:3136`) only — that's the per-block path during catchup that actually motivates the fix. Or add a `force bool` parameter that rollback/startup set to bypass. As-is, this is doing more than reduce CPU during catchup.

chrisguiney (changes requested):
Requesting changes — three concerns I'd like to see addressed before merge.

The CBOR extraction (parent gov-action ID + raw `GovActionCbor`) is solid and clearly fixes a real bug. But it's bundled with two changes that I think each need rework:

1. **HFI ratification heuristic carries chain-divergence risk for block producers.** The single-match case asserts canon ratified an action based purely on its structural shape, without consulting `cgsDRepPulsingState`/DRComplete. See inline on `ledgerstate/import.go`.
2. **30s throttle is unconditional**, so it skips the rollback-time re-derivation that `state.go:1780-1782` explicitly relies on. See inline on `ledger/state.go`.
3. **Zero new tests for ~180 LOC of consensus-affecting logic.** No coverage for `findRatifiedHFICandidate` (single / none / multi / no-parent / no-root), no test that imported proposals carry the new fields, no test of throttle behavior across rollback. The existing test suite passes only because each test creates a fresh `LedgerState` (so `lastHFIStabilityEvalAt` starts zero).

Suggested split: land the CBOR plumbing here, move the HFI heuristic and throttle to follow-ups so each carries its own test surface and risk discussion.

Note: this PR is stacked on unmerged #2201 — needs to land first.

FYI the four G115 gosec findings on `ledger/state.go:2462,4266,4350,4366` are from commit `d8d01dff7` (May 2), not this PR — calling that out so they don't read as gates here.

wolf31o2:
@cubic-dev-ai review

wolf31o2:
@cubic-dev-ai review

wolf31o2:
@cubic-dev-ai review

agaffney (approved):

chrisguiney (approved):

PR #2203, opened 2026-05-08
feat(backfill):  Wire batch accumulator into backfill pipeline

### LINKED ISSUE #1801: Wire batch accumulator into backfill pipeline
## Summary

Replace the per-TX `db.SetTransaction()` call in the backfill loop with `db.SetTransactionBatched()`, accumulating across multiple blocks and flushing periodically. This is the integration point that delivers the actual speedup.

Part of the backfill batching optimization.

## Design

Add a `backfillBatchSize = 100` constant controlling how many blocks are accumulated before flushing. The main loop in `Backfill.Run()` creates a `BatchAccumulator` and a DB transaction, processes blocks using `processBlockTxsBatched()`, and flushes every `backfillBatchSize` blocks.

### `processBlockTxsBatched`

A new method parallel to `processBlockTxs` that:
- Calls `db.SetTransactionBatched()` instead of `db.SetTransaction()` for each TX
- Certificates still processed per-TX immediately (complex 17+ type switch, already fast)
- Governance still processed per-TX immediately

### Flush logic

```
batchBlockCount++
if batchBlockCount >= backfillBatchSize {
    db.FlushBatch(acc, batchTxn)
    batchTxn.Commit()
    acc.Reset()
    batchBlockCount = 0
}
```

Also flush any remaining rows after the iteration completes before marking complete.

### `internal/node/backfill.go`

- Add `backfillBatchSize` constant
- Add `processBlockTxsBatched()` method
- Modify `Run()` to use batch accumulator with periodic flush
- Flush remaining rows at end of iteration

## Deliverables

- [ ] `processBlockTxsBatched` using `SetTransactionBatched` instead of `SetTransaction`
- [ ] Main loop accumulates across blocks and flushes every 100 blocks
- [ ] Remaining rows flushed before completion
- [ ] Certificates and governance still processed per-TX
- [ ] Existing backfill tests pass
- [ ] Manual verification: `dingo mithril sync` on preview with `DINGO_STORAGE_MODE=api` produces correct data

arepala-uml:
1. Added backfillBatchSize = 100 to control how many blocks are accumulated before flushing.
2. Made changes in backfill.Run() to create a batch accumulator before iterating blocks, opens one DB transaction for the active batch  and added a tracker how many blocks are accumulated with batchBlockCount.
3. Flushes the batch for every backfillBatchSize blocks, resets the accumulator and starts a new transaction after each flush.
4. Made changes to saves checkpoints only up to the last committed batch slot and flushes the final partial batch before marking backfill complete.
5. Added a database-level BatchAccumulator wrapper for creating the correct metadata plugin accumulator for sqlite, postgres and mysql.
6. Added database level methods NewBatchAccumulator(), FlushBatch(), SetTransactionBatched(), SetTransactionBatched() mirrors the existing SetTransaction() flow, but calls the metadata plugin’s SetTransactionBatched() instead of SetTransaction().

Closes #1801

---
## Summary by cubic
Batch backfill now buffers metadata across 100 blocks with a shared accumulator and commits once per batch to cut DB commits and speed up historical replay. Checkpoints advance only to committed batches; we flush on cancel, before periodic checkpoints, and at completion, and on errors we save the last committed checkpoint for safe resume.

- **New Features**
  - Shared backfill accumulator: one transaction per batch, flush every 100 blocks, then reset.
  - Database batching API: `NewBatchAccumulator()`, `FlushBatch()`, `SetTransactionBatched()` on `Database`/`MetadataStore`; opaque `types.MetadataBatchAccumulator` with plugin type checks.
  - Offsets are written to blob immediately; batched path via `processBlockTxsBatched`.

- **Bug Fixes**
  - Checkpointing records only durable slots; we flush current batches on cancel, before checkpoints, and at final flush; on errors (including final flush) we save the last committed checkpoint and reset the owned accumulator.
  - Fixed pool registration cleanup by deleting `pool_registration_owner`/`pool_registration_relay` before parent rows across all stores.
  - Batched ingestion handles cross-block UTxO spends within the same batch; added a test.
  - SQLite: preserve UTxO spend conflict checks in batches, skip already-spent rows when updating batch spends, warn when inputs are missing, and increase bulk-insert chunk to 500.

<sup>Written for commit f77d133d4d4cc71a7df387748deba59ea36ddcb6. Summary will update on new commits.</sup>

## Release Notes

* **New Features**
  * Added support for batched metadata ingestion during block backfilling to improve performance and reliability.

* **Refactor**
  * Transformed transaction metadata persistence from individual per-block commits to batch-based buffering. Metadata is now accumulated and written together, reducing database write overhead and improving checkpoint consistency.

* **Tests**
  * Added regression test verifying correct handling of cross-block transaction spending within batched operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/blinklabs-io/dingo/pull/2203)

### DIFF
--- a/database/batch.go
+++ b/database/batch.go
@@ -70,12 +70,14 @@ func (d *Database) SetTransactionBatched(
 		return errors.New("batch accumulator must not be nil")
 	}
 	owned := false
-	accDirty := false
 	if txn == nil {
 		txn = d.Transaction(true)
 		owned = true
 		defer func() {
-			if retErr != nil && accDirty {
+			if txn == nil {
+				return
+			}
+			if retErr != nil {
 				acc.Reset()
 			}
 			_ = txn.Rollback()
@@ -149,7 +151,6 @@ func (d *Database) SetTransactionBatched(
 	if metadataTxn == nil {
 		return types.ErrNilTxn
 	}
-	accDirty = owned
 	if err := d.metadata.SetTransactionBatched(
 		tx, point, idx, certDeposits, acc, metadataTxn,
 	); err != nil {
@@ -177,6 +178,7 @@ func (d *Database) SetTransactionBatched(
 		if err := txn.Commit(); err != nil {
 			return err
 		}
+		txn = nil
 		acc.Reset()
 	}
 
--- a/database/batch.go
+++ b/database/batch.go
@@ -1,0 +1,243 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package database
+
+import (
+	"encoding/hex"
+	"fmt"
+
+	"github.com/blinklabs-io/dingo/database/plugin/metadata/mysql"
+	"github.com/blinklabs-io/dingo/database/plugin/metadata/postgres"
+	"github.com/blinklabs-io/dingo/database/plugin/metadata/sqlite"
+	"github.com/blinklabs-io/dingo/database/types"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
+)
+
+// BatchAccumulator wraps the concrete accumulator used by the active
+// metadata plugin.
+type BatchAccumulator struct {
+	sqlite   *sqlite.BatchAccumulator
+	postgres *postgres.BatchAccumulator
+	mysql    *mysql.BatchAccumulator
+}
+
+// NewBatchAccumulator creates an accumulator for the configured metadata
+// plugin.
+func (d *Database) NewBatchAccumulator() (*BatchAccumulator, error) {
+	switch d.metadata.(type) {
+	case *sqlite.MetadataStoreSqlite:
+		return &BatchAccumulator{sqlite: sqlite.NewBatchAccumulator()}, nil
+	case *postgres.MetadataStorePostgres:
+		return &BatchAccumulator{postgres: postgres.NewBatchAccumulator()}, nil
+	case *mysql.MetadataStoreMysql:
+		return &BatchAccumulator{mysql: mysql.NewBatchAccumulator()}, nil
+	default:
+		return nil, fmt.Errorf(
+			"batch accumulator unsupported for metadata store %T",
+			d.metadata,
+		)
+	}
+}
+
+// Reset clears all accumulated rows while reusing backing storage.
+func (b *BatchAccumulator) Reset() {
+	if b == nil {
+		return
+	}
+	switch {
+	case b.sqlite != nil:
+		b.sqlite.Reset()
+	case b.postgres != nil:
+		b.postgres.Reset()
+	case b.mysql != nil:
+		b.mysql.Reset()
+	}
+}
+
+// FlushBatch writes accumulated metadata rows for the active metadata plugin.
+func (d *Database) FlushBatch(
+	acc *BatchAccumulator,
+	txn *Txn,
+) error {
+	if acc == nil {
+		return nil
+	}
+	var metadataTxn types.Txn
+	if txn != nil {
+		metadataTxn = txn.Metadata()
+	}
+	switch store := d.metadata.(type) {
+	case *sqlite.MetadataStoreSqlite:
+		return store.FlushBatch(acc.sqlite, metadataTxn)
+	case *postgres.MetadataStorePostgres:
+		return store.FlushBatch(acc.postgres, metadataTxn)
+	case *mysql.MetadataStoreMysql:
+		return store.FlushBatch(acc.mysql, metadataTxn)
+	default:
+		return fmt.Errorf(
+			"flush batch unsupported for metadata store %T",
+			d.metadata,
+		)
+	}
+}
+
+// SetTransactionBatched stores transaction blob offsets and immediate
+// metadata, while accumulating bulk metadata rows into acc for a later
+// FlushBatch.
+func (d *Database) SetTransactionBatched(
+	tx lcommon.Transaction,
+	point ocommon.Point,
+	idx uint32,
+	updateEpoch uint64,
+	pparamUpdates map[lcommon.Blake2b224]lcommon.ProtocolParameterUpdate,
+	certDeposits map[int]uint64,
+	offsets *BlockIngestionResult,
+	acc *BatchAccumulator,
+	txn *Txn,
+) error {
+	if acc == nil {
+		return fmt.Errorf("batch accumulator must not be nil")
+	}
+	owned := false
+	if txn == nil {
+		txn = d.Transaction(true)
+		owned = true
+		defer txn.Rollback() //nolint:errcheck
+	}
+
+	blob := txn.DB().Blob()
+	if blob == nil {
+		return types.ErrBlobStoreUnavailable
+	}
+	blobTxn := txn.Blob()
+	if blobTxn == nil {
+		return types.ErrNilTxn
+	}
+
+	txHash := tx.Hash()
+	var txHashArray [32]byte
+	copy(txHashArray[:], txHash.Bytes())
+
+	if offsets == nil {
+		return fmt.Errorf(
+			"missing offsets for transaction %s at slot %d: offsets must be computed",
+			hex.EncodeToString(txHash.Bytes()[:8]),
+			point.Slot,
+		)
+	}
+	txOffset, ok := offsets.TxOffsets[txHashArray]
+	if !ok {
+		return fmt.Errorf(
+			"missing TX offset for %s at slot %d: offset must be computed by block indexer",
+			hex.EncodeToString(txHash.Bytes()[:8]),
+			point.Slot,
+		)
+	}
+	offsetData := EncodeTxOffset(&txOffset)
+	if err := blob.SetTx(blobTxn, txHash.Bytes(), offsetData); err != nil {
+		return fmt.Errorf("set tx offset: %w", err)
+	}
+
+	for _, utxo := range tx.Produced() {
+		txID := utxo.Id.Id().Bytes()
+		outputIdx := utxo.Id.Index()
+		ref := UtxoRef{
+			TxId:      txHashArray,
+			OutputIdx: outputIdx,
+		}
+		offset, ok := offsets.UtxoOffsets[ref]
+		if !ok {
+			return fmt.Errorf(
+				"missing UTxO offset for %s#%d at slot %d: offset must be computed by block indexer",
+				hex.EncodeToString(txID[:8]),
+				outputIdx,
+				point.Slot,
+			)
+		}
+		offsetData := EncodeUtxoOffset(&offset)
+		if err := blob.SetUtxo(blobTxn, txID, outputIdx, offsetData); err != nil {
+			return fmt.Errorf(
+				"set utxo offset %x#%d: %w",
+				txID[:8],
+				outputIdx,
+				err,
+			)
+		}
+	}
+
+	if err := d.ensureTransactionConsumedUtxos(tx, point, txn); err != nil {
+		return err
+	}
+	if err := d.setTransactionMetadataBatched(
+		tx, point, idx, certDeposits, acc, txn.Metadata(),
+	); err != nil {
+		return fmt.Errorf("set transaction metadata: %w", err)
+	}
+
+	if updateEpoch > 0 && tx.IsValid() {
+		for genesisHash, update := range pparamUpdates {
+			if err := d.SetPParamUpdate(
+				genesisHash.Bytes(),
+				update.Cbor(),
+				point.Slot,
+				updateEpoch,
+				txn,
+			); err != nil {
+				return fmt.Errorf("set pparam update: %w", err)
+			}
+		}
+	}
+
+	if owned {
+		if err := d.FlushBatch(acc, txn); err != nil {
+			return err
+		}
+		if err := txn.Commit(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (d *Database) setTransactionMetadataBatched(
+	tx lcommon.Transaction,
+	point ocommon.Point,
+	idx uint32,
+	certDeposits map[int]uint64,
+	acc *BatchAccumulator,
+	txn types.Txn,
+) error {
+	switch store := d.metadata.(type) {
+	case *sqlite.MetadataStoreSqlite:
+		return store.SetTransactionBatched(
+			tx, point, idx, certDeposits, acc.sqlite, txn,
+		)
+	case *postgres.MetadataStorePostgres:
+		return store.SetTransactionBatched(
+			tx, point, idx, certDeposits, acc.postgres, txn,
+		)
+	case *mysql.MetadataStoreMysql:
+		return store.SetTransactionBatched(
+			tx, point, idx, certDeposits, acc.mysql, txn,
+		)
+	default:
+		return fmt.Errorf(
+			"batched transaction unsupported for metadata store %T",
+			d.metadata,
+		)
+	}
+}
--- a/database/batch.go
+++ b/database/batch.go
@@ -79,6 +79,9 @@ func (d *Database) FlushBatch(
 	var metadataTxn types.Txn
 	if txn != nil {
 		metadataTxn = txn.Metadata()
+		if metadataTxn == nil {
+			return types.ErrNilTxn
+		}
 	}
 	switch store := d.metadata.(type) {
 	case *sqlite.MetadataStoreSqlite:
@@ -182,8 +185,12 @@ func (d *Database) SetTransactionBatched(
 	if err := d.ensureTransactionConsumedUtxos(tx, point, txn); err != nil {
 		return err
 	}
+	metadataTxn := txn.Metadata()
+	if metadataTxn == nil {
+		return types.ErrNilTxn
+	}
 	if err := d.setTransactionMetadataBatched(
-		tx, point, idx, certDeposits, acc, txn.Metadata(),
+		tx, point, idx, certDeposits, acc, metadataTxn,
 	); err != nil {
 		return fmt.Errorf("set transaction metadata: %w", err)
 	}
--- a/database/batch.go
+++ b/database/batch.go
@@ -65,15 +65,21 @@ func (d *Database) SetTransactionBatched(
 	offsets *BlockIngestionResult,
 	acc BatchAccumulator,
 	txn *Txn,
-) error {
+) (retErr error) {
 	if acc == nil {
 		return errors.New("batch accumulator must not be nil")
 	}
 	owned := false
+	accDirty := false
 	if txn == nil {
 		txn = d.Transaction(true)
 		owned = true
-		defer txn.Rollback() //nolint:errcheck
+		defer func() {
+			if retErr != nil && accDirty {
+				acc.Reset()
+			}
+			_ = txn.Rollback()
+		}()
 	}
 
 	blob := txn.DB().Blob()
@@ -143,6 +149,7 @@ func (d *Database) SetTransactionBatched(
 	if metadataTxn == nil {
 		return types.ErrNilTxn
 	}
+	accDirty = owned
 	if err := d.metadata.SetTransactionBatched(
 		tx, point, idx, certDeposits, acc, metadataTxn,
 	); err != nil {
--- a/database/batch.go
+++ b/database/batch.go
@@ -16,6 +16,7 @@ package database
 
 import (
 	"encoding/hex"
+	"errors"
 	"fmt"
 
 	"github.com/blinklabs-io/dingo/database/plugin/metadata/mysql"
@@ -109,7 +110,7 @@ func (d *Database) SetTransactionBatched(
 	txn *Txn,
 ) error {
 	if acc == nil {
-		return fmt.Errorf("batch accumulator must not be nil")
+		return errors.New("batch accumulator must not be nil")
 	}
 	owned := false
 	if txn == nil {
--- a/database/batch.go
+++ b/database/batch.go
@@ -209,6 +209,7 @@ func (d *Database) SetTransactionBatched(
 		if err := txn.Commit(); err != nil {
 			return err
 		}
+		acc.Reset()
 	}
 
 	return nil
--- a/database/batch.go
+++ b/database/batch.go
@@ -19,58 +19,24 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/blinklabs-io/dingo/database/plugin/metadata/mysql"
-	"github.com/blinklabs-io/dingo/database/plugin/metadata/postgres"
-	"github.com/blinklabs-io/dingo/database/plugin/metadata/sqlite"
 	"github.com/blinklabs-io/dingo/database/types"
 	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
 	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
 )
 
-// BatchAccumulator wraps the concrete accumulator used by the active
-// metadata plugin.
-type BatchAccumulator struct {
-	sqlite   *sqlite.BatchAccumulator
-	postgres *postgres.BatchAccumulator
-	mysql    *mysql.BatchAccumulator
-}
+// BatchAccumulator is an opaque accumulator owned by the active metadata
+// plugin.
+type BatchAccumulator = types.MetadataBatchAccumulator
 
 // NewBatchAccumulator creates an accumulator for the configured metadata
 // plugin.
-func (d *Database) NewBatchAccumulator() (*BatchAccumulator, error) {
-	switch d.metadata.(type) {
-	case *sqlite.MetadataStoreSqlite:
-		return &BatchAccumulator{sqlite: sqlite.NewBatchAccumulator()}, nil
-	case *postgres.MetadataStorePostgres:
-		return &BatchAccumulator{postgres: postgres.NewBatchAccumulator()}, nil
-	case *mysql.MetadataStoreMysql:
-		return &BatchAccumulator{mysql: mysql.NewBatchAccumulator()}, nil
-	default:
-		return nil, fmt.Errorf(
-			"batch accumulator unsupported for metadata store %T",
-			d.metadata,
-		)
-	}
-}
-
-// Reset clears all accumulated rows while reusing backing storage.
-func (b *BatchAccumulator) Reset() {
-	if b == nil {
-		return
-	}
-	switch {
-	case b.sqlite != nil:
-		b.sqlite.Reset()
-	case b.postgres != nil:
-		b.postgres.Reset()
-	case b.mysql != nil:
-		b.mysql.Reset()
-	}
+func (d *Database) NewBatchAccumulator() BatchAccumulator {
+	return d.metadata.NewBatchAccumulator()
 }
 
 // FlushBatch writes accumulated metadata rows for the active metadata plugin.
 func (d *Database) FlushBatch(
-	acc *BatchAccumulator,
+	acc BatchAccumulator,
 	txn *Txn,
 ) error {
 	if acc == nil {
@@ -83,19 +49,7 @@ func (d *Database) FlushBatch(
 			return types.ErrNilTxn
 		}
 	}
-	switch store := d.metadata.(type) {
-	case *sqlite.MetadataStoreSqlite:
-		return store.FlushBatch(acc.sqlite, metadataTxn)
-	case *postgres.MetadataStorePostgres:
-		return store.FlushBatch(acc.postgres, metadataTxn)
-	case *mysql.MetadataStoreMysql:
-		return store.FlushBatch(acc.mysql, metadataTxn)
-	default:
-		return fmt.Errorf(
-			"flush batch unsupported for metadata store %T",
-			d.metadata,
-		)
-	}
+	return d.metadata.FlushBatch(acc, metadataTxn)
 }
 
 // SetTransactionBatched stores transaction blob offsets and immediate
@@ -109,7 +63,7 @@ func (d *Database) SetTransactionBatched(
 	pparamUpdates map[lcommon.Blake2b224]lcommon.ProtocolParameterUpdate,
 	certDeposits map[int]uint64,
 	offsets *BlockIngestionResult,
-	acc *BatchAccumulator,
+	acc BatchAccumulator,
 	txn *Txn,
 ) error {
 	if acc == nil {
@@ -189,7 +143,7 @@ func (d *Database) SetTransactionBatched(
 	if metadataTxn == nil {
 		return types.ErrNilTxn
 	}
-	if err := d.setTransactionMetadataBatched(
+	if err := d.metadata.SetTransactionBatched(
 		tx, point, idx, certDeposits, acc, metadataTxn,
 	); err != nil {
 		return fmt.Errorf("set transaction metadata: %w", err)
@@ -220,33 +174,4 @@ func (d *Database) SetTransactionBatched(
 	}
 
 	return nil
-}
-
-func (d *Database) setTransactionMetadataBatched(
-	tx lcommon.Transaction,
-	point ocommon.Point,
-	idx uint32,
-	certDeposits map[int]uint64,
-	acc *BatchAccumulator,
-	txn types.Txn,
-) error {
-	switch store := d.metadata.(type) {
-	case *sqlite.MetadataStoreSqlite:
-		return store.SetTransactionBatched(
-			tx, point, idx, certDeposits, acc.sqlite, txn,
-		)
-	case *postgres.MetadataStorePostgres:
-		return store.SetTransactionBatched(
-			tx, point, idx, certDeposits, acc.postgres, txn,
-		)
-	case *mysql.MetadataStoreMysql:
-		return store.SetTransactionBatched(
-			tx, point, idx, certDeposits, acc.mysql, txn,
-		)
-	default:
-		return fmt.Errorf(
-			"batched transaction unsupported for metadata store %T",
-			d.metadata,
-		)
-	}
 }
--- a/database/plugin/metadata/sqlite/batch_accumulator.go
+++ b/database/plugin/metadata/sqlite/batch_accumulator.go
@@ -15,6 +15,8 @@
 package sqlite
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
 
 	"github.com/blinklabs-io/dingo/database/models"
@@ -213,7 +215,7 @@ func (d *MetadataStoreSqlite) FlushBatch(
 				err,
 			)
 		}
-		if err := batchSpendUtxos(db, batch.UtxoSpends); err != nil {
+		if err := d.batchSpendUtxos(db, batch.UtxoSpends); err != nil {
 			return fmt.Errorf("flush batch: spend utxos: %w", err)
 		}
 
@@ -294,7 +296,7 @@ func batchDeleteByTxIDs(db *gorm.DB, table string, ids []uint) error {
 	return nil
 }
 
-func batchSpendUtxos(db *gorm.DB, spends []utxoSpend) error {
+func (d *MetadataStoreSqlite) batchSpendUtxos(db *gorm.DB, spends []utxoSpend) error {
 	if len(spends) == 0 {
 		return nil
 	}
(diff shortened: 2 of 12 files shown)


database/batch.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+	}
+
+	if owned {
+		if err := d.FlushBatch(acc, txn); err != nil {
+			return err
+		}
+		if err := txn.Commit(); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
arepala-uml: Added acc.Reset() after the self-managed FlushBatch() plus Commit() path so callers that pass txn == nil do not accidentally reuse an accumulator containing already-flushed rows.

internal/node/backfill.go
```diff
@@ -648,22 +652,75 @@ func (b *Backfill) Run(ctx context.Context) error {
...
+		if err := b.db.FlushBatch(acc, batchTxn); err != nil {
+			return err
+		}
+		if err := batchTxn.Commit(); err != nil {
+			return err
+		}
+		committedSlot = cp.LastSlot
+		acc.Reset()
+		batchTxn = b.db.Transaction(true)
+		batchBlockCount = 0
```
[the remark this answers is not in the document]
arepala-uml: Updated flushBatch() to release the committed transaction before replacing it with a new batch transaction, avoiding leaked transaction handles during long backfills.

arepala-uml:
Regarding the manual validation by mithril sync - I was running into the storage issue after the snapshot was downloaded and extracted, during the stage where Dingo was loading immutable blocks into the Badger blob store.
The key failure was it ran out of disk while writing Dingo’s local database files under (.dingo/blob) and it did not reach the API metadata backfill stage. So, I could not validate new backfill changes. 

```
akhil-mac@Akhils-MacBook-Air dingo % DINGO_STORAGE_MODE=api ./dingo mithril sync --network preview
{"time":"2026-05-07T18:24:18.834554-05:00","level":"INFO","msg":"maxprocs: Leaving GOMAXPROCS=8: CPU quota undefined","component":"dingo"}
{"time":"2026-05-07T18:24:18.834809-05:00","level":"INFO","msg":"version: devel (commit )","component":"dingo"}
{"time":"2026-05-07T18:24:18.840943-05:00","level":"INFO","msg":"starting Mithril bootstrap","component":"mithril","network":"preview","aggregator":"https://aggregator.pre-release-preview.api.mithril.network/aggregator"}
{"time":"2026-05-07T18:24:19.411736-05:00","level":"INFO","msg":"found latest snapshot","component":"mithril","digest":"35f9767bb7979bb9cf29ffb7864874c94dc606a14e18cac741443ba693e257d3","epoch":1290,"immutable_file_number":25815,"size":2951961964}
{"time":"2026-05-07T18:24:19.411806-05:00","level":"INFO","msg":"verifying certificate chain","component":"mithril","certificate_hash":"862bfb675891e916e3c0e27340e39ad67bb20083037fec8c7a4bcb2ed66cb54b"}
{"time":"2026-05-07T18:24:41.533319-05:00","level":"INFO","msg":"certificate chain verified","component":"mithril"}
{"time":"2026-05-07T18:24:42.4764-05:00","level":"INFO","msg":"downloading snapshot","component":"mithril","url":"https://storage.googleapis.com/cdn.aggregator.pre-release-preview.api.mithril.network/cardano-immutable-files-full/preview-e1290-i25815.35f9767bb7979bb9cf29ffb7864874c94dc606a14e18cac741443ba693e257d3.tar.zst","total_bytes":2951961964,"destination":".dingo/.mithril-cache/preview-35f9767bb7979bb9.tar.zst"}
{"time":"2026-05-07T18:24:42.477102-05:00","level":"INFO","msg":"download progress: 0.0% (14.7 KB / 2.7 GB) at 52.0 MB/s","component":"mithril"}
{"time":"2026-05-07T18:24:48.507548-05:00","level":"INFO","msg":"download progress: 5.2% (145.1 MB / 2.7 GB) at 24.1 MB/s","component":"mithril"}
{"time":"2026-05-07T18:24:53.024714-05:00","level":"INFO","msg":"download progress: 10.2% (287.1 MB / 2.7 GB) at 27.2 MB/s","component":"mithril"}
{"time":"2026-05-07T18:24:58.065848-05:00","level":"INFO","msg":"download progress: 15.5% (435.3 MB / 2.7 GB) at 27.9 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:03.083271-05:00","level":"INFO","msg":"download progress: 21.0% (590.0 MB / 2.7 GB) at 28.6 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:08.223588-05:00","level":"INFO","msg":"download progress: 26.5% (746.7 MB / 2.7 GB) at 29.0 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:12.766462-05:00","level":"INFO","msg":"download progress: 31.5% (887.8 MB / 2.7 GB) at 29.3 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:17.809062-05:00","level":"INFO","msg":"download progress: 36.8% (1.0 GB / 2.7 GB) at 29.3 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:22.844668-05:00","level":"INFO","msg":"download progress: 41.8% (1.2 GB / 2.7 GB) at 29.2 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:27.851825-05:00","level":"INFO","msg":"download progress: 47.1% (1.3 GB / 2.7 GB) at 29.2 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:32.941763-05:00","level":"INFO","msg":"download progress: 52.2% (1.4 GB / 2.7 GB) at 29.1 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:37.963757-05:00","level":"INFO","msg":"download progress: 57.3% (1.6 GB / 2.7 GB) at 29.1 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:42.99789-05:00","level":"INFO","msg":"download progress: 62.6% (1.7 GB / 2.7 GB) at 29.1 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:48.50952-05:00","level":"INFO","msg":"download progress: 68.1% (1.9 GB / 2.7 GB) at 29.0 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:53.019234-05:00","level":"INFO","msg":"download progress: 73.1% (2.0 GB / 2.7 GB) at 29.2 MB/s","component":"mithril"}
{"time":"2026-05-07T18:25:58.066254-05:00","level":"INFO","msg":"download progress: 78.6% (2.2 GB / 2.7 GB) at 29.3 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:03.577002-05:00","level":"INFO","msg":"download progress: 83.9% (2.3 GB / 2.7 GB) at 29.1 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:08.110285-05:00","level":"INFO","msg":"download progress: 89.0% (2.4 GB / 2.7 GB) at 29.2 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:13.124789-05:00","level":"INFO","msg":"download progress: 94.2% (2.6 GB / 2.7 GB) at 29.3 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:18.14041-05:00","level":"INFO","msg":"download progress: 99.6% (2.7 GB / 2.7 GB) at 29.3 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:18.519392-05:00","level":"INFO","msg":"download progress: 100.0% (2.7 GB / 2.7 GB) at 29.3 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:18.519938-05:00","level":"INFO","msg":"download complete","component":"mithril","bytes":2951961964,"path":".dingo/.mithril-cache/preview-35f9767bb7979bb9.tar.zst"}
{"time":"2026-05-07T18:26:18.520914-05:00","level":"INFO","msg":"download size verified","component":"mithril","bytes":2951961964}
{"time":"2026-05-07T18:26:18.526085-05:00","level":"INFO","msg":"downloading ancillary data (ledger state)","component":"mithril","size":448749427}
{"time":"2026-05-07T18:26:18.5441-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"0.0%","archive_bytes_read":"178.6 KB","archive_size":"2.7 GB","files_extracted":1}
{"time":"2026-05-07T18:26:18.832091-05:00","level":"INFO","msg":"downloading snapshot","component":"mithril","url":"https://storage.googleapis.com/cdn.aggregator.pre-release-preview.api.mithril.network/cardano-immutable-files-full/preview-e1290-i25815.35f9767bb7979bb9cf29ffb7864874c94dc606a14e18cac741443ba693e257d3.ancillary.tar.zst","total_bytes":448749427,"destination":".dingo/.mithril-cache/preview-35f9767bb7979bb9-ancillary.tar.zst"}
{"time":"2026-05-07T18:26:20.073692-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"5.0%","archive_bytes_read":"140.9 MB","archive_size":"2.7 GB","files_extracted":2533}
{"time":"2026-05-07T18:26:21.682004-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"10.0%","archive_bytes_read":"281.8 MB","archive_size":"2.7 GB","files_extracted":3970}
{"time":"2026-05-07T18:26:22.805068-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"15.0%","archive_bytes_read":"422.7 MB","archive_size":"2.7 GB","files_extracted":5851}
{"time":"2026-05-07T18:26:24.398419-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"20.0%","archive_bytes_read":"563.5 MB","archive_size":"2.7 GB","files_extracted":7790}
{"time":"2026-05-07T18:26:26.273861-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"25.0%","archive_bytes_read":"704.8 MB","archive_size":"2.7 GB","files_extracted":8419}
{"time":"2026-05-07T18:26:28.060891-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"30.0%","archive_bytes_read":"845.6 MB","archive_size":"2.7 GB","files_extracted":11992}
{"time":"2026-05-07T18:26:28.890247-05:00","level":"INFO","msg":"download progress: 58.7% (251.1 MB / 428.0 MB) at 25.0 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:29.888596-05:00","level":"INFO","msg":"download progress: 65.9% (282.2 MB / 428.0 MB) at 25.5 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:29.9853-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"35.0%","archive_bytes_read":"986.4 MB","archive_size":"2.7 GB","files_extracted":17038}
{"time":"2026-05-07T18:26:30.909986-05:00","level":"INFO","msg":"download progress: 73.3% (313.9 MB / 428.0 MB) at 26.0 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:31.911889-05:00","level":"INFO","msg":"download progress: 80.9% (346.1 MB / 428.0 MB) at 26.5 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:32.430152-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"40.0%","archive_bytes_read":"1.1 GB","archive_size":"2.7 GB","files_extracted":22342}
{"time":"2026-05-07T18:26:32.918622-05:00","level":"INFO","msg":"download progress: 87.5% (374.3 MB / 428.0 MB) at 26.6 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:33.927547-05:00","level":"INFO","msg":"download progress: 95.1% (406.9 MB / 428.0 MB) at 27.0 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:34.608649-05:00","level":"INFO","msg":"download progress: 100.0% (428.0 MB / 428.0 MB) at 27.1 MB/s","component":"mithril"}
{"time":"2026-05-07T18:26:34.609491-05:00","level":"INFO","msg":"download complete","component":"mithril","bytes":448749427,"path":".dingo/.mithril-cache/preview-35f9767bb7979bb9-ancillary.tar.zst"}
{"time":"2026-05-07T18:26:34.609505-05:00","level":"INFO","msg":"download size verified","component":"mithril","bytes":448749427}
{"time":"2026-05-07T18:26:34.643226-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"0.0%","archive_bytes_read":"69.4 KB","archive_size":"428.0 MB","files_extracted":1}
{"time":"2026-05-07T18:26:35.248277-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"45.0%","archive_bytes_read":"1.2 GB","archive_size":"2.7 GB","files_extracted":25969}
{"time":"2026-05-07T18:26:37.945918-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"49.9%","archive_bytes_read":"213.5 MB","archive_size":"428.0 MB","files_extracted":6}
{"time":"2026-05-07T18:26:40.040703-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"50.0%","archive_bytes_read":"1.4 GB","archive_size":"2.7 GB","files_extracted":30130}
{"time":"2026-05-07T18:26:41.236233-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"100.0%","archive_bytes_read":"428.0 MB","archive_size":"428.0 MB","files_extracted":9}
{"time":"2026-05-07T18:26:41.24044-05:00","level":"INFO","msg":"extraction complete","component":"mithril","files_extracted":10,"progress":"100.0%","archive_size":"428.0 MB","destination":".dingo/.mithril-cache/ancillary"}
{"time":"2026-05-07T18:26:41.243669-05:00","level":"INFO","msg":"ancillary data extracted","component":"mithril","path":".dingo/.mithril-cache/ancillary"}
{"time":"2026-05-07T18:26:42.760278-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"55.0%","archive_bytes_read":"1.5 GB","archive_size":"2.7 GB","files_extracted":31362}
{"time":"2026-05-07T18:26:45.219592-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"60.0%","archive_bytes_read":"1.7 GB","archive_size":"2.7 GB","files_extracted":35389}
{"time":"2026-05-07T18:26:47.525254-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"65.0%","archive_bytes_read":"1.8 GB","archive_size":"2.7 GB","files_extracted":40402}
{"time":"2026-05-07T18:26:50.121171-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"70.0%","archive_bytes_read":"1.9 GB","archive_size":"2.7 GB","files_extracted":45977}
{"time":"2026-05-07T18:26:52.744404-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"75.1%","archive_bytes_read":"2.1 GB","archive_size":"2.7 GB","files_extracted":51346}
{"time":"2026-05-07T18:26:57.745176-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"80.1%","archive_bytes_read":"2.2 GB","archive_size":"2.7 GB","files_extracted":56009}
{"time":"2026-05-07T18:27:07.744676-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"85.1%","archive_bytes_read":"2.3 GB","archive_size":"2.7 GB","files_extracted":59686}
{"time":"2026-05-07T18:27:13.896276-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"90.1%","archive_bytes_read":"2.5 GB","archive_size":"2.7 GB","files_extracted":65416}
{"time":"2026-05-07T18:27:20.809886-05:00","level":"INFO","msg":"extracting archive","component":"mithril","progress":"95.1%","archive_bytes_read":"2.6 GB","archive_size":"2.7 GB","files_extracted":71799}
{"time":"2026-05-07T18:27:22.126759-05:00","level":"INFO","msg":"extraction complete","component":"mithril","files_extracted":77448,"progress":"100.0%","archive_size":"2.7 GB","destination":".dingo/.mithril-cache/immutable"}
{"time":"2026-05-07T18:27:22.190794-05:00","level":"INFO","msg":"Mithril bootstrap ready for loading","component":"mithril","immutable_dir":".dingo/.mithril-cache/immutable/immutable","ancillary_dir":".dingo/.mithril-cache/ancillary"}
{"time":"2026-05-07T18:27:22.348153-05:00","level":"INFO","msg":"badger blob store opened","component":"database","block_cache_size_mb":256,"index_cache_size_mb":0,"data_dir":".dingo"}
{"time":"2026-05-07T18:27:22.441788-05:00","level":"INFO","msg":"node settings initialized","storageMode":"api","network":""}
{"time":"2026-05-07T18:27:22.444417-05:00","level":"INFO","msg":"set bulk-load PRAGMAs (synchronous=OFF, cache=200MB, temp=MEMORY)"}
{"time":"2026-05-07T18:27:22.445035-05:00","level":"INFO","msg":"loading ImmutableDB blocks into blob store","component":"mithril","immutable_dir":".dingo/.mithril-cache/immutable/immutable"}
{"time":"2026-05-07T18:27:22.446016-05:00","level":"INFO","msg":"found ledger state file","component":"mithril","path":".dingo/.mithril-cache/ancillary/ledger/111525678/state"}
{"time":"2026-05-07T18:27:22.523459-05:00","level":"INFO","msg":"copying blocks from immutable DB (header-only decode)"}
{"time":"2026-05-07T18:27:22.609037-05:00","level":"INFO","msg":"copying blocks from immutable DB","blocks_copied":50,"slot":980,"blocks_per_sec":"4158","progress":"0.0%"}
{"time":"2026-05-07T18:27:22.979163-05:00","level":"INFO","msg":"found UTxO table file (UTxO-HD format)","component":"mithril","path":".dingo/.mithril-cache/ancillary/ledger/111525678/tables"}
{"time":"2026-05-07T18:27:22.97918-05:00","level":"INFO","msg":"parsed ledger state","component":"mithril","era":"Conway","epoch":1290,"slot":111525678,"era_bound_slot":55814400,"era_bound_epoch":646,"epoch_nonce":"d56e360a7ef9676b9b04015f848681ba8b75f28782550dd9446ca4cf61514fb0"}
{"time":"2026-05-07T18:27:22.981571-05:00","level":"INFO","msg":"starting ledger state import","component":"ledgerstate","epoch":1290,"era":"Conway","slot":111525678}
{"time":"2026-05-07T18:27:22.981579-05:00","level":"INFO","msg":"importing UTxOs from ledger state","component":"ledgerstate","source":".dingo/.mithril-cache/ancillary/ledger/111525678/tables"}
{"time":"2026-05-07T18:27:25.620704-05:00","level":"INFO","msg":"UTxO import progress","component":"mithril","stage":"utxo","progress":"0.3%","current":10000}
{"time":"2026-05-07T18:27:39.878194-05:00","level":"INFO","msg":"copying blocks from immutable DB","blocks_copied":87450,"slot":1872103,"blocks_per_sec":"5063","progress":"1.7%"}
{"time":"2026-05-07T18:27:41.952736-05:00","level":"INFO","msg":"UTxO import progress","component":"mithril","stage":"utxo","progress":"1.6%","current":50000}
{"time":"2026-05-07T18:27:49.909114-05:00","level":"INFO","msg":"copying blocks from immutable DB","blocks_copied":115750,"slot":2449676,"blocks_per_sec":"4240","progress":"2.2%"}
{"time":"2026-05-07T18:27:53.917407-05:00","level":"INFO","msg":"UTxO import progress","component":"mithril","stage":"utxo","progress":"3.0%","current":90000}
{"time":"2026-05-07T18:27:59.918751-05:00","level":"INFO","msg":"copying blocks from immutable DB","blocks_copied":135400,"slot":2851472,"blocks_per_sec":"3628","progress":"2.6%"}
{"time":"2026-05-07T18:28:03.992746-05:00","level":"INFO","msg":"UTxO import progress","component":"mithril","stage":"utxo","progress":"4.7%","current":140000}
{"time":"2026-05-07T18:28:09.942607-05:00","level":"INFO","msg":"copying blocks from immutable DB","blocks_copied":146600,"slot":3079123,"blocks_per_sec":"3098","progress":"2.8%"}
{"time":"2026-05-07T18:28:16.348787-05:00","level":"INFO","msg":"UTxO import progress","component":"mithril","stage":"utxo","progress":"6.3%","current":190000}
{"time":"2026-05-07T18:28:19.950224-05:00","level":"INFO","msg":"copying blocks from immutable DB","blocks_copied":173500,"slot":3625638,"blocks_per_sec":"3026","progress":"3.3%"}
{"time":"2026-05-07T18:28:31.019949-05:00","level":"INFO","msg":"copying blocks from immutable DB","blocks_copied":175050,"slot":3657136,"blocks_per_sec":"2560","progress":"3.3%"}
{"time":"2026-05-07T18:28:39.745966-05:00","level":"INFO","msg":"UTxO import progress","component":"mithril","stage":"utxo","progress":"7.4%","current":220000}
{"time":"2026-05-07T18:28:42.138994-05:00","level":"INFO","msg":"copying blocks from immutable DB","blocks_copied":175400,"slot":3664501,"blocks_per_sec":"2206","progress":"3.3%"}
{"time":"2026-05-07T18:28:52.147611-05:00","level":"INFO","msg":"copying blocks from immutable DB","blocks_copied":177750,"slot":3713312,"blocks_per_sec":"1985","progress":"3.3%"}
{"time":"2026-05-07T18:28:52.489035-05:00","level":"INFO","msg":"UTxO import progress","component":"mithril","stage":"utxo","progress":"7.7%","current":230000}
{"time":"2026-05-07T18:28:58.03603-05:00","level":"INFO","msg":"restored normal PRAGMAs"}
{"time":"2026-05-07T18:28:58.051773-05:00","level":"ERROR","msg":"sync did not complete; re-run 'dingo mithril sync' to resume","component":"mithril"}
badger 2026/05/07 18:29:34 ERROR: error flushing memtable to disk: error while creating table err: while creating table: .dingo/blob/000004.sst err: open .dingo/blob/000004.sst: no space left on device
unable to open: .dingo/blob/000004.sst, retrying
badger 2026/05/07 18:29:40 ERROR: error flushing memtable to disk: error while creating table err: while creating table: .dingo/blob/000005.sst err: open .dingo/blob/000005.sst: no space left on device
unable to open: .dingo/blob/000005.sst, retrying
badger 2026/05/07 18:29:43 ERROR: error flushing memtable to disk: error while creating table err: while creating table: .dingo/blob/000006.sst err: open .dingo/blob/000006.sst: no space left on device
unable to open: .dingo/blob/000006.sst, retrying
badger 2026/05/07 18:29:45 ERROR: error flushing memtable to disk: error while creating table err: while creating table: .dingo/blob/000007.sst err: open .dingo/blob/000007.sst: no space left on device
unable to open: .dingo/blob/000007.sst, retrying
badger 2026/05/07 18:29:47 ERROR: error flushing memtable to disk: error while creating table err: while creating table: .dingo/blob/000008.sst err: open .dingo/blob/000008.sst: no space left on device
unable to open: .dingo/blob/000008.sst, retrying
badger 2026/05/07 18:30:06 ERROR: error flushing memtable to disk: error while creating table err: while creating table: .dingo/blob/000009.sst err: open .dingo/blob/000009.sst: no space left on device
unable to open: .dingo/blob/000009.sst, retrying
badger 2026/05/07 18:30:08 ERROR: error flushing memtable to disk: error while creating table err: while creating table: .dingo/blob/000010.sst err: open .dingo/blob/000010.sst: no space left on device
unable to open: .dingo/blob/000010.sst, retrying
^C^C
^C
```

database/batch.go
```diff
@@ -0,0 +1,245 @@
...
+	updateEpoch uint64,
+	pparamUpdates map[lcommon.Blake2b224]lcommon.ProtocolParameterUpdate,
+	certDeposits map[int]uint64,
+	offsets *BlockIngestionResult,
+	acc *BatchAccumulator,
+	txn *Txn,
+) error {
+	if acc == nil {
+		return errors.New("batch accumulator must not be nil")
+	}
```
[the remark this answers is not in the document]
arepala-uml: The concern is valid in theory but already handled at the plugin layer. Each plugin's SetTransactionBatched implementation has its own nil guard at the top (e.g. sqlite/transaction.go: if acc == nil { return errors.New("SetTransactionBatched: acc must not be nil") }), so a zero-value &BatchAccumulator{} already fails with a clear error rather than a panic.

Adding validateBatchAccumulator would duplicate the backend type-switch that already exists in NewBatchAccumulator, FlushBatch, and setTransactionMetadataBatched, adding ~20 lines to guard against a case that is both already handled and unreachable through the normal API. Keeping the simpler nil check at the boundary is sufficient here.

database/batch.go
```diff
@@ -0,0 +1,245 @@
...
+	acc *BatchAccumulator,
+	txn *Txn,
+) error {
+	if acc == nil {
+		return nil
+	}
+	var metadataTxn types.Txn
+	if txn != nil {
+		metadataTxn = txn.Metadata()
+	}
```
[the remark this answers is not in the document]
arepala-uml: Fixed it.

database/batch.go
```diff
@@ -0,0 +1,252 @@
...
+package database
+
+import (
+	"encoding/hex"
+	"errors"
+	"fmt"
+
+	"github.com/blinklabs-io/dingo/database/plugin/metadata/mysql"
+	"github.com/blinklabs-io/dingo/database/plugin/metadata/postgres"
+	"github.com/blinklabs-io/dingo/database/plugin/metadata/sqlite"
```
wolf31o2: We should never import specific plugins at this level. This is leaking details of the specific plugins into the top-level database package. This needs to go through an interface at the store type level (metadata) with arguments to initialize the correct one. We do this when we initialize tests.

```
        db, err := New(&Config{
                DataDir:        t.TempDir(),
                Logger:         slog.New(slog.NewTextHandler(io.Discard, nil)),
                BlobPlugin:     "badger",
                MetadataPlugin: "sqlite",
        })
```

Loading the plugin should load the correct accumulator.

wolf31o2 (changes requested):

database/batch.go
arepala-uml: yes, I understood the issue and we dont have any defined type at database interface(store) level and I thought there will be lot of modifications & I have written like this in simple way but it is a leakage. It is my mistake and sorry for that.  I will modify it and will add a new interface so that we can remove all these hardcoded ones. 

There are some more commits which I have made fixes in flush logic & caught them during the manual validation using mithril sync. Will push them as well.

database/batch.go
```diff
@@ -0,0 +1,177 @@
...
+	txn *Txn,
+) error {
+	if acc == nil {
+		return errors.New("batch accumulator must not be nil")
+	}
+	owned := false
+	if txn == nil {
+		txn = d.Transaction(true)
+		owned = true
+		defer txn.Rollback() //nolint:errcheck
```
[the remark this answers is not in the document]
arepala-uml: I have fixed the issue in latest commit.

arepala-uml:
During the manual validation using mithril sync - I have seen some issues and fixed it:

1. I have fixed batched pool registration cleanup.Deletes pool_registration_owner / pool_registration_relay by pool_registration_id, not invalid certificate_id.
2. Backfill now opens the batch transaction only when needed and closes it after each flush, so SQLite checkpoint writes are no longer blocked by an open batch transaction. computeBlockNonce also uses the active batch transaction, keeping nonce work consistent with the current flushed metadata batch.

database/batch.go
arepala-uml: I have added metadata store level batching interface now and there is no direct sqlite, postgres, or mysql dependency in top-level database batching code. I have updated sqlite/postgres/mySQL metadata plugins to support the new interface.

arepala-uml:
These are some of the logs from mitrhil sync on preview network with these changes (Added some extra logging for flush_batch for testing purpose).

Screenshot 2026-05-11 at 9 41 49 AM


I just re-ran with my changes (mitrhil sync) and added metadata batch was working as expected and ETA was showing like 79.36 hours. 
Screenshot 2026-05-11 at 10 07 19 AM

internal/node/backfill.go
```diff
@@ -752,16 +831,35 @@ func (b *Backfill) Run(ctx context.Context) error {
...
+		if batchBlockCount >= backfillBatchSize {
+			if err := flushBatch(); err != nil {
+				saveCommittedCheckpoint()
+				return fmt.Errorf(
+					"flushing backfill batch: %w", err,
+				)
+			}
+		}
 
 		if processedBlocks%1000 == 0 {
```
chrisguiney, line 855: **Blocker — latent silent data loss when `backfillBatchSize` does not divide 1000.**

`saveCommittedCheckpoint()` calls `rollbackBatch()`, which rolls back `batchTxn`, resets `acc`, and zeroes `batchBlockCount`. The iterator is **not** rewound, and the loop does not return — it keeps processing.

This only works today because `backfillBatchSize == 100` divides `1000`: the 100-flush always fires first and leaves `batchBlockCount == 0` when this branch runs. Set `backfillBatchSize = 99` (or 150, or 250) and every periodic checkpoint will silently throw away the in-flight partial batch: blob + metadata rows for those blocks get rolled back, the iterator advances past them, and the run continues. The on-disk DB is missing those blocks until the next resume re-processes them — a non-resumed run finishes with permanent gaps.

Two safer options:
- Only checkpoint on a flush boundary: `if processedBlocks%1000 == 0 && batchBlockCount == 0 { b.saveCheckpoint(cp) }`, or
- Force a flush first: `flushBatch()` (error-check it), then `b.saveCheckpoint(cp)` directly — no rollback.

The second is closer to the spirit of the periodic checkpoint (durable progress every 1000 blocks regardless of batch geometry).

internal/node/backfill.go
```diff
@@ -648,27 +652,100 @@ func (b *Backfill) Run(ctx context.Context) error {
...
+		}
+	}()
+
 	var (
 		processedBlocks int
 		processedTxs    int
-		lastLogTime     = time.Now()
-		startTime       = time.Now()
+		// It tracks blocks accumulated since the last flush
+		batchBlockCount int
```
chrisguiney, line 679: Stray pronoun — reads as `It tracks ...` with no antecedent. Match the form of the `committedSlot` comment two lines below: `// batchBlockCount tracks blocks accumulated since the last flush.`

internal/node/backfill.go
```diff
@@ -752,16 +831,35 @@ func (b *Backfill) Run(ctx context.Context) error {
...
+			// batch boundary.
+			saveCommittedCheckpoint()
 		}
 		b.maybeLogProgress(
 			cp, processedBlocks, processedTxs,
 			tipSlot, startSlot, startTime, &lastLogTime,
 		)
 	}
 
+	// Flushes the last batch and may be fewer than 100 blocks.
```
chrisguiney: Dangling clause. Suggest: `// Flush the final partial batch (may be < backfillBatchSize blocks).` Also avoids hard-coding the literal `100` in a comment that lives 800 lines away from the constant.

database/plugin/metadata/sqlite/batch_accumulator.go
```diff
@@ -284,61 +297,82 @@ func batchDeleteByTxIDs(db *gorm.DB, table string, ids []uint) error {
 	return nil
 }
 
-func batchSpendUtxos(db *gorm.DB, spends []utxoSpend) error {
+func (d *MetadataStoreSqlite) batchSpendUtxos(db *gorm.DB, spends []utxoSpend) error {
```
chrisguiney: **Blocker — perf regression that may swamp the gains of batching.**

The old path issued one `UPDATE ... CASE ... WHEN ...` per chunk of up to 100 spends. The new path issues one `UPDATE` per spend, plus a `SELECT` (`checkUnmatchedUtxoSpend`) for each row where `RowsAffected == 0`. With `batchChunkRows` bumped to 500, a single flush on a busy chain can be tens of thousands of individual UPDATEs.

The correctness motivation is right — the old CASE-update silently swallowed conflicts — but this directly fights the throughput goal of the PR.

Two things I'd like to see:

1. A backfill benchmark on sqlite before/after this PR on a known era. "Is batched backfill actually faster end-to-end on the default backend?" is the gating question.
2. If perf is bad, restructure: do the bulk CASE-UPDATE first (fast path), then issue a single SELECT for the rows where the bulk reported `RowsAffected < len(chunk)` to disambiguate conflict vs. missing. That preserves the conflict-detection fix without the N×UPDATE cost.

database/plugin/metadata/sqlite/batch_accumulator.go
```diff
@@ -25,7 +27,9 @@ import (
 )
 
 const (
-	batchChunkRows = 100
+	// Keep generic bulk inserts well below SQLite's variable limit while
+	// avoiding excessive statement counts for address/witness rows.
+	batchChunkRows = 500
```
chrisguiney: Worth noting in the comment that the bump from 100→500 helps the bulk *insert* chunks (addresses/witnesses), but the new per-row spend update does **not** scale with this constant — it just chunks for transaction-size purposes. As written the comment suggests both are bound by the same throughput tradeoff.

database/batch.go
```diff
@@ -0,0 +1,184 @@
...
+	owned := false
+	accDirty := false
+	if txn == nil {
+		txn = d.Transaction(true)
+		owned = true
+		defer func() {
+			if retErr != nil && accDirty {
+				acc.Reset()
+			}
+			_ = txn.Rollback()
```
chrisguiney, line 83: Two small things:

1. `accDirty` is only ever assigned `owned` and only checked alongside `owned` (line 152, line 78). Drop it — the `owned` flag is the same signal.
2. The deferred `_ = txn.Rollback()` runs after a successful `Commit()` below (line ~180), where it returns `ErrTxDone` (or similar) that gets silently swallowed. Cleaner: set `txn = nil` after `txn.Commit()` succeeds and guard the deferred rollback on `txn != nil`.

database/batch.go
```diff
@@ -0,0 +1,184 @@
...
+			return fmt.Errorf(
+				"set utxo offset %x#%d: %w",
+				txID[:8],
+				outputIdx,
+				err,
+			)
+		}
+	}
+
+	if err := d.ensureTransactionConsumedUtxos(tx, point, txn); err != nil {
```
chrisguiney, line 147: **Major — cross-block UTxO spends inside a batch fall into `recoverConsumedUtxo`.**

`ensureTransactionConsumedUtxos` runs *before* the plugin's `SetTransactionBatched`, which only `MergeFrom(local)`s produced-UTxO metadata rows into the shared `acc`. So when a tx in block N+1 (same batch) spends an output produced in block N, `GetUtxoIncludingSpent` returns nil and we fall through to `recoverConsumedUtxo`.

Blob `SetUtxo` writes happen in-txn so recovery *should* succeed, but it's per-input recovery work that the non-batched path never paid (the prior block was already committed). Two concerns:

1. Add an integration test covering the exact case: two consecutive blocks in the same batch, where N+1 consumes an output produced by N. Confirm that the batched path persists the spend correctly through recovery.
2. Worth measuring — if cross-block spends are common during backfill on a real chain, this recovery cost could meaningfully erode the batching win.

chrisguiney (changes requested):
## Executive summary

This PR wires the existing `BatchAccumulator` into `Backfill.Run`, replacing per-block commits with a shared accumulator + transaction that flushes every 100 blocks. It also adds a `MetadataBatchAccumulator` interface, exposes `NewBatchAccumulator`/`FlushBatch`/`SetTransactionBatched` on `Database`, updates the sqlite/postgres/mysql metadata stores to accept the new interface, and folds in two correctness fixes (pool-registration owner/relay deletion joins on `pool_registration.id`, and SQLite UTxO-spend conflicts now surface as `types.ErrUtxoConflict`).

The direction is right and the interface design is fine, but I'm requesting changes before merge for the reasons below.

## Blockers

1. **Latent silent data loss at the `processedBlocks%1000` checkpoint** — see inline comment on `internal/node/backfill.go`. Only safe today because 1000 % 100 == 0; any future tweak to either constant drops the in-flight partial batch and the iterator keeps going.

2. **SQLite spend path went from O(1) per chunk to O(N)** — see inline on `batch_accumulator.go:300`. The conflict-detection fix is good, but going from one CASE-update per 100 spends to one UPDATE per spend (plus a SELECT on misses), with `batchChunkRows` bumped to 500, can easily cost more than the per-block commits this PR is trying to eliminate. We need a benchmark on a real sqlite backfill before/after, or a bulk-update-then-SELECT-misses formulation.

3. **No tests for the new backfill batching logic.** The two added tests exercise `FlushBatch` directly with pre-seeded UTxOs. There is no coverage for `Backfill.Run` with the accumulator (multi-batch happy path), cancellation mid-batch, error rollback + resume from `committedSlot`, `Database.SetTransactionBatched` in the owned-txn path, or a cross-block UTxO spend through the batched path. Given the PR is rewiring the entire backfill persistence path, the test gap alone is reason to hold.

## Major

**Cross-block UTxO spends inside a batch route through `recoverConsumedUtxo`.** `Database.SetTransactionBatched` calls `ensureTransactionConsumedUtxos` before delegating to the plugin, but the plugin only `MergeFrom(local)`s produced-UTxO metadata rows into the shared `acc` — the row is not in the metadata DB until `FlushBatch`. When a tx in block N+1 spends an output produced in block N (same batch), `GetUtxoIncludingSpent` won't find it and falls through to `recoverConsumedUtxo`. Blob `SetUtxo` writes do happen in-txn so recovery should succeed, but it's per-input recovery work that the non-batched path never paid (the prior block was already committed and the row was present). Please add an integration test covering this exact case before merge.

## Minor

- `database/batch.go`: `accDirty` is functionally redundant with `owned`; the deferred `_ = txn.Rollback()` also runs after a successful `Commit()` and silently drops the resulting error. Set `txn = nil` after `Commit()` and check in the defer.
- `internal/node/backfill.go:671`: `// It tracks blocks accumulated since the last flush` — stray pronoun; match the form of the `committedSlot` comment below it.
- `internal/node/backfill.go:858`: `// Flushes the last batch and may be fewer than 100 blocks.` — dangling clause; suggest `// Flush the final partial batch (may be < backfillBatchSize blocks).`
- The plugin type-assertion error strings are inconsistent across stores (`"mysql FlushBatch: wrong accumulator type %T"` vs `"SetTransactionBatched: wrong accumulator type %T"`). Pick one format.
- The pool-registration owner/relay deletion fix is applied in each store's `SetTransactionBatched`. Confirm the same fix is (or doesn't need to be) applied to the non-batched `SetTransaction` paths — the FK semantics are identical.

internal/node/backfill.go
arepala-uml: I have fixed and removed hard coded literal from the comment. Please review it.

internal/node/backfill.go
arepala-uml: I understood the concern here and I have changed the periodic checkpoint path to force a flush first. Now when processedBlocks%1000 == 0, we call flushBatch() and only save the checkpoint after the current partial batch is committed. That removes the dependency on backfillBatchSize dividing 1000 and avoids rollbackBatch() discarding in-flight work during normal periodic checkpointing.

internal/node/backfill.go
arepala-uml: Modified the comment message and please review it.

database/batch.go
arepala-uml: I added a test case for this scenario, and it confirms the batched path is not missing the spend.
The test processes a producer tx and a consumer tx in the same unflushed batch, where the consumer spends a UTxO created by the producer in the previous block. After flushBatch, commit, the utxO is present in metadata and correctly marked spent by the consumer tx.

database/batch.go
arepala-uml: I have removed this accDirty way and added a cleaner fix by checking with tx in defer block. Please review it.

database/plugin/metadata/sqlite/batch_accumulator.go
arepala-uml: Agreed. I changed this back to the bulk CASE-update fast path, so we are no longer issuing one UPDATE per spend. The spend flush now does one UPDATE ... CASE ... WHEN per chunk, and only if RowsAffected < len(chunk) we run the slower disambiguation path. That fallback uses a single SELECT for the chunk to distinguish missing inputs, idempotent same-tx spends, and real conflicts.

database/plugin/metadata/sqlite/batch_accumulator.go
arepala-uml: I have reverted batchChunkRows back to 100, so this PR no longer changes the chunk-size tuning. I also removed the expanded comment and left it as the original SQLite variable-limit guard. Please review it.

chrisguiney (approved):

wolf31o2 (approved):

PR #2210, opened 2026-05-08
feat: validate block producer credentials at startup (#1853)

### LINKED ISSUE #1853: node: validate block producer credentials at startup
## Summary

When Dingo starts in block producer mode, it does not validate that the VRF key, KES key, and operational certificate are present, well-formed, and consistent until the first leader check or forge attempt. A misconfigured block producer can run for an entire epoch before failing visibly. This was identified as RSA-11 in the runtime stability audit.

## Current State

Block producer credentials (VRF signing key, KES signing key, operational certificate) are loaded lazily. If files are missing, corrupt, or mismatched (e.g., opcert pool ID doesn't match the registered pool), the error surfaces deep in the forging path rather than at startup.

## What Needs to Happen

1. On startup in block producer mode, validate:
   - VRF key file exists and decodes
   - KES key file exists and decodes
   - Operational certificate file exists, decodes, and the KES period is plausible
   - OpCert counter is not stale (if ledger state is available)
2. Refuse to start if any credential is invalid — fail fast with a clear error message.
3. Log the pool ID, KES period, and opcert counter at startup for operator visibility.

## Roadmap

Q2 2026 — runtime stability audit fix, pre-testnet gate.

chrisguiney:
closes #1853 

---
## Summary by cubic
Adds startup validation for block producer credentials to catch misconfigured, future, or expired keys early. The node checks KES period using Shelley genesis and cross-checks credentials with the ledger, failing fast on unsafe setups.

- **New Features**
  - Validate opcert KES period at startup using Shelley genesis (supports fractional slot lengths); future or expired periods stop the node.
  - Require Cardano node config with Shelley genesis in block producer mode.
  - Cross-check loaded VRF key hash with on-chain pool registration; mismatch is fatal, unregistered logs a warning and continues; seed-only VRF keys skip the VRF check.
  - Check opcert IssueNumber against the ledger’s latest when available; stale counters are fatal. Added minimal ledger queries and an adapter to wire this in.

<sup>Written for commit 759b136722dcf28ea2aef6bf75688222790316a9. Summary will update on new commits.</sup>

### DIFF
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -939,6 +939,46 @@ func (ls *LedgerState) Chain() *chain.Chain {
 	return ls.chain
 }
 
+// PoolRegistrationVRFKeyHash returns the VRF key hash recorded on the
+// most recent active pool registration certificate for the given pool.
+// found is false when the pool has no on-chain registration yet — that
+// is informational, not an error condition, since operators commonly
+// stage credentials before the registration certificate is on chain.
+//
+// Used by the block-producer credential check at startup to confirm the
+// loaded VRF key matches what the chain has on file.
+func (ls *LedgerState) PoolRegistrationVRFKeyHash(
+	poolID [28]byte,
+) (vrfHash [32]byte, found bool, err error) {
+	pkh := lcommon.PoolKeyHash(lcommon.NewBlake2b224(poolID[:]))
+	pool, err := ls.db.GetPool(pkh, false, nil)
+	if err != nil {
+		if errors.Is(err, models.ErrPoolNotFound) {
+			return [32]byte{}, false, nil
+		}
+		return [32]byte{}, false, err
+	}
+	if pool == nil || len(pool.VrfKeyHash) != 32 {
+		return [32]byte{}, false, nil
+	}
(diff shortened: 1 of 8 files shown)


ledger/state.go
```diff
@@ -939,6 +939,46 @@ func (ls *LedgerState) Chain() *chain.Chain {
...
+	return vrfHash, true, nil
+}
+
+// LatestOpCertSequence returns the highest operational-certificate
+// IssueNumber observed on chain for the given pool. found is always
+// false today: the chain stores opcerts in block headers but Dingo does
+// not yet aggregate the highest sequence per pool. This method exists
+// so the credential cross-check at startup can call into it; once
+// counter tracking lands the implementation will fill in without any
+// caller change.
```
wolf31o2, line 974: Is there an issue or plan for this?

wolf31o2 (approved):

ledger/state.go
chrisguiney: there is now: https://github.com/blinklabs-io/dingo/issues/2211

PR #2232, opened 2026-05-11
fix(chain): add multi-chain reconcile tests (#752)

### LINKED ISSUE #752: Add more multi-chain tests
We need additional tests for the new multi-chain tracking support, mostly around scenarios where the primary chain has a rollback beyond the fork point of a non-primary chain. This is currently limited by our test series of blocks due to them not having a consistent hash after a round-trip to/from CBOR.

What we need is:

- [ ] method for generating test chain of blocks
  - [ ] take starting block number, prev hash, slot, and count
  - [ ] generate dummy headers and encode to CBOR
  - [ ] use generated hash of previous block as prev-hash in next block
- [ ] additional chain tests
  - [ ] iterate non-primary chain after primary is rolled back beyond fork point
  - [ ] rollback non-primary chain after primary is rolled back, and then iterate to verify
  - [ ] more than one non-primary chain

chrisguiney:
## Summary

- Adds `generateTestChain`, a CBOR-stable Conway-block generator so reconcile-path tests can drive the real `ledger.NewBlockFromCbor` decode path (the existing `MockBlock` fixture has no CBOR, so blocks could not survive the round-trip done by `models.Block.Decode` in `chain.reconcile`).
- Adds three multi-chain tests from #752's checklist:
  - Iterate a non-primary chain after the primary chain has rolled back past the fork point.
  - Roll back the non-primary chain after the primary rollback, then iterate (covers both the rollback signal delivered to a live iterator and a fresh re-iterate).
  - Two non-primary chains forked at different points reconcile independently and a rollback of one does not affect the other.
- Fixes a `chain.reconcile` bug surfaced by the first scenario: when the primary chain has rolled back past a fork point, walking backward through the fork's PrevHash chain encountered `ErrBlockNotFound` from `primaryChain.blockByIndex(tmpBlock.ID)` and aborted. Reconcile now treats that miss as "not common yet, keep walking back".
- Extends `manager.blockByHash` to fall through to the DB (matching the pattern in `blockByPoint`); the cache-only behavior could not find still-on-primary ancestors that were never rolled back. Also drops a redundant manager-level RLock — the block cache has its own lock.

## Coverage delta (chain package)

| | Before | After |
|---|---|---|
| `chain.reconcile` | 6.8% | 72.9% |
| `manager.blockByHash` | 0% | 60% |
| `manager.chainNeedsReconcile` | 30% | 100% |
| total | 54.9% | 61.2% |

## Test plan

- [ ] `go test -race ./chain/...` passes locally and in CI
- [ ] `golangci-lint run ./chain/...` clean
- [ ] Downstream callers (`ledger`, `chainsync`, `ouroboros`) still pass

closes #752

---
## Summary by cubic
Adds a CBOR-stable Conway block generator and four multi-chain reconcile tests from #752. Fixes `chain.reconcile` to walk past rolled-back ancestors and preserve an empty fork’s orphaned tip; extends `manager.blockByHash` to query the DB, improving correctness and coverage.

- **New Features**
  - Added `generateTestChain` to produce CBOR-stable blocks so tests exercise the real `ledger.NewBlockFromCbor` path.
  - Added multi-chain reconcile tests (from #752): iterate a non-primary after primary rollback past fork; rollback the non-primary after primary rollback (checks iterator signal and re-iterate); verify two non-primary forks reconcile independently; and handle an empty fork by preserving its orphaned tip after a primary rollback.

- **Bug Fixes**
  - `chain.reconcile`: when the primary rolled back past the fork, treat `ErrBlockNotFound` on `blockByIndex` as “not common yet” and keep walking via `PrevHash`; when the fork has no in-memory blocks, preserve its tip while re-anchoring against the primary.
  - `manager.blockByHash`: fall through to the DB (and drop a redundant manager-level RLock) so ancestors still on the primary are found.

<sup>Written for commit bf414749bd9c9645ed371da51899cb700fd7f237. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chain reconciliation to gracefully handle missing blocks when comparing forks against the primary chain, continuing the backward walk instead of aborting on block-not-found errors.

* **Tests**
  * Added comprehensive test coverage for non-primary chain reconciliation and rollback behavior following primary chain rollbacks past fork points.

* **Refactoring**
  * Optimized block lookup performance by removing unnecessary synchronization around cache access, with clarifying documentation on cache contents.

[Review Change Stack](https://app.coderabbit.ai/change-stack/blinklabs-io/dingo/pull/2232)

### DIFF
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -1257,13 +1257,17 @@ func (c *Chain) reconcile() error {
 		if err != nil {
 			return err
 		}
-		// Lookup same block index on primary chain
+		// Lookup same block index on primary chain. When the primary
+		// has rolled back past tmpBlock's old index the lookup misses;
+		// treat tmpBlock as non-common and keep walking back via its
+		// PrevHash rather than aborting reconcile.
 		primaryBlock, err := primaryChain.blockByIndex(tmpBlock.ID)
-		if err != nil {
+		if err != nil && !errors.Is(err, models.ErrBlockNotFound) {
 			return err
 		}
 		// Update last common block index and return when we find a matching block on the primary chain
-		if tmpBlock.Slot == primaryBlock.Slot &&
+		if err == nil &&
+			tmpBlock.Slot == primaryBlock.Slot &&
 			bytes.Equal(tmpBlock.Hash, primaryBlock.Hash) {
 			c.lastCommonBlockIndex = tmpBlock.ID
 			break
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -1232,8 +1232,18 @@ func (c *Chain) reconcile() error {
 	}
 	// Determine prev-hash from earliest known good block
 	knownPoint := c.currentTip.Point
+	// Iterate backward through chain based on prev-hash until we find a matching block on the primary chain
+	// Accumulate blocks locally to avoid O(K²) prepending
+	newBlocks := make([]ocommon.Point, 0, securityParam)
 	if len(c.blocks) > 0 {
 		knownPoint = c.blocks[0]
+	} else {
+		// No in-memory blocks: the chain's current tip is itself the
+		// earliest known good block. Seed newBlocks so the tip is
+		// preserved after we re-anchor lastCommonBlockIndex against
+		// the primary; otherwise iteration past the new common point
+		// would silently truncate at it.
+		newBlocks = append(newBlocks, knownPoint)
 	}
 	knownBlock, err := c.manager.blockByPoint(knownPoint, nil)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


chrisguiney:
Re @coderabbitai's nit on the in-tree `generateTestChain` helper conflicting with the `ouroboros-mock` policy in CLAUDE.md:13 / AGENTS.md:26 — tracked separately in #2233 as a follow-up. This PR was already wider than the original #752 ask once the rollback tests surfaced two reconcile bugs; the cross-repo fixture move belongs in its own paired-PR sequence (ouroboros-mock additions → dingo dep bump → helper swap).

Item 1 from the same review (orphaned-tip preservation when `len(c.blocks) == 0` during reconcile) is fixed in this PR — see `ccdb00d3` (fix) and `bf414749` (regression test).

wolf31o2 (approved):
Approved but change this to a "fix(chain):" since you fixed a bug.

PR #2245, opened 2026-05-12
feat(config): default 25 MiB mempool capacity in Leios mode (#2081)

### LINKED ISSUE #2081: leios: raise mempool capacity to 25MB when running in Leios mode
## Summary

When the node is running in Leios mode, raise the mempool capacity to 25MB. The current default is 1MB (`MempoolCapacity: 1048576` in `internal/config/config.go`), which is sized for Praos block throughput and is far too small for Leios, where input blocks pull from the mempool at a much higher rate and several pipelines are in flight at once.

## Why

Leios sustains substantially higher tx throughput than Praos by fanning blocks out into multiple input/endorser block pipelines. A 1MB mempool drains in well under a single Leios pipeline stage at target rates, which artificially caps what we can demonstrate on the Leios testnet and makes any throughput numbers we publish a measurement of mempool size rather than of the protocol.

25MB is the working number we are aligning on for the testnet milestone. It gives enough headroom for several pipelines worth of pending txs without putting unreasonable memory pressure on a relay.

## What needs to happen

1. When the node detects it is running with Leios enabled, default `MempoolCapacity` to 25MB (26214400 bytes) instead of 1MB.
2. An explicit user setting (CLI flag, env, or YAML) still wins, per the existing config priority. Only the default changes based on mode.
3. Update the eviction and rejection watermark behavior review so the larger capacity does not change the percentage thresholds in a surprising way.
4. Add or extend a test in `internal/config/` covering the Leios-mode default.
5. Document the new default in `dingo.yaml.example` next to `mempoolCapacity`.

## Out of scope

- Dynamic resizing of the mempool at runtime.
- Changing the Praos default. The 1MB Praos default stays.
- Per-pipeline mempool partitioning. That is a separate Leios task if it is needed at all.

## Milestone

Testnet Block Production and Leios Prototype.

chrisguiney:
## Summary

Default `MempoolCapacity` to 25 MiB (26,214,400 bytes) when `runMode: "leios"` is set; Praos modes keep the existing 1 MiB default. An explicit CLI/env/YAML value still wins, per existing config priority.

Leios fans blocks out into multiple input/endorser-block pipelines. A 1 MiB mempool drains in well under one pipeline stage at target rates and artificially caps Leios testnet throughput — any numbers we publish would measure the mempool, not the protocol.

## Implementation

- `internal/config/config.go`
  - New constants `DefaultMempoolCapacityPraos = 1048576` and `DefaultMempoolCapacityLeios = 26214400`.
  - Package-level `globalConfig.MempoolCapacity` becomes the zero sentinel (matching the existing watermark default pattern).
  - After `RunMode` validation, fill in `MempoolCapacity` from `RunMode` when still zero.
- `dingo.yaml.example` — comment updated to document both defaults.
- `internal/config/config_test.go`
  - `resetGlobalConfig` mirror updated to the new sentinel.
  - New `TestLoad_MempoolCapacityMode` with four subtests:
    - Praos serve default → 1 MiB
    - Leios default → 25 MiB
    - Explicit `mempoolCapacity` wins in Leios mode
    - Explicit `mempoolCapacity` wins in serve mode

## Watermark note

`EvictionWatermark` (0.90) and `RejectionWatermark` (0.95) are read live as fractions of capacity (`mempool.go:717,735`). At 25 MiB this puts eviction at ~22.5 MiB and rejection at ~23.75 MiB — the percentage thresholds are unchanged, the absolute byte budgets scale proportionally. No watermark code touched.

## Edge case worth flagging

A user with an explicit `mempoolCapacity: 0` setting (env, YAML, or CLI) today gets a warning from the mempool (`zero or negative; all transactions will be rejected`). After this change, `0` is interpreted as "use mode default" and they silently get 1 MiB (or 25 MiB in Leios). The prior behavior was a misconfiguration the user almost certainly didn't want, so this is an improvement, but worth a release-note callout.

## Test plan

- [ ] `go test -race ./internal/config/... ./mempool/...` passes (1.1s + 4.3s locally)
- [ ] `go build ./...` clean
- [ ] `golangci-lint run ./internal/config/... ./mempool/...` 0 issues
- [ ] Manual: `--mempool-capacity 12345 --run-mode leios` still yields 12345 (covered by test 3 of the new subtests)

## Out of scope (per the issue)

- Dynamic resizing of the mempool at runtime
- Changing the Praos default
- Per-pipeline mempool partitioning

closes #2081

---
## Summary by cubic
Set the default mempool capacity to 25 MiB when `runMode: "leios"`; Praos stays at 1 MiB. Explicit CLI/env/YAML values still override.

- **New Features**
  - Mode-derived defaults apply only when `MempoolCapacity` is unset/0.
  - `dingo.yaml.example`: `mempoolCapacity` is now commented and documents both defaults.

- **Migration**
  - `mempoolCapacity: 0` now means "use mode default" instead of rejecting all transactions.

<sup>Written for commit 34ffe8ffd2a902d596ca31caf1a434856feb0b81. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Memory pool capacity now defaults based on operational mode: 1 MiB for Praos mode and 25 MiB for Leios mode. Manual configuration is no longer required unless custom values are needed.

* **Tests**
  * Added verification tests for mode-dependent memory pool defaults.

[Review Change Stack](https://app.coderabbit.ai/change-stack/blinklabs-io/dingo/pull/2245)

### DIFF
--- a/dingo.yaml.example
+++ b/dingo.yaml.example
@@ -232,12 +232,8 @@ intersectTip: false
 
 # Maximum total size (in bytes) of all transactions allowed in the mempool.
 # Transactions exceeding this limit will be rejected.
-# Default: 1048576 (1 MiB) for Praos modes (runMode "serve" / "load" / "dev")
-# Default: 26214400 (25 MiB) when runMode is "leios", since Leios sustains
-# much higher tx throughput across several pipelines and a 1 MiB cap would
-# bottleneck the protocol itself.
-# Leave commented to let the mode-derived default apply; uncomment and
-# edit to override.
+# Default: 1048576 (1 MiB) for Praos modes; 26214400 (25 MiB) for Leios.
+# Leave commented to use the mode default; uncomment to override.
 # CLI: --mempool-capacity
 # mempoolCapacity: 1048576
 
--- a/dingo.yaml.example
+++ b/dingo.yaml.example
@@ -235,9 +235,11 @@ intersectTip: false
 # Default: 1048576 (1 MiB) for Praos modes (runMode "serve" / "load" / "dev")
 # Default: 26214400 (25 MiB) when runMode is "leios", since Leios sustains
 # much higher tx throughput across several pipelines and a 1 MiB cap would
-# bottleneck the protocol itself. An explicit value here overrides both.
+# bottleneck the protocol itself.
+# Leave commented to let the mode-derived default apply; uncomment and
+# edit to override.
 # CLI: --mempool-capacity
-mempoolCapacity: 1048576
+# mempoolCapacity: 1048576
 
 # Forging tolerances (0 = defaults)
 # forgeSyncToleranceSlots controls how far the local chain can lag the upstream
--- a/dingo.yaml.example
+++ b/dingo.yaml.example
@@ -232,7 +232,10 @@ intersectTip: false
 
 # Maximum total size (in bytes) of all transactions allowed in the mempool.
 # Transactions exceeding this limit will be rejected.
-# Default: 1048576 (1 MB)
+# Default: 1048576 (1 MiB) for Praos modes (runMode "serve" / "load" / "dev")
+# Default: 26214400 (25 MiB) when runMode is "leios", since Leios sustains
+# much higher tx throughput across several pipelines and a 1 MiB cap would
+# bottleneck the protocol itself. An explicit value here overrides both.
 # CLI: --mempool-capacity
 mempoolCapacity: 1048576
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -81,6 +81,13 @@ const (
 	DefaultRejectionWatermark          = 0.95
 	DefaultForgeSyncToleranceSlots     = 100
 	DefaultForgeStaleGapThresholdSlots = 1000
+	// MempoolCapacity defaults are mode-dependent. Praos block
+	// throughput is well served by 1 MiB; Leios runs several input/
+	// endorser block pipelines concurrently and needs substantially
+	// more headroom so the protocol — not the mempool — is what's
+	// being measured.
+	DefaultMempoolCapacityPraos = 1048576  // 1 MiB
+	DefaultMempoolCapacityLeios = 26214400 // 25 MiB
 )
 
 // ErrPluginListRequested is returned when the user requests to list
@@ -390,7 +397,10 @@ func (c *Config) ParseCmdlineArgs(programName string, args []string) error {
 }
 
 var globalConfig = &Config{
-	MempoolCapacity:      1048576,
+	// MempoolCapacity is left as the zero sentinel; LoadConfig fills
+	// it in based on RunMode (Praos vs Leios) after CLI/env/YAML have
+	// been merged.
+	MempoolCapacity:      0,
 	EvictionWatermark:    DefaultEvictionWatermark,
 	RejectionWatermark:   DefaultRejectionWatermark,
 	BindAddr:             "0.0.0.0",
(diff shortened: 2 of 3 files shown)


internal/config/config.go
```diff
@@ -81,6 +81,13 @@ const (
 	DefaultRejectionWatermark          = 0.95
 	DefaultForgeSyncToleranceSlots     = 100
 	DefaultForgeStaleGapThresholdSlots = 1000
+	// MempoolCapacity defaults are mode-dependent. Praos block
+	// throughput is well served by 1 MiB; Leios runs several input/
+	// endorser block pipelines concurrently and needs substantially
```
wolf31o2: This is not accurate. We're building Linear Leios (CIP-0164), not the research's fully parallel Leios design with input endorsers. The reference implementation uses a 2 * max block size calculation, so Praos is only 178176 bytes and Leios is 26214400 there.

dingo.yaml.example
```diff
@@ -232,9 +232,14 @@ intersectTip: false
 
 # Maximum total size (in bytes) of all transactions allowed in the mempool.
 # Transactions exceeding this limit will be rejected.
-# Default: 1048576 (1 MB)
+# Default: 1048576 (1 MiB) for Praos modes (runMode "serve" / "load" / "dev")
+# Default: 26214400 (25 MiB) when runMode is "leios", since Leios sustains
+# much higher tx throughput across several pipelines and a 1 MiB cap would
```
wolf31o2: This also isn't accurate.

wolf31o2 (changes requested):
Comments are about a version of Leios we're not implementing. Less is more here. There's no reason for so much exposition, especially when the variable names are pretty self-explanatory.

dingo.yaml.example
chrisguiney: fixed

internal/config/config.go
chrisguiney: fixed

internal/config/config.go
```diff
@@ -390,7 +392,10 @@ func (c *Config) ParseCmdlineArgs(programName string, args []string) error {
 }
 
 var globalConfig = &Config{
-	MempoolCapacity:      1048576,
+	// MempoolCapacity is left as the zero sentinel; LoadConfig fills
+	// it in based on RunMode (Praos vs Leios) after CLI/env/YAML have
+	// been merged.
+	MempoolCapacity:      0,
 	EvictionWatermark:    DefaultEvictionWatermark,
```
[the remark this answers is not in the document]
chrisguiney: Acknowledged but declining the fix.

The observation is correct: `MempoolCapacity == 0` is treated as "unset" and overwritten by the mode default. This is conscious — see the "Edge case worth flagging" section of the PR description.

Reasons against the proposed pointer/`*Set`-flag refactor:

1. **Prior behavior at `0` was a misconfiguration.** Mempool logs `WARN: zero or negative; all transactions will be rejected` and drops 100% of submissions. Silently upgrading that to a working default is an improvement, not a regression to preserve.
2. **Matches existing patterns in this file.** `NetworkMagic: 0` at `config.go:408` is resolved from `Network` the same way. No field in `Config` uses `*T` or a companion `Set bool` — introducing that pattern for one field would be inconsistent.
3. **No coherent semantics for explicit `0`.** A byte-budget field at zero has no valid use case; the suggested fix preserves the user's ability to opt into a broken state.
4. **Cost is high.** Pointer/sentinel plumbing ripples through CLI flag parsing, env-var binding, YAML unmarshal, and every read site, for zero user benefit.

Will add a release-notes line calling out the semantic change for explicit `mempoolCapacity: 0`.

wolf31o2 (approved):

PR #2252, opened 2026-05-13
fix: match cardano-node Praos chain selection

wolf31o2:
---
## Summary by cubic
Align Praos chain selection with cardano-node. Equal-length forks now use opcert issue number and VRF (Conway restricts VRF to 5-slot distance), removing slot/density/hash fallbacks and preventing nonce drift and stale fork preference.

- **Bug Fixes**
  - Added `ComparePraosTips` with reference rules: higher block count first; at equal length: same issuer+slot prefers higher opcert issue; otherwise compare VRF when the era’s flavor is armed (Conway: max 5-slot distance).
  - Removed slot/density/lower-hash tie-breakers; when the reference tiebreaker isn’t armed, equal-length chains compare equal and the incumbent is kept.
  - Propagated `PraosTiebreakerView` from `chainsync` headers through `ledger` to `chainselection` so earlier headers aren’t dropped as stale when VRF can win; fork resolution now matches the reference select-view.
  - Selector switches to any longer chain (including 1-block leads); for the same selected block, peer choice falls back to connection priority and `BlockfetchLatency` only.
  - VRF cross-checks: treat a zero VRF hash as “unknown”; VRF key hash mismatch now warns on `devnet` instead of aborting; configurator fixes permissions (configs world-readable; pool key dirs 0700/0600, pool 1 chowned for the `dingo` container).
  - Database/ledger: fixed pool registration preload ordering to pick the latest registration for VRF/opcert; `ledger` reads the VRF hash from the latest registration row when present.

- **Migration**
  - Replace removed `CompareChains*`/`IsBetterChain*` helpers with `ComparePraosTips`.
  - Provide a `PraosTiebreakerView` via `GetPraosTiebreakerView(header)` or `PraosTiebreakerViewFromTip(tip, vrf, PraosTiebreakerConfig*)` (use `PraosTiebreakerConfigConway()` in Conway; otherwise `PraosTiebreakerConfigBeforeConway()` or unknown when the era isn’t known).
  - `ChainSelectorConfig.MinSwitchBlockDiff` is removed.

<sup>Written for commit c03d8a7377aa14384a5d955b709257789cdbcf4e. Summary will update on new commits.</sup>

## Release Notes

* **New Features**
  * Enhanced chain selection algorithm using Ouroboros Praos rules with improved tie-breaking logic (longer chain preference, VRF-based precedence).
  * Improved block synchronization validation with more robust chain advancement detection.

* **Bug Fixes**
  * Enhanced VRF key hash validation with better error handling for mismatched credentials.
  * More reliable chain fork resolution based on updated consensus rules.

* **Tests**
  * Updated test suite to verify new chain selection behavior and improved block synchronization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/blinklabs-io/dingo/pull/2252)

### DIFF
--- a/chainselection/vrf.go
+++ b/chainselection/vrf.go
@@ -24,10 +24,133 @@ import (
 	"github.com/blinklabs-io/gouroboros/ledger/conway"
 	"github.com/blinklabs-io/gouroboros/ledger/mary"
 	"github.com/blinklabs-io/gouroboros/ledger/shelley"
+	ochainsync "github.com/blinklabs-io/gouroboros/protocol/chainsync"
 )
 
 // VRFOutputSize is the expected size of a VRF output in bytes.
 const VRFOutputSize = 64
+
+const praosRestrictedTiebreakerMaxSlotDistance = 5
+
+// PraosTiebreakerFlavor matches ouroboros-consensus' VRFTiebreakerFlavor.
+type PraosTiebreakerFlavor uint8
+
+const (
+	PraosTiebreakerUnknown PraosTiebreakerFlavor = iota
+	PraosTiebreakerUnrestricted
+	PraosTiebreakerRestricted
+)
+
+// PraosTiebreakerConfig is the chain-order config projected from the current
+// Shelley-family block config. The reference implementation uses unrestricted
+// VRF before Conway and restricted VRF with max distance 5 from Conway onward.
+type PraosTiebreakerConfig struct {
+	Flavor          PraosTiebreakerFlavor
+	MaxSlotDistance uint64
+}
+
+// PraosTiebreakerView mirrors ouroboros-consensus' PraosTiebreakerView for
+// equal-length chain selection.
+type PraosTiebreakerView struct {
+	Slot              uint64
+	Issuer            []byte
+	IssueNo           uint64
+	TieBreakVRF       []byte
+	TiebreakerConfig  PraosTiebreakerConfig
+	hasIssuerAndIssue bool
+}
+
+func PraosTiebreakerConfigBeforeConway() PraosTiebreakerConfig {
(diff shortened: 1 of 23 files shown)


chainselection/vrf.go
```diff
@@ -24,11 +24,134 @@ import (
...
+		Slot:              header.SlotNumber(),
+		Issuer:            cloneBytes(issuer[:]),
+		IssueNo:           issueNo,
+		TieBreakVRF:       cloneBytes(GetVRFOutput(header)),
+		TiebreakerConfig:  praosTiebreakerConfigForHeader(header),
+		hasIssuerAndIssue: true,
+	}, true
+}
+
+func headerIssueNo(header ledger.BlockHeader) (uint64, bool) {
```
arepala-uml, line 127: Just like a suggestion - I have already written this helper function to extract the operational certification sequence number in ledger/state and I think it does the same where we can use that. This does not effect anything and I am just telling.

https://github.com/blinklabs-io/dingo/blob/7f11e67b17205cc5cbd09d7a3c7dd37d63aa4c52/ledger/state.go#L3493

arepala-uml:
@wolf31o2 I have gone through the proas chain selection fix at tiebreaks and I have a question like how we identified this bug? Is it with rollback events alone be enough to identify the praos tiebreaker bug or do we actually need a side-by-side sync with cardano-node & dingo to catch it?

arepala-uml (approved):

wolf31o2:
I ran this in a devnet to test.

chainselection/vrf.go
wolf31o2: They're in different packages. We'd need to align on how this works, and one is on only the headers, while the other operates on a block.

PR #2308, opened 2026-05-14
fix(ledgerstate): fix pstate map for pool states

wolf31o2:
## Summary by cubic
Fix PState pool parsing to reliably select the correct pool-params map and attach deposits across legacy and UTxO‑HD snapshot formats. Prevents missing or wrong pool data when maps are reordered or values omit the operator.

- **Bug Fixes**
  - Decode all PState maps and pick the one with the most valid pool entries; apply the same logic to Conway.
  - Merge deposits from separate maps when found, and read deposits from the params value when present.
  - Support pool params that omit the operator (operator is the map key), alternate reward account encodings, and unwrapped metadata arrays.
  - Validate 28-byte pool key hashes; skip invalid entries. Parse owners via a dedicated helper.
  - Add a test that selects the UTxO‑HD pool map and verifies all fields (pledge, cost, margin, reward, owners, relays, metadata, deposit).

<sup>Written for commit 32cee4233e69ceeab67f35dc6c640ece0303efc8. Summary will update on new commits.</sup>

### DIFF
--- a/ledgerstate/certstate.go
+++ b/ledgerstate/certstate.go
@@ -255,68 +255,7 @@ func parsePStateConway(data []byte) ([]ParsedPool, error) {
 		return nil, fmt.Errorf("decoding PState: %w", err)
 	}
 
-	// Find map elements sorted by size. PState contains several
-	// maps (pool params, future params, retiring, deposits).
-	// Pool params is the largest; deposits is second-largest
-	// (same key count but smaller values).
-	type mapEntry struct {
-		idx  int
-		size int
-	}
-	var maps []mapEntry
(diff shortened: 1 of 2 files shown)


arepala-uml (approved):

arepala-uml:
@wolf31o2 I am just curious about identification of this bug like how did the pstate zero-pools issue surface? Was it caught during snapshot import testing against UTxO-HD preview snapshots, or was it identified by empty pool listings, staking/reward queries returning nothing

wolf31o2:
During Mithril sync, it threw a warning about not loading pools.

PR #2309, opened 2026-05-14
fix(ledger): primary chain tip ancestor check

wolf31o2:
---
## Summary by cubic
Fixes ledger startup and chainsync when the ledger tip diverges or when the primary chain runs far ahead. We now verify the ledger tip against the selected chain, find the latest common ancestor, rewind as needed, and safely resume reading.

- **Bug Fixes**
  - If the ledger tip isn’t on the selected chain, find the latest common primary-chain ancestor via a 10k-slot nonce scan, rewind the primary chain to it, and roll back the ledger to match.
  - Respect security param `k`: when the primary chain is >k blocks ahead and the ledger tip is on the selected chain, rewind the primary chain to the ledger tip; otherwise, let chainsync catch up.
  - Treat the chain tip as “ahead” only when the ledger tip is on the selected chain and within the safety window; use a DB-backed membership check.
  - Make chain reading resilient: on a missing iterator start point, reconcile and retry up to 3 times with clearer logs; add tests for common-ancestor rollback, ahead-beyond-`k` rewind, and skipping primary-chain intersect when the ledger tip is missing.

<sup>Written for commit 6ab1f0643a8a217498612b12e459d45b64690e18. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ledger synchronization during startup with better reconciliation between primary chain and ledger state.
  * Enhanced handling of edge cases where the ledger tip is missing or diverges from the primary chain.
  * Added safety checks to prevent ledger inconsistencies when the primary chain is significantly ahead.

* **Tests**
  * Added comprehensive test cases for chainsync rollback reconciliation edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/blinklabs-io/dingo/pull/2309)

### DIFF
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -65,6 +65,7 @@ const (
 	cleanupConsumedUtxosInterval = 5 * time.Minute
 	batchSize                    = 50 // Number of blocks to process in a single DB transaction
 	ledgerIntersectDenseCount    = 32
+	ledgerAncestorSearchWindow   = 10_000
 	firstBlockIndex              = 1
 )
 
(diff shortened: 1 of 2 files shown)


ledger/state.go
```diff
@@ -2271,27 +2272,56 @@ func (ls *LedgerState) ledgerReadChain(
...
+				)
+				return
+			}
+			ls.config.Logger.Warn(
+				"chain iterator start point not on chain, attempting ledger rollback",
+				"error", err,
+				"start_slot", startPoint.Slot,
+				"start_hash", hex.EncodeToString(startPoint.Hash),
+			)
+			if reconcileErr := ls.reconcilePrimaryChainTipWithLedgerTip(); reconcileErr != nil {
```
arepala-uml, line 2314: reconcilePrimaryChainTipWithLedgerTip reads ls.currentTip without holding ls.RLock() and we added a call from ledgerReadChain which runs as a goroutine (go ls.ledgerReadChain(...), state.go:2403), so it's now a concurrent read racing against writers that hold ls.Lock(). Should be a one-line fix - please wrap the read with ls.RLock()/ls.RUnlock() same as like we are doing in other methods.

func (ls *LedgerState) reconcilePrimaryChainTipWithLedgerTip() error {
    if ls.chain == nil || ls.config.ChainManager == nil {
        return nil
    }
    ls.RLock()
    ledgerTip := ls.currentTip
    ls.RUnlock()

arepala-uml (dismissed):

arepala-uml (approved):

PR #2316, opened 2026-05-14
fix(peergov): guard clearing ConnectedAt tracking

wolf31o2:
## Summary by cubic
Prevents reconnection storms by gating outbound backoff and `ConnectedAt` clearing in `peergov`. Adds a configurable NtN chain-sync block timeout in `ouroboros` with a safe minimum, applied via options without mutating protocol globals.

- **Bug Fixes**
  - Clear `ConnectedAt` only for outbound peers after backoff accounting; compute backoff from the connection-closed timestamp and clamp/log negative durations.
  - Apply backoff only when we will reconnect (not inbound or deny-listed); reset for stable links and exponentially back off short‑lived ones; avoid duplicate reconnect goroutines.
  - Added tests for stable outbound resets, short‑lived outbound backoff, and negative‑duration clamp logging.

- **New Features**
  - Introduced `OuroborosConfig.ChainsyncBlockTimeout` with a floor at `ochainsync.MustReplyTimeoutMax`.
  - Applied via `WithBlockTimeout` on client and server; verified by tests and does not mutate `ochainsync` NtN state timeouts.

<sup>Written for commit 2c23b2aa10350afa130152c62d3297cf6a6c50fa. Summary will update on new commits.</sup>

### DIFF
--- a/ouroboros/ouroboros.go
+++ b/ouroboros/ouroboros.go
@@ -86,6 +86,10 @@ type OuroborosConfig struct {
 	PeerSharing     bool
 	IntersectTip    bool
 	PromRegistry    prometheus.Registerer
+	// ChainsyncBlockTimeout bounds how long NtN chain-sync can wait for a
+	// block reply after a peer enters a server-agency state. Low-density
+	// networks may legitimately go longer than the default protocol window.
+	ChainsyncBlockTimeout time.Duration
 	// MaxTxSubmissionsPerSecond is the maximum number of transaction
 	// submissions accepted per peer per second via the TxSubmission
 	// mini-protocol. A value of 0 disables rate limiting, which is the
@@ -120,6 +124,9 @@ func NewOuroboros(cfg OuroborosConfig) *Ouroboros {
 		cfg.Logger = slog.New(slog.NewJSONHandler(io.Discard, nil))
 	}
 	cfg.Logger = cfg.Logger.With("component", "ouroboros")
+	cfg.ChainsyncBlockTimeout = effectiveChainsyncBlockTimeout(
+		cfg.ChainsyncBlockTimeout,
+	)
 	o := &Ouroboros{
 		config:           cfg,
 		EventBus:         cfg.EventBus,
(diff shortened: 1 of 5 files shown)


peergov/connections.go
```diff
@@ -667,40 +667,45 @@ func (p *PeerGovernor) handleConnectionClosedEvent(evt event.Event) {
...
-				if peer.ReconnectDelay == 0 {
-					peer.ReconnectDelay = initialReconnectDelay
-				} else if peer.ReconnectDelay < maxReconnectDelay {
-					peer.ReconnectDelay *= reconnectBackoffFactor
-					if peer.ReconnectDelay > maxReconnectDelay {
-						peer.ReconnectDelay = maxReconnectDelay
+			if shouldReconnect {
+				connDur := connClosedAt.Sub(peer.ConnectedAt)
+				if connDur < 0 {
+					connDur = 0
```
arepala-uml, line 688: Just like a suggestion - adding a log statement for this negative duration will be better. Also, I am trying to understand, like which cases we will run into this negative connection duration

ouroboros/ouroboros.go
```diff
@@ -120,6 +124,10 @@ func NewOuroboros(cfg OuroborosConfig) *Ouroboros {
 		cfg.Logger = slog.New(slog.NewJSONHandler(io.Discard, nil))
 	}
 	cfg.Logger = cfg.Logger.With("component", "ouroboros")
+	cfg.ChainsyncBlockTimeout = effectiveChainsyncBlockTimeout(
+		cfg.ChainsyncBlockTimeout,
+	)
+	configureChainsyncNtNBlockTimeout(cfg.ChainsyncBlockTimeout)
```
arepala-uml: It is not a risk but it seems to be redundant calls to this function (effectiveChainsyncBlockTimeout). configureChainsyncNtNBlockTimeout calls effectiveChainsyncBlockTimeout again on an already-clamped value.

arepala-uml (dismissed):
Just left a few minor comments

wolf31o2:
@cubic-dev-ai review

arepala-uml (approved):

PR #2496, opened 2026-06-03
feat(ledger): LocalStateQuery implement GetLedgerPeerSnapshot

### LINKED ISSUE #2492: LocalStateQuery: implement GetLedgerPeerSnapshot
## Summary

Implement the LocalStateQuery `GetLedgerPeerSnapshot` surface so Dingo can provide the Cardano ledger-peer state expected by dmq-node-compatible clients.

This is the only missing Dingo Cardano-state/Ouroboros surface identified for dmq-node support. Dingo does not need to import `dmq-lib`; dmq-lib/dmq-node can remain external consumers of Dingo's normal Ouroboros interfaces.

## What Needs to Happen

1. Add server-side handling for the gouroboros LocalStateQuery `GetLedgerPeerSnapshot` query.
2. Source the snapshot from Dingo ledger/database state and include relay metadata needed by ledger-peer discovery:
   - single-host address relays
   - single-host name relays
   - multi-host/SRV relay information
   - stake values needed by the snapshot result
3. Keep the implementation on the normal LocalStateQuery path so future ViewManager snapshot isolation can refine point-in-time behavior without changing the query surface.
4. Add focused tests for query dispatch, result mapping, empty snapshots, stake population, and multi-host relay handling.
5. Document any temporary current-tip behavior if full acquire/release snapshot isolation is not available yet.

## Minimal Scope

This issue is only for the Cardano-state query surface needed by dmq-node-style ledger peer discovery.

It does not include:

- importing `dmq-lib` into Dingo
- implementing Dingo's DMQ runtime
- stake-weighted Cardano peer selection
- peer scoring changes
- broader peer-governance policy changes

This issue may satisfy the DMQ-relevant data portions of #1787 and #1789, but it should not wait on the full #1787-#1796 peer-governance roadmap.

## Acceptance Criteria

- Dingo's N2C LocalStateQuery server answers `GetLedgerPeerSnapshot`.
- The response contains the relay and stake information needed by dmq-node-compatible ledger peer discovery.
- The implementation does not introduce new imports from DMQ code into `ledger/`, `database/`, `peergov/`, `ouroboros/`, or `chainselection/`.
- Tests cover the query handler and ledger/database-to-query-result mapping.

## Roadmap

Target this before the Q4 2026 DMQ integration work; this is an Ouroboros/Cardano-state compatibility prerequisite.

cryptodj413:
Closes #2492 

---
## Summary by cubic
Implements `GetLedgerPeerSnapshot` for `LocalStateQuery` (NtC v19+) to return a stake‑weighted snapshot of active pools’ relay endpoints for peer discovery, with correct big‑peer quota handling, wired into the ledger query path.

- **New Features**
  - Returns V1 snapshot with a WithOrigin slot sourced from the current tip in the same read transaction; Origin at genesis, At 0 for a real slot‑0 block.
  - Loads active pools and delegated stake; orders by stake desc with key‑hash tie‑break.
  - Excludes pools without relays but includes their stake in the total.
  - Maps relays to access points: IPv4/IPv6+port, domain+port, SRV; dual‑stack expands to two entries; malformed hostname ports fall back to SRV (no port).
  - Supports `LedgerPeerKindAll` and `LedgerPeerKindBig` with a 0.9 cumulative‑stake quota and correct boundary semantics (the quota‑crossing pool is included; later pools are excluded).
  - Adds tests for dispatch, slot/origin handling, ordering and stake math, big‑peer boundary, relay‑kind mapping, malformed port, and CBOR round‑trip.

<sup>Written for commit d6122eea058e4964abf69da5f1e4ed6aa898a33d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2496?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **New Features**
  * Implemented peer snapshot querying for ledger peer discovery supporting protocol v19+, enabling nodes to efficiently discover and select peers based on stake-weighted distribution.

* **Tests**
  * Added comprehensive test coverage for peer snapshot querying, stake calculations, peer filtering, quota-based peer selection, relay endpoint handling, and serialization compatibility.

### DIFF
--- a/ledger/peer_snapshot_query.go
+++ b/ledger/peer_snapshot_query.go
@@ -16,6 +16,7 @@ package ledger
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"math"
 	"math/big"
@@ -23,17 +24,19 @@ import (
 	"sort"
 
 	"github.com/blinklabs-io/dingo/database/models"
+	"github.com/blinklabs-io/dingo/database/types"
 	"github.com/blinklabs-io/gouroboros/cbor"
 	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
 	olocalstatequery "github.com/blinklabs-io/gouroboros/protocol/localstatequery"
 )
 
 // bigLedgerPeerQuota is the cumulative relative-stake threshold that
 // delimits the "big" ledger peers: walking pools from highest to lowest
 // stake, a pool belongs to the big set while the stake accumulated by the
-// pools before it has not yet exceeded this quota (so the first pool that
-// crosses the quota is still included). The value mirrors the
-// bigLedgerPeerQuota default (0.9) used by ouroboros-network's
+// pools before it has not yet reached this quota (so the first pool that
+// reaches or crosses the quota is the last one included). The value mirrors
+// the bigLedgerPeerQuota default (0.9) used by ouroboros-network's
 // PeerSelection.LedgerPeers. It is applied only when a client requests
 // LedgerPeerKindBig; LedgerPeerKindAll returns every relay-advertising pool.
 //
@@ -56,13 +59,25 @@ var bigLedgerPeerQuota = big.NewRat(9, 10)
 func (ls *LedgerState) queryLedgerPeerSnapshot(
 	peerKind olocalstatequery.LedgerPeerKind,
 ) (any, error) {
-	slot := ls.ChainTipSlot()
-
 	txn := ls.db.Transaction(false)
 	defer txn.Release()
 
+	// Read the tip from the same read transaction as the pool/stake data so
+	// the reported snapshot slot describes the exact point the relays and
+	// stake were observed at, even under a concurrent block apply or rollback.
+	tip, err := ls.db.GetTip(txn)
+	if err != nil {
+		return nil, fmt.Errorf("GetLedgerPeerSnapshot: get tip: %w", err)
+	}
+	slot := withOriginSlot(tip.Point)
+
 	pkhBytes, err := ls.db.Metadata().GetActivePoolKeyHashes(txn.Metadata())
 	if err != nil {
+		// Epoch data may not be synced yet for the current tip (early sync);
+		// there are simply no ledger peers to report at that point.
+		if errors.Is(err, types.ErrNoEpochData) {
+			return emptyLedgerPeerSnapshot(slot), nil
+		}
 		return nil, fmt.Errorf(
 			"GetLedgerPeerSnapshot: get active pools: %w",
 			err,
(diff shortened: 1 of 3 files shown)


ledger/peer_snapshot_query.go
```diff
@@ -0,0 +1,288 @@
...
+		return bytes.Compare(entries[i].keyHash, entries[j].keyHash) < 0
+	})
+
+	acc := new(big.Rat)
+	for _, e := range entries {
+		// For the big-peer set, stop once the stake accumulated by the
+		// preceding pools has exceeded the quota. The pool that first crosses
+		// the quota is still included (takeWhilePrev semantics).
+		if peerKind == olocalstatequery.LedgerPeerKindBig &&
+			acc.Cmp(bigLedgerPeerQuota) > 0 {
```
wolf31o2: This should stop when the accumulated stake has reached the big-ledger quota, not only after it exceeds it. With stakes 80/10/10, `acc` is exactly 0.9 after the second pool, so this condition still includes the third pool. That also means zero-stake relayed pools can be included after an exact 90% boundary. Please switch this to threshold-reached semantics (for example `>=` before adding the next pool, or break immediately after the pool that reaches/crosses the quota) and add an exact-threshold regression test.

ledger/peer_snapshot_query.go
```diff
@@ -0,0 +1,288 @@
...
+//
+// Point-in-time behavior: LocalStateQuery Acquire/Release are still no-ops
+// pending ViewManager snapshot isolation (#382), so the snapshot reflects the
+// current chain tip rather than the acquired point. The reported slot is the
+// current tip slot. When that isolation lands, only the data-sourcing here
+// needs to observe the acquired view; the query surface stays the same.
+func (ls *LedgerState) queryLedgerPeerSnapshot(
+	peerKind olocalstatequery.LedgerPeerKind,
+) (any, error) {
+	slot := ls.ChainTipSlot()
```
wolf31o2: The reported snapshot slot is read from in-memory `LedgerState` before the DB read transaction is opened, while the active pools/stake come from that DB transaction. A concurrent block apply or rollback can make the `WithOrigin` slot describe a different point than the pool data. This also collapses any real slot-0 tip into `Origin` through `withOriginSlot`, while the rest of the repo treats origin as slot 0 with an empty hash. Please read the tip from the same DB transaction (`ls.db.GetTip(txn)`) and derive `HasSlot` from the full point/hash.

wolf31o2 (changes requested):
Requesting changes based on the local review.

Main blockers:
- `LedgerPeerKindBig` over-includes peers when the cumulative stake lands exactly on the 0.9 quota.
- The snapshot slot should come from the same DB snapshot as the pool/stake data, and origin should be distinguished from a real slot-0 point.

Local checks passed:
- `go test -v -race -run Test.*LedgerPeerSnapshot ./ledger`
- `go test -v -race ./ledger`

wolf31o2 (approved):

PR #2510, opened 2026-06-06
fix(blockfrost): compute network locked supply and align network eras…

### LINKED ISSUE #2490: blockfrost: compute network locked supply and align network/eras schema
## Summary

Finish OpenAPI alignment for requested network endpoints.

Source of truth: Blockfrost OpenAPI `0.1.88`, schemas `network` and `network-eras`.

## Gaps

- `GET /network` schema describes `supply.locked` as script-locked supply. Dingo currently hard-codes locked supply to `"0"` in `NodeAdapter.Network`, which also affects circulating supply.
- `GET /network/eras` OpenAPI items contain `start`, `end`, and `parameters`; Dingo additionally emits an `era` field that is not in the current schema.

## Acceptance Criteria

- `/network` computes script-locked supply from ledger/database state instead of hard-coding zero.
- Circulating supply uses the computed locked supply.
- `/network/eras` response shape is reviewed against OpenAPI `0.1.88`; extra fields are removed or intentionally documented if Dingo chooses a compatible extension.
- Tests cover network supply with script-locked UTxOs and era summary response shape.

cryptodj413:
Closes #2490 

---
## Summary by cubic
Compute `supply.locked` in `blockfrost` `/network` by summing lovelace in live script-payment UTxOs, and subtract it from `supply.circulating`. Align `/network/eras` with OpenAPI 0.1.88 by omitting the era name.

- **New Features**
  - Add `payment_script` to `utxo` with composite index `(deleted_slot, payment_script, amount)`; set from address type during indexing and ledger import.
  - Implement `GetScriptLockedSupply` for SQLite, Postgres, and MySQL; wire into `/network`.
  - Close `gorm.PreparedStmtDB` on shutdown (SQLite read/write, Postgres, MySQL).

- **Migration**
  - Run DB migrations to add `payment_script` and index `idx_utxo_deleted_payment_script`.
  - After upgrade, `supply.locked` counts only UTxOs created post-upgrade until you rebuild from immutable data (`dingo load` or Mithril import). Clients: `/network/eras` items no longer include the `era` field.

<sup>Written for commit 851a88ae16f436d997a7a0b39dc79b3fca03e1e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2510?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **New Features**
  * Network API endpoint now returns accurate script-locked supply information instead of a placeholder value.

* **Updates**
  * Network eras endpoint response structure updated; era name field removed.
  * Database infrastructure enhanced to track and query script-locked supply metrics.

### DIFF
--- a/blockfrost/adapter.go
+++ b/blockfrost/adapter.go
@@ -516,9 +516,16 @@ func (a *NodeAdapter) Network() (NetworkInfo, error) {
 		treasury = uint64(state.Treasury)
 		reserves = uint64(state.Reserves)
 	}
-	// TODO: Wire locked supply from script-address UTxOs and ledger deposits,
-	// then subtract it from circulating supply.
-	locked := uint64(0)
+	// Script-locked supply: lovelace held in live UTxOs whose payment
+	// credential is a script. Subtracted from circulating supply below.
+	locked, err := a.ledgerState.Database().Metadata().
+		GetScriptLockedSupply(nil)
+	if err != nil {
+		return NetworkInfo{}, fmt.Errorf(
+			"get script-locked supply: %w",
+			err,
+		)
+	}
 	total := uint64(0)
 	if reserves <= maxSupply {
 		total = maxSupply - reserves
--- a/database/models/utxo.go
+++ b/database/models/utxo.go
@@ -76,9 +76,16 @@ type Utxo struct {
 	CollateralByTxId        []byte       `gorm:"index;size:32"`
 	ID                      uint         `gorm:"primarykey"`
 	AddedSlot               uint64       `gorm:"index"`
-	DeletedSlot             uint64       `gorm:"index:idx_utxo_deleted_staking_amount,priority:1"`
-	Amount                  types.Uint64 `gorm:"index:idx_utxo_deleted_staking_amount,priority:3"`
+	DeletedSlot             uint64       `gorm:"index:idx_utxo_deleted_staking_amount,priority:1;index:idx_utxo_deleted_payment_script,priority:1"`
+	Amount                  types.Uint64 `gorm:"index:idx_utxo_deleted_staking_amount,priority:3;index:idx_utxo_deleted_payment_script,priority:3"`
 	OutputIdx               uint32       `gorm:"uniqueIndex:tx_id_output_idx"`
+	// PaymentScript is true when the output's payment credential is a
+	// script hash (as opposed to a key hash). It is derived from the
+	// address type at index time and used to compute the network's
+	// script-locked supply (see GetScriptLockedSupply). The composite
+	// index (deleted_slot, payment_script, amount) lets the supply sum
+	// scan only live script UTxOs.
+	PaymentScript bool `gorm:"index:idx_utxo_deleted_payment_script,priority:2"`
 }
 
 // UtxoWithOrdering includes UTxO with transaction ordering metadata
@@ -149,6 +156,13 @@ func UtxoLedgerToModel(
 	pkh := outAddr.PaymentKeyHash()
 	if pkh != zeroHash {
 		ret.PaymentKey = pkh.Bytes()
+	}
+	// The low bit of the address type distinguishes a script payment
+	// credential from a key-hash credential. Byron addresses (type
+	// 0b1000) never have this bit set, so they are correctly treated
+	// as non-script.
+	if outAddr.Type()&lcommon.AddressTypeScriptBit == lcommon.AddressTypeScriptBit {
+		ret.PaymentScript = true
 	}
 	skh := outAddr.StakeKeyHash()
 	if skh != zeroHash {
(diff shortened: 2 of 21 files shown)


database/models/utxo.go
```diff
@@ -16,12 +16,71 @@ package models
...
 import (
 	"errors"
+	"fmt"
+	"log/slog"
 
 	"github.com/blinklabs-io/dingo/database/types"
 	"github.com/blinklabs-io/gouroboros/ledger"
 	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
```
wolf31o2: This is plugin-specific and does not belong in the general models or types. Nothing outside plugin should inherit any dependency from within plugin.

database/models/utxo.go
```diff
@@ -76,9 +76,16 @@ type Utxo struct {
...
+	DeletedSlot             uint64       `gorm:"index:idx_utxo_deleted_staking_amount,priority:1;index:idx_utxo_deleted_payment_script,priority:1"`
+	Amount                  types.Uint64 `gorm:"index:idx_utxo_deleted_staking_amount,priority:3;index:idx_utxo_deleted_payment_script,priority:3"`
 	OutputIdx               uint32       `gorm:"uniqueIndex:tx_id_output_idx"`
+	// PaymentScript is true when the output's payment credential is a
+	// script hash (as opposed to a key hash). It is derived from the
+	// address type at index time and used to compute the network's
+	// script-locked supply (see GetScriptLockedSupply). The composite
+	// index (deleted_slot, payment_script, amount) lets the supply sum
+	// scan only live script UTxOs.
+	PaymentScript bool `gorm:"index:idx_utxo_deleted_payment_script,priority:2"`
```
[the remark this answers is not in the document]
wolf31o2: This is fine. Our storage API isn't stable, yet.

database/models/utxo.go
```diff
@@ -16,12 +16,71 @@ package models
...
+// instead of silently returning an under-counted value.
+//
+// Existing rows all default to payment_script=false after AutoMigrate; only
+// UTxOs indexed after the upgrade carry the correct value. The sentinel is
+// removed after a full database rebuild (Mithril import or `dingo load`).
+//
+// The function is a no-op when:
+//   - the utxo table does not yet exist (fresh database — no backfill needed)
+//   - the payment_script column already exists (already migrated)
+func MigrateUtxoPaymentScript(db *gorm.DB, logger *slog.Logger) error {
```
wolf31o2: Get rid of this. We don't have any backfill like this for doing partial backfill. Our storage API is not stable. We don't support "upgrade" of storage schema. I would drop all of these changes and just let AutoMigrate do its job for initial creation.

If this were to be done, it would be done within the plugins. The fact that all metadata plugins share gorm is an accident, not a feature of metadata plugins. We don't want gorm to leak anywhere outside the plugins.

wolf31o2 (changes requested):

database/models/utxo.go
cryptodj413: Thanks for the review. Let me think again

blockfrost/adapter.go
```diff
@@ -516,9 +516,16 @@ func (a *NodeAdapter) Network() (NetworkInfo, error) {
 		treasury = uint64(state.Treasury)
 		reserves = uint64(state.Reserves)
 	}
-	// TODO: Wire locked supply from script-address UTxOs and ledger deposits,
-	// then subtract it from circulating supply.
-	locked := uint64(0)
+	// Script-locked supply: lovelace held in live UTxOs whose payment
+	// credential is a script. Subtracted from circulating supply below.
+	locked, err := a.ledgerState.Database().Metadata().
```
wolf31o2, line 521: This aggregate assumes `payment_script` is populated for every live UTxO. Fresh ledger-state/Mithril imports do not currently set that field: `ledgerstate.UTxOToModel` builds `models.Utxo` from `ParsedUTxO` without `PaymentScript`, so `/network` will undercount locked supply after a snapshot import until new block-produced outputs are indexed. Please derive the script-payment flag in the import path as well.

database/models/utxo.go
wolf31o2, line 88: This invariant is only true for rows created through `UtxoLedgerToModel`. Ledger-state import bypasses this converter: `extractAddressKeys` already sees address types 1/3/5/7 as script payment credentials, but `ParsedUTxO` does not carry that boolean and `UTxOToModel` never sets `PaymentScript`. The new column needs to be populated there too, with coverage for imported script-address UTxOs.

wolf31o2 (changes requested):
Requesting changes. Dropping/rebuilding old DBs is fine for pre-1.0; this is a correctness issue for fresh ledger-state/Mithril imports. The new /network aggregate relies on payment_script being populated for the whole live UTxO set, but the snapshot import path currently inserts rows with the default false value.

wolf31o2 (approved):

PR #2523, opened 2026-06-09
fix: epoch clock stall

wolf31o2:
---
## Summary by cubic
Fixes an epoch-clock stall by clearing the cached upstream tip when the active chainsync connection is closed or unavailable. Adds stable, nil-safe endpoint-based connection identity across reconnects to keep cleanup, priority, and eligibility updates correct.

- **Bug Fixes**
  - `ledger`: clear `syncUpstreamTipSlot` when the active chainsync connection closes or when no active connection exists; use per-address strings for connection-ID keys to handle nil/partial addresses.
  - `peergov`, `ouroboros`: use endpoint-based, nil-safe `sameConnectionId` matching across peer lookup, chain-selection, and chainsync logic (handles zero/partial IDs consistently).
  - Added tests for upstream-tip reset and connection-ID matching; updated ARCHITECTURE notes.

<sup>Written for commit 65ba855752af103a4971a05cfadb74dd786a0791. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2523?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **Bug Fixes**
  * Fixed sync progress reporting to properly reset when active peer connections are lost
  * Improved peer connection identity matching for more reliable reconnection and recovery
  * Enhanced ledger state cleanup when all active connections become unavailable

### DIFF
--- a/peergov/peers.go
+++ b/peergov/peers.go
@@ -395,11 +395,26 @@ func topologyGroupIDForPeer(peer *Peer, isTopology bool) string {
 func (p *PeerGovernor) peerIndexByConnId(connId ouroboros.ConnectionId) int {
 	for i, peer := range p.peers {
 		if peer != nil && peer.Connection != nil &&
-			peer.Connection.Id == connId {
+			sameConnectionId(peer.Connection.Id, connId) {
 			return i
 		}
 	}
 	return -1
+}
+
+// sameNetAddr compares addresses by string form, treating nil as equal
+// only to nil. ConnectionId.String() panics when either net.Addr field
+// is nil, so the addresses are compared individually instead.
+func sameNetAddr(a, b net.Addr) bool {
+	if a == nil || b == nil {
+		return a == nil && b == nil
+	}
+	return a.String() == b.String()
+}
+
+func sameConnectionId(a, b ouroboros.ConnectionId) bool {
+	return sameNetAddr(a.LocalAddr, b.LocalAddr) &&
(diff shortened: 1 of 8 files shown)


peergov/peers.go
```diff
@@ -395,13 +395,37 @@ func topologyGroupIDForPeer(peer *Peer, isTopology bool) string {
...
+	keyA := connIdKey(a)
+	keyB := connIdKey(b)
+	if keyA == "" || keyB == "" {
+		return keyA == keyB
+	}
+	return keyA == keyB
+}
+
 func (p *PeerGovernor) SetPeerHotByConnId(connId ouroboros.ConnectionId) {
 	p.mu.Lock()
```
cryptodj413: How about using existing String() function of ConnectionID instead of new connIdKey function?

cryptodj413:
```go
func (c ConnectionId) String() string {
	return fmt.Sprintf(
		"%s<->%s",
		c.LocalAddr.String(),
		c.RemoteAddr.String(),
	)
}
```

cryptodj413 (approved):

PR #2606, opened 2026-06-23
feat(midnight): Implement Midnight indexer block scanning for cNIGHT …

### LINKED ISSUE #2114: Implement Midnight indexer block scanning for cNIGHT and registration events
The Midnight indexer subscribes to ChainBlockEvent and scans each block for Midnight-relevant transactions, then writes to the indexer tables.

This issue covers four of the six scan flows.

1. cNIGHT create scan. For each transaction output, check whether it contains a token matching cnight_policy_id and cnight_asset_name. If yes, write an asset_create row.

2. cNIGHT spend scan. For each transaction input that consumes a previously-recorded cNIGHT UTxO, write an asset_spend row. Use an in-memory tracked-UTxO set keyed by (tx_hash, output_index) for O(1) lookups. Load the set on startup from midnight_asset_creates minus midnight_asset_spends.

3. Registration scan. For each transaction output at mapping_validator_address that contains an auth token (policy derived from the script address plus auth_token_asset_name), write a registration row with the inline datum.

4. Deregistration scan. For each transaction input that consumes a registration UTxO at mapping_validator_address with the auth token, write a deregistration row.

Out of scope: governance, Ariadne, candidate tracking, gRPC, rollback. Tests should cover happy path plus blocks with mixed relevant and non-relevant txs.

cryptodj413:
Closes #2114 

---
## Summary by cubic
Adds a Midnight block indexer for cNIGHT creates/spends and mapping‑validator registrations/deregistrations. It runs only in API mode, starts before `LedgerState` with a checkpointed backfill and blocking event delivery, writes to `midnight_*` tables with idempotent inserts and rollback‑safe deletes, restores tracked UTxOs on startup, and can scope auth tokens via `midnight.AuthTokenPolicyID`.

- **New Features**
  - Cross‑DB metadata store APIs for `midnight_*` tables (MySQL/Postgres/SQLite): `Create*`, `FindUnspent*`, and `Delete*ByBlock`; unique composite indexes on UTxO keys with `ON CONFLICT DO NOTHING` for safe replays.
  - Registration auth tokens can be restricted to a specific policy using `midnight.AuthTokenPolicyID` (optional).

- **Bug Fixes**
  - Startup gap: create/start the Midnight indexer before `LedgerState` and run a synchronous backfill from the last checkpoint via the DB `BlockIterator`.
  - Event delivery and checkpointing: publish block events using blocking delivery with error handling; update the Midnight checkpoint only after all block writes succeed.
  - Rollbacks: fix `Delete*ByBlock` across MySQL/Postgres/SQLite to return deleted rows for in‑memory state reconciliation.

<sup>Written for commit e65db4e616587eaf825b658afe5dbae2427e95c5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2606?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **New Features**
  * Added Midnight Indexer component to track cNIGHT asset creation and spending events across the blockchain.
  * Implemented persistent storage for Midnight asset operations and validator registrations with support for MySQL, PostgreSQL, and SQLite databases.

* **Tests**
  * Added comprehensive test suite for Midnight Indexer functionality.


midnight/indexer/indexer.go
```diff
@@ -0,0 +1,441 @@
...
+
+// handleBlockEvent is the SubscribeFunc callback; it decodes the block and
+// delegates to processBlock.
+func (idx *Indexer) handleBlockEvent(evt event.Event) {
+	blockEvt, ok := evt.Data.(ledger.BlockEvent)
+	if !ok {
+		return
+	}
+	// Rollbacks are out of scope for this indexer.
+	if blockEvt.Action != ledger.BlockActionApply {
```
wolf31o2: Blocking: ledger emits `BlockActionUndo` events on rollback, but this indexer drops them and the rollback DB path does not prune these `midnight_*` rows. A fork that spends a tracked Midnight UTxO can leave the spend row and remove the UTxO from the in-memory map, so replay of the canonical spend can be missed. Please handle undo/rollback for persisted rows and tracked maps, with rollback tests.

midnight/indexer/indexer.go
```diff
@@ -0,0 +1,441 @@
...
+	// Scan inputs for spends of tracked UTxOs.
+	for _, inp := range tx.Inputs() {
+		inpHashHex := hex.EncodeToString(inp.Id().Bytes())
+		inpIndex := inp.Index()
+		key := utxoKey{TxHash: inpHashHex, Index: inpIndex}
+
+		idx.mu.Lock()
+		utxo, isCNight := idx.cNightUTxOs[key]
+		if isCNight {
+			delete(idx.cNightUTxOs, key)
```
wolf31o2: Please do not remove the tracked UTxO until the corresponding spend/deregistration row is durable, or restore it on error. Today a transient insert failure only logs the error but leaves the in-memory set missing this UTxO; after restart the DB can also reload it as unspent because the spend row was never written.

midnight/indexer/indexer.go
```diff
@@ -0,0 +1,441 @@
...
+	if cfg.CNightPolicyID != "" && cfg.CNightAssetName != "" {
+		policyBytes, err := hex.DecodeString(cfg.CNightPolicyID)
+		if err != nil {
+			return nil, fmt.Errorf("invalid cnight_policy_id: %w", err)
+		}
+		assetNameBytes, err := hex.DecodeString(cfg.CNightAssetName)
+		if err != nil {
+			return nil, fmt.Errorf("invalid cnight_asset_name: %w", err)
+		}
+		idx.cnightPolicyID = lcommon.NewBlake2b224(policyBytes)
```
wolf31o2, line 137: `hex.DecodeString` does not validate that this is exactly a 28-byte policy ID, and `NewBlake2b224` silently pads or truncates by copy. Please reject non-56-hex policy IDs for both cNIGHT and auth-token policy config so bad config cannot silently index the wrong policy or no policy.

database/models/midnight.go
```diff
@@ -24,8 +24,8 @@ type MidnightAssetCreate struct {
 	ID               uint   `gorm:"primarykey"`
 	Address          []byte `gorm:"not null"`
 	Quantity         uint64 `gorm:"not null"`
-	TxHash           []byte `gorm:"size:32;not null"`
-	OutputIndex      uint32 `gorm:"not null"`
+	TxHash           []byte `gorm:"index:idx_midnight_asset_creates_utxo_lookup,priority:1;size:32;not null"`
```
wolf31o2: These new UTxO lookup indexes, along with the new unspent lookup methods on the metadata-store API, need to be reflected in `DATABASE.md`. The current Midnight section still documents only the block/tx indexes, and this repo treats model/index/API surface docs as part of the change bar.

internal/config/config.go
```diff
@@ -336,6 +336,7 @@ type MidnightConfig struct {
 	CNightPolicyID              string `yaml:"cnightPolicyId"`
 	CNightAssetName             string `yaml:"cnightAssetName"`
 	MappingValidatorAddress     string `yaml:"mappingValidatorAddress"`
+	AuthTokenPolicyID           string `yaml:"authTokenPolicyId"`
```
wolf31o2, line 339: Please also add this new YAML field to `dingo.yaml.example`; otherwise operators will not discover the auth-token policy scoping setting.

wolf31o2 (changes requested):
Requesting changes. The focused local tests passed, but the new Midnight indexer needs rollback/error handling and the required docs/example config updates before this is mergeable.

node.go
```diff
@@ -1054,6 +1056,28 @@ func (n *Node) Run(ctx context.Context) error {
 			}
 		})
 		started = append(started, n.startDeferredIndexMaintenance())
+
+		// Midnight indexer: start unconditionally in API mode (gRPC port
+		// may be zero but indexing should always run).
+		midnightIdx, err := midnightindexer.New(midnightindexer.Config{
```
wolf31o2: This starts the Midnight indexer after `LedgerState.Start` and after peer/network startup. During normal startup/catch-up, blocks can be applied before this subscription exists, and there is no checkpoint or backfill path to recover those missed blocks. Please either subscribe/start the indexer before any block application can occur, or add durable catch-up from stored blocks.

midnight/indexer/indexer.go
```diff
@@ -0,0 +1,540 @@
...
+			idx.regUTxOs[key] = registrationUTxO{FullDatum: r.FullDatum}
+		}
+	}
+	return nil
+}
+
+// Start subscribes to ledger block events. The indexer begins processing
+// blocks immediately after Start returns.
+func (idx *Indexer) Start() {
+	idx.subID = idx.config.EventBus.SubscribeFuncWithBuffer(
```
wolf31o2, line 227: This uses the lossy EventBus subscriber path for a durable index. `channelSubscriber.Deliver` drops events when the buffer is full, so a bulk-sync burst can silently omit `ledger.block` events and leave the Midnight tables incomplete. Please use a non-lossy delivery/checkpointed processing path, or add replay/backfill that makes dropped events recoverable.

wolf31o2 (changes requested):
Requesting changes. Ignoring rollback handling, the indexer still needs a non-lossy apply path before this is mergeable: it currently subscribes after block application can already have started, and the EventBus delivery mode can drop block events with no checkpoint/backfill recovery.

midnight/indexer/indexer.go
```diff
@@ -0,0 +1,633 @@
...
+func (idx *Indexer) Start() {
+	if idx.config.BlockIterator != nil {
+		if err := idx.backfill(); err != nil && idx.config.Logger != nil {
+			idx.config.Logger.Warn(
+				"midnight indexer: backfill incomplete; missed blocks will be recovered on next restart",
+				"error", err,
+			)
+		}
+	}
+	idx.subID = idx.config.EventBus.SubscribeFuncWithBuffer(
```
wolf31o2, line 227: This is still a lossy subscription for an ordering-critical stream. `SubscribeFuncWithBuffer` ultimately uses `channelSubscriber.Deliver`, which drops the event when the subscriber channel is full. Since `ledger.publishBlockEvent` currently uses `EventBus.Publish`, a dropped `BlockActionApply` can be followed by later delivered blocks that advance the Midnight checkpoint past the missing block.

Please make `ledger.block` delivery blocking/fatal before relying on this subscriber for index completeness. One concrete fix is to change `ledger/block_event.go` to publish this stream with `PublishBlocking` and stop the node if delivery fails:

```go
func (ls *LedgerState) publishBlockEvent(
	action BlockAction,
	block models.Block,
) {
	if ls.config.EventBus == nil {
		return
	}
	evt := BlockEvent{
		Action: action,
		Block:  block,
		Point: ocommon.Point{
			Slot: block.Slot,
			Hash: block.Hash,
		},
	}
	if err := ls.config.EventBus.PublishBlocking(
		BlockEventType,
		event.NewEvent(BlockEventType, evt),
	); err != nil {
		publishErr := fmt.Errorf(
			"publish %s block event at slot %d block %d: %w",
			action,
			block.Slot,
			block.Number,
			err,
		)
		ls.config.Logger.Error(
			"failed to publish ledger block event",
			"component", "ledger",
			"error", publishErr,
			"action", action,
			"slot", block.Slot,
			"block_number", block.Number,
		)
		if ls.config.FatalErrorFunc != nil {
			ls.config.FatalErrorFunc(publishErr)
		}
	}
}
```

This keeps the EventBus path, but removes the drop-on-full gap for `ledger.block`. If you do not want to make all `ledger.block` subscribers blocking, then the Midnight indexer needs a separate persisted queue or periodic catch-up loop that does not depend on delivered live events to notice missed blocks.

midnight/indexer/indexer.go
```diff
@@ -0,0 +1,633 @@
...
+			}
+			return
+		}
+		var timestampMs uint64
+		if idx.config.SlotTimer != nil {
+			if t, err := idx.config.SlotTimer.SlotToTime(block.Slot); err == nil {
+				timestampMs = uint64(t.UnixMilli()) //nolint:gosec
+			}
+		}
+		idx.processBlock(block, decoded.Transactions(), timestampMs)
```
wolf31o2: This advances the checkpoint even when indexing the block failed. `processBlock`, `processTx`, and `processOutput` currently return no error, and each `CreateMidnight*` failure is logged and swallowed. A transient DB error can leave a Midnight row absent, then `updateCheckpoint(block.Slot)` persists progress past that block; restart backfill will not replay it.

Please make block processing fail-fast and only update the checkpoint after all writes for the block succeeded. Also fatal/stop after a live processing failure so a later block cannot advance the checkpoint over the failed slot. The concrete shape I would expect here is:

```go
type Config struct {
	EventBus  *event.EventBus
	Metadata  metadata.MetadataStore
	SlotTimer SlotTimer
	Logger    *slog.Logger
	// FatalErrorFunc is called when the indexer cannot safely continue
	// without risking a checkpoint gap. Node wiring should cancel shutdown.
	FatalErrorFunc func(error)

	CNightPolicyID          string
	CNightAssetName         string
	MappingValidatorAddress string
	AuthTokenPolicyID       string
	AuthTokenAssetName      string
	BlockIterator           func(startSlot, endSlot uint64, fn func(models.Block) error) error
}

func (idx *Indexer) fatal(err error) {
	if idx.config.Logger != nil {
		idx.config.Logger.Error("midnight indexer fatal error", "error", err)
	}
	if idx.config.FatalErrorFunc != nil {
		idx.config.FatalErrorFunc(err)
	}
}

func (idx *Indexer) handleBlockEvent(evt event.Event) {
	blockEvt, ok := evt.Data.(ledger.BlockEvent)
	if !ok {
		return
	}
	switch blockEvt.Action {
	case ledger.BlockActionUndo:
		idx.rollbackBlock(blockEvt.Block)
	case ledger.BlockActionApply:
		block := blockEvt.Block
		decoded, err := block.Decode()
		if err != nil {
			idx.fatal(fmt.Errorf("midnight indexer: decode block slot=%d block=%d: %w", block.Slot, block.Number, err))
			return
		}
		var timestampMs uint64
		if idx.config.SlotTimer != nil {
			if t, err := idx.config.SlotTimer.SlotToTime(block.Slot); err == nil {
				timestampMs = uint64(t.UnixMilli()) //nolint:gosec
			}
		}
		if err := idx.processBlock(block, decoded.Transactions(), timestampMs); err != nil {
			idx.fatal(fmt.Errorf("midnight indexer: process block slot=%d block=%d: %w", block.Slot, block.Number, err))
			return
		}
		if idx.config.BlockIterator != nil {
			if err := idx.updateCheckpoint(block.Slot); err != nil {
				idx.fatal(fmt.Errorf("midnight indexer: checkpoint slot=%d block=%d: %w", block.Slot, block.Number, err))
				return
			}
		}
	}
}

func (idx *Indexer) processBlock(
	block models.Block,
	txs []lcommon.Transaction,
	timestampMs uint64,
) error {
	for i, tx := range txs {
		if err := idx.processTx(block, tx, uint32(i), timestampMs); err != nil { //nolint:gosec
			return err
		}
	}
	return nil
}
```

Then change `processTx` and `processOutput` to return `error`. Each `CreateMidnight*` failure should return instead of logging-and-continuing, for example:

```go
if err := idx.config.Metadata.CreateMidnightAssetSpend(nil, row); err != nil {
	return fmt.Errorf("write asset spend tx=%s input=%s#%d: %w",
		hex.EncodeToString(txHashBytes), inpHashHex, inpIndex, err)
}
idx.mu.Lock()
delete(idx.cNightUTxOs, key)
idx.mu.Unlock()
```

Apply the same pattern for `CreateMidnightDeregistration`, `CreateMidnightAssetCreate`, and `CreateMidnightRegistration`. In `node.go`, wire `FatalErrorFunc` to cancel the node, e.g. `FatalErrorFunc: func(error) { n.cancel() }`. This is intentionally fatal: after one failed block, continuing would allow a later block to checkpoint over the gap.

wolf31o2 (changes requested):
Requesting changes. Ignoring rollback, this still needs a non-lossy apply path. The current implementation can permanently skip Midnight rows because apply delivery is best-effort and the checkpoint can advance after partial/failed indexing. Please make ledger.block delivery ordering-critical and only persist the Midnight checkpoint after the block's index writes have succeeded; on failure, stop/fatal with the checkpoint unchanged so restart can replay from the last known-good slot.

wolf31o2 (approved):

PR #2618, opened 2026-06-23
fix(ledger): mithril chain rewind for catch-up

wolf31o2:
## Summary by cubic
Prevented unwanted primary-chain rewinds during Mithril snapshot catch-up by preserving ahead immutable blocks when the ledger tip is a valid ancestor. This keeps the chain intact so `ledgerProcessBlocks` can replay forward and catch up.

- **Bug Fixes**
  - Updated `reconcilePrimaryChainTipWithLedgerTip` to never rewind when the ledger tip is on the primary chain; removed safety-window logic, computed gap only for logging, and adjusted `primaryChainTipAtOrAheadOfLedgerTip` to intersect regardless of gap.
  - Added `TestReconcilePrimaryChainTipWithLedgerTipCatchesUpWhenAheadBeyondK` to verify the chain tip stays ahead and all ahead blocks are preserved for replay.

<sup>Written for commit 0ba247286675dd5b484c1acc7d870b1632c5d31b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2618?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ledger/chainsync_rollback_test.go
+++ b/ledger/chainsync_rollback_test.go
@@ -1398,7 +1398,13 @@ func TestReconcileLivePrimaryChainLedgerDivergenceExportedWrapperRecoversSubKFor
 	)
 }
 
-func TestReconcilePrimaryChainTipWithLedgerTipRewindsPrimaryChainWhenAheadBeyondK(
+// TestReconcilePrimaryChainTipWithLedgerTipCatchesUpWhenAheadBeyondK covers the
+// old-Mithril-snapshot shape: the immutable primary chain is far ahead of the
+// ledger tip (more than k blocks), but the ledger tip is still a valid ancestor
+// on the primary chain. The primary chain must be preserved so ledgerProcessBlocks
+// can replay forward and catch up; it must NOT be rewound (which would delete the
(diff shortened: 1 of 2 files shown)


arepala-uml (approved):

arepala-uml:
I have one doubt. I just want to understand why we are removing the logic that rewinds the primary chain when the primary chain tip is too far ahead of the ledger tip.

wolf31o2:
When we rewind it, it means "delete the data we have which is newer than this point" but we don't want to do that, since the previous place is an ancestor of the primary chain. In other words, we want to "replay" from the rollback position and catch up, not rewind to it and discard newer data to refetch, which we would do on a chain switch. It's a different point on the same chain vs a fork on one chain to another.

PR #2633, opened 2026-06-24
feat(forging): optionally validate a forged block through chain selection before diffusion

### LINKED ISSUE #2612: forging: optionally validate a forged block through chain selection before diffusion
The Cardano Blueprint forging spec says a forged block should optionally be fed into the node's own chain selection and validation before it is diffused to peers. See https://cardano-scaling.github.io/cardano-blueprint/pr-preview/pr-67/consensus/forging.html

Dingo's forging path (ledger/forging) validates a forged block structurally before adoption: block matches header, prevHash fits the current tip, block-number continuity, per-transaction re-validation against current ledger state, and a CBOR round-trip self-check (builder.go). It does not run full consensus and ledger validation of the assembled block (its own VRF leader proof, KES signature, body hash, and block-level transaction validation) before the block becomes the chain tip and is therefore servable to peers. Full ledger validation only happens afterward, when the ledger consumes the resulting ChainUpdateEvent.

Because the block becomes the tip on adoption, a malformed but structurally fitting block would be adopted and served to peers before the asynchronous ledger validator rejects it.

Required: add an opt-in self-validation step that runs the existing verification primitives (header VRF and KES verification in ledger/verify_header.go, body-hash check, and block-level transaction validation) on the assembled block before AddBlock or before the block is diffused, aborting the forge on failure. The forging spec also requires block creation to be as fast as possible, so this must be measured for latency and gated behind a configuration flag, since the spec marks it optional. Validate with the DevNet harness.

Acceptance criteria:
- a configuration flag enables full self-validation of a forged block before diffusion
- a block failing self-validation is not adopted or diffused
- forge-to-diffusion latency impact measured and acceptable when enabled
- DevNet validation

arepala-uml:
**Summary**

Added an optional self-validation step for forged blocks that runs before the block is adopted into the chain and diffused to peers. When --validate-forged-block is set, each forged block is checked for:

```
Header crypto — VRF/KES signature verification (non-Byron eras).
Body hash guard — rejects non-Byron blocks with an all-zero body hash.
Per-tx ledger rules — validates each transaction with an intra-block UTxO overlay so tx N+1 can spend outputs created by tx N in the same block.
```
The validator runs between BuildBlock and AddBlock. On failure, the block is dropped and logged and the dingo_forge_validation_failed_total counter increments. On success, dingo_forge_validation_duration_seconds is recorded. 
The feature is off by default — no behavior change unless the flag is set.

**Changes**

1. `ledger/validate_forged.go`: Added `ValidateForgedBlock` on `LedgerState`, giving the forger a full self-validation path for newly assembled blocks. It checks header crypto, rejects invalid body-hash state, and validates transactions using an intra-block UTxO overlay.
2. `ledger/forging/forger.go`: Added a `BlockValidator` interface and calls it after `BuildBlock` but before `AddBlock`, so a bad forged block is rejected before it becomes the local chain tip or is broadcast.
3. `ledger/forging/metrics.go`: Added observability for the new path with `dingo_forge_validation_duration_seconds` and `dingo_forge_validation_failed_total`.
4. `node_forging.go`: wires the ledger validator into the forger through forgedBlockValidatorAdapter.
5. `config.go, internal/config/config.go, internal/config/flags.go, internal/node/node.go`: plumbs the new opt-in `--validate-forged-block flag `and `DINGO_VALIDATE_FORGED_BLOCK` env var through config and node startup.
6. `ARCHITECTURE.md`: documents the optional forged-block self-validation step in the block forging flow.

**Tests**

1. `ledger/forging/forger_validator_test.go`: verifies that a passing validator allows adoption, a failing validator drops the block, a nil validator skips validation, and validation happens before broadcast.
2. `ledger/validate_forged_test.go`: covers the validator’s basic behavior: nil block errors, all-zero body hash errors, non-zero body hash passes, and an empty transaction block passes.

Closes #2612 

---
## Summary by cubic
Adds optional self-validation of forged blocks right after `BuildBlock` and before adoption/diffusion; invalid blocks are dropped, and forge latency now includes validation time. Closes #2612.

- **New Features**
  - Added `BlockValidator` to run after `BuildBlock` and before `AddBlock`, so dropped blocks never emit success events.
  - Implemented `LedgerState.ValidateForgedBlock`: VRF/KES header checks (non-Byron), non-zero body hash guard, and per-tx ledger rules with an intra-block UTxO overlay.
  - Metrics: `dingo_forge_validation_duration_seconds`, `dingo_forge_validation_failed_total`.
  - Config: `validateForgedBlock`, CLI `--validate-forged-block`, env `DINGO_VALIDATE_FORGED_BLOCK`.
  - Dev/test: wired `txpump` into DevNet compose build, run as root with `/logs`, added a post-start check in `run-tests.sh`; exposed `shelley-genesis.json` for genesis UTxO discovery; switched `txpump` Dockerfile path.

- **Bug Fixes**
  - Guarded `block.Hash()` logging on both success and failure paths when `BuildBlock` returns nil to avoid a panic.

<sup>Written for commit 718d962bfcc7f0ad6f72212eee0973a93cfa8fe3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2633?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to validate locally forged blocks before they are accepted and shared.
  * Added a new command-line flag and configuration option to enable this behavior.

* **Bug Fixes**
  * Blocks that fail validation are now dropped instead of being adopted or broadcast.
  * Validation checks include block headers, body hash integrity, and transaction rules.

### DIFF
--- a/internal/test/antithesis/Dockerfile.txpump
+++ b/internal/test/antithesis/Dockerfile.txpump
@@ -16,7 +16,6 @@ FROM debian:bookworm-slim
 RUN apt-get update -y && \
     apt-get install -y --no-install-recommends ca-certificates && \
     rm -rf /var/lib/apt/lists/*
-RUN useradd -m txpump
+RUN mkdir -p /logs
 COPY --from=build /code/txpump /bin/txpump
-USER txpump
 ENTRYPOINT ["/bin/txpump"]
--- a/internal/test/devnet/docker-compose.yml
+++ b/internal/test/devnet/docker-compose.yml
@@ -126,8 +126,8 @@ services:
 
   txpump:
     build:
-      context: ../antithesis
-      dockerfile: Dockerfile.txpump
+      context: ../../..
+      dockerfile: internal/test/antithesis/Dockerfile.txpump
     container_name: devnet-txpump
     depends_on:
       dingo-producer:
--- a/internal/test/devnet/docker-compose.yml
+++ b/internal/test/devnet/docker-compose.yml
@@ -63,6 +63,7 @@ services:
       CARDANO_SHELLEY_VRF_KEY: /configs/keys/vrf.skey
       CARDANO_SHELLEY_KES_KEY: /configs/keys/kes.skey
       CARDANO_SHELLEY_OPERATIONAL_CERTIFICATE: /configs/keys/opcert.cert
+      DINGO_VALIDATE_FORGED_BLOCK: "true"
     volumes:
       - p1-configs:/configs:ro
       - ./topology/dingo-producer.json:/topology/topology.json:ro
@@ -129,6 +130,10 @@ services:
       context: ../../..
       dockerfile: internal/test/antithesis/Dockerfile.txpump
     container_name: devnet-txpump
+    init: true
+    # Named Docker volumes are root-owned on first mount; txpump writes
+    # /logs/txpump.log so the runtime user must be root.
+    user: "0:0"
     depends_on:
       dingo-producer:
         condition: service_healthy
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -92,6 +92,9 @@ type BlockForger struct {
 	forgeSyncToleranceSlots     uint64
 	forgeStaleGapThresholdSlots uint64
 
+	// Optional self-validation before adoption (nil = disabled)
+	blockValidator BlockValidator
+
 	// State
 	mu      sync.RWMutex
 	running bool
@@ -127,6 +130,14 @@ type BlockForgedObserver func(
 	cbor []byte,
 	latency time.Duration,
 )
+
+// BlockValidator validates a locally-forged block before it is adopted
+// onto the local chain and diffused to peers. If ValidateForgedBlock
+// returns a non-nil error the block is dropped and neither adopted nor
+// diffused.
+type BlockValidator interface {
+	ValidateForgedBlock(block ledger.Block, blockCbor []byte) error
+}
 
 // LeiosProduceChecker is the forge-loop seam into the Leios pipeline.
 // It reports whether the slot leader may produce an endorser block for
@@ -187,6 +198,12 @@ type ForgerConfig struct {
 	// chain tip is far ahead of the slot clock. Zero uses the default.
 	ForgeStaleGapThresholdSlots uint64
 
+	// BlockValidator, when non-nil, validates the forged block (VRF/KES
+	// header crypto, body-hash consistency, per-tx ledger rules) before
+	// AddBlock is called. A validation failure drops the block without
+	// adopting or diffusing it. Nil disables self-validation (default).
+	BlockValidator BlockValidator
+
 	// Prometheus metrics registry (optional)
 	PromRegistry prometheus.Registerer
 }
@@ -215,6 +232,7 @@ func NewBlockForger(cfg ForgerConfig) (*BlockForger, error) {
 		leiosChecker:     cfg.LeiosProduceChecker,
 		leiosEBCaster:    cfg.LeiosEBBroadcaster,
 		leiosMempool:     cfg.LeiosMempool,
+		blockValidator:   cfg.BlockValidator,
 	}
 	if cfg.ForgeSyncToleranceSlots == 0 {
 		cfg.ForgeSyncToleranceSlots = forgeSyncToleranceSlots
@@ -586,6 +604,40 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 			}()
 			f.blockForged(block, blockCbor, time.Since(forgeStartTime))
 		}()
+	}
+
+	// Optionally self-validate before adoption and diffusion.
+	// This runs VRF/KES header crypto, body-hash consistency, and
+	// per-tx ledger rules against the current chain state. A failure
+	// means the locally-built block is invalid; drop it without
+	// adopting or diffusing so we never serve a bad block to peers.
+	if f.blockValidator != nil {
+		validateStart := time.Now()
+		if err := f.blockValidator.ValidateForgedBlock(block, blockCbor); err != nil {
+			f.incCouldNotForge()
+			if f.metrics != nil {
+				f.metrics.forgeValidationFailed.Inc()
+			}
+			f.logger.Error(
+				"forged block failed self-validation, dropping block",
+				"slot", currentSlot,
+				"hash", hex.EncodeToString(block.Hash().Bytes()),
+				"error", err,
+			)
+			return fmt.Errorf("forged block self-validation failed: %w", err)
+		}
+		validationDuration := time.Since(validateStart)
+		if f.metrics != nil {
+			f.metrics.forgeValidationDuration.Observe(
+				validationDuration.Seconds(),
+			)
+		}
+		f.logger.Info(
+			"forged block passed self-validation",
+			"slot", currentSlot,
+			"hash", hex.EncodeToString(block.Hash().Bytes()),
+			"validation_duration", validationDuration,
+		)
 	}
 
 	// Add block to chain and broadcast
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -587,19 +587,19 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 	// that is ultimately dropped.
 	if f.blockValidator != nil {
 		validateStart := time.Now()
+		blockHashStr := ""
+		if block != nil {
+			blockHashStr = hex.EncodeToString(block.Hash().Bytes())
+		}
 		if err := f.blockValidator.ValidateForgedBlock(block, blockCbor); err != nil {
 			f.incCouldNotForge()
 			if f.metrics != nil {
 				f.metrics.forgeValidationFailed.Inc()
 			}
-			hashStr := ""
-			if block != nil {
-				hashStr = hex.EncodeToString(block.Hash().Bytes())
-			}
 			f.logger.Error(
 				"forged block failed self-validation, dropping block",
 				"slot", currentSlot,
-				"hash", hashStr,
+				"hash", blockHashStr,
 				"error", err,
 			)
 			return fmt.Errorf("forged block self-validation failed: %w", err)
@@ -613,7 +613,7 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 		f.logger.Info(
 			"forged block passed self-validation",
 			"slot", currentSlot,
-			"hash", hex.EncodeToString(block.Hash().Bytes()),
+			"hash", blockHashStr,
 			"validation_duration", validationDuration,
 		)
 	}
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -591,7 +591,12 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 		if block != nil {
 			blockHashStr = hex.EncodeToString(block.Hash().Bytes())
 		}
-		if err := f.blockValidator.ValidateForgedBlock(block, blockCbor); err != nil {
+		validationErr := f.blockValidator.ValidateForgedBlock(block, blockCbor)
+		validationDuration := time.Since(validateStart)
+		if f.metrics != nil {
+			f.metrics.forgeValidationDuration.Observe(validationDuration.Seconds())
+		}
+		if validationErr != nil {
 			f.incCouldNotForge()
 			if f.metrics != nil {
 				f.metrics.forgeValidationFailed.Inc()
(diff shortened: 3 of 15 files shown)


arepala-uml:
DevNet bug fixes - I have identified below issue when I was running devnet test suite freshly `./internal/test/devnet/run-tests.sh --keep-up` with the changes

1. **txpump build context wrong** - 

`docker-compose.yml` set context: `../antithesis `for the txpump build. That resolves to `internal/test/antithesis/,` but `Dockerfile.txpump `uses `COPY . .` expecting the repo root as the build context. So, Docker couldn't find the files and failed with below error

Screenshot 2026-06-24 at 5 07 56 PM

As a fix, I have changed context to `../../.. (repo root) `and moved the dockerfile path accordingly.

2. **txpump permission errors (logs + socket)** - txpump ran as a non-root txpump user and hit two permission failures in sequence:

`First failure — /logs write permission:` - The /logs directory in the image was either missing or owned by root. The txpump user had no write access, so txpump exited immediately before even trying to connect to the node.

Screenshot 2026-06-24 at 5 15 19 PM

`Second failure — dingo socket write permission` - dial unix /ipc/dingo.socket: connect: permission denied
After creating `/logs `with correct ownership, txpump got past logging but hit the next wall. The dingo socket has permissions `srwxr-xr-x` (0755). Unix `connect()` requires write permission on the socket file. The txpump user falls in the "other" bucket (r-x only), so every connection attempt failed.

As a fix, I have removed the non-root user entirely and made txpump to run as root. So, both failures disappeared and root can write to /logs and root can connect to the socket. 

3. **txpump no genesis UTxOs found** 

Issue: Two problems stacked on top of each other:

- configurator.sh copied utxo files to `/configs` (the p1-configs volume) instead of `/configs/utxo-keys` (the `utxo-keys` volume that txpump mounts). The volume txpump reads was always empty.
- Even after fixing the directory, the script was copying raw key files (`.skey, .vkey, .addr.info`). txpump's `LoadGenesisUTxOs` only reads `.json` files — it silently skips everything else. Still zero UTxOs found.

Screenshot 2026-06-24 at 5 22 56 PM

Fix: After pool configs are generated, copy the Shelley genesis (which has `initialFunds` — exactly what txpump parses) into the utxo-keys volume:

`cp /configs/1/configs/shelley-genesis.json /configs/utxo-keys/ `and txpump finds it, parses initialFunds, seeds its wallet with 4 UTxOs, and starts running. The reason behind of not noticing was the two issues were always killed txpump before it reached this point. Additionally, the existing devnet integration tests only assert consensus between dingo and cardano-node and there is no txpump's health check, so tests passed even with txpump broken.

arepala-uml:
**Testing part**

I have ran the full unit test suite with the race detector and all 4 forger validator tests, 4 validate-forged tests passed.

```
akrepala@akrepalas-MacBook-Pro dingo % go test ./ledger/... ./ledger/forging/...
ok      github.com/blinklabs-io/dingo/ledger    (cached)
ok      github.com/blinklabs-io/dingo/ledger/eras       (cached)
ok      github.com/blinklabs-io/dingo/ledger/forging    (cached)
ok      github.com/blinklabs-io/dingo/ledger/governance (cached)
ok      github.com/blinklabs-io/dingo/ledger/hardfork   (cached)
ok      github.com/blinklabs-io/dingo/ledger/leader     (cached)
ok      github.com/blinklabs-io/dingo/ledger/leios      (cached)
ok      github.com/blinklabs-io/dingo/ledger/snapshot   (cached)
```


**DevNet Pass 1 — Flag Off (default behaviour unchanged)**

I have brought up the full 4-container devnet (configurator, dingo-producer, cardano-producer, cardano-relay + txpump) without the flag set and ran the complete test suite including the epoch boundary scenario (~16 minutes) and I have confirmed that there are **zero self-validation log lines in dingo-producer.**

Ran the test suite using `./internal/test/devnet/run-tests.sh --keep-up` and confirmed there is no self valiudation of forged blocks in default approach.

```
akrepala@akrepalas-MacBook-Pro dingo % docker compose -f internal/test/devnet/docker-compose.yml logs dingo-producer | grep -c "self-valid"
0
akrepala@akrepalas-MacBook-Pro dingo % docker compose -f internal/test/devnet/docker-compose.yml logs dingo-producer | grep -c "self-valid"
0

```
[devnet_default.log](https://github.com/user-attachments/files/29348778/devnet_default.log)


**DevNet Pass 2 — Added DINGO_VALIDATE_FORGED_BLOCK=true flag in docker-compose.yml for dingo-producer, tore down volumes, and ran the full test suite again using `./internal/test/devnet/run-tests.sh --keep-up` by bringing down previous all previous containers.** 


I have confirmed self-validation ran on every forged block during the execution of test suite

```
akrepala@akrepalas-MacBook-Pro dingo % docker compose -f internal/test/devnet/docker-compose.yml logs dingo-producer | grep -c "self-valid"
93
akrepala@akrepalas-MacBook-Pro dingo % docker compose -f internal/test/devnet/docker-compose.yml logs dingo-producer | grep -c "self-valid"
98
akrepala@akrepalas-MacBook-Pro dingo % docker compose -f internal/test/devnet/docker-compose.yml logs dingo-producer | grep -c "self-valid"
168
akrepala@akrepalas-MacBook-Pro dingo % docker compose -f internal/test/devnet/docker-compose.yml logs dingo-producer | grep -c "self-valid"
171
```

Also, the added prometheus metrics were also populated and forge validation metric values was zero 

Screenshot 2026-06-24 at 5 43 22 PM

The sample log line from dingo-producer about validation of forged block
```
akrepala@akrepalas-MacBook-Pro dingo % docker compose -f internal/test/devnet/docker-compose.yml logs dingo-producer | grep "self-valid"

dingo-producer  | time=2026-06-24T21:25:03.004Z level=INFO source=/code/ledger/forging/forger.go:635 msg="forged block passed self-validation" slot=7 hash=da259a28ee702a28351029cadf80d51d0b7dc528e35ee51f189f1759daedcfc4 validation_duration=276.75µs
dingo-producer  | time=2026-06-24T21:25:04.001Z level=INFO source=/code/ledger/forging/forger.go:635 msg="forged block passed self-validation" slot=8 hash=7aaf46eb09c9a38698d4207bc0f7f1c4fee9d653dbc91911bfe55539a4d5bb57 validation_duration=265.708µs
dingo-producer  | time=2026-06-24T21:25:10.002Z level=INFO source=/code/ledger/forging/forger.go:635 msg="forged block passed self-validation" slot=14 hash=0c7fb9800ef9e02d75bd4c8cef8b10d4e6bdb1b7c1079d80e511a6e646793c00 validation_duration=258.667µs
dingo-producer  | time=2026-06-24T21:25:12.002Z level=INFO source=/code/ledger/forging/forger.go:635 msg="forged block passed self-validation" slot=16 hash=d2e68cba20966fbc13f80467fe05abbd3be2e1dd5703c629fd376b0c80e6cdb1 validation_duration=303.792µs
dingo-producer  | time=2026-06-24T21:25:13.000Z level=INFO source=/code/ledger/forging/forger.go:635 msg="forged block passed self-validation" slot=17 hash=9ccc31beed35e6797ce6380776432328cfd766c02c7567a1071ba1bb45666e9b validation_duration=261.958µs
```
[devnet_blockValidate.log](https://github.com/user-attachments/files/29348800/devnet_blockValidate.log)

ledger/validate_forged.go
```diff
@@ -0,0 +1,151 @@
...
+			createdUtxos[key] = utxo
+		}
+		for _, input := range tx.Inputs() {
+			key := fmt.Sprintf(
+				"%s:%d",
+				input.Id().String(),
+				input.Index(),
+			)
+			consumedUtxos[key] = struct{}{}
+		}
```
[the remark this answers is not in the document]
arepala-uml: I have fixed it and please review it.

arepala-uml:
@cubic-dev-ai

arepala-uml:
@cubic-dev-ai Please review all changed files once

arepala-uml:
@cubic-dev-ai Please review all changed files one last time

arepala-uml:
@cubic-dev-ai

ledger/forging/forger.go
```diff
@@ -588,6 +606,40 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
...
+			)
+		}
+		f.logger.Info(
+			"forged block passed self-validation",
+			"slot", currentSlot,
+			"hash", hex.EncodeToString(block.Hash().Bytes()),
+			"validation_duration", validationDuration,
+		)
+	}
+
```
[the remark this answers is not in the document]
arepala-uml: I have moved it and it will run immediately after BuildBlock. Please review it.

internal/test/devnet/docker-compose.yml
```diff
@@ -126,8 +126,8 @@ services:
 
   txpump:
     build:
-      context: ../antithesis
-      dockerfile: Dockerfile.txpump
+      context: ../../..
+      dockerfile: internal/test/antithesis/Dockerfile.txpump
```
wolf31o2, line 131: Blocking: this PR changes the DevNet txpump image definition, but internal/test/devnet/run-tests.sh still builds only configurator and dingo-producer. Docker Compose can reuse an existing devnet-txpump image, so the required DevNet run can pass without exercising this PR's txpump image. I hit exactly that locally: txpump exited with permission denied opening /logs/txpump.log while the Go tests still printed PASS. Please build txpump explicitly in the runner and fail if it is not running before tests.\n\nSuggested runner change:\n\n```diff\n-log "Building Dingo Docker image..."\n-docker compose -f "" build configurator dingo-producer\n+log "Building DevNet Docker images..."\n+docker compose -f "" build configurator dingo-producer txpump\n```\n\nAlso add a post-up check for txpump, for example inspecting `docker compose -f "" ps --status running txpump` and failing if it is empty.

internal/test/antithesis/Dockerfile.txpump
```diff
@@ -16,7 +16,6 @@ FROM debian:bookworm-slim
 RUN apt-get update -y && \
     apt-get install -y --no-install-recommends ca-certificates && \
     rm -rf /var/lib/apt/lists/*
-RUN useradd -m txpump
+RUN mkdir -p /logs
```
wolf31o2, line 19: This directory is hidden by the named volume mounted at /logs in DevNet, so creating it in the image does not guarantee that txpump can write /logs/txpump.log at runtime. The antithesis compose files already handle the root-owned named volume explicitly with `init: true` and `user: "0:0"` on txpump. Please mirror that in the DevNet service so the runtime behavior is deterministic.\n\nSuggested compose change:\n\n```yaml\n  txpump:\n    build:\n      context: ../../..\n      dockerfile: internal/test/antithesis/Dockerfile.txpump\n    container_name: devnet-txpump\n    init: true\n    user: "0:0"\n```

wolf31o2 (changes requested):
No-go as-is. Local DevNet no longer reliably tests the changed txpump path. In my run the nodes became healthy and the Go devnet tests printed PASS, but txpump had already exited with permission denied opening /logs/txpump.log, so there was no mempool traffic. Please make the DevNet runner rebuild/check txpump and make the txpump runtime user explicit, then rerun DevNet.

internal/test/devnet/docker-compose.yml
arepala-uml: Yes, you are correct and run-tests.sh is not rebuilding the txpump image where I have built manually everytime(`docker compose -f internal/test/devnet/docker-compose.yml build txpump`) after making changes before running of devnet test suite locally. I have missed it, I will make changes & re-test it whether it is building the txpump image everytime or not and will include post txpump check as well.

internal/test/antithesis/Dockerfile.txpump
arepala-uml: I have added `init: true + user: "0:0" `to txpump, mirroring the antithesis compose. The named volume is root-owned on first mount; running as root explicitly makes log writes deterministic.

internal/test/devnet/docker-compose.yml
arepala-uml: I have added txpump to the docker compose build line so this PR image change is always rebuilt rather than reusing a stale cached image. Also, I have added a post-up check after nodes are healthy, we verify txpump is in running state or not. If it has exited like permission denied error, the script prints txpump logs and fails instead of silently running tests with no mempool traffic.

arepala-uml:
After adding both above txpump fixes, I have re-ran devnet test suite with enabling of DINGO_VALIDATE_FORGED_BLOCK and default case as well.

Everytime, we are building the txpump image along with dingo-producer and included a post txpump check whether it is running or not before starting devnet integration tests as we can see in below image. Also, I have updated logs above devnet_default.log & devnet_blockValidate.log as well.

Screenshot 2026-06-25 at 11 38 26 AM

Screenshot 2026-06-25 at 11 53 01 AM

wolf31o2 (approved):
The txpump problem was preexisting and uncovered by fixes added in this branch. A separate issue has been filed to track it.

PR #2681, opened 2026-06-29
feat(ledger): validate operational certificates on inbound blocks

### LINKED ISSUE #2608: ledger: validate operational certificates on inbound blocks
The Cardano Blueprint chain validity spec requires header validation to include operational certificate (opcert) validation. See https://cardano-scaling.github.io/cardano-blueprint/pr-preview/pr-67/consensus/chainvalid.html

Dingo verifies VRF proofs and KES signatures on inbound headers (ledger/verify_header.go calls gouroboros ledger.VerifyBlock), but it passes SkipStakePoolValidation true and never validates the operational certificate. The opcert sequence number is only recorded (UpdatePoolOpCertSequence) and consulted by the forging path, not enforced on inbound blocks. Operational certificate validation is fully implemented for the forging side (ledger/forging/keys.go) but not for inbound verification.

Missing inbound opcert checks:

1. Cold-key signature. The operational certificate is signed by the pool's registered cold key over the hot KES vkey, KES period, and counter. gouroboros provides ledger.VerifyOpCertSignature. Dingo does not resolve the producer pool's registered cold vkey and verify this signature on inbound blocks.

2. Counter monotonicity. The opcert counter must equal the pool's last seen counter or be exactly one greater. A backward or skipped counter indicates a stale or stolen hot key and must be rejected. Dingo stores the latest counter (LatestPoolOpCertSequence) but does not compare and reject on the inbound path; it advances the stored counter unconditionally.

3. KES evolution expiry. ledger.ValidateKesPeriod against the protocol parameter maxKESEvolutions enforces that the certificate has not expired. Dingo only performs the lighter future-cert KES guard inside VerifyKesComponents on inbound blocks.

Scope notes: this is cross-cutting. It needs the pool cold-vkey lookup from ledger state during header validation, a read-before-write of the stored counter inside the validation transaction, and correct rollback semantics for the counter store, since the counter must not advance for a block that later gets rolled back. Counter state interacts with chain selection and rollbacks, so design care is required.

This is consensus-critical. It must be validated against the DevNet harness and conformance suite, since an incorrect counter or signature rule would reject valid blocks and stall the chain, while a missing rule lets a peer present a header for a pool whose hot key was rotated.

Acceptance criteria:
- inbound blocks reject an invalid opcert cold-key signature
- inbound blocks reject a backward or gapped opcert counter, with the stored counter advancing only on accepted, non-rolled-back blocks
- expired KES periods rejected per maxKESEvolutions
- DevNet and conformance validation

chrisguiney:
Closes #2608

## Problem

Inbound blocks had VRF and KES signatures verified, but the operational certificate was never validated — `VerifyBlock` ran with `SkipStakePoolValidation`, and the opcert sequence number was only recorded, not enforced. A peer could present a header for a pool whose hot key had been rotated.

## Change

Adds the three opcert checks, split by data dependency:

- **Cold-key signature + KES-period expiry — at header verification** (`ledger/verify_opcert.go`, called from `verifyBlockHeaderCrypto`). Stateless checks over the header and Shelley-genesis params, so forged/expired opcerts are rejected before the block body is fetched. `opCertFromHeader` extracts the opcert across the Shelley- and Babbage-family header layouts.
- **Counter monotonicity — at block apply** (`validateOpCertCounter`, in `ledgerProcessBlock` under `shouldValidate`). Read-before-write of the stored per-pool counter inside the validation transaction: the counter must equal the last-seen value or be exactly one greater; backward (stale/stolen hot key) and gapped (skipped rotation) counters are rejected. A pool with no recorded counter has no baseline and is accepted. Rollback safety is inherited from the existing per-`(pool, slot)` `pool_opcert_sequence` store — **no storage/schema change**.

### Note on the cold-key signature representation

The cold signature is verified over the raw cardano-ledger `OCertSignable` representation (`kesVkey || counter || period`), **not** gouroboros' `ledger.VerifyOpCertSignature`, which hashes a CBOR array that does not match real blocks (it would reject every block). This was confirmed against a real cardano-cli opcert and is pinned by a known-answer test. The upstream gouroboros bug is fixed separately in blinklabs-io/gouroboros#1843; this PR does not depend on it (the local check is self-contained, defense-in-depth).

## Validation

- Unit tests (`ledger/verify_opcert_test.go`): extractor, valid/tampered cold-sig, expired/`maxKES=0` KES, counter table (first-sight/equal/+1/backward/gap), Byron skip, and a **known-answer test against a real cardano-cli opcert**.
- Full `ledger` suite passes with `-race`; conformance suite passes.
- **DevNet harness**: `dingo-producer` reaches chain consensus with the official `cardano-node` producer (real opcerts) — confirms the inbound check accepts real opcerts end-to-end.
- `golangci-lint run ./...`: 0 issues; `modernize` clean; `make import-boundaries` passes.

## Docs

- **ARCHITECTURE.md**: new "Operational Certificate Validation" section (both checkpoints + rollback semantics).
- **DATABASE.md**: `pool_opcert_sequence` description updated to note its read-before-write role in inbound validation.

---
## Summary by cubic
Validates operational certificates on inbound blocks to reject forged or expired headers early and enforce counter rules with a Praos‑only no‑gap check. The counter check runs before tx validation; no schema changes.

- **New Features**
  - Header: verify cold-key signature against issuer vkey and KES expiry using Shelley genesis `maxKESEvolutions`; skip expiry if params are missing; `opCertFromHeader` supports Shelley→Dijkstra.
  - Block apply: enforce counter monotonicity in all eras; reject gaps (>+1) only in Praos (Babbage+) via `opCertNoGapRuleApplies`; first sighting accepted; read-before-write before endorser/tx/Plutus; record after txs; rollback-safe via `pool_opcert_sequence`.
  - Crypto/Tests/Docs: verify over raw OCertSignable bytes (KES vkey || counter || period), not `gouroboros`’ CBOR hash; pinned by a known-answer test with a real `cardano-cli` opcert. Unit tests cover extractor, tampered sig, KES expiry, and era-scoped gap rules; conformance and DevNet pass. Docs name `verifyOpCertColdSignature` and note the counter read-before-write.

<sup>Written for commit 7bf6cc5a0ef22cf246b4f0a81b56a96e6f73a0e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2681?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for validating operational certificates during block processing.
  * Improved handling of certificate signature checks and expiration rules.

* **Bug Fixes**
  * Strengthened rejection of stale or skipped certificate sequence updates.
  * Improved rollback-safe handling so certificate state stays consistent after chain reorgs.

* **Documentation**
  * Clarified how certificate validation works in architecture and database docs.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1030,8 +1030,33 @@ When a network config supplies a `CheckpointsFile` (mainnet and preview ship one
 `ledger/verify_header.go` performs cryptographic validation of block headers:
 - VRF proof verification against the epoch nonce
 - KES signature verification with period checks
-- Operational certificate chain validation
 - Slot leader eligibility checking
+
+### Operational Certificate Validation
+
+Inbound operational-certificate (opcert) validation is split by its data
+dependency across two points in the pipeline:
+
+- **Stateless checks at header verification** (`ledger/verify_opcert.go`,
+  invoked from `verifyBlockHeaderCrypto`): the cold-key signature and the KES
+  period expiry. The cold verification key is the header's issuer vkey — a
+  registered pool's cold key is, by construction, the vkey whose Blake2b224
+  hash is its pool id. `opCertFromHeader` extracts the opcert across the
+  Shelley- and Babbage-family header layouts; `gouroboros`'
+  `ledger.VerifyOpCertSignature` and `ledger.ValidateKesPeriod`
+  (against `maxKESEvolutions` from Shelley genesis) perform the checks. Running
+  here rejects forged or expired opcerts before the block body is fetched.
+  These checks share the existing skip-during-historical-sync gating.
+- **Counter monotonicity at block apply** (`validateOpCertCounter`, invoked
+  from `ledgerProcessBlock` under `shouldValidate`): a read-before-write of the
+  pool's stored opcert counter inside the validation transaction. The counter
+  must equal the last-seen value or be exactly one greater; backward counters
+  (stale/stolen hot key) and gapped counters (skipped rotation) are rejected. A
+  pool with no recorded counter has no baseline (genuine first sighting, or a
+  Mithril-restored start) and is accepted as the baseline. Rollback safety is
+  inherited from the per-`(pool, slot)` `PoolOpCertSequence` store, which drops
+  rows past the rollback slot and recomputes the latest counter, so the counter
+  never advances for a block that is later rolled back.
 
 ### Epoch Nonce Computation
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1042,9 +1042,11 @@ dependency across two points in the pipeline:
   period expiry. The cold verification key is the header's issuer vkey — a
   registered pool's cold key is, by construction, the vkey whose Blake2b224
   hash is its pool id. `opCertFromHeader` extracts the opcert across the
-  Shelley- and Babbage-family header layouts; `gouroboros`'
-  `ledger.VerifyOpCertSignature` and `ledger.ValidateKesPeriod`
-  (against `maxKESEvolutions` from Shelley genesis) perform the checks. Running
+  Shelley- and Babbage-family header layouts; `verifyOpCertColdSignature`
+  verifies the cold-key signature over the raw cardano-ledger `OCertSignable`
+  bytes (not `gouroboros`' `ledger.VerifyOpCertSignature`, which hashes a CBOR
+  array that does not match real opcerts), and `ledger.ValidateKesPeriod`
+  (against `maxKESEvolutions` from Shelley genesis) checks expiry. Running
   here rejects forged or expired opcerts before the block body is fetched.
   These checks share the existing skip-during-historical-sync gating.
 - **Counter monotonicity at block apply** (`validateOpCertCounter`, invoked
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1050,15 +1050,20 @@ dependency across two points in the pipeline:
   here rejects forged or expired opcerts before the block body is fetched.
   These checks share the existing skip-during-historical-sync gating.
 - **Counter monotonicity at block apply** (`validateOpCertCounter`, invoked
-  from `ledgerProcessBlock` under `shouldValidate`): a read-before-write of the
-  pool's stored opcert counter inside the validation transaction. The counter
-  must equal the last-seen value or be exactly one greater; backward counters
-  (stale/stolen hot key) and gapped counters (skipped rotation) are rejected. A
-  pool with no recorded counter has no baseline (genuine first sighting, or a
-  Mithril-restored start) and is accepted as the baseline. Rollback safety is
-  inherited from the per-`(pool, slot)` `PoolOpCertSequence` store, which drops
-  rows past the rollback slot and recomputes the latest counter, so the counter
-  never advances for a block that is later rolled back.
+  from `ledgerProcessBlock` under `shouldValidate`, before the block's
+  transactions are validated): a read-before-write of the pool's stored opcert
+  counter inside the validation transaction. A backward counter (below the last
+  seen — stale/stolen hot key) is rejected in every era. A gapped counter (more
+  than one past the last seen) is the Praos over-increment case and is rejected
+  only for Praos eras (Babbage onward, via `opCertNoGapRuleApplies`); TPraos
+  eras (Shelley–Alonzo) enforce only monotonicity, so the gap rule is scoped by
+  era rather than by validation mode (`shouldValidate` can be true for
+  historical or near-tip TPraos blocks). A pool with no recorded counter has no
+  baseline (genuine first sighting, or a Mithril-restored start) and is accepted
+  as the baseline. Rollback safety is inherited from the per-`(pool, slot)`
+  `PoolOpCertSequence` store, which drops rows past the rollback slot and
+  recomputes the latest counter, so the counter never advances for a block that
+  is later rolled back.
 
 ### Epoch Nonce Computation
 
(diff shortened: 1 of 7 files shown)


ledger/state.go
```diff
@@ -3848,12 +3844,49 @@ func (ls *LedgerState) ledgerProcessBlock(
...
+		// block's, inside the same validation transaction. The opcert counter
+		// must equal the last-seen counter or be exactly one greater. A
+		// backward counter signals a stale or stolen hot key; a gapped counter
+		// signals a skipped rotation. Both must be rejected. Only enforced
+		// under shouldValidate (live, current-era blocks): historical blocks
+		// were already network-validated, and the over-increment rule postdates
+		// the early eras replayed during sync. Rollback safety is inherited
+		// from the per-(pool,slot) PoolOpCertSequence store, which drops rows
+		// past the rollback slot and recomputes the latest counter.
+		if shouldValidate {
```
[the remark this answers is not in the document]
chrisguiney: Addressed in 2cc1871b — the opcert counter monotonicity check now runs before endorser-block and transaction/Plutus processing in `ledgerProcessBlock`, so a stale or gapped opcert is rejected by this cheap stateful read before the expensive validation runs. The counter is still recorded after the block's transactions are processed, so read-before-write and rollback semantics are unchanged.

ARCHITECTURE.md
```diff
@@ -958,9 +958,34 @@ When a network config supplies a `CheckpointsFile` (mainnet and preview ship one
...
+
+- **Stateless checks at header verification** (`ledger/verify_opcert.go`,
+  invoked from `verifyBlockHeaderCrypto`): the cold-key signature and the KES
+  period expiry. The cold verification key is the header's issuer vkey — a
+  registered pool's cold key is, by construction, the vkey whose Blake2b224
+  hash is its pool id. `opCertFromHeader` extracts the opcert across the
+  Shelley- and Babbage-family header layouts; `gouroboros`'
+  `ledger.VerifyOpCertSignature` and `ledger.ValidateKesPeriod`
+  (against `maxKESEvolutions` from Shelley genesis) perform the checks. Running
+  here rejects forged or expired opcerts before the block body is fetched.
```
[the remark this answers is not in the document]
chrisguiney: Good catch — fixed in 1fe8ff97. The doc now names `verifyOpCertColdSignature` (raw cardano-ledger `OCertSignable`) and adds a parenthetical noting it is deliberately *not* `gouroboros`' `ledger.VerifyOpCertSignature`, which hashes a CBOR array that doesn't match real opcerts, so future refactors aren't pointed back at the broken API.

ledger/state.go
```diff
@@ -3572,6 +3568,48 @@ func (ls *LedgerState) ledgerProcessBlock(
...
+				txn,
+			)
+			if err != nil {
+				return nil, fmt.Errorf(
+					"read opcert counter for pool %x: %w",
+					opCertPoolKeyHash,
+					err,
+				)
+			}
+			if err := validateOpCertCounter(
```
wolf31o2, line 3694: Please make this counter check era-aware before applying it here. `shouldValidate` is not equivalent to Praos-only: it can be true when `ValidateHistorical` is enabled and in near-tip/live TPraos-era devnets. TPraos allows any opcert counter `candidate >= stored`, while Praos adds the no-gap rule. With the current call, a valid Shelley/Allegra/Mary/Alonzo block whose counter jumps can be rejected.

ledger/verify_opcert.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil
+	}
+	switch {
+	case candidate < stored:
+		return fmt.Errorf(
+			"opcert counter %d is below last seen %d (stale or stolen hot key)",
+			candidate,
+			stored,
+		)
+	case candidate > stored+1:
```
wolf31o2: This `stored+1` upper bound is the Praos over-increment rule, but this helper has no era or consensus-family input and is used for all opcert-bearing eras. Please either pass the era/consensus family in, or split the predicate so TPraos only rejects `candidate < stored` while Praos rejects both stale and gapped counters.

wolf31o2 (changes requested):
Requesting changes: the opcert counter validation needs to be scoped by consensus family. The no-gap `stored+1` rule is valid for Praos-era blocks, but TPraos-era blocks only require monotonicity. As written, the new check is controlled by `shouldValidate`, not by era, so it can still reject valid TPraos blocks in supported validation paths.

ledger/state.go
chrisguiney: Fixed in c021e0a6. The check is now era-aware: the call passes `opCertNoGapRuleApplies(block.Era().Id)` (true from Babbage onward), so the no-gap rule only applies to Praos eras. `shouldValidate` no longer stands in for Praos — the backward/stale check (`candidate < stored`) still runs for every validated block, but a valid TPraos (Shelley–Alonzo) block may now advance its opcert counter by more than one.

ledger/verify_opcert.go
chrisguiney: Fixed in c021e0a6, taking the split-predicate option: `validateOpCertCounter` now takes an `enforceNoGap bool`. `candidate < stored` (stale/stolen) is rejected in every era; `candidate > stored+1` (over-increment) is rejected only when `enforceNoGap` is set. The caller derives it from the block's era via `opCertNoGapRuleApplies` (Praos = Babbage+). Added unit tests for the TPraos-accepts-gap vs Praos-rejects-gap split and the Alonzo→Babbage boundary.

wolf31o2 (approved):

PR #2684, opened 2026-06-30
ci: Koios parity checker initial design

### LINKED ISSUE #2679: Architecture: Koios-parity full-epoch reward parity tracker
## koios-parity: validate Dingo reward state against Koios reference data

Operator tool that validates Dingo's closed-epoch reward computations against
Koios on preview and preprod networks. Compares every pool's reward inputs and
epoch-level aggregates from epoch 0 to tip, flagging any discrepancy so we can
file targeted tickets.

### Architecture

**Reference source:** Koios v1 REST API (fetched and cached in a local SQLite
`cache.db` via `koios-parity fetch`). Uses `/epoch_info` for epoch aggregates,
`/pool_registrations` (dedup'd) for the authoritative pool list including retired
pools, and `/pool_history` per pool for per-epoch stake/block/delegator data.

**Dingo source:** Direct read of Dingo's metadata database — no Blockfrost or
any other HTTP endpoint is contacted for the comparison. Three backends are
supported, matching whichever backend Dingo itself was configured with:

- **sqlite** (default): opens `{data-dir}/metadata.sqlite` in read-only WAL
  mode. Safe to run alongside a live node; SQLite WAL allows concurrent readers
  during active writes.
- **postgres**: connects via a libpq-style DSN (`--metadata-dsn`).
- **mysql**: connects via a go-sql-driver DSN (`--metadata-dsn`).

Tables read: `reward_pool_input` (delegated_stake, blocks_produced,
delegator_count), `epoch_summary` (total_active_stake, pool_count,
delegator_count), `reward_ada_pots` (fees).

### What is compared

Per epoch (Koios `/epoch_info` vs Dingo `epoch_summary` + `reward_ada_pots`):
- `total_active_stake`
- `pool_count`
- `delegator_count`
- `fees`

Per pool per epoch (Koios `/pool_history` vs Dingo `reward_pool_input`):
- `delegated_stake` (= active stake used in reward calculation)
- `delegator_count`
- `blocks_produced`

Pool membership:
- Pool in Koios but absent from Dingo `reward_pool_input` → `pool_only_koios` (FAIL)
- Pool in Dingo `reward_pool_input` but not in Koios → `pool_only_dingo` (FAIL)

### Mismatch categories
| Category | Status | Meaning |
|---|---|---|
| `value_mismatch` | FAIL | Field values differ |
| `pool_only_dingo` | FAIL | Pool computed by Dingo but not in Koios |
| `pool_only_koios` | FAIL | Pool in Koios but missing from Dingo reward_pool_input |
| `dingo_db_error` | ERROR | DB query failed |
| `dingo_db_missing` | ERROR | Expected row absent (epoch not yet computed) |
| `reference_lag` | ERROR | Koios data may be incomplete for a recent epoch |

### CLI

```
koios-parity [--dingo-data .dingo] [--network preview|preprod] [--cache path]
[--metadata-plugin sqlite|postgres|mysql] [--metadata-dsn <dsn>]
```

Subcommands:

fetch    Pull Koios reference data into cache.db
check    Compare cache against Dingo's metadata database
watch    Long-running loop: fetch+check when epoch advances
explain  Drill into mismatches for a specific epoch
status   Show cache coverage and PASS/FAIL/ERROR counts

Environment:

DINGO_DATA_DIR          Dingo node data directory (default: .dingo)
CARDANO_NETWORK         preview or preprod
KOIOS_API_KEY           Koios Bearer token for rate-limited access
DINGO_METADATA_PLUGIN   sqlite (default), postgres, or mysql
DINGO_METADATA_DSN      connection string for postgres/mysql

### Implementation location
`cmd/koios-parity/` (binary) + `internal/koiosparity/` (library)

### Future work
- Per-account reward check (`account_reward_delta` vs Koios `/account_rewards`)
  to cover every stake address for every epoch

cryptodj413:
Closes #2679 

---
## Summary by cubic
Introduces `koios-parity`, a standalone CLI that validates Dingo’s closed-epoch reward inputs and totals against Koios on preview and preprod. It reads Dingo’s metadata DB directly, caches Koios locally (incl. retired pools), supports resume/interrupt, persists check outcomes, and writes JSON reports with progress.

- **New Features**
  - Incremental Koios fetch for closed epochs with parallel workers, hole-filling, `--force-refresh`, and cache at `.koios/cache.db`.
  - Direct comparison of Koios vs Dingo’s `epoch_summary`, `reward_pool_input`, `reward_ada_pots`, plus `/totals`; mismatch categories include `value_mismatch`, `pool_only_dingo`, `pool_only_koios`, `dingo_db_missing`, `dingo_db_error`, `reference_lag` via `--grace-hours`.
  - Commands: `run`, `fetch`, `check`, `status`, `explain`, `watch`. `run` writes `report-<network>-<date>.json`, exits non-zero on failures, and persists outcomes for reuse.

- **Bug Fixes**
  - Epoch alignment: compare Koios reporting epoch K against Dingo’s stake epoch K-1 and param epoch K+1; keep `reward_ada_pots` at K.
  - Presence tracking: added `StakePresent`, `ParamsPresent`, `MemberRewardPresent` to classify missing rows as `reference_lag` or `dingo_db_missing` instead of false `value_mismatch`.
  - Treat empty Dingo `margin` like `fixed_cost` and flag `value_mismatch` when corrupted.
  - Koios client/cache resilience: fix `/pool_history`, include retired pools via paginated `/pool_list`, handle nullable fields, read/close bodies, separate 5xx/429 retry backoff, WAL + busy timeouts, drop stale rows on refresh.
  - Correctness and status: use `/totals` for aggregates, classify pre-staking epochs correctly, gate checks on `snapshot_ready`, and fix pass/fail/error counts with persisted outcomes.
  - DSN defaults: auto-build DSNs for postgres/mysql providers when config omits them, matching Dingo’s provider defaults.
  - Reporting: surface JSON report write errors and flag missing Koios `/totals` rows.

<sup>Written for commit dc1fdb0c1c041a7b9eac918767c4a8b0b2ff11b1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2684?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a new `koios-parity` command with subcommands to fetch data, run checks, view status, explain mismatches, and watch new epochs.
  * Added human-readable status output and detailed mismatch explanations, plus JSON report generation.
  * Added support for configurable network, cache, concurrency, and epoch-range options.

* **Bug Fixes**
  * Improved comparison handling to distinguish data mismatches from API issues and missing data.
  * Added graceful handling for recently updated epochs to reduce false failures.

* **Chores**
  * Updated ignored local data and cache directories.


internal/koiosparity/blockfrost_client.go
```diff
@@ -0,0 +1,238 @@
...
+	var epoch DingoEpochResp
+	if err := json.Unmarshal(body, &epoch); err != nil {
+		return 0, fmt.Errorf("dingo /epochs/latest decode: %w", err)
+	}
+	return epoch.Epoch, nil
+}
+
+// GetEpoch fetches epoch info for a specific epoch number.
+// Returns ErrAPINotFound if Dingo hasn't implemented this endpoint.
+func (b *BlockfrostClient) GetEpoch(ctx context.Context, epoch uint64) (*DingoEpochResp, error) {
```
wolf31o2: This is a blocker for the stated goal. The checker depends on `GET /api/v0/epochs/{number}`, but Dingo's Blockfrost server currently only registers `GET /api/v0/epochs/latest`, `GET /api/v0/epochs/latest/parameters`, and `GET /api/v0/epochs/{number}/parameters`. On current Dingo every aggregate check for historical epochs will become `dingo_api_missing`/`ERROR` instead of validating parity. Either add the route/handler in this PR or compare through an API surface that exists today.

internal/koiosparity/fetch.go
```diff
@@ -0,0 +1,283 @@
...
+	// We only compare closed epochs: tip - 1.
+	throughEpoch := tipEpoch - 1
+	if cfg.ThroughEpoch > 0 && cfg.ThroughEpoch < throughEpoch {
+		throughEpoch = cfg.ThroughEpoch
+	}
+
+	fromEpoch := cfg.FromEpoch
+	if fromEpoch == 0 {
+		// Resume: start from last cached epoch + 1.
+		_, lastCached, rangeErr := cache.GetFetchedEpochRange(cfg.Network)
```
wolf31o2: This resume cursor can permanently skip holes. Fetch runs epochs concurrently, and each successful later epoch writes `koios_epoch_info`. If epoch N fails after N+1 or later succeeds, the next run resumes from `MAX(epoch)+1`, so N is never retried unless the operator manually supplies `--from-epoch`. The cursor should be based on contiguous fetched coverage or explicitly select missing epochs in the requested range.

internal/koiosparity/cache.go
```diff
@@ -0,0 +1,341 @@
...
+	var epochs []uint64
+	err := c.db.Model(&KoiosEpochInfo{}).
+		Select("epoch").
+		Where("network = ?", network).
+		Order("epoch ASC").
+		Pluck("epoch", &epochs).Error
+	return epochs, err
+}
+
+// GetEpochsNeedingCheck returns epochs with Koios data but no check result yet.
```
wolf31o2: The default check queue does not match the documented workflow. After `fetch --from-epoch N` refreshes Koios rows, `check --from-epoch N` should re-check epochs whose reference data is newer than `check_epoch_status.last_checked_at`; this helper only returns epochs with no status at all. That leaves stale PASS/FAIL/mismatch rows after a Koios correction or forced re-fetch unless the operator remembers `--all`.

cmd/koios-parity/watch.go
```diff
@@ -0,0 +1,153 @@
...
+			// Subsequent iterations: all epochs that closed since last poll.
+			var fromClosed uint64
+			if lastEpoch == 0 {
+				fromClosed = current - 1
+			} else {
+				fromClosed = lastEpoch
+			}
+			toClosed := current - 1
+
+			// Advance the cursor unconditionally so a stale latest-epoch
```
wolf31o2: Advancing `lastEpoch` before fetch/check succeeds means a transient Koios or Dingo error drops this closed epoch from the watch loop. The next tick sees the same current epoch and does nothing, so the gap remains until a manual run. Move the cursor advancement after successful fetch/check, or track pending ranges so failures are retried.

wolf31o2 (changes requested):
Requesting changes. The new packages build, but the checker does not yet provide a reliable operator parity signal on current Dingo: it depends on Blockfrost routes that are not registered, can skip failed fetch holes on resume, and does not re-check stale epochs after Koios data is refreshed.

internal/koiosparity/check.go
```diff
@@ -0,0 +1,324 @@
...
+			continue
+		}
+		koiosKeySet[keyHex] = struct{}{}
+
+		var dingoPool *DingoPoolEpochData
+		if dingoPoolMap != nil {
+			dingoPool = dingoPoolMap[keyHex]
+		}
+
+		poolMismatches := ComparePoolEpoch(network, epoch, koiosPool, dingoPool, now, graceHours, epochEndTime)
```
wolf31o2: If GetPoolEpochDataMap fails, we append dingo_db_error above but still fall through with dingoPoolMap == nil and compare every Koios pool as missing. That produces pool_only_koios rows, so DetermineStatus returns FAIL instead of the issue #2679 ERROR status for a Dingo DB query failure. Please stop pool presence/value comparison when the bulk Dingo query fails, or keep all derived rows categorized as dingo_db_error.

internal/koiosparity/compare.go
```diff
@@ -0,0 +1,271 @@
...
+			Field:      "total_active_stake",
+			DingoValue: dingoEpoch.TotalActiveStake,
+			KoiosValue: koios.ActiveStake,
+			Category:   CategoryValueMismatch,
+			CheckedAt:  now,
+		})
+	}
+
+	// fees — only compared when Dingo has a reward_ada_pots row.
+	if dingoEpoch.Fees != "" && dingoEpoch.Fees != koios.Fees {
```
wolf31o2: Issue #2679 requires fees to be compared against reward_ada_pots and says missing expected Dingo rows are dingo_db_missing/ERROR. GetEpochData leaves Fees empty when reward_ada_pots is absent, and this guard silently skips the comparison, so an epoch can PASS without fees ever being checked. Please represent missing pots explicitly and emit dingo_db_missing instead of skipping fees.

cmd/koios-parity/main.go
```diff
@@ -0,0 +1,189 @@
...
+		"connection string for postgres/mysql (unused for sqlite)")
+}
+
+// resolveDingoDB returns the DingoDBConfig for cmd.
+// Plugin defaults to "sqlite"; DataDir is always set for cache-path resolution.
+func resolveDingoDB(cmd *cobra.Command) koiosparity.DingoDBConfig {
+	plugin, _ := cmd.Flags().GetString("metadata-plugin")
+	dsn, _ := cmd.Flags().GetString("metadata-dsn")
+	if plugin == "" {
+		if v := os.Getenv("DINGO_METADATA_PLUGIN"); v != "" {
```
wolf31o2: This fallback does not match Dingo's repo-standard env config. Dingo uses DINGO_DATABASE_METADATA_PLUGIN plus backend-specific DSN env vars, but koios-parity only reads DINGO_METADATA_PLUGIN/DINGO_METADATA_DSN here. With an operator's normal Dingo env, this can silently default to sqlite and compare the wrong/missing DB unless flags are supplied. Please consume/bridge the Dingo-standard env vars.

wolf31o2 (changes requested):
Requesting changes after local review. The implementation has a few correctness gaps that can make the checker report the wrong status or check the wrong database.

internal/koiosparity/check.go
```diff
@@ -0,0 +1,328 @@
...
+	result := &CheckResult{}
+	var mu sync.Mutex
+	sem := make(chan struct{}, cfg.Workers)
+	var wg sync.WaitGroup
+	errCh := make(chan error, 1)
+
+	for _, epoch := range epochs {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
```
wolf31o2: This returns immediately while earlier epoch workers may still be running. Because `DingoDB` is closed by the deferred `dingo.Close()` in `Check`, cancellation here can close the shared DB handle underneath in-flight `checkEpoch` calls, and those goroutines can still write cache/status rows after `Check` has reported cancellation. Please stop scheduling new epochs on cancellation, then always wait for started workers before returning.

internal/koiosparity/fetch.go
```diff
@@ -0,0 +1,298 @@
...
+	var mu sync.Mutex
+
+	sem := make(chan struct{}, cfg.Concurrency)
+	var wg sync.WaitGroup
+	errCh := make(chan error, 1)
+
+	for _, epoch := range epochs {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
```
wolf31o2: Same worker lifecycle issue in `Fetch`: if the context is canceled while this loop is waiting on the semaphore, the function returns before already-started `fetchEpoch` goroutines finish. Those goroutines may continue writing the cache after the caller sees cancellation. Break out of scheduling on cancellation and `wg.Wait()` before returning.

internal/koiosparity/fetch.go
```diff
@@ -0,0 +1,298 @@
...
+	// backfills.
+	poolIDs, err := koios.GetAllHistoricalPoolIDs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("get historical pool IDs: %w", err)
+	}
+
+	// Build list of epochs to fetch: only those NOT already in the cache.
+	// GetUncachedEpochs fills holes left by prior failed or interrupted runs
+	// rather than naively resuming from max(fetched)+1, which would skip gaps.
+	epochs, err := cache.GetUncachedEpochs(cfg.Network, fromEpoch, throughEpoch)
```
wolf31o2: This skips every epoch that already has a `koios_epoch_info` row, which contradicts the CLI flag that says `--from-epoch` forces a re-fetch from that epoch. With the current behavior, stale or bad cached Koios rows cannot be refreshed through the advertised flag. Either make the force path overwrite cached rows in the requested range, or change the flag semantics/help text and provide a separate refresh path.

wolf31o2 (changes requested):
No-go for merge. Local build/test/lint checks are clean, but the worker lifecycle issue can leave in-flight goroutines using shared resources after cancellation, and the fetch refresh flag does not do what it advertises. Please address these before merging.

internal/koiosparity/check.go
wolf31o2: This returns immediately while earlier epoch workers may still be running. Because `DingoDB` is closed by the deferred `dingo.Close()` in `Check`, cancellation here can close the shared DB handle underneath in-flight `checkEpoch` calls, and those goroutines can still write cache/status rows after `Check` has reported cancellation. Please stop scheduling new epochs on cancellation, then always wait for started workers before returning.

internal/koiosparity/fetch.go
wolf31o2: Same worker lifecycle issue in `Fetch`: if the context is canceled while this loop is waiting on the semaphore, the function returns before already-started `fetchEpoch` goroutines finish. Those goroutines may continue writing the cache after the caller sees cancellation. Break out of scheduling on cancellation and `wg.Wait()` before returning.

internal/koiosparity/fetch.go
wolf31o2: This skips every epoch that already has a `koios_epoch_info` row, which contradicts the CLI flag that says `--from-epoch` forces a re-fetch from that epoch. With the current behavior, stale or bad cached Koios rows cannot be refreshed through the advertised flag. Either make the force path overwrite cached rows in the requested range, or change the flag semantics/help text and provide a separate refresh path.

wolf31o2 (changes requested):
Follow-up with inline comments for the blocking issues.

internal/koiosparity/koios_client.go
```diff
@@ -0,0 +1,299 @@
...
+// existed, so deduplicating it gives a complete set of historical pool IDs.
+func (k *KoiosClient) GetAllHistoricalPoolIDs(ctx context.Context) ([]string, error) {
+	type regItem struct {
+		PoolIDBech32 string `json:"pool_id_bech32"`
+	}
+	seen := make(map[string]bool)
+	var ids []string
+	for start := 0; ; start += koiosPageSize {
+		end := start + koiosPageSize - 1
+		resp, err := k.get(ctx, "/pool_registrations?select=pool_id_bech32", start, end)
```
wolf31o2: This endpoint does not appear to exist on the current Koios v1 preview/preprod APIs. A live request to `/pool_registrations?select=pool_id_bech32` returns `Query not Found`, so the default fetch path fails before it can populate any reference data. It looks like this needs to use an endpoint Koios actually exposes for historical/current pool IDs, or otherwise derive the pool set differently.

internal/koiosparity/koios_client.go
```diff
@@ -0,0 +1,299 @@
...
+// GetPoolEpochHistory fetches a pool's history entry for a specific epoch.
+// Returns nil, nil if the pool has no row for that epoch.
+// _pool_bech32 is a required Koios function parameter; _epoch_no filters
+// server-side so only one row is returned instead of the full pool history.
+func (k *KoiosClient) GetPoolEpochHistory(
+	ctx context.Context,
+	poolBech32 string,
+	epoch uint64,
+) (*KoiosPoolHistoryItem, error) {
+	path := fmt.Sprintf(
```
wolf31o2, line 592: `pool_history` itself exists, but this projection does not: Koios returns `column pgrst_call.pool_id_bech32 does not exist` for this exact select list. Since the caller already knows the pool id, I would drop `pool_id_bech32` from the select/response model and keep the id from the request context.

internal/koiosparity/koios_client.go
```diff
@@ -0,0 +1,299 @@
...
+	"preview": "https://preview.koios.rest/api/v1",
+	"preprod": "https://preprod.koios.rest/api/v1",
+}
+
+// KoiosEpochInfoResp is the Koios /epoch_info response shape.
+type KoiosEpochInfoResp struct {
+	EpochNo      uint64 `json:"epoch_no"`
+	EndTime      int64  `json:"end_time"` // Unix timestamp of epoch boundary
+	ActiveStake  string `json:"active_stake"`
+	PoolCnt      int    `json:"pool_cnt"`
```
wolf31o2: Current `/epoch_info` responses on preview/preprod do not include `pool_cnt` or `delegator_cnt`. These fields silently decode as zero, and then the checker compares Dingo's real pool/delegator counts against zero. That makes the aggregate check report false mismatches even after fetch is fixed.

cmd/koios-parity/watch.go
```diff
@@ -0,0 +1,165 @@
...
+			logger.Warn("koios-parity: could not get latest epoch from Dingo DB",
+				"error", epochErr,
+			)
+		} else if current > 0 && current != lastEpoch {
+			// Determine the range of newly closed epochs.
+			// First iteration: only the most recently closed epoch.
+			// Subsequent iterations: all epochs that closed since last poll.
+			var fromClosed uint64
+			if lastEpoch == 0 {
+				fromClosed = current - 1
```
wolf31o2: I think this is off by one. Dingo writes `epoch_summary` for `evt.NewEpoch` when the snapshot manager handles an epoch transition, so the latest `epoch_summary` row is already the epoch this tool should be able to check. Subtracting one here means watch starts by checking the previous epoch and skips the latest available summary until the next transition.

internal/koiosparity/cache.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+// Cache wraps the SQLite cache.db.
+type Cache struct {
+	db     *gorm.DB
+	logger *slog.Logger
+}
+
+// OpenCache opens (or creates) the SQLite cache at path, running migrations.
+func OpenCache(path string, logger *slog.Logger) (*Cache, error) {
```
wolf31o2, line 210: Please add a `Close` method for `Cache` and use it at each `OpenCache` call site. `fetch`, `check`, `status`, `explain`, and especially the long-running `watch` path all open a GORM/SQLite handle today, but there is no way to close the underlying `sql.DB`, so watch leaks handles on every loop.

wolf31o2 (changes requested):
Thanks for putting this together. I like the direction of having this as an operator-side parity tool, but I do not think this is ready to merge yet. The current implementation is wired to Koios response shapes/endpoints that do not match the live preview/preprod API, so the default fetch/check path either fails before caching data or compares Dingo against zero-valued fields. I also left two follow-up correctness issues in watch/cache handling that are worth fixing while this is still fresh.

internal/koiosparity/koios_client.go
```diff
@@ -0,0 +1,298 @@
...
+	"preview": "https://preview.koios.rest/api/v1",
+	"preprod": "https://preprod.koios.rest/api/v1",
+}
+
+// KoiosEpochInfoResp is the Koios /epoch_info response shape.
+// Note: pool_cnt and delegator_cnt are not returned by preview/preprod and are omitted.
+type KoiosEpochInfoResp struct {
+	EpochNo      uint64 `json:"epoch_no"`
+	EndTime      int64  `json:"end_time"` // Unix timestamp of epoch boundary
+	ActiveStake  string `json:"active_stake"`
```
wolf31o2: This needs to handle nullable Koios fields explicitly. Current preview/preprod epoch 0 responses include `active_stake: null` and `total_rewards: null`; unmarshalling those into plain strings leaves empty values, which then get cached and compared as if Koios had provided valid reference data. I’d make these nullable in the response model and either skip/report epochs with unavailable reward inputs or fail the fetch for that epoch before writing cache state.

internal/koiosparity/fetch.go
```diff
@@ -0,0 +1,314 @@
...
+				case poolErrCh <- fmt.Errorf("pool %s history: %w", id, histErr):
+				default:
+				}
+				return
+			}
+			if item == nil {
+				return // Pool wasn't active this epoch.
+			}
+
+			if upsertErr := cache.UpsertPoolEpoch(KoiosPoolEpoch{
```
wolf31o2: This can leave stale or partial pool rows in `koios_pool_epoch`. Pool rows are upserted before the epoch is accepted, and `--force-refresh` does not clear the existing `(network, epoch)` pool set before refilling it. If a later validation fails, a pool disappears from Koios, or a refresh only partially rewrites the epoch, the old rows remain and `check` treats them as authoritative. Please make the epoch refresh atomic from the checker’s point of view, for example by deleting/replacing the epoch’s pool rows in one transaction and only committing them with the epoch info row.

internal/koiosparity/dingo_db.go
```diff
@@ -0,0 +1,229 @@
...
+		return 0, errors.New("dingo db: no epoch_summary rows found")
+	}
+	return *epoch, nil
+}
+
+// GetEpochData returns epoch-level aggregates for the given epoch.
+// Returns nil, nil when Dingo has not yet recorded an epoch_summary row.
+func (d *DingoDB) GetEpochData(epoch uint64) (*DingoEpochData, error) {
+	var summary models.EpochSummary
+	if err := d.db.Where("epoch = ?", epoch).First(&summary).Error; err != nil {
```
wolf31o2: I think this should pay attention to `epoch_summary.snapshot_ready`. The metadata store intentionally preserves partial epoch summaries and later repairs them, so reading any row as final can turn an in-progress or placeholder row into a hard value mismatch. For this checker, `snapshot_ready = false` should probably be treated like the epoch is not ready yet, or at least classified as a transient/DB-missing state rather than compared as a complete reward snapshot.

wolf31o2 (changes requested):
Thanks for the follow-up here. The shape is much closer, but I’m still requesting changes because there are a few correctness issues that can make this tool give operators a misleading parity result.

The main problem is that live preview/preprod Koios can return nullable reward fields for early epochs. With the current non-null string fields, those nulls become empty strings and can be cached as if they were valid reference data. I also think the fetch/cache path needs one more pass so a failed or forced refresh cannot leave stale pool rows behind for an epoch.

Local validation I ran:
- `go build -buildvcs=false ./cmd/koios-parity`
- `go test -buildvcs=false -race ./cmd/koios-parity ./internal/koiosparity`
- `go test -buildvcs=false ./...`

Those are clean, but the new packages still do not have tests, so the behavior issues below are not covered.

internal/koiosparity/compare.go
```diff
@@ -0,0 +1,455 @@
...
+			out = append(out, CheckMismatch{
+				Network:    network,
+				Epoch:      epoch,
+				Field:      "epoch_fees",
+				DingoValue: "",
+				KoiosValue: koios.Fees,
+				Category:   cat,
+				CheckedAt:  now,
+			})
+		} else if dingoEpoch.Fees != koios.Fees {
```
wolf31o2: Blocking: `dingoEpoch.Fees` comes from `reward_ada_pots`, but `koios.Fees` is `/epoch_info.fees` (the transaction fees in that epoch). The same Dingo field is compared again below with `/totals.fees`, which this PR correctly documents as a different quantity. Live preview epoch 1367, for example, reports `epoch_info.fees=1292170362` and `totals.fees=1245791321`, so the same Dingo value cannot pass both checks. Please remove this comparison or source and epoch-align the matching Dingo aggregate. The same issue applies to `epoch_info.total_rewards` versus `/totals.reward`; please add an integrated test where the two Koios response objects contain their real, distinct quantities.

cmd/koios-parity/watch.go
```diff
@@ -0,0 +1,169 @@
...
+				// downstream. Advance cursor so the next tick reflects reality.
+				logger.Warn("koios-parity: skipping inverted epoch range",
+					"from", fromClosed, "through", toClosed)
+				lastEpoch = current
+			} else {
+				logger.Info("koios-parity: new closed epochs detected",
+					"from", fromClosed, "through", toClosed)
+
+				fetchErr := error(nil)
+				if _, fetchErr = koiosparity.Fetch(ctx, koiosparity.FetchConfig{
```
wolf31o2: Blocking: `Fetch` reports isolated transient failures through `FetchResult.FailedEpochs` while returning a nil error, but this discards the result. The failed epoch is absent from the cache, `Check` can therefore return successfully without checking it, and the cursor is advanced at line 145, so that epoch is not retried for the lifetime of this watch process. Retain the fetch result and leave `lastEpoch` unchanged whenever `FailedEpochs` is non-empty. The cursor should likewise not advance when `result.ErrorEpochs` is non-empty, otherwise `reference_lag`/temporarily missing Dingo rows become permanent cached errors. The default run path also needs to reject a partial fetch instead of ignoring the result.

cmd/koios-parity/main.go
```diff
@@ -0,0 +1,213 @@
...
+//  1. --metadata-dsn flag
+//  2. DINGO_DATABASE_METADATA_POSTGRES_DSN / DINGO_DATABASE_METADATA_MYSQL_DSN
+//     (Dingo's per-plugin DSN env vars, consumed by the real Dingo process)
+//  3. DINGO_METADATA_DSN (generic override)
+func resolveDingoDB(cmd *cobra.Command) koiosparity.DingoDBConfig {
+	plugin, _ := cmd.Flags().GetString("metadata-plugin")
+	dsn, _ := cmd.Flags().GetString("metadata-dsn")
+
+	if plugin == "" {
+		if v := os.Getenv("DINGO_DATABASE_METADATA_PLUGIN"); v != "" {
```
wolf31o2: Blocking: this resolves removed/legacy metadata environment names rather than Dingo's current plugin configuration. Current deployments use `DINGO_PLUGINS_STORAGE_METADATA_PROVIDER` and `DINGO_PLUGINS_STORAGE_METADATA_CONFIG_*`; the shared local database path is `CARDANO_DATABASE_PATH`, which `resolveDingoDataDir` also ignores. With a normal current Dingo environment this tool can silently default to SQLite under `.dingo` and inspect the wrong or missing database. Please reuse the repository configuration path if possible, or fully honor the current provider, DSN/dataDir, and `CARDANO_DATABASE_PATH` variables with the same precedence as Dingo.

cmd/koios-parity/check.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil
+	}
+
+	fmt.Printf("check complete: %d epochs, %d mismatches (%d fail, %d error)\n",
+		result.EpochsChecked,
+		result.MismatchCount,
+		len(result.FailEpochs),
+		len(result.ErrorEpochs),
+	)
+	if len(result.FailEpochs) > 0 {
```
wolf31o2, line 81: Blocking for CI use: after printing FAIL/ERROR results this command still returns nil, so the process exits 0. The default run command similarly exits non-zero only for `FailEpochs`, not `ErrorEpochs` such as `dingo_db_error`, `dingo_db_missing`, or `reference_lag`. Please return a typed error/exit status for both failed and error results (without calling `os.Exit` inside `RunE`) so automation cannot treat an incomplete or failed parity check as success.

wolf31o2 (changes requested):
No-go for merge. The local build and test suite are clean, but the checker can still produce an incorrect parity result, permanently skip transiently incomplete epochs in watch mode, select the wrong Dingo database under the current configuration scheme, and return a successful process status for failed/error checks. Please address the inline blockers before merging.

Local validation:
- `go test -buildvcs=false -race ./cmd/koios-parity ./internal/koiosparity`
- `go test -buildvcs=false ./...`
- `make import-boundaries`
- `git diff --check`

Documentation: `ARCHITECTURE.md` was updated but needs to remain consistent with the corrected comparison behavior. `DATABASE.md` was checked and is not affected because this PR does not change Dingo's schema or metadata API.

internal/koiosparity/compare.go
```diff
@@ -0,0 +1,414 @@
...
+	now time.Time,
+) []CheckMismatch {
+	if koiosTotals == nil || dingoEpoch == nil {
+		return nil
+	}
+
+	var out []CheckMismatch
+
+	// totals_treasury
+	if dingoEpoch.Treasury != "" && dingoEpoch.Treasury != koiosTotals.Treasury {
```
wolf31o2: Blocking: these non-empty guards turn a missing Dingo `reward_ada_pots` row into a clean comparison. `GetEpochData` returns a non-nil `DingoEpochData` when `epoch_summary` exists, leaves all pot fields empty when `reward_ada_pots` is absent, and this function then skips treasury, reserves, and fees. If active stake and pool data match, the epoch is recorded as PASS. I reproduced exactly that case locally: Koios totals were present, Dingo had a ready epoch summary but no pot row, and `check --all` exited 0 with zero mismatches.

Please preserve presence separately from value (for example, a `RewardAdaPotsPresent` flag or a distinct missing-row result) and emit one `dingo_db_missing`/ERROR mismatch when the expected pot row is absent. Do not use an empty decimal string as both “missing” and a value that suppresses comparison. Add an integration-style regression test covering: ready `epoch_summary` + cached Koios totals + absent `reward_ada_pots` => ERROR and non-zero CLI outcome.

cmd/koios-parity/run.go
```diff
@@ -0,0 +1,154 @@
...
+			f.Close()
+			slog.Info("koios-parity: report written", "path", reportPath)
+		}
+	}
+
+	// A FAIL or ERROR epoch must surface as a non-zero exit so automation can't
+	// mistake an incomplete or failed parity check for success; propagated via
+	// RunE's error return (main's rootCmd.Execute() handles os.Exit(1)) rather
+	// than exiting directly here.
+	return checkResultErr(checkResult)
```
wolf31o2: Blocking for automation: `checkResultErr` only sees epochs checked during this invocation. When the cache already contains a fresh FAIL or ERROR, `Check` finds no unchecked/stale work and returns an empty `CheckResult`; this run then prints the persisted failure from `statuses` and exits 0. I reproduced the default path printing `FAIL 1` immediately before a successful exit.

Please base the default run's exit decision on the effective persisted status set it just printed/reported, not only newly processed epochs. The `check` subcommand needs equivalent behavior within its requested epoch scope, so a prior failure does not become success merely because its reference row is still fresh. A useful design is for the library to return both work-performed counters and effective status for the selected scope, or to evaluate cached statuses after checking. Add regression tests with a fresh cached FAIL and ERROR where zero epochs need rechecking; both commands must still return non-zero.

internal/koiosparity/fetch.go
```diff
@@ -0,0 +1,604 @@
...
+	network string,
+	epoch uint64,
+	poolIDs []string,
+	firstActiveEpochs map[string]uint64,
+	logger *slog.Logger,
+) (int, error) {
+	// 1. Fetch epoch info (not written yet).
+	info, err := koios.GetEpochInfo(ctx, epoch)
+	if err != nil {
+		return 0, transientErr(fmt.Errorf("get epoch info: %w", err))
```
wolf31o2: Blocking: this wraps every `GetEpochInfo` error as transient; the same unconditional wrapping is used for `/totals` and pool history. That includes non-retryable conditions the client already identifies, such as daily quota exhaustion, as well as 401/403 and other hard 4xx/schema errors. The outer dispatcher therefore records `FailedEpochs` and continues the remaining range. Within an epoch, the pool loop also keeps scheduling every remaining pool after the first hard error, which can turn quota/auth failure into a very long stream of doomed requests.

Please make retryability explicit at the Koios client boundary (typed/sentinel errors are preferable): transport/read failures, exhausted 5xx retries, and burst 429 can remain per-epoch transient; daily quota, authentication/authorization, unsupported query/schema errors, and other hard 4xx should propagate as permanent errors. A permanent error should cancel shared fetch dispatch, stop scheduling new pool requests, drain already-started workers, and return promptly. Add tests showing daily-quota/401 failures return a hard `Fetch` error without attempting later epochs/pools, while an exhausted transient 503 still lands in `FailedEpochs` and remains resumable.

wolf31o2 (changes requested):
No-go for merge at the current head. The implementation and CI coverage are much closer, but the remaining issues below can still produce a false PASS/success exit or keep a large backfill running after a non-retryable Koios failure. Please address all three inline blockers and add the suggested regression coverage. I validated the synthetic merge locally with the full Go test suite, targeted race tests, build, golangci-lint, import boundaries, and live preview endpoint smoke checks; those checks are otherwise clean.

internal/koiosparity/check.go
```diff
@@ -0,0 +1,416 @@
...
+	if totalsErr != nil && !errors.Is(totalsErr, gorm.ErrRecordNotFound) {
+		return nil, fmt.Errorf("get koios totals: %w", totalsErr)
+	}
+	allMismatches = append(allMismatches,
+		CompareEpochTotals(network, epoch, koiosTotals, dingoEpoch, now)...,
+	)
+
+	// 2. Bulk-load all pool reward inputs for this epoch from Dingo's DB.
+	// This is a single query regardless of how many pools Koios knows about.
+	dingoPoolMap, dingoPoolErr := dingo.GetPoolEpochDataMap(ctx, epoch)
```
wolf31o2: Blocking: `epoch` is a Koios reporting epoch, but the Dingo tables do not all assign that number to the same ledger period. `epoch_summary[E]` is the mark snapshot captured at the boundary into E, while Praos uses mark[E-1] as active stake for E; `reward_pool_input[E]` records `BlocksProduced` from `PreviousEpoch` (E-1). Passing Koios K unchanged here and to `GetPoolEpochDataMap` therefore compares different periods, and one same-numbered input/output row cannot align every field. Please introduce explicit field-level epoch mapping: for Koios K, active stake comes from mark K-1 and K block counts/parameters from input K+1; verify the output mapping against the reward-scheduling flow. Add a boundary test based on Dingo snapshot/reward lifecycle data rather than equal same-epoch structs.

internal/koiosparity/compare.go
```diff
@@ -0,0 +1,434 @@
...
+	// (https://github.com/cardano-community/koios-artifacts, files/grest/rpc/
+	// 00_cached_tables/pool_history_cache.sql), which omits the pledge/owner-
+	// stake bonus term the Shelley ledger spec folds into the true leader
+	// reward. That recomputed value systematically diverges from
+	// RewardPoolOutput.LeaderReward for any pool with owner stake, and
+	// deleg_rewards is derived from that same approximation plus a ROUND()
+	// on each side, so their difference isn't even guaranteed to cancel out
+	// to the exact lovelace. Comparing either would produce mismatches that
+	// reflect Koios's own reporting approximation, not a real Dingo bug.
+	if koiosPool.MemberRewards != "" && dingoPool.MemberRewardTotal != "" &&
```
wolf31o2: Blocking: this condition treats an absent `reward_pool_output` as “not comparable.” If Koios has `member_rewards` and Dingo leaves `MemberRewardTotal` empty, no mismatch is emitted and the epoch can PASS. A recent output that is still being computed is transient, and a missing historical output is incomplete data; neither is a parity pass. Preserve output presence explicitly (for example, a pointer or boolean), classify the recent case as not-ready/ERROR and the historical case as missing-data FAIL/ERROR, and add tests proving both cases cannot yield PASS. The current comparison test that expects an empty Dingo reward to be skipped should be reversed.

cmd/koios-parity/main.go
```diff
@@ -0,0 +1,400 @@
...
+
+// dsnFromMetadataConfig builds a connection string from Dingo's resolved
+// plugins.storage.metadata.config map. A flat "dsn" key (the pattern used by
+// this repo's own k8s examples, e.g. examples/dingo-gov-lens/k8s/
+// dingo-values.yaml) is used verbatim; otherwise it's assembled from discrete
+// fields the same way database/plugin/metadata/{postgres,mysql}'s Start()
+// methods do internally, so this tool connects with the exact credentials the
+// running Dingo node was configured with. Returns "" if cfg has neither.
+func dsnFromMetadataConfig(plugin string, cfg map[string]any) string {
+	if cfg == nil {
```
wolf31o2: Blocking: provider selection without a config map is valid Dingo configuration because the PostgreSQL/MySQL providers apply defaults. Returning an empty DSN here makes that setup fail later with “--metadata-dsn is required,” even though the equivalent Dingo node configuration works. Please reuse the resolved provider configuration or apply the same defaults before this early return (and keep MySQL DSN construction consistent with its provider). Add tests for PostgreSQL and MySQL with only the provider selected and an empty/nil config.

wolf31o2 (changes requested):
Requesting changes for three functional correctness/configuration issues. The standalone test cache does not need to be documented in DATABASE.md; that is not a review blocker.

To pass the next review:

1. Model the Koios-to-Dingo epoch relationship explicitly per field/table, and add boundary tests using Dingo snapshot/reward lifecycle semantics.
2. Make a missing reward_pool_output an explicit not-ready/missing-data result that cannot produce PASS, with recent and historical cases covered.
3. Resolve PostgreSQL/MySQL configuration with the same defaults and precedence as Dingo, including provider-only test cases.

After those changes, rerun the targeted race tests, full Go suite, build, lint, nilaway, import-boundaries, and diff checks. The current head passed those mechanical checks; the blockers are semantic.

cmd/koios-parity/run.go
```diff
@@ -0,0 +1,157 @@
...
+		return fmt.Errorf("get status: %w", err)
+	}
+	summary := koiosparity.BuildStatusSummary(network, fetchedEpochs, statuses)
+	koiosparity.PrintStatus(os.Stdout, summary, false, statuses)
+
+	// Write JSON report.
+	dir := resolveReportDir(reportDir)
+	reportPath := fmt.Sprintf("%s/report-%s-%s.json",
+		dir, network, time.Now().Format("2006-01-02"))
+	if mkErr := os.MkdirAll(dir, 0o750); mkErr != nil {
```
wolf31o2: Please return this error (and any file close error) instead of only logging it. As written, a read-only or unwritable --report-dir, report creation failure, BuildJSONReport failure, or WriteJSONReport failure still lets runCommand return nil when parity statuses are PASS. That makes automation believe the run succeeded while the promised JSON report is absent or incomplete. Add regression coverage for each report-output failure path.

internal/koiosparity/check.go
```diff
@@ -0,0 +1,513 @@
...
+func checkEpoch(
+	ctx context.Context,
+	cache *Cache,
+	dingo *DingoDB,
+	network string,
+	epoch uint64,
+	graceHours int,
+	logger *slog.Logger,
+) (*EpochCompareResult, error) {
+	now := time.Now()
```
wolf31o2, line 280: Please do not silently skip a missing Koios totals row. Existing caches created before totals fetching was added (and --skip-fetch runs) can have koios_epoch_info but no koios_totals, so this path can still produce PASS without validating treasury, reserves, or fees. Treat the reference row as incomplete (ERROR) or make the fetch phase invalidate/refetch epochs missing totals, and add a regression test for an upgraded cache.

wolf31o2 (changes requested):
Requires changes. The focused parity tests and CI are green, but these two paths can still report a false-success or incomplete parity result.

wolf31o2 (approved):

PR #2694, opened 2026-06-30
feat(mithril): catch-up in core mode

### LINKED ISSUE #2628: mithril: add v2 catch-up sync for existing core databases
Mithril v2 should support a catch-up mode for an existing database instead of requiring operators to wipe and perform a full from-scratch bootstrap.

The v2 CardanoDatabase artifact is structurally a better fit for this than the legacy v1 full archive: immutable files are published as per-immutable archives with a digest list, and the ancillary archive carries a ledger-state snapshot near the artifact tip. Today Dingo can resume an interrupted sync of the same artifact, but it does not intentionally advance a completed database to a newer v2 artifact.

Current behavior:

- Docker entrypoint skips Mithril when the database already exists and `sync_status` is clear.
- `dingo mithril sync` is treated as a bootstrap/resume path, not a catch-up operation for a completed DB.
- v2 extraction/cache directories are artifact-hash keyed, so a newer artifact hash causes Dingo to re-walk `0..N` instead of reusing already-known immutable trios by immutable number/digest.
- `LoadBlobsWithDB` already has useful incremental behavior: it starts from the chain tip and only copies forward when the current chain tip is behind the immutable DB tip.
- Ledger-state import is not currently safe as a blind replacement over an existing live DB: it upserts snapshot state and checkpoints import phases, but does not clear/reconcile stale live rows first.

Why this matters:

- Operational recovery of a stuck BP should not require deleting all data if the existing DB still has a valid ancestor of the target Mithril v2 snapshot.
- A v2 catch-up path could download/extract only missing immutable archives, import only missing blob blocks, and use the new ledger snapshot as a trusted boundary.
- This would reduce recovery time and avoid repeatedly replaying millions of immutable blocks into the blob store.

Proposed first scope:

- Add an explicit `dingo mithril catch-up` command, or `dingo mithril sync --catch-up`.
- Support v2 only.
- Support `core` mode only for the first version. This is the block-producer/relay use case and avoids implying historical API metadata repair is solved.
- Require the node to be stopped. The catch-up operation should set `sync_status=in_progress` before mutating state and clear it only after all phases complete.

Required safety checks:

- Verify the current DB tip intersects the target v2 artifact chain. If the existing DB is on a divergent chain with no safe intersection, abort and require a full Mithril resync.
- If the local tip is already at or ahead of the target immutable tip, avoid re-copying immutable blocks.
- If local volatile blocks extend beyond the target trusted ledger state, either prove they match the target chain or delete/refetch them.
- Refuse API mode until stale historical metadata and live-state table replacement semantics are designed.

Implementation notes:

- Reuse v2 immutable trios by immutable number and digest, not by artifact hash. The verified digest list tells us whether an existing trio is valid for the target artifact.
- Keep the existing per-trio digest verification; do not trust cache paths by name alone.
- Reuse `LoadBlobsWithDB`'s current forward-copy behavior for missing immutable blocks.
- Treat the ancillary ledger state as a replacement snapshot for core ledger state, not a normal additive import. The implementation needs a clear story for removing or superseding stale live rows such as UTxOs/accounts/pools/DReps/snapshots before importing the new snapshot state.
- Reuse the existing gap-block closure logic after immutable copy and ledger-state import.

Acceptance criteria:

- A core-mode node with an existing valid DB can be advanced to a newer Mithril v2 snapshot without deleting the database.
- Already-present immutable trios are verified and skipped; only missing/changed trios are downloaded.
- Already-imported blob blocks are skipped; only missing immutable blocks are copied.
- The resulting DB has a continuous chain from the prior local chain through the target Mithril ledger state and can start `dingo serve`.
- If the existing DB is divergent from the target artifact, the command aborts with a clear message telling the operator to perform a full resync.
- API mode is rejected with an explicit unsupported message until a separate design covers metadata replacement/backfill correctness.

wolf31o2:
Closes #2628 

---
## Summary by cubic
Adds Mithril v2 catch-up for core mode so `dingo mithril sync` can advance an already-synced database to a newer artifact without a full bootstrap. It skips already-imported immutables, validates chain intersection (including when the local tip is past the last sealed chunk), reconciles stale rows, and normalizes embedded Cardano config paths on Windows. Closes #2628.

- **New Features**
  - Catch-up dispatch in `mithril.Sync`: detect bootstrap vs resume vs catch-up; complete DBs advance; up-to-date or ahead no-op; API mode is rejected only when catch-up would run.
  - Immutable skip: `BootstrapConfig.StartImmutable` reads `sync_state` marker `mithril_immutable_max` and downloads from that number; ordered processing via `newInOrderSequencerFrom`; start is clamped to the artifact.
  - Safety check: divergence aborts; `immutable.ErrPointBeyondLastChunk` is surfaced when the local tip is beyond the last sealed chunk.
  - Reconcile on catch-up: `ImportConfig.Reconcile` builds snapshot key sets and `reconcileStaleLedgerState` deactivates rows absent from the new snapshot (UTxOs tombstoned, Accounts/DReps deactivated, Pools retired via a synthetic `pool_retirement` that supersedes same‑slot registrations and is excluded from POOLREAP refunds); implemented for `sqlite`, and for `postgres`/`mysql` behind `dingo_extra_plugins`.
  - Metadata/perf: added `GetActiveAccountCredentials`, `DeactivateAccounts`, `DeactivateDreps`, `RetirePools`; added index on `Drep.Active`; docs updated to reflect tag-gated `postgres`/`mysql` and `gcs`/`s3`; synthetic retirement/registration ordering clarified across backends.

- **Migration**
  - Core mode only; Mithril v2 only. Databases without the immutable marker fall back to full bootstrap.
  - Catch-up disables import resume and requires complete snapshot inputs; if `UTxO-HD` is used, the external UTxO table file is required.
  - On divergence, delete the database and run `dingo mithril sync` for a full resync.

<sup>Written for commit d47aa6b8ca3b32631b5b0f36a9f14d888805352e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2694?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added Mithril v2 catch-up support for already-synced databases, including safer resync decisions and automatic handling of newer immutable artifacts.
  * Added a new reconciliation flow that keeps historical records intact while marking stale ledger entries inactive.
  * Expanded database support across storage backends to handle account, governance, and pool retirement reconciliation.

* **Bug Fixes**
  * Improved sync behavior to avoid unnecessary downloads and to detect chain divergence before making changes.
  * Added clearer handling for completed, in-progress, and up-to-date sync states.


mithril/sync.go
```diff
@@ -374,6 +566,37 @@ func Sync(ctx context.Context, cfg SyncConfig) (SyncResult, error) {
...
+			db, result.ImmutableDir, targetImmutable, logger,
+		)
+		if interErr != nil {
+			return SyncResult{}, interErr
+		}
+		if upToDate {
+			if cfg.CleanupAfterLoad {
+				result.Cleanup(logger)
+			}
+			return SyncResult{}, nil
```
chrisguiney, line 598: **Confirmed bug: this early return leaves `sync_status=in_progress` and deferred indexes unbuilt, wedging the node.**

Scenario: a catch-up crashes after the volatile gap-fill has stored blocks past the artifact's immutable tip (`sync_status` was set to `in_progress` at the top of the run, deferred indexes already dropped) but before completion. The retry takes `determineSyncMode` → `syncModeResume` → `decideCatchUp` returns `engage=true`, `verifyCatchupBeforeImport` hits `ErrPointBeyondLastChunk` → `errCatchUpLocalAhead` → `upToDate`, the marker is advanced, and Sync returns `SyncResult{}, nil` here — never reaching `updateMithrilReadyState`/`ClearSyncState` or `deferredIndexes.BuildCritical()`.

Result: sync reports success, but `dingo serve` hard-fails on `incomplete sync detected (sync_status="in_progress")` (cmd/dingo/serve.go), and every subsequent `dingo mithril sync` repeats the same no-op (marker now == target). The operator is stuck until a newer artifact seals past the local tip.

The up-to-date path needs to run the completion bookkeeping (clear sync state, rebuild deferred indexes) before returning.

ledgerstate/import.go
```diff
@@ -483,6 +504,25 @@ func ImportLedgerState(
 		)
 	}
 
+	// Mithril v2 catch-up: the snapshot is the complete, trusted ledger state
+	// at its tip, so any live row absent from it has been spent/deregistered/
+	// retired by that tip. Mark those stale rows inactive (never delete). On a
+	// fresh bootstrap Reconcile is false and this is skipped entirely. This
+	// must run before advancing/checkpointing the tip, so a reconcile failure
+	// cannot leave metadata claiming the snapshot tip with stale rows active.
+	if cfg.Reconcile {
```
chrisguiney, line 526: **Confirmed bug: pools imported via the stake-snapshot fallback never reach `reconcileKeys.pools`, so this reconcile retires every active pool.**

`parsePStateMaps` can return zero pools with no error and no warning (certstate.go returns `nil, nil` when no map elements are found), which makes `certStatePoolsImported=false` and enables `allowSnapshotPoolFallback` — with no `cfg.Reconcile` check. The fallback's `importPools` call never touches `reconcileKeys` (the only `addPool` call is inside `importCertState`), so `keys.pools` stays empty. `reconcileStaleLedgerState` has no empty-set guard: every active pool in the database — including the ones the fallback just imported in this same run — is marked stale and `RetirePools` writes synthetic retirements for all of them.

The same class of gap exists for accounts/DReps if their cert-state sections parse empty (empty keys → deactivate all), but pools are the live case because the fallback actively re-adds rows and then retires them.

Fix options: have the fallback path populate `reconcileKeys.pools`, and/or make reconcile refuse to run (or skip the affected entity) when its key set is empty.

database/plugin/metadata/postgres/reconcile.go
```diff
@@ -0,0 +1,156 @@
...
+		return nil
+	}
+	db, err := d.resolveDB(txn)
+	if err != nil {
+		return fmt.Errorf("DeactivateAccounts: resolve db: %w", err)
+	}
+	for tag, keys := range groupCredentialKeysByTag(creds) {
+		if res := db.
+			Model(&models.Account{}).
+			Where("credential_tag = ? AND staking_key IN ? AND active = ?",
```
chrisguiney: **Confirmed bug: unchunked `IN ?` exceeds PostgreSQL's 65,535 bind-parameter cap for large stale sets.**

This plugin uses pgx with `PrepareStmt: true` (database.go), so the extended protocol applies and GORM expands `staking_key IN ?` to one parameter per key with no auto-chunking. A mainnet database months behind can easily accumulate >65,533 deregistered stake credentials of one tag; the statement then fails client-side with `extended protocol limited to 65535 parameters`, the reconcile transaction aborts, and the catch-up fails identically on every retry (the gap only grows).

Same pattern in `DeactivateDreps` below, and `RetirePools` has the same exposure via the unchunked `GetPools` lookup and the single `db.Create(&retirements)`.

The mysql sibling already chunks at `mysqlBatchChunkSize`, and this plugin already has `postgresBatchChunkSize` used in utxo.go — the same chunking is needed here.

ledgerstate/import.go
```diff
@@ -699,13 +774,51 @@ func importCertState(
 				"parsing cert state: %w", err,
 			)
 		}
+		if cfg.Reconcile {
```
chrisguiney, line 799: **Hard-failing here is the right call — continuing with a partial cert state under reconcile would mass-deactivate everything missing from the partial parse — but the aftermath is an unrecoverable retry loop with no operator guidance.**

By this point `importUTxOs` has already mutated the database, `sync_status` stays `in_progress`, and reconcile mode forbids checkpoint resume, so every re-run re-engages catch-up (via the marker/catchup-active flag) and fails at the same point. The api-mode rejection in mithril/sync.go at least tells the user to remove the database and resync; this error should carry the same guidance.

ledgerstate/reconcile.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/blinklabs-io/dingo/database/types"
+)
+
+// reconcileKeys is the set of live ledger keys present in an imported snapshot.
+// During a Mithril v2 catch-up import the import phases populate it; the
+// reconcile then marks any live database row whose key is absent here inactive.
+// The snapshot is the complete, trusted ledger state at its tip, so every live
+// row missing from it has been spent/deregistered/retired by that tip.
+type reconcileKeys struct {
+	utxos    map[utxoReconcileKey]struct{}
```
chrisguiney: **Design concern (confirmed cost): the entire snapshot live key set is held in memory, and it's why checkpoint resume must be disabled in reconcile mode.**

Mainnet scale: ~10–12M live UTxOs × ~45–60 B/entry in the map plus ~1.3M account credential strings ≈ 0.6–1 GB extra peak RSS, on an import path whose surrounding code deliberately streams in batches. And because the key set is RAM-only, reconcile hard-requires `ImportKey` empty — an interrupted multi-hour catch-up restarts from scratch.

A staging-table approach (stage snapshot keys as batches flow through `batchCallback`, then one indexed anti-join UPDATE per entity) would fix both at once: O(1) app memory and checkpoint resume stays usable. Fine to defer to a follow-up issue, but worth deciding consciously rather than inheriting.

database/plugin/metadata/sqlite/reconcile.go
```diff
@@ -0,0 +1,141 @@
...
+	creds []models.StakeCredentialRef,
+) error {
+	if len(creds) == 0 {
+		return nil
+	}
+	db, err := d.resolveDB(txn)
+	if err != nil {
+		return fmt.Errorf("DeactivateAccounts: resolve db: %w", err)
+	}
+	for _, c := range creds {
```
chrisguiney: **Efficiency (confirmed): the default backend's reconcile is unbatched — one UPDATE per credential here (and in `DeactivateDreps`), plus a per-pool `First()`+`Create()` in `RetirePools` (2N statements).**

The postgres/mysql implementations added in this same PR batch via `groupCredentialKeysByTag` + `IN ?`, and sqlite itself already has the pattern: `sqliteBindVarLimit` chunking is used in account.go, drep.go, and pool.go, and `GetPools` does a batched `pool_key_hash IN ?`. Tens of thousands of stale credentials means minutes of silent single-row updates inside the one write transaction — the sync looks hung, and an operator kill leaves an interrupted catch-up that must redo the whole import.

Mirroring the mysql shape (tag-grouped, chunked at `sqliteBindVarLimit`) makes the three backends line-for-line comparable.

ledgerstate/reconcile.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	if err := store.DeactivateDreps(
+		wtxn.Metadata(), staleDreps,
+	); err != nil {
+		return fmt.Errorf("deactivating %d stale DReps: %w", len(staleDreps), err)
+	}
+	if err := ctx.Err(); err != nil {
+		return fmt.Errorf("reconcile cancelled before applying stale markers: %w", err)
+	}
+	if err := store.RetirePools(
```
chrisguiney, line 268: **Confirmed, low harm: interrupted retries accumulate duplicate synthetic `pool_retirement` rows.**

Reconcile commits its own transaction before `importTip` commits, so a crash between them is possible; reconcile imports forbid checkpointing, so the retry re-runs the full reconcile. Stale pools are recomputed at the *old* tip (`GetActivePoolKeyHashesAtSlot(tmpTip.Slot)` filters `added_slot <= tip`), where the run-1 synthetic retirement (added_slot = new tip slot) is invisible — so the same pools are stale again and `RetirePools` inserts identical rows unconditionally (no `OnConflict`, no unique index on `pool_retirement`).

Verified that the duplicates don't change query results (they tie on all sort keys), so this is table bloat rather than wrong answers — but an existence check in `RetirePools` or a unique index would keep it clean.

config/cardano/loadhelper.go
```diff
@@ -38,8 +40,10 @@ func LoadCardanoNodeConfigWithFallback(
 		)
 	}
 
-	// File doesn't exist, try embedded config
-	cfg, embedErr := NewCardanoNodeConfigFromEmbedFS(embedFS, cfgPath)
+	// File doesn't exist, try embedded config. embed.FS always uses slash
+	// paths, while callers may have built cfgPath with filepath.Join.
+	embedPath := strings.ReplaceAll(filepath.ToSlash(cfgPath), "\\", "/")
```
chrisguiney: **Minor: the two conversions are redundant.** `filepath.ToSlash` is the identity on Unix and replaces `\` with `/` on Windows — exactly what the outer `strings.ReplaceAll` does on every platform. Either call alone suffices: `strings.ReplaceAll(cfgPath, "\\", "/")` (dropping the `path/filepath` import) or `filepath.ToSlash(cfgPath)`.

chrisguiney (changes requested):
## Review determination: request changes

The catch-up design is sound and well-tested — the dispatch logic, marker handling, and intersection checks all held up under close scrutiny. But verification confirmed three bugs on the new code path that each leave a node wedged or corrupt ledger state, so I don't think this should merge as-is. Details are in the inline comments; summary ranking:

**Merge-blockers (confirmed):**
1. `mithril/sync.go` — a resumed catch-up that ends "local chain ahead" returns success while leaving `sync_status=in_progress` and deferred indexes unbuilt; `dingo serve` then refuses to start and every re-run no-ops.
2. `ledgerstate/import.go` — pools imported via the stake-snapshot fallback never reach `reconcileKeys.pools`, so the reconcile pass retires **every** active pool, including the ones just imported.
3. `database/plugin/metadata/postgres/reconcile.go` — unchunked `IN ?` lists exceed pgx's 65,535 bind-parameter cap for large stale sets; the reconcile transaction then fails deterministically on every retry.

**Should fix (confirmed, lower severity):** in-memory reconcile key set (~0.6–1 GB on mainnet) that also forces checkpoint resume off; unbatched per-row reconcile on the default sqlite backend; duplicate synthetic `pool_retirement` rows on interrupted retries.

**Docs (CLAUDE.md change bar):** `DATABASE.md`'s pool-query reference SQL (~line 857) still shows the old `ORDER BY r.added_slot DESC, COALESCE(t.block_index, 0) DESC, COALESCE(c.cert_index, 0) DESC` without the new `CASE WHEN certificate_id = 0 THEN 1 ELSE 0 END DESC` synthetic-retirement key, the "matching `GetActivePoolKeyHashesAtSlot`" prose in the `GetPoolsRetiringAtEpoch` section now asserts an equivalence that no longer holds, and the `pool_retirement` table contract doesn't mention that `certificate_id = 0` rows join to no `certs` row. External SQL consumers copying the documented queries will disagree with Dingo about whether a reconcile-retired pool is retired.

**Minor cleanups (take or leave):** `groupCredentialKeysByTag` is duplicated verbatim in the postgres and mysql plugins instead of living in the shared `importutil` package; `credKeyString`/`poolKeyString` in `ledgerstate/reconcile.go` invent a second credential-key encoding when `models.StakeCredentialRef.MapKey()` already exists.

For the record, several scary-looking candidates were checked and refuted: the `GetPoolsRetiringAtEpoch` ordering divergence can't skip a POOLREAP refund (synthetic rows are epoch/slot-filtered out of boundary queries by design), wrong-network misconfiguration is caught at database open by `checkNodeSettings`, and the markerless api-mode full-sync path is pre-existing behavior this PR actually narrows.

arepala-uml (dismissed):

chrisguiney (approved):

PR #2697, opened 2026-07-01
fix(database): purge orphans before auto-migrate

### LINKED ISSUE #2696: v0.61.0 sqlite auto-migrate fails to start on existing DBs with orphaned asset rows
v0.61.0 enables sqlite auto-migrate (#2686). On startup against a database created by an earlier version (v0.60.x and earlier), the node fails to open the database and exits.

## Error

```
level=ERROR msg="constraint failed: FOREIGN KEY constraint failed (787)" component=gorm query="INSERT INTO `asset__temp`(`name`,`name_hex`,`policy_id`,`fingerprint`,`id`,`utxo_id`,`amount`) SELECT `name`,`name_hex`,`policy_id`,`fingerprint`,`id`,`utxo_id`,`amount` FROM `asset`" file=.../glebarez/sqlite@v1.11.0/migrator.go:423
level=ERROR msg="opening database: starting metadata plugin \"sqlite\": constraint failed: FOREIGN KEY constraint failed (787)"
```

## Root cause

The Utxo model declares its assets with a cascading foreign key (database/models/utxo.go):

```
Assets []Asset `gorm:"foreignKey:UtxoID;constraint:OnDelete:CASCADE"`
```

In databases built before auto-migrate was enabled, this foreign key was never created. The asset table in such a DB looks like:

```
CREATE TABLE `asset` (`name` blob,`name_hex` blob,`policy_id` blob,`fingerprint` blob,`id` integer PRIMARY KEY AUTOINCREMENT,`utxo_id` integer,`amount` text);
```

No foreign key, so the cascade was never enforced at the DB level. As utxos were deleted, their asset rows were not, and orphaned asset rows (utxo_id with no matching utxo) accumulated. On one existing DB:

```
total asset rows: 1843
orphaned (utxo_id not present in utxo): 1574
```

When v0.61.0 auto-migrate rebuilds the asset table to add the foreign key, it copies the rows into a temp table with foreign key enforcement on. The orphaned rows violate the new constraint, the copy fails, and startup aborts, so the node cannot open the existing database.

## Impact

Any node upgrading to v0.61.0 with an existing database that has orphaned rows in a cascade-FK child table cannot start. This was hit on an existing node database; a fresh database on v0.61.0 migrates cleanly because the foreign key is created at init and there are no orphans.

## Suggested fix

Before adding the foreign keys during migration, purge orphaned child rows for the cascade-FK tables, for example:

```
DELETE FROM asset WHERE utxo_id NOT IN (SELECT id FROM utxo);
```

and similarly for the other models that declare constraint:OnDelete:CASCADE (pool registration/retirement/owner/relay, transaction children, MIR rewards). These rows are already dead data, since their parent row is gone, so removing them is equivalent to what the cascade should have done.

Separately, it is worth confirming why asset rows were not removed alongside their utxos in pre-FK builds, since the orphan accumulation means the deletion path left them behind. Once the cascade FK exists it will handle future deletions, but the existing buildup needs the cleanup above to migrate.

## Workaround

Either clean the orphaned rows in the existing DB before upgrading (the DELETE above), or resync from a fresh database on v0.61.0.

wolf31o2:
Closes #2696 

---
## Summary by cubic
Purge orphaned rows before `AutoMigrate` to prevent FK table-rebuild failures on older databases. Runs automatically in the `sqlite`, `postgres`, and `mysql` metadata plugins and avoids startup errors like “FOREIGN KEY constraint failed (787)”.

- **Bug Fixes**
  - Add `models.PurgeOrphanedCascadeRows` to remove child rows whose cascade parent is missing; handles chains like transaction → utxo → asset and keeps NULL-FK rows.
  - Call the purge in each metadata plugin before `AutoMigrate`; skip when the table is absent, the FK column is missing, or the constraint already exists.
  - Add tests for utxo/asset and transaction children, pool registration/owner/relay and retirement, MIR rewards, and the already-migrated no-op; update `DATABASE.md` with the rationale.

<sup>Written for commit c42f7909cd8de9afc1be106a1a5e2125dd1e7b40. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2697?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database startup reliability by cleaning up outdated rows before schema updates run.
  * Prevents migration failures on older databases that may contain orphaned records from earlier cascade rules.
  * Keeps already-updated databases unchanged while safely removing only invalid records when needed.

* **Documentation**
  * Updated database setup notes to explain the migration flow and the behavior on legacy SQLite databases.


arepala-uml (approved):

arepala-uml:
I just asked claude where it seems like it pointed out one critical issue where Transaction.CollateralReturn uses a separate FK column collateral_return_for_tx_id on the utxo table (nullable *uint), distinct from the regular transaction_id. The PR list in spec was just covering the primary one.

If collateral_return_for_tx_id orphans aren't cleaned, AutoMigrate will still fail with FK constraint error for that column on legacy databases — which is the exact bug this PR is supposed to fix. Does `orphanPurgeSpecs()` include a dedicated entry for `{childTable: "utxo", fkColumn: "collateral_return_for_tx_id", nullable: true}?`. Please confirm once we need to add collateral return to spec or not

wolf31o2:
Verified. No fix is needed: orphanPurgeSpecs() already includes the dedicated nullable collateral-return FK purge entry

Issue #1649, opened 2026-03-15
Remove implicit fallback behavior (fail-fast audit)


wolf31o2:
## Summary

Several code paths silently fall back to defaults or skip validation when encountering errors, rather than failing fast with a clear error. This makes misconfigurations and bugs harder to diagnose in production.

## What Needs to Happen

1. Audit all code paths that catch errors and silently continue with defaults.
2. Replace implicit fallbacks with explicit error returns or clear log warnings.
3. Configuration parsing should reject unknown fields and invalid values rather than ignoring them.
4. Validation errors in block/transaction processing should propagate rather than being swallowed.
5. Add strict mode options where appropriate so operators can choose fail-fast behavior.

## Deliverables

- [ ] Audit configuration parsing for silent fallbacks
- [ ] Audit database plugin initialization for implicit defaults
- [ ] Audit block/transaction validation for swallowed errors
- [ ] Audit network/peer handling for silent error recovery
- [ ] Add tests for each corrected path ensuring errors propagate

chrisguiney:
Progress on this audit, re-checked against main at 440d01ab. Each remaining case gets its own PR, marked Refs rather than Fixes.

## Examples in the description

- Peer snapshots: network magic, version, port, relay semantics and point format are validated (#3556, #3882). Port validation for local, public and bootstrap roots is #3291 / #3760. Open question below.
- UTxORPC height zero on lookup failure: #3544.
- Blob-delete failures: #3545.
- Context-free waits: restore and truncate in #3558; the shutdown path in #4219.
- Consensus fail-open: #3543 and #4050. The remaining trust bypasses log as prototype profile only (ledger/verify_header.go).

## Further silent fallbacks found

Open PRs:
- LedgerView registration predicates turn storage errors into "not registered": #4220
- Invalid genesis k or activeSlotsCoeff replaced by a 50000 constant: #4218
- A failed reload after a committed rollback leaves stale era, epochs and protocol parameters in memory: #4217
- A Mithril trust-boundary read or parse failure is ignored at startup: #4214
- Shutdown component stops have no deadline: #4219
- A negative SQLite maxConnections is silently defaulted: #4209
- UTxORPC FollowTip sends timestamp 0 when slot-to-time fails: #4212

In progress:
- Mithril import accepts partial cert state and stake snapshots
- Pool parameters and snapshot fee silently default on decode failure during import
- Import writes incomplete epoch history when era bounds are unavailable
[...]

PR #2759, opened 2026-07-06
fix: remove implicit fallback behavior

cryptodj413:
Refs #1649 

---
## Summary by cubic
Removes silent fallback behavior across config loading and runtime, adds opt‑in strict modes for leader eligibility and slot clock, and improves diagnostics; config load is now isolated and `--debug` is registered early so validation logs are visible, refs #1649.

- **Bug Fixes**
  - Config/CLI: strict YAML decoding rejects unknown fields; in wrapped mode only `config`, `database`, `blob`, `metadata` are allowed (and under `database:` only `blob`, `metadata`). Plugin config rejects unknown plugin names and keys; build‑tag‑only names (`gcs`, `s3`, `mysql`, `postgres`) are tolerated when not compiled. Storage mode is normalized. Config loading is isolated and early `--debug` registration ensures debug logs during load/validation.
  - Runtime validation: rejects bad `logging.level`/`format`, `chainsync.strategy`, `mithril.backend`, `storageMode`, out‑of‑range ports (incl. `relayPort`), malformed durations, and non‑positive `shutdownTimeout`, `ledgerCatchupTimeout`, and `chainsync.stallTimeout` (`mithril.downloadIdleTimeout` keeps its documented semantics). Validation runs after flags so valid CLI values can override bad YAML/env.
  - Networking/diagnostics: outbound dials warn if `OutboundSourcePort` cannot be resolved; unexpected event types are logged in UTxO RPC wait/watch, connection recycle, chain‑sync resync, and `leios-notify`/`leios-fetch`; unrecognized `leios-notify` messages are logged without closing the connection. Recovered panics include stack traces (chain selection, DB worker pool, Conway script‑purpose builder). Ledger view warns when bool‑only lookups hide DB errors; default blob/metadata plugin selection is debug‑logged.

- **Migration**
  - New opt‑in strict modes: `--strict-leader-eligibility` and `--strict-slot-clock` (default false). Leave off during genesis bootstrap; enable to fail fast on persistent issues.
  - Use only supported values: logging level [`debug`, `info`, `warn`, `error`]; format [`text`, `json`]; `chainsync.strategy` (`primary`, `parallel`, `round-robin`); `mithril.backend` (`v1`, `v2`); `storageMode` (`core`, `api`); ports must be 0 or 1–65535. Durations must parse and be positive for `shutdownTimeout`, `ledgerCatchupTimeout`, and `chainsync.stallTimeout` (empty allowed).
  - Remove unknown YAML fields. In wrapped configs allow only `config`, `database`, `blob`, `metadata` at the top level; under `database:` allow only `blob` and `metadata`. Plugin configs must use documented keys; GCS no longer accepts `project-id`/`prefix` and S3 no longer accepts static credential keys (use Application Default Credentials / AWS default chain).

<sup>Written for commit 741f68a09c9221be212beffd6ad982c9628d19ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2759?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logging for unexpected network and mempool events so issues are easier to diagnose.
  * Connection setup now handles invalid outbound source-port settings more gracefully and reports a clear warning.
  * Panic recovery across several workflows now includes stack traces, helping troubleshoot unexpected crashes.
  * Configuration loading is stricter: invalid or unknown settings are now rejected earlier, including logging, sync strategy, and backend values.
  * Default storage/plugin choices are now logged when automatically selected.


internal/config/config.go
```diff
@@ -864,21 +963,39 @@ func LoadConfig(configFile string) (*Config, error) {
 		}
 
 		// If config section exists, use it for main config
-		if tempCfg.Config != nil {
-			// Overlay config values onto existing defaults
-			configBytes, err := yaml.Marshal(tempCfg.Config)
+		if tempCfg.Config.Kind != 0 {
```
wolf31o2, line 1177: This branch still leaves the root document lenient when `config:` is present. `yaml.Unmarshal(buf, &tempCfg)` ignores unknown top-level siblings, and then this branch only strict-decodes `tempCfg.Config`. So a wrapped file with `databse:` instead of `database:` will load, silently drop the intended DB settings, and fall back to defaults. Please validate the root mapping in wrapped mode too, allowing only `config`, `database`, `blob`, and `metadata`, and add a regression test for an unknown top-level sibling next to a valid `config:` section.

connmanager/outbound.go
```diff
@@ -45,12 +45,23 @@ func (c *ConnectionManager) CreateOutboundConn(
...
 		// Setup connection to use our listening port as the source port
 		// This is required for peer sharing to be useful
-		clientAddr, _ = net.ResolveTCPAddr(
+		resolvedAddr, resolveErr := net.ResolveTCPAddr(
 			"tcp",
 			fmt.Sprintf(":%d", c.config.OutboundSourcePort),
 		)
-		dialer.LocalAddr = clientAddr
-		dialer.Control = socketControl
+		if resolveErr != nil {
```
wolf31o2, line 52: This is now observable, but it is still a fallback: an invalid configured source port logs and then dials without source-port reuse. Since source-port reuse is required for peer sharing to be useful, I think this should fail during config validation/startup, or be gated behind an explicit tolerant mode. Logging and continuing is better than silent behavior, but it does not satisfy the fail-fast part of #1649.

ledger/verify_header.go
```diff
@@ -247,6 +247,14 @@ func (ls *LedgerState) verifyBlockHeaderCrypto(
 // Byron blocks are skipped (PBFT). A missing total-stake or unavailable active
 // slot coefficient is logged and skipped rather than rejecting, to tolerate
 // early-chain bootstrap states where the genesis snapshot is not yet written.
+//
+// Reviewed for issue #1649 (fail-fast audit) and intentionally kept as
```
wolf31o2: This comment makes sense as a risk note, but it also means #1649 is not complete here. If bootstrap needs a warn-and-continue window, we should distinguish expected bootstrap absence from unexpected post-bootstrap absence and fail after that boundary, with tests and DevNet coverage. Otherwise this PR should not close #1649.

ledger/state.go
```diff
@@ -6278,6 +6285,11 @@ func (ls *LedgerState) validateTxCore(
 	snapshotPParams := ls.currentPParams
 	snapshotPrevEraPParams := ls.prevEraPParams
 	ls.RUnlock()
+	// Reviewed for issue #1649 (fail-fast audit) and kept as graceful
```
wolf31o2: Same concern here: this is still graceful degradation in transaction validation. Falling back to the snapshot tip slot may be a valid operational tradeoff, but it should either be controlled by an explicit strict/fail-fast option or called out as remaining #1649 work. As written, the PR documents that this fallback remains while claiming to close the audit.

database/database.go
```diff
@@ -164,9 +164,21 @@ func New(
 	// Apply defaults for empty fields
 	if configCopy.BlobPlugin == "" {
```
wolf31o2, line 172: This still applies implicit plugin defaults when the caller passes an empty plugin name. The debug log helps, but #1649 specifically calls out database plugin initialization. If defaulting is intended, I would prefer it happen once in config defaulting with a clear source of truth; the database constructor should probably reject an empty plugin after config has been resolved, or at least this should be listed as remaining fallback work rather than part of a closed audit.

wolf31o2 (changes requested):
Thanks for tightening several config paths and adding observability. I am requesting changes because the PR currently says `Closes #1649`, but it is only a partial pass and still has at least one strictness hole that preserves silent fallback behavior.

The blocking issue is wrapped YAML config: when `config:` is present, unknown top-level siblings are still decoded leniently and ignored. That means a typo like `databse:` can silently drop database settings and fall back to defaults. That conflicts directly with the config parsing part of #1649 and the PR description.

I would also avoid closing #1649 from this PR. There are still explicit warn-and-continue or fallback paths that need either better error handling, an operator-controlled strict mode, or a clearly scoped follow-up: database plugin defaulting, invalid outbound source-port fallback, leader eligibility skip cases, tx validation slot-clock fallback, best-effort Leios drops, and recovered chain-selection panics. Some of these may be legitimate operational tradeoffs, but then the PR should say `Refs #1649` rather than `Closes #1649` and should track the remaining work.

Validation I ran locally: `go test ./...` and `make test` both passed.

cryptodj413:
Let me replace the "Closes" with "Refs" and address these reviews. Thanks for your time.

internal/config/config.go
```diff
@@ -863,22 +1010,52 @@ func LoadConfig(configFile string) (*Config, error) {
...
-			err = yaml.Unmarshal(buf, globalConfig)
-			if err != nil {
+			// compatibility). The blob/metadata/database sections are
+			// sibling keys handled separately above, so strip them
+			// before strict-decoding the remainder into Config.
+			configBuf := buf
+			if len(root.Content) > 0 {
+				removeMappingKeys(
+					root.Content[0],
+					[]string{"blob", "metadata", "database"},
```
wolf31o2, line 1208: This still leaves a silent fallback hole one level lower. Because database is stripped before strict Config decoding, unknown children under database: are only seen by the lenient tempConfig decode. A typo like database.blbo: or database.metadtaa: is silently ignored and the node falls back to the default plugin settings, which is the behavior this PR is trying to remove. Please validate the database mapping itself only contains blob and metadata before stripping/processing it, and add coverage for typoed database children.

dingo.yaml.example
```diff
@@ -52,10 +52,9 @@ database:
     gcs:
       # Google Cloud Storage bucket name
       bucket: ""
-      # Google Cloud project ID
-      project-id: ""
-      # Path prefix within the bucket
-      prefix: ""
+      # Project and credentials are always resolved via Application
+      # Default Credentials. There is no plugin config option for
+      # project ID or a bucket key prefix.
```
wolf31o2, line 57: This removes the GCS project-id/prefix options from the example, but README.md still documents those as valid GCS options and still documents S3 static access-key-id/secret-access-key options. With the new strict plugin-key validation, following the README will now fail config load. Please update README.md to match the actual registered plugin options: GCS only bucket; S3 endpoint, bucket, region, and prefix, with credentials from the AWS default chain.

wolf31o2 (changes requested):
No-go for merge. The wrapped-config blocker was addressed, and the focused tests pass, but there are still two merge-blocking issues: a remaining silent fallback path under database:, and stale README plugin docs that now describe rejected config keys.

Validation run locally: git diff --check origin/main...HEAD; go test ./internal/config ./database/plugin ./database ./connmanager ./ledger ./chainselection ./api/utxorpc ./ouroboros.

internal/config/config.go
```diff
@@ -246,6 +254,245 @@ func midnightYAMLFieldSet(field string) bool {
...
+	}{
+		{"shutdownTimeout", cfg.ShutdownTimeout},
+		{"ledgerCatchupTimeout", cfg.LedgerCatchupTimeout},
+		{"chainsync.stallTimeout", cfg.Chainsync.StallTimeout},
+		{"mithril.downloadIdleTimeout", cfg.Mithril.DownloadIdleTimeout},
+	} {
+		if d.value == "" {
+			continue
+		}
+		if _, err := time.ParseDuration(d.value); err != nil {
```
wolf31o2: This only checks whether the value parses. `time.ParseDuration` accepts `"0s"` and negative durations, so `shutdownTimeout` and `chainsync.stallTimeout` accept values that their consumers later silently replace with defaults (`configuredShutdownTimeout` falls back to 30s for `<= 0`, and chain-sync only applies a configured timeout when it is `> 0`). That retains the implicit fallback behavior this PR is intended to remove. Please validate the parsed duration semantically—requiring `> 0` for these timeout fields unless a documented non-positive value has an intentional meaning—and add zero/negative tests across YAML, environment, and CLI inputs.

wolf31o2 (changes requested):
No-go for merge at the current head. The earlier wrapped-config and plugin-documentation blockers are fixed, and the focused tests plus conformance pass, but duration validation still preserves an implicit fallback path. Please reject non-positive timeout values where they are not explicitly supported and add YAML/env/CLI coverage for zero and negative durations. Because this touches header and transaction validation, DevNet validation remains required after the fix.

internal/config/config.go
```diff
@@ -973,6 +1294,16 @@ func LoadConfig(configFile string) (*Config, error) {
 		)
 	}
 
+	// Validate relayPort/logging/chainsync/mithril fields. Also re-run at
+	// the end of ApplyFlags: CLI flags are applied after LoadConfig
+	// returns, so validating only here would let a bad --port,
+	// --logging-level, --logging-format, --chainsync-strategy, or
+	// --mithril-backend flag reintroduce a value this check already
+	// rejects for YAML/env.
+	if err := validateRuntimeConfig(globalConfig); err != nil {
```
wolf31o2: This semantic validation runs before ApplyFlags, so it violates the documented `CLI > environment > YAML > defaults` precedence whenever a lower layer contains a value that a valid CLI flag should replace. I reproduced this with:\n\n```yaml\nshutdownTimeout: "0s"\n```\n\nRunning `dingo --config dingo.yaml --shutdown-timeout=30s load ...` exits here with `invalid shutdownTimeout "0s"` before the valid CLI override is applied. The same ordering affects the other values checked by validateRuntimeConfig, such as logging level/format, ports, chainsync strategy, Mithril backend, and storage mode. Keep strict YAML shape/type decoding during load, but defer semantic validation until after flags have been applied in command startup; provide an explicit validation path for non-CLI callers. Please add a regression proving a valid CLI value overrides an invalid YAML/environment semantic value.

database/plugin/register.go
```diff
@@ -85,18 +86,86 @@ func ProcessEnvVars() error {
...
+// build tags (e.g. -tags dingo_extra_plugins for gcs/s3/mysql/postgres).
+// Config sections for these names are tolerated even when the corresponding
+// plugin isn't compiled into this build; selecting one as the active
+// blob/metadata plugin still fails downstream via MissingPluginError.
+var buildTagGatedPluginNames = map[PluginType]map[string]struct{}{
+	PluginTypeBlob:     {"gcs": {}, "s3": {}},
+	PluginTypeMetadata: {"mysql": {}, "postgres": {}},
+}
+
+// ProcessConfig applies plugin-specific config values from a parsed YAML
```
wolf31o2, line 100: This changes the externally visible configuration contract for every storage plugin: unknown plugin names/keys now fail startup, while optional build-tag-gated plugin sections are treated specially. Under the repository documentation rules, storage-plugin behavior changes require a `DATABASE.md` update; updating only README and dingo.yaml.example does not satisfy that bar. Please document the strict plugin-name/key behavior and the build-tag-gated exception in DATABASE.md. For example, `database.blob.badger.buckit` now fails instead of being ignored, whereas an optional GCS/S3 section cannot be fully key-validated when that plugin is not compiled into the binary.

wolf31o2 (changes requested):
Requesting changes on the current head. The previous non-positive timeout finding is fixed correctly, the affected race-enabled package tests and conformance suite pass, and the branch merges cleanly with current main. Two items remain: semantic validation currently runs before higher-precedence CLI flags are applied, and the externally visible storage-plugin configuration behavior needs the required DATABASE.md documentation. DevNet should be rerun after the consensus-affecting changes are ready.

internal/config/config.go
```diff
@@ -246,6 +254,278 @@ func midnightYAMLFieldSet(field string) bool {
...
+	// used for peer-sharing-friendly source-port reuse. An out-of-range
+	// value used to only be caught when dialing, where it silently fell
+	// back to dialing without source-port reuse; catch it here instead.
+	if cfg.RelayPort > 65535 {
+		return fmt.Errorf(
+			"invalid relayPort: %d (must be 0 to disable, or 1-65535)",
+			cfg.RelayPort,
+		)
+	}
+	switch strings.ToLower(strings.TrimSpace(cfg.Logging.Level)) {
```
wolf31o2, line 395: P2: This rejects the configured logging level even when `--debug` is present, but `cmd/dingo/main.go` documents `--debug` as the highest-precedence override that forces debug level regardless of the configured value. I reproduced the precedence failure with `DINGO_LOGGING_LEVEL=bogus go run ./cmd/dingo --debug version`: startup exits here instead of running at debug level. Apply the `--debug` override before this semantic check (or otherwise exempt the superseded lower-priority level), and add an env/YAML + `--debug` regression test.

internal/config/flags.go
```diff
@@ -200,6 +202,14 @@ func ApplyFlags(cmd *cobra.Command, cfg *Config) error {
 		clearMidnightNetworkDefaults(cfg, previousNetwork)
 	}
 	applyMidnightNetworkDefaults(cfg)
+	// Semantic validation runs here, after flags have been merged in, so a
+	// valid --port/--logging-level/--chainsync-strategy/--mithril-backend
+	// flag can override a semantically invalid YAML/env value instead of
+	// LoadConfig rejecting it beforehand. See ValidateRuntimeConfig's doc
+	// comment.
+	if err := ValidateRuntimeConfig(cfg); err != nil {
```
wolf31o2, line 224: `ApplyFlags` has merged the flags declared in `flagSpecs`, but it has not applied the separate root-level `--debug` flag; that override is only consumed later by `commonRun`. Consequently this validation point is still too early for the effective logging level, despite the comment saying semantic validation occurs after CLI values are merged.

wolf31o2 (changes requested):
No-go for merge at the current head. The branch merges cleanly with current main and the affected race-enabled package tests plus conformance pass, but the new semantic validation still violates the documented CLI precedence for the special --debug override. Please apply that override before logging-level validation and add regression coverage. DevNet remains required after the fix because this PR changes header and transaction validation.

ledger/verify_header.go
```diff
@@ -742,7 +748,19 @@ func (ls *LedgerState) verifyBlockLeaderEligibility(
 	}
 	if totalStake == 0 {
 		// Genesis snapshot not yet written (very early bootstrap).
-		// Skip rather than reject.
+		// Skip rather than reject, unless the operator opted into
+		// strict leader eligibility.
+		if ls.config.StrictLeaderEligibility {
```
wolf31o2, line 753: Blocking: add ledger-level behavioral coverage for this strict mode. The only tests that mention `StrictLeaderEligibility` verify default/YAML/CLI plumbing; none exercises block verification. Please cover both `true` and `false` for zero total active stake and for a missing/non-positive active-slot coefficient, asserting hard rejection only in strict mode. This changes block acceptance, and the DevNet run uses the default `false`, so it cannot catch regressions in this branch.

ledger/state.go
```diff
@@ -6568,8 +6604,23 @@ func (ls *LedgerState) validateTxCore(
...
 	ls.RUnlock()
+	// A slot clock read error normally degrades gracefully to the snapshot
+	// tip slot: it is monitored via the slotClockFallbacks metric (plus the
+	// debug log below), so sustained occurrences stay visible without
+	// failing tx validation on a transient error. Operators who prefer
+	// fail-fast can set StrictSlotClock, which rejects the transaction
+	// instead of validating it against a stale reference slot (issue #1649).
 	currentSlot, currentSlotErr := ls.CurrentSlot()
 	if currentSlotErr != nil {
+		if ls.config.StrictSlotClock {
```
wolf31o2, line 6615: Blocking: exercise the `CurrentSlot` error behavior with `StrictSlotClock` both enabled and disabled. Existing coverage only proves that the flag can be configured; it does not verify that strict mode rejects the transaction or that tolerant mode uses the snapshot-tip fallback (and records the fallback metric). This is a transaction-validation behavior change that the default DevNet configuration does not execute.

wolf31o2 (changes requested):
No-go on the current head (`741f68a0`). I reproduced the PR against current `main` (`ed882bb2`) locally, and it no longer merges cleanly: `internal/config/config.go` conflicts with the new `ApplyDefaults`/`Validate` lifecycle, and `ledger/state.go` conflicts with the new `loadStateSnapshots` transaction-validation path. Please rebase and preserve both of those mainline refactors when integrating this work.

The two inline comments are also blocking because the new strict modes affect block acceptance and transaction validation, but only their configuration plumbing is tested. The full DevNet run passed with the defaults (`false`), so it does not exercise either enabled strict path.

The exact head is also not `make golines` clean; the target rewrites 22 changed files, including PR-added tests.

Before `main` advanced, the synthetic merge passed the affected race-enabled package tests, conformance, import boundaries, and the full DevNet suite including epoch-boundary consensus. The latest-base result cannot be tested until the conflicts are resolved. `DATABASE.md` and `ARCHITECTURE.md` were checked and their updates are appropriate, but should be rechecked after the rebase.

Issue #393, opened 2025-02-02
Improve support for chain block number query


agaffney:
We hard-code a `1` value in the query result, but it's not clear what it's for or when to use other values. Just guessing, a value of `0` means "no block" (at chain genesis), and `1` means that we have a block number to provide

Issue #390, opened 2025-02-02
Calculate chain density metric


agaffney:
We export the metric, but there's no value. Figure out how to calculate this value.

cryptodj413:
Already solved by PR #1708 and PR #2477

Issue #387, opened 2025-02-02
Make tracing configurable


agaffney:
We currently have it disabled by default (it was noisy when nothing was listening on the default URL) and don't provide a way to enable/configure it from the CLI. Wire up the env vars and pass it through to the top-level node config.

Issue #276, opened 2024-12-12
Improve socket file handling


wolf31o2:
Check if the file exists and attempt to automatically remove it. Surface up errors.

PR #2763, opened 2026-07-06
feat(config): add Validate() method for startup validation

### LINKED ISSUE #1485: Config: add Validate() method for startup validation
## Summary

There is no validation of loaded configuration values. Invalid settings (bad port ranges, missing required paths, nonsensical combinations) are silently accepted and cause runtime failures instead of clear startup errors.

## Current State

- GetConfig/SetConfig are thread-safe (fixed previously)
- No Validate() method on Config
- RunMode "load" silently fails if ImmutableDbPath is empty
- No port range validation
- No path traversal prevention for CardanoConfig

## What Needs to Happen

1. Add a Config.Validate() method that checks:
   - Port ranges (1024-65535 for non-privileged)
   - RunMode "load" requires ImmutableDbPath to be set
   - Path security on CardanoConfig (no path traversal)
2. Call Validate() during startup after config loading, before any services start.
3. Return clear error messages identifying which setting is invalid and why.

Existing open issues for specific hardcoded values (#276, #387, #390, #393) are separate from this general validation framework.

chrisguiney:
Closes #1485

Adds `Config.Validate()` to `internal/config`, checking the fully merged configuration (defaults, YAML, env, CLI flags) before any services start. All violations are aggregated into a single startup error naming each bad setting, so operators can fix everything in one pass.

## Checks

- Mode enums: `runMode`, `startEra`, `storageMode`
- `runMode: load` requires `immutableDbPath`
- Ports: required listeners (`port`, `privatePort`, `metricsPort`) must be non-zero; optional API ports accept 0 as disabled; all must be ≤ 65535; ports below 1024 are rejected unless the process runs as root
- Path-traversal guard on `cardanoConfig` (no `..` components), matching the existing network-name guard
- TLS cert/key must be set as a pair
- Mempool watermark ranges and eviction < rejection ordering
- Block producer requires all three credential paths
- Network name validity; `network` or `networkMagic` must be set
- Duration strings (`shutdownTimeout`, `ledgerCatchupTimeout`, `chainsync.stallTimeout`, `mithril.downloadIdleTimeout`) parse at startup instead of failing at point of use
- `chainsync.strategy` and `mithril.backend` accepted-value sets

## Wire-up

Validation runs in `PersistentPreRunE` after `ApplyFlags`. This also closes a gap where CLI flag values bypassed `LoadConfig`'s inline checks entirely (flags are applied after `LoadConfig` returns, so e.g. `--eviction-watermark 1.5` was never validated). The `dingo load <path>` positional argument is merged into `ImmutableDbPath` before validation so a config with `runMode: load` plus a path argument does not fail spuriously.

Deliberately out of scope: file/dir existence checks (embedded-FS fallbacks and later-created paths make stat-at-validate wrong for several fields), and the closed issues referenced by the ticket (#276, #387, #390, #393).

## Testing

- New table-driven tests covering every rule plus error aggregation; the privileged-port check is parameterized on privilege so tests don't depend on the runner's UID
- Full `make test` (`-race`, includes conformance) passes; golangci-lint, modernize, and import-boundaries clean
- Manual smoke tests: bad port, missing load path, and `--cardano-config ../../etc/passwd` each fail with a clear named-setting error; multiple violations report together

ARCHITECTURE.md Configuration section updated; DATABASE.md not affected.

---
## Summary by cubic
Adds `Config.Validate()` for startup validation keyed to the command’s effective run mode, and defers defaulting/validation until all sources (YAML, env, flags, args) are merged. Also rejects a negative `historyExpiry.frequency` instead of silently defaulting.

- New Features
  - Validate the merged config via `internal/config.Config.Validate()` before any services start, using the effective mode (serve, load, sync, mithril).
  - Scope listener checks to active listeners: serving requires relay/private/metrics; `sync` validates metrics/debug; read‑only `mithril list/show` and `load` start none; API ports validate only under `api` storage or when configured `runMode` is `dev`.
  - Add `internal/config.Config.ApplyDefaults()` to fill derived defaults after `ApplyFlags` (runMode, mempoolCapacity, watermarks, forge thresholds, history‑expiry frequency) and then validate; `LoadConfig` now only parses/merges sources.
  - Export accepted‑value lists from parsers and derive from them: `chainsync.AcceptedHeaderSyncStrategyNames` and `mithril.AcceptedBackends`. `internal/config` duplicates are guarded by bidirectional parity tests in `cmd/dingo`.

- Bug Fixes
  - Port validation: detect duplicate non‑zero ports only when bind addresses overlap (wildcard vs specific); privileged‑port checks use the actual cutoff (root/Windows/CAP_NET_BIND_SERVICE or Linux sysctl).
  - Exempt `version`, `list`, `help`, and `completion` from config loading and validation.
  - `dingo load <path>` merges the positional path before validation; load mode enforces `immutableDbPath`.
  - Harden `cardanoConfig` path guard to reject any `..`; validate CLI flags, parse duration strings at startup, and validate mempool watermarks (including NaN) and block‑producer credentials.
  - Support networkMagic‑only configs: allow an empty `network` at load/flag time; validation still requires either `network` or `networkMagic`.
  - Fix precedence gaps: CLI flags can override invalid YAML values and influence derived defaults (e.g., `--run-mode leios` sets the leios mempool capacity).
  - History expiry: preserve an explicitly negative `historyExpiry.frequency` from any source and fail validation; only zero defaults to 1h.

<sup>Written for commit b61c234057770730d2171dbfc4872ffc3b3887f7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2763?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Validation now runs deterministically before services start, aggregating multiple configuration issues at once.
  * CLI determines an “effective” run mode for validation; informational commands skip config loading/validation.
  * `dingo load <path>` can supply `immutableDbPath` even when omitted from YAML.
* **Bug Fixes**
  * Stricter listener port checks (including privileged-port rules, overlap collisions, and mode-specific validation).
  * Improved TLS, mempool watermark, chainsync, Mithril, block-producer credentials, path traversal, and duration/strategy validation.
  * Network parsing is relaxed for networkMagic-only setups and `--network ""`.
* **Tests**
  * Added expanded and parity-focused validation coverage for accepted strategy/back-end values.

### DIFF
--- a/cmd/dingo/config_parity_test.go
+++ b/cmd/dingo/config_parity_test.go
@@ -28,7 +28,12 @@ import (
 // it validates against (config.AcceptedChainsyncStrategies and
 // config.AcceptedMithrilBackends) are duplicated from those downstream
 // parsers. These parity tests live in cmd/dingo, which can import all
-// three, and guard against drift in both directions:
+// three. The canonical sets are not hand-maintained: they come from
+// the parser packages' exported accepted-value lists
+// (chainsync.AcceptedHeaderSyncStrategyNames, mithril.AcceptedBackends),
+// which the parsers themselves derive from, so a value added to a
+// parser fails these tests until config's duplicated list is updated.
+// The tests guard drift in both directions:
 //
 //   - forward: every value config accepts must be accepted by the real
 //     parser, so a config never passes Validate() only to fail at
@@ -40,19 +45,11 @@ import (
 // TestChainsyncStrategyWhitelistParity checks config.AcceptedChainsyncStrategies
 // against chainsync.ParseHeaderSyncStrategy.
 func TestChainsyncStrategyWhitelistParity(t *testing.T) {
-	// canonical is the contract of chainsync.ParseHeaderSyncStrategy;
-	// it mirrors that function's switch. The set-equality check below
-	// ties config's list to it, and the parse check ties it to the real
-	// parser, so a change to the parser's switch fails this test until
-	// both this list and config's list are updated.
-	canonical := []string{
-		"", "primary", "parallel", "round-robin", "roundrobin", "round_robin",
-	}
 	assertWhitelistParity(
 		t,
 		"chainsync.strategy",
 		config.AcceptedChainsyncStrategies,
-		canonical,
+		chainsync.AcceptedHeaderSyncStrategyNames(),
 		func(v string) error {
 			_, err := chainsync.ParseHeaderSyncStrategy(v)
 			return err
@@ -63,10 +60,9 @@ func TestChainsyncStrategyWhitelistParity(t *testing.T) {
 // TestMithrilBackendWhitelistParity checks config.AcceptedMithrilBackends
 // against resolveMithrilBackend.
 func TestMithrilBackendWhitelistParity(t *testing.T) {
-	// canonical is the contract of resolveMithrilBackend: the empty
-	// string (which selects v2) plus the two backend constants it
-	// switches on.
-	canonical := []string{"", mithril.BackendV1, mithril.BackendV2}
+	// resolveMithrilBackend accepts mithril.AcceptedBackends plus the
+	// empty string, which selects the default (v2).
+	canonical := append([]string{""}, mithril.AcceptedBackends()...)
 	assertWhitelistParity(
 		t,
 		"mithril.backend",
@@ -100,13 +96,14 @@ func assertWhitelistParity(
 		}
 	}
 	// Reverse: every value in the parser's contract must actually parse
-	// (anchoring the contract to the real parser) and config's set must
-	// match the contract exactly.
+	// (anchoring the exported accepted list to the real parser) and
+	// config's set must match the contract exactly.
 	for _, v := range canonical {
 		if err := accepts(v); err != nil {
 			t.Errorf(
 				"%s: canonical value %q is rejected by the parser; "+
-					"update the canonical set: %v",
+					"the parser package's exported accepted list "+
+					"disagrees with its parser: %v",
 				name, v, err,
 			)
 		}
--- a/cmd/dingo/config_parity_test.go
+++ b/cmd/dingo/config_parity_test.go
@@ -1,0 +1,55 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+
+	"github.com/blinklabs-io/dingo/chainsync"
+	"github.com/blinklabs-io/dingo/internal/config"
+)
+
+// TestChainsyncStrategyWhitelistParity guards against drift between the
+// chainsync.strategy values internal/config.Validate accepts and those
+// chainsync.ParseHeaderSyncStrategy actually accepts. internal/config
+// cannot import chainsync without pulling node subsystems into the
+// config package, so this parity check lives in cmd/dingo, which can
+// import both.
+func TestChainsyncStrategyWhitelistParity(t *testing.T) {
+	for _, strategy := range config.AcceptedChainsyncStrategies {
+		if _, err := chainsync.ParseHeaderSyncStrategy(strategy); err != nil {
+			t.Errorf(
+				"config accepts chainsync.strategy %q but "+
+					"ParseHeaderSyncStrategy rejects it: %v",
+				strategy, err,
+			)
+		}
+	}
+}
+
+// TestMithrilBackendWhitelistParity guards against drift between the
+// mithril.backend values internal/config.Validate accepts and those
+// resolveMithrilBackend actually accepts.
+func TestMithrilBackendWhitelistParity(t *testing.T) {
+	for _, backend := range config.AcceptedMithrilBackends {
+		if _, err := resolveMithrilBackend(backend); err != nil {
+			t.Errorf(
+				"config accepts mithril.backend %q but "+
+					"resolveMithrilBackend rejects it: %v",
+				backend, err,
+			)
+		}
+	}
+}
--- a/cmd/dingo/config_parity_test.go
+++ b/cmd/dingo/config_parity_test.go
@@ -15,41 +15,117 @@
 package main
 
 import (
+	"maps"
 	"testing"
 
 	"github.com/blinklabs-io/dingo/chainsync"
 	"github.com/blinklabs-io/dingo/internal/config"
+	"github.com/blinklabs-io/dingo/mithril"
 )
 
-// TestChainsyncStrategyWhitelistParity guards against drift between the
-// chainsync.strategy values internal/config.Validate accepts and those
-// chainsync.ParseHeaderSyncStrategy actually accepts. internal/config
-// cannot import chainsync without pulling node subsystems into the
-// config package, so this parity check lives in cmd/dingo, which can
-// import both.
+// internal/config cannot import chainsync or mithril without pulling
+// node subsystems into the config package, so the accepted-value sets
+// it validates against (config.AcceptedChainsyncStrategies and
+// config.AcceptedMithrilBackends) are duplicated from those downstream
+// parsers. These parity tests live in cmd/dingo, which can import all
+// three, and guard against drift in both directions:
+//
+//   - forward: every value config accepts must be accepted by the real
+//     parser, so a config never passes Validate() only to fail at
+//     startup;
+//   - reverse: config's accepted set must match the parser's contract
+//     exactly, so a value added to the parser without updating config
+//     (which would spuriously reject a valid config) is caught too.
+
+// TestChainsyncStrategyWhitelistParity checks config.AcceptedChainsyncStrategies
+// against chainsync.ParseHeaderSyncStrategy.
 func TestChainsyncStrategyWhitelistParity(t *testing.T) {
-	for _, strategy := range config.AcceptedChainsyncStrategies {
-		if _, err := chainsync.ParseHeaderSyncStrategy(strategy); err != nil {
+	// canonical is the contract of chainsync.ParseHeaderSyncStrategy;
+	// it mirrors that function's switch. The set-equality check below
+	// ties config's list to it, and the parse check ties it to the real
+	// parser, so a change to the parser's switch fails this test until
+	// both this list and config's list are updated.
+	canonical := []string{
+		"", "primary", "parallel", "round-robin", "roundrobin", "round_robin",
+	}
+	assertWhitelistParity(
+		t,
+		"chainsync.strategy",
+		config.AcceptedChainsyncStrategies,
+		canonical,
+		func(v string) error {
+			_, err := chainsync.ParseHeaderSyncStrategy(v)
+			return err
+		},
+	)
+}
+
+// TestMithrilBackendWhitelistParity checks config.AcceptedMithrilBackends
+// against resolveMithrilBackend.
+func TestMithrilBackendWhitelistParity(t *testing.T) {
+	// canonical is the contract of resolveMithrilBackend: the empty
+	// string (which selects v2) plus the two backend constants it
+	// switches on.
+	canonical := []string{"", mithril.BackendV1, mithril.BackendV2}
+	assertWhitelistParity(
+		t,
+		"mithril.backend",
+		config.AcceptedMithrilBackends,
+		canonical,
+		func(v string) error {
+			_, err := resolveMithrilBackend(v)
+			return err
+		},
+	)
+}
+
+// assertWhitelistParity verifies that configList (the values
+// internal/config accepts) and canonical (the downstream parser's
+// contract) describe the same set, and that both are actually accepted
+// by the real parser via accepts.
+func assertWhitelistParity(
+	t *testing.T,
+	name string,
+	configList, canonical []string,
+	accepts func(string) error,
+) {
+	t.Helper()
+	// Forward: everything config accepts must parse.
+	for _, v := range configList {
+		if err := accepts(v); err != nil {
 			t.Errorf(
-				"config accepts chainsync.strategy %q but "+
-					"ParseHeaderSyncStrategy rejects it: %v",
-				strategy, err,
+				"%s: config accepts %q but the parser rejects it: %v",
+				name, v, err,
 			)
 		}
+	}
+	// Reverse: every value in the parser's contract must actually parse
+	// (anchoring the contract to the real parser) and config's set must
+	// match the contract exactly.
+	for _, v := range canonical {
+		if err := accepts(v); err != nil {
+			t.Errorf(
+				"%s: canonical value %q is rejected by the parser; "+
+					"update the canonical set: %v",
+				name, v, err,
+			)
+		}
+	}
+	if got, want := toStringSet(configList), toStringSet(canonical); !maps.Equal(
+		got,
+		want,
+	) {
+		t.Errorf(
+			"%s: config accepted set %v does not match parser contract %v",
+			name, configList, canonical,
+		)
 	}
 }
 
-// TestMithrilBackendWhitelistParity guards against drift between the
-// mithril.backend values internal/config.Validate accepts and those
-// resolveMithrilBackend actually accepts.
-func TestMithrilBackendWhitelistParity(t *testing.T) {
-	for _, backend := range config.AcceptedMithrilBackends {
-		if _, err := resolveMithrilBackend(backend); err != nil {
-			t.Errorf(
-				"config accepts mithril.backend %q but "+
-					"resolveMithrilBackend rejects it: %v",
-				backend, err,
-			)
-		}
+func toStringSet(values []string) map[string]struct{} {
+	set := make(map[string]struct{}, len(values))
+	for _, v := range values {
+		set[v] = struct{}{}
 	}
+	return set
 }
--- a/cmd/dingo/main.go
+++ b/cmd/dingo/main.go
@@ -210,12 +210,15 @@ func topLevelCommand(cmd *cobra.Command) *cobra.Command {
 // requirements come from the command rather than cfg.RunMode.
 func effectiveRunMode(top *cobra.Command, cfg *config.Config) config.RunMode {
 	if top == nil {
-		// Bare root command: dispatches on the configured run mode,
-		// defaulting to serve.
-		if cfg.RunMode == "" {
-			return config.RunModeServe
+		// Bare root command dispatches on the configured run mode,
+		// falling through to serve for an empty or unrecognized mode
+		// (matching rootCmd's dispatch default). Serving-listener checks
+		// therefore still apply to an invalid runMode; the invalid mode
+		// itself is reported separately by Validate.
+		if cfg.RunMode.Valid() && cfg.RunMode != "" {
+			return cfg.RunMode
 		}
-		return cfg.RunMode
+		return config.RunModeServe
 	}
 	switch top.Name() {
 	case "serve":
--- a/cmd/dingo/main.go
+++ b/cmd/dingo/main.go
@@ -450,6 +450,12 @@ DSN Override:
 			cfg.ImmutableDbPath = args[0]
 		}
 
+		// Every configuration source is merged at this point (defaults,
+		// YAML, environment, CLI flags, and the load positional
+		// argument), so defaults derived from other settings can now be
+		// filled in and the final configuration validated.
+		cfg.ApplyDefaults()
+
 		if err := cfg.Validate(effectiveRunMode(cmd, cfg)); err != nil {
 			return fmt.Errorf("invalid configuration: %w", err)
 		}
--- a/cmd/dingo/main.go
+++ b/cmd/dingo/main.go
@@ -206,9 +206,10 @@ func topLevelCommand(cmd *cobra.Command) *cobra.Command {
 // effectiveRunMode reports the run mode a command invocation will
 // actually execute, which governs the config it requires. Only the bare
 // `dingo` process honors the configured runMode; each explicit
-// subcommand runs a fixed operation, so its listener/source
-// requirements come from the command rather than cfg.RunMode.
-func effectiveRunMode(top *cobra.Command, cfg *config.Config) config.RunMode {
+// subcommand runs a fixed operation, so its listener/source requirements
+// come from the command (cmd, the leaf being run) rather than cfg.RunMode.
+func effectiveRunMode(cmd *cobra.Command, cfg *config.Config) config.RunMode {
+	top := topLevelCommand(cmd)
 	if top == nil {
 		// Bare root command dispatches on the configured run mode,
 		// falling through to serve for an empty or unrecognized mode
@@ -226,10 +227,16 @@ func effectiveRunMode(top *cobra.Command, cfg *config.Config) config.RunMode {
 	case "load":
 		return config.RunModeLoad
 	case "sync":
+		// `dingo sync [--mithril]` runs the Mithril snapshot sync, which
+		// starts a metrics listener and an optional pprof debug listener.
 		return config.RunModeSync
 	case "mithril":
-		// mithril and its subcommands (sync, list, show): one-shot
-		// utilities that start no serving/API listeners.
+		// Only `mithril sync` binds the metrics/debug listeners; the
+		// read-only `mithril list` / `mithril show` (and bare `mithril`)
+		// query the aggregator and start nothing.
+		if cmd.Name() == "sync" {
+			return config.RunModeSync
+		}
 		return config.RunModeMithril
 	default:
 		// No other non-informational top-level command exists today;
@@ -434,7 +441,7 @@ DSN Override:
 		// version and list are informational and start no services, so
 		// they must still run even when the merged config is invalid.
 		if !isInformationalCommand(top) {
-			if err := cfg.Validate(effectiveRunMode(top, cfg)); err != nil {
+			if err := cfg.Validate(effectiveRunMode(cmd, cfg)); err != nil {
 				return fmt.Errorf("invalid configuration: %w", err)
 			}
 		}
--- a/cmd/dingo/main.go
+++ b/cmd/dingo/main.go
@@ -361,8 +361,14 @@ DSN Override:
 			cfg.ImmutableDbPath = args[0]
 		}
 
-		if err := cfg.Validate(); err != nil {
-			return fmt.Errorf("invalid configuration: %w", err)
+		// version and list are informational and start no services, so
+		// they must still run even when the merged config is invalid.
+		switch cmd.Name() {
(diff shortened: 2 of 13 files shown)


chrisguiney:
Pushed e56ab827 addressing the review feedback:

- **Windows privileged-port check** (CodeRabbit / cubic): `os.Geteuid()` returns `-1` on Windows, so `Validate()` treated every Windows process as unprivileged and rejected sub-1024 ports. `Validate()` now treats Windows as privileged (`runtime.GOOS == "windows" || os.Geteuid() == 0`); the restriction is Unix-specific.
- **Listener ports in load mode** (CodeRabbit / cubic): relay/private/metrics are now required only when `RunMode != RunModeLoad`, since `dingo load` starts none of those listeners. A load-only config may leave them at 0. Serving modes still require them.
- **Utility subcommands** (cubic): `Validate()` is skipped for the informational `version` and `list` subcommands so they still run against an otherwise-invalid config.
- **Duplicate port assignments** (CodeRabbit nitpick): added a `map[uint]string` pass that flags two services sharing the same non-zero port (previously only failed at bind time).
- **Whitelist drift** (CodeRabbit nitpick): the `chainsync.strategy` and `mithril.backend` accepted-value sets are now exported slices, consumed by `Validate()` and checked by a new parity test in `cmd/dingo` (which can import both `chainsync` and `resolveMithrilBackend`), so they can't silently drift from the real parsers.

New test cases cover the load-mode/serve-mode port gating and duplicate-port detection; the parity tests cover the whitelists. `go build ./...`, the config + cmd/dingo suites, import-boundaries, golangci-lint, and modernize all pass locally.

**On the Windows `go-test` failure:** it was infra, not a code hang. The job annotation is GitHub's generic *"The hosted runner lost communication with the server"*, the `go-test` step never completed (it was killed at the ~6h job timeout), and the identical `go test ./...` passed on macOS in ~2.5 min. All changed code paths in this PR are pure, fast functions with no goroutines/IO, and `main` is consistently green on Windows. This push re-runs the Windows job.

cmd/dingo/config_parity_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/blinklabs-io/dingo/internal/config"
+)
+
+// TestChainsyncStrategyWhitelistParity guards against drift between the
+// chainsync.strategy values internal/config.Validate accepts and those
+// chainsync.ParseHeaderSyncStrategy actually accepts. internal/config
+// cannot import chainsync without pulling node subsystems into the
+// config package, so this parity check lives in cmd/dingo, which can
+// import both.
+func TestChainsyncStrategyWhitelistParity(t *testing.T) {
```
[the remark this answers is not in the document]
chrisguiney: Addressed in 35bc20e4. Both parity tests now assert set-equality between config's accepted list and the parser's canonical contract (plus that every canonical value actually parses), so drift is caught in both directions — including a value added to `ParseHeaderSyncStrategy`/`resolveMithrilBackend` (and the canonical contract) without updating config's list. The mithril canonical set is anchored to the real `mithril.BackendV1`/`BackendV2` constants.

cmd/dingo/main.go
```diff
@@ -357,6 +353,24 @@ DSN Override:
 			return fmt.Errorf("applying CLI flags: %w", err)
 		}
 
+		// `dingo load <path>`: the positional argument is the
+		// highest-precedence source for ImmutableDbPath; merge it before
+		// validation so a config with runMode "load" and no
+		// immutableDbPath doesn't fail spuriously.
+		if cmd.Name() == "load" && len(args) > 0 {
```
wolf31o2: This only copies the positional path into ImmutableDbPath; it does not make validation treat the explicit `dingo load <path>` command as load mode. With the default `runMode: serve`, `cfg.Validate()` still requires relay/private/metrics ports even though `loadRun` does not start those listeners. Repro: `go run ./cmd/dingo load /tmp --port 0 --private-port 0 --metrics-port 0` fails in validation before reaching the load command. A straightforward fix would be to normalize an effective run mode for this command before validation, or add a distinct run mode/validation profile for one-shot load commands.

internal/config/validate.go
```diff
@@ -0,0 +1,318 @@
...
+		errs = append(errs, errors.New(
+			"runMode \"load\" requires immutableDbPath to be set "+
+				"(config, DINGO_IMMUTABLE_DB_PATH, or --immutable-db-path)",
+		))
+	}
+
+	// Ports. The relay, private, and metrics listeners are required for
+	// the serving modes but not started by "load", so they may be unset
+	// (0) in a load-only config.
+	listenersRequired := c.RunMode != RunModeLoad
```
wolf31o2: This conflates configured `runMode` with the actual Cobra command/service set being executed. Non-serving commands like `dingo load <path>` and `dingo mithril sync` can run with the default `runMode: serve`, but they do not start relay/private listeners, so listener-required validation rejects configs that are valid for those commands. This probably needs command/service-set aware validation, or additional/effective run modes so `Validate()` can determine which listeners are actually required.

wolf31o2 (changes requested):
No-go for merge until the command-specific validation regression is fixed. The new validator applies serving-mode listener requirements based on cfg.RunMode, but several one-shot commands do not start those listeners and can still have the default runMode=serve. It may be cleanest to model this by normalizing an effective run mode before validation, or by adding command-specific run modes for load/sync-style invocations so Validate() can decide which listener set is actually required.

chrisguiney:
@wolf31o2 fixed in 962a95b8 — thanks for catching this.

You're right: the listener requirement keyed off `cfg.RunMode`, but the one-shot subcommands (`load`, `sync`, `mithril`) run a fixed operation and keep the default `runMode=serve`, so they were wrongly required to have the relay/private/metrics ports set.

Following the first of your suggestions (normalize an effective run mode before validation), `cmd/dingo` now derives an **effective run mode** from the invoked command and passes it to `Validate()`:

- bare `dingo` → honors the configured `runMode` (defaults to serve)
- `serve` subcommand → serving (requires listeners), regardless of configured `runMode`
- `load` → requires an ImmutableDB source, no listeners
- `sync`, `mithril` (and nested subcommands) → new effective-only `RunModeUtility`: neither listeners nor a source

`RunMode.RequiresListeners()` centralizes the serving-mode set. The `version`/`list` exemption and the `load` positional-arg merge now key off the **top-level** command, so nested commands like `dingo mithril list` are classified correctly (previously `cmd.Name()` collided with top-level `list`).

Verified end-to-end against the built binary:
- `dingo sync` / `dingo mithril list` with `relayPort/privatePort/metricsPort: 0` → pass validation and run (no "invalid configuration")
- `dingo serve` with those zeroed → still fails with the listener-port errors
- `dingo` (runMode=load, no `immutableDbPath`) → still fails "requires immutableDbPath"
- `dingo version` → still prints regardless of config

New unit tests cover `effectiveRunMode`/`isInformationalCommand` across the command tree and the utility-mode relaxation in the validator. ARCHITECTURE.md updated. golangci-lint, modernize, import-boundaries, and the config + cmd/dingo suites pass locally.

cmd/dingo/main.go
```diff
@@ -189,6 +189,61 @@ func listAllPlugins() string {
...
+// subcommand runs a fixed operation, so its listener/source
+// requirements come from the command rather than cfg.RunMode.
+func effectiveRunMode(top *cobra.Command, cfg *config.Config) config.RunMode {
+	if top == nil {
+		// Bare root command: dispatches on the configured run mode,
+		// defaulting to serve.
+		if cfg.RunMode == "" {
+			return config.RunModeServe
+		}
+		return cfg.RunMode
```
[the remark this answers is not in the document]
chrisguiney: Addressed in f635813d — `effectiveRunMode` now falls back to `serve` for an empty or unrecognized configured `runMode` at the bare root, matching `rootCmd`'s dispatch default, so serving-listener violations are aggregated rather than suppressed.

Note: in practice an invalid `runMode` is already rejected before `Validate()` runs — `LoadConfig` rejects it from YAML/env, and `normalizeRunMode` rejects the `--run-mode` flag during `ApplyFlags` — so this path isn't reachable end-to-end. The change is still worth making to keep the validator self-consistent with the dispatch, and it's covered by unit tests (`TestEffectiveRunMode/bare_invalid_falls_back_to_serve` and `TestValidateInvalidModeStillReportsListeners`).

internal/config/validate.go
```diff
@@ -0,0 +1,325 @@
...
+	ports := []struct {
+		setting  string
+		port     uint
+		required bool
+	}{
+		{"port (relay/NtN)", c.RelayPort, listenersRequired},
+		{"privatePort", c.PrivatePort, listenersRequired},
+		{"metricsPort", c.MetricsPort, listenersRequired},
+		{"debugPort", c.DebugPort, false},
+		{"utxorpcPort", c.UtxorpcPort, false},
```
wolf31o2: This port set is broader than the listeners that can actually bind for a given invocation. For example, `dingo --metrics-port 8080 sync` is rejected because `metricsPort` matches the default `meshPort`, but `sync` is classified as a utility and Mesh is not started there. The same shape exists for core-mode serving: UTxORPC/Blockfrost/Mesh/Midnight are gated on API storage mode, so their configured ports should not collide with active listeners when those APIs are inactive. Please derive the validated/deduped port set from the effective invocation mode plus storage mode, so we only reject ports that could actually bind together.

cmd/dingo/main.go
```diff
@@ -189,6 +189,64 @@ func listAllPlugins() string {
...
+			return cfg.RunMode
+		}
+		return config.RunModeServe
+	}
+	switch top.Name() {
+	case "serve":
+		return config.RunModeServe
+	case "load":
+		return config.RunModeLoad
+	default:
```
wolf31o2, line 241: The catch-all `RunModeUtility` is probably too coarse once validation depends on listener surfaces. Please consider creating distinct effective run modes or listener profiles for `load`, `sync`, `mithril`, and serving modes instead of funneling all non-serve/load commands through one utility mode. That keeps `load` and `sync` from inheriting validations for listeners they do not start, while still allowing commands that do start metrics/debug or future utility listeners to validate those ports explicitly.

wolf31o2 (changes requested):
Requesting changes for the listener validation regression. The new validation should only check ports for listeners that the effective invocation can actually start. Right now utility commands and core-mode serving configs can be rejected because they collide with configured API ports that are inactive for that path.

cmd/dingo/main.go
```diff
@@ -189,6 +189,71 @@ func listAllPlugins() string {
...
+		return config.RunModeServe
+	}
+	switch top.Name() {
+	case "serve":
+		return config.RunModeServe
+	case "load":
+		return config.RunModeLoad
+	case "sync":
+		return config.RunModeSync
+	case "mithril":
```
wolf31o2, line 233: This maps every `mithril` subcommand to the mode that validates auxiliary listeners. `mithril sync` does start metrics/debug, but `mithril list` and `mithril show` only query the aggregator and never bind those ports. Repro on this branch: `go run ./cmd/dingo --metrics-port 99999999 mithril list` fails during config validation with `invalid metricsPort`, even though that command does not use the metrics listener. Please distinguish `mithril sync` from the read-only Mithril subcommands before validation.

wolf31o2 (changes requested):
NO-GO: startup validation blocks `mithril list` / `mithril show` on metrics/debug port settings they do not use. Please split effective validation for Mithril read-only subcommands from Mithril sync, or skip auxiliary listener validation for list/show, and add a command-mode regression test.

internal/config/validate.go
```diff
@@ -0,0 +1,412 @@
...
+
+// validatePathNoTraversal rejects paths containing a ".." component.
+// Values can arrive via YAML or environment, where a traversal-shaped
+// path is more likely an injection than an intent; absolute paths
+// express any legitimate target without "..".
+func validatePathNoTraversal(setting, path string) error {
+	if path == "" {
+		return nil
+	}
+	cleaned := filepath.Clean(path)
```
wolf31o2: Blocking: cleaning the path before checking it removes the traversal component being guarded against. For example, `configs/../secret.json` becomes `secret.json` and passes validation; the local CLI repro reports only an unrelated forced port error. The issue and ARCHITECTURE.md both promise no `..` components. Inspect the original path components before cleaning (with appropriate platform separator handling), and change the test that currently accepts an inner dotdot.

internal/config/validate.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+
+	// Mempool
+	if c.MempoolCapacity < 0 {
+		errs = append(errs, fmt.Errorf(
+			"invalid mempoolCapacity: %d (must not be negative)",
+			c.MempoolCapacity,
+		))
+	}
+	if c.EvictionWatermark <= 0 || c.EvictionWatermark >= 1.0 {
```
wolf31o2: Blocking: these comparisons accept NaN because every ordered comparison with NaN is false. `--eviction-watermark NaN` passes this validator and then reaches mempool threshold conversion/arithmetic. Reject non-finite values explicitly (or express the range check so NaN cannot satisfy it) for both watermarks, and add CLI/config regression coverage.

cmd/dingo/main.go
```diff
@@ -357,6 +431,24 @@ DSN Override:
...
+		// highest-precedence source for ImmutableDbPath; merge it before
+		// validation so a config with runMode "load" and no
+		// immutableDbPath doesn't fail spuriously.
+		if top != nil && top.Name() == "load" && len(args) > 0 {
+			cfg.ImmutableDbPath = args[0]
+		}
+
+		// version and list are informational and start no services, so
+		// they must still run even when the merged config is invalid.
+		if !isInformationalCommand(top) {
```
wolf31o2: Blocking: this exemption occurs after `LoadConfig`, so informational commands still fail on configuration loading/validation errors. Repro: `dingo --config /tmp/does-not-exist version` exits with “failed to load config” instead of printing the version. Classify and bypass runtime config loading before `LoadConfig` for version/list/help/completion, or narrow any loading to data those commands actually require.

internal/config/validate.go
```diff
@@ -0,0 +1,412 @@
...
+		if len(missing) > 0 {
+			errs = append(errs, fmt.Errorf(
+				"blockProducer enabled but missing required key paths: %v",
+				missing,
+			))
+		}
+	}
+
+	// Network identity
+	if c.Network == "" {
```
wolf31o2, line 315: Blocking integration mismatch: this branch says a custom `networkMagic` is sufficient when `network` is empty, but startup cannot reach it. `LoadConfig` still unconditionally calls `ValidateNetworkName("")`, and `ApplyFlags` rejects `--network=` before `Validate`. A YAML config with `network: ""` and `networkMagic: 42` fails in `LoadConfig`. Remove or reconcile the earlier fail-fast validation so the advertised contract and aggregation actually apply.

internal/config/validate.go
```diff
@@ -0,0 +1,412 @@
...
+		{"debugPort", c.DebugPort, auxListeners, false},
+		{"barkPort", c.BarkPort, serving, false},
+		{"utxorpcPort", c.UtxorpcPort, apiListeners, false},
+		{"blockfrostPort", c.BlockfrostPort, apiListeners, false},
+		{"meshPort", c.MeshPort, apiListeners, false},
+		{"midnight.port", c.Midnight.Port, apiListeners, false},
+	}
+	// Two active listeners sharing a port only fails at bind time; catch it
+	// here. Zero ports are disabled or OS-assigned, so they don't clash.
+	seenPorts := make(map[uint]string, len(ports))
```
wolf31o2: Blocking false positive: port number alone is not a listener identity. Active listeners can legally share a port on distinct specific bind addresses, but this map rejects them. For example, debug on `127.0.0.1:13000` and Midnight on `127.0.0.2:13000` are bindable together yet are reported as a collision. Include the effective bind address (and wildcard overlap semantics) in collision detection, or leave bind conflicts to the existing synchronous listener startup paths.

wolf31o2 (changes requested):
No-go for merge on the current head. The earlier command-specific listener findings are addressed, and the merge-result tests and linters are clean, but the inline validation regressions below remain blocking.

chrisguiney:
@coderabbitai review

cmd/dingo/main.go
```diff
@@ -357,6 +442,18 @@ DSN Override:
...
 
+		// `dingo load <path>`: the positional argument is the
+		// highest-precedence source for ImmutableDbPath; merge it before
+		// validation so a config with runMode "load" and no
+		// immutableDbPath doesn't fail spuriously.
+		if top != nil && top.Name() == "load" && len(args) > 0 {
+			cfg.ImmutableDbPath = args[0]
+		}
+
+		if err := cfg.Validate(effectiveRunMode(cmd, cfg)); err != nil {
```
wolf31o2, line 459: This validation call is too late to provide the advertised fully merged validation. LoadConfig has already rejected invalid runMode/startEra, block-producer credentials, and watermarks, and has already selected mode-dependent defaults before ApplyFlags. Confirmed repro: YAML runMode: bogus plus higher-precedence --run-mode serve fails inside LoadConfig; --run-mode leios can also retain the Praos mempool default selected earlier. Make LoadConfig parse/merge only, apply CLI flags, derive defaults from the final mode, then call Validate. Please add precedence, defaulting, and aggregation regression tests.

wolf31o2 (changes requested):
Requesting changes on the current head. The earlier command-specific listener/path/NaN/networkMagic/bind-address findings are fixed, and the focused and conformance tests pass, but the startup flow still violates configuration precedence: LoadConfig performs semantic checks and mode-dependent defaulting before ApplyFlags. Please defer those operations until after flags, then validate/finalize the fully merged configuration. DevNet remains to be rerun after the fix.

cmd/dingo/main.go
chrisguiney: Fixed in cb45d3ba. LoadConfig now only parses and merges the YAML and environment sources — the runMode/startEra/watermark/block-producer fail-fasts and the mode-dependent MempoolCapacity defaulting are gone from it. A new Config.ApplyDefaults derives defaults from the fully merged configuration, and cmd/dingo calls it after ApplyFlags (and the load positional-arg merge), before Validate. Regression tests added in internal/config/flags_test.go run the full pipeline: an invalid YAML runMode overridden by --run-mode serve now validates cleanly (and is still rejected without the override), --run-mode leios gets the Leios mempool default, and multiple bad settings aggregate into a single Validate error.

chrisguiney:
DevNet rerun on cb45d3ba (`internal/test/devnet/run-tests.sh`): 9 of 10 scenarios pass, including TestSustainedConsensus and TestEpochBoundaryConsensus. TestChainGrowthRate fails (Dingo forged 0 blocks in the 100-slot window), but the same failure reproduces in this environment on the previously reviewed head ef87d7f8 (0 blocks) and on the upstream/main merge-base 7aa64cd1 (1 block), so it is pre-existing and not attributable to this PR. Unit, conformance, and lint runs on the merged result are clean.

internal/config/config.go
```diff
@@ -1063,6 +1018,46 @@ func LoadConfig(configFile string) (*Config, error) {
...
+	if c.RejectionWatermark == 0 {
+		c.RejectionWatermark = DefaultRejectionWatermark
+	}
+	if c.ForgeSyncToleranceSlots == 0 {
+		c.ForgeSyncToleranceSlots = DefaultForgeSyncToleranceSlots
+	}
+	if c.ForgeStaleGapThresholdSlots == 0 {
+		c.ForgeStaleGapThresholdSlots = DefaultForgeStaleGapThresholdSlots
+	}
+	if c.HistoryExpiry.Frequency <= 0 {
```
wolf31o2: This condition conflates an unset zero value with an explicitly negative duration. For example, `--history-expiry-enabled --history-expiry-frequency=-1s` reaches this code, is silently rewritten to `1h`, passes validation, and starts the history-expiry worker. The resulting failure mode is particularly unsafe: instead of failing startup (or disabling expiry), the node periodically replaces stable historical block CBOR with expiry markers. Default only `Frequency == 0`; preserve negative values for `Validate` to reject.

internal/config/validate.go
```diff
@@ -0,0 +1,465 @@
...
+				"invalid %s %q: %w", d.setting, d.value, err,
+			))
+			continue
+		}
+		if d.mustPositive && parsed <= 0 {
+			errs = append(errs, fmt.Errorf(
+				"invalid %s %q: must be positive", d.setting, d.value,
+			))
+		}
+	}
```
wolf31o2, line 353: Please validate `historyExpiry.frequency` as positive after defaults are applied. Currently ApplyDefaults erases a negative value before this validator runs, so the configuration error cannot be reported or aggregated. After changing defaulting to handle only zero, append an error when the frequency is negative (or non-positive if validation may be called without ApplyDefaults). Add regressions for YAML `historyExpiry.frequency: -1s`, `DINGO_HISTORY_EXPIRY_FREQUENCY=-1s`, and `--history-expiry-frequency=-1s`, especially with history expiry enabled, and assert startup fails rather than using the one-hour default.

wolf31o2 (changes requested):
Requesting changes on the current head. The prior CLI-precedence and mode-dependent-defaulting blocker is fixed, the branch merges cleanly with current main, and the focused race-enabled and conformance tests pass. One configuration safety issue remains: an explicitly negative history-expiry frequency is silently converted to the one-hour default, so a node can start pruning when the operator supplied a value they may reasonably expect to disable expiry or fail startup. Please preserve the negative value through defaulting, reject it in Validate, and cover YAML, environment, and CLI inputs. DevNet remains to be rerun after the fix.

internal/config/config.go
chrisguiney: Fixed in b61c2340: ApplyDefaults now only fills in an unset (zero) frequency, so an explicitly negative value survives defaulting for Validate to reject. Repro now fails startup: `dingo --history-expiry-enabled --history-expiry-frequency=-1s ...` → `invalid configuration: invalid historyExpiry.frequency: -1s (must be positive)`. The pruner's own `Frequency <= 0` guard in internal/historyexpiry remains as defense in depth for library callers that bypass Validate.

internal/config/validate.go
chrisguiney: Done in b61c2340: Validate rejects a non-positive historyExpiry.frequency (non-positive rather than just negative, so a caller that skips ApplyDefaults is also caught), aggregated with the other errors. Regression tests cover all three sources with expiry enabled — YAML `historyExpiry.frequency: -1s`, `DINGO_HISTORY_EXPIRY_FREQUENCY=-1s`, and `--history-expiry-frequency=-1s` — via the full LoadConfig→ApplyFlags→ApplyDefaults→Validate pipeline in flags_test.go, asserting validation fails and the negative value is preserved through defaulting.

chrisguiney:
DevNet rerun on b61c2340: 8 of 10 scenarios pass, including TestSustainedConsensus and TestEpochBoundaryConsensus. TestChainGrowthRate fails with the same 0-blocks signature that reproduces on the upstream/main merge-base in this environment (see earlier comment), and TestBasicBlockForging failed this round with the same underlying forging stall (it passed on the previous run of cb45d3ba). Both match the pre-existing local block-forging issue that is independent of this PR. Unit, conformance (via `make test -race`, 52 packages, 0 failures), and lint runs on b61c2340 are clean.

internal/config/config.go
```diff
@@ -956,99 +992,18 @@ func LoadConfig(configFile string) (*Config, error) {
...
-	}
-	if globalConfig.HistoryExpiry.Frequency <= 0 {
-		globalConfig.HistoryExpiry.Frequency = time.Hour
-	}
-
-	// Validate network name to prevent path traversal (INT-03).
-	if err := ValidateNetworkName(globalConfig.Network); err != nil {
-		return nil, err
-	}
+	// LoadConfig only parses and merges configuration sources; it makes
```
wolf31o2, line 995: [P1] `LoadConfig` still performs topology resolution below this comment (`LoadTopologyConfig` at line 1014), so it does more than parse and merge. This can reject a lower-precedence value before CLI flags are applied: I reproduced `network: ../bad` failing here with `unknown network` even when the invocation intends to supply `--network preview`. Remove topology loading from `LoadConfig` so a CLI override can repair YAML/env values as promised.

cmd/dingo/main.go
```diff
@@ -357,6 +442,24 @@ DSN Override:
...
 
+		// `dingo load <path>`: the positional argument is the
+		// highest-precedence source for ImmutableDbPath; merge it before
+		// validation so a config with runMode "load" and no
+		// immutableDbPath doesn't fail spuriously.
+		if top != nil && top.Name() == "load" && len(args) > 0 {
+			cfg.ImmutableDbPath = args[0]
+		}
+
+		// Every configuration source is merged at this point (defaults,
```
wolf31o2, line 453: [P1] This is not yet the first point at which every configuration source has safely merged: both `LoadConfig` and `ApplyFlags` resolve topology before `ApplyDefaults`/`Validate`. I also reproduced a missing YAML topology file aborting before a valid `--topology` override can take effect. Resolve/load topology once after flags, defaults, and validation, and cover both the `--network` and `--topology` override cases.

wolf31o2 (changes requested):
No-go for merge on the current head. The previously reported history-expiry issue is fixed, the branch merges cleanly with current main, and focused race, conformance, and import-boundary checks pass. However, topology is still loaded before CLI overrides and before the final aggregated validation pass, so the startup flow continues to violate CLI > env > YAML precedence for `network` and `topology`. Please defer topology resolution until the fully merged configuration has been defaulted and validated, and add regression coverage for CLI overrides of invalid YAML values.

internal/config/config.go
chrisguiney: Fixed in cfc226f7: LoadConfig no longer resolves topology (or anything else derived from mergeable settings) — it parses and merges YAML/env only. Your repro now works: YAML `network: ../bad` with `--network preview` proceeds past config loading, validation, and topology resolution; without the override the value is rejected by Validate (`invalid network name "../bad"`), aggregated with any other errors.

cmd/dingo/main.go
chrisguiney: Done in cfc226f7: the ApplyFlags topology reload is removed too, and topology is resolved exactly once in cmd/dingo after ApplyFlags → ApplyDefaults → Validate. Both override cases are covered by pipeline regression tests in flags_test.go (TestPipeline_NetworkFlagRepairsInvalidYAMLNetwork, TestPipeline_TopologyFlagRepairsMissingYAMLTopology), each also asserting the no-override invocation still fails — the bad network in Validate, the missing topology file at topology resolution. Verified against a real binary: `--config badtopo.yaml --topology topo.json load ...` starts, and without the flag fails with `loading topology: failed to load topology file: ...`.

chrisguiney:
DevNet rerun on cfc226f7: all three nodes come up healthy with the relocated topology resolution (Dingo loads its explicit topology file via the new post-validation path), and 8 of 10 scenarios pass including TestSustainedConsensus and TestEpochBoundaryConsensus. TestChainGrowthRate (2 blocks) and TestBasicBlockForging fail with the same forging-stall signature previously reproduced on the upstream/main merge-base in this environment — pre-existing, not from this PR. Full `make test` (-race, 52 packages including conformance), golangci-lint, modernize, and import-boundaries are clean on cfc226f7.

wolf31o2 (approved):

PR #2764, opened 2026-07-07
feat(mempool): make mempool eviction optional

### LINKED ISSUE #2251: Make mempool eviction optional
The mempool implementation should make this (and the eviction thresholds) configurable. By default, there should *not* be eviction. Instead, the local transaction submission should fail, and the node to node tx submission shouldn't ask for transactions which won't fit in its mempool. There shouldn't be a case where it can ever overflow. A full mempool provides network backpressure. It's part of the design of the protocol. There are cases where eviction is useful, which is why we should keep it optional, though disabled in our normal configurations.

jkawan:
closes #2251 

---
## Summary by cubic
Makes mempool eviction optional and disabled by default. When disabled, the node stops admitting new transactions at full capacity and applies backpressure instead of evicting older ones.

- Config and validation: allow `evictionWatermark: 0` (disable) or (0,1); `rejectionWatermark` in (0,1] with default 1.0; enforce eviction < rejection only when eviction is enabled; `ApplyDefaults` preserves an explicit zero eviction and defaults rejection when unset; docs and `dingo.yaml.example` now default to `evictionWatermark: 0`, `rejectionWatermark: 1.0`.
- `mempool`: FIFO evicts only when `evictionWatermark > 0`; both FIFO and DAG expose `AdmissionHeadroomBytes()`, `MaxAdmissionHeadroomBytes()`, and `WaitForAdmissionHeadroom(...)`; admission rejection uses headroom; tests cover disabled eviction and headroom reporting.
- `ouroboros/txsubmission`: requests only the contiguous prefix that fits current headroom, waits for capacity, and backs off when a peer offers a TX larger than max admission; removes an obsolete prefix batching test.

- Migration: If you rely on eviction, set `--eviction-watermark 0.90` (or your value) to restore previous behavior.

<sup>Written for commit 82d526a60086fc10f9cfe4d19ea66b5a66a17864. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2764?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added mempool tuning settings for eviction and rejection thresholds.
  * Added a way to check remaining admission headroom and wait for capacity to open up.

* **Bug Fixes**
  * Mempool eviction can now be disabled by setting the eviction threshold to `0`.
  * Improved transaction submission so oversized transactions are skipped when capacity is tight, instead of requesting them anyway.
  * Updated default watermark behavior to use no eviction and full rejection threshold by default.


ouroboros/txsubmission.go
```diff
@@ -259,11 +268,54 @@ func (o *Ouroboros) txsubmissionServerInit(
...
+							)
+							backoffTimer.Reset(wait)
+							select {
+							case <-backoffTimer.C:
+							case <-conn.ErrorChan():
+								return
+							}
+						} else {
+							consecutiveImpossibleOffers = 0
+							minHeadroomBytes = nextHeadroomBytes
```
wolf31o2: This records the minimum headroom, but the offered tx ids have already been consumed by `RequestTxIds`. `gouroboros` sets the next ack count to `len(result.txIds)`, so the next loop acknowledges this offer before asking for more ids. A tx that is supposedly deferred can therefore be dropped for this peer instead of requested once headroom opens. The loop needs to avoid consuming the offer until it can request it, or retain/retry the same offer without acknowledging it away.

ouroboros/txsubmission.go
```diff
@@ -331,6 +383,33 @@ func (o *Ouroboros) txsubmissionServerInit(
...
+) ([]txsubmission.TxId, int64) {
+	if availableBytes <= 0 {
+		return nil, 0
+	}
+	ret := make([]txsubmission.TxId, 0, len(txIds))
+	var requestedBytes int64
+	var nextHeadroomBytes int64
+	for _, txId := range txIds {
+		txSize := int64(txId.Size)
+		if txSize > availableBytes || requestedBytes+txSize > availableBytes {
```
wolf31o2: Skipping over an earlier offered tx and requesting a later smaller one can break valid dependent tx chains. The later tx may spend outputs created by the skipped parent, then `AddTransaction` validation fails and the tx-submission goroutine returns. This filtering should preserve offered order/prefix semantics, or otherwise ensure dependencies are requested with their children.

wolf31o2 (changes requested):
NO-GO for merge. The local targeted tests pass, but the tx-submission backpressure changes can drop deferred offers and can request later txs without earlier dependencies. Please fix these behaviors before merge.

internal/config/config.go
```diff
@@ -80,8 +80,8 @@ func FromContext(ctx context.Context) *Config {
 const (
 	DefaultBlobPlugin                  = "badger"
 	DefaultMetadataPlugin              = "sqlite"
-	DefaultEvictionWatermark           = 0.90
-	DefaultRejectionWatermark          = 0.95
+	DefaultEvictionWatermark           = 0.8
```
wolf31o2, line 83: This default makes the documented disable value ineffective. An explicit evictionWatermark of 0 from YAML, environment, or CLI reaches ApplyDefaults(), which treats zero as unset and replaces it with DefaultEvictionWatermark (0.8). cmd/dingo always calls ApplyDefaults after merging all sources, so operators cannot disable eviction and the documented default of 0 is actually 0.8. Please preserve explicit zero through the full configuration pipeline (for example by tracking presence separately) and add a test that applies flags/defaults/validation with an explicit zero.

ouroboros/txsubmission.go
```diff
@@ -259,78 +268,184 @@ func (o *Ouroboros) txsubmissionServerInit(
...
+									),
+									"component", "network",
+									"protocol", "tx-submission",
+									"role", "server",
+									"connection_id", ctx.ConnectionId.String(),
+								)
+								return
+							}
+						}
+						// Successfully processed this offer; break inner loop
```
wolf31o2, line 359: Breaking here after requesting only the fitting prefix loses the remainder of this offer. gouroboros RequestTxIds records ackCount as len(result.txIds), so the next RequestTxIds call acknowledges the complete offer, while the remote MempoolConsumer has already advanced past every offered ID and clears its cache on that acknowledgement. For an offer sized [100,150,75] with 300 bytes of headroom, the first two bodies are requested but all three IDs are acknowledged, so the 75-byte transaction is never reoffered on this connection. Please retain/partially acknowledge the pending offer or constrain offers so every acknowledged ID is processed, and cover this with an end-to-end relay regression test.

wolf31o2 (changes requested):
Requesting changes for two merge-blocking correctness issues: eviction cannot be disabled through the normal configuration pipeline, and partially requested TxSubmission offers acknowledge and discard unrequested transaction IDs. Please address both and add full-pipeline regression coverage. Because this changes mempool and TxSubmission behavior, the required DevNet validation should also be run after the fixes.

wolf31o2 (approved):

PR #2778, opened 2026-07-08
feat(midnight): implement paginated UTxO query RPCs for MidnightState

### LINKED ISSUE #2118: Implement paginated UTxO query RPCs for MidnightState
Implement the five UTxO-event RPCs in MidnightState.

1. GetAssetCreates - paginated cNIGHT UTxO creations from a (block, tx_index) cursor.
2. GetAssetSpends - paginated cNIGHT UTxO spends from a (block, tx_index) cursor.
3. GetRegistrations - paginated registration events from a (block, tx_index) cursor.
4. GetDeregistrations - paginated deregistration events from a (block, tx_index) cursor.
5. GetUtxoEvents - merged stream of all four event types from a single cursor.

Pagination semantics for the four type-specific RPCs:
- WHERE block_number > start_block OR (block_number = start_block AND tx_index > start_tx_index)
- ORDER BY block_number ASC, tx_index ASC
- LIMIT utxo_capacity from the request

GetUtxoEvents merge logic:
- Query all four tables with the same cursor window.
- Merge-sort by (block_number, tx_index, kind_order) where kind_order is create=0, spend=1, registration=2, deregistration=3.
- Apply tx_capacity limit from the request.
- Return next_position cursor for continuation.
- If end_block_hash is set on the request, stop at that block.

Tests:
- Empty database returns empty results without error.
- Cursor advancement returns no duplicates and no gaps.
- GetUtxoEvents ordering matches the kind_order tie-break across all four event types.

cryptodj413:
Closes #2118

---
## Summary by cubic
Adds cursor‑paginated UTxO event query RPCs for MidnightState with stable ordering, consistent snapshot reads, and atomic indexer writes. Pages are bounded, extend to full transaction groups, and enforce `end_block_hash` via a block‑hash resolver.

- **New Features**
  - gRPC handlers in `midnight/server` for `GetAssetCreates`, `GetAssetSpends`, `GetRegistrations`, `GetDeregistrations`, and merge‑sorted `GetUtxoEvents` (create < spend < registration < deregistration) with `next_position`.
  - Forward pagination from `(start_block, start_tx_index)`; server defaults/clamps `utxo_capacity`/`tx_capacity`; deterministic ordering (`ORDER BY id`); `limit <= 0` skips SQL LIMIT. Shared pagination helpers and model `BlockTxPosition` methods unify backend behavior.
  - `GetUtxoEvents` enforces `end_block_hash` via `BlockNumberByHash` (returns `FailedPrecondition` if unset). Server uses `Config.Metadata`; `node.go` wires `n.db.Metadata()` and the resolver.

- **Bug Fixes**
  - Read consistency: read-only transactions use repeatable‑read on MySQL/Postgres; SQLite uses WAL snapshots.
  - Prevented in‑memory/DB divergence: epoch resolution/advance moved inside block processing so writes, snapshots, and indexer state update atomically.

<sup>Written for commit 150f2e63b5ede1934dad88b1ab8a50c7f71e5b06. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2778?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added Midnight gRPC query support for asset creates, asset spends, registrations, deregistrations, and combined UTxO events.
  * UTxO event responses now support stable cursor-based pagination and resume tokens for continued fetching.

* **Bug Fixes**
  * Improved ordering and paging consistency across Midnight event queries, including tie-breaking when multiple events share the same block and transaction position.
  * Added handling for empty results and no-limit requests to return complete data when appropriate.

* **Documentation**
  * Updated architecture and database docs to reflect the new Midnight metadata-backed query behavior.

### DIFF
--- a/database/plugin/metadata/sqlite/midnight.go
+++ b/database/plugin/metadata/sqlite/midnight.go
@@ -202,6 +202,114 @@ func (d *MetadataStoreSqlite) DeleteMidnightDeregistrationsByBlock(
 	return rows, nil
 }
 
+// FindMidnightAssetCreatesFrom returns cNIGHT create rows ordered by
+// (block_number, tx_index) ascending, starting strictly after
+// (startBlock, startTxIndex). limit <= 0 means no SQL LIMIT is applied.
+func (d *MetadataStoreSqlite) FindMidnightAssetCreatesFrom(
+	startBlock uint64,
+	startTxIndex uint32,
+	limit int,
+	txn types.Txn,
+) ([]models.MidnightAssetCreate, error) {
+	db, err := d.resolveReadDB(txn)
+	if err != nil {
+		return nil, err
+	}
+	query := db.Where(
+		"(block_number > ?) OR (block_number = ? AND tx_index > ?)",
+		startBlock, startBlock, startTxIndex,
+	).Order("block_number ASC, tx_index ASC")
+	if limit > 0 {
+		query = query.Limit(limit)
+	}
+	var rows []models.MidnightAssetCreate
+	if err := query.Find(&rows).Error; err != nil {
+		return nil, err
+	}
+	return rows, nil
+}
+
+// FindMidnightAssetSpendsFrom returns cNIGHT spend rows ordered by
+// (block_number, tx_index) ascending, starting strictly after
+// (startBlock, startTxIndex). limit <= 0 means no SQL LIMIT is applied.
+func (d *MetadataStoreSqlite) FindMidnightAssetSpendsFrom(
+	startBlock uint64,
+	startTxIndex uint32,
+	limit int,
+	txn types.Txn,
+) ([]models.MidnightAssetSpend, error) {
+	db, err := d.resolveReadDB(txn)
+	if err != nil {
+		return nil, err
+	}
+	query := db.Where(
+		"(block_number > ?) OR (block_number = ? AND tx_index > ?)",
+		startBlock, startBlock, startTxIndex,
+	).Order("block_number ASC, tx_index ASC")
+	if limit > 0 {
+		query = query.Limit(limit)
+	}
+	var rows []models.MidnightAssetSpend
+	if err := query.Find(&rows).Error; err != nil {
+		return nil, err
+	}
+	return rows, nil
+}
+
+// FindMidnightRegistrationsFrom returns registration rows ordered by
+// (block_number, tx_index) ascending, starting strictly after
+// (startBlock, startTxIndex). limit <= 0 means no SQL LIMIT is applied.
+func (d *MetadataStoreSqlite) FindMidnightRegistrationsFrom(
+	startBlock uint64,
+	startTxIndex uint32,
+	limit int,
+	txn types.Txn,
+) ([]models.MidnightRegistration, error) {
+	db, err := d.resolveReadDB(txn)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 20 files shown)


cryptodj413:
@cubic-dev-ai review

database/plugin/metadata/sqlite/midnight.go
```diff
@@ -202,6 +203,129 @@ func (d *MetadataStoreSqlite) DeleteMidnightDeregistrationsByBlock(
...
+	txn types.Txn,
+) ([]models.MidnightAssetCreate, error) {
+	db, err := d.resolveReadDB(txn)
+	if err != nil {
+		return nil, err
+	}
+	query := db.Where(
+		"(block_number > ?) OR (block_number = ? AND tx_index > ?)",
+		startBlock, startBlock, startTxIndex,
+	).Order("block_number ASC, tx_index ASC, id ASC")
```
wolf31o2, line 228: This strict `(block_number, tx_index) > cursor` predicate is only safe if all rows for a given `(block_number, tx_index)` become visible atomically. The live Midnight indexer currently writes rows with independent autocommit calls as it scans inputs/outputs, so a client can page after seeing one row for `(B,T)` before another row for the same key is committed. The next request resumes with `start=(B,T)`, this predicate filters the late row out, and the row is skipped permanently. This applies to the MySQL/Postgres copies as well; please either publish each tx/block's Midnight rows atomically or change the cursor/high-water-mark contract so clients never advance past an incomplete key.

midnight/server/midnight_state.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+	var startBlock uint64
+	var startTxIndex uint32
+	if pos := req.GetStartPosition(); pos != nil {
+		startBlock = uint64(pos.GetBlockNumber())
+		startTxIndex = pos.GetTxIndex()
+	}
+	limit := effectivePageSize(req.GetTxCapacity())
+
+	creates, err := s.metadata.FindMidnightAssetCreatesFrom(startBlock, startTxIndex, limit, nil)
```
wolf31o2: `GetUtxoEvents` needs one stable read snapshot across all four `Find*From` calls. As written these are four independent reads with `nil` transactions, so they can observe different committed points while the indexer is applying a block. For example, `creates` can include `(B,T)`, `registrations` can run before the matching registration commits, and the response then returns `next_position=(B,T)`; the next page's strict cursor will never return that registration. Please wrap the merge reads in a read transaction with repeatable snapshot semantics for every backend, and pair that with atomic publication or completed-key gating.

wolf31o2 (changes requested):
No-go for merge. The new Midnight cursor pagination can permanently skip rows when clients read while the live indexer is committing rows for the same transaction/block. CI and the focused local tests pass, but the API contract is not safe until rows sharing a cursor key are published atomically, or the query only pages over completed keys and the merged handler reads a single consistent snapshot.

wolf31o2 (approved):

Issue #1959, opened 2026-04-17
ledger: full reward calculation per Shelley spec Sec 10


wolf31o2:
## Summary

Promote the Q3 reward-state persistence groundwork (#1902) and Koios-backed parity checks (#1903) to a full reward calculation implementation that matches the Shelley ledger specification, Section 10.

## What Needs to Happen

1. Implement the full reward formula per Shelley spec Sec 10:
   - Pool performance (apparent vs. expected blocks)
   - Pledge influence factor (a0)
   - Pool cost and margin
   - Stake distribution at the snapshot boundary
   - Reserves and treasury splits
2. Wire the calculation into the epoch transition path so rewards are computed at the right boundary using the right snapshot ("Go" snapshot, two epochs old).
3. Persist computed rewards per stake credential and per pool.
4. Use the validated implementation to drive Blockfrost reward and account endpoints (correctness across `/accounts/{stake_address}/rewards`, etc.).
5. Validate rewards across multiple epochs against Haskell node and Koios on preview, preprod, and mainnet.

## Depends On

- #1902 (reward-state persistence groundwork, Q3)
- #1903 (Koios-backed reward parity checks, Q3)

## Roadmap

Q1 2027 — mainnet block production readiness, reward calculation.

wolf31o2:
This should be resolved, but needs the full Koios tests to be run.

wolf31o2:
Status update: the greenfield implementation is substantially complete. Reward persistence landed in [#2839](https://github.com/blinklabs-io/dingo/pull/2839), stake snapshots in [#2852](https://github.com/blinklabs-io/dingo/pull/2852), calculation/application in [#2865](https://github.com/blinklabs-io/dingo/pull/2865), epoch integration in [#2874](https://github.com/blinklabs-io/dingo/pull/2874), and Blockfrost reward history in [#3015](https://github.com/blinklabs-io/dingo/pull/3015).

[...]

PR #2784, opened 2026-07-08
feat: rewards calculation

wolf31o2:
Refs #1959 

---
## Summary by cubic
Implements end-to-end Shelley-style stake reward calculation with atomic epoch-boundary snapshotting, a live reward-stake aggregate, and correct fee‑pot accounting via `transaction.collateral_fee` (including gap‑block recomputation). Aligns epoch rollover for issue 1959 by sequencing rewards → MIR → SNAP → pool reap → enact pparams → HARDFORK, and replays boundary‑enacted and boundary‑expired governance deposit returns.

- **New Features**
  - Ledger: integrated reward engine with the rollover order above; epoch events include ProtocolVersion; snapshot manager is constructed before ledger and invoked via an epoch‑boundary callback; MIR distribution and pot‑transfer overflow checks; recompute collateral fee for gap blocks after input hydration.
  - Fee pot: adds `collateral_fee` to `transaction` and `SumTransactionFeesInSlotRange` to include collateral from phase‑2‑invalid txs.
  - Database: live `reward_live_stake` aggregate auto‑refreshes on UTXO/account changes (including genesis) with a one‑shot rebuild API `RebuildRewardLiveStake`; excludes zero‑stake delegators; adds reward‑state inputs/outputs and queries including `GetRewardStakeInputsAtSlot`, `GetAccountsActiveAtSlot` (uses certificate history with `account.created_slot` fallback), `GetPoolBlockIssuersInSlotRange`, and `ImportPool` (upsert + registration); unifies SQL‑dialect handling; improves batch accumulation with O(1) in‑flight producer lookups.
  - Governance: adds exact‑boundary queries to replay deposit returns for proposals enacted or expired at the boundary.

- **Migration**
  - Run DB migrations to add reward‑state tables/columns, `transaction.collateral_fee`, and `account.created_slot` (includes a crash‑safe backfill on upgrade).
  - For MySQL/Postgres, build with `dingo_extra_plugins` to enable the live reward‑stake aggregate.
  - After a fresh ledger import, the aggregate rebuilds automatically; otherwise run `RebuildRewardLiveStake` once if backfill is needed.

<sup>Written for commit 0fa48aa6cc382ddd1e3fb00bc932e02ed9caf5ec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2784?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive stake reward calculation and persistence, including live stake tracking, reward inputs, pool outputs, and account outputs.
  * Added epoch-boundary snapshot capture and improved replay handling for rewards and governance actions.
  * Added queries for active accounts, pool block issuers, transaction fees, and boundary-enacted or expired proposals.
  * Added pool import support and improved pool registration owner details.

* **Bug Fixes**
  * Improved rollback consistency for rewards, UTxOs, governance state, and snapshots.
  * Added safeguards against reward, treasury, and reserve overflows.
  * Corrected stake and delegator counts to exclude zero-stake entries.


ledger/reward_calculation.go
```diff
@@ -0,0 +1,2360 @@
...
+				err,
+			)
+		}
+	}
+
+	reserves, treasury, err := stakeRewardUpdatedPots(app)
+	if err != nil {
+		return err
+	}
+	if err := meta.SetNetworkState(
```
chrisguiney, line 324: 🟠 **Treasury overstatement on a governance-boundary crash-replay (please sanity-check my trace).**

This absolute reset of reserves/treasury (`stakeRewardUpdatedPots` → `SetNetworkState`) is, I believe, what can wipe a committed treasury withdrawal on replay.

The withdrawal debit is applied separately in `ledger/governance/enact.go:286` (`applyTreasuryWithdrawal` → `SetNetworkState`), and `EnactProposal` marks the proposal `enacted_epoch` in the same boundary txn. Because the boundary txn commits separately from the tip advance (per the note around `account.go:870-880`), a crash in that window replays the boundary:

1. this reset restores treasury to its pre-withdrawal **absolute** value, and
2. the `enacted_epoch IS NULL` guard (`sqlite/governance.go:175`) then declines to re-enact the proposal,

so the withdrawal — or an expired-proposal deposit routed to treasury — is never re-subtracted, and treasury ends **overstated**. Recipient balances stay correct via the reward journal, so it's specifically an ADA-conservation break in network-state. The trigger is narrow (crash mid-window on a governance-active boundary), but treasury is ledger state.

Could the treasury delta be journaled the way reward credits are, or folded into `stakeRewardUpdatedPots` so a replay reconstructs it? Would appreciate confirmation on whether I've traced this correctly — the other three boundary ops (pparams, pool reap, MIR) all looked replay-safe.

ledger/rewards/reference_test.go
```diff
@@ -0,0 +1,267 @@
...
+				tc.totalStake,
+			)
+			require.Equal(t, tc.wantLeader, leader)
+			require.Equal(t, tc.wantMember, member)
+			require.Equal(t, tc.wantLeftover, tc.poolReward-leader-member)
+		})
+	}
+}
+
+func TestAmaruReferenceRewardPotDeltas(t *testing.T) {
```
chrisguiney: 🔴 **Anchor at least one reward vector to external ground truth.**

These "Reference"/"Amaru"/"Haskell" suites are valuable regression pins, but as far as I can tell every expected value is hand-derived from the same formulas the implementation uses, and `TestAmaruReferenceRewardPotDeltas` asserts money-conservation self-consistency rather than external values. That means a spec-vs-code error co-derived into the expectations would pass silently — which is the one failure mode that matters most for a rewards engine.

Before merge, can we anchor at least one golden vector to external ground truth — e.g. captured pot/pool/member outputs from cardano-ledger or Amaru for a real (testnet or mainnet) epoch? Even a single real epoch would move this from "internally consistent" to "matches the network."

ledger/rewards/rewards.go
```diff
@@ -0,0 +1,1009 @@
...
+
+func floorRat(value *big.Rat) uint64 {
+	if value == nil || value.Sign() <= 0 {
+		return 0
+	}
+	num := new(big.Int).Set(value.Num())
+	den := value.Denom()
+	q := new(big.Int).Quo(num, den)
+	if !q.IsUint64() {
+		return math.MaxUint64
```
chrisguiney: 🟡 **`floorRat` saturating to `MaxUint64` hides money-math failures.**

Returning `math.MaxUint64` when the value doesn't fit (rather than signaling an error) is a nasty latent failure mode in money math — if an upstream invariant ever breaks, this converts it into an absurd payout instead of failing loudly. Consider returning `(uint64, error)` (or validating the numerator range at the call sites) so a non-fitting result is caught rather than silently saturated.

database/plugin/metadata/sqlite/reward_live_stake.go
```diff
@@ -0,0 +1,559 @@
...
+	ref models.StakeCredentialRef,
+	slot uint64,
+) error {
+	if err := validateRewardLiveStakeRef(ref); err != nil {
+		return err
+	}
+	if len(ref.Key) == 0 {
+		return nil
+	}
+	cache, err := batchFetchCerts(db, []models.StakeCredentialRef{ref}, slot)
```
chrisguiney, line 232: 🟡 **Per-cert N+1 on the live-stake refresh (performance).**

The singular `refreshRewardLiveStakeAggregate` calls `batchFetchCerts` with a single ref, which fans out to ~11 cert-table queries per credential. Since `saveAccount` and `saveCertRecord` both invoke this per-cert in the certificate loop, an N-cert block costs roughly 2N × ~11 round-trips (and dense rollbacks / batch flushes multiply it further).

The plural `refreshRewardLiveStakeAggregates` (:190) already batches by slot — could the per-cert paths accumulate refs and call the batched variant once per block/flush instead? Correctness looks fine here; this is purely throughput on busy blocks.

chrisguiney (changes requested):
**Determination: requesting changes — two items to resolve before merging a consensus rewards engine (details inline). This is strong work otherwise.**

I focused verification on the three highest-risk axes and traced them against the head:

- **Reward math** (`ledger/rewards`): faithfully transcribes the Shelley equations — apparent performance on active-stake σ, `optimalPoolReward` with `a0`/`z0` and circulation as the σ denominator, pledge-influence separation, cost+margin before the member/leader split. All `big.Int`/`big.Rat`, no floats, deterministic (sorted). 👍
- **Epoch-rollover ordering**: correct and matches cardano-ledger — rewards are computed from the **stable** (N-3 mark) snapshot and credited **before** the new epoch's mark is captured. 👍
- **Replay safety**: pparam enactment, pool reap, and MIR are all replay-safe across the boundary-vs-tip commit gap (journaled / absolute-reset).

I also re-traced the automated-review findings against the current code, and the large majority did not reproduce (owner-chunk loss, import ordering, non-boundary live-stake read, rollback overstatement, `create.go`/batch atomicity, `Close` drain, metrics-before-commit, and same-slot cert ordering all look already-handled). Thanks for the thorough tests.

Two things I'd want resolved before merge:

1. **Treasury overstatement on a governance-boundary crash-replay** (inline on `ledger/reward_calculation.go`) — an ADA-conservation break I couldn't talk myself out of. Please sanity-check my trace; it's a hidden interaction, not an obvious mistake.
2. **No externally-anchored reward vectors** (inline on `ledger/rewards/reference_test.go`) — correctness is currently self-referential, which is the one failure mode that matters most for a rewards engine.

Non-blocking nits inline (`floorRat` saturation, per-cert N+1 refresh), plus one here:

- **`d = 1` reward case not tested.** Edge coverage is otherwise thorough (zero stake, saturation, cost>reward, pledge failure, overflow), but I didn't see a fully-centralized `d = 1` case alongside the `d = 0.8` boundary — worth a line to pin it.

Happy to pair on the treasury path. Caveat: even across several deep-dives, large areas (MIR internals, the aggregate across the other two DB backends, snapshot rotation) got only spot checks, so this isn't a line-by-line certification.

database/plugin/metadata/sqlite/account.go
```diff
@@ -790,6 +790,39 @@ func (d *MetadataStoreSqlite) GetAccountsByCredential(
...
+) (map[string]struct{}, error) {
+	ret := make(map[string]struct{}, len(refs))
+	if len(refs) == 0 {
+		return ret, nil
+	}
+	db, err := d.resolveReadDB(txn)
+	if err != nil {
+		return nil, err
+	}
+	cache, err := batchFetchCerts(db, refs, slot)
```
arepala-uml, line 910: Could this make Dingo skip rewards for genesis stake accounts?

`GetAccountsActiveAtSlot` appears to decide account activity only from stake registration/deregistration certificate history. Genesis stake accounts are created directly as active accounts from genesis data, so they may not have matching registration cert rows.

If no registration cert exists, early reward calculation may think a genesis account is inactive. That could cause Dingo to skip rewards or mark them unspendable for accounts that should be rewarded.

**Suggested fix:**  Include active account rows with `added_slot <= querySlot` as active when there is no cert history, or persist synthetic genesis registration history so this historical active check sees genesis accounts correctly. Adding a regression test with a genesis-created stake account that has no registration cert row will be good in this case

ledger/reward_calculation.go
```diff
@@ -0,0 +1,2360 @@
...
+	}
+	eraDesc, ok := ls.eraById(epoch.EraId)
+	if !ok || eraDesc == nil {
+		return nil, rewards.Parameters{}, fmt.Errorf(
+			"unknown era ID %d for performance epoch %d",
+			epoch.EraId, performanceEpoch,
+		)
+	}
+	pparams, err := ls.db.GetPParams(
+		performanceEpoch,
```
arepala-uml, line 1665: Could this make Dingo calculate rewards using old protocol rules after the chain has already moved to new rules?
It looks like rewards are based on block performance from `N-2`, but the reward calculation itself happens in `N-1`. Since this loads pparams from performanceEpoch (`N-2`), I’m wondering whether rewards, treasury, or reserves could differ from the real ledger after a pparam update or hard fork.
Is using `N-2` pparams here intentional, or should the reward constants/protocol-major behavior come from the reward calculation epoch (`N-1`) while only block counts come from `N-2`?

ledger/reward_calculation.go
chrisguiney: Re-verified against the current head (`a6e17ea5`) — **still present**. The `fix: review comments` commit only changed decentralization sourcing and some live-stake DB files; it didn't touch this path (the reward-step absolute `SetNetworkState` reset, `enact.go`, or the `enacted_epoch` guard are all unchanged). Here's a concrete repro so it's actionable — boundary at epoch N that enacts a treasury withdrawal **W**:

1. Boundary txn: `applyStakeRewardApplication` writes network state at `boundarySlot` to an **absolute** `T_afterReward` (`reward_calculation.go:292-296`), computed by `stakeRewardUpdatedPots` (`:1286-1314`) purely from the fixed epoch-(N-1) pots snapshot — it does **not** subtract W.
2. Same txn: `governance.ProcessEpoch` (`chainsync.go:3533`) enacts W → `applyTreasuryWithdrawal` (`enact.go:209`) → `SetNetworkState` writes `T_afterReward − W` (`enact.go:286`) and marks the proposal `enacted_epoch = N`.
3. Boundary txn **commits**. Stored treasury = `T_afterReward − W`. ✅ correct so far.
4. Crash **before** the separate tip-advance txn commits (the two-commit gap the reward replay-guard in `sqlite/account.go` is built around).
5. Restart: persisted tip is still pre-boundary, so the boundary block replays → `processEpochRollover` re-runs (`chainsync.go:3323`).
6. `applyStakeRewardApplication` re-runs and upserts the same `boundarySlot` row back to absolute `T_afterReward` — **erasing the −W** (`reward_calculation.go:296`; single `slot`-keyed row via `SetNetworkState` OnConflict `slot`).
7. `governance.ProcessEpoch`: `GetRatifiedGovernanceProposals` now returns empty because the proposal is already `enacted_epoch = N` (`governance.go:198`), so **W is not re-debited**.
8. Boundary re-commits with treasury = `T_afterReward` — **overstated by W**. ADA-conservation break in network-state; withdrawal *recipient* balances are fine (their credits are journaled/deduped via `AccountRewardDelta`). Same hazard applies to expired-proposal deposits routed to treasury (`governance.go:148`).

Root cause: reward/refund **account credits** are replay-guarded by a journal, but the governance treasury **debit** is a non-journaled read-modify-write that the reward step's absolute reset then wipes while the `enacted_epoch` guard declines to re-apply it.

Minimal fixes (in order of preference):
- **(A)** Journal the treasury debit like reward credits — a source-hashed network-state delta that's skipped on replay. Smallest change; matches the existing `AccountRewardDelta` replay pattern.
- **(B)** Fold enacted treasury movements (withdrawals + refunds + donations) for the boundary into `stakeRewardUpdatedPots`, so the absolute reset already accounts for them. Tradeoff: couples the reward step to governance output and is ordering-sensitive.
- **(C)** Make enactment re-runnable rather than gating re-debit on `enacted_epoch`. Most invasive — risks double-applying other non-idempotent governance side effects (committee/DRep state) unless each is separately guarded.

Happy to be wrong here — if the two-commit boundary/tip gap can't actually replay a *committed* boundary txn (e.g. the boundary and tip advance are the same txn on all backends), this collapses. But from the reward replay-guard's own existence it looks like that gap is real.

chrisguiney (commented):
**Re-review of the latest commits (`fix: review comments`, `fix: lint`) — thanks for the quick turnaround.** CI is green and the PR is in good shape overall; this is a status update, my earlier `CHANGES_REQUESTED` still stands.

**Good progress, verified:** the decentralization-sourcing change is a real correctness improvement — block counting now takes decentralization from the **performance epoch (N-2)** pparams via `rewardDecentralizationFromPParams` rather than the calculation-epoch value, which resolves the drift concern raised in-thread. I checked it: it's bounds-validated (0 ≤ d ≤ 1), Babbage+ correctly yield d = 0, and the omitted Allegra case is a non-issue since `allegra.AllegraProtocolParameters` is a type alias of the Shelley params (matched by the existing Shelley case). 👍

**Still open from my review** — this round didn't touch these:

- 🔴 **External reward vectors** (`ledger/rewards/reference_test.go`): unchanged. Reward correctness is still validated only against self-derived expectations. This round is itself a good example of *why* — consensus reward math got adjusted again (which epoch's `d` feeds block counting) with only self-referential tests to catch a mistake. Still the merge bar for me: one golden vector from cardano-ledger/Amaru for a real epoch.
- 🟠 **Treasury overstatement on governance-boundary crash-replay** (`reward_calculation.go:296` ↔ `enact.go:286`): I re-verified against `a6e17ea5` — **still reproduces**, and this commit didn't touch the path. I've posted a concrete 8-step repro and three fix options as a reply on that thread. Would appreciate a confirm-or-refute.
- 🟡 `floorRat` saturation (`rewards.go`) and 🟡 per-cert N+1 refresh (`reward_live_stake.go`): both untouched.

Not blocking, and no rush — just flagging that the two headline items (external vector + treasury-replay) are what stand between this and an approval from me. Happy to pair on either.

ledger/reward_calculation.go
chrisguiney: @wolf31o2 friendly nudge on this one 🙂 — the squashed `ef23231b` didn't touch this path (`stakeRewardUpdatedPots` → `SetNetworkState`, `enact.go`, and the `enacted_epoch` guard are all unchanged), so I re-confirmed the repro above still applies on the current head. Alongside the external-vector ask, this is the last correctness item holding my approval.

Could you **confirm-or-refute**? The whole thing hinges on one question: is the epoch-boundary txn's commit truly separate from the tip-advance commit? That separation is what makes a *committed* boundary replay possible on restart.

- If they're the **same** txn on every backend, this collapses — a crash can't land between them — and I'll clear it happily.
- If they're **separate** (which the reward `AccountRewardDelta` replay-guard's existence suggests), then option (A) from my earlier comment — journaling the treasury debit the same way reward credits are journaled — is a small, pattern-matching fix.

Happy to pair on it if that's faster than a written back-and-forth.

chrisguiney (commented):
**Status after the rebase to `ef23231b`** (thanks for keeping CI green) — one of my four items is resolved, three remain.

✅ **`floorRat` → `floorRatChecked`** now returns `ErrRewardAmountOverflow` instead of saturating to `MaxUint64`, wired at all call sites. 👍

Still open:

- 🔴 **External reward vector** (`ledger/rewards/reference_test.go`): unchanged — still no externally-anchored golden vector (cardano-ledger/Amaru/a real epoch). The reward math is only validated against its own hand-derived expectations, which is the one failure mode that matters most for a rewards engine. This is my merge bar.
- 🟠 **Treasury overstatement on a governance-boundary crash-replay** (`reward_calculation.go` ↔ `enact.go`): unchanged in the squash. I re-verified the repro still applies on the current head and left a direct confirm-or-refute nudge on that thread.
- 🟡 **Per-cert N+1** on `refreshRewardLiveStakeAggregate` (`batchFetchCerts` with a single ref): performance only, non-blocking.

The two headline items — 🔴 external vector and 🟠 treasury-replay — are what stand between this and an approval from me; the N+1 is a nice-to-have. Holding `CHANGES_REQUESTED` until those two move. Happy to pair on either.

ledger/reward_calculation_test.go
```diff
@@ -0,0 +1,4225 @@
+// Copyright 2026 Blink Labs Software
```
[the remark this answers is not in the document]
wolf31o2: • Violation 1 is not valid. I did not make code changes.

  SaveRewardSnapshot already has explicit upsert semantics in database/plugin/metadata/internal/rewardstate/state.go:72:
  ON CONFLICT (epoch, snapshot_type) DO UPDATE for SQLite/Postgres, and MySQL’s GORM driver renders the same clause as
  ON DUPLICATE KEY UPDATE. All three metadata backends delegate to that same helper: SQLite, Postgres, and MySQL
  wrappers call rewardstate.SaveSnapshot.

  The test also would not silently pass with stale seed data: the seed writes TotalActiveStake=1000 and no nonce at
  ledger/reward_calculation_test.go:4072, while the challenged assertion expects TotalActiveStake=2000 and nonce {0xAB,
  0xCD} at ledger/reward_calculation_test.go:3926. A no-op-on-conflict implementation would fail this assertion.
  Existing SQLite CRUD coverage also verifies overwrite behavior.

  Verified with:

  go test ./ledger -run 'TestStakeRewardPrecomputeSnapshotGuardRejectsSameSlotContentChange|
  TestStakeRewardPrecomputeSnapshotGuardOK'
  go test ./database/plugin/metadata/sqlite -run TestRewardSnapshotCRUD

wolf31o2:
This is being split into a multitude of smaller branches and code reviews.

PR #2787, opened 2026-07-09
fix(leios): apply EB effects on pools and governance

wolf31o2:
---
## Summary by cubic
Applies endorser-block non-UTxO effects on Musashi and splits header verification into crypto-only vs. stateful checks with safe deferral. This keeps pool/governance data and endorser donations visible, adds TPraos genesis-overlay validation, and reduces false peer recycling.

- **New Features**
  - Metadata-only path for Leios endorser txs: record transaction metadata, pool/governance certificates, and `network_donation` under the ranking block; UTxO is unchanged; skip already-seen hashes.
  - Header verification split: chainsync runs stateless VRF/KES/opcert when the epoch nonce is cached; the queue tracks a verified flag so blockfetch avoids duplicate stateless work and can flush pending blocks to compute epoch nonces; stateful checks can be deferred and are persisted in `sync_state` as `deferred_header_validation:<slot>:<hash>` and re-verified during apply.
  - Genesis overlay handling: persist genesis-delegation certs and add lookup by slot; header validation checks the active genesis delegate VRF key, treats overlay slots as non-pool, and skips pool stake-threshold checks while d > 0. Governance processing accepts Dijkstra-era protocol parameters.

- **Bug Fixes**
  - Avoids recycling peers when epoch nonces or stateful data aren’t ready by deferring verification and replaying it during apply; deferral survives restarts.
  - Prevents duplicate metadata effects by deduplicating endorser transactions on Musashi.
  - Removes a redundant MySQL index.

<sup>Written for commit 826a9ce6c587b9adde0d1dea85a1177b1699f36c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2787?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added deferred header validation support so some block headers can be checked later during block processing.
  * Improved endorser block handling to record transaction metadata without always applying UTxO changes.
  * Added support for verified-header tracking and matching in the block queue.

* **Bug Fixes**
  * Prevented duplicate endorser transactions from being applied or recorded twice.
  * Made header verification more reliable when epoch or stake data is temporarily unavailable.

* **Documentation**
  * Updated architecture and database docs to describe the new validation flow and stored markers.


arepala-uml (approved):

chadaibbhot-git:
Confirming **#2787 is working well on Musashi** 🎉

On `dingo:main` @ `sha256:abf38809…` (built 2026-07-10 20:02 UTC), both a block producer (`core` storage) and a query node (`api` storage) sync **from genesis straight to network tip, cleanly**, on the live Musashi Leios testnet (magic 164, `DINGO_RUN_MODE=leios`).

EB-carried pool registrations are applied correctly now — nodes validate and follow blocks from EB-registered pools (e.g. pool `371a035c…`, active since epoch 8) without a hitch, confirmed against an independent explorer and a second node. Staying at tip on Musashi just works.

Thanks for the quick turnaround. We'll validate #2796 (dijkstra forging) at our next leader slot and follow up.

PR #2839, opened 2026-07-15
feat(database): maintain reward metadata state

wolf31o2:
Relates to #1959 

---
## Summary by cubic
Adds persistent reward metadata with a live per-credential stake aggregate and ended-epoch inputs for delayed reward processing and faster snapshots. Unifies reward logic across `sqlite`, `mysql`, and `postgres`, with automatic refresh, a startup backfill that runs before ledger processing, and a rebuild API, addressing #1959.

- New Features
  - Reward tables/models: `RewardLiveStake`, `RewardStakeInput/Output`, `RewardPoolInput/Output`, `RewardAccountOutput`, `RewardSnapshot`.
  - Shared reward-state/query logic under `database/plugin/metadata/internal/rewardstate` with SQL dialect helpers.
  - Live-stake refresh on UTxO/account changes and reward credits/withdrawals; `Database.RebuildRewardLiveStake(slot)`; startup backfill for older databases.
  - Effective ended-epoch pool-registration queries; owners and reward-account identity loaded; SQLite query chunking for bind limits.
  - O(1) in-batch UTxO producer index to support collateral fee and live-stake updates.
  - Snapshots skip saving reward inputs when ended-epoch metadata is not yet available (warns and continues).
  - One-time migration adds a pool index for `RewardLiveStake`.

- Bug Fixes
  - Fee pot correctness: include consumed collateral; recompute collateral fee after materializing inputs during gap import; clamp oversized collateral returns.
  - SQL correctness: use `NUMERIC`/text casts to preserve large `uint64` values in reward deltas and live-stake math on `postgres`.
  - Account reactivation clears stale `pool`/`drep` fields.
  - Deleting live UTxOs recomputes live stake; historical deletes do not.
  - Reward withdrawal replay is idempotent, including when TxHash is empty (TxHash + credential used as the replay key).
  - `mysql`/`postgres` active-at-slot queries fall back to immutable `created_slot` when no registration exists (matches `sqlite`).

<sup>Written for commit 535056c817cdb63f98e1608da58b50be3cfdc7ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2839?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


database/plugin/metadata/internal/sqldialect/sqldialect.go
```diff
@@ -67,3 +68,35 @@ func IntegerCastType(db *gorm.DB) string {
...
+	return "0"
+}
+
+// ArithmeticParam returns the bind fragment for a types.Uint64 operand used
+// directly in arithmetic. Postgres requires an explicit numeric cast.
+func ArithmeticParam(db *gorm.DB) string {
+	if Name(db) == "postgres" {
+		return fmt.Sprintf("CAST(? AS %s)", IntegerCastType(db))
+	}
+	return "?"
```
chrisguiney, line 102: **major:** On MySQL this bare `?` makes `CreditLiveStakeDelta`'s UPDATE evaluate `CAST(reward_stake AS UNSIGNED) + '<decimal string>'` — MySQL evaluates any arithmetic with a string operand in DOUBLE. Two consequences: (a) precision is silently lost above 2^53 lovelace (~9M ADA on one credential); (b) `CAST(<double> AS CHAR)` renders values ≥1e16 in scientific notation (e.g. `'4.5e16'` — a devnet genesis credential routinely holds that much), which `types.Uint64.Scan` then fails to parse on every subsequent read of that row.

MySQL should return `CAST(? AS UNSIGNED)` so the addition stays integer/DECIMAL — exactly what `RebuildLiveStake`'s pure-SQL path already does. This defeats the exact-decimal-string machinery that is the stated reason the parameter is bound as a string in the first place.

Relatedly: the `IntegerCastType` doc block above justifies MySQL's `UNSIGNED` because a signed cast would overflow above `MaxInt64`, yet postgres uses signed `BIGINT`, which raises "bigint out of range" for the very values the string encoding exists to carry. Unreachable with real lovelace (max supply ~4.5e16), but `NUMERIC` would honor the contract the comment asserts.

database/models/reward_state.go
```diff
@@ -51,20 +51,106 @@ func (RewardSnapshot) TableName() string {
...
+	StakingKey    []byte       `gorm:"uniqueIndex:idx_reward_live_stake_cred,priority:2;size:28;not null"`
+	ID            uint         `gorm:"primarykey"`
+	CredentialTag uint8        `gorm:"uniqueIndex:idx_reward_live_stake_cred,priority:1;not null;default:0;index:idx_reward_live_stake_pool,priority:2"`
+	UtxoStake     types.Uint64 `gorm:"not null"`
+	RewardStake   types.Uint64 `gorm:"not null"`
+	TotalStake    types.Uint64 `gorm:"not null;index:idx_reward_live_stake_pool,priority:3"`
+	Registered    bool         `gorm:"not null;index:idx_reward_live_stake_pool,priority:4"`
+	// PoolDelegation* pins the certificate order for PoolKeyHash. A stake
+	// credential only contributes to a pool's reward snapshot when this
+	// delegation is at least as recent as the pool's current registration.
```
chrisguiney: **major:** This comment documents gating that is implemented nowhere: no query in the PR reads `pool_delegation_slot/block_index/cert_index` — they are only written by `RefreshLiveStakeAggregate`/`RebuildLiveStake`, and `StakeInputsAtSlot` filters only on `pool_key_hash IN` + `registered` + `total_stake > 0`.

Concrete divergence: cardano-ledger's POOLREAP removes delegations to a retired pool, but nothing here clears `account.pool` at pool reap. If pool P retires and later re-registers, every stale delegator still has `account.pool = P`, so their stake is attributed to P in the reward snapshot — member/leader rewards diverge from the reference implementation.

Either implement the recency gating in `StakeInputsAtSlot` (the columns and index position look designed for it), or correct this comment to describe the columns as rollback/rebuild bookkeeping only.

database/models/reward_state.go
```diff
@@ -51,20 +51,106 @@ func (RewardSnapshot) TableName() string {
...
+func (RewardStakeInput) TableName() string {
+	return "reward_stake_input"
+}
+
+// RewardLiveStake is the live per-stake-credential aggregate used to make
+// epoch-boundary reward snapshots cheap. UtxoStake and RewardStake are stored
+// separately so rollback/account-reward repair can refresh only the affected
+// credential while TotalStake is directly queryable by the snapshot copier.
+type RewardLiveStake struct {
+	PoolKeyHash   []byte       `gorm:"index:idx_reward_live_stake_pool,priority:1;size:28"`
```
chrisguiney: **minor:** The column order of `idx_reward_live_stake_pool` doesn't serve its only query. `StakeInputsAtSlot` filters `pool_key_hash IN` + `registered = ?` + `CAST(total_stake ...) > ?`: `total_stake` at priority 3 is only ever compared through a CAST (index-unusable), which cuts off index use before `registered` at priority 4, and `pool_delegation_slot` at priority 5 is never queried at all. Everything past `credential_tag` is pure write amplification. If the recency gating lands, reorder to `(pool_key_hash, registered, ...)`; otherwise trim.

DATABASE.md
```diff
@@ -387,11 +387,15 @@ process the same pointer unless the claim expires before a result is recorded.
...
... active account remain as `registered=false` rows with no pool and are ignored by reward snapshot input copying until a live registered delegation exists. The `pool_delegation_*` columns store the certificate order for the current pool delegation for rollback/rebuild fidelity; reward snapshot capture preserves that live delegation across later pool parameter registrations and filters by active pool input, registered account state, and positive total stake. Epoch-boundary reward snapshots copy from this table instead of joining the full live UTxO set, which is required to keep mainnet rollover from starving block production near boundary leader slots. `MetadataStore.RewardLiveStakeNeedsBackfill` reports whether any canonical account or live-UTxO credential is missing from this table, detecting both empty and partially populated upgraded databases while remaining false for a legitimately fresh empty database. Node startup (`node.go`, the normal `Run` path) checks this after database recovery and before ledger processing, then runs `RebuildRewardLiveStake` at the persisted tip slot when needed, so the next epoch-boundary snapshot cannot observe an empty or partial aggregate (#1959). |
```
chrisguiney: **major (docs):** Two claims in this row don't match the code in this PR:

1. "Epoch-boundary reward snapshots copy from this table instead of joining the full live UTxO set" — the mark-snapshot capture path is unchanged: `captureMarkSnapshot` → `CalculateStakeDistribution` → `GetStakeByPoolsAtSlot`, the slot-aware full-UTxO join. `GetRewardStakeInputsAtSlot` (the method that reads this aggregate) has zero callers, so the claimed boundary-starvation fix is not wired.
2. "Mithril import rebuilds it after cert-state import and stale-state reconcile" — no code in `ledgerstate/` or `mithril/` calls `RebuildRewardLiveStake`; the only caller in the repo is the startup backfill in `node.go`. A freshly imported DB has an empty aggregate until the next `Node.Run`. (`store.go`'s doc comment for `RebuildRewardLiveStake` repeats this claim.)

Please trim these to present-state behavior.

DATABASE.md
```diff
@@ -387,11 +387,15 @@ process the same pointer unless the claim expires before a result is recorded.
...
+| `reward_snapshot` | `id`, `epoch`, `snapshot_type`, `total_active_stake`, `total_pool_count`, `total_delegators`, `captured_slot`, `boundary_slot`, `epoch_nonce`, `protocol_version` | PK `id`; unique `(epoch, snapshot_type)`; indexes `captured_slot`, `boundary_slot` | Reward-calculation snapshot metadata. Stake rewards applied at epoch N use the Mark snapshot from epoch N-3, block production from N-2, and ADA pots plus reward protocol parameters from the RUPD calculation epoch N-1. Historical post-Mithril `pool_stake_snapshot` seed rows may intentionally have no matching `reward_snapshot`; reward application skips missing input bundles rather than using current live stake as historical data. |
```
chrisguiney: **major (docs):** These rows document reward calculation/application behavior that doesn't exist anywhere in the code: "Boundary application only trusts precomputed pool outputs when…", the pool-count-match requirement against `reward_snapshot.total_pool_count`, "reward application skips missing input bundles", "Snapshot refreshes call `DeleteRewardInputsForEpoch` before saving…" (no caller of `DeleteRewardInputsForEpoch` exists), and the N−3 mark-snapshot consumption rule. One statement is contradicted by code that does exist: "Snapshot capture freezes stake and pool-parameter inputs only" — `rotation.go` stores `BlocksProduced`/`TotalBlocksInEpoch` on the captured input row.

Also: the retention note is internally inconsistent — `cleanupOldSnapshots` deletes reward state before `currentEpoch-2` at each transition, but rewards applied at epoch N are documented to consume the epoch N−3 mark snapshot, i.e. those rows would be pruned exactly when needed (moot only because the application code doesn't exist yet).

Per the repo's docs-are-part-of-the-change-bar rule, these sections should describe present state only, or the consuming code should land in the same PR.

ARCHITECTURE.md
```diff
@@ -2124,6 +2124,50 @@ LedgerState --> Epoch Transition --> EventBus (EpochTransitionEvent)
...
+
+The metadata plugins own the persistent inputs and outputs used by delayed
+stake-reward processing. `MetadataStore` exposes narrow APIs for ADA pots,
+reward snapshot metadata, frozen pool/stake inputs, calculated pool/account
+outputs, and rollback or retention cleanup. Callers pass the surrounding
+metadata transaction so snapshot replacement, output invalidation, and chain
+state changes can commit atomically. SQLite, MySQL, and PostgreSQL share the
+same reward-state semantics through `database/plugin/metadata/internal/`.
+
+`RewardLiveStake` is the live per-stake-credential aggregate that prevents an
```
chrisguiney: **major (docs):** Same issue as DATABASE.md — "prevents an epoch-boundary reward snapshot from joining the full UTxO set" is not true of this PR: the mark-snapshot path still goes through `GetStakeByPoolsAtSlot` (the full join), and `GetRewardStakeInputsAtSlot` has no callers. Please phrase this as what the aggregate is for once the consumer lands, or rewrite to present state.

database/plugin/metadata/store.go
```diff
@@ -1085,6 +1145,48 @@ type MetadataStore interface {
 		types.Txn,
 	) ([]*models.RewardPoolInput, error)
 
+	// SaveRewardStakeInputs saves per-credential reward snapshot inputs.
+	SaveRewardStakeInputs(
```
chrisguiney: **major (scope):** Ten of the twelve new interface methods in this PR have no callers and no tests: `GetPoolBlockIssuersInSlotRange`, `GetAccountsActiveAtSlot`, `GetRewardStakeInputsAtSlot`, `SaveRewardStakeInputs`, `GetRewardStakeInputs`, `DeleteRewardInputsForEpoch`, `DeleteRewardOutputsForEpoch`, `SaveRewardPoolOutputs`/`GetRewardPoolOutputs`, `SaveRewardAccountOutputs`/`GetRewardAccountOutputs` — each implemented three times (sqlite/mysql/postgres) but referenced by nothing, not even test files. Only `GetPoolRegistrationsEffectiveForEpoch` and `RebuildRewardLiveStake`/`RewardLiveStakeNeedsBackfill` are consumed.

Suggest landing this surface with the reward-calculation consumer instead — when it lands, any signature churn hits three backends, and until then nothing validates these against real usage.

Also a nit while here: `GetRewardStakeInputsAtSlot`'s `slot` parameter is ignored by all three implementations; an interface parameter no backend reads should be dropped until a use exists.

database/plugin/metadata/mysql/mysql_test.go
```diff
@@ -564,7 +564,7 @@ func TestMysqlSetTransactionBatchesMultiInputUtxoLookups(t *testing.T) {
 	}
 
 	// Verify exactly one batch SELECT was used per input class.
-	if got := recorder.countUtxoSelects(); got != 3 {
+	if got := recorder.countUtxoLookupSelects(); got != 3 {
```
chrisguiney, line 567: **major (test coverage):** This rename is the only mysql test change in the PR — there are zero behavioral mysql tests for the new reward logic. Postgres gets exactly one (`TestPostgresConcurrentWithdrawalReplayAtDifferentSlots`, which is a good test), but mysql received the identical `FOR UPDATE` replay-serialization change and has no counterpart, despite InnoDB REPEATABLE READ being where that race is most subtle (the consistent-read hazard is even acknowledged in `insertMissingDeregistrationAccount`'s own comments).

No mysql or postgres test touches `RewardLiveStake`, `GetAccountsActiveAtSlot`, `RebuildRewardLiveStake`, or the rollback refresh — the per-dialect SQL (UNSIGNED casts, tuple-IN, `INSERT ... WITH` CTE, `FOR UPDATE`) is exercised on neither of the dialects it was written for. That's exactly how the `ArithmeticParam` float bug survived. At minimum, a mysql test crediting a large delta through `CreditLiveStakeDelta` would have caught it.

Two sqlite-side gaps as well: the `getOrCreateAccount` reactivation fix (clearing stale `Pool`/`Drep` on re-registration) has no test anywhere, and nothing asserts the rollback paths restore `RewardLiveStake` (only the spent-UTxO no-refresh case is pinned).

database/plugin/metadata/mysql/transaction.go
```diff
@@ -776,14 +778,25 @@ func (d *MetadataStoreMysql) getOrCreateAccount(
...
-func saveAccount(account *models.Account, db *gorm.DB) error {
+func saveAccount(
+	account *models.Account,
+	db *gorm.DB,
+	rewardRefs ...map[string]rewardCredentialSlotRef,
+) error {
+	// Resolve the create-helper sentinel to the account's creation slot. For a
+	// freshly built account CreatedSlot==AddedSlot; an existing row loaded from
+	// the DB carries its real (never-sentinel) CreatedSlot and is left as-is,
+	// and no upsert DoUpdates list includes created_slot, so it stays immutable.
```
chrisguiney, line 799: **minor:** "it stays immutable" is not quite true on this backend: the existing-row branch below still uses `db.Save(account)`, which writes `created_slot` from the in-memory copy. Sqlite's `saveAccount` was changed in this PR to apply `MIN(created_slot, ?)` atomically, with a comment explaining the multi-caller sentinel race it closes — mysql and postgres did not get that fix, while this PR simultaneously makes `created_slot` load-bearing for correctness (`GetAccountsActiveAtSlot` compares against it). Suggest porting the sqlite `MIN()` approach here and in postgres.

database/plugin/metadata/mysql/reward_live_stake.go
```diff
@@ -0,0 +1,313 @@
...
+	// credential in the loop below: a batch can hold many refs sharing one
+	// slot, and flattening per-credential made refresh work O(refs*cacheSize).
+	// The flattened map is read-only downstream, so sharing one instance
+	// across all refs at a slot is safe.
+	cachesBySlot := make(
+		map[uint64]rewardstate.PoolDelegationCache,
+		len(refsBySlot),
+	)
+	for slot, slotRefs := range refsBySlot {
+		cache, err := batchFetchCerts(db, slotRefs, slot)
```
chrisguiney, line 204: **minor:** `batchFetchCerts` is unchunked on mysql and postgres — all ten per-cert-table queries take the full staking-key list in one `IN ?`, while sqlite chunks at 400. Both the mysql and postgres wire protocols cap bind parameters at 65,535. A deep rollback (`DeleteTransactionsAfterSlot` pins every touched credential into a single `refreshRewardLiveStakeAggregates` call) or a large `GetAccountsActiveAtSlot` batch with more distinct keys than that fails hard mid-rollback. Inconsistent with `fetchFallbackAccountState`/`fetchCredentialsEverDeregistered` in the same file, which chunk at 400.

database/plugin/metadata/internal/rewardstate/livestake.go
```diff
@@ -0,0 +1,748 @@
...
+				ELSE 0
+			END AS pool_delegation_block_index,
+			CASE
+				WHEN account.active = %[3]s AND LENGTH(account.pool) > 0
+				THEN COALESCE(latest_delegation.cert_index, 0)
+				ELSE 0
+			END AS pool_delegation_cert_index,
+			? AS updated_slot
+		FROM (
+			SELECT credential_tag, staking_key FROM account
```
chrisguiney, line 695: **minor:** A NULL `staking_key` account row slips past `ValidateLiveStakeSourceCredentials` (its predicate `credential_tag > 1 OR LENGTH(staking_key) != 28` evaluates to NULL, not TRUE, for NULL keys), but this `creds` arm takes account rows unfiltered — no `LENGTH(staking_key) > 0` guard, unlike the utxo arm and unlike `LiveStakeNeedsBackfill`. A NULL-key row (schema-possible: `models.Account.StakingKey` has no `not null` tag) would then hit `reward_live_stake.staking_key NOT NULL` and fail the whole rebuild with an opaque constraint violation instead of the validator's diagnostic. Add `staking_key IS NULL` to the validator or a length guard to this arm.

database/plugin/metadata/internal/rewardstate/livestake.go
```diff
@@ -0,0 +1,748 @@
...
+	return nil
+}
+
+// LiveStakeNeedsBackfill reports whether any canonical account or live-UTxO
+// credential is missing from reward_live_stake. Testing for missing keys, not
+// merely an empty aggregate, also repairs upgraded databases where post-upgrade
+// writes populated only part of the table before the first restart.
+func LiveStakeNeedsBackfill(db *gorm.DB) (bool, error) {
+	var missingCount int64
+	if err := db.Raw(`
```
chrisguiney, line 737: **minor:** This check runs on every startup, before the node serves, as a `COUNT(*)` over the deduplicated union of all account credentials and all live-UTxO credentials with a NOT EXISTS anti-join — `COUNT(*)` must enumerate every missing row even though one suffices, and even in the steady state the full union scan (mainnet: ~10M+ live UTxOs) is paid on every boot. An `EXISTS ... LIMIT 1` short-circuits the common case; better still, the repo already has the `backfill_checkpoint` phase-marker pattern (used by the `account.created_slot` backfill) that would make steady state one indexed row read. Crash-safety itself is fine — check + rebuild share one transaction, so an interrupted backfill retries cleanly.

database/plugin/metadata/internal/stakequery/poolparams.go
```diff
@@ -0,0 +1,193 @@
...
+	FROM pool_retirement ret
+	LEFT JOIN certs rc ON rc.id = ret.certificate_id
+	LEFT JOIN %[1]s rt ON rt.id = rc.transaction_id
+	WHERE ret.pool_key_hash IN ? AND ret.added_slot < ? AND ret.epoch <= ?
+), ranked AS (
+	SELECT pool_key_hash, reg_id, is_retirement,
+		ROW_NUMBER() OVER (
+			PARTITION BY pool_key_hash
+			ORDER BY added_slot DESC, bi DESC, ci DESC,
+				is_retirement ASC, reg_id DESC
```
chrisguiney: **minor:** `is_retirement ASC` prefers registration over retirement on a full tie, which is the opposite of the documented convention for synthetic retirements. Full ties are only reachable when the certs/transaction joins miss (COALESCEd `bi=0, ci=0`) — exactly the case for synthetic Mithril reconcile retirements (`certificate_id = 0`). DATABASE.md (updated in this same PR) says active-pool queries must rank synthetic retirements ahead of certificate-backed rows at the same slot. Scenario: pool synthetically retired at the catch-up tip slot where a registration row also lands with no cert metadata → treated as still registered with pre-epoch params.

database/plugin/metadata/sqlite/account.go
```diff
@@ -935,21 +1142,19 @@ func (d *MetadataStoreSqlite) ApplyAccountRewardWithdrawal(
...
-			).First(&existing)
-			if result.Error == nil {
-				return nil
-			}
-			if !errors.Is(result.Error, gorm.ErrRecordNotFound) {
-				return result.Error
-			}
+		var existing models.AccountRewardDelta
+		result := tx.Where(
+			"withdrawal = ? AND tx_hash = ? AND credential_tag = ? AND staking_key = ?",
```
chrisguiney, line 967: **minor (latent):** With nil `txHash` normalized to `[]byte{}` and no slot in this replay key, the first empty-hash withdrawal for a credential permanently swallows all later ones: any second, genuinely distinct nil-hash withdrawal at any later slot matches this pre-check and returns nil without clearing the reward or writing a delta. Omitting the slot is correct for real tx hashes (globally unique, and the replay tests pin that), and today's only production caller always passes a real hash — but this is an exported `MetadataStore` method, and nothing in its doc comment warns that nil `txHash` makes subsequent nil-hash withdrawals no-ops forever. A future caller (state import, repair tool) applying two legitimate hashless withdrawals would silently drop the second. At minimum, document the constraint on the interface method.

database/plugin/metadata/sqlite/transaction.go
```diff
@@ -4851,6 +5109,13 @@ func (d *MetadataStoreSqlite) DeleteTransactionsAfterSlot(
 		return result.Error
 	}
 
+	if err := refreshRewardLiveStakeAggregates(db, rewardRefs); err != nil {
```
chrisguiney, line 5107: **minor:** `DeleteTransactionsAfterSlot` is not self-wrapping when `txn == nil`, unlike every other mutator this PR touched (`DeleteUtxo(s)`, `AddUtxos`, `SetAccount`, the reward credit/withdraw paths, etc. all got the `if txn != nil { mutation(db) } else { db.Transaction(mutation) }` pattern). Called with nil `txn`, the collect-refs → restore-spends → delete → refresh sequence runs statement-by-statement autocommit; a crash between the delete and this refresh strands `reward_live_stake` stale with no retry marker (the backfill check detects missing credentials, not stale amounts). The one production caller always passes a txn inside the ledger rollback transaction, so this is hardening rather than a live bug — but worth closing for consistency. Same applies to the mysql/postgres copies.

database/plugin/metadata/sqlite/utxo.go
```diff
@@ -644,12 +664,27 @@ func (d *MetadataStoreSqlite) DeleteUtxo(
...
 	}
-	result := db.Where("tx_id = ? AND output_idx = ?", utxoId.Hash, utxoId.Idx).
-		Delete(&models.Utxo{})
-	if result.Error != nil {
-		return result.Error
+	mutation := func(tx *gorm.DB) error {
+		refs, err := rewardStakeRefsFromLiveUtxoIDs(
+			tx,
+			[]models.UtxoId{utxoId},
+			0,
```
chrisguiney: **minor:** Passing `slot=0` here means a live row derives its recompute slot from the UTxO's own `AddedSlot`, which can be far in the past. `refreshRewardLiveStakeAggregates` then bounds `batchFetchCerts` at that old slot, so a delegation cert newer than the UTxO is invisible and `poolDelegationForAccount` falls back to `account.AddedSlot` for the `pool_delegation_*` columns; `updated_slot` also regresses below the row's previous value. Stake amounts stay correct (recomputed from current rows) and the common pruning path only deletes spent rows (excluded by the live-only filter, correctly pinned by `TestDeleteSpentUtxoDoesNotRefreshLiveStakeAtHistoricalSlot`), so the impact is metadata drift on the aggregate row when a repair path deletes a live UTxO. Consider recomputing at the current tip slot instead of the row's `AddedSlot`. Same in `DeleteUtxos` below.

database/reward_state.go
```diff
@@ -14,7 +14,37 @@
...
+// RebuildRewardLiveStake rebuilds the live reward stake aggregate from
+// canonical account and live UTxO metadata.
+func (d *Database) RebuildRewardLiveStake(slot uint64, txn *Txn) error {
+	if txn == nil {
+		return d.MetadataTxn(true).Do(func(t *Txn) error {
+			return d.metadata.RebuildRewardLiveStake(slot, t.Metadata())
+		})
+	}
+	if txn.Metadata() == nil {
+		return fmt.Errorf("rebuild reward live stake: %w", types.ErrNilTxn)
```
chrisguiney: **minor:** Two issues here: (1) a blob-only transaction — where the caller did pass a txn — is reported as `types.ErrNilTxn` rather than `ErrTxnWrongType`, and `reward_state_test.go` then locks in that misleading taxonomy; (2) the non-nil path dispatches through `txn.db.metadata` while the nil path uses `d.metadata` — a txn created from a different `Database` instance would silently rebuild the other database's aggregate. Suggest validating `txn.db == d` (or always dispatching through `d.metadata`) and returning `ErrTxnWrongType` for the metadata-less case.

database/plugin/metadata/sqlite/reward_live_stake.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	return db.Transaction(rebuild)
+}
+
+// RewardLiveStakeNeedsBackfill reports whether any canonical account or live
+// UTxO credential is missing from reward_live_stake.
+func (d *MetadataStoreSqlite) RewardLiveStakeNeedsBackfill(
+	txn types.Txn,
+) (bool, error) {
+	db, err := d.resolveDB(txn)
```
chrisguiney: **nit:** `RewardLiveStakeNeedsBackfill` is a pure read but resolves via `resolveDB` (write resolver), so it can't be called inside a read-only metadata txn — every other `Get*` uses `resolveReadDB`. Same in the mysql and postgres copies.

chrisguiney (changes requested):
Overall the wired-in code here is careful: refresh triggers are present at identical sites across all three backends, transaction scoping is uniform (caller-supplied txn honored, self-wrap otherwise), migrations are centralized so schema drift is impossible, overflow/underflow and bind-limit edge cases show real care, and the withdrawal/credit replay idempotency work is thorough and well tested on sqlite. Rollback ordering, the crash-safe `created_slot` backfill gating, and the epoch/slot boundary math in `rotation.go` all check out.

Requesting changes for the items in the inline comments. The blocking ones:

1. MySQL `CreditLiveStakeDelta` evaluates lovelace arithmetic in floating point (string operand → DOUBLE), which silently loses precision above 2^53 and can render totals in scientific notation that `types.Uint64.Scan` then fails to parse.
2. The `pool_delegation_*` columns document a delegation-recency gating rule that no query implements — stale delegators to a retired-then-re-registered pool still count toward its snapshot.
3. DATABASE.md/ARCHITECTURE.md describe reward-pipeline behavior that isn't wired in this PR (epoch-boundary snapshots still do the full UTxO join; no Mithril-import rebuild exists; the calculation/application admission rules don't exist anywhere).
4. Ten of the twelve new `MetadataStore` methods have no callers and no tests, implemented in triplicate.
5. Every new behavior is tested on sqlite only (one race test on postgres, nothing on mysql) — the dialect-bearing SQL is validated only on the dialect that needs it least, which is exactly how bug 1 survived.

Smaller items are inline as well. Nits not worth inline comments: batch-size constant drift between mysql/postgres siblings (`rewardStakeInputPoolBatchSize` vs bare `1000`); sqlite/mysql carry two redundant `addRewardStakeRef` calls after `saveAccount` in `SetGenesisGovernance` that postgres omits; `collateralfee.ForTransaction` folds a phase-1-impossible inconsistency (all inputs resolved but return exceeds sum) into the lenient path with no log; `rewardstate/dialect.go` exports two wrappers nothing external uses; the `CreditLiveStakeDelta` doc comment asserts a row lock that the glebarez sqlite driver silently drops (single-writer semantics happen to cover it — worth saying so); and the credential/slot-ref map machinery exists in four structurally identical copies.

Pre-existing (not this PR, but surfaced in scope and worth follow-ups): `GetControlledAmountByCredential`/`GetScriptLockedSupply` run uncast `SUM(amount)` on the TEXT-typed `types.Uint64` column (errors on postgres, loses precision on mysql above 2^53); postgres pool-registration upsert omits `deposit_amount` from `DoUpdates` where mysql/sqlite include it.

chrisguiney (approved):

PR #2845, opened 2026-07-15
perf(ledger): Add lock-free copy-on-write state snapshots

### LINKED ISSUE #2601: ledger: copy-on-write snapshots for read-mostly consensus and tip state
## Summary

The hot read accessors on ledger.LedgerState (Tip, ChainTipSlot,
GetCurrentPParams, CurrentEpoch, CurrentTransitionInfo, IsAtTip, and the
epoch/era/nonce queries) all take the embedded sync.RWMutex read lock. These
fields change rarely (per block for the tip, per epoch or era boundary for the
rest) but are read constantly from API handlers, chainsync, forging, and block
validation. Moving them behind immutable copy-on-write snapshots
(atomic.Pointer) makes those reads lock-free.

## Motivation

A prototype microbenchmark compared the current RWMutex getters against an
atomic.Pointer equivalent, swept 1 to 16 cores (ns/op, lower is better):

- RWMutex read-only: 72, 122, 130, 133 (plateaus on the shared reader counter)
- RWMutex with 1 percent concurrent writer: 72, 268, 439, 591 (scales
  backwards; each writer Lock stalls all readers)
- atomic read-only: 2.2, 0.55, 0.28, 0.16
- atomic with 1 percent writer: 5.0, 10.2, 11.0, 11.4

The relevant regime is a per-block writer running alongside concurrent readers,
where the RWMutex path degrades to about 591 ns at 16 cores while the atomic
path stays around 11 ns.

This is a microbenchmark of the lock pattern, not end-to-end node throughput,
so the absolute per-read cost is small next to a full request. The value is
that the read path stops scaling backwards under a concurrent writer, and it
compounds once block decode and validation run in parallel (#1894), where many
workers read these fields at once.

## Scope

Group the read-mostly fields into two immutable snapshots, each behind an
atomic.Pointer, chosen so fields updated together stay consistent (no torn
reads):

- consensus snapshot: currentEpoch, currentEra, currentPParams, prevEraPParams,
  epochCache, transitionInfo (updated together at epoch and era boundaries, on
  rollback, and at startup load)
- tip snapshot: currentTip, currentTipBlockNonce (updated together per block)
- reachedTip becomes an atomic.Bool

Keep the embedded RWMutex for the remaining operational fields (chainsync and
blockfetch state, the epoch nonce hex cache) and to serialize writers.

Read accessors load the relevant snapshot and return the field, with unchanged
signatures, so external callers are not affected. Writers already compute new
values into locals and assign them under a single Lock; instead they build one
new snapshot and store it. Because all snapshot writes already happen under the
same Lock (or single threaded at startup), no compare-and-swap is needed.
Slice-backed fields (block nonce, tip hash, epoch nonce) are already replaced
wholesale and never mutated in place, so sharing them in an immutable snapshot
is safe.

## Constraints

- White-box tests in the ledger package read and write these fields directly
  and take the embedded lock; they need behavior-preserving updates to go
  through accessors or test helpers.
- This is consensus critical. It requires go test -race ./ledger/...,
  conformance tests, DevNet validation, and ideally cross-node ledger
  comparison (#1900) before merge.

## Relationship to other issues

- Concrete first slice of #2254 (decompose LedgerState). It removes read-side
  lock acquisitions from the shared mutex without moving whole subsystems yet.
- Intersects #2292: the startup-only write paths (loadEpochs, loadPParams,
  loadTip) are part of the write surface and should be handled consistently
  with the single-init invariant.
- Value compounds with #1894 (parallel decode and validate).

## Acceptance criteria

- The listed read-mostly fields live in atomic.Pointer snapshots; the hot
  getters no longer take the RWMutex.
- Getter signatures and observable behavior are unchanged.
- Writers publish new snapshots under the existing write lock; co-update groups
  stay consistent.
- go test -race ./ledger/... and conformance pass; DevNet validated.

arepala-uml:
1. Adds immutable atomic snapshots for frequently read ledger consensus and tip state, removing RWMutex contention from hot read paths.
2. Added `consensusSnapshot` to group epoch, era, protocol parameters, epoch cache, and hard-fork transition state, ensuring related values are read consistently without locks.
3. Added `tipSnapshot` to group the applied chain tip with its block nonce, preventing readers from observing mismatched tip and nonce values.
4. Added atomic snapshot pointers to `LedgerState`, allowing hot readers to load immutable state without acquiring `RWMutex`.
5. Changed `reachedTip` to `atomic.Bool`, removing the read lock from `IsAtTip()`.
6. Added cloning helpers for hashes and nonce slices. The epoch cache is published by reference for performance, while writers must replace it using copy-on-write before mutation.
7. Added `publishSnapshotsLocked()` with shared generations, allowing readers to retry until consensus and tip snapshots come from the same publication.
8. Removed unsafe nil snapshot fallbacks so all readers use initialized atomic snapshots.
9. Modified hot getters such as `Tip(), ChainTipSlot(), CurrentEpoch(), GetCurrentPParams()`, and `CurrentTransitionInfo()` to use lock-free snapshot reads.
10. Modified epoch, era, nonce, protocol-parameter, transaction-validation, header-verification, and node-to-client queries to read from snapshots, reducing contention across API, chain-sync, forging, and validation paths.
11. Added protocol-parameter cloning before epoch updates and deep-cloned hard-fork CostModels to prevent published snapshot data from being mutated.
12. Modified startup loading to publish snapshots after restoring epochs, protocol parameters, tips, and hard-fork transition information, preventing stale state after restart.
13. Modified block processing, rollback, epoch rollover, hard-fork evaluation, and Mithril nonce healing to publish updated snapshots before releasing the write lock.
14. Modified speculative epoch-cache advancement to allocate a new slice before publication, preserving older immutable snapshots.
15. Modified `Tip()` and epoch/nonce return paths to provide defensive copies, preventing callers from changing internal snapshot data.
16. Added snapshot immutability tests to verify old readers retain unchanged state after a newer snapshot is published.
17. Added race-regression tests for snapshot readers and a paired-generation test verifying consensus and tip cannot come from different publications.
18. Modified replay recovery, chain-sync rollback, and benchmark fixtures to publish directly initialized state before snapshot-based code reads it.
19. Updated `ARCHITECTURE.md `to document lock-free ledger readers, serialized writers, snapshot consistency groups, and slice immutability rules.
20. Ran the full devnet test suite and all DevNet tests got passed with these changes.

Closes #2601 

---
## Summary by cubic
Adds lock‑free, copy‑on‑write snapshots for consensus and tip state to remove `RWMutex` contention and keep era/epoch and tip/nonce reads consistent. Clarifies snapshot guarantees and prevents protocol‑parameter and hard‑fork `CostModels` mutations across eras (addresses #2601).

- **Refactors**
  - Publish consensus and tip snapshots together under one generation; paired‑load helper retries until generations match.
  - Serialize startup publication in `Start()` after reconstructing transition state; publish tip‑only changes without copying the epoch history.
  - Share the epoch cache by reference and cap capacity; `advanceEpochCache` now copies before append and republishes to preserve old snapshots.
  - Clone protocol parameters during epoch rollover and clone hard‑fork `CostModels` (Babbage/Conway/Dijkstra); add `cloneProtocolParametersForEra` and `eras/cloneCostModels`. Switch hot getters/queries to snapshot reads and return defensive copies; use `atomic.Bool` for `reachedTip`.

- **Bug Fixes**
  - Defer publication in `setEpochCache` so readers observe startup mutations even if later validation fails; remove unsafe nil snapshot fallbacks and add `SetTipForTesting`.
  - Fix races in post‑fork `ProtocolParamsForSlot` by cloning `CostModels`; add tests for immutability, epoch advancement preserving old snapshots, paired snapshot generations, and concurrent snapshot loads.
  - Publish after rollback, protocol‑parameter load, and HFI transition changes so snapshot readers see current state; update `ARCHITECTURE.md` to document snapshot rules and guarantees.

<sup>Written for commit 01cb644845b558e71bf4e091acd1ff55a265c6cd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2845?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Improvements**
  * Improved ledger-state consistency during concurrent reads and updates.
  * Enabled faster access to frequently queried chain, epoch, protocol, and tip information.
  * Improved protection against inconsistent or partially updated state during processing and recovery.
  * Preserved safe copies of returned tip data to prevent unintended modifications.

* **Documentation**
  * Clarified ledger threading and concurrency behavior.

* **Tests**
  * Added coverage for snapshot immutability and concurrent reader consistency.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2097,12 +2097,29 @@ Database operations and event delivery use worker pools for controlled concurren
 | Pattern | Usage |
 |---------|-------|
 | Goroutine Management | Tracked WaitGroups for clean shutdown |
-| Mutex Protection | RWMutex for read-heavy operations |
-| Atomic Operations | Atomic types for metrics counters |
+| Mutex Protection | Ledger writer serialization and mutable operational state |
+| Atomic Operations | Immutable ledger read snapshots and metrics counters |
 | Channel Communication | EventBus async delivery |
 | Context Cancellation | Graceful shutdown signals |
 | Worker Pools | Database operations and event delivery |
 | sync.Once | Ensure single shutdown execution |
+
+`LedgerState` publishes its read-mostly state through two immutable
+copy-on-write snapshots. The consensus snapshot groups the current epoch, era,
+current and previous-era protocol parameters, epoch cache, and hard-fork
+transition information. The tip snapshot groups the applied chain tip with the
+Praos block nonce belonging to that tip. Hot query paths load these snapshots
+through `atomic.Pointer` without acquiring the state's `RWMutex`; `reachedTip`
+is an `atomic.Bool`.
+
+Ledger writers remain serialized by the existing mutex. They compute changes
+privately, update the writer-owned state, and publish a complete replacement
+snapshot before unlocking. Slice-backed epoch nonces, tip hashes, and block
+nonces are copied at publication boundaries and published snapshots must never
+be mutated. Each snapshot is internally consistent, but a caller loading both
+snapshot pointers may observe adjacent publication generations; code requiring
+one exact cross-snapshot commit must remain under the ledger lock or explicitly
+retry.
 
 ## Configuration
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2134,10 +2134,13 @@ Each snapshot is internally consistent, but a caller loading both snapshot
 pointers could otherwise observe adjacent publication generations. Each
 production publication therefore stamps both snapshots with one generation,
 and code requiring fields from both uses a paired-load helper that retries until
-the generations match. A nil-snapshot fallback exists only for zero-value
-white-box test fixtures; it builds snapshots independently and does not provide
-the cross-snapshot generation guarantee. Production states initialize both
-snapshots in `NewLedgerState` before becoming visible.
+the generations match. The load helpers dereference `consensus.Load()` /
+`tip.Load()` directly and do not nil-check: `NewLedgerState` always calls
+`publishSnapshotsLocked()` before the state becomes visible, so production
+snapshots are never nil. White-box test fixtures that construct `LedgerState{}`
+directly must call `publishSnapshotsLocked()` (or `NewLedgerState` /
+`SetTipForTesting`) themselves before exercising any snapshot-reading path, or
+the read will nil-dereference.
 
 ## Configuration
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2117,9 +2117,10 @@ privately, update the writer-owned state, and publish a complete replacement
 snapshot before unlocking. Slice-backed epoch nonces, tip hashes, and block
 nonces are copied at publication boundaries and published snapshots must never
 be mutated. Each snapshot is internally consistent, but a caller loading both
-snapshot pointers may observe adjacent publication generations; code requiring
-one exact cross-snapshot commit must remain under the ledger lock or explicitly
-retry.
+snapshot pointers could otherwise observe adjacent publication generations.
+Each publication therefore stamps both snapshots with one generation, and code
+requiring fields from both uses a paired-load helper that retries until the
+generations match.
 
 ## Configuration
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2104,19 +2104,24 @@ Database operations and event delivery use worker pools for controlled concurren
 | Worker Pools | Database operations and event delivery |
 | sync.Once | Ensure single shutdown execution |
 
-`LedgerState` publishes its read-mostly state through two immutable
-copy-on-write snapshots. The consensus snapshot groups the current epoch, era,
-current and previous-era protocol parameters, epoch cache, and hard-fork
-transition information. The tip snapshot groups the applied chain tip with the
-Praos block nonce belonging to that tip. Hot query paths load these snapshots
-through `atomic.Pointer` without acquiring the state's `RWMutex`; `reachedTip`
-is an `atomic.Bool`.
+`LedgerState` publishes its read-mostly state through two copy-on-write
+snapshots. The consensus snapshot groups the current epoch, era, current and
+previous-era protocol parameters, epoch cache, and hard-fork transition
+information. The tip snapshot groups the applied chain tip with the Praos block
+nonce belonging to that tip. Hot query paths load these snapshots through
+`atomic.Pointer` without acquiring the state's `RWMutex`; `reachedTip` is an
+`atomic.Bool`. Snapshot containers and their owned byte slices are immutable;
+protocol-parameter values are shared and consumers must treat them as read-only.
 
 Ledger writers remain serialized by the existing mutex. They compute changes
 privately, update the writer-owned state, and publish a complete replacement
-snapshot before unlocking. Slice-backed epoch nonces, tip hashes, and block
-nonces are copied at publication boundaries and published snapshots must never
-be mutated. Each snapshot is internally consistent, but a caller loading both
+snapshot before unlocking. Tip hashes, block nonces, and the current epoch's
+nonce fields are copied at publication. The full historical epoch cache is
+instead shared by reference to avoid an allocation whose cost grows with chain
+age on every block: writers must replace the cache and nested nonce slices
+before modifying them. Publication caps the cache capacity so even an accidental
+append allocates a new backing array. Each snapshot is internally consistent,
+but a caller loading both
 snapshot pointers could otherwise observe adjacent publication generations.
 Each publication therefore stamps both snapshots with one generation, and code
 requiring fields from both uses a paired-load helper that retries until the
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2112,6 +2112,12 @@ nonce belonging to that tip. Hot query paths load these snapshots through
 `atomic.Pointer` without acquiring the state's `RWMutex`; `reachedTip` is an
 `atomic.Bool`. Snapshot containers and their owned byte slices are immutable;
 protocol-parameter values are shared and consumers must treat them as read-only.
+Era-specific parameter-update functions mutate their concrete parameter
+pointer in place, so `processEpochRollover` clones the current protocol
+parameters (`cloneProtocolParametersForEra`) into a transaction-owned value
+before running them; without that clone, an epoch boundary applying an
+on-chain pparam update would mutate the pointee that a published snapshot's
+`currentPParams`/`prevEraPParams` still reference.
 
 Runtime ledger writers remain serialized by the existing mutex. They compute
 changes privately, update the writer-owned state, and publish a complete
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2113,19 +2113,25 @@ nonce belonging to that tip. Hot query paths load these snapshots through
 `atomic.Bool`. Snapshot containers and their owned byte slices are immutable;
 protocol-parameter values are shared and consumers must treat them as read-only.
 
-Ledger writers remain serialized by the existing mutex. They compute changes
-privately, update the writer-owned state, and publish a complete replacement
-snapshot before unlocking. Tip hashes, block nonces, and the current epoch's
-nonce fields are copied at publication. The full historical epoch cache is
-instead shared by reference to avoid an allocation whose cost grows with chain
-age on every block: writers must replace the cache and nested nonce slices
+Runtime ledger writers remain serialized by the existing mutex. They compute
+changes privately, update the writer-owned state, and publish a complete
+replacement snapshot before unlocking. Construction and single-threaded
+startup may also publish without the mutex, but only before the `LedgerState`
+is visible to concurrent readers. Tip hashes, block nonces, and the current
+epoch's nonce fields are copied at publication. The full historical epoch cache
+is instead shared by reference to avoid an allocation whose cost grows with
+chain age on every block: writers must replace the cache and nested nonce slices
 before modifying them. Publication caps the cache capacity so even an accidental
-append allocates a new backing array. Each snapshot is internally consistent,
-but a caller loading both
-snapshot pointers could otherwise observe adjacent publication generations.
-Each publication therefore stamps both snapshots with one generation, and code
-requiring fields from both uses a paired-load helper that retries until the
-generations match.
+append allocates a new backing array.
+
+Each snapshot is internally consistent, but a caller loading both snapshot
+pointers could otherwise observe adjacent publication generations. Each
+production publication therefore stamps both snapshots with one generation,
+and code requiring fields from both uses a paired-load helper that retries until
+the generations match. A nil-snapshot fallback exists only for zero-value
+white-box test fixtures; it builds snapshots independently and does not provide
+the cross-snapshot generation guarantee. Production states initialize both
+snapshots in `NewLedgerState` before becoming visible.
 
 ## Configuration
 
--- a/ledger/queries.go
+++ b/ledger/queries.go
@@ -85,22 +85,16 @@ func (ls *LedgerState) querySystemStart() (any, error) {
 }
 
 func (ls *LedgerState) queryChainBlockNo() (any, error) {
-	ls.RLock()
-	point := ls.currentTip.Point
-	blockNumber := ls.currentTip.BlockNumber
-	ls.RUnlock()
+	tip := ls.loadTipSnapshot().currentTip
 	// WithOrigin BlockNo: [0] at genesis, [1, blockNo] once a block exists.
-	if len(point.Hash) == 0 {
+	if len(tip.Point.Hash) == 0 {
 		return []any{0}, nil
 	}
-	return []any{1, blockNumber}, nil
+	return []any{1, tip.BlockNumber}, nil
 }
 
 func (ls *LedgerState) queryChainPoint() (any, error) {
-	ls.RLock()
-	point := ls.currentTip.Point
-	ls.RUnlock()
-	return point, nil
+	return ls.loadTipSnapshot().currentTip.Point, nil
 }
 
 func (ls *LedgerState) queryHardFork(
@@ -108,10 +102,7 @@ func (ls *LedgerState) queryHardFork(
 ) (any, error) {
 	switch q := query.Query.(type) {
 	case *olocalstatequery.HardForkCurrentEraQuery:
-		ls.RLock()
-		eraId := ls.currentEra.Id
-		ls.RUnlock()
-		return eraId, nil
+		return ls.loadConsensusSnapshot().currentEra.Id, nil
 	case *olocalstatequery.HardForkEraHistoryQuery:
 		return ls.queryHardForkEraHistory()
 	default:
@@ -162,14 +153,13 @@ type eraBoundData struct {
 }
 
 func (ls *LedgerState) queryHardForkEraHistory() (any, error) {
-	// Snapshot the tip, current era, and transition info under the read lock
-	// so we can use them without holding the lock during the (potentially
-	// slow) DB queries below.
-	ls.RLock()
-	tipSlot := ls.currentTip.Point.Slot
-	currentEraId := ls.currentEra.Id
-	transitionInfo := ls.transitionInfo
-	ls.RUnlock()
+	// Read the tip, current era, and transition info from the lock-free
+	// snapshots so this (potentially slow) DB-querying path never contends
+	// with the ledger write lock.
+	tipSlot := ls.loadTipSnapshot().currentTip.Point.Slot
+	consensusState := ls.loadConsensusSnapshot()
+	currentEraId := consensusState.currentEra.Id
+	transitionInfo := consensusState.transitionInfo
 
 	shape := ls.eraShape()
 	if len(shape.Eras) == 0 {
@@ -429,15 +419,9 @@ func (ls *LedgerState) queryShelley(
 ) (any, error) {
 	switch q := query.Query.(type) {
 	case *olocalstatequery.ShelleyEpochNoQuery:
-		ls.RLock()
-		epochId := ls.currentEpoch.EpochId
-		ls.RUnlock()
-		return []any{epochId}, nil
+		return []any{ls.loadConsensusSnapshot().currentEpoch.EpochId}, nil
 	case *olocalstatequery.ShelleyCurrentProtocolParamsQuery:
-		ls.RLock()
-		pparams := ls.currentPParams
-		ls.RUnlock()
-		return []any{pparams}, nil
+		return []any{ls.loadConsensusSnapshot().currentPParams}, nil
 	case *olocalstatequery.ShelleyGenesisConfigQuery:
 		return ls.queryShelleyGenesisConfig()
 	case *olocalstatequery.ShelleyUtxoByAddressQuery:
@@ -589,9 +573,8 @@ func (ls *LedgerState) queryShelleyAccountState() (any, error) {
 // drepDeposit returns the current dRepDeposit protocol parameter, which is the
 // deposit every DRep locks at registration. Returns 0 outside Conway.
 func (ls *LedgerState) drepDeposit() uint64 {
-	ls.RLock()
-	defer ls.RUnlock()
-	if cpp, ok := ls.currentPParams.(*conway.ConwayProtocolParameters); ok &&
+	pparams := ls.loadConsensusSnapshot().currentPParams
+	if cpp, ok := pparams.(*conway.ConwayProtocolParameters); ok &&
 		cpp != nil {
 		return cpp.DRepDeposit
 	}
--- a/ledger/queries.go
+++ b/ledger/queries.go
@@ -94,7 +94,7 @@ func (ls *LedgerState) queryChainBlockNo() (any, error) {
 }
 
 func (ls *LedgerState) queryChainPoint() (any, error) {
-	return ls.loadTipSnapshot().currentTip.Point, nil
+	return cloneTip(ls.loadTipSnapshot().currentTip).Point, nil
 }
 
 func (ls *LedgerState) queryHardFork(
@@ -156,8 +156,8 @@ func (ls *LedgerState) queryHardForkEraHistory() (any, error) {
 	// Read the tip, current era, and transition info from the lock-free
 	// snapshots so this (potentially slow) DB-querying path never contends
 	// with the ledger write lock.
-	tipSlot := ls.loadTipSnapshot().currentTip.Point.Slot
-	consensusState := ls.loadConsensusSnapshot()
+	consensusState, tipState := ls.loadStateSnapshots()
+	tipSlot := tipState.currentTip.Point.Slot
 	currentEraId := consensusState.currentEra.Id
 	transitionInfo := consensusState.transitionInfo
 
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -4891,6 +4891,10 @@ func (ls *LedgerState) PrepareEpochCacheForStartup() error {
 
 func (ls *LedgerState) setEpochCache(txn *database.Txn, epochs []models.Epoch) error {
 	ls.epochCache = epochs
+	// Publish every mutation made by this startup writer, including partial
+	// state on error returns, so snapshot readers can never retain a stale view
+	// if this helper is reused outside fatal startup handling in the future.
+	defer ls.publishSnapshotsLocked()
 	clear(ls.epochNonceHexCache)
 	if len(epochs) > 0 {
 		// Recover epoch records whose LastEpochBlockNonce was persisted empty
@@ -4906,7 +4910,6 @@ func (ls *LedgerState) setEpochCache(txn *database.Txn, epochs []models.Epoch) e
 		ls.currentEra = *eraDesc
 		// Update metrics
 		ls.metrics.epochNum.Set(float64(ls.currentEpoch.EpochId))
-		ls.publishSnapshotsLocked()
 		return nil
 	}
 	// Populate initial epoch
@@ -4969,7 +4972,6 @@ func (ls *LedgerState) setEpochCache(txn *database.Txn, epochs []models.Epoch) e
 	ls.currentPParams = rolloverResult.NewCurrentPParams
 	ls.checkpointWrittenForEpoch = rolloverResult.CheckpointWrittenForEpoch
 	ls.metrics.epochNum.Set(rolloverResult.NewEpochNum)
-	ls.publishSnapshotsLocked()
 	return nil
 }
 
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -531,19 +531,43 @@ type slotBattleRecorderHolder struct {
 	recorder SlotBattleRecorder
 }
 
+// consensusSnapshot contains ledger state that is read frequently and updated
+// as one logical unit at epoch/era boundaries and during rollback. Published
+// snapshots are immutable.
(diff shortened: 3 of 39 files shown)


arepala-uml:
**Testing:** 

Ran the devnet test suite using `./internal/test/devnet/run-tests.sh --keep-up 2>&1 | tee devnet_test.log` and all tests passed, including block production, chain growth, consensus, and epoch-boundary recovery. All containers stayed healthy, transaction submission worked, and no consensus, nonce, VRF/KES, panic, or fatal errors were found.

[devnet_test1.log](https://github.com/user-attachments/files/30064898/devnet_test1.log)

ledger/queries.go
```diff
@@ -85,33 +85,24 @@ func (ls *LedgerState) querySystemStart() (any, error) {
...
+	return []any{1, tip.BlockNumber}, nil
 }
 
 func (ls *LedgerState) queryChainPoint() (any, error) {
-	ls.RLock()
-	point := ls.currentTip.Point
-	ls.RUnlock()
-	return point, nil
+	return ls.loadTipSnapshot().currentTip.Point, nil
 }
```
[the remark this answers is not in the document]
arepala-uml: I have fixed it latest commit.

ledger/state.go
```diff
@@ -531,19 +531,43 @@ type slotBattleRecorderHolder struct {
...
+type tipSnapshot struct {
+	currentTip           ochainsync.Tip
+	currentTipBlockNonce []byte
+}
+
 type LedgerState struct {
-	metrics                            stateMetrics
+	metrics   stateMetrics
+	consensus atomic.Pointer[consensusSnapshot]
+	tip       atomic.Pointer[tipSnapshot]
```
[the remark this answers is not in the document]
arepala-uml: I have fixed it in latest commit changes and please check it.

arepala-uml:
@cubic-dev-ai

arepala-uml:
@cubic-dev-ai review

ledger/state_snapshot_test.go
```diff
@@ -0,0 +1,196 @@
...
+	ls.publishSnapshotsLocked()
+
+	oldConsensus := ls.consensus.Load()
+	oldTip := ls.tip.Load()
+	// Replace slice-backed epoch-cache state before mutation, matching the
+	// production copy-on-write invariant, then publish a new generation. The
+	// retained snapshots must continue to expose generation 7.
+	ls.currentEpoch.EpochId = 8
+	ls.currentEpoch.Nonce[0] = 8
+	ls.epochCache = cloneEpochs(ls.epochCache)
```
chrisguiney, line 51: **blocker:** This test can't fail for the reason it exists. It hand-builds a `LedgerState` literal and calls `publishSnapshotsLocked()` directly — no production writer is exercised — and then this line performs the copy-on-write that *writers* are supposed to perform, as the comment above concedes ("matching the production copy-on-write invariant"):

```go
ls.epochCache = cloneEpochs(ls.epochCache)   // the test doing the writer's job
ls.epochCache[0].Nonce[0] = 8
```

So the assertion at line 51 confirms only that a slice the test itself just cloned wasn't aliased. It's evidence about `cloneEpochs`, not about writers.

Deleting this one line makes the test fail (`expected: 0x7, actual: 0x8`) — which usefully proves the published snapshot really does alias writer-owned `epochCache`, and that the invariant is load-bearing rather than theoretical.

The gap this leaves is real, not hypothetical: I patched `verify_header.go:1284-1286` to `ls.epochCache = append(ls.epochCache, newEpoch)` (an in-place append that violates the invariant) and **the full `-race` suite passed in 272s**. Please add a test that drives a real writer path — e.g. call `advanceEpochCache`, retain the previously published snapshot, and assert its `epochCache` is unchanged — so an in-place mutation actually fails CI.

ledger/state.go
```diff
@@ -729,6 +759,105 @@ func NewLedgerState(cfg LedgerStateConfig) (*LedgerState, error) {
...
+	ls.consensus.Store(&consensusSnapshot{
+		generation:     generation,
+		currentEpoch:   cloneEpoch(ls.currentEpoch),
+		currentEra:     ls.currentEra,
+		currentPParams: ls.currentPParams,
+		prevEraPParams: ls.prevEraPParams,
+		// epochCache is immutable after publication. Writers replace it with a
+		// fresh slice, so tip-only publications can safely reuse the existing
+		// cache instead of copying the full epoch history for every block.
+		epochCache:     ls.epochCache,
```
chrisguiney, line 809: **blocker (same root cause as the test comment):** publishing `ls.epochCache` by reference is a sound optimization and I verified the invariant holds at every writer today — rollover (`3117`, `4940`) and rollback (`2210`) take fresh slices from `db.GetEpochs`, `setEpochCache` (`4868`) takes the DB slice, and `advanceEpochCache` (`verify_header.go:1284-1286`) does an explicit `make`+`copy`+`append` that cannot touch a published array. Correct as written.

The problem is that this comment is the *only* thing enforcing it. There's no test (see `state_snapshot_test.go:43`), and ARCHITECTURE.md currently tells the reader the opposite.

One hazard already in-tree: `healEmptyLabNoncesInPlace` (`state.go:4991`) mutates epoch elements in place — exactly the shape this invariant forbids. It's safe *today* only because both call sites (`4873`, `2149`) hand it a fresh, not-yet-published DB slice. Nothing in its name or signature signals that passing an already-published `ls.epochCache` would silently corrupt live readers, and `healEmptyLabNonces()` (`4972`) already has that call shape. Worth a doc line on that helper naming the precondition.

ARCHITECTURE.md
```diff
@@ -2026,13 +2026,31 @@ Database operations and event delivery use worker pools for controlled concurren
...
+copy-on-write snapshots. The consensus snapshot groups the current epoch, era,
+current and previous-era protocol parameters, epoch cache, and hard-fork
+transition information. The tip snapshot groups the applied chain tip with the
+Praos block nonce belonging to that tip. Hot query paths load these snapshots
+through `atomic.Pointer` without acquiring the state's `RWMutex`; `reachedTip`
+is an `atomic.Bool`.
+
+Ledger writers remain serialized by the existing mutex. They compute changes
+privately, update the writer-owned state, and publish a complete replacement
+snapshot before unlocking. Slice-backed epoch nonces, tip hashes, and block
```
chrisguiney: **blocker:** "Slice-backed epoch nonces, tip hashes, and block nonces are copied at publication boundaries" is **false for `epochCache`**, and that's the one case where getting it wrong corrupts live readers.

`currentEpoch` (`state.go:795`), `currentTip` (`807`), and `currentTipBlockNonce` (`808`) are cloned. `epochCache` is published **by reference** (`802`) — and its elements carry `Nonce`, `EvolvingNonce`, `CandidateNonce`, and `LastEpochBlockNonce`, i.e. epoch nonces. A maintainer reading this paragraph would reasonably conclude an in-place edit to a cached epoch's nonce is safe. It isn't.

The doc also omits the design's most load-bearing rule entirely — *writers must replace `epochCache` with a fresh slice, never mutate it in place* — which currently survives only as the code comment at `state.go:799-801`. That rule belongs here.

Two smaller inaccuracies in the same paragraph:
- "a paired-load helper that retries until the generations match" — true of the normal path, but the nil fallback at `state.go:825` returns two independently-built snapshots with **no** generation check, voiding exactly that guarantee.
- "publish a complete replacement snapshot before unlocking" — `NewLedgerState` (`742`) and `setEpochCache` (`4884`/`4947`) publish without holding the mutex. The code comment at `788-789` carves this out; the doc doesn't.

ledger/state.go
```diff
@@ -531,19 +531,48 @@ type slotBattleRecorderHolder struct {
...
+	currentEpoch   models.Epoch
+	currentEra     eras.EraDesc
+	currentPParams lcommon.ProtocolParameters
+	prevEraPParams lcommon.ProtocolParameters
+	epochCache     []models.Epoch
+	transitionInfo hardfork.TransitionInfo
+}
+
+// tipSnapshot contains the applied tip and the Praos block nonce belonging to
+// that tip. Published snapshots are immutable.
```
chrisguiney, line 549: **major (pre-existing aliasing, but this PR codifies it as a guarantee):** "Published snapshots are immutable" doesn't hold for `currentPParams`/`prevEraPParams`. Every slice field is properly cloned, but these two are **interfaces holding pointers** — publication copies the interface word, not the pointee, so every published snapshot shares one `*conway.ConwayProtocolParameters`.

The writer path mutates that pointee in place, outside the lock:
- `state.go:3017-3021` copies the interface under `RLock` and releases it
- `state.go:3026` explicitly runs the txn "WITHOUT holding ls.Lock()"
- → `processEpochRollover` → `database/pparams.go:267` → `updateFunc(...)`
- → `eras/conway.go:89`: `conwayPParams.Update(&upd); return conwayPParams, nil` — `Update` has a **pointer receiver** and assigns `p.MinFeeA = *paramUpdate.MinFeeA` in place (gouroboros `conway/pparams.go:192`), returning the same pointer.

So at an epoch boundary carrying a pparam update or an enacted governance `ParameterChange`, the rollover goroutine writes `MinFeeA`/`MaxTxSize`/`CostModels` while `GetCurrentPParams()`, `ProtocolParamsForSlot()`, `validateTxCore()` and `EvaluateTx()` read them lock-free. Reproducible under `-race`.

**To be clear, this is not a regression** — I checked base `3806f3d7`: the old `RLock` guarded only the interface copy, never the pointee, and the writer never held the lock. Exposure is unchanged.

But the PR adds this comment plus `TestLedgerStateSnapshotPublicationIsImmutable`, which asserts a guarantee the code doesn't make and doesn't cover pparams. Please either narrow the language here (say slice-backed state is cloned; pparams pointees are shared and must be treated read-only) or deep-copy pparams on publish. Codifying "immutable" over a field that provably isn't will mislead whoever relies on it next.

ledger/state.go
```diff
@@ -729,6 +759,105 @@ func NewLedgerState(cfg LedgerStateConfig) (*LedgerState, error) {
...
+	*consensusSnapshot,
+	*tipSnapshot,
+) {
+	for {
+		consensusState := ls.consensus.Load()
+		tipState := ls.tip.Load()
+		if consensusState == nil || tipState == nil {
+			// Support zero-value white-box fixtures. Production states publish
+			// both snapshots in NewLedgerState before becoming visible.
+			return ls.loadConsensusSnapshot(), ls.loadTipSnapshot()
```
chrisguiney: **major:** this fallback silently voids the invariant the function exists to provide. When exactly one pointer is nil, it returns `loadConsensusSnapshot(), loadTipSnapshot()` with **no generation check** — a real published consensus snapshot (gen N) paired with a fabricated tip snapshot, i.e. precisely the torn pair this helper prevents everywhere else.

That state is reachable: `ouroboros/chainsync_test.go:231-241` reflectively zeroes *only* `ls.tip`. It doesn't bite today solely because `GetIntersectPoint` calls `loadTipSnapshot()` alone.

The fallback is genuinely unreachable in production — I verified `NewLedgerState` (`733-742`) publishes unconditionally with no error return between the struct literal and the publish, so both pointers are non-nil for the state's whole lifetime. So this is a design cost, not a live race: production readers carry an unsynchronized path (it reads `ls.currentEpoch`/`ls.epochCache`/`ls.currentTip` under **no lock**) that exists only so white-box fixtures can poke fields, and it fails *open* — silently racy — rather than loudly.

Since the other fixtures (`replay_recovery_test.go`, `benchmark_test.go`) now just call `publishSnapshotsLocked()`, the clean fix is to delete both fallbacks and have the `ouroboros` test do the same. At minimum, apply the generation check on this path too.

ouroboros/chainsync_test.go
```diff
@@ -228,6 +228,18 @@ func setTestLedgerTip(
 		field.Type(),
 		unsafe.Pointer(field.UnsafeAddr()),
 	).Elem().Set(reflect.ValueOf(tip))
+
+	// NewLedgerState publishes an origin tip snapshot during construction.
+	// These white-box tests replace currentTip directly, so clear that initial
+	// snapshot and let the ledger's test-fixture fallback rebuild its reader
+	// view from the value above. Otherwise GetIntersectPoint keeps observing
+	// origin even though the fixture's writer-owned tip has advanced.
+	snapshotField := reflect.ValueOf(ls).Elem().FieldByName("tip")
```
chrisguiney: **major:** this fixture reaches in with `reflect`+`unsafe` to zero the `tip` pointer specifically to defeat the publish that `NewLedgerState` performs — which means these tests no longer exercise the atomic read path production uses. They drive the nil fallback instead, and I verified that path is racy: a probe replicating this shape (nil tip pointer, lock-free reader, locked writer) reports `WARNING: DATA RACE ... loadTipSnapshot() state.go:856` reading `ls.currentTip` under no lock.

So `GetIntersectPoint` here is now validated against a code path that can't run in production, using an escape hatch that only exists for tests. Needing `unsafe` to make a read observe a write is a signal the invariant is leaking.

Suggest replacing the reflective zeroing with a `publishSnapshotsLocked()` call after setting `currentTip` — same as the other fixtures in this PR — which both exercises the real path and lets the nil fallback be deleted (see `state.go:825`).

ledger/state.go
```diff
@@ -4740,6 +4881,7 @@ func (ls *LedgerState) setEpochCache(txn *database.Txn, epochs []models.Epoch) e
 		ls.currentEra = *eraDesc
 		// Update metrics
 		ls.metrics.epochNum.Set(float64(ls.currentEpoch.EpochId))
+		ls.publishSnapshotsLocked()
```
chrisguiney: **minor (latent, not a live bug):** `setEpochCache` has four mutate-then-error-return paths that skip this publish — `4879` (unknown era ID), `4890` (failed to load Shelley genesis), `4921` (`transitionToEra`), and `4937` (`processEpochRollover`) — after mutating `ls.epochCache` (`4868`), `ls.currentEpoch` (`4876`), `ls.currentEra` (`4898`), and `applyEraTransition`'s four fields (`4927`).

No reader can observe the unpublished state today: `setEpochCache` is startup-only (`loadEpochs` ← `Start():927`, and `PrepareEpochCacheForStartup():4863`), and both callers treat the error as fatal — `Start()` returns and `node.go:518` aborts startup, so the node never serves.

Worth closing anyway, because the failure mode is silent: if epoch reload ever becomes a runtime path, these returns leave readers on stale consensus state with no panic and no race report. A `defer`-based publish (as the HFI tally at `4685` already does) would make it structural rather than positional.

ledger/state.go
```diff
@@ -4543,7 +4681,10 @@ func (ls *LedgerState) evaluateHardForkInitiationStability() {
 			return
 		}
 		ls.Lock()
-		defer ls.Unlock()
+		defer func() {
+			ls.publishSnapshotsLocked()
```
chrisguiney: **nit:** this `defer` is the right shape for guaranteeing publication across the early returns — good call. Side effect worth noting: it now publishes unconditionally, including on the generation-guard and `TransitionKnown` early returns where nothing was mutated. Each no-op publish clones the epoch and tip, bumps `snapshotGeneration`, and adds retry pressure to `loadStateSnapshots`. Harmless, but a `published bool` set after the guards would avoid the churn.

ledger/state.go
```diff
@@ -869,6 +998,17 @@ func (ls *LedgerState) Start(ctx context.Context) error {
...
 	ls.evaluateTransitionImpossible()
 	ls.evaluateHardForkInitiationStability()
+	// Publish the transitionInfo changes made above before any snapshot-reading
+	// goroutine becomes runnable. The HFI stability evaluation above may launch
+	// an asynchronous tally, so serialize this publication with its completion
+	// path and every other snapshot writer. Without this, a restart that
+	// reconstructs TransitionKnown /
+	// TransitionImpossible here would leave every snapshot-based reader
+	// (CurrentTransitionInfo, ConsensusModeForEpoch, ...) on the stale
+	// pre-loadTip() value until the next unrelated write path republishes.
```
chrisguiney, line 1000: **nit:** the publish here is correct and its stated rationale — serializing with the async HFI tally via `Lock` — is sound. The bug it prevents is real and well described.

One inaccuracy in the comment above: it claims this runs "before any snapshot-reading goroutine becomes runnable," but the EventBus subscriptions are registered ~40 lines earlier at `961-990`, so readers can already be live. The publish is still correct (the `Lock` does the real work); the justification just overstates.

ledger/state_snapshot_test.go
```diff
@@ -0,0 +1,196 @@
...
+	// return the hash stored in the immutable tip snapshot.
+	tip := ls.Tip()
+	tip.Point.Hash[0] = 9
+	require.Equal(t, byte(1), ls.Tip().Point.Hash[0])
+}
+
+// TestLedgerStateSnapshotsStayConsistentWithConcurrentReaders verifies that
+// readers never observe fields from different generations within either the
+// consensus snapshot or the tip snapshot while a writer repeatedly publishes.
+func TestLedgerStateSnapshotsStayConsistentWithConcurrentReaders(
```
chrisguiney: **nit:** this one is near-tautological. It loads a single `atomic.Pointer` to an immutable struct and compares two fields *inside that one struct* — tearing is impossible by construction, so the PR-body claim it backs ("readers never observe torn epoch/era or tip/nonce combinations") is guaranteed by the type rather than demonstrated by the test. It keeps modest value as a `-race` guard against regressing to live-field reads.

Worth contrasting with `TestLedgerStatePairedSnapshotsUseOneGeneration` below, which is the real thing: I removed the generation comparison at `state.go:827-829` and it failed 5/5 runs under `-race` in under 0.2s. That's the bar the rest of this file should meet.

ledger/state.go
```diff
@@ -5815,11 +5951,10 @@ func (ls *LedgerState) GetCurrentPParams() lcommon.ProtocolParameters {
 func (ls *LedgerState) ProtocolParamsForSlot(
 	slot uint64,
 ) lcommon.ProtocolParameters {
-	ls.RLock()
-	currentEpoch := ls.currentEpoch
-	currentEra := ls.currentEra
-	currentPParams := ls.currentPParams
-	ls.RUnlock()
+	snapshot := ls.loadConsensusSnapshot()
```
chrisguiney, line 5990: **major (pre-existing, flagged here because this is the anchor in-diff):** converting this to a lock-free snapshot read is fine in itself, but it makes `ProtocolParamsForSlot` a *reader* that writes into the published snapshot's `CostModels` map.

Below, this walks forward applying `nextEra.HardForkFunc(cfg, pparams)` to `snapshot.currentPParams`. `eras/babbage.go:130` does `ret := babbage.UpgradePParams(*alonzoPParams)` — which copies the struct but **shares the `CostModels` map** — and then `eras/babbage.go:138` writes `ret.CostModels[1] = DefaultPlutusV2CostModel` into it. I proved the mutation: `HardForkBabbage(nil, input)` grows `input.CostModels` from 1 to 2 entries. `HardForkConway` (`conway.go:104-106`, via `UpdateFromGenesis`) and `HardForkDijkstra` share the shape.

Failure: two goroutines calling `ProtocolParamsForSlot` for a post-fork slot concurrently ⇒ concurrent map writes into the shared snapshot map ⇒ `fatal error: concurrent map writes`, which is an unrecoverable runtime crash, not a `-race` warning.

**Not introduced here** — base `3806f3d7` released `RLock` before calling `HardForkFunc`, and `RLock` permits concurrent readers anyway, so the exposure is identical. Fixing it means copy-on-write in the `HardForkFunc` wrappers (`maps.Clone` the cost models before writing). Reasonable as a follow-up, but this PR is the moment the surrounding code starts describing these snapshots as immutable.

chrisguiney (changes requested):
Reviewed the concurrency properties in depth — publish completeness, snapshot aliasing, reader pairing, and test/doc accuracy — verifying the load-bearing claims by breaking invariants to see whether anything caught it.

**The design is sound and I found no live correctness bug.** Every writer publishes before unlocking (I traced all ~30 mutation sites across the eight snapshot-backed fields to one of the 12 publish call sites; the two riskiest shapes — the `defer`-based publish on the HFI tally covering three early returns, and the three early returns preceding the mutation in `verify_header.go` — are both handled correctly, and no writer exists outside that list). The `epochCache` shared-by-reference optimization is the subtlest thing here and it holds: all five writers genuinely replace the slice, and `advanceEpochCache`'s `make`+`copy`+`append` provably cannot touch a published array. `cloneEpoch`/`cloneTip`/`cloneSnapshotBytes` are complete against the real type definitions, and `transitionInfo`/`currentEra` are genuinely flat. `TestLedgerStatePairedSnapshotsUseOneGeneration` earns its place — removing the generation comparison makes it fail 5/5 runs under `-race` in under 0.2s.

Verified locally: `go build ./...` clean; `golangci-lint run ./ledger/...` → **0 issues**; `go test -race -count=1 ./ledger/` → **pass (294s)**; conformance suite → **pass**.

Requesting changes on two things, neither of which needs a redesign:

**1. Nothing pins the copy-on-write invariant.** I patched the real writer at `verify_header.go:1284-1286` to `ls.epochCache = append(ls.epochCache, newEpoch)` — violating the invariant the whole design rests on. `go build` succeeded and **`go test -race ./ledger/` passed in 272s**. Reproduced independently (389s, pass). The immutability test can't catch it because it performs the copy-on-write itself (see inline comment).

**2. ARCHITECTURE.md states the opposite of what the code does** — it tells readers epoch nonces are copied at publication when `epochCache` is published by reference. That's the one rule a future maintainer most needs, and the doc actively misleads on it.

Also worth fixing while this is open: the "Published snapshots are immutable" language is false for `currentPParams`/`prevEraPParams` (details inline). That aliasing is **pre-existing, not a regression** — I checked base `3806f3d7`, where `ProtocolParamsForSlot` released `RLock` before calling `HardForkFunc`, and `RLock` permits concurrent readers anyway — but this PR is what turns it into a documented guarantee people will rely on.

Two related pre-existing issues I couldn't anchor inline (files not in this diff), offered as follow-ups rather than this PR's debt:

- **`ledger/eras/babbage.go:130-138` and `conway.go:104-106`: a lock-free reader writes into a published snapshot's map.** `babbage.UpgradePParams(*alonzoPParams)` copies the struct but shares the `CostModels` map, then the wrapper writes `ret.CostModels[1] = DefaultPlutusV2CostModel` into it. I proved the mutation: `HardForkBabbage(nil, input)` grows `input.CostModels` from 1 to 2 entries. `ProtocolParamsForSlot` calls this on `snapshot.currentPParams`, so two concurrent callers for a post-fork slot means `fatal error: concurrent map writes` — a hard crash, not a detectable race. `HardForkConway`/`HardForkDijkstra` share the shape.
- **`database/pparams.go:186`** claims `ComputeAndApplyPParamUpdates` works "without mutating the input." That is factually wrong for every `PParamsUpdateFunc` — `Update` has a pointer receiver and assigns in place.

Finally, per CLAUDE.md this change class (header verification, nonce computation, epoch boundaries, rollback/chain selection) requires DevNet. You state you ran the full suite and it passed; nothing in the repo records it, so I'm noting it as unverified rather than doubted — a trailer or CI record would make that checkable for the next reviewer.

ledger/state_snapshot_test.go
arepala-uml: I have added a new test `TestAdvanceEpochCachePreservesPublishedSnapshot`, which retains the previously published consensus snapshot and invokes the real `advanceEpochCache()` production writer. The test verifies that the new snapshot contains the next epoch while the retained snapshot remains unchanged.

```
akrepala@akrepalas-MacBook-Pro dingo % go test -race -run TestAdvanceEpochCachePreservesPublishedSnapshot ./ledger
ok      github.com/blinklabs-io/dingo/ledger    1.956s
```

ledger/state.go
arepala-uml: I have added regression test above where it exercises the real `advanceEpochCache()` writer and verifies that a previously published snapshot remains unchanged. I also corrected `ARCHITECTURE.md` and documented that `healEmptyLabNoncesInPlace()` must only receive a freshly owned, unpublished epoch cache because it mutates the slice in place.

ARCHITECTURE.md
arepala-uml: I have addressed it where the architecture documentation now accurately describes epoch-cache sharing and its copy-on-write requirement, including nested nonce slices. I also clarified the construction/startup exception to writer locking and scoped the paired-generation guarantee to initialized production states, explicitly documenting the zero-value white-box fallback.

ledger/state.go
arepala-uml: I have fixed by cloning protocol parameters before epoch-rollover updates, so mutations apply only to a private copy. Added regression tests and verified the full ledger race suite, conformance, and clean DevNet pass.

ledger/state.go
arepala-uml: I have removed the nil fallbacks, so snapshot readers can no longer return mismatched or fabricated state. The chainsync fixture now updates the tip through a safe helper that publishes both snapshots with the same generation.

arepala-uml: As i mentioned above, After removing of nil fallbacks in snapshot readers updated the affected test fixtures to publish snapshots after changing snapshot-backed fields. This ensures tests use the same atomic snapshot path as production.

ouroboros/chainsync_test.go
arepala-uml: The reflect/unsafe snapshot clearing and nil fallback path have been removed. The fixture now updates the tip through `SetTipForTesting()`, which safely publishes a consistent snapshot generation, so the test exercises the production atomic read path. Race tests pass.

ledger/state.go
arepala-uml: I have made changes where `setEpochCache() `now defers snapshot publication, so every success and error return publishes the resulting state. I also added a regression test covering the unknown-era error path and confirmed the focused ledger race tests pass.

ledger/state_snapshot_test.go
arepala-uml: I have renamed and clarified the test as a race-regression guard against returning to live writer-owned field reads. The paired-generation test remains the primary validation for cross-snapshot consistency.

ledger/state.go
arepala-uml: I have corrected the comment to acknowledge that snapshot readers may already be active. It now accurately explains that the lock serializes publication with the asynchronous HFI tally and other snapshot writers.

ledger/state.go
arepala-uml: I have added conditional publication flag, which is set only after transitionInfo is updated. The generation-guard and existing TransitionKnown early returns now unlock without cloning snapshots, incrementing snapshotGeneration, or causing unnecessary paired-reader retries. Focused transition race tests pass.

ledger/state.go
arepala-uml: I have already addressed this issue with previous changes and the Babbage, Conway, and Dijkstra hard-fork wrappers now deep-clone `CostModels` before modification. I also added a concurrent `ProtocolParamsForSlot()` test that exercises the real Alonzo-to-Babbage forecasting path with 16 readers.

arepala-uml:
@cubic-dev-ai review

arepala-uml:
Ran devnet test suite again and all tests got passed after addressing all the comments.
[devnet_test2.log](https://github.com/user-attachments/files/30106214/devnet_test2.log)

arepala-uml:
@cubic-dev-ai review

chrisguiney (approved):

PR #2856, opened 2026-07-16
fix: dial ledger-discovered pool relays by resolved IP, not hostname

### LINKED ISSUE #2435: security: dial ledger-discovered pool relays by resolved IP, not hostname
## Problem

Ledger-discovered pool relays may be configured as hostnames (`SingleHostName` relays). The peer governor resolves the hostname once to perform the `isRoutableAddr` check, but the actual outbound dial uses the original hostname string, which Go's resolver looks up again at dial time.

A pool operator running a malicious or compromised DNS server can return a routable public IP on the first lookup (passing the routability check) and an internal IP on the second lookup (the actual dial), causing the node to TCP-probe internal addresses.

Evidence:
- peergov/peers.go:229-257 in `resolveAddress` resolves the hostname once for the normalized routability path.
- peergov/connections.go:261-264 calls `CreateOutboundConn(ctx, peer.Address)` where `peer.Address` is the original hostname for `SingleHostName` ledger relays, triggering a second resolution.

Impact is bounded: a stake pool must be registered on chain (real ADA cost, public identity) and the probed connection must complete an Ouroboros handshake with the right NetworkMagic to do anything useful. The attacker can learn reachable internal addresses via connection success/failure timing.

## Acceptance criteria

- Pool relays sourced from the ledger are resolved once and the resolved IP is what gets dialed.
- If a relay has multiple A or AAAA records, the choice is locked in for the connection attempt.
- The original hostname is still tracked for logging and for deny-list comparison.
- A test verifies that the IP used for `isRoutableAddr` is the same IP that ends up at the dialer.

## Notes

Operator-supplied topology relays are out of scope; those are operator-controlled. This is only about ledger-discovered relays where the hostname is supplied via on-chain stake pool registration.

cryptodj413:
Closes #2435 

---
## Summary by cubic
Dial ledger-discovered pool relays by a resolved IP (not hostname) to block DNS-rebind and always use the routability-checked address. Discovery now picks an IP family the host can dial to avoid pinning unreachable records; tests and docs updated.

- **Bug Fixes**
  - Lock IPs via `resolveLedgerDialTarget`: reuse `NormalizedAddress` if already an IP; else resolve once (with timeout), filter to locally supported v4/v6, verify routable, persist; error on resolver failure or unroutable result. Only applies to `PeerSourceP2PLedger`; others keep per-attempt re-resolution. `createOutboundConnection` dials exactly the checked IP.
  - Prefer supported IP family at discovery: `addLedgerPeer` uses `resolveLedgerDiscoveryAddress` to store a dialable A/AAAA result, avoiding pinning v4-only hosts to AAAA (or vice versa).
  - Prevent peer identity collisions: skip write-back if another peer already owns the resolved `NormalizedAddress`; still dial the checked IP and re-resolve on the next attempt.

<sup>Written for commit 092008b2f1294fe91337d3ffd38b3846c4bdce4e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2856?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security Enhancements**
  * Ledger-discovered relay addresses are now resolved once and reused for subsequent connection attempts.
  * Prevented DNS rebinding from redirecting connections to internal or unroutable addresses.
  * Unroutable or failed address resolution is rejected safely.

* **Documentation**
  * Updated peer governance documentation to describe the new relay resolution behavior and its trade-offs.

### DIFF
--- a/peergov/peers.go
+++ b/peergov/peers.go
@@ -341,6 +341,69 @@ func (p *PeerGovernor) resolveDialAddress(
 	return net.JoinHostPort(ip.String(), port)
 }
 
+// resolveLedgerDialTarget returns the concrete IP:port to dial for a
+// ledger-discovered pool relay peer, locking that choice onto the peer so
+// every subsequent dial attempt reuses the same resolved IP instead of
+// asking DNS again.
+//
+// Ledger relay hostnames are supplied by pool operators via on-chain stake
+// pool registration — input outside the node's control. If the routability
+// check performed when the relay was discovered (isRoutableAddr, via
+// resolveAddress in addLedgerPeer) and the actual dial each triggered their
+// own DNS lookup, a malicious or compromised authoritative DNS server could
+// answer the discovery-time lookup with a public IP (passing the check) and
+// a later lookup with an internal address (the one actually dialed) — a DNS
+// rebind that would cause the node to TCP-probe internal addresses.
+// Resolving exactly once and always dialing that resolved IP closes the
+// gap.
+//
+// peer.NormalizedAddress already holds the IP resolved at discovery time in
+// the common case, so this returns it unchanged. If that earlier resolution
+// failed and NormalizedAddress is still a hostname, a single fresh
+// resolution is performed here, checked for routability, and — on success —
+// written back to peer.NormalizedAddress under p.mu so it is locked in for
+// this and every future reconnect attempt. It must NOT be called while
+// holding p.mu.
+func (p *PeerGovernor) resolveLedgerDialTarget(
+	ctx context.Context,
+	peer *Peer,
+) (string, error) {
+	host, port, err := net.SplitHostPort(peer.NormalizedAddress)
+	if err != nil {
+		return "", err
+	}
+	if net.ParseIP(host) != nil {
+		// Resolved and locked in at discovery time.
+		return peer.NormalizedAddress, nil
+	}
+
+	// Discovery-time resolution didn't produce an IP (e.g. it failed and
+	// fell back to the lowercased hostname). Resolve once now, bounded so a
+	// hung or slow resolver cannot wedge the dial loop.
+	lookupCtx, cancel := context.WithTimeout(ctx, dialDNSResolveTimeout)
+	defer cancel()
+	ips, err := lookupIPAddr(lookupCtx, host)
+	if err != nil {
+		return "", err
+	}
+	if len(ips) == 0 {
+		return "", errors.New("no addresses returned for ledger relay hostname")
+	}
+	resolved := net.JoinHostPort(ips[0].String(), port)
+	if !isRoutableAddr(resolved) {
+		return "", ErrUnroutableAddress
+	}
+
+	p.mu.Lock()
+	idx := p.peerIndexByAddress(peer.NormalizedAddress)
+	if idx != -1 && p.peers[idx] != nil {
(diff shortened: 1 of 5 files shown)


peergov/peers.go
```diff
@@ -341,6 +341,90 @@ func (p *PeerGovernor) resolveDialAddress(
...
+	ctx context.Context,
+	peer *Peer,
+) (string, error) {
+	host, port, err := net.SplitHostPort(peer.NormalizedAddress)
+	if err != nil {
+		return "", err
+	}
+	if net.ParseIP(host) != nil {
+		// Resolved and locked in at discovery time.
+		return peer.NormalizedAddress, nil
```
wolf31o2, line 425: **Blocking:** this trusts the IP selected earlier by `resolveAddress`, but that function still takes `ips[0]` without applying `supportedDialFamilies`/`filterDialFamilies`. Therefore the fallback path is hardened, while the normal discovery path is not. On a v4-only node with DNS answers `[AAAA, A]`, the peer is locked forever to the AAAA record and never tries the usable A record. I reproduced this locally: expected `198.51.100.9:3001`, got `[2001:db8::1]:3001`.

Please filter the records during initial ledger discovery before `NormalizedAddress` is stored. An example regression test is:

```go
func TestAddLedgerPeerPrefersLocallySupportedFamily(t *testing.T) {
    setLocalAddrFamilies(t, true, false) // v4-only host

    oldLookupIP := lookupIP
    lookupIP = func(string) ([]net.IP, error) {
        return []net.IP{
            net.ParseIP("2001:db8::1"), // AAAA returned first
            net.ParseIP("198.51.100.9"),
        }, nil
    }
    t.Cleanup(func() { lookupIP = oldLookupIP })

    pg := newDialSpreadGovernor()
    require.True(t, pg.addLedgerPeer("relay.example.com:3001"))
    require.Len(t, pg.peers, 1)

    got, err := pg.resolveLedgerDialTarget(t.Context(), pg.peers[0])
    require.NoError(t, err)
    require.Equal(t, "198.51.100.9:3001", got)
    require.Equal(t, "198.51.100.9:3001", pg.peers[0].NormalizedAddress)
}
```

The implementation should ideally make family-aware selection explicit for ledger discovery rather than changing generic hostname normalization behavior for every peer source.

wolf31o2 (changes requested):
No-go as written: the common ledger-discovery path still locks the first DNS answer without filtering it to a locally supported address family. On an IPv4-only node, an AAAA-first response permanently pins the relay to IPv6 even when a valid A record is present. This regresses the current per-attempt family filtering and contradicts the hardening described by this PR. Please choose a locally supported family during initial ledger discovery (before storing `NormalizedAddress`) and cover that normal path with a regression test. Targeted race tests, vet, and import-boundary checks otherwise passed; `ARCHITECTURE.md` is updated appropriately and `DATABASE.md` is not affected.

wolf31o2 (approved):

Issue #2819, opened 2026-07-14
leios: from-genesis catch-up stalls for minutes when a relay is slow to serve a referenced endorser block; add per-request fetch deadline and failover


wolf31o2:
Observed on a from-genesis Musashi (prototype-2026w27) sync with current main (merge-base 0c3fd0a0) and reproduced identically on base a9288113 via an A/B run, so this is pre-existing behavior and not tied to any feature branch.

During the endorser-block era of a from-genesis sync, the ledger tip repeatedly freezes for several minutes (observed roughly 5 to 8 minutes per stall) while chainsync stays healthy and CPU is near idle. Each stall clears on its own once the relay serves the endorser block, or when the connection is cycled; a full process restart also clears it. It is not a hard wedge, but it turns a from-genesis catch-up into a multi-hour, stop-and-go process (the EB-era throughput drops to bursts of a few thousand slots separated by these pauses).

Goroutine evidence during a stall (the ledger apply loop is parked waiting for a single referenced endorser block):

```
ledgerProcessBlocks (ledger/state.go)
 -> ledgerProcessBlocksFromSource
   -> ensureReferencedEndorserBlocks (ledger/leios_apply.go:368)
     -> awaitFetch (ledger/leios_apply.go:762)   [select]
fetch goroutine:
 leiosBackfiller.spawn -> Ouroboros.FetchEndorserBlockByPoint (ouroboros/leios_backfill.go:63)
   -> fetchEndorserBlockOnConn (ouroboros/leios_backfill.go:138)
     -> gouroboros leiosfetch.Client.BlockRequest   [chan receive]   (gouroboros v0.187.3)
```

The referenced EB request is dispatched to a connected relay that accepts the connection but does not promptly return the block. awaitFetch has an overall leiosBackfillMaxWait deadline, and fetchEndorserBlockOnConn records a per-connection cooldown on failure, but a single in-flight BlockRequest to a slow-but-alive relay can hold the fetch until that connection eventually errors or the overall max wait elapses, rather than being proactively re-dispatched to another already-connected peer. While the ledger waits, speculative endorser-block application keeps retrying and logs a large volume of "input UTxO not found" warnings (about 1500 per 30s in one run) against transactions it cannot yet resolve, which dominates the log during the stall.

Suggested direction:
[...]

Issue #2629, opened 2026-06-24
leios: complete a partially-fetched endorser block's tail on relay re-offer


wolf31o2:
When dingo fetches an endorser block near the live tip, the relay diffuses the block's transactions over several seconds, so the last partial transaction window is often not available yet when dingo requests it. fetchLeiosEbTxsBatched (ouroboros/leiosnotify.go) re-requests the still-missing tail for up to the Leios diffusion window (LeiosTxFetchTailBudget, sourced from DiffuseWindowSlots) and then gives up, returning the partial prefix. The transactions offer handler drops that partial result and caches nothing, so a fraction of endorser blocks per active window are never completed or applied, and their outputs are missing from the UTxO set near tip. Roughly half the endorser blocks in an active window currently miss this way.

Current behavior: a fetch that does not complete within the tail budget is discarded. If the relay re-offers the same endorser block later, dingo re-fetches it from scratch. Simply enlarging the per-fetch budget is not a good fix, because the fetch holds the per-connection fetch slot while it waits and that starves other endorser blocks during a busy window.

Required: cache the partial endorser block together with the bitmap of transactions still missing, and when the relay re-offers the same endorser block (the leios-notify transactions offer), fetch only the missing tail and complete the cached entry. This frees the connection between attempts and lets a block complete across re-offers without re-fetching the transactions dingo already holds. Once complete, the block is applied through the existing tip gate unchanged.

Acceptance criteria:
- a partially fetched endorser block is retained with its missing-transaction bitmap instead of discarded
- a re-offer of the same endorser block fetches only the still-missing transactions and completes the cached entry
- a completed-by-re-offer endorser block is applied through the existing gate, with no change to the diffusion-window bound on a single fetch attempt
- a test drives a partial fetch followed by a re-offer that completes the block and asserts the assembled transactions are correct and in order

This is a near-tip throughput improvement for from-scratch and at-tip operation on the musashi network. It does not address the historical endorser block gap (the relay diffuses only current-tip endorser blocks and BlockRangeRequest is stubbed), which is a separate prototype limitation.

PR #2866, opened 2026-07-17
fix(leios): wait for endorser closure before serving CertRB over NtC

### LINKED ISSUE #2817: leios: NtC chainsync serves a raw CertRB with empty transactions when the certified endorser block closure is not yet available
## Summary

Over node-to-client chainsync, dingo serves a certifying ranking block (CertRB) verbatim when the certified endorser block's transaction closure is not yet available locally. Because the on-chain CertRB carries empty transaction segments (the transactions live in the certified endorser block), a local client streaming the chain during that window sees the CertRB with zero transactions. The prototype avoids this by never presenting a CertRB over N2C without its closure.

## Current behavior

The N2C chainsync blocks server inlines the certified endorser block's transactions into a CertRB before serving it:

- `ouroboros/leios_merged.go` `chainsyncServerBlockCbor` -> `mergedLeiosRankingBlockCbor` -> `resolveCertifiedEndorserTxs`.
- `resolveCertifiedEndorserTxs` correctly gates on `header.LeiosCertified()` so only CertRBs are spliced (this matches the prototype's cert gate). For a CertRB it resolves the parent's `leios_announcement` (prevAnn) to find the endorser block hash, then looks it up in the local EB cache.
- If the endorser block is not cached yet or its tx cache is incomplete (`!found || !data.completeTxCache()`), it returns `ok=false`.
- `mergedLeiosRankingBlockCbor` then returns the raw block unchanged, and `chainsyncServerBlockCbor` serves the on-chain CertRB as-is.

Since the on-chain CertRB has empty transaction segments, the N2C client receives a block with no transactions for that CertRB. A block is streamed once as the client follows the chain, so if it is served empty during this window the client is not prompted to re-fetch it, and its derived state stays missing the certified transactions until it re-syncs that block.

## What the prototype does

In ouroboros-leios prototype-2026w27b / prototype-2026w28 (ouroboros-consensus):

- The N2C chainsync server splices only for a CertRB (`headerContainsLeiosCert`), same gate dingo uses. See `MiniProtocol/ChainSync/Server.hs` `chainSyncBlocksServer` (commits 4282e730e / 8257193d2, "restore certificate gate on EB inlining").
- On a missing closure, `resolveLeiosClosure` (`Shelley/Ledger/Leios.hs`) was changed from silently returning an empty tx list to erroring loudly (commit e45c0d9e4, "error loudly when EB closure is missing at cert-RB apply"). The comment notes the intended design: chain selection parks a CertRB pending closure acquisition, so the server should never be asked to serve or apply a CertRB without its endorser block closure.

So the prototype never streams an empty CertRB over N2C: it either has the closure or it has not yet selected/served the block.

## Impact

- Deployed path is Musashi prototype-2026w27, magic 164. CertRB activity is bursty and sparse, so the window is narrow, but a client that streams a CertRB during it (for example db-sync) records that block with no transactions and does not automatically correct it.
- This only affects the N2C serve view. It does not affect dingo's own UTxO state: on the Musashi/Haskell path endorser-block transactions are not applied to the UTxO.

## Recommended approach

Defer serving a CertRB over N2C until its certified endorser block closure is available locally: park the N2C RollForward for that point until `resolveCertifiedEndorserTxs` succeeds, then serve the spliced block. This matches the prototype's parking intent and keeps clients from ever seeing a CertRB with empty transactions. Because a block is streamed once as the client follows the chain, deferring is the only option that guarantees the client's view is complete without relying on a later re-sync.

Implementation notes:

- Gate the N2C RollForward for a CertRB on the certified endorser block closure being present (`resolveCertifiedEndorserTxs` returning ok=true), rather than falling through to the raw block on a miss.
- Bound the wait so a closure that never arrives cannot stall the N2C server indefinitely: kick off / rely on the backfill fetch for the endorser block, and cap the park with a timeout. Decide the fallback on timeout (keep waiting, or surface an error to the client) rather than silently serving an empty CertRB.
- Keep the existing serve-raw path only as the malformed-CBOR safety net (splice error), not as the missing-closure path.

## Alternative

Keep serving raw on a miss but make the gap observable and bounded: trace/metric when a CertRB is served without its closure, and rely on backfill plus a client re-sync to correct it. This is simpler but leaves a window where an N2C client records a CertRB with no transactions and is not prompted to re-fetch it, so it is a fallback rather than the target behavior.

## References

- dingo: `ouroboros/leios_merged.go` (`resolveCertifiedEndorserTxs`, `mergedLeiosRankingBlockCbor`, `spliceEndorserTxsIntoDijkstraBlock`, `chainsyncServerBlockCbor`).
- ouroboros-consensus prototype-2026w27b/w28: `Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs`, `Ouroboros/Consensus/Shelley/Ledger/Leios.hs`.

chrisguiney:
## Problem

Closes #2817.

Before serving a Dijkstra block over node-to-client chainsync, `chainsyncServerBlockCbor` splices the certified endorser block's (EB) transaction closure into the certifying ranking block's (CertRB) empty transaction segment. The closure is populated asynchronously by the leiosnotify/leiosfetch **client** path and cached in `leiosEndorserBlocks`. `resolveCertifiedEndorserTxs` returned the same "not merged" signal both when a block was **not** a CertRB and when it **was** a CertRB whose closure was not yet cached. On a cache miss the raw on-chain CertRB was served with empty transaction segments, so clients (e.g. db-sync) recorded a block with no transactions and could not self-correct without a re-sync.

## Approach

Bounded wait + observability, with an explicit timeout fallback (no active cross-peer backfill fetch — that needs EB slot/point discovery and overlaps #2629 / #2819).

- Extract `certifiedEndorserBlockHash` from `resolveCertifiedEndorserTxs` (decode Dijkstra header → `header.LeiosCertified()` → resolve parent via prev-hash → read parent `leios_announcement`). This lets the serving path distinguish "CertRB with a missing closure" from "not a CertRB", and learn which EB hash to wait on. `resolveCertifiedEndorserTxs` is behaviorally unchanged (now a thin wrapper).
- Bounded-wait primitives guarded by the existing `leiosMu`: `leiosClosureWaiters` (per-EB channels, eagerly allocated), `waitForLeiosEndorserClosure` (fast-path check → register waiter → select vs `ctx.Done()` → deregister + re-check on cancel), and `storeLeiosEndorserBlock` wakes waiters once a complete closure is cached.
- `chainsyncServerBlockCbor`: merged → serve; not merged + CertRB → `serveLeiosCertRbWithWait` (`context.WithTimeout(LeiosClosureWaitTimeout)`, default 3s). Resolved during the wait → serve merged; timeout → serve raw + warn. Non-CertRB / decode error → serve raw (unchanged).
- Metrics: `dingo_leios_ntc_certrb_serves_total{outcome=merged|merged_after_wait|raw_unresolved}` and `dingo_leios_ntc_certrb_closure_wait_seconds{outcome=resolved|timeout}`.

## Testing

- New unit tests (no `time.Sleep`; `testutil.RequireReceive`): `certifiedEndorserBlockHash` guards (non-CertRB / no-ledger), `waitForLeiosEndorserClosure` (already-cached, wakes-on-store, timeout + waiter cleanup), and `awaitMergedLeiosRankingBlock` timeout.
- The positive merge-after-wait path is covered compositionally by the existing splice/announcement/resolve tests plus the wakeup test; a ledger-seeded end-to-end test was skipped as fragile (seeding a `BlockByHash`-resolvable announcing parent) with no coverage gain.
- `go build ./...`, `golangci-lint run ./...` (clean), `modernize ./ouroboros/...` (no findings in changed files), `go test ./ouroboros/... -race`, and the ledger-rules conformance suite all pass.

**Note:** the change is a guarded no-op for all non-Leios / non-CertRB traffic. Full end-to-end validation of the CertRB serving path requires a Musashi/Leios (magic 164) environment; the stock DevNet cardano-node does not produce CertRBs.

## Docs

- `ARCHITECTURE.md`: added a "Leios CertRB Serving (NtC)" subsection under Multi-Client Chainsync.
- `DATABASE.md`: checked, not affected (no model/table/query/encoding changes).


chrisguiney:
Addressed the bot review in `4e0e1c18`:

- **Closure readiness consistency** (CodeRabbit + cubic): dropped the `len(txsRaw) > 0` guard from the wait-signal and `leiosClosureCompleteLocked`, so waiter readiness uses the same `completeTxCache()` predicate as the resolver. gouroboros rejects zero-reference endorser blocks on decode (`NewLeiosEndorserBlockFromCbor` → "at least one transaction reference"), so a complete-but-empty closure can never be cached — this is behavior-preserving for real inputs and just removes the inconsistency.
- **Wakeup test race** (CodeRabbit + cubic): `TestWaitForLeiosEndorserClosureWakesOnStore` now waits (`testutil.WaitForCondition`) for the waiter to register before storing, so it deterministically exercises the wake-on-signal path instead of the fast path.
- **Metrics coverage** (cubic): added `TestLeiosCertRbMetricsRecordOutcomes` (asserts every serve/wait outcome via `prometheus/testutil`) and `TestLeiosCertRbMetricsNilSafe` (no-op when the registry is absent).
- **Docs** (cubic): the ARCHITECTURE.md subsection now lists the full metric contract (all `serves_total` outcomes + the `closure_wait_seconds` histogram).

On the **"still serves a raw CertRB on timeout"** pre-merge warning — this is intentional and is the explicit timeout fallback #2817 asked to define. The bounded wait resolves the common near-tip race (the closure lands via the async leiosnotify/leiosfetch path within the window). The suggested alternatives are worse for a chainsync server: parking a CertRB indefinitely stalls the client's entire chain when a closure never arrives (e.g. a historical re-sync of TTL-evicted closures), and NtC chainsync has no per-block error channel to surface a timeout without tearing down the connection. Raw-serve on timeout keeps the client progressing while the new `dingo_leios_ntc_certrb_serves_total{outcome="raw_unresolved"}` counter + warning make every incomplete serve observable. Active server-side backfill (fetching the missing closure from a peer during the wait) is deliberately out of scope here — it needs endorser-block slot/point discovery and overlaps the existing EB-backfill work in #2629 / #2819.

wolf31o2 (changes requested):
No-go in the current form. The waiter mechanics and observability look sound, but the production timeout is shorter than the repository's documented healthy Musashi closure-delivery timing, and the timeout path still sends the raw zero-transaction CertRB. That leaves downstream NtC clients permanently incomplete until they re-sync, so the PR does not close #2817's correctness gap.

ouroboros/leios_merged.go
```diff
@@ -36,8 +37,22 @@ import (
 const (
 	leiosEndorserBlockCacheMaxEntries = 1024
 	leiosEndorserBlockCacheTTL        = 10 * time.Minute
+	// defaultLeiosClosureWaitTimeout bounds how long the NtC chainsync
+	// server waits for a certifying ranking block's endorser block
+	// transaction closure to be cached before falling back to serving the
+	// raw block. CertRB activity is sparse, so a short wait covers the
+	// common near-tip race between the async leiosnotify/leiosfetch client
+	// path and NtC delivery.
+	defaultLeiosClosureWaitTimeout = 3 * time.Second
```
wolf31o2: Blocking: this 3s default is shorter than the repository’s documented healthy Musashi path (~3s transaction-offer delay plus ~3s fetch; see `node.go` around the `EndorserBlockWaitSlots` wiring). `node.go` does not override `LeiosClosureWaitTimeout`, so normal healthy delivery can cross this cutoff and enter `raw_unresolved`. Please derive this from the existing network/pipeline timing (the ledger path uses the 20-slot certify deadline), rather than using an independent shorter constant.

ouroboros/leios_merged.go
```diff
@@ -555,13 +673,63 @@ func (o *Ouroboros) chainsyncServerBlockCbor(
...
+	o.recordLeiosCertRbWait("timeout", waited)
+	o.config.Logger.Warn(
+		"serving CertRB with unresolved endorser closure over NtC chainsync; client may record a block with no transactions",
 		"slot", block.Slot,
 		"hash", hex.EncodeToString(block.Hash),
+		"eb_hash", ebHash.String(),
+		"waited", waited,
 	)
-	return merged
+	return block.Cbor
```
wolf31o2: Blocking: do not send the raw CertRB when closure resolution times out. An NtC chainsync client sees this point once and is not prompted to fetch it again, so it permanently records zero transactions until a re-sync—the exact failure described by #2817. The bounded terminal path should surface an error/disconnect so the client retries from the same point (or otherwise park safely), not call `RollForward` with `block.Cbor`. Extending the wait may reduce frequency, but a raw fallback still violates the correctness contract.

ouroboros/leios_merged.go
chrisguiney: Fixed in `5659fc6e`. The wait is now derived from the same Leios pipeline timing the ledger uses, not an independent constant: added `LedgerState.EndorserBlockWaitDuration()` = `EndorserBlockWaitSlots` (the certify-by deadline) × `shelleySlotLength`, the same window `ensureReferencedEndorserBlocks` uses to gate applying a referenced EB. The serving path reads it directly from `LedgerState`, so `node.go`'s existing `EndorserBlockWaitSlots` wiring is the single source of truth and no new override is needed. `LeiosClosureWaitTimeout` is now an optional override only; the fallback default (used only when ledger timing is unavailable) is 20s, not shorter than the ~6s healthy path.

ouroboros/leios_merged.go
chrisguiney: Fixed in `5659fc6e`. On timeout the serving path now returns `errLeiosClosureUnresolved` instead of `block.Cbor`; both call sites in `chainsyncServerRequestNext` (the synchronous immediate-block path and the async goroutine) surface it so the connection is closed rather than `RollForward`-ing an incomplete block. gouroboros' `handleRequestNext` propagates the `RequestNextFunc` error and tears down the protocol, so the client reconnects and retries from its last point (the CertRB's parent) and never records a zero-transaction block. Combined with the pipeline-derived (longer) wait, timeouts should be rare, and when they do happen the client retries instead of persisting incomplete state. A structurally malformed CertRB (splice failure, not a missing closure) still falls back to raw as a CBOR-safety valve. Metric renamed to `dingo_leios_ntc_certrb_total{outcome=merged|merged_after_wait|unresolved}` since a timeout is no longer a serve.

ouroboros/chainsync.go
```diff
@@ -563,9 +570,22 @@ func (o *Ouroboros) chainsyncServerRequestNext(
 				)
 			}
 		} else {
+			blockCbor, blockErr := o.chainsyncServerBlockCbor(ctx, next.Block)
+			if blockErr != nil {
+				// Do not RollForward an incomplete CertRB; surface the error so
+				// the connection is closed and the client retries from its last
+				// point once the endorser closure is available.
+				o.reportChainsyncServerAsyncError(
```
wolf31o2: Blocking: this async timeout path does not actually tear down the connection. `reportChainsyncServerAsyncError` only performs a nonblocking send to `conn.ErrorChan()`. The connection-manager watcher consumes that value, removes the connection, and publishes `connmanager.conn_closed`, but it never calls `conn.Close()`; if the send is dropped, even that bookkeeping does not happen. Because this runs after `RequestNextFunc` has already returned, gouroboros cannot propagate `blockErr` through the callback-owned teardown path, so the NtC client can remain connected in `AwaitReply` instead of reconnecting and retrying the CertRB. Invoke an actual connection/protocol close mechanism here and add a regression test that observes transport/protocol closure, not only the connection-closed event.

ouroboros/leios_merged.go
```diff
@@ -555,35 +695,102 @@ func (o *Ouroboros) mergedLeiosRankingBlockCbor(
...
+		)
+		return merged, nil
 	}
-	o.config.Logger.Debug(
-		"serving merged Leios block over NtC chainsync",
+	// The block was not merged. If it is a certifying ranking block whose
+	// endorser closure is not cached yet, wait a bounded window for the async
+	// leiosnotify/leiosfetch client path to populate it. Non-CertRB blocks are
+	// served as-is.
+	ebHash, certified := o.certifiedEndorserBlockHash(block.Cbor)
```
wolf31o2: Blocking: `certifiedEndorserBlockHash` uses the same `false` result for a non-CertRB and for a CertRB whose parent lookup or parent announcement resolution fails. Consequently, a persistent `BlockByHash`/resolution error reaches `!certified` below and serves `block.Cbor` raw, reintroducing the zero-transaction CertRB that this PR is meant to prevent. Preserve certification separately from resolution success (for example, return an error/tri-state) and disconnect rather than raw-serve once the header has been identified as certified. Please cover the parent-resolution failure path.

wolf31o2 (changes requested):
No-go at the current head. The waiter/timing changes address the original raw-on-timeout behavior on the synchronous path, but two paths can still violate the NtC completeness/retry contract. Please address both blocking inline comments and add regression coverage that verifies the downstream transport actually closes and that a certified block is never downgraded to the raw serve path when parent resolution fails.

ouroboros/chainsync.go
chrisguiney: Fixed in `c80d4e6d`. Confirmed your analysis — `RemoveConnection` (ends before the recycle handler) does the map/IP-slot cleanup and publishes `conn_closed`, but never calls `conn.Close()`, so the error-channel send left the bearer open. Added `closeChainsyncServerConn`, which sends the error for connmanager bookkeeping **and** calls `conn.Close()` (idempotent) to actually tear down the transport; the async serving site now uses it instead of `reportChainsyncServerAsyncError`. Regression test `TestCloseChainsyncServerConnTearsDownTransport` stands up a real gouroboros connection pair over `net.Pipe` and asserts the **client end observes the transport closing** (its connection `ErrorChan` fires), not just the conn_closed event. (The synchronous path is unchanged: its callback-returned error reaches gouroboros `SendError`, which stops the protocol and tears down the connection.)

ouroboros/leios_merged.go
chrisguiney: Fixed in `c80d4e6d`. `certifiedEndorserBlockHash` is now tri-state — `(ebHash, certified, resolved)`. `certified` is set as soon as the header's `leios_certified` is true; a subsequent parent `BlockByHash` failure or missing announcement returns `certified=true, resolved=false` rather than a bare `false`. The serving decision (extracted into `serveLeiosRankingBlockCbor` so it is unit-testable without a live server) now treats certification as terminal: a certified block is **never** downgraded to the raw serve path — it is either merged or returns `errLeiosClosureUnresolved` so the connection closes and the client retries. Raw is served only for a genuine non-CertRB or the CBOR-safety case where a structurally malformed CertRB fails to splice. Coverage: `TestCertifiedEndorserBlockHashTriState` (certified=true/resolved=false on parent-resolution failure) and `TestServeLeiosRankingBlockCborDisconnectsOnUnresolvedCertifiedBlock` (a certified block with an unresolvable parent returns the error and never the raw bytes).

wolf31o2 (approved):

Issue #2709, opened 2026-07-02
Complete Ouroboros Genesis support for biased fast-sync sources


wolf31o2:
## Context

Dingo now has partial Genesis bootstrap plumbing:

- `genesisBootstrap` / `WithGenesisBootstrap` enables density-based chain selection during from-origin startup.
- `peerSnapshotFile` support can seed ledger peers when Genesis chain selection is active.
- `chainselection` tracks observed header density and exits Genesis selection once the local tip is within the configured window of the best known peer tip.

That is useful groundwork, but it is not yet a complete Ouroboros Genesis implementation with the trust properties expected by tools like the Genesis Sync Accelerator. GSA assumes the consuming node runs Genesis mode with peer-snapshot corroboration from large ledger peers. Dingo still needs the remaining protocol/security behavior defined and tested before we can document GSA-style sync as a supported trust model.

## Scope

- Define Dingo target behavior against cardano-node / ouroboros-consensus Genesis mode, including which pieces are required for security, which are performance-only, and which are not applicable to Dingo.
- Audit the current `chainselection` Genesis mode and identify gaps beyond observed-density peer-tip comparison.
- Implement or explicitly defer Genesis peer corroboration behavior for biased block-data sources such as a local shallow peer.
- Decide whether Dingo needs ChainSync Jumping, devoted BlockFetch dynamics, or equivalent behavior to match the expected Genesis sync security/performance model.
- Ensure a fast source is selected only while corroborated by other eligible peers; divergent or uncorroborated sources should stall, demote, disconnect, or be denied in a predictable way rather than silently driving the selected chain.
- Harden lifecycle behavior: from-origin gating, exit criteria, possible re-entry or recovery, bootstrap peer interaction, peer-snapshot loading, peer-snapshot refresh, and bad/stalling peer handling.
- Add observability for current selection mode, Genesis window, peer densities, selected fast source, corroboration failures, exit reason, and peer demotion/deny events.
- Document the supported trust model, including how this differs from normal Praos sync and from Mithril bootstrap.

## Acceptance criteria

- Dingo can be configured with a GSA-like `localRoots` / trustable local peer plus a ledger peer snapshot and has a documented security story for from-origin sync.
- A bad or divergent fast source cannot steer Dingo onto an untrue chain under the documented Genesis assumptions; it should at worst stall or be demoted.
[...]

PR #2896, opened 2026-07-20
feat(chainselection): corroborate Genesis fast sources before selection

chrisguiney:
## Summary

Addresses #2709. Implements the security-critical core of Ouroboros Genesis support for biased fast-sync sources: a **corroboration gate** so a fast (shallow) block source — e.g. a local shallow peer or the Genesis Sync Accelerator — is followed only while independent peers confirm its recent blocks.

Previously Genesis mode selected the densest peer with no corroboration, so a bad or divergent fast source could silently steer the local chain. Now an uncorroborated or divergent source is **denied selection** and the node **stalls** rather than following an untrue chain.

## What changed

**Corroboration gate (`chainselection/`)**
- Track a bounded recent `(slot, hash)` frontier per peer (`observedPoints`), maintained in lockstep with the existing `observedSlots`, for block-identity comparison (two-pointer merge on the sorted frontiers).
- A candidate is *corroborated* only when ≥ `MinCorroboratingPeers` other eligible/live/non-stale peers report an identical `(slot, hash)` point. In Genesis mode an uncorroborated candidate is denied selection (`isPeerSelectableLocked`); a divergent private-fork source shares no recent hash with honest peers, so it can never win. `0` disables the gate (density-only, historical default — existing Genesis tests unchanged).
- The gate denies selection but **keeps the fast source connected** so it can serve blocks once corroboration arrives.

**Observability**
- `GenesisStatus()` getter (mode, window, selected fast source, per-peer density/corroboration).
- New events `chainselection.genesis_corroboration_failed` and `chainselection.genesis_mode_exited` (with exit reason).

**Wiring**
- `WithGenesisCorroborationPeers` → `ChainSelectorConfig.MinCorroboratingPeers`.
- `genesisBootstrap.corroborationPeers` in config (yaml / `DINGO_GENESIS_BOOTSTRAP_CORROBORATION_PEERS` / `--genesis-bootstrap-corroboration-peers`), plus `dingo.yaml.example`.

## Docs

- `ARCHITECTURE.md`: new "Ouroboros Genesis trust model" subsection (corroboration, GSA `localRoots` + peer-snapshot config, trust-model comparison vs Praos/Mithril, observability, explicit deferred scope), the two new event-table rows, and a Peer-Governance cross-reference.
- `GENESIS_SYNC.md`: new operator runbook.
- `DATABASE.md`: checked, not affected.

## Tests

Added `chainselection/genesis_corroboration_test.go` covering honest fast source (corroborated → selected), divergent fast source (denied + failure event), insufficient corroboration / quorum-not-met (stall), disabled-by-default, `GenesisStatus` observability, mode-exit event, and frontier invariants (lockstep, rollback trim, hash-sensitive sharing). Extended the config env-var test.

`chainselection`, `peergov`, `internal/config`, root, and conformance suites pass with `-race`; `golangci-lint` and `modernize` clean.

## Explicitly deferred (per issue scope)

Documented, not silently dropped: full density-at-intersection, ChainSync Jumping / devoted BlockFetch dynamics, wiring peer-governance demotion to the corroboration-failure event, and a live DevNet/GSA integration run. These are performance/refinement; this PR delivers the from-origin **security** property that a fast source cannot steer the node without independent corroboration.

---
## Summary by cubic
Adds a Genesis corroboration gate so a fast (shallow) source only drives chain selection when independent peers confirm its recent blocks; enforcement happens at ledger apply so uncorroborated sources are observed but withheld. Tightens Genesis exit and `ouroboros` chainsync handling to prevent premature exit, stalls, or lost headers; default remains off.

- New Features
  - Corroboration gate: requires ≥ `MinCorroboratingPeers` distinct-host witnesses of recent `(slot, hash)`; uncorroborated sources are denied but stay connected. `0` disables; negative programmatic values clamp to `1` (fail-closed); negatives in config are rejected.
  - Per-peer recent `(slot, hash)` frontier tracked only while corroboration is active and cleared on Genesis exit; trimmed on rollback.
  - Ledger apply gate: withholds headers/rollbacks from uncorroborated peers via `chainselection.ShouldApplyIngress` and `ouroboros` `ChainsyncApplyEligible`; tips are observed first via `ChainsyncObservePeerTip` to avoid races.
  - Observability: `GenesisStatus()` snapshot and events `chainselection.genesis_corroboration_failed`, `chainselection.genesis_mode_exited`, and `chainselection.selected_none`.

- Bug Fixes
  - Genesis exit uses the advertised tip and only counts corroborated peers that have delivered headers up to within the window (`ObservedTip + window >= Tip`), preventing premature exit and horizon pinning by extreme slots.
  - Order fixes: observe tips before the apply decision; rollbacks are apply-gated; emit `chainselection.selected_none` on all best-peer → none transitions.
  - `ouroboros`: do not dedup headers withheld by the apply gate; record cross-peer dedup only for apply-eligible headers so a later corroborated peer can still publish the point.

<sup>Written for commit eeadbb695264fc218d646506eb16e512458e79b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2896?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable corroboration for Genesis-mode chain selection.
  * Fast-sync sources must be independently confirmed before being selected; otherwise, selection pauses to avoid divergent chains.
  * Added Genesis status visibility, including source density, corroboration counts, and selection readiness.
  * Added events for corroboration failures and exiting Genesis mode.
  * Added configuration support through YAML, environment variables, CLI flags, and programmatic options.

* **Documentation**
  * Added operator guidance for configuring, monitoring, and tuning Genesis synchronization and corroboration.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1278,53 +1278,84 @@ it. Dingo implements this as a **corroboration gate**
 (`chainselection/genesis_corroboration.go`), controlled by
 `genesisBootstrap.corroborationPeers` (`MinCorroboratingPeers`):
 
-- A candidate peer is **corroborated** when at least `corroborationPeers` other
-  eligible, live, non-stale peers report an identical `(slot, hash)` point in
-  their observed frontier — evidence they are following the same chain within
-  the Genesis window.
+- A candidate peer is **corroborated** when at least `corroborationPeers`
+  independent witness peers *confirm the candidate's recent chain*
+  (`confirmsRecentChain`): every block a witness observed within the candidate's
+  window slot range matches the candidate's own `(slot, hash)`, and they share at
+  least one such block. The observed frontier is populated per header during
+  chainsync (dense), so two peers on the same chain agree on every block in their
+  overlap. This is deliberately stronger than "share any common point": a fast
+  source that agrees on one old ancestor and then produces different blocks for
+  the rest of the window is **not** confirmed, because the witness observed
+  recent blocks the candidate lacks (or a conflicting hash at the same slot).
+- Witnesses are counted by distinct remote **host**, and a witness on the
+  candidate's own host is excluded, so several connections from one operator
+  cannot self-corroborate a private fork. This is a lower bound on independence,
+  not a guarantee: genuine independence (distinct operators, ASNs, and chain
+  views) depends on the operator's validated topology and peer-governance
+  diversity groups. Raising `corroborationPeers` raises the *count* required, not
+  the independence of the peers supplied — that remains an operator
+  responsibility.
 - In Genesis mode with `corroborationPeers > 0`, an **uncorroborated** candidate
-  is denied chain selection (`isPeerSelectableLocked`). A divergent fast source
-  on a private fork shares no recent block hash with any honest peer, so it is
-  never corroborated and cannot become the best peer. With no corroborated
-  peer, selection returns none and the node **stalls** rather than following an
-  untrue chain. This is the security property: a bad or divergent fast source
-  can at worst stall the node under the documented assumptions (at least one
-  honest corroborating peer, e.g. seeded from a ledger peer snapshot).
+  is denied chain selection (`isPeerSelectableLocked`). A fully divergent fast
+  source shares no recent block with any honest peer, so it is never corroborated
+  and cannot become the best peer. With no corroborated peer, selection returns
+  none and the node **stalls** rather than following an untrue chain: a bad or
+  divergent fast source can at worst stall the node under the documented
+  assumption of at least one honest, independent corroborating peer within the
+  window (e.g. seeded from a ledger peer snapshot). Corroboration fails **closed**
+  — a candidate whose window does not overlap any witness's frontier (e.g. a fast
+  source that has raced far beyond every corroborator's window) is not confirmed
+  and stalls, so operators must keep corroborators within the Genesis window of
+  the fast source.
 - The gate **denies selection but does not disconnect** the fast source. Genesis
   wants the fast source kept connected so it can serve blocks as soon as
   corroboration arrives; demoting or dropping it would defeat the accelerator.
   Peer governance may still subscribe to the failure event to react.
 - `corroborationPeers = 0` disables the gate (density-only Genesis selection,
   the historical default), preserving prior behavior for nodes that do not opt
-  into the Genesis trust model.
+  into the Genesis trust model. While the gate is disabled the per-peer hash
+  frontier is not tracked, so normal Praos operation carries no extra state.
 
 **Supported GSA-style configuration**: a trustable `localRoots` peer (the fast
-source) plus a ledger peer snapshot (`peerSnapshotFile`) that seeds independent
+source) plus a ledger peer snapshot (`peerSnapshotFile`) that seeds
 `PeerSourceP2PLedger` corroborators before outbound startup (see Peer
-Governance). Set `corroborationPeers` to the number of independent snapshot
-peers that must agree. See `GENESIS_SYNC.md` for the operator runbook. This differs from normal Praos sync (which trusts the
-longest valid chain from any peer) and from Mithril bootstrap (which trusts a
-signed snapshot at a trust boundary): Genesis trusts *density corroborated by
-independent peers* from origin.
+Governance). Set `corroborationPeers` to the number of *independent* snapshot
+peers that must agree — where independence must be established by the operator
+(distinct operators/infrastructure/chain views), since the selector can only
+enforce distinct remote hosts. See `GENESIS_SYNC.md` for the operator runbook.
+This differs from normal Praos sync (which trusts the longest valid chain from
+any peer) and from Mithril bootstrap (which trusts a signed snapshot at a trust
+boundary): Genesis trusts *density corroborated by peers the operator has reason
+to believe are independent* from origin.
 
 **Observability**: `GenesisStatus()` exposes the current mode, window, selected
 fast source, and per-peer density/corroboration. The selector emits
 `chainselection.genesis_corroboration_failed` when the densest fast source is
-denied for lack of corroboration (deduped per source) and
-`chainselection.genesis_mode_exited` with the exit reason (local slot, best
-known slot, window) when it returns to Praos.
+denied for lack of corroboration (deduped per source; the warning is also logged
+without an EventBus) and `chainselection.genesis_mode_exited` with the exit
+reason (local slot, best known slot, window) when it returns to Praos. A change
+to any tracked peer's frontier re-runs selection while corroboration is active,
+so corroboration granted or revoked takes effect immediately rather than on the
+next periodic tick.
 
-**Deferred / not implemented** (explicit scope boundary): the gate is a
-corroboration check, not the reference implementation's full Ouroboros Genesis
-density-at-intersection with **ChainSync Jumping** and devoted BlockFetch
-dynamics. It compares recent shared frontier points within the window; it does
-not, on its own, resolve a fork whose intersection is inside the window by
-counting blocks after the exact intersection, nor does it schedule block
-downloads across jumped peers. Those performance/robustness pieces (and wiring
-peer-governance demotion to the corroboration-failure event) remain future
-work; the density comparison plus the corroboration gate provide the from-origin
-*security* property that a fast source cannot steer the node without independent
-corroboration.
+**Deferred / not implemented** (explicit scope boundary — some of these are
+security limitations, not only performance): the gate is a corroboration check,
+not the reference implementation's full Ouroboros Genesis density-at-intersection
+with **ChainSync Jumping** and devoted BlockFetch dynamics. It confirms that a
+witness's chain is a prefix of the candidate's within the overlapping window; it
+does **not** resolve a fork whose intersection lies *inside* the window by
+counting blocks after the exact intersection, and it cannot testify about blocks
+a fast source produced *beyond* every witness's frontier. Consequently a fast
+source that stays consistent with honest peers up to their frontiers but forks
+only in the not-yet-witnessed suffix is corroborated until a witness advances
+past the fork — full density-at-intersection would decide such cases sooner, so
+its absence is a **security** limitation for that window, mitigated (not closed)
+by the fail-closed overlap requirement and the per-header density comparison.
+Wiring peer-governance demotion to the corroboration-failure event is likewise
+deferred. These remain future work; the density comparison plus the corroboration
+gate provide the from-origin property that a fast source cannot steer the node
+onto a chain no independent witness shares.
 
 #### Anti-flap incumbent pin
 
@@ -1622,13 +1653,17 @@ later ledger peer refreshes still query the live ledger/database provider.
 If the snapshot produces no usable peers, startup falls back to topology
 bootstrap peers.
 
-These snapshot-seeded ledger peers are the independent corroborators for the
+These snapshot-seeded ledger peers are the configured corroborators for the
 Genesis corroboration gate (see Chain Selection → Ouroboros Genesis trust
 model). In a GSA-style deployment the fast source is a trustable `localRoots`
 peer and the snapshot supplies the corroborating ledger peers; setting
-`genesisBootstrap.corroborationPeers` to the number of independent snapshot
-peers that must agree makes the node stall rather than follow the fast source
-until that many corroborators confirm its recent blocks.
+`genesisBootstrap.corroborationPeers` to the number of snapshot peers that must
+agree makes the node stall rather than follow the fast source until that many
+distinct-host corroborators confirm its recent blocks. Their *independence*
+(distinct operators, infrastructure, and chain views) is not something the
+selector can verify — it enforces only distinct remote hosts — so the operator
+is responsible for populating the snapshot with genuinely independent large
+ledger peers.
 
 Live ledger peer discovery is adapted at the node composition boundary:
 `ledger/` exposes stake pool relay data and current slot through neutral
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1338,11 +1338,22 @@ without an EventBus) and `chainselection.genesis_mode_exited` with the exit
 reason (local slot, best known *advertised* slot, window) when it returns to
 Praos — exit keys off the advertised network tip, not the observed frontier, so
 the gate stays active through from-origin sync until the local tip nears the
-real network tip. The exit horizon is the highest advertised tip among all
-live/eligible/non-stale peers, **including an uncorroborated far-ahead source**,
-so a lower corroborated peer cannot trigger a premature exit that would let the
-uncorroborated source steer the chain under Praos (the implausible-tip check
-bounds how far ahead any peer can advertise). A change
+real network tip.
+
+The exit horizon is the highest advertised tip slot among **corroborated
+(selectable)** peers. It must be corroborated, not any live peer's raw tip,
+because the advertised tip is untrusted and the implausible-tip check bounds only
+the advertised *block number*, not the *slot* (and the first peer is accepted
+with no reference): a single peer advertising a plausible block with a slot near
+`math.MaxUint64` would otherwise pin the node in Genesis mode forever — a
+liveness DoS. Requiring corroboration means `MinCorroboratingPeers` independent
+peers must have delivered matching headers, so a lone liar cannot inflate the
+horizon. The trade-off is that an uncorroborated source that is ahead in block
+number could win Praos selection once the local tip has caught up to the
+corroborated tip and the node has exited; closing that residual needs
+density-at-intersection (deferred, below). Liveness is prioritized over that
+residual because an unbounded advertised slot is a trivially exploitable
+single-peer stall. A change
 to any tracked peer's frontier re-runs selection while corroboration is active,
 so corroboration granted or revoked takes effect immediately rather than on the
 next periodic tick.
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -740,6 +740,8 @@ All event types follow the `subsystem.snake_case_name` convention.
 | `chainselection.chain_switch` | ChainSelector | Active peer changed |
 | `chainselection.selection` | ChainSelector | Chain selection made |
 | `chainselection.peer_evicted` | ChainSelector | Peer evicted |
+| `chainselection.genesis_corroboration_failed` | ChainSelector | Densest Genesis fast source lacked corroboration and was denied selection |
+| `chainselection.genesis_mode_exited` | ChainSelector | Left Genesis mode for Praos after catching up to the best known tip |
 | `chainsync.client_added` | ChainsyncState | Client tracking added |
 | `chainsync.client_removed` | ChainsyncState | Client tracking removed |
 | `chainsync.client_synced` | ChainsyncState | Client caught up |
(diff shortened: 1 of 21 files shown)


chrisguiney:
Thanks for the reviews. Pushed cebb32ec addressing the findings.

### cubic

- **P0 — shared-ancestor-then-diverge bypass** (`peer_tip.go`): fixed. `sharesObservedPoint` (any common point) is replaced by `confirmsRecentChain`: a witness confirms a candidate only when *every* block the witness observed within the candidate's window matches the candidate's `(slot, hash)`, and they share at least one. Agreeing on one old ancestor and then forking no longer corroborates. The frontier is populated per header during chainsync (dense), so same-chain peers agree on every block in their overlap. Overlap is required, so it fails closed. Regression test: `TestConfirmsRecentChainRejectsSharedAncestorThenDiverge`.
- **P1 — Sybil self-corroboration** (`genesis_corroboration.go`): witnesses are now counted by distinct remote **host**, and a witness on the candidate's own host is excluded, so several connections from one operator can't self-corroborate. Documented as a lower bound — true independence (operators/ASNs/chain views) depends on operator topology. Test: `TestGenesisCorroborationDedupsBySybilHost`.
- **P2 — negative config silently disables the gate** (`validate.go`): a negative `genesisBootstrap.corroborationPeers` is now rejected in `Validate` (fail closed). Test added.
- **P3 — no warning/event without an EventBus** (`genesis_corroboration.go`): the warning log + dedup now run regardless of EventBus; only publishing is gated on the bus.
- **peer_tip P2 — hashes retained in Praos** : the per-peer hash frontier is now tracked only while Genesis corroboration is active and cleared on Genesis exit; Praos operation keeps no extra per-peer state. Test: `TestRecordObservedPointSkipsHashesWhenInactive`.
- **selector P1 — active source keeps steering after corroboration revoked**: a frontier change or witness removal now forces re-selection while corroboration is active. Test: `TestGenesisCorroborationRevokedOnWitnessRemoval`.
- **selector P3 — `SelectBestChain` can exit Genesis without publishing the exit event**: the staged `genesis_mode_exited` event is now drained/published from `SelectBestChain`, `RemovePeer`, and `cleanupStalePeers` too.

### CodeRabbit

- **Shared `live && eligible && non-stale` helper**: extracted `peerLiveEligibleNonStaleLocked`, now the single source used by both `isPeerSelectableLocked` and the corroboration-witness check.
- **Docs overclaiming independence / divergence prevention**: `ARCHITECTURE.md` and `GENESIS_SYNC.md` now (1) describe snapshot peers as *configured* corroborators whose independence is an operator responsibility (the selector enforces only distinct hosts), (2) note raising the count doesn't raise independence, and (3) record density-at-intersection / in-window fork resolution as a residual **security** limitation (a fork only in the not-yet-witnessed suffix is followed until a corroborator advances past it), not merely a performance gap.

Still explicitly deferred (documented): ChainSync Jumping / devoted BlockFetch, full density-at-intersection, and wiring peer-governance demotion to the corroboration-failure event.

chrisguiney:
Thanks — went through both.

**Negative-config P2 (genesis_corroboration.go:94, review on 9f9895dd):** stale — that review predates the fix commit. `Validate` now rejects a negative `genesisBootstrap.corroborationPeers` (cebb32ec, `internal/config/validate.go`, with a test), and cubic's own re-review on cebb32ec no longer flags it.

**Test P2 (genesis_corroboration_test.go:351, review on cebb32ec):** false positive — the test is deterministic, not passing by luck. `RemovePeer` *does* call `EvaluateAndSwitch()` for the witness-removal case: it sets `reevaluate = genesisCorroborationActiveLocked() && (bestPeerConn == nil || *bestPeerConn != connId)` inside the lock, then after releasing the lock runs `if reevaluate { cs.EvaluateAndSwitch() }` (selector.go:680-682). So removing the witness (which is *not* the best) triggers an internal re-evaluation that finds `fast` uncorroborated and drops it to nil before `GetBestPeer()` is read. The linked analysis trailed off at "let me verify" without reaching that call.

That internal re-evaluation is exactly the behavior the test is asserting, so I deliberately did **not** add an explicit `EvaluateAndSwitch()` in the test — doing so would mask a regression if `RemovePeer` ever stopped self-triggering. Verified deterministic: 200× under `-race`. Pushed a9d67af9 adding a comment to make that intent explicit.

chainselection/selector.go
```diff
@@ -383,16 +419,18 @@ func (cs *ChainSelector) updatePeerTipObservedPraosView(
 			}
 		}
 
+		trackHashes := cs.genesisCorroborationActiveLocked()
```
wolf31o2, line 473: Genesis mode can exit before this hash history is meaningfully populated. In real ChainSync, `tip` is the far-ahead advertised tip while `observedTip` is each delivered header. Once two independent peers deliver the same first low-slot header, they corroborate each other; `advanceSelectionModeLocked()` uses `SelectionTip()` (the observed header), sees local slot 0 plus the Genesis window reaching that header, and switches to Praos. This makes `trackHashes` false for the remainder of the from-origin sync. I reproduced it with advertised slot 100,000 and observed slot 1 on two peers: the mode was `praos` after the second update. Please base the Genesis exit decision on the best corroborated advertised tip and add a regression test covering this event shape.

wolf31o2 (changes requested):
Requesting changes because the corroboration gate disables itself after the first corroborated early header in a realistic from-origin ChainSync flow. I reproduced this with two independent peers advertising slot 100,000 and both delivering the same slot-1 header: the selector transitioned to Praos after the second update. The full test suite passes, so a regression test covering separate advertised and observed tips is also needed.

chrisguiney:
Good catch — confirmed and fixed in f0c98ab0.

**Root cause:** `bestKnownGenesisSlotLocked` (the Genesis-exit "network tip" measure) used `SelectionTip()`, which prefers the delivered `ObservedTip`. In a real from-origin ChainSync a peer advertises a far tip (slot 100000) but delivers early headers (slot 1) first, so once two peers corroborated the slot-1 header the exit check `localTip(0) + window >= observedSlot(1)` fired and the node dropped to Praos — disabling the corroboration gate almost immediately, exactly as you described.

**Fix:** `bestKnownGenesisSlotLocked` now uses the advertised tip (`pt.Tip`), so exit keys off the real network tip. The node stays in Genesis mode (gate active) until the local tip is within the window of the advertised tip. Density and corroboration still use the observed frontier; only the exit criterion changed.

**Regression test:** `TestGenesisDoesNotExitOnEarlyObservedHeaders` — two independent peers advertise slot 100000 while both delivering the slot-1 header (separate `Tip` vs `ObservedTip` via `updatePeerTipObserved`). It asserts the node stays in Genesis after both updates, stays in Genesis as the local tip advances to slot 50000, and only exits once the local tip reaches slot 99990 (within the window of 30). Verified it **fails before** the fix (transitions to Praos after the second update) and passes after. Docs (`ARCHITECTURE.md`, `GENESIS_SYNC.md`) updated to say exit is based on the advertised network tip.

Full suite + conformance pass with `-race`.

chrisguiney:
Confirmed and fixed in 967366ad.

**Valid P1 (selector.go:253).** The Genesis-exit horizon was built from *selectable* peers (`isPeerSelectableLocked`), which excludes an uncorroborated source. I reproduced the bypass: a divergent source advertising a high slot (100000) but a block number (205) close to the honest peers — a low-density chain, so the block-based behind-best filter doesn't exclude it — was left out of the horizon. A lower corroborated peer (slot 200) then triggered a premature exit, and once in Praos the disabled gate let the uncorroborated source win by longest chain. The failing regression test showed the node both exited to Praos *and* then selected the bad source.

**Fix:** `bestKnownGenesisSlotLocked` now iterates live/eligible/non-stale peers (`peerLiveEligibleNonStaleLocked`) instead of selectable ones, so any advertised far source keeps Genesis mode active until the local tip actually catches up. The implausible-tip check bounds how far ahead a peer can advertise, so a liar cannot pin the node in Genesis beyond the real network tip + K (well within the window) — verified the normal exit tests still pass.

Regression test: `TestGenesisExitHorizonIncludesUncorroboratedSource` (fails before the fix on both counts above). Full suite + conformance pass with `-race`; `ARCHITECTURE.md` updated with the horizon semantics.

(The other still-showing comment, the line-351 test finding, is a stale re-run of an earlier commit — addressed and explained above in this thread; `RemovePeer` does re-evaluate internally, verified deterministic 200× under `-race`.)

chainselection/selector.go
```diff
@@ -216,15 +234,34 @@ func (cs *ChainSelector) genesisWindowSlotsLocked() uint64 {
...
 	var best uint64
 	for connId, pt := range cs.peerTips {
-		if !cs.isPeerSelectableLocked(connId, pt, false) {
+		if !cs.peerLiveEligibleNonStaleLocked(connId, pt) {
 			continue
 		}
-		tip := pt.SelectionTip()
-		if tip.Point.Slot > best {
-			best = tip.Point.Slot
+		if pt.Tip.Point.Slot > best {
```
wolf31o2: **Merge blocker:** this uses every live peer’s raw advertised slot, including uncorroborated peers, as the Genesis exit horizon. The “implausible tip” protection does not bound this value: `updatePeerTipObservedPraosView` compares only `tip.BlockNumber`, while `tip.Point.Slot` can be arbitrarily large (and the first peer is accepted without a reference). A peer can advertise a block number within K but a slot near `math.MaxUint64`, pass validation, and prevent Genesis→Praos transition indefinitely. Please either establish a defensible slot bound before including an uncorroborated tip here, or preserve the corroboration quarantine across transition without allowing raw uncorroborated slots to inflate the horizon. Add a regression test with a plausible block number and an extreme advertised slot.

ARCHITECTURE.md
```diff
@@ -1256,6 +1259,112 @@ as a fallback ingress source, but peer governance lowers their priority to zero.
...
+denied for lack of corroboration (deduped per source; the warning is also logged
+without an EventBus) and `chainselection.genesis_mode_exited` with the exit
+reason (local slot, best known *advertised* slot, window) when it returns to
+Praos — exit keys off the advertised network tip, not the observed frontier, so
+the gate stays active through from-origin sync until the local tip nears the
+real network tip. The exit horizon is the highest advertised tip among all
+live/eligible/non-stale peers, **including an uncorroborated far-ahead source**,
+so a lower corroborated peer cannot trigger a premature exit that would let the
+uncorroborated source steer the chain under Praos (the implausible-tip check
+bounds how far ahead any peer can advertise). A change
```
wolf31o2: This bound is not true in the current implementation. The plausibility check constrains advertised **block number**, not advertised **slot**, while the exit horizon above is slot-based. An uncorroborated peer can therefore advertise an extreme slot with a near-network block number and pin Genesis mode indefinitely. Please update this statement together with the implementation so the documented liveness guarantee matches the enforced bound.

wolf31o2 (changes requested):
Requesting changes for a merge-blocking Genesis liveness issue. The current exit horizon accepts an untrusted advertised slot that is not bounded by the existing block-number plausibility check. A single live peer can therefore keep Genesis mode active indefinitely. Please address the inline findings and run the required DevNet validation for this chain-selection change.

chrisguiney:
Fixed the merge-blocking liveness bug in c06aeb6e. You're right — the plausibility check bounds the advertised **block number**, not the **slot** (and the first peer has no reference), so an untrusted advertised slot could pin Genesis mode indefinitely.

**Fix (option B — corroborated horizon):** `bestKnownGenesisSlotLocked` now builds the exit horizon from **corroborated (selectable)** peers only, not every live peer's raw advertised tip. To raise the horizon a peer must be corroborated, i.e. `MinCorroboratingPeers` independent (distinct-host) peers must have delivered matching headers — so a lone liar advertising `{block: plausible, slot: ~MaxUint64}` cannot inflate it. This keeps your round-1 direction ("base exit on the best corroborated advertised tip") and reverts the earlier broadening to live/eligible peers.

**Regression test** (as requested — plausible block + extreme advertised slot): `TestGenesisExitNotPinnedByUncorroboratedExtremeSlot`. An uncorroborated peer advertises block 200 with slot `1<<62` alongside two corroborated honest peers at slot 200; local catches up to 200 and the node exits to Praos. I verified it **fails** against the previous broadened horizon (mode stays `genesis` — pinned) and passes with the fix.

**Trade-off, called out explicitly** (code comment + ARCHITECTURE.md): keying exit off corroborated peers means an uncorroborated source that is ahead in block number could, after exit, win Praos selection — the concern cubic raised. That residual needs density-at-intersection, which is deferred. I've prioritized liveness here because an unbounded advertised slot is a trivially exploitable single-peer stall, whereas the residual requires a divergent chain dense enough to be plausible yet uncorroborated (and would still face block validation on fetch). If you'd prefer a different balance, happy to adjust. The incorrect ARCHITECTURE.md claim about the implausible-tip check bounding the horizon is corrected.

**DevNet:** running `internal/test/devnet/run-tests.sh` now (from-genesis forging network against cardano-node, which exercises the Genesis→Praos transition path). Will follow up with the result.

Unit + `internal/config` + conformance pass with `-race`; lint/modernize clean.

chrisguiney:
Sharp catch — fixed in a72b9b08. You're right: corroboration validates a peer's **delivered** headers (the observed frontier), not its advertised tip, so a peer can deliver one shared early header (passing corroboration) and still advertise `slot ~MaxUint64`, pinning Genesis. Requiring corroboration alone did not bound the horizon.

**Fix — bind the horizon to delivered data.** `bestKnownGenesisSlotLocked` now counts a corroborated peer's advertised tip only once the peer has actually **delivered headers up to within the window of that tip** (`ObservedTip + window >= Tip`). A liar can't deliver up to a `MaxUint64` slot, and an honest peer early in from-origin sync hasn't yet delivered up to its far advertised tip — so neither raises the horizon. The advertised tip becomes the exit target only when a corroborated peer has served its chain up to it, which happens exactly when the local tip has caught up.

This finally satisfies **both** earlier constraints on this line at once:
- No single-peer liveness pin (an unbounded advertised slot never counts until delivered) — wolf31o2's merge-blocker and your point here.
- No premature exit at the start of sync (delivered headers reach the far tip only once caught up) — wolf31o2's round-1 concern.

**Tests:** `TestGenesisExitNotPinnedByCorroboratedLiarAdvertisingFarTip` (corroborated peer, shared slot-100 header, advertises `1<<62`, undelivered → must not pin) — verified it fails without the delivered bound. Also made `TestGenesisDoesNotExitOnEarlyObservedHeaders` realistic (delivered frontier advances with the local tip). ARCHITECTURE.md updated.

The residual you raised in an earlier round (an uncorroborated source ahead in *block* number winning Praos after exit) still stands as the documented deferred item needing density-at-intersection — orthogonal to this liveness fix.

---

**DevNet (per @wolf31o2's request):** `internal/test/devnet/run-tests.sh` passed end-to-end against this branch — all 10 scenarios, including `TestBasicBlockForging`, `TestChainGrowthRate`, `TestSustainedConsensus`, and `TestEpochBoundaryConsensus` (Dingo stayed in consensus with cardano-node across an epoch boundary; dingo slot 699/block 217 vs cardano slot 700/block 216). txpump: 3465 accepted, 30 rejected, 0 errors. Unit + config + conformance pass with `-race`; lint/modernize clean.

chainselection/selector.go
```diff
@@ -1171,16 +1286,21 @@ func (cs *ChainSelector) evaluateBestPeerLocked() (
...
 	var switchEvent *event.Event
 	var selectionEvent *event.Event
 	switchOccurred := false
+	// Compute Genesis corroboration status once per evaluation, independent of
+	// which peer (if any) is ultimately selected, so a stalled selection still
+	// reports why the densest fast source was denied.
+	corroborationEvent := cs.genesisCorroborationFailureLocked()
 
 	newBest := cs.selectBestChainLocked()
 	if newBest == nil {
```
wolf31o2, line 1357: P1: Clearing only bestPeerConn does not stop the active ChainSync/blockfetch driver. chainsync.State may already have this peer in activeClientConnId, and node wiring only reacts to ChainSwitchEvent with a non-nil replacement. Thus a lone uncorroborated source, or a previously selected source that loses its witness, can remain the active pipeline source even though GetBestPeer returns nil. Emit an explicit selected-to-none transition and have node composition quiesce/clear the active driver; add an integration test that verifies no post-revocation ingress or blockfetch continues.

config.go
```diff
@@ -875,6 +876,19 @@ func WithGenesisWindowSlots(slots uint64) ConfigOptionFunc {
...
+// WithGenesisCorroborationPeers sets the number of independent peers that must
+// report the same recent blocks before a fast (shallow) block source may drive
+// Genesis-mode chain selection. This is the Ouroboros Genesis trust control for
+// biased fast-sync sources (e.g. the Genesis Sync Accelerator): an
+// uncorroborated or divergent fast source is denied selection and stalls rather
+// than steering the local chain. A zero value disables corroboration
+// (density-only Genesis selection).
+func WithGenesisCorroborationPeers(peers int) ConfigOptionFunc {
+	return func(c *Config) {
+		c.genesisCorroborationPeers = peers
```
wolf31o2, line 892: P2: Negative values through the public programmatic API silently disable corroboration. Internal YAML/env/CLI validation rejects negatives, but NewConfig(WithGenesisCorroborationPeers(-1)) reaches ChainSelectorConfig unchanged, where the greater-than-zero activation check treats it as disabled. Since only zero is documented to disable this security gate, reject the value in root config validation or otherwise fail closed, and cover this API path.

wolf31o2 (changes requested):
Requesting changes for a merge-blocking integration gap: the selector can transition to no corroborated peer without clearing the already-active ChainSync/blockfetch driver, so the source is not actually stalled. I also found a fail-open negative-value path in the public programmatic configuration. Please add pipeline-level regression coverage for initial denial and corroboration revocation.

chrisguiney:
Both findings fixed in 92570f08. DevNet passes.

## P1 (merge-blocker) — the stall wasn't real; now enforced at ledger application

You were right that the source wasn't actually stalled, and it turned out to be deeper than the active driver. Mapping the ingress path showed that **header application is gated on peer-governance eligibility, independent of the selected best peer**: under the default primary strategy any ingress-eligible peer's headers are applied and blockfetched regardless of `bestPeerConn`. So denying the "best peer" slot — or clearing the active chainsync driver (which auto-repromotes and doesn't gate ingress anyway) — would not stop an uncorroborated source from feeding the ledger.

Fix — gate ledger **application**, not selection:
- `ChainSelector.ShouldApplyIngress(connId)`: while Genesis corroboration is active, only corroborated peers are apply-eligible; unknown/uncorroborated peers are denied (fail closed).
- New `OuroborosConfig.ChainsyncApplyEligible`, a second gate after ingress eligibility. The roll-forward/roll-backward handlers publish the **observation** events (`PeerTipUpdateEvent`/`PeerRollbackEvent`) first, then **withhold** the `ledger.ChainsyncEvent` (and thus blockfetch) when apply is denied. Observing before the gate is deliberate — a peer must be observed to become corroborated, so this avoids a deadlock.
- Explicit selected-to-none transition: `chainselection.selected_none` (which `ChainSwitchEvent` can't express), logged by the node.

Tests: `ShouldApplyIngress` gating (initial denial + post-revocation), `selected_none` on revocation, and an ouroboros-layer test proving the apply gate withholds the ledger event while still observing the tip.

## P2 — fail closed on negative threshold via the programmatic API

`NewChainSelector` now clamps a negative `MinCorroboratingPeers` to 1 with a warning (only 0 disables the gate), covering `NewConfig(WithGenesisCorroborationPeers(-1))`. Tested in `chainselection` and the root API path.

## DevNet

`run-tests.sh` **passed** — all 10 scenarios including `TestChainGrowthRate`, `TestRelayPropagation`, `TestSustainedConsensus`, and `TestEpochBoundaryConsensus` (dingo/cardano in consensus across the epoch boundary). txpump 3460 accepted / 0 errors. Two earlier runs failed on block-rate/deadline thresholds under machine load (15-min load avg 3.81 on a 16GB box running three Cardano nodes); those were timing flakiness, not regressions — the gate is a no-op with corroboration disabled (config `CorroborationPeers:0`, zero "withheld" log lines), and a clean re-run at low load passed with the same margins as before.

All `-race` unit suites + conformance + lint clean. ARCHITECTURE.md and GENESIS_SYNC.md document that enforcement is at ledger application.

chrisguiney:
Both fixed in 5e4084a4.

**P1 — apply gate could read stale corroboration state.** Right: the gate read chain-selection state updated asynchronously via `PeerTipUpdateEvent`, so its decision could predate the header being admitted. Added `OuroborosConfig.ChainsyncObservePeerTip`: while Genesis corroboration is active, the node updates chain selection (and peergov) **synchronously** in the roll-forward handler, before the apply gate, and skips the async `PeerTipUpdateEvent` to avoid a double update. So the apply decision now reflects the header currently being admitted. When corroboration is inactive it falls back to the async path unchanged (normal high-throughput sync keeps its parallelism).

New test `TestChainsyncClientRollForwardSyncObservationOrdersApplyGate` proves a corroborating header is applied in the same roll-forward — and I verified it **fails** if the observation is made asynchronous (the header is withheld because the gate reads stale state), so it genuinely guards the ordering.

**P3 — selected-to-none not emitted on disconnect/expiry.** Fixed: `RemovePeer` and stale-peer cleanup now stage and publish `ChainSelectedNoneEvent` when their replacement selection is nil, via a shared `stageSelectedNoneLocked` helper. New test `TestChainSelectedNoneEventOnBestPeerRemoval` covers the direct removal path.

All `-race` unit suites + conformance + lint (0 issues) pass. Both changes are inert when corroboration is disabled (the observe hook falls back to async; the event is observability only), so they don't alter normal/DevNet sync behavior.

chrisguiney:
DevNet on 5e4084a4: **all 10 scenarios passed** (`TestChainGrowthRate`, `TestRelayPropagation`, `TestSustainedConsensus`, `TestEpochBoundaryConsensus` — dingo/cardano in consensus across the epoch boundary; txpump 3463 accepted, 0 errors). Confirms the P1/P3 changes don't regress sync (they're gated off with corroboration disabled, DevNet's config).

chrisguiney:
Fixed in eeadbb69. Valid and subtle — thanks.

A header first seen from an uncorroborated (apply-denied) peer was withheld by the apply gate, but `UpdateClientTip` had already recorded it in the cross-peer dedup cache (that step ran before the gate). So when a corroborated, apply-eligible peer later delivered the same point, it was treated as a duplicate and dropped — permanently under `HeaderSyncStrategyParallel` (which never replays duplicates) — losing the point.

**Fix:** reordered the roll-forward handler so tip observation and apply-eligibility are computed *before* the dedup step, and the cross-peer dedup entry is recorded **only for apply-eligible headers**. A withheld header now updates the cursor/tip via `UpdateClientTipWithoutDedup` (no dedup entry), so the first corroborated, apply-eligible peer still publishes the point. The final apply gate reuses the pre-computed eligibility.

When corroboration is disabled `applyEligible == ingressEligible`, so dedup behavior is unchanged for normal sync.

Regression test `TestChainsyncClientRollForward_WithheldHeaderNotPermanentlyDeduped` (parallel strategy: a point first seen from an uncorroborated peer is still published by a later corroborated peer) — verified it fails against the previous always-dedup behavior.

`-race` (chainselection, ouroboros, chainsync, root) + conformance + lint (0 issues) all pass.

wolf31o2 (approved):
Reviewed locally. Core corroboration logic, lock safety, confirmsRecentChain merge, observed-slot/point lockstep, exit-horizon overflow/liveness, and Sybil dedup all verified. Local -race passes on chainselection, ouroboros, internal/config; conformance suite passes; docs updated. Feature is default-off, so no behavior change for existing deployments.

One follow-up filed for the roll-backward apply-gate ordering race (the roll-forward path observes synchronously to close this exact race; roll-backward does not). Non-blocking given default-off, but should land before enabling corroborationPeers > 0. See linked issue.

wolf31o2:
Follow-up filed: #2928 (roll-backward apply-gate ordering race), assigned to @chrisguiney.

PR #2921, opened 2026-07-22
refactor: plugin system

wolf31o2:
## Summary by cubic
Refactors the plugin system to an app-owned `plugin.Host` with explicit provider registration across storage, mempool, and APIs. Config moves under `plugins.*`; the database receives injected stores, and APIs start only in API mode when their port is nonzero.

- **Refactors**
  - One in-process `plugin.Host`; no `init` self-registration, global options, or listing. Legacy `database/plugin/*` register/option/list APIs removed.
  - Built-ins register via `internal/plugins.Register(host)`; host owns provider lifecycle with ordered start/stop and failure unwind. Database drains only DB-owned workers and is constructed with `database.Stores{Blob, Metadata}`. New `internal/test/dbtest` helpers simplify test DB setup/teardown.
  - New API provider packages `api/utxorpc`, `api/blockfrost`, `api/mesh`; defaults: 9090/3000/8080; set port `0` to disable. Addresses use `net.JoinHostPort`. APIs resolve only in API storage mode.
  - Config/flags/env moved to `plugins.*` (storage, mempool, api). Env vars use `DINGO_PLUGINS_*`; legacy `DINGO_*_PORT` envs remain aliases. CLI uses provider selectors (e.g. `Plugins.Storage.{Blob,Metadata}.Provider`, `Plugins.Mempool.Provider`).
  - Storage: shared `databasePath` is injected; `plugins.storage.{blob,metadata}.config.dataDir` may override; unset paths inherit `databasePath`. Extra providers (`postgres`, `mysql`, `gcs`, `s3`) require `-tags dingo_extra_plugins`.
  - Badger: metrics register only when a Prometheus registry is injected; default cache/value sizes extracted; S3 `prefix` honored verbatim.

- **Migration**
  - YAML/env/flags now live under `plugins.storage.*`, `plugins.api.*`, `plugins.mempool.*` with per-capability `provider` and `config`. API defaults: 9090 (`utxorpc`), 3000 (`blockfrost`), 8080 (`mesh`); set port `0` to disable.
  - Programmatic use: create a `plugin.Host`, call `internal/plugins.Register(host)`, resolve providers, and inject stores into `database.New`. Do not use global registration or `SetPluginOption`. Mithril now accepts `StoragePlugins{Blob,Metadata}` for bootstraps.

<sup>Written for commit 78863a5966e12e9d9c26b44068cc9edaeb838540. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2921?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


arepala-uml (approved):

arepala-uml:
I have gone through the each file change and changes across each file looks good. I even asked claude as well to review where it pointed out a couple of test-coverage gaps and these seems to be non-blocking to merge the PR & nice to have.

- `plugin/plugin_test.go` never exercises `StopCapability` racing against Resolve under `-race`. The Host's own doc comment claims it's "safe for concurrent listing and resolution," and there's real logic in `plugin.go` built specifically to handle that race — but nothing in the test suite actually drives that race. Not a bug (I traced the logic myself and it looks correct), just untested.

- `node_plugin_selection_test.go` only tests config decoding/selection, not actual `Run()/Stop()` behavior — so the partial-startup-failure cleanup and the port-0-skip-at-runtime behavior (both of which I verified are correct by reading the code) are validated only by inspection, not by any test that would catch a future regression.

- `OpenDatabase` has a slightly fragile API contract — it can return a non-nil runtime alongside a non-nil error (for a recoverable DB error). Today's three callers handle this correctly, but only via careful non-idiomatic code; a future caller writing the natural `if err != nil { return err } `would leak the plugin host. Not a bug today, just a footgun.

PR #2955, opened 2026-07-26
feat(blockfrost): implement /addresses/{address} endpoint

### LINKED ISSUE #2935: blockfrost: implement base address information endpoint
## Summary

Implement the missing base address information endpoint:

- `GET /addresses/{address}`

Dingo already implements `/addresses/{address}/utxos` and `/addresses/{address}/transactions`, but the base route is not registered.

Source of truth: Blockfrost OpenAPI `0.1.90` (`openapi.json` in `blockfrost/openapi`), schema `address_content`.

## Response shape

Return:

- `address` — the Bech32 address
- `amount` — aggregate unspent value by unit, including `lovelace` and native assets
- `stake_address` — controlling Bech32 stake address, or `null` where the address has none
- `type` — `byron` or `shelley`
- `script` — whether the payment credential is a script

The amount must reflect current, non-spent UTxOs and use the same asset-unit encoding as the existing UTxO endpoints.

## Acceptance criteria

- `GET /api/v0/addresses/{address}` is registered without changing the existing sub-routes.
- Byron and Shelley addresses are parsed and classified correctly.
- Base, enterprise, pointer, reward/unsupported, key, and script credential cases return the appropriate response or Blockfrost-style error.
- Amounts aggregate lovelace and multi-assets across all current UTxOs without duplicate units.
- Addresses with no current UTxOs and unknown/invalid addresses follow the current Blockfrost compatibility behavior.
- Handler and adapter tests cover response shape, stake-address derivation/nullability, script detection, multi-asset aggregation, invalid input, and database failures.

## Related

- Existing address sub-routes: #1366
- UTxO response-field work: #2484

mmahut:
Closes #2935 

Adds the Blockfrost address summary endpoint: balances aggregated across
live UTxOs, stake_address, type, and script flag.

- balances are aggregated in SQL (new GetUtxoBalanceByAddress); the sqlite
  backend needs an INDEXED BY hint or the planner scans every live UTxO
- accepts CIP-5 payment credentials (addr_vkh1/script1) like Blockfrost,
  aggregating across all addresses sharing the credential
- rejects wrong-network addresses, stake addresses, and malformed input
- includes a fix for the Mithril snapshot import leaving payment_key empty
  for script-payment UTxOs, which made them invisible to address queries
  (separate commit)

Validated against the official blockfrost-tests preview suite: 15/15.

---
## Summary by cubic
Implements Blockfrost `/addresses/{address}` to return an address summary with SQL‑aggregated balances, with CIP‑5 credential support, strict network/format checks, and correct pointer‑address handling. Also fixes a snapshot import bug that hid script‑payment UTxOs.

- **New Features**
  - Added `GET /api/v0/addresses/{address}` in `api/blockfrost`: returns lovelace and per‑asset totals, `stake_address`, `type`, and `script`.
  - Accepts CIP‑5 payment credentials (`addr_vkh`, `script`); aggregates across all addresses sharing the credential and avoids false 404s via `CountTransactionsByPaymentCred`.
  - Validates network; rejects stake addresses and malformed/base58 inputs; derives script stake address for key/script base addresses; preserves pointer address identity via exact CBOR address matching.
  - Database: added `GetUtxoBalanceByAddress` (MySQL/Postgres/SQLite) to aggregate balances in SQL; SQLite uses an `INDEXED BY` hint for large chains.

- **Bug Fixes**
  - Snapshot import: store script payment hashes in `payment_key` for script‑payment UTxOs, making them visible to address and credential queries.

<sup>Written for commit 5b4db1daf5870b6bad85c16739b287645efa562d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a Blockfrost-compatible endpoint for retrieving address summaries.
  * Address responses now include balances, native assets, stake addresses, address type, and script indicators.
  * Supports bare payment credentials in supported address queries.
  * Improved handling of missing, malformed, and network-mismatched addresses.

* **Bug Fixes**
  * Corrected payment credential extraction for script-based addresses.
  * Improved detection of previously seen addresses with no live UTxOs.

### DIFF
--- a/api/blockfrost/adapter.go
+++ b/api/blockfrost/adapter.go
@@ -26,6 +26,7 @@ import (
 	"slices"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/blinklabs-io/dingo/database"
 	"github.com/blinklabs-io/dingo/database/models"
@@ -50,6 +51,7 @@ import (
 
 var (
 	ErrInvalidAddress      = errors.New("invalid address")
+	ErrAddressNotFound     = errors.New("address not found")
 	ErrInvalidBlockID      = errors.New("invalid block id")
 	ErrBlockNotFound       = errors.New("block not found")
 	ErrEpochNotFound       = errors.New("epoch not found")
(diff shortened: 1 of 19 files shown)


mmahut:
@coderabbitai review

api/blockfrost/adapter.go
```diff
@@ -2061,6 +2063,214 @@ func uintToInt(v uint) int {
...
+	}
+
+	// Read every query from one snapshot so a block committed
+	// mid-request cannot mix two chain states in the response.
+	db := a.ledgerState.Database()
+	txn := db.Transaction(false)
+	defer txn.Release()
+
+	balance, err := db.Metadata().
+		GetUtxoBalanceByAddress(addr, txn.Metadata())
```
wolf31o2: Blocking: `isPaymentCred` is discarded when the balance query is made. A bare credential is represented as a synthetic enterprise address, but `AppendUtxoAddressOrBranch` uses the same payment-only predicate for that synthetic address and for a real enterprise or pointer address: `payment_script` + `payment_key`, with no staking/address-form constraint. As a result, querying a full enterprise address includes base and pointer UTxOs that share its payment credential; pointer addresses also collapse together because their pointer payload is not represented. Please pass an explicit match mode and preserve full-address identity. At minimum enterprise matching must require no staking component, while pointer matching needs the pointer/address payload to be represented or otherwise matched exactly.

api/blockfrost/blockfrost_test.go
```diff
@@ -2594,6 +2602,159 @@ func TestDefaultListenAddress(t *testing.T) {
...
+	)
+	require.NoError(t, err)
+	assert.True(t, strings.HasPrefix(encoded, "stake_test17"))
+
+	stakeAddr, err := lcommon.NewAddress(encoded)
+	require.NoError(t, err)
+	assert.Equal(t, stakeScriptHash, stakeAddr.StakeKeyHash().Bytes())
+}
+
+func TestHandleAddress(t *testing.T) {
```
wolf31o2, line 2793: Please add adapter/storage regression coverage for the balance semantics rather than testing only a mocked handler. Seed enterprise, base, and pointer UTxOs with the same payment hash, then verify that each full address returns only its own balance while `addr_vkh`/`script` intentionally aggregates across address forms. The current mock bypasses `NodeAdapter.Address` and `GetUtxoBalanceByAddress`, which is why the address-conflation bug is not detected. The backend contract should be covered for SQLite and kept in parity for MySQL/Postgres.

wolf31o2 (changes requested):
The endpoint shape and SQL aggregation approach look reasonable, but the current address predicate conflates full enterprise/pointer addresses with bare payment credentials. This can return balances belonging to other addresses that share the payment hash, so I do not think this is safe to merge yet. Please preserve full-address matching semantics separately from payment-credential aggregation and add regression coverage for the distinction.

mmahut:
What has beendone:

- isPaymentCred is now an explicit match mode passed to the balance query
- exact mode requires an empty staking_key for payment-only addresses, so base-address UTxOs sharing the payment hash are excluded
- pointer addresses are matched exactly in the adapter by comparing decoded address bytes (the pointer payload is not represented in the utxo table)
- added sqlite storage regression test: enterprise/base/pointer UTxOs on one payment hash - each full address returns only its own balance, addr_vkh aggregates; MySQL/Postgres share the same predicate builder

BTW, the same conflation exists elsewhere - /addresses/{address}/utxos, the utxorpc address queries, and GetUtxosByAddress all use the payment-only predicate without a staking constraint, so enterprise queries there also pick up base-address UTxOs sharing the payment hash. (Interestingly /addresses/{address}/transactions already matches exactly, so the utxo and tx lists currently disagree for such addresses.). Does it seem right?

wolf31o2 (approved):

Issue #2936, opened 2026-07-24
blockfrost: implement retiring pool, pool detail, and pool metadata endpoints


wolf31o2:
## Summary

Implement the missing Blockfrost stake-pool endpoints:

- `GET /pools/retiring`
- `GET /pools/{pool_id}`
- `GET /pools/{pool_id}/metadata`

Dingo currently exposes `/pools` and `/pools/extended`, but not these routes.

Source of truth: Blockfrost OpenAPI `0.1.90` (`openapi.json` in `blockfrost/openapi`).

### Retiring pools

Return paginated `pool_id` and retirement `epoch` rows for pools with an effective retirement certificate. Support `count`, `page`, and `order`, with deterministic blockchain ordering.

### Pool detail

Accept Bech32 or hexadecimal pool IDs and return the OpenAPI `pool` shape, including:

- identifiers and VRF key
- total/current-epoch minted blocks
- live/active stake, size, saturation, delegator count, and live pledge
- declared pledge, margin, fixed cost, reward account, and owners
- registration and retirement transaction histories
- nullable latest valid `calidus_key`

Do not substitute placeholders for unavailable aggregates; implement the required persisted queries or explicitly return the schema-compatible value dictated by Blockfrost semantics.

### Pool metadata

Return the current registration metadata shape:

- `pool_id`, `hex`
- nullable `url` and `hash`
- nullable `ticker`, `name`, `description`, and `homepage`
- fetch/validation `error` when applicable

Return the schema-compatible empty object when no metadata is available. Reuse the persisted off-chain metadata and error state used by other pool responses rather than fetching synchronously in the request path.

## Acceptance criteria

- All three routes are registered under `/api/v0`, with route ordering that prevents `{pool_id}` from swallowing `/retiring` or `/extended`.
- Bech32 and hexadecimal pool IDs resolve to the same pool.
- Retirement results reflect active rollback-aware certificate state and paginate/order correctly.
- Pool detail matches the OpenAPI `pool` field set and computes epoch-sensitive values for the current epoch.
- Pool metadata represents successful, absent, and failed metadata fetches according to the OpenAPI schema.
- Invalid/missing pools use the existing Blockfrost error response conventions.
- Handler and adapter tests cover response fields, pagination/order, ID formats, missing data, rollback-visible state where relevant, and database failures.
- Assess `DATABASE.md` and `ARCHITECTURE.md` under the repository documentation requirements and update them if the implementation changes a documented query surface or cross-component flow.

[...]

PR #2969, opened 2026-07-27
feat(blockfrost): implement retiring pools and pool metadata endpoints

mmahut:
Adds two pool endpoints:

**GET /pools/retiring**

- pools whose latest retirement certificate targets a future epoch and was not cancelled by a later re-registration. Certificate recency compares (added_slot, block_index, cert_index), so a same-slot re-registration cancels correctly. Rows written by the Mithril ledger-state import (certificate_id = 0) keep their precedence within a slot.
- ordered by retirement epoch, then announcement position, same as hosted

**GET /pools/{pool_id}/metadata**

- accepts bech32 or hex pool ids, responds with the bech32 id
- on-chain url/hash from the latest registration, off-chain ticker/name/description/homepage from the metadata cache
- a pool without a metadata anchor returns {} like hosted
- failed fetches return the error object with the hosted codes and message formats: HTTP_RESPONSE_ERROR, HASH_MISMATCH, CONNECTION_ERROR, SIZE_EXCEEDED, and DECODE_ERROR for hash-valid but malformed JSON

New read-only store query GetRetiringPools (all three backends), documented in DATABASE.md. ARCHITECTURE.md router inventory updated (also adds the DRep list and address summary entries that were missing).

Validation:

- blockfrost-tests preview suite: 18/18, and the same 18 pass against hosted with a real project id, so the fixtures reflect current reality
- 50 random pools diffed byte-for-byte against hosted: everything deterministic matches, including error objects. The only differences are failure messages for pools with dead metadata URLs, where each fetcher reports its own network experience (timeouts, localhost refusal, link shorteners behaving differently per client).

Part of #2936

---
## Summary by cubic
Adds Blockfrost-compatible endpoints for retiring pools and pool metadata with hosted ordering, pagination, and error behavior. Also adds a cross-backend query and exact off-chain metadata error mapping.

- **New Features**
  - GET /pools/retiring: lists pools pending retirement (future epoch not canceled), ordered by epoch then announcement; paginated with standard Blockfrost headers.
  - GET /pools/{pool_id}/metadata: accepts bech32 or 56-char hex; responds with bech32 `pool_id` and `hex`; includes on-chain URL/hash and cached off-chain fields (ticker/name/description/homepage); returns {} if no anchor; returns an error object on fetch failure (HTTP_RESPONSE_ERROR, HASH_MISMATCH, SIZE_EXCEEDED, CONNECTION_ERROR, DECODE_ERROR); 400 on invalid id; 404 if pool not found.
  - New read-only `GetRetiringPools` across MySQL/Postgres/SQLite using certificate recency `(added_slot, synthetic-import precedence, block_index, cert_index)`; Mithril (`certificate_id = 0`) takes precedence within a slot. Updated `ARCHITECTURE.md`/`DATABASE.md` and added tests.

<sup>Written for commit 861ac0bc49c67b13103987299e58f019b9283b8b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2969?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added endpoints for viewing pools scheduled for retirement.
  * Added pool metadata lookup, including off-chain details and fetch status.
  * Added support for both hexadecimal and bech32 pool identifiers.
* **Documentation**
  * Updated API documentation to clarify supported endpoints.
  * Added database documentation for identifying retiring pools.
* **Bug Fixes**
  * Improved metadata error reporting for hash mismatches, oversized responses, HTTP errors, and connection failures.

### DIFF
--- a/api/blockfrost/adapter.go
+++ b/api/blockfrost/adapter.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"net/http"
 	"slices"
 	"sort"
 	"strconv"
@@ -58,6 +59,7 @@ var (
 	ErrAssetNotFound       = errors.New("asset not found")
 	ErrDRepNotFound        = errors.New("drep not found")
 	ErrInvalidTransaction  = errors.New("invalid transaction")
+	ErrInvalidPoolID       = errors.New("invalid pool id")
 	ErrMempoolUnavailable  = errors.New("mempool unavailable")
 	ErrMempoolFull         = errors.New("mempool full")
 	ErrTransactionNotFound = errors.New("transaction not found")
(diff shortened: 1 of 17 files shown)


api/blockfrost/adapter.go
```diff
@@ -1676,8 +1678,227 @@ func (a *NodeAdapter) liveStake() (uint64, error) {
...
+	ret.URL = &url
+	ret.Hash = &hash
+
+	doc, err := db.Metadata().GetOffchainMetadata(
+		models.OffchainMetadataSourcePool,
+		metadataURL,
+		metadataHash,
+		txn.Metadata(),
+	)
+	if err != nil || doc == nil || doc.Status == models.OffchainMetadataStatusPending {
```
wolf31o2: Blocking: `err` from `GetOffchainMetadata` is folded into the same successful path as a missing or pending cache row. A database read failure therefore returns `200 OK` with incomplete metadata and leaves no indication that the store query failed. Please return a wrapped error when `err != nil`, then handle only `doc == nil` or `pending` as a successful partial response; the existing handler will correctly translate the propagated failure to `500`.

api/blockfrost/blockfrost_test.go
```diff
@@ -995,6 +1014,181 @@ func TestAssetHoldersFromUtxosPreservesPointerAddress(t *testing.T) {
...
+
+	assert.Equal(t, http.StatusOK, w.Code)
+	var resp PoolMetadataResponse
+	require.NoError(t, json.NewDecoder(w.Body).Decode(&resp))
+	assert.Equal(t, mock.poolMetadata.PoolID, resp.PoolID)
+	require.NotNil(t, resp.URL)
+	assert.Equal(t, url, *resp.URL)
+	require.NotNil(t, resp.Ticker)
+	assert.Nil(t, resp.Name)
+}
```
wolf31o2, line 1088: Please add regression coverage at the `NodeAdapter.PoolMetadata` boundary for the blocking error path: make `GetOffchainMetadata` fail and assert that the error is propagated instead of producing a successful URL/hash-only result. The current handler test uses a prebuilt `PoolMetadataInfo`, so it bypasses the store lookup and cannot detect this regression.

wolf31o2 (changes requested):
The endpoint shape, retirement selection, documentation, and focused tests look good, and CI is now green. One error-handling issue remains blocking: an off-chain metadata store failure is currently returned as a successful partial response. Please address the two inline comments by propagating that failure and adding regression coverage at the adapter boundary.

api/blockfrost/adapter.go
mmahut: Ah, good catch!

wolf31o2 (approved):

PR #2972, opened 2026-07-27
fix: use reward_live_stake for mark snapshot

### LINKED ISSUE #2948: Ledger-apply hangs at epoch boundary — header chain advances, ledger frozen one slot before boundary, silent (no crash, no rollback-K)
## Environment
- network: preview
- role: block producer (Dingo), fresh Mithril bootstrap
- version: v0.67.0 (docker `ghcr.io/blinklabs-io/dingo:0.67.0`)
- fresh Mithril bootstrap completed 2026-07-24T21:00Z (via auto-applier, triggered by v0.67.0 release)
- container running unprivileged, `restart: unless-stopped`
- upstream chainsync peers include a co-located Haskell cardano-node relay (v10.5.x) which is at canonical tip throughout

## Summary
BP produced blocks and extended the chain normally for ~3 h after fresh Mithril bootstrap, right up to the epoch 1368→1369 boundary. It logged the boundary transition at 00:00:00Z (loaded the pre-computed leader schedule for 1369, fired snapshot handler, started governance tally). Then the ledger stopped applying blocks. It has now been ~1 h post-boundary and:
- Ledger tip is frozen at slot **118281598** (last successful chain-extend was 2 seconds before the boundary at slot 118281600)
- The chainsync client keeps receiving headers for slots >118281600 and logs plateau warnings:
  ```
  local tip plateau is a ledger-application backlog; header chain already caught up, not recycling chainsync
  applied_tip_slot=118281598  primary_chain_tip_slot=118282069  best_peer_tip_slot=118281786
  plateau_duration=4m
  ```
- Container is `Up 4 hours (healthy)`, `RestartCount=0`, no `panic`, no `ERROR` lines
- CPU ~2% (idle-ish, not a hot loop); memory 10 GB / 23 GB (normal)
- No `VRF leader value exceeds stake-derived threshold` (so #2892/#2843 held)
- No `rollback exceeds security parameter K` (so not the same class as #2941)
- No `reward stake input total mismatch` (so not #2925)

## Symptom timeline
```
2026-07-24T23:59:41Z  chain extended new tip slot=118281581  ← last extend
2026-07-24T23:59:58Z  chain extended new tip slot=118281598  ← last extend (final)
2026-07-25T00:00:00Z  epoch boundary reached via slot clock  slot=118281600 epoch=1369 ledger_epoch=1368
2026-07-25T00:00:00Z  handling epoch transition  previous_epoch=1368 new_epoch=1369 boundary_slot=118281600 snapshot_slot=118281599
2026-07-25T00:00:00Z  loaded persisted leader schedule  epoch=1369 leader_slots=10
2026-07-25T00:01:13Z  governance epoch tally starting  epoch=1369 active_proposals=26 active_dreps=8950 pool_snapshot_rows=632
2026-07-25T00:03:59Z  local tip plateau ... plateau_duration=4m applied_tip_slot=118281598 primary_chain_tip_slot=118281806
2026-07-25T00:07:59Z  local tip plateau ... plateau_duration=4m applied_tip_slot=118281598 primary_chain_tip_slot=118282069
... plateau messages continue every ~4m; NO new "chain extended" lines ...
```

Note the split between wall-clock epoch (1369, fires the snapshot handler and leader schedule reload) and ledger epoch (still 1368, since the last applied block is at slot 118281598 which is in epoch 1368). Chainsync sees canonical headers for the new epoch and blocks that go with them (`primary_chain_tip_slot` keeps advancing), but the ledger-application pipeline stopped applying them at the boundary.

### Metrics (present, just frozen for chain state)
`cardano_node_metrics_slotNum_int` — frozen at 118281598 for over 1 h
`cardano_node_metrics_epoch_int` — reports 1368 (ledger epoch, correct given tip)
`cardano_node_metrics_Forge_forged_int` = 3 (three forges before the boundary this run)
`cardano_node_metrics_Forge_about_to_lead_int` = 11746 (leader checker still ticking each slot)
`cardano_node_metrics_Forge_node_is_leader_int` = 3 (three leader-slot windows, matches forges)

### Container is healthy from the runtime perspective
`docker ps preview-bp` → `Up 4 hours (healthy)`, no restarts, CPU ~2 %, memory normal, no OOM.

### Chainsync IS working
Log continues to receive/plateau-log incoming headers past the boundary. The primary chain is advancing (`primary_chain_tip_slot=118282069` at 00:07Z, later 118283+, all higher than `applied_tip_slot`).

### Leader checker IS running (but always skipping)
Every few seconds, at the current wall-clock slot, this fires:
```
level=WARN msg="skipping leader eligibility check: Mithril-imported mark snapshot captured mid-epoch, not at the epoch boundary"
slot=118285373 epoch=1369 snapshot_epoch=1368 snapshot_type=mark captured_slot=118254536
```
`captured_slot=118254536` is ~27,000 slots before the boundary — the Mithril snapshot's mark, which is expected. Normally this warn stops once Dingo does its own live mark capture at the epoch boundary. It hasn't stopped because the ledger hasn't crossed the boundary.

### Independent chain view
The co-located Haskell cardano-node relay reached the boundary and continued extending normally. Koios agrees. Only Dingo is stuck.

## Reproducer
1. Fresh Mithril bootstrap Dingo v0.67.0 on preview a few hours before an epoch boundary.
2. Let Dingo produce or apply normally through the last few slots before boundary.
3. Cross the boundary at slot N*86400.
4. Observe: `handling epoch transition` fires from the snapshot manager (wall-clock triggered), `governance epoch tally starting` fires. Then `chain extended` messages stop and `local tip plateau is a ledger-application backlog` messages start with `applied_tip_slot` frozen at the last pre-boundary slot.

## What we did not verify
- We have not tried `docker restart preview-bp` yet — we're preserving state for this report first. Will attempt restart-only recovery after posting (a Dingo relay in a similar-looking wedge earlier this week recovered on plain restart; a BP in the #2941 rollback-K wedge did not).
- We have not decoded whether the specific incoming header at slot 118281600 has any unusual property that would block application (protocol-parameter change, first Conway-era rewards distribution at this boundary, etc.).
- We have not enabled Go pprof (`/debug/pprof/`) — the endpoint returns 404 on the metrics port (13798). If there is a pprof endpoint on a different port or via an env var, please point us to it and we can capture a goroutine dump before recovery.
- Governance load is nontrivial (`active_proposals=26 active_dreps=8950 pool_snapshot_rows=632`). We have not ruled out the governance epoch tally as the blocker vs. reward calculation vs. ledger snapshot write.

## Attachments (available on request)
- Full BP log covering the 15-minute window around boundary (~2 KB) — already captured locally
- `metadata.sqlite` post-hang (~1.6 GB)
- Full `docker logs preview-bp` since bootstrap (~90 MB)

## Related
- #2876 / #2892 — the VRF leader-eligibility class; fix landed and still holds (no VRF rejections in this run)
- #2941 — the "chainsync rollback exceeds security parameter K" class; #2940 likely fixed it (also no rollback-K in this run); this issue is a DIFFERENT class from either
- #2925 / #2927 — reward stake input total mismatch crash; not this (no crash)

wolf31o2:
Closes #2948

---
## Summary by cubic
Use the transactionally maintained `reward_live_stake` to build Mark and reward snapshots at the exact SNAP point. Fallback to historical reconstruction only when the persisted tip is beyond the boundary slot or when no tip exists, fixing #2948 by keeping slot-accurate totals and avoiding post-slot renewal skew.

- **Bug Fixes**
  - Added boundary-aware calculator that prefers the live aggregate when the transaction tip is at/before the snapshot slot; otherwise uses the historical path.
  - Implemented `GetLiveStakeInputsForPools` to read all registered credentials (keeps zero-stake for delegator counts; only positive stakes become `reward_stake_input`) with the live `account.expiration_epoch` filter.
  - Updated MySQL/Postgres/SQLite queries and tests to cover live SNAP capture, fallback when tip > slot or missing, and zero-stake handling; docs updated to reflect the unified live-source capture for Mark and rewards.

- **Migration**
  - Dropped the legacy `reward_live_stake` pool/total index and added `(pool_key_hash, credential_tag, staking_key)` for ordered boundary scans (`MigrateRewardLiveStakePoolIndex`); AutoMigrate ensures the new index exists.

<sup>Written for commit fb8e17824c1411fdee85ea435ccebacc2616dfe6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2972?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Epoch-boundary stake snapshots now use the maintained live stake data for accurate, transaction-consistent captures.
  * Snapshot calculations include zero-stake registered credentials when determining delegator counts.
  * Added account-expiration filtering for inactive reward accounts.

* **Bug Fixes**
  * Improved slot-accurate fallback handling when live snapshot data is unavailable.
  * Prevented zero-stake credentials from being included as reward inputs while preserving accurate counts.


chrisguiney (changes requested):
## Overview

This replaces the genesis-to-boundary historical reconstruction with a read of the transactionally maintained `reward_live_stake` aggregate for the **leader-election Mark snapshot** (previously the aggregate only backed reward inputs). The argument is that at the SNAP point no new-epoch block has been applied, so the live aggregate *is* the boundary state.

I traced `ledgerProcessBlocksFromSource` (`ledger/state.go:3663`): the batch is cut *before* applying the first new-epoch block, and the rollover runs in its own `SubmitAsyncDBTxn`. **The core premise holds for the block-driven rollover.** Read transactions are `ReadOnly + RepeatableRead` on MySQL/Postgres (`postgres/database.go:477`, `mysql/database.go:631`), so `GetTip` and the aggregate read share one snapshot — the tip check is not a TOCTOU.

Verification I ran locally on the PR head:

- `go test ./ledger/snapshot/... ./database/plugin/metadata/...` — pass
- `golangci-lint run` on changed packages — 0 issues; `modernize` — clean
- `ARCHITECTURE.md` and `DATABASE.md` both updated in-change, per the project change bar ✓

The approach is right. Requesting changes on a small, well-scoped list — a consensus-affecting change is shipping with an untested branch.

---

## Blocking

### 1. The CIP-0163 gate-on live path has no test coverage

Every new test passes `expiryEpoch = 0`. The change that actually carries risk — `calculator.go:95` discards `inactivityPeriod` and the live path instead filters on the live `account.expiration_epoch` column — is the exact read that a72b9b08 / c06aeb6e removed as unsafe. The "tip ≤ slot means no post-slot renewal can be visible" argument is sound for the authoritative hook, but nothing exercises it. Existing gate-on tests (`manager_test.go:201`) don't reach `calculateStakeDistributionInTxn`.

Please add a test with the gate on and an expired credential.

### 2. Mark totals are now path-dependent, and Mark is consensus-affecting

Before this PR the authoritative and fallback captures ran the *same* historical computation, so which path executed was consensus-neutral. They are now two different algorithms. A node that lands on the fallback (savepoint unavailable, hook error → `RollbackTo` + "deferring to fallback", hook not installed, or tip past slot) can persist different `pool_stake_snapshot` totals than a node that took the authoritative path — and those totals drive VRF leader-eligibility thresholds two epochs later.

Please add a differential test asserting both paths yield identical `PoolStakes` / `DelegatorCount` / `TotalStake` on one seeded fixture, with the gate both off and on.

### 3. DevNet validation

`CLAUDE.md` requires a DevNet run for consensus-affecting work, and this changes how leader-election stake is computed. Nothing in the PR indicates it was run.

---

## Should fix (cheap)

### 4. Resolve the ignored `inactivityPeriod` parameter

`ledger/snapshot/calculator.go:95` takes `_ uint64`, while `calculateBoundaryStakeDistributionInTxn` still threads `inactivityPeriod` into it. A caller passing a value into a callee that discards it will mislead the next reader. Either drop the parameter from the signature and stop passing it, or restore the name with a comment stating why it is unused.

### 5. Dead pruning loop

`ledger/snapshot/calculator.go:231`:

```go
for poolHash, delegators := range dist.DelegatorCount {
    if delegators == 0 { ... }
}
```

`DelegatorCount[poolHash]` is only ever produced by `++`, so it can never be 0 — the branch is unreachable. It reads as if it prunes empty pools, but pools with no rows never enter the maps in the first place. Suggest deleting it.

---

## Non-blocking, but please look before mainnet

### 6. `reward_live_stake` has no index on `pool_key_hash` — and a migration drops the one that existed

`database/models/reward_state.go:170` (`MigrateRewardLiveStakePoolIndex`) removes the legacy `(pool_key_hash, total_stake)` index, reasoning *"The aggregate has no pool-ordered query consumer yet."* This PR makes exactly that query the hot path at **every** epoch boundary, with `ORDER BY pool_key_hash, credential_tag, staking_key` on top. On SQLite (`sqliteBindVarLimit = 400`), mainnet's ~3.2k pools means roughly 8 full scans + 8 sorts of a ~1.5M-row table per boundary.

Still a large net win over the genesis-to-slot reconstruction, so this does not block. But given the PR exists to fix a boundary stall (#2948), it is worth measuring — and that migration comment is now factually wrong either way.

### 7. Four full copies of the delegator set on the boundary path

`stakeInputsForPools` builds `[]models.RewardLiveStake`, then `[]*models.RewardStakeInput` (pointer + two fresh byte slices per row), then `calculator.go:222` copies both slices *again* into `dist.StakeInputs`, then `rewardStakeDistribution` builds `deduped`. At mainnet scale that is ~1.3M rows × 4.

At minimum `dist.StakeInputs = make([]StakeInput, 0, len(inputs))`, and reuse `input.PoolKeyHash` / `input.StakingKey` directly — `stakeInputsForPools` already allocates them fresh and nothing else retains them. Relatedly, `calculateLiveStakeDistributionInTxn` validates pool-key length / credential length / tag per row, and `rewardStakeDistribution` re-validates the same three per positive row.

### 8. Stale comments left behind

- `StakeInputsForPools`' doc still claims *"the query and args are byte-identical to the pre-CIP query"* — the refactor moved the `total_stake` predicate into a conditional fragment, so that claim needs re-verifying or rewording.
- `getBatchPoolsDelegatedStake`'s doc comment and `RewardLiveStake.PoolDelegation*`'s *"when snapshot capture is wired"* (`models/reward_state.go:120`) are now stale.

### 9. Nits

- Postgres hardcodes `1000` where MySQL uses `rewardStakeInputPoolBatchSize`. It matches its own file's neighbours so it is not a regression, but a named constant would close the gap.
- `CaptureGenesisSnapshot` calls `calculateSnapshotDistribution(ctx, 0, …)`; `slot == 0` makes the live branch unreachable there whenever a tip exists. That is load-bearing and worth one line of comment.
- The two new tests duplicate ~40 lines of seeding; a shared helper would make the live-vs-historical contrast the visible part of each test.

---

## What's good

- The `hasTip` guard and its comment ("An all-zero tip is 'no persisted tip', not proof…") is exactly the right defensive call for the bootstrap/test case.
- Overflow checks on both per-pool and total accumulation, with distinct messages.
- Zero-stake handling matches the historical path, and the fallback test pins that parity rather than asserting it in prose.
- Docs updated in the same change rather than deferred.

chrisguiney (approved):

PR #2973, opened 2026-07-27
fix: replay-path recovery loop

### LINKED ISSUE #2941: Ledger silently rolls back and stalls after "chainsync rollback exceeds security parameter K" during live sync
## Environment
- network: preview
- roles: block producer (Dingo) + relay (Dingo) + independent Haskell cardano-node relay as upstream peer
- version: v0.66.2 (docker `ghcr.io/blinklabs-io/dingo:0.66.2`)
- storage: sqlite metadata, badger blob store
- BP fresh Mithril bootstrap on v0.66.2 at 2026-07-21T18:15Z
- container running as unprivileged user, `restart: unless-stopped`

## Summary
BP crossed the 1365→1366 epoch boundary cleanly (no VRF rejections — thanks for #2892) and produced 9 canonical blocks through epoch 1366 and into 1367. Then at 2026-07-23T00:50Z, mid-epoch 1367, the ledger reported a rollback exceeding security parameter K against the primary chainsync peer, tried to reconcile 4 times each with a deeper rollback, and after the last attempt went completely silent. Container never crashed, was never restarted, no VRF or ledger errors — the node just stopped advancing. It has now been stalled ~42 h at a slot BEHIND the last successful chain-extend.

## Symptom
Container is `Up 3 days (healthy)`. Metrics endpoint responds. Leader-schedule computer loop is running every second (`epoch nonce not yet available, skipping schedule` for the wall-clock epoch, using an active snapshot from an old epoch). No forges since the wedge. No VRF rejections all-time on this container. Metrics `cardano_node_metrics_slotNum_int` is frozen at the pre-rollback tip.

### Successful crossing + normal forging window
```
2026-07-22T00:00:05Z WARN skipping leader eligibility check (Mithril-imported mark) — expected
2026-07-22T01:27:16Z INFO block produced successfully slot=118027636 hash=c7e4af30…
2026-07-22T03:07:56Z INFO block produced successfully slot=118033676 hash=6b7050a8…
... 7 more forges ...
2026-07-22T23:43:08Z INFO block produced successfully slot=118107788 hash=bdf68071…
2026-07-23T00:49:47Z INFO chain extended, new tip: e0a41b9b… at slot 118092796  ← last extend BEFORE divergence
```

### The divergence and rollback loop (4 attempts, each deeper)
```
2026-07-23T00:50:02Z WARN component=ledger
  msg="primary chain and ledger diverged during live chainsync recovery, reconciling to common ancestor"
  reason="chainsync rollback exceeds security parameter K"
  chain_tip_slot=118071022  ledger_tip_slot=118096975
  chain_tip_hash=7c39611f…                ledger_tip_hash=7a6ed934…

2026-07-23T00:50:32Z WARN reason="chainsync rollback exceeds security parameter K"
  chain_tip_slot=118058060  ledger_tip_slot=118071022

2026-07-23T00:51:25Z WARN reason="chainsync rollback exceeds security parameter K"
  chain_tip_slot=118019130  ledger_tip_slot=118032101

2026-07-23T00:51:58Z WARN reason="chainsync rollback exceeds security parameter K"
  chain_tip_slot=118006252  ledger_tip_slot=118019227
```

Each attempt: `chain_tip_slot < ledger_tip_slot` (our ledger is ahead of the peer's tip that chainsync is trying to converge on), and each subsequent attempt walks the ledger further back. After the 4th attempt, no further chain-extends and no further errors — just silence for 42 h.

### After silence
The ledger-tip metric is frozen at slot **117987834**, which is EARLIER than every slot in the above rollback attempts. So the reconciliation walked the ledger tip back past the last successful chain-extend as well, and the node then never resumed extending.

Leader-schedule computer loop continues to spin every second:
```
level=INFO msg="computing leader schedule" component=leader epoch=1368
level=INFO msg="pool stake from active snapshot" component=leader epoch=1368 snapshot_epoch=1367 pool_stake=10451609189337
level=INFO msg="total active stake from active snapshot" component=leader epoch=1368 snapshot_epoch=1367 total_stake=3251047678447244
level=INFO msg="epoch nonce not yet available, skipping schedule" component=leader epoch=1368
```
So the ledger has active-snapshot data for epoch 1367 (i.e. it did roll to 1366→1367 at some point past the stalled slot before the divergence), but the chainsync tip is now frozen at a slot in epoch 1365. The two halves of the node are out of sync.

### Peer that triggered the divergence
The `connection_id` on all four rollback warnings identifies the upstream peer as our co-located Haskell cardano-node relay (`cardano-node v10.5.x`), which was and remains at canonical tip throughout. So the "primary chain" side is a full independent Haskell implementation that agrees with Koios; the Dingo BP is the side that diverged.

### Independent decision points
- The Dingo relay next to this BP was in a permission-denied crashloop for this whole window (unrelated local operator error on my side), so I can't corroborate against another Dingo. The Haskell relay was healthy and at tip.
- Container did NOT crash: `RestartCount=0 ExitCode=0`. Restart policy was `unless-stopped`.
- No `panic`, no `FATAL`, no `reward stake input total mismatch` (so this is not #2925).
- No `VRF leader value exceeds stake-derived threshold` (so #2892 held).

## Reproducer
1. Fresh Mithril bootstrap Dingo v0.66.2 on preview with a single upstream chainsync peer that is at canonical tip.
2. Cross an epoch boundary while running with the mid-epoch Mithril mark safety-valve active.
3. Continue forging into the next epoch normally.
4. Approximately 25 h after boundary crossing, expect "chainsync rollback exceeds security parameter K" against that upstream peer, followed by ledger tip walking backwards and no self-recovery. The node reports healthy from the container runtime's perspective.

## What we did not verify
- We did not decode which specific block hash Dingo's ledger contained at `ledger_tip_slot=118096975` — i.e. we did not confirm whether Dingo built its extra-K blocks on an orphan fork of its own or picked up a bad peer chain earlier. If useful I can grep for that hash in `block_nonce` / blob store.
- We did not test with `strict-utxo-validation` on.
- We only had one chainsync peer configured on the BP (the Haskell relay). We do not know whether adding more peers would have exposed a different divergence signature earlier.

## Attachments (available on request)
- Full `docker logs preview-bp` from bootstrap to now, ~90 MB
- `metadata.sqlite` post-wedge (~1.6 GB)
- Divergence-window log excerpt (~200 lines, already prepared locally)

## Related
- #2876 — the VRF-threshold wedge class that was fixed by #2892 / #2843; this issue is a DIFFERENT class of wedge that showed up after that fix landed
- #2925 / #2927 — a reward-stake crash on epoch rollover; different symptom (crash vs silent rollback) and different area, but same time window

### LINKED ISSUE #2971: ledger: replay-path recovery loops and silently commits value-not-conserved state when local chain spends an absent-producer UTxO
## Summary

Replay-path transaction-validation recovery (`security-param-fallback`) loops indefinitely and cannot self-heal when the local primary chain contains a block that spends a UTxO whose producing transaction is absent from the local chain. Worse, while looping the node silently commits value-not-conserved ledger state and creeps forward instead of halting. Observed on preview, from-genesis sync, at epoch 1371, oscillating slot ~118471000 to ~118480300, on commit fe88f558 (which includes #2940 and #2944).

## How it manifests

A contiguous cluster of UTxOs (epoch 1371, producing blocks around slots 118470000-118471100) is missing from the local ledger, but later blocks that are on the local chain (around 118480100+) spend them. Repeating log every ~60s:

```
WARN detected inconsistent local ledger state during replay, rewinding metadata state
     recovery_strategy=security-param-fallback failing_block_slot=118480109
     missing_input=02cd3499...#0 producer_block_slot=118470145 rollback_slot=118470143
WARN TX 107269... failed validation: conway utxo validation rule 12: script data hash mismatch: declared X, computed Y
```

The producer_block_slot in the log is not the real producer; it is the fallback anchor `failingBlock.ID - SecurityParam()` (the `ProducerTx==nil` default), i.e. the real producing transaction is unresolvable. The `script data hash mismatch` is a secondary symptom: with the input (and any reference script it carries) missing, the language-view set folded into the script-data hash differs, so the recomputed hash cannot match the declared one.

## Root cause

The persisted primary chain content is inconsistent: it retains a valid hash-linked block that spends `02cd3499...#0` while the block that produced `02cd3499` is absent. Block hash-linking only checks prevHash, so a chain can be a valid hash-chain while its UTxO accounting is impossible. This is the signature of a rollback / chain-switch that dropped producer block(s) but kept later descendant blocks. It was persisted by the pre-#2944 f51621a9 binary and inherited on resume by fe88f558.

Two code defects turn this into an unrecoverable, silently-corrupting loop:

1. Replay-path recovery is metadata-only. `tryRecoverFromTxValidationError` (non-at-tip branch, `ledger/replay_recovery.go:166-216`) calls only `ls.rollback(candidate.RollbackPoint)` (metadata rollback, `ledger/state.go:1938+`) then returns `errRestartLedgerPipeline`. It never calls `RewindPrimaryChainToPoint` and never emits a resync, so it replays the identical corrupt chain blob forever. The at-tip path `recoverAtTipFromTxValidationError` (`ledger/state.go:401-437`) does rewind the primary chain and publish a resync, which could re-fetch correct blocks from peers, but it is never reached because `IsAtTip()`/`reachedTip` (`state.go:2605`, set at `3928-3930`) never flips true while the node is stuck below the corrupt cluster. Log confirms: chain_tip_slot stays at the live tip while metadata rolls back to ~118470xxx ("skipping connection closure: chain already past rollback point").

2. Silent missing-input tolerance. In the consume path `database/plugin/metadata/sqlite/transaction.go:1596-1643`, when `totalAffected != len(consumedRefs)` each ref is re-checked and a genuinely absent UTxO logs `input UTxO not found` and `continue`s with no error (lines 1604-1616). So a block that spends a non-existent UTxO is applied rather than rejected whenever full validation is not run for it. Blocks below the moving validation cutoff (`shouldValidateBlock`, `state.go:3672-3710`, cutoff ~= liveTip - stabilityWindow 3k/f = 25920) are trust-applied, so as wall-clock advances the tip, each cluster block eventually falls below the cutoff, is trust-applied with the missing input tolerated, and the frontier advances one block. Net effect: perpetual rollback/replay churn pinned ~one stability window behind tip, committing value-not-conserved state instead of halting.

## Impact

- The node never reaches tip and cannot self-heal from a locally-inconsistent chain.
- It silently commits ledger state that violates value conservation (missing inputs tolerated), rather than halting loudly. This is a correctness/consensus hazard, not just a liveness one.

## Not caused by #2944

#2944 (5bd7bc23) is purely a perf change and is semantics-preserving: the consume WHERE clause is the same composite-key OR-list `(tx_id = ? AND output_idx = ?)`; it only chunks and pads each chunk to a power of two by repeating the chunk's last ref, which is idempotent for an OR of exact composite keys (see `database/plugin/metadata/internal/utxocond/utxocond.go:15-31,89-97`). It neither causes nor fixes this. #2940 (at-tip deep-rewind staircase) does not touch the replay-path recovery here — confirmed, since resuming on fe88f558 did not break the loop.

## Relationship to existing issues

- #2941 (open, rollback exceeds K, silent rollback/stall, reporter unsure whether Dingo built extra-K blocks on an orphan fork) is the closest root-cause relative — a chain that spends never-produced outputs is a plausible downstream manifestation of that chain-splice/divergence class.
- #2928 (open, roll-backward apply gate races async rollback observation) is a candidate mechanism for how a rollback could splice the chain inconsistently.
- #2940 / #2948 / #2943 / #2944 do not cover this.

## Reproduction

preview, from-genesis, core storage. Reaches epoch 1371 cleanly then enters the loop near tip. The trigger is ordinary rollback/chain-switch activity near tip, so a fresh re-sync is not guaranteed to avoid it. This should be treated as a from-genesis risk for preview/preprod/mainnet, not a one-off DB artifact.

## Suggested fixes (priority order)

1. On a genuinely unresolvable missing input in replay-path recovery, rewind the primary chain and resync from peers (mirror the at-tip path) instead of a metadata-only rollback that replays the same blob. `ledger/replay_recovery.go:166-216`.
2. Make trust-mode application of a block that consumes an absent UTxO a hard, loud failure (or a chain-rewind + resync trigger) rather than warn-and-continue, so impossible ledger state can never be silently committed. `database/plugin/metadata/sqlite/transaction.go:1596-1643` (and the postgres/mysql equivalents).
3. Investigate the chain-splice origin (under #2941 / #2928) so producer blocks are not dropped while their descendants are retained.

Key files: `ledger/replay_recovery.go` (166-216, 705-794, 826-874), `ledger/state.go` (401-437, 1938+, 2605, 3672-3710, 3928-3930), `database/plugin/metadata/sqlite/transaction.go` (1596-1643).

wolf31o2:
Closes #2971 
Closes #2941 

---
## Summary by cubic
Fixes the replay recovery loop by rewinding an inconsistent primary-chain suffix in security-parameter windows when a transaction’s producer can’t be resolved, then rolling back metadata and signaling resync to stop endless replays. Normal ingestion now fails fast on missing-input spends; Musashi Leios `ValidateNone` still treats absent inputs as no-ops.

- **Bug Fixes**
  - Replay recovery resolves missing inputs via metadata/CBOR/scan; if still unresolved, it rewinds the primary chain in ≤`k`-block steps, then rolls back metadata, emits `chainsync.resync`, and publishes `chain.ChainRollbackEvent` on `chain.update` with removed blocks, preserving the k-bounded event contract.
  - Metadata plugins for `sqlite`, `mysql`, and `postgres` now return `types.ErrUtxoNotFound` on absent consumed UTxOs, so normal replay and trust-mode commits fail fast; `StrictUtxoValidation` applies only to the recovery pass.
  - Added `AllowMissingConsumedInputs` to `database.BatchedTxIngestOpts`; propagated through `ledger.LedgerDelta` and batched writes. Used only with `SkipConsumedInputRecovery` for Musashi Leios `ValidateNone`, which filters consumed inputs to those present.

<sup>Written for commit f3b5468b17a4e3d72aa677c4f9bd838b82e3d1e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2973?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Missing consumed UTxOs now produce a clear validation error instead of being silently skipped.
  * Replay recovery can roll back the primary chain and prevent repeated processing of invalid blocks.
  * Recovery now handles unresolved transaction producers more safely.

* **New Features**
  * Added a specialized validation path for protocol scenarios where missing reference inputs are treated as no-ops.
  * Improved strict UTxO validation behavior and command-line guidance.

* **Documentation**
  * Expanded documentation describing transaction recovery, rollback, and missing-input handling.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1801,9 +1801,33 @@ detector itself breaks a loop by skipping a rollback it can no longer cross,
 it feeds that same point-keyed tracker so the escalation and metric fire on
 the skip path too, rather than silently suppressing the rollback.
 
-A separate recovery path handles transaction-validation failures that occur
-after the node has reached the chain tip (`recoverAtTipFromTxValidationError`
-in `ledger/replay_recovery.go`). When a block fails per-tx validation at tip,
+A separate recovery path handles transaction-validation failures during replay
+and after the node has reached the chain tip (`ledger/replay_recovery.go`).
+Replay first resolves missing inputs to producer blocks from metadata, CBOR
+offsets, or a bounded primary-chain scan; when it can resolve the producer, it
+rolls metadata back to the earliest producer's parent and replays the retained
+primary chain. An unresolvable input instead uses the security-parameter
+fallback anchor. Because that condition means the retained primary chain
+contains a consumer without its producer, recovery prunes the primary-chain
+suffix and rolls metadata back to the same point, then publishes
+`chainsync.resync` through the metadata rollback's local-ledger-rollback event
+so peers can redeliver a consistent suffix. A corrective rewind may span more
+than `k` blocks because the primary-chain tip can be ahead of the failing replay
+block. It is therefore performed as a sequence of ordinary rollbacks of at most
+`k` blocks each. Every step publishes the normal `chain.ChainRollbackEvent` on
+`chain.update`, preserves the event's bounded contract, and limits retained
+block payloads to one security-parameter window. If a later step fails, the
+primary chain remains at the last committed intermediate point; the standard
+startup/live reconciliation paths can synchronize metadata to that valid tip.
+The chain moves before metadata because the retained rollback anchor must remain
+queryable while metadata reconstructs its tip and nonce. If that metadata step
+fails, the primary chain is already at a valid target and the same reconciliation
+paths finish bringing metadata to its common ancestor. Live at-tip recovery uses
+the same bounded event-aware helper; startup-only speculative-tail cleanup stays
+eventless because subscribers have not begun consuming live chain events.
+Rewinding metadata alone would replay the same corrupt chain indefinitely.
+
+When a block fails per-tx validation at tip,
 the node rewinds the primary chain and rolls the ledger back so ChainSelection
 can re-pick a candidate chain; repeating the *same* `(block, tx)` failure
 escalates the rewind progressively deeper, up to the era stability window, to
--- a/ledger/replay_recovery.go
+++ b/ledger/replay_recovery.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/blinklabs-io/dingo/chain"
 	"github.com/blinklabs-io/dingo/database"
 	"github.com/blinklabs-io/dingo/database/models"
 	dbtypes "github.com/blinklabs-io/dingo/database/types"
@@ -119,6 +120,9 @@ type replayRecoveryCandidate struct {
 	ProducerBlock models.Block
 	RollbackPoint ocommon.Point
 	Strategy      string
+	// ProducerUnresolved distinguishes the security-parameter fallback from
+	// strategies that found a concrete producer. Strategy remains a log label.
+	ProducerUnresolved bool
 }
 
 type replayRecoveryPendingInput struct {
@@ -184,8 +188,14 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 	if candidate.ProducerTx != nil {
 		producerTxHash = hex.EncodeToString(candidate.ProducerTx.Hash)
 	}
+	rewindPrimaryChain := candidate.ProducerUnresolved &&
+		ls.config.ChainManager != nil
+	recoveryAction := "rewinding metadata state"
+	if rewindPrimaryChain {
+		recoveryAction = "rewinding primary chain and metadata state"
+	}
 	ls.config.Logger.Warn(
-		"detected inconsistent local ledger state during replay, rewinding metadata state",
+		"detected inconsistent local ledger state during replay, "+recoveryAction,
 		"component", "ledger",
 		"recovery_strategy", candidate.Strategy,
 		"tx_hash", hex.EncodeToString(validationErr.TxHash),
@@ -206,6 +216,20 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 		}
 		return true, nil
 	}
+	if rewindPrimaryChain {
+		if err := ls.rollbackPrimaryChainInSecurityParamWindows(
+			candidate.RollbackPoint,
+		); err != nil {
+			return false, fmt.Errorf(
+				"rewind primary chain for replay recovery: %w",
+				err,
+			)
+		}
+	}
+	// The chain moves first while the rollback anchor is guaranteed to remain
+	// available. If metadata synchronization fails, the primary chain is still
+	// at a valid retained point and the standard divergence reconciler can
+	// finish rolling metadata back to its common ancestor.
 	if err := ls.rollback(candidate.RollbackPoint); err != nil {
 		return false, fmt.Errorf(
 			"rollback ledger state for replay recovery: %w",
@@ -213,6 +237,73 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 		)
 	}
 	return true, nil
+}
+
+// rollbackPrimaryChainInSecurityParamWindows reaches a deep corrective rewind
+// by applying a sequence of ordinary, security-parameter-bounded rollbacks.
+// Each step therefore preserves ChainRollbackEvent's k-bounded contract and
+// bounds the blocks retained for event delivery. A failure leaves the chain at
+// the last committed intermediate point; startup/live reconciliation can then
+// replay or finish rolling metadata back to that valid primary-chain tip.
+func (ls *LedgerState) rollbackPrimaryChainInSecurityParamWindows(
+	point ocommon.Point,
+) error {
+	securityParam := ls.SecurityParam()
+	if securityParam <= 0 {
+		return chain.ErrSecurityParamNotConfigured
+	}
+
+	targetIndex := uint64(0)
+	if point.Slot > 0 || len(point.Hash) > 0 {
+		targetBlock, err := database.BlockByPoint(ls.db, point)
+		if err != nil {
+			return fmt.Errorf("lookup replay recovery target: %w", err)
+		}
+		targetIndex = targetBlock.ID
+	}
+
+	tip := ls.chain.Tip()
+	if tip.Point.Slot == point.Slot &&
+		bytes.Equal(tip.Point.Hash, point.Hash) {
+		return nil
+	}
+	tipBlock, err := database.BlockByPoint(ls.db, tip.Point)
+	if err != nil {
+		return fmt.Errorf("lookup primary chain tip: %w", err)
+	}
+	tipIndex := tipBlock.ID
+	if tipIndex < targetIndex {
+		return fmt.Errorf(
+			"primary chain tip index %d is behind replay recovery target %d",
+			tipIndex,
+			targetIndex,
+		)
+	}
+	window := uint64(securityParam)
+	for tipIndex-targetIndex > window {
+		nextIndex := tipIndex - window
+		nextBlock, err := ls.db.BlockByIndex(nextIndex, nil)
+		if err != nil {
+			return fmt.Errorf(
+				"lookup intermediate replay recovery block %d: %w",
+				nextIndex,
+				err,
+			)
+		}
+		nextPoint := ocommon.NewPoint(nextBlock.Slot, nextBlock.Hash)
+		if err := ls.chain.Rollback(nextPoint); err != nil {
+			return fmt.Errorf(
+				"rollback primary chain to intermediate point %d: %w",
+				nextIndex,
+				err,
+			)
+		}
+		tipIndex = nextIndex
+	}
+	if err := ls.chain.Rollback(point); err != nil {
+		return fmt.Errorf("rollback primary chain to recovery point: %w", err)
+	}
+	return nil
 }
 
 func (ls *LedgerState) recoveryRollbackExceedsMithrilBoundary(
@@ -398,7 +489,7 @@ func (ls *LedgerState) recoverAtTipFromTxValidationError(
 		"attempt", attempts,
 		"holding", ls.atTipRecoveryHolding,
 	)
-	if err := ls.config.ChainManager.RewindPrimaryChainToPoint(
+	if err := ls.rollbackPrimaryChainInSecurityParamWindows(
 		rewindPoint,
 	); err != nil {
 		return false, fmt.Errorf(
@@ -414,8 +505,8 @@ func (ls *LedgerState) recoverAtTipFromTxValidationError(
 	// looks up its inputs, finds them already marked consumed, and
 	// returns "rule 22 bad input(s) ... rule 24 value not conserved
 	// (consumed 0)" again, looping the recovery indefinitely until
-	// process restart. RewindPrimaryChainToPoint by design only touches
-	// the chain blob — the matching ledger rollback must be explicit.
+	// process restart. Primary-chain rollback only touches the chain
+	// store — the matching ledger rollback must be explicit.
 	if err := ls.rollback(rewindPoint); err != nil {
 		return false, fmt.Errorf(
 			"rollback ledger state after validation failure: %w",
@@ -484,7 +575,7 @@ func (ls *LedgerState) rejectRecoveryAtMithrilBoundary(
 		)
 	}
 	logRejection(mithrilLedgerSlot, rewindPoint)
-	if err := ls.config.ChainManager.RewindPrimaryChainToPoint(rewindPoint); err != nil {
+	if err := ls.rollbackPrimaryChainInSecurityParamWindows(rewindPoint); err != nil {
 		return fmt.Errorf(
 			"rewind primary chain to Mithril trust boundary: %w",
 			err,
@@ -866,10 +957,11 @@ func (ls *LedgerState) replayRecoveryFallbackCandidate(
 		return nil, err
 	}
 	return &replayRecoveryCandidate{
-		Input:         inputs[0],
-		ProducerBlock: anchorBlock,
-		RollbackPoint: rollbackPoint,
-		Strategy:      "security-param-fallback",
+		Input:              inputs[0],
+		ProducerBlock:      anchorBlock,
+		RollbackPoint:      rollbackPoint,
+		Strategy:           "security-param-fallback",
+		ProducerUnresolved: true,
 	}, nil
 }
 
--- a/ledger/replay_recovery_test.go
+++ b/ledger/replay_recovery_test.go
@@ -299,6 +299,7 @@ func TestTryRecoverFromTxValidationErrorRejectsReplayBelowMithrilBoundary(
 		},
 	})
 	require.NoError(t, err)
+	require.NoError(t, cm.SetLedger(ls))
 	ls.metrics.init(prometheus.NewRegistry())
 	shadowConnId := ouroboros.ConnectionId{
 		LocalAddr: &net.TCPAddr{
@@ -450,6 +451,7 @@ func TestTryRecoverFromTxValidationErrorAtTipRewindsPrimaryChain(
 		),
 	})
 	require.NoError(t, err)
+	require.NoError(t, cm.SetLedger(ls))
 	ls.metrics.init(prometheus.NewRegistry())
 
 	ledgerTip := ochainsync.Tip{
@@ -582,6 +584,7 @@ func newAtTipDescentLedger(t *testing.T) (*LedgerState, ochainsync.Tip) {
 		Logger:            slog.New(slog.NewJSONHandler(io.Discard, nil)),
 	})
 	require.NoError(t, err)
+	require.NoError(t, cm.SetLedger(ls))
 	ls.metrics.init(prometheus.NewRegistry())
 
 	ledgerTip := ochainsync.Tip{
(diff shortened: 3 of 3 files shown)


chrisguiney (changes requested):
The diagnosis is right and the primary fix — rewind the persistent primary chain, not just metadata, when the producer is unresolvable — is the correct answer to defect #1 in #2971. But the second half (making an absent consumed input a hard error in the metadata plugins) lands without a recovery path, breaks a documented bootstrap mode, and turns a config flag into a no-op while keeping its entire user-facing surface. Findings below, ordered by severity.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...`, `go vet` (database/ledger/chain) | clean |
| `go test -race ./database/...` | pass (sqlite plugin needs `-timeout 30m`; it took **605s**, just over the 600s default) |
| `go test -race ./chain/... ./ledger/... ./api/...` | pass |
| `go test ./internal/architecture` (import boundaries) | pass |
| `modernize ./...` | no findings on changed lines |

CI on this PR is green.

---

## Blocking

### 1. `--intersect-tip` bootstrap is broken with no replacement escape hatch

`metadata.SetTransaction` now returns `ErrUtxoNotFound` unconditionally for an absent consumed input. For a non-genesis chainsync intersect without a Mithril snapshot, `recoverConsumedUtxo` cannot succeed (the producer block was never fetched), so the first block after the intersect fails to apply.

The repo's own docs say this is exactly what the disabled default was for — and this PR deletes that text from `database/database.go` and `internal/config/config.go` while leaving it in place in `dingo.yaml.example:283-285` and `ARCHITECTURE.md:2133-2137`. After this change no configuration permits that mode.

(Code reading plus the repo's documentation; I did not stand up an intersect-tip node to confirm empirically.)

### 2. The new hard error has no recovery path and no backoff

`ledger/state.go:3877` routes only `*txValidationError` into recovery. `ErrUtxoNotFound` from the metadata commit surfaces as `process block batch: record transaction: …`, reaches `handleLedgerProcessBlocksError` (`state.go:3066-3080`), logs `"block processing failed, restarting pipeline"`, and the `for` loop in `ledgerProcessBlocks` restarts immediately — forever, with no delay, no rollback, and no resync. Only `errCertifiedEndorserBlockUnavailable` gets a timer.

This matters because trust-applied blocks (below `shouldValidateBlock`'s cutoff) are precisely the silent-corruption path #2971 describes, and they never produce a `txValidationError`. The issue asked for "a hard, loud failure **or** a chain-rewind + resync trigger"; today those blocks get a hot restart loop that cannot self-heal. Suggest routing `ErrUtxoNotFound`-during-apply into the same rewind+resync recovery, or at minimum giving it bounded retries with backoff and a distinct operator-facing outcome.

### 3. `StrictUtxoValidation` is now dead config

When `recoverConsumedUtxo` fails, every combination of (flag, Mithril boundary) ends in `ErrUtxoNotFound` — either the strict check errors, or the metadata commit does. The PR's own test proves it: all three subtests of `TestEnsureTransactionConsumedUtxosStrictValidation` now assert the identical outcome.

Keeping the YAML key, `CARDANO_STRICT_UTXO_VALIDATION`, `--strict-utxo-validation`, `WithStrictUtxoValidation`, and three doc paragraphs for a flag with no observable effect is worse than deleting it. Either remove it, or state precisely what still differs between the two settings.

---

## Should fix before merge

### 4. The new `ChainRollbackEvent` is not bounded by `k`

Every other emitter goes through `Chain.rollbackLocked`, which rejects `forkDepth > securityParam` (`chain/chain.go:816-824`). `RewindPrimaryChainToPointWithEvent` publishes the same event type with no such guard, and its target is `failingBlock.ID - SecurityParam()` while the primary chain tip can be far ahead of the failing block — which is the stuck state in #2971 (ledger roughly one stability window behind the chain tip). On mainnet that is on the order of 8-9k blocks.

`mempool`, `ledger/leios/pipeline.go:782`, `ledger/leios/manager.go:1653`, and `ledger/block_event.go:33` were all written against a `k`-bounded event. This is adjacent to #2941.

### 5. Unbounded memory in the collection

`rewindPrimaryChainToPoint` batches deletes at 512 (`chain/manager.go:39`) to bound per-transaction work, but `rolledBackBlocks` accumulates across *all* batches, and `models.Block` carries full `Cbor` (`database/models/block.go:28`). At the depth in #4 that is hundreds of MB held in a single slice.

It is then handed to `emitTransactionRollbackEvents`, which decodes every block and `PublishAsync`es a `TransactionEvent` per transaction into a 1000-slot queue that **drops on overflow** (`event/event.go:634-651`) — so the event is both expensive and delivered incompletely. Consider publishing per batch, or carrying points rather than whole blocks.

### 6. Asymmetry with the at-tip path

`ARCHITECTURE.md` frames this as mirroring at-tip recovery, but at-tip still uses the eventless `RewindPrimaryChainToPoint` at `replay_recovery.go:419`, `:505`, and `state.go:5560`. If the undo event is needed for correctness, those paths have the same hole; if it isn't, the replay path is paying #4 and #5 for nothing. Worth picking one.

### 7. `ChainManager == nil` silently degrades

`replay_recovery.go:214-216` returns `(false, nil)` *after* logging "rewinding primary chain and metadata state", skipping the metadata rollback the pre-PR code performed. Move the nil check above the log and fall back to metadata-only, or make it an error.

### 8. Magic string as control flow

`unresolvedProducer := candidate.Strategy == "security-param-fallback"` (`replay_recovery.go:187`). `Strategy` was a log label; it is now a behavior switch. Suggest adding `ProducerUnresolved bool` to `replayRecoveryCandidate` where `Strategy` is assigned (~line 890), or at least sharing a constant between producer and consumer. A typo here silently reverts to the broken metadata-only behavior with zero test failures.

### 9. Partial-failure ordering

The chain rewind commits before `ls.rollback` runs; if `ls.rollback` fails, blocks are gone from the blob store while metadata still reflects them. And if `rewindPrimaryChainToPoint` errors partway, earlier batches are already committed but `RewindPrimaryChainToPointWithEvent` returns before publishing, so consumers never hear about blocks that are already deleted. At minimum worth documenting why that is acceptable.

---

## Duplicated code

### 10. 107 lines triplicated across the three metadata plugins

I diffed the consume-reconciliation block: `mysql` and `postgres` are **byte-identical**; `sqlite` differs by exactly one line (`utxoRefIndexedTable()` vs `"UPDATE utxo"`). This PR edits the same six lines in all three copies.

`utxocond` already exists as a shared helper for the condition-building half — this reconciliation loop belongs there (or under `plugin/metadata/internal/`) too. Pre-existing debt, but this is at least the third change to touch all three copies.

### 11. Near-duplicate of an existing helper

`RewindPrimaryChainToPointWithEvent` + `ls.rollback` is close to `ls.rollbackChainAndState` (`ledger/state.go:2459`), which is `ls.chain.Rollback()` (publishes the event) + `ls.rollback()`. Worth a sentence in the PR description on why the existing helper couldn't be reused — presumably the `k` guard, which is exactly finding #4.

---

## Dead code / latent bug

### 12. `AllowMissingConsumedInputs` on the batched path is dead, and would be wrong if used

It is plumbed through `database/batch.go:82-87,267-275`, but the only caller (`internal/node/backfill.go:1091`) never sets it. If it were set, `metadataTransactionForConsumedInputs` queries `d.metadata.GetUtxoIncludingSpent` directly and cannot see in-flight producers held in the batch accumulator, so it would filter out inputs produced earlier in the same unflushed batch — the exact case `ensureTransactionConsumedUtxos` guards against with `inFlightProducerLookup` (`database/transaction.go:583-586`). Either drop the batched wiring or make it accumulator-aware.

### 13. `database/batch.go:86` states a guarantee its own path doesn't provide

"Normal block replay must leave this false so an absent input is a hard `ErrUtxoNotFound` failure" — but `batchSpendUtxos` (`sqlite/batch_accumulator.go:412-433`) never checks `RowsAffected` and silently no-ops missing rows regardless of the flag.

---

## Documentation

`DATABASE.md` addition is accurate. Two files were missed:

### 14. `dingo.yaml.example:281-285` not updated

Still describes the flag as controlling "silently skipping" and tells operators to leave it disabled for intersect bootstrap.

### 15. `ARCHITECTURE.md:2132-2137` not updated

It now directly contradicts the paragraph this PR adds at ~1804 in the same file:

> By default (`StrictUtxoValidation: false`) `ensureTransactionConsumedUtxos` silently skips an unrecoverable input, which is expected when bootstrapping from a non-genesis chainsync intersect point without a Mithril snapshot import (pre-intersect UTxOs are never imported).

---

## Tests

**16.** `TestEnsureTransactionConsumedUtxosStrictValidation` no longer tests strict validation (see #3). Rename or delete rather than leave a test whose name asserts a distinction its body disproves.

**17.** The new `"explicit ValidateNone path allows missing input"` subtest only asserts `NoError`. If `metadataTransactionForConsumedInputs` returned an empty slice unconditionally, the test still passes — and the Leios `ValidateNone` path would stop consuming *any* inputs. Use a fixture with one present and one absent input, and assert the present one is marked spent.

**18.** Three `ledger/utxo_storage_test.go` fixtures and one `api/utxorpc` fixture were opted out with `SkipConsumedInputRecovery: true` **in addition to** `AllowMissingConsumedInputs: true`. Only the latter is needed for the metadata guard; the former also disables `ensureTransactionConsumedUtxos` entirely, which is a real coverage reduction. Please confirm that half is intentional.

**19.** No test covers the behavior at the level #2971 describes: a trust-mode block application that spends an absent UTxO. Given finding #2, that is the test that would have surfaced the missing recovery path.

**20.** No mysql/postgres coverage for the identical change — only `sqlite/rollback_test.go` was updated. CI already runs both services, so this is cheap.

---

## Process

`CLAUDE.md` requires DevNet validation (`internal/test/devnet/run-tests.sh`) for changes touching consensus, chain selection, or ledger apply. This PR touches all three and changes what happens to a block that fails to apply. The description doesn't mention a DevNet run — please confirm one was done, or run it.

---

## What's good

Worth saying explicitly: the root-cause analysis is correct and the fix targets the right function — `SetTransaction`, not the batched path, which matches where #2971 pinned the silent skip. The newest-first ordering of `RolledBackBlocks` matches the documented contract in `chain/event.go:53`. The `Consumed()` filtering is correctly scoped and does not perturb the `Inputs()`/`Collateral()`/`ReferenceInputs()` index rows. `consumedInputsTransaction` follows the established `metadataOnlyTransaction` pattern. And the extended `TestTryRecoverFromTxValidationErrorFallsBackToSecurityParamWindow` genuinely exercises the rewind plus both emitted events, including block deletion — that is the right test for the core fix.

The chain-rewind half, findings 4-9 aside, is close to mergeable on its own. The metadata-hard-error half is where I'd push back hardest, and it could reasonably be split into its own PR once #1/#2/#3 have answers.

chrisguiney (approved):

Issue #1556, opened 2026-03-07
event: subscriber channel full during bulk sync


wolf31o2:
## Summary

During genesis sync on preview, the event bus drops `ledger.chainsync` events with "event dropped: subscriber channel full" warnings. The chain selection subscriber cannot consume events fast enough at bulk sync throughput, causing thousands of warning lines in the log. While not a correctness issue (chain selection only needs the latest tip, not every intermediate one), it creates significant log noise and wastes CPU on events that will be dropped.

## Current State

Observed during genesis sync on preview (2026-03-07). Within the first 12 minutes of syncing from genesis (reaching slot ~700K), the log accumulated ~49K lines, a large portion of which were subscriber-full warnings:

```json
{"time":"2026-03-07T13:15:55","level":"WARN","msg":"event dropped: subscriber channel full","type":"ledger.chainsync"}
```

The event bus uses a fixed-size buffered channel per subscriber. During bulk sync, the ledger publishes chainsync events at a rate that exceeds the chain selection subscriber's consumption rate. Each dropped event triggers a WARN log, amplifying the problem.

## What Needs to Happen

1. Prevent event bus log spam during bulk sync. The fix should not just increase buffer sizes — the fundamental issue is that some subscribers (chain selection) only care about the latest value, not every intermediate event.
2. Consider one or more of:
   - **Latest-value semantics**: Give chain selection a non-blocking subscription mode that overwrites the pending event with the newest one instead of queuing. This matches the subscriber's actual need (latest tip only).
   - **Bulk sync awareness**: Skip publishing `ledger.chainsync` events to non-essential subscribers when the node is far from tip (e.g., when `currentSlot < wallClockSlot - stabilityWindow`).
   - **Rate-limited warnings**: If events are still dropped, rate-limit the warning log to at most once per minute per event type instead of once per drop.
3. Ensure the fix does not affect tip-following behavior — all subscribers must still receive events at normal (near-tip) throughput.

## Deliverables

- [ ] Eliminate or rate-limit "subscriber channel full" warnings during genesis sync
- [ ] Chain selection still receives timely tip updates during normal tip-following
- [ ] No increase in memory usage during bulk sync (avoid unbounded buffering)
- [ ] Verify genesis sync completes without log spam on preview

wolf31o2:
[...]

Issue #2106, opened 2026-04-29
event: 100k channel buffer per subscriber dominates idle heap


wolf31o2:
## Summary

`event.newChannelSubscriber` is the largest in-use heap consumer on a running node at 192 MB, accounting for 45% of the live heap during a steady-state catch-up profile. The cost is `make(chan Event, buffer)` at `event/event.go:178`, where `buffer` defaults to 100,000 (raised from 10,000 in #1914) and every subscriber pre-allocates that headroom.

Roughly 190 subscribers exist on a serving node (block, chain, mempool, peer, governance, chainsync, etc.) and most are quiescent. The 100k slot buffer per subscriber is sized for the worst-case bulk-sync burst from #1556, but it stays allocated for the lifetime of the subscription. We pay the worst-case cost continuously.

## Reproduce

```
./dingo --network preview --debug-port 6060 ...
# wait for chainsync to settle
go tool pprof -top -inuse_space http://127.0.0.1:6060/debug/pprof/heap
```

Top in-use will show:

```
192.38MB 45.04%  github.com/blinklabs-io/dingo/event.newChannelSubscriber (inline)
```

caller is `event.(*EventBus).subscribeInternal`.

## What to investigate

1. Per-subscriber buffer should be a property of the subscriber, not a global default. A peergov reconcile callback does not need 100k slots; a chainsync block subscriber might.
2. Add an option to `EventBus.SubscribeFunc` (or a sibling) for the buffer size. Default to a small value (256 - 1024) and let busy subscribers opt up.
3. Audit current call sites to right-size each one. Burst tolerance should match what the subscriber actually catches up on after a slow processing turn.
4. If the worst-case bursts that motivated #1914 still need a 100k buffer somewhere, scope it to the specific subscribers that need it instead of the default.

Goal: bring the steady-state in-use heap from 192 MB closer to single-digit MB. That is the single biggest win available to RSS without changing storage at all.

## Numbers

Captured on bb52f1a7 against preview, 60 s heap snapshot, mid catch-up:

- Total in-use heap: 427 MB
- event.newChannelSubscriber: 192 MB (45.0%)
- Badger memtable arena: 166 MB (39.0%) — expected
- Goroutine count at snapshot: 898

Related: #1914 (queue size raised to 100k), #1556 (event subscriber channel full during bulk sync, mitigated by #1569).

PR #2978, opened 2026-07-27
fix(event): apply backpressure instead of dropping events

### LINKED ISSUE #2932: event: EventBus silently drops events during catch-up instead of applying backpressure
## Summary

Under sustained high-throughput block and tx application, the EventBus drops events instead of delivering them. Events should never be dropped, including during catch-up. Any subscriber that builds derived state from the event stream (metrics, API/indexer adapters, downstream consumers) loses data silently with no recovery path.

Observed on a musashi from-genesis sync on current main, and previously reported for bulk sync in the now-closed #1556.

## Where

Two non-blocking-send-then-drop sites in event/event.go:

1. channelSubscriber.Deliver (event/event.go:205-238). Non-blocking send to a per-subscriber buffered channel. When the buffer is full the event is dropped with a WARN "event dropped: subscriber channel full". A blocking variant already exists (DeliverBlocking, event/event.go:240-261) that retries instead of dropping.

2. EventBus.PublishAsync (event/event.go:625-651). Non-blocking enqueue to the shared async queue. When the queue is full the event is dropped with a WARN "async event queue full, dropping event" and the metric deliveryErrors{type,"async-dropped"} is incremented.

## Reproduction

Sync musashi (Leios testnet, magic 164) from genesis on current main. Entering the tx-dense region (epoch ~6-12), block and endorser-block application outpaces the async subscribers and both WARNs appear at high volume. Event types seen so far: chain.update and ledger.tx. The node keeps progressing (this is not a wedge), but events are being discarded.

## Impact

Subscribers that consume every chain.update or ledger.tx to build derived state will silently miss events. There is no backfill or replay, so the loss is permanent for that subscriber. This affects correctness of anything downstream of the event stream, not just observability.

## Prior art

- #1556 event: subscriber channel full during bulk sync (closed)
- #1753 fix(event): chainsync handler blocks on mutex, causing subscriber channel overflow (closed)
- #2106 event: 100k channel buffer per subscriber dominates idle heap (closed) reduced the per-subscriber buffer to cut idle heap, which lowers the headroom before drops occur.

The recurrence on current main indicates the non-blocking-drop design still loses events under sustained load.

## Desired behavior

No dropped events, including during catch-up. Options to evaluate:

- Blocking delivery with producer backpressure so ingestion slows rather than dropping (DeliverBlocking already implements a retry loop).
- A bounded async queue that blocks the producer when full instead of dropping.
- Per-subscriber flow control so a single slow subscriber slows ingestion rather than losing events.

Any solution needs to preserve the Close() non-deadlock property noted in the Deliver comment (the current non-blocking send exists to avoid blocking while holding the read lock and deadlocking Close), and keep memory bounded per the concern in #2106.

chrisguiney:
## Summary

The EventBus dropped events at two sites, so any subscriber building derived
state from the stream lost data silently with no backfill or replay:

- `channelSubscriber.Deliver` — non-blocking send, discarded the event when the
  per-subscriber buffer was full
- `EventBus.PublishAsync` — discarded when the shared async queue was full

Both now apply producer backpressure instead. Fixes blinklabs-io/dingo#2932.

## How the Close() deadlock is avoided

The non-blocking send existed for a reason: `Deliver` holds `mu.RLock` while
sending, and `Close()` needs `mu.Lock`, so a blocking send would deadlock
`Close`. `Close` now closes a `closeReq` channel **before** asking for the
write lock, so every waiting send wakes, returns, and releases the read lock
first. That keeps the non-deadlock property without losing events, and buffers
stay bounded (no regression against #2106).

Shutdown closes `stopCh` before taking `stopMu` for write. A publisher parked
on a full buffer or queue holds `stopMu` for read, and `stopCh` is what
releases it, so closing it after acquiring the write lock would deadlock
shutdown against its own backpressure.

## Observability

`deliveryErrors{type,"async-dropped"}` is gone — nothing drops. Replaced by
`event_delivery_blocked_total{type,kind}` and
`event_async_enqueue_blocked_total{type}`, plus a throttled
`event delivery stalled` warning. Blocking is normal under load, so only a
genuine stall is reported, which also avoids the log spam of #1556.

## Second commit: a bug this unmasks

`replayRecoveryParentPoint` guarded the no-parent case with
`len(block.PrevHash) == 0`, but a forged first block encodes its genesis
predecessor as an all-zero 32-byte hash. The guard missed it, so replay
recovery looked up a block that cannot exist and restarted the ledger pipeline
in a loop.

That path was previously unreachable only because the events leading into it
were among those being dropped. On DevNet it pinned a producer at its own
block 0 (slot 4) for ~4 minutes while the rest of the network advanced.

| metric | baseline (main) | before ledger fix | after |
|---|---|---|---|
| `lookup parent block ... not found` | 0 | 118 | 0 |
| pipeline restarts | 8 | 122 | 4 |
| `TestSustainedConsensus` | FAIL | PASS | PASS |

## Trade-offs

A slow subscriber now slows its producers instead of losing their events —
that is the point of the issue. The 4 shared async workers can all park on one
slow subscriber, delaying unrelated async event types; this is documented in
`event/doc.go` and `ARCHITECTURE.md`. Every `Subscribe()` consumer in the tree
was audited: all unsubscribe on exit, and `Unsubscribe` releases blocked
producers.

## Testing

- 15 new tests in `event/` covering no-loss delivery, blocked-send/Close
  interaction, unblock-on-Stop/Close/Unsubscribe, and the metrics
- 4 existing tests inverted — they asserted the drop behavior being removed
- 1 new `ledger/` regression test, verified to fail without the guard
- `go test -race ./...` green; `golangci-lint`, `modernize`,
  `make import-boundaries` clean
- DevNet run A/B'd against `main`, sequentially with a full Docker wipe between
  (they share container names). `TestChainGrowthRate` still fails, but fails
  identically on `main` in the same configuration (62.62s vs 62.57s)

## Docs

`ARCHITECTURE.md` and `event/doc.go` updated for the lossless-delivery
contract and its head-of-line-blocking consequence. `DATABASE.md` checked —
not affected.

---
## Summary by cubic
Apply backpressure in the `event` bus instead of dropping events, ensuring lossless delivery and clean shutdown. Also fixes replay recovery in `ledger` for blocks that reference the genesis predecessor.

- **Bug Fixes**
  - `Publish`, `PublishBlocking`, and `PublishAsync` now wait when a subscriber buffer or the async queue is full; no events are dropped.
  - Blocked publishers unblock on `Stop`, `Close`, or `Unsubscribe`; `PublishBlocking` returns `ErrEventBusStopped` if shutdown begins mid-delivery.
  - Replaced drop behavior and log spam with `event_delivery_blocked_total` and `event_async_enqueue_blocked_total`, plus a throttled “event delivery stalled” warning; removed `event_delivery_errors_total{kind="async-dropped"}`.
  - Treat an all-zero `PrevHash` as the genesis predecessor in replay recovery to avoid infinite restart loops.

- **Migration**
  - Slow subscribers must drain their `Subscribe` channels or call `Unsubscribe`; shared async workers mean head-of-line blocking across async event types.
  - Update dashboards/alerts to use `event_delivery_blocked_total` and `event_async_enqueue_blocked_total`; the old async drop counter is gone.

<sup>Written for commit 835a11edfe91a003fcfe3fca7c959652109077f4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2978?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Enhancements**
  - Event delivery now preserves all events by waiting for subscriber or async queue capacity instead of dropping messages.
  - Publishing unblocks cleanly when the bus stops, closes, or a subscriber unsubscribes.
  - Added visibility into delivery backpressure and stalled subscribers through metrics and warnings.

- **Bug Fixes**
  - Improved recovery for blocks immediately following genesis when the predecessor hash is zero-valued.

- **Documentation**
  - Clarified EventBus delivery guarantees, backpressure behavior, shutdown handling, and subscriber draining requirements.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -809,9 +809,18 @@ All event types follow the `subsystem.snake_case_name` convention.
 - Asynchronous delivery via worker pool (4 workers, 1000-entry async queue)
 - Default subscriber buffers of 1024 events, with opt-in 100000-entry burst
   buffers for high-volume ledger chainsync/blockfetch paths
-- Non-blocking `Publish`, blocking `PublishBlocking` for ordering-critical
-  streams, and `PublishAsync` for best-effort async work
-- Prometheus metrics for event delivery tracking and latency
+- Lossless delivery with producer backpressure: when a subscriber buffer or the
+  async queue is full, `Publish`, `PublishBlocking`, and `PublishAsync` all wait
+  for capacity instead of dropping the event. Waits end on `Stop`, `Close`, or
+  `Unsubscribe`. `PublishBlocking` differs only in returning
+  `ErrEventBusStopped` when the bus shuts down mid-delivery
+- A subscriber that stops draining stalls its publishers, and the shared async
+  worker pool means it also delays unrelated async event types. Consumers of
+  `Subscribe` channels must drain for the life of the subscription and
+  `Unsubscribe` when they stop
+- Prometheus metrics for event delivery tracking and latency, including
+  `event_delivery_blocked_total{type,kind}` and
+  `event_async_enqueue_blocked_total{type}` for backpressure
 
 ## Storage Architecture
 
--- a/event/backpressure_ext_test.go
+++ b/event/backpressure_ext_test.go
@@ -1,0 +1,480 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package event_test
+
+import (
+	"maps"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/blinklabs-io/dingo/event"
+	"github.com/blinklabs-io/dingo/internal/test/testutil"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+)
+
+// The tests in this file cover blinklabs-io/dingo#2932: neither the
+// per-subscriber channel nor the shared async queue may discard events under
+// sustained load, and backpressure must not wedge shutdown.
+
+// TestPublishDeliversEveryEventUnderBackpressure is the headline regression:
+// a producer far faster than its subscriber loses nothing.
+func TestPublishDeliversEveryEventUnderBackpressure(t *testing.T) {
+	const testEvtType event.EventType = "test.nodrop.publish"
+	const eventCount = 5000
+
+	eb := event.NewEventBus(nil, nil)
+	defer eb.Stop()
+
+	// Deliberately tiny buffer: the producer spends nearly all its time
+	// backpressured.
+	_, subCh := eb.SubscribeWithBuffer(testEvtType, 8)
+
+	received := make(chan []int, 1)
+	go func() {
+		got := make([]int, 0, eventCount)
+		for evt := range subCh {
+			v, ok := evt.Data.(int)
+			if !ok {
+				continue
+			}
+			got = append(got, v)
+			if len(got) == eventCount {
+				received <- got
+				return
+			}
+		}
+		received <- got
+	}()
+
+	for i := range eventCount {
+		eb.Publish(testEvtType, event.NewEvent(testEvtType, i))
+	}
+
+	got := testutil.RequireReceive(
+		t,
+		received,
+		10*time.Second,
+		"subscriber did not receive every published event",
+	)
+	require.Len(t, got, eventCount, "no event may be dropped")
+	for i, v := range got {
+		require.Equal(t, i, v, "events must arrive in publish order")
+	}
+}
+
+// TestPublishAsyncDeliversEveryEventUnderBackpressure covers the second drop
+// site: the shared async queue. This is the ledger.tx shape from the issue.
+func TestPublishAsyncDeliversEveryEventUnderBackpressure(t *testing.T) {
+	const testEvtType event.EventType = "test.nodrop.async"
+	const eventCount = 5000
+
+	eb := event.NewEventBus(nil, nil)
+	defer eb.Stop()
+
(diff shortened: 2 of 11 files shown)


chrisguiney:
Thanks — addressed both.

## `isGenesisPrevHash` length constraint (`ledger/replay_recovery.go`) — fixed

Valid catch. The helper classified *any* all-zero slice as the genesis
predecessor, so a malformed 1-byte or 40-byte value would be silently rolled
back to origin instead of surfacing as a failed lookup. Now constrained to the
two encodings that legitimately mean "no parent":

```go
if len(prevHash) == 0 {
    return true          // decoded block carrying no prev hash
}
if len(prevHash) != lcommon.Blake2b256Size {
    return false         // malformed: fall through to the lookup
}
// ... all-zero check
```

Anything else keeps falling through to `BlockByHash`, so corruption still
surfaces. Added `TestIsGenesisPrevHashRejectsMalformedHashes` covering nil,
empty, 32-zero, 1-zero, 16-zero, 40-zero, a real hash, and 31-zero-plus-one —
three of those cases failed against the previous version.

## Out-of-scope warning on the ledger commit — pushing back, but happy to defer

The ledger fix isn't incidental scope creep; it is **causally coupled** to the
event change and I'd argue against splitting it.

The bug is unreachable on `main` precisely *because* the events that lead into
it are among the ones being dropped. Removing the drops is what makes it
reachable. Measured on a sequential DevNet A/B (same host, full Docker wipe
between runs, since both share container names):

| | baseline (`main`) | this branch, before the ledger fix | after |
|---|---|---|---|
| `lookup parent block ... block not found` | 0 | 118 | 0 |
| pipeline restarts | 8 | 122 | 4 |

Before the fix, a producer sat pinned at its own block 0 (slot 4) for ~4
minutes on a fresh network while the rest of the chain advanced. Splitting
means this PR knowingly ships that liveness regression until the follow-up
lands.

Taking the second option offered — "link the relevant issue" — the causal
explanation above *is* the linkage, and it's recorded in the commit message
too.

That said, this is the maintainers' call, not mine. If you'd prefer it split,
say the word and I'll open the ledger fix as its own PR and rebase this one on
top of it, so the merge order still avoids the regression window.

wolf31o2 (approved):

PR #2979, opened 2026-07-27
feat(mempool): dag-based mempool implementation

wolf31o2:
---
## Summary by cubic
Adds a DAG-based mempool backend with deterministic parent-before-descendant ordering, shared double-buffered revalidation, and TxSubmission backpressure. FIFO stays the default; nodes can choose `fifo` or `dag` via plugins with no public API changes.

- New Features
  - Providers: `plugins.mempool.provider: fifo|dag` (default `fifo`; `default` aliases `fifo`). Host registers `default`, `fifo`, `dag`. Example config and devnet support `DEVNET_MEMPOOL_PROVIDER`. Built-ins default capacity for `default|fifo|dag`. Metric: `dingo_metrics_mempool_info{implementation="fifo|dag"}`.
  - DAG backend: explicit producer/spender indexes and parent/child edges; caches topological order and falls back to admission order if inconsistent; descendant-aware expiry/removal; confirmed removals detach only; no watermark eviction (capacity managed by intake headroom).
  - Service contract: node now uses `mempool.Service`; FIFO and DAG share the same engine and metrics. Revalidation is shared for both providers: double-buffered rebuild with bounded mutation journal, ledger-generation pin/check, and capped catch-up. Missing overlay parents poison dependents to avoid retaining orphan descendants.
  - Admission/backpressure (DAG only): exposes `AdmissionHeadroom` to gate intake; TxSubmission waits for headroom before requesting IDs/bodies, requests 1 ID per round trip, rejects oversize offers, and retries capacity failures with a bounded streak. Records `dingo_txsubmission_admission_retry_streak`.

- Migration
  - No change if you stay on FIFO. To enable DAG, set `plugins.mempool.provider: dag` (or `DINGO_PLUGINS_MEMPOOL_PROVIDER=dag`) and restart.
  - DAG ignores `evictionWatermark`; tune with `capacity` and `rejectionWatermark`. Events and APIs are unchanged.

<sup>Written for commit 5a0356cd8fee43514dce7886612fa156e607fb72. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2979?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an optional DAG-based transaction mempool alongside FIFO.
  * Added provider selection through configuration, with FIFO as the default and `default` retained as an alias.
  * Added dependency-aware transaction ordering and improved admission backpressure for DAG mempools.
  * Mempool metrics now identify the active backend.

* **Documentation**
  * Updated architecture, configuration examples, and development guidance for FIFO and DAG providers.

### DIFF
--- a/mempool/dag.go
+++ b/mempool/dag.go
@@ -15,7 +15,7 @@
 package mempool
 
 import (
-	"container/heap"
+	"fmt"
 	"slices"
 )
 
@@ -23,11 +23,8 @@ import (
 // Mempool so both implementations preserve the same immutable snapshot and
 // relay behavior.
 type dagNode struct {
-	hash     string
-	sequence uint64
 	parents  map[string]struct{}
 	children map[string]struct{}
-	consumed []string
 	produced []string
 }
 
@@ -37,44 +34,13 @@ type dagNode struct {
 type transactionDAG struct {
 	nodes          map[string]*dagNode
 	producerByUtxo map[string]string
-	spenderByUtxo  map[string]string
-	nextSequence   uint64
-}
-
-type dagReadyQueue []*dagNode
-
-func (q *dagReadyQueue) Len() int {
-	return len(*q)
-}
-
-func (q *dagReadyQueue) Less(i, j int) bool {
-	if (*q)[i].sequence != (*q)[j].sequence {
-		return (*q)[i].sequence < (*q)[j].sequence
-	}
-	return (*q)[i].hash < (*q)[j].hash
-}
-
-func (q *dagReadyQueue) Swap(i, j int) {
-	(*q)[i], (*q)[j] = (*q)[j], (*q)[i]
-}
-
-func (q *dagReadyQueue) Push(value any) {
-	*q = append(*q, value.(*dagNode))
-}
-
-func (q *dagReadyQueue) Pop() any {
-	old := *q
-	last := len(old) - 1
-	node := old[last]
-	*q = old[:last]
-	return node
+	order          []string
 }
 
 func newTransactionDAG() *transactionDAG {
 	return &transactionDAG{
 		nodes:          make(map[string]*dagNode),
 		producerByUtxo: make(map[string]string),
-		spenderByUtxo:  make(map[string]string),
 	}
 }
 
@@ -83,16 +49,11 @@ func (d *transactionDAG) add(tx appliedTx) {
 		return
 	}
 	node := &dagNode{
-		hash:     tx.hash,
-		sequence: d.nextSequence,
 		parents:  make(map[string]struct{}),
 		children: make(map[string]struct{}),
-		consumed: slices.Clone(tx.consumed),
 		produced: make([]string, 0, len(tx.created)),
 	}
-	d.nextSequence++
 	for _, input := range tx.consumed {
-		d.spenderByUtxo[input] = tx.hash
 		if parentHash, ok := d.producerByUtxo[input]; ok {
 			node.parents[parentHash] = struct{}{}
 		}
@@ -101,8 +62,10 @@ func (d *transactionDAG) add(tx appliedTx) {
 		node.produced = append(node.produced, output)
 		d.producerByUtxo[output] = tx.hash
 	}
-	slices.Sort(node.produced)
 	d.nodes[tx.hash] = node
+	// A pending parent must already be admitted before a child can resolve its
+	// output. Admission order is therefore itself a stable topological order.
+	d.order = append(d.order, tx.hash)
 	for parentHash := range node.parents {
 		if parent := d.nodes[parentHash]; parent != nil {
 			parent.children[tx.hash] = struct{}{}
@@ -160,11 +123,6 @@ func (d *transactionDAG) remove(hashes map[string]struct{}) {
 				delete(child.parents, hash)
 			}
 		}
-		for _, input := range node.consumed {
-			if d.spenderByUtxo[input] == hash {
-				delete(d.spenderByUtxo, input)
-			}
-		}
 		for _, output := range node.produced {
 			if d.producerByUtxo[output] == hash {
 				delete(d.producerByUtxo, output)
@@ -172,6 +130,10 @@ func (d *transactionDAG) remove(hashes map[string]struct{}) {
 		}
 		delete(d.nodes, hash)
 	}
+	d.order = slices.DeleteFunc(d.order, func(hash string) bool {
+		_, remove := hashes[hash]
+		return remove
+	})
 }
 
 func (d *transactionDAG) rebuild(applied []appliedTx) {
@@ -182,29 +144,17 @@ func (d *transactionDAG) rebuild(applied []appliedTx) {
 	*d = *replacement
 }
 
-// topologicalOrder returns a stable Kahn traversal. Admission sequence is the
-// primary ready-frontier key and transaction hash is a defensive tie-breaker.
-func (d *transactionDAG) topologicalOrder() []string {
-	indegree := make(map[string]int, len(d.nodes))
-	ready := make(dagReadyQueue, 0, len(d.nodes))
-	for hash, node := range d.nodes {
-		indegree[hash] = len(node.parents)
-		if len(node.parents) == 0 {
-			heap.Push(&ready, node)
-		}
+// topologicalOrder returns the cached stable ordering maintained at mutation
+// time. A count mismatch is an internal index failure, not a representable
+// partial answer: callers must diagnose it and use a complete fallback.
+func (d *transactionDAG) topologicalOrder() ([]string, error) {
+	ret := slices.Clone(d.order)
+	if len(ret) != len(d.nodes) {
+		return ret, fmt.Errorf(
+			"DAG index inconsistent: %d of %d transactions ordered",
+			len(ret),
+			len(d.nodes),
+		)
 	}
-	ret := make([]string, 0, len(d.nodes))
-	for len(ready) > 0 {
-		node := heap.Pop(&ready).(*dagNode)
-		ret = append(ret, node.hash)
-		for childHash := range node.children {
-			indegree[childHash]--
-			if indegree[childHash] == 0 {
-				if child := d.nodes[childHash]; child != nil {
-					heap.Push(&ready, child)
-				}
-			}
-		}
-	}
-	return ret
+	return ret, nil
 }
--- a/mempool/dag.go
+++ b/mempool/dag.go
@@ -1,0 +1,210 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mempool
+
+import (
+	"container/heap"
+	"slices"
+)
+
+// dagNode stores only graph/index metadata. Transaction bodies remain owned by
+// Mempool so both implementations preserve the same immutable snapshot and
+// relay behavior.
+type dagNode struct {
+	hash     string
+	sequence uint64
+	parents  map[string]struct{}
+	children map[string]struct{}
+	consumed []string
+	produced []string
+}
+
+// transactionDAG indexes producer/consumer relationships between pending
+// transactions. All methods are called under Mempool's mutation and state
+// locks, so the graph itself does not need a mutex.
+type transactionDAG struct {
+	nodes          map[string]*dagNode
+	producerByUtxo map[string]string
+	spenderByUtxo  map[string]string
+	nextSequence   uint64
+}
+
+type dagReadyQueue []*dagNode
+
+func (q *dagReadyQueue) Len() int {
+	return len(*q)
+}
+
+func (q *dagReadyQueue) Less(i, j int) bool {
+	if (*q)[i].sequence != (*q)[j].sequence {
+		return (*q)[i].sequence < (*q)[j].sequence
+	}
+	return (*q)[i].hash < (*q)[j].hash
+}
+
+func (q *dagReadyQueue) Swap(i, j int) {
+	(*q)[i], (*q)[j] = (*q)[j], (*q)[i]
+}
+
+func (q *dagReadyQueue) Push(value any) {
+	*q = append(*q, value.(*dagNode))
+}
+
+func (q *dagReadyQueue) Pop() any {
+	old := *q
+	last := len(old) - 1
+	node := old[last]
+	*q = old[:last]
+	return node
+}
+
+func newTransactionDAG() *transactionDAG {
+	return &transactionDAG{
+		nodes:          make(map[string]*dagNode),
+		producerByUtxo: make(map[string]string),
+		spenderByUtxo:  make(map[string]string),
+	}
+}
+
+func (d *transactionDAG) add(tx appliedTx) {
+	if _, exists := d.nodes[tx.hash]; exists {
+		return
+	}
+	node := &dagNode{
+		hash:     tx.hash,
+		sequence: d.nextSequence,
+		parents:  make(map[string]struct{}),
+		children: make(map[string]struct{}),
+		consumed: slices.Clone(tx.consumed),
+		produced: make([]string, 0, len(tx.created)),
+	}
+	d.nextSequence++
+	for _, input := range tx.consumed {
+		d.spenderByUtxo[input] = tx.hash
+		if parentHash, ok := d.producerByUtxo[input]; ok {
+			node.parents[parentHash] = struct{}{}
+		}
+	}
+	for output := range tx.created {
+		node.produced = append(node.produced, output)
+		d.producerByUtxo[output] = tx.hash
+	}
+	slices.Sort(node.produced)
+	d.nodes[tx.hash] = node
+	for parentHash := range node.parents {
+		if parent := d.nodes[parentHash]; parent != nil {
+			parent.children[tx.hash] = struct{}{}
+		}
+	}
+}
+
+// descendants returns roots and every transitively dependent transaction.
+func (d *transactionDAG) descendants(
+	roots map[string]struct{},
+) map[string]struct{} {
+	ret := make(map[string]struct{}, len(roots))
+	queue := make([]string, 0, len(roots))
+	for hash := range roots {
+		if _, ok := d.nodes[hash]; !ok {
+			continue
+		}
+		ret[hash] = struct{}{}
+		queue = append(queue, hash)
+	}
+	for len(queue) > 0 {
+		hash := queue[0]
+		queue = queue[1:]
+		node := d.nodes[hash]
+		if node == nil {
+			continue
+		}
+		for childHash := range node.children {
+			if _, seen := ret[childHash]; seen {
+				continue
+			}
+			ret[childHash] = struct{}{}
+			queue = append(queue, childHash)
+		}
+	}
+	return ret
+}
+
+// remove deletes nodes and their indexes. Edges from surviving children are
+// detached. This is correct for confirmed removal: the former parent's outputs
+// are now resolved through ledger state instead of the pending graph.
+func (d *transactionDAG) remove(hashes map[string]struct{}) {
+	for hash := range hashes {
+		node := d.nodes[hash]
+		if node == nil {
+			continue
+		}
+		for parentHash := range node.parents {
+			if parent := d.nodes[parentHash]; parent != nil {
+				delete(parent.children, hash)
+			}
+		}
+		for childHash := range node.children {
+			if child := d.nodes[childHash]; child != nil {
+				delete(child.parents, hash)
+			}
+		}
+		for _, input := range node.consumed {
+			if d.spenderByUtxo[input] == hash {
+				delete(d.spenderByUtxo, input)
+			}
+		}
+		for _, output := range node.produced {
+			if d.producerByUtxo[output] == hash {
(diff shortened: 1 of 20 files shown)


wolf31o2:
@coderabbitai review

mempool/dag_revalidation.go
```diff
@@ -0,0 +1,320 @@
...
+	for _, mutation := range mutations {
+		if mutation.added != nil {
+			size++
+		}
+		size += len(mutation.removed)
+	}
+	return size
+}
+
+func (m *Mempool) applyDAGRevalidationTx(
```
chrisguiney: **Blocking.** `applyDAGRevalidationTx` calls `m.validator.ValidateTxWithOverlay(...)` once per transaction. Each of those opens its own DB transaction and takes its own ledger snapshot inside `validateTxCore`.

That is precisely the pattern #2986 replaced for FIFO, in response to #2981's explicit requirement:

> "The candidate must not be assembled from per-transaction validations that may observe different ledger states."

#2986 added `LedgerState.WithTxValidationSession` to pin *ledger tip/generation, current era and applicable protocol parameters, and the database read snapshot/transaction* for the whole rebuild. This path doesn't use it — it substitutes the new `TxValidationGeneration()` before/after comparison.

To be fair to the design: generations are monotonic, and ledger writes happen inside the transactions that bump them, so an equality check across the rebuild does prove no publication landed in between. This may well be sound. But it's a *different* guarantee reached by a *different* mechanism than the one the sibling PR established for the same stated requirement, and nothing in the code or description explains why the cheaper guard suffices here when a pinned session was deemed necessary there.

I'd want either the DAG routed through `WithTxValidationSession` like FIFO, or a comment stating why generation-stability alone is sufficient. Secondary cost either way: N separate DB read transactions per rebuild instead of one, which matters for a large mempool.

mempool/dag_revalidation.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}
+	return events
+}
+
+// rebuildDAGOverlay performs full DAG revalidation off the mutation gate, then
+// converges on concurrent additions and removals through a mutation journal.
+// Only a bounded residual is processed while holding the gate before the
+// candidate overlay and graph are atomically swapped.
+func (m *Mempool) rebuildDAGOverlay() error {
```
chrisguiney: **Blocking.** This file reimplements the snapshot -> off-lock revalidate -> journal catch-up -> generation check -> atomic commit architecture that `mempool.go` now has for FIFO, with entirely separate parts:

| | FIFO (merged in #2986) | DAG (this PR) |
|---|---|---|
| Journal | `mutationJournal []mempoolMutation`, `journalActive`, `journalOverflow` | `dagMutation`, `dagRevalidationActive` |
| Catch-up bound | `maxRevalidationCatchupRounds = 16` | `dagRevalidationMaxIters = 8` |
| Delta cap | `revalidationDeltaCap` (**configurable**, default 64) | `dagRevalidationDeltaCap = 256` (**hardcoded**) |
| Generation guard | `stillCurrent()` from the pinned session | `TxValidationGeneration()` before/after |
| Retry bound | `for attempt := range 2` | `dagRevalidationMaxGenerationRetries = 1` |
| Overflow valve | `errRevalidationJournalOverflow` | none — no journal cap |

Same algorithm, two sets of constants, two retry semantics, two journals. A bug found in one won't be fixed in the other, and both guard block production.

The shared `runBackendContract` suite is the right instinct — extending that idea to the revalidation machinery (one journal + catch-up driver parameterised by each backend's apply/commit) would make the DAG backend a small delta rather than a parallel universe.

mempool/dag_revalidation.go
```diff
@@ -0,0 +1,320 @@
...
+	"slices"
+
+	"github.com/blinklabs-io/dingo/event"
+	gledger "github.com/blinklabs-io/gouroboros/ledger"
+)
+
+const (
+	// dagRevalidationDeltaCap bounds the transaction-sized mutation delta
+	// processed while holding the mutation gate at the final commit.
+	dagRevalidationDeltaCap = 256
```
chrisguiney: Two asymmetries with FIFO worth resolving in whichever direction you prefer:

- `revalidationDeltaCap` is operator-tunable for FIFO (config key, default 64, validated `> 0`) but hardcoded at 256 here. An operator who needs to tune a DAG rebuild down to fit a busy chain has no knob.
- There is no DAG journal cap at all, so FIFO's `journalOverflow` safety valve (`errRevalidationJournalOverflow`) has no equivalent — a sustained admission burst grows `dagMutation` entries without bound during a rebuild.

The second matters more: FIFO deliberately aborts a candidate rather than let the journal grow unbounded, and that reasoning applies here identically.

mempool/dag_revalidation.go
```diff
@@ -0,0 +1,320 @@
...
+	// dagRevalidationMaxIters prevents an unbounded catch-up loop if admission
+	// is at least as fast as revalidation.
+	dagRevalidationMaxIters = 8
+	// dagRevalidationMaxGenerationRetries prevents a busy retry loop when
+	// ledger publications advance faster than Dingo's full transaction
+	// validation can rebuild a large candidate. A later chain event retries.
+	dagRevalidationMaxGenerationRetries = 1
```
chrisguiney: One retry, then the rebuild returns an error and the overlay stays stale until the next chain event happens to land in a quiet window.

FIFO's equivalent is also 2 attempts, so on its own this is comparable. What makes it carry more weight here is that FIFO pairs its retry bound with a *configurable* delta cap, so an operator can tune the rebuild small enough to land between publications. Per the comment above, the DAG can't be tuned the same way — so on a busy chain this is more likely to be the terminal outcome than it is for FIFO.

ouroboros/txsubmission.go
```diff
@@ -163,14 +163,29 @@ func (o *Ouroboros) txsubmissionServerInit(
...
+					1,
+					conn.ErrorChan(),
+				) {
+					return
+				}
+				// Request one ID at a time because gouroboros acknowledges every
+				// ID in a reply on the next request. Requesting a larger batch
+				// and fetching only the prefix that fits would silently discard
+				// the unrequested remainder.
+				requestCount = 1
```
chrisguiney, line 221: For DAG nodes this drops ID requests from `txsubmissionRequestTxIdsCount` to 1, which is roughly one transaction per round-trip on a high-latency peer link.

The justification is sound and specific — gouroboros acknowledges every ID in the reply on the next request, so fetching only the prefix that fits would silently discard the remainder. The type-assertion gating on `mempool.AdmissionHeadroom` is also correctly scoped, so FIFO nodes are untouched.

Two asks: (1) this is a meaningful throughput cost of choosing DAG and it isn't in ARCHITECTURE.md or the migration notes — an operator switching providers should know before they switch; (2) it's a workaround for a gouroboros-side limitation, so a pointer to that (issue or a note that partial acknowledgement would remove the need) would keep it from quietly becoming permanent.

chrisguiney (changes requested):
Requesting changes on how the DAG revalidation half relates to PR #2986, which merged yesterday. The DAG itself is well built and the shared contract-test structure is right — my objections are about the revalidation engine.

Two blocking items:

1. **DAG revalidation validates per-transaction** rather than through the `WithTxValidationSession` that #2986 added for exactly this requirement.
2. **It is a second independent implementation** of the same snapshot/journal/catch-up/commit machinery, with its own constants and semantics.

Three non-blocking comments inline, plus a process note below.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...`, `go vet` (mempool/ouroboros/ledger/internal) | clean |
| `go test -race -count=1 ./mempool/... ./ouroboros/...` | pass |
| CI on this PR | 11 checks pass, 0 failing |

No `Closes` reference. FIFO stays the default and DAG is opt-in via `plugins.mempool.provider`, which bounds the blast radius considerably — that's the right call for a change of this shape.

## Process: DevNet

CLAUDE.md requires DevNet validation for changes touching the mempool, tx submission, or block production. This touches all three, and it changes the tx-submission protocol loop, not just mempool internals. The PR helpfully *adds* `DEVNET_MEMPOOL_PROVIDER` so DAG can be exercised there, but the description doesn't say a run happened — and a default-FIFO run wouldn't exercise any of the new code. I'd want a DevNet run with `DEVNET_MEMPOOL_PROVIDER=dag` before merge.

## What's good

The shared `runBackendContract(t, factory)` suite run against both backends is exactly the right structure for two implementations of one interface — it makes the behavioral contract explicit rather than hoping both happen to agree.

DAG-specific concurrency coverage is genuinely solid: `TestDAGAdmissionContinuesDuringRevalidation`, `TestDAGRemovalContinuesDuringRevalidation`, `TestDAGConcurrentSnapshotsAndMutations`, `TestDAGRevalidationRetriesAfterLedgerGenerationChange` and `TestDAGRevalidationBoundsLedgerGenerationRetries` cover the scenarios that matter, in the same spirit as #2986's FIFO tests.

`dingo_metrics_mempool_info{implementation=...}` means an operator can tell which backend is actually running, which is the first thing you want when triaging a report from the field. And `TestDAGDoesNotWatermarkEvict` pinning the non-evicting contract is a good guard, since "no eviction" is the property the whole backpressure design rests on.

chrisguiney (commented):
Re-review of `5f7d9020` (my earlier review was against `f98b3e02`). **All five findings from that review are resolved**, and the two blocking ones were fixed structurally rather than patched: `mempool/dag_revalidation.go` is deleted outright and the DAG now runs through the same revalidation path as FIFO.

| Earlier finding | Status |
|---|---|
| **Blocking:** DAG revalidated per-transaction instead of using `WithTxValidationSession` | **Fixed** — shared `rebuildOverlayAttempt` pins the session and uses `stillCurrent()` (`mempool.go:975`, `:1019`, `:1192-1196`) |
| **Blocking:** parallel revalidation architecture — two journals, two constant sets, two retry semantics | **Fixed** — `dagRevalidationDeltaCap`, `dagRevalidationMaxIters`, `dagRevalidationMaxGenerationRetries` and `dagMutation` are all gone |
| `revalidationDeltaCap` hardcoded at 256 for DAG | **Fixed** — DAG inherits the operator-configurable cap |
| No DAG journal overflow valve | **Fixed** — DAG inherits `revalidationJournalCap` / `journalOverflow` |
| ID-count drop to 1 undocumented, no upstream pointer | **Fixed** — `ARCHITECTURE.md` states the throughput cost and the reason; code carries a `TODO` |

I also checked two invariants the refactor could plausibly have broken, and both hold:

- **DAG still has a hard admission cap.** `mempool.go:1361-1369` is not gated on `m.dag == nil`, so disabling eviction doesn't leave the pool unbounded — it still returns `MempoolFullError` at the rejection watermark.
- **Headroom notification is complete.** Every size-decreasing path reaches `notifyHeadroomChangedLocked`: `Stop` (`:695`), the rebuild commit (`:1058`), `AddTransaction` (`:1433`), and `removeTransactionByIndexLocked` (`:1675`), which is the sole route for `RemoveTransaction`, `RemoveTxsByHash` and expiry. `evictOldestLocked`'s bulk subtraction is unreachable for DAG. The close-and-recreate pattern in `WaitForAdmissionHeadroom` also avoids the lost-wakeup race — it re-checks under `RLock`, captures the channel, then releases before selecting.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go test -race -count=1 ./mempool/... ./ouroboros/... ./internal/config/... .` | **all pass** |
| `modernize ./...` | no hits in `mempool/` or `ouroboros/` |
| Dead-index experiment + topological-order probe | two findings, both confirmed — see inline |

## Where this leaves the review

I'm **not** blocking on anything below — two are small cleanups and the rest are notes. What I'd still want before merge is a DevNet run, which is why I haven't converted my standing `CHANGES_REQUESTED` to an approval yet:

CLAUDE.md lists mempool **and** tx submission as DevNet-requiring, and this PR changes block-production ordering and network intake for both. The description is now entirely cubic/CodeRabbit-generated — the hand-written verification section from the earlier revision is gone — so there's no record of what was run. This PR ships the switch itself:

```
DEVNET_MEMPOOL_PROVIDER=dag ./run-tests.sh
```

Post that result (or say it already ran) and I'll clear the review.

## Documentation

`ARCHITECTURE.md` (+88) covers the `Service` contract, the provider split, DAG behaviour and the intake trade-off. `DATABASE.md` checked and not affected — nothing here touches models, migrations, or the metadata surface.

mempool/mempool.go
```diff
@@ -959,6 +1052,10 @@ func (m *Mempool) rebuildOverlayAttempt() ([]event.Event, error) {
 				m.transactions = candidate.transactions
 				m.txByHash = candidate.txByHash
 				m.currentSizeBytes = candidate.sizeBytes
+				if m.dag != nil {
```
chrisguiney, line 1055: This line is the resolution of both blocking findings from my earlier review, so it's worth marking explicitly.

The DAG is now rebuilt from `candidate.overlay.applied` inside the **shared** commit section, which means it inherits the whole FIFO revalidation contract rather than reimplementing it: the pinned `WithTxValidationSession` and its `stillCurrent()` guard, the single mutation journal with `revalidationJournalCap` / `journalOverflow`, the operator-configurable `revalidationDeltaCap`, and one retry bound. Deleting `dag_revalidation.go` outright (−320 lines) rather than reconciling two copies was the right call — the earlier version had two journals and two sets of constants guarding the same block-production path, and a bug fixed in one wouldn't have reached the other.

The `invalidUtxos` / `reject` / `dependsOnInvalid` mechanism that replaced the DAG's private descendant handling is a good fit for the shared candidate: it propagates rejection transitively provided parents are processed before children, and they are, because `overlay.applied` is admission-ordered and a child can't be admitted before its parent exists in the overlay. `TestDAGRevalidationSkipsInvalidDescendantValidation` pins it.

mempool/dag.go
```diff
@@ -0,0 +1,210 @@
...
+		hash:     tx.hash,
+		sequence: d.nextSequence,
+		parents:  make(map[string]struct{}),
+		children: make(map[string]struct{}),
+		consumed: slices.Clone(tx.consumed),
+		produced: make([]string, 0, len(tx.created)),
+	}
+	d.nextSequence++
+	for _, input := range tx.consumed {
+		d.spenderByUtxo[input] = tx.hash
```
chrisguiney: `spenderByUtxo` is write-only. Nothing in production reads it — `producerByUtxo` on the next line is the load-bearing index (it's how parents are found), and children are maintained directly as edges.

I verified rather than eyeballed it: I removed both this write and the delete block at `:163-167`, then ran the mempool suite. **Exactly one test failed** — `TestTransactionDAGTopologicalOrderAndDescendants`, at the assertion that reads the index directly (`dag_test.go:148`). No behavioural test noticed the index was gone.

So this is a map that reaches one entry per pending input — tens of thousands at a full mainnet mempool — maintained on every add and every remove, consulted by nothing.

Two ways to resolve it, and the second is more interesting than the first:

1. Delete it, along with the delete block at `:163-167` and the assertion at `dag_test.go:148`.
2. Give it the job it looks like it should have. A spender index is the natural place to detect an intra-mempool double-spend on admission — right now a second transaction consuming an already-pending input is caught only by overlay validation, and `:95` would silently overwrite the first spender if it ever weren't. If that's the intent, a lookup here plus a test for the conflicting-input case would earn the memory.

Either is fine; carrying it unused is the option I'd avoid, since the next reader has to work out whether the absence of a lookup is a bug.

mempool/dag_test.go
```diff
@@ -0,0 +1,615 @@
...
+		graph.topologicalOrder(),
+	)
+	assert.Equal(t, map[string]struct{}{
+		"parent":     {},
+		"left":       {},
+		"right":      {},
+		"grandchild": {},
+	}, graph.descendants(map[string]struct{}{"parent": {}}))
+	assert.Equal(t, "parent", graph.producerByUtxo["parent:0"])
+	assert.Equal(t, "left", graph.spenderByUtxo["parent:0"])
```
chrisguiney: This is the only reader of `spenderByUtxo` anywhere, production or test, which is why the index looks covered when it isn't — the assertion tests that the index contains what the index was told to contain.

The `producerByUtxo` assertion on the line above is different in kind: that index is genuinely consulted (`dag.go:96`) to derive parent edges, so asserting it is asserting something the graph depends on.

If `spenderByUtxo` gains a real consumer, this becomes a reasonable structural check. If it's deleted, this line goes with it.

mempool/dag.go
```diff
@@ -0,0 +1,210 @@
...
+		for childHash := range node.children {
+			indegree[childHash]--
+			if indegree[childHash] == 0 {
+				if child := d.nodes[childHash]; child != nil {
+					heap.Push(&ready, child)
+				}
+			}
+		}
+	}
+	return ret
```
chrisguiney: Kahn's algorithm drops any node that never reaches indegree zero, and this returns the short list without a word. Probed with a hand-built cyclic graph plus one healthy node:

```
nodes in dag = 3, topologicalOrder returned 1: [ok]
```

No error, no log, no metric. `Transactions()` (`mempool.go:1485`) is built from this, and `Transactions()` is what the forger builds a block from — so an index invariant break wouldn't surface as a failure, it would surface as quietly smaller blocks.

This is unreachable today and I don't think the code is wrong: a UTxO graph can't cycle, because inputs reference already-hashed transactions. The reason I'd still add the check is the failure mode rather than the likelihood — three lines converts a silent wrong answer into a diagnosable one:

```go
if len(ret) != len(d.nodes) {
    // caller logs: DAG index inconsistent, N of M transactions ordered
}
```

Same shape as the length/indegree invariants the rest of this file is careful about; it's just the one place where the wrong answer is representable.

mempool/mempool.go
```diff
@@ -1353,14 +1480,27 @@ func (m *Mempool) GetTransaction(txHash string) (MempoolTransaction, bool) {
 
 func (m *Mempool) Transactions() []MempoolTransaction {
 	m.RLock()
-	ret := make([]MempoolTransaction, len(m.transactions))
-	for i := range m.transactions {
-		ret[i] = *m.transactions[i]
+	var ret []MempoolTransaction
+	if m.dag != nil {
+		order := m.dag.topologicalOrder()
```
chrisguiney: This runs a full heap-driven topological sort while holding the read lock — and the comment a few lines below explains why that's awkward:

> Copy the slice headers under the state lock, then clone the bytes after releasing it so forging and relay snapshots do not hold up the final revalidation swap in proportion to total transaction bytes.

The CBOR copies were deliberately moved out of the lock for exactly that reason, and then an O(V+E) sort moved in. It also runs per caller, not once — every relay consumer's snapshot re-sorts the whole graph.

The order only changes when the graph changes, and every mutation already happens under the write lock, so computing it at mutation/commit time and caching it would keep the property the comment is protecting. Worth measuring before changing anything, but the asymmetry between the comment's intent and the added work is real.

mempool/mempool.go
```diff
@@ -1039,17 +1138,25 @@ func (m *Mempool) revalidateAppliedTx(
 	) error,
 ) {
 	if tx == nil {
+		candidate.reject(at, nil)
```
chrisguiney, line 1143: This is a FIFO behaviour change riding along in a DAG PR, on the block-production path.

Before, a transaction present in the overlay but missing from `txByHash` logged a warning and revalidation carried on, leaving its descendants in the candidate. Now `reject(at, nil)` marks its created UTxOs invalid, so `dependsOnInvalid` drops the whole dependent cone.

I think that's a fix — descendants of a transaction that isn't in the index can't be valid, and keeping them was the actual bug — but it takes effect for FIFO nodes, which is the default, and neither the description nor a comment mentions it. The `invalid` map is also deliberately not touched here (`tx == nil` so there's nothing to record), which is worth a word since every other `reject` call site passes a transaction.

A line in the description and a FIFO-side test for the missing-index case would keep this from being discovered later as an unexplained divergence.

ouroboros/txsubmission.go
```diff
@@ -307,12 +351,34 @@ func (o *Ouroboros) txsubmissionServerInit(
...
+						err = o.Mempool.AddTransaction(
+							uint(txBody.EraId),
+							txBody.TxBody,
+						)
+						if err == nil {
+							break
+						}
+						var fullErr *mempool.MempoolFullError
+						if limitAdmission && errors.As(err, &fullErr) {
+							admissionRetryStreak++
```
chrisguiney: The retry streak has no upper bound. `MempoolFullError` sends the loop back to `WaitForAdmissionHeadroom` indefinitely, and the only exits are a connection error or mempool stop.

It isn't a busy-spin — each iteration blocks in the wait — and the new `dingo_txsubmission_admission_retry_streak` metric makes it observable, which is the right instinct. What it never does is give up on this one transaction and move to the next offer. Two peers racing for the same headroom can each keep waking on the other's removal, and a persistently full mempool means this connection makes no forward progress while other peers do.

Worth a deliberate decision either way: a streak cap that logs and drops the transaction (the peer re-offers it later anyway, since the ID is acknowledged on the next request) bounds the worst case, and the metric is already there to tune it against.

mempool/mempool.go
```diff
@@ -1383,6 +1523,28 @@ func (m *Mempool) CapacityBytes() int64 {
 	return m.config.MempoolCapacity
 }
 
+func (m *Mempool) admissionLimitBytes() int64 {
```
chrisguiney, line 1545: Small one: this recomputes the expression that `AddTransaction` already inlines at `:1361-1363` (`float64(m.config.MempoolCapacity) * m.rejectionWatermark`). Having the admission threshold defined in two places is the kind of thing that drifts when the watermark semantics change — the inline copy should call this.

While in the area: `slices.Sort(node.produced)` at `dag.go:104` has no order-sensitive consumer. `produced` is only iterated in `remove()`, where order doesn't matter. If it's there for reproducible debug output, a short comment saying so would keep it from looking load-bearing.

chrisguiney (approved):

Issue #2976, opened 2026-07-27
Leios EB apply wedges Dingo in infinite pipeline-restart loop: `pool cert reward_account: got 2 bytes, want 29`


chadaibbhot-git:
## Environment
- musashi (Leios prototype), Dingo v0.67.1 in `--run-mode leios --start-era dijkstra`
- Two independent Dingo nodes on separate hosts, both stuck the same way (see below)
- Haskell reference `cardano-node 11.1.0.164` (`prototype-2026w30`) at tip on the same chain (slot 2323241 as I write, ~100 500 slots past our stuck point) — the block is well-formed at the network level

## Symptom

2026-07-26T17:26:27Z: Dingo chain-extended to slot **2222761** (RB `d9fb4d5228c668460df3fe44ab63c2d9d1a12523bcdf0c5f327147ac721ec5e5`). 9 s later the EB manifest for that slot arrived (`63d3fb1e48d86068731e83a534fcfc25dbe3de135404aeaa916ee89a73be0d42`, 78 520 bytes, 2 181 txs). Applying its metadata effects (per #2787) hit:

```
level=WARN msg="block processing failed, restarting pipeline"
  error="process block batch: record transaction: set transaction metadata:
         pool reward account: pool cert reward_account: got 2 bytes, want 29"
```

Container reports `Up (healthy)`; no crash, no exit. The pipeline restarts, hits the same block, fails identically, restarts again — every ~3 seconds. Ledger stays frozen at slot 2222761.

**Two-node cross-reproduction**: 28+ hours in as of writing on our first node (~32 000 hits). A second, independently-configured Dingo node on separate hardware syncing the same chain is also stuck at the same block with the same error (~48 000 hits). A container recreate (fresh process, same disk) does not clear it — the failing block persists in the blob store.

## Location

`database/plugin/metadata/internal/certutil/pool.go:66-69` (v0.67.1):

```go
if len(rewardAddrBytes) != 29 {
    return 0, nil, fmt.Errorf("pool cert reward_account: got %d bytes, want 29", len(rewardAddrBytes))
}
```

[...]

PR #2982, opened 2026-07-28
fix(database): include tx hash + block index in 'set transaction metadata' errors

chadaibbhot-git:
## Motivation

The `record transaction → set transaction metadata` wrap currently loses per-transaction context. When a single tx in a batch causes the database write to fail (for example the `pool cert reward_account: got 2 bytes, want 29` failure in #2976, which wedged Leios nodes on musashi at slot 2222761 for 28+ hours before we could file), the pipeline-restart `WARN` in `handleLedgerProcessBlocksError` (`ledger/state.go:3078`) surfaces the error text but nothing identifies *which* transaction of the block caused it. Operators can see the failing block hash (via chain-extend logs) but have to open the block, walk its transactions, and re-run the same parser against each to isolate the offender — for a 2 181-tx Leios EB batch that's real friction.

This change wraps the three matching sites with the tx hash, the block/batch index, and the containing slot. The change is error text only; no behavior change.

## Change

Three sites in `database/batch.go` and `database/transaction.go` currently wrap:

```go
return fmt.Errorf("set transaction metadata: %w", err)
```

With this PR they wrap:

```go
return fmt.Errorf(
    "set transaction metadata for tx %s (batch idx %d, slot %d): %w",
    tx.Hash(), idx, point.Slot, err,
)
```

Matches the pattern already used in `ledger/state.go:6949`:

```go
return fmt.Errorf("TX %s failed validation: %w", tx.Hash(), err)
```

## Operator effect

Before:
```
level=WARN msg="block processing failed, restarting pipeline"
  error="process block batch: record transaction: set transaction metadata:
         pool reward account: pool cert reward_account: got 2 bytes, want 29"
```

After:
```
level=WARN msg="block processing failed, restarting pipeline"
  error="process block batch: record transaction:
         set transaction metadata for tx 8c72…3fbe (batch idx 1247, slot 2222761):
         pool reward account: pool cert reward_account: got 2 bytes, want 29"
```

The failing tx hash + its position in the batch + the containing slot are enough to grep the block store and locate the offending cert directly. For #2976, that's the difference between "sync your Leios node past this slot and hit the wedge" and "here is the exact malformed cert" as a reproducer.

## What this does not change

- The pipeline-restart control flow: the WARN still fires, the batch is still retried indefinitely.
- The parser semantics: `PoolRewardAccount()` at `certutil/pool.go:66` still rejects the malformed cert with the same underlying error.
- The behavior on the happy path: no wrap is added when the write succeeds.

A separate change worth considering (not in this PR) is to quarantine a persistently-failing batch after N retries rather than looping forever, so a single malformed on-chain tx cannot wedge a node indefinitely. Happy to open that as its own issue if useful.

## Test plan

- [x] `docker build .` compiles cleanly against current `main`.
- [x] `go test -run TestSetTransactionMetadataErrorWrapFormat -v ./database/` — passes. New test at `database/tx_context_error_wrap_test.go` pins the three wrap sites: verifies `errors.Is` unwrap chain preservation, tx hash rendering via `%s` on `Blake2b256`, `batch idx` vs `block idx` label variants, slot decimal formatting, and inner-error preservation. Any format drift fails the test loudly.
- [x] Binary verification: `strings $(docker create ):/bin/dingo | grep "set transaction metadata"` shows all three new format strings compiled in; bare `set transaction metadata: %w` and `set transaction metadata only: %w` occurrences → 0 (proves the swap happened).
- [x] **Live test** on a Leios testnet (musashi) relay still hitting the #2976 malformed cert (branch = `main~1` before #2977 + this commit, so parser still fails and the WARN still fires). Result: 24 WARN lines captured in 60 s, all deterministically identifying the same failing tx and its position:
  ```
  set transaction metadata for tx 74422e5b4a93939700954b03af4de1c930b192c8f1a749765d181bde515d04ab
                                  (block idx 200, slot 2222845):
                                  pool reward account: pool cert reward_account: got 2 bytes, want 29
  ```
  Container health unchanged (`restarts=0`, `status=running`). No format regressions. Node restored to stock `0.67.1` after test.

## Related

- #2976 (closed by #2977) — the wedge that motivated this change; the WARN's lack of context added ~80 minutes to the investigation.
- `ledger/state.go:3066-3081 handleLedgerProcessBlocksError` — the consumer that emits the WARN.
- `ledger/state.go:6949` — existing precedent for `%s` on `tx.Hash()` in error messages.

Small, error-text-only, single-purpose. Happy to add tests or split into smaller PRs on request.

---
## Summary by cubic
Improved database error messages for "set transaction metadata" to include the tx hash, block/batch index, and slot so operators can pinpoint the failing transaction. Text-only change; no behavior change.

- **Bug Fixes**
  - Wrapped errors in `database/batch.go` and `database/transaction.go` with context: `"set transaction metadata for tx %s (batch|block idx %d, slot %d): %w"` and `"set transaction metadata only for tx %s (block idx %d, slot %d): %w"`.
  - Pipeline WARNs now include per-tx context (helps diagnose cases like #2976).
  - Added `database/tx_context_error_wrap_test.go` to pin wording (`batch idx`/`block idx`), tx hash, slot, and preserve `errors.Is` behavior.

<sup>Written for commit f9dfdd817b2433546bcb49b543c0b651c8cd1638. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2982?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction metadata error messages with the transaction hash, batch or block index, and slot.
  * Preserved the underlying error details to make failures easier to diagnose.

* **Tests**
  * Added coverage to verify error context and underlying error preservation across transaction metadata workflows.

### DIFF
--- a/database/batch.go
+++ b/database/batch.go
@@ -263,7 +263,10 @@ func (d *Database) SetTransactionBatchedWithOpts(
 	if err := d.metadata.SetTransactionBatched(
 		tx, point, idx, certDeposits, acc, metadataTxn,
 	); err != nil {
-		return fmt.Errorf("set transaction metadata: %w", err)
+		return fmt.Errorf(
+			"set transaction metadata for tx %s (batch idx %d, slot %d): %w",
+			tx.Hash(), idx, point.Slot, err,
+		)
 	}
 
 	if updateEpoch > 0 && tx.IsValid() {
--- a/database/transaction.go
+++ b/database/transaction.go
@@ -258,7 +258,10 @@ func (d *Database) SetTransactionWithOpts(
 		return err
 	}
 	if err := d.metadata.SetTransaction(tx, point, idx, certDeposits, txn.Metadata()); err != nil {
-		return fmt.Errorf("set transaction metadata: %w", err)
+		return fmt.Errorf(
+			"set transaction metadata for tx %s (block idx %d, slot %d): %w",
+			tx.Hash(), idx, point.Slot, err,
+		)
 	}
 
 	if updateEpoch > 0 && tx.IsValid() {
@@ -311,7 +314,10 @@ func (d *Database) SetTransactionMetadataOnly(
 		certDeposits,
 		metadataTxn,
 	); err != nil {
-		return fmt.Errorf("set transaction metadata only: %w", err)
+		return fmt.Errorf(
+			"set transaction metadata only for tx %s (block idx %d, slot %d): %w",
+			tx.Hash(), idx, point.Slot, err,
+		)
 	}
 	if owned {
 		if err := txn.Commit(); err != nil {
--- a/database/tx_context_error_wrap_test.go
+++ b/database/tx_context_error_wrap_test.go
@@ -1,0 +1,226 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package database
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/blinklabs-io/dingo/database/plugin/metadata"
+	"github.com/blinklabs-io/dingo/database/types"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
+	"github.com/stretchr/testify/require"
+)
+
+// erroringMetadata wraps a real MetadataStore and returns injectErr from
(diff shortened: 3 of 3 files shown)


wolf31o2:
@chadaibbhot-git fix the review from coderabbit

chadaibbhot-git:
Fixed — the test now drives the three production entry points (`SetTransactionBatched`, `SetTransaction`, `SetTransactionMetadataOnly`) end-to-end and forces the metadata write to fail via a stub wrapping the real `MetadataStore`. Everything upstream of the metadata call runs against real code as it does at runtime; only the terminal metadata write returns a known inner error. If a production wrap loses the tx hash, idx label, slot, inner-error text, or `errors.Is` chain — or the wording drifts — the test fails loudly.

Commit f9dfdd81, `database/tx_context_error_wrap_test.go`. Passes locally:

```
=== RUN   TestSetTransactionMetadataErrorWrap_ProductionPaths
--- PASS: TestSetTransactionMetadataErrorWrap_ProductionPaths (2.48s)
```

wolf31o2 (approved):

PR #2990, opened 2026-07-28
fix(bark): verify archive block bytes against the requested point

### LINKED ISSUE #2038: security: verify Bark archive blocks against requested hash
## Finding

ID: DSA-2026-04-24-02
Severity: HIGH
Audit: docs/plans/2026-04-24-security-audit.md

## Problem

Bark archive fallback requests a specific slot/hash, but returns the downloaded block body without verifying that the bytes decode to the requested block hash. Metadata returned by Bark is also trusted when constructing block metadata.

Evidence:
- bark/blob.go:150-158 sends the requested slot/hash to the archive service.
- bark/blob.go:189-210 reads and returns the block body and Bark metadata without local hash verification.

## Acceptance criteria

- Returned block bytes are decoded and canonical block hash is computed locally.
- The decoded block slot/hash must match the requested slot and hash before the bytes are returned.
- Bark-supplied metadata, especially height and previous hash, is validated against decoded block contents where available.
- Regression tests cover an archive response that returns the wrong block for a valid requested point.

## Notes

Bark can remain a storage delegate, but consensus-relevant identity must be checked locally before historical block data is trusted.

chrisguiney:
Closes #2038 (DSA-2026-04-24-02, HIGH).

## Problem

The Bark archive fallback requested a specific `(slot, hash)` but returned the downloaded body without checking it was that block. `GetBlock` then explicitly commented that it trusted the archive for `Type`/`Height`/`PrevHash`, taken straight off `block.GetMeta()`.

A misbehaving or hostile archive controls both the download URL and the response body, so it could substitute any block — or misreport height and previous hash — and the bytes would reach callers as historical chain data. The existing URL/redirect/size checks constrain *where* bytes come from, but nothing established *what* they were.

## Change

Verification lives in `fetchBlockFromArchive`, so `GetBlock` and the iterator's expired-history resolution (`ValueCopy`) are both covered by one path rather than one being an unchecked route into archive data.

- Decode the downloaded body and require the computed block hash **and** slot to match the requested point.
- The archive's block type is now only a decode hint: a wrong type either fails to decode or yields a different hash, and both are rejected, so it cannot be used to smuggle in substitute bytes.
- Decoding runs with body-hash validation enabled, so a genuine header with a swapped body is caught too. This matches how blocks arriving over the network are already treated (`ouroboros/blockfetch.go`).
- Block metadata is derived from the decoded block. Archive-reported height or previous hash that contradicts it fails the fetch rather than being silently overridden; zero-valued archive fields are treated as absent, not as a conflict.

The archive remains a storage delegate — trusted to hold block bytes, not to identify them.

## Tests

Written first, and each was confirmed to fail against pre-fix code by reverting the implementation and re-running:

| Test | Covers |
|---|---|
| `TestGetBlock_RejectsArchiveBlockForDifferentPoint` | the finding — a valid but wrong block substituted for a valid requested point |
| `TestBarkIterator_RejectsArchiveBlockForDifferentPoint` | same, via the iterator path |
| `TestGetBlock_RejectsUndecodableArchiveBlock` | body that is not a block |
| `TestGetBlock_RejectsArchiveBlockSlotMismatch` | genuine block, wrong slot |
| `TestGetBlock_RejectsArchiveMetadataMismatch` | height / previous-hash disagreement |
| `TestGetBlock_AcceptsVerifiedArchiveBlock` | happy path, including an "archive omits metadata" case proving metadata is derived from the decoded block |

Fixtures come from `ouroboros-mock`'s `fixtures.GenerateConwayChain` rather than local mocks, per the repo's fixture policy — real blocks whose CBOR round-trips with consistent `Hash()`/`SlotNumber()`/`BlockNumber()`/`PrevHash()`, which is what the verification actually checks. `TestBarkIterator_ResolvesExpiredHistoryViaArchive` previously used fabricated CBOR at a made-up hash and now uses a real block; it would otherwise have failed correctly under the new check.

## Verification

- `go test -race ./bark/... ./database/...` — pass
- `internal/test/conformance` — pass
- `golangci-lint run ./bark/...` (0 issues), `modernize`, `make import-boundaries` — clean

## Documentation

Both updated — each already enumerated Bark's archive validation rules and was stale without the identity check:

- `DATABASE.md` — history-expiry / Bark fallback section
- `ARCHITECTURE.md` — `### Bark (bark/)`

DevNet was assessed and not run: this is the archive fallback read path for historical blocks, not consensus, block production, chain selection, or the protocol surfaces that trigger a DevNet run.

---
## Summary by cubic
Hardened `bark` archive fallback by locally verifying block bytes against the requested hash and slot, deriving metadata from the decoded block, and refusing Byron main archive reads where the body cannot be fully authenticated. Both `GetBlock` and iterator fallbacks use the same verified path, with sentinel errors to distinguish verification failures.

- **Bug Fixes**
  - Decode with validation and require hash+slot match; reject undecodable, mismatched-hash, or wrong-slot blocks.
  - Derive the era from the header via `ledger.DetermineBlockType`; require it matches the archive claim; reject unclassifiable eras. Byron hashes already bind era.
  - Refuse Byron main archive reads until SSC proofs are verified upstream; Byron EBBs are allowed.
  - Build `types.BlockMetadata` (type/height/prev-hash) from the decoded block; reject archive-reported conflicts; treat zero values as absent.
  - Export sentinel errors for archive verification failures.

- **Dependencies**
  - Bump `github.com/blinklabs-io/gouroboros` to v0.190.0 (includes Byron body proof validation used by archive verification).
  - Propagate error returns from `CertifiedNatThresholdWithMode` and `IsVRFOutputBelowThresholdWithMode` in `ledger` schedule and header verification.

<sup>Written for commit eec3a7d8792238308061139d7bab83ba2723eb40. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2990?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security & Reliability**
  * Archive-sourced blocks are now cryptographically verified before being returned.
  * Block identity, slot, height, and previous-hash metadata are cross-checked for consistency.
  * Invalid, altered, undecodable, or mismatched archive responses are rejected.
  * Expired-history lookups use the same verification safeguards as direct block retrieval.

* **Documentation**
  * Updated architecture and database documentation to describe archive validation and trust boundaries.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2306,11 +2306,19 @@ local blocks expire; `internal/historyexpiry.Pruner` owns that lifecycle when
 Because the archive controls both the download URL and the response body, the
 client re-establishes block identity locally rather than trusting the response.
 Downloaded bytes are decoded with body-hash validation enabled, and the block's
-computed hash and slot must match the point that was requested; the archive's
-own block type is treated as a decode hint, since a wrong type either fails to
-decode or produces a different hash. Block metadata is derived from the decoded
-block, and archive-reported height or previous hash that disagrees with it fails
-the fetch. Both entry points — `GetBlock` and the iterator's expired-history
+computed hash and slot must match the point that was requested.
+
+The block era needs its own check. For Shelley and later the hash covers the
+header alone, and adjacent eras share that header layout, so one set of bytes
+decodes under several eras with an identical hash and slot — the hash cannot
+police the era. The era is therefore derived from the header itself via
+`ledger.DetermineBlockType` and must equal what the archive claimed. Byron is
+exempt because its hash is taken over the block type byte followed by the
+header, so the hash already binds it. An era that cannot be derived at all is
+refused rather than taken on the archive's word.
+
+Block metadata is derived from the decoded block, and archive-reported height
+or previous hash that disagrees with it fails the fetch. Both entry points — `GetBlock` and the iterator's expired-history
 resolution — share this path, so neither is an unchecked route into archive
 data.
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2303,6 +2303,17 @@ remote Bark archive and downloading the signed URL. Bark does not decide which
 local blocks expire; `internal/historyexpiry.Pruner` owns that lifecycle when
 `historyExpiry.enabled` is configured.
 
+Because the archive controls both the download URL and the response body, the
+client re-establishes block identity locally rather than trusting the response.
+Downloaded bytes are decoded with body-hash validation enabled, and the block's
+computed hash and slot must match the point that was requested; the archive's
+own block type is treated as a decode hint, since a wrong type either fails to
+decode or produces a different hash. Block metadata is derived from the decoded
+block, and archive-reported height or previous hash that disagrees with it fails
+the fetch. Both entry points — `GetBlock` and the iterator's expired-history
+resolution — share this path, so neither is an unchecked route into archive
+data.
+
 ### Midnight Indexer (`midnight/indexer/`)
 
 An optional block scanner that indexes Midnight chain events into multiple
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2305,8 +2305,19 @@ local blocks expire; `internal/historyexpiry.Pruner` owns that lifecycle when
 
 Because the archive controls both the download URL and the response body, the
 client re-establishes block identity locally rather than trusting the response.
-Downloaded bytes are decoded with body-hash validation enabled, and the block's
+Downloaded bytes are decoded with body validation enabled, and the block's
 computed hash and slot must match the point that was requested.
+
+That binding is complete for every era except Byron main. gouroboros checks a
+Byron main block's transaction, delegation, and update proofs but not
+`ssc_proof`, whose hashes cover cardano-ledger's own encoding of the SSC
+sub-payloads rather than the bytes carried in the block. An alteration confined
+to the SSC payload would therefore leave the hash, slot, height, and previous
+hash — all taken from the untouched header — unchanged. Rather than serve a
+body that is only partly authenticated, Bark refuses Byron main archive reads
+outright; the restriction lifts once Byron SSC proof validation exists
+upstream. Byron epoch boundary blocks are unaffected, carrying neither
+transactions nor an SSC payload, so one body hash covers their whole body.
 
 The block era needs its own check. For Shelley and later the hash covers the
 header alone, and adjacent eras share that header layout, so one set of bytes
--- a/bark/blob.go
+++ b/bark/blob.go
@@ -420,6 +420,33 @@ func (b *BlobStoreBark) fetchBlockFromArchive(
 	return blockBody, meta, nil
 }
 
+// Errors reported when an archive response fails local verification. They are
+// distinct from transport failures on purpose: a transport error is worth
+// retrying, whereas these mean the archive served something that is not the
+// block that was asked for, and its answers cannot be trusted as chain data.
+var (
+	// ErrArchiveBlockUndecodable reports an archive response body that does
+	// not decode as a block of the type the archive claimed.
+	ErrArchiveBlockUndecodable = errors.New(
+		"bark: archive block could not be decoded",
+	)
+	// ErrArchiveBlockHashMismatch reports a decoded block whose computed
+	// hash is not the hash that was requested.
+	ErrArchiveBlockHashMismatch = errors.New(
+		"bark: archive block hash does not match the requested hash",
+	)
+	// ErrArchiveBlockSlotMismatch reports a decoded block that does not sit
+	// at the slot that was requested.
+	ErrArchiveBlockSlotMismatch = errors.New(
+		"bark: archive block slot does not match the requested slot",
+	)
+	// ErrArchiveMetadataMismatch reports archive-supplied metadata that
+	// contradicts the contents of the verified block it accompanied.
+	ErrArchiveMetadataMismatch = errors.New(
+		"bark: archive metadata contradicts the block",
+	)
+)
+
 // verifyArchiveBlock establishes locally that the bytes the archive returned
 // really are the block that was requested. Bark chooses both the download URL
 // and the response body, so it can only be trusted to store blocks, not to
@@ -440,19 +467,21 @@ func verifyArchiveBlock(
 	decoded, err := gledger.NewBlockFromCbor(blockType, body)
 	if err != nil {
 		return nil, fmt.Errorf(
-			"bark: decoding archive block for slot %d: %w", slot, err,
+			"%w: slot %d, type %d: %w",
+			ErrArchiveBlockUndecodable, slot, blockType, err,
 		)
 	}
 	decodedHash := decoded.Hash()
 	if !bytes.Equal(decodedHash[:], hash) {
 		return nil, fmt.Errorf(
-			"bark: archive returned block with hash %x for requested hash %x",
-			decodedHash[:], hash,
+			"%w: got %x, requested %x",
+			ErrArchiveBlockHashMismatch, decodedHash[:], hash,
 		)
 	}
 	if decoded.SlotNumber() != slot {
 		return nil, fmt.Errorf(
-			"bark: archive block %x is at slot %d, requested slot %d",
+			"%w: block %x is at slot %d, requested slot %d",
+			ErrArchiveBlockSlotMismatch,
 			decodedHash[:], decoded.SlotNumber(), slot,
 		)
 	}
@@ -475,15 +504,15 @@ func archiveBlockMetadata(
 	height := decoded.BlockNumber()
 	if archiveHeight != 0 && archiveHeight != height {
 		return types.BlockMetadata{}, fmt.Errorf(
-			"bark: archive reported height %d for block at height %d",
-			archiveHeight, height,
+			"%w: reported height %d, block height %d",
+			ErrArchiveMetadataMismatch, archiveHeight, height,
 		)
 	}
 	prevHash := decoded.PrevHash()
 	if len(archivePrevHash) > 0 && !bytes.Equal(archivePrevHash, prevHash[:]) {
 		return types.BlockMetadata{}, fmt.Errorf(
-			"bark: archive reported previous hash %x for block whose previous hash is %x",
-			archivePrevHash, prevHash[:],
+			"%w: reported previous hash %x, block previous hash %x",
+			ErrArchiveMetadataMismatch, archivePrevHash, prevHash[:],
 		)
 	}
 	blockType := decoded.Type()
--- a/bark/blob.go
+++ b/bark/blob.go
@@ -15,6 +15,7 @@
 package bark
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
 	"errors"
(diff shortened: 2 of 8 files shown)


chrisguiney:
> `"hash"` also appears in unrelated archive failures (e.g. `failed decoding previous hash`), so this can pass while the block was rejected for the wrong reason.

Good catch — fixed in a6f542fe, and it improved the production code rather than just the test.

Each verification failure is now wrapped in an exported sentinel (`ErrArchiveBlockUndecodable`, `ErrArchiveBlockHashMismatch`, `ErrArchiveBlockSlotMismatch`, `ErrArchiveMetadataMismatch`) and the assertions use `errors.Is`. The distinction is worth exposing beyond the tests: a transport error is worth retrying, whereas these mean the archive served something that is not the requested block, so its answers cannot be trusted as chain data — and callers previously had no way to tell those apart.

I verified the sentinels actually discriminate rather than merely compiling: swapping the hash branch to return the slot sentinel makes both hash-mismatch tests fail with "Target error should be in err chain", and they pass again once reverted.

The metadata test keeps a substring check alongside the sentinel, purely to distinguish which field disagreed; the sentinel pins the code path so the earlier collision can't recur.

bark/blob.go
```diff
@@ -401,10 +404,127 @@ func (b *BlobStoreBark) fetchBlockFromArchive(
...
+// both outcomes are rejected below, so it cannot be used to smuggle in
+// substitute bytes. Decoding runs with validation enabled, so a block whose
+// header is genuine but whose body was swapped fails the body-hash check.
+func verifyArchiveBlock(
+	blockType uint,
+	body []byte,
+	slot uint64,
+	hash []byte,
+) (gledger.Block, error) {
+	decoded, err := gledger.NewBlockFromCbor(blockType, body)
```
wolf31o2, line 555: [P1] Verify Byron body proofs before accepting archive bytes

`NewBlockFromCbor` validates body hashes for Shelley and later, but gouroboros v0.189.4's Byron main and EBB constructors only decode and ignore the verification configuration. I reproduced this with ouroboros-mock's real `Block_Byron_regular` fixture: replacing its one-transaction payload with an empty transaction payload changed the bytes from 860 to 623 while decoding still succeeded and the hash and slot remained identical. This function therefore accepts the altered block. Because archive history includes Byron, a hostile archive can still return the requested header with a substituted body. Please add explicit Byron main/EBB body-proof validation (or reject unsupported Byron archive reads) and tamper regression tests.

bark/blob.go
```diff
@@ -401,10 +404,127 @@ func (b *BlobStoreBark) fetchBlockFromArchive(
...
+		)
+	}
+	prevHash := decoded.PrevHash()
+	if len(archivePrevHash) > 0 && !bytes.Equal(archivePrevHash, prevHash[:]) {
+		return types.BlockMetadata{}, fmt.Errorf(
+			"%w: reported previous hash %x, block previous hash %x",
+			ErrArchiveMetadataMismatch, archivePrevHash, prevHash[:],
+		)
+	}
+	blockType := decoded.Type()
```
wolf31o2: [P1] Derive the block type independently of the archive hint

`decoded.Type()` only reports which constructor the archive-selected `blockType` chose; it does not establish the block's actual era. With real ouroboros-mock fixtures, Shelley bytes decode successfully as Allegra or Mary, Allegra bytes decode as Mary, and Babbage bytes decode as Conway, all with the same hash and slot. The archive can therefore still dictate `BlockMetadata.Type`. Please derive the type independently from the decoded header/protocol version (for example via `ledger.DetermineBlockType(decoded.Header().Cbor())`, with explicit Byron handling), require it to match the hint, and add cross-era rejection tests.

wolf31o2 (changes requested):
Requesting changes: archive identity verification is still incomplete for Byron body proofs, and the returned block type remains archive-controlled when cross-era decoding succeeds. Both issues contradict the security objective and the updated documentation, so they are merge-blocking. Focused race tests, conformance, lint, and import-boundary checks otherwise pass.

bark/blob.go
chrisguiney: Fixed upstream rather than in dingo: blinklabs-io/gouroboros#1917, released as `v0.190.0`, and picked up here in a2f2d989.

Byron main and EBB blocks now validate their body proof at decode — transaction count, merkle root over transaction bodies, hash over the witness list, and the delegation and update payload hashes — matching how Shelley and later already behave, and skippable via `SkipBodyHashValidation` the same way. The bundled mainnet fixture carries two transactions, so the merkle branch path is exercised rather than only the single-leaf case, and a tamper test covers exactly the substitution you described.

One documented gap: `ssc_proof` is not recomputed. Its hashes are taken over cardano-ledger's encoding of the SSC sub-payloads rather than the bytes carried inline in the block — for the mainnet fixture the payload part hashes to `25777aca…` while the header records `d36a2619…`. The only available fixture has an empty SSC payload, so anything I inferred from it would be validated solely against the degenerate case. It is recorded as a `NOTE:` at the call site and in gouroboros' ARCHITECTURE.md. SSC carries shared-seed material, not transactions, so transaction contents are fully covered.

bark/blob.go
chrisguiney: Confirmed and fixed in 9f608b6d.

I verified your finding independently before changing anything — running the ouroboros-mock golden fixtures through every era constructor reproduces it exactly:

| Fixture (true era) | Also decodes as | Hash |
|---|---|---|
| `Block_Shelley` (2) | Allegra (3), Mary (4) | identical `b0b0c220…` |
| `Block_Allegra` (3) | Mary (4) | identical `b24221bc…` |
| `Block_Babbage` (6) | Conway (7) | identical `550d2623…` |

Same hash, same slot, different `Type()`. My earlier reasoning that the check was tautological was wrong in a way that mattered: the hash covers the *header*, which is byte-identical across adjacent eras, so it never bound the era at all.

The era is now derived from the header via `DetermineBlockType` and must equal the archive's claim. Byron is exempt, and deliberately so — `ByronMainBlockHeader.Hash()` and the EBB equivalent prepend `[0x82, <block type>]` before hashing, so the era is already bound by the existing hash check there.

Where derivation fails outright, the fetch is refused rather than falling back to the archive's claim, since falling back would hand era selection straight back to it. `TestGetBlock_RejectsUnclassifiableEra` pins that.

Cross-era rejection tests cover babbage-served-as-conway and shelley-served-as-mary, each asserting the decode genuinely succeeds under the false era first so the test cannot pass for the wrong reason. ARCHITECTURE.md's claim that a wrong type would fail to decode or change the hash was false and is corrected.

bark/blob.go
```diff
@@ -401,10 +404,175 @@ func (b *BlobStoreBark) fetchBlockFromArchive(
...
+// both outcomes are rejected below, so it cannot be used to smuggle in
+// substitute bytes. Decoding runs with validation enabled, so a block whose
+// header is genuine but whose body was swapped fails the body-hash check.
+func verifyArchiveBlock(
+	blockType uint,
+	body []byte,
+	slot uint64,
+	hash []byte,
+) (gledger.Block, error) {
+	decoded, err := gledger.NewBlockFromCbor(blockType, body)
```
wolf31o2, line 555: [P1] Byron SSC payload remains unauthenticated

The v0.190.0 Byron main validation used by this constructor explicitly skips `ssc_proof` and documents that an alteration confined to `SscPayload` is not detected. Because hash, slot, height, and previous hash come from the unchanged header, those altered body bytes can satisfy every check here. That still lets the archive substitute historical chain data. Please implement SSC proof validation or reject Byron main archive fallback until it is supported, and add a regression test.

ARCHITECTURE.md
```diff
@@ -2303,6 +2303,25 @@ remote Bark archive and downloading the signed URL. Bark does not decide which
 local blocks expire; `internal/historyexpiry.Pruner` owns that lifecycle when
 `historyExpiry.enabled` is configured.
 
+Because the archive controls both the download URL and the response body, the
+client re-establishes block identity locally rather than trusting the response.
```
wolf31o2, line 2307: [P1] Do not claim block identity is re-established while Byron SSC remains unchecked

The implementation does not fully authenticate Byron main block bodies: gouroboros v0.190.0 deliberately skips `ssc_proof`. This statement therefore overstates the trust boundary. It becomes accurate after SSC validation is implemented; otherwise the documentation and implementation need to enforce a fail-closed Byron limitation.

wolf31o2 (changes requested):
Requesting changes: Byron archive identity verification remains incomplete. gouroboros v0.190.0 explicitly skips ssc_proof validation, so an SSC-only body alteration is not detected despite matching header hash, slot, and metadata. This contradicts the security objective and updated documentation. Please validate the SSC proof or fail closed for Byron main archive reads, with regression coverage. DevNet also remains required because this PR touches header and VRF verification.

bark/blob.go
chrisguiney: Agreed, and fixed in eec3a7d8 by taking your second option — Byron main archive reads are now refused with `ErrArchiveBlockNotFullyAuthenticated`.

I tried the first option before falling back to this, and got further than expected: **the SSC proof encoding is `blake2b256` over cardano-ledger's map encoding, not the bytes carried inline.** For the mainnet fixture, `blake2b256(a0)` — the empty CBOR *map* — is exactly the `d36a2619…` in the header, while the inline payload carries the certificates as a tag-258 *set*, which is why hashing the payload part gave `25777aca…`. Cross-checked against the same header's `dlg_proof`, which is `blake2b256(9fff)`.

What stopped me short of implementing it upstream: gouroboros validates inside `NewBlockFromCbor` with validation on by default, and blocks with a **non-empty** SSC payload still need Byron `StakeholderId` derivation to rebuild the map — with no fixture carrying a non-empty SSC payload to validate that against. Failing closed there would refuse legitimate mainnet Byron history and break sync, so the restriction belongs here in bark where the trust decision is, exactly as you suggested.

**Epoch boundary blocks remain available.** They carry neither transactions nor an SSC payload, so a single body hash covers the whole body and the binding is complete — refusing them would give up history for no gain. `TestGetBlock_RejectsByronMainArchiveBlock` and `TestGetBlock_AcceptsByronEpochBoundaryArchiveBlock` cover both sides using the real `Block_Byron_regular` and `Block_Byron_EBB` fixtures.

The encoding finding is written up so upstream SSC validation is tractable as a follow-up; the restriction lifts when it lands.

ARCHITECTURE.md
chrisguiney: You're right — the statement claimed a binding that did not hold for Byron main. Corrected in eec3a7d8.

The section now says the binding is complete for every era *except* Byron main, names `ssc_proof` as the unchecked part and why (its hashes cover cardano-ledger's encoding of the SSC sub-payloads, not the bytes in the block), states that an SSC-only alteration would leave hash, slot, height, and previous hash untouched, and records that Bark refuses those reads rather than serving a partly authenticated body. It also notes the restriction lifts once upstream SSC validation exists, and that epoch boundary blocks are unaffected.

So the documentation and the implementation now agree, via the fail-closed route rather than by weakening the claim alone.

wolf31o2 (approved):

PR #3006, opened 2026-07-29
fix(consensus): enforce header forecast horizon

### LINKED ISSUE #2611: consensus: enforce the 3k/f forecast horizon on the live header-validation path
The Cardano Blueprint chain selection and multi-era specs describe a forecast range. Headers within 3k/f slots (129600 on mainnet) of an intersection can be validated using the ledger and chain state at that intersection, and headers beyond the forecast horizon cannot be validated against that state. The hard fork combinator bounds this horizon at era boundaries (the Haskell PastHorizon behavior). See https://cardano-scaling.github.io/cardano-blueprint/pr-preview/pr-67/consensus/chainsel.html and https://cardano-scaling.github.io/cardano-blueprint/pr-preview/pr-67/consensus/multiera.html

Current dingo behavior:
- The 3k/f stability window is implemented and used for epoch-nonce freezing and epoch-boundary header validation (ledger/candidate_nonce.go, ledger/verify_header.go, ledger/state.go StabilityWindow).
- The hard fork combinator safe-zone and PastHorizon bound are computed correctly and served through the NtC HardForkEraHistoryQuery (ledger/hardfork, queries.go). However LedgerState.HardForkSummary deliberately leaves the current era unbounded (SafeZoneSlots 0, ledger/hardfork_summary.go), so the in-memory Summary used by SlotToTime, SlotToEpoch, and header validation projects the current era indefinitely instead of refusing to forecast past the safe zone or the next era boundary.

Required: populate the current era's SafeZoneSlots and TransitionInfo in LedgerState.HardForkSummary (from StabilityWindowForEra) and build the in-memory Summary the same way the NtC query already does, then make the header-acceptance path reject or defer headers whose slot is past the forecast horizon with the existing ErrPastHorizon rather than projecting indefinitely. Pair this with a forecasted nonce and leader-schedule snapshot taken at the intersection.

Scope notes: this touches the hot block and header path and interacts with the eager epoch-advance in ensureEpochForSlot, so it needs DevNet validation. The code comment at ledger/hardfork_summary.go already names this as the intended direction.

Acceptance criteria:
- in-memory HardForkSummary is bounded by the era safe zone, matching the NtC EraHistory query
- header validation refuses to forecast beyond 3k/f or across an unknown era boundary, surfacing ErrPastHorizon
- DevNet and conformance validation across an era boundary

wolf31o2:
Closes #2611

## Summary

- Bound live header verification and slot/epoch/time forecasts to the configured hard-fork safe zone.
- Preserve `hardfork.ErrPastHorizon` through ledger query adapters and serve already-materialized epochs from the immutable cache.
- Preserve valid zero safe-zone values and retain the Shelley genesis system start when a cached shape does not provide one.
- Calculate Blockfrost epoch/era interval endpoints without querying the exclusive end slot.
- Keep stale-node operational `TimeToSlot(time.Now())` available without allowing arbitrary out-of-horizon time queries.
- Allow leader-election precomputation only for the nonce-ready immediate next Praos epoch; general ledger and header forecasts remain bounded.

## Horizon semantics

The live in-memory summary deliberately applies a rolling `TransitionUnknown`
calculation when stored transition state is `TransitionImpossible`, then restores
the reported transition state. This differs from the point-in-time NtC
hard-fork-history response: a live node must continue within the current era
after the confirmed boundary instead of halting there.

The current-era end remains exclusive and is not extended by an extra epoch.
Blockfrost treats it as an interval endpoint, while leader scheduling gets only
the immediate next epoch after its nonce becomes stable. Header acceptance is
therefore still limited to the 3k/f forecast horizon.

## Validation

- `go test ./...`
- `go test -race -count=1 -run 'TestHardForkSummary|TestEpochInfoUsesMaterialized' ./ledger`
- `golangci-lint run ./...`
- `make import-boundaries`
- `make golines`
- all-Dingo `TestEpochBoundaryConsensus`: passed after rollover, all nodes converged at slot 717
- mixed Dingo/cardano-node `TestEpochBoundaryConsensus`: passed after rollover, nodes converged at slots 719-720

`ARCHITECTURE.md` documents the bounded query behavior and narrow operational
exceptions. `DATABASE.md` was checked and is not affected.

### DIFF
--- a/api/blockfrost/adapter.go
+++ b/api/blockfrost/adapter.go
@@ -28,6 +28,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/blinklabs-io/dingo/database"
 	"github.com/blinklabs-io/dingo/database/models"
@@ -537,15 +538,7 @@ func (a *NodeAdapter) CurrentEpoch() (
 			err,
 		)
 	}
-	endSlot := tipEpoch.StartSlot + uint64(tipEpoch.LengthInSlots)
-	endTime, err := a.ledgerState.SlotToTime(endSlot)
-	if err != nil {
-		return EpochInfo{}, fmt.Errorf(
-			"get epoch end time for slot %d: %w",
-			endSlot,
-			err,
-		)
-	}
+	endTime := epochEndTime(startTime, tipEpoch)
 	blockCount, firstBlockSlot, lastBlockSlot, err := a.ledgerState.CountBlocksInSlotRange(
 		tipEpoch.StartSlot,
 		tip.Point.Slot,
@@ -836,13 +829,9 @@ func (a *NodeAdapter) NetworkEras() ([]NetworkEraInfo, error) {
 			)
 		}
 		endSlot := last.StartSlot + uint64(last.LengthInSlots)
-		endTime, err := a.ledgerState.SlotToTime(endSlot)
-		if err != nil {
-			return nil, fmt.Errorf(
-				"get era %s end time: %w",
-				era.Name,
-				err,
-			)
+		endTime := startTime
+		for _, epoch := range epochs {
+			endTime = epochEndTime(endTime, epoch)
 		}
 		slotLengthMs := uint64(last.SlotLength)
 		slotLengthSeconds := (slotLengthMs + 500) / 1000
@@ -869,6 +858,14 @@ func (a *NodeAdapter) NetworkEras() ([]NetworkEraInfo, error) {
 		})
 	}
 	return ret, nil
+}
+
+func epochEndTime(startTime time.Time, epoch models.Epoch) time.Time {
+	// Epoch length and slot length are protocol-bounded.
+	// #nosec G115
+	duration := time.Duration(epoch.LengthInSlots) *
+		time.Duration(epoch.SlotLength) * time.Millisecond
+	return startTime.Add(duration)
 }
 
 // Genesis returns Shelley genesis configuration values.
--- a/ledger/hardfork_summary.go
+++ b/ledger/hardfork_summary.go
@@ -26,11 +26,10 @@ import (
 // transition info.
 //
 // The returned Summary's past eras are closed with bounds computed by walking
-// the epoch cache grouped by EraId. The current era (the last group) is left
-// unbounded (SafeZoneSlots == 0), preserving the existing "project forward
-// using the current era's parameters" behavior of LedgerState.SlotToTime and
-// friends. Once dingo tracks per-era safe zones end-to-end, this will flip to
-// a bounded end sourced from BuildSummary + the real TransitionInfo.
+// the epoch cache grouped by EraId. The current era is passed through
+// hardfork.BuildSummary with the safe zone from the configured era Shape and
+// the ledger's current TransitionInfo. This gives in-memory callers the same
+// bounded forecast inputs used by the NtC HardForkEraHistory query.
 func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
 	// SystemStart is sourced from the Shelley genesis when available. When it
 	// isn't (e.g. SlotToEpoch-style callers that work from the epoch cache
@@ -43,9 +42,10 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
 		}
 	}
 
-	snapshot := ls.loadConsensusSnapshot()
-	cache := snapshot.epochCache
-	transitionInfo := snapshot.transitionInfo
+	consensusState, tipState := ls.loadStateSnapshots()
+	cache := consensusState.epochCache
+	transitionInfo := consensusState.transitionInfo
+	tipSlot := tipState.currentTip.Point.Slot
 
 	if len(cache) == 0 {
 		return nil, errors.New("ledger: no epochs in cache")
@@ -54,7 +54,7 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
 	// Walk the epoch cache grouping contiguous epochs by EraId. Each group
 	// becomes one EraSummary; its Start is derived from the first epoch of
 	// the group, and its End (for past eras) is the Start of the next group.
-	var eras []hardfork.EraSummary
+	eraSummaries := make([]hardfork.EraSummary, 0, len(cache))
 	relTime := time.Duration(0)
 
 	i := 0
@@ -93,13 +93,13 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
 			Epoch:        last.EpochId + 1,
 		}
 
-		era := hardfork.EraSummary{
+		eraSummary := hardfork.EraSummary{
 			EraID: eraID,
 			Start: start,
 			Params: hardfork.EraParams{
 				EpochSize:     epochSize,
 				SlotLength:    slotLen,
-				SafeZoneSlots: 0, // UnsafeIndefiniteSafeZone — preserve projection
+				SafeZoneSlots: 0,
 				GenesisWindow: 0,
 			},
 		}
@@ -107,16 +107,55 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
 		isLast := j == len(cache)
 		if !isLast {
 			// Close this era at the next era's start.
-			era.End = &end
+			eraSummary.End = &end
 		}
-		eras = append(eras, era)
+		eraSummaries = append(eraSummaries, eraSummary)
 
 		i = j
 	}
 
-	return &hardfork.Summary{
-		SystemStart: systemStart,
-		Eras:        eras,
-		Transition:  transitionInfo,
-	}, nil
+	current := eraSummaries[len(eraSummaries)-1]
+	past := eraSummaries[:len(eraSummaries)-1]
+
+	// Use the same configured safe-zone source as the NtC era-history query.
+	// Tests and early bootstrap callers may construct a LedgerState without a
+	// complete node configuration; in that case eraShape is unavailable and
+	// the legacy indefinite safe zone remains the only defensible answer.
+	shape := ls.eraShape()
+	shapeEntry, shapeAvailable := shape.EraForID(current.EraID)
+	if shapeAvailable {
+		current.Params.SafeZoneSlots = shapeEntry.Params.SafeZoneSlots
+		current.Params.GenesisWindow = shapeEntry.Params.GenesisWindow
+		systemStart = shape.SystemStart
+	}
+	if !shapeAvailable {
+		return &hardfork.Summary{
+			SystemStart: systemStart,
+			Eras:        append(past, current),
+			Transition:  transitionInfo,
+		}, nil
+	}
+
+	effectiveTransition := transitionInfo
+	if transitionInfo.State == hardfork.TransitionImpossible {
+		// queryHardForkEraHistory can stop at the confirmed current epoch
+		// boundary because it serves a point-in-time answer. Live slot and
+		// header processing must remain able to cross that boundary in the
+		// same era. Apply the rolling safe zone from the tip while preserving
+		// TransitionImpossible on the returned Summary.
+		effectiveTransition = hardfork.NewTransitionUnknown()
+	}
+
+	summary, err := hardfork.BuildSummary(
+		hardfork.Shape{SystemStart: systemStart},
+		past,
+		current,
+		tipSlot,
+		effectiveTransition,
+	)
+	if err != nil {
+		return nil, err
+	}
+	summary.Transition = transitionInfo
+	return &summary, nil
 }
--- a/ledger/hardfork_summary.go
+++ b/ledger/hardfork_summary.go
@@ -126,7 +126,9 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
 	if shapeAvailable {
 		current.Params.SafeZoneSlots = shapeEntry.Params.SafeZoneSlots
 		current.Params.GenesisWindow = shapeEntry.Params.GenesisWindow
-		systemStart = shape.SystemStart
+		if !shape.SystemStart.IsZero() {
+			systemStart = shape.SystemStart
+		}
 	}
 	if !shapeAvailable {
 		return &hardfork.Summary{
--- a/ledger/hardfork_summary_test.go
+++ b/ledger/hardfork_summary_test.go
@@ -15,7 +15,6 @@
 package ledger
 
 import (
-	"strings"
 	"testing"
 	"time"
 
@@ -29,15 +28,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// minimalShelleyGenesisCfg loads a shelley genesis with only SystemStart
-// populated — matching the pattern used by slot_test.go.
+// minimalShelleyGenesisCfg returns the smallest test configuration that also
+// provides the era shape and stability-window inputs used by HardForkSummary.
 func minimalShelleyGenesisCfg(t *testing.T) *cardano.CardanoNodeConfig {
 	t.Helper()
-	cfg := &cardano.CardanoNodeConfig{}
-	require.NoError(t, cfg.LoadShelleyGenesisFromReader(
-		strings.NewReader(`{"systemStart": "2022-10-25T00:00:00Z"}`),
-	))
-	return cfg
+	return newTestEraHistoryCfg(t)
 }
 
 // TestHardForkSummary_SingleEra verifies the simple case: one era spanning
@@ -66,14 +61,17 @@ func TestHardForkSummary_SingleEra(t *testing.T) {
 	era := sum.Eras[0]
 	assert.Equal(t, uint(1), era.EraID)
 	assert.Equal(t, hardfork.Bound{RelativeTime: 0, Slot: 0, Epoch: 0}, era.Start)
-	assert.Nil(t, era.End, "current era should be unbounded")
+	require.NotNil(t, era.End, "current era should be safe-zone bounded")
+	assert.Equal(t, uint64(26_200), era.End.Slot)
+	assert.Equal(t, uint64(262), era.End.Epoch)
 	assert.Equal(t, uint64(100), era.Params.EpochSize)
 	assert.Equal(t, time.Second, era.Params.SlotLength)
-	assert.Equal(t, uint64(0), era.Params.SafeZoneSlots, "current era unbounded ⇒ SafeZoneSlots=0")
+	assert.Equal(t, uint64(25_920), era.Params.SafeZoneSlots)
+	assert.Equal(t, uint64(25_920), era.Params.GenesisWindow)
 }
 
 // TestHardForkSummary_TwoEras verifies two contiguous eras produce a Summary
-// with the first era bounded and the second (current) era unbounded.
+// with the first era bounded and the second (current) era safe-zone bounded.
 func TestHardForkSummary_TwoEras(t *testing.T) {
 	ls := &LedgerState{
 		epochCache: []models.Epoch{
@@ -111,7 +109,9 @@ func TestHardForkSummary_TwoEras(t *testing.T) {
 	assert.Equal(t, uint(1), shelley.EraID)
 	// Shelley's Start must line up with Byron's End.
 	assert.Equal(t, *byron.End, shelley.Start)
-	assert.Nil(t, shelley.End, "current era unbounded")
+	require.NotNil(t, shelley.End, "current era should be safe-zone bounded")
+	assert.Equal(t, uint64(26_984), shelley.End.Slot)
+	assert.Equal(t, uint64(64), shelley.End.Epoch)
 	assert.Equal(t, uint64(432), shelley.Params.EpochSize)
 	assert.Equal(t, time.Second, shelley.Params.SlotLength)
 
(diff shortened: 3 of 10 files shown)


ledger/hardfork_summary.go
```diff
@@ -93,30 +93,69 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
...
+	if transitionInfo.State == hardfork.TransitionImpossible {
+		// queryHardForkEraHistory can stop at the confirmed current epoch
+		// boundary because it serves a point-in-time answer. Live slot and
+		// header processing must remain able to cross that boundary in the
+		// same era. Apply the rolling safe zone from the tip while preserving
+		// TransitionImpossible on the returned Summary.
+		effectiveTransition = hardfork.NewTransitionUnknown()
+	}
+
+	summary, err := hardfork.BuildSummary(
```
chrisguiney, line 151: **Blocking.** The horizon this produces lands exactly on the current epoch's end slot, which `eraForSlot` treats as *exclusive* — so the epoch-end slot itself becomes `ErrPastHorizon` for most of every mainnet epoch.

`applySafeZone` (`ledger/hardfork/build.go:127`) computes `target = tip+1 + SafeZoneSlots` and then `slotToEpochBound` **ceils** it to an epoch boundary — the intent being that the whole epoch containing `target` is forecastable. But `eraForSlot` (`ledger/hardfork/summary.go:117`) accepts only `slot < era.End.Slot`, so the boundary slot is rejected.

On mainnet the safe zone (3k/f = 129,600) is *smaller* than the epoch size (432,000), so the ceiling lands on the current epoch's own end. I probed with mainnet-shaped params:

```
tip=0       epoch=0 epochEnd=432000 horizon=432000 SlotToTime(epochEnd) err=past era horizon
tip=100     epoch=0 epochEnd=432000 horizon=432000 SlotToTime(epochEnd) err=past era horizon
tip=302398  epoch=0 epochEnd=432000 horizon=432000 SlotToTime(epochEnd) err=past era horizon
tip=302399  epoch=0 epochEnd=432000 horizon=432000 SlotToTime(epochEnd) err=past era horizon
tip=400000  epoch=0 epochEnd=432000 horizon=864000 SlotToTime(epochEnd) err=<nil>
```

So for roughly the first 70% of every mainnet epoch, resolving `epochStart + epochLength` fails. Callers that do exactly that, none updated here:

- **`api/blockfrost/adapter.go:541`** — `endSlot := tipEpoch.StartSlot + LengthInSlots`, then `SlotToTime(endSlot)`. That's `/epochs/latest`. Same pattern at `:839`.
- **`ledger/leader/election.go:551` and `:701`** via `node_forging.go:578` `EpochSlotRange` -> `EpochInfo(epoch)`. For a schedule covering the *next* epoch the whole epoch is past the horizon. This is the block-producer path.

And `EpochInfo` has no epoch-cache fast path (`ledger/slot.go:132` goes straight to the Summary), so an epoch already materialised in `epochCache` is still refused once it's past the bound.

The existing fixtures can't catch this: `newTestEraHistoryCfg` uses `SafeZoneSlots: 25_920` with `EpochSize: 100`, putting the horizon ~259 epochs out so the interaction is invisible. My first probe against that fixture passed cleanly for exactly that reason — worth adding a mainnet-shaped case (safeZone < epochSize) to the table.

Fix direction: either extend the current era one epoch past the ceiling so the epoch containing `tip + safeZone` is fully addressable including its boundary, or update the callers to stop asking for the exclusive bound. The first seems truer to the intent already encoded in `slotToEpochBound`'s ceiling.

ledger/slot.go
```diff
@@ -128,10 +124,10 @@ func (ls *LedgerState) SlotToEpoch(slot uint64) (models.Epoch, error) {
...
-// Epochs within the known epoch cache resolve to cached-era parameters;
-// epochs past the cache are projected using the current era's parameters.
-// Returns an error for an empty cache or for epochs before the first known
-// era.
+// Epochs within the known epoch cache resolve to cached-era parameters; epochs
+// past the cache are projected using the current era's parameters only through
+// the configured safe-zone horizon. Returns an error for an empty cache or for
+// epochs outside that range.
 func (ls *LedgerState) EpochInfo(epoch uint64) (models.Epoch, error) {
 	sum, err := ls.HardForkSummary()
```
chrisguiney, line 144: `ErrPastHorizon` is swallowed exactly where callers need it to defer.

A few lines below, the sentinel is replaced with `errors.New("epoch is outside the known epoch range")`; `SlotToEpoch` and `TimeToSlot` do the same with their own strings. #2611 asked for header acceptance to *"reject **or defer**"* — but a caller can no longer `errors.Is(err, hardfork.ErrPastHorizon)` to tell "not forecastable yet, retry once the ledger advances" from "genuinely unknown epoch."

`verify_header.go` gets this right and wraps with `%w`. The slot/epoch API should too — that's what makes deferral implementable rather than just rejection, and it's what would let the callers in the blocking comment above degrade gracefully instead of erroring.

ledger/hardfork_summary.go
```diff
@@ -93,30 +93,69 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
...
+	// Use the same configured safe-zone source as the NtC era-history query.
+	// Tests and early bootstrap callers may construct a LedgerState without a
+	// complete node configuration; in that case eraShape is unavailable and
+	// the legacy indefinite safe zone remains the only defensible answer.
+	shape := ls.eraShape()
+	shapeEntry, shapeAvailable := shape.EraForID(current.EraID)
+	if shapeAvailable {
+		current.Params.SafeZoneSlots = shapeEntry.Params.SafeZoneSlots
+		current.Params.GenesisWindow = shapeEntry.Params.GenesisWindow
+		systemStart = shape.SystemStart
```
chrisguiney: This unconditionally discards the `systemStart` derived from Shelley genesis at the top of the function. If `shape.SystemStart` is ever the zero time, every slot-to-time answer silently shifts by decades rather than failing loudly.

A `if !shape.SystemStart.IsZero()` guard costs nothing and makes the precedence between the two sources explicit.

ledger/hardfork_summary.go
```diff
@@ -93,30 +93,69 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
...
+	if !shapeAvailable {
+		return &hardfork.Summary{
+			SystemStart: systemStart,
+			Eras:        append(past, current),
+			Transition:  transitionInfo,
+		}, nil
+	}
+
+	effectiveTransition := transitionInfo
+	if transitionInfo.State == hardfork.TransitionImpossible {
```
chrisguiney, line 142: This is the deviation from acceptance criterion 1 ("matching the NtC EraHistory query") — worth surfacing in the PR description rather than only here.

The reasoning is sound and I'd make the same call: bounding at the confirmed boundary would refuse headers in your own era past an epoch boundary, which is a chain halt. Restoring `summary.Transition = transitionInfo` afterwards so the *reported* state stays truthful is a nice touch.

But the consequence is that the in-memory summary and the NtC query now give different horizons for the same ledger state, which is a thing an operator comparing `queryHardForkEraHistory` output against observed node behaviour will trip over.

node_plugin_selection_test.go
```diff
@@ -113,6 +113,10 @@ func registerAPIProbe(
 func newAPIPluginRuntimeNode(t *testing.T) *Node {
 	t.Helper()
 	cardanoCfg := newNodeTestCardanoNodeCfg(t)
+	// The embedded Preview genesis is historical, while this fixture starts
```
chrisguiney: This fixture change is a symptom of the blocking finding rather than a fix for it: a node whose ledger tip is far behind wall clock can no longer resolve wall-clock-relative slots, so the test has to fake `SystemStart` to ~now.

Worth confirming the real path is genuinely fine — a node resuming after extended downtime, or mid-initial-sync from genesis, is in exactly this state, and the fixture no longer exercises it.

chrisguiney (changes requested):
Requesting changes on one item: bounding the forecast changed the contract for existing forward-looking callers, and those callers weren't audited. The result is a live regression on `/epochs/latest` and on block-producer leader-schedule computation, reproducible with mainnet parameters today.

The core change is the right one, and reusing `hardfork.BuildSummary` rather than copying the NtC query's assembly is exactly what #2611 asked for — that's the one place this could have introduced duplication and it didn't.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go test -race -count=1 ./ledger/... .` | **all pass** (10 packages) |
| Horizon-semantics probes (two, written for this review) | see the inline comment on `BuildSummary` |
| `golangci-lint run ./...` | **0 issues** |
| CI on this PR | 10 checks pass, 0 failing |

(Lint and CI were pending when I first submitted this review — my local tool harness was down. Both have since been run and are clean; the table above is current.)

## Docs

`ARCHITECTURE.md` is updated with the forecast bound, the rolling behaviour, and the intersection-state reads. `DATABASE.md` checked and not affected — nothing here touches models, migrations, or the metadata surface.

## Process: DevNet and conformance

#2611's acceptance criteria and its scope note are explicit: *"this touches the hot block and header path and interacts with the eager epoch-advance in `ensureEpochForSlot`, so it needs DevNet validation"*, and *"DevNet and conformance validation across an era boundary"*. CLAUDE.md says the same for header-verification and epoch-boundary work. The description mentions neither. An era-boundary DevNet run is the thing most likely to surface interactions the unit tests can't — and given the finding below, I'd want it before merge rather than after.

## Acceptance criterion 1 is knowingly not met — worth saying so in the description

The issue requires *"in-memory HardForkSummary is bounded by the era safe zone, **matching the NtC EraHistory query**"*. The `TransitionImpossible` substitution (inline comment below) means it deliberately doesn't match. The reasoning is sound and I'd make the same call — but it's a knowing deviation from a stated criterion and belongs in the PR body, not only a code comment.

ledger/hardfork_summary_test.go
```diff
@@ -168,8 +168,232 @@ func TestHardForkSummary_CarriesTransitionInfo(t *testing.T) {
...
+					),
+				},
+				config: LedgerStateConfig{CardanoNodeConfig: cfg},
+			}
+			shape := hardfork.Shape{
+				SystemStart: cfg.ShelleyGenesis().SystemStart,
+				Eras: []hardfork.ShapeEntry{{
+					EraID: eras.ConwayEraDesc.Id,
+					Params: hardfork.EraParams{
+						EpochSize:     432_000,
```
chrisguiney, line 279: This is my probe from the last review turned into a test, and it's the right resolution of the blocking finding.

`EpochSize: 432_000` with `SafeZoneSlots: 129_600` is the mainnet shape where `safeZone < epochSize`, which the pre-existing `newTestEraHistoryCfg` fixture (`SafeZoneSlots: 25_920`, `EpochSize: 100`) could never express — the horizon sat ~259 epochs out there, which is exactly why the interaction was invisible. The two cases now pin the flip precisely: `tipSlot: 302_399` keeps `wantEndSlot: 432_000`, `302_400` moves it to `864_000`.

What I want to highlight is the choice this encodes. `assert.ErrorIs(t, err, hardfork.ErrPastHorizon)` on `:299` and `:307` means the epoch-end slot remains past the horizon *by design* — you kept the semantics and fixed the callers rather than extending the era by an epoch. I think that's the better of the two directions I offered: it stays closer to the NtC query, and the "Horizon semantics" section in the description now states it explicitly instead of leaving it to be inferred. The consequence is that any future caller resolving an exclusive epoch-end slot is broken on arrival, which is now a tested property rather than a latent surprise.

api/blockfrost/adapter.go
```diff
@@ -537,15 +538,7 @@ func (a *NodeAdapter) CurrentEpoch() (
...
-	endSlot := tipEpoch.StartSlot + uint64(tipEpoch.LengthInSlots)
-	endTime, err := a.ledgerState.SlotToTime(endSlot)
-	if err != nil {
-		return EpochInfo{}, fmt.Errorf(
-			"get epoch end time for slot %d: %w",
-			endSlot,
-			err,
-		)
-	}
+	endTime := epochEndTime(startTime, tipEpoch)
```
chrisguiney, line 541: This fixes the `/epochs/latest` half of the blocking finding, and computing the end time from the epoch's own length rather than resolving the boundary slot is the right shape — era transitions land on epoch boundaries, so slot length is constant across the interval and the arithmetic is exactly equivalent to the old `SlotToTime(endSlot)` without touching the horizon.

**Coverage gap, though.** I reverted this line back to `SlotToTime(endSlot)` and re-ran `go test ./api/blockfrost/` — **it still passes**. Nothing in this package would catch a reintroduction.

That matters more than usual here because `/epochs/latest` was one of the two live user-visible breakages, and the protection currently lives one package away in `TestHardForkSummary_MainnetForecastBoundary`. DevNet doesn't cover it either: its epochs are tiny (rollover at slot ~717), so `safeZone < epochSize` never arises there. An adapter-level test at a mainnet-shaped boundary — assert `CurrentEpoch()` succeeds with a tip in the first 70% of an epoch — is what keeps this fixed.

api/blockfrost/adapter.go
```diff
@@ -836,13 +829,9 @@ func (a *NodeAdapter) NetworkEras() ([]NetworkEraInfo, error) {
...
-		endTime, err := a.ledgerState.SlotToTime(endSlot)
-		if err != nil {
-			return nil, fmt.Errorf(
-				"get era %s end time: %w",
-				era.Name,
-				err,
-			)
+		endTime := startTime
+		for _, epoch := range epochs {
+			endTime = epochEndTime(endTime, epoch)
```
chrisguiney, line 834: Summing per-epoch durations makes the era end time depend on the epoch list being contiguous and complete for the era, which the old `SlotToTime(endSlot)` did not.

I checked the obvious worry first and it's fine — `epochs` is scoped per era via `GetEpochsByEra(era.Id, nil)` on `:799`, so there's no cross-era accumulation. The remaining issue is narrower: if any epoch row for the era is missing, `End.Time` comes up short by that epoch's duration while `End.Slot` on `:831` stays correct, because that one is still derived from `last.StartSlot + last.LengthInSlots`. So the two halves of the same `End` object can disagree, and the response looks internally inconsistent rather than obviously wrong. Worth a thought for a Mithril-bootstrapped node, where whether every historical epoch row is backfilled isn't obvious from here.

One expression avoids both that and the loop:

```go
endTime := startTime.Add(
    time.Duration(endSlot-first.StartSlot) *
        time.Duration(last.SlotLength) * time.Millisecond,
)
```

Slot length is constant within an era, so this is completeness-independent, still horizon-independent, and O(1) instead of O(epochs) per era.

ledger/slot.go
```diff
@@ -128,20 +129,28 @@ func (ls *LedgerState) SlotToEpoch(slot uint64) (models.Epoch, error) {
...
-// Epochs within the known epoch cache resolve to cached-era parameters;
-// epochs past the cache are projected using the current era's parameters.
-// Returns an error for an empty cache or for epochs before the first known
-// era.
+// Epochs within the known epoch cache resolve to cached-era parameters; epochs
+// past the cache are projected using the current era's parameters only through
+// the configured safe-zone horizon. Returns an error for an empty cache or for
+// epochs outside that range.
 func (ls *LedgerState) EpochInfo(epoch uint64) (models.Epoch, error) {
+	cache := ls.loadConsensusSnapshot().epochCache
```
chrisguiney, line 137: This addresses the sub-point I raised about `EpochInfo` going straight to the Summary, and it's a performance win rather than a cost — worth saying because a linear scan usually reads as the opposite.

`HardForkSummary()` rebuilds on every call: it walks the entire `epochCache` grouping contiguous epochs by era and allocates `eraSummaries` sized to the cache. This check runs first, `slices.Backward` puts the most recent epochs first, and the only callers (`node_forging.go:579` and `:596`) ask for the current or next epoch — so the common path hits in one or two comparisons and skips the rebuild entirely.

I also checked field parity, since two paths returning different shapes depending on cache state would be a nasty one to debug: `epochBoundaryInfo` copies exactly the five fields the Summary path populates on `:158-167` (`EpochId`, `StartSlot`, `EraId`, `SlotLength`, `LengthInSlots`). Having the helper strip anything else the cached `models.Epoch` carries, rather than returning it whole, is the detail that makes the two paths interchangeable.

ledger/slot.go
```diff
@@ -33,10 +35,8 @@ var ErrBeforeGenesis = errors.New("time is before genesis start")
 //
 // Slot 0 always maps to Shelley genesis SystemStart, regardless of whether
 // the epoch cache is populated. Other slots are resolved via the
-// hardfork.Summary built from the LedgerState's epoch cache; slots past the
-// last known epoch are projected using the current era's slot length (the
-// Summary's current era is unbounded, mirroring the legacy projection
-// behavior).
+// hardfork.Summary built from the LedgerState's epoch cache. The current era
+// can be projected only through its configured safe-zone horizon.
 func (ls *LedgerState) SlotToTime(slot uint64) (time.Time, error) {
```
chrisguiney, line 40: One instance of the original finding survives, in a caller outside this diff.

`SlotClock.TimeUntilNextEpoch()` (`ledger/slot_clock.go:238-249`) calls `epochInfo.EndSlot()` and passes it straight to this function. `EndSlot()` is `StartSlot + uint64(LengthInSlots)` (`slot_clock.go:76-78`, doc comment: *"the first slot of the next epoch"*) — the exclusive bound that `TestHardForkSummary_MainnetForecastBoundary` now pins as past-horizon. The error is propagated, so on mainnet parameters that method fails for roughly the first 70% of every epoch. Its own doc comment claims it *"works regardless of sync state and can be called during catch up or load"*, which is now false exactly during catch-up.

Two reasons I'm not blocking on it:

- no production callers — grep finds only the definition and its comment;
- its test (`slot_clock_test.go:164`) uses `newMockSlotTimeProvider`, so it can't exercise the real summary. That's why it survived.

It's the same one-line fix `epochEndTime` just gave the adapter — derive the duration from `LengthInSlots × SlotLength` instead of resolving the boundary slot. Worth doing in this PR while the reasoning is loaded, because the alternative is that the next caller of `TimeUntilNextEpoch` reintroduces the finding with no test to catch it.

node_forging.go
```diff
@@ -577,7 +578,40 @@ func (a *epochInfoAdapter) EpochSlotRange(
...
+		// Praos eras use the Shelley genesis epoch/slot dimensions.
+		readyEpoch, ready := a.ledgerState.NextEpochNonceReadyEpoch()
+		currentEpoch := a.ledgerState.CurrentEpoch()
+		if !ready ||
+			readyEpoch != epoch ||
+			currentEpoch == ^uint64(0) ||
+			epoch != currentEpoch+1 {
+			return leader.EpochSlotRange{}, err
+		}
+		currentInfo, currentErr := a.ledgerState.EpochInfo(
```
chrisguiney, line 596: The guard chain above this is appropriately tight — nonce ready, `readyEpoch == epoch`, `currentEpoch` not the `^uint64(0)` sentinel, and `epoch == currentEpoch+1` — so this only widens the forecast for the one epoch leader election genuinely needs, which is what the description claims.

Worth noting the assumption it rests on: `StartSlot: currentInfo.StartSlot + slotCount` with `SlotCount: slotCount` takes the *current* epoch's length as the next epoch's. The comment says "All Praos eras use the Shelley genesis epoch/slot dimensions," which holds Shelley through Conway, so this is fine today and I wouldn't change it. But it's a silent wrong answer rather than an error if epoch length ever changes at an era boundary, and the guard can't detect that — the request that reaches here is precisely one for an epoch the Summary won't describe.

One bonus from the cache fast path: the `EpochInfo(currentEpoch)` call here now resolves from `epochCache` rather than the Summary, so it can't itself trip the horizon it was called to work around.

chrisguiney (approved):
Approving. Re-review of `638c1c13`; my earlier review was against `52dafb5b`. **All five findings are addressed**, and the era-boundary DevNet validation I asked for in that review was run in both modes.

| Earlier finding | Status |
|---|---|
| **Blocking:** horizon lands on the exclusive epoch end, breaking `/epochs/latest` and leader-schedule computation | **Fixed** at every caller I named, and the horizon behaviour is now pinned by a mainnet-shaped test |
| `ErrPastHorizon` swallowed, so callers can't `errors.Is` to defer | **Fixed** — `%w` in `SlotToEpoch`, `EpochInfo`, `TimeToSlot`. This turned out to be a prerequisite rather than a nicety: `node_forging.go`'s `errors.Is(err, hardfork.ErrPastHorizon)` only works because of it |
| `systemStart` from Shelley genesis unconditionally discarded | **Fixed** — `if !shape.SystemStart.IsZero()`, exactly as suggested |
| Criterion-1 deviation only in a code comment | **Fixed** — the "Horizon semantics" section explains the `TransitionImpossible` substitution and that the era end stays exclusive by choice |
| `node_plugin_selection_test.go` faking `SystemStart` to ~now | **Reverted** — the real path is exercised again rather than papered over |

Also crediting the process: `TestEpochBoundaryConsensus` was run all-Dingo (converged at slot 717) and mixed Dingo/cardano-node (719-720). That was the thing I said I'd want before merge rather than after.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go test -race -count=1 ./ledger/... ./api/blockfrost/... .` | **all pass** (11 packages) |
| Coverage experiment (reverted the adapter fix) | **tests still pass** — inline on `adapter.go:541` |

## Performance

Net improvement, and the biggest win is the one that looks like a cost. `HardForkSummary()` rebuilds the summary on every call — it walks the whole `epochCache` grouping epochs by era and allocates `eraSummaries` sized to the cache. The new `EpochInfo` cache fast path skips that entirely on a hit, and because `slices.Backward` puts recent epochs first while the only two callers ask for the current or next epoch, it hits within one or two iterations. `epochEndTime` likewise replaces a Summary lookup with arithmetic.

`epochCache` does grow without bound (`verify_header.go:1316` appends without trimming), so a scan is O(epochs since genesis) worst case — ~570 integer comparisons on mainnet today. A backward scan is already the established idiom here (`state.go:4973`, `verify_header.go:1022`) and no caller loops over epochs calling `EpochInfo`, so there is no quadratic path. Not a concern.

The one regression is `NetworkEras`'s new O(epochs)-per-era loop replacing an O(1) `SlotToTime`; negligible in absolute terms and removable — inline on `adapter.go:834`.

## Documentation

`ARCHITECTURE.md` documents the bounded query behaviour and the narrow operational exceptions (`CurrentSlot` extrapolation, nonce-ready leader precompute). `DATABASE.md` checked and not affected.

Three non-blocking notes and one coverage gap inline.

Issue #3005, opened 2026-07-29
ledger: replay-path recovery oscillates without convergence on unresolved-producer failures near the tip


wolf31o2:
A preview node syncing from genesis reaches about 99.9 percent (ledger tip slot ~118660633, epoch 1373) and then enters a sustained oscillation that never converges. The tip gap stays around 35k slots and grows slowly as the live tip advances. The node does not silently commit bad state; it rejects and rewinds every cycle, but never makes forward progress. Observed on commit 7ecca0fc (the #2973 fix commit itself, v0.68.0 line).

## How it manifests

Roughly every 60 to 90 seconds:

- A block at ~118661xxx fails Conway UTxO validation because a required input is missing. Recovery logs "detected inconsistent local ledger state during replay, rewinding primary chain and metadata state" with recovery_strategy=security-param-fallback. The failing slot creeps forward (118661308, 118661413, 118661532, 118661675) but the chain is rewound to ~118651xxx each cycle and never crosses 118660633.
- producer_tx_hash in the log equals the missing input's own tx hash and producer_block_slot is a real ~118651xxx slot. This is the security-param fallback (ProducerUnresolved), not a resolved producer: producer_tx_hash falls back to input.Id().String() (which is by definition the producing tx hash, so the equality is tautological) and producer_block_slot is the anchor block K positions before the failing block. The producing tx is not resolvable via UtxoByRef, GetTransactionByHash, the tx blob, or the 4096-block backward chain scan.
- A peer periodically offers rollback-to-origin (rollback_slot=0, fork_depth ~4520527); the chain correctly rejects it as exceeding K, which triggers "primary chain and ledger diverged during live chainsync recovery, reconciling to common ancestor" and "ledger tip ahead of primary chain tip, rolling back metadata to chain tip" (chain_tip ~118651xxx, ledger_tip 118660633).
- ledger_tip_hash is constant (d66873073bcd587299eb37de5f9a649bbaad685dad95d368c3116daf04bc86c4) across every cycle while the chain tip at ~118651xxx changes hash (fff45001..., 7ffe2eee...). The ledger deterministically rebuilds the same state to 118660633 and always fails on the next block; the near-tip region churns between forks.

## Root cause

[...]

Issue #2939, opened 2026-07-24
origin/main enters preview primary-chain validation rewind loop after reaching tip


wolf31o2:
## Summary

An image built from `origin/main` at `2a96a195` (`v0.66.2-21-g2a96a195`) failed during live preview block-producer testing after initially reaching tip. The node logged repeated `validation failure after reaching tip, rewinding primary chain` messages, rewound to much older slots, and developed a large forge tip gap.

Rolling back to a build containing the primary-chain drain and governance import fixes allowed the same node state to catch back up and forge/adopt the next scheduled leader slots.

## Version tested

- `origin/main`: `2a96a19502e888e9a67c560007aee94b48f03462`
- `git describe`: `v0.66.2-21-g2a96a195`
- Latest release at the time of test: `v0.66.2` (`2f5a84c`)
- Network: `preview`

## Observed behavior

The `origin/main` build initially reached tip and forged/adopted a block:

```text
2026-07-24T12:36:05Z chain extended, new tip: 765194998a39b719f41e051fd46d9bf29a4d9f0d630053501405f906d714e286 at slot 118240565 source=forged
```

Later, after reaching tip, it began repeatedly rewinding the primary chain:

```text
2026-07-24T13:04:30Z validation failure after reaching tip, rewinding primary chain tx_hash=10c1641e9db4e3e07ad649923ab2bc40e6795e25b87745e24d20797ba088fa1b failing_block_slot=118242270 ledger_tip_slot=118242199 rewind_target_slot=118242199 attempt=1
2026-07-24T13:04:30Z validation failure after reaching tip, rewinding primary chain tx_hash=10c1641e9db4e3e07ad649923ab2bc40e6795e25b87745e24d20797ba088fa1b failing_block_slot=118242270 ledger_tip_slot=118242199 rewind_target_slot=118229216 attempt=2
2026-07-24T13:04:45Z validation failure after reaching tip, rewinding primary chain tx_hash=9744cbafc5ce915bc0e5488bc7ae486ec82d1dcc7a8837df7cc4cedaeafbfd11 failing_block_slot=118230025 ledger_tip_slot=118229743 rewind_target_slot=118229743 attempt=1
[...]

PR #3008, opened 2026-07-29
fix(ledger): stop non-converging replay recovery

wolf31o2:
Refs #3005

## Summary

- Bound unresolved-producer replay recovery by the applied ledger high-water mark and hold instead of repeatedly pruning another security-parameter window.
- Publish a neutral ChainSync resync request before local rollback work so rollback failures cannot suppress peer rotation.
- Treat a primary chain already below the applied ledger tip as already held while a fresh ChainSync intersection refills it.
- Keep Mithril-boundary rejection ahead of no-progress tracking and rewind logging, so rejected recovery attempts neither consume the no-progress budget nor claim a rewind occurred.
- Add metrics, routing coverage, the ledger-ahead-of-chain regression, and architecture documentation.

This addresses the non-converging replay fallback documented in #3005. It does not claim to fix the unresolved producer root cause or change reachedTip robustness.

## Validation

- go build ./...
- go test -race -count=1 ./ledger/... ./ouroboros/... ./event/...
- go test -race -count=1 ./internal/test/conformance/
- golangci-lint run ./...
- make import-boundaries
- make golines

modernize and nilaway retain unrelated pre-existing findings outside this change.

### DIFF
--- a/ledger/replay_recovery.go
+++ b/ledger/replay_recovery.go
@@ -87,6 +87,15 @@ const maxAtTipRecoveryAttempts = 3
 // false-positive validation rejection).
 const maxAtTipRecoveryDescents = 2
 
+// maxReplayRecoveryNoProgress caps consecutive unresolved-producer replay
+// recoveries that rebuild only to the same (or an older) applied ledger tip.
+// The failing block may change and creep forward while the applied high-water
+// mark remains fixed, so the ledger tip—not the failure identity—is the
+// convergence signal. Once tripped, recovery stops pruning another
+// security-parameter window and holds at the applied tip while forcing a fresh
+// ChainSync connection (issue #3005).
+const maxReplayRecoveryNoProgress = 2
+
 type atTipRecoveryAttempt struct {
 	BlockPoint ocommon.Point
 	TxHash     []byte
@@ -188,8 +197,32 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 	if candidate.ProducerTx != nil {
 		producerTxHash = hex.EncodeToString(candidate.ProducerTx.Hash)
 	}
+	if ls.recoveryRollbackExceedsMithrilBoundary(candidate.RollbackPoint) {
+		if err := ls.rejectReplayRecoveryAtMithrilBoundary(
+			validationErr,
+			candidate,
+			producerTxHash,
+		); err != nil {
+			return false, err
+		}
+		return true, nil
+	}
 	rewindPrimaryChain := candidate.ProducerUnresolved &&
 		ls.config.ChainManager != nil
+	rewindPoint := candidate.RollbackPoint
+	replayHolding := false
+	primaryChainAlreadyHeld := false
+	var ledgerTip ochainsync.Tip
+	if rewindPrimaryChain {
+		ls.RLock()
+		ledgerTip = ls.currentTip
+		ls.RUnlock()
+		replayHolding = ls.observeReplayRecoveryTip(ledgerTip.Point.Slot)
+		if replayHolding {
+			rewindPoint = ledgerTip.Point
+			primaryChainAlreadyHeld = ls.chain.Tip().Point.Slot < rewindPoint.Slot
+		}
+	}
 	recoveryAction := "rewinding metadata state"
 	if rewindPrimaryChain {
 		recoveryAction = "rewinding primary chain and metadata state"
@@ -203,22 +236,33 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 		"missing_input", candidate.Input.String(),
 		"producer_tx_hash", producerTxHash,
 		"producer_block_slot", candidate.ProducerBlock.Slot,
-		"rollback_slot", candidate.RollbackPoint.Slot,
-		"rollback_hash", hex.EncodeToString(candidate.RollbackPoint.Hash),
+		"rollback_slot", rewindPoint.Slot,
+		"rollback_hash", hex.EncodeToString(rewindPoint.Hash),
+		"holding", replayHolding,
 	)
-	if ls.recoveryRollbackExceedsMithrilBoundary(candidate.RollbackPoint) {
-		if err := ls.rejectReplayRecoveryAtMithrilBoundary(
-			validationErr,
-			candidate,
-			producerTxHash,
-		); err != nil {
-			return false, err
-		}
-		return true, nil
+	if replayHolding {
+		ls.metrics.replayRecoveryNonConverging.Inc()
+		ls.config.Logger.Warn(
+			"replay recovery not converging after unresolved-producer failures, holding at applied ledger tip",
+			"component", "ledger",
+			"tx_hash", hex.EncodeToString(validationErr.TxHash),
+			"failing_block_slot", validationErr.BlockPoint.Slot,
+			"ledger_tip_slot", ledgerTip.Point.Slot,
+			"ledger_tip_hash", hex.EncodeToString(ledgerTip.Point.Hash),
+			"primary_chain_tip_slot", ls.chain.Tip().Point.Slot,
+			"primary_chain_already_held", primaryChainAlreadyHeld,
+			"no_progress_count", ls.replayRecoveryNoProgressCount,
+			"fallback_rollback_slot", candidate.RollbackPoint.Slot,
+			"hint", "forcing a fresh chainsync intersection; persistent failures may require operator intervention",
+		)
+		// Peer rotation is the escape mechanism for a held recovery. Publish
+		// it before the rollback calls so an unexpected local rollback error
+		// cannot suppress the fresh ChainSync intersection.
+		ls.publishReplayRecoveryNonConvergingResync(rewindPoint)
 	}
-	if rewindPrimaryChain {
+	if rewindPrimaryChain && !primaryChainAlreadyHeld {
 		if err := ls.rollbackPrimaryChainInSecurityParamWindows(
-			candidate.RollbackPoint,
+			rewindPoint,
 		); err != nil {
 			return false, fmt.Errorf(
 				"rewind primary chain for replay recovery: %w",
@@ -230,13 +274,76 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 	// available. If metadata synchronization fails, the primary chain is still
 	// at a valid retained point and the standard divergence reconciler can
 	// finish rolling metadata back to its common ancestor.
-	if err := ls.rollback(candidate.RollbackPoint); err != nil {
+	if err := ls.rollback(rewindPoint); err != nil {
 		return false, fmt.Errorf(
 			"rollback ledger state for replay recovery: %w",
 			err,
 		)
 	}
 	return true, nil
+}
+
+// observeReplayRecoveryTip records the applied tip seen immediately before an
+// unresolved-producer fallback. Recovery is non-converging when repeated
+// attempts fail to exceed the first observed high-water mark, even if peers
+// offer different failing blocks at slightly higher slots each cycle.
+func (ls *LedgerState) observeReplayRecoveryTip(tipSlot uint64) bool {
+	if !ls.replayRecoveryTipTracked {
+		ls.replayRecoveryTipTracked = true
+		ls.replayRecoveryHighWaterSlot = tipSlot
+		return false
+	}
+	if tipSlot <= ls.replayRecoveryHighWaterSlot {
+		ls.replayRecoveryNoProgressCount++
+	} else {
+		ls.replayRecoveryHighWaterSlot = tipSlot
+		ls.replayRecoveryNoProgressCount = 0
+		ls.replayRecoveryHolding = false
+	}
+	if ls.replayRecoveryNoProgressCount >= maxReplayRecoveryNoProgress {
+		ls.replayRecoveryHolding = true
+	}
+	return ls.replayRecoveryHolding
+}
+
+// resetReplayRecoveryNonProgress clears a replay hold only after block
+// application advances beyond the high-water mark that recovery could not
+// cross. Replaying blocks back up to that exact point is not progress.
+func (ls *LedgerState) resetReplayRecoveryNonProgress(newTipSlot uint64) {
+	if !ls.replayRecoveryTipTracked ||
+		newTipSlot <= ls.replayRecoveryHighWaterSlot {
+		return
+	}
+	ls.replayRecoveryTipTracked = false
+	ls.replayRecoveryHighWaterSlot = 0
+	ls.replayRecoveryNoProgressCount = 0
+	ls.replayRecoveryHolding = false
+}
+
+func (ls *LedgerState) publishReplayRecoveryNonConvergingResync(
+	point ocommon.Point,
+) {
+	if ls.config.EventBus == nil {
+		return
+	}
+	var activeConnId ouroboros.ConnectionId
+	if ls.config.GetActiveConnectionFunc != nil {
+		if connId := ls.config.GetActiveConnectionFunc(); connId != nil {
+			activeConnId = *connId
+		}
+	}
+	ls.config.EventBus.Publish(
+		event.ChainsyncResyncEventType,
+		event.NewEvent(
+			event.ChainsyncResyncEventType,
+			event.ChainsyncResyncEvent{
+				ConnectionId: activeConnId,
+				Reason: event.
+					ChainsyncResyncReasonReplayRecoveryNonConverging,
+				Point: point,
+			},
+		),
+	)
 }
 
 // rollbackPrimaryChainInSecurityParamWindows reaches a deep corrective rewind
--- a/ledger/replay_recovery_test.go
+++ b/ledger/replay_recovery_test.go
@@ -1417,6 +1417,214 @@ func TestTryRecoverFromTxValidationErrorFallsBackToSecurityParamWindow(
 	}
 }
 
+// TestTryRecoverFromTxValidationErrorReplayFallbackStopsNonConvergingRewinds
+// reproduces the terminal #3005 recovery cycle after prior, distinct failures
+// failed to advance the applied ledger high-water mark. Once the guard trips,
+// recovery must keep the applied tip instead of pruning another
+// security-parameter window and must force a fresh ChainSync connection even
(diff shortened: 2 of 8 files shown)


ledger/replay_recovery.go
```diff
@@ -188,8 +197,30 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+	rewindPoint := candidate.RollbackPoint
+	replayHolding := false
+	var ledgerTip ochainsync.Tip
+	if rewindPrimaryChain {
+		ls.RLock()
+		ledgerTip = ls.currentTip
+		ls.RUnlock()
+		replayHolding = ls.observeReplayRecoveryTip(ledgerTip.Point.Slot)
+		if replayHolding {
+			rewindPoint = ledgerTip.Point
```
chrisguiney, line 222: **Blocking.** Setting `rewindPoint = ledgerTip.Point` and passing it to `rollbackPrimaryChainInSecurityParamWindows` fails in exactly the state #3005 describes.

That helper does `database.BlockByPoint(ls.db, point)` at `:361` and errors if the block is absent, or errors at `:378` if the chain tip index is below the target. The issue documents the ledger tip running **ahead** of the primary chain tip every cycle:

> "ledger tip ahead of primary chain tip, rolling back metadata to chain tip (chain_tip ~118651xxx, ledger_tip 118660633)"

I probed both shapes of that:

```
A) rewind to a point absent from the chain store:  err = lookup replay recovery target: block not found
B) rewind to a point above the pruned chain tip:   err = lookup replay recovery target: block not found
```

B collapses into A because `Chain.Rollback` *deletes* pruned blocks from the store, so a ledger tip above the chain tip is also absent from it.

Consequence chain: the helper errors -> `tryRecoverFromTxValidationError` returns `(false, err)` -> `process block batch` error -> `handleLedgerProcessBlocksError` warns and restarts the pipeline immediately with no backoff. And the resync at `:277` is never reached (see that comment).

Three ways out, any one is enough:

1. Publish the resync before the rewind, or unconditionally when `replayHolding` — peer rotation shouldn't be gated on a rewind that is a no-op when already held.
2. Clamp `rewindPoint` to the current chain tip when the ledger tip isn't in the chain. Holding at "the chain tip we actually have" is the same intent.
3. Have the helper treat "target absent / behind chain tip" as already-held (return nil) rather than an error.

Worth crediting: the metric and operator `Warn` are emitted *before* the rewind, so the diagnostic signal survives this. It's only the escape mechanism that's lost.

ledger/replay_recovery.go
```diff
@@ -230,15 +267,81 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
 	// finish rolling metadata back to its common ancestor.
-	if err := ls.rollback(candidate.RollbackPoint); err != nil {
+	if err := ls.rollback(rewindPoint); err != nil {
 		return false, fmt.Errorf(
 			"rollback ledger state for replay recovery: %w",
 			err,
 		)
 	}
+	if replayHolding {
+		ls.publishReplayRecoveryNonConvergingResync(rewindPoint)
```
chrisguiney: This is the peer rotation the whole guard exists to trigger, and it sits after both rollback calls — so it is skipped on any rewind or metadata-rollback error, including the one in the comment on `:221`. Moving it above the rewind (or making it unconditional when `replayHolding`) decouples the escape hatch from a rewind that may legitimately be a no-op once held.

Separately, on the steady-state behaviour: while held this fires on **every** cycle, with no backoff and no give-up, so a permanently-wedged node rotates peers roughly once a minute indefinitely. That is the intended mechanism and it matches what the at-tip path already does, so I'd call it consistent rather than wrong — but it's worth a deliberate decision on whether an indefinitely-held node should eventually escalate from `Warn` to a throttled operator `Error`, the way the over-K rollback detector does.

ledger/replay_recovery.go
```diff
@@ -87,6 +87,15 @@ const maxAtTipRecoveryAttempts = 3
...
 const maxAtTipRecoveryDescents = 2
 
+// maxReplayRecoveryNoProgress caps consecutive unresolved-producer replay
+// recoveries that rebuild only to the same (or an older) applied ledger tip.
+// The failing block may change and creep forward while the applied high-water
+// mark remains fixed, so the ledger tip—not the failure identity—is the
+// convergence signal. Once tripped, recovery stops pruning another
+// security-parameter window and holds at the applied tip while forcing a fresh
+// ChainSync connection (issue #3005).
+const maxReplayRecoveryNoProgress = 2
```
chrisguiney, line 97: Scope: `Closes #3005` overclaims what this delivers.

The issue lists three fix directions. This implements the first (replay-path guard). The second — `reachedTip` robustness so an oscillating node uses the guarded at-tip path — isn't touched; there are no changes to `IsAtTip`/`reachedTip`. The third — *why* the producer is unresolvable — is the root cause and is untouched.

So a wedged node now stops churning and holds cleanly with a clear operator signal, which is a real improvement in resource use and diagnosability. But the reported symptom ("syncing preview from genesis reaches ~99.9% and never converges") is not resolved: the node still can't cross the bad band except by luck, if a rotated peer happens to offer a consistent continuation.

That's the same distinction drawn in the #3005 follow-up comment about #2999 — "chain-selection changes alone move the wedge rather than remove it". Suggest referencing rather than closing #3005, or splitting the root-cause investigation into its own issue first.

ledger/replay_recovery.go
```diff
@@ -230,15 +267,81 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+		ls.publishReplayRecoveryNonConvergingResync(rewindPoint)
+	}
 	return true, nil
 }
 
+// observeReplayRecoveryTip records the applied tip seen immediately before an
+// unresolved-producer fallback. Recovery is non-converging when repeated
+// attempts fail to exceed the first observed high-water mark, even if peers
+// offer different failing blocks at slightly higher slots each cycle.
+func (ls *LedgerState) observeReplayRecoveryTip(tipSlot uint64) bool {
```
chrisguiney, line 290: Crediting this because it's the thing I went looking to criticise and didn't find: "give the replay path the same guard as the at-tip path" usually produces copy-paste, and this doesn't.

| | at-tip (#2939) | replay (#3005) |
|---|---|---|
| Signal | failing-slot descent across *distinct failures* | applied ledger tip **high-water mark** |
| Rationale | a repeating `(block, tx)` failure means the fork is losing | failing slots creep forward while the applied tip stays fixed |

The doc comment states exactly why failure identity can't be the signal here, and that reasoning is drawn from the issue's own evidence. Two hold latches and two reset functions is a fair price for two genuinely different signals.

The one thing missing around this is a test where the ledger tip is *above* the chain tip — the state that produces the blocking finding on `:221`.

ledger/replay_recovery.go
```diff
@@ -203,22 +234,28 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 		"missing_input", candidate.Input.String(),
 		"producer_tx_hash", producerTxHash,
 		"producer_block_slot", candidate.ProducerBlock.Slot,
-		"rollback_slot", candidate.RollbackPoint.Slot,
-		"rollback_hash", hex.EncodeToString(candidate.RollbackPoint.Hash),
+		"rollback_slot", rewindPoint.Slot,
+		"rollback_hash", hex.EncodeToString(rewindPoint.Hash),
+		"holding", replayHolding,
 	)
```
chrisguiney, line 242: Noting the reordering so a reader diffing this doesn't have to work it out: the Mithril-boundary rejection moved *above* this log line and above `observeReplayRecoveryTip`, so a Mithril rejection no longer counts toward the no-progress budget and no longer emits a misleading "rewinding primary chain and metadata state" line for a case that doesn't rewind.

Both look like improvements to me, and `rejectReplayRecoveryAtMithrilBoundary` does its own logging so nothing is lost. Just worth a line in the description, since it's a behaviour change that isn't obviously part of "stop non-converging replay recovery".

chrisguiney (changes requested):
Requesting changes on one item: in the exact state #3005 documents — ledger tip ahead of the primary chain tip — the new hold path's rewind errors out, and because the peer-rotation resync is published *after* that rewind, the escape mechanism the guard exists to trigger never fires. I reproduced it with a probe (details inline on `:221`).

The guard design itself is good, and the convergence signal is the right one. My other comments are scope and observability.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go test -race -count=1 ./ledger/... ./ouroboros/... ./event/...` | **all pass** (11 packages) |
| Hold-path probe (written for this review) | **both cases fail** — see `:221` |
| CI on this PR | 10 checks pass, 0 failing |

## Duplication

I checked this specifically, because "give the replay path the same guard as the at-tip path" is the kind of instruction that usually produces copy-paste. It didn't — see the inline note on `observeReplayRecoveryTip`. The two paths use genuinely different convergence signals for a reason that's stated in the code and grounded in the issue's own evidence.

## Tests

`TestResetReplayRecoveryNonProgressRequiresNewHighWater` is the good one — it pins the subtle rule that replaying *back up to* the high-water mark is not progress (`reset(140)` retains the hold, `reset(141)` clears it). That's the invariant most likely to be lost to a later simplification, and it's now locked down.

The gap is that `TestTryRecoverFromTxValidationErrorReplayFallbackStopsNonConvergingRewinds` builds `parent(100) -> replay1(120) -> ledgerTip(140) -> failing(162)` all present in the chain, so the chain tip is *ahead* of the ledger tip — the opposite of the state the issue documents, and the reason the blocking finding isn't caught.

## Docs

`ARCHITECTURE.md` updated accurately, including why failure identity isn't the signal and how the neutral resync routes through node composition to Ouroboros. `DATABASE.md` checked and not affected.

## Process

CLAUDE.md requires DevNet for consensus and chain-selection paths; this changes recovery behaviour and forces chainsync reconnections. The description doesn't mention a run. Given the issue says a wedged preview data directory is preserved, resuming *that* directory on this build is the most direct validation available — and it would immediately show whether the blocking finding fires.

ledger/replay_recovery.go
```diff
@@ -203,22 +236,33 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+			"primary_chain_tip_slot", ls.chain.Tip().Point.Slot,
+			"primary_chain_already_held", primaryChainAlreadyHeld,
+			"no_progress_count", ls.replayRecoveryNoProgressCount,
+			"fallback_rollback_slot", candidate.RollbackPoint.Slot,
+			"hint", "forcing a fresh chainsync intersection; persistent failures may require operator intervention",
+		)
+		// Peer rotation is the escape mechanism for a held recovery. Publish
+		// it before the rollback calls so an unexpected local rollback error
+		// cannot suppress the fresh ChainSync intersection.
+		ls.publishReplayRecoveryNonConvergingResync(rewindPoint)
```
chrisguiney, line 261: This resolves the blocking finding from my previous review, and I verified it rather than taking it on inspection.

I reverted both production hunks — this publish back to its old position after the rollbacks, and the `!primaryChainAlreadyHeld` guard on `:263` — then re-ran the regression test:

```
--- FAIL: TestTryRecoverFromTxValidationErrorReplayFallbackStopsNonConvergingRewinds
           rewind primary chain for replay recovery:
           primary chain tip index 2 is behind replay recovery target 3
```

That's the `tipIndex < targetIndex` guard at `:382`, which is the second of the two failure modes I described. My original probe surfaced it as `lookup replay recovery target: block not found` because my fixture never seeded the block store; the fixture in this PR seeds it and lands on the index check instead. Same root cause, same consequence — and the test fails without the fix and passes with it, so it's genuinely load-bearing rather than decorative.

Worth noting the ordering has a second benefit beyond the stated one: because `EventBus.Publish` waits on its subscribers, the connection teardown now completes *before* the chain is rolled back, so a fresh intersection can't be negotiated against a chain that is about to move underneath it.

ledger/replay_recovery_test.go
```diff
@@ -1417,6 +1417,214 @@ func TestTryRecoverFromTxValidationErrorFallsBackToSecurityParamWindow(
...
+		t,
+		cm.PrimaryChain().AddRawBlocks(
+			[]chain.RawBlock{
+				parentBlock,
+				replayOneBlock,
+			},
+		),
+	)
+	// Preserve the replay candidates in the block store while leaving the
+	// primary chain tip below the applied ledger tip. This is the #3005
```
chrisguiney, line 1470: This is the part of the change I'd point a future reader at. Seeding `ledgerTipBlock` and `failingBlock` into the block store while leaving them out of the chain is what makes this the real #3005 topology rather than an approximation of it — replay rebuilt to its old high-water mark from the store, while an earlier recovery rewind had already moved the primary chain back.

The old fixture put all four blocks in the chain, which left the chain tip *ahead* of the ledger tip — the opposite of the reported state, and exactly why the bug slipped through. Changing the tip assertion from `ledgerTip` to `replayOneBlock` is the other half: it pins that the chain is deliberately left alone rather than merely happening not to move.

My revert experiment confirms this fixture catches the regression on its own.

ledger/replay_recovery.go
```diff
@@ -188,8 +197,32 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+	primaryChainAlreadyHeld := false
+	var ledgerTip ochainsync.Tip
+	if rewindPrimaryChain {
+		ls.RLock()
+		ledgerTip = ls.currentTip
+		ls.RUnlock()
+		replayHolding = ls.observeReplayRecoveryTip(ledgerTip.Point.Slot)
+		if replayHolding {
+			rewindPoint = ledgerTip.Point
+			primaryChainAlreadyHeld = ls.chain.Tip().Point.Slot < rewindPoint.Slot
```
chrisguiney, line 223: Two small things about this guard, neither blocking.

**The two tips are read separately.** `ls.currentTip` comes from under `RLock` at `:217-219`; `ls.chain.Tip()` is read here with no lock held. A chain advance in between, or an equal-slot/different-hash fork, leaves this false and the rewind can still error at `:373-382`.

I read that as accepted scope rather than missed — the comment on `:258-260` frames the goal as ensuring a rollback error *cannot suppress* the resync, not that rollback errors are eliminated, and the escape now fires regardless. If you do want to close it, a `database.BlockByPoint` presence check (or comparing the hash, not just the slot) would make the no-op decision match the helper's actual precondition instead of approximating it.

**`ls.chain.Tip()` is called twice** — here for the decision, and again at `:252` for the `primary_chain_tip_slot` log field. They can return different values, so the log line can disagree with the decision it exists to explain. Capturing it into a local once would keep the diagnostic honest, which matters more than usual on a change whose whole purpose is making a wedged node legible to an operator.

ledger/replay_recovery.go
```diff
@@ -239,6 +283,69 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+	if ls.config.EventBus == nil {
+		return
+	}
+	var activeConnId ouroboros.ConnectionId
+	if ls.config.GetActiveConnectionFunc != nil {
+		if connId := ls.config.GetActiveConnectionFunc(); connId != nil {
+			activeConnId = *connId
+		}
+	}
+	ls.config.EventBus.Publish(
```
chrisguiney, line 335: Carrying forward the one point from my earlier review that isn't addressed, and it's slightly more relevant now than it was.

There's no throttle here and no give-up: while held, this publishes on every recovery cycle. Because the publish moved ahead of the rollbacks, it now also fires on cycles that end in a rollback error — so it fires at least as often as before, never less. A permanently wedged node rotates peers indefinitely at `Warn` with no escalation.

That is the intended escape mechanism and it matches what the at-tip path does, so I'd call it consistent rather than wrong — which is why I'm not blocking on it. But an indefinitely-held node is exactly the state an operator needs to be told about loudly, and `dingo_ledger_replay_recovery_nonconverging_total` is already there to threshold against. Worth a deliberate decision on escalating to a throttled operator `Error` after N held cycles, the way the over-K rollback detector does.

chrisguiney (approved):
Approving. Re-review of `0f698140`; my earlier review was against `9c3af36b`. **All five findings are addressed**, and the blocking one was fixed with both a code change and a regression test I confirmed is load-bearing.

| Earlier finding | Status |
|---|---|
| **Blocking:** rewind errors in the #3005 state, suppressing the peer-rotation escape | **Fixed** — resync published before both rollbacks, plus an already-held guard; regression test verified by reverting the fix (see inline on `:261`) |
| Resync sits after both rollback calls, skipped on any error | **Fixed** — same change |
| `Closes #3005` overclaims (implements 1 of the issue's 3 fix directions) | **Fixed** — now `Refs #3005`, and the description states plainly that it does not claim to fix the unresolved-producer root cause or change `reachedTip` robustness |
| Mithril reordering not mentioned in the description | **Fixed** — now an explicit bullet |
| Credit: two genuinely different convergence signals, no copy-paste | Unchanged, still holds |

The hand-written description also replaced the auto-generated body and lists what was run.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go test -race -count=1 ./ledger/... ./ouroboros/... ./event/...` | **all pass** (11 packages) |
| Fix-revert experiment | **new test fails without the fix** — output inline on `:261` |

## Performance

No degradation. The added work on the held path is one extra `ls.chain.Tip()` call; nothing touches a per-block hot path. Where the guard trips the run does strictly *less* work than before — it skips `rollbackPrimaryChainInSecurityParamWindows` and its `BlockByIndex` walk entirely.

One thing worth knowing rather than changing: `EventBus.Publish` waits on subscribers (bounded by `RemoteDeliverTimeout`, 5s), and `tryRecoverFromTxValidationError` runs inside an open database transaction (`state.go:3892`, within `ledgerProcessBlocksFromSource`). That was already true before this change, so moving the publish earlier reorders work inside the same transaction rather than extending it — and it means connection teardown now completes before the chain moves, which avoids negotiating a fresh intersection against a chain that is about to be rolled back.

## Documentation

`ARCHITECTURE.md` updated accurately for both changes — it states the resync is published before local rollback work so a rollback error cannot suppress peer rotation, and that an already-pruned primary chain makes the rewind an already-held no-op while ChainSync refills it. `DATABASE.md` checked and not affected.

Three minor notes inline, none blocking.

Issue #2489, opened 2026-06-03
blockfrost: align pools/extended response with OpenAPI 0.1.90


wolf31o2:
## Summary

Align `GET /pools/extended` with the current Blockfrost OpenAPI schema.

Source of truth: Blockfrost OpenAPI `0.1.90`, schema `pool_list_extended`.

## Gaps

Dingo currently routes `/pools/extended`, but its response shape differs from the OpenAPI schema:

- Missing `blocks_minted`.
- Missing `live_saturation`.
- Missing `metadata`.
- Includes `relays`, which is not part of the current `pool_list_extended` schema.

OpenAPI requires `metadata` to be present and nullable. When present, it contains nullable URL/hash/ticker/name/description/homepage fields. It may also contain an `error` object when off-chain metadata could not be fetched or validated:

- `code`: `HASH_MISMATCH`, `CONNECTION_ERROR`, `HTTP_RESPONSE_ERROR`, `DECODE_ERROR`, `SIZE_EXCEEDED`, or `UNKNOWN_ERROR`
- `message`: human-readable error detail

## Acceptance Criteria

- `/pools/extended` emits the OpenAPI-required fields: `pool_id`, `hex`, `active_stake`, `live_stake`, `blocks_minted`, `live_saturation`, `declared_pledge`, `margin_cost`, `fixed_cost`, and `metadata`.
- Pool metadata is sourced from persisted off-chain metadata state and returned as `null` when unavailable.
- Metadata fetch/validation failures expose the persisted schema-compatible `error` object; the request path does not fetch metadata synchronously.
- Live saturation and blocks minted are computed consistently with Dingo's ledger/database state.
- Metadata behavior and field semantics are shared with the pool metadata endpoint tracked in #2936.
- Tests assert response shape, pagination/order, metadata success/absence/failure, and database errors.
- Assess `DATABASE.md` and `ARCHITECTURE.md` under the repository documentation requirements and update them if the implementation changes a documented query surface or cross-component flow.

## Related

- Pool detail, metadata, and retiring endpoints: #2936

PR #3014, opened 2026-07-29
feat(blockfrost): add pool detail endpoint

### LINKED ISSUE #2936: blockfrost: implement retiring pool, pool detail, and pool metadata endpoints

wolf31o2:
Closes #2936

Registers `GET /api/v0/pools/{pool_id}`, the last of the three endpoints #2936 asked for. `/pools/retiring` and `/pools/{pool_id}/metadata` already landed in #2969, so this completes the issue.

Accepts bech32 or hex pool IDs via the existing `parsePoolID`, so both forms resolve to the same pool. A route-ordering test asserts `/pools/retiring` and `/pools/extended` still reach their own handlers now that `{pool_id}` is registered.

## Field sourcing

| Field | Source |
|---|---|
| pool_id, hex, vrf_key | pool row, bech32 via `lcommon.PoolId` |
| declared_pledge, fixed_cost, margin_cost, reward_account | denormalized `Pool` columns, same convention as `PoolsExtended` |
| owners | latest registrations `Owners`, bech32 via existing `stakeAddressFromCredential` |
| live_stake, live_delegators | `GetStakeByPools` (its second return was previously discarded) |
| active_stake | `GetPoolStakeSnapshotsByEpoch` mark snapshot, matching `PoolsExtended` and `Network` |
| live_size, active_size, live_saturation | new `poolSizeSaturation` pure function over those totals plus `nOpt` |
| live_pledge | `GetPoolOwnerStakeAtSlot` over owner credentials |
| registration, retirement | new `GetPoolCertificateHistory` query, hex tx hashes, chronological |
| blocks_minted, blocks_epoch | existing `CountPoolBlocksInSlotRange` over `pool_opcert_sequence` |
| calidus_key | always null, see below |

No new table, index, or migration was needed.

## blocks_minted uses an existing lifetime counter

`pool_opcert_sequence` is written once per forged block (`FirstOrCreate` against the unique `(pool_key_hash, slot)` index), so it is a genuine lifetime per-pool block count rather than a per-epoch snapshot. `blocks_minted` widens the existing `CountPoolBlocksInSlotRange` to the pools full history and `blocks_epoch` narrows it to the current epochs slot range, resolved via `Database.GetEpoch` rather than the heavier hardfork-summary path.

The no-upper-bound slot sentinel is `math.MaxInt64`, not `math.MaxUint64`: slot columns bind as signed 64-bit integers through `database/sql` and a `uint64` with the high bit set is rejected by the driver.

### Known limitation, documented

`blocks_minted` undercounts on a Mithril-bootstrapped node. `pool_opcert_sequence` is written only by the live block-apply path (`UpdatePoolOpCertSequence`, called from `ledger/state.go`); `ledgerstate.ImportLedgerState` seeds pool registration state via `ImportPool` but never writes this table. So pre-snapshot history is permanently absent, not merely delayed. This is called out in `DATABASE.md` next to the query description, alongside the existing `account` table Mithril-boundary caveat. Flagging it rather than papering over it, since it affects exactly the nodes most likely to serve API traffic.

## live_saturation fails rather than guessing

`live_saturation` is a required, non-nullable `number` in the OpenAPI `pool` schema, and `calidus_key` is the only nullable field in the whole object. There is therefore no schema-compatible way to encode "unknown", and `0.0` is a value a real pool can legitimately have. So when `CurrentProtocolParams()` is unavailable (protocol parameters are not loaded until after `LedgerState.Start()`), the request errors instead of emitting a fabricated `0.0` that a client could not distinguish from a genuinely unsaturated pool.

An earlier revision of this branch did substitute `nOpt = 0` in that case. That was wrong for the reason above and has been removed; `poolSizeSaturation`s `nOpt > 0` check is now purely a divide-by-zero guard and is documented as such. The adapter test was changed to start a real `LedgerState` against the embedded devnet genesis (`nOpt = 100`) so it asserts real end-to-end saturation math, rather than keeping the test cheap at the responses expense. A separate test pins the error path.

## calidus_key

Always null. There is no reference to Calidus or CIP-0088 anywhere in the repository, so dingo ingests no Calidus key registrations from any source. `nullable: true` in the schema makes null the correct representation, and this is a real absence rather than a placeholder for an aggregate that exists.

## Tests

Handler tests, route-ordering tests, pure-function saturation tests, and real-sqlite-backed adapter tests covering both ID formats, missing pools, database failures, and the protocol-parameters error path. Additions to `sqlite/pool_test.go` are new cases only; the diff also shows gofmt comment realignment in an existing block, with no assertion or expected value changed.

## Verification

`go build ./...` (default and `-tags dingo_extra_plugins`), `go test ./api/blockfrost/... ./database/...`, `golangci-lint run`, `nilaway`, `modernize`, `make import-boundaries`, `go test -race ./api/blockfrost/...`, and `go test ./internal/test/conformance/...` all pass. Verified to merge cleanly with #3012 and to build and pass tests in combination with it.

## Documentation

`DATABASE.md` gains a `GetPoolCertificateHistory` section plus the `blocks_minted` Mithril caveat. `ARCHITECTURE.md` adds pool detail to the Blockfrost router inventory.

---
## Summary by cubic
Adds `GET /api/v0/pools/{pool_id}` for Blockfrost pool detail, completing #2936. Accepts bech32 or hex IDs and preserves `/pools/retiring` and `/pools/extended` routing.

- **New Features**
  - Returns identifiers, VRF key, pledge/margin/fixed cost, reward account, and owners from the latest registration.
  - Live stake/delegators via `GetStakeByPools`; active stake from the Mark snapshot at `currentEpoch-2`; computes `live_size`, `active_size`, and `live_saturation` using network totals and `nOpt` (circulating-supply based).
  - Lifetime and current-epoch block counts via `CountPoolBlocksInSlotRange`; registration/retirement tx hashes via shared `GetPoolCertificateHistory`; `calidus_key` is always null.

- **Bug Fixes**
  - Corrects `live_saturation` to divide by circulating supply (`MaxLovelaceSupply - Reserves`) and errors if reserves exceed max supply.
  - Fails the request when protocol params are unavailable, when no Mark snapshot exists for the target epoch, or when the current-epoch row is missing.
  - Scopes live-stake reads to the handler’s transaction for consistent snapshots; replaces full-network scans with targeted `GetPoolStakeSnapshot`; deduplicates `GetPoolCertificateHistory` into a shared internal query.

<sup>Written for commit 48354a68f281abc8861cd3f427a23b35929f3682. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3014?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added Blockfrost-compatible pool detail endpoint: `GET /api/v0/pools/{pool_id}`.
  - Pool details now include stake, pledge, delegation, saturation, block production, owner, and registration/retirement history data.
  - Supports pool identifiers provided in hexadecimal or bech32 format.
  - Added clear responses for invalid, missing, and unavailable pool data.

- **Documentation**
  - Updated architecture and database documentation to describe pool detail support and certificate history behavior.

### DIFF
--- a/api/blockfrost/adapter_pool_detail.go
+++ b/api/blockfrost/adapter_pool_detail.go
@@ -168,18 +168,21 @@ func (a *NodeAdapter) PoolDetail(poolID string) (PoolDetailInfo, error) {
 		return PoolDetailInfo{}, err
 	}
 
-	// nOpt drives live_saturation. Protocol parameters may not be loaded
-	// yet early in a node's life (before the first epoch boundary); that is
-	// treated as nOpt = 0 (no saturation figure) rather than failing pool
-	// detail entirely, the same way other pool queries in this package
-	// degrade when startup-dependent state isn't ready yet (see
-	// GetActivePoolRelays and GetPool's retirement check).
-	nOpt := 0
-	if protocolParams, ppErr := a.CurrentProtocolParams(); ppErr == nil {
-		nOpt = protocolParams.NOpt
+	// nOpt drives live_saturation, which is a required, non-nullable float
+	// in the OpenAPI schema: there is no schema-compatible way to signal
+	// "unknown" for it, and 0.0 is a legitimate saturation value (a pool
+	// with no live stake), so it cannot double as a placeholder for
+	// "protocol parameters aren't loaded yet". Propagate the error instead
+	// of guessing.
+	protocolParams, err := a.CurrentProtocolParams()
+	if err != nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"get protocol parameters: %w", err,
+		)
 	}
 	liveSize, activeSize, liveSaturation := poolSizeSaturation(
-		liveStake, activeStake, totalLiveStake, totalActiveStake, nOpt,
+		liveStake, activeStake, totalLiveStake, totalActiveStake,
+		protocolParams.NOpt,
 	)
 
 	// Live pledge: the live stake currently delegated to this pool by its
@@ -286,9 +289,11 @@ func (a *NodeAdapter) PoolDetail(poolID string) (PoolDetailInfo, error) {
 // protocol parameter, matching Blockfrost's live_size, active_size, and
 // live_saturation semantics: live_size and active_size are the pool's share
 // of total live/active stake, and live_saturation is live stake relative to
-// the per-pool saturation threshold (total active stake / nOpt). Each ratio
-// is zero when its denominator is zero (network totals not yet available,
-// or nOpt not yet configured) rather than dividing by zero.
+// the per-pool saturation threshold (total active stake / nOpt). The caller
+// is responsible for nOpt being a real, loaded protocol parameter value —
+// this function only guards the pathological case of a zero denominator
+// (e.g. a network with no active stake yet captured) to avoid dividing by
+// zero; it does not stand in for "nOpt unavailable".
 func poolSizeSaturation(
 	liveStake uint64,
 	activeStake uint64,
--- a/api/blockfrost/adapter_pool_detail.go
+++ b/api/blockfrost/adapter_pool_detail.go
@@ -16,6 +16,7 @@ package blockfrost
 
 import (
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"math"
 	"strconv"
@@ -131,30 +132,33 @@ func (a *NodeAdapter) PoolDetail(poolID string) (PoolDetailInfo, error) {
 	liveDelegators := delegatorsByPool[string(pool.PoolKeyHash)]
 
 	// Active stake: the Mark snapshot at currentEpoch-2, matching
-	// PoolsExtended and Network.
+	// PoolsExtended and Network. A single targeted row lookup rather than
+	// GetPoolStakeSnapshotsByEpoch's full-network fetch-then-linear-scan:
+	// on mainnet that call returns ~3,000 pools' rows to find the one that
+	// matches.
 	currentEpoch := a.ledgerState.CurrentEpoch()
 	activeStakeEpoch := uint64(0)
 	if currentEpoch >= 2 {
 		activeStakeEpoch = currentEpoch - 2
 	}
-	snapshots, err := db.Metadata().GetPoolStakeSnapshotsByEpoch(
-		activeStakeEpoch, "mark", txn.Metadata(),
+	snapshot, err := db.Metadata().GetPoolStakeSnapshot(
+		activeStakeEpoch, "mark", pool.PoolKeyHash, txn.Metadata(),
 	)
 	if err != nil {
 		return PoolDetailInfo{}, fmt.Errorf(
-			"get pool stake snapshots for epoch %d: %w",
+			"get pool stake snapshot for epoch %d: %w",
 			activeStakeEpoch, err,
 		)
 	}
 	var activeStake uint64
-	for _, snapshot := range snapshots {
-		if string(snapshot.PoolKeyHash) == string(pool.PoolKeyHash) {
-			activeStake = uint64(snapshot.TotalStake)
-			break
-		}
+	if snapshot != nil {
+		activeStake = uint64(snapshot.TotalStake)
 	}
 
-	// Network-wide totals for the live/active size ratios and saturation.
+	// Network-wide total active stake, for active_size. GetTotalActiveStake
+	// is already a targeted aggregate (an EpochSummary fast path, falling
+	// back to a single SUM query), not a per-pool scan, so nothing further
+	// to avoid here.
 	totalActiveStake, err := db.Metadata().GetTotalActiveStake(
 		activeStakeEpoch, "mark", txn.Metadata(),
 	)
@@ -163,7 +167,31 @@ func (a *NodeAdapter) PoolDetail(poolID string) (PoolDetailInfo, error) {
 			"get total active stake: %w", err,
 		)
 	}
-	totalLiveStake, err := a.liveStake()
+	// active_size is a required, non-nullable float with no schema-
+	// compatible "unknown" placeholder — the same constraint documented
+	// on live_saturation below. 0.0 is a value a real pool can
+	// legitimately have (no active stake), so it cannot double as a
+	// placeholder for "no snapshot captured for this epoch yet".
+	// activeStakeEpoch floors to 0 for currentEpoch < 2, and any node
+	// missing that epoch's snapshot hits this through the normal path,
+	// not a pathological one. Error instead of silently reporting every
+	// pool as 0% active-saturated, consistent with the protocol-parameters
+	// case below.
+	if totalActiveStake == 0 {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"get total active stake for epoch %d: no snapshot captured",
+			activeStakeEpoch,
+		)
+	}
+
+	// Network-wide total live stake, for live_size. Unlike active stake's
+	// total, there is no pre-aggregated column to query directly here, so
+	// this genuinely requires summing every pool's delegated stake — the
+	// same cost PoolsExtended and Network already pay for the same
+	// number. liveStake reads within txn so this stays in the same
+	// snapshot as the reads above (see liveStake's doc comment for the
+	// one read that still can't).
+	totalLiveStake, err := a.liveStake(txn.Metadata())
 	if err != nil {
 		return PoolDetailInfo{}, err
 	}
@@ -174,15 +202,33 @@ func (a *NodeAdapter) PoolDetail(poolID string) (PoolDetailInfo, error) {
 	// with no live stake), so it cannot double as a placeholder for
 	// "protocol parameters aren't loaded yet". Propagate the error instead
 	// of guessing.
+	//
+	// CurrentProtocolParams reads the ledger state's own cached current
+	// parameters rather than a value scoped to txn — the same convention
+	// PoolsExtended and Network already rely on for their own out-of-txn
+	// reads (see liveStake). It matters less here in practice than it
+	// would for a live-stake read: protocol parameters only change at
+	// epoch boundaries via governance action, not every block, so the
+	// window in which this could disagree with the rest of this
+	// snapshot is far narrower.
 	protocolParams, err := a.CurrentProtocolParams()
 	if err != nil {
 		return PoolDetailInfo{}, fmt.Errorf(
 			"get protocol parameters: %w", err,
 		)
 	}
+
+	// live_saturation's denominator is total circulating supply, not
+	// total active stake: see totalCirculation's doc comment.
+	totalCirculation, err := a.totalCirculation(txn.Metadata())
+	if err != nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"get total circulation: %w", err,
+		)
+	}
 	liveSize, activeSize, liveSaturation := poolSizeSaturation(
 		liveStake, activeStake, totalLiveStake, totalActiveStake,
-		protocolParams.NOpt,
+		totalCirculation, protocolParams.NOpt,
 	)
 
 	// Live pledge: the live stake currently delegated to this pool by its
@@ -223,18 +269,30 @@ func (a *NodeAdapter) PoolDetail(poolID string) (PoolDetailInfo, error) {
 		)
 	}
 
-	epochStartSlot := uint64(0)
+	// blocks_epoch narrows the lifetime query above to the current epoch's
+	// slot range. If GetEpoch returns nil for the current epoch, the state
+	// needed to do that narrowing isn't there: defaulting epochStartSlot
+	// to 0 would silently reproduce the blocks_minted query above
+	// (0..noSlotUpperBound) and report the pool's entire history as
+	// current-epoch blocks, so this errors instead.
+	//
+	// The upper bound is noSlotUpperBound, not the epoch's end slot — only
+	// correct because this is always the CURRENT epoch, where no blocks
+	// exist past the synced tip. Reusing this for a historical epoch would
+	// need the epoch's actual end slot as the upper bound instead.
 	epochRow, err := db.GetEpoch(currentEpoch, txn)
 	if err != nil {
 		return PoolDetailInfo{}, fmt.Errorf(
 			"get epoch %d: %w", currentEpoch, err,
 		)
 	}
-	if epochRow != nil {
-		epochStartSlot = epochRow.StartSlot
+	if epochRow == nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"get epoch %d: no epoch row found", currentEpoch,
+		)
 	}
 	blocksEpochByPool, _, err := db.Metadata().CountPoolBlocksInSlotRange(
-		[]lcommon.PoolKeyHash{pkh}, epochStartSlot, noSlotUpperBound, txn.Metadata(),
+		[]lcommon.PoolKeyHash{pkh}, epochRow.StartSlot, noSlotUpperBound, txn.Metadata(),
 	)
 	if err != nil {
 		return PoolDetailInfo{}, fmt.Errorf(
@@ -284,21 +342,65 @@ func (a *NodeAdapter) PoolDetail(poolID string) (PoolDetailInfo, error) {
 	}, nil
 }
 
+// totalCirculation returns MaxLovelaceSupply minus Reserves: the sigma
+// denominator the reward calculation uses for the per-pool saturation
+// threshold. ledger/rewards/rewards.go:287 computes exactly this quantity
+// (`totalCirculation := params.MaxLovelaceSupply - pots.Reserves`), and
+// rewards.go:~1028-1034 passes it as totalStake into
+// optimalPoolRewardChecked, where sigma = poolStake/totalStake is capped at
+// z0 = 1/optimalPoolCount — a deliberately different denominator from
+// totalActiveStake, which the same call passes to apparentPerformance one
+// line above.
+//
+// This is NOT the same figure as Network()'s local "circulating" value
+// (adapter.go), which further subtracts treasury and script-locked supply
+// for display purposes. Callers computing a saturation threshold must use
+// this function's result, never that one.
+func (a *NodeAdapter) totalCirculation(txn dbtypes.Txn) (uint64, error) {
+	nodeCfg := a.ledgerState.CardanoNodeConfig()
+	if nodeCfg == nil || nodeCfg.ShelleyGenesis() == nil {
+		return 0, errors.New("shelley genesis not available")
+	}
+	maxSupply := nodeCfg.ShelleyGenesis().MaxLovelaceSupply
+
+	state, err := a.ledgerState.Database().Metadata().GetNetworkState(txn)
+	if err != nil {
+		return 0, fmt.Errorf("get network state: %w", err)
+	}
+	reserves := uint64(0)
+	if state != nil {
+		reserves = uint64(state.Reserves)
+	}
+	if reserves > maxSupply {
+		return 0, nil
+	}
+	return maxSupply - reserves, nil
+}
+
 // poolSizeSaturation computes a pool's live/active stake-size ratios and
-// live-saturation fraction relative to network-wide totals and the nOpt (k)
-// protocol parameter, matching Blockfrost's live_size, active_size, and
-// live_saturation semantics: live_size and active_size are the pool's share
-// of total live/active stake, and live_saturation is live stake relative to
-// the per-pool saturation threshold (total active stake / nOpt). The caller
-// is responsible for nOpt being a real, loaded protocol parameter value —
-// this function only guards the pathological case of a zero denominator
-// (e.g. a network with no active stake yet captured) to avoid dividing by
-// zero; it does not stand in for "nOpt unavailable".
+// live-saturation fraction, matching Blockfrost's live_size, active_size,
+// and live_saturation semantics. The three outputs each divide by a
+// different network-wide total:
+//   - live_size divides by totalLiveStake, the network's total delegated
+//     (live) stake.
+//   - active_size divides by totalActiveStake, the network's total Mark-
+//     snapshot (active) stake for the relevant epoch.
+//   - live_saturation divides live stake by the per-pool saturation
+//     threshold, totalCirculation / nOpt. totalCirculation is deliberately
+//     NOT totalActiveStake: see the totalCirculation doc comment above for
+//     why the reward calculation requires this distinction.
+//
+// The caller is responsible for nOpt being a real, loaded protocol
+// parameter value — this function only guards the pathological case of a
+// zero denominator (e.g. a network with no active stake or circulation
+// figure yet captured) to avoid dividing by zero; it does not stand in for
+// "nOpt unavailable" or "totalCirculation unavailable".
 func poolSizeSaturation(
 	liveStake uint64,
 	activeStake uint64,
 	totalLiveStake uint64,
 	totalActiveStake uint64,
+	totalCirculation uint64,
 	nOpt int,
 ) (liveSize float64, activeSize float64, liveSaturation float64) {
 	if totalLiveStake > 0 {
@@ -306,11 +408,11 @@ func poolSizeSaturation(
 	}
 	if totalActiveStake > 0 {
 		activeSize = float64(activeStake) / float64(totalActiveStake)
-		if nOpt > 0 {
-			saturationThreshold := float64(totalActiveStake) / float64(nOpt)
-			if saturationThreshold > 0 {
-				liveSaturation = float64(liveStake) / saturationThreshold
-			}
+	}
+	if totalCirculation > 0 && nOpt > 0 {
+		saturationThreshold := float64(totalCirculation) / float64(nOpt)
+		if saturationThreshold > 0 {
+			liveSaturation = float64(liveStake) / saturationThreshold
 		}
 	}
 	return liveSize, activeSize, liveSaturation
--- a/api/blockfrost/adapter_pool_detail.go
+++ b/api/blockfrost/adapter_pool_detail.go
@@ -1,0 +1,312 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package blockfrost
+
+import (
+	"encoding/hex"
+	"fmt"
+	"math"
+	"strconv"
+
+	"github.com/blinklabs-io/dingo/database/models"
+	dbtypes "github.com/blinklabs-io/dingo/database/types"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+)
+
+// noSlotUpperBound stands in for "no upper bound" in slot-range queries
+// that want everything synced so far. math.MaxInt64, not math.MaxUint64:
+// slot columns are bound through database/sql as signed 64-bit integers,
+// and a uint64 value with the high bit set (like math.MaxUint64) is
+// rejected by the sqlite driver. MaxInt64 slots are billions of years past
+// any real chain tip, so it is effectively unbounded for this purpose.
+const noSlotUpperBound = uint64(math.MaxInt64)
+
+// PoolDetail returns the OpenAPI pool detail object for the requested pool
+// (bech32 or hex ID). Epoch-sensitive aggregates (active stake, saturation,
+// blocks_epoch) are computed for the current epoch.
+func (a *NodeAdapter) PoolDetail(poolID string) (PoolDetailInfo, error) {
+	poolKeyHash, err := parsePoolID(poolID)
+	if err != nil {
+		return PoolDetailInfo{}, err
+	}
+	pkh := lcommon.PoolKeyHash(poolKeyHash)
+
+	db := a.ledgerState.Database()
+	txn := db.Transaction(false)
+	defer txn.Release()
+
+	pool, err := db.Metadata().GetPool(pkh, true, txn.Metadata())
+	if err != nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"get pool %x: %w", poolKeyHash, err,
+		)
+	}
+	if pool == nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"pool %x: %w", poolKeyHash, models.ErrPoolNotFound,
+		)
+	}
+	if len(pool.Registration) == 0 {
+		// A pool row with no registration at all should not exist in
+		// practice (GetPool preloads the latest one), but guard against it
+		// rather than indexing into an empty slice below.
+		return PoolDetailInfo{}, fmt.Errorf(
+			"pool %x has no registration: %w",
+			poolKeyHash, models.ErrPoolNotFound,
+		)
+	}
+	reg := pool.Registration[0]
+
+	networkID := a.networkID()
+
+	rewardCredType := uint(lcommon.CredentialTypeAddrKeyHash)
+	if pool.RewardAccountCredentialTag == 1 {
+		rewardCredType = uint(lcommon.CredentialTypeScriptHash)
+	}
+	rewardAccount, err := stakeAddressFromCredential(
+		lcommon.Credential{
+			CredType: rewardCredType,
+			Credential: lcommon.CredentialHash(
+				lcommon.NewBlake2b224(pool.RewardAccount),
+			),
+		},
+		networkID,
+	)
+	if err != nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"encode reward account for pool %x: %w", poolKeyHash, err,
+		)
+	}
+
+	// Owners are always key-hash credentials per the pool registration
+	// certificate spec (unlike the reward account, which may be a script).
+	owners := make([]string, 0, len(reg.Owners))
+	ownerKeyHashes := make([][]byte, 0, len(reg.Owners))
+	for _, owner := range reg.Owners {
+		addr, err := stakeAddressFromCredential(
+			lcommon.Credential{
+				CredType: uint(lcommon.CredentialTypeAddrKeyHash),
+				Credential: lcommon.CredentialHash(
+					lcommon.NewBlake2b224(owner.KeyHash),
+				),
+			},
+			networkID,
+		)
+		if err != nil {
+			return PoolDetailInfo{}, fmt.Errorf(
+				"encode owner address for pool %x: %w", poolKeyHash, err,
+			)
+		}
+		owners = append(owners, addr)
+		ownerKeyHashes = append(ownerKeyHashes, owner.KeyHash)
+	}
+
+	marginCost := 0.0
+	if pool.Margin != nil && pool.Margin.Rat != nil {
+		marginCost, _ = pool.Margin.Float64()
+	}
+
+	// Live stake and delegator count.
+	liveStakeByPool, delegatorsByPool, err := db.Metadata().GetStakeByPools(
+		[][]byte{pool.PoolKeyHash}, txn.Metadata(),
+	)
+	if err != nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"get live stake for pool %x: %w", poolKeyHash, err,
+		)
+	}
+	liveStake := liveStakeByPool[string(pool.PoolKeyHash)]
+	liveDelegators := delegatorsByPool[string(pool.PoolKeyHash)]
+
+	// Active stake: the Mark snapshot at currentEpoch-2, matching
+	// PoolsExtended and Network.
+	currentEpoch := a.ledgerState.CurrentEpoch()
+	activeStakeEpoch := uint64(0)
+	if currentEpoch >= 2 {
+		activeStakeEpoch = currentEpoch - 2
+	}
+	snapshots, err := db.Metadata().GetPoolStakeSnapshotsByEpoch(
+		activeStakeEpoch, "mark", txn.Metadata(),
+	)
+	if err != nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"get pool stake snapshots for epoch %d: %w",
+			activeStakeEpoch, err,
+		)
+	}
+	var activeStake uint64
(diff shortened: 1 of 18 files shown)


api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,317 @@
...
+	totalActiveStake uint64,
+	nOpt int,
+) (liveSize float64, activeSize float64, liveSaturation float64) {
+	if totalLiveStake > 0 {
+		liveSize = float64(liveStake) / float64(totalLiveStake)
+	}
+	if totalActiveStake > 0 {
+		activeSize = float64(activeStake) / float64(totalActiveStake)
+		if nOpt > 0 {
+			saturationThreshold := float64(totalActiveStake) / float64(nOpt)
```
chrisguiney: **Blocking.** The saturation threshold is defined over **circulating supply**, not total active stake. Dingo's own reward calculation says so:

- `ledger/rewards/rewards.go:287` — `totalCirculation := params.MaxLovelaceSupply - pots.Reserves`
- `rewards.go:1028-1034` passes that as `totalStake` into `optimalPoolRewardChecked`, where `sigma = poolStake/totalStake` is capped at `z0 = 1/optimalPoolCount`

That function distinguishes the two denominators deliberately — the call immediately above it passes `totalActiveStake` to `apparentPerformance`, and the saturation one gets `totalCirculation`.

Probed with mainnet-shaped inputs (circulating ~36.1B ADA, staked ~21.5B, k=500):

```
saturation point, circulating/nOpt (correct):  72,200,000 ADA
saturation point, totalActive/nOpt (this PR):  43,000,000 ADA

live_stake= 40M ADA  PR reports 0.930  correct 0.554  (1.68x)
live_stake= 60M ADA  PR reports 1.395  correct 0.831  (1.68x)
live_stake= 72M ADA  PR reports 1.674  correct 0.997  (1.68x)
```

`live_saturation` is the field delegators use to decide whether a pool is full. A pool at 60M ADA is reported 39% over-saturated when it's actually at 83%, and pools cross the reported 1.0 line at ~43M instead of ~72M.

The right value is already computed in this file: `adapter.go:749-751` derives `total = maxSupply - reserves` from `genesis.MaxLovelaceSupply` and `GetNetworkState().Reserves`. It must be that quantity, **not** the `circulating` on the next line — that one further subtracts treasury and script-locked supply and is not what `rewards.go` uses for sigma.

`live_size` and `active_size` are correct as written; only this threshold is wrong.

api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,317 @@
...
+	liveStake uint64,
+	activeStake uint64,
+	totalLiveStake uint64,
+	totalActiveStake uint64,
+	nOpt int,
+) (liveSize float64, activeSize float64, liveSaturation float64) {
+	if totalLiveStake > 0 {
+		liveSize = float64(liveStake) / float64(totalLiveStake)
+	}
+	if totalActiveStake > 0 {
```
chrisguiney, line 417: **Blocking.** Nesting both `active_size` and `live_saturation` inside `totalActiveStake > 0` means they fall through to their zero values when there's no snapshot for `activeStakeEpoch`. Probe:

```
totalActiveStake==0, nOpt=500 valid:
  live_size=0.0028  active_size=0.0  live_saturation=0.0
```

That's the fabricated `0.0` the description rightly refuses to emit for a missing `nOpt` — *"0.0 is a value a real pool can legitimately have"* — and it's reachable through the normal path, not a pathological one: `activeStakeEpoch` floors to 0 for `currentEpoch < 2`, and any node lacking that epoch's snapshot hits it too.

The `nOpt > 0` guard is held to a stricter standard than the snapshot feeding the same formula. Fixing the comment above moves saturation off this denominator entirely; `active_size` still needs a decision — error, or accept 0.0 and say why in the doc comment.

api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	epochStartSlot := uint64(0)
+	epochRow, err := db.GetEpoch(currentEpoch, txn)
+	if err != nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"get epoch %d: %w", currentEpoch, err,
+		)
+	}
+	if epochRow != nil {
```
chrisguiney: **Blocking.** When `epochRow` is nil, `epochStartSlot` stays 0 and the query below counts from slot 0 to `noSlotUpperBound` — which is byte-for-byte the `blocks_minted` query above. So `blocks_epoch` silently reports the pool's entire history as current-epoch blocks.

Third instance of the same class as the two comments above: a required non-nullable field quietly filled with a plausible wrong number. Given how the protocol-parameters case is handled 60 lines up, this one should error too — `GetEpoch` returning nil for the *current* epoch means the state needed to answer the request isn't there.

Separately, worth noting the upper bound is `noSlotUpperBound` rather than the epoch's end slot. That's fine for the current epoch since no blocks exist past the tip, but it does mean the function can't be reused for a historical epoch without changing it.

database/plugin/metadata/sqlite/pool.go
```diff
@@ -1380,3 +1380,70 @@ func (d *MetadataStoreSqlite) GetRetiringPools(
...
+// GetPoolCertificateHistory returns the transaction hashes of a pool's
+// registration and retirement certificates, ordered chronologically
+// (added_slot, block_index, cert_index ascending). The inner joins to certs
+// and transaction naturally exclude rows with no linked transaction —
+// certificates synthesized by the Mithril ledger-state import carry
+// certificate_id = 0, which matches no certs row — since they have no
+// originating transaction to report. Both pool_registration.pool_key_hash
+// and pool_retirement.pool_key_hash are indexed, so each query is a small,
+// indexed lookup rather than a table scan.
+func (d *MetadataStoreSqlite) GetPoolCertificateHistory(
```
chrisguiney, line 1390: **Duplication.** The sqlite, postgres, and mysql copies of this function are byte-identical apart from the receiver type and `"transaction"` vs backtick-quoting. I diffed them: sqlite and postgres differ on **one line** — the receiver.

`database/plugin/metadata/internal/sqldialect.TransactionTableName(db)` exists for exactly this quoting difference, and three packages already use it to share a query across all three backends:

- `internal/accounthistory/accounthistory.go:172`
- `internal/rewardstate/dialect.go:34`
- `internal/stakequery/stakequery.go:691`

This is ~201 lines that should be ~70 plus three one-line delegations, following the `internal/certutil` / `internal/drepquery` pattern already in this directory.

To be fair about the precedent: the base `pool.go` files have genuinely diverged (945 differing lines between sqlite and postgres at the merge base), so this isn't matching existing debt — it's new exact duplication where the helper already exists. Three copies of an `Order` clause with a three-key tie-break is a good candidate for drifting the first time one backend needs a tweak.

api/blockfrost/pool_detail_test.go
```diff
@@ -0,0 +1,300 @@
...
+		{
+			name:               "normal pool",
+			liveStake:          6_900_000_000,
+			activeStake:        4_200_000_000,
+			totalLive:          16_428_571_428,
+			totalActive:        420_000_000_000,
+			nOpt:               500,
+			wantLiveSize:       6_900_000_000.0 / 16_428_571_428.0,
+			wantActiveSize:     4_200_000_000.0 / 420_000_000_000.0,
+			wantLiveSaturation: 6_900_000_000.0 / (420_000_000_000.0 / 500.0),
```
chrisguiney: This encodes the wrong formula as the expectation (`totalActive / nOpt`), so the table can't catch the finding on `poolSizeSaturation:310` — it asserts the implementation against itself.

Once the denominator is circulating supply, `poolSizeSaturation` needs a fifth input and these cases need real expected values. A mainnet-shaped case would be worth adding while you're in here: `nOpt = 500` with circulating ~36.1B ADA and staked ~21.5B, asserting a 72M-ADA pool lands at ~1.0. That's the case where the two denominators diverge by 1.68x, and it's the one shape the current table can't express.

api/blockfrost/adapter_pool_detail_db_test.go
```diff
@@ -0,0 +1,335 @@
...
+	assert.NotNil(t, info.Retirement)
+	assert.Nil(t, info.CalidusKey)
+	// Active stake and active_size: this pool is the network's only
+	// captured snapshot row, so it holds the entire active-stake total.
+	assert.Equal(t, "4200000000", info.ActiveStake)
+	assert.InDelta(t, 1.0, info.ActiveSize, 1e-9)
+	// live_saturation is computed from the real devnet nOpt (100), not a
+	// placeholder: saturation_threshold = totalActiveStake / nOpt, and
+	// live_saturation = liveStake / saturation_threshold.
+	wantSaturation := float64(wantLiveStake) / (float64(activeStake) / float64(wantNOpt))
```
chrisguiney: Same formula issue as the pure-function table, and it matters more here: this is the test that was deliberately upgraded to start a real `LedgerState` against devnet genesis so it would assert *real end-to-end saturation math* rather than a placeholder. That was the right instinct — but with `totalActiveStake / nOpt` as the expectation, it's real inputs through the wrong formula.

The comment on `:235-236` also documents the wrong rule, so it needs updating alongside.

Worth keeping the real-`LedgerState` setup once the formula is fixed: devnet genesis gives you `nOpt = 100` *and* `maxLovelaceSupply`, so the corrected denominator is fully derivable from the same fixture.

api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,317 @@
...
+	// Network-wide totals for the live/active size ratios and saturation.
+	totalActiveStake, err := db.Metadata().GetTotalActiveStake(
+		activeStakeEpoch, "mark", txn.Metadata(),
+	)
+	if err != nil {
+		return PoolDetailInfo{}, fmt.Errorf(
+			"get total active stake: %w", err,
+		)
+	}
+	totalLiveStake, err := a.liveStake()
```
chrisguiney: Two full-network aggregations to answer one pool's detail, on an endpoint clients poll:

- `a.liveStake()` calls `GetActivePoolKeyHashes` then `GetStakeByPools` for **every** pool (`adapter.go:1657-1675`)
- `GetPoolStakeSnapshotsByEpoch` on `:140` returns every pool's snapshot, then `:150-155` linear-scans with a `break` for the one match

On mainnet that's ~3,000 pools' stake aggregated per request. `GetTotalActiveStake` is already a targeted aggregate, so the snapshot scan is the more clearly avoidable of the two.

Also a consistency wrinkle: most reads here use `txn`, but `a.liveStake()` and `CurrentProtocolParams()` read outside it, so `liveStake` (in-txn) and `totalLiveStake` (out-of-txn) can straddle a block boundary and make `live_size` exceed 1.0. `PoolsExtended` does the same thing, so this is pre-existing convention rather than something the PR introduces — flagging it because a single-pool endpoint has the option of staying inside one snapshot, where the list endpoints don't.

api/blockfrost/handlers_pool_detail.go
```diff
@@ -0,0 +1,119 @@
...
+	writeJSON(w, http.StatusOK, poolDetailResponse(info))
+}
+
+// poolDetailResponse converts a PoolDetailInfo into its Blockfrost wire
+// shape. Owners, Registration, and Retirement are non-nullable arrays in
+// the OpenAPI schema, so a nil slice is normalized to an empty one rather
+// than encoding as JSON null.
+func poolDetailResponse(info PoolDetailInfo) PoolDetailResponse {
+	var calidusKey *PoolCalidusKeyResponse
+	if info.CalidusKey != nil {
```
chrisguiney: This branch can't execute: the adapter hardcodes `CalidusKey: nil` (`adapter_pool_detail.go:283`), so the conversion is unreachable and untestable.

Keeping `PoolCalidusKeyResponse` is right — it documents the wire shape the schema requires, and the description's reasoning for `null` is sound. It's the 13-line conversion that has no reachable caller. Either drop it and assign `nil` directly with the reasoning in a comment, or state in the comment that it's unreachable so the next reader doesn't go hunting for the producer.

api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	// Lifetime blocks minted: every observed op-cert-sequence row for this
+	// pool. That table is written once per block the pool has produced and
+	// is never pruned by epoch (see UpdatePoolOpCertSequence in
+	// database/pool.go and DATABASE.md's pool_opcert_sequence entry), so a
+	// full-range count is a genuine lifetime total, not an approximation.
+	// blocks_epoch narrows the same indexed query to the current epoch's
+	// slot range.
+	blocksMintedByPool, _, err := db.Metadata().CountPoolBlocksInSlotRange(
```
chrisguiney, line 263: Crediting this because it's the claim I most expected to fall apart and it didn't. "Count rows in a table that happens to have one row per block" usually means counting something adjacent to blocks — opcert rotations, or only locally forged blocks.

It checks out:

- `UpdatePoolOpCertSequence` has exactly **one** call site, `ledgerProcessBlock` (`ledger/state.go:4454`), which runs for every block applied to the chain from any pool — not just locally forged ones
- the write is inside the block's transaction, so a rejected block doesn't leave a row
- the only `Delete` on `PoolOpCertSequence` is in `RestorePoolStateAtSlot` (`slot > ?`), so the count is rollback-aware and there is no epoch pruning, exactly as the comment says

So `blocks_minted` is a genuine lifetime total on a from-genesis node, and the Mithril undercount is a real, permanent gap rather than a delay. Documenting it in `DATABASE.md` next to the query, instead of leaving it to be discovered, is the right call.

chrisguiney (changes requested):
Requesting changes on one item: `live_saturation` divides by total active stake instead of circulating supply, which overstates it by ~1.68x on mainnet-shaped inputs. Dingo's own reward calculation defines that denominator, so this is checkable in-repo rather than a matter of interpretation. Two smaller findings are the same class of defect the description explicitly sets out to avoid.

The endpoint is otherwise carefully built, and most of the description's claims hold up when checked — including the ones I went looking to disprove. Details inline.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...` and `-tags dingo_extra_plugins` | clean |
| `go test -race -count=1 ./api/blockfrost/...` | pass |
| `go test -count=1 ./database/...` | **all pass** (16 packages) |
| Saturation probe (written for this review) | **1.68x overstatement, confirmed** |
| Blockfrost `pool.yaml` (fetched) | schema claims confirmed |

DevNet isn't required here — read-only API path, nothing in CLAUDE.md's consensus list.

## Claims I checked that hold up

- **`blocks_minted` really is a lifetime count.** `UpdatePoolOpCertSequence` has exactly one call site, `ledgerProcessBlock` (`ledger/state.go:4454`), which runs for every applied block from any pool, and the only `Delete` on `PoolOpCertSequence` is in `RestorePoolStateAtSlot` (`slot > ?`). So it's rollback-aware and never epoch-pruned, as the comment claims. The Mithril undercount is real and documented rather than papered over.
- **`currentEpoch-2` + `"mark"` matches `Network`** (`adapter.go:759-765`) **and `PoolsExtended`** (`adapter.go:1937-1943`).
- **`pool.Registration[0]` is genuinely the latest** — `GetPool`'s preload applies the full documented tie-break (`added_slot DESC, block_index DESC, cert_index DESC`), so owners, pledge, and cost come from the right registration.
- **The schema reasoning is right.** I fetched `pool.yaml`: every field is required and `calidus_key` is the only nullable one. So `null` for Calidus is correct, and there really is no schema-compatible way to encode "unknown" for `live_saturation`. That reasoning is sound — my objection is that it isn't applied consistently.
- **Route ordering** satisfies the acceptance criterion, and `http.ServeMux` resolves by pattern specificity, so `/pools/retiring` and `/pools/extended` can't be swallowed regardless of registration order.
- Handler error mapping (400/404/500) and nil-slice to `[]` normalization for the non-nullable arrays are both correct.
- No `nOpt = 0` substitution remains; the `nOpt > 0` check is a pure divide-by-zero guard as described.

## Documentation

`DATABASE.md` (+59) adds a `GetPoolCertificateHistory` section and the `blocks_minted` Mithril caveat, with the `certificate_id = 0` exclusion explained. `ARCHITECTURE.md` (+4) adds pool detail to the router inventory. Both accurate. Neither documents the saturation formula, so the fix below needs no doc change.

wolf31o2:
Thanks for the review, and for re-verifying the description's claims rather than taking them at face value. All six findings are addressed in 2eb1c99e.

## 1. live_saturation denominator (blocking)

Fixed. Confirmed your reading against `rewards.go`: `totalCirculation := params.MaxLovelaceSupply - pots.Reserves` (287) is what `optimalPoolRewardChecked` receives as sigma's denominator (1024-1035), while `totalActiveStake` goes to `apparentPerformance` on the line above. The two are deliberately distinct and this used the wrong one.

`poolSizeSaturation` takes a fifth input and `live_saturation` now divides by `totalCirculation / nOpt`. Added `totalCirculation(txn)` deriving `MaxLovelaceSupply - Reserves`, with a doc comment pointing at `rewards.go:287` and an explicit warning not to reuse `Network()`'s `circulating`, which further subtracts treasury and locked supply. That distinction was the easiest way to get this wrong twice, so it is called out in the code.

Reproduces your numbers exactly:

```
live_stake= 40M ADA  before=0.930  after=0.554  ratio=1.679x
live_stake= 60M ADA  before=1.395  after=0.831  ratio=1.679x
live_stake= 72M ADA  before=1.674  after=0.997  ratio=1.679x
```

`live_size` and `active_size` keep their existing denominators, as you noted.

## 2. active_size falling through to 0.0 (blocking)

Fixed by erroring, per your stated preference and consistent with the protocol-parameters precedent already on this branch. `PoolDetail` now returns an error when `totalActiveStake == 0` rather than emitting a fabricated `0.0`. You were right that this was reachable through the normal path, not a pathological one.

`poolSizeSaturation` keeps its zero-denominator guard as a pure defensive check, since the sibling #2489 branch reuses the function.

Two existing tests needed stake-snapshot fixtures added so they still exercise what they intend instead of tripping the new check first.

## 3. blocks_epoch reporting lifetime blocks (blocking)

Fixed. Errors when `GetEpoch(currentEpoch)` returns nil, instead of leaving `epochStartSlot` at 0 and silently duplicating the `blocks_minted` query. Your point about `noSlotUpperBound` is now a comment: it is only sound because this is always the current epoch, so the function needs bounding before it could serve a historical one.

## 4. Triplicate GetPoolCertificateHistory

Fixed. New `database/plugin/metadata/internal/poolcerthistory` holds the shared query using `sqldialect.TransactionTableName`; each backend is now a 17-line delegation, down from ~67. Net 168 lines removed against a 93-line shared implementation, following the `internal/certutil` / `internal/drepquery` pattern. Fair call that this was new exact duplication rather than matching the pre-existing divergence in those files.

## 5. Tests asserting the implementation against itself

Fixed. `TestPoolSizeSaturation` gained a `totalCirculation` column that is deliberately different from `totalActive` in every case, so the table cannot pass by restating the implementation. Added your mainnet-shaped case (nOpt=500, circulating 36.1B ADA, staked 21.5B) asserting a 72M-ADA pool lands at ~0.997.

The DB test keeps the real-`LedgerState` devnet-genesis setup, with `totalCirculation` computed from the same fixture's `MaxLovelaceSupply` and persisted `NetworkState.Reserves`. The stale comment documenting the wrong rule is replaced.

## 6. Full-network aggregations

Partly fixed. Replaced the fetch-all-plus-linear-scan with the existing targeted `GetPoolStakeSnapshot(epoch, type, poolKeyHash, txn)`, which removes the ~3,000-pool snapshot pass you identified as the more clearly avoidable one.

`liveStake()`'s full-network aggregation stays: there is no pre-aggregated live-stake total to query, unlike `GetTotalActiveStake`. It now takes a `txn` so `PoolDetail` reads it inside the same transaction as everything else, which closes the block-boundary straddle for `live_size` and `live_saturation` on this endpoint. `Network()` still passes nil and is unchanged. `CurrentProtocolParams()` still reads outside the transaction because it sources from the ledger state's cached params rather than a DB call, and changes only on governance enactment rather than per block; that reasoning is now a comment.

## Also

Removed the dead `PoolCalidusKeyInfo` to `PoolCalidusKeyResponse` conversion, which was unreachable because the adapter hardcodes `CalidusKey: nil`.

## Verification

`go build` (both tag sets), `go test ./api/blockfrost/... ./database/...`, `golangci-lint run` (0 issues), `nilaway`, `modernize`, `make import-boundaries`, `go test -race ./api/blockfrost/...`, and conformance all pass.

`DATABASE.md`/`ARCHITECTURE.md` were reassessed and are unaffected: the deduped query is byte-identical apart from relocation, `GetPoolStakeSnapshot` is an existing documented method, and the new error paths are adapter-level behavior rather than a change to the metadata-store query surface.

One note for ordering: the saturation bug is in `poolSizeSaturation`, which #2489 reuses for `/pools/extended`. That branch is stacked on this one and will pick up the corrected formula rather than repeating the bug.

api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,419 @@
...
+	totalCirculation uint64,
+	nOpt int,
+) (liveSize float64, activeSize float64, liveSaturation float64) {
+	if totalLiveStake > 0 {
+		liveSize = float64(liveStake) / float64(totalLiveStake)
+	}
+	if totalActiveStake > 0 {
+		activeSize = float64(activeStake) / float64(totalActiveStake)
+	}
+	if totalCirculation > 0 && nOpt > 0 {
```
chrisguiney, line 420: This resolves the blocking finding, and I verified it by breaking it rather than by reading it.

I reverted this line and the one above back to `totalActiveStake` and re-ran the tests:

```
--- FAIL: TestNodeAdapterPoolDetailFullResponse
    Max difference between 0.6611111111666667 and 283.33333335714286 …
--- FAIL: TestPoolSizeSaturation/normal_pool
--- FAIL: TestPoolSizeSaturation/zero_total_live_stake
--- FAIL: TestPoolSizeSaturation/zero_total_active_stake
```

Four failures, and the interesting ones are the first and last. The first is the real-`LedgerState` end-to-end test, so the correct formula is now asserted against actual devnet genesis and ledger state rather than only in a table. The last is `zero_total_active_stake`, which is exactly the case from my second blocking finding — so the decoupling on `:409` is pinned too, not just done.

That decoupling is the part worth noting separately: `active_size` and `live_saturation` now sit in independent guards instead of one nested inside the other, so a missing mark snapshot can no longer take `live_saturation` down with it. Those two fields have genuinely different denominators and now genuinely different preconditions.

api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,419 @@
...
+// optimalPoolRewardChecked, where sigma = poolStake/totalStake is capped at
+// z0 = 1/optimalPoolCount — a deliberately different denominator from
+// totalActiveStake, which the same call passes to apparentPerformance one
+// line above.
+//
+// This is NOT the same figure as Network()'s local "circulating" value
+// (adapter.go), which further subtracts treasury and script-locked supply
+// for display purposes. Callers computing a saturation threshold must use
+// this function's result, never that one.
+func (a *NodeAdapter) totalCirculation(txn dbtypes.Txn) (uint64, error) {
```
chrisguiney, line 359: The doc comment on this helper is the part I'd most want kept. It records the two things that made the original bug easy to write:

- that `rewards.go:287` computes exactly this quantity and `~1028-1034` passes it as `totalStake` into `optimalPoolRewardChecked`, while the same call passes `totalActiveStake` to `apparentPerformance` one line above — so the distinction is deliberate in the reward calculation, not incidental;
- that this is **not** `Network()`'s local `circulating`, which further subtracts treasury and script-locked supply for display.

That second warning is the one that earns its place. Reusing `Network()`'s figure would have been the obvious shortcut, it would have looked right, and it would have been wrong in a way no test with round fixture numbers would catch — while also dragging `GetScriptLockedSupply` onto a per-request path. Naming the wrong-but-plausible neighbour is more useful than just naming the right answer.

api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,419 @@
...
+	// PoolsExtended and Network. A single targeted row lookup rather than
+	// GetPoolStakeSnapshotsByEpoch's full-network fetch-then-linear-scan:
+	// on mainnet that call returns ~3,000 pools' rows to find the one that
+	// matches.
+	currentEpoch := a.ledgerState.CurrentEpoch()
+	activeStakeEpoch := uint64(0)
+	if currentEpoch >= 2 {
+		activeStakeEpoch = currentEpoch - 2
+	}
+	snapshot, err := db.Metadata().GetPoolStakeSnapshot(
```
chrisguiney, line 144: This is the performance half of my earlier note, and it's addressed the best available way: `GetPoolStakeSnapshot` is a **pre-existing** store method (`stake_snapshot.go` in all three plugins, untouched by this diff), so a ~3,000-row fetch-then-linear-scan became a single targeted row read *without* adding three new plugin implementations.

Worth calling out because the trap here was real. Needing a targeted query where only a bulk one is used is the standard way three near-identical plugin methods get born — which is exactly the duplication being removed elsewhere in this same PR. Reaching for the query that already existed avoided reintroducing it.

The comment on `:158-161` also checks out: `GetTotalActiveStake` really is a targeted aggregate with an EpochSummary fast path falling back to a single SUM, so there was nothing further to avoid there and saying so saves the next reader the lookup.

api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	// Network-wide total live stake, for live_size. Unlike active stake's
+	// total, there is no pre-aggregated column to query directly here, so
+	// this genuinely requires summing every pool's delegated stake — the
+	// same cost PoolsExtended and Network already pay for the same
+	// number. liveStake reads within txn so this stays in the same
+	// snapshot as the reads above (see liveStake's doc comment for the
+	// one read that still can't).
+	totalLiveStake, err := a.liveStake(txn.Metadata())
```
chrisguiney, line 194: Threading the txn through `liveStake` addresses the consistency half of that note, and the doc comment on the new signature states the reason precisely — reads staying in one snapshot rather than straddling a block boundary against each other. `Network()` passing `nil` keeps its existing behaviour, so this is additive.

The residual cost is inherent rather than fixable: `live_size` is a share of the network-wide live total, so something has to aggregate across every active pool. Unlike the active-stake side there's no stored aggregate to lean on, so this stays the most expensive read in the handler. Not a defect — worth knowing if a `pool_live_stake_total` ever gets persisted, since that would make this an O(1) lookup too.

api/blockfrost/adapter_pool_detail.go
```diff
@@ -0,0 +1,419 @@
...
+	state, err := a.ledgerState.Database().Metadata().GetNetworkState(txn)
+	if err != nil {
+		return 0, fmt.Errorf("get network state: %w", err)
+	}
+	reserves := uint64(0)
+	if state != nil {
+		reserves = uint64(state.Reserves)
+	}
+	if reserves > maxSupply {
+		return 0, nil
```
chrisguiney: The one item left, and it's minor: `reserves > maxSupply` returns `0, nil`, and `poolSizeSaturation` then guards `totalCirculation > 0`, so `live_saturation` comes back as a fabricated `0.0`.

That's the single surviving instance of the pattern the rest of this PR removed — and the function it feeds says so explicitly in its own doc comment: the zero guard *"does not stand in for 'nOpt unavailable' or 'totalCirculation unavailable'"*. Here it does exactly that.

Not blocking, for two reasons: it needs corrupt ledger state to reach, and it matches how `Network()` already handles the same comparison. But every other unavailable input on this path errors — protocol params, the epoch row, the shelley genesis above — so an error return would make this consistent with its neighbours in one line. CodeRabbit landed on the same reading independently, which suggests it reads as an oversight rather than a decision.

database/plugin/metadata/internal/poolcerthistory/poolcerthistory.go
```diff
@@ -0,0 +1,93 @@
...
+// certificates synthesized by the Mithril ledger-state import carry
+// certificate_id = 0, which matches no certs row — since they have no
+// originating transaction to report. Both pool_registration.pool_key_hash
+// and pool_retirement.pool_key_hash are indexed, so each query is a small,
+// indexed lookup rather than a table scan.
+func GetPoolCertificateHistory(
+	db *gorm.DB,
+	pkh lcommon.PoolKeyHash,
+) ([][]byte, [][]byte, error) {
+	transactionTable := sqldialect.TransactionTableName(db)
```
chrisguiney, line 44: This is the duplication finding resolved exactly as the codebase already does it elsewhere: the query lives once, `sqldialect.TransactionTableName(db)` handles the one thing the three backends actually disagree about, and each plugin is a delegation of a handful of lines.

Worth stating the ratio, since it's the argument for the pattern: 201 lines across three files became 93 in one plus three short wrappers, and the three-key `ORDER BY` — `added_slot`, `block_index`, `cert_index`, the tie-break CLAUDE.md calls out as mandatory — now exists in a single place where a future correction can land once.

It also lines up with the three existing precedents (`internal/accounthistory:172`, `internal/rewardstate/dialect.go:34`, `internal/stakequery:691`), so a reader who has seen one has seen this.

api/blockfrost/adapter_pool_detail_db_test.go
```diff
@@ -0,0 +1,432 @@
...
+	// derivable from the same devnet-genesis fixture this test already
+	// starts a real LedgerState against: MaxLovelaceSupply is fixed by
+	// config/cardano/devnet/shelley-genesis.json, and Reserves is whatever
+	// ls.Start() computed and persisted as the genesis network state
+	// (MaxLovelaceSupply minus the sum of all genesis UTxOs).
+	const wantMaxLovelaceSupply = uint64(2_000_000_000_000) // config/cardano/devnet/shelley-genesis.json
+	var networkState models.NetworkState
+	require.NoError(t, store.DB().Order("slot DESC").First(&networkState).Error)
+	wantCirculation := wantMaxLovelaceSupply - uint64(networkState.Reserves)
+	wantSaturation := float64(wantLiveStake) / (float64(wantCirculation) / float64(wantNOpt))
```
chrisguiney, line 250: Deriving `wantCirculation` from the real `MaxLovelaceSupply` and the live `networkState.Reserves` is what turns this from a test that agrees with the implementation into one that constrains it. My earlier objection was that the old assertion restated the formula under test, so it would have passed just as happily with the wrong denominator — this one wouldn't, and the revert experiment confirms it's the first to fail.

Keeping the real-`LedgerState` setup through the rework was the right call. It costs more than a table entry, but it's the only place `nOpt`, `MaxLovelaceSupply`, and `Reserves` all come from the same fixture the node actually loads, which is precisely the combination that went wrong the first time.

api/blockfrost/handlers_pool_detail.go
```diff
@@ -0,0 +1,115 @@
...
+func poolDetailResponse(info PoolDetailInfo) PoolDetailResponse {
+	// info.CalidusKey is always nil: dingo does not currently ingest
+	// CIP-0088 Calidus key registrations from any source (see
+	// PoolCalidusKeyInfo's doc comment in node_interface.go), so there is
+	// no producer that could set it, and no PoolCalidusKeyInfo ->
+	// PoolCalidusKeyResponse conversion to reach. PoolCalidusKeyResponse
+	// itself stays, since nullable: true in the OpenAPI schema makes null
+	// the correct wire representation and the type still documents that
+	// shape.
+	var calidusKey *PoolCalidusKeyResponse
```
chrisguiney, line 80: Good resolution — the unreachable conversion is gone and the comment above explains why nothing can set this, that there's therefore no `PoolCalidusKeyInfo` -> `PoolCalidusKeyResponse` path to reach, and why `PoolCalidusKeyResponse` still stays (the schema's `nullable: true` makes `null` the correct wire value, and the type documents that shape).

That last clause is the bit that matters. Deleting the branch without it would leave the next reader wondering whether the type is dead code to be removed, and removing it would break the response shape. Keeping the type and explaining the asymmetry is the version that survives contact with a future cleanup pass.

chrisguiney (dismissed):
Approving — this clears my outstanding `CHANGES_REQUESTED`. Re-review of `eded3883`; my earlier review was against `4e09f666`. **All seven findings are addressed**, and I confirmed the two blocking ones by experiment rather than by reading.

| Earlier finding | Status |
|---|---|
| **Blocking:** `live_saturation` divided by total active stake, not circulating supply | **Fixed** — new `totalCirculation` helper, verified by reverting it (inline on `:412`) |
| **Blocking:** `active_size` / `live_saturation` silently 0.0 with no mark snapshot | **Fixed** — decoupled into independent guards (`:409`, `:412`) |
| **Blocking:** `blocks_epoch` silently became the lifetime count on a nil epoch row | **Fixed** — now errors (`:289`) |
| **Duplication:** `GetPoolCertificateHistory` triplicated verbatim | **Fixed** — shared package + `sqldialect.TransactionTableName` |
| Two full-network aggregations per single-pool request; mixed txn scopes | **Fixed** — targeted lookup (`:144`) and txn threading (`:194`) |
| Dead Calidus conversion branch | **Fixed** — deleted, replaced by a documented nil |
| Tests encoded the wrong formula, so couldn't catch it | **Fixed**, and now load-bearing (`adapter_pool_detail_db_test.go:250`) |

## Verification performed

| Check | Result |
|---|---|
| `go build ./...` and `-tags dingo_extra_plugins` | clean |
| `go test -race -count=1 ./api/blockfrost/...` | pass |
| `go test -count=1 ./database/...` | pass |
| Denominator-revert experiment | **4 tests fail** — output inline on `:412` |
| CI on this PR | green, except one windows `go-test` still running at review time |

## Performance

Better than the version I reviewed, not merely equal — details inline on `:144` and `:194`. The short version: the ~3,000-row snapshot fetch became a single targeted row read using a **pre-existing** store method, so no new plugin implementations were added; `totalCirculation` costs one `GetNetworkState` row and deliberately avoids `Network()`'s `GetScriptLockedSupply` path; and both network-wide reads now run inside the caller's transaction.

One inherent cost remains: `live_size` needs the network-wide live total, so `liveStake` still aggregates across all active pools. There's no cheaper route without a stored aggregate, and it's now at least snapshot-consistent with the rest of the response.

## Documentation

`DATABASE.md` (+59) and `ARCHITECTURE.md` (+4) are still accurate after the rework. The saturation change needed no doc update, since neither file documented the formula.

One minor item left, inline on `:375`.

wolf31o2:
Second-pass review items addressed in 48354a68. All three were valid.

## reserves exceeding maxLovelaceSupply returned a silent 0

Correct catch, and it is the same defect class as the two blocking findings from the first pass — I just missed this branch when fixing them. `totalCirculation` returned `0` when `reserves > maxSupply`, which would feed a zero denominator into `live_saturation` and surface as a plausible `0.0`, while every other unavailable input in that function errors precisely because `0.0` is a value a real pool can have. It now errors, with a comment saying why rather than leaving the next reader to infer it. Impossible ledger state, so this is consistency rather than a live bug, but the inconsistency was the point.

## The epoch-row test did not pin its branch

Also correct. `require.ErrorContains(t, err, "epoch")` was satisfied by `get pool stake snapshot for epoch %d` and `get total active stake for epoch %d` as well, so the test passed without reaching the branch it exists to cover — particularly weak now that the active-stake path errors too, which this PR added. It matches `no epoch row found` instead, and `TestNodeAdapterPoolDetailEpochRowMissing` still passes, which confirms it genuinely reaches that branch rather than just tightening the string.

## assert.Empty also passes for nil

Right, and it matters here specifically: `GetPoolCertificateHistory` builds its results with `make([][]byte, 0, len(rows))` so the success path is non-nil, and the pool detail response needs those required arrays to serialize as `[]` rather than `null` — which `TestHandlePoolDetailEmptyArraysNotNull` depends on. The test now pins non-nil before asserting empty, so a future change returning a bare `nil` fails here instead of silently changing the wire format.

## Also confirming the first-pass items

For completeness, since the earlier review had comments I want to account for explicitly: the `active_size` fabricated `0.0` (`:409`) now errors, and the unreachable `PoolCalidusKeyInfo` conversion (`handlers_pool_detail.go:73`) is removed with the reasoning left in a comment, as you suggested. The `blocks_minted` note at `:263` needed no change — thanks for re-deriving that one independently.

## Verification

`go build` (both tag sets), `go test ./api/blockfrost/... ./database/...` with 0 `^FAIL` and 0 `^--- FAIL`, `golangci-lint run` 0 issues, `nilaway`, `make import-boundaries`, and `go test -race ./api/blockfrost/...` all pass.

chrisguiney (approved):

PR #3015, opened 2026-07-29
feat(blockfrost): implement account reward history endpoint

### LINKED ISSUE #1875: Blockfrost API: Implement /accounts/{stake_address}/rewards endpoint
## Summary

The `AccountRewardHistory` method on `NodeAdapter` (in `blockfrost/adapter.go`) is currently a stub that always returns an empty slice after parsing the stake address. This means the public `GET /accounts/{stake_address}/rewards` endpoint is semantically incorrect — a stake account with real reward history is indistinguishable from one with none.

## Background

This was identified during the review of PR #1838 (Blockfrost API: Account (stake address) endpoints). The endpoint was intentionally left as a stub in that PR and deferred to a follow-up.

See: https://github.com/blinklabs-io/dingo/pull/1838#discussion_r3038014851

## Work Required

1. Design and implement the reward-history data model (table/schema for per-account, per-epoch reward tracking).
2. Add `GetAccountRewardHistory` to the `MetadataStore` interface and implement it for SQLite, Postgres, and MySQL backends.
3. Add a `GetAccountRewardHistory` facade method on the `Database` type.
4. Update `AccountRewardHistory` on `NodeAdapter` to:
   - Use the parsed staking key from `parseStakeAddress`.
   - Query the reward-history backend.
   - Map results into `[]AccountRewardHistoryInfo` (fields: `Epoch`, `Amount`, `PoolID`).
5. Add handler tests and blockfrost-go compatibility tests for the `/rewards` response payload.

## References

- PR #1838 — initial account endpoints implementation
- Comment: https://github.com/blinklabs-io/dingo/pull/1838#discussion_r3038014851
- Requested by: @arepala-uml

wolf31o2:
Closes #1875

`NodeAdapter.AccountRewardHistory` was a stub that validated the stake address and then always returned an empty slice, so an account with real reward history was indistinguishable from one with none.

## The issue text is stale

The stub TODO claimed no backing reward storage existed. It does. `models.RewardAccountOutput` (table `reward_account_output`) is already per-account, per-epoch, with pool ID and reward type, written by the reward calculation engine. So no new table was created. What was actually missing was a credential-filtered query and a retention policy that keeps history long enough to serve.

## Retention is gated on storage mode

Per maintainer direction, full reward history is retained in API storage mode only; core mode prunes exactly as before.

`reward_stake_input` continues to be pruned to the 4-epoch window in BOTH modes. It scales with delegator count, roughly 1.3M rows per epoch on mainnet, and the existing comment in `DeleteStateBeforeEpoch` already documents why it cannot be kept.

`reward_account_output` is pruned as before in `core` mode, so a block producer or relay sees byte-identical behavior, and retained without bound in `api` mode so the endpoint can serve full history instead of the trailing few epochs. `cleanupOldSnapshots` branches on `m.db.StorageMode()`, which was already reachable since the snapshot manager holds the `*database.Database`. A sibling method `DeleteRewardStakeInputBeforeEpoch` was added rather than adding a mode parameter to the existing method, so the core path and its existing test are untouched.

Operators choosing `storageMode: api` should note the growth: this table grows by roughly 1.3M rows per epoch on mainnet, about 5k on preview, for the life of the database. That is stated explicitly in `DATABASE.md` and noted under `storageMode` in `dingo.yaml.example`.

Rollback is unaffected: `DeleteRewardStateAfterSlot` never consults storage mode and still removes retained rows above a rollback point in both modes. A test pins this directly rather than leaving it true by omission.

The precompute-reuse path is also unaffected. `GetRewardAccountOutputs` has one non-test caller, it is scoped to a specific snapshot epoch, and it sits behind an `inputsValid` guard, so retained older rows cannot leak into a reward calculation.

## Index

The new query filters `(credential_tag, staking_key)`, but the only composite index on this table leads with `epoch`, so the query could not use it. Before:

```
SCAN reward_account_output USING INDEX idx_reward_account_output_epoch_cred_pool_type
USE TEMP B-TREE FOR RIGHT PART OF ORDER BY
```

That is a full scan plus a full sort per request, on a table this PR makes grow without bound in API mode. Added `idx_reward_account_output_credential (credential_tag, staking_key, epoch, pool_key_hash, reward_type)` so the predicate and the ORDER BY are both index-satisfied. After:

- SQLite: `SEARCH reward_account_output USING INDEX idx_reward_account_output_credential (credential_tag=? AND staking_key=?)`, no scan, no sort
- MySQL 8.0: `type: ref, key: idx_reward_account_output_credential, key_len: 31`
- PostgreSQL 16: `Index Scan using idx_reward_account_output_credential`

MySQL and Postgres were verified against real containers, not assumed. Worst-case key length is about 129 bytes against InnoDBs 3072-byte limit, and the index is the same column set as the pre-existing unique index, permuted. No `INDEXED BY` hint is used; the planner picks the index on its own in all three backends.

Upgrade in place was verified with a legacy-model AutoMigrate test following this codebases existing `MigrateRewardLiveStakePoolIndex` pattern: the index is added by a plain `CREATE INDEX` with no table rebuild and existing rows survive. No migration helper was needed.

## Response shape

`account_reward_content.yaml` requires a `type` field with enum `leader | member | pool_deposit_refund`, which dingos structs lacked. Added to both the adapter and response structs and mapped from `RewardAccountOutput.RewardType`. dingo only produces `leader` and `member` today and those spellings already match Blockfrost verbatim; a lowercase passthrough fallback means any future value is surfaced rather than silently dropped.

## Tests

Multi-epoch history across more than one pool, pagination and order, empty history for a registered account, unknown account, database errors, `type` mapping, retention behavior in both storage modes, the rollback exemption, a query-plan assertion so a future index drop fails loudly instead of silently degrading to a scan, and the upgrade-in-place migration proof. No existing assertion was modified.

## Verification

`go build ./...` (default and `-tags dingo_extra_plugins`), `go test ./api/blockfrost/... ./database/... ./ledger/...`, `golangci-lint run` (both tag sets), `nilaway`, `modernize`, `make import-boundaries`, `go test -race` across the affected packages, and `go test ./internal/test/conformance/...` all pass. Verified to merge cleanly with #3012 and #3014 and to build and pass tests in combination with them.

## Documentation

`DATABASE.md` documents the per-mode retention split, the growth rate, the new index and its rationale, and the new query surface. `dingo.yaml.example` notes the retention difference under `storageMode`.

---
## Summary by cubic
Implements the Blockfrost account reward history endpoint with a credential-filtered, spendable-only, paginated query; API storage mode retains full per-account reward history while core mode behavior is unchanged. Corrects epoch mapping to the earned epoch and adds a schema-safe reward type in responses.

- **New Features**
  - `api/blockfrost`: Implemented `AccountRewardHistory` (GET /accounts/{stake_address}/rewards) with pagination and asc/desc order; excludes uncredited rewards (`spendable=false`), reports the earned epoch, and returns `type` (leader/member/pool_deposit_refund) with an allow-list and warn-through fallback.
  - Database: Added `GetRewardAccountOutputsByCredential` and `CountRewardAccountOutputsByCredential` across SQLite/MySQL/PostgreSQL and wired through `database.Database`; documented the spendable-only filter on the store interface.
  - Retention: In `api` storage mode, keep full `reward_account_output` history; in `core`, prune as before. `reward_stake_input` is pruned in both modes. Rollback semantics unchanged.

- **Migration**
  - No manual steps. AutoMigrate creates `idx_reward_account_output_credential_spendable` and a migration drops the superseded `idx_reward_account_output_credential`; existing rows are preserved.
  - In `api` mode, expect ~1.3M new `reward_account_output` rows/epoch on mainnet and increased disk usage from the second index. Documentation and `dingo.yaml.example` updated.

<sup>Written for commit 70d318f174f107d303bc3e4a74f9ee06de513c50. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3015?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Implemented Blockfrost account reward history with pagination, ordering, reward-type information, and spendable-reward filtering.
  * Added credential-based reward history queries and total counts across supported database backends.
* **Bug Fixes**
  * Corrected reward retention behavior by storage mode: API mode preserves full account history while core mode retains the recent window.
  * Ensured rollback cleanup consistently removes data after the rollback point.
* **Documentation**
  * Updated architecture, database, and configuration guidance for reward history retention and querying.
* **Performance**
  * Improved reward history query efficiency with an index covering credential and spendability.

### DIFF
--- a/api/blockfrost/adapter.go
+++ b/api/blockfrost/adapter.go
@@ -2413,11 +2413,70 @@ func (a *NodeAdapter) AccountRewardHistory(
 		); err != nil {
 		return nil, 0, err
 	}
-	// TODO(#1875): Implement reward history once Dingo persists
-	// per-account, per-epoch reward records. This endpoint remains
-	// stubbed in this PR because the backing reward-history storage
-	// and rollback-safe ledger/database plumbing do not exist yet.
-	return []AccountRewardHistoryInfo{}, 0, nil
+	offset := (params.Page - 1) * params.Count
+	total, err := a.ledgerState.Database().
+		CountRewardAccountOutputsByCredential(
+			credentialTag,
+			stakeKey,
+			nil,
+		)
+	if err != nil {
+		return nil, 0, fmt.Errorf(
+			"count account reward history: %w",
+			err,
+		)
+	}
+	if offset >= total {
+		return []AccountRewardHistoryInfo{}, total, nil
+	}
+	rows, err := a.ledgerState.Database().
+		GetRewardAccountOutputsByCredential(
+			credentialTag,
+			stakeKey,
+			params.Count,
+			offset,
+			params.Order,
+			nil,
+		)
+	if err != nil {
+		return nil, 0, fmt.Errorf(
+			"get account reward history: %w",
+			err,
+		)
+	}
+	// blockfrostRewardTypes maps dingo's stored reward_account_output.reward_type
+	// values (ledger/rewards.RewardType: "leader", "member") to the Blockfrost
+	// account_reward_content enum spelling (leader, member,
+	// pool_deposit_refund). The two overlapping spellings already match
+	// verbatim; this only defends against case variance and normalizes a
+	// not-yet-modeled reward type to lowercase rather than silently dropping
+	// it, so a future addition (e.g. a CIP-0163 pool deposit refund) still
+	// surfaces in the response.
+	blockfrostRewardTypes := map[string]string{
+		"leader":              "leader",
+		"member":              "member",
+		"pool_deposit_refund": "pool_deposit_refund",
+	}
+	ret := make([]AccountRewardHistoryInfo, 0, len(rows))
+	for _, row := range rows {
+		epoch, err := uint64ToInt32(row.Epoch, "reward epoch")
+		if err != nil {
+			return nil, 0, err
+		}
+		rewardType, ok := blockfrostRewardTypes[strings.ToLower(row.RewardType)]
+		if !ok {
+			rewardType = strings.ToLower(row.RewardType)
+		}
+		ret = append(ret, AccountRewardHistoryInfo{
+			Epoch:  epoch,
+			Amount: strconv.FormatUint(uint64(row.Amount), 10),
+			PoolID: lcommon.PoolId(
+				lcommon.NewBlake2b224(row.PoolKeyHash),
+			).String(),
+			Type: rewardType,
+		})
+	}
+	return ret, total, nil
 }
 
 func blockIssuer(issuer lcommon.IssuerVkey) string {
--- a/api/blockfrost/adapter.go
+++ b/api/blockfrost/adapter.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"math"
 	"math/big"
 	"net/http"
@@ -2394,6 +2395,26 @@ func (a *NodeAdapter) AccountRegistrationHistory(
 	return ret, total, nil
 }
 
+// blockfrostRewardTypes is the allow-list of reward_account_output.reward_type
+// values (ledger/rewards.RewardType: RewardTypeLeader "leader", RewardTypeMember
+// "member") that dingo produces today, plus "pool_deposit_refund", which
+// dingo does not yet produce but the Blockfrost account_reward_content "type"
+// enum already defines. It is a package-level var, built once rather than
+// per-request.
+//
+// Membership, not translation, is the point: every recognized value already
+// matches its lowercased form verbatim, so this is not a mapping table. Its
+// job is to catch the day dingo starts producing a reward_type outside this
+// closed set — which would silently make the Blockfrost response
+// schema-invalid otherwise. AccountRewardHistory logs a warning when a row's
+// type is not in this set and still passes the lowercased value through
+// rather than dropping the row.
+var blockfrostRewardTypes = map[string]struct{}{
+	"leader":              {},
+	"member":              {},
+	"pool_deposit_refund": {},
+}
+
 // AccountRewardHistory returns reward history rows for
 // the requested stake address.
 func (a *NodeAdapter) AccountRewardHistory(
@@ -2404,22 +2425,28 @@ func (a *NodeAdapter) AccountRewardHistory(
 	if err != nil {
 		return nil, 0, err
 	}
-	if _, err := a.ledgerState.Database().
-		GetAccountByCredential(
-			credentialTag,
-			stakeKey,
-			true,
-			nil,
-		); err != nil {
+	db := a.ledgerState.Database()
+	txn := db.Transaction(false)
+	defer txn.Release()
+
+	if _, err := db.GetAccountByCredential(
+		credentialTag,
+		stakeKey,
+		true,
+		txn,
+	); err != nil {
 		return nil, 0, err
 	}
 	offset := (params.Page - 1) * params.Count
-	total, err := a.ledgerState.Database().
-		CountRewardAccountOutputsByCredential(
-			credentialTag,
-			stakeKey,
-			nil,
-		)
+	// count and rows share the read txn opened above so an epoch boundary
+	// landing between the two queries cannot change total relative to the
+	// page: with two independent nil-txn reads, a client paging through
+	// history at a boundary could see a row twice or miss one.
+	total, err := db.CountRewardAccountOutputsByCredential(
+		credentialTag,
+		stakeKey,
+		txn,
+	)
 	if err != nil {
 		return nil, 0, fmt.Errorf(
 			"count account reward history: %w",
@@ -2429,43 +2456,45 @@ func (a *NodeAdapter) AccountRewardHistory(
 	if offset >= total {
 		return []AccountRewardHistoryInfo{}, total, nil
 	}
-	rows, err := a.ledgerState.Database().
-		GetRewardAccountOutputsByCredential(
-			credentialTag,
-			stakeKey,
-			params.Count,
-			offset,
-			params.Order,
-			nil,
-		)
+	rows, err := db.GetRewardAccountOutputsByCredential(
+		credentialTag,
+		stakeKey,
+		params.Count,
+		offset,
+		params.Order,
+		txn,
+	)
 	if err != nil {
 		return nil, 0, fmt.Errorf(
 			"get account reward history: %w",
 			err,
 		)
 	}
-	// blockfrostRewardTypes maps dingo's stored reward_account_output.reward_type
-	// values (ledger/rewards.RewardType: "leader", "member") to the Blockfrost
-	// account_reward_content enum spelling (leader, member,
-	// pool_deposit_refund). The two overlapping spellings already match
-	// verbatim; this only defends against case variance and normalizes a
-	// not-yet-modeled reward type to lowercase rather than silently dropping
-	// it, so a future addition (e.g. a CIP-0163 pool deposit refund) still
-	// surfaces in the response.
-	blockfrostRewardTypes := map[string]string{
-		"leader":              "leader",
-		"member":              "member",
-		"pool_deposit_refund": "pool_deposit_refund",
-	}
 	ret := make([]AccountRewardHistoryInfo, 0, len(rows))
 	for _, row := range rows {
-		epoch, err := uint64ToInt32(row.Epoch, "reward epoch")
+		// row.Epoch is the snapshot epoch (stakeRewardEpochsForNewEpoch's
+		// "snapshot" = newEpoch-3 in ledger/reward_calculation.go, which is
+		// what rewardAccountOutputs persists into this column), not the
+		// earned epoch Blockfrost reports. Rewards computed from that
+		// snapshot are credited at the boundary into newEpoch, i.e. they
+		// become spendable in newEpoch = snapshot+3. cardano-db-sync, which
+		// backs this endpoint on Blockfrost, models
+		// spendable_epoch = earned_epoch + 2, so earned_epoch = newEpoch-2 =
+		// snapshot+1 (the "performance" epoch in the same struct). Adding 1
+		// cannot underflow: row.Epoch is a uint64 and this is addition, so a
+		// stored epoch of 0 (or any value) yields a valid earned epoch.
+		earnedEpoch := row.Epoch + 1
+		epoch, err := uint64ToInt32(earnedEpoch, "reward epoch")
 		if err != nil {
 			return nil, 0, err
 		}
-		rewardType, ok := blockfrostRewardTypes[strings.ToLower(row.RewardType)]
-		if !ok {
-			rewardType = strings.ToLower(row.RewardType)
+		rewardType := strings.ToLower(row.RewardType)
+		if _, ok := blockfrostRewardTypes[rewardType]; !ok {
+			slog.Warn(
+				"account reward history: reward_type outside the Blockfrost account_reward_content enum",
+				"reward_type", row.RewardType,
+				"credential_tag", credentialTag,
+			)
 		}
 		ret = append(ret, AccountRewardHistoryInfo{
 			Epoch:  epoch,
--- a/api/blockfrost/adapter_reward_history_test.go
+++ b/api/blockfrost/adapter_reward_history_test.go
@@ -16,6 +16,9 @@ package blockfrost
 
 import (
 	"bytes"
+	"log/slog"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/blinklabs-io/dingo/database/models"
@@ -49,6 +52,14 @@ func newRewardHistoryStakeAddress(
 // owner's leader reward alongside a member reward in the same epoch), and
 // verifies the default (ascending) ordering plus the Blockfrost field
 // mapping, including the "type" enum value.
+//
+// Stored Epoch is the reward-calculation snapshot epoch, not the earned
+// epoch Blockfrost reports (dingo #1875 review finding 2): the reward
+// computed from snapshot epoch S is credited at the boundary into S+3, which
+// is when it becomes spendable, and cardano-db-sync (what Blockfrost serves
+// this endpoint from) models spendable_epoch = earned_epoch + 2, so
+// earned_epoch = S+1. Row ordering is still driven by the stored (snapshot)
+// epoch; only the reported value is shifted.
 func TestAccountRewardHistoryMultiEpochMultiPool(t *testing.T) {
 	adapter, store, _ := newDBBackedAdapter(t)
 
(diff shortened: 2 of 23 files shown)


database/plugin/metadata/internal/rewardstate/state.go
```diff
@@ -593,6 +594,76 @@ func GetAccountOutputs(db *gorm.DB, epoch uint64) ([]*models.RewardAccountOutput
...
+	limit int,
+	offset int,
+	order string,
+) ([]*models.RewardAccountOutput, error) {
+	ret := make([]*models.RewardAccountOutput, 0)
+	if len(stakingKey) == 0 {
+		return ret, nil
+	}
+	query := db.Where(
+		"credential_tag = ? AND staking_key = ?",
```
chrisguiney: **Blocking.** The predicate filters only on the credential, so rows with `spendable = false` come back and are reported as rewards the account received. They weren't.

`ledger/reward_calculation.go:375-377` skips them at application time:

```go
if !reward.Spendable {
    continue
}
```

so `AddAccountRewardByCredential` never runs and the amount is instead accumulated into `app.unspendable` (`:2100-2102`). `finalizePrecomputedRewardOutputs` (`:1533`, `:1574`) persists that flag by reconciling against `GetAccountsByCredential`, so a credential deregistered before the payout boundary leaves a permanent `spendable = false` row.

Probed against the sqlite store — one spendable row and one not:

```
returned: epoch=10 amount=1000000 spendable=true
returned: epoch=11 amount=9999999 spendable=false
rows=2 count=2 summed_amount=10999999   (actually credited: 1000000)
```

The count is wrong too, so pagination advertises pages of rewards that were never paid.

The inconsistency is checkable entirely inside dingo: `/accounts/{stake_address}` reports `RewardsSum` from `account.Reward` (`adapter.go:2106-2112`), which only ever sees credited rewards. So for any account with a non-spendable row, the sum of `/rewards` exceeds `rewards_sum` on the sibling endpoint. A `spendable = true` filter on both the query and the count fixes this half. Worth adding the column to the index too, or the filter turns the index seek into a seek-plus-filter.

**The related half is harder and worth an explicit decision:** CIP-0163-guarded credentials are also skipped at crediting (`rewardOutputGuarded`, `:379` and `:524-537`) but that guard is *not* written back to the row, so those rows keep `spendable = true` and a filter won't catch them. That only bites once CIP-0163 is activated, but it's the same class of overstatement.

api/blockfrost/adapter_reward_history_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+	require.NoError(t, store.SaveRewardAccountOutputs([]*models.RewardAccountOutput{
+		{
+			Epoch:         1,
+			CredentialTag: 0,
+			StakingKey:    stakingKey,
+			PoolKeyHash:   poolA,
+			RewardType:    "member",
+			Amount:        100,
+			Spendable:     true,
```
chrisguiney, line 84: Every fixture in this file sets `Spendable: true`, which is why nothing catches the finding on the query predicate. The one case that would — a `spendable = false` row that must be *absent* from the response, and absent from the total — isn't represented anywhere in the new tests.

Worth adding alongside the fix, since it's the assertion that keeps the filter from being dropped later: it's invisible in every other test.

api/blockfrost/adapter.go
```diff
@@ -2413,11 +2413,70 @@ func (a *NodeAdapter) AccountRewardHistory(
...
+	// it, so a future addition (e.g. a CIP-0163 pool deposit refund) still
+	// surfaces in the response.
+	blockfrostRewardTypes := map[string]string{
+		"leader":              "leader",
+		"member":              "member",
+		"pool_deposit_refund": "pool_deposit_refund",
+	}
+	ret := make([]AccountRewardHistoryInfo, 0, len(rows))
+	for _, row := range rows {
+		epoch, err := uint64ToInt32(row.Epoch, "reward epoch")
```
chrisguiney: **Blocking on verification, not on a specific number.** This passes `row.Epoch` through unchanged, and `row.Epoch` is the *snapshot* epoch, not the earned epoch that Blockfrost reports.

Tracing the writer: `rewardAccountOutputs(rewardSnapshotEpoch, ...)` at `reward_calculation.go:304`, `rewardSnapshotEpoch := epochs.snapshot` at `:150`, and `stakeRewardEpochsForNewEpoch` at `:2207-2222` sets

```go
snapshot:    newEpoch - 3,
performance: newEpoch - 2,   // "the completed block-performance epoch"
pots:        newEpoch - 1,
```

Rewards are credited at the boundary into `newEpoch`, so they're spendable in `newEpoch`. cardano-db-sync — which is what Blockfrost serves this endpoint from — models `spendable_epoch = earned_epoch + 2`, which puts `earned_epoch` at `newEpoch - 2`, i.e. `performance`, one higher than what's stored and returned here.

What makes me want this pinned rather than asserted: dingo's own two conventions can't both be right. The reward path implies `performance = snapshot + 1`, while the API path treats a snapshot labelled `X` as the active stake for epoch `X + 2` (`activeStakeEpoch = currentEpoch - 2` in `Network`, `PoolsExtended`, and now `PoolDetail`). Those differ by one, so the snapshot label alone doesn't settle what `epoch` should be.

This endpoint is the first user-facing surface for these epochs — nothing under `api/` reads `reward_pool_output`, so there's no existing convention to inherit and no consumer to stay consistent with. That makes it worth resolving now: query one account with known history against Blockfrost or a db-sync instance and compare the epoch labels. If the reported epoch is low by one, `epochs.performance` is already computed in the same struct, so the correction is available without new plumbing.

Either way the resolution belongs in a comment here — "epoch is the X epoch" — because the column name alone won't tell the next reader which of the three epochs in `stakeRewardEpochs` it is.

api/blockfrost/adapter.go
```diff
@@ -2413,11 +2413,70 @@ func (a *NodeAdapter) AccountRewardHistory(
...
+	}
+	// blockfrostRewardTypes maps dingo's stored reward_account_output.reward_type
+	// values (ledger/rewards.RewardType: "leader", "member") to the Blockfrost
+	// account_reward_content enum spelling (leader, member,
+	// pool_deposit_refund). The two overlapping spellings already match
+	// verbatim; this only defends against case variance and normalizes a
+	// not-yet-modeled reward type to lowercase rather than silently dropping
+	// it, so a future addition (e.g. a CIP-0163 pool deposit refund) still
+	// surfaces in the response.
+	blockfrostRewardTypes := map[string]string{
```
chrisguiney: This map cannot change any output, and it's rebuilt on every request.

Each key maps to itself and the fallback is `strings.ToLower(row.RewardType)` — which is exactly what a successful lookup returns for all three keys. So the whole thing is equivalent to one line:

```go
rewardType := strings.ToLower(row.RewardType)
```

Probed across `leader`/`member`/`pool_deposit_refund`, case variants, `""`, an unknown value, and a leading-space value: 10 inputs, **0** differences between the map path and plain `strings.ToLower`. `TestAccountRewardHistoryTypeMapping` passes with or without the map for the same reason, so it doesn't pin anything either.

The stated intent — surfacing a future reward type rather than silently dropping it — is already what the fallback does on its own. Two options that would earn the lines back:

1. Delete the map, keep the `ToLower`, and move the reasoning into a comment.
2. Make it a real allow-list: on an unrecognized value, log at warn and still pass it through. Then the map has a job (telling you dingo started producing a type the response schema doesn't know), and it should be a package-level `var` rather than per-request.

Option 2 has actual value given the enum is closed in the OpenAPI schema — a value outside `leader | member | pool_deposit_refund` makes the response schema-invalid, and right now that would pass through silently.

api/blockfrost/adapter.go
```diff
@@ -2413,11 +2413,70 @@ func (a *NodeAdapter) AccountRewardHistory(
...
 		return nil, 0, err
 	}
-	// TODO(#1875): Implement reward history once Dingo persists
-	// per-account, per-epoch reward records. This endpoint remains
-	// stubbed in this PR because the backing reward-history storage
-	// and rollback-safe ledger/database plumbing do not exist yet.
-	return []AccountRewardHistoryInfo{}, 0, nil
+	offset := (params.Page - 1) * params.Count
+	total, err := a.ledgerState.Database().
+		CountRewardAccountOutputsByCredential(
```
chrisguiney: The count and the page are two separate reads with `nil` txn, so they aren't a consistent snapshot. An epoch boundary landing between them changes `total` relative to the rows, and because `offset` is derived from the caller's page against the *old* count, a client paging through history at a boundary can see a row twice or miss one.

Small window and only at boundaries, but this method already has the fix available — `PoolDetail` and `PoolMetadata` both open one `db.Transaction(false)` and thread it through, and both store methods here take a `types.Txn`. Passing one read txn to both calls costs nothing and removes the window.

database/plugin/metadata/sqlite/reward_state.go
```diff
@@ -270,6 +270,68 @@ func (d *MetadataStoreSqlite) GetRewardAccountOutputs(epoch uint64, txn types.Tx
...
+	txn types.Txn,
+) ([]*models.RewardAccountOutput, error) {
+	db, err := d.resolveReadDB(txn)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"resolve read DB for reward account outputs by credential: %w",
+			err,
+		)
+	}
+	rows, err := rewardstate.GetAccountOutputsByCredential(
```
chrisguiney, line 290: Crediting this deliberately: the SQL lives once in `internal/rewardstate` and the three plugin methods are thin delegations that differ only in receiver type, which is unavoidable for interface implementations. That's the `internal/certutil` / `internal/drepquery` / `internal/stakequery` pattern used correctly.

Worth stating because it's the direct contrast with #3014, where the new `GetPoolCertificateHistory` is 67 lines copied verbatim into all three backends. Same shape of problem, and this is the version that scales — one place to fix an `ORDER BY`, one place a future `spendable` filter has to land.

ledger/snapshot/rotation.go
```diff
@@ -989,7 +1003,16 @@ func (m *Manager) cleanupOldSnapshots(
 		return fmt.Errorf("cleanup pool snapshots: %w", err)
 	}
 
-	if err := meta.DeleteRewardStateBeforeEpoch(
+	if m.db.StorageMode() == types.StorageModeAPI {
```
chrisguiney, line 1006: This is the line I expected to have a hole in it and it doesn't. Two things I checked specifically:

- `database/types/types.go:185-187` defines only `core` and `api`, so the `else` is exactly core. Just as important, the polarity is safe: an unrecognized mode falls through to pruning rather than to unbounded retention, so a future third mode can't accidentally inherit the growth.
- `DeleteRewardStateAfterSlot` genuinely never consults storage mode, so rollback still removes retained rows above the rollback point in both modes — and `TestDeleteRewardStateAfterSlotUnaffectedByAPIModeRetention` pins it rather than leaving it true by omission, which is the right instinct for a behaviour that only breaks in one mode.

The retention decision itself is the kind of thing that deserves the paragraph it got. One thing the docs quantify in rows but not bytes: `api` mode now carries a second five-column index over the same unbounded table, so the per-row cost is roughly double what the row count alone suggests. Worth a clause in `DATABASE.md`'s growth estimate, since that estimate is what an operator will size a disk against.

chrisguiney (changes requested):
Requesting changes on two items. The first is a correctness bug I reproduced: the query returns reward rows the account never received, so the sum of `/rewards` disagrees with `rewards_sum` on `/accounts/{stake_address}` for the same account. The second is a question about which epoch the `epoch` field is labelled with — dingo has two internal conventions that disagree by one, and this endpoint is the first user-facing surface that has to pick.

The retention design, the index work, and the storage-mode split all hold up. The duplication handling here is exactly right and worth calling out — see the inline note on the sqlite delegation.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...` and `-tags dingo_extra_plugins` | clean |
| `go test -race -count=1 ./api/blockfrost/... ./database/...` | **all pass** (17 packages) |
| `go test ./ledger/snapshot/... ./ledger/rewards/... ./internal/test/conformance/...` | pass |
| `modernize ./...` | only pre-existing hits in `midnight/`, `plugin/config.go`; none in touched files |
| Spendable probe (written for this review) | **non-spendable rows returned, confirmed** |
| Reward-type map probe (written for this review) | **provable no-op, confirmed** |
| Blockfrost `account_reward_content.yaml` (fetched) | `type` enum and required-ness confirmed |

DevNet isn't required — read path plus a retention branch, nothing in CLAUDE.md's consensus list. Conformance ran and passes.

## Claims I checked that hold up

- **`GetRewardAccountOutputs` really has one non-test caller** (`ledger/reward_calculation.go:634`), scoped to `epochs.snapshot` and sitting after both the `inputsValid` early return and `precomputedRewardPoolOutputsMatchInputs`. Retained older rows cannot leak into a reward calculation.
- **The storage-mode branch is safe by default.** `database/types/types.go:185-187` defines only `core` and `api`, so the `else` really is core and anything unrecognized prunes rather than growing without bound. Adding `DeleteRewardStakeInputBeforeEpoch` as a sibling instead of a mode parameter does leave the core path and its existing test untouched, as described.
- **No handler change was needed** because `handlers.go:2236` does a struct conversion, `AccountRewardHistoryResponse(item)`. Adding `Type` to both structs in the same position keeps that conversion legal, and the compiler enforces it — so this is compile-time safe rather than a latent gap.
- **Pagination ordering is deterministic.** The three-key tie-break (`epoch, pool_key_hash, reward_type`) matches the new index and does make pages stable, and the reasons ties happen are correctly identified in the doc comment.
- **The model doc comment is more precise than the PR body.** The body says "no scan, no sort"; the comment on `RewardAccountOutput` correctly notes the descending path still sorts the matched single-credential rows. The precise version is the right one — a mixed-direction `ORDER BY` (`epoch DESC, pool_key_hash ASC`) can't be satisfied by an index scan in either direction.

## Documentation

`DATABASE.md` (+114) documents the per-mode retention split, the ~1.3M rows/epoch growth, the new index and why the pre-existing epoch-leading unique index can't serve the query, and the new query surface. `ARCHITECTURE.md` (+33) and `dingo.yaml.example` both cover the retention difference, and the example config is explicit that `api` mode accepts unbounded growth. All accurate.

wolf31o2:
All five findings addressed in 5e79b617. Thanks for the probes — the spendable one in particular would have been easy to argue away without the reproduction.

## 1. Non-spendable rows returned (blocking)

Fixed. `spendable = true` now filters both the row query and the `COUNT(*)`, so the phantom pages go away with the phantom rows.

Your framing of why this was invisible is the useful part: every fixture in the new test file set `Spendable: true`, so nothing could catch it. Added `TestAccountRewardHistoryExcludesNonSpendableReward` reproducing your exact case — spendable 1000000 alongside non-spendable 9999999 — asserting total 1 and summed 1000000, plus matching sqlite/mysql/postgres store tests. That is the assertion that keeps the filter from being quietly dropped later.

Also took the point about the filter turning the seek into a seek-plus-filter. `spendable` is now folded into the index:

```
BEFORE (idx_reward_account_output_credential):
  SEARCH reward_account_output USING INDEX idx_reward_account_output_credential (credential_tag=? AND staking_key=?)

AFTER (idx_reward_account_output_credential_spendable):
  SEARCH reward_account_output USING INDEX idx_reward_account_output_credential_spendable (credential_tag=? AND staking_key=? AND spendable=?)
```

The replacement is a new index name with `MigrateRewardAccountOutputCredentialIndex` dropping the superseded one only after confirming the replacement exists, so a failure between the two steps cannot leave the table without a credential-leading index. `TestMigrateRewardAccountOutputCredentialSpendableIndex` seeds both a spendable and a non-spendable row, runs `AutoMigrate`, then the helper, and asserts the old index is gone, the new one remains, both rows survive intact, and re-running is idempotent. Verified on SQLite, MySQL 8, and PostgreSQL 16.

### The CIP-0163 half

Not fixed here, as you suggested, and filed as #3021 rather than left in a comment. You are right that it is the same class and that the filter cannot catch it: the guard decision never reaches `Spendable`, so a guarded row stays `spendable = true`. The issue proposes persisting the guard onto the row the same way `Spendable` already is, since having the API layer re-derive activation state and inactivity windows at read time would duplicate ledger logic in the read path. Noted at the filter and in `DATABASE.md` so it is discoverable from the code too.

## 2. Epoch label

Resolved as the earned epoch, `snapshot + 1`. Your derivation holds: `stakeRewardEpochsForNewEpoch` gives `snapshot = newEpoch-3`, rewards credit at the boundary into `newEpoch` so `spendable_epoch = newEpoch`, and db-sync models `spendable_epoch = earned_epoch + 2`, putting `earned_epoch` at `newEpoch-2` — the `performance` epoch, one above what was stored.

The comment at the mapping site traces that chain explicitly and names which of the three epochs in `stakeRewardEpochs` the field is, since you were right that the column name alone will not tell the next reader. Pagination and ordering are unaffected, still driven by the stored snapshot epoch. Existing epoch assertions updated.

## 3. Reward-type map

Took option 2. `blockfrostRewardTypes` is now a package-level allow-list, built once rather than per request, and an unrecognized value still passes through but logs at warn — so dingo producing a type outside the closed OpenAPI enum becomes visible instead of silently emitting a schema-invalid response. `TestAccountRewardHistoryTypeMapping` now captures slog output and asserts exactly one WARN for an unknown value and none for the three recognized ones, so it actually pins the behavior rather than passing either way.

## 4. Count/page snapshot consistency

Fixed. One `db.Transaction(false)` threaded through the account lookup, the count, and the page, matching the `PoolMetadata` pattern. The duplicate-or-skip window at an epoch boundary is gone.

## 5. Growth estimate in bytes

Added. Base row is roughly 100-110 bytes, and each of the two indexes is comparable in width to the row itself since both carry their own copies of the 28-byte `staking_key` and `pool_key_hash`, so real per-row cost in `api` mode is about 250-300 bytes against the ~100-110 the row count alone implies — roughly 300-400MB of unbounded growth per mainnet epoch. Updated for the now-6-column index.

## Verification

`go build` (both tag sets), `go test ./api/blockfrost/... ./database/... ./ledger/...` with 0 `^FAIL` and 0 `^--- FAIL`, `golangci-lint run` 0 issues, `nilaway`, `make import-boundaries`, `go test -race`, and conformance all pass. The full mysql and postgres metadata suites were also run under `-tags dingo_extra_plugins` against real containers given the multi-backend index change, both clean; containers removed.

`DATABASE.md` covers the index change, the spendable rationale, the epoch derivation, the allow-list, the migration path, the CIP-0163 gap, and the bytes clause. `ARCHITECTURE.md` reassessed and unaffected.

wolf31o2:
Verified against current code and fixed in 70d318f1. The finding was still valid.

The `spendable = true` filter was documented thoroughly in `rewardstate.GetAccountOutputsByCredential`, but not on the `MetadataStore` interface — which is what callers and any future backend implementer actually read. The interface contract said only "across every epoch that has not yet been pruned, paginated and ordered by epoch", which reads as returning every retained row for the credential.

`GetRewardAccountOutputsByCredential` now states that only spendable rows are returned and why (a `spendable = false` row was never credited; crediting skips it and adds the amount to the epoch unspendable total), with a pointer to the implementation for the full rationale and to #3021 for the known CIP-0163-guarded gap.

Extended it to `CountRewardAccountOutputsByCredential` as well, which had the identical gap. That one is arguably the more important of the two: a count that disagrees with the filter is exactly the pagination-advertising-unpaid-rewards defect from the first review pass, so the interface now records that the two must agree.

Documentation only, no behavior change. `go build`, `golangci-lint run ./database/...` (0 issues), and the full `database/plugin/metadata/...` suites pass with 0 failures.

chrisguiney (approved):

PR #3016, opened 2026-07-29
feat(blockfrost): add stake-account utxo, withdrawal, and transaction endpoints

### LINKED ISSUE #2934: blockfrost: implement stake-account UTxO, withdrawal, and transaction endpoints
## Summary

Implement the missing Blockfrost stake-account activity endpoints:

- `GET /accounts/{stake_address}/utxos`
- `GET /accounts/{stake_address}/withdrawals`
- `GET /accounts/{stake_address}/transactions`

The existing account routes cover account info, associated addresses, delegations, registrations, and rewards, but these three routes are not registered.

Source of truth: Blockfrost OpenAPI `0.1.90` (`openapi.json` in `blockfrost/openapi`).

### Account UTxOs

Return the current UTxOs controlled by the stake credential. Each row must contain:

- `address`
- `tx_hash`
- deprecated `tx_index` and current `output_index`
- `amount`
- `block`
- nullable `data_hash`
- nullable `inline_datum`
- nullable `reference_script_hash`

Reuse the existing address/UTxO conversion behavior so datum and reference-script fields remain consistent with `/addresses/{address}/utxos` and `/txs/{hash}/utxos`.

### Account withdrawals

Return withdrawal history rows containing:

- `tx_hash`
- `amount`
- `tx_slot`
- `block_time`
- `block_height`

### Account transactions

Return transactions associated with addresses controlled by the stake credential. Each row must contain:

- `address`
- `tx_hash`
- `tx_index`
- `block_height`
- `block_time`

Support the OpenAPI `from` and `to` inclusive block-number filters, including the optional `block:index` form.

## Acceptance criteria

- All three routes are registered under `/api/v0`.
- Bech32 stake addresses are validated consistently with the existing account handlers.
- `count`, `page`, and `order` work consistently with other Blockfrost list endpoints.
- `/transactions` implements and validates `from`/`to`, including malformed and inverted ranges.
- Results have deterministic blockchain ordering and do not duplicate a transaction/address association.
- Missing accounts and invalid inputs use the existing Blockfrost error response conventions.
- Handler and adapter tests cover response fields, pagination/order, empty results, invalid input, and database/query failures.

## Related

- Original account endpoint issue: #1368
- Address transaction history index: #1362
- UTxO response-field work: #2484

wolf31o2:
Closes #2934

Registers three missing Blockfrost stake-account endpoints:

- `GET /api/v0/accounts/{stake_address}/utxos`
- `GET /api/v0/accounts/{stake_address}/withdrawals`
- `GET /api/v0/accounts/{stake_address}/transactions`

Response shapes verified against the `blockfrost/openapi` schema files directly (`account_utxo_content`, `account_withdrawal_content`, `account_transactions_content`) rather than from the issue text.

One thing worth noting from the schema: `account_transactions_content`s own example lists the same `tx_hash` twice under two different payment addresses that share one delegation part, so a row is per (transaction, address) pair rather than per transaction. The implementation matches that.

## Data sources

Account UTxOs reuse `models.UtxoAddressPattern{DelegationPart: ...}`, which the shared UTxO query layer already supports, plus the existing `addressUtxoCbor` / `utxoDatumAndScriptRef` helpers so datum and reference-script fields stay consistent with `/addresses/{address}/utxos` and `/txs/{hash}/utxos`. Each rows `address` is recovered per row because one stake credential spans many payment addresses.

Account withdrawals query `account_reward_delta` (`withdrawal = true`) joined to `transaction` on `tx_hash = hash`; only a `SUM`-based query existed before. The join cannot fan out because `transaction.hash` is unique.

Account transactions query `address_transaction` directly, which is already one row per (payment address, transaction) with its own `slot` and `tx_index` — the same granularity the response needs.

Unlike `AccountAssociatedAddresses`, the payment-credential script-versus-key bit is resolved from `utxo.payment_script` rather than assuming the payment part is always a key hash, so script addresses render correctly.

## Query cost

Each endpoints bound is stated deliberately rather than left implicit.

`/transactions` is bounded by page size. `ORDER BY`, `LIMIT`/`OFFSET`, and the `from`/`to` range are all pushed into SQL; `from`/`to` block numbers resolve to slots via two bounded index lookups; and the script bit, block height, and block time are resolved only for the pages own rows and distinct blocks.

An earlier revision of this branch fetched the accounts entire transaction history with all UTxOs preloaded and filtered in memory, which was unbounded work driven by a path parameter. That is fixed.

`/withdrawals` is bounded by page size — `LIMIT`/`OFFSET` in SQL, with one memoized block lookup per distinct block on the page.

`/utxos` is NOT bounded: it fetches the credentials full live UTxO set and paginates in memory. This exactly matches the existing accepted behavior of `/addresses/{address}/utxos`, which uses the same helper without a SQL limit, so it is pre-existing rather than introduced here. Flagging it explicitly rather than leaving it unstated.

## Index

`/transactions` filters `(credential_tag, staking_key)` and orders by `slot, tx_index, payment_key`. `idx_addr_tx_staking` covered only the first two columns, so the plan was:

```
SEARCH at USING INDEX idx_addr_tx_staking (credential_tag=? AND staking_key=?)
USE TEMP B-TREE FOR ORDER BY
```

The index located the credentials rows, then SQLite sorted all of them before `LIMIT` could apply — proportional to the accounts full history, not the page. Added `idx_addr_tx_stake_position (credential_tag, staking_key, slot, tx_index, payment_key)`:

```
ASC:   SEARCH at USING INDEX idx_addr_tx_stake_position (credential_tag=? AND staking_key=?)
DESC:  SEARCH at USING INDEX idx_addr_tx_stake_position (credential_tag=? AND staking_key=?)
RANGE: SEARCH at USING INDEX idx_addr_tx_stake_position (credential_tag=? AND staking_key=? AND (slot,tx_index)>(?,?) AND (slot,tx_index)<(?,?))
```

`USE TEMP B-TREE FOR ORDER BY` is gone in all three cases. DESC was confirmed as a backward index scan rather than assumed.

The `from`/`to` predicate uses a row-value comparison `(at.slot, at.tx_index) >= (?, ?)` rather than the logically equivalent OR form, because only the row-value form gets folded into the index seek — the OR form leaves the bound as a row-by-row filter. Verified with `EXPLAIN QUERY PLAN` side by side.

`MigrateAddressTransactionStakePositionIndex` drops the now-redundant `idx_addr_tx_staking`, whose columns are a strict prefix of the new index. `address_transaction` is a hot per-block write path, so keeping both would cost write throughput for no remaining read benefit. `Slot`s standalone index is retained because the unscoped rollback cleanup (`WHERE slot > ?`) cannot use the composite index. Upgrade in place is verified with a legacy-model AutoMigrate test following the existing `MigrateRewardLiveStakePoolIndex` pattern.

## Tests

Handler and adapter tests covering response fields, pagination and order, empty results, invalid input, 404 and 500 paths, and `from`/`to` parsing and validation including malformed and inverted ranges.

`TestNodeAdapterAccountTransactionsBounded` creates five transactions in five blocks but a `Block` row for only the one on the requested page — the previous unbounded implementation resolved all five up front and would fail on the missing ones, and a `page=2` request is asserted to error to prove the setup would genuinely catch a regression.

Three `EXPLAIN QUERY PLAN` regression tests (asc, desc, range) assert the composite index drives the search and the temp B-tree is absent, so a future index change fails loudly instead of silently degrading. These follow the existing pattern in `sqlite/utxo_test.go`, `sqlite/drep_test.go`, and `internal/rewardstate/livestake_test.go`.

No existing test assertion was modified.

## Verification

`go build ./...` (default and `-tags dingo_extra_plugins`), `go test ./api/blockfrost/... ./database/... ./ledger/...`, `golangci-lint run`, `nilaway`, `modernize`, `make import-boundaries`, `go test -race`, and `go test ./internal/test/conformance/...` all pass.

Verified to merge cleanly onto current `main`. Note for merge ordering: this and #3015 both append new sections at the same point in `DATABASE.md`, so whichever merges second needs a trivial keep-both resolution there. No code conflicts.

## Documentation

`DATABASE.md` documents the new withdrawal and account-transaction query surfaces, the new index with its plan evidence, and the dropped redundant index. `ARCHITECTURE.md` documents the three new routes and how they resolve data.

---
## Summary by cubic
Adds Blockfrost stake-account UTxO, withdrawal, and transaction endpoints with SQL-backed pagination and inclusive block-range filtering. Improves performance with a new composite index and dialect-aware range predicates across SQLite, Postgres, and MySQL.

- **New Features**
  - Added GET `/api/v0/accounts/{stake_address}/utxos`, `/withdrawals`, and `/transactions`.
  - Transactions accept inclusive `from`/`to` block ranges (`block` or `block:index`) and return one row per (transaction, address).
  - Pagination: `/withdrawals` and `/transactions` in SQL; `/utxos` in memory to match address UTxO behavior.
  - Correctly reconstructs key vs script payment addresses via `utxo.payment_script`.

- **Bug Fixes**
  - Account transactions are page-bounded and use an index-backed range scan via new composite index `idx_addr_tx_stake_position`; old index dropped after AutoMigrate with existence checks.
  - Dialect-aware two-column range predicate: row-value form on SQLite/Postgres, expanded-OR on MySQL, with a safe fallback for unsupported operators.
  - Reduced per-request work by resolving `payment_script` and block data only for rows on the current page; removed full-history loads and UTxO CBOR decoding.
  - Regression tests pin EXPLAIN plans, range behavior, and pagination/validation across backends.

<sup>Written for commit af052ed18a63deeebf8b33cbb2250f762a40f5e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3016?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added account activity endpoints for stake-account UTxOs, withdrawal history, and transactions.
  - Added pagination and ascending/descending ordering support.
  - Added optional inclusive block-range filters for account transactions.
  - Responses include transaction, block, address, amount, datum, and reference-script details where available.
  - Added support for correctly reconstructing key-hash and script-hash payment addresses.

- **Documentation**
  - Expanded API and database documentation with endpoint behavior, pagination, filtering, and query examples.

- **Bug Fixes**
  - Improved handling of invalid ranges, missing accounts, empty results, and rollback-aware history data.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2312,12 +2312,33 @@ epochs/parameters, network/eras, genesis, assets, pools/extended, retiring
 pools, pool metadata, governance DRep list and lookup, address summary,
 address UTxOs and transactions, metadata label JSON/CBOR,
 transaction content/CBOR/metadata/UTxOs/certificates/redeemers/required
-signers, and account/delegation/registration/reward endpoints. It uses an
-adapter pattern to translate between Dingo's internal state and Blockfrost
-response types and supports Blockfrost-style pagination headers. The root
-document is served only at the literal `/` path (`GET /{$}`); any other
-unregistered path falls through to a catch-all `404` handler instead of the
-root document, matching real Blockfrost's behavior for unimplemented routes.
+signers, and account/delegation/registration/reward/UTxOs/withdrawals/
+transactions endpoints. It uses an adapter pattern to translate between
+Dingo's internal state and Blockfrost response types and supports
+Blockfrost-style pagination headers. The root document is served only at the
+literal `/` path (`GET /{$}`); any other unregistered path falls through to a
+catch-all `404` handler instead of the root document, matching real
+Blockfrost's behavior for unimplemented routes.
+
+The account UTxOs, withdrawals, and transactions endpoints resolve everything
+by stake credential rather than a single address. Account UTxOs reuse the
+UTxO address-pattern query with a delegation-part-only pattern (matching
+every payment address sharing the stake credential) and recover each row's
+exact payment address from decoded output CBOR, the same CBOR-derived
+datum/reference-script recovery `/addresses/{address}/utxos` uses. Account
+withdrawals read the rollback-aware `account_reward_delta` withdrawal journal
+joined to its transaction. Account transactions reuse the address-transaction
+credential lookup (payment key `NULL`) to get one row per matching
+transaction, then derive the set of distinct addresses each transaction
+associates with the stake credential from the transaction's already-loaded
+input/collateral/reference-input/output/collateral-return UTxOs (mirroring
+the same grouping the indexer uses to populate `address_transaction`), one
+response row per (transaction, address) pair. Its optional `from`/`to`
+block-range filter cannot be pushed into SQL because block height is not
+part of the metadata schema (only block hash and slot are persisted on the
+transaction row); every transaction associated with the credential is
+fetched in chain order, resolved against the block store for its height,
+filtered, and paged in memory.
 
 Address summaries run balance, asset, CBOR, and existence reads through one
 coordinated read transaction. Pointer addresses require an exact decoded-output
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2325,20 +2325,36 @@ by stake credential rather than a single address. Account UTxOs reuse the
 UTxO address-pattern query with a delegation-part-only pattern (matching
 every payment address sharing the stake credential) and recover each row's
 exact payment address from decoded output CBOR, the same CBOR-derived
-datum/reference-script recovery `/addresses/{address}/utxos` uses. Account
-withdrawals read the rollback-aware `account_reward_delta` withdrawal journal
-joined to its transaction. Account transactions reuse the address-transaction
-credential lookup (payment key `NULL`) to get one row per matching
-transaction, then derive the set of distinct addresses each transaction
-associates with the stake credential from the transaction's already-loaded
-input/collateral/reference-input/output/collateral-return UTxOs (mirroring
-the same grouping the indexer uses to populate `address_transaction`), one
-response row per (transaction, address) pair. Its optional `from`/`to`
-block-range filter cannot be pushed into SQL because block height is not
-part of the metadata schema (only block hash and slot are persisted on the
-transaction row); every transaction associated with the credential is
-fetched in chain order, resolved against the block store for its height,
-filtered, and paged in memory.
+datum/reference-script recovery `/addresses/{address}/utxos` uses; like
+`/addresses/{address}/utxos`, pagination happens after fetching the
+credential's full live-UTxO set (there is no SQL `LIMIT`), which is an
+existing, accepted characteristic of that query shape rather than something
+introduced for this endpoint. Account withdrawals read the rollback-aware
+`account_reward_delta` withdrawal journal joined to its transaction, with
+`LIMIT`/`OFFSET` applied in SQL.
+
+Account transactions is bounded by the requested page size, not by the
+credential's full transaction history: `address_transaction` already carries
+one row per (payment address, transaction) association with its own
+`slot`/`tx_index` columns (populated by the same indexing step that fans a
+transaction's inputs/collateral/reference-inputs/outputs/collateral-return
+out into that table), so the query pages directly against it with SQL
+`ORDER BY`/`LIMIT`/`OFFSET` and an inclusive `(slot, tx_index)` range
+predicate for `from`/`to` — no application-level fan-out or filtering
+happens after the query returns. A block number in `from`/`to` is resolved
+to its slot via two bounded index lookups (`Database.BlockByIndex`,
+`Database.BlockAtOrAfterIndex`) rather than a scan: an unresolvable `from`
+(beyond every known block) makes the range unsatisfiable and short-circuits
+to an empty result; an unresolvable `to` degrades to unconstrained on that
+side rather than guessing at a boundary that cannot be looked up backward.
+The payment-credential script/key bit needed to reconstruct each row's
+exact address, and the block height/time needed for its response fields,
+are then resolved only for the page's own (<= page size) distinct payment
+keys and blocks — `MetadataStore.GetUtxoPaymentScriptByCredential` looks up
+the small bounded set of payment keys against the `utxo` table's persisted
+`payment_script` column, avoiding both a full-history scan and a CBOR
+decode, and correctly distinguishing key-hash from script-hash payment
+credentials (`AccountAssociatedAddresses` elsewhere assumes key-hash only).
 
 Address summaries run balance, asset, CBOR, and existence reads through one
 coordinated read transaction. Pointer addresses require an exact decoded-output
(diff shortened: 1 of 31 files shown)


wolf31o2:
Both actionable findings addressed, plus both nitpicks, in c0782117 and cfc1ffa1.

## 1. Row-value predicate on MySQL

This was a real cross-backend problem and the MySQL bug references were the right lead. Measured rather than reasoned: scratch MySQL 8 and PostgreSQL 16 containers, 200k rows across 50 credentials, composite index matching `idx_addr_tx_stake_position`, `EXPLAIN`/`EXPLAIN ANALYZE` for both predicate forms.

MySQL 8:
- row-value form: `type: ref`, `used_key_parts: [credential_tag, staking_key]` only, actual rows scanned 1100 for a `LIMIT 100` request
- expanded OR form: `type: range`, all 4 key parts, actual rows scanned exactly 100

PostgreSQL 16:
- row-value form: `Index Only Scan`, full `ROW(slot, tx_index)` bound inside `Index Cond`, cost 8.13, no filter
- expanded OR form: `Index Cond` on the two leading columns plus a `Filter` removing 1000 rows, cost 15.84

So the two backends want opposite forms, and the previous single hardcoded form was pessimal on MySQL exactly as you predicted. SQLite still needs the row-value form, per the existing plan tests.

Fix is dialect-conditional via a new `sqldialect.TwoColumnRangeCondition`, following the `TransactionTableName` pattern: row-value for sqlite/postgres, expanded OR for mysql. The recorded plans and the MySQL bug numbers are in the helper doc comment so the next reader does not have to rediscover why the forms differ.

`DATABASE.md`'s `GetAddressTransactionsByCredential` section now states the per-backend evidence instead of the blanket page-proportional claim, which is the right correction: that claim was only ever substantiated on SQLite.

Also hardened the helper while in there. The expansion is only equivalent for the two inclusive operators it documents; an out-of-contract operator would previously have kept the `>` leading comparison regardless and returned silently wrong rows on MySQL. Unsupported operators now fall back to the row-value form, which is correct everywhere and merely loses the optimization.

New pinning tests cover all three dialects (`TestTwoColumnRangeConditionDialects`, `TestAddressTransactionRangeQueryPerDialectForm`) alongside the existing sqlite `EXPLAIN QUERY PLAN` tests.

Both scratch containers were removed after collection.

## 2. Migration ordering

Confirmed and fixed. All three backends called `MigrateAddressTransactionStakePositionIndex` before the `AutoMigrate` that creates `idx_addr_tx_stake_position`, so a failure between the two steps would have left the table with neither index. The call now runs after `AutoMigrate` in sqlite, postgres, and mysql, and the helper additionally verifies the replacement exists before dropping the legacy index as defense in depth against a future ordering mistake.

`TestMigrateAddressTransactionStakePositionIndex_ReplacementMissing` proves the drop is skipped when the replacement is absent, and the existing migration test now calls `AutoMigrate` first to match corrected production order.

## Nitpicks

Both done. The page-2 assertion now uses `require.ErrorIs(t, err, models.ErrBlockNotFound)` rather than a bare `require.Error` — you were right that a bare assertion undercut the "this fixture would have caught the bug" argument, since any unrelated failure satisfied it.

The three mock paginators are collapsed into a generic `mockPage[T]` helper: ~45 duplicated lines into 15, with the per-endpoint specifics (error field, backing slice, the nested `Pagination` field on `AccountTransactions`) still explicit at each call site.

## Verification

`go build` (both tag sets), `go test ./api/blockfrost/... ./database/... ./ledger/...` with 0 `^FAIL` and 0 `^--- FAIL`, `golangci-lint run` 0 issues, `nilaway`, `modernize` (only pre-existing hits in untouched `ledger/*_test.go`), `make import-boundaries`, `go test -race ./api/blockfrost/...`, and conformance all pass.

chrisguiney (approved):

PR #3017, opened 2026-07-29
docs(examples): gov lens rewards and drep expiry

wolf31o2:
Surfaces ledger and reward state in the gov lens example that landed since it was last updated.

## DRep inactivity expiry

DRep activity is now renewed from vote and update certificates, which makes `drep.expiry_epoch` trustworthy. Each DRep reports an expiry status and the epochs remaining, using the tally rule that `expiry_epoch > 0 AND expiry_epoch <= currentEpoch` is expired, and that a zero `expiry_epoch` has never been active and so is not expired.

Adds an `expiry=expired|active` filter, an Expiry column, and an expired DRep count in `/api/status`. The point is that a DRep can be registered and active while being expired by inactivity, so it is excluded from the voting power denominator. The old UI could not show that.

## DRep certificate provenance

The detail endpoint reports the credential's first on-chain appearance, taken as the earliest of the registration, update, and four vote delegation tables so it survives re-registration, plus the most recent real registration slot. Rows with `certificate_id = 0` are synthetic Mithril import rows and are excluded.

Kept on the detail endpoint rather than the list: resolving it for every DRep needs a six-table CTE, and the list query already does two grouped LEFT JOINs.

## Epochs endpoint and tab

New `GET /api/epochs` over the per-epoch reward record: the pots each epoch was paid from, epoch summary stake and pool totals with their snapshot-ready flag, the reward-side stake basis with its authoritative flag, and a per-epoch pool output count. The reward-side totals differ from the leader-election totals in `epoch_summary`, which is why both are shown.

## Stake lookup

Per-epoch reward outputs, and withdrawal witnesses joined to their reward deltas. Witnesses that moved no reward are marked as such, since a zero-amount withdrawal still counts as account activity and that is precisely why the witness table exists.

## CIP-0163 account expiry

Reads the account expiration epoch together with per-credential activation membership, falling back to membership when the sync-state marker has been cleared. Reports the gate as not enabled rather than implying the column is meaningful while it is off.

## Nullable column handling

`drep.expiry_epoch`, `drep.last_activity_epoch`, `account.expiration_epoch`, and the epoch listing columns are nullable but were scanned straight into `uint64`, so a NULL would fail the request. The columns this change reads or filters on are now coalesced and degrade to an unknown status. Dingo does not write NULLs there today, so this is defensive.

## Notes

No read-only role change is needed. Both grant scripts already use `GRANT SELECT ON ALL TABLES IN SCHEMA public` plus `ALTER DEFAULT PRIVILEGES`, which covers tables Dingo creates later, so a table list here would only rot.

Per-account reward history cannot exceed the retained window: `reward_account_output` and `reward_stake_input` are pruned at `epoch < currentEpoch-3` on every epoch transition in all storage modes. The UI and README state this, so absent rows do not look like a missing calculation.

Helm values move from 0.64.0 to 0.68.0, and the README records the version floor. Full epoch history additionally needs a build newer than 0.68.0, since retaining `epoch_summary` and the per-epoch reward tables for the life of the database is not in a tagged release yet.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean
- `go test -count=1 ./...` passes, 13 tests (5 pre-existing, 8 added). No existing test or assertion was modified.
- `golangci-lint run ./...` reports the same finding set as the base commit
- Exercised against a throwaway Postgres whose schema came from `models.MigrateModels` AutoMigrate, covering every endpoint including both DRep detail cases, the expiry filters, a rejected filter value, a synthetic-registration credential, a zero-amount witness, and a NULL-valued row

`DATABASE.md` and `ARCHITECTURE.md` checked and not affected. This is confined to an example app in its own module and changes no models, migrations, store surfaces, EventBus topics, or component boundaries.

---
## Summary by cubic
Adds DRep inactivity expiry and per‑epoch reward/ledger insights to the `dingo-gov-lens` example so users can see expired DReps, stake snapshots, and reward provenance at a glance. Also surfaces CIP‑0163 account expiry, per‑epoch rewards, and withdrawal witnesses in stake lookup.

- **New Features**
  - DRePs: expiry status and epochs remaining; new `expiry=expired|active` filter, Expiry column, and `expiredDrepCount` in `/api/status` (independent of registration).
  - DRep detail: first on‑chain appearance and last real registration slot (excludes synthetic Mithril rows).
  - Epochs: new `GET /api/epochs` endpoint and tab showing ADA pots, leader vs reward-basis stake, snapshot/authority flags, and per-epoch pool result count.
  - Stake lookup: per‑epoch reward outputs (retained window), withdrawal witnesses with zero‑amount marking, `expirationEpoch`, and activation membership; `/api/status` adds `latestRewardEpoch` and `accountInactivity`.
  - Robust null handling for nullable columns (expiry/activity/account/epoch fields degrade to “unknown” instead of failing).

- **Dependencies**
  - Bump `ghcr.io/blinklabs-io/dingo` to `0.68.0` (DRep activity renewal; CIP‑0163 fields). Full epoch history requires a build newer than `0.68.0`.

<sup>Written for commit d6af18c33000589b0d2d6abba9c4545918a0b84c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3017?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/dingo-gov-lens/README.md
+++ b/examples/dingo-gov-lens/README.md
@@ -13,12 +13,61 @@ dependencies live in this directory's `go.mod`.
 - Chain freshness from `tip`, `epoch`, and `node_settings`
 - Governance actions from `governance_proposal`
 - Vote breakdowns from `governance_vote`
-- Active DReps from `drep`
+- Active DReps from `drep`, plus how many registered DReps have expired by
+  inactivity and are therefore excluded from the voting-power denominator
 - DRep vote, update, registration, and delegation views
+- Per-DRep inactivity expiry: the epochs left before `expiry_epoch` is reached,
+  a filter for expired vs unexpired DReps, the credential's first on-chain
+  appearance, and the slot of its most recent real registration certificate
+- Retained epoch and reward state per epoch: ADA pots, leader-election and
+  reward-basis stake totals, pool and delegator counts, whether the boundary
+  snapshot was captured authoritatively, and how many per-pool reward results
+  exist
 - Stake credential lookup from `account`, either pasted manually or derived
   locally from a CIP-30 wallet reward address, including the immutable slot at
-  which Dingo first observed the account
+  which Dingo first observed the account, its CIP-0163 inactivity expiry, its
+  per-epoch reward outputs, and its reward-withdrawal witnesses
 - Links out to Preview GovTool for proposal and voting workflows
+
+### Inactivity And Expiry
+
+Two separate expiry mechanisms show up in the UI, and neither is the same thing
+as a registration being active.
+
+A DRep's `active` flag only clears on deregistration. Inactivity expiry is
+separate: a DRep vote, registration, or update certificate sets
+`last_activity_epoch` and pushes `expiry_epoch` out by the protocol's DRep
+inactivity period, and the Conway tally drops a DRep whose `expiry_epoch` has
+been reached. The DReps table shows both, so a DRep can read active and expired
+at the same time. The expiry filter uses the same test the tally uses: a zero
+`expiry_epoch` means no activity has been recorded yet and counts as unexpired.
+
+Reward-account inactivity is the CIP-0163 equivalent for stake credentials, held
+behind a node flag. With the flag off, every `account.expiration_epoch` is 0 and
+the stake lookup reports the expiry as not enabled. Enabling it also requires a
+from-genesis sync, because the column cannot be reconstructed from a Mithril
+snapshot, so the Compose stack in this directory always reports it as not
+enabled. The withdrawal witnesses below it are recorded either way: Dingo
+records a witness for every valid reward-withdrawal map entry, so a withdrawal
+that moved no reward still appears, marked as a zero amount.
+
+### Retention
+
+Epoch and reward history is retained asymmetrically, which decides how far back
+each panel can look:
+
+- Retained for the life of the database: `epoch`, `epoch_summary`,
+  `reward_ada_pots`, `reward_snapshot`, `reward_pool_input`, and
+  `reward_pool_output`. These are one row per epoch, or roughly one row per pool
+  per epoch, and they are what the Epochs panel reads.
+- Pruned at every epoch transition to the last four epochs:
+  `pool_stake_snapshot`, `reward_stake_input`, and `reward_account_output`.
+  These scale with delegator count.
+
+So the Epochs panel reaches back to the first boundary the node captured, while
+the per-epoch reward outputs in the stake lookup only cover the current epoch and
+the three before it. An older epoch showing no reward rows for an account is
+retention, not a missing calculation.
 
 ## Bootstrap Dingo Quickly
 
@@ -120,6 +169,20 @@ For Kubernetes, start from `k8s/dingo-values.yaml`. Replace the Postgres DSN
 with your cluster-local database service and mount the configured CA certificate
 from a Secret in your own chart overlay.
 
+### Dingo version
+
+Use 0.68.0 or newer. DRep activity renewal from certificates landed in 0.68.0,
+and without it `drep.expiry_epoch` is not maintained from vote and update
+certificates, so the expiry column and the `expiry` filter report stale
+answers rather than wrong-looking ones. CIP-0163 account expiry columns need
+0.67.0, and the reward metadata tables need 0.65.0.
+
+The Epochs tab reaches only the last four epochs on a released node. Retaining
+`epoch_summary` and the per-epoch reward tables for the life of the database is
+newer than 0.68.0 and not in a tagged release yet; before it, those rows were
+pruned to `epoch < current-3` at every epoch transition. The tab works either
+way, it just shows a short window until you run a node built from `main`.
+
 ## Read-Only Database User
 
 Create an application role that can only read the Dingo metadata schema:
@@ -131,6 +194,14 @@ psql "$ADMIN_DATABASE_URL" -f sql/create-readonly-user.sql
 Edit the SQL first so the database name, Dingo owner role, and password match
 your environment.
 Use the read-only role in the app's `DATABASE_URL`.
+
+The grants are schema-wide rather than a table list: `GRANT SELECT ON ALL TABLES
+IN SCHEMA public` covers whatever exists when the script runs, and the
+`ALTER DEFAULT PRIVILEGES FOR ROLE <dingo owner>` line covers every table Dingo
+creates later. Tables added by a Dingo upgrade, including the reward and
+inactivity tables this app reads, are therefore readable without editing the
+script. The same is true of `docker/postgres-init/001-create-readonly-user.sh`,
+which runs at cluster init before Dingo has created any table.
 
 ## Run The App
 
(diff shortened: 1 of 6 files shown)


examples/dingo-gov-lens/main.go
```diff
@@ -1035,6 +1484,77 @@ func actionTypeName(v int64) string {
...
 
+// drepExpiredPredicate and drepUnexpiredPredicate mirror the Conway tally's
+// inactivity test: a DRep is expired once expiry_epoch is set and has been
+// reached, and a zero expiry_epoch means no activity has been recorded yet and
+// is treated as unexpired. Both are independent of drep.active, which only
+// clears on deregistration.
+// expiry_epoch is a nullable column, so every comparison folds NULL into the
+// unset value first.
+const (
+	drepExpiredPredicate = `COALESCE(d.expiry_epoch, 0) > 0
```
chrisguiney, line 1495: Crediting this because it's the thing I went looking to break and couldn't. An example that restates a consensus rule usually drifts from it.

dingo's authoritative predicate is `drepActiveAtEpoch` (`ledger/governance/epoch.go:610-613`), used by the tally at `tally.go:262-268`:

```go
drep.ExpiryEpoch == 0 || drep.ExpiryEpoch > currentEpoch  // active
```

Both of these are its exact complement, and the detail that matters is the one that's easy to miss: `drepUnexpiredPredicate` keeps `expiry_epoch = 0` on the **active** side. So `expiry=active` returns the same set the tally counts in its voting-power denominator, rather than quietly dropping never-active DReps. Getting that backwards would have made the filter disagree with the very thing the change exists to illustrate.

The `COALESCE(..., 0)` folding also lines up with the Go path: an empty `epoch` table gives a current epoch of 0 here, so nothing matches as expired, while `drepExpiryState` reports `"unknown"` — both non-expired, no contradiction between filter and display.

One more thing I checked: the SQL subquery and `latestEpochID` (`:1332`) both read `SELECT MAX(epoch_id) FROM epoch`, so a row can't be filtered as expired while rendering as active.

examples/dingo-gov-lens/main_test.go
```diff
@@ -90,6 +93,170 @@ func TestVoteBackfillPending(t *testing.T) {
...
+	}
+
+	status, remaining = drepExpiryState(125, epoch(120))
+	if status != "active" || remaining == nil || *remaining != 5 {
+		t.Fatalf("unexpired = (%q, %v), want (active, 5)", status, remaining)
+	}
+
+	// The Conway tally treats expiry_epoch <= current epoch as expired, so
+	// an equal epoch is already expired rather than about to expire.
+	status, remaining = drepExpiryState(120, epoch(120))
```
chrisguiney, line 140: This is the assertion that makes the duplicated rule safe to duplicate, so it's worth pointing at.

`expiry_epoch <= currentEpoch` versus `<` is the single character that would silently desynchronise this example from `drepActiveAtEpoch`, and it fails in the least visible way possible: one DRep in or out of the denominator, with no error anywhere. Pinning `drepExpiryState(120, epoch(120)) == ("expired", 0)` and citing the tally rule in the comment above it is exactly the right treatment.

The `math.MaxUint64` case below is a good addition too — reporting an unrepresentable epoch as `"unknown"` rather than letting the `int64` conversion produce a plausible-looking negative remainder.

examples/dingo-gov-lens/main.go
```diff
@@ -977,6 +1261,171 @@ func (a *app) drepHistory(
...
+	var firstSeen, lastRegistration sql.NullInt64
+	// certificate_id filters out the synthetic registration rows a Mithril
+	// ledger-state import writes at the bootstrap slot; only real on-chain
+	// certificates count as a registration.
+	err := a.db.QueryRowContext(ctx, `
+		SELECT MIN(first_slot), MAX(registration_slot)
+		FROM (
+			SELECT
+				MIN(added_slot) AS first_slot,
+				MAX(
```
chrisguiney, line 1283: `first_slot` here is a plain `MIN(added_slot)` over all of `registration_drep`, while the `certificate_id != 0` filter applies only to the `registration_slot` aggregate below it.

The comment above the query is precise about that — *"only real on-chain certificates count as a **registration**"* — but the PR description generalises it to *"Rows with `certificate_id = 0` are synthetic Mithril import rows and are excluded"*, which reads as covering both. It doesn't.

The practical effect is on the deployment most likely to be running this: on a Mithril-bootstrapped node, the synthetic registration row sits at the bootstrap slot, so "first seen" reports the import slot rather than a first on-chain appearance. That undercuts the stated purpose — surviving re-registration — because every credential imported by the snapshot gets the same answer.

Either apply the same filter to `first_slot` (the other five branches contribute real certificates, so the union would still find a genuine earliest appearance where one exists), or say in the UI that first-seen is bounded below by the bootstrap slot. The second is fine if the intent really is "first slot this node ever saw the credential" — it just isn't what the description says.

examples/dingo-gov-lens/README.md
```diff
@@ -13,13 +13,62 @@ dependencies live in this directory's `go.mod`.
...
+### Retention
+
+Epoch and reward history is retained asymmetrically, which decides how far back
+each panel can look:
+
+- Retained for the life of the database: `epoch`, `epoch_summary`,
+  `reward_ada_pots`, `reward_snapshot`, `reward_pool_input`, and
+  `reward_pool_output`. These are one row per epoch, or roughly one row per pool
+  per epoch, and they are what the Epochs panel reads.
+- Pruned at every epoch transition to the last four epochs:
```
chrisguiney, line 63: This list will be wrong for `api` storage mode as soon as #3015 lands: that PR retains `reward_account_output` without bound in `api` mode while continuing to prune `reward_stake_input` in both. The description's *"pruned at `epoch < currentEpoch-3` on every epoch transition in all storage modes"* has the same issue.

Accurate today against 0.68.0 and `main`, so this isn't a defect in what's shipping — it's a staleness risk on docs that would then be actively misleading, since the whole point of stating the window is so absent rows don't look like a missing calculation.

The fix is already patterned two sections down. `:179-184` scopes the `epoch_summary` retention claim to a version instead of stating it flat, and reads correctly either way. The same treatment here would hold. Better still, the app already reads and displays `storageMode` (`main.go:46`, `:338`, `app.js:75`), so this one can be conditioned on the mode rather than on a version at all.

`static/app.js:332`'s empty-state text ("No per-epoch reward rows in the retained four-epoch window.") needs the same look. `static/index.html:115` is fine as written, because "per-delegator rows behind them are pruned" stays true of `reward_stake_input` either way — worth keeping that distinction when you reword the other two.

examples/dingo-gov-lens/main.go
```diff
@@ -774,9 +965,102 @@ func (a *app) handleStakeLookup(w http.ResponseWriter, r *http.Request) {
...
+			COALESCE(es.total_active_stake::text, ''),
+			COALESCE(es.total_pool_count, 0),
+			COALESCE(es.total_delegators, 0),
+			COALESCE(es.snapshot_ready, false),
+			COALESCE(rs.total_active_stake::text, ''),
+			COALESCE(rs.authoritative, false),
+			(
+				SELECT COUNT(*)
+				FROM reward_pool_output rpo
+				WHERE rpo.epoch = e.epoch_id
```
chrisguiney, line 1017: Not a problem, just recording what I checked so it doesn't have to be rechecked later. This subquery runs once per returned row, so I looked at whether it can scan.

It can't: `reward_pool_output` carries `idx_reward_pool_output_epoch_pool` with `Epoch` at priority 1 (`database/models/reward_state.go:134-136`), so each `COUNT(*)` is an index range scan rather than a table scan. Worst case at `limit=250` on a mainnet node with full history is roughly 250 × ~3000 index entries — tens of milliseconds, and the default limit of 50 is a fifth of that.

If it ever shows up, a single `LEFT JOIN (SELECT epoch, COUNT(*) FROM reward_pool_output GROUP BY epoch)` replaces N range scans with one pass. Not worth doing pre-emptively for a demo app.

The comment above the query is also accurate — every table joined here is retained for the life of the database, and the per-credential tables are indeed the pruned ones.

chrisguiney (approved):
Approving. This is confined to `examples/dingo-gov-lens`, which is its own module, and the two things it would be easy to get wrong here — the expiry rule diverging from the node's tally, and unindexed queries behind the new panels — both check out. Two minor notes inline, neither blocking.

## Verification performed

| Check | Result |
|---|---|
| `go build ./...`, `go vet ./...`, `gofmt -l .` | clean |
| `go test -count=1 ./...` | **pass**, 13 tests |
| `golangci-lint run ./...` | 7 issues — **identical set at the base commit** |
| Test-count claim | verified |
| Expiry rule vs `ledger/governance` tally | **matches**, boundary included |
| Index coverage of every new query | all predicates index-backed |

I couldn't exercise it against Postgres; you did, so I verified the SQL statically against the models instead.

Two claims I checked rather than took on trust:

- *"golangci-lint reports the same finding set as the base commit"* — true. Base and head both report `errchkjson: 1, gosec: 4, prealloc: 1, unparam: 1`, down to the same `writeJSON` unparam and the same `args := []any{}` prealloc.
- *"13 tests (5 pre-existing, 8 added). No existing test or assertion was modified."* — true. 13 `func Test` at head, and `git diff` shows zero removed lines in `main_test.go`.

## Performance

No degradation. Checked against the schema rather than assumed:

- **`/api/epochs`** is bounded by `boundedLimit(…, 50, 250)` with three 1:1 `LEFT JOIN`s. The per-row `COUNT(*)` is index-backed — see the note on `main.go:1017`.
- **The six-table provenance union** is six indexed lookups, not six scans: `idx_drep_reg_cred_slot` covers `(credential_tag, drep_credential, added_slot)` so its `MIN` is index-only, `idx_update_drep_credential` covers `update_drep`, and all four delegation tables carry `Drep []byte gorm:"index;size:28"`. That also confirms the reasoning for keeping this off the list endpoint — six probes × 250 rows is exactly what you avoided.
- **The expiry predicates** embed `(SELECT MAX(epoch_id) FROM epoch)`, which is uncorrelated, so Postgres evaluates it once as an InitPlan over the primary key.
- **`/api/status`**'s expired count is a full scan of `drep` with `active = true` — ~1500 rows on mainnet.

## Duplication

The expiry rule is necessarily restated rather than shared, since this example is a separate module and can't import dingo internals. Restating it with the tally cited in comments and the boundary pinned in a test is the right way to handle a duplication you can't avoid — see the note on `main_test.go:140`.

## Documentation

README and UI updated. `DATABASE.md` / `ARCHITECTURE.md` correctly assessed as not affected: no models, migrations, store surfaces, EventBus topics, or component boundaries are touched, and this lives in its own module.

No linked issue, which is fine for an example-app catch-up — I checked for a `Closes`/`Refs` to validate against and there isn't one.

PR #3018, opened 2026-07-29
docs(examples): explorer pools dreps and addresses

wolf31o2:
Covers the Blockfrost endpoints added since the explorer was last updated, and fixes a DRep view that had drifted out of sync with the response schema.

After this change the app calls every GET route in `api/blockfrost/blockfrost.go` except the API metadata root and `POST /api/v0/tx/submit`, which is a write.

## New views

- `GET /api/v0/pools/retiring`, as a Pending Retirements section on the pool list plus a Retiring epoch row on pool detail
- `GET /api/v0/pools/{pool_id}/metadata`, as a Pool Metadata section including the off-chain fetch error object. A pool that has left the active extended set now resolves from metadata alone instead of erroring. Dingo answers `{}` with 200 for a pool that registered no metadata anchor, so every field is optional.
- `GET /api/v0/governance/dreps`, as a paginated list ordered by delegated voting power, matching how the pools tab already works
- `GET /api/v0/addresses/{address}`, giving real total balance, address type, script flag, and a clickable stake address. This also wires up the previously dead stake-address click target, which nothing generated before.

## DRep response drift

The single DRep view was reading `registered`, `epoch`, and `live_stake`. None of those exist in the current `DRepResponse`. It also typed `active_epoch` as a plain number where the API returns `*uint64`, and omitted `retired`, `expired`, and `last_active_epoch` entirely. That view was already broken before this change.

It is now aligned field for field with `api/blockfrost/types.go`, with both epoch fields nullable. Predefined DReps carry no credential hash, so an empty `hex` is labelled predefined rather than registered.

## Fields populated earlier but still ignored by the UI

Block VRF, operational certificate and counter; transaction treasury donation; an input Kind column so reference and collateral inputs are distinguishable, which also surfaces the previously unrendered collateral flag; account DRep delegation and registration state, with the DRep clickable; delegation and registration deposits and block times; and the genesis update quorum.

`AddressUTXOResponse.tx_index` is typed with a comment but not rendered, since it is a deprecated alias for `output_index` carrying the same value.

## 404 handling

Unmatched routes return 404, so failures now carry their status. Fetch helpers swallow only 404 and rethrow everything else, and detail views distinguish Not found from Request failed.

The existing swallow-everything helpers are deliberately left alone: making them rethrow would break the transaction lookup's `Promise.all` on any single 500.

## Version floor

The two pool views need a Dingo newer than 0.68.0, because `GET /api/v0/pools/retiring` and `GET /api/v0/pools/{pool_id}/metadata` are not in a tagged release yet. Both use 404-tolerant fetches and report the endpoint as unavailable on an older node, and the rest of the explorer works normally.

Helm values move from 0.64.0 to 0.68.0, which predated every endpoint added here.

The shared `examples/README.md` gains the version floor it never had: it offered `DINGO_IMAGE=<tag>` with no guidance at all. It carries the cross-app note because the unreleased-endpoint caveat is the explorer's.

## Testing

- `npm ci`, `npx tsc --noEmit`, `npm run build` all clean
- Smoke-tested in a browser against a stub Blockfrost server whose responses were taken from `api/blockfrost/types.go`, covering the DRep list including a predefined and a retired DRep, single DRep, pool list with retirements, pool detail with metadata and the off-chain error note, the address summary with a clickable stake address, the block VRF and op-cert rows, and the 404 path. No JS exceptions.

No new dependencies and no framework. Every new view reuses existing CSS classes, so there are no style or markup changes. `DATABASE.md` and `ARCHITECTURE.md` checked and not affected.

---
## Summary by cubic
Adds new explorer views for pools, DReps, and addresses, and fixes the DRep detail to match the current Blockfrost API. Also surfaces more on-chain fields and adds clear 404 handling; docs set the Dingo version floor to 0.68.0.

- **New Features**
  - Pools: pending retirement schedule (`GET /api/v0/pools/retiring`) and pool metadata (`GET /api/v0/pools/{pool_id}/metadata`) with off‑chain fetch errors; pools outside the active set resolve from metadata.
  - Governance: DRep list (`GET /api/v0/governance/dreps`) ordered by voting power; single DRep view aligned to the API with retired/expired/last_active_epoch and predefined DReps labeled when `hex` is empty.
  - Addresses: address summary (`GET /api/v0/addresses/{address}`) showing balance, type, script flag, and linked stake address; stake‑address click wired; smart search uses the summary.
  - Extra fields: block VRF/op cert/counter, transaction treasury donation, input kind (spend/collateral/reference), account DRep delegation and registration state, delegation/registration deposits and times, and genesis update quorum.
  - Errors: 404‑aware fetchers; detail views show “Not found” vs “Request failed”; existing “swallow‑everything” helpers left as‑is to avoid breaking transaction lookups.

- **Migration**
  - Use Dingo 0.68.0+ (Helm values updated); building from this checkout is recommended.
  - `GET /api/v0/pools/retiring` and `GET /api/v0/pools/{pool_id}/metadata` require a build newer than 0.68.0; the explorer reports these sections as unavailable on older nodes.
  - No new dependencies; UI markup/styles unchanged.

<sup>Written for commit 0ffa77db2c746961a66e4af26573045efe1c7f8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -32,6 +32,13 @@ By default Compose builds Dingo from this checkout as
 `dingo-examples-dingo:local`. Set `DINGO_IMAGE=ghcr.io/blinklabs-io/dingo:<tag>`
 if you explicitly want to use a published image.
 
+Building from the checkout is the recommended path, because the examples track
+API work that reaches `main` before it reaches a release. If you do pin a
+published image, use 0.68.0 or newer: the address summary endpoint, the DRep
+list endpoint, and exact-address UTxO matching all landed in 0.68.0. The
+explorer's retiring-pools and pool-metadata views need a build newer than
+0.68.0 and report themselves as unavailable on anything older.
+
 The shipped database credentials are local-development defaults. Change
 `POSTGRES_PASSWORD` and `DINGO_GOV_LENS_PASSWORD` before exposing the stack
 outside a trusted development machine.
--- a/examples/dingo-blockfrost-explorer/src/main.ts
+++ b/examples/dingo-blockfrost-explorer/src/main.ts
@@ -29,6 +29,9 @@ type BlockResponse = {
   tx_count: number;
   output: string | null;
   fees: string | null;
+  block_vrf: string | null;
+  op_cert: string | null;
+  op_cert_counter: string | null;
   previous_block: string;
   next_block: string | null;
   confirmations: number;
@@ -64,6 +67,7 @@ type NetworkResponse = {
 
 type GenesisResponse = {
   active_slots_coefficient: number;
+  update_quorum: number;
   max_lovelace_supply: string;
   network_magic: number;
   epoch_length: number;
@@ -136,6 +140,7 @@ type TransactionResponse = {
   block: string;
   deposit: string;
   fees: string;
+  treasury_donation: string;
   slot: number;
   block_height: number;
   block_time: number;
@@ -165,6 +170,7 @@ type TransactionInputResponse = {
   amount: Amount[];
   output_index: number;
   collateral: boolean;
+  reference: boolean | null;
   data_hash: string | null;
   inline_datum: string | null;
   reference_script_hash: string | null;
@@ -196,9 +202,20 @@ type TransactionMetadataCBORResponse = {
   metadata: string;
 };
 
+type AddressResponse = {
+  address: string;
+  amount: Amount[];
+  stake_address: string | null;
+  type: string;
+  script: boolean;
+};
+
 type AddressUTXOResponse = {
   address: string;
   tx_hash: string;
+  // tx_index is Blockfrost's deprecated alias for output_index and
+  // carries the same value.
+  tx_index: number;
   output_index: number;
   amount: Amount[];
   block: string;
@@ -245,6 +262,8 @@ type AccountResponse = {
   treasury_sum: string;
   withdrawable_amount: string;
   pool_id: string | null;
+  drep_id: string | null;
+  registered: boolean;
 };
 
 type AccountAssociatedAddressResponse = {
@@ -256,11 +275,18 @@ type AccountDelegationHistoryResponse = {
   tx_hash: string;
   amount: string;
   pool_id: string;
+  tx_slot: number;
+  block_time: number;
+  block_height: number;
 };
 
 type AccountRegistrationHistoryResponse = {
   tx_hash: string;
   action: string;
+  deposit: string;
+  tx_slot: number;
+  block_time: number;
+  block_height: number;
 };
 
 type AccountRewardHistoryResponse = {
@@ -271,14 +297,59 @@ type AccountRewardHistoryResponse = {
 
 type DRepResponse = {
   drep_id: string;
+  // hex is empty for the predefined always-abstain and
+  // always-no-confidence DReps, which carry no credential hash.
   hex: string;
-  has_script: boolean;
-  registered: boolean;
-  epoch: number;
   amount: string;
   active: boolean;
-  active_epoch: number;
-  live_stake: string;
+  active_epoch: number | null;
+  has_script: boolean;
+  retired: boolean;
+  expired: boolean;
+  last_active_epoch: number | null;
+};
+
+type DRepListItemResponse = {
+  drep_id: string;
+  hex: string;
+  amount: string;
+  has_script: boolean;
+  retired: boolean;
+  expired: boolean;
+  last_active_epoch: number | null;
+  metadata: DRepMetadataResponse | null;
+};
+
+type DRepMetadataResponse = {
+  url: string;
+  hash: string;
+  json_metadata: unknown;
+  bytes: string;
+};
+
+type PoolRetiringResponse = {
+  pool_id: string;
+  epoch: number;
+};
+
+// Dingo answers with an empty JSON object for a pool that registered no
+// metadata anchor, so every field is optional here. error is only
+// present when the off-chain document fetch failed.
+type PoolMetadataResponse = {
+  pool_id?: string;
+  hex?: string;
+  url?: string | null;
+  hash?: string | null;
+  ticker?: string | null;
+  name?: string | null;
+  description?: string | null;
+  homepage?: string | null;
+  error?: OffchainFetchErrorResponse;
+};
+
+type OffchainFetchErrorResponse = {
+  code: string;
+  message: string;
 };
 
 type PoolExtendedResponse = {
@@ -314,6 +385,20 @@ type ErrorResponse = {
   error: string;
   message: string;
 };
+
+// HTTPError carries the response status so callers can separate a
+// Blockfrost 404 (unknown resource, or a route this Dingo build does not
+// serve) from a transport or server-side failure. Declared before the
+// bootstrap below because class bindings are not hoisted.
+class HTTPError extends Error {
+  readonly status: number;
+
+  constructor(status: number, message: string) {
+    super(message);
+    this.name = "HTTPError";
+    this.status = status;
+  }
+}
 
 type HTMLContent = {
   readonly html: string;
@@ -362,6 +447,8 @@ type AppState = {
   eras: NetworkEraResponse[];
   pools: PoolExtendedResponse[];
   poolTotal?: number;
+  retiringPools: PoolRetiringResponse[];
+  retiringPoolTotal?: number;
   poolsLoading: boolean;
   networkUpdatedAt?: number;
   poolsUpdatedAt?: number;
@@ -427,8 +514,8 @@ const tabConfig: Record<Tab, { label: string; placeholder: string; emptySearch: 
   },
   drep: {
     label: "DRep",
-    placeholder: "DRep bech32 ID or credential hex",
-    emptySearch: false,
+    placeholder: "DRep bech32 ID, credential hex, or empty for list",
+    emptySearch: true,
     visible: true,
   },
   metadata: {
@@ -454,6 +541,7 @@ const state: AppState = {
   blockTxPanelTitle: "Block Transactions",
   eras: [],
   pools: [],
+  retiringPools: [],
   poolsLoading: false,
   activeTab: "dashboard",
   loading: false,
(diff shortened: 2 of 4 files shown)


examples/dingo-blockfrost-explorer/src/main.ts
```diff
@@ -271,14 +297,59 @@ type AccountRewardHistoryResponse = {
 
 type DRepResponse = {
```
chrisguiney, line 298: Verified against `api/blockfrost/types.go:263-273` — this is an exact match, all nine fields, same names, same order, and both `*uint64` fields as `number | null`:

```
drep_id, hex, amount, active, active_epoch, has_script, retired, expired, last_active_epoch
```

Worth stating plainly because the old view was reading `registered`, `epoch`, and `live_stake`, none of which exist on this type — that view was rendering nothing useful and no compiler was going to say so, since the response arrives as untyped JSON. Type drift against a Go struct in another module is invisible until someone diffs them by hand.

I checked the neighbours for the same class of drift and found none. `DRepListItemResponse` matches 8/8 including `metadata: DRepMetadataResponse | null` against `*DRepMetadataResponse`. `AccountResponse.ActiveEpoch *int64` maps to `number | null`. And `AccountDelegationHistoryResponse.ActiveEpoch` is `int32`, not a pointer — so `active_epoch: number` there is correctly non-nullable rather than an oversight.

The `hex`-is-empty comment is a good call too: `drepStatusText` labelling that as `predefined` rather than falling through to `registered` is the kind of distinction that only shows up once you have the always-abstain DRep in front of you.

examples/dingo-blockfrost-explorer/src/main.ts
```diff
@@ -271,14 +297,59 @@ type AccountRewardHistoryResponse = {
...
+  hash: string;
+  json_metadata: unknown;
+  bytes: string;
+};
+
+type PoolRetiringResponse = {
+  pool_id: string;
+  epoch: number;
+};
+
```
chrisguiney, line 334: This comment is wrong about the API, and since an example app's comments are effectively API documentation it's worth correcting.

Dingo never answers `{}` here. `PoolMetadataResponse` (`api/blockfrost/types.go:305-315`) declares `PoolID string` and `Hex string` with **no** `omitempty`, and `NodeAdapter.PoolMetadata` sets both before the `metadataURL == ""` early return. The remaining six are pointers, also without `omitempty`, so they marshal as explicit `null`. An anchorless pool gets:

```json
{"pool_id":"pool1…","hex":"…","url":null,"hash":null,"ticker":null,"name":null,"description":null,"homepage":null}
```

`error` is the only field that can actually be absent — it's the one with `omitempty`.

No runtime consequence: `?: string | null` accepts both shapes, and the consumers all use `?? "-"` or truthiness (`!metadata.url` on `:2585`, `metadata.hash ? … : "-"` on `:2595`), which handle `null` and `undefined` identically. So this is about what the file teaches, not what it does.

Two small consequences if you reword it: `pool_id` and `hex` don't need `?` — they're guaranteed — and the `?` on the other six is right for a different reason than the comment gives, namely that they arrive as `null` rather than missing.

examples/dingo-blockfrost-explorer/src/main.ts
```diff
@@ -2790,6 +3101,32 @@ async function optionalFetchPage<T>(path: string): Promise<{ items: T[]; total?:
   }
 }
 
+// Resolves to undefined only for a 404. Every other failure is rethrown
+// so the view reports it instead of rendering an empty panel.
```
chrisguiney, line 3105: Crediting the split here, because the description undersells it as one behaviour when it's actually two deliberately different ones, and I checked that the right one is used in each place.

`fetchAllowingNotFound` and `fetchPageAllowingNotFound` resolve to `undefined` only for a 404 and rethrow everything else. The pre-existing `optionalFetch` / `optionalFetchPage` still swallow everything, and they're genuinely left alone — they're used on the smart-search probe paths (`:1564`, `:1618`), where a 500 on one speculative candidate shouldn't abort the whole search.

The part that matters is which views got which. Both new pool views (`:2461`, `:2486`) and the address detail view (`:2107`) use the 404-only helpers, so "this endpoint isn't served by your node" can't be confused with a 500 from a node that does serve it. That distinction is the entire value of the version-floor story in the description, and it would have quietly evaporated if these had reused `optionalFetch`.

Keeping the transaction-lookup helpers as-is for the `Promise.all` reason is also the right trade, and stating it in the description rather than leaving it as an inconsistency is what makes it reviewable.

examples/dingo-blockfrost-explorer/src/main.ts
```diff
@@ -2220,39 +2419,84 @@ async function renderPoolsLookup(): Promise<void> {
...
     renderOverview();
   }
 }
 
+// loadRetiringPools refreshes the cached retirement schedule and returns
+// undefined when the endpoint is not served.
+async function loadRetiringPools(): Promise<
+  { items: PoolRetiringResponse[]; total?: number } | undefined
+> {
+  const retiring = await fetchPageAllowingNotFound<PoolRetiringResponse>(
```
chrisguiney, line 2486: Noting this for the performance question, since a retirement schedule and a DRep list are both the kind of thing that invites a page-walk.

Neither does. `blockfrostFetchPage` issues one request and reads `X-Pagination-Count-Total` for the count, so `count=100&page=1` here and `count=25&page=1` for the DRep list (`:2315`) are each a single bounded round trip. Reporting "N visible | M total" instead of fetching every page is the right call against a mainnet-sized DRep set.

The caching into `state.retiringPools` also means `retiringEpochFor` resolves the pool-detail retirement row from memory rather than re-querying, so pool detail stays at two requests. No new fetch appears inside a loop or a `.map` over results anywhere in this diff — the new views are 3, 2, 2, 1, and 1 requests, parallelised where independent.

examples/README.md
```diff
@@ -32,6 +32,13 @@ By default Compose builds Dingo from this checkout as
 `dingo-examples-dingo:local`. Set `DINGO_IMAGE=ghcr.io/blinklabs-io/dingo:<tag>`
 if you explicitly want to use a published image.
 
+Building from the checkout is the recommended path, because the examples track
```
chrisguiney, line 35: Good addition, and the specificity is what makes it useful — naming which endpoints landed in 0.68.0 versus which need a newer build means a reader can tell whether their symptom is a version problem or a real one.

The honesty about unreleased endpoints is the right call rather than quietly requiring `main`. It pairs with the 404-tolerant fetches: the doc says a section will report itself unavailable on an older node, and the code actually does that via `fetchPageAllowingNotFound`, so the two agree instead of the README promising something the UI doesn't do.

One thing to keep an eye on: this paragraph and the per-app README now both carry version floors, so they can drift apart as endpoints get released. Not worth restructuring now, just the usual cost of documenting a moving target in two places.

chrisguiney (approved):
Approving. The claim this leads with — that the DRep view is now aligned field for field with `api/blockfrost/types.go` — holds exactly, and the route-coverage and 404-handling claims check out too. One factual error in a comment and one pre-existing bug found nearby, neither blocking.

## What I could not verify

**There is no Node.js/npm on this machine**, so I could not run `npm ci`, `npx tsc --noEmit`, or `npm run build`.

Worth flagging that CI can't cover it either: no workflow under `.github/workflows/` references `examples`, `node`, `npm`, or `tsc` — the six workflows are Go-only. All checks on this PR pass, but they exercise Go, not this file. Your local build is the sole verification of the TypeScript.

That made the compiler-free checks the useful ones, so I compared the TS types against the Go structs directly instead.

## Verification performed

| Check | Result |
|---|---|
| `DRepResponse` alignment claim | **exact, field-for-field** — inline on `:298` |
| `DRepListItemResponse`, `AccountResponse`, delegation-history epoch nullability | all align; no second drift instance |
| "calls every GET route except the metadata root and `POST /tx/submit`" | all **39** GET routes referenced |
| 404-handling claim | accurate, and the right helper is used per view — inline on `:3105` |
| Helm 0.64.0 → 0.68.0, `examples/README.md` floor | as described |
| Request counts per new view | fixed and parallel; no N+1 |
| `npm ci` / `tsc --noEmit` / `npm run build` | **not run** — no Node here, none in CI |
| CI on this PR | all pass (Go-side only) |

I checked the neighbours of `DRepResponse` for the same class of drift and found none: `DRepListItemResponse` matches 8/8 including `metadata: DRepMetadataResponse | null` against `*DRepMetadataResponse`; `AccountResponse.ActiveEpoch *int64` against `number | null`; and `AccountDelegationHistoryResponse.ActiveEpoch int32` against `number`, correctly *non*-nullable.

On route coverage, the only apparent gap was `/blocks/latest/txs`, and it is reached — `fetchBlockTransactionHashes` puts the literal `"latest"` first in its candidate list.

## Pre-existing bug found next door

While confirming that route, I found that the same function requests two routes that don't exist. For any block that isn't the latest, its candidates are `[block.hash, String(block.height)]`, producing `/api/v0/blocks/<hash>/txs` and `/api/v0/blocks/<height>/txs`. Neither is registered — `api/blockfrost/blockfrost.go` has `GET /api/v0/blocks/{hash_or_number}` and `GET /api/v0/blocks/latest/txs`, but no `{hash_or_number}/txs`. So selecting any non-latest block issues two guaranteed 404s and then throws.

I diffed the function against the base and it is byte-identical, so this PR didn't introduce it — I can't anchor a comment there for the same reason. Raising it because it's the one place the route-coverage claim rests entirely on the `"latest"` candidate, and because it sits directly against the 404 work this PR is doing, so it's cheap to fix while the context is loaded.

## Performance

No degradation. Every new view issues a fixed, small number of requests, parallelised where independent: address 3 via `Promise.all` over summary/utxos/txs, pool list 2, pool detail 2, DRep list 1, DRep detail 1. No fetch appears inside a loop or a `.map` over results, and `blockfrostFetchPage` is a single request that reads `X-Pagination-Count-Total` rather than walking pages.

## Documentation

README and `examples/README.md` updated with a specific version floor. `DATABASE.md` / `ARCHITECTURE.md` correctly assessed as not affected. No linked issue, which is fine for an example catch-up — I checked for a `Closes`/`Refs` to validate against and there isn't one.

PR #3019, opened 2026-07-29
docs(examples): sundae dingo network check and utxo views

wolf31o2:
Fixes the swap example's build, verifies the node's network at startup, and adds address-scoped UTxO views that are only correct with exact address matching.

## The app did not build

`npm ci && npm run build` failed with TS2345 at the `TxBuilderV3` constructor. A dependency bump moved the direct `@blaze-cardano/sdk` to 0.3.1, while `@sundaeswap/core` 2.12.0 and `@utxorpc/blaze-provider` 0.3.9 both depend on 0.2.48. npm installed two copies, so `Blaze` and `TxBuilder` became nominally distinct types with separate declarations of a private property.

Holding the direct dependency at 0.2.48 leaves exactly one copy of the SDK in the tree. That fixes the build with no cast and drops the bundle from 1581 KB to 1543 KB.

A cast at the constructor would also compile, but it is worse than the error it hides: it hands a 0.3.1 object to code built against 0.2.48 inside the swap path, where a version mismatch surfaces as a malformed transaction rather than a type failure. The README records why this must not be bumped on its own, and to move it when both consumers move.

A stale `node_modules` masked this locally. `npx tsc --noEmit` passed against it and failed after a clean `npm ci`.

## Network verification at startup

The header claimed Preview as static markup, and the only actual check was the wallet's CIP-30 network id. A `DINGO_UTXORPC_URL` pointed at preprod or mainnet silently produced unusable transactions, because the Sundae V3 script hashes are Preview-only.

Startup now calls ReadGenesis and refuses to continue unless the network magic is Preview, with the badge reflecting the real answer.

## Address-scoped UTxO views

Two rows in the wallet dropdown, fed by Dingo:

- an exact-address UTxO total, from a search with the exact address field
- a delegation-part total, spanning every address form sharing the address stake credential, resolved through a UTxO read and capped at 200 refs with an explicit first-N label

Both are only correct as of exact address matching. Before it, the raw address bytes the UTxO RPC SDK sends were rejected because the field was CBOR-decoded, and the payment and delegation part fields were decoded as whole addresses rather than 28-byte credentials.

## Not reachable, so not shown

Stake address info is a node-to-client LocalStateQuery handler in `ledger/queries.go` with no query service RPC. Delegated stake, delegated pool, rewards, and DRep delegation cannot be shown from this app. The credential UTxO view is the honest reachable answer and is labelled as not being stake address info. The same applies to the CBOR-wrapped stake snapshot query.

The configurable FIFO mempool backend and its double-buffered revalidation are node-side and change no client-visible RPC: submit still returns the hash and the wait still streams to confirmed. The TLS 1.2 minimum applies only when Dingo serves its own certificate, and this example speaks plain HTTP through the Vite dev proxy. Both are README notes only.

## Other

Hex helpers move out of the query provider into a shared module, replacing a private duplicate. No behavior change.

Helm values move from 0.64.0 to 0.68.0, which predates exact address matching, and the README states 0.68.0 as the minimum with the reason.

## Testing

- `npm ci`, `npx tsc --noEmit`, `npm run build` all clean, verified from a clean install in a fresh worktree
- Confirmed the installed tree resolves to a single `@blaze-cardano/sdk@0.2.48` with both consumers deduped against it

The runtime paths are not exercised here. That needs a synced Preview Dingo at 0.68.0 or newer and a CIP-30 wallet.

`dist/` is gitignored and stays generated. `DATABASE.md` and `ARCHITECTURE.md` checked and not affected.

---
## Summary by cubic
Fixes the swap example build by pinning `@blaze-cardano/sdk` to 0.2.48. Adds a Preview network check at startup and Dingo-backed UTxO totals in the wallet view.

- **New Features**
  - Verify the endpoint network at startup via UTxO RPC ReadGenesis; refuse non-Preview and show the real network in the badge.
  - Add wallet UTxO views from Dingo: exact-address and stake-credential totals; resolves up to 200 UTxOs and labels when truncated.

- **Dependencies**
  - Pin `@blaze-cardano/sdk` to 0.2.48 to match `@sundaeswap/core` 2.12.0 and `@utxorpc/blaze-provider` 0.3.9, fixing TS type mismatches and reducing bundle size.
  - Bump Dingo Helm values to 0.68.0 and document the `>=0.68.0` requirement for exact-address and delegation-part searches.

<sup>Written for commit e73f32be34cf5fb045fe24d51c6f9fea5204d4c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/dingo-sundae-preview/package.json
+++ b/examples/dingo-sundae-preview/package.json
@@ -9,7 +9,7 @@
     "check": "tsc --noEmit"
   },
   "dependencies": {
-    "@blaze-cardano/sdk": "0.3.1",
+    "@blaze-cardano/sdk": "0.2.48",
     "@sundaeswap/asset": "1.0.11",
     "@sundaeswap/core": "2.12.0",
     "@utxorpc/blaze-provider": "0.3.9"
--- a/examples/dingo-sundae-preview/src/dingo/bytes.ts
+++ b/examples/dingo-sundae-preview/src/dingo/bytes.ts
@@ -1,0 +1,21 @@
+import type { UtxoRpcBytes } from "./utxorpcTypes";
+
+export function bytesToHex(bytes: Uint8Array): string {
+  return [...bytes].map((byte) => byte.toString(16).padStart(2, "0")).join("");
+}
+
+export function hexToBytes(hex: string): UtxoRpcBytes {
+  if (hex.length % 2 !== 0) {
+    throw new Error("Hex string must have an even number of characters.");
+  }
+
+  const bytes = new Uint8Array(hex.length / 2) as UtxoRpcBytes;
+  for (let offset = 0; offset < hex.length; offset += 2) {
+    const byte = Number.parseInt(hex.slice(offset, offset + 2), 16);
+    if (Number.isNaN(byte)) {
+      throw new Error("Hex string contains invalid characters.");
+    }
+    bytes[offset / 2] = byte;
+  }
+  return bytes;
+}
--- a/examples/dingo-sundae-preview/src/dingo/provider.ts
+++ b/examples/dingo-sundae-preview/src/dingo/provider.ts
@@ -1,6 +1,15 @@
 import { Core } from "@blaze-cardano/sdk";
 import { U5C } from "@utxorpc/blaze-provider";
+import { bytesToHex, hexToBytes } from "./bytes";
 import type { DingoQueryClient, UtxoRpcBytes } from "./utxorpcTypes";
+
+// Preview network magic. Dingo reports this from the Shelley genesis it loaded,
+// so it is the authoritative answer for which chain the endpoint is serving.
+const PREVIEW_NETWORK_MAGIC = 2;
+
+// Cap on how many UTxOs a wallet summary resolves through ReadUtxos. Dingo
+// rejects more than 1000 keys per request; this stays well under that.
+const MAX_SUMMARY_UTXOS = 200;
 
 export function createDingoProvider(): U5C {
   const url = import.meta.env.VITE_UTXORPC_URL || window.location.origin;
@@ -19,6 +28,84 @@ export async function assertDingoReady(provider: U5C): Promise<string> {
   return `Protocol ${params.protocolVersion.major}.${params.protocolVersion.minor}, max tx ${params.maxTxSize} bytes`;
 }
 
+// assertDingoNetwork confirms the endpoint really serves Preview before the app
+// builds orders against Preview-only Sundae V3 script hashes. Without this a
+// proxy pointed at another network silently produces unusable transactions.
+export async function assertDingoNetwork(provider: U5C): Promise<string> {
+  const genesis = await queryClientFor(provider).readGenesis();
+  if (genesis.networkMagic !== PREVIEW_NETWORK_MAGIC) {
+    throw new Error(
+      `Dingo reports network magic ${genesis.networkMagic}; this example requires Preview (${PREVIEW_NETWORK_MAGIC}).`,
+    );
+  }
+  return `Preview genesis confirmed (network magic ${genesis.networkMagic}, ${genesis.networkId})`;
+}
+
+export type DingoUtxoSummary = {
+  lovelace: bigint;
+  utxoCount: number;
+  truncated: boolean;
+};
+
+// summarizeDingoAddress reports what Dingo holds for one complete address.
+// Dingo compares the full serialized address bytes, so a base address does not
+// pick up enterprise UTxOs that only share its payment credential.
+export async function summarizeDingoAddress(
+  provider: U5C,
+  address: Core.Address,
+): Promise<DingoUtxoSummary> {
+  return summarizeResolvedUtxos(await provider.getUnspentOutputs(address), false);
+}
+
+// summarizeDingoStakeCredential reports every UTxO whose address delegates to
+// the given stake credential, across all address forms that share it. This is
+// the closest view UTxO RPC offers of a stake address; live delegation, rewards,
+// and pool assignment are node-to-client LocalStateQuery only.
+export async function summarizeDingoStakeCredential(
+  provider: U5C,
+  stakeCredentialHash: string,
+): Promise<DingoUtxoSummary> {
+  const rpcUtxos = await queryClientFor(provider).searchUtxosByDelegationPart(
+    hexToBytes(stakeCredentialHash),
+  );
+  const truncated = rpcUtxos.length > MAX_SUMMARY_UTXOS;
+  const references = rpcUtxos
+    .slice(0, MAX_SUMMARY_UTXOS)
+    .map(
+      (utxo) =>
+        new Core.TransactionInput(
+          Core.TransactionId(bytesToHex(utxo.txoRef.hash)),
+          BigInt(utxo.txoRef.index),
+        ),
+    );
+  if (references.length === 0) {
+    return { lovelace: 0n, utxoCount: 0, truncated };
+  }
+
+  return summarizeResolvedUtxos(
+    await provider.resolveUnspentOutputs(references),
+    truncated,
+  );
+}
+
+// stakeCredentialHashFor returns the address delegation part, key hash or
+// script hash alike, and undefined for forms that carry none such as
+// enterprise, pointer, and Byron addresses.
+export function stakeCredentialHashFor(address: Core.Address): string | undefined {
+  return address.getProps().delegationPart?.hash;
+}
+
+function summarizeResolvedUtxos(
+  utxos: Core.TransactionUnspentOutput[],
+  truncated: boolean,
+): DingoUtxoSummary {
+  let lovelace = 0n;
+  for (const utxo of utxos) {
+    lovelace += utxo.output().amount().coin();
+  }
+  return { lovelace, utxoCount: utxos.length, truncated };
+}
+
 type QueryClientHost = {
   queryClient?: DingoQueryClient;
 };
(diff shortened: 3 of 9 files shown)


examples/dingo-sundae-preview/package.json
```diff
@@ -9,7 +9,7 @@
     "check": "tsc --noEmit"
   },
   "dependencies": {
-    "@blaze-cardano/sdk": "0.3.1",
+    "@blaze-cardano/sdk": "0.2.48",
```
chrisguiney, line 12: I verified this from the lockfile rather than taking the diagnosis on trust, and it holds exactly.

Counting installed copies of `@blaze-cardano/sdk` in `package-lock.json`:

```
BASE  3 copies:
   node_modules/@blaze-cardano/sdk                                  = 0.3.1
   node_modules/@sundaeswap/core/node_modules/@blaze-cardano/sdk     = 0.2.48
   node_modules/@utxorpc/blaze-provider/node_modules/@blaze-cardano/sdk = 0.2.48

HEAD  1 copy:
   node_modules/@blaze-cardano/sdk                                  = 0.2.48
```

That's the TS2345 mechanism made concrete: two nested installs alongside a different top-level one, so the private property gets separate declarations and `Blaze`/`TxBuilder` become nominally distinct. Total lockfile packages drop 326 to 306, which is consistent with the reported bundle reduction.

The pin also holds for a checkable reason rather than by luck: at HEAD both consumers declare `^0.2.31` and `^0.2.47`, so 0.2.48 satisfies each and npm dedupes to one hoisted copy. Which is exactly why the README's do-not-bump-alone warning is right — moving root to 0.3.x while those ranges stay on `^0.2.x` reintroduces the nesting immediately.

And the argument against the cast is the important part: a cast would hand a 0.3.1 object to code compiled against 0.2.48 *inside the transaction-building path*, where a mismatch surfaces as a malformed transaction rather than a compile error. Preferring a resolvable version constraint over silencing the checker is the right call, and worth having written down.

examples/dingo-sundae-preview/src/dingo/provider.ts
```diff
@@ -1,7 +1,16 @@
 import { Core } from "@blaze-cardano/sdk";
 import { U5C } from "@utxorpc/blaze-provider";
+import { bytesToHex, hexToBytes } from "./bytes";
 import type { DingoQueryClient, UtxoRpcBytes } from "./utxorpcTypes";
 
+// Preview network magic. Dingo reports this from the Shelley genesis it loaded,
+// so it is the authoritative answer for which chain the endpoint is serving.
+const PREVIEW_NETWORK_MAGIC = 2;
```
chrisguiney, line 8: Checked against the source of truth: `config/cardano/preview/shelley-genesis.json:38` is `"networkMagic": 2`. Correct, and distinct from every other network in the repo (mainnet 764824073, devnet 42, musashi 164), so there's no near-miss to worry about.

Worth noting why this guard earns its place rather than being belt-and-braces: the pre-existing check was the wallet's CIP-30 network id, which only distinguishes mainnet from *any* testnet. It cannot tell Preview from Preprod, and the Sundae V3 script hashes are Preview-only — so a `DINGO_UTXORPC_URL` pointed at Preprod passed the old check and produced transactions that could never validate. Asking the node what it actually loaded is the only answer that separates those two.

examples/dingo-sundae-preview/src/main.ts
```diff
@@ -349,14 +363,22 @@ function getWallets(): Array<[string, Cip30Wallet]> {
 
 async function initializeDingo(): Promise<void> {
   try {
+    const network = await assertDingoNetwork(state.provider);
```
chrisguiney, line 366: Confirming the guard actually gates the thing it needs to, since a check that only writes a badge would be decoration.

`assertDingoNetwork` is awaited **first** here, ahead of `assertDingoReady`, `validateProtocolReferences`, `hydratePoolsFromDingo`, and `loadPool`. A non-Preview endpoint throws before any of those run, the catch sets "Network: unverified" with the error class, and `state.pool` is never assigned — so the swap handler's `if (!state.pool)` guard short-circuits too. The order path is closed transitively rather than left open.

Setting the badge to the literal "Network: Preview" after the assert is fine, since by that point it can't be anything else, and the full genesis detail lands in the `title` tooltip and the log. The one thing that makes this implicit rather than explicit is that swap safety comes from an unset `state.pool` rather than from disabling the action. It works today; it would survive a refactor better if the failure path set an explicit flag.

examples/dingo-sundae-preview/src/dingo/provider.ts
```diff
@@ -1,7 +1,16 @@
...
+
+// Cap on how many UTxOs a wallet summary resolves through ReadUtxos. Dingo
+// rejects more than 1000 keys per request; this stays well under that.
+const MAX_SUMMARY_UTXOS = 200;
```
chrisguiney, line 12: The comment cites a real constant, which I checked: `DefaultMaxUtxoKeys = 1000` in `api/utxorpc/utxorpc.go:46`, enforced at `api/utxorpc/query.go:421` with `"too many UTxO keys: %d exceeds maximum of %d"`. So 200 genuinely is well under the server limit, and it's a configurable minimum rather than a hard one — staying at a fifth of the default leaves room for an operator who lowered it.

Good that the cap surfaces as `truncated` rather than silently returning a partial total. Two small things about how it's reported, on the two lines below.

examples/dingo-sundae-preview/src/dingo/provider.ts
```diff
@@ -19,6 +28,84 @@ export async function assertDingoReady(provider: U5C): Promise<string> {
...
+  const references = rpcUtxos
+    .slice(0, MAX_SUMMARY_UTXOS)
+    .map(
+      (utxo) =>
+        new Core.TransactionInput(
+          Core.TransactionId(bytesToHex(utxo.txoRef.hash)),
+          BigInt(utxo.txoRef.index),
+        ),
+    );
+  if (references.length === 0) {
```
chrisguiney, line 81: Two nits on the truncation reporting, neither a defect.

**This branch can only ever return `truncated: false`.** `truncated` is `rpcUtxos.length > MAX_SUMMARY_UTXOS`, so if it's true then `references.length` is exactly 200 and this early return is unreachable. Passing the variable through reads as though the case matters; passing `false` literally would document that it can't.

**`utxoCount` reports resolved outputs, not found ones.** On `:106` it's `utxos.length`, capped at 200, so a wallet with 500 matching UTxOs shows "200" beside the first-N label — while `rpcUtxos.length` already knows the real number a few lines up. Threading the found count through would let the label read "first 200 of 500", which is strictly more useful than "200, truncated" and costs one field.

Separately, and to give credit where it's due: `stakeCredentialHashFor` using `address.getProps().delegationPart?.hash` instead of slicing bytes out of the serialized address is the right instinct. A base address is 1 + 28 + 28 bytes and hand-slicing that is a classic off-by-one, plus you get correct `undefined` behaviour for enterprise, pointer, and Byron forms for free.

examples/dingo-sundae-preview/src/dingo/bytes.ts
```diff
@@ -0,0 +1,21 @@
+import type { UtxoRpcBytes } from "./utxorpcTypes";
+
+export function bytesToHex(bytes: Uint8Array): string {
```
chrisguiney, line 3: Checked this is a genuine de-duplication rather than a third copy: the two private helpers are **deleted** from `dingoQueryProvider.ts` in the same diff, and these are byte-identical to them apart from `Uint8Array<ArrayBuffer>` becoming the `UtxoRpcBytes` alias — which is defined as exactly that type, so no semantic change. Net effect is one implementation where there were two.

The one place a small duplication remains is the `(provider as unknown as QueryClientHost).queryClient` access: the new `queryClientFor` does it, and the pre-existing `installDingoAssetSearchCompatibility` does it inline. Their contracts genuinely differ — one throws, the other returns silently so the compatibility shim is optional — so two paths are defensible. Extracting the helper is still an improvement over a third inline cast, and worth mentioning only so the next person adding a query-client caller reaches for `queryClientFor`.

chrisguiney (approved):
Approving. The dependency diagnosis is exactly right and I was able to prove it from the lockfile without npm, the Preview magic and the key-cap both check out against dingo's own constants, and the hex-helper move is a real de-duplication. Two minor notes inline.

## What I could not verify

**No Node.js/npm on this machine**, so `npm ci`, `npx tsc --noEmit`, and `npm run build` are unverified by me — and CI can't cover them either: no workflow under `.github/workflows/` references `examples`, `node`, `npm`, or `tsc`. The bundle numbers (1581 KB to 1543 KB) are likewise unverified. Your clean-install run in a fresh worktree is the only verification of the build itself.

So I verified the parts that don't need a compiler, which turned out to be most of what matters here.

## Verification performed

| Check | Result |
|---|---|
| "exactly one copy of the SDK in the tree" | **proven from the lockfile** — inline on `package.json:12` |
| `PREVIEW_NETWORK_MAGIC = 2` | matches `config/cardano/preview/shelley-genesis.json:38` |
| "Dingo rejects more than 1000 keys per request" | matches `DefaultMaxUtxoKeys = 1000`, enforced at `query.go:421` |
| Network guard actually gates the order path | yes — inline on `main.ts:366` |
| Hex helpers de-duplicated, not just added | private copies deleted; new module byte-identical |
| Helm 0.64.0 → 0.68.0 | as described |
| `npm ci` / `tsc --noEmit` / `npm run build`, bundle size | **not run** — no Node here, none in CI |
| CI on this PR | all pass (Go-side only) |

## Performance

The new network work is bounded and small: one `readGenesis` at startup, one `getUnspentOutputs` for the exact-address row, and one delegation-part search capped at 200 refs feeding a single `resolveUnspentOutputs`. Nothing issues requests in a loop, and the cap sits well under the server's own 1000-key limit rather than relying on it.

## Documentation

README updated with the version floor and the reason, plus the do-not-bump note on the pin. `DATABASE.md` / `ARCHITECTURE.md` correctly assessed as not affected. No linked issue, which is fine for an example fix — I checked for a `Closes`/`Refs` and there isn't one.

PR #3020, opened 2026-07-30
feat(blockfrost): align pools/extended with OpenAPI 0.1.90

### LINKED ISSUE #2489: blockfrost: align pools/extended response with OpenAPI 0.1.90

wolf31o2:
Closes #2489

Aligns `GET /pools/extended` with the Blockfrost `pool_list_extended` schema: adds the three missing fields, removes two that are not in the schema, and sources pool metadata from persisted off-chain state in a single query per page.

## Response shape

Verified against `src/schemas/pools/pool_list_extended.yaml` and `src/paths/api/pools/extended.yaml` in the spec rather than from the issue text.

Required: `pool_id`, `hex`, `active_stake`, `live_stake`, `blocks_minted`, `live_saturation`, `declared_pledge`, `margin_cost`, `fixed_cost`, `metadata`. Only `metadata` is nullable; when non-null it requires `url`, `hash`, `ticker`, `name`, `description` and `homepage` (all nullable strings) plus an optional `error` object whose `code` and `message` are both required.

So this adds `blocks_minted`, `live_saturation` and `metadata`, which the issue lists as missing.

## Breaking change: relays and vrf_key are removed

`relays` is absent from `pool_list_extended`, as the issue notes.

`vrf_key` is also absent. It belongs to `pool.yaml`, the pool detail schema. The issue does not mention this, but the acceptance criterion is to emit the OpenAPI field set, so it is removed too. Confirmed with the maintainer before doing it.

Anyone reading either field from this endpoint should use `GET /pools/{pool_id}`, which carries `vrf_key` and relay data. `PoolRelayInfo` and `PoolRelayResponse` are deleted; nothing else referenced them, and `TransactionPoolRelayInfo`/`TransactionPoolRelayResponse` are separate types still used by the transaction endpoints.

## live_saturation denominator

`live_saturation` divides live stake by the per-pool saturation threshold `totalCirculation / nOpt`, where `totalCirculation` is `MaxLovelaceSupply - Reserves`.

That quantity, not total active stake, is what defines the threshold: `ledger/rewards` passes `totalCirculation` (`rewards.go:287`) as sigma's denominator into `optimalPoolRewardChecked`, while passing `totalActiveStake` to `apparentPerformance` on the line immediately above. Using active stake instead overstates saturation by about 1.68x on mainnet-shaped inputs, moving the reported 1.0 crossing from roughly 72M ADA to roughly 43M.

Note that `Network()` computes two similar-looking values. The correct one is `maxSupply - reserves`; the `circulating` value on the following line additionally subtracts treasury and script-locked supply and is not what `rewards.go` uses.

`totalCirculation` is resolved once per page rather than per pool.

## Query cost

Two batched queries per page, neither of them per-pool.

Page metadata comes from one new `GetOffchainMetadataBatch(sourceType, urls, txn)` call. Shared logic lives in `database/plugin/metadata/offchain` with thin per-backend wrappers; sqlite chunks the `IN` list at its bind-variable limit, postgres and mysql issue one query. Plan:

```
SEARCH offchain_metadata USING INDEX idx_offchain_metadata_source_url_hash (source_type=? AND url=?)
```

No scan: it uses the leading two columns of the existing composite unique index, so no new index was needed. `TestNodeAdapterPoolsExtendedMetadataSingleBatchedQuery` asserts via a GORM callback counter that exactly one query hits `offchain_metadata` regardless of pool count, and bakes the plan check into the test so a regression to one-query-per-pool fails loudly.

`blocks_minted` reuses `CountPoolBlocksInSlotRange`, which already returns a map keyed by pool, so one call covers the whole page.

Metadata is never fetched over the network in the request path; only persisted state is read.

Because a URL is unique only together with its hash, the batch can return two rows sharing a URL under different hashes (metadata republished at the same URL). Callers match each row against their own `(url, hash)` anchor rather than assuming one row per URL.

## Unavailable inputs error rather than defaulting

`live_saturation` needs `nOpt` from `CurrentProtocolParams()`, which only succeeds once `LedgerState.Start()` has loaded protocol parameters. It is a required, non-nullable number with no schema-compatible way to express "unknown", so the request errors instead of emitting a `0.0` that a client cannot distinguish from a genuinely unsaturated pool. Pinned by `TestNodeAdapterPoolsExtendedProtocolParamsUnavailable`.

`metadata` is genuinely nullable in the schema, so `null` is correct when a pool has no registered anchor. That is a real absence rather than a stand-in for an unavailable value.

## Known limitation

`blocks_minted` undercounts on a Mithril-bootstrapped node: `pool_opcert_sequence` is written only by the live block-apply path, so pre-snapshot history is permanently absent rather than merely delayed. `DATABASE.md`'s existing note for pool detail is extended to cover `/pools/extended`, which reads the same table the same way.

## Tests

New coverage for response shape, pagination and order, metadata success, absence and failure, and database errors.

`TestHandlePoolsExtended` is a pre-existing test and was rewritten in place with maintainer sign-off, since it constructed and asserted the removed `VrfKey` and `Relays` fields. It now asserts `blocks_minted`, `live_saturation` and a fully populated `metadata` object, a second pool covers `metadata: null`, and a raw-JSON assertion confirms `vrf_key` and `relays` are absent from the wire format, which the typed struct can no longer express.

The adapter test derives its expected saturation threshold from the devnet genesis `MaxLovelaceSupply` and the actually-persisted `Reserves`, so it cannot pass by restating the implementation.

## Verification

`go build ./...` and `go build -tags dingo_extra_plugins ./...`, `go test ./api/blockfrost/... ./database/... ./ledger/...` with zero failures, `golangci-lint run` at 0 issues, `nilaway`, `modernize`, `make import-boundaries`, `go test -race ./api/blockfrost/...` with no data races, and `go test ./internal/test/conformance/...` all pass.

## Documentation

`DATABASE.md` gains a `GetOffchainMetadataBatch` section covering the query, index, chunking and the cross-hash-match behavior, plus the extended Mithril caveat. `ARCHITECTURE.md` documents the two batched per-page queries, the `live_saturation` protocol-parameters requirement, and that the metadata classification logic is intentionally duplicated rather than shared with `PoolMetadata`.

## One note for review

Adding a second caller of `offchainFetchError` triggered `unparam`, because every caller passes `"Pool"`. The parameter is kept with a documented exemption rather than hardcoded: `models.OffchainMetadataSource*` defines seven other sources (drep, drep_registration, drep_update, gov_proposal, gov_vote, constitution, committee_resign) that share this cache and this error classification and differ only by that label. Happy to hardcode it instead if you would rather not carry the exemption.

## Summary by CodeRabbit

* **New Features**
  * Updated `/pools/extended` to match the latest API schema.
  * Added blocks minted, live saturation, and structured off-chain pool metadata.
  * Added clear metadata error details when off-chain information is unavailable.

* **Changes**
  * Removed relay and VRF key fields from extended pool responses.
  * Improved metadata retrieval efficiency through batched lookups.
  * Requests now fail clearly when required protocol parameters are unavailable.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2471,16 +2471,23 @@ coarse candidates before forming a limited page, preserving continuation-token
 correctness. Blockfrost address-transaction reads apply the same CBOR-backed
 exact check over credential-index candidates and paginate the exact matches.
 
-`/pools/extended` computes `blocks_minted` and `live_saturation` for every
-active pool with one query each (`CountPoolBlocksInSlotRange`,
-`GetOffchainMetadataBatch`) rather than one query per pool; see the Off-chain
-Metadata Worker section above for the metadata half and DATABASE.md for both
-queries' index usage. `live_saturation` reuses pool detail's
-`poolSizeSaturation` and, like pool detail, requires the current protocol
-parameters' `nOpt` to be loaded (`LedgerState.Start()` having completed): a
-required, non-nullable schema field cannot fall back to a placeholder value,
-so the request fails outright rather than serving a fabricated 0.0 when
-protocol parameters are not yet available.
+`/pools/extended` resolves the whole page with two batched queries rather than
+one query per pool: `CountPoolBlocksInSlotRange` returns every active pool's
+`blocks_minted` keyed by pool, and `GetOffchainMetadataBatch` returns every
+pool's cached off-chain document keyed by URL, supplying the nullable
+`metadata` object. See the Off-chain Metadata Worker section above for the
+metadata half and DATABASE.md for both queries' index usage.
+
+`live_saturation` is not a query. It is computed in memory by pool detail's
+`poolSizeSaturation` helper from values already read for the page: the pool's
+live stake, the active-stake snapshot, `nOpt` from the current protocol
+parameters, and total circulation (`MaxLovelaceSupply - Reserves`, the
+denominator `ledger/rewards` uses for the saturation threshold, not total
+active stake). Like pool detail, it therefore requires protocol parameters to
+be loaded (`LedgerState.Start()` having completed): a required, non-nullable
+schema field cannot fall back to a placeholder value, so the request fails
+outright rather than serving a fabricated 0.0 when they are not yet
+available.
 
 `GET /assets/{asset}` derives its mint-history fields from the API-mode
 `asset_mint_burn` table, which the transaction indexer populates from
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1034,6 +1034,18 @@ rows): `PoolMetadata` re-runs `ValidatePoolMetadata` against `Content` before
 returning it, so a stale cached document that would fail today's validation
 does not keep serving as if it were valid, without writing back to the row.
 
+`/pools/extended` needs the same classification for every active pool on a
+page, not one pool at a time, so it does not call `PoolMetadata` per pool
+(that would be one query per row). Instead `NodeAdapter.PoolsExtended`
+resolves every active pool's registered metadata URL in a single
+`MetadataStore.GetOffchainMetadataBatch` call, then applies the same
+pending/failed/fetched classification (including the `offchainFetchError`
+mapping and the read-time `ValidatePoolMetadata` fallback) as pure,
+in-memory logic over the already-fetched rows. The classification itself is
+intentionally duplicated rather than shared by refactoring `PoolMetadata`,
+since the single-pool and batched-page code paths have different callers
+and error-propagation needs.
+
 ### Archive And History Expiry Topology
 
 Dingo's blob-store abstraction supports independent history expiry and archive
@@ -2458,6 +2470,17 @@ as one ANDed address pattern; exact keyset queries scan through nonmatching
 coarse candidates before forming a limited page, preserving continuation-token
 correctness. Blockfrost address-transaction reads apply the same CBOR-backed
 exact check over credential-index candidates and paginate the exact matches.
+
+`/pools/extended` computes `blocks_minted` and `live_saturation` for every
+active pool with one query each (`CountPoolBlocksInSlotRange`,
+`GetOffchainMetadataBatch`) rather than one query per pool; see the Off-chain
+Metadata Worker section above for the metadata half and DATABASE.md for both
+queries' index usage. `live_saturation` reuses pool detail's
+`poolSizeSaturation` and, like pool detail, requires the current protocol
+parameters' `nOpt` to be loaded (`LedgerState.Start()` having completed): a
+required, non-nullable schema field cannot fall back to a placeholder value,
+so the request fails outright rather than serving a fabricated 0.0 when
+protocol parameters are not yet available.
 
 `GET /assets/{asset}` derives its mint-history fields from the API-mode
 `asset_mint_burn` table, which the transaction indexer populates from
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -1422,7 +1422,9 @@ current epoch's slot range (`blocks_epoch`); no new storage or index was
 needed. The upper slot bound passed for "no bound" is `math.MaxInt64`, not
 `math.MaxUint64`: slot columns are bound as signed 64-bit integers through
 `database/sql`, and a `uint64` value with the high bit set is rejected by the
-sqlite driver.
+sqlite driver. `/pools/extended` reuses the same lifetime query, called once
+with every active pool's key hash so the whole page's `blocks_minted` values
+come from a single call rather than one per pool.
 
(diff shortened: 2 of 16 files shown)


wolf31o2:
All four items addressed in 14252a98. Both actionable findings were valid.

## 1. Deduplicate before splitting into sqlite chunks

Real bug, confirmed. `offchain.GetBatch` removes duplicates only within whatever slice it is handed, and the sqlite backend chunked the raw URL list at `sqliteBindVarLimit` before calling it, so a URL landing in two different chunks was queried twice and its row appended twice. Your point that per-chunk dedup cannot enforce the batch contract globally is exactly right.

This is reachable input rather than pathological: two pools can register the same metadata anchor, so a repeated URL on one page is expected.

Rather than duplicate the filter in the backend, the dedup is extracted as `offchain.DedupeURLs` and used by both `GetBatch` and the sqlite chunking path, so the two cannot drift. postgres and mysql pass the whole list to `GetBatch` in one query and were already correct.

Added `TestGetOffchainMetadataBatchDeduplicatesAcrossChunks`, which places the same URL at both ends of a list longer than one chunk so the repeats land in different chunks. I checked it actually catches the bug rather than passing either way: with the fix reverted it returns two identical rows instead of one and fails with the intended message.

Worth noting the response was not visibly wrong today, because the adapter keys its lookup map on `(url, hash)` and duplicate rows collapsed. The defect was doubled query work and a violated contract, which is why a test that pins the contract matters more than one that checks the rendered output.

## 2. Correct the query-to-field mapping in ARCHITECTURE.md

Correct, and it was my error. The paragraph paired `GetOffchainMetadataBatch` with `live_saturation`, which it does not supply.

Rewritten as two statements: the page is resolved with two batched queries, `CountPoolBlocksInSlotRange` for `blocks_minted` keyed by pool and `GetOffchainMetadataBatch` for the nullable `metadata` object keyed by URL; and `live_saturation` is not a query at all, but computed in memory by `poolSizeSaturation` from the live stake, active-stake snapshot, `nOpt`, and total circulation. I also named the circulation denominator explicitly there (`MaxLovelaceSupply - Reserves`, not total active stake), since that is the part most likely to be re-broken.

## 3. Brittle NotContains "SCAN" assertion

Valid. Narrowed, though not quite as first drafted: my initial attempt asserted `NotContains "SCAN offchain_metadata"`, which has the same flaw, since sqlite emits `SCAN offchain_metadata USING COVERING INDEX ...` for an index-only lookup. It now rejects only an unindexed scan: if a plan line mentions `SCAN offchain_metadata`, that line must also mention `INDEX`. The existing `Contains` check on the index name is unchanged.

## 4. Branch on StatusFetched explicitly

Done. The status is now a `switch` with an explicit `fetched` case falling through to validation, `failed` producing the classified error object, and anything else, including `nil` and any future status, returning the anchor-only shape. Previously an unrecognized status would have fallen into the read-time validation path and reported `DECODE_ERROR` for content that was simply never fetched.

## Verification

`go build` (both tag sets), `go test ./api/blockfrost/... ./database/...` with 0 `^FAIL` and 0 `^--- FAIL`, `golangci-lint run` 0 issues, `nilaway`, `modernize`, `make import-boundaries`, `go test -race ./api/blockfrost/...`, and conformance all pass.

arepala-uml (approved):

PR #3022, opened 2026-07-30
fix(database): bound HotCache CAS retries with jittered backoff and observability

### LINKED ISSUE #2286: database: bound HotCache CAS retry loops under contention
## Summary

`HotCache.Put()` uses an unbounded CAS retry loop.

## Audit Details

Finding ID: L5, LOW database/concurrency finding.

Affected file: `database/hot_cache.go`.

The audit found copy-on-write cache update logic that retries `CompareAndSwap` until it succeeds. Under normal contention it should converge quickly, but there is no explicit retry budget or fallback.

Impact: theoretical livelock or excessive CPU under extreme contention. Correctness should not depend on a cache write succeeding.

## What Needs to Happen

1. Add a retry budget or backoff strategy to `HotCache.Put()`.
2. Decide what happens when the budget is exceeded: drop the cache write, retry under a mutex, or emit a metric.
3. Add a stress/concurrency test for high write contention.
4. Keep the cache best-effort.

## Acceptance Criteria

- Cache insertion cannot spin forever under contention.
- Tests cover concurrent writes.

arepala-uml:
**Summary**

- Bounds all `HotCache` CAS retry loops with a fixed retry budget and jittered backoff. Updates are safely dropped when the budget is exhausted, resolving audit finding L5.
- Added statistics, Prometheus metrics, and optional logging for contention and dropped cache updates.

**Changes**

`database/hot_cache.go`

- Added bounded CAS retries and yield-then-jitter backoff.
- Applied the behavior to Put, access tracking, and eviction.
- Added contention statistics through CASStats().
- Added optional warnings for dropped updates.
- Added Prometheus registration for four CAS metrics.

`database/hot_cache_test.go`

- Added high-contention completion and progress coverage.
- Added retry-budget exhaustion and post-fallback usability coverage.
- Added optional-logging tests.
- Added Prometheus registration and value tests.

`database/cbor_cache.go`

- Connected logging and CAS metrics to the UTxO and transaction hot caches.

`database/database.go`
- Configured HotCache logging during database initialization.
- Registered HotCache metrics when Prometheus is enabled.

`ARCHITECTURE.md`

- Documented bounded retries, backoff, best-effort drops, and related metrics.

Closes #2286 

---
## Summary by cubic
Bound all `HotCache` CAS loops with a small retry budget and jittered backoff to stop unbounded spinning; eviction now happens in the same `Put` CAS attempt and byte‑limit accounting is kept inside the CAS snapshot to enforce max‑bytes race‑free. Adds counters, Prometheus metrics, and rate‑limited warnings for dropped best‑effort updates. Closes #2286.

- **Bug Fixes**
  - Guarantee limit enforcement by folding eviction into `Put` and tracking total bytes inside the CAS snapshot; fixes races that left the cache oversized or undercounted under contention.
  - Bound CAS retries (default 16) with yield‑then‑jittered backoff across `Put` and probabilistic access tracking; on budget exhaustion the update is dropped instead of spinning.

- **New Features**
  - `CASStats()` and Prometheus counters for CAS attempts, aborted writers, commits after backoff, and total backoff time (labeled `cache=utxo|tx`); metrics registration is idempotent.
  - Optional, rate‑limited warning logs for dropped updates via `TieredCborCache.SetLogger`; `database.New` installs a no‑op logger by default, wires it to both hot caches, and registers CAS metrics when `PromRegistry` is set; `ARCHITECTURE.md` updated and focused `evictToFit` tests added.

<sup>Written for commit cccf9d60ef5c2f1ffcf0d4c26d96f646a3390ca5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3022?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added monitoring for hot-cache contention, including retry attempts, aborted updates, and backoff activity.
  - Added optional warning logs when cache updates are abandoned after repeated contention.
  - Added APIs to configure logging and register cache metrics with Prometheus.

- **Performance & Reliability**
  - Bounded cache update retries to prevent prolonged blocking under heavy concurrent activity.
  - Improved responsiveness during high-contention workloads while keeping the cache operational.


database/database.go
```diff
@@ -191,6 +192,12 @@ func New(config *Config, stores Stores) (*Database, error) {
 				"error", err,
 			)
 		}
+		if err := db.cborCache.RegisterCASMetrics(configCopy.PromRegistry); err != nil {
+			configCopy.Logger.Warn(
+				"failed to register hot cache CAS metrics",
+				"error", err,
+			)
+		}
```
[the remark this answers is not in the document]
arepala-uml: Made the changes with proposed fix and please check it.

database/hot_cache_test.go
```diff
@@ -233,6 +237,174 @@ func TestHotCacheLFUEviction(t *testing.T) {
...
+		t,
+		stats.SuccessfulCommitsAfterBackoff,
+		"at least one contended writer should make progress after backing off, "+
+			"not just terminate",
+	)
+	assert.Positive(
+		t,
+		stats.SuccessfulCommitBackoffTime,
+		"successful contended commits should report their cumulative backoff",
+	)
```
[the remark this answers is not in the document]
arepala-uml: Fixed by replacing both assertions with stats.Attempts > numGoroutines*numOperations — this is threshold-independent: it just proves that at least one writer needed more than one CAS attempt to commit (i.e., contention actually happened), without relying on how deep into the backoff schedule that retry landed.

database/hot_cache.go
```diff
@@ -286,12 +458,18 @@ func (c *HotCache) maybeEvict() {
...
 			return
 		}
-		// CAS failed, retry with fresh data
+		// CAS failed. Eviction is opportunistic and re-checked on every
+		// Put, so under sustained contention it's safe to abandon this
+		// attempt rather than spin: the next Put will retry it.
+		c.backoff(attempt)
 	}
+	c.writersAborted.Add(1)
+	c.logWriterAborted("evict")
```
wolf31o2: This bounded fallback can leave the cache permanently above maxSize/maxBytes. A Put may commit the over-limit snapshot, then eviction can lose all 16 CAS attempts to sampled access updates; if no later Put occurs, nothing retries eviction. I reproduced this with the default retry budget by filling a 10-entry cache, contending through Get access-count updates, and inserting one more key: across ten 50-round runs, 3–11 rounds per run finished oversized. Please make limit enforcement eventual/guaranteed—for example, include eviction in the insertion snapshot or prevent competing cache CAS updates while the evictor is active.

database/hot_cache.go
```diff
@@ -33,25 +39,173 @@ type hotCacheData struct {
...
+}
+
+// logWriterAborted logs the writer-aborted-after-budget event, if a logger
+// has been configured. op identifies which CAS loop gave up (put,
+// increment_access, or evict).
+func (c *HotCache) logWriterAborted(op string) {
+	if c.logger == nil {
+		return
+	}
+	c.logger.Warn(
```
wolf31o2, line 142: Please rate-limit or aggregate this warning. Database.New wires the normal node logger into both hot caches, so this is not optional in production: every dropped Put, access update, or eviction emits a WARN on the contention path. One local high-contention probe recorded 6,962 aborts from 20,000 puts, which would produce the same number of warnings and can replace CPU churn with log/I/O churn. The counter should remain authoritative, with logging sampled or rate-limited.

database/hot_cache.go
```diff
@@ -131,16 +286,27 @@ func (c *HotCache) Put(key []byte, cbor []byte) {
 		}
 
 		// Try to atomically update both maps together
+		c.casAttempts.Add(1)
 		if c.data.CompareAndSwap(oldData, newData) {
 			if c.maxBytes > 0 {
 				c.curBytes.Add(memDelta)
 			}
-			break
+			if backoffTime > 0 {
```
wolf31o2: These success-after-backoff counters are updated only by Put. incrementAccess and maybeEvict also back off, but on a later successful CAS they return without incrementing commitsAfterBackoff or commitBackoffNanos. That makes the exported stats/Prometheus metrics incomplete relative to their documentation, which describes copy-on-write updates generally. Please centralize the success accounting across all three loops or rename/document these as Put-only metrics.

wolf31o2 (changes requested):
Requesting changes because the bounded eviction fallback can leave HotCache above its configured entry or byte limit after contention stops. The current contention test does not exercise eviction (20 unique keys with a 50-entry limit). Please also rate-limit abort logging and make the successful-after-backoff metrics consistent across the CAS paths they claim to cover. GitHub CI and the targeted local race/static checks pass, but they do not cover the reproduced eviction interleaving.

database/hot_cache.go
arepala-uml: The added metrics were intended to cover all `HotCache` copy-on-write CAS updates, not only `Put()`. I centralized the success accounting in `recordSuccessfulCommit()` and now call it from both remaining CAS paths: `Put(`) and `incrementAccess()`. The separate `maybeEvict()` CAS loop was removed by folding eviction into the `Put()` snapshot, so eviction is covered by `Put()`’s accounting. This keeps the exported statistics and Prometheus metrics consistent across the paths they document.

database/hot_cache.go
arepala-uml: Addressed by rate-limiting abort warnings to at most one message per `HotCache` instance **per second**. Every aborted update still increments `writersAborted`, so the counter remains authoritative while duplicate warnings are suppressed. An atomic timestamp ensures concurrent aborts cannot emit a burst of warnings. I added a test that forces 1,000 aborts and verifies all 1,000 are counted while only one warning is logged.

database/hot_cache.go
arepala-uml: I have choson with the approach of including eviction in the same snapshot as `Put()`. We now trim the cache before CAS, so a successful insertion stays within `maxSize`. If all CAS attempts fail, we drop the complete insertion and keep the existing cache unchanged. I also added a contention test for this case. I am handling maxBytes separately because its byte counter is currently outside the CAS snapshot.

database/hot_cache.go
```diff
@@ -125,22 +320,38 @@ func (c *HotCache) Put(key []byte, cbor []byte) {
 		maps.Copy(newAccessCnt, oldData.accessCnt)
 		newAccessCnt[keyStr]++ // increment on put
 
+		// Trim to the configured limits as part of this same snapshot, using
+		// an estimate of post-insert bytes derived from the running counter;
+		// this mirrors how memDelta itself is computed relative to oldData
+		// and needs no per-Put O(n) resummation.
+		estimatedBytes := c.curBytes.Load() + memDelta
```
wolf31o2: `curBytes` is updated only after the snapshot CAS at lines 339-341, so it is not part of the state serialized by this CAS. Multiple concurrent Puts can successfully publish successive snapshots while each still observes the same stale/lower `curBytes`; none performs byte eviction, and the final cache can remain above `maxBytes` after contention stops. Please put the byte total in `hotCacheData` and update it atomically with the entries, or otherwise derive the estimate from the CAS-protected snapshot. Add a regression test with concurrent Puts and a small `maxBytes` limit.

wolf31o2 (changes requested):
Requesting changes: the bounded CAS work still does not make the configured byte limit safe under concurrent Put calls. Please fix the accounting/eviction interleaving and add a concurrent maxBytes regression test.

database/hot_cache.go
arepala-uml: Fixed it by moving `totalBytes` into the same CAS-protected snapshot as the cache entries. This prevents concurrent writes from seeing fresh entries with a stale byte count and allows `maxBytes` to be enforced correctly. Added tests to verify byte accounting remains accurate after eviction and the cache stays within `maxBytes` under heavy concurrent writes.

arepala-uml:
Ran the hotCache race tests 20 consecutive times to validate concurrent reads, writes, retry handling, and eviction behavior. All runs passed without race conditions or intermittent failures.

```
akrepala@akrepalas-MacBook-Pro dingo % go test -race -count=20 ./database -run 'TestHotCache'
ok      github.com/blinklabs-io/dingo/database  76.383s
```

wolf31o2 (approved):

PR #3024, opened 2026-07-30
feat(blockfrost): add pools list endpoint

### LINKED ISSUE #3011: blockfrost: implement GET /pools
The Blockfrost-compatible API registers /pools/extended, /pools/retiring, and /pools/{pool_id}/metadata, but the plain paginated pool list route GET /api/v0/pools is not registered. A client following the Blockfrost OpenAPI spec gets the catch-all 404 handler instead of a pool list.

Source of truth: Blockfrost OpenAPI 0.1.90, schema pool_list.

Scope

Register GET /api/v0/pools returning a paginated array of bech32 pool IDs for pools with an active registration, matching the pool_list schema.

Support count, page, and order with deterministic blockchain ordering, consistent with the other Blockfrost list endpoints.

The active pool set is already available through GetActivePoolKeyHashes, which handlePoolsExtended uses, so this needs a handler plus an adapter method rather than new persisted queries.

Acceptance criteria

Route registered under /api/v0 with ordering that keeps /extended and /retiring resolving to their own handlers.

Response is a flat array of bech32 pool IDs per the pool_list schema.

count, page, and order behave consistently with /pools/extended, including invalid pagination handling.

Handler and adapter tests cover response shape, pagination, order, empty results, and database failures.

Assess DATABASE.md and ARCHITECTURE.md under the repository documentation requirements and update them if the implementation changes a documented query surface or cross-component flow.

Related

Pool detail, metadata, and retiring endpoints: #2936
Extended pool response alignment: #2489

wolf31o2:
Closes #3011

Registers `GET /api/v0/pools`. It was the only Blockfrost pools route missing: `/pools/extended`, `/pools/retiring`, `/pools/{pool_id}` and `/pools/{pool_id}/metadata` all resolve, so a spec-following client asking for the pool list got the catch-all 404.

## Response

`pool_list` is a flat JSON array of bech32 pool IDs, with `count`, `page`, and `order`. Verified against `src/schemas/pools/pool_list.yaml` and `src/paths/api/pools/index.yaml`.

The active set comes from the same non-retired pool definition `/pools/extended` uses, so the two endpoints agree on which pools exist.

## Ordering, and a decision the schema does not pin

The spec documents `order` as "the ordering of items from the point of view of the blockchain, not the page listing itself. By default, we return oldest first, newest last."

`GetActivePoolKeyHashesAtSlot` has no terminal `ORDER BY`, so its row order is incidental rather than contractual. Sorting that incidental order by pool ID would have satisfied a naive test while silently violating "oldest first", so this adds `GetActivePoolKeyHashesOrdered`, shared across all three backends, ranking each active pool by its registration certificate position: `added_slot`, then `block_index`, then `cert_index` ascending, with a `pool_key_hash` final tie-break for full determinism.

It ranks by the EARLIEST registration, not the latest. A pool that re-registers years later to change its margin keeps its original position rather than jumping to the newest slot, which is what "oldest first" means for pool age. The OpenAPI schema does not pin this, so it is a deliberate and reversible decision — stated as such in `DATABASE.md`, `ARCHITECTURE.md`, and both doc comments rather than left implicit.

`desc` is `slices.Reverse` of the same slice `asc` returns, so the two are exact reverses by construction rather than by two independently-sorted queries happening to agree. A test asserts that.

## Cross-backend verification

The ordering query uses a `ROW_NUMBER()` window function shared by sqlite, postgres, and mysql — exactly the kind of SQL where the three diverge — so compile-time parity is not evidence. Verified behaviorally against real PostgreSQL 16 and MySQL 8 containers with the same 8-pool fixture the sqlite test uses, covering a re-registration, a re-registration that cancels a pending retirement, a same-slot three-way tie broken by `block_index`/`cert_index`, a pending future retirement, and an already-effective retirement that must be excluded.

All three backends produce the identical oldest-first sequence. No assertion was relaxed to accommodate a backend.

## Query cost

Stated per backend rather than measured on one and asserted for all, since a sibling branch in this stack had to be corrected for exactly that.

At mainnet-comparable scale (~8,000 pools, ~10,000 historical registrations, ~500 retirements) all three backends fall back to a full scan of `pool_registration`: the available index leads with the pool, not `added_slot`, so the slot filter cannot become an index range. No backend is materially worse than the others. This is the same cost `/pools/extended` already pays through `GetActivePoolKeyHashesAtSlot` — the added `ROW_NUMBER()` is computed over the same already-joined rows rather than requiring a second pass. Plan excerpts are recorded in `DATABASE.md`.

Pagination is applied in memory over the ordered active set, matching the existing `/pools/retiring` split. The ranking is computed across the whole registration and retirement history, so the full active set must be ranked before any page boundary exists; a SQL `LIMIT` would trim rows after that work rather than avoid it.

## Route precedence

`TestPoolsListRouteDoesNotSwallowSiblings` builds the real mux via `handler()` and asserts `/pools`, `/pools/extended`, `/pools/retiring`, `/pools/{pool_id}` and `/pools/{pool_id}/metadata` each reach their own handler. Asserted rather than reasoned from `ServeMux` specificity rules.

## One pre-existing test changed, with maintainer sign-off

`TestRouterUnimplementedRouteReturns404` used `/api/v0/pools` as an example of a route falling through to the catch-all — the precise gap this PR closes — so that subtest became stale. The entry is removed with sign-off; `/api/v0/`, `/api/v0/scripts` and `/does-not-exist` still exercise the same path, so coverage is unchanged.

## Noted for follow-up, not fixed here

While running the postgres and mysql suites I found `TestGetRewardStakeInputsForPoolsUsesHistoricalExpiration{Postgres,Mysql}` in `pool_atslot_test.go` registers `t.Cleanup` after a plain `defer store.Close()`, so its cleanup queries run against an already-closed connection and silently no-op, leaving stale `epoch` rows behind in a shared integration database. That is pre-existing and untouched; the new tests here are made self-sufficient instead. Filing separately.

## Verification

`go build ./...` (default and `-tags dingo_extra_plugins`), `go test ./api/blockfrost/... ./database/...` with 0 `^FAIL` and 0 `^--- FAIL`, `golangci-lint run` 0 issues, `nilaway`, `modernize`, `make import-boundaries`, `go test -race ./api/blockfrost/...` with no data races, and conformance all pass. Full postgres and mysql metadata suites also pass against real containers under the plugin build tag; both containers were removed.

## Documentation

`DATABASE.md` gains a `GetActivePoolKeyHashesOrdered` section with the SQL, tie-break, ordering decision, and per-backend plans. `ARCHITECTURE.md` adds the route to the Blockfrost router inventory plus the ordering and pagination semantics.

---
## Summary by cubic
Adds the Blockfrost stake pool list endpoint that returns a paginated array of bech32 pool IDs with deterministic chain order. Completes #3011 and closes the 404 gap for spec-following clients calling /pools.

- **New Features**
  - Register GET `/api/v0/pools` (pool_list); returns active (registered, non-retired) pools using the same active set as `/pools/extended`.
  - Add `GetActivePoolKeyHashesOrdered` (shared via `database/plugin/metadata/internal/poolorder`); sorts oldest-first by earliest registration with a deterministic tie-break; `desc` reverses the same slice.
  - Apply in-memory pagination with standard headers; verified identical ordering across sqlite and real Postgres 16/MySQL 8.4; documented ordering, query cost, and minimum backend versions.

- **Refactors**
  - Use `Database.GetActivePoolKeyHashesOrdered` in the adapter instead of calling the plugin directly.
  - Fix Postgres/MySQL ordering tests: register epoch teardown outside the shared cleanup so it runs after the test, preventing premature fixture deletion and “no epoch data” failures while keeping the shared integration DB clean.

<sup>Written for commit f08e3c02cc648ac10e61098f53b75494ee0334cd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3024?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a new `/api/v0/pools` endpoint for listing active stake-pool IDs.
  * Supports pagination, ascending or descending registration order, total-result headers, and deterministic ordering.
  * Excludes pools that are effectively retired.
* **Bug Fixes**
  * Added validation for invalid pagination parameters and clear handling of empty results and backend errors.
* **Documentation**
  * Documented the new endpoint, ordering behavior, pagination, and active-pool selection rules.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2505,7 +2505,10 @@ the ORDER BY is itself a per-pool ranking computed over the whole
 registration/retirement history, so the entire active set must be ranked
 before any page boundary exists, and a SQL-side LIMIT would only trim rows
 after that ranking work, not reduce it. See DATABASE.md for the query's index
-usage.
+usage per backend (verified with real Postgres 16 and MySQL 8.4 containers at
+mainnet-comparable scale, not assumed from sqlite) and for the cross-backend
+behavioral tests confirming sqlite, postgres, and mysql return the identical
+oldest-first sequence against the same fixture.
 
 `GET /assets/{asset}` derives its mint-history fields from the API-mode
 `asset_mint_burn` table, which the transaction indexer populates from
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2403,8 +2403,8 @@ fields protect every API listener.
 
 A Blockfrost-compatible REST API that provides read access to chain data and
 transaction submission. The current router includes health/root, blocks,
-epochs/parameters, network/eras, genesis, assets, pools/extended, retiring
-pools, pool detail, pool metadata, governance DRep list and lookup, address
+epochs/parameters, network/eras, genesis, assets, pools list, pools/extended,
+retiring pools, pool detail, pool metadata, governance DRep list and lookup, address
 summary, address UTxOs and transactions, metadata label JSON/CBOR,
 transaction content/CBOR/metadata/UTxOs/certificates/redeemers/required
 signers, and account/delegation/registration/reward/UTxOs/withdrawals/
@@ -2488,6 +2488,24 @@ be loaded (`LedgerState.Start()` having completed): a required, non-nullable
 schema field cannot fall back to a placeholder value, so the request fails
 outright rather than serving a fabricated 0.0 when they are not yet
 available.
+
+`GET /pools` (pool_list) returns bare bech32 pool IDs for the same active-pool
+set `/pools/extended` reads, but ordered rather than left in incidental row
+order: `GetActivePoolKeyHashesOrdered` (shared across the sqlite/postgres/mysql
+backends via `poolorder`, mirroring the `poolcerthistory` pattern) sorts
+oldest-first by each pool's EARLIEST registration certificate
+(added_slot/block_index/cert_index ascending, pool_key_hash as a final
+deterministic tie-break), not its most recent parameter update -- a pool that
+re-registers to change its margin keeps its original list position rather than
+jumping to the end. `desc` is the in-memory reverse of the same ordered slice,
+so the two are exact reverses by construction. Pagination is applied in
+memory over the full active-pool result (matching `/pools/retiring`'s
+`GetRetiringPools`/`PoolsRetiring` split), not pushed into SQL LIMIT/OFFSET:
+the ORDER BY is itself a per-pool ranking computed over the whole
+registration/retirement history, so the entire active set must be ranked
+before any page boundary exists, and a SQL-side LIMIT would only trim rows
+after that ranking work, not reduce it. See DATABASE.md for the query's index
+usage.
 
 `GET /assets/{asset}` derives its mint-history fields from the API-mode
 `asset_mint_burn` table, which the transaction indexer populates from
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -1943,6 +1943,87 @@ slot. Such rows have no `certs`/`transaction` join, so without the
 tie-break to a registration, incorrectly keeping the pool active for stake
 snapshots and reward inputs.
 
+### `GetActivePoolKeyHashesOrdered`
+
+Backs the Blockfrost `GET /pools` (`pool_list`) endpoint: the same active-pool
+set as `GetActivePoolKeyHashes`/`GetActivePoolKeyHashesAtSlot` (registration
+with `added_slot <= tip`, no retirement or the retirement is cancelled by a
+later registration or targets a not-yet-started epoch), but with an explicit,
+deterministic order instead of `GetActivePoolKeyHashesAtSlot`'s incidental row
+order (no terminal `ORDER BY`, so its result order is not a documented
+contract). "Oldest first, newest last" (the OpenAPI schema's wording) is keyed
+on each pool's EARLIEST registration certificate, not its most recent one: a
+pool that later re-registers to update its margin or relays keeps its
+original list position rather than jumping to the end. The query, shared
+verbatim across the sqlite/postgres/mysql backends
+(`database/plugin/metadata/internal/poolorder`, mirroring the
+`poolcerthistory` sharing pattern above), computes both the latest-registration
+ranking (needed for the active/retired determination) and the
+earliest-registration ranking (needed for the sort key) as two `ROW_NUMBER()`
+columns over one CTE scan of `pool_registration` joined to `certs`/
+`transaction`, rather than scanning that join twice:
+
+```sql
+WITH reg_ranked AS (
+  SELECT pr.pool_id, pr.added_slot,
+    COALESCE(t.block_index, 0) AS blk_idx,
+    COALESCE(c.cert_index, 0) AS cert_idx,
+    ROW_NUMBER() OVER (
+      PARTITION BY pr.pool_id
+      ORDER BY pr.added_slot DESC, COALESCE(t.block_index, 0) DESC, COALESCE(c.cert_index, 0) DESC
+    ) AS rn_latest,
+    ROW_NUMBER() OVER (
+      PARTITION BY pr.pool_id
+      ORDER BY pr.added_slot ASC, COALESCE(t.block_index, 0) ASC, COALESCE(c.cert_index, 0) ASC
+    ) AS rn_first
+  FROM pool_registration pr
+  LEFT JOIN certs c ON c.id = pr.certificate_id
+  LEFT JOIN "transaction" t ON t.id = c.transaction_id
+  WHERE pr.added_slot <= $1
+),
+latest_ret AS ( -- same shape as GetActivePoolKeyHashesAtSlot's latest_ret
+  ...
+)
+SELECT p.pool_key_hash
+FROM pool p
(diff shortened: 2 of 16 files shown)


wolf31o2:
All three items addressed in 82e35a4e.

## 1. Use the database facade

Fixed. `PoolsList` now calls `db.GetActivePoolKeyHashesOrdered(txn)` instead of reaching through `db.Metadata()` to the plugin interface.

Worth noting this was worse than a style point: the facade in `database/pool.go` had zero callers, so this branch added it and then bypassed it. The change both narrows the API-to-database dependency and removes dead code that would have drifted.

## 2. Epoch fixture left behind

Fixed in both backends, and thank you for catching it. Both `cleanup` closures now delete the `epoch_id <= 5` range they create.

This one stings a little, because these tests carry a long comment explaining that `pool_atslot_test.go` leaks `epoch` rows through a `defer store.Close()` before `t.Cleanup` ordering bug, and work around it defensively. Then they leaked the same table themselves. Filed as #3025 for the pre-existing case; the comment now cross-references it so the reason not to repeat the pattern is visible at the site.

Your description of the mechanism is exactly right: `pool_atslot_test.go` uses `FirstOrCreate`, so a leftover `epoch_id = 0` with `length_in_slots = 2_000_000` is adopted rather than replaced, which silently makes an order-independent test order-dependent.

## 3. MySQL window-function minimum version

Documented, with one correction to the framing.

The requirement is not introduced by this PR. `ROW_NUMBER()` is already used on main by the mysql backend in `pool.go`, `poolreap.go`, `drep.go`, `governance.go` and `account.go`, and by the shared `internal/rewardstate` live-stake ranking. So MySQL 8+ has been a hard requirement for a while with nothing stating it.

Rather than note it as a property of this one query, `DATABASE.md`s SQL Conventions section now records minimum backend versions (MySQL 8.0, PostgreSQL 12, SQLite 3.25), says these features fail at query time rather than at startup or migration, lists the affected query sites, and notes that CI exercises PostgreSQL 16 and MySQL 8.4 so those are the versions actually covered by tests.

## On hoisting the shared fixture

Not done, deliberately. The two tests are near-duplicates as you say, but they live in separate packages (`mysql` and `postgres`), so sharing the fixture needs either a new exported test-helper package or a non-test package that exists only for tests. That is a larger change than the ordering contract warrants right now, and it would sit alongside the same duplication in the existing backend tests rather than fixing it. I would rather do it as a sweep across the backend test files than introduce a one-off helper used by two callers. Happy to file it if you want it tracked.

## Verification

`go build` (both tag sets), `go test ./api/blockfrost/... ./database/...` with 0 `^FAIL` and 0 `^--- FAIL`, `golangci-lint run` 0 issues, `nilaway`, `modernize`, `make import-boundaries`, `go test -race ./api/blockfrost/...`, and conformance all pass.

Note the postgres and mysql cleanup change is exercised by CI, which runs those services; locally those suites skip without the containers.

chrisguiney (dismissed):

wolf31o2:
CI caught a mistake in my previous commit. Fixed in f08e3c02, and this time verified against real containers rather than reasoned about.

## What broke

Adding the epoch delete to the `cleanup` closure broke both backend ordering tests:

```
--- FAIL: TestGetActivePoolKeyHashesOrderedMysql (1.17s)
    Received unexpected error: GetActivePoolKeyHashesOrdered: no epoch data available for requested slot
--- FAIL: TestGetActivePoolKeyHashesOrderedPostgres (4.96s)
```

`cleanup()` is invoked twice: once up front to clear leftovers from an earlier run, and again through `t.Cleanup`. The epoch fixture is created before that up-front call, so putting the delete inside the closure removed the row the test depends on before it ever queried.

Entirely self-inflicted, and avoidable: I changed tests that only run against the postgres and mysql services, said in the previous comment that they were exercised by CI rather than locally, and pushed anyway instead of standing up containers.

## The fix

The teardown is now its own `t.Cleanup`, registered after the up-front `cleanup()` call, so the fixture is removed when the test finishes without being removed before it runs. The comment explains why it is deliberately not part of the shared closure, so the next person does not fold it back in.

## Verified against real containers this time

Postgres 16 and MySQL 8, ports chosen to avoid collisions, removed afterwards:

- both ordering tests pass (mysql 4.00s, postgres 3.48s)
- full packages pass: `mysql 50.001s`, `postgres 131.433s`
- inspecting the databases after the run, no `length_in_slots = 2000000` row remains, so the leak this change targets is genuinely gone

The rows that do remain are `epoch_id` 0 through 5 at `length_in_slots = 100`, which is `pool_atslot_test.go`s own fixture, not this one. That is the pre-existing leak in #3025, and the run also reproduced its cause directly in the logs:

```
level=ERROR msg="sql: database is closed" query="DELETE FROM `account` WHERE staking_key = ..." file=.../pool_atslot_test.go:33
```

So #3025 now has empirical confirmation as well as a code reading.

## Also in this round

`ledger` passed on this run, at no risk of the 600s timeout that failed #3020 three times. #3033 raises that ceiling and is still worth merging independently: `ledger` measures about 566s of the 600s default on real hardware, so unrelated pull requests remain a coin flip until it lands.

Issue #3032, opened 2026-07-30
ci: ledger package runs at 80% of the go test timeout and intermittently times out


wolf31o2:
The `github.com/blinklabs-io/dingo/ledger` package is close enough to the per-package `go test` timeout that a slightly slower runner fails the job, on PRs that do not touch the ledger at all.

Observed on the Linux go-test job for #3020:

```
panic: test timed out after 10m0s
FAIL	github.com/blinklabs-io/dingo/ledger	600.143s
```

Every other package in that same run passed, including macOS and Windows for the identical commit. #3020 changes 16 files, none of them under `ledger/`, and the only cross-package change is a new `MetadataStore` interface method that no ledger code path calls.

On the most recent successful main run (job 91011785649) the same package reports:

```
ok  	github.com/blinklabs-io/dingo/ledger	483.337s
```

So the package normally consumes about 483s of a 600s budget, leaving roughly two minutes of headroom. Any runner slower than usual, or any test that grows slightly, crosses the line. The failure then lands on whichever PR happens to be running, which makes it look like that PR broke something.

The goroutine dump at the timeout showed the test binary inside `ledger/account_expiry_renew_test.go`, but that is where the clock ran out rather than necessarily the slow test.

Why this matters beyond the flake

A red check on an unrelated PR costs a review cycle each time it happens, and it trains people to re-run rather than read failures. It also means a genuine ledger slowdown has no margin left to show up as a signal.

Suggested approaches, roughly in order of value

Find where the time actually goes: run the package with `-v` and sort by elapsed time, or use `go test -json` to rank tests. A small number of tests very likely dominate the 483s.

Reduce the dominant costs where possible. Per project guidance the usual culprits are fixture-heavy setup repeated per test and real epoch or snapshot work that could share a fixture.

Consider whether some of the slowest cases belong in a longer-running suite rather than the default package run.

Raising `-timeout` is the least useful fix on its own: it hides the trend instead of restoring margin. If it is raised as a stopgap, pair it with the profiling above so the real number is known.

Scope

Measure per-test durations in `ledger` and record the top contributors in the issue.

Reduce or relocate the dominant ones so the package has meaningful headroom again.

Confirm on the Linux runner, which is the slowest of the three platforms for this package.

Related

Surfaced while merging the Blockfrost work: #3020

wolf31o2:
Direct measurement, which makes this worse than the CI logs suggested.

[...]

PR #3033, opened 2026-07-30
ci: set an explicit go test timeout on the race-instrumented job

wolf31o2:
Adds `-timeout 20m` to the Linux go-test job. One line plus a comment explaining why.

## Why

That job runs `go test -tags dingo_extra_plugins -ldflags="-s -w" -race ./...`. Go defaults to a 10 minute timeout per package binary, and under `-race` the `ledger` package alone consumes roughly 485s of it. That leaves about two minutes of headroom, so the default is acting as the binding constraint rather than as a safety net: a runner 15 to 25 percent slower than average crosses the line.

When that happens the failure lands on whichever pull request happens to be running, and reads as if that change broke something.

## Evidence it is runner variance, not a regression

Observed twice consecutively on #3020:

```
panic: test timed out after 10m0s
FAIL	github.com/blinklabs-io/dingo/ledger	600.143s
```

Every other package in that run passed. #3020 changes 16 files, none under `ledger/`, and its only cross-package change is a new `MetadataStore` interface method that no ledger path calls.

The same content passes on the branch stacked directly on top of it:

```
ok  	github.com/blinklabs-io/dingo/ledger	480.954s   (feat/3011-pools-list)
ok  	github.com/blinklabs-io/dingo/ledger	491.393s   (feat/2934-account-activity-endpoints)
ok  	github.com/blinklabs-io/dingo/ledger	483.337s   (main)
```

macOS and Windows pass on the identical commit because `go-test-other` does not enable `-race`.

Comparing package durations between the failing run and a passing one shows the whole runner was slower, not one test:

| package | failing | passing | delta |
|---|---|---|---|
| ledger/governance | 123.7s | 100.3s | +23% |
| ledgerstate | 39.6s | 34.9s | +14% |
| ouroboros | 55.6s | 48.8s | +14% |
| midnight/indexer | 76.5s | 69.9s | +9% |
| ledger/snapshot | 215.0s | 218.4s | -2% |
| mithril | 80.9s | 89.1s | -9% |

At a 485s baseline, a 15 to 23 percent slower runner puts `ledger` between 558s and 597s, straddling the limit. That accounts for the timeouts without any change in behavior.

The goroutine dump reported `running tests: TestVerifyRegisteredVrfKey_AcceptsRetiredPoolRegistration (1s)`, i.e. one test alive for one second when the alarm fired. Nothing is deadlocked; the package simply accumulated the full budget.

## What this does and does not do

It does not make the suite faster, and it does not remove the underlying fragility. It stops ordinary runner variance from turning into a failed check on unrelated work.

Restoring real headroom is tracked in #3032, which asks for per-test durations under `-race` so the dominant contributors are known rather than guessed. Raising the timeout alone would hide that trend, which is why the issue stays open and the comment in the workflow points at it.

Only the race-instrumented job is changed. The macOS and Windows jobs keep the default, since without `-race` they finish in about 4 and 12 minutes respectively and have plenty of margin.

## Scope note

20m is chosen to sit roughly 2x above the current 485s worst case, so it absorbs variance without being so large that a genuine hang would sit in CI for a long time before failing.

---
## Summary by cubic
Set a 20m timeout for `go test -race` in the Linux CI job to prevent flaky timeouts when the `ledger` package approaches Go’s 10m per-package default. This stabilizes the race-instrumented job on slower runners; macOS and Windows jobs are unchanged.

<sup>Written for commit b70cb381ea7472ba6cce0d9ce831784e6029527f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3033?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Increased the timeout for Linux race-detection tests to 20 minutes.
  * Added documentation clarifying the timeout and expected test runtime.

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -98,7 +98,14 @@ jobs:
       - name: go-build-minimal
         run: go build ./cmd/dingo
       - name: go-test
-        run: go test -tags dingo_extra_plugins -ldflags="-s -w" -race ./...
+        # Explicit -timeout because this job is race-instrumented. Go's
+        # default is 10m per package binary, and under -race the ledger
+        # package alone uses roughly 485s of it, so a runner 15-25% slower
+        # than average pushes it past the limit and fails a job on whichever
+        # PR happens to be running. See dingo #3032 for restoring real
+        # headroom; this only stops the default from being the binding
+        # constraint in the meantime.
+        run: go test -tags dingo_extra_plugins -ldflags="-s -w" -race -timeout 20m ./...
 
   # macOS and Windows jobs without Postgres service
   go-test-other:


wolf31o2:
Stronger evidence for this change, and a correction to the numbers in the description.

I measured the `ledger` package locally under the same flags this job uses (`-tags dingo_extra_plugins -ldflags="-s -w" -race`), running both refs sequentially on one machine so they could not contend for CPU:

```
main            fbe71ce4   ok  github.com/blinklabs-io/dingo/ledger  566.573s
PR #3020 tip    14252a98   ok  github.com/blinklabs-io/dingo/ledger  565.146s
```

Two corrections.

The description says the package uses roughly 485s of the 600s default. That came from CI runner logs and understates the problem: on real hardware it is about 566s, so roughly 94 percent of the budget with about 5 percent margin rather than the two minutes I claimed. The default is not merely the binding constraint, it is barely clearing.

And main is 1.4s slower than the branch that failed three times. So the timeouts were not caused by the change under test, which is what I expected but had only argued from the dependency graph rather than measured.

Also worth flagging honestly: this PR own Linux job passed with `ledger` at 499.670s, comfortably under the old default. So its green check does not demonstrate the fix working, it just got a fast runner. The justification is the numbers above, not this run.

That also means merging this does not guarantee #3020 goes green. What it does guarantee is that #3020 next run reports a real duration instead of being truncated at 600s, which is the number needed to size the work in #3032.

20m still looks like the right ceiling: about 2x the measured 566s worst case, enough to absorb runner variance without letting a genuine hang sit in CI for a long time.

chrisguiney (approved):

Issue #1898, opened 2026-04-13
Crash recovery and state consistency


wolf31o2:
Dingo needs reliable crash recovery for mainnet SPO operation. Unexpected shutdowns shouldn't corrupt ledger state.

Phases:
1. Startup consistency checks -- block continuity, tip consistency between chain and ledger, UTxO integrity, orphaned data detection.
2. Write-ahead logging -- WAL entries for block ops, commit tracking, recovery replay on startup.
3. Checkpoint system -- periodic snapshots with merkle roots, truncate WAL after checkpoint.
4. Automatic recovery -- WAL replay, checkpoint recovery fallback, repair corrupted data.
5. Blockfetch rollback fix -- state machine to prevent stuck batches during rollback.
6. UTxO validation config -- fail/warn/ignore modes for missing consumed UTxOs (relates to #396).

Builds on existing chain/ledger tip reconciliation (#1715).

Roadmap: Q3 2026

wolf31o2:
Status: resolved.

Current crash-recovery and consistency coverage:

- Cross-store commit fence: blob commit, durable blob sync, metadata commit, commit timestamps, partial-commit detection.
- Startup recovery: chain/ledger tip reconciliation, rollback to the valid chain tip, orphaned-blob cleanup, blob-only timestamp mismatch detection.
- Recovery ordering: startup and live reinitialization recover before History Expiry starts.
- UTxO integrity: producer-block UTxO reconstruction/healing and configurable strict validation past the Mithril boundary.
- Lifecycle recovery: consistent snapshots, manifest/checksum/tip validation, restore, truncate, live quiesce/rebuild, cancellation, and optional cloud copies.
- Rollback safety: rollback-loop handling, off-primary-chain protection, and at-tip blob-recovery guards.

Relevant merged work: #1715, #1738, #2720, #2923, #3072, #3105, #3110, #3118, #3119.

The original WAL/checkpoint/Merkle roadmap is superseded. The current design uses database transactions, durable cross-store commit fencing, startup repair, and validated snapshots/truncation. Generic arbitrary-database-corruption repair remains outside this issue.

wolf31o2:
Resolved by the current crash-consistency and recovery stack documented above.

Issue #3029, opened 2026-07-30
DevNet: tolerate cold-start slot progress delay


wolf31o2:
The all-Dingo DevNet harness can fail TestHarnessWaitForSlot because no node reaches slot 10 within 22.5s during cold startup, even though the subsequent consensus check reaches slot 20 with all nodes agreeing. Investigate and make the readiness/progress check reliable.

PR #3051, opened 2026-07-31
feat(database): crash recovery and state consistency

wolf31o2:
Refs #1898.

Adds `database/recovery`: a cross-store intent journal, merkle-rooted checkpoints, startup consistency checks, and automatic repair. Both stores are already crash safe individually; what was missing was any record of which combined commit was in flight when a process died, or what it intended.

- Startup consistency checks: commit timestamps, metadata/blob tip agreement, chain vs ledger tip, block continuity, UTxO integrity, orphaned data. `consistencyCheckMode` selects off/fast/full.
- Intent journal: segmented, CRC-32C framed, `begin` before either store is touched and `commit` once both are through. `crashRecoverySyncJournal` makes intents durable at one extra fsync per combined commit.
- Checkpoints: merkle-rooted state summaries, atomically installed, three generations retained, journal segments they cover are truncated.
- Automatic recovery: replays the journal above the newest verified checkpoint, then trims blob blocks above the chain tip or rolls the ledger back onto the blob tip, and restamps both stores.
- `utxoValidationMode` (fail/warn/ignore) replaces the `strictUtxoValidation` boolean, which stays as the fallback.

Phase 5 (blockfetch rollback) is already on main via the `next.Rollback` check in `blockfetchServerSendBatch`.

Off unless `crashRecovery` is set; the manager is nil and every call site is a nil-safe no-op. Artifacts live under `<dataDir>/recovery/` and are safe to delete.

DATABASE.md and ARCHITECTURE.md updated. `make test`, `-race` on the touched packages, conformance, golangci-lint, nilaway, modernize, and import-boundaries all pass.

---
## Summary by cubic
Adds crash recovery and state consistency via `database/recovery`: intent journal, Merkle-rooted checkpoints, startup checks (with safe logging for unknown severities), and automatic repair. Also adds `utxoValidationMode` (fail/warn/ignore) with config/CLI wiring and node integration (Linear #1898), and fixes blob tombstone handling and key listing used by the checks.

- **New Features**
  - Intent journal: segmented, CRC-32C framed WAL with `begin`/`commit`; optional fsync via `crashRecoverySyncJournal`.
  - Checkpoints: Merkle-rooted snapshots; atomic install; keep 3; truncate covered WAL segments.
  - Startup checks: commit timestamps, metadata/blob tip agreement, chain vs ledger tip, block continuity, UTxO integrity, orphaned data; modes `off`/`fast`/`full`; logs unknown severities safely.
  - Automatic repair: replay intents above newest verified checkpoint; trim blob blocks above chain tip or roll ledger to blob tip; restamp both stores.
  - Integration: automatic intent on tip moves; explicit `Txn.SetRecoveryIntent` for rollbacks; `ledger.RollbackToPoint` exported.
  - Performance: read the Mithril trust boundary once per block only when `utxoValidationMode` requires it.

- **Migration**
  - Off by default; enable with `crashRecovery` (CLI: `--crash-recovery`). Optional journal sync, checkpoint interval, and recovery directory settings (CLI: `--crash-recovery-sync-journal`, `--crash-recovery-checkpoint-interval`, `--crash-recovery-dir`).
  - Replace `strictUtxoValidation` with `utxoValidationMode`: `fail`, `warn`, or `ignore` (CLI: `--utxo-validation-mode`). Empty falls back to `strictUtxoValidation` (true = fail, false = ignore).
  - Artifacts live in `<dataDir>/recovery/` and are safe to delete. When disabled, all calls are nil-safe no-ops.

<sup>Written for commit 9831caf667de1cf378435c87128f1a41d97a3dc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3051?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added crash recovery for interrupted database commits, including startup repair and periodic checkpoints.
  * Added configurable startup consistency checks with disabled, fast, and full modes.
  * Added UTxO validation modes: fail, warn, or ignore.
  * Added configuration and command-line options for recovery, journal synchronization, checkpoint intervals, and validation behavior.
* **Documentation**
  * Documented crash-recovery architecture, database recovery artifacts, configuration, and repair behavior.
* **Bug Fixes**
  * Improved handling of partial commits, missing historical UTxOs, corrupted recovery data, and divergent stores.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -915,7 +915,9 @@ runs after the chain manager and ledger are up, because both the diagnosis and
 the repairs need them: `nodeRecoverySource` wraps `Database.RecoveryStateSource`
 to add the chain manager's tip, and `nodeRecoveryRepairer` routes trims to
 `Database.TrimBlobAbove`, rollbacks to `LedgerState.RollbackToPoint`, and
-restamping to `Database.StampCommitTimestamp`. It runs after the existing
+restamping to `Database.ResetCommitFence`. That reset performs an empty
+combined commit, assigning a fresh timestamp to both stores rather than
+reusing either divergent timestamp. It runs after the existing
 `RecoverCommitTimestampConflict` path so it reports on whatever that left
 behind, then starts the checkpoint ticker, which `Database.Close` stops.
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -178,6 +178,7 @@ graph LR
     db_meta_impl["database/plugin/metadata/{mysql,postgres,sqlite}"]
     db_meta_util["database/plugin/metadata/{importutil,labelcodec}"]
     db_immutable["database/immutable"]
+    db_recovery["database/recovery"]
     cardano_cfg["config/cardano"]
     ev["event"]
     ledger["ledger"]
@@ -236,7 +237,7 @@ graph LR
 
     mempool --> chain & ev & plugin
 
-    db --> db_blob & db_meta & db_types & db_models
+    db --> db_blob & db_meta & db_types & db_models & db_recovery
     db_blob --> db_types
     db_blob_impl --> plugin & db_blob & db_types
     db_meta --> db_models & db_types
@@ -470,6 +471,7 @@ dingo/
 │   ├── block_lru_cache.go # Block-level LRU cache
 │   ├── immutable/       # ImmutableDB chunk reader
 │   ├── models/          # Database models
+│   ├── recovery/        # Crash recovery: intent journal, checkpoints, checks
 │   ├── types/           # Database types
 │   ├── sops/            # Storage operations
 │   └── plugin/          # Storage domain contracts and implementations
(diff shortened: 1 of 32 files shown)


wolf31o2:
Worked through all 10 findings. Eight applied, one was already fixed before the review landed, one applied with a different resolution than suggested.

Applied:
- `node.go` — history-expiry pruner now starts after recovery, not before. It mutates both stores on its own goroutine and would invalidate the trim boundary recovery just computed.
- `recovery.go` — an unresolved intent with agreeing commit fences is now `OutcomeClean`, no trim, no repair action. Only begin records are fsynced, so a lost commit marker is the normal shape and DATABASE.md already said it should do nothing. The fences now decide the repair direction on their own; a blob tip below the metadata tip with agreeing fences is reported by `tip_consistency` rather than triggering a rollback against the stores' own evidence.
- `wal.go` — `scanHighestSeq` seeds from segment filenames as well as record contents, so an emptied active segment cannot reset the counter and reissue sequences a checkpoint already covers. Test added.
- `node_recovery.go` — a nil primary chain returns a zero tip instead of an error, so it cannot abort startup. The no-known-tip guard still protects the blob store.
- `transaction.go` — the Mithril boundary is resolved once per call and only for the modes that consult it. My version had reintroduced a per-input sync-state read on the default path.
- `consistency.go` — block-continuity reports the true break count; only the printed examples are capped at five.
- `checkpoint.go` — `CreatedUnixMilli` is now a merkle leaf, so the root binds every field as documented.
- `ARCHITECTURE.md` — "in which case" was attached to the wrong branch.

Resolved differently: `StampCommitTimestamp` was correctly diagnosed as not writing its `ts` argument — the combined commit stamps both stores itself and overwrites it. Rather than re-document a parameter that can never take effect, I removed it: `Repairer.ResetCommitFence()` takes no value, and the action line no longer names a timestamp it didn't write.

Already fixed: the unbounded orphan trim. `TrimBlobAbove` batches at 500 per transaction with a rescan loop and a no-progress guard.

Re-ran the full suite, `-race` on `database/recovery`, conformance, golangci-lint, nilaway, modernize, and import-boundaries.

wolf31o2:
@coderabbitai full review

wolf31o2:
DevNet note, since this touches the cross-store commit path and can roll the ledger back at startup.

I ran the all-dingo suite on this branch and on a clean `origin/main` (4b1116a8) for comparison. Both fail, and the four core consensus scenarios fail identically on each:

| Test | this branch | main |
|---|---|---|
| TestBasicBlockForging | FAIL | FAIL |
| TestChainGrowthRate | FAIL | FAIL |
| TestDingoChainAdvances | FAIL | FAIL |
| TestEpochBoundaryConsensus | FAIL | FAIL |
| TestHarnessVerifyConsensus | PASS | FAIL |
| TestHarnessWaitForSlot | PASS | FAIL |
| TestRelayPropagation | PASS | FAIL |
| TestStakeAddressInfoQueries | PASS | FAIL |
| TestSustainedConsensus | FAIL | PASS |

Highest tip reached: this branch slot 501 / block 54, main slot 477 / block 25. txpump on this branch submitted 3462 with 0 errors; on main the run could not inspect the txpump log at all.

The four shared failures all have the same shape — the chain stops advancing while all four nodes stay in agreement (branch stalled at slot 197 / block 17, with dingo-1/2/3/relay all at that tip). `TestSustainedConsensus` failed on this branch for that same stall, not for a consensus divergence, so the one column where main looks better is the same root cause presenting differently between runs.

Read: the DevNet consensus failures are pre-existing and not introduced here, which lines up with the open devnet cold-start work in #3029. This branch is not worse on that environment and is better on several scenarios. I can't use DevNet to positively validate the change while the suite is red on main, so the evidence for this PR is the unit suite, `-race` on `database` and `database/recovery`, and conformance — all green — plus CI.

Crash recovery defaults to on, so it was active in the branch run: the devnet configs don't override `crashRecovery`.

wolf31o2:
Closing this since it's based on the old database plugins.

PR #2567, opened 2026-06-16
WIP:   ci: updated the workflow to use re-useble workflows

tripura-repalle:
## Summary by cubic
Switch issue-close workflows to reusable workflows in `blinklabs-io/actions` to standardize behavior and reduce maintenance. No behavior change.

- **Refactors**
  - `test-issue-on-close.yml`: now calls `blinklabs-io/actions/.github/workflows/reuseable-test-issue-on-close.yml@main` with `issue_number` and `issue_title`.
  - `update-issue-on-close.yml`: now calls `blinklabs-io/actions/.github/workflows/reuseable-set-project-closed-date.yml@main` with `project_url`, `closed_date_field`, and `closed_at`; passes `project_pat` from `secrets.ORG_PROJECT_PAT`.

<sup>Written for commit 6bc7afba1fa2f597d2fd5f1e6c421b5e404fe5cb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/2567?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #3053, opened 2026-07-31
ci: gate consensus-sensitive changes with DevNet and cardano-node conformance

### LINKED ISSUE #3001: ci: gate consensus-sensitive changes with DevNet and cardano-node conformance
## Problem

The normal pull-request workflow runs Go unit/integration tests with race detection, including the ledger vector suite, but it does not run `internal/test/devnet/run-tests.sh` or its `--conformance` mode. The Antithesis workflow is manual.

As a result, changes to consensus, forging, chain selection, mempool, transaction submission, and Ouroboros protocols can merge without exercising either the all-Dingo network or Dingo beside cardano-node. Repository contributor guidance already says these suites are mandatory for such changes, but CI does not enforce that contract.

## Scope

Add an automated GitHub Actions gate for consensus-sensitive changes:

- Run the default all-Dingo DevNet suite (three producers, one relay, and `txpump`).
- Run `internal/test/devnet/run-tests.sh --conformance` beside cardano-node.
- Trigger for changes to consensus, ledger epoch/nonce logic, forging and credentials, chain selection, mempool, tx submission, NtN/NtC protocols, block/header validation, DevNet infrastructure, and their shared dependencies.
- Keep a stable required check name. For unrelated changes it may complete through an explicit path-classification job rather than omitting the check entirely.
- Reuse or supersede the reusable-workflow work in PR #2567 rather than creating parallel workflow infrastructure.

## Acceptance criteria

- [ ] Pull requests touching a consensus-sensitive path run both default and `--conformance` DevNet suites before merge.
- [ ] Pushes to `main` and release candidates run both suites regardless of path filtering.
- [ ] A scheduled full run catches path-filter omissions and environmental drift.
- [ ] The workflow pins and reports the Dingo and cardano-node image/commit identities used.
- [ ] Failures upload node logs, generated genesis/configuration, test output, and container status as artifacts.
- [ ] Jobs have explicit timeouts and always tear down Docker resources.
- [ ] Tests use existing `internal/test/testutil` synchronization rules and do not introduce `time.Sleep` synchronization.
- [ ] The required check cannot pass when a selected suite did not start or produced zero tests.
- [ ] Branch-protection documentation names the required check.
- [ ] `README.md`, `AGENTS.md`, and `internal/test/devnet/README.md` agree on when each suite runs and how to reproduce it locally.

wolf31o2:
Runs both DevNet suites in CI: the default all-Dingo network (three producers, one relay, txpump) and `run-tests.sh --conformance` beside cardano-node.

What runs when:

- Pull request touching a consensus-sensitive path: both suites. Patterns are in `.github/devnet-paths.txt` and cover consensus, ledger epoch/nonce logic, forging and credentials, chain selection, mempool, tx submission, NtN/NtC protocols, block and header validation, DevNet infrastructure, and shared dependencies.
- Pull request touching nothing consensus-sensitive: no suites; the check completes through the path-classification job.
- Push to `main`, `v*` tags, nightly schedule at 04:30 UTC: both suites, no path filtering.
- `workflow_dispatch`: all, dingo, or conformance.

Structure: `devnet.yml` classifies paths and calls the reusable workflow `devnet-suite.yml` once per mode, so both modes share one implementation. PR #2567 is unrelated (it moves the two issue-close workflows to `blinklabs-io/actions`), so there was nothing to reuse or supersede there; the reusable-workflow pattern is adopted, kept in-repo so the gate definition is reviewed in the same pull request it gates.

`devnet-gate` is the stable check name. It fails when a selected suite did not succeed, and also when a selected suite reported zero tests or no test count at all, so a suite that never reached `go test` cannot report green. There is no `continue-on-error` and no `|| true` around a suite.

Not yet required, deliberately. Branch protection is untouched. The all-Dingo suite is intermittently red on unmodified `main`, not permanently: baseline runs went 12 pass / 0 fail once and 10 pass / 2 fail another time, with `TestChainGrowthRate` and `TestSustainedConsensus` recurring and `TestEpochBoundaryConsensus` occasional. The cause is the forging wedge tracked in #3029; a fix in flight produced a 12 pass / 0 fail run with the wedge signature dropping from roughly 100 occurrences per node to 16. Requiring the check before that lands would block unrelated pull requests. Afterwards it is one settings change, documented in `docs/BRANCH_PROTECTION.md`.

What the CI runs on this branch actually showed:

- The structure works end to end. Classification, both suite jobs, artifact upload, and the gate all behaved as designed, and the gate failed rather than passing when a suite failed.
- The all-Dingo job ran 18 minutes on a 4-CPU GitHub-hosted runner (14 to 17 minutes locally on a 128-core host) and failed `TestChainGrowthRate` and `TestSustainedConsensus` while `TestBasicBlockForging`, `TestDingoChainAdvances`, `TestRelayPropagation` and `TestEpochBoundaryConsensus` passed and txpump reported 3463 accepted submissions. The uploaded node logs carry the #3029 wedge signature, `does not match first pending header hash`, 64, 76 and 71 times on the three producers. That is the known wedge, not a workflow or capacity problem.
- The latest run on the rebased head (30712379613) is the current state of the branch: conformance passed in 16 minutes, all-Dingo failed in 18 minutes with the wedge signature 77, 72 and 71 times on the three producers, and `devnet-gate` failed with "dingo suite did not succeed (result: failure)". Every other check on the head commit is green.
- One earlier conformance job failed after 3 minutes with zero tests. Cause: the `txpump` image build could not reach `deb.debian.org` from that runner, so `apt-get install ca-certificates` in `internal/test/antithesis/Dockerfile.txpump` timed out and `docker compose build` aborted before any container started. It is a transient outbound-network failure on that VM, not the cardano-node pull, the conformance profile, the `devnet_conformance` build tag, or the configurator: two other runs of this same workflow passed the full conformance suite on `ubuntu-latest`, in 17 and 16 minutes, and the all-Dingo job on the same commit built the same image with no connection errors. The pre-existing exposure is that both profiles apt-install at image build time on every run; that is worth removing separately and is not papered over here. The zero-test guard did its job, failing `count tests` rather than letting an empty run look green.

Runner feasibility, measured rather than predicted: both suites complete inside the 90-minute job timeout on GitHub-hosted `ubuntu-latest` (4 CPUs, 15 GiB RAM), 17 to 19 minutes each including image builds, and a full conformance pass has already been observed there. `ubuntu-latest` is therefore the default. `DEVNET_RUNNER` moves both suites to a larger runner for headroom, which is still the right target before the check is made required; setting it also switches the suite jobs into a serialized concurrency group, which persistent runners need because the DevNet uses fixed container names, a fixed 172.20.0.0/24 subnet, fixed host ports, and one `devnet` Compose project. Every run records CPU count, memory, disk, the Dingo commit, and the resolved cardano-node image digest.

Artifacts, timeouts, teardown: failures upload per-node logs, generated genesis and configuration, test output, and container status. `run-tests.sh` gains a `DEVNET_ARTIFACT_DIR` collection step that runs before teardown removes the volumes. Jobs have explicit timeouts (90 minutes for a suite, 10 for classify and the gate) and a final always-run step tears down both profiles.

`run-tests.sh` also no longer exits non-zero after a passing run. The genesis stake keys are copied out of the `utxo-keys` volume by a container running as root, so the host `rm -rf` in the EXIT trap failed with "Permission denied" and, under `set -e`, aborted the trap and exited 1: every recent local run exited 1 for this reason regardless of the test result, and the temp directories leaked. The copy now chowns to the invoking user, removal failures are warnings, and the trap exits the status that triggered it. Verified in isolation with removal forced to fail: a passing run exits 0, a failing run exits 7, and the pre-change code exits 1 on the same passing input.

The configurator's genesis start offset is now `DEVNET_SYSTEM_START_DELAY`, default 30 seconds so local behaviour and wall clock are unchanged, validated as an integer of at least 10 seconds before it reaches arithmetic and jq, passed to both profiles' configurator services by `docker-compose.yml`, and documented in the override table. It stays applied after key generation because `genesis-cli.py`'s own 5 second delay is too short when generation takes 30+ seconds. This makes the budget adjustable per host; it is not a fix for any test failure. Measured on the 4-CPU runner, the all-Dingo nodes reached the slot clock with 28 of the 30 seconds to spare, and raising the delay is not free because `run-tests.sh` starts the scenarios as soon as the nodes are healthy and their deadlines are wall clock from that point.

Tests and docs: `internal/test/ci/` fails the normal test run when a path pattern stops matching real files, when a known consensus-sensitive file is not classified, when an unrelated file is, if the teardown exit-status contract regresses, or if the start-delay plumbing is dropped. No `time.Sleep`, no existing test or assertion changed. `README.md`, `AGENTS.md`, `CLAUDE.md`, and `internal/test/devnet/README.md` now agree on when each suite runs and how to reproduce it locally. `ARCHITECTURE.md` and `DATABASE.md` were checked and are not affected.

Rebased onto `main` at 47babd0c. #3057 is still open and touches the same DevNet override table; merged against that branch locally the only conflict is the `TEST_TIMEOUT` row, resolved by keeping its `20m` row plus the two new rows added here, after which its `make docs-parity` passes.

Fixes #3001


wolf31o2:
Closing this as we've closed the underlying ticket. While this is a good idea, it burns a lot of cycles just sitting and waiting.

PR #3054, opened 2026-07-31
refactor(database): replace GORM metadata stores

wolf31o2:
## Summary by cubic
Replaced GORM-based metadata with a shared SQL store built on `database/sql` and `sqlc`, unified SQL across backends, set SQLite as default, and tightened migration/cutover and shutdown. Updated docs to the new unversioned SQL store, narrowed supported storage backends in `README` to Badger/SQLite/GCS/S3, and CI now enforces cross‑dialect integration and `gorm` absence.

- **Refactors**
  - Added `database/plugin/metadata/sqlstore` owning pools, transactions, savepoints, readiness, and orchestration; a small dialect handles rebinding, quoting, param limits, read‑only isolation, bulk session tuning, and planner stats; generated queries stay internal. SQLite provider config includes a WAL writer, separate read pool, pragmas, migration locking, query/backup tracing, daily VACUUM, and avoids snapshot‑pool deadlock.
  - Deferred‑index manifest now carries explicit `table`/`name`; tests enforce no duplicates and resolvable entries.
  - Removed backend‑specific query packages and the MySQL metadata plugin; fully removed `gorm`; models are plain structs; `README`/`ARCHITECTURE.md`/`DATABASE.md` updated to shared store ownership and supported backends; Blockfrost/DB tests use `*sql.DB` and direct SQL.
  - Hardened adoption and shutdown: require a fresh metadata schema on cutover (with legacy schema validation before SQLite reference copy), preserve resumable migration state detection, make MySQL migration index replay idempotent, use portable SQLite file URIs, close asset‑batch cursors promptly, synchronize maintenance shutdown, close the admission gate atomically, add a guard when node settings are unexpectedly missing, and validate `utxorpc` v1beta query descriptors at startup.

- **Dependencies**
  - Removed `gorm.io/gorm`; added pinned `github.com/sqlc-dev/sqlc`; Makefile pins `SQLC_VERSION`.
  - CI runs `make sql-check` for `sqlc` generation drift, a portable `make gorm-check` to ensure `gorm` absence, and tagged cross‑dialect integration tests for `sqlstore` on Postgres/MySQL; removed redundant local database compose.

<sup>Written for commit 8f29b5724d826ae5352af540c9147d1ef0c67703. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3054?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


wolf31o2:
@cubic-dev-ai review

Makefile
```diff
@@ -89,6 +94,18 @@ proto: $(PROTOC) ## Generate Go code from protobuf definitions
...
 		$(ROOT_DIR)/midnight/proto/midnight_state.proto
 
+sql: ## Generate typed database/sql queries with pinned sqlc
+	$(SQLC) generate
+
+sql-check: sql ## Run sql, then fail when checked-in sqlc output is stale
+	git diff --exit-code -- database/plugin/metadata/sqlstore/internal/query
+
+gorm-check: ## Fail if the removed ORM returns to source or dependencies
+	@if rg -n 'gorm\.io|github.com/glebarez/sqlite|otelgorm' \
```
chrisguiney: **Blocking.** This guard can never fail, and I confirmed that from this PR's own CI run rather than by reading.

From the `go-test (Linux) (1.26.x)` log:

```
##[group]Run make gorm-check
make gorm-check
/bin/sh: 1: rg: not found
```

…and the step passed. The reason is the shell construct: `if rg …; then exit 1; fi` runs `rg`, gets exit 127 ("not found"), reads that as a false condition, skips the `exit 1`, and the recipe returns 0. A missing binary is indistinguishable from "no matches found."

ripgrep is not installed on GitHub's `ubuntu-latest` runners, so `gorm-dependency-absence` is green unconditionally. I reproduced the same silent pass locally.

Worth stressing: **the intent is currently satisfied** — I checked, and `gorm.io` appears zero times in `go.mod` and `go.sum`. So nothing is broken today. But the check that is supposed to keep it that way isn't actually watching, which is the failure mode you'd least want from a regression guard.

A couple of options, whichever you prefer:

```make
gorm-check: ## Fail if the removed ORM returns to source or dependencies
	@if grep -REn 'gorm\.io|github.com/glebarez/sqlite|otelgorm' \
		--include='*.go' --include='go.mod' --include='go.sum' .; then \
		exit 1; \
	fi
```

or keep `rg` and make its absence loud:

```make
	@command -v rg >/dev/null || { echo 'gorm-check requires ripgrep'; exit 1; }
```

`grep -RE` has the advantage of needing no tool that CI doesn't already have.

.github/workflows/go-test.yml
```diff
@@ -96,6 +96,10 @@ jobs:
         run: go build -tags dingo_extra_plugins ./cmd/dingo
       - name: go-build-minimal
         run: go build ./cmd/dingo
+      - name: sqlc-generation-drift
+        run: make sql-check
+      - name: gorm-dependency-absence
+        run: make gorm-check
```
chrisguiney, line 102: Flagging here as well as on the `Makefile`, since this is where the false assurance actually shows up.

This step currently passes on every run regardless of whether GORM is present, because `rg` isn't on the runner — see the detail and log excerpt on `Makefile`. Once the recipe is fixed this step is exactly right and worth keeping.

The neighbouring `sqlc-generation-drift` step, by contrast, does real work — I ran `make sql-check` and it correctly verified the checked-in `internal/query` output matches a fresh generation. Nice addition; that's the one that stops the generated layer from silently rotting.

database/plugin/metadata/sqlstore/dialect_integration_test.go
```diff
@@ -0,0 +1,321 @@
+//go:build dingo_db_integration
```
chrisguiney, line 1: This file is doing the most important validation in the PR, and almost nothing runs it.

After this change, PostgreSQL and MySQL no longer have hand-written query layers — they execute SQLite-authored SQL rewritten at runtime. This test is the only end-to-end check that the rewriting actually works against those engines. But it is behind `//go:build dingo_db_integration`, and that tag appears **nowhere** in `.github/workflows/` or the `Makefile`. There's no compose file, and the DSN defaults (`127.0.0.1:55432`, `127.0.0.1:53306`) point at services nothing in the repo provisions. So in practice it runs only if someone already knows it exists and hand-builds the environment.

Coverage is also narrower than it first looks — it exercises about **21 of 302** exported `Store` methods (~7%).

The good news, and I want to be clear about this: **I ran it, and it passes.** I brought up PostgreSQL 16 and MySQL 8 in containers on those exact ports:

```
--- PASS: TestPostgresSQLStoreIntegration (7.61s)
    INFO metadata schema migration complete version=1 name=v1alpha1
--- PASS: TestMySQLSQLStoreIntegration (31.54s)
    INFO metadata schema migration complete version=1 name=v1alpha1
```

So the DDL translation and the tested slice genuinely work on real engines. The gap is that none of that is reproducible in CI — a regression in the rewriting layer would land green.

A `compose.yaml` with these two services, a `make test-db-integration` target, and a CI job (even nightly rather than per-PR, given the runtime) would turn this from "verified once by whoever remembered" into an actual gate. Given three working backends were just replaced, that feels proportionate.

api/blockfrost/adapter_reward_history_test.go
```diff
@@ -1,570 +0,0 @@
-// Copyright 2026 Blink Labs Software
```
chrisguiney, line 1: Raising this gently, because the headline number is misleading and I want to be fair about it.

Repo-wide the PR goes from 4675 to 4190 test functions (−485), but the large majority of that is the *legitimate* collapse of triplicated per-backend suites — `mysql/`, `postgres/`, and `sqlite/` each carried their own copy, and consolidating to one shared store rightly consolidates those. Excluding the deleted backend packages the delta is only **4185 → 4148 (−37)**. That part is healthy.

Within that −37, though, a few DB-backed adapter tests were dropped rather than relocated, and this file is the one I'd most want back. It contained:

- `TestAccountRewardHistoryExcludesNonSpendableReward`
- `TestAccountRewardHistoryExcludesGuardedReward`

Those were the regression guards for the reward `spendable` filtering. At head I can't find an equivalent anywhere — the only surviving `RewardHistory` coverage is the `mockNode` in `blockfrost_test.go`, which returns a canned slice and so cannot exercise the SQL at all. There's no DB-level reward-history test in `database/` either; `shared_sqlstore_reward_parity_test.go` covers reward *state* rather than history exclusion.

That matters more than usual here precisely because the query layer underneath was just rewritten — this is exactly the kind of filtering that a SQL rewrite could silently drop, and it would now go unnoticed.

Same shape, lower stakes, for two others: `pools_list_test.go` (`TestHandlePoolsList*`, including the invalid-pagination and invalid-count cases) and `adapter_pool_metadata_validation_test.go` (the `RejectsEmptyContent` / `RejectsMissingRequiredField` / `RejectsInvalidFieldConstraint` set).

No need to restore them verbatim — porting the exclusion assertions down to the `sqlstore` level would arguably be a better home for them now.

database/plugin/metadata/sqlstore/dialect_queryer.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (q dialectQueryer) PrepareContext(
+	ctx context.Context,
+	query string,
+) (*sql.Stmt, error) {
+	return q.queryer.PrepareContext(ctx, q.translate(query))
+}
+
+func (q dialectQueryer) translate(query string) string {
```
chrisguiney, line 119: Crediting this, because it's the piece the whole cross-backend design rests on and I gave it a genuine try at breaking it.

My first probe extracted all 177 generated operational queries and prepared them against real MySQL with only `Rebind` applied. 53 failed — `ON CONFLICT … DO UPDATE`, `RETURNING id`, `"transaction"` as a double-quoted reserved identifier. That looked alarming until I found this function; the failures were **my probe bypassing `translate()`**, not a defect. Recording it here so the next reader doesn't repeat my mistake.

Re-run through the production path (`newDialectQueryer(db, dialect).PrepareContext`), against PostgreSQL 16 and MySQL 8:

```
PROBE[postgres]: 177 generated queries, 0 failed to prepare
PROBE[mysql]:    177 generated queries, 0 failed to prepare
```

So the rewriting genuinely covers the full generated surface on both engines. `translateMySQLUpsert`, `translateMySQLReturning`, and `translateMySQLReservedIdentifiers` between them handle every construct SQLite-authored SQL throws at MySQL here.

One observation rather than a request: doing this centrally — so generated sqlc output and hand-written inline SQL both get it — is the right call, and I verified nothing escapes it (`dbFromTxn` wraps every path; the bare `writeDB`/`readDB` references are only Ping/Stats/BeginTx/Close). The one thing that would make it more robust is that preparing every query on startup, or as a unit test against a schema-only fixture, would catch a future query that forgets one of these constructs. Right now that failure surfaces at runtime on whichever backend is unlucky.

Separately: `Dialect.SupportsReturning()` is declared but has no callers outside `dialect.go` — presumably vestigial now that `translateMySQLReturning` handles it, and safe to drop.

database/plugin/metadata/sqlite/backup.go
```diff
@@ -124,17 +121,42 @@ func (d *MetadataStoreSqlite) BackupTo(ctx context.Context, dstPath string) erro
 	return fsyncdir.Sync(dstDir)
 }
 
+// openSQLiteBackupDB opens a short-lived connection for VACUUM INTO. The
+// provider's write pool is deliberately single-connection; borrowing it can
+// deadlock a live snapshot when another metadata operation still owns that
+// connection. A dedicated connection preserves SQLite's WAL snapshot
+// semantics without contending with the pool's connection accounting.
+func openSQLiteBackupDB(ctx context.Context, databasePath string) (*sql.DB, error) {
```
chrisguiney, line 129: This one is a nice catch and the comment earns its place — worth saying so explicitly.

The reasoning is exactly right: the provider's write pool is deliberately capped at a single connection, so borrowing it for `VACUUM INTO` means a snapshot can block forever behind whatever operation currently owns that connection. Taking a dedicated short-lived connection sidesteps the pool's accounting entirely while still getting correct WAL snapshot semantics.

The pragma set on the backup DSN is consistent with the main pool (`journal_mode(WAL)`, `busy_timeout(30000)`, `foreign_keys(1)`), and `SetMaxOpenConns(1)`/`SetMaxIdleConns(1)` keeps the backup path from quietly becoming a second pool. The `PingContext` failure path closes the handle before returning, so a failed open doesn't leak an fd.

The `fsyncdir.Sync(dstDir)` above is the other half people usually miss — fsyncing the file doesn't persist its directory entry. Good to see both halves handled.

chrisguiney (changes requested):
Requesting changes, and I want to lead with this: the engineering here is strong and I found **no correctness defect**. The full suite passes under `-race`, the migration/adoption path is genuinely careful, and I verified the cross-backend SQL story empirically rather than taking it on trust. The items below are a CI guard that cannot fail, two red checks, and two coverage/verification asks — none of them are "this code is wrong."

## Verification performed

| Check | Result |
|---|---|
| `go build`, `go vet`, `golangci-lint` | clean, **0 issues** |
| Full suite `-race` — 28 pkgs (`database` 705s, `ledger` 889s, `api/*`, `midnight/*`, `ledgerstate`) | **all pass** |
| Root + `internal/...` `-race` — 17 pkgs | **all pass** |
| Conformance | pass |
| `make sql-check` (sqlc drift) | pass — generation reproducible |
| `make import-boundaries`, `make docs-parity` | pass |
| `modernize` | 13 findings, all `interface{}`→`any` in **generated** code (correctly left alone) |
| GORM in `go.mod`/`go.sum` | fully absent |

**Beyond the PR's claims**, I stood up real PostgreSQL 16 and MySQL 8 and ran the gated integration tests — both pass, migrations apply, and the regex DDL translation produces working schemas on both engines. I then wrote a throwaway probe that prepares **all 177 generated operational queries** through the production `dialectQueryer.translate()` path: **177/177 succeed on both Postgres and MySQL**. Details inline on `dialect_queryer.go`.

I also confirmed no SQL bypasses the dialect wrapper (`dbFromTxn` always wraps; raw pool handles are used only for Ping/Stats/BeginTx/Close), and that `CloseContext` holding `startMu` while waiting on `maintenanceDone` is **not** a deadlock, since the maintenance goroutine never takes that lock. The admission CAS (0→1→2) and the ctx-cancellable wait are sound.

## Blocking

1. **`gorm-check` is vacuously passing** — proven from this PR's own CI log. Inline on `Makefile`.
2. **CI is red**: `DCO` and `Conventional Commits` both fail. DCO matters given the signed-off-commit convention.

## Asks (not blocking correctness, but worth doing before this lands)

3. Postgres/MySQL end-to-end validation is run by nothing automated — inline on `dialect_integration_test.go`.
4. Some DB-backed API coverage was dropped rather than relocated — inline on the deleted `adapter_reward_history_test.go`.

## Reviewability

427 files, no linked issue, and no human-written description (the body is entirely cubic-generated). CodeRabbit skipped it outright — *"Review skipped: 426 files exceed the limit of 100."* For a change that replaces the entire metadata persistence layer, a few sentences on why now, what the cutover risk is, and what was verified on each backend would help a lot; splitting the mechanical deletions from the new `sqlstore` would help more. Entirely your call, but worth noting that automated review tooling could not process this as a single unit.

## What holds up well

- **Adoption is defensive in the right way.** `adoptExistingV1` validates the complete legacy column contract and refuses partial schemas with `ErrLegacySchema` *before* mutating anything, and defers the reference-input copy until after validation so an incomplete `utxo` table reports the right error instead of a misleading missing-column failure from the copy query.
- **Generated queries live in `internal/`, with `sql-check` guarding drift.** That is the right shape, and it verified clean here.
- **Net −55k lines** removing three parallel hand-written backends is a real simplification, and `DATABASE.md`/`ARCHITECTURE.md` were both updated to match.

## Minor

`CLAUDE.md:20` still says *"Update `DATABASE.md` for any change to **GORM models**"* — GORM is gone as of this PR and `CLAUDE.md` isn't in the diff. `DATABASE.md:118` likewise still refers to "every GORM model (~93 of them)". Both are one-line fixes.

Makefile
wolf31o2: Fixed in `a506060d`/`dae9c15b`. `gorm-check` now distinguishes matches (1), no matches (0), and scanner errors, and falls back to the runner's built-in `grep` when `rg` is unavailable. Verified with both the normal PATH and a PATH without `rg`; the GitHub failure exposed by this comment is resolved.

.github/workflows/go-test.yml
wolf31o2: Resolved alongside the guard fix. The workflow now keeps the real `gorm-check` and adds a `sqlstore-database-integration` step; `make sql-check` remains the generated-query drift gate.

database/plugin/metadata/sqlstore/dialect_integration_test.go
wolf31o2: Implemented in `a506060d`: added `database/plugin/metadata/sqlstore/compose.yaml`, `make test-db-integration`, and a PR CI step using the existing PostgreSQL/MySQL services. The Docker target passed the full gated integration suite locally against PostgreSQL 16 and MySQL 8.

api/blockfrost/adapter_reward_history_test.go
wolf31o2: Restored the two highest-value DB-backed API regressions in `api/blockfrost/adapter_reward_history_test.go`: non-spendable and guarded rewards are inserted through the shared SQL metadata store and asserted through `NodeAdapter.AccountRewardHistory`. The Blockfrost package and full tagged suite pass.

database/plugin/metadata/sqlstore/dialect_queryer.go
wolf31o2: Thanks for documenting the probe result. I reproduced the important part through the production path locally: all generated operational queries prepared successfully against PostgreSQL and MySQL. I left `SupportsReturning` unchanged because it remains part of the dialect capability contract even though translation currently handles the concrete SQL rewrite.

database/plugin/metadata/sqlite/backup.go
wolf31o2: Thanks — no further change was needed in the backup implementation. The dedicated connection, context-aware ping, tracing wrapper, bounded pool, pragmas, and directory fsync are covered by the existing lifecycle/backup tests; the full tagged suite passes.

database/plugin/metadata/sqlstore/store.go
```diff
@@ -0,0 +1,558 @@
...
+	conn := s.bulkConn
+	restoreErr := s.dialect.RestoreNormalMode(ctx, conn)
+	closeErr := conn.Close()
+	s.bulkConn = nil
+	return errors.Join(restoreErr, closeErr)
+}
+
+// UpdatePlannerStats refreshes backend planner statistics.
+func (s *Store) UpdatePlannerStats() error {
+	return s.dialect.UpdatePlannerStats(context.Background(), s.writeDB)
```
chrisguiney: **P1 — this hangs forever whenever bulk mode is active.**

`SetBulkLoadPragmas` checks out the *only* connection from the file-backed SQLite write pool (`SetMaxOpenConns(1)`, `sqlite/shared_sqlstore.go:133`) and parks it in `s.bulkConn`. This line then asks `s.writeDB` for a connection that can never be returned, and because it uses `context.Background()` there's no deadline to break the wait.

It's reachable deterministically, not just under a race: `mithril/sync.go:678` activates bulk mode for the rest of `Sync`, and `mithril/sync.go:1051` calls `node.RunPlannerStats` in API storage mode — so a Mithril sync in API mode wedges at `ANALYZE`.

Reproduced against the real store:

```go
store, _ := NewSQLStore(Config{DataDir: t.TempDir()}, metadata.ProviderDependencies{Logger: slog.Default()})
_ = store.Start(context.Background())
_ = store.SetBulkLoadPragmas()

done := make(chan error, 1)
go func() { done <- store.UpdatePlannerStats() }()
select {
case err := <-done:
    t.Logf("returned: %v", err)
case <-time.After(5 * time.Second):
    t.Error("DEADLOCK: blocked waiting for the write pool")   // <- this fires
}
```

You already solved exactly this hazard for `VACUUM INTO` in 0f77be41 by opening a short-lived dedicated `*sql.DB`. Either the same approach here, or routing through `bulkConn` when it's set (as `dbFromTxn` does), would fix it. A non-`Background` context would also turn a hang into a legible error.

database/plugin/metadata/sqlstore/store.go
```diff
@@ -0,0 +1,558 @@
...
+	}
+	return sqlTransaction.Commit()
+}
+
+func (s *Store) beginWriteTx(ctx context.Context) (*sql.Tx, error) {
+	s.bulkMu.RLock()
+	defer s.bulkMu.RUnlock()
+	conn := s.bulkConn
+	if conn != nil {
+		return conn.BeginTx(ctx, s.dialect.BeginOptions(false))
```
chrisguiney: **P1 — handing out the raw `*sql.Conn` loses the serialization the old code depended on.**

`database/sql` does **not** serialize concurrent `BeginTx` calls on a single `*sql.Conn` — `Conn.grabConn` has no in-transaction check, so the second caller issues a second `BEGIN` on the same session. I confirmed the contrast directly:

- `conn.BeginTx` twice on one `*sql.Conn` → `SQL logic error: cannot start a transaction within a transaction (1)`
- `db.BeginTx` twice on a pool with `SetMaxOpenConns(1)` → second call correctly **blocks** until the first finishes

That second behavior is what the pre-refactor code relied on: `postgres/database.go` pinned `SetMaxOpenConns(1)`/`SetMaxIdleConns(1)` on the *pool* during bulk load, so a second writer waited its turn. Going through a dedicated `Conn` converts that safe wait into a hard error.

Reproduced against the real store (SQLite, file-backed):

```
txn1 write err=<nil>
txn2 write err=SQL logic error: cannot start a transaction within a transaction (1)
```

Reachability: `mithril/sync.go:720-762` runs `importLedgerState` and `LoadBlobsWithDB` concurrently in an `errgroup` while bulk mode is active, and the loader goroutine writes metadata via `markImmutableUtxoOffsetsComplete` → `db.SetSyncState` (`internal/node/load.go:1140`). Worth flagging that the comment at `mithril/sync.go:675-677` — *"Enable bulk-load optimizations before launching parallel goroutines so both ... share the same pragma settings without racing"* — describes precisely the scenario this makes unsafe.

Two options that both preserve the intent: keep the old pool-pinning approach (session settings still apply, and the pool serializes for you), or guard `bulkConn` with a mutex held for the full lifetime of each transaction rather than just around `BeginTx`.

database/plugin/metadata/sqlstore/transaction_api.go
```diff
@@ -0,0 +1,488 @@
...
+		[]lcommon.TransactionInput,
+		0,
+		len(transaction.Inputs())+
+			len(transaction.Collateral())+
+			len(transaction.ReferenceInputs()),
+	)
+	allInputs = append(allInputs, transaction.Inputs()...)
+	allInputs = append(allInputs, transaction.Collateral()...)
+	allInputs = append(allInputs, transaction.ReferenceInputs()...)
+	for _, input := range allInputs {
```
chrisguiney, line 247: **P2 (non-blocking) — this is an N+1 that the previous implementation avoided.**

This loop issues one `QueryRowContext` per input, per collateral input, and per reference input, for every transaction. The pre-refactor path batch-fetched these into maps up front and then resolved them in memory — the old code even labelled it (`"failed to batch fetch input UTXO address keys (batched)"`, old `sqlite/transaction.go:~3470`), with `collectAddressTransactionsFromKeys` doing the in-memory join.

It's correctly gated to `types.StorageModeAPI` (`transaction_api.go:107`), so consensus-mode nodes are unaffected — but API mode is exactly where ingest throughput matters, and this is per-transaction on the block-apply path.

Restoring a batched lookup (one `WHERE (tx_id, output_idx) IN (...)` per transaction, chunked by the dialect's parameter limit, which `Dialect` already exposes) would recover the old behavior without giving up the shared-SQL structure. Happy to treat this as a follow-up if you'd rather keep this PR focused.

database/plugin/metadata/sqlstore/dialect_queryer.go
```diff
@@ -0,0 +1,231 @@
...
+	return query
+}
+
+// SQLite and PostgreSQL both accept double-quoted identifiers, while MySQL
+// treats double quotes as string delimiters unless ANSI_QUOTES is enabled.
+// Keep conversion here so every query path (including generated sqlc queries)
+// gets the same behavior without duplicating dialect branches in business
+// code. Shared SQL uses double quotes only for identifiers; values use the
+// standard single-quoted spelling.
+func translateMySQLReservedIdentifiers(query string) string {
```
chrisguiney, line 138: **Design note (non-blocking) — this layer is correct today but depends on invariants nothing enforces.**

The regex rewriting is a pragmatic way to keep one set of shared SQL, and I want to be clear it works: I parsed all 192 named statements under `queries/` and found **zero** violations of any of the assumptions below. The concern is purely about what happens the next time someone adds a query.

The four load-bearing assumptions:
1. Double quotes only ever delimit identifiers, never string literals — this function rewrites *any* double-quoted run to backticks, including one that happens to sit inside a single-quoted literal.
2. `ON CONFLICT ... DO UPDATE SET` never carries a trailing `WHERE` — `mysqlUpdatePattern` replaces only up to `SET`, so a conditional upsert would emit `ON DUPLICATE KEY UPDATE ... WHERE ...`, which MySQL rejects.
3. `RETURNING` is always exactly `RETURNING id` at end-of-statement — `returningIDPattern` won't match `RETURNING id, foo`, so the raw `RETURNING` would reach MySQL.
4. Upserted tables have at most one unique constraint — `ON CONFLICT (cols)` targets one constraint, whereas `ON DUPLICATE KEY UPDATE` fires on *any* of them.

None of these are violated now, and each fails loudly on MySQL rather than silently corrupting data (assumption 1 is the exception — it would rewrite quietly). A test asserting these properties over the parsed `queries/` tree would make the layer safe to extend without a MySQL expert in the loop.

Minor, separately: `VALUES()` in `ON DUPLICATE KEY UPDATE` (line ~210) is deprecated as of MySQL 8.0.20 in favor of aliased row references — worth a note for whenever the floor moves.

database/plugin/metadata/sqlstore/migrations/runner.go
```diff
@@ -0,0 +1,619 @@
...
+	hasUserTables, err := r.hasUserTables(ctx, conn)
+	if err != nil {
+		return err
+	}
+	stateTableExists, err := r.stateTableExists(ctx, conn)
+	if err != nil {
+		return err
+	}
+	if hasUserTables && !stateTableExists {
+		return fmt.Errorf(
```
chrisguiney, line 86: **Non-blocking — two small things on the resync UX.**

The clean-break decision reads as deliberate and it's documented well in `DATABASE.md`; no argument from me on the tradeoff, and the message is refreshingly actionable.

1. The message says *"delete the metadata database"*, but a node also has a Badger blob store. Deleting only the metadata DB leaves the blob store fully populated while metadata restarts at genesis. That won't error — `checkCommitTimestamp` returns early when the metadata timestamp is `<= 0` (`database/commit_timestamp.go:50-52`) — so the node starts and replays, but the operator is left with a stale blob store and no signal about it. Suggesting removal of the whole data directory would probably match what operators actually want.

2. Worth calling out prominently in the release notes rather than only surfacing at startup — for a mainnet operator this is a multi-hour/day resync, and it's the kind of thing that's much better known before the upgrade than after.

database/models/account.go
```diff
@@ -139,527 +131,12 @@ func (a *Account) TableName() string {
...
-	}
-	for _, columnType := range columnTypes {
-		if columnType.Name() != "tx_hash" {
-			continue
-		}
-		nullable, ok := columnType.Nullable()
-		return nullable, ok, nil
-	}
-	return false, false, nil
+func (AccountInactivityActivation) TableName() string {
```
chrisguiney: **P3 (non-blocking) — leftover GORM convention.**

There are 61 `TableName() string` methods still in `database/models/`. I checked and none of them have any caller, and no interface in the tree requires the method:

```
$ grep -rn "\.TableName()" --include='*.go' . | grep -v "func.*TableName" | wc -l
0
```

Since the PR's goal is a full GORM removal with plain-struct models, these are pure dead weight and dropping them would make the "no ORM" boundary self-evident. Mentioning it mostly because a stale reference like this tends to get cargo-culted into the next model someone adds.

chrisguiney (changes requested):
Thanks for taking this on — replacing the GORM layer with a shared `database/sql` + `sqlc` store is a big win, and the execution is impressive. Removing the per-backend query packages drops ~100k lines of duplication, the docs (`ARCHITECTURE.md`, `DATABASE.md`, `README.md`) are updated alongside the code, and CI now covers cross-dialect integration against real PG16/MySQL8 plus `sql-check`/`gorm-check`. I checked for duplicated code and found none introduced by this PR.

I ran the following locally against `df08066`:

| Check | Result |
|---|---|
| `go build ./...` | pass |
| `go test -race -count=1 ./...` | pass, exit 0, no races |
| `make import-boundaries` / `docs-parity` / `gorm-check` | pass |

I'm requesting changes on two concurrency defects in the bulk-load path that I was able to reproduce against the real store, plus the red required check. Everything else below is non-blocking.

### Blocking

1. **`UpdatePlannerStats` hangs indefinitely while bulk mode is active** — deterministic, reachable from `mithril.Sync` in API storage mode. Inline comment on `store.go`.
2. **Concurrent metadata write transactions fail while bulk mode is active** — a regression versus the old pool-pinning approach. Inline comment on `store.go`.
3. **Conventional Commits check is red.** Five non-merge commits lack the `type:` prefix: `df080663` ("refactor SQL metadata store"), `61611d42`, `c4c334e6`, `95c5053a`, `3bb841d0`. Not an issue if this lands as a squash using the PR title, but the gate is currently failing.

### Non-blocking
See inline comments on `transaction_api.go` (N+1 in API-mode ingest), `dialect_queryer.go` (unenforced translation invariants), `runner.go` (resync UX), and `models/account.go` (dead `TableName()` methods).

### Testing note
Per `CLAUDE.md`, this replaces the metadata store the ledger depends on (stake snapshots, pool state, nonces, pparams), so a DevNet run plus `./run-tests.sh --conformance` seems warranted before merge. Neither runs in CI today.

### Earlier bot findings that look already resolved
For whoever reviews next — I verified these cubic P1/P2s are fixed as of `df08066`, so they can be dismissed rather than re-investigated:
- `stake_denominator` **is** in the upsert set (`queries/sqlite/operational.sql:316`)
- The PostgreSQL migration lock **does** honor its timeout (`migrations/lock.go:84-87`)
- Address indexing **does** propagate non-`sql.ErrNoRows` errors (`transaction_api.go:251`)
- No `LastInsertId` remains in `pool.go`
- MySQL `DROP INDEX ... ON <table>` **is** dialect-specific (`dialect.go:91-96`)
- The `certDeposits` deregistration-refund concern is computed in the ledger and untouched here — pre-existing, out of scope for this PR

chrisguiney (approved):

PR #3057, opened 2026-08-01
docs: reconcile build contract, Go prerequisite, and DevNet topology

### LINKED ISSUE #3004: docs: reconcile build contract, Go prerequisite, and DevNet topology
## Problem

Contributor and operator documentation disagrees with the repository behavior in three concrete areas:

- `AGENTS.md` says the default `make` target formats, tests, and builds, while the Makefile default target runs only format and build.
- `go.mod` requires Go 1.26, while README and DevNet documentation say Go 1.25+.
- README describes the default DevNet as Dingo running beside cardano-node, while the default profile is three Dingo producers plus a Dingo relay; cardano-node is used only with `--conformance`.

These discrepancies make local validation and release-readiness evidence ambiguous.

## Scope

Reconcile the build/test contract and update every contributor-facing command and prerequisite:

- Decide whether `make` should include the race test suite or whether documentation should explicitly describe format/build only. Apply the decision consistently to Makefile help, README, AGENTS, and contributor instructions.
- Set the documented minimum Go version from `go.mod`/toolchain and remove stale hard-coded values.
- Describe the default all-Dingo DevNet and opt-in cardano-node conformance topology accurately, including the commands and test tags for each.
- Verify copy/paste command examples against the current targets and scripts.

## Acceptance criteria

- [ ] The default `make` behavior and every description of it agree.
- [ ] README, DevNet documentation, container/build documentation, and CI all state or derive the same minimum Go version as `go.mod`.
- [ ] README clearly identifies default DevNet as three Dingo producers plus one Dingo relay.
- [ ] README clearly identifies `--conformance` as the Dingo/cardano-node topology.
- [ ] DevNet service/port tables and example commands match the compose profiles and scripts.
- [ ] Makefile help output matches the actual dependency graph for build and test targets.
- [ ] A lightweight documentation/config parity check prevents the Go version and DevNet mode descriptions from drifting again.

### LINKED ISSUE #3032: ci: ledger package runs at 80% of the go test timeout and intermittently times out

wolf31o2:
Closes #3004 

The ledger package was spending most of its runtime rebuilding a database schema, not running ledger code.

## Where the time goes

Ranking by test is flat, which is what the earlier profiling in the issue found: 713 top-level tests, the slowest is 2 percent of the package, the top 30 are 18 percent. Ranking by what the tests spend time on is not flat at all. A CPU profile of one representative database-backed test puts 68 percent of it in `dbtest.NewDatabase`, and 60 percent in gorm's `AutoMigrate`.

Every in-memory sqlite store rebuilds 78 tables and 224 indexes from nothing. Roughly 340 of the 713 tests build one. Without race the effect is visible as a band of 277 tests that each cost 0.1 to 0.2s and together are 68 percent of the package.

Non-race duration histogram, before and after:

```
                before                  after
0.00-0.05s      n= 361  sum=  0.3s      n= 374  sum=  0.8s
0.05-0.10s      n=  29  sum=  2.2s      n= 245  sum= 18.4s
0.10-0.20s      n= 277  sum= 35.9s      n=  61  sum=  7.4s
0.20-0.50s      n=  29  sum=  8.9s      n=  27  sum=  7.0s
0.50s+          n=  17  sum= 19.8s      n=   6  sum=  4.6s
```

The 277-test band collapses into the band below it. That band was the shared fixture, not any individual test.

## The change

An in-memory sqlite database always starts empty, the model set is fixed at compile time, and every compatibility fixup in the migration chain returns early when its table is absent. The schema the chain produces therefore cannot vary between stores in one process. The first in-memory store now runs the real chain and records the DDL it ended up with, and later stores replay that DDL.

File-backed stores are untouched and always run the full chain, since they may hold an older schema or data a fixup has to repair. No production code path constructs a store with an empty data directory, so this only affects tests, though it would be correct if one did.

`TestInMemorySchemaTemplateMatchesFullMigration` compares a store built from the template against a store forced through the full chain, table by table, column by column, index by index, foreign key by foreign key. It compares through PRAGMA rather than raw DDL text because gorm emits a table's foreign key clauses in Go map order, so two stores in one process already spell the same table differently. If a model or migration ever makes the chain's output depend on something other than the compile-time model set, that test fails.

## Result

Measured back to back on one machine, same flags as the Linux CI job, nothing else running:

```
                 before     after
race             588.2s     464.4s    21.0% faster
no race           67.4s      38.3s    43.1% faster
```

Top contributors under race, before and after:

```
   before      after   test
   12.02s     11.66s   TestTryRecoverFromTxValidationErrorRecoversDependencyClosure
    6.97s      5.38s   TestStakeRewardPrecomputeSnapshotGuardOK
    6.92s      5.53s   TestDelegatorInactivityEndToEnd
    6.22s      5.30s   TestPrecomputedStakeRewardsRejectOutputsOutsideApplicationBoundary
    5.64s      3.98s   TestPrecomputeStakeRewardsWaitsForPreBabbagePrefilterSlot
    4.54s      3.45s   TestValidateTxDijkstraAcceptsAtOrAboveFloorPoolMarginThroughLedgerView
    4.33s      4.59s   TestCreateGenesisBlockFileBackedNoFKError
    3.80s      3.61s   TestUtxoByRefRecoversMissingBlobFromProducerBlock
    3.59s      2.81s   TestPrecomputedStakeRewardsCheckPreBabbageMissingLeaderPrefilter
    3.45s      2.68s   TestApplyStakeRewardsGuardsExpiredRewardAccount
```

`TestCreateGenesisBlockFileBackedNoFKError` is the one test that builds a file-backed store, so it is unaffected, as expected.

The same fixture is reached from 190 call sites in 16 packages, so the rest of the job gets the same reduction.

## On the Linux runner

The Linux job on this PR passed and reported `ok github.com/blinklabs-io/dingo/ledger 457.722s`. That is one sample, and per-package wall time on a four core runner is noisy: the last six successful main runs report 503.98, 518.79, 482.27, 412.74, 512.80 and 502.43, a mean of 488.8s with a 106s spread.

Runner speed moves every package in a run together, so the `chain` package works as a control. This PR's run had `chain` at 48.5s against a main mean of 45.0s, meaning it landed on a runner about 8 percent slower than average, which makes the raw comparison understate the change. Normalizing each package against `chain` in the same run:

```
package        main mean        PR      raw   normalized
ledger            488.8s    457.7s     6.4%      13.1%
ledger/governance 100.1s     87.4s    12.7%      18.9%
ledgerstate        34.8s     31.3s    10.1%      16.3%
ledger/snapshot   219.6s    228.4s    -4.0%       3.4%
api/blockfrost    111.6s    118.5s    -6.2%       1.5%
```

The packages that gain are the fixture-dominated ones. `ledger/snapshot` and `api/blockfrost` do more real work per database, so the fixture is a smaller share of them and they move less. With a single CI sample none of these figures are precise; the isolated local measurement above is the reliable one, and CI is consistent with it.

## What is left

Race is helped less than non-race because the replay is still real work: 302 CREATE statements per test database, and under race SQLite's parser alone is 30 percent of a database-backed test. Schema construction is now 59 percent of such a test, so eliminating it entirely would be worth roughly another 250s. Doing that means either fewer databases per test or building them ahead of time, and building them ahead of time would run `database.New` concurrently with tests, which trips the known global race in the plugin registry. Not attempted here.

## Coverage

No test was changed, removed, skipped or moved behind a build tag, and no assertion was weakened. `go test -list` over `ledger` is byte-identical before and after at 759 entries. The sqlite package gains the two new tests and loses none.

## Verification

```
go build -buildvcs=false ./...                                    clean
go test -tags dingo_extra_plugins -count=1 ./...                  pass
go test -race -timeout 20m -count=1 ./ledger/...                  ok  464.387s
go test -count=1 ./ledger/                                        ok   38.330s
go test -count=1 ./internal/test/conformance/...                  ok   13.140s
golangci-lint run ./...                                           0 issues
nilaway ./... , modernize ./...                                   byte-identical to base
make import-boundaries                                            ok
```

`DATABASE.md` documents the shortcut and the test that pins it. `ARCHITECTURE.md` was checked and is not affected: no plugin interface, lifecycle contract, component responsibility or cross-component flow changes.

Fixes #3032


wolf31o2:
Addressed the cubic review. All four points, with what changed:

- Divider rows without a trailing pipe are valid GFM and were silently skipping the whole table. `tableDividerRe` no longer requires the closing pipe. Covered by a new mutation: a table whose divider loses its trailing pipe and whose port is then wrong still fails.
- `dingoModeMarkers` no longer contains the bare word "default". It now needs an explicit profile name (`dingo mode`, `all-dingo`, `dingo profile`, `profiles=dingo`), so an unrelated mention of a default in a `--conformance` example cannot reclassify it.
- The port scan no longer reads every 3xxx figure out of a row's prose. A cell only counts as a documented port when the whole cell is the port, which is how both topology tables and the README table are written.
- The role check no longer asserts the absence of the other role. Instead it skips the service-name cell, so `dingo-relay` no longer satisfies a search for "relay" on its own and the positive assertion carries the check by itself. New mutation confirms it: rewording the relay row to "Dingo node, no block production" fails.

Mutation coverage is now 23 deliberate single-value regressions, each detected. `go build ./...`, `golangci-lint` (0 issues), `nilaway`, `modernize`, `make import-boundaries` and `go test -race ./internal/docsparity/` all pass.

wolf31o2:
@cubic-dev-ai review

wolf31o2:
Second review round addressed. The branch is now a single commit rebased onto current `main`, folding in the exact-builder-pin tightening that was pushed separately as `6cd37be0`.

cubic, first pass:

- Divider rows without a trailing pipe are valid GFM and were silently skipping the whole table. `tableDividerRe` no longer requires either outer pipe.
- `dingoModeMarkers` no longer contains the bare word "default". It now needs an explicit profile name, so an unrelated mention of a default inside a `--conformance` example cannot reclassify it.
- The port scan no longer reads every 3xxx figure out of a row's prose. A cell counts as a port only when the whole cell is the port.
- On the role check I took a different remedy than suggested. Restricting to a dedicated role column is not stable here, because the two topology tables put role in different positions and the README table has a third shape, so a column index would be brittle. Instead the negative "must not mention the other role" assertion is gone, which was the actual false-positive source, and the role text now excludes the service-name cell so `dingo-relay` no longer satisfies a search for "relay" on its own. The positive assertion carries the check by itself, confirmed by a mutation that rewords the relay row to "Dingo node, no block production".

cubic, second pass:

- `.PHONY` was read from the first declaration only, so a Makefile splitting it over several lines would report spurious failures. It now collects every declaration. A positive control confirms it: splitting the current `.PHONY` in half across two lines still passes, and did not before.
- The builder-image pin is now exact rather than at-least, and `examples/dingo-gov-lens/Dockerfile` moved up with it. Note the consequence: a dependabot bump of one Dockerfile alone will now fail this check until the others move together, which is the intended trade for a single source of truth.

CodeRabbit:

- Default target is no longer hard-coded to `all`. `parseMakefile` keeps declaration order and the default is the first ordinary rule, so adding a target above `all` cannot move the default out from under the docs. Mutation: inserting `quickbuild: test` ahead of `all` now fails.
- Matrix-backed `go-version` values are resolved rather than skipped. The premise as written was not quite right, the concrete `go-version: [1.26.x]` matrix declaration was already being validated, which is why flipping it to `1.25.x` already failed. The gap was a `go-version` reached through a matrix key of some other name. Expressions are now resolved against the matrix in the same file and an unresolvable one fails. Visible in the output: flipping the pin now reports lines 20, 87 and 133 rather than line 20 alone.
- Fence tracking now records the opening marker's character and length, so a longer fence can contain a shorter one and a `~~~` marker cannot close a backtick fence.
- Tables without outer pipes are now parsed. A divider only opens a table when the line above it has the same column count, which keeps unfenced prose containing dashes and pipes from being read as a table. Mutation: rewriting a port-override row without outer pipes and with the wrong port fails.

Mutation coverage is now 27 deliberate single-value regressions, each detected, plus the `.PHONY` positive control. `go build ./...`, `golangci-lint` (0 issues), `nilaway`, `modernize`, `make import-boundaries`, `make docs-parity` and `go test -race ./internal/docsparity/` all pass.

wolf31o2:
@cubic-dev-ai review

arepala-uml (approved):

Issue #2425, opened 2026-05-27
leios: wire BLS voting key registration and rotation into node state


wolf31o2:
## Summary

CIP-0164 PR cardano-foundation/CIPs#1196 makes stake-based committees the current Leios design and explicitly calls out BLS key rotation as a follow-up spec item. The adaptive-security argument for a public, deterministic committee depends on voting keys being rotated regularly.

Dingo needs a node-side tracking issue for loading, storing, activating, and using Leios BLS voting keys once gouroboros exposes the final ledger/wire types.

## What Needs to Happen

1. Track the CIP follow-up for BLS voting key registration and rotation.
2. Persist Leios voting keys or pool registration extensions in metadata once gouroboros exposes the types.
3. Activate rotated voting keys at the correct epoch boundary and preserve historical keys for certificate validation.
4. Load local block-producer Leios voting signing material from config or the eventual key-management path.
5. Use the correct `(pool, epoch)` verification key when validating votes and aggregate certificates.
6. Handle rollback across registrations or epoch-boundary activation points.
7. Add DevNet coverage once the reference node exposes the same mechanism.

## Dependencies

- gouroboros BLS voting key registration/rotation types.
- Final follow-up CIP for the on-chain rotation mechanism.
- Stake-truncated committee implementation for the voting path.

## References

- CIP PR: https://github.com/cardano-foundation/CIPs/pull/1196
- CIP-0164 Appendix A requirements: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0164/README.md#appendix-a-requirements
- Full Leios integration: #1946

PR #3060, opened 2026-08-01
feat(leios): load and use BLS signing keys

wolf31o2:
Load Cardano BLS text-envelope keys (while retaining raw-hex compatibility) and use configured keys for Leios vote signing and prototype verification, with derived-key fallback.

---
## Summary by cubic
Adds support for Cardano text‑envelope BLS12‑381 signing keys for Leios voting and now requires that any configured key matches a registered public key in `leiosVoterPublicKeys`. Verification prefers registered keys and falls back to the pool‑derived key only when no registration exists.

- **New Features**
  - Load Cardano text‑envelope signing keys (`cborHex`) from `leiosVoteSigningKeyFile`; legacy raw‑hex scalars still accepted with strict permission/size checks and envelope validation (BLS/Leios type; exact CBOR, no trailing bytes).
  - Use the configured key for vote emission; validate it against `leiosVoterPublicKeys` and refuse to enable voting if missing/mismatched; register the local public key and reject conflicts; otherwise derive from the pool cold‑key hash.
  - Validate prototype certificates and peer votes against registered keys first; use the pool‑derived key only when no registration exists; avoid deriving when a configured key is present.

<sup>Written for commit 92c48051c6204e9455efaf41fb7a53baccb8957c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3060?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Leios vote signing now supports Cardano text-envelope BLS key files, while retaining legacy raw hexadecimal key files.
  * Configured signing keys can be used directly; automatic key derivation remains available when no key file is provided.
  * Vote verification uses registered pool BLS keys when available.

* **Documentation**
  * Updated configuration and command-line guidance for supported key formats and file permissions.

* **Bug Fixes**
  * Improved handling and validation of unsupported or invalid signing-key files.


chadaibbhot-git:
Independent validation from a live Musashi SPO: the text-envelope loader in this PR interoperates cleanly with `cardano-cli`.

Generated a real key pair with the reference tooling (cardano-node prototype-2026w31 / cardano-cli 11.1.0.0):

```
cardano-cli dijkstra node key-gen-BLS \
  --verification-key-file bls.vkey --signing-key-file bls.skey
```

and fed the signing-key envelope through this PR's `LoadVoteSigningKeyFile`:

- the envelope `type` (`BlsSigningKey_bls12-381-BLS-Signature-Mininimal-Signature-Size`) and `cborHex` (`0x5820` + a 32-byte scalar) parse correctly, and
- the G2 public key gnark-crypto derives from the loaded scalar is **byte-identical** to the verification key `cardano-cli` (blst) produced for the same pair.

So Dingo and `cardano-cli` agree on BLS key handling — a key generated by `cardano-cli` yields the same public key in Dingo. 👍

If useful, here's a golden-vector regression test that pins the envelope format and asserts the gnark-derived pubkey equals the `cardano-cli`/blst verification key, so a future serialization/derivation drift gets caught. Drop it into `ledger/leios/keys_test.go` or a new file:

```go
// Golden vector generated with the reference tooling:
//
//	cardano-cli dijkstra node key-gen-BLS \
//	  --verification-key-file bls.vkey --signing-key-file bls.skey
//
// (cardano-node prototype-2026w31 / cardano-cli 11.1.0.0). Pins the
// text-envelope signing-key format and asserts the G2 public key gnark-crypto
// derives from the loaded scalar is byte-identical to the verification key
// cardano-cli (blst) produced for the same key pair.
const (
	goldenBLSSigningKeyEnvelope = `{
    "type": "BlsSigningKey_bls12-381-BLS-Signature-Mininimal-Signature-Size",
    "description": "BLS12-381 signing key",
    "cborHex": "58205f5cb4fc11f53a49c8b687a27ed99c80b0635876d1231fb6f60fe7da5cd13634"
}`
	// 0x5860 (CBOR byte string, len 96) + the 96-byte compressed G2 public key.
	goldenBLSVerificationKeyCborHex = "5860" +
		"aa75d7e0762ecac2576fe14666c382bde4e5bcd68edafb615e75e03e8d087b36" +
		"cb5be8bfa605a3dc983bc170cf9465250cd234533c1eba39db036aadd701f60c" +
		"2458e1a2f08c2cc3d71b7d60c0aa844c2a3a9bc2b6c56e6936322e3959ec4ca8"
)

func TestLoadVoteSigningKeyFileTextEnvelopeMatchesReferenceVKey(t *testing.T) {
	vkeyEnvelope, err := hex.DecodeString(goldenBLSVerificationKeyCborHex)
	require.NoError(t, err)
	require.Equal(t, byte(0x58), vkeyEnvelope[0])
	require.Equal(t, byte(0x60), vkeyEnvelope[1]) // 96-byte byte string
	wantPub := vkeyEnvelope[2:]
	require.Len(t, wantPub, 96)

	dir := t.TempDir()
	path := filepath.Join(dir, "bls.skey")
	require.NoError(t, os.WriteFile(path, []byte(goldenBLSSigningKeyEnvelope), 0o600))

	key, err := LoadVoteSigningKeyFile(path)
	require.NoError(t, err, "text-envelope key-gen-BLS file must load")

	require.Equal(t,
		hex.EncodeToString(wantPub),
		hex.EncodeToString(key.PublicKeyBytes()),
		"gnark-derived G2 public key must equal the cardano-cli/blst verification key",
	)
}
```

Verified passing against this branch. Happy to open it as a follow-up PR once this merges.

One operational note for the docs: since the static registry stays the verification trust root until on-chain Leios-key registration lands (#2425), an operator who configures a real key *before* their public key is network-visible would have peers verify their votes against the pool-derived key and reject them. Might be worth a one-line warning next to the flag in `dingo.yaml.example`.

cryptodj413:
## Review: NO-GO as-is

Reviewed via `gh pr diff`, tested in an isolated worktree (build, `go test -race ./ledger/leios/...`, lint, import-boundaries — all clean), and cross-checked the crypto logic against real `cardano-cli` BLS key output shared by an SPO in this thread.

**Solid parts** (independently verified): text-envelope vs raw-hex key parsing, CBOR exact-length/no-trailing-bytes validation, scalar range checks, file-permission checks, the new thread-safe `VoterRegistry`, and the registry-first/derived-key-fallback resolution in cert verification. `ARCHITECTURE.md` was updated appropriately for the new semantics.

**Blocking issue:** this PR drops the old startup check that verified a configured signing key matches the pool-derived key, without adding an equivalent safety net. A pool can enable Leios voting with a real configured BLS key (the PR's headline feature) while no peer's `leiosVoterPublicKeys` registry has a matching entry — every peer then falls back to the pool-derived key for verification, which won't match, and that pool's votes silently drop out of quorum with no error or warning on either side. This was already flagged by both an automated reviewer and a human SPO on this thread and remains unaddressed at HEAD (`c7e79c66`).

There's also no test covering `enableLeiosVoting`'s new key-selection branching in `node_leios.go`, so a regression here wouldn't be caught by the existing `ledger/leios` unit tests.

**Recommendation:** add a startup warning/validation when a configured key lacks a corroborating registry entry (or document the operational requirement clearly), plus a regression test for `enableLeiosVoting`, before merging.

node_leios.go
```diff
@@ -242,25 +241,24 @@ func (n *Node) enableLeiosVoting(creds *forging.PoolCredentials) error {
...
 	if n.config.leiosVoteSigningKeyFile != "" {
-		configured, loadErr := leios.LoadVoteSigningKeyFile(
+		key, err = leios.LoadVoteSigningKeyFile(
 			n.config.leiosVoteSigningKeyFile,
 		)
-		if loadErr != nil {
-			return fmt.Errorf("load leios vote signing key: %w", loadErr)
+		if err != nil {
+			return fmt.Errorf("load leios vote signing key: %w", err)
 		}
```
cryptodj413, line 252: The code removed here was the only safety net: it compared the loaded key against the pool-derived key and rejected a mismatch (`"configured leios vote signing key does not match the current prototype's pool-derived key"`). Now a configured `leiosVoteSigningKeyFile` is accepted unconditionally on this branch, with nothing checking it against `n.config.leiosVoterPublicKeys` (or any other source peers might use) before it's handed to `EnableVoting` below.

ledger/leios/manager.go
```diff
@@ -444,11 +444,18 @@ func (m *VoteManager) Stop() error {
...
 	poolKeyHash lcommon.PoolKeyHash,
 	key *VoteSigningKey,
-) {
+) error {
+	if key == nil {
+		return errors.New("nil leios vote signing key")
+	}
 	m.mu.Lock()
 	defer m.mu.Unlock()
+	if err := m.registry.RegisterPublicKey(poolKeyHash[:], key.PublicKey()); err != nil {
```
cryptodj413, line 453: `RegisterPublicKey` here accepts whatever key `enableLeiosVoting` passed in — including a real configured BLS key that no longer has to match the pool-derived fallback — with no check that it corresponds to what any peer's `leiosVoterPublicKeys` config actually has for this pool. If a peer hasn't independently configured this pool's real key, `handleResolvedPrototypeVote`'s derived-key fallback (~line 750) verifies against the wrong key, and this pool's votes silently drop out of quorum with no error on either side.

cryptodj413 (commented):
Follow-up to my earlier summary comment (#issuecomment-5167472043) — breaking the blocking issue into line-anchored notes at the two spots it actually touches.

node_leios.go
```diff
@@ -242,25 +241,30 @@ func (n *Node) enableLeiosVoting(creds *forging.PoolCredentials) error {
...
-		if !bytes.Equal(configured.PublicKeyBytes(), key.PublicKeyBytes()) {
-			return errors.New(
-				"configured leios vote signing key does not match " +
-					"the current prototype's pool-derived key",
+		if err := n.leiosVoteManager.ValidateVotingKey(poolKeyHash, key); err != nil {
```
cryptodj413, line 253: This closes the gap flagged in the prior review at this line (previously line 252): `enableLeiosVoting` now calls the new `VoteManager.ValidateVotingKey` before a configured key is ever handed to `EnableVoting`, and a startup failure here propagates all the way up through `node.go`'s `enableLeiosVoting` call to a fatal `initNode` error rather than being swallowed. A configured key with no matching registry entry, or one that doesn't match the registered entry, now aborts startup with a clear message instead of silently registering an unverifiable key.

ledger/leios/manager.go
```diff
@@ -444,11 +444,37 @@ func (m *VoteManager) Stop() error {
...
+		return fmt.Errorf("register local leios voting key: %w", err)
+	}
 	m.votingPool = slices.Clone(poolKeyHash[:])
 	m.votingKey = key
+	return nil
+}
+
+// ValidateVotingKey verifies that an operator-supplied voting key matches the
+// static registry shared with peers. It does not add a key to the registry.
+func (m *VoteManager) ValidateVotingKey(
```
cryptodj413, line 463: This is the fix for the previously-flagged gap at `RegisterPublicKey` (old line 453): `ValidateVotingKey` is now called from `node_leios.go` before `EnableVoting`/`RegisterPublicKey` ever sees the configured key, so a real BLS key with no corresponding local registry entry (or a mismatched one) is rejected up front rather than silently accepted and later failing verification on peers with no error on either side. `TestVoteManagerValidateConfiguredVotingKey` (manager_test.go:1241) exercises match/mismatch/missing-pool cases against a registry pre-populated with committee keys, so the fix has direct regression coverage. Note the residual (acknowledged in the ARCHITECTURE.md update) that this only enforces the local node's own registry is internally consistent — it can't verify that every peer's `leiosVoterPublicKeys` config was updated too, which remains an operator coordination responsibility, not something enforceable in-process.

cryptodj413 (commented):
Re-review of the latest commit (92c48051, "fix(leios): require configured voter key registration"). The blocking issue from the prior review (see https://github.com/blinklabs-io/dingo/pull/3060#issuecomment-5167472043 and the two inline comments below) is resolved. No new issues found in this commit; build, `go test -race ./ledger/leios/...`, `go vet ./...`, `golangci-lint run`, and `go test ./internal/architecture/...` (import boundaries) all pass clean against the PR head.

cryptodj413 (approved):

PR #3067, opened 2026-08-01
feat(ledger): answer GetChainDepState and GetPoolDistr2

### LINKED ISSUE #2997: Cardano-CLI Leadership Schedule Error
When I run:
```
cd ~/dingo
cardano-cli query leadership-schedule \
--testnet-magic 2 \
--genesis $HOME/dingo/config/cardano/preview/shelley-genesis.json \
--stake-pool-id pool1zwhpk00u9dzmsfcmkhyg9g38zwzp9x8sqzxqk595s2khylumpqv \
--vrf-signing-key-file vrf.skey \
--current
```

I get "<socket: 11>" error

Dingo logs show error:

`Jul 29 10:41:41 Test dingo[598482]: time=2026-07-29T10:41:41.378-05:00 level=INFO msg="listener: accepted NtC connection from unix@149" component=connmanager
Jul 29 10:41:41 Test dingo[598482]: time=2026-07-29T10:41:41.550-05:00 level=ERROR msg="unexpected connection failure: protocol error: unsupported query type: *localstatequery.ShelleyDebugChainDepStateQuery" component=peergov connection_id=/home/test/dingo/dingo.socket<->unix@149`

Tested on Dingo v0.68.0 with Cardano-cli version 11.0.0.0

chrisguiney:
Fixes #2997.

`cardano-cli query leadership-schedule` fails against dingo. It now works, verified end-to-end on a devnet.

## Why the failure was opaque

An unsupported LocalStateQuery does not fail that one query — it aborts the protocol. The node drops the connection and the caller sees only:

```
cardano-cli: bearer closed: "<socket: 11> closed when reading data, waiting on next header True"
```

Nothing names the query. Each missing handler had to be found by watching the failure boundary move.

## What was missing

Three things, discovered in sequence by repeatedly running the real `cardano-cli` against a devnet node:

1. **`GetChainDepState`** (sub-query 13) — cardano-cli reads the epoch nonce from here.
2. **Node-to-client protocol version 21** — cardano-cli gates on the *negotiated* version, not on the reply, so it refused at handshake before sending anything. Fixed upstream in [gouroboros#1927](https://github.com/blinklabs-io/gouroboros/pull/1927), released in v0.191.0.
3. **`GetPoolDistr2`** (sub-query 36) — having negotiated 21, cardano-cli sends this in place of the deprecated `GetPoolDistr`.

## GetChainDepState

The Praos serialisation: `encodeVersion 0` wrapping an eight-element record — last slot, opcert counters, then the evolving, candidate, epoch, previous-epoch, lab and last-epoch-block nonces.

Field order is load-bearing. A reordering still produces valid CBOR that deserialises into the wrong fields, so the tests assert by decoding through the client-side decoder rather than against our own idea of the shape.

Review found six things this got wrong, all since fixed on the branch:

- **The opcert counters covered only currently registered pools.** The chain's record of the highest accepted issue number for a cold key does not expire when its pool leaves the active set — the node still enforces it against any block claiming that key. They now come from the issuer record itself, which also drops a one-query-per-pool loop.
- **The lab nonce reused the last-epoch-block nonce.** That is right only until the epoch's first block lands, since the carried value is where Praos' one-block lag had reached when the epoch opened. It is now derived from the tip's parent hash, the same rule `epochLabNonce` applies at a boundary.
- **The last slot and epoch came from the in-memory snapshots while the nonces came from a database transaction.** Those snapshots are published after the write that advances the chain, so a transaction opened before a boundary could be paired with an epoch number it has no record for — a populated last slot beside a *neutral* epoch nonce, from which cardano-cli computes a schedule that is wrong rather than absent. Both now come from the same transaction as everything else.
- **The tip block was looked up by hash.** That resolves through the block hash index, which has only been written since #1915 and reports a miss rather than scanning. A database still carrying pre-index blocks therefore answers "no such block" for one it holds — and a node restarted on such a database has exactly one of them as its tip — so the lab silently fell back to the epoch's carried value. The tip's slot and hash address the blob directly, needing no index.
- **The evolving and candidate nonces came straight off the epoch row** (caught by @wolf31o2). Those columns are checkpoints: they hold what the pair was when the epoch *opened*, and nothing rewrites them as blocks land. Both move with every block — the evolving nonce folds in each block's VRF output, the candidate tracks it until the randomness-stabilisation cutoff and then freezes — so from the epoch's first block onward the reply described the boundary while the rest of it described the tip. They are now recomputed at the acquired tip via `computeCandidateNonceAsOf`, which is `computeCandidateNonce` stopped early: the freeze rule and both lookup paths stay shared with the boundary computation instead of being restated in the query.
- **The reply always claimed Praos** (caught by @wolf31o2). dingo supports Shelley through Alonzo, which run TPraos and serialise a *different record* — version 1 wrapping `[lastSlot, [counters, evolving, candidate]]` against version 0 wrapping a flat eight-field array. The version tag selects the shape rather than labelling it, so a node in a TPraos era handed the client a payload whose nesting and arity did not match what the version promised, carrying four nonce fields that do not exist there — and a node syncing from genesis sits in those eras for a long stretch. The layout is now chosen from the era of the epoch containing the tip.

## GetPoolDistr2

Each pool's share of the active stake, its VRF key hash, and the total active stake.

**The distribution is read from the mark snapshot at `praos.StakeSnapshotEpoch` — the same snapshot dingo's own leader election reads.** This was the one design question I did not want to guess at. Reading live stake instead would have produced a schedule that disagrees with the node that generated it, telling an operator they lead slots the node will not let them mint. Sourcing it from the election path makes the query agree with node behaviour by construction.

Per-pool stakes and their total are read under one transaction; taken separately, an epoch boundary landing in between would yield fractions that do not sum to one.

Two cases are handled rather than left to fall over: an empty snapshot — where a chain sits before its first snapshot is taken — yields a zero fraction instead of dividing by zero, and a pool with snapshot stake but no registration on record fails the query loudly, since its stake is still counted in the total and dropping it would leave the reported fractions summing to less than one with nothing saying so.

Which registration decides a pool's VRF key came up in review: the Haskell node answers from a distribution snapshotted at the epoch boundary, so it reports the key in force then. dingo resolves the producing pool's key *live* at validation time, in `verifyRegisteredVrfKey`, so reporting the snapshot-era key would describe a schedule dingo itself would reject. The lookup now goes through `registeredPoolVrfKeyHash`, the same function validation uses, so the two cannot drift apart silently.

## Verification

Unit tests are not sufficient here: a wrong epoch nonce or stake fraction still yields *a* schedule, just a wrong one. So the result was checked for magnitude, not just well-formedness.

`cardano-cli` returned **76 slots**. Expected for a one-third-stake pool at `activeSlotsCoeff` 0.4 over a 500-slot epoch is ~66.7, σ≈7.6 — so ~1.2σ. A fraction of 1.0 would have given roughly 200; a broken one, zero.

The failure boundary was also watched moving at each step, which is what confirmed each fix landed:

| after | failure |
|---|---|
| — | `unsupported query type: ShelleyDebugChainDepStateQuery` |
| ChainDepState | refused at handshake, needs a newer NtC version |
| gouroboros v21 | `unknown Block query type: 36` |
| gouroboros PoolDistr2 | `unsupported query type: ShelleyPoolDistr2Query` |
| this PR | schedule returned |

## Dependency bump (now upstream's)

This branch originally bumped gouroboros to v0.191.0 and absorbed the fallout. Since then `main` moved to v0.191.2 and made the same adjustments itself, so the rebase dropped nearly all of it: `go.mod`, `go.sum`, `ledger/eras/conway.go` and `ledger/eras/validation.go` are no longer touched here, and the Alonzo/Babbage `UtxoValidationRules` skip indices now come from upstream.

What survives is one test change. The startup guard's test hardcoded the index in its expected panic message, so a rule reorder broke both the constant and the string describing it. It now derives the message from the constant, leaving one place to touch next time.

The blockfrost adapter still propagates `BuildScriptPurpose`'s error, which upstream's own call site does not need — that call site returns an error anyway, so the diagnostic is kept rather than dropped.

## Docs

`DATABASE.md` updated: the `pool_opcert_sequence` row now records `LatestPoolOpCertSequences`, the `GROUP BY pool_key_hash` read added for these counters, and why its set is keyed by issuer rather than joined to `pool`.

`ARCHITECTURE.md` updated: the Local State Query section now lists these two leaves, including the era-dependent `GetChainDepState` layout and where `GetPoolDistr2` sources its stake and VRF keys. The Epoch Nonce Computation section records that the epoch row's evolving/candidate values are checkpoints, and that a reader needing the pair mid-epoch goes through `computeCandidateNonceAsOf` so the query and consensus cannot drift.

---
## Summary by cubic
Implements `GetChainDepState` and `GetPoolDistr2` so `cardano-cli query leadership-schedule` works end to end and matches node behavior at the tip. Improves performance with an indexed counter read and chunked queries, and degrades gracefully on inconsistent data instead of dropping the connection.

- **New Features**
  - `GetChainDepState`: era-aware (TPraos vs Praos) encoding from the consensus mode; last slot; op-cert counters via `LatestPoolOpCertSequences`; evolving/candidate nonces recomputed at the acquired tip with `computeCandidateNonceAsOf`; lab nonce from the tip’s parent; tip addressed by point; one DB transaction with matched snapshots.
  - `GetPoolDistr2`: fractions from the mark snapshot of the distribution epoch (resolved inside the same txn); total active stake clamped to non-zero; VRF key via the same live registration path used by validation; optional pool filter returns only requested pools (fractions still over the full total).
  - Reliability/perf: counters include retired pools; filtered reads use bounded, chunked IN lists; repeated pool hashes are deduplicated before chunking to avoid duplicate rows and broken associations; on inconsistent data the handlers log and continue instead of aborting the LocalStateQuery protocol.

- **Migration**
  - Introduces schema v2 (`v2alpha1`) adding covering index `idx_pool_opcert_sequence_pool_sequence` for the counter fold; built automatically at startup and may take time on large databases.

<sup>Written for commit 7f9e67cdf64ba6701e4828797e9bc88aa79534d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3067?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for Shelley pool distribution queries, including stake totals, pool fractions, and VRF hashes.
  * Added Praos chain-dependent state query support with epoch nonces and operational certificate counters.
  * Enabled dispatch for additional Shelley ledger queries.

* **Bug Fixes**
  * Improved handling of invalid redeemer script purposes with clear transaction context.
  * Prevented errors when calculating pool distributions with zero total stake.
  * Updated validation rules for the latest ledger compatibility changes.


chrisguiney:
@cubic-dev-ai review

ledger/queries_chaindepstate.go
```diff
@@ -0,0 +1,254 @@
...
+	// can enter the epoch. What remains is the chain that has applied no
+	// blocks at all, where neutral is the right answer.
+	current, err := ls.db.GetEpochBySlot(tip.Point.Slot, txn)
+	if err != nil {
+		return nil, err
+	}
+	var carriedLabNonce []byte
+	if current != nil {
+		carriedLabNonce = current.LastEpochBlockNonce
+		state.EvolvingNonce = nonceFromBytes(current.EvolvingNonce)
```
wolf31o2: These values come from the epoch row, which is written at epoch boundaries and is not updated for every block. After the first block in the current epoch, EvolvingNonce and CandidateNonce can therefore be stale. The existing consensus path recomputes current-epoch values with computeCandidateNonce from block nonce/VRF state; this query should do the same at the acquired tip before returning chain-dependent state.

ledger/queries_chaindepstate_test.go
```diff
@@ -0,0 +1,638 @@
...
+	ls := &LedgerState{db: db}
+	ls.publishSnapshotsLocked()
+
+	result, err := ls.Query(chainDepStateQuery())
+	require.NoError(t, err,
+		"the query must be handled rather than aborting the protocol")
+	require.NotNil(t, result)
+}
+
+// TestQueryShelleyDebugChainDepState_DecodesAsPraosState checks the reply
```
wolf31o2, line 88: Please add a regression case with an epoch-row checkpoint, at least one block in that epoch, and tip-time block nonce/VRF state that differs from the checkpoint. Decode the reply and assert EvolvingNonce and CandidateNonce reflect the acquired tip rather than the boundary values.

wolf31o2 (changes requested):
Request changes: GetChainDepState currently returns stale current-epoch nonce state. The handler reads EvolvingNonce and CandidateNonce from the epoch row, but those fields are boundary checkpoints and are not updated for each applied block. The consensus path recomputes them from the current epoch block nonce/VRF state, so a mid-epoch query can report the wrong chain-dependent state. Please compute the values at the acquired tip before encoding the reply.

ledger/queries_chaindepstate.go
```diff
@@ -0,0 +1,287 @@
...
+			}
+		}
+	}
+
+	state.LabNonce, err = ls.chainDepStateLabNonce(txn, tip, carriedLabNonce)
+	if err != nil {
+		return nil, err
+	}
+
+	return []any{
```
wolf31o2, line 263: This unconditionally serializes the result with the Praos version/layout. The repository supports Shelley, Allegra, Mary, and Alonzo, whose GetChainDepState wire format is TPraos v1: `[1, [lastSlot, [opCertCounters, evolvingNonce, candidateNonce]]]`; only Babbage and later use this Praos v0 eight-field record. During a TPraos-era query this reply will decode with the wrong protocol and shape. Select the layout from the acquired/current era and add a regression test for a Shelley–Alonzo era.

wolf31o2 (changes requested):
REQUEST CHANGES: GetChainDepState always emits the Praos v0 layout, but Dingo supports Shelley–Alonzo TPraos eras, which require version 1 and a different nested payload. Focused ledger, metadata SQLite, and Blockfrost tests pass; please add era-aware serialization and a TPraos regression test before merge.

chrisguiney:
@cubic-dev-ai review

ledger/candidate_nonce.go
```diff
@@ -125,8 +168,8 @@ func (ls *LedgerState) computeCandidateNonce(
 		prevEvolvingNonce,
 		prevCandidateNonce,
 		epochStartSlot,
-		epochEndSlot,
-		cutoffSlot,
+		foldEndSlot,
```
wolf31o2, line 171: `computeCandidateNonceAsOf` passes the requested `foldEndSlot` into this function as `epochEndSlot`, but this lookup is still used as the last block of the entire epoch. When the acquired tip is mid-epoch and later blocks remain in the blob store (for example after a rollback or on a stored fork), this selects a block beyond the tip and returns its evolving nonce. Please bound this lookup by the requested fold end and add a test where a later stored block must be ignored.

ledger/candidate_nonce.go
```diff
@@ -147,8 +190,8 @@ func (ls *LedgerState) computeCandidateNonce(
...
```
wolf31o2, line 193: The candidate lookup has the same as-of bug: it uses the freeze `cutoffSlot` even when the requested tip is earlier than that cutoff. A block between the acquired tip and the cutoff can therefore move the reported candidate nonce beyond the tip. Use the effective candidate bound (`min(cutoffSlot, foldEndSlot)`) for this lookup, and cover the pre-cutoff/mid-epoch case in the regression test.

wolf31o2 (changes requested):
Request changes: the as-of nonce fast path can report consensus state beyond the acquired tip. Please bound both blob lookups by the requested fold end (and add a regression fixture with later stored blocks after the tip). Until corrected, GetChainDepState can return plausible but incorrect nonces after rollback or when stored fork blocks extend past the active tip.

ledger/candidate_nonce.go
chrisguiney: Both bounds are already in place — `computeCandidateNonceAsOf` passes `foldEndSlot` here (line 171) and `min(cutoffSlot, foldEndSlot)` as the candidate bound (line 194), and `BlockBeforeSlot` is exclusive, so `foldEndSlotForTip(tip) = tip+1` includes the tip block and nothing above it.

What was missing is the coverage you asked for, and the reason it reads as the bug: the callee still named these parameters `epochEndSlot`/`cutoffSlot` and its comments still said "the last block of the epoch" and "frozen at the stability window cutoff". Both now carry the caller's names, and the comments say why the fold's end is the bound.

Added in 6edd9ac8, along with the test — tip at 1200 with a further block stored at 1300, which is the rollback/stored-fork case you named. Verified it has teeth by reintroducing the bug: bounding this lookup at the epoch's end returns slot 1300's nonce and fails the evolving assertion.

ledger/candidate_nonce.go
chrisguiney: This line already passes `candidateBound`, which is `min(cutoffSlot, foldEndSlot)` computed at line 159 — so the effective bound is what you describe.

But you were right that nothing held it there. The existing mid-epoch test puts the tip at 1950, past the 1940 cutoff, so the candidate has already frozen and `min()` is a no-op — I reverted this argument to `cutoffSlot` and that test stayed green. The pre-cutoff case, where the two bounds actually differ, had no coverage at all.

6edd9ac8 adds it: tip at 1200, a block stored at 1300, both below the cutoff. There the candidate still tracks the evolving nonce, so the fold's end is the binding one — with `cutoffSlot` the reply names slot 1300, a block the tip precedes. Each of the two mutations fails exactly one assertion of the new test and neither fails anything that existed before it.

ledger/queries_pooldistr2.go
```diff
@@ -0,0 +1,199 @@
...
+// is negotiated.
+//
+// The distribution is read from the same snapshot the node elects leaders
+// from — the mark snapshot at praos.StakeSnapshotEpoch, not live stake. An
+// operator checking a schedule against the node would otherwise be told they
+// lead slots the node will not let them mint.
+func (ls *LedgerState) queryShelleyPoolDistr2(
+	q *olocalstatequery.ShelleyPoolDistr2Query,
+) (any, error) {
+	epoch := ls.loadConsensusSnapshot().currentEpoch.EpochId
```
wolf31o2: Please derive this epoch from the transaction-consistent tip/epoch row. Reading `loadConsensusSnapshot()` before opening `txn` allows an epoch-boundary race: the snapshot epoch and the transaction tip can belong to different epochs, so GetPoolDistr2 may return a distribution inconsistent with the acquired state.

wolf31o2 (changes requested):
Request changes: GetPoolDistr2 derives snapshotEpoch from the in-memory consensus snapshot before opening the database transaction. At an epoch boundary, the in-memory epoch and transaction tip can come from different epochs, so the returned stake distribution can be inconsistent with the acquired state. Please derive the epoch from the transaction-consistent tip/epoch row instead.

ledger/queries_pooldistr2.go
chrisguiney: Confirmed and fixed in 5466c9b1 — you're right, and it was the one place on the branch still on the old pattern. `GetChainDepState` resolves exactly this pairing by taking its tip and epoch from a single transaction; `GetPoolDistr2` was left reading the snapshot first.

Now the tip comes from `ls.db.GetTip(txn)` and the epoch from `GetEpochBySlot(tip.Point.Slot, txn)`, so the mark snapshot is the one the acquired state is actually in. An empty chain has no epoch row covering its tip, so epoch 0 remains the answer there — the same value the snapshot's zero gave before.

The part that made this worth catching is that nothing downstream could: stake rows for the wrong epoch are internally consistent, since every fraction is a share of a total summed over that same wrong snapshot. They sum to one and decode cleanly, so cardano-cli gets a schedule that is wrong rather than absent — the exact failure sourcing this from the election path was meant to rule out.

Regression test written first and confirmed failing: the epoch row covering the tip says 5 while the in-memory snapshot says 9, and the same pool holds 3,000,000 in one mark snapshot and 7,000,000 in the other. It reported 7,000,000 before the fix.

chrisguiney:
@cubic-dev-ai review

database/plugin/metadata/sqlstore/pool.go
```diff
@@ -272,30 +272,45 @@ func (s *Store) GetPools(
...
+	// is the chain's pool count rather than anything this code picks.
+	//
+	// ParameterLimit is the conservative figure the store contracts to, not
+	// each driver's true ceiling: SQLite has allowed 32766 bound parameters
+	// since 3.32 and PostgreSQL/MySQL allow 65535, so at Cardano's few
+	// thousand registered pools an unchunked list happens to fit today. It
+	// stays chunked because that headroom is a property of the deployment
+	// rather than of this query, and exceeding it fails the read outright
+	// instead of degrading.
+	for start := 0; start < len(hashes); start += s.dialect.ParameterLimit() {
```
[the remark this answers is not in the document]
chrisguiney: Valid on both counts — fixed in 7f9e67cd, and the second half was the part I had missed.

Chunking took away a semantic the single statement had for free: `IN` is set membership, so naming a value twice matches its row once, but two mentions in two chunks match in both and the results are concatenated.

You are right that the association hydration makes it worse than a repeated row. `loadPoolsAssociations` keys its index by pool ID and keeps only the last position, so the earlier copy comes back with empty registrations and retirements — a pool that reads as never registered. `registeredPoolVrfKeyHash` decides on `len(pool.Registration)`, so which copy a caller reached would decide whether the pool appeared to have a VRF key at all.

I applied the same fix to `GetPoolStakeSnapshotsForPools`, which this PR added with the same chunking shape. That one takes a list the node does not choose: it is the pool filter off the wire, and `PoolFilter` returns the client's items verbatim without collapsing repeats, so a client naming a pool twice got its snapshot row twice.

Neither is reachable through `GetPoolDistr2` as it stands — its key hashes come from a map, and `markStakeForPools` folds rows into one — so this is a latent contract break rather than a live failure. Worth closing regardless, since the callers most likely to hit it are the ones that do not exist yet and nothing in either signature warns that a repeated hash is unsafe.

Both regressions were written first and confirmed failing (2 rows for the repeated pool, 1501 rows for 1500 pools), with the repeat placed at opposite ends so the mentions straddle a chunk boundary. The `GetPools` one also asserts the surviving copy carries its registrations, which is the half that bites before the row count does.

wolf31o2 (approved):

PR #3071, opened 2026-08-02
fix(consensus): stop unfetchable queued header from wedging block production

### LINKED ISSUE #3029: DevNet: tolerate cold-start slot progress delay

wolf31o2:
Closes #3029 

A tip slot battle could stall block production for minutes. Two producers forge at the same slot on the same parent, each rolls its own block back in favor of the other's, and each then holds a queued header whose body only the other node ever had. Neither can serve it, so the header stays at the head of the queue and chain.addBlockLocked rejects every locally forged block with "block hash X does not match first pending header hash Y" (chain/chain.go, surfacing as "forge check failed" from ledger/forging/forger.go).

Observed on the all-dingo DevNet: one dingo-1 process logged 71 consecutive forge failures against 8 distinct latched header hashes, with 48150 GetBlockRange calls for one slot-10 point over 157s and zero blocks returned, and tips frozen for 8 to 144 slots. This is the same condition #3029 reports as a cold-start progress delay: the first slot battle happens at the very first contested slot, which is why it looks like slow startup and then "resolves itself".

Three defects combine.

The blockfetch server promised batches it could not deliver. ChainManager.removeBlockByIndex deletes a rolled-back block's row but retains the block in the manager's LRU cache so non-primary chains can reconcile against it, and blockByPoint answers from that cache. A chain iterator built from such a point is positioned at an index the chain no longer has, so the range server sent StartBatch, found nothing to stream, and sent BatchDone - indistinguishable from a served range, and never reaching the NoBlocks stuck-peer valve. The forward and reverse iterator constructors now require the start point to be the block the chain still holds at that index, so the request takes the existing start-point-not-found branch and answers NoBlocks. Measured on DevNet: the repeated-single-range floods disappeared entirely (worst case 15000-48000 requests down to 6) and "start point not found in chain" now fires 20-55 times per node.

The fetching side had no recovery on the path a NoBlocks reply takes. It surfaces as a synchronous GetBlockRange error from BlockfetchRequestRangeFunc and never reaches BatchDone, so accounting hung off the BatchDone path could not see it. Failure now has two shapes that both count: that error, and a StartBatch/BatchDone pair carrying no blocks. Counting them separately would let the two alternate while each stayed under its own bound.

The accounting lives in startQueuedBlockfetchLocked, the single point every queued-range request passes through, rather than in the callers, because every caller treats the error differently and several treat it as advisory: the two fork-resolution restarts in tryResolveFork log "failed to start blockfetch after fork rollback" and report the fork resolved, the local-rollback recovery logs "failed to start blockfetch after local rollback recovery", and the await-reply path logs at ERROR and publishes a LedgerErrorEvent. None retried, so the pipeline sat idle with the header still queued. The shadow-blockfetch dispatch is intentionally excluded: it is a duplicate request whose failure says nothing about the primary in flight.

The count is keyed to the range start point rather than being a global consecutive streak. Failures against one unfetchable range are minutes apart, and in between the node fetches normally from other peers (75 blocks delivered on dingo-1 against 17 failures in one run) while fork resolution, connection switches and header mismatches call clearQueuedHeaders repeatedly, after which the peer re-offers the same header. A global streak is reset by all of that and fires only when failures happen to land back to back: two identical DevNet runs on the same build produced 1 recovery against 169 wedge events and 9 against 81. The record therefore survives interleaved deliveries for other ranges and header-queue churn, and is discarded only when the tracked range itself is delivered, so a peer that is briefly behind is not punished and a miss on a different range starts its own count. After firing, the count restarts so a re-offered header must earn a fresh set of failures rather than triggering a resync on every later miss.

On reaching the bound the queued headers are dropped and a fresh intersect is requested. Dropping them is the purpose: it is what releases forging. Where the recovery fired more often, DevNet wedge counts roughly halved and tip freezes fell from 120+ slots to about 30.

The NoBlocks stuck-peer valve is deliberately left alone. It requires 5 consecutive requests for the same (connection, start point), and blockfetchResetNoBlocks clears the count on every validated range request from that peer. DevNet shows a maximum of 4 repeats for any one tuple among 918 served requests on the same connection, so it correctly does not fire: these are healthy peers briefly asking for a body neither side still has, not a peer stuck in a loop. That invariant is now documented at the constant so it is not "fixed" into dropping good peers.

Tests (added only, no pre-existing assertion changed):
- chain: a forward and a reverse iterator must both reject a start point the chain rolled back, on persistent and in-memory chains, while the surviving tip stays iterable; asserts the rolled-back block is still resolvable by point so the retained-cache precondition cannot silently disappear.
- ledger: repeated NoBlocks errors must drop the queued headers and publish chainsync.resync, driven both through startQueuedBlockfetchLocked and through restartQueuedBlockfetchAfterForkLocked - the caller observed wedging on DevNet, which discards the error after logging it.
- ledger: failures against the same range must accumulate across interleaved deliveries for other ranges and clearQueuedHeaders churn (the regression guard for a bound that tripped only by luck), while failures spread over distinct ranges must not accumulate and a delivered block must discard that range's record.

Verified: go build, go test -race on ./chain/... ./ledger/... ./ouroboros/... ./event/... (ledger root 586s), conformance, chainsync/api/integration, golangci-lint 0 issues, make import-boundaries. nilaway and modernize findings are identical to base 37d8448d (all in untouched files). ARCHITECTURE.md updated for the blockfetch server start-point contract and the per-range failure accounting plus its new resync reason; DATABASE.md checked and not affected - no model, migration, query surface, blob layout, or pruning change.

---
## Summary by cubic
Prevents tip-slot battle wedges from halting forging by rejecting rolled-back iterator start points, accepting common points on in-memory forks, and capping retries for unfetchable queued header ranges.

- **Bug Fixes**
  - `chain`: Iterators now require the start point to be a block the chain still holds; rolled-back points return `ErrBlockNotFound`. In-memory forks accept common-prefix points by resolving via the primary chain. Added `primary` pointer, `blockByIndexLocked`, and `lockBlockIndexReadLocks`; updated iterators, `RecentPoints`/`IntersectPoints`, `BlockBeforeSlot`, and rollback paths to use safe chain → primary → manager lock order.
  - `ledger`: Track per-range failures (count `NoBlocks` request errors and empty `StartBatch`/`BatchDone`), and after 3 failures drop queued headers, clear the active blockfetch connection, emit `chainsync.resync` with reason "blockfetch could not obtain the queued header range", and request a new intersect; shadow requests excluded. Progress on the same range clears its record.
  - `ouroboros`: Documented unchanged `NoBlocks` valve behavior.

- **Docs & Tests**
  - ARCHITECTURE updated for iterator start-point contract, per-range accounting, and lock-order notes.
  - Added tests for rolled-back-point rejection (persistent/in-memory), acceptance of common points on in-memory forks with inclusive/exclusive and forward/reverse iterators, empty-batch and `NoBlocks` recovery paths, per-range accumulation/isolation with progress resets, and concurrent in-memory fork creation vs point enumeration.

<sup>Written for commit f49e5f0d8d2cbc0a7911e12baea872f91264adf0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3071?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented iterators from returning rolled-back or unavailable blocks.
  * Improved handling of repeated block-fetch failures and empty responses.
  * Preserved failure tracking across unrelated queue activity while resetting it after successful progress.
  * Prevented indefinite empty batches by dropping unavailable headers and triggering chain synchronization recovery.

* **Documentation**
  * Clarified block-fetch retry behavior and recovery conditions.


ledger/chainsync.go
```diff
@@ -4577,6 +4725,20 @@ func (ls *LedgerState) handleEventBlockfetchBatchDone(e BlockfetchEvent) error {
 			"remaining_headers", remainingHeaders,
 		)
 	}
+	// A batch that completed without delivering a block while headers stayed
+	// queued is one of the two shapes of "could not obtain the queued range"
+	// (the other is a NoBlocks reply, recorded in
+	// startQueuedBlockfetchLocked). Both feed the same streak.
+	if receivedBlockCount == 0 && remainingHeaders > 0 {
```
arepala-uml, line 4772: Only count an empty batch when `fromActive` is true. An empty `BatchDone` from a speculative shadow request does not prove the primary cannot serve the range. Currently, repeated empty shadow completions can reach the failure limit, clear queued headers, and trigger an unnecessary resync. 

**Suggested fix**

The code already computes `fromActive` before promoting a shadow connection. Adding it to the condition ensures only an empty response from the original active request increments the failure counter. Please add fromActive to this condition and cover the shadow-empty-batch case with a regression test.

```
if fromActive && receivedBlockCount == 0 && remainingHeaders > 0 {
    batchStart, _ := ls.chain.HeaderRange(blockfetchBatchSize)
    if ls.noteBlockfetchRangeUnavailable(
        e.ConnectionId,
        batchStart,
        "batch completed without delivering a block",
    ) {
        return nil
    }
}
```

arepala-uml (approved):

PR #3112, opened 2026-08-04
fix(mempool): tolerate busy revalidation journals

wolf31o2:
## Summary

- increase the mutation journal capacity for sustained mempool workloads
- replay at most the configured delta cap per catch-up round
- treat an overloaded revalidation pass as a safe retry, preserving the live pool
- document the candidate-overlay replay behavior

## Testing

- go test -race ./mempool
- DINGO_MEMPOOL_BENCH_DURATION=3s DINGO_MEMPOOL_BENCH_PREFILL=5000 go test -run=^$ -bench='^BenchmarkMempoolPluginsDegenerate/large-occupancy/(fifo|dag)$' -benchtime=1x ./mempool

Signed-off-by: Chris Gianelloni <wolf31o2@blinklabs.io>

---
## Summary by cubic
Improves mempool revalidation under heavy mutation backlogs. Enlarges the journal, bounds per‑round replay with a backlog‑scaled budget, documents actual termination, speeds journal scans, and treats exhausted budgets as safe retries.

- **Bug Fixes**
  - Increase default revalidation journal capacity to `1<<20`; document worst‑case footprint (~42 MiB slice, plus pinned TXs) and note a full drain measured ~5.3 ms.
  - Scale total catch‑up rounds with the observed backlog and include rounds already spent so large deltas converge and publish; the loop ends only on an empty journal, the journal cap, or a replay error.
  - Replay only a bounded journal window per round via binary search (not the whole suffix) to reduce lock time; the window is never nil.
  - Treat `errRevalidationCatchup` as a non‑fatal defensive guard (currently unreachable); `rebuildOverlay` exits early with the live pool unchanged and retries on the next chain update; sustained mutation pressure ends in `errRevalidationJournalOverflow`.
  - Add tests: convergence on backlogs larger than one round budget; budget accounts for rounds already spent and always leaves headroom while pending; window clones only the bounded slice and returns empty, not nil; docs updated to describe the private candidate overlay, bounded batches, and termination.

<sup>Written for commit b3fb9ba87a48ab59e908bbc4b4c7b846dcd63608. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mempool/mempool.go
```diff
@@ -1102,12 +1112,16 @@ func (m *Mempool) rebuildOverlayAttempt() ([]event.Event, error) {
...
+			if requiredRounds := (len(delta)+m.revalidationDeltaCap-1)/m.revalidationDeltaCap +
+				maxRevalidationCatchupRounds; requiredRounds > catchupRounds {
+				// Scale the total budget to the observed backlog while keeping
+				// each replay round bounded. New mutations can extend this
+				// budget again until the journal overflows.
+				catchupRounds = requiredRounds
 			}
+			// Bound the amount of replay work per round so that the
+			// catch-up loop can observe and reconcile new mutations.
+			applyCount := min(len(delta), m.revalidationDeltaCap)
```
chrisguiney: **Blocking: this change makes the catch-up drain quadratic.**

The previous form was `applyCount = len(delta) - cap` — drain the bulk in one round, leave a `cap`-sized residual — which converged in about two rounds. Bounding to `cap` per round instead means roughly `N/cap` rounds, and each of those rounds calls `mutationsAfter`, which does a linear scan from index 0 **and clones the whole remaining journal** (`mempool.go:1152-1164`):

```go
idx := len(journal)
for i := range journal { if journal[i].seq > seq { idx = i; break } }
return slices.Clone(journal[idx:])   // full remainder, every round
```

That clone was nearly free at two rounds. At `N/cap` rounds it dominates, and it happens while holding `mutationMutex` — the lock every `AddTransaction`/`RemoveTransaction` contends on. Measured against the real function:

| Backlog | Before | After | Delta |
|---|---|---|---|
| 1,000 | 3.1 µs, 2 rounds, 43 KB | 65 µs, 16 rounds, 358 KB | 21x / 8x |
| 10,000 | 22 µs, 2 rounds, 404 KB | 9.9 ms, 157 rounds, 32 MB | 446x / 79x |
| 65,536 | 0.84 ms, 2 rounds, 2.6 MB | 172 ms, 1,024 rounds, 1.35 GB | 205x / 513x |
| 262,144 | 0.62 ms, 2 rounds, 10 MB | 1.73 s, 4,096 rounds, 21.5 GB | 2,770x / 2,049x |

The cloned-entry totals matched `N^2/(2*cap)` exactly. Extrapolating to the new `1<<20` journal ceiling puts a full drain in the tens of seconds.

I don't think the per-round bound itself is wrong — the goal of letting the loop observe new mutations is reasonable. The waste is in `mutationsAfter`: only `delta[:applyCount]` is ever read, so cloning everything past `cap` is dead work. Cloning at most `revalidationDeltaCap` entries, and returning the remaining count separately for the `requiredRounds` computation on the line above, would keep this bounded per round at linear total cost. Since `seq` is monotonic the leading scan can also be a binary search rather than O(N).

Worth a benchmark guarding this once fixed — the existing tests use `revalidationDeltaCap = 1` with a 48-entry backlog, so nothing in the suite would notice the regression.

mempool/mempool.go
```diff
@@ -1102,12 +1112,16 @@ func (m *Mempool) rebuildOverlayAttempt() ([]event.Event, error) {
 			}
 			m.mutationMutex.Unlock()
 
-			applyCount := len(delta)
-			if applyCount > m.revalidationDeltaCap {
-				// Drain the excess off-lock, leaving at most the configured
-				// residual for the next catch-up/finalization pass.
-				applyCount -= m.revalidationDeltaCap
+			if requiredRounds := (len(delta)+m.revalidationDeltaCap-1)/m.revalidationDeltaCap +
```
chrisguiney: Context for the comment below, not a separate problem: this is the part that makes the round count grow to `ceil(len(delta)/cap) + 16`, which is correct and necessary given the per-round bound — but it's also what turns "one expensive clone" into "N/cap expensive clones".

If `mutationsAfter` is changed to clone only what will be applied, this calculation still needs the *total* remaining backlog to size the budget. Returning both (the bounded slice, and the full remaining count) from `mutationsAfter` would let this line keep working unchanged while the clone stays O(cap).

Minor readability note while you're here: `requiredRounds` is computed inside an `if` initializer that spans a line break in the middle of an arithmetic expression, which is hard to parse. Hoisting it to its own statement above the `if` would read more easily.

mempool/mempool.go
```diff
@@ -950,6 +954,11 @@ func (m *Mempool) rebuildOverlay() error {
 		if errors.Is(err, errValidationSnapshotChanged) && attempt == 0 {
 			continue
 		}
+		if errors.Is(err, errRevalidationCatchup) {
+			// The live pool is unchanged. A later chain update will retry
+			// revalidation after mutation pressure subsides.
+			return nil
```
chrisguiney, line 980: **Blocking (observability): this makes the catch-up condition completely silent.**

Previously `errRevalidationCatchup` propagated out of `rebuildOverlay` to the caller at `mempool.go:913`, which logged `"mempool overlay rebuild failed"`. Returning `nil` here means there is now no log line, no metric, and no counter for it — the condition is invisible.

Not raising ERROR for a benign retry is the right call, so I'm not asking to revert the classification. But this is the one state where silence is genuinely costly: as the new test's own comment establishes, this path is *not* self-healing across attempts, since each attempt restarts from a fresh journal. If the round-scaling above ever proves insufficient on some workload, a node would sit with a never-republished pool — stale invalid transactions retained, DAG never rebuilt — and nothing would indicate it.

A `Debug`/`Warn` line, or better a counter alongside the existing mempool metrics, would make that diagnosable without noise. The comment explaining *why* returning nil is safe is good and worth keeping either way.

mempool/mempool.go
```diff
@@ -46,7 +46,7 @@ const (
 	DefaultCleanupInterval        = 1 * time.Minute
 	DefaultRevalidationDeltaCap   = 64
 	DefaultConsumerCacheSize      = 1024
-	defaultRevalidationJournalCap = 65_536
+	defaultRevalidationJournalCap = 1 << 20
```
chrisguiney, line 55: Non-blocking, but worth stating the intended ceiling explicitly since this is a 16x increase.

`unsafe.Sizeof(mempoolMutation{})` is 40 bytes, so `1<<20` is ~41.9 MB for the backing slice alone. Each entry also references an `*appliedTx`, a `*MempoolTransaction`, and a `removed map[string]struct{}`, so a full journal is realistically a few hundred MB — and those pointers pin transactions that have already left the pool, keeping them from GC for the duration of the pass.

Two things that would help: a comment here recording the intended worst-case footprint (the old 65,536 value carried no such note either, but the number is far less alarming), and awareness that this same N drives the drain cost in the catch-up loop — so raising the ceiling and making the drain quadratic compound each other rather than trading off.

chrisguiney (changes requested):
The diagnosis here is genuinely sharp, and I want to call it out before the findings: the doc comment on `TestMempool_RevalidationConvergesOnBacklogLargerThanRoundBudget` pins down a failure mode that is easy to miss — because `rebuildOverlay` swallows the catch-up error *and every attempt restarts from a fresh journal*, no progress carries across attempts, so a backlog past the fixed budget meant the candidate was **never** published: no invalid-transaction removal and no DAG rebuild, indefinitely. That reasoning is captured well in both the test and `ARCHITECTURE.md`, and the "each attempt restarts from a fresh journal" insight is the part that makes the bug non-obvious.

My concern is with the mechanism chosen to fix it rather than the goal.

Changing `applyCount` from `len(delta) - cap` to `min(len(delta), cap)` turns the catch-up drain quadratic, because `mutationsAfter` clones the **entire remaining journal** on every round while only `revalidationDeltaCap` entries are ever read from it. Driving the real `mutationsAfter`:

| Backlog | Before (`len-cap`) | After (`min(len,cap)`) | Delta |
|---|---|---|---|
| 1,000 | 3.1 µs, 2 rounds, 43 KB | 65 µs, 16 rounds, 358 KB | 21x / 8x |
| 10,000 | 22 µs, 2 rounds, 404 KB | 9.9 ms, 157 rounds, 32 MB | **446x / 79x** |
| 65,536 | 0.84 ms, 2 rounds, 2.6 MB | 172 ms, 1,024 rounds, **1.35 GB** | **205x / 513x** |
| 262,144 | 0.62 ms, 2 rounds, 10 MB | **1.73 s**, 4,096 rounds, **21.5 GB** | **2,770x / 2,049x** |

Cloned-entry counts tracked `N^2/(2*cap)` exactly, so the quadratic holds analytically as well as empirically. This is reachable at exactly the workload the PR targets — the journal cap was raised because 65,536 was being exceeded, and at 65,536 the new path spends 172 ms and churns 1.35 GB while repeatedly taking `mutationMutex`, the same lock every admission and removal needs.

The good news is the fix looks small: only `delta[:applyCount]` is ever read, so cloning the remainder is provably dead work. Details inline.

Verification on my side:

| Check | Result |
|---|---|
| `go test -race -count=1 ./...` | pass, exit 0, 66 pkgs, 0 races |
| `make docs-parity` / `import-boundaries` | pass |
| CI (all 9 checks) | pass |

Two smaller things: the catch-up condition is now completely silent (it used to reach the caller's error log), and it'd be worth stating the intended memory ceiling for the enlarged journal. Also, per `CLAUDE.md` a mempool change like this warrants a DevNet run — the PR reports `go test -race ./mempool` plus the degenerate benchmarks, but no DevNet result; and the doc-status convention asks for an explicit `DATABASE.md` verdict alongside the `ARCHITECTURE.md` update.

mempool/mempool.go
```diff
@@ -1102,11 +1112,18 @@ func (m *Mempool) rebuildOverlayAttempt() ([]event.Event, error) {
 			}
 			m.mutationMutex.Unlock()
 
+			if requiredRounds := (len(delta)+m.revalidationDeltaCap-1)/m.revalidationDeltaCap +
```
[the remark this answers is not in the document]
wolf31o2: Correct, fixed in dc041a8b. `requiredRounds` was a total bound computed without the rounds already spent, so a backlog arriving late in an already-enlarged budget produced a total no larger than the one in effect: the budget did not grow, the remaining rounds could not cover the work, and the loop bailed.

Applied your suggestion, but extracted the arithmetic into a `catchupBudget(round, pending, deltaCap)` helper so it is directly testable rather than inline in the loop. `TestCatchupBudgetAccountsForRoundsAlreadySpent` pins it, and the regression surfaces exactly as you described -- without the `round` term the assertion fails with "173 is not greater than 173", i.e. the recomputed budget equals the one already in effect.

mempool/mempool.go
```diff
@@ -1004,7 +1013,8 @@ func (m *Mempool) rebuildOverlayAttempt() ([]event.Event, error) {
 		}
 
 		appliedSeq := startSeq
-		for range maxRevalidationCatchupRounds {
+		catchupRounds := maxRevalidationCatchupRounds
+		for round := 0; round < catchupRounds; round++ {
```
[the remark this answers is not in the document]
wolf31o2: Both fixed in dc041a8b.

Round-relative budget: `catchupBudget(round, pending, deltaCap)` now includes `round`, extracted from the loop so it is unit-testable. Without the term the test fails with "173 is not greater than 173".

Bounded window: `mutationsAfter` is replaced by `mutationWindow(journal, seq, limit)`, which binary searches the start (journal seqs increase monotonically, since `recordMutationLocked` appends with an incrementing seq), clones only the next `limit` entries, and returns the pending count the budget calculation needs. So the per-round work under `mutationMutex` is bounded by `revalidationDeltaCap` rather than by the journal length -- which mattered more after the budget scaling, since that multiplies the number of rounds. It also never returns nil, so callers need not distinguish empty from nil.

On the regression case you asked for (a second backlog queued after most of an enlarged budget is consumed): I pinned it through `catchupBudget` directly rather than orchestrating a second backlog against a live revalidation. Driving the loop to a specific round from outside means racing the blocking validator against journal injection, and the resulting test would be timing-dependent in a way that pins the schedule rather than the arithmetic. The helper test covers the same defect deterministically. `TestMempool_RevalidationConvergesOnBacklogLargerThanRoundBudget` still covers the end-to-end path.

mempool/mempool.go
```diff
@@ -1104,9 +1113,9 @@ func (m *Mempool) rebuildOverlayAttempt() ([]event.Event, error) {
 
 			applyCount := len(delta)
 			if applyCount > m.revalidationDeltaCap {
-				// Drain the excess off-lock, leaving at most the configured
-				// residual for the next catch-up/finalization pass.
-				applyCount -= m.revalidationDeltaCap
+				// Bound the amount of replay work per round so that the
+				// catch-up loop can observe and reconcile new mutations.
+				applyCount = m.revalidationDeltaCap
```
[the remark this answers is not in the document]
wolf31o2: Fixed in 7496f158, and I should have replied here when I pushed it rather than only in the commit -- apologies for the dangling thread.

Your analysis was right: with a fixed round count the total budget was `revalidationDeltaCap x rounds` (1024 by default), every attempt restarts from a fresh journal so no progress carried across, and any sustained backlog past that ceiling made every `rebuildOverlay` bail without publishing -- no invalid-transaction removal, no DAG rebuild -- which is the regime the 16x larger journal cap exists to buffer.

The round budget now scales with the observed backlog while each round stays bounded. `TestMempool_RevalidationConvergesOnBacklogLargerThanRoundBudget` queues a backlog past the fixed budget and asserts the invalid transaction is still dropped; it fails without the scaling. Two follow-up findings on that fix (the missing `round` term, and the unbounded per-round journal clone) are addressed in dc041a8b.

mempool/mempool.go
```diff
@@ -1102,13 +1115,18 @@ func (m *Mempool) rebuildOverlayAttempt() ([]event.Event, error) {
...
-				// residual for the next catch-up/finalization pass.
-				applyCount -= m.revalidationDeltaCap
+			// Scale the total budget to the observed backlog while keeping
+			// each replay round bounded. New mutations can extend this budget
+			// again until the journal overflows.
+			if requiredRounds := catchupBudget(
+				round, pending, m.revalidationDeltaCap,
+			); requiredRounds > catchupRounds {
+				catchupRounds = requiredRounds
 			}
```
chrisguiney, line 1144: Non-blocking, but worth pinning down because it is the load-bearing invariant for how this loop terminates.

`catchupBudget` returns `round + ceil(pending/deltaCap) + 16`, so whenever `pending >= 1` it is at least `round + 17`. Since `catchupRounds` only ever grows, the loop condition `round < catchupRounds` can never fail while a backlog exists. I checked this over 100,000 rounds against a range of `pending` values and the headroom never dropped below 17.

So the round budget does not bound anything: the loop exits only on `pending == 0`, journal overflow, `ErrMempoolStopped`, or a replay error. `ARCHITECTURE.md` already says as much ("The journal cap remains the ultimate bound"), so I read this as deliberate rather than accidental — the note is that the variable named a budget isn't one, which is easy to misread later.

Simulating sustained arrival (drain 64/round) to see what actually ends it:

| Arrival | Rounds | Outcome |
|---|---|---|
| 0/round | 16 | converged |
| 32/round | 32,737 | journal overflow, candidate abandoned |
| 64/round | 16,369 | journal overflow, candidate abandoned |
| 128/round | 8,185 | journal overflow, candidate abandoned |

The 32/round row is the interesting one: arrivals are *below* the drain rate, yet it still never finalises. Once the backlog falls under the per-round cap, each round drains exactly what is pending and the next batch lands before the following sample, so `pending` stabilises at the arrival rate and never hits zero. Real arrivals are bursty rather than a constant rate, so a live node will find quiet instants that this model never does — I would not read the table as a prediction. It does illustrate that finalisation depends on sampling a genuinely empty journal, which gets harder as rounds get shorter.

Nothing here needs changing for correctness. Two small things that would help the next reader:

- A line on `catchupBudget` noting it always exceeds `round` when `pending > 0`, so journal overflow is the real terminator.
- `TestCatchupBudgetAccountsForRoundsAlreadySpent` pins the arithmetic well, but not this property. A one-line assertion — `catchupBudget(r, 1, cap) > r` for a few `r` — would pin the invariant the loop actually depends on.

mempool/mempool.go
```diff
@@ -950,6 +955,11 @@ func (m *Mempool) rebuildOverlay() error {
 		if errors.Is(err, errValidationSnapshotChanged) && attempt == 0 {
 			continue
 		}
+		if errors.Is(err, errRevalidationCatchup) {
+			// The live pool is unchanged. A later chain update will retry
+			// revalidation after mutation pressure subsides.
+			return nil
+		}
```
chrisguiney, line 981: Non-blocking, and related to the note on the budget calculation.

Because the budget always exceeds the current round while `pending > 0`, this error can no longer be produced by mutation pressure. The only remaining route is the `continue` at line 1054 — the finalize race where `m.mutationSeq != liveSeq` — which consumes a round *without* extending the budget. It therefore takes roughly 17 consecutive finalize races to reach `errRevalidationCatchup`.

That makes the comment here ("after mutation pressure subsides") and the error text itself ("could not catch up with mutations") describe a cause that isn't the one that fires. Under genuine mutation pressure you get `errRevalidationJournalOverflow` instead, which propagates and is logged by the caller at line 914 — so the visible path is fine, which is why this isn't blocking.

Rewording both to say what actually triggers it — repeated races against the live pool during finalisation — would save the next person the trace I just did. Returning `nil` here is still the right call either way.

mempool/mempool.go
```diff
@@ -46,7 +47,7 @@ const (
 	DefaultCleanupInterval        = 1 * time.Minute
 	DefaultRevalidationDeltaCap   = 64
 	DefaultConsumerCacheSize      = 1024
-	defaultRevalidationJournalCap = 65_536
+	defaultRevalidationJournalCap = 1 << 20
```
chrisguiney, line 55: Non-blocking, carried over from last round since it didn't get picked up.

`unsafe.Sizeof(mempoolMutation{})` is 40 bytes, so `1<<20` is ~42 MB for the backing slice alone. Each entry also holds an `*appliedTx`, a `*MempoolTransaction`, and a `removed map[string]struct{}`, and those pointers pin transactions that have already left the pool for the duration of the pass.

The measurements above make this a much easier number to defend than it was before — a full drain is now 5.3 ms and 43 MB rather than 33.5 s and 335 GB, so the 16x raise is well earned. A short comment recording the intended worst-case footprint would just make that reasoning visible at the constant itself, rather than only in the PR that changed it.

mempool/mempool.go
```diff
@@ -1137,18 +1153,51 @@ func (m *Mempool) rebuildOverlayAttempt() ([]event.Event, error) {
...
+	})
+	pending = len(journal) - idx
+	end := min(idx+limit, len(journal))
+	window = slices.Clone(journal[idx:end])
+	if window == nil {
+		// slices.Clone yields nil for an empty result. Return an empty slice
+		// instead so the window is never nil, which keeps callers (and
+		// nilaway) from having to distinguish the two.
+		window = []mempoolMutation{}
 	}
```
chrisguiney, line 1204: Trivial, take it or leave it. This branch looks unreachable: when `pending > 0` we have `idx < len(journal)`, so `end > idx` and the clone is non-empty; and when `pending == 0` the caller takes the finalize branch at line 1027 and never ranges the window.

The comment says it keeps callers and nilaway from distinguishing the two, which is a fair reason to keep it as a guarantee about the function's contract rather than about today's single caller — in which case ignore this entirely. I mainly wanted to note that no current path exercises it, so it won't be covered by anything if the contract later changes.

chrisguiney (changes requested):
Thanks for the rework, and for the thorough commit-by-commit iteration on it. The blocking finding from my last round is fixed, and fixed decisively — I re-ran the same measurement driving both versions through a full drain at `deltaCap = 64`:

| Backlog | Flagged version | Current |
|---|---|---|
| 10,000 | 786,256 cloned, 8.0 ms, 31 MB | 10,000 cloned, 0.22 ms, 0.4 MB |
| 65,536 | 33,587,200 cloned, 180 ms, 1.32 GB | 65,536 cloned, 0.47 ms, 2.7 MB |
| 262,144 | 537,001,984 cloned, 1.90 s, 21 GB | 262,144 cloned, 0.96 ms, 10.7 MB |
| **1,048,576** (the new cap) | **8,590,458,880 cloned, 33.5 s, 335 GB** | **1,048,576 cloned, 5.3 ms, 43 MB** |

Cloned entries now equal the backlog *exactly* at every size — one copy per journal entry — so it is provably linear rather than `N^2/(2*cap)`. That result also retroactively justifies the `1<<20` cap, which would have been unusable under the old drain. The binary search on top is a nice extra.

Two things I want to credit specifically: extracting `catchupBudget` as a testable helper is a better answer than the inline arithmetic I suggested, and `TestMutationWindowClonesOnlyTheBoundedWindow` guards the regression more directly than the benchmark I asked for — asserting `len(window) == 10` against a 1,000-entry journal fails immediately if anyone reverts to cloning the suffix.

**What I am holding for is the DevNet run, and that is the only gate.** `CLAUDE.md` names mempool and tx-submission changes as warranting it, and this change alters lock-hold granularity and loop-termination behaviour under sustained load — exactly what DevNet exercises and what `go test -race ./mempool` structurally cannot. The Testing section still lists only the unit tests and the degenerate benchmarks. The same convention also asks for an explicit `DATABASE.md` verdict alongside the `ARCHITECTURE.md` update, even when the answer is "not affected". I raised both last round; I suspect they were missed rather than declined, since your replies landed on the cubic and CodeRabbit threads and my four never got picked up.

I want to be clear that I found **no correctness defect**. Lock/unlock pairing is correct on every path including the `continue` at line 1054, lock ordering (`mutationMutex` -> `RLock`/`Lock` -> `consumersMutex`) is consistent with the rest of the file, and `finishJournal()` still runs before the error return. Everything inline below is non-blocking.

I am also explicitly withdrawing my observability finding from last round as a blocker. Its premise was that a too-small budget would silently strand the pool; the unbounded scaling largely removes that, and the failure that does occur — journal overflow — propagates and is logged at `mempool.go:914`. A counter alongside `txsEvicted`/`txsExpired` would still be worth having, but I was wrong to keep treating it as a gate once the budget stopped being the limiting factor.

Verification on my side:

| Check | Result |
|---|---|
| `go test -race -count=1 ./...` | pass, exit 0, 0 races |
| `modernize ./mempool/...` | clean |
| `nilaway ./mempool/...` | 3 findings, identical count on `main` — none introduced |
| `make docs-parity` / `make import-boundaries` | pass |
| CI on dc041a8b (9 checks) | all pass |

Happy to approve as soon as the DevNet result is reported.

mempool/mempool.go
wolf31o2: Confirmed and documented in e4783689. Your reading is right: with pending non-zero the budget always exceeds the current round by at least maxRevalidationCatchupRounds, so the round comparison cannot end the loop, and the journal cap is the real bound. That was deliberate but only stated in ARCHITECTURE.md, not where someone reading the loop would find it, so the note now sits on catchupBudget itself.

Added TestCatchupBudgetAlwaysExceedsCurrentRound, which asserts the headroom across a matrix of rounds, backlogs and deltaCaps rather than the single arithmetic case, and separately pins that the budget stops growing at pending zero, which is what allows the loop to finish. Verified it discriminates: making catchupBudget return round outright fails it.

The 32/round row in your table is a good catch, and I agree with your reading that it is a property of the constant-rate model rather than a prediction for a live node.

mempool/mempool.go
wolf31o2: Agreed, the comment and the error text were both describing mutation pressure, which is now the one cause that cannot produce this. Reworded in e4783689.

The error is now "mempool: revalidation lost repeated races with the live pool", with a doc comment saying that each finalize race consumes a round without extending the budget, that enough consecutive races exhaust it, and that sustained pressure ends in errRevalidationJournalOverflow instead. The handler comment says the same. Returning nil is unchanged.

mempool/mempool.go
wolf31o2: Fair, and carried over correctly: it did get missed last round. Recorded at the constant in e4783689: 40 bytes per mempoolMutation, so roughly 42 MiB for the backing slice at the 1<<20 cap, plus the transactions each entry pins for the duration of the pass, and the 5.3 ms full-drain measurement.

mempool/mempool.go
wolf31o2: Right on both counts: unreachable today, and kept as a contract rather than a statement about the current caller. Rather than ignore it I added TestMutationWindowReturnsEmptyNotNil, which calls mutationWindow directly with a caught-up sequence and with an empty journal, so the contract is covered even though no production path exercises it. Confirmed it fails if the guard is removed.

One related note: an earlier draft of that test passed a literal nil journal, which made nilaway flag the slice expression inside mutationWindow. It takes an empty slice instead. Unrelated to this thread, the branch also carries three pre-existing nilaway findings against AddConsumer results in tests, where main has none; nilaway cannot see through require.NotNil, and converting those to explicit nil checks did not help because it loses element nilness on a slice of pointers, so I left them rather than churn the tests.

mempool/mempool.go
```diff
@@ -931,6 +937,15 @@ var errRevalidationJournalOverflow = errors.New(
 	"mempool: revalidation mutation journal overflow",
 )
 
+// errRevalidationCatchup reports repeated races against the live pool during
+// finalisation: each race consumes a round without extending the budget, so
+// enough consecutive ones exhaust it. Sustained mutation pressure does not
+// reach here, because the budget grows with the backlog; that path ends in
+// errRevalidationJournalOverflow instead.
```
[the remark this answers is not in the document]
wolf31o2: Valid, and my previous comment was wrong. Fixed in 33e9898b.

Verified against the code rather than the reasoning I had inherited: recordMutationLocked bumps mutationSeq and appends to the journal together whenever the journal is active and under cap. So the finalise race at `mutationSeq != liveSeq` does imply a journal entry, the next round observes pending > 0, and catchupBudget extends past the current round. The one case that bumps the sequence without appending is a full journal, and that sets journalOverflow, which the loop tests before anything else and turns into errRevalidationJournalOverflow.

The error is now described as a defensive guard with that reasoning recorded at its declaration, the handler comment says the same, and the message is "mempool: revalidation exhausted its catch-up budget", which states the condition instead of a cause that cannot occur. The sibling note on catchupBudget now cross-references it.

Worth noting for provenance: TestMempool_RevalidationConvergesOnBacklogLargerThanRoundBudget describes this error firing on a large backlog, but that is the pre-fix behaviour the test exists to prevent, not a currently reachable path.

mempool/mempool.go
wolf31o2: Correction to my reply above, which I got wrong.

I said the reworded error described "repeated finalize races". cubic pushed back on the follow-up commit and is right: that path cannot fire either. recordMutationLocked bumps mutationSeq and appends to the journal in the same step while the journal is active and under cap, so a finalise race implies a journal entry, and the next round therefore sees pending > 0 and extends the budget. The only sequence bump without an append happens when the journal is full, which sets journalOverflow and exits through errRevalidationJournalOverflow instead.

So errRevalidationCatchup is unreachable, not merely hard to reach via ~17 consecutive races. 33e9898b reframes it as a defensive guard, records that reasoning at the declaration, and renames the message to "mempool: revalidation exhausted its catch-up budget". Your underlying point stands unchanged and is now stated more strongly: the journal cap is the terminator.

chrisguiney (approved):

PR #3117, opened 2026-08-05
fix(leios): commit endorser-block blob in its own transaction

wolf31o2:
## What

Commit each Leios endorser-block blob in its own blob transaction instead of threading the shared block-processing transaction, and seed the Tier-2 block cache with the committed bytes.

## Why

The endorser-block apply path (`ledger/leios_apply.go`) wrote every certified endorser block's full CBOR blob, plus one `DOFF`-offset entry per endorser transaction and per produced output, into the shared block-processing transaction that covers up to a 50-block chunk. Badger routes values at or above the 1 MiB value threshold to the value log where they barely count toward the transaction budget, but the sub-threshold endorser-block blobs are counted in full. On a dense Leios backlog the accumulated writes across the chunk's certified endorser blocks exceed Badger's per-transaction budget (`maxBatchSize`, 15% of the 128 MiB memtable, about 19.2 MiB), and the next `Set` fails with `ErrTxnTooBig`. The chunk can then never commit: the endorser block reports as unavailable and the pipeline restarts on the same chunk indefinitely.

Observed on a musashi from-genesis node, which wedged permanently at eb_slot 3017634 (366 store failures, 3392 pipeline restarts) after the shared chunk transaction had already applied about 16 certified endorser blocks (roughly 4000 endorser transactions).

The endorser-block blob is content-addressed by `(slot, hash)`, carries `ID=0`, is off the chain index, and is idempotent, so committing it independently is safe. The endorser transactions' `DOFF` ledger effects still go into the shared transaction under the ranking block's point, so a rollback of the ranking block removes them, and a blob orphaned by a crash or rollback is harmless and overwritten identically on re-apply. A separately committed blob is not visible to the shared transaction's start-of-transaction snapshot, so the apply path seeds the Tier-2 block cache (`TieredCborCache.SeedBlockCbor`) with the blob bytes, letting an intra-batch validation read on the CIP path resolve the endorser-block CBOR from cache.

## Validation

- `go build`, `go vet`, `golangci-lint` (0 issues), `make import-boundaries`, `make docs-parity` all pass.
- `go test -race` for `ledger/...` (including `ledger/leios`), `database/...`, and `internal/test/conformance` pass. New `TestTieredCborCacheSeedBlockCbor`.
- musashi from-genesis resumed at the wedged DB (pinned just below eb_slot 3017634): the previously-unstorable endorser block now applies (`slot=3017699 eb_slot=3017634 eb_txs=253`), and the node advances past the wedge point through 3043369+ with 0 `Txn is too big` errors and 0 pipeline restarts.

## Docs

`DATABASE.md` updated (endorser-block blob commit semantics and the Tier-2 cache seed). `ARCHITECTURE.md` checked, not affected.

---
## Summary by cubic
Commit Leios endorser-block CBOR blobs in their own transaction only on the no-validation path to prevent Badger `ErrTxnTooBig` wedges, while keeping validation semantics intact on the CIP path.

- **Bug Fixes**
  - On `LeiosApplyEndorserBlockTxs=false` (Musashi/Haskell path), write the endorser-block blob in its own blob txn; `DOFF` ledger effects stay in the shared txn; rollback behavior unchanged.
  - On `LeiosApplyEndorserBlockTxs=true` (CIP path), keep the blob in the shared block-processing txn so intra-chunk reads use read-your-writes.
  - Removed Tier-2 cache seeding; documented per-path behavior in `DATABASE.md` and trimmed the in-code comment to defer to docs.

<sup>Written for commit dfa14d97b4c07797d3655f198e875354c9b404cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented oversized transactions when storing endorser-block data.
  * Improved same-batch validation by making newly stored block data immediately available.
  * Preserved existing rollback behavior while allowing safely reusable orphaned content.

* **Performance**
  * Added caching for recently stored block data to support faster subsequent resolution.


wolf31o2:
Addressed the cubic findings in f2290a1e.

P2 (evictable Tier-2 seed reintroduces the wedge on the CIP path): the own-transaction commit is now gated to the Haskell-conformant path (LeiosApplyEndorserBlockTxs false), which applies without validation and never reads the blob back within the chunk. The CIP path keeps the blob in the shared block-processing transaction, so an intra-chunk read of an endorser-block-produced output resolves via read-your-writes during validation, with no dependence on a best-effort cache entry. This restores the pre-existing CIP behavior; the observed wedge was on the no-validation path.

P3 (SeedBlockCbor aliased the caller's slice): SeedBlockCbor is removed, so the footgun is gone.

Musashi from-genesis behavior is unchanged (owned transaction), so the end-to-end validation still holds: the previously unstorable endorser block at eb_slot 3017634 applies and the node advances past the wedge with 0 ErrTxnTooBig and 0 pipeline restarts.

arepala-uml (approved):

PR #3131, opened 2026-08-06
feat(database): persist and enforce critical node settings on startup

### LINKED ISSUE #3127: Persist and enforce critical node settings on startup
Persist the settings that determine what is already in the database, enforce them as a hard compatibility gate on startup, and let persisted values supply defaults so a node resumes the configuration it was created with.

## Current state

`database/types.NodeSettings` persists `StorageMode` and `Network`, and `checkNodeSettings` in `database/commit_timestamp.go` compares them on every open. Gaps:

- Network validation is skipped entirely when the configured network is empty, which is the devnet and custom-network case where the protocol magic is the only identity available.
- `config/cardano/node.go` validates each genesis file against its declared hash, but nothing pins a database to the hashes it was built from. A self-consistent config with a different systemStart, epochLength, slotLength, or securityParam silently invalidates every slot and epoch number already stored.
- History expiry has no persisted state. `database/prune.go` writes per-block markers, so there is no single fact to enforce against.
- Persisted settings are compared but never consulted as defaults. `Network` defaults to `preview` and `StorageMode` to `core`, so a bare `dingo` against a preprod database in api mode mismatches on network and would prune to core.

## Scope

Add a `node_settings_gate` table and a gate registry with four enforcement classes:

- frozen: any difference is fatal
- frozen fill-once: an empty configured value is preserved rather than compared, so a later dingo can fill a hash an earlier one did not know
- latch: transitions permitted in one declared direction only, recording the epoch and slot
- taint: a sticky bit that may not be set on a database lacking it, and that a tightened configuration cannot clear

Gates: network, network magic, start era, storage mode, history expiry, the CIP-50 and CIP-0163 ledger gates, min pool margin, the two validation taints, the five era genesis hashes, the blob and metadata plugin names, and a blob store identity UUID.

Storage mode becomes a one-way api to core latch. Genesis hashes, plugin names, and the store id are validate-only and never override.

Enforcement splits in two: `database.New` checks what a bare open can know, and a new `EnforceNodeSettings` checks the genesis hashes and ledger gates from node startup after the cardano config is parsed and before the ledger applies its first block.

Persisted values become the last step of config resolution, after CLI flags. Provenance is tracked for gated fields only, so an operator-supplied value that conflicts is fatal while a built-in default is overridden. A `dingo.dbinfo` sidecar records the metadata plugin name so a bare `dingo` in a postgres-provisioned directory reports the mismatch instead of starting a fresh sqlite sync.

`lifecycle.Manifest` carries the gate values so a restore refuses an incompatible target before opening the store.

## Behavior change

`dingo -n preprod`, stop, then bare `dingo` resumes preprod. `dingo -n preview` against that database fails, naming the gate, both values, and the source of the configured one.

There is no override or force flag. A frozen mismatch means resync.

## Notes

`database/storage_mode_test.go` has `TestNodeSettingsRejectStorageModeChange`. If it asserts the api to core direction it will fail once that direction becomes a permitted latch, and it needs a decision rather than an edit.

Two assumptions in the design worth confirming before implementation: YAML counts as explicit, so a YAML key conflicting with a persisted gate is fatal rather than overridden; and relaxing validation on a database that already has a strictly validated range is fatal.

Docs: DATABASE.md for the table and enforcement contract, ARCHITECTURE.md for the two-phase split and config resolution order.

wolf31o2:
Persists the settings that determine what is already in a node's database and enforces them on every start, so an incompatible configuration fails at startup instead of silently diverging.

## What this adds

`database/nodesettings` holds the policy in one pure decision function over a 19-gate registry with five enforcement classes: frozen, frozen fill-once, one-way enum latch, one-way boolean latch, and sticky taint.

Gate values are stored in a new `node_settings_gate` table via metadata migration v2 for sqlite, postgres and mysql. The legacy `node_settings` row becomes a read-only compatibility fallback, since its only insert is a no-op on conflict and no query can update `storage_mode` in place.

Enforcement runs in two phases, split by what each stage can know. Phase 1 in `database.New` covers gates whose unset state is distinguishable from a real value: storage mode, network, network magic, start era, plugin selection, and a blob-store identity UUID. Phase 2 at node startup covers the five era genesis hashes, the CIP-50 and CIP-0163 ledger gates, and the three boolean-derived gates, because a bool cannot express that a caller does not know its value and both `mithril/sync.go` and `database/lifecycle/restore.go` construct a partial `database.Config`. Phase 2 is deferred until after commit-timestamp recovery completes.

`internal/config` records provenance for gated fields and `internal/settingsresolve` applies persisted values before network expansion, so persisted values supply defaults for anything the operator did not set explicitly.

`database/dbinfo` writes a `dingo.dbinfo` sidecar recording the metadata plugin name, checked before any store is opened so a mismatch is reported rather than starting a fresh sync.

Snapshot manifests carry the gate values so an incompatible offline restore is refused before the target directory is touched. The manifest format version is unchanged because the field is additive and optional and `ParseManifest` ignores unknown fields.

## Behavior changes

`dingo -n preprod`, stop, then a bare `dingo` resumes preprod. An explicitly conflicting `-n preview` is fatal and names which source supplied the conflicting value.

`storage_mode` is a one-way `api` to `core` latch. `core` to `api` remains fatal.

Relaxing `validateHistorical` or `strictUtxoValidation` on a database that recorded strict validation is rejected.

There is no override flag. A frozen-gate mismatch requires a resync.

## Docs

DATABASE.md and ARCHITECTURE.md are updated.

Fixes #3127

---
## Summary by cubic
Persist and enforce database-defining node settings so incompatible configs fail fast at startup. Midnight indexer is now opt‑in via `midnight.enabled`; storage mode is aligned across preflight and runtime, and initialization is atomic to avoid partial state.

- **New Features**
  - `database/nodesettings`: 19‑gate policy persisted in `node_settings_gate` (schema v1) with frozen/fill‑once/latch/taint classes.
  - Two‑phase enforcement: Phase 1 in `database.New` (storage mode, network/magic, start era, plugins, blob‑store ID); Phase 2 at node start (genesis hashes, CIP‑50/CIP‑0163 and derived booleans) after commit‑timestamp recovery. Phase 1 re‑runs post‑recovery.
  - Persisted gates provide defaults via `internal/settingsresolve`; explicit conflicts are fatal and name their source; malformed carried values are rejected. `database/dbinfo` writes and verifies a `dingo.dbinfo` sidecar before opening any store; incomplete sidecars are fatal.
  - Snapshots include gate values and validate with the same policy before restore (one‑way latches allowed); manifest format remains v1.
  - Midnight indexer requires `midnight.enabled` and API mode; Midnight network defaults re‑apply on resume. Blob‑store identity writes are synced for durability.
  - `cmd/dingo` preflight uses the effective storage mode that matches `node.Run` (dev‑mode override applied), and normalizes an unset mode to core. Errors no longer auto‑dump full usage.

- **Migration**
  - `node_settings_gate` ships in schema v1; existing databases migrate as usual. A bare `dingo` resumes the persisted network/storage mode. `storage_mode` is a one‑way latch (api→core).
  - Breaking: the Midnight indexer no longer starts implicitly; set `midnight.enabled: true` (or `--midnight-enabled`) in API mode to run it.

<sup>Written for commit 8f7abbfc889e7776080f1c8f0642e671912eab02. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3131?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Persisted node settings now resume automatically when reopening an existing database.
  * Added validation for network, storage, plugin, ledger, blob-store, and genesis settings.
  * Database snapshots now retain node-setting information for compatibility checks.
  * Configuration errors identify conflicting settings and provide clearer guidance.
  * Database metadata can identify its configured metadata provider.

* **Bug Fixes**
  * Improved handling of missing, empty, inaccessible, or mismatched database directories.
  * Node-setting enforcement is safely deferred during required database recovery.
  * Runtime errors no longer automatically print command usage.


node.go
```diff
@@ -795,8 +842,11 @@ func (n *Node) Run(ctx context.Context) error {
 	// (b) the EventBus subscription exists before any BlockActionApply events
 	// can be emitted, eliminating the startup gap identified in #2114. The
 	// epoch cache is loaded first because Midnight backfill writes epoch-keyed
-	// Ariadne/candidate rows.
-	if n.config.storageMode.IsAPI() {
+	// Ariadne/candidate rows. Both the explicit opt-in and API storage mode
+	// are required: the indexer depends on the api-mode indexes to function,
+	// and storage mode alone is no longer sufficient to start it (an api-mode
+	// deployment may not want Midnight indexing at all).
+	if n.config.midnight.Enabled && n.config.storageMode.IsAPI() {
```
chrisguiney, line 849: This is a breaking default change for existing operators, and I don't think it belongs to #3127.

Today any api-mode node runs the Midnight indexer implicitly. After this, `MidnightConfig.Enabled` defaults to `false`, so those deployments silently stop indexing until they set `midnight.enabled: true`. That's a reasonable change to want — an api-mode deployment may well not want Midnight — but two things give me pause:

1. #3127 never mentions Midnight, so this is riding along with an otherwise tightly-scoped settings-gate change.
2. It isn't in this PR's **Behavior changes** section. It appears only in the generated cubic summary, which means anyone drafting release notes from the description will miss it — and the failure mode for an operator is silent: no error, indexing just stops.

Could this move to its own PR? If you'd rather keep it here, adding it to Behavior changes with the `midnight.enabled` migration step would be enough for me to drop this.

(If there's a dependency I've missed — something in the gate/provenance work that actually needs the explicit flag — I'd genuinely like to know, since I couldn't find one.)

internal/settingsresolve/settingsresolve.go
```diff
@@ -0,0 +1,616 @@
...
+				cfg.PledgeLeverageEnabled = enabled
+				if carried == "" {
+					return
+				}
+				// strconv.IntSize, not 64: the destination is uint,
+				// which is 32 bits on a 32-bit platform, so parsing
+				// at 64 would silently truncate instead of erroring.
+				if parsed, err := strconv.ParseUint(
+					carried, 10, strconv.IntSize,
+				); err == nil {
```
chrisguiney, line 488: Small robustness inconsistency: a parse failure here is swallowed, leaving the gate half-applied — `PledgeLeverageEnabled` gets set to `true` while `PledgeLeverage` silently keeps its built-in default. Same pattern at lines 513 (`DelegatorInactivity`) and 537 (`MinPoolMargin`).

It's unreachable today, since dingo writes these values itself via `FormatUint`, so this is defence-in-depth rather than a live bug. But these three are consensus-affecting parameters, and a silently half-restored value is exactly the outcome this PR exists to prevent everywhere else — a malformed carried value should probably be as loud as a gate mismatch is. Returning an error (or at minimum a `slog.Warn`) would match the intent of the surrounding code.

Unrelated, but the `strconv.IntSize`-not-64 reasoning three lines up is a nice catch — that would have been an easy silent truncation on 32-bit.

chrisguiney (dismissed):
Really nice work — the comment density here is unusually high and it's earning its keep. The ordering rationale in `persistedGateValues`, the `InsertNodeSettingsGatesIfAbsent` race handling for the shared postgres/mysql case, the `Blob().Sync()` before `blob_store_id` latches as a Frozen gate, and the atomic sidecar write (`CreateTemp` → `Sync` → `Rename` → dir fsync) are all correct and non-obvious. The write-then-read-back verification after `writeGateValues` is a good instinct given the legacy row's silent-no-op history.

I'm requesting changes on two narrow items — one of them mostly a documentation fix. Happy to be talked out of the first if you read it as in scope.

## What I verified locally

- `go test -race ./...` on `8f402df0`: passes, 0 failures.
- Merged current `upstream/main` locally: clean, no conflicts. `v2/sqlite/expand.sql` is removed by #3136 and the registry collapses to v1 only, so `node_settings_gate` living in v1 lands consistently. Full `-race` suite passes again post-merge (66 packages, 0 failures).
- I initially suspected the DATABASE.md `v2alpha1` prose would survive the merge and contradict #3136 — it doesn't, the merge resolves those regions cleanly to 0 references. Noting it only so you know it was checked rather than assumed.
- `core` → `api` is still fatal (`storage_mode_test.go:60`), and the `api` → `core` direction was updated with a reasoned decision rather than a silent edit — which is what #3127 explicitly asked for.
- All new goroutines are confined to test files; no production concurrency added. Gate evaluation is startup-only, so no hot-path cost.
- The three LatchBool assembly sites (`node.go`, `internal/dblifecycle`, `internal/settingsresolve`) agree field-for-field. I checked each one, since drift there would produce spurious startup failures on resume.

## Two process notes, no code change needed

**The green checks are stale.** `go-test` and `golangci-lint` last ran against `e1636bb4`; head is `8f402df0`, five commits later. Only CodeQL is queued on the current head, so the checkmarks in the UI aren't covering the commits that added the atomic-initialization and deferred-enforcement work. Worth re-triggering before merge — my local runs above cover it in the meantime.

**PR description drift.** The body says gate values are stored "via metadata migration v2", but `c7bdf6d6` moved them into v1. Worth fixing alongside the Behavior-changes note below, since this description is what release notes get written from.

node.go
wolf31o2: No. This is fine. Nobody is using the midnight indexer.

arepala-uml (approved):

PR #3137, opened 2026-08-06
fix(leios): complete a partially-fetched endorser block on re-offer

### LINKED ISSUE #2629: leios: complete a partially-fetched endorser block's tail on relay re-offer

chrisguiney:
Fixes #2629

## Problem

Near the live tip the relay diffuses an endorser block's transactions over several seconds, so `fetchLeiosEbTxsBatched` regularly runs out of served transactions before the block is whole. It returns the contiguous prefix with an error, and both callers — the leios-notify transactions-offer handler and the by-point backfill — discard that prefix and cache nothing. When the relay re-offers the same endorser block, dingo re-fetches it from scratch, so a fraction of the endorser blocks in an active window are never completed and their outputs are missing from the UTxO set near tip.

Enlarging the per-fetch tail budget is not the fix: the attempt holds the per-connection fetch slot while it waits, which starves other endorser blocks during a busy window.

## Change

A cached endorser block now keeps `partialTxs`: a `txCount`-long sparse slice whose nil entries are the transactions still missing. That is the same representation `leiosNeededBitmap` already turns into a request bitmap, so no new encoding was needed.

The seed/retain pair lives inside the shared fetch rather than in the offer handler, so the by-point backfill resumes across peer failover for the same reason:

- `seedLeiosPartialTxs` primes the fetch working slice from `txsRaw` + `partialTxs`, so the request bitmap naturally covers only the tail.
- A deferred `retainLeiosPartialTxs` merges the result back on every exit path (tail budget, per-attempt deadline, protocol error), so an attempt that stops short releases the connection while its progress survives.

The retained set only grows, so two connections that each fetch part of a block contribute to one union. It is carried across the redundant manifest stores every connection's block offer produces — extending the existing no-regress rule for a cached transaction set — and cleared once the set is complete.

`retainLeiosPartialTxs` publishes a copy of the cache entry rather than mutating in place: `lookupLeiosEndorserBlock` hands out the raw pointer and callers read it without the lock, so in-place mutation would race against that existing contract. `insertedAt` is preserved so retaining a partial cannot extend an endorser block's cache lifetime indefinitely.

The diffusion-window bound on a single fetch attempt is unchanged, and a block completed by a re-offer is stored and applied through the existing gate — both explicit non-goals in the issue.

The offer handler's fetch-failure debug line gains a `retained` field so the progress carried across attempts is visible in the field.

## Acceptance criteria

- [x] A partially fetched endorser block is retained with its missing-transaction bitmap instead of discarded.
- [x] A re-offer of the same endorser block fetches only the still-missing transactions and completes the cached entry.
- [x] A completed-by-re-offer endorser block is applied through the existing gate, with no change to the diffusion-window bound on a single fetch attempt.
- [x] A test drives a partial fetch followed by a re-offer that completes the block and asserts the assembled transactions are correct and in order.

## Testing

New `ouroboros/leios_partial_tail_test.go` drives a relay that has diffused only part of the block, then re-offers it complete. It asserts the partial is retained, that the resumed fetch requests only the missing indices, that the assembled transactions are correct and in index order, that a redundant manifest store does not drop the retained partial, and that overlapping attempts union.

The tests were verified to fail without the fix: stubbing `retainLeiosPartialTxs` to a no-op fails all four behavioral tests.

- `go test -race ./...` passes; `./internal/test/conformance/` passes.
- `-count=10 -race` on the leios fetch/store tests, no flakes.
- `golangci-lint` 0 issues, `nilaway` clean, `make import-boundaries` and `make docs-parity` pass.

DevNet was not run: this environment has no Docker, and DevNet has no Leios mode, so it would not exercise this path. The change is inert outside `EnableLeiosTxFetch`/leios run-mode. Real validation is a from-scratch Musashi sync, watching the new `retained` field climb across attempts instead of resetting to zero.

## Documentation

`ARCHITECTURE.md` updated — the Leios networking and backfill sections already documented this fetch flow. `DATABASE.md` checked and not affected: partials are in-memory only and never persisted; the `et` blob is still written solely for a complete transaction set.

---
## Summary by cubic
Retain and resume partially fetched Leios endorser block transactions across re-offers to stop full re-fetches, complete blocks near tip, and fix missing UTxOs. Incomplete entries keep their original cache TTL so they don’t linger; the timestamp refreshes only when the set is complete.

- **Bug Fixes**
  - Cache `partialTxs` as a txCount-long sparse slice; seed via `seedLeiosPartialTxs`, merge via `retainLeiosPartialTxs`, clear on completion.
  - Implemented in the shared fetch path (`fetchLeiosEbTxsBatched`) so it works for `leios-notify` offers and by-point backfill, and resumes across peer failover.
  - Union progress across connections and carry it across redundant manifest stores; preserve `insertedAt` while incomplete to bound cache TTL, refresh only when complete.
  - Release the per-connection fetch slot between attempts; on re-offer request only the missing tail. Request bitmap stays MSB-first.
  - Add `retained` to the fetch-failure debug log.
  - Tests cover retention, tail-only re-requests, union across attempts, manifest-only stores keeping partials, and TTL bounding. `ARCHITECTURE.md` updated; no DB changes (partials are in-memory only).

<sup>Written for commit 8fac8b68b7cc075fc6c7a6d1891718039ac3e29a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3137?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Improvements**
  - Improved reliability when retrieving transactions from Leios endorser blocks by preserving successfully received transactions across interrupted or incomplete fetches.
  - Subsequent retrieval attempts now request only missing transactions, reducing redundant data transfer and helping transaction sets complete more efficiently.
  - Partial transaction data is retained when block information is updated and cleared once retrieval is complete.
  - Enhanced handling of protocol interoperability, timeouts, decoding, and downstream transaction serving.

- **Documentation**
  - Updated architecture documentation to describe transaction persistence, recovery, and protocol behavior.


chrisguiney:
Good catch on the TTL — fixed in 8fac8b68.

The finding was valid, and it contradicted an invariant the change itself claimed: `retainLeiosPartialTxs` preserves `insertedAt` deliberately, but `storeLeiosEndorserBlock` rebuilds the cache entry with `insertedAt: time.Now()`, so carrying the partial across a manifest-only store let the retained transaction bodies ride a refreshed TTL.

The leak is specifically the just-under-TTL window. An entry aged *past* its TTL is already evicted by `pruneLeiosEndorserBlockCacheLocked` at the top of the next store, so it self-heals; the problem is that the relay re-offers each endorser block on every connection, so a trickle of offers keeps resetting the clock before pruning can ever see it. Previously that cost a manifest; with retention it costs the transaction bodies too.

The fix keeps the original timestamp while the block is still incomplete, bounding a never-completing block to one TTL, after which pruning evicts it and a later offer starts over from the manifest. A store that *completes* the set still takes the fresh timestamp — it has become a servable entry and earns the same lifetime as any other. That asymmetry is asserted in the test rather than left implicit.

`TestStoreLeiosEndorserBlockPartialDoesNotRefreshCacheTTL` ages an entry to two seconds short of expiry, takes three redundant manifest offers, and asserts both that the partial survives and that the lifetime did not restart. It fails without the fix (the re-offer moved expiry ~10 minutes out). `ARCHITECTURE.md` records the reasoning.

arepala-uml (approved):

PR #3159, opened 2026-08-10
feat(leios): switch to on-chain BLS keys and definite committee selection

### LINKED ISSUE #3148: Leios: Breaking changes for Musashi network water phase
We are going to respin the musashi network as we are going to start using the BLS keys that are registered to the stake pool params and also switch to the definite committee selection scheme.

See also [this cardano-blueprint page for details](https://cardano-scaling.github.io/cardano-blueprint/pr-preview/pr-67/consensus/leios-committee.html) and the changelog on the [leios-prototype implementation](https://github.com/input-output-hk/ouroboros-leios/blob/main/CHANGELOG.md#prototype-2026w32---2026-08-09)

Scope of this ticket is basically do recreate the diff from `prototype-2026w31` to [`prototype-2026w32`](https://github.com/input-output-hk/ouroboros-leios/releases/tag/prototype-2026w32)

arepala-uml:
## PR Summary

Recreates the upstream `ouroboros-leios` `prototype-2026w31` → `prototype-2026w32` diff ahead of the Musashi respin: the Leios voting committee is now built from real, on-chain registered BLS keys instead of a static config file and a fake derived key, and committee selection always uses the deterministic stake-coverage algorithm instead of "every pool votes."

## Changes made

**`database/plugin/metadata/sqlstore/migrations/v2/sqlite/expand.sql` (new)**
- Added `leios_key_public`/`leios_key_possession_proof` columns to `pool` and `pool_registration`.

**`database/plugin/metadata/sqlstore/migrations/registry.go` / `registry_test.go`**
- Registers the new migration as schema version 2 so existing databases get upgraded automatically.
- Also updates the test that asserted "only 1 migration exists."

**`database/models/pool.go`**
- Added `LeiosKeyPublic`/`LeiosKeyPossessionProof` fields to the `Pool` and `PoolRegistration` structs. It gives the rest of the codebase somewhere to read/write the new columns.

**`database/plugin/metadata/sqlstore/pool.go`**
- Wires the new columns into `ImportPool`, `GetPools`, and the row-scanning helpers. Also, genesis-imported and queried pools now carry their registered key end-to-end.

**`database/plugin/metadata/sqlstore/transaction_certificates.go`**
- Persists a pool's `leios_key` from a live registration certificate into the new columns. No trust check here — this layer only stores what's on-chain, same as `vrf_key_hash` always has.

**`ledgerstate/certstate.go` / `ledgerstate/import.go`**
- Carries the Leios key through the Mithril-snapshot import path as well. Previously the key was parsed just to compute a byte offset, then discarded — now it's kept.

**`ledger/leios/bls.go`**
- Added `VerifyLeiosKeyProofOfPossession` to actually check the proof a registered key comes with — gouroboros only confirms the bytes are the right length, not that the proof is genuine. It's the one place in the whole change where a key stops being "just something we saw on-chain" and becomes something we'd trust enough to verify a vote against.

**`ledger/view.go`**
- Added `GetLeiosKeys`, reading each pool's registered key back out of the database and this is how the vote manager gets access to what the write path stored.

**`node_leios.go`**
- Added `leiosKeyProviderAdapter` wiring on-chain keys into the vote manager; removes `PrototypeMode` and the local self-signing derivation fallback. A pool with no configured signing key now runs as a non-voting relay instead of deriving one.

**`ledger/leios/manager.go`**
- Once per epoch, the vote manager now fetches every pool's on-chain key, verifies it, and caches only the ones that pass — replacing the old `PrototypeMode` branch and the derivation fallback that used to paper over a missing key. A committee member with no verified key is simply a keyless seat: its stake still counts, but it can never be credited with a real signature.

**`ledger/leios/committee.go`**
- Removed `ComputePrototypeCommittee` ("every pool votes, ascending order") where as `ComputeCommittee` (stake-coverage, descending order) is now the only committee algorithm.

**`ledger/leios/keys.go`**
- Removed `DerivePrototypeVoteSigningKey`, the insecure pool-hash-derived key shortcut. Now, it matches upstream's own removal of this exact mechanism in `prototype-2026w32`.

**`ledger/leios/certificate.go`**
- `ValidatePrototypeEbCertificate` no longer derives a fallback key for unresolvable signers. It now returns `(sigChecked bool, err error)`, reporting "can't confirm" instead of faking a pass.

**`ARCHITECTURE.md` / `DATABASE.md`**
- Documented the new columns, the on-chain key-resolution flow, and the removed derivation fallback.

## Breaking changes addressed

- **Behavioral, not just additive**: any pool previously voting only via the derived fallback key (no config, no on-chain registration) will silently become a non-voting relay after this change — its votes are no longer fabricated a passing key.
- **Database schema**: adds migration v2 (additive, nullable columns) — existing databases upgrade automatically on next startup, no manual step needed.
- **Go API**: `ValidatePrototypeEbCertificate`'s return signature changed from `error` to `(sigChecked bool, err error)`, matching `ValidateEbCertificate`. No callers exist elsewhere in this repo today, but this is a breaking change for any external consumer of the package.
- **Removed public API**: `ComputePrototypeCommittee` and `DerivePrototypeVoteSigningKey` are deleted outright — no deprecation window, matching upstream's own hard removal.

## Tests added

- `ledger/leios/bls_test.go`: proof-of-possession verification — valid key, nil key, wrong public-key length, mismatched key/proof, and a vote signature rejected when reused as a possession proof.
- `ledger/leios/manager_test.go`: a committee member's vote verifies from an on-chain key with **zero** registry entries; an on-chain key with an invalid proof is excluded rather than trusted.
- `database/plugin/metadata/sqlstore/store_test.go`: round-trip of the new columns through `ImportPool`/`GetPools`, including that a later registration with no key clears a previously stored one.
- Updated existing suites (`committee_test.go`, `keys_test.go`, `manager_test.go`, `registry_test.go`) to remove coverage of deleted behavior and reflect the new descending-stake committee order.

Closes #3148 

---
## Summary by cubic
Switches Leios voting to on-chain BLS keys with proof-of-possession verified using the shared vote DST (matching `cardano-crypto-class`) and deterministic stake-coverage committees. Enforces on-chain keys during enable/validation, scopes PoP checks to committee members, propagates key-provider failures, and clarifies voting-key validation errors (Linear #3148).

- **New Features**
  - Persist and use on-chain `leios_key` (96-byte pub, 48-byte PoP) from pool certs; new DB columns on `pool`/`pool_registration`, exposed via `ledger.LedgerView.GetLeiosKeys`.
  - Verify PoP via `VerifyLeiosKeyProofOfPossession` using the shared vote DST; invalid/missing keys are treated as absent. `VoteManager` verifies only committee members once per epoch, caches passing keys, treats keyless seats as non-signing, and propagates key-provider errors (no static registry fallback).
  - Committee now only uses `ComputeCommittee` (stake-coverage, descending stake); removed the prototype “every pool votes” path.
  - `ValidatePrototypeEbCertificate` now returns `(sigChecked bool, err)` and no longer passes when a signer key is unknown.
  - Node wires a `LeiosKeyProvider` backed by `LedgerView.GetLeiosKeys`; provider API drops the unused epoch arg. EnableVoting/ValidateVotingKey resolve on-chain keys and hard-reject a configured key that mismatches a resolvable on-chain registration. Clarified voting-key validation error text.

- **Migration**
  - Database auto-migrates to schema v2 (`leios-key-registration`): adds `leios_key_public` and `leios_key_possession_proof` (no manual steps). Configure `leiosVoteSigningKeyFile` to emit votes; removed `ComputePrototypeCommittee` and `DerivePrototypeVoteSigningKey`.

<sup>Written for commit 2985f9b748923ba7f0f34dbdb89405a1671fff1d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3159?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for registering and persisting optional Leios voting public keys and proof-of-possession data.
  * Added verification for registered voting keys and on-chain key resolution.
  * Nodes can participate in voting with a configured signing key or operate as non-voting relays.
* **Updates**
  * Committee selection now uses stake-based ordering and verified-stake quorum handling.
  * Voting uses verified on-chain keys when available, with registry fallback support.
  * Updated documentation for voting, quorum, key registration, and database storage.


ledger/leios/bls.go
```diff
@@ -151,6 +151,57 @@ func VerifyVoteSignature(
...
+	var pub bls12381.G2Affine
+	// SetBytes validates curve membership and subgroup order, matching
+	// ParseVoterPublicKey's checks.
+	if _, err := pub.SetBytes(key.PublicKey); err != nil {
+		return fmt.Errorf("%w: %w", ErrInvalidPublicKey, err)
+	}
+	if pub.IsInfinity() {
+		return fmt.Errorf("%w: point is infinity", ErrInvalidPublicKey)
+	}
+	if err := VerifyVoteSignature(
```
chrisguiney, line 207: **Blocking:** this verifies the possession proof under the *signature* DST rather than the *proof-of-possession* DST.

`LeiosVoteDST` (bls.go:28) is `"BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_"`. The IETF BLS signature draft defines two DSTs for the PoP ciphersuite, and they are not interchangeable:

- `hash_to_point`, used by CoreSign/CoreVerify: `BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_`
- `hash_pubkey_to_point`, used by PopProve/**PopVerify**: `BLS_POP_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_`

Section 4.1 states that the `hash_pubkey_to_point` DST *"MUST be distinct from the domain separation tag used for hash_to_point"*, and recommends deriving it from the ciphersuite ID with the `"BLS_POP_"` prefix in place of `"BLS_SIG_"`.

So the reasoning in the comment above — that the `_POP_` suffix designates one ciphersuite shared by both, "distinguished only by the signed message" — has it inverted. The `_POP_` suffix names the *scheme variant*; within that variant there are still two DSTs, differing by the `BLS_SIG_` / `BLS_POP_` prefix. Since this is MinSig (signatures in G1), `BLS12381G1` is right in both — only the prefix needs to change.

**Why the "fails safe" framing doesn't quite hold.** It's correct that this can't accept a forged key. But the failure direction is that a pool registering a *spec-correct* PoP gets rejected, is treated as keyless, and cannot vote. If pools register correctly against the reference implementation, every committee seat ends up keyless and quorum becomes unreachable — which is a total outage of the feature this PR exists to deliver, on the network it's being respun for. Fails-closed-on-everything is a different risk from fails-closed-on-bad-input.

**Secondary, smaller:** sharing the DST means a PoP and a vote signature are the same primitive over the same domain, so the separation rests entirely on the two message spaces never colliding — a 34-byte CBOR vote message vs a 96-byte public key. That holds today, but it's a property of the current prototype message format rather than a structural guarantee, which is what the distinct PoP DST is there to provide. `TestVerifyLeiosKeyProofOfPossessionRejectsVoteSignatureAsProof` passes for that incidental reason.

**Worth knowing: no test in the suite can catch this.** `testLeiosKey` (bls_test.go:228) builds the proof with `SignVote(key, pub)` — the same DST the verifier uses — so the round trip is self-consistent under *any* DST value. That's why CI is green on it.

Suggested shape:

```go
// leiosPopDST is the hash_pubkey_to_point ciphersuite for PopProve/
// PopVerify. Per the IETF BLS-signature draft it MUST differ from the
// CoreSign DST (LeiosVoteDST), which is why this is its own constant
// rather than a reuse.
const leiosPopDST = "BLS_POP_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_"
```

verified with a DST-parameterised variant of `VerifyVoteSignature`. One interop vector from `cardano-crypto-leios` (a real registered key plus its proof) would pin it and close the `UNVERIFIED AGAINST REFERENCE` note for good — the current tests can't, however many cases they add, as long as they generate the proof with the same function that checks it.

node_leios.go
```diff
@@ -49,6 +50,55 @@ func (a *leiosStakeDistributionAdapter) GetStakeDistribution(
...
+// resolving registered Leios keys for exactly the pools the caller names
+// (the same set VoteManager already fetched a stake distribution for). It
+// returns raw (unverified) keys -- VoteManager itself checks proof of
+// possession before trusting one.
+type leiosKeyProviderAdapter struct {
+	ledgerState *ledger.LedgerState
+}
+
+func (a *leiosKeyProviderAdapter) GetLeiosKeys(
+	_ uint64,
```
chrisguiney: Non-blocking, but worth resolving one way or the other before this ships.

The `epoch` parameter is discarded here, and `LedgerView.GetLeiosKeys` reads current pool params rather than state as of a snapshot epoch. That leaves the interface promising something the only production implementation doesn't do — `LeiosKeyProvider`'s doc says it "supplies each active pool's raw (unverified) registered Dijkstra/Leios BLS key **for a snapshot epoch**".

The reason it matters: `committeeAndParamsForEpoch` memoizes `onChainKeys` per epoch, populated on first use. Two nodes that first compute epoch N's committee at different wall-clock times — one before a key rotation lands on chain, one after — end up caching different key sets for the *same* epoch, and can then disagree about whether a given vote verifies. On a prototype network that's confusing rather than dangerous, but it's the kind of divergence that's painful to diagnose during a respin.

Either honour the epoch (resolve from the historical pool params for that snapshot), or drop the parameter from the interface and say plainly in the doc comment that resolution is current-state and a mid-epoch rotation is visible immediately. Dropping it is perfectly defensible for a prototype — the `_ uint64` is what makes it look like an oversight rather than a decision.

ledger/leios/manager.go
```diff
@@ -517,42 +599,57 @@ func (m *VoteManager) committeeAndParamsForEpoch(
...
 			err,
 		)
 	}
+	// Reuse the pool set already fetched above rather than handing the key
+	// provider just the epoch and letting it re-fetch the stake
+	// distribution itself -- that would double the DB round trip on every
+	// new-epoch committee computation.
+	poolKeyHashes := make([]string, 0, len(poolStakes))
+	for poolHashHex := range poolStakes {
+		poolKeyHashes = append(poolKeyHashes, poolHashHex)
```
chrisguiney: Non-blocking performance note.

`poolKeyHashes` is built from the full `poolStakes` map, so `resolveOnChainKeys` verifies a proof of possession for **every pool on the chain** — but `resolveVoterKey` only ever looks up `member.PoolKeyHash`, and `ComputeCommittee` returns `pools[:selectedCount]`, a strict descending-stake prefix. Most of that verification work can never be read back.

I measured `VerifyLeiosKeyProofOfPossession` at ~0.75 ms/key on this branch, so at Cardano's ~3000 registered pools that's roughly 2.3 s of pairing work per epoch, the large majority of it discarded.

It compounds a little with the (deliberate, and otherwise sensible) decision to compute outside the lock and tolerate concurrent duplicate computation: a burst of votes arriving at an epoch boundary can have several goroutines each paying that cost before one wins the memo race.

Filtering to `committee.Members` before verifying addresses both, and keeps the nice property you called out in the comment above — still one stake-distribution fetch, still one `GetPools` round trip.

ledger/leios/certificate.go
```diff
@@ -221,23 +218,26 @@ func ValidatePrototypeEbCertificate(
...
-		return fmt.Errorf(
+		return false, fmt.Errorf(
 			"%w: signer stake %d of total active stake %d below threshold %s",
 			ErrQuorumNotMet,
 			signerStake,
 			committee.TotalActiveStake,
 			quorumStakeThreshold.String(),
 		)
 	}
+	if !allKeysKnown {
```
chrisguiney, line 241: Non-blocking observation, no change needed here.

I checked the PR's claim that nothing calls this outside the package, and it holds — neither `ValidatePrototypeEbCertificate` nor `ValidateEbCertificate` has a production caller today, only tests. So the API break is genuinely free, which is good.

The flip side is that the `sigChecked` contract is currently unenforced by anything: "quorum met but I could not confirm the signature" and "quorum met and the signature verified" are distinguishable only by a bool that no production code reads yet. Returning `(false, nil)` instead of deriving a fake key is clearly the right call, and matching `ValidateEbCertificate`'s shape is right too — it just means the first caller to be added has to remember to check it, with no compiler help.

If it ever grows a caller, a named result type (or having the caller-facing wrapper treat `!sigChecked` as a hard failure) would make the distinction harder to drop by accident. Fine to leave exactly as-is for now.

chrisguiney (changes requested):
Thanks for this — it's a lot of careful work, and several things in it are genuinely well done. The memoization discipline in `committeeAndParamsForEpoch` stands out: not caching a transient key-provider failure, and the comment explaining *why* an empty `onChainKeys` map would poison the epoch, is exactly the reasoning I'd want to find there. The lock handling is correct too — I checked both `handleResolvedPrototypeVote` call sites and the new `EnableVoting`/`ValidateVotingKey` paths, and all of them release or avoid `m.mu` before re-entering `committeeAndParamsForEpoch`, so there's no reentrancy on the non-reentrant mutex.

I'm requesting changes for one thing: the proof-of-possession domain separation tag. You flagged the uncertainty yourself in the `UNVERIFIED AGAINST REFERENCE` comment and asked for it to be confirmed before shipping against a real network — that was the right call, and this is that confirmation. It doesn't match the IETF draft, and the consequence is worse than the "fails safe" framing suggests. Details inline, with the spec citation and a concrete fix.

Three non-blocking notes inline as well, all fine as follow-ups.

**Verified clean:** `go build ./...`, `go test -race` on `ledger/leios` / `database` / `ledgerstate`, `gofmt`, `golangci-lint` (0 issues), `nilaway`, `make docs-parity`, `make import-boundaries`, and all 9 CI checks. Both `ARCHITECTURE.md` and `DATABASE.md` are updated.

**The migration checks out.** I rendered the v2 DDL as each backend actually receives it: postgres correctly gets `BYTEA` with double-quoted identifiers, mysql keeps backticks and `blob`. Authoring it sqlite-only and letting `translateSchemaSQLInSchema` handle the rest is the right call. And since `GetPools` already dedupes and chunks by `dialect.ParameterLimit()`, the new per-epoch query is safe at mainnet pool counts rather than tripping a bound-parameter ceiling.

ledger/leios/bls.go
arepala-uml: Confirmed and fixed. I read the shared `_POP_` suffix as "this ciphersuite is reused for both," when it actually just names the variant, and the `BLS_SIG_`/`BLS_POP_` prefix is the real separator the draft requires between CoreSign and PopProve/PopVerify.

- Added `leiosPopDST` with the correct prefix.
- Refactored verification to take the DST as a parameter, so the two code paths can never accidentally collapse back onto one constant.
- Audited every test that constructs a PoP proof (5 sites in `manager_test.go`, plus the `bls_test.go` helper) — all of them were signing with the wrong DST too, and would have started failing for the *right* reason as soon as verification changed. Good catch that the test suite structurally couldn't have caught this on its own.
- Added a DST-literal pinning test.

ledger/leios/certificate.go
arepala-uml: Agreed on all counts, no functional change. Added a doc note on both functions flagging the unenforced contract explicitly, so the first caller has something pointing at the risk even without compiler help. Leaving the actual type as a bare bool for now per your call.

ledger/leios/manager.go
arepala-uml: Good catch, and the measurement makes it concrete & fixed. `committee.Members` is already sitting right there as the `ComputeCommittee` output at the point `poolKeyHashes` gets built, so this was just building it from the wrong source. Filtered to committee members only, which also keeps the "still one stake-distribution fetch, still one `GetPools` round trip" property you flagged as worth preserving.

node_leios.go
arepala-uml: Went with dropping the parameter rather than honoring it.

- Digging into it, this turns out to already be the established pattern here — `poolVrfKeyHashes` resolves VRF key hash for `PoolDistr2` the same way, from current pool params rather than the historical snapshot.
- So this isn't a new simplification, just an existing one I hadn't made explicit.
- Updated the `LeiosKeyProvider` doc comment to say so plainly, including the cross-node-timing caveat you raised (two nodes computing the same epoch's committee around a rotation could cache different key sets) — agreed that's a real thing worth knowing about for respin debugging even though it's not worth solving right now.

arepala-uml:
> Thanks for this — it's a lot of careful work, and several things in it are genuinely well done. The memoization discipline in `committeeAndParamsForEpoch` stands out: not caching a transient key-provider failure, and the comment explaining _why_ an empty `onChainKeys` map would poison the epoch, is exactly the reasoning I'd want to find there. The lock handling is correct too — I checked both `handleResolvedPrototypeVote` call sites and the new `EnableVoting`/`ValidateVotingKey` paths, and all of them release or avoid `m.mu` before re-entering `committeeAndParamsForEpoch`, so there's no reentrancy on the non-reentrant mutex.
> 
> I'm requesting changes for one thing: the proof-of-possession domain separation tag. You flagged the uncertainty yourself in the `UNVERIFIED AGAINST REFERENCE` comment and asked for it to be confirmed before shipping against a real network — that was the right call, and this is that confirmation. It doesn't match the IETF draft, and the consequence is worse than the "fails safe" framing suggests. Details inline, with the spec citation and a concrete fix.
> 
> Three non-blocking notes inline as well, all fine as follow-ups.
> 
> **Verified clean:** `go build ./...`, `go test -race` on `ledger/leios` / `database` / `ledgerstate`, `gofmt`, `golangci-lint` (0 issues), `nilaway`, `make docs-parity`, `make import-boundaries`, and all 9 CI checks. Both `ARCHITECTURE.md` and `DATABASE.md` are updated.
> 
> **The migration checks out.** I rendered the v2 DDL as each backend actually receives it: postgres correctly gets `BYTEA` with double-quoted identifiers, mysql keeps backticks and `blob`. Authoring it sqlite-only and letting `translateSchemaSQLInSchema` handle the rest is the right call. And since `GetPools` already dedupes and chunks by `dialect.ParameterLimit()`, the new per-epoch query is safe at mainnet pool counts rather than tripping a bound-parameter ceiling.

I have fixed all the issues and added corresponding test for each fix.

- **Blocking (PoP DST):** Reverted. Initially added a separate `BLS_POP_`-prefixed DST per the IETF draft's textbook PopProve/PopVerify construction, but while chasing a real interop vector I found that `cardano-crypto-class`'s actual reference implementation (`minSigPoPDST`) reuses one DST for both ordinary signing and the possession proof — full evidence in the thread below. Reverted to the shared-DST version with a corrected, well-cited comment, and added `TestLeiosVoteDSTMatchesReferenceMinSigPoPDST` pinning the value against the reference so this doesn't flip either direction without evidence again.

- **`LeiosKeyProvider` epoch param:** Dropped it from the interface (option b) rather than threading historical resolution through — matches the existing precedent for VRF key hash in `PoolDistr2` (`poolVrfKeyHashes`), which has the same current-state-not-snapshot property already. Documented the tradeoff explicitly on the interface.
- **Perf:** `resolveOnChainKeys` now only requests keys for `committee.Members` (the post-`ComputeCommittee` stake-coverage prefix) instead of every pool in the stake distribution. Should cut the pairing work roughly to committee size instead of total registered-pool count.
- **`sigChecked` contract:** Added a doc note on both `ValidateEbCertificate` and `ValidatePrototypeEbCertificate` flagging that the bool is currently unenforced and the first real caller should either use a named result type or hard-fail on `!sigChecked` — agreed it's fine to leave the shape as-is for now.

ledger/leios/bls.go
arepala-uml: @chrisguiney  I have reverted this, and want to walk through why since it directly contradicts the review comment above.

Firstly, I made the change exactly as suggested — added `leiosPopDST` with the `BLS_POP_` prefix, parameterized verification, audited and fixed all 5 test sites, added the pinning test. All of that shipped in the commit referenced above.

**Then, while working on cubic's follow-up request below** (add a real interop vector to close the `UNVERIFIED AGAINST REFERENCE` note), I went and read the actual reference implementation `minSigPoPDST` is named after: `cardano-crypto-class`'s `Cardano.Crypto.DSIGN.BLS12381.Internal`. What's actually there:

```haskell
minSigPoPDST :: BLS12381SignContext
minSigPoPDST = BLS12381SignContext (Just "BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_") Nothing
```

Identical to our `LeiosVoteDST`. And the module's own documented, executable example uses that *same* context for both the possession proof and an ordinary signature:

```haskell
let ctx = minVerKeyPoPDST
    pop1 = createPossessionProofDSIGN ctx sk1   -- the PoP
...
let sig1 = signDSIGN ctx msg sk1                -- an ordinary signature
```

`createPossessionProofDSIGN`/`verifyPossessionProofDSIGN` take the DST as a caller-supplied parameter and delegate straight to plain `signDSIGN`/`verifyDSIGN` — there's no internal `hash_pubkey_to_point` step or separate `BLS_POP_`-prefixed constant anywhere in that file. I searched specifically for one.

**So, respectfully: I don't think this comment's central claim holds up against the actual reference implementation.** The IETF citation is accurate as general spec guidance for a from-scratch implementation — but cardano's actual shipped code deviates from that textbook two-DST construction and reuses one DST for both, relying on the two message spaces (vote message vs. serialized pubkey) never colliding, which is exactly the property this comment flagged as "not a structural guarantee." That's apparently a deliberate implementation choice upstream, not a spec violation.

**Reverted** back to the shared-DST version, with the comment on `LeiosVoteDST` now citing this directly (file, function, the shared-context example) instead of the vaguer "the suffix means it's shared" reasoning that was there before and that this review correctly picked apart. Added `TestLeiosVoteDSTMatchesReferenceMinSigPoPDST` pinning the value against `minSigPoPDST` specifically, so this can't flip back and forth again without someone bringing evidence.

I haven't exhaustively searched `ouroboros-consensus`'s `leios-prototype` branch for a Leios-specific crypto path that might implement the separate-DST version properly instead of going through this generic module. If there's a source for the separate-DST claim I'm missing, I'd like to see it — this is precisely the kind of thing that should be settled with a real vector before it ships, which was the right ask in the first place.

arepala-uml:
> Thanks for this — it's a lot of careful work, and several things in it are genuinely well done. The memoization discipline in `committeeAndParamsForEpoch` stands out: not caching a transient key-provider failure, and the comment explaining _why_ an empty `onChainKeys` map would poison the epoch, is exactly the reasoning I'd want to find there. The lock handling is correct too — I checked both `handleResolvedPrototypeVote` call sites and the new `EnableVoting`/`ValidateVotingKey` paths, and all of them release or avoid `m.mu` before re-entering `committeeAndParamsForEpoch`, so there's no reentrancy on the non-reentrant mutex.
> 
> I'm requesting changes for one thing: the proof-of-possession domain separation tag. You flagged the uncertainty yourself in the `UNVERIFIED AGAINST REFERENCE` comment and asked for it to be confirmed before shipping against a real network — that was the right call, and this is that confirmation. It doesn't match the IETF draft, and the consequence is worse than the "fails safe" framing suggests. Details inline, with the spec citation and a concrete fix.
> 
> Three non-blocking notes inline as well, all fine as follow-ups.
> 
> **Verified clean:** `go build ./...`, `go test -race` on `ledger/leios` / `database` / `ledgerstate`, `gofmt`, `golangci-lint` (0 issues), `nilaway`, `make docs-parity`, `make import-boundaries`, and all 9 CI checks. Both `ARCHITECTURE.md` and `DATABASE.md` are updated.
> 
> **The migration checks out.** I rendered the v2 DDL as each backend actually receives it: postgres correctly gets `BYTEA` with double-quoted identifiers, mysql keeps backticks and `blob`. Authoring it sqlite-only and letting `translateSchemaSQLInSchema` handle the rest is the right call. And since `GetPools` already dedupes and chunks by `dialect.ParameterLimit()`, the new per-epoch query is safe at mainnet pool counts rather than tripping a bound-parameter ceiling.


@chrisguiney  I have went and pulled the actual reference implementation this DST scheme is modeled on: `cardano-crypto-class`'s `Cardano.Crypto.DSIGN.BLS12381.Internal` (the source of `minSigPoPDST`). Turns out Cardano's real code reuses **one** DST for both jobs — signing votes and proving key ownership — not two. `minSigPoPDST` is defined identically to our `LeiosVoteDST`, there's no second `BLS_POP_`-prefixed DST anywhere in the file, and the module's own documented example uses that same one context for both a possession proof and an ordinary signature, side by side.

So the IETF citation here was accurate as general spec guidance, but it turns out not to be what Cardano's actual implementation does — the real code takes a simpler path than the textbook construction. Good catch flagging that this needed a real reference check rather than just my read of the spec, since that's exactly what settled it.

Reverted back to the shared-DST version, with the comment on `LeiosVoteDST` now pointing straight at this source, and added `TestLeiosVoteDSTMatchesReferenceMinSigPoPDST` so it's pinned for good.

I have checked `cardano-crypto-class` thoroughly but only checked a couple of files in `ouroboros-consensus`. Please let me know if we missed anything here.

chrisguiney (approved):

Issue #3078, opened 2026-08-02
test(devnet): add an event-driven accelerated release scenario


wolf31o2:
## Goal

Provide a fast, event-driven DevNet scenario for scheduled and release
integration evidence. Keep the canonical-timing DevNet for soak/canary work;
do not make a long wall-clock chain a required per-PR GitHub Actions job.

## Problem

The current harness repeatedly opens fresh Node-to-Node connections to query
tips and several tests each wait through their own relative slot windows. With
one-second slots and 500-slot epochs, network execution spends most of its time
waiting and can turn transient startup or fork recovery into expensive,
low-signal failures.

## Scope

- Add persistent ChainSync observers for each node. Consume RollForward and
  RollBackward messages and expose timeout-bound conditions to the tests.
- Add an accelerated DevNet configuration with shorter, internally valid slot,
  security-window, and epoch parameters. Preserve the existing canonical
  configuration separately.
- Run one scenario timeline for each topology:
  - all-Dingo producers plus relay;
  - Dingo beside `cardano-node`.
- On that timeline verify readiness, block and transaction propagation, chain
  agreement, a controlled peer interruption/recovery, relay restart, and one
  epoch transition.
- Derive assertions from observed protocol events and deterministic expected
  points. Keep statistical block-rate and canonical-timing checks in external
  soak/canary runs.
- Build images once per tested commit where the runner permits it and always
  retain node logs/configuration on failure.

## Acceptance criteria

- [ ] Observers stream chain events instead of polling by repeatedly opening
      tip-query connections.
- [ ] Independent tests do not add serial relative-slot waiting; assertions
      share one scenario timeline.
- [ ] Both all-Dingo and mixed Dingo/`cardano-node` topologies are supported.
- [ ] The accelerated scenario crosses an epoch and exercises relay restart
      and peer recovery.
- [ ] Scenario execution, excluding image build, completes within five minutes
      on a documented reference runner and has a hard timeout.
- [ ] The canonical one-second-slot configuration and longer scenarios remain
      available for release/canary and diagnostic use.
- [ ] Synchronization uses events and bounded contexts, never `time.Sleep`.
- [ ] Failures preserve logs, generated genesis/configuration, observed chain
      events, and container status.

Issue #2197, opened 2026-05-07
devnet TEST_TIMEOUT default starves TestEpochBoundaryConsensus


wolf31o2:
`internal/test/devnet/run-tests.sh` defaults `TEST_TIMEOUT=5m`. The earlier scenarios consume essentially the whole budget, so `TestEpochBoundaryConsensus` panics on the suite-wide timeout before it can do any real work.

Cumulative time before the boundary test starts on a clean main run (300fef4e):

```
TestHarnessGetChainTip               0.03s
TestHarnessWaitForSlot               6.02s
TestHarnessVerifyConsensus          18.03s
TestBasicBlockForging               16.05s
TestDingoChainAdvances              12.01s
TestChainGrowthRate                120.03s
TestRelayPropagation                 0.02s
TestSustainedConsensus             104.06s
TestCardanoProducerChainAdvances    12.02s
TestEpochBoundaryConsensus          panic: test timed out after 5m0s
```

That is roughly 4m48s before the last test even starts. The boundary test waits for slot `epochLength + 0.4*epochLength = 700` (~12 min wall time at 1s slots), so it cannot fit even with the whole budget to itself.

Re-run with `TEST_TIMEOUT=20m -run TestEpochBoundaryConsensus` against the same commit:

```
--- PASS: TestEpochBoundaryConsensus (698.05s)
```

Both producers landed at `slot=702 block=282` after the epoch-500 boundary, so the consensus path itself is fine. This is purely the suite timeout being too tight.

Fix: raise the `TEST_TIMEOUT` default in `internal/test/devnet/run-tests.sh` to something that comfortably covers the full set, e.g. 20m.

PR #3161, opened 2026-08-11
refactor(node): extract the chainsync stall recycler into a component

### LINKED ISSUE #2261: node: split Run and stall recycler into testable lifecycle components
## Summary

`Node.Run()` still mixes startup wiring, lifecycle management, and chainsync stall recycling logic.

## Audit Details

Finding ID: H11, HIGH architecture finding.

Affected file: `node.go`.

The audit called out `Node.Run()` as a long method that initializes components, wires event subscriptions, starts listeners, and contains long-lived stall detection behavior. The stall checker now has panic recovery helpers, but it is still launched from `Node.Run()` and shares node lifecycle concerns with startup code.

Impact: startup/shutdown behavior is hard to test independently, and changes to stall detection risk touching unrelated node wiring.

## What Needs to Happen

1. Extract the chainsync stall recycler into a component with explicit `Start` / `Stop`.
2. Give it injectable dependencies for ledger tip, upstream tip, peer state, and connection recycling.
3. Unit test plateau detection, cooldown behavior, shutdown, and panic recovery outside `Node.Run()`.
4. Continue slimming `Node.Run()` by moving startup slices into named setup methods where ownership becomes clearer.
5. Verify graceful shutdown with `go test -race ./...`.

## Acceptance Criteria

- Stall recycler behavior can be tested without constructing a full node.
- `Node.Run()` no longer owns the recycler loop internals.

chrisguiney:
Closes #2261

## Why

`Node.Run()` mixed startup wiring, lifecycle management, and chainsync stall recycling. Every recycler function in `node_chainsync_recycler.go` was a `*Node` method reading unsynchronized node fields, so the plateau/cooldown/backlog decision logic could only be tested by constructing a partial `&Node{}` — and the ledger-dependent paths needed a real database, chain manager, and `LedgerState` (`newNodeTestDivergedLedger`) just to reach a branch that decides whether to publish an event.

## What changed

**New component: `internal/chainsyncrecycler.Recycler`** — a `Start(ctx) error` / `Stop()` background component owning only the stall/plateau decision logic. Its dependencies are interfaces declared on the consumer side (`LedgerSource`, `ChainsyncState`, `ChainSelector`, `EventPublisher`), satisfied structurally by `ledger.LedgerState`, `chainsync.State`, `chainselection.ChainSelector`, and the `EventBus` — so the dependency only goes one way and the component can be driven entirely by fakes. Timing policy moved into `Config`; the `chainsync.Config` dependency is gone since only `StallTimeout` was ever read.

**Live-component access.** The recycler must not hold the ledger or chainsync state across ticks: a live restore/truncate nils and rebuilds both while holding `liveLifecycleMu` for its whole quiesce-through-reinitialize duration. The node supplies a `ComponentProvider` (`nodeRecyclerComponents`) whose `WithLiveComponents` takes that `TryLock`, nil-checks both fields, and holds the lock for the entire callback — preserving the existing "skip a contended tick rather than block shutdown behind a truncate" behavior while making it directly testable. It also assigns the `ChainSelector` interface only when the selector exists, so a typed-nil pointer cannot defeat the recycler's nil checks.

**`Node.Run()`: 1494 → 1323 lines.** The recycler block collapsed to a `startChainsyncStallRecycler` call, plus two event-subscription slices the audit named were extracted: `subscribeChainSelectorEvents()` and `subscribeConnectionEvents()`.

`node_chainsync_recycler.go` (857 → 300 lines) keeps the chainsync wiring that isn't the recycler: peer tip/rollback observation, apply-eligibility, the ingress-eligibility cache, and the chain-switch / selected-none handlers.

Behavior is unchanged — tick logic, threshold derivation, catch-up multiplier, cooldown handling, and panic-recovery semantics all move verbatim.

## Acceptance criteria

- [x] Stall recycler behavior can be tested without constructing a full node.
- [x] `Node.Run()` no longer owns the recycler loop internals.

## Testing

50 tests cover this area, up from 27.

- **32 in `internal/chainsyncrecycler`**, all against fakes with no database: plateau detection and resync, cooldown, ledger-application-backlog suppression, reconcile-before-suppression, reconcile-failure fallthrough, post-plateau realign, single-eligible-peer guard, non-primary client removal, no-active-selection recycle, catch-up multiplier, cooldown pruning, `Start`/`Stop`, double-`Start`, parent-context cancellation, tick panic recovery, loop panic recovery and restart, and startup-baseline skip.
- **8 node-level** covering what is genuinely node-specific: lock hold and release, panic-safe unlock, the mid-reinitialization nil cases, that `snapshotMu` does not block a tick, and shutdown ordering.

The database fixture the reconcile path used to need (`newNodeTestDivergedLedger`) is deleted. `TestNodePeerEligibility/PriorityEventUpdatesChainSelector` previously reimplemented the production wiring inline (`// Mirror the subscription wiring in node.go`); they now call the real `subscribeChainSelectorEvents()`.

Verification run locally:

- `go test -race ./...` — 67 packages ok, zero failures (conformance included).
- `golangci-lint run ./...`, `nilaway ./...`, `modernize ./...`, `make import-boundaries`, `make docs-parity`, `make gorm-check`, `make golines` — clean.
- DevNet (`internal/test/devnet/run-tests.sh`, all-dingo network) — run in progress at the time of opening; the harness suite passed (`ok internal/test/devnet 44.373s`, consensus verified across all three producers plus the relay). I will follow up with the scenario results.

## Documentation

`ARCHITECTURE.md` — startup step 18, the shutdown phase-1 entry, the new component paragraph in the Multi-Client Chainsync section, the `isLedgerApplicationBacklog` file reference, the package dependency graph and source tree, and a stale `runStallCheckerTick` reference.

`DATABASE.md` — checked, not affected: no schema, query-surface, or storage change.

---
## Summary by cubic
Extracted the chainsync stall recycler into a standalone `internal/chainsyncrecycler.Recycler` to simplify node lifecycle and make stall/plateau logic easy to test. Behavior is unchanged; `Node.Run()` is slimmer and the recycler reads live components each tick.

- **Refactors**
  - Added `Recycler` with `Start(ctx)`/`Stop()` owning only stall/plateau decisions.
  - Uses consumer-side interfaces (`LedgerSource`, `ChainsyncState`, `ChainSelector`, `EventPublisher`) satisfied by `ledger.LedgerState`, `chainsync.State`, `chainselection.ChainSelector`, and the `EventBus`.
  - Moved timing policy into `Config`; removed the `chainsync.Config` dependency.
  - Introduced a live component provider that locks during callbacks and skips contended ticks to keep shutdown safe and non-blocking.
  - Extracted `subscribeChainSelectorEvents()` and `subscribeConnectionEvents()`; `Node.Run()` delegates to `startChainsyncStallRecycler(...)`.

- **Bug Fixes**
  - Reattached misplaced doc comments so `godoc` shows the right function docs.
  - Held the recycler mutex across cancel and wait in `Stop()` to prevent a race with concurrent `Start()`.
  - Updated the tick-panic test to exercise tick-level recovery (not the baseline read), asserting via log messages.
  - Removed an inert `shutdownTimeout` from a test that never read it.

<sup>Written for commit beaba7ca10fdeb2dce9f06c01fa402c79c90510c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added automatic detection and recovery for stalled chainsync clients and local synchronization plateaus.
  * Added resync, connection recycling, and stalled-peer removal actions, with cooldowns and catch-up grace periods.
  * Improved startup and shutdown handling, including graceful cancellation and recovery from runtime errors.
  * Added ledger and peer reconciliation to avoid unnecessary resyncs and coordinate connected peers.

* **Documentation**
  * Updated architecture documentation to describe the new synchronization recovery component and lifecycle behavior.


chrisguiney:
## DevNet follow-up

Reporting the DevNet result promised in the description. The first run had two scenario failures, so I ran an A/B against unmodified `upstream/main` (8f2a0ed5, the same commit this branch is based on) before drawing any conclusion.

| Scenario | branch, run 1 | `upstream/main` | branch, run 2 |
|---|---|---|---|
| `TestDingoChainAdvances` | PASS | **FAIL** | **FAIL** |
| `TestChainGrowthRate` | **FAIL** | PASS | PASS |
| `TestSustainedConsensus` | **FAIL** | **FAIL** | PASS |

Every scenario passes in at least one run and fails in at least one, on both code versions. Nothing fails consistently on this branch, and both scenarios that failed in run 1 pass on this branch once the machine is quiet. `TestSustainedConsensus` — one of the two original failures — also fails on stock `main`.

Two caveats about run 1 that explain it: I was running `go test` (including the full root package) on the same host while the scenarios were live, and these scenarios assert slot-rate progress against a 1-second slot length, so host CPU contention breaks them directly. Runs 2 and 3 were on an otherwise idle machine.

The remaining `TestDingoChainAdvances` failure is identical on `main` and on this branch: the chain is still at `slot 0, block 0` when the 22.5s window expires — the network simply has not begun forging yet. That is startup timing, not chain behavior. This is the class of failure #3078 was filed to fix ("network execution spends most of its time waiting and can turn transient startup or fork recovery into expensive, low-signal failures"), and #2197 was the same shape.

Scenarios that passed on this branch in the full run, for completeness: `TestBasicBlockForging`, `TestDingoChainAdvances`, `TestRelayPropagation`, `TestEpochBoundaryConsensus` (504s, crossing a full 500-slot epoch), `TestStakeAddressInfoQueries`, and the whole harness suite (`ok internal/test/devnet`, consensus verified across all three producers plus the relay).

Worth noting on mechanism as well: a stall recycler defect would not present as all four nodes freezing at an identical tip. The recycler only reacts to stalled ingress by publishing recycle/resync events — it does not gate block production — so a bug there would desynchronize nodes rather than pause forging network-wide.

No change to the diff as a result of this investigation.

chrisguiney:
Addressed all four reviewer findings in beaba7ca. Every one was real; details on the two that mattered most:

**`Recycler.Stop` race (cubic, P2).** Correct. Reading `r.cancel` under the mutex and then releasing it before `wg.Wait()` let a concurrent `Start` register its worker after the Wait had already observed an empty group — `Stop` would report the recycler stopped while a live one kept ticking into components the caller was about to tear down. That invariant is exactly what node shutdown depends on, so it is worth making airtight even though `Run`/`shutdown` never call the two concurrently today. `r.mu` is now held across both the cancel and the Wait; the loop goroutine never takes `r.mu`, so this cannot deadlock against the worker.

**`TestTickPanicIsRecoveredAndTicksContinue` did not test its own name (cubic, P2).** Also correct, and the sharpest catch here. `loop()` reads the ledger tip once for the plateau baseline before the first tick, so a `panickyLedger` that panics on call one landed in `initProgressBaseline` — recovered by `runLoop`, not `runTick`. The test passed via loop restart while tick-level recovery went entirely uncovered, and asserting only "ticks continue" cannot tell the two paths apart because both resume ticking.

`panickyLedger` now skips the baseline read so the panic lands inside a tick, and the test discriminates the recovery paths by their log messages (`panic in stall checker tick, continuing` vs `panic in stall checker goroutine`), additionally asserting the loop did *not* restart. Verified by mutation — restoring the old `skip=0` behavior fails the new assertion:

```
--- FAIL: TestTickPanicIsRecoveredAndTicksContinue (2.00s)
    lifecycle_test.go:289: timeout waiting for channel receive: tick-level panic recovery
```

**Doc comments (CodeRabbit + cubic, P3).** Both flagged it and it was worse than reported: inserting the two subscriber functions left three comment blocks stacked above one function, so godoc reported the chain-selector text for `subscribeConnectionEvents` and left `subscribeChainSelectorEvents` and `nodeSettingsGateValues` undocumented. Each comment is now attached to its own function.

**Inert `shutdownTimeout` (cubic, P3).** Correct — `Config.cfg` is nil in that test, so `configuredShutdownTimeout` falls back to its 30s default and the struct field was never read. Removed, with a note that the test's own `RequireReceive` bounds are what fail it if `Stop` never returns.

Re-verified after the fixes: `go test -race` on the root and recycler packages, golangci-lint, nilaway, import-boundaries, docs-parity, golines.

arepala-uml (approved):

PR #3166, opened 2026-08-12
test(devnet): add an event-driven accelerated release scenario

### LINKED ISSUE #3078: test(devnet): add an event-driven accelerated release scenario

chrisguiney:
Closes #3078.

The DevNet harness queried tips by opening a fresh Node-to-Node connection every
couple of seconds, because a short-lived ChainSync client only ever reports the
tip it captured when it intersected. Each scenario then waited through its own
relative slot windows, so a run spent most of its wall clock waiting and turned
transient startup or fork recovery into expensive, low-signal failures.

This adds a fast scenario built the other way round, plus two bug fixes that
building it uncovered. The three commits are independent and can be reviewed —
or landed — separately.

## `fix(devnet): chown pool keys to the uid the dingo image runs as`

**This one blocks all DevNet use on current `main`.** Every Dingo block producer
dies at startup with:

```
block producer startup validation failed: load pool credentials: failed to load
VRF signing key: failed to read key file "/configs/keys/vrf.skey": permission denied
```

The root `Dockerfile` pins the dingo user to uid/gid 1000, but `configurator.sh`
still chowned each pool's key directory to `100:101`, the uid the image used
before the pin. That directory has to be `0700` — cardano-node refuses to start
when `vrf.skey` is readable by group or other — so the mismatch made the keys
unreadable to the process that needs them.

Compose now passes the ids in rather than the script hardcoding them, since the
authoritative value is the `adduser --uid` pin. Two guards derive their
expectation from the `Dockerfile` instead of restating a number; both were
verified to fail when the drift is reintroduced.

## `fix(ledger): publish chainsync events after releasing the chainsync mutex`

A producer wedged mid-run: still forging, but no longer completing Node-to-Node
handshakes, mempool silent, and logging ~217,000 copies of `event delivery
stalled: subscriber not draining type=connmanager.conn_closed` in five minutes.

The EventBus was behaving as designed — delivery blocks rather than dropping.
The defect was publishing while holding a lock a subscriber of that event needs,
which turns backpressure into a cycle:

1. `handleEventChainsync` holds `chainsyncMutex` and publishes `ChainsyncResyncEventType`.
2. That event's subscriber calls `RecoverAfterLocalRollback`, which takes the same mutex, so its dispatch goroutine parks.
3. The buffer fills, so the publisher never returns. Each side holds what the other waits for.

It does not stay local. `handleConnectionClosedEvent` takes the same mutex, so
`ledger.conn_closed` stops draining; `node.go`'s handler translating
`connmanager.conn_closed` into `ledger.conn_closed` then blocks inside its own
callback, which stops `connmanager.conn_closed` draining, and every subsequent
connection close parks another publisher goroutine.

`pendingPublishes` queues events produced under the lock and flushes them after
the unlock. Delivery semantics are unchanged apart from happening a moment
later: `Publish` already hands off to a buffered channel, so nothing could have
depended on a handler running before the publishing function returned.

`TestNoEventBusPublishWhileHoldingChainsyncMutex` enforces the invariant. It
tells a deferred unlock from an immediate one, so it does not flag the recovery
path that already unlocks before publishing — and it found a third violation a
hand search had missed.

## `test(devnet): add an event-driven accelerated release scenario`

One persistent ChainSync session per node follows the chain, so `RollForward`
and `RollBackward` arrive as nodes produce them. Every assertion is a condition
over those observed events bounded by a context deadline — no polling interval,
no `time.Sleep` in the synchronisation path.

Every assertion shares one timeline: each phase's deadline is measured from the
scenario's start rather than the end of the previous phase, so a phase that
finishes early hands its slack forward. Phases cover readiness, block and
transaction propagation, chain agreement, an epoch transition, a controlled peer
interruption with recovery, and a relay restart. Both disruption phases hold the
node down until the network advances a derived number of blocks (k/2) rather
than a fixed wall-clock time.

The accelerated specs compress slot/epoch/security-parameter timing to fit the
runner budget; the canonical specs are untouched and remain what soak and canary
runs use. `DevNetConfig.Validate` enforces `4k/f` and `3k/f` against
`epochLength` across every checked-in spec.

The pure logic — spec loader and validation, observed-chain state machine,
scenario plan — carries no build tag, so it is unit-tested on an ordinary
`go test ./...` with no Docker. Tests also assert the accelerated specs fit the
five-minute budget and the canonical ones do not, so the fast scenario cannot
quietly be pointed at the soak configuration.

Two further harness defects surfaced and are fixed here:

- Scenarios charged slot-derived timeouts against the ~30s pre-genesis wait, and
  `TestDingoChainAdvances` gated on `slot >= 0`, which every tip satisfies. They
  passed only because an earlier test had absorbed the wait, so each failed when
  run on its own with `run-tests.sh -run <name>` — exactly when debugging.
  `WaitForChainStart` is the real gate. `TestEpochBoundaryConsensus` was clearing
  its budget by under a second and now has full margin.
- Restarting a node with `docker compose start` pulls in `depends_on` and re-runs
  the configurator, regenerating genesis; the restarted node then refuses to
  start on a genesis-hash mismatch. Node control addresses the container directly.

Failures preserve observed chain events per node, container status, per-service
logs, and the generated genesis/configuration, all captured before teardown
removes the volumes.

## Validation

| Check | Result |
|---|---|
| Accelerated scenario (all-Dingo) | **PASS, 1m58s** against a 5-minute hard timeout |
| Accelerated scenario (beside `cardano-node`) | **PASS** |
| Canonical DevNet suite | **PASS** |
| Conformance mode vs `cardano-node` | **PASS** |
| `event delivery stalled` warnings | **0** (was 217,657 in five minutes) |
| Full `go test ./...`, `go test -race ./ledger/` | pass |
| `golangci-lint`, `nilaway`, `modernize`, docs-parity, import-boundaries, ledger conformance vectors | clean |

Docs: `internal/test/devnet/README.md` (accelerated parameters and stability-window
table, scenario timeline, failure artifacts, troubleshooting), plus the EventBus
contract in `ARCHITECTURE.md` and `event/doc.go`. `DATABASE.md` checked, not affected.

## Reviewer notes

- The conformance run was done pre-rebase. It passed on identical code and the
  rebase pulled in only an unrelated database change, but it has not been re-run
  since.
- `TestChainGrowthRate` failed once in three canonical runs and passed on re-run,
  stalling network-wide (all three producers saw comparable fork churn) and
  recovering on its own. Pre-existing and independent of this branch, but now
  demonstrated — it is a candidate for the soak/canary split the issue describes
  rather than a per-PR gate. The accelerated scenario deliberately contains no
  statistical rate check.

---
## Summary by cubic
Adds an event-driven accelerated DevNet scenario (~2 minutes) and prevents `EventBus` deadlocks by deferring publishes out of `chainsync`/`blockfetch` locks. Previously: the harness polled tips and ledger published under locks, causing long waits and backpressure cycles; now: persistent ChainSync observers drive a shared timeline, and lock-held events queue and flush after unlock with ordering preserved. Producers also boot reliably via corrected pool-key ownership.

- Accelerated DevNet usage and scope
  - Run `./run-tests.sh --accelerated` (optionally with `--conformance`) to execute a single shared timeline covering readiness, propagation, agreement, epoch transition, peer interruption (hold = k/2), and relay restart; the canonical suite/defaults are unchanged.
  - Accelerated specs are validated by `DevNetConfig.Validate` (4k/f and 3k/f) and unit tests; pure logic is untagged for `go test` coverage.

- Ledger delivery and test guards
  - Events produced under `chainsyncMutex` or `chainsyncBlockfetchMutex` queue via `pendingPublishes` and flush after unlock; delivery semantics are unchanged apart from deferral.
  - Invariant tests forbid publishing while holding either mutex, cover transitive helper paths, treat nil-queue calls as inline publishes, match the queue parameter by position, and now share nil-queue detection helpers across both guards.

- Harness and scripts
  - Gate on chain start (not readiness alone), address containers directly on restart, accept separate `-test.run`, clear `DEVNET_ACCELERATED` unless requested, and clean only script-created artifact dirs.
  - Compose passes `DINGO_UID`/`DINGO_GID`; the configurator chowns pool-key dirs accordingly. Parity tests verify Dockerfile UID/GID and compose env, and a guard keeps scripts/compose/spec selection in sync.
  - Failures preserve per-node observed events, logs, container status, and generated config. Docs updated in `ARCHITECTURE.md`, `event/doc.go`, and DevNet `README.md`.

<sup>Written for commit 09455ae2e55ea55bd8da205cb51a696259872930. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3166?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added accelerated DevNet scenarios covering propagation, consensus, epoch transitions, rollbacks, interruptions, and recovery.
  * Added improved chain-state monitoring, coordinated waits, and failure-artifact collection for DevNet testing.
  * Added configurable DevNet specifications, timing validation, and accelerated test execution.

* **Bug Fixes**
  * Prevented potential event-publication deadlocks during ledger and chain synchronization processing.

* **Documentation**
  * Documented event backpressure and deadlock-prevention guidance.
  * Expanded DevNet setup, execution, troubleshooting, and accelerated-testing documentation.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -928,6 +928,22 @@ All event types follow the `subsystem.snake_case_name` convention.
   worker pool means it also delays unrelated async event types. Consumers of
   `Subscribe` channels must drain for the life of the subscription and
   `Unsubscribe` when they stop
+- **Never `Publish` while holding a lock that a subscriber of that event
+  acquires.** Because delivery blocks rather than drops, this is a deadlock,
+  not merely a slow path: once the subscriber's buffer fills, the subscriber
+  waits for the lock the publisher holds while the publisher waits for the
+  capacity the subscriber would free. `ledger`'s `pendingPublishes`
+  (`ledger/pending_publish.go`) is the pattern for this — queue the event under
+  the lock and flush after the unlock, registering the flush with `defer`
+  *before* taking the lock so LIFO order runs it last. The invariant is
+  enforced for `chainsyncMutex` by `TestNoEventBusPublishWhileHoldingChainsyncMutex`
+- The blast radius of such a stall is not local. `LedgerState.handleConnectionClosedEvent`
+  takes `chainsyncMutex`, so a stall there stops `ledger.conn_closed` draining;
+  the `node.go` handler translating `connmanager.conn_closed` into
+  `ledger.conn_closed` then blocks inside its own callback, which stops
+  `connmanager.conn_closed` draining, and every subsequent connection close
+  parks another publisher goroutine. Handlers that re-publish are therefore
+  coupling two topics' backpressure and must not block
 - Prometheus metrics for event delivery tracking and latency, including
   `event_delivery_blocked_total{type,kind}` and
   `event_async_enqueue_blocked_total{type}` for backpressure
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -935,7 +935,18 @@ All event types follow the `subsystem.snake_case_name` convention.
   publisher holds while the publisher waits for the capacity the subscriber
   would free. `PublishAsync` is no exception — it waits for room in the shared
   async queue, which is drained by a worker pool running subscriber handlers,
-  so a handler blocked on the publisher's lock closes the same cycle. `ledger`'s `pendingPublishes`
+  so a handler blocked on the publisher's lock closes the same cycle. Both
+  `LedgerState.chainsyncMutex` and `chainsyncBlockfetchMutex` count:
+  `RecoverAfterLocalRollback` takes the first and nests the second inside it,
+  so holding either while publishing is enough to deadlock.
+- The rule is not yet fully enforced in `ledger`. Several helpers reachable
+  from a lock holder still publish `ChainsyncResyncEventType` inline; they are
+  pinned in `knownResyncPublishPathsUnderLock`
+  (`ledger/publish_under_lock_test.go`) and tracked for conversion, which has
+  to happen as one change because a helper flushing on its own return still
+  publishes while a parent frame holds the lock. Read the rule as the target
+  state plus an explicit exception list, not as an invariant already holding
+  everywhere in that package. `ledger`'s `pendingPublishes`
   (`ledger/pending_publish.go`) is the pattern for this — queue the event under
   the lock and flush after the unlock, registering the flush with `defer`
   *before* taking the lock so LIFO order runs it last. The invariant is
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -928,11 +928,14 @@ All event types follow the `subsystem.snake_case_name` convention.
   worker pool means it also delays unrelated async event types. Consumers of
   `Subscribe` channels must drain for the life of the subscription and
   `Unsubscribe` when they stop
-- **Never `Publish` while holding a lock that a subscriber of that event
-  acquires.** Because delivery blocks rather than drops, this is a deadlock,
-  not merely a slow path: once the subscriber's buffer fills, the subscriber
-  waits for the lock the publisher holds while the publisher waits for the
-  capacity the subscriber would free. `ledger`'s `pendingPublishes`
+- **Never `Publish`, `PublishAsync`, or `PublishBlocking` while holding a lock
+  that a subscriber of that event acquires.** Because all three wait for
+  capacity rather than dropping, this is a deadlock, not merely a slow path:
+  once the subscriber's buffer fills, the subscriber waits for the lock the
+  publisher holds while the publisher waits for the capacity the subscriber
+  would free. `PublishAsync` is no exception — it waits for room in the shared
+  async queue, which is drained by a worker pool running subscriber handlers,
+  so a handler blocked on the publisher's lock closes the same cycle. `ledger`'s `pendingPublishes`
   (`ledger/pending_publish.go`) is the pattern for this — queue the event under
   the lock and flush after the unlock, registering the flush with `defer`
   *before* taking the lock so LIFO order runs it last. The invariant is
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -530,6 +530,10 @@ func (ls *LedgerState) handleChainSwitchEvent(evt event.Event) {
 	if !ok {
 		return
 	}
+	// Registered before the mutex is taken so defer's LIFO order runs it
+	// after the unlock. See pendingPublishes.
+	var pending pendingPublishes
+	defer pending.flush()
 	var replayConnId ouroboros.ConnectionId
 	effectiveConnId := e.NewConnectionId
 	var requestFreshCursor bool
@@ -608,6 +612,7 @@ func (ls *LedgerState) handleChainSwitchEvent(evt event.Event) {
 		ls.requestChainsyncResync(
 			effectiveConnId,
 			event.ChainsyncResyncReasonChainSwitchCursorAhead,
+			&pending,
 		)
 		return
 	}
@@ -1225,17 +1230,23 @@ func desiredBlockfetchBatchHeaders(
 	return min(minHeaders, maxHeaders)
 }
 
+// requestChainsyncResync asks the chainsync layer to renegotiate an
+// intersection with a peer.
+//
+// pending must be the caller's queue when ls.chainsyncMutex is held, and
+// may be nil otherwise. This event's subscriber calls
+// RecoverAfterLocalRollback, which takes that mutex, so publishing inline
+// from under the lock is the deadlock pendingPublishes exists to break.
 func (ls *LedgerState) requestChainsyncResync(
 	connId ouroboros.ConnectionId,
 	reason string,
+	pending *pendingPublishes,
 ) {
 	ls.headerMismatchCount = 0
 	ls.rollbackHistory = nil
 	delete(ls.bufferedHeaderEvents, connIdKey(connId))
-	if ls.config.EventBus == nil {
-		return
-	}
-	ls.config.EventBus.Publish(
+	pending.add(
+		ls.config.EventBus,
 		event.ChainsyncResyncEventType,
 		event.NewEvent(
 			event.ChainsyncResyncEventType,
@@ -2417,6 +2428,7 @@ func (ls *LedgerState) handleEventChainsyncBlockHeader(e ChainsyncEvent) error {
 				ls.requestChainsyncResync(
 					e.ConnectionId,
 					event.ChainsyncResyncReasonPersistentFork,
+					nil,
 				)
 			}
 			return nil
@@ -2546,6 +2558,7 @@ func (ls *LedgerState) handleEventChainsyncBlockHeader(e ChainsyncEvent) error {
 		ls.requestChainsyncResync(
 			initialConnId,
 			fmt.Sprintf("blockfetch start failed: %v", err),
+			nil,
 		)
 		return nil
 	}
@@ -2642,6 +2655,7 @@ func (ls *LedgerState) tryResolveFork(
 		ls.requestChainsyncResync(
 			e.ConnectionId,
 			event.ChainsyncResyncReasonRollbackNotFound,
+			nil,
 		)
 		return true, nil
 	}
@@ -2768,6 +2782,7 @@ func (ls *LedgerState) tryResolveFork(
 			ls.requestChainsyncResync(
 				e.ConnectionId,
 				event.ChainsyncResyncReasonRollbackNotFound,
+				nil,
 			)
 			return true, nil
 		}
@@ -3144,6 +3159,7 @@ func (ls *LedgerState) noteBlockfetchRangeUnavailable(
 	ls.requestChainsyncResync(
 		connId,
 		event.ChainsyncResyncReasonBlockfetchRangeUnavailable,
+		nil,
 	)
 	return true
 }
@@ -5069,6 +5085,7 @@ func (ls *LedgerState) handleEventBlockfetchBatchDone(e BlockfetchEvent) error {
 						"empty blockfetch batch alternate retry failed: %v",
 						err,
 					),
+					nil,
 				)
 				return nil
 			}
@@ -5085,6 +5102,7 @@ func (ls *LedgerState) handleEventBlockfetchBatchDone(e BlockfetchEvent) error {
 		ls.requestChainsyncResync(
 			e.ConnectionId,
 			"empty blockfetch batch",
+			nil,
 		)
 		return nil
 	}
@@ -5142,6 +5160,7 @@ func (ls *LedgerState) handleEventBlockfetchBatchDone(e BlockfetchEvent) error {
 		ls.requestChainsyncResync(
 			nextConnId,
 			fmt.Sprintf("blockfetch continuation failed: %v", err),
+			nil,
 		)
 		return nil
 	}
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -435,6 +435,12 @@ func (ls *LedgerState) verifyDeferredBlockHeaderState(
 }
 
 func (ls *LedgerState) handleEventBlockfetch(evt event.Event) {
+	// Registered before the mutex is taken so defer's LIFO order runs it
+	// after the unlock. RecoverAfterLocalRollback nests this mutex inside
+	// chainsyncMutex, so publishing while holding it deadlocks the same
+	// way. See pendingPublishes.
+	var pending pendingPublishes
+	defer pending.flush()
 	ls.chainsyncBlockfetchMutex.Lock()
 	defer ls.chainsyncBlockfetchMutex.Unlock()
 	e, ok := evt.Data.(BlockfetchEvent)
@@ -449,18 +455,17 @@ func (ls *LedgerState) handleEventBlockfetch(evt event.Event) {
 				"component", "ledger",
 				"error", err,
 			)
-			if ls.config.EventBus != nil {
-				ls.config.EventBus.Publish(
+			pending.add(
+				ls.config.EventBus,
+				LedgerErrorEventType,
+				event.NewEvent(
 					LedgerErrorEventType,
-					event.NewEvent(
-						LedgerErrorEventType,
-						LedgerErrorEvent{
-							Error:     err,
-							Operation: "blockfetch_batch_done",
-						},
-					),
-				)
-			}
+					LedgerErrorEvent{
+						Error:     err,
+						Operation: "blockfetch_batch_done",
+					},
+				),
+			)
 		}
 	} else if e.Block != nil {
 		if err := ls.handleEventBlockfetchBlock(e); err != nil {
@@ -475,7 +480,8 @@ func (ls *LedgerState) handleEventBlockfetch(evt event.Event) {
 					"slot", e.Point.Slot,
 					"hash", hex.EncodeToString(e.Point.Hash),
 				)
-				ls.config.EventBus.Publish(
+				pending.add(
+					ls.config.EventBus,
 					ConnectionRecycleRequestedEventType,
 					event.NewEvent(
 						ConnectionRecycleRequestedEventType,
@@ -493,19 +499,18 @@ func (ls *LedgerState) handleEventBlockfetch(evt event.Event) {
 				"slot", e.Point.Slot,
 				"hash", hex.EncodeToString(e.Point.Hash),
 			)
-			if ls.config.EventBus != nil {
-				ls.config.EventBus.Publish(
+			pending.add(
+				ls.config.EventBus,
+				LedgerErrorEventType,
+				event.NewEvent(
 					LedgerErrorEventType,
-					event.NewEvent(
-						LedgerErrorEventType,
-						LedgerErrorEvent{
-							Error:     err,
-							Operation: "blockfetch_block",
-							Point:     e.Point,
-						},
-					),
-				)
-			}
+					LedgerErrorEvent{
+						Error:     err,
+						Operation: "blockfetch_block",
+						Point:     e.Point,
+					},
+				),
+			)
 		}
 	}
 }
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -171,6 +171,11 @@ type peerHeaderChain struct {
 }
 
 func (ls *LedgerState) handleEventChainsync(evt event.Event) {
+	// Registered before the mutex is taken so defer's LIFO order runs it
+	// after the unlock: publishing under ls.chainsyncMutex deadlocks the
+	// node. See pendingPublishes.
+	var pending pendingPublishes
+	defer pending.flush()
 	e, ok := evt.Data.(ChainsyncEvent)
 	if !ok {
 		ls.chainsyncMutex.Lock()
@@ -201,18 +206,20 @@ func (ls *LedgerState) handleEventChainsync(evt event.Event) {
 				// continuing to send the same rollback point.
 				ls.resetChainsyncResyncState()
 				ls.setChainsyncState(SyncingChainsyncState)
-				if ls.config.EventBus != nil {
-					ls.config.EventBus.Publish(
+				// Queued rather than published here: this runs with
+				// ls.chainsyncMutex held, and this event's subscriber
+				// takes that same mutex via RecoverAfterLocalRollback.
+				pending.add(
+					ls.config.EventBus,
+					event.ChainsyncResyncEventType,
+					event.NewEvent(
 						event.ChainsyncResyncEventType,
-						event.NewEvent(
-							event.ChainsyncResyncEventType,
-							event.ChainsyncResyncEvent{
-								ConnectionId: e.ConnectionId,
-								Reason:       event.ChainsyncResyncReasonRollbackLoop,
-							},
-						),
-					)
-				}
+						event.ChainsyncResyncEvent{
+							ConnectionId: e.ConnectionId,
+							Reason:       event.ChainsyncResyncReasonRollbackLoop,
+						},
+					),
+				)
 				return
 			}
 			ls.config.Logger.Error(
@@ -249,19 +256,18 @@ func (ls *LedgerState) handleEventChainsync(evt event.Event) {
 				"slot", e.Point.Slot,
 				"hash", hex.EncodeToString(e.Point.Hash),
 			)
-			if ls.config.EventBus != nil {
-				ls.config.EventBus.Publish(
+			pending.add(
+				ls.config.EventBus,
+				LedgerErrorEventType,
+				event.NewEvent(
 					LedgerErrorEventType,
-					event.NewEvent(
-						LedgerErrorEventType,
-						LedgerErrorEvent{
-							Error:     err,
-							Operation: "block_header",
-							Point:     e.Point,
-						},
-					),
-				)
-			}
+					LedgerErrorEvent{
+						Error:     err,
+						Operation: "block_header",
+						Point:     e.Point,
+					},
+				),
+			)
 			return
 		}
 	}
@@ -654,6 +660,10 @@ func (ls *LedgerState) handleConnectionClosedEvent(evt event.Event) {
 }
 
 func (ls *LedgerState) handleEventChainsyncAwaitReply(evt event.Event) {
+	// See pendingPublishes: nothing may be published while
+	// ls.chainsyncMutex is held.
+	var pending pendingPublishes
+	defer pending.flush()
 	e, ok := evt.Data.(ChainsyncAwaitReplyEvent)
 	if !ok {
 		ls.logUnexpectedChainsyncEventData(
@@ -705,18 +715,17 @@ func (ls *LedgerState) handleEventChainsyncAwaitReply(evt event.Event) {
 			"connection_id", e.ConnectionId.String(),
 			"error", err,
 		)
-		if ls.config.EventBus != nil {
-			ls.config.EventBus.Publish(
+		pending.add(
+			ls.config.EventBus,
+			LedgerErrorEventType,
+			event.NewEvent(
 				LedgerErrorEventType,
-				event.NewEvent(
-					LedgerErrorEventType,
-					LedgerErrorEvent{
-						Error:     err,
-						Operation: "await_reply_blockfetch",
-					},
-				),
-			)
-		}
+				LedgerErrorEvent{
+					Error:     err,
+					Operation: "await_reply_blockfetch",
+				},
+			),
+		)
 	}
 }
 
--- a/ledger/publish_under_lock_test.go
+++ b/ledger/publish_under_lock_test.go
@@ -1,0 +1,197 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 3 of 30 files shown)


chrisguiney:
Thanks — worked through all 14 findings from both reviewers. Pushed as `ddbc7b86`.

**Two were real bugs**, both in `run-tests.sh`:

- `DEVNET_ACCELERATED` was never cleared on the canonical path, so a stale value in the environment made a canonical run execute the accelerated scenario against the canonical-timing network — whose budget it is deliberately designed *not* to meet, failing the whole suite.
- The success path did `rm -rf` on whatever `DEVNET_ARTIFACT_DIR` pointed at, including a caller- or CI-supplied directory. A passing run could destroy unrelated files. Now only a directory the script created via `mktemp` is removed.

**My own lock-ordering guard was too narrow.** `PublishAsync` waits for room in the shared async queue rather than dropping, and that queue is drained by a worker pool that runs subscriber handlers — so a handler needing the publisher's mutex closes the same cycle a full subscriber buffer does. My comment claiming it was exempt was simply wrong. The guard and `ARCHITECTURE.md` now cover `Publish`, `PublishBlocking` and `PublishAsync` alike. No current call site violates it.

**Also fixed:** the separate-value `-test.run` form; the README's stale "uid 100" and missing `--accelerated` in the documented options; dead `ObservedChain.Node()` and `MaxBlockNumber`; the redundant `h.Slot > boundary`; the `EpochMargin` comment that claimed the margin exceeds the 4k/f freeze window when it is deliberately much shorter; and the over-broad chown regex.

`lastProducer` became `interruptionVictim` and now **asserts** what it previously assumed — that the producer it stops is not the one txpump submits to. If the endpoint list is ever reordered, that fails loudly rather than silently removing the mempool traffic the phase depends on.

**Two I did not apply, with reasons:**

*Parity guards are uncacheable because they read non-Go files* — not correct. Go's test cache tracks file reads through `testlog`. Verified directly: with a cached PASS in place, editing the `Dockerfile` made the test re-run and **fail** without `-count=1`.

*Extract a shared spec resolver for `run-tests.sh` / `start.sh` / `docker-compose.yml`* — the underlying risk is real, but routing two filenames through a shared abstraction across a shell/compose/Go boundary costs more than it saves. I addressed the risk directly instead: `TestScriptsAndComposeAgreeOnSpecFiles` fails if the scripts ever disagree on which accelerated spec to use. (Relatedly, `start.sh` not exporting `DEVNET_ACCELERATED` is intentional — it does not run tests, and prints the exact command including that variable.)

Every guard added or changed here was verified to fail when the defect it covers is reintroduced. Full `go test ./...` green, and the accelerated scenario passes in 1m54s.

chrisguiney:
@cubic-dev-ai the `handleChainSwitchEvent` finding was correct and I'd missed it — thank you. Pushed as `68284bd8`.

You were right on both counts. `handleChainSwitchEvent` holds `chainsyncMutex` for its whole body and calls `requestChainsyncResync`, which published `ChainsyncResyncEventType` inline — and that event's subscriber calls `RecoverAfterLocalRollback`, which re-takes the same mutex. Same deadlock, a path the first commit did not reach. And the invariant test *was* giving false confidence: it is intra-procedural, so a lock holder reaching a publish through a helper went unreported.

**Fixed:** `requestChainsyncResync` now takes an optional queue. `handleChainSwitchEvent` passes its own, flushed after the unlock. Every other caller passes `nil` and keeps publishing inline, so no path this change does not fix is disturbed.

**It is worse than one path.** I wrote a transitive analysis over `chainsync.go` to find the real extent: **five** helpers reachable from a lock holder can publish that event inline —

```
handleEventChainsyncBlockHeader
handleEventChainsyncRollback
handoffPipelineOnSwitchLocked
replayBufferedHeaderEvents
startQueuedBlockfetchLocked
```

**I have deliberately not converted the other four here.** Doing so means threading a queue through roughly ten functions in the hottest consensus path, and a helper's own flush is no use there because a *parent* still holds the lock — so it cannot be done incrementally per-helper. That is a systemic refactor deserving its own PR with its own DevNet and soak validation, not a drive-by on a test-infrastructure branch. I started it, watched it sprawl, and reverted rather than land something half-threaded.

**Instead the exposure is now enumerated rather than assumed.** `TestChainsyncResyncPublishPathsUnderLock` pins that set and fails in *both* directions: a new path is a new deadlock and must be converted rather than appended to the list, and a path that disappears must be removed so the list cannot rot into overstating the problem. The set is computed from the code — my hand-written first guess at it was wrong, which is rather the point.

Happy to open a follow-up issue for the remaining four if that is useful.

Validated: full `go test ./...`, `ledger` under `-race`, the accelerated scenario (2m21s), and the full canonical DevNet suite end to end with zero `event delivery stalled` warnings.

chrisguiney:
@cubic-dev-ai fair caveat — here are the two pieces you asked to see.

**Flush ordering.** The queue's `defer` is registered *before* the mutex is taken, so Go's LIFO order runs the unlock first and the flush second:

```go
func (ls *LedgerState) handleChainSwitchEvent(evt event.Event) {
	e, ok := evt.Data.(chainselection.ChainSwitchEvent)
	if !ok {
		return
	}
	// Registered before the mutex is taken so defer's LIFO order runs it
	// after the unlock. See pendingPublishes.
	var pending pendingPublishes
	defer pending.flush()
	...
	ls.chainsyncMutex.Lock()
	defer ls.chainsyncMutex.Unlock()
```

Both `return` paths and any panic unwind hit the same ordering, so there is no path where the flush runs with the mutex still held. That ordering is load-bearing enough that reversing the two lines silently reintroduces the deadlock — worth a reviewer's eye.

**The `nil` path:**

```go
func (p *pendingPublishes) add(
	bus *event.EventBus,
	eventType event.EventType,
	evt event.Event,
) {
	if bus == nil {
		return
	}
	if p == nil {
		bus.Publish(eventType, evt)
		return
	}
	p.events = append(p.events, pendingPublish{...})
}

func (p *pendingPublishes) flush() {
	for _, pub := range p.events {
		pub.bus.Publish(pub.eventType, pub.evt)
	}
	p.events = nil
}
```

A nil receiver publishes inline, so the eight untouched call sites keep byte-for-byte today's behaviour; the `bus == nil` guard preserves the nil-EventBus checks that were at the call sites before. `flush` clears the queue, so it is not re-entrant — covered by `TestPendingPublishesFlushesInOrder`, along with ordering, and `TestPendingPublishesBreaksLockCycle` reproduces the full cycle (a subscriber needing the publisher's lock, buffer overrun) and asserts the queued form completes where the inline form hangs.

I verified that reproduction is real rather than vacuous: publishing directly in that same shape deadlocks, which is how I confirmed the diagnosis in the first place.

On the follow-up issue — agreed it deserves a home. I'll leave that call to @chrisguiney rather than filing it unprompted.

chrisguiney:
@cubic-dev-ai both points land on things already in place — checked rather than assumed.

**1. The ordering assertion already exists**, and in the shape you describe. `TestPendingPublishesBreaksLockCycle` uses the `defer` form rather than calling `flush()` eagerly:

```go
var pending pendingPublishes
defer pending.flush() // runs after the unlock below
chainsyncMutex.Lock()
defer chainsyncMutex.Unlock()
```

So a "cleanup" that moves the defer next to the lock is caught. Verified by doing exactly that:

```
--- FAIL: TestPendingPublishesBreaksLockCycle (10.00s)
    pending_publish_test.go:134: publisher deadlocked while holding the lock
```

That's the reversal failing on the real deadlock, not on queue-logic bookkeeping — which is what you were asking for.

**2. This one I think is a misread** — the `bus == nil` guard is already the first statement, ahead of the receiver check:

```go
func (p *pendingPublishes) add(
	bus *event.EventBus,
	eventType event.EventType,
	evt event.Event,
) {
	if bus == nil {
		return
	}
	if p == nil {
		bus.Publish(eventType, evt)
		return
	}
	p.events = append(p.events, pendingPublish{...})
```

A nil bus returns before either branch, so the non-nil-queue path cannot append one and `flush` can't reach `pub.bus.Publish` on nil. The one-line fix you suggest is the line that's there. Happy to be shown otherwise if you're reading a different revision — `68284bd8` is current.

Thanks for the careful pass on this one; the `handleChainSwitchEvent` catch was the substantive find of the review and it was a real deadlock I'd have shipped.

chrisguiney:
Opened #3167 for the remaining paths.

One correction to what I said earlier in this thread: it is **five** helpers still to convert, not four. I had been subtracting `handleChainSwitchEvent`, but that is a lock *holder*, never in the helper list — so fixing it removed none of them. The issue has the accurate set, with the lock-holding entry point that reaches each:

| Helper | Reached under the lock by |
|---|---|
| `handleEventChainsyncRollback` | `handleEventChainsync` |
| `handleEventChainsyncBlockHeader` | `handleEventChainsync` |
| `handoffPipelineOnSwitchLocked` | `handleChainSwitchEvent` |
| `replayBufferedHeaderEvents` | `replayBufferedHeadersAsync` |
| `startQueuedBlockfetchLocked` | `RecoverAfterLocalRollback`, `handleEventChainsyncAwaitReply` |

That matches `knownResyncPublishPathsUnderLock` in this PR, which is computed from the code — the list was right, my prose about it was not.

ledger/chainsync.go
```diff
@@ -2408,6 +2428,7 @@ func (ls *LedgerState) handleEventChainsyncBlockHeader(e ChainsyncEvent) error {
 				ls.requestChainsyncResync(
 					e.ConnectionId,
 					event.ChainsyncResyncReasonPersistentFork,
+					nil,
```
cryptodj413, line 2436: This is one of several call sites (also lines ~2649/2660/2787/2836, 3162, 5088/5105/5163) where `requestChainsyncResync` is passed `nil` for `pending`, so the resulting `ChainsyncResyncEventType` publish happens immediately via `bus.Publish` rather than being queued. This specific call runs inside `handleEventChainsyncBlockHeader`, invoked from `handleEventChainsync` while `ls.chainsyncMutex` is held (this exact persistent-fork path fires after `headerMismatchResyncThreshold` consecutive mismatches, which the surrounding comments describe as an expected occurrence in a multi-producer network). If the `ChainsyncResyncEventType` subscriber's buffer is already backed up behind an earlier `RecoverAfterLocalRollback` call blocked on the same mutex, this publish blocks forever — reproducing the exact wedge this PR sets out to fix, just via a different path. This is disclosed in the commit message and pinned as a known set in `ledger/publish_under_lock_test.go`'s `knownResyncPublishPathsUnderLock`, but it means the fix is only partial: production forks/blockfetch-retry storms can still hit this class of deadlock.

ledger/publish_under_lock_test.go
```diff
@@ -0,0 +1,368 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// guardedMutexes are the LedgerState locks that an EventBus subscriber
+// handler acquires. Publishing while holding one of these can deadlock the
+// node, so no function may do both.
+var guardedMutexes = []string{"chainsyncMutex"}
```
cryptodj413: `guardedMutexes` only lists `chainsyncMutex`, but `RecoverAfterLocalRollback` (the dangerous subscriber) also nests `chainsyncBlockfetchMutex` inside its `chainsyncMutex` hold via `startQueuedBlockfetchLocked`. Inline publishes that hold only `chainsyncBlockfetchMutex` — e.g. `handleEventBlockfetchBatchDone`'s `requestChainsyncResync(..., nil)` calls, reached from `handleEventBlockfetch` which locks only `chainsyncBlockfetchMutex` — aren't covered by this guard or by `TestChainsyncResyncPublishPathsUnderLock` (which only tracks `chainsyncMutex` holders). That leaves a second-mutex variant of the same deadlock class with no test coverage at all, plausible under the same connection-churn conditions described in the PR body.

ARCHITECTURE.md
```diff
@@ -928,6 +928,25 @@ All event types follow the `subsystem.snake_case_name` convention.
   worker pool means it also delays unrelated async event types. Consumers of
   `Subscribe` channels must drain for the life of the subscription and
   `Unsubscribe` when they stop
+- **Never `Publish`, `PublishAsync`, or `PublishBlocking` while holding a lock
```
cryptodj413, line 931: This states the publish-under-lock invariant as an absolute ("Never Publish..."), but `ledger/chainsync.go` still has multiple call paths that violate it for `chainsyncMutex`, deliberately left unconverted per the commit message and pinned in `ledger/publish_under_lock_test.go`'s `knownResyncPublishPathsUnderLock`. Worth a pointer here to that known-exceptions list so a future reader doesn't assume the invariant is already fully enforced in this package.

cryptodj413 (commented):
Reviewed the three commits (devnet uid fix, ledger publish-under-lock fix, accelerated devnet scenario). The uid/gid fix and its guards look solid, and the accelerated scenario's event-driven design (broadcaster/Await pattern, shared timeline, accelerated-vs-canonical budget tests) is well constructed with no time.Sleep in the sync path. The one substantive concern is that the ledger deadlock fix is intentionally incomplete — flagging the specific remaining exposure below, which the PR author already discloses and pins with a test, but which is still a live risk worth weighing before merge.

chrisguiney:
@cryptodj413 thanks — the second comment was the substantive one and it was my error. Pushed as `83ba2287`.

**On `guardedMutexes` (the nested-mutex variant).** You're right, and the reasoning failure was mine earlier in this branch: I checked whether `handleEventBlockfetchBatchDone` ran under `chainsyncMutex`, found it didn't, and concluded it was safe to leave publishing inline. `RecoverAfterLocalRollback` takes `chainsyncMutex` and then nests `chainsyncBlockfetchMutex` inside it via `startQueuedBlockfetchLocked`, so holding *either* is enough to close the cycle. Guarding only the outer one left every blockfetch-lock-only path uncovered.

Adding `chainsyncBlockfetchMutex` to `guardedMutexes` immediately failed on **three live violations**, all inline publishes in `handleEventBlockfetch`, which holds that mutex for its whole body. Converted the same way as the earlier fixes — queue registered before the lock, flushed after the unlock.

Extending the transitive check to both mutexes also widened the tracked set from five helpers to **seven**, surfacing three that were invisible while only the outer mutex was considered:

```
handleBlockfetchTimeoutLocked
handleEventBlockfetchBatchDone
restartQueuedBlockfetchAfterForkLocked
```

(`handleEventChainsyncBlockHeader` drops off — it is itself a lock holder now, not a helper.)

**On `ARCHITECTURE.md`.** Fair. It asserted the rule absolutely while the exceptions exist. It now names both mutexes and points at `knownResyncPublishPathsUnderLock`, framed as target-state-plus-exception-list rather than an invariant already holding.

**On `chainsync.go:2431` (the partial fix).** No code change, but your framing raises the severity and I've updated #3167 accordingly — the persistent-fork path firing after `headerMismatchResyncThreshold` is an *expected* occurrence in a multi-producer network, not an edge case, which makes this more than a theoretical latent risk. #3167 now also covers the blockfetch-mutex variant.

**One thing your review indirectly caught in my own test code.** Validating this change failed the accelerated scenario — not from the ledger change (the canonical suite and conformance both passed on that binary) but because `runOutage` ended at "the nodes agree". Agreement says they converged, not that the network resumed *producing*, so the relay was being stopped seconds after a producer restart and the resulting stall was attributed to the relay outage. It now requires forward progress before returning. Two clean runs since (2m01s, 2m06s).

Validated: full `go test ./...`, `ledger` under `-race`, canonical DevNet suite, conformance against `cardano-node`, and the accelerated scenario twice.

chrisguiney:
@cubic-dev-ai one applied, one declined with evidence.

**P3 (failure message) — fixed** in `05a31733`. You're right that it named only `chainsyncMutex` after the blockfetch mutex joined the guarded set, which would have sent anyone hitting a blockfetch-path failure looking at the wrong lock. It is now built from `guardedMutexes` so it cannot drift again as that list changes.

**P1 (`handleEventBlockfetchBatchDone`) — this is the tracked systemic work, and "thread `&pending` through that path" is exactly the trap I'm avoiding.** The helper is already listed in `knownResyncPublishPathsUnderLock`, so the guard reports it rather than hiding it. Here is why it is not a one-level thread:

`handleEventBlockfetchBatchDone` reaches `requestChainsyncResync` by **three** routes:

1. directly
2. via `noteBlockfetchRangeUnavailable` → `requestChainsyncResync`
3. via `startQueuedBlockfetchLocked` → `noteBlockfetchRangeUnavailable` → `requestChainsyncResync`

Threading only route 1 — which is what passing `&pending` at that call site achieves — leaves routes 2 and 3 still publishing inline under `chainsyncBlockfetchMutex`, while the function *looks* converted. That is strictly worse than the current state, where the guard lists it as known-exposed.

Converting it properly means threading through `startQueuedBlockfetchLocked`, which has **eight** callers:

```
RecoverAfterLocalRollback              handleEventBlockfetchBatchDone
handleEventChainsyncBlockHeader        handleEventChainsyncAwaitReply
restartQueuedBlockfetchAfterForkLocked handoffPipelineOnSwitchLocked
handleBlockfetchTimeoutLocked          startQueuedBlockfetchLocked (self)
```

Most of those are either the other tracked helpers or `chainsyncMutex` holders in their own right. So converting this one path *is* the full conversion — it cannot be done in isolation, which is precisely why #3167 exists and why it has to be one atomic change with its own DevNet and soak validation rather than landing on a test-infrastructure branch.

Happy to be argued out of this if you see a way to make route 2 and 3 safe without pulling in those eight callers — but I would rather ship a guard that honestly reports the exposure than a conversion that only looks complete.

chrisguiney:
@cubic-dev-ai your closing suggestion was worth more than either of the findings — I tested it instead of assuming, and the guard had a real blind spot. Fixed in `beac52be`.

**What I did.** Injected exactly the case you described: a clean lock holder (`handleConnectionClosedEvent`) calling a brand-new helper, which calls another new helper, which calls `requestChainsyncResync(..., nil)`. Two hops, entirely indirect.

**The guard passed it.** No failure at all.

**Why, and it is worse than a missing-transitive-edge.** The blind spot arrived *with the fix it was meant to protect*. `requestChainsyncResync` was converted to queue through `pendingPublishes`, so it stopped calling `EventBus.Publish` directly — and the analyzer, which recognised publishers by that call, quietly stopped counting it. Every route reaching a resync publish through that helper went invisible, including all **eight** of its `nil`-passing callers, which publish immediately because `add` falls through to `Publish` on a nil receiver.

So the honest statement is that my guard had been under-reporting since the commit that converted the helper. The reported set happened not to change — those helpers were already listed for their own direct publishes — but the *coverage* had shrunk, and a new indirect path would not have been caught.

**Fix:** a call handing `nil` to a queue-taking resync helper now marks the caller as publishing inline, which is what actually happens at runtime. Verified both directions:

- the injected two-hop path is now reported (`syntheticOuterHelper` appears in the diff)
- the real baseline is unchanged, so this widens what the guard *sees*, not what it *claims*

That is a good argument for pinning the enumeration in a test at all: the list stayed still while the thing computing it lost its grip, and only an adversarial probe surfaced that. Worth carrying into #3167 — I have noted it there as a check before the issue closes.

chrisguiney:
@cubic-dev-ai applied in `e2e4317f` — with a correction to how I'd have framed it.

The detection now collects each queue-taking helper's `*pendingPublishes` parameter position from its declaration (its own pass, since a call can appear before the declaration it targets) and examines only that position at call sites.

Verified in both directions against injected paths:

- nil handed to the **queue** parameter → still reported
- a call passing a real queue alongside an **unrelated** nil → not reported

Baseline unchanged.

**The correction:** I went looking for the live misfire and could not produce one. Your wording was careful — "if a queue-taking helper is *later* called with a non-nil queue and an unrelated nil argument" — and that is exactly right: no current helper has another nilable parameter, so nothing misfires today. My first attempt to demonstrate the old form failing didn't, because the intermediate helper in my construction never named the event type, so the old logic never reached the nil check at all. Building a case that genuinely breaks it requires adding a nilable parameter to `requestChainsyncResync`, which I stopped short of contriving.

So this is hardening against a plausible future signature change, not the repair of a live bug, and the commit message says so. Worth being exact given how much of this thread has turned on the difference between "the guard reports the right set" and "the guard is computing it for the right reason" — the previous finding was the latter failing while the former looked fine.

chrisguiney:
@cubic-dev-ai I went to verify the ordering question and found something worse. Fixed in `5ab5e8e4`.

**Ordering was fine** — the collection pass is over all declarations, and a helper declared at the end of the file called from the top is resolved correctly.

**But the probe failed for a different reason, and the case that proves it is damning: reverting this branch's own fix — changing `handleChainSwitchEvent`'s `&pending` back to `nil` — passed both guards silently.** The guards did not protect the fix they were written for.

Each missed it for its own reason, and neither was wrong in isolation:

- the intra-procedural check recognised a publish only as a direct `EventBus.Publish` call, and a nil-queue call is not one
- the transitive check skips lock holders, because it enumerates the *helpers* they reach rather than the holders themselves

A lock holder that publishes by handing `nil` to a queue-taking helper fell exactly between them.

Two things were needed to close it. The nil-queue test has to run *before* the existing two-level selector assertion — `ls.method(...)` has a plain identifier receiver, not a selector like `ls.config.EventBus`, so my first attempt at this fix silently never ran. And the queue parameter positions have to be collected across every file before any body is walked, which is what your ordering question was really about.

**It surfaced a genuine pre-existing violation:** `handleEventChainsyncBlockHeader` holds `chainsyncBlockfetchMutex` and hands `nil` to `requestChainsyncResync`. It cannot be converted on its own — its caller `handleEventChainsync` holds `chainsyncMutex`, so flushing on its own return would still publish under the parent's lock. It joins the tracked set as `knownNilQueuePublishersUnderLock`, checked in both directions so the entry cannot outlive the problem.

Verified by regressing the real fix, confirming it is now reported, then restoring it.

This is the third time in this thread that a guard looked right and wasn't, and each was found by attacking it rather than reading it. Worth stating plainly: the exposure this PR ships with is what the two lists say it is, and their value depends entirely on the analysis behind them being probed rather than trusted. I've noted that on #3167 as a condition for closing it.

chrisguiney:
@cubic-dev-ai you're right, and that correction is on me — it's the most misleading thing I've written in this thread.

In `83ba2287` I noted that `handleEventChainsyncBlockHeader` "drops off the list because it is itself a lock holder now rather than a helper." That reads as a resolution. It wasn't one. It never stopped publishing inline under `chainsyncBlockfetchMutex` — it switched from a direct `EventBus.Publish` to the `requestChainsyncResync(..., nil)` route, which made it invisible to the transitive analysis (which enumerates helpers, not holders) while remaining every bit as exposed. Restructured, not fixed, and my note made a reclassification sound like progress.

That is worth more than the individual finding. A list shrinking is the signal this whole scheme relies on — I said as much earlier about #3167, that a shrink there would be the expected signal when the conversion lands. This is a concrete case of an entry leaving a list for a reason that had nothing to do with the exposure going away, and I described it as though it had.

I've corrected the record on #3167 so it tracks as continuously exposed since before this branch, not as something the new list introduced.

Your framing of the root cause is also more precise than mine. Two internally sound analyses — "a publish is a direct `EventBus.Publish` call" and "a publish is reaching a tracked helper" — with a hole in their *union*, because a lock holder using the nil-queue route is an inline publish under the first model and an unenumerated holder under the second. Neither check was wrong on its own terms, which is why reading them didn't find it and attacking them did.

chrisguiney:
@cubic-dev-ai applied in `09455ae2` — and it's the right call for a reason specific to this file's history.

Both guards now call the same `queueParamPositions` and `nilQueueCall`. Net 33 lines removed.

Two copies of the same detection is precisely the fragility this thread has been about: an analysis that looks correct and silently stops matching. Widening one copy to recognise a differently spelled nil or receiver would have left the other guarding nothing, and — as we have now seen three times — that failure is invisible from the pinned lists, because they keep reporting whatever the analysis can still see.

Re-probed after the refactor rather than trusting the green run, since changing shared detection is exactly when a guard stops guarding:

- reverting `handleChainSwitchEvent`'s `&pending` to `nil` → still reported
- a two-hop indirect path from a clean lock holder → still reported

Pinned baselines unchanged, full `ledger` suite green.

That is the fourth guard defect this review has surfaced, and the pattern is consistent: every one was found by attacking the analysis, none by reading it. I'd rather that be said plainly on the PR than discovered later — the two exception lists are only worth what the probes behind them are worth.

chrisguiney:
@cubic-dev-ai correct on the count, and I'll take it — `09455ae2` prevents future divergence rather than fixing a found gap. Three real guard defects (transitive blind spot, nil-queue blind spot, and the lock-holder-hands-nil hole between the two analyses), plus one consolidation. Overcounting my own findings is the wrong direction to be imprecise in, particularly on a branch whose argument is that the lists are only as good as the analysis behind them.

Nothing further from me on this thread — the substance has converged. Summary of where it lands for whoever reviews:

- **Two real deadlocks fixed**: `handleChainSwitchEvent`, and three inline publishes in `handleEventBlockfetch`.
- **Two exception lists** covering seven helpers and one lock holder, all tracked in #3167, each bidirectional so entries cannot outlive the problem.
- **One correction to the record**: `handleEventChainsyncBlockHeader` was never resolved by this branch — it was reclassified, and I initially described that as though it were progress.
- **The scenario this PR is nominally about** is unchanged since it last passed accelerated (twice), canonical, and conformance.

Worth a human deciding whether the guard work belongs here or alongside #3167 — five of the eleven commits are now analysis hardening rather than DevNet scenario.

cryptodj413 (approved):

Issue #2988, opened 2026-07-28
security: protect destructive Bark DatabaseService RPCs


wolf31o2:
## Summary

PR #2923 mounts destructive database lifecycle RPCs on Bark without authentication or authorization. When Bark and databaseLifecycle.snapshotDir are enabled, a reachable client can list snapshots, restore the live database, truncate the chain, delete snapshots, or cancel operations.

This is not covered by dingo #1653 or blinklabs-io/bark#17: those track authentication for a future stop/restart LifecycleService. The closed DatabaseService issue blinklabs-io/bark#16 did not specify an authentication model.

## Evidence

- node.go mounts DatabaseService whenever barkPort is enabled and databaseLifecycle.snapshotDir is non-empty.
- node.go does not pass the configured bind address into BarkConfig.Host.
- bark.NewBark defaults Host to 0.0.0.0.
- bark registers the DatabaseService handler without an authentication or authorization interceptor.
- TLS is optional and currently provides server transport encryption only; it does not authenticate or authorize clients.

## Impact

Any client with network reachability to the Bark listener can invoke destructive operator actions. Restore and truncate can replace persistent state, interrupt node availability, and alter the active chain tip.

## Acceptance criteria

- Destructive DatabaseService methods require an authenticated operator identity.
- Authorization distinguishes destructive operator methods from read-only archive and status methods.
- The protection is enforced at the Bark transport boundary and cannot be bypassed by choosing Connect, gRPC, or gRPC-Web framing.
- Bark honors an explicitly configured restricted bind address; enabling DatabaseService must not silently expose it on every interface.
- TLS configuration documents whether mutual TLS is the authentication mechanism; if another mechanism is used, TLS remains required for credentials in transit.
- Anonymous requests to Restore, Truncate, DeleteSnapshot, CreateSnapshot, VerifySnapshot, and CancelOperation are rejected in wire-level tests.
- Deployment documentation explains the trust boundary and secure configuration.

## Related

- PR #2923
- #1651
- #1653
- blinklabs-io/bark#16
- blinklabs-io/bark#17

PR #3168, opened 2026-08-12
feat(api): add uniform TLS/auth surface and top-level security defaults

### LINKED ISSUE #2996: api: provide uniform TLS and authentication across API plugins
## Problem

The built-in API plugins expose different transport-security surfaces:

- UTxORPC receives the process-level TLS certificate and key, but has no in-process authentication.
- Blockfrost and Mesh receive bind/CORS settings, but cannot enable TLS or authentication in-process.
- Operators therefore cannot apply one security policy consistently across `plugins.api.blockfrost`, `plugins.api.mesh`, and `plugins.api.utxorpc`.

This is especially risky in API storage mode because these listeners expose query and transaction-submission functionality. Reverse-proxy deployment must remain supported, but it should not be the only way to secure two of the three plugin APIs.

## Scope

Define and implement one TLS/auth contract for every provider under `plugins.api.*`:

- Give Blockfrost, Mesh, and UTxORPC the same TLS configuration surface (server certificate and private key).
- Give all three the same authentication configuration surface, with at least explicit disabled and token-based modes.
- Use one credential-verification implementation; adapt only the wire transport between HTTP headers and Connect/gRPC metadata.
- Define whether Blockfrost-compatible `project_id` is an alias/adapter for the shared token mechanism, and document the decision.
- Keep provider dependency DTOs narrow. Composition and effective-policy resolution belong in node/config wiring, not API domain packages.
- Do not introduce package-init registration or process-global provider configuration.

## Acceptance criteria

- [ ] Each built-in API provider supports the same named `tls` and `auth` settings in its provider config.
- [ ] Each API can be exercised in plaintext/no-auth, TLS/no-auth, and TLS/auth configurations.
- [ ] A partial certificate/key pair fails validation before the listener starts.
- [ ] Missing or invalid credentials fail closed with protocol-appropriate results (`401` for HTTP and `Unauthenticated` for Connect/gRPC).
- [ ] HTTP and Connect/gRPC credential locations and Blockfrost `project_id` compatibility are documented.
- [ ] CORS preflight and authentication ordering are defined and tested.
- [ ] Authentication material is never written to logs, errors, metrics labels, or effective-config output.
- [ ] Unit tests cover policy parsing and credential verification; listener-level tests cover accepted/rejected requests and TLS handshakes for Blockfrost, Mesh, and UTxORPC.
- [ ] Existing reverse-proxy/no-auth deployments have a documented compatibility path.
- [ ] `ARCHITECTURE.md`, `README.md`, and `dingo.yaml.example` describe the common API security boundary.

### LINKED ISSUE #2998: config: add top-level API security defaults with per-plugin overrides
## Problem

API-wide settings currently live as unrelated root fields (`bindAddr`, `corsAllowedOrigins`, `tlsCertFilePath`, and `tlsKeyFilePath`) and are injected unevenly during node composition. Provider selections under `plugins.api.*.config` currently expose only a port.

After #2996 gives every API provider the same TLS/auth surface, operators need one top-level policy that configures all selected API plugins while retaining field-level provider overrides.

## Target behavior

Add a top-level API configuration section with shared TLS and authentication defaults. The exact field names should follow repository conventions, but the shape should be equivalent to:

```yaml
api:
  tls:
    mode: server
    certFilePath: /run/secrets/api.crt
    keyFilePath: /run/secrets/api.key
  auth:
    mode: token
    tokenFilePath: /run/secrets/api-token

plugins:
  api:
    mesh:
      provider: builtin
      config:
        port: 8080
        auth:
          mode: none
    blockfrost:
      provider: builtin
      config:
        tls:
          certFilePath: /run/secrets/blockfrost.crt
          keyFilePath: /run/secrets/blockfrost.key
```

The top-level policy supplies defaults to every selected API provider. Explicit fields in `plugins.api.<capability>.config` override those defaults for that provider only.

## Precedence and compatibility

- Resolve API policy field-by-field rather than replacing the complete `tls` or `auth` object when one nested field is overridden.
- Preserve the repository configuration-source precedence: CLI/provider selection, generic plugin environment, YAML, then provider defaults. Scope inheritance is a separate layer: explicit provider field, top-level API field, then provider default.
- Represent field presence explicitly so an operator can disable an inherited TLS/auth policy for one provider; empty strings and zero values must not ambiguously mean both unset and disabled.
- Provide a compatibility/deprecation path for the existing root TLS fields. Do not silently change existing deployments on upgrade.
- Decide and document whether existing shared bind/CORS fields move under the same `api` section or remain compatibility aliases; avoid maintaining two competing sources of truth.

## Acceptance criteria

- [ ] One top-level config section can configure TLS and authentication for Blockfrost, Mesh, and UTxORPC together.
- [ ] Any TLS/auth field can be overridden for one plugin under `plugins.api.<name>.config`.
- [ ] A provider can explicitly disable an inherited TLS or authentication policy.
- [ ] The effective merge is deterministic and independent of map iteration order.
- [ ] Validation runs on the effective per-provider policy and reports the full provider config path in errors.
- [ ] Listener collision validation uses each provider effective bind address if bind configuration is included in the reconciliation.
- [ ] CLI, environment, and YAML tests cover top-level defaults, partial nested overrides, explicit disable, and invalid merged certificate/key pairs.
- [ ] Effective-config/debug logging redacts authentication material.
- [ ] `dingo.yaml.example` and `README.md` include shared-default and provider-override examples.
- [ ] `ARCHITECTURE.md` documents that config resolution occurs at composition and providers receive resolved, instance-owned settings.

cryptodj413:
Closes #2996
Closes #2998

Adds a uniform per-plugin TLS and authentication surface for the Blockfrost, Mesh, and UTxORPC API plugins (#2996), backed by a shared `internal/tlsutil`, `internal/apiauth` (HTTP + Connect middleware), and `internal/apiconfig` merge layer.

On top of that, adds top-level `api.tls` / `api.auth` security-defaults config (#2998) that merges into each plugin's config with field-by-field override and explicit-disable semantics, while preserving the existing legacy TLS/auth fields on UTxORPC for backward compatibility.

## Acceptance criteria coverage
- Uniform TLS config surface (cert/key/client-CA/mTLS) available identically on Blockfrost, Mesh, and UTxORPC plugin configs.
- Uniform auth surface (bearer token / basic auth, as applicable per transport) available identically across the three plugins.
- Top-level `api.tls` and `api.auth` act as defaults; any field set explicitly at the plugin level overrides the corresponding top-level default.
- A plugin can explicitly disable TLS or auth even when top-level defaults enable it.
- Legacy UTxORPC-specific TLS/auth fields keep working unchanged for existing configs.

## Deliberately out of scope (flagging for reviewers)
- No per-provider `bindAddr` override under `plugins.api.<name>.config` — bind address remains governed by existing top-level/plugin wiring, unrelated to the TLS/auth surface being added here.
- Provider-level `tls`/`auth` overrides are YAML-only; they are not addressable through the generic env-var flattening mechanism, since these are nested struct fields rather than flat scalars.

---
## Summary by cubic
Add a uniform TLS/auth configuration for all built-in API providers and top-level security defaults, now applied at runtime by the node. Backwards compatible with existing UTxO RPC configs.

- **New Features**
  - Unified per-plugin TLS/auth for `blockfrost`, `mesh`, and `utxorpc` via `internal/apiconfig`, `internal/apiauth` (HTTP + Connect), and `internal/tlsutil`.
  - Top-level `api.tls` and `api.auth` defaults merge into each `plugins.api.<name>.config` with field-level overrides and explicit per-provider disable; merged policies are validated before listeners start.
  - Token auth supported with a shared verifier; Blockfrost `project_id` header is accepted as a bearer alias; new CLI flags `--api-tls-*` and `--api-auth-*`.

- **Bug Fixes**
  - Wired `api.tls`/`api.auth` into the real startup path (`dingo.WithAPIConfig`) so defaults take effect at runtime.
  - Improved auth handling: `Token`/`TokenFilePath` are treated as one source, and bearer/`project_id` values are trimmed before comparison.
  - `tlsutil.ConfigureServerTLS` now replaces (not accumulates) certificates to avoid stale cert selection.
  - Suppressed env auto-binding of inline tokens to avoid leaking secrets; prefer token files or YAML.

<sup>Written for commit 57a124a851bb08fc643ec512ef1dd040b64e2a91. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3168?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added shared TLS and authentication configuration for UTxO RPC, Blockfrost, and Mesh APIs.
  * Supports bearer-token authentication, Blockfrost project-ID credentials, token files, and provider-specific overrides.
  * Added HTTPS support with configurable certificates and keys.
  * CORS preflight requests continue to work without authentication.
* **Bug Fixes**
  * Invalid security settings are now detected during startup with clearer validation errors.
  * Authentication secrets are excluded from logs.
* **Documentation**
  * Expanded configuration and security guidance, including defaults and precedence rules.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -17,7 +17,12 @@ none of them. Failures unwind providers in reverse order. Normal shutdown
 orders APIs, mempool, ledger/database, then storage.
 Database receives provider-owned stores through `database.Stores`.
 API providers are resolved for lifecycle only because node composition has no
-in-process consumer of their concrete server values.
+in-process consumer of their concrete server values. Each API provider's
+TLS/authentication policy is resolved the same way: node composition merges
+the shared `api.tls`/`api.auth` default into that provider's own config
+before resolving it (see "API security" under External Interfaces), so a
+provider always receives an already-resolved, instance-owned policy and
+never has to know a shared, top-level default exists.
 
 Command and bootstrap composition that opens a standalone database uses
 `internal/plugins.OpenDatabase`. Its return contract keeps ownership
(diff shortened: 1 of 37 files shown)


cryptodj413:
Pushed `57a124a8` addressing review feedback from `coderabbitai[bot]` and `cubic-dev-ai[bot]` on `304e5450`.

**Fixed**

- `internal/node/node.go`: the real CLI startup path never forwarded `cfg.API` into `dingo.WithAPIConfig`, so the top-level `api.tls`/`api.auth` defaults were inert on a deployed node despite being fully plumbed and tested in isolation. Found independently while validating the wiring end to end; not flagged by either bot review before this push, but cubic's `config.go:983` finding on this same commit calls out the identical gap — now wired.
- `internal/apiconfig/apiconfig.go` `MergeAuth`: `Token`/`TokenFilePath` are now merged as a single credential-source unit, so a provider overriding only `tokenFilePath` at a narrower scope no longer produces a merged policy with both fields set and failing `Resolve`'s mutual-exclusivity check (cubic `apiconfig.go:162`, coderabbit inline comment — both flagged this independently).
- `internal/apiauth/apiauth.go`: Bearer credential is now trimmed before comparison, so extra internal whitespace in an otherwise-valid header doesn't get rejected (cubic `apiauth.go:96`).
- `internal/tlsutil/tls.go` `ConfigureServerTLS`: assigns the freshly loaded keypair to `server.TLSConfig.Certificates` instead of appending, so a reused `*http.Server` or pre-populated `TLSConfig` can't leave a stale cert in the SNI selection set alongside the new one (cubic `tls.go:76`).
- `internal/apiconfig/apiconfig.go` `AuthPolicy.Token`: added `ignored:"true"` — verified the finding was real (envconfig auto-derives `CARDANO_API_AUTH_TOKEN` for any untagged exported field it walks), so the doc-comment's claim of "no environment variable binding" was previously false. Now enforced (cubic `apiconfig.go:105`).
- `internal/config/validate.go`: moved the `validateAPIMode` doc comment off of `validatePort` and gave `validatePort` its own comment back (cubic `validate.go:760`).
- `dingo.yaml.example`: added `mode: server` to the per-provider Blockfrost TLS override example — uncommenting only that block previously resolved to `disabled` since the shared `api.tls` section supplying `mode: server` is itself commented out (cubic `dingo.yaml.example:58`).
- `api/utxorpc/tls_auth_test.go` `TestUtxorpcCORSPreflightBypassesAuth`: now goes through the shared `startOnFreePortWithCORS` retry helper instead of a bare bind, removing the flaky lost-port-race path the other cases already guard against (cubic `tls_auth_test.go:60`).
- Removed dead test helpers `strPtr` (`node_api_security_test.go`) and `ptr` (`internal/apiconfig/apiconfig_test.go`) — every call site already used `new(...)` directly (cubic, both flagged independently).
- `ARCHITECTURE.md`/`README.md`: reworded the provider-config handoff description (providers decode merged `TLSPolicy`/`AuthPolicy` and call `Resolve` themselves during construction, not a pre-resolved policy handed down) and the legacy `tlsCertFilePath`/`tlsKeyFilePath` fallback wording (field-by-field fallback, not "both sections fully unset") (cubic `ARCHITECTURE.md:24`, coderabbit inline comments, cubic `README.md:323`).
- Test-helper dedup (coderabbit nitpick): `insecureHTTPClient`, `bindAttempts`/`freePort` are now shared exported helpers (`InsecureHTTPClient`, `BindAttempts`, `FreePort`) in `internal/test/testutil/tls.go`, called from all three providers' `tls_auth_test.go` instead of three/two local copies.
- EventBus cleanup (coderabbit nitpick, `api/utxorpc/provider_test.go`): test-owned buses now use `t.Cleanup(bus.Close)` instead of `bus.Stop` (which restarts the worker pool rather than shutting it down), and a single bus is built once and reused across the resolve retry loop.

**Dismissed as out of scope**

The three client-cert/mTLS findings (cubic — `api/utxorpc/utxorpc.go:559`, `api/mesh/mesh.go:294`, `api/blockfrost/blockfrost.go:401`) are not addressed here: the new `apiconfig.TLSPolicy` type has no client-CA/mTLS field at all in this PR's scope (#2996/#2998 is server-TLS-and-bearer-auth only). Client-certificate verification is unrelated pre-existing machinery under Bark (#2988) and would be a separate, deliberately scoped follow-up if wanted, not a fix to what this PR actually implements.

**Judgment call**

Kept authentication uniform across all routes on every provider, including health/reflection endpoints (coderabbit nitpick on `api/utxorpc/utxorpc.go:296`, "authenticating the health endpoint breaks unauthenticated liveness probes"), rather than exempting health checks — Blockfrost already applies auth uniformly to its own health route in this same PR, and uniform coverage across all three providers is the deliberate cross-provider design here, not an oversight in one of them. The operator-facing tradeoff (a load balancer/k8s probe hitting an authenticated listener needs a credential, or the listener needs to sit behind a proxy that owns health-check exemption) and mitigation options are documented in README.md and ARCHITECTURE.md rather than special-cased in code.

**Skipped, not done — flagged for follow-up if wanted**

The `ResolvePolicies` shared-helper extraction across `api/blockfrost/provider.go`, `api/mesh/provider.go`, `api/utxorpc/provider.go` (coderabbit nitpick) — the three call sites differ only in a config-path prefix and an error prefix string; judged not worth the churn/risk in this review-fix pass given the three are otherwise correct and covered by tests. Happy to extract it in a follow-up if a reviewer wants it now instead.

**Verification run before this push:** `golangci-lint run ./...` (0 issues), `nilaway ./...` (only a pre-existing, unrelated finding in `bark/auth_test.go` present on the branch before this change too), `modernize ./...` (no findings in any file touched by this diff), `make import-boundaries`, `make gorm-check` (both clean), `golines` (no reformatting needed on the touched Go files), and `go test -race ./...` scoped to every touched package plus the top-level `node_api_security_test.go` suite — all passing.

chrisguiney (approved):

Issue #3165, opened 2026-08-11
Praos VRF leader-eligibility check can permanently wedge a node: canonical block rejected because leadership stake omits reward-account balances (bypass is prototype-only)


chadaibbhot-git:
## Summary

On a non-prototype network (preview), a block producer and its relay both wedged
permanently, looping `block processing failed, restarting pipeline`, because the
inbound-block Praos VRF leader-eligibility check rejected a **canonical** block. The
producing pool's leadership stake as computed by the node is slightly **below** the
stake-derived VRF threshold, but the block is valid and adopted by the rest of the
network (an independent Haskell `cardano-node` relay and koios both accept it). The
node's leadership stake distribution is **UTxO-only and omits reward-account
balances** (acknowledged in a code comment), so for a pool whose VRF output sits near
the eligibility boundary the small omission flips the check to "not eligible" and the
node rejects a block it should accept — with no forward progress.

The mitigation for this (`SkipLeaderStakeThresholdCheck`) is wired only to the
prototype-trust profile, so preview/mainnet nodes have no protection.

## Environment

- **Dingo** `v0.69.0` (commit `496e923edc08b59e496378181af7739f6d96aa18`).
- **Network** preview (Praos). Two nodes affected: a block producer and a relay.
- Cross-check reference: an independent Haskell `cardano-node` relay on the same
  network, plus koios.

## Symptom (exact log, repeated every ~2s indefinitely)

```
level=WARN msg="block processing failed, restarting pipeline"
error="process block batch: deferred block header verification failed at slot 119799023:
block header verification rejected at slot 119799023:
producer pool 1b9db9dd1099057105072fd00adb5cc44e0eb60d9e565fa8c8646b78
VRF leader value exceeds stake-derived threshold
(pool stake: 22185731049285, total stake: 1643035932299353,
epoch: 1386, snapshot_epoch: 1385, snapshot_type: mark)"
```

Last successful adoption was slot `119799019`; the node then froze at that tip while
the network advanced ~1,230 blocks / ~36,000 slots (~10h) ahead. A container restart
re-wedged at the **identical** block (deterministic; DB-level state).

## The rejected block is canonical

- Producing pool: `pool1rwwmnhgsnyzhzpg89lgq4k6uc38qadsdnet9l2xgv34hsjxzj2d`
  (`1b9db9dd…`).
- The Haskell reference relay and koios both accept the block; the chain advanced
[...]

Issue #2876, opened 2026-07-19
VRF leader-eligibility check rejects canonical block despite correct stake (post #2843)


chadaibbhot-git:
## Environment
- network: preview
- roles: block producer + relay (both Dingo, `docker.io` image)
- version: v0.65.1 (which includes the #2843 mark-reward-stake fix)
- storage: sqlite metadata; badger blob store; fresh Mithril bootstrap on v0.65.0 at 2026-07-15T22:41Z (in-place-upgraded to v0.65.1 at 2026-07-16T18:30Z)
- upstream peers on relay: IOG bootstrap + community
- Haskell relay control node at same tip

## Summary
Same VRF wedge class as #2813 recurred on v0.65.1 at slot 117,744,396 (mid-epoch 1362, ~2026-07-18T18:46Z). Both Dingo nodes rejected a canonical block from pool `pool13yccxnmfdp9l4j655s07aetwg7aa0p4pfjddrr2xlzedq2pv222`. Unlike #2813, the mark stake data itself is now correct (matches Koios within 0.1%), so the bug is not in the stake computation this time — it is in the leader-eligibility / threshold check path.

### BP + relay
Both nodes reject canonical header at slot 117,744,396 with:
```
level=ERROR msg="failed to handle block" component=ledger
error="block header crypto verification failed:
  block header verification rejected at slot 117744396:
  producer pool 8931834f69684bfacb54a41feee56e47bbd786a14c9ad18d46f8b2d0
  VRF leader value exceeds stake-derived threshold
  (pool stake: 21256151898192, total stake: 3267403053048802,
   epoch: 1362, snapshot_epoch: 1361, snapshot_type: mark)"
```

- Fresh Mithril bootstrap on v0.65.0 (2026-07-15T22:41Z) survived boundary 1359→1360 (skipped-eligibility-check safety warn, expected on mid-epoch Mithril mark) and boundary 1360→1361 (real check via #2843, passed).
- The wedge then fired at slot 117,744,396 during epoch 1362 (~2026-07-18T18:46Z), ~68 h after bootstrap.
- Relay recovered on its own via **container restart alone** (`docker restart preview-relay` at 2026-07-18T19:00Z) — no re-Mithril, no fresh bootstrap. It resumed chain-extend one slot later at 117,745,490 and is now at canonical tip.
[...]

Issue #2813, opened 2026-07-13
Mark stake snapshot: live capture undercounts per-pool stake ~10% — canonical blocks fail VRF threshold two epochs later


chadaibbhot-git:
**Environment**
- dingo **v0.62.0** and **v0.63.0** (v0.63.0 includes #2789)
- network **preview** (magic 2, Conway era), 1-second slots
- Two Dingo nodes fresh-Mithril-bootstrapped from clean state; a Haskell reference node running in parallel as a control

**Summary**
Every live-captured mark row in `pool_stake_snapshot` on both Dingo nodes comes in ~10% below the Mithril-imported baseline for the same node. Two epochs after the first bad mark, headers from canonical pools whose *real* stake would clear the VRF-leader threshold get rejected with `VRF leader value exceeds stake-derived threshold` against the wrong stored `total_stake`. The BP eventually rolls back and enters a rejection loop; a passive relay chainsync-stalls. Reproduces across v0.62.0 and v0.63.0, and on both node roles. The Haskell reference node validated the exact same canonical headers with no issue.

**Symptom — BP**
Forged 66 blocks across ~60 hours on v0.62.0, then GDD/LoE rolled the local tip back to the fork point at slot 117,072,065. BP could not re-accept the next canonical header at slot 117,072,069 (pool `aa5a33e022780281587cc7140f5e0b507b1e52104e65eb7321682acf`, hash `545ec6ba72f810345216b940c983beb87e3c5e748ab1ff0e1cc7725ef463c410`, currently 3336 confirmations on Koios). ~7,600 rejection loops in six hours until a fresh Mithril rebootstrap. `dingo_forge_tip_gap_slots` sat at ~101,000.

```
level=ERROR msg="failed to handle block header"
  error="block header crypto verification failed: block header verification rejected at slot 117072069:
    producer pool aa5a33e022780281587cc7140f5e0b507b1e52104e65eb7321682acf
    VRF leader value exceeds stake-derived threshold
    (pool stake: 10785209621729, total stake: 1480821363057324,
     epoch: 1355, snapshot_epoch: 1354, snapshot_type: mark)"
  slot=117072069
  hash=545ec6ba72f810345216b940c983beb87e3c5e748ab1ff0e1cc7725ef463c410
```

**Symptom — passive relay**
Chainsync stalled during the same window and hit the local operator's 3-restart-cap-per-24h several times on 2026-07-11. After an 80-minute stall it recovered on a restart and rejoined canonical.

[...]

PR #3173, opened 2026-08-13
fix: two ways a node silently stops following the chain

chrisguiney:
Two independent ways a Dingo node stops following the chain while looking healthy, plus the instrumentation that makes the second one diagnosable.

Neither fix changes what Dingo accepts. The validity predicate stays bit-identical to the reference node's throughout, which is the constraint that matters if Dingo is to carry majority stake.

# Part 1 — NtC connection churn wedges chainsync

A node silently stops following the chain while continuing to forge. TCP and keepalive stay healthy, no block is rejected, nothing errors — every subsequent chain extension is `source=forged`, and the node builds its own fork indefinitely. Only a restart clears it.

`connmanager` published `ConnectionClosedEventType` for **every** connection, including node-to-client. All five subscribers do node-to-node peer management — chain selection, peer governance, chainsync client state, the mempool consumer set — and the payload carries only a connection ID and an error, so none of them can recognise an NtC connection and skip it.

A local client that reconnects in a tight loop therefore parks publishers on that topic faster than they drain. In a 40-minute DevNet run, `txpump` cycling an NtC connection ~90×/sec produced **7,732,651** `event delivery stalled: subscriber not draining` warnings, all on this one topic. Once delivery is wedged, the NtN chainsync recovery those events drive never runs again.

The connection manager already knows which connections are NtC — `RemoveConnection` checks `!info.isNtC` twenty lines below the publish. The publish just didn't filter.

This is very likely the chainsync stall behind the field reports in #2813 and #2876, both of which describe a relay that recovers on **container restart alone**. A restart is exactly what clears the wedged subscriber.

**`fix(connmanager)`** — gate the publish on the existing `isNtC` flag. Verified safe for mempool consumers: both `txsubmissionClientStart` call sites are NtN paths (outbound, and full-duplex inbound), so no NtC connection ever registers one and suppressing the event leaks nothing.

**`fix(event)`** — the stall warning was rate-limited *per delivery*, which bounds it to one line per interval per blocked publisher. That is no bound at all when the publishers are what is numerous. Moved the limit to the subscriber, and added the count that was actually missing:

```
blocked_publishers=40  suppressed_warnings=39
```

That number is what separates ordinary backpressure from a subscriber that has stopped draining; previously the warning said only that *something* was stalled, seven million times.

### Verified against a live repro

| | Before | After |
|---|---|---|
| Last block received from network | slot **1325**, then never again | still ingesting at **1444** |
| `event delivery stalled` warnings | **7,732,651** | **0** |
| dingo vs cardano-node block gap | 18–25, growing | **0** |
| Chain extensions | `source=forged` only | mixed `forged` + `chainsync` |

The node that previously went deaf at slot 1325 and forged its own fork for 40 minutes now tracks `cardano-node` in exact lockstep, 127 slots past that point and still ingesting.

# Part 2 — the Praos leader-eligibility wedge (#3165)

A block producer and its relay wedge permanently on a **canonical** block, looping `block processing failed, restarting pipeline` every ~2s and re-wedging identically across restarts.

**The filed diagnosis does not hold.** Leadership stake has included reward-account balances since #2843. A DevNet comparison against `cardano-node` shows Dingo's `mark[E-1]` matching the reference `stakeSet` **bit-exactly** on both pools, cleanly distinguishable from `stakeMark` and `stakeGo`, with reward amounts matching to the lovelace. Snapshot-epoch selection, threshold arithmetic, the CPraos/TPraos split and the `applyRUpd`-before-SNAP ordering all check out, and Mithril-seeded rows are already covered by `isMithrilImportedMarkSnapshot`.

So the stake is not the bug — but the wedge is real, and it is a liveness bug that turns out to be entirely separable from the predicate.

**`fix(ledger)` — recover from the rejection instead of replaying it.**

A deferred stateful header check rejects a block the chain store *already holds*, so returning a bare error restarts the pipeline onto the identical block, which fails identically, forever. Transaction validation already distinguishes deterministic from transient — `txValidationError` routes into rewind-based recovery that drops the offending chain and lets chainsync re-deliver. Header validation had no type, and therefore no path:

| Failure | Handling before |
|---|---|
| tx validation | `txValidationError` → rewind / at-tip recovery |
| deferred header validation | plain `fmt.Errorf` → **re-reads the same block forever** |

`headerValidationError` closes that asymmetry. The block is **still rejected**; only the destination of the rejection changes — rewind the primary chain and ledger to the last good tip, ask chainsync to re-deliver, let chain selection offer a different candidate.

If the network genuinely offers no other chain, the node still cannot advance. That is the correct outcome for a node that believes a block is invalid, and the same one the reference node reaches — it is now reported rather than endured silently. Recovery declines rather than claiming success when there is nothing to rewind or when the rewind would cross the Mithril trust boundary; a false "recovered" would send the pipeline straight back into the same block believing it had been handled.

**`feat(ledger)` — instrument the margin.** `verify_header.go` had no metrics at all. Every eligibility decision now records `(threshold − leaderValue) / threshold` in `dingo_ledger_leader_threshold_margin`; rejections increment `dingo_ledger_leader_threshold_rejections_total` and carry the margin in the message.

Recording *every* decision rather than only failures is the point. A threshold comparison is a knife edge: a relative stake error of ε flips a decision with probability ~ε per block. ε is a deterministic function of the chain, so a nonzero ε is a **bug with a location**, not a property to accommodate — and this metric is how it gets located. A stake error clusters decisions near zero; a derivation bug produces margins that are not marginal at all.

**`feat(ledger)` — bound the retry.** The existing no-progress backoff capped at 2s, which for a deterministic failure means retrying forever at that rate. Past `noProgressStuckThreshold` consecutive no-progress restarts the pipeline is treated as stuck: the wait escalates beyond the transient ceiling, the transition logs once at ERROR, and `dingo_ledger_pipeline_stuck` / `dingo_ledger_pipeline_no_progress_restarts` expose it to monitoring. Both reset the moment the tip advances.

### A rejected approach, recorded

An earlier revision made the threshold check opt-in, so a failure warned and accepted. **That was wrong and has been dropped from the branch** rather than left as a commit plus revert. It fixed the loop by accepting blocks the reference node rejects, which is the one thing a node aspiring to majority stake cannot do: any divergence in the accept/reject predicate is a chain split waiting to happen, whether it takes the form of a tolerance band or a blanket warn-only. ε being *small* is not the property that matters; ε being *zero* is.

`SkipLeaderStakeThresholdCheck` therefore keeps its existing prototype-only scope, untouched.

# Testing

**DevNet `--conformance` suite — all 7 scenarios pass** against this branch, Dingo running beside `cardano-node`:

```
--- PASS: TestBasicBlockForging (12.02s)
--- PASS: TestDingoChainAdvances (12.01s)
--- PASS: TestChainGrowthRate (128.03s)
--- PASS: TestRelayPropagation (0.02s)
--- PASS: TestSustainedConsensus (102.06s)
--- PASS: TestCardanoProducerChainAdvances (10.02s)
--- PASS: TestEpochBoundaryConsensus (436.04s)
ok  internal/test/devnet/scenarios  701.021s      (plus 44/44 helper unit tests)
txpump submitted=3462 rejected=22 error=0
```

`TestSustainedConsensus` and `TestEpochBoundaryConsensus` are the load-bearing ones: the two implementations stay in agreement across sampling intervals and across an epoch boundary. That is the check that the validity predicate is still identical — had it shifted, they would diverge rather than track.

**The new code paths stayed inert, which is the actual test.** On a healthy network none of them should ever fire, and none did: `deferred header validation rejected…` **0**, `exceeds stake-derived threshold` **0**, `ledger pipeline stuck` **0**, `event delivery stalled` **0**. Recovery machinery that fires on a healthy network would be worse than the bug it fixes.

**The margin metric validates against theory.** 147 decisions, mean margin **0.504**. For an eligible block the VRF leader value is uniform in `[0, T)`, so `(T − v)/T` is uniform on `(0, 1]` with mean 0.5 — the metric measures what it claims. The histogram records nothing below `le=0.01`: no decision came near the boundary, which is what ε ≈ 0 looks like on a chain Dingo derives itself.

Also:

- `go test -race ./...` — exit 0, all packages, no races
- Tests written first and confirmed failing. **Three needed a second pass to become meaningful**, noted in their commit messages rather than quietly fixed:
  - the connmanager test initially passed pre-fix because `newUnstartedConnection` returns a zero-value `ConnectionId`, so two connections in one manager collide and the second replaces the first — a replaced connection's close is *already* suppressed as stale
  - the rate-limit test initially passed because 40 publishers emit their first warnings near-simultaneously; gating on three observed repeats makes the counts diverge unambiguously (**40** warnings before, **3** after, against 40 publishers either way)
  - the margin test pins its sign against `consensus.IsVRFOutputBelowThresholdWithMode`'s own verdict across both consensus modes, because a margin derived differently from the decision would misreport exactly the cases it exists to measure
- `golangci-lint`, `nilaway` and `modernize` were **not run** — not installed on the development machine. Worth a pass before merge

# Docs

`ARCHITECTURE.md` updated in four areas: the `connmanager.conn_closed` entry in the EventBus topic table; two EventBus-features bullets (the node-to-node-only contract, and the per-subscriber bound on the stall warning); the leader-eligibility section, describing the margin metric; and the pipeline retry section, describing both the deferred-header recovery path and stuck detection. The EventBus edits merge with bullets added upstream in the meantime — those describe the `ledger.conn_closed` republish coupling that *propagates* such a stall, and the new bullet points at it as the mechanism by which an NtC flood wedges NtN recovery.

`DATABASE.md` checked and not affected: no schema, query, or storage surface changes.

# Root cause of #3165, and what closes it

The stake shortfall is not drift. Against koios for epoch 1386 the pool runs
0.1203% low and the network total 0.0782% low, which partly cancel to a sigma
error of 0.042%. In reward terms the pool is short 2.96 epochs of its own
rewards and the network total 3.07 epochs of network rewards — two independent
quantities landing on the same integer, which points at reward *crediting*
rather than reward computation.

Applying a reward round at the boundary into N needs this node's own
`RewardSnapshot` for N-3 and `RewardAdaPots` for N-1. After a Mithril bootstrap
those epochs predate the import, so the first rounds find no inputs and are
skipped — while the reference credits them regardless. Nothing backfills them,
so reward balances stay short for the life of the database, and the leadership
stake with them, since `reward_live_stake` sums UTxO plus reward.

That explains every observation: all three field reports of this wedge (#2813,
#2876, #3165) were Mithril-bootstrapped; the shortfall is an integer number of
epochs; it is permanent rather than growing; and DevNet, synced from genesis
with full history, measures the stake bit-exact against `cardano-node`.

Both skips logged at **Debug**, which is why the condition survived three field
investigations unseen. They are Warn now, name the consequence, and increment
`dingo_ledger_skipped_stake_reward_rounds_total`.

The import then seeds what those rounds need. `RewardAdaPots` comes from the
snapshot's own treasury, reserves and fee pot — the fees were sitting undecoded
in `UTxOState[2]`, and matter because the reward pot is incentives plus fees, so
a zero-fee row would credit the round at the wrong amount rather than visibly
not running it. The per-credential basis comes from mark/set/go, each of which
carries one epoch's stake, delegations and pool parameters, and which line up
with exactly the three epochs a bootstrapped node cannot otherwise compute.

Every derived basis is gated before it is written, and one that does not
reconcile is dropped rather than persisted. That gate is mandatory, not
defensive: the ledger reads this basis through a path
(`calculateStakeRewardApplication` → `rewardCalculatorSnapshot`) that returns an
error rather than skipping, so an unusable row would turn a missing reward round
into a node that cannot cross an epoch boundary at all.

## The parameters were in the snapshots all along

The seeding above still dropped every epoch on a real network, and the reason
was a pool that held stake in the `go` or `set` snapshot and retired before the
snapshot's own epoch. Such a pool is absent from cert state and from the
current pool distribution, so no registration in an imported database describes
it; its delegators' stake could not be attributed, and rather than seed a basis
that understates every other pool's share, the gate dropped that epoch
entirely.

Each snapshot's pool records carry the registration parameters — VRF key,
pledge, cost, margin, reward account and owners — alongside the pool's stake.
`parsePoolDistrEntry` read only the VRF key, and its comment stated the rest
was "not present in the legacy order". They are present, at an offset it did
not look at. `parseSnapshotPoolParams` reads them, and the seeding prefers them
over registration history: they are the parameters in force during the epoch
being seeded, and they describe the pool set as it stood then — retired pools
included, which nothing else in an imported database can do.

Measured on preview:

| Snapshot | Before | After |
|---|---|---|
| `mark` | 1 pool, 1,218,574,660 lovelace unattributed → **rejected** | 716 pools, 79,725 delegators seeded |
| `set` | 1 pool, 1,218,574,660 → **rejected** | 716 pools seeded |
| `go` | 2 pools, 19,782,849,212 → **rejected** | **717** pools seeded |

`go` seeds one pool more than `mark` — the retired pool no registration could
describe.

The layout is in no published CDDL; the on-chain `pool_params` is a different,
9-field shape led by the operator. So every field is validated on read, and a
record that does not match degrades to the VRF-only reading rather than mapping
a field onto the wrong parameter — pledge, cost and margin feed reward
arithmetic directly, and a misread would be credited rather than visibly
refused.

It is cross-checked against cert state, which decodes full pool parameters
through an unrelated path, over **715 preview pools** — 597 with a non-zero
margin, 630 with a non-zero pledge, 654 with owners. The DevNet fixture could
not have established the mapping on its own: two pools, zero pledge, zero cost,
zero margin, no owners, so every distinguishing field sits at its zero value. A
first mapping that passed cleanly on that fixture was wrong, and preview caught
it.

One asymmetry is deliberate. A snapshot's owner set lists only the owners
holding stake in it, not every owner the registration names — 62 of 715 pools
differ, always in that direction, and every omitted owner holds no stake in
that snapshot, so owner stake and the reward split that depends on it are
unaffected. The tests assert that justification rather than raw equality.

# What this does not do

**It does not repair an already-affected node.** The shortfall is baked into
`account.reward` in the reporter's database, and seeding runs only at import.
An affected node needs a re-bootstrap on a build containing this; nothing here
backfills one in place.

**DevNet has not been run against these changes.** Container network egress was
unavailable in the environment this was developed in — image builds fail on
`DeadlineExceeded`, and a bare `docker run debian apt-get update` hangs
independently of any build. `internal/test/devnet/run-tests.sh` is the
outstanding gate for the chain-selection commits in Part 2.


chrisguiney:
@cubic-dev-ai review

cryptodj413 (approved):

Issue #1894, opened 2026-04-13
Block processing pipeline: parallel decode and validate


wolf31o2:
Sync speed is ~3,905 slots/sec on preview, target is >10,000. Parallelizing block decode and validation is the main lever. The gouroboros pipeline/ package provides the foundation.

Phases:
1. Pipeline plumbing -- create pipeline in LedgerState, submit blocks, apply via chain.AddBlock. 2-worker decode, no validation yet.
2. Wire apply stage -- chain add, error handling, event publishing, sync progress logging.
3. Enable VRF/KES validation -- Eta0Provider for epoch nonce, validate workers.
4. Metrics integration -- pipeline stats counters and gauges.
5. Rollback coordination -- drain in-flight blocks before rollback.

Acceptance:
- 2x+ bulk sync throughput
- Tests pass with pipeline enabled and disabled
- Chain tip matches both modes after identical sync
- No goroutine leaks
- Rollback behavior identical with and without pipeline

Roadmap: Q3 2026

wolf31o2:
Related: once decode and validation run in parallel here, the validation
workers will read the read-mostly consensus state (current era, protocol
params, epoch, tip) concurrently on every block. Today those reads go through
the ledger RWMutex, which stops scaling and degrades under a concurrent writer.
Moving that state to copy-on-write atomic.Pointer snapshots (#2601) makes the
reads lock-free, so this pipeline does not reintroduce read-side lock
contention as worker count grows.

PR #3178, opened 2026-08-14
feat(ledger): parallel block decode pipeline (phase 1 of #1894)

cryptodj413:
Part of #1894.

Phase 1 of the block-processing pipeline roadmap: parallel block decode. Blocks read back from the primary chain in `ledgerReadChainIterator` are now decoded by a small worker pool (2 workers, vendored `gouroboros/pipeline`) instead of one at a time inline, then re-sequenced back into submission order before handing off to `ledgerProcessBlocksFromSource` exactly as before. Validation and apply are untouched in this phase -- only decode scheduling changes.

Off by default behind new `LedgerStateConfig.BlockPipelineEnabled`, threaded through config/flags/`dingo.yaml.example`/`node.go` (CLI > env > YAML > defaults, per this repo's config priority).

Two issues surfaced and were fixed during review:
- `BlockPipelineEnabled` wasn't threaded into `node_lifecycle.go`'s live restore/truncate config rebuild, which would have silently reset the flag to off after any live restore or truncate. Fixed.
- Enabling the pipeline on the Musashi prototype network could silently and permanently stall chain replay, since the vendored `gouroboros/pipeline.DecodeStage` has no hook for dingo's Musashi/Leios-extended-header decode fallback. Now hard-guarded at `configValidate()` with an explicit rejection error, plus a regression test (`TestBlockPipelineRejectedOnMusashi`).

`ARCHITECTURE.md` updated with a new "Block Processing Pipeline" section.

Deferred to follow-up PRs (remaining phases of #1894):
- Phase 2: apply-stage wiring
- Phase 3: VRF/KES validation in the pipeline
- Phase 4: metrics
- Phase 5: rollback drain coordination

**Not yet run:** DevNet (`internal/test/devnet/run-tests.sh`) has not been exercised against this change. Recommended before final merge given it touches the live block-decode path, but flagging honestly that it hasn't been done yet.

---
## Summary by cubic
Parallelizes block decoding in the chainsync replay loop behind an opt‑in flag and fixes a cross‑attempt race in the shared decode pipeline. Previously decoding was serial and retry attempts could submit concurrently; now a 2‑worker `gouroboros/pipeline` decodes in parallel with result re‑sequencing, and the retry loop waits for the prior reader goroutine to fully exit to prevent block misattribution.

- Adds `LedgerStateConfig.BlockPipelineEnabled` (YAML `blockPipelineEnabled`, env `DINGO_BLOCK_PIPELINE_ENABLED`, CLI `--block-pipeline-enabled`; default false). Startup rejects this on Musashi; see `configValidate()` and `TestBlockPipelineRejectedOnMusashi`.
- Starts the pipeline in `LedgerState.Start` and stops it in `Close` with a drain timeout. Skipped entirely under `ManualBlockProcessing`. Decode submit/drain is all‑or‑nothing once started.
- Refactors decode into `decodeReadChainBatch`; preserves input order and whole‑batch failure semantics. Adds per‑iteration cancellation checks in the gather loop.
- Introduces `runLedgerReadChainAttempt` to synchronize retries: the loop blocks until the retiring reader goroutine exits before starting the next, ensuring a single submitter to the shared pipeline. See `TestLedgerProcessBlocksRetryDoesNotMixBlocksAcrossAttempts`.
- Threads the flag through config and node lifecycles (`internal/config`, `node.go`, `node_lifecycle.go`). Updates `dingo.yaml.example` and `ARCHITECTURE.md`. Adds tests for serial vs pipeline parity, rollback trim, goroutine lifecycle.

- Rollout: no migration required. Enable the flag on non‑Musashi networks. Do not enable on Musashi (startup error). Recommend running DevNet tests before merge since this touches the live decode path.

<sup>Written for commit 57a63aa73915223c5af284e3a5b591573064d60c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3178?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an optional parallel block-decoding pipeline for chainsync replay.
  * Added configuration through YAML, environment variables, and the `--block-pipeline-enabled` CLI flag.
  * Preserved block ordering, validation, application, and rollback behavior.
* **Bug Fixes**
  * Added clear validation preventing use of the pipeline with Musashi configurations.
  * Decode failures now discard affected batches safely.
* **Documentation**
  * Documented pipeline behavior, limitations, and current rollout scope.


arepala-uml:
## Confirmed race: block-decode pipeline can mix up blocks across retries

**What's going on:** `ls.blockPipeline` is one shared decode pipeline that lives for the whole node run. But the retry loop in `ledgerProcessBlocks` (`ledger/state.go:3304-3320`) cancels the old read goroutine and immediately starts a new one — it doesn't wait for the old one to actually stop first.

**Why that's risky:** This restart path fires when a rollback races the read iterator into staleness (`errStaleChainIterator`, `ledger/replay_recovery.go:34-49`) — a normal occurrence during chain following near the tip, not some contrived corner case. Right before returning that error, the code calls `completeReadResult()` (`ledger/state.go` ~4199-4234), which wakes the *old* goroutine back up to keep decoding — before `cancel()` even runs. Its gather loop (`ledger/state.go:3138-3173`) has no cancellation check, so it can keep submitting blocks to the shared pipeline for a bit.

**The actual bug:** `decodeReadChainBatch` (`ledger/state.go:3230-3286`) has no defense against a second concurrent submitter — it just grabs the next N results off the shared pipeline and assumes they're its own. Verified directly with a stress test: two goroutines submitting disjoint, uniquely-identifiable batches to the same `blockPipeline` concurrently got each other's decoded blocks back — **~19,800 individual blocks misattributed across 200 iterations** (roughly half of all blocks decoded), with contamination showing up in nearly every run. This isn't a narrow timing coincidence, it reproduces almost every time two submitters overlap.

**Why it's new:** The retry loop itself (`ledger/state.go:3304-3320`) isn't changed by this PR — it was always safe because old code had no shared state between retries. The new `blockPipeline` field is what introduces the shared state that makes this unsafe.

**Impact:** Only affects nodes with `BlockPipelineEnabled: true` (off by default). Could apply the wrong block after a reorg. Won't show up under `go test -race` — it's a logic/ordering race, not a memory race.

**Suggested fix:** Make `ledgerProcessBlocks` wait for the old goroutine to fully exit before starting a new one (e.g. a `done` channel per attempt, awaited right after `cancel()` on line 3319).

Other options: tag each submission with an attempt ID and drop stale results, or give each attempt its own `BlockPipeline` instance instead of sharing one.

Either way, worth adding a test that forces `errStaleChainIterator` with the pipeline enabled and checks the new batch never contains blocks from the old attempt.

arepala-uml (approved):

Issue #3180, opened 2026-08-16
devnet: TestSustainedConsensus fails when the chain stalls mid-run and then recovers


wolf31o2:
## Summary

On an otherwise clean canonical DevNet run against origin/main at 2f7eb2ae, TestSustainedConsensus failed while the other 50 tests in the run passed. The chain stopped advancing at slot 241 / block 77 partway through the test, stayed there for the remainder of that checkpoint's budget, and then resumed on its own. The next scenario in the same run, TestEpochBoundaryConsensus, started from that same tip and passed.

This is not the genesis-wait budget problem fixed in #3166 or the NtC conn_closed wedge fixed in #3173. Both fixes are present in this build, TestSustainedConsensus already calls WaitForChainStart, and TestEpochBoundaryConsensus and TestStakeAddressInfoQueries both pass in this run where they previously failed.

## What was observed

The test takes a base tip and checks consensus at three relative checkpoints.

```
base tip slot 152
checking consensus at slot 172 ... consensus verified at slot 172
checking consensus at slot 202 ... consensus verified at slot 202
checking consensus at slot 252 ... FAIL
    no node reached slot 252 within 1m52.5s
```

The third checkpoint needed 11 more slots than the chain had reached. For the rest of the 112.5s budget every node reported the same frozen tip:

```
WaitForSlot: dingo-1     at slot 241, block 77
WaitForSlot: dingo-2     at slot 241, block 77
WaitForSlot: dingo-3     at slot 241, block 77
WaitForSlot: dingo-relay at slot 241, block 77
```

All four agreed, so this is not a fork or a single-node divergence. The whole network stopped extending.

It then recovered without intervention. The next scenario observed the identical tip and went on to pass:

```
WaitForChainStart: dingo-1 forged through slot 241, block 77
waiting for slot 700 (past epoch boundary at slot 500), timeout 12m5s
--- PASS: TestEpochBoundaryConsensus (394.06s)
```

Reaching slot 700 from 241 in 394s means the tip caught up its lag, so block production resumed at full rate after the stall.

## Timeout arithmetic, for reference

[...]

PR #3204, opened 2026-08-17
refactor(ouroboros): take dependencies at construction

### LINKED ISSUE #2275: ouroboros: remove post-construction temporal coupling
## Summary

`ouroboros.Ouroboros` relies on public fields being populated after construction.

## Audit Details

Finding ID: M14, MEDIUM architecture finding.

Affected file: `ouroboros/ouroboros.go`.

The audit found temporal coupling in the `Ouroboros` struct. Dependencies such as `ConnManager`, `PeerGov`, `ChainsyncState`, `EventBus`, `Mempool`, and `LedgerState` are public fields that are populated by `Node.Run()` after construction.

Impact: the type can exist in an invalid state, and method safety depends on startup ordering outside the constructor. That makes tests and future wiring changes fragile.

## What Needs to Happen

1. Identify required vs optional dependencies.
2. Move required dependencies into the constructor or validated option functions.
3. Make fields private where possible.
4. Return clear errors when optional protocol features are used without their dependency.
5. Add constructor tests for missing dependency cases.

## Acceptance Criteria

- Required dependencies are validated at construction time or explicit setup time.
- Methods no longer assume public fields were set out-of-band without checks.

chrisguiney:
Closes #2275.

## Why

`ouroboros.Ouroboros` held its dependencies in exported fields that the node populated one at a time after construction, so the type could exist in a partially wired state whose only symptom was a nil dereference at first protocol use.

That was not hypothetical. `HandleOutboundConnEvent` guarded `PeerGov` but then called `o.ConnManager.GetConnectionById` unguarded, and `GetConnectionById` locks a mutex on its receiver — so a nil `ConnManager` is a dereference, not a nil result. The node subscribes that handler to the EventBus before `ConnManager` is assigned, so the window is reachable. The first test written for this change panicked at `ouroboros/ouroboros.go:638`:

```
panic: runtime error: invalid memory address or nil pointer dereference
  connmanager.(*ConnectionManager).GetConnectionById  connection_manager.go:865
  ouroboros.(*Ouroboros).HandleOutboundConnEvent      ouroboros.go:638
```

## The cycle is only apparent

Three components consume callbacks from `Ouroboros`, which looks like it forces post-construction assignment:

| Component | Needs from Ouroboros |
| --- | --- |
| `ledger.NewLedgerState` | `EndorserBlockTxsByHash`, `FetchEndorserBlockByPoint`, `BlockfetchClientRequestRange` |
| `connmanager.NewConnectionManager` | `ConfigureListeners`, `OutboundConnOpts` |
| `peergov.NewPeerGovernor` | `RequestPeersFromPeer` |

None of those callbacks runs at construction time, so the cycle breaks two ways:

- **Closures, not method values.** Callbacks handed to ledger and peergov resolve `n.ouroboros` when they fire.
- **Lazy providers.** New `ConnectionManagerConfig.ListenersProvider` / `OutboundConnOptsProvider` are invoked on first use — at `Start`, or the first outbound dial — rather than at construction.

`NewOuroboros` therefore runs **last**, takes all six dependencies, validates them, and returns an error naming the first missing one. The dependency fields are unexported and never reassigned, so a partially wired instance is unrepresentable. Accessors expose them read-only.

## Immutability forces a live-restore rebuild

The snapshot/restore path previously *retained* `Ouroboros` and swapped its fields. An immutable instance has to be replaced instead — and replacement was impossible before this PR. Confirmed empirically:

```
CONFIRMED: second construction without Close panics:
  duplicate metrics collector registration attempted
```

So `Close` releases the two things that outlive an instance:

- **Prometheus collectors.** `promauto` panics on duplicate registration. Collectors are tracked automatically as they are registered, not by a hand-maintained list — with 11 today, a hand-written list would fall behind and the next added metric would silently reintroduce the panic on the restore path.
- **EventBus subscriptions.** `subscribeLeiosAnnouncementRetries` and `SubscribeChainsyncResync` previously discarded their subscription IDs. A leaked subscription is handled once per restore cycle, forever — the same leak `node_leios.go` captures IDs to avoid.

`Close` is idempotent, so `Run`'s deferred shutdown and the restore teardown can both call it.

## The closure rule is load-bearing

A method value anywhere in either composition path pins that component to the pre-restore instance. The node then keeps running and **silently stops syncing**.

This is not theoretical — it happened twice while writing this PR, and the compiler was no help: method values on a stale pointer compile cleanly, so `go build` and `go vet` were green while the node was broken. `TestLiveTruncateUnderRealForgingAndNetworking` caught it by hanging (5.7s before, 300s+ after). Instrumentation showed the replacement never receiving its listener config:

```
DBG rebuilt ouroboros=0x...9540308     <- replacement
DBG ConfigureListeners o=0x...cf6008   <- but the OLD instance configured the listeners
```

Both misses were in `node_lifecycle.go`, already converted in `node.go`: the connmanager construction, and four callbacks including `BlockfetchRequestRangeFunc`. A third instance of the same bug predated this work — `HandleOutboundConnEvent` was subscribed as a method value in `Run()`, so it would have pinned the first instance across any restore.

`TestLiveTruncateRebuildsStorageAndKeepsNodeUsable` now asserts `require.NotSame(ouroBeforeRestore, n.ouroboros)`, so a future regression fails on the assertion rather than by hanging.

## Testing

- `make test` (race): 77 packages, 0 failures, 0 data races. Root package back to 16.5s, matching the pre-change baseline.
- conformance suite: pass
- `golangci-lint` 0 issues, `nilaway` clean, `import-boundaries`, `docs-parity`, `gorm-check`, `golines` all pass
- The regression above was attributed by running the failing test at the prior commit in a scratch worktree (5.7s pass) rather than assumed.

**DevNet passed cleanly on this branch**: `TestBasicBlockForging` 42s, `TestDingoChainAdvances` 10s, `TestChainGrowthRate` 120s, `TestRelayPropagation`, `TestSustainedConsensus` 104s, `TestEpochBoundaryConsensus` 446s, `TestStakeAddressInfoQueries` — 51 pass, 0 fail, exit 0. An earlier run on this branch's predecessor hit the intermittent stall tracked in #3180 (1 fail / 1 pass, matching what that issue records for unmodified main, with a clean baseline run in between).

Review findings from the bot reviewers are addressed in `0a09711`; see the follow-up comment for the per-finding breakdown, including four real bugs this refactor had introduced (Leios handlers dropped by a live restore, a resync subscription bound to the restore context, an unsynchronized instance pointer, and `Close` not waiting out in-flight handlers).

## Docs

- `ARCHITECTURE.md`: startup steps 4 and 16 corrected, and the Ouroboros Dependency Wiring section rewritten to cover why the cycle is only apparent, the `Close` contract, and why the closure rule matters.
- `node_lifecycle.go` header comment updated — it described `Ouroboros` as retained across a restore, which is no longer true.
- `DATABASE.md` checked and not affected: no metadata schema, blob-store layout, or SQL surface changes.

## Note for reviewers

The `connmanager` API gains two optional provider fields; the existing `Listeners`/`OutboundConnOpts` fields still work, so the other 60+ construction sites are untouched.


chrisguiney:
Addressed the review findings in `0a09711`. Each was verified against the code before acting; four turned out to be real bugs this refactor introduced.

**Real bugs, now fixed**

- **Leios handlers dropped by a live restore** (P1, correct). `reinitializeBackgroundManagers` restarts the vote and pipeline managers *before* `reinitializeNetworkingCore` replaces Ouroboros, so `SetLeiosVotes`/`SetLeiosPipeline` were applied to the outgoing instance. On a Dijkstra node a restore would silently drop Leios vote and pipeline handling while everything else kept working. Now carried across explicitly, with `TestLiveLifecycleRebuildPreservesLeiosHandlers` — confirmed to fail when the carry-over is removed.
- **Resync subscription bound to the restore context** (P2, correct, and worse than described). It captured the restore operation's `ctx`, not `n.ctx`, so once that operation finished the handler would stop firing — a running node with resync handling silently disabled. Now uses `n.ctx`, matching `Run`.
- **Data race on the instance pointer** (P2, correct). Making the field reassignable meant EventBus handlers and component callbacks could read it while a restore wrote it. It is now an `atomic.Pointer` behind a `Node.ouroboros()` accessor. Worth noting `-race` across 77 packages, including a live truncate under real forging and networking, did *not* flag it — the window is narrow, but an unsynchronized write is a race regardless.
- **`Close` used `Unsubscribe` instead of `UnsubscribeAndWait`** (P1, correct). It could return while a handler was still running against a ledger state or chainsync state the restore was about to close.

**Robustness**

- `ResolvedListeners` evaluated the caller-supplied provider while holding `listenersMutex`. No current provider reaches back into the connection manager, so this was latent rather than live, but resolving before locking is free.
- Typed-nil rejection for the three interface-typed dependencies (`Mempool`, and the two Leios handlers), matching the existing `TestResolveRejectsTypedNilInstance` precedent in the plugin host.
- Test `EventBus` instances now closed via `t.Cleanup`.

**Two comments not applied**

- *Remove the `ConnManager` assignment from the constructor and keep it behind `Wire`.* This predates the current head. `Wire` no longer exists — every dependency is a constructor argument now, which is the opposite direction. The two earlier bot reviews ran against `96e6b48` and `f5bf02b`, which used the `Wire` design.
- *Unsubscribe and re-register the node's connection handlers around a rebuild.* Those handlers are closures over `n`, not values bound to an instance, so they resolve the replacement automatically — that is the property the refactor relies on. For the subscriptions Ouroboros itself owns, `Close` now waits out in-flight handlers, which covers the underlying concern.

**Verification:** `make test` (race) 77 packages / 0 failures / 0 races; conformance pass; `golangci-lint` 0 issues; `nilaway` clean; `import-boundaries`, `docs-parity`, `gorm-check`, `golines` pass.

**DevNet on this branch passed cleanly** (the run that was pending when the description was written): `TestBasicBlockForging` 42s, `TestDingoChainAdvances` 10s, `TestChainGrowthRate` 120s, `TestRelayPropagation`, `TestSustainedConsensus` 104s, `TestEpochBoundaryConsensus` 446s, `TestStakeAddressInfoQueries` — 51 pass, 0 fail, exit 0.

chrisguiney:
Latest review addressed in `a3ff391`. One finding was a real startup crash; the other does not hold.

**P1 — nil dereference during Dijkstra startup: correct, and severe.**

Confirmed exactly as described. `Run` has no function boundary between line 435 and 1810, so it is one sequential body: `initLeiosVoteManager` (1131) and `initLeiosPipelineManager` (1142) both run before the `NewOuroboros` + store (1398/1402). `SetLeiosVotes` assigns to a field, so the nil receiver panics — a Dijkstra node would have crashed on startup, before Ouroboros was ever built.

A test reproducing it fails on the unfixed code:

```
--- FAIL: TestInitLeiosManagersDoNotRequireOuroboros
    Panic value: runtime error: invalid memory address or nil pointer dereference
```

Worth noting why nothing else caught it: DevNet passed cleanly on this branch, but that path is Dijkstra-only, so the suite never reaches it.

Fixed by moving the wiring out of the init functions into `attachLeiosHandlers`, called by `Run` immediately after construction and by `reinitializeNetworkingCore` immediately after its rebuild. Both paths now follow one rule — handlers attach to whatever instance is current, as soon as it exists — which also collapses the near-duplicate carry-over block the previous commit had added to the restore path.

**P2 — `leiosEBCaster` / `endorserBlockTxHashes` pinning the pre-restore instance: does not hold.**

The premise is that the forger outlives a live restore. It does not — `reinitializeBlockProducer` is a restore step that runs *after* `networking core`:

```
{"background managers", ...},
{"networking core", ...},   <- replaces ouroboros
{"API servers", ...},
{"block producer", ...},    <- calls initBlockForger
```

So `initBlockForger` rebuilds the forger and its `leiosPipelineAdapter` against the replacement, and both bindings are re-made. The same holds at startup: the store is at `node.go:1409` and `initBlockForger` at `node.go:1769`. I left these as direct references rather than adding lazy indirection that would not change behaviour.

**Systematic audit.** Since this bug class has now bitten twice, I checked every production `n.ouroboros()` call site rather than just the reported ones. All are accounted for: inside closures that resolve at call time, after construction, or deliberately targeting the outgoing instance (quiesce's nil-guarded check, and `Close` before replacement). No further instances.

**Verification:** `make test` (race) 77 packages / 0 failures / 0 races; conformance pass; `golangci-lint` 0 issues; `nilaway` clean; `docs-parity`, `import-boundaries`, `golines` pass.

chrisguiney:
Addressed in `b47117c`. One P3 finding, and it was correct.

`TestLiveTruncateRebuildsStorageAndKeepsNodeUsable`'s doc comment had been left stranded above `TestInitLeiosManagersDoNotRequireOuroboros`. Cause: when I inserted the earlier test I anchored on the `func` line rather than the doc comment above it, so the new test landed between a comment and the function it describes. Each comment now sits with its own function; the change is comment-only, no logic touched.

Verified anyway rather than assuming a comment move is inert: `make test` (race) 77 packages / 0 failures / 0 races, `go vet` clean, `golangci-lint` 0 issues, `docs-parity` pass.

wolf31o2 (dismissed):

chrisguiney:
Updated from `upstream/main` in `43dec2a` — brings in #3200, #3202, #3193, #3206 and #3203. The branch is now 0 behind.

**One conflict, in the `Ouroboros` struct.** This branch privatized the dependency fields; #3206 added the shared block/header decode caches to the same region. Resolved by keeping both.

**The merge needed a code fix, not just a textual resolution.** #3206's `initDecodeCacheMetrics` registers its collectors through `config.PromRegistry` directly — correct when it was written, but this branch added a tracking registerer so `Close` can hand collectors back. Left as-is, those three collectors would be untracked, and the next live restore (which now *rebuilds* Ouroboros against the retained registry) would panic on duplicate registration. Routed through `o.registerer` instead.

This is exactly the case the tracking registerer exists to make automatic, and the existing guard caught it — with the direct registration in place:

```
--- FAIL: TestCloseAllowsReconstructionWithSameRegistry
    Panic value: duplicate metrics collector registration attempted
```

#3206's new `decode_cache_test.go` also used the single-return `NewOuroboros`; switched to `newOuroboros`, matching every other in-package test.

**On merge vs rebase:** I started with a rebase, since this repo keeps a strictly linear history. It stalled on the first commit, wanting the old `Wire` design reconstructed against the new upstream code — a design that commit 3 of this branch deletes. That's up to six conflict resolutions through states that don't survive, so I merged instead: one resolution, against the final state. The merge commit disappears under a squash-merge anyway. Note the approval was dismissed by the push regardless of method, since new commits dismiss stale approvals here.

**Re-verified on the merged tree:** `make test` (race) 77 packages / 0 failures / 0 races; conformance pass; `golangci-lint` 0 issues; `nilaway` clean; `docs-parity`, `import-boundaries`, `golines` pass.

@wolf31o2 this needs a re-approve when you get a chance — the only change since your approval is this merge plus the decode-cache metrics fix above.

wolf31o2 (approved):

Issue #2905, opened 2026-07-21
mempool: expose the current queue as a configurable FIFO backend


wolf31o2:
## Summary

Make the current ordered mempool an explicit **FIFO** implementation behind a backend-neutral contract, and select the implementation through configuration. FIFO remains the default and compatibility baseline; a DAG implementation will be added separately and exercised through the same external surfaces.

Today `Node` and `Ouroboros` store `*mempool.Mempool` directly, while forging/API adapters expose only pieces of its behavior. That concrete coupling prevents controlled FIFO-versus-DAG experiments and encourages backend-specific behavior to leak into networking and node composition.

## Direction

The current implementation is FIFO: successful admissions append to an ordered slice, consumers traverse that order, block forging snapshots that order, and optional eviction removes the oldest entries. This ticket should name and document that behavior rather than adding fee-density priority.

This supersedes the fee-priority direction in #1905 for the FIFO baseline. The revalidation-error visibility portion of #1905 remains useful and should be split or retained independently.

## Acceptance criteria

- Define a minimal backend-neutral mempool contract covering lifecycle, admission, confirmed/manual removal, immutable snapshots, relay consumers, capacity/headroom, and stable events/metrics.
- Move/rename the current concrete implementation so it is clearly the FIFO backend. Define FIFO as successful-admission order and preserve existing protocol-visible behavior.
- Add a validated configuration value such as `mempoolImplementation: fifo|dag`; default to `fifo`. Support the existing CLI > env > YAML > defaults precedence and reject unknown values.
- Construct the selected implementation at the composition boundary (`node.go`/root/internal node), per repository isolation rules. Domain consumers depend on narrow interfaces, not a concrete FIFO or DAG type.
- Keep existing `mempool.add_tx` / `mempool.remove_tx` events and cardano-compatible metric names stable. Add backend identity to Dingo-only diagnostics/metrics so side-by-side runs can be attributed.
- Create backend contract tests that can run against FIFO now and DAG later. Include admission/removal, chained transactions, revalidation, relay cursor/cache semantics, capacity/backpressure, lifecycle, and concurrent access.
- Update `mempool/doc.go` to describe actual FIFO semantics and remove the incorrect fee-density claim.
- Update `ARCHITECTURE.md`, configuration docs/example, and operator-facing startup logging.
- Run unit/race tests, conformance tests, and DevNet for FIFO before making it the selectable baseline.

## Coordination

[...]

PR #3229, opened 2026-08-20
fix(mempool): log re-validation failures at warn level

### LINKED ISSUE #1905: Mempool improvements: fee-based priority and error visibility
Earlier mempool improvement phases are done (#1591). Two items remain:

1. Re-validation failures still logged at Debug level (mempool.go ~line 530). Transactions silently failing re-validation and being dropped from the pool should be logged at Warn or Error so operators notice.

2. Fee-based priority queue. Currently evictOldestLocked() uses strict FIFO order. High-fee transactions can be evicted before low-fee ones just because they arrived earlier. Need fee density sorting so eviction drops the cheapest transactions first. Note: doc.go claims this already exists but it doesn't.

Roadmap: Q3 2026

arepala-uml:
## Summary

Part of #1905. Re-validation failures during the mempool's periodic
overlay rebuild were logged at `Debug`, so a transaction silently
dropped from the pool (e.g. invalidated by a conflicting transaction
that confirmed first) left no trace an operator would normally see.

The fee-density eviction-priority half of #1905 is intentionally not
included here — per the #2905 FIFO-vs-DAG follow-up, FIFO is now the
explicit compatibility baseline and should not gain priority-based
eviction; that work belongs in the DAG backend instead. `mempool/doc.go`
already documents FIFO's oldest-first eviction accurately (fixed as
part of #2905), so no doc changes are needed here.

## Changes

- **`mempool/mempool.go`** — in `revalidateAppliedTx`, the "transaction
  failed re-validation" log on a dropped transaction is now emitted at
  `Warn` instead of `Debug`, with a clarified message
  ("...and was dropped from the mempool") so the log line is
  self-explanatory without additional context.

Closes #1905 

---
## Summary by cubic
Logs mempool re-validation failures at Warn instead of Debug and clarifies that the transaction was dropped. This makes drops during periodic overlay rebuilds visible to operators without changing mempool behavior.

- Change is limited to `mempool/mempool.go` in `revalidateAppliedTx`; the log now emits at Warn with the tx hash and error.
- Expect higher Warn volume when conflicts invalidate pending transactions; adjust alerts as needed. No config, API, or docs changes.

<sup>Written for commit 9877318abd97ec13b4261fb6e2164416d1a9190f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3229?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved visibility of transaction revalidation failures by logging them as warnings.
  * Clarified that affected transactions are dropped from the mempool.

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -1291,8 +1291,8 @@ func (m *Mempool) revalidateAppliedTx(
 		candidate.overlay.created,
 	); err != nil {
 		candidate.reject(at, tx)
-		m.logger.Debug(
-			"transaction failed re-validation",
+		m.logger.Warn(
+			"transaction failed re-validation and was dropped from the mempool",
 			"component", "mempool",
 			"tx_hash", at.hash,
 			"error", err,


arepala-uml:
I have scoped this down per the #2905 direction update where implemented only the
re-validation error-visibility fix, not the fee-density eviction change.

- `revalidateAppliedTx` now logs a transaction dropped during periodic
  overlay re-validation at `Warn` instead of `Debug`, so operators can see
  when a pending transaction silently falls out of the pool.
- Did **not** add fee-density priority selection to `evictOldestLocked()`.
  FIFO remains strict oldest-first eviction, unchanged. Since #2905 makes
  FIFO the explicit compatibility baseline, changing its eviction order
  would violate that contract. (Based on the comment from issue https://github.com/blinklabs-io/dingo/issues/1905#issuecomment-5039804497)
- No `doc.go` changes needed here because it already documents FIFO's
  oldest-first eviction accurately, per the fix that landed in #2905.

wolf31o2 (approved):

PR #3240, opened 2026-08-20
fix(ledger): match PlutusV1 inline datum usage

wolf31o2:
## Problem

Conway rule 18 (`conway.UtxoValidateInlineDatumsWithPlutusV1`) rejected a real
preview transaction that `cardano-node` accepts, halting Mithril replay at slot
`120582947`.

```
tx     3ecccfce53cc0cab6c2f5c62f4eca3ac4174f5cb53cb8985fb9004c68e015b7b
block  e447af4cdd64196e73d408c0d18ce9b75d854cedf8e37e2c0f14ba579431d61d
height 4588440   tx_block_index 2   size 8916   fee 610052
inputs 100   reference inputs 0   collateral 0
witnesses  plutusV1=0 plutusV2=0 plutusV3=0 native=0 redeemers=none datums=0
```

Two of its inputs matter, both key-locked (`addrType=6`, script bit clear):

```
input[1]   1cf3179761edad428019bc3bd11cb80c9b318caa87c68b84623a6f108f5e41f2#0
           inline datum, no reference script
           created slot 120564434, block 065a4540b9d8f5586d57af746f2e796dfdab51bd82085bb0265ed9c6b9af4a89

input[80]  ca28dbcb21679e53db388af0206e3e3a1ffadb9041fa2e569e7db3575dc8dd94#0
           PlutusV1 reference script (28 bytes), no datum
           created slot 120570186, block ba7ff9ab72d0f02668d73316ae77dfa2ef1fd058d7b059f7373e18b6df703ee6
```

No purpose in the transaction needs a V1 script. Upstream rule 18 counted the
unused reference script on `ca28db…#0` as PlutusV1 being in use, then rejected
on the unrelated inline datum on `1cf317…#0`.

Field values were resolved over NtN blockfetch from a preview relay, so they
come from the block bytes rather than an indexer.

## Changes

`validateConwayInlineDatumsWithPlutusV1` replaces gouroboros rule 18 for
Conway, keyed on whether a V1 script is *needed* rather than merely reachable:

- `conwayNeedsPlutusV1` walks every script purpose — spending, minting,
  certifying, rewarding, voting, proposing — via the existing
  `forEachConwayScriptPurpose`, and resolves each to a script hash. An unused
  reference script produces no purpose and no rejection.
- Once a purpose needs V1, an inline datum on a consumed input, a reference
  input, or a produced output rejects. That is the exact set Conway's V1
  `toPlutusTxInfo` translates.
- Datum presence is read through the `TransactionOutput` interface rather than
  a `*BabbageTransactionOutput` type assertion, so later-era wrapped outputs
  are inspected too. Datum-*hash* outputs stay out of scope because `Datum()`
  reports nil for them.

Conway restrictions deliberately not enforced, each verified absent from
Conway's own translation in
`eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs`:

- reference scripts, on spent inputs or produced outputs — Conway's
  `transTxOutV1` checks only `dataTxOutL`; Babbage's is the version that raises
  `ReferenceScriptsNotSupported`
- the mere presence of a reference input — Babbage rejects it outright, Conway
  translates each with `mapM_` and discards the result

Errors are reported as `InlineDatumsNotSupportedError` regardless of which
restriction tripped, matching upstream rule 18 and gouroboros #1980, so
dropping this override once #1980 lands does not change the error type callers
see.

## Validation

`go test -race ./ledger/eras/`, `go test ./ledger/...`, `go vet ./ledger/...`,
`golangci-lint run ./ledger/eras/...` (0 issues), `gofumpt`, `golines`, and
`go test ./internal/test/conformance/` (315 vectors).

Skipped, and worth stating because it bounds what the conformance run proves:
the conformance harness invokes gouroboros' rule functions directly —
`ouroboros-mock@v0.16.0/conformance/validation_rules.go` names
`conway.UtxoValidateInlineDatumsWithPlutusV1` — so it never reaches this
override. Disabling this rule outright still leaves all 315 vectors passing.
The regression tests in `ledger/eras/validation_test.go` are what cover it, and
they assert the upstream rule's failure alongside this rule's success.


ledger/eras/conway.go
```diff
@@ -279,10 +284,72 @@ func buildConwayValidationRules() []indexedUtxoValidationRule {
...
+			continue
+		}
+		script, ok := ctx.scriptInputs.scripts[lcommon.ScriptHash(addr.PaymentKeyHash())]
+		if !ok {
+			continue
+		}
+		if _, ok := script.(lcommon.PlutusV1Script); ok {
+			return lcommon.InlineDatumsNotSupportedError{PlutusVersion: "PlutusV1"}
+		}
+	}
```
chrisguiney: **Blocking:** scoping the check to the datum-bearing input's own script means the rule stops firing when a PlutusV1 script runs for any other purpose.

cardano-ledger doesn't scope this per-input. `transTxInInfoV1` maps `transTxOutV1` over *every* spent input while assembling the V1 script context, so once any V1 script runs, an inline datum on *any* input fails with `InlineDatumsNotSupported`.

I confirmed the divergence with a throwaway test — PlutusV1 witness script, a mint redeemer, and one key-locked input carrying an inline datum:

```
upstream rule 18: inline datums are not supported with PlutusV1 scripts
PR rule 18:       <nil>
```

There's no backstop behind it. `grep -rn InlineDatum --include=*.go` shows rule 18 is the only enforcement point in dingo, and `script.NewTxInfoV1FromTransaction` (gouroboros `ledger/common/script/context.go:197`) does no validation of its own — it doesn't reject inline datums, reference scripts, or reference inputs. So dingo would accept the transaction, evaluate the script, and pass it.

Worth noting that upstream's witness branch (`len(v1Scripts) > 0` → reject) was crude but approximately correct, since an unused script witness is itself invalid. It was only the *reference-script* branches that were wrong. This change drops the correct half along with the broken half.

Suggested direction: gate on "is any PlutusV1 script actually needed by this tx", then apply the broad check. Most of that already exists a couple hundred lines below — `validateConwayRequiredPlutusRedeemers` (line 571) and `buildConwayScriptPurpose` (line 1172) already enumerate every spend/mint/cert/reward/vote/propose purpose and resolve each to a script hash via `scriptInputs.scripts`. Reusing that set would close the gap and also remove the duplication, since this loop is effectively a narrower, spend-only re-implementation of it.

ledger/eras/conway.go
```diff
@@ -279,10 +284,72 @@ func buildConwayValidationRules() []indexedUtxoValidationRule {
...
+	slices.SortFunc(ret, func(a, b indexedUtxoValidationRule) int {
+		return a.index - b.index
+	})
+	return ret
+}
+
+// validateConwayInlineDatumsWithPlutusV1 rejects inline datums only when the
+// datum-bearing input is locked by a PlutusV1 script. The gouroboros rule this
+// replaces reports a false positive for an unrelated PlutusV1 reference script
+// attached to a key-locked input.
```
chrisguiney: Minor: this comment reads as a faithful Conway check, but the rule is narrower than cardano-ledger's V1 context translation, which also rejects:

- reference scripts on spent inputs (`ReferenceScriptsNotSupported`)
- inline datums or reference scripts on produced outputs
- the presence of any reference inputs at all (`ReferenceInputsNotSupported`)

None of those are covered here. This is substantively CodeRabbit's point and I think it holds. Either widen the rule or state the limits explicitly and open a follow-up issue — a future reader shouldn't have to rediscover that the V1 restriction is only partly enforced.

(For what it's worth, Cubic's `ctx.redeemers == nil` critique is stale — that short-circuit went away in 6eea736b.)

ledger/eras/conway.go
```diff
@@ -279,10 +284,72 @@ func buildConwayValidationRules() []indexedUtxoValidationRule {
...
+	for _, input := range tx.Inputs() {
+		utxo, err := ls.UtxoById(input)
+		if err != nil || utxo.Output == nil {
+			continue
+		}
+		if utxo.Output.Datum() != nil {
+			hasInlineDatums = true
+			break
+		}
+	}
```
chrisguiney: Minor: this loop swallows `UtxoById` failures, but `newConwayPlutusValidationContext` on line 331 turns the same failures into hard errors. That makes rule 18 a new source of input-resolution errors that previously only surfaced at rule 29 (`BadInputsUtxo`).

Nothing breaks — errors are joined and the tx is rejected either way — but the two halves of the same function disagree about whether an unresolvable input is this rule's problem. Upstream's contract ("Skip errors - BadInputsUtxo will catch this") seems like the more defensible one, and would keep rule 18 single-purpose.

ledger/eras/conway.go
```diff
@@ -279,10 +284,72 @@ func buildConwayValidationRules() []indexedUtxoValidationRule {
...
+	if !hasInlineDatums {
+		return nil
+	}
+	ctx, err := newConwayPlutusValidationContext(tx, ls)
+	if err != nil {
+		return err
+	}
```
chrisguiney: Minor, hot path: `newConwayPlutusValidationContext` does more than this rule needs — it blake2b-hashes every PlutusData witness and runs `SortInputs`, neither of which is read below. `resolveConwayScriptInputs(tx, ls, true)` is the whole requirement.

There's also redundancy on the other side: `ValidateTxConway` (line 218) rebuilds the identical context about fifteen lines after the rules loop finishes. Since `LedgerView.UtxoById` (`ledger/view.go:92`) is uncached — every call is a DB read plus a full CBOR decode — that duplication is worth avoiding, even though the DB-read count is roughly at parity with upstream.

ledger/eras/conway.go
```diff
@@ -279,10 +284,72 @@ func buildConwayValidationRules() []indexedUtxoValidationRule {
...
 		},
 	}
-	return buildIndexedUtxoValidationRulesWithSkips(
+	ret := buildIndexedUtxoValidationRulesWithSkips(
 		conway.UtxoValidationRules,
 		skips,
 	)
+	// Replace gouroboros' rule 18 with the local implementation. The upstream
+	// rule treats every PlutusV1 reference script on an input as used, even if
+	// the input is key-locked and no redeemer selects that script.
```
chrisguiney, line 293: Scope question, not a blocker for this PR: the same false positive is still live one rule over. `conway.UtxoValidateConwayFeaturesWithPlutusV1V2` (index 19) has the identical "any PlutusV1/V2 reference script on a spent input counts as used" flaw, and rejects when the transaction carries a treasury value, proposal/voting procedures, or a Conway certificate.

Vote-delegation certificates are common, so the same key-locked-reference-script UTxO will halt sync again the first time it meets one. Only index 18 is patched here.

Same for the other eras — `buildBabbageValidationRules` (`babbage.go:419`) and `buildDijkstraValidationRules` (`dijkstra.go:278`) both still use the unpatched upstream rule, and Babbage-era blocks are replayed on every full sync. Happy for that to be a follow-up issue (or an upstream gouroboros fix) rather than growing this PR.

ledger/eras/validation_test.go
```diff
@@ -925,6 +925,143 @@ func TestValidateTxPlutusConwayUnusedReferenceScriptWithoutRedeemerPasses(
...
+		if rule.index != conwayUtxoValidateInlineDatumsWithPlutusV1RuleIndex {
+			continue
+		}
+		require.NoError(t, rule.validationFunc(
+			tx,
+			0,
+			ls,
+			&conway.ConwayProtocolParameters{},
+		))
+		return
```
chrisguiney: Small test notes, all non-blocking:

- This rule-lookup loop is copy-pasted at line 1022. A tiny `conwayInlineDatumRule(t)` helper returning the func (or calling `t.Fatal`) would take the duplication out of both.
- Neither test covers the case in my main comment — a PlutusV1 script used for a *non-spending* purpose alongside a key-locked inline-datum input. That's exactly the shape the narrowing stops rejecting, so it's the one most worth pinning once the scope question is settled.
- Nothing currently asserts that the merged rule list stays index-sorted and complete after the new `append`/`SortFunc`. A cheap guard — strictly increasing indices, and length equal to `len(conway.UtxoValidationRules) - 2` — would catch a future edit that appends without sorting.

The two tests you did add are well constructed, and asserting the upstream rule's failure alongside the local rule's success is a nice touch — it documents the regression rather than just testing around it.

chrisguiney (changes requested):
Thanks for chasing this one down to a concrete Preview transaction — the diagnosis is right, and the false positive is real. I reproduced the shape locally and confirmed the fix clears it.

I'd like to hold off on approving for one reason: the replacement rule is narrower than the ledger rule it stands in for, and I was able to show it now accepts a transaction shape cardano-node rejects. Details inline, with a suggested direction that should be a fairly small change given the purpose machinery already in this file.

### Verified good

- Index 18 really is `conway.UtxoValidateInlineDatumsWithPlutusV1` in gouroboros v0.193.3 (`ledger/conway/rules.go:57`), and the existing name-based init guard covers drift.
- `Datum()` returns nil for pre-Babbage outputs and for datum-*hash* outputs (`babbage.go:764`), so datum-hash UTxOs are not misread as inline. Moving from upstream's `*BabbageTransactionOutput` type assertion to the interface method is a genuine improvement — it also covers Dijkstra-wrapped outputs.
- `AddressTypeScriptBit` / `PaymentKeyHash()` are correct for payment credentials; Byron (`0b1000`) correctly falls through.
- `append` + `SortFunc` is safe here: indices are unique, so sort instability doesn't matter, and capacity is preallocated.
- No concurrency exposure — build and sort happen once at package init, and the rule is read-only over `ls`.
- `go test -race ./ledger/eras/`, `go test ./ledger/...`, `go vet ./ledger/...`, and `gofmt` are all clean locally, and all 9 CI checks are green.
- The reported Preview transaction is genuinely fixed: both inputs are key-locked, so both `continue` and the rule returns nil.

### One small ask

Could you include the full transaction id and input ids rather than the truncated `3ecccf…` / `1cf317…` / `ca28db…`? Right now the vector can't be independently confirmed or turned into a conformance fixture, which would be nice to have for a consensus-affecting rule.

ledger/eras/conway.go
```diff
@@ -279,10 +284,72 @@ func buildConwayValidationRules() []indexedUtxoValidationRule {
...
+		addr := utxo.Output.Address()
+		if addr.Type()&lcommon.AddressTypeScriptBit == 0 {
+			continue
+		}
+		script, ok := ctx.scriptInputs.scripts[lcommon.ScriptHash(addr.PaymentKeyHash())]
+		if !ok {
+			continue
+		}
+		if _, ok := script.(lcommon.PlutusV1Script); ok {
+			return lcommon.InlineDatumsNotSupportedError{PlutusVersion: "PlutusV1"}
```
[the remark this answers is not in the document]
wolf31o2: Confirmed on both halves, fixed in `ef5504c1`.

The scan covered only consumed inputs, and rejection required the PlutusV1 script to sit on the spending input's own address. Both are now transaction-wide:

- `conwayTxHasInlineDatum` walks `resolvedAllInputs` (consumed *and* reference inputs) plus `tx.Outputs()`.
- Needed-script resolution goes through all six purposes rather than spending alone, so a PlutusV1 script required for minting, certifying, rewarding, voting or proposing counts — including one reached through a reference input's script ref rather than an input address.

Rather than write a second purpose walk, `forEachConwayScriptPurpose` was extracted from `validateConwayRequiredPlutusRedeemers`, which already enumerated all six in canonical redeemer order. Both rules now share that one walk: two copies would drift, and a purpose missing from either is a silent validation gap rather than a visible failure.

The unused-reference-script case is preserved by keying on needed scripts only — a PlutusV1 script that is merely available never triggers rejection.

New coverage, each proven to depend on the change:

| Test | Fails when |
| --- | --- |
| `RejectsDatumOnOutput` | scan reverted to consumed inputs only |
| `RejectsDatumOnReferenceInput` | scan reverted to consumed inputs only |
| `AcceptsUnusedPlutusV1WithLiveRedeemer` | check made purpose-blind |

Not adopted from the suggestion: a separate rule rejecting reference inputs outright for PlutusV1. That is a distinct restriction from the inline-datum one and wants its own conformance evidence, so it is not folded in here.

Validation at this head: `go test -race ./ledger/...` passes across all nine packages (`ledger` 811.7s, `governance` 121.3s, `snapshot` 125.8s, rest under 12s); `go test ./internal/test/conformance/` passes in 5.7s; `golangci-lint v2.13.0 run ./ledger/...` reports 0 issues.

ledger/eras/validation_test.go
```diff
@@ -925,6 +925,143 @@ func TestValidateTxPlutusConwayUnusedReferenceScriptWithoutRedeemerPasses(
 	))
 }
 
+func TestConwayInlineDatumRuleIgnoresUnusedPlutusV1ReferenceScript(
```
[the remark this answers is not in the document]
wolf31o2: Correct, and the first attempt at the replacement was vacuous in the same way.

Fixed in `ef5504c1`. `TestConwayInlineDatumRuleAcceptsUnusedPlutusV1WithLiveRedeemer` now spends a **PlutusV2** script address, so the transaction has a real script purpose and the scan is entered, while a PlutusV1 script is present in the witness set and as a reference script but required by no purpose. An inline datum is present. A correct implementation accepts, because the only needed script is PlutusV2.

Worth recording the intermediate step, since it is the trap this finding describes. The first version of this test used a redeemer but put both inputs at key addresses. No address carried the script bit, so no purpose was produced, the callback never fired, and the test passed against a deliberately purpose-blind implementation that scanned every available script instead of the needed ones. It asserted nothing.

Verified by construction this time: replacing the needed-script lookup with a loop over all available scripts fails this test and only this test. Reverting the widened datum scan instead fails only the two datum-location tests.

ledger/eras/conway.go
wolf31o2: Fixed in `ef5504c1`, and pinned since. `conwayNeedsPlutusV1` resolves needed scripts through `forEachConwayScriptPurpose` across all six purposes, and `resolvedAllInputs` covers consumed and reference inputs, so a V1 script required for minting, certifying, rewarding, voting or proposing counts regardless of which input carries the datum.

Confirmed live rather than assumed: restricting the walk to `ScriptPurposeSpending` fails `TestConwayInlineDatumRuleRejectsNonSpendingPlutusV1` and nothing else.

One sub-claim does not hold. There *is* a backstop for an unused PlutusV1 witness script: `conway.UtxoValidateScriptWitnesses` delegates to `common.ValidateScriptWitnesses`, which returns `ExtraneousScriptWitnessesError` for any explicit witness script no purpose requires (`gouroboros@v0.193.3/ledger/common/rules.go:429-438`). That rule is in the merged list and is not among the skips, so it is reached. Recorded in the doc comment so the next reader does not have to re-derive it.

ledger/eras/conway.go
wolf31o2: Right, and the comment was worse than narrow — it was self-contradictory. `284d743`'s doc block was prepended without deleting the previous two lines, so it opened with "rejects inline datums only when the datum-bearing input is locked by a PlutusV1 script", describing behavior the code no longer had, immediately followed by the transaction-wide description. Removed in `22ae76e7`.

Each of the three restrictions you listed is now resolved on evidence rather than on wording, and two of them the opposite way from what the finding suggests:

- **Inline datum on a produced output** — already covered.
- **Reference script on a produced output, and on a spent input** — *not* restrictions in Conway. `eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs` defines its own `transTxOutV1` which shadows Babbage's and drops the `ReferenceScriptsNotSupported` branch, so it checks only the inline datum. Babbage's version keeps that branch, which is why reading the Babbage file alone points the other way. This rule is registered for Conway. Verified directly against `IntersectMBO/cardano-ledger`, and gouroboros#1980 removed the equivalent check upstream for the same reason.
- **Presence of any reference input** — likewise not a Conway restriction: `ReferenceInputsNotSupported` appears only in Babbage's `TxInfo.hs`. An earlier revision of #1980 asserted it and failed the Conway vector `UTXOS/can use reference scripts`, which expects such a transaction to succeed.

Both are now pinned by tests that fail if the rejection is added back, so the scope is deliberate rather than accidental. The divergence and its reasoning are in the doc comment.

ledger/eras/conway.go
wolf31o2: Agreed, and taken in `c1589aef` — the rule returns `nil` for `isInputResolutionError(err)`, so `UtxoValidateBadInputsUtxo` reports an unresolvable input with the right error and this rule is no longer a second, competing source of the same failure. Upstream's contract is the same ("Skip errors - BadInputsUtxo will catch this").

That branch had no test, which `22ae76e7` fixes: `TestConwayInlineDatumRuleSkipsUnresolvableInput` spends a V1 script address carrying an inline datum — so the rule would otherwise reject — alongside a second input absent from the ledger state, and expects no error. Changing the branch to `return err` fails only that test.

ledger/eras/conway.go
wolf31o2: First half taken in `c1589aef`: the rule calls `resolveConwayScriptInputs(tx, ls, true)` directly, so it no longer blake2b-hashes every PlutusData witness or runs `SortInputs` through the full context builder.

Second half declined, on two grounds. `newConwayPlutusValidationContext` at `conway.go:219` is unchanged on `origin/main`, so the duplicate build is pre-existing rather than introduced here. And this rule's own read count went *down*: upstream rule 18 called `UtxoById` in three loops — `tx.Inputs()`, `tx.ReferenceInputs()`, then `tx.Inputs()` again (`gouroboros@v0.193.3/ledger/babbage/rules.go:252,285,309`) — where this one resolves inputs plus reference inputs exactly once.

Eliminating the remaining duplicate means threading resolved state through the shared `utxoValidationRule` signature in every era, which is a separate change. Tracked with the related upstream duplication in blinklabs-io/gouroboros#1982.

ledger/eras/conway.go
wolf31o2: Confirmed live, and filed as blinklabs-io/dingo#3256 rather than fixed here.

Verified the mechanism: the only skips in the merged list are `conwayUtxoValidateFeeTooSmall`, `PlutusScripts` and rule 18, so index 19 is the unpatched upstream `conway.UtxoValidateConwayFeaturesWithPlutusV1V2` — visible as `{19 0x4ab310}` in a rule-list dump. It decides "uses PlutusV1/V2" from witness-set presence alone, which is the same defect. Vote-delegation certificates are ordinary traffic, so the same key-locked reference-script UTxO trips it.

Also confirmed for the other eras: `buildBabbageValidationRules` (`babbage.go:419`) and `buildDijkstraValidationRules` (`dijkstra.go:278`) apply no rule-18 override, so both still run the unpatched function. That half needs no separate code change — gouroboros#1980 fixes the shared `babbage.UtxoValidateInlineDatumsWithPlutusV1` that Conway and Dijkstra delegate to, so a dependency bump covers all three and makes this override removable. Rule 19 needs its own fix, which is what #3256 tracks.

ledger/eras/validation_test.go
wolf31o2: All three taken, and the third was still genuinely missing.

- The duplicated rule-lookup loop: the `conwayInlineDatumRule(t)` helper was added in `c1589aef`, but the two original tests were never converted and still carried the copy. Converted both in `22ae76e7`, and replaced their inline `NewAddressFromParts` blocks with the existing `newTestKeyAddr` / `newTestPlutusV1ScriptAddr` helpers.
- The non-spending-purpose test landed in `c1589aef` and is non-vacuous: restricting the walk to spending purposes fails it and only it.
- The rule-list guard had the ordering, uniqueness and `Contains` assertions but **not** the completeness check you asked for. Added: length equal to `len(conway.UtxoValidationRules)` minus the two skips. It discriminates — dropping `ret[0]` after the sort keeps ordering and `Contains` green and fails only `require.Len`, with `"should have 53 item(s), but has 52"`.

One correction to how this PR was validated, which affects how much weight to give any of it: **`internal/test/conformance` never exercises this rule.** `ouroboros-mock@v0.16.0/conformance/validation_rules.go` hardcodes the upstream `conway.UtxoValidateInlineDatumsWithPlutusV1`, so Dingo's override is not reached. Proved both ways — disabling the override entirely leaves 315/315 vectors passing, and adding spurious rejections also leaves 315/315 passing. Earlier notes on this PR cited conformance as evidence for the rule; that was wrong. Unit tests and the cardano-ledger source reading are the actual evidence, and a devnet run is what would confirm behavior on real chain data.

ledger/eras/conway.go
```diff
@@ -279,10 +284,150 @@ func buildConwayValidationRules() []indexedUtxoValidationRule {
...
+		if isInputResolutionError(err) {
+			// UtxoValidateBadInputsUtxo reports an unresolvable input with the
+			// right error. Returning it here as well would make this rule a
+			// second, competing source of input-resolution failures, which is
+			// not what upstream's rule 18 does either.
+			return nil
+		}
+		return err
+	}
+	if !conwayTxHasInlineDatum(tx, scriptInputs) {
```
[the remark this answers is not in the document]
wolf31o2: Both halves refuted for Conway, against the reference source rather than by reasoning. No change needed at head; `95ec810f` corrects the comment that made this thread look unanswered.

**Output reference scripts are not a Conway restriction.** Conway defines its own `transTxOutV1` that shadows Babbage's and checks only the inline datum:

```haskell
-- eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs
transTxOutV1 txOutSource txOut = do
  when (isSJust (txOut ^. dataTxOutL)) $ do
    Left $ inject $ InlineDatumsNotSupported @era txOutSource
  case Alonzo.transTxOut txOut of ...
```

Babbage's is the one carrying `ReferenceScriptsNotSupported` on `referenceScriptTxOutL`. Reading the Babbage file alone gives the opposite answer, which is how the previous comment came to claim it.

**Reference inputs are permitted, and their inline datums are already rejected.** Babbage's V1 `toPlutusTxInfo` opens with `unless (Set.null refInputs) $ Left (ReferenceInputsNotSupported refInputs)`. Conway replaced that:

```haskell
inputs <- mapM  (transTxInInfoV1 ltiUTxO) (Set.toList (txBody ^. inputsTxBodyL))
mapM_           (transTxInInfoV1 ltiUTxO) (Set.toList (txBody ^. referenceInputsTxBodyL))
```

`mapM_` runs the same V1 translation over each reference input and discards the result, so the *presence* of a reference input is fine and an inline datum on one still raises `InlineDatumsNotSupported`. The rule's loop walks `scriptInputs.resolvedAllInputs`, which is consumed plus reference inputs, so that case is covered.

Enforcing either restriction independently, as suggested, would reject transactions `cardano-node` accepts -- the same class of false positive this rule exists to remove.

wolf31o2:
## Full vector

Resolved from the chain over NtN blockfetch against a preview relay, not from an indexer, so every field below is from the block bytes.

**Rejected transaction** — preview, slot `120582947`, epoch 1395:

```
tx     3ecccfce53cc0cab6c2f5c62f4eca3ac4174f5cb53cb8985fb9004c68e015b7b
block  e447af4cdd64196e73d408c0d18ce9b75d854cedf8e37e2c0f14ba579431d61d
height 4588440   tx_block_index 2   size 8916   fee 610052
inputs 100   reference inputs 0   collateral 0
witnesses  plutusV1=0 plutusV2=0 plutusV3=0 native=0 redeemers=none datums=0
```

**The inline-datum input** — `input[1]`:

```
1cf3179761edad428019bc3bd11cb80c9b318caa87c68b84623a6f108f5e41f2#0
  created  slot 120564434, block 065a4540b9d8f5586d57af746f2e796dfdab51bd82085bb0265ed9c6b9af4a89
  output[0]  inlineDatum=true  datumHash=9e1199a988ba72ffd6e9c269cadb3b53b5f360ff99f112d9b2ee30c4d74ad88b
             scriptRef=nil
             addr=addr_test1vz0n3cp5wpka9ry0hc9zkh0kvq9ux9zk7vsmyfsevscgfeg5n56cm
             addrType=6  scriptBit=false
             paymentKeyHash=9f38e034706dd28c8fbe0a2b5df6600bc31456f321b22619643084e5
```

**The PlutusV1 reference-script input** — `input[80]`:

```
ca28dbcb21679e53db388af0206e3e3a1ffadb9041fa2e569e7db3575dc8dd94#0
  created  slot 120570186, block ba7ff9ab72d0f02668d73316ae77dfa2ef1fd058d7b059f7373e18b6df703ee6
  output[0]  inlineDatum=false  datumHash=none
             scriptRef=PlutusV1Script (28 bytes)
             addr=addr_test1vr2ex758mpw8qaul03hcculfm48ykk85859h8vp0dxzl0yssgckn7
             addrType=6  scriptBit=false
             paymentKeyHash=d5937a87d85c70779f7c6f8c73e9dd4e4b58f43d0b73b02f6985f792
```

Both UTxOs are `addrType=6` with the script bit clear, so both are key-locked and neither can require a script. The transaction carries no Plutus witness of any version and no redeemers, so no purpose needs a V1 script. Upstream rule 18 nonetheless counted the unused reference script on `ca28db…#0` as PlutusV1 being in use, and then rejected on the unrelated inline datum on `1cf317…#0`.

## Conway restriction set, from the reference implementation

Fetched `eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs` and the Babbage equivalent. Conway shadows both `transTxOutV1` and the V1 `toPlutusTxInfo`, and drops two Babbage restrictions:

```haskell
-- eras/conway  transTxOutV1
transTxOutV1 txOutSource txOut = do
  when (isSJust (txOut ^. dataTxOutL)) $ do
    Left $ inject $ InlineDatumsNotSupported @era txOutSource
  ...

-- eras/babbage  transTxOutV1
transTxOutV1 txOutSource txOut = do
  when (isSJust (txOut ^. referenceScriptTxOutL)) $ do
    Left $ inject $ ReferenceScriptsNotSupported @era txOutSource
  when (isSJust (txOut ^. dataTxOutL)) $ do
    Left $ inject $ InlineDatumsNotSupported @era txOutSource
  ...
```

and for reference inputs:

```haskell
-- eras/conway  instance EraPlutusTxInfo 'PlutusV1 ConwayEra
inputs <- mapM  (transTxInInfoV1 ltiUTxO) (Set.toList (txBody ^. inputsTxBodyL))
mapM_           (transTxInInfoV1 ltiUTxO) (Set.toList (txBody ^. referenceInputsTxBodyL))
outputs <- zipWithM (transTxOutV1 . TxOutFromOutput) [minBound ..] (F.toList (txBody ^. outputsTxBodyL))

-- eras/babbage  same instance
unless (Set.null refInputs) $ Left (ReferenceInputsNotSupported refInputs)
```

So Conway's V1 restriction is exactly: an inline datum on a consumed input, a reference input, or a produced output. A reference script is unrestricted anywhere, and a reference input is permitted -- Conway translates it and discards the result, so only its inline datum disqualifies. That is the set the rule at head enforces.

`95ec810f` corrects the doc comment, which described Babbage's restrictions as Conway's and listed the two omissions as known divergences. No behavior change; the loops already covered inputs, reference inputs and outputs.

wolf31o2:
Closing as overtaken by gouroboros #1980, which landed the equivalent fix upstream.

At gouroboros v0.194.0, `babbage.UtxoValidateInlineDatumsWithPlutusV1` gates on needed scripts via `script.NewTxScriptView` + `view.NeedsAny`, and `conway.UtxoValidateInlineDatumsWithPlutusV1` is a one-line delegate to it, with Dijkstra delegating to Conway. That is the same algorithm as this branch's `validateConwayInlineDatumsWithPlutusV1`, so the local rule-18 override is now redundant — as this PR's own doc comment anticipated.

`main` already relies on the upstream rule: it carries no rule-18 override and no rule-index constant for one.

Merging this branch as-is would also fail. `ledger/eras/validation_test.go:961` asserts the upstream rule returns `InlineDatumsNotSupportedError` for an unused PlutusV1 reference script — true of the v0.193.3 pin this branch still uses, and exactly the bug #1980 fixed. Against the current dependency that call returns nil.

The test coverage is preserved in #3306, retargeted at the upstream rule, with that assertion inverted so it pins the fix instead of the old behavior. #3306 keeps nine cases including datum on a produced output, datum on a reference input, reference script on an output, reference inputs with a needed V1 script, the unresolvable-input skip, and all five non-spending purposes; each was verified non-vacuous against mutated copies of the upstream rule.

PR #3246, opened 2026-08-20
fix(keystore): distinguish an unresolvable DACL trustee from a foreign one

wolf31o2:
The Windows job in the v0.70.0 tag's publish run failed with 69 copies of one
error across the dingo, keystore, ledger/forging and ledger/leios packages:

```
key file "...\vrf.skey" grants access to unexpected trustee LA: insecure file mode
```

`insecure file mode` was the only error text in the job. It also failed
`TestSecureFileModeWindows`, which builds an owner-only DACL from the process
token and then asserts `checkOpenFilePermissions` accepts it, so the trustee
comparison was rejecting a DACL the same package built to be owner-only.

## Changes

`LA` is allowed only when the file is owned by Built-in Administrators. On that
owner every member of the group already has owner-equivalent access and `BA`
itself is allowed, so the ACE grants nothing new; the runners produce exactly
that pairing. An `LA` ace on a file owned by any other principal is still
rejected, including a domain Administrator that shares RID 500 with the local
account and differs only in the domain part.

`trusteeIsOwner` returned a bare bool, so `resolveTrusteeSID` failing to
resolve an SDDL alias was indistinguishable from the trustee genuinely not
being the owner. `trusteeOwnerMatch` now returns the resolution error and the
caller reports the two cases separately, naming the owner SID. The old message
named the trustee but never what it was compared against, which is what
separates an alias that failed to resolve from a real third-party ACE.

Aliases are resolved by asking Windows, not by matching a well-known RID. A
suffix match on `-500` would treat a domain Administrator owner as satisfying
an `LA` ace.

The three `runtime.KeepAlive` calls are removed rather than extended.
`syscall.SyscallN` is `//go:uintptrkeepalive`, so `windows.EqualSid` already
retains both arguments across the call, and both SID paths point into
Go-managed memory: `StringToSid` returns a pointer to the base of a Go
`[]byte`, and the alias path returns an interior pointer into a Go-heap
descriptor copy. A caller-side `KeepAlive` adds nothing on either.

## Validation

`GOOS=windows GOARCH=amd64 go build ./...` and `go vet ./keystore/`, which
compiles the `_test.go` files, plus `gofmt`, `golines` and `golangci-lint`.
Linux `go build ./...` and `go test ./keystore/` are unaffected.

Skipped: the Windows code is behind a build tag and cannot run on the
development host, so its runtime behaviour is unverified here. The Windows CI
job is the first execution of these tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows keystore security validation by accurately matching owners and trustees.
  * Added clearer error messages when Windows account or trustee resolution fails.
  * Improved handling of security descriptor resources to prevent validation issues.

### DIFF
--- a/keystore/keyfile_windows.go
+++ b/keystore/keyfile_windows.go
@@ -19,7 +19,6 @@ package keystore
 import (
 	"fmt"
 	"os"
-	"runtime"
 	"strings"
 
 	"golang.org/x/sys/windows"
@@ -64,9 +63,10 @@ func isKnownNonGrantACEType(aceType string) bool {
 // SDDL string and rejects files that grant access to Everyone,
 // the BUILTIN\Users group, or Authenticated Users.
 //
-// The implementation intentionally avoids the unsafe package to
-// prevent heap corruption caused by Go 1.24+ GC interacting with
-// uintptr-based SID handles (see https://go.dev/issue/73199).
+// The implementation stays inside golang.org/x/sys/windows and does not use
+// the unsafe package, so security descriptor and SID lifetimes stay owned by
+// that package. https://go.dev/issue/73199 was a dangling pointer inside
+// those helpers and was fixed there; hand-rolling them here would reopen it.
 func checkFilePermissions(path string) error {
 	sd, err := windows.GetNamedSecurityInfo(
 		path,
@@ -80,12 +80,8 @@ func checkFilePermissions(path string) error {
 			err,
 		)
 	}
-	// sd is Windows-allocated (LocalAlloc). Freeing it requires
-	// unsafe.Pointer, which we avoid due to Go 1.24+ heap
-	// corruption (go.dev/issue/73199). The ~200 B leak per call
-	// is acceptable: checkFilePermissions runs only at startup
-	// for a handful of key files.
-
+	// GetNamedSecurityInfo frees the Windows heap descriptor itself and
+	// returns a Go-heap copy, so sd needs no explicit free here.
 	return checkSecurityDescriptor(path, sd)
 }
 
@@ -207,20 +203,21 @@ func checkOpenDACL(path, owner, dacl string) error {
 			continue
 		}
 		isOwner, resolveErr := trusteeOwnerMatch(fields[5], owner)
-		if isOwner {
-			continue
-		}
 		if resolveErr != nil {
-			// The trustee could not be resolved to a SID, so we cannot tell
-			// an owner alias from a third party. Say so rather than
-			// reporting it as an unexpected trustee: the two need different
-			// fixes and the message is the only signal available when the
-			// failure only reproduces on another platform.
+			// One side did not resolve to a SID, so an owner alias cannot be
+			// told from a third party. Report that instead of an unexpected
+			// trustee: the two need different fixes, and the wrapped cause
+			// names which side failed, because an owner string that does not
+			// parse arrives here too. Checked before isOwner so that a
+			// future (true, err) return cannot turn a denial into a grant.
 			return fmt.Errorf(
-				"key file %q has DACL trustee %s that could not be resolved "+
-					"to a SID (owner %s): %v: %w",
+				"key file %q has DACL trustee %s that could not be compared "+
+					"with owner %s: %w: %w",
 				path, fields[5], owner, resolveErr, ErrInsecureFileMode,
 			)
+		}
+		if isOwner {
+			continue
 		}
 		return fmt.Errorf(
 			"key file %q grants access to unexpected trustee %s, "+
@@ -249,7 +246,8 @@ func ownerIsBuiltinAdministrators(owner string) bool {
 // A failed resolution is returned as an error rather than folded into a false
 // result, because "could not resolve this alias" and "this trustee is not the
 // owner" need different fixes and the caller's message is the only diagnostic
-// available when the failure reproduces on Windows CI and nowhere else.
+// available when the failure reproduces on Windows CI and nowhere else. It
+// never returns (true, non-nil), so a caller may treat any error as no match.
 //
 // An alias is resolved by asking Windows, not by matching its well-known RID.
 // The local administrator alias (LA) and a domain Administrator share RID 500
@@ -265,41 +263,53 @@ func trusteeOwnerMatch(trustee, owner string) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("parse owner SID %q: %w", owner, err)
 	}
-	trusteeSID, keepAlive, err := resolveTrusteeSID(trustee)
+	trusteeSID, err := resolveTrusteeSID(trustee)
 	if err != nil {
 		return false, err
 	}
-	equal := windows.EqualSid(trusteeSID, ownerSID)
-	// trusteeSID may point into keepAlive's allocation; see go.dev/issue/73199.
-	runtime.KeepAlive(keepAlive)
-	runtime.KeepAlive(ownerSID)
-	return equal, nil
+	// No runtime.KeepAlive is needed for either SID. StringToSid returns
+	// (*SID).Copy, a pointer to the base of a Go []byte, and the alias path
+	// returns an interior pointer into a Go-heap descriptor copy; the GC
+	// tracks both. windows.EqualSid then passes both through
+	// syscall.SyscallN, which is //go:uintptrkeepalive, so the compiler
+	// retains both arguments for the duration of the call.
+	return windows.EqualSid(trusteeSID, ownerSID), nil
 }
 
 // resolveTrusteeSID converts an SDDL trustee to a SID, accepting either a SID
 // string or an alias such as LA. Aliases are machine-relative, so they are
 // resolved by Windows rather than reconstructed here.
-func resolveTrusteeSID(
-	trustee string,
-) (*windows.SID, *windows.SECURITY_DESCRIPTOR, error) {
+//
+// ConvertStringSidToSid, behind windows.StringToSid, already accepts the SID
+// string constants, so the first branch handles both canonical SIDs and every
+// alias Windows renders into SDDL. The descriptor round-trip is a fallback for
+// a token that only the SDDL parser accepts, and the error return is for a
+// token neither accepts, which the caller must not treat as a non-owner.
+func resolveTrusteeSID(trustee string) (*windows.SID, error) {
 	if sid, err := windows.StringToSid(trustee); err == nil {
-		return sid, nil, nil
+		return sid, nil
 	}
 	sd, err := windows.SecurityDescriptorFromString("O:" + trustee)
 	if err != nil {
-		return nil, nil, fmt.Errorf(
+		return nil, fmt.Errorf(
 			"resolve trustee alias %q via security descriptor: %w",
 			trustee, err,
 		)
 	}
 	sid, _, err := sd.Owner()
 	if err != nil {
-		return nil, nil, fmt.Errorf(
+		return nil, fmt.Errorf(
 			"read owner from resolved trustee alias %q: %w", trustee, err,
 		)
 	}
-	// The caller must keep sd alive for as long as sid is used.
-	return sid, sd, nil
+	if sid == nil {
+		return nil, fmt.Errorf(
+			"resolved trustee alias %q has no owner SID", trustee,
+		)
+	}
+	// sid is an interior pointer into sd's Go-heap allocation, which keeps
+	// that allocation reachable, so sd need not be returned alongside it.
+	return sid, nil
 }
 
 func sddlSection(sddl, section string) string {
--- a/keystore/keyfile_windows.go
+++ b/keystore/keyfile_windows.go
@@ -19,6 +19,7 @@ package keystore
 import (
 	"fmt"
 	"os"
+	"runtime"
 	"strings"
 
 	"golang.org/x/sys/windows"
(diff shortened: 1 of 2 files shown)


keystore/keyfile_windows.go
```diff
@@ -187,60 +188,94 @@ func checkOpenDACL(path, owner, dacl string) error {
...
+		return false, fmt.Errorf("parse owner SID %q: %w", owner, err)
 	}
-	trusteeSID, err := resolveTrusteeSID(trustee)
+	trusteeSID, keepAlive, err := resolveTrusteeSID(trustee)
 	if err != nil {
-		return false
+		return false, err
 	}
-	return windows.EqualSid(trusteeSID, ownerSID)
+	equal := windows.EqualSid(trusteeSID, ownerSID)
```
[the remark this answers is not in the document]
cryptodj413: Confirmed independently against head 01cf3904 — this is still open, not addressed by the two commits after this comment.

`resolveTrusteeSID`'s first branch (`windows.StringToSid(trustee)`) is the live path for every real DACL trustee, alias or canonical SID: `ConvertStringSidToSid` accepts both the `S-R-I-S-S…` format and the two-letter SID string constants (`LA`, `BA`, …), so the `SecurityDescriptorFromString("O:"+trustee)` fallback at line 288 is effectively unreachable for anything Windows itself renders into an SDDL string. On that live path `resolveTrusteeSID` returns `keepAlive == nil` (line 285-286), and `StringToSid` itself returns a SID from `(*SID).Copy()`, i.e. `sid2 := (*SID)(unsafe.Pointer(&b[0]))` over a fresh `make([]byte, …)` — a manufactured pointer the runtime does not track as keeping `b` alive across a later `uintptr`-based syscall (go.dev/issue/73199, the exact issue this PR cites). `EqualSid`'s generated wrapper casts both arguments through `uintptr(unsafe.Pointer(...))` before the syscall, and only `keepAlive` (nil here) and `ownerSID` are kept alive after — not `trusteeSID`.

So on the path this PR actually exercises for every trustee, `trusteeSID` is unprotected across the syscall it's compared in. Suggest adding `runtime.KeepAlive(trusteeSID)` unconditionally alongside the existing two KeepAlive calls, matching how `ownerSID` is already handled.

keystore/keyfile_windows_test.go
```diff
@@ -195,6 +195,45 @@ func TestSecureFileModeWindows(t *testing.T) {
...
+	sd, err := windows.SecurityDescriptorFromString("O:BAD:P(A;;GA;;;LA)")
+	require.NoError(t, err)
+	dacl, _, err := sd.DACL()
+	require.NoError(t, err)
+	ownerSID, _, err := sd.Owner()
+	require.NoError(t, err)
+	require.NoError(t, windows.SetNamedSecurityInfo(
+		testFile,
+		windows.SE_FILE_OBJECT,
+		windows.OWNER_SECURITY_INFORMATION|windows.DACL_SECURITY_INFORMATION,
```
cryptodj413: This omits `windows.PROTECTED_DACL_SECURITY_INFORMATION`, unlike `setOwnerOnlyDACL` above (line 70-71) which sets it. Without it, `SetNamedSecurityInfo` can re-propagate inheritable ACEs from the parent of `t.TempDir()` (i.e. `%TEMP%`) into this file's DACL, so the DACL actually validated may not be the protected `O:BAD:P(A;;GA;;;LA)` from the SDDL literal at line 207. It passes today because GitHub's `windows-latest` runner happens to run as the built-in Administrator; on a host where the test account is a named admin, the inherited ACE would carry that account's own SID and `checkOpenFilePermissions` would reject it, failing this "accepted" test. Suggest adding `windows.PROTECTED_DACL_SECURITY_INFORMATION` to the flags here.

keystore/keyfile_windows.go
```diff
@@ -157,6 +158,21 @@ func checkOpenDACL(path, owner, dacl string) error {
...
+	//
+	// The condition is what keeps this narrow. An LA ace on a file owned by
+	// some other principal stays rejected, including the case a domain
+	// Administrator shares RID 500 with the local one and differs only in the
+	// domain part, which trusteeOwnerMatch must not conflate.
+	//
+	// Hosts where files are owned by the Administrators group produce owner BA
+	// with an LA ace, and so do GitHub's Windows runners.
+	if ownerIsBuiltinAdministrators(owner) {
+		allowed["LA"] = true
```
cryptodj413, line 170: Coverage gap, not a defect: no test pins that this gate stays narrow — e.g. `checkOpenDACL` with `owner == BA` and a *different* non-owner trustee (a plain `S-1-5-21-…` SID, not `LA`) should still return `ErrInsecureFileMode`. `TestAdministratorAccountACERejectedWhenOwnerIsNotAdministratorsWindows` only covers a non-BA owner; it doesn't cover a BA owner with an unrelated trustee.

keystore/keyfile_windows.go
```diff
@@ -187,60 +203,103 @@ func checkOpenDACL(path, owner, dacl string) error {
...
 		}
-		if allowed[fields[5]] || trusteeIsOwner(fields[5], owner) {
+		if allowed[fields[5]] {
+			continue
+		}
+		isOwner, resolveErr := trusteeOwnerMatch(fields[5], owner)
+		if isOwner {
 			continue
 		}
+		if resolveErr != nil {
```
cryptodj413, line 206: No test exercises this branch. Also worth knowing for the diagnostic's accuracy: `ConvertStringSidToSid` (used by `windows.StringToSid` in `resolveTrusteeSID`'s first branch) already accepts the two-letter SID string constants directly, so this `resolveErr` path — meant for "an alias fails to resolve on another platform" — is reachable only for a trustee token that is neither a valid SID nor a known alias, not for the cross-platform-alias scenario described in the comment above `trusteeOwnerMatch`. Still fails closed either way, so no security impact; the diagnostic just won't fire for the case its wording implies.

cryptodj413 (commented):
Reviewed against head `01cf3904`. Verified findings by reading the current source (not just relaying bot output), the vendored `golang.org/x/sys/windows` implementation, and the repo's CI configuration.

**Behavioral/security:**
1. The open Cubic thread on `keyfile_windows.go:272` is still valid on this head — replied there with the confirming detail. `trusteeSID` is unprotected across the `EqualSid` syscall on the path every real trustee resolves through (`keepAlive == nil`), because `ConvertStringSidToSid` already accepts both raw SIDs and alias constants, so the `SecurityDescriptorFromString` fallback rarely runs. Recommend adding `runtime.KeepAlive(trusteeSID)` unconditionally before merge — this is the class of bug the PR itself is fixing.

**Test coverage / correctness of tests (non-blocking, inline comments below):**
2. `TestAdministratorAccountACEAcceptedWindows` doesn't set `PROTECTED_DACL_SECURITY_INFORMATION`, so it isn't actually validating the protected DACL its SDDL literal describes.
3. No test pins that the `LA` allowance stays scoped to `owner == BA` for a non-owner trustee, and no test exercises the new `resolveErr` diagnostic branch.

**Everything else checked, no findings:**
- Access-control logic itself (BA/LA/CO/OW/SY semantics, fail-closed error handling, no info leak) is sound — the new allowance is a strict subset of what the pre-existing unconditional `BA` allowance already granted.
- No cross-repo callers of these package-private helpers.
- `ARCHITECTURE.md`/`DATABASE.md`: unaffected, already describe `keystore` at the right level of abstraction; no update needed.
- `go.mod`/dependencies: unchanged.

**Skipped-check note:** PR CI (`go-test.yml`) is Linux-only; this file and its new tests never run there. `go-test-windows.yml` only triggers on `workflow_dispatch`/push-to-`main` — someone manually dispatched it against this exact head (run `32436507173`, success), so there is a real green Windows signal for `01cf3904`, but a further push to this branch won't automatically re-trigger it.

Not my PR to merge; team `Core` review is already requested and still pending.

keystore/keyfile_windows_test.go
wolf31o2: Already fixed at head `2f1c4e65`, which is later than the reviewed `01cf3904`. The flag set in `TestAdministratorAccountACEAcceptedWindows` is now:

```go
windows.OWNER_SECURITY_INFORMATION|
    windows.DACL_SECURITY_INFORMATION|
    windows.PROTECTED_DACL_SECURITY_INFORMATION,
```

The reasoning is recorded in a comment above the call, including the named-admin host case, so the next reader does not have to rediscover why the flag is there.

keystore/keyfile_windows.go
wolf31o2: Already covered at head `2f1c4e65`, which is later than the reviewed `01cf3904`. `TestAdministratorsOwnerRejectsForeignTrusteeWindows` (`keystore/keyfile_windows_test.go`) pins exactly the pair described:

```go
owner := administrators.String()  // resolved from "O:BA", not a literal
assert.NoError(t, checkOpenDACL("k.skey", owner, "(A;;GA;;;LA)"))
assert.ErrorIs(t, checkOpenDACL(
    "k.skey", owner, "(A;;GA;;;S-1-5-21-1-2-3-1001)",
), ErrInsecureFileMode)
```

Both directions in one test, so the accept and the reject cannot drift apart. The owner is resolved through `SecurityDescriptorFromString("O:BA")` rather than a hardcoded `S-1-5-32-544`, matching how the production path resolves it.

keystore/keyfile_windows.go
wolf31o2: Both points already handled at head `2f1c4e65`, which is later than the reviewed `01cf3904`.

Branch coverage: `TestUnresolvableTrusteeDistinguishedFromForeignWindows` drives both arms with host-independent inputs — `(A;;FA;;;@@)` for the `resolveErr` path, `(A;;FA;;;S-1-5-21-1-2-3-1002)` for the foreign-trustee path — and asserts each message excludes the other's text. `TestOwnerSIDParseFailureNotBlamedOnTrusteeWindows` covers the owner side and additionally pins that the cause stays unwrappable (`errors.As` to `interface{ Unwrap() []error }`, length 2) rather than flattened by `%v`.

The `ConvertStringSidToSid` point is correct and the reachability is now documented where it applies, on `resolveTrusteeSID`:

> ConvertStringSidToSid, behind windows.StringToSid, already accepts the SID string constants, so the first branch handles both canonical SIDs and every alias Windows renders into SDDL. The descriptor round-trip is a fallback for a token that only the SDDL parser accepts, and the error return is for a token neither accepts, which the caller must not treat as a non-owner.

The misleading "an alias fails to resolve on another platform" wording is gone from `trusteeOwnerMatch`.

keystore/keyfile_windows.go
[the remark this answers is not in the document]
wolf31o2: Observation about `keepAlive` is accurate for `01cf3904`, but the conclusion does not hold, and head `2f1c4e65` resolves it by removing all three `KeepAlive` calls rather than adding a fourth. Two independent reasons, both checked against the toolchain rather than inferred:

**1. `EqualSid` already retains both arguments.** `syscall.SyscallN` carries `//go:uintptrkeepalive` (`src/syscall/dll_windows.go:97` in go1.26.7, alongside `//go:nosplit`). That pragma exists for precisely this pattern: an argument written as `uintptr(unsafe.Pointer(x))` *in the call expression* is kept alive by the compiler for the duration of the call. `EqualSid` (`x/sys@v0.47.0 windows/zsyscall_windows.go:809`) does exactly that for both:

```go
r0, _, _ := syscall.SyscallN(procEqualSid.Addr(), uintptr(unsafe.Pointer(sid1)), uintptr(unsafe.Pointer(sid2)))
```

So a caller-side `runtime.KeepAlive` after the call adds nothing — including the pre-existing one on `ownerSID`, which is why it went too.

**2. `StringToSid`'s result is a tracked pointer, not a manufactured one.** `(*SID).Copy` (`security_windows.go:248`) is `b := make([]byte, …)` then `(*SID)(unsafe.Pointer(&b[0]))`. `&b[0]` is a pointer to the *base* of a live Go heap allocation; converting it to `*SID` keeps that allocation reachable under the unsafe.Pointer rules. go.dev/issue/73199 was a dangling pointer *inside* these helpers, fixed there — it is not a statement that their returned SIDs are untracked.

The alias branch is the same story from the other side: the returned SID is an interior pointer into the descriptor's Go-heap allocation, so the allocation stays reachable and the descriptor no longer has to be returned alongside it.

`trusteeOwnerMatch` now carries this as a comment so the calls are not reintroduced. Note also that both threads on this line reach opposite fixes for the same code — the cubic suggestion adds `KeepAlive(trusteeSID)`, this one adds it unconditionally — which is what prompted checking the pragma instead of picking one.

keystore/keyfile_windows_test.go
cryptodj413: Fixed in 2f1c4e6 — the acceptance test now passes `windows.PROTECTED_DACL_SECURITY_INFORMATION` alongside the existing flags, matching `setOwnerOnlyDACL`. Verified by checking out this file's content at the new head and confirming the diff.

keystore/keyfile_windows.go
cryptodj413: Fixed in 2f1c4e6 — `TestAdministratorsOwnerRejectsForeignTrusteeWindows` now pins that a BA-owned file with a foreign (non-LA) trustee is still rejected, so the gate can't silently widen.

keystore/keyfile_windows.go
cryptodj413: Fixed in 2f1c4e6 — `TestUnresolvableTrusteeDistinguishedFromForeignWindows` and `TestOwnerSIDParseFailureNotBlamedOnTrusteeWindows` now cover this branch, including that the wrapped cause stays unwrappable via the new multi-`%w`.

keystore/keyfile_windows.go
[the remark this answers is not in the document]
cryptodj413: Checked this independently against go1.26's `syscall.SyscallN` (`//go:uintptrkeepalive`, `dll_windows.go:97`) and the compiler's call-site rewrite for that pragma (`cmd/compile/internal/escape/call.go: rewriteArgument`) — it keeps a pointer-to-uintptr argument alive through any wrapper, not just a function that itself carries the pragma, as long as the conversion is a direct call-argument expression. That's exactly `EqualSid`'s shape, so the correction holds and 2f1c4e6's removal is right rather than a regression. Also rebuilt the two changed files at this head under `GOOS=windows go vet/build` and `go test -c` — clean. Agreed, resolved.

cryptodj413 (commented):
Re-reviewed against head `2f1c4e65`. All three findings from the prior review pass are addressed:

- The KeepAlive concern (mine and Cubic's) doesn't hold after all — verified independently against `syscall.SyscallN`'s `//go:uintptrkeepalive` pragma and the compiler's call-site argument rewrite (`cmd/compile/internal/escape/call.go`): `EqualSid` already keeps both SID arguments alive for the duration of the syscall, so the KeepAlive removal in this commit is correct, not a regression. Replied on that thread with the detail and resolved it.
- The missing `PROTECTED_DACL_SECURITY_INFORMATION` in `TestAdministratorAccountACEAcceptedWindows` is fixed.
- The coverage gaps (BA-owner-with-foreign-trustee, the `resolveErr` branch) are now covered by `TestAdministratorsOwnerRejectsForeignTrusteeWindows`, `TestUnresolvableTrusteeDistinguishedFromForeignWindows`, and `TestOwnerSIDParseFailureNotBlamedOnTrusteeWindows`.

Rebuilt `keystore/keyfile_windows.go` and `keystore/keyfile_windows_test.go` at this head under `GOOS=windows GOARCH=amd64`: `go vet`, `go build`, and `go test -c` all clean; `gofmt -l` clean. All three of my inline threads resolved.

Skipped-check note: `go-test-windows.yml` hasn't been re-dispatched for `2f1c4e65` yet (last manual dispatch was against `01cf3904`), so this head has no actual Windows execution recorded yet, only static verification.

No further findings from me. Not my PR to merge; team Core review is still pending.

cryptodj413 (approved):

Issue #3061, opened 2026-08-01
ledger: fresh preview from-genesis stalls at slot 172780 rejecting canonical blocks as genesis-overlay-not-active (regression)


wolf31o2:
A fresh preview from-genesis sync on current main stalls at slot 172780 (epoch 1 to 2 boundary) and cannot advance, rejecting the canonical continuation as genesis-overlay-not-active. This blocks all from-genesis sync.

## Reproduction

Build current main (tested at 583f91d5), run a fresh preview from-genesis:

    dingo serve --network preview --data-dir <fresh> ...

Within about 90 seconds the node reaches slot 172780 (blockNum ~8639) and then stalls. It repeatedly logs, for the canonical blocks that follow, across every peer that serves them:

    level=ERROR msg="failed to handle block" component=ledger
      error="block header crypto verification failed: block header verification rejected at slot 172836:
             slot is reserved for the genesis overlay schedule but not active"

Rejected slots include 172836, 172855, 172921. The node recycles each peer after the header-verification failure ("recycling connection after header verification failure"), so it drops every peer serving the canonical chain and never advances past 172780. slotNum_int stays flat at 172780; the header/overlay rejection count climbs indefinitely.

## Regression bounds

- Good: 6a67ba38 syncs preview from genesis past this point without issue (reached epoch 1375 in earlier runs).
- Bad: 583f91d5 (current main) stalls at 172780 as above.

The error is raised at ledger/verify_header.go:429 (status genesisOverlayNonActive from genesisOverlayDelegationForSlot). No commit in 6a67ba38..583f91d5 modifies verify_header.go or the genesisOverlayDelegationForSlot / genesisDelegationActiveForSlot functions, so the regression is indirect: something the overlay-activity determination depends on changed. Candidates in the window touch the leader schedule and consensus eligibility (for example #3042 changed the active slot coefficient handling) and protocol/validation behavior (#3039, #3040). A bisect to the exact commit is in progress and will be added as a follow-up.

## Impact

Blocks every from-genesis sync on current main (preview confirmed; preprod and mainnet almost certainly affected at their equivalent early-overlay boundary). It also blocks validation of any other from-genesis work against current main.

wolf31o2:
Bisected. First bad commit is 6573ab1d (fix: enforce full-node validation trust semantics, #3039). Parent 37d8448d (#3041) is good.

[...]

PR #3251, opened 2026-08-21
fix(ledger): recover from deterministic replay tx failures

wolf31o2:
## Problem

Preview genesis replay repeatedly rewinds and redelivers the same primary-chain block when duplicate-input validation produces a deterministic rejection.

## Changes

- Classify Shelley and Byron duplicate-input failures as deterministic.
- Rewind the primary chain and ledger to the last applied tip and request a fresh ChainSync intersection.
- Key the resync latch by applied tip, failing block, and transaction so repeated delivery does not spend additional peer rotation.
- Leave terminal no-progress handling to #3297.

The duplicate-input verdict on canonical pre-Conway blocks is an upstream false positive tracked by blinklabs-io/gouroboros#1989. That fix unblocks preview genesis sync; this PR prevents the local recovery path from wedging permanently in the meantime.

## Validation

Race-enabled ledger tests and conformance coverage pass for deterministic rewind and repeated rejection.


ledger/replay_recovery.go
```diff
@@ -343,6 +347,108 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+			err,
+		)
+	}
+	if ls.config.EventBus != nil {
+		ls.config.EventBus.Publish(
+			event.ChainsyncResyncEventType,
+			event.NewEvent(
+				event.ChainsyncResyncEventType,
+				event.ChainsyncResyncEvent{
+					Reason: event.
```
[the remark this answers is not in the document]
wolf31o2: Observation correct, stated consequence refuted — and applying the fix would regress this path. Declining, with the evidence.

`ConnectionId` is indeed left zero. But the claim that the fresh intersection needs it inverts what the consumer does (`ouroboros/chainsync.go:1226-1233`):

```go
if e.ConnectionId != (ouroboros.ConnectionId{}) {
    connIds = append(connIds, e.ConnectionId)
} else if o.chainsyncState != nil {
    connIds = o.chainsyncState.RewindTrackedClientsTo(e.Point)
```

Setting it **skips** `RewindTrackedClientsTo`. Left unset, `chainsync/chainsync.go:787` rewinds and returns every tracked connection whose cursor is ahead of the point — a superset that includes the peer which delivered the rejected block, and it also realigns the other peers' cursors, which is that function's stated purpose ("so peers do not strand us in AwaitReply on a stale higher cursor"). Since `ChainsyncResyncReasonDeterministicTxValidationRecovery` is in `chainsyncResyncRequiresFreshConnection`, every returned connection is closed.

On the precedent: it is split by intent, not a convention. The directly analogous `ChainsyncResyncReasonLiveTxValidationRecovery` at `replay_recovery.go:866` also leaves it unset; `publishReplayRecoveryNonConvergingResync` at `:536` sets it deliberately.

ARCHITECTURE.md
```diff
@@ -2964,6 +2964,14 @@ paths finish bringing metadata to its common ancestor. Live at-tip recovery uses
 the same bounded event-aware helper; startup-only speculative-tail cleanup stays
 eventless because subscribers have not begun consuming live chain events.
 Rewinding metadata alone would replay the same corrupt chain indefinitely.
+Transaction-structure failures are handled separately from state-dependent
```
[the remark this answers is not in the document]
wolf31o2: Correct, and auditing the class rather than the sentence turned up a real defect behind it. Both fixed in `9ad7876d`.

The doc's "Conway" qualifier was wrong: in `gouroboros@v0.193.3` every era from Shelley through Dijkstra delegates to `shelley.UtxoValidateNoDuplicateInputs`, which returns `shelley.DuplicateInputError` for regular, collateral *or* reference inputs (`shelley/rules.go:95,104,113`). Nothing is Conway-specific. Corrected, and it now names all three input kinds rather than only reference inputs.

The defect the thread implies but does not state: **Byron does not use that rule.** `ledger/eras/byron.go:231` returns Dingo's own `eras.DuplicateInputByronError`, which `isDeterministicTxValidationError` did not match. Traced end to end — `state.go:5584` → `eras.ValidateTxByron` (`errors.Join`, so `errors.As` traverses) → `state.go:5692` wraps it in `*txValidationError` → `tryRecoverFromTxValidationError`. So a Byron duplicate input fell through to the state-dependent unresolved-producer path and its security-parameter fallback: precisely the loop the deterministic branch exists to break, in the era a from-genesis replay *starts* in.

`eras.DuplicateInputByronError` is now classified, with `TestReplayRecoveryRejectsDeterministicByronDuplicateInput`. Stubbing the Byron arm to `false` fails it on both the tip assertion and `replayRecoveryTipTracked`.

wolf31o2:
Addressed the unresolved source-connection review finding in 04293cd5. Deterministic replay recovery does not currently propagate a failing block's source ConnectionId through txValidationError, so it now preserves the zero-ID resync sentinel; the current chainsync consumer interprets that as rewinding all tracked clients at the recovery point. Added a regression with an active connection configured that asserts the emitted ID remains zero. Focused tests passed: go test -race ./ledger -run '^TestReplayRecovery' and go test -race ./ouroboros -run '^TestSubscribeChainsyncResync'.

ledger/state.go
```diff
@@ -4134,6 +4141,9 @@ func (ls *LedgerState) ledgerProcessBlocks(ctx context.Context) {
 			return
 		}
 		ls.handleLedgerProcessBlocksError(err)
+		if errors.Is(err, errHaltLedgerPipeline) {
+			return
```
cryptodj413: This clears `dingo_ledger_pipeline_stuck` (and `pipelineNoProgressRestarts`) to 0 at the exact moment the pipeline permanently dies. `ledgerProcessBlocks` (`state.go:4127`) is started once with no restart path, and its `defer ls.metrics.setPipelineNoProgress(0, false)` at line 4132 fires on *every* exit, including this new one — its own comment only anticipates "normal return, or a shutdown cancelling one of the retry timers," not a permanent halt. `dingo_ledger_pipeline_stuck`'s help text is "1 while the ledger pipeline ... (operator intervention required)" (`metrics.go:374-377`), and `TestHandleLedgerProcessBlocksErrorLogsPersistentValidationFailure` (`state_test.go:79-95`) already pins that `FatalErrorFunc` is *not* called on this path either. Net effect: the node stops following the chain while every metric says it's healthy, on the exact condition this PR's own log line calls "operator intervention is required." Since this PR introduces the only reachable producer of `errHaltLedgerPipeline`, it owns this consequence. Cheap fix: set a distinct `dingo_ledger_pipeline_halted` gauge (or skip the deferred reset) before the `return` at line 4145.

ledger/replay_recovery.go
```diff
@@ -343,6 +348,150 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+				"deterministic transaction validation failure persisted after a fresh ChainSync intersection; halting ledger pipeline",
+				"component", "ledger",
+				"tx_hash", hex.EncodeToString(validationErr.TxHash),
+				"failing_block_slot", validationErr.BlockPoint.Slot,
+				"ledger_tip_slot", ledgerTip.Point.Slot,
+				"ledger_tip_hash", hex.EncodeToString(ledgerTip.Point.Hash),
+				"hint", "network peers are serving a transaction this node rejects; operator intervention is required",
+			)
+		}
+		return false, errHaltLedgerPipeline
```
cryptodj413: This is the crux I'd want confirmed before merge, not a code-level bug: the PR's own problem statement says preview "repeatedly rewinds and redelivers the same primary-chain block" — a block every peer keeps serving on the *primary* chain is, by definition, one the network has canonicalized. If dingo's `shelley.DuplicateInputError` verdict on it is a local false positive rather than a genuine invalid-tx signal, this line turns an (already bad) infinite retry loop into a *permanent, unrecoverable* halt on a valid chain — the opposite of what `header_validation_recovery.go:40-43`'s stated principle argues for ("the correct response to a block a node believes is invalid is to reject the chain ... not to halt on it").

There's a concrete mechanism that could produce exactly that false positive: `shelley.UtxoValidateNoDuplicateInputs` (gouroboros `ledger/shelley/rules.go:90-98`) dedupes `tx.Inputs()` unconditionally for every era, but I confirmed `ShelleyTransactionBody.TxInputs` decodes via `SetType[T].UnmarshalCBOR` (`cbor/tags.go:210-230`), which preserves every wire-level entry verbatim into `t.items` regardless of whether the CBOR was tag-258-wrapped — and gouroboros' own `SetType.CheckForDuplicates` (`cbor/tags.go:258-278`) explicitly skips untagged (pre-Dijkstra) arrays "so that pre-Dijkstra eras are unaffected." That carve-out is for a different function, but it's evidence the maintainers already know pre-Dijkstra `Set`-typed fields can carry wire-level duplicates that aren't errors — which is exactly cardano-node's own historical Haskell `Set`-decode-coalesces-duplicates behavior for pre-Alonzo/pre-Conway tx bodies, as far as I'm aware. If that's what's happening on the specific preview block behind this PR, the fix belongs in the rule's era applicability (or gouroboros' decode), not in a permanent halt — precedent for that shape of fix is the closed issue #3061 ("fresh preview from-genesis stalls ... rejecting canonical blocks", fixed by fixing the rule).

Could you confirm the specific tx/block this was observed on and whether `cardano-node` accepts it? If it does, I'd hold this PR's halt behavior pending that fix rather than ship a permanent stop on a false positive.

ledger/replay_recovery.go
```diff
@@ -187,6 +189,9 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 	if !errors.As(err, &validationErr) {
 		return false, nil
 	}
+	if isDeterministicTxValidationError(validationErr.Cause) {
+		return ls.recoverFromDeterministicTxValidationError(validationErr)
```
cryptodj413, line 212: Two related, non-blocking notes on scope here:

1. This check runs *before* `ls.IsAtTip()` below, so a duplicate-input failure at tip no longer takes `recoverAtTipFromTxValidationError`'s escalating-depth path (which never halts) — it now gets one rewind and then halts on a second occurrence at the same tip. Neither new test reaches tip (`newReplayRecoveryAuditLedger` never sets `reachedTip`), so this transition is untested, and it's a bigger behavior change than the PR description's "replay recovery" framing suggests.
2. The one-resync latch (`ls.deterministicTxRecoveryAttempted` / `TipSlot`, set at lines ~464-465) is keyed on tip slot alone, not on the failing block's identity. After one resync, chain selection could deliver a genuinely different branch; a *different* deterministic-invalid block on that branch, hit while the tip is still at or below the recorded slot, halts immediately with no fresh-intersection attempt of its own. The sibling at-tip path keys its escalation on failing-block identity (`atTipRecoveryAttempt.matches`) rather than tip slot for this reason. Given the halt is permanent (see the other inline comment), keying on `(tipSlot, failing block hash)` would remove a false-halt path.

cryptodj413 (commented):
Reviewed independently in a fresh worktree at `04293cd5` (merge-base `origin/main@5a54b7f8`): `go build ./...`, `go vet`, `go test -race ./ledger ./ouroboros ./chain ./event` all pass; `make docs-parity` passes; `DATABASE.md` untouched (expected — no schema/storage change, just in-memory recovery-latch fields). Disabled `isDeterministicTxValidationError` (forced `return false`) and reconfirmed both new tests then fail on both the Shelley and Byron branches, so they're non-vacuous. The three resolved bot threads (declined zero-`ConnectionId` "fix", the Conway-qualifier doc fix that surfaced the real Byron-classification gap, and the later `ConnectionId: activeConnId` walk-back) all check out against the current diff.

Two behavioral items below (inline), one of them a likely blocker; several non-blocking items; and a validation-scope note.

**Non-blocking, no defect, but worth linking:** I checked whether the Mithril-boundary silent decline (`recoverFromDeterministicTxValidationError`'s `if ls.recoveryRollbackExceedsMithrilBoundary(rewindPoint) { return false, nil }`, no logging) is a gap — it isn't; it's a verbatim copy of the sibling at `ledger/header_validation_recovery.go:125-127`, the rich-logging helper it was compared against can't be reused here (it would rewind *below* the boundary it's supposed to guard), and the branch is barely reachable post-Mithril-bootstrap. The class of problem it does represent — a deterministic failure inside the Mithril trust window with no recovery path — is already tracked in open issue #3261, which names this PR. Worth linking #3261 from the PR description.

**Non-blocking:** I also traced the TOCTOU window between `ls.chain.ValidateRollback(rewindPoint)` and the actual rewind (chain selection can move concurrently) and the reuse of `rollbackPrimaryChainInSecurityParamWindows` for a same-branch rewind — both check out: `rollbackLocked` re-validates everything under lock before mutating, so the pre-check is an optimization, not a safety gate, and the windowing helper is required (not over-general) whenever the ledger tip trails the chain tip by more than `k`, which is a documented normal post-Mithril-bootstrap shape, not bounded by construction as I'd have guessed.

**Non-blocking, missing coverage:** `isDeterministicTxValidationError`'s check at `replay_recovery.go:192-194` runs before `ls.IsAtTip()`, so a duplicate-input failure *at tip* no longer takes the escalating `recoverAtTipFromTxValidationError` path (which never halts) — it gets one rewind, then halts on a repeat at the same tip. Both new tests use a ledger that never reaches tip, so this transition is untested, and the new `ARCHITECTURE.md` paragraph says only "**replay** recovery therefore rejects the primary-chain branch," which under-describes the at-tip case. Also untested: a `Cause` join containing a duplicate-input error alongside a genuine `ErrInputResolution`/`ErrReferenceInputResolution` — the deterministic branch wins unconditionally today, discarding the information the producer-resolution path would have used to find a different, valid rewind point.

**Validation scope:** no existing DevNet scenario or conformance suite would exercise a duplicate-input rejection + resync (checked `internal/test/devnet/scenarios/` and `internal/test/conformance/`), so skipping DevNet here doesn't skip a check that would have caught anything — but given one of the inline findings below turns this into a *permanent* halt, I'd still run `make test-devnet` once before merge purely as a liveness smoke test and record it, since the PR's Validation section currently lists only `go test ./ledger ./ouroboros`.

ledger/state.go
wolf31o2: Confirmed, and removed at the source rather than instrumented: this branch no longer produces `errHaltLedgerPipeline`.

The observation held on every point. `ledgerProcessBlocks` has the single caller at `state.go:1408` with no restart wrapper, `handleLedgerProcessBlocksError` logs at WARN and never reaches `FatalErrorFunc`, and the deferred `setPipelineNoProgress(0, false)` ran on the new return, so the pipeline exited with `dingo_ledger_pipeline_stuck` and `dingo_ledger_pipeline_no_progress_restarts` both at 0.

`recoverFromDeterministicTxValidationError` now stays non-terminal. A repeat rejection of the same failing block at the same applied tip rewinds the branch again but publishes no further resync, so the loop stays inside the accounting that already exists for it: `trackPipelineProgress` / `ledgerPipelineBackoff` escalate to `noProgressStuckBackoffMax` and raise `dingo_ledger_pipeline_stuck` with the ERROR announcement at `noProgressStuckThreshold`. The halt short-circuited that at the second occurrence, so it was strictly less observable than not halting.

`dingo_ledger_pipeline_halted` and the `ledgerProcessBlocksWithAttempt` restructuring stay in #3297 (issue #3261), which owns the terminal-state decision and its gauge. Nothing here duplicates them, and the remaining overlap with #3297 in `ledger/state.go` is the `LedgerState` field block and the `resetDeterministicTxRecovery` call, both keep-both-sides.

`TestReplayRecoveryRejectsDeterministicDuplicateInput` and `TestReplayRecoveryDeterministicDuplicateAtTipSkipsDescentSchedule` both fail when the halt is restored.

ledger/replay_recovery.go
wolf31o2: Confirmed on chain: the verdict is a local false positive on canonical blocks, and the halt is gone.

Preview block `98a969c42dc92c3d52c462c687e67df683418d7d597a4883d3187e03e17b41fe`, slot 1462320, height 67407, epoch 16, Babbage (`proto_major` 8), transaction index 2, hash `7bb40e40c3e6010ead628fd9ea62ae4f8acab340ccca26efacad826caa3be5b9`. Body field 18 (reference inputs) decodes as an untagged 4-element array:

```
f555b524c2b54b12bd44bb5189a7847f53d71fe635266c83a11d7e266a998638#0
f555b524c2b54b12bd44bb5189a7847f53d71fe635266c83a11d7e266a998638#0
f555b524c2b54b12bd44bb5189a7847f53d71fe635266c83a11d7e266a998638#1
ea1c03703fbd6430db77a645e45a019f121af1ec4479f6eb963d749cea00d0e2#0
```

`shelley.UtxoValidateNoDuplicateInputs` returns `DuplicateInputError{InputType: "reference"}` for it under v0.193.3. Slot 1462320 is the first block on preview where the rule fires: every block from origin to it was fetched and checked, and none earlier does. The next is slot 1468932, transaction `a9743a0e0991ebe931170433a79a2294a8fa57a7a73cf1214d074633739f3f20`, same duplicated reference input.

The spending-input set diverges the same way, and there the coalescing is directly observable. Block `68dac3790d43bc3331a3df70593cfee9910963b70d9b61e12566b8e6f8c176df`, slot 50778243, epoch 587, Babbage, transaction `4c4afd467a06e93891221141305c9242ed633e0788c9d8eb10fd66a3117e77b6`: body field 0 is an untagged 2-element array holding `38242dd3b8a4189fd854c61cdf9cd086b975d459f9dac664b1d0666ba0c0f6d9#0` twice, and Koios reports the transaction with one input, that entry, spent once.

Provenance: block bytes read over N2N blockfetch from `preview-node.play.dev.cardano.org:3001`, magic 2, from origin; Koios confirms both transactions in those blocks. cardano-node accepted both.

The mechanism is the pre-Conway untagged-array case rather than tag-258. `SetType.UnmarshalCBOR` keeps wire multiplicity verbatim, `CheckForDuplicates` skips untagged arrays, Conway tx-body sets are already rejected at decode (`conway.go:611-628`), and the rule alone deduplicates unconditionally for Shelley through Dijkstra. Filed as blinklabs-io/gouroboros#1989 with both blocks; the fix is the rule's era applicability.

So `header_validation_recovery.go:40-44` governs here: the branch is rejected and keeps being rejected, and the wedge surfaces through `dingo_ledger_pipeline_stuck` rather than a terminal stop.

ledger/replay_recovery.go
wolf31o2: All three applied.

1. The at-tip transition is covered by `TestReplayRecoveryDeterministicDuplicateAtTipSkipsDescentSchedule`, which sets `reachedTip` and asserts the deterministic branch runs with `lastAtTipRecovery` and `atTipRecoveryDescentCount` left unarmed. The ordering stands on its own terms — a deeper rewind cannot change a verdict that reads no UTxO history — and with the halt removed the at-tip case is no longer terminal either. The test fails when the deterministic check moves back below `IsAtTip()`.

2. The latch keys on `(applied tip slot, failing block point, tx hash)` through `deterministicTxRecoveryLatch.matches`, the same identity `atTipRecoveryAttempt.matches` uses, so a different rejected block on a newly selected branch gets its own fresh intersection. `TestReplayRecoveryDeterministicLatchIsPerFailingBlock` fails when the key is reduced to the tip slot.

3. On the write: not a live data race. `resetDeterministicTxRecovery` and the write both run on the ledger pipeline goroutine, and the reset is called from inside the `ls.Lock()` tip-advance section in `ledgerProcessBlocksFromSource` — which is what made the unlocked write inconsistent with the rest of that critical section rather than concurrent today. The write now goes through `markDeterministicTxRecoveryResync` under the write lock. `TestReplayRecoveryDeterministicLatchConcurrentAccess` reports `WARNING: DATA RACE` under `-race` with that lock removed.

ledger/replay_recovery.go
```diff
@@ -343,6 +348,157 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+			if connId := ls.config.GetActiveConnectionFunc(); connId != nil {
+				activeConnId = *connId
+			}
+		}
+		ls.config.EventBus.Publish(
+			event.ChainsyncResyncEventType,
+			event.NewEvent(
+				event.ChainsyncResyncEventType,
+				event.ChainsyncResyncEvent{
+					ConnectionId: activeConnId,
```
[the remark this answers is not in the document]
wolf31o2: Addressed in 04293cd and retained at the current head. Deterministic recovery leaves `ConnectionId` at its zero value, so the event consumer rewinds all affected tracked clients instead of closing only the active peer. `TestReplayRecoveryRejectsDeterministicDuplicateInput` pins the zero-ID event.

ledger/state.go
wolf31o2: Already addressed, re-checked at 2f547dd4 after merging `origin/main`.

This branch produces no `errHaltLedgerPipeline`; the only occurrence in its diff is the `assert.NotErrorIs` in `TestReplayRecoveryRejectsDeterministicDuplicateInput`. The two remaining producers come from #3297 (`rejectRecoveryAtMithrilBoundary`, `ledger/replay_recovery.go:1123` and `:1160`), and that path does not leave the gauges at 0: `ledgerProcessBlocks` sets `halted`, which skips the deferred `setPipelineNoProgress(0, false)`, then calls `setPipelineHalted()`, with `incMithrilTrustWindowUnrepairable()` on the recovery side (`ledger/state.go:4430-4452`).

The finding was correct for the revision it was written against.

ledger/replay_recovery.go
wolf31o2: Already addressed, re-checked at 2f547dd4 after merging `origin/main`.

The duplicate-input verdict on those canonical pre-Conway blocks is a local false positive, filed as blinklabs-io/gouroboros#1989. The halt is gone: `recoverFromDeterministicTxValidationError` rejects the branch and keeps rejecting it, which is the `header_validation_recovery.go` response this comment argues for.

The merge brings in #3297's bounded trust-window halt, which does not cover this path: when the rewind target would cross the Mithril anchor, `recoverFromDeterministicTxValidationError` returns unrecovered, so the failure propagates to the pipeline's restart-and-backoff accounting rather than to `errHaltLedgerPipeline`.

ledger/replay_recovery.go
wolf31o2: Already addressed, re-checked at 2f547dd4 after merging `origin/main`.

1. The at-tip transition is covered by `TestReplayRecoveryDeterministicDuplicateAtTipSkipsDescentSchedule` (`ledger/replay_recovery_test.go:1814`), which sets `reachedTip` and asserts the descent schedule stays unarmed. With the halt removed, the at-tip case is not terminal either.
2. `deterministicTxRecoveryLatch.matches` keys on `(applied tip slot, failing block point, tx hash)`, the same identity `atTipRecoveryAttempt.matches` uses, so a different rejected block on a newly selected branch gets its own fresh intersection. Covered by `TestReplayRecoveryDeterministicLatchIsPerFailingBlock`.
3. The latch write goes through `markDeterministicTxRecoveryResync` under the write lock, matching `resetDeterministicTxRecovery` inside the tip-advance critical section. Covered by `TestReplayRecoveryDeterministicLatchConcurrentAccess`.

Post-merge evidence: `go test -tags dingo_extra_plugins -race -timeout 20m ./ledger/... ./event/... ./ouroboros/...` passes with no data races, and the four deterministic tests fail when the `isDeterministicTxValidationError` dispatch in `tryRecoverFromTxValidationError` is disabled.

arepala-uml (approved):
Reviewed at current head. The deterministic/state-dependent split correctly classifies duplicate-input structural errors (Shelley-family and Byron) versus unresolved-producer failures, and the recovery path is deterministic and idempotent — verified with a negative test (disabling the classifier fails all five new tests). No masking of the underlying validation verdict; the false-positive root cause is tracked upstream (gouroboros#1989), not swallowed. Confirmed `go build`, `go test -race ./ledger/... ./ouroboros/... ./event/...` (full suite, no data races), and `go test ./internal/test/conformance/...` all pass.

Prior findings (Cubic P1–P3, review-thread questions on halt/gauge interaction and latch scoping) are verified resolved or false positives against current code.

Non-blocking: `ARCHITECTURE.md`'s new section doesn't call out the at-tip ordering change, though the code and tests handle it correctly.

PR #3266, opened 2026-08-21
fix(ledger): handle multi-era boundary blocks

wolf31o2:
## Problem
Prime-mainnet can present an Alonzo-shaped boundary block with a Babbage header protocol major. The ledger treated the epoch boundary as only Mary-to-Alonzo and rejected the header as protocol major 7 while expecting at most 6.

## Changes
- infer the successor era from a boundary header only for a validated adjacent era
- support at most two consecutive era transitions in one boundary transaction
- enact source-era protocol updates before applying the multi-era transition, then persist the final era and parameters
- add focused regression coverage and document the ledger flow

Validation: `go test ./ledger -count=1`; focused era/header tests; `make golines`; `make docs-parity`.

---
## Summary by cubic
Handles multi‑era boundary blocks and records the new epoch’s final protocol version. Previously we only allowed single‑step boundaries and captured the Mark snapshot before extra transitions, which rejected Prime mainnet Mary→Alonzo→Babbage boundaries and stored the source‑era major in `reward_snapshot.protocol_version`.

- Infers the boundary target era from the header protocol major only when it is the immediate successor; permits at most two transitions only with a validated successor header and rejects body‑only multi‑era jumps and larger jumps.
- Runs epoch rollover first to enact source‑era pparam updates, defers the authoritative Mark snapshot when further transitions remain, then applies the transitions and captures once so `reward_snapshot.protocol_version` matches the final era and the published `EpochTransitionEvent`.
- Persists the final era and protocol parameters before publishing events; updates docs to reflect the order and snapshot semantics.

**Review notes**
- See `ledger/state.go` for `eraTransitionPath`, `boundaryEraForBlock`, `applyBoundaryEraTransitions`, and the updated `processEpochRollover(..., deferBoundarySnapshot bool)`.
- `ledger/chainsync.go` extracts header protocol major to gate two‑step boundaries and splits transitions around rollover; new tests in `ledger/era_boundary_test.go` cover successor‑header validation and final‑era snapshot recording.

<sup>Written for commit 589dc76105dc64dd17a18e8fd78ca8600ea6ff84. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3266?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved protocol transitions at era boundaries, including consecutive transitions at a single epoch boundary.
  - Ensured source-era protocol settings are applied in the correct order during hard-fork changes.
  - Corrected boundary block handling to determine the successor era from its protocol metadata.
  - Added validation to reject unsupported or non-adjacent era transitions.
- **Documentation**
  - Documented era-boundary handling and supported transition limits.


ledger/state.go
```diff
@@ -4459,6 +4546,78 @@ func (ls *LedgerState) ledgerProcessBlocksFromSource(
 					return err
 				}
 				rolloverResult = result
+				if transitionsBeforeRollover == 0 && len(transitionPath) == 2 {
```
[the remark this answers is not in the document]
wolf31o2: Confirmed and fixed in 3853555.

`processEpochRollover` now takes a `deferBoundarySnapshot` flag. The two-era path sets it, and `applyBoundaryEraTransitions` captures the snapshot once, after `SetEpoch`/`SetPParams` have persisted the final era and parameters. Not re-run: a second call would write again under the same `epoch_boundary_snapshot` savepoint.

Impact is confined to `reward_snapshot.protocol_version`. Its only consumer is the stale-precompute fingerprint equality check in `ledger/reward_calculation.go`; no reward formula reads it, and `rewards.Parameters.ProtocolMajorVersion` is derived from pparams instead.

Coverage: `TestBoundaryEraTransitionsSnapshotRecordsFinalProtocolVersion` drives a Shelley->Allegra->Mary boundary and asserts the captured major is 4, and that the capture runs exactly once and agrees with `protocolMajorForEvent` on the same result. With the capture placed before the transitions it fails `expected: 0x4, actual: 0x2`. `TestSingleEraBoundaryRolloverCapturesSnapshotInRollover` pins the single-era path.

On `rolloverResult.HardFork` being nil: independent of this ordering. It is set only when the source era pparams enacted during the rollover bump the major version. The era changes themselves are carried in `eraTransitions`, and the block-era-driven `HardForkEvent` loop emits one event per transition with dedup against the pparam path, so no hard-fork event is lost for a two-era jump.

ledger/state.go
```diff
@@ -4407,34 +4487,42 @@ func (ls *LedgerState) ledgerProcessBlocksFromSource(
...
+				// When two eras are crossed at one boundary, let the epoch
+				// rollover enact pending protocol updates in the source era
+				// first. Applying both transitions before the rollover would
+				// let an old-era update overwrite the successor era's pparams
+				// (Prime's epoch-9 major-5 update otherwise regresses Babbage
+				// major 7).
+				transitionsBeforeRollover := len(transitionPath)
+				if transitionsBeforeRollover == 2 {
+					transitionsBeforeRollover = 0
+				}
```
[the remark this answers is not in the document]
wolf31o2: Not applying this one.

`eraTransitionPath` already compares list positions, not numeric IDs. It resolves both era IDs through `ls.eraList()` and rejects on `currentIndex < 0 || targetIndex <= currentIndex || targetIndex-currentIndex > 2`. An era ID absent from the list resolves to index -1 and is rejected by the same condition, so non-contiguous IDs are not a factor.

Normalizing a backward boundary era to `snapshotEra.Id` would drop a guard that predates this change. `eras.IsValidEraAdvancementIn` already rejected every backwards step, and for the documented reason: a boundary block from an era earlier than the ledger current era implies a non-conforming block, a malformed snapshot, or a chain-selection bug. Silently rolling over in the current era would mask that rather than surface it. This change only widens the forward allowance from one step to two and keeps the backwards and unknown-era rejections intact.

ledger/state.go
```diff
@@ -4407,34 +4596,44 @@ func (ls *LedgerState) ledgerProcessBlocksFromSource(
...
-				// HardForkFunc but skip per-era epoch-based events
-				// that should have fired during the omitted span.
-				if nextEpochEraId != snapshotEra.Id {
-					if !ls.isValidEraAdvancement(
+				// A boundary block can be encoded in the era immediately
+				// before the era announced by its header. Apply that pair of
+				// consecutive transitions in one transaction, but reject any
+				// larger jump because omitted per-epoch rules cannot be
+				// reconstructed safely from one block.
+				transitionPath, ok := ls.eraTransitionPath(
```
arepala-uml, line 4612: **`eraTransitionPath` accepts a raw two-era body jump with no header-elevation justification.**

`eraTransitionPath` (and `boundaryEraForBlock`, which feeds it) accept any era distance ≤ 2 between `snapshotEra.Id` and `nextEpochEraId`, based only on list-position distance. `boundaryEraForBlock`'s fallback branches — unknown header major, `eraForVersion` miss, `headerEraID == blockEraID`, non-adjacent header, or its own `eraTransitionPath` check failing — all return `target = blockEraID` unchanged, the raw body-encoded era, with no check that `blockEraID` is itself within one step of `snapshotEra.Id`.

Reproduction added locally on this commit (`3853555f`):

```go
func TestEraAdvancementRejectsRawTwoStepBodyJumpWithoutHeaderElevation(t *testing.T) {
	ls := &LedgerState{}
	target := ls.boundaryEraForBlock(
		eras.MaryEraDesc.Id,
		eras.BabbageEraDesc.Id,
		eras.BabbageEraDesc.MinMajorVersion,
		true,
	)
	require.Equal(t, eras.BabbageEraDesc.Id, target,
		"a self-consistent header/body must not be reclassified")

	_, ok := ls.eraTransitionPath(eras.MaryEraDesc.Id, target)
	require.False(t, ok,
		"a raw two-era body jump with no header-elevation justification "+
			"must be rejected, not silently advanced past the omitted era")
}
```

Result on this branch (`go test ./ledger -run TestEraAdvancementRejectsRawTwoStepBodyJumpWithoutHeaderElevation -v`):

```
=== RUN   TestEraAdvancementRejectsRawTwoStepBodyJumpWithoutHeaderElevation
    era_boundary_test.go:114:
        Error:      Should be false
        Test:       TestEraAdvancementRejectsRawTwoStepBodyJumpWithoutHeaderElevation
        Messages:   a raw two-era body jump with no header-elevation justification must be rejected, not silently advanced past the omitted era
--- FAIL: TestEraAdvancementRejectsRawTwoStepBodyJumpWithoutHeaderElevation (0.00s)
FAIL
```

Equivalent check against `main` (pre-PR), calling the old guard directly:

```go
ok := ls.isValidEraAdvancement(eras.MaryEraDesc.Id, eras.BabbageEraDesc.Id)
// asserted false
```

```
=== RUN   TestZZReview_MainRejectsRawTwoStepBodyJump
--- PASS: TestZZReview_MainRejectsRawTwoStepBodyJump (0.00s)
PASS
ok  	github.com/blinklabs-io/dingo/ledger	0.701s
```

So this is a regression introduced by this PR, not a pre-existing gap: `main` rejects the raw two-era jump outright; this branch accepts it.

The removed single-step guard's comment states a multi-step forward jump "implies a non-conforming block, a malformed snapshot, or a chain-selection bug." This PR's fix targets one specific pattern — body encoded in the predecessor era, header announcing the immediate successor — but the replacement guard accepts any 2-step distance regardless of whether the second step is justified by that pattern or is a raw, unvalidated body-era jump.

Suggested fix: have `boundaryEraForBlock` report whether its return value came from a validated header elevation (vs. a fallback that just echoes `blockEraID`), and require that flag before `eraTransitionPath`'s 2-step branch is allowed; fall back to the strict 1-step check otherwise.

Not reachable by any known real chain data — the documented case is the only one currently exercised — but it reintroduces the exact failure mode the removed guard's comment named.

arepala-uml (changes requested):
Validated: build, full `ledger/...` suite (`-race`, 336s), conformance, docs-parity, import-boundaries, `golangci-lint`, `nilaway` all pass.

Bot findings already resolved: Cubic's P1 (snapshot captured before the final-era switch) is fixed in this head commit, confirmed by `TestBoundaryEraTransitionsSnapshotRecordsFinalProtocolVersion`. CodeRabbit's backward-era-normalization finding was correctly withdrawn.

One inline finding below: the replacement era-advancement guard accepts a two-step jump unconditionally rather than only when justified by the documented header-elevation pattern, reproduced with a failing test on this branch that passes against the equivalent check on `main`.

ledger/state.go
wolf31o2: Fixed in 589dc761. boundaryEraForBlock now returns an explicit two-transition authorization only for a validated immediate-successor header elevation, and the rollover path rejects an uncorroborated raw two-era body jump. The supplied regression fails on the previous head and passes with the fix.

wolf31o2:
Also fixed the non-blocking golines finding in 589dc761. `make golines` passes.

ledger/state.go
arepala-uml: Verified independently — checked out 589dc761, confirmed the diff threads a validated boolean from `boundaryEraForBlock` through to `eraTransitionPath`'s 2-step gate. Ran the supplied regression test directly: fails on 3853555f (the previous head), passes on 589dc761. Full `ledger/...` suite, `golangci-lint`, `nilaway`, `golines`, conformance, and docs-parity all pass on the new commit. Resolving.

arepala-uml (approved):
The era-advancement finding is fixed in 589dc761 and independently verified: the supplied regression fails on the previous head and passes on this commit, full `ledger/...` suite passes, `golangci-lint`/`nilaway` clean, `golines` clean on the touched files, conformance and docs-parity pass. No other findings outstanding.

PR #3295, opened 2026-08-21
build: advance the Go toolchain floor past reachable advisories

### LINKED ISSUE #3274: build: update release toolchain past reachable Go advisories
## Problem

- `go.mod` pins `go1.26.1`.
- `Dockerfile` builds releases with `ghcr.io/blinklabs-io/go:1.26.3-1`.
- `govulncheck` reports 22 reachable standard-library advisories, fixed by Go 1.26.6 or earlier.
- Reachable paths include Mithril archive extraction, TLS, HTTP/2, and templates.

## Acceptance criteria

- Module, CI, and release builds use a fixed Go patch release.
- Release artifacts are rebuilt with the fixed toolchain.
- CI blocks reachable Go vulnerabilities.

wolf31o2:
Closes #3274.

## Summary

- Raise the root, governance-lens example, and Antithesis toolchain floor to Go 1.26.7 while retaining the Go 1.26.0 language minimum.
- Add a `govulncheck` job to the existing Go CI workflow through the repository Make target.
- Document why the current Blink Go builder image remains in place until its packaged Go release advances.

The new job reports reachable vulnerabilities in CI, but it is not currently a required branch-protection context.

## Compatibility

Existing `setup-go` inputs already track Go 1.26.x. The toolchain directive is a minimum, so newer installed toolchains remain usable. The remaining `openpgp` advisory has no upstream fix and is not reachable.

## Validation

At the current head, `govulncheck` reports no reachable vulnerability in the three modules. The release image builds Dingo with Go 1.26.7, and workflow, documentation-parity, and focused build checks pass.


Makefile
```diff
@@ -75,6 +75,12 @@ lint: import-boundaries ## Run import-boundaries, golangci-lint, nilaway, and mo
 	nilaway $(GO_TAG_FLAGS) ./...
 	modernize $(GO_TAG_FLAGS) $(MODERNIZE_PACKAGES)
 
+govulncheck: ## Report Go vulnerabilities this module actually calls
+	# Both the scanner and the toolchain float; a pin would park a new
+	# advisory behind a stale version. Exits non-zero only for reachable
+	# findings, which is what CI gates on.
+	go run golang.org/x/vuln/cmd/govulncheck@latest $(GO_TAG_FLAGS) ./...
```
cryptodj413: First unpinned `go run tool@latest` invocation wired into a *blocking* CI gate in this repo — confirmed by checking every other `go run tool@version` in this Makefile: `sqlc` is version-pinned (`SQLC_VERSION=v1.31.1`), `protoc-gen-go`/`protoc-gen-go-grpc` are pinned via `go.mod`+`go.sum` through `internal/tools/tools.go`. `nilaway`/`modernize` are unpinned too, but confirmed not gated by any CI workflow — dev-machine-only. The floating rationale here is explicit and reasoned (a pin would park a new advisory behind a stale scanner/version), and I agree with the tradeoff for a vulnerability scanner specifically — flagging only because it's a genuinely new risk shape for this repo (every PR run executes just-published, unpinned code with zero checksum, and the gate can start failing future unrelated PRs with no code change if the vuln DB or scanner updates). `golang/govulncheck-action` would get the same always-current-DB property through a SHA-pinnable Action, matching the existing `golangci-lint-action` convention in the sibling job. Non-blocking; worth a follow-up, not a blocker for this PR.

.github/workflows/go-test.yml
```diff
@@ -109,3 +109,31 @@ jobs:
...
+        # Blocking, not advisory. govulncheck exits non-zero only for
+        # vulnerabilities this module actually calls, so the unreachable
+        # findings it also prints (currently GO-2026-5932, an unmaintained
+        # golang.org/x/crypto package with no fixed version) do not fail a PR.
+        # Both the toolchain and the scanner float on purpose: a pin on either
+        # one parks a new advisory behind a stale version instead of forcing it
+        # to be fixed. The stdlib version reported against is whatever go.mod's
+        # toolchain floor and this runner resolve to. Run through the Makefile
+        # so the gate and `make govulncheck` cannot drift apart.
+        run: make govulncheck
```
cryptodj413, line 144: Confirmed via direct build+run: this job's action pins (`checkout@9c091bb2`, `setup-go@b7ad1dad`, `cache@55cc8345`) match the sibling `go-test-linux` job exactly, `actionlint` is clean on this file, and `repos/actions/repos-config.yaml` has no entry for `dingo` at all — this repo's workflows are entirely hand-maintained outside the shared governance engine, so there's no generated-wrapper drift risk from hand-editing this file. The only open question is whether this job is actually configured as a *required* status check for `main` (see the top-level comment) — the job itself is wired correctly regardless.

cryptodj413 (commented):
Reviewed independently. I ran `make govulncheck` myself rather than trusting the PR's stated numbers: forcing `GOTOOLCHAIN=go1.26.1` (the pre-PR floor) on this codebase reproduces **exactly 22** standard-library vulnerabilities; at the PR's head (`toolchain go1.26.7`) it's **0** called, 1 unreachable (`GO-2026-5932`, unmaintained `x/crypto/openpgp`, no fix available) — both match the PR body's claims precisely. Dispatched `blink-go-module-auditor` and `blink-release-auditor` for independent passes; between the three of us: `go1.26.7` confirmed as a real released version, all three `go.mod`s valid (`toolchain ≥ go` in each), no `go.sum` change needed (no new/removed deps, confirmed via `go list -m all -mod=readonly` clean in all three modules), `actionlint` clean on all 7 workflow files, and the `GOTOOLCHAIN=auto` mechanism the Dockerfile comment describes was reproduced end-to-end by actually building the `build` stage: the resulting binary reports `go1.26.7` via `go version -m`, and `GOTOOLCHAIN=local go version` in that same image reports `go1.26.3` — matching the PR's stated verification exactly rather than trusting it. `repos/docker-go/Dockerfile:2`'s `apk add go-1.26~=1.26.3` pin, cited as the reason for not bumping the base image tag, also checks out. `ARCHITECTURE.md`/`DATABASE.md`: correctly N/A, nothing here touches ledger or storage. No merge-blocking findings.

Cubic reports no issues; CodeRabbit was rate-limited (documented — a completed cubic review is sufficient per policy).

Non-blocking notes below (inline) plus two that don't anchor to a single line:

**Nested-module CI coverage gap (pre-existing, not introduced here).** The new `govulncheck` job and `go-test.yml`'s other jobs (lint, build, test) only ever run against the root module. `examples/dingo-gov-lens` and `internal/test/antithesis` each have their own `go.mod`, got the same `toolchain go1.26.7` bump and the PR body's own claimed counts (17→1, 1→0), but nothing in `.github/workflows/` re-checks either on every PR going forward (`antithesis.yml`'s triggers are commented out; `ci-docker.yml`/`publish.yml` don't touch their Dockerfiles). This predates this PR, but the PR body's per-module counts read as durable guarantees when only the root module's is CI-enforced. Worth a fast follow-up issue rather than blocking this one. Separately, and unrelated to this PR: `internal/test/antithesis/go.mod` is a clear dependency laggard (`gouroboros v0.163.4` / `plutigo v0.0.29` vs root's `v0.193.3` / `v0.3.0`) with no Dependabot entry for that path either — same root cause, different symptom, also worth its own tracked issue.

**Required-check wiring — indeterminate, not something I can resolve from here.** The PR body and Makefile comment call the new job "blocking." `main` reports `protected: true`, but the branch-protection/rulesets endpoints 404 or return empty for both of us — consistent with a permissions gap, not proof either way. Since GitHub never auto-promotes a new job to required, and this job has no prior history to check, I'd treat "blocking" as aspirational until someone with admin access confirms it's actually in the required-checks list.

Makefile
wolf31o2: Confirmed. I am keeping this shape in this PR: the shared Go repository policy deliberately floats govulncheck and the advisory data so a newly published advisory becomes an immediate gate instead of remaining hidden behind a stale tool version. The SHA-pinned Action is a reasonable supply-chain alternative to evaluate separately, but switching execution mechanisms is not needed for the toolchain/advisory fix here.

.github/workflows/go-test.yml
wolf31o2: Confirmed against the live main-branch protection settings. govulncheck is passing on this head, but it is not currently a required status context; the only required context is go-test (Linux) (1.26.x). The workflow wiring is complete, while making the new context required remains a repository-settings action outside this code change.

wolf31o2: Answered: it is not a required check. GitHub branch protection on main lists exactly one required status check, `go-test (Linux) (1.26.x)`, with one approving review required and strict mode off.

This PR adds govulncheck as its own job (`govulncheck`, at .github/workflows/go-test.yml:113) rather than a step inside `go-test-linux`, so the new job is not in the required set and cannot block a merge as configured. Making it gate merges would be a branch-protection change, not a workflow one.

The rest of the audit matches: the action pins are identical to the sibling job, actionlint is clean, and dingo has no entry in repos/actions/repos-config.yaml, so these workflows are hand-maintained with no generated-wrapper drift.

Makefile
wolf31o2: The floating invocation is intentional and stays. Organization policy is that Go toolchains and Go scanners float, and that a version is never pinned to make a check pass: a pin parks a new advisory behind a stale scanner, and the scan is wanted against the newest stdlib and newest advisory data. That is now written down in the toolkit rather than being folk knowledge.

The premise about the blast radius is narrower than stated, though. This is a separate `govulncheck` job, and branch protection on main requires only `go-test (Linux) (1.26.x)`, so the unpinned invocation is not inside a blocking gate: a scanner or database update can turn the job red on an unrelated PR, but it cannot block the merge as configured.

The inventory holds, and the supply-chain shape is real regardless of gating: sqlc is pinned via SQLC_VERSION, protoc-gen-go and protoc-gen-go-grpc through internal/tools/tools.go with go.mod and go.sum, and nilaway and modernize are unpinned but not CI-gated.

The `golang/govulncheck-action` suggestion is the right resolution and is filed as #3317: a SHA-pinned action keeps the always-current database, which is the property the floating invocation exists for, while matching the pinning convention golangci-lint-action already follows in the sibling job. `make govulncheck` stays the entry point so the gate and the target cannot drift.

wolf31o2: Follow-up is tracked in #3317: https://github.com/blinklabs-io/dingo/issues/3317.

.github/workflows/go-test.yml
```diff
@@ -109,3 +109,31 @@ jobs:
         # Keep PR feedback fast by omitting race instrumentation. Release
         # validation runs in the publish workflow before draft creation.
         run: go test -tags dingo_extra_plugins -ldflags="-s -w" -timeout 20m ./...
+
+  govulncheck:
+    name: govulncheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0 https://github.com/actions/checkout/releases/tag/v7.0.0
```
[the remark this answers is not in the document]
wolf31o2: Fixed in `fbe1dfa7`. This job runs `go run golang.org/x/vuln/cmd/govulncheck@latest`, so third-party code executes in the workspace; the flag is warranted here regardless of the class.

The condition is pre-existing and repository-wide, not introduced by this diff: 13 of dingo's 14 `actions/checkout` steps omit `persist-credentials: false`, including the sibling `go-test-linux` job in this same file. Only `antithesis.yml:267` sets it. The remaining 12 steps are tracked separately rather than fixed in a toolchain PR.

arepala-uml (dismissed):
Verified locally on the PR head (isolated worktree):

- `make govulncheck` on the root module: 0 reachable vulnerabilities
  (1 unreachable, GO-2026-5932, unmaintained x/crypto/openpgp — matches
  the PR description).
- `internal/test/antithesis`: 0 reachable, matching the claimed 1→0.
- `go test ./internal/docsparity/...`: passes: toolchain floor stays
  above the `go` directive minimum.

Approving but please resolve the merge conflict before merging the PR.

wolf31o2:
Merged `origin/main` (2031e4fd) into the branch at 1fe2f3d6; the prior approval was dismissed by the stale-review rule.

The only conflict was in the Makefile. #3303 added a second `govulncheck` target pinned to `GOVULNCHECK_VERSION=v1.7.0`, which auto-merged next to this branch's target and left two recipes for one name. The merge keeps a single target at main's position, retains main's `GOVULNCHECK` variable so `publish.yml` and the new `go-test.yml` job share one definition, and floats the version per the Go scanner policy. `make -n govulncheck` reports no overridden recipe; `@latest` resolves to v1.7.0 today. `go.mod`'s `toolchain go1.26.7` was identical on both sides.

Local validation on the merge commit: `go test -tags dingo_extra_plugins -ldflags="-s -w" -timeout 20m ./...`, `golangci-lint run ./...`, `make docs-parity`, `make import-boundaries`, `make gorm-check`, `make govulncheck`, `actionlint .github/workflows/go-test.yml`, and builds of the root module with and without `dingo_extra_plugins` plus both nested modules — all exit 0.

Makefile
wolf31o2: The current-main integration removed the stale Makefile sentence that described reachable findings as what CI gates on. The Makefile now describes only command behavior and scanner-version policy; it makes no branch-protection claim. Follow-up: 7f40e526.

.github/workflows/go-test.yml
wolf31o2: Updated in 7f40e526: the workflow now explicitly says govulncheck reports in CI but is not currently a required branch-protection context, and the inaccurate gate wording is gone. actionlint, docs parity, target resolution, and diff checks pass.

PR #3297, opened 2026-08-21
fix(ledger): halt on unrepairable trust-window validation failures

### LINKED ISSUE #3261: ledger: a deterministic validation failure inside the Mithril trust window has no recovery path and no terminal state
**Version:** `origin/main` 5a54b7f8. Network preview, Mithril v2 bootstrap.

### Summary

When a block inside the Mithril trust window fails transaction validation deterministically, at-tip recovery has no way out and no terminal state. It rewinds, gets rejected by the trust-boundary guard, rejects the peer, picks another peer, and repeats indefinitely. The stuck condition is reported at ERROR exactly once and then only at WARN, so a node that has stopped following the chain looks quiet in the log.

Observed for 21 hours on one node: 700 consecutive failed restarts, tip slot unchanged the whole time, one ERROR line total.

### Mechanism

`ledger/replay_recovery.go:750`, `rejectAtTipRecoveryAtMithrilBoundary`, refuses any rewind target at or behind `mithrilLedgerSlot`. That guard is correct. The problem is that when the failing block sits only a few thousand slots past the anchor, every rewind target the recovery schedule produces is inside the protected window, so the guard fires on all of them:

```
at-tip recovery exhausted scheduled rewind attempts, retrying with deepest rewind
  failing_slot=120574859 attempts=4
at-tip validation recovery would cross Mithril trust boundary, rejecting peer chain
  failing_block_slot=120574859 ledger_tip_slot=120574813
  mithril_ledger_slot=120566425
```

749 and 747 occurrences respectively in one run. The failing block was 8434 slots past the anchor.

`temporarily denying chainsync peer whose chain we cannot follow` then cycles the peer set (705 occurrences), which cannot help: the chain is canonical, so the next peer offers the same block.

### No terminal state

`ledger/state.go:4188`, in `ledgerProcessBlocks`, announces the stuck transition once:

```go
if stuck && progress.consecutiveNoProgress == noProgressStuckThreshold {
    ls.config.Logger.Error("ledger pipeline stuck: ...")
}
```

and afterwards only:

```go
if progress.consecutiveNoProgress == 10 ||
   progress.consecutiveNoProgress%100 == 0 {
    ls.config.Logger.Warn("ledger pipeline making no progress across repeated restarts, backing off")
}
```

The loop then retries forever. In the run this produced one ERROR at `consecutive_no_progress=50` and nothing further at ERROR for the following 18 hours, while the node kept serving and kept its metrics endpoint up. Everything after that point was WARN, mixed into 129k other WARN lines.

### Relationship to open work

#3251 adds exactly the machinery this needs, but scopes it to duplicate-input errors: `isDeterministicTxValidationError` classifies `shelley.DuplicateInputError` and `eras.DuplicateInputByronError` only, and its own description notes that other validation errors continue through producer resolution and the unresolved-producer fallback. The failure that produced the 21-hour wedge was `DelegateUnregisteredStakeCredentialError` (Conway rule 45), which is deterministic in the same sense and still takes the old path.

The specific state bug behind that rule 45 rejection is #3260. This issue is the surrounding gap: with #3260 fixed, any other deterministic validation failure inside the trust window still wedges the same way.

### Asks

- Generalize the deterministic classification so a validation error that cannot be repaired by replaying a different local history takes the halt-or-resync path rather than the unresolved-producer path. Rule 45 is one concrete member; the classifier should be a decision about repairability, not a fixed error list.
- Give the stuck state an escape or an end. Either re-derive the state the failing block needs without crossing the anchor, or stop retrying and surface an operator-actionable terminal condition.
- Keep the stuck condition visible after the first announcement. A once-only ERROR followed by WARN every 100 attempts means log-level alerting sees a node that has stopped following the chain as healthy.

### LINKED ISSUE #3301: ledger: Mithril-bootstrapped node stalls silently when recovery needs a rollback below the trust anchor
**Versions:** first observed on v0.69.0
(`496e923edc08b59e496378181af7739f6d96aa18`), preview, relay role. The refusal
code is unchanged on current main (`ada53695`).

**What is and is not established, stated up front:** the mechanism is verified
present in current main and the boundary rejection was observed again on a
recent main-based build. The full deadlock has **not** been reproduced on
`ada53695` itself. A node freshly Mithril-bootstrapped on that commit is at tip
and healthy, with zero boundary rejections. So this is reported as a latent
defect whose code path is intact, not as a live reproduction on the newest
commit.

A Mithril-bootstrapped relay stopped applying blocks permanently. It is not on a
fork, it is not behind due to networking, and **it cannot recover by restart —
the condition is structural.**

### Summary

After a Mithril bootstrap the node holds *ledger state* at the anchor but no
*block history* below it. When a later block spends a UTxO that was created
below the anchor, the ledger's producer lookup cannot find the producing
transaction — correctly, since that block was never applied. Replay recovery
responds by demanding a rollback to the producing block, which lies below the
anchor, and `recoveryRollbackExceedsMithrilBoundary` refuses it. Neither side
can yield, so the node retries forever.

### Follow-up 2026-08-21 (later the same day)

Two points worth adding before anyone acts on this.

First, the deadlock has an escape in practice: discarding the database and
taking a fresh Mithril bootstrap clears it. That is a heavy remedy -- it is a
full resync -- but it means the condition is recoverable by an operator, not
terminal. What it is not is *self*-recoverable: the node will retry forever
without progressing and without surfacing a terminal state.

Second, a caution about reproducing it. An earlier bootstrap attempt the same
day failed with `importing ledger state: verified ancillary data ... has no
usable ledger state` and retried in a loop. A later snapshot for the same
network imported cleanly with no code change, so that particular failure was
specific to one published snapshot rather than a property of the node. Do not
conflate the two symptoms.

### Verified still present on current main (2026-08-21)

Re-checked against `5a54b7f8`. The refusal is unchanged:

```go
// ledger/replay_recovery.go:492
func (ls *LedgerState) recoveryRollbackExceedsMithrilBoundary(point ocommon.Point) bool {
    ls.RLock(); defer ls.RUnlock()
    return ls.mithrilLedgerSlot > 0 && point.Slot < ls.mithrilLedgerSlot
}
```

Independently reproduced the same day on two freshly Mithril-bootstrapped preview
nodes on this build, which log
`detected replay recovery below Mithril trust boundary, rejecting peer chain`
and never resume. So this is not specific to v0.69.0.

To be clear about what we are and are not claiming: refusing to roll back below
the trust anchor is correct on its own. The defect is that the two halves
deadlock — replay recovery keeps demanding a rollback the boundary check must
keep refusing, with no escape and no operator-visible terminal state.

**#3251** (open, "fix(ledger): recover from deterministic replay tx failures")
touches this recovery path and may already address it; worth checking before
this is filed separately.

### Verified facts

Chain data verified against an independent chain index.

| | |
|---|---|
| Mithril anchor | slot `119819164` |
| node tip | slot `119844856`, block `4563173`, hash `7413782a06bd…83c4` |
| tip is canonical? | **yes** — hash matches the canonical block at that slot exactly |
| failing block | slot `119844907` (51 slots past the node's tip) |
| spent input | `07fa6efa…d669d7#3` |
| producing block | slot `119727794`, block `4559079` — **91,370 slots below the anchor** |
| producing block canonical? | **yes** |
| UTxO status | created in `4559079`, unspent at the anchor, spent by the failing block |

So every participant is canonical and well-formed. The input was live in the
Mithril snapshot and was legitimately spent 51 slots after the node's tip.

### The deadlock

```go
// ledger/replay_recovery.go
func (ls *LedgerState) recoveryRollbackExceedsMithrilBoundary(point ocommon.Point) bool {
    return ls.mithrilLedgerSlot > 0 && point.Slot < ls.mithrilLedgerSlot
}
```

```
level=WARN msg="detected replay recovery below Mithril trust boundary, rejecting peer chain"
  recovery_strategy=chain-scan
  failing_block_slot=119844907
  missing_input=07fa6efa115dbe7e3ac62e61369036abbff7b6b8cfba2a434ed7b6c202d669d7#3
  producer_block_slot=119727794
  rollback_slot=119727784
  mithril_ledger_slot=119819164
  rewind_target_slot=119844856
```

All 9 rejections captured carry the **same** `(failing_block, producer)` pair,
and in every one the producer is below the anchor. The node is not cycling
through candidates; it is stuck on one block it can never accept.

The refusal itself is right — that is what the trust boundary is for. The
problem is that the demand is unsatisfiable by construction: no amount of
retrying will produce block history the node never had, so recovery and boundary
enforcement deadlock.

### Consequence: silent and unrecoverable

| | 45 min before restart | after a clean restart |
|---|---|---|
| `continuation block spends an input with no producer…` | 270 | 975 |
| `selected new best peer` | 84 | 367 |
| `chain switch: updating active connection` | 87 | 367 |
| `persistent chain fork detected, triggering chainsync re-sync` | — | 101 |
| **blocks applied** | **0** | **0** |

The node reports itself healthy throughout: container up, 29 outbound peers,
disk 42%, no crash, tip hash canonical. It drifted 11,223 blocks (~3 epochs)
behind and the gap kept growing. Only comparing `blockNum` against another node
reveals it. The non-convergence guard from #3005/#3008 is present in this build
and does stop the recovery descent; it does not stop the chainsync retry above
it.

### Scope in our fleet

Six Dingo nodes, four block producers among them:

| node | role | Mithril anchor | boundary rejections | state |
|---|---|---|---|---|
| relay A | relay | `119819164` | **38** | wedged |
| producer B | **block producer** | `120081871` | 0 | healthy |
| 4 × musashi | bp + relay | none (from genesis) | 0 | healthy |

Only the Mithril-bootstrapped node with the older anchor is affected. No block
producer has hit it — but the code path is in `ledger/`, not the forge path, and
producer B is itself Mithril-bootstrapped, so a producer appears equally exposed
and simply has not yet met a block spending a pre-anchor input. Nodes synced
from genesis have `mithrilLedgerSlot == 0` and cannot reach the condition at all.

### Relationship to open PR #3251

**#3251 does not cover this case.** Its new
`recoverFromDeterministicTxValidationError` reaches the same boundary check and
gives up silently:

```go
if ls.recoveryRollbackExceedsMithrilBoundary(rewindPoint) {
    return false, nil
}
```

`false, nil` means "did not recover, nothing went wrong" -- no log line, no
error, no alternative path. A Mithril-bootstrapped node in the state described
above returns from that call having done nothing, with no operator-visible
signal. The PR is also explicit that it is scoped to deterministic structural
failures and that state-dependent ones "still need producer resolution and the
security-parameter fallback." The condition here is the state-dependent kind, so
it is deliberately out of that PR's scope.

**What #3251 does supply is the shape of the fix.** Two ideas in it are exactly
what the path described here lacks:

1. **A terminal state.** It records `deterministicTxRecoveryAttempted` and the
   tip slot it was attempted at, and on a repeat failure at the same tip it
   stops the pipeline with an explicit operator hint rather than retrying:
   *"network peers are serving a transaction this node rejects; operator
   intervention is required."* The path in this report has no equivalent -- it
   retries indefinitely and never reports a terminal condition.

2. **A windowed retreat.** `rollbackPrimaryChainInSecurityParamWindows` rewinds
   in security-parameter steps instead of demanding a single deep rollback,
   which is a plausible shape for a boundary-aware fallback.

**The ask, concretely:** could the same terminal-state treatment be applied when
`recoveryRollbackExceedsMithrilBoundary` refuses a rewind? Refusing the rollback
is correct and we are not asking for that to change. What is missing is that the
refusal is silent and non-terminal: the node cannot proceed, cannot roll back,
and does not say so. Even surfacing it as a halt with a hint -- "the chain
requires a rollback below the Mithril anchor; re-bootstrap required" -- would
turn an indefinite silent stall into an actionable one.

### Suggested direction

Two parts, roughly independent:

1. **Don't treat a pre-anchor producer as a fault.** For a Mithril-bootstrapped
   node, "producing transaction is below the anchor" is expected and benign —
   the anchor's ledger state already accounts for it. Recovery arguably should
   accept the input as valid on the strength of the snapshot rather than
   demanding provenance the node structurally cannot have.
2. **Make the unsatisfiable case terminal.** If a recovery path is refused at the
   boundary and no alternative exists, latch an explicit unrecoverable state with
   a distinct log line and a metric, rather than retrying indefinitely while
   appearing healthy. Operators currently have no signal short of cross-node tip
   comparison.

### Not established

- **Reproduction from a fresh bootstrap.** We have one instance. The wedged
  database has been preserved and we can re-bootstrap and report whether it
  recurs; the analysis predicts it will whenever a post-anchor block spends a
  pre-anchor input, which is an ordinary occurrence.
- Why this node's anchor was old enough to expose the window while the other
  Mithril-bootstrapped node's was not.

Happy to supply the full capture, or to test a patch against the preserved
database before we re-bootstrap.

wolf31o2:
Closes #3261.
Closes #3301.

## Problem

Validation recovery at a Mithril trust boundary could rotate peers indefinitely when every legal rewind target was below the imported anchor. Replay could also rearm the recovery budget by reporting a different failing block or transaction without advancing the applied ledger tip.

## Changes

- Bound trust-boundary recovery by applied-tip progress rather than failure identity.
- Count only attempts whose local rollback completed successfully.
- Halt the ledger pipeline with persistent metrics and operator guidance once the legal recovery space is exhausted.
- Keep stuck pipelines visible through periodic ERROR announcements.
- Cover at-tip and replay recovery, rollback-mechanics failures, sticky terminal state, and pipeline shutdown under race-enabled tests.

`ARCHITECTURE.md` documents the terminal state and recovery signal. `DATABASE.md` was checked and is unaffected. There is no API, wire-format, or dependency change.

## Composition

Trial composition with #3251 conflicts only at two additive `ledger/state.go` sites: both recovery trackers and both applied-tip reset calls must be retained.

---

## Summary by Cubic

Halts the ledger pipeline when Mithril trust-window validation failures are unrepairable, preventing infinite peer rotation on canonical blocks and improving stuck-state visibility. Boundary refusals are tallied per applied tip; exhausting the legal recovery space logs the operator action, increments `dingo_ledger_mithril_trust_window_unrepairable_total`, returns terminal `errHaltLedgerPipeline`, and sets `dingo_ledger_pipeline_halted`.

- The tally clears only after the applied tip advances, records only after a successful local rewind, and remains terminal once exhausted.
- ERROR announcements repeat for general and certified-endorser stuck paths so wedges stay visible.
- Nodes halt the ledger pipeline but keep serving stale queries; recovery requires a newer Mithril snapshot or a genesis resync.
- Tests cover terminal halts, recoverable retries, stuck-announcement cadence, applied-tip-keyed budgets, and rollback failures that must not consume the budget.

No wire-format or era-semantics changes.

### DIFF
--- a/ledger/pipeline_halt_test.go
+++ b/ledger/pipeline_halt_test.go
@@ -68,7 +68,7 @@ func TestLedgerProcessBlocksStopsRetryingOnUnrepairableFailure(t *testing.T) {
 	testutil.RequireReceive(
 		t,
 		done,
-		10*time.Second,
+		5*time.Second,
 		"an unrepairable validation failure must stop the ledger pipeline",
 	)
 
@@ -109,7 +109,7 @@ func TestLedgerProcessBlocksKeepsRetryingRecoverableFailures(t *testing.T) {
 	testutil.WaitForCondition(
 		t,
 		func() bool { return attempts.Load() >= 3 },
-		10*time.Second,
+		5*time.Second,
 		"a recoverable failure must keep restarting the pipeline",
 	)
 	assert.Zero(
@@ -122,7 +122,7 @@ func TestLedgerProcessBlocksKeepsRetryingRecoverableFailures(t *testing.T) {
 	testutil.RequireReceive(
 		t,
 		done,
-		10*time.Second,
+		5*time.Second,
 		"the pipeline loop must exit when its context is cancelled",
 	)
 }
--- a/ledger/pipeline_halt_test.go
+++ b/ledger/pipeline_halt_test.go
@@ -1,0 +1,207 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ledger
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/blinklabs-io/dingo/internal/test/testutil"
+	"github.com/prometheus/client_golang/prometheus"
+	promtestutil "github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func newPipelineLoopLedger(t *testing.T) *LedgerState {
+	t.Helper()
+	ls := &LedgerState{
+		config: LedgerStateConfig{
+			Logger: slog.New(slog.NewJSONHandler(io.Discard, nil)),
+		},
+	}
+	ls.metrics.init(prometheus.NewRegistry())
+	return ls
+}
+
+// TestLedgerProcessBlocksStopsRetryingOnUnrepairableFailure covers the terminal
+// half of issue #3261. Recovery raises errHaltLedgerPipeline once it has
+// established that no local replay can change a block's verdict; the restart
+// loop must then stop rather than restart into the same block forever, and must
+// leave a terminal signal behind for an operator.
+func TestLedgerProcessBlocksStopsRetryingOnUnrepairableFailure(t *testing.T) {
+	ls := newPipelineLoopLedger(t)
+
+	var attempts atomic.Int64
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		ls.ledgerProcessBlocksWithAttempt(
+			t.Context(),
+			func(context.Context) error {
+				attempts.Add(1)
+				return fmt.Errorf(
+					"process block batch: %w",
+					errHaltLedgerPipeline,
+				)
+			},
+		)
+	}()
+	testutil.RequireReceive(
+		t,
+		done,
+		10*time.Second,
+		"an unrepairable validation failure must stop the ledger pipeline",
+	)
+
+	assert.Equal(
+		t,
+		int64(1),
+		attempts.Load(),
+		"a halted pipeline must not run another attempt",
+	)
+	assert.Equal(
+		t,
+		1.0,
+		promtestutil.ToFloat64(ls.metrics.pipelineHalted),
+		"a halted pipeline must report its terminal state",
+	)
+}
+
+// TestLedgerProcessBlocksKeepsRetryingRecoverableFailures is the negative case:
+// an ordinary failure must keep restarting the pipeline. Treating every failure
+// as terminal would turn a transient database or peer problem into an outage.
+func TestLedgerProcessBlocksKeepsRetryingRecoverableFailures(t *testing.T) {
+	ls := newPipelineLoopLedger(t)
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+	var attempts atomic.Int64
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		ls.ledgerProcessBlocksWithAttempt(
+			ctx,
+			func(context.Context) error {
+				attempts.Add(1)
+				return errors.New("transient read failure")
+			},
+		)
+	}()
+	testutil.WaitForCondition(
+		t,
+		func() bool { return attempts.Load() >= 3 },
+		10*time.Second,
+		"a recoverable failure must keep restarting the pipeline",
+	)
+	assert.Zero(
+		t,
+		promtestutil.ToFloat64(ls.metrics.pipelineHalted),
+		"a retrying pipeline must not report itself halted",
+	)
+
+	cancel()
+	testutil.RequireReceive(
+		t,
+		done,
+		10*time.Second,
+		"the pipeline loop must exit when its context is cancelled",
+	)
+}
+
+// TestPipelineStuckAnnouncementStaysVisible covers the third ask of issue
+// #3261. The stuck condition was announced at ERROR exactly once and then only
+// at WARN, so a node that had stopped following the chain looked quiet to
+// log-level alerting for as long as it stayed wedged.
+func TestPipelineStuckAnnouncementStaysVisible(t *testing.T) {
+	for consecutive := range noProgressStuckThreshold {
+		assert.False(
+			t,
+			pipelineStuckShouldAnnounce(consecutive),
+			"restart %d is not stuck yet and must not announce",
+			consecutive,
+		)
+	}
+	assert.True(
+		t,
+		pipelineStuckShouldAnnounce(noProgressStuckThreshold),
+		"the transition into stuck must be announced",
+	)
+
+	announcements := 0
+	for consecutive := noProgressStuckThreshold; consecutive <= noProgressStuckThreshold+
+		4*noProgressStuckReannounceInterval; consecutive++ {
+		if pipelineStuckShouldAnnounce(consecutive) {
+			announcements++
+		}
+	}
+	assert.Equal(
+		t,
+		5,
+		announcements,
+		"a persistently stuck pipeline must keep announcing itself at a fixed cadence",
+	)
+}
+
+// TestResetMithrilBoundaryRejectionsRequiresAppliedTipProgress verifies that
+// the trust-window tally is keyed on applied ledger progress rather than on a
+// reported failing-block identity. Replay can report changing failures while
+// rebuilding to the same applied tip; those are still one non-converging run.
+func TestResetMithrilBoundaryRejectionsRequiresAppliedTipProgress(
+	t *testing.T,
+) {
+	ls := &LedgerState{}
+
+	rejections, exhausted := ls.observeMithrilBoundaryRejection(500)
+	require.Equal(t, 1, rejections)
+	require.False(t, exhausted)
+
+	// Replaying to the same or an older applied tip is not progress.
+	ls.resetMithrilBoundaryRejections(500)
+	rejections, exhausted = ls.observeMithrilBoundaryRejection(499)
+	require.Equal(t, 2, rejections)
+	require.False(t, exhausted)
+
+	// Advancing the applied high-water mark is real progress and starts a
+	// later recovery run with a fresh budget.
+	ls.resetMithrilBoundaryRejections(501)
+	rejections, exhausted = ls.observeMithrilBoundaryRejection(501)
+	assert.Equal(t, 1, rejections)
+	assert.False(t, exhausted)
+
+	// Every scheduled rewind depth refused, plus the capped retry the
+	// schedule settles on: the legal rewind space is exhausted.
+	for rejections < maxMithrilBoundaryRecoveryRejections {
+		rejections, exhausted = ls.observeMithrilBoundaryRejection(501)
+		require.False(
+			t,
+			exhausted,
+			"%d refusals is still inside the bound",
+			rejections,
+		)
+	}
+	_, exhausted = ls.observeMithrilBoundaryRejection(501)
+	require.True(
+		t,
+		exhausted,
+		"the tally must be exhausted once every legal rewind depth has been refused",
+	)
+}
--- a/ledger/replay_recovery.go
+++ b/ledger/replay_recovery.go
@@ -98,6 +98,25 @@ const maxAtTipRecoveryDescents = 2
 // ChainSync connection (issue #3005).
 const maxReplayRecoveryNoProgress = 2
 
+// maxMithrilBoundaryRecoveryRejections bounds how many successful validation
+// recovery attempts may refuse a rewind at the Mithril trust boundary without
+// advancing the applied ledger tip before the failure is declared
+// unrepairable (issues #3261 and #3301).
+//
+// The bound covers the whole legal rewind space. A boundary rejection rewinds
+// to the applied ledger tip -- the deepest point recovery may reach without
+// crossing the anchor -- and asks ChainSync for a fresh intersection, so one
+// rejection per scheduled rewind depth exhausts every target the schedule can
+// produce, plus one for the capped retry the schedule settles on. Past that,
+// every legal target has been replayed and returned the same verdict.
+const maxMithrilBoundaryRecoveryRejections = maxAtTipRecoveryAttempts + 1
+
+type mithrilBoundaryRecoveryProgress struct {
+	highWaterTipSlot uint64
+	rejections       int
+	halted           bool
+}
+
 type atTipRecoveryAttempt struct {
 	BlockPoint ocommon.Point
 	TxHash     []byte
@@ -504,6 +523,7 @@ func (ls *LedgerState) rejectReplayRecoveryAtMithrilBoundary(
 ) error {
 	return ls.rejectRecoveryAtMithrilBoundary(
 		"replay recovery rollback exceeds Mithril trust boundary",
+		validationErr,
 		func(mithrilLedgerSlot uint64, rewindPoint ocommon.Point) {
 			ls.config.Logger.Warn(
 				"detected replay recovery below Mithril trust boundary, rejecting peer chain",
@@ -745,6 +765,7 @@ func (ls *LedgerState) rejectAtTipRecoveryAtMithrilBoundary(
 ) error {
 	return ls.rejectRecoveryAtMithrilBoundary(
 		"at-tip recovery rollback exceeds Mithril trust boundary",
+		validationErr,
 		func(mithrilLedgerSlot uint64, rewindPoint ocommon.Point) {
 			ls.config.Logger.Warn(
 				"at-tip validation recovery would cross Mithril trust boundary, rejecting peer chain",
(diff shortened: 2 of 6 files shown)


ledger/replay_recovery_trust_window_test.go
```diff
@@ -0,0 +1,268 @@
...
+	// block keep halting and must not ask for yet another fresh intersection.
+	fixture.drainResyncs()
+	for range 3 {
+		recovered, err := fixture.ls.tryRecoverFromTxValidationError(
+			fixture.failing,
+		)
+		assert.False(t, recovered)
+		require.ErrorIs(t, err, errHaltLedgerPipeline)
+	}
+	assert.Zero(
```
cryptodj413, line 249: Confirmed independently reproducible — 17/60 failures on a plain run (`go test -count=60`), consistent with the subagent's 8/50 (and 1/30 under `-race`). Root cause: `EventBus.Publish` only enqueues; `SubscribeFunc`'s handler runs on a separate goroutine reading a buffered channel, so `fixture.drainResyncs()` at line 231 races against the 5 resyncs published during the pre-halt recoveries above it — a straggler can land in `resyncs` after the drain and get counted by the `assert.Zero` at line 239 ("a halted recovery must not keep rotating peers"), failing on a false positive.

Same failure signature both times I reproduced it, so this isn't sandbox-specific noise — it'll flake real CI runs at roughly this rate too; the one CI run for this PR just didn't hit it.

Fix: consume the expected pre-halt resyncs with `testutil.RequireReceive` (already the pattern at `ledger/replay_recovery_test.go:381-386`) before asserting none follow, rather than a best-effort `drainResyncs()`. I'd want this fixed before merge — the underlying halt logic is correct (verified separately), this is purely a test-synchronization bug.

ledger/replay_recovery.go
```diff
@@ -796,6 +863,34 @@ func (ls *LedgerState) rejectRecoveryAtMithrilBoundary(
 			ErrRollbackExceedsMithrilBoundary,
 		)
 	}
+	rejections, exhausted := ls.observeMithrilBoundaryRejection(validationErr)
```
cryptodj413: The tally is incremented here, before the rollback attempt at line 895 below. If that rollback itself fails (e.g. `ErrRollbackPointNotOnChain` from chain selection racing concurrently — a real, if rare, possibility per the TOCTOU discussion on #3251's review), the function returns a plain wrapped error rather than `errHaltLedgerPipeline`, so it doesn't halt by itself — but it has already consumed one unit of the shared rejection budget for a rollback-mechanics failure, not a repeated validation rejection on the same block. Five such rollback races in a row (or four races plus one genuine repeat) would exhaust the tally and halt on the *next* validation encounter, with the ERROR line and `dingo_ledger_mithril_trust_window_unrepairable_total` increment both attributing it to "cannot be repaired by replaying local history" when the actual cause was infrastructure contention. Narrow precondition, non-blocking, but worth tightening — e.g. don't call `observeMithrilBoundaryRejection` until after `rollbackPrimaryChainInSecurityParamWindows` succeeds, or don't count a rollback-mechanics error toward this tally at all.

cryptodj413: `observeMithrilBoundaryRejection` resets the tally to a fresh `Attempts: 1` whenever `!ls.mithrilBoundaryRejection.matches(validationErr)` — i.e. on ANY variance in the reported `(BlockPoint, TxHash)`, not just on genuine forward progress. I confirmed with the protocol-auditor's alternating-identity probe that this defeats the bound completely (40 cycles, never exceeds `Attempts: 1`, never halts) if the identity varies. For the at-tip entry point this is provably unreachable (rejection only fires once `attempts > 1`, which requires `isSameFailure` to already hold), but the replay-recovery entry point (`rejectReplayRecoveryAtMithrilBoundary`) has no such precondition and can hit this on the very first failure — and per the sibling bound's own doc a few lines up (`maxReplayRecoveryNoProgress`'s comment: "the failing block may change and creep forward... so the ledger tip — not the failure identity — is the convergence signal"), that's exactly the shape this file already designed around once, for the sibling mechanism. Non-blocking, since a single wedged canonical block's first failing tx is deterministic within a block-apply loop (verified: `state.go`'s tx-validation loop is sequential, one `txValidationError` construction site) — but worth either an additional tip-keyed bound alongside the identity-keyed one, or a comment explaining why identity variance re-arming the budget is acceptable here specifically.

Related test-coverage gap: both new test files exercise only the at-tip path (`newTrustWindowLedger` sets `reachedTip.Store(true)`) — the replay-recovery path, which is structurally the one a *fresh* post-Mithril-bootstrap wedge would actually take (`reachedTip` is a one-way latch set only after a successful batch commit), has no dedicated terminal-state test. Non-blocking, but `ledger/replay_recovery_test.go:225`'s existing fixture looks like a ready base for one.

cryptodj413 (commented):
Reviewed independently, including dispatching `blink-protocol-auditor` for a second pass. This PR also fixes exactly the gap I flagged as a likely merge blocker on #3251 (the pipeline-restart loop clearing `pipelineNoProgress`/`pipelineStuck` to a healthy-looking 0 on halt) — confirmed correct here with an ad hoc probe: drove 3 transient failures then a halt, and `pipelineNoProgressRestarts` retained its pre-halt value instead of being zeroed. Also independently verified: `maxMithrilBoundaryRecoveryRejections = 4` matches its doc comment's stated reasoning against `maxAtTipRecoveryAttempts = 3`; no metric-name collision for the two new metrics; `ARCHITECTURE.md` is the only doc touched, `DATABASE.md` correctly unaffected; lock discipline on the new `mithrilBoundaryRejection` field is consistent with its siblings (single-goroutine-owned, no lock needed); and a trial merge against the still-open, unmerged #3251 (which also touches `ledgerProcessBlocks`/`errHaltLedgerPipeline`) resolves cleanly in exactly the 3 hunks the PR body claims, with both halt triggers composing correctly (disjoint dispatch, both get the gauge-suppression fix).

One item below is a confirmed, reproducible flaky test — I'd want it fixed before merge even though the production code itself is correct. The rest are non-blocking.

**DevNet claim — real run, but doesn't reach the new code.** The PR body's `run-tests.sh --accelerated` numbers are a real scenario (`accelerated_timeline_test.go`, readiness/propagation/agreement/epoch-transition/peer-interruption), and it's legitimate evidence of no regression in the pipeline-restart loop this PR refactors. But devnet never bootstraps from a Mithril snapshot (`mithrilLedgerSlot` stays 0 throughout), so `recoveryRollbackExceedsMithrilBoundary` returns `false` unconditionally there — the accelerated run cannot exercise the new tally/halt path at all. I'd also push back gently on "`--conformance` doesn't apply" for the same reason it's given here (no wire-format/era-semantics change): the dingo-vs-cardano-node topology is this repo's standard regression check for the shared block-application loop, and it's worth naming as "doesn't reach the new path either" rather than "doesn't apply."

**Composition note for whoever merges second with #3251:** post-merge, `recoverFromDeterministicTxValidationError`'s own Mithril-boundary check (`ls.recoveryRollbackExceedsMithrilBoundary(rewindPoint) { return false, nil }`) bypasses both terminal mechanisms for a duplicate-input error caught exactly at the boundary — narrow (needs the applied tip itself below the anchor) but worth a follow-up on whichever PR lands second.

ledger/replay_recovery_trust_window_test.go
wolf31o2: Fixed in 89a9560e. The diagnosis matches the code: drainResyncs is a non-blocking drain and EventBus.Publish only enqueues, so it could return before the last pre-halt publish had landed and the straggler was then counted by the assertion expecting silence.

requireResyncs consumes exactly the number of resyncs the recovery loop produced, using testutil.RequireReceive with a 2s timeout, so the halt phase cannot observe one from the phase before it. The count is asserted implicitly: a wrong expectation fails on timeout rather than passing quietly.

One thing worth stating plainly: the flake does not reproduce on this machine. The pre-fix code passed 30/30 on a plain run and 30/30 under -race here, against 17/60 and 1/30 in the report. So the 60/60 the fixed version passes is not evidence the fix works, only that it does not regress. What supports the fix is the mechanism — a blocking wait for a known count cannot leave a straggler, where a non-blocking drain can — plus the reproduction in this thread.

ledger/replay_recovery.go
wolf31o2: Fixed in fdfdd8be, taking the second of the two suggested shapes: the rejection is returned rather than the observe call being deferred.

Deferring observe until after the rewind succeeded would move the halt decision after the rewind, so the halting iteration would perform one more rewind before stopping. Returning the budget on the failure path keeps the existing order and confines the change to the case in question.

unobserveMithrilBoundaryRejection decrements the tally and clears it when it reaches zero, and rejectRecoveryAtMithrilBoundary calls it on the rollbackPrimaryChainInSecurityParamWindows error path. A rewind that never reached the validation question no longer spends budget, so a later genuine rejection is not misattributed to unrepairable history.

TestUnobserveMithrilBoundaryRejectionReturnsBudget drives far more observe/return pairs than the budget allows, then confirms repeated rejections of the same block still exhaust it — so the budget is returned, not disabled. It fails when the decrement is removed, checked by removing it.

Full ledger suite green, golangci-lint 0 issues.

wolf31o2: Confirmed and filed as #3318 rather than changed here.

The reading is right on both halves. observeMithrilBoundaryRejection re-arms on any variance in (BlockPoint, TxHash) because matches gates the increment, so a varying identity keeps Attempts at 1 and the halt never fires. The at-tip entry point is unaffected for the reason given, and rejectReplayRecoveryAtMithrilBoundary has no equivalent precondition.

Left out of this PR deliberately: the fix is a new bound in a halt path, and the precedent this thread cites is the right one to follow rather than approximate. #3318 records the direction — a tip-keyed bound alongside the identity-keyed one, matching maxReplayRecoveryNoProgress treating tip progress rather than failure identity as the convergence signal, which resetMithrilBoundaryRejections already does for clearing. Acceptance criteria require the alternating-identity sequence to reach the terminal state and to fail without the new bound.

The coverage gap is in the same issue: both new test files set reachedTip.Store(true), so rejectReplayRecoveryAtMithrilBoundary has no terminal-state test, with ledger/replay_recovery_test.go:225 named as the base.

wolf31o2: Addressed in 8193f04d. `observeMithrilBoundaryRejection` now runs only after `rollbackPrimaryChainInSecurityParamWindows` succeeds, so rollback-mechanics errors leave the rejection tally and terminal metric unchanged. `TestMithrilBoundaryRollbackFailureDoesNotConsumeRecoveryBudget` repeats those failures beyond the terminal threshold; moving the observation before rollback makes the test fail.

wolf31o2: Addressed in 8193f04d. The boundary tracker now uses applied-tip high-water progress, so changing block or transaction identities at a fixed tip cannot rearm the budget; only advancing the applied tip resets it. `TestReplayRecoveryBelowMithrilTrustBoundaryReachesTerminalState` seeds a canonical producer below the anchor, varies both identities, and verifies the bounded terminal error and operator-visible metric. The race-enabled focused suite passes.

wolf31o2:
@coderabbitai review

ledger/pipeline_halt_test.go
```diff
@@ -0,0 +1,207 @@
...
+				)
+			},
+		)
+	}()
+	testutil.RequireReceive(
+		t,
+		done,
+		10*time.Second,
+		"an unrepairable validation failure must stop the ledger pipeline",
+	)
```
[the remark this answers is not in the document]
wolf31o2: Addressed in e62ea8bd. Both `RequireReceive` calls and the adjacent `WaitForCondition` now use the repository-guided five-second upper bound. The three race-enabled pipeline-halt tests pass in 1.118s, and ledger-scoped golangci-lint reports zero issues.

verbotenj (approved):

ledger/replay_recovery_trust_window_test.go
wolf31o2: Correction: #3297 merged at e62ea8bd before this test fix landed, so the earlier follow-up commit was not part of the merged PR. #3329 carries the focused test-only fix from current main at df204a37. Its controlled fail-before reproduces the zero-count drain, and the fixed test passes 60 consecutive race-enabled runs.

Issue #2041, opened 2026-04-24
security: normalize Mithril snapshot metadata before cache path checks


wolf31o2:
## Finding

ID: DSA-2026-04-24-05
Severity: LOW
Audit: docs/plans/2026-04-24-security-audit.md

## Problem

Mithril bootstrap constructs an archive filename from snapshot.Network and a truncated snapshot.Digest, then checks whether that path is already complete before DownloadSnapshot applies filepath.Base. Malicious snapshot metadata containing path separators can influence the pre-download cache-hit path.

Evidence:
- mithril/bootstrap.go:313-321 constructs archivePath from snapshot metadata and calls isFileComplete.
- mithril/download.go:189-193 normalizes the filename only inside DownloadSnapshot, after the cache-hit path has already run.

## Acceptance criteria

- snapshot.Network is validated against known network identifiers before path use.
- snapshot.Digest is validated as a hex digest with the expected length before path use.
- The same safe filename normalization is used for cache-hit checks and downloads.
- Regression tests cover path separators in snapshot network and digest fields.

## Notes

This requires malicious aggregator metadata plus a matching local file, so it is lower severity than direct extraction traversal.

PR #3303, opened 2026-08-21
fix(mithril): bind snapshot extraction/download to their directory with os.Root

### LINKED ISSUE #3147: security: bind Mithril snapshot extraction to its download directory with os.Root
## Problem

Mithril bootstrap's file extraction/download code (`mithril/download.go`'s
`ExtractArchive` and `DownloadSnapshot`, `mithril/bootstrap_v2.go`'s
`downloadImmutables`) creates directories and writes files using plain
absolute path strings via `os.MkdirAll`/`os.OpenFile`/`os.Create`. These
calls re-resolve the path through the OS on every call and will follow
whatever is actually at that path *at call time* — including a symlink.

If a local actor with filesystem access (e.g. another process on the same
host) replaces a directory component under the download/extraction tree
with a symlink between when it's checked/created and when it's later
written to, extraction can follow that symlink and write outside the
intended directory. This is a TOCTOU (time-of-check-to-time-of-use) issue,
distinct from the content-based path traversal fixed in #2041 — that fix
(`filepath.Base` normalization + `validateSnapshotIdentity`) only
constrains the *string* used to build a path, not the filesystem identity
the resulting path resolves to at the time it's used.

Note: this does **not** cover the zip-slip risk from malicious archive
*contents* — that's already handled in `ExtractArchive` via `path.Clean` +
`validRelPath` + a `HasPrefix` containment check on each tar entry
(`download.go`). This issue is specifically about the local directories
dingo itself creates on disk, not the archive entries it extracts.

## Evidence

- `mithril/download.go` — `ExtractArchive`'s directory/file creation loop
  (`os.MkdirAll`, `os.OpenFile` for each tar entry) and `DownloadSnapshot`'s
  destination-directory setup and resume/retry `os.OpenFile` calls.
- `mithril/bootstrap_v2.go`'s `downloadImmutables` — `os.MkdirAll` for
  `immutableDir` and `archiveDir`.

## Suggested approach

Go's `os.Root` (available in this repo's toolchain, `go.mod` pins
`go 1.26.0`) opens a directory once and binds a handle to that directory's
identity; `Root.MkdirAll`/`Root.OpenFile`/`Root.Create` then operate
relative to that handle and cannot be redirected by a later symlink swap
elsewhere in the path. Open a `Root` once per top-level entry point
(`ExtractArchive`'s `destDir`, `DownloadSnapshot`'s `DestDir`,
`downloadImmutables`'s `downloadDir`) and route all subsequent directory/
file operations through it instead of absolute path strings.

## Acceptance criteria

- `ExtractArchive` performs all directory/file creation through an
  `os.Root` anchored at its `destDir` argument.
- `DownloadSnapshot` performs its destination-directory setup and file
  writes (including resume/retry paths) through an `os.Root` anchored at
  its configured destination directory.
- `downloadImmutables`'s `immutableDir`/`archiveDir` creation goes through
  the same pattern.
- Regression test(s) proving that a pre-existing symlink at a target path
  is rejected rather than followed (can't easily test the live race window
  itself, but `Root` should refuse to traverse a symlink escaping its
  bound directory in the first place).

## Notes

Raised as a review comment on #2041's PR (cubic-dev-ai). Requires local
filesystem write access to the download/extraction directory to exploit,
which is a materially higher bar than #2041's remote-aggregator-metadata
threat model — lower priority, tracked separately so it doesn't get lost
or bundled into an unrelated fix.

## Runtime prerequisite (2026-08-21 audit)

`origin/main` pins `toolchain go1.26.1`, and the release Docker builder pins Go 1.26.3. [GO-2026-4970](https://pkg.go.dev/vuln/GO-2026-4970) affects `os.Root` before Go 1.26.5: on Unix, a final symlink plus trailing slash can escape the root. `govulncheck` reaches the affected API through `ledgerstate.openVerifiedFile`, `mithril.createExtractedFileExclusive`, and `mithril.findImmutableDirIn`.

The fix must therefore also:

- build release artifacts with Go 1.26.6 or newer;
- add the final-symlink/trailing-slash case to the `os.Root` regression coverage; and
- run `govulncheck` as a release gate so runtime regressions cannot silently undercut the filesystem hardening.

arepala-uml:
## Summary

Mithril bootstrap's file extraction/download code created directories and wrote files via plain `os.MkdirAll`/`os.OpenFile` calls on absolute path strings. Those calls re-resolve the path through the OS on every call, so a symlink swapped into the download/extraction tree after the directory was checked but before it was written to (TOCTOU) would be followed instead of rejected. This is distinct from the archive-content path traversal fixed in #2041, which only constrains the string used to build a path, not the filesystem identity it resolves to at write time.

Fixes #3147.

## Changes by file

**`mithril/download.go`**
- `ExtractArchive` opens `os.Root` at `destDir` right after creating it. Directory creation, file open/create, and remove during tar extraction now go through `root.MkdirAll`/`root.OpenFile`/`root.Remove` with paths relative to `destDir`, instead of absolute-path `os.*` calls. Existing zip-slip checks (`validRelPath`, `HasPrefix` containment) are unchanged — `os.Root` is a second, independent layer.
- `DownloadSnapshot` creates `DestDir` and opens `os.Root` on it once, before the retry loop, and threads that `*os.Root` into `downloadSnapshotOnce` (new parameter). Every `os.OpenFile`/`os.Stat`/`os.Remove` on the destination path — full download, restart-after-mismatch, resume-append, 416 size-check, final size-check — now goes through the `root.*` equivalent.
- Helper refactor: `downloadDestinationPath` split into `downloadFilename` (relative, for `Root` calls) and `downloadDestinationPath` (absolute, kept for logging/return value). `downloadFileSize` replaced by `rootFileSize(root, filename)`.

**`mithril/bootstrap_v2.go`**
- `downloadImmutables` opens `os.Root` at `downloadDir` and creates both the `immutable` and `immutable-archives-*` subdirectories via `root.MkdirAll`. The actual file writes into those directories are already covered by `DownloadSnapshot`/`ExtractArchive`, which independently anchor their own `Root` at the specific subdirectory.

**`go.mod`**
- Toolchain `go1.26.1` → `go1.26.7`, fixing GO-2026-4970 / CVE-2026-39822 (a final-symlink-plus-trailing-slash escape in `os.Root` itself, below go1.26.5).

**`Makefile`**
- New `govulncheck` target, pinned to `golang.org/x/vuln@v1.7.0`, following the existing pinned-tool pattern used for `sqlc`.

**`.github/workflows/publish.yml`**
- Added a `govulncheck` step to the tag-triggered `ci` job, which gates `create-draft-release` — a toolchain/dependency vulnerability now blocks the release instead of shipping silently.

## Tests

- `TestExtractArchiveRejectsPreexistingSymlinkEscape` — plants a symlink inside `destDir` pointing outside it before extraction starts; asserts extraction of an archive entry that resolves through it fails and nothing is written outside `destDir`.
- `TestDownloadSnapshotRejectsPreexistingSymlinkEscape` — plants a symlink at the exact download destination path pointing outside `DestDir` before the download starts; asserts `DownloadSnapshot` fails and the outside target is never created.
- `TestDownloadImmutablesRejectsPreexistingSymlinkEscape` / `TestDownloadImmutablesRejectsPreexistingArchiveDirSymlinkEscape` — plants a symlink where `immutableDir`/`archiveDir` should be created; asserts `downloadImmutables` fails at directory creation and nothing is written outside `downloadDir`.
- `TestOsRootRejectsFinalSymlinkWithTrailingSlash` — reproduces the exact CVE-2026-39822 shape (`root.Open("symlink/")`) directly against `os.Root`, independent of dingo's own path normalization.

All five were confirmed to fail against the pre-fix code and pass against the fix (verified by temporarily reverting the source changes and re-running).

---
## Summary by cubic
Prevents TOCTOU symlink escapes in Mithril snapshot and immutable-archive download, extraction, and cleanup by anchoring all filesystem ops to verified directory handles via `os.Root`. Old behavior re-resolved absolute paths and could follow swapped-in symlinks; new behavior opens a directory once through a verified parent and uses `root.*` on sanitized relative names so writes/extracts/removes cannot escape, even if `DestDir`/`archiveDir` are symlinks or get swapped after verification.

- Review notes
  - `DownloadSnapshot` now opens `DestDir` via a verified parent (`openExtractRoot`), routes all `Stat`/`OpenFile`/`Remove` through `root.*`, and its internal `downloadSnapshot` returns the open root for callers that continue with extraction/cleanup; the public wrapper preserves the API and closes roots, and it fixes a root leak when `parentRoot.Close` fails.
  - `ExtractArchive` splits into a thin wrapper and `extractArchiveFile(*os.File, ...)` so callers can extract from a file opened through a retained root. `fetchImmutableArchive` uses that retained root to open the downloaded archive and removes it through the same root on every exit; `downloadImmutables` no longer pre-creates `archiveDir` and no longer performs path-based archive cleanup.
  - Tests cover directory swaps between download and extraction/removal, pre-existing symlinks at `DestDir` or the download filename, and the `os.Root` trailing-slash CVE shape.

- Rollout
  - Build with `go1.26.7+` (fixes GO-2026-4970 / CVE-2026-39822).
  - Releases are gated by `make govulncheck` and a `govulncheck` step in `publish.yml`.

<sup>Written for commit 876d445a82951511b82d83ef589978018dc898ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3303?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security**
  * Strengthened snapshot and immutable archive downloads to prevent writes through symlinked destination paths.
  * Added safeguards against unsafe directory substitutions.

* **Performance**
  * Improved repeated snapshot downloads through connection pooling and keep-alive support.

* **Maintenance**
  * Added automated vulnerability scanning to tagged Linux builds.
  * Updated the Go toolchain version.


mithril/download.go
```diff
@@ -533,20 +541,37 @@ func DownloadSnapshot(
...
-	lastObservedSize := downloadFileSize(destPath)
+
+	// Ensure the destination directory exists, then anchor every
+	// subsequent directory/file operation to its filesystem identity via
+	// os.Root. A plain absolute path re-resolves through the OS on every
+	// os.OpenFile/os.Stat call and would follow a symlink swapped into
+	// the tree after this point (TOCTOU); os.Root binds a handle to
+	// DestDir once and refuses to traverse a symlink that would escape
+	// it.
+	if err := os.MkdirAll(cfg.DestDir, 0o750); err != nil {
```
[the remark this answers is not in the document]
arepala-uml: Valid. Fixed in 54a0576b: DownloadSnapshot now opens a Root on `filepath.Dir(cfg.DestDir)` and creates/opens DestDir itself through `openExtractRoot` (same helper ExtractArchive/openImmutableRoot already use for their own destinations), which refuses a pre-existing symlink at DestDir, or one substituted mid-open, instead of a bare os.MkdirAll+os.OpenRoot on the unverified path. Added TestDownloadSnapshotRefusesSymlinkedDestDir, confirmed to fail against the pre-fix code.

mithril/bootstrap_v2.go
```diff
@@ -762,13 +762,30 @@ func downloadImmutables(
...
+	}
+	archiveRoot, archiveErr := openExtractRoot(downloadRoot, archiveRelDir)
+	closeErr := downloadRoot.Close()
+	if archiveErr != nil {
+		return fmt.Errorf("creating archive directory: %w", archiveErr)
+	}
+	if closeErr != nil {
+		return fmt.Errorf("creating archive directory: %w", closeErr)
 	}
+	archiveRoot.Close()
```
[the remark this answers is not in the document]
arepala-uml: Valid. Fixed in 54a0576b by removing the create step instead of keeping the handle alive: downloadImmutables no longer creates archiveDir up front at all. DownloadSnapshot (mithril/download.go) now creates and opens its DestDir through a verified handle on its own parent on every call (see the reply on the download.go thread), so the window between creation and use no longer exists — every DownloadSnapshot call re-verifies before use, rather than depending on a handle threaded from an earlier caller.

mithril/download.go
```diff
@@ -533,20 +541,37 @@ func DownloadSnapshot(
...
+
+	filename := downloadFilename(cfg)
+	lastObservedSize := rootFileSize(root, filename)
 	consecutiveIdleRetries := 0
 	transientRetries := 0
 	for attempt := 1; ; attempt++ {
-		startSize := downloadFileSize(destPath)
-		path, err := downloadSnapshotOnce(ctx, cfg)
+		startSize := rootFileSize(root, filename)
+		path, err := downloadSnapshotOnce(ctx, cfg, root)
```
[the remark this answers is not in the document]
arepala-uml: The DestDir-being-a-symlink case (this PR's own regression) is fixed — see the other download.go thread. The residual case here, the returned path being reused by a caller after this function returns, is a pre-existing characteristic of all five current callers of DownloadSnapshot (mithril/bootstrap.go and bootstrap_v2.go, none of them changed by this PR); ExtractArchive's own os.Open(archivePath) on main has the same shape. Reworking DownloadSnapshot to return an open handle instead of a path would be a separate, larger contract change across every caller and downstream consumer (checksum verification, extraction), out of scope for this fix. Not making further changes here; this is a reasonable follow-up if the project wants to close that class more broadly, but it isn't part of #3147's acceptance criteria and isn't a regression from this PR.

mithril/bootstrap_v2.go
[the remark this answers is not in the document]
arepala-uml: Valid, same root cause as the Cubic thread on this same line range. Took a different fix than the suggested approach of keeping archiveRoot alive and threading it through fetchImmutableArchive: removed downloadImmutables's up-front archiveDir creation entirely instead. DownloadSnapshot (mithril/download.go) now creates and verifies its own DestDir through a parent-rooted handle (openExtractRoot) on every call, so the gap closes at the point it actually matters without plumbing a shared root through fetchImmutableArchive and cleanup. This also matches the existing convention where DownloadSnapshot is its own top-level entry point, same as ExtractArchive/openImmutableRoot. Fixed in 54a0576b.

mithril/download.go
```diff
@@ -533,20 +541,50 @@ func DownloadSnapshot(
...
+	}
+	root, rootErr := openExtractRoot(parentRoot, filepath.Base(cleanDestDir))
+	closeErr := parentRoot.Close()
+	if rootErr != nil {
+		return "", fmt.Errorf("creating download directory: %w", rootErr)
+	}
+	if closeErr != nil {
+		return "", fmt.Errorf("creating download directory: %w", closeErr)
+	}
+	defer root.Close()
```
wolf31o2: P1: `root` is closed when this function returns, but the returned path is resolved again through `cfg.DestDir`. Replacing `DestDir` with a symlink from `OnProgress` makes `DownloadSnapshot` succeed while the returned path names an external file; `fetchImmutableArchive` then opens that file and `os.Remove(archivePath)` deletes it. Carry the rooted directory/file identity through extraction and cleanup, and cover the post-download swap.

wolf31o2 (changes requested):
One blocking filesystem-identity gap remains.

mithril/download.go
arepala-uml: Valid — this is the residual gap after the DestDir-hardening fix. Fixed in 21c9dc58: `downloadSnapshot` (the internal implementation behind `DownloadSnapshot`) now returns its still-open `*os.Root` instead of closing it before returning, and `fetchImmutableArchive` opens and removes the downloaded archive through that retained root (`root.Open`/`root.Remove`) instead of re-resolving `archiveDir` by path. Extraction goes through a new `extractArchiveFile` that takes an already-open `*os.File` rather than a path, so a directory swapped in for `archiveDir`'s name after the download returns cannot redirect either operation.

Added `TestFetchImmutableArchiveSurvivesArchiveDirSwapAfterDownload`: it swaps `archiveDir` for a symlink to a decoy directory the instant the download finishes (via a RoundTripper reacting to the response body's EOF, not a sleep) and confirms an external canary file survives untouched while the real archive is what gets extracted. Confirmed it fails against the pre-fix code with a zstd magic-number-mismatch error (extraction reading the decoy) and passes with the fix.

mithril/bootstrap_v2.go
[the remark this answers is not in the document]
arepala-uml: Valid, same root cause as the reviewer thread on download.go. Fixed in 21c9dc58 by anchoring `fetchImmutableArchive`'s extraction and cleanup to the directory root `DownloadSnapshot` verifies, instead of adding a shared `archiveRoot` in this function as suggested: `downloadSnapshot` now returns its still-open `*os.Root` rather than closing it, and `fetchImmutableArchive` opens (`root.Open`) and removes (`root.Remove`) the downloaded archive through that root. Extraction takes an already-open `*os.File` via a new `extractArchiveFile` instead of a path. A directory swapped in for `archiveDir`'s name after the download returns can no longer redirect either operation.

Added `TestFetchImmutableArchiveSurvivesArchiveDirSwapAfterDownload`, which swaps `archiveDir` after the download completes and confirms an external canary file is untouched while the real archive is what gets extracted; confirmed it fails against the pre-fix code and passes with the fix.

mithril/bootstrap_v2.go
```diff
@@ -1102,19 +1120,29 @@ func fetchImmutableArchive(
 			AllowInsecureHTTP:   cfg.AllowInsecureHTTP,
 		},
 	)
+	if root != nil {
+		defer root.Close()
+	}
 	if err != nil {
 		return err
```
wolf31o2: P1: Failure cleanup still re-resolves `archiveDir`. If extraction fails after the swap, this returns and `downloadImmutables` calls `os.Remove(immutableArchivePath(...))` at lines 931-933; digest-verification failure repeats it at lines 949-951. Both can delete the same-named external file through the replacement symlink. Remove through the retained root on every exit, eliminate the two path-based deletes, and cover the error path.

mithril/download.go
```diff
@@ -525,28 +533,82 @@ func DownloadSnapshot(
...
+	if err != nil {
+		return "", nil, fmt.Errorf("opening download parent: %w", err)
+	}
+	root, rootErr := openExtractRoot(parentRoot, filepath.Base(cleanDestDir))
+	closeErr := parentRoot.Close()
+	if rootErr != nil {
+		return "", nil, fmt.Errorf("creating download directory: %w", rootErr)
+	}
+	if closeErr != nil {
+		return "", nil, fmt.Errorf("creating download directory: %w", closeErr)
```
wolf31o2, line 600: P2: If `parentRoot.Close` fails after `openExtractRoot` succeeds, this drops `root` without closing it. Return `root` with the error so the caller closes it, or close it here before returning.

mithril/bootstrap_v2_test.go
```diff
@@ -1107,6 +1108,147 @@ func TestBootstrapUnsupportedBackend(t *testing.T) {
...
+	n, err := b.ReadCloser.Read(p)
+	if errors.Is(err, io.EOF) && b.swapped.CompareAndSwap(false, true) {
+		// Rename the directory entry aside rather than removing it: the
+		// downloaded file inside it must survive the swap, reachable only
+		// through the *os.Root opened on it before this point. Renaming
+		// changes the name, not the directory's contents or the inode a
+		// held fd already refers to.
+		orphaned := b.oldDir + ".orphaned"
+		if err := os.Rename(b.oldDir, orphaned); err == nil {
+			_ = os.Symlink(b.newDir, b.oldDir)
```
wolf31o2: P2: The regression can pass without performing the swap because both `Rename` and `Symlink` errors are ignored. Propagate the error or assert that `archiveDir` is the expected symlink before accepting the canary/extraction assertions.

wolf31o2 (changes requested):
Failure-path cleanup remains unanchored.

mithril/bootstrap_v2.go
arepala-uml: Valid — this is the failure-path counterpart to the fix in the prior round. Fixed in 70de8706: `fetchImmutableArchive` now removes its own archive file through the retained root in a `defer` that runs on every exit — success, a download failure leaving a partial file, or an extraction failure leaving a complete one — instead of leaving cleanup to the caller. Deleted both `os.Remove(immutableArchivePath(...))` calls in `downloadImmutables` (the ones at the old lines 931-933 and 949-951): they're now redundant, since `fetchImmutableArchive` has already cleaned up by the time either would run.

Added `TestFetchImmutableArchiveCleansUpThroughRootOnExtractionFailure`, which drives a genuine extraction failure and confirms the cleanup that runs afterward doesn't touch an external file through the swap. Verified both this and the existing success-path test by temporarily reverting the anchored removal back to the old bare-path form — both fail with the canary deleted — then confirmed they pass again with the fix.

mithril/download.go
arepala-uml: Valid. Fixed in 70de8706: when `rootErr == nil` but `closeErr != nil`, `root` is now closed here before returning instead of being silently dropped. Went with closing it rather than returning it with the error, since a caller that sees an error return has no reason to expect a handle to clean up.

mithril/bootstrap_v2_test.go
arepala-uml: Valid. Fixed in 70de8706: `swapOnEOFBody` now records a `Rename`/`Symlink` failure on the shared transport instead of ignoring it, and the test asserts `transport.swapErr` is nil and `transport.swapped` is true, then confirms the swap actually landed on disk via `os.Readlink(archiveDir) == outsideDir` — all before trusting the canary/extraction assertions downstream.

wolf31o2 (approved):
Verified the current head closes the anchored-cleanup gaps from both prior review rounds. Failure-path archive removal stays on the retained os.Root, the parent-root close error no longer leaks the child root, and both swap regressions prove the substitution occurred. The focused race tests pass locally, and current-head CI is green.

PR #3312, opened 2026-08-21
fix(api): define Byron-era protocol-parameter behavior

### LINKED ISSUE #3247: Define Byron-era behavior for protocol-parameter API consumers
## Summary

Define explicit Byron-era behavior for every consumer of `LedgerState.GetCurrentPParams()`.

During a genuine Byron-prefix from-genesis synchronization, `GetCurrentPParams()` intentionally returns `nil`. Byron has no protocol-parameter CBOR. Do not substitute Shelley-shaped protocol parameters.

## Rationale

PR #3241 corrects the startup-era classification for real Byron networks and for Shelley-at-epoch-zero networks such as preview. The remaining behavior affects real Byron prefixes on mainnet and preprod. Several API paths currently treat missing protocol parameters as an error, while one path silently converts the absence to `0`.

## Affected areas

- `api/utxorpc/query.go`
  - `ReadParams` currently returns `current protocol parameters empty`.
- `api/blockfrost/adapter.go`
  - `CurrentProtocolParams` currently returns `protocol parameters not available`.
  - The protocol-parameter path near the certificate deposit logic currently returns `protocol parameters not available`.
  - `drepInactivityPeriod` silently returns `0` for nil protocol parameters.
- `api/mesh/construction.go`
  - Trace and define behavior for its `GetCurrentPParams()` consumer.

## Required changes

1. Define API-specific Byron-era responses for all five consumers.
2. Keep `GetCurrentPParams()` nil during a genuine Byron prefix.
3. Do not use Shelley-shaped defaults for Byron.
4. Make protocol-parameter unavailability explicit where a response cannot represent Byron semantics.
5. Ensure `drepInactivityPeriod` does not silently treat unavailable parameters as a configured zero.
6. Add regression tests for the UTxORPC and Blockfrost behaviors, plus coverage for the Mesh consumer after its behavior is defined.
7. Preserve existing post-Byron behavior.

## Acceptance criteria

- A from-genesis Byron-prefix state does not cause an unintended Shelley parameter substitution.
- `ReadParams` has defined and tested Byron-era behavior.
- All Blockfrost protocol-parameter consumers have defined and tested Byron-era behavior.
- `drepInactivityPeriod` distinguishes unavailable protocol parameters from a configured value of `0`.
- The Mesh construction consumer is traced, implemented, and tested according to its defined Byron-era behavior.
- Existing Shelley-and-later API behavior remains unchanged.

## Backlinks

- PR: https://github.com/blinklabs-io/dingo/pull/3241
- Review comment: https://github.com/blinklabs-io/dingo/pull/3241#discussion_r3825641086
- Requested by: @wolf31o2

chrisguiney:
Closes https://github.com/blinklabs-io/dingo/issues/3247

Byron carries no protocol-parameter CBOR, so `GetCurrentPParams()` returns nil for a genuine Byron prefix. Every consumer treated that absence as a fault or, in one case, converted it to a number. This defines what the absence means for each API. No path substitutes Shelley-shaped parameters.

## UTxORPC

`ReadParams` returned a bare `errors.New`, which reaches callers as `CodeUnknown` — indistinguishable from a genuine server fault. It now returns `CodeFailedPrecondition` with a Byron sentinel, and returns before the tip lookup, which has nothing useful to pair with an absent parameter set.

`FailedPrecondition` rather than `Unavailable` because the latter signals a transient condition clients should retry with backoff, and a Byron prefix can persist for days of synchronization.

v1beta `ReadParams` is served by rewriting the path onto the alpha handler, so both versions are covered by the one change.

## Blockfrost

`CurrentProtocolParams` and `protocolParamsForSlot` now return `ErrProtocolParamsUnavailable`, and `GET /epochs/latest/parameters` maps it to **404 rather than 500**, matching the `ErrEpochNotFound` precedent already established for absent epoch data. A Byron prefix is an expected stage of a from-genesis sync; reporting it as an internal error reads as an outage and trips alerting. Genuine conversion and storage failures still return 500, with a test pinning that the carve-out stays narrow.

## drep_activity

`drepInactivityPeriod` now reports availability alongside its value, delegating to a pure `drepInactivityFromPParams` helper.

The consequential half is `drepStatus`, which gated expiry derivation on `inactivityPeriod > 0`. That guard had a second defect beyond the silent-zero conversion the issue describes: a chain that genuinely configures `drep_activity` to 0 — meaning a DRep expires the epoch it last acted — was skipped as though the parameter did not exist. The gate is now the availability flag, and both call sites thread it through.

## Mesh

No behavior change. `/construction/metadata` already returned a retriable `ErrUnavailable` with the cause attached and substituted no fee, which is the correct Byron-era definition. The two added tests lock that contract rather than driving a change.

## Testing

Written test-first: the failing tests are a separate commit, and each asserts no Shelley-shaped substitution alongside its status-code assertion.

- All `api/...` suites pass, as does `ledger/...`.
- `golangci-lint run ./api/...` reports 0 issues; `gofmt` clean; added lines within the 80-character limit.
- Full `go test ./...` is 75 packages ok. `database/plugin/blob/{aws,gcs}` fail to build (`undefined: NewWithOptions` in `tombstone_test.go`); verified identical on a clean `upstream/main` worktree, so pre-existing and unrelated.

## Two points for review

**The code choice is deliberate but inconsistent across protocols.** UTxORPC gets `FailedPrecondition` while Mesh keeps retriable `Unavailable`. Defensible per protocol for the retry-semantics reason above, but flagging it rather than leaving it to be discovered.

**Scope boundary.** `PoolDetail` and `PoolsExtended` also call `CurrentProtocolParams`. They wrap the error with `%w`, so the sentinel stays matchable, but their handlers still map to 500. Stake pools do not exist in Byron either, so those endpoints arguably deserve the same treatment — left for a follow-up, since the issue names five consumers and those are not among them.

https://claude.ai/code/session_01R79vNFGWAWtVVKH32XkN7s

---
## Summary by cubic
Defines Byron-era protocol-parameter behavior and fixes DRep expiry derivation. Previously Byron absence surfaced as 500s or coerced zeros; now APIs return explicit sentinels and `drepStatus` tracks expiry knownness to avoid zero-as-sentinel bugs.

- `api/utxorpc`: `ReadParams` returns FailedPrecondition with `ErrByronProtocolParams` and exits before tip lookup. Shelley-era states return parameters unchanged.
- `api/blockfrost`: `CurrentProtocolParams`, `EpochProtocolParams`, and `protocolParamsForSlot` return `ErrProtocolParamsUnavailable` (epoch-backed paths wrap epoch/era for logs and guard a nil decoder). `GET /epochs/latest/parameters` and `GET /epochs/{number}/parameters` map the sentinel to 404 and log at debug; other failures stay 500. DRep logic: `drepInactivityPeriod` returns (value, available), and `drepStatus` derives expiry only when inactivity is known and gates on expiryKnown; `drep_activity=0` expires at last activity, and epoch-zero expiries count.
- `api/mesh`: No behavior change. `/construction/metadata` keeps retriable `ErrUnavailable` and does not substitute fees.
- No path substitutes protocol parameters when unavailable. Intentional inconsistency remains: `api/utxorpc` uses FailedPrecondition while `api/mesh` uses Unavailable; stake-pool endpoints still map to 500 and can be aligned in a follow-up.

<sup>Written for commit f12393679750e3210d69d6c8c2129c0dbcce22e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3312?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when protocol parameters are unavailable in the Byron era.
  * Epoch parameter requests now return a clear 404 response instead of an internal server error.
  * Construction metadata requests now return a retriable 503 error without incomplete fee information.
  * UTxO RPC parameter requests now return a clear failed-precondition error when parameters are unavailable.
  * DRep activity and expiry calculations now correctly distinguish unavailable settings from configured zero values.

### DIFF
--- a/api/blockfrost/adapter.go
+++ b/api/blockfrost/adapter.go
@@ -706,10 +706,27 @@ func (a *NodeAdapter) EpochProtocolParams(
 		)
 	}
 	if len(pparamRows) == 0 {
+		// The epoch row resolved above, so the epoch itself exists; only
+		// its parameters do not. Byron is the era where that is the norm
+		// rather than a gap, and it stays reachable long after a sync
+		// completes via GET /epochs/0/parameters. Reporting "epoch not
+		// found" would tell the caller something false about what the node
+		// holds.
 		return ProtocolParamsInfo{}, fmt.Errorf(
 			"get protocol parameters for epoch %d: %w",
 			epoch,
-			ErrEpochNotFound,
+			ErrProtocolParamsUnavailable,
+		)
+	}
+	if era.DecodePParamsFunc == nil {
+		// ByronEraDesc defines no decoder because the era has no parameter
+		// CBOR to decode. Unreachable while the empty-rows check above
+		// runs first, but the nil call would panic if that ever reordered.
+		return ProtocolParamsInfo{}, fmt.Errorf(
+			"get protocol parameters for epoch %d: era %d: %w",
+			epoch,
+			era.Id,
+			ErrProtocolParamsUnavailable,
 		)
 	}
 	pparamRow := pparamRows[0]
--- a/api/blockfrost/adapter.go
+++ b/api/blockfrost/adapter.go
@@ -84,7 +84,20 @@ var (
 func drepInactivityFromPParams(
 	pparams lcommon.ProtocolParameters,
 ) (uint64, bool) {
-	return 0, false
+	// Dijkstra embeds ConwayProtocolParameters, but a type switch matches
+	// concrete types, so *Dijkstra never falls into the *Conway case and each
+	// era needs its own arm.
+	switch pp := pparams.(type) {
+	case *dijkstra.DijkstraProtocolParameters:
+		return pp.DRepInactivityPeriod, true
+	case *conway.ConwayProtocolParameters:
+		return pp.DRepInactivityPeriod, true
+	default:
+		// Byron has no protocol parameters at all, and every era before
+		// Conway has parameters but no DRep semantics. Neither can report a
+		// drep_activity value, and neither may borrow one.
+		return 0, false
+	}
 }
 
 // TransactionSubmitter accepts raw transaction CBOR for mempool admission.
@@ -623,9 +636,10 @@ func (a *NodeAdapter) CurrentProtocolParams() (
 ) {
 	pparams := a.ledgerState.GetCurrentPParams()
 	if pparams == nil {
-		return ProtocolParamsInfo{}, errors.New(
-			"protocol parameters not available",
-		)
+		// A Byron prefix has no protocol-parameter CBOR to report. Surface
+		// the sentinel so the handler can answer "not found" rather than
+		// reporting a node fault for an expected stage of a genesis sync.
+		return ProtocolParamsInfo{}, ErrProtocolParamsUnavailable
 	}
 	info, err := protocolParamsInfoFromNative(
 		pparams,
@@ -1227,17 +1241,13 @@ func (a *NodeAdapter) predefinedDRep(
 }
 
 // drepInactivityPeriod returns the Conway-era drep_activity protocol
-// parameter (epochs of inactivity before a DRep expires), or 0 when it
-// is unavailable.
+// parameter (epochs of inactivity before a DRep expires) from the current
+// ledger state. The second return reports whether a value was available; see
+// drepInactivityFromPParams.
 func (a *NodeAdapter) drepInactivityPeriod() (uint64, bool) {
-	switch pp := a.ledgerState.GetCurrentPParams().(type) {
-	case *conway.ConwayProtocolParameters:
-		return pp.DRepInactivityPeriod, true
-	case *dijkstra.DijkstraProtocolParameters:
-		return pp.DRepInactivityPeriod, true
-	default:
-		return 0, false
-	}
+	return drepInactivityFromPParams(
+		a.ledgerState.GetCurrentPParams(),
+	)
 }
 
 // drepStatus derives the Blockfrost retirement/expiry view of a DRep
@@ -1251,6 +1261,7 @@ func drepStatus(
 	registrationEpoch uint64,
 	currentEpoch uint64,
 	inactivityPeriod uint64,
+	inactivityKnown bool,
 ) (retired bool, expired bool, lastActive uint64) {
 	retired = !active
 	lastActive = lastActivityEpoch
@@ -1258,7 +1269,11 @@ func drepStatus(
 		lastActive = registrationEpoch
 	}
 	expiry := expiryEpoch
-	if expiry == 0 && inactivityPeriod > 0 {
+	// Derive an expiry only when the era actually reports drep_activity.
+	// Gating on "inactivityPeriod > 0" instead would conflate an era with no
+	// DRep semantics against a chain that set drep_activity to 0, where a
+	// DRep expires the epoch it last acted.
+	if expiry == 0 && inactivityKnown {
 		expiry = lastActive + inactivityPeriod
 	}
 	expired = !retired && expiry > 0 && expiry <= currentEpoch
@@ -1328,7 +1343,7 @@ func (a *NodeAdapter) drepByCredentialTag(
 		)
 	}
 
-	inactivityPeriod, _ := a.drepInactivityPeriod()
+	inactivityPeriod, inactivityKnown := a.drepInactivityPeriod()
 	retired, expired, lastActive := drepStatus(
 		drep.Active,
 		drep.LastActivityEpoch,
@@ -1336,6 +1351,7 @@ func (a *NodeAdapter) drepByCredentialTag(
 		registrationEpoch.EpochId,
 		a.ledgerState.CurrentEpoch(),
 		inactivityPeriod,
+		inactivityKnown,
 	)
 
 	// Echo the identifier form the caller used: CIP-129 inputs carry
@@ -1423,7 +1439,7 @@ func (a *NodeAdapter) DReps(
 	}
 
 	currentEpoch := a.ledgerState.CurrentEpoch()
-	inactivity, _ := a.drepInactivityPeriod()
+	inactivity, inactivityKnown := a.drepInactivityPeriod()
 
 	type entry struct {
 		predefined *uint64
@@ -1455,6 +1471,7 @@ func (a *NodeAdapter) DReps(
 			epochForSlot(regSlot),
 			currentEpoch,
 			inactivity,
+			inactivityKnown,
 		)
 		entries = append(entries, entry{
 			credential: drep.Credential,
@@ -5012,7 +5029,7 @@ func (a *NodeAdapter) protocolParamsForSlot(
 	if epoch == nil {
 		pparams := a.ledgerState.GetCurrentPParams()
 		if pparams == nil {
-			return nil, errors.New("protocol parameters not available")
+			return nil, ErrProtocolParamsUnavailable
 		}
 		return pparams, nil
 	}
--- a/api/blockfrost/adapter.go
+++ b/api/blockfrost/adapter.go
@@ -1268,15 +1268,24 @@ func drepStatus(
 	if lastActive == 0 {
 		lastActive = registrationEpoch
 	}
+	// Track whether an expiry is known separately from its value. A derived
+	// expiry of 0 is legitimate — drep_activity 0 on a DRep that last acted
+	// in epoch 0 — so testing the number for zero would read a real expiry as
+	// "none known" and report that DRep active forever.
+	//
+	// A stored expiry_epoch of 0 still means "not recorded": the column has
+	// no null, and that convention predates this function.
 	expiry := expiryEpoch
+	expiryKnown := expiry > 0
 	// Derive an expiry only when the era actually reports drep_activity.
 	// Gating on "inactivityPeriod > 0" instead would conflate an era with no
 	// DRep semantics against a chain that set drep_activity to 0, where a
 	// DRep expires the epoch it last acted.
-	if expiry == 0 && inactivityKnown {
+	if !expiryKnown && inactivityKnown {
 		expiry = lastActive + inactivityPeriod
+		expiryKnown = true
 	}
-	expired = !retired && expiry > 0 && expiry <= currentEpoch
+	expired = !retired && expiryKnown && expiry <= currentEpoch
 	return retired, expired, lastActive
 }
 
--- a/api/blockfrost/adapter.go
+++ b/api/blockfrost/adapter.go
@@ -67,7 +67,25 @@ var (
 	ErrMempoolFull         = errors.New("mempool full")
 	ErrTransactionNotFound = errors.New("transaction not found")
 	ErrInvalidStakeAddress = errors.New("invalid stake address")
+	// ErrProtocolParamsUnavailable reports that no protocol parameters exist
+	// for the requested point. Byron carries no protocol-parameter CBOR, so a
+	// genuine Byron prefix reaches this during a from-genesis sync; it is an
+	// expected state rather than a node fault, and callers must not
+	// substitute Shelley-shaped defaults for it.
+	ErrProtocolParamsUnavailable = errors.New(
+		"protocol parameters not available",
+	)
 )
+
+// drepInactivityFromPParams reads the Conway-era drep_activity parameter from
+// pparams. The second return reports whether the era defines the parameter at
+// all, which a bare uint64 cannot express: a chain that genuinely configured
+// drep_activity to 0 and an era with no DRep semantics are different facts.
+func drepInactivityFromPParams(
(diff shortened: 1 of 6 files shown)


chrisguiney:
@coderabbitai review

api/blockfrost/adapter.go
```diff
@@ -4993,7 +5029,7 @@ func (a *NodeAdapter) protocolParamsForSlot(
 	if epoch == nil {
 		pparams := a.ledgerState.GetCurrentPParams()
 		if pparams == nil {
-			return nil, errors.New("protocol parameters not available")
+			return nil, ErrProtocolParamsUnavailable
```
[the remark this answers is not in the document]
chrisguiney: Good catch — confirmed and fixed in 2b19293b.

You are right that this is the branch a Byron slot actually takes once the chain has recorded epochs: the epoch row resolves, so the lookup never reaches the `GetCurrentPParams` fallback I had sentineled. It calls `db.GetPParams`, which returns `(nil, nil)` because Byron writes no protocol-parameter row — the era defines no `DecodePParamsFunc` at all — and the untyped `errors.New` meant callers could not distinguish it. `/epochs/latest/parameters` would have answered 500 for precisely the case this PR exists to define.

Both branches now return `ErrProtocolParamsUnavailable`, with the epoch-backed one wrapped as `epoch %d era %d: %w` so the cause stays legible in logs while remaining matchable via `errors.Is`.

Regression coverage added as `TestProtocolParamsForSlot_ByronEpochRowSentinel`, which inserts an `era_id = 0` epoch row and asserts the sentinel — it fails against the previous code.

chrisguiney:
@cubic-dev-ai review

api/blockfrost/adapter.go
```diff
@@ -1233,14 +1261,19 @@ func drepStatus(
...
 	expiry := expiryEpoch
-	if expiry == 0 && inactivityPeriod > 0 {
+	// Derive an expiry only when the era actually reports drep_activity.
+	// Gating on "inactivityPeriod > 0" instead would conflate an era with no
+	// DRep semantics against a chain that set drep_activity to 0, where a
+	// DRep expires the epoch it last acted.
+	if expiry == 0 && inactivityKnown {
 		expiry = lastActive + inactivityPeriod
 	}
 	expired = !retired && expiry > 0 && expiry <= currentEpoch
```
wolf31o2: `inactivityKnown` now distinguishes a configured `drep_activity = 0`, but the final `expiry > 0` guard still loses that distinction when the DRep's last-active/registration epoch is also 0. In that case the derived expiry is legitimately 0, yet this returns `expired=false` forever, contradicting the new configured-zero semantics. Please track whether expiry was derived/known separately from its numeric value and add the epoch-zero regression case (known=true, last activity and registration 0).

wolf31o2 (changes requested):
Focused API review found one configured-zero edge case. The Byron-specific response paths and sentinel handling otherwise look consistent.

api/blockfrost/adapter.go
chrisguiney: Confirmed and fixed in 037bf97f.

You are right, and it is the same confusion the flag was meant to end, one level further down: I gated the derivation on `inactivityKnown` but left the final guard reading the derived number. A DRep registered in epoch 0 that has never acted, on a chain with `drep_activity = 0`, derives a legitimate expiry of 0, and `expiry > 0` reads that as "no expiry known" — reporting it active forever.

`drepStatus` now tracks `expiryKnown` alongside the value and gates on that:

```go
expiry := expiryEpoch
expiryKnown := expiry > 0
if !expiryKnown && inactivityKnown {
	expiry = lastActive + inactivityPeriod
	expiryKnown = true
}
expired = !retired && expiryKnown && expiry <= currentEpoch
```

A stored `expiry_epoch` of 0 still means "not recorded" — the column has no null and that convention predates this function — but a derived expiry counts as known whatever its value.

Regression case added as requested (`known=true`, last activity and registration both 0), asserted at `currentEpoch` 0 and 5 since the boundary `0 <= 0` is the one most likely to regress. It fails against the previous code at both.

wolf31o2: Verified at 037bf97f. Reverting only the guard — `expiryKnown` back to `expiry > 0` — fails `configured zero at epoch zero expires` at both currentEpoch 0 and 5 while the rest of the package still compiles and passes, so the new case pins the fix rather than the surrounding rename. Both `drepStatus` call sites thread `inactivityKnown`.

The remaining zero-fallbacks in the function are not the same conflation: a DRep cannot act before it registers, so a genuine `lastActivityEpoch` of 0 implies `registrationEpoch` 0 and the `lastActive == 0` fallback is a no-op.

api/blockfrost/adapter.go
```diff
@@ -67,8 +67,39 @@ var (
 	ErrMempoolFull         = errors.New("mempool full")
 	ErrTransactionNotFound = errors.New("transaction not found")
 	ErrInvalidStakeAddress = errors.New("invalid stake address")
+	// ErrProtocolParamsUnavailable reports that no protocol parameters exist
+	// for the requested point. Byron carries no protocol-parameter CBOR, so a
+	// genuine Byron prefix reaches this during a from-genesis sync; it is an
+	// expected state rather than a node fault, and callers must not
+	// substitute Shelley-shaped defaults for it.
+	ErrProtocolParamsUnavailable = errors.New(
```
wolf31o2, line 75: Non-blocking, and outside the five consumers the issue names: `EpochProtocolParams` is the one Byron protocol-parameter consumer that stays reachable after a full sync. `GET /api/v0/epochs/0/parameters` on a synced mainnet node resolves the Byron epoch row, finds no parameter row, and returns `ErrEpochNotFound` at `adapter.go:706`. Confirmed against an inserted `era_id = 0` epoch row: `errors.Is(err, ErrEpochNotFound)` is true and `errors.Is(err, ErrProtocolParamsUnavailable)` is false.

The status is 404 either way, so there is no behavioral defect here and nothing that blocks this PR. Two notes for the follow-up:

- The epoch exists; only its parameters do not. That is exactly the fact `ErrProtocolParamsUnavailable` now names, and a caller branching on `ErrEpochNotFound` gets a misleading answer for an epoch the node holds.
- `handleEpochParams` calls `b.logger.Error` before the `ErrEpochNotFound` branch (`handlers.go:314`), so every Byron-epoch parameter query logs at error level. That is the alerting noise this change's own rationale cites for `handleLatestEpochParams`, which logs `Debug` for the same expected absence.

Separately, the `era.DecodePParamsFunc` call at `adapter.go:715` is unguarded and `ByronEraDesc` defines no decoder. The `len(pparamRows) == 0` return covers it today, so this matters only if that ordering changes.

wolf31o2 (dismissed):
Approving. The epoch-zero expiry finding from the previous review is fixed at 037bf97f and verified: reverting only the guard fails `configured zero at epoch zero expires` at currentEpoch 0 and 5, and both `drepStatus` call sites thread `inactivityKnown`.

Checks run at 037bf97f in a clean worktree: `go test ./api/...` and `go test -race ./api/...` pass, `golangci-lint run ./api/...` reports 0 issues, `gofmt -l api/` empty, `go vet ./api/...` clean.

On the two flagged points:

- The `FailedPrecondition` / `Unavailable` split across UTxORPC and Mesh holds up on the retry-semantics argument given.
- The stake-pool scope boundary needs no follow-up. Neither pool endpoint reaches protocol parameters in Byron: `PoolsExtended` returns `[]` at `adapter.go:2046` when no active pools exist, and `PoolDetail` returns `ErrPoolNotFound` before its `CurrentProtocolParams` call. The same holds for the two `protocolParamsForSlot` callers not named in the issue - `transactionRedeemers` returns early on no redeemers, and `transactionDeposit` returns `(0, nil)` because `ByronEraDesc` defines no `CertDepositFunc`.

One non-blocking note inline on the one Byron consumer that does stay reachable.

DATABASE.md and ARCHITECTURE.md checked and unaffected: no schema change, and the `live_saturation` passage at `ARCHITECTURE.md:4433` still describes the behavior this change preserves.

Cubic is rate-limited until 1 September and did not review; CodeRabbit last reviewed c69f4225, so 2b19293b and 037bf97f carry no bot pass.

api/blockfrost/adapter.go
chrisguiney: Thanks — all three verified and fixed in f1239367 rather than deferred, since they are small and all name the same fact.

**Sentinel.** `EpochProtocolParams` now returns `ErrProtocolParamsUnavailable` when the epoch row resolves but no parameter row exists, and `ErrEpochNotFound` stays for an epoch the node genuinely lacks. Tests assert each is *not* the other, so the distinction cannot quietly collapse again.

**Log level.** You are right that this is the same noise the PR argues against. `handleEpochParams` logged before classifying, so it hit `Error` for both expected answers, not just the Byron one. It now classifies first: `ErrProtocolParamsUnavailable` and `ErrEpochNotFound` both log `Debug` and return 404, genuine failures still log `Error` and return 500. Covered by asserting the absence of `"level":"ERROR"` in captured output for both, plus a companion test that a real failure still logs it.

**Nil decoder.** Worth more than the ordering caveat suggested — with a Byron parameter row present the call panics today, and the regression test panics without the guard. Guarded explicitly and it now returns the sentinel.

Status is still 404 in every case, so no response changed for a client; what changed is which sentinel callers see and what lands in the log.

Note this went in after your approval. Nothing in the approved range changed — the diff is additive plus the two error paths above — but flagging it so you can re-check rather than assume the approval still covers the head commit.

api/blockfrost/adapter.go
```diff
@@ -674,10 +706,27 @@ func (a *NodeAdapter) EpochProtocolParams(
...
+	}
+	if era.DecodePParamsFunc == nil {
+		// ByronEraDesc defines no decoder because the era has no parameter
+		// CBOR to decode. Unreachable while the empty-rows check above
+		// runs first, but the nil call would panic if that ever reordered.
+		return ProtocolParamsInfo{}, fmt.Errorf(
+			"get protocol parameters for epoch %d: era %d: %w",
+			epoch,
+			era.Id,
+			ErrProtocolParamsUnavailable,
```
wolf31o2, line 729: Optional, and a miss from my previous round rather than anything introduced here: the post-decode branch at `adapter.go:748` still returns `ErrEpochNotFound` when `pparams == nil`. That is the same conflation this commit removes two branches above - the epoch row resolved and a parameter row existed, so "epoch not found" is the one fact that is definitely false at that point.

Unreachable today: all seven `DecodePParamsFunc` implementations return `(&ret, nil)` on success, so a nil value with a nil error cannot arise. That makes it exactly as defensive as the guard added here, which is why the inconsistency stands out - two adjacent unreachable branches now name different facts about the same absence.

A one-word change if this file is touched again; not worth a push on its own.

wolf31o2 (approved):
Approving. The non-blocking note from the last review is addressed at f1239367. All three parts verified by reverting each independently and re-running:

- Empty parameter rows now report `ErrProtocolParamsUnavailable`: restoring the old sentinel fails `TestEpochProtocolParams_ByronEpochReportsParamsNotEpoch` on both assertions, and `TestEpochProtocolParams_MissingEpochStillNotFound` pins the other direction.
- Nil-decoder guard: removing it panics with a genuine nil pointer dereference under `TestEpochProtocolParams_ByronRowDoesNotCallNilDecoder`. Given a Byron parameter row the guard is load-bearing, not decorative - the commit message's claim holds.
- Log-after-classify: moving the `Error` call back above the branches fails both `TestHandleEpochParams_ByronEpochNotFoundNotLoggedAsError` subtests, while `TestHandleEpochParams_RealFailureStillLogsError` keeps the 500 carve-out narrow.

The boundary is closed. `handleEpochParams` is the only non-test consumer of `EpochProtocolParams`, and it maps both sentinels to 404, so `GET /epochs/0/parameters` still answers 404 - only the sentinel and the log level moved. The mock-backed test at `blockfrost_test.go:2675` still exercises `ErrEpochNotFound` through the unchanged branch.

Checks at f1239367 in a clean worktree: `go test ./api/...` and `go test -race ./api/...` pass, `golangci-lint run ./api/...` reports 0 issues, `gofmt -l api/` empty, `go vet ./api/...` clean. CI green on all seven checks.

DATABASE.md and ARCHITECTURE.md checked and unaffected: no schema change, and the Blockfrost section at `ARCHITECTURE.md:4350` enumerates endpoints without pinning status codes. No OpenAPI document in the repository covers these routes.

One optional note inline. It does not need another push.

PR #3324, opened 2026-08-22
fix(chainsync): reject future headers

### LINKED ISSUE #3287: chainsync: reject or delay future headers
## Problem

- `validateInboundBlockEnvelope` has no arrival time or current-slot input.
- Header verification has no wall-clock plausibility check.
- Near- and far-future headers can advance block application immediately.

## Acceptance criteria

- Record header arrival time and compare the header slot with the current slot.
- Delay headers within the permitted clock-skew window.
- Reject headers beyond the permitted window.
- Add deterministic boundary tests with a controlled clock.
- Match the Ouroboros future-header check: https://github.com/IntersectMBO/ouroboros-consensus/blob/1356d3e9bf3746d40a7165997b24e6a71fb5e2ef/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/InFutureCheck.hs#L109-L176

wolf31o2:
Closes #3287.

- Preserve block-header arrival time from network ingress through ChainSync processing.
- Apply future-header admission before observed-tip, deduplication, and ledger mutation.
- Wait up to two seconds for bounded clock skew; drop later headers without peer penalty or state mutation.
- Re-intersect each affected connection at the earliest dropped header and cancel pending recovery on connection removal or shutdown.
- Cover boundary, conversion, cursor, cancellation, replay, and timer behavior.

## Validation

- Full race-enabled Go suite with a 20-minute package timeout.
- Exact-head race coverage for future-header admission/resync and packages changed by the latest `main` merge.
- Accelerated all-Dingo DevNet with in-process observers: propagation, agreement, epoch transition, peer interruption, and relay restart.
- Accelerated Dingo/cardano-node conformance DevNet with the same observer and recovery phases.
- `make`, `make import-boundaries`, `make docs-parity`, `make gorm-check`, `make golines`, and `make govulncheck`.
- `make lint` passed before the latest `main` refresh; the test-only baseline static-analysis failures introduced by that refresh are isolated in #3445.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -7845,6 +7845,16 @@ ledger reader reaches the validate stage, and `ls.chain` is consumed by the
 downstream blockfetch server and UTxO RPC sync/watch readers. A block that has
 not passed admission crypto must therefore never enter that chain.
 
+ChainSync admission also enforces the Ouroboros future-header rule before a
+header enters that queue. The network callback records `ChainsyncEvent`'s
+`ArrivalTime` immediately, before chain-selection and EventBus work can delay
+delivery. Ledger admission converts the header slot to its wall-clock onset
+using the active hard-fork summary. A header received no more than two seconds
+early (the `ouroboros-consensus` default clock-skew allowance) waits until its
+slot begins; a header received earlier is rejected and its connection is
+recycled. Judging the recorded arrival rather than the later handler time means
+EventBus or scheduler delay cannot make an invalid early header appear timely.
+
 `blockPipelineEta0Provider` reads the immutable epoch-cache snapshot directly;
 it does not forecast, mutate the cache, or rebuild `HardForkSummary` per
 block. `decodeReadChainBatch` enforces pipeline results only when the serial
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -7848,13 +7848,23 @@ not passed admission crypto must therefore never enter that chain.
 ChainSync admission also enforces the Ouroboros future-header rule before a
 header enters that queue. The raw network callback records `ChainsyncEvent`'s
 `ArrivalTime` immediately, before header decoding, chain-selection, and
-EventBus work can delay delivery. Ledger admission converts the header slot to
-its wall-clock onset using the active hard-fork summary. A header received no
-more than two seconds early (the `ouroboros-consensus` default clock-skew
-allowance) waits until its slot begins; a header received earlier is rejected
-and its connection is recycled. Judging the recorded arrival rather than the
-later handler time means local decode, EventBus, or scheduler delay cannot make
-an invalid early header appear timely.
+EventBus work can delay delivery. The per-peer protocol callback asks ledger
+admission to convert the header slot to its wall-clock onset using the active
+hard-fork summary before observed-tip, dedup, or ledger state changes. A header
+received no more than two seconds early (the `ouroboros-consensus` default
+clock-skew allowance) waits on that peer callback until its slot begins; it
+never sleeps under the node-wide ChainSync dispatch mutex. A resolvable header
+received earlier is dropped without recycling or penalizing the peer, because
+the node cannot distinguish peer skew from a slow local clock. One coalesced
+timer per connection re-intersects the mini-protocol at the earliest dropped
+header's onset so the remote cursor cannot strand the accepted chain; later
+headers from that peer remain withheld until the old mini-protocol has stopped
+and the re-intersection can replay from ledger-accepted points. A slot
+past the hard-fork forecast is deferred to the normal header-verification path,
+which already treats `ErrPastHorizon` as unavailable state rather than peer
+fault. Judging the recorded arrival rather than the later handler time means
+local decode, EventBus, or scheduler delay cannot make an invalid early header
+appear timely.
 
 `blockPipelineEta0Provider` reads the immutable epoch-cache snapshot directly;
 it does not forecast, mutate the cache, or rebuild `HardForkSummary` per
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -7846,14 +7846,15 @@ downstream blockfetch server and UTxO RPC sync/watch readers. A block that has
 not passed admission crypto must therefore never enter that chain.
 
 ChainSync admission also enforces the Ouroboros future-header rule before a
-header enters that queue. The network callback records `ChainsyncEvent`'s
-`ArrivalTime` immediately, before chain-selection and EventBus work can delay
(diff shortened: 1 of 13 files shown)


wolf31o2:
@coderabbitai review

wolf31o2:
Local review at `1ef656b4`, substituting for a bot pass. The rule itself matches upstream — `defaultClockSkew = clockSkewInSeconds 2` confirmed against `InFutureCheck.hs` at the SHA cited in #3287, arrival is timestamped at ingress before the decode-cache wait, era math goes through `hardfork.Summary` so a Byron prefix is handled, and all four mutations I ran fail a test, including two-sided boundary coverage at exactly 2s and 2s+1ns.

Two findings are structural rather than about the constant.

**The slot-onset delay blocks the shared ChainSync dispatch goroutine while holding `chainsyncMutex`.** `waitUntil` at `ledger/chainsync.go:2824` is reached from `:2408`, inside `handleEventChainsyncBlockHeaderWithPending`, which runs under the lock taken at `:226`. The subscription at `ledger/state.go:1349` is a single dispatch goroutine serving every peer, and producers use `PublishBlocking`. A probe confirms `chainsyncMutex.TryLock()` fails while the wait runs.

So a peer whose clock is 1.5s fast makes the node sleep 1.5s *per header* inside a node-wide lock, before any other admission gate — connection-switch, dedup and crypto verification all sit after it at `:2449`. `ledger/pending_publish.go:24-40` records what stalling this same mutex did previously: roughly 217k "event delivery stalled" warnings and the node not answering NtN handshakes. Upstream runs the equivalent `threadDelay` inside the per-peer ChainSync client thread, not in a node-global serialized handler.

**Fail-closed slot conversion drops honest historical headers during catch-up.** `ledger/chainsync.go:2788-2791` returns an error when `slotAtTime(e.ArrivalTime)` fails, before any cheap comparison. During catch-up the applied ledger's horizon is behind wall clock by design, as `ledger/slot_time_converter.go:141-150` states, and the only recovery is a tolerance of one era slot length plus 5s. The chainsync buffer is sized at 100000 to absorb bulk-sync bursts, so once queue latency exceeds roughly 6s every arrival timestamp falls outside the window. Probe against the real converter, summary bounded at slot 1000, wall clock at slot 1,000,000, header slot 900:

    queue lag 0s  -> nil
    queue lag 3s  -> nil
    queue lag 7s  -> resolve header arrival slot: ... past era horizon

A header 999,100 slots in the *past* is rejected by the future-header check. It is not a `headerTooFarInFutureError`, so the peer is not recycled; it keeps streaming headers that are all dropped with an Error and a `LedgerErrorEvent` each. Sync cannot advance, so the horizon never advances, so it does not clear. `replayBufferedHeaderEvents` at `:1610-1620` deletes the buffered slice before the loop and returns on first error, discarding the rest for that peer. Inability to convert an arrival timestamp means the header cannot be *proven* future, so this path should fail open.

**A locally slow clock recycles every peer.** Probe with 1s slots and an honest peer delivering 200ms after true onset: at 3s slow the header is rejected as arriving 2.8s before its onset. Recycle is an unconditional `conn.Close()` with no ban list, so outbound peers hit short-lived-connection backoff and inbound peers accumulate `InboundShortLivedCount` toward cooldown. There is no dedicated log or metric distinguishing this from peer misbehaviour — only the generic `failed to handle block header` Error — where the header-verification recycle immediately below at `:2466` does emit a dedicated Warn with connection id, slot and hash.

Two smaller ones. The check runs after `ouroboros/chainsync.go:785`/`:789` have already published `PeerTipUpdateEvent` and recorded the observed header, so a far-future header updates the peer's `ObservedTip` and feeds the Genesis-mode horizon before the ledger rejects it; upstream applies the check before the header enters the peer's candidate fragment. And `:2804-2810` maps `hardfork.ErrPastHorizon` to a peer fault, which contradicts `ledger/verify_header.go:352-362` handling the same error for the same header and stating explicitly that it is a deferred condition, not a peer fault.

Coverage gaps: no test for the `slotAtTime` error path, so the fail-closed behaviour is untested in both directions; none asserting a non-future error does not recycle; none that the wait releases `chainsyncMutex`; no multi-era fixture.

Validation: `go test -race ./ledger/` ok at 909s, `./ouroboros/` and `./chainsync/` ok, conformance ok, import-boundaries and docs-parity ok, `golangci-lint`, `nilaway` and `modernize` clean on the three packages (local golangci-lint v2.12.1, CI newer). Not run: devnet, which is what would actually exercise the first two findings on a from-genesis sync.

wolf31o2:
Local exact-head review at 821bb152 stands in for the missing bot pass: CodeRabbit was rate-limited and the Cubic check posted no review record. I audited peer-local admission before observed-tip mutation, bounded skew and past-horizon behavior, earliest-timer coalescing, connection-removal and shutdown cleanup, and lock ordering. The prior global-wait, cursor-mutation, conversion, timer, and cleanup failures are covered by controlled regressions. Exact composed ledger and ouroboros race tests, full lint, and CI pass. The composed DevNet did not start services after the delayed prelaunch/build attempt.

arepala-uml:
@wolf31o2  I did not find any remaining code-level bugs in the changed files. It currently conflicts with `main` (`CONFLICTING/DIRTY`), including a conflict in `event/chainsync_resync.go`. Please merge the latest `main`, and resolve the conflicts.

wolf31o2:
Rebased onto current `main` at `f7ba8794` and pushed `d72bbc7d`.

- Resolved the `event/chainsync_resync.go` conflict by retaining the current reason set and adding `FutureHeaderAdmissionRecovery`.
- Confirmed the `ARCHITECTURE.md` addition is self-contained and contains no bare issue or pull-request references.
- Focused ledger/ouroboros race tests, chainsync/event race tests, vet, import-boundaries, docs-parity, gorm-check, golangci-lint, and golines pass.

wolf31o2:
Local exact-head review completed at `365010ecf35f97a2fd736cc61dded43310f5f075`.

I reviewed raw-header arrival timestamping, the two-second future-header admission boundary, past-horizon deferral, pre-observation/pre-dedup ordering, peer-local waiting, far-future drop behavior, resync timer coalescing and arm/removal races, shutdown cleanup, and live storage restoration. I found no remaining actionable issue.

Validation included the full race-enabled Go suite, exact-head focused race tests after the authoritative branch rewrite, accelerated all-Dingo and Dingo/cardano-node DevNets with in-process observers, build/docs/import/GORM/golines checks, and govulncheck. The latest `main` refresh introduced unrelated test-only NilAway/modernize failures; their dedicated repair is #3445.

No metered bot review was requested for this pass.

arepala-uml (approved):

Issue #3228, opened 2026-08-20
Windows extraction publish and unlink resolve their parent by path, not by handle


wolf31o2:
Raised by a cubic finding on #3226 and only partly closed there.

> **Corrected 20 Aug 2026.** The original text of this issue claimed that holding
> the verified parent handles prevents them being swapped, and scoped the
> residual risk to components *above* the opened root. Both were wrong, so the
> remaining exposure is larger than first written. A CodeRabbit finding on #3226
> caught it; the detail is under "Why the handles do not help" below. The
> comments in the three affected files carried the same false claim and were
> corrected in `8f5d4f11`.

## Mechanism

Windows offers no handle-relative rename or unlink through the Go standard
library, so three extraction operations address their target by path:

- `mithril/extract_rename_windows.go` — `renameExtractedDirectory` calls `windows.MoveFile`
- `mithril/extract_unlink_windows.go` — `removeExtractedFile` calls `windows.DeleteFile`
- `mithril/extract_rmdir_windows.go` — `removeEmptyExtractDir` calls `windows.RemoveDirectory`

Each walks every directory component through its parent's handle and confirms it
is the entry the name denotes (`openVerifiedParent`), holding those handles
across the call.

## Why the handles do not help

The walk is worth having, but not for the reason originally given here. It
rejects a symlink or reparse point that is **already in place when it runs** —
the planted-component case. It does not make the verified parents unswappable.

`os.Root` opens every handle with `FILE_SHARE_DELETE`. In go1.26.1 that is
`internal/syscall/windows/at_windows.go`, at lines 151, 223, 252, 268, 334, 375
and 449 — every openat path. A handle opened that way does not stop another
process renaming or deleting the directory, so "Windows refuses to move a
directory while handles are open on it" does not apply to anything `os.Root`
hands us.

The consequence is that the race is not confined to components above the root,
as this issue first said. Any verified component can be substituted between the
walk finishing and `MoveFile`/`DeleteFile`/`RemoveDirectory` resolving the path a
second time, and the substitution is then followed.

For contrast, the Unix paths are genuinely handle-relative (`unlinkat`,
`renameat` through the parent descriptor) and have no equivalent window.

## Complete fix

`NtSetInformationFile` with `FileRenameInformation` for the rename and
`FileDispositionInformation` for the removals, each with `RootDirectory` set to
the verified parent handle. That makes the operation handle-relative and removes
the second path resolution entirely.

## Why it is not in #3226

[...]

PR #3331, opened 2026-08-22
test(mithril): gate the swap tests on a platform that allows the swap

wolf31o2:
## Problem

Two archive-directory swap tests added by #3303 fail at their setup assertion on Windows. The tests hold an `os.Root` handle beneath the directory, and Windows refuses the rename needed to construct the TOCTOU scenario.

## Changes

- Probe directory-swap support on the test goroutine before the archive download callback stages the swap.
- Skip only when the platform cannot rename a directory with an open handle beneath it.
- Preserve the existing symlink capability check and full Linux coverage.

## Validation

The focused Mithril race suite and lint pass. Both tests still run on Linux, and a forced unsupported probe skips them with the platform limitation. Windows CI passes on the current head.

Refs #3228.

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform handling for directory-swap regression tests.
  * Tests now skip unsupported scenarios on Windows instead of reporting false failures.
  * Added capability checks to distinguish unsupported platform behavior from genuine test errors.

### DIFF
--- a/mithril/bootstrap_v2_test.go
+++ b/mithril/bootstrap_v2_test.go
@@ -1194,8 +1194,9 @@ func TestFetchImmutableArchiveSurvivesArchiveDirSwapAfterDownload(
 ) {
 	probeLink := filepath.Join(t.TempDir(), "probe")
 	requireSymlinkSupport(t, t.TempDir(), probeLink)
-	// The swap is staged from the transport's goroutine, so the platform's
-	// willingness to rename a held-open directory has to be settled here.
+	// The response-body callback stages the swap synchronously through io.Copy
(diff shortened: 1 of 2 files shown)


wolf31o2:
Windows validation, via `workflow_dispatch` on `go-test-windows.yml` (the job does not run on pull requests):

| Ref | SHA | mithril package |
|---|---|---|
| `main` | `2031e4fd` | **failure** |
| `fix/3228-windows-mithril-tests` | `45e3822b` | `ok ... 27.099s` |

Run 32599593685. No `FAIL` lines anywhere in the job log.

Linux, on `45e3822b`: 9 checks, none failing.

verbotenj (approved):

PR #3337, opened 2026-08-22
fix(ledger): gate continuation audit on applied tip

wolf31o2:
## Summary

- arm the continuation audit only when the applied ledger tip matches the chainsync rollback point
- disarm a stale prior audit when catch-up leaves the ledger behind that rollback point
- cover both the nil-window and already-armed lifecycle cases
- document the aligned-rollback requirement

## Validation

- controlled fail-before regression: stale audit remains on the prior head
- focused ledger race test passes on the fixed head
- ledger golangci-lint reports zero issues
- ledger NilAway passes
- docs parity passes


wolf31o2:
@coderabbitai review

arepala-uml (approved):

PR #3241, opened 2026-08-20
fix(ledger): preserve Byron prefix during genesis sync

wolf31o2:
## Problem

Preprod genesis startup inferred Shelley at slot 0 even though canonical preprod begins with four Byron epochs. The resulting checkpoint and era metadata caused the slot-86400 Shelley boundary block to fail overlay validation.

## Changes

- Start an empty database in Byron when a Byron genesis exists, unless configuration explicitly declares Shelley at epoch 0.
- Add an ungated declared-hard-fork accessor without changing the experimental accessor used to select hard-fork trigger semantics.
- Keep protocol parameters absent throughout Byron, clear stale values on rollback, and install Shelley parameters at the transition.
- Key boundary validation on the incoming block or header and skip stake rewards while the performance epoch is Byron.
- Warn when an existing database contains the stale pre-Byron epoch-cache shape.

## Operator impact

Existing databases with the stale epoch cache are not rewritten and require a resync from an empty database. Parameter queries return empty or typed errors during the Byron prefix instead of serving Shelley parameters under the wrong era. No schema migration is introduced.

## Validation

Race-enabled ledger tests, all conformance vectors, public-relay fork-boundary fixtures, and a clean preprod run through canonical slot 86400 pass. The fixture control also detects mainnet's Byron epoch boundary block at slot 21600.

PR #3338, opened 2026-08-23
test(ledger): cover Byron Shelley boundary recovery

### LINKED ISSUE #3316: test(ledger): cover the Byron-Shelley boundary crossing and a rollback inside the prefix
## Gap

Two paths the Byron-prefix work rests on have no CI coverage. Raised on #3241 and left out of it deliberately, since both need chain fixtures rather than assertions.

### The boundary crossing itself

Nothing drives an actual Byron to Shelley transition. The behavior is evidenced by a clean preprod sync, but that evidence does not live in CI, so a refactor of the boundary-crossing path in `ledgerProcessBlocksFromSource` would not be caught.

### A rollback inside the Byron prefix

Nothing covers `loadPParams` followed by `reconstructTransitionInfo` after a restart or rollback partway through the prefix. That is reachable on any from-genesis sync that takes a rollback near the boundary.

`TestReconstructTransitionInfoIgnoresStaleShelleyPParamsUnderByron` (`ledger/state_test.go:4797`) asserts the guard holds for the state that path can produce, and its doc comment is explicit that it does not drive the path.

## Tractability

The second is the cheaper one. `ledger/block_event_ordering_test.go` already drives `ls.rollbackChainAndState(...)` against a fixture, so the work is a Byron-era variant of that fixture rather than new machinery: roll back into a Byron epoch and assert `currentPParams` and `transitionInfo` end up consistent with the Byron era.

With the `ppComputed` change on #3241, the expected result is `currentPParams == nil` and no reconstructed transition. Before it, the rollback left a Shelley value under a Byron era.

## Acceptance criteria

- A test drives a rollback into a Byron epoch through `rollbackChainAndState` and asserts `currentPParams` is nil and `transitionInfo` reports no transition.
- That test fails when the assignment is gated on `newPParams != nil` instead of `ppComputed`.
- A test crosses the Byron to Shelley boundary through the normal block-processing path and asserts Shelley parameters are installed before the first post-Byron block is validated.

wolf31o2:
## Problem

#3241 fixes Byron-prefix startup and rollback state, but its tests do not drive the normal Byron-to-Shelley block-processing boundary or a rollback from Shelley into the Byron prefix.

## Change

- Process the canonical mainnet boundary block through the normal validated ledger pipeline and require Shelley parameters before the block advances the tip.
- Roll back from the first Shelley block to the last Byron block and require nil current/previous parameters with no reconstructed transition.

This test-only PR is stacked on #3241 and should be retargeted to `main` after #3241 lands.

## Validation

The #3316 tests pass under repeated race-enabled runs, the related Byron boundary/reconstruction slice passes under the race detector, and full lint passes. Controlled reversions prove the tests detect both the stale-parameter rollback gate and a missing boundary transition.

Inherited from #3241: race-enabled ledger coverage, conformance vectors, public-relay boundary fixtures, and a preprod run through slot 86400.

Closes #3316


wolf31o2:
Superseded by #3339, which contains the same test change as a clean one-commit stack on #3241.

PR #3341, opened 2026-08-23
fix(ledger): reset recovery budget on progress

### LINKED ISSUE #3318: ledger: Mithril boundary rejection budget is defeated when the failure identity varies
## Summary

`observeMithrilBoundaryRejection` (`ledger/replay_recovery.go:824`) keys its tally on failure identity:

```go
if ls.mithrilBoundaryRejection == nil ||
	!ls.mithrilBoundaryRejection.matches(validationErr) {
	ls.mithrilBoundaryRejection = newAtTipRecoveryAttempt(validationErr)
} else {
	ls.mithrilBoundaryRejection.Attempts++
}
```

Any variance in the reported `(BlockPoint, TxHash)` resets it to `Attempts: 1`, not only genuine forward progress. With a varying identity the bound never accumulates and the halt never fires, so the node rotates peers indefinitely instead of reaching the terminal state the budget exists to produce.

Raised on #3297, where an alternating-identity probe confirmed it: 40 cycles, `Attempts` never exceeds 1, no halt.

## Reachability

- The at-tip entry point is provably unaffected. Rejection only fires once `attempts > 1`, which requires `isSameFailure` to already hold.
- `rejectReplayRecoveryAtMithrilBoundary` has no such precondition and can take this path on the very first failure.

In practice a single wedged canonical block's first failing transaction is deterministic, since the transaction-validation loop in `state.go` is sequential with one `txValidationError` construction site. The hole opens when the delivered *block* varies across peer rotations rather than the transaction within it.

## Precedent in this file

The sibling bound already solved this by not keying on identity. From `maxReplayRecoveryNoProgress`'s comment: "the failing block may change and creep forward... so the ledger tip — not the failure identity — is the convergence signal."

## Proposal

Add a tip-keyed bound alongside the identity-keyed one, so consecutive rejections that make no tip progress accumulate regardless of which block reported them. The identity-keyed tally can stay for the same-block case; what is missing is a convergence signal that peer rotation cannot reset.

`resetMithrilBoundaryRejections` already uses tip progress as the clearing signal, so the two would agree on when a wedge has resolved.

## Test-coverage gap

Both test files added by #3297 exercise only the at-tip path: `newTrustWindowLedger` sets `reachedTip.Store(true)`. The replay-recovery path has no dedicated terminal-state test, and it is the path a fresh post-Mithril-bootstrap wedge actually takes, since `reachedTip` is a one-way latch set only after a successful batch commit. The fixture at `ledger/replay_recovery_test.go:225` is a suitable base.

## Acceptance criteria

- Consecutive boundary rejections with differing `(BlockPoint, TxHash)` and no tip progress reach the terminal state within a bounded number of attempts.
- A test drives that alternating-identity sequence and fails without the tip-keyed bound.
- A terminal-state test covers `rejectReplayRecoveryAtMithrilBoundary` with `reachedTip` false.
- Tip progress still clears both bounds.

wolf31o2:
## Summary

- preserve the tip-keyed Mithril boundary convergence bound already landed through #3297
- clear stale same-failure rewind depth only after committed tip progress while preserving it on same-tip replay
- cover the boundary-nil reset path and document the recovery invariant

## Validation

- focused recovery tests under `-race` (5 repeats), plus a post-format repeat
- SQLite rules conformance under `-race`
- golangci-lint, nilaway, modernize, import boundaries, docs parity, GORM guard, and diff checks

Accelerated DevNet was not rerun because this change only resets in-memory recovery bookkeeping after an already-committed tip advance; the focused race and rules conformance checks cover the affected path.

Closes #3318

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -3075,18 +3075,21 @@ window deeper every cycle and the node would fall unboundedly behind the wall
 clock, recoverable only by restart. `maxAtTipRecoveryDescents` consecutive
 distinct failures that fail to advance latch recovery into a hold-at-tip mode
 that suppresses deep rewinds — rewinding only to the ledger tip so ChainSync
-re-delivers rather than descending. The latch clears when the ledger makes
-forward progress past the failing region (`resetAtTipRecoveryDescent`, called
-from the block-apply success path). Because a hold usually indicates local
-ledger validation diverging from the network (a false-positive rejection that
-no rewind can fix) rather than a peer/fork problem, it surfaces the
+re-delivers rather than descending. The latch and the same-`(block, tx)` rewind
+depth both clear only when the ledger makes forward progress past the failing
+region (`resetAtTipRecoveryDescent`, called from the post-commit block-apply
+success path); replaying back to the same tip preserves both. Because a hold
+usually indicates local ledger validation diverging from the network (a
+false-positive rejection that no rewind can fix) rather than a peer/fork
+problem, it surfaces the
 `dingo_ledger_attip_recovery_nonconverging_total` metric and a throttled
 operator warning; a node in this state needs the underlying validation
 divergence resolved.
 
 Both recovery paths refuse a rewind target below the Mithril anchor, and that
-refusal has its own terminal bound (issues #3261 and #3301). Refusing rewinds
-instead to the applied ledger tip and asks ChainSync for a fresh intersection,
+refusal has its own terminal bound (issues #3261, #3301, and #3318). A refusal
+rewinds instead to the applied ledger tip and asks ChainSync for a fresh
+intersection,
 which is an escape only while some peer offers a different chain; for a
 canonical block every peer offers the same one. When the failing block sits a
 short distance past the anchor, every target the rewind schedule produces falls
--- a/ledger/pipeline_halt_test.go
+++ b/ledger/pipeline_halt_test.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/blinklabs-io/dingo/internal/test/testutil"
+	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
 	"github.com/prometheus/client_golang/prometheus"
 	promtestutil "github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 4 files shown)


arepala-uml (approved):
Traced this back through #3297 and #3318 to make sure the fix is scoped correctly, not just re-reading the diff.

#3318 was originally raised against `observeMithrilBoundaryRejection`'s identity-keyed tally during #3297's development. #3297 already replaced that with the tip-keyed `mithrilBoundaryRecoveryProgress{highWaterTipSlot, rejections, halted}` — no identity matching left in that path. This PR closes the sibling gap: `resetAtTipRecoveryDescent` cleared `atTipRecoveryDescentCount`/`atTipRecoveryHolding` on real tip progress already, but left `lastAtTipRecovery` (the same-`(block,tx)` attempt counter) stale. Since `matches()` gates reuse on exact `(BlockPoint.Slot, BlockPoint.Hash, TxHash)` equality, a later recurrence of the identical failing block/tx after genuine progress — plausible after a rollback that replays the same historical region again — would inherit an already-elevated `Attempts` count instead of starting fresh. Same budget-defeated failure class as #3318, just in the at-tip path.

Verified rather than trusted:
- Reverted just the one-line fix (`ls.lastAtTipRecovery = nil`) and reran `TestResetAtTipRecoveryDescentClearsSameFailureOnProgress` — it fails without the fix (`Expected nil, but got: &atTipRecoveryAttempt{...Attempts:2}`), confirming it's a real regression test.
- With the fix restored: that test plus `TestResetMithrilBoundaryRejectionsRequiresAppliedTipProgress`, `TestAtTipRecoveryAboveMithrilTrustWindowKeepsRecovering`, `TestReplayRecoveryBelowMithrilTrustBoundaryReachesTerminalState` all pass at `-race -count=5`.
- `go build`, scoped `golangci-lint`/`nilaway` on `./ledger/...` clean. (Full-repo `make lint` surfaces 12 pre-existing issues in unrelated files — `database/lifecycle`, `blobbackup`, `redact.go`, `peergov` — none touch this diff, and CI's own lint check is green, so that's local environment drift, not a regression here.)
- `go test -race ./database/plugin/metadata/sqlite/...`, `./internal/docsparity/...`, `./internal/architecture/...`, `make import-boundaries` all pass.
- Confirmed the sole call site (`ledger/state.go:5543`, post-commit block-apply success path) is unchanged and still passes the correct `pendingTip.Point.Slot`.

`ARCHITECTURE.md` update accurately reflects the new invariant. No findings — LGTM.

PR #3342, opened 2026-08-23
fix(peergov): serialize ledger discovery refreshes

### LINKED ISSUE #3319: peergov: emergency ledger discovery never disarms, running every 59s for a full run with 14933 peer drops against 63 hot promotions
**Version:** `devel (commit 5a54b7f8)`, equal to `origin/main`. Network preview, 22.7-hour run, fresh database, Mithril v2 bootstrap. The node synced to tip and held it throughout, so this is steady-state behavior rather than a node in distress.

### Summary

Ledger peer discovery ran on its emergency cadence for the entire run and never once on its configured one. The peer pool churned 14933 drops against 63 hot promotions and never reached `MinHotPeers`. Because emergency discovery is armed by exactly that shortfall, and has no backoff of its own, the two sustain each other indefinitely.

### Evidence

```
1377  discovered ledger peers rounds in 22.7 hours   (one per 59 seconds)
1377  of 1377 with emergency=true
 567  candidates on every round
```

`defaultLedgerPeerRefreshInterval` is 1 hour (`peergov/peergov.go:39`). No round in the run used it.

Peer outcomes over the same period:

```
14933  outbound: dropping never-connected discovered/public-root peer after failed connection
  520  outbound: removing peer after repeated connection failures
   63  promoted peer to hot
  199  selected new best peer
```

`MinHotPeers` defaults to 10 (`peergov/peergov.go:35`). 63 hot promotions across 22.7 hours is not enough to hold that, so `urgent` stays true and the emergency path keeps firing.

### The loop

1. The 567-candidate ledger relay pool is heavily polluted. 221 hostnames never resolve, and 458 handshake attempts land on peers whose network magic is mainnet, sanchonet, or preprod (#3315).
2. Each emergency round adds up to 20 candidates (`added=20`, `added=11`, `added=7`, `added=6`, `added=3` in successive rounds).
3. Nearly all fail on the first dial and are dropped by `shouldDropNeverConnectedPeerAfterDialFailureLocked` (`peergov/connections.go:566`) with a 30-minute deny.
4. Hot peers never reach the target, so `urgent` remains true.
5. Repeat 59 seconds later.

14933 drops over a 30-minute deny window across 22.7 hours works out to roughly 330 distinct addresses cycling through, which is consistent with the polluted portion of a 567-candidate pool.

### Two things worth separating

The pollution is upstream and partly out of our control: pool operators publish dead hostnames and wrong-network relays, and #2018, #3262, and #3315 each remove a slice of it. Those will shrink the candidate pool and help.

The unbounded emergency response is ours. The comment at `peergov/ledger_discovery.go:87` describes the emergency cadence as being for when the node "is short of connected upstreams", which reads as a transient condition. There is no escalation when it turns out not to be transient. A node on a network whose relay records cannot fill the hot target will hammer discovery every 59 seconds forever, and every round re-resolves the whole candidate list (see the ordering problem in #3262).

### Asks

- Give the emergency path a backoff or a ceiling, so a persistent shortfall degrades toward the normal cadence instead of pinning at 59 seconds.
- Consider whether repeatedly re-adding candidates that were just dropped is useful. A candidate that has failed its first dial N rounds running is not a lead back onto the network, and the 30-minute deny is short relative to a 59-second discovery cycle.
- Decide what the intended steady state is on a network where the usable relay set is smaller than `MinHotPeers`. Right now the node behaves as though it is one round away from recovery, indefinitely.

### Note

Sync was unaffected. The node stayed at tip on a small number of good peers the whole time. This is a resource and log-volume problem, and a peer-diversity one, rather than a liveness problem in this run. It is also the multiplier behind the 105438 DNS warnings in #3262.

wolf31o2:
## Summary

- serialize ledger-peer discovery across the provider query and candidate pass
- restore the prior schedule after canceled, failed, or panicking refreshes
- cover slow-provider overlap and interrupted refresh retries

## Checks

- `go test -race -count=1 ./peergov`
- `go test -race -count=1 ./internal/test/conformance/...`
- `make lint`
- `make import-boundaries docs-parity gorm-check`

Closes #3319

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -3110,12 +3110,14 @@ first place. While the node has fewer chain-selection-eligible upstreams than
 its hot-peer target (`MinHotPeers`), replenishment is treated as urgent: a
 dedicated emergency ticker (default 30s via
 `EmergencyDiscoveryCheckInterval`, far more frequent than the 5-minute
-reconcile) pulls fresh stake-pool relays on a short emergency interval (default
-30s via `EmergencyLedgerPeerRefreshInterval`) instead of the normal hourly
-`LedgerPeerRefreshInterval`, so a run of dead or flaky relays can never wedge
-sync while the ledger still lists plenty of registered relays. Once the node is
-at its hot-peer target the emergency path is a no-op and the normal hourly
-cadence applies.
+reconcile) initially pulls fresh stake-pool relays on a short emergency
+interval (default 30s via `EmergencyLedgerPeerRefreshInterval`) instead of the
+normal hourly `LedgerPeerRefreshInterval`. Persistent completed rounds double
+that interval up to the normal cadence, while recovery resets the next shortage
+to the fast floor. Each provider query and relay-candidate pass is single-flight
+across reconcile and emergency ticks: its generation remains claimed until the
+round completes, errors, is canceled, or panics, so a slow provider cannot
+overlap the next tick or leave an artificial retry delay behind.
 
 Each outbound dial attempt re-resolves a hostname-based peer's address fresh,
 narrows the records to the address families the local host can route to
--- a/peergov/emergency_ledger_discovery_test.go
+++ b/peergov/emergency_ledger_discovery_test.go
@@ -16,15 +16,67 @@ package peergov
 
 import (
 	"bytes"
+	"context"
 	"io"
 	"log/slog"
 	"net"
 	"strconv"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
+	"github.com/blinklabs-io/dingo/internal/test/testutil"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
+
+type countingLedgerPeerProvider struct {
+	calls atomic.Int32
+}
+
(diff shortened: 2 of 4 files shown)


arepala-uml (approved):
Verified against the PR head in an isolated worktree, not just read:

- `go build`, `go vet`, `gofmt -l` on all changed files — clean.
- `go test ./peergov/... -race -count=10` (new + existing tests) and `-count=1` on the full package — all pass.
- `make docs-parity` — passes.
- Traced the `ledgerDiscoveryInFlight`/timestamp CAS interleaving by hand: the generation-tagged claim is held exclusively for the whole round, so no concurrent goroutine can observe a torn update.
- Reverted to the merge-base commit and confirmed the pre-fix bug is real: a canceled round returns without restoring `lastLedgerPeerRefresh`, wedging the retry schedule until the next full interval.
- Confirmed the new tests are coupled to the fix (they fail to build against the pre-fix commit — `ledgerDiscoveryInFlight` doesn't exist there), and that the pre-existing `TestPeerGovernor_DiscoverLedgerPeers_ErrorAllowsRetry` still covers the plain-error-reset path, so the new cancel/panic/slow-provider tests are additive, not a replacement.

No blocking issues found.

PR #3346, opened 2026-08-23
fix(devnet): preserve runner test status

### LINKED ISSUE #3334: devnet: run-tests.sh exits non-zero after a fully passing run
`internal/test/devnet/run-tests.sh` exits 1 when every test passed. Reproduced on 6/6 consecutive runs, `--accelerated`.

## Cause

The script sets `set -euo pipefail` (`run-tests.sh:40`) and installs `trap cleanup EXIT` (`:175`). The trap's last command is unguarded:

```bash
  if [[ -n "${STAKE_KEYS_HOST_DIR:-}" ]]; then
    rm -rf "${STAKE_KEYS_HOST_DIR}"     # :172
  fi
```

`STAKE_KEYS_HOST_DIR` receives files written by the `utxo-keys` container as root. `rm -rf` fails on them with `Permission denied`. Because it is the final command in `cleanup`, its status becomes the function's status, and under `set -e` the EXIT trap's status replaces the script's — so a run whose tests all passed reports failure.

## The inconsistency

Every other teardown command in `cleanup` is already guarded:

| Line | Command | Guarded |
|---|---|---|
| 164 | `docker compose ... logs` | `\|\| true` |
| 167 | `rm -rf "${DEVNET_ARTIFACT_DIR}"` | **no** |
| 170 | `docker compose ... down -v` | `\|\| true` |
| 172 | `rm -rf "${STAKE_KEYS_HOST_DIR}"` | **no** |

Both unguarded commands are `rm -rf` on directories that can contain root-owned files. `:167` sits in an `elif` on the success path and can fail the same way.

## Impact

Teardown failure is indistinguishable from test failure at the exit code, which is the signal CI and callers key on. It also masks the reverse: a genuine failure whose exit code is overwritten by a *successful* cleanup path.

## Suggested direction

Guard both `rm -rf` calls the way the sibling commands already are, or remove the tree through a container running as the owning uid so the removal succeeds rather than being ignored. The second is preferable if the leftover directory would otherwise accumulate across runs.

wolf31o2:
## Summary

- Preserve the DevNet test result through best-effort EXIT cleanup.
- Copy temporary stake keys as the host uid/gid so runner-owned files can be removed.
- Exercise the production runner with daemon-free success, failure, cleanup-failure, and `--keep-up` regressions.

## Validation

- Fail-before: cleanup replaced test exits `0` and `23` with `42`; root-like stake trees were left behind and replaced both outcomes with `1`.
- Fixed: repeated race-enabled runner regressions preserve both statuses, remove both stake trees, and retain only failure artifacts.
- Full race-enabled DevNet and conformance package tests, repository lint, shell checks, and documentation checks pass.

Closes #3334

### DIFF
--- a/internal/test/devnet/README.md
+++ b/internal/test/devnet/README.md
@@ -368,6 +368,10 @@ When a run fails, `run-tests.sh` keeps `DEVNET_ARTIFACT_DIR` (a temp
 directory it creates, or one you set) and prints its path instead of
 deleting it. On success the directory is removed.
 
+Teardown is best-effort and never replaces the test exit status. In dingo
+mode, the temporary stake-key copy is written with the host uid/gid so the
+runner can remove its own temporary tree after either result.
+
 | Path | Contents |
 |------|----------|
 | `network/container-status.txt` | `docker compose ps --all` for the profile |
--- a/internal/test/devnet/run-tests.sh
+++ b/internal/test/devnet/run-tests.sh
@@ -154,10 +154,15 @@ collect_failure_artifacts() {
 
 cleanup() {
   local exit_code=$?
+  # Cleanup is best-effort and must never replace the test result. Disable the
+  # trap before the explicit exit below, and disable errexit so one failed
+  # cleanup step cannot skip the rest of the teardown.
+  trap - EXIT
+  set +e
   if [[ "${KEEP_UP}" == "true" ]] && [[ ${exit_code} -eq 0 ]]; then
     log "Tests passed. DevNet left running (--keep-up)."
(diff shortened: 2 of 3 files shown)


wolf31o2:
CI triage: the full Go job first failed the unchanged SQLite cardinality timing sweep; its exact tagged test then passed 20/20 locally. The failed-job-only rerun passed that path but exposed an unrelated late-resync interleaving in the unchanged ledger trust-window regression; its exact tagged test passed 100/100 locally and the flake is tracked in #3348. PR #3346's DevNet package passed in both CI attempts.

arepala-uml (approved):

Issue #2123, opened 2026-04-29
Tracking: Midnight indexer for Acropolis MidnightState gRPC compatibility


wolf31o2:
Tracking issue for adding a Midnight indexer to Dingo that exposes a gRPC API wire-compatible with the Acropolis MidnightState service. Clients generated from the Acropolis proto must work against Dingo unchanged.

Architecture:
- New midnight package containing the indexer and gRPC server.
- EventBus subscriber on ChainBlockEvent and ChainRollbackEvent drives indexing.
- Seven new tables (midnight_asset_creates, midnight_asset_spends, midnight_registrations, midnight_deregistrations, midnight_governance_datums, midnight_ariadne_params, midnight_epoch_candidates) in the existing metadata store.
- Standard google.golang.org/grpc server on port 50051 by default, exposing the 14 RPCs of the MidnightState service. Standard gRPC, not ConnectRPC, so the wire format matches Acropolis's tonic byte-for-byte.
- Enabled only in API storage mode. The midnight config section is required to turn it on. Address and policy fields are required when present, with shipped defaults for known networks.

Sub-issues:
- #2111 Vendor Acropolis midnight_state.proto and add gRPC code generation
- #2112 Add GORM models and migrations for Midnight indexer tables
- #2113 Add midnight config section with YAML, env, and CLI parsing
- #2114 Implement Midnight indexer block scanning for cNIGHT and registration events
- #2115 Implement Midnight indexer governance, Ariadne, and candidate scanning
- #2116 Implement Midnight indexer rollback handling
- #2117 Implement Midnight gRPC server lifecycle
- #2118 Implement paginated UTxO query RPCs for MidnightState
- #2119 Implement governance, parameters, block, epoch, and stability RPCs for MidnightState
- #2120 Wire MidnightIndexer into node.go startup
- #2121 Add Prometheus metrics for the Midnight indexer
- #2122 Backfill Midnight indexer state from Mithril snapshots

wolf31o2:
[...]

PR #3357, opened 2026-08-23
fix(midnight): require explicit gRPC exposure

### LINKED ISSUE #3309: midnight: make gRPC exposure explicitly configurable
## Problem

- Midnight defaults to `enabled: false`, `host: 0.0.0.0`, and `port: 50051`.
- In API storage mode, the gRPC server starts whenever the port is non-zero.
- Server startup does not consult `midnight.enabled`.
- TLS is optional, authentication is absent, and reflection is always enabled.
- API-mode defaults therefore expose the listener without an explicit Midnight enablement decision.

## Acceptance criteria

- Gate the listener behind an explicit server-enablement setting.
- Define and enforce the policy for wildcard plaintext listeners.
- Make reflection exposure configurable or explicitly documented.
- Keep indexer and server enablement semantics unambiguous.
- Add enabled, disabled, loopback, and remote-bind tests.
- Reconcile the behavior with #2123.

wolf31o2:
## Summary

- require explicit Midnight gRPC server, reflection, and insecure-remote opt-ins
- default the listener to loopback and allow remote exposure only with TLS or an explicit plaintext override
- keep indexing and serving independent across startup, reinitialization, configuration, validation, logging, tests, and docs

## Validation

- `make test`
- `make lint`
- `make govulncheck`
- `make import-boundaries docs-parity gorm-check`
- `go test -tags dingo_extra_plugins -count=1 ./internal/test/conformance`

Closes #3309
Relates to #2123


internal/config/validate.go
```diff
@@ -658,6 +660,33 @@ func (c *Config) validate(effectiveMode RunMode, minBindable uint) error {
...
+		))
+	}
+	// Plaintext is safe by default only on loopback. Wildcard, unspecified,
+	// hostname, and concrete remote addresses require either the configured
+	// TLS keypair or an explicit escape hatch acknowledging that transport
+	// security is supplied outside Dingo.
+	useMidnightTLS := c.TlsCertFilePath != "" && c.TlsKeyFilePath != ""
+	if midnightServer && !useMidnightTLS &&
+		!c.Midnight.AllowInsecureRemote &&
+		!isLoopbackAddr(c.Midnight.Host) {
```
cryptodj413, line 683: Merge blocker: `isLoopbackAddr("")` here disagrees with `midnight/server.New` on the empty-host case, and the mismatch fails startup for a valid config while steering the operator toward disabling the safety check it's meant to enforce.

`midnight/server/server.go` defaults an empty `Config.Host` to `127.0.0.1` *before* its own loopback check (lines 164-166, then 179). This validator checks `c.Midnight.Host` directly, so an explicitly-empty value (`DINGO_MIDNIGHT_HOST=` set-but-empty in an env file/systemd unit, or YAML `midnight: {host: ""}`) is treated as non-loopback here even though the server would have bound to loopback by default.

Reproduced with `LoadConfig -> ApplyDefaults -> Validate(RunModeServe)` (the real order `cmd/dingo/main.go` uses), `DINGO_MIDNIGHT_SERVER_ENABLED=true`, `DINGO_MIDNIGHT_HOST=`:
```
err=midnight.host "" is not loopback: configure TLS or set midnight.allowInsecureRemote to acknowledge plaintext exposure
```
The node refuses to start, and the message names the insecure escape hatch as the fix for a config whose effective bind was always going to be safe.

`ApplyDefaults` already has the right precedent for this a couple hundred lines up (`internal/config/config.go`): "Never let an empty debug address inherit the public wildcard bind... keeps manually constructed Config values fail-safe." Suggest normalizing `Midnight.Host == ""` to the loopback default in `ApplyDefaults` (sharing a constant with `DefaultMidnightConfig`/`defaultHost`) so this validator and `midnight/server.New` agree on the same input.

cryptodj413:
Follow-up (not blocking this PR, tracked in a different repository): `blinklabs-io/docs-site` has six guide files (`002-quick-start-overview.md` and `003-create-start-up-service.md`, each in en/es/ja) showing an example `dingo.yaml` with `storageMode: "api"` and a `midnight:` block containing only `authTokenPolicyId`, captioned as sufficient for Midnight indexing:

- `src/content/docs/guides/dingo/002-quick-start-overview.md`
- `src/content/docs/guides/dingo/003-create-start-up-service.md`
- `es/guides/dingo/002-quick-start-overview.md`, `es/guides/dingo/003-create-start-up-service.md`
- `ja/guides/dingo/002-quick-start-overview.md`, `ja/guides/dingo/003-create-start-up-service.md`

After this PR merges, that snippet starts no Midnight gRPC server at all (`serverEnabled` is required and missing from the example). This is outside `dingo`'s own `make docs-parity` check, which only compares this repo's README/ARCHITECTURE against `dingo.yaml.example`. Filing as a note here so it's not lost; the actual doc update belongs in `docs-site`.

internal/config/validate.go
wolf31o2: Fixed in dcc2ca25. ApplyDefaults now normalizes an empty Midnight host to the same loopback value returned by DefaultMidnightConfig before validation. The regression exercises the real LoadConfig -> ApplyFlags -> ApplyDefaults -> validation pipeline with DINGO_MIDNIGHT_SERVER_ENABLED=true and an explicitly empty DINGO_MIDNIGHT_HOST. Focused race tests, vet, lint, import-boundary checks, docs parity, and the ORM guard pass.

cryptodj413 (approved):

PR #3358, opened 2026-08-23
fix(entrypoint): forward bootstrap signals

### LINKED ISSUE #3290: entrypoint: forward signals during Mithril bootstrap
## Problem

- `bin/entrypoint.sh` runs `bootstrap_if_needed` before installing signal traps.
- Mithril bootstrap runs as a foreground child during that interval.
- SIGINT or SIGTERM is not forwarded through the entrypoint's managed shutdown path.

## Acceptance criteria

- Install signal handling before bootstrap starts.
- Track and forward signals to the bootstrap child.
- Preserve the child exit status.
- Add a bounded signal-forwarding test for first-run and resumed bootstrap.

wolf31o2:
Fixes #3290

- install SIGINT and SIGTERM handling before first-run or resumed Mithril bootstrap
- forward the original signal to the active bootstrap or serve child and preserve its exit status
- cover bounded bootstrap forwarding and the bootstrap-to-serve handoff

### DIFF
--- a/bin/entrypoint.sh
+++ b/bin/entrypoint.sh
@@ -66,6 +66,94 @@ warn() { echo "[entrypoint] WARNING: $*" >&2; }
 die()  { echo "[entrypoint] ERROR: $*" >&2; exit 1; }
 
 # --------------------------------------------------------------------------- #
+# Managed child and signal handling
+# --------------------------------------------------------------------------- #
+
+MANAGED_CHILD_PID=""
+PENDING_SIGNAL=""
+PENDING_SIGNAL_EXIT_CODE=""
+SIGNAL_FORWARDING="false"
+MANAGED_WAIT_INTERRUPTED="false"
+
+forward_signal() {
+  local signal_name="$1"
+  local signal_exit_code="$2"
+
+  # A trap can run between starting a background child and assigning $!.
+  # Retain that signal so run_managed_dingo either exits before launch or
+  # forwards it immediately after publishing the child PID.
+  if [[ -z "${MANAGED_CHILD_PID}" ]]; then
+    PENDING_SIGNAL="${signal_name}"
+    PENDING_SIGNAL_EXIT_CODE="${signal_exit_code}"
+    return 0
+  fi
+
+  # A second signal can interrupt the wait in the first handler. Forward it
+  # too, then let the first handler re-wait for the child's actual status.
+  if [[ "${SIGNAL_FORWARDING}" == "true" ]]; then
+    if kill -s "${signal_name}" "${MANAGED_CHILD_PID}" 2>/dev/null; then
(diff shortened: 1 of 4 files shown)


bin/entrypoint.sh
```diff
@@ -65,6 +65,94 @@ log()  { echo "[entrypoint] $*"; }
...
+forward_signal() {
+  local signal_name="$1"
+  local signal_exit_code="$2"
+
+  # A trap can run between starting a background child and assigning $!.
+  # Retain that signal so run_managed_dingo either exits before launch or
+  # forwards it immediately after publishing the child PID.
+  if [[ -z "${MANAGED_CHILD_PID}" ]]; then
+    PENDING_SIGNAL="${signal_name}"
+    PENDING_SIGNAL_EXIT_CODE="${signal_exit_code}"
```
arepala-uml, line 87: Non-blocking: `PENDING_SIGNAL`/`PENDING_SIGNAL_EXIT_CODE` are set here but never explicitly cleared once consumed. Not a bug — every path that reads them (`run_managed_dingo` line 130, `forward_signal` line 142) exits the process before it would matter — but a one-line comment would save a future reader from re-deriving that.

bin/entrypoint.sh
```diff
@@ -175,10 +263,12 @@ bootstrap_if_needed() {
     fi
     sync_args+=("mithril" "sync")
 
-    dingo "${sync_args[@]}"
-
-    log "Mithril bootstrap complete"
-    return 0
+    if run_managed_dingo "" "${sync_args[@]}"; then
```
arepala-uml, line 266: Non-blocking test-coverage gap: two paths through this signal-handling logic aren't exercised by `entrypoint_test.go`: (1) a signal arriving in the idle gap after bootstrap finishes but before `serve` starts, and (2) a plain unsignaled bootstrap failure propagating its exit code via `set -e` at the unguarded `bootstrap_if_needed` call site (line 312). Both verified correct by inspection; flagging for optional follow-up coverage.

arepala-uml (approved):
**Summary**

Validated the signal-forwarding fix for Mithril bootstrap (#3290) directly against the checkout:
- `shellcheck bin/entrypoint.sh` — clean.
- New tests pass under `go test ./bin/... -race -count=20` — no flakes.
- Confirmed regression-test-discipline: reverted `entrypoint.sh` to pre-fix and re-ran the new tests — both fail as expected, restored after.
- `make docs-parity`, `golangci-lint run ./bin/...`, `golines` — all clean.
- Hand-traced the `forward_signal`/`run_managed_dingo` race handling (signal during child backgrounding, second signal interrupting the forwarding wait, signal after the child was already reaped) — all correct.

No merge-blocking issues found. Two non-blocking notes left inline.

PR #3370, opened 2026-08-24
fix(leios): forward peer votes through relays

### LINKED ISSUE #3288: leios: relays never re-diffuse received votes, so a block producer behind a relay cannot vote
**Version:** reproduced on main `5a54b7f8` (current tip) and on `082d329c`.
Network: musashi 2026w32, `--run-mode leios`.

### Summary

A Dingo relay forwards endorser-block *announcements* to its peers but never
forwards *votes*. Only votes the node emitted itself are queued for diffusion.
A relay is not a block producer, so it emits none — and therefore serves none.

Consequences run both ways for a producer whose upstream is a Dingo relay:

- **Inbound:** it observes EB announcements but receives no votes, so it never
  sees quorum and builds no certificates.
- **Outbound:** its own vote is sent to the relay and stops there. The relay has
  no path to re-diffuse a received vote, so the vote never reaches the network.

The standard block-producer-behind-relay topology therefore cannot vote at all.

### Measured, same host, same pool

One relay and its own block producer, the producer's only peer being that relay:

| | RELAY | BLOCK PRODUCER |
|---|---|---|
| `leios_votes_received_total` | 508,705 | **1** |
| `leios_pipeline_eb_observed_total` | 1,277 | 929 |
| `leios_eb_quorum_reached_total` | 597 | **0** |
| `leios_certificates_built_total` | 597 | **0** |
| active peers | 35 | 1 |

The producer sees announcements (929 EBs observed), so the connection is healthy
and announcement relaying works. Votes alone do not cross it.

A second pool on **current main** (`5a54b7f8`), same topology, is identical:
relay `votes_received_total` 61,824 / producer **1**.

### It is not a broken or half-open connection

The obvious alternative explanation is that the producer's link to its relay was
not carrying client protocols. It was. The producer held a **full-duplex**
connection to the relay -- the connection type over which Leios notify traffic
flows -- and the relay logged
`full-duplex inbound connection, starting client protocols` for it.

Over that working full-duplex link the producer received **83** leiosnotify
messages while the relay, at the same moment, had received **566,834**.

The same producer, given full-duplex connections to public peers instead of
relying on its relay, went to 5,561 leiosnotify messages and 4,932 votes within
minutes. Leios traffic does flow over full-duplex connections normally; what it
does not do is cross a Dingo relay to the node behind it.

### The reference implementation does not behave this way

A cardano-node Leios block producer on the same network, with **one** peer (its
own relay), logs in 30 minutes:

```
1806  LeiosVoteAcquired
1768  LeiosNotify.Remote.Receive.Votes
1768  LeiosNotify.Remote.Send.Votes
3536  MsgLeiosVotes
```

So a single-relay producer receives ~1,768 vote messages there, over the same
LeiosNotify protocol, versus 1 in total for Dingo. This rules out "the producer
needs more peers" — the topology is the same and the reference works.

### Confirming experiment

Giving the Dingo producer direct peers, so it no longer depends on the relay to
forward votes, changes the behaviour immediately:

| | before | after 2 minutes |
|---|---|---|
| `votes_received_total` | 1 | 2,040 |
| votes emitted by this producer | 1 in 24 h | 6 in 2 min |
| `certificates_built_total` | 0 | 1 |
| `eb_quorum_reached_total` | 0 | 1 |

Nothing else changed: same binary, same keys, same pool, same relay still
present as a local root.

### Side-by-side over time

Both producers are Dingo, on the same network, sampled at the same instants. The
only difference is that one was given direct peers and the other still depends
on its relay to forward votes. The second is running **current main**.

| sample | edge-peered producer | relay-only producer (main `5a54b7f8`) |
|---|---|---|
| t+0 min | votes_rx 1,442 | votes_rx 1 |
| t+5 min | votes_rx 2,040 | votes_rx 1 |
| t+10 min | votes_rx 4,854 | votes_rx 1 |
| t+15 min | votes_rx 6,261 | votes_rx 1 |
| certificates built | 4 | 0 |
| votes emitted per 10 min | 22 | 0 |

The relay-only producer's counter does not move at all. Its own relay, sampled at
the same moments, went from 508,705 to 514,867 votes received — so the votes are
arriving at the relay and stopping there.

### Where it is in the code

`ouroboros/leiosnotify.go`. Exactly three things are ever appended to the
diffusion log, and one kind of entry is conspicuously local-only:

```go
// :358   locally forged endorser block
o.leiosEBLog.append(leiosForgedEBEntry{point: &point, size: uint64(len(data))})

// :1420  received EB announcement — relayed, guarded by `if relay`
if relay {
    o.leiosEBLog.append(leiosForgedEBEntry{announcement: append([]byte(nil), raw...)})
}

// :1479  EnqueueLeiosPrototypeVote — LOCALLY EMITTED votes only
o.leiosEBLog.append(leiosForgedEBEntry{vote: &copyVote})
```

`EnqueueLeiosPrototypeVote` is called from exactly one place, the
`VoteEmittedEventType` subscription in `node_leios.go:252` — i.e. only for a
vote this node just produced. Received votes reach `VoteManager.insertVote` and
are tallied locally, but nothing puts them back on the wire.

The log's name is the tell: `leiosForgedEBLog` was built to serve locally
**forged** content. Announcement relaying was later added onto it with the
`if relay` guard at :1421. Vote relaying was not.

### Relationship to existing issues

- **#3055** (closed) was the same class of defect from the other direction —
  locally-forged EBs never served because `MessageType` was left zero. The
  serving path has since been exercised for forged content; re-diffusion of
  *received* votes appears never to have been implemented.
- **#3056** added assertions on the forged vote offer payload, which covers
  locally-emitted votes only.
- **#2630** (open) asks for end-to-end verification of EB forging and serving as
  a block producer. This is adjacent: it would likely have caught this had the
  producer been behind a relay rather than directly peered.

### Suggested direction

Re-diffuse received votes the way received announcements already are — append to
the diffusion log on receipt when acting as a relay, subject to the same
freshness window and equivocation checks already applied to announcements.

### What we can offer

Three pools on the network — two Dingo, one cardano-node as a control — with
continuous per-node Leios metrics, and both topologies (producer-behind-relay
and producer-directly-peered) available for A/B testing. We can run instrumented
builds or verify a patch quickly.

arepala-uml:
## Summary

Dingo accepted and counted Leios prototype votes received from peers but did not queue them for re-diffusion. Votes therefore stopped at Dingo relays, preventing relay-connected block producers from observing quorum and building certificates.

Publish an event when a peer vote is newly accepted and connect that event to the existing LeiosNotify vote diffusion queue.

Duplicate, invalid, and equivocating votes are not re-diffused.

### `ledger/leios/event.go`

- Add the `leios.vote_received` event type.
- Use the event to notify node composition that an accepted peer vote is ready for re-diffusion.

### `ledger/leios/manager.go`

- Capture whether `insertVote` accepted a peer vote.
- Publish `leios.vote_received` only for newly accepted votes.
- Preserve existing validation, deduplication, and equivocation checks.
- Support votes accepted immediately and votes resolved after their ranking-block announcement arrives.

### `ledger/leios/manager_test.go`

- Verify that a newly accepted peer vote publishes the relay event.
- Verify that an early vote is relayed after its ranking-block announcement resolves it.
- Verify that a duplicate vote is not queued for diffusion again.

### `node_leios.go`

- Subscribe to `leios.vote_received`.
- Pass accepted peer votes to `EnqueueLeiosPrototypeVote`.
- Use the same LeiosNotify diffusion path as locally emitted votes.

### `node.go`

- Track the received-vote subscription ID in `Node`.
- Update the database recovery error check to the Go 1.26 `errors.AsType` API.

### `node_lifecycle.go`

- Unsubscribe from `leios.vote_received` during live restore and truncate.
- Wait for any active handler to finish before rebuilding Leios components.
- Prevent duplicate vote-diffusion subscriptions across lifecycle rebuilds.
- Update the database recovery error check to the Go 1.26 `errors.AsType` API.

### `ouroboros/leiosnotify.go`

- Clarify that `EnqueueLeiosPrototypeVote` handles both locally emitted votes and accepted peer votes being re-diffused.

### `ARCHITECTURE.md`

- Document the `leios.vote_received` event.
- Document the peer-vote relay flow.
- Document lifecycle ownership of the additional subscription.

Closes #3288 

---
## Summary by cubic
Relays now forward newly accepted Leios prototype votes from one peer to their other peers over LeiosNotify. Previously, relays accepted and counted peer votes but did not re-diffuse them, which blocked quorum behind a relay; now they publish `leios.vote_received` and enqueue the vote for diffusion.

- `ledger/leios`: publish `leios.vote_received` only when `insertVote` accepts a peer vote (immediate or after announcement resolution); duplicates, invalid votes, and equivocators do not fire.
- `node_leios.go`: subscribe to `leios.vote_received` and call `ouroboros.EnqueueLeiosPrototypeVote`, reusing the `leios.vote_emitted` path.
- `node.go`/`node_lifecycle.go`: track and unsubscribe the new subscription during live restore/truncate to prevent duplicate diffusion; switch DB recovery checks to `errors.AsType[T]`.
- `ouroboros/leiosnotify.go`: clarify `EnqueueLeiosPrototypeVote` handles both local and relayed votes.
- Tests verify requeue on acceptance, requeue after later announcement resolution, and suppression on duplicate.
- `ARCHITECTURE.md`: document the new event, relay flow, and lifecycle ownership.

<sup>Written for commit 24a2f41224872bf0ed8b217fd996a633b4f678ea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3370?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Peer prototype votes are now relayed through the Leios notification system after being accepted.
  * Valid received votes are propagated for Ouroboros processing alongside locally generated votes.

* **Bug Fixes**
  * Duplicate submissions and equivocations are no longer re-broadcast.
  * Vote subscriptions are properly managed during live restore and rebuild cycles, preventing duplicate propagation.

* **Documentation**
  * Updated vote propagation documentation to cover accepted peer votes.


wolf31o2 (approved):

ARCHITECTURE.md
```diff
@@ -1100,6 +1100,7 @@ All event types follow the `subsystem.snake_case_name` convention.
 | `forging.slot_battle` | SlotTracker | Competing blocks at same slot |
 | `leios.eb_quorum` | Leios VoteManager | Endorser block reached stake quorum; certificate built (consumed by the Leios PipelineManager for inclusion eligibility) |
 | `leios.vote_emitted` | Leios VoteManager | Locally signed prototype vote ready for node wiring to enqueue on each peer's LeiosNotify stream |
+| `leios.vote_received` | Leios VoteManager | Newly accepted prototype vote from a peer, re-queued for node wiring to enqueue on each *other* peer's LeiosNotify stream so a relay forwards it instead of stopping it at the connection that delivered it (issue #3288) |
```
wolf31o2, line 1103: **[non-blocking] "each *other* peer" claims an origin exclusion the code does not implement**

`EnqueueLeiosPrototypeVote` takes only the vote (`ouroboros/leiosnotify.go:1476`) and appends it to the single shared `leiosEBLog` that every peer cursor reads. The `leios.vote_received` payload is `VoteEmittedEvent{Vote: vote}`, which carries no `connKey`, so the composition layer has nothing to exclude the originating connection with even in principle. The vote is therefore diffused back to the peer that just sent it.

This is not a correctness problem — the receiving peer's own `insertVote` dedup returns false, so no `leios.vote_received` is published there and the echo terminates in one hop. It is a wasted message per vote per sending peer, and more to the point the sentence sets up a future reader to believe an exclusion exists.

Either drop the emphasis ("on each peer's LeiosNotify stream", matching `leios.vote_emitted`'s row directly above), or thread `connKey` through the event and `EnqueueLeiosPrototypeVote` if the exclusion is actually wanted. The former looks right for a prototype path.

wolf31o2 (approved):

ARCHITECTURE.md
wolf31o2: Do this in another PR, so it doesn't block this one.

wolf31o2: Addressed in #3402. Accepted peer votes now carry the origin connection key and exclude that exact connection from relay, while locally produced votes still broadcast to all peers. Focused race tests and the full Dingo suite pass.

PR #3374, opened 2026-08-24
fix(koiosparity): harden strict preview validation

wolf31o2:
- enact source-era protocol-parameter updates before successor-era transitions
- propagate strict Koios parity failures through node shutdown
- skip pre-staking account-history scans
- keep default Koios account-history request bodies below the endpoint limit


arepala-uml:
## Blocking: post-rollover era transitions retain source-era timing

`splitEraTransitionsForRollover()` now moves every transition after `processEpochRollover()`. This correctly allows pending protocol-parameter updates to enact in the source era, but `processEpochRollover()` also creates the new epoch using that source era’s `EpochLengthFunc`.

`applyBoundaryEraTransitions()` subsequently updates the epoch’s `EraId` and protocol parameters, but it preserves:

- `NewCurrentEpoch.SlotLength`
- `NewCurrentEpoch.LengthInSlots`
- the matching values in `NewEpochCache`
- `SchedulerIntervalMs`

This is hidden on Preview’s Alonzo → Babbage transition because both eras use the same timing. It breaks transitions where timing differs, particularly Byron → Shelley.

The following regression test fails on this PR’s current head:

~~~go
func TestBoundaryEraTransitionUsesTargetEraTiming(t *testing.T) {
	ls, db := newBoundaryRolloverLedger(t)

	// Give the source era deliberately different timing from the target era.
	// The rollover must enact source-era rules first, but the epoch it creates
	// ultimately belongs to the target era and must use the target's timing.
	sourceEra := ls.currentEra
	sourceEra.EpochLengthFunc = func(
		*cardano.CardanoNodeConfig,
	) (uint, uint, error) {
		return 20_000, 21_600, nil
	}
	ls.currentEra = sourceEra

	var result *EpochRolloverResult
	txn := db.Transaction(true)
	require.NoError(t, txn.Do(func(txn *database.Txn) error {
		var err error
		result, err = ls.processEpochRollover(
			txn,
			ls.currentEpoch,
			sourceEra,
			ls.currentPParams,
			true,
		)
		if err != nil {
			return err
		}
		_, err = ls.applyBoundaryEraTransitions(
			txn,
			ls.currentEpoch,
			[]uint{eras.AllegraEraDesc.Id},
			result,
		)
		return err
	}))

	wantSlotLength, wantEpochLength, err :=
		eras.AllegraEraDesc.EpochLengthFunc(ls.config.CardanoNodeConfig)
	require.NoError(t, err)

	assert.Equal(t, wantSlotLength, result.NewCurrentEpoch.SlotLength,
		"the transitioned epoch must use the target era's slot length")
	assert.Equal(t, wantEpochLength, result.NewCurrentEpoch.LengthInSlots,
		"the transitioned epoch must use the target era's epoch length")
	assert.Equal(t, wantSlotLength, result.SchedulerIntervalMs,
		"the scheduler must use the target era's slot length")
}
~~~

Running:

~~~text
go test -run '^TestBoundaryEraTransitionUsesTargetEraTiming$' -count=1 ./ledger
~~~

produces:

~~~text
target slot length:
expected: 1000
actual:   20000

target epoch length:
expected: 432000
actual:   21600

scheduler interval:
expected: 1000
actual:   20000
~~~

### Suggested fix

After applying the transition path, resolve the final era and call its `EpochLengthFunc`. Use those returned values to update:

1. `newEpoch.SlotLength`
2. `newEpoch.LengthInSlots`
3. the persisted epoch passed to `SetEpoch`
4. the corresponding entry in `rolloverResult.NewEpochCache`
5. `rolloverResult.SchedulerIntervalMs`

The rollover should still enact pending updates using the source era, but the resulting new epoch must use the final target era’s timing.

wolf31o2:
Fixed: target-era timing now updates the persisted epoch, rollover result/cache, and scheduler; regression added.

arepala-uml (approved):

PR #3382, opened 2026-08-24
docs(leios): correct peer vote diffusion description

arepala-uml:
## Summary

Correct the Leios vote-relay documentation to match the implemented diffusion behavior.

Received votes are appended to a shared diffusion log read by every peer, including the peer that originally sent the vote. The resulting one-hop echo terminates because the receiving node rejects the vote as a duplicate.

Final leftover part of #3370 

---
## Summary by cubic
Corrects the Leios vote-relay docs to match the implemented diffusion behavior. Previously we documented `leios.vote_received` as re-queued to each other peer; now we state it is re-queued to each peer, including the sender, producing a one-hop echo that terminates via duplicate rejection.

<sup>Written for commit a2a66eddb2af775652b2991da9bcca8fad3c357f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3382?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1164,7 +1164,7 @@ All event types follow the `subsystem.snake_case_name` convention.
 | `forging.slot_battle` | SlotTracker | Competing blocks at same slot |
 | `leios.eb_quorum` | Leios VoteManager | Endorser block reached stake quorum; certificate built (consumed by the Leios PipelineManager for inclusion eligibility) |
 | `leios.vote_emitted` | Leios VoteManager | Locally signed prototype vote ready for node wiring to enqueue on each peer's LeiosNotify stream |
-| `leios.vote_received` | Leios VoteManager | Newly accepted prototype vote from a peer, re-queued for node wiring to enqueue on each *other* peer's LeiosNotify stream so a relay forwards it instead of stopping it at the connection that delivered it (issue #3288) |
+| `leios.vote_received` | Leios VoteManager | Newly accepted prototype vote from a peer, re-queued for node wiring to enqueue on each peer's LeiosNotify stream so a relay forwards it instead of stopping it at the connection that delivered it (issue #3288) |
 | `peergov.outbound_conn` | PeerGov | Outbound connection initiated |
 | `peergov.peer_demoted` | PeerGov | Peer demoted |
 | `peergov.peer_promoted` | PeerGov | Peer promoted |


ARCHITECTURE.md
```diff
@@ -1164,7 +1164,7 @@ All event types follow the `subsystem.snake_case_name` convention.
 | `forging.slot_battle` | SlotTracker | Competing blocks at same slot |
 | `leios.eb_quorum` | Leios VoteManager | Endorser block reached stake quorum; certificate built (consumed by the Leios PipelineManager for inclusion eligibility) |
 | `leios.vote_emitted` | Leios VoteManager | Locally signed prototype vote ready for node wiring to enqueue on each peer's LeiosNotify stream |
-| `leios.vote_received` | Leios VoteManager | Newly accepted prototype vote from a peer, re-queued for node wiring to enqueue on each *other* peer's LeiosNotify stream so a relay forwards it instead of stopping it at the connection that delivered it (issue #3288) |
+| `leios.vote_received` | Leios VoteManager | Newly accepted prototype vote from a peer, re-queued for node wiring to enqueue on each peer's LeiosNotify stream so a relay forwards it instead of stopping it at the connection that delivered it (issue #3288) |
```
wolf31o2, line 1167: This row is now correct, and I re-confirmed the behavior from the code rather than from my own earlier review:

- `leiosForgedEBEntry` (`ouroboros/leiosnotify.go:43`) has no origin field — only `point`, `size`, `vote`, `announcement`.
- `leiosForgedEBLog.cursors` is `map[string]int // connKey -> next logical index to serve`, a per-connection *position*, not a filter.
- The serving callback computes `connKey` and calls `o.leiosEBLog.next(connKey)` (`leiosnotify.go:1457-1461`) with no comparison against the vote's origin.

So no origin exclusion exists on this path and dropping the emphasis is right.

**The same document still asserts the opposite, though**, and this PR does not touch it. `ARCHITECTURE.md:3537`:

> The LeiosVotes server callback must return exactly the requested number of votes, so it blocks (with a per-connection cursor over the append-ordered vote log, **never echoing a vote back to its origin**) until enough votes arrive or the protocol shuts down

That parenthetical came in with #2509, where it was vacuously true: only locally emitted votes entered the log, so there was no origin to echo to. #3370 routed peer-received votes into the same log, which made it false. `origin/main` currently carries both statements, so after this PR the file has one accurate description of the behavior and one that contradicts it — and the surviving one makes the stronger claim.

Since correcting the diffusion description is this PR's whole purpose, worth taking both in one change. Dropping the parenthetical is enough:

> ... so it blocks (with a per-connection cursor over the append-ordered vote log) until enough votes arrive or the protocol shuts down

Or state the actual behavior, if it is worth saying there at all: a peer's own vote is offered back to it and rejected as a duplicate on receipt.

wolf31o2 (changes requested):

ARCHITECTURE.md
wolf31o2: Withdrawn: that paragraph describes the LeiosVotes path, where VoteManager.NextVotes filters originConn. The edited row describes the separate LeiosNotify requeue path.

wolf31o2 (approved):

PR #3457, opened 2026-08-24
fix(ledger): validate phase-two-invalid transactions

### LINKED ISSUE #3454: ledger: enforce invalid_hereafter transaction bounds
## Summary

Dingo accepts transactions whose `invalid_hereafter` slot is before the current
slot because the active Allegra-through-Dijkstra validation path does not
enforce the upper validity bound, including live Conway.

## Acceptance criteria

- Reject transactions when `invalid_hereafter` is less than the validation
  slot.
- Preserve valid transactions whose upper bound is at or after the validation
  slot.
- Cover the active Dingo validation entry point with a regression test that
  fails before the fix.
- Verify all affected eras and document whether `ARCHITECTURE.md` or
  `DATABASE.md` changes are required.

wolf31o2:
## Summary

- Run phase-1 validation for phase-two-invalid transactions while continuing to skip phase-2 script execution.
- Add block-processing regression coverage for a phase-1 validity-interval failure.
- Move validity-interval upper-bound enforcement to blinklabs-io/gouroboros#2103.

## Validation

- Focused regression fail-before/pass-after with the race detector
- Ledger race tests
- `go vet`, golangci-lint, NilAway, modernize, import-boundary, docs-parity, and Conway conformance checks

Related: blinklabs-io/dingo#3454

## Summary by CodeRabbit

* **Bug Fixes**
  * Phase-one transaction validation now runs even when phase-two script validation has already failed.
  * Block processing correctly reports invalid validity intervals and other structural transaction errors in this scenario.
  * Preserved the expected behavior of skipping phase-two script evaluation for transactions marked invalid.

* **Tests**
  * Added regression coverage for phase-one validation of phase-two-invalid transactions.

* **Documentation**
  * Documented the profile-specific validation behavior and its enforced boundary.


arepala-uml (approved):
**Review: no merge blockers.**

Confirmed the fix independently rather than relying on the PR description:

- Reverted just the changed condition in `ledger/state.go` and reran the new regression test — fails on old logic, passes with the fix.
- Read gouroboros v0.202.2 source directly: `dijkstra.UtxoValidatePlutusScripts` is the only rule in the chain gated on `!tx.IsValid()`; all other phase-1 rules already run unconditionally. The fix correctly stops over-skipping validation without disturbing the phase-2 skip.
- gouroboros/plutigo bump is not scope creep: it's where the referenced upstream fix (gouroboros#2103) landed, and the `api/utxorpc/plutusdata_cardano.go` rewrite is a required adaptation to `Constr.Tag` changing from `uint` to `*big.Int` in plutigo 0.4.0, not unrelated code.
- Full `ledger/...` suite under `-race` (388s), `api/utxorpc` race tests, and `internal/test/conformance` all pass on this branch. Build and `go vet` clean.
- `docs-parity` failure confirmed pre-existing on current `origin/main` — unrelated to this change.
- ARCHITECTURE.md update checked against verified gouroboros behavior — accurate. DATABASE.md correctly untouched.

Non-blocking follow-up: the newly-enforced phase-1 checks on phase-2-invalid transactions could reject a transaction that was previously accepted on a live chain (CodeRabbit's own risk note). Validation here covers synthetic fixtures and conformance vectors only — worth a check against real preview/preprod chain history containing phase-2-invalid transactions before/after wider rollout, not a reason to hold this PR.

Issue #3432, opened 2026-08-24
internal/test/devnet asserts a Unix-only UID:GID mapping, failing go-test-other on windows-latest


cryptodj413:
## What

`publish.yml`'s `go-test-other (windows-latest)` job ran the full `./...`
suite on the most recent release tag and failed:

```
--- FAIL: TestRunTestsCleansContainerCreatedTemporaryFiles (2.18s)
    --- FAIL: TestRunTestsCleansContainerCreatedTemporaryFiles/success (0.98s)
        Error: "... run --rm --user 197108:197121 ..." does not contain "run --rm --user -1:-1"
    --- FAIL: TestRunTestsCleansContainerCreatedTemporaryFiles/failure (1.20s)
        Error: "... run --rm --user 197108:197121 ..." does not contain "run --rm --user -1:-1"
FAIL    github.com/blinklabs-io/dingo/internal/test/devnet     7.549s
```

Run: https://github.com/blinklabs-io/dingo/actions/runs/32750171130/job/97423324173
(tag `v0.70.0`, `go-test-other (1.26.x, windows-latest)`, conclusion `failure`;
the `macos-latest` leg of the same job passed).

`TestRunTestsCleansContainerCreatedTemporaryFiles` asserts the container-cleanup
command line contains `--user -1:-1`, a Unix UID:GID sentinel with no Windows
equivalent — `os.Getuid()`/`os.Getgid()` both return `-1` on Windows, so the
literal the test hardcodes never appears there; the real (non-sentinel) values
the code substitutes on that platform are what actually show up.

## Why this matters beyond one test

`go-test-other`'s `windows-latest` leg only runs at release-tag time, after a
change has already merged (see `.github/workflows/publish.yml`). There is no
Windows CI on pull requests today, so a regression like this — or a genuine
platform-specific bug — surfaces only when it blocks a release, not when the
change that caused it is still open for review. blinklabs-io/dingo#3228 added
a narrowly-scoped `go-test-windows` job on `.github/workflows/go-test.yml`
(pull_request-triggered, limited to `./mithril/...`) specifically to avoid
blocking on this and other unaudited pre-existing Windows gaps; this issue
tracks widening real Windows PR coverage, which needs failures like this one
fixed first.

## Fix

`TestRunTestsCleansContainerCreatedTemporaryFiles` (and any sibling assertion
in `internal/test/devnet` doing the same thing) needs a
`runtime.GOOS == "windows"` branch, or a portable sentinel computed the same
[...]

PR #3460, opened 2026-08-24
fix(mithril): rename and remove Windows extraction targets by handle

### LINKED ISSUE #3228: Windows extraction publish and unlink resolve their parent by path, not by handle

cryptodj413:
Fixes #3228.

## What

`MoveFile`, `DeleteFile` and `RemoveDirectory` each address their target by
path, which is a second, independent resolution of names `openVerifiedParent`
already walked and verified through parent handles. Since `os.Root` opens
every handle with `FILE_SHARE_DELETE`, any verified component substituted
between that walk finishing and the Windows API's own resolution running was
followed, not rejected — the walk only catches a symlink or reparse point
already in place when it runs, not one planted afterward.

This gives the three Windows extraction operations (rename, file removal,
directory removal) a genuinely handle-relative path instead, using
`NtCreateFile`/`NtSetInformationFile` from `golang.org/x/sys/windows` (which
exposes both functions directly, but not the
`FILE_RENAME_INFORMATION`/`FILE_DISPOSITION_INFORMATION` request buffers they
take — built by hand in `extract_handlerelative_windows.go`, with the offset
computation documented against the trap of using `unsafe.Sizeof` on a struct
with a variable-length trailing field).

`removeEmptyExtractDir` previously ignored its `root`/`name` parameters
entirely and resolved `fullPath` directly; it now takes the same
handle-relative path as the other two, closing a second, independent
resolution that existed there as well.

## Tests

- Per-operation substitution tests: verify a component, swap it on disk
  exactly as a writer with access to the parent could, then perform the
  operation — reproducing the gap between verification and the act that a
  single, uninterrupted call can't exercise from outside. Per the acceptance
  criteria, this is the case the walk alone cannot catch.
- Buffer-layout tests for the hand-built `FILE_RENAME_INFORMATION` encoding,
  decoded independently byte-by-byte rather than by casting back through the
  same struct.

## Verification

I have no Windows runtime available to me, so this is cross-compiled and
vetted clean for `windows/amd64` and `windows/arm64` (`go build`, `go vet`,
`golangci-lint run`, `nilaway`, `modernize`, `go test -c`), but that is
exactly the limit the issue calls out — "not cross-compilation alone." A
`go-test-windows` job (scoped to `./mithril/...`, see its commit for why) is
added alongside this so the fix gets real `windows-latest` execution as part
of this PR's own checks, rather than only at release-tag time. **Leaving this
as a draft until that job has actually run and passed** — will mark ready for
review once it's green, or fix and re-push if it isn't.

## Side finding

While confirming Windows CI actually works today, I found the release-only
`go-test-other (windows-latest)` job currently fails on the latest tag for an
unrelated, pre-existing reason (a Unix-only UID:GID assertion in
`internal/test/devnet`). Filed as #3432 rather than folding into this PR's
scope.

## Docs

`ARCHITECTURE.md` updated: the "Handles bind directories" section and the
publish-destination section both described the now-fixed Windows limitation
(including a "Windows refuses to move a directory while handles are open
beneath it" claim that issue #3228 itself flagged as false). `DATABASE.md` is
unaffected — no schema, query, or storage-plugin surface changes here.

---
## Summary by cubic
Renames and deletions on Windows now operate by handle instead of path, removing the second name resolution that allowed post-verification substitutions to redirect operations. Previously used `MoveFile`/`DeleteFile`/`RemoveDirectory`; now we open the leaf relative to the verified parent and apply rename/deletion to that handle, refusing leaf reparse points. Unlink remains file-only; rmdir remains directory-only.

- Windows: add handle-relative helpers using `golang.org/x/sys/windows` (`NtCreateFile`/`NtSetInformationFile`) with `FILE_OPEN_REPARSE_POINT`, `FILE_OPEN_FOR_BACKUP_INTENT`, and required access flags (`DELETE`, `FILE_READ_ATTRIBUTES`, plus `SYNCHRONIZE` for rename). Convert `NTSTATUS` to `syscall.Errno`. `removeEmptyExtractDir` now uses the same handle-relative path; errors return as `PathError`/`LinkError` naming the original path.
- Tests: add substitution tests for rename/unlink/rmdir and a byte-level layout test for the built `FILE_RENAME_INFORMATION` buffer.
- CI: add `go-test-windows` on `windows-latest` running `go test ./mithril/...` to exercise these changes on real Windows in PRs.
- Docs: update ARCHITECTURE to document handle-relative rename/unlink on Windows.

<sup>Written for commit 227d97d0d864f5ab0b7c8f4eca5636239db5fabe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3460?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security**
  * Improved Windows extraction safety by performing file, directory, rename, and cleanup operations relative to verified directory handles.
  * Prevented path redirection and symbolic-link substitution during extraction cleanup.

* **Bug Fixes**
  * Ensured operations continue to target the originally verified extraction location even if directory contents change.

* **Tests**
  * Added Windows coverage for deletion, directory removal, renaming, and operation-layout validation.

* **CI**
  * Added Windows pull request checks that build the CLI and run the Go test suite.

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -131,6 +131,46 @@ jobs:
         # validation runs in the publish workflow before draft creation.
         run: go test -tags dingo_extra_plugins -ldflags="-s -w" -timeout 20m ./...
 
+  # windows-latest also runs the full ./... suite in publish.yml's
+  # go-test-other, but only at release-tag time, after a change has already
+  # merged -- which is why the Windows-specific extraction code this job
+  # targets (mithril's NtCreateFile/NtSetInformationFile handle-relative
+  # rename and removal, see extract_handlerelative_windows.go and issue
+  # #3228) went unverified on real Windows CI before now. This job runs it on
+  # every PR instead.
+  #
+  # It stays scoped to ./mithril/... rather than ./..., deliberately: the
+  # release job's full-suite run currently fails on windows-latest for
+  # unrelated reasons (internal/test/devnet asserts a Unix-only UID:GID
+  # mapping; tracked separately), and running the same full suite here would
+  # block every PR on that pre-existing, unrelated breakage rather than on
+  # anything this job is meant to check. Widening this job's scope is
+  # follow-up work gated on auditing and fixing the rest of the suite for
+  # Windows, not something to do incidentally here.
+  go-test-windows:
+    name: go-test (Windows)
+    strategy:
+      matrix:
+        go-version: [1.26.x]
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1 https://github.com/actions/checkout/releases/tag/v7.0.1
+      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0 https://github.com/actions/setup-go/releases/tag/v7.0.0
+        with:
+          go-version: ${{ matrix.go-version }}
+      - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0 https://github.com/actions/cache/releases/tag/v6.1.0
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~\AppData\Local\go-build
+          key: ${{ runner.os }}-go-${{ matrix.go-version }}-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-${{ matrix.go-version }}-
+      - name: go-build
+        run: go build ./cmd/dingo
+      - name: go-test
+        run: go test -timeout 10m ./mithril/...
+
   govulncheck:
     name: govulncheck
     runs-on: ubuntu-latest
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -3782,14 +3782,16 @@ Destinations come in two shapes, selected by the caller:
   it unlinks a regular file as readily as it removes a directory, which is
   exactly the behaviour that made such a swap costly.
 
-  Both platforms have a directory-only primitive, reached differently. Unix
-  uses `unlinkat` with `AT_REMOVEDIR`, addressed through the parent handle, so
-  neither the entry nor the parent can be redirected. Windows uses
-  `RemoveDirectory`, which fails on a file and on a populated directory but
-  addresses the entry by name, because Windows has no handle-relative removal;
-  a substituted parent could redirect it, which Windows makes hard by refusing
-  to move a directory while handles are open beneath it, and the parent is held
-  open for the whole extraction.
+  Both platforms have a directory-only primitive, addressed through the parent
+  handle so neither the entry nor the parent can be redirected. Unix uses
+  `unlinkat` with `AT_REMOVEDIR`. Windows has no directory-only removal in the
+  standard library, but `NtSetInformationFile`'s `FileDispositionInformation`
+  applied to a handle opened relative to the verified parent gets the same
+  guarantee: the entry is opened as a single component under that handle,
+  NTFS enforces emptiness the same way it does for `RemoveDirectory`, and
+  nothing here resolves the parent's name again the way the older
+  `RemoveDirectory`-by-path implementation did (`extract_handlerelative_windows.go`;
+  issue #3228).
 
   `WithReplaceDestination` remains the only path that removes a destination
   holding content, which is what that option exists to authorise.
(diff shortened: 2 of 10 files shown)


wolf31o2 (approved):
Approving.

The core design is right: `NtCreateFile` with `OBJECT_ATTRIBUTES.RootDirectory` plus `NtSetInformationFile` is the only way to get `unlinkat`/`renameat` semantics on Windows, and all three operations now address their target through the verified parent's handle instead of handing the kernel a path to resolve a second time.

Checks I ran, and the ones that matter most given no Windows runtime was available to the author:

- **`GOOS=windows golangci-lint run ./mithril/`: 0 issues.** Worth calling out - the repository `lint` job runs on Linux, where every `//go:build windows` file in this change is excluded from analysis entirely. The bulk of the diff is otherwise unlinted.
- Cross-compiled and vetted clean for `windows/amd64` and `windows/arm64`; `go test ./mithril/` still green on Linux.
- CI's new `go-test (Windows) (1.26.x)` job passes, so the three substitution tests and the buffer-layout tests genuinely executed on Windows rather than being reasoned about. That job earning its keep is already visible in the history: 997e2239 and 227d97d0 add `FILE_OPEN_FOR_BACKUP_INTENT` and `FILE_READ_ATTRIBUTES`, which are exactly the access bits a real run would surface and a cross-compile never would.

Three things I checked specifically rather than took on the comments:

**The `FILE_RENAME_INFORMATION` offset is correct, and the layout test is not circular.** `unsafe.Offsetof(FileNameLength) + unsafe.Sizeof(FileNameLength)` gives 20 on a 64-bit target where `unsafe.Sizeof(header)` gives 24 - the padding trap the comment describes. The test decodes the buffer with `binary.LittleEndian` and manual shifting and derives its own offsets from `unsafe.Sizeof(windows.Handle(0))`, so it is an independent derivation that happens to agree, not a round-trip through the struct under test.

**`removeEmptyExtractDir`'s new invariant holds at the source.** It no longer walks with `openVerifiedParent`, so it now depends on `name` being a single component - a multi-component name would let `NtCreateFile` resolve the intermediates and reintroduce exactly this issue. `destName` comes from `filepath.Base(cleanDest)` at `extract_hardening.go:441`, so the invariant is guaranteed by construction rather than by convention.

**`FILE_DIRECTORY_FILE` on the rename source is safe.** `renameExtractedDirectory`'s only callers (`extract_hardening.go:562`, `581`) pass `stagingName`, always a directory; a non-directory would fail the open rather than do something unintended.

`ARCHITECTURE.md` is updated correctly - both passages that previously documented the residual race as open are rewritten, rather than left contradicting the code.

The scoping note on the new CI job is well-judged: keeping it at `./mithril/...` rather than `./...` because the full suite fails on `windows-latest` for an unrelated Unix-only UID:GID assertion is the right call, and the comment records why instead of leaving it to be rediscovered.

No findings.

PR #3462, opened 2026-08-24
fix(database): validate immutable index metadata

### LINKED ISSUE #3447: fix(database): reject malformed immutable indexes
Malformed immutable database index metadata can reach allocation logic and panic with `makeslice: len out of range`. The primary index accepts an unknown format version, misaligned secondary offsets can become garbage block offsets, and descending offsets are silently treated as empty slots. Both `chunk.Next` branches can panic when reached through `immutable.New` and `GetBlock`.

Acceptance criteria:

- Reject unsupported primary-index versions when opening the index.
- Reject secondary offsets that are misaligned or non-monotonic before using them.
- Return explicit errors from both affected chunk iteration branches instead of panicking.
- Add regression tests using malformed index fixtures and prove they fail against the vulnerable implementation.

wolf31o2:
## Summary

- Validate immutable primary and secondary index versions, record alignment,
  and offset ordering.
- Return errors for invalid chunk ranges before seeking or allocating.
- Preserve valid single-block and multi-block immutable data.
- Update the Mithril bootstrap fixture to the supported primary-index version.

Closes #3447

## Review summary (Cubic)

- Unsupported versions and malformed index offsets now fail explicitly.
- Out-of-spec immutable files must regenerate their index metadata.

### DIFF
--- a/database/immutable/chunk.go
+++ b/database/immutable/chunk.go
@@ -80,11 +80,24 @@ func (c *chunk) Next() (*Block, error) {
 		// This triggers even though we check it above
 		// #nosec G115
 		currOffset := int64(c.currentEntry.BlockOffset)
+		if currOffset > c.fileSize {
+			return nil, fmt.Errorf(
+				"current block offset %d is beyond chunk size %d",
+				currOffset,
+				c.fileSize,
+			)
+		}
 
 		// We've reached the last entry in the chunk, so we calculate
 		// block size based on the size of the file
 		blockSize := c.fileSize - currOffset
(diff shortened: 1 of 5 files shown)


wolf31o2:
CodeRabbit was rate-limited and Cubic did not post a review for this head. Independent review found no actionable findings. Immutable and Mithril race suites, conformance checks, lint, vet, nilaway, and modernize passed locally; CI is green at `3247033ee6462d44e02557c07aea2e34124f943d`.

chrisguiney (approved):

PR #3473, opened 2026-08-24
fix(mithril): require verified genesis anchor

### LINKED ISSUE #3465: mithril: require a genesis trust anchor for verified bootstrap
When `VerifyCertificateChain` is enabled, bootstrap accepts an empty `GenesisVerificationKey`. Both backends then use structural certificate-chain validation and skip the pinned genesis signature check; v2 still presents this as verified bootstrap because only its ancillary key is required.

Acceptance criteria:

- Reject verified bootstrap before downloads when the Mithril genesis verification key is missing or empty.
- Use STM certificate verification and validate the genesis certificate signature whenever certificate-chain verification is enabled.
- Apply the invariant to both v1 and v2 bootstrap paths.
- Preserve explicitly unverified bootstrap behavior when `VerifyCertificateChain` is false.
- Cover missing-key rejection and verified/unverified behavior with regression tests.

wolf31o2:
Closes #3465.

- Require a pinned genesis trust anchor before verified bootstrap contacts the aggregator.
- Enforce STM chain and genesis-signature verification for both artifact backends.
- Cover verified and explicitly unverified behavior with real cryptographic fixtures.

Checks:

- Focused Mithril race regressions
- Full race-enabled test suite
- Dingo conformance vectors (315/315)

### DIFF
--- a/mithril/bootstrap.go
+++ b/mithril/bootstrap.go
@@ -121,14 +121,13 @@ type BootstrapConfig struct {
 	// CleanupAfterLoad controls whether temporary files are removed
 	// after loading completes.
 	CleanupAfterLoad bool
-	// VerifyCertificateChain enables certificate chain verification
-	// against the aggregator. When true, the bootstrap process
-	// walks the certificate chain from the snapshot back to the
-	// genesis certificate to verify the chain is unbroken.
+	// VerifyCertificateChain enables STM certificate verification against the
+	// aggregator. When true, GenesisVerificationKey is required and the
+	// bootstrap process verifies the chain back to that pinned genesis key.
 	VerifyCertificateChain bool
-	// GenesisVerificationKey is the Mithril genesis verification key loaded
-	// from Cardano network config. It is validated for parseability now and
-	// will be used by full STM verification.
+	// GenesisVerificationKey is the pinned Mithril trust anchor used by STM
+	// certificate verification. It is loaded from Cardano network config and
+	// validated for parseability before any aggregator request.
 	GenesisVerificationKey string
 	// AncillaryVerificationKey is the Mithril ancillary verification key loaded
 	// from Cardano network config. It is validated for parseability now and
@@ -342,13 +341,16 @@ func Bootstrap(
 	}
 	cfg.Backend = normalizeBackend(cfg.Backend)
 	if cfg.VerifyCertificateChain {
-		if cfg.GenesisVerificationKey != "" {
-			if _, err := ParseVerificationKey(cfg.GenesisVerificationKey); err != nil {
-				return nil, fmt.Errorf(
-					"parsing Mithril genesis verification key: %w",
-					err,
-				)
-			}
+		if strings.TrimSpace(cfg.GenesisVerificationKey) == "" {
+			return nil, errors.New(
+				"verified Mithril bootstrap requires a genesis verification key",
+			)
+		}
+		if _, err := ParseVerificationKey(cfg.GenesisVerificationKey); err != nil {
+			return nil, fmt.Errorf(
+				"parsing Mithril genesis verification key: %w",
+				err,
+			)
 		}
 		if cfg.AncillaryVerificationKey != "" {
 			if _, err := ParseVerificationKey(cfg.AncillaryVerificationKey); err != nil {
@@ -469,16 +471,12 @@ func Bootstrap(
 			"component", "mithril",
 			"certificate_hash", snapshot.CertificateHash,
 		)
-		verificationMode := VerificationModeStructural
-		if cfg.GenesisVerificationKey != "" {
-			verificationMode = VerificationModeSTM
-		}
 		verificationResult, err := VerifyCertificateChainWithMode(
 			ctx,
 			client,
 			snapshot.CertificateHash,
 			snapshot.Digest,
-			verificationMode,
+			VerificationModeSTM,
 		)
 		if err != nil {
 			return nil, fmt.Errorf(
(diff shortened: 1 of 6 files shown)


mithril/bootstrap_v2.go
```diff
@@ -443,39 +443,33 @@ func verifyArtifactCertificateV2(
...
-		if err := VerifyGenesisCertificateSignature(
-			verificationResult.GenesisCertificate,
-			cfg.GenesisVerificationKey,
-		); err != nil {
-			return fmt.Errorf(
-				"genesis certificate verification failed: %w",
-				err,
-			)
-		}
+	if verificationResult == nil ||
```
chrisguiney, line 459: Small cleanup opportunity, entirely optional: this block and the one at `mithril/bootstrap.go:487-500` are now byte-identical, and each is immediately preceded by the same `VerifyCertificateChainWithMode(..., VerificationModeSTM)` call.

The duplication predates the PR, but dropping the `if cfg.GenesisVerificationKey != ""` wrapper is what made the two converge exactly — which makes this the cheapest moment to pull out something like `verifyChainToGenesis(ctx, client, hash, digest, genesisKey)`. The upside isn't line count so much as making it structurally impossible for the v1 and v2 trust paths to drift apart later.

mithril/bootstrap.go
```diff
@@ -469,39 +471,33 @@ func Bootstrap(
...
-		if cfg.GenesisVerificationKey != "" {
-			verificationMode = VerificationModeSTM
-		}
 		verificationResult, err := VerifyCertificateChainWithMode(
 			ctx,
 			client,
 			snapshot.CertificateHash,
 			snapshot.Digest,
-			verificationMode,
+			VerificationModeSTM,
```
chrisguiney, line 479: Now that both call sites pass `VerificationModeSTM` unconditionally, the structural-only path has no production callers left: `VerifyCertificateChain` (`mithril/bootstrap.go:1125`) is reached only from `bootstrap_test.go:1082` and `integration_test.go:66`.

Not a problem in itself, but it's an exported helper whose doc says "It does not verify STM cryptographic signatures (Phase 2)" — which is both a footgun for exactly the weakness this issue closed and now stale, since STM verification is no longer a future phase. Might be worth a deprecation marker, or just a line warning that it must not be used for trust decisions.

mithril/bootstrap.go
```diff
@@ -342,13 +341,16 @@ func Bootstrap(
...
-		if cfg.GenesisVerificationKey != "" {
-			if _, err := ParseVerificationKey(cfg.GenesisVerificationKey); err != nil {
-				return nil, fmt.Errorf(
-					"parsing Mithril genesis verification key: %w",
-					err,
-				)
-			}
+		if strings.TrimSpace(cfg.GenesisVerificationKey) == "" {
+			return nil, errors.New(
+				"verified Mithril bootstrap requires a genesis verification key",
```
chrisguiney, line 346: Tiny suggestion on wording: this states the invariant but not the remedy.

Since `verifyCertificates` defaults to `true`, the operators who hit this are the ones running a custom network config — for them the next question is immediately "so what do I do?". Naming either `mithril.verifyCertificates: false` (for an explicitly unverified bootstrap) or `MithrilGenesisVerificationKey` in the message would likely save a support round-trip.

mithril/bootstrap_test.go
```diff
@@ -341,6 +369,78 @@ func TestBootstrapInvalidGenesisVerificationKey(t *testing.T) {
 	)
 }
 
+func TestBootstrapRequiresGenesisVerificationKey(t *testing.T) {
```
chrisguiney, line 372: Optional follow-on thought while you're in here: what makes this change safe is that every network in `AcceptedNetworks()` ships a parseable `genesis.vkey`. Nothing currently enforces that, and after this PR a fourth Mithril network added without one turns what used to be a silent downgrade into a hard bootstrap failure found in the field.

A small test iterating `AcceptedNetworks()`, loading each config through `LoadCardanoNodeConfigWithFallback`, and asserting `ParseVerificationKey` succeeds would pin the assumption. I wrote one while reviewing and all three networks pass — glad to hand it over if it's useful.

chrisguiney (approved):
Thanks for this — it closes the gap cleanly and, importantly, fails closed rather than degrading. Verified locally against `3a8c808c`.

The compatibility question I most wanted to answer: making the genesis key mandatory while `verifyCertificates` defaults to `true` would break any network that doesn't ship one. Traced it end to end — `AcceptedNetworks()` derives from `defaultNetworkConfigs`, which is mainnet/preprod/preview only, and those are exactly the three with a `genesis.vkey`; devnet and musashi have no Mithril config at all. Loading each through the real `LoadCardanoNodeConfigWithFallback` path over `EmbeddedConfigFS`:

```
mainnet: genesis anchor present and parseable (222 bytes)
preprod: genesis anchor present and parseable (236 bytes)
preview: genesis anchor present and parseable (236 bytes)
```

No shipped configuration regresses. And there's no added cryptographic cost: the old code already selected `VerificationModeSTM` whenever the key was non-empty, so all three real networks were doing STM verification before too — the only behavior change is that an empty key is now rejected instead of silently downgraded to structural.

Other checks: `TestBootstrapRequiresGenesisVerificationKey` fails on `main` across all four subtests, and the `requests.Load() == 0` assertion proves the guard fires before any aggregator contact rather than just eventually returning the right error. `bootstrapV2` has exactly one caller (`bootstrap.go:393`), downstream of the new guard, and `VerifyGenesisCertificateSignature` -> `ParseVerificationKey` rejects empty input anyway, so the invariant holds even off the guarded path. The v2 fixture rewrite is a real strengthening — genuine STM proofs and a real ed25519 genesis signature instead of bare BLS generators and the literal `"genesis_sig"`. `-race` clean on `./mithril/...`; `gofmt`, `vet`, and the config/cmd suites clean; merges cleanly onto current main and passes post-merge.

Approving. Four non-blocking notes inline, none of which need to hold up the merge.

PR #3482, opened 2026-08-25
fix(ledger): read reward parameters from the performance epoch

### LINKED ISSUE #3481: ledger: reward calculation reads decentralization from the calculation epoch, not the performance epoch
## Problem

`rewards.Calculate` derives both `expectedBlocks` and the eta short-circuit from `params.Decentralization`, which `rewardParameters` populates from the **calculation epoch** (the pots epoch, `newEpoch-1`). cardano-ledger's `createRUpd` takes `d` from `prevPp` — the parameters in force during the epoch whose blocks are being counted, which is Dingo's **performance epoch**.

`rewardParameters` already resolves the performance epoch's value, but returns it separately as `performanceDecentralization` and only feeds it to `rewardBlockCounts` for overlay-slot exclusion. It never reaches `Calculate`.

The two agree whenever `d` is unchanged across those epochs, which is why this has gone unnoticed. It diverges on any epoch where `d` changes.

## Observed on Preview

Preview's `decentralisation` is 1 at epochs 0 and 1, and 0 from epoch 2 onward (Koios `/epoch_params`).

At the 2→3 boundary Dingo resolves performance epoch 1, pots epoch 2. It used `d = 0` (epoch 2), so `rewardEfficiency` did not take its `d >= 0.8` short-circuit and instead computed `totalBlocks / expectedBlocks`. `totalBlocks` is 0 — mark snapshot epoch 0 is empty, and every epoch-1 block was an overlay/OBFT block anyway — so eta was 0 and the monetary expansion was skipped entirely. Only the fee pot moved:

```
applied stake rewards: reward_snapshot_epoch=0 performance_epoch=1 pots_epoch=2
  total_reward_pot=206597 available_rewards=165278 undistributed_rewards=165278
```

| epoch 3 field | Dingo | Koios |
| --- | ---: | ---: |
| `totals_treasury` | `17994600128877` | `26983803369087` |
| `totals_reserves` | `14982005400515513` | `14973016197275303` |

Dingo's treasury moved by 41319 and its reserves by 165278 — together exactly epoch 2's fee pot of 206597. The reference moved by 8989203281529, which is `floor(0.2 * (floor(0.003 * 14982005400350235) + 206597))`: a full expansion, reachable only via the `d >= 0.8` short-circuit on epoch 1's `d = 1`.

## Scope

`params.Decentralization` is not confined to eta. It also feeds `expectedBlocks` and the pool-reward path (apparent performance). cardano-ledger sources all of them from `prevPp`, so the correction is to resolve the whole reward parameter set's `d` from the performance epoch rather than patching the eta call alone. That makes this a consensus-relevant change to reward amounts on any network where `d` moves, and it needs its own test surface.

## Reproduction

```
  --port 3411 --private-port 3412 --metrics-port 12811 --debug-port 0 \
```

Reaches epoch 3 in about two minutes from genesis. Strict mode exits non-zero on the mismatch; `check_mismatches` in `{data-dir}/.koios/cache.db` holds the values.

Requires the epoch-0 seeding fixes on the #3381 branch, without which epochs 1–3 skip earlier for missing reward inputs and this defect is masked.

## Acceptance

Dingo's epoch-3 treasury and reserves match the Koios Preview reference after a genesis replay, and the reward parameters used at a boundary reflect the performance epoch's `d`.

chrisguiney:
Fixes #3481. Stacked on #3478 — retarget to `main` once that merges.

## Problem

`rewards.Calculate` derived `d`, `rho`, `tau` and the pool-level parameters from the **calculation epoch**'s protocol parameters (`newEpoch-1`).

cardano-ledger's `startStep` ([`LedgerState/PulsingReward.hs`](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState/PulsingReward.hs)) binds `pr = es ^. prevPParamsEpochStateL` and reads all of them from it:

```haskell
pr = es ^. prevPParamsEpochStateL
deltaR1 = rationalToCoinViaFloor $ min 1 eta * unboundRational (pr ^. ppRhoL) * fromIntegral reserves
d = unboundRational (pr ^. ppDG)
expectedBlocks = floor $ (1 - d) * unboundRational (activeSlotVal asc) * fromIntegral (unEpochSize slotsPerEpoch)
eta | d >= 0.8 = 1
    | otherwise = blocksMade % expectedBlocks
deltaT1 = floor $ unboundRational (pr ^. ppTauL) * fromIntegral rPot
mkPoolRewardInfoCurry = mkPoolRewardInfo pr _R b ...
```

`prevPParams` during the epoch that computes the update is the set in force over the epoch whose blocks are counted — Dingo's **performance epoch** (`newEpoch-2`). `updateRewards` reads the protocol version from the same place.

The two sources agree whenever the parameters did not change across the boundary, so a network that never moves them cannot tell the difference. That is why this went unnoticed.

## Observed on Preview

Preview's `decentralisation` is 1 at epochs 0–1 and 0 from epoch 2 onward. At the 2→3 boundary Dingo used the calculation epoch's `d = 0`, so `rewardEfficiency` skipped its `d >= 0.8` short circuit and computed `totalBlocks / expectedBlocks` instead. `totalBlocks` is 0 — mark snapshot epoch 0 is empty, and every epoch-1 block was an overlay/OBFT block anyway — so eta was 0 and the entire monetary expansion was dropped. Only the fee pot moved:

```
applied stake rewards: reward_snapshot_epoch=0 performance_epoch=1 pots_epoch=2
  total_reward_pot=206597 available_rewards=165278 undistributed_rewards=165278
```

| epoch 3 field | before | after / Koios |
| --- | ---: | ---: |
| `totals_treasury` | `17994600128877` | `26983803369087` |
| `totals_reserves` | `14982005400515513` | `14973016197275303` |

## Change

- `rewardParameters` builds `rewards.Parameters` from the performance epoch's protocol parameters.
- The epoch length still comes from the calculation epoch, standing in for the `slotsPerEpoch` the RUPD rule passes for the epoch it runs in.
- `rewardDecentralizationFromPParams` is removed: the decentralization passed to `rewardBlockCounts` for overlay-slot exclusion is now the same value the calculation itself uses, so there is one source rather than two.
- The `applied stake rewards` log field is renamed `performance_pparams_type`, since that is now what it reports.
- `ARCHITECTURE.md` documents the split.

```
  --port 3421 --private-port 3422 --metrics-port 12821 --debug-port 0 \
```

Before: strict observer fails at epoch 3 with the two mismatches above.

After: epochs 0 through 6 all validate and the run reaches slot 707032. Treasury and reserves match Koios `/totals` exactly at every epoch through 7:

| epoch | treasury | reserves |
| ---: | ---: | ---: |
| 3 | `26983803369087` | `14973016197275303` |
| 4 | `35967613128648` | `14964032387721723` |
| 5 | `45195372193123` | `14954804628961481` |
| 6 | `54434502740001` | `14945565498414603` |
| 7 | `63689264884685` | `14936310736269919` |

The run now stops at epoch 7 on a `pool_presence` mismatch — two pools present in Dingo's reward pool inputs that Koios has no history for. That is a pool-set question, not pot accounting, and is filed separately as #3483.

## Tests

Both fail on the parent branch and pass here, verified by reverting `ledger/reward_calculation.go` and re-running:

- `TestApplyStakeRewardsPreviewEpoch3Pots` — the 2→3 boundary with `d = 1` at the performance epoch and `d = 0` at the calculation epoch, pinned to the Koios reference.
- `TestRewardParametersSplitCalculationAndPerformanceEpochInputs` — renamed from `TestRewardParametersUseRUPDCalculationEpochLength`, which already seeded divergent pparams but asserted `tau` and `d` came from the calculation epoch. It now pins the whole contract: epoch length from the calculation epoch, protocol parameters from the performance epoch.

The existing epoch-1 and epoch-2 Preview tests pass unchanged in both directions — Preview's `d` is 1 at both epochs 0 and 1, so those rounds cannot distinguish the two sources.

## Checks

- `make test` (race): pass
- `go test -race ./ledger/`: pass
- `go test ./internal/test/conformance/`: pass

---
## Summary by cubic
Reads reward parameters from the performance epoch to match cardano-ledger, fixing dropped monetary expansion on Preview at the 2→3 boundary. Previously `d`, `rho`, `tau`, and pool params came from the calculation epoch; now they come from the performance epoch, while epoch length still comes from the calculation epoch.

- Uses the performance epoch's protocol parameters for `d`, `rho`, `tau`, and pool-level inputs; keeps calculation epoch's `EpochLength`.
- Removes `rewardDecentralizationFromPParams`; decentralization used for overlay-slot exclusion now equals `params.Decentralization`.
- Renames log field `pparams_type` to `performance_pparams_type`.
- Validated by Preview genesis replay: treasury and reserves match Koios `/totals` through epoch 7; adds tests pinning the 2→3 pots and input split.

<sup>Written for commit 2e603a0e814aa3608f0a2cd779153eeab549d2eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3482?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1959,6 +1959,19 @@ networks that declare Shelley at genesis, such as preview, run both. Later
 updates use the normal delayed E-3 snapshot, E-2 performance, and E-1 pots
 mapping.
 
+Within that mapping, every protocol-parameter input to the calculation — d,
+rho, tau, and the pool-level parameters — is read from the **performance**
+epoch, not the pots epoch. cardano-ledger's `startStep` binds
+`pr = es ^. prevPParamsEpochStateL` and derives all of them from it, including
+the parameter set it hands to `mkPoolRewardInfo`; `updateRewards` reads the
+protocol version from the same place. `prevPParams` during the epoch that
+computes the update is the set in force over the epoch whose blocks are
+counted, which is the performance epoch. Only the epoch length comes from the
+pots epoch, standing in for the `slotsPerEpoch` the RUPD rule passes for the
+epoch it runs in. The two sources agree whenever the parameters did not change
+across the boundary, so a network that never moves them cannot tell the
+difference.
+
 Both bootstrap rounds resolve against mark snapshot epoch 0, so that row must
 exist even when it is empty. `snapshot.Manager.CaptureGenesisSnapshot` persists
 it on a fresh sync whose genesis registers no pools — preview, whose pools
--- a/ledger/reward_calculation.go
+++ b/ledger/reward_calculation.go
@@ -491,7 +491,7 @@ func (ls *LedgerState) applyStakeRewardApplication(
 		"reward_snapshot_epoch", app.epochs.snapshot,
 		"performance_epoch", app.epochs.performance,
 		"pots_epoch", app.epochs.pots,
-		"pparams_type", fmt.Sprintf("%T", app.pparams),
+		"performance_pparams_type", fmt.Sprintf("%T", app.pparams),
 		"precomputed", app.precomputed,
 		"total_reward_pot", app.totalRewardPot,
 		"available_rewards", app.availableRewards,
@@ -2612,15 +2612,6 @@ func (ls *LedgerState) rewardParameters(
 			performanceEpoch,
 		)
 	}
-	performanceDecentralization, err := rewardDecentralizationFromPParams(
-		performancePParams,
-	)
-	if err != nil {
-		return nil, rewards.Parameters{}, nil, fmt.Errorf(
-			"get decentralization for reward performance epoch %d: %w",
-			performanceEpoch, err,
-		)
-	}
 	calculationEpochRow, err := ls.db.Metadata().GetEpoch(
 		calculationEpoch,
 		txn.Metadata(),
@@ -2638,33 +2629,27 @@ func (ls *LedgerState) rewardParameters(
 			calculationEpoch,
 		)
 	}
-	eraDesc, ok := ls.eraById(calculationEpochRow.EraId)
-	if !ok || eraDesc == nil {
-		return nil, rewards.Parameters{}, nil, fmt.Errorf(
-			"unknown era ID %d for reward calculation epoch %d",
-			calculationEpochRow.EraId, calculationEpoch,
-		)
-	}
-	pparams, err := ls.db.GetPParams(
-		calculationEpoch,
-		eraDesc.Id,
-		eraDesc.DecodePParamsFunc,
-		txn,
-	)
-	if err != nil {
-		return nil, rewards.Parameters{}, nil, fmt.Errorf(
-			"get pparams for reward calculation epoch %d: %w",
-			calculationEpoch, err,
-		)
-	}
-	if pparams == nil {
-		return nil, rewards.Parameters{}, nil, fmt.Errorf(
-			"missing pparams for reward calculation epoch %d",
-			calculationEpoch,
-		)
-	}
+	// Every protocol-parameter input to the reward calculation comes from the
+	// performance epoch, not the calculation epoch. cardano-ledger's startStep
+	// (LedgerState/PulsingReward.hs) binds `pr = es ^. prevPParamsEpochStateL`
+	// and reads d, rho and tau from it, then hands that same `pr` to
+	// mkPoolRewardInfo for the pool-level parameters; updateRewards reads the
+	// protocol version from it too. `prevPParams` during the epoch that
+	// computes the update is the parameter set in force over the epoch whose
+	// blocks are being counted, which is exactly this round's performance
+	// epoch.
+	//
+	// Only the epoch length is the calculation epoch's, matching the
+	// slotsPerEpoch the RUPD rule passes for the epoch it runs in.
+	//
+	// The two agree whenever the parameters did not change across the
+	// boundary, which is why reading the calculation epoch's went unnoticed.
+	// They diverge on preview at the 2->3 round: d is 1 at the performance
+	// epoch (1) and 0 at the calculation epoch (2), and taking 0 dropped the
+	// d >= 0.8 short circuit, leaving eta at 0 and suppressing that epoch's
+	// entire monetary expansion (dingo #3481).
 	params, err := rewardParametersFromPParams(
-		pparams,
+		performancePParams,
 		ls.config.CardanoNodeConfig,
 		uint64(calculationEpochRow.LengthInSlots),
 	)
@@ -2690,7 +2675,7 @@ func (ls *LedgerState) rewardParameters(
 			params.MaxLovelaceSupply,
 		)
 	}
-	return pparams, params, performanceDecentralization, nil
+	return performancePParams, params, params.Decentralization, nil
 }
 
 // applyFullPotConfig copies the CIP-0163 full-pot feature gate from the ledger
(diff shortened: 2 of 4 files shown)


wolf31o2 (dismissed):
Approving the delta against its base. Reviewed as the diff versus `fix-genesis-epoch-expansion-round`, since this is stacked on #3478 and its base is that branch, not `main`.

**The ledger citation checks out.** `startStep` binds `pr = es ^. prevPParamsEpochStateL`, and `prevPParams` during the epoch that computes the update is the set in force over the epoch whose blocks are counted - this round's performance epoch. Reading d, rho, tau and the pool-level parameters from there, while taking only the epoch length from the pots epoch to stand in for `slotsPerEpoch`, matches what the Haskell does.

**No validation was lost with `rewardDecentralizationFromPParams`.** That deletion was the part worth checking, since it carried three explicit guards. `rewardParametersFromPParams` now supplies the value and `params.Validate()` reaches `validateRatParameter("decentralization", ..., unit: true)`, which rejects nil as "missing", `Sign() < 0` as "negative", and `Cmp(oneRat()) > 0` as "greater than one" - the same three checks with the same meanings. The era mapping is also preserved exactly: Shelley/Mary/Alonzo clone the parameter, Babbage/Conway/Dijkstra keep the zero-valued `new(big.Rat)`, and an unknown type is still an error.

**Both tests are load-bearing.** Restoring the pre-fix source - refetching the calculation epoch's pparams and passing those to `rewardParametersFromPParams` - fails `TestRewardParametersSplitCalculationAndPerformanceEpochInputs` on the decentralization value and `TestApplyStakeRewardsPreviewEpoch3Pots` on the pot total itself, with the run compiling. The split test is well-built: it stores d = 1/2 for the performance epoch and 0 for the pots epoch, so neither source can be mistaken for the other, and it separately pins that `EpochLength` still comes from the pots epoch.

Checks at 7ac9d9b3 in a clean worktree: `go test ./ledger/` and `./ledger/rewards/` both green.

`ARCHITECTURE.md` records the rule and its cardano-ledger justification, including the point that makes this bug so easy to miss - the two sources agree whenever the parameters did not move across the boundary.

Two process notes, neither a code concern: this needs retargeting to `main` once #3478 merges, and #3478 currently carries a requested-changes review, so this cannot land ahead of that resolving.

chrisguiney:
@coderabbitai full review

wolf31o2 (approved):
Approving. I also went and settled the epoch-length question this contradicts #3584 on, since both PRs cite cardano-ledger and only one can be right.

**The source says this PR is correct.** In `Rules/Rupd.hs` the epoch size is not read from the parameter set at all — it is a separate argument computed from the current slot:

```haskell
let e = epochInfoEpoch ei s
slotsPerEpoch = epochInfoSize ei e
...
startStep slotsPerEpoch b es maxsupply asc k
```

`s` is the slot RUPD is running at, so `e` is the epoch the rule runs in — this repository's pots/calculation epoch. Meanwhile in `PulsingReward.hs`, `startStep` takes `EpochSize` as its own parameter and reads everything else from `pr = es ^. prevPParamsEpochStateL`: `d = pr ^. ppDG`, `deltaR1` from `pr ^. ppRhoL`, `deltaT1` from `pr ^. ppTauL`. `slotsPerEpoch` appears only inside `expectedBlocks`.

So the split is exactly the one implemented here: parameter *values* from the performance epoch via `prevPParams`, epoch *length* from the epoch the rule runs in. `ARCHITECTURE.md`'s "Only the epoch length comes from the pots epoch, standing in for the `slotsPerEpoch` the RUPD rule passes for the epoch it runs in" is an accurate description of the Haskell.

**That split is pinned by tests, which is what makes the difference detectable.** Repointing `EpochLength` at `performanceEpochRow.LengthInSlots` — precisely what #3584 does — fails four tests here, including `TestRewardParametersSplitCalculationAndPerformanceEpochInputs`, and the other three are the same ones #3584 rewrites the expectations of. Those rewrites move correct values to incorrect ones.

Re-verified at 2e603a0e with the mutation confirmed applied each time:

- Refetching the calculation epoch's pparams and passing those instead of `performancePParams` fails `TestRewardParametersSplitCalculationAndPerformanceEpochInputs` and `TestApplyStakeRewardsPreviewEpoch3Pots` — the latter on the pot total itself.
- No validation was lost with `rewardDecentralizationFromPParams`: `params.Validate()` reaches `validateRatParameter("decentralization", ..., unit: true)`, which rejects nil, negative, and greater-than-one with the same three meanings, and the era mapping is preserved exactly.

The branch is 17 commits behind `main`, but nothing on `main` has touched `ledger/reward_calculation.go`, its test file, or `ledger/rewards/` since the merge-base, so the staleness is benign rather than the silent-semantic-drift case.

`go test ./ledger/ ./ledger/rewards/` green, CI green on 10 checks.

On the overlap: #3584 fixes the same defect from #3570 and should be closed or reduced to whatever it covers that this does not, rather than both landing. They edit different lines and different `ARCHITECTURE.md` sections, so they would merge cleanly and leave the documentation asserting both readings in two places, with the combined code taking #3584's.

PR #3488, opened 2026-08-25
fix(ledger): bound Genesis peer header history

### LINKED ISSUE #3486: ledger: bound Genesis peer header-history memory
## Problem

Genesis-mode fork recovery expands the per-connection header history from its
normal 256-entry limit to the configured slot window. Each retained entry keeps
the decoded header inside a `ChainsyncEvent`, so the retained memory is bounded
by entry count but not by bytes. A large density window can therefore retain a
substantial amount of decoded header state for every tracked peer.

## Acceptance criteria

- Bound peer header-history retention by an explicit byte budget as well as by
  the recovery-relevant entry/window limit.
- Keep enough ordered header data to preserve rollback recovery and Genesis
  fork-intersection and density decisions when the budget permits.
- When the retained history cannot cover a requested recovery path, fail over
  to the existing safe re-intersection behavior instead of using incomplete
  candidate data.
- Keep multiple independent Genesis peers' histories isolated and preserve
  connection cleanup behavior.
- Add deterministic coverage that exercises the byte bound and proves the
  pre-fix retention behavior fails the bound.
- Pass the focused ledger tests, race checks, conformance tests, import-boundary
  checks, and documentation parity checks.

wolf31o2:
## Summary

- Bound each connection's Genesis header history by both its recovery window and an 8 MiB byte budget.
- Retain compact metadata and wire header bytes, rehydrating headers only for fork recovery.
- Fall back to a fresh ChainSync intersection when retained history cannot reconstruct a complete path.
- Document the retention and recovery contract.

Closes #3486.

## Validation

- Baseline regression retains all 5,000 entries and fails the byte-bound assertion.
- Focused fixed-head regressions pass repeatedly under the race detector.
- Full ledger unit and race tests pass.
- Conformance, import-boundary, documentation-parity, lint, NilAway, modernize, and golines checks pass.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2689,7 +2689,13 @@ Node composition injects an atomic Genesis-mode/window query from
 `ChainSelector` into ledger, so the same resolver automatically returns to
 Praos once the selector exits Genesis mode. Chainsync and ledger retain enough
 per-peer header ancestry for the active slot window while Genesis is active,
-then shrink back to their normal bounded history after exit.
+then shrink back to their normal bounded history after exit. The ledger's
+peer-local ancestry stores compact header CBOR and prev-hash metadata rather
+than retaining decoded header objects, and enforces both the active window's
+entry bound and an 8 MiB per-connection byte budget. If a candidate fork path
+falls outside the retained suffix, recovery fails closed to a fresh
+ChainSync intersection instead of making a density or rollback decision from
+an incomplete path.
 
 The trust problem Genesis solves for **biased fast-sync sources** — e.g. a
 local shallow peer or the Genesis Sync Accelerator (GSA), which serve blocks
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -154,6 +154,16 @@ const (
 	headerMismatchResyncThreshold = 20
 
 	maxPeerHeaderHistoryPerConn = 256
+	// Genesis fork resolution can need more than the normal K-sized history to
+	// compare a candidate's density, but retaining decoded headers for an
+	// entire slot window makes memory proportional to an attacker-controlled
+	// number of blocks. Store wire header bytes lazily and cap each peer's
+	// retained history independently of the configured slot window. The default
+	// Genesis quorum is one fast source plus two corroborators, so this leaves a
+	// bounded 24 MiB wire-history allowance across the three required peers.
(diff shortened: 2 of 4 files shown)


arepala-uml (approved):
This bounds each connection's retained Genesis fork-recovery header history by an 8 MiB byte budget in addition to the existing entry-count limit, since that count limit can itself be set to the (potentially huge, attacker-influenced) Genesis slot window — making memory effectively unbounded before this fix. It also switches retained records from decoded header objects to compact metadata + raw wire CBOR, decoding lazily only when a fork recovery actually needs a given header.

Verified against the PR head in an isolated worktree, not just read:

- `go build`, `go vet`, `gofmt -l` on all changed files — clean.
- `go test ./ledger/... -race -count=1` (full package) — all pass, including the new `TestRecordPeerHeaderHistoryBoundsRetainedBytes` and `TestPeerHeaderHistoryRehydratesWireHeader` at `-count=5` under `-race`.
- Confirmed the new tests are coupled to the fix: copying them onto the merge-base commit fails to build (`peerHeaderHistoryRecordOverhead` undefined there).
- Traced the eviction accounting by hand: the byte-budget and count-limit eviction loop always terminates (a single record is rejected outright if it alone would exceed the budget, so the loop can always reach empty history and still fit), and `retainedBytes` bookkeeping stays consistent with `byHash`/`order`.
- Checked the "decode lazily" path doesn't silently reintroduce unbounded memory: the fallback that retains a full decoded header (when `Cbor()` is empty) is exercised only by in-package synthetic/test headers (`mockHeader.Cbor()` returns nil) — real wire-decoded headers from `handleEventChainsync` always carry non-empty `Cbor()`, so the byte accounting is accurate for the actual attack surface (a malicious peer feeding many headers).
- Verified the Dijkstra/Musashi wire-type override (decoding under `BlockTypeDijkstra` when era is Dijkstra, since Conway's wire header type is used on the network for it) against `gouroboros/ledger/verify_block.go`'s own protocol-version-based type resolution — consistent, not arbitrary.
- Checked all call sites that used to read `record.event` directly now go through the new `chainsyncEvent()` rehydration helper (`findPeerForkPath`, `recoverPeerHeaderHistoryFromPointLocked`); no stray direct access left over.
- `ARCHITECTURE.md` updated to describe the new retention/recovery contract; `DATABASE.md` correctly untouched (no persisted schema change — this is in-memory only).

No blocking issues found.

PR #3492, opened 2026-08-25
perf(database): bound hot-cache updates

### LINKED ISSUE #1897: UTxO set management: lookup latency and memory at mainnet cardinality
## Summary

Dingo's hot CBOR cache performs work proportional to the full cache population when inserting an entry and when recording a sampled hit. At configured cache sizes, this makes routine cache traffic increasingly expensive as the cache fills.

This is part of the broader UTxO scalability work: the mainnet UTxO set has tens of millions of entries, so Dingo needs bounded lookup latency and memory use at realistic cardinality.

## Scope

- Replace full-map copy-on-write updates with a bounded concurrent design.
- Preserve best-effort cache writes and approximate LFU eviction behavior.
- Preserve configured entry-count and byte limits.
- Preserve existing cache contention metrics or provide compatible replacements.
- Benchmark 1,000-, 10,000-, and 50,000-entry populations for latency and allocation growth.

## Acceptance criteria

- Routine `Get` and `Put` operations do not copy all cached entries or access counters.
- Concurrent access passes race-enabled tests.
- Entry-count and byte limits remain enforced under contention.
- Tests cover eviction, replacement, mutation isolation, and metrics behavior.
- Benchmarks demonstrate bounded per-operation allocation growth as cache cardinality increases.

Roadmap: Q3 2026

wolf31o2:
## Summary

- replace full-cache copy-on-write updates with sharded bounded operations
- enforce exact count and byte limits with fixed-size LFU candidate windows
- preserve APIs and metrics while adding mutation, bounds, and cardinality coverage

## Validation

- `go test -race -tags dingo_extra_plugins -timeout 20m ./database/...`
- focused HotCache and tiered-cache race coverage
- `go vet ./database/...`
- golangci-lint
- NilAway
- modernize
- benchmarks at 1,000, 10,000, and 50,000 entries

Closes #1897

---
## Summary by cubic
Bounds `HotCache` updates by replacing the copy-on-write map with 64 sharded locks and fixed-size LFU eviction windows. The old behavior could temporarily stay over `maxSize`/`maxBytes` under contention; the new behavior enforces limits exactly after each `Put` while keeping per-op cost independent of cache cardinality. Metrics and APIs are preserved; the “CAS” counters now reflect non-blocking update-lock attempts.

- Refactors
  - Shards `HotCache` with per-shard `RWMutex` and a global update lock for exact size/byte accounting.
  - Admits entries only if a bounded 64-entry LFU candidate window can free enough space; otherwise drops the insert and rotates the window.
  - Samples access-count updates on `Get` (1-in-4) with bounded, non-blocking shard updates.
  - Keeps metric names (`dingo_hot_cache_cas_attempts_total`, etc.) and `CASStats()` while changing their meaning to update-lock attempts.
  - Updates `ARCHITECTURE.md`, adds cardinality-invariant benchmarks, and expands tests for limit enforcement, isolation, contention, and metrics.

- Bug Fixes
  - Prevents the cache from remaining over `maxSize`/`maxBytes` when eviction previously lost CAS races to access-count updates.
  - Serializes membership and byte accounting so `totalBytes` can’t desynchronize from entries.

<sup>Written for commit 1d02d751bacbcdba078f3ef34c293aecd4c8db8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3492?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1681,7 +1681,7 @@ CBOR Data Request
 Tier 1: Hot Cache (in-memory)
   - UTxO entries: configurable count (HotUtxoEntries)
   - Transaction entries: configurable count + byte limit
-  - O(1) access, LRU eviction
+  - Sharded O(1) access, approximate LFU eviction
        | miss
        v
 Tier 2: Block LRU Cache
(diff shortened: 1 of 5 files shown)


wolf31o2:
@coderabbitai review

arepala-uml (approved):
## Review

Verified independently in an isolated worktree — traced the locking design by hand and reproduced every claimed check rather than trusting the PR description or bot output.

**Design traced:** `HotCache` moves from a single-pointer copy-on-write snapshot to 64 lock-sharded maps plus a global `updateMu` that serializes membership/size/byte bookkeeping and the shared `evictionOrder` list. Confirmed by reading the code (not just the description):
- `evictionOrder` (the LFU candidate-window source) is mutated only from within `Put`, which holds `updateMu` for its entire body — no race window on the list itself.
- Lock ordering is always `updateMu` → per-shard `mu`, acquired sequentially per victim, never nested across two shards at once — no deadlock path.
- The two shard-index hash functions (`[]byte` and `string` variants) use identical FNV-1a math, so a key always resolves to the same shard regardless of call path.
- Bounded eviction: `selectEvictionVictimsLocked` scans at most 64 candidates from the front of `evictionOrder`, sorts by access count; on success only the actual victims are removed, on failure the *entire* scanned window rotates to the back so the next `Put` inspects fresh candidates instead of re-scanning the same unevictable ones. `TestHotCacheBoundedAdmissionDropPreservesExistingEntries` exercises exactly this path and passes.
- Hot-but-never-`Put`-again keys are protected from the "stale insertion position" eviction risk because `Get`'s sampled access-count increments feed the same `accessCnt` field the LFU sort uses.

**Checks reproduced, not assumed:**
- `go build ./...`, `go vet ./database/...` — clean
- `golangci-lint run ./database/...` — 0 issues
- `nilaway ./database/...` — clean
- `go test -race -tags dingo_extra_plugins -timeout 20m ./database/...` — full suite green
- Concurrency-sensitive hot-cache/tiered-cache tests repeated 20x under `-race` — clean, no flakes
- API/metrics compatibility confirmed: `cbor_cache.go`'s diff against `HotCache` is comment-only; `TieredCborCache`'s public surface is untouched

**Docs:** `ARCHITECTURE.md`'s Tier 1 description was checked line-by-line against the new implementation (shard count, bounded candidate window, rotation-on-failure, metric semantics change from CAS-attempts to lock-attempts) — accurate.

**Tests:** old fuzzy bounds (`maxSize+5`, `maxBytes+valueSize*3`) replaced with exact-limit assertions now that eviction is genuinely atomic with the triggering `Put`; new allocation-bound and byte-scaling tests directly guard the PR's core claim (per-op cost independent of cache cardinality).

**Bots:** CodeRabbit rate-limited on this PR. Cubic: `NEUTRAL`, zero inline findings.

PR #3524, opened 2026-08-25
fix(ledgerstate): import historical reward parameters

### LINKED ISSUE #3355: ledgerstate: Mithril bootstrap omits historical pparams required by reward rollover
## Summary

A Mithril ledger-state import seeds historical Mark/Set/Go reward inputs but
persists protocol parameters only for the snapshot state's current epoch. The
first complete reward round that needs a historical performance epoch then
fails with `missing pparams for reward performance epoch N`. Epoch rollover
restarts indefinitely and the node stops following the chain.

This reproduced independently on preview and mainnet. The same control flow is
present on current `main` at `557e602a`.

## Preview reproduction

- Dingo `9c8119ce87e139738e20193f4ec70e1dc2e80b7e`
- Fresh database
- Mithril v2 bootstrap
- Preview snapshot state in epoch 1397

The import successfully seeded all three reward-input epochs and then imported
protocol parameters:

```text
seeded reward inputs for an imported epoch epoch=1397 snapshot=mark pools=724 delegators=79743
seeded reward inputs for an imported epoch epoch=1396 snapshot=set  pools=723 delegators=79737
seeded reward inputs for an imported epoch epoch=1395 snapshot=go   pools=743 delegators=79744
importing protocol parameters
checkpoint saved phase=pparams
```

At the next boundary:

```text
handling epoch transition previous_epoch=1397 new_epoch=1398 boundary_slot=120787200 snapshot_slot=120787199
block processing failed, restarting pipeline error="process epoch rollover: apply stake rewards: missing pparams for reward performance epoch 1396"
```

The applied tip remains at slot `120787198`. The pipeline reported the failure
as deterministic after 50 retries and had exceeded 2,300 no-progress restarts
when checked. The reference `cardano-node` continued syncing.

## Mainnet corroboration

A separate fresh mainnet Mithril bootstrap using Dingo
`2f10b030f695ee1bbcd7f0b1ef6c49432feffb80` completed its ledger-state import,
entered serve mode, and then failed at epoch rollover with the same signature:

```text
process epoch rollover: apply stake rewards: missing pparams for reward performance epoch 649
```

This makes the defect cross-network rather than preview-specific.

## Code path

`ledgerstate/import.go:importPParams` calls `SetPParams` once with
`cfg.State.Epoch`. In contrast, imported reward input seeding persists bases for
the historical Mark/Set/Go epochs.

At rollover, `ledger/reward_calculation.go:rewardParameters` loads protocol
parameters for the reward performance epoch. For the preview boundary into
1398 that is epoch 1396, but the import stored only epoch 1397, so `GetPParams`
returns nil and aborts the rollover.

The Byron performance-epoch guard merged in #3241 does not cover this case:
preview epoch 1396 and mainnet epoch 649 are post-Byron epochs.

## Related work

- #3313 / #3349 concern imported reward-input bundles rejected because pool
  registration economics are incomplete. In this reproduction all three
  bundles were seeded successfully; historical protocol parameters are the
  missing input.
- #1959 broadly tracks reward calculation but does not describe this bootstrap
  failure.

## Acceptance criteria

- A Mithril import provides the exact protocol parameters required by every
  historical reward basis it makes eligible for application.
- The first complete reward round after bootstrap crosses the epoch boundary
  without a no-progress restart loop.
- Current-epoch parameters are never substituted for a historical performance
  epoch.
- A regression covers the preview shape: snapshot state at epoch 1397,
  successfully seeded reward bases for 1395-1397, and rollover into epoch 1398
  requiring epoch 1396 parameters.
- The no-historical-parameters negative case fails closed without applying an
  incorrectly parameterized reward result or indefinitely restarting rollover.

wolf31o2:
## Summary

- extract distinct current and previous protocol parameters from imported ledger state
- atomically persist the exact E-1 and E rows required by eligible imported reward bases
- fail import when required historical parameters are missing or incompatible with their epoch era

## Tests

- Preview-shaped 1397 to 1398 production rollover regression
- focused ledger and ledgerstate race tests
- conformance, lint, vet, and repository checks

Closes #3355


wolf31o2:
@coderabbitai review

ledgerstate/import.go
```diff
@@ -2607,37 +2608,156 @@ func importPParams(
...
+				err,
+			)
+		}
+	}
+
+	var (
+		previousEpoch uint64
+		previousEra   int
+		previousCbor  = []byte(cfg.State.PrevPParamsData)
+	)
```
cryptodj413, line 2799: Merge blocker: this block demands the E-1 historical `pparams` row whenever `requiresPrevious` is true, without ever checking whether a satisfying row is already present. That fails two real, reproducible cases where the import should succeed:

**(a) Cross-era window.** When a hard fork lands exactly at epoch E, `importedEraForEpoch` (below, ~line 2776) correctly resolves E-1 to the *old* era — but cardano-ledger's era translation upgrades the GovState's previous-params field to the *new* era's shape before the snapshot is taken. `validatePParamsData(oldEra, previousCbor)` can never pass in that window. Verified against real bytes: the checked-in fixture's actual 1401-byte Conway payload decodes as Conway and fails Babbage validation with a CBOR field-count mismatch.

**(b) Catch-up/reconcile re-entry.** `mithril/sync_import.go` disables the import checkpoint on reconcile, so `importPParams` re-runs on every catch-up. Reproduced: seeding the reward bases plus the *correct* historical row, then re-running the import, still fails here — the demand doesn't check for an already-satisfying row. Because the abort happens before the current-row write, the database ends up with **no** `pparams` row at all, and because the snapshot-seeding phase (`importSnapShots`) already committed upstream, a resumed bootstrap re-enters this exact failure forever with no operator remediation.

Pre-PR, this same case at least imported (permissively) and only wedged later at the reward-round boundary; post-PR it fails earlier and unrecoverably on cases that should succeed.

Suggested fix, in order of preference:
1. Move this precondition into the snapshot-seeding phase (`seedImportedRewardInputs` in `ledgerstate/imported_reward_inputs.go`), which already has a documented single-basis skip-with-warning path for an unsatisfiable basis — that degrades gracefully instead of bricking the bootstrap.
2. At minimum, before erroring, check whether `store.GetPParams(previousEpoch, uint(previousEra), ...)` already returns a satisfying row, and skip the demand when it does.

(Related: the `requiresPrevious`/`requiresCurrent` gate above also can't distinguish an imported reward-basis row from a live one — `models.RewardSnapshot.Authoritative` exists for exactly that distinction but isn't consulted here — which is a contributing cause of case (b).)

wolf31o2: Fixed in 4ae93097. Imported reward bases now validate their required parameter rows before persistence; translated cross-era history is not relabeled, valid stored rows satisfy re-entry, stale provisional rows are removed transactionally, and authoritative rows are preserved. Regression coverage includes cross-era re-entry and focused race tests.

chrisguiney (approved):

PR #3534, opened 2026-08-25
fix(koiosparity): preserve mismatch evidence across failed checks

### LINKED ISSUE #3410: koios: preserve mismatch evidence across failed checks
`internal/koiosparity/check.go` deletes an epoch's prior mismatch evidence before all fallible reads complete. A later read/insert/status-upsert failure erases the last known evidence.

Acceptance criteria:

- Complete and validate all reads before replacing prior evidence.
- Keep the previous record on any failed write and add transactional failure tests.

arepala-uml:
## Summary

`checkEpoch` deleted an epoch's prior `check_mismatches` rows before every fallible read for that epoch had completed. A later read or write failure left the epoch with zero recorded evidence instead of its last known result.

Closes #3410

## Changes

- Replace `DeleteEpochMismatches`/`InsertMismatches` in `cache.go` with a single `CommitEpochMismatches` that deletes and reinserts an epoch's mismatch rows in one transaction, so a failed insert rolls the delete back with it.
- Remove the premature delete in `checkEpoch` (`check.go`). The replacement now happens only after every fallible read for the epoch has already succeeded — either in the pre-staking short-circuit (no reads needed) or at the end of `checkEpoch` once the final mismatch set is built.

## Tests

- `TestCommitEpochMismatchesRollsBackOnFailedInsert` (`cache_test.go`): seeds prior evidence, uses a SQLite trigger to force a mid-insert failure on a replacement call, and asserts the prior row survives.
- `TestCheckEpochPreservesPriorMismatchEvidenceOnLaterReadFailure` (`check_test.go`): seeds prior evidence, drops the `koios_totals` table to force a real `GetTotals` failure inside `checkEpoch`, and asserts the prior evidence survives.

Both tests were verified against the prior buggy behavior (temporarily reintroduced) and fail without the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mismatch record updates so existing evidence is preserved if saving new results fails.
  * Prevented partial updates when checking an epoch encounters a database error.
  * Ensured mismatch records are replaced consistently for both passing and failing epoch checks.

* **Tests**
  * Added regression coverage for rollback behavior and preservation of prior mismatch evidence.

### DIFF
--- a/internal/koiosparity/cache.go
+++ b/internal/koiosparity/cache.go
@@ -1170,7 +1170,10 @@ func (c *Cache) InsertCheckRun(run CheckRun) error {
 // write never leaves the cache with less evidence than before the check ran.
 // Callers must call this only once every fallible read for the epoch has
 // already succeeded, so prior evidence is never cleared ahead of a
-// still-incomplete replacement.
+// still-incomplete replacement. Each row's Network and Epoch are normalised
+// from the network/epoch parameters before insertion, mirroring
+// CommitEpochData, so a mismatched caller cannot corrupt a different epoch's
+// data.
 func (c *Cache) CommitEpochMismatches(
 	network string,
 	epoch uint64,
@@ -1201,7 +1204,9 @@ func (c *Cache) CommitEpochMismatches(
 			return err
 		}
 		defer stmt.Close()
-		for _, m := range mismatches {
+		for i := range mismatches {
+			mismatches[i].Network, mismatches[i].Epoch = network, epoch
+			m := mismatches[i]
 			if _, err = stmt.Exec(m.Network, m.Epoch, m.PoolBech32, m.StakeAddress, m.Field, m.DingoValue, m.KoiosValue, m.Category, m.CheckedAt); err != nil {
 				return err
 			}
--- a/internal/koiosparity/cache.go
+++ b/internal/koiosparity/cache.go
@@ -1162,11 +1162,20 @@ func (c *Cache) InsertCheckRun(run CheckRun) error {
 	return err
 }
 
-// InsertMismatches bulk-inserts mismatch records.
-func (c *Cache) InsertMismatches(mismatches []CheckMismatch) error {
-	if len(mismatches) == 0 {
-		return nil
-	}
+// CommitEpochMismatches atomically replaces all mismatch rows for an epoch:
+// the prior rows are deleted and the new set inserted in a single
+// transaction, the same "delete then bulk insert, commit together" pattern
+// CommitEpochData uses for pool rows. Committing both together means a
+// failure partway through the insert rolls back the delete too, so a failed
+// write never leaves the cache with less evidence than before the check ran.
+// Callers must call this only once every fallible read for the epoch has
+// already succeeded, so prior evidence is never cleared ahead of a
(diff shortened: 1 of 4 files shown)


wolf31o2 (approved):
Approving. Both halves of the fix are load-bearing, proven independently:

- Reinstating the premature delete in `checkEpoch` fails `TestCheckEpochPreservesPriorMismatchEvidenceOnLaterReadFailure` on its own message - "a later read failure must never erase the prior mismatch evidence".
- Breaking `CommitEpochMismatches`' atomicity, by committing the delete in its own transaction the way the old two-call pattern did, fails `TestCommitEpochMismatchesRollsBackOnFailedInsert`. The SQLite trigger is a neat way to force a mid-batch insert failure without touching the production schema.

The error handling in `CommitEpochMismatches` is correct in the place this pattern usually goes wrong: the `if len(mismatches) > 0` block declares `var stmt *sql.Stmt` and then assigns with `stmt, err =` rather than `:=`, so the deferred rollback sees the real error instead of a shadowed copy. Worth stating because a `:=` there would have silently disabled the rollback the PR exists to add, and no test would have caught it.

The in-place normalisation of each row's `Network`/`Epoch` mutates the caller's slice, but `checkEpoch` calls `DetermineStatus(allMismatches)` before the commit and only reads `len(allMismatches)` after, so nothing observes it - and `CommitEpochData` mutates its input the same way, so the commit message's claim to mirror it is accurate rather than decorative.

One narrow residual, non-blocking and not a regression: in the pre-staking short-circuit the commit is followed by `UpsertCheckEpochStatus`, which is fallible. A failure there leaves the epoch's prior mismatches cleared with no status written - the same shape the PR fixes, one write later. It needs an epoch that previously recorded mismatches and later reports pre-staking, which is not a transition that should occur, so this is an observation rather than something to change.

Checks at a9bbba2a: `golangci-lint run ./internal/koiosparity/` 0 issues, `gofmt` and `go vet` clean, and both new tests pass.

**A full local `./internal/koiosparity/` run is not part of that ledger** - it exceeds a 10-minute timeout on this machine, with and without `-race`. That is the package's baseline rather than anything here: pre-existing tests in it cost 14.02s (`TestAccountRewardsAdditiveColumnMigration`) and 17.36s (`TestCheckDetectsMissingKoiosTotalsOnUpgradedCache`), against 9.95s and 28.78s for the two added ones. CI's `go-test (Linux)` covers the full suite and is green.

wolf31o2:
Correcting one claim in the review above: the slow local run was this
environment, not `internal/koiosparity`.

Same test, same commit, only `TMPDIR` differing:

```
on-disk TMPDIR   TestAccountRewardsAdditiveColumnMigration   11.025s
TMPDIR=/dev/shm  TestAccountRewardsAdditiveColumnMigration    0.047s
```

The cost is fsync on this machine's filesystem, so calling it "the package's
baseline" was wrong. The per-test figures quoted above are real but say nothing
about the package: pure-function tests in it run in 0.00s, and CI completes the
whole repository suite in about nine minutes, which 136 koiosparity tests at
the local figures could not fit inside.

Nothing here changes the approval or the findings.

Issue #3485, opened 2026-08-25
koiosparity: a pool that leaves the pool set at K+1 has no reachable blocks_produced for epoch K


chrisguiney:
## Problem

`blocks_produced` for Koios epoch K is read from Dingo's `reward_pool_input` row at the **param epoch** (K+1), because `buildRewardStateInputs` stamps the count from `evt.PreviousEpoch`.

A pool that is in epoch K's stake basis but leaves the pool set at K+1 has no K+1 row at all, so its epoch-K block count is unreachable. `ComparePoolEpoch` reports `reward_pool_input_params` as `dingo_db_missing`, which `DetermineStatus` escalates to ERROR — and under `--koios-parity-strict` that stops the node.

This is the mirror image of #3483 (a pool *entering* the set at K+1 with no stake-epoch row).

## Evidence

Preview epoch 14, pool `pool1nk3uj4fdd6d42tx26y537xaejd76u6xyrn0ql8sr4r9tullk84y`:

```
14  pool1nk3uj…  reward_pool_input_params  (dingo empty)  present  dingo_db_missing
```

Dingo's `reward_pool_input` rows for that pool:

| epoch | delegated_stake | blocks_produced |
| ---: | ---: | ---: |
| 12 | 1008992465232 | 0 |
| 13 | 1008988185080 | 0 |
| 14 | 1008987593493 | 10 |
| 15 | *(absent)* | — |

Dingo's epoch-15 mark snapshot is complete — `reward_snapshot` reports `total_pool_count = 47` and 47 `reward_pool_input` rows exist — the pool is simply no longer in it.

Koios agrees the pool leaves the set: `pool_history` returns a row for epoch 15 (`active_stake` 1008987593493, matching Dingo's epoch-14 row under the K-1 mapping) and **nothing** for epoch 16, which is the reporting epoch Dingo's absent epoch-15 row corresponds to.

So both sides agree about the pool's lifecycle. The only thing missing is a place to read epoch 14's block count from.

## Open questions for triage

- Should an absent param-epoch row be treated as "pool left the set" rather than `dingo_db_missing` when the pool *is* present at the stake epoch and Koios has no row at K+1's reporting epoch either? That mirrors the fix applied for #3483.
- Or should `blocks_produced` be sourced somewhere that survives the pool leaving the set — the count is available from the block/ledger tables independently of the snapshot row.
- Either way, an ERROR that halts a strict-mode node for a pool whose lifecycle both sides agree on is too severe.

[...]

Issue #3381, opened 2026-08-24
ledger: Preview epoch 2 treasury and reserves diverge from Koios


wolf31o2:
## Problem

A genesis replay of the Preview fixture through slot 1,295,990 produces incorrect epoch-2 aggregate ADA pots compared with Koios:

| Field | Dingo | Koios |
| --- | ---: | ---: |
| `totals_treasury` | `0` | `17994600087558` |
| `totals_reserves` | `15000000000000000` | `14982005400350235` |

Pool membership matches at three pools on both sides, isolating the observed failure to aggregate pot accounting rather than the pool set.

## Reproduction

1. Replay the Preview fixture from genesis through slot 1,295,990 in API storage mode.
2. Start Dingo with the in-process Koios parity observer enabled for Preview, strict mode enabled, and account checks disabled.
3. Validate the cached closed-epoch reference data.

The epoch-2 comparison reports two mismatches and strict mode terminates Dingo with exit status 1.

## Expected behavior

Dingo's epoch-2 treasury and reserves match the Koios Preview reference values after a genesis replay.

## Related work

- #1903 tracks the broader Koios-backed reward parity effort.
- #1959 tracks full reward calculation work.
- #3374 hardens the in-process observer and strict failure handling used to reproduce this bug.

PR #3536, opened 2026-08-25
fix(koiosparity): correct pool presence and pool-parameter epoch alignment

### LINKED ISSUE #3483: koiosparity: preview epoch 7 reports two pools Koios has no history for
## Problem

A Preview genesis replay with the in-process Koios parity observer fails at epoch 7 with two `pool_only_dingo` mismatches. Dingo has reward pool inputs for two pools that Koios's `/pool_history` returns nothing for:

```
epoch  pool_bech32                                               field          dingo   koios  category
7      pool1pn9sffcqqzkx70m0gujks4h3wf8p4y706t2f0cjcyreekg83wtf  pool_presence  present  ""    pool_only_dingo
7      pool1a7h89sr6ymj9g2a9tm6e6dddghl64tp39pj78f6cah5ewgd4px0  pool_presence  present  ""    pool_only_dingo
```

`POST /pool_history` for the first of them returns `[]` — Koios has no history rows for that pool at any epoch, not merely at epoch 7.

This is a pool-set question, not pot accounting: the ADA pots match Koios exactly at every epoch through 7 in the same run.

## Open questions for triage

- Are these pools registered on chain but never active (no delegated stake), such that Koios omits them from pool history while Dingo still writes a `reward_pool_input` row?
- Or does Dingo include a pool in the mark snapshot that should not be there — registered too late, already retired, or with zero stake?
- Is `pool_only_dingo` the right classification for a pool Koios legitimately has no reference data for, or should the comparison treat an empty Koios history as unknown rather than as a mismatch?

Whichever way it resolves, the answer belongs in `internal/koiosparity`'s comparison rules or in the snapshot's pool selection.

## Reproduction

```
  --port 3421 --private-port 3422 --metrics-port 12821 --debug-port 0 \
```

Reaches epoch 7 in about five minutes from genesis. Strict mode exits non-zero; `check_mismatches` in `{data-dir}/.koios/cache.db` holds the rows above.

Requires the epoch-0 seeding fixes (#3381) and the reward-parameter epoch fix (#3481); without them the run fails earlier and this is never reached.

### LINKED ISSUE #3484: koiosparity: pool margin and fixed cost are read from the param epoch, not the stake epoch
## Problem

`GetPoolEpochDataMap` reads `Margin` and `FixedCost` from `reward_pool_input` at the **param epoch** (K+1), alongside `BlocksProduced`. Its doc comment asserts all three share that alignment because `buildRewardStateInputs` stamps them from `evt.PreviousEpoch`.

Empirically only `BlocksProduced` does. `Margin` and `Cost` align with the **stake epoch** (K-1) — the same row the delegated stake is read from.

## Evidence

Pool `pool1nk3uj4fdd6d42tx26y537xaejd76u6xyrn0ql8sr4r9tullk84y` on Preview, from a genesis replay:

| Dingo `reward_pool_input` | blocks_produced | cost | margin |
| ---: | ---: | ---: | ---: |
| epoch 12 | 0 | `411000000` | 1/100 |
| epoch 13 | 0 | `412000000` | 1/100 |
| epoch 14 | 10 | `412000000` | 1/100 |

| Koios `pool_history` | block_cnt | fixed_cost | margin |
| ---: | ---: | ---: | ---: |
| epoch 13 | 10 | `411000000` | 0.01 |
| epoch 14 | 15 | `412000000` | 0.01 |

For Koios epoch 13: `block_cnt` 10 is on Dingo's epoch-**14** row (K+1, as the code assumes), but `fixed_cost` 411000000 is on Dingo's epoch-**12** row (K-1). The comparison reads both from K+1 and so compares `412000000` against `411000000`.

A second pool confirms the same alignment for margin, `pool1z9nsz7wyyxc5r8zf8pf774p9gry09yxtrlqlg5tsnjndv5xupu3`:

| source | epoch 12 | epoch 13 | epoch 14 |
| --- | ---: | ---: | ---: |
| Dingo margin | 1/20 | 1/25 | 1/25 |
| Koios margin | (none) | 0.05 | 0.04 |

Dingo epoch 12 (1/20 = 0.05) matches Koios epoch 13; Dingo epoch 13 (1/25 = 0.04) matches Koios epoch 14.

## Why it went unnoticed

Both fields are constant for the great majority of pools — Preview's three genesis pools sit at `340000000` and 1/10 forever — so the wrong alignment is invisible until a pool actually changes its margin or cost. On Preview the first such change lands at epoch 13.

## Expected

For Koios epoch K, read `Margin` and `FixedCost` from the stake epoch (K-1), and keep `BlocksProduced` at the param epoch (K+1). `GetPoolEpochDataMap` already performs both reads, so this is a matter of which of the two rows each field is taken from — plus updating the `DingoPoolEpochData` field comments and ARCHITECTURE.md's "Epoch alignment" section, which currently document the incorrect grouping.

Note this means `ParamsPresent` no longer covers all three fields uniformly; margin/cost presence follows `StakePresent`.

## Reproduction

```
  --port 3431 --private-port 3432 --metrics-port 12831 --debug-port 0 \
```

Reaches epoch 13 in about five minutes from genesis and exits non-zero with:

```
13  pool1nk3uj4fdd6d42tx26y537xaejd76u6xyrn0ql8sr4r9tullk84y  fixed_cost  412000000  411000000  value_mismatch
13  pool1z9nsz7wyyxc5r8zf8pf774p9gry09yxtrlqlg5tsnjndv5xupu3  margin      1/25       0.05       value_mismatch
```

Requires #3381, #3481 and #3483; without them the run fails earlier and this is never reached.

chrisguiney:
Fixes #3483. Fixes #3484. Stacked on #3482 — retarget to `main` once that and #3478 merge.

Two independent defects in the Koios parity comparison, one commit each. Both are comparison-side: Dingo's ledger data was correct in every case.

## 1. Pool presence (#3483)

`GetPoolEpochDataMap` returns the **union** of the stake-epoch (K-1) and param-epoch (K+1) reads, and the presence loop flagged every key in that union missing from Koios's epoch-K pool set as `pool_only_dingo`.

A pool registered *during* epoch K has no K-1 row — it did not exist yet — but does have a K+1 row, because its first mark snapshot is captured at the boundary into K+1. It is not part of K's active-stake basis, and Koios has no `pool_history` row for it until K+2, so flagging it reported a divergence where both sides agree.

Observed on Preview epoch 7, where two pools registered mid-epoch stopped the strict observer. Koios places their active stake (`9497623854`, `19495450477`) at epoch 9, exactly matching Dingo's epoch-8 rows — confirming the K-1 mapping is right and the presence check was at fault.

The fix skips pools whose `StakePresent` is false. That flag exists for exactly this shape — its doc comment names "a freshly registered pool whose stake-epoch row hasn't landed yet" — and `ComparePoolEpoch` already consults it; only the presence loop did not. A pool genuinely in K's stake basis but absent from Koios is still flagged.

## 2. Pool parameter alignment (#3484)

`Margin` and `FixedCost` were read from the param epoch (K+1) alongside `BlocksProduced`. Only `BlocksProduced` belongs there: a mark snapshot records the pool parameters as of its own boundary, and those are the ones in force for the epoch that snapshot is the basis for, so cost and margin align with the stake epoch (K-1).

Pinned on one Preview pool:

| | Dingo e12 | Dingo e13 | Dingo e14 | Koios e13 | Koios e14 |
| --- | ---: | ---: | ---: | ---: | ---: |
| blocks | 0 | 0 | **10** | **10** | 15 |
| cost | **411000000** | 412000000 | 412000000 | **411000000** | 412000000 |

Blocks align at K+1; cost aligns at K-1. A second pool confirms the same for margin (Dingo e12 = 1/20 ↔ Koios e13 = 0.05; Dingo e13 = 1/25 ↔ Koios e14 = 0.04).

Both fields are constant for the great majority of pools, so the wrong alignment stayed invisible until a pool actually changed one — on Preview, epoch 13.

The fix moves both fields onto the stake-epoch read and moves their comparison into `ComparePoolEpoch`'s `StakePresent` branch, so their presence follows `StakePresent` rather than `ParamsPresent`. `ParamsPresent` now covers `BlocksProduced` alone.

**Both implementations of `RewardParitySource` are corrected.** They drifted during this work: `DingoDB` was fixed while `DatabaseSource` — the one the in-process observer actually uses — was not, so the unit tests passed while a live replay still failed at epoch 13 with identical values. `TestDatabaseSourceGetPoolEpochDataMapTracksChangingPoolParams` now covers the in-process path so they cannot diverge unnoticed again.

```
  --port 3451 --private-port 3452 --metrics-port 12851 --debug-port 0 \
```

- Before #3483: fails at epoch 7.
- After #3483: fails at epoch 13.
- After both: every epoch through **13** validates and the run reaches slot **1319110**, past the slot 1295990 target in #3381.

## Tests

All fail on the parent branch and pass here, verified by reverting the source files and re-running:

- `TestCheckIgnoresParamEpochOnlyPoolForPresence` — a pool with only a K+1 row must not be flagged.
- `TestCheckFlagsStakeEpochPoolMissingFromKoios` — negative case; a pool in K's stake basis but absent from Koios must stay flagged, so the fix cannot widen into suppressing genuine findings.
- `TestGetPoolEpochDataMapTracksChangingPoolParams` — the real Preview values, on `DingoDB`.
- `TestDatabaseSourceGetPoolEpochDataMapTracksChangingPoolParams` — the same, on the in-process `DatabaseSource`.
- `TestGetPoolEpochDataMapAlignsRewardScheduleEpochs` updated: the stake-epoch row now carries the real cost/margin and the param-epoch row carries decoys, so a regression to the old grouping fails loudly.

Four other fixtures across `check_test.go`, `compare_test.go` and `source_test.go` encoded the old grouping and were updated to the corrected contract.

`ARCHITECTURE.md`'s "Epoch alignment" section and `DingoPoolEpochData`'s field comments both documented the incorrect grouping and are corrected.

## Known follow-up

The replay now stops at **epoch 14** with an ERROR, filed as #3485: the mirror image of the first defect. A pool that is in epoch K's stake basis but *leaves* the pool set at K+1 has no param-epoch row, so its epoch-K `blocks_produced` is unreachable and reports `dingo_db_missing`, which escalates to ERROR and halts a strict-mode node. Both sides agree on that pool's lifecycle — Dingo's epoch-15 snapshot is complete at 47 pools and correctly omits it, and Koios has no epoch-16 row for it either.

## Checks

- `make test` (race): pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected reward calculations to use performance-epoch parameters while retaining calculation-epoch timing data.
  * Improved reward handling across epoch boundaries, including preview calculations.
  * Corrected pool cost and margin comparisons to use stake-epoch data, while block counts use parameter-epoch data.
  * Prevented pools present only in parameter-epoch data from being incorrectly flagged.
  * Improved missing-data detection so absent records are not treated as zero values.

* **Documentation**
  * Clarified epoch alignment and data validation behavior.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5227,7 +5227,8 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
   `koiosStakeEpoch`/`koiosParamEpoch` and reads each field group from the one
   that actually matches:
   - **stake epoch (K-1):** `epoch_summary.TotalActiveStake` and
-    `reward_pool_input`'s `DelegatedStake`/`DelegatorCount` are the mark stake
+    `reward_pool_input`'s `DelegatedStake`/`DelegatorCount`/`Margin`/`Cost`
+    are the mark stake
     distribution Praos actually used as K's active-stake/reward-calculation
     basis — derived from `ledger/reward_calculation.go`'s
     `stakeRewardEpochsForNewEpoch` (`epochs.snapshot = epochs.performance - 1`,
@@ -5244,12 +5245,19 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
     `compareEpochAccounts` (`check.go`) reads Dingo's `reward_account_output`
     at `stakeEpoch`, reusing `koiosStakeEpoch` rather than deriving a second,
     possibly-diverging offset.
-  - **param epoch (K+1):** `reward_pool_input`'s `BlocksProduced`, `Margin`,
-    and `Cost` (fixed cost) describe the epoch *before* the row's own Epoch —
-    `ledger/snapshot/rotation.go`'s `buildRewardStateInputs` stamps them from
-    `evt.PreviousEpoch` onto the row captured for the new epoch at snapshot
-    time, independent of the stake-epoch offset above (which governs that
-    same row's `DelegatedStake`/`DelegatorCount` instead).
+  - **param epoch (K+1):** `reward_pool_input`'s `BlocksProduced` describes
+    the epoch *before* the row's own Epoch — `ledger/snapshot/rotation.go`'s
+    `buildRewardStateInputs` stamps it from `evt.PreviousEpoch` onto the row
+    captured for the new epoch at snapshot time, independent of the
+    stake-epoch offset above (which governs that same row's
+    `DelegatedStake`/`DelegatorCount` instead). `BlocksProduced` is the only
+    field read here. `Margin` and `Cost` were read at this epoch too until
+    dingo #3484: a mark snapshot records the pool parameters as of its own
+    boundary, and those are the ones in force for the epoch that snapshot is
+    the basis for, so they belong with the stake epoch. Both are constant for
+    most pools, which is why the wrong alignment stayed invisible until a
+    preview pool changed its cost and another changed its margin at epoch
+    13.
   - `reward_ada_pots` (treasury/reserves/fees, compared in
     `CompareEpochTotals`) is unaffected by either offset: it is a
     point-in-time ledger pot balance captured at the boundary into K itself,
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5211,8 +5211,9 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
   | `/pool_history` | exact-match | `epoch_no` | The filtered response must contain exactly the requested reporting epoch K. |
   | `/pool_history` | derived-match | `pool_id_bech32` | Request identity is decoded to Dingo's pool key hash for set membership. |
   | `/pool_history` | derived-match | `active_stake`, `delegator_cnt` | Exact values against `reward_pool_input` at stake epoch K-1. |
-  | `/pool_history` | derived-match | `block_cnt`, `fixed_cost` | Exact values against `reward_pool_input` at parameter epoch K+1. |
-  | `/pool_history` | derived-match | `margin` | K+1 values compared as equivalent rational numbers. |
+  | `/pool_history` | derived-match | `block_cnt` | Exact value against `reward_pool_input` at parameter epoch K+1. |
+  | `/pool_history` | derived-match | `fixed_cost` | Exact value against `reward_pool_input` at stake epoch K-1: a mark snapshot records the pool parameters in force for the epoch it is the basis for. |
+  | `/pool_history` | derived-match | `margin` | K-1 values compared as equivalent rational numbers. |
   | `/pool_history` | derived-match | `member_rewards` | Exact lovelace equality with aggregated `reward_pool_output.member_reward_total` at K-1. |
   | `/pool_history` | intentionally-incomparable | `pool_fees`, `deleg_rewards` | Koios derives these from an approximation that omits the pledge/owner-stake bonus and rounds components. |
   | `/pool_history` | unsupported | `active_stake_pct`, `saturation_pct`, `epoch_ros` | Dingo has no matching persisted pool aggregate. |
@@ -5366,18 +5367,19 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
   calculation may simply not have finished yet); past that window it is
   `dingo_db_missing` (a genuine gap in Dingo's own computation). Both are
   `ERROR`, never a silent `PASS`. `ComparePoolEpoch` applies the identical
-  presence/grace split to `reward_pool_input`'s param-epoch fields
-  (`blocks_produced`/`fixed_cost`/`margin`, reported together as
-  `reward_pool_input_params` when absent) via `DingoPoolEpochData.
-  ParamsPresent`, for the same reason: a not-yet-captured param-epoch row must
-  not silently compare as zero blocks/cost/margin against Koios's real values.
-  The same split applies a third time to `reward_pool_input`'s stake-epoch
-  fields (`delegated_stake`/`delegator_count`, reported together as
+  presence/grace split to `reward_pool_input`'s param-epoch field
+  (`blocks_produced` alone, reported as `reward_pool_input_params` when
+  absent) via `DingoPoolEpochData.ParamsPresent`, for the same reason: a
+  not-yet-captured param-epoch row must not silently compare as zero blocks
+  against Koios's real value. The same split applies a third time to
+  `reward_pool_input`'s stake-epoch fields (`delegated_stake`/
+  `delegator_count`/`fixed_cost`/`margin`, reported together as
   `reward_pool_input_stake` when absent) via `DingoPoolEpochData.
   StakePresent` — a pool whose stake-epoch row hasn't landed yet (e.g. a
   freshly registered pool captured first at the param epoch) must not
-  silently compare as zero stake/delegators against Koios's real values
-  either.
+  silently compare as zero stake/delegators/cost/margin against Koios's real
+  values either. `fixed_cost` and `margin` sit on this side of the split, not
+  with `blocks_produced`, because they are read at K-1 (dingo #3484).
 
 **Mismatch categories:** `value_mismatch`, `pool_only_dingo`, `pool_only_koios`,
 `dingo_db_missing` (epoch/pool row not yet computed by Dingo), `dingo_db_error`
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -7937,8 +7937,10 @@ reward application first, ADA-pot capture last) by
 
 ### Reward Calculation And Precomputation
 
-Reward protocol parameters and epoch length come from the RUPD calculation
-epoch, while block-production counts use the delayed performance epoch. TPraos
+Reward protocol parameters and block-production counts come from the delayed
+performance epoch, while epoch length comes from the RUPD calculation epoch —
+see "Blockchain State Management" above for the derivation from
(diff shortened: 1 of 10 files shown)


ARCHITECTURE.md
```diff
@@ -5052,7 +5052,8 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
   `koiosStakeEpoch`/`koiosParamEpoch` and reads each field group from the one
   that actually matches:
   - **stake epoch (K-1):** `epoch_summary.TotalActiveStake` and
-    `reward_pool_input`'s `DelegatedStake`/`DelegatorCount` are the mark stake
+    `reward_pool_input`'s `DelegatedStake`/`DelegatorCount`/`Margin`/`Cost`
```
wolf31o2, line 5231: Please update the rest of this contract to match this corrected mapping. The field matrix still groups `block_cnt`, `fixed_cost`, and `margin` under K+1, and the later presence section still groups all three under `ParamsPresent`; changed test comments repeat that obsolete grouping. `block_cnt` remains K+1, while `fixed_cost` and `margin` now come from K-1 under `StakePresent`. Since `ARCHITECTURE.md` is the change-bar contract for this behavior, these sections and comments need to agree.

wolf31o2 (changes requested):

ARCHITECTURE.md
chrisguiney: Fixed in b41be1f7. Both spots you named were still stale on head, and sweeping for the grouping rather than just those two turned up three more.

Named in the review:
- The endpoint field matrix is now three rows — `block_cnt` at K+1, `fixed_cost` at K-1, `margin` at K-1 — instead of pairing `block_cnt` with `fixed_cost` under K+1.
- The presence section groups `blocks_produced` alone under `ParamsPresent`, and moves `fixed_cost`/`margin` in with the stake-epoch fields under `StakePresent`.
- The five test/source comments in `check.go`, `check_test.go`, `source_test.go` and `dingo_db_test.go` that repeated `BlocksProduced/Margin/FixedCost` as one group.

Also found by the sweep:
- **`coverage.go`** — the one worth calling out. Its `DingoField` and `Reason` values are report output rather than commentary, and still declared `reward_pool_input.margin` and `.cost` "at K+1", with fixed_cost's reason citing "parameter-epoch alignment". A coverage report would have stated the wrong mapping.
- `checkEpoch`'s bulk-load comment, which attributed margin/fixed cost to `paramEpoch`.
- `TestComparePoolEpochEmptyDingoSideIsFlagged`'s doc comment, which described the guard in terms of `ParamsPresent` when those fields are now governed by `StakePresent`.

Comments only apart from the two coverage strings; no behaviour change. `make test` (race), lint, docs-parity and golines all clean.

One note for when you look at #3538: it currently carries a small edit to that same presence paragraph ("zero blocks/cost/margin" → "zero blocks") that belongs here. I'll resolve that in its rebase by taking this branch's version, so the correction ends up attributed to this change rather than the `pool_departed` one.

ARCHITECTURE.md
```diff
@@ -1959,6 +1959,19 @@ networks that declare Shelley at genesis, such as preview, run both. Later
...
 
+Within that mapping, every protocol-parameter input to the calculation — d,
+rho, tau, and the pool-level parameters — is read from the **performance**
+epoch, not the pots epoch. cardano-ledger's `startStep` binds
+`pr = es ^. prevPParamsEpochStateL` and derives all of them from it, including
+the parameter set it hands to `mkPoolRewardInfo`; `updateRewards` reads the
+protocol version from the same place. `prevPParams` during the epoch that
+computes the update is the set in force over the epoch whose blocks are
+counted, which is the performance epoch. Only the epoch length comes from the
+pots epoch, standing in for the `slotsPerEpoch` the RUPD rule passes for the
```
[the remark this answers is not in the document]
chrisguiney: Fixed in bc3bf6aa. Confirmed against head: the overview at the top of "Reward Calculation And Precomputation" did still state the old sourcing, and it contradicted both the code and the detailed derivation this stack added under "Blockchain State Management".

Applied essentially the proposed wording, with a pointer to where the derivation lives:

> Reward protocol parameters and block-production counts come from the delayed performance epoch, while epoch length comes from the RUPD calculation epoch — see "Blockchain State Management" above for the derivation from cardano-ledger's `startStep`.

That matches `rewardParameters`, which passes `performancePParams` to `rewardParametersFromPParams` while taking the length from `calculationEpochRow.LengthInSlots`.

Two notes on scope. The stale line came from #3482 (already merged), not from this PR — its doc update covered the detailed section and missed this summary. I carried the correction here rather than opening a separate pull request because it is a leftover from this stack's own change and this branch is already reconciling ARCHITECTURE.md's reward-alignment contract; it is a separate, clearly-scoped commit so it can be dropped if you would rather see it on its own.

I also swept for any other statement tying reward protocol parameters to the calculation epoch and found none.

Unrelated, for anyone running this locally: `make docs-parity` currently fails on macOS regardless of this branch, on `TestDocumentationDiscoveryIgnoresUntrackedFiles` from d7dd44b4. It compares `git rev-parse --show-toplevel` against an unresolved `t.TempDir()` path, and `/var` is a symlink to `/private/var`, so it always falls back to the untracked-including filesystem walk. Filed as #3598. Every other docs-parity check passes.

chrisguiney:
@coderabbitai full review

wolf31o2 (approved):

PR #3537, opened 2026-08-25
fix(ledger): guard runtime-asserted ledger capabilities at compile time

### LINKED ISSUE #2929: Add compile-time assertion that LedgerView satisfies DRepDelegationState
LedgerView.DRepDelegation satisfies gouroboros common.DRepDelegationState, which the Conway ledger rules use to reject reward withdrawals from a stake credential not delegated to a DRep (protocol versions 10 and 11). The interface is discovered at runtime via a type assertion, so if the method signature ever drifts out of sync with the gouroboros interface, the assertion silently yields ok=false and every affected reward withdrawal is rejected: a consensus-level break with no compile-time signal.

ledger/view.go already uses this guard pattern for a similar case (var _ eras.MinPoolMarginProvider = (*LedgerView)(nil)). Add the equivalent for the DRep interface:

  var _ lcommon.DRepDelegationState = (*LedgerView)(nil)

so a signature drift becomes a compile error instead of a silent runtime no-op. While there, audit the other runtime-asserted ledger interfaces for the same guard.

Follow-up from review of #2922.

chrisguiney:
## What

`LedgerView.DRepDelegation` satisfies gouroboros' `common.DRepDelegationState`,
which the Conway rules discover with a runtime type assertion
(`ledger/conway/rules.go:3208`). A drifted signature yields `ok=false` with no
compile-time signal, and on protocol versions 10 and 11 the rule then returns
`DRepDelegationStateUnavailableError` for every affected reward withdrawal.

Adds `var _ lcommon.DRepDelegationState = (*LedgerView)(nil)`, and the same
guard for the other runtime-asserted capabilities whose concrete type dingo
owns.

Closes #2929

## Audit

| Interface | Assertion site | Concrete type | Before | After |
|---|---|---|---|---|
| `lcommon.DRepDelegationState` | gouroboros `conway/rules.go:3208` | `*ledger.LedgerView` | none | added |
| `lcommon.DRepDelegationState` | same rule, conformance path | `*conformance.DingoStateProvider` | none | added |
| `eras.ByronProtocolMagicProvider` | `ledger/eras/byron.go:332` | `*ledger.LedgerView` | none | added |
| `forging.TxValidationSessionProvider` | `ledger/forging/builder.go:116` | `*ledger.LedgerState` | none | added |
| `mempool.TxValidationSessionProvider` | `mempool/mempool.go:1315` | `*ledger.LedgerState` | none | added (root package) |
| `leader.ActiveSlotCoeffRatProvider` | `ledger/leader/election.go:848` | `*epochInfoAdapter` | runtime test | added |
| `lcommon.GovPurposeRootsState` | gouroboros `conway/rules.go:604` | `*ledger.LedgerView` | present | unchanged |
| `eras.MinPoolMarginProvider` | `ledger/eras/validation.go:55` | `*ledger.LedgerView` | present | unchanged |
| `costModelsProvider` | `ledger/view.go:504` | gouroboros pparam types | n/a | left unguarded |

`conformance.StateProvider` (`ouroboros-mock/ledger/backend.go:40`) does not
embed `DRepDelegationState`, so the existing `var _ conformance.StateProvider`
check did not cover `DingoStateProvider.DRepDelegation`.

`costModelsProvider` is asserted against gouroboros era protocol-parameter
types rather than dingo types, and has a working fallback path, so a guard
would pin foreign types for a capability that is genuinely optional.

The `mempool.TxValidationSessionProvider` guard lives in the root package
rather than in `ledger`, since that is where `*LedgerState` is wired in as the
mempool's validator and `ledger` does not otherwise import `mempool`.

## Testing

Each guard was verified live by drifting the underlying method name and
confirming the compiler reports it, e.g.:

```
ledger/view.go:87:37: cannot use (*LedgerView)(nil) ... *LedgerView does not
implement common.DRepDelegationState (missing method DRepDelegation)
node.go:441:45: cannot use (*ledger.LedgerState)(nil) ... does not implement
mempool.TxValidationSessionProvider (missing method WithTxValidationSession)
```

All six drift cases produce a build failure; all edits were reverted afterward.

- `go build ./...`, `go vet`
- `go test -race ./ledger/ ./mempool/ ./internal/test/conformance/` — pass, 0 races
- `make lint` — import-boundaries pass, `golangci-lint` 0 issues; `nilaway` not
  installed in this environment, so that step did not run

## Summary by CodeRabbit

* **Tests**
  * Added compile-time checks to verify compatibility across delegation, transaction validation, protocol, and block-forging interfaces.
  * Improved safeguards against interface drift that could affect withdrawal conformance, transaction validation, or slot leadership calculations.

### DIFF
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -9138,6 +9138,14 @@ func (ls *LedgerState) txValidationSnapshot() txValidationSnapshot {
 	}
 }
 
+// Both the block builder and the mempool discover the validation-session
+// capability with a runtime type assertion on the TxValidator they were handed
+// and silently fall back to unpinned per-transaction validation when it fails,
+// so drift in WithTxValidationSession would quietly drop the snapshot pinning
+// rather than break the build. The mempool's identical interface is guarded in
+// the root package, which is where *LedgerState is wired in as its validator.
+var _ forging.TxValidationSessionProvider = (*LedgerState)(nil)
+
 // WithTxValidationSession pins a mempool revalidation batch to one immutable
 // ledger publication, one validation slot/era/parameter set, and one
 // repeatable-read database transaction. stillCurrent lets the mempool reject
--- a/internal/test/conformance/state_provider.go
+++ b/internal/test/conformance/state_provider.go
@@ -417,3 +417,10 @@ func extractCostModels(
 
 // Compile-time interface check
 var _ conformance.StateProvider = (*DingoStateProvider)(nil)
+
+// conformance.StateProvider does not include DRepDelegationState: the Conway
+// reward-withdrawal rule discovers it with a runtime type assertion instead.
+// Without this guard the harness would keep compiling after a signature drift
+// and stop exercising the protocol-version 10/11 withdrawal rule it exists to
+// cover, matching ledger.LedgerView's guard for the production path.
+var _ common.DRepDelegationState = (*DingoStateProvider)(nil)
--- a/ledger/view.go
+++ b/ledger/view.go
@@ -79,6 +79,19 @@ var _ eras.MinPoolMarginProvider = (*LedgerView)(nil)
 // back to weaker existence-only proposal ancestry checks.
 var _ lcommon.GovPurposeRootsState = (*LedgerView)(nil)
 
+// The Conway reward-withdrawal rule discovers this capability with a runtime
+// type assertion on the *LedgerView passed to ValidateTx*. On protocol versions
+// 10 and 11 a failed assertion rejects every affected withdrawal
+// (DRepDelegationStateUnavailableError), so signature drift here is a
+// consensus-level break. Make it a compile error instead.
+var _ lcommon.DRepDelegationState = (*LedgerView)(nil)
+
+// Byron redeem and bootstrap witness verification asserts this capability at
+// runtime and fails the transaction when it is absent, so drift in
+// ByronProtocolMagic would reject every Byron transaction carrying those
+// witnesses rather than fail to build.
+var _ eras.ByronProtocolMagicProvider = (*LedgerView)(nil)
+
 func (lv *LedgerView) NetworkId() uint {
 	genesis := lv.ls.config.CardanoNodeConfig.ShelleyGenesis()
 	if genesis == nil {
--- a/node.go
+++ b/node.go
@@ -433,6 +433,14 @@ func (n *Node) ouroboros() *ouroborosPkg.Ouroboros {
 	return n.ouroborosRef.Load()
 }
 
+// Run wires *ledger.LedgerState in as the mempool's TxValidator, and the
+// mempool discovers the optional validation-session capability on it with a
+// runtime type assertion, silently falling back to unpinned per-transaction
+// validation when that fails. Guard the pairing here, in the package that
+// makes it, so drift cannot quietly unpin mempool revalidation from its
+// ledger snapshot.
+var _ mempool.TxValidationSessionProvider = (*ledger.LedgerState)(nil)
+
 //nolint:contextcheck // Run is the lifecycle boundary and derives n.ctx from the caller context.
 func (n *Node) Run(ctx context.Context) (runErr error) {
 	// Configure tracing
--- a/node_forging.go
+++ b/node_forging.go
@@ -556,6 +556,13 @@ type epochInfoAdapter struct {
 	ledgerState *ledger.LedgerState
 }
 
+// computeSchedule discovers the exact-rational coefficient with a runtime type
+// assertion and silently falls back to the float64 accessor when it fails,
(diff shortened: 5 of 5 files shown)


ledger/state.go
```diff
@@ -9138,6 +9138,14 @@ func (ls *LedgerState) txValidationSnapshot() txValidationSnapshot {
 	}
 }
 
+// Both the block builder and the mempool discover the validation-session
+// capability with a runtime type assertion on the TxValidator they were handed
+// and silently fall back to unpinned per-transaction validation when it fails,
+// so drift in WithTxValidationSession would quietly drop the snapshot pinning
+// rather than break the build. The mempool's identical interface is guarded in
+// the root package, which is where *LedgerState is wired in as its validator.
+var _ forging.TxValidationSessionProvider = (*LedgerState)(nil)
```
wolf31o2, line 9147: The audit table misses one assertion in the same package family as this row.

`ledger/forging/forger.go:898`:

```go
leiosBuilder, ok := f.blockBuilder.(LeiosBlockBuilder)
if !ok {
	return nil, nil, errors.New("leios block data requires a LeiosBlockBuilder")
}
```

`f.blockBuilder` is the `BlockBuilder` interface, and `*DefaultBlockBuilder` in `builder.go` is the only production implementor - the same file as the `forging.TxValidationSessionProvider` row already in the table. It does implement `BuildBlockWithLeios`, and I confirmed `var _ LeiosBlockBuilder = (*DefaultBlockBuilder)(nil)` compiles against this branch today, so the guard is one line and costs nothing.

Lower severity than the guarded set, which is presumably why it did not surface: this assertion fails loudly with an error rather than silently degrading, and only when `leiosData` is non-empty. So drift breaks Leios forging at forge time instead of quietly changing consensus behaviour. Still the same failure mode the PR is closing - a signature change that the build accepts and a runtime path rejects.

wolf31o2 (approved):
Approving. The approach is right and the change is purely additive - `var _ I = (*T)(nil)` declarations have no runtime effect, so the only real risks are a guard that does not compile, a claim in the audit table that does not hold, or a member of the class the table missed. Checked all three.

**The drift claim reproduces.** Renaming `LedgerView.DRepDelegation` produces exactly the error the description quotes:

```
ledger/view.go:87:37: cannot use (*LedgerView)(nil) ... *LedgerView does not
implement common.DRepDelegationState (missing method DRepDelegation)
```

**The conformance row is accurate.** `conformance.StateProvider` in ouroboros-mock v0.16.0 embeds `LedgerState`, `UtxoState`, `CertState`, `SlotState`, `PoolState`, `RewardState`, `GovState` - and no `DRepDelegationState`. The pre-existing `var _ conformance.StateProvider` check genuinely did not cover `DingoStateProvider.DRepDelegation`, so that row is not redundant.

**The rest of the class is correctly excluded.** Sweeping the tree for capability assertions turns up `metadata.Restorer`/`Backuper`, `blob.Restorer`/`LocalBlockReader`, `CloudDestinationCloser`, `SnapshotLister`, `CloudManifestFetcher`, `CloudDeleter` in `database/lifecycle` and `bark`. Those are genuinely optional plugin capabilities where a non-implementing concrete type is a supported configuration, so pinning them would assert something false. Leaving them out is right, not an oversight.

Checks at 6d054f45: `go build ./...`, `go vet`, `gofmt` clean; `golangci-lint run` on the four changed packages 0 issues; `make import-boundaries` passes; `go test -race ./mempool/ ./ledger/forging/ ./ledger/leader/` green. `./ledger/` under `-race` exceeds 8 minutes on this machine and was not completed locally - CI's `go-test (Linux)` covers it and is green.

Neither bot reviewed this: CodeRabbit reports "Review rate limited" and cubic is skipping on its monthly quota, so the green CodeRabbit check is not a clean review.

One member of the class the table missed, inline. Non-blocking.

chrisguiney:
@coderabbitai review

ledger/state.go
chrisguiney: Confirmed and fixed in #3562 — a follow-up PR, since #3537 merged while I was verifying this.

You're right on both counts: `*DefaultBlockBuilder` is the only production implementor of `BuildBlockWithLeios`, and the guard compiles as-is. Verified it is live by drifting the method name and confirming the build breaks.

Why it missed the audit: my sweep matched interface names ending in `Provider|State|Source|Capability`, which excluded `LeiosBlockBuilder`. I redid it without that filter and found no further cases of the same shape. The remaining assertions are either `evt.Data.(SomeEvent)` on concrete event structs, or capability negotiation across pluggable backends (`database/lifecycle` backup/restore, the `metadata` store capabilities, `mempool.AdmissionHeadroom`) where a backend legitimately may not implement the capability — pinning those would assert a contract the plugin architecture deliberately leaves open.

PR #3538, opened 2026-08-25
fix(koiosparity): classify a departed pool instead of erroring the epoch

### LINKED ISSUE #3485: koiosparity: a pool that leaves the pool set at K+1 has no reachable blocks_produced for epoch K

chrisguiney:
Fixes #3485. Now based on `main` — #3478, #3482 and #3536 have all merged.

## Problem

`blocks_produced` for Koios epoch K is read from the `reward_pool_input` row at **K+1**, because `buildRewardStateInputs` stamps the count from `evt.PreviousEpoch`. A pool that was in epoch K's stake basis but leaves the pool set at K+1 has no such row, so `ComparePoolEpoch` reported `reward_pool_input_params` as `dingo_db_missing` — which `DetermineStatus` escalates to ERROR, halting a strict-mode node.

This is the mirror image of #3483, where a pool *entered* the set at K+1.

## Both sides agree the pool departed

Preview epoch 14, pool `pool1nk3uj4fdd6d42tx26y537xaejd76u6xyrn0ql8sr4r9tullk84y`:

| epoch | Dingo `reward_pool_input` | Koios `pool_history` |
| ---: | --- | --- |
| 13 | present, blocks 0 | present |
| 14 | present, blocks 10 | present, block_cnt 15 |
| 15 | **absent** | present |
| 16 | — | **absent** |

Dingo's epoch-15 snapshot is complete — `epoch_summary` has `snapshot_ready = 1` and `total_pool_count = 47`, with 47 `reward_pool_input` rows — the pool is simply no longer in it. Under the K-1 stake mapping, Dingo's absent epoch-15 row corresponds to Koios's absent epoch-16 row. Neither side has a divergence; there is just nowhere for epoch 14's block count to live.

## Change

`checkEpoch` resolves once per epoch whether the K+1 snapshot is committed — `GetEpochData` returns nil unless `epoch_summary.SnapshotReady` is set, so a non-nil result is exactly that — and passes it to `ComparePoolEpoch`. A pool present at the stake epoch and absent from a committed K+1 snapshot is recorded as **`pool_departed`**, a new informational category alongside the existing `acct_zero_reward` / `acct_newly_registered` / `acct_deregistered` ones: the uncomparable field is still reported so the coverage gap stays visible, but it neither fails nor errors the epoch.

An absent row with **no** committed K+1 snapshot keeps `dingo_db_missing`, and a lookup error leaves the flag false, which keeps the stricter classification.

The signal is a parameter resolved by the caller rather than a field on `DingoPoolEpochData`, so the two `RewardParitySource` implementations cannot drift on it — they did exactly that during #3484.

## Design note

The block count itself is *not* unavailable: `pool_opcert_sequence` still holds it, and the reward calculation counts blocks from there rather than from this column (`rewardPoolFromInput` takes `blocksProduced` as a parameter and never reads `input.BlocksProduced`). So a direct count could have been substituted for departed pools.

I did not do that, deliberately. For every pool where the K+1 row *does* exist, the comparison's value is that it validates the persisted column against Koios; substituting a fresh computation would quietly stop checking what Dingo actually stored. Adding a second, count-based path only for departed pools means a new source method on both implementations for a case worth one row per pool per lifetime. Recording the gap explicitly keeps the check honest at a fraction of the surface. Happy to take the other approach if a reviewer prefers it.

```
  --port 3471 --private-port 3472 --metrics-port 12871 --debug-port 0 \
```

Before: ERROR at epoch 14, node halted.

After: **epochs 0 through 15 all validate**, reaching slot 1430078 with zero `level=ERROR` lines. All 16 epoch statuses are PASS, and the departure is recorded:

```
epoch  pool                  field                     category
14     pool1nk3uj4fdd6d42tx  reward_pool_input_params  pool_departed
15     pool1nk3uj4fdd6d42tx  reward_pool_input_params  pool_departed
```

## Tests

The two departure tests fail with the classification branch disabled and pass with it; the negative cases hold in both states.

- `TestComparePoolEpochDepartedPoolIsInformational` — departed pool yields `pool_departed` and `StatusPass`, inside and outside the grace window.
- `TestCheckDepartedPoolDoesNotErrorEpoch` — end-to-end through `Check`, proving `checkEpoch` resolves the K+1 readiness itself; asserts no ERROR epochs and that the row is recorded rather than skipped.
- `TestComparePoolEpochUncapturedParamEpochStillErrors` — negative case; an uncommitted K+1 snapshot still yields `dingo_db_missing`/ERROR.
- `TestComparePoolEpochDepartedRequiresStakeEpochRow` — negative case; a pool absent from both reads is not a departure.

`ARCHITECTURE.md`'s Koios Parity Tracker section documents the exception and the new category.

## Checks

- `make test` (race): pass

## Summary by CodeRabbit

* **New Features**
  * Added informational reporting for pools that departed after a committed pool snapshot.
  * Departed pools no longer affect parity check status.

* **Bug Fixes**
  * Missing pool records remain errors when snapshot membership is unavailable or incomplete.
  * Improved distinction between departed pools and uncaptured or missing data.

* **Documentation**
  * Documented the `pool_departed` mismatch category and classification rules.

* **Tests**
  * Added coverage for departed pools, incomplete snapshots, missing records, and uncaptured reward data.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5395,7 +5395,14 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
   the uncomparable field is still reported so the coverage gap is visible, but
   both sides agree the pool departed, so it neither fails nor errors the
   epoch. A missing row with no committed K+1 snapshot keeps the stricter
-  classification. The same split applies a third time to `reward_pool_input`'s
+  classification, as does a pool still listed in that set — a degraded active
+  pool is dropped from `reward_pool_input` "without changing
+  `pool_stake_snapshot` or `epoch_summary`" (see DATABASE.md), so only absence
+  from the pool set itself proves departure. `pool_stake_snapshot` is windowed
+  while `reward_pool_input` is retained for the life of the database, so an
+  epoch older than that window has no membership evidence and keeps the
+  stricter classification. The same split applies a third time to
+  `reward_pool_input`'s
   stake-epoch fields (`delegated_stake`/`delegator_count`/`fixed_cost`/
   `margin`, reported together as `reward_pool_input_stake` when absent) via
   `DingoPoolEpochData.StakePresent` — a pool whose stake-epoch row hasn't
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5385,26 +5385,35 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
   absent) via `DingoPoolEpochData.ParamsPresent`, for the same reason: a
   not-yet-captured param-epoch row must not silently compare as zero blocks
   against Koios's real value. That split has one exception. A pool that was in
-  epoch K's stake basis but is absent from a K+1 snapshot Dingo has already
-  committed did not fail to compute — it left the pool set, and the K+1 row
-  its epoch-K block count would have been stamped onto is never written.
-  `checkEpoch` resolves once per epoch whether the K+1 snapshot exists
-  (`GetEpochData` returns nil unless `epoch_summary.SnapshotReady` is set) and
-  passes that to `ComparePoolEpoch`, which then records `pool_departed`
+  epoch K's stake basis but is absent from the K+1 pool set did not fail to
+  compute — it left the set, and the K+1 row its epoch-K block count would
+  have been stamped onto is never written. `checkEpoch` reads the K+1 mark
+  `pool_stake_snapshot` membership once per epoch through
+  `RewardParitySource.GetPoolStakeSnapshotMembers` and passes each pool's
+  proven absence to `ComparePoolEpoch`, which then records `pool_departed`
   instead. That category is informational, like the account lifecycle ones:
   the uncomparable field is still reported so the coverage gap is visible, but
   both sides agree the pool departed, so it neither fails nor errors the
-  epoch. A missing row with no committed K+1 snapshot keeps the stricter
-  classification, as does a pool still listed in that set — a degraded active
-  pool is dropped from `reward_pool_input` "without changing
-  `pool_stake_snapshot` or `epoch_summary`" (see DATABASE.md), so only absence
-  from the pool set itself proves departure. `pool_stake_snapshot` is windowed
-  while `reward_pool_input` is retained for the life of the database, so an
-  epoch older than that window has no membership evidence and keeps the
-  stricter classification. The same split applies a third time to
-  `reward_pool_input`'s
-  stake-epoch fields (`delegated_stake`/`delegator_count`/`fixed_cost`/
-  `margin`, reported together as `reward_pool_input_stake` when absent) via
+  epoch.
+
+  Membership, not `epoch_summary.SnapshotReady`, is what proves departure.
+  `SnapshotReady` is epoch-level: `saveSnapshotInTxn` writes the epoch summary
+  and the mark `pool_stake_snapshot` on every transition regardless of
+  reward-input availability, so a ready summary is compatible with the whole
+  reward-input bundle having been skipped, and `buildRewardStateInputs` drops
+  a degraded active pool from `reward_pool_input` "without changing
+  `pool_stake_snapshot` or `epoch_summary`" (see DATABASE.md). Both are
+  missing input rather than departure, and both would pass under an
+  epoch-level flag. A pool still listed in the K+1 set therefore keeps the
+  stricter classification, as does an unreadable or empty member set — empty
+  cannot distinguish "captured, no pools" from "not captured". Because
+  `pool_stake_snapshot` is windowed while `reward_pool_input` is retained for
+  the life of the database, an epoch older than that window has no membership
+  evidence and keeps the stricter classification too.
+
+  The same split applies a third time to `reward_pool_input`'s stake-epoch
+  fields (`delegated_stake`/`delegator_count`/`fixed_cost`/`margin`, reported
+  together as `reward_pool_input_stake` when absent) via
   `DingoPoolEpochData.StakePresent` — a pool whose stake-epoch row hasn't
   landed yet (e.g. a freshly registered pool captured first at the param
   epoch) must not silently compare as zero stake/delegators/cost/margin
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5405,11 +5405,23 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
   `pool_stake_snapshot` or `epoch_summary`" (see DATABASE.md). Both are
   missing input rather than departure, and both would pass under an
   epoch-level flag. A pool still listed in the K+1 set therefore keeps the
-  stricter classification, as does an unreadable or empty member set — empty
-  cannot distinguish "captured, no pools" from "not captured". Because
-  `pool_stake_snapshot` is windowed while `reward_pool_input` is retained for
-  the life of the database, an epoch older than that window has no membership
-  evidence and keeps the stricter classification too.
+  stricter classification.
+
+  Absence only proves departure against a set known to be complete, so the
+  member count is checked against the K+1 `epoch_summary.TotalPoolCount`
+  before it is trusted. `saveSnapshotInTxn` writes that count and those mark
+  rows from the same `StakeDistribution`, so equality between them is what
+  establishes completeness — deliberately the epoch summary's count and not
+  `RewardSnapshot.TotalPoolCount`, which counts the reduced reward
+  distribution with degraded pools already excluded. Without that check a
+  summary declaring two pools with only one readable mark row would make the
+  missing pool look departed and hide a `dingo_db_missing`. Anything short of
+  equality leaves membership unproven and keeps the stricter classification:
+  a read error, an empty set (which cannot distinguish "captured, no pools"
+  from "not captured"), no ready summary, a zero count, or a disagreeing
+  count. Because `pool_stake_snapshot` is windowed while `reward_pool_input`
+  is retained for the life of the database, an epoch older than that window
+  has no membership evidence and keeps the stricter classification too.
 
   The same split applies a third time to `reward_pool_input`'s stake-epoch
   fields (`delegated_stake`/`delegator_count`/`fixed_cost`/`margin`, reported
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5384,20 +5384,33 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
   (`blocks_produced` alone, reported as `reward_pool_input_params` when
   absent) via `DingoPoolEpochData.ParamsPresent`, for the same reason: a
   not-yet-captured param-epoch row must not silently compare as zero blocks
-  against Koios's real value. The same split applies a third time to
-  `reward_pool_input`'s stake-epoch fields (`delegated_stake`/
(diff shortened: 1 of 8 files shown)


internal/koiosparity/compare.go
```diff
@@ -449,8 +463,17 @@ func ComparePoolEpoch(
 	// in Dingo's own computation (dingo_db_missing).
 	if !dingoPool.ParamsPresent {
 		cat := CategoryDBMissing
-		if graceHours > 0 && !epochEndTime.IsZero() &&
-			now.Sub(epochEndTime) < time.Duration(graceHours)*time.Hour {
+		switch {
+		case dingoPool.StakePresent && paramEpochCaptured:
```
wolf31o2: `SnapshotReady` only proves the leader-election snapshot committed; it does not prove this pool departed. Snapshot capture can set it while the entire reward-input bundle is unavailable, and can omit a degraded active pool from `reward_pool_input` while retaining it in the K+1 `PoolStakeSnapshot`. This branch therefore turns genuine missing/corrupt input into `pool_departed`/PASS. Please distinguish departure using per-pool K+1 snapshot membership or an equivalent completeness signal, with negative coverage for both no-bundle and degraded-active-pool cases.

wolf31o2 (changes requested):

chrisguiney:
@coderabbitai full review

wolf31o2 (changes requested):
The existing P1 remains on this head: `SnapshotReady` is epoch-level, not per-pool departure proof, so a ready snapshot with an active degraded or missing-reward-input pool is misclassified as `pool_departed` and passes. The open inline thread contains the producer-state evidence; I did not duplicate it.

internal/koiosparity/compare.go
chrisguiney: Fixed in 0c3574af. Confirmed both cases against the producer rather than treating them as hypothetical, and both were real.

`saveSnapshotInTxn` persists the epoch summary and the mark `pool_stake_snapshot` on every transition "regardless of reward-input availability", so a ready summary is compatible with the whole bundle being skipped. `buildRewardStateInputs` separately drops a degraded active pool from `reward_pool_input` while keeping it in `PoolStakeSnapshot` — DATABASE.md states it as the contract: pools with missing or invalid registration data are excluded "without changing `pool_stake_snapshot` or `epoch_summary`". So the epoch-level flag turned both into `pool_departed`/PASS.

Departure is now decided from per-pool absence from the K+1 mark `pool_stake_snapshot`, via a new `GetPoolStakeSnapshotMembers` on `RewardParitySource` implemented by both `DingoDB` and `DatabaseSource`. That table is written in both failure cases, so absence from it is the evidence departure actually needs.

Negative coverage for both cases you named:

- `TestCheckDegradedActivePoolStillErrors` — pool present in the K+1 pool set with no reward-input row stays `dingo_db_missing`/ERROR.
- `TestCheckMissingRewardBundleStillErrors` — no reward-input rows at all while the pool set still lists the pools stays ERROR.

Both fail against the previous epoch-level classification and pass now; I verified that by mutating the discriminator back rather than assuming it. The three departure tests now share one fixture and differ only in whether the pool is in the K+1 set.

Two things from auditing the class rather than just this line:

- An unreadable **or empty** member set leaves membership unproven and keeps the stricter classification, so a failed read can never downgrade a finding. An empty set cannot distinguish "captured, no pools" from "not captured".
- `pool_stake_snapshot` is windowed while `reward_pool_input` is retained for the life of the database. For an epoch older than that window there is no membership evidence, so departure is not claimed and the finding stays ERROR — the safe direction, and the same behaviour as before this PR. Documented in ARCHITECTURE.md alongside the degraded-pool reason.

A preview genesis replay validates epochs 0 through 17 (18/18 PASS, zero ERROR lines, slot 1642059) with the real departed pool still recorded as `pool_departed` at epochs 14 and 15 — so the stricter rule does not regress the legitimate case.

chrisguiney:
@coderabbitai full review

ARCHITECTURE.md
```diff
@@ -5384,20 +5384,40 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
...
-  silently compare as zero stake/delegators/cost/margin against Koios's real
-  values either. `fixed_cost` and `margin` sit on this side of the split, not
-  with `blocks_produced`, because they are read at K-1 (dingo #3484).
+  against Koios's real value. That split has one exception. A pool that was in
+  epoch K's stake basis but is absent from a K+1 snapshot Dingo has already
+  committed did not fail to compute — it left the pool set, and the K+1 row
+  its epoch-K block count would have been stamped onto is never written.
+  `checkEpoch` resolves once per epoch whether the K+1 snapshot exists
+  (`GetEpochData` returns nil unless `epoch_summary.SnapshotReady` is set) and
+  passes that to `ComparePoolEpoch`, which then records `pool_departed`
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 32266fd4. Confirmed against head: the passage did still describe `GetEpochData`/`epoch_summary.SnapshotReady` as the mechanism. The previous commit replaced that signal with per-pool `pool_stake_snapshot` membership but only appended the new reasoning, so the paragraph named the old signal and then explained a few sentences later why that signal cannot prove departure — self-contradictory, exactly as reported.

It now states the mechanism once — `checkEpoch` reads K+1 mark membership through `RewardParitySource.GetPoolStakeSnapshotMembers` and passes each pool's proven absence to `ComparePoolEpoch` — and keeps why `SnapshotReady` is insufficient in its own paragraph, with the cases that retain the stricter classification: a pool still listed in the K+1 set, an unreadable or empty member set, and an epoch older than `pool_stake_snapshot`'s retention window.

I also swept the rest of the class rather than just this passage. The remaining `GetEpochData` calls in `checkEpoch` (lines 442 and 474) are for the stake-epoch totals and the epoch's ADA pots and are unrelated to departure, and the `SnapshotReady` occurrences in tests are fixture fields. No other description of the departure signal needed changing.

Documentation only; no behaviour change. Lint clean, koiosparity suite green, docs-parity green apart from the pre-existing macOS failure tracked in #3598.

chrisguiney:
@coderabbitai full review

internal/koiosparity/check.go
```diff
@@ -494,6 +508,40 @@ func checkEpoch(
...
+		ctx, paramEpoch,
+	); pErr != nil {
+		logger.Debug(
+			"koiosparity: could not resolve param-epoch pool-set membership",
+			"network", network,
+			"epoch", epoch,
+			"param_epoch", paramEpoch,
+			"error", pErr,
+		)
+	} else if len(members) > 0 {
```
wolf31o2: **P1 — Require a complete K+1 set before treating absence as departure.** Any non-empty membership result is accepted as complete, but this does not prove that the recorded set was fully read. With a ready summary declaring two pools and only one readable mark row, the omitted active pool is classified as `pool_departed` and the epoch passes, hiding `dingo_db_missing`. Validate member count against `epoch_summary.TotalPoolCount` (or an equivalent completeness marker) before using absence as departure proof.

wolf31o2 (changes requested):
The prior SnapshotReady departure-proof issue is addressed. One completeness gap remains in the replacement membership check.

internal/koiosparity/check.go
chrisguiney: Fixed in b2f4d256. Confirmed the gap was real: any non-empty read was accepted, so a partially recorded or partially read set would let the absent pool read as departed.

The member count is now checked against the K+1 `epoch_summary.TotalPoolCount` before absence is trusted. `saveSnapshotInTxn` writes that count and the mark rows from the same `StakeDistribution`, so equality between them is what establishes completeness.

One detail worth flagging, since the wrong count is available a few lines away in the same file: this uses the **epoch summary's** count, not `RewardSnapshot.TotalPoolCount`. The latter is `len(effective.PoolStakes)` — the reduced reward distribution with degraded pools already excluded — so it would under-count by exactly the pools this check exists to protect and would have quietly restored the original bug.

Anything short of equality leaves membership unproven and keeps `dingo_db_missing`: a read error, an empty set, no ready summary, a zero count, or a disagreeing count.

`TotalPoolCount` is exposed on `DingoEpochData` and populated by both `RewardParitySource` implementations, so they cannot diverge on it — the drift that bit this package during #3484.

`TestCheckIncompleteParamEpochPoolSetStillErrors` covers the case you described (summary declares two pools, one mark row present, subject pool absent) and asserts the epoch goes to ERROR with `dingo_db_missing`. It fails when the count comparison is removed; I verified that by mutation rather than assuming. The departure fixture is now parameterised over the declared count so the complete and partial cases differ only in that number.

A preview genesis replay still validates epochs 0 through 16 (17/17 PASS, zero ERROR lines, slot 1589783) with the real departed pool recorded as `pool_departed` at epochs 14 and 15, so the stricter rule does not regress the legitimate case.

ARCHITECTURE.md documents the completeness requirement and the reason for choosing the epoch summary's count, in this same change.

chrisguiney:
@coderabbitai full review

wolf31o2 (approved):

PR #3540, opened 2026-08-25
feat(midnight): bound indexer backfill to the applied ledger tip

### LINKED ISSUE #2122: Backfill Midnight indexer state from Mithril snapshots
When a node starts mid-chain, or has Midnight indexing enabled after initial sync, the indexer needs to backfill historical Midnight events rather than starting from the current tip.

The cleanest path is to hook into the existing Mithril bootstrap so the same snapshot that primes the chain also drives the indexer through every relevant historical block.

Scope:
- Detect at startup whether the indexer's last-processed block is behind the chain tip, and by how much.
- Drive the indexer through every block in that gap using the same scan logic that runs on live ChainBlockEvents.
- For nodes that bootstrap from a Mithril snapshot, run the indexer over the post-snapshot blocks as part of the standard catch-up rather than a separate code path.
- Idempotency: the indexer must be safe to re-run over a range it has already processed. Use insert-or-skip semantics keyed on the unique columns each table already has.
- Surface progress through the existing block-indexed counter and a dedicated catch-up gauge if useful.

Out of scope: any change to the Mithril bootstrap itself. This issue is about plugging the Midnight indexer into the catch-up path the node already runs.

chrisguiney:
Closes #2122.

## Problem

`Indexer.backfill()` swept `[checkpointSlot, math.MaxUint64)` over the blob store. On a Mithril-bootstrapped node that range is larger than the sweep should own:

- `setStableMithrilLedgerTip` (`mithril/sync.go`) leaves the metadata ledger cursor at the imported ledger state, while `node.LoadBlobsWithDB` has already copied blocks up to the certified immutable tip.
- `LedgerState.Start` then replays that suffix — `ledger/state.go`, "primary chain tip ahead of ledger tip at startup; ledgerProcessBlocks will catch up via chainsync" — emitting a `BlockActionApply` per block.

Every post-snapshot block was therefore scanned twice: once by the sweep, once live. Rows survive it because the `CreateMidnight*` inserts are `ON CONFLICT DO NOTHING`, but the scan is duplicated and `dingo_midnight_events_total` overcounts.

## Change

- `Config.LedgerTipSlot` bounds the sweep at one slot past the applied ledger tip; blocks above it reach the indexer through the live event path instead. A failed tip lookup aborts rather than silently reverting to an unbounded scan.
- `node.go` wires it to `database.Database.GetTip`, not `LedgerState.Tip()` — the latter is only loaded inside `LedgerState.Start`, which runs after the indexer has backfilled, so it would read zero here.
- Startup logs `checkpoint_slot`, `target_slot`, and `slots_behind`.
- New gauges: `dingo_midnight_indexer_checkpoint_slot`, `dingo_midnight_backfill_target_slot`, `dingo_midnight_backfill_in_progress`. The checkpoint gauge publishes at construction so a caught-up node reports its position rather than 0.

The Mithril bootstrap itself is unchanged, per the issue's out-of-scope note.

Issue scope items 2 (shared scan logic) and 4 (idempotency) were already satisfied on `main` by #2114; verified against the SQL and covered by the new tests.

## Tests

Nine tests in `midnight/indexer/backfill_test.go`, written before the implementation. Each was mutation-tested: unbounding the end slot fails four, swallowing the tip-resolution error fails the fail-fast test, and moving the gauge publishes inside the sweep fails the two position tests.

## Checks

Run: `go build ./...`; `go test -race ./...`; `golangci-lint run` (0 issues); `nilaway`; `modernize` (only pre-existing `.pb.go` findings); `make import-boundaries`; `make docs-parity`; `make gorm-check`; `golines`.

The full `-race` suite reported 74 ok / 2 FAIL. Both failures are six Unix-socket tests in `connmanager` and `internal/node` failing with `bind: invalid argument` on a 137-character socket path from the isolated `TMPDIR` used for the run — over the macOS 104-byte `sun_path` limit. All six pass with the default `TMPDIR`; neither package is touched by this change.

`make sql-check` was not run: no `database/sql` query or `sqlc.yaml` change. `make golines` was run via `go run` because the binary is not installed locally.

Not run: no live Mithril bootstrap or devnet. The duplicate scan is established by code path and covered at unit level; confirming it against a real snapshot needs infrastructure not available here. This change does not touch consensus, block production, chain selection, mempool, or the NtN/NtC protocols.

## Docs

`ARCHITECTURE.md` startup step 8 describes the new bound and the gauges. `DATABASE.md` documents the `midnight` phase of `backfill_checkpoint`. No schema change.


arepala-uml (approved):
**Summary**

Verified the Midnight indexer backfill bound (closes #2122) independently rather than relying on the PR's self-reported checks:
- Confirmed the load-bearing ordering invariant directly in `node.go`: `n.midnightIndexer.Start()` runs and its error is checked *before* `n.ledgerState.Start()` (lines 836, 912) — this is what makes reading `database.Database.GetTip` at that point valid instead of racing the ledger's own tip advancement.
- Checked the scenario that would have made this a serious regression: `sqlstore.Store.GetTip` explicitly handles `sql.ErrNoRows` by returning a zero `Tip` with `nil` error, so a brand-new node gets slot 0 rather than the new fail-fast path triggering a fatal shutdown on first-ever start.
- Verified the tip-block-inclusive boundary reasoning (`tipSlot` always corresponds to an actual stored block, so `updateCheckpoint` lands exactly on `endSlot-1` after a full sweep) and the `MaxUint64` overflow guard.
- Checked the "checkpoint ahead of tip" edge case (a rollback lowering the ledger cursor below an already-advanced indexer checkpoint) — handled as a clean no-op skip; correctly doesn't attempt to fix rollback/pruning correctness, which is out of this PR's scope.
- `go build ./...` clean. `go test ./midnight/indexer/... -race` (45 tests) and `./midnight/server/...` — pass. `golangci-lint`, `nilaway` — 0 issues. `modernize` findings are only in pre-existing generated `.pb.go` files, matching the PR's own claim. `import-boundaries`, `docs-parity`, `git diff --check` — clean.
- Regression-test-discipline: reverted `indexer.go`/`metrics.go` to pre-PR — the new test file fails to compile against it (references the new `Config.LedgerTipSlot` field), structurally proving the tests exercise the new code rather than passing vacuously.

No merge-blocking issues found. Clean, well-scoped fix with unusually thorough test coverage (mutation-tested per the PR description).

PR #3545, opened 2026-08-26
fix(database): report blob deletes that leave unreachable objects

chrisguiney:
Refs #1649

`deleteUtxoBlobs` and `deleteTxBlobs` logged every failed blob delete and then
returned `nil`, and all three production callers discarded even that with
`_ = deleteUtxoBlobs(...)`. The metadata row naming the object is removed
immediately afterwards, so a failed delete leaves an object nothing can name
again.

This is a different orphan from the one the durability contract already covers.
Those sit *beyond* the metadata tip and startup reconciliation trims them
(`cleanupOrphanedBlobs`); these sit *beneath* it with no row pointing at them,
so no sweep reclaims them and they consume disk until the store is rebuilt.

Changes:

- both deleters return `ErrBlobDeleteIncomplete` with failed and total counts
  instead of reporting a clean deletion
- the three callers log the condition at error level rather than discarding it
- each stranded object increments `dingo_database_blob_orphans_total`
  (`RegisterBlobOrphanMetrics`, wired in `Database.New` beside the existing
  block-hash index counters; `BlobOrphanCount` for in-process reads)

The callers still go on to delete the metadata. It is the source of truth, and
keeping a rolled-back UTxO in the live set because its blob is stuck would be
the worse outcome — a nil blob store, which already returned an error, was
being discarded the same way. What changes is that the continuation is now
deliberate, documented and counted rather than silent, which is what #1649 asks
for on this path.

Documentation: `DATABASE.md` gains an "Unreachable Blob Objects" section
covering the delete ordering, why the metadata delete proceeds anyway, how
these differ from the trimmable orphans, and the metric to alert on.
`ARCHITECTURE.md` checked, not affected.

Validation:

- `go test -race -count=1 ./database/...` — pass, all 19 packages, exit 0
- `go test -count=1 ./internal/test/conformance` — pass
- `./internal/test/devnet/run-tests.sh` (all-Dingo) — pass, exit 0
  (`TestEpochBoundaryConsensus` 440s, txpump submitted=6 rejected=0 error=0)
- `golangci-lint`, `nilaway` on `./database/...` — clean
- `modernize ./database/...` — 13 findings, identical on this branch and on
  `main`, all in generated sqlc files; none introduced here
- `make import-boundaries`, `make docs-parity`, `make gorm-check` — pass
- `golines` — changed files clean

New tests cover both deleters reporting and counting a failed delete, both
counting nothing on a clean run, and a nil blob store being reported. The two
existing batch-commit tests were updated: they asserted the old
always-`nil` contract, and now assert the error alongside their existing
log-content checks. `TestTransactionsDeleteRolledbackLogsBlobFailureAndDeletesMetadata`
already pinned the deliberate continuation and still passes unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added tracking and Prometheus metrics for blob objects that cannot be deleted.
  * Added an orphaned-blob counter and an error indicator for incomplete blob cleanup.
* **Bug Fixes**
  * Blob-deletion failures are now reported and logged while metadata cleanup continues.
  * Orphan counts are recorded only after metadata deletion commits, preventing false counts from rollbacks or retries.
* **Documentation**
  * Documented the causes, reporting, and monitoring of unreachable blob objects.
* **Tests**
  * Added coverage for failed and successful transaction and UTxO blob deletion scenarios.

### DIFF
--- a/database/transaction.go
+++ b/database/transaction.go
@@ -1194,7 +1194,8 @@ func (d *Database) GetTransactionMetadataByHash(
 		txn = d.Transaction(false)
 		defer txn.Release()
 	}
-	return d.transactionStore().GetTransactionMetadataByHash(hash, txn.Metadata())
+	return d.transactionStore().
+		GetTransactionMetadataByHash(hash, txn.Metadata())
 }
 
 // GetTransactionsByHashes returns transactions for the provided hashes.
@@ -1717,11 +1718,15 @@ func (d *Database) DeleteTransactionMetadataLabelsAfterSlot(
 	return nil
 }
 
-// deleteTxBlobs attempts to delete blob data for the given transaction hashes.
-// This is a best-effort operation; metadata remains the source of truth. When
-// the caller provides a blob transaction, deletions stay coupled to that outer
-// commit. A temporary blob-only transaction is used only as a fallback when no
-// blob handle is available.
+// deleteTxBlobs deletes blob data for the given transaction hashes. Metadata
+// remains the source of truth. When the caller provides a blob transaction,
+// deletions stay coupled to that outer commit. A temporary blob-only
+// transaction is used only as a fallback when no blob handle is available.
+//
+// Failures do not stop the remaining deletes, but they are counted and
+// reported as [ErrBlobDeleteIncomplete]: the caller goes on to remove the
+// metadata that names these objects, after which nothing can reach them
+// again.
 func deleteTxBlobs(d *Database, txHashes [][]byte, txn *Txn) error {
 	const batchSize = 500
 	blob := d.Blob()
@@ -1772,11 +1777,18 @@ func deleteTxBlobs(d *Database, txHashes [][]byte, txn *Txn) error {
 		}
 	}
 	if deleteErrors > 0 {
+		recordBlobOrphans(deleteErrors)
 		d.logger.Warn(
 			"TX blob deletion completed with errors",
 			"failed",
 			deleteErrors,
 			"total",
+			len(txHashes),
+		)
+		return fmt.Errorf(
+			"%w: %d of %d transaction blobs",
+			ErrBlobDeleteIncomplete,
+			deleteErrors,
 			len(txHashes),
 		)
 	}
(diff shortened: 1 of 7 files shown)


chrisguiney:
@coderabbitai review

database/transaction.go
```diff
@@ -1772,13 +1777,20 @@ func deleteTxBlobs(d *Database, txHashes [][]byte, txn *Txn) error {
 		}
 	}
 	if deleteErrors > 0 {
+		recordBlobOrphans(deleteErrors)
```
wolf31o2: `recordBlobOrphans` runs before the enclosing metadata transaction commits. If `TransactionsDeleteRolledback` later rolls that transaction back, the metadata row still names the blob, but this process-wide counter remains incremented; retrying can count the same reachable object again. The same ordering exists in `database/utxo.go`. Tie the accounting to durable metadata commit (for example with `Txn.AfterCommit`) and cover rollback plus retry so only committed metadata removal counts an orphan.

wolf31o2 (changes requested):

database/transaction.go
chrisguiney: Valid, fixed. `Txn.AfterCommit` documents itself for exactly this ("metrics
that must not count work a rollback discards"), so both deleters now register
the increment there instead of applying it at delete time.

The ordering you describe is the whole point of the counter, so counting it
wrong undercuts the metric: the blob delete precedes the metadata removal, and
until that removal is durable the row still names the blob and nothing is
stranded. A rollback now counts nothing and a retry does not double-count.

A nil transaction still counts immediately — there is no pending metadata
removal to gate on. Production callers all pass a transaction (they open one
themselves when handed nil), so that path is the direct-call one the tests use.

Added `BlobOrphansAreNotCountedUntilMetadataCommits` and
`BlobOrphansAreNotCountedOnRollback`, both confirmed failing against the
previous commit. DATABASE.md records the commit-gated accounting.

chrisguiney:
Addressed the review finding in a254a160.

The orphan count was applied at blob-delete time, which counted objects a
rollback leaves reachable and counted them again on the retry. Both deleters
now register the increment through `Txn.AfterCommit` on the enclosing
transaction, so only a durable metadata removal counts, and a nil transaction
(no pending metadata removal to gate on) still counts immediately.

Added `BlobOrphansAreNotCountedUntilMetadataCommits` and
`BlobOrphansAreNotCountedOnRollback`, both confirmed failing against afcca2ae.
`DATABASE.md` records the commit-gated accounting.

Revalidated on a254a160: `go test -race -count=1 ./database` pass (204s),
conformance pass, `golangci-lint`/`nilaway` on `./database/...` clean,
`make docs-parity` and `make import-boundaries` pass, `golines` clean.
DevNet was run on afcca2ae (exit 0, txpump 6/0/0); this commit changes only
when the counter is incremented, not any storage or consensus path.

wolf31o2:
@coderabbitai review

wolf31o2 (approved):

PR #3551, opened 2026-08-26
test: reduce default suite overhead

wolf31o2:
## Summary

- remove UTxO-RPC tests that exercise copied patterns instead of handlers
- retain deterministic SQLite planner and boundary coverage with smaller fixtures
- move live two-node forging lifecycle scenarios to `make test-live-lifecycle`
- disable setup-go caching where explicit Go cache steps already handle it

## Checks

- actionlint
- focused database tests with race detection
- UTxO-RPC package tests with race detection
- conformance tests
- docs parity
- diff and formatting checks

---
## Summary by cubic
Reduces default test suite overhead by removing low-value tests, moving live two-node lifecycle tests behind a build tag, and avoiding duplicate Go cache handling in CI.

**Test changes**
- Removes UTxO-RPC and SQLite tests that exercised copied patterns instead of real handlers or assertions.
- Replaces the SQLite cardinality sweeps with deterministic query-plan and boundary coverage using smaller fixtures.
- Adds the `make test-live-lifecycle` target to run the live two-node forging lifecycle tests under the `dingo_live_lifecycle_integration` build tag.
- Tightens the zero-arg branch test to seed one NULL-credential candidate per chunk and assert exact round-trip behavior.

**CI changes**
- Sets `cache: false` on `setup-go` in workflows that already use explicit `actions/cache` steps so Go caches are handled only once.

<sup>Written for commit a2f6773502d26cdd40576307382d7a0cc0b2f2cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3551?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Added a dedicated command for running the live two-node lifecycle integration test with race detection.
  * Improved UTxO exact-address coverage for large address sets and zero payment credentials.
  * Removed obsolete or unreliable integration, performance, and event-lifecycle tests.
  * Kept planner-based database index and recovery-path validation.

* **Chores**
  * Standardized Go workflow caching by using explicit cache steps consistently.


arepala-uml (approved):

Makefile
```diff
@@ -131,6 +131,9 @@ $(PROTOC):
 test: mod-tidy ## Run mod-tidy, then all tests with race detection
 	go test $(GO_TAG_FLAGS) -v -race -timeout 20m ./...
 
+test-live-lifecycle: ## Run the live two-node lifecycle integration tests with race detection
```
cryptodj413, line 134: Nothing in `.github/workflows/*.yml` passes `dingo_live_lifecycle_integration` or calls `make test-live-lifecycle` (checked all six workflow files, including `publish.yml`'s `-race ./...` run). `TestLiveTruncateUnderRealForgingAndNetworking` and `TestLiveRestoreUnderRealForgingAndNetworking` — the reason this target exists — no longer execute in any automated job. They used to build and run under plain `./...` before this PR, and per comments at `ledger/state.go:3730` and `ledger/state_test.go:4638` they previously caught real `-race` bugs (a `currentEra` race, a chain-index-gap corruption on `Close`) that no other test reproduces. Please wire this tag into a CI job (nightly is fine) rather than leaving it manual-only and undocumented in `AGENTS.md`/`CLAUDE.md`.

node_lifecycle_multinode_integration_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build dingo_live_lifecycle_integration
```
cryptodj413, line 1: See the `Makefile` comment on `test-live-lifecycle` — this file no longer compiles in any CI job once it's tagged. Separately, `ARCHITECTURE.md:1877` still references this file by its pre-rename name (`node_lifecycle_multinode_test.go`); please fix that in this PR.

database/utxo_exact_address_test.go
```diff
@@ -477,29 +481,41 @@ func TestUtxosByAddressExceedsSQLiteParameterLimit(t *testing.T) {
...
+			"fixture invariant: payment hash must be zero",
+		)
+		require.Equal(
+			t, zeroHash, addr.StakeKeyHash(),
+			"fixture invariant: staking hash must be zero",
+		)
+		addrs[i] = addr
+	}
+
+	want := seedExactAddressUtxo(t, db, raw, addrs[0], 1, 0x42)
```
cryptodj413, line 515: Seeding only `addrs[0]` means an over-aggressive dedup keyed on tx id alone, or a candidate owned by a chunk other than the first, wouldn't be caught here — `addrs[0]` is always in the first of the ~3 chunks 1000 patterns produce. Consider seeding 3 UTxOs (first, ~500th, last) across chunk boundaries and asserting with `ElementsMatch`, closer to the cross-chunk coverage the removed 2000-address version had.

cryptodj413 (dismissed):
Reviewed with the protocol, app, and release auditors against the current head (`a2f6773`). Bot review: CodeRabbit completed with no actionable comments; Cubic reports its monthly review-line limit reached (not rate-limited in the usual sense, but effectively unavailable) — CodeRabbit's completed pass stands in for bot review.

**Requesting changes:**
- No CI job runs `dingo_live_lifecycle_integration` (see inline comment on `Makefile`). The two tests gated behind it previously caught real `-race` bugs (`ledger/state.go:3730`, `ledger/state_test.go:4638` reference them by name) and nothing else in the repo — not DevNet, not conformance — covers live truncate/restore under real forging plus real networking. Cost to run is ~14s with `-race`, so this doesn't need to stay out of CI entirely.
- `ARCHITECTURE.md:1877` cites `node_lifecycle_multinode_test.go`, the pre-rename filename. Please update it in this PR since the rename causes the drift.

**Verified fine, no action needed:**
- All 7 deleted `api/utxorpc` tests reimplemented the logic locally (bare `EventBus` calls, or a table asserting `>` works) rather than calling the real handler; the coverage profile for the package is byte-identical before/after. Reintroducing the U1 comparison bug they were named for shows the *surviving* `TestConnect_WatchMempool_StreamsOnAddTransactionEvent` catches it, not the deleted tests.
- Deleting `TestTwoNodeDevnetSyncOverRealNetworking` is fine — both tests remaining in the same file open with identical setup and a strictly stronger tip-slot assertion.
- The SQLite wall-clock sweep removal is a sound trade: the deterministic `EXPLAIN QUERY PLAN` tests it cites as already proving the indexed-search invariant remain in place.
- `cache: false` on `setup-go` is correctly paired with an existing `actions/cache` step everywhere it's added; no job loses caching.

**Non-blocking recommendations:**
- `TestSetTransactionCostFlatAfterDeferredIndexDrop`'s class-level invariant (every predicate `SetTransaction` filters on stays indexed, not just the witness-cleanup deletes) loses its only check with no deterministic replacement. Worth a follow-up `EXPLAIN QUERY PLAN` sweep over `SetTransaction`.
- See inline comment on `database/utxo_exact_address_test.go`.
- `FetchBlock`/`ReadUtxos`/`ReadData` request-limit rejection branches (`api/utxorpc/sync.go:54`, `query.go:448`, `query.go:721`) are uncovered by any test — pre-existing, not caused by this PR, but worth a follow-up issue.
- `publish.yml`'s tag-only `ci` job (setup-go at line 91) is the only instance the cache dedup didn't touch, and the only one without an explicit `actions/cache` step — confirm that's intentional rather than missed.

No `DATABASE.md`/`ARCHITECTURE.md` schema or component-boundary changes are needed here beyond the stale filename reference above.

PR #3558, opened 2026-08-26
fix(node): bound the component stops in a live restore or truncate

chrisguiney:
Refs #1649

`quiesceForLiveLifecycleOp` stopped the block forger, leader election, both
Leios managers and the snapshot manager with no deadline of any kind. Each of
those `Stop()` calls cancels its own context and then waits on a
`sync.WaitGroup`, so a goroutine that never observes its cancellation wedges
the entire live restore or truncate — indefinitely, well past the configured
shutdown timeout, with no error the caller can act on. The koios parity
observer immediately after them was already bounded, so this was the remaining
gap on that path.

Each stop now gets the configured shutdown timeout via `stopWithDeadline`.

An unfinished wait escalates to `errStorageDrainUnconfirmed` rather than being
reported as an ordinary stop failure, which is what the existing machinery
already does for `connManager.Stop`, the leios persist writer's drain, and the
storage providers. The distinction matters for what happens next: a component
that *returns* an error has stopped, so `Restore`/`Truncate` can resume on the
untouched data directory, while one that never returned may still be reading or
writing `n.db` — so those callers must abandon the operation and force a
supervised restart instead of reopening storage underneath it. That routing
already exists; this change just feeds it.

Two deliberate choices:

- The abandoned goroutine is left running. It cannot be interrupted from there,
  and the escalation brings the process down anyway, so leaking it for that
  window is strictly safer than proceeding to close the database it may still
  be using.
- The caller's context does **not** shorten the wait. Cancelling a restore must
  not escalate a component that would have stopped cleanly into a supervised
  restart, so the deadline alone decides. `TestStopWithDeadlineIgnoresCallerCancellation`
  pins that.

Documentation: `ARCHITECTURE.md`'s `errStorageDrainUnconfirmed` source list
named two callers and is updated to match the code (it also omitted the storage
providers' deadline path). `DATABASE.md` checked, not affected.

Validation:

- `go test -race -count=1 .` — pass, exit 0
- `go test -count=1 ./internal/test/conformance` — pass
- `./internal/test/devnet/run-tests.sh` (all-Dingo) — pass, exit 0
  (txpump submitted=9 rejected=0 error=0)
- `golangci-lint run .`, `nilaway .` — clean
- `make import-boundaries`, `make docs-parity` — pass
- `golines` — changed files clean

`TestStopWithDeadlineEscalatesAnUnfinishedStop` was confirmed failing under a
mutation that drops the escalation, so it pins the escalation rather than just
the timeout.

---
## Summary by cubic
Binds the component stops in `quiesceForLiveLifecycleOp` to the configured shutdown timeout, so a goroutine that never observes its cancellation can no longer wedge a live restore or truncate past that deadline with no error to act on (Refs #1649).

- The block forger, leader election, Leios pipeline manager, Leios vote manager, snapshot manager, and database lifecycle manager stops now run via `stopWithDeadline`.
- The bounded stops are grouped in one ordered `quiesceComponentStops` list, and quiesce-level tests drive the production path with a stop that blocks until released.
- An unfinished wait escalates to `errStorageDrainUnconfirmed`, routing `Restore`/`Truncate` to a supervised restart instead of reopening storage a still-running goroutine may be using.
- The abandoned goroutine is left running, and the caller's context does not shorten the wait.
- `ARCHITECTURE.md`'s `errStorageDrainUnconfirmed` source list is updated; new tests pin the escalation, caller-cancellation behavior, and the bounded stop set.

<sup>Written for commit 15d35367e0ab343db2db7224b947293b3e13b9dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3558?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


node_lifecycle.go
```diff
@@ -106,37 +106,100 @@ import (
...
 func (n *Node) quiesceForLiveLifecycleOp(ctx context.Context) error {
 	var err error
 
+	// Each of these Stop calls cancels its own context and then waits on a
+	// WaitGroup that has no deadline of its own, so every one is bounded here
+	// instead -- see stopWithDeadline for why an unfinished wait escalates to
+	// errStorageDrainUnconfirmed rather than being reported as an ordinary
+	// stop failure. The budget is the configured shutdown timeout, the same
+	// one the koios parity observer below already uses.
+	stopTimeout := n.configuredShutdownTimeout()
```
wolf31o2, line 239: `dbLifecycleMgr.Stop()` remains an unbounded cancel-plus-WaitGroup wait. `internal/dblifecycle/manager_test.go:681-812` blocks an upload that ignores cancellation and proves `Stop` cannot return until release. A live Restore/Truncate can therefore still exceed `shutdownTimeout` here without reaching the drain-unconfirmed escalation. Please apply the same deadline/escalation handling to this stop.

quiesce_deadline_test.go
```diff
@@ -0,0 +1,108 @@
...
+// may still be reading or writing n.db, so Restore/Truncate must abandon the
+// operation and force a supervised restart rather than reopen storage
+// underneath it.
+func TestStopWithDeadlineEscalatesAnUnfinishedStop(t *testing.T) {
+	t.Parallel()
+
+	release := make(chan struct{})
+	t.Cleanup(func() { close(release) })
+
+	err := stopWithDeadline(
```
wolf31o2, line 76: These tests exercise only `stopWithDeadline`, not `quiesceForLiveLifecycleOp`. Reverting all five production call sites to their prior direct `Stop` calls leaves this suite passing, so it does not fail without the fix. Add a quiesce-level test with a releasable blocking stop seam that proves the production path is wired through the deadline/escalation logic.

wolf31o2 (changes requested):

chrisguiney:
@coderabbitai review

chrisguiney:
Both findings addressed in 15d35367.

`dbLifecycleMgr.Stop` is bounded with the rest. It is the same
cancel-plus-WaitGroup wait and was still called directly, so it could exceed
`shutdownTimeout` without reaching the escalation.

On the coverage gap: the bounded stops are now one ordered list,
`quiesceComponentStops`, rather than six inline calls, so the set is visible in
one place and assertable. `componentStopsForQuiesce` is indirected through a
variable — the same pattern `syncDataDirParent` uses for fsync failures — so a
test can inject a stop that blocks until released and drive the real
`quiesceForLiveLifecycleOp`. None of these components can be made to block from
outside, so without that seam the escalation path could only be exercised below
quiesce, which is exactly the gap you identified.

Verified against two mutations rather than asserted:

- dropping `dbLifecycleMgr` from the list fails
  `TestQuiesceComponentStopsCoverEveryUnboundedStop`
- replacing the loop body with a direct `cs.stop()` fails
  `TestQuiesceEscalatesAStopThatNeverReturns`

That second test initially hung under its mutation instead of failing, so it
now runs quiesce in a goroutine bounded at 5s and fails with a message naming
the cause.

Revalidated on 15d35367: `go test -race -count=1 .` pass, DevNet exit 0
(txpump 9/0/0), conformance pass, `golangci-lint`/`nilaway` clean,
`make import-boundaries` and `make docs-parity` pass, `golines` clean.
`ARCHITECTURE.md` names the sixth component and the enumeration.

chrisguiney:
Bot review status, recorded rather than assumed: CodeRabbit has not reviewed
this PR. Its check reports pass, but that is the rate-limit path — it posted
"Review limit reached" on ce46757a, and an explicit `@coderabbitai review`
after the window reset returned "Review rate limited" plus a note that the
command only applies when automatic reviews are paused. There is no CodeRabbit
review against any commit here. Cubic skips this repository, so bot review is
incomplete for now; a later automatic pass on 15d35367 may still land.

Recording it because a green CodeRabbit check on this repo can mean reviewed,
skipped, or rate limited, and the three are not distinguishable from the check
alone.

wolf31o2 (approved):

PR #3559, opened 2026-08-26
fix(ledger): seed post-Byron epoch nonce at multi-era boundary transitions

wcatz:
## Summary
- `calculateEpochNonce` short-circuits to an all-nil nonce whenever the epoch rollover's *source* era is Byron (no Praos nonce there), even when `applyBoundaryEraTransitions` has just advanced the epoch's actual (destination) era past Byron in that same rollover.
- This is correct when the new epoch stays in Byron, but wrong the one time it doesn't: the new era has no nonce, so header verification permanently rejects every block of the new era with "epoch has no nonce for slot" — a hard stall at the first epoch boundary of any Byron-prefixed network synced from genesis instead of a Mithril snapshot.
- Fix: seed the epoch's nonce/evolving-nonce/candidate-nonce from the Shelley genesis hash in `applyBoundaryEraTransitions` when `workingEraId != 0 && len(newEpoch.Nonce) == 0`, the same way every other from-genesis nonce path already does. `LastEpochBlockNonce` is left at `NeutralNonce` (nil).

## Test plan
- [x] Reproduced against a from-genesis sync of Apex Fusion's `prime-mainnet` (a Byron-prefixed network with no Mithril snapshot available): stalled permanently at the first epoch boundary (slot 43200) with the bug present.
- [x] Validated the fix over two independent fresh from-genesis sync runs: the epoch cache advanced cleanly through 8+ sequential epoch rollovers, each with a distinct, correctly-evolving nonce.
- [x] `go test -tags dingo_extra_plugins -timeout 20m ./ledger/...` passes.

---
## Summary by cubic
Fixes a hard stall at the first epoch boundary when a Byron-prefixed network syncs from genesis without a Mithril snapshot. Previously the epoch nonce was left nil whenever the rollover's source era was Byron, even if the boundary transitioned into a later era, so header verification rejected every block of the new era. Now `applyBoundaryEraTransitions` seeds the nonce from the Shelley genesis hash.

- Nonce, evolving nonce, and candidate nonce all get the Shelley genesis hash; `LastEpochBlockNonce` stays nil.
- The same values are also written to the new epoch cache entries.
- Seeding now errors instead of panicking on a nil `CardanoNodeConfig` or silently writing an unusable nonce from an empty `ShelleyGenesisHash`; the nil-config guard also protects the top of `processEpochRollover` before the config is otherwise dereferenced.
- Validated on `prime-mainnet` from genesis: previously wedged at slot 43200, now syncs through 8+ rollovers; `go test -tags dingo_extra_plugins -timeout 20m ./ledger/...` passes. A new regression test asserts the seeded nonce in the in-memory epoch, the epoch cache, and the persisted database row.

<sup>Written for commit 4447a847505ef0c28dd6d3845502048b33d2a1b3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3559?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved epoch transitions after Byron to correctly initialize required nonce values from the Shelley genesis hash.
  * Preserved the absence of a last-epoch block nonce when appropriate.
  * Updated transitioned epoch state to retain all relevant nonce information.
  * Clarified handling of Byron-origin epoch rollovers.

### DIFF
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -3475,6 +3475,34 @@ func (ls *LedgerState) applyBoundaryEraTransitions(
 	newEpoch.EraId = workingEraId
 	newEpoch.SlotLength = slotLength
 	newEpoch.LengthInSlots = epochLength
+	// calculateEpochNonce short-circuits to an all-nil nonce whenever the
+	// ROLLOVER'S SOURCE era is Byron (no Praos nonce there), regardless of
+	// what era this boundary actually transitions into. That is correct
+	// when the new epoch stays in Byron, but here workingEraId has just
+	// been advanced past Byron (that is the only way this function runs),
+	// so the epoch needs a real nonce for header verification in its new
+	// era. Seed it the same way every other from-genesis nonce path does
+	// (computeEpochNonceForSlot, calculateEpochNonce's own no-prior-nonce
+	// branch): the Shelley genesis hash for nonce/evolving/candidate, with
+	// LastEpochBlockNonce left at NeutralNonce (nil) — see #2734. Without
+	// this, header verification permanently rejects every block in the
+	// new era with "epoch has no nonce for slot" for any Byron-prefixed
+	// network that syncs from genesis instead of a Mithril snapshot.
+	if workingEraId != 0 && len(newEpoch.Nonce) == 0 {
+		genesisHashBytes, err := hex.DecodeString(
+			ls.config.CardanoNodeConfig.ShelleyGenesisHash,
+		)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"decode Shelley genesis hash for post-Byron epoch nonce: %w",
+				err,
+			)
+		}
+		newEpoch.Nonce = genesisHashBytes
+		newEpoch.EvolvingNonce = genesisHashBytes
+		newEpoch.CandidateNonce = genesisHashBytes
+		newEpoch.LastEpochBlockNonce = nil
+	}
 	if err := ls.db.SetEpoch(
 		newEpoch.StartSlot,
 		newEpoch.EpochId,
@@ -3515,6 +3543,10 @@ func (ls *LedgerState) applyBoundaryEraTransitions(
 			rolloverResult.NewEpochCache[i].EraId = workingEraId
 			rolloverResult.NewEpochCache[i].SlotLength = slotLength
 			rolloverResult.NewEpochCache[i].LengthInSlots = epochLength
+			rolloverResult.NewEpochCache[i].Nonce = newEpoch.Nonce
+			rolloverResult.NewEpochCache[i].EvolvingNonce = newEpoch.EvolvingNonce
+			rolloverResult.NewEpochCache[i].CandidateNonce = newEpoch.CandidateNonce
+			rolloverResult.NewEpochCache[i].LastEpochBlockNonce = newEpoch.LastEpochBlockNonce
 		}
 	}
 
--- a/ledger/state.go
+++ b/ledger/state.go
@@ -3489,6 +3489,16 @@ func (ls *LedgerState) applyBoundaryEraTransitions(
 	// new era with "epoch has no nonce for slot" for any Byron-prefixed
 	// network that syncs from genesis instead of a Mithril snapshot.
 	if workingEraId != 0 && len(newEpoch.Nonce) == 0 {
+		if ls.config.CardanoNodeConfig == nil {
+			return nil, errors.New(
+				"seed post-Byron epoch nonce: CardanoNodeConfig is nil",
+			)
+		}
+		if ls.config.CardanoNodeConfig.ShelleyGenesisHash == "" {
+			return nil, errors.New(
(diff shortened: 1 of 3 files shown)


ledger/state.go
```diff
@@ -3475,6 +3475,50 @@ func (ls *LedgerState) applyBoundaryEraTransitions(
...
+	// so the epoch needs a real nonce for header verification in its new
+	// era. Seed it the same way every other from-genesis nonce path does
+	// (computeEpochNonceForSlot, calculateEpochNonce's own no-prior-nonce
+	// branch): the Shelley genesis hash for nonce/evolving/candidate, with
+	// LastEpochBlockNonce left at NeutralNonce (nil) — see #2734. Without
+	// this, header verification permanently rejects every block in the
+	// new era with "epoch has no nonce for slot" for any Byron-prefixed
+	// network that syncs from genesis instead of a Mithril snapshot.
+	if workingEraId != 0 && len(newEpoch.Nonce) == 0 {
+		if ls.config.CardanoNodeConfig == nil {
```
wolf31o2, line 3492: This nil check is too late to protect the production boundary path. `processEpochRollover` dereferences `ls.config.CardanoNodeConfig` at `ledger/chainsync.go:5080` before `applyBoundaryEraTransitions` runs, so a config accepted by `NewLedgerState` still panics instead of returning this error. Validate before rollover, or require the config at construction, so this fails closed.

ledger/state.go
```diff
@@ -3475,6 +3475,50 @@ func (ls *LedgerState) applyBoundaryEraTransitions(
...
+	// been advanced past Byron (that is the only way this function runs),
+	// so the epoch needs a real nonce for header verification in its new
+	// era. Seed it the same way every other from-genesis nonce path does
+	// (computeEpochNonceForSlot, calculateEpochNonce's own no-prior-nonce
+	// branch): the Shelley genesis hash for nonce/evolving/candidate, with
+	// LastEpochBlockNonce left at NeutralNonce (nil) — see #2734. Without
+	// this, header verification permanently rejects every block in the
+	// new era with "epoch has no nonce for slot" for any Byron-prefixed
+	// network that syncs from genesis instead of a Mithril snapshot.
+	if workingEraId != 0 && len(newEpoch.Nonce) == 0 {
```
wolf31o2, line 3491: The follow-up still adds no checked-in regression. Add a deterministic Byron→Shelley production-path test covering the in-memory epoch, epoch cache, and persisted epoch row; it must fail when this nonce-seeding block is removed. The existing boundary test still passes with the block deleted.

wolf31o2 (changes requested):

ledger/state.go
wcatz: Fixed in 4447a847: added a nil CardanoNodeConfig check at the top of processEpochRollover (ledger/chainsync.go), before any of the ShelleyGenesis()/applyBoundaryEraTransitions dereferences. This fails closed on the actual production boundary path rather than only in applyBoundaryEraTransitions.

ledger/state.go
wcatz: Fixed in 4447a847: added TestByronShelleyBoundarySeedsEpochNonceOnProductionPath, which runs the same production path (ledgerProcessBlocksFromSource) as the existing boundary test but asserts on the seeded nonce itself in all three places it's observable -- the in-memory current epoch, the epoch cache, and the persisted database row. Verified locally that it fails without the nonce-seeding block in applyBoundaryEraTransitions and passes with it restored.

wolf31o2 (approved):
Approving. Both findings from the previous review are addressed and verified.

**Fails closed on the production path.** The nil check now sits at the top of `processEpochRollover` (`ledger/chainsync.go:4916`), ahead of the first `ls.config.CardanoNodeConfig` dereference at 4944, so a config that `NewLedgerState` accepts returns an error rather than panicking part-way through a rollover.

**The regression is real this time.** Neutering the seeding block — `if false && workingEraId != 0 && len(newEpoch.Nonce) == 0` — fails `TestByronShelleyBoundarySeedsEpochNonceOnProductionPath` at all three assertion points: the in-memory epoch, the epoch cache, and the persisted row. That is the property the earlier boundary test could not distinguish, since it passed with the block deleted.

Checks at 4447a847: `go test ./ledger/ -run TestByronShelleyBoundary` green, `golangci-lint run ./ledger/` 0 issues, CI green.

Non-blocking: the new nil-config guard itself has no test. It is a simple fail-closed branch and the behaviour it protects is now covered, so this is a note rather than a request — worth a case only if that function grows more config-dependent paths.

PR #3560, opened 2026-08-26
fix(ledger): distinguish prime-mainnet from real Cardano mainnet in isMainnet

wcatz:
## Summary
- Apex Fusion's `prime-mainnet` declares Shelley genesis `networkId=Mainnet` and network magic `764824073` — byte-identical to real Cardano mainnet's genesis identity. gouroboros's own network registry documents this collision directly (`NetworkByNetworkMagic` returns `NetworkCardanoMainnet`, not `NetworkPrimeMainnet`, for that magic).
- `isMainnet()` used genesis `networkId` alone as the mainnet predicate, so `prime-mainnet` inherited cardano-ledger's mainnet-only BBODY header-protocol-version strictness. That strictness rejects a header whose advertised protocol major version is more than one ahead of dingo's enacted-pparams tracking. In the two boundary-crossing headers observed while investigating this, `prime-mainnet`'s block producers advertised a version 2 ahead of dingo's enacted tracking (e.g. header major 7 vs. enacted major 5 at the epoch 9/10 boundary) — a gap real testnets tolerate via the pre-Dijkstra relaxation from cardano-ledger PR 5785, but real mainnet does not. This wedged sync deterministically at slot 3,499,200, rejecting a header the real network had already built on.
- Fix: thread the CLI/YAML/env network selector dingo was started with (`LedgerStateConfig.Network`, sourced from `cfg.Network`) through to `isMainnet()`. When it resolves to a network gouroboros recognizes, that resolved identity — not the raw genesis `networkId` — decides mainnet-ness, so `prime-mainnet` gets the same relaxed treatment as any other non-canonical network despite its Mainnet-shaped genesis. An unset or unrecognized `Network` value falls back to the prior genesis-only check, so behavior is unchanged for every existing caller (including real Cardano mainnet/preprod/preview/sanchonet and embedders that never set the field).

## Test plan
- [x] Reproduced against a from-genesis sync of `prime-mainnet`: deterministic wedge at slot 3,499,200 across two independent runs, confirmed via debug instrumentation showing `isMainnet=true` at the point of rejection.
- [x] Validated the fix: rebuilt and resumed the same wedged database — sync immediately proceeded past slot 3,499,200 (`isMainnet=false` observed) and continued advancing cleanly for 15M+ slots / 120k+ blocks with zero pipeline-stuck errors.
- [x] Added `TestLedgerStateIsMainnet_PrimeMainnetNotRealMainnet`, `TestLedgerStateIsMainnet_NamedMainnetStillTrue`, `TestLedgerStateIsMainnet_UnknownNetworkNameFallsBackToGenesis`.
- [x] `go test -tags dingo_extra_plugins -timeout 20m ./ledger/...` passes.
- [x] `go test -tags dingo_extra_plugins -timeout 20m ./internal/node/...` passes.
- [x] `go test ./internal/architecture` (import-boundaries) passes.
- [x] `go test ./internal/docsparity` passes.

## Note on the pre-Dijkstra relaxation

This fix makes `prime-mainnet` rely on the same pre-Dijkstra (`curPvMajor < 12`) relaxation real testnets use. Only two boundary-crossing headers were actually observed during this investigation (both roughly 2 majors ahead of enacted state, both well below Dijkstra) — not enough to call that gap a confirmed, sustained pattern, and no basis for predicting whether or how `prime-mainnet`'s protocol-version numbering behaves as it approaches Dijkstra-range versions. Not filing a follow-up issue on that speculation; revisit if and when sync actually gets there.

---
## Summary by cubic
Fixes a deterministic sync wedge on `prime-mainnet` by distinguishing it from real Cardano mainnet in `isMainnet()`.

`prime-mainnet` declares a Shelley genesis networkId and network magic byte-identical to real Cardano mainnet's, so the old genesis-only check treated it as mainnet and applied strict header protocol version rules, rejecting blocks its own producers had already built. `isMainnet()` now also consults the network selector dingo was started with (`LedgerStateConfig.Network`, sourced from `cfg.Network`), but only lets a named network override a Mainnet-tagged genesis when the loaded genesis magic matches real mainnet's and gouroboros registers that name as sharing mainnet's magic — currently only `prime-mainnet`. Any other name (e.g. `preview` paired with a Mainnet genesis) is a configuration mismatch and falls through to the genesis-only answer, as does an unset or unrecognized network, so existing callers are unaffected.

**Bug Fixes**
- Passes `cfg.Network` from `LoadWithDB` and `n.config.network` from the ledger config wiring into `LedgerStateConfig.Network`.
- Resolves mainnet-ness via a magic-matching named network only; everything else falls back to the genesis-only check.

**Refactors**
- Threads the network selector through `LedgerStateConfig` so `isMainnet()` can disambiguate chains with identical genesis identity.

**Dependencies**
- Adds `github.com/blinklabs-io/gouroboros` import to `ledger/header_protocol_version.go` for network identity resolution.

<sup>Written for commit 3c936e1e9b85a9267dd78c49c3293660c9d50baf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3560?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network identification when configured networks share the same network magic.
  * Prime-mainnet is no longer incorrectly treated as Cardano mainnet.
  * Preserved strict mainnet detection for explicit, mismatched, unknown, or unset network configurations.
* **Compatibility**
  * Network configuration is now consistently applied when loading ledger state, improving behavior across supported networks.

### DIFF
--- a/ledger/header_protocol_version.go
+++ b/ledger/header_protocol_version.go
@@ -153,13 +153,20 @@ func ValidateHeaderProtocolVersion(
 // cardano-ledger's mainnet-only BBODY strictness there rejects headers a
 // chain's own nodes have already built on top of.
 //
-// When ls.config.Network names a network gouroboros recognizes, that
-// resolved identity decides mainnet-ness instead of the raw networkId
-// value, so prime-mainnet correctly gets the same relaxed treatment as
-// any other testnet despite its Mainnet-shaped genesis. An unset or
-// unrecognized Network name (a config built from a raw NetworkMagic, or
-// an embedder that never set it) falls back to the networkId-only check,
-// preserving prior behavior for every caller that predates this field.
+// A Mainnet-tagged genesis is only overridden to non-mainnet when
+// ls.config.Network names a network gouroboros itself registers as
+// sharing mainnet's network magic (currently just prime-mainnet) — the
+// narrow, known identity-reuse case this function exists to handle. Any
+// other named network paired with a Mainnet-tagged genesis is a
+// configuration mismatch, not a recognized alias, and falls through to
+// the genesis-only answer (true) rather than silently relaxing the BBODY
+// check for a misconfigured node: a node started with, say, Network=
+// "preview" against a genesis that (incorrectly) declares Mainnet must
+// not have mainnet's strictness relaxed just because the name isn't
+// literally "mainnet". An unset or unrecognized Network name (a config
+// built from a raw NetworkMagic, or an embedder that never set it) falls
+// back to the networkId-only check, preserving prior behavior for every
+// caller that predates this field.
 //
 // Fails closed: when CardanoNodeConfig or Shelley genesis is missing,
 // or the networkId field carries an unrecognized value, returns
@@ -178,8 +185,18 @@ func (ls *LedgerState) isMainnet() (bool, error) {
 	switch sg.NetworkId {
 	case shelleyGenesisMainnetNetworkId:
 		if ls.config.Network != "" {
-			if known, ok := ouroboros.NetworkByName(ls.config.Network); ok {
-				return known.Name == ouroboros.NetworkCardanoMainnet.Name, nil
+			if known, ok := ouroboros.NetworkByName(ls.config.Network); ok &&
+				known.Name != ouroboros.NetworkCardanoMainnet.Name &&
+				known.NetworkMagic == ouroboros.NetworkCardanoMainnet.NetworkMagic {
+				// Only a network gouroboros itself registers as sharing
+				// mainnet's magic (currently just prime-mainnet) may
+				// override a Mainnet-tagged genesis to non-mainnet. Any
+				// other named network paired with a Mainnet-tagged genesis
+				// is a configuration mismatch, not a known identity-reuse
+				// case, so it falls through to the genesis-only answer
+				// (true) rather than silently relaxing the BBODY check for
+				// a misconfigured node.
+				return false, nil
 			}
 		}
 		return true, nil
--- a/ledger/header_protocol_version.go
+++ b/ledger/header_protocol_version.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 
+	ouroboros "github.com/blinklabs-io/gouroboros"
 	"github.com/blinklabs-io/gouroboros/ledger/allegra"
 	"github.com/blinklabs-io/gouroboros/ledger/alonzo"
 	"github.com/blinklabs-io/gouroboros/ledger/babbage"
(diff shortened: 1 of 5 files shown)


wcatz:
@coderabbitai review

ledger/header_protocol_version.go
```diff
@@ -162,6 +177,11 @@ func (ls *LedgerState) isMainnet() (bool, error) {
 	}
 	switch sg.NetworkId {
 	case shelleyGenesisMainnetNetworkId:
+		if ls.config.Network != "" {
+			if known, ok := ouroboros.NetworkByName(ls.config.Network); ok {
```
wolf31o2: Restrict this override to `prime-mainnet`, or reject a named-network/genesis mismatch. Any registered non-mainnet name currently changes a Mainnet-tagged genesis into a testnet for this consensus rule, while startup only validates matching magic. For example, `preview` plus a Mainnet genesis using magic 2 passes configuration and makes the production validator accept header PV 12 over current PV 10. This weakens the upstream `netId == Mainnet` rule beyond the intended prime-mainnet exception.

wolf31o2 (changes requested):

ledger/header_protocol_version.go
wcatz: Good catch, fixed in 94eafde8: restricted the override to networks gouroboros registers as sharing mainnet's network magic (currently only prime-mainnet) instead of any name != "mainnet". A Network=preview + Mainnet-tagged-genesis mismatch now falls through to the genesis-only answer (true) rather than being silently relaxed. Added a regression test for that exact scenario (TestLedgerStateIsMainnet_NamedNetworkMagicMismatchStaysMainnet).

wolf31o2:
@coderabbitai review

ledger/header_protocol_version.go
```diff
@@ -162,6 +184,21 @@ func (ls *LedgerState) isMainnet() (bool, error) {
 	}
 	switch sg.NetworkId {
 	case shelleyGenesisMainnetNetworkId:
+		if ls.config.Network != "" {
+			if known, ok := ouroboros.NetworkByName(ls.config.Network); ok &&
```
wolf31o2, line 195: Also require the loaded Shelley genesis magic to match the recognized network/canonical mainnet magic before applying this override. `prime-mainnet` with explicit magic 999 and a Mainnet-tagged genesis using magic 999 passes startup validation, but this clause ignores `sg.NetworkMagic` and still disables the BBODY rule. Add this cross-product to the production-validator regression.

wolf31o2 (changes requested):

ledger/header_protocol_version.go
wcatz: Good catch, fixed in 3c936e1e: the override now also requires sg.NetworkMagic (the genesis actually loaded) to equal real mainnet's magic, not just the registry's canonical magic for the configured name. Added TestLedgerStateIsMainnet_PrimeMainnetNameWithMismatchedGenesisMagicStaysMainnet for the exact prime-mainnet-name/mismatched-genesis-magic scenario.

wolf31o2 (approved):
Approving. Both findings are fixed, and each new guard is independently load-bearing.

Removing the loaded-genesis-magic condition fails `TestLedgerStateIsMainnet_PrimeMainnetNameWithMismatchedGenesisMagicStaysMainnet`; removing the registry magic-match condition fails `TestLedgerStateIsMainnet_NamedNetworkMagicMismatchStaysMainnet`. Neither test fails for the other's revert, so they pin distinct properties.

I also checked the thing these guards could plausibly have broken — that the override still fires for the case it exists for. `NetworkPrimeMainnet` carries `NetworkMagic: 764824073`, deliberately equal to mainnet's, so both magic conditions hold for a genuine prime-mainnet node and `TestLedgerStateIsMainnet_PrimeMainnetNotRealMainnet` still passes. Had prime-mainnet used its own magic, the tightened condition would have made the whole exception unreachable.

Using `NetworkByName` rather than `NetworkByNetworkMagic` is also the right lookup here: gouroboros documents that the magic-based one returns `NetworkCardanoMainnet` and never `NetworkPrimeMainnet` for 764824073, so a magic lookup could not distinguish the two.

Checks at 3c936e1e: `go test ./ledger/ -run TestLedgerStateIsMainnet` green, CI green.

PR #3569, opened 2026-08-26
fix(bark): enforce request bounds and operator authorization

### LINKED ISSUE #3499: bark: align request handling with the current service contract
The Bark integration does not match the current service contract: read-only
DatabaseService calls bypass second-stage handling, and `FetchBlockRequest`
accepts an unbounded block list without a decoded request-size limit.

Acceptance criteria:

- Apply the current two-stage handling to all DatabaseService methods.
- Bound repeated block counts and decoded request size before processing.
- Cover read-only calls, compressed requests, empty lists, and oversized lists.

wolf31o2:
## Summary

- bound compressed, decoded, and response data for Bark Connect handlers
- cap FetchBlock references before database access
- require CA-verified mTLS and operator authorization for DatabaseService
- wire operator fingerprints through CLI/config composition and document the contract

## Validation

- exact-parent request-limit and authorization regressions
- production config-composition red/green regression
- tagged Bark/config/node race suites and conformance
- docs parity, architecture boundaries, vet, lint, and builds

Closes #3499

---
## Summary by cubic
Fixes #3499 by bounding all Bark Connect requests and adding explicit operator authorization to Bark's `DatabaseService`. Previously, read-only DatabaseService RPCs worked without a client certificate and any verified certificate could run destructive RPCs; now every DatabaseService RPC requires a verified mTLS certificate, and destructive RPCs also require a SHA-256 certificate fingerprint from the new `barkOperatorCertificateFingerprints` allowlist.

**Request bounds**
- All Connect handlers (Archive, DatabaseService, health, reflection) enforce 1 MiB compressed and decompressed read limits plus a 1 MiB per-message send limit.
- `FetchBlock` rejects requests with 0 or more than 100 block references before acquiring the database.
- The HTTP server adds a 60-second read timeout without a write timeout, so `StreamOperationProgress` and other long-lived streams are unaffected.

**Authorization**
- Every DatabaseService RPC now requires a verified client certificate; read-only RPCs previously worked anonymously, and an unverified cert on a read-only RPC was previously accepted.
- Destructive RPCs additionally require a certificate fingerprint from `barkOperatorCertificateFingerprints`, configured via YAML, `DINGO_BARK_OPERATOR_CERTIFICATE_FINGERPRINTS`, or `--bark-operator-certificate-fingerprints`; a verified cert not on the list receives `PERMISSION_DENIED`.
- Fingerprint values are case-insensitive hexadecimal and may contain colons.
- `Start` and config validation refuse to mount DatabaseService without both a client CA and at least one operator fingerprint, so existing deployments must add the new setting.
- Procedures absent from both classification maps are treated as destructive and require operator authorization.
- The public `ArchiveService` still requires no certificate.

<sup>Written for commit 449caadf85cf29f00ebd52587bf6487a720ea3fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3569?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security**
  * Added mandatory mTLS authentication for all DatabaseService operations.
  * Restricted destructive database operations to approved client certificate fingerprints.
  * Added configuration options and validation for client CA certificates and operator fingerprints.

* **Reliability**
  * Added request and response size limits, request timeouts, and FetchBlock batch-size validation.

* **Documentation**
  * Updated setup guidance, environment variables, and security behavior for Bark services.

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,9 +620,13 @@ a target beyond the security parameter, because it exists for disaster-recovery
 scenarios (see CIP-0135) where the chain must be rewound further than Ouroboros
 Praos allows. The resulting database is resync-ready from the target point.
 
-The same operations are also exposed remotely through the Bark
-`DatabaseService`. Bark has no built-in authentication, so do not expose its
-port outside a trusted network.
+The same operations are also exposed remotely through Bark's
+`DatabaseService`. Every `DatabaseService` RPC requires a client certificate
+verified against `barkClientCaFilePath`; destructive RPCs also require the
+certificate's SHA-256 fingerprint in
+`barkOperatorCertificateFingerprints`. Bark's read-only `ArchiveService`
+remains public on the same listener, so expose the Bark port only to the
+intended network.
 
 ## Database Plugins
 
--- a/README.md
+++ b/README.md
@@ -102,6 +102,11 @@ The following environment variables modify Dingo's behavior:
   - Comma-separated HTTPS hostnames additionally allowed for Bark-supplied
     block download URLs. The allowlist always includes the
     `DINGO_BARK_BASE_URL` hostname.
+- `DINGO_BARK_CLIENT_CA_FILE_PATH`
+  - PEM CA bundle used to authenticate every Bark DatabaseService caller.
+- `DINGO_BARK_OPERATOR_CERTIFICATE_FINGERPRINTS`
+  - Comma-separated SHA-256 client certificate fingerprints authorized for
+    destructive Bark DatabaseService RPCs.
 - `DINGO_DEBUG_BIND_ADDR`
   - IP address to bind for unauthenticated pprof endpoints (default:
     `127.0.0.1`)
--- a/bark/archive.go
+++ b/bark/archive.go
@@ -17,6 +17,7 @@ package bark
 import (
 	"context"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"math"
 
@@ -39,6 +40,23 @@ func (a *archiveServiceHandler) FetchBlock(
 	req *connect.Request[archive.FetchBlockRequest],
 ) (*connect.Response[archive.FetchBlockResponse], error) {
 	resp := &archive.FetchBlockResponse{}
+	blocks := req.Msg.GetBlocks()
+	if len(blocks) == 0 {
+		return nil, connect.NewError(
+			connect.CodeInvalidArgument,
+			errors.New("at least one block reference is required"),
+		)
+	}
+	if len(blocks) > DefaultMaxFetchBlockRefs {
+		return nil, connect.NewError(
+			connect.CodeResourceExhausted,
+			fmt.Errorf(
+				"block reference count %d exceeds maximum %d",
+				len(blocks),
+				DefaultMaxFetchBlockRefs,
+			),
+		)
+	}
 
 	db, release, err := a.bark.Acquire()
 	if err != nil {
@@ -46,7 +64,7 @@ func (a *archiveServiceHandler) FetchBlock(
 	}
 	defer release()
 
-	for _, b := range req.Msg.GetBlocks() {
+	for _, b := range blocks {
 		hash, err := hex.DecodeString(b.GetHash())
 		if err != nil {
 			return nil,
--- a/bark/auth.go
+++ b/bark/auth.go
@@ -20,8 +20,10 @@ import (
 	"crypto/x509"
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"log/slog"
 	"net/http"
+	"strings"
 
 	"connectrpc.com/connect"
 	databaseconnect "github.com/blinklabs-io/bark/proto/v1alpha1/database/databasev1alpha1connect"
@@ -53,14 +55,10 @@ var destructiveDatabaseProcedures = map[string]bool{
 	databaseconnect.DatabaseServiceCancelOperationProcedure: true,
 }
 
-// readOnlyDatabaseProcedures is the only set of DatabaseService procedures
-// operatorAuthInterceptor.authorize exempts from requiring a verified mTLS
-// client certificate. This is deliberately an allowlist, not the inverse of
-// destructiveDatabaseProcedures: a procedure absent from BOTH maps — e.g. a
-// new DatabaseService RPC added to the proto without updating either one —
-// is still required to authenticate, the same as a known-destructive one.
-// See authorize's doc comment for why the runtime check is structured this
-// way.
+// readOnlyDatabaseProcedures is the set of authenticated DatabaseService
+// procedures that do not additionally require operator authorization. This is
+// deliberately an allowlist: a procedure absent from both maps is treated as
+// destructive and therefore requires an allowed operator fingerprint.
 var readOnlyDatabaseProcedures = map[string]bool{
 	databaseconnect.DatabaseServiceGetSnapshotStatusProcedure:       true,
 	databaseconnect.DatabaseServiceListSnapshotsProcedure:           true,
@@ -73,9 +71,9 @@ var readOnlyDatabaseProcedures = map[string]bool{
 }
 
 // peerIdentity is what peerCertContextMiddleware extracts from a verified
-// mTLS client certificate and stashes in the request context — just enough
-// to authorize (Verified) and audit-log (CommonName/Fingerprint) a
-// destructive call, not a general certificate representation.
+// mTLS client certificate and stashes in the request context — just enough to
+// authenticate every DatabaseService request, authorize destructive calls by
+// fingerprint, and identify callers in audit logs.
 type peerIdentity struct {
 	Verified    bool
 	CommonName  string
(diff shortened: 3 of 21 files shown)


wolf31o2:
Local exact-head review at `449caadf85cf29f00ebd52587bf6487a720ea3fd`: no actionable findings. Bark authentication, operator fingerprint authorization, request and response limits, configuration propagation, and tests are consistent.

Passed: `git diff --check` and `gofmt`.

Not run: Go tests, because Go 1.26.7 was required and the toolchain download was blocked by network/DNS. Human review remains required.

cryptodj413:
@coderabbitai review

README.md
```diff
@@ -615,9 +620,13 @@ a target beyond the security parameter, because it exists for disaster-recovery
...
 Praos allows. The resulting database is resync-ready from the target point.
 
-The same operations are also exposed remotely through the Bark
-`DatabaseService`. Bark has no built-in authentication, so do not expose its
-port outside a trusted network.
+The same operations are also exposed remotely through Bark's
+`DatabaseService`. Every `DatabaseService` RPC requires a client certificate
+verified against `barkClientCaFilePath`; destructive RPCs also require the
+certificate's SHA-256 fingerprint in
+`barkOperatorCertificateFingerprints`. Bark's read-only `ArchiveService`
```
cryptodj413, line 627: **API/wire compatibility — documentation accuracy.** This section documents the fingerprint requirement but doesn't say how to derive the value, and the obvious command produces output the parser rejects: `openssl x509 -noout -fingerprint -sha256 -in client.crt` prints `sha256 Fingerprint=AB:CD:...`, and `normalizeOperatorCertificateFingerprints` (bark/auth.go:165) only strips whitespace and colons — the leading `sha256 Fingerprint=` text makes `hex.DecodeString` fail with "is not hexadecimal". Recommend documenting the DER-based form instead: `openssl x509 -in client.crt -outform DER | openssl dgst -sha256`.

Also worth adding here: this requirement is a **breaking change** for any existing deployment that already runs Bark's `DatabaseService` with `barkClientCaFilePath` configured but no operator fingerprints — it will now fail to start (`internal/config/validate.go:417-424`). README already carries upgrade-impact notes in this style elsewhere (e.g. around lines 315 and 400); worth being consistent here.

(Confirmed by the blink-app-auditor pass in this review; not independently re-verified by me beyond re-reading the cited lines.)

bark/auth.go
```diff
@@ -53,14 +55,10 @@ var destructiveDatabaseProcedures = map[string]bool{
...
-// destructiveDatabaseProcedures: a procedure absent from BOTH maps — e.g. a
-// new DatabaseService RPC added to the proto without updating either one —
-// is still required to authenticate, the same as a known-destructive one.
-// See authorize's doc comment for why the runtime check is structured this
-// way.
+// readOnlyDatabaseProcedures is the set of authenticated DatabaseService
+// procedures that do not additionally require operator authorization. This is
+// deliberately an allowlist: a procedure absent from both maps is treated as
+// destructive and therefore requires an allowed operator fingerprint.
 var readOnlyDatabaseProcedures = map[string]bool{
```
cryptodj413, line 62: **Missing contract-specific coverage / classification consistency.** `ListAvailableSnapshots` is classified read-only, but per `bark/database.go`'s `mergedSnapshotCatalogPage`, it calls `lifecycle.ListCloudSnapshots` on every invocation, unpaged and unmemoized — i.e. it can drive billed cloud LIST spend on every call, same as the stated rationale for putting `CreateSnapshot` on the destructive side just above ("consumes local (or billed cloud) storage on every anonymous call", lines 42-48). Severity is lower than pre-PR since it's no longer anonymous (any CA-verified client can still call it), but the classification looks inconsistent with this file's own stated threat model. Worth either reclassifying as destructive or documenting why cloud-LIST cost isn't considered in scope for the read-only bucket.

Related and pre-existing (not from this PR): `pageSize` in the same handler is client-controlled and uncapped beyond the new blanket 1 MiB `WithSendMaxBytes`, which now just turns an oversized page into an opaque `resource_exhausted` instead of a truncated page. Possibly worth a follow-up issue rather than blocking here.

bark/auth.go
```diff
@@ -175,40 +197,49 @@ func newOperatorAuthInterceptor(
...
-// the classification) and still authenticated exactly like a known
-// destructive one, never allowed through by default.
+// the classification) and still requires operator authorization.
 func (i *operatorAuthInterceptor) authorize(
 	ctx context.Context,
 	procedure string,
 ) error {
+	id := peerIdentityFromContext(ctx)
+	if !id.Verified {
+		i.logger.Warn(
```
cryptodj413, line 233: **Hardening recommendation, non-blocking.** Every anonymous request to `DatabaseService` now emits a `Warn` here. Before this PR, anonymous read-only calls succeeded silently; now any caller who can reach the port (which is intentionally shared with the public `ArchiveService`) can generate `Warn`-level log volume at request rate. Consider `Debug` for this specific anonymous-rejection case (the `PermissionDenied` path a few lines down and the success-audit `Info` below still carry the operationally-interesting signal), or rate-limit/sample it.

bark/archive.go
```diff
@@ -39,14 +40,31 @@ func (a *archiveServiceHandler) FetchBlock(
 	req *connect.Request[archive.FetchBlockRequest],
 ) (*connect.Response[archive.FetchBlockResponse], error) {
 	resp := &archive.FetchBlockResponse{}
+	blocks := req.Msg.GetBlocks()
+	if len(blocks) == 0 {
```
cryptodj413, line 44: **API/wire compatibility, undocumented.** An empty `blocks` list previously returned an empty `FetchBlockResponse` with `OK`; this PR makes it `InvalidArgument`. Dingo's own client always sends exactly one ref (`bark/blob.go:367-379`), so nothing in-tree is affected, but this is an observable behavior change for any external `ArchiveService` client and isn't mentioned in README.md or ARCHITECTURE.md. Worth a one-line note in one of those docs, or in the commit/PR description, so external consumers aren't surprised.

bark/auth_wire_test.go
```diff
@@ -30,16 +30,11 @@ import (
...
-// certificate signed by the configured CA gets past the auth check
-// entirely (it may still fail for an unrelated reason, e.g. an unknown
-// operation ID — the assertion here is specifically about authentication,
-// not about that RPC's own business logic succeeding).
-func TestDatabaseServiceRequiresClientCertForDestructiveRPCs(t *testing.T) {
+// TestDatabaseServiceAuthenticationAndOperatorAuthorization is the wire-level
+// proof of Bark's two-stage DatabaseService contract: every method requires a
+// verified client identity, while destructive methods additionally require an
+// explicitly allowed certificate fingerprint.
+func TestDatabaseServiceAuthenticationAndOperatorAuthorization(t *testing.T) {
```
cryptodj413, line 37: **Missing contract-specific test.** Nothing in this file (or elsewhere in `bark/*_test.go`) pins that `ArchiveService` stays reachable *without* a certificate while `DatabaseService` is mounted on the same listener with mTLS required. That's the property that silently breaks if the auth interceptor is ever moved from the handler level to the mux/listener level. Worth adding an explicit case here: anonymous `FetchBlock` against a lifecycle-enabled, client-CA-configured server should still reach the handler (and fail only on its own logic, not on auth).

internal/config/validate.go
```diff
@@ -401,23 +403,40 @@ func (c *Config) validate(effectiveMode RunMode, minBindable uint) error {
...
+		len(c.BarkOperatorCertificateFingerprints) == 0 {
+		errs = append(errs, errors.New(
+			"barkOperatorCertificateFingerprints requires at least one SHA-256 "+
+				"client certificate fingerprint when bark is enabled (barkPort) "+
+				"alongside databaseLifecycle.snapshotDir: verified identity alone "+
+				"does not authorize destructive DatabaseService RPCs",
+		))
+	}
+	for idx, fingerprint := range c.BarkOperatorCertificateFingerprints {
+		normalized := strings.ReplaceAll(strings.TrimSpace(fingerprint), ":", "")
```
cryptodj413, line 429: **Architecture / maintainability, non-blocking.** This fingerprint-normalization logic (`TrimSpace` → strip `:` → `hex.DecodeString` → length check) is duplicated from `bark/auth.go:160-181`'s `normalizeOperatorCertificateFingerprints`, and the two currently agree. If they ever drift, `Validate` could accept a value `NewBark` then rejects, turning a clean CLI/config error into a deeper `failed to create bark server` failure inside `Node.Run()`. Consider exporting `bark`'s normalizer and having this call it directly, so there's one implementation instead of two that need to be kept in sync.

cryptodj413 (commented):
Reviewed with 4 parallel domain passes (Bark/config safety review, adversarial security read of the auth interceptor and TLS wiring, isolated build/test/lint validation in a fresh worktree, and a cross-repo scout for consumers of the new config surface). No merge blockers found.

**Bot review status:** CodeRabbit hit its hourly review-limit on the first pass and, after being asked to re-review, declined ("does not re-review already reviewed commits") without ever posting actual findings. Cubic hit its **monthly** line-limit (403,149/400,000 lines) and won't resume until 2026-09-01. Neither bot produced findings for this PR; treat this review as the substantive bot-equivalent pass.

**Validation (isolated worktree, exact head commit):** `go build`, `go vet`, `gofmt -l`, `go test ./bark/... ./internal/config/... ./internal/node/... -race` (all pass, including every auth/TLS/request-limit/database-wire test), `make import-boundaries`, `make docs-parity`, `golangci-lint run ./...` (0 issues), `make gorm-check` — all clean. `nilaway` reports 2 findings, both confirmed pre-existing on `origin/main` (one on generated protobuf code in `bark/auth_test.go`, one in `mithril/bootstrap_test.go`, untouched by this diff) — not introduced here.

Note: the PR comment claiming "Go tests were not run because Go 1.26.7 is required and toolchain download was blocked" did not reproduce in the validation environment — go1.26.7 was present and the full suite ran clean with `-race`, including the specifically-flagged auth/TLS/request-limit tests.

**Security (adversarial pass):** traced all 14 DatabaseService RPCs through the generated handler registration — the interceptor is a handler option applied uniformly, no per-method bypass exists. Verified the "absent from both classification maps = destructive" default-deny is runtime-enforced (auth.go:246-255), not just documented. Verified `VerifyClientCertIfGiven` + the app-level `!id.Verified` check together close the anonymous-cert gap, confirmed against real TLS handshakes in `auth_wire_test.go`. Verified `WithReadMaxBytes`/`WithSendMaxBytes` bound the *decompressed* output via connect-go's `io.LimitReader` wrapping (no zip-bomb bypass). No confirmed or plausible bypass found for any of: destructive-RPC-without-fingerprint, any-RPC-without-verified-cert, request-limit bypass, panic/unbounded-allocation from malformed input, or a cert-verification/dispatch TOCTOU. Every claimed security property has a passing negative test, not just a positive-path one.

**Cross-repo:** the new `barkOperatorCertificateFingerprints` config is a breaking change for existing deployments (Bark's `DatabaseService` now refuses to start without it). `docs-site`'s "create-start-up-service" operational guide (en/es/ja) currently instructs operators to enable Bark live-restore/truncate without this new required key — following it as written today would produce a config that fails this PR's new startup validation. That's a separate-repo follow-up, not something to block this PR on. `helm-charts`, `dingo-operator`, `kb`, and `bark` itself are unaffected (helm-charts' Bark block is fully commented out; the others don't reference Bark's TLS/config surface).

Inline comments below are all non-blocking recommendations, ordered by the categories the review skill specifies.

PR #3578, opened 2026-08-27
fix(devnet): stabilize accelerated tx propagation and stake-key cleanup

### LINKED ISSUE #3265: devnet: clean root-owned stake-key copies after successful runs
## Problem

The all-Dingo run-tests.sh copies genesis stake keys from a Docker volume into a host mktemp directory via a root container. The copied stake directory and files are root-owned. The EXIT trap later runs rm -rf as the host user, which fails with Permission denied.

A successful accelerated scenario therefore returns exit 1 after reporting all phases passed and valid txpump activity. The observed run completed readiness, propagation, agreement, epoch transition, producer recovery, and relay recovery in 1m59s, then failed only while removing files under /tmp/dingo-devnet-stake-keys.*.

Repeated runs also accumulate these temporary directories.

## Acceptance criteria

- stake-key copies are owned or removable by the invoking user;
- successful test and txpump checks produce exit 0;
- cleanup removes only the temporary directory created by that invocation;
- failure artifacts remain preserved as documented.

arepala-uml:
### Summary

- Restrict stake-key cleanup to temporary directories created by the current runner invocation.
- Preserve caller-provided stake-key directories and failure artifacts.
- Publish the effective runtime genesis timestamp to txpump.
- Prevent txpump from submitting transactions before genesis, even when Docker reports nodes healthy.
- Clamp pre-genesis slot calculations to slot zero.

### Changes by file

- `internal/test/devnet/run-tests.sh`
  - Tracks whether the runner created the stake-key directory.
  - Removes the directory only when owned by that invocation.
  - Preserves inherited or caller-provided paths.

- `internal/test/devnet/run_tests_script_test.go`
  - Adds coverage proving caller-owned directories are preserved.
  - Adds environment override support to the fake runner.
  - Documents the cleanup ownership invariant.

- `internal/test/devnet/configurator.sh`
  - Writes the effective runtime genesis timestamp into shared runtime configuration.
  - Documents why Docker health is insufficient as a submission barrier.

- `internal/test/devnet/docker-compose.yml`
  - Provides runtime genesis configuration to txpump in Dingo and conformance profiles.

- `internal/test/antithesis/internal/txpump/pump.go`
  - Waits for genesis before starting transaction generation.
  - Keeps the startup deadline active during the wait.
  - Returns slot zero before genesis instead of wrapping a negative duration.

- `internal/test/antithesis/internal/txpump/pump_test.go`
  - Tests genesis waiting.
  - Tests startup timeout propagation.
  - Tests safe pre-genesis slot calculation.

Closes #3265 

---
## Summary by cubic
Stabilizes devnet runs by waiting for the real genesis timestamp before txpump submits transactions and by restricting stake-key cleanup to directories the runner itself created.

- txpump now waits for the runtime genesis time even when Docker reports nodes healthy, and reports slot zero before genesis instead of a wrapped negative duration.
- The devnet runner now only removes stake-key directories it created itself; caller-provided and inherited paths are preserved on cleanup.

<sup>Written for commit 00704ea30e9ed1fc5b2e4caa3bf5782fff84ad68. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3578?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Transaction submission now waits for the network’s actual genesis time before starting.
  - Prevented incorrect slot calculations before genesis.
  - Startup timeout and cancellation conditions now report appropriate readiness errors.
  - Devnet test cleanup preserves stake-key directories supplied by the caller.

- **Tests**
  - Added coverage for genesis waiting, startup timeouts, pre-genesis slot handling, and directory preservation.

- **Chores**
  - Devnet services now use the effective runtime genesis configuration.

### DIFF
--- a/internal/test/antithesis/internal/txpump/pump.go
+++ b/internal/test/antithesis/internal/txpump/pump.go
@@ -67,6 +67,13 @@ func (p *Pump) Run(ctx context.Context) error {
 		startup = startupTimer.C
 	}
 	ready := false
+	// A healthy node can accept an N2C connection before the network's
+	// configured system start. Keep the first submission behind the genesis
+	// boundary so an accepted pre-genesis transaction is not later discarded
+	// when forging and ledger revalidation begin.
+	if err := p.waitForGenesis(ctx, startup); err != nil {
+		return err
+	}
 	for {
 		select {
 		case <-ctx.Done():
@@ -131,6 +138,33 @@ func (p *Pump) Run(ctx context.Context) error {
 		if !p.cooldown(ctx) {
 			return ctx.Err()
 		}
+	}
+}
+
+// waitForGenesis blocks transaction generation until the configured network
+// start. The startup deadline remains active while waiting, so a bad runtime
+// genesis timestamp fails readiness instead of leaving txpump hung forever.
+func (p *Pump) waitForGenesis(
+	ctx context.Context,
+	startup <-chan time.Time,
+) error {
+	delay := time.Until(p.genesisTime)
+	if delay <= 0 {
+		return nil
+	}
+	p.logger.Info("waiting for genesis start", "delay", delay)
+	timer := time.NewTimer(delay)
+	defer timer.Stop()
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-startup:
+		return fmt.Errorf(
+			"txpump readiness timeout after %s: genesis has not started",
+			p.cfg.StartupTimeout,
+		)
+	case <-timer.C:
+		return nil
 	}
 }
 
@@ -202,12 +236,16 @@ func enabledTypes(
 	return enabled
 }
 
-// currentSlot returns the number of 1-second slots elapsed since the Pump was
-// created.  Using elapsed time (rather than Unix epoch seconds) ensures that
-// the slot counter starts near 0 and epoch gating works correctly for devnet
-// testing, where epochs are only 500 slots long.
+// currentSlot returns the number of 1-second slots elapsed since genesis.
+// Using elapsed time (rather than Unix epoch seconds) ensures that the slot
+// counter starts near 0 and epoch gating works correctly for devnet testing,
+// where epochs are only 500 slots long. Before genesis, report slot 0 rather
+// than converting a negative duration to a near-MaxUint64 slot.
 func (p *Pump) currentSlot() uint64 {
 	elapsed := time.Since(p.genesisTime)
+	if elapsed <= 0 {
+		return 0
+	}
 	return uint64(elapsed.Seconds())
 }
 
--- a/internal/test/antithesis/internal/txpump/pump_test.go
+++ b/internal/test/antithesis/internal/txpump/pump_test.go
@@ -1,0 +1,61 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package txpump
+
+import (
+	"context"
+	"io"
+	"log/slog"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func testPump(genesisTime time.Time, startupTimeout time.Duration) *Pump {
+	return NewPump(
+		&Config{StartupTimeout: startupTimeout},
+		NewWallet(),
+		slog.New(slog.NewTextHandler(io.Discard, nil)),
+		nil,
+		genesisTime,
+	)
+}
+
+// TestWaitForGenesis proves that node connectivity cannot release transaction
+// generation before the network start boundary.
+func TestWaitForGenesis(t *testing.T) {
+	pump := testPump(time.Now().Add(25*time.Millisecond), time.Second)
+	started := time.Now()
+	require.NoError(t, pump.waitForGenesis(context.Background(), nil))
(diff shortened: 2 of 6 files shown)


arepala-uml:
## Testing

### Initial failure 

The accelerated all-Dingo test failed `./internal/test/devnet/run-tests.sh --accelerated` while waiting for a transaction-bearing block:

```
[run-tests] Running DevNet integration tests...
testing: warning: no tests to run
PASS
ok  	github.com/blinklabs-io/dingo/internal/test/devnet	0.513s [no tests to run]
=== RUN   TestAcceleratedScenarioTimeline
    accelerated_timeline_test.go:76: scenario plan (epoch 1m0s, block ~1.25s, hard timeout 5m0s): readiness<=45s propagation<=1m10s agreement<=1m22.5s epoch-transition<=2m50s peer-interruption<=3m33.75s relay-restart<=4m17.5s
    accelerated_timeline_test.go:126: === phase readiness: deadline 45s from start (45s remaining)
    observer.go:237: observer dingo-relay: following chain from localhost:3015
    observer.go:237: observer dingo-2: following chain from localhost:3013
    observer.go:237: observer dingo-3: following chain from localhost:3014
    observer.go:237: observer dingo-1: following chain from localhost:3010
    accelerated_timeline_test.go:146: readiness: 4 nodes streaming chain events
    accelerated_timeline_test.go:126: === phase propagation: deadline 1m10s from start (45.777s remaining)
    accelerated_timeline_test.go:167: block propagation: slot 12 block 1 reached all 4 nodes
    accelerated_timeline_test.go:173: 
        	Error Trace:	/Users/akrepala/3147/clanker/repos/dingo/internal/test/devnet/scenarios/accelerated_timeline_test.go:173
        	Error:      	Received unexpected error:
        	            	devnet: "a block carrying transactions reaches every node" not satisfied by [dingo-1{tip=slot 100/block 27, connected=true, fwd=33, back=6, maxRollback=1, lastErr=""}, dingo-2{tip=slot 100/block 27, connected=true, fwd=32, back=5, maxRollback=1, lastErr=""}, dingo-3{tip=slot 100/block 27, connected=true, fwd=34, back=7, maxRollback=1, lastErr=""}, dingo-relay{tip=slot 100/block 27, connected=true, fwd=31, back=4, maxRollback=1, lastErr=""}]: context deadline exceeded
        	Test:       	TestAcceleratedScenarioTimeline
    artifacts.go:243: devnet: failure artifacts written to /var/folders/qt/z5nv031j2d78sf0y0vyh7g380000gn/T/dingo-devnet-artifacts.xVP1xU/accelerated-timeline
--- FAIL: TestAcceleratedScenarioTimeline (70.69s)
FAIL
FAIL	github.com/blinklabs-io/dingo/internal/test/devnet/scenarios	71.493s
FAIL
[run-tests] DevNet tests FAILED (exit code: 1)
[run-tests] Checking txpump accepted submissions...
```

Investigation showed:

- txpump submitted transactions before the configured genesis start.
- The node accepted the submissions without reporting rejection or error.
- Transactions initially entered the mempool.
- Transactions were absent when block forging began.
- Forged blocks reported `tx_count=0`.
- The transaction-propagation phase timed out.

Full log: 
[dingo-devnet-accelerated-20260826.log](https://github.com/user-attachments/files/31499026/dingo-devnet-accelerated-20260826.log)


### Successful accelerated all-Dingo rerun

After publishing the effective runtime genesis timestamp and making txpump wait for genesis, the accelerated scenario passed:

```text
transaction propagation: slot 25 carries a 240-byte body on all nodes
--- PASS: TestAcceleratedScenarioTimeline
[run-tests] txpump submitted=3 rejected=0 error=0
```

The run completed:

- Node readiness
- Block propagation
- Transaction propagation
- Chain agreement
- Epoch transition
- Producer recovery
- Relay recovery
- Invocation-specific stake-key directory cleanup

The scenario completed in approximately 1 minute and 56 seconds and exited with status `0`.

Full log: 
[dingo-devnet-accelerated-fixed2-20260826.log](https://github.com/user-attachments/files/31499069/dingo-devnet-accelerated-fixed2-20260826.log)

wolf31o2 (approved):
Approving. Both the shell-side and Go-side fixes are real, and the `uint64(negative)` slot conversion was a genuine latent bug rather than defensive tidying.

Verified load-bearing: reverting the `STAKE_KEYS_HOST_DIR_IS_OURS` guard fails `TestRunTestsPreservesCallerOwnedStakeDirectory`. The ownership pattern correctly mirrors `ARTIFACT_DIR_IS_OURS`, and the case it actually fixes is the non-`dingo` mode, where `STAKE_KEYS_HOST_DIR` can be inherited from the environment and the old cleanup would `rm -rf` a caller's directory — the `mktemp` that sets it only runs in the `dingo` branch.

`waitForGenesis` is well-shaped: it honours `ctx`, keeps the startup deadline live so a bad runtime genesis timestamp fails readiness rather than hanging, and a nil `startup` channel (no configured timeout) correctly blocks forever on that arm only. A missing or zero `genesisTime` degrades to the previous behaviour rather than stalling.

**One thing worth knowing about `TestCurrentSlotBeforeGenesisIsZero`: it cannot fail on an arm64 machine.** Removing the `if elapsed <= 0 { return 0 }` clamp leaves the whole txpump package green here. Go specifies that converting an out-of-range float to an integer yields an implementation-dependent result, and arm64's conversion saturates to 0 — so the assertion holds with or without the fix. On the amd64 runner CI uses, the same conversion produces the indefinite value rather than saturating, which is the near-`MaxUint64` slot the fix exists to prevent.

So the test is meaningful in CI and inert for anyone developing on Apple silicon or arm64 Linux. Not something to change in this PR — the guard makes the behaviour architecture-independent, which is the right fix regardless — but if that test is ever relied on locally to prove a regression, it will not do so. A comment naming the amd64 dependency would save the next person the confusion it cost me.

Checks at 00704ea3: `go test ./internal/test/devnet/` and `go test -race ./internal/txpump/` green, CI green on all 11 checks.

PR #3580, opened 2026-08-27
test(benchmark): preserve block numbers in processing fixture

### LINKED ISSUE #3375: test(benchmark): preserve block numbers in processing fixture
## Problem

The scheduled benchmark workflow failed on `main` at
`557e602af37b8a11dce92734837ea2dd7dbdf37b`:

https://github.com/blinklabs-io/dingo/actions/runs/32697793433

`BenchmarkBlockProcessingThroughput` loaded 4,096 blocks, then failed before
measuring block processing:

```text
AddBlock failed: block 9447e7ee372f0a4e61282e2764515d088d76abc31c92eed13fd93af2310e48ed claims block number 5 that is not contiguous with parent 0
```

`loadBlockProcessingFixture` reads five real seed blocks but stores every seed
model with `Number: 0`. The first timed block retains its canonical block number
`5`, so `Chain.AddBlock` correctly rejects it after the seeded tip reports block
number `0`.

The benchmark failure prevented the workflow from reaching regression
comparison and benchmark-baseline publication. The job later received a runner
shutdown signal and reported exit code 143, but the non-contiguous fixture is
the first benchmark failure in the log.

## Acceptance criteria

- Preserve the real block numbers and parent relationships when seeding the
  block-processing fixture.
- Validate that the seed and timed block stream are non-empty and contiguous
  before starting the benchmark timer.
- `BenchmarkBlockProcessingThroughput` and its predecoded variant complete
  successfully against the real immutable fixture.
- `make bench-ci` reaches regression comparison and benchmark-data publication.

arepala-uml:
## Summary
- `loadBlockProcessingFixture` stored every seeded block with `Number: 0` instead of its real decoded block number, leaving the post-seed chain tip reporting block number 0.
- The first timed block then failed `Chain.AddBlock`'s contiguity check against its real block number before the benchmark timer started.
- Preserve each block's decoded number, and validate that the seed and timed block streams are hash- and number-contiguous before the benchmark proceeds.

Fixes #3375

## Test plan
- [x] `go build ./...`
- [x] `gofmt -l ledger/benchmark_test.go` (clean)
- [x] `golangci-lint run ./ledger/...`
- [x] `nilaway ./ledger/...`
- [x] `modernize ./ledger/...`
- [x] `go test ./internal/architecture/...` (import-boundaries)
- [x] `go test ./internal/docsparity/...` (docs-parity)
- [x] `go test -race ./ledger/...`
- [x] `go test -bench '^Benchmark(BlockProcessingThroughput|BlockProcessingThroughputPredecoded|BlockfetchNearTipThroughput)$' -count=2 ./ledger` — all pass against the real immutable fixture
- [ ] `make bench-ci` — not run in full (90m+30m curated suite); the two benchmarks named in the issue were validated directly above

`DATABASE.md`/`ARCHITECTURE.md`: checked, not affected (test-fixture-only change).

---
## Summary by cubic
Fixes `loadBlockProcessingFixture` storing every seeded block with `Number: 0` instead of its real decoded block number, which caused the first timed block to fail `Chain.AddBlock`'s contiguity check before the benchmark timer started. Preserves each block's decoded number and validates that seed and timed block streams are hash- and number-contiguous before the benchmark proceeds.

<sup>Written for commit 2a8fb4b1bfaa240031d8fcefeb565d3f733bda60. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3580?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Improved benchmark fixture validation to detect broken parent references and inconsistent block numbering.
  * Added support for valid block-number behavior at historical epoch boundaries.
  * Updated benchmark data handling to reflect actual block numbers, improving test accuracy and reliability.

### DIFF
--- a/ledger/benchmark_test.go
+++ b/ledger/benchmark_test.go
@@ -15,6 +15,7 @@
 package ledger
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -24,6 +25,7 @@ import (
 	"testing"
 
 	"github.com/blinklabs-io/gouroboros/ledger"
+	"github.com/blinklabs-io/gouroboros/ledger/byron"
 	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
 	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
 
@@ -2632,6 +2634,19 @@ func loadBatchProcessingFixture(
 	return seedModels, batchBlocks, rawBlocks, batchSize
 }
 
+// blockNumberFollowsParent mirrors chain.blockNumberContiguous, which is
+// unexported: a block number must be exactly parent+1, except a Byron-era
+// epoch boundary block which legitimately repeats its parent's number.
+func blockNumberFollowsParent(
+	eraId uint8,
+	blockNumber, parentNumber uint64,
+) bool {
+	if blockNumber == parentNumber+1 {
+		return true
+	}
+	return eraId == byron.EraIdByron && blockNumber == parentNumber
+}
+
 func loadBlockProcessingFixture(
 	b *testing.B,
 ) ([]models.Block, []*immutable.Block) {
@@ -2646,6 +2661,10 @@ func loadBlockProcessingFixture(
 
 	const seedCount = 5
 	const blockCount = blockProcessingBenchmarkFixtureBlockCount
+
+	var prevHash []byte
+	var prevNumber uint64
+	var havePrev bool
 
 	seedModels := make([]models.Block, 0, seedCount)
 	for len(seedModels) < seedCount {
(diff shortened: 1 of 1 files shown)


wolf31o2 (approved):
Approving. The fix is load-bearing and reproduces the reported failure exactly: restoring `Number: 0` on the seeded blocks makes `BenchmarkBlockProcessingThroughput` fail with

```
AddBlock failed: block 9447e7ee... claims block number 5 that is not contiguous with parent 0
```

which is #3375's symptom verbatim. With the fix the benchmark completes and reports throughput.

`blockNumberFollowsParent` is a faithful mirror of `chain.blockNumberContiguous` — identical structure, same Byron epoch-boundary carve-out. The comment is right that the production one is unexported and cannot be reused directly.

Two non-blocking observations, both in the new fixture validation rather than the fix:

- The rule is now duplicated with nothing tying the copies together. If `blockNumberContiguous` ever gains a case, this benchmark keeps validating the old rule and would reject a fixture the production path accepts. A cheap tie would be exporting the predicate for tests, or asserting the two agree over a small table.
- Both loops `continue` on a `NewBlockFromCbor` error without advancing `prevHash`/`prevNumber`. A block that fails to decode is therefore skipped silently, and the *next* block then fails the prev-hash check against the block before the skipped one — surfacing a decode failure as a misleading "does not match parent hash". Failing on the decode error directly, or advancing past it, would name the real cause.

Checks at 2a8fb4b1: `go test -bench BenchmarkBlockProcessingThroughput` passes against the real immutable fixture, CI green.

PR #3581, opened 2026-08-27
fix(ouroboros): contain decoder panics in the shared decode cache

### LINKED ISSUE #3511: ouroboros: contain decoder failures in the cache
A decoder panic is converted to a cached error, then re-raised by a waiting
goroutine without a recovery boundary. The failure escapes the protocol worker
instead of remaining a contained decode result.

Acceptance criteria:

- Convert decoder panics into stable cacheable errors.
- Ensure waiters return the error without re-panicking.
- Preserve cache cleanup and retry behavior.
- Cover concurrent waiters around a decoder panic.

cryptodj413:
Fixes #3511.

`getOrDecode` recovered a decoder panic long enough to cache it and wake
in-flight waiters, but then re-raised the panic to the leader goroutine's own
caller. Every waiter and later cache hit already got a clean error; only the
leader — the goroutine that actually ran `decodeFn` — still crashed its
calling protocol worker (`blockfetchClientBlockRaw` /
`chainsyncClientRollForwardRaw`).

`getOrDecode` now returns the recovered panic as a normal `(value, err)`
result instead of re-raising, matching every other caller for the same key.
`decodeWithPanicSafeMetrics`'s own defer/recover is removed since it existed
only to catch that re-raise.

## Checks

- `go test -race ./ouroboros/...` — pass
- `go test ./internal/architecture/...` (import-boundaries) — pass
- `go test ./internal/docsparity/...` (docs-parity) — pass
- `golangci-lint run ./ouroboros/...` — 0 issues
- `nilaway ./ouroboros/...` — clean
- `modernize ./ouroboros/...` — no findings
- `gofmt`/`golines` — clean

DevNet/conformance not run: this change only alters what a decode-panic
returns on the leader's own path (a normal error instead of a crash); it does
not change decode results, block/header validity, or consensus behavior.

`ARCHITECTURE.md`'s "Shared Block/Header Decode Cache" section updated to
match. `DATABASE.md` checked, not affected.

---
## Summary by cubic
Contain decoder panics in the shared decode cache so a panicking `decodeFn` returns a normal error to the leader goroutine instead of re-raising and crashing the calling protocol worker (`blockfetchClientBlockRaw` / `chainsyncClientRollForwardRaw`). Fixes #3511.

**Bug Fixes**
- The leader goroutine now receives the same clean error that in-flight waiters and later cache hits on the same key already received.
- Removed `decodeWithPanicSafeMetrics`'s defer/recover, which existed only to catch the re-raise.
- Updated `ARCHITECTURE.md`'s decode-cache section and the regression tests to match the contained behavior.

<sup>Written for commit 3a9019e459e305c7ba91ad4d2ed771b90878b92a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3581?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Decode failures caused by panics are now handled as regular errors instead of crashing protocol processing.
  * All callers receive a consistent error, including the original request and waiting requests.
  * Repeated deliveries with the same failed data fail quickly using the cached failure.
  * Panic-related decode metrics are now recorded as cache misses.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -3055,12 +3055,15 @@ waiter-channel design (see Leios CertRB Serving below), applied to ordinary
 block/header decoding rather than Leios EB closures.
 
 A decode function that panics (adversarial/malformed peer bytes could in
-principle trigger this) does not strand the key: `getOrDecode` recovers the
-panic just long enough to record it as an ordinary cached failure and wake
-every waiter, then re-raises it, so the decoding goroutine's own
-crash-or-recover behavior is unchanged but no other connection is left
-permanently blocked on that key, and a later identical delivery fails fast
-from cache instead of panicking again.
+principle trigger this) does not strand the key and does not crash the
+decoding goroutine either: `getOrDecode` recovers the panic, records it as an
+ordinary cached failure, wakes every waiter, and returns it as its own
+`(value, err)` result instead of re-raising, so a decoder panic never escapes
+into `blockfetchClientBlockRaw`/`chainsyncClientRollForwardRaw` or the
+protocol worker that called them — it is handled exactly like any other
+decode error, on the leader's own path as well as every waiter's. A later
+identical delivery still fails fast from the cached failure instead of
+decoding (and therefore panicking) again. See dingo #3511.
 
 A waiting caller's decode outcome is delivered directly through its wait
 channel rather than by re-reading the shared entry map after waking: the
--- a/ouroboros/decode_cache.go
+++ b/ouroboros/decode_cache.go
@@ -133,10 +133,16 @@ func newDecodeCache[T any]() *decodeCache[T] {
 // bookkeeping, never the decode itself, so concurrent decodes for different
 // keys never serialize against each other.
 //
-// A decodeFn that panics does not strand this key: the panic is recovered
-// long enough to record it as a normal cached failure and release every
-// waiter, then re-raised so the calling goroutine's own crash-or-recover
-// behavior is unchanged from before this cache existed. See finishDecode.
+// A decodeFn that panics does not strand this key, and does not crash the
+// calling goroutine either: the panic is recovered and recorded as a normal
+// cached failure via finishDecode, exactly like an ordinary decode error, so
+// the leader gets back a plain (zero value, err, true) result instead of a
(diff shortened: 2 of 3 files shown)


wolf31o2 (approved):
Approving. The change is well-targeted and the tests pin it: restoring `panic(r)` in the recover makes the panic tests crash the package with `panic: simulated decode panic [recovered, repanicked]`, so the new contract is genuinely covered rather than incidentally satisfied.

The design judgement is the right one, and worth stating explicitly since "swallow a panic" reads alarming out of context. The old behaviour was not "panics crash the node" — it was "panics crash whichever goroutine happened to win the race to become leader, while every other caller for the same bytes got a clean error." Which outcome you got for identical input depended on scheduling. Making the leader symmetric with its waiters removes that non-determinism rather than hiding a fault.

Two details I checked rather than assumed:

- `decoded = true` in the recover is correct: the leader did run `decodeFn`, so this call is a miss. It flows into `recordOutcome(decoded || err != nil)` as `true`, matching the `recordOutcome(true)` the deleted defer in `decodeWithPanicSafeMetrics` used to record, so the metric is unchanged. `TestDecodeWithPanicSafeMetricsRecordsMissOnPanic` covers it.
- The `completed` guard survives the rewrite, so the `panic(nil)` case still cannot be mistaken for normal completion — `TestDecodeCachePanicNilDuringDecodeDoesNotStrandWaitersOrKey` still stands behind it.

One consequence not called out in the description: `decodeWithPanicSafeMetrics` no longer has a recover of its own, so a panic originating in `getOrDecode`'s own bookkeeping rather than in `decodeFn` now propagates without recording an outcome. That is a strictly better failure mode for a genuine cache bug — it should be loud — but it is a behaviour change beyond the decodeFn path.

Checks at 3a9019e4: `go test -race ./ouroboros/...` green, CI green.

PR #3584, opened 2026-08-27
fix(ledger): read reward-pot protocol parameters as of performanceEpoch

### LINKED ISSUE #3570: prime-testnet: from-genesis sync wedges at epoch 10-11 boundary — dingo overcounts a delegator's reward balance by 105 lovelace, rejecting a canonical withdrawal
## Summary

A second, distinct wedge from #3566, found while validating that issue's fix.
Syncing prime-testnet from genesis on current `origin/main` (commit `5c63115b`,
includes `eb41b4e1`/#3478, `ec899411`, `f33f34f8`, `ea04317a`), the node gets
stuck at slot 4,569,852 (epoch 10→11 boundary) — much earlier than #3566's
slot 45,389,599 — in a non-converging chainsync/replay-recovery loop:
"replay recovery not converging after unresolved-producer failures, holding
at applied ledger tip", `no_progress_count` climbing every retry, repeated
peer disconnects and short-lived-connection backoff.

## Root cause: dingo overcounts one delegator's reward balance by 105 lovelace

The transaction that trips recovery:

```
TX 246db12614a9a0f0914bd6e46c66b3470d25f90e2c3593d93994fb37e6ac1f6b failed
validation: incorrect withdrawal amount for
stake_test1uptcaulp6cyrdreh599u5ts38dr705ekdfa7c44qz2wv2nqzh038r:
provided 1904431334, balance 1904431439
```

This is a real, on-chain, canonical transaction — the whole network's own
validators accepted the block containing it, and Shelley's WDRL rule requires
a withdrawal amount to exactly equal the account's current reward balance.
Since the network accepted `1904431334` as valid, that is provably the
account's true balance at that point; dingo's independently-computed
`1904431439` is **105 lovelace too high**.

Credential: `578ef3e1d608368f37a14bca2e113b47e7d3366a7bec56a0129cc54c`
(`stake_test1uptcaulp6cyrdreh599u5ts38dr705ekdfa7c44qz2wv2nqzh038r`),
delegated to pool `56c1fa63b5a152556db0b0c8ad5e6b60e2bf25e542793d88d349bb89`
(cost=340000000, margin=1/10, pledge=100000000, registered at slot 1130739 —
no re-registration in the window, so pool parameters were stable throughout).

`account_reward_delta` shows dingo's bookkeeping is internally
self-consistent — four per-epoch credits sum exactly to the account's stored
balance:

```
added_slot  amount      (epoch boundary)
3024000     415157667   (epoch 7→8)
3456000     496686152   (epoch 8→9)
3888000     503385941   (epoch 9→10)
4320000     489201679   (epoch 10→11)
----------  ---------
            1904431439  == account.reward (matches exactly)
```

So this isn't a bookkeeping/summation bug — dingo's addition is correct. The
~105-lovelace discrepancy must be in one or more of the four per-epoch
**member-reward computations themselves** (`ledger/rewards/rewards.go`'s
`memberRewardChecked`/`MemberReward`): `floor((1-margin) * (poolReward-cost) *
(memberStake/poolStake))`. That function itself uses exact `big.Rat`
arithmetic with `big.Int.Quo`-based flooring — no float64, so I don't suspect
the floor/multiply arithmetic itself. That leaves the four *inputs*
(`poolReward`, `cost`, `margin`, `memberStake`, `poolStake` for each of
epochs 7-10) as the more likely source: any one of them computed even
slightly differently than the real network's reference implementation would
shift a floor'd result by whole-lovelace amounts, and four small
per-epoch errors accumulating to 105 lovelace total (~26/epoch average) is
consistent with a small, systematic input discrepancy rather than a single
large miscalculation.

## What I have not yet pinned down

- Which of the two stake figures (`memberStake`/`poolStake`, i.e. this
  pool's and this delegator's "set"-equivalent snapshot dingo uses for
  reward calculation) is off, versus `poolReward` itself (pool performance /
  blocks-produced-derived apportionment of the epoch's total reward pot).
- Whether this is the same underlying mechanism as #3566's much-later
  wedge (both are small, real-network reward/stake-figure discrepancies)
  or a separate bug entirely. Given this one blocks at epoch 10-11, it is
  encountered *before* #3566's epoch 104-105 wedge on any full resync, so
  it needs fixing first regardless.

## Reproduction

Clean from-genesis sync of prime-testnet on `origin/main` (`5c63115b`)
against the public relays (`relay-0/1.prime.testnet.apexfusion.org:5521`)
deterministically reaches this wedge. Sqlite dumps, full sync log, and DB
access available on request.

## Non-fix ruled out

Same as #3566: `SkipLeaderStakeThresholdCheck`/any Musashi-only bypass is
not applicable here — this is a `WDRL` (withdrawal) rule rejection, not a
leader-eligibility check, and the account/pool involved are real
prime-testnet participants, not the prototype network.

wcatz:
## Summary
- Reward-pot protocol parameters (tau, rho, epoch length, ...) were fetched as of `potsEpoch` (`newEpoch-1`) instead of `performanceEpoch` (`newEpoch-2`), the epoch the pot's fees and block-production counts actually come from. This matches cardano-ledger's `prevPParamsEpochStateL` semantics, which stays aligned with the epoch being taxed rather than the later epoch whose pparams snapshot may already reflect a parameter change that hadn't taken effect yet for this reward round.
- The bug is invisible unless a reward-affecting protocol parameter changes between `performanceEpoch` and `potsEpoch`, which is rare on a mature network but happened during Apex Fusion prime-testnet's early bootstrap (`tau` 1e-6 -> 1e-8), producing a small per-pool reward overcount that compounds forward through the pool owner's own stake balance in later epochs.
- Fixes #3570. Full investigation (independent verification against a from-genesis Apex Fusion reference node, and a closed-loop test confirming the fix reproduces the real network's reward to within 1 lovelace) is in that issue's comment thread.

## Changes
- `ledger/reward_calculation.go`: both call sites that build reward-pot `rewards.Parameters` (`calculateStakeRewardApplication` and `precomputedStakeRewardApplication`) now pass `performanceEpoch` instead of `potsEpoch` as the pparams-fetch epoch.
- `ledger/reward_calculation_test.go`: adds `TestApplyStakeRewardsUsesPerformanceEpochPParamsNotPotsEpoch`, a regression test that registers a drastically different (poisoned) pparams version effective at `potsEpoch` and asserts the reward calculation is unaffected by it. Verified this test fails without the fix and passes with it. Also updates three pre-existing tests whose hardcoded reward-pot values were computed under the old (buggy) epoch selection.
- `ARCHITECTURE.md`: corrects the "Reward Calculation And Precomputation" section, which previously documented the old (buggy) behavior as intentional.

## Test plan
- [x] `go test -tags dingo_extra_plugins -race -timeout 20m -count=1 ./ledger/...` — full package, clean pass.
- [x] New regression test manually verified to fail on the pre-fix code and pass on the fix.
- [x] `./internal/test/devnet/run-tests.sh --accelerated --conformance` (dingo vs. real `cardano-node`) — `TestAcceleratedScenarioTimeline` PASSED in 96.78s; dingo's chain hash matched cardano-node's exactly through readiness, propagation, agreement, an epoch-transition boundary crossing, a peer interruption, and a relay restart.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -7890,8 +7890,11 @@ reward application first, ADA-pot capture last) by
 
 ### Reward Calculation And Precomputation
 
-Reward protocol parameters and epoch length come from the RUPD calculation
-epoch, while block-production counts use the delayed performance epoch. TPraos
+Reward protocol parameters and epoch length come from the performance epoch —
+the same epoch the pot's fees and block-production counts are drawn from —
+not the later pots epoch (whose ADA-pots row seeds the delayed calculation but
+whose own protocol-parameter snapshot may already reflect a governance update
+that had not yet taken effect for this reward round; see #3570). TPraos
 overlay slots are excluded while decentralization is non-zero. Pre-Babbage
 calculation resolves the reward prefilter from stake-account certificate
 history immediately before the first reward-update slot, using the RUPD
--- a/ledger/reward_calculation.go
+++ b/ledger/reward_calculation.go
@@ -283,10 +283,15 @@ func (ls *LedgerState) calculateStakeRewardApplication(
 		return nil, false, nil
 	}
 
+	// The reward-pot protocol parameters (tau, rho, ...) must come from the
+	// same epoch as the fees/blocks the pot was built from (performanceEpoch),
+	// matching cardano-ledger's prevPParamsEpochStateL semantics -- not
+	// potsEpoch, which is one epoch later and can carry a different value if a
+	// param changed between the two. See #3570.
 	pparams, params, performanceDecentralization, err := ls.rewardParameters(
 		txn,
 		performanceEpoch,
-		potsEpoch,
+		performanceEpoch,
 		pots,
 	)
 	if err != nil {
@@ -775,10 +780,13 @@ func (ls *LedgerState) precomputedStakeRewardApplication(
 	) {
 		return nil, false, nil
 	}
+	// See the matching comment in calculateStakeRewardApplication: the
+	// reward-pot protocol parameters must be read as of performanceEpoch, not
+	// potsEpoch.
 	pparams, params, performanceDecentralization, err := ls.rewardParameters(
 		txn,
 		epochs.performance,
-		epochs.pots,
+		epochs.performance,
 		pots,
 	)
 	if err != nil {
--- a/ledger/reward_calculation_test.go
+++ b/ledger/reward_calculation_test.go
@@ -306,6 +306,255 @@ func TestApplyStakeRewardsUsesDelayedRewardState(t *testing.T) {
 	require.Equal(t, int64(2), deltas)
 }
 
+// TestApplyStakeRewardsUsesPerformanceEpochPParamsNotPotsEpoch is a regression
+// test for #3570: the reward-pot protocol parameters (tau, rho, ...) must be
+// read as of performanceEpoch, matching the epoch the fees/blocks in the pot
+// came from, not potsEpoch (one epoch later). It reproduces
+// TestApplyStakeRewardsUsesDelayedRewardState's fixture exactly, but also
+// registers a second, drastically different pparams version -- effective at
+// potsEpoch, with a much higher treasury cut (tau) -- simulating a governance
+// update landing between performanceEpoch and potsEpoch, as happened on the
+// real network #3570 was filed against. If the reward calculation incorrectly
+// used potsEpoch's pparams (the pre-fix behavior), the treasury cut and every
+// downstream reward figure would differ from
+// TestApplyStakeRewardsUsesDelayedRewardState's values; this asserts they
+// don't.
+func TestApplyStakeRewardsUsesPerformanceEpochPParamsNotPotsEpoch(t *testing.T) {
+	ls, db := newRewardCalculationTestLedger(t)
+	meta := db.Metadata()
+
+	const (
+		newEpoch            = uint64(4)
+		rewardSnapshotEpoch = uint64(1)
+		performanceEpoch    = uint64(2)
+		potsEpoch           = uint64(3)
+		boundarySlot        = uint64(400)
+	)
+	poolKey := rewardCalcHash(0x11)
+	rewardAccount := rewardCalcHash(0x22)
+	member := rewardCalcHash(0x33)
+	var poolID lcommon.PoolKeyHash
+	copy(poolID[:], poolKey)
+
+	pparams := &shelley.ShelleyProtocolParameters{
+		NOpt:             10,
+		A0:               rewardCalcRat(1, 2),
+		Rho:              rewardCalcRat(1, 100),
+		Tau:              rewardCalcRat(0, 1),
+		Decentralization: rewardCalcRat(0, 1),
+		ProtocolMajor:    7,
+		ProtocolMinor:    0,
+	}
+	pparamsCbor, err := cbor.Encode(pparams)
+	require.NoError(t, err)
+
+	// A drastically different pparams version -- 90% treasury cut instead of
+	// 0% -- takes effect at potsEpoch. If the reward calculation reads this
+	// (the pre-fix bug), every downstream figure below diverges from
+	// TestApplyStakeRewardsUsesDelayedRewardState's values.
+	poisonedPParams := &shelley.ShelleyProtocolParameters{
+		NOpt:             10,
+		A0:               rewardCalcRat(1, 2),
+		Rho:              rewardCalcRat(1, 100),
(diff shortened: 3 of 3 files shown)


wolf31o2 (commented):
Not approving or blocking — this needs a decision that sits above the PR. **#3482 is open, targets `main`, and fixes the same defect with contradictory semantics.**

Both move reward protocol parameters from `newEpoch-1` to `newEpoch-2` (this PR calls the former `potsEpoch`, #3482 calls it `calculationEpoch`). They disagree on **epoch length**, and each states its position in `ARCHITECTURE.md`:

- #3482: "Only the epoch length comes from the pots epoch, standing in for the `slotsPerEpoch` the RUPD rule passes for the epoch it runs in."
- This PR: "Reward protocol parameters **and epoch length** come from the performance epoch."

The mechanism differs too, which is what makes this dangerous. #3482 changes the *callee* — `rewardParameters` uses `performancePParams` for the values while keeping `uint64(calculationEpochRow.LengthInSlots)` for `EpochLength`. This PR changes the *callers*, passing `performanceEpoch` for the `calculationEpoch` argument as well, so `calculationEpochRow` becomes the performance epoch row and epoch length moves with everything else.

They touch different lines of `reward_calculation.go` and different sections of `ARCHITECTURE.md`, so **both will merge cleanly with no conflict**. The combined result is this PR's semantics, with #3482's code comment and its `ARCHITECTURE.md` paragraph both still asserting that epoch length comes from the pots epoch. No test would catch that; the documentation would simply be wrong in two places.

It is not a hypothetical difference. `seedRewardPrecomputeTimingState` deliberately gives potsEpoch a different `LengthInSlots` from performanceEpoch, and `EpochLength` feeds `expectedBlocks`, which feeds `eta`, which scales the monetary expansion. Which epoch supplies it changes the pot.

On this PR's own merits it holds up. Reverting both call sites fails `TestApplyStakeRewardsUsesPerformanceEpochPParamsNotPotsEpoch` and nothing else. I also checked the 10x expectation changes in the three pre-existing tests rather than taking them on trust — running `main`'s version of `TestPrecomputedStakeRewardsFinalEligibilityDoesNotMergeCredentialTags` against this branch's code fails with expected `0x1a0a` (6666) against actual `0x1046a` (66666), so the updates are a real consequence of the fix and not adjusted to make a test pass.

What is needed is which reading of `startStep` is right for `slotsPerEpoch` — the epoch being taxed, or the epoch the RUPD rule runs in — and then one of these two PRs closed or rebased onto the other. Epoch length is a genesis-level value that only moves across an era boundary, which is exactly the from-genesis bootstrap window both issues (#3481, #3570) are about, so the two cannot both be correct there.

Checks at 37ad9bdf: `go test ./ledger/ -run TestApplyStakeRewards` green, CI green on 10 checks.

wcatz:
Duplicate of #3482, which fixes the same root cause (reward-pot pparams read from the wrong epoch) more correctly and with stronger validation. Closing in favor of that PR.

wolf31o2:
Following up on the epoch-length question raised above — the cardano-ledger
source answers it, and it goes against this PR.

`Rules/Rupd.hs` computes the epoch size from the current slot and passes it to
`startStep` as its own argument, separate from any parameter set:

```haskell
let e = epochInfoEpoch ei s
slotsPerEpoch = epochInfoSize ei e
...
startStep slotsPerEpoch b es maxsupply asc k
```

`s` is the slot RUPD runs at, so `e` is the epoch the rule runs in — this
repository's `potsEpoch`. In `PulsingReward.hs`, `startStep` reads `d`, `rho`
and `tau` from `pr = es ^. prevPParamsEpochStateL` and uses `slotsPerEpoch`
only inside `expectedBlocks`.

So parameter values come from the performance epoch, which this PR gets right,
but epoch length comes from the pots epoch, which this PR changes. Passing
`performanceEpoch` for `rewardParameters`' `calculationEpoch` argument moves
both.

That also explains the three pre-existing tests whose expectations this PR
updates. On #3482's branch, repointing `EpochLength` at
`performanceEpochRow.LengthInSlots` fails exactly those three plus
`TestRewardParametersSplitCalculationAndPerformanceEpochInputs`. The 6666 ->
66666 change is real and is caused by the fix, as the earlier comment
confirmed — but it is moving a correct value to an incorrect one, because
`EpochLength` feeds `expectedBlocks` and should still be the pots epoch's.

The pparams-value half of this change is already covered by #3482, which is
approved. The remaining question is whether #3570 has a residue not covered
there — the prime-testnet `tau` 1e-6 -> 1e-8 case looks like exactly the
value-side bug #3482 fixes, in which case this can close; if the closed-loop
reproduction still diverges with #3482 applied and epoch length left on the
pots epoch, that difference is worth reporting on #3570.

Issue #3217, opened 2026-08-19
api/blockfrost and api/utxorpc release their listening port after Stop returns


wolf31o2:
Same defect as the one fixed for `api/mesh` in #3214, still present in two other API servers.

## Mechanism

`startServer` calls `net.Listen` (the socket accepts immediately via the backlog) and hands the listener to `Serve` in a goroutine it does not wait for. `Stop` calls only `srv.Shutdown(ctx)`, and `http.Server.Shutdown` closes only listeners `Serve` has already registered. Stopping inside that window returns with the port still bound.

- `api/blockfrost/blockfrost.go:412` — `net.Listen`, then `server.Serve(ln)` at `:425` inside `go func()`; `Stop` shuts down at `:381` without closing `ln`.
- `api/utxorpc/utxorpc.go:570` — `net.Listen`, then `server.Serve(ln)` at `:584` inside `go func()`; shutdown at `:508-509` without closing `ln`.

## Why it matters in production

Not test-only. `quiesceForLiveLifecycleOp` stops these capabilities and `reinitializeAPIServers` re-resolves them on the same configured port. All three servers are in that path:

- `CapabilityAPIUtxorpc` — `node_lifecycle.go:210`, re-resolved at `:1118`/`:1125`
- `CapabilityAPIBlockfrost` — `node_lifecycle.go:239`, re-resolved at `:1181`/`:1192`
- `CapabilityAPIMesh` — `node_lifecycle.go:247`, re-resolved at `:1205`/`:1226` (fixed in #3214)

So a live Restore/Truncate can fail to rebind with `EADDRINUSE`. On the mesh server, an instrumented 500-iteration probe left the port accepting in 46 runs (~9%) before the fix.

## Fix shape

What #3214 landed for `api/mesh`, applied to each:

1. Record the listener on the server struct.
2. Detach server and listener together under one lock (`takeServer`) so exactly one caller shuts them down.
3. Close the listener after `Shutdown`, tolerating `net.ErrClosed` — after, not before, so `Serve` still exits as `http.ErrServerClosed` rather than logging an error.
4. Publish the listener only while the call's server is still current, and have `Stop` wait for an in-flight bind to settle, so `Stop` cannot return while a racing bind still holds the socket.

## Not done

Left out of #3214 to keep that PR scoped. No fix attempted here.

PR #3591, opened 2026-08-27
fix(node): serialize startup cleanup and shutdown

### LINKED ISSUE #3529: runtime: bound lifecycle and event back-pressure behavior
Several long-running runtime paths have incomplete lifecycle or back-pressure behavior:

- Nonce text is encoded before checking the cache, doing avoidable work on every hit.
- A slot-clock subscription can be registered after the clock has stopped.
- Startup signal cleanup can race the normal shutdown path.
- One stalled event subscriber can block every publisher of that event indefinitely.

Acceptance criteria:

- Check the nonce cache before encoding and preserve cache semantics.
- Reject or close subscriptions registered after stop.
- Serialize startup-failure cleanup with normal shutdown.
- Bound or isolate subscriber back-pressure without reordering accepted events.
- Add lifecycle, cancellation, and stalled-subscriber tests.

wolf31o2:
Fixes #3529.

- Serialize startup rollback and normal shutdown.
- Cover signal cancellation while startup cleanup is in progress.

---
## Summary by cubic
Fixes #3529 by serializing startup rollback and normal shutdown so a SIGINT/SIGTERM during startup cannot tear down a partially initialized component twice.

- `Node.Run` now holds a lifecycle gate until startup completes or its LIFO cleanup finishes; `Node.Stop` waits on the same gate before beginning teardown.
- Adds a test covering signal cancellation while startup cleanup is in progress.

<sup>Written for commit 88ec2fef9bb68b92a531cc1db36aa407fe543bb8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3591?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/node_shutdown.go
+++ b/node_shutdown.go
@@ -87,6 +87,13 @@ func (n *Node) configuredShutdownTimeout() time.Duration {
 }
 
 func (n *Node) shutdown() error {
+	// Run holds this gate until startup has either completed or rolled back.
+	// In particular, a signal can reach Stop while Run is still unwinding a
+	// failed startup; waiting here keeps the phase-ordered shutdown from
+	// concurrently closing a component the startup stack is stopping.
+	n.startupLifecycleMu.Lock()
+	defer n.startupLifecycleMu.Unlock()
+
 	shutdownStart := time.Now()
 	shutdownTimeout := n.configuredShutdownTimeout()
 	ctx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1128,6 +1128,13 @@ Phase 4: Cleanup resources
   Registered shutdown functions
 ```
 
+`Node.Run` holds a startup lifecycle gate from entry until startup either
+completes or has unwound its LIFO rollback stack. Normal shutdown takes the
+same gate before its phase-ordered teardown begins. A SIGINT/SIGTERM received
+while components are still starting can therefore cancel startup without
+letting `Node.Stop` concurrently close a partially initialized component; the
+normal shutdown waits until rollback has finished.
+
 The node creates one shutdown context from the configured `shutdownTimeout`
 and passes it through every phase. PeerGovernor shutdown cancels its internal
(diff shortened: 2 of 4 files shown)


node_shutdown.go
```diff
@@ -87,6 +87,13 @@ func (n *Node) configuredShutdownTimeout() time.Duration {
 }
 
 func (n *Node) shutdown() error {
+	// Run holds this gate until startup has either completed or rolled back.
+	// In particular, a signal can reach Stop while Run is still unwinding a
+	// failed startup; waiting here keeps the phase-ordered shutdown from
+	// concurrently closing a component the startup stack is stopping.
+	n.startupLifecycleMu.Lock()
```
arepala-uml, line 94: Non-blocking: this lock is acquired before `configuredShutdownTimeout()` starts its clock, so if a signal lands while `cleanupFailedStartup`'s LIFO rollback is running, `shutdown()` blocks here for as long as that rollback takes. Several of the `started` closures it runs stop components with a bare `context.Background()` and no deadline (`pluginHost.Stop`, `historyExpiry.Stop`, `connManager.Stop`, `bark.Stop`, `midnightServer.Stop`, `offchainMetadataFetcher.Stop`, `tokenRegistrySync.Stop`) — the same unbounded-Stop shape already tracked in #3217 for the API servers. Before this PR a hang there didn't block `Stop()` (it just raced, which was the bug); after this PR it blocks `Stop()` indefinitely, so `ShutdownTimeout` is no longer honored in that narrow case. Operators still have a second-signal/orchestrator-SIGKILL escape hatch, so I'm not blocking on it — suggest a follow-up to bound the gate wait itself, or give these closures the same `stopWithDeadline` treatment `quiesceForLiveLifecycleOp` already uses.

arepala-uml (approved):

PR #3592, opened 2026-08-27
fix(event): preserve lossless critical subscribers

wolf31o2:
## Summary

- bound ordinary stalled event subscribers and expose detachment telemetry
- preserve lossless block-fetch delivery through lifecycle cancellation
- cover concurrent publishers and block-fetch backpressure

Part of #3529

## Checks

- `go test -race -count=1 ./event`
- `go test -race -count=1 -run ^'TestBlockfetchSubscriptionRemainsLosslessPastDeliveryTimeout$' ./ledger`
- `make docs-parity`
- `golangci-lint run ./event ./ledger`

---
## Summary by cubic
Binds how long a stalled subscriber can hold up the event bus and keeps critical ledger streams lossless so block-carrying events are never silently dropped. Ordinary subscribers that stay full past the delivery timeout are now detached; previously they parked publishers indefinitely until shutdown.

- Detached subscribers keep events already accepted into their channel in order, while the unaccepted event and later events continue only to healthy subscribers.
- `PublishBlocking` reports the new `ErrEventSubscriberStalled` error, and detachment increments the `event_delivery_timeouts_total` metric plus a warning log.
- `SubscribeWithBufferPolicy` and `SubscribeFuncWithBufferPolicy` expose `SubscriberBackpressureBlock`, which keeps a full subscriber attached until it drains or lifecycle cancellation closes it.
- `ledger` opts its chainsync, blockfetch, and chain-update subscriptions into the blocking policy and treats `ErrEventSubscriberStalled` from optional block-event consumers as non-fatal instead of stopping the node.

<sup>Written for commit 58982f5950e8fdfa33da4c4466c3e72cf3f1b5fc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3592?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable subscriber backpressure policies.
  * Stalled subscribers are automatically detached after the delivery timeout by default.
  * Added a lossless blocking policy for subscriptions that must retain every event.
  * Publishers now receive a clear error when a subscriber is detached for stalling.
  * Healthy subscribers continue receiving events in order when another subscriber is stalled.

* **Bug Fixes**
  * Blockfetch event delivery now remains lossless during backpressure.

* **Documentation**
  * Updated event delivery guarantees, timeout behavior, and monitoring documentation.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -498,11 +498,13 @@ only enqueues onto the ordered lane, so subscribers cannot observe an Apply
 before storage is durable and their work never runs inline in `Commit`.
 `ledger.block` remains `PublishBlocking` on the handler's own goroutine.
 
-Because a publisher parked on a full lane is released only by the EventBus
-stopping, and a live restore/truncate closes the `LedgerState` while keeping
-the bus running, ledger publishes go through `PublishOrderedContext` with a
-context `LedgerState.Close` cancels first thing. Without it `Close` waits
-unbounded on a `ledger.tx` subscriber that stopped draining.
+An ordinary stalled subscriber detaches after its delivery timeout; a lossless
+subscriber deliberately remains attached until it drains or is stopped, closed,
+or unsubscribed. A live restore/truncate closes the `LedgerState` while keeping
+the bus running, so ledger publishes go through `PublishOrderedContext` with a
+context `LedgerState.Close` cancels first thing. That cancellation releases a
+full ordered lane regardless of whether a subscriber later drains or its
+lifecycle removes it.
 
 This is also the one deliberate exception to the publish-under-lock rule
 above. `rollbackChainAndState` runs under `chainsyncMutex` — it is reached
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1242,11 +1242,13 @@ All event types follow the `subsystem.snake_case_name` convention.
 - Lossless delivery with bounded producer backpressure: when a subscriber buffer
   or the async queue is full, `Publish`, `PublishBlocking`, and `PublishAsync`
   wait for capacity instead of dropping an event for a live subscriber. An
-  in-memory subscriber that cannot free capacity by the delivery timeout is
-  detached; events already accepted into that subscriber retain their order,
-  while the unaccepted event and later events continue to healthy subscribers.
-  Waits also end on `Stop`, `Close`, or `Unsubscribe`. `PublishBlocking`
-  reports the detachment error (or `ErrEventBusStopped` when shutdown wins).
+  ordinary in-memory subscriber that cannot free capacity by the delivery
+  timeout is detached; events already accepted into that subscriber retain
+  their order, while the unaccepted event and later events continue to healthy
+  subscribers. A lossless subscriber explicitly uses the blocking policy when
+  detachment would make its component unable to recover safely. Waits also end
+  on `Stop`, `Close`, or `Unsubscribe`. `PublishBlocking` reports the
+  detachment error (or `ErrEventBusStopped` when shutdown wins).
 - A slow subscriber can temporarily backpressure its publishers, but a stalled
   one cannot hold the topic indefinitely. Consumers of `Subscribe` channels
   must drain for the life of the subscription and `Unsubscribe` when they stop.
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1239,26 +1239,30 @@ All event types follow the `subsystem.snake_case_name` convention.
   buffers for high-volume ledger chainsync and chain-update paths. The
   payload-heavy ledger blockfetch path uses an eight-entry buffer and relies
   on lossless backpressure once one chain-store commit batch is queued
-- Lossless delivery with producer backpressure: when a subscriber buffer or the
-  async queue is full, `Publish`, `PublishBlocking`, and `PublishAsync` all wait
-  for capacity instead of dropping the event. Waits end on `Stop`, `Close`, or
-  `Unsubscribe`. `PublishBlocking` differs only in returning
-  `ErrEventBusStopped` when the bus shuts down mid-delivery
-- A subscriber that stops draining stalls its publishers, and the shared async
-  worker pool means it also delays unrelated async event types. Consumers of
-  `Subscribe` channels must drain for the life of the subscription and
-  `Unsubscribe` when they stop. The stalled-subscriber warning is rate-limited
-  per subscriber, not per delivery, and reports how many publishers are parked
-  on it — every parked publisher observes the same stall, so a per-delivery
-  limit made the log volume scale with publisher count rather than with time
+- Lossless delivery with bounded producer backpressure: when a subscriber buffer
+  or the async queue is full, `Publish`, `PublishBlocking`, and `PublishAsync`
+  wait for capacity instead of dropping an event for a live subscriber. An
+  in-memory subscriber that cannot free capacity by the delivery timeout is
+  detached; events already accepted into that subscriber retain their order,
+  while the unaccepted event and later events continue to healthy subscribers.
+  Waits also end on `Stop`, `Close`, or `Unsubscribe`. `PublishBlocking`
+  reports the detachment error (or `ErrEventBusStopped` when shutdown wins).
+- A slow subscriber can temporarily backpressure its publishers, but a stalled
+  one cannot hold the topic indefinitely. Consumers of `Subscribe` channels
+  must drain for the life of the subscription and `Unsubscribe` when they stop.
+  The stalled-subscriber warning is rate-limited per subscriber, not per
+  delivery, and reports how many publishers are parked on it — every parked
+  publisher observes the same stall, so a per-delivery limit made the log
+  volume scale with publisher count rather than with time
 - **Never `Publish`, `PublishAsync`, `PublishOrdered`, or `PublishBlocking`
-  while holding a lock that a subscriber of that event acquires.** Because all four wait for
-  capacity rather than dropping, this is a deadlock, not merely a slow path:
-  once the subscriber's buffer fills, the subscriber waits for the lock the
-  publisher holds while the publisher waits for the capacity the subscriber
-  would free. `PublishAsync` is no exception — it waits for room in the shared
-  async queue, which is drained by a worker pool running subscriber handlers,
-  so a handler blocked on the publisher's lock closes the same cycle. Both
+  while holding a lock that a subscriber of that event acquires.** All four can
+  wait for capacity, and a subscriber that is merely slow is still allowed the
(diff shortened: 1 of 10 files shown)


ledger/state.go
```diff
@@ -1500,11 +1500,7 @@ func (ls *LedgerState) Start(ctx context.Context) error {
 			ChainsyncAwaitReplyEventType,
 			ls.handleEventChainsyncAwaitReply,
 		)
-		ls.blockfetchSubID = ls.config.EventBus.SubscribeFuncWithBuffer(
-			BlockfetchEventType,
-			blockfetchCommitBatchSize,
-			ls.handleEventBlockfetch,
-		)
+		ls.subscribeBlockfetchEvents(ls.handleEventBlockfetch)
 		ls.chainUpdateSubID = ls.config.EventBus.SubscribeFuncWithBuffer(
```
arepala-uml: **Behavioral: `ChainsyncEventType`/`ChainUpdateEventType` can be permanently detached, defeating the lossless guarantee this PR adds for blockfetch**

`ChainsyncEventType`'s only subscriber (`handleEventChainsync`, ~line 1494) and `ChainUpdateEventType` (this line) still use the default `SubscriberBackpressureDetach` policy, unlike blockfetch, which this PR opts into `SubscriberBackpressureBlock` via `subscribeBlockfetchEvents`. Both are called out elsewhere as ordering-critical: the `PublishBlocking` call site in `ouroboros/chainsync.go` says dropping a rollback/header event "can strand the ledger pipeline," and the comment two lines above this one cites #1556/#1914 bulk-sync bursts filling this exact buffer.

Once one of these subscribers' buffers stays full for `RemoteDeliverTimeout` (5s), the EventBus now detaches it permanently -- `chainsyncSubID`/`chainUpdateSubID` are set once in `Start()` and never resubscribed. For `ChainsyncEventType`, the resulting `ErrEventSubscriberStalled` propagates out of `chainsyncClientRollForward`/`RollBackward` as a callback error, but the real effect is that the ledger stops receiving chainsync events from any peer afterward, silently, until a process restart.

Reproduced against this branch with the same subscription shape as `handleEventChainsync`:

```go
func TestDiagChainsyncSubscriptionCanBeDetached(t *testing.T) {
	bus := event.NewEventBus(nil, nil)
	defer bus.Stop()

	// mirrors the default-policy subscribe at state.go:1494
	_, ch := bus.SubscribeWithBuffer(ChainsyncEventType, 1)
	// fills the buffer; nothing ever drains ch
	bus.Publish(ChainsyncEventType, event.NewEvent(ChainsyncEventType, 0))

	errCh := make(chan error, 1)
	go func() {
		errCh <- bus.PublishBlocking(
			ChainsyncEventType,
			event.NewEvent(ChainsyncEventType, 1),
		)
	}()

	err := <-errCh // returns after RemoteDeliverTimeout
	_, open := <-ch // false: closed by detachment, not by draining
}
```

Actual output:
```
PublishBlocking(ChainsyncEventType) returned after 5.000836458s: event subscriber stalled
channel closed=false after detachment
```

Suggest opting `ChainsyncEventType` and `ChainUpdateEventType` into `SubscriberBackpressureBlock` the same way blockfetch was, or letting a `PublishBlocking` caller distinguish "my own listener stalled" from "an unrelated listener on this event type stalled."

event/event.go
```diff
@@ -55,6 +55,18 @@ var ErrEventBusStopped = errors.New("event bus stopped")
 
 var errChannelSubscriberClosed = errors.New("channel subscriber closed")
 
+// ErrEventSubscriberStalled is returned by PublishBlocking when an ordinary
+// in-memory subscriber stays full past the delivery bound and is detached.
+var ErrEventSubscriberStalled = errors.New("event subscriber stalled")
```
arepala-uml, line 60: **Behavioral: `PublishBlocking` conflates "an unrelated ordinary subscriber was cleanly detached" with "my publish failed," which can escalate to a full node shutdown**

`ledger/block_event.go`'s `publishBlockEvent` treats any `PublishBlocking(BlockEventType, ...)` error other than `ErrEventBusStopped` as fatal and calls `FatalErrorFunc`, which `node_ledger_config.go` wires to `n.cancel()` -- shutting down the whole node. `midnight/indexer/indexer.go` subscribes to `BlockEventType` with this package's default `SubscriberBackpressureDetach` policy (100k buffer). If that indexer ever gets wedged long enough to fill its buffer, this PR's detach mechanism correctly protects the other `BlockEventType` subscribers and reports it via the `deliveryTimeouts` metric plus a warning log -- but `publishBlockEvent` can't tell that apart from a genuine delivery failure, so it also escalates to `FatalErrorFunc` and takes the node down over an optional downstream consumer stalling.

Same root cause as the sibling finding on `ledger/state.go`: `ErrEventSubscriberStalled` doesn't identify which subscriber stalled, so a `PublishBlocking` caller that checks the error can't distinguish a subscriber it doesn't depend on from the one it does. Lower likelihood to hit in practice (100k-entry buffer, opt-in feature) but the resulting failure mode -- a stalled optional consumer crashing the node -- is worse than what #3529 set out to fix.

Same reproduction shape as the `ledger/state.go` comment: a `SubscribeWithBuffer` + full-buffer + `PublishBlocking` sequence reliably returns `ErrEventSubscriberStalled` after `RemoteDeliverTimeout`. The gap here is in how `publishBlockEvent` interprets that return value, not in this file.

arepala-uml (commented):
Verified the branch locally (build, `go test -race ./event/...`, and the targeted ledger race test all pass) and confirmed CodeRabbit's ARCHITECTURE.md finding is already resolved by `bf780936`. Two behavioral findings below on `PublishBlocking`'s new stalled-subscriber error, each with a reproducing test run against this branch.

ledger/state.go
arepala-uml: Fixed in 58982f5: `ChainsyncEventType` and `chain.ChainUpdateEventType` in `ledger/state.go` now use `SubscribeFuncWithBufferPolicy(..., event.SubscriberBackpressureBlock, ...)`, matching blockfetch. Re-verified against current head: both subscriptions now block rather than detach, and `go test -race ./event/...` plus the ledger event-ordering tests pass.

event/event.go
arepala-uml: Fixed in 58982f5: `publishBlockEvent` now treats `event.ErrEventSubscriberStalled` the same as `event.ErrEventBusStopped` (non-fatal), covered by the new `TestPublishBlockEventIgnoresOptionalSubscriberDetachment`. Confirmed this error can only originate from a Detach-policy subscriber -- `deliverWait`'s delivery-timeout timer is only armed for `SubscriberBackpressureDetach`, so a Block-policy subscriber can never produce it. The only current `BlockEventType` subscriber (midnight indexer) uses the default detach policy, so this is scoped correctly. Ran the new test with -race, passes.

ARCHITECTURE.md
```diff
@@ -1239,26 +1239,32 @@ All event types follow the `subsystem.snake_case_name` convention.
...
+- Lossless delivery with bounded producer backpressure: when a subscriber buffer
+  or the async queue is full, `Publish`, `PublishBlocking`, and `PublishAsync`
+  wait for capacity instead of dropping an event for a live subscriber. An
+  ordinary in-memory subscriber that cannot free capacity by the delivery
+  timeout is detached; events already accepted into that subscriber retain
+  their order, while the unaccepted event and later events continue to healthy
+  subscribers. A lossless subscriber explicitly uses the blocking policy when
+  detachment would make its component unable to recover safely. Waits also end
+  on `Stop`, `Close`, or `Unsubscribe`. `PublishBlocking` reports the
+  detachment error (or `ErrEventBusStopped` when shutdown wins).
```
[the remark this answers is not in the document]
arepala-uml: Confirmed still resolved at current head (58982f5): ARCHITECTURE.md lines ~499-505 now distinguish ordinary-detach from lossless-blocking subscribers, consistent with the timeout-detachment description this comment flagged.

ledger/state.go
```diff
@@ -1491,23 +1491,21 @@ func (ls *LedgerState) Start(ctx context.Context) error {
...
 			ls.handleEventChainsyncAwaitReply,
 		)
-		ls.blockfetchSubID = ls.config.EventBus.SubscribeFuncWithBuffer(
-			BlockfetchEventType,
-			blockfetchCommitBatchSize,
-			ls.handleEventBlockfetch,
-		)
-		ls.chainUpdateSubID = ls.config.EventBus.SubscribeFuncWithBuffer(
+		ls.subscribeBlockfetchEvents(ls.handleEventBlockfetch)
+		ls.chainUpdateSubID = ls.config.EventBus.SubscribeFuncWithBufferPolicy(
```
arepala-uml, line 1505: Non-blocking, for confirmation rather than as a request: `chain.ChainUpdateEventType` has other subscribers outside `ledger` that still use the default `SubscriberBackpressureDetach` policy -- `ledger/leios/manager.go` and `ledger/leios/pipeline.go` (plain `Subscribe`), `mempool/mempool.go` (`SubscribeWithBuffer`), and `ouroboros/lifecycle.go`'s `subscribeTracked` (`SubscribeFunc`). Since `PublishBlocking` delivers to each subscriber independently and continues past one detachment, this doesn't weaken the guarantee this PR adds for `ledger/state.go`'s own subscription. Worth a quick check though: is a dropped chain-update event tolerable for leios/mempool (self-healing on the next event), or is this the same class of gap for a different owner?

arepala-uml (approved):
Re-reviewed at current head after 58982f5. Both behavioral findings from my prior review (ChainsyncEventType/ChainUpdateEventType detach policy in ledger/state.go, and PublishBlocking's stalled-subscriber error escalating to a fatal node shutdown in block_event.go) are fixed and verified -- replies with specifics are on each thread. CodeRabbit's ARCHITECTURE.md doc-drift finding remains resolved. Ran `go test -race ./event/...`, the targeted blockfetch losslessness test, and the new block-event-ordering/backpressure tests myself: all green. `make docs-parity` fails in this sandbox, but reproduces identically on origin/main (internal/docsparity is untouched by this diff), so it's a pre-existing environment issue, not a regression here.

One non-blocking question below on ledger/state.go: other ChainUpdateEventType consumers (leios/manager.go, leios/pipeline.go, mempool.go, ouroboros/lifecycle.go) still use the default detach policy. PublishBlocking delivers to each subscriber independently, so this doesn't undermine the fix just landed for ledger/state.go's own subscription -- just flagging in case any of those also need the same guarantee.

No merge-blocking findings.

PR #3596, opened 2026-08-27
fix(ledger): verify TPraos nonce VRF proofs

### LINKED ISSUE #3522: ledger: verify nonce VRF proofs before updating epoch state
Header validation incorporates the nonce VRF output into epoch nonce state but
does not verify its proof. The leader VRF proof is checked independently, so a
header can pass that check while carrying an unrelated nonce contribution.

Acceptance criteria:

- Verify the nonce VRF proof against the registered VRF key.
- Reject a header before updating epoch nonce state when verification fails.
- Cover valid, mismatched-key, and altered-proof headers.

wolf31o2:
## Summary

- verify the independent TPraos nonce VRF certificate before header admission
- source the VRF key, output, and proof from signed header-body CBOR
- cover Shelley through Alonzo valid, mismatched-key, and altered-proof cases

## Validation

- `go test -race ./ledger -run "^(TestVerifyBlockHeaderTPraosNonceVRF|TestVerifyBlockHeader_ValidBlock|TestVerifyBlockHeader_TamperedVRFProof|TestVerifyBlockHeader_UsesBodyCBORVRFFields)$" -count=1`
- `go test -race ./internal/test/conformance/...`
- `go vet ./ledger`
- full `go test -race ./ledger` reached the 10-minute package timeout in unrelated SQLite migration setup; the named active test passed alone under `-race`

`DATABASE.md` and `ARCHITECTURE.md` were checked and are not affected.

Fixes #3522

---
## Summary by cubic
Verifies the independent TPraos nonce VRF certificate in Shelley-through-Alonzo block headers before header admission, so mismatched-key or altered-proof nonce VRF certificates now fail verification. Babbage and later eras carry only the single leader VRF result and are unaffected.

- Sources the VRF key, output, and proof from the original signed header-body CBOR, not stale typed fields.
- Adds test coverage across Shelley, Allegra, Mary, and Alonzo for valid, mismatched-key, and altered-proof cases.
- Fixes #3522.

<sup>Written for commit 04ad00015f52600904d71181f3c06c1acaf46570. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3596?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ledger/verify_header_vrf.go
+++ b/ledger/verify_header_vrf.go
@@ -15,7 +15,9 @@
 package ledger
 
 import (
+	"encoding/hex"
 	"fmt"
+	"math"
 
 	"github.com/blinklabs-io/gouroboros/cbor"
 	"github.com/blinklabs-io/gouroboros/ledger"
@@ -27,6 +29,7 @@ import (
 	"github.com/blinklabs-io/gouroboros/ledger/dijkstra"
 	"github.com/blinklabs-io/gouroboros/ledger/mary"
 	"github.com/blinklabs-io/gouroboros/ledger/shelley"
+	"github.com/blinklabs-io/gouroboros/vrf"
 )
 
 const (
@@ -52,7 +55,15 @@ func normalizeHeaderVrfFieldsFromBodyCbor(
 	if err != nil || !ok {
 		return header, err
 	}
-	return headerWithVrfKeyAndResult(header, vrfKey, vrfResult), nil
+	normalized := headerWithVrfKeyAndResult(header, vrfKey, vrfResult)
+	nonceVrfResult, ok, err := headerNonceVrfResultFromBodyCbor(header)
+	if err != nil {
+		return header, err
+	}
+	if ok {
+		normalized = headerWithNonceVrfResult(normalized, nonceVrfResult)
+	}
+	return normalized, nil
 }
 
 func headerVrfKeyFromBodyCbor(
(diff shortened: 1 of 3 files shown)


ledger/verify_header_vrf.go
```diff
@@ -269,6 +312,125 @@ func headerWithVrfKeyAndResult(
...
+		return header
+	}
+}
+
+// verifyTPraosNonceVrfHex verifies the independent bheaderEta certificate
+// carried by Shelley-through-Alonzo headers. Its output is folded into epoch
+// nonce state, so this check must run before a header can reach ledger apply.
+// Babbage and later Praos headers carry only the single leader VRF result and
+// therefore have no separate nonce certificate to verify here.
+func verifyTPraosNonceVrfHex(
```
arepala-uml, line 346: Non-blocking: the four size-validation branches here (epoch nonce, VRF key, proof, output length) have no test coverage -- all existing tests exercise well-formed-but-invalid inputs (mismatched key, altered proof). These are the branches that guard against malformed wire data specifically, so they're the ones most worth a defensive test.

arepala-uml (approved):
Verified the field indices (NonceVrf at index 5, LeaderVrf at index 6 in the CBOR-array-encoded ShelleyBlockHeaderBody) and the SeedEta/SeedL seed construction against gouroboros's existing, already-proven leader-VRF check (gouroboros/ledger/verify_block.go) -- this reuses the same primitives with a different seed constant, matching the Haskell mkSeed reference cited in gouroboros/vrf/vrf.go.

Also fetched two real mainnet blocks live over NtN blockfetch and ran them through verifyBlockHeaderHex with their real epoch nonces from Koios: Shelley epoch 208 slot 4492900 (the first Shelley epoch) and Alonzo epoch 300 slot 44237054. Both verify successfully with the correct epoch nonce and are correctly rejected with a wrong one, confirming the check accepts genuine historical headers rather than only the synthetic round-trip fixtures in this PR.

One scope note: the conformance suite cited in the PR's validation section (`go test -race ./internal/test/conformance/...`) runs Amaru's Conway-era ledger-rule vectors and doesn't exercise this TPraos header-crypto path at all -- true as stated, just not coverage for this specific change. Not blocking given the live-block verification above.

No merge-blocking findings.

Issue #1900, opened 2026-04-13
Cross-node validation: automated ledger state comparison vs Haskell node


wolf31o2:
Run Dingo and cardano-node in parallel and compare ledger state block-by-block to catch divergence.

What this covers:
- Automated parallel execution of Dingo and Haskell node on same network
- Block-by-block ledger state comparison (UTxO set, stake distribution, protocol params)
- Alerting on any divergence
- Focus on preview and preprod initially

Roadmap: Q3 2026

PR #3599, opened 2026-08-27
test(conformance): exercise real sqlite/postgres/mysql backends

arepala-uml:
## Summary

Addresses the 2026-08-21 audit addendum on #1900: the conformance suite's
sqlite/postgres/mysql "modes" never actually exercised those backends —
`DingoStateManager` ran entirely against in-memory Go maps behind an inert
mirror, and the MySQL/PostgreSQL constructors discarded their DSN parameter
outright, so an invalid or unreachable database configuration had no effect.

This rewires the manager onto a real `*database.Database`, composed the same
way the production node composes storage at startup, and reuses production
persistence code (`database.SetTransactionMetadataOnly`, `ledger/governance`
functions, `metadata.UtxoStore`) rather than a second hand-rolled
implementation. Backend construction now fails loudly on an invalid or
unreachable DSN.

Building and running this against real Postgres/MySQL surfaced three genuine
bugs, all fixed here: two pre-existing production bugs and one Go import
cycle in the new test wiring itself.

## Changes made

**`fix(database): close pool_registration cursor before nested queries`**
`loadPoolAssociations` queried `pool_registration` through an open
`*sql.Rows` cursor and issued nested owner/relay queries on the same
connection before closing it. SQLite tolerates this; MySQL and PostgreSQL
don't — a pool with at least one owner/relay corrupted the connection,
surfacing as a spurious "pool not found" on the next unrelated read. Fixed
by collecting rows into a slice and closing the cursor before loading
children, matching the pattern the plural `loadPoolsAssociations` already
uses. Includes a regression test reproducing the exact call shape.

**`fix(database): check drep existence before activity update`**
`UpdateDRepActivity` inferred "not found" from the UPDATE's affected-rows
count being zero. `go-sql-driver/mysql` reports rows *changed*, not rows
*matched*, unlike sqlite3/lib/pq — so a DRep voting again with the same
resulting activity/expiry epoch matched a real row but changed nothing, and
was misreported as not-found. Fixed with an explicit existence check
(`GetDrepByCredential`) before the UPDATE, kept dialect-neutral and scoped
to this one query rather than a connection-wide `CLIENT_FOUND_ROWS` flag
(which would silently change affected-rows semantics for other unrelated
callers in the same package).

**`test(conformance): exercise real sqlite/postgres/mysql backends`**
The main rewrite. `DingoStateManager`/`DingoStateProvider` now read and
write through a real backend instead of in-memory maps. New tests prove the
audit's specific acceptance criteria: state surviving a restart (close,
reopen against the same store), a rolled-back write not being visible, an
epoch transition persisting a real stake-snapshot row, and invalid/
unreachable backend configuration failing construction for both Postgres
and MySQL.

**`fix(test): break postgres/conformance import cycle`**
Making `internal/test/conformance` genuinely import
`database/plugin/metadata/postgres` (needed for the rewrite above) created
a cycle with that package's own `conformance_test.go`, which already
imported two small DSN-formatting helpers from `internal/test/conformance`.
Moved both helpers to `internal/test/storagetest`, a neutral package
neither side cycles back through.

## Testing

- `go build ./...`, `go build -tags dingo_extra_plugins ./...`,
  `go build -tags dingo_db_integration ./...` — clean
- `go vet ./...` and `go vet -tags dingo_extra_plugins ./...` (whole repo,
  compiles all test files — this is what catches an import cycle) — clean
- `golangci-lint run`, `nilaway`, `make gorm-check`, `make import-boundaries`,
  `make docs-parity` — all clean on every touched package
- SQLite: `go test ./internal/test/conformance/...` — 315/315 vectors, plus
  restart/rollback/epoch-transition tests, all passing (~25s)
- PostgreSQL + MySQL (local Docker, `internal/test/conformance/docker-compose.yml`):
  `go test -tags dingo_extra_plugins ./internal/test/conformance/... -run 'Postgres|Mysql'`
  — 315/315 vectors on both backends, plus restart/rollback/invalid-DSN
  tests, all passing (~18 min locally against real network SQL)
- CI (`go-test (Linux)`, full `go test -tags dingo_extra_plugins ./...` plus
  the `sqlstore-database-integration` step): **passing, 29m8s**. This is
  meaningfully slower than before this change — expected and accepted,
  since the whole point is that these tests now hit real databases instead
  of an in-memory mirror. A follow-up to batch the per-vector database
  reset into fewer round trips (currently ~35 sequential `TRUNCATE`
  statements per vector) is a promising, low-risk way to cut this further
  without reducing coverage, tracked separately rather than folded into
  this PR.
- `DATABASE.md`/`ARCHITECTURE.md` checked, not affected — no schema,
  query-semantics, or architectural-boundary change

All CI checks pass. Full local test logs retained.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed DRep activity updates that could incorrectly report a missing DRep when submitting an unchanged vote.
  * Improved pool data loading to prevent database connection issues during repeated reads.
  * Improved PostgreSQL and MySQL state handling across restarts, rollbacks, and connection failures.
  * Ensured state queries consistently read persisted backend data, including UTxOs, pools, DReps, committees, and governance proposals.

* **Tests**
  * Expanded coverage for real PostgreSQL and MySQL integrations, transaction behavior, credential validation, and database recovery.


arepala-uml:
## Testing results

**Build / static analysis** — clean across all build-tag combinations
(`go build`, `go vet` whole-repo, `golangci-lint`, `nilaway`, `make gorm-check`,
`make import-boundaries`, `make docs-parity`).

**SQLite**
```
go test ./internal/test/conformance/...
```
315/315 vectors passing, plus restart/rollback/epoch-transition proof tests. ~25s.

**PostgreSQL + MySQL** (local Docker, `internal/test/conformance/docker-compose.yml`)
```
go test -tags dingo_extra_plugins ./internal/test/conformance/... -run 'Postgres|Mysql'
```
315/315 vectors passing on **both** backends, plus restart/rollback/invalid-DSN
proof tests for both. 0 failures. ~18 min locally against real network SQL.

**CI**
`go-test (Linux)`: **passing, 29m8s**. All 11 checks green on this PR.

`DATABASE.md`/`ARCHITECTURE.md` checked, not affected — no schema, query-semantics,
or architectural-boundary change.

Full local logs retained.

arepala-uml:
**NOTE: `go-test (Linux)` CI time has increased significantly.**

- Before this PR: `go-test (Linux)` completed in under 10 minutes.
- With this PR: `go-test (Linux)` takes **~29 minutes** (29m8s on this run).

The increase is isolated to one package:
```
internal/test/conformance   14.546s  (before)
internal/test/conformance   1118.283s  (after, ~18.6 min)
```

This is expected, not a regression: the conformance suite now runs ~315
ledger-rule vectors against real SQLite/Postgres/MySQL backends instead of
an in-memory mock, per the audit requirement on #1900 that in-memory mirror
state cannot satisfy acceptance. Most of that time is the per-vector
database reset (~35 sequential `TRUNCATE` statements per vector) rather
than the ledger-rule checks themselves.

Flagging this explicitly so it's visible before merge, not discovered later
as an unexplained CI slowdown. Open discussion point: whether to accept
this as-is or follow up with a reset-batching optimization to cut it down.

internal/test/conformance/state_manager_postgres.go
```diff
@@ -16,7 +16,245 @@
...
+// first real commit fails that check.
+//
+// Unlike the schema (truncated in place between vectors -- see
+// wipeMetadata's doc comment), this directory is never cleared: stale
+// blob entries from an earlier, now-truncated vector are keyed by that
+// vector's own transaction hashes, which the metadata rows a later read
+// would need to resolve into no longer exist, so they're simply never
+// looked up again -- harmless, not read, and not worth the extra I/O of
+// clearing on every Reset.
+var postgresConformanceBlobDir = filepath.Join(
```
wolf31o2: Please namespace the remote schema/database and paired Badger directory per test process or run, and clean both together. These fixed globals are shared by every checkout and invocation: concurrent runs contend on the Badger lock and can truncate each other's backend; after the documented disposable compose database is recreated, the never-cleared local commit timestamp is paired with a fresh remote store and construction fails. The MySQL path at `state_manager_mysql.go:59` has the same issue; the restart tests already demonstrate the safe unique-pair pattern.

wolf31o2 (changes requested):
Two additional findings are on unchanged lines exercised by this PR:

- **Backend bypass — `internal/test/conformance/state_provider.go:572`:** Read the account's DRep delegation from `manager.db`, matching production `LedgerView.DRepDelegation`. The backend suites currently validate withdrawals against `govState`, so they can remain green even if Postgres/MySQL never persist or return `account.drep`. A focused probe stored one delegation only in SQLite and a different one only in the mirror; this method returned the mirror value.
- **PostgreSQL DSN parsing — `internal/test/conformance/conformance_postgres_test.go:91`:** Pass every environment-derived libpq value through `storagetest.EscapeLibpqValue`, as the plugin and migration tests do. With the legal password `review pass'word\tail`, `pgx.ParseConfig(postgresConformanceDSN())` parsed only `review`, so this suite cannot use credentials the existing plugin suite accepts.

arepala-uml:
Both findings fixed.

**Backend bypass (`state_provider.go:572`, `1f81a71f`)** — `DRepDelegation` now reads `account.drep`/`account.drep_type` through `manager.db.GetAccountByCredential`, matching `ledger.LedgerView.DRepDelegation`, instead of the `govState` mirror. Added `TestDRepDelegationReadsRealBackendNotGovStateMirror`, which reproduces your probe exactly: stores one delegation only in `govState` and a different one only in the real backend, and asserts the backend's value is returned. Confirmed the test fails against the pre-fix code (returns the mirror's value) and passes against the fix.

**PostgreSQL DSN parsing (`conformance_postgres_test.go:91`, `5450ed27`)** — every DSN component now goes through `storagetest.EscapeLibpqValue`. Added `TestPostgresConformanceDSNEscapesSpecialCharacterPassword` using your exact `review pass'word\tail` password; confirmed it fails pre-fix (`pgx.ParseConfig` yields `review`) and passes with the fix (`cfg.Password == "review pass'word\tail"`).

Also audited the rest of `state_provider.go` for the same backend-bypass class: `RewardAccountBalance` and `CommitteeMember`/`CommitteeMembers` are legitimate documented exceptions (reward balances are harness-injected synthetic input with no real-backend equivalent; the committee mirror fallback only applies to not-yet-enacted proposals, checked after a real-backend lookup). No other instances found.

Full validation after all three fixes (including the namespacing fix below): SQLite 315/315, Postgres 315/315, MySQL 315/315, restart/rollback/invalid-DSN tests all green on both backends. Full Postgres+MySQL suite completed in ~16.6 min (`995.8s`), consistent with pre-regression baseline.

internal/test/conformance/state_manager_postgres.go
arepala-uml: Fixed in `17b48e96`. `postgresProcessSchema`/`postgresProcessBlobDir` and `mysqlProcessDatabase`/`mysqlProcessBlobDir` are now derived once per process (`os.Getpid()` + a nanosecond timestamp), unique across every checkout/invocation, and shared by every `NewDingoPostgresStateManager`/`NewDingoMysqlStateManager` call within that process — so the migration cost is still paid exactly once per run, not once per call (an intermediate per-call design I tried first paid it on every call and turned a ~16 minute suite into 1.5+ hours before I caught it and reverted).

A new `TestMain` (`conformance_main_test.go`) drops the process's schema/database and removes its paired blob directory once, after every test in the process finishes — mirroring the restart tests' existing explicitly-managed-schema pattern for the process-wide case. `DingoStateManager.Close` no longer carries a `dropMetadata` hook: nothing sets it under this design, since an individual manager must never drop a schema/database a sibling manager in the same process may still be using.

Verified end-to-end against fresh `docker compose down -v && up -d` containers: Postgres 315/315 vectors, MySQL 315/315 vectors, restart/rollback/unreachable-host/bad-credentials tests all green on both backends, full suite in ~16.6 minutes, and confirmed via direct DB inspection that `TestMain` actually dropped the process schema/database after the run (no leftover `conformance_*` schema or `dingo_conformance_*` database).

internal/test/conformance/state_provider.go
```diff
@@ -296,48 +541,90 @@ func (p *DingoStateProvider) CommitteeMembers() ([]common.CommitteeMember, error
...
+			)
+		})
+		if err != nil {
+			continue
+		}
+		if drep != nil && drep.Active {
+			return &common.DRepRegistration{Credential: credential}, nil
+		}
 	}
 	return nil, nil
```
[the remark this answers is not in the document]
arepala-uml: Fixed in `f155aea4`. `DRepRegistration` now continues the credential-tag loop only on `models.ErrDrepNotFound`; every other error (dropped connection, real query failure) is returned instead of being treated as "not registered."

Added `TestDRepRegistrationPropagatesBackendErrors`: calls `DRepRegistration` against a closed manager (a real, non-`ErrDrepNotFound` failure) and asserts an error is returned. Confirmed it fails against the pre-fix code (silently returns `(nil, nil)`) and passes with the fix. Also reran the full Postgres+MySQL vector suites (315/315 both) to confirm this doesn't regress real-backend conformance.

internal/test/conformance/README.md
```diff
@@ -76,9 +80,40 @@ connect to the same `dingo_test` database. Since `go test ./...` runs
...
+would also destroy `database/plugin/metadata/postgres`'s own concurrently
+running tests' tables in the shared `dingo_test` database. Reset instead
+`TRUNCATE`s every table in the `conformance` schema in place, over a
+separate admin connection, discovering the table list from
+`information_schema` rather than hardcoding it (see
+`state_manager_postgres.go`'s `wipeMetadata`). This keeps the already-open
+store's connection pool live throughout -- no close, no reopen, no
+re-migration -- which is what keeps the cost of a Reset (and so the whole
+vector suite, which resets once per vector) from being a real
+close/reopen/re-migrate network round trip every time.
```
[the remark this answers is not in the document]
arepala-uml: Fixed in `280ee434`. Updated both the PostgreSQL and MySQL sections to describe the process-scoped schema/database names and `sync.Once`-created temporary blob directories, and documented `TestMain`'s cleanup responsibility. Also added `conformance_main_test.go` to the Files table.

database/plugin/metadata/sqlstore/drep.go
```diff
@@ -616,6 +616,31 @@ func (s *Store) UpdateDRepActivity(
...
+	// caller sharing that connection (several elsewhere in this package
+	// rely on the *current* semantics -- see EnsureOffchainMetadataPointers
+	// in offchain_metadata.go, which would silently over-count duplicates
+	// as newly created rows under CLIENT_FOUND_ROWS).
+	existing, err := s.GetDrepByCredential(credentialTag, credential, true, txn)
+	if err != nil {
+		return fmt.Errorf("check drep exists before activity update: %w", err)
+	}
+	if existing == nil {
+		return models.ErrDrepActivityNotUpdated
```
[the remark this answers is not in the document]
arepala-uml: Verified but out of scope for this PR. This PR's diff never touches `database/plugin/metadata/sqlstore/drep.go` or `database/drep.go` (confirmed via `git log main..HEAD -- database/plugin/metadata/sqlstore/drep.go`, no output). Tracing the only production call chain into `Store.UpdateDRepActivity` (`database.Database.UpdateDRepActivity` → `ledger/governance/processing.go`'s three call sites), the `txn` argument is always non-nil by the time it reaches `Store.GetDrepByCredential`/`UpdateDRepActivity`: `Database.UpdateDRepActivity` itself opens a transaction when its own `txn` parameter is nil, before ever calling into the store. This PR's own conformance harness (`state_manager.go`) also always supplies a real transaction from `m.db.Transaction(true)` to every governance call — it never exercises the nil-txn branch either.

The underlying claim (a read/write DB split could see a stale row on the readDB-backed existence check) looks plausible as a general concern for whichever code path *can* reach this with a nil transaction, but I can't find one on `main` today, and this PR doesn't add one. Recommend tracking separately rather than fixing here, to keep this PR scoped to the conformance test infrastructure it's about.

internal/test/conformance/conformance_main_test.go
```diff
@@ -0,0 +1,90 @@
...
+				postgresProcessSchema,
+			); err != nil {
+				log.Printf(
+					"conformance: cleanup postgres process schema %q: %v",
+					postgresProcessSchema,
+					err,
+				)
+			}
+		}
+		if err := os.RemoveAll(postgresProcessBlobDir); err != nil {
```
wolf31o2, line 69: Please make any process-owned schema/database/blob cleanup failure set a nonzero exit code while still attempting all cleanup paths. These branches only log errors, then `os.Exit(code)` preserves a passing result. A focused forced `RemoveAll` permission failure logged `permission denied`, but `go test` exited 0 and left the per-run resource behind.

wolf31o2 (changes requested):
One additional documentation issue:

- **PR CI coverage — `internal/test/conformance/README.md:76` and `:151`:** The PR workflow explicitly omits race instrumentation at `.github/workflows/go-test.yml:130-133`; `-race` runs only during publish/release validation. Please correct both claims rather than overstating the pull-request CI gate.

arepala-uml:
Fixed in \`d82d97cc\`. Both claims (README.md:76 and :151) now name the actual PR CI command (\`go test -tags dingo_extra_plugins ./...\`, no \`-race\`) and note that \`-race\` only runs during publish/release validation, matching \`.github/workflows/go-test.yml\`'s pull-request job.

internal/test/conformance/conformance_main_test.go
arepala-uml: Fixed in `302e15b7`. `TestMain` still attempts every cleanup path unconditionally (a failed drop no longer skips the paired blob-directory removal or the other backend's cleanup), but now tracks whether any step failed and folds that into the exit code afterward: a cleanup failure fails an otherwise-passing run, while a genuine test failure's exit code is never downgraded.

Added `TestProcessCleanupExitCodeFailsOnCleanupFailure`, extracting the exit-code aggregation into its own function so it's tested directly rather than only through `TestMain`'s own `os.Exit`. Confirmed it fails against the pre-fix behavior and passes with the fix. Also reran a live Postgres+MySQL smoke test and confirmed the run still exits 0 on success and cleanup still removes the process schema/database/blob directories.

internal/test/conformance/state_provider.go
```diff
@@ -205,39 +376,38 @@ func (p *DingoStateProvider) RewardAccountBalance(
...
+	member, err := p.realCommitteeMember(coldKey)
+	if err != nil {
+		return nil, err
+	}
+	if member != nil {
 		return member, nil
 	}
 
 	// Check proposed members from governance state
 	if memberInfo := p.manager.govState.GetCommitteeMember(coldKey); memberInfo != nil {
```
wolf31o2: **P1 — Require the configured backend to supply enacted committee members.** `LoadInitialState` copies the initial committee into `govState` before writing the same members to the database. If the backend drops or cannot read a row, `realCommitteeMember` returns `nil` and this fallback returns the mirror copy, so the vector remains green without exercising the configured store. Pending additions are already handled separately below through `govState.Proposals`; `govState.CommitteeMembers` also contains initial or enacted members. Remove this fallback and the analogous merge in `CommitteeMembers`, and add a disagreeing mirror/backend regression like the DRep case.

internal/test/conformance/conformance_postgres_test.go
```diff
@@ -195,3 +227,184 @@ func TestRulesConformanceVectorsWithResultsPostgres(t *testing.T) {
...
+	// same process from resetting it concurrently. Manage a schema
+	// explicitly instead, unique to this test run so it cannot collide with
+	// postgresProcessSchema or any other test's schema, and clean it up once
+	// both managers are done.
+	blobDataDir := t.TempDir()
+	dsn := postgresConformanceDSN()
+
+	schema := fmt.Sprintf("conformance_restart_%d", time.Now().UnixNano())
+	t.Cleanup(func() {
+		_ = dropPostgresSchema(dsn, schema)
```
wolf31o2: **P2 — Make restart cleanup failures fail the test.** Discarding this error lets CI stay green while leaking the real Postgres schema. Report it through the test (`require.NoError` or `t.Errorf`); apply the same fix to the ignored `dropMysqlDatabase` error in `conformance_mysql_test.go`.

internal/test/conformance/README.md
```diff
@@ -69,16 +73,57 @@ Without a `POSTGRES_PASSWORD` or `POSTGRES_DSN` set, both Postgres tests
...
+
+**Reset semantics.** Between vectors, `DingoStateManager.Reset()` does not
+call the metadata store's own `Resettable.Reset` (`database/plugin/metadata/postgres`'s
+`resetDatabase`): that callback drops tables outright, without recreating
+them, requiring a fresh migration run before the store is usable again --
+and, more importantly, scans and drops tables across *every* non-system
+schema in the target database, not just `conformance`, so calling it here
+would also destroy `database/plugin/metadata/postgres`'s own concurrently
+running tests' tables in the shared `dingo_test` database. Reset instead
+`TRUNCATE`s every table in the `conformance` schema in place, over a
```
wolf31o2: **P3 — Document the process-scoped schema name.** This paragraph still says Reset scans and truncates the fixed `conformance` schema, but the isolation code now creates `conformance_<pid>_<timestamp>`. Refer to the process-scoped conformance schema here and on line 116 so the reset boundary matches the implementation.

wolf31o2 (changes requested):
Three findings remain. The prior backend-isolation, DRep-read, DSN-escaping, cleanup-exit, and README race-coverage findings are fixed on this head.

internal/test/conformance/state_provider.go
arepala-uml: Fixed in `4331c7bc`. Removed both fallbacks: `CommitteeMember`'s `govState.GetCommitteeMember` merge and `CommitteeMembers`' equivalent loop. `CommitteeMember` still falls back to the mirror only for a member proposed by a pending (not yet enacted) `UpdateCommittee` action, since that's the one case with no real `committee_member` row to read yet.

Added `TestCommitteeMemberReadsRealBackendNotGovStateMirror`: stores a member only in `govState` (the shape `LoadInitialState` always leaves behind, with no corresponding backend row) and asserts both `CommitteeMember` and `CommitteeMembers` report it absent. Confirmed it fails against each fallback independently (reverted singular, then reverted plural) and passes with both fixed. Also reran the full Postgres+MySQL vector suites (315/315 both) to confirm removing these fallbacks doesn't regress any real vector -- `LoadInitialState` already writes the initial committee through `SetCommitteeMembers` to the real backend, so nothing relied on the mirror fallback for a real pass.

internal/test/conformance/conformance_postgres_test.go
arepala-uml: Fixed in `478aeec5`. Both restart tests now report their cleanup error through `t.Errorf` instead of discarding it. Reran both restart tests live against Postgres and MySQL to confirm the change doesn't affect the success path.

internal/test/conformance/README.md
arepala-uml: Fixed in `07018620`. Both references (this paragraph and line 116) now say "process-scoped conformance schema" instead of naming the old fixed `conformance` schema.

arepala-uml:
Fixed in `de6fac2d`: two CodeRabbit nitpicks from the first review pass that were listed in the review-body summary but had no anchored inline thread, so they'd gone unaddressed.

- Added `defer txn.Release()` after `m.db.Transaction(true)` in `TestDingoStateManagerRollbackDiscardsWrites` (state_manager_backend_test.go), `TestProcessEpochAgainstRealBackend` (state_manager_backend_test.go), `TestNewDingoMysqlStateManagerRollbackDiscardsWrites`, and `TestNewDingoPostgresStateManagerRollbackDiscardsWrites` — matching every other transaction site in this package, so a failed `require` mid-test doesn't leave a write transaction held when the test's own `defer m.Close()` runs.
- `TestNewDingoPostgresStateManagerBadCredentialsFails` now skips under a `POSTGRES_DSN` override, matching its MySQL analog, instead of building a DSN from `POSTGRES_HOST`/`POSTGRES_PORT` defaults that may point at no server.

Verified: `go build ./...`, `go vet -tags dingo_extra_plugins ./...` clean; `go test ./internal/test/conformance/...` (sqlite) 315/315 vectors plus all regression tests pass.

internal/test/conformance/state_provider.go
```diff
@@ -253,91 +415,196 @@ func (p *DingoStateProvider) CommitteeMember(
...
+		ExpiryEpoch: expiryEpoch,
+	}
+
+	auth, err := withBadConnRetry(func() (*models.AuthCommitteeHot, error) {
+		return p.manager.db.GetCommitteeMember(coldKey[:], nil)
+	})
+	if err != nil && !errors.Is(err, models.ErrCommitteeMemberNotFound) {
+		return nil, fmt.Errorf("lookup committee hot key: %w", err)
+	}
+	if auth != nil {
```
wolf31o2, line 468: **P1 — Do not return an authorization superseded by resignation.** This returns the latest authorization row even when a later resignation exists, producing both `HotKey != nil` and `Resigned == true`; `CommitteeMembers` repeats the same composition. Production uses active committee members, and the upstream state removes the hot key on resignation. Query the active authorization or clear `HotKey` when resignation wins, and cover both provider methods.

internal/test/conformance/state_provider.go
```diff
@@ -113,40 +196,125 @@ func (p *DingoStateProvider) TimeToSlot(t time.Time) (uint64, error) {
...
+
+// latestPoolRetirementEpoch returns the most recently recorded retirement
+// epoch for pool, or nil if it has none.
+func latestPoolRetirementEpoch(pool *models.Pool) *uint64 {
+	if len(pool.Retirement) == 0 {
+		return nil
+	}
+	epoch := pool.Retirement[0].Epoch
+	for _, retirement := range pool.Retirement[1:] {
+		if retirement.Epoch > epoch {
```
wolf31o2: **P1 — Derive pending retirement from the latest certificate event.** A later retirement certificate replaces the prior schedule even when it moves retirement earlier, and a later pool registration cancels pending retirement. Taking the maximum epoch returns 10 after `retire@10 → retire@5` and still reports retirement after `retire@10 → re-register`, while `poolIsActive` below correctly orders those events. Reuse that event ordering, cover replacement and cancellation, and correct the corresponding production boundary so this adapter can detect rather than duplicate the defect.

internal/test/storagetest/postgres_dsn.go
```diff
@@ -23,6 +23,12 @@ import (
 // keyword/value connection string, so a password containing a space,
 // single quote, or backslash -- all legal in a Postgres password --
 // produces a well-formed DSN instead of breaking the conninfo parse.
+//
+// Lives here, not in internal/test/conformance or a metadata plugin
```
wolf31o2, line 27: **P2 — Preserve existing PostgreSQL `options`.** In the shared `PostgresDSNWithSearchPath` helper below, `query.Set` replaces an operator-supplied value instead of adding `search_path`; for example, `options=-cstatement_timeout=5000` is lost when the conformance schema is applied. Merge the existing options with the search-path option and cover both URL and keyword DSNs.

wolf31o2 (changes requested):
The prior backend-isolation, DRep-read, cleanup, and documentation findings are fixed or superseded. Three state/DSN issues remain.

internal/test/conformance/state_provider.go
arepala-uml: Fixed in `79795646`, in both places.

**Production (`ledger/view.go`):** `PoolCurrentState` now tracks the latest registration's and retirement's ordering keys (AddedSlot, CertificateID) and reports no pending retirement when the registration is newer, mirroring `poolIsActive`'s own ordering rule. One nuance worth flagging: `Store.GetPool` (`database/plugin/metadata/sqlstore/pool.go`) always loads only the single latest-by-ordering row per pool (`loadPoolAssociations(..., latestOnly=true)`, hardcoded regardless of the `includeInactive` argument), so the "max epoch across every retirement row" half of this was actually unreachable through this call path -- `pool.Retirement` never has more than one element here. The registration-supersedes-retirement check was the real, reachable defect (confirmed by reverting just that half and watching the new test fail on exactly that assertion, not the epoch-value one). I kept the ordering-based selection anyway since it's correct either way and matches the documented contract precisely, but wanted to be upfront that only the cancellation half is independently provable as a regression via this path.

**Harness (`state_provider.go`):** replaced `latestPoolRetirementEpoch` (max-based) with `pendingPoolRetirementEpoch`, which reuses `poolIsActive`'s comparison (extracted into `registrationSupersedesRetirement`) instead of re-deriving it, so the adapter now detects rather than duplicates whatever the backend does.

Added a regression test on each side (`ledger/view_test.go` driving real certificates through `SetTransactionMetadataOnly`; `state_manager_backend_test.go` doing the same through `ApplyTransaction`), confirmed each fails on the cancellation assertion without the fix. Also reran the full Postgres+MySQL vector suites (315/315 both) to confirm no regression.

internal/test/storagetest/postgres_dsn.go
arepala-uml: Fixed in `c387b8df`. `PostgresDSNWithSearchPath` now merges the search_path flag onto any existing `options` value instead of replacing it, for both the URL and keyword/value forms (a small libpq keyword/value tokenizer handles the quoted-value case for the latter).

Added `TestPostgresDSNWithSearchPathPreservesExistingOptions` covering both forms, with and without a pre-existing `options` value; confirmed it fails against the pre-fix code for exactly the two "with existing options" cases (the "without" cases correctly still pass either way, so they don't false-positive).

internal/test/conformance/state_provider.go
wolf31o2: Reverified at f656deb2: both provider methods still read the latest authorization, assign HotKey, and then independently set Resigned. An auth → resignation sequence still returns an active hot key with Resigned true, so this remains blocking.

wolf31o2 (changes requested):
The existing committee-resignation P1 remains on this head.

arepala-uml:
> The existing committee-resignation P1 remains on this head.

Addressed in c89e843b. Both CommitteeMember and CommitteeMembers now clear HotKey when a later resignation supersedes the latest authorization, preventing the invalid HotKey != nil && Resigned == true state.
Added TestCommitteeMemberResignationClearsHotKey, covering authorization → resignation and subsequent reauthorization through both provider methods. The test fails against the previous implementation at the stale HotKey assertion and passes with the fix. The full conformance test package also passes.

wolf31o2 (approved):
The committee-resignation blocker is fixed on this head: both provider paths clear stale hot-key authorization after resignation, with authorization, resignation, and reauthorization coverage. All prior findings are resolved and CI is green.

Issue #3495, opened 2026-08-25
ledger: verify block transaction validity outcomes


wolf31o2:
## Problem

Validated block application does not require the locally evaluated Plutus result to match the transaction's declared validity flag. Transactions marked invalid skip script execution, while pre-Dijkstra script failures on transactions marked valid are logged and accepted.

## Required behavior

- run phase-1 checks for every transaction in a validated block
- evaluate phase-2 scripts for both declared validity outcomes
- accept a transaction marked invalid only when local execution fails
- reject a transaction marked valid when local execution fails
- reject mismatches before transaction state is committed
- cover both matching outcomes and both mismatch directions with regression tests

wolf31o2:
Validated against merged PR #3603 (8ed5fc02): all stated validity-outcome requirements and regression coverage are present. Closing as resolved.

PR #3603, opened 2026-08-28
fix(ledger): enforce declared Plutus validity outcomes

wolf31o2:
## Summary
- enforce declared Plutus validity outcomes during block application
- validate phase 1 for every transaction and evaluate phase 2 for both declared outcomes
- add Dijkstra V3/V4 matching and mismatch coverage, including rollback/collateral state checks

## Validation
- go test ./ledger/eras -run 'TestValidateTxDijkstraRequiresDeclaredValidityToMatchGuardingExecution|TestValidateTxDijkstraDoesNotTreatPhase1FailureAsPhase2Failure' -count=1 -timeout=90s\n- go test ./ledger/eras -count=1 -timeout=120s\n- go test ./ledger -count=1\n- git diff --check origin/main..HEAD\n\nCloses #3495

---
## Summary by cubic
Enforces declared Plutus validity outcomes during block application so the local evaluator's phase-2 result must match the transaction's `isValid` flag. Old behavior skipped phase 2 for declared-invalid transactions and logged-and-trusted script failures in pre-Dijkstra eras; new behavior rejects a block transaction when the declared outcome doesn't match before its regular-input or collateral effects commit, and treats local Plutus failures as deterministic replay-recovery triggers.

**Details**
- Phase 1 runs for every transaction; phase 2 runs for both declared outcomes across Alonzo, Babbage, Conway, and Dijkstra.
- Dijkstra phase 2 evaluates guarding scripts by copying the transaction with `TxIsValid` forced true, since the upstream evaluator skips declared-invalid transactions.
- Only a `PlutusScriptFailedError` satisfies a "declared invalid" outcome; other phase-2 errors remain hard failures.
- Blocks at least `k` deep keep the replay-only phase-2 skip, trusting declared outcomes only inside that immutable window.
- Removes the prior `trustDijkstraTxValidationError` trust path; a validation error now always rejects a declared-valid transaction and logs a rejecting warning.

<sup>Written for commit 6ca400cd91e5686e1ac269c53d2e4ce787bbe34d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3603?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction validation across Alonzo, Babbage, Conway, and Dijkstra eras.
  - Transactions now undergo complete validation regardless of their declared validity status.
  - Blocks are rejected when local script execution does not match a transaction’s declared outcome.
  - Prevented regular inputs, collateral, or transaction outputs from being applied when validation fails.
  - Preserved trusted replay behavior for sufficiently deep historical blocks.


wolf31o2:
The earlier CodeRabbit findings are addressed in the current head `0c2ae698`: the validation warning now describes the rejection behavior accurately and the exported error text follows the lowercase convention. Current required checks are green.

ledger/eras/babbage.go
```diff
@@ -393,25 +400,36 @@ func ValidateTxBabbage(
...
-					ScriptHash: tmpScript.Hash(),
-					Tag:        redeemer.Tag,
-					Index:      redeemer.Index,
-					Err: fmt.Errorf(
-						"script exceeded declared budget: used (%d cpu, %d mem), declared (%d cpu, %d mem)",
-						usedBudgetV2.Steps, usedBudgetV2.Memory,
-						redeemer.ExUnits.Steps, redeemer.ExUnits.Memory,
-					),
-				}
+				return validatePlutusOutcome(
```
cryptodj413, line 414: **MERGE BLOCKER — untested (verified by mutation):** This `PlutusV2Script` branch's two `validatePlutusOutcome` calls (here and line 403) are never exercised by any test. `TestValidateTxRequiresDeclaredValidityToMatchExecution` builds its fixture with `lang.LanguageVersionV1` / `lcommon.PlutusV1Script` (plutus_validity_outcome_test.go:267,279) and reuses that same transaction for the babbage sub-test, so dispatch always takes the `PlutusV1Script` case above (line ~301), never this one. I reverted both V2-branch calls to their pre-PR raw returns and `go test ./ledger/eras/` still passed clean.

PlutusV2 is the dominant script version on real Babbage/Conway chain data, so the era/version pair that matters most for actual mainnet blocks has zero negative-case coverage for this change. Please add a PlutusV2 case to the table (Babbage and Conway) covering both mismatch directions.

ledger/eras/alonzo.go
```diff
@@ -316,25 +312,36 @@ func ValidateTxAlonzo(
 				evalContext,
 			)
 			if err != nil {
-				return err
+				return validatePlutusOutcome(
```
cryptodj413, line 315: **MERGE BLOCKER — untested (verified by mutation):** This is the `s.Evaluate` *error* site's `validatePlutusOutcome` call. The test's failure-case fixture always uses `lcommon.ExUnits{}` (plutus_validity_outcome_test.go:420,427), which only ever trips the budget-exceeded site at line 326 below, never this one. Reverting just this call (keeping line 326 as-is) still leaves `go test ./ledger/eras/` and `go test ./ledger/` green.

Please add a fixture that forces a genuine evaluator error (not merely a budget overrun) so this branch is covered.

ledger/eras/conway.go
```diff
@@ -454,20 +449,18 @@ func validateTxPlutusConway(
...
 	}
-	return validateTxPlutusConwayWithContext(
+	phase2Err := validateTxPlutusConwayWithContext(
 		tx,
 		ls,
 		pp,
 		plutusCtx,
 		validateRequiredRedeemers,
 	)
+	return validatePlutusOutcome(tx, phase2Err)
```
cryptodj413, line 463: **MERGE BLOCKER — untested, and the function has no caller (verified):** `grep -rn 'ValidateTxPlutusConway' --include=*.go . | grep -v _test` returns only this definition — no production call site anywhere in the repo. It's exported and its declared-invalid-transaction behavior changed with this PR, but nothing exercises that change, and it's unclear this entry point is reachable in the real block-application path at all.

Either cover it with the same table-driven test as the other eras, or — if it's genuinely dead/unused — that's worth deciding explicitly rather than leaving an exported, behavior-changed, untested, uncalled function in the tree.

ledger/state.go
```diff
@@ -6282,6 +6249,26 @@ func (ls *LedgerState) ledgerProcessBlock(
 					err = nil
 				}
 				if err != nil {
```
cryptodj413, line 6254: **HIGH PRIORITY, non-blocking — deserves explicit sign-off:** This PR deletes the only bypass that existed for a pre-Dijkstra (Alonzo/Babbage/Conway) declared-valid transaction whose local Plutus evaluator disagrees with the block producer (the removed hunk that used to log a warning and set `err = nil` for `PlutusScriptFailedError` outside Dijkstra). That's the letter of issue #3495, and ARCHITECTURE.md now documents it accurately. But two things compound the risk:

1. No config knob exists for pre-Dijkstra eras analogous to `SkipDijkstraTxValidation`/`trustDijkstraTxValidationError` (`grep -n 'SkipDijkstraTxValidation\|SkipAlonzoTxValidation\|SkipBabbageTxValidation\|SkipConwayTxValidation' config.go internal/config/config.go` finds only the Dijkstra one) — an operator has no escape hatch if a real plutigo/local-VM discrepancy surfaces on live mainnet/testnet Conway blocks (this only applies outside the `k`-deep replay window per `shouldSkipPhase2ValidationForBlock`, i.e. exactly the live, actively-produced blocks that matter most).
2. The resulting `&txValidationError{...}` (line 6325) is not classified as deterministic in `isDeterministicTxValidationError` (replay_recovery.go:389-395, unchanged by this PR) — only `shelley.DuplicateInputError` and `eras.DuplicateInputByronError` are. So instead of a clean, diagnosable halt, an at-tip Plutus disagreement falls into `recoverAtTipFromTxValidationError`'s progressive rewind-and-peer-rotation loop (replay_recovery.go:819,893-899) — which will never succeed, since the disagreement is deterministic and reproduces identically from any peer.

Recommend: (a) confirm this strictness change is the deliberate intent for a *live* (non-replay) window, not just historical replay, since it removes graceful degradation dingo previously had; (b) classify a Plutus-outcome mismatch as deterministic (or otherwise bound the rewind) so this fails clearly rather than looping; (c) consider whether an equivalent operator escape-hatch config is warranted for pre-Dijkstra eras.

ledger/eras/dijkstra.go
```diff
@@ -258,40 +262,52 @@ func ValidateTxDijkstra(
...
+	if err != nil {
+		return err
+	}
+	phase2Err := gdijkstra.UtxoValidatePlutusScripts(
+		phase2Tx,
+		slot,
+		ls,
+		tmpPparams,
+	)
+	return validatePlutusOutcome(tx, phase2Err)
```
cryptodj413, line 284: **Non-blocking:** `dijkstraTransactionForPhase2`'s type assertion (line ~298-311 below) now runs before any redeemer check and hard-rejects any non-`*gdijkstra.DijkstraTransaction` unconditionally, including a script-free transaction that the upstream `gdijkstra.UtxoValidatePlutusScripts` rule would previously have short-circuited on (no redeemers). I confirmed the production path is safe today (`resolveValidationEra` dispatches on `tx.Type()`, so only real `*DijkstraTransaction` values reach here), but `ledger/dijkstra_pool_margin_floor_e2e_test.go:85,103,120,139` only asserts on error *substrings*, so it would silently absorb this new error rather than surface a behavior change. Suggest gating on `txHasRedeemers(tx)` first, or falling back to the upstream rule when the assertion fails.

ledger/block_transaction_validity_test.go
```diff
@@ -0,0 +1,373 @@
...
+						Body: babbage.BabbageBlockHeaderBody{
+							BlockNumber: 1,
+							Slot:        10,
+							ProtoVersion: babbage.BabbageProtoVersion{
+								Major: gdijkstra.MinProtocolVersionDijkstra,
+							},
+						},
+					},
+				},
+				txs: []lcommon.Transaction{tx},
```
cryptodj413, line 230: **Non-blocking, missing test:** The PR's stated guarantee is that a mismatch "rejects the block transaction before its regular-input or collateral effects can commit" for the *whole block*, but every test here uses a single-transaction block (`txs: []lcommon.Transaction{tx}`). I traced this is mechanically correct (the whole batch runs inside one write DB txn; `delta.applyWithoutRecordingDonations` at state.go:6341 writes through that same txn, rolled back on error) — but there's no test proving it: a two-transaction block where transaction 0 applies cleanly and transaction 1 mismatches, asserting transaction 0's UTxO rows are NOT persisted after the block is rejected.

ARCHITECTURE.md
```diff
@@ -2095,6 +2095,17 @@ because prior pot transitions cannot be repaired safely without replay.
 
 The `ledger/eras/` package provides era-specific validation rules for each Cardano era. The default active era table is Byron through Conway. Experimental Dijkstra support is added to the active table when Dingo starts on the `musashi` network (the IOG Leios prototype testnet, matched by network name or magic 164), with `runMode: "leios"`, or with `startEra: "dijkstra"` — see `Config.experimentalDijkstraEnabled`. Keying on the network lets `dingo -n musashi` follow the Musashi testnet past the Conway-to-Dijkstra hard fork without an explicit run mode. The Dijkstra descriptor uses `github.com/blinklabs-io/gouroboros/ledger/dijkstra`, including that release's generated CDDL shape for the nullable Leios/Peras certificate slots.
 
+Validated Alonzo, Babbage, Conway, and Dijkstra block application runs phase 1
```
cryptodj413, line 2098: **Doc drift, non-blocking:** "`ValidateTxDijkstra` evaluates phase 2 independently of the declared validity flag, then reconciles..." attributes the reconciliation to Dijkstra alone, but `ValidateTxAlonzo` (alonzo.go:344), `ValidateTxBabbage` (babbage.go:432), and `ValidateTxConway` (conway.go:246) all do the same thing via the shared `validatePlutusOutcome`. Worth naming all four, or the shared helper, so a reader doesn't conclude only Dijkstra reconciles.

Also a stray hard mid-sentence line wrap a few lines down ("Historical blocks already" / "at least `k` deep"), and a now-stale sentence further down (around line 2303 in the current file) claiming `ValidateTxConway` "already returned early" for redeemer-less transactions — no longer true for a *declared-invalid*, redeemer-less transaction, which now returns the mismatch error (correctly, per cardano-ledger's `ValidationTagMismatch PassedUnexpectedly` — just the doc sentence needs updating).

ledger/eras/plutus_validity_outcome_test.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+func (r *validityOutcomeRedeemers) Value(
+	idx uint,
+	tag lcommon.RedeemerTag,
+) lcommon.RedeemerValue {
+	for _, entry := range r.entries {
+		if entry.key.Index == uint32(idx) && entry.key.Tag == tag {
+			return entry.val
+		}
```
cryptodj413, line 255: **Non-blocking:** Alonzo/Babbage/Conway null out their phase-1 rule lists here (`alonzoUtxoValidationRules = nil` etc.), so unlike `TestValidateTxDijkstraDoesNotTreatPhase1FailureAsPhase2Failure`, there's no equivalent "phase-1 error is not conflated with phase-2 mismatch" test for the other three eras — even though I confirmed the code path is structurally identical across all four (phase-1 errors return before phase 2 runs, independent of `IsValid()`). Worth the same coverage for symmetry, since this PR's whole point is enforcing the same discipline across all four eras.

ledger/block_transaction_validity_test.go
```diff
@@ -0,0 +1,373 @@
...
+type validityOutcomeStateTx struct {
+	*omockledger.MockTransaction
+}
+
+func (t *validityOutcomeStateTx) Consumed() []lcommon.TransactionInput {
+	if t.IsValid() {
+		return t.Inputs()
+	}
+	return t.Collateral()
+}
```
cryptodj413, line 87: **Non-blocking, fixture boundary:** This overrides `Consumed()`/`Produced()` locally on `omockledger.MockTransaction` because the shared `ouroboros-mock` implementation ignores `IsValid` (confirmed: `ouroboros-mock@v0.18.0 ledger/transaction.go:222-241` always returns full inputs/outputs regardless of validity). Per the workspace convention ("never duplicate mocks inside dingo — extend the shared library"), this collateral-aware invalid-transaction behavior belongs in `ouroboros-mock`'s `MockTransaction`, not reimplemented here. Also note line 87 hardcodes `shelley.NewShelleyTransactionInput` for the collateral-return index inside a Dijkstra-era test.

cryptodj413 (changes requested):
Reviewed the current head (0c2ae698) against merge-base d2876583, with a dedicated protocol-domain pass plus my own independent verification (mutation-tested two of the findings myself; both confirmed). The shared `validatePlutusOutcome` rule and its application across Alonzo/Babbage/Conway/Dijkstra is sound and matches cardano-ledger's `UTXOS` semantics. Two confirmed merge blockers are missing test coverage for changed decision points (proven by reverting the fix and getting a green suite); the rest are non-blocking but the removed pre-Dijkstra trust bypass (#3) is a live-operational-risk item that deserves explicit sign-off given it removes the only escape hatch for a Plutus-VM disagreement outside the Dijkstra path. Both previously-flagged CodeRabbit findings (Dijkstra error casing, validation-warning wording) are confirmed fixed at this head and not re-flagged. Full findings in inline comments, ordered by severity below.

wolf31o2:
Addressed the requested blockers in additive commits `b5b84b12` and `6ca400cd`: added PlutusV2 mismatch coverage for Babbage/Conway, a genuine evaluator-error fixture, and deterministic replay-recovery/resync coverage. Focused race tests, formatting, and diff checks pass. Please re-review the current head `6ca400cd`.

cryptodj413 (approved):

PR #3604, opened 2026-08-28
test(docsparity): resolve symlinked repository paths

### LINKED ISSUE #3598: test: TestDocumentationDiscoveryIgnoresUntrackedFiles always fails on macOS
## Problem

`TestDocumentationDiscoveryIgnoresUntrackedFiles` (`internal/docsparity/parity_helpers_test.go`), added in d7dd44b4, fails on every run on macOS:

```
parity_helpers_test.go:238: workflowFiles() = [.github/workflows/ignored.yaml .github/workflows/tracked.yml],
                            want [.github/workflows/tracked.yml]
```

Reproduced on a clean `origin/main` checkout, so it is not branch-specific. `make test` and `make docs-parity` both fail as a result.

## Cause

`filesMatching` decides whether to use `git ls-files` or fall back to a filesystem walk by comparing git's reported top level against the requested root:

```go
topLevel, err := exec.Command("git", "-C", root, "rev-parse", "--show-toplevel").Output()
if err != nil || filepath.Clean(strings.TrimSpace(string(topLevel))) != filepath.Clean(rootAbs) {
    return filesMatchingWalk(t, root, match)
}
```

On macOS `t.TempDir()` returns a path under `/var/folders/...`, and `/var` is a symlink to `/private/var`. `git rev-parse --show-toplevel` resolves symlinks; `filepath.Clean` does not. The two never compare equal:

```
raw temp path:     /var/folders/8c/.../T/tmp.mfZdGUosTq
git show-toplevel: /private/var/folders/8c/.../T/tmp.mfZdGUosTq
```

So the comparison always fails, `filesMatching` always falls back to `filesMatchingWalk`, and the walk returns untracked files — which is exactly what the test asserts must not happen. Linux CI has no such symlink, so the job passes there and the breakage is macOS-only.

## Suggested fix

Resolve symlinks on both sides before comparing, e.g. `filepath.EvalSymlinks(rootAbs)` (and on the git output) before `filepath.Clean`. That keeps the intended guard — only trust `git ls-files` when the root really is the repository top level — while letting it match on macOS.

Worth confirming the same comparison is not relied on elsewhere in the helper for the nested-repository case further down the test, which passes today only because it takes the walk path deliberately.

wolf31o2:
## Summary

- normalize Git repository roots before selecting index-backed discovery
- cover discovery through a directory symlink

Closes #3598

## Checks

- `GOWORK=off GOCACHE=/tmp/dingo-3598-gocache go test -race -count=1 -run "^TestDocumentationDiscoveryIgnoresUntrackedFiles$" ./internal/docsparity`
- `GOWORK=off GOCACHE=/tmp/dingo-3598-gocache go test -race -count=1 ./internal/docsparity`
- `GOWORK=off GOCACHE=/tmp/dingo-3598-darwin-gocache GOOS=darwin GOARCH=arm64 go test -c -o /tmp/dingo-3598-docsparity-darwin.test ./internal/docsparity`

---
## Summary by cubic
Fixes `docsparity` discovery so index-backed file lists are used even when the repository path is accessed through a symlink. Previously, the unresolved root path was compared to Git's resolved toplevel, causing fallback to the slower walk method. Now both paths are resolved before comparison. Adds a regression test for symlinked repo roots. Closes #3598.

<sup>Written for commit 196906187198af37ddfa07b0139986a2527791b5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3604?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/docsparity/parity_helpers_test.go
+++ b/internal/docsparity/parity_helpers_test.go
@@ -104,8 +104,16 @@ func filesMatching(
 	if err != nil {
 		return filesMatchingWalk(t, root, match)
 	}
+	rootResolved, err := filepath.EvalSymlinks(rootAbs)
+	if err != nil {
+		return filesMatchingWalk(t, root, match)
+	}
 	topLevel, err := exec.Command("git", "-C", root, "rev-parse", "--show-toplevel").Output()
-	if err != nil || filepath.Clean(strings.TrimSpace(string(topLevel))) != filepath.Clean(rootAbs) {
+	if err != nil {
+		return filesMatchingWalk(t, root, match)
+	}
+	topLevelResolved, err := filepath.EvalSymlinks(strings.TrimSpace(string(topLevel)))
+	if err != nil || filepath.Clean(topLevelResolved) != filepath.Clean(rootResolved) {
 		return filesMatchingWalk(t, root, match)
 	}
 	cmd := exec.Command("git", "-C", root, "ls-files", "-z")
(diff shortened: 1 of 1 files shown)


wolf31o2:
Superseded by PR #3610, which contains this symlink-root fix plus the broader cross-platform/generated-worktree discovery corrections. Retaining #3610 as the single review lane.

PR #3606, opened 2026-08-28
fix(leios): validate and bound endorser-block cache entries

### LINKED ISSUE #3512: leios: validate and bound endorser-block cache entries
The endorser-block cache is limited by entry count rather than bytes, and the
advertised body size is not checked against the stored payload. Retained data
can exceed the configured expectation while metadata and body contents disagree.

Acceptance criteria:

- Reject offers whose declared size differs from the received body.
- Enforce per-entry and aggregate byte limits.
- Define eviction behavior at the byte budget.
- Cover valid, mismatched, oversized, and eviction cases.

cryptodj413:
Fixes #3512

## Summary
- Rejects a `MsgBlockOffer`-driven manifest fetch whose received body length
  does not match the offer's declared `Size`.
- Enforces a per-entry byte budget (manifest + transaction bodies) on the
  in-memory endorser-block cache (`Ouroboros.leiosEndorserBlocks`),
  independent of the existing entry-count cap.
- Enforces an aggregate byte budget across the cache, evicting the
  oldest-inserted entries first when either budget is exceeded — the same
  policy already used for the entry-count cap.
- Extracts the manifest fetch/validate step into
  `fetchAndValidateLeiosEbManifest` behind a narrow `leiosBlockRequester`
  interface for unit testing without a live connection.

## Tests
- `TestFetchAndValidateLeiosEbManifestAcceptsMatchingSize` (valid)
- `TestFetchAndValidateLeiosEbManifestRejectsSizeMismatch` (mismatched)
- `TestStoreLeiosEndorserBlockRejectsOversizedEntry` (oversized)
- `TestLeiosEndorserBlockCacheEvictsOldestFirstAtByteBudget` (eviction)

## Validation
- `go build ./...`
- `go test ./ouroboros/... -race` (full package)
- `go test ./... -short` (full module)
- `golangci-lint run ./ouroboros/...`, `nilaway ./ouroboros/...`, `modernize ./ouroboros/...`
- `make import-boundaries`, `make docs-parity`, `make golines`
- `ARCHITECTURE.md` updated (cache bound description); `DATABASE.md` checked, not affected (no blob-key/schema change)
- Skipped: `make govulncheck` (network), DevNet/conformance (not required — no consensus/protocol wire-format change, only cache admission/eviction policy)

---
## Summary by cubic
Fixes #3512 by validating endorser-block manifest fetches against the offer's declared size and bounding the Leios endorser-block cache by bytes, so mismatched or oversized entries are rejected instead of cached and the cache can no longer balloon toward multiple GiB under the entry-count cap alone.

- Rejects a `MsgBlockOffer`-driven manifest fetch when the served body length does not match the offer's `Size`, and skips the fetch entirely when the offer's declared size already exceeds the per-entry budget.
- Rejects a store whose manifest plus held transaction bodies exceed a per-entry `16 MiB` budget, leaving any smaller existing entry for that hash untouched; the partial-transaction retention path enforces the same budget.
- Serves blob-store reloads that exceed the per-entry budget to the caller uncached, and re-checks the aggregate budget after an admitted reload.
- Evicts the oldest-inserted entries first once the cache exceeds a `256 MiB` aggregate budget, ordering by insertion sequence rather than wall-clock timestamp so a delayed goroutine cannot evict a newer entry first.
- Extracts fetch and size validation into `fetchAndValidateLeiosEbManifest` behind a `leiosBlockRequester` interface so it is unit-testable without a live connection.
- Adds tests for matching-size acceptance, size-mismatch rejection, oversized-entry rejection, oldest-first byte-budget eviction, seq-based eviction ordering, partial-retention budget rejection, uncached oversized reloads, and aggregate-budget eviction after a reload.

<sup>Written for commit 604748aeb8bac31b384995ae1dea9e22fb13d223. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3606?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added byte-size limits for cached Leios endorser blocks, including per-entry and aggregate cache budgets.
  * Cache entries are now evicted based on age when count or total byte limits are exceeded.

* **Bug Fixes**
  * Rejects blocks whose fetched size does not match the declared offer size.
  * Prevents oversized blocks from being fetched or cached.
  * Preserves existing cached entries when an oversized replacement is rejected.

* **Documentation**
  * Updated architecture documentation to describe the cache’s TTL, entry-count, and byte-size limits.


ouroboros/leios_merged.go
```diff
@@ -308,6 +332,18 @@ func (o *Ouroboros) storeLeiosEndorserBlock(
...
+	// manifest-only store) is left untouched, since the map has not been
+	// mutated yet at this point.
+	if n := data.approxBytes(); n > leiosEndorserBlockCacheMaxEntryBytes {
+		o.leiosMu.Unlock()
+		return fmt.Errorf(
+			"leios endorser block cache: entry size %d exceeds max %d",
+			n,
+			leiosEndorserBlockCacheMaxEntryBytes,
+		)
+	}
```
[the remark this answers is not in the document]
cryptodj413: Confirmed and fixed in a7566b56: retainLeiosPartialTxs now checks approxBytes() against leiosEndorserBlockCacheMaxEntryBytes before publishing the merged entry, rejecting the merge (keeping the existing smaller cached entry) when it would exceed the per-entry budget. Added TestRetainLeiosPartialTxsRejectsMergeOverEntryByteBudget, verified to fail without the fix.

ouroboros/leios_merged.go
```diff
@@ -548,11 +605,34 @@ func (o *Ouroboros) pruneLeiosEndorserBlockCacheLocked(now time.Time) {
 	slices.SortFunc(blocks, func(a, b *leiosEndorserBlockData) int {
 		return a.insertedAt.Compare(b.insertedAt)
 	})
```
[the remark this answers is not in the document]
cryptodj413: Confirmed and fixed in a7566b56: added a monotonic leiosEndorserBlockSeq counter assigned to each entry while leiosMu is held (Ouroboros.leiosEndorserBlockSeq / leiosEndorserBlockData.seq), and eviction now sorts by seq instead of insertedAt. Added TestLeiosEndorserBlockCacheEvictionOrdersBySeqNotInsertedAt, verified to fail against the old insertedAt-based sort.

cryptodj413:
@coderabbitai review

cryptodj413:
@coderabbitai review

cryptodj413:
@cubic-dev-ai review this PR

cryptodj413:
Status update:

- Rebased onto current \`main\` (was conflicting in ARCHITECTURE.md with #3601's edits to the same paragraph; merged both sets of prose). Force-pushed, conflict resolved (\`mergeable: MERGEABLE\`).
- All required CI checks pass on the current head (0c1a4e63): CodeQL, Analyze (actions/go), Conventional Commits, DCO, lint, govulncheck, go-test (Linux), go-test (Windows).
- \`go-test (Linux)\` failed once with a 20-minute hang in \`TestCaptureLoadGenesisSnapshot_RelayWarnsAndContinues\` (Badger commit-watermark wait) — unrelated to this PR's changes (doesn't touch \`internal/node\`, \`ledger/snapshot\`, or the Badger plugin). Re-ran the identical job with no code change and it passed clean; filed blinklabs-io/dingo#3609 to track the flake.
- Bot review: CodeRabbit fully reviewed the substantive diff (2 actionable findings, both fixed and confirmed with regression tests that fail without the fix — see thread replies) but is now rate-limited (1 review/hour, already used) so it hasn't re-run against the rebased head, whose diff content is unchanged by the rebase. Cubic's org-wide monthly review quota is exhausted (403,149/400,000 lines; resumes 2026-09-01) and cannot review at all right now.

Ready for human review.

ouroboros/leios_merged.go
```diff
@@ -682,6 +789,8 @@ func (o *Ouroboros) loadLeiosEBFromDB(
 	if existing := o.leiosEndorserBlocks[cacheKeys[0]]; existing == nil ||
 		existing.expired(time.Now()) {
 		o.pruneLeiosEndorserBlockCacheLocked(time.Now())
+		o.leiosEndorserBlockSeq++
```
wolf31o2, line 808: This database-reload insertion bypasses the new limits: pruning happens before `data` is added, with no `approxBytes` admission check or post-insert prune. `leiosfetchServerBlockRequest` reaches this path, so persisted or legacy endorser blocks can repopulate the cache beyond both budgets. Enforce the same limits before publishing (or serve the entry uncached), then prune after insertion, with a persisted-reload regression test.

wolf31o2 (changes requested):

ouroboros/leios_merged.go
cryptodj413: Confirmed and fixed in 12e49214: loadLeiosEBFromDB now checks approxBytes() against leiosEndorserBlockCacheMaxEntryBytes before admitting the reload into the map (serving it to the caller uncached when over budget), and re-prunes after insertion so the aggregate budget is re-checked too, matching storeLeiosEndorserBlock's before/after pattern. Added TestLoadLeiosEBFromDBServesOversizedEntryUncached, verified to fail without the fix.

cryptodj413:
@wolf31o2 fixed in 12e49214: \`loadLeiosEBFromDB\` now runs the same per-entry \`approxBytes()\` check as \`storeLeiosEndorserBlock\` before admitting a reloaded entry into the map (serving it to the caller uncached, not dropped, when over budget), and prunes again after insertion so the aggregate budget is re-checked too. Added \`TestLoadLeiosEBFromDBServesOversizedEntryUncached\`, confirmed to fail without the fix. Requested another review.

cryptodj413:
@coderabbitai review

ouroboros/leios_merged.go
```diff
@@ -682,9 +805,16 @@ func (o *Ouroboros) loadLeiosEBFromDB(
...
+		data.seq = o.leiosEndorserBlockSeq
 		for _, key := range cacheKeys {
 			o.leiosEndorserBlocks[key] = data
 		}
+		// Prune again after inserting so a reload that fits the per-entry
+		// budget but pushes the aggregate over its budget is evicted
+		// promptly, the same way storeLeiosEndorserBlock prunes both before
+		// and after admitting an entry.
+		o.pruneLeiosEndorserBlockCacheLocked(time.Now())
```
wolf31o2, line 817: **P2 — Cover aggregate budgeting on database reload**

Please add a regression that reloads two persisted entries which each fit the per-entry limit but together exceed the aggregate limit, then asserts the oldest entry is evicted and total retained bytes remain bounded. The existing reload test covers only the oversized-entry early return; deleting this post-insert prune otherwise escapes the submitted suite.

wolf31o2 (changes requested):
The original database-reload budget bypass is fixed, but the new aggregate reload guard still needs regression coverage that fails when the post-insert prune is removed.

ouroboros/leios_merged.go
cryptodj413: Added in 604748ae: TestLoadLeiosEBFromDBPrunesAggregateBudgetAfterReload reloads two persisted entries (each within the per-entry budget, together over the 1500-byte aggregate budget), asserts the older one is evicted and total retained bytes stay bounded. Verified it fails with the post-insert prune removed, and that the existing oversized-entry test alone does not catch that removal -- confirming the gap you flagged.

cryptodj413:
@wolf31o2 added in 604748ae: \`TestLoadLeiosEBFromDBPrunesAggregateBudgetAfterReload\` reloads two persisted entries that each fit the per-entry budget but together exceed the aggregate budget, and asserts the older is evicted with total retained bytes bounded. Confirmed it fails with the post-insert prune removed, and that the prior oversized-entry test alone does not catch that removal. Requested another review.

wolf31o2 (approved):

Issue #3523, opened 2026-08-25
leios: bound fetch bitmaps by transaction count


wolf31o2:
The Leios fetch response bitmap is expanded without checking its size against
the transaction count for the requested block. A small transaction set can
therefore produce a disproportionately large decoded bitmap.

Acceptance criteria:

- Reject bitmaps that exceed the expected length for the transaction count.
- Bound decoded bitmap allocation before expansion.
- Cover empty, exact-boundary, oversized, and malformed replies.

PR #3607, opened 2026-08-28
fix(leios): bind endorser-block entries to their announced point

### LINKED ISSUE #3513: leios: bind endorser blocks to their announced point
Endorser-block entries are accepted with a first-writer-wins slot association
without verifying that the supplied slot and body match the announced hash and
point. Different connections can populate conflicting metadata for one entry.

Acceptance criteria:

- Verify body hash, slot, and point against the announcement.
- Reject conflicting later offers.
- Keep valid retransmission idempotent.
- Cover cross-connection conflicts.

cryptodj413:
Fixes #3513

## Problem

`storeLeiosEndorserBlock` cached an endorser-block entry's slot on a
first-writer-wins basis: the first connection to store a body for a
given hash decided the cached slot, and only a *later* conflicting
offer was checked (against that first store, never against the
ranking-block announcement that actually vouches for the hash). A
single connection offering a real, previously-announced endorser block
at a fabricated slot could poison the entry before any conflicting
offer arrived to trip the existing-entry check.

## Fix

- Record the slot each ranking-block announcement declares for its
  endorser-block hash (`leiosAnnouncementSlots`), mirroring the
  existing per-hash size consistency check in `recordLeiosAnnouncement`.
- `storeLeiosEndorserBlock` now rejects a store whose point conflicts
  with the recorded announcement for that hash, when one is known. A
  hash with no recorded announcement (backfill of a block whose
  announcement aged out of the bounded ten-minute window, or a
  locally-forged block) is unaffected and still gets the existing
  point-hash and existing-entry checks.

## Tests

`ouroboros/leios_eb_announced_point_test.go` adds:
- rejection of a store whose point conflicts with the announcement,
  on the very first store attempt for the hash (no prior cache entry
  to otherwise catch it) -- confirmed to fail without the fix;
- acceptance of a store matching the announced point, and idempotent
  retransmission of it;
- a cross-connection conflict rejected regardless of arrival order.

## Validation

- `go build ./...`
- `go vet ./ouroboros/...`
- `golangci-lint run ./ouroboros/...` (0 issues)
- `nilaway ./ouroboros/...`
- `gofmt -l` / `golines -l` (clean)
- `go test ./ouroboros/... -race` (pass)

Full-repo `make test` (race, `dingo_extra_plugins`) was attempted but
was externally terminated partway through (before reaching the
`ouroboros` package) and was not retried; the above targeted `-race`
run of the changed package is the evidence for this change.
`DATABASE.md` is unaffected (no schema/blob-key changes); `ARCHITECTURE.md`'s
Leios Networking narrative is updated for the new invariant.

---
## Summary by cubic
Fixes #3513 by keying in-memory Leios endorser-block entries, fetches, and durable storage by `(slot, hash)` instead of hash alone. Previously, a peer could cache a real block under a fabricated slot before its announcement, and same-hash occurrences collided; now occurrences stay independent and unverified peer offers cannot affect voting, persistence, or ledger application.

**Bug Fixes**

- Track live announced slots per hash, allowing identical manifests to be announced at multiple slots while expiring old bindings.
- Authoritative announcements, local forging, and chain-derived backfills verify the matching occurrence; unverified entries remain withheld until then.
- Pass the expected slot through fetching, ledger application, closure checks, notification handling, and forging so only the requested occurrence satisfies a consumer.
- Key backfill, offer-fetch, and persistence job deduplication by `(slot, hash)` so one occurrence cannot suppress another.
- Keep announcement/store reconciliation atomic and promote entries with copy-on-write; publication runs only after releasing the announcements lock.
- Persistent `em`/`et` blobs are keyed by `(slot, hash)` too, so each occurrence is durably re-served independently.
- Reads of `em`/`et` blobs fall back to legacy hash-only keys only on a genuine miss; other read errors propagate unchanged, so pre-change data stays servable.

<sup>Written for commit 62eb1d50e411ecc0c12f68cd521d6d418b579ccf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3607?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Leios endorser-block validation by ensuring stored blocks match the slot announced for their hash.
  - Rejects conflicting offers and prevents invalid data from poisoning the cache.
  - Preserves valid entries across retransmissions and conflicting connections.

- **Documentation**
  - Updated networking architecture documentation to describe announcement-based slot validation and handling of unannounced hashes.

- **Tests**
  - Added coverage for conflicting slots, valid retransmissions, and cross-connection ordering scenarios.


cryptodj413:
@coderabbitai review

cryptodj413:
@coderabbitai full review

cryptodj413:
Bot review status: CodeRabbit hit its OSS rate limit on the initial automatic pass; a manual `@coderabbitai review` retry (after the stated 20-minute window) and a `@coderabbitai full review` retry both also came back rate limited, the latter citing a 55-minute wait. Cubic is separately unavailable — it has reached its monthly line-review limit, resuming 1 September 2026 (~4 days out), unrelated to this PR.

Both configured review bots are therefore unavailable for this PR despite good-faith retries. Requesting human review per the workspace's bot-first sequence; will re-check CodeRabbit if a review is requested before it becomes available again.

cryptodj413:
@coderabbitai review

cryptodj413:
@coderabbitai full review

cryptodj413:
CodeRabbit follow-up: after the rate-limit window reset (~05:00 UTC), `@coderabbitai full review` was triggered and acknowledged ("Full review finished", 05:17:02Z). No review body, summary, or inline comments were posted afterward (0 PR reviews, 0 review comments via the GitHub API, checked 15+ minutes later) — treating this as a completed pass with nothing flagged, not another rate limit. Cubic remains unavailable for ~4 days (monthly line-review limit). Standing down on further bot-review retries; PR remains with human review requested from Core/Docs/PMs.

ouroboros/leios_merged.go
```diff
@@ -242,6 +242,26 @@ func (o *Ouroboros) storeLeiosEndorserBlock(
...
+	// checked against that first store, never against the announcement), so
+	// a single connection offering a real, previously-announced endorser
+	// block at a fabricated slot could poison the entry before any
+	// conflicting offer arrived to trip the existing-entry check below
+	// (issue #3513). A hash with no recorded announcement (backfill of an
+	// old block whose announcement already expired the bounded window, or a
+	// locally-forged block not sourced from a peer announcement at all)
+	// skips this check; it still gets the existing-entry and body-hash
+	// checks above and below.
+	if announcedSlot, ok := o.leiosAnnouncedSlot(point.Hash); ok &&
```
wolf31o2: This lookup protects only announcement-first ordering. `BroadcastEndorserBlock` queues the block offer before the ranking-block announcement, and the receiver stores fetched offers asynchronously, so an authentic manifest can be stored under a fabricated slot while this map is empty. A deterministic probe retained slot 42 after accepting the valid announcement for slot 41. Make publication contingent on a validated matching announcement, or reconcile/quarantine store-first entries before persistence, pipeline publication, and ledger consumption; then add the reverse-order regression.

wolf31o2 (changes requested):

ouroboros/leios_merged.go
cryptodj413: Confirmed and fixed in 86ec183f. Your probe reproduced exactly: with the offer stored first, the fabricated slot 42 was retained after the valid slot-41 announcement, and the vote had already fired with 42. Since `BroadcastEndorserBlock` (and the relay) queue the offer before the announcement, the original lookup was inert on the normal ordering.

Each store now carries a `leiosStoreOrigin`. Locally forged blocks and by-point backfills (whose point comes from the ranking block being applied) are authoritative; leios-notify offers are not. A peer-offered store contradicting a recorded announcement is rejected outright; an unannounced one is cached but marked `slotVerified: false` rather than rejected, since rejecting would drop endorser blocks on the normal ordering.

Everything keyed on the slot is withheld while unverified: vote emission, pipeline observation, blob persistence, and the slot `EndorserBlockTxsByHash` hands the ledger for the endorser blob key. `bindLeiosEndorserBlockSlot` reconciles when an authority arrives — a validated announcement, or the chain-derived point in `FetchEndorserBlockByPoint` — promoting and publishing a matching entry and evicting a contradicting one.

Reverse-order regressions added, each confirmed failing without the fix: publication withheld until an announcement binds a peer offer; a fabricated slot evicted by the genuine announcement without ever reaching the vote handler; and a complete-but-unverified entry withheld from the ledger provider.

cryptodj413:
Pushed 86ec183f addressing the review.

The announcement lookup alone only protected announcement-first ordering; the normal ordering is offer-first, so an authentic manifest could be bound to a fabricated slot and published before anything contradicted it. Publication keyed on the slot (vote emission, pipeline observation, blob persistence, and the slot handed to the ledger for the endorser blob key) is now withheld until the binding is verified by a validated announcement or the ledger's chain-derived point, with reconciliation promoting a matching entry and evicting a contradicting one.

Validation: `go test ./ouroboros/... -race` and `go test ./ledger/...` pass; build, `golangci-lint` (0 issues), `nilaway`, `gofmt`/`golines`, `make import-boundaries`, `make docs-parity` clean. The three new reverse-order regressions were each confirmed failing without the fix. `ARCHITECTURE.md` updated; `DATABASE.md` unaffected.

ouroboros/leios_merged.go
```diff
@@ -242,18 +266,44 @@ func (o *Ouroboros) storeLeiosEndorserBlock(
...
+	//
+	// The announcement need not have arrived yet -- the relay, and dingo's own
+	// forge path, queue the block offer before the ranking-block announcement
+	// -- so an unannounced peer-offered store is cached but left unverified
+	// rather than rejected, which would drop endorser blocks on the normal
+	// ordering. slotVerified gates everything keyed on the slot; the binding is
+	// reconciled later by bindLeiosEndorserBlockSlot when the announcement (or
+	// the ledger's chain-derived reference) arrives.
+	verified := origin == leiosStoreAuthoritative
+	if announcedSlot, ok := o.leiosAnnouncedSlot(point.Hash); ok {
```
wolf31o2: **P1 — Make announcement/store reconciliation atomic**

This samples `leiosAnnouncementSlots` before decoding and before the cache insertion, while `recordLeiosAnnouncement` performs only one bind after dropping the announcement lock. If a matching announcement records and its bind runs during that gap, it sees no cache entry; the later store then inserts `slotVerified=false` after the authority has already passed, so the block remains permanently withheld. A coordinated store/announcement regression reproduced this. Recheck or bind after insertion under one consistent lock order, and cover this interleaving.

ouroboros/leios_backfill.go
```diff
@@ -111,8 +111,15 @@ func (o *Ouroboros) FetchEndorserBlockByPoint(
 	ebSlot uint64,
 	ebHash []byte,
 ) error {
+	// The caller's point comes from the ranking block being applied, so it is
+	// authoritative for this endorser block's slot. Reconcile any entry cached
+	// from a peer offer before an announcement corroborated it: a matching
+	// entry is promoted (and published) here, a contradicting one is evicted so
+	// the fetch below replaces it rather than serving a poisoned slot to the
+	// ledger (issue #3513).
+	o.bindLeiosEndorserBlockSlot(ebHash, ebSlot)
```
wolf31o2: **P1 — Rebind entries loaded from persistence**

This bind runs before a cache miss triggers the database load. `loadLeiosEBFromDB` reconstructs the entry with zero-value `slotVerified`; `fetchEndorserBlockOnConn` then skips its authoritative store when the reload is complete, and the final success check tests only completeness. The fetch therefore returns nil while `EndorserBlockTxsByHash` still reports the closure unavailable. A store → drain → clear-memory → provider regression fails. Load before binding and validate the persisted slot against `ebSlot`, or otherwise restore the binding safely, and require `slotVerified` in the post-fetch success check.

ouroboros/leios_merged.go
```diff
@@ -338,7 +412,48 @@ func (o *Ouroboros) storeLeiosEndorserBlock(
...
+			"evicted leios EB cached under a slot its announcement contradicts",
+			"component", "network",
+			"protocol", "leios-notify",
+			"cached_slot", data.point.Slot,
+			"announced_slot", slot,
+			"hash", hex.EncodeToString(ebHash),
+		)
+		return
+	}
+	data.slotVerified = true
```
wolf31o2: **P1 — Keep cached entries immutable after publication**

`lookupLeiosEndorserBlock` returns this shared pointer after releasing `leiosMu`, and readers inspect `slotVerified` outside the lock. Mutating the field here violates this file’s existing copy-on-update invariant; a focused `-race` test reports the write at this line racing with a reader of the returned snapshot. Replace the cached entry with a verified copy, as the other cache-update paths already do, then publish that copy.

ouroboros/leiosnotify.go
```diff
@@ -1569,6 +1625,18 @@ func (o *Ouroboros) recordLeiosAnnouncement(
...
 
+// leiosAnnouncedSlot returns the slot a previously observed ranking-block
+// announcement recorded for the endorser block identified by ebHash, and
+// whether one is known. It lets a leios-fetch offer or store be bound to the
+// point its announcement actually vouched for, rather than trusting whatever
+// point the offering connection supplies for that hash (issue #3513).
+func (o *Ouroboros) leiosAnnouncedSlot(ebHash []byte) (uint64, bool) {
+	o.leiosAnnouncementsMu.Lock()
+	defer o.leiosAnnouncementsMu.Unlock()
+	slot, ok := o.leiosAnnouncementSlots[string(ebHash)]
```
wolf31o2: **P2 — Expire slot bindings before validating a new offer**

This lookup returns the slot index without aging it, while pruning runs only when another announcement is accepted. After ten idle minutes, the normal offer-before-announcement ordering can therefore reject an offer for the same content hash at a new slot against an expired binding. A review regression that advances the announcement ledger past the expiry window fails with the stale slot mismatch. Prune or check expiry during this lookup and cover the idle-expiry path.

ouroboros/leios_eb_zero_refs_test.go
```diff
@@ -167,7 +167,7 @@ func TestStoreLeiosEndorserBlockEmptyManifestIsHashMismatch(t *testing.T) {
 	emptyManifest := []byte{0xa0}
 
 	o := newOuroboros(OuroborosConfig{EnableLeios: true})
-	err := o.storeLeiosEndorserBlock(point, emptyManifest, nil)
+	err := o.storeLeiosEndorserBlock(point, emptyManifest, nil, leiosStoreAuthoritative)
```
wolf31o2: **P3 — Apply the repository’s golines formatting**

The required `golines -l --ignore-generated --chain-split-dots --max-len=80 --reformat-tags` check lists this file and four other PR-touched test files after the new origin argument made these calls overlong. Run `make golines` so the changed tests satisfy the repository’s formatting contract.

wolf31o2 (changes requested):
The sequential reverse-order case is improved, but concurrent and persisted-cache binding paths still have blocking defects.

ouroboros/leios_merged.go
cryptodj413: Confirmed and fixed in 40e22115. storeLeiosEndorserBlock now holds `leiosAnnouncementsMu` from the announcement check through the cache insertion, and `recordLeiosAnnouncement` holds the same lock across recording the announcement and calling `bindLeiosEndorserBlockSlot` -- both in the same order (announcements before the endorser-block cache), so the two check-then-act sequences can no longer interleave. Added `TestStoreAndAnnouncementRaceAlwaysEndsVerified`, which races a peer-offered store against its matching announcement across 64 independent hashes concurrently under `-race`; confirmed it fails on every one of 15 runs against the prior code (unverified entries left behind) and passes reliably against the fix.

ouroboros/leios_backfill.go
cryptodj413: Confirmed and fixed in 40e22115, two parts. loadLeiosEBFromDB now reconstructs the reload as already `slotVerified`, since the blob store is only ever written from publishLeiosEndorserBlock once an entry is verified -- a persisted (slot, hash) pair is already authoritatively bound. Separately, fetchEndorserBlockOnConn's initial lookup previously took the completeness shortcut for an already-cached-but-unverified entry without ever binding it; it now calls bindLeiosEndorserBlockSlot in place when the entry exists but isn't verified (no network fetch needed, the bytes are already held), and FetchEndorserBlockByPoint's per-attempt success check now requires slotVerified in addition to completeness. Added TestEndorserBlockTxsByHashAvailableAfterDBReload (store -> drain -> clear-memory -> provider), confirmed failing without the loadLeiosEBFromDB fix.

ouroboros/leios_merged.go
cryptodj413: Confirmed and fixed in 40e22115. bindLeiosEndorserBlockSlot now replaces the cached entry with a verified copy (`verified := *data; verified.slotVerified = true`) and swaps the map pointer under the lock, the same pattern retainLeiosPartialTxs already uses, instead of flipping the field on the shared struct. Added TestBindLeiosEndorserBlockSlotDoesNotMutateSharedEntry, which races a reader of an already-held pointer against a concurrent bind under -race; confirmed it reproduces the data race (and the semantic mutation) against the prior code and is clean against the fix.

ouroboros/leiosnotify.go
cryptodj413: Confirmed and fixed in 40e22115. The (now-internal) leiosAnnouncedSlotLocked treats a binding as expired, and so absent, once its slot ages past the same leiosNotifyMaxAnnouncementAge window pruneLeiosAnnouncements enforces elsewhere -- consistent with pruneLeiosAnnouncements' own no-op when the ledger isn't wired. Added TestLeiosAnnouncedSlotIgnoresExpiredBinding (fake ledger reporting the binding's slot as long past the window), confirmed failing without the fix.

ouroboros/leios_eb_zero_refs_test.go
cryptodj413: Fixed in 40e22115 -- ran `make golines` (also picked up leios_eb_cache_bounds_test.go, which needed the new store-origin argument added regardless). `gofmt -l`/`golines -l` are clean now.

cryptodj413:
Pushed 40e22115 addressing this review round, plus a merge with origin/main resolving the conflict with #3606 (merged since this PR opened, same files).

All 5 findings confirmed and fixed:
- **P1** (atomic reconciliation): `storeLeiosEndorserBlock` and `recordLeiosAnnouncement` now hold `leiosAnnouncementsMu` across their full check-then-act sequence, in a consistent lock order, closing the interleaving where a one-shot reconciliation could run against a cache not yet holding the entry.
- **P1** (rebind from persistence): `loadLeiosEBFromDB` reconstructs a reload as already verified (the blob store is never written from an unverified entry); `fetchEndorserBlockOnConn`/`FetchEndorserBlockByPoint` now bind and require `slotVerified` rather than completeness alone.
- **P1** (immutable entries): `bindLeiosEndorserBlockSlot` publishes a verified copy instead of mutating the shared cached pointer, matching `retainLeiosPartialTxs`'s existing pattern.
- **P2** (expire slot bindings): the announcement-slot lookup now treats a binding as expired (and so absent, not conflicting) once it ages past the same window `pruneLeiosAnnouncements` enforces.
- **P3** (golines): `make golines` run; clean.

Four new regressions added, each confirmed failing without its corresponding fix and passing with it (details in the inline replies).

Validation: `go build ./...`, `go vet`, `golangci-lint` (0 issues), `nilaway` (clean), `gofmt`/`golines` (clean), `go test ./ouroboros/... -race` and `go test ./ledger/...` pass, `make import-boundaries`/`make docs-parity` pass. `ARCHITECTURE.md` updated for the additional invariants; `DATABASE.md` unaffected.

ouroboros/leios_backfill.go
```diff
@@ -111,8 +111,15 @@ func (o *Ouroboros) FetchEndorserBlockByPoint(
...
 	if data, ok := o.lookupLeiosEndorserBlock(ebHash); ok &&
```
wolf31o2: **P1 — Validate a persisted slot against the authoritative point.** `loadLeiosEBFromDB` marks every persisted entry `slotVerified`, while `bindLeiosEndorserBlockSlot` returns immediately for any verified entry. This success path checks completeness and the flag but never requires `data.point.Slot == ebSlot`: a focused regression persisted a valid block at slot 41, requested the same hash authoritatively at slot 42, and this returned `nil` while the provider still exposed slot 41. This affects databases written by the previous peer-supplied-slot behavior and the supported same-hash rebind-after-expiry path. Compare, invalidate, or rebind the persisted entry against `ebSlot` before accepting it, and cover the mismatched-slot reload.

wolf31o2 (changes requested):
One new blocking correctness issue is inline. The prior announcement/store interleaving, same-slot reload, shared-pointer race, and stale-binding issues are fixed; the existing `golines` thread remains unresolved.

ouroboros/leios_backfill.go
cryptodj413: Confirmed and fixed in 30d417a1. `loadLeiosEBFromDB` marking every reload verified unconditionally was correct only for whatever occurrence wrote it; nothing compared that reloaded slot against `ebSlot`, so a hash persisted at 41 and requested authoritatively at 42 reported success while the provider still exposed 41.

Fixed in three parts:
- `bindLeiosEndorserBlockSlot` now looks up through `lookupLeiosEndorserBlock` (so a blob-only reload is reconciled, not just an in-memory entry) and re-checks the slot even when already verified, evicting on a mismatch.
- Since eviction alone can't correct a stale blob record (the next reload just repopulates it), `FetchEndorserBlockByPoint`'s two success checks and `fetchEndorserBlockOnConn`'s initial lookup now additionally require the point to match the caller's authoritative slot before treating anything as already satisfied.
- `storeLeiosEndorserBlock` now lets an authoritative store (never a peer-offered one) override a cached entry under a contradicting slot instead of rejecting it, so a genuine re-fetch actually corrects the record (and the blob store's single record for that hash) going forward.

`TestFetchEndorserBlockByPointRejectsStaleReloadedSlot` reproduces your scenario (persist at 41, drain, clear memory, request authoritatively at 42) and is confirmed failing without either half of the fix. `TestStoreLeiosEndorserBlockRejectsSlotMismatchBeforeVote` is updated to exercise the peer-offered side of the reject/override distinction, since the authoritative side is now intentionally allowed to override.

cryptodj413:
Pushed 30d417a1 addressing this round's finding.

Confirmed: `loadLeiosEBFromDB` marking every reload verified unconditionally was correct only for whatever occurrence wrote it, and nothing compared that against the caller's authoritative `ebSlot`. Fixed with three coordinated changes (details in the inline reply): `bindLeiosEndorserBlockSlot` reconciles through the DB-inclusive lookup and re-checks the slot even when already verified; `FetchEndorserBlockByPoint`/`fetchEndorserBlockOnConn` require the point to actually match before treating a lookup as satisfied (eviction alone can't fix a stale blob record, since the next reload just repopulates it); `storeLeiosEndorserBlock` lets an authoritative store override a contradicting cached entry so a genuine re-fetch corrects the record.

New regression (`TestFetchEndorserBlockByPointRejectsStaleReloadedSlot`) reproduces the reported scenario and is confirmed failing without either half of the fix. An existing test (`TestStoreLeiosEndorserBlockRejectsSlotMismatchBeforeVote`) is updated to test the peer-offered side specifically, since an authoritative override is now intentionally allowed.

Validation: `go build ./...`, `go vet`, `golangci-lint` (0 issues), `nilaway` (clean), `gofmt`/`golines` (clean), `go test ./ouroboros/... -race` and `go test ./ledger/...` pass, `make import-boundaries`/`make docs-parity` pass. `ARCHITECTURE.md` updated.

ouroboros/leios_eb_announced_point_test.go
```diff
@@ -0,0 +1,623 @@
...
+	// o.connManager is nil, so a genuine cache miss (or a correctly-rejected
+	// stale reload) must return an error here, not silently succeed --
+	// there is no way to actually fetch anything without a connection.
+	err = o.FetchEndorserBlockByPoint(authoritativeSlot, hash)
+	require.Error(
+		t,
+		err,
+		"must not silently accept a reload bound to a different slot",
+	)
+	// Without a connection to actually re-fetch and re-persist, the blob
```
wolf31o2, line 722: **P1 — Do not leave the stale slot visible to the ledger provider**

This narrow contract is unsafe because each ledger availability gate calls `EndorserBlockProvider(hash)` before `FetchEndorserBlockByPoint` and discards the returned slot (`ensureReferencedEndorserBlocks`, `classifyEndorserBlockFetches`, and `leiosBackfiller.spawn`). A complete persisted entry at slot 99 therefore satisfies a certified reference at slot 100, so the corrected fetch is never invoked; `ledgerProcessBlock` later applies the closure under the stale slot. A focused test through `ensureReferencedEndorserBlocks` returned `nil` for this mismatch. Make availability and final application slot-aware so a mismatched provider result is treated as missing and triggers the authoritative fetch.

wolf31o2 (changes requested):
The prior storage and rebinding fixes are present, but one ledger-facing P1 remains. The existing `golines` P3 also remains open.

ouroboros/leios_eb_announced_point_test.go
cryptodj413: Confirmed and fixed in 8eec3e7b. Every ensureReferencedEndorserBlocks/classifyEndorserBlockFetches/leiosBackfiller call site already knows the slot its own reference requires (leiosEbRef pairs them) but discarded the provider's returned slot and trusted ok=true alone.

Added `endorserBlockAvailableAt`, requiring the provider's returned slot to match the caller's expected one, and used it at every call site: `ensureRequiredAvailable`, `classifyEndorserBlockFetches`'s `cached` check, the backfill/tipWait loops, `leiosBackfiller.spawn`/`awaitFetch`, and `waitForEndorserBlock`. `leiosEndorserBlockForApply` now also returns the expected slot (the block's own slot on the CIP path, the certifying block's resolved parent's slot on the Musashi path), and `ledgerProcessBlock` uses that -- not the provider's own reported slot -- both to gate availability and as `applyEndorserBlock`'s `ebSlot`.

`TestEnsureReferencedEndorserBlocksRejectsProviderResultAtWrongSlot` reproduces your scenario through `ensureReferencedEndorserBlocks` directly and is confirmed failing without the fix.

While re-checking this round end to end I also found and closed a symmetric gap in the prior fix: `storeLeiosEndorserBlock`'s announcement check still unconditionally rejected an authoritative store contradicting a live announcement record, even though the equivalent cache-entry check was already relaxed for authoritative origin. Fixed the same way, with `TestStoreLeiosEndorserBlockAuthoritativeOverridesStaleAnnouncement` confirmed failing without it.

cryptodj413:
Pushed 8eec3e7b addressing this review round, plus a broader re-check across the whole PR (per request).

**The reported finding is fixed:** every `EndorserBlockProvider` call site in `ensureReferencedEndorserBlocks`/`classifyEndorserBlockFetches`/`leiosBackfiller` now validates the provider's returned slot against the slot its own reference requires (shared `endorserBlockAvailableAt` helper), instead of trusting `ok=true` alone. `leiosEndorserBlockForApply` now also returns the structurally-derived expected slot, and `ledgerProcessBlock` uses that (not the provider's own reported value) both to gate availability and for `applyEndorserBlock`'s `ebSlot`. `TestEnsureReferencedEndorserBlocksRejectsProviderResultAtWrongSlot` reproduces the exact scenario through `ensureReferencedEndorserBlocks`, confirmed failing without the fix.

**Also fixed, found during the broader re-check:** `storeLeiosEndorserBlock`'s announcement check still unconditionally rejected an authoritative store contradicting a live announcement record, a symmetric gap to the cache-entry check already relaxed for authoritative origin in an earlier round. Same fix applied; `TestStoreLeiosEndorserBlockAuthoritativeOverridesStaleAnnouncement` confirmed failing without it.

**Re-check performed:** re-read the full locking design (lock order `leiosAnnouncementsMu` → `leiosMu` everywhere, no reverse acquisition found, no reentrancy through persist/vote/pipeline callbacks) for deadlock risk; re-traced `storeLeiosEndorserBlock`'s branch logic and `fetchEndorserBlockOnConn`'s restructured flow end-to-end against the original attack scenario; re-verified every `EndorserBlockProvider` call site in the ledger package (6 production sites + test fakes) for the same class of bug.

Resolved the golines (P3) thread, already fixed in 40e22115 and clean in every check since.

Validation: `go build ./...`, `go vet ./...`, `golangci-lint` (0 issues), `nilaway` (clean; one pre-existing, unrelated finding in `ledger/leios/manager_test.go` not touched by this PR), `gofmt`/`golines` (clean), `go test ./ouroboros/... -race` (77s) and targeted `go test ./ledger/... -race` for every Leios-related test (12s) pass; full `go test ./ledger/...` (non-race) passes in 63s. A full-package `-race` run of `./ledger/...` hit this shared machine's memory pressure (swap saturated by unrelated processes) and timed out at the default 10m — confirmed environmental, not a hang: the non-race run and every directly-relevant `-race`-scoped subset both pass cleanly. `make import-boundaries`/`make docs-parity` pass. `ARCHITECTURE.md` updated for the new invariant.

ledger/leios_apply.go
```diff
@@ -853,7 +902,7 @@ func (b *leiosBackfiller) spawn(r leiosEbRef) {
 	if _, loaded := b.inflight.LoadOrStore(key, struct{}{}); loaded {
```
wolf31o2, line 931: P1: In-flight backfill deduplication is keyed only by hash, but this change makes cache availability slot-bound and allows the same manifest hash at different slots. Concurrent requests for slots A and B can suppress B's fetch; when A completes, B's await path has no provider entry and can leave the B-bound closure unavailable. Include the slot in the in-flight key and add a same-hash/different-slot concurrent test.

wolf31o2 (changes requested):
One ledger-facing concurrency issue remains on this head.

ledger/leios_apply.go
cryptodj413: Confirmed and fixed in aa31e373. Deduping in-flight fetches by hash alone broke once cache/provider availability became slot-bound: the same hash can now legitimately be required at two different slots concurrently, but a hash-only key let one slot's still-in-flight fetch silently suppress `spawn` for the other (found the key already loaded, no-opped without ever calling fetch), and let `awaitFetch`'s "not in flight" skip-fast fire the moment the *first* slot's fetch cleared the shared key -- reporting the second slot's requirement as resolved when it was never fetched at all.

Keyed `leiosBackfiller`'s in-flight map by (slot, hash) instead of hash alone, in both `spawn` and `awaitFetch`, so two different slots of the same hash are dispatched and awaited independently.

Added two regressions, each confirmed failing without the fix:
- `TestLeiosBackfillerSpawnDedupsByHashAndSlotIndependently" -- spawns slot 100 (blocks in flight), then spawns slot 200 for the same hash, and asserts `fetch` is actually invoked for both.
- `TestLeiosBackfillerAwaitFetchDoesNotSkipFastOnDifferentSlotCompletion` -- spawns both slots concurrently, lets slot 100 finish first while slot 200 is still genuinely in flight, and asserts `awaitFetch` for slot 200 does not return until slot 200 itself completes.

cryptodj413:
Pushed aa31e373 addressing this round.

`leiosBackfiller` deduped in-flight fetches by hash alone, which broke once the prior fix made availability slot-bound: the same hash can now legitimately be required at two different slots concurrently. A hash-only key let one slot's still-in-flight fetch silently suppress `spawn` for a different slot of the same hash, and let `awaitFetch`'s skip-fast wrongly treat the first slot's completion as satisfying the second's requirement.

Fixed by keying the in-flight map by (slot, hash) in both `spawn` and `awaitFetch`. Two new regressions target `spawn`'s dispatch suppression and `awaitFetch`'s skip-fast separately, each confirmed failing without the fix (details in the inline reply).

Validation: `go build ./...`, `go vet ./...`, `golangci-lint` (0 issues), `nilaway` (clean; same pre-existing unrelated finding in `ledger/leios/manager_test.go`), `gofmt`/`golines` (clean), targeted `go test ./ledger/... -race` for every Leios-related test (12s) and full `go test ./ledger/...` non-race (63s) pass. `make import-boundaries`/`make docs-parity` pass. `ARCHITECTURE.md` updated.

ledger/leios_apply.go
```diff
@@ -698,13 +739,13 @@ func classifyEndorserBlockFetches(
...
 	certDrivenHistorical bool,
-	cached func(ebHash lcommon.Blake2b256) bool,
+	cached func(r leiosEbRef) bool,
 ) (backfill, tipWait []leiosEbRef) {
 	backfillSeen := make(map[string]struct{})
 	tipWaitSeen := make(map[string]struct{})
 	appendRef := func(dst *[]leiosEbRef, seen map[string]struct{}, r leiosEbRef) {
 		key := string(r.hash.Bytes())
-		if _, ok := seen[key]; ok || cached(r.hash) {
+		if _, ok := seen[key]; ok || cached(r) {
```
wolf31o2, line 769: P1: `classifyEndorserBlockFetches` still deduplicates by hash alone even though provider/cache availability is slot-bound. A second reference with the same hash at a different slot can be omitted from backfill/tip-wait, so its slot-specific fetch never occurs. Key these maps by `(slot, hash)`; apply the same fix to `requiredCertifiedEndorserBlocks` at line 667.

wolf31o2 (changes requested):
The worker key is now slot-aware, but the upstream reference collectors still collapse distinct slot-bound references. `requiredCertifiedEndorserBlocks` at `ledger/leios_apply.go:667` also keys `seen` by hash alone; key both collectors by `(slot, hash)` and add a same-hash/different-slot regression test.

ledger/leios_apply.go
cryptodj413: Confirmed and fixed the remaining half in b6cfbd6a (merged with your 187084f2 in 3e803432, which also picks up ~59 commits from main since this branch's last sync).

`classifyEndorserBlockFetches`'s `appendRef` was still deduping `backfillSeen`/`tipWaitSeen` by hash alone. Introduced a single `leiosEbRefKey(r)` (slot, hash) helper used everywhere in this file now -- `requiredCertifiedEndorserBlocks` (replacing the inline formatting from 187084f2 with a call to the shared helper), `classifyEndorserBlockFetches`'s `appendRef`, and `leiosBackfiller`'s in-flight map (dropping the now-redundant `leiosBackfillInflightKey` wrapper).

`TestClassifyEndorserBlockFetchesKeepsDistinctSlotsOfSameHash` reproduces the same-hash-different-slot scenario through `classifyEndorserBlockFetches` directly and is confirmed failing without the fix, alongside your `TestRequiredCertifiedEndorserBlocksKeepsDistinctSlots`.

Validation: `go build ./...`, `go vet ./...` (whole repo, clean), `golangci-lint`/`nilaway` on `ouroboros`+`ledger` (0 issues from this change; a few pre-existing findings in files this PR doesn't touch, from the main merge), `gofmt`/`golines` clean, targeted `-race` tests for every Leios-related test in both packages pass, full `ouroboros` `-race` (25s) and full `ledger` non-race (42s) pass. `make import-boundaries`/`docs-parity` pass.

cryptodj413:
Pushed 3e803432 (b6cfbd6a merged with your 187084f2, plus ~59 commits from main since this branch's last sync -- resolved one conflict in ouroboros/leios_eb_cache_bounds_test.go, an unrelated pre-existing call site missing the new origin argument).

Your commit fixed `requiredCertifiedEndorserBlocks`; `classifyEndorserBlockFetches`'s `appendRef` -- the other half your review comment named -- was still hash-only. Fixed with a single shared `leiosEbRefKey(r)` (slot, hash) helper now used by both collectors and `leiosBackfiller`'s in-flight map, replacing three separate ad hoc key implementations. `TestClassifyEndorserBlockFetchesKeepsDistinctSlotsOfSameHash` confirmed failing without the fix.

Validation: full-repo `go build`/`go vet` clean, `golangci-lint`/`nilaway` clean on the changed packages, `gofmt`/`golines` clean, targeted `-race` Leios tests and full `ouroboros` `-race` (25s) / full `ledger` non-race (42s) pass, `make import-boundaries`/`docs-parity` pass.

ouroboros/leiosnotify.go
```diff
@@ -637,6 +644,7 @@ func (o *Ouroboros) leiosnotifyClientNotification(
 				point,
 				blockRaw,
 				nil,
+				leiosStorePeerOffered,
```
wolf31o2, line 651: P1: This hash-only cache hit drops a valid offer when the same manifest hash recurs at a different slot, which this PR explicitly supports. If slot A is cached and an offer for slot B arrives before B's announcement, this returns without fetching B; the later announcement binds and evicts A, leaving B unavailable. The MsgBlockTxsOffer path has the same hash-only suppression via `leiosFetchInProgress` at lines 703-708. Make the offer/in-progress decision slot-aware and add an offer-first same-hash/different-slot regression through the notify handler.

ouroboros/leios_merged.go
```diff
@@ -268,8 +268,36 @@ type leiosEndorserBlockData struct {
...
+	// pipeline observation, blob persistence, and the slot handed to the
+	// ledger by EndorserBlockTxsByHash -- is withheld until it is true, so a
+	// peer cannot bind an authentic manifest to a fabricated slot by offering
+	// it before its announcement arrives (issue #3513). This field is never
+	// mutated on a cache entry that is already reachable by another reader:
+	// the cache stores *leiosEndorserBlockData, and lookupLeiosEndorserBlock
+	// hands that pointer to callers outside leiosMu, so promoting an entry
+	// (bindLeiosEndorserBlockSlot) replaces it with a new, distinct value
+	// under the lock rather than flipping this field on the shared one.
+	slotVerified bool
```
wolf31o2, line 283: P1: A complete peer-offered entry can still be `slotVerified == false`, but this accessor returns its transaction hashes based only on completeness. The forging adapter uses this result to exclude the certified EB's transactions from a newly announced EB, so an unbound peer claim can affect forging before announcement/point binding. Require `slotVerified` here, and apply the same gate to `resolveCertifiedEndorserTxs`/`leiosClosureCompleteLocked`; otherwise node-to-client merge and closure waiters can also consume an unverified slot. Add a regression proving a complete unverified entry is hidden from each consumer.

wolf31o2 (changes requested):
Two current-head blocking correctness issues remain in the slot-binding boundary.

ouroboros/leiosnotify.go
cryptodj413: Fixed in 1463882b.

- MsgBlockOffer's cache-hit skip now requires `data.point.Slot == point.Slot`, not just a hash hit.
- MsgBlockTxsOffer's top-level skip gets the same slot check.
- The dispatched fetch closure independently re-checked the same cache entry via a second hash-only "nothing left to fetch" test (line ~723 pre-fix) — fixed too, since the top-level fix alone doesn't stop that inner re-check from swallowing the offer.
- `leiosFetchInProgress` is now keyed by `leiosFetchClaimKey(point)` (slot + hash), not hash alone.

Regression: `TestLeiosNotifyBlockOfferFetchesSameHashAtDifferentSlot` and `TestLeiosNotifyBlockTxsOfferFetchesSameHashAtDifferentSlot` in `ouroboros/leiosnotify_slot_test.go` drive the notify handler through a full mock leios-fetch conversation with a stale-slot cache entry present; both fail (conversation never completes) without the fix.

ouroboros/leios_merged.go
cryptodj413: Fixed in 1463882b.

- `EndorserBlockTxHashesByHash` now requires `data.slotVerified` alongside completeness.
- `resolveCertifiedEndorserTxs` gets the same gate on its cache lookup.
- `leiosClosureCompleteLocked` gets the same gate, so `waitForLeiosEndorserClosure` (the NtC merge wait path) no longer reports — or wakes a waiter on — a complete-but-unverified closure.
- `storeLeiosEndorserBlock`'s closure-wake signal is withheld along with it (it used completeness alone before). Since that leaves a closure that was already complete at store time unsignaled forever, `bindLeiosEndorserBlockSlot`'s promotion now signals it once the slot is verified.

Regression: `TestEndorserBlockTxHashesByHashWithholdsUnverifiedSlot`, `TestLeiosClosureCompleteLockedWithholdsUnverifiedEntry` (also covers the waiter-wakes-on-promotion path), and `TestResolveCertifiedEndorserTxsWithholdsUnverifiedSlot` (drives `certifiedEndorserBlockHash` through a real ledger-backed parent block, not just header decoding) in `ouroboros/leios_eb_announced_point_test.go` / `leios_merged_test.go`. All three fail without their respective fix.

cryptodj413:
Addressed both P1 findings from the latest review round (1463882b):

- **leiosnotify.go offer/in-progress hash-only checks**: MsgBlockOffer's cache-hit skip, MsgBlockTxsOffer's top-level skip, its dispatched fetch closure's own re-check, and `leiosFetchInProgress`'s dedup key are all now slot-aware (`leiosFetchClaimKey`), not hash-only. A same-hash offer at a different slot is fetched instead of silently swallowed.
- **Missing slotVerified gates**: `EndorserBlockTxHashesByHash`, `resolveCertifiedEndorserTxs`, and `leiosClosureCompleteLocked` now withhold a complete-but-unverified entry, matching `EndorserBlockTxsByHash`. `bindLeiosEndorserBlockSlot`'s promotion now wakes a closure waiter that was already complete when stored (previously unsignaled).

New regression tests (each confirmed to fail without its fix):
- `TestLeiosNotifyBlockOfferFetchesSameHashAtDifferentSlot`, `TestLeiosNotifyBlockTxsOfferFetchesSameHashAtDifferentSlot`, `TestLeiosFetchClaimKeyDistinguishesSlots` (`ouroboros/leiosnotify_slot_test.go`)
- `TestEndorserBlockTxHashesByHashWithholdsUnverifiedSlot`, `TestLeiosClosureCompleteLockedWithholdsUnverifiedEntry` (`ouroboros/leios_eb_announced_point_test.go`)
- `TestResolveCertifiedEndorserTxsWithholdsUnverifiedSlot` (`ouroboros/leios_merged_test.go`, drives `certifiedEndorserBlockHash` through a real ledger-backed parent block)

Also did a full pass over every `slotVerified`/hash-only-lookup site in `ouroboros/` for the same two bug classes; no further instances found. `ARCHITECTURE.md` updated for both invariants.

Validation: `go build ./...`, `go vet ./...`, `gofmt`, `golines`, `nilaway ./ouroboros/...` (clean), `golangci-lint run ./ouroboros/...` (0 issues), `modernize` (clean), full `ouroboros` and `ledger` package tests (including `-race`), `import-boundaries`, `docs-parity` — all pass.

ouroboros/ouroboros.go
```diff
@@ -202,6 +206,11 @@ type Ouroboros struct {
 	leiosDeferredMu            sync.Mutex
 	leiosDeferredAnnouncements map[string]leiosDeferredAnnouncement
 	leiosAnnouncementSizes     map[string]uint64
+	// leiosAnnouncementSlots records the slot each announced endorser-block
+	// hash was declared at, so a later leios-fetch offer or store can be
+	// bound to the point its announcement actually vouched for instead of
+	// trusting whatever point the offering connection supplies (issue #3513).
+	leiosAnnouncementSlots map[string]uint64
```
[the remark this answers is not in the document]
cryptodj413: Confirmed, fixed in 125c23fb.

`recordLeiosAnnouncementLocked`'s own slot-consistency check read `leiosAnnouncementSlots[ebKey]` directly instead of going through `leiosAnnouncedSlotLocked`'s expiry-aware lookup, so it never got the same treatment the offer/store path already has: a genuinely new announcement of the same content-addressed hash at a different slot, arriving after the earlier binding aged out of `leiosNotifyMaxAnnouncementAge`, was rejected as "inconsistent" forever. Switched that check to `leiosAnnouncedSlotLocked(ebHash.Bytes())`. The map itself stays one-slot-per-hash (a scalar), which is now correct: it only ever needs to hold the single live (non-expired) binding for a hash, not every historical occurrence.

Regression: `TestRecordLeiosAnnouncementAcceptsNewSlotAfterExpiry` in `ouroboros/leios_eb_announced_point_test.go`, confirmed to fail with exactly "announcement slot is inconsistent with a previously observed endorser block" without the fix.

ouroboros/leios_merged.go
```diff
@@ -537,7 +672,91 @@ func (o *Ouroboros) storeLeiosEndorserBlock(
...
+	// slot, or a waiter parked on it would otherwise sit until its wait
+	// window times out instead of waking on the closure it is already
+	// holding (issue #3513).
+	if verified.completeTxCache() {
+		for _, key := range data.cacheKeys {
+			o.signalLeiosClosureWaitersLocked(key)
+		}
+	}
+	o.leiosMu.Unlock()
+	o.publishLeiosEndorserBlock(
```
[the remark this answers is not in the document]
cryptodj413: Confirmed, fixed in 125c23fb.

`bindLeiosEndorserBlockSlot` now returns the publish step (vote emission, pipeline observation, persistence enqueue) as a closure instead of calling `publishLeiosEndorserBlock` inline. `recordLeiosAnnouncement` runs the closure only after unlocking `leiosAnnouncementsMu`, so those external handlers never run under a lock shared by every concurrent announcement. The two backfill callers (`FetchEndorserBlockByPoint`, `fetchEndorserBlockOnConn`), which hold no outer lock across the call, just invoke the closure immediately -- unchanged behavior for them.

Regression: `TestRecordLeiosAnnouncementPublishesAfterReleasingAnnouncementsLock` in `ouroboros/leios_eb_announced_point_test.go` -- a vote handler that itself acquires `leiosAnnouncementsMu` in `HandleEndorserBlock`, run through a 2s-bounded goroutine so a regression fails cleanly instead of hanging the test binary. Confirmed this reproduces a real self-deadlock without the fix.

ouroboros/leios_eb_announced_point_test.go
```diff
@@ -0,0 +1,793 @@
...
+			_ = o.storeLeiosEndorserBlock(
+				points[i],
+				blocks[i],
+				nil,
+				leiosStorePeerOffered,
+			)
+		}(i)
+		go func(i int) {
+			defer wg.Done()
+			recordTestLeiosAnnouncement(t, o, headers[i])
```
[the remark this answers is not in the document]
cryptodj413: Confirmed, fixed in 125c23fb.

Added `recordTestLeiosAnnouncementNoFail`, an error-returning twin of `recordTestLeiosAnnouncement` with no `require`/`t.FailNow` calls, and switched the goroutines in `TestStoreAndAnnouncementRaceAlwaysEndsVerified` to call it and collect the result into a per-index `[]error` slice, asserted with `require.NoError` back on the test goroutine after `wg.Wait()`.

cryptodj413:
Addressed all 3 findings from cubic's latest review (125c23fb):

- **P1 (`ouroboros.go:213`)**: `recordLeiosAnnouncementLocked`'s slot-consistency check now goes through `leiosAnnouncedSlotLocked`'s expiry-aware lookup instead of a raw map read, so a genuine new announcement of the same hash at a different slot is accepted once the earlier binding ages out, instead of being rejected as "inconsistent" forever.
- **P2 (`leios_merged.go:754`)**: `bindLeiosEndorserBlockSlot` now returns its publish step as a closure instead of running it inline, so `recordLeiosAnnouncement` can release `leiosAnnouncementsMu` before running it -- avoiding stalling (or deadlocking, if a handler needed that same lock) every concurrent announcement for the duration of vote/pipeline/persistence handling.
- **P3 (`leios_eb_announced_point_test.go:504`)**: `TestStoreAndAnnouncementRaceAlwaysEndsVerified`'s goroutines no longer call `require` (and so `t.FailNow`) off the test goroutine; a new error-returning helper collects results instead.

New regression tests, each confirmed to fail without its fix: `TestRecordLeiosAnnouncementAcceptsNewSlotAfterExpiry`, `TestRecordLeiosAnnouncementPublishesAfterReleasingAnnouncementsLock` (a bounded-timeout deadlock reproduction using a vote handler that itself takes the announcement lock).

Validation: build, vet, gofmt, golines, nilaway, golangci-lint (0 issues), full `ouroboros` package tests with `-race`, import-boundaries, docs-parity — all pass. `ARCHITECTURE.md` updated.

ouroboros/leiosnotify.go
```diff
@@ -1589,6 +1682,21 @@ func (o *Ouroboros) recordLeiosAnnouncement(
...
+	// leiosAnnouncedSlotLocked, not a raw map read: unlike ebSize (a property
+	// of the content-addressed bytes the hash commits to, so always identical
+	// across every legitimate occurrence), the slot is extrinsic and the same
+	// hash can legitimately recur at a different slot once the earlier
+	// binding has aged out. A raw read here would keep rejecting a genuine
+	// later announcement of the same hash as "inconsistent" forever, the same
+	// bug leiosAnnouncedSlotLocked's expiry check was already added to fix
+	// for the offer/store path (cubic review; issue #3513).
+	slot := header.SlotNumber()
+	if previousSlot, exists := o.leiosAnnouncedSlotLocked(ebHash.Bytes()); exists &&
```
wolf31o2: This still stores only one active slot per content hash and rejects every different-slot announcement while that binding is live. The manifest hash is content-addressed and the code/comments explicitly allow the same hash to recur at another slot after expiry; there is no protocol uniqueness guarantee here that prevents recurrence before the ten-minute window ends. A legitimate second occurrence is therefore dropped, so its offer/fetch and associated endorser data never become available for the ranking block that referenced it. Track active bindings by (hash, slot), or otherwise establish/enforce a protocol-level uniqueness rule before rejecting this case, and add a regression for a same-hash different-slot announcement while the first binding is still unexpired.

wolf31o2 (changes requested):
Requesting changes: one active-binding correctness issue remains for recurring Leios endorser-block content.

ouroboros/leiosnotify.go
cryptodj413: Implemented the full (hash, slot) tracking, in commit c182dd17.

- `Ouroboros.leiosEndorserBlocks` and `leiosFetchInProgress` are now keyed by `(slot, hash)` (`leiosBlockKey`), not hash alone, so two live occurrences of the same hash at different slots coexist as independent entries. This removes the old reject/evict-on-mismatch branches in `storeLeiosEndorserBlock`/`bindLeiosEndorserBlockSlot` entirely -- a composite-key lookup can only return the exact occurrence asked for.
- `leiosAnnouncementSlots` is now a set of live slots per hash, not a scalar. `recordLeiosAnnouncementLocked` no longer rejects a second, independently live announcement of the same hash as "inconsistent"; `leiosAnnouncementBindsSlotLocked" (replacing `leiosAnnouncedSlotLocked`) is a membership-plus-expiry check.
- `EndorserBlockProviderFunc`, `EndorserBlockTxsByHash`, `EndorserBlockTxHashesByHash`, `certifiedEndorserBlockHash`, `waitForLeiosEndorserClosure`, and the forge loop's `CertifiedEndorserBlockTxHashes` now take (or structurally derive) the caller's own expected slot and resolve exactly that occurrence.
- Deliberately scoped out: the persistent blob store (`em`/`et` keys) stays hash-only -- durable re-serving of a second concurrent occurrence after in-memory TTL eviction is a disclosed, narrower limitation (documented in `DATABASE.md`), separate from the in-memory offer/fetch/serving path this review is about.

Regression: `TestStoreLeiosEndorserBlockAcceptsDifferentSlotOfSameHashWhileFirstIsLive` in `ouroboros/leios_eb_announced_point_test.go` -- two independent, live (no expiry involved) announcements of the same hash at different slots, both stored, both independently verified, and both reachable via `EndorserBlockTxsByHash` at once. Also rewrote the existing tests whose premise (rejection/eviction on a different slot) no longer holds, and added a forge-loop regression (`TestCheckAndForgeProductionCertifiesLeiosEBAfterAdoption`) proving the correct slot is threaded through to `CertifiedEndorserBlockTxHashes`, confirmed to fail if the wrong slot is passed.

cryptodj413:
Implemented the full (hash, slot) redesign wolf31o2 requested (c182dd17), rather than the narrower expiry-only fix from the previous round:

- `Ouroboros.leiosEndorserBlocks` and `leiosFetchInProgress` are now keyed by `(slot, hash)`, not hash alone, so two live occurrences of the same content-addressed hash at different slots coexist as independent entries instead of one masking or evicting the other. This removes the old reject/evict-on-mismatch logic in `storeLeiosEndorserBlock`/`bindLeiosEndorserBlockSlot` entirely -- a composite-key lookup can only ever return the exact occurrence asked for.
- `leiosAnnouncementSlots` is now a set of live slots per hash. A second, independently live announcement of the same hash at a different slot is accepted, not rejected as inconsistent.
- `EndorserBlockProviderFunc`, `EndorserBlockTxsByHash`, `EndorserBlockTxHashesByHash`, `certifiedEndorserBlockHash`, `waitForLeiosEndorserClosure`, and the forge loop's `CertifiedEndorserBlockTxHashes` all now take (or structurally derive) the caller's own expected slot and resolve exactly that occurrence.
- Deliberately out of scope: the persistent blob store (`em`/`et` keys) stays hash-only; durable re-serving of a second concurrent occurrence after in-memory TTL eviction is a disclosed, narrower limitation (documented in `DATABASE.md`).

Rewrote every existing test whose premise (rejection/eviction on a different slot) no longer holds under the new design, and added new regressions:
- `TestStoreLeiosEndorserBlockAcceptsDifferentSlotOfSameHashWhileFirstIsLive` / `...CrossConnectionDifferentSlotsCoexistRegardlessOfOrder` / `...AuthoritativeAndAnnouncedOccurrencesCoexist` (`ouroboros/leios_eb_announced_point_test.go`) -- two live, unexpired occurrences of the same hash at different slots, both independently verified and reachable via `EndorserBlockTxsByHash` at once.
- `TestCheckAndForgeProductionCertifiesLeiosEBAfterAdoption` (`ledger/forging/forger_test.go`) extended to assert the forge loop passes the certified EB's own slot (not zero or the ranking block's slot) to `CertifiedEndorserBlockTxHashes`; confirmed to fail if the wrong slot is threaded through.
- `TestLeiosAnnouncementBindsSlotIgnoresExpiredBinding`, `TestPeerOfferedStoreUnderFabricatedSlotStaysPermanentlyUnverified`, and others updated for the new semantics.

Validation: `go build ./...`, `go vet ./...`, gofmt, golines, nilaway (clean on every touched file), golangci-lint (0 issues on `ouroboros`, `ledger`, and root), modernize (clean on touched files), full test suites for `ouroboros`, `ledger`, `ledger/forging`, and root — including `-race` on `ouroboros` and `ledger` — import-boundaries, docs-parity. `ARCHITECTURE.md` and `DATABASE.md` updated.

ouroboros/leios_merged.go
```diff
@@ -426,18 +465,49 @@ func (o *Ouroboros) storeLeiosEndorserBlock(
...
+	o.leiosAnnouncementsMu.Lock()
+	verified := origin == leiosStoreAuthoritative ||
+		o.leiosAnnouncementBindsSlotLocked(point.Hash, point.Slot)
 	block, err := lcommon.NewLeiosEndorserBlockFromCbor(blockRaw)
 	if err != nil {
+		o.leiosAnnouncementsMu.Unlock()
 		return fmt.Errorf("decode leios endorser block: %w", err)
 	}
-	cacheKeys := []string{leiosBlockKey(point.Hash)}
+	cacheKeys := []string{leiosBlockKey(point.Slot, point.Hash)}
```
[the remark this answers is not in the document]
cryptodj413: Confirmed, fixed in baaf5803.

The persistent blob store (`em`/`et` keys) is now keyed by (slot, hash) too, matching the in-memory cache. `types.LeiosEBManifestKey`/`LeiosEBTxsKey` append the slot after the hash; `SetLeiosEB`/`GetLeiosEBManifest`/`GetLeiosEBTxs` take the caller's expected slot directly (no more reporting a slot for the caller to compare). The async persistence writer's job-coalescing key (`ouroboros/leios_persist.go`) is also `(slot, hash)` now, so a manifest-only job for one occurrence can't be superseded by a complete job for a different occurrence of the same hash.

Regression: `TestLeiosPersistTwoOccurrencesOfSameHashPersistIndependently` in `ouroboros/leios_persist_test.go` -- two live occurrences of the same hash at different slots, both persisted with distinct tx bodies, both read back correctly. Confirmed to fail (reads back the wrong occurrence's tx body) when the key functions are reverted to hash-only.

ouroboros/leios_eb_announced_point_test.go
```diff
@@ -0,0 +1,973 @@
...
+		leiosStorePeerOffered,
+	))
+
+	data, ok := o.lookupLeiosEndorserBlock(point.Slot, point.Hash)
+	require.True(t, ok)
+	require.Equal(t, point.Slot, data.point.Slot)
+	require.True(t, data.slotVerified)
+}
+
+// TestStoreLeiosEndorserBlockCrossConnectionDifferentSlotsCoexistRegardlessOfOrder
```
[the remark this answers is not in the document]
cryptodj413: Confirmed, fixed in baaf5803.

Made the test table-driven over both arrival orders (`first-then-second` and `second-then-first`), both asserting the same coexistence invariants.

cryptodj413:
Addressed both of cubic's findings on the multi-slot redesign (baaf5803):

- **Test coverage gap**: `TestStoreLeiosEndorserBlockCrossConnectionDifferentSlotsCoexistRegardlessOfOrder` only tested one arrival order; made it table-driven over both orders.
- **P2, real gap**: the persistent blob store (`em`/`et` keys) was still hash-only, so of two live in-memory occurrences of the same hash, only the most-recently-persisted one survived for historical re-serving. `database/types.LeiosEBManifestKey`/`LeiosEBTxsKey` now append the slot; `SetLeiosEB`/`GetLeiosEBManifest`/`GetLeiosEBTxs` and the async persistence writer's job-coalescing key are all `(slot, hash)`-aware now. This closes the scoping decision I'd disclosed in the previous round rather than leaving it as a documented limitation.

Regression: `TestLeiosPersistTwoOccurrencesOfSameHashPersistIndependently` (`ouroboros/leios_persist_test.go`), confirmed to fail (reads back the wrong occurrence's tx body) when the key functions are reverted to hash-only.

Also: CI's full-repo `go-test (Linux, race)` job failed on `TestQuiesceForLiveLifecycleOpHandlesUninitializedOuroborosAndUnconfirmedConnShutdown` in `node_lifecycle_test.go`, a file this PR never touches. It passed 20/20 locally under `-race`, and the `ouroboros` package itself passed cleanly in that same CI run -- a pre-existing timing-dependent flake (the test's own comment concedes it relies on `select` tie-break behavior), not a regression from this change. Re-ran that job.

Validation: `go build ./...`, `go vet ./...`, gofmt, golines, nilaway (clean), golangci-lint (0 issues), modernize (clean on touched files), full test suites for `ouroboros`, `database` (including all plugin subpackages), `ledger`, `ledger/forging`, and root, plus `-race` on `ouroboros` -- all pass. `ARCHITECTURE.md` and `DATABASE.md` updated.

database/types/keys.go
```diff
@@ -136,33 +136,44 @@ func PoolCredentialStakeKey(
...
+	// available.
 	LeiosEBTxsKeyPrefix = "et"
 )
 
-// LeiosEBManifestKey builds the blob key for a Leios endorser-block manifest.
-func LeiosEBManifestKey(hash []byte) []byte {
-	key := make([]byte, 0, len(LeiosEBManifestKeyPrefix)+len(hash))
+// LeiosEBManifestKey builds the blob key for one occurrence of a Leios
+// endorser-block manifest, identified by its slot and hash together.
+func LeiosEBManifestKey(hash []byte, slot uint64) []byte {
```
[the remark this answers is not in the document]
cryptodj413: Confirmed, fixed in d053b323.

`database/types.LegacyLeiosEBManifestKey`/`LegacyLeiosEBTxsKey` rebuild the pre-issue-#3513 hash-only key format. `GetLeiosEBManifest`/`GetLeiosEBTxs` fall back to it on a miss under the new (hash, slot) key, validating the legacy manifest value's embedded 8-byte slot prefix against the requested slot before trusting it (and, for `GetLeiosEBTxs`, confirming that same match before trusting the paired legacy txs value, since the old format could only ever track one occurrence per hash).

Regression: `TestGetLeiosEBManifestFallsBackToLegacyKey` and `TestGetLeiosEBTxsFallsBackToLegacyKey` in the new `database/leios_test.go`, each confirmed to fail (blob key not found) when its respective fallback is reverted.

ARCHITECTURE.md
```diff
@@ -2334,8 +2334,7 @@ revoke-on-conflict closure, and the associated `MetadataStore` methods
...
...`/`et` blob keys so historical EBs can be re-served after the cache TTL. A missing endorser block or incomplete transaction set is declined with the protocol's typed `ErrBlockNotFound` or `ErrBlockTxsNotFound`, while malformed transaction bitmaps retain their validation error. Pushed votes (notify `MsgVotesOffer` tag 4) are forwarded to the `ledger/leios` vote manager. The LeiosVotes and LeiosFetch vote handlers delegate vote collection, serving, and emission to the `ledger/leios` vote manager (see "Leios Voting"); LeiosFetch vote requests return a valid empty `MsgVotes` when that optional manager is absent, while standalone LeiosVotes remains unavailable and returns its explicit error. A cached or reloaded endorser block is also handed to the `ledger/leios` pipeline manager (see "Leios Pipeline") for stage/timing tracking and equivocation detection. LeiosVotes pull requests remain outstanding while no servable votes exist and complete when new vote material arrives or the protocol shuts down, so an empty relay store is not treated as a mini-protocol error. Durable vote storage is still future work; CertRB production uses the current gouroboros/Dijkstra prototype certificate shape.
```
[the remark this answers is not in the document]
cryptodj413: Confirmed, fixed in d053b323. Updated the Era-Specific Validation paragraph's persistence description (blob keys and job coalescing) to (hash, slot), matching the Leios Networking section and DATABASE.md's rows. Also documented the new legacy-key read fallback there.

DATABASE.md
```diff
@@ -1145,8 +1145,8 @@ flowchart LR
...
+| `em` + endorser-block hash bytes (32) + big-endian slot `uint64` (8 bytes) | The raw endorser-block manifest CBOR received over leios-fetch `MsgBlock`. Written by `Database.SetLeiosEB` (the merged manifest+txs single-commit writer) on the asynchronous background persistence writer, off the leios-fetch hot path; the granular `Database.SetLeiosEBManifest` remains available. Read by `Database.GetLeiosEBManifest`. Used so a synced node can serve historical EB manifests to downstream peers via leios-fetch `MsgBlockRequest` after the in-memory 10-minute TTL has expired. Keyed by (hash, slot) together, matching the in-memory `Ouroboros.leiosEndorserBlocks` cache: the manifest is content-addressed and can be a live, independently required occurrence at more than one slot at once, and a hash-only key would let a second occurrence's persist silently overwrite the first (issue #3513 review; see `ARCHITECTURE.md`'s Leios Networking section). The slot is part of the key rather than the value now, so `GetLeiosEBManifest`/`GetLeiosEBTxs` take the caller's expected slot and resolve exactly that occurrence rather than reporting one and letting the caller compare. | Leios EB manifest serving |
```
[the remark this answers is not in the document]
cryptodj413: Confirmed, fixed in d053b323 -- this was actually ARCHITECTURE.md's Era-Specific Validation paragraph that was stale, not these rows. Updated it to match; also documented the new legacy-key read fallback in both docs.

cryptodj413:
Addressed cubic's three findings on the (slot, hash) blob-key migration (d053b323):

- **Real upgrade hazard (P2)**: manifests/transaction bodies persisted under the old hash-only `em`/`et` keys would become silently unreachable after upgrading, since `GetLeiosEBManifest`/`GetLeiosEBTxs` only looked up the new `(hash, slot)` key. Added `types.LegacyLeiosEBManifestKey`/`LegacyLeiosEBTxsKey` and a read-path fallback: on a miss under the new key, the legacy hash-only key is tried, with its embedded slot validated against the request before the record is trusted.
- **Docs inconsistency (P2/P3)**: `ARCHITECTURE.md`'s Era-Specific Validation section still described the persistence writer as hash-only keyed/coalescing. Updated it to match the Leios Networking section and `DATABASE.md`'s rows (which were actually correct), and documented the new legacy-key fallback in both.

Regression: `TestGetLeiosEBManifestFallsBackToLegacyKey`, `TestGetLeiosEBTxsFallsBackToLegacyKey` (new `database/leios_test.go`), each confirmed to fail when its respective fallback is reverted.

Also confirming: CI's full-repo `go-test (Linux, race)` passed cleanly on this push, confirming the earlier failure on `TestQuiesceForLiveLifecycleOpHandlesUninitializedOuroborosAndUnconfirmedConnShutdown` was the pre-existing flake I diagnosed, not a regression.

Validation: `go build ./...`, `go vet ./...`, gofmt, golines, nilaway (clean), golangci-lint (0 issues), full `database`/`ouroboros`/`ledger`/root test suites, docs-parity -- all pass.

database/leios.go
```diff
@@ -127,9 +159,23 @@ func (d *Database) GetLeiosEBTxs(hash []byte) ([]cbor.RawMessage, error) {
...
 	if err != nil {
-		return nil, err
+		if !errors.Is(err, types.ErrBlobKeyNotFound) {
+			return nil, err
+		}
+		legacyManifest, legacyErr := blob.Get(
+			blobTxn,
+			types.LegacyLeiosEBManifestKey(hash),
+		)
+		if legacyErr != nil || len(legacyManifest) < 8 ||
```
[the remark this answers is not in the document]
cryptodj413: Fixed in 6641a664: `GetLeiosEBTxs` now checks `errors.Is(legacyErr, types.ErrBlobKeyNotFound)` before falling through, so a real legacy-read failure propagates instead of collapsing into the exact-key not-found error. Covered by `TestGetLeiosEBTxsPropagatesRealLegacyReadError` (confirmed to fail without the fix via revert-and-rerun).

database/leios.go
```diff
@@ -54,38 +51,63 @@ func (d *Database) SetLeiosEBManifest(
...
-		return 0, nil, errors.New("GetLeiosEBManifest: stored value too short")
+	if !errors.Is(err, types.ErrBlobKeyNotFound) {
+		return nil, err
 	}
-	return binary.BigEndian.Uint64(val[:8]), val[8:], nil
+	legacyVal, legacyErr := blob.Get(
+		blobTxn,
+		types.LegacyLeiosEBManifestKey(hash),
+	)
+	if legacyErr != nil || len(legacyVal) < 8 {
```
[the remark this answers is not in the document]
cryptodj413: Fixed in 6641a664: `GetLeiosEBManifest` now checks `errors.Is(legacyErr, types.ErrBlobKeyNotFound)` before falling through, so a real legacy-read failure propagates instead of collapsing into the exact-key not-found error. Covered by `TestGetLeiosEBManifestPropagatesRealLegacyReadError` (confirmed to fail without the fix via revert-and-rerun).

cryptodj413:
Pushed two commits:

- `0fdc87be` — merged `origin/main` (9 commits, unrelated to Leios). Resolved a content conflict in `ARCHITECTURE.md` by combining this branch's (slot, hash) redesign prose with upstream's independent doc updates (leios-fetch request-timeout wording, and the `MsgBlockRangeRequest`/#3623 paragraph from #3772).
- `6641a664` — fixes the two remaining cubic-dev-ai findings on `database/leios.go`: `GetLeiosEBManifest`/`GetLeiosEBTxs` were collapsing any error reading the legacy fallback key (storage/network/auth failures included) into the exact-key's ordinary not-found error. Both now check `errors.Is(legacyErr, types.ErrBlobKeyNotFound)` first and propagate a real error as-is. New regression tests confirmed to fail without the fix via revert-and-rerun.

Full validation post-merge: build, vet, gofmt, golines, full test suite (`database`, `ouroboros`, `ledger`, `ledger/forging`, root) all pass; `import-boundaries`/`docs-parity` pass; `golangci-lint` clean across all module dirs and Windows GOOS. `nilaway` reports 17 pre-existing findings, all in files this PR does not touch — it isn't part of any CI workflow.

Both cubic threads replied to and resolved. Waiting on the fresh CI run and cubic's re-review of this push before re-requesting @wolf31o2.

database/leios_test.go
```diff
@@ -0,0 +1,156 @@
...
+	_, err = d.GetLeiosEBTxs(hash, slot+1)
+	require.Error(t, err)
+}
+
+// TestGetLeiosEBManifestPropagatesRealLegacyReadError is the cubic
+// regression: a genuine failure reading the legacy fallback key (storage,
+// network, auth) must surface as-is, not collapse into the exact-key's
+// ordinary ErrBlobKeyNotFound and look like the manifest was simply never
+// persisted.
+func TestGetLeiosEBManifestPropagatesRealLegacyReadError(t *testing.T) {
```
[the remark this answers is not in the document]
cryptodj413: Not actually a leak in this case, but switched to `newTestDB` anyway since it's simpler and matches the sibling fallback tests in this file (70350a83).

For the record on why it wasn't a leak: `openTestDB`'s real Badger store is tracked independently of `d.blob` -- `newTestDatabase` registers its own `t.Cleanup` that looks up the underlying `plugin.Host` via a package-level `testDatabaseHosts` map keyed by the `*Database` pointer itself, and stops it there. Overwriting `d.blob` afterward doesn't touch that mapping, so the real store was always closed correctly regardless. Still, no reason to open one here when every read is mocked.

wolf31o2 (changes requested):
Reviewed current head 70350a8. I cannot approve while the branch is merge-conflicting with main. Resolve the conflict, rerun the required checks, and request a fresh review on the resolved head.

cryptodj413:
Pushed `62eb1d50` — merged `origin/main` (9 more commits, including @wolf31o2's #3810 leios-fetch bitmap bounds fix, which touched `ouroboros/leiosnotify.go` and `ARCHITECTURE.md`).

`ouroboros/leiosnotify.go` auto-merged cleanly. `ARCHITECTURE.md` had a content conflict in the Leios Networking section, resolved by combining this branch's (slot, hash) redesign prose with upstream's new bitmap-validation sentence (#3523).

While resolving it I found and fixed a mistake from my *previous* merge-conflict resolution on this PR: an off-by-one in the line range I used to splice the resolved paragraph back in had silently dropped the following paragraph ("For the current respun prototype, the notify vote dialect is specifically the three-field `(announcing_rb_hash, voter_id, signature)` form...") entirely. It's restored in this commit -- confirmed via `grep` that it's back and the file has no stray conflict markers.

Validation: build, vet, gofmt, full `ouroboros`/`database`/`ledger`/root test suites, `import-boundaries`, `docs-parity` all pass.

wolf31o2 (approved):

PR #3608, opened 2026-08-28
fix(leios): bound NtC CertRB closure waits to the serving connection

### LINKED ISSUE #3514: leios: bound closure waits in NtC serving
NtC serving can remain parked for the full endorser-block closure wait while a
remote request is pending. The wait context is not tied closely enough to the
request or connection lifecycle.

Acceptance criteria:

- Apply a request-scoped closure deadline.
- Cancel waits when the request or connection ends.
- Return a defined incomplete-closure result.
- Cover completed, timed-out, and cancelled waits.

cryptodj413:
Fixes #3514

## Problem

`serveLeiosCertRbWithWait` (the NtC chainsync serving path for a certifying
ranking block's endorser closure) derived its wait context from
`context.Background()`, bounded only by the configured closure-wait timeout.
The wait was not tied to the request or connection that triggered it, so it
could remain parked for the full window even after the client had already
disconnected.

## Changes

- Add `leiosConnDoneContext`, a minimal `context.Context` wrapping a
  connection's done channel.
- Thread the chainsync server protocol instance's `DoneChan()` through
  `chainsyncServerBlockCbor` -> `serveLeiosRankingBlockCbor` ->
  `serveLeiosCertRbWithWait`, and combine it with `context.WithTimeout` so the
  wait ends on whichever comes first: the configured window, or the
  connection ending.
- Add a `wait_outcome` distinction (`resolved` / `timeout` / `cancelled`) to
  the returned error, the warning log, and the
  `dingo_leios_ntc_certrb_closure_wait_seconds` metric label, so a
  connection-driven cancellation is distinguishable from a natural timeout.
- Update `ARCHITECTURE.md`'s Leios CertRB Serving (NtC) section for the new
  bound and metric label.

## Tests

- `TestServeLeiosCertRbWithWaitCancelsOnConnectionDone` and
  `TestServeLeiosCertRbWithWaitCancelsWhenConnectionEndsMidWait`: configure a
  1-hour closure-wait timeout and verify the call still returns within 2s once
  the connection's done channel closes (already-done and closes-mid-wait
  cases). Verified both fail (time out) against the pre-fix code.
- Updated `TestServeLeiosCertRbWithWaitErrorsOnTimeout` and the metrics test
  for the new `connDone` parameter and `cancelled` wait-outcome label.

## Checks

- `make` (fmt + build): clean
- `golangci-lint run ./...`, `nilaway ./...`, `modernize ./...`: clean
- `make import-boundaries`, `make docs-parity`, `make golines` (no diff),
  `make gorm-check`: clean
- `make govulncheck`: 0 reachable vulnerabilities
- `go test ./ouroboros/... -race`: pass
- `go test ./internal/test/conformance/... -run TestRulesConformanceVectors`: pass
- `make test` (full repo, `-race`): pass. One unrelated flake
  (`TestQuiesceForLiveLifecycleOpHandlesUninitializedOuroborosAndUnconfirmedConnShutdown`,
  in `node_lifecycle_test.go`, unrelated to this diff) on the first full run
  under load; reproduced reliably in isolation with no failure, and a second
  full-suite run passed clean.

## Skipped

- DevNet (`internal/test/devnet/run-tests.sh`) and `--conformance` mode: not
  run — this change does not touch consensus, block production, chain
  selection, or epoch/nonce logic; it only bounds an existing NtC serving
  wait to connection lifetime.
- Postgres/MySQL conformance variants: skipped (no `POSTGRES_PASSWORD` /
  `MYSQL_ROOT_PASSWORD` set locally); CI's `go-test-linux` job runs these.

---
## Summary by cubic
Fixes #3514 by bounding the Leios NtC CertRB closure wait to the serving connection's lifetime, so a disconnected client no longer holds the wait for the full timeout window. The wait is now keyed by `ConnectionId` and released from the node's connection-closed handler rather than the protocol's done channel, which cannot close while the serving callback runs.

- Registers a per-connection waiter in `serveLeiosCertRbWithWait` and releases it via `ReleaseLeiosServeWaiters` from `handleConnManagerClosed`, which the connection manager drives independently of the protocol callback.
- Re-checks connection liveness after registering so an already-closed connection does not produce a wait nothing will release.
- Adds a `wait_outcome` label (`resolved` / `timeout` / `cancelled`) to the error, warning log, and `dingo_leios_ntc_certrb_closure_wait_seconds` metric so cancellation is distinguishable from a natural timeout.
- Adds regression tests using the real chainsync harness with a live peer disconnect, verifying both the disconnect release and that an up connection still ends at the timeout.

<sup>Written for commit e76317c25c43dc482f5254a302fced4c784e8069. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Leios serving waits now end immediately when the client connection closes.
  * Prevented waits from remaining open when a connection is already closed.
  * Timeout behavior remains bounded when the connection stays active.

* **Monitoring**
  * Added a `cancelled` outcome for waits ended by connection closure.
  * Cancellation is now reflected in unresolved serving metrics.

* **Documentation**
  * Updated architecture documentation to describe connection-aware wait handling.


cryptodj413:
@coderabbitai review

cryptodj413:
## Bot review status

- **Cubic**: completed, conclusion NEUTRAL, no actionable findings posted.
- **CodeRabbit**: rate-limited (OSS free-tier review quota) on both the initial run and a manual re-request; next available review window is ~51 minutes out as of this comment. Per project policy, a completed Cubic review is sufficient for bot review when CodeRabbit is rate-limited.

All CI checks are green (Conventional Commits, go-test Linux/Windows, golangci-lint, govulncheck, DCO). Ready for human review; @blinklabs-io/core has been requested via CODEOWNERS.

ouroboros/leios_merged.go
```diff
@@ -1061,7 +1094,10 @@ func (o *Ouroboros) chainsyncServerBlockCbor(
 	if p == nil || p.Mode() != protocol.ProtocolModeNodeToClient {
 		return block.Cbor, nil
 	}
-	return o.serveLeiosRankingBlockCbor(block)
+	// p.DoneChan() ties a pending closure wait to this connection's lifetime:
+	// it closes as soon as the connection ends, so the wait does not remain
+	// parked for the full window after the request it serves is gone.
+	return o.serveLeiosRankingBlockCbor(block, p.DoneChan())
```
wolf31o2: `Protocol.DoneChan()` cannot close while this synchronous `RequestNext` callback owns gouroboros's receive loop: protocol completion waits for `recvDoneChan`, which closes only after the callback returns. A real NtC harness confirmed that disconnect left `DoneChan()` open until the callback was manually released. An already-available CertRB therefore still waits for the full timeout after disconnect. Use a connection or transport lifecycle signal that closes independently of this callback, and cover it through the real chainsync protocol path rather than a manually closed helper channel.

wolf31o2 (changes requested):

ouroboros/leios_merged.go
cryptodj413: Confirmed and fixed — you're right on both counts, and the original tests were passing only because they closed a channel they made up themselves.

Traced in gouroboros v0.202.2: `recvLoop` dispatches `handleMessage` synchronously (`protocol/protocol.go:911`), `recvDoneChan` closes only in `recvLoop`'s deferred exit (`:884-886`), and `doneChan` closes only after `<-recvDoneChan` *and* `<-sendDoneChan` (`:186-190`). So `DoneChan()` cannot close while the callback it would release is still on the stack. `muxerDoneChan` does close independently, but it isn't exposed on `Protocol` or `Muxer`.

The release now comes from connmanager's per-connection `ErrorChan` watcher goroutine (`connmanager/connection_manager.go:843-882`), which is independent of the protocol callback and is the only close notification an NtC connection gets:

- `Ouroboros.leiosServeWaiters` keyed by `ConnectionId`; `serveLeiosCertRbWithWait` takes `ctx.ConnectionId` and registers a waiter instead of taking a channel.
- `ReleaseLeiosServeWaiters` is called from `Node.handleConnManagerClosed`, alongside the existing NtC `RemoveClient`.
- `registerLeiosServeWaiter` re-checks liveness after registering, since connmanager removes the connection from its map before invoking `ConnClosedFunc` — otherwise a close that already ran would leave a wait nothing releases.

Coverage is now through the real chainsync protocol path. The two hand-made-channel tests are deleted. New tests use the shared `ouroboros-mock` chainsync harness via `newChainsyncServerFixtureWithConfig(t, csmock.ModeNtC, ...)` — the same fixture that wires Dingo's real `chainsyncServerConnOpts` callbacks onto a real connection — and disconnect with `Harness.Disconnect()`:

- `TestLeiosServeWaitReleasedByRealPeerDisconnect` — 1h wait window, real peer disconnect, asserts prompt release.
- `TestLeiosServeWaitStillBoundedByTimeout` — timeout bound still reported as `timeout`, not `cancelled`.
- `TestLeiosServeWaiterNotRegisteredForClosedConnection` — the liveness race.
- `TestHandleConnManagerClosed_NtC_ReleasesLeiosServeWaiters` / `_NilOuroboros` — node wiring, built through the validating `NewOuroboros` with the full dependency set.

Each was verified to fail without its fix: substituting a never-closing release channel (what `DoneChan()` effectively is during the callback) makes the disconnect tests time out, and removing the `ReleaseLeiosServeWaiters` call from `node.go` fails the wiring test.

`ARCHITECTURE.md` now records why `DoneChan()` is not usable here so this isn't reintroduced.

cryptodj413:
Pushed `8b7d233f` addressing the review.

The `Protocol.DoneChan()` finding was correct — verified in gouroboros v0.202.2, and the original tests only passed because they closed a channel the test itself created rather than anything the protocol owns. The release now comes from connmanager's per-connection `ErrorChan` watcher goroutine (`ReleaseLeiosServeWaiters`, called from `Node.handleConnManagerClosed`), which runs independently of the parked server callback. Details in the inline reply.

Coverage moved onto the real chainsync protocol path via the shared `ouroboros-mock` chainsync harness and `Harness.Disconnect()`; the two hand-made-channel tests are deleted. Each new test was verified to fail without its fix.

Re-validated on the new head: `make` , `golangci-lint` (0 issues), `nilaway`, `modernize`, `make import-boundaries`, `make docs-parity`, `make golines` (no diff), `make gorm-check`, and `make test` (full repo, `-race`) all pass. nilaway/modernize report only pre-existing findings in files this change does not touch (`ledger/leios/manager_test.go`, `mithril/bootstrap_test.go`, generated `midnight/*.pb.go`, `internal/test/devnet/`). DevNet and the Postgres/MySQL conformance variants remain unrun locally, as noted in the description.

@wolf31o2 re-requested.

cryptodj413:
@coderabbitai Valid — fixed in `e76317c2`. The two manual `select` blocks in `TestHandleConnManagerClosed_NtC_ReleasesLeiosServeWaiters` now use `testutil.RequireNoReceive` for the precondition and `testutil.RequireReceive` for the release assertion, matching the repository's channel-assertion rule. Re-verified the test still fails when the `ReleaseLeiosServeWaiters` call is removed from `node.go`.

wolf31o2 (approved):

PR #3619, opened 2026-08-28
fix(blockfrost): bound pagination and address UTxO loading

### LINKED ISSUE #3520: blockfrost: bound pagination and address UTxO loading
One Blockfrost pagination path accepts an unbounded page value, while address
UTxO queries load and optionally reverse the complete result set before paging.
Large page values and large addresses cause unnecessary arithmetic and memory
work.

Acceptance criteria:

- Validate page and count ranges before calculating offsets.
- Stop address queries at the requested result window.
- Preserve ordering and response compatibility.
- Cover large pages, empty pages, and large address result sets.

arepala-uml:
## Summary

One Blockfrost pagination path accepted an unbounded `page` value, and the
address/account UTxO endpoints always loaded and fully materialized an
address's or account's entire live UTxO set before slicing out one page
(full-reversing it for descending order). This bounds both paths to the
requested page instead.

- `ParsePagination` now clamps `page` the same way it already clamped
  `count`.
- `AccountUTXOs` (stake-credential UTxOs) now uses a cheap SQL `COUNT` plus
  a `LIMIT`/`OFFSET`/`ORDER`-bound fetch — its address pattern only needs a
  coarse SQL match, so this is safe and complete.
- `AddressUTXOs` still needs CBOR-based exact-address matching to compute
  an accurate total (a coarse `COUNT`/`OFFSET` is not safe there), so it
  now does a reference-only scan (no assets, no full rows) for the total
  and ordering, then fetches full data for only the requested page.

## Changes, file by file

- `api/blockfrost/pagination.go` — `ParsePagination` clamps `page` to
  `MaxPaginationPage`; adds `paginationOffset` (overflow-safe page→offset).
- `api/blockfrost/adapter_account_activity.go` — `AccountUTXOs` rewritten
  to use a `COUNT` plus a bounded `Limit`/`Offset`/`Descending` fetch
  instead of loading and reversing the full result set.
- `api/blockfrost/adapter.go` — `AddressUTXOs` rewritten to fetch matching
  UTxO references only (no assets/full rows), then fetch full data for
  just the requested page via the existing `UtxosByRefs` path. Removes the
  now-unused `paginateUtxos` helper.
- `database/models/utxo.go` — adds `Offset`, `Descending`, and
  `SkipAssets` to `UtxoWithOrderingQuery`, plus two new sentinel errors for
  invalid combinations.
- `database/plugin/metadata/sqlstore/utxo.go` — `GetUtxosByAddressWithOrdering`
  honors the new query options; adds `CountUtxosByAddressWithOrdering`
  (plain `COUNT(*)`, rejected for exact-address patterns).
- `database/plugin/metadata/store.go` — declares
  `CountUtxosByAddressWithOrdering` on the store interface.
- `database/utxo.go` — adds `Database.CountUtxosByAddressWithOrdering` and
  `Database.MatchingUtxoRefsByAddressWithOrdering` (the bounded
  reference-only scan for exact-address patterns).
- `ledger/state.go` — thin `LedgerState` wrappers for both new database
  methods.
- `DATABASE.md` — documents the new query options and the
  count-vs-reference-scan split between `AccountUTXOs` and `AddressUTXOs`.

Closes #3520 

---
## Summary by cubic
Fixes dingo/3520 by bounding Blockfrost pagination and the address UTxO endpoints: `page` is now clamped like `count`, and loading a single page no longer materializes an account's or address's entire live UTxO history.

**Bug Fixes**

- `ParsePagination` now clamps `page` to `MaxPaginationPage`; the page-to-offset conversion is overflow-safe.
- Huge `page` values now return an empty page past the end instead of reaching unchecked offset arithmetic.

**Refactors**

- `AccountUTXOs` uses a SQL `COUNT` plus a `LIMIT`/`OFFSET`/`ORDER`-bound fetch instead of loading and reversing the full result set.
- `AddressUTXOs` fetches matching UTxO references only (no assets, no full rows), then loads full data for just the requested page via `UtxosByRefs`.
- Both endpoints pass one shared read transaction to their count/scan and page-fetch calls, so the reported total and returned page describe the same snapshot under concurrent chain updates.
- Both exact-address scans — `MatchingUtxoRefsByAddressWithOrdering` and `UtxosByAddressWithOrdering`'s page-fill loop — now carry the last transaction ID in their keyset cursor, so snapshot-imported rows tied on slot, block index, and output index survive batch boundaries.
- The reference scan no longer applies the page-fill candidate cap that had turned large address listings into errors.
- `GetUtxosByAddressWithOrdering` gains `Offset`, `Descending`, and `SkipAssets` query options, with `Offset` and `Descending` rejected for keyset `After` and `Offset` rejected for exact-address patterns.
- Adds `CountUtxosByAddressWithOrdering` and `MatchingUtxoRefsByAddressWithOrdering` database methods backing the two endpoints.

<sup>Written for commit 03f737b54bb7f499676970f0235a1b550c54726d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3619?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added reliable pagination and total counts for address and account UTxO queries.
  - Supports ascending or descending ordering, offsets, page limits, and efficient large-result handling.
  - Preserves native asset details when requested, with options to omit asset data for faster responses.
  - Ensures totals and page results reflect the same database snapshot.

- **Bug Fixes**
  - Prevents invalid combinations of keyset pagination, offsets, and descending ordering.
  - Correctly handles empty results, oversized pages, and pages beyond the available data.

### DIFF
--- a/api/blockfrost/adapter_account_activity.go
+++ b/api/blockfrost/adapter_account_activity.go
@@ -42,8 +42,15 @@ func (a *NodeAdapter) AccountUTXOs(
 	if err != nil {
 		return nil, 0, err
 	}
+	// Shared across the existence check, count, and page fetch below so
+	// the total and the returned page describe the same snapshot: two
+	// separate (nil-txn) calls could otherwise straddle a concurrent
+	// commit and return a page inconsistent with the reported total.
+	txn := a.ledgerState.Database().Transaction(false)
+	defer txn.Release()
+
 	if _, err := a.ledgerState.Database().
-		GetAccountByCredential(credentialTag, stakeKey, true, nil); err != nil {
+		GetAccountByCredential(credentialTag, stakeKey, true, txn); err != nil {
 		return nil, 0, err
 	}
 
@@ -56,8 +63,9 @@ func (a *NodeAdapter) AccountUTXOs(
 	addressPatterns := []models.UtxoAddressPattern{
 		{DelegationPart: stakeKey},
 	}
-	total, err := a.ledgerState.CountUtxosByAddressWithOrdering(
+	total, err := a.ledgerState.Database().CountUtxosByAddressWithOrdering(
 		&models.UtxoWithOrderingQuery{AddressPatterns: addressPatterns},
+		txn,
 	)
 	if err != nil {
 		return nil, 0, fmt.Errorf(
@@ -71,13 +79,14 @@ func (a *NodeAdapter) AccountUTXOs(
 		return []AccountUTXOInfo{}, total, nil
 	}
 
-	paged, err := a.ledgerState.UtxosByAddressWithOrdering(
+	paged, err := a.ledgerState.Database().UtxosByAddressWithOrdering(
 		&models.UtxoWithOrderingQuery{
 			AddressPatterns: addressPatterns,
 			Limit:           params.Count,
 			Offset:          offset,
 			Descending:      params.Order == PaginationOrderDesc,
 		},
+		txn,
 	)
 	if err != nil {
 		return nil, 0, fmt.Errorf(
--- a/api/blockfrost/adapter_account_activity.go
+++ b/api/blockfrost/adapter_account_activity.go
@@ -47,11 +47,36 @@ func (a *NodeAdapter) AccountUTXOs(
 		return nil, 0, err
 	}
 
-	utxos, err := a.ledgerState.UtxosByAddressWithOrdering(
+	// A stake credential's UTxO set is matched entirely by SQL (payment_key
+	// is irrelevant; staking_key alone is exact), so it does not need the
+	// CBOR-based exact-address filtering that AddressUTXOs does. That makes
+	// a cheap SQL COUNT and a LIMIT/OFFSET-bound fetch safe: unlike
+	// AddressUTXOs, this query never has to materialize the full UTxO
+	// history to page or total a large stake account.
+	addressPatterns := []models.UtxoAddressPattern{
+		{DelegationPart: stakeKey},
+	}
+	total, err := a.ledgerState.CountUtxosByAddressWithOrdering(
+		&models.UtxoWithOrderingQuery{AddressPatterns: addressPatterns},
+	)
+	if err != nil {
+		return nil, 0, fmt.Errorf(
+			"count account UTxOs for %q: %w",
+			stakeAddress,
+			err,
+		)
+	}
+	offset, ok := paginationOffset(params)
+	if !ok || offset >= total {
+		return []AccountUTXOInfo{}, total, nil
+	}
+
+	paged, err := a.ledgerState.UtxosByAddressWithOrdering(
 		&models.UtxoWithOrderingQuery{
-			AddressPatterns: []models.UtxoAddressPattern{
-				{DelegationPart: stakeKey},
-			},
+			AddressPatterns: addressPatterns,
+			Limit:           params.Count,
+			Offset:          offset,
+			Descending:      params.Order == PaginationOrderDesc,
 		},
 	)
 	if err != nil {
@@ -61,14 +86,7 @@ func (a *NodeAdapter) AccountUTXOs(
 			err,
 		)
 	}
-	total := len(utxos)
-	if params.Order == PaginationOrderDesc {
-		for left, right := 0, len(utxos)-1; left < right; left, right = left+1, right-1 {
-			utxos[left], utxos[right] = utxos[right], utxos[left]
-		}
-	}
 
-	paged := paginateUtxos(utxos, params)
 	txBlockHashes, err := a.addressUtxoBlockHashes(paged)
 	if err != nil {
 		return nil, 0, fmt.Errorf(
--- a/database/plugin/metadata/sqlstore/utxo.go
+++ b/database/plugin/metadata/sqlstore/utxo.go
@@ -1089,6 +1089,12 @@ func (s *Store) GetUtxosByAddressWithOrdering(
 			models.ErrDescendingKeysetUnsupported,
 		)
 	}
+	if query.After != nil && query.Offset > 0 {
+		return nil, fmt.Errorf(
+			"GetUtxosByAddressWithOrdering: %w",
+			models.ErrOffsetKeysetUnsupported,
+		)
+	}
 	if query.Offset > 0 && models.RequiresExactAddressFilter(query.AddressPatterns) {
 		return nil, fmt.Errorf(
 			"GetUtxosByAddressWithOrdering: %w",
--- a/database/plugin/metadata/sqlstore/utxo.go
+++ b/database/plugin/metadata/sqlstore/utxo.go
@@ -1025,20 +1025,13 @@ func (s *Store) GetUtxosByAddress(
 	return ret, nil
 }
 
-func (s *Store) GetUtxosByAddressWithOrdering(
+// utxoOrderingPredicate builds the WHERE predicate shared by
+// GetUtxosByAddressWithOrdering and CountUtxosByAddressWithOrdering. The
+// returned predicate excludes any keyset (query.After) bound, which only
+// GetUtxosByAddressWithOrdering applies.
+func utxoOrderingPredicate(
 	query *models.UtxoWithOrderingQuery,
-	txn types.Txn,
-) ([]models.UtxoWithOrdering, error) {
-	if query == nil {
-		return nil, fmt.Errorf(
-			"GetUtxosByAddressWithOrdering: %w",
-			models.ErrNilUtxoWithOrderingQuery,
-		)
-	}
-	db, ctx, err := s.readDBFromTxn(txn)
-	if err != nil {
-		return nil, err
-	}
+) (string, []any, error) {
 	predicate := "utxo.deleted_slot = 0"
 	args := []any{}
 	switch {
@@ -1053,10 +1046,7 @@ func (s *Store) GetUtxosByAddressWithOrdering(
 				&args,
 				pattern,
 			); err != nil {
-				return nil, fmt.Errorf(
-					"GetUtxosByAddressWithOrdering: %w",
-					err,
-				)
+				return "", nil, err
 			}
 		}
 		if len(branches) == 0 {
@@ -1067,11 +1057,7 @@ func (s *Store) GetUtxosByAddressWithOrdering(
 	}
 	if query.FilterByAsset {
 		if len(query.AssetPolicyID) == 0 {
-			return nil, fmt.Errorf(
-				"GetUtxosByAddressWithOrdering: "+
-					"asset filter requires non-empty policy id: %w",
-				models.ErrEmptyAssetPolicyID,
-			)
+			return "", nil, models.ErrEmptyAssetPolicyID
 		}
 		predicate += `
  AND EXISTS (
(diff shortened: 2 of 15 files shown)


arepala-uml:
## Testing

**New/updated tests**

- `api/blockfrost/pagination_test.go` — `page` clamping (zero, above max,
  a huge value), and `paginationOffset` (valid pages/counts, zero
  page/count, overflow-safe at max page × max count).
- `api/blockfrost/adapter_account_utxos_test.go` (new) — `AccountUTXOs`
  against a real (temp) DB with 250 seeded UTxOs: correct ascending page,
  correct descending page, a page past the end (empty, correct total), an
  absurdly large page (empty, no error), and a zero-UTxO account.
- `api/blockfrost/adapter_address_utxos_pagination_test.go` (new) —
  `AddressUTXOs` with 250 seeded UTxOs: ascending/descending paging, a page
  past the end, and a dedicated case proving a native asset on a UTxO
  still comes through correctly via the new reference-fetch path.
- `database/utxo_exact_address_test.go` — `CountUtxosByAddressWithOrdering`
  and `Offset`/`Descending` pagination on a coarse (credential) pattern;
  rejection of `Offset` and `Count` on exact-address patterns; rejection of
  `Descending` combined with keyset (`After`) pagination;
  `MatchingUtxoRefsByAddressWithOrdering` correctly excluding a
  pointer-address lookalike; and a 1,100-row case crossing the new scan's
  internal 1,024-row batch boundary (proving no dropped/duplicated rows).
- `database/plugin/metadata/sqlstore/utxo_test.go` — `SkipAssets` actually
  omits assets from the result without affecting which rows come back.
- `database/plugin/metadata/domain_stores_test.go` — updated golden method
  list for the new `CountUtxosByAddressWithOrdering` store method.

**Verification**

- Confirmed the new tests fail (don't even compile) with the fix reverted,
  so they're exercising the real change.
- `go build ./...`, `go vet ./...`, `golangci-lint run ./...`, `nilaway`,
  `modernize`, `gofmt` — all clean, repo-wide.
- `make import-boundaries`, `make docs-parity`, `make golines` — all pass.
- `go test ./...` (full repo) — all packages pass. Two unrelated tests
  (`bark`, `bin` — neither touched by this change) failed once under full
  parallel-suite load; reran both in isolation on this branch and on a
  clean `origin/main` checkout and they passed both times, confirming
  pre-existing timing/subprocess flakiness rather than a regression.
- `go test ./internal/test/conformance/...` — passes (no consensus/ledger
  behavior changed by this PR, so devnet was not run).

arepala-uml:
## Manual verification (real HTTP server, not just unit tests)

To close the gap between "unit tests call the adapter directly" and "this
actually works when dingo is serving requests," I started the real
`Blockfrost` HTTP server (`New(...).Start(ctx)` — the exact same code path
`dingo` runs in API mode) against a throwaway seeded database (300 live
UTxOs on one address, one carrying a native asset, and 300 on one stake
account), bound it to a real local port, and hit it with real HTTP
requests. Results below; the harness itself was a temporary local test
file, not something added to the repo.

**Address UTxOs — ascending, page 1 (`count=5`)**

Total is 300, page-total 60 (300/5), first item's amount includes the
native asset:

```
GET /api/v0/addresses/{addr}/utxos?page=1&count=5
Status: 200 OK
X-Pagination-Count-Total: 300
X-Pagination-Page-Total: 60
Body: [..., {"amount":[{"unit":"lovelace","quantity":"1000000"},
             {"unit":"...4d414e55414c534d4f4b45","quantity":"7"}], ...}, ...]
```

**Address UTxOs — ascending, page 3** → correctly items 11–15
(`amount` 11000000..15000000).

**Address UTxOs — page 999999 (huge page)** → no crash, correct total/page
totals preserved, empty body:

```
GET /api/v0/addresses/{addr}/utxos?page=999999&count=5
Status: 200 OK
X-Pagination-Count-Total: 300
X-Pagination-Page-Total: 60
Body: []
```

**Address UTxOs — descending, page 2** → correctly the 6th–10th newest
(amounts 295000000 down to 291000000, out of 300).

**Account UTxOs — ascending, page 1** → correct amounts (2000000,
4000000, ... 10000000) and each row's `address` recovered per-UTxO.

**Account UTxOs — descending, page 3** → correct window (amounts
580000000 down to 572000000).

**Account UTxOs — page 999999, count=100 (huge page)** → empty, no crash,
`X-Pagination-Count-Total: 300`, `X-Pagination-Page-Total: 3`.

**Address UTxOs — `page=0`** → `400 Bad Request` with
`{"error":"Bad Request","message":"Invalid pagination parameters."}`,
confirming the strict validator on this endpoint still rejects an
out-of-range page rather than silently clamping it.

I checked every returned amount against the expected math by hand (offset
= `(page-1)*count`, ascending vs. reversed windows) — all matched. No
crash, no timeout, and no unbounded work on any of the huge-page requests
against either endpoint.

api/blockfrost/adapter_account_activity.go
```diff
@@ -47,11 +47,36 @@ func (a *NodeAdapter) AccountUTXOs(
...
+
+	paged, err := a.ledgerState.UtxosByAddressWithOrdering(
 		&models.UtxoWithOrderingQuery{
-			AddressPatterns: []models.UtxoAddressPattern{
-				{DelegationPart: stakeKey},
-			},
+			AddressPatterns: addressPatterns,
+			Limit:           params.Count,
+			Offset:          offset,
+			Descending:      params.Order == PaginationOrderDesc,
```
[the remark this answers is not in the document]
arepala-uml: Fixed: `AccountUTXOs` and `AddressUTXOs` now open one `Database.Transaction(false)` and pass it to both their count/reference-scan call and their page-fetch call, matching the existing `AccountWithdrawals` pattern. Added `TestCountAndFetchShareSnapshotWithinOneTxn`, which proves a write committed between the two calls is invisible to a read reusing the shared txn (confirmed it fails if either call reverts to a nil txn). See the latest commit.

database/plugin/metadata/sqlstore/utxo.go
```diff
@@ -1084,6 +1070,39 @@ func (s *Store) GetUtxosByAddressWithOrdering(
...
+			"GetUtxosByAddressWithOrdering: %w",
+			models.ErrDescendingKeysetUnsupported,
+		)
+	}
+	if query.Offset > 0 && models.RequiresExactAddressFilter(query.AddressPatterns) {
+		return nil, fmt.Errorf(
+			"GetUtxosByAddressWithOrdering: %w",
+			models.ErrOffsetRequiresCoarseMatch,
+		)
+	}
```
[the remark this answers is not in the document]
arepala-uml: Fixed: added an `After != nil && Offset > 0` check alongside the existing `After`/`Descending` guard, returning a new `models.ErrOffsetKeysetUnsupported`. Added `TestUtxosByAddressWithOrderingRejectsOffsetKeyset` covering it. See the latest commit.

arepala-uml:
## Addressed CodeRabbit findings

Both findings were valid — fixed in the latest commit:

1. **Snapshot consistency across count and page fetch.** `AccountUTXOs`
   and `AddressUTXOs` each issued their count/reference-scan call and their
   page-fetch call with a `nil` transaction, so each opened and released
   its own read transaction. A commit landing between the two (a live node
   adding or spending a UTxO for the same address mid-request) could make
   the reported total and the returned page describe different UTxO sets.
   Both now open one `Database.Transaction(false)` and pass it to both
   calls, matching the existing `AccountWithdrawals` pattern elsewhere in
   this file. Added `TestCountAndFetchShareSnapshotWithinOneTxn`, which
   commits an out-of-band write between the two calls and proves a read
   reusing the shared txn doesn't see it (I confirmed the test fails, as
   expected, if either call is reverted to a `nil` txn).

2. **`After` + `Offset` combination not rejected.** The existing guard only
   rejected `Offset` combined with exact-address patterns, not `Offset`
   combined with keyset (`After`) pagination — mixing the two would filter
   to rows after the cursor and then additionally skip `Offset` rows within
   that filtered set, silently returning a page shifted by both controls.
   Added the missing check and a new `models.ErrOffsetKeysetUnsupported`,
   plus `TestUtxosByAddressWithOrderingRejectsOffsetKeyset`.

As a side effect, `LedgerState.CountUtxosByAddressWithOrdering` and
`LedgerState.MatchingUtxoRefsByAddressWithOrdering` are removed — both call
sites now go through `Database` directly (as `AccountWithdrawals` already
did) so they can share a transaction, leaving those wrappers unused.

**Re-verified:** full build/vet/lint/test suite still clean, and I reran
the manual HTTP smoke test from the earlier comment against the fixed code
— identical, correct results for every request.

database/utxo.go
```diff
@@ -723,6 +723,133 @@ func (d *Database) UtxosByAddressWithOrdering(
...
+					Hash: batch[i].TxId,
+					Idx:  batch[i].OutputIdx,
+				})
+			}
+		}
+		if len(batch) < scanQuery.Limit || len(batch) == 0 {
+			break
+		}
+		last := batch[len(batch)-1]
+		scanQuery.After = &models.UtxoOrderingCursor{
```
wolf31o2, line 826: **P1 — Carry the final transaction ID into the keyset cursor.** Snapshot imports can share the snapshot slot, block index zero, and output index because they have no producing transaction. The SQL order and bound include `utxo.tx_id`, but this continuation leaves `TxId` empty, so the next batch revisits tied rows. Address scans then duplicate results until the candidate cap and valid snapshot-backed requests fail. Set `TxId: last.TxId` and cover a fixture with more than 1024 same-slot/imported rows.

database/utxo.go
```diff
@@ -723,6 +723,133 @@ func (d *Database) UtxosByAddressWithOrdering(
...
+	scanQuery.SkipAssets = true
+	scanQuery.Offset = 0
+	scanQuery.Descending = false
+	refs := []models.UtxoId{}
+	candidatesProcessed := 0
+	for {
+		remainingCandidates := exactAddressCandidateScanLimit -
+			candidatesProcessed
+		if remainingCandidates <= 0 {
+			return refs, errExactAddressCandidateScanLimit
```
wolf31o2: **P1 — Do not make valid high-cardinality addresses unlistable.** This scan returns `errExactAddressCandidateScanLimit` after 10,000 coarse candidates. The previous `AddressUTXOs` query used `Limit == 0` and did not impose that cap, so an address with more than 10,000 live UTxOs (or enough pointer siblings) now fails even for page 1. Keep materialization bounded while continuing/streaming the exact scan, or define and expose an API-level limit instead of converting a valid listing into a server error.

wolf31o2 (changes requested):
Two pagination/UTxO correctness regressions remain; existing bot findings are fixed and resolved.

database/utxo.go
arepala-uml: Fixed: the exact-address scan's keyset cursor now carries `TxId: last.TxId`, matching `UtxoOrderingCursor`'s documented tie-break contract. Added `TestMatchingUtxoRefsByAddressWithOrderingSnapshotTieBreak`, which seeds >1024 snapshot-imported rows sharing slot/block index/output index and proves the batch-boundary continuation neither drops nor duplicates a tied row (confirmed it fails without the fix). See commit eeefa10c.

database/utxo.go
arepala-uml: Fixed: removed exactAddressCandidateScanLimit from this scan. It was borrowed from UtxosByAddressWithOrdering's page-fill loop, which can stop once a bounded page's worth of matches is found; this scan must visit every coarse candidate to produce an accurate total and cannot stop early, so the cap only ever turned a valid, merely large, address listing into a hard error. SkipAssets and the reference-only result already bound the per-candidate cost. Added TestMatchingUtxoRefsByAddressWithOrderingExceedsOldCandidateScanLimit, seeding 10,050 exact-matching rows and asserting the listing completes (confirmed it fails with the old cap). See commit eeefa10c.

arepala-uml:
@coderabbitai review

arepala-uml:
## Addressed CodeRabbit's outside-diff finding (`database/utxo.go` 717-721)

Valid — fixed in the latest commit. `UtxosByAddressWithOrdering`'s exact-address
page-fill scan built the same keyset cursor as
`MatchingUtxoRefsByAddressWithOrdering` and had the same missing `TxId`, so
snapshot-imported rows tied on slot/block index/output index could be dropped
or duplicated across the scan's batch boundary.

Added `TestUtxosByAddressWithOrderingSnapshotTieBreak`, which seeds 118
pointer-sibling rows followed by 12 exact-address matches, all tied on
slot/block index/output index, so the matches straddle the scan's 128-row
batch boundary. Confirmed it drops the 2 matches past the boundary without
the fix (10 of 12 returned) and passes with it.

See commit 03f737b5.

arepala-uml:
@coderabbitai review

arepala-uml:
## Status

Both correctness regressions from @wolf31o2's review are fixed (commits
eeefa10c, 03f737b5), each with a regression test confirmed to fail without
its fix and pass with it. CodeRabbit's own re-review independently
surfaced the same missing-`TxId` class in the sibling scan
(`UtxosByAddressWithOrdering`), also fixed and tested.

Bot re-review status:
- CodeRabbit: rate limited on the second manual trigger ("Review rate
  limited", plan allows 1 included review/hour). The first re-trigger did
  complete and found the sibling issue above, which is now fixed.
- Cubic: unavailable — org has exceeded its monthly line-review quota,
  resuming 2026-09-01.

`go test ./database/... ./api/blockfrost/...`, `golangci-lint run`,
`go vet`, and `nilaway` all pass locally on the current head.

Re-requested review from @wolf31o2.

wolf31o2 (approved):

PR #3620, opened 2026-08-28
fix(devnet): isolate concurrent DevNet runs across worktrees

### LINKED ISSUE #3264: devnet: isolate Docker Compose resources across worktrees
## Problem

The DevNet scripts are not isolated across worktrees. Services pin global container names such as dingo-1, while Compose project resources also share the devnet namespace. Concurrent validation runs can recreate or remove each other’s containers and volumes even when host ports are unique.

Observed while validating #3254:

- an accelerated run passed readiness, transaction propagation, block propagation, and four-node agreement, then all four containers disappeared simultaneously before the epoch-transition checkpoint;
- the preserved Compose status contained no containers and the per-node logs were unavailable, which rules out an ordinary single-node failure;
- a serialized retry failed during docker compose up with No such container immediately after Compose recreated that container;
- dedicated host ports did not prevent the collision.

NodeControl also invokes docker stop/start using the pinned global container names, so setting COMPOSE_PROJECT_NAME alone cannot isolate the scenario.

## Acceptance criteria

- concurrent DevNet runs from separate worktrees use distinct containers, networks, volumes, and Compose project state;
- NodeControl targets the containers belonging to its own run;
- teardown removes only resources created by that run;
- accelerated peer-interruption and relay-restart phases continue to exercise real container stop/start behavior.

arepala-uml:
## Summary

Concurrent DevNet runs from separate worktrees could collide: pinned
`container_name`s meant one worktree's containers could be recreated or
removed by another, and a hardcoded `172.20.0.0/24` network meant a second
worktree's `docker compose up` failed outright with `Pool overlaps with
other one on this address space`. Each worktree now gets its own Compose
project, container namespace, network/subnet, and rendered topology config.

## Changes by file

- **`internal/test/devnet/compose-project.sh`** (new) — Shared helper
  sourced by the other scripts. Derives a stable `COMPOSE_PROJECT_NAME` and
  a distinct network subnet (`DEVNET_NET_BASE`) from the worktree's absolute
  path, and renders the checked-in `topology/*.json` peer lists into a
  per-run directory with that subnet's addresses substituted in.

- **`internal/test/devnet/docker-compose.yml`** — Removed the
  `container_name` pin on every service so Compose scopes each container to
  its own project. Parameterized the network subnet, every service's static
  `ipv4_address`, and the topology bind mounts via `DEVNET_NET_BASE` /
  `DEVNET_TOPOLOGY_DIR` (both default to the original literal values, so a
  bare `docker compose up` still works for a single run).

- **`internal/test/devnet/start.sh`** — Derives this run's Compose project
  and renders its topology config before bringing containers up.

- **`internal/test/devnet/stop.sh`** — Derives this run's Compose project,
  tears down only that project, and removes its rendered topology
  directory.

- **`internal/test/devnet/run-tests.sh`** — Same project/topology setup as
  `start.sh`, plus scopes the health-check and volume-discovery logic (which
  used to assume fixed container/volume names) to this run's project, and
  cleans up the rendered topology directory in its exit trap.

- **`internal/test/devnet/nodectl.go`** — `NodeControl` now requires the
  Compose project and resolves each service's container ID through
  `docker compose -p <project> ps` before stopping/starting it, instead of
  assuming the container name equals the service name.

- **`internal/test/devnet/isolation_test.go`** (new) — Tests covering the
  isolation behavior (see testing comment below).

- **`internal/test/devnet/README.md`** — Documents the new project/subnet
  derivation and how to override it.

Closess #3264 

---
## Summary by cubic
Fixes #3264: concurrent DevNet runs from separate worktrees no longer collide. Each run now derives its own Compose project, a collision-checked network subnet, a free block of host ports, and a rendered topology config (via `DEVNET_*` variables that keep the original literals as defaults), so one run can no longer recreate, remove, or fail against another's containers, networks, or ports.

**Bug Fixes**
- Dropped all `container_name` pins so Compose scopes containers, volumes, and networks per project.
- `devnet_net_base` checks Docker's existing networks (skipping IPv6) and walks to a genuinely free /24; `devnet_compose_up` retries "Pool overlaps" and host-port bind failures after recomputing the subnet or port block. `devnet_ports` similarly shifts the 11 published host ports to a free block and backs off if the caller set any override.
- `start.sh --accelerated` now prints the exact `go test` command with this run's derived project and port addresses, since auto-derived ports may differ from the defaults.
- `NodeControl` resolves containers via `docker compose -p <project> ps` and errors when the compose project is unset.
- New `isolation_test.go` covers project, subnet, port, IPv6, retry, and topology-render derivation, including that rendering never mutates the checked-in sources.

**Migration**
- Hand-rolled scenario invocations must now export `DEVNET_COMPOSE_PROJECT` or `COMPOSE_PROJECT_NAME`; `run-tests.sh` sets it automatically.
- To run more than one DevNet in the same worktree, explicitly set `COMPOSE_PROJECT_NAME`, `DEVNET_NET_BASE`, and `DEVNET_TOPOLOGY_DIR`.

<sup>Written for commit ff4186c35882f90de2c2df8471320bf24887c617. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3620?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added worktree-isolated DevNet environments with unique Compose projects, networks, ports, and rendered topology files.
  * Added automatic collision detection and retry during DevNet startup.
  * Improved startup output with the exact commands and connection settings for accelerated scenarios.
  * Added project-scoped startup, testing, and cleanup workflows.

* **Bug Fixes**
  * Prevented conflicts between concurrent DevNet worktrees and existing Docker networks or host ports.

* **Tests**
  * Added end-to-end coverage for isolation, allocation, retries, topology rendering, and cleanup.

### DIFF
--- a/internal/test/devnet/compose-project.sh
+++ b/internal/test/devnet/compose-project.sh
@@ -40,13 +40,34 @@ _devnet_cidr_overlaps() {
   [[ ${a_lo} -le ${b_hi} && ${b_lo} -le ${a_hi} ]]
 }
 
-# Every subnet Docker currently has allocated, one per line. Best-effort:
-# an unreachable daemon just yields no exclusions, same as today.
+# True (exit 0) when the string is a dotted-decimal IPv4 CIDR ("a.b.c.d/n").
+# Docker networks can be IPv6 (a ULA /64, /8, etc.), and _devnet_ip_to_int
+# only understands IPv4 — an IPv6 subnet reaching it would abort
+# start.sh/run-tests.sh with a bash arithmetic error before Compose ever
+# runs, rather than just being (correctly) treated as no constraint on our
+# IPv4-only range.
+_devnet_is_ipv4_cidr() {
+  [[ "$1" =~ ^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}/[0-9]{1,2}$ ]]
+}
+
+# Every IPv4 subnet Docker currently has allocated, one per line. IPv6
+# subnets are filtered out (see _devnet_is_ipv4_cidr). Best-effort: an
+# unreachable daemon just yields no exclusions, same as today.
 _devnet_used_subnets() {
   local name
   docker network ls --format '{{.Name}}' 2>/dev/null | while read -r name; do
     docker network inspect "${name}" \
       --format '{{range .IPAM.Config}}{{.Subnet}}{{"\n"}}{{end}}' 2>/dev/null
+  done | while read -r subnet; do
+    # Not `_devnet_is_ipv4_cidr "${subnet}" && printf ...`: under
+    # pipefail+errexit, a false condition on the very last line (the
+    # closing "no subnet" line every `docker network inspect` emits for a
+    # network without one) makes `&&` short-circuit non-zero, which kills
+    # the whole run-tests.sh/start.sh process right here. `if` doesn't
+    # have that problem — a false condition with no `else` still exits 0.
+    if _devnet_is_ipv4_cidr "${subnet}"; then
+      printf '%s\n' "${subnet}"
+    fi
   done
 }
 
@@ -161,14 +182,18 @@ _devnet_port_free() {
 # (3010/3013-3015, 3020-3023, and conformance's 3010-3012) collide outright
 # with "port is already allocated". Skips entirely if the caller has
 # already set any of _DEVNET_PORT_VARS, so a manual override stays in full
-# manual control.
+# manual control. Sets DEVNET_PORTS_AUTO=1 when it actually derived the
+# block, so devnet_compose_up knows it's safe to pick a new one on a bind
+# failure rather than fighting a caller's explicit choice.
 devnet_ports() {
+  DEVNET_PORTS_AUTO=""
   local var
   for var in "${_DEVNET_PORT_VARS[@]}"; do
     if [[ -n "${!var:-}" ]]; then
       return
     fi
   done
+  DEVNET_PORTS_AUTO=1
 
   local project_root project_hash block_size base attempt candidate i port all_free
   project_root="$(cd "${SCRIPT_DIR}/../../.." && pwd -P)"
@@ -200,11 +225,30 @@ devnet_ports() {
   return 1
 }
 
-# `docker compose up -d`, retrying if this run's subnet lost a race with a
-# concurrent worktree between devnet_net_base's check and the network
-# actually being created. On that failure the losing worktree's network
-# now shows up in `docker network ls`, so recomputing DEVNET_NET_BASE (and
-# re-rendering topology against it) naturally avoids it on the next try.
+# True (exit 0) when docker compose's failure output looks like a host port
+# already being bound, as opposed to some other bring-up failure. Docker's
+# wording for this varies by version/OS ("port is already allocated" on
+# older Docker, "Ports are not available: ... already in use" on newer).
+_devnet_looks_like_port_conflict() {
+  case "$1" in
+    *"port is already allocated"* | *"Ports are not available"* | \
+      *"address already in use"*)
+      return 0
+      ;;
+    *) return 1 ;;
+  esac
+}
+
+# `docker compose up -d`, retrying twice if this run's subnet or host ports
+# lost a race with a concurrent worktree between the earlier check
+# (devnet_net_base / devnet_ports) and Docker actually claiming them:
+#
+#   - subnet: a "Pool overlaps" failure means the losing worktree's network
+#     now shows up in `docker network ls`, so recomputing DEVNET_NET_BASE
+#     (and re-rendering topology against it) naturally avoids it next try.
+#   - ports: a port-bind failure only gets retried when devnet_ports itself
+#     derived the block (DEVNET_PORTS_AUTO=1) — a caller's explicit port
+#     override is never second-guessed, so that case still fails through.
 devnet_compose_up() {
   local compose_file="$1" attempt out
   for (( attempt = 1; attempt <= 3; attempt++ )); do
@@ -213,12 +257,27 @@ devnet_compose_up() {
       return 0
     fi
     printf '%s\n' "${out}" >&2
-    if [[ "${out}" != *"Pool overlaps"* ]] || [[ ${attempt} -eq 3 ]]; then
+    if [[ ${attempt} -eq 3 ]]; then
       return 1
     fi
-    echo "[compose-project] subnet collided with a concurrent worktree;" \
-      "picking a new one and retrying (attempt ${attempt}/3)" >&2
-    unset DEVNET_NET_BASE
-    devnet_render_topology
+    if [[ "${out}" == *"Pool overlaps"* ]]; then
+      echo "[compose-project] subnet collided with a concurrent worktree;" \
+        "picking a new one and retrying (attempt ${attempt}/3)" >&2
+      unset DEVNET_NET_BASE
+      devnet_render_topology
+      continue
+    fi
+    if [[ "${DEVNET_PORTS_AUTO:-}" == "1" ]] &&
+      _devnet_looks_like_port_conflict "${out}"; then
+      echo "[compose-project] host port collided with a concurrent" \
+        "worktree; picking a new block and retrying (attempt ${attempt}/3)" >&2
+      local var
+      for var in "${_DEVNET_PORT_VARS[@]}"; do
+        unset "${var}"
+      done
+      devnet_ports
+      continue
+    fi
+    return 1
   done
 }
--- a/internal/test/devnet/compose-project.sh
+++ b/internal/test/devnet/compose-project.sh
@@ -12,14 +12,59 @@ devnet_compose_project() {
   export COMPOSE_PROJECT_NAME="dingo-devnet-${project_hash}"
 }
 
-# Select a stable, worktree-specific /24 for the DevNet bridge network.
-# Compose project-scopes the network's *name*, but its subnet is a
-# separate axis: the checked-in topology/*.json files address peers by a
-# hardcoded 172.20.0.0/24 IP (devnet_render_topology rewrites those), and
-# Docker refuses to create two networks with the same subnet even under
-# different project names ("Pool overlaps with other one on this address
-# space"). Concurrent worktrees therefore each need their own range, not
-# just their own network name.
+# --- IPv4 CIDR helpers, used to check a candidate subnet against every
+# --- network Docker already knows about (see devnet_net_base). --------
+
+_devnet_ip_to_int() {
+  local IFS=.
+  # shellcheck disable=SC2206 # splitting "a.b.c.d" on IFS=. is the point
+  local -a o=(${1})
+  echo $(( (o[0] << 24) + (o[1] << 16) + (o[2] << 8) + o[3] ))
+}
+
+# Prints "<first> <last>", the inclusive integer host range for a CIDR.
+_devnet_cidr_range() {
+  local cidr="$1" addr prefix base mask
+  addr="${cidr%/*}"
+  prefix="${cidr#*/}"
+  base=$(_devnet_ip_to_int "${addr}")
+  mask=$(( (0xFFFFFFFF << (32 - prefix)) & 0xFFFFFFFF ))
+  base=$(( base & mask ))
+  echo "${base} $(( base + (1 << (32 - prefix)) - 1 ))"
+}
+
+_devnet_cidr_overlaps() {
+  local a_lo a_hi b_lo b_hi
+  read -r a_lo a_hi < <(_devnet_cidr_range "$1")
+  read -r b_lo b_hi < <(_devnet_cidr_range "$2")
+  [[ ${a_lo} -le ${b_hi} && ${b_lo} -le ${a_hi} ]]
+}
+
+# Every subnet Docker currently has allocated, one per line. Best-effort:
+# an unreachable daemon just yields no exclusions, same as today.
+_devnet_used_subnets() {
+  local name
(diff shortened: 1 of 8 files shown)


arepala-uml:
## Testing

**Unit tests** — full package, including the new `isolation_test.go` tests
(`TestComposeServicesDoNotPinContainerNames`,
`TestComposeProjectIsStableAndWorktreeSpecific`,
`TestDevNetScriptsSelectComposeProject`,
`TestComposeNetworkSubnetIsParameterizedPerWorktree`,
`TestComposeTopologyMountsUseRenderedDirectory`,
`TestNetBaseIsStableAndWorktreeSpecific`,
`TestRenderTopologyRewritesAddressesWithoutMutatingSource`) and the
pre-existing shell-harness tests in `run_tests_script_test.go`:

```
$ go test ./internal/test/devnet/... -v
...
PASS
ok  	github.com/blinklabs-io/dingo/internal/test/devnet	3.587s
```

69 subtests passed, 0 failed.

**Live Docker verification** (proves the actual bug, since unit tests can't call the real Docker daemon):

Before the fix, two networks requesting the old hardcoded `172.20.0.0/24` subnet collide exactly as the issue describes:

```
$ docker network create --driver bridge --subnet 172.20.0.0/24 repro-net-a   # succeeds
$ docker network create --driver bridge --subnet 172.20.0.0/24 repro-net-b
Error response from daemon: invalid pool request: Pool overlaps with other one on this address space
```

After the fix, two simulated worktrees derive distinct projects/subnets and both `docker network create` calls succeed concurrently:

```
worktree A: project=dingo-devnet-4111070087  subnet=172.31.112.0/24  -> created OK
worktree B: project=dingo-devnet-2279779412  subnet=172.28.138.0/24  -> created OK
```

Also confirmed with `docker compose config`: unset env vars resolve to the original literal `172.20.0.x` values (bare `docker compose up` unaffected); with `DEVNET_NET_BASE`/`DEVNET_TOPOLOGY_DIR` set, they resolve to the overridden values.

**Live end-to-end run**: `./run-tests.sh --accelerated` (full build, 4-node bring-up, accelerated scenario timeline including the peer-interruption/relay-restart phases that exercise the changed `NodeControl` code, teardown) 

Log: 
[devnet-live-run-accelerated.log](https://github.com/user-attachments/files/31576834/devnet-live-run-accelerated.log)

**Result:** PASS (exit code 0)

--- PASS: TestAcceleratedScenarioTimeline (120.82s)
[run-tests] txpump submitted=3 rejected=0 error=0

The peer-interruption and relay-restart phases (which stop/start `dingo-3`
and `dingo-relay` mid-run) both recovered successfully — this is the part
of the scenario that exercises the `NodeControl` container-resolution
change directly.

internal/test/devnet/run-tests.sh
```diff
@@ -42,6 +42,10 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROJECT_ROOT="$(cd "${SCRIPT_DIR}/../../.." && pwd)"
 COMPOSE_FILE="${SCRIPT_DIR}/docker-compose.yml"
+# shellcheck source=compose-project.sh
+source "${SCRIPT_DIR}/compose-project.sh"
+devnet_compose_project
```
wolf31o2, line 47: **P1 — Derive per-run host ports too.** This establishes a distinct Compose project but leaves every published host port at its shared default (3010/3013/3014/3015 and 3020–3023; conformance also publishes 3010–3012). A second default `run-tests.sh` from another worktree therefore fails during `compose up` with `port is already allocated`; the README still requires manual overrides. Derive and collision-check a port block alongside the project, or stop claiming concurrent worktrees run unmodified.

internal/test/devnet/compose-project.sh
```diff
@@ -0,0 +1,61 @@
...
+# 172.24-172.31 stays clear of the static subnets the antithesis/
+# archive-demo (172.21.0.0/24) and erastest (172.22.0.0/24) stacks pin, and
+# of Docker's own default address pool, which starts allocating from
+# 172.17.0.0/16. Callers may override it, same as COMPOSE_PROJECT_NAME.
+devnet_net_base() {
+  if [[ -n "${DEVNET_NET_BASE:-}" ]]; then
+    return
+  fi
+  local project_root project_hash second third
+  project_root="$(cd "${SCRIPT_DIR}/../../.." && pwd -P)"
```
wolf31o2, line 99: **P2 — Make subnet allocation collision-free across concurrent worktrees.** This reduces the path hash to only 8×256 = 2,048 `/24` networks. Distinct worktree roots can therefore derive the same `DEVNET_NET_BASE`, so their differently named Compose networks still fail with Docker's overlapping-pool error. Allocate and lock a free subnet, or add a collision check, rather than assuming this truncated hash is unique.

wolf31o2 (changes requested):
The worktree-specific project and topology paths are useful, but two concurrent-run collision paths remain.

internal/test/devnet/compose-project.sh
```diff
@@ -0,0 +1,224 @@
...
+  local project_root project_hash
+  project_root="$(cd "${SCRIPT_DIR}/../../.." && pwd -P)"
+  project_hash="$(printf '%s' "${project_root}" | cksum | awk '{print $1}')"
+  export COMPOSE_PROJECT_NAME="dingo-devnet-${project_hash}"
+}
+
+# --- IPv4 CIDR helpers, used to check a candidate subnet against every
+# --- network Docker already knows about (see devnet_net_base). --------
+
+_devnet_ip_to_int() {
```
wolf31o2, line 18: P2: `_devnet_used_subnets` feeds every Docker IPAM subnet into `_devnet_cidr_range`, but `_devnet_ip_to_int` assumes dotted-decimal IPv4. If any existing Docker network has an IPv6 subnet such as `fd00::/64`, this arithmetic parser fails and can abort `start.sh`/`run-tests.sh` before Compose starts. Filter to IPv4 CIDRs before parsing or make unsupported families no-ops, and add coverage for mixed IPv4/IPv6 networks.

wolf31o2 (changes requested):
The previous port and subnet-collision issues are addressed, but one portability gap remains.

internal/test/devnet/run-tests.sh
arepala-uml: Fixed in 00174ca7: `devnet_ports` now derives a worktree-specific block of all 11 host ports `docker-compose.yml` publishes, verifies each is actually free on `127.0.0.1` (shifting the whole block forward if any is occupied), and exports them as the `DEVNET_*_PORT` variables `docker-compose.yml` already reads. It backs off entirely if the caller has already set any one of them, so a manual override stays in full manual control. Verified live: bound a real socket on the derived base port and confirmed the whole block shifts (`TestPortsAvoidOccupiedPorts`), plus a real `start.sh`/`stop.sh` cycle showing the derived ports actually applied.

internal/test/devnet/compose-project.sh
arepala-uml: Fixed in 00174ca7: `devnet_net_base` no longer trusts the hash alone — it queries every subnet `docker network ls`/`inspect` currently reports (real CIDR overlap math, not string equality) and walks forward from the hash-derived starting point until it finds one with zero overlap. Added `devnet_compose_up` to cover the residual race between that check and the network actually being created: on a "Pool overlaps" failure it recomputes `DEVNET_NET_BASE` (which now sees the other worktree's just-created network) and retries, up to 3 attempts. Verified with a fake `docker` reporting the hash-derived candidate as taken (`TestNetBaseAvoidsSubnetsDockerReports`) and a fake `docker` that fails the first `up -d` with the real "Pool overlaps" error then succeeds (`TestComposeUpRetriesOnPoolOverlap`).

internal/test/devnet/compose-project.sh
arepala-uml: Good catch — fixed in ff4186c3: `_devnet_used_subnets` now filters to dotted-decimal IPv4 CIDRs before they ever reach `_devnet_ip_to_int`, so an IPv6 network (a ULA `/64`, `/8`, etc.) is treated as no constraint instead of aborting the script. Verified with a fake `docker` reporting `fd00::/64` (`TestNetBaseIgnoresIPv6Subnets`) — confirmed `devnet_net_base` no longer crashes and still lands on a valid IPv4 candidate. (Also caught and fixed an unrelated regression my first attempt at this introduced: filtering with `cond && printf` inside a `while read` loop made the last line's failed match propagate a non-zero exit through pipefail+errexit, silently killing the whole script — switched to `if`, verified live under real `set -euo pipefail`.)

internal/test/devnet/compose-project.sh
```diff
@@ -0,0 +1,224 @@
...
+# re-rendering topology against it) naturally avoids it on the next try.
+devnet_compose_up() {
+  local compose_file="$1" attempt out
+  for (( attempt = 1; attempt <= 3; attempt++ )); do
+    if out=$(docker compose -f "${compose_file}" up -d 2>&1); then
+      printf '%s\n' "${out}"
+      return 0
+    fi
+    printf '%s\n' "${out}" >&2
+    if [[ "${out}" != *"Pool overlaps"* ]] || [[ ${attempt} -eq 3 ]]; then
```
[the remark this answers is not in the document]
arepala-uml: Fixed in ff4186c3: `devnet_compose_up` now also recognizes a host-port-bind failure (in addition to the subnet "Pool overlaps" case) and re-derives the port block via `devnet_ports` when it allocated the ports in the first place (`DEVNET_PORTS_AUTO=1`) — a caller's explicit port override is never retried away. Verified with `TestComposeUpRetriesOnPortConflict` (two subtests: auto-derived ports get retried, an explicit override doesn't).

internal/test/devnet/isolation_test.go
```diff
@@ -0,0 +1,562 @@
...
+			filepath.Join(root, "internal", "test", "devnet"), 0o755,
+		))
+	}
+
+	baseA := deriveNetBase(t, helper, worktreeA, "")
+	require.Equal(t, baseA, deriveNetBase(t, helper, worktreeA, ""))
+	baseB := deriveNetBase(t, helper, worktreeB, "")
+	require.NotEqual(t, baseA, baseB,
+		"two worktrees derived the same subnet; concurrent runs would race"+
+			" to create the same Docker network")
```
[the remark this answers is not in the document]
arepala-uml: Fixed in ff4186c3: removed the `require.NotEqual(baseA, baseB)` assertion — agreed it wasn't something the design actually guarantees on its own (2048-candidate hash space, no real Docker network created in that test), just something usually true. `TestNetBaseAvoidsSubnetsDockerReports` and `TestComposeUpRetriesOnPoolOverlap` already cover the real guarantee: a subnet actually in use gets skipped, and a collision surfaced by `docker compose up` gets retried onto a different one.

arepala-uml:
Verified locally at commit `ff4186c3`.

## Verification performed

### Unit and configuration checks

- Uncached DevNet isolation tests passed with the race detector.
- Shell syntax validation passed.
- Docker Compose configuration validation passed.
- The worktree remained clean after testing.

### Complete default DevNet suite

The complete default all-Dingo DevNet suite passed:

- All four Dingo nodes became healthy.
- DevNet package tests passed.
- Scenario tests passed.
- Epoch-boundary consensus passed.
- `txpump`: 9 submitted, 0 rejected, 0 errors.
- Containers, networks, volumes, and rendered topology were cleaned up successfully.

### Concurrent live DevNet verification

Two accelerated four-node DevNet suites were run concurrently.

Both runs initially selected the same candidate subnet, `172.28.182.0/24`. This exercised both live collision-recovery paths in the second run:

1. Docker rejected the initial network with `Pool overlaps`.
2. The script selected `172.28.183.0/24` and retried.
3. Docker then reported `port is already allocated`.
4. The script selected a new host-port block and retried successfully.

Both concurrent suites passed:

- Four-node readiness
- Block propagation
- Transaction propagation
- Steady-state chain agreement
- Epoch transition
- `dingo-3` interruption and recovery
- Relay restart and recovery
- Final chain agreement

Results:

- Suite A: PASS in 121.83s
  - `txpump`: 2 submitted, 0 rejected, 0 errors
- Suite B: PASS in 112.41s
  - `txpump`: 3 submitted, 0 rejected, 0 errors

Both Compose projects tore down independently with no remaining containers or networks.

### Mixed IPv4/IPv6 Docker IPAM verification

A real dual-stack Docker network was created with:

- IPv4: `172.28.182.0/24`
- IPv6: `fd00:3620::/64`

The allocator correctly:

- Ignored the IPv6 subnet.
- Recognized the occupied IPv4 subnet.
- Selected the next available IPv4 subnet, `172.28.183.0/24`.

## Logs

- [concurrent-a.log](https://github.com/user-attachments/files/31652206/concurrent-a.log)

- [concurrent-b.log](https://github.com/user-attachments/files/31652210/concurrent-b.log)


The attached logs include the complete builds, initial subnet collision, host-port collision, successful retries, scenario results, `txpump` counts, and teardown.

Concurrent DevNet isolation—including real subnet and host-port collision recovery—and mixed IPv4/IPv6 Docker IPAM handling are verified.

wolf31o2 (approved):
Prior port, subnet, and IPv6 portability findings are fixed on this head. CI is green; no new blockers found.

PR #3625, opened 2026-08-29
fix(ledger): validate phase two during historical replay

### LINKED ISSUE #3503: ledger: keep phase-2 validation enabled during historical replay
Historical replay can bypass local phase-2 checks when a block falls behind the
configured stability boundary. Historical and live paths therefore apply
different validation behavior during catch-up even when historical validation
is enabled.

Acceptance criteria:

- Apply the configured phase-2 checks across the historical boundary.
- Define an explicit result when required ledger context is not yet available.
- Cover replay, catch-up, and near-tip cases with both validity outcomes.

wolf31o2:
Historical replay currently skips local Plutus phase-2 evaluation for blocks
outside the stability window even when historical validation is enabled.

Keep phase-2 validation enabled for that configured mode while preserving the
trusted-replay shortcut when historical validation is disabled. Add focused
coverage for both behaviors.

Fixes #3503

---
## Summary by cubic
Fixes historical replay so local Plutus phase-2 evaluation runs when historical validation is enabled, instead of being skipped for blocks outside the stability window. Fixes #3503.

- Keeps the trusted-replay shortcut, which skips phase 2 only when historical validation is disabled.
- Adds coverage for the skip decision and for phase-2 behavior at the block-application boundary across historical, trusted-replay, and unvalidated scenarios.

<sup>Written for commit d4ad5d3c86ae5258b939f707ecf1d03e5c2cb1ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3625?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Historical validation now remains active during deep block processing when configured, improving validation accuracy across stability boundaries.
  * Trusted replay shortcuts no longer bypass phase-two validation when historical validation is enabled.

* **Tests**
  * Added coverage for historical validation, trusted replay, and unvalidated block scenarios.

### DIFF
--- a/ledger/block_transaction_validity_test.go
+++ b/ledger/block_transaction_validity_test.go
@@ -287,6 +287,131 @@ func TestLedgerProcessBlockDijkstraValidityOutcomeStateTransitions(
 		})
 	}
 }
+
+// TestLedgerProcessBlockHistoricalValidationRunsPhase2 verifies the
+// configuration decision at the real block-application boundary.  A
+// historical-validation replay must pass a LedgerView with phase-two
+// validation enabled; the trusted-replay control retains the skip shortcut.
+func TestLedgerProcessBlockHistoricalValidationRunsPhase2(t *testing.T) {
+	t.Parallel()
+
+	for _, tt := range []struct {
+		name              string
+		validationEnabled bool
+		wantValidationErr bool
+	}{
+		{
+			name:              "historical validation evaluates phase two",
+			validationEnabled: true,
+			wantValidationErr: true,
+		},
+		{
+			name:              "trusted replay keeps phase two skipped",
+			validationEnabled: false,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			db := newTestDB(t)
+			tx := omockledger.NewTransactionBuilder()
+			tx.WithId(bytes.Repeat([]byte{0x52}, 32))
+			tx.WithType(gdijkstra.TxTypeDijkstra)
+			tx.WithValid(true)
+			var txHash [32]byte
+			copy(txHash[:], tx.Hash().Bytes())
+			offsets := &database.BlockIngestionResult{
+				TxOffsets: map[[32]byte]database.CborOffset{
+					txHash: {BlockSlot: 10, ByteLength: 1},
+				},
+			}
+
+			phase2Called := false
+			testEra := eras.DijkstraEraDesc
+			testEra.ValidateTxFunc = func(
+				_ lcommon.Transaction,
+				_ uint64,
+				state lcommon.LedgerState,
+				_ lcommon.ProtocolParameters,
+			) error {
+				phase2Called = true
+				skipper, ok := state.(interface {
+					SkipPhase2Validation() bool
+				})
+				if ok && skipper.SkipPhase2Validation() {
+					return nil
+				}
+				return conway.PlutusScriptFailedError{
+					Err: errors.New("phase-two validation mismatch"),
+				}
+			}
+			pparams := &gdijkstra.DijkstraProtocolParameters{
+				ConwayProtocolParameters: conway.ConwayProtocolParameters{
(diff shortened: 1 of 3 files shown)


arepala-uml (changes requested):
## Request changes

The production change in `ledger/state.go` appears logically correct: deep historical blocks no longer skip phase-two validation when historical validation is enabled.

However, the new test only exercises the extracted boolean helper. It does not prove that phase-two validation runs through the real historical replay path. The test could remain green if `ledgerProcessBlocksFromSource` passed incorrect inputs, ignored the result, or otherwise failed to evaluate phase two.

Issue #3503 requires coverage for:

- Historical replay, catch-up, and near-tip processing.
- Valid and invalid phase-two outcomes.
- An explicit result when required ledger context is unavailable.

Please add a regression test through the actual block-processing path. For example:

    func TestHistoricalReplayRejectsPhase2MismatchWhenValidationEnabled(t *testing.T) {
        t.Parallel()

        // Configure the ledger to validate historical blocks.
        ls := newHistoricalReplayLedger(t, LedgerStateConfig{
            ValidateHistorical: true,
        })

        // Keep the chain tip more than k blocks ahead so the tested block is
        // classified as deep historical.
        setChainTipBeyondStabilityWindow(t, ls)

        // Construct a block containing a transaction whose declared validity
        // disagrees with local Plutus evaluation. The block should otherwise be
        // valid so the rejection specifically proves phase-two evaluation ran.
        block := makeHistoricalBlockWithPhase2Mismatch(t, ls)

        err := processHistoricalReplayBlock(t, ls, block)

        require.Error(t, err)
        var plutusErr conway.PlutusScriptFailedError
        require.ErrorAs(t, err, &plutusErr)
    }

The exact implementation should reuse existing repository or `ouroboros-mock` fixtures.

The regression test should:

1. Fail against the PR’s base commit because the deep historical path sets `skipPhase2Validation=true`.
2. Pass with this PR’s implementation.
3. Include a valid historical control block that succeeds.
4. Include a `ValidateHistorical=false` control proving the trusted-replay shortcut remains unchanged.
5. Assert the explicit behavior when phase-two evaluation cannot obtain required ledger context.
6. Exercise replay, catch-up, and near-tip classifications, as required by issue #3503.

Validation performed:

- Focused race-enabled ledger tests passed.
- Full `go test ./ledger` passed.
- `go test ./internal/test/conformance/` passed.
- `git diff --check` passed.
- Current GitHub CI checks are green.
- `DATABASE.md` and `ARCHITECTURE.md` were checked and are unaffected.

CodeRabbit reviewed the previous commit rather than the current head. Its earlier request for independent predicate cases was addressed, although the thread remains unresolved. The current-head CodeRabbit review was skipped, and Cubic completed with a neutral result.

I recommend requesting changes until a regression test through the real historical replay path proves that an invalid phase-two result is rejected and satisfies the acceptance criteria from issue #3503.

wolf31o2:
Added a regression through `ledgerProcessBlock` in `d4ad5d3c`. The test drives the configured historical-validation decision into the real block-application path, verifies phase-two mismatch rejection, and preserves the trusted-replay control. Focused validation: `GOWORK=off GOCACHE=/tmp/dingo-3503-review-cache go test ./ledger -run TestLedgerProcessBlockHistoricalValidationRunsPhase2 -count=1`.

wolf31o2:
The current head `d4ad5d3c` includes the real block-application regression test requested in review and the focused test passes. The changes-requested review is from the prior head; please re-review the current head.

arepala-uml (approved):

PR #3636, opened 2026-08-29
fix(config): make the bundled devnet usable

### LINKED ISSUE #3631: devnet genesis crosses an epoch boundary every 0.5 seconds
`config/cardano/devnet/shelley-genesis.json` sets `epochLength` 5 with `slotLength` 0.1, so the bundled devnet starts a new epoch twice a second. `byron-genesis.json` sets `protocolConsts.k` 2160 and `blockVersionData.slotDuration` 250, which makes a Byron epoch 21600 slots, so the two files also disagree about epoch length by more than three orders of magnitude.

Reported against 0.69.0 and 0.70.2 by a downstream consumer generating a single node devnet from the release image:

- after a few hundred epochs `GetStakePools` stops answering. cardano-cli issues that query unconditionally while balancing, so `cardano-cli transaction build` fails and the devnet can no longer be funded
- on 0.70.2 the ledger tip also stopped advancing while the forger kept producing

The reporter's validation only passed originally because it funded within the first minute.

The Conway randomness stabilisation window is 4k/f (`ledger/candidate_nonce.go:548`). With k=2160 and f=1.0 that is 8640 slots against a 5 slot epoch, so `computeCandidateNonceAsOf` takes the `stabilityWindow >= epochLengthInSlots` branch at `ledger/candidate_nonce.go:145` and freezes the candidate nonce at every epoch start.

Proposed genesis change, keeping this a single node devnet:

- `slotLength` 0.1 -> 1
- `epochLength` 5 -> 4320
- `securityParam` 2160 -> 432
- byron `protocolConsts.k` 2160 -> 432
- byron `blockVersionData.slotDuration` 250 -> 1000

That gives one second slots, one block per slot at `activeSlotsCoeff` 1.0, a 72 minute epoch, and 4k/f = 1728 which fits inside the epoch. Byron and Shelley then agree on a 4320 slot epoch.

Also add a load time check in `config/cardano/genesis_consistency.go` that rejects a Shelley genesis whose `epochLength` is smaller than 4k/f, so this cannot regress. Every other config passes today:

| config | k | f | epochLength | 4k/f |
| --- | --- | --- | --- | --- |
| mainnet | 2160 | 0.05 | 432000 | 172800 |
| preprod | 2160 | 0.05 | 432000 | 172800 |
| preview | 432 | 0.05 | 86400 | 34560 |
| sanchonet | 432 | 0.05 | 86400 | 34560 |
| musashi | 108 | 0.05 | 21600 | 8640 |
| internal/test/devnet generated | 40 | 0.4 | 500 | 400 |

The same genesis files are mirrored in blinklabs-io/docker-cardano-configs, which is what the release image ships as `/opt/cardano/config/devnet` and what downstream consumers actually copy. That repository needs the same change.

Found via input-output-hk/cardano-init#84.

### LINKED ISSUE #3632: devnet genesis ships pre-Chang Plutus cost models at protocol version 11
`config/cardano/devnet/shelley-genesis.json` declares `protocolVersion.major` 11, but the devnet genesis carries the original Conway cost models:

- `alonzo-genesis.json` `costModels.PlutusV1`: 166 entries
- `alonzo-genesis.json` `costModels.PlutusV2`: 175 entries
- `conway-genesis.json` `plutusV3CostModel`: 251 entries

Mainnet, preprod and preview all run protocol version 11 today with 332 / 332 / 350 entries, and the three networks carry byte identical values (checked against Koios `epoch_params` for mainnet epoch 652, preprod 310, preview 1404). Those counts match `lang.CostModelParamNamesV1/V2/V3` in plutigo v0.5.0 exactly. plutigo v0.4.0, which dingo currently pins, has 328/328/346 and tolerates the longer list because `costModelFromList` iterates parameter names and breaks at `len(data)`, so shipping the current models does not require a plutigo bump.

The script data hash is computed over the cost models, so any builder using current cost models produces a hash the devnet rejects, and no builtin added after Chang is priced at all. This surfaced in input-output-hk/cardano-init#84: the MeshJS template pinned Mesh's built-in cost models and every script transaction failed against the devnet, because Mesh's PlutusV3 constant has 297 entries and the devnet chain has 251.

Fix: replace the three devnet cost models with the protocol version 11 models the public networks run.

The same genesis files are mirrored in blinklabs-io/docker-cardano-configs, which is what the release image ships as `/opt/cardano/config/devnet`. That repository needs the same change.

Separately worth noting: plutigo v0.4.0 knows 328/328/346 parameters while the live networks publish 332/332/350, so dingo currently ignores the last four parameters of each on chain cost model. That is a plutigo bump, tracked separately from this issue.

wolf31o2:
## config/cardano/devnet

Timing, retimed to Yaci DevKit's default local cluster:

- `slotLength` 0.1 -> 1
- `epochLength` 5 -> 600
- `securityParam` 2160 -> 100, from Yaci's `k = ((epochLength * f) / 3) * 0.5`
- byron `protocolConsts.k` 2160 -> 60, so the Byron epoch (10k slots) is also 600
- byron `blockVersionData.slotDuration` 250 -> 1000

`activeSlotsCoeff` stays 1.0, so the single producer forges one block per second. 4k/f is 400 inside a 600 slot epoch.

Cost models, replaced with the protocol version 11 models mainnet, preprod and preview all publish:

- `alonzo-genesis.json` `costModels.PlutusV1`: 166 -> 332 entries
- `alonzo-genesis.json` `costModels.PlutusV2`: 175 -> 332 entries
- `conway-genesis.json` `plutusV3CostModel`: 251 -> 350 entries

The three networks carry byte identical values, and the counts match `lang.CostModelParamNamesV1/V2/V3` in plutigo v0.5.0. plutigo v0.4.0, pinned here, has 328/328/346 and tolerates the longer list because `costModelFromList` iterates parameter names and breaks at `len(data)`.

No other field in either genesis file changed.

## config/cardano/genesis_consistency.go

`validateGenesisConsistency` gains a second assertion: the Shelley `epochLength` must be strictly longer than the randomness stabilisation window 4k/f, computed the same way as `ledger.nonceStabilityWindow`. The comparison is strict because `computeCandidateNonceAsOf` (`ledger/candidate_nonce.go:145`) pins the cutoff to the epoch's first slot at `>=`, not `>`. A genesis missing `securityParam`, `epochLength`, or `activeSlotsCoeff` passes rather than being rejected on a zero value.

Margins for the bundled configs: mainnet and preprod 172800/432000, preview and sanchonet 34560/86400, musashi 8640/21600, devnet 400/600, generated `internal/test/devnet` config 400/500.

## Tests

- `TestValidateEpochLengthFitsNonceWindow`: per network margins, the pre-fix devnet, exact window/epoch equality, and ceiling rounding
- `TestValidateEpochLengthFitsNonceWindowIncompleteGenesis`: each missing input
- `TestEmbeddedConfigsPassGenesisConsistency`: loads every embedded network through `NewCardanoNodeConfigFromEmbedFS`
- `TestDevnetGenesisIsUsable`: pins the devnet timing values
- `TestDevnetCostModelsCoverEveryPricedParameter`: pins the counts and asserts each is at least the length of plutigo's parameter table

Each fails against the pre-fix genesis.

## Docs

`ARCHITECTURE.md` gains a Genesis Consistency Validation section. `README.md` records the devnet parameters. `DATABASE.md` checked and not affected: no metadata schema, query, blob layout, or pruning change.

The same genesis files are mirrored in blinklabs-io/docker-cardano-configs, which is what the release image ships as `/opt/cardano/config/devnet`.

Fixes #3631
Fixes #3632

---
## Summary by cubic
Makes the bundled devnet usable: the old genesis put an epoch boundary every 0.5 seconds (5-slot epoch at 0.1s slots) with a randomness stabilisation window longer than the epoch, freezing the candidate nonce, and shipped protocol v9-era cost models while declaring protocol version 11. Retimes to Yaci DevKit's local cluster defaults (1s slots, 600-slot epoch, `securityParam` 100, Byron k=60 with 1-second slots so a Byron epoch spans the same 600 seconds as a Shelley epoch) and replaces the cost models with the protocol v11 values mainnet, preprod, and preview publish. Bumps the pinned `cardano-configs` image to 20260829-1 so the embedded configs match what the release image ships as `/opt/cardano/config/devnet`. Byron k is inert here since Shelley hard-forks at epoch 0.

**Genesis validation**
- `validateGenesisConsistency` now rejects a genesis whose epoch length is not strictly longer than the 4k/f randomness stabilisation window, matching `ledger.nonceStabilityWindow`.
- Geneses missing `securityParam`, `epochLength`, or `activeSlotsCoeff` pass rather than failing on the zero value.
- Tests verify every embedded network passes, pin the devnet timing and cost model counts, and cover the window math; `ARCHITECTURE.md` documents the check and `README.md` records the devnet parameters.

<sup>Written for commit 5016e4505fc537aca8bda6ff546cf0578909892e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3636?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


wolf31o2:
Added `config-parity`, a CI check that diffs the embedded network configs against blinklabs-io/docker-cardano-configs.

`bin/config-parity.sh` compares every git-tracked file under the networks named in `config/cardano/embed.go`'s embed directive, so adding a network there brings it under the check. It covers all five embedded networks, 54 files, not just the devnet. `preview/README.md` is allowlisted as the one file with no counterpart upstream; anything else missing there is reported as drift.

`.github/workflows/config-parity.yml` runs it per pull request, which catches drift introduced here, and weekly, since drift introduced in docker-cardano-configs produces no event in this repository. It runs through `make config-parity` so the CI check and the local target cannot diverge. The workflow checks the configs repository out with `actions/checkout` and passes the path in `CARDANO_CONFIGS_DIR`; run locally without that variable the script clones the repository itself.

Note on ordering: the check compares against docker-cardano-configs `main`, so it fails on this PR until blinklabs-io/docker-cardano-configs#79 merges. Against that PR's branch it reports 54 files matching across 5 networks.

wolf31o2:
Split the config-parity CI check out to #3637 so it can merge on its own. Reverted here in 130f5b8c; this branch is back to config and docs only.

The check compares `config/cardano/` against the cardano-configs image tag the Dockerfile pins, so on this branch it stays red until docker-cardano-configs releases an image carrying the devnet change and this branch bumps the pinned tag. Keeping it here would have gated a genesis fix on a release in another repository.

Merge order for this branch:

1. blinklabs-io/docker-cardano-configs#79
2. a cardano-configs image release carrying it
3. a commit here bumping `ghcr.io/blinklabs-io/cardano-configs` in the Dockerfile to that tag

Step 3 is only required once #3637 has merged and the check is running. Until then this branch does not depend on it.

wolf31o2:
docker-cardano-configs 20260829-1 is released with the devnet change, so 5016e450 bumps the pinned tag in the Dockerfile. The embedded copy and the image the build pulls now carry the same devnet genesis.

Verified against the released image: all 54 files across the five embedded networks match 20260829-1, so the release carried the devnet change and nothing else. Running #3637's check against this branch reports `54 file(s) across 5 network(s) match ghcr.io/blinklabs-io/cardano-configs:20260829-1`.

This branch no longer depends on merge order. #3637 is based on main, which pins 20260817-1 and carries the matching pre-change configs, so it passes there too, and whichever of the two merges first leaves main consistent.

arepala-uml (approved):

PR #3637, opened 2026-08-29
ci: fail when embedded network configs drift from the shipped image

wolf31o2:
## What

`bin/config-parity.sh` diffs the network configs embedded in the binary against the docker-cardano-configs copy shipped in the release image. `make config-parity` runs it locally, and `.github/workflows/config-parity.yml` runs it in CI.

The embedded and image copies are the same configuration reached through two runtime paths. `CARDANO_NETWORK=<net>` with no config file serves `config/cardano/`; the release image ships the docker-cardano-configs copy as `/opt/cardano/config/`.

## Scope

The script compares every tracked file under the networks named by `config/cardano/embed.go`: 54 files across devnet, mainnet, musashi, preprod, and preview. `preview/README.md` is the only allowlisted file without an upstream counterpart.

## Comparison target

The PR job compares against the cardano-configs image tag pinned in the Dockerfile, currently `20260829-1`. A config update can therefore bump the pin and update `config/cardano/` in the same PR. A separate weekly job compares against upstream `main` and reports when a new image pin is due.

Repository branch protection must require the `config-parity` check context before this becomes a merge gate.

## Sources

`CARDANO_CONFIGS_DIR` uses an already extracted tree, `CARDANO_CONFIGS_REF` clones a git ref, and the default extracts `/config` from the pinned image.

## Verification

- passes for 54 files across 5 networks against `20260829-1`
- reports unified diffs for content drift
- reports missing counterpart paths
- `shellcheck` and `actionlint` pass
- `make docs-parity` passes with the new Makefile target

Split out of #3636 so it can merge independently.

## Summary by CodeRabbit

* **Chores**
  * Added automated checks to verify that embedded network configuration files remain identical to the corresponding release configurations.
  * Pull requests now receive configuration parity validation before merge.
  * Scheduled checks compare configurations with the upstream source weekly, with optional manual runs.
  * Added a pre-commit checklist step to detect configuration drift and report affected files with remediation guidance.

### DIFF
--- a/.github/workflows/config-parity.yml
+++ b/.github/workflows/config-parity.yml
@@ -1,0 +1,58 @@
+name: config-parity
+
+# The network configs under config/cardano/ exist twice: embedded in the
+# binary here (config/cardano/embed.go), and in
+# blinklabs-io/docker-cardano-configs, which the release image ships as
+# /opt/cardano/config/ (see the cardano-configs stage in the Dockerfile) and
+# which downstream tooling copies out of the image. CARDANO_NETWORK=<net>
+# resolves to one or the other depending on how dingo was deployed, so the
+# same network name means the same chain only while the two stay identical.
+on:
+  pull_request:
+  schedule:
+    - cron: '0 7 * * 1' # Monday 07:00 UTC
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # Gates merges. Compares against the cardano-configs image tag the
+  # Dockerfile pins, which is what a built image actually contains, so a
+  # config change is satisfiable inside one pull request here: bump the
+  # pinned tag and update config/cardano/ in the same commit. Comparing
+  # against the configs repository's main branch instead would pass as soon
+  # as that branch moved, while the image this repository builds still
+  # shipped the old config.
+  config-parity:
+    name: config-parity
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1 https://github.com/actions/checkout/releases/tag/v7.0.1
+        with:
+          persist-credentials: false
+      - name: config-parity
+        # Run through the Makefile so this CI check and `make config-parity`
+        # cannot drift apart.
+        run: make config-parity
+
+  # Does not gate merges. The job above pins to a released image, so it stays
+  # green while docker-cardano-configs moves ahead of that pin; this reports
+  # when it has, meaning a tag bump is due. Drift introduced in that
+  # repository produces no event here, hence the schedule.
+  config-upstream:
+    name: config-upstream
+    if: github.event_name != 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1 https://github.com/actions/checkout/releases/tag/v7.0.1
+        with:
+          persist-credentials: false
+      - name: config-parity-against-upstream-main
+        env:
+          CARDANO_CONFIGS_REF: main
+        run: make config-parity
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -32,6 +32,7 @@ Go Cardano node (Ouroboros). Derivable info (build targets, flags, package layou
 make lint         # import-boundaries, all modules, windows, nilaway, modernize
 modernize --fix ./...   # optional: auto-apply modernize's findings
 make docs-parity
+make config-parity  # embedded configs vs the pinned cardano-configs image; needs network + docker
 make golines
 make sql-check    # only when database/sql queries or sqlc.yaml changed
 make govulncheck  # reachable Go vulnerabilities; needs network
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ GO_TAG_FLAGS=$(if $(strip $(BUILD_TAGS)),-tags "$(BUILD_TAGS)",)
 # run modernize only against hand-written packages to avoid generator drift.
 MODERNIZE_PACKAGES=$(shell go list $(GO_TAG_FLAGS) -f '{{if .GoFiles}}{{.ImportPath}}{{end}}' ./... | grep -Ev '/database/plugin/(blob/(aws|gcs)|metadata/(mysql|postgres)|metadata/sqlstore/internal/query/(mysql|postgres|sqlite))$$|/midnight$$')
 
-.PHONY: all build help install uninstall mod-tidy clean format golines lint import-boundaries docs-parity proto sql sql-check govulncheck test test-live-lifecycle bench bench-ci bench-mempool bench-mempool-normal bench-mempool-degenerate bench-mempool-revalidation test-load test-load-log test-load-profile test-devnet
+.PHONY: all build help install uninstall mod-tidy clean format golines lint import-boundaries docs-parity config-parity proto sql sql-check govulncheck test test-live-lifecycle bench bench-ci bench-mempool bench-mempool-normal bench-mempool-degenerate bench-mempool-revalidation test-load test-load-log test-load-profile test-devnet
 
 # Default target
 all: format build ## Format and build (default)
@@ -95,6 +95,9 @@ import-boundaries: ## Check reviewed package import boundaries
 
 docs-parity: ## Check docs against go.mod, the Makefile, and the DevNet compose file
 	go test ./internal/docsparity
+
+config-parity: ## Fail if the embedded network configs drift from docker-cardano-configs
+	./bin/config-parity.sh
 
 proto: $(PROTOC) ## Generate Go code from protobuf definitions
 	go build -o $(TOOLS_BIN)/protoc-gen-go google.golang.org/protobuf/cmd/protoc-gen-go
--- a/bin/config-parity.sh
+++ b/bin/config-parity.sh
@@ -1,0 +1,183 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 Blink Labs Software
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Fail when the network configs embedded in the dingo binary have drifted
+# from the docker-cardano-configs copy the release image ships.
+#
+# The two copies must stay byte-identical because they are the same config
+# reached by two different runtime paths. `CARDANO_NETWORK=<net>` with no
+# config file serves the copy embedded here via config/cardano/embed.go,
+# while the release image ships the docker-cardano-configs copy as
+# /opt/cardano/config/ and downstream tooling copies it out of there. A
+# change landing in only one of them makes the Docker and binary
+# deployments run different chains under the same network name.
+#
+# The default comparison is against the cardano-configs image tag the
+# Dockerfile pins, not against that repository's main branch, because the
+# pinned tag is what a built image actually contains. That also makes the
+# check satisfiable inside a single dingo pull request: bump the pinned tag
+# and update the embedded copy together and it goes green, with no window
+# where it is red waiting on another repository to merge. Comparing against
+# main instead would go green the moment that branch moved, while the image
+# this repository builds still shipped the old config.
+#
+# Sources, in precedence order:
+#   CARDANO_CONFIGS_DIR  an existing checkout or extracted config tree
+#   CARDANO_CONFIGS_REF  a branch, tag, or commit SHA of the repository
+#   (default)            the image tag pinned in the Dockerfile
+
+set -euo pipefail
+
+configs_repo="https://github.com/blinklabs-io/docker-cardano-configs"
+configs_image="ghcr.io/blinklabs-io/cardano-configs"
+repo_root="$(cd "$(dirname "$0")/.." && pwd)"
+cd "$repo_root"
+
+# Files that exist here with no counterpart in docker-cardano-configs, and
+# are expected to. Anything else missing from that repository is drift.
+dingo_only=(
+	"preview/README.md"
+)
+
+# These files are shipped by cardano-configs but are intentionally not embedded
+# by dingo. Keep the allowlist explicit so any new upstream file still fails.
+upstream_only=(
+	"preview/db-sync-config.json"
+	"preview/tracer-config.json"
+	"mainnet/db-sync-config.json"
+	"mainnet/tracer-config.json"
+	"preprod/db-sync-config.json"
+	"preprod/tracer-config.json"
+	"musashi/tracer-config.json"
+)
+
+configs_dir="${CARDANO_CONFIGS_DIR:-}"
+configs_ref="${CARDANO_CONFIGS_REF:-}"
+source_desc=""
+cleanup_dir=""
+if [[ -n "${configs_dir}" ]]; then
+	source_desc="${configs_dir}"
+else
+	cleanup_dir="$(mktemp -d)"
+	# shellcheck disable=SC2064 # expand cleanup_dir now, not at trap time
+	trap "rm -rf '${cleanup_dir}'" EXIT
+	if [[ -n "${configs_ref}" ]]; then
+		source_desc="${configs_repo} at ${configs_ref}"
+		echo "cloning ${source_desc}"
+		git clone --quiet --filter=blob:none "${configs_repo}" \
+			"${cleanup_dir}/configs"
+		git -C "${cleanup_dir}/configs" fetch --quiet origin "${configs_ref}"
+		git -C "${cleanup_dir}/configs" checkout --quiet FETCH_HEAD
+		configs_dir="${cleanup_dir}/configs"
+	else
+		tag="$(sed -n "s|^FROM ${configs_image}:\\([^ ]*\\) .*|\\1|p" Dockerfile | head -1)"
+		if [[ -z "${tag}" ]]; then
+			echo "error: no ${configs_image} tag found in Dockerfile" >&2
+			exit 1
+		fi
+		source_desc="${configs_image}:${tag}"
+		echo "extracting /config from ${source_desc}"
+		container="$(docker create "${source_desc}")"
+		# shellcheck disable=SC2064 # expand both now, not at trap time
+		trap "docker rm --force '${container}' >/dev/null 2>&1 || true; rm -rf '${cleanup_dir}'" EXIT
+		mkdir -p "${cleanup_dir}/configs"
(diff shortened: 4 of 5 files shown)


bin/config-parity.sh
```diff
@@ -0,0 +1,146 @@
...
+embed_line="$(grep -m1 '^//go:embed ' config/cardano/embed.go)"
+read -r -a networks <<<"${embed_line#//go:embed }"
+if [[ ${#networks[@]} -eq 0 ]]; then
+	echo "error: no networks found in config/cardano/embed.go" >&2
+	exit 1
+fi
+
+drift=0
+checked=0
+for network in "${networks[@]}"; do
```
arepala-uml, line 126: **Blocking: the comparison is one-directional.** This loop only walks `git ls-files config/cardano/<network>` (dingo's own tracked files) and checks each has an upstream counterpart. It never walks the upstream tree to check for files dingo is missing.

Verified: extracted the real `20260829-1` image, added a new file under its `mainnet/` directory (simulating docker-cardano-configs shipping a file dingo hasn't picked up), and re-ran the script - it printed `54 file(s) across 5 network(s) match` and exited 0.

This is the exact failure mode the PR description says the check exists to catch ("the same config reached by two different runtime paths ... must stay byte-identical"). As written, if docker-cardano-configs adds a file that dingo's embedded copy doesn't pick up, this check - and the required CI job built on it - passes indefinitely. Suggest adding a reverse pass over each network's upstream directory (e.g. diffing `find` output or `git ls-files`-equivalent on the extracted tree) before declaring parity.

CLAUDE.md
```diff
@@ -31,6 +31,7 @@ Go Cardano node (Ouroboros). Derivable info (build targets, flags, package layou
 make lint         # import-boundaries, all modules, windows, nilaway, modernize
 modernize --fix ./...   # optional: auto-apply modernize's findings
 make docs-parity
+make config-parity  # embedded configs vs the pinned cardano-configs image; needs network + docker
```
arepala-uml, line 35: AGENTS.md carries a near-identical Pre-commit list but wasn't updated with this new `make config-parity` step - it still only lists lint/docs-parity/golines/sql-check/govulncheck. `make docs-parity` won't catch this since it doesn't cross-check AGENTS.md against CLAUDE.md. Worth adding the same line there so an agent following AGENTS.md doesn't skip the check.

bin/config-parity.sh
```diff
@@ -0,0 +1,146 @@
...
+# pinned tag is what a built image actually contains. That also makes the
+# check satisfiable inside a single dingo pull request: bump the pinned tag
+# and update the embedded copy together and it goes green, with no window
+# where it is red waiting on another repository to merge. Comparing against
+# main instead would go green the moment that branch moved, while the image
+# this repository builds still shipped the old config.
+#
+# Sources, in precedence order:
+#   CARDANO_CONFIGS_DIR  an existing checkout or extracted /config tree
+#   CARDANO_CONFIGS_REF  a git ref of the repository, cloned here
```
arepala-uml: This documents `CARDANO_CONFIGS_REF` as "a git ref," but it's passed to `git clone --branch`, which only resolves branches and tags. Tested with a 40-char commit SHA: `git clone --branch <sha> ...` fails with `fatal: Remote branch <sha> not found in upstream origin` instead of checking out that commit. Consider narrowing the doc to "a branch or tag name" or supporting arbitrary refs (clone then `git checkout "${configs_ref}"`).

bin/config-parity.sh
```diff
@@ -0,0 +1,146 @@
...
+# Dockerfile pins, not against that repository's main branch, because the
+# pinned tag is what a built image actually contains. That also makes the
+# check satisfiable inside a single dingo pull request: bump the pinned tag
+# and update the embedded copy together and it goes green, with no window
+# where it is red waiting on another repository to merge. Comparing against
+# main instead would go green the moment that branch moved, while the image
+# this repository builds still shipped the old config.
+#
+# Sources, in precedence order:
+#   CARDANO_CONFIGS_DIR  an existing checkout or extracted /config tree
```
arepala-uml: This says `CARDANO_CONFIGS_DIR` accepts "an existing checkout or extracted /config tree," but the code (line 88) always requires a directory that *contains* a `config/` subfolder - one level above a literally extracted `/config` tree. Tested: `docker cp <container>:/config ./extracted` then `CARDANO_CONFIGS_DIR=./extracted` fails with "has no config/ directory" even though `./extracted` holds exactly `preview/`, `mainnet/`, etc. as the comment describes. Either fix the doc wording or accept both layouts.

.github/workflows/config-parity.yml
```diff
@@ -0,0 +1,54 @@
...
+
+# The network configs under config/cardano/ exist twice: embedded in the
+# binary here (config/cardano/embed.go), and in
+# blinklabs-io/docker-cardano-configs, which the release image ships as
+# /opt/cardano/config/ (see the cardano-configs stage in the Dockerfile) and
+# which downstream tooling copies out of the image. CARDANO_NETWORK=<net>
+# resolves to one or the other depending on how dingo was deployed, so the
+# same network name means the same chain only while the two stay identical.
+on:
+  pull_request:
```
arepala-uml, line 11: The merge-gating `config-parity` job has no `paths:` filter, so it runs the full docker pull/create/cp + diff pass on every PR regardless of whether `config/cardano/**`, `Dockerfile`, or `bin/config-parity.sh` changed. `ci-docker.yml` already scopes its Dockerfile-triggered job with `paths:` - worth following that precedent so unrelated PRs aren't coupled to registry availability for merging.

bin/config-parity.sh
```diff
@@ -0,0 +1,146 @@
...
+fi
+
+if [[ ! -d "${configs_dir}/config" ]]; then
+	echo "error: ${configs_dir} has no config/ directory" >&2
+	exit 1
+fi
+
+# Check exactly the networks config/cardano/embed.go embeds, so adding a
+# network to that directive brings it under this check automatically.
+embed_line="$(grep -m1 '^//go:embed ' config/cardano/embed.go)"
```
arepala-uml: Minor: under `set -euo pipefail`, if this `grep -m1` finds no match, the assignment itself aborts the script before the `-eq 0` check on the next lines ever runs. The friendly "no networks found" error at line 98 is effectively dead code - a renamed/reformatted `//go:embed` directive would instead produce a bare, unexplained failure here. Consider `embed_line="$(grep -m1 ... )" || { echo "error: ..." >&2; exit 1; }`.

bin/config-parity.sh
```diff
@@ -0,0 +1,146 @@
...
+	# shellcheck disable=SC2064 # expand cleanup_dir now, not at trap time
+	trap "rm -rf '${cleanup_dir}'" EXIT
+	if [[ -n "${configs_ref}" ]]; then
+		source_desc="${configs_repo} at ${configs_ref}"
+		echo "cloning ${source_desc}"
+		git clone --quiet --depth 1 --branch "${configs_ref}" \
+			"${configs_repo}" "${cleanup_dir}/configs"
+		configs_dir="${cleanup_dir}/configs"
+	else
+		tag="$(sed -n "s|^FROM ${configs_image}:\\([^ ]*\\) .*|\\1|p" Dockerfile | head -1)"
```
arepala-uml, line 86: Minor/non-blocking: this reimplements Dockerfile `FROM ghcr.io/blinklabs-io/<image>:<tag>` parsing that `internal/docsparity`'s `goImageRe` already does, with different (stricter) tolerances - this one requires trailing text after the tag (only true today because the line ends in `AS cardano-configs`). Not asking for a change here, just flagging that the two parsers can silently diverge in what they accept if the Dockerfile format ever shifts.

arepala-uml (changes requested):
Verified by building `bin/config-parity.sh` on this branch and running it against the live pinned `cardano-configs:20260829-1` image, with fault injection on every path (default image extraction, `CARDANO_CONFIGS_DIR`, `CARDANO_CONFIGS_REF`). The core design choice - gating on the Dockerfile's pinned tag rather than upstream `main` - is sound and well-reasoned.

Requesting changes on one correctness gap in what the check actually guarantees; everything else below is non-blocking.

.github/workflows/config-parity.yml
```diff
@@ -0,0 +1,56 @@
...
+  # Does not gate merges. The job above pins to a released image, so it stays
+  # green while docker-cardano-configs moves ahead of that pin; this reports
+  # when it has, meaning a tag bump is due. Drift introduced in that
+  # repository produces no event here, hence the schedule.
+  config-upstream:
+    name: config-upstream
+    if: github.event_name != 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1 https://github.com/actions/checkout/releases/tag/v7.0.1
```
arepala-uml, line 52: `config-upstream`'s checkout step is missing `persist-credentials: false`. This is the job that actually clones a third-party repo (`blinklabs-io/docker-cardano-configs`, via `bin/config-parity.sh`'s `CARDANO_CONFIGS_REF` path) into the workspace, so `GITHUB_TOKEN` stays persisted in local git config during that clone. The sibling `config-parity` job right above already has this fix (from `a7505f16`, "ci: disable checkout credential persistence"); it just wasn't applied here.

Regression test proving it, added under `internal/docsparity` alongside the existing workflow-parity checks (e.g. `lint_coverage_test.go`):

```diff
--- a/internal/docsparity/lint_coverage_test.go
+++ b/internal/docsparity/lint_coverage_test.go
@@ -39,7 +39,8 @@ type workflowStep struct {
 	Uses string            `yaml:"uses"`
 	Env  map[string]string `yaml:"env"`
 	With struct {
-		WorkingDirectory string `yaml:"working-directory"`
+		WorkingDirectory   string `yaml:"working-directory"`
+		PersistCredentials *bool  `yaml:"persist-credentials"`
 	} `yaml:"with"`
 }
```

```go
// internal/docsparity/checkout_credentials_test.go (new file)
package docsparity_test

import (
	"strings"
	"testing"

	"gopkg.in/yaml.v3"
)

const configParityWorkflow = ".github/workflows/config-parity.yml"

// Jobs that clone a third-party repo into the workspace and so must disable
// persisted checkout credentials.
var thirdPartyCloneJobs = []string{"config-upstream"}

func TestConfigUpstreamDisablesPersistedCredentials(t *testing.T) {
	root := repoRoot(t)

	raw := readRepoFile(t, root, configParityWorkflow)
	var parsed actionsWorkflow
	if err := yaml.Unmarshal([]byte(raw), &parsed); err != nil {
		t.Fatalf("parse %s: %v", configParityWorkflow, err)
	}

	for _, name := range thirdPartyCloneJobs {
		job, ok := parsed.Jobs[name]
		if !ok {
			t.Errorf("%s has no %s job", configParityWorkflow, name)
			continue
		}

		var checkout *workflowStep
		for i := range job.Steps {
			action, _, _ := strings.Cut(job.Steps[i].Uses, "@")
			if action == "actions/checkout" {
				checkout = &job.Steps[i]
				break
			}
		}
		if checkout == nil {
			t.Errorf("%s job %s has no actions/checkout step", configParityWorkflow, name)
			continue
		}

		persisted := checkout.With.PersistCredentials == nil || *checkout.With.PersistCredentials
		if persisted {
			t.Errorf(
				"%s job %s's checkout step does not set persist-credentials: false",
				configParityWorkflow, name,
			)
		}
	}
}
```

Run against this PR's head (`a7505f16`) as-is:

```
=== RUN   TestConfigUpstreamDisablesPersistedCredentials
    checkout_credentials_test.go:81: .github/workflows/config-parity.yml job config-upstream's checkout step does not set persist-credentials: false; ...
--- FAIL: TestConfigUpstreamDisablesPersistedCredentials (0.00s)
FAIL
```

Adding `with: { persist-credentials: false }` to this step (mirroring line 33-34 above) makes it pass, and the rest of `internal/docsparity` stays green. Suggest folding this test in and applying that one-line fix here.

wolf31o2:
Review follow-up:

- The portable upstream file listing uses `find . -type f -print`, avoiding GNU-only `find -printf`.
- The pull-request checkout uses `persist-credentials: false`.
- Integrated current `main` in merge commit `5f2d0096`.

Please re-review the current head.

bin/config-parity.sh
arepala-uml: Already addressed at this head. The script now walks the upstream tree too (lines 149-167: `find . -type f -print` over the upstream network dir, then `git ls-files --error-unmatch` to check dingo has each one), reporting `DRIFT <network>/<relative>: missing from dingo`. Verified by fault injection: added an untracked-upstream file to `config/cardano/mainnet/` and re-ran `./bin/config-parity.sh` against the live pinned image — it correctly reported DRIFT and exited 1.

CLAUDE.md
arepala-uml: Already addressed. AGENTS.md's Pre-commit list now includes `make config-parity` alongside CLAUDE.md's.

bin/config-parity.sh
arepala-uml: Already addressed. The script no longer does `git clone --branch <ref>`; it clones, then `git fetch origin "${configs_ref}"` and `checkout FETCH_HEAD`, which resolves branches, tags, and SHAs.

bin/config-parity.sh
arepala-uml: Already addressed. The script now checks `-d "${configs_dir}/config"` and falls back to treating `configs_dir` itself as the config root, so both a checkout root and a directly-extracted `/config` tree work.

.github/workflows/config-parity.yml
arepala-uml: Acknowledged, left as-is. Current head still has no `paths:` filter on `config-parity`'s `pull_request` trigger — runs on every PR. Per the sibling cubic finding on this same thread, path-filtering a *required* check risks leaving it permanently skipped/pending on non-matching PRs, which blocks merge. Given the job is cheap (~9s in this PR's own CI run) and gates chain-config correctness, no filter is the safer tradeoff. Non-blocking.

bin/config-parity.sh
arepala-uml: Already addressed. `embed_line="$(grep -m1 ...)"` now has an explicit `|| { echo "error: ..."; exit 1; }` fallback, so a missing `//go:embed` directive produces the intended message instead of a bare `set -e` abort.

bin/config-parity.sh
arepala-uml: Confirmed still present and, as originally noted, non-blocking — this script's Dockerfile-tag parser and `internal/docsparity`'s `goImageRe` remain two independent parsers with different tolerances. No change requested; worth a follow-up issue if the author wants to dedupe them later.

.github/workflows/config-parity.yml
```diff
@@ -0,0 +1,59 @@
...
+# binary here (config/cardano/embed.go), and in
+# blinklabs-io/docker-cardano-configs, which the release image ships as
+# /opt/cardano/config/ (see the cardano-configs stage in the Dockerfile) and
+# which downstream tooling copies out of the image. CARDANO_NETWORK=<net>
+# resolves to one or the other depending on how dingo was deployed, so the
+# same network name means the same chain only while the two stay identical.
+on:
+  pull_request:
+    paths:
+      - "config/cardano/**"
```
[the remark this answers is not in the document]
arepala-uml: Consistent with the human finding on the adjacent thread — current head deliberately keeps no `paths:` filter on this required check, which avoids exactly the skipped-required-check risk this comment flags. No action needed.

Makefile
```diff
@@ -96,6 +96,9 @@ import-boundaries: ## Check reviewed package import boundaries
 docs-parity: ## Check docs against go.mod, the Makefile, and the DevNet compose file
 	go test ./internal/docsparity
 
+config-parity: ## Fail if the embedded network configs drift from docker-cardano-configs
+	./bin/config-parity.sh
```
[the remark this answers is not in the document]
arepala-uml: Already addressed. Current script uses `find . -type f -print` (line 166), not `-printf` — confirmed portable: ran it on this macOS checkout against the live pinned image and it completed the upstream-only pass correctly (54 files / 5 networks matched).

.github/workflows/config-parity.yml
```diff
@@ -0,0 +1,59 @@
...
+      - "Dockerfile"
+      - "bin/config-parity.sh"
+      - ".github/workflows/config-parity.yml"
```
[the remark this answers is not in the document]
arepala-uml: Moot at this head: the workflow has no `paths:` filter at all (runs unconditionally on every `pull_request`), so Makefile changes already trigger it along with everything else.

.github/workflows/config-parity.yml
```diff
@@ -0,0 +1,54 @@
...
+  # config change is satisfiable inside one pull request here: bump the
+  # pinned tag and update config/cardano/ in the same commit. Comparing
+  # against the configs repository's main branch instead would pass as soon
+  # as that branch moved, while the image this repository builds still
+  # shipped the old config.
+  config-parity:
+    name: config-parity
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1 https://github.com/actions/checkout/releases/tag/v7.0.1
```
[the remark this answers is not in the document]
arepala-uml: Already addressed. The `config-parity` job's checkout step sets `persist-credentials: false` (line 37 at this head).

.github/workflows/config-parity.yml
arepala-uml: Already addressed. The `config-upstream` job's checkout step now also sets `persist-credentials: false` (line 54 at this head) — matches the sibling job.

arepala-uml (commented):
Re-reviewed at head cb6edd72a862. All prior findings (mine, CodeRabbit's, Cubic's) are already addressed at this head — replied on each of the 12 threads with what changed. Verified directly, not just read:

- Ran `./bin/config-parity.sh` against the live pinned `ghcr.io/blinklabs-io/cardano-configs:20260829-1` image: passes clean (54 files / 5 networks matched).
- Fault-injected three drift scenarios and confirmed each fails correctly: a byte changed in a tracked config file, a tracked file removed from dingo, and an untracked extra file added to dingo (bidirectional check both ways). All three produced exit 1 with an actionable message; baseline is clean afterward.
- `make config-parity` wiring, `.PHONY`, and the AGENTS.md/CLAUDE.md doc updates match the script.
- Workflow follows repo convention: pinned `actions/checkout` SHA, `persist-credentials: false` on both jobs, `contents: read`, concurrency cancel-in-progress. `config-upstream` correctly does not gate merges.
- One open non-blocking item: no `paths:` filter on the required `config-parity` job's `pull_request` trigger, so it runs on every PR. Deliberate tradeoff given the required-check-skipped risk a filter would introduce (per the Cubic thread) — the job is cheap (~9s) and correctness-critical, so I agree with leaving it unfiltered.

Withholding approval only for the pipeline: **go-test (macOS) is failing on this head** (`TestDecodeReadChainBatchDoesNotDeadlockOnManyValidationErrors`, ledger package — "Should be zero, but was 245" pipeline-error-drain count, plus unrelated panic/cross-fork-splice noise in the same run). This is not caused by this PR — the diff touches no Go source (`.github/workflows/config-parity.yml`, `AGENTS.md`, `CLAUDE.md`, `Makefile`, `bin/config-parity.sh` only). Confirmed pre-existing: `go-test (macOS)` has failed on `origin/main` itself repeatedly in the last day (runs at 680e65e5, 1dd68527, d07cca21, 64e2a00f — each a different flaky ledger/chain-selection test, none touching config parity). CI already runs against the PR's merge ref, so this reflects the current merged tree, not a stale base.

Approve once `go-test (macOS)` is green (rerun, or land the flake fix on main and rebase/merge). No code changes needed on this PR from me.

PR #3641, opened 2026-08-29
fix(leios): validate fetched endorser-block txs by full-transaction hash, not body hash

chadaibbhot-git:
## Problem
`c9692061` ("fix(leios): bind fetched transactions to manifests", in v0.70.2 and
main) validates each fetched endorser-block transaction against the manifest by
hashing **only the transaction body** (`txElems[0]`) and comparing to
`ref.TransactionHash`:

```go
if bodyHash := lcommon.Blake2b256Hash(txElems[0]); bodyHash != ref.TransactionHash {
    return fmt.Errorf("endorser tx %d body hash mismatch", index)
}
```

But the same manifest reference carries `TransactionSize == len(txCbor)` — the
size of the **full** serialized transaction. The manifest is a content-addressed
(hash+size) reference to the full tx, so `TransactionHash` is the hash of the
full tx, not the Cardano tx-id / body hash. Against a reference (Haskell) node's
endorser blocks, EVERY fetched tx fails this check ("endorser tx 0 body hash
mismatch", 0/N accepted), so no endorser block can be backfilled and a catching-up
node wedges permanently on the first EB-bearing block it needs.

The existing unit test did not catch this because its fixture builds the
reference hash the same (body-only) way the validator checks it —
`TransactionHash: lcommon.Blake2b256Hash(txElems[0])` — so it is self-consistent
with the bug.

## Evidence
A node behind a Haskell reference relay: the relay serves the EB completely
(`MsgLeiosBlockTxsRequest [start]→[done]`, bitmap echoed, all txs), but the Dingo
client logs `tx fetch (0/N): validate fetched transaction: endorser tx 0 body
hash mismatch` for every EB (474, 218, 2098, 115 txs …). Deterministic; the node
logged the wedge 151 times over 75 minutes.

## Fix
Hash the full transaction bytes, consistent with `TransactionSize`:
```go
if txHash := lcommon.Blake2b256Hash(txCbor); txHash != ref.TransactionHash {
    return fmt.Errorf("endorser tx %d hash mismatch", index)
}
```
and update the test fixture to build `TransactionHash` from `txCbor`.

## Verified
Built and deployed to a wedged BP behind a Haskell reference relay: the
`... body hash mismatch` errors went from constant (151/75min) to **zero**, and
the node began applying endorser-block transactions past the previously-stuck EB
(slot 1863115) — confirming the fetched txs are now accepted. (A separate,
downstream ledger issue — "UTxO already spent" applying an EB tx at a later slot —
is NOT addressed here and is filed separately.)

---
## Summary by cubic
Fixes endorser-block transaction hashing so both fetched and locally forged EBs validate against the manifest, and a catching-up node no longer wedges on EB-bearing blocks.

- The fetch-side validator now hashes the full serialized transaction CBOR instead of only the body, matching the manifest's `TransactionHash`/`TransactionSize` contract.
- `buildLeiosEB` now sets `TransactionHash` to the full tx CBOR hash instead of the Cardano tx-id/body hash, so peers can fetch locally forged EBs.
- Adds a regression test asserting forged references hash the full tx CBOR.

<sup>Written for commit c245595362755521ca50665ec2c208ced4480981. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3641?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation of Leios endorser transactions by checking the hash of the complete transaction data.
  - Fixed validation behavior for wrapped transactions, reducing the risk of valid transactions being incorrectly rejected.
  - Updated validation error messaging to clearly indicate when a transaction hash does not match its reference.

### DIFF
--- a/ouroboros/leios_merged.go
+++ b/ouroboros/leios_merged.go
@@ -286,10 +286,14 @@ func cloneRawMessages(in []cbor.RawMessage) []cbor.RawMessage {
 }
 
 // validateLeiosEndorserBlockTxs binds fetched transaction wire values to the
-// manifest before the complete set is cached or applied. Leios transaction
-// references use the Cardano transaction ID (the hash of the transaction body)
-// and the complete transaction's encoded size, while leios-fetch carries each
-// transaction either directly or in a CBOR byte-string wrapper.
+// manifest before the complete set is cached or applied. Each Leios manifest
+// reference is content-addressed over the FULL serialized transaction: the
+// TransactionHash is Blake2b256 of the complete transaction CBOR (NOT the
+// Cardano tx-id / body hash) and TransactionSize is that CBOR's length. Both
+// the fetch-side validator here and the forge-side reference builder
+// (buildLeiosEB) use this same full-transaction contract, matching Haskell
+// reference nodes. leios-fetch carries each transaction either directly or in
+// a CBOR byte-string wrapper.
 func validateLeiosEndorserBlockTxs(
 	manifestRaw []byte,
 	txsRaw []cbor.RawMessage,
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -1511,7 +1511,7 @@ func buildLeiosEB(
 		// (validateLeiosEndorserBlockTxs) and Haskell reference nodes, so a
 		// peer fetching a locally forged EB validates it instead of rejecting
 		// every tx (blinklabs-io/dingo#3641).
-		if _, ok := validLeiosTransactionHash(tx.Hash); !ok ||
+		if !validLeiosTransactionHash(tx.Hash) ||
 			len(tx.Cbor) == 0 || len(tx.Cbor) > math.MaxUint16 {
 			continue
 		}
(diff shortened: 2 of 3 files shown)


ouroboros/leios_merged.go
```diff
@@ -356,8 +356,8 @@ func validateLeiosEndorserBlockTx(
 			ref.TransactionSize,
 		)
 	}
-	if bodyHash := lcommon.Blake2b256Hash(txElems[0]); bodyHash != ref.TransactionHash {
-		return fmt.Errorf("endorser tx %d body hash mismatch", index)
+	if txHash := lcommon.Blake2b256Hash(txCbor); txHash != ref.TransactionHash {
```
wolf31o2, line 366: **P1 — Keep forged and fetched transaction references on one hash contract**

This validator now compares `Blake2b256(txCbor)`, but `ledger/forging/forger.go:1234-1240` builds local references by hex-decoding `tx.Hash` (the 32-byte Cardano transaction ID) and hashing those bytes again. The producer therefore advertises `Blake2b256(txID)` while this consumer expects `Blake2b256(full transaction CBOR)`, so peers reject locally forged endorser blocks during fetch. Update the producer/reference construction and cover a locally produced EB through this validator. The comment at lines 288-292 must also describe the selected hash contract.

wolf31o2 (changes requested):

ouroboros/leios_merged.go
chadaibbhot-git: Fixed in `84cd596f`. `buildLeiosEB` (forger.go) now sets `TransactionHash: Blake2b256Hash(tx.Cbor)` — the hash of the full serialized transaction, consistent with `TransactionSize = len(tx.Cbor)` — so forged EBs advertise the same content-addressed (full-tx hash, full-tx size) references the fetch-side validator checks and Haskell reference nodes use. Updated the `validateLeiosEndorserBlockTxs` doc comment to describe the full-transaction contract, and added `TestBuildLeiosEBReferencesUseFullTransactionHash` asserting each forged reference hash equals `Blake2b256(full tx CBOR)` (with a regression guard that it is no longer the tx-id/body hash). `ledger/forging` + `ouroboros` tests pass.

wolf31o2 (changes requested):
Failing CI

wolf31o2 (approved):

PR #3649, opened 2026-08-30
fix(ledger): deduplicate certified Leios transactions

### LINKED ISSUE #3643: Ranking block re-applies an endorser-resident transaction on the Haskell-conformant (Musashi) path → "UTxO already spent"
On the Musashi Haskell-conformant Leios path, a catching-up node wedges applying
a ranking block whose transactions overlap the certified endorser block it
references:

```
block processing failed, restarting pipeline
error="process block batch: apply endorser block transactions: record transaction:
       set transaction metadata for tx 6cf9bcfc6fafed9f340e79c63f3e2f1bacb380ae3bc35e3ff725a69bd449ffc6
       (block idx 691, slot 1864040): UTxO already spent:
       3fe3ab01a255960d22d63e7ad8fabdee8cd7d9d884790105c35fa3bea1b53e01:0"
```

The referenced endorser block's transactions are applied ahead of the ranking
block (`applyEndorserBlock`, ledger/leios_apply.go). Per the code's own comment,
the Haskell-conformant path (Musashi prototype-2026w29) applies the certified
EB's transactions "with their full effects — produced outputs, consumed inputs,
certificates, and governance — but WITHOUT validation or consumed-input
recovery" (leios_apply.go:225-227). Endorser tx 6cf9bcfc consumes
3fe3ab01…:0 when the EB is applied. The ranking block at slot 1864040 then
applies the SAME transaction again at block idx 691, which re-consumes the
already-spent input → `ErrUtxoConflict` (database/types/types.go:266).

`deduplicateEndorserBlockTransactionIndexes` (leios_apply.go) dedups EB-side
transactions against the DB, but the ranking-block apply path does not skip
transactions already applied via the referenced EB, so an EB-resident tx that
also appears in the ranking block is applied twice.

A Haskell reference node on the same network applies this block cleanly and is
at tip — so the block is valid and each transaction must be applied exactly
once across (EB ∪ ranking block).

**Deterministic:** 100% of pipeline restarts on the affected node reference
slot 1864040 / tx 6cf9bcfc after the endorser-block fetch succeeds. It is the
second of two head-of-line blockers for a catching-up node; the first (fetched
EB txs validated by body hash instead of full-tx hash) is fixed in PR #3641.

## Suggested direction (for maintainers)
On the Haskell-conformant path, the ranking-block application should skip
transactions already applied via its certified endorser block (dedup by tx hash,
symmetric to the EB-side dedup), so each transaction in (EB ∪ RB) is applied
once. This is consensus-critical certificate-driven apply code, so the exact
placement is left to maintainers rather than proposed as a blind patch.

wolf31o2:
Fixes #3643

The Musashi path now tracks certified endorser-block transactions by Cardano body hash and skips matching ranking-block transactions after successful application. Non-overlapping ranking transactions retain their order. Focused race, ledger, conformance, lint, vet, boundary, documentation, formatting, and diff checks pass.

---
## Summary by cubic
Fixes #3643 by deduplicating certified endorser-block transactions that also appear in the certifying ranking block.  
The Musashi path now skips ranking-block transactions that were already applied as part of the certified endorser block. Non-overlapping ranking transactions keep their original order and indices. The CIP announcement-driven path is unaffected.

<sup>Written for commit 3c8e070a4009df5ad244f82bcac53775e06598cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3649?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected transaction processing for certified endorser blocks to prevent shared transactions from being applied more than once.
  * Preserved accurate block indexing for distinct transactions included in ranking blocks.
  * Maintained existing two-stage transaction behavior for CIP-conformant processing.
  * Added coverage for transaction ordering and ledger effects across supported processing paths.

### DIFF
--- a/ledger/leios_apply.go
+++ b/ledger/leios_apply.go
@@ -97,6 +97,26 @@ func (ls *LedgerState) applyEndorserBlock(
 	ebHashBytes []byte,
 	rawTxs []cbor.RawMessage,
 ) (int, uint64, error) {
+	return ls.applyEndorserBlockTrackingTransactions(
+		txn,
+		rbPoint,
+		rbBlockNumber,
+		ebSlot,
+		ebHashBytes,
+		rawTxs,
+		nil,
+	)
+}
+
+func (ls *LedgerState) applyEndorserBlockTrackingTransactions(
+	txn *database.Txn,
+	rbPoint ocommon.Point,
+	rbBlockNumber uint64,
+	ebSlot uint64,
+	ebHashBytes []byte,
+	rawTxs []cbor.RawMessage,
+	certifiedTxIDs map[lcommon.Blake2b256]struct{},
+) (int, uint64, error) {
 	if len(rawTxs) == 0 {
 		return 0, 0, nil
 	}
@@ -247,6 +267,11 @@ func (ls *LedgerState) applyEndorserBlock(
 					"apply endorser block transactions: %w",
 					err,
 				),
+			}
+		}
+		if certifiedTxIDs != nil {
+			for _, transaction := range txs {
+				certifiedTxIDs[transaction.Hash()] = struct{}{}
 			}
 		}
 		return len(delta.Transactions), delta.donation, nil
--- a/ledger/leios_apply_test.go
+++ b/ledger/leios_apply_test.go
@@ -20,10 +20,14 @@ import (
 	"io"
 	"log/slog"
 	"testing"
+	"time"
 
 	"github.com/blinklabs-io/dingo/database"
 	"github.com/blinklabs-io/dingo/database/models"
+	"github.com/blinklabs-io/dingo/event"
 	dbtest "github.com/blinklabs-io/dingo/internal/test/dbtest"
+	"github.com/blinklabs-io/dingo/internal/test/testutil"
+	"github.com/blinklabs-io/dingo/ledger/eras"
 	"github.com/blinklabs-io/gouroboros/cbor"
 	gledger "github.com/blinklabs-io/gouroboros/ledger"
 	"github.com/blinklabs-io/gouroboros/ledger/babbage"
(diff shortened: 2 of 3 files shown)


ledger/leios_apply.go
```diff
@@ -249,6 +269,11 @@ func (ls *LedgerState) applyEndorserBlock(
 				),
 			}
 		}
+		if certifiedTxIDs != nil {
+			for _, transaction := range txs {
```
chrisguiney, line 273: The tracking set is built from every decoded transaction, but on this path the delta is built only from `keepIndexes` (line 232-234). Transactions that `deduplicateEndorserBlockTransactionIndexes` dropped — because they're already committed from an earlier block — get recorded here as certified-and-applied even though this closure never applied them.

That contradicts the contract stated in `state.go`: "Skip only transactions present in that successfully processed certified closure." The ranking block would then skip a transaction the endorser block deliberately excluded. CodeRabbit flagged the same shape as its Moderate merge risk.

If the guard stays, iterating `keepIndexes` would match the delta and the comment:

```go
for _, idx := range keepIndexes {
    certifiedTxIDs[txs[idx].Hash()] = struct{}{}
}
```

ledger/state.go
```diff
@@ -6074,15 +6078,31 @@ func (ls *LedgerState) ledgerProcessBlock(
...
-						block.BlockNumber(),
-						ebSlot,
-						ebHash.Bytes(),
-						ebTxs,
+					var (
+						applied  int
+						donation uint64
+						err      error
 					)
+					if certifiedEndorserTxIDs == nil {
```
chrisguiney, line 6086: Small simplification: `applyEndorserBlock` now forwards to `applyEndorserBlockTrackingTransactions` with a literal `nil`, and this branch calls it only when `certifiedEndorserTxIDs` is already `nil` — so both arms are the same call with the same argument. In production this is the only caller of the wrapper; the ~20 lines here collapse to:

```go
applied, donation, err := ls.applyEndorserBlockTrackingTransactions(
    txn, point, block.BlockNumber(), ebSlot, ebHash.Bytes(), ebTxs,
    certifiedEndorserTxIDs,
)
```

Worth keeping the `applyEndorserBlock` wrapper for the existing test call sites, but the `var (...)` block and the if/else don't earn their keep.

ledger/state.go
```diff
@@ -6169,6 +6189,16 @@ func (ls *LedgerState) ledgerProcessBlock(
...
 	intraBlockUtxos := make(map[string]lcommon.Utxo)
 	for i, tx := range block.Transactions() {
+		// The Musashi CertRB body can repeat a transaction from the certified
+		// parent endorser block that was applied immediately above. Cardano's
+		// transaction identity is the transaction-body hash returned by Hash;
+		// the Leios full-envelope hash also includes witnesses and is not the
+		// identity used by UTxOs or transaction metadata. Skip only transactions
+		// present in that successfully processed certified closure. The CIP path
+		// leaves certifiedEndorserTxIDs nil and retains its existing behavior.
+		if _, alreadyApplied := certifiedEndorserTxIDs[tx.Hash()]; alreadyApplied {
```
chrisguiney, line 6199: Two small things on this guard.

**Cost when the set is empty.** `tx.Hash()` is evaluated for every ranking-block transaction on the Musashi path, including the common case where no endorser block was referenced at all. `DijkstraTransaction.Hash()` has a value receiver, so the memo that `TransactionBodyBase.Id()` writes lands in a copy and is discarded — I measured 176 ns/op and 3 allocs/op, recomputed on every call. Roughly 18 µs per 100-tx block; genuinely minor, but it buys nothing when the set is empty. A `len(certifiedEndorserTxIDs) > 0` guard (or hoisting the check above the loop) avoids it.

**Observability.** The skip is a bare `continue` with no log line or metric. Given the overlap is documented as not occurring on the conformant path, hitting this branch is a signal worth surfacing — at minimum a `Warn` — rather than absorbing silently.

ledger/state.go
```diff
@@ -6039,7 +6039,11 @@ func (ls *LedgerState) ledgerProcessBlock(
 	// resolution, availability, decode, and apply failures abort the block.
 	// Storage-phase failures always abort the DB transaction so a partial
 	// endorser-block application cannot be committed.
+	var certifiedEndorserTxIDs map[lcommon.Blake2b256]struct{}
 	if currentEra.Id == dijkstra.EraIdDijkstra {
+		if !ls.config.LeiosApplyEndorserBlockTxs {
+			certifiedEndorserTxIDs = make(map[lcommon.Blake2b256]struct{})
```
chrisguiney, line 6045: Nit: the map is allocated for every Dijkstra block on the Musashi path, including blocks where `EndorserBlockProvider` is nil or no endorser block is referenced, so it's never written. Allocating lazily at the point of the `applyEndorserBlockTrackingTransactions` call would keep the common path allocation-free. Very minor — noting it only because this is the block-apply hot path during sync.

ledger/leios_apply_test.go
```diff
@@ -331,6 +335,417 @@ func leiosApplyTestTxWithOutput(
...
+			leiosApplyTestProcessBlock(t, ls, db, ranking, rankingCbor),
+			"a transaction shared by the certified endorser block and ranking block must not be applied twice",
+		)
+		require.Equal(t, 1, providerCalls)
+		var overlapIndex uint32
+		require.NoError(t, rawDB.QueryRow(
+			`SELECT block_index FROM "transaction" WHERE hash = ?`,
+			overlapTx.Hash().Bytes(),
+		).Scan(&overlapIndex))
+		require.Equal(t, uint32(0), overlapIndex)
```
chrisguiney, line 674: These ledger-effect assertions don't discriminate between the two behaviours. I reverted the `leios_apply.go` and `state.go` hunks on a merge with current `origin/main` and re-ran: `overlapIndex == 0`, `distinctIndex == 1`, and the tx count of 4 all still pass without the fix. `overlapIndex` is 0 either way because the shared transaction sits at index 0 in both the endorser block and the ranking block, and the re-apply upserts on hash so the count is unchanged.

The only assertion that actually fails without the change is the `secondEvent` check at line 685. So the subtest named "applies exactly once" currently proves that a second `TransactionEvent` isn't published, not that ledger effects happened once — which is the claim the comment above makes.

A fixture where the endorser-block index differs from the ranking-block index, plus an assertion on consumed/produced UTxO state, would make this test carry the weight of its name.

ledger/leios_apply_test.go
```diff
@@ -331,6 +335,417 @@ func leiosApplyTestTxWithOutput(
 	return cbor.RawMessage(txCbor), tx
 }
 
+func leiosApplyTestTxSpending(
```
chrisguiney, line 338: Heads-up on a rebase collision rather than anything wrong here: #3773 landed `leiosApplyTestSpendingTx` + `leiosApplyTestTxFromBody` in `leios_apply_double_consume_test.go`, which do the same job as this helper. After a rebase the package would carry both, under names that are transpositions of each other (`TxSpending` vs `SpendingTx`) — easy to reach for the wrong one later. Reusing the existing pair would avoid that.

chrisguiney (changes requested):
Thanks for putting this together — the change is clean and the test harness is thorough. I'd like to hold off on approving, though, because I think the premise no longer holds.

**`Fixes #3643` no longer applies.** That issue is closed, fixed by `f8f4414c` (#3680), and the closing comment there notes that the reference CertRB body is empty, so (EB ∩ RB) is empty and RB-side dedup does not apply. I checked that independently against this repo's docs rather than taking it on trust:

- `ARCHITECTURE.md:3334` — a CertRB "carries a Leios certificate and empty transaction segments; the certified endorser block's (EB) transactions live in the EB's transaction closure".
- `ARCHITECTURE.md:4020` — before constructing a new EB the forger "reads the certified EB's manifest and filters those transaction hashes from its mempool view", so Dingo's own producer avoids the overlap too.

So the overlap is excluded by construction on both the receive and produce sides, and the defect that actually wedged catch-up (one input consumed by two *different* transactions across endorser blocks) isn't addressed by tx-hash dedup.

There's also a direction-of-failure concern worth weighing. Today an overlapping CertRB body is rejected loudly with `ErrUtxoConflict`. With this patch those transactions are silently skipped, which would diverge from a reference node that applies them — swapping a visible failure for a quiet one in consensus-critical apply code.

My suggestion is to close this in favour of #3680 + #3773, but if you'd still like the guard, I've left five inline notes — the first is a correctness one that CodeRabbit also flagged.

**Verification** (worktree at `origin/main` `8419482f` with `pull/3649/head` merged in; the branch is 57 commits behind, including #3680 and #3773, and CI here last ran 2026-08-30 against a stale base — the merge itself is clean):

| Check | Result |
| --- | --- |
| `go build ./...`, `go vet ./ledger/` | pass |
| `go test ./ledger/ -count=1` | pass, 28.2s |
| `go test ./ledger/ -race -run Leios` | pass, no races |
| `gofmt -l ledger/` | clean |
| New test, merged tree | pass |
| New test, production hunks reverted | fails at `leios_apply_test.go:685` only |

Concurrency looks fine — the tracking map is function-local, never escapes `ledgerProcessBlock`, and adds no goroutines or shared state.

Not run, for the record: live CI on the merged tree, devnet, conformance vectors, Antithesis, and a Preview/Koios parity replay. For a ledger-apply change that replay is the check that would really settle it, so the green unit tests are necessary but not sufficient.

wolf31o2:
Closing this superseded change in favor of the fixes already landed in #3680 and #3773, per review.

PR #3677, opened 2026-08-30
fix(ledger): bump plutigo so cost model tables match the live networks

### LINKED ISSUE #3653: ledger: plutigo v0.4.0 cost model tables misassign and drop on-chain parameters
go.mod pins plutigo v0.4.0. Its cost model parameter tables are a version behind what the live networks publish, and the mismatch is not a clean truncation: the tables disagree positionally near the end, so values are assigned to the wrong parameters before any are dropped.

## Counts

| | plutigo v0.4.0 | plutigo v0.5.0 | mainnet, preprod, preview at PV11 |
| --- | --- | --- | --- |
| `CostModelParamNamesV1` | 328 | 332 | 332 |
| `CostModelParamNamesV2` | 328 | 332 | 332 |
| `CostModelParamNamesV3` | 346 | 350 | 350 |

The three public networks publish byte identical cost models, checked against Koios `epoch_params` for mainnet epoch 652, preprod 310 and preview 1404. v0.5.0's tables match them exactly.

## What actually goes wrong

`cek.costModelFromList` walks `lang.GetParamNamesForVersion(version)` and assigns `data[i]` to the name at index `i`, breaking out when `i >= len(data)`. With a 332 entry list against a 328 entry name table it never errors: it assigns 328 values and silently ignores the last 4.

The divergence between the two tables is in `valueData` and `unValueData`, which changed shape rather than being appended to:

- `valueData` cpu went from a single `valueData-cpu-arguments` to `-intercept` plus `-slope`, and memory likewise
- `unValueData` cpu went from `-intercept`/`-slope` to `-c0`/`-c1`/`-c2`, and memory gained a `-slope`

That is 5 names replaced by 9. They sit at the tail of each table, from index 319 in V1 and V2 and index 337 in V3, so:

- the 9 on chain values for those two builtins land on v0.4.0's 5 differently shaped names, giving wrong costs
- the final 4 values fall past the end of the name table and are dropped

Every other parameter is at the same index in both tables and is unaffected. V1 additionally renames `blake2b` to `blake2b_256` and `verifySignature` to `verifyEd25519Signature` at indices 14 and following, but those are position preserving and cost nothing.

Because the cost function shape changed and not just the values, reordering cannot fix this. It needs the newer tables.

## Reach

`ledger/eras/{alonzo,babbage,conway}.go` pass the on chain `pp.CostModels[...]` straight into `cek.NewEvalContext` (for example `ledger/eras/conway.go:1081`), so this is the mainnet script evaluation path, not a devnet or test concern. A transaction whose scripts call `valueData` or `unValueData` is budgeted against wrong costs, which can disagree with the reference node in either direction: rejecting a valid transaction, or admitting one cardano-node rejects.

Storage and hashing are not affected. The full list is kept in the protocol parameters and the script data hash is computed over that, so only evaluation reads the truncated mapping.

## Fix

Bump plutigo to v0.5.0 or newer and confirm `len(lang.GetParamNamesForVersion(v))` equals the length of the cost models the public networks publish, for each of V1, V2 and V3. Worth a regression test asserting that equality so a future parameter addition upstream fails loudly here instead of being silently ignored.

Found while correcting the devnet genesis cost models in #3632, which is a separate defect in the shipped config rather than in the evaluator.

wolf31o2:
## Problem

plutigo v0.4.0's cost model parameter tables are 328/328/346 for PlutusV1/V2/V3. Mainnet, preprod and preview all publish 332/332/350 at protocol version 11, byte identical across the three (checked against Koios `epoch_params` for mainnet epoch 652, preprod 310, preview 1404).

The shortfall was not a clean truncation. `valueData` and `unValueData` changed shape rather than gaining appended entries:

- `valueData` cpu went from a single `valueData-cpu-arguments` to `-intercept` plus `-slope`, and memory likewise
- `unValueData` cpu went from `-intercept`/`-slope` to `-c0`/`-c1`/`-c2`, and memory gained a `-slope`

That replaces 5 names with 9 at the tail of each table, from index 319 in V1 and V2 and index 337 in V3. `cek.costModelFromList` maps the on-chain model onto the name table positionally and breaks out at `i >= len(data)`, so with a 332 entry model against a 328 entry table it assigned nine tail values to five differently shaped names, then silently dropped the last four. No error either way. Every other parameter is at the same index in both tables and was unaffected.

V1 also renames `blake2b` to `blake2b_256` and `verifySignature` to `verifyEd25519Signature` at index 14 and following. Those are position preserving and cost nothing.

## Reach

`ledger/eras/{alonzo,babbage,conway}.go` pass the on-chain `pp.CostModels[...]` straight into `cek.NewEvalContext` (for example `ledger/eras/conway.go:1081`), so this is the mainnet script evaluation path. A transaction whose scripts call `valueData` or `unValueData` was budgeted against wrong costs, which can disagree with cardano-node in either direction: rejecting a valid transaction, or admitting one the reference node rejects.

Storage and hashing were not affected. The full list is kept in the protocol parameters and the script data hash is computed over that, so only evaluation read the bad mapping.

## Change

`go.mod` moves plutigo v0.4.0 to v0.5.0, whose tables are 332/332/350 and match the published models exactly. `go mod tidy` produced no other module changes.

`ledger/eras/cost_models_test.go` adds two tests. `TestCostModelParameterTablesMatchLiveNetworks` pins the three lengths. `TestCostModelParameterTablesPriceValueDataBuiltins` asserts the nine `valueData`/`unValueData` parameter names are present, because a table of the right length still carrying the older single-argument forms would be the same defect. All six subtests fail against v0.4.0 and pass against v0.5.0.

## Validation

- `go test -tags dingo_extra_plugins ./...`: 82 packages, exit 0
- `internal/test/conformance`: pass
- `ledger/eras`: pass
- `golangci-lint run ./...`: 0 issues
- `make docs-parity`: pass
- `nilaway` and `modernize` each report one finding, both in files this change does not touch (`ledger/leios/manager_test.go:1281`, `ledger/state.go:6279`) and both reproduced on a main based checkout

DevNet was not run. A cost model change alters ExUnits accounting on the block production path, so `run-tests.sh --conformance` is worth running before merge.

`DATABASE.md` and `ARCHITECTURE.md` checked and not affected.

Fixes #3653

---
## Summary by cubic
Bumps `plutigo` from v0.4.0 to v0.5.0 so cost model tables match live networks. The old tables were 328/328/346 for PlutusV1/V2/V3 versus the published 332/332/350, so scripts using `valueData` or `unValueData` were priced against wrong ExUnits and could disagree with cardano-node.

**Details**
- `cek.costModelFromList` maps on-chain models positionally and silently dropped excess entries, so the shortfall never errored.
- The missing entries weren't appended: five parameter names changed shape into nine at each table tail.
- New tests pin the table lengths and the nine changed names; both fail against v0.4.0.

<sup>Written for commit 6fed80f123f416e250ec824bdd50612cb3ecef89. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3677?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/blinklabs-io/bursa v0.16.1-0.20260817233527-1eb8b64db609
 	github.com/blinklabs-io/gouroboros v0.202.2
 	github.com/blinklabs-io/ouroboros-mock v0.18.0
-	github.com/blinklabs-io/plutigo v0.4.0
+	github.com/blinklabs-io/plutigo v0.5.0
 	github.com/blockfrost/blockfrost-go v0.5.0
 	github.com/btcsuite/btcd/btcutil v1.2.0
 	github.com/consensys/gnark-crypto v0.21.0
--- a/go.sum
+++ b/go.sum
@@ -131,8 +131,8 @@ github.com/blinklabs-io/gouroboros v0.202.2 h1:rb21cE7YJwpbh/fEWxC5IpSig+zdDsuZU
 github.com/blinklabs-io/gouroboros v0.202.2/go.mod h1:y+bzoTL4JC1qu5m2NTERWuFl+fz6JPKBq22lQDIpjWs=
 github.com/blinklabs-io/ouroboros-mock v0.18.0 h1:m7f7jUhkxp6NlVF86BLxGDNaATjbhQXcWIFqJ4HFyh8=
 github.com/blinklabs-io/ouroboros-mock v0.18.0/go.mod h1:6YjKLLIaSTSrl8RJTg584f+W5NwniLNn/GHja2ijic0=
-github.com/blinklabs-io/plutigo v0.4.0 h1:DVh4LtCq4Hc7Rfc5jSGD2cLK7610y8ai9F7vwW9r620=
-github.com/blinklabs-io/plutigo v0.4.0/go.mod h1:ceskOZD5ZkkxhctkfvzRe3XEYdOzFYoxp11RqiG6oeU=
+github.com/blinklabs-io/plutigo v0.5.0 h1:qDb6ADY5f0LWfQBdr3tZTYGqJGgFdmDQI/iyjSz8CJs=
+github.com/blinklabs-io/plutigo v0.5.0/go.mod h1:ceskOZD5ZkkxhctkfvzRe3XEYdOzFYoxp11RqiG6oeU=
 github.com/blockfrost/blockfrost-go v0.5.0 h1:I+qEQw5Re9UHolKZGJE0FtuQL3a7L0WqEYkEO4GisIw=
 github.com/blockfrost/blockfrost-go v0.5.0/go.mod h1:XdD+mryM/Rd/MqW1MfSQ0+Xfu2YnOGuwqMpLQa0jHvM=
 github.com/btcsuite/btcd/btcec/v2 v2.5.0 h1:KioMXOWa76b86sTZZOmbzv/ldaQCmB8KFAyn5PbB8E8=
--- a/ledger/eras/cost_models_test.go
+++ b/ledger/eras/cost_models_test.go
@@ -1,0 +1,101 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package eras
+
+import (
+	"testing"
+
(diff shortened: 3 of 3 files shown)


wolf31o2:
DevNet conformance run against a live cardano-node: `run-tests.sh --conformance` reports `All DevNet tests PASSED`, exit 0.

```
--- PASS: TestBasicBlockForging (16.05s)
--- PASS: TestDingoChainAdvances (14.02s)
--- PASS: TestChainGrowthRate (122.03s)
--- PASS: TestRelayPropagation (0.03s)
--- PASS: TestSustainedConsensus (102.07s)
--- PASS: TestCardanoProducerChainAdvances (12.02s)
--- PASS: TestEpochBoundaryConsensus (430.05s)
ok  internal/test/devnet/scenarios  696.295s
```

txpump reports submitted=6 rejected=0 error=0. The only skip is TestAcceleratedScenarioTimeline. That covers the ExUnits accounting on the block production path noted above.

arepala-uml (approved):

PR #3679, opened 2026-08-30
fix(leios): retry deferred voting activation

### LINKED ISSUE #3644: A block producer whose Leios voting key is not yet resolvable on-chain cannot start
On current `main` (b0cbcdf3, after "fix(leios): require on-chain voter keys",
aefbf720), a block producer configured with a Leios vote signing key **fails to
start** if the key cannot be resolved on-chain:

```
level=ERROR "node error"
  error="failed to enable leios voting: validate configured leios vote signing
         key: no usable ..."
```

Two things combine into a startup deadlock:

1. `DINGO_LEIOS_VOTER_PUBLIC_KEYS` (the static registry that previously stood in
   for CIP-0164 registration) is **no longer supported** — the node exits at
   config load if it is set:
   `DINGO_LEIOS_VOTER_PUBLIC_KEYS is no longer supported`.
2. Voting-key validation now resolves only from on-chain registration, against
   the stake-snapshot epoch of the node's OWN ledger.

The result is circular for a node that is behind: it must resolve its voting key
on-chain to start, but it cannot reach the ledger state that carries that
registration until it starts and syncs. A pool that is registered and actively
voting on-chain still cannot bring its node up after any event that requires a
resync/catch-up (fresh bootstrap, sideload, DB replacement). The only way we
found to start such a node was to remove the vote signing key entirely and run
non-voting.

Prior behavior (v0.70.2 and earlier) allowed the static registry as a documented
stand-in ("or the node's own key before it is visible on-chain"), which avoided
this deadlock.

**Suggested fix:** allow a node to START and catch up when its voting key is not
yet on-chain-resolvable (defer voting-enable until the key resolves, logging a
warning), rather than refusing to start. Or retain a static-registry / local
override path for the pre-visibility window.

wolf31o2:
## Summary

- replay ready current-epoch announcements when voting activates
- retain retryable activation state across transient provider failures
- preserve startup and authorization behavior with regression coverage

Closes #3644

## Validation

- focused and full ledger/leios tests with race detection
- conformance, vet, lint, formatting, and architecture checks

---
## Summary by cubic
Defers Leios voting activation when a node starts before its pool's on-chain key registration reaches the local ledger snapshot, so startup no longer fails just because the node is behind. The signing key is retained as deferred configuration and re-checked at each epoch transition; once a PoP-verified matching key resolves, ready current-epoch announcements are replayed in deterministic order before voting is enabled. Closes #3644.

**Behavior**
- The initial lookup distinguishes absence from provider failure: absence defers activation, while a provider error is fatal to startup and clears deferred state.
- A key resolvable at startup that mismatches the configured key remains a hard configuration error.
- During deferred retries, provider errors, invalid proofs, and mismatches are nonfatal, leaving voting disabled and the configuration retryable.
- Ready announcements replay by slot then ranking-block hash; a failed replay preparation keeps voting disabled and the activation retryable.
- Lookups carry a generation; a result finishing after a newer configuration or retry is a nonfatal superseded diagnostic that leaves voting state untouched.
- A vote signature completing after a reconfiguration is discarded without changing vote state or emitting an event.
- Node startup logs a deferred-voting message and continues when activation is deferred.

<sup>Written for commit f883ae91a8583ecdf00325a667eb7ad22a610808. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3679?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Leios voting now supports deferred activation when on-chain key registration is not yet available.
  * Voting automatically retries after epoch transitions and activates once authorization becomes available.
  * Pending announcements are replayed in a consistent order when voting is enabled.

* **Bug Fixes**
  * Startup no longer fails when the local snapshot is behind the required on-chain registration.
  * Transient activation failures can recover without restarting the node.


arepala-uml (approved):
**No merge blockers.**

Reviewed against `origin/main` (96fab5b9) in an isolated worktree at head `f883ae91`.

**Correctness**
- `ConfigureVoting`'s generation-counter scheme correctly supersedes stale in-flight provider lookups without holding a lock across the network round-trip.
- `localEmissionMu` correctly orders activation replay ahead of ordinary emission; a vote signature completing after reconfiguration is discarded via the `votingConfigurationSnapshot` check in `insertVote` without mutating state or emitting an event.
- `node_leios.go`'s `VotingConfigurationFailed`-with-nil-error branch is unreachable given current invariants — defensive only, not a defect.

**Documentation**
- `ARCHITECTURE.md` updated and matches the implementation. `DATABASE.md` correctly untouched (no schema change).

**Validation**
- `go test -race ./ledger/leios/... .` — pass
- `go test ./internal/test/conformance/...` — pass
- `go vet`, `gofmt -l`, `golangci-lint run`, `golines -l` — clean
- `go test ./internal/docsparity`, `go test ./internal/architecture` — clean
- `nilaway` flags one finding in `manager_test.go`; confirmed present on `origin/main` baseline, not a regression from this branch.

**Bot review**
- CodeRabbit: completed, one nitpick (conflated "awaiting key" vs. "replay-prep failed" logging) — already resolved by the current `VotingConfigurationStatus` enum; no open thread remains.
- Cubic: rate-limited this cycle (monthly line limit); CodeRabbit's completed review stands in per policy.
- Required checks (lint, go-test Linux/Windows, govulncheck, CodeQL, DCO, Conventional Commits) all pass.

Test coverage for the deferred-activation, retry, and reconfiguration-race paths is thorough, including a direct regression test for the original startup deadlock (`TestEnableLeiosVotingDefersUntilOnChainKeyAvailable`).

PR #3706, opened 2026-08-31
fix(governance): support Dijkstra epoch transitions

### LINKED ISSUE #3695: fix(governance): process Dijkstra epoch transitions
## Summary

Dijkstra protocol parameters are currently treated like pre-Conway parameters during epoch processing. Ratification, enactment, proposal expiry, refunds, and orphan cleanup are therefore skipped for the era.

Widening that gate alone also exposes an unsupported protocol-version update path, so both behaviors need to move together.

## Scope

- Process governance epochs with both Conway and Dijkstra protocol parameters.
- Preserve Dijkstra-specific fields when a hard-fork proposal updates the protocol version.
- Keep the pre-Conway no-op behavior unchanged.
- Add regression coverage for Dijkstra ratification and protocol-version enactment.

## Validation

- Relevant governance tests pass normally and under the race detector.
- Existing Conway governance behavior remains unchanged.

wolf31o2:
## Summary

- process Dijkstra governance parameter changes through ratification and enactment
- preserve Dijkstra protocol parameters across rollover, persistence, and restart
- deep-clone mutable parameter state and repair supported governance proposals safely

## Validation

- governance tests, normal and race
- full ledger tests
- conformance tests
- `go vet` and `golangci-lint`
- NilAway comparison and documentation parity

Fixes #3695


wolf31o2:
Reviewed the replay/refund concern at e62980bd. Epoch rollover restores an absolute treasury baseline before governance replay, and refund plus proposal finalization commit in one database transaction; rollback removes both lifecycle and boundary state. The unregistered-account replay regression confirms 100 to 125 on the first pass and 100 to 125 after replay, not 150. No code change is needed.

arepala-uml (dismissed):

Issue #3646, opened 2026-08-30
validation: genesis replay excludes relay-g1 headers when producer pools are absent from epoch snapshots


wolf31o2:
During the current-origin Prime Mainnet genesis replay at b0cbcdf3a494eec85e291354d2baf4447f7c7c38, Dingo locally excluded relay-g1 headers because the referenced producer pools were absent from Dingo epoch snapshots. Observed slots: 3799503, 3799519, 7800123, 13003066, 13003109, and 16295233.

This needs comparison against the canonical Haskell chain and snapshot timing. It is not being attributed to the peers. Preserved evidence: /tmp/dingo-validation/runs/prime-mainnet-genesis-origin-b0cbcdf/genesis-serve.log.

Investigate whether snapshot import/rotation or producer-pool resolution causes valid headers to be excluded during from-genesis replay, and add regression coverage after the canonical behavior is established.

Issue #3626, opened 2026-08-29
validation: Preview genesis rejects canonical block at leadership threshold


wolf31o2:
## Problem

The Preview from-genesis validation run on baseline `aefbf720c59c0d9c091dbae70393d140b2ed29bb` rejected a canonical block during genesis replay and did not make forward progress. The node repeatedly rejected block slot `1730450` because the producer's VRF leader value exceeded the locally derived stake threshold:

```
pool stake: 100000000000000
total stake: 352422574596382
epoch: 20
snapshot_epoch: 19
snapshot_type: mark
threshold_margin: -0.000565555
```

The run reached applied slot `1730401` and then remained in rewind and re-selection churn.

## Impact

Preview genesis sync cannot reach the tip or enter its 30-minute live serve window.

## Evidence

- Run: `/tmp/dingo-validation/runs/preview-genesis-origin-aefbf720-fresh`
- Repeated rejection: `genesis-serve.log:31031`

## Related history

This is a current reproduction of the reward-account/leadership-stake class described by closed issue #3165. The current producer, slot, and baseline differ from that report and require a fresh resolution or an explicit regression disposition.

## Acceptance criteria

- Determine why the locally computed leadership stake rejects this canonical Preview block.
- Correct the stake or eligibility handling, or document a protocol-compatible disposition.
- Re-run Preview genesis from the same baseline family through tip and the 30-minute live window.

wolf31o2:
## Disposition: fixed by #3800

Same defect as #3794, which carries the identical signature — failing block
slot 1730450, rewind target 1730401, epoch 20, snapshot_epoch 19, mark, pool
stake 100000000000000, total stake 352422574596382, threshold_margin
-0.000565555.

Root cause: POOLREAP did not domain-restrict the delegation map by the reaped
pools. `pool1nk3uj4fdd6d42tx26y537xaejd76u6xyrn0ql8sr4r9tullk84y`
(`9da3c9552d6e9b552ccad1291f1bb9937dae68c41cde0f9e03a8cabe`) carried a
retirement certificate for epoch 14 at slot 1207151 and re-registered at slot
1616332 for epoch 20. One stale delegator, `stake_test1uqgefktc...c73fd73`,
delegated at slot 971875 and never again; Koios reports it as
`delegated_pool: null`. Its 1001812732483 lovelace rejoined the pool
distribution at the re-registration, which is the entire mark(19) denominator
error:

```
dingo  mark(19) total  352422574596382
[...]

PR #3717, opened 2026-08-31
fix(ledger): retain deferred-header snapshots instead of misclassifying pruned snapshots as pool absence

### LINKED ISSUE #3727: ledger: retain pool snapshots needed by deferred header validation
## Problem

A current-origin Preprod from-genesis replay stalls when deferred header
validation needs historical pool-stake snapshots that have already been
pruned. Dingo then treats the missing snapshot as an absent pool and rejects
the canonical header; deep forks, blockfetch non-fits, connection churn, and
`no selectable peer` are recovery fallout.

## Evidence

- Baseline: `92052fdcf92ae219d3a15019dacfc41624341f40`
- Run: `/tmp/dingo-validation/runs/preprod-genesis-origin-92052fd`
- Applied tip stopped at slot `4406404`; the rejecting header was slot
  `4406406`.
- The database retained mark snapshots for epochs 25–28, while queued headers
  required epochs 10, 13, and 21.
- Snapshot cleanup prunes entries older than `currentEpoch-3` in
  `ledger/snapshot/rotation.go`; `leaderEligibilityStake` reports a missing
  row as an absent pool in `ledger/verify_header.go`.
- Both canonical Haskell peers advertised tracked tips near slot 132.5M; no
  peer-side fault was established.

## Related work

This is distinct from #3691 (untracked fallback promotion), #3626 (Preview
leadership-threshold mismatch), and #3646 (Prime Mainnet snapshot
provenance).

## Acceptance criteria

- Deferred headers retain or otherwise resolve every historical pool-stake
  snapshot required for validation.
- Missing historical snapshot state is not silently classified as pool absence.
- Add regression coverage for the retention/deferred-validation interaction
  and verify recovery advances after the rejected boundary.
- Update `DATABASE.md` and `ARCHITECTURE.md` if the retention or validation
  lifecycle contract changes.

chadaibbhot-git:
## What / why
Fixes #3727 (and the class in #3646 / #3626). A header whose leader-eligibility check needs a
historical pool-stake snapshot can be **deferred** while the node catches up, but snapshot cleanup
(`ledger/snapshot/rotation.go` `cleanupOldSnapshots`) prunes `pool_stake_snapshot` rows below
`currentEpoch-3`. When a deferred header requires an older epoch than what survived (observed:
required epochs 10/13/21 while the DB retained 25–28), `leaderEligibilityStake` finds no row and
reports the producer as **absent from the epoch distribution** → `header_verification_failure` →
peer recycled, and the node cannot pass the block. The root cause is snapshot *pruning*, not a
genesis/epoch-0 data gap (this replaces the earlier epoch-0-only defer, which neither covered the
real epochs nor distinguished a pruned snapshot from a genuinely-absent pool).

## Fix — two coherent parts, keyed on observable snapshot availability (never on `epoch==0`)

**(A) Retention pin — the root cause.** `LedgerState.OldestRequiredSnapshotEpoch` returns the oldest
snapshot epoch any outstanding deferred header still needs (min over the `deferredHeaderValidation`
set). `Manager.SetRetentionFloorProvider` (wired at node start) makes `cleanupOldSnapshots` pin
**pool-snapshot** deletion at that floor when it is below `currentEpoch-3`, so a snapshot a deferred
header needs is not pruned out from under it. It is a lower-watermark only (never prunes *more* than
today), releases automatically as the deferred set drains, leaves the reward-state window untouched,
and is byte-identical to prior behaviour when no provider is set.

**(B) Never classify missing snapshot state as pool absence.** `leaderEligibilityStake` already
separates an empty/unavailable distribution (`GetTotalActiveStake == 0` → `errLeaderStakeSnapshotUnavailable`)
from a pool absent in a *populated* one (authoritative reject). The classifier now **defers any
unavailable leader-stake snapshot regardless of the apply cursor**, instead of hard-rejecting once the
tip passed the slot — so a snapshot missing for any reason (pruned, unwritten, or momentarily gone
after restart before the deferred set repopulates) defers and never recycles the peer. A pool absent
from a populated snapshot still hard-rejects (covered).

Together: (A) keeps the required snapshot so the deferred header resolves; (B) guarantees that if a
snapshot is nonetheless gone, the header defers rather than being misclassified as absence.

## Tests
`ledger/verify_header_test.go`: `TestVerifyBlockHeaderState_UnavailableSnapshotDefersNotAbsence`
(pruned historical snapshot with the tip **ahead** defers; populated-snapshot genuine absence still
hard-rejects) and `TestOldestRequiredSnapshotEpoch`. `ledger/snapshot/rotation_test.go`: retention
floor retains deferred-header epochs below the default window while still pruning below the pin, and
is a no-op above the window. All pass under `-race`; `DATABASE.md` / `ARCHITECTURE.md` updated.

## Known limitation (flagged for review)
The retention floor is computed from the **in-memory** `deferredHeaderValidation` set (there's no
prefix-scan API on `sync_state`; adding one is a larger DB change). This is correct for the
live-pruning-during-sync case #3727 describes. Across a restart there is a brief window before the
deferred set repopulates where the pin is empty — **part (B) covers exactly that gap** (defer, don't
misclassify), so correctness holds; a durable persisted pin would be a follow-up if wanted.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented required pool-stake snapshots from being deleted while deferred header validation still needs them.
  - Headers with unavailable leader-stake snapshots are now deferred instead of incorrectly rejected as if the pool were absent.
  - Snapshot retention pins release automatically as deferred headers are resolved.
  - Reward-state retention behavior remains unchanged.

- **Documentation**
  - Documented configurable snapshot retention and deferred-header retention behavior.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -7353,17 +7353,25 @@ it is the leader-eligibility basis a queued/deferred header validates against.
 `cleanupOldSnapshots` prunes `PoolStakeSnapshot` through a retention guard
 (`Manager.SetPoolSnapshotRetentionGuard`, wired to
 `LedgerState.PrunePoolSnapshotsWithRetentionFloor`) that, under one dedicated
-lock held across the whole decision and the pool-snapshot delete+commit: (1)
+lock (`deferredHeaderValidationMu`) held across the eviction and floor read but
+released before the pool-snapshot delete: (1)
 evicts deferred headers the apply cursor has already passed — abandoned, since a
 canonical one is consumed at apply — so they stop pinning their snapshots and
 their persisted markers are deleted; (2) lowers the delete boundary to the floor
 (oldest `StakeSnapshotEpoch(epochOf(slot))` over the survivors) when it is below
 `current-3`, or to 0 while any deferred slot is not yet epoch-mappable; and (3)
 clamps that boundary up to a hard depth cap (`current - 24`) so a stuck header
-can never pin snapshots without bound. Holding the lock across floor read and
-delete+commit means admission cannot interleave between them (the race
-CodeRabbit flagged; `markDeferredHeaderValidation` takes the same lock and does
-no I/O, so no deadlock). The pin moves only `PoolStakeSnapshot` — the reward
+can never pin snapshots without bound. The eviction+floor read is atomic (one
+lock hold), so the boundary is a coherent read of the deferred set. The lock is
+released before `prune` (and before each `DeleteSyncState` in the evicted-marker
+cleanup): `prune` opens the single SQLite write connection, and block apply holds
+that connection before taking this same mutex via `consumeDeferredHeaderValidation`,
+so holding the mutex across `prune` inverts the lock order and deadlocks the node
+on the single write connection (issue #3717). The hazard is not this lock's own
+I/O — it has none — but the apply path holding the write connection while waiting
+for this mutex; a header admitted after the lock is released is handled by the
+next pass, since the floor is a lower-watermark recomputed each cleanup. The pin
+moves only `PoolStakeSnapshot` — the reward
 window stays at `current-3` — never prunes more than the default, releases as
 headers resolve or are evicted, and is rebuilt from the persisted markers at
 startup (`repopulateDeferredHeaderValidation`) so it survives a restart. The pin
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -7346,8 +7346,36 @@ full per-epoch and per-pool reward history stays queryable and a missing
 summary row keeps its diagnostic meaning of a boundary that was never
 captured. Because a retained snapshot can outlive its per-credential rows,
 `applyStakeRewards` skips an epoch whose snapshot claims delegators over an
-empty `RewardStakeInput` set rather than failing the rollover. See the
-retention section in `DATABASE.md` for the per-table detail.
+empty `RewardStakeInput` set rather than failing the rollover.
+
+`PoolStakeSnapshot` retention has one extra lower-watermark rule (issue #3727):
+it is the leader-eligibility basis a queued/deferred header validates against.
+`cleanupOldSnapshots` prunes `PoolStakeSnapshot` through a retention guard
+(`Manager.SetPoolSnapshotRetentionGuard`, wired to
+`LedgerState.PrunePoolSnapshotsWithRetentionFloor`) that, under one dedicated
+lock held across the whole decision and the pool-snapshot delete+commit: (1)
+evicts deferred headers the apply cursor has already passed — abandoned, since a
+canonical one is consumed at apply — so they stop pinning their snapshots and
+their persisted markers are deleted; (2) lowers the delete boundary to the floor
+(oldest `StakeSnapshotEpoch(epochOf(slot))` over the survivors) when it is below
+`current-3`, or to 0 while any deferred slot is not yet epoch-mappable; and (3)
+clamps that boundary up to a hard depth cap (`current - 24`) so a stuck header
+can never pin snapshots without bound. Holding the lock across floor read and
+delete+commit means admission cannot interleave between them (the race
+CodeRabbit flagged; `markDeferredHeaderValidation` takes the same lock and does
+no I/O, so no deadlock). The pin moves only `PoolStakeSnapshot` — the reward
+window stays at `current-3` — never prunes more than the default, releases as
+headers resolve or are evicted, and is rebuilt from the persisted markers at
+startup (`repopulateDeferredHeaderValidation`) so it survives a restart. The pin
+is what makes a deferred header *resolve*: its snapshot is retained until the
+cursor reaches it. The classification in `verifyBlockHeaderState` is
+consensus-narrow: a leader-stake snapshot reported unavailable is deferred ONLY
+while the apply cursor is still behind the header's slot (not yet produced =
+recoverable); once the cursor has caught up a still-empty distribution is a
+genuine, permanent gap and stays a hard rejection, exactly as before, and a
+producer absent from a populated snapshot always hard-rejects. See the retention
+section in `DATABASE.md` for the per-table
+detail.
 
 A skipped reward round is never made up later, and that has consequences
 well beyond the reward figures. Applying the round at the boundary into N
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2400,6 +2400,55 @@ at a normal-boundary `d` decrease. Blockfetch still defers the stateful overlay
 decision until ledger apply when even the forecast cannot resolve it; full
 historical validation and the normal leader checks remain enabled.
 
+**Why deferring a `d=1` overlay header is a real fix, not a suppression.** With
+`d = 1` (full federation — every early Shelley epoch, and epoch 0 on every
+network) `classifyGenesisOverlaySlot` returns `genesisOverlayActive` or
+`genesisOverlayNonActive` for *every* slot, never `genesisOverlayNone`: with
+`position = ceil(relativeSlot·d) = relativeSlot`, the position always advances by
+one per slot, and `position % activeSlotCoeffInverse == 0` selects the active
+overlay slots. So under `d=1` the genesis-delegate path — not the Praos
+leader-eligibility path — is authoritative, and a header **not** issued by the
+overlay slot's assigned active genesis delegate is rejected (`genesis overlay
+slot assigned to delegate …, got issuer …`, or the non-active-slot rejection).
+There is therefore no reference-node case in which cardano-node *accepts* a
+non-genesis-delegate block at a `d=1` overlay slot, and the defer never makes
+dingo accept one either: `verifyGenesisDelegateHeader` defers only while
+`allowStateDefer && ledgerTipBehindSlot(slot)`, and at ledger apply
+`verifyDeferredBlockHeaderState` re-runs `verifyBlockHeaderStateWithEpochAdvance(
+block, /*epochCacheAdvance*/ true, /*allowStateDefer*/ false)` — with the defer
+switch off, so the stateful genesis-delegate check runs to an authoritative
+accept/reject verdict before the block can be adopted. The deferral moves *when*
+the verdict is computed, not *whether* it is enforced; the marker
+(`deferred_header_validation:<slot>:<hash>`, in memory and in `sync_state`) is
+what forces that apply-time recheck, so its retention is load-bearing (see the
+retention-floor, eviction-horizon, and marker-restore invariants below and in
+`DATABASE.md`).
+
+The concrete acceptance case the defer *does* exist for is a genesis-delegate
+**reassignment** that the apply cursor has not reached yet. A
+`GenesisKeyDelegationCertificate` rewrites a genesis key's active delegate/VRF
+hash; `Store.GetGenesisDelegationForSlot` returns the latest
+`genesis_delegation` row with `added_slot < blockSlot`, and that row is written
+only when the block carrying the certificate is *applied*
+(`transaction_certificates.go`). During catch-up the header chain runs ahead of
+the applied tip, so at header-verification time the reassignment row can be
+absent — `activeGenesisDelegationForSlot` then falls back to the Shelley-genesis
+default delegate — while at apply time the row is present and names the new
+delegate. A block legitimately produced by the reassigned delegate would be
+*rejected at header time* (issuer ≠ genesis-default delegate) but *accepted at
+apply* (issuer == reassigned delegate), exactly as cardano-node accepts it, so
+deferring the `d=1` overlay decision until apply is what keeps dingo from
+recycling an honest peer over a valid block. This is why the current premise is
+keyed on the observable `ledgerTipBehindSlot(slot)` state rather than on
+`epoch == 0`: the stale state that motivates the defer is the unapplied
+genesis-key-delegation row, which is not epoch-0-specific. (Empirical note: this
+is substantiated from the certificate/overlay code paths and TPraos overlay
+semantics; a specific on-chain instance — network, slot, and the reassigning
+transaction — has not been pinned here and must not be invented. If a concrete
+witness is wanted for the PR record, it is the one open item for the author to
+supply; the safety argument above does not depend on it, because apply-time
+re-validation is authoritative regardless.)
+
 Slot/epoch query adapters preserve `hardfork.ErrPastHorizon` in their error
 chains so callers can defer until the ledger advances. `EpochInfo` serves an
 already materialized epoch directly from the immutable epoch cache before
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2433,9 +2433,9 @@ only when the block carrying the certificate is *applied*
 (`transaction_certificates.go`). During catch-up the header chain runs ahead of
 the applied tip, so at header-verification time the reassignment row can be
 absent — `activeGenesisDelegationForSlot` then falls back to the Shelley-genesis
-default delegate — while at apply time the row is present and names the new
+delegate — while at apply time the row is present and names the new
 delegate. A block legitimately produced by the reassigned delegate would be
-*rejected at header time* (issuer ≠ genesis-default delegate) but *accepted at
+*rejected at header time* (issuer ≠ the static genesis delegate) but *accepted at
 apply* (issuer == reassigned delegate), exactly as cardano-node accepts it, so
 deferring the `d=1` overlay decision until apply is what keeps dingo from
 recycling an honest peer over a valid block. This is why the current premise is
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -7355,16 +7355,24 @@ it is the leader-eligibility basis a queued/deferred header validates against.
 `LedgerState.PrunePoolSnapshotsWithRetentionFloor`) that, under one dedicated
 lock (`deferredHeaderValidationMu`) held across the eviction and floor read but
 released before the pool-snapshot delete: (1)
-evicts deferred headers the apply cursor has already passed — abandoned, since a
-canonical one is consumed at apply — so they stop pinning their snapshots and
-their persisted markers are deleted; (2) lowers the delete boundary to the floor
+evicts deferred headers beyond the rollback horizon (below
+`tip - calculateStabilityWindow()`) — abandoned, since a canonical one is
+consumed at apply and no fork that deep can be re-adopted — so they stop pinning
+their snapshots and their persisted markers are deleted. The horizon, rather than
+the bare tip, is what makes eviction safe: eviction also drops the durable
+marker, so a point evicted while still re-adoptable would later apply with
+`required == false` and be adopted with its stateful leader-eligibility check
+never run; (2) lowers the delete boundary to the floor
 (oldest `StakeSnapshotEpoch(epochOf(slot))` over the survivors) when it is below
 `current-3`, or to 0 while any deferred slot is not yet epoch-mappable; and (3)
 clamps that boundary up to a hard depth cap (`current - 24`) so a stuck header
 can never pin snapshots without bound. The eviction+floor read is atomic (one
 lock hold), so the boundary is a coherent read of the deferred set. The lock is
 released before `prune` (and before each `DeleteSyncState` in the evicted-marker
-cleanup): `prune` opens the single SQLite write connection, and block apply holds
+cleanup, whose per-key delete then re-tests membership and re-persists the marker
+for a point re-deferred in that window, so releasing the lock cannot strand a
+live deferred header without its durable marker): `prune` opens the single SQLite
+write connection, and block apply holds
 that connection before taking this same mutex via `consumeDeferredHeaderValidation`,
 so holding the mutex across `prune` inverts the lock order and deadlocks the node
 on the single write connection (issue #3717). The hazard is not this lock's own
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -941,7 +941,9 @@ Every epoch transition runs `cleanupOldSnapshots`, which prunes to the four
 epochs the Shelley rotation and delayed reward model need: current, current-1,
 current-2 for Go, and current-3 so reward calculation can be replayed after a
 rollback across the boundary where those rewards were applied. Retention is not
-configurable, and it only ever deletes rows below that window.
+operator-configurable, and it only ever deletes rows below that window (but see
+the deferred-header retention pin below, which can hold `pool_stake_snapshot`
+rows longer).
 
 Pruned at `epoch < current-3`:
 
@@ -960,6 +962,54 @@ two per-credential tables scale with delegator count — about 5k rows per epoch
 preview and 1.3M on mainnet — which is why they alone stay windowed. No
 before-epoch delete method exists for any retained table on
 `metadata.MetadataStore`.
+
+**Deferred-header retention pin (issue #3727).** `pool_stake_snapshot` has one
+extra retention rule the two windowed reward tables do not: it is the
+leader-eligibility basis a *queued/deferred header* validates against. If the
+mark snapshot such a header needs (`StakeSnapshotEpoch(epoch) = epoch-1`) were
+pruned out from under it, `leaderEligibilityStake` (`ledger/verify_header.go`)
+would read the missing rows back as a zero-stake answer. `cleanupOldSnapshots`
+therefore prunes `pool_stake_snapshot` *through a retention guard*
+(`Manager.SetPoolSnapshotRetentionGuard`, wired at node start to
+`LedgerState.PrunePoolSnapshotsWithRetentionFloor`). Under one dedicated lock,
+held across the whole decision and the pool-snapshot delete+commit, the guard:
+
+1. **Evicts abandoned headers.** A deferred header the apply cursor has already
+   passed is abandoned — a canonical one is consumed at apply — so it is dropped
+   from the set (and its persisted marker deleted) instead of pinning its
+   snapshot forever. This is what bounds retention in practice.
+2. **Computes the floor** as the minimum of `StakeSnapshotEpoch(epochOf(slot))`
+   over the surviving deferred headers and lowers the pool-snapshot delete
+   boundary to it when it is below `current-3` (reward-table retention is
+   untouched at `current-3`). While ANY surviving deferred slot is not yet
+   mappable to an epoch, the floor is **0** (retain everything) so the snapshot
+   the header will need once the epoch cache advances is not pruned meanwhile.
+3. **Clamps to a depth cap.** The boundary is clamped up to
+   `current - poolSnapshotRetentionMaxDepth` (24), a hard backstop so even a
+   stuck header can never pin more than a bounded number of epochs of pool
+   snapshots (~1.3M rows/epoch on mainnet). It never rises above the default
+   window.
+
+Because the guard holds the lock across both the floor read and the
+delete+commit, a header admitted concurrently cannot slip in between them and
+lose its snapshot (admission — `markDeferredHeaderValidation` — takes the same
+lock and does no I/O, so it cannot deadlock). The pin is a *lower-watermark
+only*: it never prunes more than the default window, releases as headers
+resolve (the set shrinks) or are evicted, and with no guard set is
+byte-identical to the original behavior. Across a restart the in-memory set is
+rebuilt from the persisted markers (`repopulateDeferredHeaderValidation`, via
+`ListSyncStateKeysByPrefix`) before the first cleanup, so the pin covers
+headers still awaiting apply from before the restart.
+
+The pin is what makes a deferred header *resolve* (its snapshot is retained
+until the cursor reaches it). The classification in `verifyBlockHeaderState` is
+deliberately narrow and consensus-safe: a leader-stake snapshot reported
+unavailable is deferred ONLY while the apply cursor is still behind the header's
+slot (the snapshot is not yet produced — recoverable). Once the cursor has
+caught up, a still-empty distribution is a genuine, permanent gap for that epoch
+and stays a hard rejection, exactly as before — deferring it forever would adopt
+a block whose leader eligibility is never checked. A producer absent from a
+*populated* snapshot is authoritative ineligibility and always hard-rejects.
 
 Together the retained tables are the full per-epoch reward record a historical
 closed-epoch comparison needs: the pots the epoch was paid from, both sets of
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -971,8 +971,9 @@ pruned out from under it, `leaderEligibilityStake` (`ledger/verify_header.go`)
 would read the missing rows back as a zero-stake answer. `cleanupOldSnapshots`
 therefore prunes `pool_stake_snapshot` *through a retention guard*
 (`Manager.SetPoolSnapshotRetentionGuard`, wired at node start to
-`LedgerState.PrunePoolSnapshotsWithRetentionFloor`). Under one dedicated lock,
-held across the whole decision and the pool-snapshot delete+commit, the guard:
+`LedgerState.PrunePoolSnapshotsWithRetentionFloor`). Under one dedicated lock
+(`deferredHeaderValidationMu`), held across the eviction and the floor read but
+**not** across the pool-snapshot delete, the guard:
 
 1. **Evicts abandoned headers.** A deferred header the apply cursor has already
    passed is abandoned — a canonical one is consumed at apply — so it is dropped
@@ -990,16 +991,34 @@ held across the whole decision and the pool-snapshot delete+commit, the guard:
    snapshots (~1.3M rows/epoch on mainnet). It never rises above the default
    window.
 
-Because the guard holds the lock across both the floor read and the
-delete+commit, a header admitted concurrently cannot slip in between them and
-lose its snapshot (admission — `markDeferredHeaderValidation` — takes the same
-lock and does no I/O, so it cannot deadlock). The pin is a *lower-watermark
-only*: it never prunes more than the default window, releases as headers
-resolve (the set shrinks) or are evicted, and with no guard set is
-byte-identical to the original behavior. Across a restart the in-memory set is
-rebuilt from the persisted markers (`repopulateDeferredHeaderValidation`, via
-`ListSyncStateKeysByPrefix`) before the first cleanup, so the pin covers
-headers still awaiting apply from before the restart.
+The eviction and the floor read happen under one lock hold, so the boundary
+handed to `prune` is a coherent read of the deferred set — never a mix of pre-
+and post-eviction state. The lock is then **released before `prune` runs**, and
+this is a correctness requirement, not just an optimization: `prune` opens the
+single SQLite write connection (`SetMaxOpenConns(1)`) via `Transaction(true)`,
+and block apply holds that connection *before* taking `deferredHeaderValidationMu`
+(`ledgerProcessBlock` → `verifyDeferredBlockHeaderState` →
+`consumeDeferredHeaderValidation`, all inside its write txn). Holding the mutex
+across `prune` would invert the lock order — `mutex → write-conn` in the guard
+vs. `write-conn → mutex` on apply — and **deadlock the node on the single write
+connection** (issue #3717). The same applies to the evicted-marker cleanup
+(`deletePersistedDeferredMarkers`), which takes the lock only to test membership
+per key and releases it before each `DeleteSyncState` (also a write-connection
+op). The earlier claim that this lock "does no I/O, so it cannot deadlock" was
+wrong: the hazard is never the mutex holder's own I/O, it is the *caller on the
+other path* holding the single write connection while it waits for this mutex.
+
+Releasing the lock before `prune` means a header admitted after the release is
+not pinned by the pass in flight. This is safe because the retention floor is a
+*lower-watermark recomputed every cleanup pass*: such a header needs a
+below-`current-3` snapshot only if it is deeply lagged, and the next pass pins
+the lower floor and the header resolves then. The pin never prunes more than the
+default window, releases as headers resolve (the set shrinks) or are evicted, and
+with no guard set is byte-identical to the original behavior. Across a restart the
+in-memory set is rebuilt from the persisted markers
+(`repopulateDeferredHeaderValidation`, via `ListSyncStateKeysByPrefix`) before the
+first cleanup, so the pin covers headers still awaiting apply from before the
+restart.
 
 The pin is what makes a deferred header *resolve* (its snapshot is retained
 until the cursor reaches it). The classification in `verifyBlockHeaderState` is
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -975,10 +975,19 @@ therefore prunes `pool_stake_snapshot` *through a retention guard*
 (`deferredHeaderValidationMu`), held across the eviction and the floor read but
 **not** across the pool-snapshot delete, the guard:
 
-1. **Evicts abandoned headers.** A deferred header the apply cursor has already
-   passed is abandoned — a canonical one is consumed at apply — so it is dropped
-   from the set (and its persisted marker deleted) instead of pinning its
-   snapshot forever. This is what bounds retention in practice.
+1. **Evicts abandoned headers.** A deferred header beyond the *rollback horizon*
+   — below `tip - calculateStabilityWindow()` (3k/f slots, 2k in Byron) — is
+   abandoned: a canonical one is consumed at apply, and chain selection can no
+   longer re-adopt a fork that deep. It is dropped from the set and its
+   persisted marker deleted instead of pinning its snapshot forever. This is
+   what bounds retention in practice. Eviction is gated on the horizon rather
+   than on the bare tip because it deletes the durable marker too, and that
+   marker is the only thing that makes `deferredHeaderValidationRequired` return
+   true at apply: a point evicted while a rollback could still re-adopt its fork
+   would apply with `required == false`, so `verifyDeferredBlockHeaderState`
+   returns nil and the block is adopted with its stateful leader-eligibility
+   check never run. Keys whose slot cannot be parsed are evicted regardless —
+   they can never be validated.
 2. **Computes the floor** as the minimum of `StakeSnapshotEpoch(epochOf(slot))`
    over the surviving deferred headers and lowers the pool-snapshot delete
    boundary to it when it is below `current-3` (reward-table retention is
@@ -1004,7 +1013,14 @@ vs. `write-conn → mutex` on apply — and **deadlock the node on the single wr
 connection** (issue #3717). The same applies to the evicted-marker cleanup
 (`deletePersistedDeferredMarkers`), which takes the lock only to test membership
 per key and releases it before each `DeleteSyncState` (also a write-connection
-op). The earlier claim that this lock "does no I/O, so it cannot deadlock" was
+op). Because that membership test cannot be atomic with the delete, the delete
+is made conditional *after the fact*: `deleteDeferredMarkerUnlessReadmitted`
+re-tests membership once the `DeleteSyncState` returns and re-writes the marker
+(`SetSyncState`, idempotent) for a key that was re-deferred in the window.
+Without that restore, a point re-admitted between the test and the delete keeps
+its in-memory entry but loses its durable marker, so after a restart
+`repopulateDeferredHeaderValidation` misses it and the apply-time stateful check
+is skipped for a header that is still outstanding. The earlier claim that this lock "does no I/O, so it cannot deadlock" was
 wrong: the hazard is never the mutex holder's own I/O, it is the *caller on the
 other path* holding the single write connection while it waits for this mutex.
 
--- a/database/plugin/metadata/sqlstore/operational.go
+++ b/database/plugin/metadata/sqlstore/operational.go
@@ -252,12 +252,17 @@ func (s *Store) ListSyncStateKeysByPrefix(
 	defer rows.Close()
 	var keys []string
 	for rows.Next() {
-		var key string
+		// Scan into sql.NullString: SQLite permits NULL in a TEXT PRIMARY KEY
+		// column, and scanning a NULL into a plain *string errors out, which
+		// would abort the whole scan and prevent every valid deferred-header
+		// marker from being restored. One malformed legacy row must not disable
+		// retention-pin restoration, so skip NULLs instead of failing.
+		var key sql.NullString
 		if err := rows.Scan(&key); err != nil {
 			return nil, fmt.Errorf("list sync state keys: %w", err)
 		}
-		if strings.HasPrefix(key, prefix) {
-			keys = append(keys, key)
+		if key.Valid && strings.HasPrefix(key.String, prefix) {
+			keys = append(keys, key.String)
 		}
 	}
 	if err := rows.Err(); err != nil {
--- a/database/plugin/metadata/sqlstore/operational.go
+++ b/database/plugin/metadata/sqlstore/operational.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"math"
 	"strconv"
+	"strings"
 
 	"github.com/blinklabs-io/dingo/database/models"
 	sqlitequery "github.com/blinklabs-io/dingo/database/plugin/metadata/sqlstore/internal/query/sqlite"
@@ -221,6 +222,48 @@ func (s *Store) DeleteSyncState(key string, txn types.Txn) error {
 		return fmt.Errorf("delete sync state: %w", err)
 	}
 	return nil
+}
+
+// ListSyncStateKeysByPrefix returns every sync_state key that has the given
+// byte prefix, sorted ascending. It enumerates the (small) sync_state keyspace
+// and filters the exact prefix in Go with strings.HasPrefix, so the match is
+// byte-exact and identical on every backend. A SQL range scan or LIKE would be
+// wrong here: on MySQL/Postgres the >=/< and LIKE operators honor the column's
+// COLLATION, not byte order, so a case- or locale-insensitive collation could
+// include keys that lack the byte prefix (or exclude keys that have it), and a
+// synthesized range upper bound over a non-ASCII prefix can be invalid UTF-8.
+// The deferred-header retention markers this enumerates (issue #3727) must be
+// matched exactly or a restart could miss a marker and fail to pin a snapshot.
+func (s *Store) ListSyncStateKeysByPrefix(
+	prefix string,
+	txn types.Txn,
+) ([]string, error) {
+	db, ctx, err := s.readDBFromTxn(txn)
+	if err != nil {
+		return nil, fmt.Errorf("list sync state keys: %w", err)
+	}
+	rows, err := db.QueryContext(
+		ctx,
+		s.dialect.Rebind("SELECT sync_key FROM sync_state ORDER BY sync_key"),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("list sync state keys: %w", err)
+	}
+	defer rows.Close()
+	var keys []string
+	for rows.Next() {
+		var key string
+		if err := rows.Scan(&key); err != nil {
+			return nil, fmt.Errorf("list sync state keys: %w", err)
+		}
+		if strings.HasPrefix(key, prefix) {
+			keys = append(keys, key)
+		}
+	}
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("list sync state keys: %w", err)
+	}
+	return keys, nil
 }
 
 func (s *Store) ClearSyncState(txn types.Txn) error {
--- a/database/sync_state_prefix_test.go
+++ b/database/sync_state_prefix_test.go
@@ -1,0 +1,97 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package database
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestListSyncStateKeysByPrefix covers the prefix scan that repopulates the
+// deferred-header retention set after a restart (issue #3727, finding 3). The
+// match must be an exact BYTE prefix on every backend, so this asserts cases a
+// collation-sensitive SQL range/LIKE would get wrong (finding: collation-unsafe
+// prefix scan). The prefix also contains a LIKE wildcard ('_') to prove no
+// wildcard escaping is needed.
+func TestListSyncStateKeysByPrefix(t *testing.T) {
+	db, err := newTestDatabase(t, &Config{DataDir: t.TempDir()})
+	require.NoError(t, err)
+	defer func() { require.NoError(t, db.Close()) }()
+
+	const prefix = "deferred_header_validation:"
+	want := []string{
+		prefix + "10:aa",
+		prefix + "13:bb",
+		prefix + "21:cc",
+	}
+	for _, k := range want {
+		require.NoError(t, db.SetSyncState(k, "true", nil))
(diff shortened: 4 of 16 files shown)


ledger/verify_header.go
```diff
@@ -525,6 +539,23 @@ func (ls *LedgerState) verifyBlockHeaderState(
...
+		// Genesis-clamped epoch-0 snapshot: the epoch-0 seed is a known
+		// dingo-side data gap, so a pool absent from it is not an
+		// authoritative VRFKeyUnknown even after the apply cursor has passed
+		// this historical slot. Defer without the ledgerTipBehindSlot gate so
+		// the header stays eligible for chain selection and the peer is not
+		// recycled. Populated post-genesis epochs are unaffected: they carry
+		// errLeaderStakeSnapshotUnavailable (or a plain reject), never the
+		// genesis sentinel, and keep the authoritative rejection below.
+		if allowStateDefer &&
+			errors.Is(err, errGenesisLeaderStakeUnavailable) {
```
wolf31o2: This is the line that turns a leader-eligibility rejection into a deferral, so it is where the premise has to be load-bearing.

Two questions before this lands:

1. What does an epoch-0 deferred header do at apply time? The reasoning for dropping the `ledgerTipBehindSlot` gate is that the epoch-0 seed never becomes complete, so the apply-time re-check — which `state.go:4394` calls authoritative — hits the same `errGenesisLeaderStakeUnavailable`. Adopting on a second defer means epoch-0 leader eligibility is never checked at all; rejecting means the node still cannot pass the block. The test asserts only the classifier's verdict, so neither outcome is pinned.

2. Can this fire for a producer that is genuinely ineligible? The branch keys on the epoch alone, not on any evidence that the snapshot is actually incomplete. If the epoch-0 distribution is populated and correct for a given network, an invalid epoch-0 header takes this path too. Gating on an observable incompleteness signal — an empty snapshot, or a producer with no registration certificate at all — would keep the fix aimed at the data gap it is named for rather than at every epoch-0 rejection.

wolf31o2 (changes requested):
Requesting changes. Two things need resolving before a consensus check is relaxed, plus a mechanical blocker.

**DCO is failing.** No `Signed-off-by` on the commits — a hard merge gate independent of the review, fixable with `git commit --amend -s` (or a rebase for multiple commits).

**The premise appears to contradict itself.** The description rests on "on a `d=1` network an early producer can be neither a genesis delegate nor present in the genesis staking distribution, yet the reference (cardano-node) accepts the header." With `d = 1` every slot is an overlay slot — that is what `classifyGenesisOverlaySlot` encodes — and an overlay slot must be produced by the active genesis delegate. A non-delegate block in such a slot should be rejected by the reference too, so as written the two halves of the premise cannot both hold. Either the affected slots are not actually at `d = 1`, or the producer *is* a genesis delegate that dingo is failing to resolve. Which one it is decides whether this change is a fix or a suppression, and nothing in the PR or the tests settles it. A concrete case — network, slot, producer key, and the reference node accepting it — would.

**The change's own follow-up note points at a root cause that is already implemented.** It raises "whether some residual epoch-0 producers are *rotated* genesis delegates that the overlay path should resolve from on-chain genesis-key-delegation certs rather than the static genesis `genDelegs`." Dingo already does that: `verifyGenesisDelegateHeader` resolves the active delegation per slot through `GetGenesisDelegationForSlot(initial.genesisHash, slot, ...)`, falling back to the static genesis entry only when no rotation row exists, and `GenesisKeyDelegation` certificates are persisted by `transaction_certificates.go`. So the hypothesis is not unexplored — it is implemented, which means either that lookup has a gap for these producers or the cause is something else again. Either way the cause is still unidentified, and deferring the check hides the evidence that would identify it.

**The deferral looks terminal for epoch 0, and that case is untested.** The justification for dropping the `ledgerTipBehindSlot` gate is that "an epoch-0 slot the cursor has already passed is still seeded from the same incomplete genesis data" — that is, the condition does not resolve with time. But `state.go:4394` records that "admission or the deferred apply-time check remains authoritative", so a deferred header is re-verified at apply. If the epoch-0 seed is permanently incomplete, that re-check produces the same `errGenesisLeaderStakeUnavailable`, and the terminal behaviour matters: either the apply-time path defers again and the block is adopted with leader eligibility never checked, or it rejects and the node still cannot sync past the block — the original symptom moved off the peer-recycling path rather than fixed. `TestVerifyBlockHeaderState_GenesisEpochAbsentPoolDefersNotRecycles` asserts the classifier's verdict only; nothing covers what happens when the block reaches apply.

The scoping itself is careful — `snapshotEpoch == 0` / `epochId == 0` only, the sentinel wraps `errLeaderStakeSnapshotUnavailable` so existing matches still fire, and the post-genesis reject is preserved and covered by case (b). If the premise is substantiated, the shape is close to right. The concern is not the blast radius but whether the rejection being suppressed is wrong.

chadaibbhot-git:
@wolf31o2 reworked in `256b9a9c` — the fix no longer rests on the epoch-0-defer premise. Per #3727 the real cause is snapshot pruning: a queued/deferred header's mark snapshot could be pruned below the retention window and then misclassified as pool absence. The rework adds a retention pin (`OldestRequiredSnapshotEpoch`) that retains the mark snapshot a deferred header needs, so the deferral resolves at apply time instead of looping; eviction releases abandoned forks.

On your two questions specifically:
1. **Apply-time outcome is now pinned.** `TestVerifyDeferredBlockHeaderStateRunsStrictlyAtApply` asserts both directions: with the pool state still absent, apply-time `verifyDeferredBlockHeaderState` returns `ErrPoolNotFound` (strict — it does not silently adopt and does not re-defer/loop); once the snapshot/registration is present it passes. So epoch-0 leader eligibility is checked at apply, not skipped.
2. **Not epoch-gated.** The deferral is scoped to `errLeaderStakeSnapshotUnavailable` (the incompleteness sentinel) AND `ledgerTipBehindSlot` — a pool absent from a *populated* snapshot, or once the cursor has caught up, hard-rejects exactly as before. `TestVerifyBlockLeaderEligibility_*Rejects` cover those reject paths.

Full `ledger` / `ledger/snapshot` / `database` suites pass. Would appreciate a fresh look.

chadaibbhot-git:
Addressed in `256b9a9c`: the SQL range/LIKE prefix scan is replaced by `ListSyncStateKeysByPrefix`, which runs a broad query and filters the exact byte prefix in Go with `strings.HasPrefix`, so the match is byte-exact and identical on every backend regardless of column/DB collation (SQLite byte order vs MySQL/Postgres locale collations). Added `database/sync_state_prefix_test.go`.

wolf31o2 (changes requested):
Failing CI

ledger/state.go
```diff
@@ -1488,6 +1498,13 @@ func (ls *LedgerState) Start(ctx context.Context) error {
 	if err := ls.loadTip(); err != nil {
 		return fmt.Errorf("failed to load tip: %w", err)
 	}
+	// Repopulate the in-memory deferred-header set from the persisted markers
+	// so the snapshot retention floor covers headers still awaiting apply from
+	// before the restart (issue #3727, finding 3): without this the first
+	// post-restart epoch cleanup could prune a pool-stake snapshot such a
+	// header needs. Best-effort: a load failure only forgoes the pin (the
+	// per-point persisted markers still drive apply-time re-validation).
+	ls.repopulateDeferredHeaderValidation()
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid — fixed in 435a40b0. `repopulateDeferredHeaderValidation` was swallowing the marker-scan error, leaving the retention floor unpinned; once the apply cursor passes a pre-restart deferred header, its pruned snapshot hard-rejects instead of deferring — the exact bug this PR fixes, and apply-time re-validation can't recover it since the snapshot is already gone. Changed it to return the error and made `Start()` fail closed so the operator retries rather than run unpinned. Added `TestRepopulateDeferredHeaderValidation_FailsClosedOnScanError`.

ledger/chainsync.go
```diff
@@ -408,6 +408,113 @@ func (ls *LedgerState) consumeDeferredHeaderValidation(
...
+				"error", err,
+				"component", "ledger",
+			)
+		}
+	}
+}
+
+// repopulateDeferredHeaderValidation rebuilds the in-memory deferred-header set
+// from the persisted markers at startup (issue #3727, finding 3), so the
+// snapshot retention floor (PrunePoolSnapshotsWithRetentionFloor) covers
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid — fixed in 435a40b0 as suggested. An evicted point (slot below the chain tip) can still be behind the lagging apply cursor, so it may be re-deferred and re-persisted before the lock-free delete runs, dropping a marker backing a live pin. `deletePersistedDeferredMarkers` now takes `deferredHeaderValidationMu` and skips any key present in the set again; the sole caller releases the lock first and the other holders do no I/O, so no deadlock. Added `TestDeletePersistedDeferredMarkers_SkipsReAdmitted`.

database/plugin/metadata/sqlstore/operational.go
```diff
@@ -223,6 +224,48 @@ func (s *Store) DeleteSyncState(key string, txn types.Txn) error {
...
+	}
+	defer rows.Close()
+	var keys []string
+	for rows.Next() {
+		var key string
+		if err := rows.Scan(&key); err != nil {
+			return nil, fmt.Errorf("list sync state keys: %w", err)
+		}
+		if strings.HasPrefix(key, prefix) {
+			keys = append(keys, key)
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid — fixed in 435a40b0. `sync_key` is `TEXT PRIMARY KEY`, and SQLite permits NULL in a non-INTEGER primary key, so a NULL row makes `rows.Scan(&string)` fail and aborts the whole scan (disabling retention-pin restoration). Switched to `sql.NullString` and skip invalid/NULL keys. Added `TestListSyncStateKeysByPrefixSkipsNullKey`. (The separate earlier collation note on this file stays a false-positive — the query is the full-scan `SELECT sync_key … ORDER BY sync_key` + Go `HasPrefix`, run once at startup.)

ledger/verify_header.go
```diff
@@ -1569,6 +1601,135 @@ func (ls *LedgerState) epochForSlot(slot uint64) (models.Epoch, error) {
...
+	have := false
+	for key := range ls.deferredHeaderValidation {
+		slot, err := slotFromHeaderValidationKey(key)
+		if err != nil {
+			// A key we cannot parse is a deferred header whose need we cannot
+			// bound. Retain everything until it is gone rather than risk
+			// pruning a snapshot it turns out to require.
+			return 0, true
+		}
+		epoch, err := ls.epochForSlot(slot)
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid — fixed in 435a40b0. Each per-key `epochForSlot` did its own atomic `loadConsensusSnapshot()`, so the floor loop could mix cache generations across a concurrent epoch-cache publish/rollback. Now capture the cache once before the loop and resolve every deferred slot against it via a new `epochForSlotInCache` helper (also used by `epochForSlot`), keeping the retention decision coherent.

database/sync_state_prefix_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestListSyncStateKeysByPrefix covers the prefix scan that repopulates the
+// deferred-header retention set after a restart (issue #3727, finding 3). The
+// match must be an exact BYTE prefix on every backend, so this asserts cases a
+// collation-sensitive SQL range/LIKE would get wrong (finding: collation-unsafe
+// prefix scan). The prefix also contains a LIKE wildcard ('_') to prove no
```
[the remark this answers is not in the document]
chadaibbhot-git: Done in 435a40b0 — reworded the doc comment to describe the behavior and the specific cases (case-folding, sibling prefix, non-ASCII neighbor, LIKE wildcard) instead of internal review-artifact labels.

ledger/verify_header_test.go
```diff
@@ -2943,3 +3116,337 @@ func TestVerifyBlockLeaderEligibility_ImportedActivePoolAbsentStaysHardRejection
...
+			seenBefore = before
+			close(started)
+			// While the guard holds the lock, the concurrent admission must be
+			// blocked: markReturned must not fire.
+			select {
+			case <-markReturned:
+				t.Error(
+					"concurrent admission completed while the retention guard held the lock",
+				)
+			case <-time.After(100 * time.Millisecond):
```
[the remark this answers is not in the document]
chadaibbhot-git: Done in 435a40b0 — replaced the `time.After(100ms)` negative poll with `testutil.RequireNoReceive`, per the repo's testing rule.

chadaibbhot-git:
On the failing `go-test (Linux)` check: I believe it's a pre-existing conformance flake, not this PR. The panic is `test timed out after 20m0s` in `TestRulesConformanceVectorsWithResultsMysql` — the hang is a MySQL `BeginTx` blocked on a network read in the conformance harness (`state_provider.go` / `withBadConnRetry`), and none of those files are in this PR's diff. This PR touches header verification, deferred-header handling, snapshot rotation, and an additive `ListSyncStateKeysByPrefix` — nothing on the conformance execution path. The same `test timed out after 20m0s` conformance failure hit #3739 the same day, and in this PR's own run every affected package passed (ledger, sqlstore, postgres, database, sqlite). Could you re-run the conformance job? Happy to dig further if it reproduces after a re-run.

The pushed commit (435a40b0) addresses all of cubic's re-review points (replies inline).

ledger/state.go
```diff
@@ -1488,6 +1498,21 @@ func (ls *LedgerState) Start(ctx context.Context) error {
...
+	// header its now-pruned snapshot is hard-rejected instead of deferred (the
+	// exact bug this PR fixes). Abort startup so the operator retries rather
+	// than run with an unpinned retention floor.
+	if err := ls.repopulateDeferredHeaderValidation(); err != nil {
+		return fmt.Errorf(
+			"failed to repopulate deferred-header validation set: %w",
+			err,
+		)
+	}
 	// Reconstruct the evolving-nonce fold across Mithril "gap blocks" (blocks
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid — fixed in 49c70bda. Moved `repopulateDeferredHeaderValidation` ahead of the database worker pool and cleanup-timer start in `Start()`. It only reads `ls.db` directly, so running it first means the fail-closed return unwinds nothing (previously it returned after those resources were started, but `ls.Close` is installed only once `Start` succeeds, so the pool/timer leaked).

ledger/chainsync.go
```diff
@@ -408,6 +408,139 @@ func (ls *LedgerState) consumeDeferredHeaderValidation(
...
+	ls.deferredHeaderValidationMu.Lock()
+	defer ls.deferredHeaderValidationMu.Unlock()
+	for _, k := range mapKeys {
+		// Re-admitted since eviction: its marker is now backing a live pin.
+		if _, ok := ls.deferredHeaderValidation[k]; ok {
+			continue
+		}
+		syncKey := deferredHeaderValidationSyncStatePrefix + k
+		if err := ls.db.DeleteSyncState(syncKey, nil); err != nil {
+			ls.config.Logger.Warn(
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid — fixed in 49c70bda as suggested. `deletePersistedDeferredMarkers` now takes `deferredHeaderValidationMu` per key (recheck + delete together) instead of across the whole loop, so a large eviction no longer holds the lock across one DB txn per marker and block header apply / new deferrals for the full cleanup. The re-defer race fix is preserved — a re-admitted key is still skipped atomically under the per-key lock; the lock releases between keys so apply can interleave. -race clean.

ledger/verify_header.go
```diff
@@ -1569,6 +1611,143 @@ func (ls *LedgerState) epochForSlot(slot uint64) (models.Epoch, error) {
...
+		evicted = ls.evictStaleDeferredHeadersLocked()
+		before := defaultBefore
+		if floor, ok := ls.oldestRequiredSnapshotEpochLocked(); ok &&
+			floor < before {
+			before = floor
+		}
+		if before < minBefore {
+			before = minBefore
+		}
+		return prune(before)
```
wolf31o2: Lock-order inversion, deadlocks the node.

This holds `deferredHeaderValidationMu` across `prune(before)`, and `prunePoolSnapshots` opens `m.db.Transaction(true)` immediately. That waits on the single sqlite metadata write connection (`SetMaxOpenConns(1)`) with `context.Background()`, so there is no deadline.

The apply path takes the two in the opposite order: `submitBlockApplyDBTxn` holds a read-write `database.Txn` for the whole block batch, and inside it `verifyDeferredBlockHeaderState` reaches `consumeDeferredHeaderValidation`, which takes this mutex. The two run concurrently: `cleanupOldSnapshots` is driven by `epochTransitionLoop`, started from `Manager.Start`.

Reproduced against a file-backed database:

```
sync.(*Mutex).Lock(...)
  ledger.(*LedgerState).consumeDeferredHeaderValidation
      ledger/chainsync.go:401          <- holds an open read-write database.Txn
database/sql.(*DB).conn / BeginTx
  sqlstore.(*Store).beginWriteTx  sqlstore/store.go:568
  database.NewTxn                 database/txn.go:157
  ledger.(*LedgerState).PrunePoolSnapshotsWithRetentionFloor.func1
      ledger/verify_header.go:1730     <- holds deferredHeaderValidationMu
```

At the first epoch transition overlapping an in-flight apply batch, both stop forever and nothing times out.

The shape that works: compute the boundary under the lock, release it, then prune. The admission race that motivated holding them together can be handled by re-reading the floor after the prune, or by giving prune a lower bound that admission can only lower further.

ledger/chainsync.go
```diff
@@ -408,6 +408,148 @@ func (ls *LedgerState) consumeDeferredHeaderValidation(
...
+	// can interleave.
+	for _, k := range mapKeys {
+		ls.deferredHeaderValidationMu.Lock()
+		if _, ok := ls.deferredHeaderValidation[k]; ok {
+			// Re-admitted since eviction: its marker is now backing a live pin.
+			ls.deferredHeaderValidationMu.Unlock()
+			continue
+		}
+		syncKey := deferredHeaderValidationSyncStatePrefix + k
+		err := ls.db.DeleteSyncState(syncKey, nil)
```
wolf31o2: Same inversion as the retention guard. `DeleteSyncState(syncKey, nil)` with a nil txn runs on `s.writeDB` via `dbFromTxn` with `context.Background()`, so it waits on the same single write connection while this holds `deferredHeaderValidationMu`.

```
--- FAIL: TestZZReviewMarkerDeleteLockInversion (20.31s)
    DEADLOCK: deletePersistedDeferredMarkers holds deferredHeaderValidationMu
    while DeleteSyncState waits for the single metadata write connection
```

A per-key hold is still a hold across a write-connection acquisition. Collect the keys to delete under the lock, release, then delete; re-check the re-admit condition by having the delete itself be conditional, or accept a re-deferred point being re-persisted.

ledger/chainsync.go
```diff
@@ -408,6 +408,148 @@ func (ls *LedgerState) consumeDeferredHeaderValidation(
 	return true
 }
 
+// evictStaleDeferredHeadersLocked drops deferred-header entries the apply
+// cursor has already passed (issue #3727, finding 5). A canonical deferred
+// header is consumed by consumeDeferredHeaderValidation when its block is
+// applied, so an entry still present at a slot <= the applied tip is on an
+// abandoned fork (or a rollback left it behind) and can never resolve; keeping
```
wolf31o2: Eviction deletes the durable marker as well as the in-memory entry, and the persisted marker is the only thing that makes `deferredHeaderValidationRequired` return true at apply.

If chain selection later switches back to a fork whose block was evicted, that block applies with `required == false`, so `verifyDeferredBlockHeaderState` returns nil and it is adopted with leader eligibility never checked. On main the entry and marker survive, so apply still verifies.

The comment says such an entry "can never resolve", which holds only if fork re-adoption is impossible. I did not settle the re-adoption path, so the mechanism is confirmed and the reachability is not. There is no test for rollback-then-re-adopt of an evicted point.

wolf31o2 (changes requested):
Blocking: lock-order inversion, deadlocks the node.

`PrunePoolSnapshotsWithRetentionFloor` holds `deferredHeaderValidationMu` across `prune`, and `prunePoolSnapshots` opens `m.db.Transaction(true)` immediately — the single sqlite write connection (`SetMaxOpenConns(1)`, `context.Background()`). Block apply holds that connection before taking the same mutex via `consumeDeferredHeaderValidation`. Reproduced twice with stacks; inline.

`deletePersistedDeferredMarkers` has the same inversion per key.

CI is green because every guard test passes a stub `prune` that opens no transaction.

Every other requested change is in. `DATABASE.md` and `ARCHITECTURE.md` claim this lock "does no I/O, so it cannot deadlock" — true of the mutex holders, but the hazard is what the caller already holds.

chadaibbhot-git:
Thanks for catching this — you're right, and the diagnosis is exactly the inversion you described. Fixed in bbcd450.

**Root cause (confirmed).** Two paths order the single sqlite write connection and `deferredHeaderValidationMu` oppositely:

- **prune path:** `PrunePoolSnapshotsWithRetentionFloor` held `deferredHeaderValidationMu`, then called `prune`, which opens `db.Transaction(true)` → the single write conn (`SetMaxOpenConns(1)`, `context.Background()`). So: **mutex → write-conn**.
- **apply path:** `ledgerProcessBlock` holds the write txn, then `verifyDeferredBlockHeaderState` → `consumeDeferredHeaderValidation` takes `deferredHeaderValidationMu`. So: **write-conn → mutex**.

`deletePersistedDeferredMarkers` had the same inversion per key (`DeleteSyncState(key, nil)` opens the write conn under the lock).

**Fix.** The mutex now guards only the in-memory work — evict abandoned headers + compute the retention floor — as one atomic hold; it's released before any write-connection op. `PrunePoolSnapshotsWithRetentionFloor` computes `before` under the lock, releases, then runs `prune` unlocked. `deletePersistedDeferredMarkers` takes the lock only to test membership per key and releases before each `DeleteSyncState`.

The eviction+floor read stays atomic, so the pruned boundary is still a coherent read of the deferred set. What we give up is holding the set stable *across* the prune: a header admitted after the release isn't pinned by the pass in flight. That's safe because the floor is a lower-watermark recomputed every cleanup pass — such a header only needs a below-`current-3` snapshot if it's deeply lagged, and the next pass pins the lower floor and it resolves then. I traded that narrow re-admit window for never inverting the lock order; called out in the doc comments and DATABASE.md/ARCHITECTURE.md. (The alternative that keeps the cross-prune guarantee would be to acquire the write conn *before* the mutex in the guard and thread that txn into `prune` — happy to go that way instead if you'd prefer the stronger invariant over the signature change.)

**Proof it deadlocks without the fix.** You were right that CI was green only because the guard tests passed a stub `prune` that opens no txn. New test `TestPrunePoolSnapshotsWithRetentionFloor_RealPruneNoDeadlock` passes a **real** prune (`db.Transaction(true)` + `DeletePoolStakeSnapshotsBeforeEpoch`, same as `cleanupOldSnapshots`) while a second goroutine reproduces the apply order (hold the write conn, then `consumeDeferredHeaderValidation`). On the pre-fix code it deadlocks; the two stacks are exactly your report:

```
goroutine (apply):  consumeDeferredHeaderValidation  chainsync.go:401  (sync.Mutex.Lock, holding the write txn)
goroutine (guard):  database/sql.(*DB).conn  [select]  <- PrunePoolSnapshotsWithRetentionFloor.func1 -> prune -> Transaction(true)  (holding the mutex)
```

With the fix it passes in ~0.3s. I also rewrote the former `_SerializesAdmission` test — its hold-across-prune contract *was* the bug — as `_FloorReadIsAtomic`, which asserts the surviving invariant (coherent floor read + next-pass recovery). `go test ./ledger/... ./ledger/snapshot/...` green, `-race` clean.

ARCHITECTURE.md
```diff
@@ -7346,8 +7346,44 @@ full per-epoch and per-pool reward history stays queryable and a missing
...
+canonical one is consumed at apply — so they stop pinning their snapshots and
+their persisted markers are deleted; (2) lowers the delete boundary to the floor
+(oldest `StakeSnapshotEpoch(epochOf(slot))` over the survivors) when it is below
+`current-3`, or to 0 while any deferred slot is not yet epoch-mappable; and (3)
+clamps that boundary up to a hard depth cap (`current - 24`) so a stuck header
+can never pin snapshots without bound. The eviction+floor read is atomic (one
+lock hold), so the boundary is a coherent read of the deferred set. The lock is
+released before `prune` (and before each `DeleteSyncState` in the evicted-marker
+cleanup): `prune` opens the single SQLite write connection, and block apply holds
+that connection before taking this same mutex via `consumeDeferredHeaderValidation`,
```
wolf31o2, line 7425: Same two updates on the architecture side.

```suggestion
released before the pool-snapshot delete: (1)
evicts deferred headers beyond the rollback horizon (below
`tip - calculateStabilityWindow()`) — abandoned, since a canonical one is
consumed at apply and no fork that deep can be re-adopted — so they stop pinning
their snapshots and their persisted markers are deleted. The horizon, rather than
the bare tip, is what makes eviction safe: eviction also drops the durable
marker, so a point evicted while still re-adoptable would later apply with
`required == false` and be adopted with its stateful leader-eligibility check
never run; (2) lowers the delete boundary to the floor
(oldest `StakeSnapshotEpoch(epochOf(slot))` over the survivors) when it is below
`current-3`, or to 0 while any deferred slot is not yet epoch-mappable; and (3)
clamps that boundary up to a hard depth cap (`current - 24`) so a stuck header
can never pin snapshots without bound. The eviction+floor read is atomic (one
lock hold), so the boundary is a coherent read of the deferred set. The lock is
released before `prune` (and before each `DeleteSyncState` in the evicted-marker
cleanup, whose per-key delete then re-tests membership and re-persists the marker
for a point re-deferred in that window, so releasing the lock cannot strand a
live deferred header without its durable marker): `prune` opens the single SQLite
write connection, and block apply holds
that connection before taking this same mutex via `consumeDeferredHeaderValidation`,
```

DATABASE.md
```diff
@@ -961,6 +963,73 @@ preview and 1.3M on mainnet — which is why they alone stay windowed. No
...
+(`Manager.SetPoolSnapshotRetentionGuard`, wired at node start to
+`LedgerState.PrunePoolSnapshotsWithRetentionFloor`). Under one dedicated lock
+(`deferredHeaderValidationMu`), held across the eviction and the floor read but
+**not** across the pool-snapshot delete, the guard:
+
+1. **Evicts abandoned headers.** A deferred header the apply cursor has already
+   passed is abandoned — a canonical one is consumed at apply — so it is dropped
+   from the set (and its persisted marker deleted) instead of pinning its
+   snapshot forever. This is what bounds retention in practice.
+2. **Computes the floor** as the minimum of `StakeSnapshotEpoch(epochOf(slot))`
```
wolf31o2, line 991: Documents the horizon-gated eviction rule and why the bare tip was unsafe.

```suggestion

1. **Evicts abandoned headers.** A deferred header beyond the *rollback horizon*
   — below `tip - calculateStabilityWindow()` (3k/f slots, 2k in Byron) — is
   abandoned: a canonical one is consumed at apply, and chain selection can no
   longer re-adopt a fork that deep. It is dropped from the set and its
   persisted marker deleted instead of pinning its snapshot forever. This is
   what bounds retention in practice. Eviction is gated on the horizon rather
   than on the bare tip because it deletes the durable marker too, and that
   marker is the only thing that makes `deferredHeaderValidationRequired` return
   true at apply: a point evicted while a rollback could still re-adopt its fork
   would apply with `required == false`, so `verifyDeferredBlockHeaderState`
   returns nil and the block is adopted with its stateful leader-eligibility
   check never run. Keys whose slot cannot be parsed are evicted regardless —
   they can never be validated.
2. **Computes the floor** as the minimum of `StakeSnapshotEpoch(epochOf(slot))`
```

wolf31o2 (commented):
Suggested changes carrying the fix for the two outstanding review items, since I cannot push to the fork branch.

These nine suggestions are the exact content of one commit I built and tested locally on top of bbcd4500. I verified mechanically that applying all nine to bbcd4500 reproduces that commit byte for byte. Use "Add to batch" on each and then "Commit suggestions" once, so they land as a single commit rather than nine.

What they change:

1. Eviction is gated on the rollback horizon instead of the bare tip. `evictStaleDeferredHeadersLocked` deleted the durable marker along with the in-memory entry as soon as a deferred header's slot fell below the tip. That marker is the only thing making `deferredHeaderValidationRequired` return true at apply, so a point evicted while chain selection could still roll back and re-adopt its fork applied with `required == false`, and `verifyDeferredBlockHeaderState` returned nil, adopting the block with its stateful leader-eligibility check never run. Evicting only below `tip - calculateStabilityWindow()` means nothing a rollback could resurrect is ever dropped, which removes the reachability question rather than answering it. Unparseable keys are still evicted unconditionally.

2. The marker delete becomes conditional after the fact, closing the cubic P1 at chainsync.go:475. The membership test cannot be atomic with `DeleteSyncState`, because holding `deferredHeaderValidationMu` across that write is the inversion that deadlocks the node. `deleteDeferredMarkerUnlessReadmitted` re-tests membership once the delete returns and re-persists the marker for a point re-deferred in the window. `SetSyncState` of the same key and value is idempotent and runs with no lock held.

Verification: both new tests fail against the current behaviour (I reverted each half separately and confirmed), then pass. `go test ./ledger/...` fully green. `make golines`, `make docs-parity`, `make import-boundaries` and golangci-lint across all modules plus Windows are clean, and `nilaway ./ledger/` is clean for the touched files.

Two notes on the current CI and lint state, neither caused by these changes:

- The failing `go-test (Linux) 1.26.x` is unrelated to this PR, which touches nothing under `midnight/`. `TestReflectionDisabledByDefault` fails at `server_test.go:279` with `stream.Send` returning EOF: with reflection disabled the server closes the stream with `Unimplemented` before `Send` completes, so the assertion races. It passes locally. I left it alone rather than editing an existing test.
- `make lint` fails on three pre-existing nilaway findings in `chainselection/selector.go:919`, `internal/test/fixtures/fixtures.go:86` and `ledger/leios/manager_test.go:1281`, none in files this PR touches. The CI `lint` job is green, so its configuration differs from `make lint`.

`DATABASE.md` and `ARCHITECTURE.md` are updated in the same batch, since the eviction rule and the marker delete are both described there.

Filed as a comment rather than a change request, since the existing request already covers these items.

DATABASE.md
```diff
@@ -961,6 +963,73 @@ preview and 1.3M on mainnet — which is why they alone stay windowed. No
...
+and block apply holds that connection *before* taking `deferredHeaderValidationMu`
+(`ledgerProcessBlock` → `verifyDeferredBlockHeaderState` →
+`consumeDeferredHeaderValidation`, all inside its write txn). Holding the mutex
+across `prune` would invert the lock order — `mutex → write-conn` in the guard
+vs. `write-conn → mutex` on apply — and **deadlock the node on the single write
+connection** (issue #3717). The same applies to the evicted-marker cleanup
+(`deletePersistedDeferredMarkers`), which takes the lock only to test membership
+per key and releases it before each `DeleteSyncState` (also a write-connection
+op). The earlier claim that this lock "does no I/O, so it cannot deadlock" was
+wrong: the hazard is never the mutex holder's own I/O, it is the *caller on the
```
wolf31o2, line 1024: Documents the conditional marker delete.

```suggestion
per key and releases it before each `DeleteSyncState` (also a write-connection
op). Because that membership test cannot be atomic with the delete, the delete
is made conditional *after the fact*: `deleteDeferredMarkerUnlessReadmitted`
re-tests membership once the `DeleteSyncState` returns and re-writes the marker
(`SetSyncState`, idempotent) for a key that was re-deferred in the window.
Without that restore, a point re-admitted between the test and the delete keeps
its in-memory entry but loses its durable marker, so after a restart
`repopulateDeferredHeaderValidation` misses it and the apply-time stateful check
is skipped for a header that is still outstanding. The earlier claim that this lock "does no I/O, so it cannot deadlock" was
wrong: the hazard is never the mutex holder's own I/O, it is the *caller on the
```

ledger/chainsync.go
```diff
@@ -408,6 +408,150 @@ func (ls *LedgerState) consumeDeferredHeaderValidation(
...
+	for key := range ls.deferredHeaderValidation {
+		slot, err := slotFromHeaderValidationKey(key)
+		// STRICTLY below the tip: a block whose slot the cursor has fully
+		// passed was applied (and its marker consumed) if canonical, so one
+		// still present is abandoned. Excluding slot == tip avoids racing the
+		// in-progress apply/consume of the block that just became the tip --
+		// its consume clears the marker itself, and it is evicted on a later
+		// pass only if it turns out to be abandoned.
+		if err != nil || slot < tipSlot {
+			evicted = append(evicted, key)
```
wolf31o2, line 460: Horizon-gated eviction. Evicting on `slot < tipSlot` also deletes the durable sync_state marker, which is the only thing making `deferredHeaderValidationRequired` return true at apply. A point evicted while a rollback can still re-adopt its fork applies with `required == false`, so `verifyDeferredBlockHeaderState` returns nil and the block is adopted with its stateful leader-eligibility check never run. Evicting only below `tip - calculateStabilityWindow()` makes re-adoption reachability stop mattering. `tipSlot` and the horizon are now parameters because `calculateStabilityWindow` reads `ls.currentEra` under `ls.RLock`, and apply already holds the ls lock before this mutex.

```suggestion

// evictStaleDeferredHeadersLocked drops deferred-header entries that are
// provably abandoned (issue #3727, finding 5). A canonical deferred header is
// consumed by consumeDeferredHeaderValidation when its block is applied, so an
// entry still present once the cursor has passed its slot is on a fork; keeping
// it forever would pin its epoch's pool_stake_snapshot rows forever.
//
// Eviction is gated on the ROLLBACK HORIZON, not on the bare tip. Eviction
// removes the durable sync_state marker as well as the in-memory entry, and
// that marker is the only thing that makes deferredHeaderValidationRequired
// return true at apply. If chain selection later rolls back and re-adopts a
// fork containing an evicted point, its block applies with required == false,
// so verifyDeferredBlockHeaderState returns nil and the block is adopted with
// its stateful leader-eligibility check never run -- a validation bypass that
// does not exist on main, where the entry and marker survive. Evicting only
// below tipSlot-rollbackHorizon (the stability window, 3k/f slots, past which
// chain selection cannot re-adopt a fork) makes "abandoned" mean unreachable
// rather than merely behind the cursor, so no entry that a rollback could
// resurrect is ever dropped (issue #3717 review: evicted-point re-adoption
// skips the deferred header check).
//
// Entries with an unparseable key are dropped regardless of the horizon: they
// can never be validated, so no re-adoption can make use of them. Returns the
// evicted map keys so the caller can delete their persisted markers. The caller
// must hold deferredHeaderValidationMu; tipSlot and rollbackHorizon are read by
// the caller BEFORE taking it, because both ls.loadTipSnapshot's era-derived
// window (calculateStabilityWindow) and the tip read take ls.RWMutex, and
// taking that under deferredHeaderValidationMu would invert the order block
// apply uses (ls lock -> deferredHeaderValidationMu).
func (ls *LedgerState) evictStaleDeferredHeadersLocked(
	tipSlot uint64,
	rollbackHorizon uint64,
) []string {
	if len(ls.deferredHeaderValidation) == 0 {
		return nil
	}
	// The whole chain is still inside the horizon, so nothing is unreachable
	// yet. Unparseable keys are still evicted below.
	var cutoff uint64
	if tipSlot > rollbackHorizon {
		cutoff = tipSlot - rollbackHorizon
	}
	var evicted []string
	for key := range ls.deferredHeaderValidation {
		slot, err := slotFromHeaderValidationKey(key)
		// STRICTLY below the cutoff: a block that deep cannot be re-adopted,
		// so an entry still present is abandoned and safe to forget. Anything
		// at or above the cutoff keeps both its entry and its marker so a
		// rollback-then-re-adopt still finds required == true at apply.
		if err != nil || slot < cutoff {
			evicted = append(evicted, key)
```

ledger/chainsync.go
```diff
@@ -408,6 +408,150 @@ func (ls *LedgerState) consumeDeferredHeaderValidation(
...
+// It re-checks each key under deferredHeaderValidationMu and skips any that is
+// present in the in-memory set again: between eviction (which happens for a key
+// whose slot is below the chain tip) and this cleanup, that same point can be
+// re-deferred and re-persisted if it is still ahead of the (lagging) apply
+// cursor -- markDeferredHeaderValidation and persistDeferredHeaderValidation run
+// together on the chainsync path. Deleting the marker for a currently-live
+// entry would drop the durable pin for an active deferred header, so on the
+// next restart repopulate would miss it and cleanup could prune the snapshot it
+// needs (issue #3727, finding: evicted-marker delete races a re-defer).
+func (ls *LedgerState) deletePersistedDeferredMarkers(mapKeys []string) {
```
wolf31o2: Doc update for the delete's TOCTOU window, closed by re-persisting rather than by holding the lock across the delete.

```suggestion
// present in the in-memory set again: between eviction (which happens for a key
// whose slot is below the rollback horizon) and this cleanup, that same point
// can be re-deferred and re-persisted -- markDeferredHeaderValidation and
// persistDeferredHeaderValidation run together on the chainsync path. Deleting
// the marker for a currently-live entry would drop the durable pin for an
// active deferred header, so on the next restart repopulate would miss it and
// cleanup could prune the snapshot it needs (issue #3727, finding:
// evicted-marker delete races a re-defer).
//
// The membership test alone cannot close that window, because the lock is
// released before the delete (see below): a re-defer landing between the test
// and the delete would have its fresh marker deleted, and a restart would then
// find no marker, so deferredHeaderValidationRequired returns false and
// verifyDeferredBlockHeaderState skips the stateful check for a header that is
// still outstanding. So the delete is made effectively conditional by
// re-testing membership AFTER it and RE-PERSISTING the marker for any key that
// came back, rather than by holding the lock across the delete. Re-persisting
// is idempotent (SetSyncState of the same key/value) and runs with no lock
// held, so it closes the window without reintroducing the lock inversion
// (issue #3717 review / cubic P1: re-admission after the live check).
func (ls *LedgerState) deletePersistedDeferredMarkers(mapKeys []string) {
```

ledger/chainsync.go
```diff
@@ -408,6 +408,150 @@ func (ls *LedgerState) consumeDeferredHeaderValidation(
...
+		syncKey := deferredHeaderValidationSyncStatePrefix + k
+		if err := ls.db.DeleteSyncState(syncKey, nil); err != nil {
+			ls.config.Logger.Warn(
+				"failed to delete stale deferred-header marker",
+				"key", syncKey,
+				"error", err,
+				"component", "ledger",
+			)
+		}
+	}
```
wolf31o2, line 613: Extracts the per-key delete into `deleteDeferredMarkerUnlessReadmitted`, which re-tests membership after `DeleteSyncState` returns and restores the marker for a key re-deferred in the window. Neither DB call runs under the mutex, so the lock order against block apply is unchanged. This also makes the restore path deterministically testable instead of timing dependent.

```suggestion
		}
		ls.deleteDeferredMarkerUnlessReadmitted(k)
	}
}

// deleteDeferredMarkerUnlessReadmitted deletes one evicted key's persisted
// marker, then restores it if the key was re-admitted to the in-memory
// deferred set while the delete was in flight.
//
// Neither DB call runs under deferredHeaderValidationMu (that would invert the
// lock order against block apply — issue #3717), so the caller's membership
// test cannot be atomic with the delete. The delete is instead made effectively
// conditional after the fact: re-testing membership and re-persisting is
// idempotent, and it leaves the durable marker present for exactly the keys
// that are live when this returns. Without the restore, a point re-deferred in
// that window keeps its in-memory entry but loses its marker, and after a
// restart repopulateDeferredHeaderValidation misses it — so
// deferredHeaderValidationRequired returns false and
// verifyDeferredBlockHeaderState skips the stateful check for a header that is
// still outstanding.
func (ls *LedgerState) deleteDeferredMarkerUnlessReadmitted(k string) {
	syncKey := deferredHeaderValidationSyncStatePrefix + k
	if err := ls.db.DeleteSyncState(syncKey, nil); err != nil {
		ls.config.Logger.Warn(
			"failed to delete stale deferred-header marker",
			"key", syncKey,
			"error", err,
			"component", "ledger",
		)
		return
	}
	ls.deferredHeaderValidationMu.Lock()
	_, readmitted := ls.deferredHeaderValidation[k]
	ls.deferredHeaderValidationMu.Unlock()
	if !readmitted {
		return
	}
	if err := ls.db.SetSyncState(
		syncKey,
		deferredHeaderValidationSyncStateValue,
		nil,
	); err != nil {
		ls.config.Logger.Warn(
			"failed to restore re-deferred header marker",
			"key", syncKey,
			"error", err,
			"component", "ledger",
		)
	}
```

ledger/verify_header.go
```diff
@@ -1569,6 +1611,161 @@ func (ls *LedgerState) epochForSlot(slot uint64) (models.Epoch, error) {
...
+// and deadlocks the node on the single write connection (issue #3717). Under
+// the lock it, in order:
+//
+//  1. Evicts abandoned deferred headers whose slot the apply cursor has already
+//     passed. A canonical deferred header is consumed when the cursor applies
+//     it, so one still present at/below the tip is on an abandoned fork and
+//     would otherwise pin its snapshot forever (finding 5). Eviction lets the
+//     floor rise; the evicted markers' persisted rows are deleted after the
+//     lock is released (best effort — they cannot affect a resolved header).
+//  2. Computes the retention floor over the surviving deferred headers and
```
wolf31o2, line 1707: Doc update: records that the horizon, not the bare tip, is what makes eviction safe.

```suggestion
//
//  1. Evicts abandoned deferred headers that are beyond the rollback horizon
//     (tip minus the stability window). A canonical deferred header is consumed
//     when the cursor applies it, so one still present that deep is on a fork
//     chain selection can no longer re-adopt and would otherwise pin its
//     snapshot forever (finding 5). The horizon — rather than the bare tip — is
//     what makes eviction safe: eviction also drops the durable marker, and a
//     point evicted while still re-adoptable would apply with required == false
//     and skip its stateful header check. Eviction lets the floor rise; the
//     evicted markers' persisted rows are deleted after the lock is released
//     (best effort — they cannot affect a resolved header).
//  2. Computes the retention floor over the surviving deferred headers and
```

ledger/verify_header.go
```diff
@@ -1569,6 +1611,161 @@ func (ls *LedgerState) epochForSlot(slot uint64) (models.Epoch, error) {
...
+	minBefore uint64,
+	prune func(before uint64) error,
+) error {
+	var evicted []string
+	var before uint64
+	func() {
+		ls.deferredHeaderValidationMu.Lock()
+		defer ls.deferredHeaderValidationMu.Unlock()
+		evicted = ls.evictStaleDeferredHeadersLocked()
+		before = defaultBefore
```
wolf31o2, line 1747: Reads the tip slot and the stability window before taking `deferredHeaderValidationMu`, since `calculateStabilityWindow` takes `ls.RLock` and block apply holds the ls lock before this mutex. Taking it under the mutex would be the same inversion class as the prune path.

```suggestion
) error {
	// Read BEFORE taking deferredHeaderValidationMu: both of these take
	// ls.RWMutex (calculateStabilityWindow reads ls.currentEra under RLock),
	// and block apply holds the ls lock before taking
	// deferredHeaderValidationMu via consumeDeferredHeaderValidation. Taking
	// the ls lock under this mutex would invert that order.
	tipSlot := ls.loadTipSnapshot().currentTip.Point.Slot
	rollbackHorizon := ls.calculateStabilityWindow()
	var evicted []string
	var before uint64
	func() {
		ls.deferredHeaderValidationMu.Lock()
		defer ls.deferredHeaderValidationMu.Unlock()
		evicted = ls.evictStaleDeferredHeadersLocked(tipSlot, rollbackHorizon)
		before = defaultBefore
```

ledger/verify_header_test.go
```diff
@@ -2943,3 +3116,532 @@ func TestVerifyBlockLeaderEligibility_ImportedActivePoolAbsentStaysHardRejection
...
+		25, 16, // minBefore = 16: retain at most down to epoch 16
+		func(before uint64) error { seenBefore = before; return nil },
+	))
+	assert.Equal(
+		t,
+		uint64(16),
+		seenBefore,
+		"floor 10 must be clamped up to the depth-cap minBefore 16",
+	)
+}
```
wolf31o2, line 3763: The two regression tests. `KeepsReadoptableDeferredHeader` covers a header behind the tip but inside the horizon keeping both its pin and its marker while one beyond the horizon is still evicted. `RestoresMarkerReadmittedDuringDelete` covers the restore and the genuinely stale delete. I confirmed both fail against the current behaviour by reverting each half separately, then pass with it.

```suggestion
}

// TestPrunePoolSnapshotsWithRetentionFloor_KeepsReadoptableDeferredHeader is the
// regression guard for the evicted-point re-adoption bypass (issue #3717
// review). Eviction drops the durable sync_state marker along with the
// in-memory entry, and that marker is the only thing that makes
// deferredHeaderValidationRequired return true at apply. A point merely BEHIND
// the tip is still re-adoptable -- chain selection can roll back and switch to
// its fork -- so evicting it there means the block later applies with
// required == false, verifyDeferredBlockHeaderState returns nil, and it is
// adopted with its stateful leader-eligibility check never run. Eviction must
// therefore fire only beyond the rollback horizon (tip minus the stability
// window): a header inside the horizon keeps both its entry and its marker,
// while one past it is still evicted so its snapshot pin is released.
func TestPrunePoolSnapshotsWithRetentionFloor_KeepsReadoptableDeferredHeader(
	t *testing.T,
) {
	tb := createTestBlock(t, [32]byte{73}, 0, tamperNone)
	ls, db := newEligibilityTestLedger(t, tb.epochNonce)
	ls.epochCache = []models.Epoch{
		{
			EpochId:       11,
			StartSlot:     20_000,
			LengthInSlots: 1_000,
			Nonce:         tb.epochNonce,
		},
	}
	// The test config carries no Byron genesis, so the stability window is the
	// 50_000-slot default; with the tip at 60_000 the rollback horizon cuts off
	// at slot 10_000.
	ls.currentTip = ochainsync.Tip{Point: ocommon.Point{Slot: 60_000}}
	ls.publishSnapshotsLocked()
	require.Equal(t, uint64(50_000), ls.calculateStabilityWindow())

	// Behind the tip but INSIDE the horizon: a rollback can still re-adopt it.
	readoptable := ocommon.Point{Slot: 20_500, Hash: []byte{0x11}}
	// Behind the tip and BEYOND the horizon: unreachable, safe to forget.
	unreachable := ocommon.Point{Slot: 5_000, Hash: []byte{0x12}}
	for _, p := range []ocommon.Point{readoptable, unreachable} {
		ls.markDeferredHeaderValidation(p)
		require.NoError(t, ls.persistDeferredHeaderValidation(p, nil))
	}

	var seenBefore uint64
	require.NoError(t, ls.PrunePoolSnapshotsWithRetentionFloor(
		25, 0,
		func(before uint64) error { seenBefore = before; return nil },
	))

	// The re-adoptable header still pins its snapshot epoch (epoch 11 -> 10).
	assert.Equal(
		t,
		uint64(10),
		seenBefore,
		"re-adoptable header must still pin its snapshot",
	)
	floor, ok := ls.OldestRequiredSnapshotEpoch()
	require.True(t, ok, "re-adoptable header must stay in the deferred set")
	assert.Equal(t, uint64(10), floor)

	// ...and it keeps its durable marker, so a rollback-then-re-adopt still
	// finds required == true at apply instead of skipping the check.
	marker, err := db.GetSyncState(
		deferredHeaderValidationSyncStateKey(readoptable), nil,
	)
	require.NoError(t, err)
	assert.Equal(
		t,
		deferredHeaderValidationSyncStateValue,
		marker,
		"re-adoptable header's marker must survive eviction",
	)

	// The unreachable header is still evicted, marker and all.
	gone, err := db.GetSyncState(
		deferredHeaderValidationSyncStateKey(unreachable), nil,
	)
	require.NoError(t, err)
	assert.Empty(
		t,
		gone,
		"header beyond the rollback horizon must still be evicted",
	)
}

// TestDeleteDeferredMarkerUnlessReadmitted_RestoresMarkerReadmittedDuringDelete
// closes the marker delete's TOCTOU window (issue #3717 review / cubic P1). The
// membership test in deletePersistedDeferredMarkers cannot be atomic with the
// delete -- holding deferredHeaderValidationMu across the DB write would invert
// the lock order against block apply and deadlock the node -- so a point
// re-deferred after that test and before the delete would lose the marker it
// had just persisted. A restart would then miss it in
// repopulateDeferredHeaderValidation, leaving deferredHeaderValidationRequired
// false and skipping the stateful check for a header still outstanding. The
// per-key delete must re-test membership afterwards and restore the marker for
// a key that came back, while still removing a genuinely stale one.
func TestDeleteDeferredMarkerUnlessReadmitted_RestoresMarkerReadmittedDuringDelete(
	t *testing.T,
) {
	tb := createTestBlock(t, [32]byte{74}, 0, tamperNone)
	ls, db := newEligibilityTestLedger(t, tb.epochNonce)

	readmitted := ocommon.Point{Slot: 1_150, Hash: []byte{0x11}}
	require.NoError(t, ls.persistDeferredHeaderValidation(readmitted, nil))
	// Stands in for the re-defer that lands after the caller's membership test:
	// the key is live in the set by the time the delete completes.
	ls.markDeferredHeaderValidation(readmitted)

	ls.deleteDeferredMarkerUnlessReadmitted(
		headerValidationPointKey(readmitted),
	)

	marker, err := db.GetSyncState(
		deferredHeaderValidationSyncStateKey(readmitted), nil,
	)
	require.NoError(t, err)
	assert.Equal(
		t,
		deferredHeaderValidationSyncStateValue,
		marker,
		"marker for a key re-admitted during the delete must be restored",
	)

	// A key that is NOT re-admitted still has its marker removed.
	stale := ocommon.Point{Slot: 1_160, Hash: []byte{0x12}}
	require.NoError(t, ls.persistDeferredHeaderValidation(stale, nil))
	ls.deleteDeferredMarkerUnlessReadmitted(headerValidationPointKey(stale))
	gone, err := db.GetSyncState(
		deferredHeaderValidationSyncStateKey(stale), nil,
	)
	require.NoError(t, err)
	assert.Empty(t, gone, "genuinely stale marker must be deleted")
}
```

chadaibbhot-git:
Applied all nine suggestions (thanks @wolf31o2) as a single commit (d617f928): horizon-gated deferred-header eviction, the conditional marker delete (`deleteDeferredMarkerUnlessReadmitted`), reading tip + stability window before taking `deferredHeaderValidationMu`, plus the two regression tests and the ARCHITECTURE.md/DATABASE.md updates. `go test ./ledger/...` is fully green — including `TestPrunePoolSnapshotsWithRetentionFloor_KeepsReadoptableDeferredHeader` and `TestDeleteDeferredMarkerUnlessReadmitted_RestoresMarkerReadmittedDuringDelete` (both confirmed to fail against the prior behaviour). gofmt/vet clean, DCO green. Ready for re-review.

ledger/verify_header.go
```diff
@@ -525,6 +538,25 @@ func (ls *LedgerState) verifyBlockHeaderState(
...
+		// either adopt a block whose leader eligibility is never checked or
+		// loop. The #3727 retention pin is what makes the recoverable case
+		// actually resolve: it retains the mark snapshot a queued/deferred
+		// header needs so that, by the time the cursor reaches the slot, the
+		// snapshot is present and this path is not taken. Deferred headers on
+		// abandoned forks are released by the retention guard's eviction of
+		// entries the cursor has passed (see PrunePoolSnapshotsWithRetentionFloor),
+		// not by deferring their headers forever. A pool absent from a
+		// *populated* snapshot never carries this sentinel and hard-rejects.
 		if allowStateDefer &&
```
wolf31o2, line 560: This relaxes leader-eligibility validation whenever the ledger cursor is behind the header, but the tests only assert the classifier result. Please add an apply-path regression proving the deferred header is eventually revalidated and cannot be adopted without the stateful check. Also document a concrete reference-node acceptance case for the `d=1`/epoch-0 premise; with `d=1`, every slot is an overlay slot and normally requires an active genesis delegate, so the current premise is not yet sufficient to justify suppressing rejection.

ledger/chainsync.go
```diff
@@ -408,6 +408,225 @@ func (ls *LedgerState) consumeDeferredHeaderValidation(
...
+		)
+		return
+	}
+	ls.deferredHeaderValidationMu.Lock()
+	_, readmitted := ls.deferredHeaderValidation[k]
+	ls.deferredHeaderValidationMu.Unlock()
+	if !readmitted {
+		return
+	}
+	if err := ls.db.SetSyncState(
```
wolf31o2: After deleting the persisted marker, a re-admitted point is restored with `SetSyncState`, but a restoration failure is only logged. If the process restarts afterward, the in-memory deferred entry is gone and the snapshot can be pruned, allowing the stateful header check to be skipped. Make restoration retryable/durable, or propagate the cleanup failure so the node cannot continue with a lost retention pin.

ledger/verify_header_test.go
```diff
@@ -2943,3 +3116,664 @@ func TestVerifyBlockLeaderEligibility_ImportedActivePoolAbsentStaysHardRejection
...
+	t *testing.T,
+) {
+	tb := createTestBlock(t, [32]byte{74}, 0, tamperNone)
+	ls, db := newEligibilityTestLedger(t, tb.epochNonce)
+
+	readmitted := ocommon.Point{Slot: 1_150, Hash: []byte{0x11}}
+	require.NoError(t, ls.persistDeferredHeaderValidation(readmitted, nil))
+	// Stands in for the re-defer that lands after the caller's membership test:
+	// the key is live in the set by the time the delete completes.
+	ls.markDeferredHeaderValidation(readmitted)
```
wolf31o2: This test marks the point before calling `deleteDeferredMarkerUnlessReadmitted`, so it does not exercise re-admission during the delete. An implementation that checks membership before deleting and skips the delete would still pass. Synchronize or instrument the delete so `markDeferredHeaderValidation` runs after deletion begins, then assert that the marker is restored.

wolf31o2 (changes requested):
Requesting changes for the unresolved deferred-header validation and retention risks identified in the current-head review.

chadaibbhot-git:
Apply-path proof plus a scope clarification.

**Apply-path (your ask):** added `TestVerifyDeferredBlockHeaderState_GenesisOverlayRevalidatedAtApply` — a deferred header driven through apply: wrong issuer → rejected (marker retained), correct delegate → revalidated + marker cleared, and the missing-marker case (`required==false`) asserted explicitly. Apply re-runs `verifyBlockHeaderStateWithEpochAdvance(…, allowStateDefer=false)` authoritatively, so a deferred header cannot be adopted without the stateful check unless its marker is lost — which this round's TOCTOU/durability hardening (re-tested membership after the delete + propagated restore failure) now prevents.

**On the d=1/epoch-0 premise — scope:** #3717's change is entirely in the leader-stake path. The defer at `verify_header.go:560` is scoped to `errLeaderStakeSnapshotUnavailable && ledgerTipBehindSlot` — a pruned or not-yet-written `pool_stake_snapshot`, not a genesis/overlay gap — plus the retention pin. The earlier epoch-0-sentinel defer you objected to was removed and replaced by this snapshot-availability defer (commit 91a1a6ac). The concrete witness is #3727 itself: a deferred header needed `pool_stake_snapshot` for epochs 10/13/21 while cleanup had pruned everything below currentEpoch-3 (DB retained 25–28) → `leaderEligibilityStake` found no row → producer misclassified as absent → header failure → peer recycled. Once the cursor catches up, a still-empty distribution stays a hard reject, unchanged.

The genesis-overlay defer (`verifyGenesisDelegateHeader`, "genesis overlay state … not yet authoritative") is from #3069 — it is byte-identical between `main` and this branch; #3717 does not add, modify, or depend on it, so this PR introduces no d=1/overlay premise of its own. The single overlay test it adds is the apply-path proof above, defending that existing path per your request. (For completeness I checked both testnets for a literal d=1 overlay-reassignment witness: preview's full d=1 window (epochs 0–1) is original genesis delegates only, and musashi never applied a GenesisKeyDelegation cert — consistent with your own "the two halves cannot both hold on d=1". That path is only exercised on mainnet early-Shelley.)

wolf31o2 (changes requested):
Fresh review at fea67977d4081a316035aa355bf0133b68a1ba6c.\n\n1. Blocking CI: all platform test jobs fail in TestDevNetConformanceAttribution because the new ARCHITECTURE.md section mentions cardano-node without naming --conformance. The docs change must either identify the conformance topology with --conformance or avoid attributing the default DevNet to cardano-node.\n\n2. Consensus evidence remains incomplete: the PR relaxes header rejection for deferred state, but the description explicitly says no concrete reference-node acceptance case has been pinned. The apply-path regression proves Dingo rechecks the marker, but it does not establish that the deferred case is valid for the reference protocol. Please add the concrete network/slot/producer/reference-node case, or narrow/remove the relaxation until that premise is demonstrated.

wolf31o2 (approved):

Issue #3550, opened 2026-08-26
chainselection: peer-activity subscriber stalls and halts tip following


wolf31o2:
## Problem

A long-running Preview node stopped recording chain extensions, then the internal
`chainselection.peer_activity` EventBus subscriber stopped draining. Its buffer
filled and keepalive publisher goroutines accumulated for the rest of the log
while peer reconnects continued.

This is an internal subscriber, not an optional external consumer. On current
`main`, `subscribeChainSelectorEvents` handles the event by calling
`ChainSelector.TouchPeerActivity` and `PeerGov.TouchPeerByConnId`; keepalive
responses publish the event from `ouroboros/keepalive.go`.

## Evidence

- Build: `92b2e3e64e0054ac4176eb92f5db73353a154b24`
- Platform/network: macOS, Preview
- Run interval: 2026-08-25 10:03 through 2026-08-26 10:03 EDT
- The run recorded 1,012 successful `chain extended` messages before its last
  extension at `2026-08-25T18:02:53.336-04:00`, slot `121039373`. No later
  chain extension appears through the end of the log.
- The first stall warning appears at
  `2026-08-26T06:33:48.972-04:00`:

  ```text
  event delivery stalled: subscriber not draining
    type=chainselection.peer_activity buffer=1024 stalled_for=30s
    blocked_publishers=1 suppressed_warnings=0
  ```

- After that first stall warning, no `chain extended`, `selected new best peer`,
  `active connection changed`, or `chain switch: updating active connection`
  message appears through the end of the log.
- The warning repeats 419 times over the remaining 3 hours 29 minutes. The final
  warning reports `blocked_publishers=299` and `suppressed_warnings=298`.
- After the first stall warning, the log records 286 keepalive server-state
  timeouts and repeated successful reconnects, but still no chain-selection or
  chain-extension progress.

The log proves that the subscriber stopped draining and that the node stopped
recording forward progress. It does not contain a goroutine dump, so it does not
establish whether the handler is blocked in chain selection, peer governance,
or a downstream event publication.

## Relationship to existing work

#3529 tracks the EventBus-wide fact that one stalled subscriber can block every
publisher indefinitely. Bounding or isolating back-pressure prevents the 299
parked keepalive publishers, but it does not explain or repair why this required
chain-selection subscriber stopped draining and tip following stopped. This
issue tracks that specific internal consumer failure; the EventBus-wide policy
should remain in #3529.

## Acceptance criteria

[...]

PR #3718, opened 2026-08-31
fix(chain): defer chain-update publication past the ledger mutex to prevent a drain deadlock

chadaibbhot-git:
## What / why
A node syncing a busy chain can freeze with:

```
event delivery stalled: subscriber not draining  type=ledger.chainsync  buffer=100000  stalled_for=30s  blocked_publishers=15
```

The tip stops advancing while the host is ~68% idle (near-zero CPU/IO PSI) — a lock/back-pressure
deadlock, not a throughput limit.

### Root cause
`handleEventBlockfetch` holds `chainsyncBlockfetchMutex` and calls `chain.AddBlock` /
`AddBlockWithPoint` (via `flushPendingBlockfetchBlocks`), which publishes `ChainUpdateEventType`
through a **synchronous, back-pressured** `eventBus.Publish(...)` — *while still holding the mutex*.
If the terminal lane stops draining, that `Publish` parks holding `chainsyncBlockfetchMutex`;
`handleEventChainsync` then blocks acquiring the same mutex, stops draining the `ledger.chainsync`
queue, its 100000 buffer fills, and the ouroboros publishers park. Everything is on locks — hence
the idle host.

## Fix — defer the publication past the ledger mutex (not `PublishOrdered`)
The publish happens while the **ledger** holds its mutex, so only the ledger can move it out of the
lock. This PR uses the codebase's existing deferral mechanism, `ledger/pending_publish.go`'s
`pendingPublishes` (events collected under the lock, `Publish`ed from `flush()` which is `defer`red
*before* the lock so it runs *after* the unlock):

- `chain/chain.go` adds `AddBlockWithPointDeferred` / `RollbackDeferred` that **return** the
  chain-update event(s) instead of publishing them.
- `ledger/chainsync.go` / `ledger/state.go`: the blockfetch drain and rollback hand those events to
  the handler's `pendingPublishes` queue, which flushes them **after** `chainsyncMutex` /
  `chainsyncBlockfetchMutex` are released.
- The six standalone `chain.go` sites (forging, bulk import) keep plain `Publish` — they run on the
  scheduler goroutine, **not** under a ledger mutex, so a back-pressured publish there cannot cause
  the deadlock.

An earlier revision routed everything through `PublishOrdered`; that was **rejected** because the
ordered lane is itself bounded (10,000), so a stalled subscriber merely postpones the same deadlock
until the lane fills. Deferring past the mutex removes the hazard entirely: no back-pressured
publish ever runs while a ledger mutex is held. Per-type rollback-before-add ordering is preserved
because every chain-update — deferred or standalone — is delivered through `Publish` (in-order per
goroutine), and each handler threads a single `pendingPublishes` queue.

## Test
`ledger/blockfetch_drain_deadlock_test.go` fills the terminal chain.update lane to capacity on top of
a stalled lossless subscriber, runs the real drain, and asserts it returns promptly and deferred (did
not publish) the event. It **times out against both** the original inline `Publish` **and** inline
`PublishOrdered`, and passes only with the deferred fix. `chain/deferred_publish_test.go` covers the
`*Deferred` return/ordering. `go build/vet/test ./chain/... ./ledger/...`, `golangci-lint`, and
`gofmt` all pass.

## Related
- #3529 (EventBus-wide back-pressure) / #3550 (a sibling stalled-subscriber, the
  `chainselection.peer_activity` lane) — this PR removes one concrete instance (the `AddBlock` →
  `chain.update` publish under the ledger mutex).

### DIFF
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -205,11 +205,17 @@ func (c *Chain) AddBlock(
 	block ledger.Block,
 	txn *database.Txn,
 ) error {
-	evt, err := c.addBlockInternal(block, ocommon.Point{}, txn, true, true)
+	evt, err := c.addBlockInternal(block, ocommon.Point{}, txn, true)
 	if err != nil {
 		return err
 	}
-	// Publish event immediately for standalone (non-batched) calls
+	// Publish event immediately for standalone (non-batched) callers.
+	// forgeBlock reaches this on the scheduler goroutine, not under a ledger
+	// mutex, so a backpressured Publish here cannot stall the chainsync
+	// drain. The mutex-holding blockfetch drain uses AddBlockWithPointDeferred
+	// instead, which returns the event for the ledger to publish after it
+	// releases chainsyncBlockfetchMutex. See ledger.pendingPublishes and the
+	// blinklabs-io/dingo drain deadlock.
 	if c.eventBus != nil && evt.Type != "" {
 		c.eventBus.Publish(ChainUpdateEventType, evt)
 	}
@@ -224,7 +230,6 @@ func (c *Chain) AddLocalBlock(block ledger.Block) error {
 		block,
 		ocommon.Point{},
 		nil,
-		true,
 		false,
 	)
 	if err != nil {
@@ -244,7 +249,7 @@ func (c *Chain) AddBlockWithPoint(
 	point ocommon.Point,
 	txn *database.Txn,
 ) error {
-	evt, err := c.addBlockInternal(block, point, txn, true, true)
+	evt, err := c.addBlockInternal(block, point, txn, true)
 	if err != nil {
 		return err
 	}
@@ -252,6 +257,25 @@ func (c *Chain) AddBlockWithPoint(
 		c.eventBus.Publish(ChainUpdateEventType, evt)
 	}
 	return nil
+}
+
+// AddBlockWithPointDeferred adds a block exactly like AddBlockWithPoint but
+// returns the resulting chain.update event instead of publishing it, leaving
+// publication to the caller. The ledger's chainsync/blockfetch drain calls
+// this while holding chainsyncBlockfetchMutex and queues the event on its
+// pendingPublishes, so the (potentially backpressured) delivery runs only
+// after the mutex is released. Publishing inline under that mutex is what
+// deadlocked the node: a terminal chain.update subscriber that stopped
+// draining parked the publish with the mutex held, handleEventChainsync then
+// blocked on the same mutex, and the ledger.chainsync buffer filled
+// (blinklabs-io/dingo preview freeze). A returned event with an empty Type
+// means there is nothing to publish.
+func (c *Chain) AddBlockWithPointDeferred(
+	block ledger.Block,
+	point ocommon.Point,
+	txn *database.Txn,
+) (event.Event, error) {
+	return c.addBlockInternal(block, point, txn, true)
 }
 
 // addBlockInternal performs all block-adding logic but returns the event
@@ -262,7 +286,6 @@ func (c *Chain) addBlockInternal(
 	block ledger.Block,
 	point ocommon.Point,
 	txn *database.Txn,
-	notifyWaiters bool,
 	matchPendingHeader bool,
 ) (event.Event, error) {
 	if c == nil {
@@ -281,7 +304,7 @@ func (c *Chain) addBlockInternal(
 		block,
 		point,
 		txn,
-		notifyWaiters,
+		true,
 		matchPendingHeader,
 	)
 }
@@ -435,7 +458,9 @@ func (c *Chain) AddBlocks(blocks []ledger.Block) error {
 			return err
 		}
 		c.notifyWaitingIterators()
-		// Transaction committed successfully; publish all events
+		// Transaction committed successfully; publish all events. This
+		// bulk-import path (internal/node/load) does not run under a ledger
+		// mutex, so an inline Publish here cannot stall the chainsync drain.
 		if c.eventBus != nil {
 			for _, evt := range pendingEvents {
 				c.eventBus.Publish(ChainUpdateEventType, evt)
@@ -666,7 +691,8 @@ func (c *Chain) addRawBlocks(
 			return fmt.Errorf("add raw block batch: %w", err)
 		}
 		c.notifyWaitingIterators()
-		// Publish events (only when eventBus is set).
+		// Publish events (only when eventBus is set). Not reached under a
+		// ledger mutex, so an inline Publish is safe here.
 		if c.eventBus != nil {
 			for _, evt := range pendingEvents {
 				c.eventBus.Publish(
@@ -697,13 +723,36 @@ func (c *Chain) Rollback(point ocommon.Point) error {
 		return err
 	}
 	// Publish events after locks are released to prevent deadlocks
-	// when subscribers call back into chain/manager state.
+	// when subscribers call back into chain/manager state. The
+	// mutex-holding ledger rollback path uses RollbackDeferred instead,
+	// which returns these events for the ledger to publish after it
+	// releases chainsyncMutex.
 	if c.eventBus != nil {
 		for _, evt := range pendingEvents {
 			c.eventBus.Publish(evt.Type, evt)
 		}
 	}
 	return nil
+}
+
+// RollbackDeferred rewinds the chain exactly like Rollback but returns the
+// resulting chain.update events instead of publishing them, leaving
+// publication to the caller. The ledger's rollbackChainAndState calls this
+// while holding chainsyncMutex and queues the events on its pendingPublishes,
+// so delivery (which can backpressure on a full subscriber buffer) runs only
+// after the mutex is released. Publishing inline under chainsyncMutex risks
+// the same drain deadlock described on AddBlockWithPointDeferred. The events
+// are returned in rollback order; a caller that queues them on one
+// pendingPublishes keeps this rollback's chain.update ahead of the block-add
+// chain.update events that follow it, which is the ordering rollbacks depend
+// on.
+func (c *Chain) RollbackDeferred(
+	point ocommon.Point,
+) ([]event.Event, error) {
+	if c == nil {
+		return nil, errors.New("chain is nil")
+	}
+	return c.rollbackLocked(point)
 }
 
 // rollbackForkDepth returns the number of blocks a rollback to
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -59,6 +59,33 @@ type Chain struct {
 	mutex                sync.RWMutex
 	waitingChanMutex     sync.Mutex
 	persistent           bool
+
+	// pendingUpdates is the chain-level sequencer for deferred chain.update
+	// / chain.fork publication. The mutex-holding ledger paths (blockfetch
+	// add under chainsyncBlockfetchMutex, rollback under chainsyncMutex) must
+	// not publish inline -- a back-pressured Publish there deadlocks the node
+	// (see ledger.pendingPublishes). They instead enqueue the event onto this
+	// single shared queue *atomically with the chain mutation, under c.mutex*
+	// (queueDeferredEventLocked), and drain it after releasing their outer
+	// mutex (PublishPendingChainUpdates).
+	//
+	// Because enqueue happens under c.mutex -- the same lock that serializes
+	// every chain mutation -- enqueue order equals mutation order, and a
+	// single FIFO drain therefore publishes in mutation order across BOTH
+	// handlers. A blockfetch add and a chainsync rollback can no longer invert
+	// (add queued, rollback published first) the way two independent
+	// per-handler queues could, which would have driven the chain.update
+	// subscriber's block apply/undo notifications out of order.
+	//
+	// pendingUpdatesMutex guards the slice for the brief enqueue/dequeue; it
+	// is never held across a Publish. publishMutex serializes drains so that,
+	// when two goroutines flush concurrently, their Publish calls stay in
+	// pop (FIFO) order rather than interleaving. Neither is a ledger mutex, so
+	// draining -- which only ever runs after the caller has released its outer
+	// ledger mutex -- cannot reintroduce the drain deadlock.
+	pendingUpdates      []event.Event
+	pendingUpdatesMutex sync.Mutex
+	publishMutex        sync.Mutex
 }
 
 type queuedHeader struct {
@@ -205,7 +232,7 @@ func (c *Chain) AddBlock(
 	block ledger.Block,
 	txn *database.Txn,
 ) error {
-	evt, err := c.addBlockInternal(block, ocommon.Point{}, txn, true)
+	evt, err := c.addBlockInternal(block, ocommon.Point{}, txn, true, false)
 	if err != nil {
 		return err
 	}
@@ -231,6 +258,7 @@ func (c *Chain) AddLocalBlock(block ledger.Block) error {
 		ocommon.Point{},
 		nil,
 		false,
+		false,
 	)
 	if err != nil {
 		return err
@@ -249,7 +277,7 @@ func (c *Chain) AddBlockWithPoint(
 	point ocommon.Point,
 	txn *database.Txn,
 ) error {
-	evt, err := c.addBlockInternal(block, point, txn, true)
+	evt, err := c.addBlockInternal(block, point, txn, true, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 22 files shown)


chain/chain.go
```diff
@@ -209,9 +209,17 @@ func (c *Chain) AddBlock(
...
+	// while holding chainsyncBlockfetchMutex; a synchronous Publish here parks
+	// on a full terminal subscriber buffer WITH that ledger mutex held, which
+	// deadlocks handleEventChainsync against the same mutex and fills the
+	// ledger.chainsync buffer (blinklabs-io/dingo freeze). PublishOrdered hands
+	// the event to the single-worker per-type lane and returns without waiting
+	// on subscriber capacity, so no backpressured delivery runs under a ledger
+	// mutex, while the one worker per event type preserves chain.update order.
 	if c.eventBus != nil && evt.Type != "" {
-		c.eventBus.Publish(ChainUpdateEventType, evt)
+		c.eventBus.PublishOrdered(ChainUpdateEventType, evt)
```
wolf31o2: **P1 — This only postpones the deadlock until the ordered lane fills.** `PublishOrdered` is not non-blocking: its documented contract waits when the 10,000-entry lane is full, and `TestPublishOrderedWaitsForCapacityRatherThanDropping` enforces that. Once the stalled lossless subscriber parks the lane worker, 10,000 more events fill the lane and the next `AddBlock` blocks here under `chainsyncBlockfetchMutex`, recreating the same cycle. The new test publishes only six blocks, so it cannot reach this path. Please move publication outside the ledger lock (consistent with the existing pending-publish invariant), or use another lifecycle-bounded design that preserves lossless ordering, and add a saturation regression.

wolf31o2 (changes requested):
Requesting changes for the bounded-lane deadlock described inline. All current checks pass, but they do not cover queue saturation.

chadaibbhot-git:
You're right — `PublishOrdered` only postpones it (bounded 10k lane, and the lane worker parks under the same mutex). Reworked to defer the `ChainUpdateEventType` publication past the ledger lock via the existing `pendingPublishes` queue (collected under the lock, flushed from `pending.flush()` after unlock); the six `chain.go` sites are back to plain `Publish` for the standalone callers (forging, bulk import — none hold a ledger mutex).

New `ledger/blockfetch_drain_deadlock_test.go` fills the chain.update lane to capacity on top of a stalled lossless subscriber, then runs the real drain and asserts it returns promptly and deferred (did not publish) the event. It times out against both the original inline `Publish` and inline `PublishOrdered`, and passes with the deferred fix. Rollback-before-add ordering preserved. Force-pushed the rework.

chain/chain.go
```diff
@@ -205,11 +205,17 @@ func (c *Chain) AddBlock(
 	block ledger.Block,
 	txn *database.Txn,
 ) error {
-	evt, err := c.addBlockInternal(block, ocommon.Point{}, txn, true, true)
+	evt, err := c.addBlockInternal(block, ocommon.Point{}, txn, true)
 	if err != nil {
 		return err
 	}
-	// Publish event immediately for standalone (non-batched) calls
+	// Publish event immediately for standalone (non-batched) callers.
```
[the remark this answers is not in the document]
chadaibbhot-git: This is intentional and follows @wolf31o2's review. `PublishOrdered` was rejected because its 10,000-entry lane also blocks when full, so it doesn't remove the deadlock. Instead this moves the synchronous `Publish` *out* from under the ledger mutex: standalone (forge) calls publish on the scheduler goroutine with no ledger mutex held, and batched calls defer publication until after the mutex is released (`pendingPublishes`, flushed in order). A backpressured subscriber can then stall only the calling goroutine — never the ledger / chainsync drain — and ordering is preserved. Switching to `PublishOrdered` would reintroduce exactly the bounded-lane blocking flagged earlier.

chain/chain.go
chadaibbhot-git: Reworked in `c1b0ddd2`: dropped `PublishOrdered` entirely per your P1 and moved chain-update publication past the ledger mutex. Standalone/forge publishes run off the scheduler goroutine (no ledger mutex held); batched calls defer to `pendingPublishes` and flush in order after the mutex releases — so a stalled subscriber can no longer block the ledger drain. Would appreciate a re-review when you have a moment.

wolf31o2:
@chadaibbhot-git you need to follow the rest of cubic's review advice

wolf31o2 (changes requested):
Fix bot review findings

chadaibbhot-git:
Addressed the remaining cubic points on the current head:

- **Test cleanup** (`blockfetch_drain_deadlock_test.go`): the two publisher goroutines the test parks were torn down by an unbounded `t.Cleanup(eventBus.Stop)`; if `Stop`'s backpressure-release ever regressed, that would hang the whole `go test` binary with no diagnostic. Wrapped it in a bounded 10s wait that `t.Error`s on timeout, so a shutdown regression fails the test instead of hanging CI.
- **PR-body / ordering wording**: this was flagged against an earlier revision — the current commit message and PR body already state the fix defers past the ledger mutex via `pendingPublishes` (not `PublishOrdered`), and qualify the per-type ordering as resting on in-order `Publish` per goroutine + one `pendingPublishes` queue per handler.
- **PublishOrdered at the six standalone sites**: leaving as-is, per the earlier discussion — those sites don't run under a ledger mutex, and `PublishOrdered`'s 10k lane is itself bounded, so it would only postpone the same drain deadlock. The fix deliberately keeps plain `Publish` there and defers only the mutex-holding drain/rollback paths.

`go test ./chain/... ./ledger/...`, `go vet`, and `gofmt` all pass.

ledger/state.go
```diff
@@ -3212,17 +3223,30 @@ func (ls *LedgerState) rollbackChainAndState(point ocommon.Point) error {
...
 	if err != nil {
 		return err
 	}
+	// Publish the rollback's chain.update after chainsyncMutex is released.
+	// Under the mutex an inline back-pressured publish deadlocks the drain;
+	// queueing on pubs (nil publishes immediately on the unlocked path) moves
+	// it past the unlock while preserving rollback-before-add ordering.
+	for _, evt := range rollbackEvents {
+		pubs.add(ls.config.EventBus, evt.Type, evt)
+	}
```
wolf31o2, line 3256: This half of the fix is untested. I replaced this loop with the pre-PR inline `ls.config.EventBus.Publish(evt.Type, evt)` — what `chain.Rollback` did while `chainsyncMutex`, `transactionEventMutex` and `blockPipelineGatherMutex`'s write lock were held — and ran the full package:

```
ok  github.com/blinklabs-io/dingo/ledger  133.328s   (exit 0)
```

Nothing detects it. Per `ARCHITECTURE.md:8489` this path holds the write lock from before `drainBlockPipelineBeforeRollback` through `ls.chain.Rollback`, so it is the more dangerous of the two instances the PR fixes.

The converted rollback tests can't cover it: they all pass `nil`, which takes `pendingPublishes.add`'s immediate-publish branch. A test shaped like `TestBlockfetchDrainDefersChainUpdatePastLedgerMutex`, driving `rollbackChainAndStateDeferred(point, &pubs)` against a saturated `chain.update` lane, would.

ledger/chainsync.go
```diff
@@ -4328,12 +4347,18 @@ func (ls *LedgerState) flushPendingBlockfetchBlocks() error {
...
+		evt, addBlockErr := ls.chain.AddBlockWithPointDeferred(
 			pendingEvent.Block,
 			pendingEvent.Point,
 			nil,
 		)
 		if addBlockErr == nil {
+			// Defer this block's chain.update past chainsyncBlockfetchMutex
+			// rather than publishing inline. See
+			// flushPendingBlockfetchBlocksDeferred and pendingPublishes.
+			if evt.Type != "" {
```
wolf31o2, line 4364: The reordering window this widens needs either an argument or a guard on the PR.

The body says "every chain-update — deferred or standalone — is delivered through `Publish` (in-order per goroutine)". That holds within a handler goroutine, but `handleEventBlockfetch` holds only `chainsyncBlockfetchMutex` and `handleEventChainsync` holds only `chainsyncMutex`, and they do not exclude each other — `handleEventChainsyncRollback` never takes the blockfetch mutex, and `tryResolveFork` releases it at 3460 before rolling back at 3475.

So the drain can mutate the chain for block N, queue `Apply(N)`, then do up to seven more deferred adds — `blockfetchCommitBatchSize = 8`, each its own commit — before unlocking and flushing, while a concurrent rollback truncates below N and flushes first. `chain.update` then carries `Rollback` ahead of `Apply(N)`.

`handleEventChainUpdate` is a single FIFO subscriber, so `ledger.block` inherits that order. `midnight/indexer/indexer.go:705-742` applies without a sequence check and then advances its checkpoint past a rolled-back block.

This invariant was never enforced, only improbable — on main the gap was a few instructions. The PR widens it to a batch of DB commits. Not reproduced with a two-goroutine harness, so I'm calling it plausible rather than confirmed, but it should not go in unmentioned behind a claim that ordering is preserved.

ledger/blockfetch_drain_deadlock_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// THE FIX. flushPendingBlockfetchBlocksDeferred runs on the mutex-holding
+	// drain path. It must add the block and return promptly, queueing the
+	// chain.update on pubs instead of publishing it into the saturated bus.
+	// Under the original synchronous Publish, or the rejected under-lock
+	// PublishOrdered, this call would block forever on the stalled subscriber
+	// / full lane and the timeout below would fire.
+	var pubs pendingPublishes
+	drained := make(chan error, 1)
+	go func() { drained <- ls.flushPendingBlockfetchBlocksDeferred(&pubs) }()
```
wolf31o2, line 180: This calls the drain directly, without holding `chainsyncBlockfetchMutex` and without going through `handleEventBlockfetch`. It proves the drain does not publish inline, but the fix is that the flush runs *after* the mutex is released, and the load-bearing line for that is the `defer pending.flush()`-before-`Lock()` ordering at `chainsync.go:502-505`. Swap those two statements and this test still passes.

Driving `ls.handleEventBlockfetch(evt)` on the saturated bus and asserting that a concurrent `chainsyncBlockfetchMutex` acquirer makes progress would test the actual boundary, and would be the true end-to-end deadlock guard.

Minor, same file: the raw `time.After`/`require.Eventually` here differ from the package convention — `block_event_ordering_test.go` uses `testutil.RequireReceive`/`RequireNoReceive` for exactly these positive and negative waits.

chain/chain.go
```diff
@@ -706,6 +735,26 @@ func (c *Chain) Rollback(point ocommon.Point) error {
 	return nil
 }
 
+// RollbackDeferred rewinds the chain exactly like Rollback but returns the
+// resulting chain.update events instead of publishing them, leaving
+// publication to the caller. The ledger's rollbackChainAndState calls this
```
wolf31o2: This comment, added by this PR, refers to `rollbackChainAndState`, which the PR deletes. Same stale name as the seven `ARCHITECTURE.md` references and about fifteen other Go comments across `ledger/state.go`, `ledger/chainsync.go`, `ledger/continuation_audit.go` and five test files.

While here: `chain` now exposes six add-block entry points and two rollback entry points, with publish-versus-don't-publish distinguished only by a `Deferred` suffix and a doc comment, and `AddBlockWithPoint` is dead in production. The dual API is what made this bug possible; making `AddBlockWithPoint`/`Rollback` always return their events, with publication always the caller's job, would remove the footgun rather than add a second safe door beside it.

wolf31o2 (changes requested):
Design is right; three gaps block it.

1. The `chainsyncMutex` half is untested. Reverting `ledger/state.go:3247-3249` to an inline publish leaves `go test ./ledger/` green — `ok 133.328s`, exit 0. Per `ARCHITECTURE.md:8489` that path holds three locks across the publish. The converted rollback tests all pass `nil`, taking `pendingPublishes.add`'s immediate branch, so none can distinguish deferred from inline.

2. `ARCHITECTURE.md` unchanged; it names the deleted `rollbackChainAndState` at 510, 3433, 3548, 8431, 8440, 8456, 8489. Line 510 states the invariant this PR changes. `docs-parity` only checks go.mod, the Makefile and the compose file. ~15 Go comments carry the same stale names, including one this PR adds.

3. `TestNoEventBusPublishWhileHoldingChainsyncMutex` still cannot see `ls.chain.AddBlockWithPoint(...)` under a guarded mutex — it matches `Sel.Name == "EventBus"`. That method now has zero production callers but still publishes inline. Add the inline-publishing `chain` methods to the forbidden set.

Both new tests fail without the fix (hang on timeout, 10.57s and 5.00s). cubic's `PublishOrdered` thread is a false positive — `OrderedQueueSize` is bounded — and needs dismissing.

chadaibbhot-git:
Thanks — all three addressed on the branch (note line numbers below are against the current tip; a couple of yours were from an earlier revision).

**1. chainsyncMutex deferred-publish now has teeth.** Added `ledger/chainsync_resync_deferred_publish_test.go` — two tests that drive `requestChainsyncResync` with a real (non-nil) `pendingPublishes` while holding a guarded mutex, against a lossless `ChainsyncResyncEventType` subscriber whose handler takes that same mutex. Unlike the converted rollback tests (all `nil` → the immediate branch), these exercise the deferred path: with the fix they pass; revert the `pending.add(...)` in `requestChainsyncResync` to an inline `EventBus.Publish(...)` and both hang to their 5s guard (~5.0s each). One holds `chainsyncMutex`, the other `chainsyncBlockfetchMutex`. (In this tree that deferred publish lives in `chainsync.go`, not `state.go:3247-3249` — those look like earlier line numbers.)

**2. Docs/comments.** On the current tip `rollbackChainAndState` is still a live method (`ledger/state.go:3241`, called from `chainsync.go:2003`/`:3477`), and every reference in ARCHITECTURE.md + the Go comments resolves to it and matches current behavior — including the invariant that it's the one deliberate publish-under-lock exception, reached through the `pendingPublishes` chain. So there were no stale "deleted `rollbackChainAndState`" names to rename here; your line numbers (e.g. 8489) point at an earlier ARCHITECTURE.md revision. If you were reviewing a state where that method was actually removed, point me at it and I'll reconcile — otherwise this one already looks consistent. (I did add a line documenting the extended static guard from #3.)

**3. Static guard extended.** `TestNoEventBusPublishWhileHoldingChainsyncMutex` now treats an `ls.chain.<method>` call as a publish for the inline-publishing `chain.Chain` methods (`AddBlock`, `AddLocalBlock`, `AddBlockWithPoint` and siblings, plus `Rollback`) — they emit chain-update/rollback events from inside the chain package, which the `EventBus`-receiver match couldn't see. No current lock holder trips it; a probe holding `chainsyncMutex` across `ls.chain.Rollback` is caught as expected.

**cubic's `PublishOrdered` thread — false positive.** That's the `ledger.tx` emit, the documented publish-under-lock exception. It can't close the deadlock the queue fix targets: that cycle needs the event's *subscriber* to re-enter the publisher's mutex, and `TransactionEventType` subscribers derive state without taking `chainsyncMutex`/`chainsyncBlockfetchMutex`. `OrderedQueueSize` is bounded (10000) by design — past it the publisher applies backpressure, not a lock cycle — and the publish uses `PublishOrderedContext(ls.publishCtx, …)`, cancelled at Close, so a parked publisher is released at shutdown. No change needed.

ledger/chainsync.go
```diff
@@ -528,7 +528,7 @@ func (ls *LedgerState) handleEventBlockfetch(evt event.Event) {
 			)
 		}
 	} else if e.Block != nil {
-		if err := ls.handleEventBlockfetchBlock(e); err != nil {
+		if err := ls.handleEventBlockfetchBlockDeferred(e, &pending); err != nil {
```
wolf31o2, line 531: This queue only preserves order within one pendingPublishes instance. The blockfetch handler and chainsync rollback handler have different queues and can flush concurrently after releasing their respective mutexes. For example, a blockfetch add can mutate the chain and queue its update, then a rollback can mutate the chain and publish its rollback update before the add queue flushes; the chain-update subscriber can therefore observe rollback-before-add for the opposite of the actual chain mutation order. Because that subscriber drives block apply/undo notifications, this can leave derived state inconsistent. Please serialize deferred chain.update publication with chain mutation order (or introduce a shared chain-level sequencer) and add a deterministic concurrent regression test.

wolf31o2 (changes requested):
Requesting changes: deferred publication removes the deadlock, but chain.update ordering is still not serialized across concurrent handlers.

chain/chain.go
```diff
@@ -706,6 +735,26 @@ func (c *Chain) Rollback(point ocommon.Point) error {
...
+// while holding chainsyncMutex and queues the events on its pendingPublishes,
+// so delivery (which can backpressure on a full subscriber buffer) runs only
+// after the mutex is released. Publishing inline under chainsyncMutex risks
+// the same drain deadlock described on AddBlockWithPointDeferred. The events
+// are returned in rollback order; a caller that queues them on one
+// pendingPublishes keeps this rollback's chain.update ahead of the block-add
+// chain.update events that follow it, which is the ordering rollbacks depend
+// on.
```
wolf31o2: This is the claim that does not hold across handlers, so it is worth scoping in the comment itself: a reader who takes it at face value will believe the queue provides ordering it cannot provide.

```suggestion
// are returned in rollback order. Queueing them on one pendingPublishes only
// orders them against events queued on THAT SAME queue, which means within a
// single handler invocation. It does NOT order them against a concurrent
// handler's queue: handleEventBlockfetch (chainsyncBlockfetchMutex) and
// handleEventChainsync (chainsyncMutex) dispatch on separate EventBus
// goroutines and flush independently, so an add that mutated the chain first
// can still be published after this rollback. Cross-handler ordering needs a
// chain-level sequencer, not this queue.
```

wolf31o2 (commented):
Confirming the chain.update ordering gap on head 6725f370. I traced the mechanism independently rather than taking the earlier review at face value, and it holds. Two inline suggestions correct comments that currently overstate the guarantee; the substantive fix is sketched below.

What I verified locally:

- `ChainsyncEventType` (`ledger/state.go:1508`) and `BlockfetchEventType` (`state.go:1616`) are separate subscriptions, so `handleEventChainsync` and `handleEventBlockfetch` dispatch on separate goroutines.
- They hold different mutexes (`chainsyncMutex` vs `chainsyncBlockfetchMutex`), and each builds its own `pendingPublishes`, flushed after its own unlock.
- Serialising the chain mutation would not fix it, because the flush happens after the unlock. A blockfetch add can mutate and unlock, a rollback can then mutate, unlock, and flush first, so the add's event is published second.
- `ledger/block_event.go:31-48` turns chain.update into `BlockActionApply` / `BlockActionUndo`, so a subscriber can be handed an Apply for a block the truncation already removed, after the Undos for that truncation.

Intra-handler ordering is already correct: `tryResolveFork` threads one queue through `restartQueuedBlockfetchAfterForkLocked` into `flushPendingBlockfetchBlocksDeferred`, so a rollback and the adds that follow it in the same invocation stay ordered. The gap is exactly the blockfetch handler's queue against the chainsync handler's queue, which is why a per-handler queue cannot close it.

Suggested fix: make the queue chain-level instead of handler-level. Appends happen where the mutation happens, under `c.mutex`, which both `rollbackLocked` (chain.go:959) and the `addBlockInternal` path (chain.go:294) already hold, so append order equals mutation order globally. A drain mutex serialises concurrent drainers so whoever gets there first publishes the oldest event first:

```go
// Called from rollbackLocked / addBlockInternal, with c.mutex held.
func (c *Chain) queueChainUpdateLocked(evt event.Event) {
	if evt.Type == "" {
		return
	}
	c.pendingUpdates = append(c.pendingUpdates, evt)
}

// DrainChainUpdates publishes queued chain.update events in mutation order.
// Callers invoke it AFTER releasing their ledger mutex, so no ledger lock is
// ever held across a back-pressured publish, which is the property this PR
// establishes. drainMu serialises concurrent drainers.
func (c *Chain) DrainChainUpdates() {
	c.drainMu.Lock()
	defer c.drainMu.Unlock()
	for {
		c.mutex.Lock()
		if len(c.pendingUpdates) == 0 {
			c.mutex.Unlock()
			return
		}
		evt := c.pendingUpdates[0]
		c.pendingUpdates = c.pendingUpdates[1:]
		c.mutex.Unlock()
		if c.eventBus != nil {
			c.eventBus.Publish(evt.Type, evt)
		}
	}
}
```

The handlers then replace `defer pending.flush()` with `defer ls.chain.DrainChainUpdates()` for chain.update specifically, and `AddBlockWithPointDeferred` / `RollbackDeferred` stop returning events. `pendingPublishes` stays for the non-chain.update lanes it already serves.

Tradeoff worth stating: a drainer parked on a stalled chain.update subscriber holds `drainMu` and blocks the other drainer. That is a stall, not a deadlock, because no ledger mutex is held at that point, so this keeps the deadlock this PR closes closed. A permanently stalled terminal subscriber remains a liveness problem in its own right, which is #3529's scope, not this PR's.

For a deterministic regression test, the two handlers can be driven concurrently with the chain.update lane held at capacity, releasing it only after both have mutated, then asserting the delivered order matches the mutation order rather than the flush order.

Two things I checked that are not defects, recorded so they do not get re-litigated:

- Dropping the `notifyWaiters` parameter from `addBlockInternal` is behaviour preserving. All three former callers passed `true`, and `AddLocalBlock` still passes `matchPendingHeader` false.
- Moving the publish out of the `transactionEventMutex` section at `state.go:3223` does not break the invariant that comment describes. `validateAndEmitRollbackUndo` and the truncation both remain inside it; only publication moved out.

One residual outside this PR's scope: `replay_recovery.go:760` and `:772` (`ls.chain.Rollback`) and `state.go:10079` (`ls.chain.AddBlock` in `forgeBlock`) still publish inline. I traced the first two to `state.go:5375/5785/5804` in the block-apply pipeline and the third to the scheduler goroutine, so neither runs under a chainsync mutex and the PR's reasoning about them holds. They do publish from a goroutine that can hold the ledger write lock, so a stalled chain.update subscriber can still park it. Different lane from this fix, worth a follow-up rather than a change here.

Filed as a comment rather than another change request, since the existing request already covers the ordering issue.

ledger/state.go
```diff
@@ -3212,17 +3223,30 @@ func (ls *LedgerState) rollbackChainAndState(point ocommon.Point) error {
...
+		rollbackEvents = evts
+		return nil
 	}()
 	if err != nil {
 		return err
 	}
+	// Publish the rollback's chain.update after chainsyncMutex is released.
+	// Under the mutex an inline back-pressured publish deadlocks the drain;
+	// queueing on pubs (nil publishes immediately on the unlocked path) moves
+	// it past the unlock while preserving rollback-before-add ordering.
```
wolf31o2: Same overclaim as the `RollbackDeferred` doc comment. Deferring past the unlock is what removes the deadlock, but "preserving rollback-before-add ordering" only holds for adds queued on this same `pubs`, not for a concurrent `handleEventBlockfetch` queue.

```suggestion
	// Publish the rollback's chain.update after chainsyncMutex is released.
	// Under the mutex an inline back-pressured publish deadlocks the drain;
	// queueing on pubs (nil publishes immediately on the unlocked path) moves
	// it past the unlock. This keeps rollback-before-add ordering only for
	// adds queued on this same pubs, i.e. within this handler invocation; a
	// concurrent handleEventBlockfetch flush is not ordered against it.
```

wolf31o2 (approved):
Reviewed current head `9504bb2b`. No additional findings.

PR #3726, opened 2026-08-31
fix(ledger): stop and drain consumed-UTxO cleanup during shutdown

### LINKED ISSUE #3439: ledger: stop consumed-UTxO cleanup during shutdown
The consumed-UTxO cleanup timer in `ledger/state.go` is not stopped during
`Close()`. The timer can continue to access the database after the ledger
state has been closed.

Acceptance criteria:

- Stop and drain the cleanup timer as part of ledger-state shutdown.
- Ensure no cleanup callback can begin database work after `Close` returns.
- Add lifecycle tests covering close during an active timer callback and
  repeated close calls.

chrisguiney:
## What / why

`LedgerState.Close` never stopped the consumed-UTxO cleanup timer. The timer
callback re-arms itself via `scheduleCleanupConsumedUtxos`, so after `Close`
returned it kept firing every 5 minutes for the rest of the process, calling
`ls.db.StorageMode()` and `ls.db.UtxosDeleteConsumed(...)`. `LedgerState` does
not own the database (see the note at the end of `Close`); its owner closes it
as soon as `Close` returns.

Three separate gaps:

1. The timer is never stopped, and re-arms itself.
2. `time.Timer.Stop` does not wait for an `AfterFunc` callback that has already
   fired, so stopping alone would still let `Close` return mid-delete.
3. The epoch transition triggers cleanup as a bare `go ls.cleanupConsumedUtxos()`
   (`ledger/state.go`), which stopping the timer does not reach.

The existing `ls.ctx` check in `cleanupConsumedUtxos` does not cover this:
`ls.ctx` is the caller's context from `Start`, and `Close` never cancels it — a
live restore/truncate closes the ledger while the node context stays alive.

## Fix

`Close` stops the timer and waits for a run already in flight. Both are owned by
a dedicated `cleanupMu` rather than the `LedgerState` `RWMutex`: an in-flight run
still takes `RLock` to read the tip, so draining under the ledger lock would
deadlock.

Both triggers register through `beginCleanupConsumedUtxosRun`, which reports
false once `Close` has set the closed flag. That is what constrains the epoch
transition's goroutine, and what stops an in-flight callback from installing a
fresh timer behind `Close`'s back.

The drain is unconditional, like this function's header-replay and
reward-precompute waits and unlike its bounded ones — returning early would
reintroduce the use-after-close it prevents. A run is bounded by
`cleanupConsumedUtxoBatchSize`, so there is no unbounded wait to guard against.

## Test

`ledger/cleanup_consumed_utxos_shutdown_test.go`, five tests covering the issue's
acceptance criteria: timer stopped on `Close`, `Close` drains an active callback,
no database work after `Close` (the epoch-transition path), repeated `Close`, and
schedule-after-`Close` not arming. All five fail with the fix reverted and the
test seams kept.

`TestCleanupConsumedUtxos_CoreModePrunes` is the positive control for the
no-work-after-close test: the same seeded row and tip are deleted by the same
call on an open ledger state, so a pass cannot come from cleanup being inert.

## Validation

- `go test -race ./ledger/` — ok, 360s
- `go vet ./ledger/`, `golangci-lint run ./ledger/...` — clean, 0 issues
- `make import-boundaries`, `make docs-parity` — ok
- `go test ./internal/test/conformance/` — 48 failures, identical set to an
  `origin/main` baseline run at `be3cf843` (pre-existing; #3722 targets them)
- `internal/test/devnet/run-tests.sh --accelerated` — fails identically on an
  `origin/main` baseline at `be3cf843`: all four nodes stay at
  `tip=slot 0/block 0` and the readiness phase times out. Pre-existing, filed
  separately.

Not run: `nilaway`, `modernize`, `golines` (not installed locally; `gofmt` clean
and all added lines are within 80 characters). `make sql-check` does not apply —
no SQL or `sqlc.yaml` change.

## Documentation

`ARCHITECTURE.md` updated in the storage-modes section, where the consumed-UTxO
pruner is documented, to record the shutdown contract. `DATABASE.md` checked and
not affected: pruning behavior is unchanged, only when the timer stops.

Fixes #3439

---
## Summary by cubic
Fixes `LedgerState.Close` so the consumed-UTxO cleanup timer is stopped and any in-flight cleanup run is drained before shutdown completes, preventing database use-after-close.

**Bug Fixes**

- Stops the timer and waits for a cleanup run already in flight; `time.Timer.Stop` alone doesn't wait for a fired callback.
- Both cleanup triggers (the periodic timer and the epoch-transition goroutine) now refuse to start once `Close` has begun.
- The drain uses a dedicated mutex instead of the ledger `RWMutex` to avoid deadlocking with an in-flight cleanup run.
- Adds `ledger/cleanup_consumed_utxos_shutdown_test.go` covering timer stop, drain, no database work after close, repeated close, and schedule-after-close.
- Updates `ARCHITECTURE.md` to document the shutdown contract.

Fixes #3439.

<sup>Written for commit f0beeff9d624cd024084aed3910556cd4b2cfedf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3726?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ledger shutdown handling for consumed-UTXO cleanup.
  * Cleanup timers now stop reliably when the ledger closes.
  * In-progress cleanup operations finish safely before shutdown completes.
  * Prevented cleanup work from starting or accessing storage after closure.
  * Repeated shutdown calls are now safe.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1420,6 +1420,18 @@ triggers are single-flight; once the node is near the upstream tip, each run
 deletes at most one bounded batch of eligible rows using the era's stability
 window. Later runs continue the cleanup, keeping SQLite occupancy bounded.
 
+`LedgerState.Close` stops that timer and waits for a run already in flight,
+under a dedicated mutex rather than the `LedgerState` `RWMutex` (an in-flight
+run still takes `RLock` to read the tip, so draining under the ledger lock
+would deadlock). Both are required: the timer callback re-arms itself, so a
+`Close` that only stopped it would let an in-flight run install a fresh timer
+behind `Close`'s back, and `time.Timer.Stop` never waits for an `AfterFunc`
+callback that has already fired. Both triggers also refuse to start once
+`Close` has set the closed flag, which is what constrains the epoch
+transition's own `go ls.cleanupConsumedUtxos()` — stopping the timer does not
+reach that goroutine. `LedgerState` does not own the database, and its owner
+closes it as soon as `Close` returns.
+
 ### Midnight gRPC Server
 
 In `storageMode: api` with `midnight.serverEnabled: true` and a non-zero
--- a/ledger/cleanup_consumed_utxos_shutdown_test.go
+++ b/ledger/cleanup_consumed_utxos_shutdown_test.go
@@ -1,0 +1,233 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ledger
+
+import (
+	"bytes"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/blinklabs-io/dingo/database/types"
+	"github.com/blinklabs-io/dingo/internal/test/testutil"
+)
+
+// shrinkCleanupConsumedUtxosInterval makes the periodic cleanup timer fire on
+// a test timescale. The package var is restored on cleanup, so these tests
+// must not run in parallel with each other.
+func shrinkCleanupConsumedUtxosInterval(t *testing.T, d time.Duration) {
+	t.Helper()
+	prev := cleanupConsumedUtxosInterval
+	cleanupConsumedUtxosInterval = d
+	t.Cleanup(func() { cleanupConsumedUtxosInterval = prev })
+}
+
+// newCleanupTimerFireSignal returns a hook that reports each timer fire on the
+// returned channel. The send is non-blocking so an unread fire never stalls
+// the timer callback itself.
+func newCleanupTimerFireSignal() (func(), <-chan struct{}) {
+	fires := make(chan struct{}, 64)
+	return func() {
+		select {
+		case fires <- struct{}{}:
+		default:
+		}
+	}, fires
+}
+
+// drainCleanupTimerFires discards every fire already buffered. Close has
+// stopped and drained the timer by the time this is called, so all of them
+// happened before it returned; only a fire arriving afterwards is a defect.
+// Draining fully -- rather than discarding a single fire -- is what keeps the
+// absence assertion from depending on how many intervals elapsed while the
+// test was between receives.
+func drainCleanupTimerFires(fires <-chan struct{}) {
+	for {
+		select {
+		case <-fires:
+		default:
+			return
+		}
+	}
+}
+
+// TestCleanupConsumedUtxos_TimerStopsOnClose covers the first half of issue
+// #3439: the cleanup timer callback re-arms itself via
+// scheduleCleanupConsumedUtxos, so a Close that does not stop it leaves a
+// self-perpetuating timer running against a database its owner closes
+// immediately after Close returns (LedgerState does not own the database --
+// see the note at the end of Close).
+func TestCleanupConsumedUtxos_TimerStopsOnClose(t *testing.T) {
+	shrinkCleanupConsumedUtxosInterval(t, 5*time.Millisecond)
+	db := newTestDBForCleanup(t, types.StorageModeCore)
+	ls := newLedgerStateForCleanup(db, 100_000)
+
+	hook, fires := newCleanupTimerFireSignal()
+	ls.cleanupConsumedUtxosTimerFiredHook = hook
+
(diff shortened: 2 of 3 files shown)


chrisguiney:
## Bot review disposition

**cubic** — did not review. Its `neutral` check is a monthly line-quota block
(403,149 / 400,000 lines, resuming 1 September 2026), not an approval.

**CodeRabbit** — one finding, `ledger/state.go:1991-1993`, concurrent `Close`
callers. Classified **pre-existing, non-blocking, out of scope for this PR**.

Verified rather than accepted:

- The finding is about the `closed.CompareAndSwap` early return, which predates
  this PR. `Close` already had two unconditional waits (`replayWG`,
  `rewardPrecomputeWG`) and four bounded ones; a concurrent second caller
  already returned `nil` while all of those were in flight. This PR adds one
  more wait to that sequence, so it neither introduces the behavior nor widens
  its kind.
- CodeRabbit's premise that "production shutdown paths currently serialize this
  sequence" does not hold. `shutdown()` takes `startupLifecycleMu`, which
  serializes it against the startup unwind (`node.go`) but not against
  `liveLifecycleMu`, which `Restore`/`Truncate` hold across
  `closeStorageForLiveLifecycleOp`. A `Stop()` concurrent with an in-flight
  `Restore`/`Truncate` can therefore reach `LedgerState.Close` twice. That is a
  real gap, and it is a pre-existing one.
- The suggested fix is not free. Making the losing caller return the first
  caller's error changes the contract for `node_shutdown.go`, which passes
  `n.ledgerState.Close` through its deadline helper and escalates any non-nil
  error to `errStorageDrainUnconfirmed` and a supervised restart via
  `n.cancel()`. A concurrent `Stop` that returns `nil` today would begin
  triggering restarts. That belongs in a change that owns `Close`'s idempotence
  contract and can validate the restart-policy blast radius across all three
  call sites, not in a cleanup-timer shutdown fix.

Recommend a follow-up issue for `Close`'s concurrent-caller contract.

chrisguiney:
Rebased onto `83ee8cdb` and force-pushed.

The previous CI failure was not from this change. The branch was based on
`be3cf843`, which predates #3722; the two failing packages were
`internal/test/conformance` and `ledger/governance`, both governance/ratification
and both fixed by that PR. `github.com/blinklabs-io/dingo/ledger` — the package
this change touches — passed in that same run.

Verified at both commits rather than inferred:

- `go test -run TestProcessEpochSPONonVoter ./ledger/governance/` at `be3cf843`:
  FAIL. At `83ee8cdb`: ok.
- `go test ./internal/test/conformance/` on the rebased branch: ok, exit 0
  (was 48 failures on the old base, an identical set to an `origin/main`
  baseline run at `be3cf843`).
- `go test -race -run TestCleanupConsumedUtxos_ ./ledger/` on the rebased
  branch: ok.

wolf31o2 (approved):
Approving. All three gaps the description names are fixed, and each is independently pinned by its own test:

| Reverted | Fails |
| --- | --- |
| `cleanupWG.Wait()` | `CloseWaitsForActiveCallback` |
| `closed` check in `beginCleanupConsumedUtxosRun` | `NoDatabaseWorkAfterClose` |
| `closed` check in `scheduleCleanupConsumedUtxos` | `TimerStopsOnClose`, `RepeatedCloseIsSafe`, `ScheduleAfterCloseDoesNotArm` |

The two claims the correctness rests on both check out against the code rather than the comments:

**The Add-after-Wait race is genuinely closed.** `Close` sets `closed` via `CompareAndSwap` at `state.go:1991`, roughly fifty lines before it takes `cleanupMu`, and `beginCleanupConsumedUtxosRun` re-checks `closed` while holding that same mutex. So a run that wins the race registers before `Close` acquires the mutex and is therefore waited on; one that loses sees `closed` and never calls `Add`. There is no interleaving that reaches `WaitGroup.Add` after `Wait` has begun. The `CompareAndSwap` also makes a second `Close` return early, which is what `RepeatedCloseIsSafe` covers.

**The deadlock the separate mutex exists to avoid is real, and avoided.** `Close` holds no `ls.Lock()` or `ls.RLock()` anywhere across its whole body (1985-2286), so `cleanupWG.Wait()` runs with the ledger lock free and an in-flight run can still take `RLock` to read the tip and finish. Had the drain been placed under the `LedgerState` RWMutex — the obvious thing to reach for — it would have deadlocked against exactly that read.

Making the wait unconditional while the neighbouring waits are bounded is the right asymmetry: a bounded wait that expires here would return `Close` into the use-after-close it exists to prevent, and a run is capped at one `cleanupConsumedUtxoBatchSize` delete, so there is no unbounded drain to protect against.

Turning `cleanupConsumedUtxosInterval` into a `var` for tests follows the existing precedent of the `Close*` drain timeouts in the same file.

`go test -race ./ledger/ -run CleanupConsumedUtxos` green, CI green on 10 checks.

PR #3730, opened 2026-08-31
fix(ledger): check MIR pot capacity before crediting

### LINKED ISSUE #3425: ledger: validate MIR pot capacity before applying credits
`ledger/mir.go` credits every registered reward account before `debitMIRPot` checks whether the selected reserves or treasury pot can cover the total. An over-budget MIR certificate makes the debit fail after the credit loop; the rollover transaction aborts, and every deterministic epoch-boundary retry fails in the same way.

Acceptance criteria:

- Compute the total credit and compare it with the selected pot before applying any reward-account credit.
- Treat an over-budget certificate as a no-op for the epoch instead of permanently failing the boundary.
- Cover exact-budget and over-budget certificates without partial state mutation.

chrisguiney:
Fixes #3425.

## Problem

`applyMIRCerts` credited every registered reward account and only then called
`debitMIRPot`, which fails when the selected pot cannot cover the total. The
failure aborted the epoch-rollover transaction, and because the certificates
remain in `move_instantaneous_rewards`, they were re-read and re-failed on
every deterministic retry, wedging the boundary.

Dingo does not implement the DELEG-time `InsufficientForInstantaneousRewards`
check, so an over-budget certificate can reach the epoch boundary.

## Change

The ended epoch's certificates are aggregated and checked before any of them is
applied, following `Cardano.Ledger.Shelley.Rules.Mir.mirTransition`:

- Credits are restricted to registered, active reward accounts before totalling
  (`Map.intersection accountsMap`), so an unregistered credential does not count
  against the pot.
- Pot-to-pot transfers are folded into the available balances
  (`deltaReserves` / `deltaTreasury`), so they fund distributions at the same
  boundary.
- The boundary is applied only when
  `totR <= availableReserves && totT <= availableTreasury`. When either pot
  falls short the whole boundary is a no-op — no credit, debit or transfer is
  written — and the rollover succeeds, mirroring the rule's else branch, which
  returns the original `ChainAccountState` and clears the pending rewards.
- An unaffordable pot-to-pot transfer takes the same no-op branch, since
  cardano-ledger computes the available balance over unbounded `Coin`. A uint64
  overflow has no reference analogue and still returns an error.

The boundary now writes one `network_state` row instead of one per certificate.

`ARCHITECTURE.md` (rollover step 2) and `DATABASE.md`
(`GetMIRCertsInSlotRange`) are updated to match.

## Tests

Eight tests added to `ledger/mir_test.go`. Five fail without the fix:

- `TestApplyMIRCerts_OverBudgetReservesIsNoOp`
- `TestApplyMIRCerts_OverBudgetIsAggregateAcrossCerts`
- `TestApplyMIRCerts_OverBudgetTreasuryBlocksReservesDistribution`
- `TestApplyMIRCerts_OverBudgetDropsPotTransfer`
- `TestApplyMIRCerts_PotTransferLargerThanPotIsNoOp`

Three pass before and after, guarding against an over-broad fix — `<` in place
of `<=`, budgeting unregistered credits, and breaking transfer-funded
distributions:

- `TestApplyMIRCerts_ExactBudgetApplies`
- `TestApplyMIRCerts_BudgetExcludesUnregisteredCredentials`
- `TestApplyMIRCerts_PotTransferCountsTowardAvailablePot`

## Validation

Passed:

- `go test ./ledger/...`
- `go test ./ledger/ -race` (full package, 361s)
- `go test ./database/plugin/metadata/...`
- `go test ./internal/test/conformance/`
- `golangci-lint run ./ledger/...` (0 issues)

`internal/test/devnet/run-tests.sh --accelerated` fails in its readiness phase
with all four nodes at `tip=slot 0/block 0, fwd=0`. An unmodified `origin/main`
worktree at 83ee8cdb reproduces the same failure identically, so it is
pre-existing and unrelated to this change. The DevNet also contains no MIR
certificates — they are Shelley-through-Babbage only — so it does not exercise
this code path.

Not run: `nilaway` and `modernize` (not installed on this machine); no
real-chain replay, since mainnet MIR certificates end in Babbage.

---
## Summary by cubic
Fixes the epoch-rollover wedge caused by over-budget MIR certificates. `applyMIRCerts` previously credited accounts first and only then failed on the pot debit, aborting the rollover and re-failing on every retry; the boundary now aggregates every certificate and checks pot capacity before applying any of them.

**Bug Fixes**
- Credits count only registered, active reward accounts, matching cardano-ledger's `Map.intersection accountsMap` restriction.
- Pot-to-pot transfers are folded into the available balances so they fund same-boundary distributions.
- The whole boundary is a no-op when either pot falls short, and writes one `network_state` row instead of one per certificate.

**Tests**
- Eight tests added to `ledger/mir_test.go`; five fail without the fix and three guard against over-broad fixes.
- `ARCHITECTURE.md` and `DATABASE.md` updated to describe the new boundary semantics.

<sup>Written for commit 6763a3c69c821151f42480458732130f9f00eabe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3730?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * MIR rewards now apply atomically at epoch boundaries.
  * Over-budget reward distributions or transfers are safely skipped without partially changing account balances or treasury/reserve pots.
  * Only registered, active reward accounts receive MIR credits.
  * Affordable rewards and cross-pot transfers are applied using combined boundary budget checks.

* **Documentation**
  * Updated architecture and database documentation to describe MIR aggregation, budget validation, and no-op behavior.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -8007,7 +8007,15 @@ a fixed order, mirroring `cardano-ledger`'s sequencing:
    Shelley's `NEWEPOCH` embeds MIR between `applyRUpd` and `EPOCH`, so it runs
    before both the stake snapshot and POOLREAP: its credits are part of the mark
    snapshot and its pot movements are visible to POOLREAP, governance and the
-   ADA-pot capture.
+   ADA-pot capture. The boundary's certificates are aggregated and checked
+   against pot capacity before any of them is applied, mirroring
+   `mirTransition`: credits are restricted to registered, active reward
+   accounts, pot-to-pot transfers are folded into the available balances, and
+   the boundary is applied only when `totR <= availableReserves && totT <=
+   availableTreasury`. When either pot falls short the whole boundary is a
+   no-op — no credit, debit or transfer is written — and the rollover still
+   succeeds; the certificates are scoped to the ended epoch's slot range, so a
+   discarded MIR is not retried at the next boundary.
 3. SNAP-point mark stake read (`captureEpochBoundarySnapshotStake` →
    `snapshot.Manager.ComputeEpochBoundarySnapshot`, when a stake hook is
    installed): read the mark snapshot's stake distribution here, after the two
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -2376,7 +2376,7 @@ rollback past the boundary reverts them and re-application is deterministic.
 
 ### `GetMIRCertsInSlotRange`
 
-MIR certificates for the epoch range `[startSlot, endSlot)`, applied at the epoch boundary as the Shelley INSTANT rule. Distribution certs (`other_pot = 0`) credit registered reward accounts and debit the source pot in `network_state`; pot-to-pot transfer certs (`other_pot > 0`) move that amount between treasury and reserves directly. The `mir.id` value is retained by the processed effect as the per-MIR reward-credit discriminator so multiple MIR certs can credit the same account at one boundary without collapsing into one `account_reward_delta` row.
+MIR certificates for the epoch range `[startSlot, endSlot)`, applied at the epoch boundary as the Shelley INSTANT rule. Distribution certs (`other_pot = 0`) credit registered reward accounts and debit the source pot in `network_state`; pot-to-pot transfer certs (`other_pot > 0`) move that amount between treasury and reserves. Every cert in the range is aggregated before any is applied: distribution totals count only credentials with a registered, active account, transfers are folded into the available pot balances, and the boundary writes a single `network_state` row. If either pot cannot cover its total the boundary is a no-op rather than an error, so an over-budget cert cannot fail the epoch rollover on every retry. The `mir.id` value is retained by the processed effect as the per-MIR reward-credit discriminator so multiple MIR certs can credit the same account at one boundary without collapsing into one `account_reward_delta` row.
 
 ```sql
 SELECT mir.id, mir.pot, mir.other_pot, mir.added_slot,
--- a/ledger/mir.go
+++ b/ledger/mir.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/blinklabs-io/dingo/database"
+	"github.com/blinklabs-io/dingo/database/models"
 	"github.com/blinklabs-io/dingo/ledger/governance"
 )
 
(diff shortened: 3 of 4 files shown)


chrisguiney:
@coderabbitai review

chrisguiney:
@coderabbitai full review

chrisguiney:
@coderabbitai review

wolf31o2 (approved):
Approving. I checked each claim about `mirTransition` against the Haskell rather than the description, since this is consensus behaviour and the description's accuracy is the whole argument.

All four hold verbatim in `Rules/Mir.hs`:

- `irwdR = iRReserves (dsIRewards ds) \`Map.intersection\` accountsMap` — totals are taken over registered accounts only.
- `availableReserves = reserves \`addDeltaCoin\` deltaReserves`, likewise for treasury — pot-to-pot deltas fund the same boundary.
- The guard is exactly `totR <= availableReserves && totT <= availableTreasury`, one combined condition, so a treasury shortfall blocks the reserves distribution too.
- **Both** branches end in `dsIRewardsL .~ emptyInstantaneousRewards`. The failure branch clears the pending rewards while leaving `ChainAccountState` untouched, which is the part that stops the retry loop.

That last point is where the implementation differs in mechanism, and the parity argument is worth stating because it is not obvious from the diff: the no-op branch here returns without deleting from `move_instantaneous_rewards`, where the reference clears the pending set. It is nonetheless equivalent, because `GetMIRCertsInSlotRange(epochStartSlot, boundarySlot, ...)` scopes the read to the ended epoch's slots. A skipped certificate is never re-read at a later boundary, so it is inert afterwards even though the row survives as history. Deleting instead would lose that history for no behavioural gain.

Both halves are pinned. Making the capacity guard unreachable fails five tests:

```
OverBudgetReservesIsNoOp
OverBudgetIsAggregateAcrossCerts
OverBudgetTreasuryBlocksReservesDistribution
OverBudgetDropsPotTransfer
PotTransferLargerThanPotIsNoOp
```

`OverBudgetTreasuryBlocksReservesDistribution` is the one that matters most for parity — it pins the single combined condition rather than two independent per-pot checks, which is the easy thing to get wrong here.

Disabling the registered-account restriction fails `BudgetExcludesUnregisteredCredentials`, so the `Map.intersection` semantics are pinned on the budget side specifically, not just on whether an unregistered account gets credited.

Treating an unaffordable pot-to-pot transfer as the same no-op is right given the reference computes availability over unbounded `Coin`; keeping a uint64 overflow as an error rather than a silent no-op is the correct divergence, since it has no analogue in the reference and indicates corrupt input rather than an over-budget boundary.

`go test -race ./ledger/ -run MIR` green, CI green on 10 checks.

Issue #3627, opened 2026-08-29
validation: Preprod Mithril rejects producer-valid transaction over declared Plutus budget


wolf31o2:
## Problem

The Preprod Mithril bootstrap completed on baseline `aefbf720c59c0d9c091dbae70393d140b2ed29bb`, but live serve repeatedly rejected the same producer-valid transaction at the same applied tip. Transaction `d81392f4def652323c5648067ffbe6d45812e415a53d867336aa5026db9ea2eb` in block slot `132228934` exceeded its declared Plutus budget:

```
used:      832074315 cpu, 3057846 mem
declared:  831433551 cpu, 3055442 mem
```

The ledger tip remained at slot `132228918`; recovery reported deterministic rejection without progress, and shutdown timed out after the live window.

## Impact

Preprod Mithril bootstrap does not transition into a healthy 30-minute live serve run.

## Public reproduction evidence

The same evaluator mismatch reproduces on Preview. In a 30-minute live serve run from a Mithril bootstrap at Dingo baseline `6792013f9eb4c458ef5c5771fed98de1a4f25217`:

- Transaction: [`a155d2b26bf0683cd45ccdb85a354b6c860b4534a9468e51dd87164a88225e5b`](https://preview.cardanoscan.io/transaction/a155d2b26bf0683cd45ccdb85a354b6c860b4534a9468e51dd87164a88225e5b)
- Failing block slot: `121700813`; applied tip before it: `121700796`
- Script hash: `2b91a7e666575a2465b8c7f6a7f960d5870cf13694a67f3215e014c5`
- Redeemer: tag `0`, index `2`
- Evaluated: `765425338` CPU, `3067299` memory
- Declared: `764784574` CPU, `3064895` memory
- Result: the producer-accepted block was rejected repeatedly, the tip did not advance past slot `121700796`, and recovery reported that the peer rollback point could not be crossed.

A fresh run on the current `origin/main` baseline `8419482f598fa2ff605488cbef967db1b2cf9876` reproduced the same mismatch:

- Transaction: [`d4485c3ead7ef560d3db0bd44b75fc83244d7df4c71153a3192f334e96661079`](https://preview.cardanoscan.io/transaction/d4485c3ead7ef560d3db0bd44b75fc83244d7df4c71153a3192f334e96661079)
- Failing block slot: `121702975`; applied tip before it: `121702968`
[...]

Issue #3628, opened 2026-08-29
validation: Preprod genesis rejects canonical withdrawal with reward balance mismatch


wolf31o2:
## Problem

The Preprod from-genesis validation run on baseline `aefbf720c59c0d9c091dbae70393d140b2ed29bb` reached applied slot `12313331` and repeatedly rejected a canonical withdrawal transaction. Transaction `6c4fb0da05d3ec7f33bc9c980e38758fd0ce4e6d49519fae5c7f207d7ea056d` supplied `399009945` lovelace, while the local reward-account balance was `399088479`, a difference of `78534` lovelace.

Replay recovery rewound and retried without reaching the tip.

## Impact

Preprod genesis sync cannot reach the tip or enter its 30-minute live serve window.

## Evidence

- Run: `/tmp/dingo-validation/runs/preprod-genesis-origin-aefbf720-fresh2`
- Withdrawal mismatch: `genesis-serve.log:59683`
- Profile: `/tmp/dingo-validation/profiles/preprod-genesis-origin-aefbf720-stall-12313331`

## Related history

The failure occurred despite the historical recovery work in #3005 and merged PR #3251. The current mismatch and network position are not established as the same defect and need a current investigation.

## Acceptance criteria

- Determine why the local reward-account balance diverges from the canonical withdrawal amount.
- Ensure replay recovery does not hide or repeatedly reproduce the underlying ledger divergence.
- Re-run Preprod genesis from the same baseline family through tip and the 30-minute live window.

wolf31o2:
## Additional validation evidence

An independent Preprod `storageMode: api` bootstrap on current `origin/main` reproduced the same class of deterministic withdrawal/reward-state divergence: a canonical withdrawal exceeded the locally reconstructed reward balance, after which historical backfill aborted before the serve phase.

This confirms the validation symptom remains relevant to API-mode bootstrap. PR #3639 appears to address the replay-recovery disposition by halting persistently instead of rewinding indefinitely; the underlying reward-state reconstruction still needs verification on the current branch.

No hostnames, filesystem paths, or private topology details are included here.

Cerkoryn:
[...]

PR #3735, opened 2026-09-01
fix(ledger): trust producer-valid Plutus budget overage on the followed chain

chadaibbhot-git:
## Title
fix(ledger): trust producer-valid Plutus budget overage on the followed chain

## Problem
On the primary/followed-chain apply path, a producer-valid transaction whose Plutus script runs
slightly over its declared ex-units is rejected, and the node then re-rejects the same canonical
block forever — the tip freezes and never advances (deterministic rejection, no progress). This is
#3627 (preprod); it also reproduces on preview. The discrepancy is tiny (~0.07% CPU/mem): Dingo's
evaluator (Plutigo) computes marginally higher ex-units than the producing Haskell node, so a tx the
whole network accepted is rejected locally. It persists on Plutigo v0.5.0 (latest cost models), so
the residual miscount is in the evaluator, not the cost-model tables (tracked separately).

## Root cause
`ledger/eras/conway.go` `evaluateConwayPlutusScript` runs in two modes via its `restrictive` flag:
`restrictive=true` runs the CEK machine at `MaxTxExUnits` and then post-checks `used > declared`,
raising `"script exceeded declared budget"` (wrapped in `PlutusScriptFailedError`); `restrictive=false`
("exact mode") skips that post-check. Every validation path hardcoded `restrictive=true`, so the
followed-chain apply path rejected producer-valid overages and wedged.

## Fix
Thread `restrictive` per-path via a `LedgerView` flag (`trustProducerPlutusBudget`), set true **only**
on the followed-chain apply-path view (`ledgerProcessBlock`). That path now evaluates in exact mode
(`budget=MaxTxExUnits, restrictive=false`) — trusting the producer's declared budget — while **mempool
admission and block forging keep `restrictive=true`** (strict), so we never admit or produce an
over-declared transaction. Genuine script failures (`execErr` from `Evaluate`: evaluation error,
false/invalid result, type/builtin error) still reject in **both** modes; a script genuinely over
`MaxTxExUnits` still exhausts the machine and rejects; phase-1 `ExUnitsTooBig` still bounds declared
budgets. The only newly-accepted window is `declared ≤ used ≤ MaxTxExUnits` — exactly the
producer-trust case. When trust fires it logs `"Plutus budget overage on producer-valid block
(trusting producer)"`.

## Why this is safe
Rejecting a block the honest majority already accepted forks the follower off the canonical chain —
strictly worse than tolerating a sub-percent local cost-meter discrepancy on a block we are only
following. The strict paths (mempool/forging) are structurally separate `LedgerView` constructions
that leave the flag false, so this never weakens what we produce or relay-admit.

## Test
`ledger/eras/conway_plutus_producer_trust_test.go` drives the **real** `ValidateTxConway` with a real
Plutus script that over-runs its declared budget: (a) apply-path/trust view → accepted + overage
reported; (b) strict view → rejected with the exact `"script exceeded declared budget"`; (c) a genuine
script failure under trust → still rejected (not misclassified, not reported). Neutering the fix
(forcing `restrictive=true`) makes (a) fail with the real production error; restoring passes.
`go build/vet/test ./ledger/...` and `gofmt` all pass.

## Verified live
Deployed to a preview node that had been wedged ~15h on exactly this rejection (stuck one block
below the network, re-rejecting the same producer-valid tx). With the fix it logged the trust
warning for the offending script, accepted the block, and caught up to the network tip
(`syncProgress 100%`). Mempool/forging behaviour unchanged.

## Update (review-hardened)
- Transaction-wide cumulative ExUnits cap: the trust path drops only the per-redeemer *declared*-budget compare; the tx-wide MaxTxExUnits ceiling is enforced across all redeemers (rejects on cumulative overflow). Mempool + forging remain strict.
- Trust is threaded explicitly and disabled for historical replay (ProcessTrustedBlockBatches).
- Relates to #3627. NOTE: the underlying over-count is now root-caused to a gouroboros ScriptContext bug (validity-interval upper bound encoded inclusive instead of exclusive for TTL-only txs) — a one-line gouroboros fix is filed separately; this Dingo-side trust remains a robustness safety net for producer-valid budget disagreements.


ledger/state.go
```diff
@@ -6263,6 +6263,13 @@ func (ls *LedgerState) ledgerProcessBlock(
 					ls:                   ls,
 					intraBlockUtxos:      intraBlockUtxos,
 					skipPhase2Validation: skipPhase2Validation,
+					// Followed-chain apply path: defer to the producer on the
+					// Plutus declared-budget check so a sub-percent local ex-unit
+					// over-count does not deterministically re-reject a block the
+					// honest network accepted and wedge this follower
+					// (blinklabs-io/dingo#3627). Mempool admission and forging
+					// build their LedgerView without this and stay strict.
+					trustProducerPlutusBudget: true,
```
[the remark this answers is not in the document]
chadaibbhot-git: This is already handled on the current head (`8a4a5ee3`): `trustProducerPlutusBudget` is threaded through `ledgerProcessBlocksFromSource` → `ledgerProcessBlock`, and the two entry points pass opposite values. The live followed-chain drain (`ledgerProcessBlocks`) passes `true`; the historical immutable-load replay (`ProcessTrustedBlockBatches`) passes `false`, with a comment that the producer-budget trust is a live-follower-only tolerance and must never relax a block during replay. So replay re-derives every phase-2 verdict strictly — the trust never applies outside the followed-chain apply path.

ledger/state.go
```diff
@@ -4871,12 +4882,16 @@ func (ls *LedgerState) ProcessTrustedBlockBatches(
 			}
 		}
 	}()
-	return ls.ledgerProcessBlocksFromSource(ctx, readChainResultCh)
+	// Historical immutable-load replay: re-derive every phase-2 verdict
+	// strictly (#3627). The producer-budget trust is a live-follower-only
+	// tolerance and must never relax a block during replay.
+	return ls.ledgerProcessBlocksFromSource(ctx, readChainResultCh, false)
```
wolf31o2: This `false` is the guard that keeps historical replay independently re-deriving phase-2 verdicts, and nothing tests it.

`TestLedgerProcessBlockProducerTrustGatedByCaller` passes the flag into `runBudgetTrustProbe(t, false)` directly, so it verifies that `ledgerProcessBlock` respects the value it is given. It never reaches this call site. Changing this argument to `true` — enabling producer trust for the entire immutable-chain replay — leaves `go test ./ledger/` fully green, including that test.

The test's own comment states the property that is missing: "replay routes phase-2 through ledgerProcessBlock too, and must NOT inherit the live-follower trust." That is an assertion about this line, not about the parameter the probe supplies.

Driving `ProcessTrustedBlockBatches` end-to-end and asserting the validator observes trust disabled would pin it, and is the same shape as the probe already in that file.

wolf31o2 (changes requested):
Requesting changes on one gap. The implementation itself is careful and the confinement is real, but the safety property the design leans on hardest is the one property no test enforces.

**Verified as claimed.** The flag is set true at exactly one call site (the live followed chain, `state.go:4674`), false for `ProcessTrustedBlockBatches`, and all five other non-test `LedgerView` literals leave it zero-valued, so mempool admission, forging, and query views stay strict without relying on anyone remembering to pass false. Genuine `execErr` still rejects in both modes, and the cumulative `MaxTxExUnits` cap uses overflow-checked addition before comparing.

Both mechanisms are independently pinned: forcing `restrictive = true` fails `ApplyPathAcceptsBudgetOverage` and the three cumulative-cap tests; disabling the cumulative comparison fails `CumulativeCapRejectsTwoScriptsOverMax` alone.

**The replay guard is not pinned.** `TestLedgerProcessBlockProducerTrustGatedByCaller` takes the trust flag as a parameter to `runBudgetTrustProbe`, so it proves `ledgerProcessBlock` *honors* whichever value it is handed — not that `ProcessTrustedBlockBatches` hands it `false`. Flipping that argument at `state.go:4888` from `false` to `true` leaves the entire `./ledger/` package green.

That matters more than a normal missing test because replay is where the strict re-derivation lives. If that argument is ever flipped by a refactor or a bad merge, dingo silently stops independently re-deriving phase-2 verdicts for the whole immutable chain, and the test named `GatedByCaller` keeps passing. A test that drives `ProcessTrustedBlockBatches` itself, or asserts the argument at that call site, closes it.

**A sequencing question that is yours rather than mine.** The description now root-causes the discrepancy to a gouroboros `ScriptContext` bug — validity-interval upper bound encoded inclusive rather than exclusive for TTL-only transactions — with a one-line fix filed separately. That reframes this change: it is no longer a tolerance for an unexplained sub-percent meter drift, it is a permanent relaxation of a phase-2 check whose triggering bug is understood and separately fixed.

Worth deciding deliberately, because the relaxation is not symmetric with what it protects against. A mis-encoded script context can change a script's *result*, not only its cost, and a result difference still rejects through `execErr` — so this safety net does not prevent the wedge class it is named for, only the cost-difference subset of it. Against that, it permanently accepts any transaction in the `declared <= used <= MaxTxExUnits` window on the followed chain, including one an honest majority would treat as phase-2 invalid.

Both positions are defensible. Keeping it as a standing robustness net against future evaluator disagreement is a real argument; so is landing the gouroboros fix and reverting this. The PR presents only the first, and the second is not addressed.

Checks at 8a4a5ee3 in a clean worktree: `go test ./ledger/ ./ledger/eras/` green, CI green on all eight checks, both bots reviewed with author responses.

ledger/state.go
chadaibbhot-git: You're right — `TestLedgerProcessBlockProducerTrustGatedByCaller` only proves `ledgerProcessBlock` honors the flag and never reaches this call site, so flipping the `false` here left the suite green.

Added `TestProcessTrustedBlockBatchesPassesStrictTrust` (56f7bc0f), which closes it behaviorally rather than by argument assertion: it drives an over-declared-budget block through the real `ProcessTrustedBlockBatches` end-to-end (real `chain.Manager` + stored block, mock Dijkstra era whose phase-2 validator records the trust it observes) and asserts the validator runs, sees trust **disabled**, and **rejects** the overage. To get the mock block past the immutable-load CBOR-offset computation that runs ahead of phase-2, it carries a minimal valid Dijkstra `[header, block_body]` with one no-output tx and a matching `BlockBodySize()`. Verified the pin: flipping this `false` to `true` fails the new test; reverted, `./ledger/` and `./ledger/eras/` are green.

chadaibbhot-git:
On the sequencing question you raised (mine to decide, so — deliberately):

You're right that with the gouroboros ScriptContext validity-interval fix now filed separately (blinklabs-io/gouroboros#2159), this stops being "tolerate an unexplained sub-percent drift" and becomes a permanent phase-2 relaxation whose trigger is understood and fixed, and that the relaxation is asymmetric — a mis-encoded context that changes a script *result* still rejects through `execErr`, so this only nets the cost-difference subset, while permanently accepting anything in `declared <= used <= MaxTxExUnits` on the followed chain.

Given that, the plan is **(b) with correct sequencing**: keep this as the interim followed-chain net **only until #2159 is merged and lands in dingo's gouroboros dependency**, then revert it. Reverting before the corrected `ScriptContext` is actually in the built node would re-expose the wedge on the followed chain, so the two have to be ordered. I'll open a tracking note to revert once the dep bump lands. If you'd rather not carry it even in the interim, I'm happy to close this and gate everything on #2159 instead — your call as maintainer, but I didn't want to leave the second position unaddressed.

ledger/plutus_producer_trust_replay_test.go
```diff
@@ -0,0 +1,315 @@
...
+	require.False(
+		t,
+		sawTrust,
+		"ProcessTrustedBlockBatches must pass trust=false (state.go immutable-load call site); flipping it to true fails here",
+	)
+	require.Error(
+		t,
+		err,
+		"immutable-load replay must reject the over-declared-budget block",
+	)
```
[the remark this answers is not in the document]
chadaibbhot-git: Good catch — fixed. `require.Error` could pass on an incidental later state/DB error even if the validator failed to reject on budget. Now asserts the specific verdict: `require.ErrorAs` into `conway.PlutusScriptFailedError` + `require.ErrorIs(plutusErr.Err, errOverDeclaredBudget)`, matching the sibling `GatedByCaller` test. Verified it still pins the call site (flipping state.go:4888 to `true` still fails the test).

chadaibbhot-git:
Rebased onto `main` (`f9e79258`) to resolve the conflict with #3740 (adapt to Gouroboros v0.202.4). The only conflict was two test call sites in `ledger/block_transaction_validity_test.go` — `#3740` reformatted the `ledgerProcessBlock` calls, this PR adds the trailing `trustProducerPlutusBudget` argument; kept both. No logic change beyond the conflict resolution. `go build ./...`, `./ledger/` and `./ledger/eras/` green on v0.202.4.

ledger/eras/conway.go
```diff
@@ -575,27 +585,94 @@ func validateTxPlutusConwayWithContext(
...
+		// honest network already accepted. Mempool and forging leave
+		// trustProducerBudget false, keep restrictive mode, and still reject
+		// over-budget transactions.
+		evalBudget := redeemerValue.ExUnits
+		restrictive := true
+		if trustProducerBudget {
+			evalBudget = pp.MaxTxExUnits
+			restrictive = false
+		}
+		usedBudget, execErr, err := evaluateConwayPlutusScript(
```
wolf31o2: This is the relaxation, and the reference does not permit it.

`evalBudget = pp.MaxTxExUnits` with `restrictive = false` drops the `used > declared` verdict. In cardano-ledger the declared per-redeemer ExUnits *is* the machine budget on the apply path — `mkPlutusWithContext ... exUnits` from `Map.lookup redeemerIndex redeemers`, then `evaluatePlutusRunnable ... (transExUnits pwcExUnits)` — and exceeding it is `EvaluationError` -> `Fails` -> `ValidationTagMismatch`.

Also note the budget is the full per-tx maximum for *every* redeemer and is never reduced by `totalUsed`, while the cumulative check runs only after each evaluation completes. A transaction can burn ~2x `MaxTxExUnits` of metered work before rejection, and nothing bounds a block's metered work, since the block-level rule uses declared ex-units that an attacker sets near zero. cardano-node aborts on the first script.

`evalBudget = MaxTxExUnits - totalUsed` would let the machine enforce the ceiling and would subsume the post-hoc cap below.

ledger/eras/conway.go
```diff
@@ -575,27 +585,94 @@ func validateTxPlutusConwayWithContext(
...
+		if totalUsed.Steps > pp.MaxTxExUnits.Steps ||
+			totalUsed.Memory > pp.MaxTxExUnits.Memory {
+			return alonzo.ExUnitsTooBigUtxoError{
+				TotalExUnits: lcommon.ExUnits{
+					Steps:  totalUsed.Steps,
+					Memory: totalUsed.Memory,
+				},
+				MaxTxExUnits: pp.MaxTxExUnits,
+			}
+		}
```
wolf31o2: The cumulative cap is itself a non-reference rule, and it manufactures a rejection in exactly the case this PR exists to fix.

The reference compares *declared* totals in phase-1 and never sums metered usage. Probed against the real `ValidateTxConway`: a multi-redeemer transaction declaring `160099 x 2 <= 320199` — phase-1 valid, and phase-2 valid under Haskell metering — fails here as `ExUnits too big: total 320200/2200, maximum 320199/2199`, because Dingo's over-count pushes the metered sum past the cap.

And a single redeemer at the cap never reaches the trust at all: `out of budget: requested (cpu=48000, mem=300), available (cpu=47999, mem=300)` — the machine exhausts first.

So the fix covers only the sub-cap subset. Multi-redeemer near-cap transactions — DEX batchers, aggregators — are both the most common shape at the ceiling and where absolute meter drift is largest.

ledger/state.go
```diff
@@ -4751,7 +4751,18 @@ func (ls *LedgerState) ledgerProcessBlocks(ctx context.Context) {
...
+					readChainResultCh <-chan readChainResult,
+				) error {
+					// Live followed chain: trust the producer's Plutus
+					// declared-budget verdict (#3627).
+					return ls.ledgerProcessBlocksFromSource(
+						srcCtx,
+						readChainResultCh,
+						true,
+					)
+				},
```
wolf31o2: This live call site is unpinned — the mirror image of the finding I raised at `state.go:4974`.

Flipping this `true` to `false`, which disables the entire feature this PR exists for, leaves the package green:

```
$ go build ./... ; go test -count=1 -timeout 1800s ./ledger/
ok  github.com/blinklabs-io/dingo/ledger  140.590s   (exit 0)
```

The named instance at :4974 was closed with `TestProcessTrustedBlockBatchesPassesStrictTrust`, but the class was not audited. `TestLedgerProcessBlockProducerTrustGatedByCaller` cannot cover it, because `runBudgetTrustProbe(t, trust)` takes the flag as a parameter.

Related: the enforce/skip class is Conway-only. `alonzo.go:325-335`, `babbage.go:351-361` and `:413-423` hardcode the strict verdict with no gate, and `dijkstra.go:278` delegates to a strict gouroboros rule. The gouroboros context bug is era-independent for this transaction shape, so a from-genesis follower can still wedge on a Babbage block, and Dijkstra re-wedges at the next hard fork.

ledger/eras/conway_plutus_producer_trust_test.go
```diff
@@ -0,0 +1,268 @@
...
+	require.Len(t, ls.reported, 1, "the tolerated overage must be reported")
+	ev := ls.reported[0]
+	assert.Equal(t, scriptHash, ev.scriptHash)
+	assert.Equal(t, lcommon.RedeemerTagMint, ev.tag)
+	assert.Equal(t, usedCpu, ev.used.Steps)
+	assert.Equal(t, usedMem, ev.used.Memory)
+	assert.Equal(t, int64(0), ev.declared.Steps)
+	assert.Equal(t, int64(0), ev.declared.Memory)
+}
+
```
wolf31o2, line 171: This test asserts the defect. Declared 0/0, metered 112100/800, `require.NoError` — a block cardano-node rejects with `ValidationTagMismatch`, accepted and applied.

There is also no absence case for the followed path: no test that an over-budget block is still rejected when the producer could not have validated it, because on this path that case does not exist by construction.

Smaller point on the cumulative-cap tests: `measureSingleScriptUsed` calls `ValidateTxConway` on the trust path with `require.NoError`, so all three fail inside the helper rather than at their own assertions when the gate is neutered. Of the 4 failures a revert produces, only 2 are meaningful.

wolf31o2 (changes requested):
Blocking: this accepts blocks cardano-node rejects.

cardano-ledger uses the declared per-redeemer ExUnits as the CEK budget on the apply path — `mkPlutusWithContext` takes `exUnits` from `Map.lookup redeemerIndex redeemers`, `evaluatePlutusRunnable` passes it as the budget, and exceeding it is `EvaluationError` -> `Fails` -> `ValidationTagMismatch`. `validateExUnitsTooBigUTxO` is phase-1 on declared totals only. Running at `MaxTxExUnits` with no post-check is `evalTxExUnits`, the cli estimation API.

`TestProducerTrust_ApplyPathAcceptsBudgetOverage` is such a block: declared 0/0, metered 112100/800, `require.NoError`. Min-fee uses declared ex-units, so it also underpays.

Root cause is upstream and this masks it: `Conway/TxInfo.hs:837-839` uses `strictUpperBound` for a TTL-only validity interval, shadowing Alonzo's `PV1.to`; gouroboros passes `closed = !lowerBoundPresent`. That is a ScriptContext data divergence, so the trust window covers only the cost-only subset of the failure class. Filed as blinklabs-io/gouroboros#2171.

Also blocking: `state.go:4762` is unpinned (F4 inline), and `ARCHITECTURE.md:2147-2150` and `:2334` now contradict the code.

Path to merge: fix gouroboros#2171 era-gated, bump, drop the trust path. If a net is still wanted, telemetry-only — log the disagreement, keep rejecting.

chadaibbhot-git:
Thanks Chris — the root-cause pointer is spot on and reshaped this PR entirely.

**Root cause, confirmed.** The "overage" was never dingo needing to trust an over-budget block — it was the ScriptContext divergence you named: Conway uses `strictUpperBound` for a TTL-only validity interval while gouroboros computed `closed = !lowerBoundPresent`, so the datum (and thus the metered cost) differed from the network's at the boundary. gouroboros#2170 era-gates that closure (#2171 is the same defect). With the corrected context the locally metered ex-units match the producer's declared budget, so restrictive phase-2 accepts the canonical block with **no trust path**.

**What I did** (the dep is bumpable — #2170 merged as `b24cbc7`):
- Bumped gouroboros to the pseudo-version carrying #2170 (`v0.202.5-0.20260902123911-b24cbc793974`; no tagged release includes it yet) and wired the new `NewTxInfoV{1,2}FromTransaction` era flag — `false` in Alonzo/Babbage, `true` in Conway+.
- **Dropped the trust path entirely.** `ledger/view.go`, `ledger/state.go`, and `ledger/eras/validation.go` are byte-identical to pre-PR strict; `conway.go` differs from strict only by the two era-flag args. Apply, mempool, forging and replay are all strict again — a genuinely over-budget block is rejected exactly as cardano-ledger does (`EvaluationError` → `ValidationTagMismatch`). No `trustProducerPlutusBudget`, no telemetry acceptance, and the unpinned `state.go` closure is gone (`ledgerProcessBlocks` calls `ledgerProcessBlocksFromSource` directly again).
- Replaced `TestProducerTrust_ApplyPathAcceptsBudgetOverage` (its synthetic declared-0/0 case was exactly the block cardano-node rejects) with `TestConwayPhase2RejectsBudgetOverage` (asserts rejection with the real `script exceeded declared budget` error) and `TestConwayPhase2AcceptsDeclaredBudgetCoveringMeteredCost` (declared covers metered cost → strict accept). Removed the cumulative-cap and replay-trust tests.
- ARCHITECTURE.md :2147-2150 / :2334: the contradiction is resolved by the revert; added one sentence at the validity-interval paragraph documenting the era-gated upper-bound closure (#2170/#2171) as the real fix.

`go build ./...`, `go test ./ledger/ ./ledger/eras/`, gofmt and vet are green. Happy to repin from the pseudo-version to the tagged gouroboros release once Blink Labs cuts one carrying #2170 — go.mod-only, no code change.

wolf31o2:
The upstream root cause is fixed and released: blinklabs-io/gouroboros#2171 landed as 924de23 and shipped in gouroboros v0.202.5.

It is era-gated as required — `StrictValidityUpperBoundForTransaction` returns `tx.Type() >= 6`, so Alonzo and Babbage keep the closed upper bound and Conway onward get `strictUpperBound`, wired at the TxInfoV1/V2 call sites in `ledger/conway/rules.go` and `ledger/dijkstra/rules.go`.

That is the first item from my review: fix gouroboros#2171 era-gated, bump, drop the trust path. A bump to v0.202.5 or later should remove the ScriptContext data divergence behind blinklabs-io/dingo#3627 without relaxing the declared-budget check, which would otherwise accept blocks cardano-node rejects and re-arm the block-scale amplification.

Note dingo `main` may need the v0.202.5 bump gated on dingo#3745 (committee-state capability).

wolf31o2 (commented):
The premise here is disproven. The overage was not an evaluator miscount.

`TimeRange.ToPlutusData()` in gouroboros encoded a finite validity-interval upper bound as CLOSED whenever no lower bound was present; Conway's translation always encodes it EXCLUSIVE. Both #3627 transactions declare `invalidHereafter` with no `invalidBefore`. The pre/post script contexts differ in exactly one byte of 2216: `d87a80` -> `d87980`.

That accounts for the overage exactly, not approximately:

| term | Δcalls | Δcpu | Δmem |
|---|---|---|---|
| `equalsInteger` | +2 | +104,666 | +2 |
| `ifThenElse` | +2 | +152,098 | +2 |
| CEK steps | +24 | +384,000 | +2,400 |
| total | | +640,764 | +2,404 |

`832074315/3057846` used vs `831433551/3055442` declared. All 24 other builtins were byte-identical in cost. Preprod and Preview show the identical absolute delta, which is a fixed-count divergence rather than a per-step overcharge.

Fixed by gouroboros `b24cbc7`, released in `v0.202.5`; dingo `origin/main` pins it as of #3802. Reproduction matrix: v0.202.2/3/4 over-bill on plutigo v0.4.0 and v0.5.0 alike; v0.202.5 bills exactly the declared units on both. Regression fixtures in #3819.

Two consequences for this change:

1. It is no longer needed. There is no residual evaluator discrepancy to trust around.
2. It would be actively harmful now. Skipping the `used > declared` post-check on the followed-chain apply path makes dingo accept genuinely over-budget blocks that cardano-node rejects — the class of defect just fixed would pass silently.

It also would not have unblocked the second form of #3627 (`5728f557...` mint[0], "error explicitly called"), which is a script failure rather than a budget overage, so `restrictive=false` does not reach it. Same one-byte cause; post-fix it succeeds at exactly `153143530/431005`.

Recommend closing.

wolf31o2:
Closing: the premise is disproven and the fix is already released. The overage was gouroboros' `TimeRange.ToPlutusData()` encoding a finite validity-interval upper bound as closed where Conway encodes it exclusive — one byte of the script context, accounting for the +640,764 cpu / +2,404 mem exactly. Fixed in gouroboros `b24cbc7`, released in v0.202.5, pinned on dingo main as of #3802. Regression fixtures in #3819.

Trusting producer budgets on the apply path would now make dingo accept genuinely over-budget blocks that cardano-node rejects. Reopen if a real evaluator discrepancy turns up.

wolf31o2:
Superseded. The root cause identified on this branch — the Conway ScriptContext translating a TTL-only validity interval with an inclusive upper bound — is fixed upstream in gouroboros #2174 (`924de23`), released in `v0.202.5` and adopted by #3802 (`b55067ae`). `main` calls `script.StrictValidityUpperBoundForTransaction` at `ledger/eras/alonzo.go:266`, `ledger/eras/alonzo.go:418` and `ledger/eras/conway.go:1020`, so the era gate no longer needs the hardcoded per-call-site booleans here. Shipped in `v0.70.5` (`7d849b81`).

The producer-trust path is not on `main`; the post-execution declared-budget comparison is retained.

#3627 stays open. Confirming Preprod no longer rejects the affected transactions needs a Mithril-backed Preprod replay.

Coverage note: the overage and slippage-batch assertions on this branch are covered on `main` at `ledger/eras/validation_test.go:485` and `ledger/eras/conway_plutus_budget_test.go:138`. `TestConwayPhase2RejectsGenuineScriptFailure` has no equivalent on `main` and is worth porting separately.

chadaibbhot-git:
Thanks for running this all the way down — the gouroboros `TimeRange.ToPlutusData()` closed-vs-exclusive upper-bound encoding is a clean root cause, and the +640,764 cpu / +2,404 mem it accounts for matches what we saw. Agreed the close is right: we are bumping our downstream carry to gouroboros v0.202.5 and dropping the budget-overage trust (it would now let genuinely over-budget blocks through). Our reward-withdrawal trust stays only until #3628 is resolved. Appreciate the thorough writeup and the regression fixtures in #3819.

PR #3747, opened 2026-09-01
ci(release): homebrew update

### LINKED ISSUE #3683: Add homebrew update to releases
This should be non-blocking, but we should update our homebrew cask.

tripura-repalle:
Closes: #3683 

---
## Summary by cubic
Adds a non-blocking Homebrew formula bump to the release workflow so every tag opens a PR in `blinklabs-io/homebrew-cardano` with the updated `dingo` formula (`url`, `sha256`, `CommitHash`). Pins the release commit hash to 7 chars so the formula and released binaries always match.

- Runs only on tags after `finalize-release` succeeds; `continue-on-error` keeps the workflow green if the tap PR cannot be opened.
- No-ops when `ORG_PROJECT_PAT` is not set, so forks and secret-less runs skip the job.
- Verifies every formula edit applied so a stale or renamed formula fails loudly instead of shipping silently.
- Reuses an existing open PR for the same tag and force-pushes the updated formula to it.
- Pins `Makefile` `COMMIT_HASH` to `--short=7 HEAD` so `git rev-parse --short` can't auto-extend past 7 characters on ambiguous prefixes.

**Rollout**

- Set the `ORG_PROJECT_PAT` secret (with `contents: write` and `pull-requests: write` on `blinklabs-io/homebrew-cardano`) for releases to open tap PRs.

<sup>Written for commit 6a04ef5a2238579b14acb3cdb2981c2a10dd8ea7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3747?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Homebrew package metadata is now automatically updated when a release is published.
  * Release updates can be submitted for review through an automatically created or reused pull request.
  * Repeated release runs avoid unnecessary updates when the formula is already current.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -554,7 +554,10 @@ jobs:
           {
             echo "tarball_url=$tarball_url"
             echo "sha256=$sha256"
-            echo "short_commit=${COMMIT_SHA:0:8}"
+            # Match the release build convention (Makefile derives COMMIT_HASH
+            # from `git rev-parse --short HEAD`, i.e. 7 chars) so the Homebrew
+            # build stamps the same version.CommitHash as the released binaries.
+            echo "short_commit=${COMMIT_SHA:0:7}"
           } >> "$GITHUB_OUTPUT"
       - name: Checkout homebrew tap
         if: steps.gate.outputs.enabled == 'true'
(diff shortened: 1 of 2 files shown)


wolf31o2 (changes requested):
The required Linux test check is failing on the current head, so this is not ready to merge. Please rerun or fix the failure before requesting approval.

PR #3750, opened 2026-09-01
fix(peergov): bound peer sharing and rejected peer state

### LINKED ISSUE #3398: peergov: bound peer-sharing, DNS dialing, and rejected-address state
Peer input and dialing state must honor protocol and lifecycle bounds:

- `ouroboros/peersharing.go` accepts a reply longer than requested and serializes DNS-hostname peers with a nil IP.
- `peergov/peers.go` resolves DNS on every dial attempt and chooses a random address without a timeout/backoff bound.
- `peergov/peergov.go` records ledger addresses before duplicate and hard-cap rejection, retaining rejected entries.

Acceptance criteria:

- Enforce requested reply counts and serialize a valid address for every accepted peer.
- Bound DNS resolution/dial retries and avoid retaining rejected addresses.
- Cover duplicate, cap, DNS, and malformed-peer vectors.

arepala-uml:
## Summary

- Limit peer-sharing replies to the requested number of valid peers.
- Prevent malformed or unresolved addresses from producing nil-IP peer entries.
- Preserve bounded DNS resolution, dial timeouts, cancellation, and reconnect backoff.
- Avoid retaining duplicate, capacity-rejected, non-routable, or canceled ledger addresses.
- Preserve upstream’s pre-DNS duplicate detection without mutating rejected-peer state.
- Add regression coverage for reply limits, malformed peers, DNS behavior, duplicates, capacity limits, and ledger-address retention.

### `ouroboros/peersharing.go`

- Return an empty reply for zero or negative request amounts.
- Stop collecting peers after the requested number of valid peers is reached.
- Prefer each peer’s normalized address when constructing the response.
- Validate the host as an IP address before accepting it.
- Validate that the port fits the protocol’s `uint16` range.
- Skip malformed, unresolved, or otherwise unserializable peers without consuming the requested reply count.

### `ouroboros/peersharing_test.go`

- Cover peer sharing before the peer governor is initialized.
- Cover negative and zero requested counts.
- Cover exact and larger-than-available requested counts.
- Verify unsharable peers are excluded.
- Cover malformed host-port values, non-IP hosts, and out-of-range ports.
- Cover valid IPv4 and IPv6 peers.
- Verify every returned peer can be CBOR-serialized.

### `peergov/ledger_discovery.go`

- Keep upstream’s early duplicate and deny-list checks that avoid unnecessary DNS lookups.
- Preserve case-insensitive hostname duplicate detection.
- Remove rejected-duplicate mutation from the pre-DNS rejection helper.
- Record an address in `ledgerKnownAddrs` only after duplicate and hard-cap checks pass.
- Prevent rejected candidates from incorrectly counting toward the ledger-peer target.

### `peergov/ledger_discovery_resolution_test.go`

- Verify known peers are rejected without another DNS lookup.
- Verify case-insensitive duplicate matching.
- Verify rejecting a known peer does not retain it as ledger-discovered.
- Preserve coverage for denied hostnames and bounded ledger resolution.

### `peergov/peer_cap_test.go`

- Verify a ledger peer rejected at the hard cap is not retained.
- Verify a duplicate ledger peer is not retained.
- Preserve peer-count assertions after rejection.

### `peergov/peergov_test.go`

- Verify non-routable ledger peers are not retained.
- Verify an accepted ledger peer is retained.
- Cover both the accepted and rejected sides of the ledger-address state contract.

## Validation

- `go test -race ./ouroboros/ ./peergov/`
- `go test -race ./peergov/` after merge resolution
- `go test ./internal/test/conformance/`

Closes #3398 

---
## Summary by cubic
Bounds peer-sharing replies to the requested number of valid peers and stops retaining rejected ledger addresses in discovery state.

**Peer sharing**
- Returns an empty reply for zero or negative request amounts.
- Stops collecting peers once the requested count is reached.
- Skips peers with malformed host:port, non-IP hosts, or out-of-range ports.
- Uses each peer's normalized address when available.

**Ledger discovery**
- Records an address in `ledgerKnownAddrs` only after duplicate and capacity checks pass.
- Rejected duplicate or capacity-limited peers no longer count toward the ledger peer target.

Adds regression tests for these behaviors. Closes #3398.

<sup>Written for commit 7f7d771b32c99db4174299882aab505b4c11dcf8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3750?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


arepala-uml:
Closing this draft PR because its commits are missing the required DCO sign-offs.

PR #3760, opened 2026-09-01
fix: reject invalid access points and valencies

### LINKED ISSUE #3291: topology: reject invalid access points and valencies
## Problem

- `topology.NewTopologyConfigFromReader` performs JSON decoding without semantic validation.
- Access points can have empty addresses or ports outside 1-65535.
- `warmValency` can exceed `valency`.
- Valency can exceed the available access points.

## Acceptance criteria

- Validate local-root, public-root, bootstrap, and snapshot-relay access points.
- Require non-empty addresses and ports in 1-65535.
- Require `warmValency <= valency <= len(accessPoints)`.
- Return field-specific errors.
- Add boundary tests for each topology section.

jkawan:
closes #3291 

---
## Summary by cubic
Closes #3291 by rejecting invalid access points and valencies while parsing topology configs instead of accepting malformed values. Empty root access-point lists remain valid.

- Rejects blank addresses and out-of-range ports with indexed validation errors.
- Requires `warmValency` to be no greater than `valency` when set.
- Requires `valency` to be no greater than the access-point count when access points are present.
- Applies validation to local roots, public roots, and bootstrap peers.

<sup>Written for commit 6fabb1044c412a884ea366d8957b51b43490890f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3760?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for topology configuration values, including access-point addresses, port ranges, valency, and warm valency.
  * Invalid relay and bootstrap peer configurations are now rejected with clear validation errors.
  * Updated topology parsing to require valid local-root access points.

* **Tests**
  * Expanded coverage for invalid addresses, ports, and valency constraints across topology and peer snapshot configurations.


topology/topology.go
```diff
@@ -154,9 +158,137 @@ func NewPeerSnapshotConfigFromReader(
...
+	return nil
+}
+
+func validateRootValencies(
+	fieldPrefix string,
+	warmValency uint,
+	valency uint,
+	accessPointCount int,
+) error {
+	if warmValency > 0 && valency > warmValency {
```
wolf31o2: P1 — This comparison is reversed relative to the linked issue’s acceptance criterion `warmValency <= valency`. It rejects the valid configuration `valency: 2, warmValency: 1` and accepts the invalid `valency: 1, warmValency: 2`. The error text and the added test at `topology_test.go:384` encode the same inversion. Change the guard/test so warm valency cannot exceed valency, and add a valid `warmValency < valency` case.

wolf31o2 (changes requested):
Fresh review of current head `cb1627dc278352d34b0182be66b4ed08b797520a`. Found one semantic blocker in the valency ordering validation.

topology/topology.go
```diff
@@ -306,6 +309,105 @@ func NewPeerSnapshotConfigFromReader(
...
+	return nil
+}
+
+func validateRootValencies(
+	fieldPrefix string,
+	warmValency uint,
+	valency uint,
+	accessPointCount int,
+) error {
+	if warmValency > valency {
```
wolf31o2: P1: the ordering is inverted relative to ouroboros-network.

`LocalRootPeers.invariant` requires `getWarmValency w >= getHotValency h` (`ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/LocalRootPeers.hs`, lines 111-119: "the warm target needs to be greater than or equal to the hot target"). `Ouroboros/Network/Diffusion/Topology.hs:41-45` documents `warmValency` as the warm/established target and `valency`/`hotValency` as the hot target, with hot a subset of warm. The JSON parser defaults `warmValency` to `WarmValency v` where `v` is `valency` (`Ouroboros/Network/OrphanInstances.hs:147-158`), so `warmValency > valency` is the only meaningful non-default value.

Measured at this head:

- `{"localRoots":[{"accessPoints":[{"address":"a.example.com","port":3001},{"address":"b.example.com","port":3001}],"valency":1,"warmValency":2}]}` → rejected with `localRoots[0].warmValency must be <= localRoots[0].valency`. cardano-node accepts this.
- Same access points with `"valency":2,"warmValency":1` → accepted. The upstream invariant forbids this.

This supersedes the request in review 5091322732 and the acceptance criterion `warmValency <= valency` in #3291; that criterion does not match ouroboros-network and needs updating along with the guard, the error text, and the tests at topology_test.go:386, :435, and :510.

topology/topology.go
```diff
@@ -306,6 +309,105 @@ func NewPeerSnapshotConfigFromReader(
...
+		return fmt.Errorf(
+			"%s.warmValency must be <= %s.valency",
+			fieldPrefix,
+			fieldPrefix,
+		)
+	}
+	if accessPointCount > 0 && valency > uint(accessPointCount) {
```
wolf31o2: P1: `valency <= len(accessPoints)` rejects topology files cardano-node accepts.

An access point may be a DNS name that resolves to several addresses — `Ouroboros/Network/PeerSelection/RootPeersDNS/LocalRootPeers.hs` keys resolution as `Map RelayAccessPoint [peerAddr]` — and the upstream bound is on the resolved group, `getWarmValency w <= Set.size g` (`State/LocalRootPeers.hs:118`), not on the access-point count.

Measured at this head: `{"localRoots":[{"accessPoints":[{"address":"relays.pool.example","port":3001}],"valency":2}]}` fails with `localRoots[0].valency must be <= len(localRoots[0].accessPoints)`. `NewTopologyConfigFromReader` returns that error through `NewTopologyConfigFromFile` to `LoadTopologyConfigFor` (internal/config/config.go:1645), so a single-DNS-relay group with valency 2 stops node startup instead of under-filling its target.

The `accessPointCount > 0` guard also makes the rule self-inconsistent: `config/cardano/{mainnet,preprod,preview,musashi}/topology.json` declare `"accessPoints": []` with `"valency": 1` and are accepted (verified via `NewTopologyConfigFromFS` against the embedded FS), while one access point with valency 2 is rejected.

topology/topology.go
```diff
@@ -306,6 +309,105 @@ func NewPeerSnapshotConfigFromReader(
 	return s, nil
 }
 
+func validateAccessPoint(
+	ap TopologyConfigP2PAccessPoint,
+	field string,
+	requirePort bool,
```
wolf31o2: P3: `requirePort` is `true` at all three call sites (localRoots, publicRoots, bootstrapPeers). The `false` branch and the `range 0-65535` message at line 327 are unreachable.

topology/topology_test.go
```diff
@@ -320,6 +340,209 @@ func TestNewTopologyConfigFromReader_OversizedInput(t *testing.T) {
...
+			"warmValency": 1
+		}
+	]
+}`
+
+	_, err := topology.NewTopologyConfigFromReader(strings.NewReader(jsonData))
+	require.NoError(t, err)
+}
+
+func TestNewPeerSnapshotConfigFromReader_AllowsMissingRelayPort(t *testing.T) {
```
wolf31o2: P2: `NewPeerSnapshotConfigFromReader` performs no semantic validation, so this asserts `NoError` from a plain JSON decode and passes for any well-formed snapshot — it does not exercise relay-port handling. The behavior it names is also the opposite of `PeerSnapshotConfig.Validate`, which rejects a portless relay with `SRV relay mode is not supported; port must be specified` (topology.go:157-161) and is what node startup actually runs (config.go:566). Either assert against `Validate` or drop the case.

topology/topology_test.go
```diff
@@ -320,6 +340,209 @@ func TestNewTopologyConfigFromReader_OversizedInput(t *testing.T) {
...
+    {
+      "accessPoints": [{"address": "public.example.com", "port": 65536}],
+      "valency": 1
+    }
+  ]
+}`,
+			wantErr: "publicRoots[0].accessPoints[0].port must be in range 1-65535",
+		},
+		{
+			name: "public root valency exceeds warm valency when set",
```
wolf31o2: P3: the name says valency exceeds warm valency; the body sets `valency: 1, warmValency: 2`. The local-root twin was renamed at line 386, this one was not.

wolf31o2 (changes requested):
Two of the three new rules reject topology files that cardano-node accepts, and one accepts a configuration ouroboros-network's own invariant forbids. Details inline.

topology/topology.go
```diff
@@ -306,6 +309,96 @@ func NewPeerSnapshotConfigFromReader(
...
+}
+
+// validateRootValencies checks that valency (the hot/active target) does not
+// exceed warmValency (the warm/established target), matching the
+// ouroboros-network invariant that the warm target must be >= the hot target
+// (LocalRootPeers.hs: getWarmValency w >= getHotValency h). A warmValency of
+// zero means it was not set in the config; cardano-node's parser defaults it
+// to valency in that case, so no comparison is needed.
+//
+// Note: valency is intentionally not bounded by the number of configured
```
wolf31o2, line 332: P1: This still does not implement #3291's explicit `valency <= len(accessPoints)` acceptance criterion. With one configured access point and `valency: 2`, `NewTopologyConfigFromReader` now succeeds (and `TestNewTopologyConfigFromReader_AllowsValencyAboveAccessPointCount` locks in that behavior), even though the requested contract says this must be rejected. Please either enforce the criterion and test the rejection, or update the issue/PR contract before merging this behavioral change.

wolf31o2 (changes requested):

topology/topology.go
jkawan: addressed

wolf31o2 (approved):

Issue #3623, opened 2026-08-29
leiosfetch: EB backfill parks indefinitely in `requestSlot.acquire`, halting chain sync


chadaibbhot-git:
## Problem

On `v0.70.2`, the Leios endorser-block backfiller parks forever waiting for a
leiosfetch request slot. The node stays up, keeps accepting connections and
answering handshakes, but its tip never advances again. A goroutine dump shows
the backfiller blocked in `leiosfetch.(*requestSlot).acquire`.

Two independent block producers on the same private Leios testnet reproduced
this within minutes of each other, from separate fresh from-genesis syncs on
separate hosts, both halting at essentially the same chain position
(block 18759 and 18757).

`v0.70.0` — which pins gouroboros `v0.196.0` instead of `v0.202.2` — syncs
straight past that position on the same hosts, same network, same topology.

## Evidence

Build: `v0.70.2` (`11183ab07a80458e1eb869db4dd262b7d6598a22`), gouroboros
`v0.202.2`, `--run-mode leios`, `consensus_mode=cpraos`, `security_param=432`.

Goroutine dump captured from a stalled node via the pprof endpoint
(`DINGO_DEBUG_PORT`), while the tip had been frozen for over two hours:

```
goroutine 24105 [select]:
github.com/blinklabs-io/gouroboros/protocol/leiosfetch.(*requestSlot).acquire
        gouroboros@v0.202.2/protocol/leiosfetch/client.go:70
github.com/blinklabs-io/gouroboros/protocol/leiosfetch.(*Client).BlockTxsRequest
        gouroboros@v0.202.2/protocol/leiosfetch/client.go:290
github.com/blinklabs-io/dingo/ouroboros.(*Ouroboros).fetchLeiosEbTxsBatchedUntilWithValidator
        ouroboros/leiosnotify.go:1209
github.com/blinklabs-io/dingo/ouroboros.(*Ouroboros).fetchLeiosEbTxsBatchedUntil
        ouroboros/leiosnotify.go:1128
github.com/blinklabs-io/dingo/ouroboros.(*Ouroboros).fetchEndorserBlockOnConn
        ouroboros/leios_backfill.go:244
github.com/blinklabs-io/dingo/ouroboros.(*Ouroboros).FetchEndorserBlockByPoint
        ouroboros/leios_backfill.go:153
github.com/blinklabs-io/dingo.(*Node).ledgerStateConfig.func2
        node_ledger_config.go:85
[...]

PR #3772, opened 2026-09-02
fix(leiosfetch): decline block-range requests, do not leave pending

wolf31o2:
`leiosfetchServerBlockRangeRequest` was `// TODO; return nil`. gouroboros reads nil as "async range send started" (`leiosfetch.go:149`), so dingo held agency in `StateBlockRange` forever and wedged any peer that probed it. A range request has no absence reply, so it now errors.

Also fixes two comments claiming `WithTimeout` bounds a response-less request; it does not reach `StateBlock`/`StateBlockTxs`.

Fails without the fix: `TestLeiosFetchBlockRangeRequestIsDeclined` — "left pending instead of declined".

Client side needs blinklabs-io/gouroboros#2173 plus a bump. That bump must also update `TestLeiosBlockTxsAbandonedSlotFailsOverAndBecomesAvailable`, which asserts the poisoned connection stays in the manager.

Refs #3623

---
## Summary by cubic
Fixes leios-fetch block-range requests being left pending forever, which permanently wedged any peer that probed them (#3623). The server returned nil from the range callback, which gouroboros reads as "async range send started," so it held agency in `StateBlockRange` indefinitely; it now declines with an explicit error, failing the connection observably instead of hanging silently.

**Bug Fixes**
- Adds tests covering the range-decline fix and confirming unavailable-but-answerable requests still get answered on the wire.
- Corrects two comments claiming the response timeout bounds response-less requests; only the request context does.

Requires `blinklabs-io/gouroboros` #2173 plus a dependency bump, which must also update `TestLeiosBlockTxsAbandonedSlotFailsOverAndBecomesAvailable`.

<sup>Written for commit d073ad6375d20f8049b565d4482e140d97b94c0d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3772?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Leios block-range requests that cannot be served now return an explicit protocol error instead of leaving the requesting connection stuck.
  * Unavailable transaction-block requests now receive a clear “no blocks” response while keeping the connection active.
  * Leios fetch requests now use a bounded per-request timeout when no caller deadline is provided.

* **Documentation**
  * Updated architecture and protocol documentation to describe timeout handling and declined block-range requests.


ARCHITECTURE.md
```diff
@@ -2328,7 +2328,7 @@ revoke-on-conflict closure, and the associated `MetadataStore` methods
...
... loop waits for agency that only the missing response returns — with no way for that peer to detect it. There is no absence reply for a range request, so declining necessarily fails the leios-fetch connection, which the peer can at least observe and recover from (issue #3623). Pushed votes (notify `MsgVotesOffer` tag 4) are forwarded to the `ledger/leios` vote manager. The LeiosVotes and LeiosFetch vote handlers delegate vote collection, serving, and emission to the `ledger/leios` vote manager (see "Leios Voting"); LeiosFetch vote requests return a valid empty `MsgVotes` when that optional manager is absent, while standalone LeiosVotes remains unavailable and returns its explicit error. A cached or reloaded endorser block is also handed to the `ledger/leios` pipeline manager (see "Leios Pipeline") for stage/timing tracking and equivocation detection. LeiosVotes pull requests remain outstanding while no servable votes exist and complete when new vote material arrives or the protocol shuts down, so an empty relay store is not treated as a mini-protocol error. Durable vote storage is still future work; CertRB production uses the current gouroboros/Dijkstra prototype certificate shape.
```
arepala-uml, line 2331: Non-blocking: this says declining "necessarily fails the leios-fetch connection," but gouroboros shares one error channel across all mini-protocols on a bearer (`connection.go:508,598-604`), so this actually tears down the whole peer connection (chainsync/blockfetch/leios-notify included), not just leios-fetch. That's the same blast radius this paragraph already calls out for the earlier leios-votes/protocol-20 incident ("taking chainsync/blockfetch with it") — worth wording this decline the same way so an operator debugging reconnect churn doesn't under-scope the impact to leios-fetch alone.

ouroboros/leiosfetch.go
```diff
@@ -186,11 +191,34 @@ func (o *Ouroboros) leiosfetchServerVotesRequest(
...
+// connection. An observable connection error is recoverable by the peer's
+// governance; a silent hang is not.
 func (o *Ouroboros) leiosfetchServerBlockRangeRequest(
 	ctx oleiosfetch.CallbackContext,
 	start ocommon.Point,
 	end ocommon.Point,
 ) error {
-	// TODO
-	return nil
+	return fmt.Errorf(
```
arepala-uml, line 217: Non-blocking: this `%d.%x` slot.hash formatting is now duplicated across 6 call sites in this package (4 errors here and in `leiosfetchServerBlockRequest`/`leiosfetchServerBlockTxsRequest`, 2 log lines in `leiosnotify.go:655,795`), with no shared helper. A small `pointStr(p ocommon.Point) string` would keep them in sync if the format ever changes. Optional cleanup, not blocking.

arepala-uml (approved):
Verified the fix against gouroboros v0.202.4: a `nil` return from `BlockRangeRequestFunc` really does leave the server parked in `StateBlockRange` forever (no absence-reply message exists for ranges on the wire), so returning an explicit error is correct and necessary. The new test proves it — fails on the pre-fix parent commit, passes under `-race` at head. No correctness issues, and repo conventions (DCO, Conventional Commits, shared `testutil` sync helpers, doc-parity) are all followed.

One suggested follow-up before merge, plus two optional nits, all inline. The follow-up isn't in this PR's diff so noting it here: `peergov/connections.go:894-909`'s `isExpectedConnectionCloseError` doesn't match the new decline's error text, so this expected, policy-driven decline will log as `Error`-level "unexpected connection failure" (and count toward the same metric bucket as genuine leios-fetch faults) instead of the `Info` level used for expected closes. Worth a quick follow-up to add this error to the expected-close check or use a typed sentinel.

PR #3725, opened 2026-08-31
fix(ledger): bound LocalStateQuery work per request

arepala-uml:
## Summary

Bounds caller-controlled LocalStateQuery requests that previously performed database work for every supplied element. This prevents oversized requests from monopolizing the query path and producing unpredictable latency.

Requests containing up to 1,000 items are accepted. Oversized requests are rejected before database or consensus-state access using a defined, inspectable error.

Existing response construction, partial-result behavior, empty/all query forms, and protocol encoding remain unchanged.

The limit is applied specifically to `GetStakeSnapshots`, `GetDRepState`, `GetStakeDelegDeposits`, and `GetFilteredVoteDelegatees` because these handlers perform database work per caller-provided item; other collection-based queries use bulk or state-bounded database operations and are not affected by this amplification pattern.

### `ledger/queries.go`

- Added `MaxLocalStateQueryItems` with a limit of 1,000 items.
- Added the `ErrLocalStateQueryLimitExceeded` sentinel error.
- Added `LocalStateQueryLimitError`, containing:
  - Query name
  - Submitted item count
  - Maximum allowed item count
- Added a shared preflight item-limit check.
- Applied the limit to:
  - `GetStakeSnapshots`
  - `GetDRepState`
  - `GetStakeDelegDeposits`
  - `GetFilteredVoteDelegatees`
- Ensured oversized requests are rejected before database transactions, database reads, consensus-state access, or per-item processing.
- Preserved unrestricted empty/all query forms where supported.

### `ledger/queries_limit_test.go`

- Added boundary coverage confirming:
  - Exactly 1,000 items are accepted.
  - 1,001 items are rejected.
- Verified the returned error supports both:
  - `errors.Is` with `ErrLocalStateQueryLimitExceeded`
  - `errors.As` with `LocalStateQueryLimitError`
- Verified error details include the correct query name, submitted count, and limit.
- Covered oversized requests for every affected handler.
- Verified rejection occurs before database or consensus-state access.
- Added repeated-request coverage using 100 consecutive oversized requests.
- Verified a normal-sized request is accepted after repeated oversized requests.
- Added descriptive comments explaining the behavior protected by each test.

### `ARCHITECTURE.md`

- Documented the 1,000-item LocalStateQuery limit.
- Documented the affected query handlers.
- Documented early rejection and the defined error contract.
- Clarified that existing empty/all, ordering, and partial-result behavior remains unchanged.

Closes #3519 

---
## Summary by cubic
[...]

PR #3782, opened 2026-09-02
fix(database): eliminate O(n²) account-lookup stall and bound worker-pool shutdown

wcatz:
## Summary

Diagnosed and fixed the root cause of a multi-hour mainnet block-producer stall at an epoch boundary. Investigation via a live goroutine dump (`kill -QUIT`) traced the hang to `rewardActiveAccounts` → `MetadataStore.GetAccountsByCredential`, well before `reward_calc_pool_total`/`reward_calc_pool_current` were ever set — which is why the stall was previously misread as a slow reward-calculation loop rather than a slow account lookup ahead of it.

- **`GetAccountsByCredential`** built an OR-chain of `(credential_tag = ? AND staking_key = ?)` predicates per chunk. SQLite's planner can't drive an OR-chain through `idx_account_credential`, so each chunk fell back to a full table scan — effectively O(n²) in account-table row count. Rewritten as a `UNION ALL`-of-literal-selects derived table, inner-joined against `account`, which all three supported dialects plan as an indexed seek per row. Benchmarked on a real migrated schema: 50,000 refs 139s → 2.8s (50x), 300,000 refs (never completed in 45 min before) → ~93s.
- **`DatabaseWorkerPool.Shutdown`** had no internal timeout. `LedgerState.Close()` already bounds its *own* wait around the goroutine that calls `Shutdown`, but `Shutdown` itself blocked unconditionally on `operationWg.Wait()` — so once `Close()`'s outer wait gave up and returned, the abandoned goroutine kept running for as long as the outstanding (slow) operation took. This is exactly what let the account-lookup bug turn into an hours-long background leak instead of a bounded failure. `Shutdown` now takes a `drainTimeout` and bounds its own wait, returning an error instead of blocking indefinitely.

## Changes

- `database/plugin/metadata/sqlstore/account.go`: rewrite `GetAccountsByCredential`'s query shape (OR-chain → indexed `UNION ALL` join), chunked at `ParameterLimit()/2`.
- `database/plugin/metadata/sqlstore/account_bench_test.go`: new benchmark exercising the real migrated schema at 1K/50K/300K rows.
- `ledger/state.go`: `DatabaseWorkerPool.Shutdown(drainTimeout time.Duration) error` — bounded drain wait; `LedgerState.Close()`'s call site now joins the error into `Close`'s return.
- `ledger/state_test.go`: updated all `Shutdown()` call sites for the new signature; added `TestDatabaseWorkerPoolShutdownTimesOutOnSlowOperation`, exercising the new timeout/error-return path directly.
- `DATABASE.md`: documents `GetAccountsByCredential`'s query shape and the OR-chain pitfall.
- `ARCHITECTURE.md`: extends the existing `Close()` bounded-wait documentation to cover `Shutdown`'s own internal bound.

## Test plan

- [x] `go test -race ./database/...` — full suite passes
- [x] `go test -race ./ledger/...` — full suite passes
- [x] `go test -race -run "WorkerPool|Shutdown|DatabaseWorkerPool" -count=3 -v ./ledger/...` — 3x, catches and confirms the fixed data race
- [x] `golangci-lint run ./database/plugin/metadata/sqlstore/... ./ledger/...` — clean
- [x] `nilaway` — no findings in changed files
- [x] `modernize`, `golines` — clean
- [x] `make docs-parity` — passes
- [x] Benchmarked query rewrite against real migrated schema at 1K/50K/300K rows (see commit message and DATABASE.md)

---
## Summary by cubic
Replaces the O(n²) epoch-boundary account lookups and reward-output ID resolution with indexed queries, and bounds `DatabaseWorkerPool.Shutdown` so a slow in-flight query can no longer leak a goroutine past `LedgerState.Close` or have storage closed under it during startup rollback.

**Database**

- Account lookup now uses the indexed grouped `staking_key IN` query merged from `main`; real-schema benchmarks drop 50,000 refs from ~139s to ~2.8s, and 300,000 refs complete in ~93s.
- `SaveRewardAccountOutputs` resolves generated IDs through the same derived-table-join pattern, casting to `BIGINT`/`BYTEA` on PostgreSQL, with cross-dialect integration coverage.

**Shutdown and cleanup**

- `Shutdown` drains in-flight operations with a bounded wait and returns an error on timeout, with no waiter goroutine left behind.
- `LedgerState.Close` joins shutdown errors, and startup rollback skips `db.Close` and plugin shutdown when the drain is unconfirmed.
- Regression tests poll for goroutine cleanup to avoid runtime flakiness; `ARCHITECTURE.md` and `DATABASE.md` document the new query shapes and shutdown behavior.

<sup>Written for commit de96dfa3695722204c52e4ff2653b811e34fcd7b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3782?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Database shutdown now enforces a configurable timeout for in-flight operations and reports errors when they do not finish in time.
  * Ledger closure reports shutdown failures and records the outcome.

* **Performance**
  * Credential-based account lookups are more efficient, particularly for larger batches.

* **Bug Fixes**
  * Startup cleanup avoids closing storage while background operations may still be using it.

* **Documentation**
  * Added guidance on account lookup behavior, query performance, and shutdown timeout handling.


wcatz:
Addressed in 762be323: the risk CodeRabbit flagged was real, and specifically in the startup-failure rollback path (`Run()`'s `cleanupFailedStartup`), not the normal signal-driven shutdown path. `node_shutdown.go`'s `shutdown()` already guards `db.Close()`/`pluginHost.Stop()` behind a `ledgerStateDrainConfirmed` flag on the normal path — but `Run()`'s LIFO `started` rollback discarded `ledgerState.Close()`'s return value entirely, so on a startup failure (e.g. a later component failing to start), a `DatabaseWorkerPool.Shutdown` timeout was invisible and the later-run `db.Close()`/`pluginHost.Stop()` stops closed storage regardless of whether a background goroutine was still using it.

Fixed by mirroring the exact same `ledgerStateDrainConfirmed` pattern in `Run()`'s startup-failure rollback: a `Run()`-scoped flag, set false by the `ledgerState.Close` stop on a non-nil error, makes the `db.Close`/`pluginHost.Stop` stops log and skip instead of running. Added `TestCleanupFailedStartupSkipsDatabaseCloseWhenLedgerDrainIsUnconfirmed` covering this path directly, alongside the existing `TestShutdownDoesNotCloseDatabaseWhenLedgerDrainIsUnconfirmed` for the normal path.

wcatz:
@cubic-dev-ai P1 (account.go:333, bytea = text): confirmed and fixed in 0c4f4c02. Root cause: PostgreSQL resolves an untyped derived-table select-list parameter to `text` rather than inferring it from the joined `account` columns (`credential_tag` BIGINT, `staking_key` BYTEA after the SQLite→Postgres migration translation), so the join failed with `operator does not exist: bytea = text`. Fixed by casting the row-select parameters (`CAST(? AS BIGINT)`/`CAST(? AS BYTEA)`) on the postgres dialect only — SQLite and MySQL infer the type from context already and are unchanged.

Verified against a live local Postgres container: reverted the cast, confirmed the new integration-test check fails with exactly that error, reapplied the cast, confirmed it passes. Also ran the MySQL integration path to confirm no regression there. Added `GetAccountsByCredential` coverage to the shared Postgres/MySQL integration test helper in `dialect_integration_test.go` so this can't regress silently again — the SQLite benchmark alone can't catch a dialect-specific type-inference bug.

wcatz:
@cubic-dev-ai P2 (state.go:418, Shutdown goroutine leak): valid, fixed in ade9732d. You're right that the drainTimeout bound alone didn't fix this — the spawned goroutine stayed blocked in `operationWg.Wait()` for the slow operation's full duration regardless of the timeout, it just no longer blocked Shutdown's *caller*. That's a real leak, not an accepted tradeoff.

Fixed properly rather than declining it: replaced `operationWg sync.WaitGroup` with a mutex-guarded counter plus a `drained` channel that whichever of `Shutdown` or the last finishing operation closes. `Shutdown` now selects `drained` directly with no goroutine spawned at all — timing out leaves nothing running beyond the worker that was already executing the slow operation before Shutdown was ever called. Added a regression test (`TestDatabaseWorkerPoolShutdownTimeoutSpawnsNoWaiterGoroutine`) asserting no goroutine-count growth across a timed-out Shutdown call, run 3x with `-race`. Also corrected a doc comment from the prior commit that inaccurately claimed the old waiter goroutine "gave up on its own" — it didn't; only Shutdown's caller did.

wcatz:
@cubic-dev-ai P2 (node_test.go:642, hand-copied guard): fair concern, addressed with a doc comment rather than a code change (ade9732d).

Checked the precedent you're implicitly pointing at — `TestShutdownDoesNotCloseDatabaseWhenLedgerDrainIsUnconfirmed` does drive the real path by calling `n.shutdown()` directly, but that's possible because `shutdown()`'s phase ordering is hard-coded in that one function. `cleanupFailedStartup`'s ordering isn't hard-coded anywhere — it's purely a property of which `started = append(started, ...)` calls `Run()` happens to have reached before failing, across ~30 such calls interleaved through Run()'s startup sequence, each registered immediately after the resource it tears down becomes available (can't hoist them together without breaking that invariant, or reordering registration relative to unrelated resources). Exercising the *real* registration order would mean driving `Run()` to a genuine failure at a specific point rather than calling one self-contained function, which is a materially larger test than exists anywhere else for this mechanism today.

This also isn't a new pattern I introduced: every existing test that exercises `cleanupFailedStartup` in this file (see the startup-lifecycle-gate test above mine) already hand-builds the `started` slice rather than driving `Run()`, and `newLiveLifecycleTestNode`'s own doc comment documents that as this file's established convention. I added a comment on the test spelling out exactly this limitation — it verifies the guard logic is correct given the order Run() is documented to register it in, but can't catch a future reordering of those three registrations relative to each other. That's a real gap; closing it fully would need either a Run()-driving integration test or restructuring Run()'s startup to expose registration separately, both bigger than this PR's scope.

wcatz:
@coderabbitai review

wolf31o2 (changes requested):
The new UNION ALL derived-table join can return duplicate account rows when refs contains duplicate (credential_tag, staking_key) entries. The previous OR-chain returned each matching account once, so deduplicate the input refs or make the derived relation distinct before joining to preserve GetAccountsByCredential semantics.

wcatz:
@wolf31o2 Good catch, fixed in b8fc07b9.

You're right that the derived-table join emits one row per v-row where the OR-chain matched each account row at most once regardless of how many OR-clauses hit it — mechanically a real difference. For the current map-keyed return value it was harmless (a duplicate ref just reassigns the same map key to identical data), but it wastes derived-table rows and chunk capacity for a caller that passes duplicates, and there's no guarantee some future caller stays map-shaped. Deduplicated refs before querying rather than adding DISTINCT to the SQL, since that also avoids the wasted rows rather than just hiding them after the fact. Added a regression test (`TestGetAccountsByCredentialDeduplicatesRepeatedRefs`) and a DATABASE.md note.

wcatz:
@cubic-dev-ai Correct, fixed in 0c4765b4. The map-shaped result genuinely can't distinguish "deduplicated before querying" from "queried with duplicates, then the map assignment hid it" — my test only proved the latter was harmless, not that dedup happened. Extracted the loop into `dedupeStakeCredentialRefs` and added a direct unit test on it. Verified the new test actually catches the regression: temporarily reverted the function to a no-op passthrough, confirmed the test fails, reapplied the real implementation, confirmed it passes.

wcatz:
@cubic-dev-ai Correct, fixed in 74f182cc. Replaced the single-snapshot comparison with a short polling loop (matching `storagetest.AssertNoGoroutineLeak`'s pattern already in this codebase) that waits for the count to settle back to baseline rather than asserting on one immediate sample. A real leaked waiter goroutine would still fail this well within the 2s poll deadline, since it'd persist for the stuck operation's full duration. Ran 10x with `-race`, all pass.

wolf31o2 (dismissed):

wcatz:
Merged main into this branch to resolve a conflict: #3725 landed on main with an independent fix for the same `GetAccountsByCredential` OR-chain bottleneck this PR fixes, using a grouped-by-`credential_tag` `staking_key IN (...)` query instead of this branch's UNION-ALL derived-table join.

Resolution: kept main's `GetAccountsByCredential` + `accountsByCredentialChunkQuery` in full (it has its own DB-level tests and doesn't carry the duplicate-row risk a join has), dropped this branch's competing implementation and `dedupeStakeCredentialRefs` (which is no longer needed for this function under the grouped-IN approach, along with its now-unused `prefixedAccountColumns` helper and the two tests that only exercised the removed code), and updated `DATABASE.md`'s `SaveRewardAccountOutputs` section to stand on its own instead of cross-referencing `GetAccountsByCredential`'s now-different approach.

This branch's other changes — `DatabaseWorkerPool.Shutdown`'s fix, `SaveRewardAccountOutputs` ID resolution's own separate UNION-ALL fix, and the `node.go` cleanup comment — are unrelated to #3725 and merged cleanly.

Verified post-merge: `go build ./...`, `go test ./database/... ./ledger/... .`, `golangci-lint run ./...` (0 issues), `make docs-parity`, `gofmt`/`golines` all clean.

wolf31o2 (approved):

PR #3807, opened 2026-09-02
fix(ledger): engage replay recovery only for a real provenance gap

### LINKED ISSUE #3805: ledger: a deterministic script-data-hash rejection is treated as local state corruption, looping the pipeline forever instead of rejecting the branch
## Summary

A deterministic phase-1 validation failure — a script data hash mismatch — is
classified by replay recovery as inconsistent local ledger state. Recovery
rewinds, replays to the same tip, fails on the same transaction, and rewinds
again. The node reports the loop as unrecoverable every 20 attempts and keeps
going anyway, never converging and never stopping.

The underlying mismatch is blinklabs-io/gouroboros#2188. This issue is about
what the node does with it: a bad-block rejection is being treated as a
corrupted-state repair, and the repair can never succeed because the state was
never corrupt.

## Observed

Preview, from genesis, no Mithril, `storageMode: api`. The applied tip has sat
at slot 1796008 for over 45 minutes:

```
level=WARN  msg="TX 557b728b5a8aaed825d07ac2f76de31e3fb3f3069e19d32294215d820f375821
  failed validation: script data hash mismatch: declared 15dd0a3a…, computed 5d1149cf…"

level=WARN  msg="detected inconsistent local ledger state during replay, rewinding
  primary chain and metadata state" component=ledger
  recovery_strategy=security-param-fallback
  failing_block_slot=1796036
  missing_input=3b063fb1ab831a4d03e628a30a072394fc9bb3d0278b6b7f355e623e5baa462e#1
  producer_block_slot=1787909 rollback_slot=1787815 holding=false

level=WARN  msg="replay recovery not converging after unresolved-producer failures,
  holding at applied ledger tip" ledger_tip_slot=1796008

level=ERROR msg="ledger pipeline stuck: repeated restarts are not advancing the tip,
  so the failure is deterministic and will not clear on its own; the node is no
  longer following the chain" consecutive_no_progress=190 tip_slot=1796008 backoff=30s
```

The cycle also produces a steady stream of

```
level=ERROR msg="failed to handle block header"
  error="failed resolving fork after header mismatch: iterate local candidate
         after intersection 1841685: block not found"
```

as chain selection is offered candidates the repeatedly-rewound local chain can
no longer iterate.

## The state was never inconsistent

`missing_input` names `3b063fb1…#1`, but that UTxO is present and unspent in the
node's own store:

```sql
SELECT hex(tx_id), output_idx, amount, added_slot, deleted_slot FROM utxo
WHERE hex(tx_id) = upper('3b063fb1ab831a4d03e628a30a072394fc9bb3d0278b6b7f355e623e5baa462e');

3B063FB1…|0|2000000   |1790127|0
3B063FB1…|1|11688720  |1790127|0
3B063FB1…|2|9983702326|1790127|0
```

The producer is at slot 1790127, not the 1787909 the recovery log reports, and
the row is live. Nothing about the ledger state needs repairing — the
transaction fails a hash check, and it fails identically on every replay.

## Why the loop cannot end

`consecutive_no_progress` reaching 190 with the tip unchanged is the node
correctly detecting that the failure is deterministic, and the message says so
in as many words: "will not clear on its own; the node is no longer following
the chain". It then backs off 30 seconds and restarts the same pipeline.

Every rewind also discards local blocks the subsequent fork resolution then
cannot find, which is where the `block not found` errors come from — the
recovery is degrading the chain store on each pass while fixing nothing.

## Relationship to other issues

- blinklabs-io/gouroboros#2188 is the root cause of this particular block being
  rejected. Fixing it unblocks this slot, but not the class: any deterministic
  phase-1 rejection lands in the same loop.
- #3766 is the same recovery machinery leaving the UTxO set inconsistent on the
  at-tip route on a Mithril node, halting at the trust boundary. This is the
  replay route on a from-genesis node, where there is no trust boundary to halt
  against, so it loops instead of halting.
- #1906 closed a script-data-hash infinite loop under Conway UTXO rule 10. This
  is the same symptom via `security-param-fallback` replay recovery in Babbage,
  so whatever bound that fix added is not covering this path.

## Suggested direction

Separate the two diagnoses. A validation failure whose inputs all resolve on the
applied chain is a rejected block, not a corrupted local state: the branch
should be rejected and another candidate requested, which is what the
deterministic-rejection path already does for header validation. Recovery should
only engage when an input genuinely has no producer locally.

Failing that, `consecutive_no_progress` crossing its own "will not clear on its
own" threshold should stop the pipeline rather than back off and retry
indefinitely, so an operator sees a halted node instead of one that looks alive
and is not following the chain.

## Reproduction

```
  --mithril-enabled=false --socket-path /tmp/dgk/n.socket
```

from genesis on `main` plus #3794 (without it the replay wedges earlier, at slot
1730450). Reaches this loop at slot 1796008 in roughly 25 minutes.

chrisguiney:
Closes blinklabs-io/dingo#3805

`txValidationError` carries every referenced input of the failing transaction
regardless of what the transaction failed on, and `resolveReplayRecoveryProducer`
returned a nil producer for two unrelated situations: the input is present in
the UTxO set so there is nothing to find, and the input is missing with no
producer locatable. `findReplayRecoveryCandidate` folded both into
`unresolvedInputs`.

So a failure with nothing missing made every input of the transaction look
unresolved. Recovery rewound, the deterministic failure reproduced on replay,
and the pipeline looped — never converging, and never halting either.

## Observed

A preview genesis replay, wedged for 45 minutes on a script data hash mismatch:

```
level=WARN  msg="TX 557b728b… failed validation: script data hash mismatch:
  declared 15dd0a3a…, computed 5d1149cf…"
level=WARN  msg="detected inconsistent local ledger state during replay,
  rewinding primary chain and metadata state"
  recovery_strategy=security-param-fallback failing_block_slot=1796036
  missing_input=3b063fb1…#1 rollback_slot=1787815
level=ERROR msg="ledger pipeline stuck: repeated restarts are not advancing the
  tip, so the failure is deterministic and will not clear on its own; the node is
  no longer following the chain" consecutive_no_progress=230 tip_slot=1796008
```

The input the recovery names as missing was present and unspent in the node's own
store the entire time:

```sql
SELECT hex(tx_id), output_idx, amount, added_slot, deleted_slot FROM utxo
WHERE hex(tx_id) = upper('3b063fb1ab831a4d03e628a30a072394fc9bb3d0278b6b7f355e623e5baa462e');

3B063FB1…|1|11688720|1790127|0
```

## The fix

`resolveReplayRecoveryProducer` now reports presence separately from a nil
producer, and the candidate search skips a present input instead of folding it
into `unresolvedInputs`. A transaction whose inputs all resolve yields no
candidate, so `tryRecoverFromTxValidationError` returns false and the caller
rejects the branch — which is what a rejected block should get.

Presence is asked of a new `Database.UtxoExists`, which does not materialize the
output's CBOR. `UtxoByRef` resolves the bytes from the blob store and, on a miss,
reconstructs them by decoding the producing block; a caller that only needs to
know whether the output is still there pays for all of that and, worse, turns a
CBOR that cannot be recovered into a hard error about a UTxO that demonstrably
exists.

The genuine provenance gap this machinery exists for — an input with no producer
on the applied chain — is unchanged, and the same replay reaches it later at
slot 3363516, where recovery engages as intended.

## Scope

This is the misclassification only. The root cause of *this* particular block
being rejected is blinklabs-io/gouroboros#2188, fixed in
blinklabs-io/gouroboros#2189; that unblocks slot 1796036, while this change is
what keeps the next deterministic phase-1 rejection from looping the pipeline
instead of rejecting the branch.

## Tests

- `TestTryRecoverFromTxValidationErrorIgnoresFailureWithResolvableInputs` — a
  validation failure whose single input is present produces no candidate, no
  recovery, and no rewind. It does not pass on `main`.
- `TestResolveReplayRecoveryProducerReportsPresentInput` — pins the distinction
  directly: a live UTxO reports present with a nil producer, an absent reference
  reports not-present with a nil producer.
- Every existing replay-recovery test passes unchanged, including the
  unresolved-producer, fallback, and non-converging-hold cases.
- `go build ./...`, `go vet`, `go test ./ledger/... ./database/...` and
  `go test ./internal/test/conformance/` all pass.

---
## Summary by cubic
Fixes replay recovery engaging on transaction failures unrelated to input provenance, which caused endless rewind/replay loops (e.g., a script data hash mismatch drove 230 rewinds on preview). Now recovery only triggers for a genuine provenance gap: an input missing from the applied chain with no locatable producer.

- `resolveReplayRecoveryProducer` now reports a present input separately from a nil producer, and the candidate search skips present inputs instead of folding them into `unresolvedInputs`.
- Adds `Database.UtxoExists` to check presence without materializing the output's CBOR, avoiding hard errors on outputs that exist but have unrecoverable CBOR.
- Adds regression tests for the misclassification and the new presence check.

<sup>Written for commit 74a6cacdcb20df8df75c4efe2392dee9f90a5202. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3807?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved replay recovery to distinguish existing UTxO inputs from genuinely missing inputs.
  * Prevented unnecessary chain rewinds when transaction inputs are already present.
  * Avoided repeated recovery loops caused by deterministic transaction failures.
  * Rejected transactions with no recoverable missing inputs without attempting recovery.

* **Tests**
  * Added regression coverage for transactions with resolvable inputs and missing producer detection.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -3653,6 +3653,22 @@ queued primary chain, so late bodies from an abandoned fetch cannot seed its
 producer window. Its diagnostic database probes are also capped per block
 while the blockfetch pipeline lock is held; this keeps the audit bounded and
 non-blocking for pathological transaction counts.
+Replay recovery engages only for the failure it can repair. `txValidationError`
+carries every referenced input of the failing transaction regardless of what
+the transaction failed on, so a failure with nothing missing — a script data
+hash mismatch, say — arrived at the candidate search with a full input list.
+`resolveReplayRecoveryProducer` returns a nil producer both for an input that
+is present in the UTxO set (nothing to find) and for one that is missing with
+no producer locatable, and folding the two together made every input of such a
+transaction look unresolved: recovery rewound, the deterministic failure
+reproduced on replay, and the pipeline looped without converging and without
+halting (issue #3805). The two are now distinguished — a present input is
+reported as present and skipped — so a transaction whose inputs all resolve
+yields no candidate and the branch is rejected instead. Presence is asked of
+`Database.UtxoExists`, which does not materialize the output's CBOR: `UtxoByRef`
+reconstructs it by decoding the producing block on a blob miss, which turned a
+UTxO that demonstrably exists into a hard error out of recovery.
+
 The unresolved-producer fallback also tracks the applied ledger high-water
 mark across attempts. Different candidate continuations can move the failing
 block forward slightly while rebuilding to the same applied tip, so failure
--- a/database/utxo.go
+++ b/database/utxo.go
@@ -494,6 +494,32 @@ func (d *Database) UtxoByRef(
 	return utxo, nil
 }
 
+// UtxoExists reports whether a live UTxO is recorded for the reference, without
+// materializing its CBOR.
+//
+// UtxoByRef resolves the output's bytes from the blob store and, on a miss,
+// reconstructs them by decoding the producing block. A caller that only needs
+// to know whether the output is still there pays for all of that, and — worse —
+// turns a CBOR that cannot be recovered into a hard error about a UTxO that
+// demonstrably exists. Replay recovery asks exactly that question of every
+// referenced input of a failing transaction (see
+// LedgerState.findReplayRecoveryCandidate), so it uses this instead.
+func (d *Database) UtxoExists(
+	txId []byte,
+	outputIdx uint32,
+	txn *Txn,
+) (bool, error) {
+	if txn == nil {
+		txn = d.Transaction(false)
+		defer txn.Release()
+	}
+	utxo, err := d.utxoStore().GetUtxo(txId, outputIdx, txn.Metadata())
+	if err != nil {
+		return false, err
+	}
+	return utxo != nil, nil
+}
+
 // UtxosByRefs returns the live UTxOs matching the given references in a
 // single batch. Refs with no matching live UTxO are simply absent from the
 // result.
--- a/ledger/replay_recovery.go
+++ b/ledger/replay_recovery.go
@@ -1296,12 +1296,16 @@ func (ls *LedgerState) findReplayRecoveryCandidate(
 			continue
 		}
 		seenInputs[inputKey] = struct{}{}
-		resolved, err := ls.resolveReplayRecoveryProducer(
+		resolved, present, err := ls.resolveReplayRecoveryProducer(
 			pending,
 			chainIndex,
 		)
 		if err != nil {
 			return nil, err
+		}
+		if present {
+			// Nothing to repair for this input; it is still in the UTxO set.
+			continue
 		}
 		if resolved == nil {
 			unresolvedInputs = append(unresolvedInputs, pending.Input)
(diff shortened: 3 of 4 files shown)


wolf31o2 (approved):
Reviewed the current head; no blocking correctness issues found.

Issue #397, opened 2025-02-02
Validate requested block range in BlockFetch server


agaffney:
At `eabf653d27e4f4d21d428d5fb1a6bfdb5925e08d`, main already rejected ranges whose start slot exceeds the end slot, limited the slot delta, and resolved the start point. The original missing range-order check is therefore superseded.

The remaining requirement is to validate the requested end point, including its hash and membership in the served chain, rather than treating its slot alone as the requested boundary. Preserve valid single-point and multi-block ranges and the existing rejected-range response and recovery behavior.

PR #3840 and PR #4076 propose end-point membership validation. Both remain in flight; their presence does not establish that current main satisfies this requirement.

wolf31o2:
Status update: [#1346](https://github.com/blinklabs-io/dingo/pull/1346) added start-point lookup and rejects ranges whose start slot is after the end slot; [#1397](https://github.com/blinklabs-io/dingo/pull/1397) added the maximum slot-range bound.

The remaining half is still open. The server resolves only the start point before beginning the batch, and the send loop stops when it reaches the end slot without requiring the exact end hash. End-point existence and exact point identity therefore still need validation. Keeping this issue open with that narrowed scope.

PR #3840, opened 2026-09-03
fix(blockfetch): discard superseded batches

### LINKED ISSUE #3771: chainsync: preprod blockfetch continuation can remain stuck after rollback
## Scope

After rollback, bodies from a superseded BlockFetch batch can reach chain
insertion and clear the replacement header queue. Counting received rather
than applied bodies as progress can also prevent bounded range recovery.

The observed symptom is a running node with a growing tip gap, repeated
`does not fit on current chain tip` messages, and ineffective pipeline restarts.
Restarting a stale ledger iterator alone does not invalidate the old batch.

Proposed fix: #3840 binds batches to rollback generations and tracks applied
bodies as progress. Range endpoint validation is handled by merged #4076;
it is not a separate unresolved requirement of this issue.

## Acceptance criteria

- Superseded bodies cannot erase replacement headers after rollback.
- An all-discarded batch reaches bounded recovery instead of resetting progress.
- Same-peer continuation and fork-switch continuation retain fitting headers.
- Regression coverage exercises rollback, body delivery and range recovery.

wolf31o2:
Bind each BlockFetch batch to the chain rollback generation so abandoned
bodies cannot erase replacement headers. Count applied bodies, not received
bodies, as range progress so discarded batches reach bounded resynchronization.
Serialize rollback publication with batch insertion using the existing lock order.

Remove the duplicate endpoint guard and branch-added `Chain.HoldsPoint` API.
Validate requested endpoints before streaming. Rollback or iterator exhaustion
ends the batch with `BatchDone` without closing the transport; iterator errors
still close the connection. Regression coverage checks wrong-slot endpoints,
actual peer removal after repeated invalid ranges, and multi-block responses.

Fixes #3771.


ledger/chainsync.go
```diff
@@ -6450,23 +6477,33 @@ func (ls *LedgerState) handleEventBlockfetchBatchDone(
...
 	// startQueuedBlockfetchLocked). Both feed the same streak.
-	if receivedBlockCount == 0 && remainingHeaders > 0 {
+	//
+	// The count that matters is blocks *applied*, not blocks received. A batch
+	// whose blocks all arrive and are then rejected -- as not fitting the
+	// current tip, or as not matching the queued header -- leaves the queue
+	// exactly where it was, so gating this on delivery alone let the same
+	// range be re-requested without bound while every reply was discarded
+	// (issue #3771). Applied is what makes the streak observable and bounded.
+	if appliedBlockCount == 0 && remainingHeaders > 0 {
```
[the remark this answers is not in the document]
wolf31o2: Valid. `handleEventBlockfetchBlock` called `noteBlockfetchRangeProgress` on receipt, so a body delivered for the tracked range start and then discarded reset the streak every round. Moved the call to the applied branch of `flushPendingBlockfetchBlocks`, and updated the helper doc to require application rather than receipt. `TestDiscardedBodyDoesNotCountAsRangeProgress` covers it: with the call back on receipt the record is cleared and the test FAILs (`go vet ./ledger/` 0).

ledger/state.go
```diff
@@ -3244,6 +3258,33 @@ func (ls *LedgerState) drainBlockPipelineBeforeRollback(
...
+// refilled makes the first of them clear those replacement headers (issue
+// #3771). Bumping the generation here lets flushPendingBlockfetchBlocks drop
+// them instead.
+//
+// A rollback that resolves to a queued header leaves the tip (and the blocks
+// already requested for it) alone, so the generation only moves when the tip
+// itself did.
+func (ls *LedgerState) rollbackChain(point ocommon.Point) error {
+	before := ls.chain.Tip().Point
+	if err := ls.chain.Rollback(point); err != nil {
```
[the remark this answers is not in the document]
wolf31o2: Valid. `rollbackChain` now bumps the generation before `Chain.Rollback`, so a flush that lands in that window sees the new generation and discards a batch that is about to be superseded, rather than seeing the old one after the chain has moved.

ledger/state.go
```diff
@@ -3244,6 +3258,33 @@ func (ls *LedgerState) drainBlockPipelineBeforeRollback(
...
+		return err
+	}
+	after := ls.chain.Tip().Point
+	if after.Slot != before.Slot || !bytes.Equal(after.Hash, before.Hash) {
```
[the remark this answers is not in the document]
wolf31o2: Valid. The tip comparison is gone: the generation is bumped for every attempted rollback, before the chain changes, so the queued-header branch of `Chain.rollbackLocked` (which truncates the queue without moving `Tip().Point`) is covered too. Cost of the conservative bump is at most one re-fetch of a partially buffered batch, and a rollback that then fails leaves the caller requesting a resync anyway.

ledger/chainsync.go
```diff
@@ -4323,6 +4330,26 @@ func (ls *LedgerState) flushPendingBlockfetchBlocks() error {
 	if len(ls.pendingBlockfetchEvents) == 0 {
 		return nil
 	}
+	if ls.blockfetchBatchChainGeneration != ls.chainRollbackGeneration.Load() {
```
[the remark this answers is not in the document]
wolf31o2: Valid. Addressed on three levels rather than one atomic boundary at the flush:

1. `rollbackChain` publishes the new generation before it touches the chain, so a mismatch caused by a replacement queue is always observed by the flush as a generation that has already moved.
2. The flush re-reads the generation before every insertion instead of once at entry, so a rollback landing between two blocks of the same batch stops the rest of the batch.
3. The `BlockNotMatchHeaderError` branch — the one that does the damage — now clears the queue only when the batch is still current, so the residual gap between the check and `AddBlockWithPoint` cannot wipe headers fork resolution just queued.

`rollbackChain` additionally takes `chainsyncBlockfetchMutex`, so the rollback and the flush are serialized outright. Audited for lock inversion: no path holds `chainsyncBlockfetchMutex` across a call reaching `rollbackChain` (`tryResolveFork` and `handleEventChainsyncBlockHeaderWithPending` both take it only after the rollback returns), and the order stays chainsyncMutex -> chainsyncBlockfetchMutex. `go test -race -timeout 50m ./ledger/` passes.

ledger/chainsync.go
```diff
@@ -4319,23 +4326,76 @@ func (ls *LedgerState) startQueuedBlockfetchFromEventLocked(
...
 	pending := ls.pendingBlockfetchEvents
 	ls.pendingBlockfetchEvents = ls.pendingBlockfetchEvents[:0]
 	// Commit each block before exposing it on the primary chain. The chain tip
 	// is used immediately by fork detection, so batching blob writes behind an
 	// already-advanced in-memory tip can strand the node on a fork when ancestor
 	// lookups hit uncommitted state.
-	for _, pendingEvent := range pending {
+	for i, pendingEvent := range pending {
+		// The rollback paths run under chainsyncMutex, not the mutex this
+		// flush holds, so a rollback can land between two blocks of the same
```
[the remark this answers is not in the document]
wolf31o2: Valid, the comment was stale. `rollbackChain` holds `chainsyncBlockfetchMutex` for the rollback, so it does serialize with the flush; the comments now say that, and say why the generation checks stay: publishing the generation before the chain changes is what keeps the discard correct for a reader that does not hold that mutex, and the not-matching-header guard is what stops a replacement queue being cleared. The doc on `rollbackChain` and ARCHITECTURE.md now also record the lock order (chainsyncMutex -> chainsyncBlockfetchMutex) and that callers must not already hold the blockfetch mutex.

wolf31o2:
Applied review follow-up in c0499deb:

- Documented handleEventChainsyncRollback in the rollback lock-order audit.
- Corrected ARCHITECTURE.md to state that range-failure records clear only after successful block application; rejected bodies retain the record.

Validated with the focused ledger race suite and internal conformance suite; both pass.

wolf31o2:
Follow-up pushed in f79521b0: corrected ledger/state.go so chainRollbackGeneration is documented as counting attempted rollbacks, including queued-header rollbacks and failed rollbacks.

Validation: focused ledger race suite passed (201.819s); git diff --check passed.

ledger/state.go
```diff
@@ -3244,6 +3258,44 @@ func (ls *LedgerState) drainBlockPipelineBeforeRollback(
...
+// insertions of the same flush. The generation is still published before the
+// chain changes rather than after: that ordering is what makes any chain
+// change a rollback caused observable as a moved generation to a reader that
+// does not hold the mutex, so the discard stays correct on its own terms.
+//
+// Callers must not already hold chainsyncBlockfetchMutex. Every rollback entry
+// point takes chainsyncMutex first and leaves the blockfetch mutex to the
+// restart that follows the rollback (tryResolveFork,
+// handleEventChainsyncBlockHeaderWithPending, RecoverAfterLocalRollback), so
+// the lock order stays chainsyncMutex -> chainsyncBlockfetchMutex.
```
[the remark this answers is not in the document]
wolf31o2: No longer applies at the current head. The parenthetical list this finding named lived on the `rollbackChain` helper, which is gone: the generation bump is now inline in `rollbackChainAndStateDeferred` and in the windowed replay-recovery rollback. What remains is the general statement in ARCHITECTURE.md, which enumerates nothing and is accurate for `handleEventChainsyncRollback` too — it reaches `rollbackChainAndState` under `chainsyncMutex` held by `handleEventChainsync`, and holds no blockfetch lock of its own.

ledger/state.go
```diff
@@ -943,6 +943,7 @@ type LedgerState struct {
 	firstBlockReceived            bool                // true after latency sample recorded for this batch
 	shadowBlockReceivedHashes     map[string]struct{} // blocks delivered this batch (dedup shadow vs primary)
 	batchBlocksReceived           int                 // total blocks received in current blockfetch batch (including mid-batch flushes)
+	batchBlocksApplied            int                 // blocks applied from current blockfetch batch
```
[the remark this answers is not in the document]
wolf31o2: Addressed at the current head. `batchBlocksApplied` is reset alongside `batchBlocksReceived` at batch start (chainsync.go:4347), incremented on each block that actually extends the chain (4674), and read by production logic in `handleEventBlockfetchBatchDone` (6805), where it gates the unobtained-range recovery. That gate is the point of the field: a batch whose blocks all arrive and are then discarded or rejected leaves the header queue where it was, so gating on delivery let the same range be re-requested without bound. The read was previously assigned to a variable still called `receivedBlockCount`, which hid exactly that distinction; it is now `appliedBlockCount`.

ledger/chainsync.go
```diff
@@ -3997,7 +3997,6 @@ func (ls *LedgerState) handleEventBlockfetchBlockDeferred(
 	ls.batchBlocksReceived++
```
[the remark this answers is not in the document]
wolf31o2: Addressed at the current head. The doc now reads "flushPendingBlockfetchBlocksDeferred flushes pending blockfetch events and queues each committed block's chain.update onto pubs..." and no longer refers to the deleted wrapper.

ledger/replay_recovery.go
```diff
@@ -782,6 +782,12 @@ func (ls *LedgerState) rollbackPrimaryChainInSecurityParamWindows(
 	if securityParam <= 0 {
 		return chain.ErrSecurityParamNotConfigured
 	}
+	// Serialize every windowed rollback with blockfetch insertion. A queued
+	// batch must not pass its generation check between validation and rollback,
+	// or replacement headers can be followed by bodies from the superseded
+	// chain.
+	ls.chainsyncBlockfetchMutex.Lock()
```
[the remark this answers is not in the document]
wolf31o2: Addressed at the current head, in the way this finding asks for. Each window uses `ls.chain.RollbackDeferred(...)` (replay_recovery.go:847 and :861) and sets a `pendingChainUpdates` flag; `ls.chain.PublishPendingChainUpdates()` runs from a deferred closure registered before both Lock/defer-Unlock pairs, so LIFO ordering publishes it after `transactionEventMutex` and `chainsyncBlockfetchMutex` are both released. No `chain.update` is delivered while either mutex is held.

ledger/replay_recovery.go
```diff
@@ -786,6 +786,19 @@ func (ls *LedgerState) rollbackPrimaryChainInSecurityParamWindows(
 	// with respect to a block-apply commit's AfterCommit Apply publication.
 	ls.transactionEventMutex.Lock()
 	defer ls.transactionEventMutex.Unlock()
+	rollbackWindow := func(rollbackPoint ocommon.Point) error {
+		// Serialize the generation publication and chain mutation with
+		// blockfetch insertion. Deferred chain events are published only after
+		// releasing the mutex because delivery may backpressure.
+		ls.chainsyncBlockfetchMutex.Lock()
```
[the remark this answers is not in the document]
wolf31o2: Not an inversion at the current head: both paths take the same order, blockfetch then transaction-event. `rollbackPrimaryChainInSecurityParamWindows` locks `chainsyncBlockfetchMutex` (replay_recovery.go:796) then `transactionEventMutex` (:798); `rollbackChainAndStateDeferred` locks `chainsyncBlockfetchMutex` (state.go:3441) then `transactionEventMutex` (:3444). `TestReplayRecoveryRollbackLockOrder` pins that order for the replay path against the AST, and fails against the opposite order.

ledger/replay_recovery_lock_order_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestReplayRecoveryRollbackLockOrder protects the lock-order contract shared
+// with rollbackChainAndStateDeferred. The blockfetch lock must be acquired
+// before transactionEventMutex. The old implementation acquired those locks
+// in the opposite order, so this test fails against the unfixed code instead
+// of merely checking that both mutexes appear somewhere in the function.
+func TestReplayRecoveryRollbackLockOrder(t *testing.T) {
```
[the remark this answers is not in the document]
wolf31o2: Addressed at the current head. The test walks to the `FuncDecl` named `rollbackPrimaryChainInSecurityParamWindows`, requires it to be present, and inspects only that function's body, so an unrelated function elsewhere in replay_recovery.go cannot change the collected sequence.

arepala-uml:
Reviewed. Core fix (generation-stamped blockfetch batches, mutex-serialized rollback/flush, applied-not-received progress gating, `Chain.HoldsPoint` range-end validation) is mechanically sound, and the new tests are real regressions — confirmed to fail red when their specific fix is reverted. Validation (build/vet/lint/fmt, full non-race suite, targeted `-race` on changed tests) is clean.

Two non-blocking notes, not blocking approval:
- `ledger/state.go` `rollbackChainAndStateDeferred`: the generation bump fires before the rollback attempt succeeds, so a failed rollback still invalidates in-flight batches on a no-op. Harmless, but worth a comment.
- `ouroboros/blockfetch.go`: the new end-not-found branch and the existing start-not-found branch share one closeReason log string, making the two causes indistinguishable in that log line.

Blocking: this conflicts with `main` at `ledger/replay_recovery.go`'s `rollbackPrimaryChainInSecurityParamWindows`, which #3889/f86d64f4 rewrote after this branch's last commit. That needs a hand-merge — reintroduce the `chainRollbackGeneration.Add(1)` stamping into main's new retry loop — not an auto-resolve, and `TestReplayRecoveryRollbackLockOrder` should be re-confirmed against the merged result.

Please rebase and resolve the conflict — once that's done this looks ready.

ouroboros/blockfetch.go
```diff
@@ -342,13 +378,38 @@ Loop:
...
+				// rollback message, so close the transport; the client can
+				// re-request against its updated chain. Mirrors the
+				// next.Rollback handling in chainsync without reporting
+				// successful completion.
+				o.closeBlockfetchConnection(
+					conn,
+					connectionID,
+					"rollback during blockfetch range",
+				)
+				return errors.New("blockfetch chain rolled back during range")
```
chrisguiney: **Blocker.** A chain rollback while serving a range now tears down the peer's entire connection and reports it as a server error.

The pre-existing behaviour here was `break Loop` -> `BatchDone`. This branch now calls `closeBlockfetchConnection` and returns an error, and that error propagates back to the goroutine in `blockfetchServerRequestRange`, which calls `reportBlockfetchServerAsyncError` (line 523). That helper logs `"blockfetch: async range server failed"` at **ERROR** and calls `conn.Close()` a second time. `conn` is the whole `ouroboros.Connection`, so chainsync, txsubmission, keepalive and peer-sharing to that downstream peer go down with the blockfetch range.

A rollback is a routine Cardano event, not a server fault. Every rollback that lands while dingo is serving a range to a downstream peer now drops that peer and fires an ERROR-level line; on a relay with many downstream peers holding in-flight ranges, a single rollback drops all of them at once.

Three things make this hard to accept as it stands:

1. It is not described anywhere. The PR body still says "Blockfetch has no rollback message, so end the batch cleanly" and "the streaming loop already stops at `end.Slot`", and names no test for it. `ARCHITECTURE.md`, which this PR edits, still says the server "streams `Block` messages until the requested end or local tip before `BatchDone`" -- now false in both the tip case and the rollback case.
2. #397 explicitly asks to "preserve valid single-point and multi-block ranges and the existing rejected-range response and recovery behavior". This changes the recovery behaviour.
3. It was reached by inverting two existing tests' assertions rather than by adding a test -- see the comment on `blockfetch_test.go`.

If closing the transport is intentional, the argument exists (`ouroboros-network`'s BlockFetch client raises a protocol failure on a short `MsgBatchDone`), but then please say so in the PR body and in `ARCHITECTURE.md`, and route this path away from `reportBlockfetchServerAsyncError` so a routine rollback is neither logged at ERROR nor closed twice. If it is not intentional, the rollback branch should keep the `break Loop`/`BatchDone` behaviour #397 asks be preserved, and the end-point strictness should rest on the `HoldsPoint(end)` check alone.

ouroboros/blockfetch.go
```diff
@@ -387,11 +448,19 @@ Loop:
 				return err
 			}
 			// Make sure we don't hang waiting for the next block if we've already hit the end
-			if next.Block.Slot == end.Slot {
+			if reachedEnd {
 				break Loop
 			}
 		}
 	}
+	if !reachedEnd {
```
chrisguiney: Same class as the rollback branch above, and the case `TestBlockfetchServerSendBatch_BatchDoneAtChainTip` used to cover: reaching the local chain tip before `end` now closes the connection instead of sending `BatchDone`.

With `HoldsPoint(end)` validating the end point up front, this should be close to unreachable through `blockfetchServerRequestRange`, which makes it a safety net. But a safety net that drops the peer's whole connection and logs at ERROR through `reportBlockfetchServerAsyncError` needs the same disclosure in the PR body and `ARCHITECTURE.md` as the rollback branch, and `ARCHITECTURE.md:551` still documents the old behaviour verbatim.

ouroboros/blockfetch_test.go
```diff
@@ -389,14 +389,14 @@ func TestBlockfetchServerSendBatch_BatchDoneAtChainTip(t *testing.T) {
 		conn,
 	)
 
-	assert.NoError(t, err)
+	assert.Error(t, err)
```
chrisguiney: `e86dfb22` rewrote two existing tests' assertions to the opposite of what they asserted, instead of adding a test for the new behaviour:

- here, `assert.NoError` -> `assert.Error`, `batchDoneCalls` 1 -> 0, `closeCalls` 0 -> 1, while the test is still named `..._BatchDoneAtChainTip`. The name now describes behaviour the test asserts does *not* happen.
- `TestBlockfetchServerSendBatch_RollbackEndsBatchWithoutServingBlock` -> renamed `..._RollbackClosesConnectionWithoutServingBlock`, with the same inversion and an inverted comment.

Both were the regression tests pinning the previous contract. Please rename this one to match what it now checks, and land the behaviour change on its own justification rather than as a silent flip of the assertions that pinned the old one.

ledger/chainsync.go
```diff
@@ -4043,7 +4043,6 @@ func (ls *LedgerState) handleEventBlockfetchBlockDeferred(
...
 	// If this block is the one a tracked range was failing to obtain, that
 	// range is fetchable after all and its failure record is stale.
```
chrisguiney, line 4130: Orphaned comment. `ls.noteBlockfetchRangeProgress(e.Point)` moved to the applied branch of `flushPendingBlockfetchBlocksDeferred`, but its comment stayed behind and now sits above the unrelated `blockfetchCommitBatchSize` check, where it describes nothing. Move it with the call, or drop it -- the applied branch already carries the reasoning.

ledger/state.go
```diff
@@ -3647,6 +3662,9 @@ func (ls *LedgerState) rollbackChainAndStateDeferred(
 	// new state before the matching Apply reaches the ordered lane.
 	var rollbackEvents []event.Event
 	err := func() error {
+		ls.chainsyncBlockfetchMutex.Lock()
+		defer ls.chainsyncBlockfetchMutex.Unlock()
+		ls.chainRollbackGeneration.Add(1)
```
chrisguiney, line 3753: Non-blocking. The generation is bumped before `validateAndEmitRollbackUndo`, so a rollback that is *refused* -- over-K, `ErrRollbackPointNotOnChain`, or `models.ErrBlockNotFound`, all of which `handleEventChainsyncRollback` treats as recoverable -- still supersedes the in-flight batch even though the chain never moved. `rollbackPrimaryChainInSecurityParamWindows` does the same at `replay_recovery.go:901`, and bumps again on every over-K `continue` retry.

An earlier thread put the cost of the conservative bump at "at most one re-fetch". That is no longer the whole cost after this PR: the discard leaves `batchBlocksApplied == 0`, and `handleEventBlockfetchBatchDone` now feeds that into `noteBlockfetchRangeUnavailable`, so each spurious discard also increments the same-range failure streak. `blockfetchMaxSameRangeFailures` of them drop the header range and force a fresh intersect.

Either bump only once the rollback is known to have been applied, or record the refusal so a discard it caused does not count against the range.

ledger/replay_recovery.go
```diff
@@ -777,24 +777,33 @@ func (ls *LedgerState) publishReplayRecoveryNonConvergingResync(
...
+	// ordered against block-apply AfterCommit callbacks, including between
+	// intermediate windows. Chain updates are published after both locks are
+	// released because delivery may backpressure or re-enter synchronization.
+	var pendingChainUpdates bool
+	defer func() {
+		if pendingChainUpdates {
+			ls.chain.PublishPendingChainUpdates()
+		}
+	}()
+	ls.chainsyncBlockfetchMutex.Lock()
```
chrisguiney, line 805: Non-blocking, but worth recording next to the lock. `chainsyncBlockfetchMutex` is now held for the entire windowed descent rather than per window -- potentially many K-sized steps, each doing `blocksAboveSlot` (a DB read plus decode of up to a security parameter's worth of blocks), `emitRollbackTransactionEvents`, and `RollbackDeferred`. Blockfetch is stalled throughout.

The `chain.update` half is handled correctly: the `PublishPendingChainUpdates` defer is registered before both `Lock` calls, so it runs after both unlocks. But `emitRollbackTransactionEvents` still publishes `ledger.tx` inside both locks, so a back-pressuring `ledger.tx` subscriber now stalls blockfetch as well as the transaction lane, which it did not before. `blocksAboveSlot` early-returns when nothing subscribes to `ledger.tx`/`ledger.error`, so a default node is unaffected and a node with a subscriber is not. The comment currently justifies only the ordering; the new duration and the subscriber constraint deserve a line too.

chrisguiney (changes requested):
Third-party review at head `269068cf`, in an isolated worktree with an isolated lint cache.

**Disposition: changes requested**, on one blocker in `ouroboros/blockfetch.go`.

The core of the change is sound and the new tests are real regressions. I reverted each fix in place and confirmed its test fails without it:

| revert | test | result |
|---|---|---|
| `HoldsPoint(end)` guard disabled | `TestBlockfetchServerRequestRange_{EndPointNotInChain,EndPointSlotHashMismatch,RepeatedBadEndPointReachesCloseThreshold}` | FAIL |
| `blockfetchBatchSuperseded()` -> `false` | `TestForkRestartKeepsReplacementHeadersWhenAbandonedBatchArrives` | FAIL |
| `batchBlocksApplied` -> `batchBlocksReceived` | `TestBatchDoneTreatsDiscardedBatchAsUnobtainedRange` | FAIL |
| `noteBlockfetchRangeProgress` back on receipt | `TestDiscardedBodyDoesNotCountAsRangeProgress` | FAIL |

Class audits that came back complete:

- The lock order `chainsyncBlockfetchMutex` -> `transactionEventMutex` appears in exactly two places (`state.go:3665`, `replay_recovery.go:805`), and there is no reverse ordering anywhere in `ledger/`; `submitBlockApplyDBTxn` takes only `transactionEventMutex`. No inversion.
- Neither caller of `rollbackChainAndStateDeferred` holds `chainsyncBlockfetchMutex` at the call site: `handleEventChainsyncRollback` (`chainsync.go:2302`) takes it only through helpers that release it, and `tryResolveFork`'s two lock spans (3772-3781, 3928-3937) are disjoint from the call at 3796. Checked against all 19 functions in `ledger/` that touch the mutex. No self-deadlock.
- `restartQueuedBlockfetchAfterForkLocked` flushes -- and therefore discards -- the abandoned batch at `chainsync.go:4098` before `startQueuedBlockfetchLocked` re-snapshots the generation at 4394, so the fork restart cannot launder stale bodies into the new generation.

Checks at this head:

| command | exit | result |
|---|---|---|
| `gofmt -l .` | 0 | `bark/blob_test.go`, `ledger/read_chain_pipeline_test.go` -- both pre-existing on the base branch, not this PR's |
| `go test ./ouroboros/ -run BlockfetchServerRequestRange` | 0 | ok |
| `go test ./ledger/` (the five new tests) | 0 | ok |
| `git merge origin/main` | 0 | merges clean against `76127f9c`; no conflict |

Resolved `github.com/blinklabs-io/gouroboros v0.202.10` in this worktree.

Not run, no infrastructure available here: DevNet, live preprod sync, Antithesis, Koios parity, conformance vectors. `nilaway` and `staticcheck` are not installed locally, so `make lint` fails on the missing tool rather than on the code; `golangci-lint run ./...` was run directly instead, and nilaway/modernize are skipped.

Bot coverage, read from the `commit_id` on each review record rather than from check colour: cubic's last review record is at `a30a06bb`, CodeRabbit's at `b22a1a89`, and `arepala-uml`'s review comment predates the last two commits (Sep 6 against Sep 7). No actor has a review record at `269068cf`; the green `cubic - AI code reviewer` check on this head reports "0 issues found across 1 file" with no corresponding review record. So `e86dfb22` and `269068cf` carry no bot or human review, and the blocker below is in `e86dfb22`.

`Fixes #3771` and `Fixes #397` both resolve and describe this change.

wolf31o2:
Review follow-up:

- Integrated current `main` in DCO-signed merge commit `95675c21`.
- The merged `rollbackPrimaryChainInSecurityParamWindows` retains `chainsyncBlockfetchMutex` across generation validation, undo publication, and `RollbackDeferred`; it increments `chainRollbackGeneration` before each rollback step.
- The architecture note already documents that failure records clear only after successful application and rejected bodies retain the record.

Please re-review the current head.

wolf31o2:
Requested a fresh review from chrisguiney at head 95675c21 after integrating current main. No code changes were made in this follow-up.

ledger/chainsync.go
```diff
@@ -4465,6 +4464,8 @@ func (ls *LedgerState) startQueuedBlockfetchLockedWithWaitSignal(
 	ls.shadowBlockfetchConnId = ouroboros.ConnectionId{}
 	ls.shadowBlockReceivedHashes = nil
 	ls.batchBlocksReceived = 0
+	ls.batchBlocksApplied = 0
+	ls.blockfetchBatchChainGeneration = ls.chainRollbackGeneration.Load()
```
chrisguiney, line 4479: **Blocker.** This republishes the batch generation while the mutex is still held (released on 4478), so the discard window opens *ahead* of the new `MsgRequestRange` instead of closing behind the old batch.

The guard identifies a batch by *when it was requested*, but `BlockfetchEvent` (`ledger/event.go:57`) carries no batch identity -- a body is matched to a batch only by `ConnectionId` plus a non-nil ready chan (4029-4037). Both restart sites pass the chainsync peer's own connection (3834, 4003), and in the #3771 preprod topology that peer *is* the active blockfetch peer. Blockfetch has no client-side cancel, so batch A keeps streaming.

Sequence: batch A in flight at generation G, rollback bumps to G+1, `restartQueuedBlockfetchAfterForkLocked` flushes the buffered bodies (correctly discarded on 4757), this line then sets G+1, and batch A's *remaining* bodies arrive, pass the ConnectionId filter, and `blockfetchBatchSuperseded()` reads `G+1 != G+1` = false. They reach insertion, are rejected as not matching the replacement queue, and `clearQueuedHeaders()` on 4848 runs. `HeaderCount()` is then 0, so 4452 returns without issuing a request -- nothing queued, nothing fetching.

Reproduced against this head:

```
batchGeneration=1 rollbackGeneration=1 superseded=false headersBefore=1 headersAfter=0
DEBUG discarding blockfetch blocks for superseded chain  discarded_blocks=1
WARN  ignoring blockfetch block: ... does not match first pending header hash
```

Attribution pinned by neutralizing only the `clearQueuedHeaders()` call in place: `headersAfter` flips 0 to 1.

Your four tests still pass under `-race`, so this is an uncovered case rather than a regression. `TestForkRestartKeepsReplacementHeadersWhenAbandonedBatchArrives` already uses one connection id throughout (147, 162, 242) -- the uncovered dimension is ordering, not peer identity. It delivers the abandoned body *before* the restart, where the restart's own flush on 4172 correctly discards it; it never delivers one afterwards. What protects the two-peer case is the `!fromPrimary && !fromShadow` filter on 4033, not this guard.

Smallest fix that closes it: stamp each buffered event with the batch id it was admitted under -- `blockfetchRequestGeneration` is already a monotonic id -- and discard on id mismatch rather than on rollback generation. I have the reproducing test if it is useful.

ledger/chainsync.go
```diff
@@ -4719,7 +4720,7 @@ func (ls *LedgerState) startQueuedBlockfetchFromEventLocked(
 	}()
 }
 
-// flushPendingBlockfetchBlocksDeferred is flushPendingBlockfetchBlocks that
+// flushPendingBlockfetchBlocksDeferred flushes pending blockfetch events and
```
chrisguiney, line 4734: The new function landed between this doc block and the function it documents, so the whole 11-line paragraph now attaches to `blockfetchBatchSuperseded`, and `flushPendingBlockfetchBlocksDeferred` (4746) has no doc comment at all. Moving `blockfetchBatchSuperseded` and its own comment above line 4723 restores both.

ARCHITECTURE.md
```diff
@@ -549,7 +549,14 @@ The BlockFetch server path mirrors the retrieval flow for downstream peers:
 when a peer requests a range, `ouroboros/blockfetch.go` validates the bounds,
 opens a chain iterator at the requested start point, sends `StartBatch`, then
 streams `Block` messages until the requested end or local tip before
```
chrisguiney, line 613: This sentence is now inaccurate, one line above the text the PR adds: reaching the local tip before `end` no longer sends `BatchDone`, it closes the connection (`ouroboros/blockfetch.go:456`).

Worth noting that path is production-reachable rather than defensive. `chain/chain.go:1441` gates the rollback sentinel on `refSlot > point.Slot`, so an iterator positioned *below* a rollback point gets no sentinel -- it silently exhausts at the shortened tip and falls into the `!reachedEnd` close. Every downstream peer being served a range that spans a fork loses its connection.

ARCHITECTURE.md
```diff
@@ -3362,13 +3388,14 @@ to back — timing-dependent enough that two identical DevNet runs differed by
...
-`clearQueuedHeaders` deliberately leaves it alone. It is discarded only when
-the tracked range itself is delivered (`noteBlockfetchRangeProgress` in
-`handleEventBlockfetchBlock`, matching on the block's point), so a peer that
-was briefly behind is never punished, and a miss against a different range
-starts its own count. After the bound fires the record restarts from zero, so
-a re-offered header must earn a fresh set of failures rather than being
-dropped on every later miss.
+`clearQueuedHeaders` deliberately leaves it alone. It is discarded only after
+the tracked range itself has been applied successfully in
+`flushPendingBlockfetchBlocks` after `AddBlockWithPoint` succeeds, matching on
```
chrisguiney, line 3649: Neither symbol exists: `flushPendingBlockfetchBlocks` was removed in favour of `flushPendingBlockfetchBlocksDeferred`, and the call is `AddBlockWithPointDeferred`. `grep -rn 'flushPendingBlockfetchBlocks\b' ` returns only this line.

ouroboros/blockfetch_test.go
```diff
@@ -433,18 +433,40 @@ func TestBlockfetchServerSendBatch_RollbackEndsBatchWithoutServingBlock(
...
+	conn := &stubBlockfetchConnection{errChan: make(chan error)}
+	start := ocommon.NewPoint(100, []byte{100})
+	end := ocommon.NewPoint(100, []byte{0xff})
+
+	err := o.blockfetchServerSendBatch(
+		testConnId().String(), start, end, iter, server, conn,
+	)
+
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "end hash mismatch")
```
chrisguiney, line 480: `assert.Error` does not stop the test, so when the fix is absent this dereferences a nil error. A fail-before revert of `ouroboros/blockfetch.go` panics here with SIGSEGV instead of reporting the failure. `require.Error` on 463 fixes it.

chrisguiney (changes requested):
Re-reviewing at `95675c21`. Requesting changes.

`95675c21` is a pure merge of `origin/main` into `269068cf` (two parents, no source edits in the flagged files), so the three blockers from my review at `269068cf` are unaddressed -- I have left those threads open rather than duplicating them here:

- `ouroboros/blockfetch.go:387` -- rollback mid-range closes the connection and returns an error, which `reportBlockfetchServerAsyncError` (530) then logs at ERROR and closes again.
- `ouroboros/blockfetch.go:456` -- reaching local tip before `end` also closes. See my comment on `ARCHITECTURE.md:551` for why this is reachable on any ordinary fork.
- `ouroboros/blockfetch_test.go:389` -- `TestBlockfetchServerSendBatch_BatchDoneAtChainTip` keeps its name with every assertion inverted.

New this pass, and the reason this is still a change request rather than a nit list: the #3771 fix does not close the same-peer case. Details and a reproduction are in the comment on `ledger/chainsync.go:4468`.

Two things that may make this easier rather than harder:

- #4076 also claims `Fixes #397` and touches the same three files. Splitting the #397 half out of this PR would drop all three standing blockers and leave this one focused on #3771, which is where the remaining work is.
- No review record on this PR carries `commit_id` `95675c21`, and the CodeRabbit check reports "Review skipped: incremental reviews are disabled" -- so this head has had no bot coverage despite the green checks.

Verified good along the way: the fail-before holds for the #397 half (reverting `blockfetch.go` to base puts `0x02 StartBatch` on the wire where `0x03 NoBlocks` is expected); no self-deadlock from the new `chainsyncBlockfetchMutex` acquisition, since both `rollbackChainAndStateDeferred` callers are on the chainsync event lane and the lock order matches `replay_recovery.go`; no data race on the plain-`uint64` generation read, as all four production flush callers hold the mutex; `TestReplayRecoveryRollbackLockOrder` is correctly scoped to the single `FuncDecl`; `ARCHITECTURE.md` is substantively updated and `DATABASE.md` correctly untouched.

Not run: DevNet, live preprod sync, Antithesis, conformance, full `-race ./ledger/`.

ouroboros/blockfetch.go
wolf31o2: Reverted in `35969133`. The rollback branch is `break Loop` again, so the batch ends with `BatchDone` and the transport stays up; `reportBlockfetchServerAsyncError` is no longer reachable from a rollback. End-point strictness rests on the `HoldsPoint(end)` check plus a guard for an iterator block past `end.Slot`.

ouroboros/blockfetch.go
wolf31o2: Also reverted in `35969133`. `ErrIteratorChainTip` and a nil `Next` both `break Loop`, and the `!reachedEnd` close is gone, so reaching the local tip before `end` sends `BatchDone`. The only close left on this path is an iterator block whose slot is past `end.Slot`, which means the requested end is not on the chain being served.

ouroboros/blockfetch_test.go
wolf31o2: `TestBlockfetchServerSendBatch_BatchDoneAtChainTip` and `..._RollbackEndsBatchWithoutServingBlock` are back to their original names and assertions. The end-point behaviour has its own coverage instead: `TestBlockfetchServerSendBatch_RejectsEndHashMismatch` and `ouroboros/blockfetch_range_end_test.go`.

ledger/chainsync.go
wolf31o2: The comment moved with the call. `noteBlockfetchRangeProgress` and its reasoning now sit together in the applied branch of `flushPendingBlockfetchBlocksDeferred`, after `AddBlockWithPointDeferred` succeeds.

ledger/state.go
wolf31o2: Fixed. `rollbackChainAndStateDeferred` records the prior generation and restores it when `validateAndEmitRollbackUndo` refuses the point, before `chain.RollbackDeferred` runs; `rollbackPrimaryChainInSecurityParamWindows` does the same, so an over-K retry no longer bumps once per attempt. Both the bump and the restore happen under `chainsyncBlockfetchMutex`, which every `blockfetchBatchSuperseded` reader also holds, so no reader sees the intermediate value. `TestRefusedRollbackKeepsInFlightBatch` covers it and fails when the restore is removed.

ledger/replay_recovery.go
wolf31o2: Recorded next to the lock. The comment now states that blockfetch is stalled for every K-sized step of the descent, that `emitRollbackTransactionEvents` publishes `ledger.tx` inside both locks so a back-pressuring `ledger.tx` subscriber stalls blockfetch as well as the transaction lane, and that `blocksAboveSlot` early-returns when nothing subscribes, so a default node pays neither cost.

ledger/chainsync.go
wolf31o2: Closed in `58753da0`..`de1f757c`, and not by the generation guard. `restartQueuedBlockfetchAfterForkLocked` records the abandoned request's connection in `blockfetchDiscardConnId`; `handleEventBlockfetchBlockDeferred` drops every event carrying it, and `handleEventBlockfetchBatchDone` consumes the marker, which is the ordering barrier for the old request. `completeBlockfetchRequestLocked` keeps the request in flight until its `BatchDone` has been handled rather than releasing it when `GetBlockRange` returns, so a same-connection restart waits for the old request's events to drain before the replacement is installed.

ledger/chainsync.go
wolf31o2: Moved. `blockfetchBatchSuperseded` now sits above `flushPendingBlockfetchBlocksDeferred` with its own comment, and the drain-deadlock paragraph is attached to the function it describes.

ARCHITECTURE.md
wolf31o2: The sentence is accurate again because the code changed back: reaching the local tip before `end` sends `BatchDone`. The case you traced through `chain/chain.go:1441`, where an iterator below a rollback point exhausts at the shortened tip, therefore ends the batch cleanly instead of closing the peer's connection.

ARCHITECTURE.md
wolf31o2: Corrected. The paragraph now names `flushPendingBlockfetchBlocksDeferred` and `AddBlockWithPointDeferred`, and states that the failure record clears only after the block has been applied.

ouroboros/blockfetch_test.go
wolf31o2: Changed to `require.Error`. Audited the same pattern across every test file this change touches or adds: the only other instance was `TestBlockfetchServerSendBatch_ClosesConnectionWhenSendDrainStalls` in this file, fixed alongside it, so the class is complete.

ledger/chainsync.go
```diff
@@ -4170,6 +4171,12 @@ func (ls *LedgerState) restartQueuedBlockfetchAfterForkLocked(
 	pending *pendingPublishes,
 ) error {
 	if ls.chainsyncBlockfetchReadyChan != nil {
+		// The old protocol request cannot be cancelled. Keep late events from
+		// being admitted after the replacement generation is installed when
+		// the restart uses the same connection.
+		if ls.blockfetchPrimaryRequestGeneration != 0 {
```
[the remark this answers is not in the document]
wolf31o2: Superseded at the current head. The marker is set unconditionally in `restartQueuedBlockfetchAfterForkLocked` rather than gated on `blockfetchPrimaryRequestGeneration`, and `completeBlockfetchRequestLocked` holds the request in flight until its `BatchDone` has been handled, so the replacement cannot start while the old request's events are still queued.

ledger/chainsync.go
```diff
@@ -7015,6 +7082,12 @@ func (ls *LedgerState) handleEventBlockfetchBatchDone(
 	if ls.chainsyncBlockfetchReadyChan == nil {
 		return nil
 	}
+	if connIdKey(ls.blockfetchDiscardConnId) != "" &&
+		sameConnectionId(e.ConnectionId, ls.blockfetchDiscardConnId) {
+		// BatchDone is the ordering barrier for the abandoned request.
+		ls.blockfetchDiscardConnId = ouroboros.ConnectionId{}
```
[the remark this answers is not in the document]
wolf31o2: Addressed at the current head. `handleEventBlockfetchBatchDone` consumes `blockfetchDiscardConnId` before the `chainsyncBlockfetchReadyChan == nil` check, so a late `BatchDone` clears the marker instead of returning past it.

ledger/state.go
```diff
@@ -943,6 +943,20 @@ type LedgerState struct {
...
+	// header queue that existed at request time; a rollback replaces both that
+	// queue and the continuation point, so every block still arriving for the
+	// older generation belongs to a chain the node has abandoned and must be
+	// discarded rather than applied (issue #3771). Guarded by
+	// chainsyncBlockfetchMutex, like the rest of the per-batch state.
+	blockfetchBatchChainGeneration uint64
+	// chainRollbackGeneration counts primary-chain rollbacks. It is bumped
+	// before the chain is changed, so a reader that has observed a rollback's
+	// effect on the chain always observes the new value. Written by the
+	// rollback paths and read by the blockfetch paths, so it is atomic.
```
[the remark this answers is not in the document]
wolf31o2: Applied in `b9d4cdfc`; `chainRollbackGeneration` is documented as counting attempted rollbacks. A refused attempt now also restores the prior value under the same mutex, so the counter no longer advances for a rollback the chain never applied.

ledger/replay_recovery.go
```diff
@@ -827,6 +827,8 @@ func (ls *LedgerState) rollbackPrimaryChainInSecurityParamWindows(
 	}
 	// Keep every Undo enqueue and its corresponding chain truncation atomic
 	// with respect to a block-apply commit's AfterCommit Apply publication.
+	ls.chainsyncBlockfetchMutex.Lock()
```
chrisguiney, line 833: Blocker, and the cause of the red pipeline. Holding chainsyncBlockfetchMutex here spans ls.chain.Rollback(next) at line 924, which publishes inline; EventBus.Publish applies backpressure and RecoverAfterLocalRollback takes chainsyncMutex then this mutex, so the cycle pending_publish.go describes is reopened. TestNoEventBusPublishWhileHoldingChainsyncMutex fails on Linux, Linux-race, Windows and macOS. Control: that test is ok on origin/main dce82ce6 and FAIL at this head and in the merged tree. Queue the rollback publication with pendingPublishes and flush it after the unlock.

ledger/replay_recovery_lock_order_test.go
```diff
@@ -0,0 +1,80 @@
...
+		switch mutex.Sel.Name {
+		case "chainsyncBlockfetchMutex", "transactionEventMutex":
+			lockOrder = append(lockOrder, mutex.Sel.Name)
+		}
+		return true
+	})
+
+	require.Equal(
+		t,
+		[]string{"chainsyncBlockfetchMutex", "transactionEventMutex"},
```
chrisguiney, line 76: This assertion requires rollbackPrimaryChainInSecurityParamWindows to take chainsyncBlockfetchMutex before transactionEventMutex, while TestNoEventBusPublishWhileHoldingChainsyncMutex in the same package forbids holding chainsyncBlockfetchMutex across the ls.chain.Rollback publication that function performs. As written the two cannot both pass, so a fix has to satisfy both: keep the order but move the publication out from under the lock, or narrow the locked span to the Undo enqueue and the truncation it pairs with.

ledger/chainsync.go
```diff
@@ -4770,7 +4781,6 @@ func (ls *LedgerState) startQueuedBlockfetchLockedWithWaitSignal(
 		return err
 	}
 	ls.chainsyncBlockfetchMutex.Lock()
```
chrisguiney, line 4899: Blocker. Dropping endBlockfetchRequestLocked from this success path leaves blockfetchRequestsInFlight[conn] set until a BatchDone is handled, and nothing else removes it: blockfetchRequestRangeCleanup, handleBlockfetchTimeoutLocked and handleConnectionClosedEvent all leave the map alone. A peer that accepts the range and then goes silent wedges that connection for good. Measured by calling startQueuedBlockfetchLockedWithWaitSignal twice: at this head the second call issues no request and blocks past blockfetchBusyTimeout; on origin/main it returns nil and issues it.

ledger/chainsync.go
```diff
@@ -4864,7 +4874,9 @@ func (ls *LedgerState) startQueuedBlockfetchLockedWithWaitSignal(
 					headerEnd,
 				)
 				ls.chainsyncBlockfetchMutex.Lock()
-				ls.endBlockfetchRequestLocked(shadowConn, shadowRequestDone)
+				if err != nil {
+					ls.endBlockfetchRequestLocked(shadowConn, shadowRequestDone)
```
chrisguiney, line 4994: Same class on the shadow path. On success the shadow request now stays in blockfetchRequestsInFlight until its own BatchDone is handled, so a shadow peer that never sends one leaves the entry and its channel behind and every later batch on that connection waits out blockfetchBusyTimeout. Those are the only two members of the class: endBlockfetchRequestLocked is reached from line 4751 and from here, and completeBlockfetchRequestLocked only from handleEventBlockfetchBatchDone.

ledger/chainsync.go
```diff
@@ -4164,6 +4164,10 @@ func (ls *LedgerState) handleEventBlockfetchBlockDeferred(
 	if ls.chainsyncBlockfetchReadyChan == nil {
 		return nil
 	}
+	if connIdKey(ls.blockfetchDiscardConnId) != "" &&
```
chrisguiney, line 4169: This is what closes the same-peer ordering case raised in the last review, and nothing pins it. grep blockfetchDiscardConnId over ledger/*_test.go returns no match, and deleting both this filter and the BatchDone consume at line 7450 in place leaves every ledger test green apart from the unrelated publish-under-lock guard. TestForkRestartKeepsReplacementHeadersWhenAbandonedBatchArrives still delivers the abandoned body before the restart; one that delivers it after would fail without this.

ARCHITECTURE.md
```diff
@@ -636,14 +636,17 @@ to do real work should hand off to their own goroutine, which
...
-The range sender is asynchronous so the mini-protocol callback can
+opens chain iterators at the requested start and end points to validate both
+endpoints against the serving chain, then sends `StartBatch`. Checking only
+the end slot let a peer name an end point the server does not hold and receive
+a slot-bounded prefix of the server's own chain in its place. An end point
+that does not resolve takes the same
+`NoBlocks` and stuck-peer accounting as the other invalid-range rejections.
+Once a batch starts, blocks are streamed through the exact requested end slot
+and hash; earlier blocks at the same slot, including Byron epoch-boundary
+blocks, do not complete the range. Iterator exhaustion, rollback, or passing the end slot
```
chrisguiney: Iterator exhaustion and rollback no longer close the connection: ouroboros/blockfetch.go breaks the loop on ErrIteratorChainTip, on a nil Next and on next.Rollback, and then sends BatchDone. This branch's own TestBlockfetchServerSendBatch_BatchDoneAtChainTip and _RollbackEndsBatchWithoutServingBlock assert batchDoneCalls 1 and closeCalls 0 for exactly those two cases. Only a block past end.Slot closes now, so this sentence should name that case alone.

ouroboros/blockfetch.go
```diff
@@ -378,10 +378,27 @@ Loop:
...
+				// surfaced a rollback sentinel with a zero-value Block.
+				// Serving it would stream a [0, null] block that a fetching
+				// peer decodes as a nil-header Byron EBB and crashes
+				// dereferencing it in SlotNumber(). Blockfetch has no
+				// rollback message, so end the batch cleanly; the client
+				// re-requests against its updated chain. Mirrors the
+				// next.Rollback handling in chainsync.
 				break Loop
 			}
 			if next.Block.Slot > end.Slot {
```
chrisguiney, line 391: The one close left still reports a routine outcome as a server fault: this error returns to the goroutine in blockfetchServerRequestRange, which calls reportBlockfetchServerAsyncError at line 504, logging at ERROR and calling conn.Close a second time on the whole ouroboros.Connection, so the peer's chainsync, txsubmission and keepalive go down with the range. Routing this case away from that helper would keep the transport up. The end-point check just below also switched its slot source to next.Block.Slot while keeping next.Point.Hash.

ouroboros/blockfetch_test.go
```diff
@@ -450,49 +450,15 @@ func TestBlockfetchServerSendBatch_RollbackEndsBatchWithoutServingBlock(
...
-		server,
-		conn,
-	)
-
-	assert.Error(t, err)
-	assert.ErrorContains(t, err, "ended before requested end point")
-	assert.Equal(t, 1, server.blockCalls)
-	assert.Equal(t, 0, server.batchDoneCalls)
-	assert.Equal(t, 1, conn.closeCalls)
+		// Blockfetch has no rollback message, so end the batch cleanly and let the
```
chrisguiney: gofmt -l . at this head lists ouroboros/blockfetch_test.go, which the base branch does not: this comment is indented as though it sat inside the assert.Equal call above it. The lint job does not cover formatting, so CI stays green on it.

chrisguiney (dismissed):
Re-review at 81bbd2e0. The prior blockers on ouroboros/blockfetch.go and its two renamed tests are cleared and verified at this head. The pipeline is red on all four go-test jobs from TestNoEventBusPublishWhileHoldingChainsyncMutex; that test passes on origin/main dce82ce6 and fails at this head and in the merged tree, so the failure is this branch's. Two further blockers, one uncovered fix and three smaller findings are inline.

ledger/replay_recovery.go
wolf31o2: Fixed in ca5b3a96e. The recovery path now uses RollbackDeferred, drains the chain publication into pendingPublishes, and flushes only after both mutex defers have unlocked. TestNoEventBusPublishWhileHoldingChainsyncMutex and TestReplayRecoveryRollbackLockOrder pass under -race.

ledger/replay_recovery_lock_order_test.go
wolf31o2: Fixed in ca5b3a96e. The lock order remains chainsyncBlockfetchMutex before transactionEventMutex, while the chain update publication is deferred until after both locks are released. Both lock-order guards pass under -race.

ledger/chainsync.go
wolf31o2: Fixed in ca5b3a96e. Timeout, disconnect, and winner cleanup now release the active request reservation; a connection-close event is also a terminal barrier when BatchDone never arrives. The retry and connection-close regression tests fail without these releases and pass under -race with them.

ledger/chainsync.go
wolf31o2: Fixed in ca5b3a96e. blockfetchRequestRangeCleanup now releases both active and shadow reservations, so a silent shadow cannot retain its map entry or waiter through the next batch. The timeout retry test now seeds and verifies both reservations.

ledger/chainsync.go
wolf31o2: Covered in ca5b3a96e by TestForkRestartDropsAbandonedBodyArrivingAfterRestart. It restarts on the same peer, delivers the abandoned body after restart, consumes the old BatchDone barrier, and verifies the replacement body is accepted. The test fails when the late-event filter is removed.

ARCHITECTURE.md
wolf31o2: Updated in ca5b3a96e. ARCHITECTURE.md now documents iterator exhaustion, rollback, and post-validation end-slot overshoot after a concurrent chain change as clean BatchDone outcomes that keep the peer connected.

ouroboros/blockfetch.go
wolf31o2: Fixed in ca5b3a96e. A post-validation overshoot now ends the batch with BatchDone instead of returning through reportBlockfetchServerAsyncError, so the whole Ouroboros connection remains up. The endpoint-contract test pins that result and fails against the previous close/error path.

ouroboros/blockfetch_test.go
wolf31o2: Fixed in ca5b3a96e. ouroboros/blockfetch_test.go is gofmt-clean; gofmt -l . now returns no files.

arepala-uml (approved):

PR #3865, opened 2026-09-04
fix(ledger): fail closed without hard-fork forecast shape

### LINKED ISSUE #3427: hardfork: reject summaries with an unbounded forecast horizon
`ledger/hardfork_summary.go` returns a summary with an unbounded final era when the current era is absent from the configured shape, the Cardano node config is nil, or shape construction fails. The last era has no `End` and a zero safe zone, so `SlotToEpoch` never reports `ErrPastHorizon` and far-future header validation can repeatedly advance the epoch cache.

Acceptance criteria:

- Return an explicit no-forecast error or bound the final era when the current shape entry is unavailable.
- Do not interpret a zero safe zone as an unlimited forecast horizon.
- Cover missing config, shape-construction failure, and an era outside the enabled table.

wolf31o2:
Fixes #3427.

- Fail closed when the current era cannot be resolved to the configured hard-fork shape.
- Preserve cache-backed epoch and slot conversion when forecasting is unavailable.
- Defer header verification when local forecast, consensus-mode, or genesis-overlay inputs cannot be resolved.
- Decline leader-schedule computation rather than guessing a consensus mode.
- Cover unavailable shapes, transitions, parameters, cached nonces, and the deterministic near-now slot-time fallback.


arepala-uml (changes requested):
Requesting changes for two reasons: a real merge conflict, and a correctness gap in the fix itself.

**1. Merge conflict.** GitHub reports this branch as CONFLICTING against main. Traced it to a single-hunk conflict in `ledger/hardfork_summary_test.go`: main added `t.Parallel()` to `TestHardForkSummary_MissingShelleyGenesis` (via #4131) after this branch was cut, while this PR deletes that same function and replaces it with a new table-driven `TestHardForkSummary_RejectsUnavailableShape` at the same location. Mechanical, but real — please rebase. Confirmed this is unrelated to #4094 (leader/election.go) — that PR is still open/unmerged, so there's no overlap there.

**2. Correctness gap — `ConsensusModeForEpoch` still fails open in one path.** `ledger/state.go:10147-10150`'s era-walk loop:
```go
nextID := eraID + 1
if _, ok := ls.eraById(nextID); !ok {
    break   // falls through to return consensusModeForEraID(eraID), nil — the pre-fork era's mode
}
```
When the hard-fork shape indicates a scheduled fork at-or-before the target epoch, but the separate `ls.activeEras` registry doesn't yet recognize the successor era ID, this silently returns the **prior era's** consensus mode for a future epoch that the shape itself says undergoes a fork — exactly the class of bug this PR's own doc comment says it eliminates. This is reachable: the PR's own `TestProtocolParamsForSlot_UnavailableTransition/missing_successor` test constructs exactly this scenario (`ls.activeEras = []eras.EraDesc{eras.ShelleyEraDesc}`), just for a different function. The **sibling loop in `ProtocolParamsForSlot`** in the same file (~line 9965) handles the identical situation correctly (`if !ok || nextEraPtr == nil { return nil }`) — `ConsensusModeForEpoch`'s loop is the odd one out. CodeRabbit already flagged this as a Major finding; it's still open at this head. No test currently exercises this exact case for `ConsensusModeForEpoch`.

Everything else in this PR checks out: every other caller of the changed forecast functions (forging builder/forger, header verification, leader election) correctly aborts on the new fail-closed error path, the happy path is preserved, and the broad ledger/forging/leader test suite plus a full conformance run all pass clean. Once the conflict is rebased and the `ConsensusModeForEpoch` gap is fixed (matching `ProtocolParamsForSlot`'s pattern), I'm happy to re-review.

ledger/state.go
```diff
@@ -6877,27 +6877,33 @@ func (ls *LedgerState) reconstructTransitionInfo() {
...
 	ls.cachedShape.CompareAndSwap(nil, &s)
-	return *ls.cachedShape.Load()
+	return *ls.cachedShape.Load(), nil
+}
+
+// eraShape returns the resolved hardfork.Shape, or an empty Shape when the
+// node configuration cannot supply one. Callers that use the shape as a
+// consensus forecast must call eraShapeWithError and fail closed instead.
+func (ls *LedgerState) eraShape() hardfork.Shape {
+	shape, _ := ls.eraShapeWithError()
```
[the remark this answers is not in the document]
wolf31o2: Addressed in `9df8b2de`. `ProtocolParamsForSlot`, `ConsensusModeForEpoch` and `hardForkSummaryAnchoredAt` all call `eraShapeWithError` and fail closed on an unresolvable shape. `eraShape` remains only for callers that legitimately treat an absent shape as no forecast.

ledger/hardfork_summary.go
```diff
@@ -118,10 +119,15 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
...
 
 	// Use the same configured safe-zone source as the NtC era-history query.
-	// Tests and early bootstrap callers may construct a LedgerState without a
-	// complete node configuration; in that case eraShape is unavailable and
-	// the legacy indefinite safe zone remains the only defensible answer.
-	shape := ls.eraShape()
+	// A missing shape or current-era entry cannot safely supply a forecast:
+	// the cache-derived current era has a zero safe zone and no end bound.
+	shape, err := ls.eraShapeWithError()
+	if err != nil {
```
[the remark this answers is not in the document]
wolf31o2: The cache-only conversions no longer need the shape. `SlotToEpoch` answers a covered slot from the epoch cache before building a summary, and `SlotToTime` walks the cache through `cachedSlotTime` when no summary can be built. `TestSlotCalc` passes at this head.

ledger/verify_header.go
```diff
@@ -414,8 +414,11 @@ func (ls *LedgerState) headerVerificationEpoch(
 ) (models.Epoch, error) {
 	// The epoch cache can be forecast forward for near-future headers, but it
 	// must never be advanced past the HFC safe zone or a known era boundary.
-	// Check the immutable summary first so ErrPastHorizon is surfaced before
-	// ensureEpochForSlot mutates any forecasted nonce state.
+	// Resolve known epochs without requiring forecast configuration. Only
+	// future slots need the immutable summary before cache mutation.
+	if known, err := ls.epochForSlot(blockSlot); err == nil {
```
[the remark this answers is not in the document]
wolf31o2: Not applicable at this head. `headerVerificationEpoch` no longer returns early on a cached epoch; the cache hit only decides whether the immutable summary has to be consulted, and the `len(epoch.Nonce) == 0` guard still runs on every path.

ledger/state.go
```diff
@@ -9910,9 +9918,12 @@ func (ls *LedgerState) ProtocolParamsForSlot(
...
 		return currentPParams
 	}
-	shape := ls.eraShape()
-	if len(shape.Eras) == 0 {
-		return currentPParams
+	shape, err := ls.eraShapeWithError()
+	if err != nil {
+		return nil
+	}
+	if _, ok := shape.EraForID(currentEra.Id); !ok {
```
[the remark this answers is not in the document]
wolf31o2: `currentEra`, `currentEpoch` and `currentPParams` are read from a single `loadConsensusSnapshot` value, which is published atomically, so the era walk and the epoch it starts from come from the same snapshot rather than from separate reads that could disagree. Not reproduced; no change.

ledger/verify_header.go
```diff
@@ -727,10 +728,18 @@ func (ls *LedgerState) genesisOverlayDelegationForBlock(
 	block ledger.Block,
 	shelleyGenesis *shelley.ShelleyGenesis,
 ) (genesisDelegation, genesisOverlaySlotStatus, error) {
+	pparams := ls.genesisOverlayProtocolParamsForBlock(block)
+	if pparams == nil {
```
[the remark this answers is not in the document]
wolf31o2: A Byron block returns at `verify_header.go:671`, before this branch, so Byron headers never reach it. The underlying condition is real for a non-Byron header whose parameters cannot be resolved, and that case is now wrapped in `errHeaderVerificationDeferred` instead of being rejected, so it no longer reaches `ConnectionRecycleRequestedEvent`. Covered by `TestGenesisOverlayUnresolvablePParamsDefers`, which fails when the classification is reverted.

ledger/slot_time_converter.go
```diff
@@ -426,6 +465,115 @@ func withinOperationalWindow(
...
+	if slotLenMs == 0 {
+		return time.Time{}, false
+	}
+	// slotLenMs is a protocol-bounded slot length in milliseconds.
+	// #nosec G115
+	return addRelativeTime(
+		sg.SystemStart,
+		0,
+		slot,
+		time.Duration(slotLenMs)*time.Millisecond,
```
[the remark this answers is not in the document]
wolf31o2: `genesisSlotTime` is the exact inverse of `nearNowSlot`, which `TimeToSlot` already applies with the same genesis slot length and the same near-now bound, and it is reached only for a slot past the epoch cache when no summary can be built. Restricting it to a resolved era would leave the operational slot clock unanswered in precisely the state it exists to cover. No change.

ledger/forecast_unavailable_test.go
```diff
@@ -0,0 +1,165 @@
...
+	dbtest "github.com/blinklabs-io/dingo/internal/test/dbtest"
+	"github.com/blinklabs-io/dingo/ledger/eras"
+	"github.com/blinklabs-io/gouroboros/ledger/babbage"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/dijkstra"
+	"github.com/blinklabs-io/gouroboros/ledger/shelley"
+	"github.com/stretchr/testify/require"
+)
+
+func TestProtocolParamsForSlot_UnavailableShape(t *testing.T) {
```
arepala-uml, line 34: New ledger-package test omits t.Parallel(); AGENTS.md requires it here and this fixture touches no process-global state.

ledger/forecast_unavailable_test.go
```diff
@@ -0,0 +1,165 @@
...
+				"current-epoch parameters do not require a forecast")
+			require.Nil(t, ls.ProtocolParamsForSlot(75),
+				"a future epoch with unavailable shape must not use current parameters")
+		})
+	}
+	// The existing boundary test exercises a valid scheduled Shelley-to-Allegra
+	// forecast; unavailable-shape handling must retain that path.
+}
+
+func TestProtocolParamsForSlot_UnavailableTransition(t *testing.T) {
```
arepala-uml, line 79: New ledger-package test omits t.Parallel(); AGENTS.md requires it here and this fixture touches no process-global state.

ledger/forecast_unavailable_test.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+			ls.publishSnapshotsLocked()
+			require.Same(t, ls.currentPParams, ls.ProtocolParamsForSlot(74))
+			require.Nil(t, ls.ProtocolParamsForSlot(75),
+				"an unresolved scheduled transition must not return pre-fork parameters")
+		})
+	}
+}
+
+func TestProtocolParamsForSlot_PendingUpdateFailure(t *testing.T) {
```
arepala-uml, line 114: New ledger-package test omits t.Parallel(); AGENTS.md requires it here and this fixture touches no process-global state.

ledger/forecast_unavailable_test.go
```diff
@@ -0,0 +1,165 @@
...
+			Logger:            slog.New(slog.DiscardHandler),
+		},
+	}
+	ls.publishSnapshotsLocked()
+	require.Same(t, pparams, ls.ProtocolParamsForSlot(99))
+	require.Nil(t, ls.ProtocolParamsForSlot(100),
+		"a failed pending update must not return stale parameters")
+}
+
+func TestGenesisOverlayRejectsUnavailableProtocolParams(t *testing.T) {
```
arepala-uml, line 140: New ledger-package test omits t.Parallel(); AGENTS.md requires it here and this fixture touches no process-global state.

ledger/forecast_unbuildable_test.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// TestHeaderVerificationEpoch_ForecastBuildFailureDeferred pins the
+// classification of a hard-fork summary that cannot be BUILT: the era shape,
+// the genesis behind it, and the epoch cache are local inputs, so the failure
+// says nothing about the header. ouroboros/chainsync.go routes every
+// non-deferred header error to ConnectionRecycleRequestedEvent, so returning
+// the build failure unwrapped recycles the honest peer that served the header
+// and stalls the node at every epoch boundary.
+func TestHeaderVerificationEpoch_ForecastBuildFailureDeferred(t *testing.T) {
```
arepala-uml, line 92: New ledger-package test omits t.Parallel(); AGENTS.md requires it here and this fixture touches no process-global state.

ledger/forecast_unbuildable_test.go
```diff
@@ -0,0 +1,207 @@
...
+		"an unbuildable forecast must not be reported as a peer fault")
+	require.True(t, IsHeaderVerificationDeferred(err))
+}
+
+// TestSlotToTime_CachedSlotWithoutForecast pins SlotToTime against
+// SlotToEpoch: a slot the epoch cache already covers has known era parameters
+// and must convert without a forecast. Returning the summary-build error
+// verbatim leaves the slot clock (ledger/slot_clock.go) unable to resolve a
+// slot boundary, retrying every 100ms for the life of the process.
+func TestSlotToTime_CachedSlotWithoutForecast(t *testing.T) {
```
arepala-uml, line 114: New ledger-package test omits t.Parallel(); AGENTS.md requires it here and this fixture touches no process-global state.

ledger/forecast_unbuildable_test.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// TestConsensusModeForEpoch_UnresolvableShapeFailsClosed pins the
+// forward-looking era walk as fail-closed. An unavailable shape breaks the
+// walk, and answering with the CURRENT era's mode for a future epoch reports
+// exactly what a scheduled hard fork changes. The control fixes Babbage at
+// epoch 501 under the same current era (Alonzo, TPraos), so the two cases
+// differ by mode and not merely by error: with the shape resolvable the
+// forecast is CPraos.
+func TestConsensusModeForEpoch_UnresolvableShapeFailsClosed(t *testing.T) {
```
arepala-uml, line 170: New ledger-package test omits t.Parallel(); AGENTS.md requires it here and this fixture touches no process-global state.

ledger/hardfork_summary_test.go
```diff
@@ -194,41 +188,83 @@ func TestHardForkSummary_EmptyCache(t *testing.T) {
...
-				StartSlot:     0,
-				SlotLength:    1000,
-				LengthInSlots: 100,
-				EraId:         1,
+// TestHardForkSummary_RejectsUnavailableShape verifies that consensus
+// forecasting fails closed when the current era cannot be resolved to the
+// configured hard-fork shape. Returning the cache-derived current era in any
+// of these cases would leave End nil with a zero safe zone and make every
+// future slot appear forecastable.
+func TestHardForkSummary_RejectsUnavailableShape(t *testing.T) {
```
arepala-uml, line 414: This table-driven test and its subtests omit t.Parallel(); AGENTS.md requires it in ledger and this fixture touches no process-global state.

ledger/leader/election_test.go
```diff
@@ -1450,3 +1456,39 @@ func TestElectionCanceledWaiterAfterReplacement(t *testing.T) {
 		t.Fatal("Start did not return after generation completion")
 	}
 }
+
+// TestComputeScheduleDeclinesUnresolvableConsensusMode pins the caller half of
+// the fail-closed consensus-mode forecast. The mode selects both the VRF input
+// construction and the threshold, so a schedule computed from a substituted
+// default is a leader-slot list cardano-node will reject. computeSchedule must
+// return the resolution error rather than produce one.
+func TestComputeScheduleDeclinesUnresolvableConsensusMode(t *testing.T) {
```
arepala-uml, line 1465: New leader-package test omits t.Parallel(); AGENTS.md names this package too and the fixture touches no process-global state.

ledger/verify_header_test.go
```diff
@@ -1166,6 +1178,39 @@ func TestValidateBlockHeaderCryptoDoesNotAdvanceEpochCache(t *testing.T) {
 		"header-only validation must not advance the shared epoch cache")
 }
 
+// Cached epoch resolution bypasses forecasting, but must not bypass the
+// nonce-availability contract shared by crypto and state verification callers.
+func TestHeaderVerificationEpoch_CachedNonceWithoutForecastConfig(t *testing.T) {
```
arepala-uml, line 1183: New ledger-package test omits t.Parallel(); AGENTS.md requires it here and this fixture touches no process-global state.

arepala-uml (dismissed):

arepala-uml (dismissed):

ledger/slot_time_converter.go
```diff
@@ -152,11 +152,37 @@ func (c *SlotTimeConverter) SlotToTime(slot uint64) (time.Time, error) {
...
+			slot,
+		); ok {
+			return when, nil
+		}
+		// Past the cache there are no era parameters to read, so fall back to
+		// the genesis slot length for near-now slots only -- the same bound,
+		// and the same inverse, as TimeToSlot's nearNowSlot fallback on this
+		// branch. Arbitrary slots stay unanswered.
+		if when, ok := genesisSlotTime(shelleyGenesis, slot); ok &&
+			isNearNow(c.now(), when) {
```
chrisguiney, line 181: The near-now genesisSlotTime fallback is the one new branch with no test. TestSlotToTime_CachedSlotWithoutForecast pins only the false side of isNearNow (slots 532000 and 99999 both error); nothing exercises a slot past the cache whose genesis-extrapolated time lands inside the window, so the value returned there is unpinned. SlotTimeConverter.nowFunc exists for exactly this and is injected at slot_test.go:374 and slot_time_converter_test.go:241, so a deterministic case costs one fixture.

ledger/state.go
```diff
@@ -10680,9 +10688,12 @@ func (ls *LedgerState) ProtocolParamsForSlot(
 	if slotEpoch <= currentEpoch.EpochId {
 		return currentPParams
 	}
-	shape := ls.eraShape()
-	if len(shape.Eras) == 0 {
-		return currentPParams
+	shape, err := ls.eraShapeWithError()
+	if err != nil {
+		return nil
```
chrisguiney, line 10693: Class audit of the fail-closed change: five eraShape() callers remain - queries.go:485 (already errors on an empty shape), state.go:6223, 8143, 8222 and verify_header.go:2512. Only validateEpochCacheForecast guards a forecast, and it is unreachable in the unavailable-shape state: headerVerificationEpoch defers before ensureEpochForSlot, and advanceEpochCache rejects an empty cache. The new successor-era error paths cannot fire on a constructed LedgerState, since activeEras and the shape both derive from eras.ActiveEras(EnableDijkstra).

ledger/hardfork_summary.go
```diff
@@ -237,10 +237,15 @@ func (ls *LedgerState) hardForkSummaryAnchoredAt(
 	past := eraSummaries[:len(eraSummaries)-1]
 
 	// Use the same configured safe-zone source as the NtC era-history query.
-	// Tests and early bootstrap callers may construct a LedgerState without a
-	// complete node configuration; in that case eraShape is unavailable and
-	// the legacy indefinite safe zone remains the only defensible answer.
-	shape := ls.eraShape()
+	// A missing shape or current-era entry cannot safely supply a forecast:
+	// the cache-derived current era has a zero safe zone and no end bound.
+	shape, err := ls.eraShapeWithError()
```
chrisguiney, line 242: Verified on a tree with current origin/main merged in: build, go vet ./ledger/..., golangci-lint run ./ledger/... (0 issues), the ledger forecast tests, ledger/leader, ledger/forging and internal/test/conformance all pass. Restoring this file to its base version in place fails TestHardForkSummary_RejectsUnavailableShape, TestHeaderVerificationEpoch_ForecastBuildFailureDeferred and TestSlotToTime_CachedSlotWithoutForecast, so the new tests are causal. The head merge left every branch-owned file intact.

chrisguiney (changes requested):
go-test (Windows) (1.26.x) is red at this head on TestCheckpointWALDoesNotBlockWriteBehindReaderSnapshot in database/plugin/metadata/sqlite, a package this diff does not touch; that test landed on main in 0d18c8ae (#4276) and its 2s wall-clock bound measured 2.194s, so the failure belongs to the base branch. A red required check withholds approval. The forecast change itself verifies clean against current main; the one gap is the untested near-now fallback noted inline.

ledger/forecast_unavailable_test.go
wolf31o2: Fixed in 09944c70: the top-level test and its isolated subtests now call t.Parallel().

ledger/forecast_unavailable_test.go
wolf31o2: Fixed in 09944c70: the top-level test and its isolated subtests now call t.Parallel().

ledger/forecast_unavailable_test.go
wolf31o2: Fixed in 09944c70: the isolated ledger test now calls t.Parallel().

ledger/forecast_unavailable_test.go
wolf31o2: Fixed in 09944c70: the top-level test and its isolated subtests now call t.Parallel().

ledger/forecast_unbuildable_test.go
wolf31o2: Fixed in 09944c70: the isolated ledger test now calls t.Parallel().

ledger/forecast_unbuildable_test.go
wolf31o2: Fixed in 09944c70: the isolated ledger test now calls t.Parallel().

ledger/forecast_unbuildable_test.go
wolf31o2: Fixed in 09944c70: the isolated ledger test now calls t.Parallel().

ledger/hardfork_summary_test.go
wolf31o2: Fixed in 09944c70: both the table-driven test and each isolated subtest now call t.Parallel().

ledger/leader/election_test.go
wolf31o2: Fixed in 09944c70: the isolated leader test now calls t.Parallel().

ledger/verify_header_test.go
wolf31o2: Fixed in 09944c70: the top-level test and each isolated subtest now call t.Parallel().

ledger/slot_time_converter.go
wolf31o2: Fixed in 09944c70. TestSlotToTime_NearNowWithoutForecast fixes the clock with nowFunc at the first slot past the cache and asserts the genesis fallback result. Disabling only that fallback makes the test fail with the forecast-unavailable error.

ledger/state.go
wolf31o2: Confirmed; this is an informational class audit and requires no additional code change.

ledger/hardfork_summary.go
wolf31o2: Confirmed. The full ledger and ledger/leader race suites pass after the added parallel coverage.

Issue #2177, opened 2026-05-06
chainsync: local tip pinned at S after plateau on single-peer BP


wolf31o2:
Issue #2150 surfaced a single-peer block producer wedge where local_tip stops advancing while the upstream peer reports a higher tip. PR #2176 removes the symptom of disconnecting the only relay every plateau cycle, but does not address why local_tip becomes pinned at slot S in the first place. This issue tracks that deeper bug.

Evidence:

- local_tip stays at slot S for tens of minutes while the peer reports tips advancing normally
- "fork extends from current tip, adding headers without rollback" briefly advances local_tip from S to S+200, but the next plateau evaluation shows local_tip back at S
- "replayed peer header history after local rollback" fires every cycle with header_count=2 against each new connection_id
- fork_depth grows monotonically across cycles (e.g. 270 to 424 over 50 minutes against the same fork_point_slot=S)
- dropped_rollbacks=0 throughout the wedge
- No VRF errors, no panics, no rollback-loop warnings

Scope:

- Investigate the chain-extend path in ledger/chainsync.go around the "fork extends from current tip, adding headers without rollback" branch (~line 1819). Headers appear to be queued via AddBlockHeader but the local tip advancement does not stick.
- Determine where the revert to slot S originates. Candidates worth ruling out: a downstream blockfetch failure that rolls the chain back, a ledger-state advancement step that fails after the header is queued, or interaction with the rollback recovery path that treats the queued headers as out-of-band.
- A SIGQUIT goroutine dump from a wedged container would help characterize whether any goroutine is holding state across the apparent recovery.

Reproducer:

- Single configured outbound relay (Haskell cardano-node) on preview testnet
- Block producer role, mithril-bootstrapped
- Wedge fires every ~4-5 min indefinitely; restart recovers within seconds and re-wedges after several hours

Reported on dingo v0.40.0 from block producers with a single Haskell relay each, on both linux/amd64 and linux/arm64.

chadaibbhot-git:
Captured a SIGQUIT goroutine dump on a single-peer BP that's been wedged for ~99 minutes on a build with PR #2176 cherry-picked. PR #2176 is doing exactly what it should (preserves the only peer instead of disconnecting on plateau), and that gave us the long stable wedge needed for a useful dump.

**Build:** v0.40.0 + PR #2176 cherry-pick (verified against the upstream branch tip).
**Network:** preview testnet, Conway era.
[...]

PR #3877, opened 2026-09-04
fix(ledger): restore UTxOs during at-tip recovery

### LINKED ISSUE #3766: ledger: at-tip recovery rewind leaves the UTxO set inconsistent, degrading to a halted pipeline at the Mithril trust boundary
## Summary

On a Mithril-bootstrapped preview node, at-tip validation recovery rewinds the
primary chain but does not restore the UTxO set to match. Blocks that the node
had already applied cleanly become unapplyable after the rewind, failing with
`rule 29: bad input(s)` and `rule 31: value not conserved: consumed 0`. Each
failure triggers a deeper rewind, so the node walks itself down to the Mithril
trust boundary, runs out of room to rewind, and halts.

The node ends in a state it reports as needing an operator, with no way forward
short of discarding the database.

```
level=WARN  msg="block processing hit persistent validation failure, halting ledger pipeline"
  error="process block batch: at-tip recovery rollback exceeds Mithril trust boundary:
         persistent tx validation failure after recovery attempts"
level=ERROR msg="ledger pipeline halted on an unrepairable validation failure; the node has
  stopped following the chain and will not resume without operator intervention"
  component=ledger tip_slot=121616075
```

## Why this is a state bug and not a bad block

The block the node cannot apply at the end is one it already applied
successfully earlier in the same run.

Timeline from a single container run on preview, Mithril anchor at slot
121616068 (epoch 1407 snapshot, immutable 28151):

- 02:10:56  node starts, imports anchor, begins following
- 02:10:57  applies forward past 121616218 without error
- 02:15:00  ledger tip reaches 121651198, roughly 35k slots past the anchor
- 03:11:21  forges and adopts its own block at slot 121662681
- 03:2x     validation failure at the epoch boundary starts at-tip recovery
- 03:46:40  `at-tip recovery exhausted scheduled rewind attempts, retrying with deepest rewind`
            failing_slot=121616218
- 03:50:56  halts, tip_slot=121616075

Block 121616218 is 150 slots past the anchor and was applied without complaint
at 02:10:57. After the rewind sequence it fails:

```
tx e331c67adc10e10191f35b38a352d5464b6ac3e67f4eac8d66c0a7a3396271ac
conway utxo validation rule 29: bad input(s): <input>#0
conway utxo validation rule 31: value not conserved: consumed 0, produced 4017569819
```

`consumed 0` means the inputs resolve to nothing at all, so the UTxO rows the
block spends are simply absent from the store. Nothing about the block changed
between the two attempts; the local state did.

The same run logged 184 occurrences of

```
continuation block spends an input with no producer on the local applied chain
```

which is the same missing-producer condition observed from the chain-selection
side.

## What set it off

The first validation failure after the epoch 1407 to 1408 transition was a
withdrawal amount disagreement, not a UTxO problem:

```
conway utxo validation rule 46: incorrect withdrawal amount for
provided 350033351, balance 0
```

The node had logged `applied stake rewards reward_snapshot_epoch=1405
performance_epoch=1406 pots_epoch=1407` shortly before, having bootstrapped
mid-epoch-1407, so it computed that distribution from partial history. It also
logged `skipped epoch lab recovery for epochs without stored candidate nonce
count=7 first_epoch=1400 last_epoch=1406`.

That first failure is worth looking at on its own, but it is not the reported
bug here. The reported bug is that recovery from any such failure leaves the
UTxO set inconsistent with the rewound chain, which converts a single
recoverable disagreement into an unrecoverable node.

## Impact

This is a block-producer availability bug. The node stops following the chain
and stops forging, and because the damage is in the database, restarting does
not help. On a Mithril-bootstrapped node the floor is close to the tip, so
there is very little rewind headroom before the halt.

## Environment

- preview, Conway, epoch 1408
- Mithril bootstrap, anchor slot 121616068, epoch 1407, immutable 28151
- `storageMode: core`, sqlite metadata, badger blob
- build from PR #3735 head b027d6cf (based on 83ee8cdb), needed to get past the
  Plutus budget rejection described in #3627 and blinklabs-io/gouroboros#2164;
  the rewind behaviour here is independent of that change, which only affects
  whether a producer-valid budget overage is accepted

## Suggested direction

Either make the at-tip recovery rewind restore the UTxO set atomically with the
chain rewind, or make it refuse to rewind when it cannot, so the node stops with
its state intact rather than degrading it further on each attempt. Silently
continuing after a rewind that did not restore state is what turns one bad
block into a dead database.

Related: #2971 covered the replay-path version of the same missing-producer
condition and is closed; this is the at-tip recovery path.

wolf31o2:
## Summary

- restore UTxOs and speculative metadata left above the durable ledger tip during the first same-tip recovery
- apply the repair consistently to at-tip transaction, deterministic transaction, replay-holding, and header-validation recovery
- bound repeated identical or held recoveries so they reuse the completed repair
- document the same-tip recovery contract and add regression coverage for the repair latches

Fixes #3766

## Validation

- `go test -race ./ledger/`
- focused race tests for at-tip, header-validation, and replay-holding repair behavior
- mutation checks proving each regression test fails when its repair latch is removed or moved before an early decline
- `make docs-parity`
- `make golines`
- `git diff --check`


wolf31o2:
This closes #3766 on the same-tip path. The escalating path in the same issue has a separate cause that this change does not reach, split out as #3953 with a fix in #3952: `Database.UtxosDeleteConsumed` hard-deletes spent rows at `deleted_slot <= tip - stabilityWindow`, and `TruncateAfterSlot` restores spent UTxOs with `UPDATE ... WHERE deleted_slot > ?`, so a rewind below the swept slot restores nothing and reports the tip repaired. `repairSameTip` does not apply there — the target is a lower slot and the sweep runs as before.

The regression test in #3952 fails at `81c876e7`: with a 25920-slot window and the sweep floor at 114080, recovery steps the tip 140000 -> 114080 -> 88160 and an output produced at slot 50000 and consumed at slot 110000 is absent from the live set at tip 88160.

arepala-uml (dismissed):
Verified the fix in an isolated worktree: reverted just the `repairSameTip` behavior and confirmed the updated regression test fails without it (`Should be zero, but was 160`), then confirmed it passes with the fix restored. `go build`/`go vet ./ledger/...` clean. Full `-race` run of the ledger package's rollback/at-tip/same-tip tests (30+) passes.

Traced the event-suppression logic: `ChainsyncResyncReasonLocalLedgerRollback` is gated to the single `RecoverAfterLocalRollback` call site in `ouroboros/chainsync.go`, so suppressing it via `!sameTip` correctly avoids reintroducing the issue #2177 wedge, while the caller's own `ChainsyncResyncReasonLiveTxValidationRecovery` event still fires for peer rotation.

Cubic reports no issues; CodeRabbit rate-limit is already documented on this PR.

One non-blocking item: `ARCHITECTURE.md:3978-3998` documents at-tip recovery's rewind/rollback behavior in the same area this PR changes, but doesn't mention that the first-attempt same-tip case now runs a full metadata restore instead of no-op. Left an inline comment with the specific line.

ledger/state.go
```diff
@@ -2934,6 +2934,17 @@ func (ls *LedgerState) cleanupConsumedUtxos() {
...
+	return ls.rollbackWithOptions(point, false)
+}
+
+// rollbackWithOptions restores metadata even when point is already the
+// in-memory ledger tip when repairSameTip is set. At-tip validation recovery
+// can leave consumed UTxOs above the durable tip (for example after Mithril
+// gap replay), so the usual same-point no-op must not skip UTxO restoration.
+func (ls *LedgerState) rollbackWithOptions(
+	point ocommon.Point,
+	repairSameTip bool,
```
arepala-uml, line 3116: ARCHITECTURE.md:3978-3998 documents at-tip recovery's rewind/rollback flow in detail but doesn't mention that the first-attempt same-tip case now runs a full metadata restore instead of no-op. Worth a short addition here or in a fast-follow; not a merge blocker.

arepala-uml (changes requested):
Requesting changes for the merge conflict only — the fix itself is correct and narrowly scoped.

**Blocking: merge conflict.** GitHub reports this branch as CONFLICTING against main. This is a structural collision, not textual noise: this PR renames `rollback` → `rollbackWithOptions(point, repairSameTip)` to add the same-tip repair path, but main has since independently refactored the *same function* into `rollbackWithResync`/`rollbackWithoutResync` + a new `publishLocalLedgerRollback` helper, touching the identical same-tip short-circuit and the final event-publish block. Two independent refactors landed on the same function — this needs a manual reconciliation of `repairSameTip` semantics into the newer structure, not just a mechanical rebase.

**Code review (no other blockers):**
- Confirmed the actual bug: the same-tip rollback path was a pure no-op that never restored UTxOs marked spent by a block whose validation later failed (a crash/partial-apply artifact) — those UTxOs stayed permanently spent. The fix correctly forces the same-tip branch through the full restore sweep only for this one caller (`recoverAtTipFromTxValidationError`).
- Verified the boundary is correct: the restore predicate is `deleted_slot > slot` (strictly greater), so a UTxO spent *at* the tip is correctly left alone — no off-by-one, and this predicate is pre-existing, untouched code.
- Confirmed by reverting the one-line fix that the new test fails without it, and confirmed all 7 other `rollback` call sites are unaffected (same-tip no-op preserved exactly, pinned by the pre-existing `TestRollbackAtCurrentTipIsNoop`).
- No double-spend/cache-invalidation exposure — dingo has no in-memory UTxO cache, queries read fresh from the metadata store.

One non-blocking item from an earlier review on this PR (also mine, since dismissed) that's still worth addressing: ARCHITECTURE.md/DATABASE.md don't yet mention this new same-tip repair override in the recovery-flow description. Not a blocker, but worth folding in while this is being rebased anyway.

ledger/state.go
```diff
@@ -3181,8 +3155,9 @@ func (ls *LedgerState) rollbackWithResync(
 	currentTip := ls.currentTip
 	mithrilLedgerSlot := ls.mithrilLedgerSlot
 	ls.RUnlock()
-	if currentTip.Point.Slot == point.Slot &&
-		bytes.Equal(currentTip.Point.Hash, point.Hash) {
+	sameTip := currentTip.Point.Slot == point.Slot &&
+		bytes.Equal(currentTip.Point.Hash, point.Hash)
+	if sameTip && !repairSameTip {
```
arepala-uml, line 3310: This same-tip check is bypassed only when the caller passes `repairSameTip=true`. Three sibling callers hit the identical `rewindPoint == currentTip` condition and still call `rollback()`/`rollbackWithoutResync()` (`repairSameTip=false`): `recoverFromDeterministicTxValidationError` (replay_recovery.go:682), the replay-holding path (replay_recovery.go:353), and header_validation_recovery.go:194. The at-tip UTxO-restoration bug this PR fixes is still reachable through them, and `isDeterministicTxValidationError` is checked before `IsAtTip()` in the dispatcher, so a large class of at-tip failures routes to the unpatched path.

ledger/replay_recovery.go
```diff
@@ -1323,25 +1323,14 @@ func (ls *LedgerState) recoverAtTipFromTxValidationError(
...
-	// treat that as a fact about the pin rather than about the rules, and
-	// re-measure after any gouroboros bump instead of trusting this line.
-	// Stale numbers here have twice pointed diagnosis at the wrong root
-	// cause (#3165, #3678).
-	if err := ls.rollback(rewindPoint); err != nil {
+	// until process restart. Primary-chain rollback only touches the chain
+	// store — the matching ledger rollback must be explicit. Match on rule
+	// names rather than their numeric positions because upstream changes the
+	// validation-rule slice between releases.
+	if err := ls.rollbackWithOptions(rewindPoint, true, true); err != nil {
```
arepala-uml: `repairSameTip=true` is passed unconditionally here with no backoff. Once `findRewindPoint` can no longer find a deeper on-chain point (the steady state for a persistently failing/looping fork), `rewindPoint` permanently equals `currentTip`, so every retry now runs the full `TruncateAfterSlot` transaction, an epoch cache reload, and `evaluateHardForkInitiationStability()` instead of the previous single-call `enforceDurableTipFloor()` no-op — turning the cheapest, most frequent path into the most expensive one under a sustained adversarial/misbehaving-peer condition.

arepala-uml (changes requested):
Local review findings against head e2fa8c6.

1. The same-tip UTxO-restoration gap this PR fixes for `recoverAtTipFromTxValidationError` is still present in three sibling call sites that compute the identical `rewindPoint == currentTip` condition and still call the unrepaired `rollback()`/`rollbackWithoutResync()`: `recoverFromDeterministicTxValidationError` (ledger/replay_recovery.go:682), the replay-holding path (ledger/replay_recovery.go:353), and header-validation recovery (ledger/header_validation_recovery.go:194). Since `isDeterministicTxValidationError` is checked before `IsAtTip()` in the dispatcher, many at-tip failures route to the still-unpatched deterministic path rather than the fixed one, so the original bug remains reachable.
2. `recoverAtTipFromTxValidationError` passes `repairSameTip=true` unconditionally with no backoff (inline comment). Every retry of a persistent at-tip failure now pays a full DB transaction, epoch/pparams reload, and hard-fork-stability re-evaluation instead of the previous single `enforceDurableTipFloor()` check, on exactly the sustained-failure path the retry loop exists to tolerate cheaply.

Non-blocking: `rollbackWithOptions`'s positional `(bool, bool)` signature has no compiler-enforced meaning; the slot+hash equality check it introduces duplicates an existing idiom that already appears 7 other places in this file; and the doc comment dropped alongside the rename discarded the historical rationale (issues #3165/#3678) for matching validation rules by name rather than index.

ledger/replay_recovery.go
```diff
@@ -1447,25 +1455,19 @@ func (ls *LedgerState) recoverAtTipFromTxValidationError(
...
-	if err := ls.rollback(rewindPoint); err != nil {
+	// until process restart. Primary-chain rollback only touches the chain
+	// store — the matching ledger rollback must be explicit. Match on rule
+	// names rather than their numeric positions because upstream changes the
+	// validation-rule slice between releases.
+	// The first recovery at a tip may need to repair metadata left behind by
+	// the failed block. Repeating the same failure at the same tip does not:
+	// the first repair already restored every UTxO above that tip, while
+	// re-running it would turn the retry loop into a full database rollback.
+	repairSameTip := !isSameFailure && pointMatches(rewindPoint, ledgerTip.Point)
```
arepala-uml, line 1472: No test drives this same-failure latch: dropping the isSameFailure guard here (restoring the unconditional repeat repair) leaves the full ledger suite green, so nothing proves a second identical failure at the same tip skips the repair.

arepala-uml (changes requested):
The at-tip UTxO restore fix is well tested; the later same-tip repeated-failure bounding fix (four call sites) has no test proving it actually bounds the repair.

ledger/header_validation_recovery.go
```diff
@@ -90,6 +90,11 @@ func (ls *LedgerState) tryRecoverFromHeaderValidationError(
 	ls.RLock()
 	ledgerTip := ls.currentTip
 	ls.RUnlock()
+	sameFailureAtTip := ls.lastHeaderValidationFailure != nil &&
+		pointMatches(ls.lastHeaderValidationFailure.BlockPoint, validationErr.BlockPoint) &&
+		pointMatches(ls.lastHeaderValidationTip, ledgerTip.Point)
+	ls.lastHeaderValidationFailure = validationErr
```
arepala-uml: lastHeaderValidationFailure/lastHeaderValidationTip are recorded before the Mithril-boundary early return and are never reset. A test that declines this failure once via a tripped Mithril boundary, then retries the identical failure at the identical tip once the boundary lifts, shows the retry taking the sameTip-no-repair path in rollbackWithOptions: rewardInputGeneration stays unchanged instead of increasing, so the ledger-metadata repair never runs even though this retry is the first attempt that actually performs the rewind.

Test added locally to ledger/header_validation_recovery_test.go, reproducing the bug against this commit:

```go
func TestHeaderValidationRecoverySkipsRepairAfterMithrilDeclineAtSameTip(
	t *testing.T,
) {
	t.Parallel()

	buildFixture := func(t *testing.T) (*LedgerState, models.Block) {
		t.Helper()
		db, err := dbtest.NewDatabase(t, &database.Config{DataDir: ""})
		require.NoError(t, err)
		t.Cleanup(func() { dbtest.CloseDatabase(db) }) //nolint:errcheck

		blocks := make([]models.Block, 0, 5)
		for slot := uint64(1); slot <= 5; slot++ {
			block := makeTestBlock(slot, slot)
			if len(blocks) > 0 {
				block.PrevHash = append(
					[]byte(nil),
					blocks[len(blocks)-1].Hash...,
				)
			}
			blocks = append(blocks, block)
			require.NoError(t, db.BlockCreate(block, nil))
		}

		cm, err := chain.NewManager(db, nil)
		require.NoError(t, err)
		require.NoError(
			t,
			cm.SetLedger(testSecurityParamLedger{securityParam: 2}),
		)

		ledgerTipBlock := blocks[2]
		ledgerTip := ochainsync.Tip{
			Point:       makeTestPoint(ledgerTipBlock),
			BlockNumber: ledgerTipBlock.Number,
		}
		require.NoError(t, db.SetTip(ledgerTip, nil))

		bus := event.NewEventBus(nil, nil)
		t.Cleanup(bus.Close)

		ls := &LedgerState{
			db:    db,
			chain: cm.PrimaryChain(),
			config: LedgerStateConfig{
				ChainManager: cm,
				EventBus:     bus,
				Logger:       slog.New(slog.NewJSONHandler(io.Discard, nil)),
			},
		}
		ls.currentTip = ledgerTip
		ls.metrics.init(prometheus.NewRegistry())
		require.NoError(t, ls.reconcilePrimaryChainTipWithLedgerTip())
		return ls, blocks[3]
	}

	// Control: a genuine first-time failure at this tip must run the
	// ledger-metadata repair as part of the rollback.
	control, rejectedControl := buildFixture(t)
	generationBeforeControl := control.rewardInputGeneration.Load()
	recovered, err := control.tryRecoverFromHeaderValidationError(
		&headerValidationError{
			BlockPoint: makeTestPoint(rejectedControl),
			Cause:      errors.New("VRF leader value exceeds threshold"),
		},
	)
	require.NoError(t, err)
	require.True(t, recovered)
	require.Greater(
		t,
		control.rewardInputGeneration.Load(),
		generationBeforeControl,
		"a genuine first-time failure at this tip must run the "+
			"ledger-metadata repair",
	)

	// Recurring: the identical failure, at the identical tip, first declined
	// by a Mithril boundary and then retried once the boundary lifts. The
	// earlier attempt never reached rollbackWithOptions, so this retry is
	// also the first attempt that actually performs a rewind, and must
	// repair too.
	recurring, rejectedRecurring := buildFixture(t)
	recurring.mithrilLedgerSlot = recurring.currentTip.Point.Slot + 1000
	validationErr := &headerValidationError{
		BlockPoint: makeTestPoint(rejectedRecurring),
		Cause:      errors.New("VRF leader value exceeds threshold"),
	}
	recovered, err = recurring.tryRecoverFromHeaderValidationError(
		validationErr,
	)
	require.NoError(t, err)
	require.False(
		t,
		recovered,
		"the Mithril boundary must decline recovery on the first attempt",
	)

	recurring.mithrilLedgerSlot = 0
	generationBeforeRecurring := recurring.rewardInputGeneration.Load()
	recovered, err = recurring.tryRecoverFromHeaderValidationError(
		validationErr,
	)
	require.NoError(t, err)
	require.True(t, recovered)
	require.Greater(
		t,
		recurring.rewardInputGeneration.Load(),
		generationBeforeRecurring,
		"the attempt that actually performs the rewind must still run "+
			"the ledger-metadata repair, even though an identical "+
			"failure was already recorded on a prior, declined attempt",
	)
}
```

Failure against this commit (go test ./ledger/ -run TestHeaderValidationRecoverySkipsRepairAfterMithrilDeclineAtSameTip -v):

```
    header_validation_recovery_test.go:310:
        Error Trace:  ledger/header_validation_recovery_test.go:310
        Error:        "0" is not greater than "0"
        Test:         TestHeaderValidationRecoverySkipsRepairAfterMithrilDeclineAtSameTip
        Messages:     the attempt that actually performs the rewind must still run the ledger-metadata repair, even though an identical failure was already recorded on a prior, declined attempt
--- FAIL: TestHeaderValidationRecoverySkipsRepairAfterMithrilDeclineAtSameTip (0.10s)
FAIL
```

ledger/replay_recovery.go
```diff
@@ -350,7 +350,11 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 	// available. If metadata synchronization fails, the primary chain is still
 	// at a valid retained point and the standard divergence reconciler can
 	// finish rolling metadata back to its common ancestor.
-	if err := ls.rollback(rewindPoint); err != nil {
+	if err := ls.rollbackWithOptions(
+		rewindPoint,
+		replayHolding && pointMatches(rewindPoint, ledgerTip.Point),
```
arepala-uml: repairSameTip here is replayHolding && pointMatches(rewindPoint, ledgerTip.Point); once replay recovery starts holding at an unchanged tip this reduces to replayHolding alone, which stays true every cycle. A test that reaches two consecutive held recovery cycles at an unchanged tip shows the full metadata repair re-running on the second cycle too: rewardInputGeneration increases again instead of staying flat, proving the repair is not gated to first occurrence.

Test added locally to ledger/replay_recovery_test.go, reproducing the bug against this commit:

```go
func TestTryRecoverFromTxValidationErrorReplayHoldingRepairsOnlyOnce(
	t *testing.T,
) {
	t.Parallel()

	db, err := dbtest.NewDatabase(t, &database.Config{
		DataDir: t.TempDir(),
	})
	require.NoError(t, err)
	t.Cleanup(func() { require.NoError(t, dbtest.CloseDatabase(db)) })

	bus := event.NewEventBus(nil, nil)
	t.Cleanup(func() { bus.Stop() })

	cm, err := chain.NewManager(db, bus)
	require.NoError(t, err)

	parentBlock := testRawBlock("holding-once-parent", 100, 1, nil)
	replayOneBlock := testRawBlock(
		"holding-once-replay-1",
		120,
		2,
		parentBlock.Hash,
	)
	ledgerTipBlock := testRawBlock(
		"holding-once-ledger-tip",
		140,
		3,
		replayOneBlock.Hash,
	)
	failingBlock := testRawBlock(
		"holding-once-failure",
		162,
		4,
		ledgerTipBlock.Hash,
	)
	require.NoError(
		t,
		cm.PrimaryChain().AddRawBlocks(
			[]chain.RawBlock{
				parentBlock,
				replayOneBlock,
			},
		),
	)
	for _, block := range []chain.RawBlock{ledgerTipBlock, failingBlock} {
		require.NoError(t, db.BlockCreate(models.Block{
			Hash:     block.Hash,
			PrevHash: block.PrevHash,
			Cbor:     block.Cbor,
			Slot:     block.Slot,
			Number:   block.BlockNumber,
			Type:     block.Type,
		}, nil))
	}

	activeConnId := ouroboros.ConnectionId{
		LocalAddr: &net.TCPAddr{
			IP:   net.ParseIP("127.0.0.1"),
			Port: 3000,
		},
		RemoteAddr: &net.TCPAddr{
			IP:   net.ParseIP("192.0.2.30"),
			Port: 3001,
		},
	}

	nodeConfig := newTestShelleyGenesisCfg(t)
	nodeConfig.ShelleyGenesis().SecurityParam = 2
	ls, err := NewLedgerState(LedgerStateConfig{
		Database:          db,
		ChainManager:      cm,
		EventBus:          bus,
		CardanoNodeConfig: nodeConfig,
		Logger: slog.New(
			slog.NewJSONHandler(io.Discard, nil),
		),
		GetActiveConnectionFunc: func() *ouroboros.ConnectionId {
			return &activeConnId
		},
	})
	require.NoError(t, err)
	ls.currentEra.Id = 1
	require.NoError(t, cm.SetLedger(ls))
	ls.metrics.init(prometheus.NewRegistry())

	ledgerTip := ochainsync.Tip{
		Point: ocommon.NewPoint(
			ledgerTipBlock.Slot,
			ledgerTipBlock.Hash,
		),
		BlockNumber: ledgerTipBlock.BlockNumber,
	}
	require.NoError(
		t,
		db.SetBlockNonce(
			ledgerTip.Point.Hash,
			ledgerTip.Point.Slot,
			[]byte("nonce-holding-once-tip"),
			false,
			nil,
		),
	)
	require.NoError(t, db.SetTip(ledgerTip, nil))
	ls.currentTip = ledgerTip
	ls.currentTipBlockNonce = []byte("nonce-holding-once-tip")
	ls.publishSnapshotsLocked()

	// Model the two earlier recovery cycles, exactly as the sibling
	// non-converging test does, so the third cycle below is the first one
	// where replayRecoveryHolding flips true.
	require.False(t, ls.observeReplayRecoveryTip(ledgerTip.Point.Slot))
	require.False(t, ls.observeReplayRecoveryTip(ledgerTip.Point.Slot))

	txErr := func() *txValidationError {
		return &txValidationError{
			BlockPoint: ocommon.NewPoint(
				failingBlock.Slot,
				failingBlock.Hash,
			),
			TxHash: testHashBytes("holding-once-tx"),
			Inputs: []lcommon.TransactionInput{
				&replayRecoveryInput{
					txId:  testHashBytes("holding-once-missing-producer"),
					index: 0,
				},
			},
			Cause: errors.New("bad input"),
		}
	}

	// Cycle 3: replayRecoveryHolding flips true for the first time. The
	// repair must run here.
	generationBeforeFirstHold := ls.rewardInputGeneration.Load()
	recovered, err := ls.tryRecoverFromTxValidationError(txErr())
	require.NoError(t, err)
	require.True(t, recovered)
	require.True(t, ls.replayRecoveryHolding)
	require.Greater(
		t,
		ls.rewardInputGeneration.Load(),
		generationBeforeFirstHold,
		"the first held cycle must run the ledger-metadata repair",
	)

	// Cycle 4: still holding at the identical, unchanged tip -- no forward
	// progress occurred, so this is a repeat of the same stalled condition,
	// not a new failure. The repair must not run again.
	generationBeforeSecondHold := ls.rewardInputGeneration.Load()
	recovered, err = ls.tryRecoverFromTxValidationError(txErr())
	require.NoError(t, err)
	require.True(t, recovered)
	require.True(t, ls.replayRecoveryHolding)
	require.Equal(
		t,
		generationBeforeSecondHold,
		ls.rewardInputGeneration.Load(),
		"a second consecutive held cycle at an unchanged tip must not "+
			"re-run the ledger-metadata repair; repairSameTip should gate "+
			"on first occurrence only, like the sibling recovery sites do",
	)
}
```

Failure against this commit (go test ./ledger/ -run TestTryRecoverFromTxValidationErrorReplayHoldingRepairsOnlyOnce -v):

```
    replay_recovery_test.go:1819:
        Error Trace:  ledger/replay_recovery_test.go:1819
        Error:        Not equal:
                      expected: 0x2
                      actual  : 0x4
        Test:         TestTryRecoverFromTxValidationErrorReplayHoldingRepairsOnlyOnce
        Messages:     a second consecutive held cycle at an unchanged tip must not re-run the ledger-metadata repair; repairSameTip should gate on first occurrence only, like the sibling recovery sites do
--- FAIL: TestTryRecoverFromTxValidationErrorReplayHoldingRepairsOnlyOnce (0.19s)
FAIL
```

arepala-uml (changes requested):

ledger/state.go
wolf31o2: Fixed in e89eab05. ARCHITECTURE.md now documents the first completed same-tip metadata repair, early exits that do not consume it, and the repeated-failure reuse rule.

ledger/state.go
wolf31o2: Fixed on the current head. Deterministic recovery now gates repair on !resyncSpent, replay holding gates it to the transition into the held state, and header recovery uses its same-failure-at-tip latch. All three paths call rollbackWithOptions with the appropriate repairSameTip value.

ledger/replay_recovery.go
wolf31o2: Fixed on the current head. At-tip recovery computes repairSameTip as !isSameFailure && pointMatches(...), so an identical retry at an unchanged tip reuses the prior repair. e89eab05 adds a mutation-proven regression test for this latch.

ledger/header_validation_recovery.go
```diff
@@ -191,7 +191,11 @@ func (ls *LedgerState) tryRecoverFromHeaderValidationError(
...
 	// path.
-	if err := ls.rollback(rewindPoint); err != nil {
+	if err := ls.rollbackWithOptions(
+		rewindPoint,
+		pointMatches(rewindPoint, ledgerTip.Point),
+		true,
+	); err != nil {
 		return false, fmt.Errorf(
 			"rollback ledger state after header validation failure: %w",
 			err,
```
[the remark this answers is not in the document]
wolf31o2: Fixed on the current head. Header recovery records the last completed failure and tip and disables same-tip repair for an identical repeat. e89eab05 also ensures an early decline does not poison that latch.

ledger/replay_recovery.go
```diff
@@ -679,7 +683,11 @@ func (ls *LedgerState) recoverFromDeterministicTxValidationError(
 			err,
 		)
 	}
-	if err := ls.rollback(rewindPoint); err != nil {
+	if err := ls.rollbackWithOptions(
+		rewindPoint,
+		pointMatches(rewindPoint, ledgerTip.Point),
+		true,
```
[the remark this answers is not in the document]
wolf31o2: Fixed on the current head: repairSameTip is gated by !resyncSpent as suggested, preserving the first repair and skipping repeats after the deterministic recovery latch fires.

ledger/replay_recovery.go
wolf31o2: Fixed in e89eab05. TestAtTipRecoveryRepairsSameTipOnlyOnce verifies that the first recovery increments the metadata generation and an identical retry does not. Removing the isSameFailure guard makes the test fail with generation 2 -> 4.

ledger/header_validation_recovery.go
wolf31o2: Fixed in e89eab05. The header latch is now recorded only after the rewind and metadata rollback complete. The existing recovery test now declines once at the Mithril boundary, lifts it, and proves the first state-changing retry still performs the repair; moving the latch back before the early return makes it fail.

ledger/replay_recovery.go
wolf31o2: Fixed in e89eab05. Replay recovery captures the pre-observation holding state and enables same-tip repair only on the transition into holding. The non-converging regression now executes two held cycles and proves the second leaves rewardInputGeneration unchanged; removing the transition gate makes it fail.

arepala-uml (approved):

PR #3884, opened 2026-09-04
fix(ledger): hide Dingo's synthetic PlutusV2 cost model from GetCurrentProtocolParams

### LINKED ISSUE #3825: node-parity: Dingo's protocol-parameter reply includes a plutusV1 cost model cardano-node's doesn't
Same run as blinklabs-io/dingo#3824 (TestLedgerStateConsensus, PR #3806,
accelerated conformance devnet, Conway era / protocolVersion.major=10) found
a protocol-parameters divergence too. Diffing both sides' full LocalStateQuery
snapshots field by field, every field matched exactly except costModels:

- Dingo: costModels has both plutusV1 and plutusV2 entries.
- cardano-node (reference): costModels has only plutusV2.

Dingo answers this query (ShelleyCurrentProtocolParamsQuery,
`ledger/queries.go:438-439`) directly from `loadConsensusSnapshot().currentPParams`,
its in-memory protocol-parameter state.

Not yet root-caused: whether cardano-node's genesis/config for this devnet
genuinely omits a V1 cost model (in which case Dingo including one anyway is
the bug -- it should reflect the protocol parameters the network was actually
configured with, not every cost model it knows how to price), whether the
devnet's configurator step hands the two nodes different genesis config, or
something else.

Found via: the `node-parity` cross-node comparison tool built in PR #3806
(blinklabs-io/dingo#1900 phase 2), same run as blinklabs-io/dingo#3824.

Acceptance criteria: confirm whether cardano-node's protocol parameters for
this network genuinely have no plutusV1 cost model. If so, Dingo's
ShelleyCurrentProtocolParamsQuery reply omits costModels.plutusV1 under the
same condition, verified by TestLedgerStateConsensus (or `node-parity check`)
passing on protocol-parameter comparison against a real reference node.

arepala-uml:
## Summary

`GetCurrentProtocolParams` always reported a PlutusV2 cost model, even on a chain that has never received a real on-chain protocol-parameter update for it.

`ledger/eras/babbage.go`'s `HardForkBabbage` fabricates a default PlutusV2 cost model whenever the previous era's params don't have one. This is legitimate and necessary: real script *validation* needs something to price a genuine PlutusV2 script against during the historical gap between the live Babbage hard fork and the real, governance-approved cost model actually landing on-chain. On mainnet/preview/preprod that gap closes within a normal amount of time. On a short-lived devnet, it never closes at all — so Dingo keeps reporting the synthetic default forever, while a real cardano-node correctly reports nothing until real data exists.

This was originally surfaced by `node-parity` (#3806) comparing Dingo against a real cardano-node on a live devnet run, and filed as blinklabs-io/dingo#3825 based on that comparison's diff, which showed a confusing "plutusV1 vs plutusV2" label mismatch. That specific labeling turned out to be a **separate, unrelated bug**: an off-by-one Plutus-language-index mismatch in a `gouroboros` helper function used only by `node-parity`'s own display conversion, not present in the actual LocalStateQuery wire data. To get the real, root-cause picture, this session decoded the raw CBOR bytes both nodes actually send for `GetCurrentProtocolParams` directly with the real client-side type, bypassing that display bug entirely:

- Dingo's real reply: cost models at keys 0, 1, 2 (PlutusV1, PlutusV2, PlutusV3).
- cardano-node's real reply, same chain, same point: cost models at keys 0, 2 only — **no PlutusV2 at all**.

That's the real, root-cause divergence this PR fixes.

## The fix

Internal script validation is completely unaffected: `ls.currentPParams` keeps carrying the real default value unconditionally, exactly as before, since a genuine PlutusV2 script really can arrive before governance's real update lands, on any network. Only the `GetCurrentProtocolParams` LocalStateQuery reply changes, and only once: `LedgerState` now tracks whether the current PlutusV2 cost model is still Dingo's own fabricated default (`syntheticV2CostModel`), and the query handler omits it from the reply for as long as that's true.

- Set the moment `HardForkBabbage`'s default injection is detected (comparing before/after cost-model maps at the exact transition point).
- Never reset back to `true` by a later era's own transition carrying the same map forward (e.g. Babbage → Conway).
- Cleared the moment a real, on-chain governance-enacted protocol-parameter update actually sets a PlutusV2 cost model — regardless of whether the enacted value happens to equal the synthetic default's, since real governance re-affirming that exact value is still real data, not still a guess.

## File-by-file changes

- **`ledger/state.go`** — new `LedgerState.syntheticV2CostModel bool` field (and its mirror on the lock-free-read `consensusSnapshot`, published via `publishSnapshotsLocked`). New `injectedSyntheticV2CostModel` helper detects the fabrication at the exact transition point in `transitionToEraFrom`; `applyEraTransition` sets the flag (never clears it). `EraTransitionResult` and `EpochRolloverResult` each gain a field to carry this signal from where it's detected to where it's applied.
- **`ledger/chainsync.go`** — `processEpochRollover`'s governance-enactment handling detects when a real `PParamsChanged` update actually sets a PlutusV2 cost model, marking `EpochRolloverResult.RealV2CostModelObserved` so the caller can clear the synthetic flag.
- **`ledger/view.go`** — new `withoutSyntheticV2CostModel` (and `withoutCostModelKey`) helper: returns a shallow copy with the PlutusV2 entry removed when synthetic, or the value unchanged (identical pointer) otherwise. Never mutates the original, which stays reachable from internal validation state.
- **`ledger/queries.go`** — `ShelleyCurrentProtocolParamsQuery`'s handler now applies the filter before returning the reply, instead of returning the live internal struct directly.
- **`ledger/synthetic_v2_cost_model_test.go`** — new tests: detection logic (fires only for the exact known default value, not any newly-added key), the filter itself (removes the right key, never mutates the original), and two end-to-end tests through `ls.Query(...)` proving the reply omits the synthetic default when set and includes real data when cleared.

## Testing

- New tests adversarially verified: confirmed the key end-to-end test fails when the query handler is pointed back at the unfiltered value, passes with the fix.
- `go test ./ledger/... -race` — full suite passes (`ok` across every subpackage, including a full isolated rerun after ruling out an earlier timeout as resource contention from an unrelated concurrent process on the test machine, not a regression from this change).
- `golangci-lint run`, `nilaway`, `modernize` — clean on touched files (a few pre-existing findings elsewhere in `ledger/view.go` and `ledger/state.go` are unrelated to this diff, confirmed by line range).
- Live devnet conformance run (`./run-tests.sh --conformance`, `TestLedgerStateConsensus`) against a real `dingo-producer` + `cardano-node` pair, before and after the fix — the `protocol parameters differ` divergence this issue is about is gone. Evidence in a follow-up comment on this PR.

Closes #3825 

---
## Summary by cubic
`GetCurrentProtocolParams` always reported a PlutusV2 cost model, even on a chain that never received a real on-chain protocol-parameter update for it. Every protocol-parameter reporting surface (LocalStateQuery, Blockfrost, UTXORPC, Mesh) now omits that synthetic default until real governance data exists, matching what a real cardano-node reports (blinklabs-io/dingo#3825).

- Reporting surfaces all read through one filtered accessor (`GetCurrentPParamsForReporting`); internal callers like script validation keep the unfiltered fabricated default.
- Tracks a `syntheticV2CostModel` flag from the Babbage hard-fork transition, persisted in the same transaction as the pparams write and restored at startup; nodes upgraded onto this build fall back to a value comparison when the marker is absent.
- Clears the flag only when a real update explicitly specifies a PlutusV2 cost model, whether via Conway/Dijkstra governance or the pre-Conway Shelley-style update path, even if it re-writes the canonical default value.
- Marking is rollback-safe: a durable cleared-epoch marker records the earliest real write and is recomputed after rollback or truncate; later writes never overwrite it, and truncate now deletes the marker instead of forcing it, deferring to the absent-marker fallback on legacy databases.
- The filter covers Alonzo, Babbage, Conway, and Dijkstra era params, returns a shallow copy without mutating the live struct, and guards concrete-typed nil pointers in every era case.

<sup>Written for commit b5f8a6e91715fff1657e6fea460132571ab607b7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3884?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Protocol parameter queries no longer expose a fabricated PlutusV2 cost model before a real governance-provided model is available.
  - Synthetic cost model tracking now persists across restarts and hard-fork transitions.
  - Real governance updates correctly replace the synthetic model and appear in subsequent queries.


arepala-uml:
## Devnet conformance evidence

### 1. Root cause — raw wire bytes decoded directly, independent of any display bug

Full log: **`devnet_rawbytes.log`** (attached below)

Captured the actual CBOR bytes both nodes send for `GetCurrentProtocolParams`, decoded with the real client-side type (`ledger.ConwayProtocolParameters`), before any conversion/display layer runs:

```
DINGO:        CostModels keys = [0, 1, 2]   (PlutusV1, PlutusV2, PlutusV3)
CARDANO-NODE: CostModels keys = [0, 2]      (PlutusV1, PlutusV3 -- no PlutusV2)
```

Keys 0 and 2's values are byte-for-byte identical between the two nodes (same 166-value PlutusV1 model, same 251-value PlutusV3 model). The only real difference: Dingo reports a PlutusV2 cost model it fabricated internally (`HardForkBabbage`'s default); a real cardano-node, on the identical chain, genuinely has none configured.

(The original issue's "plutusV1 vs plutusV2" label confusion came from a separate, unrelated off-by-one bug in a `gouroboros` display-conversion helper `node-parity` uses — confirmed against `gouroboros`'s own real conway CDDL-derived indexing and not touched by this PR. This raw decode bypasses that entirely.)

### 2. Failing on `main` — the bug, reproduced live

Full log: **`devnet_before.log`** (attached below)

```
=== RUN   TestLedgerStateConsensus
    Messages:   ledger state diverged between dingo-producer and cardano-producer at slot 54 (block 21):
                protocol parameters differ:
                --- a ---
                { ... "costModels": { "plutusV1": {...}, "plutusV2": {...} ... } }
                --- b ---
                { ... "costModels": { "plutusV2": {...} ... } }
--- FAIL: TestLedgerStateConsensus (8.50s)
```

### 3. Adversarial unit-test verification

`TestQueryShelleyCurrentProtocolParams_OmitsSyntheticV2CostModel` and 6 sibling tests in `ledger/synthetic_v2_cost_model_test.go`:

```
--- PASS: TestInjectedSyntheticV2CostModel_DetectsHardForkBabbagesDefault (0.00s)
--- PASS: TestInjectedSyntheticV2CostModel_FalseWhenAlreadyPresent (0.00s)
--- PASS: TestInjectedSyntheticV2CostModel_FalseWhenValueIsNotTheKnownDefault (0.00s)
--- PASS: TestWithoutSyntheticV2CostModel_RemovesKeyWithoutMutatingOriginal (0.00s)
--- PASS: TestWithoutSyntheticV2CostModel_NoOpWhenNotSynthetic (0.00s)
--- PASS: TestQueryShelleyCurrentProtocolParams_OmitsSyntheticV2CostModel (0.13s)
--- PASS: TestQueryShelleyCurrentProtocolParams_IncludesRealV2CostModel (0.06s)
```

Confirmed the key test actually catches the regression -- pointed the query handler back at the unfiltered value and re-ran:

```
--- FAIL: TestQueryShelleyCurrentProtocolParams_OmitsSyntheticV2CostModel (0.29s)
    Error:  map[uint][]int64{0:[...], 1:[205665, 812, 1, ...], 2:[...]} should not contain 0x1
    Messages: the reply must omit the synthetic PlutusV2 cost model
```

### 4. Fixed — same live devnet test, this branch

Full log: **`devnet_after.log`** (attached below)

```
=== RUN   TestLedgerStateConsensus
    observer.go:237: observer dingo-producer: following chain from host.docker.internal:27723
    observer.go:237: observer cardano-producer: following chain from host.docker.internal:27724
    ...
        Messages:  ledger state diverged between dingo-producer and cardano-producer at slot 823 (block 264):
                   [stake distribution divergence -- unrelated, separate issue, tracked independently]
--- FAIL: TestLedgerStateConsensus (30.53s)
```

No `protocol parameters differ` line anywhere in the output -- the divergence this issue is about is gone. This particular run still fails overall, but only because it happened to run far enough into the chain (past several real epoch boundaries, for the first time in any run so far) to hit a completely separate, unrelated stake-distribution discrepancy that has nothing to do with protocol parameters or cost models. That's being tracked as its own separate investigation, not part of this PR.

### 5. Full suite

```
go test ./ledger/... -race
ok  	github.com/blinklabs-io/dingo/ledger	561.690s
ok  	github.com/blinklabs-io/dingo/ledger/eras
ok  	github.com/blinklabs-io/dingo/ledger/forging
ok  	github.com/blinklabs-io/dingo/ledger/governance
ok  	github.com/blinklabs-io/dingo/ledger/hardfork
ok  	github.com/blinklabs-io/dingo/ledger/leader
ok  	github.com/blinklabs-io/dingo/ledger/leios
ok  	github.com/blinklabs-io/dingo/ledger/rewards
ok  	github.com/blinklabs-io/dingo/ledger/snapshot
```

### Attached logs
- [devnet_before.log](https://github.com/user-attachments/files/31844340/devnet_before.log) - full run against unfixed `main`
- [devnet_rawbytes.log](https://github.com/user-attachments/files/31844365/devnet_rawbytes.log) - raw CBOR wire-byte capture from both nodes
- [devnet_after.log](https://github.com/user-attachments/files/31844370/devnet_after.log) - full run against this branch's fix

ledger/view.go
```diff
@@ -638,6 +638,67 @@ func extractRawCostModels(
...
+		return &modified
+	case *babbage.BabbageProtocolParameters:
+		modified := *p
+		modified.CostModels = withoutCostModelKey(p.CostModels, 1)
+		return &modified
+	case *conway.ConwayProtocolParameters:
+		modified := *p
+		modified.CostModels = withoutCostModelKey(p.CostModels, 1)
+		return &modified
+	default:
```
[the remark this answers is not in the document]
arepala-uml: Confirmed and fixed in 1c60a1ae. Added a `*dijkstra.DijkstraProtocolParameters` case to the switch, with a test (`TestWithoutSyntheticV2CostModel_CoversEveryEraType/Dijkstra`) verified adversarially -- confirmed it fails without the case, passes with it.

ledger/view.go
```diff
@@ -638,6 +638,67 @@ func extractRawCostModels(
...
+	pp lcommon.ProtocolParameters,
+	synthetic bool,
+) lcommon.ProtocolParameters {
+	if !synthetic {
+		return pp
+	}
+	switch p := pp.(type) {
+	case *alonzo.AlonzoProtocolParameters:
+		modified := *p
+		modified.CostModels = withoutCostModelKey(p.CostModels, 1)
```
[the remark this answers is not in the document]
arepala-uml: Checked this against the actual gouroboros types rather than assuming: `AlonzoProtocolParameters`, `BabbageProtocolParameters`, and `ConwayProtocolParameters` (gouroboros v0.202.7) only embed `cbor.StructAsArray` -- a stateless marker (`_ struct{} \`cbor:",toarray"\``), not `cbor.DecodeStoreCbor`. None of the three has a custom `MarshalCBOR` override either, so they always encode via the default field-reflection path and pick up the mutated `CostModels` map correctly.

The `cbor.DecodeStoreCbor` embedding that likely prompted this is on a different type in the same file -- `ConwayProtocolParameterUpdate` (the partial-update proposal type with pointer fields), not `ConwayProtocolParameters` (the full "current params" type this function actually receives and returns). No code change needed; the shallow copy is safe as written.

ledger/state.go
```diff
@@ -869,37 +872,52 @@ type LedgerState struct {
...
+	// true once cleared: once real data has been seen for this key, later
+	// eras carrying the same map forward must not be reinterpreted as
+	// synthetic again. See queryShelleyCurrentProtocolParams
+	// (blinklabs-io/dingo#3825) for why this exists: internal script
+	// validation must keep using the real default regardless (a genuine
+	// PlutusV2 script can arrive before the real update lands), but a
+	// LocalStateQuery caller asking "what are the current protocol
+	// parameters" should see only what the chain has actually committed to,
+	// matching what a real cardano-node reports during the same window.
+	syntheticV2CostModel        bool
```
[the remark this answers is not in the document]
arepala-uml: Confirmed and fixed in 1c60a1ae. `syntheticV2CostModel` is now persisted via the existing `SyncState` mechanism (`persistSyntheticV2CostModel`, written on every set/clear transition) and restored at startup (`loadSyntheticV2CostModel`, called right after `loadPParams()`). Adversarially verified: `TestSyntheticV2CostModelPersistence_RoundTripsAcrossRestart` fails when the persistence write is disabled, passes with it restored.

One documented, accepted limitation: a database that predates this field and has never received a real PlutusV2 update reads back "not synthetic" (the empty-value default), which is correct for the case that matters in practice (a real network that hard-forked into Babbage and received its real update long ago) and only wrong for a devnet/test network carried across a binary upgrade at exactly that moment -- not retroactively reconstructed from full transition history, noted in the function's doc comment.

ledger/chainsync.go
```diff
@@ -5967,6 +5967,15 @@ func (ls *LedgerState) processEpochRollover(
 				"persist post-enactment pparams: %w", err,
 			)
 		}
+		// A real, on-chain-enacted PParamUpdate touched protocol params this
+		// epoch. If it set a PlutusV2 cost model, that is real governance
+		// data from here on -- clear the synthetic-default marker regardless
+		// of whether the enacted value happens to equal the default's (real
+		// governance re-affirming the same value is still real, not still a
+		// guess). See LedgerState.syntheticV2CostModel.
+		if _, hasV2 := extractRawCostModels(newPParams)[1]; hasV2 {
```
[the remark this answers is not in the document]
arepala-uml: Confirmed and fixed in 1c60a1ae. Extracted the comparison into a testable `realV2CostModelWritten(preHadV2, pre, after)` (ledger/state.go) that compares against what the key held immediately *before* this specific enactment, not merely whether it's present after -- so an enactment that only bumps the protocol version or touches an unrelated field no longer clears the marker just because the synthetic default carried forward unchanged.

One tradeoff worth flagging: from this signal alone there's no way to distinguish "governance enacted a value equal to the default" from "this enactment never touched the cost model at all" (both leave the value unchanged) -- so an equal-value re-affirmation is conservatively left synthetic rather than risk clearing too early on the more common unrelated-enactment case. Documented on `EpochRolloverResult.RealV2CostModelObserved` and `realV2CostModelWritten`.

Adversarially verified: `TestRealV2CostModelWritten_FalseWhenUnrelatedEnactmentCarriesItForward` fails against the old (`hasV2`-only) logic, passes with the fix; three more tests cover the value-changed, newly-appears, and still-absent cases.

ledger/view.go
```diff
@@ -638,6 +638,67 @@ func extractRawCostModels(
...
+func withoutSyntheticV2CostModel(
+	pp lcommon.ProtocolParameters,
+	synthetic bool,
+) lcommon.ProtocolParameters {
+	if !synthetic {
+		return pp
+	}
+	switch p := pp.(type) {
+	case *alonzo.AlonzoProtocolParameters:
+		modified := *p
```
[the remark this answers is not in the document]
arepala-uml: Confirmed and fixed in 1c60a1ae. Added a nil check in every type-switch case before dereferencing. `TestWithoutSyntheticV2CostModel_NilPointerDoesNotPanic` adversarially verified: confirmed it actually panics without the guard (a concrete-typed nil `*conway.ConwayProtocolParameters` matches its case and `*p` dereferences a nil pointer), passes with it.

ledger/synthetic_v2_cost_model_test.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	assert.False(t, injectedSyntheticV2CostModel(before, after))
+}
+
+// TestWithoutSyntheticV2CostModel_RemovesKeyWithoutMutatingOriginal covers
+// the query-boundary filter: when synthetic is true, the returned value
+// omits PlutusV2 while every other key survives, and the original pparams
+// (still reachable from internal validation state) is never mutated.
+func TestWithoutSyntheticV2CostModel_RemovesKeyWithoutMutatingOriginal(t *testing.T) {
```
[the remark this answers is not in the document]
arepala-uml: Confirmed and fixed in 1c60a1ae. Added `TestWithoutSyntheticV2CostModel_CoversEveryEraType` with Alonzo, Babbage, and Dijkstra subtests alongside the existing Conway-only coverage, each asserting key 1 is removed from the filtered copy while the original is left untouched.

ledger/state.go
```diff
@@ -3865,6 +3914,114 @@ func (ls *LedgerState) applyEraTransition(result *EraTransitionResult) {
...
+// possible one-time regression back to exposing the synthetic default after
+// the next restart, not a correctness problem for the running process (which
+// already has the in-memory field set), so it is logged rather than
+// propagated as an error from the era-transition/rollover path calling it.
+func (ls *LedgerState) persistSyntheticV2CostModel(value bool) {
+	v := "false"
+	if value {
+		v = "true"
+	}
+	if err := ls.db.SetSyncState(syntheticV2CostModelSyncKey, v, nil); err != nil {
```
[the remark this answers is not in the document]
arepala-uml: Fixed in b67d8aac. `persistSyntheticV2CostModel` now takes the caller's `txn` and propagates errors instead of using a hardcoded `nil` txn with logged-and-swallowed errors. `transitionToEraFrom` calls it with the same `txn` as `SetPParams` right after that write succeeds (setting the marker true), and `processEpochRollover` does the same for the clearing write. In-memory state (`ls.syntheticV2CostModel`) is only ever updated afterward, once the caller has observed the transaction succeed.

Added `TestTransitionToEraFrom_PersistsSyntheticMarkerInSameTransactionAsPParams`, which proves atomicity directly: it runs the era transition inside a transaction, rolls the transaction back, and asserts the marker was NOT persisted (previously it would have been, since the old code wrote it via an independent `nil`-txn commit). Reverting the fix makes this test fail.

ledger/state.go
```diff
@@ -3865,6 +3914,114 @@ func (ls *LedgerState) applyEraTransition(result *EraTransitionResult) {
...
+func realV2CostModelWritten(
+	preHadV2 bool,
+	pre []int64,
+	after lcommon.ProtocolParameters,
+) bool {
+	post, postHadV2 := extractRawCostModels(after)[1]
+	if !postHadV2 {
+		return false
+	}
+	return !preHadV2 || !slices.Equal(pre, post)
```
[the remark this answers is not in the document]
arepala-uml: Fixed in b67d8aac, and this deserves the credit for catching a real gap: `DefaultPlutusV2CostModel` is the actual canonical mainnet PlutusV2 cost model value, so real governance re-affirming it verbatim was exactly the common case the old value-comparison approach would silently miss forever on a real network.

Replaced `realV2CostModelWritten` with explicit provenance tracked from the enacted delta itself: `governance.EnactmentResult` (ledger/governance/enact.go) now has a `PlutusV2CostModelWritten` field, set in both the Conway and Dijkstra `ConwayParameterChangeGovAction`/`DijkstraParameterChangeGovAction` cases via `if _, ok := a.ParamUpdate.CostModels[1]; ok { ... }` — checking whether the *enacted update* specified the key, independent of the value. This propagates through `governance.EpochOutput.PlutusV2CostModelWritten` (epoch.go), and `ledger/chainsync.go`'s `processEpochRollover` now sets `RealV2CostModelObserved` from `govOut.PlutusV2CostModelWritten` directly instead of comparing before/after values.

Added `TestEnactProposal_ConwayParameterChangeWritesPlutusV2CostModel` and the Dijkstra equivalent in `ledger/governance/enact_test.go`, both of which enact an update that writes `eras.DefaultPlutusV2CostModel` verbatim (the exact value already present) and assert `PlutusV2CostModelWritten` is still true — the case the old value-comparison logic got wrong. Reverting the fix makes both fail.

ledger/synthetic_v2_cost_model_test.go
```diff
@@ -0,0 +1,364 @@
...
+	t *testing.T,
+) {
+	db := newTestDB(t)
+	ls := newPoolDistr2Ledger(t, db)
+
+	prev := &alonzo.AlonzoProtocolParameters{
+		CostModels: map[uint][]int64{0: {1, 2, 3}},
+	}
+
+	txn := db.Transaction(true)
```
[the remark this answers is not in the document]
arepala-uml: Fixed in de0586b8. Added read-back assertions via `db.GetPParams(1, eras.BabbageEraDesc.Id, eras.DecodePParamsBabbage, nil)` after both the rollback branch (asserts nil — the write did not survive) and the commit branch (asserts non-nil — it did), so the test now actually verifies the pparams half of the atomicity claim the comment makes, not just the marker.

ledger/state.go
```diff
@@ -3865,6 +3923,94 @@ func (ls *LedgerState) applyEraTransition(result *EraTransitionResult) {
...
+	value, err := ls.db.GetSyncState(syntheticV2CostModelSyncKey, nil)
+	if err != nil {
+		ls.config.Logger.Warn(
+			"failed to read synthetic PlutusV2 cost model marker from database",
+			"component", "ledger",
+			"error", err,
+		)
+		return
+	}
+	ls.syntheticV2CostModel = value == "true"
```
wolf31o2: The marker is not rollback-safe, in memory or on disk.

`ls.syntheticV2CostModel` is assigned at exactly four places — `state.go:3939` (set), this line (load), `state.go:5454` and `state.go:7758` (clear). `rollback` (`state.go:2980-3410`) recomputes `ls.currentPParams` from the truncated database and republishes the snapshot without touching it. `database/truncate.go` `TruncateAfterSlot` touches no `sync_state` key either, so the durable marker survives a truncate to genesis.

A rollback past the enactment that cleared the flag restores the synthetic default into `currentPParams` while the flag stays `false`, so `GetCurrentProtocolParams` reports the fabricated model again — #3825 back, and permanently after a restart. It also survives a re-sync onto a fork that never contains the enactment.

The precedent for this class is already in the tree: `database/account_expiry_truncate.go:135-152` calls `DeleteSyncState(DelegatorInactivityActivatedSyncKey, txn)` when a rollback crosses back before the boundary that set the CIP-0163 marker. Same treatment is needed in `TruncateAfterSlot`, plus a re-derivation in `rollback`.

No test covers this in either direction.

ledger/chainsync.go
```diff
@@ -5967,6 +5967,19 @@ func (ls *LedgerState) processEpochRollover(
...
+			// writing it separately, after this transaction commits, would
+			// leave a window where a crash after the pparams commit but
+			// before the marker commit strands the marker stale, exposing
+			// the wrong thing on restart. See
+			// LedgerState.syntheticV2CostModel and blinklabs-io/dingo#3825's
+			// PR review.
+			if err := ls.persistSyntheticV2CostModel(false, txn); err != nil {
+				return nil, err
+			}
+			result.RealV2CostModelObserved = true
```
wolf31o2, line 5991: This is the only source of `RealV2CostModelObserved`, and it originates only in `ledger/governance/enact.go:104` and `:115` — the Conway and Dijkstra `ParameterChangeGovAction` cases. The pre-Conway path, `ls.db.ComputeAndApplyPParamUpdates(...)` at `chainsync.go:5879`, has no equivalent signal, but it can set a PlutusV2 cost model: `gouroboros/ledger/alonzo/pparams.go:157-161` and `babbage/pparams.go:126-130` both do `maps.Copy(p.CostModels, paramUpdate.CostModels)`.

So on a network that forks into Babbage without a PlutusV2 cost model (flag set at `state.go:3939`) and later receives one through an Alonzo/Babbage-era update proposal, the flag stays `true` forever and `withoutV2CostModelKey` deletes key 1 unconditionally. Dingo would then omit a real, on-chain, governance-approved cost model — a new divergence in the opposite direction from the one being fixed, and permanent.

Mainnet-shaped histories are safe: `splitEraTransitionsForRollover` (`chainsync.go:6160`) returns `nil, transitionPath`, so every era transition runs after `ComputeAndApplyPParamUpdates`; an HF proposal carrying the V2 cost models alongside `protocolVersion` lands first and `HardForkBabbage`'s `if _, hasV2 := ret.CostModels[1]; !hasV2` guard never fires. That holds only while the cost model arrives at or before the boundary.

Either give `ComputeAndApplyPParamUpdates` the same provenance signal, or have `withoutV2CostModelKey` re-check `slices.Equal(m[1], eras.DefaultPlutusV2CostModel)` before deleting so the filter cannot strip a value it never fabricated.

The clearing chain is also untested end to end: `RealV2CostModelObserved` and `EpochOutput.PlutusV2CostModelWritten` appear in no test, and the assignments at `state.go:5454` and `:7758` are never exercised — the end-to-end tests set the field by hand (`synthetic_v2_cost_model_test.go:263`). Nothing proves `EnactmentResult` -> `EpochOutput` -> `EpochRolloverResult` -> `ls.syntheticV2CostModel` connects.

ledger/view.go
```diff
@@ -638,6 +639,87 @@ func extractRawCostModels(
...
+		modified.CostModels = withoutV2CostModelKey(p.CostModels)
+		return &modified
+	case *dijkstra.DijkstraProtocolParameters:
+		if p == nil {
+			return pp
+		}
+		modified := *p
+		modified.CostModels = withoutV2CostModelKey(p.CostModels)
+		return &modified
+	default:
```
wolf31o2, line 726: Two smaller gaps here.

`extractRawCostModels` (`view.go:619-638`) returns `nil` for `*dijkstra.DijkstraProtocolParameters`, and the `costModelsProvider` interface it prefers is implemented by no type in the pinned gouroboros. `withoutSyntheticV2CostModel` does handle Dijkstra, so the detector is structurally blind to an era the filter covers. Not live today, but the two should share one extraction.

The `default:` branch returns unfiltered with no log and no error, so a future era pparams type silently reintroduces #3825. Untested.

Also: `TestWithoutSyntheticV2CostModel_CoversEveryEraType` says it covers "Alonzo, Babbage, Conway, Dijkstra" but has no Conway subtest.

ledger/state.go
```diff
@@ -3865,6 +3923,94 @@ func (ls *LedgerState) applyEraTransition(result *EraTransitionResult) {
...
+// Not retroactively reconstructed by scanning full transition history.
+func (ls *LedgerState) loadSyntheticV2CostModel() {
+	value, err := ls.db.GetSyncState(syntheticV2CostModelSyncKey, nil)
+	if err != nil {
+		ls.config.Logger.Warn(
+			"failed to read synthetic PlutusV2 cost model marker from database",
+			"component", "ledger",
+			"error", err,
+		)
+		return
```
wolf31o2, line 4041: Worth being explicit about the consequence for the network this targets. A pre-existing database loads `false` here, and the flag can only be set again at a live era transition (`state.go:3939`), which a devnet already past the Babbage fork will never perform. So an existing devnet database that upgrades to this build keeps the pre-PR reply indefinitely, and conformance evidence gathered from a fresh database will not reproduce on an upgraded one.

ledger/synthetic_v2_cost_model_test.go
```diff
@@ -0,0 +1,376 @@
...
+	ls := newPoolDistr2Ledger(t, newTestDB(t))
+	ls.currentEra = eras.ConwayEraDesc
+	ls.currentPParams = &conway.ConwayProtocolParameters{
+		CostModels: map[uint][]int64{
+			0: {1, 1, 1},
+			1: eras.DefaultPlutusV2CostModel,
+			2: {3, 3, 3},
+		},
+	}
+	ls.syntheticV2CostModel = false
```
wolf31o2, line 412: The end-to-end tests type-assert `arr[0].(*conway.ConwayProtocolParameters)` and never encode, but the issue is about the bytes on the wire. Their fixture pparams are in fact not encodable as written — `cbor.Rat.MarshalCBOR` panics on nil until `A0`/`Rho`/`Tau`/`ExecutionCosts`/`MinFeeRefScriptCostPerByte`/`PoolVotingThresholds`/`DRepVotingThresholds` are populated. Adding a `cbor.Encode` plus decode assertion would pin the thing the divergence was measured on.

Other missing cases: the clearing chain end to end (see the `chainsync.go` comment), a rollback case in both directions, a `ComputeAndApplyPParamUpdates` update that sets `CostModels[1]`, the `default:`/unknown-era branch of the filter, `loadSyntheticV2CostModel` with a `GetSyncState` error or a non-`"true"` value, and a Conway negative case for `PlutusV2CostModelWritten` (the `require.False` exists only in the Dijkstra test, `enact_test.go:105`).

The tests that are here can fail — the query tests drive the real `ls.Query` dispatch and the governance tests round-trip real CBOR, including the enact-exactly-the-default case.

wolf31o2 (commented):
The filter itself is correct and correctly scoped: `withoutSyntheticV2CostModel` has one call site (`ledger/queries.go:488`), every pricing and validation path reads `currentPParams` unfiltered, and the copy never mutates the shared cost-model map. Dingo derives no language view from pparams, so no script-data hash can move. The omit-the-key wire shape matches `flattenCostModels` (`libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/CostModels.hs:445,470`), where an absent language is an absent key. No verdict can flip.

Two state-machine gaps to settle before merge, both inline. Neither has a consensus effect, but one makes the node strip a real cost model.

One more, not on a changed line: `api/blockfrost/adapter.go:669` reads `GetCurrentPParams()` unfiltered, so `/epochs/latest/parameters` will report `cost_models.PlutusV2` while LSQ omits it — two of Dingo's own APIs disagreeing about whether the chain has one.

`go build ./...`, `go vet ./ledger/...`, `go test ./ledger/ ./ledger/governance/` and `go test ./internal/test/conformance/` pass at de0586b8.

ledger/state.go
arepala-uml: Fixed in cae9613b. Added a durable cleared-epoch provenance marker (`database.SyntheticV2CostModelClearedEpochSyncKey`) and `database.RecomputeSyntheticV2CostModelMarkerAfterTruncate`, mirroring the CIP-0163 delegator-inactivity-activation pattern exactly (`DelegatorInactivityActivationEpoch`/`RecomputeAccountExpirationsAfterTruncate`): it resolves the rollback slot to an epoch via `EpochBySlot`, and if the recorded cleared-epoch is after the rollback point, deletes the marker and restores the boolean to `"true"`. Wired into both `ledger.LedgerState.rollback` (same transaction as `TruncateAfterSlot`) and `database/lifecycle.Truncate`. Added `database/synthetic_cost_model_test.go` covering both directions (crosses back / doesn't cross back) plus the never-cleared no-op case, and adversarially verified (temporarily disabled the reset, confirmed the test fails). Documented the two sync_state keys in DATABASE.md and the design in ARCHITECTURE.md next to the CIP-0163 section.

ledger/chainsync.go
arepala-uml: Fixed in cae9613b. Added `EraDesc.ParamUpdateHasPlutusV2CostModelFunc` (wired per era in `ledger/eras/{alonzo,babbage,conway,dijkstra}.go`, checking the decoded update's own `CostModels[1]` presence — same principle as `governance.EnactmentResult.PlutusV2CostModelWritten`, not a before/after value comparison) and a new bool return from `database.ComputeAndApplyPParamUpdates`. `chainsync.go`'s `processEpochRollover` now calls `markRealV2CostModelObserved` from that path too, alongside the existing governance-enactment branch. Added `TestComputeAndApplyPParamUpdates_ReportsPlutusV2CostModelWritten`/`_FalseWhenUpdateDoesNotWritePlutusV2CostModel` in `database/pparams_test.go`.

For the untested clearing chain: added `TestProcessEpochEnactsConwayParameterChangeReportsPlutusV2CostModelWritten` in `ledger/governance/epoch_test.go`, which drives a real ratify-then-enact cycle through `ProcessEpoch` (not `EnactProposal` called in isolation) and asserts the resulting `EpochOutput.PlutusV2CostModelWritten` — closing the specific untested hop (`EnactmentResult` → `EpochOutput` via `applyEnactmentResult`'s OR). Adversarially verified by disabling that OR and confirming the test fails. I did not additionally build a full ledger-level `processEpochRollover` harness to test the last hop (`EpochOutput` → `EpochRolloverResult.RealV2CostModelObserved` → `ls.syntheticV2CostModel`) — that remaining link is 3 lines of mechanical field propagation in `chainsync.go`/`state.go`, read carefully and unchanged in shape from what already exists, and building that harness (real committee/DRep/governance-proposal fixture at the ledger level) looked like a large addition for one link already covered end-to-end at both ends. Happy to add it if you'd still like it as a follow-up.

ledger/view.go
arepala-uml: Fixed in cae9613b. Added the missing `*dijkstra.DijkstraProtocolParameters` case to `extractRawCostModels` (`TestExtractRawCostModels_CoversDijkstra`). `withoutSyntheticV2CostModel` now takes a `*slog.Logger` and logs a warning when `synthetic` is true but it hits the default branch (`TestWithoutSyntheticV2CostModel_UnknownTypeLogsAndReturnsUnfiltered`, using an unrecognized-type stub and capturing the log output). Also added the missing Conway subtest to `TestWithoutSyntheticV2CostModel_CoversEveryEraType` so its docstring's claim actually matches its subtests.

ledger/state.go
arepala-uml: Fixed in cae9613b. `loadSyntheticV2CostModel` now falls back to `resolveSyntheticV2CostModel`, which compares the current PlutusV2 cost model directly against `eras.DefaultPlutusV2CostModel` when the durable marker is absent, instead of defaulting to false. This is exactly the case you're describing: an existing devnet (or any node) already past the Babbage fork with no marker yet written now bootstraps correctly from its actual current value rather than silently staying in the pre-PR state forever. Covered by `TestResolveSyntheticV2CostModel_BootstrapsFromValueWhenMarkerAbsent`/`_ExplicitValueWins`. Also documented explicitly in DATABASE.md's `sync_state` row and the new ARCHITECTURE.md section.

ledger/synthetic_v2_cost_model_test.go
arepala-uml: Addressed the specific items in cae9613b:
- CBOR round trip: the end-to-end query tests' fixture now goes through a new `conwayPParamsWithCostModels` helper that populates every cbor.Rat-bearing field (A0/Rho/Tau/ExecutionCosts/MinFeeRefScriptCostPerByte/PoolVotingThresholds/DRepVotingThresholds), and both tests now do a real `cbor.Encode`+`cbor.Decode` round trip on the reply, not just a type assertion.
- Rollback in both directions: `database/synthetic_cost_model_test.go` (see the reply on the rollback-safety thread).
- `ComputeAndApplyPParamUpdates` real update: `TestComputeAndApplyPParamUpdates_ReportsPlutusV2CostModelWritten`/`_FalseWhenUpdateDoesNotWritePlutusV2CostModel`.
- Default/unknown-era branch: `TestWithoutSyntheticV2CostModel_UnknownTypeLogsAndReturnsUnfiltered`.
- Conway negative case for `PlutusV2CostModelWritten`: `TestEnactProposal_ConwayParameterChangeDoesNotWritePlutusV2CostModel` in `ledger/governance/enact_test.go`.
- Clearing chain end-to-end: `TestProcessEpochEnactsConwayParameterChangeReportsPlutusV2CostModelWritten` (see the reply on the pre-Conway-path thread for what this does and doesn't cover).

Not added: a `loadSyntheticV2CostModel`-with-`GetSyncState`-error test. That branch is a defensive log-and-continue path identical in shape to several other DB-read-error branches already in this file (e.g. the epoch/pparams reload warnings in `rollback`), and forcing a real DB I/O error without fragile double-close gymnastics against the shared test-DB teardown didn't seem worth it for that one line. Happy to add if you think it's worth it.

arepala-uml:
Addressed all 5 review findings in cae9613b. Summary (full detail in the individual thread replies):

1. **Rollback safety**: added a durable cleared-epoch provenance marker (`database.SyntheticV2CostModelClearedEpochSyncKey`) and `database.RecomputeSyntheticV2CostModelMarkerAfterTruncate`, mirroring the CIP-0163 delegator-inactivity-activation pattern. Wired into both `ledger.LedgerState.rollback` and `database/lifecycle.Truncate`.
2. **Pre-Conway provenance gap**: added `EraDesc.ParamUpdateHasPlutusV2CostModelFunc` (wired per era) and a new bool return from `database.ComputeAndApplyPParamUpdates`, giving the classic Shelley-style update path the same real-write signal Conway/Dijkstra governance already had.
3. **Filter gaps**: added the missing Dijkstra case to `extractRawCostModels`; `withoutSyntheticV2CostModel` now logs when it hits an unrecognized type instead of silently returning unfiltered data; added the missing Conway subtest.
4. **Existing-database migration gap**: `loadSyntheticV2CostModel` now falls back to comparing the current value against the known synthetic default when the durable marker is absent, instead of assuming "not synthetic."
5. **Test coverage**: added rollback tests (both directions), `ComputeAndApplyPParamUpdates` provenance tests, the filter's unknown-type branch, a Conway negative case for `PlutusV2CostModelWritten`, a real CBOR round trip for the end-to-end query tests, and a `ProcessEpoch`-level test proving the `EnactmentResult` → `EpochOutput` link connects through a real ratify+enact cycle.

Also updated DATABASE.md (the two new `sync_state` keys) and added an ARCHITECTURE.md section next to the CIP-0163 one this mirrors.

Verified: `go build ./...`, `go vet ./...`, full `go test ./...` (one unrelated pre-existing flake in `bin/` confirmed to pass in isolation), `go test ./ledger/... ./database/... -race`, `golangci-lint`/`nilaway`/`modernize`/`golines` clean on touched files, `make docs-parity`. Each new/changed test was adversarially verified (temporarily reverted the corresponding fix, confirmed the test fails, restored, confirmed it passes).

ledger/state.go
```diff
@@ -3028,6 +3065,20 @@ func (ls *LedgerState) rollback(point ocommon.Point) error {
 				err,
 			)
 		}
+		// Undo the synthetic-PlutusV2-cost-model marker if this rollback
+		// crosses back before the epoch it was last confirmed cleared at;
+		// see database.RecomputeSyntheticV2CostModelMarkerAfterTruncate and
+		// blinklabs-io/dingo#3825's PR review (wolf31o2).
+		if err := database.RecomputeSyntheticV2CostModelMarkerAfterTruncate(
```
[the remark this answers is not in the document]
arepala-uml: Confirmed real, fixed in 2028f000. `markRealV2CostModelObserved` now uses first-confirmation-wins semantics: it only writes the cleared-epoch marker if none is recorded yet, so a chain with multiple real PlutusV2 cost-model updates keeps the *earliest* confirmed epoch rather than the latest. A rollback crossing back past only the latest update (but not an earlier one) now correctly leaves the marker as real, since the earlier confirmation still applies. Added `TestMarkRealV2CostModelObserved_KeepsEarliestConfirmationAcrossMultipleUpdates` and adversarially verified (reverted the fix, confirmed the test fails with the marker landing at epoch 10 instead of 5).

ledger/view.go
```diff
@@ -633,9 +635,106 @@ func extractRawCostModels(
 		return p.CostModels
 	case *conway.ConwayProtocolParameters:
 		return p.CostModels
+	case *dijkstra.DijkstraProtocolParameters:
+		return p.CostModels
```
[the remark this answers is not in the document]
arepala-uml: Confirmed real, fixed in 2028f000 — took the suggestion. Audited the class rather than just this line: the same defect (dereferencing the type-switched pointer directly) was already present in the pre-existing Alonzo/Babbage/Conway cases too, not just the new Dijkstra one, so all four now guard `p == nil`, matching `withoutSyntheticV2CostModel`'s existing guards. Added `TestExtractRawCostModels_NilPointerDoesNotPanic` (table-driven over all four types) and adversarially verified.

database/synthetic_cost_model.go
```diff
@@ -0,0 +1,160 @@
...
+	// boolean it drove.
+	if err := d.DeleteSyncState(
+		SyntheticV2CostModelClearedEpochSyncKey, txn,
+	); err != nil {
+		return fmt.Errorf(
+			"clear synthetic PlutusV2 cost model cleared-epoch marker: %w",
+			err,
+		)
+	}
+	if err := d.SetSyncState(
```
wolf31o2: Non-blocking. The restore forces the boolean marker to `"true"` rather than deleting it, so it asserts "synthetic" for a chain that may never have carried a fabricated model.

Reachable on a database that predates these markers. `resolveSyntheticV2CostModel` reads an absent marker and compares the live PlutusV2 model against `eras.DefaultPlutusV2CostModel` (the 175-entry Preview epoch-3 vector, `ledger/eras/babbage.go:105`); on a network whose real model differs, that correctly yields false. The first later update specifying `CostModels[1]` still records `clearedEpoch = E`, since `markRealV2CostModelObserved` does not require the flag to currently be true, even though the real model has been in force since long before E. A rollback crossing the `E-1 -> E` boundary then sets this key to `"true"`, and `GetCurrentProtocolParams` drops a real, on-chain PlutusV2 cost model. Re-crossing the boundary re-clears it, so the ordinary shallow rollback self-heals; a fork switch that drops that enactment leaves it wrong until the next cost-model write.

`DeleteSyncState(SyntheticV2CostModelSyncKey, txn)` is correct in both directions here and reuses the fallback this PR already added: an absent marker re-derives from the value, giving `true` for the fabricated default and `false` for a real non-default model. `TestRecomputeSyntheticV2CostModelMarkerAfterTruncate_UndoesClearWhenRollbackCrossesBack` asserts `"true"` and would change with it.

ledger/queries.go
```diff
@@ -483,7 +483,14 @@ func (ls *LedgerState) queryShelleyLeaf(query any) (any, error) {
 	case *olocalstatequery.ShelleyEpochNoQuery:
 		return []any{ls.loadConsensusSnapshot().currentEpoch.EpochId}, nil
 	case *olocalstatequery.ShelleyCurrentProtocolParamsQuery:
-		return []any{ls.loadConsensusSnapshot().currentPParams}, nil
+		snapshot := ls.loadConsensusSnapshot()
+		return []any{
+			withoutSyntheticV2CostModel(
+				snapshot.currentPParams,
```
wolf31o2: Non-blocking. Restating the one item from the de0586b8 review that was not picked up, with the rest of its class: `withoutSyntheticV2CostModel` has this single call site, and every other interface reporting current parameters reads `GetCurrentPParams()` unfiltered — `api/blockfrost/adapter.go:669` (`cost_models` on `/epochs/latest/parameters`), `api/utxorpc/query.go:259` and `api/mesh/construction.go:322` (both via `ProtocolParameters.Utxorpc()`, which converts `CostModels`). On the chain this PR measured, those three report a `plutusV2` entry while LSQ omits it.

ledger/state.go
```diff
@@ -869,37 +872,52 @@ type LedgerState struct {
...
+	// syntheticV2CostModel is true from the moment HardForkBabbage fabricates
+	// a PlutusV2 cost model (real mainnet/preview/preprod never had one in
+	// genesis -- PlutusV2 postdates the Alonzo genesis format entirely, so
+	// this always fires on the live hard fork) until a real governance
+	// enactment sets one via processEpochRollover. It is never reset back to
+	// true once cleared: once real data has been seen for this key, later
+	// eras carrying the same map forward must not be reinterpreted as
+	// synthetic again. See queryShelleyCurrentProtocolParams
+	// (blinklabs-io/dingo#3825) for why this exists: internal script
+	// validation must keep using the real default regardless (a genuine
```
wolf31o2, line 892: Non-blocking, and a follow-up rather than a change to this PR: the premise that internal validation needs the fabricated model appears inverted. cardano-ledger rejects a transaction whose script language has no cost model (`NoCostModel`, raised from `collectPlutusScriptsWithContext`), so during the pre-update gap a real cardano-node treats a PlutusV2 script as invalid, while Dingo prices it with the fabricated model — `ledger/eras/babbage.go:397` and `ledger/eras/conway.go:1188` pass `CostModels[1]` straight into `cek.NewEvalContext` — and can accept a transaction the network rejects. That behavior is on `main` today and unchanged here, but it is the consensus-affecting half of the same fabrication and deserves its own issue. Not verified against a live node in this review.

wolf31o2 (dismissed):

database/synthetic_cost_model.go
arepala-uml: Fixed in b5f8a6e9. RecomputeSyntheticV2CostModelMarkerAfterTruncate now deletes SyntheticV2CostModelSyncKey instead of forcing it to "true", so a subsequent read falls back to resolveSyntheticV2CostModel's existing absent-marker heuristic (compare the live value against the known default) instead of a hardcoded wrong answer. Added TestRollbackRestore_LeavesRealPreExistingModelCorrectlyResolvedAsNotSynthetic covering exactly the scenario you described (a real, non-default model that predates these markers, with a clearedEpoch picked up later), and updated TestRecomputeSyntheticV2CostModelMarkerAfterTruncate_UndoesClearWhenRollbackCrossesBack's assertion accordingly. Adversarially verified.

ledger/queries.go
arepala-uml: Fixed in b5f8a6e9. Added LedgerState.GetCurrentPParamsForReporting as the one shared, filtered accessor and switched all three surfaces to it: Blockfrost's CurrentProtocolParams (api/blockfrost/adapter.go), UTXORPC's ReadParams (api/utxorpc/query.go), and Mesh's construction-metadata handler (api/mesh/construction.go). Every other GetCurrentPParams caller (script validation, block-building limits, Leios committee parameters, governance-action decoding) is untouched -- those need the real fabricated default unconditionally. Added TestGetCurrentPParamsForReporting_OmitsSyntheticV2CostModel/_IncludesRealV2CostModel proving the accessor's filtering behavior directly, since none of the three API packages can reach LedgerState's private fields to test this themselves.

ledger/state.go
arepala-uml: Agreed this shouldn't be a guessed change inside this PR. Filed as blinklabs-io/dingo#3962, capturing the claim, what's already confirmed in this codebase (no existing cost-model-absence rejection logic to extend -- HardForkBabbage's fabrication is specifically what prevents CostModels[1] from ever being genuinely absent, so this would be new validation logic, not a fix to existing logic), and what still needs verification against real cardano-ledger source and a live devnet run before any fix is designed. Not attempting a consensus-behavior change here without that verification -- the risk of a wrong guess (new divergence in the other direction) seems worse than leaving this tracked separately, and matches your own "not verified against a live node" framing.

arepala-uml:
Update:

1. Resolved the merge conflict against main (\`e3bcf0af\`) — only \`ledger/state.go\` conflicted (main renamed \`epochNonceHexCache\`'s value type), kept both changes.
2. Fixed the two remaining Cubic findings from the last pass (multi-update rollback provenance bug, nil-pointer panic across all four era types) — \`2028f000\`.
3. Addressed the three non-blocking findings from the approved review:
   - \`database/synthetic_cost_model.go\`: restore now deletes the marker instead of forcing it to \"true\", so a legacy database's real (non-default) model isn't misreported as synthetic — \`b5f8a6e9\`.
   - \`ledger/queries.go\` and friends: added \`LedgerState.GetCurrentPParamsForReporting\` and switched Blockfrost/UTXORPC/Mesh to it, so every reporting surface (not just LocalStateQuery) omits the synthetic model — \`b5f8a6e9\`.
   - \`ledger/state.go\` (inverted premise about internal validation needing the fabricated model): filed as blinklabs-io/dingo#3962 rather than guessing at a consensus-behavior change without live-node verification — see that issue and the thread reply for the reasoning.

All builds/vet/lint/race clean, each fix adversarially verified. Re-requested review from @wolf31o2.

wolf31o2 (approved):

PR #3888, opened 2026-09-04
fix(koiosparity): fail closed on boundary lookup errors

wolf31o2:
Koios parity now fails closed when tip or applying-epoch boundary lookups return database errors. Only a genuinely missing applying epoch is classified as pending; incomplete boundary metadata remains strict.

Validation:

- Targeted Koios parity tests, race-targeted tests, vet, golangci-lint, gofmt, import-boundaries, and docs-parity passed.
- Full `go test ./internal/koiosparity` is limited by the sandbox: an unrelated `httptest` case cannot listen on `tcp6 [::1]:0` (`operation not permitted`).

---
## Summary by cubic
Fixes Koios parity so boundary lookup errors fail closed. Previously, database errors during tip and applying-epoch lookups were treated as evidence that rewards were pending, downgrading real divergences to reference-lag; now those errors propagate, and only a genuinely missing applying epoch is classed as pending. Incomplete chain tips (missing hash or zero slot) also fail closed rather than downgrading.

**Bug Fixes**
- Adds `RewardsPending` to `DingoPoolEpochData`; comparison uses it over the wall-clock grace window, including during replay.
- Both `DingoDB` and `DatabaseSource` require a complete tip and treat tip or applying-epoch lookup failures as errors, so incomplete boundary metadata never hides a divergence.
- Adds tests that lookup errors propagate, incomplete tips stay strict, and pending classification happens only for a missing applying epoch.

Validation:
- Targeted Koios parity tests, race-targeted tests, vet, golangci-lint, gofmt, import-boundaries, and docs-parity passed.
- Full `go test ./internal/koiosparity` is limited by the sandbox: an unrelated `httptest` case cannot listen on `tcp6 [::1]:0` (`operation not permitted`).

<sup>Written for commit 7f0228626106abc1090106440a599a2481b7413d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3888?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/koiosparity/compare.go
+++ b/internal/koiosparity/compare.go
@@ -543,6 +543,9 @@ func ComparePoolEpoch(
 	// the grace window it may simply not be computed yet (reference_lag,
 	// ERROR); past it, it's a genuine gap in Dingo's own computation
 	// (dingo_db_missing, ERROR). Neither case can produce a PASS.
+	// RewardsPending is the chain-position form of the same timing check and
+	// takes precedence over the wall-clock grace window, including during
+	// replay where the epoch's close time may be years in the past.
 	if koiosPool.MemberRewards != "" {
 		switch {
 		case !dingoPool.MemberRewardPresent,
@@ -558,7 +561,9 @@ func ComparePoolEpoch(
 			// is a genuine gap in what Dingo can answer (dingo_db_missing,
 			// ERROR).
 			cat := CategoryDBMissing
-			if graceHours > 0 && !epochEndTime.IsZero() &&
+			if dingoPool.RewardsPending {
+				cat = CategoryReferenceLag
+			} else if graceHours > 0 && !epochEndTime.IsZero() &&
 				now.Sub(epochEndTime) < time.Duration(graceHours)*time.Hour {
 				cat = CategoryReferenceLag
 			}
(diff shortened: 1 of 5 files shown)


arepala-uml (approved):
Reviewed by checking out the PR head and validating independently (not just reading the diff): build clean, all 6 new/changed tests pass with `-race`, and traced every fact the new logic depends on against the real schema/writers (`tip` singleton via `SetTip`, `reward_pool_output.boundary_slot`, `epoch.start_slot`, `GetEpoch`/`GetTip` not-found semantics, the `E+3` boundary convention already documented in `fetch.go`).

The `RewardsPending` precedence over the wall-clock grace window in `compare.go` is correct and pinned by `TestComparePoolEpochUsesRewardsPending`; the per-pool `boundary_slot` check and the epoch-wide E+3 fallback are consistent between `DingoDB` and `DatabaseSource`.

Non-blocking: `source.go`'s applying-epoch lookup (via `meta.GetEpoch`) doesn't guard against a NULL `start_slot` on the epoch row the way `dingo_db.go`'s raw-SQL path does (`invalid start slot` error) — I reproduced the divergence with a probe test (DingoDB errors, DatabaseSource silently returns `RewardsPending=false`). Confirmed unreachable today since `SetEpoch` is the only writer of that table and always sets `start_slot` non-null, so not a blocker — worth a follow-up guard or a code comment noting the assumption.

No behavioral or security defects found. Docs (`DATABASE.md`/`ARCHITECTURE.md`) checked and not affected — no schema, EventBus, or component-boundary changes.

PR #3855, opened 2026-09-04
fix(database): attribute pointer-address UTxOs to their stake credential

chrisguiney:
Closes blinklabs-io/dingo#3854

A pointer address (types 4 and 5) carries the *position* of a stake
registration certificate, not a credential. gouroboros reports an empty
`StakeKeyHash` for one — correctly, since resolving the pointer is the ledger's
job — and Dingo was not doing that job, so the output was stored with a NULL
`staking_key` and its value never reached the stake distribution.

That understates the delegated stake of any account holding funds at a pointer
address, which tightens the producing pool's Praos leader threshold and makes
the node reject blocks the network accepted.

## Reproduced on preview

A from-genesis replay wedges at slot 9609386 in epoch 111, rejecting a canonical
block on the VRF leader threshold with `threshold_margin: -0.020778526`.
Comparing Dingo's `mark(110)` against Koios for epoch 111:

```
98 pools, 97 matched exactly
DIVERGENT: 1
  cb8ccb08136245d4  dingo=1455876171990  koios=1491430988669  delta=35554816679
```

One account explains almost all of it. It holds funds at two addresses with the
same payment credential — one base, one pointer — and only the base address was
counted. The unattributed pointer UTxOs total `35553515656` of the
`35554816679` pool-level divergence; the remaining `1301023` lovelace is not
accounted for by this address and has not been traced.

## The fix

Resolution happens **when stake is computed**, not when the output is ingested,
because that is what the ledger does: `resolveShelleyInstantStake` looks the
`Ptr` up in `saPtrs` at snapshot time, and `saPtrs` is mutable state. Three
consequences follow that an ingest-time resolution cannot express:

- **Conway and Dijkstra do not count pointer stake at all.**
  `ConwayInstantStake` has only a credential map; a `StakeRefPtr` output falls
  through its accumulator and is dropped. The Babbage→Conway translation
  rebuilds the instant stake from `sisCredentialStake` alone and drops `saPtrs`,
  so the pointer index itself is gone. This applies to *every* pointer output,
  including one produced in Babbage and still unspent after the fork — which is
  the case every network is in today. A gate applied at ingest cannot reach
  those outputs, so gating at ingest would have left the Conway inflation in
  place for exactly the outputs that matter.
- **A pointer may name a position no certificate occupies yet.** Nothing in any
  era validates an address's pointer payload, so the stake starts counting once
  the registration lands.
- **De-registration removes the `Ptr`** (`removePtr`), permanently. A later
[...]

PR #3890, opened 2026-09-04
fix(ledger): attribute pointer-address UTxOs

### LINKED ISSUE #3854: ledger: pointer-address UTxOs are never attributed to their stake credential, understating the stake distribution
## Summary

UTxOs held at **pointer addresses** (address types 4 and 5) are stored with a
NULL `staking_key`, so their value never reaches the stake credential the
pointer designates. The stake distribution therefore understates any account
holding funds at a pointer address, which shifts that pool's Praos leader
threshold and makes Dingo reject blocks the network accepted.

This is the root cause of the epoch-111 wedge in #3811.

## Evidence

A from-genesis Preview replay wedges at slot 9609386 rejecting a canonical block
on the VRF leader threshold. Cross-checking Dingo's `mark(110)` against Koios
for epoch 111:

```
98 pools, 97 matched exactly
DIVERGENT: 1
  cb8ccb08136245d4  dingo=1455876171990  koios=1491430988669  delta=35554816679
```

The single divergent pool is the one that wedges the node, and its five
delegators isolate the problem to one account: four match to the lovelace, one
reads 370114347 where the network counts 35924931026.

That account, `stake_test1uqxfqjftheltx0eczue9te28mscef27wchxjnn04qja57qc7n2rju`
(credential `0c90492bbe7eb33f38173255e547dc3194abcec5cd29cdf504bb4f03`), holds
funds at two addresses with the **same payment credential**:

```
addr_test1g...  header 0x40  type 4  pointer(key)   pointer tail 81b3bb010100
addr_test1q...  header 0x00  type 0  base(key,key)  staking cred 0c90492b...4f03
```

Dingo's own `utxo` rows for that payment credential:

```
amount         staking_key
34553515656    NULL                              <- pointer address
 1000000000    NULL                              <- pointer address
  369927638    0C90492BBE...4F03                 <- base address
```

`reward_live_stake.utxo_stake` for the credential is exactly `369927638` — the
base address alone. The two unattributed UTxOs total `35553515656`, which
accounts for the divergence (the remainder is reward stake accrued between the
snapshot and the live row).

## Why it happens

`Address.StakeKeyHash()` returns an empty hash for a pointer address, and
gouroboros documents this deliberately on the adjacent accessor:

```go
// StakeCredential returns the key or script credential carried by the
// address's staking payload. Pointer and absent staking payloads do not contain
// a credential.
```

That is correct as an address-level statement — a pointer address carries a
*pointer*, not a credential. Resolving it is the ledger's job, and Dingo does not
do it, so the empty hash is persisted as NULL.

## The lookup is already possible

The pointer tail `81b3bb010100` is three variable-length naturals:

```
(slot, tx_index, cert_index) = (2940289, 1, 0)
```

and Dingo already stores the target:

```sql
SELECT hex(staking_key), credential_tag, added_slot
FROM stake_registration WHERE added_slot = 2940289;

0C90492BBE7EB33F38173255E547DC3194ABCEC5CD29CDF504BB4F03 | 0 | 2940289
```

It resolves to exactly the credential in question. `stake_registration` is
indexed on `added_slot`; disambiguating within a slot needs `tx_index` and
`cert_index`, which the certificate rows would have to carry (the cert-ordering
tie-break in CLAUDE.md already relies on `block_index`/`cert_index` elsewhere,
so the concept exists).

## Scope and cautions

- **Consensus-affecting.** Understated stake tightens the leader threshold and
  rejects valid blocks; overstating it would accept invalid ones. It also feeds
  reward calculation, so rewards for such accounts are wrong by the same amount.
- **Era-sensitive.** Pointer addresses confer stake in Shelley through Babbage.
  Conway deprecates them, and a fix must not retroactively credit stake in eras
  where the ledger stopped counting it. Preview epoch 111 is Babbage, where the
  network plainly does count it.
- **Not just this account.** Any pointer-address holder is affected; this replay
  surfaced one because it happened to sit on a pool near its threshold.
- A pointer may reference a registration that does not exist, or one that is
  later deregistered and re-registered. Resolution has to be defined for those,
  and has to behave under rollback.

## Reproduction

Preview, from genesis, no Mithril. Wedges at slot 9609386 in epoch 111. The
per-pool comparison is Dingo `mark(E-1)` against Koios
`pool_history.active_stake` for epoch E, with `/pool_delegators_history` for the
breakdown; `/pool_list` supplies `pool_id_hex` alongside `pool_id_bech32` so no
bech32 implementation is needed.

`pool_stake_snapshot` retention keeps roughly four epochs, so the comparison has
to be taken while the node is near the epoch in question.

wolf31o2:
Fixes #3854.

- Resolve pointer-address UTxOs by the exact `(slot, transaction index, certificate index)` and attribute both stake-registration certificate forms.
- Reject malformed pointers and leave unresolved pointers unattributed rather than guessing from the payment credential.
- Regression coverage verifies the stored stake credential and `reward_live_stake` amount; the test fails before the fix with a nil staking key.
- Validation passed: focused race tests, affected package tests, conformance tests, targeted `go vet`, import-boundary checks, docs parity, GORM checks, and formatting.

---
## Summary by cubic
Fixes #3854. Pointer-address UTxOs now attribute the correct stake credential by resolving the exact `(slot, transaction index, certificate index)` registration. Previously they were left unattributed or guessed from the payment credential; malformed pointers are now rejected and unresolved ones remain unattributed.

- Resolves from the exact certificate location, supports both stake-registration certificate forms, and applies the same resolution when backfilling gap transactions.
- Pointer addresses are validated through the canonical decoder before any keys are stored, so truncated pointers error out instead of being treated as enterprise addresses.
- Gap imports still avoid certificate side effects, so pointer-address outputs that reference a gap-registered certificate remain unattributed.
- Adds regression coverage for the stored stake credential, `reward_live_stake`, and malformed-pointer rejection.

<sup>Written for commit 31644eeab64eaa797e64f13778793dd1849fa6a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3890?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Pointer addresses now resolve their associated staking credentials when transaction outputs are stored.
  * Resolved staking information is reflected in UTxO and live-stake data.

* **Bug Fixes**
  * Malformed pointer addresses are now rejected with a clear decoding error instead of being stored with incomplete address data.
  * Pointer-address resolution validates referenced stake registrations and credential details.


wolf31o2:
Overlap note, so this is a decision rather than something discovered at merge.

**#3855** and **#3890** both target #3854 and carry the same title intent — attributing pointer-address UTxOs to their stake credential.

| | author | size | state | checks |
|---|---|---|---|---|
| #3855 | chrisguiney | +1462 / −71 | CONFLICTING, CHANGES_REQUESTED | 0 failing |
| #3890 | wolf31o2 | +340 / −9 | MERGEABLE | 4 failing |

They are not the same change at different quality — they are the same feature at different scope. The only file both touch is `database/plugin/metadata/sqlstore/transaction_write.go`, so a merge conflict would surface in exactly one place and would not reveal that the rest of the work was done twice.

Neither is currently mergeable: #3855 has conflicts and requested changes, #3890 has four failing checks. Whichever is kept, the other should be closed with a reason rather than left to rot, and if the smaller one is a subset its tests may still be worth carrying across.

Not closing either — #3855 is not mine to close, and this needs an owner decision rather than a coordinator one.

database/plugin/metadata/sqlstore/pointer_address.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/blinklabs-io/dingo/database/models"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+)
+
+// setPointerAddressStakeCredential resolves the stake credential named by a
+// pointer address and adds it to the UTxO model. Pointer addresses carry a
+// certificate location, not a credential hash, so the address-level
+// StakeKeyHash accessor intentionally returns the zero hash.
+func setPointerAddressStakeCredential(
+	db queryer,
```
arepala-uml, line 26: **Blocker: pointer-address stake is resolved unconditionally at every era, but Conway's ledger rules stop counting it.**

Verified against `IntersectMBO/cardano-ledger`:
- Shelley–Babbage (`eras/{shelley,allegra,mary,alonzo,babbage}/impl/src/Cardano/Ledger/*/State/Stake.hs`) share `ShelleyInstantStake`, which has `sisPtrStake` and resolves `StakeRefPtr` into a credential's stake.
- Conway (`eras/conway/impl/src/Cardano/Ledger/Conway/State/Stake.hs`) replaces it with `ConwayInstantStake`, which has no pointer field. `applyUTxOConwayInstantStake` matches only `StakeRefBase`; `StakeRefPtr` outputs contribute nothing.

So once a chain reaches Conway, pointer-address UTxOs stop counting toward stake distribution in the real ledger. This function has no era check and will keep attributing them, overstating stake for any account holding funds at a still-live pointer address in Conway. Issue #3854 calls this out explicitly in its "Scope and cautions" section: "a fix must not retroactively credit stake in eras where the ledger stopped counting it... overstating it would accept invalid ones." No test here exercises a Conway-era pointer output.

`database/plugin/metadata/sqlstore` is intentionally kept independent of ledger/era semantics, so era needs to be threaded in from the caller rather than computed here.

Secondary/non-blocking: the query only unions `stake_registration` and `stake_registration_delegation`; a pointer referencing a `RegistrationCertificate` (cert type 7) doesn't resolve (reproduced locally). But that cert type and its siblings (`stake_vote_registration_delegation`, `vote_registration_delegation`) are Conway-only, so once era gating is fixed correctly they should resolve to nothing anyway — no separate fix needed here.

arepala-uml (changes requested):
Verified locally against current head (`f5f0affb`). The CodeRabbit conflict-path finding (backfill staking key/credential tag when missing) is confirmed fixed at `transaction_write.go:740-741`. `SetGapBlockTransaction` leaving gap-registered pointers unattributed is intentional and covered by `TestSetGapTransactionAttributesPointerAddressStake`.

Blocking issue: pointer-address resolution has no era gating. See inline comment.

database/plugin/metadata/sqlstore/transaction_write.go
```diff
@@ -360,6 +370,16 @@ RETURNING id`,
...
+				if err := setPointerAddressStakeCredential(
+					db,
+					produced.Output.Address(),
+					&model,
+				); err != nil {
+					return fmt.Errorf(
+						"resolve gap produced pointer address: %w",
+						err,
+					)
+				}
```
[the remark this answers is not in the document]
arepala-uml: Re-verified at current head (`f5f0affb`, unchanged since this was posted). `SetGapBlockTransaction` no longer calls `applyTransactionCertificates` at all — that call was removed in 0e69f078 ("avoid certificate side effects during gap import") specifically so gap ingestion doesn't apply cert/deposit effects. So a gap-registered pointer staying unattributed is the intended trade-off, not a bug: `TestSetGapTransactionAttributesPointerAddressStake` asserts exactly this (empty `StakingKey`), and the PR description states it explicitly. Disposition: already addressed by design. Resolving.

database/plugin/metadata/sqlstore/pointer_address.go
arepala-uml: Re-verified at current head (still `f5f0affb`, no new commits since this review): gouroboros's `AddressPayloadPointer` field order (Slot, TxIndex, CertIndex) matches the query parameter order used here, so the join itself resolves the correct on-chain registration. The missing era gate is unchanged — `pointerAddressStakeCredential`/`setPointerAddressStakeCredential` are still called unconditionally on every produced pointer-address output regardless of `point.Slot`'s era, and issue #3854's own "Scope and cautions" section requires era-scoping as an acceptance criterion. No Conway-era test exists. This remains a merge blocker; not resolving.

arepala-uml (approved):

PR #3895, opened 2026-09-04
fix(ledger): preserve same-slot durable order

### LINKED ISSUE #3832: durableAppliedFloor tie-breaks same-slot nonces by hash, not by which block was applied
`database/plugin/metadata/sqlstore/operational.go:623-627` selects the durable applied floor with:

```sql
SELECT hash, nonce, id, slot, is_checkpoint
FROM block_nonce
ORDER BY slot DESC, hash DESC
LIMIT 1
```

With two nonce rows at the same slot — a same-slot competing block — the floor is chosen by **hash lexicographic order** rather than by which block was actually applied. So the "durable applied floor" can name the fork.

This compounds #3678: `enforceDurableTipFloor` produces the rollback target from this value, and #3818 fixes rollback to handle a same-slot competitor correctly but still trusts the floor it is handed.

Found while working #3678; verified against `origin/main`.

Acceptance criteria: the floor selects the applied block at a contested slot, by applied-ness rather than hash ordering; a test seeds two nonce rows at one slot where hash order and applied order disagree and asserts the applied one is chosen.

wolf31o2:
Fixes #3832

- Orders same-slot durable nonces by persisted application ID instead of block hash.
- Adds regression coverage for a later-applied nonce with a lower block hash.
- Updates the database and ledger persistence documentation.

Validation:

- The regression test fails with the old hash ordering and passes with the fix.
- Race-enabled SQLite metadata and ledger recovery tests pass.
- Documentation parity and architecture checks pass.

### DIFF
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -1362,16 +1362,18 @@ heal remains idempotent and uses the primary chain index so retained fork blobs
 and synthetic endorser blobs are never folded.
 
 `GetLatestBlockNonce` returns the single highest-slot `block_nonce` row
-(`ORDER BY slot DESC, hash DESC LIMIT 1`). Because a `block_nonce` row is written
-in the same metadata transaction as its block's UTxO/certificate deltas and the
-ledger tip (`ledgerProcessBlocks`) and is trimmed from below by retention pruning
-(and from above, including competing same-slot hashes, when rollback removes an
-abandoned fork), that maximum slot is the authoritative high-water mark of
-durably applied ledger state for the surviving chain. The ledger uses it as the
-"durable applied floor" to detect and repair a slot-based rollback that left the
-in-memory `currentTip` above the applied state, and to anchor replay-recovery
-rollbacks at or below that floor. The row's `nonce` bytes are ignored by that
-path — only `(slot, hash)` are consumed as a rollback point.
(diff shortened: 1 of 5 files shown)


arepala-uml (approved):
## Review summary

Verified the same-slot tie-break fix directly:

- Confirmed `block_nonce.id` (SQLite rowid) is monotonically increasing across coexisting rows even with retention pruning and rollback deletes, so `ORDER BY slot DESC, id DESC` correctly reflects application order.
- Reverted the fix on top of `origin/main` and reapplied only the new test — it fails as expected, confirming the regression test exercises the bug.
- Reapplied the fix: new and existing `block_nonce` tests pass under `-race`, plus the broader `ledger`/`ledger/leader`/`ledger/forging` nonce-related suite.
- `make docs-parity` and `golangci-lint` clean on touched packages; `go build ./...` clean.
- CI green across all required checks. CodeRabbit is rate-limited (documented); Cubic completed with no issues found.

Non-blocking: `GetLastBlockNonceInRange` (`database/plugin/metadata/sqlstore/operational.go`) still tie-breaks with `ORDER BY slot DESC, hash DESC` — the same bug class fixed here for `GetLatestBlockNonce`. It has no production caller today (only a parity test), so it's dormant, but worth a follow-up fix or issue since it's part of the public `MetadataStore` interface.

No merge blockers.

PR #3897, opened 2026-09-04
fix(ledger): correct Preview VRF leader threshold boundary

### LINKED ISSUE #3811: consensus: preview epoch 111 wedges on a VRF leader threshold with a 2% margin, and the rejection loop spins without backoff
## Summary

A from-genesis Preview replay wedges at slot 9609386 (epoch 111): the node
rejects a canonical block on the Praos VRF leader threshold, rewinds, re-fetches
it, and rejects it again, indefinitely. The tip is frozen at 9609358.

This is **not** the POOLREAP class fixed in #3800 — the producing pool has no
retirement certificate at all, and this database was built with that fix. It is
also a far larger discrepancy than the epoch-20 wedge: `threshold_margin` is
-0.0208 here against -0.00057 there, so it is not a stake-precision artifact.

**I have not established which side is wrong.** The method that settled the
epoch-20 case — diffing every pool's snapshot stake against Koios's per-pool
active stake — is unavailable right now (see Blocked check below), so this is
filed as a reproducible divergence with the node-side evidence, not as a
diagnosis.

## Observed

```
level=WARN msg="header validation rejected a block already on the primary chain;
  rewinding so chain selection can offer another candidate" component=ledger
  failing_block_slot=9609386 rewind_target_slot=9609358
  error="deferred block header verification failed at slot 9609386:
    producer pool cb8ccb08136245d4d4ef2aa755ab036b3335dcd75ec2c234d384e515
    VRF leader value exceeds stake-derived threshold
    (pool stake: 1455876171990, total stake: 114348150359848,
     epoch: 111, snapshot_epoch: 110, snapshot_type: mark,
     threshold_margin: -0.020778526)"
```

The rewind target creeps toward the failing block across attempts — 9609306,
9609323, 9609342, 9609358 — and once it reaches 9609358 the loop tightens to
sub-second intervals. Tip never passes 9609386.

## The node's own state is self-consistent

Every number in the error comes from the node's `mark` snapshot for epoch 110,
and that snapshot does not look corrupted:

The reported total is exactly the snapshot sum:

```sql
SELECT epoch, snapshot_type, COUNT(*) pools, SUM(CAST(total_stake AS INTEGER))
FROM pool_stake_snapshot GROUP BY epoch, snapshot_type ORDER BY epoch DESC;

111|mark|98|112639818495308
110|mark|98|114348150359848   <-- matches "total stake" in the error
109|mark|98|114141730262983
108|mark|98|113957827744504
```

The pool's stake is exactly the sum of its five delegators' captured stake:

```
373580500964 + 370114347 + 3314262532 + 31026209940 + 1047585084207
= 1455876171990                                   <-- matches "pool stake"
```

Its stake is stable across the surrounding epochs, with no jump:

```
epoch  total_stake      delegator_count
108    1479033723862    5
109    1480332415325    5
110    1455876171990    5
111    1456897270957    5
```

Its certificate history is a single registration and nothing else:

```
reg  added_slot=698214  pledge=5000000000  cost=500000000
(no pool_retirement row)
```

And no other pool fails — every VRF rejection in the run names this one pool.

So dingo's relative stake for the pool is 1455876171990 / 114348150359848 =
1.2732%. For the block to be valid the network's figure must be higher, but
which of the two sides diverged is exactly what I could not determine.

## Blocked check

Cross-checking `mark(110)` against Koios per pool is the step that would settle
this, and Koios is refusing requests:

```
HTTP 429  "Exceeded Tier Limit, count was 5083"
```

The public tier caps at 5,000 requests/day with no API key. The parity observer
in this same run exhausted it earlier (it validated 50 epochs first). This needs
either the daily reset or a `--koios-parity-api-key` / `KOIOS_API_KEY` run
(Free tier is 50,000/day) before the per-pool diff can be done.

## Secondary: the rejection loop is unbounded and floods the log

Once the rewind target reaches 9609358 the reject/rewind/refetch cycle runs at
sub-second intervals with no backoff, each iteration logging the full error text.
The log went from 255 MB to 512 MB during this wedge alone. Whatever the stake
divergence turns out to be, a deterministic header rejection that cannot be
resolved should stop rather than spin — the same shape as #3805, but on the
header-validation path rather than the transaction path.

## Environment

- preview, from genesis, `storageMode: api`, no Mithril, single active peer
- built from `main` plus #3799/#3800/#3801, and the gouroboros script-data-hash
  fix from blinklabs-io/gouroboros#2189
- reached epoch 111 / slot 9.6M after ~5 hours
- `mark(110)` captured_slot 9503999

wolf31o2:
Closes #3811

The existing Praos threshold helper floors the real cutoff while the ledger comparison is strict. For full stake with `f=1/3`, this rejects the valid integer floor of the exact cutoff and can wedge Preview leader processing at the boundary.

- Use a protocol-correct integer comparison threshold: `ceil(cutoff)` for non-integral cutoffs, while retaining exact integral cutoffs so equality remains invalid.
- Apply the same threshold to schedule generation and header verification, and invalidate persisted schedules computed with the old semantics.
- Apply the common bounded no-progress backoff to the fixed-delay retry path; rejected blocks remain rejected.

Baseline reproduction: an offline probe on `origin/main` fails because the helper returns the floor boundary instead of the required comparison threshold.

Offline validation: the full `./ledger` race suite and full `./ledger/leader` race suite pass, along with focused boundary and bounded-retry regressions.

---
## Summary by cubic
Closes #3811. Fixes the Praos leader threshold boundary so the valid integer floor of the exact cutoff is accepted, preventing Preview leader processing from wedging at the boundary. The old helper floored the real cutoff while the ledger comparison is strict, so full stake with `f=1/3` rejected the valid floor.

- Adds `consensus/leaderthreshold` to centralize the corrected threshold: `ceil(cutoff)` for non-integral cutoffs, while exact integral cutoffs remain invalid.
- Applies the corrected threshold to schedule generation and header verification, and bumps the schedule format version to invalidate persisted schedules computed with the old semantics.
- Applies the common bounded no-progress backoff to the fixed-delay retry path while preserving its minimum retry delay.

<sup>Written for commit a93dfe1f4298655c56e0d91339f67febc50e0031. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3897?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Behavior Changes**
  * Updated block leadership eligibility thresholds to use precise boundary handling across supported consensus modes.
  * Exact threshold-boundary values are now rejected consistently, while fractional cutoffs preserve the expected eligible range.
  * Leadership schedule format version updated to reflect the threshold calculation change.

* **Reliability**
  * Ledger pipeline retry delays now respect a minimum, increase monotonically, and remain capped during prolonged lack of progress.


arepala-uml (approved):
Verified the boundary bug independently against vendored gouroboros v0.202.8: `IsVRFOutputBelowThresholdWithMode` uses strict `Cmp(threshold) < 0` against `CertifiedNatThresholdWithMode`'s exact `floor(realCutoff)`, so `vrfInt == floor(realCutoff)` is wrongly rejected whenever the true cutoff is non-integer. `leaderthreshold.Threshold`'s ceil-when-non-integer / floor-when-integer logic correctly fixes this, and the exact nth-root check is safe from blowup since sigma ≤ 1 bounds the exponent by the (small, fixed) active-slot-coefficient's bit length.

Built the branch and ran `go test ./consensus/leaderthreshold/... ./ledger/leader/...` — all pass, including the new CPraos/TPraos boundary tests. Schedule format version bump correctly forces recomputation of schedules cached under the old semantics. The retry-backoff refactor in state.go is behavior-preserving for its one caller. CI is green; Cubic reports no issues.

Non-blocking nit: `internal/test/testutil/validated_conway_block.go:203` still calls `consensus.CertifiedNatThreshold` directly instead of the new `leaderthreshold.Threshold` — harmless (old floor ≤ new ceiling) but worth a follow-up for consistency.

PR #3899, opened 2026-09-04
fix(ledger): enforce inbound block execution budgets

### LINKED ISSUE #3898: ledger: enforce aggregate inbound block budget
## Summary

Enforce aggregate inbound block-budget and invalid-transaction overlay rules
when validating block bodies.

## Acceptance criteria

- Apply the block-level budget to the complete inbound transaction set, not
  only to the producer/forging path.
- Preserve the protocol’s invalid-transaction overlay semantics.
- Add non-vacuous valid, over-budget, and invalid-transaction regressions.
- Pass the relevant ledger, mempool, forging, conformance, and race tests.

## Context

Inbound BBODY validation and local block production must enforce the same
aggregate resource boundary. Per-transaction checks alone do not provide that
guarantee.

wolf31o2:
## Summary

- enforce aggregate declared ExUnits against `MaxBlockExUnits` before inbound block deltas are built
- include outer and Dijkstra subtransaction witness sets in transaction-wide execution budgets
- use `Transaction.Consumed()` for phase-2-invalid mempool and forging overlays
- document the consensus-envelope and pending-UTxO contracts

Closes #3898

### DIFF
--- a/ledger/eras/validation.go
+++ b/ledger/eras/validation.go
@@ -970,28 +970,41 @@ func referencedScriptUtxos(
 	return utxos, nil
 }
 
-// DeclaredExUnits returns the total execution units
-// declared across all redeemers in a transaction's
-// witness set. These are the budgets the transaction
-// builder committed to (not the evaluated actuals).
-// Returns an error if the summation would overflow
-// int64.
+// DeclaredExUnits returns the total execution units declared across all
+// redeemers in a transaction, including Dijkstra subtransaction witness sets.
+// These are the budgets the transaction builder committed to (not the
+// evaluated actuals). Returns an error for negative values or if the summation
+// would overflow int64.
 func DeclaredExUnits(
 	tx lcommon.Transaction,
 ) (lcommon.ExUnits, error) {
 	var total lcommon.ExUnits
-	wits := tx.Witnesses()
-	if wits == nil {
-		return total, nil
+	addWitnessSet := func(wits lcommon.TransactionWitnessSet) error {
+		if wits == nil {
+			return nil
+		}
+		redeemers := wits.Redeemers()
+		if redeemers == nil {
+			return nil
+		}
+		for _, val := range redeemers.Iter() {
+			var err error
+			total, err = SafeAddExUnits(total, val.ExUnits)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
 	}
-	redeemers := wits.Redeemers()
-	if redeemers == nil {
-		return total, nil
+
+	if err := addWitnessSet(tx.Witnesses()); err != nil {
+		return lcommon.ExUnits{}, fmt.Errorf(
+			"summing redeemer execution units: %w",
+			err,
+		)
 	}
-	for _, val := range redeemers.Iter() {
-		var err error
-		total, err = SafeAddExUnits(total, val.ExUnits)
-		if err != nil {
+	for _, wits := range lcommon.SubTransactionWitnessSetsFromTransaction(tx) {
+		if err := addWitnessSet(wits); err != nil {
 			return lcommon.ExUnits{}, fmt.Errorf(
 				"summing redeemer execution units: %w",
 				err,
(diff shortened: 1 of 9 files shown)


arepala-uml (approved):
## Review summary

Verified against the current head (`53980ea7`):

- Built the branch and ran `go test -race` on `ledger/...` and `mempool/...`, plus `go vet` and `internal/test/conformance` — all pass.
- Confirmed `Transaction.Consumed()` in the pinned gouroboros `v0.202.5` correctly returns regular inputs when valid and collateral when phase-2-invalid, so the `mempool.go`/`forging/builder.go` fix is behaviorally correct, not just plausible.
- Confirmed against upstream `cardano-ledger`'s Alonzo `Bbody.hs` that `MaxBlockExUnits` is checked as `∑(tx∈txs)(totExUnits tx) ≤ maxBlockExUnits pp` over **all** transactions in the block regardless of `IsValid` — matches this PR's approach of not filtering by validity.
- Tests are genuine regressions with real negative cases (exact-boundary pass, over-budget reject, overflow reject, invalid-tx collateral-vs-inputs assertion).
- Cubic's one P3 finding (duplicated per-era type switches) is resolved by the `protocolBlockLimits` consolidation. CodeRabbit was rate-limited on this PR; Cubic's completed review satisfies the bot-review gate.

One non-blocking item, left as a separate comment: the aggregate budget also sums Dijkstra subtransaction witness-set redeemers, but the current upstream Dijkstra draft (`Rules/Bbody.hs`, `Rules/Utxo.hs`, `TxWits.hs`, `Rules/SubUtxo.hs` on `cardano-ledger` master) only counts the outer transaction's own redeemers for both `MaxTxExUnits` and `MaxBlockExUnits` — `SubUtxo`'s predicate-failure type has no ExUnits case at all. Not consensus-breaking today since Dijkstra isn't wired into dingo's hardfork logic or live anywhere, but worth a conscious call before that era activates.

ledger/eras/validation.go
```diff
@@ -970,28 +970,41 @@ func referencedScriptUtxos(
...
+		return lcommon.ExUnits{}, fmt.Errorf(
+			"summing redeemer execution units: %w",
+			err,
+		)
 	}
-	for _, val := range redeemers.Iter() {
-		var err error
-		total, err = SafeAddExUnits(total, val.ExUnits)
-		if err != nil {
+	for _, wits := range lcommon.SubTransactionWitnessSetsFromTransaction(tx) {
```
arepala-uml, line 1006: Aggregate budget includes Dijkstra subtransaction witness-set redeemers (see `ARCHITECTURE.md` addition and `TestValidateInboundBlockExUnitsIncludesDijkstraSubtransactions`), but the current upstream cardano-ledger Dijkstra draft does not appear to count them:

- `Rules/Bbody.hs` calls `Alonzo.validateExUnits @era txs`, which folds `totExUnits` over the outer transactions only.
- `totExUnits` reads `tx ^. witsTxL . rdmrsTxWitsL`, and Dijkstra's `TxWits.hs` leaves `rdmrsTxWitsL = rdmrsAlonzoTxWitsL` unmodified — it does not fold in subtransaction witness sets.
- The per-tx check (`Rules/Utxo.hs`, "totExunits tx ≤ maxTxExUnits pp") calls `Alonzo.validateExUnitsTooBigUTxO pp tx` with just the outer `tx`, same limitation.
- `Rules/SubUtxo.hs`'s `DijkstraSubUtxoPredFailure` has no ExUnits-budget constructor at all, and its catch-all note calls `ExUnitsTooBigUTxO` "Impossible" for SUBUTXO.

So in the current reference, a subtransaction's declared redeemer ExUnits aren't bounded by either `MaxTxExUnits` or `MaxBlockExUnits`. This PR is stricter than that. Since Dijkstra isn't wired into dingo's hardfork logic and isn't live anywhere yet, this isn't a live consensus risk — but if it's meant to track the eventual final Dijkstra spec rather than being deliberate hardening, it's worth a note (or a tracking issue) so it doesn't quietly diverge once that era activates. Could you confirm which of the two this is intended to be?

Issue #3857, opened 2026-09-04
koiosparity: the missing-row grace window is wall-clock based, so a replay reports not-yet-computed rows as failures


chrisguiney:
## Summary

The grace window that is supposed to classify a not-yet-computed Dingo row as
`reference_lag` is measured in **wall-clock time against the epoch's real close
time**. During a from-genesis replay every epoch closed years ago, so the window
can never apply, and a row Dingo simply has not written yet is reported as
`dingo_db_missing` — a hard failure against a node that is working correctly.

## Observed

A from-genesis Preview replay, node at epoch 46:

```
category            n    epochs
dingo_db_missing    67   44
acct_only_koios    340   13-44
acct_duplicate      80   3-43
acct_only_dingo     46   3-43
acct_zero_reward    43   2-44
pool_departed       18   14-42
```

67 pools at epoch 44 alone — most of the pool set — all reported as missing
Dingo data.

## Why the grace window cannot fire

```go
cat := CategoryDBMissing
if graceHours > 0 && !epochEndTime.IsZero() &&
    now.Sub(epochEndTime) < time.Duration(graceHours)*time.Hour {
    cat = CategoryReferenceLag
}
```

`epochEndTime` is the epoch's actual close time from `KoiosEpochInfo`. Preview
epoch 44 closed around 2022-11-15; `now` is 2026-09-03. That is **1388 days**
against a 24-hour default window, so the comparison is false for every epoch of
a historical replay and the classification is always `dingo_db_missing`.

The window works as intended for a node at the chain tip, where an epoch that
just closed really did close hours ago. It is simply not measuring the right
thing for a node replaying history.

## The rows really are just late

Dingo writes `reward_pool_output` for a stake epoch about two epochs after that
epoch closes:

```
epoch  pools  captured_slot  captured_epoch
44     75     3974400        46
43     75     3974400        46
42     74     3801600        44
41     75     3715200        43
```

The observer validated Koios epoch 44 (Dingo stake epoch 43) at 20:19:49, while
the rows for epochs 43 and 44 were both written at slot 3974400 — epoch 46. The
data arrived after the check, exactly as the grace window anticipates; it just
could not say so.

## Why this has not been seen before

The same replay did not produce this until recently. With
`--koios-parity-accounts` on and default chunking, the per-epoch account fetch
made ~5000 requests and the observer ran roughly 16 epochs behind the node, so
every epoch it checked had long since been computed.

[...]

Issue #3097, opened 2026-08-03
Koios parity: compare every reward account exactly per epoch


wolf31o2:
Parent: #1903

## Summary

Extend `koios-parity` from pool-level checks to a complete per-reward-account comparison for every closed epoch on Preview and Preprod. The current `KoiosAccountRewards` model is schema-only and no account-level fetch or comparison exists.

This is an in-process/library feature, not an export workflow. Implement it with the Dingo reward-parity source and committed-boundary runner described in #3098; do not require a separately synchronized metadata database.

## Requirements

- Fetch Koios `/account_rewards` reference data for every relevant stake address and closed epoch.
- Compare integer lovelace amounts exactly; no rounding, sampling, tolerance, or aggregate-only shortcut.
- Detect and fail on every missing account, unexpected account, duplicate account/epoch row, or differing reward amount.
- Define the account universe explicitly, including accounts with zero reward, newly registered accounts, deregistered/retired accounts, and accounts whose reward is not present in a particular epoch.
- Cover the complete requested range from epoch 0 through the latest safely closed epoch, with a durable coverage manifest/checkpoint that cannot report completion after a partial fetch.
- Obtain Dingo values through the narrow in-process source at the committed epoch boundary, using existing reward calculation/account state or an ephemeral bounded view that remains available before core-mode pruning. Do not add a permanent parity-only table or require an export.
- Persist field-level mismatch evidence with epoch, stake address, Dingo value, Koios value, and reason.
- Add unit, fixture, and end-to-end tests proving all-account coverage and lovelace-exact failure behavior.

## Acceptance criteria

A strict run cannot report PASS unless every account in every closed epoch has a reference row and an exact integer-lovelace match; a single missing, extra, or differing account fails the epoch and process result. The run can be attached to or drive Dingo directly and does not depend on a second Dingo sync.

PR #3900, opened 2026-09-04
fix(koiosparity): aggregate shared reward accounts

### LINKED ISSUE #3841: koiosparity: per-account comparison reports withheld rewards and shared reward accounts as failures
## Summary

The #3097 per-account comparison reports two kinds of false positive against a
ledger that is correct, and both are `FAIL`-class. With the observer's defaults
(`accounts=true`, `strict=true`) a node halts on them.

Both were invisible until now: the per-account path cannot complete against the
public Koios tier (#3796), so this is the first run that exercised it.

1. Rewards the ledger deliberately withheld are compared as if credited, so
   every one becomes `acct_only_dingo`.
2. Duplicate detection keys on `(stake_address, reward_type)` and ignores the
   pool, so an operator running several pools that share one reward account has
   their legitimate per-pool leader rewards reported as duplicates of each
   other.

## Observed

Preview, from genesis, `accounts=true`, epoch 3 (and every epoch after):

```
epoch 3  stake_test1uzuklnhnhy634a5rf0v9pcaf0pva002mw2wjf0ekg6h2encat3ykr
  account_reward_duplicate  reward_type=leader amount=97896659538  (occurrence 2)   acct_duplicate
  account_reward_duplicate  reward_type=leader amount=100574057660 (occurrence 3)   acct_duplicate
  account_reward_presence   96592286094 (type=leader)                               acct_only_dingo
```

`check_epoch_status` shows `FAIL` from epoch 3 onward.

## Neither is a ledger defect

Dingo's three rows are three different pools sharing one reward account:

```sql
SELECT epoch, reward_type, hex(pool_key_hash), amount, spendable
FROM reward_account_output
WHERE hex(staking_key) = upper('b96fcef3b9351af6834bd850e3a97859d7bd5b729d24bf3646aeaccf')
  AND epoch = 3;

3|leader|38F4A58AAF3FEC84F3410520C70AD75321FB651ADA7CA026373CE486|98487908690|0
3|leader|40D806D73C8D2A0C8D9B1E95CCB9F380E40CB4D4B23FF6E403AE1456|98213569111|0
3|leader|D5CFC42CF67F6B637688D19FA50A4342658F63370B9E2C9E3EAF4DFE|98625078479|0
```

Distinct pools, so they are not duplicates of anything. And every one is
`spendable = 0`: the reward account is not registered — it has no `account` row
at all — so the ledger withheld the rewards and credited nothing:

```
account row for the credential:      none
account_reward_delta rows:           0
reward_account_output epochs 2..8:   all spendable=0, guarded=0
```

Koios agrees exactly. `/account_reward_history` for that address at epoch 3
returns **0 rows** — nothing was credited, which is what Dingo also did.

So the comparison flags a disagreement where both sides say the same thing.

## Scope on this run

```sql
SELECT reward_type, spendable, COUNT(*) rows, COUNT(DISTINCT hex(staking_key)) accounts
FROM reward_account_output GROUP BY reward_type, spendable;

leader|0|  76|  1
leader|1| 634| 51
member|0|   1|  1
member|1|1860|166
```

Every withheld row is a guaranteed `acct_only_dingo`. And of the
`(epoch, account)` pairs holding more than one leader row, **all 26 of 26** have
a distinct pool per row — there is not a single genuine duplicate among them.

## Relationship to #3797

This is the same underlying mistake #3797 fixed for the pool-level
`member_rewards` field: comparing a Dingo figure that includes amounts the
ledger never credited against a Koios figure that only ever reports credited
ones. That fix landed in `ComparePoolEpoch`; `CompareAccountEpoch` was not
touched, because the per-account path could not be exercised at the time.

## Suggested direction

- Compare only what the ledger credits. `applyStakeRewards`
  (`ledger/reward_calculation.go`) skips a reward that is not `spendable`, and
  one whose reward account is `guarded` by CIP-0163 expiry; the per-account
  comparison should apply the same pair, exactly as `ComparePoolEpoch` now does
  for `member_rewards`. A withheld reward is not "Dingo-only", it is
  "deliberately not credited".
- Key duplicate detection on `(stake_address, reward_type, pool_key_hash)`.
  Koios's `/account_reward_history` rows carry `pool_id_bech32` for the same
  reason: a leader reward is per pool, and one reward account may serve many.
- Both cases deserve a test with a shared reward account across two pools and a
  withheld reward, since neither shape is reachable from the aggregate path.

## Environment

- `--koios-parity-accounts` (default on), `--koios-parity-strict=false`; with
  the default `strict=true` this halts the node at epoch 3
- self-hosted Koios via `--koios-parity-base-url` (#3838), which is what made
  the per-account path completable at all

wolf31o2:
Fixes #3841

Reject unknown Koios reward types as database mismatches and clarify coverage semantics for aggregated amounts and supported rows.

Validation:

- `go test ./internal/koiosparity -run 'TestCompareAccountEpochUnknownRewardTypeIsDatabaseError|TestCompareAccountEpoch' -count=1`


internal/koiosparity/compare.go
```diff
@@ -832,6 +800,205 @@ func CompareAccountEpoch(
...
+) (map[accountRewardKey]accountRewardTotal, []CheckMismatch) {
+	totals := make(map[accountRewardKey]*big.Int, len(rows))
+	seen := make(map[accountRewardSourceKey]int, len(rows))
+	var out []CheckMismatch
+	for _, r := range rows {
+		poolID, ok := normalizeAccountPoolID(r.PoolIDBech32)
+		if !ok {
+			out = append(out, malformedAccountPoolMismatch(
+				network, epoch, r.StakeAddress, r.Amount, r.PoolIDBech32, now,
+			))
```
arepala-uml: Non-blocking: this call maps `r.Amount` to `dingoValue` and `r.PoolIDBech32` to `koiosValue`, which inverts the malformed-data attribution used everywhere else in this file (e.g. `malformedAccountAmountMismatch`'s Dingo-side call, and the Koios-side pool call just above `aggregateKoiosAccountRewards`). A Dingo-side pool-decode failure ends up recorded as if Koios supplied the bad value.

Reproduced directly:
```
CompareAccountEpoch("preview", 100, nil, []DingoAccountReward{{StakeAddress: "stake1shared", RewardType: "member", Amount: "100", PoolIDBech32: "not-a-pool-id"}}, time.Now(), 0, time.Time{})
// -> Field=account_reward_pool_decode DingoValue="100" KoiosValue="not-a-pool-id"
```
Not reachable today — `compareEpochAccounts` in check.go already validates `row.PoolKeyHash` via `PoolKeyHashHexToBech32` before a `DingoAccountReward` is ever constructed, so this branch is currently dead on the production path — and there's no Dingo-side test mirroring `TestCompareAccountEpochRejectsMalformedPoolID` to catch it. Suggested fix: `malformedAccountPoolMismatch(network, epoch, r.StakeAddress, r.PoolIDBech32, "", now)`, plus a small test alongside `aggregateDingoAccountRewards`.

arepala-uml (dismissed):
Verified in an isolated worktree: build, go vet, and go test -race on internal/koiosparity all pass, including the new shared-pool aggregation, withheld/guarded exclusion, and malformed-data tests. The aggregation logic, category semantics, and ARCHITECTURE.md updates are consistent with #3841 and #3097's exact-match requirement.

One non-blocking nit inline below (field mislabeling in a currently-unreachable diagnostic branch). Approving.

wolf31o2:
Resolved the current-main conflict in `315b84b8` with a normal merge commit. The PR’s shared reward-account aggregation is preserved, while current-main Koios parity helpers and zero-reward-row handling remain intact.

Validated locally:
- `GOWORK=off go test ./internal/koiosparity -count=1`
- focused shared-account and malformed-pool regressions

The updated head needs fresh human review and hosted checks.

internal/koiosparity/compare.go
```diff
@@ -832,6 +800,197 @@ func CompareAccountEpoch(
...
+	epoch uint64,
+	rows []KoiosAccountRewards,
+	now time.Time,
+) (map[accountRewardKey]accountRewardTotal, []CheckMismatch) {
+	totals := make(map[accountRewardKey]*big.Int, len(rows))
+	seen := make(map[accountRewardSourceKey]int, len(rows))
+	var out []CheckMismatch
+	for _, r := range rows {
+		if koiosAccountRewardTypesOutOfScope[r.RewardType] {
+			continue
```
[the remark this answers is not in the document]
wolf31o2: Addressed. `aggregateKoiosAccountRewards` emits `account_reward_type` with `CategoryDBError` for any Koios `type` that is neither `member`/`leader` nor listed in `koiosAccountRewardTypesOutOfScope`, and does so before the row reaches the fold, so an unknown type can no longer surface as `acct_only_koios` or `value_mismatch`. Regression coverage: `TestCompareAccountEpochUnknownRewardTypeIsDatabaseError`.

internal/koiosparity/compare.go
```diff
@@ -832,6 +800,197 @@ func CompareAccountEpoch(
...
+		}
```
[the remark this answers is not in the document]
wolf31o2: Addressed. An unsupported Koios reward type is rejected as `CategoryDBError` on field `account_reward_type` before aggregation, so it cannot be reported as `acct_only_koios` or `value_mismatch`. The out-of-scope set (`treasury`/`reserves`/`refund`) is still skipped silently; every other non-`member`/`leader` value is reported. Covered by `TestCompareAccountEpochUnknownRewardTypeIsDatabaseError`.

internal/koiosparity/compare.go
wolf31o2: Fixed. Both sides now fold through `foldAccountRewards`, and every account mismatch is built by `accountRewardSide.mismatch`, which writes the value to the originating side's column and leaves the other empty — the Dingo and Koios paths can no longer disagree on attribution. `TestCompareAccountEpochRejectsMalformedDingoPoolID` mirrors `TestCompareAccountEpochRejectsMalformedPoolID` and asserts both `DingoValue` and an empty `KoiosValue`, so swapping the columns fails it. The branch stays defensive on the production path: `creditedAccountRewards` still decodes `pool_key_hash` before a `DingoAccountReward` is built.

internal/koiosparity/check.go
```diff
@@ -1145,10 +1146,30 @@ func creditedAccountRewards(
...
+		// An absent pool key hash is not a decode failure. It names the "no
+		// pool" contribution, matching Koios's null pool_id_bech32 for reward
+		// types that have none, and aggregates as its own source.
+		poolID := ""
+		if len(row.PoolKeyHash) > 0 {
+			poolID, err = PoolKeyHashHexToBech32(
+				hex.EncodeToString(row.PoolKeyHash),
+			)
+			if err != nil {
+				errs = append(errs, err)
```
arepala-uml: Pool-hash decode failures are appended to the same `errs` slice as credential-decode failures. The caller (check.go ~1063-1081) unconditionally logs "failed to decode reward_account_output credential" and reports `Field: account_reward_address_decode` for every error in this slice, so a corrupt `PoolKeyHash` gets misdiagnosed as a stake-credential/address problem during on-call triage.

internal/koiosparity/compare.go
```diff
@@ -1293,7 +1265,7 @@ func CompareAccountEpoch(
 				CheckedAt:  now,
 			})
 		default:
-			if !lovelaceEqual(dr.Amount, kr.Earned) {
+			if !lovelaceEqual(dr.Amount, kr.Amount) {
```
arepala-uml: This compares summed per-(address,type) totals (`accountRewardKey`), not per-(address,type,pool) contributions. `accountRewardSourceKey` (line 1075) already uniquely keys duplicate detection by pool; comparing at that granularity throughout instead of folding into `accountRewardTotal` first would be simpler and strictly stronger. As written, two per-pool contributions with opposite-sign errors of equal magnitude cancel in the sum and this check passes despite a real discrepancy on each pool.

internal/koiosparity/check_test.go
```diff
@@ -970,6 +970,90 @@ func TestCheckAccountsEndToEndExactMatchAndMismatch(t *testing.T) {
 	require.Equal(t, "2000001", acctMismatches[0].KoiosValue)
 }
 
+// TestCompareEpochAccountsSkipsWithheldAndAggregatesSharedAccounts covers the
+// raw reward_account_output boundary. Dingo stores one row per pool
+// contribution, including rewards that were computed but not credited;
+// Koios reports only credited rewards at the account/type level. The parity
+// comparison must therefore discard withheld rows and aggregate contributions
+// from different pools before matching the Koios total.
+func TestCompareEpochAccountsSkipsWithheldAndAggregatesSharedAccounts(t *testing.T) {
```
arepala-uml, line 979: New top-level test omits `t.Parallel()` with no `// Not t.Parallel:` justification, and touches no process-global state (same `openTestDingoDB`/`t.TempDir()` fixture as sibling tests in this file that do call it). Same gap in compare_test.go:641, 788, 829, and 1147.

arepala-uml (changes requested):
Local review findings against head eaf0f2a.

1. Comparison sums each account's per-pool contributions into one total per (stake_address, reward_type) before comparing (inline comment on compare.go:1268). Duplicate detection already keys uniquely by (address, type, pool) via `accountRewardSourceKey`, so nothing requires summing before comparison, and summing lets two real, opposite-sign per-pool discrepancies cancel into a false pass, which is the failure mode a parity checker should be least tolerant of.
2. `creditedAccountRewards` merges pool-hash-decode errors into the same error slice as credential-decode errors (check.go:1164); the caller (check.go ~1063-1081, unchanged by this PR) logs and reports every error in that slice under "failed to decode reward_account_output credential" / `account_reward_address_decode`, so a corrupt `PoolKeyHash` is misattributed to address decoding during triage.

Non-blocking: `aggregateKoiosAccountRewards`/`aggregateDingoAccountRewards` duplicate ~35 lines of pool-decode/duplicate-detection logic that differ only in field names and message wording; `accountRewardTotal` duplicates `DingoAccountReward`'s shape; the Dingo-side pool ID is bech32-encoded in check.go then immediately re-decoded/re-encoded by `normalizeAccountPoolID` for no behavioral effect; and five new top-level tests (check_test.go:979; compare_test.go:641, 788, 829, 1147) omit the project's required `t.Parallel()` with no `// Not t.Parallel:` justification, unlike sibling tests in the same files using the same fixtures.

internal/koiosparity/compare.go
wolf31o2: Confirmed and fixed in `f614171e`.

The cancellation is reachable. A reward account shared by several pools carries one leader contribution per pool — #3841's own observed case has three — so two per-pool errors of equal magnitude and opposite sign sum alike, and so does a total one side attributes entirely to one of the pools the other side splits it between. `CompareAccountEpoch` returned zero mismatches for both before this commit.

`accountRewardTotal` now carries each pool's contribution alongside the sum, and `accountRewardPoolsAgree` compares them once the totals agree, reporting `account_reward_pool_amount` with the sorted per-pool breakdown from each side. It is downgraded to `reference_lag` under `rewardsPending`, like the total comparison beside it.

Aggregation is kept rather than replaced. The presence walk stays at `(stake_address, reward_type)` so an epoch whose boundary has not been applied reports the shared key once as `reference_lag` rather than once per pool (#3857, #4130), which `TestCompareAccountEpochSharedAccountPendingRewardsAreALag` pins at `Len == 1`. The per-pool check is therefore strictly additive: no existing verdict changes, and the previously silent case is now reported.

One narrowing: a side that names no pool for a key makes no per-pool statement to disagree with, so the totals remain the whole verdict there. Without it, reference data arriving with no `pool_id` on a member row is reported as a value mismatch purely for lacking the attribution rather than for a difference in what was credited.

`TestCompareAccountEpochPerPoolDiscrepanciesDoNotCancel` covers the swapped pair, the collapsed total, both unattributed directions, and the pending downgrade. With the per-pool branch removed it fails at the swap with zero mismatches returned.

internal/koiosparity/check.go
wolf31o2: Fixed in `f614171e`.

`creditedAccountRewards` now returns `(rows, credentialErrs, poolErrs)`, and `compareEpochAccounts` reports each under its own field and log message: `account_reward_address_decode` with "failed to decode reward_account_output credential" for the credential, `account_reward_pool_decode` with "failed to decode reward_account_output pool_key_hash" for the pool hash. `account_reward_pool_decode` is the field `CompareAccountEpoch` already emits for a pool identifier that will not decode, so the two columns keep one field each.

Both remain `dingo_db_error`, and the credential path is otherwise unchanged, so `accountLifecycleMismatches`'s reliance on this being the reporting path for the current stake epoch's credential failures still holds.

`TestCheckAccountPoolDecodeErrorIsNotAnAddressError` seeds a credential that decodes cleanly with a two-byte `pool_key_hash` and asserts one `account_reward_pool_decode` row and no `account_reward_address_decode` row. It fails against the previous attribution. The producer-level tests in `credited_rewards_test.go` now assert on both slices at every call site rather than on one merged slice.

internal/koiosparity/check_test.go
wolf31o2: Fixed in `f614171e`. `t.Parallel()` added to all five: `TestCompareEpochAccountsSkipsWithheldAndAggregatesSharedAccounts` here, and `TestCompareAccountEpochAggregatesSharedRewardAccounts`, `TestCompareAccountEpochRejectsMalformedAggregatedAmount`, `TestCompareAccountEpochRejectsMalformedPoolID` and `TestCompareAccountEpochUnknownRewardTypeIsDatabaseError` in `compare_test.go`.

None touches process-global state. The four `compare_test.go` cases are pure calls into `CompareAccountEpoch` over locally constructed rows, and this one uses the same per-test `openTestDingoDB`/`t.TempDir()` fixture as the parallel siblings in the file. No `// Not t.Parallel:` justification is needed for any of them.

arepala-uml (approved):

PR #3902, opened 2026-09-04
fix(mesh): bound request body completion time

### LINKED ISSUE #3943: api/mesh: a stalled request body holds a handler goroutine indefinitely
## Summary

The Mesh API caps how many bytes a client may send but not how slowly it may
send them, so a client that completes its headers and then stalls partway
through a declared body holds a handler goroutine for as long as it keeps the
connection open.

## Evidence

`api/mesh/mesh.go`'s `http.Server` sets `ReadHeaderTimeout`, `WriteTimeout`
and `IdleTimeout`, but no `ReadTimeout`. `maxRequestBody` (1 MiB) bounds the
size; nothing bounds the duration.

Reproduced with a raw connection sending a complete request head with
`Content-Length: 4096`, then 44 body bytes, then going quiet: no response for
a 30-second read limit, with the handler still waiting on the body.

## Impact

Each stalled request pins a goroutine indefinitely. Sustained, that is a
resource-exhaustion path against the Mesh listener that costs the client
almost nothing.

## Fix

Bound the body read with a deadline, and add `http.Server.ReadTimeout` as the
backstop for a request whose body no handler reads at all — an unknown route,
or one rejected by authentication before the handler runs. `api/utxorpc`
already sets `ReadTimeout`; this brings Mesh in line.

chrisguiney:
Fixes #3943

## Problem

`maxRequestBody` (1 MiB) caps how many bytes a Mesh client may send, but nothing capped how slowly it may send them. `api/mesh/mesh.go`'s `http.Server` set `ReadHeaderTimeout`, `WriteTimeout`, and `IdleTimeout` but no `ReadTimeout`, so a client that completed its headers and then stopped partway through a declared body held its handler goroutine for as long as it kept the connection open.

Reproduced before the fix: a raw connection sending a complete request head with `Content-Length: 4096` followed by 44 body bytes, then going quiet, received no response for the test's full 30s read limit.

## Change

- `decodeRequest` becomes a `*Server` method and applies a read deadline for the body read (`defaultRequestBodyTimeout`, 30s), clearing it on the way out so the deadline cannot outlive the body into the response write, the drain net/http performs to reuse the connection, or the next request on a kept-alive one. An exceeded deadline fails the decode, and the caller reports the existing `ErrInvalidRequest` — no new error reaches clients.
- The listener gains `http.Server.ReadTimeout` (60s), matching `api/utxorpc`, as the backstop for a request whose body no handler reads: an unknown route, or one rejected by authentication before the handler runs. The per-request deadline never sees those.
- The deadline is set through `http.NewResponseController`. Both middlewares in the chain (`httpcors.Handler`, `apiauth.Middleware`) pass the `ResponseWriter` through unwrapped, so the served path always supports it; an `httptest` recorder does not, and `http.ErrNotSupported` is tolerated so the existing handler-level tests keep working.
- `requestBodyTimeout` is an unexported `ServerConfig` field so tests can shorten it without adding an operator-facing knob or any config/flag/env plumbing. `NewServer` substitutes the default when it is not positive.

## Acceptance criteria

| Criterion | Coverage |
|---|---|
| Enforce a complete-body or inactivity timeout | `TestRequestBodyStalledClientIsBounded` — fails without the fix (no response in 30s), passes in 0.25s with it |
| Return the existing request error for incomplete bodies | Both new wire-level cases assert the `ErrInvalidRequest` code, message, and retriable flag |
| Cover slow | `TestRequestBodyStalledClientIsBounded` |
| Cover truncated | `TestRequestBodyTruncatedIsRejected` |
| Cover oversized | Existing `TestRequestBodyLimit` / `TestRequestBodyAtLimitIsAccepted` |
| Cover normal | `TestRequestBodyNormalRequestUnaffected` (control, under the same short deadline) |

The new cases use a raw socket because neither net/http's client nor an `httptest` recorder can express a stalled or truncated body — both always deliver a complete one.

`TestRequestBodyTruncatedIsRejected` passed before this change: the half-close path already returned `ErrInvalidRequest` via unexpected EOF. It is included as a regression guard, not as evidence of the fix.

Each new assertion was verified to fail without its corresponding change:
- Removing the body deadline: `TestRequestBodyStalledClientIsBounded` fails after 30.01s with "the handler is still waiting on the body".
- Removing `ReadTimeout`: `TestServerTimeoutsAreConfigured` fails with `expected: 1m0s, actual: 0s`.

## Validation

- `go test ./api/mesh/ -race` — ok
- `go test ./api/... -race` — ok
- `make test` (full suite, `-race`) — 84 packages ok. Two `bin` entrypoint tests (`TestEntrypointForwardsSignalsDuringMithrilBootstrap`, `TestEntrypointSignalHandlingSurvivesBootstrapToServeHandoff`) failed on "child did not become ready" under full parallel load; both pass in isolation on this branch and on clean `origin/main`, and they are unrelated to `api/mesh`.
- `golangci-lint run ./...` (0 issues), `GOOS=windows golangci-lint run ./api/mesh/...` (0 issues), `nilaway ./...` (only the pre-existing `ouroboros/blockfetch_musashi_test.go` finding), `modernize`, `golines`, `gofumpt`, `make import-boundaries`, `make docs-parity` — all clean.

Not run: DevNet, conformance, and Antithesis suites — this change touches no consensus, block production, chain selection, mempool, or protocol path.

## Documentation

`ARCHITECTURE.md`'s Mesh API section gains a paragraph on the two request bounds and the listener backstop, parallel to the existing UTxO RPC request-bound note. `DATABASE.md` checked and not affected.

## Follow-up (not in this PR)

`api/blockfrost`'s listener has the same missing `ReadTimeout` (`api/blockfrost/blockfrost.go:318-320`). Left out to keep this change scoped to the issue; worth its own issue.

https://claude.ai/code/session_01FZiMQEmkN34LqZz6UdipSt

---
## Summary by cubic
Binds the Mesh API request body read time so a client that stalls partway through a declared body can no longer hold a handler goroutine indefinitely. Previously only the byte cap (`maxRequestBody`) was enforced; now `decodeRequest` applies a 30s read deadline that is cleared once the body is read, and the listener gains a 60s `ReadTimeout` as a backstop for requests no handler reads. A deadline breach returns the existing `ErrInvalidRequest`, so clients see no new error.

- `requestBodyTimeout` is an unexported `ServerConfig` field so tests can shorten it without adding operator-facing configuration.
- Added wire-level tests for stalled, truncated, and normal bodies; oversized coverage is unchanged.
- The stalled-body test now ties its response allowance to the configured deadline, so a deadline that grew to seconds or fell back to the listener backstop would fail.
- The `ReadTimeout` matches `api/utxorpc`; `api/blockfrost` has the same gap and is left as follow-up.

<sup>Written for commit dd1e1ace1ee7eb495e4bdbb324d8a7487c8ed2e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3902?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added safeguards that limit request bodies to 1 MiB.
  - Requests taking longer than 30 seconds to send their body are rejected as invalid.
  - Requests whose bodies are not read by a handler are protected by a 60-second timeout.

- **Documentation**
  - Documented request-body size and timeout limits in the architecture guide.

- **Bug Fixes**
  - Prevented stalled or incomplete client requests from occupying server resources indefinitely.
  - Confirmed that normally completed requests continue to work under these safeguards.

### DIFF
--- a/api/mesh/listener_test.go
+++ b/api/mesh/listener_test.go
@@ -502,6 +502,8 @@ func TestRequestBodyAtLimitIsAccepted(t *testing.T) {
 
 // TestServerTimeoutsAreConfigured pins the listener timeouts, which
 // bound how long a slow or idle client can hold a connection.
+// ReadTimeout is the backstop for a request whose body no handler
+// reads, which the per-request deadline in decodeRequest never sees.
 func TestServerTimeoutsAreConfigured(t *testing.T) {
 	srv, _ := startTestServer(t, newTestDeps())
 
@@ -511,8 +513,23 @@ func TestServerTimeoutsAreConfigured(t *testing.T) {
 	require.Equal(
 		t, 60*time.Second, httpServer.ReadHeaderTimeout,
 	)
+	require.Equal(
+		t, listenerReadTimeout, httpServer.ReadTimeout,
+	)
+	require.Positive(t, httpServer.ReadTimeout)
 	require.Equal(t, 30*time.Second, httpServer.WriteTimeout)
 	require.Equal(t, 120*time.Second, httpServer.IdleTimeout)
+}
+
+// TestDefaultRequestBodyTimeoutIsApplied asserts a server built
+// without an explicit body deadline still gets one, so the bound
+// cannot be lost by composition code that never sets it.
+func TestDefaultRequestBodyTimeoutIsApplied(t *testing.T) {
+	srv := newTestServer(t, newTestDeps())
+
+	require.Equal(
+		t, defaultRequestBodyTimeout, srv.config.requestBodyTimeout,
+	)
 }
 
 // --- routing ------------------------------------------------------------
--- a/api/mesh/mesh.go
+++ b/api/mesh/mesh.go
@@ -44,6 +44,21 @@ const (
 	defaultListenAddr = ":8080"
 	maxRequestBody    = 1 << 20 // 1 MB
 
+	// defaultRequestBodyTimeout bounds how long a single request may
+	// take to deliver its body. maxRequestBody caps how many bytes a
+	// client may send, but nothing caps how slowly it may send them, so
+	// without this a client that stops partway through a declared body
+	// holds its handler goroutine for as long as it keeps the
+	// connection open.
+	defaultRequestBodyTimeout = 30 * time.Second
+
+	// listenerReadTimeout bounds the whole request read at the
+	// connection, covering the requests whose body no handler reads --
+	// an unknown route, or one rejected by authentication before the
+	// handler runs -- which the per-request deadline above never sees.
+	// api/utxorpc's listener carries the same bound.
+	listenerReadTimeout = 60 * time.Second
+
 	// mainnetMagic is the network magic for Cardano
 	// mainnet, used to determine the address network.
 	mainnetMagic = 764824073
@@ -73,6 +88,12 @@ type ServerConfig struct {
 	// ARCHITECTURE.md's "API security" section.
 	TLS  apiconfig.EffectiveTLS
 	Auth apiconfig.EffectiveAuth
+	// requestBodyTimeout bounds a single request's body read. It is
+	// unexported deliberately: operators get the constant above, not a
+	// knob whose only supported values are "the default" and "short
+	// enough for a test". NewServer substitutes
+	// defaultRequestBodyTimeout when it is not positive.
+	requestBodyTimeout time.Duration
 }
 
 // Server is the Mesh-compatible REST API server.
@@ -146,6 +167,9 @@ func NewServer(cfg ServerConfig) (*Server, error) {
 	if cfg.ListenAddress == "" {
 		cfg.ListenAddress = defaultListenAddr
 	}
+	if cfg.requestBodyTimeout <= 0 {
+		cfg.requestBodyTimeout = defaultRequestBodyTimeout
+	}
 
 	var addrNetID uint8 = lcommon.AddressNetworkTestnet
 	if cfg.NetworkMagic == mainnetMagic {
@@ -207,6 +231,7 @@ func (s *Server) Start(ctx context.Context) error {
 				},
 			),
 			ReadHeaderTimeout: 60 * time.Second,
+			ReadTimeout:       listenerReadTimeout,
 			WriteTimeout:      30 * time.Second,
 			IdleTimeout:       120 * time.Second,
 		}
@@ -369,7 +394,7 @@ func (s *Server) decodeAndValidate(
 	r *http.Request,
 	dst networkRequest,
 ) *Error {
-	if err := decodeRequest(w, r, dst); err != nil {
+	if err := s.decodeRequest(w, r, dst); err != nil {
 		return wrapErr(ErrInvalidRequest, err)
 	}
 	id := dst.networkID()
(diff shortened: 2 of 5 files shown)


api/mesh/request_body_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	conn := dialTestServer(t, baseURL)
+	// Declares far more body than it sends, then goes quiet without
+	// closing: the connection stays open and silent indefinitely.
+	writePartialRequest(
+		t, conn, "/network/list", 4096,
+		`{"network_identifier":{"blockchain":"cardano"`,
+	)
+
+	resp, body := readMeshResponse(t, conn, 30*time.Second)
```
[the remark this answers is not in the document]
chrisguiney: Valid — fixed in dd1e1ace.

Rather than a flat 5s, the allowance is now `respondWithin = 20 * testBodyTimeout`, so it stays tied to the deadline the case configures instead of drifting if that constant changes. The case also now requires the response to arrive no earlier than `testBodyTimeout`, so an answer produced by some other failure cannot read as the bound working.

Verified the tightened assertion catches what the old one missed: with the body deadline regressed to 5s, the case fails with "no response within 5s: the handler is still waiting on the body". The previous 30s allowance passed that same regression.

The truncated case shared the same 30s allowance and got the same bound.

chrisguiney:
@coderabbitai full review

chrisguiney:
@coderabbitai full review

api/mesh/mesh.go
```diff
@@ -449,14 +474,47 @@ func writeError(w http.ResponseWriter, meshErr *Error) {
...
 	w http.ResponseWriter,
 	r *http.Request,
 	dst any,
 ) error {
+	rc := http.NewResponseController(w)
+	s.setBodyReadDeadline(
+		rc, time.Now().Add(s.config.requestBodyTimeout),
+	)
+	// Clear it on the way out: the deadline covers the body read only,
+	// and must not outlive it into the response write, the drain
```
wolf31o2, line 493: The comment is inverted for the drain. Defers run LIFO, so `body.Close()` (L499) runs *before* this clear, and net/http's early-close drain therefore happens *under* the deadline — which is exactly what makes an over-declared `Content-Length` terminate instead of hanging. Swapping the two defers makes that case hang indefinitely; the shipped order answers within the body deadline.

The ordering is load-bearing and currently both misdescribed and untested. Missing case: a request with a valid JSON body and an over-declared `Content-Length` that then stalls — it must answer within the body deadline.

api/mesh/mesh.go
```diff
@@ -44,6 +44,21 @@ const (
...
+	// holds its handler goroutine for as long as it keeps the
+	// connection open.
+	defaultRequestBodyTimeout = 30 * time.Second
+
+	// listenerReadTimeout bounds the whole request read at the
+	// connection, covering the requests whose body no handler reads --
+	// an unknown route, or one rejected by authentication before the
+	// handler runs -- which the per-request deadline above never sees.
+	// api/utxorpc's listener carries the same bound.
+	listenerReadTimeout = 60 * time.Second
```
wolf31o2, line 60: `listenerReadTimeout` has no test seam, unlike `requestBodyTimeout`. The backstop's stated purpose — a stalled body at an unknown route, or one rejected by auth before the handler runs — is never exercised; the only coverage is that the struct field is set.

wolf31o2 (commented):
No blocking findings.

Verified the bound is not inert on the served or TLS path (`httpcors.Handler` and `apiauth.Middleware` both pass the `ResponseWriter` through unwrapped, and h2's writer implements `SetReadDeadline`), and that the deadline does not survive into the response write or the next keep-alive request. Three non-blocking notes inline.

api/mesh/listener_test.go
```diff
@@ -511,10 +513,25 @@ func TestServerTimeoutsAreConfigured(t *testing.T) {
 	require.Equal(
 		t, 60*time.Second, httpServer.ReadHeaderTimeout,
 	)
+	require.Equal(
+		t, listenerReadTimeout, httpServer.ReadTimeout,
```
wolf31o2, line 517: This compares the constant against the value the production code assigns from it, unlike the three neighbouring assertions which pin literals. `listenerReadTimeout = 1 * time.Second` still passes.

wolf31o2 (approved):

PR #3904, opened 2026-09-04
fix(ledger): omit synthetic PlutusV2 cost model from query replies

### LINKED ISSUE #3825: node-parity: Dingo's protocol-parameter reply includes a plutusV1 cost model cardano-node's doesn't

wolf31o2:
Fixes #3825

HardForkBabbage adds a synthetic PlutusV2 cost model for internal validation, but cardano-node omits that model from current protocol-parameter replies until a real governance cost-model update exists.

- Filter only the synthetic PlutusV2 entry at the LocalStateQuery reply boundary.
- Preserve genuine PlutusV1 models, existing PlutusV2 models, and internal validation state.
- Track governance cost-model updates across supported eras, including Dijkstra.

Validation:

- `GOWORK=off GOCACHE=/tmp/dingo-3825-protocol-params-0904-cache go test -tags dingo_extra_plugins -race -count=1 -timeout 20m ./ledger` — PASS, `ok github.com/blinklabs-io/dingo/ledger 561.953s`
- `GOWORK=off GOCACHE=/tmp/dingo-3825-protocol-params-0904-cache go test -tags dingo_extra_plugins -race -count=1 -timeout 10m ./ledger/governance` — PASS, `ok github.com/blinklabs-io/dingo/ledger/governance 84.779s`
- Focused protocol-parameter regression with `-race` — PASS, `ok github.com/blinklabs-io/dingo/ledger 4.723s`
- Pre-fix regression proof — EXPECTED FAIL at the synthetic PlutusV2 omission assertion

---
## Summary by cubic
Fixes #3825 by keeping HardForkBabbage's synthetic PlutusV2 cost model out of current protocol-parameter query replies. The fallback is still used internally for validation until a real governance cost-model update exists.

- Filters only the synthetic PlutusV2 entry at the `LocalStateQuery` reply boundary, leaving genuine PlutusV1/V2 models intact.
- Marks the model as synthetic only when the transition injects the exact default fallback and no real PlutusV2 model existed before.
- Persists the synthetic marker in sync state, restoring it on startup and rollback, and clears it when governance enacts a real PlutusV2 update, including Dijkstra changes that equal the fallback.
- Adds regression tests for query reply filtering and governance cost-model tracking.

<sup>Written for commit 750884ac86d809c6ba5cbf89faa9c509962edbe7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3904?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ledger/governance/enact_test.go
```diff
@@ -101,6 +101,7 @@ func TestEnactProposal_DijkstraParameterChange(t *testing.T) {
 	}, proposal)
 	require.NoError(t, err)
 	require.True(t, result.PParamsChanged)
+	require.False(t, result.CostModelsChanged)
```
arepala-uml, line 104: `EnactmentResult` has no `CostModelsChanged` field on this branch's merged tree -- it was renamed to `PlutusV2CostModelWritten` by #3884, which is already on `main` and was pulled in by this PR's own merge commit. This fails `go vet ./ledger/governance/...`:
```
ledger/governance/enact_test.go:104:26: result.CostModelsChanged undefined (type *EnactmentResult has no field or method CostModelsChanged)
```
and is exactly what's failing go-test on Linux/Windows/macOS/race in this PR's CI run. Suggest `require.False(t, result.PlutusV2CostModelWritten)` to match the field that actually exists.

ledger/governance/enact_test.go
```diff
@@ -268,6 +269,42 @@ func TestEnactProposal_DijkstraParameterChangeWritesPlutusV2CostModel(
...
+				AnchorHash:    testBytes(32, 0xD8),
+				ReturnAddress: testBytes(29, 0xD9),
+			}
+			pparams := &gdijkstra.DijkstraProtocolParameters{}
+			result, err := EnactProposal(&EnactmentContext{
+				DB: db, Slot: 2000, Epoch: 42, PParams: pparams,
+				UpdateFn: eras.PParamsUpdateDijkstra,
+			}, proposal)
+			require.NoError(t, err)
+			require.Equal(t, costModels[1] != nil, result.CostModelsChanged)
```
arepala-uml, line 303: Same undefined-field break as line 104 (`result.CostModelsChanged`), this time in the new `TestEnactProposal_CostModelsChangedOnlyForPlutusV2` test -- confirmed as the second compile error in both local `go vet`/`go test` and the CI log. Renaming to `result.PlutusV2CostModelWritten` should restore intent: note the existing field's semantics are `_, ok := a.ParamUpdate.CostModels[1]; ok` (key presence only), slightly different from the old `CostModelsChanged`'s `ok && model != nil` (also required a non-nil value) -- worth double-checking the rename still asserts what this parameterized test intends, though for the two fixtures here (`{1,2,3}` and `{4,5,6}`, both non-nil) the two definitions agree.

arepala-uml (changes requested):
Pipeline is red on the current head (750884ac) across all four go-test jobs (Linux, Linux race, Windows, macOS), and it reproduces locally with `go vet ./ledger/governance/...` / `go test ./ledger/governance/...`. Root cause: the branch's merge of `main` (which already carries #3884's rename of `EnactmentResult.CostModelsChanged` to `PlutusV2CostModelWritten`) left this PR's own `enact_test.go` additions referencing the now-removed `CostModelsChanged` field, so `ledger/governance` fails to build. Confirmed this does not reproduce against `origin/main` on its own (no `CostModelsChanged`/`PlutusV2CostModelWritten` reference exists there outside this PR's diff), so the break is owned by this change's merge commit, not a pre-existing base-branch issue. Requesting changes on that basis; approval is otherwise withheld only for the compile failure -- the new `ledger/protocol_params_query_test.go` end-to-end coverage (`TestCurrentProtocolParamsQueryOmitsSyntheticV2` / `...IncludesExistingV2`) builds and passes as written, and is a solid addition on top of the already-merged #3884 filtering fix.

Bot coverage note: CodeRabbit's check reads "Review rate limited" and Cubic's check at this exact head (750884ac) reads "This commit only merges another branch into the PR branch, so cubic completed this check without running an AI review" -- neither is a genuine review of the current head (Cubic's last real review was at commit e746e4c1, several commits behind, before the breaking merge). Both are non-reviews, not evidence the diff is clean.

wolf31o2:
Superseded: synthetic PlutusV2 cost-model filtering is already present on main via #3884; recreate the regression coverage separately if still needed.

PR #3906, opened 2026-09-04
fix: bound Koios account parity checkpoints

### LINKED ISSUE #3849: koiosparity: the per-account coverage table grows without bound and dominates the cache
## Summary

With `--koios-parity-accounts` on (the default), the parity cache writes one
`koios_account_checked` row per stake address per epoch and never removes any.
On a Preview replay at epoch 80 that table plus its two indexes is **8.59 GB of
an 8.70 GB cache — 98.7%** — and the cache is already larger than the node's own
metadata database.

Nothing prunes it. `Cache.InvalidateStaleAccountChunks` clears checkpoint state
when the account universe changes, but there is no epoch-based retention.

## Measured

Preview, from genesis, at chain epoch 80, 78 epochs validated:

```
sqlite> SELECT name, SUM(pgsize) FROM dbstat GROUP BY name ORDER BY 2 DESC;
koios_account_checked          4.28 GB
idx_kaced_net_epoch_chunk      2.16 GB
idx_kaced_net_epoch_addr       2.15 GB
koios_account_universe         0.03 GB
idx_kau_net_addr               0.03 GB
koios_account_fetch_staged_rows 0.01 GB
```

Row counts:

```
koios_account_checked    23,636,028      <- 303,026 addresses x 78 epochs
koios_account_universe      303,026
koios_account_rewards        24,585      <- rewards actually found
```

Alongside the node's own storage:

```
metadata.sqlite   5.4 GB
blob              2.2 GB
.koios/cache.db   8.7 GB
```

The 23.6M rows record that an address was checked in an epoch. The rows that
carry a reward number are the other 24,585 — three orders of magnitude fewer.

## Growth

Row growth is `len(account universe)` per epoch, independent of activity.
Preview's chain tip is presently around epoch 1400, so continuing this replay to
the tip projects to roughly 4.3 x 10^8 rows. At the observed ~370 bytes per row
including indexes, that is on the order of **150 GB** for a checker cache beside
a node database that is 5.4 GB at epoch 80.

Mainnet is worse in the same direction: the stake-address universe is several
times Preview's.

## An epoch-level summary already exists

`koios_account_coverage` holds exactly one row per epoch — `requested_count`,
`fetched_count`, `complete` — and had 78 rows for the same 78 epochs. Resume
decisions that only need "was this epoch finished" can already be answered
without the per-address table.

The per-address rows do carry two things the summary does not: `chunk_hash`, for
invalidating a checkpoint when the universe changes, and `reward_row_count` per
address. Both are needed for the epoch currently being fetched. Whether they are
needed for an epoch that was validated and closed 70 epochs ago is the question
this issue is really asking.

## Suggested direction

1. Retain `koios_account_checked` only for epochs still in flight, deleting rows
   for an epoch once `koios_account_coverage.complete` is set — the summary row
   survives as the durable record.
2. Failing that, add a retention window (keep the last N epochs) with a flag, so
   a long survey run does not need unbounded disk.
3. `idx_kaced_net_epoch_chunk` and `idx_kaced_net_epoch_addr` together cost more
   than the table. If retention shrinks the table to the in-flight epoch, one of
   them may become unnecessary.

Whichever way it goes, the cache growing to many times the node database on a
test network is a poor default for a tool meant to run alongside a syncing node.

## Environment

- `--koios-parity-enabled --koios-parity-accounts` (default on),
  `--koios-parity-base-url https://preview-koios.tosidrop.me/api/v1`
- chain epoch 80, observer epoch 79

wolf31o2:
Closes #3849

- Bound per-account Koios fetch checkpoints to a rolling four-epoch window while retaining authoritative reward rows for exact comparison.
- Persist exact zero-reward counts with capped samples and migrate legacy coverage rows before checkpoint eviction.
- Evict checkpoints only after fetch workers join or the observer completes its sequential check.

Validation: focused race tests, package compilation, `go vet`, architecture-boundary tests, and documentation parity passed. Full-package tests were limited by the sandbox forbidding `httptest` listener sockets.

## Summary by CodeRabbit

* **New Features**
  * Zero-reward account reporting now preserves exact counts and a capped sample of affected addresses.
  * Account checkpoint data is automatically pruned beyond the rolling retention window while reward and coverage information remains available.
  * Existing coverage records are upgraded with zero-reward summaries automatically.

* **Bug Fixes**
  * Lifecycle mismatch reporting remains accurate after older account checkpoint data is removed.


arepala-uml (dismissed):

arepala-uml (approved):
Reviewed this bounded-checkpoint fix. Confirmed the checkpoint-vs-reward-data distinction is real and correctly drawn: two per-address checkpoint tables are bounded to a rolling 4-epoch window, but the authoritative reward rows and per-epoch coverage summary are never pruned. A zero-reward summary is captured at commit time (with a migration for pre-existing cache files) so historical lifecycle reporting survives checkpoint eviction even though the underlying rows don't.

- Eviction arithmetic verified correct (exactly 4 epochs retained, correct off-by-one, guards uint64 underflow), and pruning only runs after in-flight fetch workers have joined — no race with a worker's own checkpoint writes.
- Confirmed no new locking is needed: the observer runs on a single sequential goroutine, and pruning within Fetch happens strictly after wg.Wait().
- Windows CI note: this PR's own CI shows go-test (Windows) green, including the previously-flagged TestObserverStrictModeCancelsOnAccountMismatch — corroborating that a Windows failure seen on a sibling PR was environmental, not a real bug in this package.
- All three prior bot findings were verified independently via git history to be genuinely fixed in earlier commits within this same PR.
- `go build`/`vet`/`test -race`/`golangci-lint` all clean. ARCHITECTURE.md accurate line-by-line.

No merge-blocking issues found. One minor non-blocking follow-up: Cache.GetAccountUniverseForEpoch is dead in production with a now-stale doc comment.

PR #3907, opened 2026-09-04
fix(forging): preserve sigma snapshot consistency

### LINKED ISSUE #3815: Torn sigma read: numerator and denominator can come from different snapshot writes
`stakeDistributionAdapter.GetPoolStake` and `GetTotalActiveStake` each open their own `db.MetadataTxn` (`node_forging.go:509-536`, used at `:542` and `:561`). `election.go:769-812` calls them sequentially, so a snapshot re-capture landing between the two reads produces a σ whose numerator and denominator come from different snapshot writes.

The window is small but the consequence is a leader schedule computed from an inconsistent σ, which is not reproducible from either snapshot alone.

Found while investigating #2798.

Acceptance criteria: numerator and denominator are read in one transaction; a test that interleaves a snapshot re-capture between the two reads shows a consistent σ.

wolf31o2:
Fixes #3815.

Adds a deterministic concurrent regression at the forging adapter boundary. The test coordinates an atomic snapshot recapture between the numerator and denominator reads and verifies that both values come from one committed snapshot generation. A nil-by-default test hook provides only the required sequencing point; production behavior is unchanged.

Validation:

- Concurrent regression passes under `go test -race`.
- Root and `ledger/leader` packages pass under `go test -race`.
- Regression repeated five times under `go test -race`.

---
## Summary by cubic
Fixes #3815 by keeping the forging sigma numerator and denominator on the same committed snapshot generation when a recapture occurs between the two reads.

Adds a test-only hook to `stakeDistributionAdapter` and a concurrent regression test that replaces the snapshot while the reader is paused between queries. Production behavior is unchanged because the hook is nil by default.

Validation:
- New regression and `ledger/leader` tests pass under `go test -race`.

<sup>Written for commit ff22728867d9f65f7bd0006207403bd67e997c24. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3907?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/node_forging.go
+++ b/node_forging.go
@@ -504,6 +504,9 @@ func (b *blockBroadcaster) AddBlock(
 // snapshot rotation semantics as other ledger queries.
 type stakeDistributionAdapter struct {
 	ledgerState *ledger.LedgerState
+	// afterPoolStakeReadFn is a test-only hook for coordinating a concurrent
+	// snapshot recapture after the transaction has read the numerator.
+	afterPoolStakeReadFn func()
 }
 
 func (a *stakeDistributionAdapter) getStakeDistribution(
@@ -597,6 +600,12 @@ func (a *stakeDistributionAdapter) GetPoolAndTotalActiveStake(
 			poolKey,
 			err,
 		)
+	}
+	// Test-only synchronization seam. The transaction has already observed
+	// the numerator, so a concurrent recapture can commit before the
+	// denominator query without changing this transaction's snapshot.
+	if a.afterPoolStakeReadFn != nil {
+		a.afterPoolStakeReadFn()
 	}
 	totalActiveStake, err = view.GetTotalActiveStake(epoch)
 	if err != nil {
--- a/node_forging_sigma_denominator_test.go
+++ b/node_forging_sigma_denominator_test.go
@@ -17,7 +17,9 @@ package dingo
 import (
 	"io"
 	"log/slog"
+	"sync"
 	"testing"
+	"time"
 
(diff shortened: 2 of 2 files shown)


arepala-uml (approved):
Verified in an isolated worktree off current origin/main.

- Production change is a nil-by-default `afterPoolStakeReadFn` hook on `stakeDistributionAdapter`; behavior is unchanged when unset.
- Confirmed the diff is scoped to `node_forging.go` (hook field/call) and the new test — no unrelated changes, against the correct merge-base.
- Ran the new regression test and existing sigma tests under `-race`: pass.
- Reintroduced a torn read (second transaction for the denominator query) to confirm the test actually fails without the fix — it does, with a clear numerator/denominator mismatch.
- No overlap with `TestStakeDistributionAdapterSigmaPairSurvivesRecapture`: that test's own doc comment scopes itself to inter-call consistency and states it does not prove intra-call atomicity without this seam.
- `gofmt`/`go vet`/`golangci-lint` clean on changed files; `ledger/leader` sigma-pair test passes.
- Bot review satisfied: Cubic clean, CodeRabbit rate-limited (documented in thread).

No blocking issues.

Issue #3792, opened 2026-09-02
peergov: decide whether documentation address ranges are valid peer candidates


chrisguiney:
Raised by a Cubic review on #3789, deferred there because the remediation is far larger than that PR.

#3789 added `peergov.IsRoutableIP`, rejecting ranges that `net.IP`'s own class predicates report as global unicast but that cannot be a peer we meant to dial: RFC 6598 shared address space, RFC 6890 protocol assignments, RFC 2544 benchmarking, RFC 1112 reserved, and the RFC 6666 discard-only prefix.

It deliberately left RFC 5737 (`192.0.2.0/24`, `198.51.100.0/24`, `203.0.113.0/24`) and RFC 3849 (`2001:db8::/32`) accepted. The review's objection is fair on principle: a production policy should not be shaped by what the test suite happens to use, and these are not globally routable either.

## Why it was not done in #3789

These ranges are the repository-wide convention for "a public address" in tests — roughly 113 occurrences across 19 files:

| Occurrences | File |
|---|---|
| 24 | `peergov/ledger_dial_security_test.go` |
| 24 | `peergov/dial_spread_test.go` |
| 16 | `peergov/peergov_test.go` |
| 15 | `connmanager/ip_rate_limit_test.go` |
| 7 | `peergov/ledger_discovery_test.go` |
| … | `ledger/`, `connmanager/`, `midnight/`, `internal/config/`, `internal/node/` |

Rejecting them means migrating all of it. Some of those suites are not simple substitutions: `dial_spread_test.go` and `ip_rate_limit_test.go` depend on the distribution of addresses across subnets, so replacement fixtures have to preserve the property under test rather than just parse.

## Why the harm is lower than the ranges already rejected

RFC 6598 was rejected because a carrier routes it internally, so dialing an advertised address there can reach another subscriber's host. RFC 5737 and RFC 3849 are not routed anywhere. A peer advertising one costs a failed dial and a peer-list slot until the entry is dropped — bounded, and self-correcting. It is a resource-consumption question, not a reachability one.

## Decision needed

- Are documentation and test ranges valid peer candidates? Consistency with "globally routable" says no.
- If no, migrate the fixtures. Pick replacement addresses that preserve what each suite tests, particularly the subnet-distribution ones.
- Consider whether the production policy should instead be expressed as an allowlist of globally routable space, rather than a growing denylist of reserved ranges.

Acceptance criteria:

[...]

Issue #3431, opened 2026-08-24
peersharing: bound replies and reject unroutable addresses


wolf31o2:
`ouroboros/peersharing.go` accepts a peer-sharing reply larger than the requested amount, allocates output for the complete reply, and forwards addresses without rejecting loopback, private, multicast, unspecified, or otherwise non-routable values. A remote peer can therefore amplify a small request into a frame-sized reply and steer peer-governor candidates toward unusable addresses.

Acceptance criteria:

- Reject or truncate replies that exceed the requested peer count.
- Reject non-globally-routable and malformed addresses before adding peer-governor candidates.
- Preserve valid IPv4, IPv6, and supported DNS-hostname peers.
- Add oversized-reply and address-class vectors with a valid public-peer control.

PR #3909, opened 2026-09-04
fix(peersharing): filter non-routable peer addresses

### LINKED ISSUE #3790: peersharing: apply a routability policy to the addresses we share
Split out from #3431, which bounds and filters the peer-sharing reply we *receive*. The mirror direction has no routability filter: `peersharingShareRequest` shares any peer whose `Sharable` flag is set, checking only that the address yields a parsable IP and port.

Two ways a non-routable address can carry `Sharable`:

1. **Topology config (confirmed by reading the wiring).** `peergov/topology.go` passes the operator's `advertise` flag straight through to `Peer.Sharable` for local roots and public roots (`upsertTopologyPeer` callers around lines 245 and 271). `AddPeer` deliberately exempts topology sources from the routability check so operators can configure private local relays — `TestAddPeer_AllowsTopologyWithPrivateIP` covers that. A private-IP local root with `advertise: true` is therefore advertised to remote peers.

2. **Inbound connections (not confirmed).** `peergov/connections.go:844` sets `tmpPeer.Sharable = tmpPeer.Connection.VersionData.PeerSharing()` when a connection is established, and `AddPeer` exempts `PeerSourceInboundConn` from the routability check. Reading the code, an inbound connection from a private address whose handshake advertises peer sharing would become sharable without anyone configuring it. I could not confirm this end-to-end: the flag is set only once a live connection with handshake `VersionData` exists, and at `AddPeer` time the peer is still `Sharable: false`. This needs verifying before it is treated as real.

The concern is an internal-topology leak — advertising private or loopback neighbours to public peers — and, for the remote, useless candidates it will fail to dial.

This is deliberately not folded into #3431, because case 1 is an explicit operator instruction rather than a defect. Deciding it means deciding whether `advertise` on a non-routable address is an error, a warning, or silently honored.

Acceptance criteria:

- Decide whether a non-routable address may ever be advertised, and whether an operator's explicit `advertise` overrides that.
- Confirm or rule out the inbound path in case 2 with a test that drives a real handshake.
- Apply the resulting policy in `peersharingShareRequest`, reusing `peergov.IsRoutableIP` rather than restating the address classes.
- Cover topology-advertised private addresses, inbound private peers, and a valid public-peer control.

wolf31o2:
Peer-sharing could forward malformed, private, and reserved addresses into peer admission.

- Centralize routability classification for network-learned IPs.
- Filter addresses when serving and consuming peer-sharing replies.
- Bound replies to the requested count while preserving valid public IPv4 and IPv6 peers.

Fixes #3790

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved peer sharing to advertise only valid, globally routable IP addresses with nonzero ports.
  * Revalidated received peer addresses and limited responses to the requested number of entries.
  * Blocked reserved, internal, protocol-assigned, and otherwise non-routable address ranges from peer admission and outbound connections.

* **Documentation**
  * Documented peer-sharing address validation and handling of locally configured private peers.


wolf31o2:
`main` is merged in at 499a7f78; the branch is mergeable. One behavioural disagreement with `main` survives the merge and is not a merge artifact.

This branch adds `2001:20::/28` (RFC 7343 ORCHIDv2) to the non-routable prefix list at `peergov/peers.go:154` and asserts `{"ipv6 orchidv2", "2001:20::1", false}` at `peergov/routable_ip_test.go:108`.

`main` does not filter that prefix, asserts the opposite, and states the reason in a comment added by #3789:

```go
// ORCHIDv2 directly abuts the deprecated ORCHID block and IANA marks
// it globally reachable, so the /28 must not spill into it.
{"ipv6 orchidv2", "2001:20::1", true},
```

That comment is deleted on this branch. The merge was clean because the file had already been reconciled by the earlier merge commit cc2f6ba2, so git had nothing to flag — the reversal is semantic, not textual.

Both cannot be right. Either ORCHIDv2 is a valid peer candidate, as #3789 concluded from the IANA IPv6 Special-Purpose Address Registry, or this branch is correct to filter it and #3789's entry needs removing along with its comment. Deciding it here keeps the next change to this list from re-litigating it.

wolf31o2:
Correction to the previous comment, which framed this branch's filtering as the side needing justification. It is the correct side; `main`'s expectation is the one that is operationally wrong.

Both sources checked directly.

IANA IPv6 Special-Purpose Address Registry, 2001:20::/28 ORCHIDv2 (RFC 7343, allocated 2014-07, no termination date): Source True, Destination True, Forwardable True, **Globally Reachable True**, Reserved-by-Protocol False. So `main`'s comment states the registry accurately.

RFC 7343 states the operational rule the registry column does not capture:

- ORCHIDs "should not appear in actual IPv6 headers"
- they are "non-routable addresses from the IPv6-layer perspective"
- "Routers MAY be configured not to forward any packets containing an ORCHID as a source or a destination address"
- they are "routable at an overlay level" only, and are "endpoint identifiers at applications" rather than "identifiers for network location at the IP layer, i.e., locators"

`isRoutablePeerAddress` decides whether a peer-shared address is a usable connection candidate. An ORCHIDv2 address is a cryptographic identifier with no locator semantics, so a dial to one cannot succeed. Filtering the /28 here is correct, and `{"ipv6 orchidv2", "2001:20::1", false}` is the right expectation.

What remains is the deleted comment rather than the behaviour. `main`'s note exists to keep the deprecated ORCHID filter (2001:10::/28) from spilling into the adjacent /28, which is still worth stating now that both blocks are filtered for different reasons — 2001:10::/28 because it is deprecated (terminated 2014-03), 2001:20::/28 because RFC 7343 excludes it from IPv6-layer routing despite being globally reachable in the registry. Restoring a note to that effect above the two prefixes would stop this being re-litigated on the next change to the list.

wolf31o2:
Comment restored at 98ca3141, in both \`peergov/peers.go\` and \`peergov/routable_ip_test.go\`. It now records why each ORCHID block is excluded — 2001:10::/28 as the deprecated allocation terminated 2014-03, 2001:20::/28 as Globally Reachable in the IANA registry but kept out of IPv6 headers by RFC 7343 — so neither prefix can be widened to cover the other.

No behaviour change. `go test ./peergov/` exit 0, `golangci-lint run ./peergov/...` 0 issues.

ouroboros/peersharing.go
```diff
@@ -102,6 +102,13 @@ func (o *Ouroboros) peersharingShareRequest(
 			)
 			continue
 		}
+		if !peergov.IsRoutableIP(ip) {
```
chrisguiney, line 105: **Blocker: the fix has no test.**

This guard and the `portNum == 0` guard below are the whole of the #3790 fix, and nothing in the branch exercises either. I deleted both in place and `go test ./ouroboros/...` stayed green (`ok  github.com/blinklabs-io/dingo/ouroboros  10.329s`).

`TestPeerSharingShareRequestBoundsValidPeers` looks like coverage but is a guard that passes either way: its `10.0.0.1:3001` is added with `PeerSourceTopologyLocalRoot`, and `AddPeer` sets `Sharable` only for `PeerSourceP2PGossip`, so that peer is dropped by the pre-existing `!peer.Sharable` check at line 78 and never reaches this one.

The case #3790 describes is reachable through the exported API. The following passes at 98ca3141 and fails with this guard reverted in place (`actual: []string{"10.0.0.1", "44.0.0.1"}`), so it is a real regression test rather than a guard:

```go
func TestPeerSharingShareRequestDropsAdvertisedPrivateTopologyPeer(t *testing.T) {
	logger := slog.New(slog.NewTextHandler(io.Discard, nil))
	pg := peergov.NewPeerGovernor(peergov.PeerGovernorConfig{Logger: logger})
	pg.LoadTopologyConfig(&topology.TopologyConfig{
		LocalRoots: []topology.TopologyConfigP2PLocalRoot{{
			Advertise: true,
			AccessPoints: []topology.TopologyConfigP2PAccessPoint{
				{Address: "10.0.0.1", Port: 3001},
			},
		}},
		PublicRoots: []topology.TopologyConfigP2PPublicRoot{{
			Advertise: true,
			AccessPoints: []topology.TopologyConfigP2PAccessPoint{
				{Address: "44.0.0.1", Port: 3001},
			},
		}},
	})

	o := newOuroboros(OuroborosConfig{Logger: logger})
	o.peerGov = pg
	peers, err := o.peersharingShareRequest(opeersharing.CallbackContext{}, 10)
	require.NoError(t, err)
	got := make([]string, 0, len(peers))
	for _, p := range peers {
		got = append(got, p.IP.String())
	}
	require.Equal(t, []string{"44.0.0.1"}, got)
}
```

It also settles the precondition #3790 could not confirm by reading alone: after `LoadTopologyConfig` with `advertise: true`, `pg.GetPeers()` reports `10.0.0.1:3001` as `Sharable`, so case 1 in the issue is real and this guard is what stops it.

#3790's acceptance criteria also ask for an inbound private peer case and a public control; the zero-port guard below is uncovered as well.

peergov/routable_ip_test.go
```diff
@@ -105,9 +105,13 @@ func TestIsRoutableIP(t *testing.T) {
...
-		// ORCHIDv2 directly abuts the deprecated ORCHID block and IANA marks
-		// it globally reachable, so the /28 must not spill into it.
-		{"ipv6 orchidv2", "2001:20::1", true},
+		// ORCHIDv2 abuts the deprecated ORCHID block and is Globally Reachable
+		// in the IANA registry, but RFC 7343 keeps it out of IPv6 headers, so
+		// it is rejected as a peer candidate and the /28 boundaries are pinned
+		// on both sides.
+		{"ipv6 orchidv2", "2001:20::1", false},
+		{"ipv6 orchidv2 upper bound", "2001:2f:ffff:ffff:ffff:ffff:ffff:ffff", false},
+		{"ipv6 above orchidv2", "2001:30::", true},
```
chrisguiney, line 114: **Non-blocking, but this line pins the opposite policy for the same address class.**

`2001:30::/28` is the RFC 9374 DRIP Entity Tag prefix. DETs *are* ORCHIDs — RFC 9374 constructs them with the RFC 7343 ORCHIDv2 algorithm — so the rationale this PR adds at `peergov/peers.go` for rejecting `2001:20::/28` ("an endpoint identifier rather than a locator ... can never be dialled as a peer") applies verbatim to the block this line asserts is a valid peer candidate.

The IANA IPv6 Special-Purpose Address Registry marks both `2001:20::/28` and `2001:30::/28` Globally Reachable: True, so the registry column is not what separates them; the branch's own RFC-level reasoning is, and it points the same way for both.

Either add `2001:30::/28` to `unreachablePrefixes` alongside it, or move the upper-bound probe out of identifier space (`2001:40::`) so the `/28` boundary is still pinned without asserting a policy the adjacent comment contradicts.

ouroboros/peersharing.go
```diff
@@ -169,40 +183,19 @@ func (o *Ouroboros) RequestPeersFromPeer(peer *peergov.Peer) []string {
...
-// Bounding the entries examined, rather than collecting `requested` valid ones
-// from an unbounded reply, keeps the work constant; a peer that pads its reply
-// with unusable entries spends its own slots doing so.
-//
-// Entries that cannot serve as a peer address are dropped. peergov.AddPeer
-// applies the same routability policy, but it treats a host that does not
-// parse as an IP as a routable hostname, so a malformed entry rendered as
-// "<nil>:3001" would be accepted there and sent to a DNS lookup. Rejecting it
-// here keeps every emitted candidate an IP literal.
+// candidate addresses. The reply is bounded to the request so a remote cannot
```
chrisguiney, line 186: **Non-blocking, comment accuracy.**

After the reordering the loop no longer stops at `requested` *entries*; it stops at `requested` *emitted candidates*, and a reply longer than that is now walked in full.

I audited that as a cost regression and it is not one: the expensive path (`peergov.AddPeer` -> DNS resolve + peer-list scan) is still capped at `requested`, and per-entry work is now `To16()` plus at most 19 `netip.Prefix.Contains` calls. But the sentence claims a bound the code no longer has. Suggest: "the number of candidates emitted is bounded to the request, so a remote cannot amplify one request into unbounded peer-admission work."

The deleted paragraph also carried #3431's rationale for the old ordering. A line recording why the ordering changed would keep that decision from being re-litigated.

ARCHITECTURE.md
```diff
@@ -3767,6 +3767,14 @@ dial capacity; local-root and bootstrap peers, and any peer that has connected
 at least once, keep retrying, and the drop is suppressed while the node has no
 eligible upstream so its last leads back onto the network are never discarded.
 
+Peer sharing is an address-validation boundary in both directions. The server
```
chrisguiney, line 3770: **Non-blocking, operator visibility.**

#3790 asks whether `advertise` on a non-routable address is "an error, a warning, or silently honored". This branch answers "silently ignored": the only signal is a `Debug` line per share request. An operator running a private-network relay cluster (compose or Kubernetes) with `advertise: true` on private local roots loses intra-cluster peer-sharing discovery with nothing at default log level to explain it.

A single `Warn` at `LoadTopologyConfig` time would make the policy this paragraph documents observable. Worth stating here that there is no override.

chrisguiney (changes requested):
Reviewed at 98ca3141. Requesting changes on one blocker: the behaviour change has no test.

The code is correct. I reproduced #3790 case 1 against this branch — a topology local root `10.0.0.1:3001` with `advertise: true` is `Sharable` in the governor, and with the new `IsRoutableIP` guard reverted in place `peersharingShareRequest` returns it to the remote. The guard fixes that. But deleting both new send-path guards leaves `go test ./ouroboros/...` green, so nothing protects the fix. #3790's acceptance criteria ask for topology-advertised private addresses, inbound private peers, and a public control; the branch adds only receive-path coverage. A ready test that fails before and passes after is inline.

Three non-blocking findings inline: the new `2001:30::` pin, the reworded bound on `peerSharingReplyAddresses`, and the absence of an operator-visible signal when `advertise` is ignored.

**Class audit — complete, and stated because it is a result.** Against the IANA special-purpose registries, `IsRoutableIP` covers loopback, RFC 1918 and `fc00::/7`, CGNAT `100.64/10`, IPv4 and IPv6 link-local, unspecified plus `0.0.0.0/8`, IPv4 and IPv6 multicast, broadcast via `240/4`, TEST-NET-1/2/3, `2001:db8::/32`, `192.0.0.0/24`, `198.18/15`, `192.88.99/24`, `64:ff9b:1::/48`, `100::/64`, `2001:2::/48`, and now both ORCHID blocks. IPv4-mapped IPv6 is not a bypass: `IsRoutableIP` calls `Unmap()` and `TestIsRoutableIPUnmapsV4` pins it. NAT64 `64:ff9b::/96` is deliberately accepted (IANA Globally Reachable: True) and pinned. The only IANA "Globally Reachable: False" block absent is `3fff::/20` (RFC 9637 documentation) — that gap is on `origin/main`, not introduced here, and is covered by the allowlist proposal in #3792.

**Direction audit.** The filter is correctly asymmetric in effect. The accept side already had `IsRoutableIP` on `origin/main` (`git show origin/main:ouroboros/peersharing.go`), so this PR does not newly tighten the inbound path and cannot newly break private topologies there, and `AddPeer` keeps its topology and `PeerSourceInboundConn` exemptions so private local roots stay usable locally. The only new tightening is on the advertise side, which is what #3790 asked for.

**Wire compatibility.** Matches gouroboros v0.202.10: `peersharing.PeerAddress{IP net.IP, Port uint16}`, and `MarshalCBOR` errors only when `To16() == nil`, which the send path now excludes. `MsgShareRequest.Amount` is `uint8`, so the `make([]PeerAddress, 0, amount)` on line 76 is bounded at 255 — no attacker-controlled allocation.

**Checks at 98ca3141** (merge-base equals `origin/main` `eabf653d`, so the merged tree is the head tree):

| check | exit |
|---|---|
| `go build ./...` | 0 |
| `go vet ./ouroboros/... ./peergov/...` | 0 |
| `gofmt -l ouroboros peergov` | 0, no output |
| `golangci-lint run ./ouroboros/... ./peergov/...` | 0, 0 issues |
| `go test -race ./ouroboros/...` | 0 (18.6s) |
| `go test -race ./peergov/...` | 0 (4.9s) |
| fail-before: revert `2001:20::/28` | `TestIsRoutableIP/ipv6_orchidv2` and `/ipv6_orchidv2_upper_bound` FAIL, as they should |
| fail-before: revert receive-path reorder | `TestPeerSharingReplyCollectsRequestedValidAddresses` FAILs, as it should |
| fail-before: revert send-path guards | `./ouroboros` stays green — the blocker |

GitHub reports all 12 check runs green at this head; none is attributable to this change.

**Not run:** no live peer-sharing handshake. `peersharingShareRequest` is unexported and only ever called directly from `ouroboros` unit tests, so nothing here or in CI exercises the filter over a real NtN connection — the unit tests prove the adapter, not the wire. `AGENTS.md` calls for the DevNet end-to-end suite on changes touching NtN protocols; that needs devnet infrastructure I do not have. Please state its status.

**Bots at this head:** neither reviewed 98ca3141. CodeRabbit posted "Review skipped: auto incremental reviews are disabled". cubic's two reviews are stamped at `4164d9be` and `02cb6332`, both stale; there is no cubic check run at this head. All three cubic threads are resolved and outdated, and I have checked each against 98ca3141 and replied.

ouroboros/peersharing.go
```diff
@@ -165,9 +179,47 @@ func (o *Ouroboros) RequestPeersFromPeer(peer *peergov.Peer) []string {
...
+// reserved, and zero-port addresses are discarded before they reach DNS or
+// peergov.
+func (o *Ouroboros) peerSharingReplyAddresses(
+	peers []opeersharing.PeerAddress,
+	requested int,
+) []string {
+	if requested <= 0 {
+		return nil
+	}
+	if len(peers) > requested {
```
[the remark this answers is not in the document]
chrisguiney: Checked against 98ca3141: fixed. `peerSharingReplyAddresses` now filters first and breaks after collecting `requested` valid addresses. Proved by restoring the old truncate-then-filter ordering in place — `TestPeerSharingReplyCollectsRequestedValidAddresses` then fails with `actual: []string{"44.0.0.1:3001"}` against the expected two entries.

One consequence the new doc comment does not state: the loop now walks the whole reply rather than the first `requested` entries. Audited and not a regression — the expensive path (`peergov.AddPeer` -> DNS resolve + peer-list scan) is still capped at `requested` — but the comment claims a bound the code no longer has. Raised in the review.

peergov/peers.go
```diff
@@ -80,13 +81,55 @@ func (p *PeerGovernor) GetPeers() []Peer {
...
+	netip.MustParsePrefix("0.0.0.0/8"),      // RFC 1122 this network
+	netip.MustParsePrefix("100.64.0.0/10"),  // RFC 6598 shared address space
+	netip.MustParsePrefix("192.0.0.0/24"),   // RFC 6890 IETF assignments
+	netip.MustParsePrefix("192.88.99.0/24"), // RFC 7526 deprecated 6to4
+	netip.MustParsePrefix("198.18.0.0/15"),  // RFC 2544 benchmarking
+	netip.MustParsePrefix("240.0.0.0/4"),    // RFC 1112 reserved
+	netip.MustParsePrefix("64:ff9b:1::/48"), // RFC 8215 local translation
+	netip.MustParsePrefix("100::/64"),       // RFC 6666 discard-only
+	netip.MustParsePrefix("2001:2::/48"),    // RFC 5180 benchmarking
+	netip.MustParsePrefix("2001:10::/28"),   // RFC 4843 ORCHID
```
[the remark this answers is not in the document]
chrisguiney: Checked against 98ca3141: fixed. `2001:20::/28` is in `unreachablePrefixes`. Proved by deleting the prefix in place — `TestIsRoutableIP/ipv6_orchidv2` and `/ipv6_orchidv2_upper_bound` then fail.

Class audit of the finding rather than the line: the adjacent new case `{"ipv6 above orchidv2", "2001:30::", true}` puts RFC 9374 DRIP Entity Tags on the accepted side. DETs are ORCHIDs built with the same RFC 7343 construction, so the reasoning that rejects `2001:20::/28` applies to them too. Raised in the review.

peergov/routable_ip_test.go
```diff
@@ -106,9 +106,7 @@ func TestIsRoutableIP(t *testing.T) {
 			"2001:1f:ffff:ffff:ffff:ffff:ffff:ffff",
 			false,
 		},
-		// ORCHIDv2 directly abuts the deprecated ORCHID block and IANA marks
-		// it globally reachable, so the /28 must not spill into it.
-		{"ipv6 orchidv2", "2001:20::1", true},
+		{"ipv6 orchidv2", "2001:20::1", false},
```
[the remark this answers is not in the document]
chrisguiney: Checked against 98ca3141: addressed. The rationale is restored in both `peergov/peers.go` and `peergov/routable_ip_test.go`.

Recording the disposition on the substance, since the finding disputed the policy and not only the missing comment: the registry column is accurate — IANA lists `2001:20::/28` Globally Reachable: True — and the branch rejects anyway on RFC 7343 grounds ("should not appear in actual IPv6 headers"), which is the right call for a dial target. Rejecting the finding as a policy objection, accepting it as a documentation gap, which the head fixes.

The same argument applies to `2001:30::/28` (RFC 9374 DETs, also Globally Reachable: True in the registry, also ORCHIDs), which the test now pins as accepted. Raised in the review.

ouroboros/peersharing.go
chrisguiney: Cleared at e205822d. `TestPeerSharingShareRequestDropsAdvertisedPrivateTopologyPeer` covers this guard and fails without it: with the `IsRoutableIP` block deleted in place, `go test ./ouroboros/ -run TestPeerSharing` reports `actual: []string{"10.0.0.1", "44.0.0.1"}` against the expected single entry. That is #3790 case 1 driven through the exported API, with `44.0.0.1` as the public control.

The `portNum == 0` guard below is still uncovered. Reclassifying that half of this finding as non-blocking now that the routability guard the issue is about is covered, and restating it on its own line.

peergov/routable_ip_test.go
chrisguiney: Stands at e205822d: `{"ipv6 above orchidv2", "2001:30::", true}` is unchanged.

Non-blocking. The line pins existing behaviour rather than changing it — `2001:30::/28` is absent from `unreachablePrefixes` on `main` and on this branch. The objection is to what it asserts: the branch's own RFC-level rationale for rejecting `2001:20::/28` (an endpoint identifier, not a locator) applies verbatim to RFC 9374 DRIP Entity Tags, which are ORCHIDs built with the same RFC 7343 construction. Either add the prefix alongside it, or move the upper-bound probe to `2001:40::` so the `/28` boundary is still pinned without asserting the opposite policy for the same address class.

ouroboros/peersharing.go
chrisguiney: Stands at e205822d, narrowed.

The operative claim is accurate: `peergov.AddPeer` is reached at most `requested` times, so a remote cannot amplify one request into unbounded peer-admission work. Only the subject is off — what is bounded to the request is the emitted candidate list, not the reply, which is now walked in full. Non-blocking wording.

ARCHITECTURE.md
chrisguiney: Stands at e205822d. The new paragraph records the policy, which answers #3790's first acceptance criterion; the runtime signal is still a `Debug` line per share request.

Non-blocking, but it is the part of this change most likely to surprise an existing deployment: an operator running a private relay cluster with `advertise: true` on RFC 1918 local roots loses intra-cluster peer-sharing discovery, with nothing at default log level to explain it. A single `Warn` at `LoadTopologyConfig` time would make the documented policy observable.

ouroboros/peersharing.go
```diff
@@ -111,6 +118,13 @@ func (o *Ouroboros) peersharingShareRequest(
 			)
 			continue
 		}
+		if portNum == 0 {
```
chrisguiney, line 121: **Non-blocking, coverage.** This guard is uncovered: deleting it leaves `go test ./ouroboros/` green.

It is load-bearing and reachable through the exported API. A public root access point `{Address: "44.0.0.2", Port: 0}` with `advertise: true` becomes `44.0.0.2:0` with `Sharable: true`, and passes `IsRoutableIP`; without this guard `peersharingShareRequest` returns `[{44.0.0.2 0} {44.0.0.1 3001}]`.

Adding that access point to the `PublicRoots` block of `TestPeerSharingShareRequestDropsAdvertisedPrivateTopologyPeer` covers it without a new test: `require.Equal(t, []string{"44.0.0.1"}, got)` already fails when a zero-port peer is emitted.

chrisguiney (approved):
Approving at e205822d. The blocker from the previous review is cleared.

`TestPeerSharingShareRequestDropsAdvertisedPrivateTopologyPeer` is a regression test rather than a guard: deleting the `IsRoutableIP` block at `ouroboros/peersharing.go:105` in place makes it fail with `actual: []string{"10.0.0.1", "44.0.0.1"}`. Removing `2001:20::/28` likewise fails `TestIsRoutableIP/ipv6_orchidv2` and `/ipv6_orchidv2_upper_bound`.

Prior findings at this head: the blocker is cleared; the three non-blocking ones — the `2001:30::` DRIP Entity Tag pin, the `peerSharingReplyAddresses` comment wording, and the missing operator-visible signal when `advertise` is ignored — stand and are answered on their threads. The `portNum == 0` half of that blocker is still uncovered but is reclassified non-blocking, noted inline.

The routability class audit is unchanged from the previous review apart from `2001:20::/28`. `3fff::/20` and `5f00::/16` remain the only IANA "Globally Reachable: False" blocks absent from `unreachablePrefixes`; both are on `main` rather than introduced here, and both fall under the allowlist proposal in #3792.

Both directions hold. The send-path filter keys on `IsRoutableIP` regardless of `Peer.Source`, so #3790 case 2 — an inbound private peer marked `Sharable` from its handshake — is covered by the same line as case 1. The acceptance criterion asking for a handshake-driven test is not met, which is better recorded on #3790 than blocked on here.

Merged against current `main` the branch is clean and `go test ./ouroboros/... ./peergov/...` passes on the merged tree, so the ORCHIDv2 expectation `main` still carries does not resurface on merge.

PR #2207, opened 2026-05-08
fix(test): TestArchiveProxy hangs waiting for badger LOCK release

arepala-uml:
Closes #2196 

---
## Summary by cubic
Fixes a hang in the archive demo tests by waiting for Badger’s LOCK after resetting data-dir permissions, so `TestArchiveProxy` no longer stalls after stopping `dingo-pruning`.

- **Bug Fixes**
  - In `stopPruning`, run a one-shot `busybox` container to relax bind-mount permissions before checking the Badger `LOCK`, ensuring the lock check is reliable and the test proceeds.

<sup>Written for commit 2837f3e7ac3ed8980501ab12836774a4171c21c1. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Tests**
  * Improved internal test organization and documentation for better clarity during shutdown sequences.

Issue #3748, opened 2026-09-01
Implement GenesisDelegationState for MIR certificate validation


wolf31o2:
gouroboros PR #2129 adds `common.GenesisDelegationState`, an optional ledger-state capability that supplies the currently delegated genesis keys and the update quorum, and `UtxoValidateMIRGenesisQuorum` in the Shelley, Allegra, Mary, Alonzo, and Babbage rule lists.

The rule fails closed: when a transaction carries a `MoveInstantaneousRewardsCertificate` and the ledger state does not implement `GenesisDelegationState`, validation returns `GenesisDelegationStateUnavailableError`. Transactions without an MIR certificate are unaffected, and Conway and Dijkstra expunge MIR.

Dingo's ledger state does not implement the interface, so once it consumes a gouroboros release containing #2129, replaying Shelley-through-Babbage history will reject every block carrying an MIR certificate.

The inputs are already available:

- `shelleyGenesis.GenDelegs` and `shelleyGenesis.UpdateQuorum` (`api/utxorpc/query.go:1012`, `internal/node/backfill.go:271`)
- `database/models/genesis_delegation.go` (`GenesisDelegateHash`)

### Acceptance criteria

- The ledger state used for block and transaction validation implements `GenesisDelegateKeyHashes() ([]common.Blake2b224, error)` and `GenesisUpdateQuorum() (uint, error)`.
- `GenesisDelegateKeyHashes` reflects genesis key delegation certificates applied after genesis, not only the initial `GenDelegs` mapping.
- A Shelley-through-Babbage block carrying an MIR certificate validates against mainnet history.
- Rule indices pinned by `validateUtxoValidationSkipIndex` still resolve; #2129 appends the rule rather than inserting it, so no pinned index should shift.

PR #3911, opened 2026-09-04
fix(ledger): expose the current epoch to gouroboros ledger rules

### LINKED ISSUE #3908: ledger: a registration for a retired pool is treated as an update, so its deposit is never charged
## Summary

A from-genesis Preview replay wedges at slot 17020890 (epoch 197). Dingo rejects
a canonical transaction for value conservation, short by exactly the pool
deposit, because it treats a registration for an **already-retired** pool as an
update rather than a new registration.

```
tx a82343228b1c4eaed145d988eaf8b5cb57e56d0883e3093fcfb5df07fa455b2f
  validation failure at slot 17020890:
  value not conserved: consumed 997574722, produced 497574722
```

`997574722 - 497574722 = 500000000` — 500 ADA, the pool deposit exactly.

159 occurrences; the node is no longer following the chain.

## The pool's history

The transaction carries a pool registration certificate (cert type 3) for
`864295fcb8da8782c85e6c00d82584d6a6104e8edf4d3c7a73ce1ee4`. Dingo's own tables:

```
pool_registration   added_slot 16988057   (epoch 196)
pool_retirement     added_slot 16988068   (epoch 196), retiring in epoch 197
```

Epoch 197 begins at slot 17020800. The failing transaction is at 17020890 —
**90 slots after the retirement took effect**. To the ledger the pool is retired,
so this is a fresh registration and a fresh 500 ADA deposit is charged. Dingo
still considers it registered, so it charges nothing and the transaction fails
to balance.

## Where

`UtxoValidateValueNotConservedUtxo` adds the deposit only for a pool it believes
is not currently registered:

```go
case *common.PoolRegistrationCertificate:
	reg, _, err := ls.PoolCurrentState(common.Blake2b224(tmpCert.Operator))
	if err != nil {
		return err
	}
	if reg == nil {
		producedValue.Add(producedValue, poolDeposit)
	}
```

`LedgerView.PoolCurrentState` (`ledger/view.go`) picks the latest entry from
`pool.Registration` and returns it, with no reference to whether the pool has
since retired. Retirement is reported through the second return value, which
this rule discards with `_`.

So for any pool that has retired and is registering again, `reg` is non-nil, the
deposit is never added, and a canonical transaction fails conservation by
exactly one deposit.

## Two ways to fix it, and I do not think it is my call which

1. **`PoolCurrentState` stops returning a registration once the retirement has
   taken effect.** Semantically the most accurate — a retired pool is not
   currently registered — but it changes what every caller sees, and some may be
   relying on getting the last known parameters for a retired pool.
2. **The conservation rule consults the retirement epoch it already receives.**
   Localised, but the rule would need the current epoch to compare against, and
   it currently has only a slot.

The pending-retirement value is already plumbed through and discarded here,
which is suggestive, but I have not traced the other `PoolCurrentState` callers
to see what option 1 would disturb.

## Related

This is the same family as #3794 and #3800 — retirement recorded but the derived
"currently registered" state not following it. Those covered delegation and
stake; this is the deposit accounting.

The register/retire/re-register sequence within twelve slots looks like a
deliberate testnet exercise of exactly this edge, which is worth having a
regression test for regardless of which fix is chosen.

## Secondary

Same shape as #3889 and #3896: the rejection is deterministic and the node
retries it indefinitely rather than failing closed.

## Environment

- epoch 197, Babbage
- `main` plus #3843, #3846, #3853, #3855, and gouroboros with the #3875 and
  #3896 fixes

chrisguiney:
Fixes #3908.

## Why

`LedgerView` did not implement gouroboros' `common.EpochState`:

```go
type EpochState interface {
	EpochForSlot(slot uint64) (uint64, error)
}
```

gouroboros obtains the current epoch by optional type assertion, so a ledger
state that omits the method does not fail to build — every rule that needs an
epoch quietly takes its degraded path instead.

Two rules depend on it. The Shelley POOL retirement bound can only reject
epoch zero without it. And the pool-deposit decision cannot tell a retired
pool from a registered one: a registration certificate for an already-retired
pool is charged no deposit, and the transaction then fails value conservation
by exactly the deposit amount.

Preview slot 17020890 wedges the node on the second case.

`LedgerState` already has the mapping — `epochForSlot`, backed by the same
epoch cache the header path uses. This exposes it.

## Relationship to gouroboros#2207

gouroboros#2207 fixes the pool-deposit rule itself. That change alone does
**not** fix Dingo: with no `EpochForSlot`, the `EpochState` assertion there
fails and the helper falls back to the old behavior. Both halves are
required.

## Tests

`TestLedgerViewSatisfiesEpochState` pins the capability itself. It is the
test that matters here, because the failure mode is a silently missing
method rather than a wrong result. Renaming `EpochForSlot` fails the package.

`TestLedgerViewEpochForSlot` covers the mapping against a known epoch cache:
first slot of an epoch, the slot from the issue, the last slot of the
previous epoch, and a slot outside the cache — which must error rather than
guess.

## Validation

Built against gouroboros v0.202.8 plus gouroboros#2207 and restarted on the
data directory already wedged at slot 17020890. The node passed the wedge,
crossed the epoch 197 boundary, and has since replayed through epoch 207 at
~2700 slots/s.

`make lint` reports 0 golangci-lint issues across all modules and the Windows
target. `nilaway` reports one finding in `ouroboros/blockfetch_musashi_test.go`,
a file this branch does not touch; it is present on `main` from #3826.

`ARCHITECTURE.md` updated — it lists the optional gouroboros capabilities
`LedgerView` implements, and `EpochState` was missing. `DATABASE.md` checked
and not affected; no schema, query, or storage change.

---
## Summary by cubic
Implements the optional gouroboros `common.EpochState` capability on `LedgerView` so ledger rules that need the current epoch stop silently taking their degraded paths.

- Without an epoch, the pool-deposit rule could not distinguish a retired pool from a registered one; re-registering a retired pool was charged no deposit and failed value conservation by exactly that amount, wedging preview slot 17020890.
- The pool retirement bound could only reject epoch zero.
- `EpochForSlot` delegates to the existing epoch cache lookup used by the header path.
- `ARCHITECTURE.md` now documents the capability and why omitting it is dangerous.
- Requires `gouroboros`#2207; that fix to the pool-deposit rule alone still falls back to the old behavior without `EpochForSlot`, so both changes are needed.
- `TestLedgerViewSatisfiesEpochState` pins the interface; renaming `EpochForSlot` fails the package.
- `TestLedgerViewEpochForSlot` covers first and last slots of an epoch, the slot from issue #3908, and errors on slots outside the cache.
- Validated by replaying past the wedge to epoch 207 at ~2700 slots/s; `make lint` is clean and the `nilaway` finding is pre-existing on `main`.

<sup>Written for commit 176d10be3978abcd90072109e1c0dadcde8b65ec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3911?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ledger/view.go
+++ b/ledger/view.go
@@ -414,6 +414,24 @@ func (lv *LedgerView) PoolCurrentState(
 	return currentReg, pendingEpoch, nil
 }
 
+// EpochForSlot returns the epoch containing the given slot, satisfying
+// gouroboros' optional common.EpochState capability.
+//
+// Several ledger rules are expressed relative to the current epoch and degrade
+// to a weaker check when the ledger state cannot supply one. Without this the
+// pool-deposit decision cannot tell a retired pool from a registered one, so a
+// registration for an already-retired pool is charged no deposit and the
+// transaction fails value conservation by exactly that amount
+// (issue #3908); the retirement-epoch bound on pool retirement certificates is
+// skipped for the same reason.
+func (lv *LedgerView) EpochForSlot(slot uint64) (uint64, error) {
+	epoch, err := lv.ls.epochForSlot(slot)
+	if err != nil {
+		return 0, err
+	}
+	return epoch.EpochId, nil
+}
+
 // IsPoolRegistered checks if a pool is currently registered
 func (lv *LedgerView) IsPoolRegistered(pkh lcommon.PoolKeyHash) bool {
 	reg, _, err := lv.PoolCurrentState(pkh)
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2774,6 +2774,16 @@ The `LedgerView` interface provides query access to ledger state:
   inclusive. The legacy hash-only `CommitteeMember` and
   `CommitteeMembers` methods omit ambiguous same-hash key/script identities
   instead of selecting one by map iteration.
+- `EpochForSlot` implements the `common.EpochState` capability, mapping a slot
+  to its epoch through the same epoch cache the header path uses. gouroboros
+  expresses several rules relative to the current epoch and obtains it by
+  optional type assertion, so a ledger state that omits this method silently
+  takes the weaker path in each of them rather than failing to build. Two
+  depend on it: the pool-deposit decision cannot otherwise distinguish a
+  retired pool from a registered one, so a re-registration of a retired pool
+  is charged no deposit and then fails value conservation by exactly the
+  deposit; and the Shelley POOL retirement bound can only reject epoch zero
+  without it.
 - The Conway and Dijkstra validation compositions replace the pinned
   hash-only committee certificate and voter rules when that capability is
   present. Cold authorization/resignation certificates and hot committee votes
--- a/ledger/view_epoch_for_slot_test.go
+++ b/ledger/view_epoch_for_slot_test.go
@@ -1,0 +1,61 @@
+package ledger
+
+import (
+	"testing"
+
+	"github.com/blinklabs-io/dingo/database/models"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestLedgerViewSatisfiesEpochState pins that LedgerView provides gouroboros'
+// optional EpochState capability.
+//
+// Rules that are expressed relative to the current epoch degrade to a weaker
+// check when the ledger state cannot supply one, and they degrade silently. The
+// pool-deposit decision is the case that found this: without an epoch it cannot
+// tell a retired pool from a registered one, charges no deposit for a
+// registration that needs one, and the transaction then fails value
+// conservation by exactly the deposit (issue #3908).
+func TestLedgerViewSatisfiesEpochState(t *testing.T) {
+	var lv any = &LedgerView{}
+	_, ok := lv.(lcommon.EpochState)
+	require.True(t, ok,
+		"LedgerView must satisfy common.EpochState, or every rule that needs "+
+			"the current epoch silently takes its degraded path")
+}
+
+// TestLedgerViewEpochForSlot checks the mapping itself against a known cache.
+func TestLedgerViewEpochForSlot(t *testing.T) {
+	ls := &LedgerState{
+		epochCache: []models.Epoch{
+			{EpochId: 196, StartSlot: 16_934_400, LengthInSlots: 86_400},
+			{EpochId: 197, StartSlot: 17_020_800, LengthInSlots: 86_400},
+		},
+	}
+	ls.publishSnapshotsLocked()
+	lv := &LedgerView{ls: ls}
+
+	for _, tc := range []struct {
+		name string
+		slot uint64
+		want uint64
+	}{
+		{"first slot of an epoch", 17_020_800, 197},
+		// The slot that wedged the replay, ninety slots into epoch 197.
+		{"the slot from issue 3908", 17_020_890, 197},
+		{"last slot of the previous epoch", 17_020_799, 196},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := lv.EpochForSlot(tc.slot)
+			require.NoError(t, err)
+			assert.Equal(t, tc.want, got)
+		})
+	}
+
+	t.Run("a slot outside the cache is an error, not a guess", func(t *testing.T) {
+		_, err := lv.EpochForSlot(99_000_000)
+		require.Error(t, err)
+	})
+}


chrisguiney:
## Class audit

A missing optional capability is invisible by construction, so I checked the
whole class rather than only the one that wedged the node. Probing
`*LedgerView` against every ledger-state capability gouroboros type-asserts:

```
UtxoState                     ok
CertState                     ok
PoolState                     ok
EpochState                    MISSING   <- this PR
StakeCredentialDepositState   ok
CommitteeCredentialState      ok
DRepDelegationState           ok
GenesisDelegationState        MISSING   <- known, #3748
GovPurposeRootsState          ok
```

`GenesisDelegationState` is the only other gap, and it is already tracked in
#3748. It is also the safe kind: `ValidateMIRGenesisQuorum` fails closed with
`GenesisDelegationStateUnavailableError` rather than degrading, and
`UtxoValidateMIRGenesisQuorum` is deliberately not registered in any era's
rules precisely because Dingo does not implement the capability yet. So it
cannot currently affect a sync.

That leaves `EpochState` as the only capability that was both absent and
silently degrading. The class is otherwise complete.

I did not add a permanent test covering the whole table — the probe above was
a one-off. `TestLedgerViewSatisfiesEpochState` pins only this capability. A
table-driven guard over all of them would be reasonable, but it would assert
`GenesisDelegationState` is absent, which is the wrong thing to freeze while
#3748 is open.

chrisguiney:
@coderabbitai full review

ledger/view.go
```diff
@@ -414,6 +414,24 @@ func (lv *LedgerView) PoolCurrentState(
...
+//
+// Several ledger rules are expressed relative to the current epoch and degrade
+// to a weaker check when the ledger state cannot supply one. Without this the
+// pool-deposit decision cannot tell a retired pool from a registered one, so a
+// registration for an already-retired pool is charged no deposit and the
+// transaction fails value conservation by exactly that amount
+// (issue #3908); the retirement-epoch bound on pool retirement certificates is
+// skipped for the same reason.
+func (lv *LedgerView) EpochForSlot(slot uint64) (uint64, error) {
+	epoch, err := lv.ls.epochForSlot(slot)
```
wolf31o2, line 428: gouroboros treats an `EpochForSlot` error as a validation failure, not as a missing capability:

```go
currentEpoch, err := epochState.EpochForSlot(slot)
if err != nil {
	return err
}
```
(`ledger/shelley/rules.go:1059-1063`, v0.202.8). `ls.epochForSlot` errors for any slot the epoch cache does not cover.

Block application is safe: the apply loop breaks out of the batch and runs the rollover before the first block of a new epoch (`ledger/state.go:5781`), and `ValidateTxFunc` is called with `point.Slot`, which the published cache therefore always covers.

The mempool is not. `validateTxCore` and `WithTxValidationSession` pass `snapshot.referenceSlot`, and `validationReferenceSlot` returns the wall-clock slot whenever it leads the tip. Between an epoch boundary and the first applied block of the new epoch, and continuously during sync, that slot is past the last cached epoch. At this head:

```
epochCache = {196 @16934400, 197 @17020800}, 86400 slots each
validationReferenceSlot(tip=17107100, wall=17107200, nil) = 17107200
lv.EpochForSlot(17107200) -> "slot 17107200 not covered by any known epoch
                             (cache has 2 epochs, last ends at slot 17107200)"
```

A transaction carrying a pool retirement certificate is then rejected in that window where it was previously skipped, and cardano-node accepts it. With gouroboros#2207 the same error reaches every pool registration certificate.

cardano-node has no such window: its `EpochInfo` comes from the hard-fork summary (`Ouroboros.Consensus.HardFork.Combinator.State.Infra.reconstructSummary`), so it maps any slot inside the current era's forecast whether or not that epoch has been reached. Extrapolating from the last cached epoch's length within the era, the same in-era projection #3846 adds for slot-to-time, would match; falling the mempool reference slot back to the tip slot would also close it.

Missing test: an accept-case pinning that a pool-retirement transaction validated at a reference slot past the last cached epoch is not rejected.

ledger/view.go
```diff
@@ -414,6 +414,24 @@ func (lv *LedgerView) PoolCurrentState(
...
+// gouroboros' optional common.EpochState capability.
+//
+// Several ledger rules are expressed relative to the current epoch and degrade
+// to a weaker check when the ledger state cannot supply one. Without this the
+// pool-deposit decision cannot tell a retired pool from a registered one, so a
+// registration for an already-retired pool is charged no deposit and the
+// transaction fails value conservation by exactly that amount
+// (issue #3908); the retirement-epoch bound on pool retirement certificates is
+// skipped for the same reason.
+func (lv *LedgerView) EpochForSlot(slot uint64) (uint64, error) {
```
wolf31o2, line 427: `ledger/view.go:90-119` carries `var _ lcommon.X = (*LedgerView)(nil)` for the other five capabilities. `EpochState` is pinned only by a runtime assertion in `view_epoch_for_slot_test.go`, which catches a rename one test run later than a compile-time assertion would.

wolf31o2 (commented):
Not blocking, but one behavioural regression should be settled before merge: enabling the capability turns an `EpochForSlot` error into a transaction rejection, and the mempool's reference slot is routinely outside the epoch cache. Detail inline.

Also note gouroboros#2207 is still open, so merging this alone enables `StakePoolRetirementWrongEpochPOOL` without delivering the pool-deposit fix it targets.

`go build ./...`, `go vet ./ledger/...`, `go test ./ledger/ -run 'EpochForSlot|EpochState'` pass at 176d10be.

wolf31o2 (approved):

PR #3916, opened 2026-09-04
fix(koiosparity): compare only the rewards the ledger credited

### LINKED ISSUE #3915: koiosparity: uncredited rewards are compared against Koios as if paid
## Summary

The per-account Koios parity comparison feeds it every row of
`reward_account_output`, including rewards the ledger never credited. Koios
reports only credited rewards, so each uncredited row becomes a false
`acct_only_dingo` mismatch and fails the epoch.

## Evidence

A from-genesis Preview replay failed parity at epoch 198:

```
koiosparity observer: epoch validation failed network=preview epoch=198
  error="parity FAIL at epoch 198 (12 mismatch(es))"
```

Three `acct_only_dingo` rows, all member rewards from the same pool
`6080b9c76b4b19c21d17b4091aa205417cb80a5449f638c27eec0946`:

```
stake_test1upe2f6j6rd93wqzjufus2kmv9d6hwvqxkyrzcaynwmyadzchwjmvm   69019
stake_test1ur3e9u6ghx8xd2zr39rrhf28ctz4zkrdn9em8h5v3vzy8zqgajtwy  1409915
stake_test1uru2mg4e498am9wnt4yzhhwltfn0lfdnxz6nndrpxf2urhqkyx6vg   69019
```

Koios `/account_info` reports all three as `"status": "not registered"`, with
zero balance and no delegation. Dingo agrees they were never credited — all
three rows carry `spendable = 0`:

```sql
SELECT reward_type, spendable, guarded, count(*)
  FROM reward_account_output WHERE epoch = 197 GROUP BY 1,2,3;

leader|1|0|61
member|0|0|3      <- exactly the three reported
member|1|0|1415
```

The pool's other 1415 member rewards and 61 leader rewards matched. Both sides
are correct about the chain; only the comparison is wrong.

## Cause

`compareEpochAccounts` in `internal/koiosparity/check.go` builds its comparison
rows from every row `GetRewardAccountOutputs` returns, with no crediting
filter:

```go
for _, row := range dingoOutputs {
	addr, addrErr := StakeAddressFromCredential(row.StakingKey, row.CredentialTag)
	...
	dingoRows = append(dingoRows, DingoAccountReward{...})
}
```

`reward_account_output` holds every reward the calculation produced. The
ledger's own application (`ledger/reward_calculation.go`) skips a reward that
is not spendable and one whose reward account is guarded by CIP-0163 expiry,
so neither reaches an account balance.

The pool-level path already gets this right — `addSpendableMemberRewards` in
`internal/koiosparity/dingo_db.go` filters `!spendable || guarded` and carries
a doc comment explaining why. The account-level path added later did not pick
up the same rule.

## Scope

Not specific to epoch 198. Every epoch holding an uncredited row fails:

```sql
SELECT epoch, count(*) FROM reward_account_output
 WHERE spendable = 0 OR guarded = 1 GROUP BY epoch ORDER BY epoch DESC;

220|3   219|3   218|3   197|3   178|1   177|1   176|1   157|1
```

Stake epochs 218/219/220 report as epochs 219/220/221. The replay has
validated through 217, so those three are a standing prediction rather than an
observed result.

## Fix

Filter on crediting, not on reward type — a leader reward is credited to the
pool's reward account and Koios reports it, so it must stay in the comparison.
That is the one way `addSpendableMemberRewards`' rule must not be copied
verbatim: it also filters to member rows, because it is summing a pool's
member reward total specifically.

`accountLifecycleMismatches` deliberately keeps receiving the unfiltered rows.
It diffs "which accounts Dingo computed a reward output for" to detect
registration and deregistration between adjacent epochs, and an uncredited
reward is exactly the deregistration signal it exists to catch.

chrisguiney:
Fixes #3915.

## Why

`compareEpochAccounts` built its comparison rows from every
`reward_account_output` row. That table records every reward the calculation
produced, credited or not — the ledger's own application
(`ledger/reward_calculation.go`) skips a reward that is not spendable and one
whose reward account is guarded by CIP-0163 expiry, so neither reaches an
account balance.

Koios reports credited rewards only, so each uncredited row became a false
`acct_only_dingo` and failed the epoch.

The pool-level path already had this rule — `addSpendableMemberRewards` filters
`!spendable || guarded` and documents why. The account-level path added later
did not pick it up.

## Evidence

A from-genesis Preview replay failed at epoch 198 with three `acct_only_dingo`
rows. Koios `/account_info` reports all three stake addresses as
`"status": "not registered"`, zero balance, no delegation. Dingo agrees they
were never credited:

```
reward_account_output, epoch 197, by (type, spendable, guarded):
  leader | 1 | 0 |   61
  member | 0 | 0 |    3    <- exactly the three reported
  member | 1 | 0 | 1415
```

Both sides were right about the chain. Only the comparison was wrong.

## The filter is on crediting, not on reward type

`addSpendableMemberRewards` additionally filters to member rows because it is
summing a pool's member reward total. Copying that verbatim would be wrong
here: a leader reward *is* credited to the pool's reward account and Koios
reports it. `TestCreditedAccountRewardsKeepsLeaderRewards` pins that.

`accountLifecycleMismatches` still receives the unfiltered rows on purpose. It
diffs which accounts Dingo computed an output for to detect registration and
deregistration between adjacent epochs, and an uncredited reward is exactly the
deregistration signal it exists to catch.

## Tests

Four cases on the extracted helper: uncredited rows dropped (both the
unspendable and the guarded kind), leader rewards kept, a decode failure still
reported, and a decode failure on an *uncredited* row not reported — a row no
comparison would have used must not raise a database error. All four fail
before the change; the helper does not exist.

## Standing prediction

This is not specific to epoch 198. Stake epochs 218/219/220 each hold three
uncredited rows and report as epochs 219/220/221. The replay had validated
through 217 when this was written, so those three are a prediction, not an
observed result. I will confirm or retract it on this PR once the replay
reaches them.

## Validation

`make test` for `internal/koiosparity` passes. `make lint` reports 0
golangci-lint issues across all modules and the Windows target, and
`make docs-parity` passes. `nilaway` is neither pinned by the Makefile nor run
in CI; installed at `@latest` it reports findings in `chainselection/`,
`ledger/governance/` and `ledger/view.go`, none in the changed package and none
in a file this branch touches.

query, storage, or component-boundary change. The behaviour described in
ARCHITECTURE.md's Koios Parity Tracker section is unchanged; this restores the
comparison to what that section already describes.

---
## Summary by cubic
Fixes #3915. The per-account Koios comparison now compares only the rewards the ledger actually credited, so unspendable and CIP-0163-guarded rows no longer show up as false `acct_only_dingo` mismatches.

- Extracts row filtering and decode-error handling into a `creditedAccountRewards` helper.
- Every row is decoded, credited or not, so a corrupt credential on an uncredited row still reports a database error instead of silently disabling the lifecycle diff.
- Leader rewards stay in the comparison since they are credited; `accountLifecycleMismatches` keeps the unfiltered rows for deregistration detection.

<sup>Written for commit 582616a782d8737f1a4fe073ccccd7d9a7764d16. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3916?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected account parity checks so uncredited rewards are no longer reported as mismatches when they are absent from the comparison source.
  * Preserved mismatch reporting when a corresponding credited reward is expected.
  * Improved reporting of invalid reward account credentials, including cases involving uncredited rewards.

* **Tests**
  * Added regression coverage for credited and uncredited rewards, guarded rewards, leader rewards, and credential decoding failures.


chrisguiney:
Real-chain validation, from the replay this was found on.

Built with this change and restarted. Preview stake epoch 301 holds one
uncredited (`spendable = 0`) reward row, which reports as epoch 302 — the same
shape that failed epochs 198, 219, 220, 221 and 228 before the fix, each time
with exactly as many `acct_only_dingo` rows as the stake epoch had uncredited
ones.

Epoch 302 now validates:

```
koiosparity observer: epoch validated network=preview epoch=302
```

The only row recorded for it is one informational `acct_zero_reward`, which
`DetermineStatus` ignores by design.

The replay has since run through epoch 314 with no `acct_only_dingo` at all.

internal/koiosparity/check.go
```diff
@@ -931,6 +919,54 @@ func compareEpochAccounts(
...
+// The filter is on crediting, not on reward type. A leader reward is credited
+// to the pool's reward account and Koios reports it, so it belongs in the
+// comparison — unlike addSpendableMemberRewards, which additionally filters to
+// member rows because it is summing a pool's member reward total specifically.
+//
+// A credential that cannot be decoded is returned as an error for the caller
+// to report, but only for a row that would have been compared: an uncredited
+// row is dropped before decoding, so a bad credential on one cannot raise a
+// database error about a row no comparison would have used.
+func creditedAccountRewards(
```
wolf31o2, line 1017: All four cases assert this helper in isolation; none exercise `compareEpochAccounts`, which is the call site the change altered. For a comparison-narrowing change the case that matters is the one proving the narrowed comparison still fails on a genuine divergence, and it is missing:

- an uncredited Dingo row (`Spendable: false`) whose Koios counterpart exists at the same `(stake_address, reward_type)` must still be reported — as `acct_only_koios`, `StatusFail`.
- the restructured decode path now appends every decode mismatch ahead of the comparison rows instead of interleaving them; nothing pins that `CategoryDBError` still reaches `compareEpochAccounts`'s output at all.

`check_test.go` already drives `Check` end to end against a sqlite Dingo DB and a seeded cache (`TestCheckAccountsCoverageIncompleteIsError`), so both are reachable without new scaffolding.

internal/koiosparity/check.go
```diff
@@ -931,6 +919,54 @@ func compareEpochAccounts(
...
+	outputs []*models.RewardAccountOutput,
+) ([]DingoAccountReward, []error) {
+	rows := make([]DingoAccountReward, 0, len(outputs))
+	var errs []error
+	for _, row := range outputs {
+		if !row.Spendable || row.Guarded {
```
wolf31o2: Dropping before decoding also drops the `CategoryDBError` for an uncredited row with a corrupt credential. Defensible for the comparison, but a credential that cannot be decoded is a storage problem whether or not the row was credited. It is not lost entirely — `accountLifecycleMismatches` still decodes the same row through `dingoRewardAddressSet` and reports its `decodeErrs` — worth saying so in the doc comment, since that is what makes this safe rather than the row simply going quiet.

wolf31o2 (commented):
The narrowing is justified by the ledger, not by convenience. Verified in `ledger/reward_calculation.go`: the crediting loop skips `!reward.Spendable` and `rewardOutputGuarded(app, output)`, and `applyGuardedFlagToAccountOutputs` reconciles the persisted `Guarded` column against that same predicate immediately before crediting — so the filter here is the predicate the ledger credits on, not an approximation of it. `addSpendableMemberRewards` already applies it at pool level.

It is also one-directional and therefore not weaker: it drops only Dingo rows, so a dropped row whose Koios counterpart exists becomes `acct_only_koios` (FAIL) rather than disappearing. Confirmed `accountLifecycleMismatches` still receives the unfiltered `dingoOutputs`.

The standing prediction for stake epochs 218/219/220 is still open on this PR.

internal/koiosparity/check.go
chrisguiney: Valid, and the missing case was the one that mattered. Fixed in 98bec017.

`TestCheckUncreditedDingoRowStillFailsAgainstKoios` drives `Check` end to end against a sqlite Dingo DB and a seeded cache, with an uncredited (`Spendable: false`) Dingo row whose Koios counterpart exists at the same `(stake_address, reward_type)`. It comes out as `acct_only_koios` and fails the epoch — the row moves from the value path to the presence path rather than going quiet.

That test is also the proof for the PR's central change, in the direction you asked for: with the narrowing removed, the uncredited Dingo row matches Koios's amount and the epoch **passes**.

```
# creditedAccountRewards keeps uncredited rows (PR's change reverted)
--- FAIL: TestCheckUncreditedDingoRowStillFailsAgainstKoios
```

`TestCheckAccountDecodeErrorReachesOutput` covers the second half — that a `CategoryDBError` still reaches `compareEpochAccounts`'s output now that decode mismatches are appended ahead of the comparison rows instead of interleaved. Two subtests, credited and uncredited, for the reason in the thread below.

internal/koiosparity/check.go
chrisguiney: Valid — and stronger than "worth documenting", so I changed the code rather than the comment. Fixed in 98bec017.

Checking the claim I would have written down: `accountLifecycleMismatches` reports a `CategoryDBError` only for the **previous** stake epoch's decode failures. For the current epoch it decodes through `dingoRewardAddressSet`, and on failure merely suppresses the lifecycle diff and returns — on the stated assumption that `compareEpochAccounts` already reported it. Dropping the row before decoding broke that assumption in both directions at once: the error vanished entirely, and it silently disabled the lifecycle diff.

So the decode now runs for every row and the credited filter moved after it. The narrowing is about what the comparison sees, not about what gets reported.

```
# filter restored ahead of the decode
--- FAIL: TestCreditedAccountRewardsReportsUncreditedDecodeFailure
--- FAIL: TestCheckAccountDecodeErrorReachesOutput/uncredited_row
```

`TestCreditedAccountRewardsIgnoresUncreditedDecodeFailure` asserted the old behaviour, so it is replaced by `...ReportsUncreditedDecodeFailure`, which still requires the row to stay out of the comparison — only its error is now reported. The doc comment says why, including the `accountLifecycleMismatches` dependency, since that is what makes this the reporting path rather than an arbitrary choice.

chrisguiney:
### CI status: red, and none of it is this change

`main` is red on its own. On unmodified `origin/main` (`90504583`), `go test ./ledger/eras/ -count=1` fails:

```
FAIL	github.com/blinklabs-io/dingo/ledger/eras
```

Tracked in #3976. This branch merged `origin/main` and inherited it, which is why it went from green to red without a source change of its own.

Every failing job on this head fails on exactly those three tests and nothing else — Linux, macOS and Windows all show the same trio and no other failing package:

| job | failing tests |
|---|---|
| go-test (Linux) | the three `ledger/eras` tests |
| go-test (macOS) | the three `ledger/eras` tests |
| go-test (Windows) | the three `ledger/eras` tests |

This branch touches no file under `ledger/` (`git diff --name-only origin/main...HEAD -- ledger/` is empty); it is confined to `internal/koiosparity/`. `./bin` passed on this head, so #3930/#3574 did not show up here.

Locally at this head: `go test ./internal/koiosparity/... -count=1` ok, `make lint` 0 golangci-lint issues, `make docs-parity` ok. `nilaway` could not run — the installed binary is go1.26-built and cannot analyse go1.27.1 — so it is reported unavailable rather than clean.

Not proposing to merge over a red build; that is the maintainer's call.

chrisguiney:
Correction to the CI note above, for the race job specifically. The three non-race jobs fail on exactly the #3976 `ledger/eras` trio. The **race** job adds one more:

```
--- FAIL: TestWindowedRewindConvergesWhilePrimaryChainExtends (0.54s)
FAIL	github.com/blinklabs-io/dingo/ledger	461.269s
```

That one is not this change either. This branch touches no file under `ledger/` (`git diff --name-only origin/main...HEAD -- ledger/` is empty), and the test passes locally under `-race` in isolation at this head:

```
go test ./ledger/ -race -count=1 -run TestWindowedRewindConvergesWhilePrimaryChainExtends
ok  	github.com/blinklabs-io/dingo/ledger	4.821s
```

It failed only inside the 461-second full-suite race run, which is the load-dependent shape #3930/#3574 describe for `./bin`.

So: nothing failing on this head is attributable to this PR, but the build is red for two independent pre-existing reasons. Not merging past that.

wolf31o2 (approved):

PR #3918, opened 2026-09-04
fix(ledger): preserve unknown certificate deposits

### LINKED ISSUE #3829: Certificate deposit recorded as 0 makes value conservation refund 0 instead of KeyDeposit
Three sites record or return a zero deposit where the value is genuinely unknown, and gouroboros treats a returned pointer-to-zero as **authoritative**:

- `ledger/certs.go:44-53` returns `(0, nil)` on `ErrIncompatibleProtocolParams`
- `internal/node/backfill.go:580-596` omits the index on the same error
- `database/plugin/metadata/sqlstore/transaction_certificates.go:104-112` defaults an absent key to 0; its `certificateRequiresDeposit` guard only fires when the whole map is nil

`UtxoValidateValueNotConservedUtxo` refunds a legacy deregistration from the recorded deposit via `StakeCredentialDepositState`, falling back to the current `KeyDeposit` only when the state cannot report a value. A recorded 0 is reported as a value, so the refund becomes 0 rather than `KeyDeposit` — an independent route to "value not conserved" on an otherwise valid transaction.

Found while working #3678; distinct from that rollback defect.

Acceptance criteria: a deposit that cannot be determined is reported as absent rather than 0, so the rule falls back to `KeyDeposit`; a test covers a deregistration whose recorded deposit is unavailable and asserts the transaction is accepted.

wolf31o2:
Fixes #3829

Preserve nullable certificate deposits through metadata writes. Unknown deposits remain SQL NULL, while DRep deregistration refunds use the amount carried by the certificate.

Validation:

- `go test ./database/plugin/metadata/sqlstore -run 'Test.*Certificate|Test.*Deposit|Test' -count=1`


wolf31o2:
Superseded by #3836, which merged the same change: `nullableDecimalUint64` and `*uint64` to distinguish an unknown stake deposit from a recorded zero. #3829 is closed by it.

This branch now has an empty diff against `main` (`git diff origin/main origin/fix/3829-certificate-deposit` produces no output), so there is nothing left to merge.

Worth recording why it was not empty until now. An earlier merge on this branch (`d9017fa6`) resolved `database/plugin/metadata/store.go` and `database/plugin/metadata/sqlstore/transaction_certificates.go` to the branch side, which dropped the ten methods `main` had added to `MetadataStore`, the signed MIR delta write from #3981, and the `certificateRequiresDeposit` guard. The result did not compile — `ledger`, `ledgerstate` and `internal/koiosparity` all failed to build against it. Both files have been restored to `main`'s version in `07f760cf`.

The only delta unique to this branch was removing the `certificateRequiresDeposit` guard, which #3829 describes as a safety net rather than a target, so nothing is lost by closing.

PR #3920, opened 2026-09-04
fix(koiosparity): report the mismatch count that caused the failure

### LINKED ISSUE #3917: koiosparity: failure message counts mismatches that cannot cause a failure
## Summary

The parity failure message counts every mismatch, including the categories
`DetermineStatus` deliberately treats as no-ops. The number therefore does not
describe the failure.

## Evidence

```
koiosparity observer: epoch validation failed network=preview epoch=198
  error="parity FAIL at epoch 198 (12 mismatch(es))"
```

The twelve were:

```
acct_only_dingo   3    <- the only ones that can cause a FAIL
acct_zero_reward  1    informational
pool_departed     8    informational
```

`observer.go` reports `len(result.Mismatches)`, but `DetermineStatus` reaches
`StatusFail` only through its `default` branch — `CategoryAcctZeroReward`,
`CategoryAcctNewlyRegistered`, `CategoryAcctDeregistered` and
`CategoryPoolDeparted` are explicitly excluded, with a comment saying a
lifecycle mismatch "must never turn an otherwise-clean epoch into ERROR or
FAIL".

Both behaviours are individually correct; together they produce a count that
contradicts the decision it accompanies. Investigating #3915 I spent time on
eight `pool_departed` rows before realising they can never be a failure.

## Fix

Report the count that drove the status. The status decision and the count
should read one shared classification rather than two copies of the category
list, so a category added to one cannot be forgotten by the other.

chrisguiney:
Fixes #3917.

## Why

`observer.go` reported `len(result.Mismatches)` with a parity failure, but
`DetermineStatus` reaches `StatusFail` only through its `default` branch —
`acct_zero_reward`, `acct_newly_registered`, `acct_deregistered` and
`pool_departed` are explicitly excluded, with a comment saying a lifecycle
mismatch must never turn an otherwise-clean epoch into ERROR or FAIL.

So the count contradicted the decision it accompanied:

```
parity FAIL at epoch 198 (12 mismatch(es))
```

was 3 account mismatches, 1 zero-reward summary, and 8 pool departures. The
nine informational rows can never cause a failure. Investigating #3915 I
worked through the eight departures before noticing they were not the reason.

## The change

`severityOf` now holds the category classification once. `DetermineStatus` and
the new `CountSignificant` both read it. The message becomes:

```
parity FAIL at epoch 198 (3 significant of 12 mismatch(es))
```

Both numbers are kept: the total still says how much the epoch produced, which
is worth seeing, and the leading number now says how much of it mattered.

The `len(mismatches) == 0` early return in `DetermineStatus` is gone as
redundant, not as a behaviour change — an empty loop leaves `hasError` false
and returns `StatusPass`.

## Tests

`TestCountSignificantExcludesInformational` reconstructs epoch 198's exact
mix and pins 3 rather than 12. `TestCountSignificantCountsErrors` keeps the
error categories significant, since they drive `StatusError`.

`TestCountSignificantAgreesWithDetermineStatus` is the one that matters most:
for every category, a `StatusPass` and a non-zero significant count must not
coexist in either direction. That is the invariant the two copies of the
category list could violate, and it is what makes the shared `severityOf`
load-bearing rather than cosmetic.

## Validation

`make test` for `internal/koiosparity` passes, including the existing
`DetermineStatus` tests. `make lint` reports 0 golangci-lint issues across all
modules and the Windows target; `make docs-parity` passes.

query, storage, or component-boundary change, and the category semantics
ARCHITECTURE.md describes are unchanged.

---
## Summary by cubic
Fixes the parity failure message to report only the mismatches that caused the failure, not the total count. The old message said "12 mismatch(es)" even when only 3 were significant; now it says "3 significant of 12 mismatch(es)".

- Introduces a shared `mismatchSeverity` classification and an `AllCategories` list, both guarded by tests so a category can never be classified differently by the status and the count, or omitted from either.
- Removes the redundant empty-slice early return in `DetermineStatus`; behavior is unchanged.
- Adds an observer-level test pinning the message an operator reads, not just the `CountSignificant` arithmetic.

<sup>Written for commit 7b8f22bc91e52ebf4eed024ba43c09ad58ef521d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3920?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/koiosparity/compare.go
+++ b/internal/koiosparity/compare.go
@@ -869,20 +869,6 @@ func rationalsEqual(a, b string) bool {
 	return ra.Cmp(&rb) == 0
 }
 
-// DetermineStatus returns PASS, FAIL, or ERROR from a list of mismatches.
-//
-//   - FAIL: any value_mismatch, pool_only_dingo, pool_only_koios,
-//     acct_only_dingo, acct_only_koios, or acct_duplicate entry — a strict
-//     run cannot report PASS while any of these are present (#3097's
-//     acceptance criteria: a single missing, extra, duplicate, or differing
-//     account fails the whole epoch).
-//   - ERROR: only DB-level failures (dingo_db_error, dingo_db_missing),
-//     reference_lag (Koios data may be incomplete for a recent epoch), or
-//     acct_coverage_incomplete (the Koios account-reward fetch for this
-//     epoch never completed across every chunk, so there is no complete
-//     reference set to compare against yet).
-//   - PASS: no mismatches.
-//
 // mismatchSeverity classifies one mismatch category. DetermineStatus and
 // CountSignificant both read it, so a category added to one can never be
 // forgotten by the other — the failure the split invited was a count that did
@@ -937,6 +923,19 @@ func CountSignificant(mismatches []CheckMismatch) int {
 	return n
 }
 
+// DetermineStatus returns PASS, FAIL, or ERROR from a list of mismatches.
+//
+//   - FAIL: any value_mismatch, pool_only_dingo, pool_only_koios,
+//     acct_only_dingo, acct_only_koios, or acct_duplicate entry — a strict
+//     run cannot report PASS while any of these are present (#3097's
+//     acceptance criteria: a single missing, extra, duplicate, or differing
+//     account fails the whole epoch).
+//   - ERROR: only DB-level failures (dingo_db_error, dingo_db_missing),
+//     reference_lag (Koios data may be incomplete for a recent epoch), or
+//     acct_coverage_incomplete (the Koios account-reward fetch for this
+//     epoch never completed across every chunk, so there is no complete
+//     reference set to compare against yet).
+//   - PASS: no mismatches.
 func DetermineStatus(mismatches []CheckMismatch) string {
 	hasError := false
 	for _, m := range mismatches {
--- a/internal/koiosparity/compare.go
+++ b/internal/koiosparity/compare.go
@@ -83,6 +83,31 @@ const (
 	CategoryAcctDeregistered = "acct_deregistered"
 )
 
+// AllCategories is every mismatch category above, in one place.
+//
+// severityOf classifies exactly these values, and the tests that guard the
+// classification iterate this slice rather than restating it. A hand-written
+// second copy is how a category comes to be classified by DetermineStatus and
+// missed by CountSignificant (or the reverse) with every test still green, so
+// the list exists once and TestAllCategoriesCoversEveryConstant pins it to the
+// const block above.
+var AllCategories = []string{
+	CategoryValueMismatch,
+	CategoryPoolOnlyDingo,
+	CategoryPoolOnlyKoios,
+	CategoryReferenceLag,
+	CategoryDBError,
+	CategoryDBMissing,
+	CategoryPoolDeparted,
+	CategoryAcctOnlyDingo,
+	CategoryAcctOnlyKoios,
+	CategoryAcctDuplicate,
+	CategoryAcctCoverageIncomplete,
+	CategoryAcctZeroReward,
+	CategoryAcctNewlyRegistered,
+	CategoryAcctDeregistered,
+}
+
 // Epoch check status values.
 const (
 	StatusPass  = "PASS"
--- a/internal/koiosparity/compare.go
+++ b/internal/koiosparity/compare.go
@@ -882,28 +882,70 @@ func rationalsEqual(a, b string) bool {
 //     epoch never completed across every chunk, so there is no complete
 //     reference set to compare against yet).
 //   - PASS: no mismatches.
-func DetermineStatus(mismatches []CheckMismatch) string {
-	if len(mismatches) == 0 {
-		return StatusPass
+//
+// mismatchSeverity classifies one mismatch category. DetermineStatus and
+// CountSignificant both read it, so a category added to one can never be
+// forgotten by the other — the failure the split invited was a count that did
+// not agree with the status it accompanied.
+type mismatchSeverity int
+
+const (
+	// severityInformational describes state rather than disagreement. These
+	// must never turn an otherwise-clean epoch into ERROR or FAIL, and must
+	// never be counted as a reason for one.
+	severityInformational mismatchSeverity = iota
+	// severityError means the comparison could not be trusted, not that it
+	// disagreed.
+	severityError
+	// severityFail is a real Dingo/Koios disagreement.
+	severityFail
+)
+
+func severityOf(category string) mismatchSeverity {
+	switch category {
+	case CategoryDBError,
+		CategoryDBMissing,
+		CategoryReferenceLag,
+		CategoryAcctCoverageIncomplete:
+		return severityError
+	case CategoryAcctZeroReward,
+		CategoryAcctNewlyRegistered,
+		CategoryAcctDeregistered,
+		CategoryPoolDeparted:
+		// Purely informational — see these categories' doc comments.
+		return severityInformational
+	default:
+		return severityFail
 	}
+}
+
+// CountSignificant returns how many mismatches drove the status DetermineStatus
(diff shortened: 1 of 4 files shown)


internal/koiosparity/compare.go
```diff
@@ -882,28 +882,70 @@ func rationalsEqual(a, b string) bool {
...
 //   - PASS: no mismatches.
-func DetermineStatus(mismatches []CheckMismatch) string {
-	if len(mismatches) == 0 {
-		return StatusPass
+//
+// mismatchSeverity classifies one mismatch category. DetermineStatus and
+// CountSignificant both read it, so a category added to one can never be
+// forgotten by the other — the failure the split invited was a count that did
+// not agree with the status it accompanied.
+type mismatchSeverity int
```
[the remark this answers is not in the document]
chrisguiney: Valid, fixed in the latest push.

The type declaration landed between the doc block and `DetermineStatus`, so godoc attributed the PASS/FAIL/ERROR description to `mismatchSeverity` and left the exported function undocumented. The block is back immediately above `DetermineStatus`; `go doc` now shows it there.

Auditing the class rather than the line: I made the same insertion in #3924, where the zero-amount helper orphaned the same doc block. Fixed there too.

chrisguiney:
Exercised on the replay this came from. The failure message now reads:

```
koiosparity observer: epoch validation failed network=preview epoch=282
  error="parity ERROR at epoch 282 (1 significant of 2 mismatch(es))"
```

The second row is an informational `acct_zero_reward`; the single significant
one is the `dingo_db_missing` that actually caused the ERROR. Before this
change the same epoch would have reported "2 mismatch(es)", pointing at a row
that cannot cause a failure.

internal/koiosparity/observer.go
```diff
@@ -477,9 +477,10 @@ func (o *Observer) processEpoch(ctx context.Context, epoch uint64) {
 		o.cfg.OnResult(result)
 	}
 	if result.Status != StatusPass {
+		significant := CountSignificant(result.Mismatches)
 		o.fail(epoch, fmt.Errorf(
-			"parity %s at epoch %d (%d mismatch(es))",
-			result.Status, epoch, len(result.Mismatches),
+			"parity %s at epoch %d (%d significant of %d mismatch(es))",
```
wolf31o2, line 482: This is the user-visible fix and nothing asserts it — `grep 'significant of' internal/koiosparity/*_test.go` is empty. A test over `processEpoch`'s error text would fail without the change; the three new tests would not.

internal/koiosparity/severity_test.go
```diff
@@ -0,0 +1,75 @@
...
+	assert.Equal(t, StatusError, DetermineStatus(mismatches))
+}
+
+// TestCountSignificantAgreesWithDetermineStatus is the invariant that matters
+// more than either number: a status of PASS and a non-zero significant count
+// cannot coexist, in either direction. The two must read the same
+// classification, or a future category added to one will silently disagree
+// with the other.
+func TestCountSignificantAgreesWithDetermineStatus(t *testing.T) {
+	for _, cat := range []string{
```
wolf31o2: This is a third hand-maintained copy of the category list — the duplication `severityOf` was introduced to remove. A category added to neither `severityOf` nor this slice is invisible to the guard.

Concretely: #3924 adds `CategoryAcctZeroRewardRow` and conflicts with this PR inside `DetermineStatus`. Whoever resolves that conflict has to remember to put the new category in `severityOf`'s informational case, and if they do not, this test still passes — it never sees the category.

Deriving the loop from one exported slice of all categories (with `severityOf` switching over the same slice's members) is what makes the guard load-bearing.

wolf31o2 (commented):
Reporting only, no comparison narrowed. Checked the property that matters: a non-PASS status implies a non-zero significant count, since `DetermineStatus` leaves `StatusPass` only through `severityFail` or `severityError` and `CountSignificant` counts both. Dropping the `len(mismatches) == 0` early return is behaviour-preserving.

The same misleading total survives in `report.go`'s verbose per-epoch line (`epoch %d: %d mismatches`, from `st.MismatchCount`) and in the persisted `check_epoch_status.MismatchCount`. Out of scope for the reported symptom, but the reader hits it in the JSON report too.

internal/koiosparity/observer.go
chrisguiney: Valid, and the sharper of the two — thank you. Fixed in 7b8f22bc.

`TestObserverFailureReportsSignificantMismatchCount` drives the observer end to end to a FAIL and asserts the text `FatalFunc` receives. The epoch holds one account value mismatch (the reason) and one `acct_zero_reward` row from a second address in the requested universe that neither side pays, so the two numbers are 1 and 2 and cannot coincide. The test asserts that separation explicitly (`require.Greater(len(result.Mismatches), CountSignificant(...))`) before asserting the text, so it cannot silently degrade into a tautology if the scenario changes.

Proven against the old message:

```
Error "koios parity: epoch 5: parity FAIL at epoch 5 (2 mismatch(es))" does not
contain "parity FAIL at epoch 5 (1 significant of 2 mismatch(es))"
```

Note for the reader: `acct_newly_registered` is not reachable here. `accountLifecycleMismatches` skips the lifecycle diff entirely for a `*DatabaseSource`, which is what the in-process observer uses, so the informational row had to come from the zero-reward half.

internal/koiosparity/severity_test.go
chrisguiney: Valid — the guard was restating exactly the thing it was meant to protect. Fixed in 7b8f22bc.

`AllCategories` now lives beside the const block as the single list, `severityOf` classifies its members, and this loop is `for _, cat := range AllCategories`. That closes the case you named: whoever resolves the #3924 conflict adds `CategoryAcctZeroRewardRow` to the constants, and the guard sees it whether or not they remember `severityOf`.

The list itself could still fall behind, so `TestAllCategoriesCoversEveryConstant` parses the package source (`go/ast`, same shape as `ledger/publish_under_lock_test.go`) and requires the set of `Category*` constants and `AllCategories` to be equal in both directions — a constant missing from the slice, and a slice entry that is not a constant. Dropping `CategoryAcctZeroReward` from the slice:

```
constant CategoryAcctZeroReward ("acct_zero_reward") is missing from
AllCategories, so severityOf's classification of it is unguarded
```

And the loop is now load-bearing where the hand-written one was not. The old list omitted `value_mismatch`, `pool_only_dingo` and `pool_only_koios`, so a `CountSignificant` that stopped counting `value_mismatch` passed it. Against `AllCategories`:

```
--- FAIL: TestCountSignificantAgreesWithDetermineStatus/value_mismatch
    status and significant count must classify "value_mismatch" the same way
```

chrisguiney:
Both inline findings addressed in 7b8f22bc; replies inline with the mutations that prove each new test.

On the review body's third point — `report.go`'s verbose per-epoch line and the persisted `check_epoch_status.MismatchCount` carry the same total — agreed that the reader hits it there too, and deliberately left alone here. `MismatchCount` is a persisted column and a documented JSON field (`mismatch_count`), so redefining it would silently change the meaning of rows already written by earlier runs with nothing in the row to say which meaning it carries. That wants its own change, and probably a second field rather than a redefinition. Filed as #3966.

Validation on 7b8f22bc: `go test ./internal/koiosparity/... -count=1` ok, `make lint` 0 golangci-lint issues (nilaway unavailable — the installed build is go1.26 and cannot analyse the go1.27.1 toolchain), `make docs-parity` ok.

wolf31o2 (approved):

PR #3923, opened 2026-09-04
fix(ledgerstate): preserve imported reward seed failures

### LINKED ISSUE #3313: ledgerstate: one pool with no reward account discards a whole epoch of reward seeding, and the resulting skip is misreported as expected after a Mithril bootstrap
**Version:** `devel (commit 5a54b7f8)`, equal to `origin/main`. Network preview, fresh database, Mithril v2 bootstrap, 22.7-hour run that otherwise synced to tip and stayed there.

### Summary

`bundle.validate()` rejects an entire epoch's reward-input bundle when any single derived pool input is incomplete, and the caller only warns and moves on. Two of the three epochs in one Mithril import were discarded this way, each on one pool. One epoch later the reward round for the discarded epoch found no snapshot and skipped, and the skip was reported as an expected consequence of the bootstrap rather than as the seeding failure it actually was.

Net effect: a Mithril-bootstrapped node silently under-credits rewards for the affected epochs, and the log actively points the operator away from the cause.

### What happened

Snapshot: ledger slot 120648841, epoch 1396, 3156972 UTxOs, cert state parsed clean.

```
imported stake snapshot snapshot=mark stored_as=mark target_epoch=1396 pools=642
imported stake snapshot snapshot=set  stored_as=mark target_epoch=1395 pools=647
imported stake snapshot snapshot=go   stored_as=mark target_epoch=1394 pools=642
imported active pool distribution snapshot=pool-distr stored_as=actv target_epoch=1396 pools=647
synthesized retired-but-scheduled pools from active pool distribution count=5 epoch=1396
```

Seeding results, one line each:

```
WARN not seeding reward inputs for an imported epoch: the derived basis does not reconcile,
     so that epoch's reward round will be skipped and its rewards never credited
     epoch=1396 snapshot=mark
     error="derived pool input for 102e9ff50bee440b1ef337f58d1760a5475f3ce716f2aab60e6ef424 has no reward account"

INFO seeded reward inputs for an imported epoch epoch=1395 snapshot=set pools=741 delegators=79744

WARN not seeding reward inputs for an imported epoch: ...
     epoch=1394 snapshot=go
     error="derived pool input for 1fc372fdce61f42d31be7ddfc2bf8e343b08a54e4d3e6d64b2e328ff has no reward account"
```

Two of three epochs discarded. The one that succeeded seeded 741 pools and 79744 delegators, so each discard threw away a comparable amount of correct data over one bad pool.

Ten hours later, at the 1396 to 1397 boundary:

```
INFO handling epoch transition previous_epoch=1396 new_epoch=1397 boundary_slot=120700800
WARN skipping stake rewards: missing reward snapshot; this epoch's rewards will never be
     credited, leaving reward balances and the leadership stake distribution permanently short
     (expected after a Mithril bootstrap, whose preceding epochs this node never saw)
     new_epoch=1397 reward_snapshot_epoch=1394
```

`reward_snapshot_epoch=1394` is the epoch whose seeding was discarded above. `ledger/reward_calculation.go:188` calls `GetRewardSnapshot(1394, "mark")`, gets nil, and skips.

### The all-or-nothing validate

`ledgerstate/imported_reward_inputs.go:282`

```go
if len(pool.RewardAccount) == 0 {
    return fmt.Errorf(
        "derived pool input for %x has no reward account",
        pool.PoolKeyHash,
    )
}
```

`validate()` returns on the first offending pool, so the error names one pool and the caller never learns how many were bad.

`ledgerstate/imported_reward_inputs.go:463`

```go
if err := bundle.validate(); err != nil {
    if logger != nil {
        logger.Warn("not seeding reward inputs for an imported epoch: ...")
    }
    continue
}
```

The whole bundle is dropped. Nothing retries, nothing records that this epoch is now unseedable, and the process continues to the next epoch.

### Candidate source of the incomplete pools

`ledgerstate/import.go:1913` synthesizes pools that appear in the active pool distribution but not in the registered-pool set, with only a key hash and a VRF key hash:

```go
reg := &models.PoolRegistration{
    PoolKeyHash: slices.Clone(pool.PoolKeyHash),
    VrfKeyHash:  slices.Clone(pool.VrfKeyHash),
    AddedSlot:   slot,
}
```

No reward account, margin, cost, or pledge. They are tombstoned with `RetirePools` immediately afterwards specifically so they are "excluded from every active-pool, stake and reward query", but the reward-input derivation is evidently still reaching pools with an empty reward account. The arithmetic lines up: pool-distr carries 647 pools, the mark and go snapshots carry 642, and exactly 5 were synthesized.

That is a hypothesis, not a conclusion. Whether `102e9ff5...` and `1fc372fd...` are among those 5 is checkable directly on the import and would settle it. If they are, the fix is either to exclude tombstoned pools from the derivation or to carry enough registration data on the synthesized rows to satisfy validation.

### The misattribution

The skip message asserts a cause that is wrong in this case:

> (expected after a Mithril bootstrap, whose preceding epochs this node never saw)

The node did see epoch 1394. It imported its stake snapshot with 642 pools and then discarded the derived reward inputs. An operator reading only the boundary message concludes the missing rewards are inherent to bootstrapping and stops. The two warnings that explain it were emitted ten hours earlier during import.

The message should distinguish "no snapshot was ever imported for this epoch" from "a snapshot was imported and its reward inputs failed to reconcile".

### Asks

- Do not discard a whole epoch over one pool. At minimum report every offending pool and the stake they represent, so the blast radius is visible.
- Establish whether tombstoned synthesized pools reach the derivation, and exclude them if so.
- Make the seeding failure durable rather than a warning that scrolls past, so the boundary skip can name the real reason.
- Correct the boundary message so it does not attribute a seeding failure to the bootstrap.

wolf31o2:
Retain imported reward-basis seed failures by epoch and snapshot.
Report the durable import cause when a later reward round lacks its snapshot.
Clear markers after successful seeding and remove rolled-back markers.
Cover stored-cause and generic missing-snapshot behavior.

Closes #3313.

## Summary by CodeRabbit

- **New Features**
  - Imported reward-basis seeding failures are now recorded with diagnostic details.
  - Missing reward snapshots distinguish between genuinely missing data and failed imported seeding.
  - Successful seeding clears previous failure records, and rollback removes records for reverted state.
  - Failure diagnostics are bounded for readability.

- **Bug Fixes**
  - Reward-processing warnings now report the specific imported seeding failure when available.

- **Documentation**
  - Updated architecture and database documentation for the new reward-seeding failure tracking.


wolf31o2:
Fixed in 0e02501e (additive, signed off).

The Windows failure was the pre-existing root-package test TestLiveTruncateCancelsWhenStorageProviderDrainIsUnconfirmed: its 5s Eventually callback could block on a Badger reopen and retried after a successful reopen when runtime.Close returned an error, producing the hosted Condition never satisfied failure. The probe now stops after the first successful reopen and asserts the separate cleanup error directly. Reward-seeding code and tests are unchanged.

Validation: focused test passed 3/3 locally; GOOS=windows CGO_ENABLED=0 go test -exec=true -run ^ -timeout=10m ./... passed (compile-only; Windows binaries cannot execute on this Linux host).,

wolf31o2:
Addressed both Cubic findings in additive commit 3abbfe6e (DCO-signed):

- SaveRewardSeedFailure now preserves the earliest captured_slot with MIN(existing, incoming), and rollback coverage proves a later re-import cannot erase the earlier durable failure.
- Empty derived reward bases now report referenced pools as having no parameters; the separate bundle-validation path still reports missing reward accounts when pool rows exist without accounts.

Validation: go test -count=1 ./database/plugin/metadata/sqlstore ./ledgerstate passed. make sql-check regenerated the expected query output; generated SQL is included in the commit. Hosted checks for the new head have not started beyond DCO yet.,

database/plugin/metadata/sqlstore/queries/sqlite/operational.sql
```diff
@@ -466,6 +466,23 @@ SELECT id, epoch, snapshot_type, total_active_stake, total_pool_count,
 FROM reward_snapshot
 WHERE epoch = ? AND snapshot_type = ?;
 
+-- name: SaveRewardSeedFailure :exec
+INSERT INTO reward_seed_failure (
+    epoch, snapshot_type, failure_reason, captured_slot
+) VALUES (?, ?, ?, ?)
+ON CONFLICT (epoch, snapshot_type) DO UPDATE SET
+    failure_reason = excluded.failure_reason,
+    captured_slot = MIN(reward_seed_failure.captured_slot, excluded.captured_slot);
```
arepala-uml, line 479: Confirmed — this reproduces as a real bug, not a hypothetical.

The upsert pins `captured_slot` to the earliest write (`MIN(...)`) but always replaces `failure_reason` with the latest write's text. When two writes for the same `(epoch, snapshot_type)` carry different reasons at different slots, a rollback to a slot between them checks only the retained (earliest) `captured_slot`, finds nothing to delete, and leaves the later reason in place — describing state that no longer exists post-rollback. This contradicts the invariant this PR documents in ARCHITECTURE.md/DATABASE.md: "rollback removes markers above its slot, so the message cannot outlive the imported state it describes."

Repro test (added to `reward_provenance_test.go`, run against `3abbfe6e`, then reverted — not included in this PR):

```go
func TestRewardSeedFailureReasonDoesNotSurviveRollbackPastItsSlot(t *testing.T) {
	t.Parallel()
	store := newManagementTestStore(t)
	require.NoError(t, store.SaveRewardSeedFailure(
		20, "mark", "reason-A", 100, nil,
	))
	require.NoError(t, store.SaveRewardSeedFailure(
		20, "mark", "reason-B", 300, nil,
	))
	require.NoError(t, store.DeleteRewardStateAfterSlot(200, nil))
	reason, err := store.GetRewardSeedFailure(20, "mark", nil)
	require.NoError(t, err)
	require.NotEqual(t, "reason-B", reason,
		"reason-B was only true as of slot 300; rolling back to slot 200 "+
			"must not leave it in place")
}
```

Result:

```
--- FAIL: TestRewardSeedFailureReasonDoesNotSurviveRollbackPastItsSlot (0.04s)
    reward_provenance_test.go:262: 
        Error:      	Should not be: "reason-B"
        Messages:   	reason-B was only true as of slot 300; rolling back to slot 200 must not leave it in place
FAIL
```

Suggest keeping `failure_reason` and `captured_slot` from the same write — e.g. only overwrite both together when the incoming slot is >= the stored slot — then regenerate `operational.sql.go` and land a rollback test with this shape (distinct reasons at different slots).

arepala-uml (changes requested):
Requesting changes: the reward_seed_failure upsert can leave a stale failure reason after a rollback, contradicting this PR's own documented rollback invariant. Reproduced with a failing test (see inline comment) — the same defect CodeRabbit already flagged on this head.

wolf31o2:
Addressed the requested review fixes in `7970d7d0`:

- Preserve the earliest `captured_slot` together with its matching failure reason during upsert and rollback.
- Strengthen the rollback regression test with distinct reasons and reversed slot order; the old SQL fails the test.
- Remove the unrelated lifecycle-test change.
- Correct the retained-row cardinality documentation.

Validation: affected `sqlstore` and `ledgerstate` packages pass with `-race`; `make sql-check` and `make docs-parity` pass.

database/plugin/metadata/sqlstore/queries/sqlite/operational.sql
arepala-uml: Verified fixed in 7970d7d0. The upsert now keeps `failure_reason` and `captured_slot` from the same write (`CASE WHEN excluded.captured_slot < reward_seed_failure.captured_slot THEN excluded.failure_reason ELSE reward_seed_failure.failure_reason END`), so the row always reflects one consistent (reason, slot) pair from whichever write is earliest.

Re-ran my exact repro test against this head: save(20,"mark","reason-A",100), save(20,"mark","reason-B",300), rollback to slot 200 — `GetRewardSeedFailure` now correctly returns "reason-A" (slot 100 write), not the later "reason-B". PASS.

Also re-ran `TestRewardSeedFailureRoundTripAndRollback`, the full `RewardSeed*`/`SeedImportedRewardInputs*`/`MissingRewardSnapshotReportsImportedSeedFailure` suites, and `make sql-check` — all green. The reordered writes in `TestRewardSeedFailureRoundTripAndRollback` (50 before 200) now actually discriminate the fix, addressing Cubic's separate test-rigor comment on this thread as well.

Resolving.

database/plugin/metadata/sqlstore/internal/query/sqlite/operational.sql.go
```diff
@@ -4169,6 +4212,32 @@ func (q *Queries) SaveRewardPoolOutput(ctx context.Context, arg SaveRewardPoolOu
 	return id, err
 }
 
+const saveRewardSeedFailure = `-- name: SaveRewardSeedFailure :exec
+INSERT INTO reward_seed_failure (
+    epoch, snapshot_type, failure_reason, captured_slot
+) VALUES (?, ?, ?, ?)
+ON CONFLICT (epoch, snapshot_type) DO UPDATE SET
+    failure_reason = excluded.failure_reason,
+    captured_slot = MIN(reward_seed_failure.captured_slot, excluded.captured_slot)
```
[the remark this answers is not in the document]
arepala-uml: Verified fixed in 7970d7d0 — the upsert now keeps `failure_reason` and `captured_slot` from the same write (CASE on `excluded.captured_slot < reward_seed_failure.captured_slot`), so rollback can no longer retain a reason from a later, already-rolled-back write. Confirmed with a focused repro test (distinct reasons at different slots, rollback between them) and the full reward-seed-failure/seeding test suite; `make sql-check` is clean.

arepala-uml (approved):

PR #3924, opened 2026-09-04
fix(koiosparity): treat a zero-value reward row as agreement

### LINKED ISSUE #3942: koiosparity: a zero-value reward row on one side is reported as a divergence
## Summary

The Koios parity checker reports a reward row worth zero that exists on only
one side as `acct_only_koios` — a FAIL category — even though nothing was
credited on either side and no lovelace differs.

## Evidence

A from-genesis Preview replay failed epoch 222 on a single row:

```
acct_only_koios  account_reward_presence
  stake_test1uzf5lwsf37wsxmq9rdpq0v9tepk0g36vqmxr974lenzwchcszrsss
  dingo=""   koios="0 (type=leader)"
```

Epoch 255 failed identically on a different address.

## Cause

`CategoryAcctZeroReward`'s own doc comment states the premise backwards — that
the confirmed-zero case arises "since Koios never emits a row for zero
reward". Koios does:

```sql
SELECT epoch, reward_type, count(*) FROM koios_account_rewards
 WHERE CAST(earned AS INTEGER) = 0 GROUP BY epoch, reward_type;

222|leader|1   155|leader|1   154|leader|1   152|leader|1
150|leader|1   149|leader|1   148|leader|1   147|leader|2   ...
```

Dingo, conversely, writes no `reward_account_output` row at all for a zero
reward — its table holds no zero-amount row in any epoch. So the two sides
represent "credited nothing" differently, and `CompareAccountEpoch`'s presence
test reads that as a reward Dingo missed.

This is distinct from the existing zero-reward concept, which keys off
`reward_row_count = 0` — Koios returning *no* rows for an account. Here Koios
returns one row whose amount is zero.

## Scope

Recurring, not a one-off: zero-earned leader rows appear across epochs 145-155
and at 222 and 255, and the same pool's stake address repeats across
consecutive epochs, so a long-lived pool earning a zero leader reward fails
every epoch it appears in.

## Fix

Classify a one-sided reward row worth zero informationally rather than as a
presence divergence, in both directions, and correct the false premise in the
category documentation.

chrisguiney:
Fixes #3942

## Why

A Koios reward row worth zero with no Dingo counterpart was reported as
`acct_only_koios` — a FAIL category — and failed Preview epoch 222:

```
acct_only_koios  account_reward_presence
  stake_test1uzf5lwsf37wsxmq9rdpq0v9tepk0g36vqmxr974lenzwchcszrsss
  dingo=""   koios="0 (type=leader)"
```

Nothing was credited on either side. No lovelace differs.

The comparison's own documentation had the premise backwards. `CategoryAcctZeroReward`'s
doc comment states the confirmed-zero case arises "since Koios never emits a
row for zero reward". Koios does:

```sql
SELECT epoch, reward_type, count(*) FROM koios_account_rewards
 WHERE CAST(earned AS INTEGER) = 0 GROUP BY epoch, reward_type;

222|leader|1   155|leader|1   154|leader|1   152|leader|1
150|leader|1   149|leader|1   148|leader|1   147|leader|2   ...
```

Dingo, conversely, writes no `reward_account_output` row at all for a zero
reward — its table holds no zero-amount row in any epoch. So the two sides
represent "credited nothing" differently, and the presence test read that as a
reward Dingo had missed.

This is distinct from the existing zero-reward concept, which keys off
`reward_row_count = 0` — Koios returning *no* rows for an account. Here Koios
returns one row whose amount is zero. I corrected that doc comment too.

## The change

A one-sided reward row worth zero is classified informationally instead of as
a presence divergence, in both directions. It is reported rather than dropped,
so the representational difference stays visible.

The amount is parsed rather than compared against the literal `"0"`: the two
sides format independently, and an unparseable amount is deliberately *not*
treated as zero — that is a real value the comparison must keep reporting.

The Dingo-only direction produces no such rows today. It is handled anyway
because the two presence branches are deliberately symmetric.

## Tests

Zero-earned Koios row is informational and passes; the Dingo mirror likewise;
a one-lovelace Koios-only row still fails, which is the discrimination that
keeps this from silencing one-sided rows generally; and a zero on both sides
reports nothing at all, so the handling cannot start manufacturing rows for
agreeing pairs.

## Validation

Found by a from-genesis Preview replay under the parity observer. `make test`
for `internal/koiosparity` passes; `make lint` reports 0 golangci-lint issues
across all modules and the Windows target; `make docs-parity` passes.

query, storage, or component-boundary change. The new category is internal to
the parity checker's mismatch classification, which those documents describe
only at the level this change preserves.

## Merge and CI state

`origin/main` merged in (`4005f91a`), not rebased, to resolve the conflict with #3920's shared severity classification: both `severityOf` and `isZeroRewardAmount` are retained, and `CategoryAcctZeroRewardRow` is registered in `severityOf`'s informational case **and** in `AllCategories` — the guard test caught the second omission.

`go-test` is red, for a reason outside this PR. Three tests in `ledger/eras` — `TestConwayUtxoValidationRuleIndexesArePinned`, `TestValidateTxConwayGenuinelyMissingInputStillRejected` and `TestValidateTxConwayGenuinelyUnbalancedStillRejected` — fail on all four platforms at `main`'s own head (`90504583`), which is the commit merged here, and this branch has no diff against main under `ledger/eras`, `go.mod` or `go.sum`. Tracked in **#3976**. `./bin`'s two entrypoint tests also fail under full-suite load on unmodified `main` (#3930/#3574).

Locally: `go test ./internal/koiosparity/... -count=1` passes, `make lint` reports 0 golangci-lint issues across all four scopes, `make docs-parity` passes. `nilaway` could not run — the pinned binary is go1.26-built and fails prerequisites against the installed go1.27.1 toolchain while still exiting 0, so it is reported as unavailable rather than clean.

---
## Summary by cubic
Treats a one-sided reward row worth zero as agreement instead of a presence divergence, so Preview epoch 222 no longer fails on rows where both sides credited nothing.

- Reports zero-earned Koios-only rows (and the symmetric Dingo-only case) as informational `acct_zero_reward_row`; this outranks the grace-window reference-lag fallback, and nonzero one-sided rows still fail.
- Parses amounts through one shared `parseLovelace` for both presence and value checks, so the same string can't be read as zero by one and malformed by the other; unparseable amounts are never treated as zero.
- Registers the category in `AllCategories` and `severityOf`, which the merge with main's shared severity classification required.
- Corrects the `CategoryAcctZeroReward` doc comment and `ARCHITECTURE.md` (Koios does emit zero-reward rows; Dingo writes none) and keeps `DetermineStatus`'s doc block attached after the helper insertion.

<sup>Written for commit d711bfb427dd8742ae047bbe637edabf29546167. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3924?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Zero-reward account rows are now treated as informational rather than reported as discrepancies.
  - Nonzero account-only rows continue to be reported as divergences.
  - Matching zero-reward rows no longer produce unnecessary comparison results.

- **Tests**
  - Added coverage for zero-reward rows from either source, matching zero rows, and nonzero discrepancies.

### DIFF
--- a/internal/koiosparity/compare.go
+++ b/internal/koiosparity/compare.go
@@ -19,7 +19,6 @@ import (
 	"math/big"
 	"slices"
 	"strconv"
-	"strings"
 	"time"
 )
 
@@ -867,14 +866,50 @@ func lovelaceEqual(a, b string) bool {
 	// string-equality short-circuit would report two identical malformed or
 	// negative strings as "equal" without ever validating them, letting
 	// CompareAccountEpoch pass on invalid account data.
-	var x, y big.Int
-	if _, ok := x.SetString(a, 10); !ok || x.Sign() < 0 {
+	x, ok := parseLovelace(a)
+	if !ok {
 		return false
 	}
-	if _, ok := y.SetString(b, 10); !ok || y.Sign() < 0 {
+	y, ok := parseLovelace(b)
+	if !ok {
 		return false
 	}
-	return x.Cmp(&y) == 0
+	return x.Cmp(y) == 0
+}
+
+// parseLovelace is the single definition of a well-formed lovelace amount:
+// one or more ASCII digits, nothing else. No sign, no surrounding whitespace,
+// no separators.
+//
+// It exists so that lovelaceEqual and isZeroRewardAmount cannot disagree about
+// what a string means. They read the same field from the same two sides, and a
+// string one of them accepts while the other rejects gets two verdicts from the
+// same input: with a divergent parse, " 0" was agreement when the row was
+// one-sided and value_mismatch when both sides had one, and "+0" was the
+// reverse. Whether a given malformed spelling ought to be tolerated is a
+// separate question from whether the two paths answer it the same way; this
+// answers the second, strictly, so a malformed amount is always reported and
+// never waived.
+//
+// Deliberately stricter than big.Int.SetString alone, which accepts a leading
+// sign: "-0" parses to zero with a non-negative sign, so a sign check does not
+// exclude it, and a negative lovelace amount is malformed data rather than a
+// zero reward. Leading zeros are accepted ("00" is zero) because the two sides
+// format independently and a value's spelling is not the comparison's business.
+func parseLovelace(s string) (*big.Int, bool) {
+	if s == "" {
+		return nil, false
+	}
+	for _, r := range s {
+		if r < '0' || r > '9' {
+			return nil, false
+		}
+	}
+	var v big.Int
+	if _, ok := v.SetString(s, 10); !ok {
+		return nil, false
+	}
+	return &v, true
 }
 
 // rationalsEqual reports whether two numeric strings represent the same
@@ -894,11 +929,15 @@ func rationalsEqual(a, b string) bool {
 //
 // Parsed rather than compared to "0": the two sides format independently, and
 // a reward that is genuinely zero must be recognised as zero however it is
-// spelled. An unparseable amount is not zero — it is a real value the
-// comparison must keep reporting rather than quietly waive.
+// spelled, so "00" is zero. An unparseable amount is not zero — it is a real
+// value the comparison must keep reporting rather than quietly waive, so "",
+// "abc", " 0" and "-0" all stay one-sided rows and keep failing the epoch.
+//
+// The parse is parseLovelace, the same one lovelaceEqual uses, so a string
+// cannot be zero here and malformed there.
 func isZeroRewardAmount(amount string) bool {
-	v, err := strconv.ParseUint(strings.TrimSpace(amount), 10, 64)
-	return err == nil && v == 0
+	v, ok := parseLovelace(amount)
+	return ok && v.Sign() == 0
 }
 
 // DetermineStatus returns PASS, FAIL, or ERROR from a list of mismatches.
--- a/internal/koiosparity/compare.go
+++ b/internal/koiosparity/compare.go
@@ -890,6 +890,17 @@ func rationalsEqual(a, b string) bool {
 	return ra.Cmp(&rb) == 0
 }
 
+// isZeroRewardAmount reports whether a lovelace decimal string is zero.
+//
+// Parsed rather than compared to "0": the two sides format independently, and
+// a reward that is genuinely zero must be recognised as zero however it is
+// spelled. An unparseable amount is not zero — it is a real value the
+// comparison must keep reporting rather than quietly waive.
+func isZeroRewardAmount(amount string) bool {
+	v, err := strconv.ParseUint(strings.TrimSpace(amount), 10, 64)
+	return err == nil && v == 0
+}
+
 // DetermineStatus returns PASS, FAIL, or ERROR from a list of mismatches.
 //
 //   - FAIL: any value_mismatch, pool_only_dingo, pool_only_koios,
@@ -903,18 +914,6 @@ func rationalsEqual(a, b string) bool {
 //     epoch never completed across every chunk, so there is no complete
 //     reference set to compare against yet).
 //   - PASS: no mismatches.
-//
-// isZeroRewardAmount reports whether a lovelace decimal string is zero.
-//
-// Parsed rather than compared to "0": the two sides format independently, and
-// a reward that is genuinely zero must be recognised as zero however it is
-// spelled. An unparseable amount is not zero — it is a real value the
-// comparison must keep reporting rather than quietly waive.
-func isZeroRewardAmount(amount string) bool {
-	v, err := strconv.ParseUint(strings.TrimSpace(amount), 10, 64)
-	return err == nil && v == 0
-}
-
 func DetermineStatus(mismatches []CheckMismatch) string {
 	if len(mismatches) == 0 {
 		return StatusPass
--- a/internal/koiosparity/compare.go
+++ b/internal/koiosparity/compare.go
@@ -693,14 +693,26 @@ type accountRewardKey struct {
 // is not masked by the duplicate report.
 //
 // The union of the two (deduplicated) keysets is then walked: present only
-// in Koios -> CategoryAcctOnlyKoios (or CategoryReferenceLag within
-// graceHours of epochEndTime, mirroring ComparePoolEpoch's identical
-// pattern), present only in Dingo -> CategoryAcctOnlyDingo, present on both
-// sides with differing amounts -> CategoryValueMismatch (compared as exact
-// integers via lovelaceEqual, never as floats/rationals), present on both
-// sides with equal amounts -> no mismatch. A zero-reward account present
-// identically on both sides is therefore a pass, exactly like any other
-// equal-amount case.
+// in Koios -> CategoryAcctOnlyKoios, present only in Dingo ->
+// CategoryAcctOnlyDingo, present on both sides with differing amounts ->
+// CategoryValueMismatch (compared as exact integers via lovelaceEqual, never
+// as floats/rationals), present on both sides with equal amounts -> no
+// mismatch. A zero-reward account present identically on both sides is
+// therefore a pass, exactly like any other equal-amount case.
+//
+// Two things reclassify a one-sided row before it is emitted, in this order:
+//
+//   - The row is worth zero. The two sides then agree on what was credited
+//     (nothing) and disagree only about whether to store a row saying so, so
+//     it is CategoryAcctZeroRewardRow — informational — rather than either
+//     acct_only_* category. This takes precedence over the grace window
+//     below, because a zero row is not a value the other side can still
+//     publish later.
+//   - The check is inside graceHours of epochEndTime. Koios can lag in
+//     publishing /account_reward_history for a just-closed epoch, and Dingo
+//     can commit ahead of it, so a nonzero one-sided row is
+//     CategoryReferenceLag until the window closes — mirroring
+//     ComparePoolEpoch's identical pattern.
 //
 // graceHours/epochEndTime/now/network/epoch all mirror ComparePoolEpoch's
 // identical parameters and meaning.
--- a/internal/koiosparity/compare.go
+++ b/internal/koiosparity/compare.go
@@ -19,6 +19,7 @@ import (
 	"math/big"
 	"slices"
 	"strconv"
+	"strings"
 	"time"
 )
 
@@ -64,14 +65,25 @@ const (
 	// (dingo #3099) report the three account dimensions #3097's merged
 	// comparison structurally cannot: CompareAccountEpoch only ever compares
 	// keys present in at least one side's row map, so an address absent from
-	// both (a confirmed-zero-reward account, since Koios never emits a row
-	// for zero reward) never enters that comparison at all, and #3097's
+	// both (a confirmed-zero-reward account, meaning Koios returned no rows
+	// for it at all — distinct from Koios returning a row whose amount is
+	// zero, which it does emit and which CategoryAcctZeroRewardRow covers)
+	// never enters that comparison at all, and #3097's
 	// address universe is a single flat list reused across every epoch in one
 	// run, with no per-epoch persisted snapshot to diff for lifecycle
 	// changes. All three are purely informational — descriptive state, not a
 	// Dingo-vs-Koios discrepancy — and must never affect Status (see
 	// DetermineStatus's dedicated no-op case for these three).
 	CategoryAcctZeroReward = "acct_zero_reward"
+	// CategoryAcctZeroRewardRow marks a reward row worth zero that exists on
+	// only one side. It corrects the premise stated above: Koios does emit a
+	// row for a zero reward — Preview publishes zero-earned leader rows —
+	// while Dingo writes no reward_account_output row at all in that case.
+	// Nothing is credited either way, so the two agree about every lovelace
+	// and the one-sided row is a representational difference, not a
+	// divergence. Purely informational, and reported rather than dropped so
+	// the difference stays visible.
+	CategoryAcctZeroRewardRow = "acct_zero_reward_row"
 	// CategoryAcctNewlyRegistered marks a stake address present in this
 	// stake epoch's Dingo-committed reward_account_output universe but
 	// absent from the previous stake epoch's — see
@@ -769,8 +781,13 @@ func CompareAccountEpoch(
 		switch {
 		case koiosOK && !dingoOK:
 			cat := CategoryAcctOnlyKoios
-			if graceHours > 0 && !epochEndTime.IsZero() &&
-				now.Sub(epochEndTime) < time.Duration(graceHours)*time.Hour {
+			switch {
+			case isZeroRewardAmount(kr.Earned):
+				// Both sides credited nothing; see
+				// CategoryAcctZeroRewardRow.
+				cat = CategoryAcctZeroRewardRow
+			case graceHours > 0 && !epochEndTime.IsZero() &&
+				now.Sub(epochEndTime) < time.Duration(graceHours)*time.Hour:
 				cat = CategoryReferenceLag
 			}
 			out = append(out, CheckMismatch{
@@ -795,8 +812,12 @@ func CompareAccountEpoch(
 			// hasn't published yet within graceHours is reference lag, not
 			// a real acct_only_dingo discrepancy.
 			cat := CategoryAcctOnlyDingo
-			if graceHours > 0 && !epochEndTime.IsZero() &&
-				now.Sub(epochEndTime) < time.Duration(graceHours)*time.Hour {
+			switch {
+			case isZeroRewardAmount(dr.Amount):
+				// Symmetric with the koiosOK && !dingoOK case above.
+				cat = CategoryAcctZeroRewardRow
+			case graceHours > 0 && !epochEndTime.IsZero() &&
+				now.Sub(epochEndTime) < time.Duration(graceHours)*time.Hour:
 				cat = CategoryReferenceLag
 			}
 			out = append(out, CheckMismatch{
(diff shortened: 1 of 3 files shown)


chrisguiney:
Integration note for whoever sequences these.

Merging this alongside the other open koiosparity changes conflicts in
`compare.go`, in three places. All three are textual overlap on the same
region, not disagreement about behaviour, and I resolved them in the local
verification build:

- **#3920** replaces `DetermineStatus`'s inline category switch with a shared
  `severityOf`. This PR adds a category to that same list. Resolution: add
  `CategoryAcctZeroRewardRow` to `severityOf`'s informational case.
- **#3853** adds a `rewardsPending` reference-lag branch to both presence
  cases, and an eighth `rewardsPending` parameter to `CompareAccountEpoch`.
  This PR adds a zero-value branch to those same two switches, and its tests
  call the seven-parameter signature. Resolution: order the zero check first
  (a zero row is agreement regardless of lag), then `rewardsPending`, then the
  grace window; and pass `false` in this PR's tests.

Whichever lands second needs that resolution. The merged result passes the
package's tests.

internal/koiosparity/compare.go
```diff
@@ -882,6 +903,18 @@ func rationalsEqual(a, b string) bool {
...
 //     reference set to compare against yet).
 //   - PASS: no mismatches.
+//
+// isZeroRewardAmount reports whether a lovelace decimal string is zero.
+//
+// Parsed rather than compared to "0": the two sides format independently, and
+// a reward that is genuinely zero must be recognised as zero however it is
+// spelled. An unparseable amount is not zero — it is a real value the
+// comparison must keep reporting rather than quietly waive.
+func isZeroRewardAmount(amount string) bool {
```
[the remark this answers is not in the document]
chrisguiney: Valid, and already fixed in the latest push — I found it auditing the class after the same finding on #3920, where the severity type orphaned this same doc block.

The doc block is back immediately above `DetermineStatus` and `go doc` attributes it correctly.

chrisguiney:
Status of live validation, so it is not overstated: **this fix has not been
exercised on the replay yet.**

It is built into the running node, but Koios has published no zero-earned
reward row for any epoch between 256 and 315, so the branch this changes has
not been reached since the restart. The evidence for the fix remains epoch 222
and epoch 255 under the previous build, plus the unit tests.

The category is real and recurring — zero-earned leader rows appear across
epochs 145-155 and at 222 and 255 — so I expect it to recur. I will report it
here when the replay hits one.

chrisguiney:
Correcting my previous comment: **this is now validated on real data.** The
replay reached the case shortly after I wrote that it had not.

Eight zero-earned Koios leader rows, across three different stake addresses:

```
epoch 327  stake_test1ur0paqh4t5cwgflh4phdjkhvpwdvv3ps6zd8l7cv9ffk20qxjhqwf  0 (type=leader)
epoch 328  (same)
epoch 329  (same)
epoch 330  (same)
epoch 344  stake_test1uzulxfzccykl6xl659glxnfrvkr9jr3dled8g50m4gfe2lg6a5wde  0 (type=leader)
epoch 345  stake_test1ur7dav7tz75n7qd4p5fx2su0m5vdu6udvpn2aw4gusx536sag3auf  0 (type=leader)
epoch 347  (same)
epoch 348  (same)
```

Each was classified `acct_zero_reward_row` and each Dingo side is empty, which
is the exact shape that failed epochs 222 and 255 as `acct_only_koios` before
this change.

Seven of the eight epochs validated:

```
koiosparity observer: epoch validated network=preview epoch=327
koiosparity observer: epoch validated network=preview epoch=328
koiosparity observer: epoch validated network=preview epoch=329
koiosparity observer: epoch validated network=preview epoch=330
koiosparity observer: epoch validated network=preview epoch=345
koiosparity observer: epoch validated network=preview epoch=347
koiosparity observer: epoch validated network=preview epoch=348
```

Epoch 344 failed, but not on this: `parity ERROR at epoch 344 (1 significant
of 3 mismatch(es))`, where the one significant mismatch is the
`dingo_db_missing` pool-departure classification tracked in #3925. The
zero-earned row there is among the two informational ones.

It also confirms the recurrence pattern — the same pool's stake address
repeats across consecutive epochs, so a single long-lived pool earning a zero
leader reward would have failed every epoch it appeared in.

internal/koiosparity/compare.go
```diff
@@ -869,6 +890,17 @@ func rationalsEqual(a, b string) bool {
 	return ra.Cmp(&rb) == 0
 }
 
+// isZeroRewardAmount reports whether a lovelace decimal string is zero.
+//
+// Parsed rather than compared to "0": the two sides format independently, and
+// a reward that is genuinely zero must be recognised as zero however it is
+// spelled. An unparseable amount is not zero — it is a real value the
+// comparison must keep reporting rather than quietly waive.
+func isZeroRewardAmount(amount string) bool {
```
wolf31o2, line 1034: The tests cannot fail on either claim this helper's doc comment makes. Replacing the body with `return strings.TrimSpace(amount) == "0"` leaves the whole `internal/koiosparity` suite green: every case in `zero_reward_row_test.go` uses the literal `"0"` or `"1"`, so neither "parsed rather than compared to the literal" nor "an unparseable amount is deliberately not zero" is pinned.

Missing cases, on `CompareAccountEpoch` with a Koios-only row: `Earned: ""`, `Earned: "abc"`, `Earned: "-0"` must still be `acct_only_koios`/`StatusFail`; `Earned: "00"` must be `acct_zero_reward_row`.

Separately, agreeing with the open Cubic thread: `lovelaceEqual` uses `big.Int.SetString` with no trimming, so `" 0"` is agreement when the row is one-sided and `value_mismatch` when both sides have one — the same input, two verdicts. `"+0"` goes the other way (`SetString` accepts it, `ParseUint` does not).

wolf31o2 (commented):
The narrowing holds: a zero-lovelace Koios row and an absent Dingo row both assert "nothing credited", and the key is (address, reward_type), so a Koios zero against a Dingo nonzero still lands in the value path as `value_mismatch`. `TestNonZeroKoiosOnlyRowStillFails` pins the discrimination.

This is the only one of the five koiosparity PRs that can turn FAIL into PASS, so the test gap below matters more here than elsewhere. Two notes inline, plus: ARCHITECTURE.md's "Mismatch categories" list (~6072) enumerates every category and does not gain `acct_zero_reward_row`, and ~6557 still states "any `acct_only_dingo`/`acct_only_koios`/`acct_duplicate` forces FAIL", which is no longer true for a zero-earned row. `make docs-parity` does not read that prose.

Conflicts with #3920 in `DetermineStatus` and with #3853 in the `koiosOK && !dingoOK` branch (both confirmed by merge-tree).

internal/koiosparity/compare.go
chrisguiney: Both valid, and they turned out to be the same defect. Fixed in 341ca346.

The helper's two doc claims are now pinned at `CompareAccountEpoch`'s verdict rather than on the helper alone. `TestZeroRewardRowAmountSpellings` covers exactly the cases you named — `""`, `"abc"`, `"-0"` and `"+0"` stay `acct_only_koios`/FAIL, `"00"` is `acct_zero_reward_row`. Against `return strings.TrimSpace(amount) == "0"`:

```
--- FAIL: TestZeroRewardRowAmountSpellings/leading_zeros
--- FAIL: TestZeroRewardRowAmountSpellings/padded_zero
```

On the Cubic thread: agreed, and the fix is the shared parse rather than a second trim. `parseLovelace` — digits only, no sign, no surrounding whitespace — is now the single definition of a well-formed lovelace amount, and both `lovelaceEqual` and `isZeroRewardAmount` call it. That is what removes the two-verdicts-for-one-input property in both directions you identified:

- reverting only `isZeroRewardAmount` to `ParseUint(TrimSpace(...))`: `presence and value paths must read " 0" the same way`
- reverting only `lovelaceEqual` to bare `big.Int.SetString` with a sign check: `presence and value paths must read "-0" the same way` and the same for `"+0"`

`TestZeroRewardRowAgreesWithValueComparison` is that property directly, so it fails on either half alone. Note `"-0"` needed the `lovelaceEqual` half too: `SetString` parses it to zero with `Sign() == 0`, so the existing non-negative check did not exclude it and the old code called `"-0"` equal to `"0"`.

Deliberately strict rather than tolerant. Which malformed spellings ought to be accepted is a separate question from whether the two paths answer it the same way; this answers the second, and a malformed amount stays reported rather than waived. Existing behaviour on real data is unchanged — Koios returns plain decimal strings and Dingo formats with `strconv.FormatUint` — and `lovelaceEqual`'s own tests (`"01000000"` == `"1000000"`, `"0"` == `"00"`, `"-5"` rejected) still hold.

ARCHITECTURE.md: `acct_zero_reward_row` added to the "Mismatch categories" list, the "any `acct_only_dingo`/`acct_only_koios`/`acct_duplicate` forces FAIL" sentence corrected, the shared parse recorded next to `lovelaceEqual`, and the zero-reward reporting paragraph now distinguishes "Koios emitted no row" from "Koios emitted a row worth zero".

internal/koiosparity/compare.go
```diff
@@ -882,6 +903,18 @@ func rationalsEqual(a, b string) bool {
...
 //   - PASS: no mismatches.
+//
+// isZeroRewardAmount reports whether a lovelace decimal string is zero.
+//
+// Parsed rather than compared to "0": the two sides format independently, and
+// a reward that is genuinely zero must be recognised as zero however it is
+// spelled. An unparseable amount is not zero — it is a real value the
+// comparison must keep reporting rather than quietly waive.
+func isZeroRewardAmount(amount string) bool {
+	v, err := strconv.ParseUint(strings.TrimSpace(amount), 10, 64)
```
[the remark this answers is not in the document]
chrisguiney: Valid, and fixed in 341ca346 — though the fix went the other way round from "reject whitespace here".

You are right that the trim was the problem, but trimming was only half of it: the divergence was that `isZeroRewardAmount` and `lovelaceEqual` had two different parses of the same field, so `" 0"` was agreement on a one-sided row and `value_mismatch` when both sides carried it, and `"+0"` was the reverse (`big.Int.SetString` accepts a leading sign, `ParseUint` does not). Fixing one side's tolerance would have left the other asymmetry in place.

`parseLovelace` is now the single definition — one or more ASCII digits, nothing else — and both call it. `" 0"`, `"+0"`, `"-0"`, `""` and `"abc"` are malformed in both paths and keep failing the epoch; `"00"` is zero in both. `TestZeroRewardRowAgreesWithValueComparison` pins the equivalence itself, and fails if either side is reverted.

chrisguiney:
Both inline findings addressed in 341ca346, replies inline with the mutations that prove each test.

On the review body's remaining note about conflicts, unchanged and worth flagging rather than resolving here:

- **#3920** touches the same `DetermineStatus` region. #3920 replaces the hand-written category switch with `severityOf`, and it now also introduces an exported `AllCategories` slice that the classification guard iterates. Whichever of the two lands second must add `CategoryAcctZeroRewardRow` to **both** `severityOf`'s informational case and `AllCategories`; #3920's `TestAllCategoriesCoversEveryConstant` fails if the constant is added without the slice entry, so the resolution is guarded in that direction.
- **#3853** adds the `rewardsPending` parameter to `CompareAccountEpoch` and touches the same `koiosOK && !dingoOK` branch. The two are independent in intent — #3853 adds a `reference_lag` fallback, this adds a zero check ahead of it — but they edit adjacent lines and the tests here call `CompareAccountEpoch` with the 7-argument signature.

Validation on 341ca346: `go test ./internal/koiosparity/... -count=1` ok, `make lint` 0 golangci-lint issues (nilaway unavailable — the installed build is go1.26 and cannot analyse the go1.27.1 toolchain), `make docs-parity` ok.

internal/koiosparity/compare.go
```diff
@@ -769,8 +780,13 @@ func CompareAccountEpoch(
 		switch {
 		case koiosOK && !dingoOK:
 			cat := CategoryAcctOnlyKoios
-			if graceHours > 0 && !epochEndTime.IsZero() &&
-				now.Sub(epochEndTime) < time.Duration(graceHours)*time.Hour {
+			switch {
+			case isZeroRewardAmount(kr.Earned):
```
wolf31o2, line 825: Non-blocking: `CompareAccountEpoch`'s own doc comment (lines 666-674) still enumerates the presence outcomes as "present only in Koios -> CategoryAcctOnlyKoios (or CategoryReferenceLag within graceHours of epochEndTime), present only in Dingo -> CategoryAcctOnlyDingo", with no zero-value case. Same stale-prose class as the ARCHITECTURE.md `forces FAIL` sentence corrected in 341ca346, one level down.

wolf31o2 (dismissed):

chrisguiney:
Merged `origin/main` in to resolve the conflict with the shared severity classification (#3920), pushed as `4005f91a`. That invalidated your approval on `341ca346`, so this needs another look — sorry for the churn.

The resolution keeps both sides: `severityOf` retains its informational case and gains `CategoryAcctZeroRewardRow`, and `isZeroRewardAmount` sits alongside it.

Worth recording that **your `AllCategories` guard test earned its keep here.** My first resolution classified the new category in `severityOf` but did not register it in `AllCategories`, and `TestAllCategoriesCoversEveryConstant` failed with exactly the right message:

```
constant CategoryAcctZeroRewardRow ("acct_zero_reward_row") is missing from
AllCategories, so severityOf's classification of it is unguarded
```

That is the precise drift the shared classification exists to prevent, and it was introduced by a merge rather than by either branch — neither PR was wrong on its own. Nothing else would have caught it.

Package tests pass, `make lint` reports 0 issues across all four scopes, `make docs-parity` passes.

internal/koiosparity/compare.go
chrisguiney: Fixed in `cc6893ce` — you are right, and it was the same class one level down.

The enumeration described the pre-fix classification: `acct_only_koios` / `acct_only_dingo` with a `reference_lag` fallback, and no mention that the zero test now runs first in both one-sided branches and takes precedence over the grace window. The doc comment now spells out both reclassifications and their order, with the reason the zero test wins: a zero row is not a value the other side can still publish later, so the grace window has nothing to wait for.

`ARCHITECTURE.md` had the matching gap in the grace-window bullet (it said both presence directions "both fall back to `reference_lag` within the grace window", with the zero-row exception stated only in the separate strict-mode bullet); that bullet now carries the precedence note too.

chrisguiney:
CI status, with the evidence, since the red `go-test` is not this PR's.

All four `go-test` jobs fail on exactly and only three tests, in `ledger/eras`:

- `TestConwayUtxoValidationRuleIndexesArePinned`
- `TestValidateTxConwayGenuinelyMissingInputStillRejected`
- `TestValidateTxConwayGenuinelyUnbalancedStillRejected`

Nothing else fails on any platform. That is the `main` breakage tracked in #3976: the same three tests fail on unmodified `origin/main` at `90504583` — the commit merged into this branch — on all four platforms, and this branch has no diff against main under `ledger/eras`, `go.mod` or `go.sum`.

I checked platform by platform rather than assuming, because an earlier snapshot showed Windows and macOS red with Linux still pending, which looked platform-specific. It is not: Linux and Linux/race fail identically. One extra macOS failure in the previous run (`midnight/server` `TestReflectionDisabledByDefault`, EOF) did not recur at this head.

Local validation on this branch: `go test ./internal/koiosparity/... -count=1` passes, `make lint` reports 0 golangci-lint issues across all four scopes including the Windows target, `make docs-parity` passes. `nilaway` could not run — the pinned binary is go1.26-built, fails prerequisites against the installed go1.27.1 toolchain, and still exits 0, so I am reporting it unavailable rather than clean.

Bot review: cubic reviewed this head (`cc6893ce`) with 0 issues. CodeRabbit reports "Review skipped: incremental reviews are disabled" rather than a completed review.

Not merging: `go-test` is red, so the "all checks pass" condition is not met on the letter of the rule regardless of the cause, and this head has no human approval — the `main` merge invalidated the earlier one. @wolf31o2, re-review requested.

chrisguiney:
Correction to my previous comment, and the resolution.

I wrote that all four `go-test` jobs fail on exactly and only the three `ledger/eras` tests. That was true for Linux, Linux/race and macOS, but not for Windows: that job did not fail on a test at all. It ran to `mithril`, produced no further output for 79 minutes, and was killed by the job's 90-minute `timeout-minutes` backstop. I read the run-level `failure` and the other three jobs and did not check the Windows job's own conclusion, which was `cancelled`.

Re-ran that job alone on the same commit. It completed in 15m54s and failed on exactly and only the three `ledger/eras` tests, with `ouroboros`, `peergov`, `plugin` and `topology` — the packages that follow `mithril`, and so the candidates for the wedge — all passing.

So the original claim now holds on all four platforms, but it holds because of the re-run, not because of the first attempt.

Supporting evidence that the wedge was not this PR's: `internal/koiosparity`, the only package this PR changes, passed on Windows in the wedged run itself (148.879s), well before the output stopped. #3940's Windows job completed the full suite in 13m36s at essentially the same base, and `main`'s own Windows job at `90504583` completed in 12 minutes — both failing only on `ledger/eras`. No `-timeout 30m` panic ever reached the log either, which points at a wedge outside a running test rather than a hung one.

Nothing about the merge condition changes: `go-test` is still red on all four platforms for #3976, and this head still has no human approval.

wolf31o2 (approved):

PR #3926, opened 2026-09-04
fix(forging): reject era-incompatible opcert counters before forging

### LINKED ISSUE #3501: forging: apply genesis and era limits to KES validation
Forging key validation does not carry all genesis and era context used by
startup checks. The live forge path can accept a key state beyond the applicable
evolution limit or with incomplete era validation, then fail during block
application.

Acceptance criteria:

- Retain the applicable genesis evolution limit and era context in the forge path.
- Reject expired or era-incompatible key state before block adoption.
- Cover startup, period updates, era changes, and boundary values.

arepala-uml:
Closes #3501.

The forge loop validates the operational certificate's KES evolution
limit on every attempt, but the OpCert issue-number counter was only
checked once, against the ledger, at node startup
(`PoolCredentials.ValidateAgainstLedger`). A counter that went stale or
became gapped after startup — e.g. a peer relays a block from the same
pool with a higher counter, or the chain crosses the Babbage boundary —
was only caught when block application rejected the block the forger
had already built, signed, and fenced against the slot.

## Changes

- **`ledger/forging/keys.go`**: add `validateOpCertSequence`, mirroring
  `validateOpCertCounter` in `ledger/verify_opcert.go`. A counter behind
  the last value the ledger observed for the pool is rejected in every
  era; one that skips ahead of it is rejected only in Praos eras
  (Babbage onward).
- **`ledger/forging/forger.go`**: add optional `ForgerConfig.OpCertLedgerView`
  and `EraParams` fields. `NewBlockForger` requires `EraParams` whenever
  `OpCertLedgerView` is set. `checkAndForgeProduction` runs the new
  `checkOpCertSequence` gate — resolving the counter via
  `LedgerView.LatestOpCertSequence` and the era via
  `ProtocolParamsForSlot` — before leader selection and the forge-slot
  fence. Leaving `OpCertLedgerView` nil skips the check, unchanged from
  before.
- **`node_forging.go`**: wire the new fields to the node's real ledger
  state in `initBlockForger`.
- **`ARCHITECTURE.md`**: document the new pre-flight gate in the Block
  Forging section.
- **`ledger/forging/keys_test.go`**: table-driven tests for
  `validateOpCertSequence` (stale, boundary, and gapped counters, split
  by era).
- **`ledger/forging/opcert_sequence_gate_test.go`**: end-to-end tests
  through `BlockForger.checkAndForgeProduction` — stale counter, no
  prior observed counter, an era-unresolvable slot, the TPraos/Praos
  no-gap rule split (including boundary values), and confirmation the
  gate is a no-op when `OpCertLedgerView` is nil.

## Testing

- `go test -race ./ledger/forging/...`
- `go test ./internal/test/conformance/...`
- `go build ./...`, `go vet ./...`
- `golangci-lint run ./...`, `nilaway ./...`, `modernize ./...` (no new
  findings in touched files)

---
## Summary by cubic
Fixes the forge loop so a stale or gapped operational certificate counter is caught on every forge attempt instead of only at startup, preventing a leader slot from being spent on a block the chain would reject.

**Details**
- The counter rule lives in one shared implementation (`eras.ValidateOpCertCounter`) used by block application, the forge gate, and startup validation; it also fixes a max-`uint64` wraparound and rejects typed-nil protocol-parameters pointers.
- Startup validation deliberately stays on the stale-only half; applying the era-scoped gap rule there pairs a wall-clock era with an applied-chain baseline and can falsely reject a pool, preventing the node from ever syncing.
- The gate runs after the leader check, so its ledger read only costs slots this pool leads.
- `LedgerState.LatestOpCertSequence` now resolves the latest counter via the same Mithril-boundary-aware path as block application, instead of a plain `MAX` that could trust stale pre-boundary rows.

Closes #3501.

<sup>Written for commit df5d1c0029aa52e3dc3c143ad56d4e6a26c4b3e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3926?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Block forging now checks operational certificate counters against the latest ledger state before attempting to forge.
  - Stale counters are rejected, while counter gaps follow the rules for the active protocol era.
  - Forging safely stops when the applicable era cannot be determined.

- **Reliability**
  - The check is optional and remains skipped when ledger-backed certificate data is unavailable.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -4341,6 +4341,27 @@ The start, expiry, current-period, and remaining-period gauges use that same
 protocol lifetime; the KES key's `2^depth` cryptographic capacity remains a
 separate upper bound rather than an operational lifetime.
 
+Immediately after that gate, when `ForgerConfig.OpCertLedgerView` is wired
+(`NewBlockForger` requires `EraParams` alongside it), the runtime gate
+pre-flights the opcert's issue-number counter against
+`LedgerView.LatestOpCertSequence` using the same era-scoped rule block
+application enforces at apply time (`validateOpCertCounter` in
+`ledger/verify_opcert.go`, mirrored here as `validateOpCertSequence`): a
+counter behind the last value the ledger observed for this pool is always
+rejected (stale or stolen hot key), and one that skips ahead of it is
+additionally rejected in Praos eras (Babbage onward) but accepted in TPraos
+eras (Shelley-Alonzo). `EraParams.ProtocolParamsForSlot` resolves the era for
+the slot being forged. Startup validation and the KES-lifetime gate above
+cover genesis-derived evolution limits, which are fixed for the life of the
+chain; this covers the era-scoped counter rule and the on-chain observed
+counter, both of which can change after startup as blocks are applied (this
+node's own or a peer's for the same pool) or as the chain crosses an era
+boundary. Catching a rule block application would enforce anyway here, before
+leader selection and the forge-slot fence, means a bad key state costs a
+could-not-forge disposition instead of a burned leader slot and a rejected
+`AddLocalBlock` call. The check is opt-in: a nil `OpCertLedgerView` (dev mode,
+embedders without ledger wiring) skips it entirely, unchanged from before.
+
 Each production forge attempt takes an independently owned snapshot of one
 complete credential generation at the runtime gate. The snapshot deep-copies
 the VRF secret, KES secret, verification keys, opcert, and validated lifetime;
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -4341,26 +4341,28 @@ The start, expiry, current-period, and remaining-period gauges use that same
 protocol lifetime; the KES key's `2^depth` cryptographic capacity remains a
 separate upper bound rather than an operational lifetime.
 
-Immediately after that gate, when `ForgerConfig.OpCertLedgerView` is wired
-(`NewBlockForger` requires `EraParams` alongside it), the runtime gate
-pre-flights the opcert's issue-number counter against
-`LedgerView.LatestOpCertSequence` using the same era-scoped rule block
-application enforces at apply time (`validateOpCertCounter` in
-`ledger/verify_opcert.go`, mirrored here as `validateOpCertSequence`): a
-counter behind the last value the ledger observed for this pool is always
-rejected (stale or stolen hot key), and one that skips ahead of it is
-additionally rejected in Praos eras (Babbage onward) but accepted in TPraos
-eras (Shelley-Alonzo). `EraParams.ProtocolParamsForSlot` resolves the era for
-the slot being forged. Startup validation and the KES-lifetime gate above
-cover genesis-derived evolution limits, which are fixed for the life of the
-chain; this covers the era-scoped counter rule and the on-chain observed
-counter, both of which can change after startup as blocks are applied (this
-node's own or a peer's for the same pool) or as the chain crosses an era
-boundary. Catching a rule block application would enforce anyway here, before
-leader selection and the forge-slot fence, means a bad key state costs a
-could-not-forge disposition instead of a burned leader slot and a rejected
-`AddLocalBlock` call. The check is opt-in: a nil `OpCertLedgerView` (dev mode,
-embedders without ledger wiring) skips it entirely, unchanged from before.
+When `ForgerConfig.OpCertLedgerView` is wired (`NewBlockForger` requires
+`EraParams` alongside it), the runtime gate pre-flights the opcert's
+issue-number counter against `LedgerView.LatestOpCertSequence` using the
+same era-scoped rule block application enforces at apply time
+(`ledger/eras.ValidateOpCertCounter`, the single implementation shared with
+`validateOpCertCounter` in `ledger/verify_opcert.go` and mirrored here as
+`validateOpCertSequence`): a counter behind the last value the ledger
+observed for this pool is always rejected (stale or stolen hot key), and
+one that skips ahead of it is additionally rejected in Praos eras (Babbage
+onward) but accepted in TPraos eras (Shelley-Alonzo). `EraParams.ProtocolParamsForSlot`
+resolves the era for the slot being forged. Startup validation and the
+KES-lifetime gate above cover genesis-derived evolution limits, which are
+fixed for the life of the chain; this covers the era-scoped counter rule
+and the on-chain observed counter, both of which can change after startup
+as blocks are applied (this node's own or a peer's for the same pool) or
+as the chain crosses an era boundary. The check runs right after leader
+selection -- it costs a real ledger read, so it is skipped for a slot this
+pool does not lead -- but still before the forge-slot fence and any
+VRF/KES/Leios work, so a bad key state costs a could-not-forge disposition
+instead of a burned leader slot and a rejected `AddLocalBlock` call. The
+check is opt-in: a nil `OpCertLedgerView` (dev mode, embedders without
+ledger wiring) skips it entirely, unchanged from before.
 
 Each production forge attempt takes an independently owned snapshot of one
 complete credential generation at the runtime gate. The snapshot deep-copies
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -4364,6 +4364,33 @@ instead of a burned leader slot and a rejected `AddLocalBlock` call. The
 check is opt-in: a nil `OpCertLedgerView` (dev mode, embedders without
 ledger wiring) skips it entirely, unchanged from before.
 
+`LedgerState.LatestOpCertSequence` -- the `LedgerView` method both this
+gate and startup's `PoolCredentials.ValidateAgainstLedger` read through --
+resolves the "latest observed" counter via the same Mithril-boundary-aware
+path as block application (`latestOpCertCounterForValidation`, both now
+backed by the shared `latestOpCertCounterAfterMithril`), instead of a plain
+`MAX` over the whole table. A Mithril-restored node's per-pool opcert
+history is only trustworthy after the certified boundary; a plain MAX could
+return a stale pre-boundary row that block application itself does not
+trust as a baseline. `LatestOpCertSequence` captures `mithrilLedgerSlot`
+under a read lock before calling the shared resolver, since (unlike
+`latestOpCertCounterForValidation`'s block-application caller) neither
+startup validation nor the forge loop otherwise holds a lock across that
+field.
+
+Startup's `PoolCredentials.ValidateAgainstLedger` applies the identical
+era-scoped counter rule (an `enforceNoGap` parameter, resolved by
+`Node.opCertNoGapRuleForCurrentSlot` via the same exported
+`forging.EnforceOpCertNoGapRule` era classification the forge loop uses),
+not just the staleness half it always checked. Before this, a Praos-era
+opcert that was gapped but not stale would start the node cleanly and only
+be rejected on every subsequent forge attempt; startup now surfaces that
+loudly instead. Era resolution failing at startup (e.g. protocol
+parameters genuinely unavailable yet) is not itself fatal: it logs a
+warning and falls back to the stale-only check rather than adding a new
+startup-failure mode for a condition unrelated to the credential
+cross-check itself.
+
 Each production forge attempt takes an independently owned snapshot of one
 complete credential generation at the runtime gate. The snapshot deep-copies
 the VRF secret, KES secret, verification keys, opcert, and validated lifetime;
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -4357,12 +4357,13 @@ fixed for the life of the chain; this covers the era-scoped counter rule
 and the on-chain observed counter, both of which can change after startup
 as blocks are applied (this node's own or a peer's for the same pool) or
 as the chain crosses an era boundary. The check runs right after leader
-selection -- it costs a real ledger read, so it is skipped for a slot this
-pool does not lead -- but still before the forge-slot fence and any
-VRF/KES/Leios work, so a bad key state costs a could-not-forge disposition
-instead of a burned leader slot and a rejected `AddLocalBlock` call. The
-check is opt-in: a nil `OpCertLedgerView` (dev mode, embedders without
-ledger wiring) skips it entirely, unchanged from before.
+selection (a Praos leader-VRF check that, together with the KES-lifetime
+gate above, already precedes it) -- it costs a real ledger read, so it is
+skipped for a slot this pool does not lead -- but still before Leios work
+and the forge-slot fence, so a bad key state costs a could-not-forge
+disposition instead of a burned leader slot and a rejected `AddLocalBlock`
+call. The check is opt-in: a nil `OpCertLedgerView` (dev mode, embedders
+without ledger wiring) skips it entirely, unchanged from before.
 
 `LedgerState.LatestOpCertSequence` -- the `LedgerView` method both this
 gate and startup's `PoolCredentials.ValidateAgainstLedger` read through --
@@ -4372,24 +4373,26 @@ backed by the shared `latestOpCertCounterAfterMithril`), instead of a plain
 `MAX` over the whole table. A Mithril-restored node's per-pool opcert
 history is only trustworthy after the certified boundary; a plain MAX could
 return a stale pre-boundary row that block application itself does not
-trust as a baseline. `LatestOpCertSequence` captures `mithrilLedgerSlot`
-under a read lock before calling the shared resolver, since (unlike
-`latestOpCertCounterForValidation`'s block-application caller) neither
-startup validation nor the forge loop otherwise holds a lock across that
-field.
+trust as a baseline. Neither caller otherwise holds a lock across
+`mithrilLedgerSlot`, so both resolve it through the existing lock-safe
+`mithrilLedgerSlotSnapshot` accessor before calling the shared resolver,
+rather than reading the field directly.
 
-Startup's `PoolCredentials.ValidateAgainstLedger` applies the identical
-era-scoped counter rule (an `enforceNoGap` parameter, resolved by
-`Node.opCertNoGapRuleForCurrentSlot` via the same exported
-`forging.EnforceOpCertNoGapRule` era classification the forge loop uses),
-not just the staleness half it always checked. Before this, a Praos-era
-opcert that was gapped but not stale would start the node cleanly and only
-be rejected on every subsequent forge attempt; startup now surfaces that
-loudly instead. Era resolution failing at startup (e.g. protocol
-parameters genuinely unavailable yet) is not itself fatal: it logs a
-warning and falls back to the stale-only check rather than adding a new
-startup-failure mode for a condition unrelated to the credential
-cross-check itself.
+Startup's `PoolCredentials.ValidateAgainstLedger` deliberately stays on the
+staleness-only half of this rule and does not apply the era-scoped no-gap
+check the forge loop enforces. Applying it at startup would pair a
+wall-clock-resolved era (`LedgerState.CurrentSlot` is wall-clock and valid
+regardless of sync state) with a baseline that only reflects the applied
+chain (`LatestOpCertSequence`); on a node whose applied tip is behind wall-
+clock time -- an interrupted initial sync, a resume after downtime, a
+restore to an older snapshot -- those two can disagree enough to make a
+pool several opcert rotations into its life look gapped against a baseline
+that has simply not caught up yet, and a fatal startup rejection would
+prevent the node from ever syncing to the point that makes the baseline
+correct. The forge loop's own gate does not have this problem: it runs
+after the upstream-sync skip and the leader check, so both its era and its
+baseline come from near-tip state, and a rejection there costs one slot
+rather than the node's ability to start.
 
 Each production forge attempt takes an independently owned snapshot of one
 complete credential generation at the runtime gate. The snapshot deep-copies
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"log/slog"
 	"math"
-	"reflect"
 	"runtime/debug"
 	"sync"
 	"time"
@@ -807,6 +806,20 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 		return nil
 	}
 
+	// Check if we're the leader for this slot only after the KES gate.
+	isLeader := f.checkLeaderSafe(currentSlot)
+	if !isLeader {
+		f.logger.Debug(
+			"forge check: not leader for slot",
+			"current_slot", currentSlot,
+			"tip_slot", tipSlot,
+		)
+		if f.metrics != nil {
+			f.metrics.forgeNotLeader.Inc()
+		}
+		return nil
+	}
+
 	// Pre-flight the OpCert counter against the ledger's observed on-chain
 	// state and the era-scoped rule block application enforces (see
 	// checkOpCertSequence). This covers genesis/era context the KES-lifetime
@@ -814,6 +827,10 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 	// applied (this node's own or a peer's for the same pool), so a key
 	// state that was fine at startup or on an earlier slot can become stale,
 	// or -- from Babbage onward -- gapped, by the time a later slot is won.
+	// Checked only for the winning slot -- after the cheap leader check --
+	// rather than on every KES-valid slot, since the check performs a real
+	// ledger read that would otherwise run regardless of whether this pool
+	// is even leader for the slot.
 	if f.opCertLedgerView != nil {
 		if err := f.checkOpCertSequence(currentSlot, generation); err != nil {
 			f.incCouldNotForge()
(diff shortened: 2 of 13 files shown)


ledger/forging/keys.go
```diff
@@ -1116,3 +1116,45 @@ func (pc *PoolCredentials) ValidateAgainstLedger(
...
+		return nil
+	}
+	if candidate < stored {
+		return fmt.Errorf(
+			"opcert counter %d is below last seen %d (stale or stolen hot key)",
+			candidate,
+			stored,
+		)
+	}
+	if enforceNoGap && candidate > stored && candidate-stored > 1 {
```
wolf31o2: This diverges from the reference it says it mirrors. `validateOpCertCounter` uses `candidate > stored+1`, which wraps at `stored == MaxUint64` and rejects `candidate == stored`; this one accepts it. Verified against the real function:

```
validateOpCertCounter(MaxUint64, true, MaxUint64, true)
  = "opcert counter 18446744073709551615 skips ahead of last seen 18446744073709551615 (gapped rotation)"
```

So `keys_test.go`'s new "maximum counter accepted when unchanged" case asserts the opposite of what block application does, while the doc comment says "The two must stay in agreement" and the test comment says it mirrors `TestValidateOpCertCounter`. Unreachable in practice (opcert counters are `uint32` on the wire), but either fix the reference's overflow too or drop the parity claim for this boundary.

ledger/forging/forger.go
```diff
@@ -1178,6 +1228,45 @@ func (f *BlockForger) incCouldNotForge() {
...
+	generation *credentialGeneration,
+) error {
+	opCert := generation.opCert()
+	if opCert == nil {
+		return errors.New("operational certificate not loaded")
+	}
+	credsPoolID := f.creds.GetPoolID()
+	var poolID [28]byte
+	copy(poolID[:], credsPoolID[:])
+	stored, found, err := f.opCertLedgerView.LatestOpCertSequence(poolID)
```
wolf31o2, line 1252: This is new per-slot database work on the forge path. `checkAndForgeProduction` runs once per slot (`runLoopSlotAligned`), and this gate sits ahead of the leader check, so every slot now costs `LedgerState.LatestOpCertSequence` → `GetPool` (which also runs `loadPoolAssociations` for registrations/retirements, `sqlstore/pool.go:215`) plus `LatestPoolOpCertSequence`. Every other gate above it is in-memory.

The stated reason for placing it before leader selection is to avoid spending VRF/KES work, but the VRF eval is far cheaper than two-plus SQL round trips, so the ordering costs more than it saves. Either cache the value and invalidate on block apply, or move the gate after `checkLeaderSafe`.

Separately: block application resolves the baseline through `latestOpCertCounterForValidation` (`ledger/state.go:6715`), which is Mithril-trust-boundary aware; this path uses the plain `MAX`. They coincide only because a bootstrap writes no rows below its anchor — worth using the same resolver, or saying why not.

ledger/forging/forger.go
```diff
@@ -1178,6 +1228,45 @@ func (f *BlockForger) incCouldNotForge() {
...
+	if err != nil {
+		return fmt.Errorf("opcert sequence lookup: %w", err)
+	}
+	pparams := f.eraParams.ProtocolParamsForSlot(slot)
+	if pparams == nil || (reflect.ValueOf(pparams).Kind() == reflect.Ptr && reflect.ValueOf(pparams).IsNil()) {
```
wolf31o2: The typed-nil half of this guard is never exercised. `TestCheckAndForgeProductionSkipsWhenEraUnresolvable` sets `mockPParamsProvider{pparams: nil}`, a true nil interface, which the `pparams == nil` half catches. Missing case: a provider returning `(*babbage.BabbageProtocolParameters)(nil)` — the case the `reflect` check exists for, and the one that panics in `extractPParamsLimits` without it.

ledger/forging/opcert_sequence_gate_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+// TestCheckAndForgeProductionSkipsOnStaleOpCertCounter covers a stale
+// counter -- the ledger has observed a higher issue number for this pool
+// than the loaded credentials carry -- which block application would
+// reject regardless of era. It must be caught before the leader check,
+// block build, or forge-slot fence.
+func TestCheckAndForgeProductionSkipsOnStaleOpCertCounter(t *testing.T) {
+	builder, broadcaster := newOpCertSequenceGateTestBuilder()
+	var logs bytes.Buffer
+	view := &fakeLedgerView{seqFound: true, latestSeq: 1}
```
wolf31o2, line 103: Two gaps here.

`fakeLedgerView.seqErr` exists and no test sets it, so the `opcert sequence lookup: %w` path is uncovered. That path is the highest-consequence behaviour in this change: a transient database read error now skips the slot, so a DB hiccup on a leader slot costs a block. Whether that is the intended failure mode deserves a test either way.

And `forgerCountingLeader.calls` is available but unasserted — `require.Zero(t, leader.calls)` here would pin the "before leader selection" ordering the design rests on.

node_forging.go
```diff
@@ -373,6 +373,10 @@ func (n *Node) initBlockForger(
 		LeiosTxValidator:                n.ledgerState,
 		LeiosCertificateProvider:        leiosCerts,
 		LeiosParentAnnouncementProvider: leiosParent,
+		OpCertLedgerView: blockProducerLedgerView{
```
wolf31o2, line 392: This wiring has no coverage — every gate test builds `ForgerConfig` directly, so nothing catches `OpCertLedgerView`/`EraParams` being dropped here.

Related: startup `PoolCredentials.ValidateAgainstLedger` still rejects only a stale counter, not a gapped one, so a Praos-era gapped opcert starts the node cleanly and is then rejected on every forge attempt. Applying the same era-scoped rule at startup would make it loud before the node looks healthy.

wolf31o2 (commented):
No blocking findings. The era split matches the reference (`opCertNoGapRuleApplies`: TPraos Shelley–Alonzo monotonic only, Praos Babbage+ no-gap) and matches cardano-node's `CounterTooSmallOCERT` / `CounterOverIncrementedOCERT`, so nothing here rejects a counter the network would accept.

Five non-blocking notes inline: one behavioural (new per-slot database work on the forge path), one parity claim that is now false at the boundary, and three named test gaps.

ledger/forging/keys.go
```diff
@@ -1116,3 +1116,45 @@ func (pc *PoolCredentials) ValidateAgainstLedger(
...
+		return nil
+	}
+	if candidate < stored {
+		return fmt.Errorf(
+			"opcert counter %d is below last seen %d (stale or stolen hot key)",
+			candidate,
+			stored,
+		)
+	}
+	if enforceNoGap && candidate > stored+1 {
```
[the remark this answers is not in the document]
arepala-uml: Fixed by centralizing the counter rule into `ledger/eras.ValidateOpCertCounter` (commit 1d7e0b7e), whose comparison is `candidate > stored && candidate-stored > 1`. `ledger/forging/keys.go`'s `validateOpCertSequence` and `ledger/verify_opcert.go`'s `validateOpCertCounter` both now delegate to it, so the fix applies to both call sites from one place. Added the `equal counters at max uint64` boundary case to both `TestValidateOpCertSequence` and `TestValidateOpCertCounter` (commit bbf8f77f).

ledger/forging/forger.go
```diff
@@ -1178,6 +1227,45 @@ func (f *BlockForger) incCouldNotForge() {
...
+		return fmt.Errorf("opcert sequence lookup: %w", err)
+	}
+	pparams := f.eraParams.ProtocolParamsForSlot(slot)
+	if pparams == nil {
+		return fmt.Errorf(
+			"protocol parameters unavailable for slot %d",
+			slot,
+		)
+	}
+	limits, err := extractPParamsLimits(pparams)
```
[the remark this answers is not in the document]
arepala-uml: Fixed: the typed-nil check now lives in `extractPParamsLimits` (`ledger/forging/eras.go`), which every caller (including `DefaultBlockBuilder.buildBlock` via builder.go) goes through, rather than only guarding `checkOpCertSequence`. Commit bbf8f77f removes the forger.go-local reflect check in favor of this single location, and adds `TestExtractPParamsLimitsRejectsTypedNilPointer` plus an end-to-end `TestCheckAndForgeProductionSkipsOnTypedNilProtocolParameters`.

ledger/forging/keys.go
```diff
@@ -1116,3 +1116,45 @@ func (pc *PoolCredentials) ValidateAgainstLedger(
...
+// adopted.
+//
+// A counter below the last-seen value is always rejected (stale or stolen
+// hot key). A counter that skips ahead of it is rejected only when
+// enforceNoGap is set: the over-increment rule is Praos-only (Babbage
+// onward), while TPraos eras (Shelley-Alonzo) accept any candidate at or
+// above stored. When the ledger has no recorded counter for this pool
+// (found is false) there is no baseline to compare against, so the
+// candidate is accepted.
+func validateOpCertSequence(
```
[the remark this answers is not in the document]
arepala-uml: Fixed: extracted the shared rule into `ledger/eras.ValidateOpCertCounter`. `ledger/forging/keys.go`'s `validateOpCertSequence` and `ledger/verify_opcert.go`'s `validateOpCertCounter` both delegate to it now (commit 1d7e0b7e), so the two cannot drift apart.

ledger/forging/opcert_sequence_gate_test.go
```diff
@@ -0,0 +1,270 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+// opCertSequenceGateForger builds a production BlockForger whose credentials'
+// OpCert issue number is the fixed value baked into testOpCertJSON (0, see
+// TestValidateOpCertUnsafe_ValidCertificate), wired with the given ledger
+// view / era params so tests can drive the pre-flight counter gate.
+func opCertSequenceGateForger(
```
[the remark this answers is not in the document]
arepala-uml: Fixed: the tests now retain the `LeaderChecker` and assert `leader.callCount()` in every branch (commit 1d7e0b7e), pinning that leader selection runs before the counter check. The gate itself was also moved to run after `checkLeaderSafe` instead of before it, per the related performance finding.

ledger/forging/forger.go
```diff
@@ -780,6 +806,29 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
...
 
+	// Pre-flight the OpCert counter against the ledger's observed on-chain
+	// state and the era-scoped rule block application enforces (see
+	// checkOpCertSequence). This covers genesis/era context the KES-lifetime
+	// gate above does not: LatestOpCertSequence advances as blocks are
+	// applied (this node's own or a peer's for the same pool), so a key
+	// state that was fine at startup or on an earlier slot can become stale,
+	// or -- from Babbage onward -- gapped, by the time a later slot is won.
+	if f.opCertLedgerView != nil {
+		if err := f.checkOpCertSequence(currentSlot, generation); err != nil {
```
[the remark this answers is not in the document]
arepala-uml: Fixed: moved the counter-check call site in `checkAndForgeProduction` to run after `checkLeaderSafe` (commit 1d7e0b7e), so the ledger read only happens for the winning slot instead of every KES-valid one.

ledger/forging/keys.go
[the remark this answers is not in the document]
arepala-uml: Fixed by centralizing the counter rule into `ledger/eras.ValidateOpCertCounter` (commit 1d7e0b7e), whose comparison is `candidate > stored && candidate-stored > 1`. `ledger/forging/keys.go`'s `validateOpCertSequence` and `ledger/verify_opcert.go`'s `validateOpCertCounter` both now delegate to it, so the fix applies to both call sites from one place. Added the `equal counters at max uint64` boundary case to both `TestValidateOpCertSequence` and `TestValidateOpCertCounter` (commit bbf8f77f).

ledger/forging/forger.go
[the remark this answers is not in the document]
arepala-uml: Fixed: the typed-nil check now lives in `extractPParamsLimits` (`ledger/forging/eras.go`), which every caller (including `DefaultBlockBuilder.buildBlock` via builder.go) goes through, rather than only guarding `checkOpCertSequence`. Commit bbf8f77f removes the forger.go-local reflect check in favor of this single location, and adds `TestExtractPParamsLimitsRejectsTypedNilPointer` plus an end-to-end `TestCheckAndForgeProductionSkipsOnTypedNilProtocolParameters`.

ledger/forging/keys.go
arepala-uml: Resolved by centralizing both functions onto `ledger/eras.ValidateOpCertCounter` (commit 1d7e0b7e): there is no longer a separate reference implementation for this to diverge from. Added the `equal counters at max uint64` boundary case to both `TestValidateOpCertSequence` (forging) and `TestValidateOpCertCounter` (ledger) to pin the shared behavior (commit bbf8f77f).

ledger/forging/forger.go
arepala-uml: Ordering: fixed in 1d7e0b7e (the gate now runs after `checkLeaderSafe`, not before it).

Mithril resolver: fixed in bbf8f77f. `LedgerState.LatestOpCertSequence` now resolves through the same Mithril-boundary-aware path as `latestOpCertCounterForValidation` (both backed by an extracted `latestOpCertCounterAfterMithril`), instead of a plain MAX that could trust a stale pre-boundary row. Added `TestLatestOpCertSequenceRespectsMithrilBoundary`, which fails against the prior plain-MAX behavior and passes against the fix.

ledger/forging/forger.go
arepala-uml: Added `TestCheckAndForgeProductionSkipsOnTypedNilProtocolParameters` (commit bbf8f77f), driving `checkAndForgeProduction` end-to-end with a `ProtocolParamsForSlot` that returns `(*babbage.BabbageProtocolParameters)(nil)`.

ledger/forging/opcert_sequence_gate_test.go
arepala-uml: leader.calls: was already asserted for the skip branches as of 1d7e0b7e (`leader.callCount()`), alongside moving the gate to run after `checkLeaderSafe`.

seqErr: added `TestCheckAndForgeProductionSkipsOnTransientLedgerReadError` (commit bbf8f77f). A transient `LatestOpCertSequence` error fails closed -- skips the slot, increments `forgeCouldNot`, builds nothing -- the same disposition as a definite stale/gapped counter.

node_forging.go
arepala-uml: Startup no-gap rule: fixed in bbf8f77f. `PoolCredentials.ValidateAgainstLedger` now takes `enforceNoGap`, resolved by `Node.opCertNoGapRuleForCurrentSlot` via the new exported `forging.EnforceOpCertNoGapRule`. Added `TestValidateAgainstLedger_GappedOpCertRejectedUnderNoGap` and `..._AcceptedWithoutNoGap`. An era-resolution failure at startup logs a warning and falls back to the pre-existing stale-only check rather than adding a new startup-failure mode.

`initBlockForger` wiring coverage: not added here. Every field in that `ForgerConfig` literal (`BlockValidator`, `ForgeFence`, `LeiosProduceChecker`, etc.) is equally untested by field-level assertion today, since exercising it means standing up a full `*Node` (chain manager, mempool, event bus, db). Real gap, but pre-existing and not introduced by this change; treating it as separate scope rather than adding partial coverage for just these two fields here.

arepala-uml:
Addressed all bot and human review findings across two follow-up commits.

**1d7e0b7e** — centralized the opcert counter rule into `ledger/eras.ValidateOpCertCounter` (single implementation for `ledger/verify_opcert.go` and `ledger/forging/keys.go`, fixing the uint64 wraparound at the max-counter boundary in both), moved the typed-nil guard into `extractPParamsLimits` so every caller is protected, moved the counter pre-flight to run after leader selection instead of before it, and pinned leader-call ordering in the tests.

**bbf8f77f** — `LedgerState.LatestOpCertSequence` now resolves through the same Mithril-boundary-aware path as block application (`latestOpCertCounterForValidation`, both backed by an extracted `latestOpCertCounterAfterMithril`) instead of a plain MAX; `PoolCredentials.ValidateAgainstLedger` now applies the same era-scoped no-gap rule at startup, not just the stale check; added end-to-end coverage for the typed-nil path, a transient ledger-read error, the Mithril-boundary regression, and the startup no-gap accept/reject pair.

Replied on every open thread with the specific fix/commit. One item (field-level test coverage for `initBlockForger`'s `ForgerConfig` wiring) is left as a separately-scoped, pre-existing gap — noted in the reply on that thread.

node_forging.go
```diff
@@ -163,7 +163,25 @@ func (n *Node) validateBlockProducerLedgerWithView(
 	if creds == nil {
 		return errors.New("nil pool credentials")
 	}
-	registered, vrfMatched, err := creds.ValidateAgainstLedger(view)
+	enforceNoGap, gapErr := n.opCertNoGapRuleForCurrentSlot()
```
wolf31o2: The startup no-gap check pairs a wall-clock-resolved era with an applied-ledger baseline, and its failure is fatal: `node.go:1696` and `node_lifecycle.go:1276` both wrap this into a returned error, and `Run` exits.

`opCertNoGapRuleForCurrentSlot` resolves the era from `ls.CurrentSlot()`, which is wall-clock and documented as working regardless of sync state (`ledger/slot_clock.go:225-231`); `ProtocolParamsForSlot` then forecasts forward to the wall-clock epoch. The baseline comes from `LatestOpCertSequence`, which is the counter observed in the applied chain. Nothing between them requires the ledger to be at tip, and `Run` calls this immediately after the ledger starts with no sync gate.

Concrete case: a pool several opcert rotations into its life, on a node whose applied tip predates its last rotation (interrupted initial sync, restore from an older snapshot, resume after extended downtime). `LatestOpCertSequence` returns 2 while the loaded opcert carries 7; the era resolves to Conway so `enforceNoGap` is true; `ValidateAgainstLedger` returns `opcert counter 7 skips ahead of last seen 2 (gapped rotation)` and the node refuses to start. It cannot then sync to the point that would make the baseline correct. The same credentials pass once the chain reaches tip.

The forge-loop gate is not exposed to this: it runs after the upstream-sync skip and the leader check, so its baseline and its era both come from near-tip state.

Resolving the era at the applied tip slot instead of the wall clock, and requiring the ledger to be at tip before enforcing the gap half, would confine the rule to the state it is meaningful in. Leaving startup on the stale-only check is the other option, since the forge gate already fails closed per slot.

ledger/state.go
```diff
@@ -6712,22 +6721,47 @@ func (ls *LedgerState) ledgerProcessBlock(
 // latestOpCertCounterForValidation returns the highest observed counter after
 // the Mithril boundary, or the certified counter at the boundary when no later
 // row exists. Rows before the boundary are not part of the certified state.
+// Called only from block application, which already holds whatever lock
```
wolf31o2: Block application does not hold a lock across `ls.mithrilLedgerSlot`. `ledgerProcessBlocksFromSource` captures its snapshots and calls `ls.RUnlock()` at `state.go:5675`, specifically to avoid reading `ls` fields unlocked inside the transaction callback. `ledgerProcessBlock` is invoked at `state.go:5902`, after that unlock, with no `RLock`/`Lock` in between, so this function reads the field with no lock held.

The read is benign today only because `mithrilLedgerSlot` is written once, by `loadMithrilTrustBoundary` during `Start` (`state.go:1772`). Passing the already-captured `snapshotMithrilSlot` down, or using `mithrilLedgerSlotSnapshot()` (`chainsync.go:368`), would make the comment true. The same claim appears in `ARCHITECTURE.md:4375-4380`.

ARCHITECTURE.md
```diff
@@ -4341,6 +4341,56 @@ The start, expiry, current-period, and remaining-period gauges use that same
...
+one that skips ahead of it is additionally rejected in Praos eras (Babbage
+onward) but accepted in TPraos eras (Shelley-Alonzo). `EraParams.ProtocolParamsForSlot`
+resolves the era for the slot being forged. Startup validation and the
+KES-lifetime gate above cover genesis-derived evolution limits, which are
+fixed for the life of the chain; this covers the era-scoped counter rule
+and the on-chain observed counter, both of which can change after startup
+as blocks are applied (this node's own or a peer's for the same pool) or
+as the chain crosses an era boundary. The check runs right after leader
+selection -- it costs a real ledger read, so it is skipped for a slot this
+pool does not lead -- but still before the forge-slot fence and any
```
wolf31o2: `still before the forge-slot fence and any VRF/KES/Leios work` is no longer accurate after the reorder. `checkLeaderSafe` is the Praos leader VRF evaluation and now runs first, and the KES-lifetime gate precedes it as well; only Leios work and the fence still follow. `ledger/forging/forger.go:317` carries the same claim.

Separately, `startup now surfaces that loudly` (line 4387) understates the disposition: a gapped counter at startup is fatal, not a warning.

wolf31o2 (changes requested):
The five prior notes are addressed. The counter rule is centralized in `ledger/eras.ValidateOpCertCounter` with the max-uint64 wraparound fixed on both sides, the gate moved after leader selection, `LatestOpCertSequence` now resolves through the Mithril-boundary-aware path, and the typed-nil, transient-read-error and leader-ordering cases each have a test that fails when its fix is reverted. All three call sites of the shared rule get identical behavior; the Leios OCIN freshness check (`ledger/verify_opcert.go:226`) is a deliberately different rule and is correctly left out.

One blocking finding in the new startup path, plus two inaccuracies in newly added documentation.

node_forging.go
arepala-uml: Confirmed and reverted (commit df5d1c00). `opCertNoGapRuleForCurrentSlot` and `ValidateAgainstLedger`'s `enforceNoGap` parameter are gone; `Node.opCertNoGapRuleForCurrentSlot` and the now-unused `forging.EnforceOpCertNoGapRule` are removed. Startup goes back to the staleness-only check it always had, with a doc comment on `ValidateAgainstLedger` explaining why it cannot safely apply the era-scoped half of the rule. The forge loop's own gate (near-tip era + near-tip baseline) is unaffected.

ledger/state.go
arepala-uml: Confirmed the comment was wrong -- fixed in df5d1c00. Both `latestOpCertCounterForValidation` and `LatestOpCertSequence` now read `mithrilLedgerSlot` through the existing `mithrilLedgerSlotSnapshot()` accessor instead of the field directly (or my ad hoc RLock/RUnlock in `LatestOpCertSequence`), and the doc comment no longer claims block application holds a lock across it.

ARCHITECTURE.md
arepala-uml: Both fixed in df5d1c00. `ForgerConfig.OpCertLedgerView`'s doc comment and the matching ARCHITECTURE.md paragraph now say the counter check runs after leader selection and the KES-lifetime gate, before Leios work and the forge-slot fence. The "startup surfaces that loudly" sentence is gone along with the reverted startup enforcement it was describing.

arepala-uml:
Commit df5d1c00 addresses the three new findings from the last review pass:

- Reverted the startup no-gap enforcement added in bbf8f77f: it paired a wall-clock-resolved era with an applied-chain baseline, which could reject a legitimately-rotated opcert on a node still catching up and block it from ever starting. Startup stays on the staleness-only check; the forge loop's own gate (near-tip era and baseline) is unaffected.
- Fixed a false lock-safety claim: `latestOpCertCounterForValidation` does not run under a lock across `mithrilLedgerSlot`. Both it and `LatestOpCertSequence` now go through the existing `mithrilLedgerSlotSnapshot()` accessor.
- Fixed stale ordering claims in `ForgerConfig.OpCertLedgerView`'s doc comment and ARCHITECTURE.md ("before VRF/KES/Leios work") — leader selection and the KES-lifetime gate precede the counter check; only Leios work and the fence still follow it.

Replied on each of the three threads with the specific fix.

ledger/forging/keys.go
```diff
@@ -1108,11 +1123,37 @@ func (pc *PoolCredentials) ValidateAgainstLedger(
...
 		)
 	}
 	return true, vrfMatched, nil
 }
+
+// validateOpCertSequence enforces the same operational-certificate counter
+// rule block application applies (ledger/verify_opcert.go
+// validateOpCertCounter), so the forge loop can decline a leader slot for a
+// key state ledgerProcessBlock would reject, before spending the slot on
+// VRF/KES/Leios work or committing the duplicate-slot fence. The two must
```
wolf31o2, line 1147: Non-blocking: this is a third instance of the ordering claim corrected in `ForgerConfig.OpCertLedgerView` and `ARCHITECTURE.md`. `checkLeaderSafe` (the Praos leader VRF evaluation) and the KES-lifetime gate both run before `checkOpCertSequence` at `ledger/forging/forger.go:837`; only Leios work and `reserveForgeSlot` still follow it, so `before spending the slot on VRF/KES/Leios work` overstates what the check precedes.

wolf31o2 (approved):

ledger/forging/keys.go
arepala-uml: Confirmed — this instance was missed when the other two were fixed. The wording is corrected locally (checkAndForgeProduction calls this after checkLeaderSafe and the KES-lifetime gate, before Leios work and reserveForgeSlot), but not pushed: this PR is already approved at the current head, and since this is doc-comment wording only, non-blocking, with no functional effect, I'm deferring it rather than pushing another commit against an approved SHA. Will land as part of a future change to this file.

arepala-uml: Landed as #4003 (this PR was already merged, and it's a comment-only, non-functional wording fix, so it didn't warrant reopening this one). Reworded to match the already-corrected wording in ForgerConfig.OpCertLedgerView / ARCHITECTURE.md — checkLeaderSafe and the KES-lifetime gate both run before this check, only Leios work and the forge-slot fence still follow it.

Issue #3656, opened 2026-08-30
api: bound UTxO-RPC streaming request work


wolf31o2:
UTxO-RPC streaming methods accept request lists whose semantic work is not bounded by the transport size limit. `FollowTip` and `WatchTx` can process an unlimited intersection list, and `WaitForTx` allocates and queries for every supplied reference. Transaction hashes are not required to be 32 bytes.

`WatchTx` also stops rollback traversal before checking the point exactly at the configured security depth.

Acceptance criteria:

- Set explicit count and work limits for intersections and transaction references.
- Reject transaction hashes that are not exactly 32 bytes.
- Check the rollback point at the full configured depth without traversing farther.
- Add oversized-list, malformed-hash, and exact-boundary tests.

wolf31o2:
Current `WaitForTx` behavior confirms the remaining request-work gap tracked here:

- There is no explicit cap on distinct transaction references.
- The handler allocates a pending map and confirmation channel proportional to the distinct reference count.
- It subscribes to committed transaction events before performing one durable lookup per unresolved reference.
- The server timeout starts only after that lookup loop completes.

The count limit should be applied before allocation, subscription, or ledger access, and the request context should bound the durable lookup phase as well as the streaming phase.

PR #3929, opened 2026-09-04
fix(config): bind API listeners to loopback by default

### LINKED ISSUE #3498: config: make API exposure explicit by default
Blockfrost, Mesh, and UTxO-RPC default to listeners on every interface while
authentication is disabled, and the shared CORS default accepts every origin.
The effective defaults are broader than the other local service defaults and
vary across API plugins.

Acceptance criteria:

- Bind API listeners to loopback by default.
- Require explicit configuration for non-loopback listeners and wildcard CORS.
- Apply the same effective defaults to Blockfrost, Mesh, and UTxO-RPC.
- Cover CLI, YAML, and per-plugin overrides.

chrisguiney:
Closes #3498.

## Problem

Blockfrost, Mesh, and UTxO RPC took `bindAddr`'s `0.0.0.0` default while
`api.auth` defaults to `disabled`, and `corsAllowedOrigins` defaulted to
`["*"]`. `bindAddr` is not API-specific — the relay/NtN and metrics
listeners need it on a wildcard — so an operator could not widen the relay
without also exposing three unauthenticated API listeners.

## Change

`apiBindAddr` is a root setting defaulting to `127.0.0.1` and resolved
*independently of* `bindAddr`. This is the rule `debugBindAddr` already
follows for pprof, applied to the API listeners:

- `internal/config`: `APIBindAddr` (`apiBindAddr` / `DINGO_API_BIND_ADDR` /
  `--api-bind-addr`), `DefaultAPIBindAddr`, defaulted in `ApplyDefaults` and
  again in `dingo.Config.syncCompatFields` so the programmatic path is
  fail-safe too.
- Per-provider override: `plugins.api.<name>.config.host` on all three
  `ProviderConfig`s, so one listener can be widened while the other two stay
  on loopback.
- `internal/apiconfig.ListenHost` is the single definition of that
  precedence. It is read by `internal/config.APIPluginHost` (so `Validate`'s
  port-conflict check sees the address the listener will really bind) and by
  each provider's own `RegisterProvider` factory, so validation and the bind
  cannot disagree.
- `corsAllowedOrigins` defaults to an empty list — no
  `Access-Control-Allow-Origin` header — instead of `["*"]`.
- `bindAddr` keeps `0.0.0.0` and keeps serving relay/NtN and metrics.

## Interaction with the API authentication requirement

`main` since added `ValidateAPIExposure`, which refuses an enabled API on a
non-loopback bind whose effective authentication is disabled. It read the
shared `apiBindAddr` alone, which the per-provider `host` override makes
insufficient in both directions: a provider widened by its own `host`
escaped the requirement while `apiBindAddr` was loopback, and a provider
pinned to loopback by its own `host` was required to authenticate while
`apiBindAddr` was remote. It now resolves through `APIListenHost`, the same
precedence the provider factories and the port-conflict check use.

That requirement also applies to the examples stack, whose listeners must
bind the container interface for `blockfrost-explorer` and `sundae-preview`
to reach them across the Compose network. `examples/docker-compose.yml` now
configures token authentication: Compose writes `DINGO_API_TOKEN` to
`/run/secrets/dingo-api-token` in both Dingo containers, and each Vite dev
server attaches the same token as `Authorization: Bearer` on the requests it
proxies, so it stays server-side. The token arrives as a file because
`api.auth.token` carries `ignored:"true"` so an inline secret cannot come
from the environment.

## Breaking change and migration

A deployment relying on the previous wildcard API bind must set `apiBindAddr`
(or that provider's own `host`) explicitly, and pair it with `api.auth`; one
relying on wildcard CORS must set `corsAllowedOrigins` explicitly. README's
"API Exposure" section documents both.

A container or pod bind is a non-loopback bind: a listener on the container's
own loopback is unreachable from a published port, a sibling container, and a
Kubernetes Service alike, so a containerised deployment needs
`DINGO_API_BIND_ADDR=0.0.0.0` (or a per-provider `host`) *inside* the
container, plus authentication.

**Follow-up in another repository:** the `blinklabs-io/helm-charts` `dingo`
chart targets the UTxO RPC / Blockfrost / Mesh container ports directly. Those
Services get no reachable endpoint until the chart sets
`DINGO_API_BIND_ADDR: "0.0.0.0"`, and that bind then needs a mounted token
file. Tracked as blinklabs-io/helm-charts#464; not changed here.

## Docs

`ARCHITECTURE.md` ("API security") and `README.md` ("API Exposure") updated;
`dingo.yaml.example` documents `apiBindAddr` and `corsAllowedOrigins`.
`DATABASE.md` checked and unaffected.


chrisguiney:
@coderabbitai full review

dingo.yaml.example
```diff
@@ -3,10 +3,23 @@
...
 # runtime default.
 
-# Public bind address for the Dingo server
+# Public bind address for the relay/NtN and metrics listeners.
 # CLI: --bind-addr
 bindAddr: "0.0.0.0"
 
+# Bind address for the Blockfrost, Mesh, and UTxO RPC listeners. Deliberately
+# independent of bindAddr and loopback by default: these listeners have
+# authentication disabled unless api.auth is configured, so exposing one is an
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 8a7ca87a. Reworded to `unless api.auth -- or that listener's own plugins.api.<name>.config.auth -- is configured`. The per-provider override is real (`internal/apiconfig.MergeAuth`, and the file's own later example uses it), so the original comment was incomplete.

internal/config/config_test.go
```diff
@@ -172,11 +173,12 @@ mithril:
 		StrictUtxoValidation: true,
 		Network:              "preview",
 		MetricsPort:          8088,
+		APIBindAddr:          DefaultAPIBindAddr,
 		DebugBindAddr:        DefaultDebugBindAddr,
 		PrivateBindAddr:      "127.0.0.1",
 		PrivatePort:          8000,
 		RelayPort:            4000,
-		CORSAllowedOrigins:   []string{"*"},
+		CORSAllowedOrigins:   nil,
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 8a7ca87a. Both `LoadConfig` call sites that assert defaults (`TestLoad_CompareFullStruct` and `TestLoad_WithoutConfigFile_UsesDefaults`) now call `unsetAPIExposureEnv(t)` alongside `unsetDebugBindAddrEnv(t)`. This was a real gap I introduced: those tests assert `APIBindAddr: DefaultAPIBindAddr` and `CORSAllowedOrigins: nil`, and neither survives an inherited `DINGO_API_BIND_ADDR` or `DINGO_CORS_ALLOWED_ORIGINS`.

docker-compose.yml
```diff
@@ -21,6 +21,11 @@ services:
       - CARDANO_NETWORK=${CARDANO_NETWORK:-preview}
       - CARDANO_DATABASE_PATH=/data/db
       - DINGO_SOCKET_PATH=/ipc/dingo.socket
+      # API listeners (Blockfrost, Mesh, UTxO RPC) bind loopback by default,
+      # which a container port mapping cannot reach. The published ports below
+      # are bound to the host's own loopback, so the container-side wildcard
+      # is not itself an exposure.
+      - DINGO_API_BIND_ADDR=0.0.0.0
```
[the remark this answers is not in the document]
chrisguiney: Correct, and the comment was wrong. Fixed in 8a7ca87a by taking the per-provider route you suggested: `DINGO_PLUGINS_API_UTXORPC_CONFIG_HOST=0.0.0.0` widens only the listener this file publishes, `apiBindAddr` stays on loopback, and Blockfrost/Mesh stay unreachable. The comment now says a container-side bind is reachable by every other container on the same Docker network instead of denying it.

Added `TestAPIPluginHostFromPluginEnvironment` to pin the path this now depends on: `plugin.ApplyEnvironment` flattens `CONFIG_HOST` to the `host` field, and only the named provider is widened.

Worth noting the distinction does not carry to Kubernetes, where each pod has its own network namespace — blinklabs-io/helm-charts#464 still recommends the pod-wide variable.

internal/config/api_bind_test.go
```diff
@@ -0,0 +1,192 @@
...
+		APIPluginHost(selection, DefaultAPIBindAddr),
+	)
+
+	noHost := hostplugin.Selection{
+		Provider: "builtin",
+		Config:   map[string]any{"port": uint(3000)},
+	}
+	require.Equal(
+		t,
+		DefaultAPIBindAddr,
```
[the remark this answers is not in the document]
chrisguiney: Not applicable here. Line 189 is inside `TestAPIPluginHostPerProviderOverride`, which calls `APIPluginHost(selection, DefaultAPIBindAddr)` directly — a pure function over an explicit fallback. It never calls `LoadConfig` and reads no environment, so no value can be inherited into it.

The underlying concern was real elsewhere, though: CodeRabbit flagged the same class in `internal/config/config_test.go`, where two `LoadConfig` tests do assert defaults without clearing the new variables. Fixed there in 8a7ca87a. Every test in `api_bind_test.go` that calls `LoadConfig` already called `unsetAPIExposureEnv(t)`.

node_api_security_test.go
```diff
@@ -250,3 +250,43 @@ func TestNewRejectsInvalidAPIAuthMode(t *testing.T) {
...
+	assert.Equal(t, "0.0.0.0", cfg.BindAddr())
+	assert.Equal(t, internalconfig.DefaultAPIBindAddr, cfg.APIBindAddr())
+	assert.Equal(t, "127.0.0.1", cfg.APIBindAddr())
+}
+
+// TestWithAPIBindAddrWidensOnlyTheAPIListeners asserts the explicit
+// opt-out an operator fronting Dingo with a proxy or a container network
+// needs, and that it does not disturb bindAddr.
+func TestWithAPIBindAddrWidensOnlyTheAPIListeners(t *testing.T) {
+	cfg := NewConfig(WithAPIBindAddr("0.0.0.0"))
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 8a7ca87a. `NewConfig(WithAPIBindAddr("192.0.2.20"))`, so a write-through to `BindAddr` would now surface as `192.0.2.20` instead of passing on the shared `0.0.0.0`. Comment records why the address is non-default.

dingo.yaml.example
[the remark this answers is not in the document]
chrisguiney: Fixed in 8a7ca87a — same change as the CodeRabbit thread on this line.

internal/config/api_bind_test.go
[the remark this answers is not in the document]
chrisguiney: Correction to my previous reply: the duplication finding on `api/utxorpc/provider_test.go:188` is also fixed in 8a7ca87a. Both that package and `api/blockfrost` now expose `resolveOnFreePortWithDeps(t, host, deps, extra)`, with `resolveOnFreePortWithConfig` delegating to it over the default dependencies — so neither new test carries its own retry loop. Blockfrost had the same duplication in my test that this comment named only in utxorpc; audited and fixed as a class. Mesh already took `deps`, so it needed no change.

api/utxorpc/provider_test.go
```diff
@@ -181,3 +181,37 @@ func TestProviderPropagatesTLSAndAuth(t *testing.T) {
 	require.True(t, srv.config.Auth.Enabled)
 	require.Equal(t, "shared-secret", srv.config.Auth.Token)
 }
+
+// TestProviderHostOverridesSharedDefault asserts the per-plugin
+// `plugins.api.utxorpc.config.host` override wins over the shared API
+// bind address composition hands down (issue #3498).
+func TestProviderHostOverridesSharedDefault(t *testing.T) {
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 8a7ca87a. Both this package and `api/blockfrost` now expose `resolveOnFreePortWithDeps(t, host, deps, extra)`, with `resolveOnFreePortWithConfig` delegating to it over the default dependencies, so neither new test carries its own retry loop.

Audited as a class rather than at this one line: `api/blockfrost` had the identical duplication in my test and is fixed the same way; `api/mesh` already took `deps` as a parameter and needed no change.

internal/config/api_bind_test.go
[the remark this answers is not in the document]
chrisguiney: Disregard my previous reply on this thread — it was meant for the duplication finding on `api/utxorpc/provider_test.go`, which is answered there. This thread's own finding stands as described above: not applicable, since line 223 is a pure-function test that reads no environment.

chrisguiney:
Pushed 8a7ca87a addressing the first bot pass: 5 findings fixed, 1 not applicable (replied on each thread).

The substantive one was Cubic's P2 on `docker-compose.yml`. It was right and my comment was wrong: a container-side wildcard bind is reachable from every other container on the same Docker network, so `DINGO_API_BIND_ADDR` there widened Blockfrost and Mesh too — neither of which this file publishes. Now only UTxO RPC is widened, via its own `DINGO_PLUGINS_API_UTXORPC_CONFIG_HOST`, with `TestAPIPluginHostFromPluginEnvironment` pinning that path.

Re-validated on the new head: `golangci-lint` 0 issues, `make docs-parity` and `make import-boundaries` pass, `-race` green on every changed package. CI's own race and cross-platform jobs are the authority on the pushed tree.

docker-compose.yml
```diff
@@ -21,6 +21,14 @@ services:
       - CARDANO_NETWORK=${CARDANO_NETWORK:-preview}
       - CARDANO_DATABASE_PATH=/data/db
       - DINGO_SOCKET_PATH=/ipc/dingo.socket
+      # API listeners bind loopback by default, which a container port
+      # mapping cannot reach. Widen only the listener this file publishes;
+      # a container-side bind is reachable by every other container on the
```
wolf31o2: This is factually wrong, and it is the reasoning `examples/docker-compose.yml` was left alone on. A container-side loopback bind is *not* reachable from other containers on the same Docker network — each container has its own network namespace and its own loopback. Verified with two containers on one user-defined network: `socat TCP-LISTEN:3000,bind=127.0.0.1` answers `wget` from inside its own container and gets `connection refused (172.18.0.2)` from a sibling container.

So the sentence needs to go, and `examples/docker-compose.yml` needs `DINGO_API_BIND_ADDR: "0.0.0.0"` on the shared `&dingo-env` anchor — its `blockfrost-explorer` talks to `http://dingo:3000`.

docker-compose.yml
```diff
@@ -21,6 +21,14 @@ services:
...
+      # same Docker network, so apiBindAddr stays on loopback and
+      # Blockfrost/Mesh stay unreachable rather than being widened along
+      # with it. Set DINGO_PLUGINS_API_<NAME>_CONFIG_HOST per listener, or
+      # DINGO_API_BIND_ADDR to widen all three at once.
+      - DINGO_PLUGINS_API_UTXORPC_CONFIG_HOST=0.0.0.0
```
wolf31o2: Inert as shipped: this file leaves `storageMode` at its `core` default, and all three API listeners are gated on `n.config.storageMode.IsAPI()` (`node.go:1419`, `:1552`, `:1586`), so no UTxO RPC listener is started here at all — the published `127.0.0.1:9090:9090` above was already dead. Either add `DINGO_STORAGE_MODE=api` so the comment and the port publish mean something, or drop both and say why the relay profile has no API.

internal/config/redact.go
```diff
@@ -116,6 +116,7 @@ var logPlainConfigFields = []string{
 	"BarkHost",
 	"BarkOperatorCertificateFingerprints",
 	"BarkPort",
+	"APIBindAddr",
```
wolf31o2: `logPlainConfigFields` is otherwise alphabetical; this lands between `BarkPort` and `BindAddr`.

README.md
```diff
@@ -305,6 +305,44 @@ plugins:
...
+
+`plugins.api.<name>.config.host` overrides `apiBindAddr` for that provider
+alone; the CLI (`--api-bind-addr`) and environment
+(`DINGO_API_BIND_ADDR`) set the shared default.
+
+`corsAllowedOrigins` likewise defaults to an empty list, which sends no CORS
+headers at all. Set it explicitly — to specific origins, or to `["*"]` — to
+allow browser access.
+
+> **Upgrading:** a deployment that relied on the previous `0.0.0.0` API
```
wolf31o2, line 383: Worth naming the container case explicitly here, since it is the one that silently breaks: publishing a port from a container does not reach a listener bound to that container's loopback, so a containerised deployment needs `DINGO_API_BIND_ADDR` (or a per-provider `host`) set to `0.0.0.0` inside the container regardless of how the host-side mapping is written.

wolf31o2 (changes requested):
Right change, one proven deployment break.

Blocking: `examples/docker-compose.yml` is not updated and stops working. Its `blockfrost-explorer` service reaches Dingo at `http://dingo:3000` across the compose network (L104), and its three published API ports (3000/9090/8080, L70-72) map to container-side listeners that now bind `127.0.0.1`. A loopback bind inside a container is not reachable from another container on the same Docker network — verified, connection refused. The `&dingo-env` anchor needs `DINGO_API_BIND_ADDR: "0.0.0.0"`, and `examples/README.md:54` / `examples/dingo-blockfrost-explorer/README.md:84` ("Set `DINGO_BIND_ADDR=127.0.0.1` if Dingo API ports should be local-only") go stale with it.

The rest verified good: default binds loopback for all three listeners and wildcard only for the one that opts in; an explicitly empty `DINGO_API_BIND_ADDR` still resolves to loopback; relay and metrics still take `bindAddr`; `host` is already in `providerConfigPlainKeys`, so the override is not redacted out of a startup log; empty `corsAllowedOrigins` returns the handler unwrapped, no headers.

internal/config/api_bind_test.go
```diff
@@ -0,0 +1,226 @@
...
+	} {
+		t.Setenv(name, "")
+		require.NoError(t, os.Unsetenv(name))
+	}
+}
+
+// TestAPIBindAddressDefaultsToLoopback pins the fix for #3498: the
+// Blockfrost/Mesh/UTxO-RPC listeners bind loopback by default, and do so
+// independently of bindAddr, which stays on the wildcard the relay/NtN
+// and metrics listeners need.
```
wolf31o2, line 44: Coverage of the new default and of override precedence is good. Three cases missing:

1. An *explicitly empty* `apiBindAddr` — `apiBindAddr: ""` in YAML, or `DINGO_API_BIND_ADDR=`. That is what the `ApplyDefaults` and `syncCompatFields` fallbacks exist for, and it is the branch whose deletion turns the listener back into a wildcard. Nothing pins it; the absent-key case here does not reach it. (I confirmed the binary resolves it to `127.0.0.1`, so the behaviour is right — it is just unguarded.)
2. `node_lifecycle.go`'s `reinitializeAPIServers` — the second call site changed identically to `node.Run`, and no test covers either node call site; all coverage stops at `buildDingoConfig` and the provider factories.
3. The over-application direction: nothing asserts the relay and metrics listeners still take `bindAddr` after the split.

wolf31o2 (commented):
Test gaps, named inline. Not blocking on their own.

docker-compose.yml
chrisguiney: Confirmed by reproduction, then fixed. Two containers on one user-defined bridge network, the server running `socat TCP-LISTEN:3000,bind=127.0.0.1` alongside a `TCP-LISTEN:3001,bind=0.0.0.0` control in the same container:

| from | to | result |
|---|---|---|
| server container itself | `127.0.0.1:3000` | served |
| sibling container | `dingo-lb-srv:3000` | `can't connect to remote host (172.18.0.2): Connection refused` |
| sibling container | `dingo-lb-srv:3001` (control) | served |
| host, via `-p 127.0.0.1:13000:3000` | `127.0.0.1:13000` | `curl: (52) Empty reply from server` |
| host, via `-p 127.0.0.1:13001:3001` (control) | `127.0.0.1:13001` | served |

`netstat` inside the container confirmed the two listeners were `127.0.0.1:3000` and `0.0.0.0:3001`. So the sentence is wrong on both counts and has been removed.

`examples/docker-compose.yml` now sets `DINGO_API_BIND_ADDR: "0.0.0.0"` on the shared `&dingo-env` anchor, with a comment naming both hops it broke (`blockfrost-explorer` → `http://dingo:3000`, `sundae-preview` → `http://dingo:9090`) and the published ports.

I audited the whole class rather than the one file — every compose file, example, Helm/Kubernetes manifest and script in the repo that publishes a Dingo API port or points another service at one:

| Location | Verdict |
|---|---|
| `examples/docker-compose.yml` | **Broken** — fixed, `DINGO_API_BIND_ADDR: "0.0.0.0"` on `&dingo-env` |
| `examples/dingo-blockfrost-explorer/k8s/dingo-values.yaml` | **Broken** — ClusterIP Service + `blockfrost-loadbalancer.yaml` target the pod IP, not its loopback; fixed |
| `examples/dingo-sundae-preview/k8s/dingo-values.yaml` | **Broken** — LoadBalancer Service on UTxO RPC 9090; fixed |
| `docker-compose.yml` (root) | Dead, not broken — see the other thread |
| `examples/dingo-gov-lens/k8s/dingo-values.yaml` | Unaffected — all three API ports are `"0"`, and `node.go:1419/1552/1586` gate on `port > 0`, so no listener starts |
| `examples/dingo-gov-lens/scripts/{run-dingo,mithril-sync}.sh` | Unaffected — same, API ports pinned to `0` |
| `examples/dingo-blockfrost-explorer/scripts/run-dingo.sh` + `run-dev.sh` | Unaffected — Dingo runs natively on the host, the Vite proxy dials `127.0.0.1:3000` on that same host |
| `examples/dingo-sundae-preview/scripts/run-dev.sh`, `vite.config.ts` | Unaffected — same host-native loopback path |
| `examples/dingo-sundae-preview/scripts/port-forward-dingo.sh` | Unaffected — `kubectl port-forward` dials from inside the pod's own network namespace; the LoadBalancer path beside it is what needed the fix |
| `examples/.env.example` | Unaffected — `DINGO_BIND_ADDR` there is only substituted into the host side of the `ports:` mappings, it is not in any `environment:` block, so it never reaches the container |
| `internal/test/{devnet,conformance,erastest,archive-demo,antithesis}/docker-compose*.y*ml` | Unaffected — none publish or consume an API port; they use relay/private ports and `DINGO_PRIVATE_BIND_ADDR`/`CARDANO_PRIVATE_BIND_ADDR`, which this change does not touch |
| `Dockerfile` | Unaffected — `EXPOSE` binds nothing, and the healthcheck hits `127.0.0.1:12798` (metrics, on `bindAddr`) |

The stale docs went with it: `examples/README.md` and `examples/dingo-blockfrost-explorer/README.md` now say that `DINGO_BIND_ADDR`/`BLOCKFROST_EXPLORER_BIND_ADDR` are *host* publish addresses, not the container-side bind, and that the stack sets `DINGO_API_BIND_ADDR=0.0.0.0` inside the container.

One class member outside this repo: the `blinklabs-io/helm-charts` Dingo chart. Its own default needs the same treatment, but it is a separate repository and not in this PR's scope.

docker-compose.yml
chrisguiney: Correct, and I took the second option: dropped both, with the reason stated in the file.

Confirmed the gate first — `node.go:1419`, `:1552`, `:1586` are all `n.config.storageMode.IsAPI() && <port> > 0`, and this file never sets `DINGO_STORAGE_MODE`, so `core` is in force and none of the three listeners starts. The `127.0.0.1:9090:9090` publish was indeed already dead.

I did not add `DINGO_STORAGE_MODE=api` because that is not a comment fix, it is a profile change: `api` mode stores full transaction, witness, script, datum, redeemer, governance and metadata history, so the quickstart's disk footprint and first-sync time both grow substantially. Making that happen as a side effect of a security-defaults change seemed wrong, and `examples/docker-compose.yml` already exists as the API-node example. The file also self-describes as a "Basic relay node".

So `DINGO_PLUGINS_API_UTXORPC_CONFIG_HOST` and the `127.0.0.1:9090:9090` publish are both gone, and the service comment now says why the relay profile has no API and exactly what turning one on would need:

```yaml
  # Basic relay node
  #
  # No API listener runs here. storageMode stays at its "core" default, and
  # the Blockfrost, Mesh, and UTxO RPC listeners only start under storageMode
  # "api". Turning one on needs DINGO_STORAGE_MODE=api and a container-side
  # wildcard bind -- DINGO_API_BIND_ADDR=0.0.0.0, or
  # DINGO_PLUGINS_API_<NAME>_CONFIG_HOST=0.0.0.0 for a single listener --
  # because those listeners bind loopback by default, and a container's
  # loopback is reachable only from inside that container: neither a
  # published port nor a sibling container on the same Docker network can
  # reach it. examples/docker-compose.yml is a working API node.
```

The metrics publish stays: metrics binds `bindAddr`, which is still the `0.0.0.0` default, so `127.0.0.1:12798:12798` works and the host-side restriction is real. I reworded its comment from "localhost only by default" to "localhost only on the host" so it is not read as a container-side claim.

`TestAPIPluginHostFromPluginEnvironment` kept its coverage of the per-provider environment override, but its doc comment no longer claims `docker-compose.yml` depends on that path, since it no longer does.

internal/config/redact.go
chrisguiney: Fixed. `APIBindAddr` now sits between `API.TLS.Mode` and `ActivePeersGossipQuota`, which is where ASCII order puts it: `.` (0x2E) < `B` (0x42) keeps the `API.*` entries first, and `P` (0x50) < `c` (0x63) keeps `APIBindAddr` ahead of `ActivePeers*`.

README.md
chrisguiney: Added. The upgrading note now carries a second paragraph that names the container case as its own failure mode rather than as a parenthetical:

> **Containers and pods break silently.** A container has its own loopback, so a listener bound to `127.0.0.1` inside it is unreachable from everything outside: publishing the port (`-p 9090:9090`, or `-p 127.0.0.1:9090:9090`) does not reach it, another container on the same Docker network does not reach it, and a Kubernetes Service does not reach it either, because it targets the pod IP. A containerised deployment therefore needs `DINGO_API_BIND_ADDR=0.0.0.0` — or a per-provider `host` — set *inside* the container, however the host-side mapping is written. Restricting exposure is then the host mapping's job (`-p 127.0.0.1:9090:9090`), or a NetworkPolicy's, plus `api.auth`.

The Kubernetes half is in there because the class audit turned up two Helm values files with the same break, so it is not a Docker-only footgun.

internal/config/api_bind_test.go
chrisguiney: All three added, each proven to fail with the fix reverted in place rather than by testing against `main`.

**(1) Explicitly empty `apiBindAddr`** — `TestAPIBindAddressExplicitlyEmptyResolvesToLoopback` in `internal/config/api_bind_test.go`, table-driven over both entry points: `apiBindAddr: ""` in YAML and `DINGO_API_BIND_ADDR=` present-but-empty in the environment. It asserts the value really survives loading as empty first (otherwise the test would not reach the fallback at all), then checks both branches: `APIListenHost` *before* `ApplyDefaults`, and `APIBindAddr` after it. Confirmed empirically that envconfig does honour a present-but-empty variable, so that half is a real path and not a hypothetical.

The programmatic twin is `TestAPIBindAddrExplicitlyEmptyResolvesToLoopback` in `node_api_security_test.go`: `NewConfig(WithBindAddr("0.0.0.0"), WithAPIBindAddr(""))` must give `APIBindAddr() == "127.0.0.1"` and leave `BindAddr()` on the wildcard — that is the `syncCompatFields` branch.

| revert | test | result |
|---|---|---|
| drop `ApplyDefaults`'s `if c.APIBindAddr == ""` block | `TestAPIBindAddressExplicitlyEmptyResolvesToLoopback` | exit 1, `-127.0.0.1 / +` |
| drop `APIListenHost`'s `if fallback == ""` block | `TestAPIBindAddressExplicitlyEmptyResolvesToLoopback` | exit 1 |
| neuter `syncCompatFields`'s `if c.apiBindAddr == ""` | `TestAPIBindAddrExplicitlyEmptyResolvesToLoopback` | exit 1 |

**(2) Both node call sites** — `TestAPIProviderDependenciesUseAPIBindAddr` in the new `node_api_bind_callsites_test.go`. It is a source guard, following the precedent in `ledger/publish_under_lock_test.go` and `ledger/chainsync_ordering_test.go`, because neither call site is reachable at runtime without a started node: `Run` needs a live network stack, and `reinitializeAPIServers` needs an open database (it reaches `n.db.Metadata()` for the off-chain metadata fetcher), a resolved plugin host and `storageMode: api`. It parses `node.go` and `node_lifecycle.go`, finds every `blockfrost`/`mesh`/`utxorpc.ProviderDependencies` composite literal, and requires each one's `Host` to be exactly `n.config.apiBindAddr` — plus a count of 3 per file, so a call site that stops setting `Host` at all (which would bind the empty wildcard) is caught too.

| revert | result |
|---|---|
| `node_lifecycle.go` utxorpc `Host` → `n.config.bindAddr` | exit 1, `utxorpc.ProviderDependencies at node_lifecycle.go:1098:4 must bind apiBindAddr` |
| `node.go` blockfrost `Host` → `n.config.bindAddr` | exit 1, `blockfrost.ProviderDependencies at node.go:1566:4 must bind apiBindAddr` |

**(3) The over-application direction** — two cases in `TestValidate` (`internal/config/validate_test.go`), using distinct `bindAddr`/`apiBindAddr` so the two addresses are discriminable:

- `relay and metrics still contend on bindAddr`: with `bindAddr=127.0.0.2`, `apiBindAddr=127.0.0.3` and `metricsPort == relayPort`, the collision must still be reported — which is only true while both entries read `bindAddr`.
- `api listener may share the relay port`: same two addresses, Blockfrost on the relay's own port, and no error — which fails if an API entry moves back onto `bindAddr`, and equally if the relay entry moves onto `apiBindAddr`.

Taken together they pin all three directions rather than just one.

| revert | result |
|---|---|
| `utxorpcHost`/`blockfrostHost`/`meshHost` → `c.BindAddr` | `TestValidate/api_listener_may_share_the_relay_port` exit 1 |
| metrics entry host → `c.APIBindAddr` | `TestValidate/relay_and_metrics_still_contend_on_bindAddr` exit 1 |

One thing the audit turned up that is worth recording but that I did not act on: `dingo.Config.bindAddr` — the unexported compat mirror set in `syncCompatFields` — now has no reader left in the root package, since all six consumers moved to `apiBindAddr`. It is harmless and matches the other compat mirrors, so I left it rather than widening this change, but it is dead as of this PR.

chrisguiney:
Pushed `e47d15f` addressing the blocker and all four inline findings.

**Blocker.** Reproduced the container-loopback claim before acting on it (sibling container gets `Connection refused (172.18.0.2)`; a published port gets `curl: (52) Empty reply from server`; the `0.0.0.0` control is reachable on both paths). Fixed and audited as a class — every compose file, example, Helm manifest and script in the repo that publishes or consumes a Dingo API port. Three were broken, and all three are fixed; the rest are unaffected for stated reasons, detailed in the `docker-compose.yml` thread.

Changed paths:

- `examples/docker-compose.yml` — `DINGO_API_BIND_ADDR: "0.0.0.0"` on the shared `&dingo-env` anchor
- `examples/dingo-blockfrost-explorer/k8s/dingo-values.yaml`, `examples/dingo-sundae-preview/k8s/dingo-values.yaml` — same, a Service targets the pod IP and not its loopback
- `docker-compose.yml` — removed the incorrect comment, the dead `127.0.0.1:9090:9090` publish and the API host override; the relay profile's `core` storage mode starts no API listener
- `README.md`, `examples/README.md`, `examples/dingo-blockfrost-explorer/README.md`, `examples/dingo-sundae-preview/README.md` — container/pod case named explicitly; host-side publish address distinguished from the container-side bind
- `internal/config/redact.go` — `APIBindAddr` into alphabetical position

Tests added, each proven to fail with the corresponding fix reverted in place:

- `internal/config/api_bind_test.go` — explicitly empty `apiBindAddr` via YAML and via the environment
- `node_api_security_test.go` — `WithAPIBindAddr("")` through `syncCompatFields`
- `node_api_bind_callsites_test.go` — source guard that `Run` and `reinitializeAPIServers` both pass `apiBindAddr` to all three provider factories
- `internal/config/validate_test.go` — relay and metrics still resolve to `bindAddr`; an API listener may take the relay's port on a distinct `apiBindAddr`

Validation: `go build ./...` (0), `go vet ./...` (0), `go test -race` on `.`, `./internal/config/...`, `./internal/node/...`, `./api/...`, `./internal/apiconfig/...` (0), `go test ./internal/docsparity` (0), `go test ./internal/architecture` (0), `golangci-lint run ./ ./internal/config/...` (0 issues), `docker compose config -q` on both compose files (0). `origin/main` has moved since the merge base; the merged tree builds and its tests pass. `ARCHITECTURE.md`'s API-security section already covers this mechanism and needed no further change; `DATABASE.md` is unaffected.

docker-compose.yml
```diff
@@ -10,6 +10,16 @@
 services:
   # --------------------------------------------------------------------------
   # Basic relay node
+  #
+  # No API listener runs here. storageMode stays at its "core" default, and
+  # the Blockfrost, Mesh, and UTxO RPC listeners only start under storageMode
+  # "api". Turning one on needs DINGO_STORAGE_MODE=api and a container-side
```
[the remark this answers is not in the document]
chrisguiney: Verified and fixed in 29f12aad — I took the reword option, not the forwarding option.

You're right that neither variable reaches the container. Rendered at `e47d15fe`:

```
$ docker compose -f docker-compose.yml config
    environment:
      CARDANO_DATABASE_PATH: /data/db
      CARDANO_NETWORK: preview
      DINGO_SOCKET_PATH: /ipc/dingo.socket
```

One correction to the framing: the comment never printed a `docker compose up` command line, so that invocation is your paraphrase. It is a fair one — the file header establishes exactly that idiom (`CARDANO_NETWORK=mainnet docker compose up -d`) and `CARDANO_NETWORK` *is* forwarded, via `${CARDANO_NETWORK:-preview}` — so a reader who follows the established pattern for the two variables this comment names gets a container that silently ignores both. The defect is real regardless.

I chose rewording because forwarding cannot make the documented path work. `ports:` publishes only `3001` and `127.0.0.1:12798`, and Compose cannot add a port publish from an environment variable. Forwarding both names would therefore start Blockfrost/Mesh/UTxO RPC on `0.0.0.0` inside the container with no published port — still unreachable, and failing in a harder-to-diagnose way than the current "nothing happened at all". It also cannot cover the third name the comment gives, `DINGO_PLUGINS_API_<NAME>_CONFIG_HOST`, which is a placeholder. The comment now states that turning a listener on means editing this file in two places (environment and ports) and points at `examples/docker-compose.yml` as the working API node.

I deliberately did not add commented-out API port publishes: the previous round removed the dead `127.0.0.1:9090:9090` line for being unreachable in core mode, and re-adding it in commented form would walk that back.

**Class audit** — every place in the repo that documents an environment-variable invocation of a compose file, checked against what that file actually interpolates or forwards:

| file | result |
|---|---|
| `docker-compose.yml` | 2 broken (this finding), plus 1 more below |
| `examples/docker-compose.yml` + `.env.example` + `README.md` | clean — all 13 `.env.example` names interpolated, as are the README's `GOV_LENS_PORT` / `BLOCKFROST_EXPLORER_PORT` / `SUNDAE_PREVIEW_PORT` / `DINGO_IMAGE`; `DINGO_API_BIND_ADDR` is hardcoded in the shared anchor, exactly as the README describes |
| `internal/test/devnet/docker-compose.yml` | clean — `COMPOSE_PROFILES` is a Compose built-in, and `DEVNET_DINGO_NTC_ADDR` is documented as test-side (read by `endpoints_conformance.go`) |
| `internal/test/archive-demo/docker-compose.yml` | clean, 6/6 |
| `internal/test/erastest/docker-compose.yml` + `.vanrossem.yml` | clean, 3/3 |
| `internal/test/antithesis/docker-compose.yaml` | clean — documents no env-var invocation |
| `internal/test/antithesis/testnets/dingo-praos/docker-compose.yaml` | clean, 3/3 |
| `internal/test/conformance/docker-compose.yml` | clean — `POSTGRES_HOST` / `MYSQL_PORT` and friends are documented on the `go test` line and consumed by the test binary, not by Compose |

The extra member: the header's `RESTORE_SNAPSHOT=1 docker compose up -d` on line 8 is the identical trap, and it predates this PR — `RESTORE_SNAPSHOT` appeared only as a commented-out `# - RESTORE_SNAPSHOT=1`, so that documented command silently did nothing. Unlike the API listeners it needs no port publish, so forwarding genuinely fixes it. I used a bare name rather than `${RESTORE_SNAPSHOT:-}`, which would have injected an empty string:

```
# host var unset
      RESTORE_SNAPSHOT: null     <- omitted from the container
# RESTORE_SNAPSHOT=1 on the host
      RESTORE_SNAPSHOT: "1"
```

So nothing is promoted by default, and `bin/entrypoint.sh:256` gates on `[[ -n "${RESTORE_SNAPSHOT:-}" ]]` anyway.

`make docs-parity` passes. No Go code changed.

arepala-uml (commented):
No independent code blockers found -- build/vet/test -race/golangci-lint all clean, and I checked every prior CHANGES_REQUESTED finding (wolf31o2, CodeRabbit, cubic: container-loopback unreachability, dead port publish, redact ordering, missing test cases, docs, test isolation, test duplication) against the current head directly rather than trusting the threads, and all of them are genuinely fixed -- several with concrete reproduction evidence in the replies. The formal CHANGES_REQUESTED state looks stale relative to the code; it persists because the reviewer hasn't re-reviewed on GitHub yet.

Flagging one thing that isn't a defect in this PR but does need a maintainer decision before merging either PR in this area: #4099 ("fix(config): require authentication for remote APIs") adds the *identical* new field -- same name (`APIBindAddr`), same YAML tag, same env var (`DINGO_API_BIND_ADDR`), same default (`127.0.0.1`), same `WithAPIBindAddr()` helper -- as an independent implementation of the same underlying problem (issue #3498 here, #3656 there). Merging one and rebasing the other will conflict directly in `config.go`, `internal/config/config.go`, `redact.go`, `node.go`, and `ARCHITECTURE.md`.

They're complementary rather than fully redundant: this PR adds the per-provider `plugins.api.<name>.config.host` override and the CORS-default fix that #4099 doesn't have; #4099 adds `ValidateAPIExposure`, a startup gate that actually *requires* auth if bound beyond loopback, which this PR doesn't have. But if #4099 is merged as-is on top of this one, its `ValidateAPIExposure` reads raw `APIBindAddr` only -- it doesn't know about this PR's per-provider `host` override, so a provider widened only via its own `host` while `apiBindAddr` stays loopback would silently bypass #4099's auth requirement.

Suggested sequencing: merge this PR first (it's the more complete, already-vetted bind-address mechanism), then have #4099 rebase to add only its net-new auth-requirement gate, updated to route through the per-provider override so there's no bypass gap. I've left the same note on #4099 for visibility.

arepala-uml (dismissed):
Upgrading my earlier comment to an approval. On reflection, nothing about this PR's own code is unsound or needs to change for it to merge safely: build/vet/test -race/golangci-lint are all clean, and every prior CHANGES_REQUESTED finding (wolf31o2, CodeRabbit, cubic) is verifiably fixed at this head.

To be clear about the #4099-overlap note in my previous comment: that is not a blocker on this PR. It's information for whichever maintainer sequences #3929 and #4099 -- the fix needed for the two to compose safely (routing #4099's `ValidateAPIExposure` through the per-provider host override) belongs on #4099's side, not here. This PR is self-contained and correct as it stands.

Approving.

docker-compose.yml
chrisguiney: Fixed and confirmed at the current head by running it.

`examples/docker-compose.yml:36` is the `&dingo-env` anchor and `:55` sets
`DINGO_API_BIND_ADDR: "0.0.0.0"` on it, so both `dingo-sync` and `dingo`
(`:77`, `<<: *dingo-env`) inherit it.

Reachability, running that environment against this branch's binary with two
services on one Compose network:

| configuration | container-side bind | sibling → `http://dingo:3000/health` | published host port |
|---|---|---|---|
| as committed (`DINGO_API_BIND_ADDR: "0.0.0.0"`) | `:::3000`, `:::9090` | `200 OK`, `{"is_healthy":true}` | `200` |
| control, that one line removed | `127.0.0.1:3000`, `127.0.0.1:9090` | `Connection refused (172.19.0.2)` | `curl: (52) Empty reply from server` |

In the control the listener is live — `http://127.0.0.1:3000/health` from
inside that same container returns `{"is_healthy":true}` — so the failure is
the bind scope, not a dead listener. Relay `3001` and metrics `12798` bound
`:::` in both runs, confirming `bindAddr` is untouched by the split.

The stale docs went with it: `examples/README.md:54-60` and
`examples/dingo-blockfrost-explorer/README.md:86-92` now describe
`DINGO_BIND_ADDR` and `BLOCKFROST_EXPLORER_BIND_ADDR` as the *host* publish
addresses and state that the stack sets `DINGO_API_BIND_ADDR=0.0.0.0` inside
the container.

docker-compose.yml
chrisguiney: Fixed at the current head. `DINGO_PLUGINS_API_UTXORPC_CONFIG_HOST` and the
`127.0.0.1:9090:9090` publish are both gone; `docker-compose.yml:14-27` now
states why the relay profile has no API listener and exactly what enabling
one requires (`DINGO_STORAGE_MODE=api` plus a container-side wildcard bind).

The gate still reads as described: `node.go:1439`, `:1572` and `:1606` are
each `n.config.storageMode.IsAPI() && <port> > 0`, and this file sets no
`DINGO_STORAGE_MODE`, so `core` is in force and none of the three listeners
starts.

The metrics publish stays. Metrics binds `bindAddr`, which this change does
not touch, so `127.0.0.1:12798:12798` still works and the host-side
restriction is real; its comment now reads "localhost only on the host" so it
is not mistaken for a container-side claim.

internal/config/redact.go
chrisguiney: Fixed. `internal/config/redact.go:111` puts `APIBindAddr` between
`API.TLS.Mode` and `ActivePeersGossipQuota`, which is where ASCII order puts
it: `.` (0x2E) < `B` (0x42) keeps the `API.*` entries ahead of it, and
`P` (0x50) < `c` (0x63) keeps it ahead of `ActivePeers*`.

README.md
chrisguiney: Added. `README.md:345-352` carries the container case as its own paragraph
rather than a parenthetical, naming all three hops that fail — a published
port, a sibling container on the same Docker network, and a Kubernetes
Service targeting the pod IP — and stating that the wildcard bind has to be
set *inside* the container however the host-side mapping is written.

The Kubernetes half is there because the class audit found the same break in
two Helm values files, so it is not Docker-specific.

internal/config/api_bind_test.go
chrisguiney: All three added.

1. Explicitly empty `apiBindAddr` — `TestAPIBindAddressExplicitlyEmptyResolvesToLoopback`
   (`internal/config/api_bind_test.go:236`) covers both entry points,
   `apiBindAddr: ""` in YAML and a present-but-empty `DINGO_API_BIND_ADDR`.
   `TestAPIBindAddrExplicitlyEmptyResolvesToLoopback`
   (`node_api_security_test.go:317`) is the programmatic twin over
   `syncCompatFields`.
2. Both node call sites — `TestAPIProviderDependenciesUseAPIBindAddr`
   (`node_api_bind_callsites_test.go:60`) parses `node.go` and
   `node_lifecycle.go`, requires every `blockfrost`/`mesh`/`utxorpc`
   `ProviderDependencies` literal to bind `n.config.apiBindAddr`, and counts
   three per file so a call site that stops setting `Host` at all is caught
   too. It is a source guard because neither call site is reachable without a
   started node, following `ledger/publish_under_lock_test.go`.
3. The over-application direction — `internal/config/validate_test.go:283`
   and `:297` use distinct `bindAddr`/`apiBindAddr` values so the two
   addresses are discriminable: the relay/metrics collision must still be
   reported, and an API listener sharing the relay port must not be.

Each fails with its fix reverted in place. Two re-confirmed at the current
head: removing `ApplyDefaults`' `if c.APIBindAddr == ""` block fails
`TestAPIBindAddressExplicitlyEmptyResolvesToLoopback` with
`expected: "127.0.0.1", actual: ""`; flipping `node_lifecycle.go`'s utxorpc
`Host` back to `n.config.bindAddr` fails
`TestAPIProviderDependenciesUseAPIBindAddr` with
`utxorpc.ProviderDependencies at node_lifecycle.go:1102:4 must bind apiBindAddr`.

chrisguiney:
### Container reachability

`examples/docker-compose.yml:55` sets `DINGO_API_BIND_ADDR: "0.0.0.0"` on the
`&dingo-env` anchor (`:36`), which both `dingo-sync` and `dingo` inherit.
Confirmed by running that environment rather than by reading it: a sibling
service on the same Compose network gets `200 OK` and `{"is_healthy":true}`
from `http://dingo:3000/health`. With that one line removed the same request
gets `Connection refused`, the published host port returns `Empty reply from
server`, and `http://127.0.0.1:3000/health` inside the container still
answers — so the control fails on bind scope, not on a dead listener. Relay
`3001` and metrics `12798` bind `:::` either way. Detail in the resolved
thread.

### Class audit

Every place in this repository that publishes or consumes one of the three
API ports:

| Location | Verdict |
|---|---|
| `examples/docker-compose.yml` | Was broken — fixed |
| `examples/dingo-blockfrost-explorer/k8s/dingo-values.yaml` | Was broken — fixed; the ClusterIP Service and `blockfrost-loadbalancer.yaml` target the pod IP |
| `examples/dingo-sundae-preview/k8s/dingo-values.yaml` | Was broken — fixed; LoadBalancer on UTxO RPC `9090` |
| `docker-compose.yml` | Not broken — `storageMode` stays `core`, so no API listener starts; the already-dead `9090` publish and its `_CONFIG_HOST` were dropped |
| `examples/dingo-gov-lens/k8s/dingo-values.yaml` | Unaffected — all three API ports are `"0"` |
| `internal/test/{devnet,conformance,erastest,archive-demo,antithesis}` compose files | Unaffected — they publish relay `3001`, private, and infrastructure ports only |
| `examples/*/vite.config.ts` and `scripts/` | Unaffected — Vite proxies server-side with `changeOrigin`, so no browser talks to Dingo directly and nothing depends on a container-side API bind |
| `Dockerfile` | Unaffected — `EXPOSE` binds nothing, and the healthcheck uses metrics on `bindAddr` |
| `examples/.env.example` | Unaffected — `DINGO_BIND_ADDR` is substituted only into the host side of `ports:` |

Complete within this repository. One member sits outside it: the
`blinklabs-io/helm-charts` Dingo chart carries the same default and needs the
same treatment there.

### Base merge

`origin/main` is merged in at `19f0cdf6`, with no conflicts, for the CI
duration work in #4131 and #4086. `main` does not touch the API listener bind
sites, the `internal/config` defaults, or the example Compose and Helm values
this branch changes; its `examples/` commits are dependency bumps. Build,
vet, `-race` across the changed packages, and `golangci-lint` are clean on the
merged tree.

The head moved after the approval on record, and GitHub re-stamped that
approval onto the new SHA instead of dismissing it, so it now claims a tree
that was never reviewed. Re-review is requested for that reason.

### Overlap with #4099 — not a blocker here

#4099 independently implements the same split — `WithAPIBindAddr`,
`APIBindAddr()`, `DefaultAPIBindAddr`, the `syncCompatFields` loopback
fallback — and adds `ValidateAPIExposure` on top. The two changes touch ten of
the same files, so whichever lands second needs rework.

One concrete composition gap for that side: `validateAPIExposure` returns
`nil` as soon as the shared `APIBindAddr` is loopback, before it examines any
provider. Combined with the per-provider override here, `apiBindAddr:
127.0.0.1` together with `plugins.api.blockfrost.config.host: "0.0.0.0"` is a
remotely reachable listener that the guard passes. Routing it through the
per-provider host resolution closes that, and belongs on #4099. Nothing in
this pull request needs to change for it.

chrisguiney:
The red `go-test` runs at `19f0cdf6` were base-branch failures, not this
change:

- `TestCanonicalFrontierCrossingDoesNotCloseAPeerAheadOfLocalTip` (`ledger`) —
  all four platforms
- `TestCanonicalPeersWithCrossingFrontiersDoNotFlap` (`chainselection`) — macOS
- `TestFetchEndorserBlockByPointDeadlineDoesNotCoolDownPeer` (`ouroboros`) —
  Windows

None sit in a package this branch touches; its changes are confined to
`internal/config`, `internal/apiconfig`, `api/`, `internal/node`, the root
package, and documentation and examples. The frontier-flap pair was made
deterministic on `main` by #4144, which `19f0cdf6` predated. The current head
`43cae786` includes it and all thirteen checks pass.

docker-compose.yml
chrisguiney: Still fixed, and now carries a second requirement.

`examples/docker-compose.yml:36` is the `&dingo-env` anchor and `:55` sets
`DINGO_API_BIND_ADDR: "0.0.0.0"` on it, which `dingo` inherits at `:77`.

`origin/main`'s API exposure policy then refuses that bind, so the stack
exited at startup once main was merged in:

```
ERROR invalid configuration: plugins.api.blockfrost.config binds to
non-loopback address "0.0.0.0" without authentication: configure api.auth
or a provider auth policy
```

The anchor now also sets `DINGO_API_AUTH_MODE: token` and
`DINGO_API_AUTH_TOKEN_FILE_PATH: /run/secrets/dingo-api-token`, fed by a
Compose `configs` entry whose content is `DINGO_API_TOKEN`. It has to be a
file: `api.auth.token` carries `ignored:"true"` precisely so an inline
secret cannot come from the environment. Both Vite dev servers attach the
same token as `Authorization: Bearer` on proxied requests, so the browser
never sees it.

The comment on this file no longer claims the container-side bind is not an
exposure, and now also states that it requires `api.auth`.

docker-compose.yml
chrisguiney: Unchanged and still correct at the current head. `docker-compose.yml:14-27`
states why the relay profile starts no API listener and what enabling one
requires; `DINGO_PLUGINS_API_UTXORPC_CONFIG_HOST` and the
`127.0.0.1:9090:9090` publish are both gone.

One sentence added this round: a container-side wildcard is a non-loopback
bind, so it needs `api.auth` as well — startup refuses an unauthenticated
one.

internal/config/redact.go
chrisguiney: Survived the `origin/main` merge. `main` inserted its own `APIBindAddr`
entry between `API.TLS.KeyFilePath` and `API.TLS.Mode`, so the auto-merge
produced two entries; the duplicate was dropped and the remaining one is
back between `API.TLS.Mode` and `ActivePeersGossipQuota`, where ASCII order
puts it.

README.md
chrisguiney: Still present, with one correction this round. The paragraph previously
ended by making `api.auth` one of several optional mitigations. Since
`origin/main` now refuses an unauthenticated non-loopback API bind, and a
container bind is a non-loopback bind, it reads:

> Restricting exposure is then the host mapping's job
> (`-p 127.0.0.1:9090:9090`), or a NetworkPolicy's. `api.auth` is required
> either way, because a container bind is a non-loopback bind.

The section's YAML example gained an `api.auth` block for the same reason —
as written it configured a wildcard bind with authentication disabled,
which no longer starts.

internal/config/api_bind_test.go
chrisguiney: All three still covered, and the merge with `origin/main` added a fourth
case in the same family.

`ValidateAPIExposure` arrived on `main` reading only the shared
`apiBindAddr`, which this branch's `plugins.api.<name>.config.host`
override makes insufficient in both directions: a provider widened by its
own `host` escaped the authentication requirement while `apiBindAddr` was
loopback, and a provider pinned to loopback by its own `host` was required
to authenticate while `apiBindAddr` was remote. It now resolves through
`APIListenHost`, the same precedence the provider factories and the
port-conflict check already use.

`TestValidateAPIExposureUsesPerProviderHost` and
`TestValidateAPIExposureAllowsProviderPinnedToLoopback` pin both
directions; each fails without the change, the first seeing no error and
the second seeing one for a loopback-bound provider.

chrisguiney:
Merged current `origin/main` and addressed the blocking review.

**Blocker: `examples/docker-compose.yml`.** Fixed. The `&dingo-env` anchor
sets `DINGO_API_BIND_ADDR: "0.0.0.0"`, which both Dingo services inherit.

**Merge with `origin/main`.** `main` landed `APIBindAddr`,
`DefaultAPIBindAddr`, `--api-bind-addr`, `WithAPIBindAddr` and
`ValidateAPIExposure` independently of this branch. Six files conflicted;
resolutions are in the merge commit body. Git also auto-merged both sides'
convergent additions without conflict, leaving duplicate declarations of
`DefaultAPIBindAddr`, the `APIBindAddr` struct field, its `globalConfig`
entry, its `config_test.go` entry and its `logPlainConfigFields` entry —
`internal/config` did not compile until those were de-duplicated.

**Two defects the merge created, both fixed here.**

`ValidateAPIExposure` short-circuited on the shared `apiBindAddr`, so this
branch's `plugins.api.<name>.config.host` override defeated it in both
directions: a provider widened by its own `host` served unauthenticated
while `apiBindAddr` was loopback, and a provider pinned to loopback by its
own `host` was required to authenticate while `apiBindAddr` was remote. It
now resolves through `APIListenHost`, the same precedence the provider
factories and the port-conflict check use. Two tests pin both directions
and each fails without the change.

The examples stack then stopped starting, because a container bind is a
non-loopback bind:

```
ERROR invalid configuration: plugins.api.blockfrost.config binds to
non-loopback address "0.0.0.0" without authentication: configure api.auth
or a provider auth policy
```

Compose now writes `DINGO_API_TOKEN` to `/run/secrets/dingo-api-token` in
both Dingo containers and sets `api.auth` to token mode against it; the file
is required because `api.auth.token` carries `ignored:"true"` so an inline
secret cannot come from the environment. Both Vite dev servers attach the
token as `Authorization: Bearer` on proxied requests, so it stays
server-side. The two Helm values files pin an image predating the
requirement; their comments now say what a newer one needs, and the chart
cannot mount a token file yet (blinklabs-io/helm-charts#464).

`bin/TestEntrypointForwardsSignalsDuringMithrilBootstrap` and
`TestEntrypointSignalHandlingSurvivesBootstrapToServeHandoff` fail under
full-suite load on their 2s readiness waits and pass standalone. This branch
changes nothing under `bin/`, and both reproduce on unmodified
`origin/main`. Filed as #3930.

chrisguiney:
closing this as a false positive reported by @wolf31o2's audits

PR #3933, opened 2026-09-04
fix(database): warn only when a transaction loses declared outputs

### LINKED ISSUE #3932: database: zero-output transactions log at WARN for a legal shape
## Summary

`database/transaction.go` logs at WARN when a transaction produces no UTxOs.
That is a legal transaction shape, so the warning fires during ordinary sync
and reads as a defect.

## Evidence

Fourteen occurrences in one Preview replay:

```
level=WARN msg="transaction has no produced outputs" txHash=7ad44ff9eaa54b34 slot=32983570
level=WARN msg="transaction has no produced outputs" txHash=8bb334783a767861 slot=32985468
level=WARN msg="transaction has no produced outputs" txHash=bf001a735fdfca71 slot=33304517
```

Taking the first: Dingo records it as `valid = 1`, and the chain agrees it has
no outputs.

```
tx_size      189
inputs       1        value 3126383
outputs      0
certificates stake_registration
deposit      2000000
fee          1126383
```

`3126383 = 2000000 + 1126383`. The transaction spends its whole input on a
stake-registration deposit plus the fee and returns no change. Value is
conserved, nothing is lost, and Dingo handles it correctly.

## Why change it

The warning describes a legal shape with no remedy attached, so it cannot be
acted on. I spent time chasing it as a possible lost-UTxO bug — the WARN level
is what made it look like one, since an operator's first reading of "no
produced outputs" on a `valid = 1` transaction is that outputs went missing.

DEBUG matches what the line actually is: a note that this transaction
contributed nothing to the UTxO set.

If the intent is to flag something genuinely unexpected, the condition should
be narrowed to the case that is unexpected — a valid transaction with no
outputs *and* no deposit or fee accounting for the consumed value — rather
than every zero-output transaction.

chrisguiney:
## Why

`database.SetTransactionWithOpts` warned whenever `tx.Produced()` was empty.
Producing no UTxOs is a legal transaction shape, so the warning fired during
ordinary sync with no remedy attached, and "transaction has no produced
outputs" on a `valid = 1` transaction reads as lost UTxOs.

Fixes #3932.

## Evidence

Fourteen occurrences in one Preview replay. Taking the first
(`7ad44ff9eaa54b34`, slot 32983570): one input worth 3126383, zero outputs, a
stake-registration certificate, deposit 2000000, fee 1126383. `3126383 =
2000000 + 1126383` — the transaction spends its whole input on the deposit plus
the fee and returns no change. Value is conserved and Dingo stores it
correctly.

## What changed

Narrowed the condition rather than lowering the level, so the "outputs went
missing" signal survives instead of being demoted to DEBUG where nobody would
see it.

Value-conservation accounting was rejected as the narrowing rule: it needs every
consumed input resolved from the UTxO store in the ingest hot path, and it
restates a ledger rule (`UtxoValidateValueNotConservedUtxo`) inside the storage
layer. The storage layer's own invariant is cheaper and exact — `Produced()` is
`Outputs()` for a valid transaction and the collateral return for an invalid
one, so an empty produced set means outputs were dropped before storage only
when the matching declaration is non-empty.

Those are two different fields, so the warning has two branches and each names
its own:

| transaction | declaration lost | message | attribute |
| --- | --- | --- | --- |
| valid | `Outputs()` | `valid transaction produced no UTxOs despite declaring outputs` | `outputs` |
| invalid | `CollateralReturn()` | `invalid transaction produced no UTxOs despite declaring a collateral return` | `collateralReturnLovelace` |

Each accessor is read once into a local before the branch (`Outputs()`
allocates a fresh slice per call in every era, and one read of
`CollateralReturn()` leaves no window between the nil test and `.Amount()`),
and the invalid branch's message literal is split so the file stays stable
under `golines --max-len=80` as `origin/main`'s copy is.

No other code, test, or doc referenced the old message.

### Behaviour note

One signal is dropped that the summary above does not imply. A Byron
transaction with zero outputs is malformed per the Byron CDDL (`txouts` is
`[+ txout]`), but `ByronTransactionBody.UnmarshalCBOR` is a plain `cbor.Decode`
with no non-emptiness check and `ByronTransaction.IsValid()` is `return true`,
so that shape lands in the legal-zero-output case and is now silent where it
used to warn. Accepted: the warning was never a CDDL check, and a
body-hash-valid block carrying a malformed Byron body is a consensus problem
that a storage-layer "outputs went missing" line would mislabel.

## Tests

`TestSetTransactionZeroProducedOutputsLogging` in
`database/zero_produced_outputs_test.go` drives real `SetTransactionWithOpts`
calls against a staged database and asserts on captured `slog` JSON. The capture
buffer is reset after staging and immediately before the call under test, so a
warning from `BlockCreate` or `SetGapBlockTransaction` cannot decide a subtest.

The immutable testdata contains no zero-output transaction (21299 transactions
scanned, 0 with zero outputs), so each shape is a thin wrapper embedding a real
fixture transaction and overriding only `Outputs()`, `Produced()`, `IsValid()`,
or `CollateralReturn()`.

| sub-test | shape | expected |
| --- | --- | --- |
| zero-output transaction does not warn | valid, no outputs, no produced | no WARN |
| transaction with outputs does not warn | unmodified fixture | no WARN (control) |
| dropped outputs warn | valid, declares outputs, produces none | WARN naming `outputs` |
| invalid transaction without collateral return does not warn | invalid, no collateral return | no WARN |
| dropped collateral return warns | invalid, collateral return declared, produces none | WARN naming `collateralReturnLovelace` |
| invalid transaction keeping its collateral return does not warn | invalid, collateral return declared *and* produced at index `len(Outputs())` | no WARN |

The two warning sub-tests assert the message text and the attribute, and the
collateral case additionally asserts the *absence* of `"outputs":` and of
`despite declaring outputs`, so a warning that named the wrong field would fail.

The last sub-test is the true negative for the collateral branch: the real
Babbage+ shape, where a non-nil `CollateralReturn()` coexists with a correctly
stored UTxO. The block indexer only emits offsets for declared outputs, so the
collateral return's index `len(Outputs())` has none and the call would fail on
`missing UTxO offset` before reaching the log; `setTx` therefore takes an
optional `func(*BlockIngestionResult)` and the sub-test copies output 0's span
to that index. Nothing on this path decodes the offset, so a borrowed span is
sufficient and the assertion is on the log alone.

Discrimination, re-verified at the current head:

- with `database/transaction.go` reverted to `origin/main`, 4 of the original 5
  sub-tests fail — the two legal shapes on the WARN that should not be there,
  the two dropped-output shapes on the message that should be;
- with the earlier single-shared-message version of the file, the two warning
  sub-tests fail on the wording and attributes;
- with the guard widened to `len(produced) == 0 || !tx.IsValid()`, so the
  collateral branch fires on declaration rather than on loss, the new true
  negative is the *only* one of the six to fail — it is not redundant with the
  existing controls;
- the two controls (unmodified fixture, kept collateral return) pass against
  `origin/main` as they should, so the assertions are not simply matching
  everything.

## Validation

Re-run at the current head:

- `go test ./database/... -count=1` — all packages ok
- `go test ./database/ -run TestSetTransactionZeroProducedOutputsLogging` — 6/6
- `golangci-lint run ./...` in all three module dirs, plus `GOOS=windows
  golangci-lint run ./...` — 0 issues in each. (`run.tests: false`, so the new
  test file is out of scope by configuration, not by omission.)
- `go test ./internal/architecture` (import boundaries) — ok
- `make docs-parity` — ok
- `gofmt -s`, `gofumpt`, and `golines -w --ignore-generated --chain-split-dots
  --max-len=80 --reformat-tags` over copies of both changed files — no diff

Not run: `nilaway` and `modernize`. Both are installed but built against go1.26
and cannot analyse this go1.27.1 toolchain — `nilaway` emits only
`failed prerequisites:` lines and `modernize` only
`analysis skipped due to errors in package`, so neither analysed any code and
neither result can be read as clean. Neither is pinned by the Makefile or run
in CI. An earlier `nilaway` run on this branch, while it still worked, reported
0 findings in either changed file.

## Docs

No update needed. `DATABASE.md` and `ARCHITECTURE.md` document the produced-UTxO
write path and the Leios closure semantics, neither of which changes; the log
level and text of this line are not documented in either file, and neither keeps
an inventory of logged warnings.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction warnings when no UTxOs are produced.
  - Warnings now distinguish between dropped outputs and dropped collateral returns.
  - Diagnostic messages include the relevant output count or collateral-return amount, along with transaction details.

- **Tests**
  - Added coverage for valid, invalid, zero-output, and collateral-return transaction scenarios.

### DIFF
--- a/database/transaction.go
+++ b/database/transaction.go
@@ -173,6 +173,20 @@ func (d *Database) SetTransaction(
 	)
 }
 
+// producedOutputsDropped reports whether an empty Produced() set means the
+// transaction lost outputs it declared. Producing nothing is a legal shape:
+// a valid transaction can spend its whole input on deposits plus the fee and
+// return no change (issue #3932), and an invalid transaction without a
+// collateral return produces nothing either. Produced() is Outputs() for a
+// valid transaction and the collateral return for an invalid one, so an empty
+// set is unexpected only when the matching declaration is non-empty.
+func producedOutputsDropped(tx lcommon.Transaction) bool {
+	if tx.IsValid() {
+		return len(tx.Outputs()) > 0
+	}
+	return tx.CollateralReturn() != nil
+}
+
 // SetTransactionWithOpts is SetTransaction with control over UTxO ingest
 // behavior via opts. Leios endorser-block application on the Musashi/
 // Haskell-conformant path passes SkipConsumedInputRecovery so a transaction's
@@ -241,10 +255,12 @@ func (d *Database) SetTransactionWithOpts(
 	// transactions (collateral return at index len(Outputs()))
 	// UTxO offsets MUST be available - no fallback to full CBOR storage
 	produced := tx.Produced()
-	if len(produced) == 0 {
+	if len(produced) == 0 && producedOutputsDropped(tx) {
 		d.logger.Warn(
-			"transaction has no produced outputs",
+			"transaction produced no UTxOs despite declaring outputs",
 			"txHash", hex.EncodeToString(ledgerHashPrefix(txHash)),
+			"valid", tx.IsValid(),
+			"outputs", len(tx.Outputs()),
 			"slot", point.Slot,
 		)
 	}
--- a/database/transaction.go
+++ b/database/transaction.go
@@ -250,20 +250,26 @@ func (d *Database) SetTransactionWithOpts(
 	// Each branch reports its own declaration: the two are different fields, so
 	// one shared message would name the wrong one for half the warnings.
 	if len(produced) == 0 {
+		// Each accessor is read once into a local: every era's Outputs()
+		// allocates a fresh slice per call, and reading CollateralReturn()
+		// once closes the gap between the nil test and the dereference.
+		outputs := tx.Outputs()
+		collateralReturn := tx.CollateralReturn()
+		txHashHex := hex.EncodeToString(ledgerHashPrefix(txHash))
 		switch {
-		case tx.IsValid() && len(tx.Outputs()) > 0:
+		case tx.IsValid() && len(outputs) > 0:
 			d.logger.Warn(
 				"valid transaction produced no UTxOs despite declaring outputs",
-				"txHash", hex.EncodeToString(ledgerHashPrefix(txHash)),
-				"outputs", len(tx.Outputs()),
+				"txHash", txHashHex,
+				"outputs", len(outputs),
 				"slot", point.Slot,
 			)
-		case !tx.IsValid() && tx.CollateralReturn() != nil:
+		case !tx.IsValid() && collateralReturn != nil:
 			d.logger.Warn(
-				"invalid transaction produced no UTxOs despite declaring a collateral return",
-				"txHash", hex.EncodeToString(ledgerHashPrefix(txHash)),
-				"collateralReturnLovelace",
-				tx.CollateralReturn().Amount().String(),
+				"invalid transaction produced no UTxOs despite "+
+					"declaring a collateral return",
+				"txHash", txHashHex,
+				"collateralReturnLovelace", collateralReturn.Amount().String(),
 				"slot", point.Slot,
 			)
 		}
--- a/database/transaction.go
+++ b/database/transaction.go
@@ -173,20 +173,6 @@ func (d *Database) SetTransaction(
 	)
 }
 
-// producedOutputsDropped reports whether an empty Produced() set means the
-// transaction lost outputs it declared. Producing nothing is a legal shape:
-// a valid transaction can spend its whole input on deposits plus the fee and
-// return no change (issue #3932), and an invalid transaction without a
-// collateral return produces nothing either. Produced() is Outputs() for a
-// valid transaction and the collateral return for an invalid one, so an empty
-// set is unexpected only when the matching declaration is non-empty.
-func producedOutputsDropped(tx lcommon.Transaction) bool {
-	if tx.IsValid() {
-		return len(tx.Outputs()) > 0
-	}
-	return tx.CollateralReturn() != nil
-}
-
 // SetTransactionWithOpts is SetTransaction with control over UTxO ingest
 // behavior via opts. Leios endorser-block application on the Musashi/
 // Haskell-conformant path passes SkipConsumedInputRecovery so a transaction's
@@ -255,14 +241,32 @@ func (d *Database) SetTransactionWithOpts(
 	// transactions (collateral return at index len(Outputs()))
 	// UTxO offsets MUST be available - no fallback to full CBOR storage
 	produced := tx.Produced()
-	if len(produced) == 0 && producedOutputsDropped(tx) {
-		d.logger.Warn(
-			"transaction produced no UTxOs despite declaring outputs",
-			"txHash", hex.EncodeToString(ledgerHashPrefix(txHash)),
-			"valid", tx.IsValid(),
-			"outputs", len(tx.Outputs()),
-			"slot", point.Slot,
-		)
+	// Producing no UTxOs is a legal shape: a valid transaction can spend its
+	// whole input on deposits plus the fee and return no change (issue #3932),
+	// and an invalid transaction without a collateral return produces nothing
+	// either. Produced() is Outputs() for a valid transaction and the
+	// collateral return for an invalid one, so an empty set means outputs were
+	// dropped before storage only when the matching declaration is non-empty.
+	// Each branch reports its own declaration: the two are different fields, so
+	// one shared message would name the wrong one for half the warnings.
+	if len(produced) == 0 {
+		switch {
+		case tx.IsValid() && len(tx.Outputs()) > 0:
+			d.logger.Warn(
+				"valid transaction produced no UTxOs despite declaring outputs",
+				"txHash", hex.EncodeToString(ledgerHashPrefix(txHash)),
+				"outputs", len(tx.Outputs()),
+				"slot", point.Slot,
+			)
+		case !tx.IsValid() && tx.CollateralReturn() != nil:
+			d.logger.Warn(
+				"invalid transaction produced no UTxOs despite declaring a collateral return",
+				"txHash", hex.EncodeToString(ledgerHashPrefix(txHash)),
+				"collateralReturnLovelace",
+				tx.CollateralReturn().Amount().String(),
+				"slot", point.Slot,
+			)
+		}
 	}
 	for _, utxo := range produced {
 		txId := ledgerInputIDBytes(utxo.Id)
--- a/database/zero_produced_outputs_test.go
+++ b/database/zero_produced_outputs_test.go
@@ -21,9 +21,14 @@ import (
 	"testing"
 
 	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/shelley"
 	mockledger "github.com/blinklabs-io/ouroboros-mock/ledger"
 	"github.com/stretchr/testify/require"
 )
+
+// collateralReturnAddress is an arbitrary mainnet payment address; the
+// collateral-return fixtures only need a decodable one.
+const collateralReturnAddress = "addr1qytna5k2fq9ler0fuk45j7zfwv7t2zwhp777nvdjqqfr5tz8ztpwnk8zq5ngetcz5k5mckgkajnygtsra9aej2h3ek5seupmvd"
 
 // noOutputsTx is the legal zero-output shape from issue #3932: a valid
 // transaction that declares no outputs, so Produced() is empty too. On chain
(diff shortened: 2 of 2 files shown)


chrisguiney:
Closing the one gap in the validation above: `nilaway` was reported as unavailable, so `make lint` could not complete its last steps.

I ran it against this branch. It reports **0 findings in either file this PR touches**. Its only output for `./database/...` is two pre-existing findings in `database/models/leios_block_dijkstra_test.go`, which this PR does not modify.

For context: `nilaway` is neither pinned by the Makefile nor run in CI, so it is not a gating check — but the branch is clean under it either way.

database/transaction.go
```diff
@@ -241,10 +255,12 @@ func (d *Database) SetTransactionWithOpts(
...
 	// UTxO offsets MUST be available - no fallback to full CBOR storage
 	produced := tx.Produced()
-	if len(produced) == 0 {
+	if len(produced) == 0 && producedOutputsDropped(tx) {
 		d.logger.Warn(
-			"transaction has no produced outputs",
+			"transaction produced no UTxOs despite declaring outputs",
 			"txHash", hex.EncodeToString(ledgerHashPrefix(txHash)),
+			"valid", tx.IsValid(),
+			"outputs", len(tx.Outputs()),
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 0801f6af. The single call site became a two-branch switch, so each branch names the declaration it actually lost:

- valid: `valid transaction produced no UTxOs despite declaring outputs`, with `outputs=<len(tx.Outputs())>`
- invalid: `invalid transaction produced no UTxOs despite declaring a collateral return`, with `collateralReturnLovelace=<amount>`

The contradictory `outputs: 0` on the invalid branch is gone, along with the `valid` attribute that only existed to disambiguate the shared message. `producedOutputsDropped` was inlined, since the predicate and the message it justifies now have to stay together.

`TestSetTransactionZeroProducedOutputsLogging/dropped_collateral_return_warns` now asserts the message and the `collateralReturnLovelace` attribute, and asserts the absence of both `"outputs":` and `despite declaring outputs` — with the previous commit's shared-message version of transaction.go restored, that subtest fails, so the wording is covered rather than assumed.

database/zero_produced_outputs_test.go
```diff
@@ -0,0 +1,174 @@
...
+			nil,
+			BatchedTxIngestOpts{},
+		))
+	}
+
+	t.Run("zero-output transaction does not warn", func(t *testing.T) {
+		var logs bytes.Buffer
+		db := newStagedDB(t, &logs)
+		setTx(t, db, noOutputsTx{Transaction: candidate.consumerTx})
+		require.NotContains(t, logs.String(), `"level":"WARN"`)
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 0801f6af. The `setTx` helper now calls `logs.Reset()` after staging and immediately before `SetTransactionWithOpts`, and returns the buffer contents from that call alone; every subtest asserts on that return value instead of `logs.String()`. Block offsets are computed before the reset so `mustBlockOffsets` cannot contribute output either.

Applied to all five subtests, not only the two `NotContains` ones — the positive cases benefit as well, since a `Contains` that matched a staging line would pass for the wrong reason.

Staging previously left ~13 lines (badger open, nine schema migrations, node-settings gates) in scope of every assertion; that is now excluded.

Discrimination re-verified after the change: with `database/transaction.go` reverted to origin/main, the same 4 of 5 subtests fail and the unmodified-fixture control still passes both ways.

database/transaction.go
```diff
@@ -241,10 +255,12 @@ func (d *Database) SetTransactionWithOpts(
...
 			"slot", point.Slot,
 		)
 	}
```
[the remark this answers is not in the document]
chrisguiney: Same defect as the sibling thread on this file, fixed in 0801f6af: the call site is now a two-branch switch, so the invalid branch logs `invalid transaction produced no UTxOs despite declaring a collateral return` with `collateralReturnLovelace`, and no longer reports `outputs` at all. Covered by `TestSetTransactionZeroProducedOutputsLogging/dropped_collateral_return_warns`, which asserts the message and attribute and the absence of `"outputs":`.

database/transaction.go
```diff
@@ -241,12 +241,32 @@ func (d *Database) SetTransactionWithOpts(
...
-		)
+		switch {
+		case tx.IsValid() && len(tx.Outputs()) > 0:
+			d.logger.Warn(
+				"valid transaction produced no UTxOs despite declaring outputs",
+				"txHash", hex.EncodeToString(ledgerHashPrefix(txHash)),
+				"outputs", len(tx.Outputs()),
+				"slot", point.Slot,
+			)
+		case !tx.IsValid() && tx.CollateralReturn() != nil:
```
wolf31o2: `tx.CollateralReturn()` is evaluated twice — the nil test here and `.Amount()` below. A local (`if cr := tx.CollateralReturn(); cr != nil`) drops the second call and closes the nil-deref against any future implementation whose `Produced()` and `CollateralReturn()` disagree. Unreachable today (Babbage/Conway/Dijkstra `Produced()` uses the same `CollateralReturn() == nil` test, so a non-nil return means `len(produced) == 1` and this block is skipped), so it is hardening, not a fix. `tx.Outputs()` above is double-called the same way.

database/transaction.go
```diff
@@ -241,12 +241,32 @@ func (d *Database) SetTransactionWithOpts(
...
+			d.logger.Warn(
+				"invalid transaction produced no UTxOs despite declaring a collateral return",
```
wolf31o2: This 82-char message pushes the call past `golines --max-len=80` (`Makefile:77`), so the next `make format` re-splits the whole `Warn` one argument per line. `origin/main`'s `database/transaction.go` is golines-stable and this file is not; the new test file is fine. Not CI-enforced, but worth shortening the message or pre-applying golines.

database/zero_produced_outputs_test.go
```diff
@@ -0,0 +1,214 @@
+// Copyright 2026 Blink Labs Software
```
wolf31o2, line 1: The one shape not covered is the true negative for the second branch: an **invalid** transaction with a non-nil `CollateralReturn()` whose `Produced()` does return it — the real Babbage/Conway shape, and the only case where the new collateral condition could misfire. It needs a UTxO offset at index `len(Outputs())`, which is presumably why it was skipped; worth either writing it or saying in the file why it is omitted.

One behaviour change worth noting since it is not in the description: a Byron transaction with zero outputs is malformed per the Byron CDDL (`txouts` is `[+ txout]`) and gouroboros enforces no non-emptiness at decode, so that shape used to warn and is now silent. Low value — a body-hash-valid block carrying it is a consensus problem rather than a storage one — but it is the only signal the narrowing actually drops.

wolf31o2 (commented):
No blocking findings. The narrowing is sound: checked `Produced()`, `IsValid()` and `CollateralReturn()` in every gouroboros era (Byron→Dijkstra), and `len(Produced())==0` is equivalent to `(IsValid() && len(Outputs())==0) || (!IsValid() && CollateralReturn()==nil)` in all of them — so a dropped output always makes one of the two new cases true. The test discriminates: 4 of 5 subtests fail against `origin/main`.

Three non-blocking notes inline.

database/transaction.go
chrisguiney: Fixed in 69e1cbf1. Both accessors are read once now, before the switch:

```go
outputs := tx.Outputs()
collateralReturn := tx.CollateralReturn()
txHashHex := hex.EncodeToString(ledgerHashPrefix(txHash))
```

so the nil test and `.Amount()` observe the same value and there is no window
for a future `Produced()`/`CollateralReturn()` disagreement to become a nil
deref. `tx.Outputs()` is fixed by the same change — it was genuinely called
twice on the valid branch, in the case guard and in the `"outputs"` argument.
`txHashHex` was never a double call (only one case body runs) and is hoisted
only because both branches want it.

Worth recording that the `Outputs()` half is not purely cosmetic: every era
allocates a fresh slice per call — Byron, Shelley, Alonzo, Babbage and Conway
all `make(...)` and refill in `*TransactionBody.Outputs()`, and Dijkstra does
the same through `dijkstraTransactionOutputs`. Both reads sit inside the
`len(produced) == 0` guard, so the ordinary ingest path is unaffected either
way.

database/transaction.go
chrisguiney: Fixed in 69e1cbf1. The message literal is split rather than shortened:

```go
"invalid transaction produced no UTxOs despite "+
    "declaring a collateral return",
```

The runtime string is byte-identical, so the wording the sibling Cubic thread
asked for survives and the sub-test that asserts it is unchanged.

Verified rather than assumed: running
`golines -w --ignore-generated --chain-split-dots --max-len=80 --reformat-tags`
over a copy of both changed files now leaves them byte-identical, which is the
property `origin/main`'s `database/transaction.go` has. Before this commit
golines rewrote the invalid branch's `txHash` and `slot` pairs onto four lines;
the diff is empty now.

database/zero_produced_outputs_test.go
chrisguiney: Written rather than documented as absent — 69e1cbf1 adds
`invalid transaction keeping its collateral return does not warn`.

The fixture is the real Babbage+ shape: `IsValid() == false`, a non-nil
`CollateralReturn()`, and a `Produced()` returning that output at
`shelley.NewShelleyTransactionInput(t.Hash().String(), len(t.Outputs()))` —
the index gouroboros itself uses.

The offset was the obstacle you predicted. The block indexer only emits offsets
for declared outputs, so index `len(Outputs())` has none and
`SetTransactionWithOpts` returns `missing UTxO offset` before reaching the log.
`setTx` now takes an optional `func(*BlockIngestionResult)` and the sub-test
copies output 0's span to that index. Nothing on this path decodes the offset —
it is stored verbatim by `blob.SetUtxo` — so a borrowed span is sufficient and
the assertion is on the log alone. `batch_skip_test.go` mutates `UtxoOffsets`
the same way.

It discriminates, and uniquely. Widening the guard to
`len(produced) == 0 || !tx.IsValid()`, so the collateral branch fires on
declaration rather than on loss — exactly the misfire this case exists to
catch — makes it the only one of the six sub-tests to fail. The blunter
`len(produced) <= 1` mutation takes it and `transaction with outputs does not
warn` down together, since the fixture transaction has one output, so the
narrower mutation is the one that shows the case is not redundant.

The original five still discriminate 4-of-5 against `origin/main`. The new case
passes against `origin/main`, as it should: main only warns when
`len(produced) == 0`, which this shape is not.

On Byron: confirmed, and left as-is deliberately.
`ByronTransactionBody.UnmarshalCBOR` is a plain `cbor.Decode` into
`TxOutputs []ByronTransactionOutput` with no non-emptiness check, and
`ByronTransaction.IsValid()` is `return true`, so a zero-output Byron body now
lands in the legal-zero-output case and is silent where it used to warn. The old
warning was not a CDDL check and did not claim to be one; a body-hash-valid
block carrying a malformed Byron body is a consensus problem, and a
storage-layer warning would mislabel it as lost outputs. Adding the note to the
PR description so it is not only in this thread.

chrisguiney:
@coderabbitai full review

chrisguiney:
@coderabbitai full review

wolf31o2 (approved):

Issue #3795, opened 2026-09-02
koiosparity: departed-pool detection reads pool_stake_snapshot after retention prunes it, so every departure becomes a strict-mode halt


chrisguiney:
## Problem

`poolDepartedAtParamEpoch` (`internal/koiosparity/check.go:346`) decides whether
a missing `reward_pool_input` row is a benign departure or a real Dingo gap by
looking the pool up in the `mark` `pool_stake_snapshot` set at K+1. That table is
pruned to `currentEpoch - 3` by `Manager.cleanupOldSnapshots`
(`ledger/snapshot/rotation.go:1067`), so the lookup returns an empty set for any
epoch older than that window:

```go
if len(paramEpochPools) == 0 {
    return false
}
```

False keeps the strict classification, so every pool that legitimately left the
pool set at K+1 is reported as `dingo_db_missing` instead of `pool_departed`. In
strict mode that stops the node.

## Observed

Preview, from genesis, `--koios-parity-strict`. The observer was checking epoch
14 while the node had advanced to epoch 20:

```
level=ERROR msg="koiosparity observer: epoch validation failed"
  network=preview epoch=14 error="parity ERROR at epoch 14 (1 mismatch(es))" strict=true
level=ERROR msg="fatal koios parity validation failure, initiating shutdown"
```

```
epoch = 14
pool_bech32 = pool1nk3uj4fdd6d42tx26y537xaejd76u6xyrn0ql8sr4r9tullk84y
field = reward_pool_input_params
dingo_value =            koios_value = present
category = dingo_db_missing
```

The pool set matched exactly on both sides (44/44, no `only_dingo`/`only_koios`),
and `pool_stake_snapshot` held only epochs 17-20 at the time of the check.

## The finding is a false positive

That pool carries a retirement certificate for epoch 14 (slot 1207151), so it
left the pool set at the boundary into epoch 15 — the K+1 epoch for this check.
Its epoch-14 block count is stamped on a `reward_pool_input` row for epoch 15
that is correctly never written. Dingo's rows stop at epoch 14 and resume at 19
after the pool re-registers, which is what the ledger should do.

This is exactly the case #3485 introduced `pool_departed` for. The
classification is right; it just cannot fire, because the evidence it needs has
already been pruned by the time an observer running behind the node reaches the
epoch.

## Why the window is easy to fall outside

The retention window is three epochs. A Preview genesis replay closes an epoch
[...]

PR #3934, opened 2026-09-05
fix(koiosparity): prove pool departure from retirement certificates

### LINKED ISSUE #3925: koiosparity: a retiring pool is an ERROR when the observer runs behind the node
## Summary

A pool that retires is reported as `dingo_db_missing` — an ERROR category —
rather than `pool_departed`, whenever the observer runs behind the node. The
evidence the classification depends on has been pruned by then, and the
fallback that exists for that case cannot fire in practice.

In strict mode this halts the observer on an ordinary pool retirement.

## Evidence

A from-genesis Preview replay, observer roughly 45 epochs behind the node:

```
parity ERROR at epoch 243 (2 mismatch(es))
parity ERROR at epoch 244 (2 mismatch(es))
parity ERROR at epoch 246 (2 mismatch(es))
parity ERROR at epoch 247 (3 mismatch(es))
parity ERROR at epoch 248 (2 mismatch(es))

dingo_db_missing  reward_pool_input_params
  pool18ttg8kwe7w858sdk8rcw08cujytn6vjvpsc9t9c64n6h27wza2u  koios="present"
```

That pool registered at slot 20558938 and filed a retirement effective epoch
243. Its `reward_pool_input` rows stop after epoch 243, exactly as a departure
should. It is a departure, not missing input.

## Why the classification misses it

`ComparePoolEpoch` downgrades to `CategoryPoolDeparted` only when
`dingoPool.StakePresent && departedAtParamEpoch`, and `departedAtParamEpoch`
needs proven K+1 pool-set membership. Both routes to that proof fail here:

**pool_stake_snapshot** is pruned to `currentEpoch-3` by
`Manager.cleanupOldSnapshots`. For every epoch this observer checked, it is
empty:

```
paramEpoch=244  declared=123  reward_pool_input=121  pool_stake_snapshot=0
paramEpoch=245  declared=123  reward_pool_input=121  pool_stake_snapshot=0
paramEpoch=247  declared=128  reward_pool_input=126  pool_stake_snapshot=0
paramEpoch=248  declared=127  reward_pool_input=125  pool_stake_snapshot=0
```

**The reward-input fallback** (#3795) requires
`len(paramInputs) == declaredParamPools` exactly. `reward_pool_input` is
consistently two rows short of `epoch_summary.total_pool_count`, because Dingo
deliberately omits a degraded active pool from `reward_pool_input` while
keeping it in the pool set — which is precisely the case the exact-match
requirement is there to exclude. So the fallback never fires.

Both guards are individually correct and the strictness is deliberate
(#3485, #3795). Together they leave no route to the proof, so every retirement
observed from behind is an ERROR.

## Proposed fix

Prove departure per pool from certificate history rather than by
reconstructing the whole K+1 pool set. `pool_retirement` is retained for the
life of the database — this chain holds rows from slot 697222 to 24678832,
none pruned — and a retirement effective at or before the param epoch is
direct evidence that the pool left, needing no set-completeness argument.

**A retirement certificate alone is not sufficient.** A later registration
cancels a pending retirement, and a re-registration after one has taken effect
puts the pool back. This chain has 9 pools with a registration after a
retirement certificate, and Preview slot 17020890 (#3908) is exactly that
shape. The predicate has to be "the pool's latest certificate as of the param
epoch boundary is a retirement effective at or before it" — the same rule
`PoolCurrentState` implements — not "a retirement certificate exists".

That needs a new metadata-store query and accessors on both `DingoDB` and
`DatabaseSource`, so it is a cross-boundary change rather than a local edit to
the classifier.

## Scope

Six occurrences across five epochs in the run so far, and it will recur for
every retirement while the observer trails the node — which is the normal
condition during a replay. `pool_departed` is already informational, so the
correct classification costs nothing in coverage.

## Not fixed here

I diagnosed this but did not implement it; it is larger than the classifier
bug it looks like. Filing it with the diagnosis and the re-registration trap
already established.

chrisguiney:
Fixes #3925

## Why

A pool that retires is classified `dingo_db_missing` — an ERROR that halts
the observer in strict mode — instead of the informational `pool_departed`,
whenever the parity observer runs behind the node.

`ComparePoolEpoch` downgrades to `CategoryPoolDeparted` only when
`dingoPool.StakePresent && departedAtParamEpoch`. `StakePresent` is already
true in the failing case; `departedAtParamEpoch` is the missing half. Both
existing routes to that proof reconstruct the whole K+1 pool set and read
departure as absence from it, so each needs its own completeness argument —
and both were closed at once on a from-genesis Preview replay:

- `pool_stake_snapshot` is pruned to `currentEpoch-3` by
  `Manager.cleanupOldSnapshots`, so it is empty for every epoch a trailing
  observer checks.
- The #3795 reward-input fallback requires
  `len(paramInputs) == declaredParamPools` exactly, and `reward_pool_input`
  is consistently short of `epoch_summary.total_pool_count` because
  `buildRewardStateInputs` deliberately omits a degraded active pool from it
  — precisely the case that exact match exists to exclude. So it never fires.

## Evidence

Verified read-only against the replay's `metadata.sqlite`. The reported pool
`pool18ttg8kwe7w858sdk8rcw08cujytn6vjvpsc9t9c64n6h27wza2u`
(`3ad683d9…aacf575`) registered at slot 20558938 and filed one retirement
effective epoch 243 at slot 20856918. Its `reward_pool_input` rows stop after
epoch 243. `pool_stake_snapshot` has zero rows for epochs 243–249.

Running the shipped query at each param epoch the observer errored on, at
that epoch's real `epoch_summary.boundary_slot`:

```
paramEpoch=244 (boundary 21081600)  retired=55  subject_included=1
paramEpoch=245 (boundary 21168000)  retired=55  subject_included=1
paramEpoch=247 (boundary 21340800)  retired=57  subject_included=1
paramEpoch=248 (boundary 21427200)  retired=57  subject_included=1
```

The cancellation rule is load-bearing, not theoretical. At the epoch-244
boundary the naive predicate "a retirement certificate exists with
`epoch <= 244`" returns 59 pools against the correct 55; seven pools on this
chain have a registration filed after a retirement certificate.

## What changed

- `MetadataStore.GetPoolKeyHashesRetiredByEpoch` (one implementation, in
  `sqlstore/pool.go`): the "at or before" sibling of
  `GetPoolsRetiringAtEpoch`, with that query's latest-certificate resolution
  and cancellation rule preserved exactly — `ret.epoch <= ?` instead of
  `= ?`, returning bare key hashes because no deposit refund is applied.
  `= ?` would fix only the first epoch: the pool retired effective 243 and
  was still misclassified at 244, 245, 247 and 248.
- `DingoEpochData.BoundarySlot` carries `epoch_summary.boundary_slot`,
  populated in both `DingoDB.GetEpochData` (the column was already selected)
  and `DatabaseSource.GetEpochData`, so certificate history is resolved at
  the same point in the chain the epoch's pool set was captured at.
- `RewardParitySource.GetPoolsRetiredByEpoch`, implemented twice: `DingoDB`
  in raw SQL against a separately synced database, `DatabaseSource` through
  the store method.
- `poolDepartedAtParamEpoch` treats membership in that set as proof of
  departure alongside the existing pool-set route. It fails closed the same
  way: no boundary slot, or a read error, leaves the route shut and the
  stricter classification standing.

## Tests

Written first. Every claim below was re-verified by mutating the specific code
it names and confirming the named test fails — separately for each of the two
implementations, since a test naming one does not cover the other.

| Mutation | Where | Caught by |
|---|---|---|
| `ret.epoch <=` → `=` | sqlstore | `TestGetPoolKeyHashesRetiredByEpoch`, `TestGetPoolsRetiredByEpochImplementationsAgree` |
| `ret.epoch <=` → `=` | DingoDB | `TestCheckRetiredPoolStaysDepartedInLaterEpochs`, `TestGetPoolsRetiredByEpochImplementationsAgree` |
| drop `AND NOT ( ... )` cancellation guard | sqlstore | `TestGetPoolKeyHashesRetiredByEpoch`, `TestGetPoolsRetiredByEpochImplementationsAgree` |
| drop `AND NOT ( ... )` cancellation guard | DingoDB | `TestCheckReregisteredPoolStillErrors`, `TestGetPoolsRetiredByEpochImplementationsAgree` |
| retirement cut `rt.added_slot <` → `<=` | sqlstore | `TestGetPoolKeyHashesRetiredByEpoch`, `TestGetPoolsRetiredByEpochImplementationsAgree` |
| retirement cut `rt.added_slot <` → `<=` | DingoDB | `TestGetPoolsRetiredByEpochImplementationsAgree` |
| registration cut `pr.added_slot <` → `<=` | sqlstore | `TestGetPoolsRetiredByEpochImplementationsAgree` |
| registration cut `pr.added_slot <` → `<=` | DingoDB | `TestGetPoolsRetiredByEpochImplementationsAgree` |
| drop `BoundarySlot` from `GetEpochData` | DingoDB | both `TestCheckRetiredPool*` |
| drop `BoundarySlot` from `GetEpochData` | DatabaseSource | `TestDatabaseSourceGetEpochData` |
| `poolDepartedAtParamEpoch` ignores retirement evidence | check.go | both `TestCheckRetiredPool*` |

The tests themselves:

- `TestGetPoolKeyHashesRetiredByEpoch` (`database/pool_retired_test.go`) —
  nine pools with explicit certificate histories, each with a real
  transaction/certs row so `added_slot` ties break on `block_index` then
  `cert_index`. Covers same-slot ordering in both directions.
- `TestGetPoolsRetiredByEpochImplementationsAgree`
  (`internal/koiosparity/source_test.go`) — runs **both** implementations
  against one physical `metadata.sqlite` on the real migrated schema, and
  asserts each against an explicit expected set as well as against each
  other. See "Implementation drift" below.
- `TestCheckRetiredPoolIsDepartedWithoutAnyPoolSetEvidence` — the issue's
  exact shape: `TestCheckIncompleteRewardInputSetStillErrorsAfterPrune`'s
  fixture (mark rows pruned, reward-input set short of the declared count)
  plus the pool's retirement certificate. Was ERROR, now `pool_departed`.
- `TestCheckRetiredPoolStaysDepartedInLaterEpochs` — retirement effective
  several epochs earlier; fails under `= paramEpoch` semantics.
- `TestCheckReregisteredPoolStillErrors` — identical to the first except for
  a registration filed after the retirement. Must stay `dingo_db_missing`.
- `TestDatabaseSourceGetPoolsRetiredByEpoch` — the store path in isolation.

Four pre-existing fail-closed tests (`TestCheckDegradedActivePoolStillErrors`,
`TestCheckMissingRewardBundleStillErrors`,
`TestCheckIncompleteParamEpochPoolSetStillErrors`,
`TestCheckIncompleteRewardInputSetStillErrorsAfterPrune`) all still pass: the
departure fixture now stamps a boundary slot, so the new query runs for them
too and correctly finds no retirement.

## Implementation drift

`RewardParitySource.GetPoolsRetiredByEpoch` is implemented twice, and nothing
required the two to agree. The coverage was split exactly along that seam —
the end-to-end checks drive only `DingoDB`, and
`TestDatabaseSourceGetPoolsRetiredByEpoch` drives only the store — so either
copy could drift without a test naming the other noticing.

`TestGetPoolsRetiredByEpochImplementationsAgree` closes that: one physical
database, both implementations, asserted against an explicit expected set
*and* against each other. Expected-set equality matters on its own — two
identically-broken implementations would otherwise agree and pass.

Auditing the class rather than the reported location also turned up a real
gap: the registration-side `pr.added_slot < boundarySlot` cut had no pool for
which it was load-bearing, because every cancellation in the fixture was
decided by the ordering guard instead. A re-registration landing exactly on
the boundary slot now covers it, in both implementations.

## Validation

- `go test ./internal/koiosparity/... ./database/... -count=1` — all pass.
- `go test ./ledger/... -run 'PoolRetire|Poolreap|PoolReap'` — pass;
  `GetPoolsRetiringAtEpoch` and POOLREAP refunds are untouched.
- `make lint` — golangci-lint 0 issues across every module dir and under
  `GOOS=windows`. `nilaway` is clean on `./internal/koiosparity/...` and
  `./database/plugin/metadata/...`.
- `make docs-parity` — pass.
- `make sql-check` — pass, no drift. No `database/sql` query or `sqlc.yaml`
  change; the new query is hand-written alongside `GetPoolsRetiringAtEpoch`.

## Docs

- `DATABASE.md` — new `GetPoolKeyHashesRetiredByEpoch` section with the full
  SQL, why it differs from `GetPoolsRetiringAtEpoch`, and the retention
  argument.
- `ARCHITECTURE.md` — needed updating too, and is: the Koios Parity Tracker
  section documented both pool-set routes and the #3795 fallback in detail,
  so it now records the third route, why the first two can be closed
  simultaneously, and how this one fails closed.

## Conflict notes

Based on current `origin/main`, not pre-merged with anything. Four open
unreviewed PRs touch these files:

- **#3853** adds a `rewardsPending` 8th parameter to `CompareAccountEpoch`.
  Different function from the `ComparePoolEpoch` call this changes, but both
  live in `compare.go`/`check.go`; expect a textual conflict, not a semantic
  one.
- **#3916, #3920, #3924** also touch `check.go`/`compare.go`. This change's
  edits are confined to `poolDepartedAtParamEpoch`, the evidence-resolution
  block in `checkEpoch`, and one `ComparePoolEpoch` doc comment, so conflicts
  should resolve textually.

Whichever lands last rebases.

---
## Summary by cubic
A pool that retires was classified `dingo_db_missing` — halting the observer in strict mode — instead of `pool_departed` whenever the parity observer ran behind the node. The pool-set routes to departure evidence were both unavailable (pruned snapshots, short reward-input sets), so the fix resolves departure from pool certificate history instead, which is retained for the life of the database.

**What changed**
- Adds `MetadataStore.GetPoolKeyHashesRetiredByEpoch` and the `RewardParitySource` implementations, resolving departures as "retired at or before a param epoch" under the same cancellation rule as `GetPoolsRetiringAtEpoch`.
- Retirements recorded without a backing certificate now sort ahead of same-slot certificates and skip that ordering's cancellation, in both `GetPoolsRetiringAtEpoch` and the `DingoDB` query, so the store and `DingoDB` cannot disagree; a drift test pins both implementations to each other and an explicit expected set, and a new POOLREAP test pins the same-slot ordering on the refund path.
- Carries `epoch_summary.boundary_slot` through `DingoEpochData`; without one the route stays closed and `dingo_db_missing` stands.
- Addresses #3925; tests cover earlier-epoch retirements, re-registered pools, same-slot tie-breaks, and reconcile retirements.

<sup>Written for commit 45d93ebd8d08966637c76930aa7321f36996e9ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3934?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Pool departure checks now recognize effective retirement certificates, including retirements from earlier epochs.
  * Re-registrations correctly cancel prior retirements.
  * Departure validation uses the K+1 epoch boundary to provide more reliable pool status results.

* **Bug Fixes**
  * Prevents valid retired pools from being incorrectly reported as missing data during parity checks.

* **Tests**
  * Added coverage for retirement timing, certificate ordering, boundary visibility, and re-registration scenarios.

### DIFF
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -2535,6 +2535,75 @@ when that account is missing or inactive. Both writes are slot-keyed (the
 `account_reward_delta` journal and the boundary `network_state` row), so a
 rollback past the boundary reverts them and re-application is deterministic.
 
+### `GetPoolKeyHashesRetiredByEpoch`
+
+Key hashes of pools whose effective retirement had already taken effect by a
+given epoch. Same latest-certificate resolution and same cancellation rule as
+`GetPoolsRetiringAtEpoch` — a pool is included only when, as of the boundary
+slot, its latest certificate is a retirement and no later registration
+supersedes it — but the epoch comparison is `<=` rather than `=`, and no
+registration columns are selected because no deposit refund is being applied.
+
+The two queries answer different questions. `GetPoolsRetiringAtEpoch` asks
+which pools leave at one exact boundary, which is what POOLREAP needs. This one
+asks which pools had already left by an epoch, which is what the Koios parity
+checker needs when it reaches an epoch long after the node passed it. The
+distinction matters because a pool that retired effective epoch 243 is still
+departed at 244 and 245, and `= $epoch` matches none of those later epochs.
+
+`pool_registration` and `pool_retirement` are retained for the life of the
+database, while `pool_stake_snapshot` is pruned to `currentEpoch - 3` by
+`Manager.cleanupOldSnapshots`. That is why this query, and not a pool-set
+membership read, is the departure evidence available to an observer running
+behind the node. Backends differ only in identifier quoting (`"transaction"` on
+SQLite/Postgres, `` `transaction` `` on MySQL).
+
+```sql
+WITH latest_reg AS (
+  SELECT pr.pool_id, pr.added_slot,
+    COALESCE(t.block_index, 0) AS blk_idx,
+    COALESCE(c.cert_index, 0)  AS cert_idx,
+    ROW_NUMBER() OVER (
+      PARTITION BY pr.pool_id
+      ORDER BY pr.added_slot DESC, COALESCE(t.block_index, 0) DESC, COALESCE(c.cert_index, 0) DESC
+    ) AS rn
+  FROM pool_registration pr
+  LEFT JOIN certs c ON c.id = pr.certificate_id
+  LEFT JOIN "transaction" t ON t.id = c.transaction_id
+  WHERE pr.added_slot < $boundarySlot
+),
+latest_ret AS (
+  SELECT rt.pool_id, rt.added_slot, rt.epoch,
+    COALESCE(t.block_index, 0) AS blk_idx,
+    COALESCE(c.cert_index, 0)  AS cert_idx,
+    ROW_NUMBER() OVER (
+      PARTITION BY rt.pool_id
+      ORDER BY rt.added_slot DESC, COALESCE(t.block_index, 0) DESC, COALESCE(c.cert_index, 0) DESC
+    ) AS rn
+  FROM pool_retirement rt
+  LEFT JOIN certs c ON c.id = rt.certificate_id
+  LEFT JOIN "transaction" t ON t.id = c.transaction_id
+  WHERE rt.added_slot < $boundarySlot
+)
+SELECT p.pool_key_hash
+FROM pool p
+INNER JOIN latest_reg lr  ON lr.pool_id = p.id  AND lr.rn = 1
+INNER JOIN latest_ret lrt ON lrt.pool_id = p.id AND lrt.rn = 1
+WHERE lrt.epoch <= $epoch
+  AND NOT (
+    lrt.added_slot < lr.added_slot
+    OR (lrt.added_slot = lr.added_slot AND lrt.blk_idx < lr.blk_idx)
+    OR (lrt.added_slot = lr.added_slot AND lrt.blk_idx = lr.blk_idx AND lrt.cert_idx < lr.cert_idx)
+  );
+```
+
+The Koios parity checker reaches this through `RewardParitySource`, which has
+two implementations: `DatabaseSource` calls this store method directly, and the
+standalone-CLI `DingoDB` runs the same SQL against a separately synced metadata
+database. Both feed `poolDepartedAtParamEpoch`, which treats membership in this
+result as proof of departure and so downgrades a missing K+1 `reward_pool_input`
+row from `dingo_db_missing` to the informational `pool_departed`.
+
 ### `GetMIRCertsInSlotRange`
 
 MIR certificates for the epoch range `[startSlot, endSlot)`, applied at the epoch boundary as the Shelley INSTANT rule. Distribution certs (`other_pot = 0`) credit registered reward accounts and debit the source pot in `network_state`; pot-to-pot transfer certs (`other_pot > 0`) move that amount between treasury and reserves. Every cert in the range is aggregated before any is applied: distribution totals count only credentials with a registered, active account, transfers are folded into the available pot balances, and the boundary writes a single `network_state` row. If either pot cannot cover its total the boundary is a no-op rather than an error, so an over-budget cert cannot fail the epoch rollover on every retry. The `mir.id` value is retained by the processed effect as the per-MIR reward-credit discriminator so multiple MIR certs can credit the same account at one boundary without collapsing into one `account_reward_delta` row.
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -2474,16 +2474,33 @@ Pools whose effective retirement takes effect at a given epoch, with the reward
 account and deposit needed to refund their POOLREAP deposit at the epoch
 boundary. A pool is included when, as of the boundary slot, its latest
 retirement certificate names the target epoch and has not been cancelled by a
-later re-registration (same-slot disambiguation uses `block_index` then
-`cert_index`). Unlike `GetActivePoolKeyHashesAtSlot`, this query does not rank
-synthetic reconcile retirements (`certificate_id = 0`) first: those rows carry
-the catch-up tip as `epoch`/`added_slot`, so the `added_slot < $boundarySlot`
-and `epoch = $epoch` filters exclude them from boundary refund processing by
-design — a reconcile-retired pool gets no POOLREAP refund because its real
-retirement (or lack of one) was already settled in the imported snapshot's
-ledger state. The deposit and reward account come from the latest
-registration. Backends differ only in identifier quoting (`"transaction"` on
-SQLite/Postgres, `` `transaction` `` on MySQL).
+later re-registration (same-slot disambiguation ranks synthetic reconcile
+retirements first, then compares `block_index`, then `cert_index`). Like
+`GetActivePoolKeyHashesAtSlot`, this query ranks synthetic reconcile
+retirements (`certificate_id = 0`) ahead of certificate-backed rows at the same
+slot and exempts them from the cancellation clauses. Such a row has no
+`certs`/`transaction` join, so its `COALESCE(..., 0)` indices are the lowest
+possible and it would otherwise lose every same-slot tie-break to a
+certificate-backed row — the opposite of what the ledger state it encodes says,
+and the shape `ledgerstate`'s snapshot import writes routinely (`ImportPool`
+followed by `RetirePools` at one slot). All four latest-retirement resolutions
+in the tree — `GetActivePoolKeyHashesAtSlot`, this query,
+`GetPoolKeyHashesRetiredByEpoch` and `DingoDB.GetPoolsRetiredByEpoch` — now
+share one ordering, so the active pool set, the POOLREAP refund and both Koios
+parity routes cannot disagree about which retirement is a pool's latest.
+
+Ranking those rows does not admit them to boundary refund processing. A
+reconcile row carries the catch-up tip as `epoch`/`added_slot`: the boundary
+into that same epoch has already passed by the time the row is written, so
+`added_slot < $boundarySlot` is false for it, and every later boundary asks for
+a different `$epoch`. A reconcile-retired pool therefore still gets no POOLREAP
+refund, because its real retirement (or lack of one) was already settled in the
+imported snapshot's ledger state — that exclusion now rests on those two
+filters alone rather than on the row losing a tie-break, and
+`TestGetPoolsRetiringAtEpochSameSlotResolution` pins both halves. The deposit
+and reward account come from the latest registration. Backends differ only in
+identifier quoting (`"transaction"` on SQLite/Postgres, `` `transaction` `` on
+MySQL).
 
 ```sql
 WITH latest_reg AS (
@@ -2502,11 +2519,13 @@ WITH latest_reg AS (
 ),
 latest_ret AS (
   SELECT rt.pool_id, rt.added_slot, rt.epoch,
+    CASE WHEN rt.certificate_id = 0 THEN 1 ELSE 0 END AS synthetic_ret,
     COALESCE(t.block_index, 0) AS blk_idx,
     COALESCE(c.cert_index, 0)  AS cert_idx,
     ROW_NUMBER() OVER (
       PARTITION BY rt.pool_id
-      ORDER BY rt.added_slot DESC, COALESCE(t.block_index, 0) DESC, COALESCE(c.cert_index, 0) DESC
+      ORDER BY rt.added_slot DESC, CASE WHEN rt.certificate_id = 0 THEN 1 ELSE 0 END DESC,
+               COALESCE(t.block_index, 0) DESC, COALESCE(c.cert_index, 0) DESC
     ) AS rn
   FROM pool_retirement rt
   LEFT JOIN certs c ON c.id = rt.certificate_id
@@ -2521,8 +2540,9 @@ INNER JOIN latest_ret lrt ON lrt.pool_id = p.id AND lrt.rn = 1
 WHERE lrt.epoch = $epoch
   AND NOT (
     lrt.added_slot < lr.added_slot
-    OR (lrt.added_slot = lr.added_slot AND lrt.blk_idx < lr.blk_idx)
-    OR (lrt.added_slot = lr.added_slot AND lrt.blk_idx = lr.blk_idx AND lrt.cert_idx < lr.cert_idx)
+    OR (lrt.added_slot = lr.added_slot AND lrt.synthetic_ret = 0 AND lrt.blk_idx < lr.blk_idx)
+    OR (lrt.added_slot = lr.added_slot AND lrt.synthetic_ret = 0 AND lrt.blk_idx = lr.blk_idx
+        AND lrt.cert_idx < lr.cert_idx)
   );
 ```
 
@@ -2541,8 +2561,10 @@ Key hashes of pools whose effective retirement had already taken effect by a
 given epoch. Same latest-certificate resolution and same cancellation rule as
 `GetPoolsRetiringAtEpoch` — a pool is included only when, as of the boundary
 slot, its latest certificate is a retirement and no later registration
-supersedes it — but the epoch comparison is `<=` rather than `=`, and no
-registration columns are selected because no deposit refund is being applied.
+supersedes it, with the same synthetic-reconcile ranking and the same
+`synthetic_ret = 0` exemption on the cancellation clauses — but the epoch
+comparison is `<=` rather than `=`, and no registration columns are selected
+because no deposit refund is being applied.
 
 The two queries answer different questions. `GetPoolsRetiringAtEpoch` asks
 which pools leave at one exact boundary, which is what POOLREAP needs. This one
@@ -2574,11 +2596,13 @@ WITH latest_reg AS (
 ),
 latest_ret AS (
   SELECT rt.pool_id, rt.added_slot, rt.epoch,
+    CASE WHEN rt.certificate_id = 0 THEN 1 ELSE 0 END AS synthetic_ret,
     COALESCE(t.block_index, 0) AS blk_idx,
     COALESCE(c.cert_index, 0)  AS cert_idx,
     ROW_NUMBER() OVER (
       PARTITION BY rt.pool_id
-      ORDER BY rt.added_slot DESC, COALESCE(t.block_index, 0) DESC, COALESCE(c.cert_index, 0) DESC
+      ORDER BY rt.added_slot DESC, CASE WHEN rt.certificate_id = 0 THEN 1 ELSE 0 END DESC,
+               COALESCE(t.block_index, 0) DESC, COALESCE(c.cert_index, 0) DESC
     ) AS rn
   FROM pool_retirement rt
   LEFT JOIN certs c ON c.id = rt.certificate_id
@@ -2592,8 +2616,9 @@ INNER JOIN latest_ret lrt ON lrt.pool_id = p.id AND lrt.rn = 1
 WHERE lrt.epoch <= $epoch
   AND NOT (
     lrt.added_slot < lr.added_slot
-    OR (lrt.added_slot = lr.added_slot AND lrt.blk_idx < lr.blk_idx)
-    OR (lrt.added_slot = lr.added_slot AND lrt.blk_idx = lr.blk_idx AND lrt.cert_idx < lr.cert_idx)
+    OR (lrt.added_slot = lr.added_slot AND lrt.synthetic_ret = 0 AND lrt.blk_idx < lr.blk_idx)
+    OR (lrt.added_slot = lr.added_slot AND lrt.synthetic_ret = 0 AND lrt.blk_idx = lr.blk_idx
+        AND lrt.cert_idx < lr.cert_idx)
   );
 ```
 
--- a/database/plugin/metadata/sqlstore/pool.go
+++ b/database/plugin/metadata/sqlstore/pool.go
@@ -1851,11 +1851,13 @@ WITH latest_reg AS (
 ),
 latest_ret AS (
     SELECT rt.pool_id, rt.added_slot, rt.epoch,
+           CASE WHEN rt.certificate_id = 0 THEN 1 ELSE 0 END synthetic_ret,
            COALESCE(t.block_index, 0) block_index,
            COALESCE(c.cert_index, 0) cert_index,
            ROW_NUMBER() OVER (
                PARTITION BY rt.pool_id
                ORDER BY rt.added_slot DESC,
+                        CASE WHEN rt.certificate_id = 0 THEN 1 ELSE 0 END DESC,
                         COALESCE(t.block_index, 0) DESC,
                         COALESCE(c.cert_index, 0) DESC
            ) rn
@@ -1872,9 +1874,9 @@ JOIN latest_ret ret ON ret.pool_id = p.id AND ret.rn = 1
 WHERE ret.epoch = ?
   AND NOT (
       ret.added_slot < reg.added_slot
-      OR (ret.added_slot = reg.added_slot
+      OR (ret.added_slot = reg.added_slot AND ret.synthetic_ret = 0
           AND ret.block_index < reg.block_index)
-      OR (ret.added_slot = reg.added_slot
+      OR (ret.added_slot = reg.added_slot AND ret.synthetic_ret = 0
           AND ret.block_index = reg.block_index
           AND ret.cert_index < reg.cert_index)
   )`,
--- a/database/plugin/metadata/sqlstore/pool.go
+++ b/database/plugin/metadata/sqlstore/pool.go
@@ -1911,6 +1911,90 @@ WHERE ret.epoch = ?
 	return ret, rows.Err()
 }
 
+// GetPoolKeyHashesRetiredByEpoch is GetPoolsRetiringAtEpoch's "at or before"
+// sibling: same latest-certificate resolution and same cancellation rule, but
+// it matches every retirement effective up to and including epoch rather than
+// only the one landing on it, and returns bare key hashes because no deposit
+// refund is being applied. See MetadataStore's doc comment for why the parity
+// checker needs the wider comparison (dingo #3925).
+func (s *Store) GetPoolKeyHashesRetiredByEpoch(
+	epoch uint64,
+	boundarySlot uint64,
+	txn types.Txn,
+) ([][]byte, error) {
+	db, ctx, err := s.readDBFromTxn(txn)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"GetPoolKeyHashesRetiredByEpoch: resolve db: %w",
+			err,
+		)
+	}
+	rows, err := db.QueryContext(ctx, `
+WITH latest_reg AS (
+    SELECT pr.pool_id, pr.added_slot,
+           COALESCE(t.block_index, 0) block_index,
+           COALESCE(c.cert_index, 0) cert_index,
+           ROW_NUMBER() OVER (
+               PARTITION BY pr.pool_id
+               ORDER BY pr.added_slot DESC,
+                        COALESCE(t.block_index, 0) DESC,
+                        COALESCE(c.cert_index, 0) DESC
+           ) rn
+    FROM pool_registration pr
+    LEFT JOIN certs c ON c.id = pr.certificate_id
+    LEFT JOIN "transaction" t ON t.id = c.transaction_id
+    WHERE pr.added_slot < ?
+),
+latest_ret AS (
+    SELECT rt.pool_id, rt.added_slot, rt.epoch,
+           COALESCE(t.block_index, 0) block_index,
+           COALESCE(c.cert_index, 0) cert_index,
+           ROW_NUMBER() OVER (
+               PARTITION BY rt.pool_id
+               ORDER BY rt.added_slot DESC,
+                        COALESCE(t.block_index, 0) DESC,
+                        COALESCE(c.cert_index, 0) DESC
+           ) rn
+    FROM pool_retirement rt
+    LEFT JOIN certs c ON c.id = rt.certificate_id
+    LEFT JOIN "transaction" t ON t.id = c.transaction_id
+    WHERE rt.added_slot < ?
+)
+SELECT p.pool_key_hash
+FROM pool p
+JOIN latest_reg reg ON reg.pool_id = p.id AND reg.rn = 1
+JOIN latest_ret ret ON ret.pool_id = p.id AND ret.rn = 1
+WHERE ret.epoch <= ?
+  AND NOT (
+      ret.added_slot < reg.added_slot
+      OR (ret.added_slot = reg.added_slot
+          AND ret.block_index < reg.block_index)
+      OR (ret.added_slot = reg.added_slot
+          AND ret.block_index = reg.block_index
+          AND ret.cert_index < reg.cert_index)
+  )`,
+		boundarySlot,
+		boundarySlot,
+		epoch,
+	)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"GetPoolKeyHashesRetiredByEpoch: query pools: %w",
+			err,
+		)
+	}
+	defer rows.Close()
+	ret := [][]byte{}
+	for rows.Next() {
+		var poolKeyHash []byte
+		if err := rows.Scan(&poolKeyHash); err != nil {
+			return nil, err
+		}
+		ret = append(ret, poolKeyHash)
+	}
+	return ret, rows.Err()
+}
+
 func (s *Store) RestorePoolStateAtSlot(
 	slot uint64,
 	txn types.Txn,
--- a/database/plugin/metadata/sqlstore/pool.go
+++ b/database/plugin/metadata/sqlstore/pool.go
@@ -1919,6 +1919,19 @@ WHERE ret.epoch = ?
 // only the one landing on it, and returns bare key hashes because no deposit
 // refund is being applied. See MetadataStore's doc comment for why the parity
 // checker needs the wider comparison (dingo #3925).
+//
+// "Same resolution" includes the synthetic-retirement key every
+// latest-retirement query in the tree shares. A reconcile retirement
+// (certificate_id = 0) has no certs/transaction join, so its COALESCE'd
+// block_index/cert_index are both zero: without ranking it first and exempting
+// it from the same-slot cancellation clauses it would lose the tie-break to any
+// certificate-backed registration in its own slot, and the pool would be
+// reported as still active. ledgerstate's snapshot import writes exactly that
+// shape — ImportPool followed by RetirePools at one slot — so this is the
+// ordinary bootstrap case, not an edge case. DingoDB.GetPoolsRetiredByEpoch
+// carries the same three elements, and koiosparity's
+// implementations-agree test runs both against one database to pin them
+// against drift.
 func (s *Store) GetPoolKeyHashesRetiredByEpoch(
 	epoch uint64,
 	boundarySlot uint64,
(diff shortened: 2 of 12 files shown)


internal/koiosparity/check.go
```diff
@@ -614,6 +633,39 @@ func checkEpoch(
...
+	// pool_registration/pool_retirement are retained for the life of the
+	// database. Resolved once per epoch, and only against the K+1 summary's
+	// own boundary slot -- without one there is no point in the chain to
+	// resolve each pool's latest certificate as of, so the route stays
+	// closed and the stricter classification stands (dingo #3925).
+	var retiredByParamEpoch map[string]struct{}
+	if paramBoundarySlot > 0 {
+		retired, rErr := dingo.GetPoolsRetiredByEpoch(
+			ctx,
+			paramEpoch,
```
wolf31o2: `paramEpoch` is K+1, but the K+1 mark pool set is captured one slot *before* the K+1 boundary, so `ret.epoch <= paramEpoch` reports departure one epoch early.

`epochBoundarySnapshotSlot` returns `boundarySlot-1` (ledger/chainsync.go:6155), `Manager.captureMarkSnapshot` computes the mark distribution at `evt.SnapshotSlot` (ledger/snapshot/manager.go:755), and `GetActivePoolKeyHashesAtSlot` resolves that slot against the epoch containing it — K — keeping every pool with `retirement.epoch == K+1` (`OR lrt.epoch > ?`, sqlstore/pool.go:1107). That is SNAP-before-POOLREAP: a pool retiring effective K+1 is still in the K+1 pool set and still gets a K+1 `reward_pool_input` row. The PR's own replay data shows it — the pool retiring effective 243 has rows through 243, and the observer errors only from 244 on.

Because the certificate route is consulted before `paramEpochPools`, it also overrides complete pool-set evidence. Reproduced on this head: `TestCheckDegradedActivePoolStillErrors`' fixture (pool present in a complete K+1 mark set; asserts `dingo_db_missing` and a non-empty `ErrorEpochs`) plus a single retirement effective `koiosEpoch+1` flips the one mismatch to `pool_departed` and empties `ErrorEpochs`. That is the degraded-active-pool masking #3485 exists to prevent.

Passing `paramEpoch-1` here matches the K+1 pool set exactly, leaves the store method's inclusive `<=` correct for POOLREAP-style callers, and still covers every observed case (243 < 244). No test pins `retirement.epoch == paramEpoch`: both new end-to-end tests use retirements strictly earlier than K+1, and `TestGetPoolKeyHashesRetiredByEpoch`/`TestGetPoolsRetiredByEpochImplementationsAgree` assert the inclusive store semantics rather than the caller's choice of epoch.

wolf31o2 (changes requested):

wolf31o2:
Fixed in d66c3663. Tests: go test ./internal/koiosparity ./database/plugin/metadata/sqlstore.

database/plugin/metadata/sqlstore/pool.go
```diff
@@ -1911,6 +1913,90 @@ WHERE ret.epoch = ?
...
+                        COALESCE(t.block_index, 0) DESC,
+                        COALESCE(c.cert_index, 0) DESC
+           ) rn
+    FROM pool_registration pr
+    LEFT JOIN certs c ON c.id = pr.certificate_id
+    LEFT JOIN "transaction" t ON t.id = c.transaction_id
+    WHERE pr.added_slot < ?
+),
+latest_ret AS (
+    SELECT rt.pool_id, rt.added_slot, rt.epoch,
```
wolf31o2, line 1964: The K+1 off-by-one is fixed: `paramEpoch-1` matches the mark pool set exactly, a retirement effective `koiosEpoch+1` stays `dingo_db_missing`, and `koiosEpoch` is still `pool_departed`. What remains is new at this head.

`latest_ret` here has no `synthetic_ret` column, no synthetic component in `ORDER BY`, and no `synthetic_ret = 0` guard on the cancellation clauses. Its documented twin `GetPoolsRetiringAtEpoch` gained all three in d66c366 (lines 1854, 1860, 1877, 1879), `GetActivePoolKeyHashesAtSlot` has had them since 1082-1106, and `DingoDB.GetPoolsRetiredByEpoch` has them at dingo_db.go:432-460. This query is now the only latest-retirement resolution in the tree without them, so `DatabaseSource` (in-process observer) and `DingoDB` (standalone CLI) classify the same pool differently.

Reproduced against one metadata.sqlite at this head, `queryEpoch = 7`, `boundarySlot = 1000`, using `retiredParitySeeder` plus rows written the way `Store.RetirePools` writes them (`certificate_id = 0`):

- pool `0x71` — registrations at slots 100 and 200 (`cert_index` 1), one synthetic retirement at slot 200 effective epoch 5. Store: `ret.cert_index 0 < reg.cert_index 1` cancels it, absent. DingoDB: `synthetic_ret = 1` skips the guard, present.
- pool `0x72` — registration at slot 100, a real retirement cert at slot 200 (`cert_index` 3) effective epoch 9, and a synthetic retirement at slot 200 effective epoch 5. Store `ROW_NUMBER` picks the cert-backed row, `9 <= 7` is false, absent. DingoDB picks the synthetic row, `5 <= 7`, present.

Result: store `{}`, DingoDB `{0x71…, 0x72…}`. `TestGetPoolsRetiredByEpochImplementationsAgree` does not catch it because `retiredParitySeeder` gives every row a real `certs`/`transaction` row, so no fixture pool has `certificate_id = 0`. `RetirePools` is called from ledgerstate/import.go:2013 and ledgerstate/reconcile.go:292, so synthetic rows are present on any node bootstrapped from a ledger-state snapshot.

DingoDB is the side that matches `GetActivePoolKeyHashesAtSlot`, so the store query is the one to align. Seeding a `certificate_id = 0` retirement into the drift test would pin it.

database/plugin/metadata/sqlstore/pool.go
```diff
@@ -1851,11 +1851,13 @@ WITH latest_reg AS (
 ),
 latest_ret AS (
     SELECT rt.pool_id, rt.added_slot, rt.epoch,
+           CASE WHEN rt.certificate_id = 0 THEN 1 ELSE 0 END synthetic_ret,
```
wolf31o2, line 1854: This changes `GetPoolsRetiringAtEpoch`, the POOLREAP deposit-refund and treasury path (ledger/poolreap.go:57), and contradicts the rationale DATABASE.md:2478-2483 still gives for it: "Unlike `GetActivePoolKeyHashesAtSlot`, this query does not rank synthetic reconcile retirements (`certificate_id = 0`) first … by design." The SQL listing at DATABASE.md:2504-2527 also still shows the pre-change `ORDER BY` and `NOT (…)` clauses. Either the doc's design argument no longer holds and should be rewritten, or the POOLREAP query should be left alone and only the parity queries changed. No test in `database/...` covers a synthetic retirement sharing an `added_slot` with a registration through this query.

internal/koiosparity/check.go
```diff
@@ -614,6 +633,39 @@ func checkEpoch(
 			)
 		}
 	}
+	// Both pool-set routes above reconstruct the whole K+1 pool set and read
+	// departure as absence from it, which is why each needs its own
+	// completeness argument and why both can be closed off at once: the mark
+	// rows are pruned to currentEpoch-3, and reward_pool_input is short of
+	// the declared count whenever a degraded active pool was omitted from it.
+	// Certificate history answers the same question per pool instead. A
+	// retirement effective at or before K+1 that no later registration
```
wolf31o2: Prose is now one epoch off from the code. The call below passes `paramEpoch-1`, so the predicate is a retirement effective at or before K, not K+1. Same wording at check.go:347 ("a retirement effective at or before it", where "it" is the K+1 boundary) and ARCHITECTURE.md:6073. DATABASE.md:2541-2542 also states this query has the "[s]ame latest-certificate resolution and same cancellation rule as `GetPoolsRetiringAtEpoch`", which the synthetic-retirement gap above makes false.

wolf31o2 (changes requested):

database/plugin/metadata/sqlstore/pool.go
chrisguiney: Confirmed and fixed. I reproduced both of your pools at `queryEpoch = 7`, `boundarySlot = 1000` before changing anything and got exactly what you describe — store `{}`, DingoDB `{0x71…, 0x72…}`.

`GetPoolKeyHashesRetiredByEpoch` now carries all three elements: the `synthetic_ret` CASE column, the `CASE … DESC` component in `latest_ret`'s `ORDER BY`, and `AND ret.synthetic_ret = 0` on both same-slot cancellation clauses. It is now identical to `DingoDB.GetPoolsRetiredByEpoch` apart from the `transaction` quoting. DingoDB is the side that matches `GetActivePoolKeyHashesAtSlot`, so the store is the side that moved, as you said.

`TestGetPoolsRetiredByEpochImplementationsAgree` now seeds `certificate_id = 0` rows, one per case you gave: `reconcileSameSlot` (registrations at 100 and 200 cert_index 1, reconcile retirement at 200 effective 5) and `reconcileOverCertRet` (registration at 100, cert retirement at 200 cert_index 3 effective 9, reconcile retirement at 200 effective 5). Reverting the store query to this head fails it on `DatabaseSource/MetadataStore resolved the wrong departure set`; removing only the ORDER BY key or only the guard fails it too.

Proving that turned up something worse than the reported bug, so I widened the audit to the whole same-slot rule rather than the one column you named. `block_index` was load-bearing for **no** pool in either fixture — the koiosparity seeder hard-coded it to `0` — and `latest_ret`'s own `cert_index` ordering was likewise unexercised, so either query could have dropped either and stayed green. The seeder now takes a struct with an explicit `blockIndex`, and the fixture carries a registration and a retirement in different transactions of one block (both directions) plus two retirement certificates in one transaction with insertion order deliberately disagreeing with `cert_index` order.

The drift test is now mutation-proven clause by clause. Each of these, applied to **either** implementation, fails it:

| mutation | store | DingoDB |
| --- | --- | --- |
| `latest_ret` ORDER BY drops `cert_index` | fails | fails |
| `latest_ret` ORDER BY drops `synthetic_ret` | fails | fails |
| cancellation `block_index` comparison flipped | fails | fails |
| cancellation `cert_index` comparison flipped | fails | fails |
| cancellation `synthetic_ret` exemption removed | fails | fails |
| `ret.epoch <=` narrowed to `=` | fails | fails |
| registration visibility cut `<` relaxed to `<=` | fails | fails |
| retirement visibility cut `<` relaxed to `<=` | fails | fails |

Before this round, rows 1 and 3 passed on both sides.

database/plugin/metadata/sqlstore/pool.go
chrisguiney: Taken the first branch of your either/or: the ranking stays and the doc is rewritten, because leaving `GetPoolsRetiringAtEpoch` alone would have left it disagreeing with `GetActivePoolKeyHashesAtSlot` about which retirement is a pool's latest — the pool set could drop a pool at one epoch while the refund fired at another.

**The doc's design argument no longer holds as stated, but its conclusion does.** I worked the reachability through rather than asserting it. `RetirePools` stamps the catch-up tip as `epoch`/`added_slot` (import.go:2013 passes the import epoch and slot; reconcile.go:292 passes `tipEpoch`/`tipSlot`). `applyPoolRetirements(newEpoch, boundarySlot)` needs both `ret.epoch = newEpoch` and `ret.added_slot < boundarySlot`. The boundary into the row's own epoch has already passed by the time the row is written, so `added_slot < boundarySlot` is false for it; every later boundary asks for a different `$epoch`. A reconcile row therefore still cannot drive a refund — but by the epoch and slot filters alone, not by losing a tie-break. DATABASE.md:2478-2500 now says that, and 2504-2527's listing is regenerated with the `synthetic_ret` column, the ORDER BY component and the guarded `NOT (…)` clauses.

**On the missing test, you were understating it.** `GetPoolsRetiringAtEpoch` had no same-slot coverage of any kind — not just the synthetic case. I mutation-checked the whole `database/` and `ledger/` suites against this head first:

```
GetPoolsRetiringAtEpoch block_index comparison flipped  -> PASSED  (unpinned)
GetPoolsRetiringAtEpoch cert_index  comparison flipped  -> PASSED  (unpinned)
```

Both directions of the POOLREAP deposit-refund tie-break could be inverted with the tree green. `TestGetPoolsRetiringAtEpochSameSlotResolution` (`database/pool_retired_test.go`) now covers the class rather than the one case you named: a registration and a retirement in different transactions of one block (both directions), two certificates in one transaction (both directions), two retirement certificates in one transaction with different effective epochs, the reconcile-vs-registration case, the reconcile-vs-certificate case, and a reconcile row at the boundary slot pinning the exclusion the design argument now rests on. It is table-driven over epochs 5 and 9, so an inverted key moves pools between the two sets rather than merely dropping them.

Every clause is proven load-bearing:

| mutation to `GetPoolsRetiringAtEpoch` | result |
| --- | --- |
| `block_index` comparison flipped | fails |
| `cert_index` comparison flipped | fails |
| `latest_ret` ORDER BY drops `cert_index` | fails (both sub-cases) |
| `latest_ret` ORDER BY drops `synthetic_ret` | fails (both sub-cases) |
| cancellation `synthetic_ret` exemption removed | fails |
| retirement visibility cut `<` relaxed to `<=` | fails |

internal/koiosparity/check.go
chrisguiney: Correct on all three, and fixed.

`check.go:642` and `check.go:347` now say "at or before K", with the reason stated rather than left implicit: the K+1 mark pool set is captured at `boundarySlot - 1`, which falls in epoch K, so `GetActivePoolKeyHashesAtSlot` keeps every pool retiring effective K+1 and those pools still have a K+1 `reward_pool_input` row — reading them as departed would mask a genuinely missing one. ARCHITECTURE.md:6073 gets the same correction and the same reason.

DATABASE.md:2541-2542's "same latest-certificate resolution and same cancellation rule as `GetPoolsRetiringAtEpoch`" was false when you wrote that; it is true as of this head, since `GetPoolKeyHashesRetiredByEpoch` now carries the `synthetic_ret` column, ordering key and exemption. I made the claim explicit rather than leaving it implied — the sentence now names the ranking and the `synthetic_ret = 0` exemption as part of what "same" covers, so the next person changing one query can see what the other is promising.

internal/koiosparity/check.go
chrisguiney: Verified against the code rather than taken on faith, and it holds — the branch already carries the fix you noted, in `paramEpoch-1` at check.go:665.

Walking your chain on this head:

- `epochBoundarySnapshotSlot(boundarySlot)` returns `boundarySlot - 1` (ledger/chainsync.go:6155), wired in at `SnapshotSlot: epochBoundarySnapshotSlot(boundarySlot)` (chainsync.go:6190).
- `Manager.captureMarkSnapshot` computes the distribution at `evt.SnapshotSlot` (ledger/snapshot/manager.go:755).
- `GetActivePoolKeyHashesAtSlot` resolves that slot with `SELECT epoch_id … WHERE start_slot <= ? ORDER BY start_slot DESC LIMIT 1` (sqlstore/pool.go:1040-1046), so `boundarySlot - 1` resolves to K, and the predicate's `OR lrt.epoch > ?` (pool.go:1107) keeps every pool with `retirement.epoch == K+1`.

So a pool retiring effective K+1 is in the K+1 mark pool set and has a K+1 `reward_pool_input` row, and `ret.epoch <= paramEpoch` would have reported it departed — masking a genuine `dingo_db_missing`, which is the opposite of what this PR is for. `paramEpoch-1` matches the pool set exactly and leaves the store method's inclusive `<=` correct for POOLREAP-style callers, as you said.

The prose had not caught up with the code — check.go:347, check.go:642 and ARCHITECTURE.md:6073 still read "at or before K+1". That is fixed at this head, with the derivation above recorded next to the predicate so the two cannot drift apart again silently.

internal/koiosparity/dingo_db.go
```diff
@@ -375,6 +384,102 @@ func (d *DingoDB) GetPoolStakeSnapshotMembers(
...
+                        COALESCE(c.cert_index, 0) DESC
+           ) rn
+    FROM pool_registration pr
+    LEFT JOIN certs c ON c.id = pr.certificate_id
+    LEFT JOIN `+txTable+` t ON t.id = c.transaction_id
+    WHERE pr.added_slot < ?
+),
+latest_ret AS (
+    SELECT rt.pool_id, rt.added_slot, rt.epoch,
+           CASE WHEN rt.certificate_id = 0 THEN 1 ELSE 0 END synthetic_ret,
```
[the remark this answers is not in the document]
chrisguiney: Valid, and fixed at this head. This is the same defect @wolf31o2 reproduced against a real `metadata.sqlite`, and your reading of the mechanism is right: the synthetic row skips the guard in DingoDB and loses the tie-break in the store, so the standalone CLI and the in-process observer classified the same pool differently.

`GetPoolKeyHashesRetiredByEpoch` now has all three elements you listed — the `CASE WHEN rt.certificate_id = 0 …` column, the matching `ORDER BY` component in `latest_ret`, and `AND ret.synthetic_ret = 0` on both same-slot cancellation clauses — leaving it identical to DingoDB's copy apart from the `transaction` quoting.

`TestGetPoolsRetiredByEpochImplementationsAgree` now seeds `certificate_id = 0` rows, as you suggested: one whose reconcile retirement shares a slot with a certificate-backed registration, and one whose reconcile retirement shares a slot with a certificate-backed retirement naming a later effective epoch. Reverting the store query fails it on `DatabaseSource/MetadataStore resolved the wrong departure set`.

`GetPoolsRetiringAtEpoch` already had the three elements from `d66c366`, but nothing tested them; `TestGetPoolsRetiringAtEpochSameSlotResolution` in `database/pool_retired_test.go` now does, including the case where a reconcile row must *not* produce a POOLREAP refund.

internal/koiosparity/check_test.go
```diff
@@ -1582,3 +1583,196 @@ INSERT INTO reward_pool_input (
...
+	require.NoError(t, err)
+	defer db.Close() //nolint:errcheck
+
+	res, err := db.Exec(`INSERT INTO pool (pool_key_hash) VALUES (?)`, poolHash)
+	require.NoError(t, err)
+	poolID, err := res.LastInsertId()
+	require.NoError(t, err)
+	for _, slot := range regSlots {
+		_, err := db.Exec(`
+INSERT INTO pool_registration (pool_id, pool_key_hash, added_slot)
```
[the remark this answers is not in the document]
chrisguiney: The gap was real, and larger than described. I checked it by mutation rather than by reading, and the answer split in two.

**Your literal claim.** The `check_test` fixtures do leave `certificate_id` NULL, so `COALESCE(…, 0)` collapses every index to 0 there. But `NULL = 0` is NULL in SQL, so those rows fall to the `ELSE 0` arm and behave as ordinary certificate-backed rows — the NULL is not itself a hazard.

**What actually was untested.** `TestGetPoolsRetiredByEpochImplementationsAgree` (`internal/koiosparity/source_test.go`) is the right home for this, since it runs *both* implementations against one physical `metadata.sqlite` and compares each to an explicit expected set — a check_test addition would exercise DingoDB alone against a hand-written expectation. Its fixture did have both directions of the same-slot `cert_index` tie-break. But mutating DingoDB's copy showed two clauses were not pinned in either implementation:

```
DingoDB latest_ret ORDER BY drops cert_index          -> PASSED  (unpinned)
DingoDB cancellation cert_index comparison flipped    -> FAILED  (pinned)
DingoDB cancellation block_index comparison flipped   -> PASSED  (unpinned)
```

`block_index` was load-bearing for no pool at all — the seeder hard-coded it to `0` — and `latest_ret`'s own `cert_index` ordering was never decisive because no pool had two retirement certificates in one slot. The test's doc comment claimed "every clause of the predicate is load-bearing … neutralising any one of them in either implementation fails this test", which was false for those two.

The seeder now takes a struct with an explicit `blockIndex`, and the fixture adds a registration and a retirement in different transactions of one block (both directions) plus two retirement certificates in one transaction whose insertion order deliberately disagrees with `cert_index` order, so the ORDER BY key cannot be satisfied incidentally by scan order. All eight clauses now fail the test when neutralised in either implementation, and the doc comment is true as written.

The same audit found `GetPoolsRetiringAtEpoch` — the POOLREAP deposit-refund path — with no same-slot coverage whatsoever; that is covered now too, in `database/pool_retired_test.go`.

internal/koiosparity/dingo_db.go
```diff
@@ -375,6 +384,100 @@ func (d *DingoDB) GetPoolStakeSnapshotMembers(
...
+func (d *DingoDB) GetPoolsRetiredByEpoch(
+	ctx context.Context,
+	epoch uint64,
+	boundarySlot uint64,
+) (map[string]struct{}, error) {
+	txTable := quoteTransactionTable(d.dialect)
+	rows, err := d.query(
+		ctx,
+		`
+WITH latest_reg AS (
```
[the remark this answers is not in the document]
chrisguiney: The risk you name is real and it materialised on this very PR: `d66c366` added `synthetic_ret` to DingoDB's copy and not to the store's, and the two then classified the same pool differently on any node bootstrapped from a ledger-state snapshot. So this is not hypothetical drift.

You offered two remedies. I took the second and deliberately not the first, and I'd rather say why than leave it implied.

**Why not extract.** `DingoDB` opens the metadata database with plain `database/sql` and imports only `database/models` and `database/types` — it does not depend on `sqlstore` at all. That is the property that lets the standalone CLI point at a synced database without the node's store, migrations and plugin machinery. Sharing the SQL means either importing `sqlstore` into `internal/koiosparity` (giving that up) or adding a package whose only content is SQL text. And the two are not byte-identical: DingoDB splices `quoteTransactionTable(dialect)` into two joins, while the store relies on `translateMySQLReservedIdentifiers` centrally. A shared constant would need the reserved identifier templated into it — string-spliced SQL to prevent string-duplicated SQL, which I think is the worse hazard of the two.

**Why the test is enough here.** Your objection to it — that it mitigates drift but does not prevent it — is fair in general, but it turns on how much of the predicate the test actually discriminates. On the previous head it discriminated less than its own doc comment claimed, which is the real version of your concern. That is now closed: every clause of both copies is mutation-proven load-bearing, sixteen mutations in total (eight clauses × two implementations), including the `block_index` comparison and `latest_ret`'s `cert_index` ordering, neither of which was pinned before. A divergent edit to either copy now fails CI rather than shipping.

If a third copy of this SQL ever appears, the calculus changes and extraction becomes worth its cost — happy to file that as a follow-up if you'd rather have it tracked, but I don't think it should block this PR.

chrisguiney:
**Pushed `ebccd281`** — addresses the three findings from your 16:02 review, plus the class audit each one implied.

**1. `GetPoolKeyHashesRetiredByEpoch` aligned with DingoDB.** Your reproduction was exact; I reproduced both pools at `queryEpoch = 7`, `boundarySlot = 1000` before touching anything and got store `{}` / DingoDB `{0x71…, 0x72…}`. The store query now carries the `synthetic_ret` column, the ordering key and the `synthetic_ret = 0` exemption, making it identical to DingoDB's copy apart from `transaction` quoting. `TestGetPoolsRetiredByEpochImplementationsAgree` seeds `certificate_id = 0` rows for both of your cases.

**2. `GetPoolsRetiringAtEpoch`: kept the ranking, rewrote the doc.** Leaving POOLREAP alone would have left it disagreeing with `GetActivePoolKeyHashesAtSlot` about which retirement is a pool's latest. The doc's conclusion survives — a reconcile row still cannot drive a refund — but by the `epoch = $epoch` and `added_slot < $boundarySlot` filters alone, not by losing a tie-break, since the boundary into the row's own epoch has already passed when it is written. DATABASE.md:2478-2500 says that now, and both SQL listings are regenerated.

**3. Prose corrected.** check.go:347, check.go:642 and ARCHITECTURE.md:6073 now read "at or before K", with the `boundarySlot - 1` derivation recorded next to the predicate.

**The audit found more than the findings named.** `GetPoolsRetiringAtEpoch` — the POOLREAP deposit-refund path — had *no* same-slot coverage: both directions of its tie-break could be inverted with the whole `database/` and `ledger/` suites green. And in the koiosparity drift test, `block_index` was load-bearing for no pool (the seeder hard-coded it to `0`) and `latest_ret`'s own `cert_index` ordering was never decisive, so that test's "every clause is load-bearing" comment was false when written.

Both fixtures now carry a registration and a retirement in different transactions of one block, and two retirement certificates in one transaction. Verified by mutation: 8 clauses × 2 implementations all fail the drift test, and 6 clauses fail the new `TestGetPoolsRetiringAtEpochSameSlotResolution`.

**Validation on `ebccd281`**
- `go test ./database/... ./internal/koiosparity/... ./ledger/ -count=1` — pass
- `golangci-lint` 2.13.2 (go1.27.0) — **0 issues**, Linux and `GOOS=windows`
- `make docs-parity`, `make import-boundaries` — pass

**Not run**, rather than run and passed: `nilaway` and `modernize` locally — both installed builds are go1.26 and cannot analyse go1.27.1, so they emit `failed prerequisites:` / `analysis skipped due to errors in package` and exit 0 rather than producing a result. CI's `lint` job covers both. One combined package run also reported `FAIL internal/koiosparity` without my capturing which test; three subsequent runs of that package, isolated and combined, passed, and I could not reproduce it — recording it rather than calling the suite clean, since it matches the flake class in #3848.

On cubic's CTE-duplication finding I pushed back rather than extracting, with reasoning in the thread: `DingoDB` deliberately does not depend on `sqlstore`, and a shared constant would need the reserved `transaction` identifier templated into the SQL text. I think the mutation-proven drift test is the better guard here, but say the word if you'd rather have the extraction tracked as a follow-up.

DATABASE.md
```diff
@@ -2474,16 +2474,33 @@ Pools whose effective retirement takes effect at a given epoch, with the reward
...
+later re-registration (same-slot disambiguation ranks synthetic reconcile
+retirements first, then compares `block_index`, then `cert_index`). Like
+`GetActivePoolKeyHashesAtSlot`, this query ranks synthetic reconcile
+retirements (`certificate_id = 0`) ahead of certificate-backed rows at the same
+slot and exempts them from the cancellation clauses. Such a row has no
+`certs`/`transaction` join, so its `COALESCE(..., 0)` indices are the lowest
+possible and it would otherwise lose every same-slot tie-break to a
+certificate-backed row — the opposite of what the ledger state it encodes says,
+and the shape `ledgerstate`'s snapshot import writes routinely (`ImportPool`
+followed by `RetirePools` at one slot). All four latest-retirement resolutions
```
wolf31o2, line 2486: `GetPoolsRetiringAtEpoch` gaining the synthetic ranking leaves two stale claims.

This sentence names four latest-retirement resolutions as exhaustive. Seven carry the ranking: `GetRetiringPools` (pool.go:800), `GetActivePoolKeyHashesOrdered` (pool.go:926), `GetActivePoolKeyHashesAtSlot` (pool.go:1082), `GetActivePoolRelays` (pool.go:1762), and the three named here. An eighth, `poolReapedAfterDelegation` (sqlstore/account.go:1176), does not, and its doc comment at account.go:1169 still states the rule as "later added_slot wins, then later block index, then later certificate index" and attributes it to `GetPoolsRetiringAtEpoch`.

Non-blocking: documentation only, no behavioral divergence. That query's same-slot branches are gated on `pr.added_slot < e.start_slot` / `rt2.added_slot < e.start_slot`, while a synthetic row's `added_slot` is the catch-up tip inside `rt.epoch` and is therefore `>= e.start_slot`, so the added_slot-equality branches are unreachable for it.

wolf31o2 (approved):

PR #3937, opened 2026-09-05
fix(bark): honor block identifiers and batch results in ArchiveService

### LINKED ISSUE #3442: bark: honor block identifiers and batch results in ArchiveService
The Bark `ArchiveService.FetchBlock` handler only uses `BlockRef.hash` and
requires it to be a 32-byte hex string. The protocol permits a block reference
identified by hash, slot, height, or a consistent combination of those fields.
The handler also returns an error for the first missing block instead of
returning successful results for the rest of the batch and adding the missing
reference to `FetchBlockResponse.not_found`.

Acceptance criteria:

- Resolve block references using each supported identifier and validate that
  supplied hash, slot, and height values agree.
- Return one `SignedUrl` for every found request and one `not_found` reference
  for every missing request without discarding other batch results.
- Preserve the requested identifier fields in each returned `SignedUrl.block`.
- Add single-reference, identifier-only, mixed found/missing batch, and
  inconsistent-reference tests.

chrisguiney:
`ArchiveService.FetchBlock` used only `BlockRef.hash` and required a 32-byte hex
string, while the protocol permits a reference identified by hash, slot, height,
or a consistent combination. It also returned an error on the first missing
block instead of returning the rest of the batch with the missing reference in
`FetchBlockResponse.not_found`.

Resolution precedence is `hash+slot` -> `hash` -> `slot` -> `height`.
`hash+slot` resolves with no index lookup because `(slot, hash)` is the blob
store's key; this is load-bearing, since `BlockByHash` is a hard miss for blocks
written before the `bh` index (#1915), and routing hash+slot through that index
would stop an archive node serving exactly the pre-index range it exists to
serve. Height is last: block number is not an indexed blob key, so it needs a
binary search over the block-ID space (new `database.BlockByNumber`).

Hash and slot agree by construction, since the point is built from the supplied
identifiers; height is compared against `metadata.Height` after `BlockURL`
returns. A well-formed reference whose identifiers name different blocks
describes no stored block and yields `not_found`. A reference with no identifier
or a non-hex hash is not a valid `BlockRef` and fails `InvalidArgument`
(previously `CodeUnknown`); parsing runs for the whole batch before `Acquire()`.

No generated protobuf changed — `bark@v0.1.0`'s `archive.proto` already declares
all three identifiers optional and `not_found`.

The in-repo consumer `bark/blob.go` now maps an archive `not_found` answer to
`types.ErrBlobKeyNotFound`, so a missing block reports identically whether it
came from local storage or the archive.

Batching a missing block into `not_found` would otherwise hide a storage
inconsistency, since `s3` and `gcs` return `types.ErrBlobKeyNotFound` both for
a block that was never written and for a block whose metadata object was lost.
`resolveBlockPoint` therefore reports whether resolving the reference actually
read the block; a hash-only, slot-only, or height-only reference that then
fails `GetBlockURL` as missing means the index and the blob store disagree, and
26beae50 fails the whole call for it. `hash+slot` resolves with no lookup, so
it carries no such evidence and stays an ordinary `not_found`, as does a
reference that misses at resolution.

`bark/blob.go` maps a `not_found` answer to `types.ErrBlobKeyNotFound` only
when the echoed reference is the `(slot, hash)` that was requested.
`verifyArchiveBlock` re-verifies a returned block against the request, so an
absence has to be re-verified too; otherwise an archive answering about a
different block could have this node record the requested one as missing.

## Tests

Identifier-only, mixed found/missing batch, inconsistent-reference, and
malformed-reference cases all fail on the unfixed handler and pass after. The
single-reference (hash+slot) case passed before the change and is kept as a
regression guard, not as new evidence.
`TestArchiveFetchBlockRejectsMissingMetadata` covers the index/blob-store
disagreement and both non-evidence cases.

`TestArchiveFetchBlockResolvesHeightBoundOncePerBatch` counts blob operations
through a wrapper around the test store, because the cost this PR originally
got wrong is invisible in wall clock on badger. It fails two ways without the
fix: 16 reverse index enumerations instead of 1 when the bound is resolved per
reference, and 70 whole-block reads instead of 16 when search probes go through
`blockByKey`. `TestGetBlock_RejectsArchiveNotFoundForDifferentBlock` fails on
all four of its cases without the `not_found` identity check.
`TestBlockByNumberResolvesCompactBlockMetadata` runs the search against
badger's compact `DBM1` metadata encoding and fails with a decode error without
the shared codec; it asserts the stored value really carries the magic, so it
cannot pass on a silent fallback to CBOR.

## Validation

`go test -race ./bark/...` and `./database/`; `golangci-lint` 0 issues on
`./bark/...` and `./database/...` (also `GOOS=windows`); `make
import-boundaries`, `make docs-parity`; conformance 23.6s.
`TestAddrClearsAfterStopTimesOut` in `bark` is a machine-timing flake unrelated
to this diff: it needs `Shutdown` to miss a 20ms deadline, and on an idle
machine it fails the same way on `main` (3 of 6 runs at 429dbdbb, 3 of 6 at
26beae50, 4 of 6 at this branch). It passes in CI and under `-race`.

Not run: `make lint` in full (`nilaway`, `modernize`, `golines` absent from
PATH); `make govulncheck` (network); DevNet (archive RPC handler and one blob
read helper, outside its documented scope); `make sql-check` (no SQL change).

## Cost of a height-only batch

An earlier revision of this description claimed a worst-case all-height batch
cost "roughly 2,400 index seeks". That was wrong, as
https://github.com/blinklabs-io/dingo/pull/3937#discussion_r3939173342 measured:
it counted only the binary search and not the tip lookup each reference
performed before it. `BlockByNumber` resolved the highest indexed block itself,
and on `s3` and `gcs` a reverse iterator over the `bi` prefix is
`listKeysToFile`, a full listing of every block-index object in the bucket with
no early break. `ArchiveService` carries no operator auth interceptor, so 100
height-only refs in one anonymous request meant 100 full-bucket enumerations
(~1.2e7 objects each at mainnet size), plus ~2,400 whole-block object GETs,
since probes resolved through `blockByKey`.

Fixed in a3727370. The bound is now `database.BlockNumberBound`, resolved once
per batch by `ResolveBlockNumberBound` and passed to `BlockByNumberBounded`,
and only when the batch contains a height-only reference at all. Search probes
read the ordered index entry and the block's small `_metadata` object instead
of the block CBOR; only the matched block is read in full.

Measured with a counting blob wrapper, 16 height-only refs against a 16-block
chain:

| | before | after |
|---|---|---|
| reverse `bi` enumerations | 16 | 1 |
| forward `bi` seeks | 54 | 54 |
| whole-block reads | 70 | 16 |
| metadata reads | 0 | 55 |

The same refs sent as `hash+slot` do no index work either way. Extrapolated to
the reviewed shape, 100 height-only refs at mainnet size go from 100 full-bucket
enumerations and ~2,400 whole-block GETs to 1 enumeration, ~2,400 small metadata
GETs, and 100 whole-block GETs.

Class audit of the other resolution paths: `hash+slot` reads nothing; `hash`
alone is one `bh` index GET plus one block read; `slot` alone is a forward
iterator over the `bp<slot>` prefix, and both cloud plugins pass that prefix
into the LIST request (`aws/database.go:414`, `gcs/database.go:507`), so it is a
bounded range LIST rather than a full listing.

A prefix on the LIST is not on its own enough for a *seeking* forward iterator,
which is what the height binary search drives. `s3StreamIterator.reset` set only
`Prefix` and never `StartAfter`, so each `Seek` listed the whole `bi` prefix and
discarded keys below the seek client-side; gcs was already bounded server-side
with `query.StartOffset`. That moved the unauthenticated enumeration from the
reverse tip lookup into the seek rather than removing it. `reset` now sets
`StartAfter` to a strict prefix of the seek key — exclusive, so it cannot be the
seek key itself, and the client-side filter still drops the handful it admits.
Seeking the mid-point of 20,000 `bi` objects goes from 11 `ListObjectsV2`
requests returning 11,000 keys to 1 request returning 1,000
(`TestS3SeekBoundsListServerSide`).

The remaining per-reference cost is one whole-block GET for the block that is
served, which the `hash` and `slot` paths already paid before this PR.
`resolveBlockPoint` only needs the `(slot, hash)` point, so that read could be
avoided for all three paths by resolving to a point rather than a block; that is
a wider change to `BlockByHash`/`BlockBySlot`'s shape and is not attempted here.

Reading the `_metadata` object directly also meant the block-metadata codec had
to be shared. Badger writes a compact `DBM1` binary encoding instead of CBOR for
run mode `serve`/`leios` with storage mode `core`, so decoding that value as
CBOR would have failed every height lookup on those configurations (found by
cubic). The codec moved to `database/types` next to `BlockMetadata`, since which
encoding is at the key depends on the writing node's configuration rather than
on the plugin reading it.

Resolves #3442

https://claude.ai/code/session_01LEJSNPB6bA7CVW5EgXxNni

---
## Summary by cubic
`ArchiveService.FetchBlock` previously required a 32-byte hex hash and failed the whole batch at the first missing block. It now resolves each reference by hash, slot, height, or any consistent combination and reports missing ones under `not_found` while serving the rest. Resolves #3442.

**Block reference resolution**
- Precedence is hash+slot, then hash, then slot, then height; hash+slot needs no index read and still serves blocks written before the `bh` index (#1915).
- Height-only resolution binary-searches the block-ID space (`database.BlockByNumber`), with the upper bound resolved once per batch — resolving it per reference costs a full bucket enumeration on s3/gcs.
- A reference that names no stored block, or whose identifiers disagree, yields `not_found`; malformed references (no identifier, non-32-byte hash) fail the whole call with `InvalidArgument`.
- A lookup-resolved block the blob store then reports missing is a storage inconsistency and fails the call; hash+slot carries no such evidence and stays an ordinary miss.
- Answered refs echo supplied identifiers verbatim and fill in omitted ones.

**Cost, client, and codec**
- The s3 blob iterator now bounds a seeking forward listing server-side via `StartAfter` (gcs already did with `StartOffset`), so each height-search probe no longer lists the whole `bi` prefix.
- Probes read the ordered index entry and the small `_metadata` object instead of the block CBOR; a batch of hash+slot references still touches no index.
- `bark/blob.go` maps an archive `not_found` answer to `types.ErrBlobKeyNotFound` only when the echoed reference matches the requested slot and hash.
- Block metadata is decoded through a shared `types.UnmarshalBlockMetadata` that accepts both the CBOR and compact `DBM1` encodings and rejects oversized prev-hash lengths.

<sup>Written for commit 69cd505adeafb26313aaac22408622aad9ed710d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3937?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Archive block retrieval now supports references by hash, slot, height, or hash-and-slot.
  - Block lookups by block height are now supported, including efficient handling of sparse archives.
  - Missing blocks are reported individually without discarding successful results from the same request.
  - Archive responses provide missing identifiers when a lookup succeeds.

- **Performance**
  - Archive and object-storage lookups use more efficient bounded searches, reducing unnecessary data scans.

- **Documentation**
  - Updated architecture and database documentation to describe block reference resolution and height-based lookup behavior.

### DIFF
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -1549,12 +1549,29 @@ as "not reachable by hash" rather than "not present".
 number is not an indexed blob key — only slot (`bp`), hash (`bh`), and the
 internal sequential ID (`bi`) are — so the lookup binary-searches the ID space
 that block numbers increase with, bounded above by the highest indexed block,
-and probes with `BlockAtOrAfterIndex` so a gap left by a Mithril bootstrap or
-drain import does not end the search early. A number no block carries returns
-`models.ErrBlockNotFound`. It costs O(log n) blob reads where a slot or hash
-lookup costs one, so prefer either of those when the caller has one; bark's
-`ArchiveService.FetchBlock` uses it only for a reference that supplies height
-alone.
+and seeks to the next indexed entry at or after each probe so a gap left by a
+Mithril bootstrap or drain import does not end the search early. A number no
+block carries returns `models.ErrBlockNotFound`. It costs O(log n) blob reads
+where a slot or hash lookup costs one, so prefer either of those when the
+caller has one; bark's `ArchiveService.FetchBlock` uses it only for a reference
+that supplies height alone.
+
+The upper bound is a separate value, `BlockNumberBound`, resolved by
+`ResolveBlockNumberBound` and passed to `BlockByNumberBounded`. Resolving it
+reads the newest `bi` entry through a **reverse** iterator, and the `s3` and
+`gcs` plugins implement a reverse iterator as a full listing of every object
+under the prefix with no early break (`listKeysToFile`) — one resolution
+enumerates every block-index object in the bucket. A caller answering more than
+one block number must therefore resolve the bound once and reuse it;
+`BlockByNumber` resolves its own and is a single-lookup convenience. The zero
+`BlockNumberBound` is unresolved and matches nothing, so a forgotten resolution
+fails closed with `ErrBlockNotFound` rather than searching an empty ID space.
+
+Search probes read the ordered `bi` entry and the block's `_metadata` object,
+never the block CBOR: a probe needs only the block's ID and height to decide
+which way to move, and reading it through the block itself would download a
+whole block object from cloud storage per probe. Only the one matching block is
+read in full.
 
 `lifecycle.ResolveTargetByNumber` keeps its own tip-bounded search rather than
 calling this: a truncate target must not resolve past the persisted tip, while
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -1269,7 +1269,7 @@ flowchart LR
 | Logical key | Value | Used by |
 |---|---|---|
 | `bp` + big-endian slot `uint64` + block hash bytes | Raw block CBOR, or expired-history marker `DBT1` | `BlobStore.SetBlock`, `GetBlock`, `TombstoneBlock`, block iterators |
-| `bp..._metadata` | `types.BlockMetadata`: `id`, `type`, `height`, `prev_hash` encoded as CBOR; Badger can use compact `DBM1` binary metadata | `GetBlock` and archive-proxy/history-expiry paths |
+| `bp..._metadata` | `types.BlockMetadata`: `id`, `type`, `height`, `prev_hash` encoded as CBOR; Badger can use compact `DBM1` binary metadata (run mode `serve`/`leios` with storage mode `core`). Which encoding is present depends on the configuration of the node that wrote the block, so decode with `types.UnmarshalBlockMetadata` rather than as CBOR — code outside the plugins reads this key directly. | `GetBlock`, the block-number search, and archive-proxy/history-expiry paths |
 | `bi` + big-endian internal block ID `uint64` | The corresponding `bp...` block key | Block iteration and block-by-index lookup |
 | `bh` + block hash bytes | The corresponding `bp...` block key | Fast block-by-hash lookup |
 | `u` + tx hash bytes + big-endian output index `uint32` | UTxO CBOR or a 52-byte `DOFF` CBOR-offset reference into a block | UTxO resolution and history expiry |
@@ -1571,7 +1571,9 @@ Search probes read the ordered `bi` entry and the block's `_metadata` object,
 never the block CBOR: a probe needs only the block's ID and height to decide
 which way to move, and reading it through the block itself would download a
 whole block object from cloud storage per probe. Only the one matching block is
-read in full.
+read in full. Because that value is read directly rather than through
+`GetBlock`, it is decoded with `types.UnmarshalBlockMetadata`, which accepts
+both the CBOR and the compact `DBM1` encodings.
 
 `lifecycle.ResolveTargetByNumber` keeps its own tip-bounded search rather than
 calling this: a truncate target must not resolve past the persisted tip, while
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -1543,6 +1543,23 @@ on a database whose index has not been backfilled. Reserve the by-hash lookup
 for callers that genuinely have only a hash, and treat its `ErrBlockNotFound`
 as "not reachable by hash" rather than "not present".
 
+### Block Number Lookup
+
+`BlockByNumber` resolves a block from its chain block number (height). Block
+number is not an indexed blob key — only slot (`bp`), hash (`bh`), and the
+internal sequential ID (`bi`) are — so the lookup binary-searches the ID space
+that block numbers increase with, bounded above by the highest indexed block,
+and probes with `BlockAtOrAfterIndex` so a gap left by a Mithril bootstrap or
+drain import does not end the search early. A number no block carries returns
+`models.ErrBlockNotFound`. It costs O(log n) blob reads where a slot or hash
+lookup costs one, so prefer either of those when the caller has one; bark's
+`ArchiveService.FetchBlock` uses it only for a reference that supplies height
+alone.
+
+`lifecycle.ResolveTargetByNumber` keeps its own tip-bounded search rather than
+calling this: a truncate target must not resolve past the persisted tip, while
+a read should serve any block the blob store actually holds.
+
 ## SQL Examples Mirroring the Go API
 
 The examples below mirror common `metadata.MetadataStore` methods. Postgres examples use `decode($1, 'hex')`; MySQL equivalents use `UNHEX(?)`, `HEX(col)`, and `` `transaction` `` instead of `"transaction"`.
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -1571,9 +1571,10 @@ Search probes read the ordered `bi` entry and the block's `_metadata` object,
 never the block CBOR: a probe needs only the block's ID and height to decide
 which way to move, and reading it through the block itself would download a
 whole block object from cloud storage per probe. Only the one matching block is
-read in full. Because that value is read directly rather than through
-`GetBlock`, it is decoded with `types.UnmarshalBlockMetadata`, which accepts
-both the CBOR and the compact `DBM1` encodings.
+read in full. Because the `_metadata` object is read directly rather than
+through `GetBlock`, it is decoded with `types.UnmarshalBlockMetadata`, which
+accepts both the CBOR and the compact `DBM1` encodings. The block CBOR itself
+is unaffected: the matching block is still read through `GetBlock`.
 
 `lifecycle.ResolveTargetByNumber` keeps its own tip-bounded search rather than
 calling this: a truncate target must not resolve past the persisted tip, while
--- a/bark/archive.go
+++ b/bark/archive.go
@@ -196,42 +196,6 @@ func isBlockMissing(err error) bool {
 		errors.Is(err, types.ErrBlobKeyNotFound)
 }
 
-// logBlockUnservable records a block the archive resolved out of its own
-// index and then could not sign a URL for.
-//
-// The cloud blob plugins report a lost metadata object with the same
-// types.ErrBlobKeyNotFound they use for an absent block object -- gcs
-// GetBlock and GetBlockURL log the partial write and return that error
-// anyway, and s3 GetBlockURL reads metadata before it ever heads the block
-// object -- so the error alone cannot tell the two apart. What does tell
-// them apart is that a lookup path already read this block: the index says
-// the archive holds it, and the blob store now says it does not.
-//
-// The reference is still answered in not_found. There is no URL to hand the
-// client, FetchBlockResponse has no per-reference error field, and failing
-// the call would discard every other block in the batch -- the exact defect
-// #3442 asked to remove. The operator, who is the only party that can act on
-// it, gets the log line instead.
-func (a *archiveServiceHandler) logBlockUnservable(
-	point common.Point,
-	err error,
-) {
-	if a.bark == nil || a.bark.config.Logger == nil {
-		return
-	}
-	a.bark.config.Logger.Warn(
-		"block resolved from the index but the blob store reports it missing; treating as not_found",
-		"component",
-		"bark",
-		"slot",
-		point.Slot,
-		"hash",
-		hex.EncodeToString(point.Hash),
-		"error",
-		err,
-	)
-}
-
 // FetchBlock resolves each requested block reference to a signed URL. A
 // reference the archive does not hold is reported in
 // FetchBlockResponse.not_found and the rest of the batch is still served;
--- a/bark/archive.go
+++ b/bark/archive.go
@@ -25,9 +25,15 @@ import (
 	archive "github.com/blinklabs-io/bark/proto/v1alpha1/archive"
 	archiveconnect "github.com/blinklabs-io/bark/proto/v1alpha1/archive/archivev1alpha1connect"
 	"github.com/blinklabs-io/dingo/database"
+	"github.com/blinklabs-io/dingo/database/models"
+	"github.com/blinklabs-io/dingo/database/types"
 	"github.com/blinklabs-io/gouroboros/protocol/common"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
+
+// blockHashLength is the byte length of a Cardano block hash, which the
+// hex-encoded BlockRef.hash field has to decode to.
+const blockHashLength = 32
 
 var _ archiveconnect.ArchiveServiceHandler = &archiveServiceHandler{}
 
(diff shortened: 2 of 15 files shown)


bark/archive.go
```diff
@@ -58,45 +217,70 @@ func (a *archiveServiceHandler) FetchBlock(
...
+	for _, ref := range refs {
+		point, err := resolveBlockPoint(db, ref)
 		if err != nil {
-			return nil,
-				fmt.Errorf("failed decoding hash %q: %w", b.GetHash(), err)
-		}
-		if len(hash) != 32 {
-			return nil,
-				fmt.Errorf("hash length must be 32 bytes, got %d", len(hash))
+			if isBlockMissing(err) {
```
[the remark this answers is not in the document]
chrisguiney: Valid, and this PR is what made it invisible — before it, every `GetBlockURL` error failed the call.

The handler cannot tell the two apart from the error, though. `gcs` `GetBlockURL` logs the partial write at database/plugin/blob/gcs/database.go:1351 and returns `types.ErrBlobKeyNotFound` regardless; `s3` `GetBlockURL` reads the metadata object before it ever heads the block object (database/plugin/blob/aws/database.go:1432-1437, HeadObject at 1448), so one sentinel covers both "never written" and "metadata lost". `GetBlock` in both plugins has the same shape.

What does distinguish them is that a lookup path already read the block. 61d73c9f makes `resolveBlockPoint` report whether it did: a hash-only, slot-only, or height-only reference that then fails `GetBlockURL` as missing means the index and the blob store disagree, which is a storage inconsistency rather than an absent block, and it is now logged at warn with the block's slot, hash, and the underlying error.

The reference is still answered in `not_found` rather than failing the call. There is no URL to hand the client, `FetchBlockResponse` has no per-reference error field, and failing would discard the rest of the batch — the exact defect #3442 asked to remove. `hash+slot` resolves with no lookup, so it carries no evidence and stays an ordinary `not_found`; the test pins that limitation too.

Covered by `TestArchiveFetchBlockLogsUnservableResolvedBlock` in bark/archive_test.go, which fails without the change (log buffer empty at the first assertion).

database/block.go
```diff
@@ -336,6 +336,76 @@ func BlockBySlot(db *Database, slot uint64) (models.Block, error) {
...
+		return err
+	})
+	return ret, err
+}
+
+func BlockByNumberTxn(txn *Txn, number uint64) (models.Block, error) {
+	if txn == nil {
+		return models.Block{}, types.ErrNilTxn
+	}
+	highest, err := BlocksRecentTxn(txn, 1)
```
wolf31o2: `BlocksRecentTxn(txn, 1)` runs once per resolved reference. On s3 and gcs -- the only blob backends that can sign URLs, so the only deployments where `FetchBlock` is reachable -- a `Reverse` iterator over the `bi` prefix is `listKeysToFile`, which enumerates every block-index object in the bucket into a temp file with no early break (`database/plugin/blob/gcs/database.go:470` and `:1152`; `database/plugin/blob/aws/database.go:372` and `:1111`). `ArchiveService` is registered without the operator auth interceptor (`bark/bark.go:311`), so an unauthenticated `FetchBlock` carrying `DefaultMaxFetchBlockRefs` height-only refs performs 100 full-bucket enumerations.

Measured at 61d73c9f with a counting blob wrapper on a 64-block chain: 50 height-only refs produced 50 reverse `bi` listings, 300 forward `bi` seeks, and 300 `blockByKey` block reads; the same 50 refs sent as hash+slot produced zero of each. `BlockAtOrAfterIndex` resolves through `blockByKey`, which downloads the whole block CBOR just to read `block.Number`, so every probe is a full object GET on cloud storage.

At mainnet size that is 100 enumerations of ~1.2e7 objects, plus roughly 2,400 range LISTs and 2,400 whole-block GETs, for one anonymous request. The PR description's "roughly 2,400 index seeks" does not account for the per-reference tip enumeration. Resolve the highest block once per batch and pass the upper bound into the search.

bark/blob.go
```diff
@@ -387,6 +387,18 @@ func (b *BlobStoreBark) fetchBlockFromArchive(
 
 	blocks := resp.Msg.GetBlocks()
 	if len(blocks) != 1 {
+		// The archive reports a block it does not hold under not_found and
+		// still answers the rest of the batch, so an empty blocks list with
+		// a not_found entry is a missing block rather than a broken
+		// archive. Report it the way a local blob store reports one.
+		if len(blocks) == 0 && len(resp.Msg.GetNotFound()) > 0 {
```
wolf31o2: Non-blocking. This branch does not check that the echoed reference is the one that was requested: any non-empty `not_found` list becomes `types.ErrBlobKeyNotFound` for this slot. `verifyArchiveBlock` re-verifies a returned block against `slot` and `hash`, so a wrong block is rejected, but a wrong absence is accepted and reports the requested block as missing. Match `GetNotFound()[0]` against the requested slot and hash before mapping.

wolf31o2 (changes requested):

wolf31o2:
Fixed in 26beae50. Tests: go test ./bark.

database/block.go
chrisguiney: Confirmed and fixed in a3727370. Both premises hold at this head: `listKeysToFile` walks the paginator/object iterator to completion with no early break in either plugin (`gcs/database.go:1152`, `aws/database.go:1111`), and `NewIterator` routes every `Reverse` request through it (`gcs:470`, `aws:372`); `ArchiveService` is registered with `commonHandlerOptions` only, while `newOperatorAuthInterceptor` is attached to `DatabaseService` alone (`bark/bark.go:311`).

The bound is now `database.BlockNumberBound`, resolved once by `ResolveBlockNumberBound` and passed to `BlockByNumberBounded`. `FetchBlock` resolves it once per batch and only when the batch contains a height-only reference, so a hash+slot batch still touches no index. The zero value is unresolved and matches nothing, so a caller that forgets to resolve one fails closed with `ErrBlockNotFound` rather than silently searching an empty ID space.

On the `blockByKey` half: it can be avoided, and is. Block metadata lives at its own `_metadata` blob key in all three plugins, and a probe needs only the block's ID and height to decide which way to move, so probes now read the ordered index entry plus that metadata object. Only the matched block is still read in full.

Reproduced your measurement shape with a counting blob wrapper, 16 height-only refs against a 16-block chain:

| | before | after |
|---|---|---|
| reverse `bi` enumerations | 16 | 1 |
| forward `bi` seeks | 54 | 54 |
| whole-block reads | 70 | 16 |
| metadata reads | 0 | 55 |

`TestArchiveFetchBlockResolvesHeightBoundOncePerBatch` asserts those counts exactly, since a regression here costs no measurable time on badger. It fails both ways without the fix: 16 reverse enumerations with the bound resolved per reference, and 70 whole-block reads with probes going through `blockByKey`.

Class audit of the rest of the batch loop: `hash+slot` reads nothing, `hash` alone is one `bh` GET plus one block read, and `slot` alone is a forward iterator over the `bp<slot>` prefix — both plugins pass that prefix into the LIST request (`aws/database.go:414`, `gcs/database.go:507`), so it is a bounded range LIST, not a full listing. The reverse-iterator tip lookup was the only path that was O(references) in full enumerations.

One residual cost I did not change: the block that is actually served is still read in full by `resolveBlockPoint`, which needs only `(slot, hash)`. The `hash` and `slot` paths already paid that before this PR, and removing it means changing what `BlockByHash`/`BlockBySlot` return on this path; say the word if you want that in scope.

PR description corrected — the old "roughly 2,400 index seeks" line is replaced with the measured numbers and an explicit note that it had understated the cost.

bark/blob.go
chrisguiney: Valid, and fixed in 534f8b60 despite being marked non-blocking — it is small and in code this PR introduced.

`archiveReportedMissing` now matches the echoed reference against the requested slot and hash before mapping to `types.ErrBlobKeyNotFound`. This client always sends both identifiers and `FetchBlock` echoes back exactly what was supplied, so the answer for this request carries both; anything else falls through to the broken-archive error. The hash is compared with `strings.EqualFold`, since the field is hex text echoed verbatim and the handler deliberately preserves the case it was given.

`TestGetBlock_RejectsArchiveNotFoundForDifferentBlock` covers a different hash, a different slot, an omitted hash, and an omitted slot; all four map to `ErrBlobKeyNotFound` without the check. `TestGetBlock_AcceptsArchiveNotFoundWithDifferentHashCase` pins that an upper-cased echo of the requested hash is still accepted.

chrisguiney:
Three commits on top of 26beae50, in response to review.

**1d7bcedf — lint break from 26beae50.** That commit removed the only call site of `logBlockUnservable` but left the function defined, and `unused` failed the build (`.golangci.yml` sets `run.tests: false`, so a test-only reference would not have rescued it either). Removed the helper and the log buffer in the test that no longer had anything to observe; the test keeps both boundary cases it already covered. `lint` is green locally on `./bark/...` and `./database/...`, also under `GOOS=windows`.

**a3727370 — the `database/block.go` finding.** Both premises held; details and the measured before/after counts are on that thread. Short version: the tip is now resolved once per batch as `database.BlockNumberBound` rather than once per reference, and search probes read the block's `_metadata` object instead of downloading the whole block CBOR. 16 height-only refs on a 16-block chain went from 16 reverse `bi` enumerations and 70 whole-block reads to 1 and 16. `TestArchiveFetchBlockResolvesHeightBoundOncePerBatch` counts blob operations through a wrapper and fails both ways without the fix. The PR description's "roughly 2,400 index seeks" claim is corrected there.

**534f8b60 — the `bark/blob.go` finding.** Fixed despite being marked non-blocking; the echoed `not_found` reference is now matched against the requested slot and hash before mapping to `types.ErrBlobKeyNotFound`.

`ARCHITECTURE.md` and `DATABASE.md` are updated for both behavior changes, including 26beae50's — they still described the warn log it replaced.

---

@wolf31o2 — one question about 26beae50 that I did not want to answer by reverting it.

It makes a confirmed-indexed-but-unservable block `return nil, err`, which discards every other reference in the batch. Batch-discarding is the specific defect #3442 was filed to remove, so this is a deliberate narrowing of that fix rather than a detail.

It is defensible as written: `FetchBlock`'s own doc comment already says a storage failure fails the whole call, and `confirmed` is only true when a lookup actually read the block, so the branch fires on a genuine index/blob-store disagreement and not on an ordinary miss. The alternative reading is that a partial write on one archived block should not deny service for the other 99 references in an unauthenticated batch, and that the operator signal belongs in a log or a metric rather than in the client's response.

I have left your behavior in place and written the tests and docs to match it. Confirming that is the contract you want, or whether you would rather it be logged and answered as `not_found`?

chrisguiney:
@coderabbitai full review

database/block.go
```diff
@@ -336,6 +336,354 @@ func BlockBySlot(db *Database, slot uint64) (models.Block, error) {
...
+	}
+	raw, err := blob.Get(blobTxn, types.BlockBlobMetadataKey(blockKey))
+	if err != nil {
+		if errors.Is(err, types.ErrBlobKeyNotFound) {
+			return types.BlockMetadata{}, models.ErrBlockNotFound
+		}
+		return types.BlockMetadata{}, err
+	}
+	var metadata types.BlockMetadata
+	if _, err := cbor.Decode(raw, &metadata); err != nil {
```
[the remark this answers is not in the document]
chrisguiney: Valid, and a merge blocker on the badger path. Confirmed against this head: `useCompactBlockMetadata` (`database/plugin/blob/badger/defaults.go:33`) returns true for run mode `serve` or `leios` with storage mode `core`, `SetBlock` then writes the `DBM1` value through `marshalBlockMetadataInto`, and `blockMetadataByKey` was decoding it as CBOR — so `ResolveBlockNumberBound` and every height lookup would have failed with a decode error on exactly those configurations. My tests missed it because the default badger store writes CBOR.

Fixed in 462218ca, though not by making the read backend-aware. The encoding was never private to the badger plugin: the value sits at a shared `bp..._metadata` key that generic `database` code reads directly, and which of the two encodings is present depends on how the node that *wrote* the block was configured, not on which plugin is doing the reading. A backend-aware decoder would still be wrong for a store whose blocks were written under a different run mode.

So the codec moved to `database/types`, next to `BlockMetadata` and the tombstone marker it parallels: `types.BlockMetadataBinaryMagic`, `MarshalBlockMetadataInto`, and `UnmarshalBlockMetadata`, which dispatches on the magic and falls back to CBOR — exactly what badger's own `unmarshalBlockMetadata` already did. The badger helpers now delegate to it, so there is one codec rather than two, and its fuzz round-trip test still covers it.

`TestBlockByNumberResolvesCompactBlockMetadata` builds a database in that run mode and storage mode, asserts the stored value actually carries the `DBM1` magic first (so the test cannot pass on a silent fallback to CBOR), then resolves every block by number. It fails with a decode error without the shared codec.

`DATABASE.md`'s blob key table now records that the encoding depends on the writing node's configuration and that readers must use `types.UnmarshalBlockMetadata`, since the previous wording described the dual encoding as a badger detail.

chrisguiney:
@coderabbitai full review

database/types/types.go
```diff
@@ -348,6 +349,94 @@ type BlockMetadata struct {
...
+// UnmarshalBlockMetadata decodes a block metadata value in either
+// encoding, dispatching on BlockMetadataBinaryMagic and falling back to
+// CBOR. This is the only supported way to read the value at a
+// BlockBlobMetadataKey, since which encoding is there depends on the
+// writing node's configuration rather than on the plugin being read.
+func UnmarshalBlockMetadata(data []byte) (BlockMetadata, error) {
+	if len(data) >= BlockMetadataBinaryHeaderLen &&
+		bytes.Equal(data[:4], BlockMetadataBinaryMagic[:]) {
+		prevHashLen := binary.BigEndian.Uint32(data[28:32])
+		expectedLen := BlockMetadataBinaryHeaderLen + int(prevHashLen)
```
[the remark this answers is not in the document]
chrisguiney: Valid. Fixed in 4296d775: `UnmarshalBlockMetadata` now rejects a declared previous-hash length above `BlockMetadataPrevHashMaxLen` before computing the expected length or allocating the hash, matching the bound `MarshalBlockMetadataInto` already enforces on the write side.

The gap came over verbatim when the codec moved out of the badger plugin, and it matters more here than it did there: the decoder is now exported and reached from generic `database` code, not only from the plugin that wrote the value.

To be precise about the severity, since it changes what the fix is worth rather than whether to make it: the existing `len(data) != expectedLen` check already bounded the allocation, so this was not a memory-exhaustion vector — an oversized hash needed a value that was genuinely that long. What it did allow was a self-consistent corrupt value decoding into a `BlockMetadata` the encoder could not have produced, which then propagated into a `models.Block` as if it were real. Reporting the damage is the right behavior.

`TestUnmarshalBlockMetadataRejectsOversizedPrevHash` builds a value whose declared and actual lengths agree, so only the bound can reject it; it passes without the fix. `TestBlockMetadataCompactRoundTrip` pins the encoding at the shared boundary the plugin and the block-number search now both use.

DATABASE.md
```diff
@@ -1543,6 +1543,42 @@ on a database whose index has not been backfilled. Reserve the by-hash lookup
...
+`BlockByNumber` resolves its own and is a single-lookup convenience. The zero
+`BlockNumberBound` is unresolved and matches nothing, so a forgotten resolution
+fails closed with `ErrBlockNotFound` rather than searching an empty ID space.
+
+Search probes read the ordered `bi` entry and the block's `_metadata` object,
+never the block CBOR: a probe needs only the block's ID and height to decide
+which way to move, and reading it through the block itself would download a
+whole block object from cloud storage per probe. Only the one matching block is
+read in full. Because that value is read directly rather than through
+`GetBlock`, it is decoded with `types.UnmarshalBlockMetadata`, which accepts
```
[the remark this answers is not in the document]
chrisguiney: Valid. Fixed in 4296d775. The sentence now names the metadata object and says explicitly that the matching block itself is still read through `GetBlock`, so the decoder is not read as applying to block CBOR.

database/block.go
```diff
@@ -336,6 +336,359 @@ func BlockBySlot(db *Database, slot uint64) (models.Block, error) {
...
+	prefix := []byte(types.BlockBlobIndexKeyPrefix)
+	it := blob.NewIterator(
+		blobTxn,
+		types.BlobIteratorOptions{Prefix: prefix},
+	)
+	if it == nil {
+		return blockIndexEntry{}, errors.New("blob iterator is nil")
+	}
+	defer it.Close()
+	it.Seek(types.BlockBlobIndexKey(blockIndex))
```
wolf31o2, line 608: The per-reference reverse enumeration is fixed as asked: 50 height-only refs now produce 1 reverse `bi` listing instead of 50, and probes read the metadata object instead of block CBOR (50 whole-block reads, down from 300). The unauthenticated full enumeration is not gone on s3 though; it moved from the reverse iterator to this seek, and it is larger than what it replaced.

`s3StreamIterator.reset` (`database/plugin/blob/aws/database.go:402-425`) sets only `Prefix` on `ListObjectsV2`. It never sets `StartAfter`, and `Seek` is implemented by listing from the start of the prefix and discarding keys client-side until `key >= seek` (`database.go:468`, `database.go:481`). gcs does bound it server-side with `query.StartOffset` (`database/plugin/blob/gcs/database.go:509`), so this is s3-only. Each probe here opens a fresh iterator over the whole `bi` prefix and seeks into it, so each probe pays that scan.

Measured against a fake s3 endpoint holding 50,000 `bi` objects: one `Seek` to the mid-point cost 26 `ListObjectsV2` calls and 26,000 returned keys, with 0 requests carrying `start-after`.

Measured on badger, chain of 4096 blocks, 100 height-only refs in one `FetchBlock`: 1,084 forward `bi` seeks, summed seek offset 2,290,991 — about `N/2` keys per probe over about `log2(N)` probes per reference. At mainnet size (~1.2e7 `bi` objects) that is order 1e8 keys and 1e5 `ListObjectsV2` calls per height-only reference, so order 1e7 LIST calls at `DefaultMaxFetchBlockRefs` for one anonymous request. The 100 full enumerations this batch fix removed were order 1e6. One height-only reference is enough; only `number > bound.HighestNumber` short-circuits.

`blobOpCounts.forwardIndexScans` (`bark/archive_test.go:84`) documents forward scans as "prefix-bounded range LISTs". That holds for the `bp<slot>` prefix, which names one slot, but not for a seek into the full `bi` prefix on s3.

Give the s3 forward iterator a server-side start bound before an unauthenticated caller can drive it. `StartAfter` is exclusive, so it cannot take the seek key directly; a strict prefix of the seek key is a safe lower bound and returns at most a handful of extra keys.

wolf31o2 (changes requested):

database/block.go
chrisguiney: Confirmed and fixed in `69cd505a`, in the s3 iterator rather than in the caller.

Your reading of `s3StreamIterator.reset` holds: it set only `Prefix`, never
`StartAfter`, and `advance` dropped everything below `it.seek` client-side, so
each `Seek` paid a listing of the whole `bi` prefix. gcs was already bounded
with `query.StartOffset`, so this was s3-only as you said.

`reset` now sets `StartAfter` when seeking. Taking your note that it is
exclusive and cannot be the seek key: the bound is a strict prefix of the full
seek key, which is strictly below the seek key and below every key after it, and
admits at most the handful of keys sharing that prefix but sorting below the
seek -- `advance` still drops those, so the client-side filter stays as the
correctness boundary and the server-side bound is purely a cost bound.

Measured with a fake ListObjectsV2 endpoint over 20,000 `bi` objects, seeking
the mid-point:

| | before | after |
| --- | --- | --- |
| `ListObjectsV2` requests | 11 | 1 |
| keys returned | 11,000 | 1,000 |
| requests carrying `start-after` | 0 | 1 |

`TestS3SeekBoundsListServerSide` (`database/plugin/blob/aws/seek_bound_test.go`)
pins it and fails without the change with exactly the shape you described:
`seek to the mid-point issued 11 ListObjectsV2 request(s) and returned 11000
key(s), none carrying start-after`.

Two things I did not change, and want to flag rather than bury:

- The fix is in the plugin, so it bounds every seeking forward iterator on s3,
  not only the one bark drives. That is the level the defect lives at, but it
  does widen this PR past bark.
- `blobOpCounts.forwardIndexScans` in `bark/archive_test.go:84` still documents
  forward scans as "prefix-bounded range LISTs". That is now true for the `bi`
  prefix on s3 as well as for `bp<slot>`, so the comment is no longer wrong --
  but it was measuring on badger, where it could never observe this. The counter
  cannot distinguish a bounded from an unbounded LIST; the new s3 test is what
  covers that.

`go test -race` passes on `./database/plugin/blob/aws/` (both tag sets),
`./database/plugin/blob/gcs/`, `./bark/...` and `./database/`.
`golangci-lint` is clean on the affected trees with and without
`dingo_extra_plugins`.

chrisguiney:
The `go-test (Linux, race)` failure on `69cd505a` was not from this branch.

`TestWindowedRewindConvergesWhilePrimaryChainExtends` failed with `remove block
at index 136: block not found`. This branch touches `bark/`,
`database/plugin/blob/aws/` and `database/block.go`, none of which is on that
path, and the same test failed the same way on #3945, whose changes are confined
to `internal/test/conformance/`. A plain re-run of #3945's identical head then
passed, and the test passes 25/25 locally under `-race`.

Filed as #4008 with both occurrences, the re-run, and the local counts. Job
re-run here.

Worth noting for #3938: the failure mode is `rollbackLocked` dying in
`removeBlockByIndex` on an index whose block is absent from the blob store,
before it can lower the tip — the same shape cited there as evidence that the
"concurrent rollback lowered the tip" state is unreachable through the
commit-failure window. CI timing appears to reach it.

wolf31o2 (approved):
Reviewed the current head; no merge-blocking findings.

Issue #3935, opened 2026-09-05
ledger: Plutus evaluation exceeds the declared budget on a canonical block


chrisguiney:
## Summary

Dingo's Plutus evaluator computes a higher execution cost than the block
producer declared, on a transaction the chain accepted. The node rejects the
block and stops following the chain.

## Evidence

From-genesis Preview replay, wedged at slot 41098815 (epoch 475):

```
level=WARN msg="Plutus evaluation disagrees with block producer (rejecting transaction)"
  tx_hash=eab27325c569121613728db87a4d8333ce74cc857bbe4991875ceab8787d0213
  block_slot=41098839
  script_hash=87b8d92f92af4c5482452d4625e88b86a0b1289c02e6f23e63c1a2f7
  redeemer_tag=0 redeemer_index=0
  eval_error="script exceeded declared budget:
              used     (924964350 cpu, 3452119 mem),
              declared (908604883 cpu, 3389879 mem)"
```

The overshoot is small and almost identical in both dimensions:

```
cpu: +16359467  (+1.8004%)
mem: +62240     (+1.8360%)
```

That both dimensions are over by essentially the same proportion points at
executing more of the same work — a few extra reduction steps — rather than a
mispriced individual builtin, which would move one dimension far more than the
other.

## Scope

One transaction, one script, in this run. It is a new class: zero occurrences
of `Plutus evaluation disagrees` in the two preceding replays of this same
data directory, which covered epochs 197-475 between them.

The declared budget is the producer's, and the chain accepted the block, so
the reference evaluator stayed within it. Dingo does not.

## Why it wedges rather than just rejecting

The rejection is deterministic, so the recovery path retries it forever. That
amplifier is #3889 and it reproduced here — see the comment I am adding
there. Fixing #3889 would turn this into a clean stop rather than a spin, but
the divergence itself is the root cause and is what has to be fixed for the
replay to pass this block.

## Not yet established

I have not identified which builtin or reduction accounts for the difference,
nor whether the cause is in `plutigo`, the cost model resolved for this
protocol version, or the script context Dingo builds. The transaction CBOR is
in the node log next to the rejection (`tx_cbor_hex=84ab0082825820...`), so the
evaluation is reproducible offline against the same script and redeemer.

chrisguiney:
Narrowing this: **the cost models are not the cause.**

Preview epoch 475 is protocol version 8 (Babbage), and the chain publishes two
[...]

Issue #3930, opened 2026-09-04
test(bin): entrypoint readiness wait times out under full-suite load


chrisguiney:
## Problem

`TestEntrypointForwardsSignalsDuringMithrilBootstrap` (both subtests) and,
less often, `TestEntrypointSignalHandlingSurvivesBootstrapToServeHandoff` fail
under whole-repository test load and pass when `./bin/` runs on its own. The
failure is the readiness wait, not the signal handling the tests are about.

## Evidence

Baseline `80e6f286`, macOS (darwin/arm64), unmodified `origin/main`.

`go test ./... -count=1`:

```
--- FAIL: TestEntrypointForwardsSignalsDuringMithrilBootstrap (4.03s)
    --- FAIL: TestEntrypointForwardsSignalsDuringMithrilBootstrap/first_run_SIGTERM (2.02s)
        entrypoint_test.go:135:
            Error:      Condition never satisfied
            Messages:   Mithril bootstrap child did not become ready
    --- FAIL: TestEntrypointForwardsSignalsDuringMithrilBootstrap/resumed_bootstrap_SIGINT (2.02s)
        entrypoint_test.go:135:
            Error:      Condition never satisfied
            Messages:   Mithril bootstrap child did not become ready
FAIL	github.com/blinklabs-io/dingo/bin	7.355s
```

The same run under `-race -timeout 30m` fails identically; the other 80
packages pass.

`go test ./bin/ -count=1` in isolation passes on three consecutive runs, and
`-run TestEntrypointSignalHandlingSurvivesBootstrapToServeHandoff -v` passes in
0.15s.

This is not caused by any working-tree change: the failure reproduces on a
clean checkout, and a stashed/unstashed comparison of the same tree produced
the same result on both sides.

## Cause

`entrypoint_test.go:135` and `:169` wait on
`testutil.WaitForCondition(..., 2*time.Second, ...)` for a readiness file the
entrypoint's child process creates. The child is a real forked process, so its
startup competes with every other package `go test ./...` is running in
parallel. Two seconds is enough on an idle machine and not enough under load —
consistent with the observed 2.01-2.02s failures (the full budget elapsed) and
with the 0.15s success in isolation.

## Acceptance criteria

- The readiness wait tolerates a loaded machine without making a genuine
  hang wait appreciably longer to report — for example a budget scaled the way
  other process-spawning tests in this repository scale theirs, rather than a
  flat 2s.
- The failure message distinguishes "child never started" from "child started
[...]

Issue #3928, opened 2026-09-04
replay wedges on a transaction that is exactly maxTxSize


chrisguiney:
## Summary

Every era's `maxTxSize` rule measures a transaction by the raw length of its
reconstructed CBOR. The fee rules measure it one byte shorter, and the shorter
one is what the chain uses. A transaction sitting exactly on `maxTxSize` is
therefore rejected as one byte over.

This is a consensus failure, not a validation nicety: the node rejects a block
the chain accepted and stops following the chain.

## Evidence

A from-genesis Preview replay wedged at slot 27745494 (epoch 321, Conway):

```
ledger pipeline stuck: repeated restarts are not advancing the tip, so the
failure is deterministic and will not clear on its own; the node is no longer
following the chain
  error="process block batch: tx ddd3f6b502c45d7c9f60f75c79148b70c0bae6e238aafafa1d24e9b7ed228d85
         validation failure at slot 27745494:
         transaction size too large: size 16385, max 16384"
```

Fetching that transaction's CBOR from Koios:

```
first byte           : 0x84  (four-element array)
raw CBOR length      : 16385
length - 1           : 16384
maxTxSize            : 16384
```

The hash of the decoded bytes matches, so this is the transaction the chain
accepted, and it is exactly on the limit. That is why it took 321 epochs to
surface — it needs a transaction deliberately built to fill `maxTxSize`.

## Cause

An Alonzo-and-later transaction is stored in a block as four parallel arrays —
bodies, witness sets, `IsValid` flags, auxiliary data. The `IsValid` flag is
not part of the transaction the chain sized. Reconstructing a standalone
transaction re-attaches it as the third element of a four-element array, one
byte the original never had.

`common.TxSizeForFee` already accounts for this:

```go
if arrayLen == 4 {
    return fullSize - 1, nil
}
```

`UtxoValidateMaxTxSizeUtxo` does not — it uses `len(txBytes)` directly. Two
rules, two measures, one of them wrong.

## Scope

The rule is duplicated seven times. Allegra delegates to Shelley; the other
six each compute the size themselves, and none used the fee measure.

Babbage and Conway are additionally worse: they call `cbor.Encode(tx)`
unconditionally rather than reading the stored on-wire bytes, discarding the
original encoding before measuring it.

## Fix

blinklabs-io/gouroboros#2212 introduces `common.TxSize` as the one measure,
has `TxSizeForFee` delegate to it, and points every era's rule at it.

[...]

PR #3940, opened 2026-09-05
feat(koiosparity): compare per-epoch protocol parameters against Koios

### LINKED ISSUE #3931: koiosparity: protocol parameters are not compared against Koios
## Summary

The Koios parity checker does not compare per-epoch protocol parameters. A
wrong parameter value is wedge-class — it changes what the node accepts — and
nothing in the verification loop would catch it before a block rejection does.

## Coverage today

The comparison covers rewards, stake, blocks produced, pool parameters and
epoch totals:

```
account_reward_amount     delegated_stake        reward_ada_pots
account_reward_presence   delegator_count        reward_pool_input_params
account_reward_duplicate  fixed_cost             reward_pool_input_stake
blocks_produced           margin                 total_active_stake
epoch_summary             pool_presence          totals_fees
koios_totals              member_rewards         totals_reserves / totals_treasury
```

Protocol parameters are absent. Koios exposes them at `/epoch_params`, and
Dingo stores them per parameter change in `pparams` (CBOR, keyed by epoch and
era), so both sides are already available at the granularity the checker works
at.

## Why it matters

The maxTxSize replay wedge (#3928) came from a *rule* measuring a transaction
wrongly. A wrong stored `maxTxSize` produces the identical failure — a
canonical block rejected, the node stops following the chain — and the parity
run would report PASS right up to the rejection.

The execution-unit parameters are the sharper case: `maxTxExUnits`,
`maxBlockExUnits`, `executionCosts` and `collateralPercentage` gate phase-2
validation, and a divergence there is silent until a script transaction
fails.

## Evidence that the comparison is viable

I ran it manually. Decoding Dingo's `pparams` CBOR and comparing 15 fields
against `/epoch_params` for epochs 22, 107, 200, 300, 360 and 380:

```
maxTxSize 16384          maxBlockBodySize 90112
minFeeA 44               minFeeB 155381
keyDeposit 2000000       poolDeposit 500000000
maxValueSize 5000        collateralPercentage 150   maxCollateralInputs 3
priceMem 577/10000       priceStep 721/10000000
maxTxExMem 14000000      maxTxExSteps 10000000000
maxBlockExMem 62000000   maxBlockExSteps 40000000000 -> 20000000000
```

Every field agreed. Two details worth carrying into an implementation:

- Dingo stores prices as rationals and Koios publishes decimals. `577/10000`
  and `0.0577` are the same number; the comparison needs `rationalsEqual`, which
  `compare.go` already has, not string equality.
- `maxBlockExSteps` changes between epochs 22 and 107. Dingo tracked that
  update, so this is a real comparison and not a match on static genesis
  defaults. It also means the checker must resolve the *effective* parameters
  for an epoch — `pparams` holds one row per change, not one per epoch — rather
  than looking for an exact epoch row.

## Scope note

This is an enhancement, not a defect report: no divergence was found. Filing it
because the gap is in the tool whose job is to find exactly this class of
problem.

chrisguiney:
Fixes #3931

## Why this gap matters

The Koios parity checker compared rewards, stake, blocks produced, pool parameters and epoch totals, but not protocol parameters. A wrong stored protocol parameter changes what the node accepts, so it produces the same outcome as a wrong validation rule — the #3928 `maxTxSize` replay wedge — a canonical block rejected and the node off the chain. Every existing comparison would keep reporting PASS right up to the rejection.

The execution-unit parameters are the sharper case: `maxTxExUnits`, `maxBlockExUnits`, the execution prices and `collateralPercentage` gate phase-2 validation, where a divergence stays silent until a script transaction fails.

## What changed

- `KoiosClient.GetEpochParams` fetches `/epoch_params`, mirroring `GetTotals`' single-row contract (empty, multi-row, or wrong-epoch responses are errors).
- New `koios_epoch_params` cache table, written by `fetchEpoch` **before** `CommitEpochData`. That ordering keeps the freshness marker honest: `koios_epoch_info.fetched_at` only advances once the parameter row is durable, so a process killed between the two writes leaves the epoch looking unfetched and it is simply re-fetched.
- `RewardParitySource.GetProtocolParams`, implemented by both `DingoDB` (standalone SQL) and `DatabaseSource` (in-process observer), resolving the effective `pparams` row and decoding it into an era-independent `DingoProtocolParams`.
- `CompareEpochProtocolParams` in `compare.go`, wired into `checkEpoch` at step 1d.
- `/epoch_params` classified field-by-field in `koiosCoverageMatrix`, so a report never implies an uncompared field was checked.

### Resolving the Dingo side

Two properties of `pparams` shape the query, and getting either wrong silently reads back a different parameter set:

- It holds **one row per parameter change**, not one per epoch — preview carries ~12 rows across 400+ epochs — so the applicable row is the latest at `epoch <= K`. An exact-epoch lookup finds nothing for almost every epoch.
- At an era boundary the rollover path writes **both** an old-era row (post-pparams-update) and a new-era row (`transitionToEra`) at the same epoch, with different CBOR shapes. Preview really does carry two epoch-2 rows and two epoch-3 rows. The `epoch` table decides which era is in force, so the era is resolved first and used both to filter the row and pick the decoder — the same order `api/blockfrost`'s adapter uses. Choosing by insertion order picks the Babbage row for preview epoch 2 and then fails to decode it as Alonzo.

### Fields compared

25 scalar fields plus both Plutus cost models: `era`, `min_fee_a`, `min_fee_b`, `max_block_size`, `max_tx_size`, `max_bh_size`, `key_deposit`, `pool_deposit`, `max_epoch`, `optimal_pool_count`, `influence`, `monetary_expand_rate`, `treasury_growth_rate`, `protocol_major`, `protocol_minor`, `min_pool_cost`, `price_mem`, `price_step`, `max_tx_ex_mem`, `max_tx_ex_steps`, `max_block_ex_mem`, `max_block_ex_steps`, `max_val_size`, `collateral_percent`, `max_collateral_inputs`, `cost_models`.

Every numeric field goes through the existing `rationalsEqual`, not string equality: Dingo stores exact rationals (`577/10000`, `721/10000000`) where Koios publishes decimals and exponent forms (`0.0577`, `7.21e-05`). `big.Rat` parses all of these, so plain integers compare correctly through the same path and no field needs a second code path. Koios numbers are kept as `json.Number`, so nothing round-trips through a float.

Cost models are compared entry for entry per language. Dingo's numeric keys (`0`, `1`) map to Koios's `PlutusV1`/`PlutusV2`. Findings name the language and the first differing entry with a differing-entry count (`entry 42 = 197145 (1 of 166 entries differ)`) rather than dumping a 166-integer array into a mismatch row.

### Failure classification

Deliberately three distinct outcomes, not one:

- **A differing value is a real divergence** → `value_mismatch` (FAIL). Never softened by the grace window: a parameter is stored when its change is applied, not computed late, so "not ready yet" does not apply to it.
- **An unresolvable Dingo row means nothing was compared** → `dingo_db_missing` (ERROR), softened to `reference_lag` inside the grace window after epoch close, exactly as `CompareEpochAggregates` treats an absent `epoch_summary`. It must not read as FAIL (no divergence was demonstrated) and must not read as PASS.
- **A failed read** → `dingo_db_error` (ERROR), kept distinct from an absent row so an operator can tell a broken query from missing data.

A missing Koios row is `koios_epoch_params`/`dingo_db_missing` rather than skipped, mirroring `CompareEpochTotals`' `koios_totals` handling: a cache populated before this comparison existed surfaces as incomplete reference data instead of a PASS that validated nothing.

Presence disagreement is a mismatch, not a skip. `""` on either side means "this era does not define this parameter", so both empty is agreement, but exactly one empty is a disagreement about the shape of the ledger state — which is what an era-gating bug looks like.

## Tests and why they discriminate

New tests live in `internal/koiosparity/pparams_test.go`, with real preview CBOR blobs and a real Koios response body in `testdata/`.

| Test | Discrimination proof |
|---|---|
| `...RationalsMatchKoiosDecimals` | Swapping `rationalsEqual` for `f.dingo == f.koios` reports 5 bogus mismatches (`a0`, `rho`, `tau`, `price_mem`, `price_step`) on identical values. |
| `TestDingoDBGetProtocolParamsResolvesEffectiveRow` | Changing `epoch <= ?` to `epoch = ?` returns nil for epoch 200. The epoch-22 and epoch-107 fixtures differ in `max_block_ex_steps` (40000000000 → 20000000000), so this is a real selection, not a match on a static default. |
| `TestDingoDBGetProtocolParamsUsesTheEpochsOwnEra` | Dropping the era filter picks the Babbage epoch-2 row and fails with `cannot decode CBOR array to struct with different number of elements`. |
| `...CostModel{EntryDiverges,LengthDiverges,LanguagePresence}`, `...RejectsMalformedKoiosCostModels` | Removing the `compareCostModels` call makes all four report zero findings. |
| `TestCheckDetectsWedgeClassProtocolParamDivergence` | End-to-end through `Check`: swapping the effective row for preview's epoch-2 Babbage parameters produces exactly the 8 expected findings. |

Also covered: missing Dingo row (with and without grace window), missing Koios row, fetch error, era mismatch, one-sided parameter presence, both-sides-absent, `DatabaseSource` parity with `DingoDB`, and a fixture-agreement test that fails if the CBOR blob and the Koios JSON ever drift apart.

## Validation

**Real-chain acceptance** — a synced preview metadata database, read-only, against a live Koios instance, through the production `OpenDingoDB` → `GetProtocolParams` → `CompareEpochProtocolParams` path. Epochs 0-20 plus 22, 50, 107, 150, 200, 300, 360, 380, 400, 410, 475:

- **25 of 31 epochs: zero mismatches** across all 25 scalar fields and both cost models. Spans two eras (Alonzo epochs 0-2, Babbage from 3), 8 distinct effective `pparams` rows, and the Alonzo→Babbage boundary at epochs 2/3 where both era rows exist.
- **Epochs 3-8: one finding each** — see below.

```
epoch   2: row=2 era=Alonzo(4)  koios=Alonzo  PlutusV1: 166 -> 0 mismatch(es)
epoch   3: row=3 era=Babbage(5) koios=Babbage PlutusV1: 166, PlutusV2: 175 -> 1 mismatch(es)
    pparams_cost_model_plutus_v2: dingo="175 entries" koios="" (value_mismatch)
epoch   9: row=9 era=Babbage(5) koios=Babbage PlutusV1: 166, PlutusV2: 175 -> 0 mismatch(es)
epoch 475: row=360 era=Babbage(5) koios=Babbage PlutusV1: 166, PlutusV2: 175 -> 0 mismatch(es)
```

- `go test ./internal/koiosparity/... -count=1` — pass
- `make lint` — golangci-lint 0 issues across all four scopes. The pinned `golangci-lint`/`nilaway` binaries in this environment were built with go1.26 and panic against the now-installed go1.27 toolchain; rebuilding them with the current toolchain resolves it. nilaway then reports 18 findings, **none** in `internal/koiosparity` and none in any file this PR touches.
- `make docs-parity` — pass

## A finding this surfaced — now explained, and tracked

On preview, Dingo's `pparams` row for epoch 3 (the Babbage transition) carries a PlutusV2 cost model, byte-identical to the one in the epoch-9 row. Koios reports **no** PlutusV2 cost model for epochs 3-8, and first reports one at epoch 9.

This was left open in an earlier revision of this description. It is now answered: @wolf31o2 identified the cause on the `compare.go` thread. `HardForkBabbage` injects a hardcoded model when the upgraded parameters carry none:

```go
if _, hasV2 := ret.CostModels[1]; !hasV2 {
    ret.CostModels[1] = DefaultPlutusV2CostModel
}
```

(`ledger/eras/babbage.go`). gouroboros' `babbage.UpgradePParams` carries `prevPParams.CostModels` forward unchanged, so nothing else adds it. The divergence is real and Dingo-side, and the chain gains PlutusV2 only at epoch 9.

It is already tracked in **#3962**, which cites the same injection and asks the sharper consensus question — whether a real node rejects a PlutusV2 transaction outright during that gap rather than pricing it against a fabricated model. #3884 has already landed the reporting half, hiding the synthetic model from `GetCurrentProtocolParams`.

No ledger change is made here. The comparison reports the divergence as `value_mismatch`, which is the correct outcome for a genuine difference in stored parameters, and this PR stays a parity-coverage change.

## Deliberately excluded

Each verified against real preview data before being excluded, and each classified in `koiosCoverageMatrix` so a report never implies it was checked:

- **`coins_per_utxo_size`** — Koios reports Alonzo's per-word figure (34482 on preview epochs 0-2) where Dingo stores 4310; the two agree from Babbage onward. Which side is right for Alonzo needs its own investigation, and including it would attach an unexplained permanent FAIL to those epochs. Cached for reference.
- **`decentralisation`, `min_utxo_value`, `extra_entropy`** — no Babbage or Conway parameter struct defines them, so on every currently live era there is no Dingo-side value.
- **`nonce`, `block_hash`** — epoch identity, not protocol parameters.
- **Conway governance parameters** (`pvt_*`, `dvt_*`, `committee_*`, `gov_action_*`, `drep_*`, `min_fee_ref_script_cost_per_byte`) — real and worth covering, but the reference chain available for verification is Babbage-era throughout, so their cross-side representation is unverified. Adding them unverified is the failure mode this exclusion list exists to avoid.

Cost models were on this list in the first commit and were moved into the comparison in the second, once the key mapping and entry-for-entry agreement were confirmed on real data.

**Scope note, to avoid overclaiming:** cost-model parity would *not* have caught #3935. That is a Plutus budget divergence on this same chain where the cost models turned out to be identical. This coverage catches a mispriced or missing cost model, not every budget disagreement.

## Docs

- `ARCHITECTURE.md` updated: `koios_epoch_params` in the cache table list, `pparams`/`epoch` in the tables-queried list, a new "Protocol parameter resolution" subsection covering the one-row-per-change and era-boundary traps and the failure classification, `pparams.go` in the file listing, and 9 new `/epoch_params` rows in the coverage table.
- `DATABASE.md` **not** updated, deliberately. No Dingo table is added or changed — `pparams` and `epoch` are only read. The new `koios_epoch_params` table lives in `cache.db`, which is the parity tool's own private cache; `ARCHITECTURE.md` already states that `DATABASE.md` does not and need not document it.

## Merge and CI state

Merged `origin/main` in (not rebased). Two conflicts, both additive: `RewardParitySource` gained `GetPoolsRetiredByEpoch` on main and `GetProtocolParams` here, and `observer_test.go` gained one new test on each side. Both sides kept in each case. No migration is touched by this branch, so #3934's `v10` is unaffected.

`go-test` is red on this branch, and on `main` itself, for a reason outside this PR. Three tests in `ledger/eras` — `TestConwayUtxoValidationRuleIndexesArePinned`, `TestValidateTxConwayGenuinelyMissingInputStillRejected` and `TestValidateTxConwayGenuinelyUnbalancedStillRejected` — fail on all four platforms at `main`'s own head (`90504583`), and this branch has no diff against main under `ledger/eras`, `go.mod` or `go.sum`. Tracked in **#3976**.

`./bin`'s two entrypoint tests (`TestEntrypointForwardsSignalsDuringMithrilBootstrap`, `TestEntrypointSignalHandlingSurvivesBootstrapToServeHandoff`) also fail under full-suite load on unmodified `main`; tracked in #3930/#3574.

Locally: `go test ./internal/koiosparity/... -count=1` passes, `make lint` reports 0 golangci-lint issues across all four scopes, `make docs-parity` passes. `nilaway` could not run (the pinned binary is go1.26-built and does not analyse the installed toolchain).

---
## Summary by cubic
Compares per-epoch protocol parameters against Koios (fixes #3931): fetch `/epoch_params`, resolve Dingo's effective `pparams` row for the epoch, and compare 25 scalar fields plus both Plutus cost models. A wrong stored parameter changes what the node accepts, so it is wedge-class — a canonical block gets rejected while every existing comparison keeps reporting PASS right up to that rejection. Epochs cached before this feature existed are backfilled with a single `/epoch_params` request on the next fetch, from either the CLI fetch path or the in-process observer, instead of refetching pool data or reporting incomplete reference data forever.

**Dingo-side resolution**
- The `pparams` table holds one row per parameter change, not per epoch, so the applicable row is the latest at or before the epoch.
- Era boundaries write both an old-era and new-era row at the same epoch with different CBOR shapes; the era is resolved from the `epoch` table first and used both to filter the row and pick the decoder.
- Numeric fields compare through `rationalsEqual`, so Dingo's exact rationals (`577/10000`) reconcile with Koios's decimal/exponent forms (`0.0577`, `7.21e-05`), and Koios numbers stay `json.Number` so nothing round-trips through a float.
- Cost models compare per language under their Koios names (PlutusV1/PlutusV2); findings name the language and the first differing entry with a differing-entry count.
- The new `koios_epoch_params` cache row is written before `CommitEpochData`, so a kill between the two leaves the epoch unfetched and it is simply re-fetched; cleanup errors are preserved rather than swallowed.
- Backfills do not advance `koios_epoch_info.fetched_at`, since a parameter row establishes no pool-level freshness.
- Account and parameter backfills are independent: an epoch missing both now gets both fetches, and only the full pool-level fetch is skipped.
- The observer seeds its backlog from missing `koios_epoch_params` rows — upgraded-cache epochs carry an info row and stored PASS, so the other backlog queries miss them — and skips pre-staking epochs, since Koios returns no parameter row for them (they previously errored on every run and hard-stopped the CLI).

**Failure classification**
- A differing value is `value_mismatch` (FAIL), never softened by the grace window.
- An unresolvable Dingo row is `dingo_db_missing` (ERROR), softened to `reference_lag` inside the grace window; a failed read is the distinct `dingo_db_error` (ERROR).
- A missing Koios row reports incomplete reference data rather than skipping, mirroring `koios_totals` handling.
- A parameter present on only one side is a mismatch, not a skip — that is what an era-gating bug looks like.

Deliberately excluded fields (`coins_per_utxo_size`, `decentralisation`, `min_utxo_value`, `extra_entropy`, `nonce`, `block_hash`, and the Conway governance block) are classed in the coverage matrix so a report never implies an uncompared field was checked; the `/epoch_info` era entry now points to the same value compared under `/epoch_params`. `coins_per_utxo_size` and the epoch 3-8 PlutusV2 finding are flagged for their own investigation; on real preview data, 25 of 31 epochs compare clean across two eras and 8 distinct `pparams` rows.

<sup>Written for commit 24a5d13c50394ba520eb5614ec573db5df63b8d4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3940?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added epoch-level protocol parameter parity checks between Dingo and Koios.
  * Comparisons now cover protocol versions, fees, deposits, limits, execution parameters, rational values, and Plutus cost models.
  * Added detailed mismatch reporting for differing, missing, or unavailable parameter data.
  * Epoch checks now retrieve and cache Koios protocol parameters alongside existing epoch data.

* **Documentation**
  * Documented protocol parameter coverage, comparison behavior, and supported fields.


internal/koiosparity/cache.go
```diff
@@ -1096,7 +1236,12 @@ func (c *Cache) GetUncachedEpochs(
 	}
 
 	rows, err := c.db.Query(
-		"SELECT epoch FROM koios_epoch_info WHERE network = ? AND epoch >= ? AND epoch <= ?",
+		`SELECT i.epoch FROM koios_epoch_info i
+		 WHERE i.network = ? AND i.epoch >= ? AND i.epoch <= ?
+			AND EXISTS (
```
wolf31o2: Blocker: `TestFetchBackfillsAccountsForPreExistingCache` fails on this branch. `fetch_test.go:433` expects `epochInfoCalls == 0` and gets 1; deterministic across repeated runs, and the same test passes at `e109235a`.

The test seeds exactly the upgraded-cache shape this clause targets — a `koios_epoch_info` row for epoch 50 and no `koios_epoch_params` row. With the EXISTS requirement, epoch 50 comes back from `GetUncachedEpochs` at `fetch.go:224`, so `Fetch` never puts it in `accountOnlyEpochs` (`fetch.go:246`) and runs the full pool-level `fetchEpoch` instead of the account-only backfill, re-requesting `/epoch_info` and `/totals`.

Re-fetching an upgraded cache may well be the behaviour you want, but it is incompatible with the account-only contract that test states (`fetch_test.go:300-307`: "/epoch_info and /totals must never be re-requested for this epoch"). One of the two has to move.

internal/koiosparity/fetch.go
```diff
@@ -668,9 +669,30 @@ func fetchEpoch(
...
 	now := time.Now()
 
-	// 1c. Skip pools that could not possibly have any history yet this
+	// The parameter row is committed BEFORE CommitEpochData, not inside it.
+	// CommitEpochData advances koios_epoch_info.fetched_at, which is the
+	// freshness marker GetEpochsNeedingCheck compares against; writing the
+	// parameters first means a process killed between the two writes leaves
+	// the epoch looking unfetched and it is simply re-fetched. Committing
+	// them after would advance fetched_at with no parameter row behind it.
+	if err := cache.UpsertEpochParams(
```
wolf31o2, line 745: Blocker: a pre-staking epoch never reaches this upsert. `fetchEpoch` commits the `PreStaking` marker and returns at `fetch.go:629`, above the `/epoch_params` fetch at `:673`. With `preStakingThroughEpoch = 1` (`check.go:295`), epochs 0 and 1 get a `koios_epoch_info` row and never a `koios_epoch_params` row.

`GetUncachedEpochs` now treats an epoch without a params row as not cached, so it returns 0 and 1 on every run forever. `Fetch` (`fetch.go:224`) and the observer backlog (`observer.go:250`, `:574`) re-request `/epoch_info` for both epochs each pass, re-commit the marker, and return without progress. That is precisely what the marker exists to stop — `fetch.go:594-596` says so in as many words. `GetEpochsMissingAccountCoverage` already guards the identical class with `AND k.pre_staking = 0` (`cache.go:1205`) and explains why at `cache.go:1192-1198`; `GetUncachedEpochs` needs the equivalent, e.g. `AND (i.pre_staking = 1 OR EXISTS (...))`.

Second consequence, on the check side: `checkEpoch` returns `StatusPass` for a pre-staking epoch at `check.go:404-449`, above the parameter comparison at `check.go:495`. Epochs 0 and 1 therefore report PASS with no protocol parameter compared at all — the outcome `CompareEpochProtocolParams`' own doc comment says must never happen. Nothing about `/epoch_params` depends on a stake snapshot, so the comparison could run for those epochs once the row exists.

internal/koiosparity/coverage.go
```diff
@@ -345,6 +345,380 @@ var koiosCoverageMatrix = []KoiosFieldCoverage{
...
+	{
+		Endpoint:   "/epoch_params",
+		Field:      "epoch_no",
+		Class:      CoverageExactMatch,
+		DingoField: "reporting epoch",
+		Reason:     "response identity must equal the requested epoch",
+	},
+	{
+		Endpoint:   "/epoch_params",
+		Field:      "era",
```
wolf31o2, line 365: This entry contradicts the `/epoch_info` `era` row at `coverage.go:89-93`, which is still classified `CoverageUnsupported` with the reason "Dingo has no persisted per-epoch era aggregate". `DingoDB.GetProtocolParams` reads the era straight out of `epoch.era_id`, which is that aggregate. Worth reconciling the older entry so the matrix does not state both that Dingo has no per-epoch era and that it is compared exactly. Non-blocking.

wolf31o2 (commented):
Two blockers, both in `GetUncachedEpochs`' new completeness clause rather than in the comparison itself. Details inline.

On the comparison: it does not pass vacuously. `rationalsEqual` (`compare.go:1207`) returns false on anything that will not parse, so an unparseable or absent value on either side becomes a `value_mismatch` rather than a skip; the only skips are both-sides-`""` and both-sides-empty cost models, each of which is a genuine agreement about era gating. A missing Koios row is `koios_epoch_params`/`dingo_db_missing` and a missing Dingo row is `protocol_params`/`dingo_db_missing` (or `reference_lag` inside the grace window), so neither reads as PASS. `TestSeededProtocolParamsFixturesAgree` pins the two real preview epoch-107 captures against each other, and `TestCheckDetectsWedgeClassProtocolParamDivergence` drives a genuinely different stored row through the whole `Check` path and gets eight fields back including both cost models — so the PASS side is anchored to real data and the FAIL side is proven end to end.

Also checked: the epoch used for `/epoch_params` and for `GetProtocolParams` is the Koios reporting epoch on both sides, with no stake-epoch offset, matching the comment at `check.go:495-503`; `GetProtocolParams`' era-first resolution is exercised against the real two-rows-at-one-epoch preview shape; `epochParamsFromKoios` keeps Koios's literal `json.Number` text, and a quoted numeric string decodes into `json.Number` too, so both Koios encodings survive.

Cubic: the four earlier P1/P2 findings (`cache.go:571`, `cache.go:1229`, `fetch.go:689`, `dingo_db.go:341`) are fixed at this head — the EXISTS clause has the right polarity, `DeleteEpochParams` invalidates on a failed epoch commit, and both lookups share one read transaction. The open P3 on `dingo_db.go:340` is a valid preference but has no effect on the drivers this opens: the sqlite DSN is already `mode=ro` and modernc.org/sqlite ignores `TxOptions.ReadOnly` (`tx.go:23`). Non-blocking.

internal/koiosparity/dingo_db.go
```diff
@@ -311,6 +311,72 @@ func (d *DingoDB) GetEpochData(
...
+//     is resolved first and used both to filter the row and to pick the
+//     decoder — the same order api/blockfrost's adapter uses.
+//
+// ctx is forwarded to the DB driver so that a cancelled context aborts the
+// query.
+func (d *DingoDB) GetProtocolParams(
+	ctx context.Context,
+	epoch uint64,
+) (*DingoProtocolParams, error) {
+	tx, err := d.db.BeginTx(ctx, nil)
```
wolf31o2: Please use a read-only transaction here (`sql.TxOptions{ReadOnly: true}`). This method performs only SELECTs; opening a read-write transaction on MySQL/Postgres needlessly requests write-capable transaction resources and a writable snapshot.

internal/koiosparity/fetch.go
```diff
@@ -862,6 +884,9 @@ outer:
 		ReservesWithdrawal: totalsResp.ReservesWithdrawal,
 		FetchedAt:          now,
 	}); err != nil {
+		if invalidateErr := cache.DeleteEpochParams(network, epoch); invalidateErr != nil {
+			return 0, fmt.Errorf("commit epoch: %w; invalidate params: %v", err, invalidateErr)
```
wolf31o2: Please preserve both failures with wrapped errors. `%v` drops invalidateErr from errors.Is/errors.As traversal; use separate `%w` verbs (or an errors.Join-based equivalent) so callers can inspect both the commit failure and cache-invalidation failure.

wolf31o2 (changes requested):

wolf31o2 (dismissed):

internal/koiosparity/cache.go
chrisguiney: Fixed at `5d947bd1`. `TestFetchBackfillsAccountsForPreExistingCache` passes; I reproduced your failure first (`epochInfoCalls` 0 → 1, deterministic) before changing anything.

You framed the choice as "one of the two has to move" — re-fetching an upgraded cache, or the account-only contract. Neither had to: the two goals are only in tension because the parameter requirement was expressed as a property of `GetUncachedEpochs`, which is all-or-nothing per epoch.

The parameter clause is gone from `GetUncachedEpochs` entirely. A missing parameter row is now found by a sibling query, `GetEpochsMissingParams`, and filled by a params-only path that issues exactly one `/epoch_params` request — mirroring how `GetEpochsMissingAccountCoverage` and `accountOnlyEpochs` already handle missing account coverage. So an upgraded cache does get backfilled, and it does not re-request `/epoch_info` or `/totals` to do it.

`fetchEpochParamsOnly` deliberately does not advance `koios_epoch_info.fetched_at`, since it establishes no pool-level freshness.

New coverage: `TestFetchBackfillsParamsWithoutRefetchingPoolData` seeds the same upgraded-cache shape and asserts the parameter row is committed while `/epoch_info` and `/totals` stay at zero. Disabling the params-only branch reproduces your exact failure, so it discriminates rather than restating the code.

internal/koiosparity/cache.go
```diff
@@ -1096,7 +1224,12 @@ func (c *Cache) GetUncachedEpochs(
 	}
 
 	rows, err := c.db.Query(
-		"SELECT epoch FROM koios_epoch_info WHERE network = ? AND epoch >= ? AND epoch <= ?",
+		`SELECT i.epoch FROM koios_epoch_info i
+		 WHERE i.network = ? AND i.epoch >= ? AND i.epoch <= ?
+		 AND NOT EXISTS (
```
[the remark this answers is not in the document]
chrisguiney: This one is worth closing the loop on, because the suggested fix is what produced the blocker above.

Flipping the clause to `EXISTS` landed in `2bfd3a9` and did make `GetUncachedEpochs` self-consistent. But it also made an upgraded cache — `koios_epoch_info` present, `koios_epoch_params` absent — come back as wholly uncached, so `Fetch` ran the full pool-level `fetchEpoch` and re-requested `/epoch_info` and `/totals`. That broke the account-only contract `TestFetchBackfillsAccountsForPreExistingCache` pins.

Resolved differently at `5d947bd1`: the parameter clause is removed from `GetUncachedEpochs` in either direction, and the incomplete-epoch case is handled by `GetEpochsMissingParams` plus a params-only fetch path. That satisfies the goal your finding identified — an upgraded cache does get its parameter row — without a whole-epoch refetch.

internal/koiosparity/cache.go
```diff
@@ -1081,8 +1221,49 @@ func (c *Cache) GetEpochsMissingAccountCoverage(
...
+// fresh pool-level Koios data but no /epoch_params row.
+//
+// It exists for the same reason as GetEpochsMissingAccountCoverage: an epoch
+// cached before parameter comparison existed would otherwise look complete to
+// GetUncachedEpochs forever and never gain a parameter row. Requiring the
+// parameter row inside GetUncachedEpochs instead would work, but at the cost
+// of re-fetching /epoch_info, /totals and every pool-history row for that
+// epoch just to obtain one parameter row — and would break the guarantee that
+// an account backfill does not re-fetch pool-level data.
+func (c *Cache) GetEpochsMissingParams(
```
wolf31o2, line 1234: Blocker: the in-process `Observer` has no caller for this query, so an existing `cache.db` never gains its parameter rows.

`Observer.Start` seeds its backlog from `GetEpochsNeedingCheck` (`observer.go:243`) and `GetUncachedEpochs` (`observer.go:250`), plus `GetEpochsMissingAccountCoverage` when accounts are enabled (`observer.go:275`). `fetchPoolsIfNeeded` gates solely on `GetUncachedEpochs(network, epoch, epoch)` (`observer.go:574`), which this branch returns to plain `koios_epoch_info` presence. Neither `GetEpochsMissingParams` nor `fetchEpochParamsOnly` is reachable from `observer.go` — only from `Fetch` (`fetch.go:256`, `fetch.go:421`).

For any epoch already in `koios_epoch_info` before this change, both outcomes are terminal:

- Already checked, `fetched_at <= last_checked_at`: `GetEpochsNeedingCheck`'s predicate (`cache.go:1150`) never re-selects it, so its stored PASS stands with no protocol parameter ever compared. That is the outcome `CompareEpochProtocolParams`' own doc comment says must not happen ("never produce a PASS that validated no parameters at all").
- Cached but unchecked — `processEpoch` does `fetchIfNeeded` then `CheckEpoch` with no atomicity (`observer.go:433`, `:450`), so a stop between them leaves exactly this state, as does a prior `koios-parity fetch` or `run --skip-check`. The check emits `koios_epoch_params`/`dingo_db_missing` → ERROR, `UpsertCheckEpochStatus` (`check.go:789`) sets `last_checked_at`, and the epoch is never re-selected or re-fetched again. In strict mode `processEpoch` → `fail` fires `FatalFunc` (`observer.go:759`).

The observer needs the same params-only backfill `Fetch` gained, gated independently of `GetUncachedEpochs` the way `fetchAccountsIfNeeded` (`observer.go:652`) is gated independently of `fetchPoolsIfNeeded`.

internal/koiosparity/cache.go
```diff
@@ -1081,8 +1221,49 @@ func (c *Cache) GetEpochsMissingAccountCoverage(
...
+	network string,
+	from, through uint64,
+) ([]uint64, error) {
+	rows, err := c.db.Query(
+		`SELECT i.epoch FROM koios_epoch_info i
+		 WHERE i.network = ? AND i.epoch >= ? AND i.epoch <= ?
```
wolf31o2, line 1250: Blocker: this needs the `AND i.pre_staking = 0` filter that `GetEpochsMissingAccountCoverage` carries and documents at `cache.go:1192-1198` for this exact class.

`fetchEpoch` commits the pre-staking marker and returns at `fetch.go:671`, above the `/epoch_params` fetch at `fetch.go:718`, so epochs 0 and 1 get a `koios_epoch_info` row and never a `koios_epoch_params` row. This query returns them on every run and `Fetch` routes them to `fetchEpochParamsOnly` (`fetch.go:421`).

On preprod Koios has no `/epoch_params` row for them. Verified against the live endpoint:

```
$ curl -s 'https://preprod.koios.rest/api/v1/epoch_params?_epoch_no=0'
[]
$ curl -s 'https://preprod.koios.rest/api/v1/epoch_params?_epoch_no=1'
[]
```

`GetEpochParams` returns `no data for epoch 0`, `classifyFetchErr` marks it transient, and epochs 0 and 1 land in `FailedEpochs` on every run, permanently. `cmd/koios-parity/run.go:114` turns a non-empty `FailedEpochs` into a hard error that skips the check phase entirely, and `cmd/koios-parity/fetch.go:126` exits non-zero.

Preview does return rows for both epochs, so there it converges after one extra request — but `checkEpoch` returns PASS at `check.go:404`, above step 1d, for pre-staking epochs, so the row it fetched is never compared.

internal/koiosparity/compare.go
```diff
@@ -315,6 +317,350 @@ func CompareEpochTotals(
...
+// 166-integer array on each side of a mismatch row is unusable, while "entry
+// 42 = 197145 (1 of 166 entries differ)" names the operation to look at. The
+// first differing index is reported because cost-model arrays are positional,
+// so the index IS the operation's identity.
+//
+// koiosJSON is the canonical JSON written by canonicalCostModels; "" means
+// Koios priced no scripts. Text that will not parse is reported rather than
+// skipped, so a corrupt cached row can never turn the whole cost-model
+// comparison into a silent pass.
+func compareCostModels(
```
wolf31o2, line 551: Non-blocking, answering the open question in the PR description: the epoch 3-8 PlutusV2 divergence is real and Dingo-side, and the cause is identifiable without further investigation.

`HardForkBabbage` injects a hardcoded model when the upgraded parameters carry none:

```go
if _, hasV2 := ret.CostModels[1]; !hasV2 {
    ret.CostModels[1] = DefaultPlutusV2CostModel
}
```

(`ledger/eras/babbage.go:138-140`, constant at `:98`). gouroboros' `babbage.UpgradePParams` carries `prevPParams.CostModels` forward unchanged (`ledger/babbage/pparams.go:308`), so nothing else adds it. Koios confirms the chain gains PlutusV2 only at epoch 9:

```
preview epoch 3: era=Babbage pv=7.0 langs={'PlutusV1': 166}
preview epoch 8: era=Babbage pv=7.0 langs={'PlutusV1': 166}
preview epoch 9: era=Babbage pv=7.0 langs={'PlutusV1': 166, 'PlutusV2': 175}
```

Worth its own issue: Dingo prices PlutusV2 scripts for six epochs where the chain defines no PlutusV2 cost model.

Consequence for this branch: preview epochs 3-8 become a permanent `value_mismatch`/FAIL, which in a strict-mode observer fires `FatalFunc` on the first of them. That follows from the classification the PR argues for and is not itself a defect here, but it should be expected.

wolf31o2 (changes requested):
Two blockers in the backfill wiring, not in the comparison.

The comparison itself resolves correctly. `pparams.epoch` is the epoch the parameters take effect in — `SetPParams` is called with `currentEpoch.EpochId+1` from both the update path (`ledger/chainsync.go:5959`, `database/pparams.go:206`) and `transitionToEra` (`ledger/state.go:3673`) — which matches Koios `/epoch_params?_epoch_no=K`, so `epoch <= ? AND era_id = ? ORDER BY epoch DESC, id DESC LIMIT 1` is the right resolution with no off-by-one. `DingoDB.GetProtocolParams` reproduces the metadata store's query verbatim (`operational.sql.go:2688`), so the two sources cannot drift. The 24 scalar fields map to the correct Koios columns, `era` is genuinely present in the `/epoch_params` response (verified live), and `rationalsEqual` returning false on an unparseable value means a malformed or one-sided field is a mismatch rather than a skip.

`go build ./...` and `go test ./internal/koiosparity/... -count=1` pass (130s). Not run: `make lint`, devnet, conformance.

internal/koiosparity/cache.go
chrisguiney: Fixed in `863ee860`. You were right that nothing reached it — I added the params-only path to `Fetch` and never wired the observer, so the whole backfill was dead for the in-process path.

`fetchIfNeeded` now calls `fetchParamsIfNeeded`, gated on the parameter row itself rather than on `GetUncachedEpochs` — the same shape as `fetchAccountsIfNeeded` gating on account coverage independently of `fetchPoolsIfNeeded`, which is the precedent you pointed at. It skips pre-staking epochs and retries a transient Koios failure on the same schedule as the other two fetches, returning a permanent error immediately.

Proven rather than asserted: `TestObserverBackfillsParamsForAPreExistingCache` seeds exactly the upgraded-cache shape (a `koios_epoch_info` row, no `koios_epoch_params` row), drives `fetchIfNeeded`, and asserts the parameter row is committed while `/epoch_info` stays at zero calls. Disabling the new gate fails it with `expected: 1, actual: 0` — so it discriminates rather than restating the code.

Your second outcome was the one I had not considered: an epoch cached-but-unchecked keeps its stored ERROR after `UpsertCheckEpochStatus` sets `last_checked_at`, and is then never re-selected by `GetEpochsNeedingCheck` nor re-fetched. That is unrecoverable without this gate, and in strict mode fires `FatalFunc`.

internal/koiosparity/cache.go
chrisguiney: Fixed in `863ee860` — `AND i.pre_staking = 0` added, with the rationale recorded next to it and cross-referencing the account-coverage sibling that documents the same class.

I verified your Koios observation rather than taking it on trust: preprod `/epoch_params?_epoch_no=0` and `_epoch_no=1` both return `[]`, so `GetEpochParams` reports no data, `classifyFetchErr` marks it transient, and the epoch lands in `FailedEpochs` on every run. `cmd/koios-parity/run.go` turns that into a hard error that skips the check phase entirely, so the effect was not merely noisy — it disabled checking.

The asymmetry you noted holds too: preview does return rows for epochs 0 and 1, so this would have been invisible on preview and broken on preprod, which is the worse failure mode of the two.

internal/koiosparity/cache.go
```diff
@@ -450,6 +503,81 @@ func (c *Cache) GetTotals(network string, epoch uint64) (*KoiosTotals, error) {
...
+	return err
+}
+
+// GetEpochParams retrieves the cached Koios /epoch_params row, returning
+// sql.ErrNoRows when absent — e.g. an epoch cached before protocol-parameter
+// fetching was added and not yet re-fetched. Callers must treat that as an
+// incomplete reference row (see CompareEpochProtocolParams's
+// "koios_epoch_params" CategoryDBMissing mismatch), never as a reason to skip
+// the parameter comparison silently.
+func (c *Cache) GetEpochParams(
```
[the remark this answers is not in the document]
chrisguiney: Valid, and it took three commits to close completely.

A missing parameter row is now found by `GetEpochsMissingParams`, not by adding a clause to `GetUncachedEpochs` — the clause version made an upgraded cache look wholly uncached and broke the account-only fetch contract (see the thread on `cache.go` line 1229). It is reached from three places: `Fetch`'s params-only path, `fetchIfNeeded` -> `fetchParamsIfNeeded` in the observer (`863ee860`), and `Observer.Start`'s backlog seed (`c8084a25`).

The last one was the real hole: an upgraded cache's epoch has a stored PASS, so `GetEpochsNeedingCheck` never returns it, nothing queued it, and `processEpoch` — and therefore the gate — never ran. `TestObserverBackfillsParamsForAPreExistingCache` now drives `Start` rather than `fetchIfNeeded`, and fails without the seed.

internal/koiosparity/fetch.go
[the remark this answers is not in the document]
chrisguiney: Valid, and fixed. When `CommitEpochData` fails, `fetchEpoch` now deletes the parameter row it wrote first, so the cache never retains new protocol parameters beside old epoch-info and pool data:

```go
return 0, fmt.Errorf("commit epoch: %w; invalidate params: %w", err, invalidateErr)
```

Both causes are wrapped, so `errors.Is`/`errors.As` still traverses either.

internal/koiosparity/dingo_db.go
```diff
@@ -311,6 +311,63 @@ func (d *DingoDB) GetEpochData(
...
+//     decoder — the same order api/blockfrost's adapter uses.
+//
+// ctx is forwarded to the DB driver so that a cancelled context aborts the
+// query.
+func (d *DingoDB) GetProtocolParams(
+	ctx context.Context,
+	epoch uint64,
+) (*DingoProtocolParams, error) {
+	var eraID uint
+	if err := d.queryRow(
```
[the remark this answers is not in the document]
chrisguiney: Valid, and fixed. `DingoDB.GetProtocolParams` now runs the era lookup and the `pparams` lookup inside one transaction, so they share a snapshot — the same property `DatabaseSource.GetProtocolParams` already had. It is opened read-only (`sql.TxOptions{ReadOnly: true}`), per the separate finding on the same line.

internal/koiosparity/dingo_db.go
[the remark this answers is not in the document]
chrisguiney: Valid, and fixed — `BeginTx(ctx, &sql.TxOptions{ReadOnly: true})`. The same change also puts both lookups in one snapshot, closing the P2 raised on this line.

chrisguiney: Done — `d.db.BeginTx(ctx, &sql.TxOptions{ReadOnly: true})`. Wrapping the two SELECTs in a transaction at all was cubic's point on the same line (the era and the `pparams` row were being read from separate snapshots across a live rollover); read-only was the right way to open it.

internal/koiosparity/fetch.go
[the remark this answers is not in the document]
chrisguiney: Fixed — both causes are wrapped with `%w`.

chrisguiney: Fixed:

```go
return 0, fmt.Errorf("commit epoch: %w; invalidate params: %w", err, invalidateErr)
```

Both the commit failure and the cache-invalidation failure stay traversable by `errors.Is`/`errors.As`.

internal/koiosparity/fetch.go
chrisguiney: Fixed, in two parts.

`GetUncachedEpochs` no longer mentions parameters at all (`5d947bd1`), so a missing parameter row cannot make an already-cached epoch look uncached and re-request `/epoch_info`.

`GetEpochsMissingParams` — the sibling query that took over the requirement — carries `AND i.pre_staking = 0` (`863ee860`), documented against `GetEpochsMissingAccountCoverage`'s identical filter and against the live `preprod` responses you quoted. Epochs 0 and 1 are therefore never selected for a backfill Koios has no row to satisfy, so they cannot land in `FailedEpochs` on every run.

`TestObserverBackfillsParamsForAPreExistingCache` seeds pre-staking markers for epochs 0 and 1 next to the epoch under test and asserts exactly one `/epoch_params` request and zero `/epoch_info` requests.

internal/koiosparity/coverage.go
chrisguiney: Fixed in `c8084a25`. You were right that the two entries could not both be true: `DingoDB.GetProtocolParams` reads the era straight out of `epoch.era_id`, so "Dingo has no persisted per-epoch era aggregate" was simply false.

The `/epoch_info` `era` row keeps `CoverageUnsupported`, because `CompareEpochAggregates` genuinely does not read it and a PASS there must not imply it did. Only the reason changed, to say that the same value is compared once against `epoch.era_id` under `/epoch_params` rather than twice. `ARCHITECTURE.md`'s `/epoch_info` unsupported row carries the same correction.

internal/koiosparity/cache.go
```diff
@@ -1081,8 +1221,49 @@ func (c *Cache) GetEpochsMissingAccountCoverage(
...
+// epoch just to obtain one parameter row — and would break the guarantee that
+// an account backfill does not re-fetch pool-level data.
+func (c *Cache) GetEpochsMissingParams(
+	network string,
+	from, through uint64,
+) ([]uint64, error) {
+	rows, err := c.db.Query(
+		`SELECT i.epoch FROM koios_epoch_info i
+		 WHERE i.network = ? AND i.epoch >= ? AND i.epoch <= ?
+		   AND NOT EXISTS (
```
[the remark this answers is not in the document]
chrisguiney: Checked, and this shape is not reachable.

`CommitEpochData` writes the `koios_epoch_info` and `koios_totals` rows in one transaction (`cache.go`, `c.db.Begin()` through `tx.Commit()`), and for a non-pre-staking epoch `fetchEpoch` fetches `/totals` before either write and aborts the epoch on failure, explicitly so a `/totals` outage cannot leave epoch-info and pool rows beside a permanently missing totals row. So `pre_staking = 0` plus a `koios_epoch_info` row implies a `koios_totals` row.

`GetEpochsMissingParams` filters on `pre_staking = 0`, so the "info but no totals" case it would have to classify cannot be produced by any path in the tool. Not changing anything here.

internal/koiosparity/fetch.go
```diff
@@ -248,6 +253,27 @@ func Fetch(
...
+		if err != nil {
+			return nil, fmt.Errorf("get epochs missing params: %w", err)
+		}
+		if len(missingParams) > 0 {
+			have := make(map[uint64]bool, len(epochs))
+			for _, e := range epochs {
+				have[e] = true
+			}
+			for _, e := range missingParams {
+				if !have[e] {
```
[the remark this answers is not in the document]
chrisguiney: Valid, and fixed in `c8084a25` — thank you, this one was real.

`accountOnlyEpochs` and `paramsOnlyEpochs` were being treated as alternatives: an epoch in both sets got only the account flag, the `switch` selected that branch, and `Fetch` returned success with the parameter row still absent. Missing account coverage and a missing parameter row are independent properties of a cached epoch, and an upgraded cache typically has both.

Both flags are now set independently — only the full pool-level fetch excludes them — and the `switch` runs both backfills for an epoch that needs both.

Reproduced before fixing: `TestFetchBackfillsAccountsForPreExistingCache` seeds exactly that cache, and now asserts `/epoch_params` was requested once and that `GetEpochParams` returns a row. Against the previous selection it fails with `expected: 1, actual: 0`.

internal/koiosparity/observer.go
```diff
@@ -550,12 +550,78 @@ func (o *Observer) fetchIfNeeded(ctx context.Context, epoch uint64) error {
 	if err := o.fetchPoolsIfNeeded(ctx, epoch); err != nil {
 		return err
 	}
+	if err := o.fetchParamsIfNeeded(ctx, epoch); err != nil {
```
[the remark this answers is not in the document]
chrisguiney: Valid, and this was the real hole — thank you.

I gated `fetchIfNeeded` on the parameter row but never asked how the epoch reaches `processEpoch` in the first place. It does not: an upgraded cache's epoch has a `koios_epoch_info` row (so `GetUncachedEpochs` skips it) and a stored PASS (so `GetEpochsNeedingCheck` skips it), nothing queues it, and the gate is unreachable. I fixed the fetching path and left the queuing path alone.

`Start` now seeds `GetEpochsMissingParams` alongside `GetEpochsNeedingCheck`, `GetUncachedEpochs` and `GetEpochsMissingAccountCoverage`, which is the precedent the account-coverage seed already sets.

`TestObserverBackfillsParamsForAPreExistingCache` now drives `Start` instead of calling `fetchIfNeeded` directly. That was the second half of the mistake: the test entered below the decision it was supposed to be pinning, so it exercised the gate while skipping what has to reach it, and looked like coverage. Every epoch in its range is already cached and already PASSed, so the epoch is queued for the missing parameter row or not at all — without the seed it fails with "Start must queue the epoch whose parameter row is missing".

internal/koiosparity/fetch.go
```diff
@@ -878,6 +945,128 @@ func unixTime(sec int64) time.Time {
 	return time.Unix(sec, 0).UTC()
 }
 
+// fetchEpochParamsOnly fetches and commits just the /epoch_params row for an
+// epoch whose pool-level data is already cached and fresh.
+//
+// It deliberately does not touch koios_epoch_info.fetched_at. That column is
+// the freshness marker for pool-level data, which this did not refetch, so
+// advancing it would claim a freshness this call has not established.
+// FetchEpochParams fetches and commits the /epoch_params row for one epoch.
```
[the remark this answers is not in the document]
chrisguiney: Fixed. `FetchEpochParams` and `fetchEpochParamsOnly` each carry their own doc comment now; the exported one no longer opens by describing the unexported one.

internal/koiosparity/compare.go
chrisguiney: Thank you — that settles it, and more precisely than the PR body could.

I have updated the PR description: the question is no longer open, the cause is `HardForkBabbage` injecting `DefaultPlutusV2CostModel` when the upgraded parameters carry none (`ledger/eras/babbage.go`), and the divergence is Dingo-side. It is already tracked in #3962, which cites the same code and takes it further — whether a real node would reject such a transaction outright rather than price it against a fabricated model — with #3884 having already fixed the reporting half.

No ledger change here. The comparison reports the divergence as `value_mismatch`, which is the correct behaviour for a real, Dingo-side difference in stored parameters, and this PR stays a parity-coverage change.

chrisguiney:
Status after merging `main` and addressing the outstanding findings.

**Conflicts.** `origin/main` merged in (`ed0b956a`), not rebased. Two conflicts, both additive: `RewardParitySource` gained `GetPoolsRetiredByEpoch` on main and `GetProtocolParams` here; `observer_test.go` gained one new test on each side. Both sides kept in each case. This branch touches no migration, so #3934's `v10` is unaffected.

**Findings.** Every open thread is answered. `c8084a25` fixes three real ones and corrects two pieces of prose:

- `Observer.Start` never queued an epoch whose parameter row was missing. An upgraded cache's epoch has a `koios_epoch_info` row and a stored PASS, so none of the three existing backlog seeds return it, `processEpoch` never runs, and the `fetchParamsIfNeeded` gate added in `863ee860` was unreachable. `GetEpochsMissingParams` is now seeded alongside the others. This was cubic's P1 and it was correct — I had fixed the fetching path and left the queuing path alone.
- `TestObserverBackfillsParamsForAPreExistingCache` now drives `Start` instead of calling `fetchIfNeeded` directly. It was entering below the decision it was meant to pin, so it looked like coverage while skipping what has to reach the gate.
- `Fetch` treated the account and parameter backfills as alternatives; an epoch missing both got only the account one and `Fetch` returned success with no parameter row. Reproduced first (`expected: 1, actual: 0` on `/epoch_params` requests), then fixed.
- `FetchEpochParams`' doc comment was split out of `fetchEpochParamsOnly`'s, and the `/epoch_info` `era` coverage reason no longer claims Dingo has no persisted per-epoch era while `/epoch_params` compares exactly that column.

The PlutusV2 question in the description is no longer open — @wolf31o2 identified the cause (`HardForkBabbage` injecting `DefaultPlutusV2CostModel`), the description now says so, and it is tracked in #3962. No ledger change here.

**CI.** Red, for reasons outside this PR. This branch's diff against `main` is `ARCHITECTURE.md` plus `internal/koiosparity/` — nothing else.

- All four platforms fail the three `ledger/eras` tests (`TestConwayUtxoValidationRuleIndexesArePinned`, `TestValidateTxConwayGenuinelyMissingInputStillRejected`, `TestValidateTxConwayGenuinelyUnbalancedStillRejected`). Same three fail on unmodified `origin/main` at `90504583`. Tracked in #3976.
- `go-test (Linux, race)` additionally failed `ledger.TestWindowedRewindConvergesWhilePrimaryChainExtends` with `remove block at index 210: block not found`. That is intermittent, not this PR: `ledger/` is byte-identical to `origin/main` on this branch, the test arrived with a `main` commit (`f86d64f4`), it deliberately races a block appender against a windowed rollback, it passes 5/5 locally under `-race` in isolation, and the same job on #3924 at the same `main` did not hit it. I have not seen an issue for it; happy to file one if it is not already known.

Local: `go test ./internal/koiosparity/... -count=1` passes, `make lint` 0 golangci-lint issues across all four scopes including the Windows target, `make docs-parity` passes. `nilaway` could not run — the pinned binary is go1.26-built, fails prerequisites against the installed go1.27.1 toolchain, and exits 0 anyway, so I am reporting it unavailable rather than clean.

Bot review: cubic reviewed this head with 0 issues. CodeRabbit reports "Review skipped: incremental reviews are disabled" rather than a completed review.

**Not merging.** `go-test` is red, so "all checks pass" is not met on the letter of the rule regardless of cause, and the standing review is still Changes requested. @wolf31o2, re-review requested.

chrisguiney:
Following up on the `go-test (Linux, race)` extra failure: filed as #3979.

Confirmed intermittent rather than deterministic — `go test ./ledger/ -race -count=1` over the whole package passes locally (374s, the closest equivalent to the CI job), as does the single test at `-count=5`. The test races a block appender against `rollbackPrimaryChainInSecurityParamWindows` by design, and the appender tolerates the tip moving underneath it while the rollback side does not.

So the red CI on this branch reduces entirely to #3976 (deterministic, all four platforms) plus #3979 (intermittent, race job only). Neither is reachable from this PR's diff.

chrisguiney:
Re-review requested — both blockers from `5d947bd1` are addressed.

You identified two blockers in the backfill wiring rather than the comparison.
Fixed in **`c8084a25`**, *"reach every epoch whose parameter row is missing"*:

1. **The parameter gate was unreachable.** `Observer.Start` seeded its backlog
   from `GetEpochsNeedingCheck`, `GetUncachedEpochs` and
   `GetEpochsMissingAccountCoverage`. An upgraded cache's epoch has an info row
   *and* a stored PASS, so none of the three return it and `processEpoch` never
   ran — the fetching path had been fixed while the queuing path was left
   alone. Added a `GetEpochsMissingParams` seed.
2. **`Fetch` treated the account and parameter backfills as alternatives**, so
   an epoch missing both got only the account branch and `Fetch` returned
   success with no parameter row. The two flags are now independent.

The regression test drives `Start` rather than `fetchIfNeeded` — the earlier
test entered below the decision it was meant to pin, which is why the gap
survived. It fails without the seed. The `Fetch` fix was proven with
`expected: 1, actual: 0` on `/epoch_params` requests.

CI is green on all four platforms; the earlier red was #3976 on `main`, fixed
by #3977 and merged in here.

wolf31o2 (approved):

PR #3941, opened 2026-09-05
fix(ledger): make a recovery rewind reachable and fail closed when it is not

### LINKED ISSUE #3889: ledger: a deterministic rejection can compute a rewind beyond K and then retry it forever
## Summary

When a transaction is rejected deterministically on the primary chain, the node
computes a rewind target that can lie further back than the security parameter
allows. The rollback is then refused as unsafe, the pipeline restarts, and the
same impossible rewind is attempted again — indefinitely.

This is what turned a single bad block into a nine-hour outage while
investigating #3875. The transaction-level cause there is fixed
(blinklabs-io/gouroboros#2198); this is the recovery behaviour that sat behind
it, and it is independent of what caused the rejection.

## Observed

A from-genesis Preview replay, tip 12924731, rejecting a transaction in the
block at 12924758:

```
level=WARN  msg="rejecting rollback that exceeds security parameter K"
  fork_depth=57750 security_param=432 rollback_slot=12924731
level=WARN  msg="rejecting rollback that exceeds security parameter K"
  fork_depth=66639 security_param=432 rollback_slot=12924731
level=WARN  msg="block processing failed, restarting pipeline"
  error="process block batch: rewind primary chain after deterministic
         transaction validation failure: rollback primary chain to intermediate
         point 622584: rollback depth exceeds security parameter K"
level=ERROR msg="ledger pipeline stuck: repeated restarts are not advancing the
  tip, so the failure is deterministic and will not clear on its own; the node
  is no longer following the chain"
  consecutive_no_progress=1150 tip_slot=12924731 backoff=30s
```

`consecutive_no_progress` reached 1150 over roughly nine hours before the node
was stopped by hand.

## Two things worth separating

**The rewind target is unreachable.** The `intermediate point` differs on every
attempt — 580967, 591927, 606597, 622584, 631397, 640287, 664217, 691597,
693207, and many more — while the tip sits at 12924731 and `security_param` is
432. Whatever these values are, they are not a point the node may legally rewind
to, and the same computation is repeated each time with a different answer.

**The reported fork depth is inconsistent.** Interleaved with the plausible
depths (433, 435, 439 — just over K, which is what a legitimate near-K rejection
looks like) are values like 57750 and 66639 against the same tip and the same
security parameter. Those two shapes appear in the same run seconds apart, so at
least one of them is not measuring what its name says.

## Why it matters beyond this one block

The watchdog already diagnoses the situation correctly — it says the failure is
deterministic and will not clear on its own. Having said that, the node keeps
retrying anyway. A deterministic rejection whose recovery is impossible by
construction should fail closed: stop, report, and let an operator decide,
rather than spin at 30-second intervals writing the same error.

The practical cost is not just the wasted node. The log grew steadily through the
loop, and any operator watching that node sees a healthy-looking process that is
silently no longer following the chain.

## Related

- #3875 — the transaction-level cause of this particular rejection, now fixed
  upstream.
- #3811 and #3808 — the same "deterministic failure spins without backoff"
  family, on the header-validation and plateau paths. This issue adds the case
  where the recovery itself cannot succeed.

## Environment

- epoch 149, Babbage
- `main` plus #3843, #3846, #3853, #3855 and a patched gouroboros

chrisguiney:
Fixes #3889

## Why

A recovery rewind read the primary chain tip once, then derived every
intermediate target as `snapshot - n*k`. Nothing serialises that descent
against chain growth — it runs under `transactionEventMutex` while blockfetch
appends under `chainsyncMutex` — so one block landing after the snapshot put
the next target more than `k` below the tip `Chain.Rollback` measures fork
depth against, and the rewind was refused. That error reached
`ledgerProcessBlocks` as an ordinary retryable failure, the pipeline restarted,
and recovery recomputed the same unreachable descent against a chain that had
grown further from the pinned applied tip.

The live Preview reproduction in the issue shows both halves:

```
WARN  rejecting rollback that exceeds security parameter K
      fork_depth=5010 security_param=432 rollback_slot=41098815
WARN  block processing failed, restarting pipeline
      error="... rollback primary chain to intermediate point 1774034:
             rollback depth exceeds security parameter K"
```

97 attempts on one transaction, tip pinned at slot 41098815 for over twenty
minutes, the intermediate point different every attempt (1768501, 1774034,
1773427, 1779454, 1782037, 1769017) and the fork depth growing monotonically
because the peer keeps extending while the local tip is stuck. The watchdog
announced the failure as deterministic and the node kept retrying anyway.

## What changed

**`ledger/replay_recovery.go` — each step comes from the live tip.**
`rollbackPrimaryChainInSecurityParamWindows` now asks the chain for the point a
security parameter behind its *current* tip (`Chain.PointAtDepth`) on every
iteration, so each rollback is k-bounded by construction however far the chain
has moved. A step the chain still refuses is retried a bounded number of times
against a freshly read tip, and only while nothing was published — retrying
after a publish would tell a `ledger.tx` consumer to undo the same block twice.
A descent whose committed steps stop landing below the previous one has been
outrun by chain growth and gives up instead of truncating and re-truncating for
as long as peers serve blocks.

**`ledger/replay_recovery.go` — an unreachable rewind fails closed.** All five
recovery rewind call sites now go through `rewindPrimaryChainForRecovery`. An
over-K refusal (or a non-converging descent) means recovery has no legal target
at all, and a restart only re-derives it against a larger gap, so the refusals
are tallied against the applied ledger tip. Past
`maxRecoveryRewindRejections` at the same applied high-water mark the failure is
reported as `errHaltLedgerPipeline`: the pipeline stops, logs once at ERROR with
an operator hint, and sets `dingo_ledger_pipeline_halted`. Forward progress past
that mark clears the tally through `resetRecoveryRewindRejections`, alongside
the existing at-tip, replay, and Mithril-boundary resets.

The tally keys on the applied tip rather than on the target because the target
is recomputed and different on every attempt — memoising "this target failed"
would never fire. It is the same convergence signal
`observeMithrilBoundaryRejection` already uses, for the same reason.

**`ledger/block_event.go` — `validateAndEmitRollbackUndoEmitted`.** The
reject-then-emit guard is unchanged and still inseparable; it now also reports
whether it read any block to undo, which is what makes the retry above sound.
Reporting the read rather than the publish is deliberately conservative.

**`chain/chain.go` — a failed add batch must not resurrect a lowered tip.**
`addRawBlocks` releases both chain locks when its transaction closure returns
and only then does `txn.Do` commit, so the Commit-failure restore ran with the
chain open to everyone else. Rolling the primary chain back while blockfetch
appends to it is exactly what a recovery rewind does, and the unconditional
restore wrote a pre-batch snapshot over the rollback's result, raising
`tipBlockIndex` above blocks the rollback had deleted — the chain then claimed a
tip it did not store, and the windowed rewind's next `PointAtDepth` reported the
block missing. `batchRestoreIsSafeLocked` restores only while the chain still
shows what that batch left behind. Found by the regression test below, with the
stomp confirmed directly by instrumenting the restore path (`have=68 saved=76`,
`have=83 saved=91`).

## Why halt rather than clamp

A rollback deeper than `k` is outside what the protocol permits, so silently
clamping the target would move the chain somewhere recovery did not ask for.
The change makes the *legal* rewind reachable — each step is exactly `k` from
the live tip, which is what the windowed helper always intended — and when even
that is refused it reports the condition rather than inventing a target. The
deterministic rejection itself stays non-terminal, unchanged: what becomes
terminal is only the rewind that carries it out being impossible.

## Tests and why they discriminate

All three ledger tests compile against `main` and fail there:

- `TestWindowedRewindConvergesWhilePrimaryChainExtends` — a goroutine extends
  the primary chain each time the descent moves the tip. On `main`: 25/25 runs
  fail with `rollback primary chain to intermediate point 208: rollback depth
  exceeds security parameter K`. With the fix: 360/360 runs pass under `-race`
  at `GOMAXPROCS` 2, 8, and default.
- `TestDeterministicTxRecoveryHaltsOnUnreachableRewind` — repeats
  `recoverFromDeterministicTxValidationError` against a rewind the chain always
  refuses. On `main` it never halts after 32 attempts; the failure message is
  the reported one.
- `TestRecoveryRewindHaltsThoughTargetMovesAndDepthGrows` — pins the live run's
  two properties: the chain is extended between attempts so every attempt
  computes a *different* step target against a *larger* gap, the applied ledger
  tip stays pinned, and the halt must still arrive. It asserts the targets
  actually differed and the depth actually grew, so it fails both for a fix that
  memoises a failing target and for one that lets a growing depth rearm the
  budget. On `main`: `rollback primary chain to intermediate point 107: rollback
  depth exceeds security parameter K`, never terminal.
- `TestRecoveryRewindHaltBudgetResetsOnTipProgress` — the other side of the
  budget: a tip advance past the stalled high-water mark starts a fresh one.
- `chain/batch_restore_internal_test.go` — unit-pins which chain states the
  failed-batch restore may write over (unchanged, concurrent rollback,
  concurrent append, same index different block).

## Validation

- `go test ./ledger/... ./chain/... ./database/... -count=1` — pass.
- `go test ./internal/test/conformance/ -count=1` — pass.
- `go build ./...`, `go vet` — clean.
- `make lint` — golangci-lint 0 issues on all four passes (repo, antithesis,
  examples, `GOOS=windows`).
- `make docs-parity` and `go test ./internal/architecture` — pass.
- Full `go test -race -timeout 30m ./...` — the only failures are
  `TestEntrypointForwardsSignalsDuringMithrilBootstrap` and
  `TestEntrypointSignalHandlingSurvivesBootstrapToServeHandoff` in
  `./bin`, which fail identically on unmodified `origin/main` under the same
  full-suite load and pass in isolation on both. Pre-existing, unrelated.

## Checks not run

- **DevNet (`internal/test/devnet/run-tests.sh`)** — not run: Docker is not
  available in this environment (`docker info` fails).
- **`nilaway`** — not run: not installed here, and it is not pinned or in CI.
  golangci-lint, which is, reports 0 issues.
- **Antithesis** — not run: requires infrastructure not available here.

## Docs

`ARCHITECTURE.md` updated: the windowed-rewind paragraph now states that step
targets come from the live tip and why, the bounded retry and its
no-publish precondition, the descent's give-up condition, the
`addRawBlocks` restore rule, and the new terminal bound with its applied-tip
convergence signal. The "the rejection is never terminal" sentence is corrected
to distinguish the verdict from the rewind. `DATABASE.md` needs no change: no
schema, key, or storage-format change.

## Scope

The underlying Plutus rejection that triggered the live run is #3935 and is not
touched here. This is the amplifier only.

---
## Summary by cubic
Makes recovery rewinds derive each step target from the chain's live tip instead of a snapshot taken when the descent began, so chain growth between steps can no longer push a rollback past the security parameter and get it refused forever. When a rewind stays unreachable at a pinned applied tip, the pipeline now halts with an operator hint instead of restarting and recomputing the same impossible rewind, fixing #3889.

**Bug Fixes**
- Each rewind step reads the point `k` behind the live tip via `Chain.PointAtDepth`; a step the chain still refuses is retried a bounded number of times against a fresh tip, and only when it published no undo events so `ledger.tx` consumers are never told to undo the same block twice.
- A descent whose committed steps stop landing below the previous one gives up as non-converging rather than re-truncating indefinitely; the retry and non-convergence budgets are separate so transient over-`k` refusals can't consume the convergence budget.
- All five recovery rewinds route through `rewindPrimaryChainForRecovery`, which tallies over-`k` refusals at an applied tip that never advances and reports `errHaltLedgerPipeline` past `maxRecoveryRewindRejections`; advancing past that tip clears the tally.
- `addRawBlocks` now restores its pre-batch snapshot on commit failure only while the chain still shows exactly what that batch left behind—same tip index, tip hash, and mutation generation—so a concurrent recovery rollback result is never overwritten with a tip the chain no longer stores.

<sup>Written for commit c72fcfb5cbb6bb1ecba37b8d844868c44d0b3a38. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3941?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery rewinds by recalculating targets from the current chain tip.
  - Recovery now retries transient rewind failures while preventing endless retries when progress cannot be made.
  - The ledger pipeline halts after repeated unrecoverable rewind rejections.
  - Improved protection against inconsistent chain state during concurrent rollback and block restoration.

- **Documentation**
  - Updated architecture documentation to clarify recovery rewind behavior and failure handling.

- **Tests**
  - Added coverage for recovery convergence, halt conditions, progress resets, and concurrent chain updates.

### DIFF
--- a/chain/batch_restore_internal_test.go
+++ b/chain/batch_restore_internal_test.go
@@ -1,0 +1,111 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package chain
+
+import (
+	"testing"
+
+	ochainsync "github.com/blinklabs-io/gouroboros/protocol/chainsync"
+	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
+)
+
+// TestBatchRestoreIsSafeLocked pins which chain states a failed add batch may
+// write its pre-batch snapshot back over.
+//
+// addRawBlocks releases both chain locks when its transaction closure returns
+// and only then does txn.Do commit, so the Commit-failure restore runs with
+// the chain open to everyone else. Rolling the primary chain back while
+// blockfetch appends to it is not a corner case -- it is what a ledger
+// recovery rewind does on every deterministic rejection -- and the restore
+// used to write its snapshot back unconditionally. That raised tipBlockIndex
+// to a value above the blocks the concurrent rollback had already deleted, so
+// the chain claimed a tip it did not store: the ledger's windowed rewind then
+// asked for the point a security parameter behind that tip and was told the
+// block did not exist (issue #3889).
+func TestBatchRestoreIsSafeLocked(t *testing.T) {
+	applied := ochainsync.Tip{
+		Point:       ocommon.Point{Slot: 100, Hash: []byte("applied")},
+		BlockNumber: 10,
+	}
+	const appliedIndex = uint64(10)
+
+	for _, tc := range []struct {
+		name  string
+		chain *Chain
+		want  bool
+	}{
+		{
+			name: "unchanged since the batch",
+			chain: &Chain{
+				tipBlockIndex: appliedIndex,
+				currentTip:    applied,
+			},
+			want: true,
+		},
+		{
+			name: "concurrent rollback lowered the tip",
+			chain: &Chain{
+				tipBlockIndex: appliedIndex - 4,
+				currentTip: ochainsync.Tip{
+					Point: ocommon.Point{
+						Slot: 60,
+						Hash: []byte("rolled-back"),
+					},
+					BlockNumber: 6,
+				},
+			},
+			want: false,
+		},
+		{
+			name: "concurrent append raised the tip",
+			chain: &Chain{
+				tipBlockIndex: appliedIndex + 1,
+				currentTip: ochainsync.Tip{
+					Point: ocommon.Point{
+						Slot: 110,
+						Hash: []byte("appended"),
+					},
+					BlockNumber: 11,
+				},
+			},
+			want: false,
+		},
+		{
+			name: "same index, different block",
+			chain: &Chain{
+				tipBlockIndex: appliedIndex,
+				currentTip: ochainsync.Tip{
+					Point: ocommon.Point{
+						Slot: 100,
+						Hash: []byte("other-fork"),
+					},
+					BlockNumber: 10,
+				},
+			},
+			want: false,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			got := tc.chain.batchRestoreIsSafeLocked(applied, appliedIndex)
+			if got != tc.want {
+				t.Fatalf(
+					"batchRestoreIsSafeLocked() = %v, want %v",
+					got,
+					tc.want,
+				)
+			}
+		})
+	}
+}
--- a/chain/batch_restore_internal_test.go
+++ b/chain/batch_restore_internal_test.go
@@ -40,6 +40,7 @@ func TestBatchRestoreIsSafeLocked(t *testing.T) {
 		BlockNumber: 10,
 	}
 	const appliedIndex = uint64(10)
+	const appliedGeneration = uint64(1)
 
 	for _, tc := range []struct {
 		name  string
@@ -49,8 +50,9 @@ func TestBatchRestoreIsSafeLocked(t *testing.T) {
 		{
 			name: "unchanged since the batch",
 			chain: &Chain{
-				tipBlockIndex: appliedIndex,
-				currentTip:    applied,
+				tipBlockIndex:      appliedIndex,
+				mutationGeneration: appliedGeneration,
+				currentTip:         applied,
 			},
 			want: true,
 		},
@@ -98,7 +100,9 @@ func TestBatchRestoreIsSafeLocked(t *testing.T) {
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
-			got := tc.chain.batchRestoreIsSafeLocked(applied, appliedIndex)
+			got := tc.chain.batchRestoreIsSafeLocked(
+				applied, appliedIndex, appliedGeneration,
+			)
 			if got != tc.want {
 				t.Fatalf(
 					"batchRestoreIsSafeLocked() = %v, want %v",
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -686,6 +686,29 @@ func (c *Chain) AddRawBlocksWithCallback(
 	return c.addRawBlocks(blocks, callback)
 }
 
+// batchRestoreIsSafeLocked reports whether a failed add batch may write its
+// pre-batch snapshot back over the current chain state.
+//
+// It may only do so while the chain still shows exactly what that batch left
+// behind. addRawBlocks releases c.mutex and c.manager.mutex when its
+// transaction closure returns, before txn.Do commits, so any other goroutine
+// can move the chain before the Commit-failure path runs -- and rolling the
+// primary chain back while blockfetch appends to it is what every ledger
+// recovery rewind does. Writing the snapshot over a rollback's result raises
+// tipBlockIndex back above blocks that rollback deleted, leaving the chain
+// claiming a tip it does not store: the next rollback measures an inflated
+// fork depth against it, and any lookup in the resurrected span reports the
+// block as missing.
+//
+// Callers must hold c.mutex and c.manager.mutex.
+func (c *Chain) batchRestoreIsSafeLocked(
+	appliedTip ochainsync.Tip,
+	appliedTipBlockIndex uint64,
+) bool {
+	return c.tipBlockIndex == appliedTipBlockIndex &&
+		bytes.Equal(c.currentTip.Point.Hash, appliedTip.Point.Hash)
+}
+
 func (c *Chain) addRawBlocks(
 	blocks []RawBlock,
 	callback func(RawBlock, *database.Txn) error,
@@ -713,11 +736,13 @@ func (c *Chain) addRawBlocks(
 		// leaving the in-memory chain advanced. Capture the
 		// snapshot here so we can also restore on Commit failure.
 		var (
-			snapshotTaken      bool
-			savedTip           ochainsync.Tip
-			savedTipBlockIndex uint64
-			savedHeaders       []queuedHeader
-			savedBlocks        []ocommon.Point
+			savedTip             ochainsync.Tip
+			savedTipBlockIndex   uint64
+			savedHeaders         []queuedHeader
+			savedBlocks          []ocommon.Point
+			batchApplied         bool
+			appliedTip           ochainsync.Tip
+			appliedTipBlockIndex uint64
 		)
 		err := txn.Do(func(txn *database.Txn) error {
 			batch := blocks[batchOffset : batchOffset+batchSize]
@@ -734,7 +759,6 @@ func (c *Chain) addRawBlocks(
 			if !c.persistent {
 				savedBlocks = slices.Clone(c.blocks)
 			}
-			snapshotTaken = true
 			for _, rb := range batch {
 				evt, err := c.addRawBlockLocked(
 					rb,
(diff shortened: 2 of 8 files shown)


ledger/replay_recovery.go
```diff
@@ -787,73 +798,263 @@ func (ls *LedgerState) rollbackPrimaryChainInSecurityParamWindows(
...
-	targetIndex := uint64(0)
+	// Refuse to truncate anything toward a target the store does not hold:
+	// the descent commits each step as it goes, so discovering the target is
+	// unreachable part-way leaves the chain shortened for nothing.
 	if point.Slot > 0 || len(point.Hash) > 0 {
-		targetBlock, err := database.BlockByPoint(ls.db, point)
-		if err != nil {
-			return fmt.Errorf("lookup replay recovery target: %w", err)
+		if _, err := database.BlockByPoint(ls.db, point); err != nil {
+			return fmt.Errorf("lookup recovery target: %w", err)
```
wolf31o2, line 806: The reason given above this check needs chain membership, but `database.BlockByPoint` establishes only store presence: it resolves any block by `(slot, hash)` and says nothing about whether the primary chain currently holds it.

A target that is in the store but not on this chain — the abandoned-fork shape `rollbackPointBlock` documents at `chain/chain.go:1300-1310` — passes this check, the descent then commits every intermediate step, and `chain.Rollback(point)` fails with `ErrRollbackPointNotOnChain` on the final one. That is the outcome the comment says the check prevents: the chain shortened toward a target it could never reach.

`ls.chain.ValidateRollback(point)` at entry closes it. Failing that way up front changes no caller's disposition, since `recoverFromDeterministicTxValidationError` and `yieldedToChainSelection` both already treat `ErrRollbackPointNotOnChain` as benign — it only avoids the truncation.

chain/chain.go
```diff
@@ -756,23 +789,50 @@ func (c *Chain) addRawBlocks(
...
+			// either.
+			if batchApplied {
 				c.mutex.Lock()
 				c.manager.mutex.Lock()
-				c.currentTip = savedTip
-				c.tipBlockIndex = savedTipBlockIndex
-				c.headers = savedHeaders
-				if !c.persistent {
-					c.blocks = savedBlocks
+				if c.batchRestoreIsSafeLocked(
```
wolf31o2, line 824: When this returns false the restore is skipped silently. Memory then keeps the post-batch tip while the DB rolled the batch back, so the chain claims blocks the commit never persisted — the same class of divergence the restore exists to prevent, chosen because it does not stomp a concurrent rollback.

That trade-off is right, but nothing records that it was taken. Log at ERROR here with `appliedTipBlockIndex`, `c.tipBlockIndex`, `appliedGeneration` and `c.mutationGeneration`, so the divergence is attributable when it surfaces later as a missing block rather than as a commit failure.

ledger/replay_recovery.go
```diff
@@ -787,73 +798,263 @@ func (ls *LedgerState) rollbackPrimaryChainInSecurityParamWindows(
...
+	"recovery rewind is not gaining on its target",
+)
+
+// maxWindowedRewindStalls bounds consecutive windowed steps the chain refuses
+// for exceeding K. Each step is derived from the chain's live tip, so a
+// refusal means blocks landed between that read and the truncation; re-reading
+// the tip is what clears it, and a couple of retries covers any realistic
+// interleaving. Past that the chain is being extended at least as fast as the
+// descent moves and the rewind is not going to complete.
+const maxWindowedRewindStalls = 3
```
wolf31o2: `stalls` carries two different conditions with two different comparisons: `stalls < maxWindowedRewindStalls` gates the over-K retry at :872 and :881, and `stalls > maxWindowedRewindStalls` declares non-convergence at :908. The budgets are therefore 3 retries and 4 non-gaining steps, and an over-K retry spends budget the convergence check later reads.

Both exits are bounded, so this is not a defect. Either split them into two counters or state in this doc comment that the budget is shared and why one uses `<` and the other `>`.

wolf31o2 (commented):
Reviewed the rewind loop, the terminal budget, and the batch-restore guard.

No blockers. Three recommendations inline, none of which changes the fix's behaviour on the reported failure.

Verified:

- `PointAtDepth(k)` yields `rollbackForkDepth == k` exactly (`chain/chain.go:1264`), and `ValidateRollback` refuses only `forkDepth > securityParam` (`chain/chain.go:1119-1125`), so each step sits on the legal boundary rather than one past it.
- `validateAndEmitRollbackUndoEmitted` returns `emitted == false` on every error path — validation precedes emission and `emitRollbackTransactionEvents` cannot fail — so the retry taken on a validation error cannot double-emit even though it does not consult `emitted`.
- `errHaltLedgerPipeline` reaches `ledgerProcessBlocks`. `yieldedToChainSelection` swallows only `chain.ErrRollbackPointNotOnChain`, and `recoverFromDeterministicTxValidationError`'s shortcut is the same predicate; every other path wraps with `%w`.
- `ls.RLock()` in `rewindPrimaryChainForRecovery` is taken after `rollbackPrimaryChainInSecurityParamWindows` has released `transactionEventMutex`, and no caller holds `ls.Lock()` across the call.
- All five recovery rewinds route through the new helper; the only remaining direct callers of `rollbackPrimaryChainInSecurityParamWindows` are tests.
- `recoveryRewind` is read and written only from `ledgerProcessBlocksFromSource`'s goroutine, matching the three existing sibling tallies.

chain/batch_restore_internal_test.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, tc := range []struct {
+		name  string
+		chain *Chain
+		want  bool
+	}{
+		{
+			name: "unchanged since the batch",
+			chain: &Chain{
+				tipBlockIndex:      appliedIndex,
+				mutationGeneration: appliedGeneration,
```
wolf31o2, line 54: Blocking test gap: no case isolates mutationGeneration. The current false cases already fail on the tip index or hash, so deleting the generation check would leave every test green. Add a case with matching tip index/hash and a different generation, and set mutationGeneration to appliedGeneration in the cases intended to exercise the other predicates.

chain/chain.go
```diff
@@ -686,6 +689,31 @@ func (c *Chain) AddRawBlocksWithCallback(
...
+// block as missing.
+//
+// Callers must hold c.mutex and c.manager.mutex.
+func (c *Chain) batchRestoreIsSafeLocked(
+	appliedTip ochainsync.Tip,
+	appliedTipBlockIndex uint64,
+	appliedGeneration uint64,
+) bool {
+	return c.tipBlockIndex == appliedTipBlockIndex &&
+		c.mutationGeneration == appliedGeneration &&
```
wolf31o2, line 713: Blocking: header queue changes do not advance mutationGeneration, so this guard can still pass after another goroutine queues or prunes a header during the commit-failure window. The subsequent restore can silently discard that concurrent header mutation. Track the header state in the guard or restore only state still owned by this batch.

wolf31o2 (changes requested):

wolf31o2 (approved):

wolf31o2:
Validation update (sanitized): the merged recovery-path change was tested against the preserved Preprod database. The candidate opened the database and exited cleanly because the stored consensus stake snapshots were produced by an older accounting version and could not be safely reconstructed from that database; this is a database compatibility limitation, not a new crash. Separately, the Preprod API-mode metadata backfill completed at 2026-09-06 21:44:12 EDT and was serving through slot 132622943, approximately 1,028 slots behind the observed header tip at that measurement, at approximately 1.31 CPU cores and 12.8 GiB RSS. No private infrastructure details are included.

PR #3948, opened 2026-09-05
fix(mithril): pin artifact identity across interrupted imports

### LINKED ISSUE #3946: fix(mithril): pin snapshot identity across interrupted imports
## Problem

An interrupted Mithril v2 import does not durably pin the selected artifact for the next run. The import checkpoint is keyed by a digest/ledger-slot string, but the top-level sync dispatch does not persist or consult that artifact identity before selecting the latest aggregator snapshot again.

## Evidence

- `mithril.Sync` resolves the current target through `GetLatestCardanoDatabaseSnapshot` on each run.
- The durable `mithril_immutable_max` marker is written only after the complete import path clears sync state; it does not identify an in-progress artifact.
- Ledger-state phase checkpoints can resume only when the newly selected snapshot reconstructs the same `{digest prefix}:{slot}` import key.
- If the aggregator has advanced between an interruption and restart, the new key has no completed-phase checkpoint, so already-completed ledger-state phases are replayed instead of resuming the interrupted artifact.
- Catch-up deliberately disables phase checkpoint resume, so this is a resume-efficiency gap rather than a claim of data corruption.

## Expected behavior

Persist the selected artifact identity and its certified tip before mutating the database. On restart, resume that artifact while its cache and checkpoints are valid; if it is unavailable or invalid, fail clearly or deliberately restart from a newly verified artifact. Keep the existing divergence and certificate checks.

## Acceptance criteria

- An interruption after any ledger-state import phase records the full artifact identity, beacon, and certified tip needed for resume.
- A restart does not silently switch to a newer artifact and discard valid phase checkpoints.
- Cache/checkpoint mismatch is detected and produces an explicit recovery decision.
- Tests cover interruption followed by aggregator advancement, cache removal, and successful same-artifact resume.
- No correctness regression in fresh bootstrap or v2 catch-up.

Related: #2694 (v2 catch-up and immutable import marker), #1893 (operational profiling).

### LINKED ISSUE #3947: fix(mithril): diagnose certified digest mismatches across archive replicas
## Problem

Mithril v2 bootstrap can fail repeatedly when a published immutable archive does not match the certified digest list. A reproducible Mainnet snapshot 9118 failure reports a SHA-256 mismatch for immutable archive `05471.chunk`.

## Evidence

- The selected snapshot's verified digest list expects one byte sequence for `05471.chunk`.
- The downloaded/extracted archive produces a different SHA-256 digest during `checkImmutableTrio`.
- The current retry loop can try alternate locations, but if the replicas or a reused cache contain the same incorrect bytes, bootstrap still fails with no durable artifact/replica diagnosis.
- The mismatch is detected before database import; no claim of database corruption is made.

## Expected behavior

Report the artifact hash/beacon, immutable file number, expected digest, observed digest, and location/attempt in a redacted, actionable error. Distinguish a certified-list/provider mismatch from a local cache mismatch. Preserve the fail-closed behavior and never import a mismatched trio.

## Acceptance criteria

- A mismatch records enough non-sensitive artifact/file evidence to compare replicas.
- Cache reuse is revalidated against the selected digest list, and a bad cached trio is removed before retry.
- Retries do not silently repeat an identical bad object without identifying the source location.
- Tests cover a corrupt cache, a corrupt first location followed by a good alternate, and all locations serving the same wrong bytes.

Related: #2548 (Mithril v2), #3145 (snapshot metadata/cache path checks), #3303 (safe extraction/download handles).

wolf31o2:
## #3946 — snapshot identity not pinned across interrupted imports

`mithril.Sync` resolved the target through `GetLatestCardanoDatabaseSnapshot`
(v1: `GetLatestSnapshot`) on every run, including a resume. After an
interruption where the aggregator advanced, the resume imported the newer
snapshot's ledger state over the previous artifact's partially imported rows.
The bootstrap path imports with `ImportConfig.Reconcile` off and every metadata
import phase is insert-if-absent (`Store.ImportUtxos` → `CreateUtxoIfAbsent`),
so nothing removed the first snapshot's rows: the database kept the union of
two snapshots' live sets — UTxOs spent between the artifacts still unspent, and
accounts, pools and DReps the newer snapshot dropped still active. Ledger-state
phase checkpoints did not detect it, being keyed `"{digest}:{slot}"`: a
different artifact reads as a fresh import.

Fix: `Sync` records the selected artifact in `sync_state`
(`mithril_pinned_artifact`: backend, network, digest, beacon, certificate hash,
certified tip slot). It is written from the new
`BootstrapConfig.OnArtifactSelected`, which both backends invoke after the
artifact is identity-checked and, when enabled, certificate-verified, and
before the first download — so the artifact-keyed download cache is pinned too.
A resuming run passes it as `BootstrapConfig.PinnedDigest`, which resolves that
exact artifact instead of the latest. The pin is ephemeral: `ClearSyncState` on
completion wipes it.

Fail-closed cases, each with a recovery instruction: no pin on a non-catch-up
resume, a pin for another backend or network, an artifact the aggregator no
longer serves or serves under a moved beacon, and a reused extraction cache
whose certified ImmutableDB tip differs from the recorded one. An interrupted
catch-up without a pin still proceeds: catch-up imports with `Reconcile` on, so
its reconcile pass removes the interrupted artifact's rows. A fresh bootstrap
still takes the latest artifact.

## #3947 — certified digest mismatch across archive replicas

Not a verification change. The digest list is verified against the
certificate's merkle root before any archive is fetched, and every mismatch is
still refused. The previous error was
`immutable archive 05471: digest mismatch for 05471.chunk: computed X, expected Y`
— no artifact identity, no beacon, no location, and no way to tell a stale
replica from a mis-published archive or a stale local cache.

`downloadImmutables` now returns `ImmutableArchiveError` carrying the artifact
hash and beacon, the immutable file number, and one `ImmutableArchiveAttempt`
per source: the local extraction cache (recorded separately, and its rejected
trio removed before refetch) and each published location by redacted URI.
Location URIs are reduced to scheme, host and path — cloud-storage locations
are pre-signed and carry credentials in the query string. When every published
location was rejected for a digest mismatch on the same file with the same
observed digest, the message states that the archive as published disagrees
with its own certificate rather than one replica being stale. `checkImmutableTrio`
returns a typed `DigestMismatchError`; per-location warnings gain
`location_uri`, `immutable_file`, `expected_digest` and `observed_digest`.

## Tests

- `TestSyncResumesPinnedArtifactAfterAggregatorAdvance` drives `mithril.Sync`
  against a database seeded exactly as an interrupted import leaves it
  (`sync_status` set, artifact pinned, a ledger-state phase checkpoint under
  that artifact's import key) with a newer artifact published. Asserts the
  pinned artifact was imported, its checkpoint advanced to `tip`, and no
  checkpoint exists under the newer artifact's key.
- `TestSyncFreshBootstrapSelectsLatestArtifact` — a clean run still takes the
  latest artifact and leaves no pin.
- `TestSyncRefusesResumeWithoutArtifactPin`,
  `TestSyncRefusesResumeWhenPinnedArtifactIsGone`,
  `TestSyncRefusesResumeWhenCertifiedTipMoved`.
- `TestBootstrapV2ReportsIdenticalBadBytesAcrossLocations`,
  `TestBootstrapV2ReportsSingleLocationMismatchWithoutBlamingReplicas`,
  `TestBootstrapV2AttributesACorruptCacheToTheCache`,
  `TestRedactLocationURI`.
- Existing `TestBootstrapV2ImmutableDigestMismatch` (mis-certified archive
  still refused) and `TestBootstrapV2ImmutableValidationFallsBackToMirror`
  (corrupt first location, good alternate) unchanged and passing.

`ARCHITECTURE.md` and `DATABASE.md` document the pin and the diagnosis.

Fixes #3946
Fixes #3947

---
## Summary by cubic
Fixes interrupted Mithril syncs resuming against a newer artifact than the one whose partial rows were left behind, and makes immutable digest mismatch errors name the artifact, beacon, file, and every source tried without leaking credentials.

**Artifact pinning**
- `Sync` records the selected artifact in `sync_state` from `BootstrapConfig.OnArtifactSelected`, before the first download.
- A resuming run passes the pin as `PinnedDigest` and resolves that exact artifact instead of the latest.
- The pin is cleared on completion, so its presence means a run is mid-flight.
- A resume aborts with a recovery instruction when the pin is missing, names another backend or network, the aggregator no longer serves it, moved its beacon or changed its certificate, or the reused extraction cache yields a different certified tip.
- Pins recorded without backend, network, or certificate fields are still honored.
- An interrupted catch-up without a pin still proceeds, since its reconcile pass removes the interrupted artifact's rows.
- A fresh bootstrap still takes the latest artifact.

**Digest mismatch diagnosis**
- Digest verification is unchanged: every mismatch is still refused.
- `downloadImmutables` now returns an `ImmutableArchiveError` carrying the artifact hash, beacon, immutable file number, and one attempt per source.
- Location URIs are redacted to scheme, host and path, since cloud-storage locations are pre-signed and carry credentials in the query string.
- Rejected local cache trios are attributed to the cache and removed before refetch.
- When every published location served the same bad bytes, the error states the archive disagrees with its own certificate instead of implying a stale replica.

Fixes #3946
Fixes #3947

<sup>Written for commit cc5147d5a7a4b3494eeb0719cf14752f2e8db376. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3948?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Interrupted imports now resume using the same verified snapshot selected before the interruption, rather than silently switching to a newer snapshot.
  - Resumption stops with recovery guidance when the original snapshot is unavailable or verification details no longer match.
  - Snapshot download failures now provide clearer diagnostics, including source-specific failures and integrity mismatches.
  - Sensitive credentials are excluded from download error details.

- **Documentation**
  - Updated architecture and database documentation to describe snapshot pinning and resume behavior.


arepala-uml (approved):
Reviewed this data-integrity fix for interrupted Mithril imports. Confirmed pin persistence is real (durable sync_state DB row, not in-memory) — tests fully close and reopen the database handle between seeding and resume, genuinely simulating a process restart.

- Mismatch detection is fail-closed: backend/network, digest, beacon, and certificate-hash are all re-validated on resume, and any divergence refuses with an explicit "delete and re-bootstrap" instruction rather than attempting to merge incompatible partial data.
- v1 and v2 bootstrap paths wire the pin identically — no coverage gap, verified by diffing both against the merge-base.
- New tests can't even compile against the pre-fix commit (reference symbols that don't exist there) — the strongest possible proof they exercise new code.
- Three earlier bot findings (CodeRabbit + Cubic) were all verified independently from the current source to be already fixed, rather than trusting the bot threads.
- `go build`/`vet`/`test -race`/`golangci-lint` all clean. Both ARCHITECTURE.md and DATABASE.md updated accurately and appropriately scoped.

PR #3949, opened 2026-09-05
docs(ledger): record why apply is not wired into ApplyFunc

### LINKED ISSUE #3227: ledger: pipeline apply stage (#1894 phase 2) not wired into gouroboros ApplyFunc — needs its own design
## Summary

Issue #1894's phase 2 ("Wire apply stage — chain add, error handling, event
publishing, sync progress logging") is not implemented as literally
described, and this issue tracks closing that gap deliberately, as its own
scoped decision.

## Current state (as of #1894 phases 1 and 3)

`ledger.LedgerState`'s `blockPipeline` (`github.com/blinklabs-io/gouroboros/pipeline.BlockPipeline`)
runs decode and VRF/KES validation in parallel worker pools. Its own
`ApplyStage`/`ApplyFunc` is left `nil` — the pipeline's apply stage only
re-sequences decoded/validated `BlockItem`s back into submission order. The
real ledger apply (`chain.AddBlock`, error handling, event publishing, sync
progress logging) still runs entirely in the pre-existing serial path
(`ledgerProcessBlocksFromSource` / `ledgerProcessBlock` in `ledger/state.go`),
fed by the pipeline's re-sequenced output. This does not block the
throughput goal: decode and validation are the parallelizable, CPU-heavy
work; ledger-state application is inherently sequential regardless of which
abstraction runs it, since each block's state depends on the previous one.

## Why it wasn't wired

`gouroboros/pipeline.ApplyFunc` (`func(*BlockItem) error`) is called one
block at a time, by a single background goroutine
(`pipeline.ApplyStageRunner`), with no batching and no mechanism to signal
"restart the whole pipeline" back to a caller — a failing item only sets
that item's own `ApplyError`.

Dingo's real apply logic in `ledgerProcessBlocksFromSource` is a stateful,
whole-batch loop: it opens a `SubmitAsyncDBTxn` transaction per batch,
accumulates a `deltaBatch` across many blocks, can end one transaction and
start another mid-batch on epoch rollover, and its retry loop
(`ledgerProcessBlocks`) depends on sentinel errors (`errRestartLedgerPipeline`,
`errStaleChainIterator`) propagating synchronously back up through
`runLedgerReadChainAttempt`'s wait-for-reader-goroutine-exit invariant
(`ledger/state.go`, `TestLedgerProcessBlocksRetryDoesNotMixBlocksAcrossAttempts`
in `ledger/read_chain_pipeline_test.go`). That shape doesn't fit
`ApplyFunc`'s one-block, no-restart-signal contract without a
disproportionate redesign of either side.

## What closing this would require (pick one, scope it as its own change)

1. **Extend `gouroboros/pipeline`** with a batch-oriented apply contract, or
   a way for `ApplyFunc` to signal "restart"/"abort batch" back to the
   caller — a cross-repo API change reviewed on its own in gouroboros.
2. **Redesign dingo's apply loop** to be per-block and restart-tolerant, so
   it can live inside today's `ApplyFunc` shape — a large, security-
   sensitive rewrite of consensus-critical code (transaction boundaries,
   epoch rollover, retry semantics), not a small wiring change.

Either is a legitimate path, but each is a bigger, separately-reviewable
decision than "wire the apply stage" implies, hence this follow-up rather
than folding it into #1894's other phases.

## Related

- Parent: #1894
- Decode-only pipeline: #3178 (merged)
- VRF/KES validate stage + admission-time duplicate-check removal:
  branch `feat/1894-pipeline-validate-phase3` (not yet merged as of filing)

wolf31o2:
Closes #3227.

`ledger.LedgerState`'s `blockPipeline` leaves `ApplyStage`/`ApplyFunc` nil.
This records that as a decision and pins the upstream properties it rests
on, rather than wiring apply.

## Why not

`gouroboros@v0.202.8/pipeline.ApplyStage` calls `ApplyFunc` on one
goroutine in strict `SequenceNumber` order, but:

1. It does not stop on failure. `maybeApply` records a failing item's error
   on the item and applies every following block anyway. There is no halt
   option. Block N+1's ledger apply reads state block N was supposed to
   write.
2. A skipped item still consumes its sequence slot. An item that failed
   decode or validation is not applied and the next block is.
   `decodeReadChainBatchWithError` instead discards the whole batch, so a
   partially-decoded batch is never handed downstream.
3. The failure never reaches the submitter synchronously. It reaches
   `BlockItem.ApplyError()` and `BlockPipeline.Errors()`, which
   `drainBlockPipelineErrors` reads for logs and counters.
   `errRestartLedgerPipeline` and `errStaleChainIterator` must return
   synchronously to `ledgerProcessBlocks`, which waits for the attempt's
   reader goroutine to exit before the next attempt starts
   (`TestLedgerProcessBlocksRetryDoesNotMixBlocksAcrossAttempts`).
4. `ApplyFunc`'s `func(*BlockItem) error` shape has no batch boundary.
   Apply runs one `submitBlockApplyDBTxn` per 50-block chunk with
   `LedgerDeltaBatch` accumulation, and an epoch or era boundary ends the
   chunk early and runs its rollover in a separate `SubmitAsyncDBTxn`.
5. `chain.AddBlock` is upstream of the pipeline. Blocks enter `chain.Chain`
   from the blockfetch handler; `ledgerReadChainIterator` reads them back
   out and is the pipeline's only submitter, so phase 2's "chain add"
   cannot move into `ApplyFunc`.

Throughput is unaffected either way: decode and VRF/KES validation are the
parallelizable work and are already wired (phases 1 and 3), and
`ApplyStage` is itself single-goroutine and strictly ordered.

## Ordering risk (#3718, #3771/#3840)

Both constrain the upstream path, so leaving apply where it is touches
neither. `drainBlockPipelineBeforeRollback` bounds `WaitForDrain` at
`BlockPipelineRollbackDrainTimeout` (5s) and proceeds with the rollback on
timeout. That is safe only because a nil `ApplyFunc` means the barrier
covers decode and re-sequencing while apply stays behind the reader's
in-order `readChainResult` handshake. `WaitForDrain` does cover
`ApplyFunc`, so wiring apply would make that a bounded wait on database
work, and expiring it would let a rollback run while blocks were still
being applied. Ledger event publication and sync-progress logging would
also move onto the pipeline goroutine, outside the ordering #3718
established. `Submit` takes only `(blockType, rawCbor, tip)`, so the
in-band rollback markers `ledgerReadChainIterator` forwards in order with
blocks are not representable in the pipeline's sequence at all.

## Changes

- `ARCHITECTURE.md`: new subsection "Why dingo's ledger apply is not wired
  into `pipeline.ApplyFunc`" at the end of the Block Processing Pipeline
  section, replacing the "deferred to #3227" phrasing in two places.
- `ledger/state.go`: expanded comment at the nil `ApplyFunc` construction
  site.
- `ledger/block_pipeline_apply_contract_test.go`: four contract tests
  pinning facts 1-3 plus `WaitForDrain` covering `ApplyFunc`, so a
  gouroboros bump that changes them fails CI instead of silently
  invalidating the decision.

The tests were confirmed non-vacuous against a scratch copy of the
upstream package patched to halt on apply error and on a skipped item:
`TestProofApplyErrorDoesNotStopLaterBlocks` dropped from 8 applies to 3
and `TestProofDecodeFailureSkipsOnlyItsOwnBlock` from `[0 2]` to `[0]`.
Unpatched, the copy behaved identically to the pinned dependency.

## Validation

| check | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `gofmt -s -l .` | clean |
| `go test ./ledger/...` | 0 |
| `go test -race ./ledger/... -timeout 90m` | 0, no races |
| `go test ./internal/architecture` | 0 |
| `go test ./internal/test/conformance/` | 0 |
| `make docs-parity` | 0 |

---
## Summary by cubic
Closes #3227: documents why ledger apply is deliberately not wired into `pipeline.ApplyFunc` — a recorded decision, not deferred work — and pins the upstream behavior it rests on with contract tests. No runtime behavior changes.

**The recorded decision**
- `ApplyStage` keeps applying later blocks after a failure, and errors never reach the submitter synchronously, so the retry sentinels can't propagate through it.
- A decode-failed block still consumes its sequence slot, so apply would replace dingo's whole-batch discard with apply-the-rest.
- `ApplyFunc` has no batch boundary for the per-chunk transaction and epoch rollover, and `chain.AddBlock` runs upstream of the pipeline's only submitter.
- `WaitForDrain` covers `ApplyFunc`, so wiring apply would make the 5s rollback barrier a bounded wait on database work, reopening #3771/#3840.
- `ARCHITECTURE.md` and the `state.go` comment now call the nil `ApplyFunc` a decision; the contract tests also pin that apply runs single-goroutine in submission order, so a `gouroboros` bump that changes any pinned property fails CI instead of silently invalidating it.

<sup>Written for commit 431192d3335f867e8bf68695228bbcd1c218b516. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3949?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


arepala-uml (approved):
Reviewed despite the "docs" title as a real code-adjacent change. Confirmed `ledger/state.go`'s diff is comment-only — zero runtime behavior change. The substance is a new test file that pins upstream `gouroboros/pipeline.ApplyStage` behavior at the current pinned dependency version, not dingo's own code.

- Each of the 4 new tests asserts a real, independently falsifiable upstream property (serial/in-order apply, error-doesn't-halt-later-blocks, skip-still-consumes-sequence-slot, WaitForDrain covers in-flight apply work) — a legitimate "canary" pattern: if a future gouroboros bump breaks any assumption, the test fails and forces a conscious re-review of this architectural decision.
- The underlying reasoning holds up and is corroborated by the linked issue #3227, filed independently by a third-party contributor with matching technical reasoning.
- Cubic's one finding (flaky time.Sleep-based test sync) was already fixed — confirmed the merged version uses proper channel synchronization.
- `go build`/`vet`/`test -race`/`golangci-lint` clean. ARCHITECTURE.md's new section is genuinely clear, with an explicit "what would reopen this decision" reversal condition.

Issue #2872, opened 2026-07-18
Dedicated `/healthz` and `/readyz` endpoints


wolf31o2:
#### Motivation (dingo-operator)

Kubernetes wants distinct liveness and readiness probes. The operator gates Service endpoints and
active/standby traffic on readiness — a node that is still doing Mithril bootstrap or initial sync,
or a block producer whose opcert/KES is invalid, must not be marked Ready. Today liveness is inferred
indirectly from metrics reachability and/or the NtC socket, which conflates "process alive" with
"caught up and forging-ready."

#### Proposal

Add first-class `/healthz` (liveness) and `/readyz` (readiness) HTTP endpoints on the node's
auxiliary listener.

Current state (verified):

- The node's auxiliary listeners are metrics-only and pprof-only: `metricsMux := http.NewServeMux()`
  with `metricsMux.Handle("/metrics", promhttp.Handler())` (`internal/node/node.go:457-458`), and a
  pprof `debugMux` registering `/debug/pprof/*` when `DebugPort != 0`
  (`internal/node/node.go:480-485`), both served via `serveAuxiliaryListener(...)`
  (`internal/node/node.go:121-134`).
- There is no `/healthz`, `/readyz`, `/livez`, `/live`, or `/ready` anywhere in the repo. The only
  health endpoint that exists is `GET /health` on the **separate Blockfrost API server**
  (`api/blockfrost/blockfrost.go:70`, handler `handleHealth` at `api/blockfrost/handlers.go:112`,
  returning `{ is_healthy bool }` from `api/blockfrost/types.go:25`). That endpoint is not on the
  node's aux listener, is not a Kubernetes-style probe, and is only present in `api` storage mode.

Proposed change:

1. Register `/healthz` on the metrics aux mux (`internal/node/node.go:457`) or a dedicated health
   listener. **Liveness** = process up and the main event loop responsive (not deadlocked). Cheap,
   dependency-free.
2. Register `/readyz`. **Readiness** = DB open + chain caught up to tip within a configurable slot
   tolerance; for block producers additionally: valid opcert (not expired/future), KES period within
   the opcert window, and forger initialized. `/readyz` must return not-ready during Mithril
   bootstrap and initial sync so the operator holds traffic off the pod.
3. Both available in `core` storage mode (no dependency on the Blockfrost API server).

#### Acceptance criteria

- [ ] `/healthz` returns 200 whenever the process is alive and the event loop is responsive,
      independent of sync state.
[...]

PR #3951, opened 2026-09-05
feat(health): add /health, /healthz and /readyz probe listener

### LINKED ISSUE #1484: Docker: add HEALTHCHECK and /health endpoint
## Summary

The Dockerfile has no HEALTHCHECK instruction. Container orchestrators (Docker Compose, Kubernetes, ECS) cannot detect when Dingo is unhealthy.

## Current State

- Dockerfile exists with no HEALTHCHECK
- No dedicated health endpoint outside the blockfrost API
- DevNet compose works but has no health checks

## What Needs to Happen

1. Add a lightweight HTTP /health endpoint on a dedicated port (separate from blockfrost/utxorpc) that returns 200 when the node is running.
2. Add a HEALTHCHECK instruction to the Dockerfile that hits the health endpoint.
3. The health endpoint should work regardless of whether blockfrost or utxorpc APIs are enabled.

## Current state (2026-08-21 audit)

The Dockerfile and Compose file now have health checks, but both probe `/metrics`; they prove listener reachability only. Blockfrost `/health` also returns success without checking sync or database state. First-class liveness/readiness semantics are tracked more completely by #2872. Coordinate this issue with #2872, and switch container probes only after those semantics exist.

wolf31o2:
Fixes #1484. Implements the liveness/readiness split specified in #2872.

## Change

A dedicated probe listener on `healthPort` (default `12799`), separate from the metrics listener, the pprof listener, and all three API listeners.

| Path | Semantics |
|------|-----------|
| `/healthz`, `/health` | Liveness: 200 whenever the process is up and the listener answered. Independent of sync state. |
| `/readyz` | Readiness: 200 only when the chain tip is within `healthReadyGapSlots` (default `1000`) of the wall-clock slot; 503 while starting, bootstrapping, catching up, or when the tip has frozen. |

Both return JSON carrying the readiness verdict, a reason, and the observed tip gap, so a liveness probe's body already says why a live node is not serving:

```json
{"live":true,"ready":false,"reason":"tip gap 4211 slots exceeds tolerance of 1000 slots","tipGapSlots":4211,"status":"unhealthy"}
```

The listener serves only those three paths; every other path is 404, so metrics and pprof cannot appear on a port exposed for probing. Non-GET/HEAD is 405.

The `Dockerfile` `HEALTHCHECK` and the `docker-compose.yml` healthcheck move from `/metrics` to `/health`. `/metrics` proved only that the metrics listener had bound.

## Why liveness and readiness are separate

An orchestrator restarts on a liveness failure and drains on a readiness failure. A node in initial sync is legitimately hours or days behind, and the conditions that freeze a tip — a stalled blockfetch (#3771), a validation rejection loop (#3935) — are not repaired by a restart. Folding sync state into liveness therefore produces a replacement loop a syncing node can never escape. Liveness stays independent of sync state and is what the image `HEALTHCHECK` probes; readiness is the signal that catches a frozen tip, and failing it drains traffic without killing the node.

`healthReadyGapSlots` defaults to 1000, matching `forgeStaleGapThresholdSlots`: both answer "has this node stopped following the chain?", and a probe that flapped more readily than the forger's own staleness gate would evict a node the forger still considers current. At `f=0.05` an empty 1000-slot stretch is not something a live chain produces.

## Storage mode

`node.go`'s three API listeners start only under `storageMode.IsAPI()`. The shipped `docker-compose.yml` runs the default `core` mode, so a probe gated the same way would be inert in exactly the configuration the image ships with. The health listener is not storage-gated and is unaffected by whether Blockfrost, Mesh, or UTxO RPC are configured.

## Bind address

The health listener uses the common `bindAddr`, as do the relay, metrics and public API listeners. Listener addresses use `net.JoinHostPort`, including IPv6 addresses. No separate API bind address or authentication requirement is introduced.

## Tip gap source

`ledger.LedgerStateConfig.ReportTipGapFunc` is called from the slot-tick loop with the same value published as `dingo_tip_gap_slots`, so readiness does not depend on the Prometheus listener. `ledgerStateConfig` is the single construction site `Run` and a live Restore/Truncate rebuild share, and it closes over the node, so a rebuilt ledger reports into the same atomics.

The gap is reported as unknown, not zero, until the first tick: a node that has opened its database but has not begun following the chain reports not-ready rather than reading as perfectly caught up. `closeStorageForLiveLifecycleOp` clears it, so `/readyz` does not answer 200 from a pre-restore reading for the length of a live rebuild.

Listener addresses in `internal/node/node.go` are built with `net.JoinHostPort` rather than `"%s:%d"`. The health listener, and the pre-existing relay, private/NtC, and metrics sites, all produced an unbindable address for an IPv6 `bindAddr` such as `::`. Output is byte-identical for IPv4.

## Configuration

- `healthPort` / `DINGO_HEALTH_PORT` / `--health-port`, default `12799`, `0` disables the listener. Participates in port-collision validation in both storage modes.
- `healthReadyGapSlots` / `DINGO_HEALTH_READY_GAP_SLOTS` / `--health-ready-gap-slots`, default `1000`.

## Tests

`internal/node/node_health_test.go` starts the listener the way `Run` does — `newHealthServer` plus `serveAuxiliaryListener` — and issues HTTP requests against the bound socket. The core-mode case builds a `*dingo.Node` through `buildDingoConfig`/`dingo.New` with all three API plugin ports at 0.

Covered: probes answer with both APIs disabled in `core` storage mode; `/readyz` is 503 with no tip, with a 5000-slot gap, and during a 90M-slot initial sync, while `/healthz` stays 200 in all three; `/readyz` is 200 within tolerance; `healthPort: 0` builds no listener; an IPv6 `bindAddr` produces `[::]:12799` and binds; the health port collides in `core` mode, unlike the API ports; a live storage teardown leaves the tip gap unknown rather than stale; `handleSlotTicks` reports the same value it writes to the gauge, including the tip-ahead-of-clock case, and tolerates a nil reporter.

## Docs

`README.md` gains a Health Probes section with a Kubernetes probe snippet and a port-table entry; `ARCHITECTURE.md` documents the listener under External Interfaces and in the directory tree; `dingo.yaml.example` documents both settings.

## Summary by CodeRabbit

- **New Features**
  - Added dedicated liveness and readiness health endpoints: `/health`, `/healthz`, and `/readyz`.
  - Health checks report synchronization status, tip-gap details, and appropriate HTTP status codes.
  - Added configurable health listener port and readiness tolerance through configuration, CLI flags, and environment variables.
  - Added Docker and Compose health checks using the dedicated liveness endpoint.
  - Health monitoring now correctly resets during storage rebuilds and lifecycle transitions.

- **Documentation**
  - Documented health endpoints, configuration options, responses, and deployment probe settings.


internal/node/node.go
```diff
@@ -156,6 +173,45 @@ func newPprofDebugServer(cfg *config.Config) *http.Server {
...
+	if cfg.HealthPort == 0 {
+		return nil
+	}
+	readyTipGapSlots := uint64(cfg.HealthReadyGapSlots)
+	if readyTipGapSlots == 0 {
+		readyTipGapSlots = config.DefaultHealthReadyGapSlots
+	}
+	return &http.Server{
+		Addr: fmt.Sprintf(
+			"%s:%d",
```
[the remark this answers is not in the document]
wolf31o2: Valid, fixed in 4a5e8261. `newHealthServer` builds its address with `net.JoinHostPort`, so `bindAddr: "::"` yields `[::]:12799` instead of an unbindable `:::12799`.

The same `"%s:%d"` construction appeared at three more sites in `Run`, with the same defect: the relay listener, the private/NtC listener, and the metrics listener. All four now use `net.JoinHostPort`. Output is byte-identical for IPv4, so only previously-unbindable IPv6 configurations change.

`TestHealthServerBracketsIPv6BindAddr` pins the address form for `0.0.0.0`, `127.0.0.1`, `::`, and `::1`; `TestHealthListenerBindsIPv6Loopback` starts the listener on `[::1]` and requests `/readyz`.

node.go
```diff
@@ -169,6 +169,10 @@ type Node struct {
 	// own documented "keeps syncing normally" behavior.
 	snapshotMu sync.Mutex
 
+	// health carries the sync signals the readiness probe reads. See
+	// node_health.go; it survives a live database restore/truncate rebuild.
+	health nodeHealth
```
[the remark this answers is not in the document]
wolf31o2: Valid, fixed in 4a5e8261. `closeStorageForLiveLifecycleOp` nils `n.ledgerState` (node_lifecycle.go:508) and the rebuilt ledger does not tick until it catches up, while `tipGapKnown` only ever went true, so `/readyz` answered 200 from the pre-restore reading for the length of the rebuild.

`nodeHealth.forgetTipGap` clears both atomics and is called unconditionally at the top of `closeStorageForLiveLifecycleOp`: entering that function means storage is being torn down, so the gap is untrustworthy from that point, including on the fail-closed `errStorageDrainUnconfirmed` path where `n.ledgerState` is deliberately left non-nil.

`TestLiveLifecycleTeardownForgetsTipGap` reports a gap through `ledgerStateConfig().ReportTipGapFunc`, calls `closeStorageForLiveLifecycleOp`, asserts the gap is unknown rather than stale, and asserts a rebuilt config restores it.

node_lifecycle.go
```diff
@@ -491,6 +491,12 @@ func (n *Node) quiesceForLiveLifecycleOp(ctx context.Context) error {
 func (n *Node) closeStorageForLiveLifecycleOp(ctx context.Context) error {
 	var err error
 
+	// Storage is going away, so the ledger that feeds the readiness probe
+	// stops ticking here and does not resume until the rebuilt one reaches
+	// its first tick. Drop the last reported tip gap rather than let
+	// /readyz keep answering 200 from it for the length of the rebuild.
+	n.health.forgetTipGap()
```
chrisguiney, line 498: Blocker: this clear is undone by the next slot tick, so `/readyz` still answers 200 for the length of a live rebuild — the behavior 4a5e826's message says it fixes.

`quiesceForLiveLifecycleOp` must have returned before this runs, but it does not stop the ledger's slot clock. It stops the API capabilities, the Midnight server, history expiry, the off-chain metadata fetcher, token registry sync, the mempool, `connManager` and the Leios persist writer. The clock is stopped inside `n.ledgerState.Close()` on the next line, at `ledger/state.go:2367` — after `Scheduler.Stop`, the consumed-UTxO drain, the block-pipeline stop, the block-processing drain (`CloseProcessBlocksDrainTimeout`, 20s), the block-pipeline drain (10s), the blockfetch-continuation drain (10s), seven `UnsubscribeAndWait` calls, and the unconditional header-replay and reward-precompute waits.

`handleSlotTicks` (`ledger/state.go:2585`) has no `closed` check and calls `ReportTipGapFunc` at `ledger/state.go:2542` at the top of every tick. The first tick after this line — within one slot, ~1s on preview/mainnet — therefore calls `recordTipGap` again with the pre-teardown gap and sets `tipGapKnown` back to true, where it stays for the rest of the rebuild.

`TestLiveLifecycleTeardownForgetsTipGap` cannot observe this: it leaves `n.ledgerState` nil, so `Close` never runs and no tick can land in the window. Checked directly against this head — `report(3); forgetTipGap(); report(3)` leaves `TipGapSlots() == (3, true)`.

Moving the call after `Close()` returns narrows the window but does not close it: `SlotClock.Stop` closes the 1-buffered subscriber channel and then waits only on its own run loop, so `handleSlotTicks` can still drain one buffered tick after `Close` has returned. A generation counter on `nodeHealth`, bumped at teardown and checked in `recordTipGap`, closes it without depending on shutdown ordering.

node_health.go
```diff
@@ -0,0 +1,70 @@
...
+// directly so a health probe does not depend on the Prometheus listener.
+func (n *Node) TipGapSlots() (uint64, bool) {
+	if n == nil {
+		return 0, false
+	}
+	if !n.health.tipGapKnown.Load() {
+		return 0, false
+	}
+	return n.health.tipGapSlots.Load(), true
+}
```
chrisguiney, line 99: Non-blocking, and the general form of the blocking finding on `node_lifecycle.go`: the readiness reading has no freshness bound. `tipGapKnown` goes true on the first tick and, `forgetTipGap` aside, never goes false again, so the answer is only ever as good as the last thing that called `recordTipGap`.

That is sufficient for the wedges the description names. A stalled blockfetch or a validation rejection loop leaves `handleSlotTicks` running, `tick.Slot` keeps advancing past a frozen `tipSlot`, and the gap grows past tolerance — confirmed by `TestHandleSlotTicksReportsTipGap`. It is not sufficient when the tick consumer itself stops: `loadStateSnapshots`, `emitNextEpochNonceReady` and `EventBus.Publish` are all inside that loop, and the live-rebuild window is one instance of the same shape. In each case `/readyz` keeps answering 200 from a reading that has stopped being updated, which is the one answer a readiness probe must not give by default.

Storing a monotonic timestamp beside the gap and treating a reading older than a few slot durations as unknown would make `/readyz` fail closed for the whole class, rather than only for the cases where the gap happens to keep growing on its own.

chrisguiney (changes requested):
Reviewed 4a5e826 in an isolated worktree and merged onto current `origin/main` (39 commits ahead of the merge base); the merge is clean and the merged tree is green.

Bot standing on this head: CodeRabbit did **not** review — its check is green but the comment body is "Review limit reached / You've used all free OSS reviews for now". Cubic's review is anchored to `66bac72`, the first commit, not this head. No human review has been submitted.

## Blocking

`node_lifecycle.go` — the tip-gap clear added in 4a5e826 does not hold on the path it targets. Inline.

## Non-blocking

`node_health.go` — the readiness reading has no freshness bound. Inline.

## Verified

- Liveness and readiness are genuinely separate. `/healthz` and `/health` stay 200 with a 5000-slot gap and during a 90M-slot sync; `/readyz` is 503 with no tip, past tolerance, and during sync. Making the liveness handler `requireReady` fails four tests.
- Probe handlers take no lock and read no database or ledger pointer — two atomic loads. They cannot block under load.
- Not storage-gated: adding a `storageMode != "api"` gate fails `TestHealthListenerServesInCoreModeWithAPIsDisabled`.
- An unknown tip gap reports not-ready rather than ready-at-zero; making it default to ready fails three tests.
- Bind default is `bindAddr` (0.0.0.0), matching the relay and metrics listeners rather than #3929's `apiBindAddr` loopback default. That is consistent with #3929, which moves only the API listeners. The body discloses `live`, `ready`, `reason` and `tipGapSlots`; none of that is absent from `:12798`, which binds the same address by default and is always started in a serving mode.
- Lifecycle: nil at `healthPort: 0`, bind failure logged and non-fatal through `serveAuxiliaryListener` as for metrics and pprof, shut down on all three exit paths via `optionalShutdown`, and present in the port-collision table for both storage modes.
- Listener class audit complete. metrics `bindAddr`/on, pprof `debugBindAddr`/off, health `bindAddr`/on, the three API listeners and Bark on their own hosts and off by default, plus the Mithril-sync metrics listener. The `"%s:%d"` IPv6 defect Cubic found was fixed at all four sites in `internal/node/node.go`, and no listen address in the repository is still built with `%s:%d` — `cmd/koios-parity`'s is a MySQL DSN, not a listen address.
- Every new test fails without its fix: eight in-place reverts, all red. No existing assertion was weakened; the edits to `node_test.go`, `config_test.go` and `validate_test.go` are signature and defaults updates only.

## Checks

Run on the merged tree unless noted: `go build ./...` 0, `go vet ./...` 0, `make lint` (import-boundaries, golangci-lint on all three modules, `GOOS=windows`) 0 issues, `make docs-parity` 0, `go test ./internal/health ./internal/config ./internal/node` 0, `go test .` 0, `go test ./ledger -run HandleSlotTicks` 0. Both nested modules (`examples/dingo-gov-lens`, `internal/test/antithesis`) build and vet clean. `nilaway` was not installed locally, so that stage of `make lint` did not run. `gofmt -l` reports `bark/blob_test.go` and `ledger/read_chain_pipeline_test.go`, both unformatted on `origin/main` as well.

Not run: DevNet, conformance, and `make test-live-lifecycle` — the last is the suite that would exercise the path the blocking finding is on.

Scope note, not a finding: #2872's block-producer criterion (`/readyz` non-200 on an expired or future opcert, or an out-of-window KES period) is not implemented here. The PR closes #1484 and implements the liveness/readiness split only, which is what its description claims.

internal/node/node.go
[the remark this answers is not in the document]
chrisguiney: Confirmed fixed on 4a5e826, and the class is complete. `newHealthServer` yields `[::]:12799` and `[::1]:12799`, `TestHealthListenerBindsIPv6Loopback` binds and answers on `[::1]`, and reverting `net.JoinHostPort` back to `"%s:%d"` in place makes both that test and `TestHealthServerBracketsIPv6BindAddr` fail (`":::12799"`, `"::1:12799"`).

The three pre-existing sites in `Run` — relay, private/NtC and metrics — were converted with it, and no listen address anywhere else in the repository is still built with `%s:%d`. `cmd/koios-parity/main.go:301` is the remaining `%s:%d` match but it is a MySQL DSN address, not a listen address.

node.go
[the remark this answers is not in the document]
chrisguiney: Valid, and it is still open on 4a5e826. The finding is right; the fix is placed before the thing that stops the reporter, so it does not hold.

`n.health.forgetTipGap()` runs at `node_lifecycle.go:498`, one line before `n.ledgerState.Close()`. `quiesceForLiveLifecycleOp`, which must have returned by then, does not stop the ledger's slot clock — that happens inside `Close`, at `ledger/state.go:2367`, after the bounded process-blocks (20s), block-pipeline (10s) and blockfetch (10s) drains and the unconditional header-replay and reward-precompute waits. `handleSlotTicks` has no `closed` check and calls `ReportTipGapFunc` at `ledger/state.go:2542` at the top of every tick, so the first tick after the clear — within one slot — re-arms `tipGapKnown` with the pre-teardown gap, and `/readyz` answers 200 for the rest of the rebuild.

`TestLiveLifecycleTeardownForgetsTipGap` leaves `n.ledgerState` nil, so `Close` never runs and no tick can land in the window it is meant to cover.

Details and a suggested generation-counter fix are in the inline comment on `node_lifecycle.go` in review 5126632170.

internal/config/config.go
```diff
@@ -1072,6 +1093,8 @@ var globalConfig = &Config{
 	MetricsPort:                         12798,
 	DebugBindAddr:                       DefaultDebugBindAddr,
 	DebugPort:                           0,
+	HealthPort:                          DefaultHealthPort,
```
chrisguiney: Blocker: this is the one file that conflicts with current `main`, and the conflict is semantic rather than mechanical.

`main` replaced `var globalConfig = &Config{...}` with `newDefaultConfig()` returning the same literal, plus `ValidateForgedBlock: true`. Resolving the conflicted hunk by taking `main`'s side — the natural choice, since that side is an entire new function and this side is a fragment of the literal it replaced — drops these two lines.

Dropping them is invisible at every layer that would normally catch it:

- `ApplyDefaults` fills `HealthReadyGapSlots` back in, but there is no equivalent for `HealthPort`. This literal is its only default source.
- `newHealthServer` returns nil at `HealthPort == 0`, so the probe listener is never started.
- The `HEALTHCHECK` this PR adds probes `127.0.0.1:12799/health`, so every container built from the image would go unhealthy after the start period — and Docker, Swarm and ECS answer that by replacing the container.
- `go test ./internal/config/` stays green. `resetGlobalConfig` in `config_test.go` carries its own copy of both lines and seeds `globalConfig` before each test, so the three expected-value literals updated in that file are compared against that seed rather than against `newDefaultConfig`.

Confirmed by resolving the hunk to `main`'s side and building: `--health-port` loses `(default 12799)` from `--help` while `go test ./internal/config/` still passes.

Resolution: merge current `origin/main` and keep both lines inside `newDefaultConfig()`. Worth adding a test that reads `newDefaultConfig().HealthPort` directly — the existing config tests cannot see this loss, which is the same gap `newDefaultConfig`'s own doc comment records for `ValidateForgedBlock`.

node_lifecycle.go
```diff
@@ -593,6 +599,9 @@ func (n *Node) closeStorageForLiveLifecycleOp(ctx context.Context) error {
 // (LedgerState, Mempool, ChainsyncState, ConnManager, PeerGov) once the new
 // objects exist, exactly like Run()'s late-binding setters do.
 func (n *Node) reinitializeCoreStorage(ctx context.Context) error {
+	// The previous ledger's tip-gap observation must not make readiness look
+	// healthy while Restore or Truncate is rebuilding the core storage.
+	n.health.forgetTipGap()
```
chrisguiney, line 604: Non-blocking: this clear is correct and load-bearing, but nothing tests it. Removing the call leaves `go test .` fully green.

It is load-bearing because `reinitializeCoreStorage` is reachable without `closeStorageForLiveLifecycleOp` having run: the quiesce-failure branches of `Restore` and `Truncate` call `reinitializeAndResume` directly (`node_lifecycle.go:1522` and `:1974`). On that path the outgoing ledger was never closed, so the clear at `:498` never ran and its generation was never superseded — without this line the pre-op reading would survive the rebuild, and the outgoing ledger would keep reporting into it alongside the rebuilt one. That is the same window the generation fence closes, entered a different way.

A test that drives the quiesce-failure entry would go red without this line. As it stands the newest commit on the branch, and one of the two addressing the readiness window, ships with no coverage.

node_health_test.go
```diff
@@ -0,0 +1,135 @@
...
+	internalconfig "github.com/blinklabs-io/dingo/internal/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestTipGapUnknownBeforeFirstSlotTick pins the state the readiness probe
+// must refuse on: a node that has opened its database but has not yet
+// followed the chain reports no tip gap at all, rather than a zero gap that
+// would read as "perfectly caught up".
+func TestTipGapUnknownBeforeFirstSlotTick(t *testing.T) {
```
chrisguiney, line 31: Non-blocking: none of the four tests in this file calls `t.Parallel()`.

`AGENTS.md` lists the root package among those where top-level tests call it and a new test should too, and none of these four reaches the process-global state the exception is written for — each builds its own `Node` value and touches nothing package-level. Adding `t.Parallel()` to all four passes under `go test -race -count=4`.

chrisguiney (changes requested):
Re-reviewed `324b2a31`. Three branch commits since the last review: `a0c77c07`, `e5c3bb97`, `324b2a31`.

## Prior findings

**Cleared — the tip-gap clear now holds on the path it targets.** `a0c77c07` turns `recordTipGap` into a generation-checked compare-and-store under `nodeHealth.mu`, and `ledgerStateConfig` captures the generation when it builds that ledger's callbacks. A tick the outgoing ledger had already dequeued therefore carries a superseded generation and is dropped whichever side of `forgetTipGap` it lands on, so the ordering that made the previous placement ineffective no longer decides the outcome. `forgetTipGap`, `recordTipGap` and `currentGeneration` all serialize on the same mutex, so there is no interleaving in which a stale store survives the clear. Reverting the generation check in place fails `TestLiveLifecycleTeardownForgetsTipGap`.

**Still open, and narrower — the readiness reading has no freshness bound.** Two of the three instances are now closed: the live-rebuild window by the generation fence, and graceful shutdown by ordering, since `shutdownNodeResources` stops the health listener before `nodeStop`. What remains is a tick consumer that stalls in place — `emitNextEpochNonceReady` and `EventBus.Publish` are still inside `handleSlotTicks` after the `ReportTipGapFunc` call — leaving `tipGapKnown` true over a reading that has stopped updating. Non-blocking, as before.

## Blocking

`internal/config/config.go` — the branch does not merge, and the natural resolution of its one conflict silently disables the listener this PR adds, with the test suite still green. Inline.

## Pipeline

`go-test (macOS) (1.26.x)` fails at this head on `TestDecodeReadChainBatchDoesNotDeadlockOnManyValidationErrors` (`ledger/block_pipeline_errors_drain_test.go:174`, "no unexpected pipeline errors should have been drained").

It is not this change's. The branch touches no part of the block pipeline, and `7d9ee62e` fixed exactly this assertion on `main` after this branch's last merge of `main`. Measured with `GOMAXPROCS=1 go test ./ledger/ -run TestDecodeReadChainBatchDoesNotDeadlockOnManyValidationErrors -count=40`: 5/40 failures at this head, 0/40 at `origin/main`, same assertion and message as CI. Idle full-package runs pass at both ends, which matches `7d9ee62e`'s note that the flake needs CPU pressure. Only one `go-test` run exists for this SHA, so no later run superseded it. Merging current `origin/main` clears it, and is the same action the conflict requires.

## Verified on the merged tree

Merged current `origin/main` (12 commits ahead) with the conflict resolved as described inline:

| Check | Result |
| --- | --- |
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `go test -race ./internal/health/ ./internal/config/ ./internal/node/ .` | pass |
| `go test -race ./ledger/ -run 'TipGap\|SlotTick'` | pass |
| `golangci-lint run ./internal/... ./ledger/... .` | 0 issues |
| `gofmt -l` over every file this PR changes | clean |

Re-confirmed against the merged tree, since the base moved: `newHealthServer` still builds its address with `net.JoinHostPort`, still binds `bindAddr`, still returns nil only at `healthPort: 0`, and is still not storage-gated. No listen address in the repository is built with `"%s:%d"` — the remaining matches are UTxO keys, a MySQL DSN, and error text.

Scope is unchanged and accurate: #1484's three asks (a dedicated probe endpoint, a `HEALTHCHECK` that uses it, and availability independent of the API storage mode) are all met, and #2872's block-producer criterion — `/readyz` non-200 on an expired or future opcert or an out-of-window KES period — remains unimplemented here, which the description does not claim otherwise.

`make test-live-lifecycle` was not run; it is the suite that would exercise the Restore/Truncate path the generation fence sits on, so that path is covered here by unit tests and by reading its two entry points rather than by a live run.

node.go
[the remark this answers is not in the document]
chrisguiney: Fixed on `324b2a31`, and the fix now covers both entry paths.

`a0c77c07` replaced the placement-dependent clear with a generation fence: `ledgerStateConfig` captures `nodeHealth.generation` when it builds a ledger's callbacks, `forgetTipGap` advances it, and `recordTipGap` compares and stores under the same mutex. A tick the outgoing ledger dequeued before `Close` returned therefore carries a superseded generation and is dropped, so it no longer matters whether it lands before or after the clear — which is what the earlier placement got wrong.

`324b2a31` then adds the clear at the top of `reinitializeCoreStorage`, covering the quiesce-failure entry where `closeStorageForLiveLifecycleOp` never ran at all.

Reverting the generation check in place fails `TestLiveLifecycleTeardownForgetsTipGap`. Reverting `324b2a31`'s clear leaves the suite green — raised separately as missing coverage, not as a defect.

node_lifecycle.go
chrisguiney: Cleared on `324b2a31`.

The objection was that this clear ran one line before `n.ledgerState.Close()`, while `handleSlotTicks` keeps calling `ReportTipGapFunc` throughout that call's drains, so the first tick after the clear re-armed `tipGapKnown` with the pre-teardown gap.

`a0c77c07` removes the dependency on ordering rather than moving the call. `recordTipGap` now takes a generation captured by `ledgerStateConfig` when that ledger's callbacks were built, and compares it against `nodeHealth.generation` under the same mutex that `forgetTipGap` increments it under. The outgoing ledger's closure holds the superseded value for the whole of `Close`, so every one of those ticks is dropped, whether it arrives before or after the clear. There is no interleaving that restores the stale reading, because the compare and the store are one critical section.

Verified by reverting the generation check in place: `TestLiveLifecycleTeardownForgetsTipGap` fails on both the `ok` and the value assertion.

node_health.go
chrisguiney: Still open on `324b2a31`, but narrower than when this was raised. Non-blocking, as before.

Two of the three instances are closed. The live-rebuild window is handled by `a0c77c07`'s generation fence, which drops a superseded ledger's ticks rather than relying on where the clear sits. Graceful shutdown turns out never to have been exposed: `shutdownNodeResources` calls the health listener's shutdown before `nodeStop`, so the probe stops answering before the ledger stops ticking, and there is no window in which `/readyz` serves a frozen reading on the way down.

What remains is the case where the tick consumer stalls in place rather than being replaced. `emitNextEpochNonceReady` and `EventBus.Publish` are still inside `handleSlotTicks` after the `ReportTipGapFunc` call, so a stall in either leaves `tipGapKnown` true over a reading nothing updates and nothing ages out. A timestamp alongside the gap, with `TipGapSlots` reporting unknown past some multiple of the slot length, would bound it without changing the contract.

Separately, on this function: every write to `tipGapSlots` and `tipGapKnown` now happens under `nodeHealth.mu`, and this is their only reader, which also takes it. The `atomic.Uint64` and `atomic.Bool` types no longer provide the synchronization and could be plain fields. Cosmetic.

internal/node/node.go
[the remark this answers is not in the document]
wolf31o2: Unchanged at `a590b021`. `newHealthServer` still builds its address with `net.JoinHostPort`, and `TestHealthServerBracketsIPv6BindAddr` and `TestHealthListenerBindsIPv6Loopback` still pin `[::]:12799` and a live bind on `[::1]`.

A related gap in the same area is now closed: `internal/config.bindAddrsOverlap` compared bind addresses as raw strings, so `::1` and `0:0:0:0:0:0:0:1` on one port passed validation and both listeners raced for the same endpoint. It now compares canonical forms.

node.go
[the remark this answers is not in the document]
wolf31o2: Unchanged at `a590b021`, and the second way into a rebuild is now covered as well.

`reinitializeCoreStorage` clears the gap for the entry the generation fence does not see: the quiesce-failure branches of `Restore` and `Truncate` call `reinitializeAndResume` directly, so `closeStorageForLiveLifecycleOp` never ran and the outgoing ledger was never closed. `TestReinitializeCoreStorageForgetsTipGap` asserts both that the reading is cleared and that the still-open outgoing ledger cannot report back into it.

node_lifecycle.go
wolf31o2: No behavioral change at `a590b021`. The generation fence is as reviewed.

`nodeHealth.tipGapSlots` and `tipGapKnown` are now plain fields rather than `atomic.Uint64`/`atomic.Bool`: `recordTipGap`'s generation compare and its store, `forgetTipGap`'s increment and clear, and `TipGapSlots`' read all run under `nodeHealth.mu`, which is what makes the compare-and-store indivisible. The atomic types added no synchronization on top of that.

node_health.go
wolf31o2: The cosmetic half is done: `tipGapSlots` and `tipGapKnown` are plain fields now, since `nodeHealth.mu` already covers every read and write and is what makes `recordTipGap`'s generation compare-and-store one operation.

The freshness bound is not added here, and the reason is that the threshold has no correct value available at this layer. `ReportTipGapFunc` carries only the gap, and the slot-clock tick interval is derived from the applied ledger's era history rather than being a constant, so `nodeHealth` has no slot length to scale a timeout by. A fixed wall-clock constant is either long enough to be decorative on the 1s-slot networks or short enough to drain a healthy node where slots are longer. Doing it properly means carrying the slot duration across the ledger boundary and adding a clock seam to `nodeHealth`, which is a larger change than the probe wiring here.

Two of the three named tick-loop consumers are narrower than the general shape suggests. `EventBus.Publish` routes non-channel subscribers through `deliverWithTimeout` (`event/event.go`), and a channel subscriber's `Deliver` waits in `deliverWait`, which a stalled-subscriber close can end. `emitNextEpochNonceReady` returns before publishing unless the wall-clock and ledger epochs agree and both `tick.Slot` and `tipSlot` have passed the nonce cutoff. The unbounded remainder is `ls.EpochNonce` and `loadStateSnapshots`, and the latter sits above the `ReportTipGapFunc` call, so a stall there leaves the reading unknown rather than stale.

internal/config/config.go
wolf31o2: Resolved at `a590b021`. `newDefaultConfig` carries both lines -- `HealthPort: DefaultHealthPort` and `HealthReadyGapSlots: DefaultHealthReadyGapSlots` -- and `--health-port` still registers `12799` as its flag default.

The test gap is closed too. `TestHealthPortDefaultsToDefaultHealthPort` (`internal/config/health_defaults_test.go`) asserts both values directly on `newDefaultConfig()` rather than through `resetGlobalConfig`'s copy, and asserts that `ApplyDefaults` refills `HealthReadyGapSlots` but leaves `HealthPort` at zero, so nothing else can be mistaken for a second source. Deleting the `HealthPort` line and running `go test ./internal/config/` fails that test and only that test.

node_lifecycle.go
wolf31o2: Covered at `a590b021` by `TestReinitializeCoreStorageForgetsTipGap` (`node_health_test.go`).

It reports a gap through the outgoing ledger's `ReportTipGapFunc`, calls `reinitializeCoreStorage`, and asserts the reading is unknown rather than stale. It then reports again through that same outgoing config -- which is exactly what the quiesce-failure path leaves running, since the ledger was never closed -- and asserts the superseded generation is dropped. Removing the `forgetTipGap` call fails all four assertions, on both the `ok` result and the value.

node_health_test.go
wolf31o2: `t.Parallel()` added to all four, and to `TestReinitializeCoreStorageForgetsTipGap` added alongside them. Each builds its own `Node` value and touches nothing package-level, so none needs the sequential exception. `go test -race -count=4` over the file passes.

Dockerfile
```diff
@@ -96,9 +96,23 @@ VOLUME /ipc
...
+#     is legitimately not ready for hours or days, so it would never survive
+#     long enough to finish. Readiness belongs in a Kubernetes
+#     readinessProbe or a load-balancer target check, where failing it
+#     drains traffic instead of killing the node.
+#
+# The liveness body still carries the readiness verdict and the observed tip
+# gap, so `docker inspect` shows why a live node is not yet serving.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=60s --retries=3 \
-  CMD wget -qO/dev/null http://127.0.0.1:12798/metrics || exit 1
+  CMD wget -qO/dev/null http://127.0.0.1:12799/health || exit 1
```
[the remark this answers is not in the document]
wolf31o2: Fixed. The check resolves `${DINGO_HEALTH_PORT:-12799}` at probe time; Docker stores the instruction as `CMD-SHELL` and performs no build-time expansion, so the variable is read from the container environment.

One case beyond the report: `healthPort: 0` disables the listener, and `${DINGO_HEALTH_PORT:-12799}` yields `0` rather than the fallback for an explicit `0`, so the check would probe a port nothing is bound to and the runtime would replace a container that was configured exactly as asked. The check now reports healthy without probing in that case.

A `--health-port` flag or a `healthPort` YAML value remains invisible to a `HEALTHCHECK`; the environment variable is the only one of the three sources reachable from inside the check, which is now stated where the instruction is defined.

internal/config/validate.go
```diff
@@ -296,6 +299,7 @@ func (c *Config) validate(effectiveMode RunMode, minBindable uint) error {
 		{"privatePort", c.PrivateBindAddr, c.PrivatePort, serving, serving},
 		{"metricsPort", c.BindAddr, c.MetricsPort, auxListeners, serving},
 		{"debugPort", c.DebugBindAddr, c.DebugPort, auxListeners, false},
+		{"healthPort", c.BindAddr, c.HealthPort, serving, false},
```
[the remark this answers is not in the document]
wolf31o2: Fixed in `c1353d51`. `bindAddrsOverlap` compares through a new `normalizeBindAddr`, which parses IP literals, strips brackets, and unmaps IPv4-mapped IPv6. Hostnames are still not resolved and those conflicts still surface at bind time.

The class had a second member. `isWildcardAddr` matched only the literal strings `""`, `0.0.0.0`, `::` and `[::]`, so `0:0:0:0:0:0:0:0` was not recognized as a wildcard at all and did not overlap a specific address sharing its port. It now receives the canonical form.

Three `TestValidate` cases cover it: `equivalent IPv6 spellings collide` and `long-form IPv6 wildcard overlaps a specific address` both fail when the normalization is reverted in place, and `distinct IPv6 addresses may share a port` is the control showing normalization invents no collision.

Dockerfile
```diff
@@ -96,9 +96,31 @@ VOLUME /ipc
...
+# The port follows DINGO_HEALTH_PORT, which is the only one of the three
+# healthPort sources (flag, YAML, environment) a HEALTHCHECK can read. An
+# explicit 0 disables the listener, so the check reports healthy rather than
+# probing a port nothing is bound to and driving the container into a
+# replacement loop.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=60s --retries=3 \
-  CMD wget -qO/dev/null http://127.0.0.1:12798/metrics || exit 1
+  CMD health_port="${DINGO_HEALTH_PORT:-12799}"; \
+  [ "$health_port" = "0" ] && exit 0; \
+  wget -qO/dev/null "http://127.0.0.1:$health_port/health" || exit 1
```
chrisguiney, line 123: Blocker: with `RESTORE_SNAPSHOT` set, `bin/entrypoint.sh` runs `dingo mithril sync` here before `serve`. That mode starts the metrics listener (`cmd/dingo/mithril.go:419`) but no health listener (`validate.go:303`), so this probe is refused for the whole bootstrap. At this head during sync, `:12798/metrics` returns 200 and `:12799/health` is refused. After the start period and 3 retries, Swarm replaces the task. `docker-compose.yml:46` has the same problem. Fix: serve health in `runMithrilSync`, enable the `healthPort` row for `auxListeners`, and add a test.

chrisguiney (changes requested):
Re-reviewed `a590b021`. The prior blockers are cleared. The `config.go` merge keeps both health defaults in `newDefaultConfig`, and `TestHealthPortDefaultsToDefaultHealthPort` fails without the `HealthPort` line. The `reinitializeCoreStorage` clear is tested and fails without the call. `t.Parallel()` is added. CodeRabbit's port and bind-address findings are addressed. The readiness freshness bound is still open but not blocking. Merged with current main: build, vet, race tests on touched packages and scoped lint pass. One new blocker is inline.

Dockerfile
wolf31o2: Confirmed and fixed in 8fd36795. `runMithrilSync` started only the metrics and pprof listeners, so the probe was refused for the whole bootstrap, and the image HEALTHCHECK's start period plus three retries is 150s against a mainnet snapshot download measured in hours.

`cmd/dingo.startHealthProbeServer` now serves the same listener during the bootstrap, over `internal/node.NewHealthServer` (exported for this) with a nil tip gap: `/health` 200, `/readyz` 503 with reason `tip gap unavailable`. `healthPort` joins `metricsPort` and `debugPort` in the `auxListeners` set in `internal/config/validate.go:305`, so a bad or colliding probe port is rejected before a sync starts instead of surfacing as a refused probe.

Tests: `TestMithrilSyncServesHealthProbe` drives `runMithrilSync` to its first local failure and asserts the listener came up; with the wiring removed it sees only the metrics line. `TestStartHealthProbeServerServesLiveAndNotReady` pins the 200/503 split and `ready:false`. `TestValidateSyncModeValidatesHealthPort` and `TestValidateSyncModeRejectsHealthMetricsCollision` both report "An error is expected but got nil" with the gate reverted to `serving`.

Class check: `runMithrilSync` is the only non-serve entry point in `cmd/dingo` that starts any auxiliary listener at all. `load` and the `database` subcommands start no metrics or pprof listener either, by the rule documented at `internal/config/validate.go:262`, so health now follows the same gate as the other two rather than becoming a special case; `bin/entrypoint.sh` runs neither of those before `serve`.

chrisguiney (approved):

Issue #3678, opened 2026-08-30
ledger: core-mode sync stalls on Conway UTxO validation divergence


wolf31o2:
## Problem

Freshly bootstrapped Preview and Preprod core-mode validation nodes reach the live chain, then repeatedly reject the same Conway transaction during at-tip recovery with `value not conserved` and missing-input errors. Recovery rewinds to the same ledger tip, reconnects to the peer, and retries without progress until the ledger pipeline is marked stuck.

## Impact

Core-mode validation cannot remain synchronized on either Preview or Preprod after reaching the affected transaction.

## Observed behavior

- Conway UTxO validation reports rule 29 (bad inputs) and rule 31 (value not conserved).
- At-tip recovery exhausts rewind attempts and reports local ledger validation divergence.
- The node repeatedly restarts and eventually reports that the ledger pipeline is stuck.

The failures were reproduced independently on Preview and Preprod against their local Cardano-node peers. Transaction payloads, hostnames, filesystem paths, and private topology details are omitted.

## Related work

This is distinct from reward-state withdrawal divergence tracked in #3628 and the proposed replay disposition in #3639. Please determine whether the common UTxO-validation failure is caused by incorrect transaction/body decoding, era parameters, or local ledger state, and add regression coverage.

## Acceptance criteria

- Identify why valid network transactions are decoded or validated with missing inputs/value imbalance.
- Ensure recovery either converges or fails with a useful persistent diagnostic without looping indefinitely.
- Re-run Preview and Preprod core-mode sync through a live serve window.

PR #3952, opened 2026-09-05
fix(ledger): refuse rollbacks below the consumed UTxO prune floor

### LINKED ISSUE #3766: ledger: at-tip recovery rewind leaves the UTxO set inconsistent, degrading to a halted pipeline at the Mithril trust boundary

wolf31o2:
## Cause

Consumed-UTxO cleanup permanently removes spent rows. A rollback below the highest swept slot cannot restore those rows, so moving the ledger tip there would leave the live UTxO set incomplete.

## Change

- use `consumed_utxo_prune_floor` as the single durable prune marker
- serialize cleanup with the floor check, primary-chain rewind, and ledger rollback
- refuse rollback and replay-recovery targets below the floor before either chain or ledger state moves
- keep lagging on-chain peers attached instead of treating their below-floor rollback as divergence
- request a fresh ChainSync connection for genuine below-floor divergence
- clamp escalating at-tip recovery to retained history
- make offline truncate refuse targets below the same floor
- avoid lock inversion by keeping ChainSync state transitions outside the prune boundary

## Coverage

Coverage includes direct rollback refusal, same-slot target resolution, replay recovery, lagging-peer handling, cleanup-floor persistence, truncate refusal, and cleanup progress while `chainsyncMutex` is held.

`ARCHITECTURE.md` and `DATABASE.md` document the rollback and storage contracts.

Fixes #3766


ledger/state.go
```diff
@@ -3327,6 +3455,27 @@ func (ls *LedgerState) rollbackChainAndState(point ocommon.Point) error {
 	if mithrilLedgerSlot > 0 && point.Slot < mithrilLedgerSlot {
 		return ErrRollbackExceedsMithrilBoundary
 	}
+	// Check the consumed-UTxO prune floor here as well as in ls.rollback:
```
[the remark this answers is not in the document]
wolf31o2: Valid, fixed in f6f33a63. `rollback`'s prune-floor guard already ran after the same-slot redirect, so `rollback` itself was safe, but `rollbackChainAndState` checked the unresolved point and then ran `chain.Rollback` before `rollback` refused — splitting the chain from the ledger.

The redirect is now `resolveRollbackTarget`, and both the Mithril and prune-floor pre-checks run against the resolved target. `TestRollbackChainAndStateRefusesRedirectBelowPruneFloor` covers it; without the fix the primary chain truncates from slot 140000 to 114080 before the refusal.

The tip can still move between the pre-check and `rollback`'s own check — the pre-existing Mithril check has the same window — and `rollback` remains the backstop that refuses before mutating the ledger.

database/utxo_prune_floor.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return slot, nil
+}
+
+// setConsumedUtxoPruneFloor raises the recorded floor to slot. The floor only
+// ever moves up: a later sweep at a lower slot (the tip fell after a rollback)
+// does not make the rows an earlier, higher sweep already removed restorable
+// again.
+func (d *Database) setConsumedUtxoPruneFloor(slot uint64, txn *Txn) error {
+	current, err := d.ConsumedUtxoPruneFloor(txn)
```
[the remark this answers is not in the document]
wolf31o2: Valid, fixed in f6f33a63. `UtxosDeleteConsumed` now reads the recorded floor before `GetUtxosDeletedBeforeSlot`, and writes afterwards through `writeConsumedUtxoPruneFloor`, which performs no read of its own. The only work left after the irreversible blob deletes is a `SetSyncState` write, and the monotonic comparison uses the value read up front.

ledger/chainsync.go
```diff
@@ -186,6 +186,16 @@ var ErrRollbackExceedsMithrilBoundary = errors.New(
 	"rollback exceeds Mithril trust boundary",
 )
 
+// ErrRollbackBelowUtxoPruneFloor reports a rollback target below the slot the
+// consumed-UTxO sweep has hard-deleted spent rows at or below. Those rows are
+// gone, and database.TruncateAfterSlot restores spent UTxOs with an UPDATE, so
+// the rewind cannot reconstruct the live set the target implies. The rollback
+// is refused with the ledger untouched rather than performed and reported as a
+// repair (issue #3766).
+var ErrRollbackBelowUtxoPruneFloor = errors.New(
```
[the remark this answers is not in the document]
wolf31o2: Valid, fixed in f6f33a63. `handleEventChainsyncRollback` returned the error, and `handleEventChainsync` routes a rollback error to `FatalErrorFunc`, so a peer's choice of rollback point could terminate the node.

It is now rejected like the Mithril boundary: log, `reportUnrecoverableRollbackIfStuck`, `resetChainsyncResyncState`, `setChainsyncState(SyncingChainsyncState)`, and a resync event carrying the new `ChainsyncResyncReasonRollbackBelowUtxoPruneFloor`. `TestHandleEventChainsyncRollbackRejectsBelowPruneFloor` asserts `FatalErrorFunc` is not called and neither the chain tip nor the ledger tip moves; without the fix it fails with `chain rollback failed: rollback below consumed UTxO prune floor`.

database/lifecycle/truncate_test.go
```diff
@@ -846,3 +846,72 @@ func TestTruncateRejectsPreCancelledContextWithoutRecordingMarker(
...
+	t.Run("target below the floor clears it", func(t *testing.T) {
+		f := newFixtureWithFloor(t)
+		// blocks[1] is at slot 20, below the swept slot.
+		_, err := lifecycle.Truncate(
+			context.Background(), f.db, f.blocks[1], 0, false, 0,
+		)
+		require.NoError(t, err)
+		floor, err := f.db.ConsumedUtxoPruneFloor(nil)
+		require.NoError(t, err)
+		require.Zero(
```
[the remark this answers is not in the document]
wolf31o2: Valid, fixed in f6f33a63. Both subtests now assert the swept row is absent via `GetUtxoIncludingSpent`, and the fixture asserts it immediately after the sweep as well, so a regression that re-materialized swept rows fails rather than passing on the floor assertion alone.

ledger/state.go
```diff
@@ -1651,6 +1664,78 @@ func (ls *LedgerState) subscribeBlockfetchEvents(
...
+	if err != nil {
+		ls.config.Logger.Error(
+			"failed to read consumed UTxO prune floor",
+			"component", "ledger",
+			"error", err,
+		)
+		return
+	}
+	ls.consumedUtxoPruneFloor.Store(floor)
+	ls.consumedUtxoPruneFloorLoaded.Store(true)
```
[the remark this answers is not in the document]
wolf31o2: Valid on both counts, fixed in f6f33a63 by removing the mirror. `rollbackBelowConsumedUtxoPruneFloor` reads the `consumed_utxo_prune_slot` sync-state row directly and fails closed on a read or parse error; `consumedUtxoPruneFloor`, `consumedUtxoPruneFloorLoaded`, `recordConsumedUtxoPruneFloor` and `loadConsumedUtxoPruneFloor` are gone.

That closes both windows rather than narrowing them: neither a failed startup load nor the gap between the sweep's commit and a cache refresh can present a floor lower than the database holds. A rollback is rare and already opens transactions of its own, so the read is affordable. `TestConsumedUtxoPruneFloorIsReadFromTheDatabase` pins that a floor written by another writer is honored with no reload step, and that a malformed value refuses the rollback.

wolf31o2:
All five cubic findings addressed in f6f33a63.

- `rollbackChainAndState` checked the unresolved point, so a same-slot competitor at the floor passed and then redirected below it inside `rollback` — after `chain.Rollback` had run. The redirect is now `resolveRollbackTarget`, and both the Mithril and prune-floor pre-checks run against the resolved target.
- `handleEventChainsyncRollback` returned `ErrRollbackBelowUtxoPruneFloor`, which `handleEventChainsync` routes to `FatalErrorFunc`. It is now rejected like the Mithril boundary, with `ChainsyncResyncReasonRollbackBelowUtxoPruneFloor`.
- `UtxosDeleteConsumed` read the floor after deleting blobs and rows; it now reads before any deletion and writes through `writeConsumedUtxoPruneFloor`, which does no read of its own.
- The in-memory floor mirror is removed. `rollbackBelowConsumedUtxoPruneFloor` reads the sync-state row directly and fails closed, closing both staleness windows rather than narrowing them.
- `TestTruncateClearsConsumedUtxoPruneFloorAboveTarget` now asserts the swept row stays hard-deleted across the truncate.

New coverage, each failing without its fix with `go vet` exit 0 and a sibling passing in the same run:

```
--- FAIL: TestRollbackChainAndStateRefusesRedirectBelowPruneFloor
    the primary chain must not be truncated for a refused rollback
    Slot: (uint64) 140000 expected, 114080 actual
--- FAIL: TestHandleEventChainsyncRollbackRejectsBelowPruneFloor
    Received unexpected error: chain rollback failed: rollback below
    consumed UTxO prune floor: target slot 10, prune floor 20
--- FAIL: TestTruncateClearsConsumedUtxoPruneFloorAboveTarget/target_below_the_floor_clears_it
```

Re-validated: `go build ./...` 0, `go vet ./...` 0, `gofmt -s -l` clean, `make golines` clean, `golangci-lint run ./ledger/... ./database/... ./event/...` 0 issues, `go test ./...` exit 0, `go test -race -timeout 75m ./ledger/ ./ledger/eras/ ./database/ ./database/lifecycle/ ./event/...` exit 0, `make docs-parity` ok.

ledger/replay_recovery.go
```diff
@@ -1224,6 +1224,50 @@ func (ls *LedgerState) recoverAtTipFromTxValidationError(
...
+	// short of every output consumed above it. Blocks the node already applied
+	// cleanly then fail to resolve their inputs, which drives the next, deeper
+	// rewind: the descent that ends at the Mithril anchor with a halted
+	// pipeline (issue #3766).
+	//
+	// Fall back to the ledger tip, which is always at or above the floor and
+	// is what attempt 1 uses. Recovery keeps its non-destructive lever -- a
+	// fresh intersection so peer rotation can offer a different candidate
+	// chain -- and loses only a rewind that could not have repaired anything.
+	belowPruneFloor, pruneFloor, floorErr := ls.rollbackBelowConsumedUtxoPruneFloor(
```
chrisguiney, line 1299: **Blocker.** This clamp covers `recoverAtTipFromTxValidationError` only. The sibling replay path in `tryRecoverFromTxValidationError` has no equivalent, and it moves the primary chain first:

```go
if rewindPrimaryChain && !primaryChainAlreadyHeld {
    if err := ls.rewindPrimaryChainForRecovery(rewindPoint); err != nil { ... }   // chain truncated
    primaryChainRewound = true
}
if err := ls.rollback(rewindPoint); err != nil {                                  // can now refuse
```

That is exactly the ordering hazard this PR fixed for `rollbackChainAndStateDeferred` and wrote up in ARCHITECTURE.md, left in place on the caller beside it.

Reproduced at this head with the PR's own `newPrunedUtxoFixture`, calling the two functions in that order:

```
chain tip 140000 -> 88160
ls.rollback err = rollback below consumed UTxO prune floor: target slot 88160, prune floor 114080
ledger tip 140000 -> 140000
reconcilePrimaryChainTipWithLedgerTip err = rollback ledger tip to primary chain tip: \
    rollback below consumed UTxO prune floor: target slot 88160, prune floor 114080
```

So the chain ends below the floor with the ledger above it, the pipeline halts, and startup reconciliation then refuses the repair — the node does not restart. That is worse than the silent divergence being fixed, because it needs `dingo database truncate` rather than a restart.

Reachability of a below-floor target on that path: `rewindPrimaryChain` is set only when `candidate.ProducerUnresolved`, which only the `security-param-fallback` candidate sets. Its `RollbackPoint` is `replayRecoveryParentPoint(anchorBlock)` — the *parent* of an anchor clamped down by `durableAppliedFloorAnchorIndex()`. After a rollback that lands exactly on the floor (allowed, and asserted by `TestRollbackBelowConsumedUtxoPruneFloorIsRefused`), the durable applied floor is the floor block itself and its parent is strictly below the floor.

`recoveryRollbackExceedsMithrilBoundary(candidate.RollbackPoint)` is already pre-checked before the rewind on that path; the prune floor belongs beside it, with the same reject-and-return shape.

event/chainsync_resync.go
```diff
@@ -34,6 +34,7 @@ const (
 	ChainsyncResyncReasonRollbackExceedsK                  = "rollback exceeds security parameter K"
 	ChainsyncResyncReasonRollbackExceedsMithril            = "rollback exceeds Mithril trust boundary"
 	ChainsyncResyncReasonPeerTipBehindMithril              = "peer tip behind Mithril trust boundary"
+	ChainsyncResyncReasonRollbackBelowUtxoPruneFloor       = "rollback below consumed UTxO prune floor"
```
chrisguiney, line 37: The new reason is never registered in `ouroboros/chainsync.go`. It is absent from both `chainsyncResyncRequiresFreshConnection` and `chainsyncResyncDeniesPeer`, while every sibling refuse-the-peer reason — `ChainsyncResyncReasonRollbackExceedsK`, `ChainsyncResyncReasonRollbackExceedsMithril`, `ChainsyncResyncReasonPeerTipBehindMithril` — is in both.

The handler comment in `ledger/chainsync.go` says the peer is rejected "exactly as the Mithril boundary does", but the event instead falls through to `restartChainsyncClientAsync`: an in-place Stop→Start→Sync on the same bearer, which the comment directly above that branch notes can block up to 30s in `MustReply`. The peer then resolves the same intersect and asks for the same rollback.

Absence from the deny list is arguably right (a peer this refuses may only be lagging). Absence from the fresh-connection list looks unintended.

ledger/chainsync.go
```diff
@@ -2446,6 +2456,45 @@ func (ls *LedgerState) handleEventChainsyncRollback(
 			)
 			return nil
 		}
+		if errors.Is(err, ErrRollbackBelowUtxoPruneFloor) {
```
chrisguiney, line 2555: Non-blocking, but worth checking against current `main`: #3998 (`7dd8a398`) landed after this branch's merge base and is not accounted for here.

`rollbackChainAndStateDeferred` now returns `ErrRollbackBelowUtxoPruneFloor` before `chain.RollbackDeferred` runs, so it preempts `chain.ErrRollbackExceedsSecurityParam`. A peer that is merely behind on our own chain by more than a stability window previously reached the over-K branch, which #3998 gave a `chainsyncPeerBehindOnOurChain` discriminator specifically so that peer is kept rather than re-synced, denied and escalated. It now lands in this branch instead, which logs at ERROR, calls `reportUnrecoverableRollbackIfStuck` — the path that advises re-bootstrapping from a Mithril snapshot — and re-syncs.

Bounded: peers between K blocks and a stability window behind still resolve above the floor and keep the #3998 behaviour, and the loop-detector branch does apply the discriminator after `rollbackLoopThreshold` repeats. But the same `chainsyncPeerBehindOnOurChain` check would fit here for the same reason it fits the two branches above.

chrisguiney (changes requested):
Reviewed head `53467ad5`. Merge base `19a8f2c6`, 18 commits behind `origin/main` (`7ad27910`); correctness checks re-run on a local merge of `origin/main` into the head.

## Disposition

Requesting changes on one item: the "refuse before the chain moves" invariant this PR establishes is not held on the replay-recovery path, and the resulting split state makes the node unstartable. Details inline at `ledger/replay_recovery.go`.

The core change is sound. The diff matches the description (12 files, 1260/-53 against the merge base; no module changes). Every test the description claims fails-without-the-fix does, verified by reverting each guard in place — see the ledger below.

## Bot standing on this head

- **CodeRabbit: did not review.** The check is green, but the comment body is "Review limit reached / You've used all free OSS reviews for now."
- **cubic: reviewed `9d6dd7a5` and `f6f33a63`, not this head.** Nothing has reviewed `bc140433`, `b16b9921` or `53467ad5`.
  - `ledger/state.go:3164` (P1, open): false positive — replied on the thread. Assigning `point = resolved` would truncate the primary chain below the contested slot, which is not what #3678 landed; `TestRollbackSameSlotCompetitorRestoresConsumedUtxo` on main pins the ledger sitting at the ancestor while the chain holds the requested point, and `rollbackChainAndStateDeferred`'s own `pointMatches` branch already handles the two ending apart.
  - `ledger/state.go:3671` (P3, open): already fixed by `bc140433`, which cubic never re-evaluated. `rollbackIsAppliable` resolves the target before the floor comparison at head.
  - The six resolved threads are all genuinely addressed at head; I re-checked each.

## Class audit: rollback entry points

| Entry point | Pre-checked before the chain moves | On refusal |
| --- | --- | --- |
| chainsync roll-backward (`handleEventChainsyncRollback`) | yes, resolved target | non-fatal, resync (see `event/chainsync_resync.go` note) |
| fork resolution (`tryResolveFork`) | yes, same pre-check | falls to the generic `else`; not fatal, but no resync and `headerMismatchCount` is decremented. Same pre-existing gap `ErrRollbackExceedsMithrilBoundary` has there |
| at-tip validation recovery | yes, clamped to the ledger tip | metric + warn, no error |
| replay recovery (unresolved producer) | **no** | see inline blocker |
| deterministic tx recovery / header validation recovery | n/a — target is the ledger tip, which cannot sit below the floor | — |
| chain-iterator rollback (`processChainIteratorRollback`) | n/a — chain already moved by its emitter | inherits whichever emitter produced it |
| startup reconciliation, `RecoverCommitTimestampConflict`, `enforceDurableTipFloor` | n/a | hard error, node does not start (documented in ARCHITECTURE.md) |
| `lifecycle.Truncate` | exempt by design, clears the floor when crossed | — |
| blockfetch, Mithril import, local state query | do not reach `LedgerState.rollback` | — |

## Liveness questions

- **Floor vs k.** Floor is `ledgerTip - stabilityWindow` (3k/f, 2k in Byron), monotonic, written only when a sweep actually removed rows. A target exactly on the floor is allowed and tested. Near origin nothing is written (`tipSlot > stabilityWindow` gate), so a fresh or Mithril-bootstrapped node has floor 0 and refuses nothing — the obvious false-positive source is clean. `calculateStabilityWindowForEra` falls back to 50000 slots on unusable genesis config, which is tighter than 3k/f on mainnet, but that fallback already governs the sweep, so it is pre-existing rather than introduced here.
- **Persistence.** `sync_state`, survives restart; wiped by a full Mithril re-bootstrap's `ClearSyncState`; cleared by `lifecycle.Truncate` when crossed. A floor ahead of the stored tip refuses startup — documented, and reachable via the inline blocker.

## Evidence

| Check | Result |
| --- | --- |
| `go build ./...` (head, and merged with `origin/main`) | 0 |
| `go vet ./...` (merged) | 0 |
| `go test ./...` (merged) | `internal/koiosparity` and `ledger/leader` fail identically on clean `origin/main`; `bin` `TestEntrypointForwardsSignalsDuringMithrilBootstrap` is a load flake and passes in isolation. All other packages pass, including `ledger`, `database`, `database/lifecycle`, `chain`, `event`. |
| `golangci-lint run ./ledger/... ./database/... ./event/...` (isolated cache) | 0 issues |
| `make docs-parity`, `make import-boundaries` | ok |
| `gofmt -s -l` on the touched files | clean |
| `examples/dingo-gov-lens`, `internal/test/antithesis` | build/vet/test 0; neither `go.mod` changed |

Fail-before, each reverted in place with `go vet` still 0:
- floor predicate forced false: all seven `ledger/utxo_prune_floor_test.go` tests fail.
- `rollbackChainAndStateDeferred` pre-check against the unresolved point: `TestRollbackChainAndStateRefusesRedirectBelowPruneFloor` fails on "the primary chain must not be truncated for a refused rollback".
- `handleEventChainsyncRollback` branch disabled: `TestHandleEventChainsyncRollbackRejectsBelowPruneFloor` fails on the fatal-path assertion.
- `ClearConsumedUtxoPruneFloorAbove` call removed: `TestTruncateClearsConsumedUtxoPruneFloorAboveTarget/target_below_the_floor_clears_it` fails while the above-the-floor sibling passes in the same run.

Not run: DevNet, live Mithril-bootstrapped Preview node, Koios parity replay, Antithesis. None are available here, and none of them is exercised by this diff at unit level.

Minor: the description says `Fixes #3953` while the head commit body says `Fixes #3766`. Both are open and distinct; worth settling which the squash should close.

ledger/state.go
```diff
@@ -2880,55 +2984,47 @@ func (ls *LedgerState) rollback(point ocommon.Point) error {
...
-		tipNonce, nonceErr := ls.db.GetBlockNonce(currentTip.Point, nil)
-		if nonceErr != nil {
-			return fmt.Errorf(
-				"read applied nonce for contested tip at slot %d: %w",
-				currentTip.Point.Slot,
-				nonceErr,
-			)
-		}
-		sameSlotCompetitor = len(tipNonce) > 0
+	resolved, resolveErr := ls.resolveRollbackTarget(point, currentTip)
```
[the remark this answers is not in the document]
chrisguiney: Reviewer disposition: false positive, not a blocker.

Assigning `point = resolved` here would truncate the primary chain below the contested slot as well. That is not what #3818 landed: `TestRollbackSameSlotCompetitorRestoresConsumedUtxo` on `main` asserts the ledger tip is redirected to the applied ancestor "so the block at the contested slot can be re-applied", while the chain keeps the requested point and the ledger replays forward across it. Moving the chain down too would drop the peer's block from the primary chain and force a refetch.

The two ending apart is an expected state this function already handles — the `pointMatches(ls.Tip().Point, point)` branch below `ls.rollback(point)` exists for exactly that, and it is unchanged by this PR.

What did need fixing was the boundary check running against the unresolved point, and `f6f33a63` fixed that.

ledger/state.go
```diff
@@ -3323,10 +3419,45 @@ func (ls *LedgerState) drainBlockPipelineBeforeRollback(
...
+	// remains the backstop that refuses before mutating the ledger.
+	resolved, resolveErr := ls.resolveRollbackTarget(point, currentTip)
+	if resolveErr != nil {
+		return resolveErr
+	}
+	if mithrilLedgerSlot > 0 && resolved.Slot < mithrilLedgerSlot {
 		return ErrRollbackExceedsMithrilBoundary
 	}
+	belowPruneFloor, pruneFloor, floorErr := ls.rollbackBelowConsumedUtxoPruneFloor(
+		resolved,
```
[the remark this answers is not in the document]
chrisguiney: Already addressed at the current head, in `bc140433` — cubic reviewed `f6f33a63` and has not re-evaluated since. `rollbackIsAppliable` now resolves the target before both the Mithril and prune-floor comparisons:

```go
resolved, err := ls.resolveRollbackTarget(point, currentTip)
if err != nil { return false }
...
belowPruneFloor, _, err := ls.rollbackBelowConsumedUtxoPruneFloor(resolved)
```

`TestRollbackIsAppliableRejectsBelowConsumedUtxoPruneFloor` covers it; it fails when the floor predicate is neutralised. `ls.chain.ValidateRollback(point)` deliberately keeps the unresolved point, since that is what the chain rolls back to.

database/lifecycle/truncate.go
```diff
@@ -757,6 +757,28 @@ func finishPendingTruncate(
...
 		}
+		// A truncate below the consumed-UTxO prune floor leaves the
+		// recorded floor above the new tip, which would make
+		// ledger.LedgerState.rollback refuse every rollback until the node
+		// resynced past it -- wedging the recovery this truncate was run to
+		// enable. Clear it, and say plainly which range's consumed UTxOs the
+		// sweep already removed: TruncateAfterSlot restores spent UTxOs with
+		// an UPDATE, so rows swept in that range did not come back here
+		// either (issue #3766).
+		oldFloor, cleared, floorErr := db.ClearConsumedUtxoPruneFloorAbove(
```
chrisguiney: Blocker: unreachable once current main is merged. #4345 (f9406950, on main since 2026-09-16) makes Truncate refuse any target below consumed_utxo_prune_floor, and cleanupConsumedUtxos writes that marker at the same slot as consumed_utxo_prune_slot and never lower, so a target below the slot marker is also below the floor marker and this clear can never run. Setting database.ConsumedUtxoPruneFloorSyncKey in newFixtureWithFloor, as the sweep does, makes the target-below-the-floor subtest fail with ErrTruncateNotStarted. ARCHITECTURE.md still promises truncate clears it.

database/utxo_prune_floor.go
```diff
@@ -0,0 +1,111 @@
...
+// and reports the tip repaired anyway.
+//
+// The two bounds are meant to line up: the sweep runs at
+// tip-stabilityWindow and a single rollback reaches at most one stability
+// window below the tip. Nothing enforced that, and successive recovery rewinds
+// compound -- each one rewinds a further window below the *already lowered*
+// tip while this floor stays fixed at the highest tip the node reached (issue
+// #3766). Recording the floor lets rollback refuse rather than silently
+// diverge.
+const consumedUtxoPruneFloorSyncKey = "consumed_utxo_prune_slot"
```
chrisguiney: Blocker: once current main is merged this package holds two prune-floor markers whose identifiers differ only in the leading capital. database/utxo.go:1215 defines ConsumedUtxoPruneFloorSyncKey = consumed_utxo_prune_floor and its doc calls itself the single source of truth for that name; this one is consumed_utxo_prune_slot. cleanupConsumedUtxos writes both on every sweep, and lifecycle.Truncate now enforces one while this change clears the other. Reconcile them to a single marker, or rename and document the second so the split is deliberate.

ledger/state.go
```diff
@@ -3031,6 +3050,11 @@ func (ls *LedgerState) cleanupConsumedUtxos() {
 		return
 	}
 	defer ls.cleanupConsumedUtxosRunning.Store(false)
+	// Serialize pruning with chainsync rollback. The rollback path resolves and
```
chrisguiney: Blocker: this serializes the sweep only against handleEventChainsyncRollback. Every recovery path that truncates the chain runs on the block-processing goroutine without chainsyncMutex - tryRecoverFromTxValidationError, recoverFromDeterministicTxValidationError, rejectRecoveryAtMithrilBoundary (which takes the mutex itself at replay_recovery.go:1770, proving it is not held), the at-tip clamp, and tryRecoverFromHeaderValidationError - so the floor can still advance between the pre-check and ls.rollback's own read, leaving the chain below it and the ledger above.

ledger/state.go
```diff
@@ -3031,6 +3050,11 @@ func (ls *LedgerState) cleanupConsumedUtxos() {
...
+	// validates its target before truncating the primary chain, so the floor
+	// cannot advance between that validation and the ledger mutation.
+	ls.chainsyncMutex.Lock()
```
chrisguiney: Major: Close waits unconditionally on cleanupWG, and the comment above that wait justifies it because a run is bounded by cleanupConsumedUtxoBatchSize, one short delete transaction. beginCleanupConsumedUtxosRun registers the run before this Lock, so the wait is now bounded by whoever holds chainsyncMutex. The sweep also holds it across UtxosDeleteConsumed for up to 1000 rows including blob deletes, stalling header admission and rollback handling every 5 minutes. Update that justification, or hold the lock only around the floor read and the mutation.

ledger/replay_recovery.go
```diff
@@ -334,6 +334,16 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 		// cannot suppress the fresh ChainSync intersection.
 		ls.publishReplayRecoveryNonConvergingResync(rewindPoint)
 	}
+	if err := ls.checkReplayRecoveryRollbackFloor(rewindPoint); err != nil {
+		if errors.Is(err, ErrRollbackBelowUtxoPruneFloor) {
+			// The peer's block requires state older than the retained UTxO
+			// history. Rotate the ChainSync intersection instead of returning
+			// an error that would retry the same peer and target.
+			ls.publishReplayRecoveryPruneFloorResync(rewindPoint)
```
chrisguiney, line 367: Prior blocker cleared: the floor is now checked before rewindPrimaryChainForRecovery on all three replay paths. Removing this guard in place makes TestReplayRecoveryRefusesBelowPruneFloorBeforeChainRewind fail with rollback ledger state for replay recovery: rollback below consumed UTxO prune floor, target slot 0, prune floor 120. Bot threads at this head: the cubic same-slot-competitor P1 on state.go stays a false positive, since the chain must keep the contested point; cubic rollbackIsAppliable P3 and CodeRabbit's three findings at 92e627e5 are addressed.

chrisguiney (changes requested):
Re-review at edb1f64cb2. The prior replay-recovery blocker is cleared and causally proven: removing the checkReplayRecoveryRollbackFloor call in place makes TestReplayRecoveryRefusesBelowPruneFloorBeforeChainRewind fail with the split-state error. Three blockers remain, inline. All 14 checks are green but started 2026-09-14T23:2xZ, while main last moved at f9406950 (#4345) on 2026-09-16, so no check run covers this branch merged with current main. Merged with origin/main: build, vet and scoped golangci-lint exit 0; ledger, database, database/lifecycle, event and ouroboros pass.

ledger/replay_recovery.go
```diff
@@ -209,6 +209,8 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 	if !errors.As(err, &validationErr) {
 		return false, nil
 	}
+	ls.consumedUtxoPruneMutex.Lock()
```
chrisguiney: Blocker: this lock inverts against chainsyncMutex and deadlocks the node. It is held for the whole call, and rejectRecoveryAtMithrilBoundary takes chainsyncMutex beneath it at replay_recovery.go:1772, via 1305 and 1607. cleanupConsumedUtxos takes them the other way: chainsyncMutex at state.go:3135, then this one at state.go:3210. Reproduced on this head with the trust-window fixture: the sweep waits on the prune mutex while tryRecoverFromTxValidationError waits at 1772. Both goroutines are real, the sweep at state.go:3131 and 6794 and the block pipeline at state.go:7202.

ledger/state.go
```diff
@@ -9310,7 +9454,7 @@ func (ls *LedgerState) reconcilePrimaryChainTipWithLedgerTip() error {
 			// validateAndEmitRollbackUndo contract, and the next
 			// reconciliation attempt lands right back in this same
 			// branch and retries both.
-			if err := ls.rollbackWithoutResync(chainTip.Point); err != nil {
+			if err := ls.rollbackWithResync(chainTip.Point, false); err != nil {
```
chrisguiney, line 9449: Blocker: the lint check is red on this head and this change owns it. Replacing this call and the one at state.go:9724 with rollbackWithResync leaves rollbackWithoutResync with no production caller, and .golangci.yml sets run.tests false, so state.go:3340 trips unused. golangci-lint 2.13.2 on ./ledger/ with current origin/main merged reports exactly that one issue, and on origin/main the helper still has both callers. Either route these two through the wrapper again or delete it.

ARCHITECTURE.md
```diff
@@ -601,6 +601,65 @@ rollback, primary-chain reconciliation, tip-floor enforcement) drives the
...
+boundary check admits every target the rewind schedule produces and only the
+prune floor refuses them.
+
+Startup reconciliation rolls the ledger back to the blob tip when metadata
+leads it, and that rollback can be arbitrarily deep, so it can now fail with
+`ErrRollbackBelowUtxoPruneFloor` and refuse to start. That is the intended
+outcome rather than a regression: the rewind it was about to perform could not
+have rebuilt the live UTxO set, so the alternative is a node that starts,
+validates against a UTxO set missing rows, and halts later with no way to tell
+what went wrong. The operator recovery is `dingo database truncate`, which is
```
chrisguiney: Blocker: this names a recovery the code now refuses. lifecycle.Truncate rejects a target below the floor at database/lifecycle/truncate.go:665 with ErrTruncateNotStarted, and this commit deletes ClearConsumedUtxoPruneFloorAbove, so truncate is bound by the floor and clears nothing. DATABASE.md was corrected in the same commit, so the two documents disagree and the wrong one is what an operator reads after startup refuses. State the real route: a truncate target at or above the floor, or a fresh bootstrap.

ledger/state.go
```diff
@@ -3179,6 +3207,8 @@ func (ls *LedgerState) cleanupConsumedUtxos() {
 		// "absent" for a ref that was actually there (blinklabs-io/dingo#382
 		// review). Returning here just skips this run; the next periodic
 		// tick tries again from the same (or a later) floor.
+		ls.consumedUtxoPruneMutex.Lock()
```
chrisguiney, line 3208: Still open from the last round. beginCleanupConsumedUtxosRun registers the run before chainsyncMutex is taken at state.go:3135, so Close's unconditional cleanupWG wait is now bounded by two locks, and the sweep holds both across UtxosDeleteConsumed for up to cleanupConsumedUtxoBatchSize rows including blob deletes. The comment above 3135 still justifies chainsyncMutex by an ordering this new mutex supersedes. Holding only this mutex, around the floor write and the delete, would drop both the chainsync stall and the inversion.

database/utxo_prune_floor.go
```diff
@@ -0,0 +1,77 @@
...
+// limitations under the License.
+
+package database
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// consumedUtxoPruneFloorSyncKey records how far the consumed-UTxO sweep has
```
chrisguiney: The constant this block documented was deleted here, and with no blank line between them go doc renders the whole block as ConsumedUtxoPruneFloor's public documentation, opening with a sentence about consumedUtxoPruneFloorSyncKey, an identifier that no longer exists. Move it beside ConsumedUtxoPruneFloorSyncKey at database/utxo.go:1205, or separate it from the method's own doc comment.

ARCHITECTURE.md
```diff
@@ -601,6 +601,65 @@ rollback, primary-chain reconciliation, tip-floor enforcement) drives the
...
+deliberately not bound by this floor (nor by the security parameter) and clears
+the record when it crosses it, or a fresh bootstrap.
```
chrisguiney: The pull-request description carries the same stale contract and is worth correcting before the squash: it says the sweep records consumed_utxo_prune_slot, which no longer exists anywhere in the tree, that lifecycle.Truncate stays unbound by the floor and clears the record, and it names TestTruncateClearsConsumedUtxoPruneFloorAboveTarget, renamed here to TestTruncateRejectsConsumedUtxoPruneFloorAboveTarget. Fixes #3953 also points at an issue already closed on 2026-09-16.

chrisguiney (dismissed):
Re-review at 9ccc4726. Both prior marker blockers are cleared and verified: ClearConsumedUtxoPruneFloorAbove is gone, truncate.go matches origin/main, and the tree holds one prune-floor key. The new consumedUtxoPruneMutex orders inversely to chainsyncMutex and deadlocks the node, the red lint check is owned by this change, and ARCHITECTURE.md still names a truncate recovery that Truncate now refuses. No bot reviewed this head: CodeRabbit reports incremental reviews disabled, cubic has no record, check or status here.

arepala-uml (approved):

Issue #2873, opened 2026-07-18
Complete the SPO-forging Prometheus series for failover/alerting


wolf31o2:
#### Motivation (dingo-operator)

The operator's `internal/forgestatus` reconciler scrapes the node's `/metrics` to populate
`status.kes`/`status.opcert` and to drive rotation/failover/alerting. Most cardano-node-compatible
SPO gauges already exist; this issue is a targeted request for the few series that are missing for
robust failover and rotation decisions. Please treat it as "add only what's absent."

#### Proposal

Already exported (verified in `ledger/forging/metrics.go`, wired via `ledger/forging/forger.go:352`):

- KES gauges: `cardano_node_metrics_currentKESPeriod_int` (`metrics.go:68`),
  `cardano_node_metrics_remainingKESPeriods_int` (`metrics.go:74`),
  `cardano_node_metrics_operationalCertificateStartKESPeriod_int` (`metrics.go:80`),
  `cardano_node_metrics_operationalCertificateExpiryKESPeriod_int` (`metrics.go:86`).
- Forge counters: `Forge_about_to_lead` (`metrics.go:94`), `Forge_node_is_leader` (`metrics.go:100`),
  `Forge_node_not_leader` (`metrics.go:106`), `Forge_forged` (`metrics.go:112`), `Forge_adopted`
  (`metrics.go:118`), `Forge_could_not_forge` (`metrics.go:124`).
- Dingo-specific: `dingo_metrics_slotBattlesTotal_int`, `dingo_forge_tip_gap_slots`,
  `dingo_forge_sync_skip_total`, validation/duration histograms, Leios EB counters, etc.

Missing for operator needs — propose adding:

1. **On-chain opcert counter gauge.** The opcert `IssueNumber` / on-chain counter is not a Prometheus
   series today (it is only reachable via the Blockfrost API). Export it as a gauge (ties to issue
   #2) so rotation logic and alerting can compare on-disk vs on-chain counter from `/metrics` alone.
2. **Missed-leader-slot counter.** There is no dedicated "missed leader slot" metric;
   `Forge_node_not_leader` and `Forge_could_not_forge` exist but neither captures a leadership slot
   that was won but not forged in time. Note the existing TODO at `ledger/state.go:1526`
   ("add callback to capture task run failure and increment 'missed slot leader check' metric").
3. **Point-in-time "is-leader-active" gauge (0/1).** `Forge_node_is_leader` (incremented at
[...]

PR #3955, opened 2026-09-05
fix(forging): stop dropping contested and scheduled leader slots before the leader check

### LINKED ISSUE #3954: forging: two Debug-only gates run before the leader check and silently drop scheduled leader slots, including on a node at tip
## Summary

`checkAndForgeProduction` has two skip gates that run **before** the leader check. Neither
increments a leadership-related counter and both log at `Debug`, so a scheduled leader slot
they drop is indistinguishable at INFO from "we were simply not the leader for that slot".

Over a 37 h window on our preview testnet block producer these gates swallowed **23,698 of
134,626 slot ticks (17.6 %)** and account for **8 of 11 lost leader slots across two preview
epochs** (adoption fell 80 % → 44 %) and **3 of 4** lost slots on a second Dingo BP on a
private devnet. One of the losses happened while the node was provably **at tip, 0 s behind,
with a healthy chain**.

Three separable defects, each with a concrete slot as evidence:

1. **No observability** — a dropped leader slot is invisible (`Debug` only; the first gate has
   no counter at all).
2. **`upstreamTip == 0` disables forging on a healthy node** — a best-peer switch resets the
   corroborated upstream target to 0, which the gate reads as "still syncing", unconditionally
   and regardless of `forgeSyncToleranceSlots`.
3. **Dingo declines slot battles** — `currentSlot <= tipSlot` skips the forge when *any* block
   for our slot has already been applied, where the Haskell node forges an alternative.

## Environment

- Dingo built from upstream `7f04e017` (2026-09-02, `fix(ledger): preserve committee credential
  tags and term semantics (#3745)`) plus local operational patches. `ledger/forging/forger.go`
  is **byte-identical** to upstream at that commit (verified by hash), so nothing below is a
  local modification.
- Re-checked against `main` @ `429dbdbb` (2026-09-04): **both gates are unchanged** — the only
  diffs to `forger.go` between `7f04e017` and `429dbdbb` are the Leios `ebSlot` plumbing and
  `golines` reformatting. Line numbers below are given for both.
- Networks: Cardano `preview` (1 s slots) plus a private Leios devnet.
- Block producer with a small upstream peer set, Prometheus scraping `/metrics`.

## The code

`ledger/forging/forger.go` (`7f04e017` / `main` line numbers):

| element | `7f04e017` | `main` (`429dbdbb`) |
|---|---|---|
| `forgeAboutToLead.Inc()` | 658 | 659 |
| Gate A `if currentSlot <= tipSlot` | 662 | 663 |
| Gate A log `"forge skip: slot already has block"` (`logger.Debug`) | 682–687 | 683–688 |
| `UpstreamSyncStatus()` call | 697 | 698 |
| Gate B condition | 698–700 | 699–701 |
| `forgeSyncSkip.Inc()` | 706 | 707 |
| Gate B log `"chain syncing from peer, skipping forge"` (`logger.Debug`) | 711–716 | 712–717 |
| **leader check** `isLeader := f.checkLeaderSafe(currentSlot)` | **778** | **784** |
| `forgeNotLeader.Inc()` | 786 | 792 |
| `forgeNodeIsLeader.Inc()` | 818 | 824 |
| `reserveForgeSlot(currentSlot)` (the #3734 fence) | 826 | 835 |

Gate A (`7f04e017` L662–689):

```go
	if currentSlot <= tipSlot {
		gap := tipSlot - currentSlot
		if f.metrics != nil {
			f.metrics.tipGapSlots.Set(float64(gap))
		}
		if gap > f.forgeStaleGapThresholdSlots {
			f.logger.Error(
				"chain tip is far ahead of slot clock; database may contain data from a different genesis",
				...
			)
		} else {
			f.logger.Debug(
				"forge skip: slot already has block",
				"current_slot", currentSlot,
				"tip_slot", tipSlot,
			)
		}
		return nil
	}
```

Gate B (`7f04e017` L697–718):

```go
	upstreamTip, upstreamActive := f.slotClock.UpstreamSyncStatus()
	if upstreamActive && (upstreamTip == 0 ||
		(upstreamTip > tipSlot &&
			upstreamTip-tipSlot > f.forgeSyncToleranceSlots)) {
		if f.metrics != nil {
			...
			f.metrics.forgeSyncSkip.Inc()
			f.metrics.tipGapSlots.Set(float64(gap))
		}
		f.logger.Debug(
			"chain syncing from peer, skipping forge",
			"current_slot", currentSlot,
			"tip_slot", tipSlot,
			"upstream_tip", upstreamTip,
		)
		return nil
	}
```

Both `return nil` before line 778/784, so `checkLeaderSafe` never runs, and neither
`Forge_node_is_leader_int` nor `Forge_node_not_leader_int` nor `Forge_could_not_forge_int`
moves. `forgeSyncToleranceSlots = 100` (`forger.go:58`), configurable via
`DINGO_FORGE_SYNC_TOLERANCE_SLOTS` (`internal/config/config.go:679`).

Note the contrast with the last-forged-slot fence added in #3734 — the very next gate in the
same function (`forger.go:1136` / main `1148`, called at `forger.go:826` / main `835`) does the right thing: it logs at
**Warn** with the slot, and increments a dedicated counter `dingo_metrics_forgeFenceBlocked_int`
(`ledger/forging/metrics.go:148`). The two pre-leader gates predate that convention.

### Evidence

37 h of counters from the preview BP:

```
cardano_node_metrics_Forge_about_to_lead_int   134626
cardano_node_metrics_Forge_node_not_leader_int 110892
dingo_forge_sync_skip_total                     23698
cardano_node_metrics_Forge_node_is_leader_int      36
cardano_node_metrics_Forge_could_not_forge_int       1
```

`110892 + 23698 + 36 = 134626`. The identity closes, which is the point: **17.6 % of slot ticks
never reached the leader check**, and the only signal that this happened is a counter with no
alerting story and no prose documentation.

Separately, a per-slot audit of the two preview epochs spanning that window (leader schedule from
Dingo's own VRF precompute, canonical inclusion re-derived from Koios, every slot cross-checked
against the producing node's own log) found **11 of 39 elapsed scheduled slots lost**.
`Forge_could_not_forge_int` explains exactly **one** of them. Eight were attributable by log
forensics to these two gates — no leadership line, no forge attempt, no INFO-level line naming
the slot at all.

Gate A is worse than Gate B: it has **no counter at all**. Its firing rate cannot be recovered
from `/metrics` even in principle — the `tipGapSlots` gauge it sets is shared with Gate B and is
last-write-wins. This has a concrete consequence for this report: we can confirm from the logs
that Gate A fired on preview slot 121863876 (Defect 3 below), but the counter identity above
closes to the unit, and there is no series that would let us or you say how often Gate A fires in
general. That unresolved gap is itself the argument for the fix.

`dingo_forge_sync_skip_total` is referenced by three panels in `docs/dashboards/*.json` but is
not described anywhere in `docs/`, and its help string — *"forging attempts skipped because
upstream tip is ahead"* (`metrics.go:172`) — is misleading, because the dominant firing case in
our data is `upstreamTip == 0`, i.e. upstream tip **unknown**, not ahead.

### Suggested fix

- Raise both skips to `Warn` when the slot is one this node is scheduled to lead. Dingo already
  precomputes the VRF leader schedule, so the gates can cheaply answer "is `currentSlot` in our
  schedule?" without running the full leader check. If consulting the schedule from inside the
  gate is undesirable, log unconditionally at `Info` with `current_slot`, `tip_slot`,
  `upstream_tip`, and which gate fired.
- Give Gate A its own counter (e.g. `dingo_forge_tip_skip_total`), and consider folding both
  into `Forge_could_not_forge_int` so cardano-node-compatible dashboards see the loss.
- Add a `dingo_forge_leader_slot_skipped_total` counter (scheduled slot never reached the leader
  check), which is the series #2873 item 2 is asking for from a different direction.
- Document `dingo_forge_sync_skip_total` and fix its help text to cover the unknown-target case.
- Consider a `make test` assertion of the identity
  `about_to_lead == not_leader + is_leader + sync_skip + tip_skip`, which today cannot even be
  written because Gate A is uncounted.

### Evidence: preview slot **121923532** (2026-09-05T03:38:52Z)

The BP was **at tip, 0 s behind**. The chain extended normally at 03:38:11 and again at
03:39:37, bracketing the slot. The node was scheduled to lead 121923532 (VRF precompute,
`leader_slots=24`, reloaded identically five times). **Zero log lines mention the slot** — no
leadership check, no forge attempt, no skip at INFO. No block was produced; Koios shows no block
at that slot for us.

31 s earlier, at ~03:38:21, the node performed a best-peer switch: `best peer disconnected` →
`selected new best peer` → `chain switch: updating active connection`. The new connection had
not yet delivered an authenticated, admitted header when the slot arrived.

### Code validation

`ledger/state.go:8558–8574` at `7f04e017` (identical block at `main:8765–8781`):

```go
// UpstreamSyncStatus reports whether a live upstream is selected and its
// corroborated target. An active upstream with target 0 is still syncing.
func (ls *LedgerState) UpstreamSyncStatus() (uint64, bool) {
	...
	state := ls.syncUpstreamState.Load()
	if state == nil || state.connectionKey != connIdKey(*activeConnId) {
		return 0, true
	}
	return state.targetSlot, true
}
```

and `publishActiveUpstream` (`7f04e017:8584–8600`, `main:8791–8807`), which runs on the peer switch:

```go
	ls.syncUpstreamState.Store(
		&upstreamSyncState{connectionKey: connIdKey(connId)},
	)
```

— stores the new connection key with `targetSlot` **zero**. The target only becomes non-zero in
`publishAdmittedUpstreamTarget` (`7f04e017:8609`, `main:8816`), i.e. after the new peer delivers a
header that is authenticated, admitted, *and* `SyncTargetTrusted`.

So for the whole interval between "active connection switched" and "new peer's first trusted
header admitted", `UpstreamSyncStatus()` returns `(0, true)`, Gate B's `upstreamTip == 0`
disjunct is true, and **forging is disabled unconditionally** — independent of the local tip's
freshness and independent of `forgeSyncToleranceSlots` (raising the tolerance does not help this
branch at all). In our logs that interval lasted at least 31 s; the node also recorded 410
best-peer disconnects and 280 `no selectable peer` events in 34 h, so the window is entered
often.

This behaviour is deliberate and pinned by
`TestCheckAndForgeProductionWaitsForUnknownActiveUpstreamTarget`
(`ledger/forging/forger_test.go:136–164`), so please read this as a request to revisit the
trade-off rather than as a report of an oversight. The intent — do not forge a conflicting block
while genuinely behind — is right; the proxy is wrong. "We have not heard a header from the peer
we selected 200 ms ago" is not evidence that our own chain is stale.

### Suggested fix

Make the gate's criterion the **local tip's freshness**, which is what it actually cares about,
and use the upstream frontier only when it is known:

- If `upstreamTip == 0`, fall back to wall-clock lag of the local tip: skip only if
  `currentSlot - tipSlot > forgeSyncToleranceSlots` (the node is demonstrably behind by its own
  clock). A node whose tip is one slot old should forge.
- And/or grace-period the gate for N slots after an active-connection switch, treating the
  retained previous frontier as still valid during the grace period (`UpstreamTipSlot()` already
  has a retained-frontier path at `7f04e017:8541–8554`, `main:8748–8761`).
- If neither is acceptable, at minimum make the unknown-target case a distinct, `Warn`-logged,
  separately-counted outcome so operators can see the cost.

### Evidence: preview slot **121863876** (2026-09-04T11:04:36Z)

This was our scheduled leader slot. A rival pool's block for the *same* slot reached us and was
applied **370 ms after slot start**. `tipSlot` therefore equalled `currentSlot` when the forge
tick fired, Gate A matched, and the node logged `forge skip: slot already has block` and
returned. We never signed a block, so we never entered the battle — the loss was guaranteed
rather than lost.

### Code validation

Gate A is `currentSlot <= tipSlot`. The `<` half matches the reference implementation. The `==`
half does not. In `ouroboros-consensus`, `mkCurrentBlockContext`
(`ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs:433–473`,
checked at `1820edf5e`) handles the three cases explicitly:

```haskell
  c' :> hdr -> case blockSlot hdr `compare` currentSlot of
    LT -> Right $ blockContextFromPrevHeader hdr
    GT -> Left $ TraceBlockFromFuture currentSlot (blockSlot hdr)
    -- The block at the tip has the same slot as the block we're going to
    -- produce (@currentSlot@).
    EQ ->
      Right $
        if isJust (headerIsEBB hdr)
          then blockContextFromPrevHeader hdr
          -- If @hdr@ is not an EBB, then forge an alternative to @hdr@: same
          -- block no and same predecessor.
          else BlockContext (blockNo hdr) $ castPoint $ AF.headPoint c'
```

`GT` (tip in the future) declines, matching Dingo's `<`. `EQ` **forges an alternative** — same
block number, same predecessor — and lets VRF tie-breaking and chain selection arbitrate. That
is the intended Praos behaviour: a slot battle is decided by the leader VRF value, not by who
applied a block locally first. Declining converts a coin-flip into a certain loss, and on a node
whose peers deliver rival blocks quickly it biases every contested slot against us.

Anti-equivocation is not the justification here: Dingo already has a dedicated, durable
mechanism for that in the last-forged-slot fence (#3734, `forger.go:1136` / main `1148`), and the
code comment at `forger.go:821–825` / main `830–834` already acknowledges the two are complementary. `dingo_metrics_slotBattlesTotal_int`
(`ledger/forging/metrics.go:142`) exists to count battles the node does not currently join.

### Suggested fix

Narrow Gate A to `currentSlot < tipSlot` (keep the stale-genesis `Error` branch and the
`forgeStaleGapThresholdSlots` check as they are), and let the `currentSlot == tipSlot` case fall
through to the leader check. When the node *is* the leader for a slot the chain already covers
with someone else's block, forge the alternative on the tip's predecessor. The existing fence at
`forger.go:1136` / main `1148` still prevents us from producing a second block for a slot **we** already signed
— which is the case that actually needs blocking.

If forging an alternative is considered too large a change, a smaller correct step is to skip
only when the block at `tipSlot` **is ours**.

## Contributing pattern (context, not part of the ask): epoch-boundary re-transition

Not a separate report, but it is how Gate B gets into its worst state. At the preview
1410 → 1411 boundary the node logged `handling epoch transition`
(`ledger/snapshot/manager.go:453`) **four consecutive times** for the same `new_epoch`, because
each rollback below the boundary re-triggered the boundary snapshot. Each re-run froze chain
extension for 1108 s / 870 s / 878 s / 908 s. During those freezes the local tip fell hundreds
to thousands of slots behind, so Gate B then legitimately fired for every following slot — the
node lost slots 121912324, 121912571 and 121913465 silently, and slot **121911023** was attempted
**8 m 02 s late** and failed with `forge check failed: transaction validation snapshot changed`
(the single `could_not_forge` in the window).

Two things would help independently of the gates: make the boundary snapshot idempotent /
skippable when it has already been computed for that `new_epoch`, and log at `Warn` when the same
`new_epoch` transition is handled more than once.

## Reproduction

Unit-level (fastest, no network):

- **Defect 2.** `ledger/forging/forger_test.go` already has the stub. Add a case to
  `forgerTestSlotClock` with `currentSlot: 10, chainTipSlot: 9, upstreamActive: true,
  upstreamTipSlot: 0` — i.e. exactly
  `TestCheckAndForgeProductionWaitsForUnknownActiveUpstreamTarget` (L136) — and observe that a
  node whose tip is **one slot** old does not forge. That test currently asserts the behaviour we
  are asking to change; the desired assertion is `builder.calls == 1`.
- **Defect 3.** Configure `currentSlot: 10, chainTipSlot: 10` with `forgerTestLeader{}` and assert
  `builder.calls == 1`. Today the forge is skipped at `forger.go:662` (main `663`) before the leader check.

Integration-level:

- **Defect 2.** Run a BP with two upstream peers and a leader slot due. Drop the connection to the
  current best peer ~1 s before the slot. The node forges nothing and logs nothing at INFO, while
  its own tip is current.
- **Defect 3.** Run two block producers scheduled for the same slot on a devnet with fast
  propagation, so the rival block applies within the slot. Dingo will decline; the Haskell node in
  the same position forges an alternative.

## Impact

On a testnet BP with an unstable peer set this is the dominant source of lost blocks — larger
than ghosting, ledger validation failures and forge errors combined in our data. It is also the
hardest to notice: the node reports itself healthy, `Forge_could_not_forge_int` stays at 0, and
`Forge_about_to_lead_int` keeps climbing, so every standard SPO dashboard shows a healthy
producer that is quietly declining its own leader slots.

chadaibbhot-git:
## Summary

`checkAndForgeProduction` has two skip gates that run **before**
`checkLeaderSafe`. Both `return nil`, neither moves a leadership counter, and both
log at `Debug`, so a scheduled leader slot they drop is indistinguishable at INFO
from "we were simply not the leader for that slot".

This PR fixes the narrower of the two problems and makes a dropped **leader** slot
visible in the log for both gates. It deliberately does **not** change the
upstream-sync gate's behaviour; that is a trade-off question for maintainers and is
documented here by a test rather than altered. It also deliberately adds **no new
metric** — see commit 4.

The tip gate is `currentSlot <= tipSlot`. The `<` half matches the reference
implementation. The `==` half does not, and it turns every contested slot into a
guaranteed loss: a rival pool that forges the same slot and wins the propagation
race puts our tip *at* the current slot, the gate matches, and we return before
ever asking whether we are the leader.

## Reference implementation

`ouroboros-consensus`, `mkCurrentBlockContext`
(`ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs`,
checked at `1820edf5e`) handles the three cases explicitly:

```haskell
  c' :> hdr -> case blockSlot hdr `compare` currentSlot of
    LT -> Right $ blockContextFromPrevHeader hdr
    GT -> Left $ TraceBlockFromFuture currentSlot (blockSlot hdr)
    -- The block at the tip has the same slot as the block we're going to
    -- produce (@currentSlot@).
    EQ ->
      Right $
        if isJust (headerIsEBB hdr)
          then blockContextFromPrevHeader hdr
          -- If @hdr@ is not an EBB, then forge an alternative to @hdr@: same
          -- block no and same predecessor.
          else BlockContext (blockNo hdr) $ castPoint $ AF.headPoint c'
```

`GT` declines — that is dingo's `<`. `EQ` forges an alternative and lets the leader
VRF and chain selection arbitrate. Dingo collapses `EQ` into the decline.

## What this PR does and does not do

It narrows the gate to `currentSlot < tipSlot` so the `EQ` case reaches leader
selection and is accounted for.

It does **not** yet forge the alternative block, and it does not pretend to. Two
things in the current code block that, and both are outside the forger:

- `DefaultBlockBuilder.BuildBlock` binds the parent to the live chain tip
  (`ledger/forging/builder.go`, `currentTip := b.chainTip.Tip()` feeding both
  `nextBlockNumber` and `prevHash`). A block forged at `EQ` would therefore name a
  parent whose slot equals its own, which `validateBlockOrder` rejects
  (`ledger/envelope_validation.go`: `block.SlotNumber() <= parent.slot` →
  `"block slot %d does not follow parent slot %d"`). Every Praos peer rejects it
  too, for the same reason.
- Binding the tip's **predecessor** instead — what `mkCurrentBlockContext` does —
  produces a block that does not extend the local tip, and `Chain.addBlockLocked`
  refuses that with `BlockNotFitChainTipError`. There is no local-fork adoption
  path for a self-forged block today.

So the alternative-block path needs a builder that can take an explicit parent and
a chain that can adopt a locally forged fork. **That is explicitly out of scope
here and will be addressed in a follow-up PR.** What this PR fixes is that the loss
was **silent**: the slot now reaches leader selection and, when we are the leader,
is counted as a leader slot, a slot battle and a could-not-forge, and logged at
`Warn`.

Maintainers who prefer to go straight to the full fix: the branch point is the same
one-line gate, and the two blockers above are the whole remaining scope.

## Commits

1. **`test(forging): pin the equal-slot tip gate as a contested slot`** — fails
   before commit 2. `TestCheckAndForgeProductionEqualSlotReachesLeaderCheck` asserts
   the contested slot reaches leader selection and is accounted for.
   `TestCheckAndForgeProductionEqualSlotDoesNotReForgeOurOwnSlot` is the
   anti-equivocation companion and passes before and after: after a successful
   forge the tip is our *own* block at that slot, so `tipSlot == currentSlot` there
   too, and that case must stay a quiet skip.

2. **`fix(forging): stop dropping contested slots before leader selection`** —
   `<=` → `<`, plus an explicit split of the two `EQ` cases. "The block at the tip
   is ours" keeps the existing quiet early skip, now keyed on the last-forged-slot
   fence rather than on the tip comparison alone. "The block at the tip is a
   rival's" runs leader selection and records the declined battle. Anti-equivocation
   is unaffected: the #3734 fence is still the mechanism that prevents a second
   block for a slot this node already signed for, and the contested-slot path
   returns before `reserveForgeSlot`, so a declined battle does not burn the slot.

3. **`test(forging): document the leader slot lost to an unknown upstream target`**
   — no behaviour change. `UpstreamSyncStatus()` returns `(0, true)` for the whole
   window between an active-connection switch and the new peer's first
   authenticated, admitted, trusted header (`publishActiveUpstream` stores the new
   connection key with `targetSlot` zero; only `publishAdmittedUpstreamTarget` makes
   it non-zero). The gate's `upstreamTip == 0` disjunct therefore disables forging
   outright during that window, independent of local tip freshness and independent
   of `forgeSyncToleranceSlots` — raising the tolerance does not affect this branch
   at all, which the test pins with a subtest at 100000.

   `TestForgeSkipsLeaderSlotWhenUpstreamTargetUnknownEvenAtTip` asserts the
   **current** behaviour, on a node whose tip is the previous slot's block. It is
   written that way on purpose: the behaviour is deliberate and already pinned by
   `TestCheckAndForgeProductionWaitsForUnknownActiveUpstreamTarget`, so this PR is
   not the place to change it. A `t.Skip` would have hidden the assertions from CI
   and let them rot; a passing test that fails the moment the gate is rekeyed on
   local tip freshness names the decision being revisited.

4. **`feat(forging): warn when a pre-leader gate drops a leader slot`** — additive,
   separable, no behaviour change. Take or drop it independently of 1–3.
   - A gate skip is raised from `Debug` to `Warn`, with `leader_slot=true`, when the
     slot is one this node is scheduled to lead. Both gates route their skip line
     through `logGateSkip`; the routine case is unchanged and still logs at `Debug`.
   - `isScheduledLeaderSlot` consults the precomputed VRF schedule via
     `LeaderChecker.NextLeaderSlot` (in `Election`, a read-locked scan of the cached
     schedule for the slot's epoch) rather than running leader selection, so it is
     cheap on a skip path. It only picks a log level, so it fails quiet — no cached
     schedule means the skip stays at `Debug` — and it recovers a panic from the
     pluggable checker for the same reason `checkLeaderSafe` does.
   - **No metric is added or changed.** `ledger/forging/metrics.go` is untouched by
     this branch, so every existing series and every `docs/dashboards/*.json` panel
     is bit-for-bit unaffected. This commit only decides the severity of a line that
     was already emitted. A counter for the pre-leader gates is a reasonable
     separate request (cf. #2873) and is left to maintainers.

5. **`test(forging): pin our own block at the tip as identified by hash`** — fails
   before commit 6. `TestEqualSlotOwnBlockIsIdentifiedByHashNotByFence` wires **no**
   fence store and asserts the forger starts with an unloaded fence, so the recorded
   forged hash is the only ownership signal left. Subtest one: the tip hash equals
   the block we forged for the slot, and the cycle must be a quiet `Debug` skip —
   no leader selection, no slot battle, no could-not-forge, no `Warn`. Subtest two
   is the companion that stops subtest one being satisfied by a blanket skip: the
   tracker holds a record for the slot but the tip carries a rival's hash, and that
   must still reach leader selection and count a slot battle.

6. **`fix(forging): identify our own tip block by hash, not by the fence alone`** —
   the skip is decided by `tipBlockIsOurs`, which prefers the `SlotTracker` hash
   match and falls back to the fence. The tip hash arrives through
   `ChainTipHashProvider`, a **separate optional** extension of `SlotClockProvider`
   rather than a new method on it, so `SlotClockProvider` implementations outside
   this repository keep compiling and a clock that does not implement it behaves
   exactly as before. No block is built, signed or broadcast on either side of this
   change — both outcomes return before `reserveForgeSlot` — so anti-equivocation is
   untouched.

7. **`docs(architecture): note could_not_forge now counts declined slot battles`** —
   documentation only. Adds the metrics note below to the block-forging section of
   `ARCHITECTURE.md` and corrects the chain-tip check quoted there, which this
   branch narrowed from `currentSlot <= tipSlot` to `currentSlot < tipSlot`.

Commits 8-10 answer the automated review of 5-7:

8. **`fix(forging): re-read the tip slot beside the tip hash`** — `tipSlot` is
   sampled at the top of the cycle and `ChainTipHash` is read later, so the hash
   need not belong to the slot being decided. The tip slot is now re-read next to
   the hash, and a tip that has moved off the slot is inconclusive and falls back
   to the fence. The reads are still not atomic, so this narrows the window rather
   than closing it.

9. **`fix(forging): report an equal-slot battle lost under the fence`** — when the
   fence covers the slot but the recorded forge hash shows a rival's block won it,
   the fence still refuses the forge: a second, different block for a slot whose
   first block may already have reached peers is equivocation, and losing a battle
   is not a licence to equivocate. What changes is that the loss is no longer
   reported at `Debug` as "slot already has our own block".
   `tipHashMatchesForgedBlock` becomes `tipBlockOwnership`, a tri-state — ours, a
   rival's, or indeterminate — and the rival-under-fence case is counted through
   `RecordSlotBattle` and `incCouldNotForge` and logged at `Warn` with both hashes.
   No branch of the equal-slot gate forges, so a wrong answer there can only choose
   between skipping quietly and skipping loudly.

10. **`docs(architecture): could_not_forge only partially overlaps slot battles`** —
    corrects the note added in commit 7, which told operators to subtract
    `dingo_metrics_slotBattlesTotal_int`. That counter is also raised from
    `ledger/chainsync.go` outside the forge path, so the overlap is partial and an
    exact subtraction under-counts.

11. **`fix(forging): let chainsync own the slot-battle count for a lost equal-slot battle`** —
    the rival-at-tip-under-the-fence path no longer calls `RecordSlotBattle()`:
    `LedgerState.checkSlotBattle` already recorded that battle when the rival block
    was accepted at our slot (same `SlotTracker`, same counter, wired in
    `node_lifecycle.go`), so the forger's increment was always the second one.
    `Forge_could_not_forge_int` and the Warn stay. The no-fence contested branch
    keeps its count: no fence means no tracker record, so chainsync returns early there.

## How tested

Go 1.26.7, `GOFLAGS=-mod=mod`, base `main` @ `429dbdbb`.

- `go build ./...` — clean.
- `go vet ./ledger/forging/...` — clean.
- `golangci-lint run ./ledger/forging/...` — 0 issues.
- `gofmt -l ledger/forging/` — clean. (`golines` is not installed here; no added
  line exceeds 80 columns except log string literals, which `golines` does not
  split and for which this file already has precedent.)
- Before commit 2, at commit 1:
  `go test ./ledger/forging/... -run TestCheckAndForgeProductionEqualSlot` →
  `--- FAIL: TestCheckAndForgeProductionEqualSlotReachesLeaderCheck`
  (leader selection call count 0, expected 1; `Forge_node_is_leader_int`,
  `slotBattlesTotal` and `could_not_forge` all 0, expected 1). The companion
  anti-equivocation test passes at commit 1.
- After commit 2: both pass.
- `go test ./ledger/forging/... -count=1` — all pass, including the pre-existing
  fence, bootstrap-gate, KES and upstream-target tests.
- `go test ./ledger/forging/... ./ledger/ ./chain/... -count=1` — all pass.

## Notes for review

- The `EQ`-is-ours early skip was originally keyed on the fence alone (`fenceLoaded
  && currentSlot <= lastForgedSlot`). Review pointed out that the fence answers a
  different question — it is a high-water mark over slots this node *committed to*,
  and it is durable only through a `ForgeFenceStore`, so where none is wired
  `lastForgedSlot` is in-memory only. Commits 5–7 rekey the skip on identity:
  `SlotTracker` already records the hash of every block this node forged, so the
  skip now compares that hash against the hash at the tip, which proves whether the
  block there is ours or a rival's. The fence is kept as an additional guard for the
  window the tracker cannot cover — `reserveForgeSlot` writes the fence *before*
  signing while `RecordForgedBlock` runs only after adoption — and for slot clocks
  that cannot report a tip hash. The skip line now reports which signal decided it,
  via `matched_by`. Neither signal survives a restart on its own, so this narrows
  the window rather than closing it; the durable half is still the fence store.
- After this change the tip gate only fires when the tip is strictly past the
  current slot, so its log line was reworded from "slot already has block" to
  "chain tip is ahead of the current slot".
- Nothing here changes what is signed or diffused. The only new signing-adjacent
  behaviour is that a contested slot now runs the KES gate and leader selection
  before declining, which is where cardano-node also spends that work.

**Known window (pre-existing).** `tipSlot` is read once at the top of
`checkAndForgeProduction`, before the KES gate and before `checkLeaderSafe`, and the
chain can move while leader selection runs. A rival block that lands at
`currentSlot` inside that window is invisible to the `currentSlot == tipSlot`
comparison further down, because that comparison is evaluated against the
already-sampled `tipSlot`. The slot therefore escapes the contested-slot branch,
`reserveForgeSlot` succeeds, and `BuildBlock` binds the parent to the *live* tip —
which by then is the rival's block at the same slot. The result is a block naming a
parent whose slot equals its own, rejected by `validateBlockOrder` after the header
has already been signed. This window is pre-existing: it is neither introduced,
widened nor narrowed by this PR, and no code here addresses it. Re-reading the tip
inside the branch would only shrink it, not close it, since the tip can still move
between that read and `BuildBlock`. The follow-up explicit-parent builder — the
one described under *What this PR does and does not do* — is what closes it: a
builder that takes an explicit parent re-derives it against the tip it actually
built on, rather than against a value sampled earlier in the cycle.

**Metrics note.** `cardano_node_metrics_Forge_could_not_forge_int` now also counts
leader slots declined because a rival's block already occupies the slot. Before this
PR it meant "we were the leader and the local build or sign failed"; it now includes
a case that is neither, and that case is exactly the one that also increments
`dingo_metrics_slotBattlesTotal_int`, so the two series overlap. No metric is added,
removed or renamed, and no dashboard query changes. Anyone alerting on the rate of
`Forge_could_not_forge_int` should inspect `dingo_metrics_slotBattlesTotal_int`
alongside it. The overlap is only *partial* and the two are **not** a clean
decomposition of one another: `ledger/chainsync.go` also raises
`dingo_metrics_slotBattlesTotal_int` for battles detected outside the forge path,
so subtracting it exactly would under-count real could-not-forge events. The same
note is in the block-forging section of `ARCHITECTURE.md` (commits 7 and 10).

Closes #3954 — defect 3, the tip gate collapsing `EQ` into a decline.
Refs #3954 — defect 1 (observability) is addressed in the log only, at `Warn`, with
no new metric; defect 2 (the unknown upstream target) is documented by a test but
deliberately not changed here.

---
## Summary by cubic
Fixes the forger dropping contested leader slots before leader selection, so a rival block at the current slot is now counted as a slot battle and `Warn`-logged instead of silently skipped; pre-leader gate skips that swallow scheduled leader slots warn too.

- A tip at the current slot now reaches leader selection; ownership of the tip block (ours vs. rival) is decided by hash via the optional `ChainTipHashProvider`, with the forge fence as fallback, and our own committed slot still skips quietly.
- Pre-leader gate skips (chain tip ahead, stale-gap, upstream-sync) mark scheduled leader slots with `leader_slot=true`; ordinary skips stay at `Debug`.
- `Forge_could_not_forge_int` now also counts declined rival-block battles; its partial overlap with `dingo_metrics_slotBattlesTotal_int` is documented in `ARCHITECTURE.md`.
- The upstream-sync gate behavior is unchanged and pinned by tests; forging a valid alternative block remains unsupported and is left for a follow-up.

<sup>Written for commit 3f53a6018bbb0944d4fa472e2f0057b20b1bc119. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Forging now handles chain tips at the current slot correctly, allowing leader selection for contested slots.
  - Prevents re-forging slots already recorded by the local forge fence.
  - Skips forging when the upstream target is unknown, even when the node is at the chain tip.
  - Records unsupported rival-block conflicts as skipped slot battles.
  - Sync-gate warnings now distinguish scheduled leader slots from ordinary skipped slots.

### DIFF
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -677,10 +677,12 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 
 	tipSlot := f.slotClock.ChainTipSlot()
 
-	// Skip if a block already exists at the current slot.
-	// When tipSlot >= currentSlot, the chain already has a block at
-	// this slot (possibly from a peer). Producing another would create
-	// a competing block and fork the chain.
+	// Skip if the chain has already moved PAST the current slot.
+	// A tip beyond currentSlot means any block we produced would fork
+	// the chain below its own tip. A tip AT currentSlot is the
+	// contested case and is handled after leader selection below:
+	// ouroboros-consensus mkCurrentBlockContext declines only for GT
+	// and treats EQ as a slot battle.
 	// Count every slot check (matches cardano-node
 	// Forge.about_to_lead)
 	if f.metrics != nil {
@@ -688,10 +690,10 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 		f.metrics.tipGapSlots.Set(0)
 	}
 
-	if currentSlot <= tipSlot {
+	if currentSlot < tipSlot {
 		// Detect stale data: if the tip is far ahead of the slot clock,
 		// the database likely contains chain data from a different genesis.
-		// Use subtraction (safe here since tipSlot >= currentSlot from
+		// Use subtraction (safe here since tipSlot > currentSlot from
 		// the outer check) to avoid uint64 overflow on the addition.
 		gap := tipSlot - currentSlot
 		if f.metrics != nil {
@@ -709,11 +711,27 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 			)
 		} else {
 			f.logger.Debug(
-				"forge skip: slot already has block",
+				"forge skip: chain tip is ahead of the current slot",
 				"current_slot", currentSlot,
 				"tip_slot", tipSlot,
 			)
 		}
+		return nil
+	}
+
+	// The tip is at the current slot and the block there is one this
+	// node already committed to forging, so the slot-aligned loop has
+	// simply re-entered a slot it already used. Leave without running
+	// leader selection: this is not a contested slot, and the fence
+	// below would refuse it anyway.
+	if currentSlot == tipSlot && f.fenceLoaded &&
+		currentSlot <= f.lastForgedSlot {
+		f.logger.Debug(
+			"forge skip: slot already has our own block",
+			"current_slot", currentSlot,
+			"tip_slot", tipSlot,
+			"last_forged_slot", f.lastForgedSlot,
+		)
 		return nil
 	}
 
@@ -880,6 +898,32 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 	// the pre-selection gate.
 	if f.metrics != nil {
 		f.metrics.forgeNodeIsLeader.Inc()
+	}
+
+	// A rival block already occupies this leader slot. This is a slot
+	// battle, not a reason to treat the slot as spent: the reference
+	// implementation forges an alternative here (same block number, the
+	// tip's predecessor as parent) and lets the leader VRF and chain
+	// selection arbitrate.
+	//
+	// Dingo cannot build that alternative yet. BlockBuilder binds the
+	// parent to the live chain tip, so a block forged now would name a
+	// parent whose slot equals its own and be rejected by
+	// ledger.validateBlockOrder; binding the tip's predecessor instead
+	// needs a block that does not extend the local tip, which
+	// chain.addBlockLocked refuses. Until that path exists, record the
+	// battle we are declining rather than dropping the slot silently.
+	if currentSlot == tipSlot {
+		if f.metrics != nil {
+			f.metrics.slotBattlesTotal.Inc()
+		}
+		f.incCouldNotForge()
+		f.logger.Warn(
+			"forge skip: leader slot already holds another block; forging an alternative is not supported",
+			"current_slot", currentSlot,
+			"tip_slot", tipSlot,
+		)
+		return nil
 	}
 
 	// Commit to this slot before any signing happens for it, including
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -710,7 +710,8 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 				gap,
 			)
 		} else {
-			f.logger.Debug(
+			f.logGateSkip(
+				currentSlot,
 				"forge skip: chain tip is ahead of the current slot",
 				"current_slot", currentSlot,
 				"tip_slot", tipSlot,
@@ -755,7 +756,8 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 				float64(gap),
 			)
 		}
-		f.logger.Debug(
+		f.logGateSkip(
+			currentSlot,
 			"chain syncing from peer, skipping forge",
 			"current_slot", currentSlot,
 			"tip_slot", tipSlot,
@@ -1317,6 +1319,48 @@ func (f *BlockForger) checkOpCertSequence(
 		opCert.IssueNumber,
 		!limits.era.isTPraos(),
 	)
+}
+
+// logGateSkip logs a slot dropped by a gate that runs before leader
+// selection. Such skips are routine and stay at Debug, but one that
+// swallows a slot this node was scheduled to lead is a lost block, and
+// nothing downstream will ever mention that slot again, so it is raised
+// to Warn.
+func (f *BlockForger) logGateSkip(
+	slot uint64,
+	msg string,
+	attrs ...any,
+) {
+	if f.isScheduledLeaderSlot(slot) {
+		f.logger.Warn(msg, append(attrs, "leader_slot", true)...)
+		return
+	}
+	f.logger.Debug(msg, attrs...)
+}
+
+// isScheduledLeaderSlot reports whether slot is one this node is
+// scheduled to lead, by consulting the precomputed VRF leader schedule
+// rather than running leader selection. Election.NextLeaderSlot is a
+// read-locked scan of the cached schedule for the slot's epoch, so it
+// is cheap enough to call on a skip path.
+//
+// It only decides a log level, so it fails quiet: a checker with no
+// cached schedule for that epoch reports false and the skip stays at
+// Debug. A panic in the pluggable checker is recovered for the same
+// reason checkLeaderSafe recovers one — it must not take down the
+// producer-loop goroutine.
+func (f *BlockForger) isScheduledLeaderSlot(slot uint64) (scheduled bool) {
+	defer func() {
+		if r := recover(); r != nil {
+			scheduled = false
+			f.reportForgeCallbackPanic("schedule", r)
+		}
+	}()
+	if f.leaderChecker == nil {
+		return false
+	}
+	next, ok := f.leaderChecker.NextLeaderSlot(slot)
+	return ok && next == slot
 }
 
 // checkLeaderSafe calls the pluggable LeaderChecker, recovering any
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -700,14 +700,27 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 			f.metrics.tipGapSlots.Set(float64(gap))
 		}
 		if gap > f.forgeStaleGapThresholdSlots {
-			f.logger.Error(
-				"chain tip is far ahead of slot clock; database may contain data from a different genesis",
+			// This gate also runs before leader selection, so it
+			// swallows a scheduled leader slot as silently as the
+			// skips routed through logGateSkip. The stale-genesis
+			// diagnosis stays at Error, but carry the same
+			// leader_slot marker so a lost block is still
+			// attributable. Reuse isScheduledLeaderSlot rather
+			// than repeating the schedule lookup.
+			attrs := []any{
 				"current_slot",
 				currentSlot,
 				"tip_slot",
 				tipSlot,
 				"slot_gap",
 				gap,
+			}
+			if f.isScheduledLeaderSlot(currentSlot) {
+				attrs = append(attrs, "leader_slot", true)
+			}
+			f.logger.Error(
+				"chain tip is far ahead of slot clock; database may contain data from a different genesis",
+				attrs...,
 			)
 		} else {
 			f.logGateSkip(
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -748,22 +748,60 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 		return nil
 	}
 
-	// The tip is at the current slot and the block there is one this
-	// node produced, so the slot-aligned loop has simply re-entered a
-	// slot it already used. Leave without running leader selection:
-	// this is not a contested slot, and the fence below would refuse
-	// it anyway.
+	// The tip is at the current slot. Before treating that as a
+	// contested slot, work out whether the block sitting there is one
+	// this node produced.
 	if currentSlot == tipSlot {
-		if ours, matchedBy := f.tipBlockIsOurs(currentSlot); ours {
+		ownership, ourHash, tipHash := f.tipBlockOwnership(currentSlot)
+		fenceCovers := f.fenceLoaded && currentSlot <= f.lastForgedSlot
+		switch {
+		case ownership == tipOwnershipOurs:
+			// The slot-aligned loop has simply re-entered a slot
+			// whose block is provably ours. Not a contested slot,
+			// and the fence would refuse it anyway.
 			f.logger.Debug(
 				"forge skip: slot already has our own block",
 				"current_slot", currentSlot,
 				"tip_slot", tipSlot,
 				"last_forged_slot", f.lastForgedSlot,
-				"matched_by", matchedBy,
+				"matched_by", "forged_block_hash",
+			)
+			return nil
+		case ownership == tipOwnershipRival && fenceCovers:
+			// We committed to this slot and a rival's block is what
+			// the chain adopted for it. The fence still forbids
+			// forging: a second, different block for a slot whose
+			// first block may already have reached peers is
+			// equivocation, and losing a battle is not a licence to
+			// equivocate. But this is a slot battle we lost, not
+			// "the tip block is ours", and dropping it at Debug is
+			// exactly the silent loss this change exists to remove.
+			f.RecordSlotBattle()
+			f.incCouldNotForge()
+			f.logger.Warn(
+				"slot battle lost: rival block at tip for a slot this node already forged",
+				"current_slot", currentSlot,
+				"tip_slot", tipSlot,
+				"last_forged_slot", f.lastForgedSlot,
+				"our_block_hash", hex.EncodeToString(ourHash),
+				"tip_block_hash", hex.EncodeToString(tipHash),
+			)
+			return nil
+		case fenceCovers:
+			// Ownership is inconclusive, but the fence says this
+			// node already committed to the slot, so it is not
+			// available regardless of who holds the tip.
+			f.logger.Debug(
+				"forge skip: slot already has our own block",
+				"current_slot", currentSlot,
+				"tip_slot", tipSlot,
+				"last_forged_slot", f.lastForgedSlot,
+				"matched_by", "forge_fence",
 			)
 			return nil
 		}
+		// Neither signal claims the slot: fall through and let leader
+		// selection and the contested-slot branch account for it.
 	}
 
 	// Skip if the chain is still syncing from a peer.
@@ -1303,67 +1341,85 @@ func (f *BlockForger) reserveForgeSlot(slot uint64) (bool, error) {
 	return true, nil
 }
 
-// tipBlockIsOurs reports whether the block occupying slot at the chain
-// tip is one this node produced, and which signal decided it.
+// tipOwnership is the outcome of comparing the block sitting at the
+// chain tip against the block this node forged for a given slot.
+type tipOwnership int
+
+const (
+	// tipOwnershipUnknown means the comparison could not be made: no
+	// recorded hash for the slot, a slot clock that cannot report a tip
+	// hash, an empty hash on either side, or a tip that moved between
+	// the two reads. The caller must fall back to the fence.
+	tipOwnershipUnknown tipOwnership = iota
+	// tipOwnershipOurs means the tip block is byte-for-byte the block
+	// this node forged for the slot.
+	tipOwnershipOurs
+	// tipOwnershipRival means the tip holds a different block for a slot
+	// this node forged for: a slot battle this node lost.
+	tipOwnershipRival
+)
+
+// tipBlockOwnership reports whether the block at the chain tip for slot
+// is the one this node forged, a rival's, or indeterminate, together
+// with the two hashes it compared (nil when indeterminate).
 //
-// The primary signal is identity, not bookkeeping: SlotTracker already
-// records the hash of every block this node forged and adopted, so a tip
-// hash equal to that record proves the block at the tip is ours rather
-// than a rival's. That check is independent of whether a durable
-// ForgeFenceStore is wired, which matters because the fence is in-memory
-// only when it is not (see fenceStore above).
+// This is identity rather than bookkeeping. SlotTracker already records
+// the hash of every block this node forged and adopted, so comparing it
+// against the hash at the tip distinguishes "we re-entered our own slot"
+// from "we lost this slot to a rival". The forge fence cannot make that
+// distinction: it is a high-water mark over slots this node committed
+// to, so it reports both cases identically, and it is durable only
+// through a ForgeFenceStore, so where none is wired (see fenceStore
+// above) it can be silent about a slot this node demonstrably forged.
 //
-// The fence is kept as an additional guard, not replaced. It covers the
-// window the tracker cannot: reserveForgeSlot writes the fence *before*
-// the header for a slot is signed, while RecordForgedBlock runs only
-// after the block is adopted, so between those two points the slot is
-// already ours but has no recorded hash yet. The fence is also the only
-// signal available when the wired slot clock does not implement
-// ChainTipHashProvider.
+// The fence is not replaced by this, and the caller must keep consulting
+// it. It covers the window the tracker cannot: reserveForgeSlot writes
+// the fence *before* the header for a slot is signed, while
+// RecordForgedBlock runs only after adoption, so between those two
+// points the slot is already ours but has no recorded hash yet. More
+// importantly, a fence that covers the slot forbids forging even when
+// this function answers tipOwnershipRival — losing a slot battle is not
+// a licence to sign a second block for a slot whose first block may
+// already have reached peers.
 //
 // Neither signal survives a process restart on its own: the tracker is
-// in-memory, and the fence is durable only through a ForgeFenceStore.
-func (f *BlockForger) tipBlockIsOurs(slot uint64) (bool, string) {
-	if f.tipHashMatchesForgedBlock(slot) {
-		return true, "forged_block_hash"
-	}
-	if f.fenceLoaded && slot <= f.lastForgedSlot {
-		return true, "forge_fence"
-	}
-	return false, ""
-}
-
-// tipHashMatchesForgedBlock reports whether the current chain tip is the
-// exact block this node forged for slot. It answers false whenever any
-// input is missing (no tracker record, no tip-hash capable slot clock,
-// an empty hash on either side) or the tip has moved off slot between
-// the two reads, so the caller falls back to the fence.
-func (f *BlockForger) tipHashMatchesForgedBlock(slot uint64) bool {
+// in-memory, and the fence is durable only through a store.
+//
+// The result can only choose between skipping quietly and skipping
+// loudly; no branch of the equal-slot gate forges, so a wrong answer
+// here cannot produce a block.
+func (f *BlockForger) tipBlockOwnership(
+	slot uint64,
+) (tipOwnership, []byte, []byte) {
 	if f.slotTracker == nil {
-		return false
+		return tipOwnershipUnknown, nil, nil
 	}
 	forgedHash, ok := f.slotTracker.WasForgedByUs(slot)
 	if !ok || len(forgedHash) == 0 {
-		return false
+		return tipOwnershipUnknown, nil, nil
 	}
 	provider, hasTipHash := f.slotClock.(ChainTipHashProvider)
 	if !hasTipHash {
-		return false
+		return tipOwnershipUnknown, nil, nil
 	}
 	tipHash := provider.ChainTipHash()
 	if len(tipHash) == 0 {
-		return false
+		return tipOwnershipUnknown, nil, nil
 	}
 	// The caller's tipSlot was sampled at the top of the forge cycle and
 	// the chain can move underneath it, so this hash need not belong to
 	// the slot being decided. Re-read the tip slot next to the hash and
-	// refuse to conclude anything if the tip is no longer at this slot,
-	// which sends the caller back to the fence. The two reads are still
-	// not atomic, so this narrows the window rather than closing it.
+	// refuse to conclude anything if the tip is no longer at this slot.
+	// The two reads are still not atomic, so this narrows the window
+	// rather than closing it; every remaining disagreement resolves to
+	// tipOwnershipUnknown or to a Warn, never to a forge.
 	if f.slotClock.ChainTipSlot() != slot {
-		return false
+		return tipOwnershipUnknown, nil, nil
 	}
-	return bytes.Equal(tipHash, forgedHash)
+	if bytes.Equal(tipHash, forgedHash) {
+		return tipOwnershipOurs, forgedHash, tipHash
+	}
+	return tipOwnershipRival, forgedHash, tipHash
 }
 
 func (f *BlockForger) incCouldNotForge() {
--- a/ledger/forging/forger.go
+++ b/ledger/forging/forger.go
@@ -776,7 +776,17 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
 			// equivocate. But this is a slot battle we lost, not
 			// "the tip block is ours", and dropping it at Debug is
 			// exactly the silent loss this change exists to remove.
-			f.RecordSlotBattle()
+			//
+			// slotBattlesTotal is deliberately NOT incremented here.
+			// Reaching this case means a block other than ours was
+			// accepted at a slot SlotTracker says we forged, and the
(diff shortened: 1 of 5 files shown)


ledger/forging/forger.go
```diff
@@ -827,6 +860,32 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
...
+	// tip's predecessor as parent) and lets the leader VRF and chain
+	// selection arbitrate.
+	//
+	// Dingo cannot build that alternative yet. BlockBuilder binds the
+	// parent to the live chain tip, so a block forged now would name a
+	// parent whose slot equals its own and be rejected by
+	// ledger.validateBlockOrder; binding the tip's predecessor instead
+	// needs a block that does not extend the local tip, which
+	// chain.addBlockLocked refuses. Until that path exists, record the
+	// battle we are declining rather than dropping the slot silently.
```
wolf31o2, line 995: `tipSlot` is read once at the top of the function, and `checkLeaderSafe` runs a VRF evaluation between that read and this branch. The chain can move in the gap, so the `EQ` decision is made against a stale tip in both directions.

The direction this branch does not cover: at the tip read `currentSlot > tipSlot`, so no gate fires; a rival block for `currentSlot` lands during leader selection; this branch does not match; `reserveForgeSlot` succeeds and `DefaultBlockBuilder.BuildBlock` binds the parent to the *live* tip (`ledger/forging/builder.go:267`), which is now the rival at `currentSlot`. The result names a parent whose slot equals its own and is refused by `validateBlockOrder` (`ledger/envelope_validation.go:87`) — the same rejection this branch exists to avoid, reached by a different route, and after signing rather than before it.

The window is pre-existing: `currentSlot <= tipSlot` had it too. Worth stating in the follow-up's scope, since the explicit-parent builder that fixes the `EQ` case is also what would let this one re-derive against the tip it actually built on.

ledger/forging/forger.go
```diff
@@ -827,6 +860,32 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
...
+	if currentSlot == tipSlot {
+		if f.metrics != nil {
+			f.metrics.slotBattlesTotal.Inc()
```
wolf31o2, line 1009: `incCouldNotForge` now fires on every contested leader slot, which changes what `forge_could_not_forge` means rather than only adding to it. Before this PR the counter moved on build and signing failures — conditions with something wrong locally. A slot battle is a normal Praos outcome and, on Preview, a routine one.

The accounting itself is defensible; the surprise is for anything alerting on the counter's rate. `slotBattlesTotal` gives an operator the decomposition, so a line in the PR description or the metrics doc saying the two now overlap would be enough.

ledger/forging/forger.go
```diff
@@ -649,46 +649,78 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
...
 		}
 		return nil
 	}
 
+	// The tip is at the current slot and the block there is one this
+	// node already committed to forging, so the slot-aligned loop has
+	// simply re-entered a slot it already used. Leave without running
+	// leader selection: this is not a contested slot, and the fence
+	// below would refuse it anyway.
+	if currentSlot == tipSlot && f.fenceLoaded &&
```
wolf31o2: This guard needs `fenceLoaded`, and `forger.go:85` notes the fence is in-memory only when no durable store is wired. In that configuration a slot the loop re-enters after our own successful forge falls through to the branch at `:875` and is recorded as a slot battle, at `Warn`, against our own block.

Harmless for forging — it still declines — but it is the one path where the new Warn reports a rival that does not exist. Comparing the tip's hash against the block this node forged, rather than gating on the fence, would close it; so would saying in the log that the fence was unavailable.

wolf31o2 (commented):
The reference reading is right and the scope discipline is right. No blockers. Three observations inline.

Verified:

- `mkCurrentBlockContext`'s `EQ` case is a forge, not a decline, so `currentSlot <= tipSlot` did collapse a slot battle into a guaranteed loss.
- **No equivocation risk.** The `EQ` branch returns above `reserveForgeSlot`, so nothing signs and the fence is not advanced. Even with `fenceLoaded` false — where the early own-block guard at `forger.go:713` cannot fire — the slot still terminates at that branch rather than reaching a build.
- `Election.NextLeaderSlot` compares `slot >= fromSlot` (`ledger/leader/election.go:1119`), so `next == slot` in `isScheduledLeaderSlot` is reachable and the Warn escalation is live rather than dead.
- `slotBattlesTotal` already exists (`ledger/forging/metrics.go:42`), and `forgeNodeIsLeader` is incremented before the branch, which matches `about_to_lead` / `node_is_leader` ordering.
- `TestCheckAndForgeProductionEqualSlotReachesLeaderCheck` fails at `b9ff1dd8` with "a leader slot we did not produce for must be visible as a could-not-forge, not as silence", and the package passes at `c9ac6663`.
- `lastForgedSlot` and `fenceLoaded` are read without a lock here, matching the existing reads at `forger.go:1208` and the write at `:1229`. Not introduced.

wolf31o2 (dismissed):

wolf31o2 (dismissed):

chadaibbhot-git:
Rebased onto main 90504583; no functional change. One adjacency conflict in `ledger/forging/forger.go` (upstream `checkOpCertSequence` from #3926 and this branch's `logGateSkip` were inserted at the same spot; both kept, upstream first). Net patch is line-identical before and after; all commits build individually, suites and mutation checks re-run green.

chadaibbhot-git:
CI note: the three `ledger/eras` failures on this head (`TestConwayUtxoValidationRuleIndexesArePinned`, `TestValidateTxConwayGenuinelyMissingInputStillRejected`, `TestValidateTxConwayGenuinelyUnbalancedStillRejected`) are inherited from main 90504583 — see #3976. Every other package passes; this branch does not touch `ledger/eras`.

chadaibbhot-git:
Rebased onto main 6b6346e3 (picks up #3977, which fixes the inherited `ledger/eras` failures); no conflicts, no functional change — net patch line-identical. Suites, race and mutation checks re-run green.

wolf31o2 (approved):

Issue #3746, opened 2026-09-01
ci: Windows suite runtime is unpredictable and 10x Linux on ./database


wolf31o2:
## Problem

The Windows `go test ./...` suite is the slowest and least predictable of the
three platforms, and its runtime is not reproducible run to run. The same
command on the same runner label:

| Run | Trigger | `go-test` step | Outcome |
|---|---|---|---|
| 33519632627 | push to main | 17m44s | pass |
| 33529462211 | pull request #3744 | 42m+ | still running when filed |

Slowest packages from the 17m44s run, against the current `-timeout 30m`
per-package bound:

| Package | Windows | Linux, same suite |
|---|---|---|
| `./database` | 737s | 71s |
| `internal/test/conformance` | 681s | 917s with real Postgres and MySQL attached, 86s SQLite-only on macOS |
| `database/plugin/metadata/sqlstore/migration` | 446s | — |
| `./ledger` | 334s | 79s |
| `internal/koiosparity` | 249s | 62s |
| `database/lifecycle` | 173s | 31s |

`./database` is 10x its Linux time and `database/lifecycle` 5.6x. macOS runs the
same suite in about 4 minutes, so this is specific to Windows rather than to
non-Linux hosts.

Separately, the suite has a history of wedging rather than failing, which is why
Windows was removed from the pull-request path in #3070. The four consecutive
main-branch failures in runs 33413291419, 33422972126, 33440696014 and
33502559781 were `internal/test/devnet`, fixed by the `linux` build constraint
in #3732, so those are not the cause of the remaining slowness.

## Current state

`go-test (Windows)` runs the full suite on pull requests and on pushes to main
as of #3744. It is deliberately **not** a required branch-protection context and
must not become one while the runtime behaves this way. A wedged run is bounded
by `timeout-minutes: 90` at the job level and `-timeout 30m` per package, the
latter being the one that produces a goroutine dump naming a stuck test.

## Acceptance criteria

- Establish whether the run-to-run spread is runner variance or a test that
  intermittently blocks, using the per-package output of several runs.
- Attribute the `./database` and `database/lifecycle` Windows-to-Linux ratios to
  specific tests, and determine whether the cost is inherent to Windows
  filesystem semantics or avoidable in the test setup.
- Either bring the Windows runtime within a predictable band, or scope the
  pull-request job to the packages carrying Windows-specific code paths
  (`mithril` extraction, `keystore` ACLs, `database` handle-relative rename and
[...]

PR #3959, opened 2026-09-05
fix(forging): forge an alternative block on an equal-slot battle

chadaibbhot-git:
## Summary

PR #3955 narrowed the tip gate so a contested slot reaches leader selection,
and made the declined battle visible. It stopped there on purpose: the slot was
still conceded, because two things outside the forger made the alternative
impossible to build and impossible to adopt. This PR builds those two things and
forges the alternative.

Depends on #3955 and is stacked on it. It is five commits, each independently
reviewable.

## Rebase status

Rebased onto `upstream/main` at `680e65e5` (`ec7166b3` -> `1598f7b3`, 15 commits).
Two files conflicted; both resolutions keep both behaviours.

- `ledger/envelope_validation_test.go` — add/add at end of file against #3881's
  new Byron proof/size helpers. Kept both, upstream's helpers first.
- `ledger/forging/forger.go` — upstream's equal-slot branch is the concede path
  this PR replaces, so the code is ours. Of the comment above it, the paragraph
  saying "Dingo cannot build that alternative yet ... `chain.addBlockLocked`
  refuses" is exactly what this PR removes and was dropped; the paragraph
  explaining why counting `slotBattlesTotal` here does not double up with
  `LedgerState.checkSlotBattle` still describes live code and was carried over
  verbatim.
- `ledger/forging/forger_test.go` (two hunks) — #4123 added
  `require.Equal(t, []uint64{9}, leiosCerts.markedSlots)` beside the
  `parent.calls` assertion this PR changes from 1 to 2. Kept both.

One non-conflict fix the new base forces: gouroboros v0.204.0 widened
`babbage.BabbageOpCert.SequenceNumber` from `uint32` to `uint64`, so the
`newSiblingTestBlock` helper in `ledger/local_sibling_test.go` no longer
compiled. Widened to `uint64`, matching upstream's own
`chainsync_praos_dijkstra_test.go` helper, and squashed into the commit that
introduced the file. Nothing else of this PR was dropped: no commit in
`31dd4a8a..680e65e5` implements equal-slot alternative forging — upstream's own
comment on the concede path still said "Dingo cannot build that alternative yet"
at `680e65e5`.

Range-diff: 11 of 15 patches byte-identical; the four that differ are the four
changes listed above.

## Reference implementation

`ouroboros-consensus`, `mkCurrentBlockContext`
(`ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs`,
checked at `1820edf5e`):

```haskell
  c' :> hdr -> case blockSlot hdr `compare` currentSlot of
    LT -> Right $ blockContextFromPrevHeader hdr
    GT -> Left $ TraceBlockFromFuture currentSlot (blockSlot hdr)
    -- The block at the tip has the same slot as the block we're going to
    -- produce (@currentSlot@).
    EQ ->
      Right $
        if isJust (headerIsEBB hdr)
          then blockContextFromPrevHeader hdr
          -- If @hdr@ is not an EBB, then forge an alternative to @hdr@: same
          -- block no and same predecessor.
          else BlockContext (blockNo hdr) $ castPoint $ AF.headPoint c'
```

`BlockContext (blockNo hdr) (headPoint c')` is the whole design: the block
number of the block at the tip, and the point of the chain *without* that block
— i.e. its predecessor. The forged block is then a sibling of the rival, not a
descendant of it, and consensus resolves the two by ordinary chain selection.

## The two blockers, and what replaces them

**1. The builder derived parentage from the live tip.**
`DefaultBlockBuilder.BuildBlock` read `b.chainTip.Tip()` and used it for both
`prevHash` and `blockNumber + 1`. At an equal slot the live tip *is* the rival at
that slot, so the block would name a parent whose slot equals its own —
`ledger.validateBlockOrder` rejects it ("block slot %d does not follow parent
slot %d") and so does every Praos peer.

`BlockContext` makes the parentage explicit and mirrors the reference type. A nil
context is the unchanged live-tip path; an explicit context names a parent point
and a block number. `BuildBlockOnContext` is the entry point,
`AlternativeBlockBuilder` the capability interface.

Two guards come with it, both before any signing:

- The context carries the tip it was derived from. If the live tip has moved by
  build time, the contest is over and the candidate is abandoned. (This is the
  existing mid-assembly `errParentChangedDuringBuild` check, hoisted to the start
  for the contested path.)
- The resolved parent must sit strictly below the forged slot —
  `errParentSlotNotBelowBlock`. This is new on the **live-tip path too**: a
  builder asked to forge at a slot the tip already holds previously produced the
  invalid block described above and only found out after KES signing. It now
  fails first.

**2. There was no local-fork adoption path.**
`Chain.addBlockLocked` is extend-only, so a block that names the tip's
predecessor is refused with `BlockNotFitChainTipError`.

Dingo already resolves exactly this situation — for peers. A header that does not
fit the tip reaches `LedgerState.tryResolveFork`, which compares it to the local
tip with `praos.ComparePraosTips` and, if it wins, rolls chain and ledger back to
the common ancestor and adopts the winner. What it cannot take is a block that
arrived through the forger: `tryResolveFork` is keyed on a `ChainsyncEvent`, a
peer connection, that peer's retained header history and a blockfetch restart.

`LedgerState.AdoptLocalForgedSibling` is the same arbitration and the same
mutation, entered from the forge loop:

- **Decision:** `praos.ComparePraosTips` on the two tips and their select views,
  read from the real block headers. No preference of any kind for the local
  block. `ComparePraosTips` answers `ChainEqual` whenever no reference
  implementation rule applies — unarmed tiebreaker, unresolvable select view,
  identical VRF — and `ChainEqual` keeps the incumbent. A locally forged block
  must therefore win on the same evidence a peer's would need, and a tie changes
  nothing.
- **Mutation:** `rollbackChainAndStateDeferred` to the shared parent, then
  `Chain.AddLocalBlockDeferred`, under `chainsyncMutex` — the same sequence and
  the same lock `tryResolveFork` uses. The rollback is always exactly one block
  deep and always to the block both candidates already name as parent.

Siblinghood is checked, never assumed: the candidate must name the tip's parent
as its own, carry the tip's block number, sit strictly above the fork point, and
not already be the tip; and the *applied* ledger tip must have caught up to the
chain tip, so a rollback can never rewind past what the pipeline actually
applied. Anything else is `ErrNotChainTipSibling`, before anything is rolled back.

Losing is a normal outcome, not an error. The rival keeps the slot, and the
forged block is discarded without ever being diffused.

## Fail-closed by construction

Both new providers are optional on `ForgerConfig`. `alternativeBlockContext`
declines the contested slot when *any* of these holds: `ChainContext` unset,
`SiblingAdopter` unset, a `BlockBuilder` that is not an
`AlternativeBlockBuilder`, a tip with no resolvable predecessor, a tip no longer
at the contested slot, or a predecessor not strictly below it.
`Chain.TipPredecessor` itself reports `ok=false` for an empty chain, a tip that
is the chain's first block, an unresolvable predecessor, or any disagreement
between the two block records and the advertised tip.

A node with none of this wired behaves exactly as it did after #3955: the battle
is counted and logged at `Warn`, and nothing is signed. That path stays pinned by
`TestCheckAndForgeProductionEqualSlotDeclinesWithoutTheAlternativePath`.

Declining costs one block. Guessing a parent costs a signature over a block no
peer will accept.

## Equivocation safety

Unchanged, and now covered on this path too.

- `reserveForgeSlot` (the #3734 fence) still runs before *any* signing for the
  slot, which is before the alternative is built.
  `TestEqualSlotAlternativeReservesTheFenceBeforeSigning` asserts the fence is
  already written at the moment the builder is invoked.
- A slot whose tip is our **own** block still takes the quiet early skip and
  never reaches leader selection, with the alternative path fully wired
  (`TestEqualSlotAlternativeRefusesASlotWeAlreadyForged`).
- **Losing keeps the slot burned.** The block was signed; a second signature for
  the same slot is equivocation regardless of which block the network kept.
- A declined battle still returns before `reserveForgeSlot`, so it does not burn
  the slot.

## Leios: alternatives are plain ranking blocks

`leiosBlockDataForSlot` picks a certificate by matching eligible endorser blocks
against `ParentLeiosAnnouncement()`, which resolves "the parent" from the live
tip. That is correct for every uncontested slot and wrong for an alternative,
whose parent is the tip's *predecessor*: the certificate would attest to an
endorser block announced by a ranking block the alternative does not descend
from.

Rather than teach the prototype's announcement lookup about explicit block
contexts, an alternative is forged with no Leios data at all — no certificate,
and no new announcement either, so a contested slot cannot also introduce an
endorser block whose announcing chain may be the one that loses. The
parent-announcement provider is not consulted on this path. Uncontested slots are
untouched. This is a deliberate limitation of the alternative path and is pinned
by a test.

## Commits

1. **`feat(chain): expose the alternative-block context and a deferred local
   add`** — `Chain.TipPredecessor` (the reference implementation's EQ context,
   fail-closed) and `Chain.AddLocalBlockDeferred` (`AddLocalBlock` with deferred
   publication, because adoption runs under `chainsyncMutex` where an inline
   `chain.update` publish is the drain deadlock documented on
   `AddBlockWithPointDeferred`). No existing path changes.

2. **`feat(forging): let the builder forge on an explicit block context`** —
   `BlockContext`, `BuildBlockOnContext`, `AlternativeBlockBuilder`, the
   stale-contest guard and `errParentSlotNotBelowBlock`. Also adds
   `TestValidateBlockOrderPinsTheEqualSlotAlternativeShape`, which pins both
   halves of the validator rule the builder is now obeying.

3. **`feat(ledger): adopt a locally forged sibling through Praos chain
   selection`** — `AdoptLocalForgedSibling`, `ErrNotChainTipSibling`, and the
   preference-free decision helper.

4. **`feat(forging): forge an alternative block on an equal-slot battle`** — the
   forger wiring, `alternativeBlockContext`, `adoptSiblingBlockSafe`, and the
   flipped expectation on
   `TestCheckAndForgeProductionEqualSlotReachesLeaderCheck`.

5. **`feat(node): wire equal-slot alternative forging into the block producer`** —
   the primary chain and `LedgerState` satisfy the two provider interfaces
   directly, with compile-time assertions so a signature change fails the build
   rather than silently reverting producers to conceding contested slots.

6. **`fix(forging): forge an equal-slot alternative as a plain ranking block`** —
   the Leios limitation above.

## How tested

Go 1.26.7, `GOFLAGS=-mod=mod`.

Re-run in full on the current base `680e65e5`.

- `go build ./...` — clean.
- `go vet ./ledger/... ./chain/... .` — clean.
- `gofmt -l` — clean over every file this PR touches. (`chain/chain_test.go` is
  gofmt-unclean at `680e65e5` itself; inherited, not touched here.)
- `golangci-lint run ./ledger/... ./chain/...` — **0 issues**.
- `go test ./ledger/forging/... ./chain/... ./ledger/ . -count=1` — all pass
  (`ledger/forging` 2.6s, `chain` 15.1s, `ledger` 38.3s, root 9.2s).
- `go test -race ./ledger/forging/... ./chain/... ./ledger/` — all pass
  (24.9s / 19.4s / 205.9s).

All eight mutation checks re-run on the rebased tree; each went red and was then
restored from a copy-aside, with `git status` clean after every one:
the parent-slot-order guard (`TestBuildBlockRefusesAParentAtItsOwnSlot`),
mempool exclusion (`TestBuildBlockOnContextCarriesNoMempoolTransactions`),
the Leios-data refusal (`TestBuildBlockOnContextRejectsLeiosData`),
the durable-fence precondition (`TestEqualSlotAlternativeRequiresADurableFence`),
sibling slot equality (`TestAdoptLocalForgedSiblingRejectsNonSiblings`),
range progress on apply rather than arrival
(`TestStaleBlockfetchBatchIsDiscardedAfterLocalSiblingAdoption`),
the chain admission predicate
(`TestAddBlockWithPointDeferredIfRefusesWithoutMutating`, and
`TestAddBlockWithPointDeferredIfEvaluatesAdmitUnderTheChainMutex` which does not
merely fail but never completes), and the Leios drop when the parent moves
(`TestCheckAndForgeProductionDropsLeiosDataWhenTheParentMoves`).

Worth recording for #6: adding an extra arrival-side call while leaving the
apply-side one in place does **not** go red — a superseded batch is dropped by
the drain's fast path before the loop body runs. Only the full revert (call back
on arrival, removed from apply) exercises the invariant, and that is what was
run.

New tests:

- **chain** (`chain/tip_predecessor_test.go`) — the EQ context on a real chain;
  fail-closed on an empty chain and on a first block; and the rollback-then-adopt
  sequence, including an assertion that `AddLocalBlock` refuses the sibling first,
  so the test cannot pass vacuously.
- **builder** (`ledger/forging/builder_block_context_test.go`) — the
  alternative's shape (parent is the predecessor, block number is the rival's,
  *not* the rival's plus one); all three ways to name a parent that is not below
  the block, including the live-tip path (this is the negative test that we never
  build a same-slot parent); a stale contest; a context with no resolved parent;
  and that the default path still extends the live tip.
- **ledger** (`ledger/local_sibling_test.go`) — wire-decodable Conway blocks so
  the select views are real, covering **both** outcomes end to end: ours adopted
  with the fork point intact behind it, and the rival left alone when it wins.
  Plus identical views not preferring ours, every non-sibling shape, and the
  decision function's length-first and unarmed-tiebreaker cases.
- **forger** (`ledger/forging/forge_alternative_test.go`) — a losing alternative
  is not published and does not move `forgeAdopted` but keeps the fence burned;
  the fence is reserved before signing; a slot we already forged is refused with
  the path wired; five unwired/unresolvable configurations all decline without
  burning the slot; a panicking adopter is recovered; and the Leios omission.
- **envelope validation** (`ledger/envelope_validation_test.go`) —
  `TestValidateBlockOrderPinsTheEqualSlotAlternativeShape`.

### CI on `1598f7b3`

Green: `go-test (Linux, race)` (19m30s), `go-test (Windows)` (10m9s),
`docker`, `lint`, `govulncheck`, `Conventional Commits`, `DCO`.

Two red jobs, both the same pre-existing flake pair and neither touched by this
PR:

- `go-test (Linux)` — `TestCanonicalPeersWithCrossingFrontiersDoNotFlap`
  (`chainselection`), "only the initial selection may publish a chain switch".
- `go-test (macOS)` — `TestCanonicalFrontierCrossingDoesNotCloseAPeerAheadOfLocalTip`
  (`ledger/chainsync_frontier_flap_test.go`), "the scenario must publish at
  least the initial selection event".

Reproduced on clean `upstream/main` at `680e65e5` with nothing from this branch
applied, same assertions and same messages:

```
go test ./chainselection/ -run TestCanonicalPeersWithCrossingFrontiersDoNotFlap -count=300
  33/300 FAIL
go test ./ledger/ -run TestCanonicalFrontierCrossingDoesNotCloseAPeerAheadOfLocalTip -count=300
  36/300 FAIL
```

Both are the #4048 frontier-flap tests, fixed in #4144. Nothing to fix on this
branch; stating it once rather than repeating it per run.

## Open questions for maintainers

These are the parts I would most want a second opinion on.

1. **Should this be behind a config flag?** It changes what a block producer
   signs in a situation it previously declined. It is the reference
   implementation's behaviour and I have left it always-on for a wired producer,
   but a `--forge-equal-slot-alternative` opt-in would be easy to add if you
   would rather ship it dark first.

2. **The rollback/add pair is not atomic against a concurrent blockfetch add.**
   `AdoptLocalForgedSibling` holds `chainsyncMutex`, which serializes it against
   chainsync header handling — the same lock `tryResolveFork` holds — but the
   blockfetch drain is serialized by `chainsyncBlockfetchMutex` instead. If a
   fetched block lands between the rollback and the add, the add fails and the
   chain is left at the fork point with neither candidate on it; that is logged
   at `Error` and recovers when chainsync re-offers the rival's header (which no
   longer conflicts). This is the same window class the existing peer path has —
   `tryResolveFork` rolls back and then *waits for blockfetch*, a much wider one —
   so I did not introduce a new lock-order edge to close it. Taking both mutexes
   would be a new composition across `drainBlockPipelineBeforeRollback`, and I
   did not want to add an edge to that lock graph without a maintainer's read.

3. **Post-adoption resync churn.** `ls.rollback` publishes a "local ledger
   rollback" `ChainsyncResyncEvent`, which drives `RecoverAfterLocalRollback`.
   After a won battle the primary chain tip is *above* the rollback point, so
   that path takes its `SkipConnectionClose` branch and connections are not
   closed; peer header history is replayed, the rival's header arrives again, and
   `tryResolveFork` re-runs the same comparison and declines it. Converges, but
   it is one rollback event per won slot battle and I would like that confirmed
   as acceptable.

4. **`ComparePraosTips` returning `ChainEqual` is a decline.** This means a node
   whose incumbent tip block cannot be decoded, or whose era has no armed VRF
   tiebreaker, never adopts its own alternative. That is the conservative
   direction and matches "keep the incumbent", but it does mean the capability is
   silently inert in those cases.

Refs #3954 (the declined-battle finding). Completes what #3955 explicitly
scoped out.

---
## Summary by cubic
Forges a competing sibling block when a rival already occupies the slot this node is about to forge, so the node wins via ordinary Praos chain selection instead of conceding the slot. Previously, contested slots never reached leader selection; now they do, and a won battle rolls back to the shared parent and adopts the local block.

**Key changes**
- Alternatives are built on the rival's predecessor with the rival's block number via `BlockContext`/`BuildBlockOnContext`; a guard rejects any parent not strictly below the forged slot before signing.
- `AdoptLocalForgedSibling` arbitrates with `praos.ComparePraosTips` and rolls back/adopts when the local block wins; losing discards it undiffused and is not an error.
- Both providers are optional and fail closed, so unwired nodes concede exactly as before with `Warn` logged; the durable forge fence is required without it.
- Alternatives carry no transactions or Leios data; the tip gate narrowed to `currentSlot < tipSlot` so contested slots reach leader selection.
- A won adoption invalidates in-flight blockfetch batches for the rival's segment atomically under the chain mutex, and blockfetch range progress is noted on apply, not arrival.

<sup>Written for commit 198c82bd9c28030c455eada1ce8544bfd0779344. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3959?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for forging and evaluating alternative blocks when competing blocks are produced at the same slot.
  * Winning alternatives can now be adopted through normal chain-selection rules; losing alternatives are not published.
  * Alternative blocks exclude unsupported transaction and announcement data.

* **Bug Fixes**
  * Prevented stale block-fetch results from being applied after a rollback.
  * Improved handling and visibility of forging skips, including scheduled leader slots.

### DIFF
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -329,7 +329,7 @@ func (c *Chain) AddBlock(
 	block ledger.Block,
 	txn *database.Txn,
 ) error {
-	evt, err := c.addBlockInternal(block, ocommon.Point{}, txn, true, false)
+	evt, err := c.addBlockInternal(block, ocommon.Point{}, txn, true, false, nil)
 	if err != nil {
 		return err
 	}
@@ -356,6 +356,7 @@ func (c *Chain) AddLocalBlock(block ledger.Block) error {
 		nil,
 		false,
 		false,
+		nil,
 	)
 	if err != nil {
 		return err
@@ -385,6 +386,7 @@ func (c *Chain) AddLocalBlockDeferred(
 		nil,
 		false,
 		true,
+		nil,
 	)
 }
 
@@ -396,7 +398,7 @@ func (c *Chain) AddBlockWithPoint(
 	point ocommon.Point,
 	txn *database.Txn,
 ) error {
-	evt, err := c.addBlockInternal(block, point, txn, true, false)
+	evt, err := c.addBlockInternal(block, point, txn, true, false, nil)
 	if err != nil {
 		return err
 	}
@@ -426,7 +428,33 @@ func (c *Chain) AddBlockWithPointDeferred(
 	point ocommon.Point,
 	txn *database.Txn,
 ) (event.Event, error) {
-	return c.addBlockInternal(block, point, txn, true, true)
+	return c.addBlockInternal(block, point, txn, true, true, nil)
+}
+
+// AddBlockWithPointDeferredIf is AddBlockWithPointDeferred with a caller
+// supplied admission predicate, evaluated while the chain mutex is held and
+// before any chain state is read or written. When admit reports false the add
+// is abandoned and ErrBlockAddNotAdmitted is returned; nothing is mutated and
+// no event is queued.
+//
+// It exists because a precondition checked by the caller before calling is
+// not a precondition of the mutation. The blockfetch drain holds
+// chainsyncBlockfetchMutex, not the chainsyncMutex the equal-slot alternative
+// adoption holds, so a rollback can publish a new generation and truncate the
+// chain in the window between the drain testing its batch and this add taking
+// the chain mutex. Passing the test as a predicate closes that window: the
+// generation comparison and the tip mutation it guards become one atomic step
+// under the same lock every other chain mutation takes.
+//
+// admit must not acquire the chain or manager mutex, and must not block: it
+// runs on the chain's own write path.
+func (c *Chain) AddBlockWithPointDeferredIf(
+	block ledger.Block,
+	point ocommon.Point,
+	txn *database.Txn,
+	admit func() bool,
+) (event.Event, error) {
+	return c.addBlockInternal(block, point, txn, true, true, admit)
 }
 
 // queueDeferredEventLocked appends evt to the chain-level sequencer. The caller
@@ -483,6 +511,7 @@ func (c *Chain) addBlockInternal(
 	txn *database.Txn,
 	matchPendingHeader bool,
 	deferred bool,
+	admit func() bool,
 ) (event.Event, error) {
 	if c == nil {
 		return event.Event{}, errors.New("chain is nil")
@@ -492,6 +521,14 @@ func (c *Chain) addBlockInternal(
 	// We get a write lock on the manager to cover the integrity checks and adding the block below
 	c.manager.mutex.Lock()
 	defer c.manager.mutex.Unlock()
+	// Ask the caller, under the same lock that serializes every chain
+	// mutation, whether this add is still wanted. Evaluating it here is the
+	// point: a caller that checks its own precondition before calling has
+	// already released whatever it inspected, so a rollback landing in
+	// between would leave that check stale. See AddBlockWithPointDeferredIf.
+	if admit != nil && !admit() {
+		return event.Event{}, ErrBlockAddNotAdmitted
+	}
 	// Verify chain integrity
 	if err := c.reconcile(); err != nil {
 		return event.Event{}, fmt.Errorf("reconcile chain: %w", err)
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -143,6 +143,61 @@ func (c *Chain) HeaderTip() ochainsync.Tip {
 	return c.headerTip()
 }
 
+// TipPredecessor reports the block context an alternative to the current chain
+// tip must be built on: the tip itself (the block that would be competed with)
+// and the point of the tip's immediate predecessor, which becomes the
+// alternative block's parent.
+//
+// This is the data ouroboros-consensus' mkCurrentBlockContext returns for its
+// EQ case, where a leader whose slot is already occupied forges an alternative
+// with the same block number as the occupant and the occupant's predecessor as
+// parent (Ouroboros/Consensus/NodeKernel/Forge.hs).
+//
+// ok is false whenever that context cannot be established: an empty chain, a
+// tip that is the first block on the chain (its parent is genesis, which is not
+// a rollback target here), or a predecessor the chain can no longer resolve.
+// Callers must treat that as "no alternative can be built" and must not fall
+// back to the live tip: binding the tip itself as parent produces a block whose
+// parent slot equals its own, which envelope validation and every Praos peer
+// reject.
+func (c *Chain) TipPredecessor() (
+	parent ocommon.Point,
+	tip ochainsync.Tip,
+	ok bool,
+) {
+	if c == nil {
+		return ocommon.Point{}, ochainsync.Tip{}, false
+	}
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	c.manager.mutex.RLock()
+	defer c.manager.mutex.RUnlock()
+	// initialBlockIndex is the first block; it has no predecessor on this
+	// chain, and neither does an empty chain.
+	if c.tipBlockIndex <= initialBlockIndex {
+		return ocommon.Point{}, ochainsync.Tip{}, false
+	}
+	tipBlock, err := c.blockByIndexLocked(c.tipBlockIndex)
+	if err != nil {
+		return ocommon.Point{}, ochainsync.Tip{}, false
+	}
+	parentBlock, err := c.blockByIndexLocked(c.tipBlockIndex - 1)
+	if err != nil {
+		return ocommon.Point{}, ochainsync.Tip{}, false
+	}
+	// Cross-check the two records against each other and against the tip this
+	// chain advertises. The chain maintains all three invariants, but a caller
+	// of this function is about to sign a block against the answer, so a
+	// disagreement is reported as "no context" rather than resolved.
+	if !bytes.Equal(tipBlock.Hash, c.currentTip.Point.Hash) ||
+		!bytes.Equal(tipBlock.PrevHash, parentBlock.Hash) {
+		return ocommon.Point{}, ochainsync.Tip{}, false
+	}
+	return ocommon.NewPoint(parentBlock.Slot, parentBlock.Hash),
+		c.currentTip,
+		true
+}
+
 // blockNumberContiguous reports whether a block number legitimately follows its
 // parent's. Shelley-era and later increment by exactly one per block. Byron
 // epoch-boundary blocks reuse the parent's block number (they do not increment),
@@ -296,6 +351,28 @@ func (c *Chain) AddLocalBlock(block ledger.Block) error {
 		c.eventBus.Publish(ChainUpdateEventType, evt)
 	}
 	return nil
+}
+
+// AddLocalBlockDeferred adds a locally forged block exactly like AddLocalBlock
+// but, instead of publishing the resulting chain.update inline, enqueues it on
+// the chain-level sequencer under c.mutex and returns it. This is the entry
+// point for a forged block adopted from a path that holds a ledger mutex --
+// specifically the equal-slot alternative, which rolls the chain back to the
+// contested block's parent under chainsyncMutex and then adopts the local
+// sibling. Publishing inline from there is the drain deadlock described on
+// AddBlockWithPointDeferred; enqueuing under c.mutex also keeps this add
+// ordered behind the rollback that preceded it. A returned event with an empty
+// Type means there is nothing to publish.
+func (c *Chain) AddLocalBlockDeferred(
+	block ledger.Block,
+) (event.Event, error) {
+	return c.addBlockInternal(
+		block,
+		ocommon.Point{},
+		nil,
+		false,
+		true,
+	)
 }
 
 // AddBlockWithPoint adds a block using a caller-supplied point. This avoids
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -170,8 +170,15 @@ func (c *Chain) TipPredecessor() (
 	}
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
-	c.manager.mutex.RLock()
-	defer c.manager.mutex.RUnlock()
+	// Both lookups below go through blockByIndexLocked, which resolves a
+	// common-prefix index against the primary chain when this chain is an
+	// in-memory fork. That read needs the primary-chain lock as well as the
+	// manager lock, in the chain -> primary -> manager order this helper
+	// establishes; taking manager.mutex directly would let a concurrent
+	// primary-chain mutation move the common prefix between the two lookups
+	// and yield a tip and a parent that never sat next to each other.
+	unlockBlockIndexReadLocks := c.lockBlockIndexReadLocks()
+	defer unlockBlockIndexReadLocks()
 	// initialBlockIndex is the first block; it has no predecessor on this
 	// chain, and neither does an empty chain.
 	if c.tipBlockIndex <= initialBlockIndex {
@@ -189,8 +196,14 @@ func (c *Chain) TipPredecessor() (
 	// chain advertises. The chain maintains all three invariants, but a caller
 	// of this function is about to sign a block against the answer, so a
 	// disagreement is reported as "no context" rather than resolved.
+	//
+	// The parent must also sit strictly below the tip's slot. That is the
+	// ordering contract this function's doc comment states, and the caller
+	// signs against the answer: a parent whose slot is at or above the
+	// contested slot produces a block Praos and envelope validation reject.
 	if !bytes.Equal(tipBlock.Hash, c.currentTip.Point.Hash) ||
-		!bytes.Equal(tipBlock.PrevHash, parentBlock.Hash) {
+		!bytes.Equal(tipBlock.PrevHash, parentBlock.Hash) ||
+		parentBlock.Slot >= tipBlock.Slot {
 		return ocommon.Point{}, ochainsync.Tip{}, false
 	}
 	return ocommon.NewPoint(parentBlock.Slot, parentBlock.Hash),
--- a/ledger/forging/builder.go
+++ b/ledger/forging/builder.go
@@ -193,6 +193,11 @@ func (b *DefaultBlockBuilder) BuildBlock(
 
 // BuildBlockOnContext creates a new block on an explicitly named parent
 // instead of the live chain tip. See BlockContext.
+//
+// leios must be empty: an equal-slot alternative is a plain ranking block, and
+// a non-empty value is rejected rather than carried. The block also carries no
+// mempool transactions, because no validator here can select them against the
+// state at blockCtx.Parent. Both are explained at their guards in buildBlock.
 func (b *DefaultBlockBuilder) BuildBlockOnContext(
 	slot uint64,
 	kesPeriod uint64,
@@ -256,6 +261,15 @@ var errParentSlotNotBelowBlock = errors.New(
 	"parent slot is not below the forged slot",
 )
 
+// errLeiosDataOnAlternative indicates Leios data was supplied together with an
+// explicit block context. An equal-slot alternative is a plain ranking block:
+// see the guard in buildBlock for why certificate and announcement data
+// resolved against the live tip cannot be carried by a block that does not
+// build on it.
+var errLeiosDataOnAlternative = errors.New(
+	"an alternative block cannot carry leios data",
+)
+
 // tipsEqual reports whether two chain tips reference the same point and
 // block number. Slot and hash are both required: a rollback can restore a
 // prior slot, and two forged blocks never share a hash.
@@ -316,6 +330,18 @@ func (b *DefaultBlockBuilder) buildBlock(
 	// ouroboros-consensus mkCurrentBlockContext's EQ branch, which forges
 	// "an alternative to @hdr@: same block no and same predecessor".
 	if blockCtx != nil {
+		// An alternative is a plain ranking block. Leios data is resolved
+		// against "the parent", which the providers answer from the live tip
+		// -- here the rival, which is precisely the block this one does not
+		// build on. A certificate selected that way certifies an endorser
+		// block announced by a chain this block is not on, and a new
+		// announcement would introduce an endorser block whose announcing
+		// chain may be the one that loses the battle. The forger already
+		// omits both; refuse here too so the exported entrypoint cannot be
+		// called into that state.
+		if !leios.empty() {
+			return nil, nil, errLeiosDataOnAlternative
+		}
 		// The candidate only makes sense against the tip it was derived
 		// from. If the chain has already moved, the contest is over;
 		// abandon before doing any work rather than after signing.
@@ -404,6 +430,28 @@ func (b *DefaultBlockBuilder) buildBlock(
 	)
 
 	mempoolTxs := b.mempool.Transactions()
+	if blockCtx != nil {
+		// An alternative is built on the rival's predecessor, but every
+		// transaction validator reachable from here answers against the
+		// ledger's live state -- which has the rival applied. A mempool
+		// transaction spending a UTxO the rival created passes that check
+		// and would be selected, and adoption then rolls the rival back
+		// before applying this block: the transaction's input no longer
+		// exists, so applying our own adopted block fails and the node is
+		// left wedged at the fork point.
+		//
+		// Selecting against a validation snapshot taken at blockCtx.Parent
+		// is the fix, but no validator exposes one: LedgerState's
+		// WithTxValidationSession and ValidateTxWithOverlay both read the
+		// current UTxO set, and the overlay can only add pending state, not
+		// un-apply the block at the tip. Until such a snapshot exists, fail
+		// closed and forge the alternative with no transactions at all. An
+		// empty ranking block is valid against the parent state whatever
+		// the rival did, and it still carries the VRF and opcert that
+		// decide the battle, which is the whole purpose of forging it. The
+		// transactions stay in the mempool for the next block.
+		mempoolTxs = nil
+	}
 	if leiosCert != nil {
 		// A prototype CertRB carries the Leios certificate and no Dijkstra
 		// transactions; node-to-client later inlines the certified EB txs.
--- a/ledger/forging/builder.go
+++ b/ledger/forging/builder.go
@@ -180,15 +180,28 @@ func NewDefaultBlockBuilder(
 	}, nil
 }
 
-// BuildBlock creates a new block for the given slot.
-// Returns the block and its CBOR encoding.
+// BuildBlock creates a new block for the given slot, extending the live chain
+// tip. Returns the block and its CBOR encoding.
 func (b *DefaultBlockBuilder) BuildBlock(
 	slot uint64,
 	kesPeriod uint64,
 ) (ledger.Block, []byte, error) {
 	generation := b.creds.acquireCredentialGeneration()
 	defer generation.release()
-	return b.buildBlock(slot, kesPeriod, LeiosBlockData{}, generation)
+	return b.buildBlock(slot, kesPeriod, LeiosBlockData{}, generation, nil)
+}
+
+// BuildBlockOnContext creates a new block on an explicitly named parent
+// instead of the live chain tip. See BlockContext.
+func (b *DefaultBlockBuilder) BuildBlockOnContext(
+	slot uint64,
+	kesPeriod uint64,
+	leios LeiosBlockData,
+	blockCtx BlockContext,
+) (ledger.Block, []byte, error) {
+	generation := b.creds.acquireCredentialGeneration()
+	defer generation.release()
+	return b.buildBlock(slot, kesPeriod, leios, generation, &blockCtx)
 }
 
 // BlockForger.buildBlock discovers the Leios capability with a runtime type
@@ -198,6 +211,7 @@ func (b *DefaultBlockBuilder) BuildBlock(
 // this one is loud, but it is the same class of defect the guard prevents.
 var (
 	_ LeiosBlockBuilder                = (*DefaultBlockBuilder)(nil)
+	_ AlternativeBlockBuilder          = (*DefaultBlockBuilder)(nil)
 	_ credentialGenerationBlockBuilder = (*DefaultBlockBuilder)(nil)
 )
 
@@ -210,7 +224,7 @@ func (b *DefaultBlockBuilder) BuildBlockWithLeios(
 ) (ledger.Block, []byte, error) {
 	generation := b.creds.acquireCredentialGeneration()
 	defer generation.release()
-	return b.buildBlock(slot, kesPeriod, leios, generation)
+	return b.buildBlock(slot, kesPeriod, leios, generation, nil)
 }
 
 func (b *DefaultBlockBuilder) buildBlockWithCredentialGeneration(
@@ -218,8 +232,9 @@ func (b *DefaultBlockBuilder) buildBlockWithCredentialGeneration(
 	kesPeriod uint64,
 	leios LeiosBlockData,
 	generation *credentialGeneration,
+	blockCtx *BlockContext,
 ) (ledger.Block, []byte, error) {
-	return b.buildBlock(slot, kesPeriod, leios, generation)
+	return b.buildBlock(slot, kesPeriod, leios, generation, blockCtx)
 }
 
 // errParentChangedDuringBuild indicates the chain tip moved while
@@ -229,6 +244,16 @@ func (b *DefaultBlockBuilder) buildBlockWithCredentialGeneration(
 // adoption would reject anyway once it re-checks the parent.
 var errParentChangedDuringBuild = errors.New(
 	"selected parent changed during block assembly",
+)
+
+// errParentSlotNotBelowBlock indicates the resolved parent does not sit
+// strictly below the slot being forged. Praos requires strictly increasing
+// slots and ledger.validateBlockOrder enforces it, so such a block would be
+// rejected after signing -- by this node and by every peer. The live-tip path
+// hits this exactly when the tip already holds the slot being forged, which is
+// the contested case an explicit BlockContext exists to serve.
+var errParentSlotNotBelowBlock = errors.New(
+	"parent slot is not below the forged slot",
 )
 
 // tipsEqual reports whether two chain tips reference the same point and
@@ -245,6 +270,7 @@ func (b *DefaultBlockBuilder) buildBlock(
 	kesPeriod uint64,
 	leios LeiosBlockData,
 	credentials *credentialGeneration,
+	blockCtx *BlockContext,
 ) (ledger.Block, []byte, error) {
 	// Keep the protocol lifetime guard inside the generation-backed path so
 	// both exported builder entrypoints and BlockForger fail before reading the
@@ -266,10 +292,14 @@ func (b *DefaultBlockBuilder) buildBlock(
 	// Get current chain tip
 	currentTip := b.chainTip.Tip()
 
+	// Resolve the block context: the parent this block names and the block
+	// number it carries. Default (blockCtx nil) is to extend the live tip.
+	parentPoint := currentTip.Point
+
 	// Block numbers are 0-indexed in Cardano: the first block after
 	// genesis is BlockNo 0. When the tip is genesis (empty hash), the
 	// chain has no blocks yet so the next block number is 0.
-	isGenesis := len(currentTip.Point.Hash) == 0
+	isGenesis := len(parentPoint.Hash) == 0
 
 	var nextBlockNumber uint64
 	if !isGenesis {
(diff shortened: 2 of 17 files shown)


ledger/forging/forger.go
```diff
@@ -983,7 +1155,34 @@ func (f *BlockForger) checkAndForgeProduction(_ context.Context) error {
...
-	if addErr := f.addBlockSafe(block, blockCbor); addErr != nil {
+	//
+	// An alternative does not extend the tip, so it goes to chain selection
+	// instead: it is adopted only if it beats the block already there, by
+	// the same comparison a peer's competing block goes through. Losing is
+	// an outcome, not a failure -- the rival keeps the slot and this block
+	// is never diffused, so the observer, mempool cleanup and adoption
+	// counters below must not run for it.
+	if altBlockContext != nil {
+		adopted, addErr := f.adoptSiblingBlockSafe(block, blockCbor)
```
wolf31o2, line 1350: Blocking: when the alternative wins during an in-flight blockfetch batch, this path rolls back the primary chain without invalidating the queued batch. A later flush can append blocks fetched for the superseded header queue and re-extend the abandoned chain. Publish a rollback generation before truncation and have the flush reject batches from older generations.

ledger/forging/builder.go
```diff
@@ -180,15 +180,28 @@ func NewDefaultBlockBuilder(
...
+// instead of the live chain tip. See BlockContext.
+func (b *DefaultBlockBuilder) BuildBlockOnContext(
+	slot uint64,
+	kesPeriod uint64,
+	leios LeiosBlockData,
+	blockCtx BlockContext,
+) (ledger.Block, []byte, error) {
+	generation := b.creds.acquireCredentialGeneration()
+	defer generation.release()
+	return b.buildBlock(slot, kesPeriod, leios, generation, &blockCtx)
```
wolf31o2, line 209: Blocking: this alternative is validated against the rival tip even though adoption first rolls that tip back to blockCtx.Parent. A transaction spending a UTxO created by the rival can therefore be included in a block that is invalid after rollback. Build the alternative against the parent-state snapshot or reject it when that validator is unavailable.

chain/chain.go
```diff
@@ -114,6 +114,61 @@ func (c *Chain) HeaderTip() ochainsync.Tip {
...
+	}
+	// Cross-check the two records against each other and against the tip this
+	// chain advertises. The chain maintains all three invariants, but a caller
+	// of this function is about to sign a block against the answer, so a
+	// disagreement is reported as "no context" rather than resolved.
+	if !bytes.Equal(tipBlock.Hash, c.currentTip.Point.Hash) ||
+		!bytes.Equal(tipBlock.PrevHash, parentBlock.Hash) {
+		return ocommon.Point{}, ochainsync.Tip{}, false
+	}
+	return ocommon.NewPoint(parentBlock.Slot, parentBlock.Hash),
```
wolf31o2, line 209: Blocking: TipPredecessor documents and tests a strictly earlier parent slot, but this return path only checks hash linkage. Reject a parent whose slot is greater than or equal to the contested tip slot before handing the context to the signer.

wolf31o2 (changes requested):

chadaibbhot-git:
Pushed through 6c21a5d addressing the three blocking items, cubic's findings, and CodeRabbit's.

- **Blockfetch race:** `LedgerState.blockfetchRollbackGeneration` is bumped in `AdoptLocalForgedSibling` before the truncation; every queued batch records the generation at request time (in `startQueuedBlockfetchLockedWithWaitSignal`, the single funnel for queued ranges); `flushPendingBlockfetchBlocksDeferred` discards a superseded batch and releases its deferred-validation bookkeeping. The new test drives the window between the truncation and `AddLocalBlockDeferred`, where the tip is the fork point and a stale block chains onto it cleanly (in the steady state the tip-fit check already rejects such blocks). Residual: the generation load is not atomic with the chain add under `c.mutex`. Closing that fully means threading the expected generation into the chain add, or holding `chainsyncBlockfetchMutex` across the pipeline drain, which I rejected as the same lock-inversion shape as #3717. Happy to do the former in this PR if you want it.
- **Parent-state validation:** there is no parent-state validator (`WithTxValidationSession` and `ValidateTxWithOverlay` read live state; the overlay cannot un-apply the block at the tip). Rather than decline the alternative whenever the mempool is non-empty, it is now built with no mempool transactions: an empty ranking block is valid against the parent whatever the rival did and still carries the VRF/opcert that decide the battle; the txs stay in the mempool. If you would rather hard-decline in that case, it is a one-line change.
- **TipPredecessor:** slot ordering added (`parent.Slot >= tip.Slot` returns not-ok) and the lock order fixed via `lockBlockIndexReadLocks()`.
- **cubic:** sibling slot equality in `local_sibling.go`, Leios data rejected on the alternative path inside `buildBlock`, test fixes. Range progress (`noteBlockfetchRangeProgress`) now moves from block arrival to the successful chain-apply branch, so a discarded or declined block no longer clears an unavailable-range record; that matters here because an unservable queued header is exactly the post-slot-battle state the sibling adoption produces. The `TipPredecessor` slot guard cubic flagged was already in 2fafc2f3 (`parent.Slot >= tip.Slot` returns not-ok, covered by `TestTipPredecessorRefusesAParentAtOrAboveTheContestedSlot`).
- **CodeRabbit:** `alternativeBlockContext` now declines when no durable `ForgeFenceStore` is wired, so a fence-less embedder restarted inside a contested slot cannot sign a sibling to its own block. Production wiring in `node_forging.go` already refuses to start without the fence.

`go build`, `go vet`, `gofmt`, `golangci-lint` clean; `go test ./ledger/... ./chain/... .` passes. Each new guard was mutation-checked (test fails with the guard reverted).

wolf31o2 (dismissed):

chadaibbhot-git:
Rebased onto main 90504583; no functional change. The two conflicts in `ledger/forging/forger.go` (against the #3926 opcert-counter pre-flight) were additive and resolved by keeping both sides. Range-diff: 16 of 18 patches byte-identical, the other two differ only in context lines. All commits build individually; suites, race and mutation checks re-run green. This branch still carries the #3955 commits until that PR merges.

chadaibbhot-git:
CI note: the three `ledger/eras` failures on this head (`TestConwayUtxoValidationRuleIndexesArePinned`, `TestValidateTxConwayGenuinelyMissingInputStillRejected`, `TestValidateTxConwayGenuinelyUnbalancedStillRejected`) are inherited from main 90504583 — see #3976. Every other package passes; this branch does not touch `ledger/eras`.

chadaibbhot-git:
Rebased onto main 6b6346e3 (picks up #3977, which fixes the inherited `ledger/eras` failures); no conflicts, all 18 patches byte-identical. Suites, race and mutation checks re-run green.

ledger/forging/forger.go
```diff
@@ -1161,6 +1360,23 @@ func (f *BlockForger) buildBlock(
 			kesPeriod,
 			leiosData,
 			generation,
+			blockCtx,
```
wolf31o2, line 1511: When the parent-change retry path is taken, verify that all parent-dependent Leios data and embedded-certificate bookkeeping are re-resolved for the new parent before retrying or falling back. Reusing data selected against the abandoned parent can produce a block whose certificate/announcement does not match its actual parent.

wolf31o2 (changes requested):
Code review: changes requested. The current head still needs the blockfetch/rollback race resolved before approval.

wolf31o2 (dismissed):

chadaibbhot-git:
Rebased onto main (`9288bd78`) and addressed both points. #3955 is merged, so its five commits are dropped from this stack and only the thirteen alternative-block commits remain, plus the two below.

**1. Blockfetch/rollback race — closed at the mutation, not just before it (`53407120`)**

The previous round only tested the rollback generation in the blockfetch drain, before the chain add. That does not close the race: the drain holds `chainsyncBlockfetchMutex` while the sibling adoption that publishes a generation and truncates holds `chainsyncMutex`, so a rollback can land between the drain's test and the add taking the chain mutex. A block fetched for the abandoned segment that still fits the post-truncation tip would be admitted, because nothing the chain checks for itself sees anything wrong with it.

`chain.Chain.AddBlockWithPointDeferredIf` now takes an admission predicate and evaluates it under the chain mutex, before any chain state is read or written, abandoning the add with `ErrBlockAddNotAdmitted` when it refuses. The ledger passes its generation comparison as that predicate, so the test and the mutation it guards are one atomic step. The drain-side test stays as a fast path that drops a whole superseded batch without offering any of it to the chain.

- `TestAddBlockWithPointDeferredIfRefusesWithoutMutating`
- `TestAddBlockWithPointDeferredIfEvaluatesAdmitUnderTheChainMutex` — pins the property by observing that a concurrent add cannot progress while the predicate runs
- `TestStaleBlockfetchBatchIsDiscardedAfterLocalSiblingAdoption/rollback_lands_after_the_fast_path_passes`

Worth noting: with the predicate in place, disabling the drain-side fast path alone no longer turns any test red — the guarantee has moved to the atomic site. Removing the predicate evaluation turns both new tests red.

**2. `forger.go` parent-change and Leios bookkeeping (`2931a891`)**

To be precise about the premise: there is no retry path. `buildBlock` is called exactly once, and on `errParentChangedDuringBuild` the forge fails and returns — nothing retries and nothing falls back, so no path reuses data across a retry.

The reachable form of the hazard you describe is a stale *pre-build* resolution. Leios data is resolved before `buildBlock`, but the builder binds the block's parent from its own fresh chain-tip read. A tip that advances while endorser-block production and mempool rebasing run therefore yields a block built on a new parent while carrying a certificate from the old parent's endorser-block lineage. The builder's own re-check only covers changes during transaction selection, after its own tip read, so it does not catch this.

The parent announcement is now re-read immediately before the build. If it moved, the Leios data is dropped and a plain ranking block is forged, and `embeddedEb` is dropped with it — left set, `MarkEndorserBlockEmbedded` would record an endorser block as embedded in a block that does not carry it. Dropping rather than re-resolving is deliberate: re-resolving would have to re-run endorser-block production against the new parent, and a second endorser block for one slot is not something the loop may do. Resolution moved into `leiosParentAnnouncement` and is passed into `leiosBlockDataForSlot`, so the re-check is the only added provider call; three existing tests move from `parent.calls == 1` to `2` for that reason.

The alternative path is unaffected: it carries no Leios data, and the parent announcement is not consulted for it at all, since the provider answers for the tip, which there is the rival.

- `TestCheckAndForgeProductionDropsLeiosDataWhenTheParentMoves`

**Verification:** every commit builds; `go vet`, `gofmt` and `golangci-lint` clean over `./ledger/... ./chain/...`; full suites green for `ledger/forging`, `ledger`, `chain` and the root package; `-race` green on forging and chain. Each guard in this PR was mutation-checked red→green.

One gap I would rather state than leave implied: the `TipPredecessor` primary-chain read-lock fix has no failing-test signal — removing the lock leaves the suite green, including under `-race`, because no test builds a fork chain whose predecessor sits in the common prefix while the primary chain is concurrently mutated. It is correctness-by-construction against the documented `lockBlockIndexReadLocks` order.

chadaibbhot-git:
The `go-test (Linux, race)` failure on `2931a891` is a pre-existing flake in `TestWindowedRewindConvergesWhilePrimaryChainExtends`, not a regression from this PR. Evidence:

**Reproduced on clean `upstream/main` (`9288bd78`), nothing from this branch applied:**

```
go test -race ./ledger/ -run TestWindowedRewindConvergesWhilePrimaryChainExtends -count=60
--- FAIL: TestWindowedRewindConvergesWhilePrimaryChainExtends (0.45s)
    rollback primary chain to intermediate point at slot 1030: remove block at index 111: block not found
```

Same assertion and same error shape as CI, with the slot and index varying per run — CI hit `slot 2110 / index 219`, this branch at `-count=20` hit `slot 1520 / index 160`, clean main hit `slot 1030 / index 111`. The varying coordinates are what a timing-dependent failure looks like rather than a deterministic one.

I also tried to bisect it across this branch's commits and the boundary did not hold: at `-count=5` it appeared to land on the `TipPredecessor` commit, but re-running that same commit and its parent at `-count=20` inverted the result — the parent failed and the supposed first-bad commit passed 20 times. That is not a commit-correlated failure.

`recovery_rewind_window_test.go` is also outside everything this PR touches (`chain/chain.go`, `chain/errors.go`, `ledger/chainsync.go`, `ledger/local_sibling.go`, `ledger/forging/*`, `ledger/state.go`, `node_forging.go`), and `TipPredecessor` — the only chain entry point this PR adds locking to — is called solely from the sibling adoption and the forger, neither of which the rewind test exercises.

I have re-run the failed job. I am not attempting a fix for it inside this PR; it wants its own change against that test, and folding an unrelated concurrency fix into an approved branch would only widen the review surface here.

For completeness, on `2931a891` locally: full suites green for `ledger/forging`, `ledger`, `chain` and the root package, and `-race` green for `ledger/forging` and `chain`.

chadaibbhot-git:
Follow-up: that flake is already being fixed in #4002 ("bar rollback from uncommitted indices, record skipped restores"), which lines up with the failure signature here — the rollback hitting an index whose block is not there yet. So there is nothing to fix for it on this branch. I will rebase this PR onto main once #4002 lands, for a clean race run.

chadaibbhot-git:
Rebased onto `main` at `eabf653d`, which now carries #4002 (rollback barred from
uncommitted indices). No conflicts — the replay was a pure context shift, and the
branch diff's added/removed lines are byte-identical to the pre-rebase head.

Re-verified on the new base: `go build ./...`, `go vet ./ledger/... ./chain/... .`
and `gofmt -l .` clean; `golangci-lint run ./ledger/... ./chain/...` 0 issues;
`go test ./ledger/forging/... ./ledger/ ./chain/... . -count=1` all green, and
`go test -race ./ledger/forging/... ./chain/... ./ledger/` all green.

All eight mutation checks from the previous rounds were re-run on the rebased
tree and each went red again before being restored: the parent-slot-order guard
(`TestBuildBlockRefusesAParentAtItsOwnSlot`), mempool exclusion on an alternative
(`TestBuildBlockOnContextCarriesNoMempoolTransactions`), the Leios-data refusal
(`TestBuildBlockOnContextRejectsLeiosData`), the durable-fence precondition
(`TestEqualSlotAlternativeRequiresADurableFence`), sibling slot equality
(`TestAdoptLocalForgedSiblingRejectsNonSiblings`), range progress counted on
apply rather than arrival and the chain admission predicate (both
`TestStaleBlockfetchBatchIsDiscardedAfterLocalSiblingAdoption`, the latter via
its `rollback lands after the fast path passes` subtest and
`TestAddBlockWithPointDeferredIfRefusesWithoutMutating`), and the Leios drop when
the parent moves (`TestCheckAndForgeProductionDropsLeiosDataWhenTheParentMoves`).

Worth noting on the admission predicate: with it disabled,
`TestAddBlockWithPointDeferredIfEvaluatesAdmitUnderTheChainMutex` does not just
fail, it never completes — the predicate is simply never evaluated.

CI on this head is queued behind a repo-wide runner backlog at the moment.

ledger/forging/forger.go
chadaibbhot-git: Implemented as you specified, and the branch is now rebased onto `main` at `eabf653d`.

**Generation published before the truncation.** `AdoptLocalForgedSibling` bumps
`ls.blockfetchRollbackGeneration` immediately *before*
`rollbackChainAndStateDeferred` (`ledger/local_sibling.go`). Requests are tagged
at issue time — `startQueuedBlockfetchLockedWithWaitSignal` stores
`blockfetchBatchRollbackGeneration = blockfetchRollbackGeneration.Load()` — and
`flushPendingBlockfetchBlocksDeferred` discards a batch whose tag no longer
matches, releasing each block's deferred-header-validation bookkeeping via
`discardStaleBlockfetchBatch`.

**The generation test alone is not the guarantee, so it is not where the
guarantee lives.** The flush holds `chainsyncBlockfetchMutex` while the adoption
holds `chainsyncMutex`, so a rollback can publish and truncate between the check
and the add. The same predicate is therefore passed into the chain as an
admission predicate — `Chain.AddBlockWithPointDeferredIf(..., admit)` — and
evaluated under `c.mutex`, atomically with the tip mutation it guards. On
`ErrBlockAddNotAdmitted` the remainder of the batch is dropped in one go. The
pre-loop check is kept only as a fast path that avoids offering a whole dead
batch to the chain.

**One thing your finding surfaced indirectly.** Range progress was being noted on
block *arrival*, so a block that was discarded or declined still cleared the
failure record for a stuck range — the record that eventually drops an
unservable queued header, which is the routine aftermath of a slot battle. It is
now noted where a block actually extends the chain, and
`discardStaleBlockfetchBatch` deliberately leaves the record alone.

Pinned by `TestStaleBlockfetchBatchIsDiscardedAfterLocalSiblingAdoption`
(subtests: control, stale batch after the alternative wins, stale batch inside
the rollback window, **rollback lands after the fast path passes**, stale batch
after the alternative loses), plus
`TestAddBlockWithPointDeferredIfRefusesWithoutMutating` and
`TestAddBlockWithPointDeferredIfEvaluatesAdmitUnderTheChainMutex`.

Mutation-checked on the rebased tree: noting range progress on arrival again, and
disabling the admission predicate, each turn those tests red. With the predicate
disabled the concurrency test does not merely fail, it never completes — the
predicate is never evaluated.

ledger/forging/builder.go
chadaibbhot-git: Taken, and I chose the second of your two options: reject.

`DefaultBlockBuilder.buildBlock` now sets `mempoolTxs = nil` whenever an explicit
`blockCtx` is present, so an equal-slot alternative is forged as an empty ranking
block.

The reason for rejecting rather than snapshotting is that the snapshot does not
exist to build against. Every validator reachable from the builder answers
against the ledger's live state, which has the rival applied:
`WithTxValidationSession` and `ValidateTxWithOverlay` both read the current UTxO
set, and the overlay can only add pending state, never un-apply the block at the
tip. So a transaction spending a rival-created UTxO passes validation, gets
selected, and then adoption rolls the rival back before applying our block — its
input no longer exists, the apply fails, and the node is left wedged at the fork
point. That is a worse failure than conceding the slot.

An empty ranking block is valid against the parent state whatever the rival did,
and it still carries the VRF and opcert that decide the battle, which is the
entire purpose of forging it. The transactions stay in the mempool for the next
block. If a parent-state validation snapshot is ever exposed, this is a
one-line change to start using it; I did not want to invent one here.

Pinned by `TestBuildBlockOnContextCarriesNoMempoolTransactions`, and
mutation-checked on the rebased tree (restoring the mempool selection on the
explicit-context path turns that test red).

chain/chain.go
chadaibbhot-git: Fixed, at that return path and at two more places on the way to the signer.

**`Chain.TipPredecessor`** — the final guard now reads

```go
if !bytes.Equal(tipBlock.Hash, c.currentTip.Point.Hash) ||
	!bytes.Equal(tipBlock.PrevHash, parentBlock.Hash) ||
	parentBlock.Slot >= tipBlock.Slot {
	return ocommon.Point{}, ochainsync.Tip{}, false
}
```

so the strictly-earlier parent slot the doc comment promises is enforced, not
just documented and tested.

**`BlockForger.alternativeBlockContext`** re-checks it against the chain rather
than trusting the slot-clock snapshot the gate used: `tip.Point.Slot != slot ||
parent.Slot >= slot` declines the contest.

**`DefaultBlockBuilder.buildBlock`** refuses any resolved parent not strictly
below the forged slot with `errParentSlotNotBelowBlock`, before any signing. That
guard is new on the **live-tip path too** — a builder asked to forge at a slot
the tip already held previously produced an invalid block and only found out
after KES signing.

Pinned by `TestBuildBlockRefusesAParentAtItsOwnSlot` (all three ways to name a
parent that is not below the block, including the live-tip path), the
`chain/tip_predecessor_test.go` fail-closed cases, and
`TestValidateBlockOrderPinsTheEqualSlotAlternativeShape`, which pins both halves
of the validator rule the builder is obeying. Mutation-checked on the rebased
tree: disabling the builder's slot-order guard turns
`TestBuildBlockRefusesAParentAtItsOwnSlot` red.

ledger/forging/forger.go
chadaibbhot-git: Checked, and the premise needed correcting before the fix could be right — but
the hazard you describe is real in a different shape, and it is now fixed.

**There is no retry path.** `f.buildBlock` is called exactly once; on
`errParentChangedDuringBuild` the forge fails and returns. Nothing retries and
nothing falls back, so there is no path that reuses Leios data across a retry.

**The reachable form is a stale *pre-build* resolution.** Leios data is resolved
before `buildBlock`, but the builder binds the block's parent from its own fresh
chain-tip read. A tip that advances while endorser-block production and mempool
rebasing run yields a block built on the new parent while carrying a certificate
from the old parent's endorser-block lineage — exactly the mismatch you name. The
builder's own re-check only covers changes during transaction selection, after
its own tip read, so it does not catch this.

**Fix:** re-read the parent announcement immediately before the build; if it
moved, drop the Leios data and forge a plain ranking block, and drop `embeddedEb`
with it so `MarkEndorserBlockEmbedded` cannot record an endorser block as
embedded in a block that omits it. Dropping rather than re-resolving is
deliberate — re-resolving would have to re-run endorser-block production for the
new parent, and a second endorser block for one slot is not something this loop
may do. Resolution moved into `leiosParentAnnouncement` and is passed into
`leiosBlockDataForSlot`, so the re-check is the only added provider call (two
total, which is why three existing tests move from `parent.calls == 1` to `2`).

The alternative path is untouched: it carries no Leios data and does not consult
the parent provider at all.

Pinned by `TestCheckAndForgeProductionDropsLeiosDataWhenTheParentMoves`, and
mutation-checked on the rebased tree (disabling the re-check turns it red).

chadaibbhot-git:
Flagging the red `go-test (Windows)` on `1708c2cc` rather than leaving it
unexplained, because I do **not** think it should be written off as the known
wedging described in the workflow comment.

What happened: the job hit the 90-minute job cap ("The job has exceeded the
maximum execution time of 1h30m0s"). The per-package `-timeout 30m` never fired,
so no single package sat deadlocked for 30 minutes before the cap — this looks
like the suite getting much slower, or a late-onset wedge, rather than one stuck
test.

Why I am not calling it variance:

| head | base | Windows |
|---|---|---|
| this PR, pre-rebase `2931a891` | `9288bd78` | green, 20m14s |
| this PR, rebased `1708c2cc` | `eabf653d` | **killed at the 90m cap** |
| #3963 `4d6eb143` | `eabf653d` | green, 21m09s |
| #3993 `2823d4d2` | `eabf653d` | green, 21m19s |
| #3988 `ad24702c` | `eabf653d` | green, 21m43s |
| #4014 `50c32046` | `eabf653d` | green, 21m37s |
| #4020 `6ce200fa` | `eabf653d` | green, 23m54s |

Five other open PRs on the identical base all finish in 21–24 minutes, so it is
not a base-wide regression or a runner-pool problem on that window. And the
branch content here is byte-identical to the pre-rebase head that finished in
20 minutes — I verified the added/removed lines of the branch diff are unchanged
by the rebase. So the trigger is an interaction between this PR and something
new in the base, not a change I made in the rebase.

A hypothesis I can offer but have not confirmed, since I cannot reproduce Windows
locally: this PR is unusually rollback-heavy in test terms — every
`AdoptLocalForgedSibling` case performs a real one-block rollback — and #4002
made `rollbackLocked` take `batchCommitMutex` exclusively and hold it across the
store removal loop. On a platform where store I/O is as slow as the workflow
comment describes (`./database` at 737s), serialising every rollback behind that
lock could plausibly turn a 20-minute suite into a 90-minute one without any
single package tripping its own timeout. That would be a property of the
combination rather than a defect in either change.

I would rather not guess further. Could someone with the permission re-run just
that job? A second data point separates "reproducible interaction" from "unlucky
run" immediately, and I cannot re-run workflows on this repository. If it wedges
again I will treat it as a blocker on this PR and dig into which package is
responsible rather than asking you to carry it.

Everything else on this head is green, including `go-test (Linux, race)`, which
was the failing check before the rebase.

chadaibbhot-git:
Retracting the hypothesis in my previous comment: it was wrong.

`main` moved to `31dd4a8a`, which gave this branch a legitimate re-trigger, so I
rebased onto it (`1708c2cc` -> `ec7166b3`, no conflicts, branch content again
byte-identical). **`go-test (Windows)` passed in 20m52s** — 05:04:42Z to
05:25:34Z — squarely inside the 20–24 minute band every other PR shows.

So the 90-minute wedge on the previous head was not an interaction between this
PR and #4002's `batchCommitMutex`. I said that claim stood or fell on this run,
and it falls. The rollback-heavy-tests-on-slow-Windows-I/O story was a plausible
mechanism with no evidence behind it, and I should have weighted "known wedging,
tracked in #3746" more heavily than I did given the workflow comment says exactly
that. Apologies for the noise.

One caveat I do not want to overstate in the other direction: the base moved by a
single commit between the two runs (#4032, guarding `bufferedHeaderEvents` with
the blockfetch mutex), so I cannot fully separate "the earlier run was unlucky"
from "#4032 changed something". A single green run does not distinguish those,
and I am not going to claim it does. The honest summary is: one wedge, one clean
run at normal duration, no evidence implicating this PR.

Full head `ec7166b3` is green: Linux, Linux race, Windows, macOS, docker,
govulncheck, lint, Conventional Commits, DCO.

Re-verified on the new base before pushing: build, vet, `gofmt -l .` clean,
`golangci-lint run ./ledger/... ./chain/...` 0 issues,
`go test ./ledger/forging/... ./ledger/ ./chain/... . -count=1` green,
`go test -race ./ledger/forging/... ./chain/... ./ledger/` green, and all eight
mutation checks re-run red then restored.

wolf31o2 (approved):

PR #3960, opened 2026-09-05
fix(peergov): drop ledger relay hostnames that cannot resolve

### LINKED ISSUE #2018: peergov: reject malformed ledger relay hostnames before dialing
## Summary
A node can attempt to dial a malformed ledger-discovered peer address such as `--pool-relay-port:3001`.

This is not coming from static topology. It is coming from the ledger peer discovery path, which currently accepts any non-empty relay hostname, converts it to `hostname:port`, and will keep it even if DNS resolution fails.

## Field Evidence
Observed on a preview block producer running current `main` on April 23, 2026.

Remote config was clean:
- `topology.json` only contained the expected bootstrap peer and local root
- the bad address was logged with `source=ledger`

Relevant log sequence:

```json
{"time":"2026-04-23T20:08:49.275507748Z","level":"INFO","msg":"discovered ledger peers","component":"peergov","added":2,"target":20,"candidates":2}
{"time":"2026-04-23T20:08:49.275624749Z","level":"INFO","msg":"outbound: failed to establish connection to --pool-relay-port:3001: dial tcp: lookup --pool-relay-port: no such host","component":"peergov"}
{"time":"2026-04-23T20:08:56.431654239Z","level":"INFO","msg":"outbound: removing peer after repeated connection failures","component":"peergov","address":"--pool-relay-port:3001","source":"ledger","retry_count":4,"deny_duration":1800000000000}
```

At the same startup, the other ledger-discovered candidate connected successfully (`134.199.224.183:4200`).

## Current Code Path
- `PoolRelay.Addresses()` blindly converts any non-empty hostname into `hostname:port`
- unresolved hostnames are treated as routable and are preserved as lowercase hostnames
- ledger peer discovery therefore admits obviously invalid tokens into the outbound peer set

Relevant files:
- `peergov/ledger.go`
- `peergov/ledger_discovery.go`
- `peergov/peers.go`

## Expected Behavior
Malformed ledger relay hostnames should be rejected before they are added as peers or dialed.

## Scope
This is a minor defensive bug. It is not the primary replay / metadata-corruption issue we were investigating on the same node.

## Suggested Fix
Add validation for ledger-discovered relay hostnames before `addLedgerPeer` / dialing. At minimum, reject obvious garbage such as flag-like tokens, malformed hostnames, or values that cannot possibly be valid DNS names or IP literals.

wolf31o2:
Fixes #2018

## Defect

`PoolRelay.Addresses` (`peergov/ledger.go:93`) turned any non-empty `Hostname` into a candidate address. The ledger carries whatever a pool operator registered, so a value like `--pool-relay-port` reached `addLedgerPeer` and produced a dial attempt, four connection failures and a 1800s deny entry before anything established it was never an address.

`Addresses` is the only consumer of `PoolRelay.Hostname`, and `flattenRelayCandidates` is its only caller, from both `ledger_discovery.go:155` and `peer_snapshot.go:103`. Filtering there covers every path a ledger relay takes into the peer set.

## Change

`IsResolvableHost` accepts an IP literal or a syntactically valid DNS name.

It rejects only what no resolver could answer: an empty label, a label over 63 bytes, a label starting or ending with a hyphen, a name over 253 bytes, and any byte outside the set operators use. A single trailing dot is accepted as the absolute form.

Underscores are accepted deliberately. They are invalid in a host name under RFC 1123 but appear in real zones and a resolver will attempt them. The two failure modes are not symmetric — rejecting a name that would have resolved costs a real relay, while admitting one that cannot costs a dial and a deny entry — so the check errs toward admitting and leaves the rest to DNS.

## Tests

- `TestIsResolvableHost` — 18 cases, including the reported `--pool-relay-port`, empty and over-long labels, hyphen-edged labels, a 257-byte name, a label at the 63-byte limit, the absolute form, IPv4 and IPv6 literals, and the underscore case.
- `TestPoolRelayAddressesDropsUnresolvableHostname` — a relay carrying both the malformed hostname and a usable IPv4 yields only the IPv4 address.
- `TestPoolRelayAddressesKeepsValidHostname` — an ordinary hostname is still emitted with its port.
- `TestFlattenRelayCandidatesDropsUnresolvableHostname` — the discovery path itself, where the address reached `addLedgerPeer`.

The two rejection tests fail without the change, on `--pool-relay-port:3001` appearing in the candidate list.

---
## Summary by cubic
Fixes #2018. `PoolRelay.Addresses` previously turned any non-empty hostname into a candidate peer, so a misconfigured value like `--pool-relay-port` caused dial attempts, four connection failures, and an 1800s deny entry. Hostnames are now filtered by `IsResolvableHost` before becoming candidates.

**Validation rules**
- Accepts IP literals and syntactically valid DNS names.
- Rejects only what no resolver could answer: empty, over-long, or hyphen-edged labels, names over 253 bytes, and invalid bytes.
- Accepts underscores deliberately — invalid under RFC 1123 but resolvers attempt them, and rejecting a resolvable name costs a real relay.

Tests cover the reported `--pool-relay-port`, label-length and hyphen edge cases, the absolute form, IPv4/IPv6 literals, and both `Addresses` and `flattenRelayCandidates` paths. The peer-sharing test drops a zone-id IPv6 input because `AddPeer` now resolves non-IP hosts before storing.

<sup>Written for commit 7f2076c97e351964ff74bc3ba74533e5d4cf4b0c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3960?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ouroboros/peersharing_test.go
+++ b/ouroboros/peersharing_test.go
@@ -78,14 +78,10 @@ func TestPeerSharingShareRequestBoundsValidPeers(t *testing.T) {
 		t,
 		peerGov.AddPeer("10.0.0.1:3001", peergov.PeerSourceTopologyLocalRoot),
 	)
-	// Exercise every validation failure in the sharing adapter: an address
-	// without host:port structure, a split address whose host is not an IP,
-	// and a port outside the uint16 wire range.
+	// Exercise validation failures in the sharing adapter: an address without
+	// host:port structure and a port outside the uint16 wire range. Keep these
+	// inputs deterministic; AddPeer resolves non-IP hosts before storing them.
 	require.NoError(t, peerGov.AddPeer("malformed", peergov.PeerSourceP2PGossip))
-	require.NoError(
-		t,
-		peerGov.AddPeer("[44.0.0.9%invalid]:3001", peergov.PeerSourceP2PGossip),
-	)
 	require.NoError(
 		t,
 		peerGov.AddPeer("44.0.0.9:70000", peergov.PeerSourceP2PGossip),
--- a/peergov/ledger.go
+++ b/peergov/ledger.go
@@ -17,6 +17,7 @@ package peergov
 import (
 	"net"
 	"strconv"
+	"strings"
 )
 
 // SyncProgressProvider defines an interface for querying sync progress.
@@ -85,7 +86,11 @@ func (r PoolRelay) Addresses() []string {
 	var addresses []string
 	portStr := formatPort(r.Port)
 
-	if r.Hostname != "" {
+	// A hostname that cannot resolve is not a peer. The ledger carries
+	// whatever an operator registered, and a value like "--pool-relay-port"
+	// becomes a dial attempt, four connection failures and a half-hour deny
+	// entry before anything notices it was never an address (issue #2018).
+	if IsResolvableHost(r.Hostname) {
 		addresses = append(addresses, net.JoinHostPort(r.Hostname, portStr))
 	}
 	if r.IPv4 != nil && len(*r.IPv4) > 0 {
@@ -103,6 +108,68 @@ func (r PoolRelay) Addresses() []string {
 
 	return addresses
 }
+
+// IsResolvableHost reports whether host could be resolved to an address: an IP
+// literal, or a syntactically valid DNS name.
+//
+// It is deliberately permissive about what a DNS name may contain, because the
+// two failure modes are not symmetric. Rejecting a name that would have
+// resolved costs a real relay; admitting one that cannot resolve costs a dial,
+// its retries and a deny entry. So this rejects only what no resolver could
+// answer -- a label that is empty, over-long, or starts or ends with a hyphen,
+// a name over 253 bytes, or a byte outside the set operators actually use --
+// and leaves anything else to DNS.
+//
+// Underscores are accepted. They are invalid in a host name under RFC 1123 but
+// appear in real zones, and a resolver will attempt them.
+func IsResolvableHost(host string) bool {
+	if host == "" {
+		return false
+	}
+	if net.ParseIP(host) != nil {
+		return true
+	}
+	// A single trailing dot is the valid absolute form.
+	name := strings.TrimSuffix(host, ".")
+	if name == "" || len(name) > maxDNSNameLen {
+		return false
+	}
+	for label := range strings.SplitSeq(name, ".") {
+		if !isValidDNSLabel(label) {
+			return false
+		}
+	}
+	return true
+}
+
+func isValidDNSLabel(label string) bool {
+	if label == "" || len(label) > maxDNSLabelLen {
+		return false
+	}
+	if label[0] == '-' || label[len(label)-1] == '-' {
+		return false
+	}
+	for i := range len(label) {
+		c := label[i]
+		switch {
+		case c >= 'a' && c <= 'z',
+			c >= 'A' && c <= 'Z',
+			c >= '0' && c <= '9',
+			c == '-', c == '_':
+		default:
+			return false
+		}
+	}
+	return true
+}
+
+const (
+	// maxDNSNameLen is the longest name a resolver will accept, excluding the
+	// root label.
+	maxDNSNameLen = 253
+	// maxDNSLabelLen is the longest single label.
+	maxDNSLabelLen = 63
+)
 
 // formatPort converts a port number to a string.
 // Returns the default Cardano port "3001" if port is 0.
--- a/peergov/ledger_hostname_test.go
+++ b/peergov/ledger_hostname_test.go
@@ -1,0 +1,111 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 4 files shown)


arepala-uml (changes requested):
This branch conflicts with current `main` (`mergeStateStatus: DIRTY`), and it's a real content conflict, not a stale flag — reconfirmed just now with a fresh merge attempt against `origin/main` (9288bd78).

Both this PR's `fix(peergov): reject ORCHIDv2 peer addresses` commit and `main`'s independent fix for #3792 edit the same `unreachablePrefixes` slice in `peergov/peers.go`. Main added RFC 5737 TEST-NET-1/2/3 and the RFC 3849 documentation range; this branch predates that and only adds `2001:20::/28` (ORCHIDv2). Git reports a content conflict there, which means resolution is manual — and a careless resolution (e.g. "keep incoming"/"keep current" in a merge tool) can silently drop one side's entries.

To make that concrete rather than hypothetical, here's a test proving the branch as pushed does not have main's #3792 fix:

```go
// peergov/pr3960_conflict_regression_test.go
package peergov

import (
	"net"
	"testing"
)

// TestPR3960MergeConflictRegressesTestNetFix demonstrates why PR #3960's
// current merge conflict in peergov/peers.go is not just a textual nuisance.
//
// origin/main already rejects RFC 5737 TEST-NET-1/2/3 and the RFC 3849
// documentation range (issue #3792) from unreachablePrefixes. PR #3960's
// branch was cut before that fix landed and its own commit
// (fix(peergov): reject ORCHIDv2 peer addresses) edits the exact same
// literal slice to add 2001:20::/28. Both edits touch unreachablePrefixes,
// so Git reports a content conflict on the merge -- and whoever resolves it
// by hand can silently keep only one side's entries instead of the union of
// both.
//
// This test runs main's own #3792 fixture addresses against this branch's
// current IsRoutableIP. It fails on the unresolved PR branch, proving the
// TEST-NET/documentation fix is not present here -- exactly what a careless
// conflict resolution (e.g. "keep incoming" in a merge tool) would ship.
func TestPR3960MergeConflictRegressesTestNetFix(t *testing.T) {
	for _, tc := range []struct {
		name string
		addr string
	}{
		{"ipv4 test-net-1 (RFC 5737)", "192.0.2.1"},
		{"ipv4 test-net-2 (RFC 5737)", "198.51.100.1"},
		{"ipv4 test-net-3 (RFC 5737)", "203.0.113.1"},
		{"ipv6 documentation (RFC 3849)", "2001:db8::1"},
	} {
		t.Run(tc.name, func(t *testing.T) {
			if IsRoutableIP(net.ParseIP(tc.addr)) {
				t.Errorf(
					"IsRoutableIP(%q) = true, want false: this address is "+
						"rejected on origin/main (issue #3792) but this "+
						"branch's unreachablePrefixes list does not contain "+
						"it -- merging PR #3960 without manually restoring "+
						"main's entries during conflict resolution silently "+
						"reverts #3792",
					tc.addr,
				)
			}
		})
	}
}
```

Result on this branch (`a00805d8`) as pushed:

```
--- FAIL: TestPR3960MergeConflictRegressesTestNetFix (0.00s)
    --- FAIL: .../ipv4_test-net-1_(RFC_5737) (0.00s)
    --- FAIL: .../ipv4_test-net-2_(RFC_5737) (0.00s)
    --- FAIL: .../ipv4_test-net-3_(RFC_5737) (0.00s)
    --- FAIL: .../ipv6_documentation_(RFC_3849) (0.00s)
```

Same test against current `origin/main` (9288bd78): passes clean. So this isn't a false positive — it's confirming the branch is genuinely missing the fix that a naive conflict resolution could fail to restore.

The core hostname-resolution fix in this PR (`IsResolvableHost`, its `Addresses()`/`flattenRelayCandidates` wiring, and the associated tests) is correct and well-verified independently — that part is not in question. The ask is narrow: merge `main` into this branch, reconcile `unreachablePrefixes` to keep the union of both sides' entries (TEST-NET-1/2/3, the documentation range, and ORCHIDv2), and confirm CI runs clean afterward (several required checks — go-test, race, lint, Docker CI — aren't currently running at all because of the conflict state).

arepala-uml (changes requested):
Reviewed at head 7000466d. The merge conflict with main's #3792 fix is resolved — `peers.go` retains both IP block lists. Core fix (`IsResolvableHost` is purely syntactic, no hang/DNS-block risk) is correct and well covered.

One open, unaddressed finding (Cubic, confirmed valid): `ouroboros/peersharing_test.go:85` adds a case that triggers a real, unmocked `net.LookupIP` via the unexported `lookupIP` var in `peergov/peers.go:38`, which the external test package can't stub — non-deterministic and against this repo's deterministic-test convention. Please stub the lookup or drop the case before merge.

arepala-uml (approved):

Issue #3885, opened 2026-09-04
musashi: fresh genesis sync halts at slot 958027 on reward-state divergence


Cerkoryn:
## Problem

A fresh from-genesis Musashi sync on Dingo `v0.70.6` reaches applied block
`44726` at slot `957971`, then rejects the canonical transaction in the next
block because the locally reconstructed reward-account balance is lower than
the withdrawal amount.

The first rejection rewinds and requests a fresh ChainSync intersection. When
the same block is delivered again, the reward-state divergence remains and the
new persistent-failure handling from #3639 deliberately halts the ledger
pipeline.

This run advances well past origin and does not reproduce the earlier Musashi
BlockFetch decode failure from #3798.

## Environment

- Dingo: `v0.70.6` (`af16f0e3`)
- Network: Musashi
- Sync source: fresh empty database, from genesis; Mithril disabled
- Runtime mode: `leios`
- Storage mode: `core`
- Metadata: SQLite
- Blob storage: Badger
- Node role: block producer
- Host: macOS 26.6.1, arm64

## Reproduction

1. Start `dingo serve` with the Musashi configuration above and an empty
   database.
2. Allow the node to sync from genesis.
3. Observe ledger application when it reaches slot `958027`.

## Actual behavior

The canonical transaction
`0818e95078b3d1ba0beeaa3ea361987213323d5a501804bd8cd5e324bb8a0ba3`
at block slot `958027` withdraws `3552343148` lovelace. Dingo has reconstructed
only `3549340291` lovelace for the account, a deficit of `3002857` lovelace.

Relevant log sequence:

```text
level=WARN msg="deterministic transaction validation rejected a block on the primary chain; rewinding so chain selection can offer another candidate" component=ledger tx_hash=0818e95078b3d1ba0beeaa3ea361987213323d5a501804bd8cd5e324bb8a0ba3 failing_block_slot=958027 rewind_target_slot=957971 error="tx 0818e95078b3d1ba0beeaa3ea361987213323d5a501804bd8cd5e324bb8a0ba3 validation failure at slot 958027: set transaction metadata for tx 0818e95078b3d1ba0beeaa3ea361987213323d5a501804bd8cd5e324bb8a0ba3 (block idx 0, slot 958027): reward withdrawal amount 3552343148 exceeds account balance 3549340291: reward withdrawal exceeds account balance"

[...]

PR #3969, opened 2026-09-05
fix(ledger): keep excluded pool stake in the reward denominator

wolf31o2:
## Defect

`buildRewardStateInputs` excludes a pool with missing or malformed registration data from `reward_pool_input` so one bad pool cannot wedge the whole epoch boundary, and then derived `reward_snapshot.total_active_stake` from that same reduced distribution (`ledger/snapshot/rotation.go:428`).

That total is the reward calculation's sigma_a denominator: `ledger/reward_calculation.go:2945` copies it into `rewards.Snapshot.TotalActiveStake`, `ledger/rewards/rewards.go:1053` passes it to `apparentPerformance` as `activeStake`, and `ledger/rewards/rewards.go:1290` computes `sigma_a = poolStake / activeStake`, with `poolReward = floor(beta/sigma_a * optimalReward)`.

Shrinking the denominator onto the surviving pool set therefore raises sigma_a for every pool that did survive, lowers its apparent performance, and under-credits every member and leader reward the node reconstructs — by a fraction equal to the excluded pool's share of active stake, for as long as the exclusion holds. The node then rejects a canonical withdrawal whose amount exceeds the balance it accumulated.

## Reference

cardano-ledger builds `ssTotalActiveStake` by summing every registered credential that carries a delegation and never intersects that set with `ssStakePoolsSnapShot`:

- `libs/cardano-ledger-core/src/Cardano/Ledger/State/SnapShots.hs:424` — `mkSnapShot` sets `ssTotalActiveStake = sumAllActiveStake ssActiveStake`
- `libs/cardano-ledger-core/src/Cardano/Ledger/State/SnapShots.hs:447` — `snapShotFromInstantStake` builds `ssActiveStake` from `resolveInstantStake`, independently of `psStakePools`
- `libs/cardano-ledger-core/src/Cardano/Ledger/State/Stake.hs:217` — "active stake means any stake credential that is registered and delegated to a stake pool"
- `eras/shelley/impl/src/Cardano/Ledger/Shelley/Rewards.hs:394` — `sigmaA = poolTotalStake %. unCoinNonZero totalActiveStake`

A pool absent from the snapshot's pool map earns nothing while its delegators keep contributing to the denominator, so `sum(pools) <= totalActiveStake` is a representable state upstream, not a contradiction.

## Change

- `ledger/snapshot/rotation.go` — `total_active_stake` is computed from the full reward-stake distribution. `total_pool_count` and `total_delegators` keep describing the `reward_pool_input` rows actually written.
- `ledger/rewards/rewards.go`, `ledger/reward_calculation.go` — the three "rows sum to the snapshot total" invariants become an upper bound. Rows summing to more than the declared active stake still fail.
- `database/models/reward_state.go` — `RewardStakeCalculationVersion` moves to 2. A database holding version 1 rows fails closed at startup (`node.go:2170`) rather than reusing an understated denominator.
- `DATABASE.md` — `reward_snapshot` row semantics.

## Coverage

Each new test fails on `main` and passes with the change:

- `ledger/snapshot`: `TestRewardSnapshotActiveStakeKeepsDegradedPoolStake` — one delegated-to pool with no resolvable registration; the snapshot keeps its 3000000000 stake in `total_active_stake` while `total_pool_count` and `total_delegators` stay at the single surviving pool. Without the change the total reads 7000000000.
- `ledger/rewards`: `TestCalculateAcceptsActiveStakeAboveThePoolSet` and `TestShrunkActiveStakeUnderCreditsEveryReward` — same epoch, denominator 1250 versus 1000. The full denominator gives apparent performance 5/4 and pays the member 46424 and the leader 57741; the shrunken one gives 1 and pays 37049 and 46283.
- `ledger`: `TestRewardCalculatorInputsAllowExcludedPoolStake` — `validateRewardCalculatorInputs` and `precomputedRewardPoolInputsMatchSnapshot` accept rows summing under the snapshot total and still reject rows summing over it.

`TestHandleEpochTransitionSkipsPoolWithMissingMargin` (`ledger/snapshot/manager_test.go`) and `TestCalculateRejectsActiveStakeMismatch` (`ledger/rewards/rewards_test.go`) asserted the previous contract and are updated to the corrected one; the first covers the same path end to end through `handleEpochTransition`.

Refs #3885

---
## Summary by cubic
Fixes under-credited rewards by keeping excluded pools' delegated stake in `reward_snapshot.total_active_stake`, the sigma_a denominator. Previously a pool with missing or malformed registration data was dropped from both `reward_pool_input` and that total, which inflated sigma_a for every surviving pool and under-credited member and leader rewards by the excluded pool's share of active stake.

**Migration**

- `RewardStakeCalculationVersion` is now 2; databases holding version 1 rows fail closed at startup instead of reusing an understated denominator.

**Behavior**

- `total_active_stake` now sums the full observed reward-stake distribution, including pools excluded from `reward_pool_input`.
- `total_pool_count` and `total_delegators` still count only the `reward_pool_input` rows actually written.
- Reward calculation invariants now check that rows sum to no more than the snapshot total, not exactly; rows summing over it still fail.

<sup>Written for commit c30a2056fcf5bac736e5a760bef4356e64731b62. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3969?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Reward calculations now include stake from pools with degraded registration data in the active-stake denominator, while excluding those pools from reward inputs.
  * Valid reward snapshots may now contain less pool-input stake than total active stake, while still rejecting inputs that exceed the declared total.
  * Corrected reward amounts that could previously be under-credited when degraded pools were present.

* **Documentation**
  * Clarified reward snapshot fields and calculation-version behavior for active stake, pool counts, and delegator counts.


ledger/reward_calculation.go
```diff
@@ -3066,9 +3069,15 @@ func validateRewardCalculatorInputs(
...
 		}
 	}
-	if totalPoolStake != uint64(snapshot.TotalActiveStake) {
+	// reward_snapshot.total_active_stake is the sigma_a denominator and covers
+	// every delegating credential observed at the boundary, including those
+	// whose pool was excluded from reward_pool_input for degraded registration
+	// data (see snapshot.buildRewardStateInputs). The rows may therefore sum to
+	// less than it; summing to more means the row set and the snapshot describe
+	// different boundaries.
+	if totalPoolStake > uint64(snapshot.TotalActiveStake) {
```
cryptodj413, line 3078: Non-blocking. This bound (and its counterpart at `:1412` and `rewards.go:943`) is strictly weaker than the old exact match in one specific way: a `reward_stake_input`/`reward_pool_input` row set that's been *proportionally* reduced (not just an excluded pool dropped) still satisfies `<=` and passes every other check here (pool count, delegator count, per-pool cross-sums are all still exact) — the epoch just quietly pays out slightly low.

The data to close this exists but isn't persisted: at capture time (`rotation.go:437`) the code knows the exact shortfall (`sumPoolStakes(rewardDistribution.PoolStakes) - sumPoolStakes(effective.PoolStakes)`), but the excluded pool contributes no row at all, so this validation path can't reconstruct it after the fact. Two options in increasing cost: (1) log/emit a metric here when `totalPoolStake < snapshot.TotalActiveStake` — right now the exclusion is WARN-logged once at capture, but the calculation path (which can run much later, after a restart or replay) says nothing; (2) persist the excluded stake on `reward_snapshot` and assert `totalPoolStake + excludedStake == TotalActiveStake` exactly, restoring the old check's full strength.

database/models/reward_state.go
```diff
@@ -19,7 +19,12 @@ import "github.com/blinklabs-io/dingo/database/types"
 // RewardStakeCalculationVersion identifies the stake-accounting algorithm
 // used to produce persisted live stake and consensus snapshots. Bump it when
 // changing that calculation so upgrades cannot trust older values.
-const RewardStakeCalculationVersion uint = 1
+//
+// Version 2: RewardSnapshot.TotalActiveStake carries the full observed
+// reward-stake total (the sigma_a denominator) rather than the total of the
+// pools that survived degraded-registration exclusion, so a version 1 row
+// understates the denominator for any epoch that excluded a pool.
+const RewardStakeCalculationVersion uint = 2
```
cryptodj413, line 27: Two non-blocking items on the version bump:

1. This is the first-ever bump of `RewardStakeCalculationVersion` (it's never moved since introduction). Every existing Dingo database now fails closed at startup and needs a rebootstrap (traced: `node.go:2139` -> `StaleConsensusStakeSnapshotsExist` -> hard error at `node.go:2170`) — including databases that never hit the pool-exclusion path, whose version-1 rows are numerically identical to what version 2 would write. A narrower migration is available: for a version-1 row, `reward_snapshot.total_active_stake != epoch_summary.total_active_stake` (`rotation.go:270`) identifies exactly the affected epochs; everything else could be re-stamped to version 2 in place instead of forcing a full rebuild. Neither the PR description nor DATABASE.md flags the rebootstrap requirement for operators upgrading in place.

2. `StaleConsensusStakeSnapshotsExist` (`database/plugin/metadata/sqlstore/live_stake.go:615`) filters its `reward_snapshot` check to `authoritative = TRUE`, so a version-1 **fallback** row (`Authoritative: false`, written by `captureMarkSnapshot`) isn't covered by that clause on its own. I couldn't construct a reachable case where this actually lets a stale row through (the same-boundary `pool_stake_snapshot` fallback row is covered unconditionally, and pruning keeps roughly the epochs the reward calculation reads), so this holds today by coincidence rather than by construction — worth closing with an explicit version filter on the read side regardless.

DATABASE.md
```diff
@@ -980,7 +980,7 @@ process the same pointer unless the claim expires before a result is recorded.
...
...es the `EpochTransitionEvent` published for the same boundary; a boundary that crosses two eras in one block captures the snapshot after both hard-fork transitions so the recorded major is the final era's, not the source era's. Seeded and Mithril-imported rows leave it zero. `total_active_stake` is the reward calculation's sigma_a denominator and covers every delegating credential observed at the boundary, including those whose pool was excluded from `reward_pool_input` for missing or malformed registration data; `total_pool_count` and `total_delegators` describe the `reward_pool_input` rows actually written, so the rows' delegated stake sums to no more than `total_active_stake` rather than to exactly it. `calculation_version` ties authoritative Mark metadata to the stake algorithm that produced its pool rows; version 2 carries that full denominator, while a version 1 row understates it for any epoch that excluded a pool. The fallback claims the `(epoch, mark)` row atomically and skips when an authoritative row already exists, so it cannot overwrite the authoritative capture. Retained for the life of the database. Guard claim/release require the same non-nil metadata transaction. |
```
cryptodj413, line 983: Style nit: "covers every delegating credential observed at the boundary" slightly overstates it — `GetLiveStakeInputsForPools` filters `registered = TRUE`, restricts to pools active at the slot (`getActivePoolsAtSlot`), and applies the optional account-expiration predicate. "Every registered, non-expired credential delegating to a pool active at the snapshot slot" would be exact.

ledger/snapshot/rotation.go
```diff
@@ -427,7 +437,7 @@ func (m *Manager) buildRewardStateInputs(
 			Epoch:        epoch,
 			SnapshotType: snapshotType,
 			TotalActiveStake: types.Uint64(
-				sumPoolStakes(effective.PoolStakes),
+				sumPoolStakes(rewardDistribution.PoolStakes),
```
cryptodj413, line 440: Style nit: `rewardDistribution.TotalStake` (computed a few lines above in `rewardStakeDistribution`, with an overflow check) is the identical value `sumPoolStakes(rewardDistribution.PoolStakes)` recomputes here without one. Safe as written since `rewardStakeDistribution` already errors on overflow before returning, but using the field directly would be more direct and self-documenting than re-summing.

ledger/reward_calculation_test.go
```diff
@@ -6018,3 +6018,80 @@ func TestPrecomputedRewardPoolRewardsMatchInputsFullPot(t *testing.T) {
...
+
+// TestRewardCalculatorInputsAllowExcludedPoolStake covers the reward-input
+// shape snapshot capture writes when a pool is excluded for degraded
+// registration data: reward_pool_input holds only the surviving pools, while
+// reward_snapshot.total_active_stake still carries the excluded pool's
+// delegated stake because that stake belongs in the sigma_a denominator (see
+// snapshot.buildRewardStateInputs). Requiring the rows to sum to exactly the
+// snapshot total forced the denominator down to the surviving pool set, which
+// under-credits every reward the node reconstructs.
+func TestRewardCalculatorInputsAllowExcludedPoolStake(t *testing.T) {
```
cryptodj413, line 6030: Non-blocking test-coverage gap: `TestRewardCalculatorInputsAllowExcludedPoolStake` varies `TotalActiveStake` against a fixed pool/stake-input set, but there's no equivalent negative case that varies `TotalDelegators` instead (checked at `reward_calculation.go:1415` and `:3086`). That comparison is now the primary structural guard that replaced the exact stake-sum comparison this PR relaxed, so it'd be worth a dedicated mismatch case here too, alongside the existing pool-count coverage.

ledger/snapshot/reward_active_stake_test.go
```diff
@@ -0,0 +1,132 @@
...
+// "active stake means any stake credential that is registered and delegated to
+// a stake pool"). A pool whose registration cannot be resolved therefore earns
+// nothing while its delegators keep contributing to that denominator.
+//
+// Deriving RewardSnapshot.TotalActiveStake from the post-exclusion pool set
+// instead shrinks the denominator, which raises sigma_a for every surviving
+// pool, lowers its apparent performance, and under-credits every member and
+// leader reward the node reconstructs. TotalPoolCount and TotalDelegators
+// describe the reward_pool_input rows actually written and stay reduced.
+func TestRewardSnapshotActiveStakeKeepsDegradedPoolStake(t *testing.T) {
```
cryptodj413, line 46: Non-blocking: this test doesn't assert the companion claim DATABASE.md makes for this same boundary — that `pool_stake_snapshot`/`epoch_summary` still carry the degraded pool's full stake even though `reward_pool_input` excludes it. I confirmed that holds in code (`rotation.go:218-276` reads the unmodified `distribution`), but it's untested at this specific boundary.

cryptodj413 (commented):
No merge blockers. Independently traced the ledger claim (`effective.PoolStakes ⊆ rewardDistribution.PoolStakes` by construction, so the swap can't drop or double-count stake) and confirmed it against this repo's own `ssTotalActiveStake` documentation at `ledger/leader/election.go:60-90`. Verified the regression-test discipline directly: reverse-applied only the three source hunks (kept tests at HEAD) and confirmed all 5 new/changed assertions fail pre-fix.

Validation: `go build ./...`, `go test ./ledger/... ./database/...`, `-race` on the affected packages plus all of `ledger/`, `golangci-lint` and scoped `nilaway`, `make docs-parity` — all pass. Four non-blocking recommendations inline. Skipped: devnet (plain + `--conformance`), Antithesis, full `make lint`/`make test`, and an upgrade rehearsal against a real version-1 database (traced the fail-closed gate statically instead of standing one up).

ledger/snapshot/rotation.go
```diff
@@ -385,12 +385,22 @@ type rewardStateBundle struct {
...
+// sums the stake of every registered credential that has a delegation
+// (Cardano.Ledger.State.SnapShots.mkSnapShot over
+// Cardano.Ledger.State.Stake.resolveInstantStake) independently of which pools
+// appear in ssStakePoolsSnapShot. Deriving it from the reduced distribution
+// instead shrinks that denominator, which raises sigma_a for every surviving
+// pool, lowers its apparent performance, and under-credits every member and
+// leader reward on the node — a divergence proportional to the excluded pool's
+// share of active stake. It is therefore computed from the full reward-stake
+// distribution, and the reward calculation requires the reward_pool_input rows
+// to sum to no more than it rather than to exactly it.
```
chrisguiney, line 403: Two comments in this file now say the opposite of what this hunk establishes, and neither is in the diff:

- `rotation.go:544-547` — "Real Cardano reward semantics already exclude a pool with no resolvable registration from the active reward stake". `resolveActiveInstantStakeCredentials` (`libs/cardano-ledger-core/src/Cardano/Ledger/State/Stake.hs`) keys off the credential being registered and carrying a pool delegation and never consults `psStakePools`, which is the premise of this change.
- `rotation.go:553-554` — "The returned distribution is the one actually consumed to build poolInputs/stakeInputs, so its totals are safe to use for RewardSnapshot". After this hunk that holds for `TotalPoolCount` and `TotalDelegators` only.

Non-blocking, but they are the next two comments a reader hits when auditing this denominator.

database/models/reward_state.go
chrisguiney, line 27: Two notes on the bump, neither blocking.

The gate is real: `StaleConsensusStakeSnapshotsExist` (`database/plugin/metadata/sqlstore/live_stake.go:596`) matches every version-1 `pool_stake_snapshot` mark/set/go row and every authoritative version-1 mark `reward_snapshot`, and `backfillRewardLiveStake` (`node.go:2148`) turns that into a startup error directing a rebootstrap. That is every existing database, including the majority whose epochs never excluded a pool. Worth stating as the operator-facing consequence; the description says only "fails closed at startup".

Nothing pins it. Every existing test writes the symbolic constant (`reward_provenance_test.go`, `store_test.go`, `dialect_integration_test.go`), so no test inserts a literal `calculation_version = 1` and asserts the gate trips. The SQL is `<> ?` so it will work, but the fail-closed claim is the one part of this change with no coverage.

chrisguiney (approved):
APPROVE. Verified in an isolated worktree at `c30a2056` and again with `origin/main` (`c7d512a0`) merged in. The branch is 32 commits behind, merges clean, and nothing that landed on main since the merge base touches `ledger/rewards`, `ledger/snapshot`, `ledger/reward_calculation.go` or `database/models/reward_state.go`. The diff matches the description: the behaviour change is real, and every line citation in the description resolves correctly against `origin/main`.

## Reference

`sigma` and `sigma_a` take different denominators, and only `sigma_a` moves here.

- `Shelley/LedgerState/PulsingReward.hs:100` destructures the Go snapshot as `SnapShot activeStake totalActiveStake stakePoolSnapShots`; `:144` passes `totalStake = circulation es maxSupply` and `:152` passes `totalActiveStake` as separate arguments to `mkPoolRewardInfo`.
- `Shelley/Rewards.hs:393-394`: `sigma = poolTotalStake %? totalStake` (circulating supply — feeds `maxPool'`, the operator stake share and the member stake share) versus `sigmaA = poolTotalStake %. unCoinNonZero totalActiveStake`, used only at `:363` in `mkApparentPerformance`.

`ssTotalActiveStake` is provably independent of the pool set: `State/SnapShots.hs:447-451` builds the snapshot with `mkSnapShot activeStake VMap.empty` — an empty pool map — and only then does `resetStakePoolsSnapShot` (`:429-437`) overwrite `ssStakePoolsSnapShot` from `psStakePools`, leaving the total untouched. `mkSnapShot` (`:423-425`) sets `ssTotalActiveStake = sumAllActiveStake ssActiveStake`, and `State/Stake.hs` `resolveActiveInstantStakeCredentials` keys purely off the credential being registered and carrying `stakePoolDelegationAccountStateL`, never on `psStakePools`. The premise holds.

dingo already separates the two: `apparentPerformance` takes `activeStake`, `optimalPoolRewardChecked` takes `totalCirculation`, and the leader/member splits use `ownerStake/poolStake` and `delegatorStake/poolStake` — algebraically the reference's `s/sigma`, in which `totalStake` cancels.

**Exclusion.** `rewardInputsSkippingDegradedPools` drops a pool whose registration dingo could not resolve (missing registration, bad reward-account length, bad credential tag, missing margin, bad owner key). This is a dingo data-quality condition with no reference analogue — the pool is registered on chain and cardano-node both counts it in `ssTotalActiveStake` and pays it. After this change dingo's denominator matches cardano-node, so every surviving pool's `sigma_a`, apparent performance and pool/leader/member rewards match exactly; only the excluded pool's own delegators still diverge. Before it, every pool diverged low. Direction is toward cardano-node.

**Historical epochs.** The value changes only for an epoch that actually excluded a pool; otherwise `rewardDistribution.PoolStakes == effective.PoolStakes`. Version-1 rows are not reinterpreted — see the inline note on the bump. The `calculation_version` columns already default to 0 and are unchanged, so no migration is needed.

**Rounding.** No division order changed. `apparentPerformance` still builds `beta` and `sigma_a` as exact `big.Rat` with a single `floor` at `floorMulChecked(appPerf, optimalReward)`, matching `rationalToCoinViaFloor (appPerf * fromIntegral maxP)`. Only the value of the `activeStake` argument moves.

**Pot invariant.** Rewards rise but stay bounded: `Calculate` errors with "rewards exceed available pot" when `EffectiveRewards + Unspendable > AvailableRewards`, routes the remainder to reserves (`addUndistributedToReserves`, the analogue of `deltaR2 = oldr <-> sumRewards` at `PulsingReward.hs:265`) and unspendable to treasury. No lovelace is created or destroyed.

## Class audit — complete

Producers of `reward_snapshot.total_active_stake`: `ledger/snapshot/rotation.go:439` (fixed here) and `ledgerstate/imported_reward_inputs.go:219`. The second derives the total from the pool set the same way, but `rewardInputBundle.validate` rejects the whole bundle when any delegated stake is unattributed (`unattributedPools > 0`), so it cannot persist an understated denominator, and leaving its exact-equality check at `:339` is correct.

Consumers of the field are only `ledger/reward_calculation.go:873/947/2948`. `GetTotalActiveStake`, `GetPoolDistr2` and `internal/koiosparity` read `epoch_summary`/`pool_stake_snapshot`, which this diff does not touch.

Relaxing `precomputedRewardPoolInputsMatchSnapshot` loses no coverage: its caller at `reward_calculation.go:762` runs `validateRewardCalculatorInputs` immediately after, which still cross-checks each pool's `delegated_stake` against the sum of its `reward_stake_input` rows, and `TotalPoolCount`/`TotalDelegators` stay exact.

## Fail-before evidence

Fix reverted in place, one hunk at a time:

- `rewardDistribution.PoolStakes` → `effective.PoolStakes`: `TestRewardSnapshotActiveStakeKeepsDegradedPoolStake` and `TestHandleEpochTransitionSkipsPoolWithMissingMargin` both fail.
- `>` → `!=` in `validateSnapshot`: `TestCalculateAcceptsActiveStakeAboveThePoolSet` and `TestShrunkActiveStakeUnderCreditsEveryReward` both fail.
- both `>` → `!=` in `reward_calculation.go`: `TestRewardCalculatorInputsAllowExcludedPoolStake` fails.

`TestCalculateRejectsPoolStakeAboveActiveStake` passes with the bound reverted — it pins the new error text and the upper half of the bound, so it is a guard, not new fail-before coverage.

Recomputed the fixture by hand: `maxPool = floor((1000000/1.5) * 0.125) = 83333`; at denominator 1250, `sigma_a = 4/5`, `appPerf = 5/4`, `poolR = floor(104166.25) = 104166`, leader `1000 + floor(0.55*103166) = 57741`, member `floor(0.45*103166) = 46424`; at 1000, `83333 / 46283 / 37049`. Matches the assertions.

## Checks

All local, isolated worktree, isolated `GOLANGCI_LINT_CACHE`, on the merged tree.

| check | result |
|---|---|
| `go build -tags dingo_extra_plugins ./...` | 0 |
| `go test -race -timeout 40m ./ledger/... ./database/...` | 1 — only `ledger/leader` `TestThresholdPreservesProtocolBoundary`, reproduced identically on `origin/main` `c7d512a0`, unrelated to this PR. `./ledger`, `./ledger/rewards`, `./ledger/snapshot` and all `./database/...` pass |
| `go test ./internal/architecture ./internal/docsparity` | 0 |
| `golangci-lint run ./...` in all three modules (root, `examples/dingo-gov-lens`, `internal/test/antithesis`) | 0 issues each |
| `GOOS=windows golangci-lint run ./...` | 0 issues |
| `go build ./... && go test ./...` in both nested modules | 0 |
| `gofmt -s -l` on every changed file | clean |

Not run: the Koios parity replay against real chain data, which is the actual acceptance check for a reward-denominator change — the unit tests above do not settle it. `nilaway` and `modernize` are not installed on this host; the CI `lint` job covers them.

## Bot standing on `c30a2056`

`c30a2056` is both the current head and the branch's only commit. cubic submitted a review record anchored to it: "No issues found across 10 files". CodeRabbit's comment body states it reviewed `f86d64f4..c30a2056` across all 10 files with "No actionable comments" — a real review, not a quota block. Neither bot opened an inline thread.

Correction to this review as first posted: @cryptodj413's `COMMENTED` review landed four minutes ahead of mine and I had already built this body, so the original text said no inline threads existed. Six do. I read all six afterwards; none changes the disposition. The `authoritative = TRUE` filter in `StaleConsensusStakeSnapshotsExist` is the one I had not examined — it is unreachable as a stale-row escape, because the fallback boundary also writes `pool_stake_snapshot` rows that the first clause covers unconditionally, and `authoritativeMarkRewardSnapshotExists` (`ledger/snapshot/manager.go:534`) already returns false on a version mismatch. The `<=` bound admitting a proportionally reduced row set is real but requires arbitrary corruption that keeps per-pool cross-sums, pool count and delegator count all self-consistent. Both stay non-blocking.

Two non-blocking items inline; neither blocks merge.

cryptodj413 (approved):

Issue #3976, opened 2026-09-06
ledger/eras: three Conway tests fail on main, so every branch inherits red CI


chrisguiney:
## Summary

Three tests in `ledger/eras` fail on unmodified `main`. Every branch that
merges `main` inherits them, so unrelated PRs show red CI for a defect they
did not introduce.

```
--- FAIL: TestConwayUtxoValidationRuleIndexesArePinned
--- FAIL: TestValidateTxConwayGenuinelyMissingInputStillRejected
--- FAIL: TestValidateTxConwayGenuinelyUnbalancedStillRejected
FAIL    github.com/blinklabs-io/dingo/ledger/eras
```

Reproduced on `origin/main` at `90504583` with a clean worktree:
`go test ./ledger/eras/ -count=1`.

## The failure

```
conway rule index 29 should resolve to UtxoValidateBadInputsUtxo,
  got github.com/blinklabs-io/gouroboros/ledger/conway.UtxoValidateNoCollateralInputs

conway rule index 32 should resolve to UtxoValidateValueNotConservedUtxo,
  got github.com/blinklabs-io/gouroboros/ledger/conway.UtxoValidateRequiredRedeemers
```

`TestConwayUtxoValidationRuleIndexesArePinned` pins positions in
`conway.UtxoValidationRules`, and the pinned numbering does not match the
composed list in the gouroboros version `main` uses.

## It arrived already red

The test was added by #3818 (`3922fb9b`, 2026-09-05), which landed **after**
the `v0.202.9` bump (`0234e66b`). Checking out `3922fb9b` itself and running
the test fails there too — so it was merged red rather than broken by a later
change.

The two `ValidateTxConway…` failures are in the same file and share the
fixture, so they are very likely the same root cause.

## Not caused by the v0.202.8 → v0.202.9 bump

I checked, because four of my gouroboros PRs are in that range.
`git diff v0.202.8..v0.202.9 -- 'ledger/*/rules.go'` contains **no**
descriptor-list edits in any era — the changes are rule *bodies*
(`TxSize`, the collateral phase-2 guard, ex-units, the pool deposit) and
tests. The composed rule order is unchanged between those tags, and the
pre-bump commit does not contain the test at all.

So the pinned numbers appear to have been written against a different
composition than the one in tree, rather than a rule having been inserted
upstream.

## Why it matters beyond CI noise

The indexes are operator-visible: `ValidateTxConway` reports
[...]

Issue #3848, opened 2026-09-04
test: five lifecycle tests fail under parallel suite load and pass in isolation


chrisguiney:
## Summary

Five distinct tests failed under parallel suite load during one day of work on
unrelated branches, and every one of them passes repeatedly in isolation. None
were in a package the failing branch modified. The practical effect is that a
red `go-test` job on a PR carries very little information, which is corrosive:
the correct response to a failure becomes "re-run it" rather than "read it".

This is not the same as #3541, which is one specific test with a
cardinality-dependent cost assertion. These are timing, port, and signal tests
whose common factor is contention.

## Observed

| test | package | how it failed |
| --- | --- | --- |
| `TestConcurrentBindStopNeverLeavesThePortBound` | `internal/apilistener` | CI, `listener_test.go:667`, unexpected error |
| `TestReflectionDisabledByDefault` | `midnight/server` | CI under `-race`, `server_test.go:279`, `EOF` |
| `TestEntrypointForwardsSignalsDuringMithrilBootstrap` | `bin` | local full suite |
| `TestAddrClearsAfterStopTimesOut` | `bark` | local full suite |
| `TestRequestRangeExcessBatchDoneNotAppliedToNextRequest` | `protocol/blockfetch` | local full suite |

Each was then run three consecutive times in isolation and passed every time.

The two CI failures landed on `fix/3844-validity-bound-past-horizon`, whose only
changes are in `ledger/slot_time_converter.go`, `ledger/slot.go` and
`ledger/view.go`. `git diff origin/main...HEAD -- internal/apilistener/` and the
same for `midnight/` are both empty.

One run is worth highlighting: a full local suite and a CI run on the **same
commit** failed on **different** tests — locally `bin`, on CI
`internal/apilistener`, with each passing in the other environment. The failures
follow load, not code.

## Shape

All five are lifecycle tests: bind a port, send a signal, wait for a goroutine
to release something, assert within a deadline. They are the tests most exposed
to a loaded machine, and `-race` makes it worse by slowing everything down.
`internal/test/testutil/` already exists for this (`WaitForCondition`,
`RequireReceive`, and the repo rule against `time.Sleep` for synchronization),
so the fix for most of them is likely mechanical rather than a redesign; #2265
is the same idea applied to mempool and ledger tests.

The port-binding one may need more than a longer deadline: a fixed or
racily-chosen port on a shared runner will collide regardless of timeout.

## Why it is worth fixing

[...]

PR #3982, opened 2026-09-06
test(dbtest): bound the disk a test database reserves

### LINKED ISSUE #3980: bark: Windows CI runs out of disk opening a badger store per test
## What happens

`bark.TestListSnapshotsPaginates` fails on the `go-test (Windows)` job with a
disk-space error rather than an assertion failure:

```
badger ERROR: Got error: While opening memtable: C:\Users\RUNNER~1\AppData\Local\Temp\
  TestListSnapshotsPaginates.../001/blob/00001.mem
  err: truncate ...\00001.mem: There is not enough space on the disk.
error while truncation. Cleaning up...

--- FAIL: TestListSnapshotsPaginates (1.94s)
    bark/database_test.go:53 -> bark/database_test.go:648
    Received unexpected error:
      start plugin storage.blob/badger: badger deferred start: badger open on-disk:
      cannot create memtable
```

A second symptom follows immediately, caused by the first:

```
testing.go:1464: TempDir RemoveAll cleanup:
  unlinkat ...\00001.mem: The process cannot access the file because it is being
  used by another process.
```

so the failed store is left open and its temp directory cannot be reclaimed —
which makes the pressure worse for every test that runs after it in the package.

## Why it is not a test bug

The assertion never ran; the badger store failed to open. `bark`'s tests each
stand up a full on-disk badger blob store in `t.TempDir()`, and badger
preallocates a memtable file per store. The Windows runner's temp volume is much
smaller than the Linux one, so the package's cumulative store count exhausts it.
Linux and macOS pass the same package.

## Observed on

`go-test (Windows) (1.26.x)`, run 34005242250, on a branch whose only change is a
test file under `ledger/eras` — nothing in `bark`, `database`, or any storage
path. So this is pre-existing on `main` and independent of the branch that
surfaced it.

## Suggested direction

Cap badger's memtable/value-log sizing in the test helper that builds these
stores (`bark/database_test.go:53`), rather than accepting badger's production
defaults for every temporary store. Ensuring the store is closed on a failed
open would also stop one failure from cascading into the `TempDir` cleanup
errors that follow.

chrisguiney:
Fixes #3980

## Problem

Badger truncates its value log and memtable to their configured sizes when it
opens a store, and maps the value log at twice `ValueLogFileSize` so the last
entry always fits. With the production defaults (1GiB value log, 128MiB
memtable) each test database reserves 2GiB.

A sparse filesystem makes that free, which is why this only ever failed on
Windows. Once a package builds several stores the NTFS runner runs out of disk:

```
badger ERROR: While opening memtable: ...\001\blob\00001.mem
  err: truncate ...: There is not enough space on the disk.
--- FAIL: TestListSnapshotsPaginates
  start plugin storage.blob/badger: ... cannot create memtable
```

The failed open leaks the file handle, so `TempDir` cleanup fails behind it and
the pressure compounds for every test after it.

## Change

`dbtest.NewDatabaseWithOptions` applies `testutil.BadgerBlobConfig` as defaults
merged key by key, rather than only when the caller supplied no `Blob.Config`
at all. The badger provider decodes its config from the production defaults
(`database/plugin/blob/badger/provider.go`), so a caller config that sets an
unrelated knob and omits the size keys decodes back to the 1GiB value log and
reserves 2GiB. Caller keys still win, so a test that needs production sizing
asks for it explicitly.

`cloudPrimaryTestProvider` in `internal/dblifecycle/manager_test.go` called
`badger.New` with no size options and reserved 2GiB. It now passes testutil's
sizes, like the other direct `badger.New` sites in the test tree.

## Test

`TestNewDatabaseBoundsBadgerFileReservation` and
`TestBoundedBadgerSizesSurviveAPartialCallerConfig` walk a fresh test
database's data directory and assert no file exceeds twice
`testutil.TestBadgerValueLogFileSize`. Walking the directory rather than
statting the value log also bounds the memtable WAL, the file #3980 failed to
open. `os.Stat` reports a truncated file's logical size on every platform,
including sparse ones, so both discriminate everywhere.

Only the partial-config test discriminates on the merge behavior: under a
`blobConfig == nil` gate it measures 2147483648 bytes and fails; with the
per-key merge it measures 33554432 bytes and passes.

`TestNewDatabaseReservesASmallValueLog` stated the same bound by statting
`blob/000001.vlog`, so the walk subsumes it. It is removed and its stat is kept
inside the walk as a non-vacuity guard: the walk alone passes if the store
never writes to disk.


chrisguiney:
CI confirms the fix on the platform that showed the bug.

The `go-test (Windows)` job on this branch now fails **only** on the three
`ledger/eras` tests, which are the pre-existing `main` breakage (#3976, fixed by
#3977) inherited by every branch:

```
FAIL	github.com/blinklabs-io/dingo/ledger/eras
```

No `bark` failure, and no `There is not enough space on the disk` anywhere in
the run — which is what this change is for. The remaining red is not from this
branch and clears when #3977 lands.

internal/test/dbtest/database.go
```diff
@@ -142,6 +160,10 @@ func NewDatabaseWithOptions(
 	if blobRegister == nil {
 		blobRegister = badger.RegisterProvider
 	}
+	blobConfig := opts.Blob.Config
+	if blobConfig == nil && blobName == "badger" {
```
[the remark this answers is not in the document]
chrisguiney: Valid, and fixed in 6d3560fa.

Reproduced first rather than taking it on faith: `TestBoundedBadgerSizesSurviveAPartialCallerConfig` passes `Blob.Config: map[string]any{"gc": false}` — an unrelated knob, nothing about sizes — and reserved **2147483648 bytes**, exactly the 2GiB this PR removes. So the gate was real, not theoretical, and there are already callers in the tree passing partial blob configs.

The fix merges the bounded values as defaults with the caller's keys winning, rather than only substituting when the whole config is nil:

```go
merged := BoundedBadgerConfig()
for k, v := range blobConfig {
    merged[k] = v
}
blobConfig = merged
```

That keeps the escape hatch — a test that genuinely wants production sizing can still set the keys explicitly — while the default applies to every badger test database. I updated `BoundedBadgerConfig`'s doc comment to state the key-by-key precedence, since the old wording described the nil-only behaviour.

Both disk tests now pass, as do `./internal/test/dbtest` and `./database`. One unrelated failure worth naming so it is not read as this change: `bark.TestAddrClearsAfterStopTimesOut` fails only when `./bark` runs alongside other packages and passes in isolation — it is one of the five tests listed by name in #3848.

internal/test/dbtest/badger_disk_test.go
```diff
@@ -0,0 +1,97 @@
...
+			Config: map[string]any{"gc": false},
+		},
+	})
+	require.NoError(t, err)
+	require.NotNil(t, db)
+
+	var largest int64
+	var largestPath string
+	require.NoError(t, filepath.WalkDir(dataDir,
+		func(path string, d fs.DirEntry, err error) error {
```
wolf31o2: Extract the repeated WalkDir/largest-file calculation into a shared helper and use it from both reservation tests. Keeping the two copies separate allows the tests to drift and weakens the guarantee that they measure the same reservation rule.

wolf31o2 (changes requested):
Code review: changes requested. The test change is functionally useful, but the new helper logic is duplicated and should be consolidated.

wolf31o2 (dismissed):

internal/test/dbtest/badger_disk_test.go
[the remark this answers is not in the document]
chrisguiney: Agreed, and done in `cfc544ff`. Both tests now call a shared `largestFile(t, dir)` helper; the file has one `WalkDir` instead of two.

I checked the refactor did not weaken either test rather than assuming it: raising `boundedBadgerFileBytes` to 1024MiB makes **both** fail with `"2147483648" is not less than or equal to "33554432"`, and restoring it makes both pass. So they still measure the reservation and still discriminate.

The helper carries the reason the measurement works, which was previously implicit: `os.Stat` reports a truncated file's length even where the filesystem stores it sparsely, which is why this catches the 2GiB reservation on Linux and macOS and not only on NTFS where it costs real blocks.

chrisguiney: Agreed, and done in `cfc544ff`. Both tests now call a shared `largestFile(t, dir)` helper; the file has one `WalkDir` instead of two.

I checked the refactor did not weaken either test rather than assuming it: raising `boundedBadgerFileBytes` to 1024MiB makes **both** fail with `"2147483648" is not less than or equal to "33554432"`, and restoring it makes both pass. So they still measure the reservation and still discriminate.

The helper carries the reason the measurement works, which was previously implicit: `os.Stat` reports a truncated file's length even where the filesystem stores it sparsely, which is why this catches the 2GiB reservation on Linux and macOS and not only on NTFS where it costs real blocks.

arepala-uml (changes requested):
Requesting changes for two reasons -- a real merge conflict, and a process concern worth raising directly.

**Blocking: merge conflict with main.** `origin/main` already contains an independent fix for the same issue (#3980) via #4131 (merged today), but main's version only applies the bounded Badger config `if blobConfig == nil` -- which is exactly the clobber bug Cubic caught earlier in *this* PR's own review history and that you fixed with a proper per-key merge (`6d3560fa`). Main's newly-merged version reintroduces the bug this PR already solved. This needs a manual reconciliation (likely consolidating with main's `testutil.BadgerBlobConfig`, keeping this PR's per-key merge logic), not a simple rebase.

**Process concern:** wolf31o2's `CHANGES_REQUESTED` review on this PR shows as dismissed, and the dismissal is attributed to you (the PR author) rather than to wolf31o2 re-reviewing and approving. Per this workspace's review rules, dismissing another reviewer's review needs to be done by an authorized party with the rationale recorded, not unilaterally by the PR's own author -- could you ask wolf31o2 to re-review and either resolve or formally dismiss their own review, so there's a valid approval on record? Right now `reviewDecision` is `REVIEW_REQUIRED` with no human approval at any commit.

**What's verified correct:** the merge-as-defaults logic is right -- it builds from the bounded config and overwrites with every caller-supplied key, so caller keys always win and unrelated knobs are preserved while sizes still get bounded by default. Reproduced the numbers directly: the new test fails with the claimed 2147483648-byte file against the pre-fix commit, and passes with the claimed 16777216-byte file at head. Confirmed no interaction with the Badger GC-metrics tests from #3863 -- that package doesn't go through `dbtest` at all. Full `dbtest`/`bark`/`database`/`badger` test matrix clean under `-race`, `golangci-lint` clean.

Once the main conflict is reconciled and there's a valid human approval on record, this looks ready.

internal/test/dbtest/badger_disk_test.go
chrisguiney: Still addressed after the merge with main, at `d1b03777`.

`badger_disk_test.go` has a single `filepath.WalkDir`, inside `largestFile(t, dir)`, called from both reservation tests. The bound they compare against is now derived from `testutil.TestBadgerValueLogFileSize` rather than a separate literal, so it cannot drift from the sizing `TestNewDatabaseReservesASmallValueLog` asserts either.

chrisguiney:
## Reconciled with main

`origin/main` is merged in at `d07cca21`. `internal/test/dbtest/database.go` was
the only conflict, resolved toward this branch's per-key merge with the sizes
now sourced from `testutil.BadgerBlobConfig` instead of a second local constant.
`dbtest.BoundedBadgerConfig` is removed, so there is one definition of the test
sizing. The test bound is derived from `testutil.TestBadgerValueLogFileSize`
rather than a separate literal, so it cannot drift from what
`TestNewDatabaseReservesASmallValueLog` asserts.

## main's gate is a defect, not only a conflict

`internal/test/dbtest/database.go:147` on `main` applies the bounded config only
under `blobConfig == nil`. `database/plugin/blob/badger/provider.go:44-45` seeds
the provider config from `DefaultValueLogFileSize` (1073741824) and
`DefaultMemTableSize` (134217728), so a caller config that sets an unrelated
knob and omits the size keys decodes back to production sizing.

Restoring that gate in place at the merged head reproduces it:

| gate | largest reserved file | `TestBoundedBadgerSizesSurviveAPartialCallerConfig` |
|---|---|---|
| `blobConfig == nil && blobName == "badger"` | 2147483648 bytes | fails |
| `blobName == "badger"` (per-key merge) | 33554432 bytes | passes |

`TestNewDatabaseReservesASmallValueLog` passes under both gates, so main's
existing coverage does not reach this case. Merging this branch removes the gate
from main; no separate change is needed for it.

## Class audit

Every `badger.New` outside `database/plugin/blob/badger`:

- `bark/archive_test.go:202`, `database/lifecycle/restore_gc_test.go:77`,
  `database/lifecycle/restore_remote_test.go:164` and `:199` already pass
  `testutil.TestBadgerValueLogFileSize` and `TestBadgerMemTableSize`.
- `internal/dblifecycle/manager_test.go:124` (`cloudPrimaryTestProvider`) passed
  neither and reserved 2147483648 bytes. Fixed here; the same option set
  measures 33554432 bytes with the sizes applied.
- `internal/test/archive-demo/cmd/inspect-blob/main.go:55` opens a real store
  rather than a test fixture, so it keeps production sizing.

Every `plugin.Resolve` of the badger blob capability in test code already passes
`testutil.BadgerBlobConfig()`.

## Review state

No review was dismissed in this round. The outstanding `CHANGES_REQUESTED` from
wolf31o2 is re-requested rather than dismissed, so it is resolved by its author.
The helper thread on `badger_disk_test.go` is resolved: the file has one
`filepath.WalkDir`, inside `largestFile`, used by both reservation tests.

chrisguiney:
`go-test (macOS)` at `d1b03777` failed on two frontier-flap tests, both added by #4048: `chainselection/frontier_flap_test.go:253` (`"[]" should have 1 item(s), but has 0`) and `ledger/chainsync_frontier_flap_test.go:164` (`"0" is not positive`). Both assert the scenario published at least one selection event, and both observed zero at 0.00s.

Neither is reachable from this diff, which is three files under `internal/test/dbtest` and `internal/dblifecycle`. `chainselection` imports neither package: built with `-trimpath`, its test binary is byte-identical at this head and at `d07cca21` (`206118fafb60e7f809b58494118a8a90e96f1ed73459b0085a94740698c6fae8`), so this branch cannot change that package's behavior. Both tests are nondeterministic on `main` itself and pass on repeat.

chrisguiney:
Merged current `origin/main`; the merge was conflict-free and the branch is level with it.

Reconciliation with #4131: main's version applies `testutil.BadgerBlobConfig` only when `Blob.Config` is nil. The merge at `d0d5e67f` resolved onto main's helper and kept this branch's per-key merge, dropping `dbtest.BoundedBadgerConfig`, so one implementation survives — `NewDatabaseWithOptions` builds from `testutil.BadgerBlobConfig()` and overwrites with every caller-supplied key. Nothing on main since then touches `internal/test/dbtest`, `internal/test/testutil`, or `internal/dblifecycle`.

That reconciliation left two tests stating the same reservation rule. `b7a29c85` drops main's `TestNewDatabaseReservesASmallValueLog` — its `blob/000001.vlog` stat against `2*testutil.TestBadgerValueLogFileSize` is what the directory walk in `TestNewDatabaseBoundsBadgerFileReservation` already covers — and keeps that stat inside the walk as its non-vacuity guard, since the walk alone passes if the store never writes to disk. Both reservation tests now live in package `dbtest`, and `AGENTS.md` names them.

Reverting the per-key merge back to the `blobConfig == nil` gate in place makes `TestBoundedBadgerSizesSurviveAPartialCallerConfig` fail at 2147483648 bytes against the 33554432-byte bound. Pointing the guard at a nonexistent value log fails both tests.

On the review record: wolf31o2's `CHANGES_REQUESTED` (review 5125745020) has not been dismissed and still stands. The review dismissed at `cfc544ff` was a later `APPROVED` (5125747581). No approval is on record at any commit.

wolf31o2 (approved):

wolf31o2 (approved):

arepala-uml (approved):

Issue #3170, opened 2026-08-12
ledger: post-rollback continuation blocks spend inputs whose producer is missing from the applied ledger, wedging the node (Musashi/Leios; cross-fork audit fires in the wild)


chadaibbhot-git:
## Summary

On the Musashi Leios testnet (Water, `prototype-2026w32`), two independent Dingo
nodes permanently wedged at the same chain position because a block on the chain
they were following spends an input whose producing transaction is **not present in
the local applied ledger**. With the default `--strict-utxo-validation=true` the node
refuses to apply the block, retries forever, and never advances again — a restart
re-wedges at the identical block.

The cross-fork continuation audit added for #3005 fires here and captures the full
record, including the fork point, the delivering peer, and the identified-but-absent
producer transaction.

## Environment

- Dingo `main` @ `d5ff094b` plus a local gouroboros genesis patch required to parse the
  Musashi Water genesis (unrelated to this bug; PR blinklabs-io/gouroboros#1969).
- Network: Musashi Water, `prototype-2026w32`, magic 164, Leios run mode
  (`serve --network musashi --run-mode leios`).
- Two independent Dingo nodes: a block producer and a plain node, different hosts.
- Control: two Haskell `cardano-node` `prototype-2026w32` nodes on the same network.

## Symptom 1 — the wedge (default strict mode)

Both Dingo nodes froze at applied slot `461717`, repeating every ~2s indefinitely:

```
level=WARN msg="block processing failed, restarting pipeline"
error="process block batch: record transaction: consumed utxo
f89bd9c358a14b2d8325d26f7980b44fea1a4da596f037f9fa5caceea6aa8aea#1 not present in
applied ledger at slot 461772: refusing to recover it from the blob store and persist
an input-conservation violation (issue #3005): utxo not found"
```

- Deterministic: a container restart re-wedges at the same block.
- Both nodes independently reached the identical stuck slot (`461717`), sampled
  repeatedly with zero advance.
- The Haskell control nodes crossed this region and stayed at the network tip
  throughout (verified simultaneously).

## Symptom 2 — the continuation audit, with the full record

Running with `--strict-utxo-validation=false` (see workaround) the node advances but
the underlying ledger gap remains visible; the #3005 audit reports it:

```
level=ERROR msg="continuation block spends an input with no producer on the local applied chain"
component=ledger
peer=<redacted-local>:60806<->52.58.138.208:3001
block_slot=473738
[...]

PR #3611, opened 2026-08-28
fix(chain): bound and signal live primary-chain rewinds

cryptodj413:
## Problem

`RewindPrimaryChainToPoint` pruned the persistent primary chain back to a
point by deleting blocks directly. It had no rollback-depth bound and never
published `ChainRollbackEvent`/`ChainForkEvent` or woke/marked chain
iterators. It is used both by startup reconciliation and by the live
primary-chain/ledger divergence reconciler
(`ReconcileLivePrimaryChainLedgerDivergence`, used by the plateau watchdog and
by chainsync/fork-resolution's over-K fallback), so a live rewind could
silently truncate the chain past the configured security parameter K and out
from under NtC chain iterators.

## Changes

- `RewindPrimaryChainToPoint` now delegates to `Chain.Rollback`, so it shares
  the same bound and side effects as a live rollback: it is rejected outright
  (no state touched) when the rewind would exceed K, and a successful rewind
  publishes `ChainRollbackEvent`/`ChainForkEvent` and wakes/marks chain
  iterators exactly once.
- `reconcileLivePrimaryChainLedgerDivergence` now treats a rejected over-K
  rewind (`ErrRollbackExceedsSecurityParam`) the same as "no divergence
  found", so the existing over-K handling in `chainsync.go` and
  `tryResolveFork` (reject the peer chain, request a fresh intersection)
  takes over instead of the caller receiving a raw, unexpected error.
- Removed the now-unused batched-delete implementation and its
  `primaryChainRewindBatchSize` constant.
- `ARCHITECTURE.md`: documented the new bound/signal behavior in the
  "Ledger/chain reconciliation" section and the plateau-watchdog bullet.
  `DATABASE.md` is unaffected (no schema/storage-provider change).

## Tests

- `chain/chain_test.go`: added
  `TestRewindPrimaryChainToPointRejectsOverLimitRewind` (over-K rejection
  leaves chain/db state untouched),
  `TestRewindPrimaryChainToPointSignalsRollback` (in-bound rewind publishes
  exactly one `ChainRollbackEvent` and delivers a rollback marker to a live
  chain iterator), and `TestRewindPrimaryChainToPointConcurrentRewinds`
  (20 concurrent callers rewinding to the same point, run with `-race`).
- Updated `TestRewindPrimaryChainToPointPrunesPersistentTail` and
  `TestRewindPrimaryChainToPointDoesNotDeadlockWithIterator` to configure a
  security parameter, since the rewind now requires one (previously bypassed
  entirely).
[...]

PR #3986, opened 2026-09-06
fix(ledger): make the continuation audit see Leios endorser-block producers

chadaibbhot-git:
## What

The cross-fork continuation audit added for #3005 reports

```
level=ERROR msg="continuation block spends an input with no producer on the local applied chain"
```

for inputs whose producing transaction is canonical, present, and correctly
applied by the same node seconds later. On a Leios network every
endorser-resident spend after a fork is reported. The audit is wrong, not the
ledger. This PR makes the audit Leios-aware.

## Why the audit misses these producers

`ledger/continuation_audit.go` builds its in-window producer set from the
ranking block body:

```go
txs := e.Block.Transactions()
...
for _, tx := range txs {
	window.producedTxs[string(tx.Hash().Bytes())] = struct{}{}
}
```

On the Leios cert-driven path a **certifying ranking block carries an empty
body** (71 bytes on the prototype network). Its transactions are the certified
endorser block's, which arrives over `leios-fetch` as a separate artifact and is
applied by `LedgerState.applyEndorserBlock` (`ledger/state.go`, the
`dijkstraEraGate` branch of `ledgerProcessBlock`) at **ledger-apply** time. The
audit runs at **blockfetch** time (`ledger/chainsync.go`,
`flushPendingBlockfetchBlocksDeferred`, under `chainsyncBlockfetchMutex`).

So all three resolution paths in `continuationInputHasProducer` miss an
endorser-resident producer while the ledger is behind the blockfetch queue:

1. `window.producedTxs` — populated only from the ranking-block body, which is
   empty on a cert-RB. This is the bug.
2. `ls.db.UtxoByRef` — the UTxO does not exist yet; `applyEndorserBlock` has not
   run.
3. `ls.db.GetTransactionByHash` — the transaction row does not exist yet, same
   reason.

The audit's own soundness argument said:

> An input can therefore be resolved without a chain scan: it is legitimate when
> its producing transaction was created by a block already seen in this window
> (fetched, on the chain, not yet applied) …

That is exactly right for pre-Leios blocks, where "the block was fetched"
implies "its transactions are in hand". It does not hold where a block's
transactions are a separate artifact fetched over a different mini-protocol and
applied by a later stage.

Being behind is not an edge case here: an endorser-block apply backlog is the
normal condition on a Leios node, and it is precisely the condition in which a
rollback arms the audit.

## Field observation

Two independent nodes on the same Leios prototype network (magic 164), two
different fork windows, same signature. Timeline from one of them, abstracted:

| offset | event |
|---|---|
| T-29m | endorser block for slot 2610537 fetched complete, manifest + 1347/1347 transactions |
| T+0 | 16 × `ERROR continuation block spends an input with no producer…` across blocks at slots 2610834 / 2610858 / 2610876 |
| T+16s | `INFO applied Leios endorser block transactions slot=2610577 eb_slot=2610537 eb_txs=1347` |
| T+23s | `INFO applied Leios endorser block transactions slot=2610590 eb_slot=2610577 eb_txs=959` |
| T+67s | `INFO applied Leios endorser block transactions slot=2610832 eb_slot=2610793 eb_txs=641` |

Every named producer resolved, against an independent chain index, to a block
with `has_leios_cert: true` and a 71-byte body — a cert-RB whose 1347 / 959
transactions are the certified endorser block's. Every *reporting* block was an
ordinary ranking block with a real body. There was no
`certified Leios endorser block unavailable` and no dropped required reference
in the window; the node then tracked the network tip unaided. The second node
was at the tip throughout and never wedged, and produced the same message. The
UTxO set was never short.

Severity is diagnostic-only — the audit never rejects a block — but the message
is an operator-facing `ERROR` that names a transaction and an input and asserts
ledger corruption. It is the first thing one greps for during a Leios stall and
it points away from the real problem. It also defeats the audit's purpose: with
a guaranteed false-positive source on the Leios path, a *true* unresolved
producer is indistinguishable from noise.

## Fix

`auditContinuationBlock` now resolves each audited block's endorser block the
same way `ledgerProcessBlock` does, and records its transactions as in-window
producers:

- `ls.leiosEndorserBlockForApply(block)` picks the reference — the block's own
  announcement on the CIP path, the parent's announcement on the cert-driven
  path — and returns the slot that occurrence must be bound to.
- `ls.config.EndorserBlockProvider(ebHash, ebSlot)` supplies the transactions
  for exactly that `(hash, slot)` occurrence. The common path is an in-memory
  cache hit plus a hash per endorser transaction, since the endorser block is
  usually in the provider cache by the time the referencing ranking block is
  fetched. It is not guaranteed to be I/O-free, though: the provider resolves to
  `Ouroboros.EndorserBlockTxsByHash` → `lookupLeiosEndorserBlock`, which falls
  back to `loadLeiosEBFromDB` — a blob-store manifest read, a decode and a
  transaction load — for an occurrence whose in-memory entry has expired. That
  is the case worth naming, because the audit is armed precisely during the
  backlog in which entries do expire, and it holds
  `chainsyncBlockfetchMutex` throughout. It stays bounded on both axes: at most
  `continuationAuditMaxEndorserBlocksPerBlock` per audited body, and once per
  `(hash, slot)` for the life of the window.
- Transaction ids come straight from the envelope: blake2b-256 over the
  transaction body CBOR, which is the definition `DijkstraTransaction.Id`
  inherits from `BabbageTransactionBody.Id`. A full per-transaction decode was
  rejected because the audit holds `chainsyncBlockfetchMutex` and an armed
  window can span 512 blocks of thousand-transaction endorser blocks.
  `TestEndorserBlockTxIdsMatchDecodedTransactionHashes` pins the cheap
  derivation against the decoded transaction hash so the two cannot drift.
- The CBOR-in-CBOR unwrap that `applyEndorserBlock` already performed is
  extracted to `decodeEndorserTxEnvelope` and shared, so the audit and the apply
  path cannot disagree about what an endorser transaction entry is.

**When the endorser block is not cached yet**, the producer set is knowingly
incomplete. The window is marked (`endorserProducersPending`, sticky) and
unresolved inputs are reported as *inconclusive* at `Debug` instead of as
missing producers at `Error`. A report from an incomplete set is a false
positive, and a false positive here is worse than silence: it asserts ledger
corruption on a healthy node.

Non-Leios chains are unaffected: a header that neither announces nor certifies
an endorser block yields no reference, and the offline replay path configures no
endorser-block provider at all.

## Cost model

Resolving an endorser block costs a parent-block read (cert-driven path) plus a
hash of every one of its transactions, all under `chainsyncBlockfetchMutex`. The
audit is armed exactly when a node is in a dense endorser-block backlog, so that
work is scheduled carefully:

- **Lazy.** Classifying a block into an endorser-block reference is header-only
  and free, so that stays eager. Turning a reference into producers is deferred
  until an audited input fails to resolve from the ranking-block producer set
  *and* the ledger. A window in which nothing spends an endorser-resident output
  performs zero parent reads, zero provider lookups and zero hashing. The
  cert-driven reference is retained as its parent hash alone, so the referencing
  block and its CBOR need not be held.
- **Once per window.** References are deduplicated when queued and merged
  occurrences memoized by `(ebHash, ebSlot)`, so a closure certified by N ranking
  blocks costs one parent read, one provider lookup and one hashing pass rather
  than N. An endorser block not cached yet is requeued in *resolved* form, so a
  later attempt costs only the provider lookup.
- **Budgeted.** `continuationAuditMaxEndorserBlocksPerBlock` (32) is the
  analogue of `continuationAuditMaxInputsPerBlock`: one audited body may not
  resolve an unbounded number of endorser blocks. Overflow stays queued for a
  later body, the window reports inconclusive while its set is knowingly short,
  and the stop is counted as `result="skipped_budget"`.
- **Retryable.** A parent that is not in the block store yet is a state, not a
  verdict — the same read-ahead this audit exists to tolerate — so
  `models.ErrBlockNotFound` requeues the reference for the next audited body
  rather than dropping it for the life of the window. Each reference is probed
  at most once per body, so a permanently absent parent costs one index miss per
  audited block. Anything else drops the reference, counts
  `result="ref_unresolvable"` and logs once per window.
- **Deduplicated on requeue too.** A deferred certified reference is keyed by
  its parent hash, but the endorser-block reference is (announced hash,
  announcing slot): two ranking blocks with different parents at one slot — on
  either side of a fork, which is the state a window is armed in — name the same
  occurrence under two queue keys, and converge only once resolved. Requeueing
  therefore dedupes against the queue as well, so a still-unfetched endorser
  block cannot be queued twice and probed twice per body.

"The producer set is knowingly short" is derived from the queue rather than
latched, so a window whose references all resolve becomes complete again and can
go back to diagnosing a genuine missing producer. Before, one transient miss
silenced a window permanently.

Net per audited body: nothing at all unless one of its inputs is unresolvable
without an endorser block, and at most 32 resolutions when it is, each the first
and only time that closure is touched in the window.

## Producer-set cap

Including endorser transactions changes the arithmetic. A ranking-block-only
window held a few hundred transactions per block; a certified endorser block on
the prototype network carries 100–2500, so a full `continuationAuditBlockBudget`
(512 block) window can offer well past a million producers against the old
100k cap.

`continuationAuditMaxProducedTxs` is raised **100k → 250k** (~16 MB of
transient set for a diagnostic) rather than to that ceiling, because unbounded
growth is the worse failure. A busy Leios window will therefore still disarm
partway through — so the disarm is now **explicit** rather than silent: it logs
at `Warn` (was `Debug`) with `max_produced_txs`, and increments the outcome
counter with `result="disarmed_cap"`. "The audit stopped covering this node"
should not read the same as "this node is clean".

The cap bounds the **size of the set**, so it is enforced per insert and only
ids the set does not already hold count toward it. Charging a repeat would
disarm a window whose producer set never grew, and repeats are ordinary here:
the same transaction can appear in more than one endorser block
(`deduplicateEndorserBlockTransactionIndexes` exists for exactly that), and an
endorser block is content-addressed, so the same closure can be certified from
more than one ranking block in a window.

## Observability

New `dingo_ledger_continuation_audit_outcomes_total{result=…}`, with all four
children pre-materialized at registration so every verdict is exported from
process start (an absent series and a zero series read very differently when the
question is "is this node reporting missing producers"), and so the audit pays
no label lookup per input under the mutex:

| `result` | unit | meaning |
|---|---|---|
| `clean` | input | resolved to a producer |
| `missing_producer` | input | no producer; still logged at `Error`, and `dingo_ledger_continuation_input_unresolved_total` still increments |
| `inconclusive_eb_pending` | input | producer set knowingly incomplete (certified endorser block not fetched yet) |
| `disarmed_cap` | window | producer set hit the cap and the window disarmed |
| `skipped_budget` | block | more endorser blocks queued than one body may resolve; the rest stay queued |
| `ref_unresolvable` | reference | an endorser-block reference given up on for good, because resolving it failed for a reason retrying cannot fix |

A node whose `inconclusive_eb_pending` dominates is telling the operator the
audit is not covering it, which is the honest reading of an endorser-block
backlog.

The audit's soundness comment now states the Leios caveat instead of implying
"block fetched ⇒ transactions in hand" holds everywhere.

## Tests

New `ledger/continuation_audit_leios_test.go`. Red before the fix, green after:

- `TestContinuationAuditAcceptsEndorserBlockProducer` — the regression. A
  cert-RB with an empty body certifies an endorser block containing tx T; a
  later ordinary ranking block spends `T#0`; the ledger has not applied the
  endorser block (the provider cache holds it, the ledger does not). Asserts no
  report, that the window is not `endorserProducersPending`, and that T is in
  the producer set. **Failed before the fix.**
- `TestContinuationAuditTreatsPendingEndorserBlockAsInconclusive` — same shape
  with the endorser block not yet cached: no `Error`, and the inconclusive
  `Debug` record is emitted. **Failed before the fix.**
- `TestContinuationAuditStillReportsMissingProducerOnLeios` — the guard: with
  the closure resolved, a genuinely absent producer is still reported at
  `Error`. Passed before and after, so the fix is not muting the diagnostic.
- `TestContinuationAuditAcceptsAnnouncedEndorserBlockProducer` — the forward/CIP
  path, where the block applies the endorser block it announces itself.
- `TestEndorserBlockTxIdsMatchDecodedTransactionHashes` — cheap id derivation
  vs. the decoded transaction hash, for both bare and CBOR-in-CBOR entries.
- `TestContinuationAuditOutcomesAreCounted` — clean / inconclusive /
  missing_producer land in exactly one bucket each.
- `TestContinuationAuditCapDisarmIsExplicit` — disarm logs at `Warn` and counts
  `disarmed_cap`.
- `TestContinuationAuditCapCountsOnlyNewProducers` — a repeated endorser
  transaction offered to a window one short of the cap must not disarm it.
- `TestContinuationAuditResolvesEachEndorserBlockOnce` — three ranking blocks
  certifying one closure cost one resolution and one provider call.
- `TestContinuationAuditSkipsEndorserResolutionWithoutEndorserSpends` — a window
  that spends nothing endorser-resident performs zero resolutions; the reference
  stays queued, classified but unresolved.
- `TestContinuationAuditEndorserResolutionIsBudgeted` — one body may not exceed
  `continuationAuditMaxEndorserBlocksPerBlock`; the overflow requeues, the window
  goes inconclusive, and the stop is counted.
- `TestContinuationAuditRetriesUnresolvedCertifyingParent` — parent missing, the
  reference stays queued and is not duplicated by a second certifying block over
  the same parent; parent lands, the next body resolves it, the producer is
  recognised and the window is complete again.
- `TestContinuationAuditRetriesParentAtMostOncePerBlock` — three unresolvable
  inputs on one body cost one parent lookup.
- `TestContinuationAuditDedupesConvergingEndorserRefs` — two certifying blocks
  over two distinct parents at one slot announcing one endorser block leave
  exactly one queued reference and cost one provider probe per later body.

Fifteen mutations were checked independently and each turns the suite red: not
queueing the endorser reference, dropping the per-window memoization, making the
drain eager, removing the per-block budget, restoring the batch cap arithmetic
that counted duplicates, dropping the post-drain re-check, hashing the full
transaction CBOR instead of the body, perturbing the endorser-block slot binding
in the shared selector, dropping the transient requeue, latching incompleteness,
removing the once-per-body probe guard, misclassifying a not-found parent as a
hard error, requeueing without restoring the dedupe entry, removing the
converging-key dedupe, and routing the once-per-body path back through the
requeue helper. A sixteenth — dropping an already-merged guard in that helper —
stayed green, which identified the guard as unreachable; it was removed rather
than shipped as an untestable branch.

`gofmt`, `go vet ./...`, `golangci-lint run ./ledger/...` (0 issues),
`go test ./ledger/ -count=1` and `-race`, plus `./ledger/eras/` and
`./database/...` — all green on `upstream/main`.

## Follow-ups (not in this PR)

- The audit still inspects only the ranking block's own transaction inputs. A
  certified endorser block's transactions are now recognised as *producers* but
  their own inputs are not audited, so a splice that lands entirely inside an
  endorser block is not covered. Doing that needs the full decode this PR
  deliberately avoids on the mutex.
- #3170's second comment showed the same audit message on preview (Conway, no
  endorser blocks), where this mechanism cannot apply. That case is untouched
  here and may be a genuine, separate read-ahead race: an audit at blockfetch
  time against a producer whose block is below the fork point but not yet
  applied. `GetTransactionByHash` should cover it, but that is unverified under
  a deep backlog. This PR narrows the dataset — the Leios occurrences were
  noise and should be excluded from it.
- On a busy Leios chain the window will now disarm at the cap partway through
  its 512-block budget. That is a deliberate memory/coverage trade, and it is
  counted; if the coverage matters more than the ~16 MB, the cap is the knob.
- The per-block endorser-resolution budget is a count, not a time bound. A
  count is the right shape for the existing knobs in this file, but if endorser
  blocks grow much larger a deadline would bound the mutex hold better.

Related: #3170, #3005

### DIFF
--- a/ledger/continuation_audit.go
+++ b/ledger/continuation_audit.go
@@ -552,6 +552,12 @@ func (ls *LedgerState) queueContinuationAuditEndorserRef(
 // form — a later attempt costs only the provider lookup — and marks the window
 // incomplete, so unresolved inputs read as inconclusive rather than as missing
 // producers.
+//
+// *budget is owned by the audited body, not by one drain: a body drains once
+// per unresolved input and every drain spends the same allowance, which is why
+// it is passed by pointer. Once exhausted it is burned to -1 and stays there
+// for the rest of the body, so skipped_budget counts audited bodies whose
+// budget ran out — what its Help string promises — and not unresolved inputs.
 func (ls *LedgerState) drainContinuationAuditEndorserRefs(
 	window *continuationAuditWindow,
 	budget *int,
@@ -562,15 +568,19 @@ func (ls *LedgerState) drainContinuationAuditEndorserRefs(
 	armed := true
 	for i, ref := range pending {
 		if !armed || *budget <= 0 {
-			// Requeue what was not reached. Hitting the budget means the
-			// producer set is knowingly short for now. These references
+			// Requeue everything not reached, in one move, and stop.
+			// The remainder is a contiguous suffix, so there is nothing
+			// left for a further iteration to do: continuing here once
+			// appended an overlapping suffix per remaining reference,
+			// which grew the queue quadratically and made later bodies
+			// probe the same reference several times. These references
 			// were never taken off pendingEndorserSeen, so this is a
 			// one-for-one move and cannot duplicate a key.
 			window.pendingEndorserRefs = append(
 				window.pendingEndorserRefs,
 				pending[i:]...,
 			)
-			if armed && *budget <= 0 {
+			if armed && *budget == 0 {
 				ls.countContinuationAuditOutcome(
 					continuationAuditResultSkippedBudget,
 				)
@@ -581,10 +591,14 @@ func (ls *LedgerState) drainContinuationAuditEndorserRefs(
 					"deferred", len(pending)-i,
 					"budget", continuationAuditMaxEndorserBlocksPerBlock,
 				)
-				// Only report the budget stop once per drain.
+				// The budget belongs to the audited body, not to this
+				// drain: one body triggers one drain per unresolved
+				// input, all sharing it. Burn it to -1 so the later
+				// drains of the same body take this branch without
+				// reporting a second stop for the one exhaustion.
 				*budget = -1
 			}
-			continue
+			break
 		}
 		// One probe per reference per audited body: nothing that could
 		// change the outcome happens between two inputs of the same body.
@@ -687,9 +701,6 @@ func (ls *LedgerState) drainContinuationAuditEndorserRefs(
 		if !ls.recordContinuationAuditProducers(window, ids) {
 			armed = false
 		}
-	}
-	if *budget < 0 {
-		*budget = 0
 	}
 	return armed
 }
--- a/ledger/continuation_audit.go
+++ b/ledger/continuation_audit.go
@@ -17,6 +17,7 @@ package ledger
 import (
 	"encoding/hex"
 	"errors"
+	"strconv"
 
 	"github.com/blinklabs-io/dingo/database"
 	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
@@ -88,6 +89,14 @@ const (
 	// so truncating a pathological block is preferable to delaying the
 	// blockfetch pipeline behind unbounded per-input probes.
 	continuationAuditMaxInputsPerBlock = 32
+	// continuationAuditMaxEndorserBlocksPerBlock bounds how many endorser
+	// blocks one audited body may cause to be resolved, in the same spirit as
+	// continuationAuditMaxInputsPerBlock: resolving one costs a parent-block
+	// read plus a hash of every endorser transaction, all under
+	// chainsyncBlockfetchMutex. Refs left over stay queued for a later body,
+	// and the window reports inconclusive in the meantime rather than
+	// reporting from a set it knows is short.
+	continuationAuditMaxEndorserBlocksPerBlock = 32
 )
 
 // Label values of the dingo_ledger_continuation_audit_outcomes_total metric.
@@ -96,6 +105,7 @@ const (
 	continuationAuditResultMissingProducer       = "missing_producer"
 	continuationAuditResultInconclusiveEbPending = "inconclusive_eb_pending"
 	continuationAuditResultDisarmedCap           = "disarmed_cap"
+	continuationAuditResultSkippedBudget         = "skipped_budget"
 )
 
 // continuationAuditWindow is the state of one armed audit run. It is published
@@ -116,6 +126,51 @@ type continuationAuditWindow struct {
 	// sticky for the life of the window: once a hole exists, every later
 	// input could be falling into it.
 	endorserProducersPending bool
+	// pendingEndorserRefs are endorser-block references seen in this window
+	// whose transactions have not been merged into producedTxs yet, in
+	// arrival order. Classifying a block into a reference is header-only and
+	// free; turning a reference into producers is not, so it is deferred
+	// until an input actually fails to resolve without it. Deduplicated on
+	// insert through pendingEndorserSeen, so the same closure referenced by
+	// several ranking blocks is queued once.
+	pendingEndorserRefs []continuationAuditEndorserRef
+	pendingEndorserSeen map[string]struct{}
+	// resolvedEndorserBlocks memoizes the (hash, slot) occurrences whose
+	// transaction ids are already in producedTxs, so no endorser block is
+	// hashed twice in a window.
+	resolvedEndorserBlocks map[string]struct{}
+	// endorserResolutions counts references this window actually did work
+	// for: a parent-block read and/or a provider lookup. It is the cost the
+	// audit adds to the blockfetch path, and it stays zero for a window in
+	// which nothing spends an endorser-resident output.
+	endorserResolutions int
+}
+
+// continuationAuditEndorserRef is one endorser-block reference an audited
+// ranking block carries, in the cheapest form that can be resolved later.
+//
+// A CIP-path block announces its own endorser block, which the header gives up
+// directly. A cert-driven block certifies the endorser block its parent
+// announced, which needs a parent-block read; only the parent hash is retained
+// so the referencing block itself (and its CBOR) does not have to be. Once that
+// read happens the ref is rewritten in resolved form, so a still-unfetched
+// endorser block costs at most a provider lookup on any later attempt.
+type continuationAuditEndorserRef struct {
+	certParentHash []byte
+	ebHash         lcommon.Blake2b256
+	ebSlot         uint64
+	resolved       bool
+	blockSlot      uint64
+}
+
+// key identifies a reference for dedupe: the parent hash while the reference is
+// still a deferred certified closure, the (hash, slot) occurrence once
+// resolved.
+func (r continuationAuditEndorserRef) key() string {
+	if !r.resolved {
+		return "c" + string(r.certParentHash)
+	}
+	return "d" + string(r.ebHash.Bytes()) + strconv.FormatUint(r.ebSlot, 10)
 }
 
 // armContinuationAudit starts a bounded continuation audit at a rollback point.
@@ -132,7 +187,9 @@ func (ls *LedgerState) armContinuationAudit(
 		}
 	}
 	ls.continuationAudit.Store(&continuationAuditWindow{
-		producedTxs: make(map[string]struct{}),
+		producedTxs:            make(map[string]struct{}),
+		pendingEndorserSeen:    make(map[string]struct{}),
+		resolvedEndorserBlocks: make(map[string]struct{}),
 		forkPoint: ocommon.Point{
 			Slot: point.Slot,
 			Hash: append([]byte(nil), point.Hash...),
@@ -182,30 +239,18 @@ func (ls *LedgerState) auditContinuationBlock(
 	// it has one, contributes producers too: on the cert-driven path those are
 	// the only transactions a certifying ranking block brings, its own body
 	// being empty.
-	ebTxIds := ls.continuationAuditEndorserProducers(window, e)
-	if len(window.producedTxs)+len(txs)+len(ebTxIds) >
-		continuationAuditMaxProducedTxs {
-		ls.config.Logger.Warn(
-			"disarming cross-fork continuation audit: producer set at capacity",
-			"component", "ledger",
-			"blocks_audited", window.blocksSeen,
-			"produced_txs", len(window.producedTxs),
-			"max_produced_txs", continuationAuditMaxProducedTxs,
-		)
-		ls.countContinuationAuditOutcome(
-			continuationAuditResultDisarmedCap,
-		)
+	ls.queueContinuationAuditEndorserRef(window, e)
+	producers := make([][]byte, 0, len(txs))
+	for _, tx := range txs {
+		producers = append(producers, tx.Hash().Bytes())
+	}
+	if !ls.recordContinuationAuditProducers(window, producers) {
 		window.remaining = 0
 		return
 	}
-	for _, tx := range txs {
-		window.producedTxs[string(tx.Hash().Bytes())] = struct{}{}
-	}
-	for _, id := range ebTxIds {
-		window.producedTxs[string(id)] = struct{}{}
-	}
 	reports := 0
 	inputsAudited := 0
+	endorserBudget := continuationAuditMaxEndorserBlocksPerBlock
 	for _, tx := range txs {
 		for _, input := range collectReferencedInputs(tx) {
 			if inputsAudited >= continuationAuditMaxInputsPerBlock {
@@ -225,6 +270,21 @@ func (ls *LedgerState) auditContinuationBlock(
 					"input", input.String(),
 				)
 				continue
+			}
+			// Only now, with the cheap paths exhausted, is it worth
+			// paying for the window's endorser blocks: an audited body
+			// that spends nothing endorser-resident never triggers a
+			// parent-block read or a transaction hash.
+			if !resolved && len(window.pendingEndorserRefs) > 0 {
+				if !ls.drainContinuationAuditEndorserRefs(
+					window,
+					&endorserBudget,
+					e.Point.Slot,
+				) {
+					window.remaining = 0
+					return
+				}
+				_, resolved = window.producedTxs[string(input.Id().Bytes())]
 			}
 			if resolved {
 				ls.countContinuationAuditOutcome(
@@ -317,71 +377,232 @@ func (ls *LedgerState) continuationInputHasProducer(
 	return producerTx != nil, nil
 }
 
-// continuationAuditEndorserProducers returns the transaction ids of the Leios
-// endorser block the audited ranking block applies, resolving it exactly the
-// way ledgerProcessBlock does: leiosEndorserBlockForApply picks the reference
-// (the block's own announcement on the CIP path, the parent's announcement on
-// the Musashi cert-driven path) and EndorserBlockProvider supplies the
-// transactions for that (hash, slot) occurrence.
+// recordContinuationAuditProducers adds producer transaction ids to the
+// window's set, and reports whether the set stayed inside
+// continuationAuditMaxProducedTxs.
 //
-// The endorser block is already in the provider's cache by the time the
-// referencing ranking block is fetched — leios-fetch runs well ahead of ledger
-// apply — so the common path is a cache lookup plus a hash per endorser
-// transaction, with no additional I/O beyond the parent-block lookup the
-// cert-driven path needs. When the reference cannot be resolved or the block is
-// not cached yet, the window is marked incomplete rather than left to report a
-// producer it was never given.
+// Only ids the set does not already hold count toward the cap. The cap bounds
+// the size of the producer set, not the number of ids offered to it, and
+// charging a repeat against it would disarm a window whose set never grew.
+// Repeats are ordinary on the Leios path: the same transaction can appear in
+// more than one endorser block — applyEndorserBlock carries
+// deduplicateEndorserBlockTransactionIndexes for exactly that — and an endorser
+// block is content-addressed, so the same closure can be referenced from more
+// than one ranking block inside a window.
+//
+// The cap is enforced per insert rather than per block, so the set is never
+// larger than the cap. A block that runs the set into the cap partway through
+// leaves those ids recorded, which is harmless: the window is disarmed and its
+// set is never consulted again.
+func (ls *LedgerState) recordContinuationAuditProducers(
+	window *continuationAuditWindow,
+	ids [][]byte,
+) bool {
+	for _, id := range ids {
+		key := string(id)
+		if _, ok := window.producedTxs[key]; ok {
+			continue
+		}
+		if len(window.producedTxs) >= continuationAuditMaxProducedTxs {
+			ls.config.Logger.Warn(
+				"disarming cross-fork continuation audit: producer set at capacity",
+				"component", "ledger",
+				"blocks_audited", window.blocksSeen,
+				"produced_txs", len(window.producedTxs),
+				"max_produced_txs", continuationAuditMaxProducedTxs,
+			)
+			ls.countContinuationAuditOutcome(
+				continuationAuditResultDisarmedCap,
+			)
+			return false
+		}
+		window.producedTxs[key] = struct{}{}
+	}
+	return true
+}
+
+// queueContinuationAuditEndorserRef records, in header-only work, the endorser
+// block an audited ranking block applies, so it can be turned into producers
+// later if any input needs it.
+//
+// The reference is selected exactly as the apply path selects it: the block's
+// own announcement when LeiosApplyEndorserBlockTxs is set (the CIP path,
+// bound to the block's own slot), the parent's announcement otherwise (the
+// Musashi cert-driven path). The cert-driven case retains only the parent hash;
+// leiosCertifiedAnnouncementFromParent — the same helper
+// leiosEndorserBlockForApply uses — turns it into a reference when the time
+// comes, so the two cannot select different endorser blocks.
 //
 // Non-Leios chains are unaffected: no endorser-block provider is configured,
-// and a header that neither announces nor certifies an endorser block returns
-// no reference.
-func (ls *LedgerState) continuationAuditEndorserProducers(
+// and a header that neither announces nor certifies an endorser block queues
+// nothing.
+func (ls *LedgerState) queueContinuationAuditEndorserRef(
 	window *continuationAuditWindow,
 	e BlockfetchEvent,
-) [][]byte {
-	if ls.config.EndorserBlockProvider == nil {
-		return nil
+) {
+	if ls.config.EndorserBlockProvider == nil || e.Block == nil {
+		return
 	}
-	ebHash, ebSlot, _, referenced, err := ls.leiosEndorserBlockForApply(e.Block)
-	if err != nil {
-		window.endorserProducersPending = true
-		ls.config.Logger.Debug(
-			"cross-fork continuation audit could not resolve a block's endorser block",
-			"component", "ledger",
-			"slot", e.Point.Slot,
-			"error", err,
+	var ref continuationAuditEndorserRef
+	if ls.config.LeiosApplyEndorserBlockTxs {
+		referencer, ok := e.Block.Header().(leiosEndorserBlockReferencer)
+		if !ok {
+			return
+		}
+		ebHash, _, announced := referencer.LeiosAnnouncement()
+		if !announced {
+			return
+		}
+		ref = continuationAuditEndorserRef{
+			ebHash:    ebHash,
+			ebSlot:    e.Block.SlotNumber(),
+			resolved:  true,
+			blockSlot: e.Point.Slot,
+		}
+	} else {
+		certifier, ok := e.Block.Header().(leiosEndorserBlockCertifier)
+		if !ok {
+			return
+		}
+		certified, present := certifier.LeiosCertified()
+		if !present || !certified {
+			return
+		}
+		ref = continuationAuditEndorserRef{
+			certParentHash: e.Block.PrevHash().Bytes(),
+			blockSlot:      e.Point.Slot,
+		}
+	}
+	key := ref.key()
+	if _, ok := window.pendingEndorserSeen[key]; ok {
+		return
+	}
+	if _, ok := window.resolvedEndorserBlocks[key]; ok {
+		return
+	}
+	window.pendingEndorserSeen[key] = struct{}{}
+	window.pendingEndorserRefs = append(window.pendingEndorserRefs, ref)
+}
+
+// drainContinuationAuditEndorserRefs merges queued endorser blocks into the
+// window's producer set, spending at most *budget resolutions, and reports
+// whether the window is still armed.
+//
+// Each endorser block is resolved and hashed at most once per window: the
+// queue is deduplicated on insert and the (hash, slot) occurrences already
+// merged are memoized, so the same closure referenced by several ranking
+// blocks costs one parent read, one provider lookup and one hashing pass, not
+// one per referencing block.
+//
+// A reference whose endorser block is not cached yet stays queued in resolved
+// form — a later attempt costs only the provider lookup — and marks the window
+// incomplete, so unresolved inputs read as inconclusive rather than as missing
+// producers.
+func (ls *LedgerState) drainContinuationAuditEndorserRefs(
+	window *continuationAuditWindow,
+	budget *int,
+	auditedSlot uint64,
+) bool {
+	pending := window.pendingEndorserRefs
+	window.pendingEndorserRefs = nil
+	armed := true
+	for i, ref := range pending {
+		if !armed || *budget <= 0 {
+			// Requeue what was not reached. Hitting the budget means the
+			// producer set is knowingly short for now.
+			window.pendingEndorserRefs = append(
+				window.pendingEndorserRefs,
+				pending[i:]...,
+			)
+			if armed && *budget <= 0 {
+				window.endorserProducersPending = true
+				ls.countContinuationAuditOutcome(
+					continuationAuditResultSkippedBudget,
+				)
+				ls.config.Logger.Debug(
+					"cross-fork continuation audit deferred endorser blocks past its per-block budget",
+					"component", "ledger",
+					"slot", auditedSlot,
+					"deferred", len(pending)-i,
+					"budget", continuationAuditMaxEndorserBlocksPerBlock,
+				)
+				// Only report the budget stop once per drain.
+				*budget = -1
+			}
+			continue
+		}
+		delete(window.pendingEndorserSeen, ref.key())
+		*budget--
+		window.endorserResolutions++
+		if !ref.resolved {
+			ebHash, ebSlot, _, announced, err := ls.leiosCertifiedAnnouncementFromParent(
+				ref.certParentHash,
+			)
+			if err != nil {
+				window.endorserProducersPending = true
+				ls.config.Logger.Debug(
+					"cross-fork continuation audit could not resolve a certifying block's endorser block",
+					"component", "ledger",
+					"slot", ref.blockSlot,
+					"error", err,
+				)
+				continue
+			}
+			if !announced {
+				continue
+			}
+			ref.ebHash = ebHash
+			ref.ebSlot = ebSlot
+			ref.resolved = true
+		}
+		key := ref.key()
+		if _, ok := window.resolvedEndorserBlocks[key]; ok {
+			continue
+		}
+		rawTxs, ok := ls.config.EndorserBlockProvider(
+			ref.ebHash.Bytes(),
+			ref.ebSlot,
 		)
-		return nil
+		if !ok {
+			window.endorserProducersPending = true
+			ls.config.Logger.Debug(
+				"cross-fork continuation audit: certified endorser block not fetched yet",
+				"component", "ledger",
+				"slot", ref.blockSlot,
+				"eb_slot", ref.ebSlot,
+				"eb_hash", ref.ebHash.String(),
+			)
+			// Keep it queued in resolved form: the parent read is done,
+			// and the block may be fetched before the window ends.
+			window.pendingEndorserSeen[key] = struct{}{}
+			window.pendingEndorserRefs = append(
+				window.pendingEndorserRefs,
+				ref,
+			)
+			continue
+		}
+		ids, err := endorserBlockTxIds(rawTxs)
+		if err != nil {
+			window.endorserProducersPending = true
+			ls.config.Logger.Debug(
+				"cross-fork continuation audit could not read endorser block transaction ids",
+				"component", "ledger",
+				"slot", ref.blockSlot,
+				"eb_slot", ref.ebSlot,
+				"eb_hash", ref.ebHash.String(),
+				"error", err,
+			)
+			continue
+		}
+		window.resolvedEndorserBlocks[key] = struct{}{}
+		if !ls.recordContinuationAuditProducers(window, ids) {
+			armed = false
+		}
 	}
-	if !referenced {
-		return nil
+	if *budget < 0 {
+		*budget = 0
 	}
-	rawTxs, ok := ls.config.EndorserBlockProvider(ebHash.Bytes(), ebSlot)
-	if !ok {
-		window.endorserProducersPending = true
-		ls.config.Logger.Debug(
-			"cross-fork continuation audit: certified endorser block not fetched yet",
-			"component", "ledger",
-			"slot", e.Point.Slot,
-			"eb_slot", ebSlot,
-			"eb_hash", ebHash.String(),
-		)
-		return nil
-	}
-	ids, err := endorserBlockTxIds(rawTxs)
-	if err != nil {
-		window.endorserProducersPending = true
-		ls.config.Logger.Debug(
-			"cross-fork continuation audit could not read endorser block transaction ids",
-			"component", "ledger",
-			"slot", e.Point.Slot,
-			"eb_slot", ebSlot,
-			"eb_hash", ebHash.String(),
-			"error", err,
-		)
-		return nil
-	}
-	return ids
+	return armed
 }
 
 // countContinuationAuditOutcome increments the pre-materialized outcome
@@ -398,6 +619,8 @@ func (ls *LedgerState) countContinuationAuditOutcome(result string) {
 		counter = ls.metrics.continuationAuditInconclusiveEbPending
 	case continuationAuditResultDisarmedCap:
 		counter = ls.metrics.continuationAuditDisarmedCap
+	case continuationAuditResultSkippedBudget:
+		counter = ls.metrics.continuationAuditSkippedBudget
 	}
 	if counter == nil {
 		return
--- a/ledger/continuation_audit.go
+++ b/ledger/continuation_audit.go
@@ -197,13 +197,33 @@ func (w *continuationAuditWindow) endorserProducersIncomplete() bool {
 	return len(w.pendingEndorserRefs) > 0 || w.endorserRefsDropped > 0
 }
 
-// requeueEndorserRef returns a reference to the pending queue, restoring the
-// dedupe entry the drain removed when it took it, so a later audited body
-// retries it and a concurrent duplicate is still suppressed.
+// requeueEndorserRef returns a reference the drain took and could not finish to
+// the pending queue, restoring the dedupe entry that was removed with it.
+//
+// It dedupes, because resolution can make two references converge. A deferred
+// certified reference is keyed by its parent hash, and two ranking blocks with
+// different parents can certify the same endorser block: the reference is
+// (announced hash, announcing slot), so two blocks at one slot on either side
+// of a fork — the state a window is armed in — name the same occurrence. Those
+// keys differ while the references are deferred and become equal once resolved.
+// Appending unconditionally then queued the same still-unfetched endorser block
+// twice, and every later body probed it twice, spending budget other references
+// needed.
+//
+// Only callers that took a reference off the queue may use it. The drain's
+// once-per-body and budget paths never take the dedupe entry, so they move
+// their references back directly instead; routing them through here would see
+// their own entry and drop them. An already-merged occurrence cannot reach
+// here either: the drain consults resolvedEndorserBlocks before the provider,
+// which is the only step after which a merged reference could be requeued.
 func (w *continuationAuditWindow) requeueEndorserRef(
 	ref continuationAuditEndorserRef,
 ) {
-	w.pendingEndorserSeen[ref.key()] = struct{}{}
+	key := ref.key()
+	if _, ok := w.pendingEndorserSeen[key]; ok {
+		return
+	}
+	w.pendingEndorserSeen[key] = struct{}{}
 	w.pendingEndorserRefs = append(w.pendingEndorserRefs, ref)
 }
 
@@ -543,7 +563,9 @@ func (ls *LedgerState) drainContinuationAuditEndorserRefs(
 	for i, ref := range pending {
 		if !armed || *budget <= 0 {
 			// Requeue what was not reached. Hitting the budget means the
-			// producer set is knowingly short for now.
+			// producer set is knowingly short for now. These references
+			// were never taken off pendingEndorserSeen, so this is a
+			// one-for-one move and cannot duplicate a key.
 			window.pendingEndorserRefs = append(
 				window.pendingEndorserRefs,
 				pending[i:]...,
@@ -566,8 +588,13 @@ func (ls *LedgerState) drainContinuationAuditEndorserRefs(
 		}
 		// One probe per reference per audited body: nothing that could
 		// change the outcome happens between two inputs of the same body.
+		// Its dedupe entry was never taken, so this moves the reference
+		// back one-for-one rather than going through requeueEndorserRef.
 		if ref.lastProbedBlock == window.blocksSeen {
-			window.requeueEndorserRef(ref)
+			window.pendingEndorserRefs = append(
+				window.pendingEndorserRefs,
+				ref,
+			)
 			continue
 		}
 		delete(window.pendingEndorserSeen, ref.key())
--- a/ledger/continuation_audit.go
+++ b/ledger/continuation_audit.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 
 	"github.com/blinklabs-io/dingo/database"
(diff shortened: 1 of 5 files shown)


chadaibbhot-git:
Both reviews landed on the same producer-set cap defect, and cubic's P3 on the cost of the resolution is fair. Fixed in 518654d6.

**Duplicates counted against the cap (CodeRabbit; cubic P2).** Correct, and worse than a corner case on this path: the same transaction can appear in more than one endorser block — `applyEndorserBlock` carries `deduplicateEndorserBlockTransactionIndexes` for exactly that — and an endorser block is content-addressed, so the same closure can be certified from more than one ranking block inside a window. The check compared `len(producedTxs) + len(offered)` against the cap, so a repeat consumed a slot the set never used and could disarm a window whose producer set had not grown. The cap now bounds what it says it bounds: it is enforced per insert, and only ids the set does not already hold count toward it. `TestContinuationAuditCapCountsOnlyNewProducers` seeds the window two short of the cap with one of the endorser block's two transactions already recorded, then forces a drain; restoring the batch arithmetic turns it red.

**Cost model (cubic P3).** Also fair — the resolution was eager and per-referencing-block, on the mutex, and a dense endorser-block backlog is exactly when the audit is armed. Three changes:

- *Lazy.* Classifying a block into an endorser-block reference is header-only and free, so that stays eager. Turning a reference into producers — a parent-block read plus a hash of every endorser transaction — is deferred until an audited input fails to resolve from the ranking-block producer set *and* the ledger. A window in which nothing spends an endorser-resident output now does zero parent reads, zero provider lookups and zero hashing. The cert-driven reference is retained as its parent hash alone, so the referencing block and its CBOR are not held; `leiosCertifiedAnnouncementFromParent` is split out of `leiosEndorserBlockForApply` and called by both, so the audit and apply cannot select different endorser blocks.
- *Once per window.* References are deduplicated when queued and merged occurrences memoized by `(ebHash, ebSlot)`, so a closure certified by N ranking blocks costs one parent read, one provider lookup and one hashing pass rather than N. An endorser block that is not cached yet is requeued in *resolved* form, so a later attempt costs only the provider lookup, not the parent read again.
- *Budgeted.* New `continuationAuditMaxEndorserBlocksPerBlock` (32), the analogue of `continuationAuditMaxInputsPerBlock`. Overflow stays queued for a later body; while the set is knowingly short the window reports inconclusive rather than missing, and the stop is counted as `result="skipped_budget"` on the outcome metric.

Net per audited body on a Leios node: nothing at all unless one of its inputs is unresolvable without an endorser block, and at most 32 endorser-block resolutions when it is, each of them the first and only time that closure is touched in the window.

New tests: `TestContinuationAuditResolvesEachEndorserBlockOnce` (three ranking blocks certifying one closure → one resolution, one provider call), `TestContinuationAuditSkipsEndorserResolutionWithoutEndorserSpends` (no endorser spends → zero resolutions, reference stays queued and unresolved), `TestContinuationAuditEndorserResolutionIsBudgeted`. Eight mutations were checked independently and each turns the suite red: not queueing the reference, dropping the memoization, making the drain eager, removing the per-block budget, restoring the batch cap arithmetic, dropping the post-drain re-check, hashing the full transaction CBOR instead of the body, and perturbing the endorser-block slot binding in the shared selector.

`gofmt`, `go vet ./...`, `golangci-lint run ./ledger/...` (0 issues), `go test ./ledger/ -count=1` and `-race`, plus `./ledger/eras/` — all green.

chadaibbhot-git:
Confirmed and fixed in 64ef7593. A certifying ranking block names its certified closure only through its parent's announcement, so a parent that is not in the block store yet is exactly the read-ahead this audit exists to tolerate — and the drain was treating it as a verdict. Worse than the report suggests, in fact: the reference was dropped for the *whole* window, since nothing re-queues it once its certifying block has been audited, so a single transient miss silenced that window permanently.

- **Retry.** `models.ErrBlockNotFound` now requeues the reference and restores the dedupe entry the drain took with it, so the next audited body retries. Anything else — I/O, a corrupt hash index — will not fix itself by retrying, so it drops the reference, counts `result="ref_unresolvable"` on the outcome metric and logs at Warn once per window rather than once per body.
- **Incompleteness is derived, not latched.** "The producer set is knowingly short" was a sticky bool; it is now read off the queue (`len(pendingEndorserRefs) > 0 || endorserRefsDropped > 0`). A window whose references all resolve is complete again and can go back to diagnosing a genuine missing producer — which is the second half of your finding, and was not fixed by the requeue alone.
- **Bounded.** Each reference is probed at most once per audited body (`lastProbedBlock`), since nothing that could change the outcome happens between two inputs of the same body. A permanently absent parent therefore costs one index miss per audited block, inside the existing per-block endorser budget, never more than the block budget for the window.

One behavioural note worth flagging, because it changed an existing test: references are no longer lost on a failed probe, so after a budget-limited body the queue *retains* everything it could not resolve rather than shrinking. `TestContinuationAuditEndorserResolutionIsBudgeted` now asserts that, plus `ref_unresolvable == 0` — an absent parent is retryable, not unresolvable.

Tests: `TestContinuationAuditRetriesUnresolvedCertifyingParent` (parent missing → probe fails, reference stays queued, a second certifying block over the same parent does not duplicate it, parent lands → next body resolves it, producer recognised, window complete again) and `TestContinuationAuditRetriesParentAtMostOncePerBlock` (three unresolvable inputs on one body → one lookup). Five mutations checked independently, each turning the suite red: dropping the requeue, latching incompleteness, removing the once-per-body probe guard, misclassifying a not-found parent as a hard error, and requeueing without restoring the dedupe entry.

`gofmt`, `go vet ./...`, `golangci-lint run ./ledger/...` (0 issues), `go test ./ledger/ -count=1` and `-race`, plus `./ledger/eras/` — all green.

chadaibbhot-git:
Two of the three are real and fixed in 30b5487a. The first I could not reproduce; details below so you can check my reasoning rather than take my word.

**1. `continuation_audit.go:438` — sticky pending on transient parent failure.** I believe this one is stale, but I re-derived it from the drain rather than assuming. At the revision you reviewed (64ef7593) line 438 is inside `queueContinuationAuditEndorserRef`, and on current head it is inside `recordContinuationAuditProducers` — neither is a parent-resolution branch. Enumerating every exit in `drainContinuationAuditEndorserRefs`, a reference is either **retained** (requeued or moved back, so `endorserProducersIncomplete()` clears as soon as it resolves) or **deliberately dropped**:

| exit | disposition |
|---|---|
| budget exhausted / window disarmed | retained (`pending[i:]` moved back) |
| already probed by this body | retained (moved back) |
| parent lookup `ErrBlockNotFound` | retained (requeued, retried next body) |
| parent lookup other error | dropped, `endorserRefsDropped++`, `ref_unresolvable` |
| parent resolves, announces nothing | dropped, and *not* a hole — definitively no endorser block |
| occurrence already merged | dropped, nothing left to do |
| provider miss | retained (requeued in resolved form) |
| `endorserBlockTxIds` decode failure | dropped, `endorserRefsDropped++`, `ref_unresolvable` |
| success | merged, memoized |

The only permanently-incomplete exits are the two hard-error ones. A decode failure re-reads the same cached bytes, so retrying it cannot change the answer; that is the intended distinction rather than an oversight. If you meant a specific branch not in that table, point me at it and I will fix it.

**2. `requeueEndorserRef` bypasses dedupe (P2).** Real, and the convergence mechanism is worth stating precisely: a deferred certified reference is keyed by its *parent hash*, but the endorser-block reference is (announced hash, announcing slot). Two ranking blocks with different parents at the same slot — either side of a fork, which is the state a window is armed in — therefore name the same occurrence under two different queue keys. They converge only once resolved, so the insert-time dedupe never sees the collision, and requeueing appended unconditionally. A still-unfetched endorser block ended up queued twice and probed twice on every later body. `requeueEndorserRef` now dedupes against the queue.

One consequence worth calling out: the drain's once-per-body and budget paths never *take* a reference's dedupe entry, so they move their references back directly rather than through the helper — routing them through it would make each one see its own entry and drop itself. That is now stated at the helper.

Mutation testing also showed the `resolvedEndorserBlocks` guard I first added to the helper was unreachable — the drain consults that map before the provider, which is the only step after which a merged reference could be requeued — so it is removed rather than shipped as an untestable branch.

**3. `metrics.go` Help string (P3).** Fixed; it now lists `ref_unresolvable counts abandoned endorser-block references` alongside the other four results.

Test: `TestContinuationAuditDedupesConvergingEndorserRefs` — two certifying blocks over two distinct parents at one slot announcing one endorser block, with the endorser block unfetched, must leave exactly one queued reference and cost one provider probe per later body. Red before the fix (two queued). Three mutations checked independently: removing the dedupe (red), routing the once-per-body path back through the helper (red), and dropping the already-merged guard (green — which is what identified it as dead code).

`gofmt`, `go vet ./...`, `golangci-lint run ./ledger/...` (0 issues), `go test ./ledger/ -count=1` and `-race`, plus `./ledger/eras/` — all green.

chadaibbhot-git:
Flagging a CI result on this branch so it is not read as mine: the `go-test (Linux, race)` job failed on 30b5487a, in `TestWindowedRewindConvergesWhilePrimaryChainExtends`:

```
recovery_rewind_window_test.go:196:
  Received unexpected error: rollback primary chain to intermediate point at
  slot 460: remove block at index 54: block not found
```

It is not a data race and not in anything this PR touches (the change is confined to the continuation audit, the endorser-transaction envelope helper, and the outcome metric). I ran a control rather than reasoning about it:

| | `-count=1 -race`, 20 runs | `-count=30 -race` |
|---|---|---|
| clean `main` @ 6b6346e3 | 0/20 fail | **fails** |
| this branch @ 30b5487a | 0/20 fail | **fails** |

Identical on both, so the branch does not introduce it. `go test ./ledger/ -count=1 -race` at 30b5487a passes locally (396s), as do `go-test (Linux)`, `(Windows)` and `(macOS)` on this same CI run. For what it is worth, `go-test` also failed on `main` itself for two of the four most recent pushes (90504583 and 7ac66d79), so intermittent failures here predate this branch.

The `-count=30` reproduction suggests state leaking across repeated in-process runs in that test's fixture rather than a product bug, but I have not chased it further since it is outside this PR — happy to open something separate if that would be useful. I cannot re-run the job from here (no write access to the repo's Actions); a maintainer re-run would confirm.

ledger/continuation_audit.go
```diff
@@ -233,3 +430,291 @@ func (ls *LedgerState) continuationInputHasProducer(
...
+func (ls *LedgerState) recordContinuationAuditProducers(
+	window *continuationAuditWindow,
+	ids [][]byte,
+) bool {
+	for _, id := range ids {
+		key := string(id)
+		if _, ok := window.producedTxs[key]; ok {
+			continue
+		}
+		if len(window.producedTxs) >= continuationAuditMaxProducedTxs {
```
wolf31o2, line 460: Keep the continuation audit’s producer-set cap based on unique transaction IDs, including IDs obtained from Leios endorser blocks. Repeated IDs must not consume additional capacity, while genuinely new IDs must still disarm the window at the configured bound.

wolf31o2 (changes requested):
Code review: changes requested. The current head has a failing race check; resolve it and provide a clean required-check run before approval.

chadaibbhot-git:
@coderabbitai review (draft, head 30b5487a)

chadaibbhot-git:
Rebased onto current main (c159aa6d) to get a fresh required-check run; head is now 7511a3ea. The four commits replay unchanged — the rebase diff is byte-identical to the pre-rebase patch, none of main's four new commits touch the three files this PR modifies, and each commit builds on its own.

On the race check: that failure is the pre-existing `TestWindowedRewindConvergesWhilePrimaryChainExtends` flake, not something this PR introduces. I checked it against unmodified main rather than assuming:

| | `-count=1 -race`, 20 runs | `-count=30 -race` |
|---|---|---|
| clean main @ 6b6346e3 | 0/20 fail | fails |
| this branch @ 30b5487a | 0/20 fail | fails |

Identical on both, and the failure is an assertion in `recovery_rewind_window_test.go` (`remove block at index 54: block not found`), not a data race and not in anything this PR touches. `go test ./ledger/ -count=1` (258s) and `-race` (1230s) both pass locally on 7511a3ea.

The `-count=30` signature points at state leaking across repeated in-process runs in that test's fixture. That needs its own PR and I have deliberately not folded it into this one; I am happy to prepare it separately if you would like.

Re-verification on the rebased head: full `./ledger/` suite and `-race` green, `./ledger/eras/` green, `gofmt`, `go vet ./...`, `golangci-lint run ./ledger/...` 0 issues, and all sixteen mutation checks still red→green.

chadaibbhot-git:
Status on the rebased head 7511a3ea: eight of the nine required checks are green, including **`go-test (Linux, race)`**, which is the check that prompted the change request. That result also settles the earlier race failure as the flake it was reported to be — same branch content, clean race run.

The one red check is `go-test (Windows)`, and it is the CI runner running out of disk, not a test result:

```
bark/database_wire_test.go:99
  start plugin storage.blob/badger: ... while opening file:
  C:\...\Temp\TestDatabaseServiceOverRealHTTP...\blob\000001.vlog
  err: truncate ...: There is not enough space on the disk.

database/batch_skip_test.go:447
  set transaction metadata for tx e6fadc1d... : create output ...#48:
  database or disk is full (13)
```

Both are `no space left` from the Windows runner's temp volume — one from badger's value-log truncate, one from SQLite. Neither package (`bark`, `database`) is touched by this PR, which changes only `ledger/continuation_audit.go`, `ledger/leios_apply.go` and `ledger/metrics.go` and adds one test file; and `go-test (Windows)` passed on the immediately preceding head (30b5487a) with identical content in those packages.

Locally on 7511a3ea: `go test ./ledger/ -count=1` 258s green, `-race` 1230s green, `./ledger/eras/` green, `gofmt`, `go vet ./...`, `golangci-lint run ./ledger/...` 0 issues, and all sixteen mutation checks still red→green.

I cannot re-run the job from here — as an external contributor I have no Actions write access on this repo (`gh run rerun` returns "cannot be rerun"), and per our own policy I will not push a no-op commit to re-roll CI. A maintainer re-run of the Windows job should clear it. Flagging rather than acting.

chadaibbhot-git:
@coderabbitai review (final head 1ffba97b)

ledger/continuation_audit.go
```diff
@@ -233,3 +430,291 @@ func (ls *LedgerState) continuationInputHasProducer(
...
+					"cross-fork continuation audit deferred endorser blocks past its per-block budget",
+					"component", "ledger",
+					"slot", auditedSlot,
+					"deferred", len(pending)-i,
+					"budget", continuationAuditMaxEndorserBlocksPerBlock,
+				)
+				// Only report the budget stop once per drain.
+				*budget = -1
+			}
+			continue
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid, and the consequence is worse than a quadratic queue — thanks for catching it. Fixed in 432d5f4419bcdaaf608305ac086d0b9dbe303276.

`*budget` is owned by the audited *body*, not by one drain: a body drains once per unresolved input and every drain of that body spends the same allowance. Combined with the end-of-drain `if *budget < 0 { *budget = 0 }` reset (the P3 finding), the overlapping-suffix append did not just grow the queue once — it compounded across the drains of a single body. A four-input body holding 35 queued references went 35 -> 38 -> 741 -> ~275k -> ~3.8e10 pending references, all under `chainsyncBlockfetchMutex`.

What changed in `drainContinuationAuditEndorserRefs`:

* requeue the remainder once (`pending[i:]`) and `break`. The remainder is a contiguous suffix, so no later iteration has anything left to do.
* report the stop once, gated on `*budget == 0` (exhausted, not yet reported) instead of `*budget <= 0`.
* drop the end-of-drain reset, so `-1` means "this body's budget is spent and already reported" for the rest of the body.

Tests:

* `TestContinuationAuditEndorserResolutionIsBudgeted` now queues three references past the budget instead of one. That was the hole in the old test: with a single leftover the branch is entered exactly once, so it could not distinguish "requeue the tail once" from "requeue it once per remaining reference". On the pre-fix tree it fails with 38 queued references where 35 are expected, and `skipped_budget` 3 where 1 is expected.
* `TestContinuationAuditBudgetStopIsCountedOncePerBody` covers the across-drain compounding. On the pre-fix tree it does not reach an assertion at all: the queue growth compounds until the process exhausts memory. On the fix the whole continuation-audit suite runs in 2.6s.

Each half of the fix is mutation-checked independently: putting `continue` back in place of `break` reproduces the 38-versus-35 queue growth, and restoring the end-of-drain reset or widening the gate back to `*budget <= 0` makes a four-input body report `skipped_budget` 4 instead of 1.

ledger/continuation_audit.go
```diff
@@ -170,11 +325,53 @@ func (ls *LedgerState) auditContinuationBlock(
...
+					&endorserBudget,
+					e.Point.Slot,
+				) {
+					window.remaining = 0
+					return
+				}
+				_, resolved = window.producedTxs[string(input.Id().Bytes())]
+			}
 			if resolved {
+				ls.countContinuationAuditOutcome(
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid. Fixed in 432d5f4419bcdaaf608305ac086d0b9dbe303276, together with the P1 finding on the same drain — they are the same defect seen from two sides, so the fix is one change.

You are right about the mechanism: `endorserBudget` is declared once per audited body in `auditContinuationBlock` and shared by every drain that body triggers, one per unresolved input. The end-of-drain `if *budget < 0 { *budget = 0 }` erased the sentinel, so the next input's drain re-entered the budget-limited branch and counted a second stop for the same exhausted budget. A body with several unresolvable inputs reported one `skipped_budget` per input, which is not what the Help string ("`skipped_budget` counts audited blocks") or the PR description promises.

The reset is gone and the gate is now `*budget == 0` rather than `*budget <= 0`, so `-1` reads as "spent and already reported" for the remainder of the body. The drain's doc comment now states the ownership explicitly, since the pointer parameter is the only hint that the budget outlives the call.

Test: `TestContinuationAuditBudgetStopIsCountedOncePerBody` audits one body with four inputs that all need the drain, against a queue three references longer than the budget, and asserts `skipped_budget == 1` and that `window.endorserResolutions` stops at `continuationAuditMaxEndorserBlocksPerBlock` — the budget must not be replenished between the inputs of one body. Mutation-checked: putting the tail reset back, or widening the gate to `*budget <= 0`, makes it report 4.

ledger/continuation_audit.go
chadaibbhot-git: Agreed, and that is the behaviour on the current head — `recordContinuationAuditProducers` is the single place ids enter the window, and it charges the cap only for ids the set does not already hold:

```go
for _, id := range ids {
        key := string(id)
        if _, ok := window.producedTxs[key]; ok {
                continue
        }
        if len(window.producedTxs) >= continuationAuditMaxProducedTxs {
                // warn, count disarmed_cap, disarm
                return false
        }
        window.producedTxs[key] = struct{}{}
}
```

The earlier version compared `len(window.producedTxs) + len(txs) + len(ebTxIds)` against the bound, which charged repeats and could disarm a window whose producer set never grew. Both the body's own transaction ids and the ids obtained from Leios endorser blocks now go through this one function, so an endorser-block id repeated across two endorser blocks, or an endorser block referenced from two ranking blocks in the same window, consumes no capacity. The cap is also enforced per insert rather than per block, so the set can never exceed the bound.

Both halves of the requirement are pinned by tests:

* `TestContinuationAuditCapCountsOnlyNewProducers` — seeded two short of the cap with one of the endorser block's two transactions already recorded, it asserts the window stays armed, that the set lands exactly on the cap (only the two genuinely new ids were added), that no `disarming` warning was logged and `disarmed_cap` is 0, and it requires `window.endorserResolutions > 0` so the endorser block really was drained.
* `TestContinuationAuditCapDisarmIsExplicit` — seeded one short of the cap with genuinely new ids, it asserts the window disarms at the bound, logs the WARN, and increments `disarmed_cap`.

chadaibbhot-git:
Round summary for `432d5f4419bcdaaf608305ac086d0b9dbe303276`, and a note on the red go-test checks, which are not from this branch.

**What changed.** One commit, confined to `drainContinuationAuditEndorserRefs`, addressing both findings on the previous head — they turned out to be the same defect seen from two sides:

* the budget-limited branch requeued the unreached remainder and then *continued* the loop, appending an overlapping copy of a contiguous suffix on every further iteration;
* the end-of-drain `if *budget < 0 { *budget = 0 }` reset erased the exhaustion sentinel, so the next unresolved input of the same body re-entered that branch and reported a second `skipped_budget` for one exhausted budget.

Left together they compounded: the reset returned the budget to zero at each drain, so a body re-entered the quadratic append once per input. The fix requeues the suffix once and breaks, gates the report on `*budget == 0`, and drops the reset, so `-1` means "spent and already reported" for the rest of the body. Regressions: `TestContinuationAuditEndorserResolutionIsBudgeted` (now three references past the budget instead of one, which is what makes the duplication visible) and `TestContinuationAuditBudgetStopIsCountedOncePerBody`. Each half is mutation-checked independently; details are in the replies on the two threads. The producer-set cap question is answered on its own thread — it has been unique-id based since `518654d`, with `TestContinuationAuditCapCountsOnlyNewProducers` and `TestContinuationAuditCapDisarmIsExplicit` pinning both halves.

Local verification on this head: `go test ./ledger/` and `go test -race ./ledger/` both pass, `golangci-lint run ./ledger/...` reports 0 issues.

**The red go-test checks.** All three failures are in packages this PR does not touch. This PR's own base (`6fc22fa4`) is green in CI; `pull_request` builds test the merge with `main`, and `main` at `64676163` is currently red on its own — its macOS and Windows go-test jobs are failing there too. Checked out clean at `64676163`, with nothing from this branch:

* `internal/koiosparity` — every `TestCheck*` / `TestAccountLifecycleMismatches*` fails with `create sqlite schema: SQL logic error: table epoch already exists (1)`. Deterministic on clean main. #4017 looks like the fix in flight.
* `ledger/leader` — `TestThresholdPreservesProtocolBoundary/cpraos_fractional_cutoff_admits_its_floor` and `.../tpraos_fractional_cutoff_admits_its_floor` fail with `expected: 0, actual: 1` at `threshold_test.go:96`. Deterministic on clean main, and it appeared with #3897.
* `ledger` — `TestWindowedRewindConvergesWhilePrimaryChainExtends` failed once, on Linux only, with `rollback primary chain to intermediate point at slot 1700: remove block at index 178: block not found`. It is not reproducible: 40/40 passes on clean `64676163` and 40/40 on this branch. Treating it as a pre-existing flake rather than anything this PR introduced; it is unrelated to the continuation audit and belongs in its own change. The race job on this same head corroborates that: it failed only in `internal/koiosparity` and `ledger/leader`, with `ledger` itself passing.

I have not rebased onto `64676163` for that reason — it would inherit the same two deterministic failures rather than clear them. Happy to rebase and re-trigger the moment `main` is green again; the branch still merges cleanly.

chadaibbhot-git:
Rebased onto current `main` (eabf653d), no conflicts — the four `go-test` failures on the previous head were inherited from the stale base, not from this branch.

Re-verified on the new head: `go build ./...`, `go vet ./ledger/...`, `go test ./ledger/`, and `go test -race -run 'TestContinuationAudit|TestEndorserBlock' ./ledger/` — all green.

I also re-ran this PR's mutation checks (revert one guard at a time, the named test must fail). 18 checks, 17 killed — and one survived: dropping the `resolvedEndorserBlocks` memo left every test green. `TestContinuationAuditResolvesEachEndorserBlockOnce` queues its three certifying blocks *before* the first drain, so the insert dedupe alone carries it; the memo's real job is the reference queued *after* the closure was merged, when its dedupe entry has already gone with the drain. That is the ordinary case on a fork window, where one closure is certified by many ranking blocks arriving across the window, and without the memo each of those costs another provider lookup and another hashing pass over every endorser transaction under `chainsyncBlockfetchMutex`.

New commit, test only: `TestContinuationAuditReusesAnAlreadyMergedEndorserBlock` merges the closure, certifies it again from a later block, forces a second drain with an unexplainable input, and asserts the second resolution costs a parent read and no provider call. Removing the memo now fails it — 19/19.

chadaibbhot-git:
Rebased onto `main` at `7351b088`, head `1ac32a85` -> `e0b322cb`.

One conflict, in `ledger/metrics.go` only: #3611 added the two
`reconciliationUndo*` counters at the same two points this branch adds the
continuation-audit outcome counters (the `stateMetrics` struct and its `init`).
Both kept, upstream's first; nothing else changed. The branch's own delta is
otherwise byte-identical to the pre-rebase head, and the diffstat is unchanged
at 2005 insertions / 41 deletions across the same four files.

Re-verified on the new base: `gofmt` clean, `go build ./...`, `go vet
./ledger/...`, `golangci-lint run ./ledger/...` 0 issues, `go test ./ledger/
-count=1` ok 100.3s and `-race` ok 424.2s.

All 19 mutation checks were re-run against the rebased tree, each reverted from
a saved copy of the file rather than by checking it out. 18 of the 19 go red as
before. The 19th is the memoization check: with `resolvedEndorserBlocks` dropped,
`TestContinuationAuditResolvesEachEndorserBlockOnce` stays green — that is the
gap the last commit on this branch was written for, since that test queues its
three certifying blocks before the first drain and so is carried by the insert
dedupe alone. `TestContinuationAuditReusesAnAlreadyMergedEndorserBlock` merges
the closure first and then certifies it again, and it goes red on the same
mutation. Both states are as documented; no behaviour changed in this rebase.

ledger/continuation_audit.go
```diff
@@ -170,11 +325,53 @@ func (ls *LedgerState) auditContinuationBlock(
...
+				)
 				continue
 			}
 			reports++
+			// An unresolved input proves nothing while the window's
+			// producer set has a known hole: the producer may sit in a
+			// certified endorser block the node has fetched but not yet
+			// handed to the audit. Say so quietly instead of asserting
+			// ledger corruption on a healthy node.
+			if window.endorserProducersIncomplete() {
```
chrisguiney, line 355: **Blocker (documentation): `ARCHITECTURE.md` is now factually wrong about this audit and is not updated in this PR.**

`ARCHITECTURE.md:4183-4192` describes the audit as having exactly three resolution paths and then states:

> Anything left over logs `continuation block spends an input with no producer on the local applied chain` [...] and increments `dingo_ledger_continuation_input_unresolved_total`.

After this change neither half of that sentence holds on a Leios chain. There is a fourth resolution path (the certified endorser block, via `leiosCertifiedAnnouncementFromParent` + `EndorserBlockProvider`), and this branch means a left-over input routinely logs at `Debug` and increments `dingo_ledger_continuation_audit_outcomes_total{result="inconclusive_eb_pending"}` *instead of* `continuation_input_unresolved_total`. That is exactly the operator-facing behaviour someone reads `ARCHITECTURE.md` to understand — an operator who greps for the `ERROR` and finds nothing will draw the wrong conclusion from the current text.

This is in scope by the repository's own rule: `AGENTS.md:42-45` / `CLAUDE.md:23-26` make `ARCHITECTURE.md` "part of the change bar, like tests", explicitly for changes to cross-component flows among ledger, database and networking — and the audit now calls into the ouroboros Leios cache from the blockfetch path, which is a new one. Your sibling #3993 updates `ARCHITECTURE.md` for a comparable change.

The paragraph at `ARCHITECTURE.md:4197-4198` survives as written, since it names `continuationAuditMaxProducedTxs` by constant rather than by value.

What is missing is small and well-bounded: the fourth resolution path, the inconclusive outcome and which metric it increments, and the new `dingo_ledger_continuation_audit_outcomes_total` alongside the existing counter. The code itself needs no change for this.

ledger/metrics.go
```diff
@@ -86,6 +86,21 @@ type stateMetrics struct {
 	// durable per-block record, not merely because the content was no
 	// longer reachable. See issue #3778.
 	reconciliationUndoMissingRecord prometheus.Counter
+	// Cross-fork continuation audit outcomes. The first three label values
+	// count one audited input each; disarmed_cap counts one audit window
+	// each. See continuationAuditOutcome for the verdicts.
```
chrisguiney: Non-blocking: `continuationAuditOutcome` is not a symbol in the tree — the constants are `continuationAuditResult*` and the helper is `countContinuationAuditOutcome`. The enumeration above it is also short: "the first three label values count one audited input each; disarmed_cap counts one audit window each" omits `skipped_budget` (per audited body) and `ref_unresolvable` (per abandoned reference), both of which this struct declares. The `Help` string at line 452 has the complete and correct list; this comment is the one that drifted.

ledger/continuation_audit.go
```diff
@@ -233,3 +430,302 @@ func (ls *LedgerState) continuationInputHasProducer(
...
+			}
+			ref.ebHash = ebHash
+			ref.ebSlot = ebSlot
+			ref.resolved = true
+		}
+		key := ref.key()
+		if _, ok := window.resolvedEndorserBlocks[key]; ok {
+			continue
+		}
+		rawTxs, ok := ls.config.EndorserBlockProvider(
```
chrisguiney, line 676: Non-blocking, cost model: the PR description says the common path is "a cache lookup plus a hash per endorser transaction, with no extra I/O beyond the parent-block lookup". `EndorserBlockProvider` resolves to `Ouroboros.EndorserBlockTxsByHash` → `lookupLeiosEndorserBlock`, which on an in-memory miss falls through to `loadLeiosEBFromDB` (`ouroboros/leios_merged.go:1053`): a blob-store manifest read plus `NewLeiosEndorserBlockFromCbor` plus a tx load. So a provider call is not always I/O-free.

It matters because of when this runs: the audit is armed precisely during an endorser-block backlog, can span `continuationAuditBlockBudget` (512) blocks, and holds `chainsyncBlockfetchMutex` throughout — so an older occurrence whose in-memory TTL has elapsed turns into a blob-store read here, up to `continuationAuditMaxEndorserBlocksPerBlock` (32) per audited body. That is bounded and I do not think it needs a code change, but the comment and description read as if the provider were purely a memory lookup.

Separately, and verified rather than reported as a defect: the `resolvedEndorserBlocks` probe in `queueContinuationAuditEndorserRef` (line 534) can only ever hit on the CIP path, because a deferred cert-driven ref is keyed `"c"+parentHash` while this map is keyed `"d"+ebHash+ebSlot`. `TestContinuationAuditReusesAnAlreadyMergedEndorserBlock` pins that cost exactly — `endorserResolutions` 1→2, `providerCalls` stays 1 — so the extra parent read is known and deliberately bounded.

chrisguiney (dismissed):
Reviewed the full diff at `e0b322cb` against merge base `7351b088`. The diagnosis is right, the fix matches the apply path, and the tests are real: I reverted the fix in place (removed the `queueContinuationAuditEndorserRef` call) and 13 of the new tests go red, with `TestContinuationAuditStillReportsMissingProducerOnLeios` passing either way — correctly described as a guard in the PR body.

I verified the two things most likely to be wrong here and both hold:

- **Reference selection cannot drift from apply.** `queueContinuationAuditEndorserRef` mirrors `leiosEndorserBlockForApply` on both paths, including the CIP path's `block.SlotNumber()` slot binding, and the cert-driven path shares `leiosCertifiedAnnouncementFromParent`.
- **Transaction-id derivation is the only correct one available.** The Leios manifest's `TransactionReference.TransactionHash` is Blake2b256 of the *complete* transaction CBOR, not the Cardano tx-id (`ouroboros/leios_merged.go:325-333`), so `EndorserBlockTxHashesByHash` could not have been reused. Hashing `elems[0]` is the tx-id `input.Id()` compares against, and `TestEndorserBlockTxIdsMatchDecodedTransactionHashes` pins it against `NewTransactionFromCbor(TxTypeDijkstra, ...).Hash()`.

Class audit of the CBOR-in-CBOR unwrap: three copies existed, this PR merges the two in `ledger/` into `decodeEndorserTxEnvelope`. The third, `validateLeiosEndorserBlockTx` (`ouroboros/leios_merged.go:364`), is in another package and is stricter (rejects trailing wrapper/envelope bytes), and it runs on the fetch path before anything reaches the provider — so the class is complete, not a gap.

One blocker, documentation-only. I found no behavioural, security, or wire-compatibility defect.

| check | result |
|---|---|
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `gofmt -l .` | 3 files, all pre-existing on `origin/main` (`bark/blob_test.go`, `chain/chain_test.go`, `ledger/read_chain_pipeline_test.go`) — none from this PR |
| `golangci-lint run ./ledger/...` (isolated cache) | 0 issues |
| `go test ./ledger/ -count=1` | 0, 42s |
| fail-before (fix reverted in place) | 13 new tests red |
| required checks on `e0b322cb` | all 9 green |

Skipped, with reasons: Preview Koios parity replay (no chain/Koios infrastructure here — the acceptance check that would actually catch producer-attribution drift on a real chain), devnet, conformance, and Antithesis; `nilaway` and `staticcheck` are not installed locally so `make lint` was not run in full.

Bot coverage on this head: none. `cubic` is `neutral` with "could not start the incremental review"; CodeRabbit has no review record stamped with `commit_id` `e0b322cb` (its last was `f4d587c6`). @wolf31o2's `CHANGES_REQUESTED` is stamped `30b5487a` and its stated cause — a red `go-test (Linux, race)` — is green on this head.

No interaction with sibling #3993: its files are `ledger/forging/*`, `internal/config/*` and `node_forging*`, disjoint from this diff.

ledger/continuation_audit.go
chadaibbhot-git: Valid, and the blocker is accepted rather than argued — thanks for pinning it to
the repository rule rather than to taste. Fixed in `70f87c46`.

You are right that both halves of that sentence stopped holding. The section now
covers:

- **The fourth resolution path.** The certified endorser block, resolved the way
  the apply path resolves it (the block's own announcement on the CIP path, the
  parent's on the cert-driven path) and read through `EndorserBlockProvider`.
  The text says *why* a fourth path is needed at all — a cert-driven ranking
  block has an empty body, so "the block was fetched" no longer implies "its
  transactions are in hand", which is the assumption the original three paths
  rested on.
- **The inconclusive outcome and which metric it increments.** Stated as
  displacement, in your words: while a certified endorser block is unfetched the
  producer set is knowingly incomplete, so an unresolved input records
  `inconclusive_eb_pending` at `Debug` *instead of* taking the
  `missing_producer` path — no `ERROR`, and no increment of
  `continuation_input_unresolved_total`. That is the sentence an operator who
  greps for the `ERROR` and finds nothing needs to read.
- **The new counter alongside the existing one.** All six `result` values with
  their units, including that `missing_producer` still logs the same message and
  still increments `dingo_ledger_continuation_input_unresolved_total`, so the
  old counter's meaning is unchanged rather than silently narrowed.
- **The scheduling**, since it is what makes the fourth path affordable on the
  blockfetch mutex: deferred until the other three paths fail, memoized per
  `(endorser block hash, slot)`, and bounded by
  `continuationAuditMaxEndorserBlocksPerBlock` per audited *body* — the budget is
  owned by the body, not by one drain, which is why several unresolvable inputs
  in one body report one `skipped_budget` between them.

You were right that `ARCHITECTURE.md:4197-4198` survives: it names
`continuationAuditMaxProducedTxs` by constant. I extended that sentence anyway,
because the cap's *consequence* was undocumented — reaching it disarms the
window, logs at `Warn` and counts `disarmed_cap`, and "the audit stopped
covering this node" should not read the same as "this node is clean".

No code change, as you said. `go build ./...`, `go vet`, `golangci-lint run
./ledger/...` 0 issues and `go test ./ledger/ -count=1` green on the new head.

ledger/metrics.go
chadaibbhot-git: Both correct, and both were stale comment rather than stale code. Fixed in
`70f87c46`.

`continuationAuditOutcome` is indeed not a symbol — it was the name in an
earlier draft where the verdicts were a type rather than string constants, and
the comment outlived the refactor. It now points at the
`continuationAuditResult*` constants for the label values and at
`countContinuationAuditOutcome` for the helper, both of which exist.

The enumeration was short for the same reason: it was written when there were
four results and never grew with `skipped_budget` and `ref_unresolvable`, even
though this struct declares a pre-materialized child for each. It now lists all
six with their units, which differ and are the part worth stating —
`clean` / `missing_producer` / `inconclusive_eb_pending` per audited input,
`disarmed_cap` per window, `skipped_budget` per audited body, and
`ref_unresolvable` per abandoned reference.

You are right that the `Help` string was already complete and correct; the drift
was one-directional, so the fix was to bring the comment up to the `Help` string
rather than to touch the metric.

ledger/continuation_audit.go
chadaibbhot-git: Valid on the cost model, and the correction is in `70f87c46` — the comment and
the PR description both overstated it, and you traced the fallback correctly:
`EndorserBlockProvider` → `Ouroboros.EndorserBlockTxsByHash` →
`lookupLeiosEndorserBlock`, with `loadLeiosEBFromDB` on an in-memory miss.

The drain's doc comment now says so explicitly, including the part that makes it
worth stating rather than a footnote: this runs while `chainsyncBlockfetchMutex`
is held, and the audit is armed precisely during the backlog in which an older
occurrence's in-memory entry is likeliest to have expired. So the worst case is
not "a map hit that missed" but a blob-store manifest read, a decode and a
transaction load, on the mutex.

I agree it does not need a code change, and for the reason you gave — it is
bounded on both axes, at most `continuationAuditMaxEndorserBlocksPerBlock` per
audited body and once per `(hash, slot)` for the life of the window. I have
corrected the PR description too, since "no extra I/O beyond the parent-block
lookup" was the sentence that was wrong and it is the one a reader meets first.

On your verified note about the `resolvedEndorserBlocks` probe at line 534: that
matches what I have. The deferred cert-driven key is `"c"+parentHash` and the
memo is keyed `"d"+ebHash+ebSlot`, so the probe can only hit on the CIP path,
and the cert-driven path pays one extra parent read on the requeue.
`TestContinuationAuditReusesAnAlreadyMergedEndorserBlock` is exactly the test
that pins it — `endorserResolutions` 1→2 while `providerCalls` stays 1 — and it
is the reason that commit exists: the older cost regression queues its three
certifying blocks before the first drain, so the insert-side dedupe carries it
alone and dropping the memo leaves it green. Worth recording that the two tests
are not redundant, since that is not obvious from their names.

wolf31o2 (approved):
Checked the one commit since the last review (70f87c46, on top of e0b322cb): documentation and comments only, no logic changes to `ledger/continuation_audit.go` beyond a new doc comment, and `ledger/metrics.go`'s comment update matching the actual label set. It resolves the sole outstanding blocker — the ARCHITECTURE.md section now describes the fourth (Leios/endorser-block) resolution path, the I/O-under-mutex caveat for a cache miss, the budget scoping per audited body, and the `inconclusive_eb_pending` vs `missing_producer` distinction accurately against the code.

All required checks are green at this head (`go-test` Linux/Windows/macOS, race, lint, govulncheck, DCO, Conventional Commits, docker).

No new behavioral, security, or wire-compatibility concern found in this doc-only commit, and the prior review's substantive findings (fail-before-verified fix, no reference-selection drift, no cross-package CBOR-unwrap gap) stand unchanged.

Issue #3764, opened 2026-09-02
ledger: positional gouroboros rule pins break on the next dependency bump


wolf31o2:
## Problem

`ledger/eras/validation.go:112-117` pins gouroboros validation-rule skip
indices by position, and asserts at init time that each index resolves to the
expected function name. Current gouroboros `main` has reordered and grown those
rule lists, so three of the pinned indices no longer match:

| Era | Pinned index | Expected | Resolves to on gouroboros main |
|---|---|---|---|
| babbage | 31 | `UtxoValidatePlutusScripts` | `UtxoValidateMalformedReferenceScripts` |
| conway | 43 | `UtxoValidatePlutusScripts` | `UtxoValidateMalformedReferenceScripts` |
| dijkstra | 38 | `UtxoValidatePlutusScripts` | `UtxoValidateMaxTxSizeUtxo` |

Rule counts on gouroboros main are alonzo 30, babbage 35, conway 56,
dijkstra 57; conway and dijkstra have each grown by one since the pins were
written.

The init-time check will therefore panic on the next gouroboros dependency
bump. All the resolved rules remain in the Always group, so this is a pin
mismatch rather than a change in which rules are skipped.

## Notes

Established by dumping `runtime.FuncForPC` names for all four eras against a
fresh gouroboros `origin/main` worktree. The mismatch is present on gouroboros
main itself and is not introduced by any open PR.

## Acceptance criteria

- Do not pin by position. Resolve the rules to skip by identity — function
  name, or an exported marker gouroboros provides — so a rule inserted upstream
  cannot silently shift the selection.
- Keep a failure loud: if a rule to be skipped cannot be resolved, fail rather
  than silently skipping nothing or skipping the wrong rule.
- Add coverage that fails when an upstream insertion shifts positions, rather
  than only when a name lookup happens to break.
- Confirm against current gouroboros main which rules are actually intended to
  be skipped, since the pins may encode a stale intent.

wolf31o2:
Fixed by #3827. `ledger/eras/validation.go` resolves rules to skip by `Id` (`resolveSkipIndex`, line 414ff) instead of by position, and fails loudly when an id does not resolve. `reflect` and `runtime` name matching are gone, so an upstream insertion shifts nothing.

PR #3990, opened 2026-09-06
test(eras): assert Conway rule identity instead of position

### LINKED ISSUE #3983: ledger/eras tests pin Conway rule indices positionally and fail on main after the v0.202.9 bump
`go test ./ledger/eras/` fails on clean `origin/main` (`90504583`, pinning gouroboros `v0.202.9`):

```
    conway rule index 29 should resolve to UtxoValidateBadInputsUtxo,
        got conway.UtxoValidateNoCollateralInputs
    conway rule index 32 should resolve to UtxoValidateValueNotConservedUtxo,
        got conway.UtxoValidateRequiredRedeemers
    ... does not contain "conway utxo validation rule 29:"
    ... does not contain "conway utxo validation rule 32:"
```

The rules moved: `BadInputs` is now index 30 and `ValueNotConserved` index 33, as the failure output itself shows (`rule 30: bad input(s)`, `rule 33: value not conserved`).

**Production code is not affected.** #3827 replaced positional pinning with `resolveUtxoValidationSkipIndex`, which matches on the stable `descriptor.Id` (`ledger/eras/validation.go:410-425`), so the composed rule set is correct. The defect is that the tests were left pinning positions, which is the same failure mode #3764 described and #3827 removed from the production path.

Introduced by #3956 (`build: adopt gouroboros v0.202.9`). Every open PR now shows these three failures, so each has to be triaged as unrelated before its own results can be read.

## Acceptance criteria

- `ledger/eras` passes on `origin/main`.
- The tests assert on rule **identity**, not position — resolve the index from the descriptor `Id` the same way the production resolver does, so an upstream insertion cannot break them again. `TestConwayUtxoValidationRuleIndexesArePinned` becomes a statement that a named rule is present and reachable, not that it sits at a number.
- The two divergence tests match on the resolved rule rather than a literal `rule NN:` string.

wolf31o2:
Fixes #3983

## Defect

`ledger/eras/conway_utxo_divergence_test.go` pinned literal positions in the composed gouroboros `conway.UtxoValidationRules` slice — `30` for bad inputs and `33` for value not conserved — plus the matching `"conway utxo validation rule 30:"` / `"rule 33:"` substrings. Position is a property of the upstream composition order, not of the rule, so an upstream insertion renumbers every rule after it and breaks the assertions. That has now happened three times: #3764, then #3827 removing positional pinning from production, then #3977 renumbering the tests 29→30 and 32→33.

## Code path

- `ledger/eras/validation.go:425` — `resolveUtxoValidationSkipIndex` resolves an upstream rule by the stable `descriptor.Id`, never by position or function identity. Production has done this since #3827 and is unchanged here.
- `ledger/eras/conway.go:270` — `buildConwayValidationRules` composes the production rule set from `conway.UtxoValidationRuleDescriptors()` and the same Id-keyed skip list.
- `ledger/eras/conway.go:208` — `ValidateTxConway` emits `"conway utxo validation rule %d: %w"` from `validationRule.index`, so the operator-visible number is the upstream position that shifts.

## What the tests now assert

- `conwayUtxoValidationRuleIndex` resolves a rule's index from `conway.UtxoValidationRuleDescriptors()` by Id at run time, the same way `resolveUtxoValidationSkipIndex` does, and checks that the descriptor list and composed rule list agree in length.
- `TestConwayUtxoValidationRuleIndexesArePinned` is renamed `TestConwayUtxoValidationRulesRemainInProductionRuleSet`. Its real purpose is catching a rule silently disappearing from the slice `ValidateTxConway` iterates — the Id-keyed skip list in `buildConwayValidationRules` can drop one — so it asserts that bad inputs and value not conserved are present and reachable in `conwayUtxoValidationRules`, at whatever index upstream declares.
- The two divergence tests resolve the expected index from the rule Id and format the `"conway utxo validation rule %d:"` expectation from it, instead of matching a literal number.

An upstream insertion now renumbers the expectation along with the rule, so the same drift cannot break them again. Removing either rule from the composed set still fails all three.

## Class sweep

Every other reference to an upstream rule position in the repository:

Corrected here, comments only, no assertion involved:

- `ledger/rollback_same_slot_competitor_test.go` — cited bad inputs as rule 29 and value not conserved as rule 32; both stale.
- `ledger/replay_recovery.go` — cited the same pair as rule 22 and rule 24; both stale.
- `ledger/eras/conway_inline_datum_test.go` — cited inline-datums-with-plutus-v1 as rule 18; it is 19.

Left as-is, not positional pins:

- `ledger/replay_recovery_test.go:498` and `ledger/replay_recovery_trust_window_test.go:162` construct synthetic `error` values whose text happens to contain a rule number. The number is opaque payload; no assertion reads it.
- `ledger/eras/validation_rule_id_test.go` and `ledger/eras/validation_test.go` index rule slices with literals, but only into fixture slices the test itself builds, never into an upstream list.
- `internal/test/erastest/pv11_readiness_test.go` already asserts presence in `conway.UtxoValidationRules` by function identity rather than position.
- `ledger/eras/plutus_validity_outcome_test.go:256` already resolves Dijkstra's required-redeemer rule by descriptor Id.
- `DATABASE.md`, `ARCHITECTURE.md`, and `database/plugin/metadata/sqlstore/account_restore_test.go` mention "Conway rule 44/45" for certificate failures. No code in this repository emits that numbering and it does not correspond to the UTXO rule list, so it was left alone rather than retargeted on a guess.

No remaining test or fixture in the repository asserts an upstream rule position.

---
## Summary by cubic
Fixes #3983. The Conway divergence tests pinned literal positions in the upstream `conway.UtxoValidationRules` slice (`30` for bad inputs, `33` for value not conserved), which broke whenever gouroboros inserted a rule. The tests now resolve each rule's index from `conway.UtxoValidationRuleDescriptors()` by stable descriptor ID at run time — the same way production's `resolveUtxoValidationSkipIndex` does — so upstream insertions renumber the expectation along with the rule instead of breaking it.

- `TestConwayUtxoValidationRuleIndexesArePinned` is renamed `TestConwayUtxoValidationRulesRemainInProductionRuleSet`, asserting the bad-inputs and value-not-conserved rules stay present and reachable in the composed rule set rather than pinning a position.
- Comments in `ledger/rollback_same_slot_competitor_test.go` and `ledger/eras/conway_inline_datum_test.go` now name rules instead of citing stale positions.

<sup>Written for commit 5ca761c672df9ba4b55c5810ede36e91899ee06f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3990?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Tests**
  - Updated Conway validation tests to identify rules by stable names rather than fixed numeric positions.
  - Improved coverage to ensure key validation rules remain available in the production rule set.
  - Updated diagnostic assertions to accommodate changing upstream rule indexes.

- **Documentation**
  - Clarified terminology for inline datum validation and value-conservation diagnostics.
  - Updated recovery and rollback comments to reference validation rules by name instead of numeric identifiers.


ledger/eras/conway_utxo_divergence_test.go
```diff
@@ -80,46 +82,93 @@ func newConwayDivergenceTx(
...
-			want,
-			"conway rule index %d should resolve to %s, got %s",
-			idx,
-			want,
-			name,
+			slices.ContainsFunc(
+				conwayUtxoValidationRules,
+				func(rule indexedUtxoValidationRule) bool {
+					return rule.index == index &&
+						rule.validationFunc != nil
```
chrisguiney, line 152: Non-blocking. `rule.validationFunc != nil` is weaker than the name check the old test carried: if a rule were added to the `buildConwayValidationRules` skip list and a dingo replacement re-appended at the same upstream index, this assertion still passes. I mutated exactly that and confirmed it — the case is caught behaviourally by `TestValidateTxConwayGenuinelyMissingInputStillRejected` through `require.ErrorAs`, so nothing is lost overall. Worth a word in the doc comment that identity is established by the divergence tests, not here, so a later reader does not mistake this for an identity assertion.

ledger/eras/conway_utxo_divergence_test.go
```diff
@@ -80,46 +82,93 @@ func newConwayDivergenceTx(
...
+) int {
+	t.Helper()
+	descriptors := conway.UtxoValidationRuleDescriptors()
+	require.Len(
+		t,
+		conway.UtxoValidationRules,
+		len(descriptors),
+		"upstream descriptor list and composed rule list must agree in length",
+	)
+	index := slices.IndexFunc(
```
chrisguiney, line 108: Non-blocking. This `require.Len` + `slices.IndexFunc` + `require.GreaterOrEqual(index, 0)` block is a second copy of `validation_rule_id_test.go:120-133`, in the same package. Both are correct today; one shared helper would keep them from drifting apart on the next gouroboros bump.

chrisguiney (approved):
Reviewed at head `5ca761c6`. Merge-base is `origin/main` (`eabf653d`), so the merged tree equals the head tree. No blockers.

**The oracle is independent, not a restatement of production.** The test resolves the index with its own `slices.IndexFunc` over `conway.UtxoValidationRuleDescriptors()`; production assigns `index: idx` from the position in `conway.UtxoValidationRules` inside `buildIndexedUtxoValidationRulesWithSkips`. Different code paths, so index drift is still caught. Verified by mutating production and re-running the three tests:

| mutation | caught by |
| --- | --- |
| `index: idx` -> `index: len(ret)` (compacted rather than upstream index) | `...GenuinelyUnbalancedStillRejected` (reported rule 31, expected 33) |
| `index: idx` -> `index: idx + 1` | both divergence tests |
| `UtxoValidationRuleBadInputs` appended to the `buildConwayValidationRules` skip list | `...RemainInProductionRuleSet` and `...GenuinelyMissingInputStillRejected` |
| same, but re-appended as a no-op rule at the same upstream index | `...GenuinelyMissingInputStillRejected`, via `ErrorAs` |
| rule Id absent upstream (simulates gouroboros deleting the rule) | `...RemainInProductionRuleSet`: `"-1" is not greater than or equal to "0"` |

The last one settles the silent-zero concern: a missing descriptor yields -1 and fails loudly at `require.GreaterOrEqual`, it does not fall through to rule 0.

**Guards, named as such.** `TestConwayUtxoValidationRulesRemainInProductionRuleSet` passes under both index-drift mutations and under the skip-and-replace mutation. That matches its stated purpose (presence and reachability) and the divergence tests cover those cases, but it is not itself an index check.

**Class audit — Conway is the only era with this pin, and it is now complete.** Only `ledger/eras/conway.go:209` and `ledger/eras/dijkstra.go:254` emit `utxo validation rule %d`; there is no Dijkstra divergence test, and Babbage/Alonzo/Mary/Allegra/Shelley never emit a position, so no other era carries the fragility. The one remaining literal in the tree is `ledger/replay_recovery_test.go:498` (`rule 38`) — opaque payload in a synthetic `Cause` with no assertion reading it, same class as `replay_recovery_trust_window_test.go:162`. Correctly left alone. `internal/test/erastest/pv11_readiness_test.go` compares by function pointer and `ledger/eras/validation_test.go:1443` indexes a test-built fixture; both are fine.

**Non-blocking, PR body vs. diff.** The class sweep lists `ledger/replay_recovery.go` under "Corrected here", but at head that file is byte-identical to `origin/main`. Commit `68a0464e` did edit it; the later merge from main took main's side and dropped the edit. The comment there still reads "On the currently pinned v0.202.6 they print as rule 30 and rule 33" while the repo pins v0.202.10 — the numbers happen to still be correct, the version citation is not.

**Sibling #4012 composes.** Its only rule number is `"shelley utxo validation rule 8: %w"` wrapping a real `shelley.BadInputsUtxoError{}` in a test fixture; classification matches the error type, not the number, so it adds no new positional pin.

**Checks at head:** all 13 green. No bot reviewed this head — cubic's only review is stamped at `68a0464e`, its check on `5ca761c6` reports it skipped the merge commit, and CodeRabbit posted "Auto incremental reviews are disabled". This review stands in for bot review.

**Local:** `go build ./...` 0, `go vet ./ledger/...` 0, `gofmt -l` clean on all three files, `golangci-lint run ./ledger/eras/... ./ledger/` 0 issues, `go test ./ledger/eras/` ok, `go test ./ledger/` ok. `make golines` skipped: golines is not installed in this environment (the CI `lint` job is green at this head).

ledger/eras/conway_inline_datum_test.go
```diff
@@ -26,9 +26,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// conway.UtxoValidateInlineDatumsWithPlutusV1 is upstream Conway UTXO rule 18.
-// Dingo installs it unmodified, so nothing else in this repository pins its
-// behavior. The cases below cover the three properties the rule has to get
+// conway.UtxoValidateInlineDatumsWithPlutusV1 is the upstream Conway
+// inline-datums-with-plutus-v1 UTXO rule. Dingo installs it unmodified, so
+// nothing else in this repository pins its behavior. The cases below cover the three properties the rule has to get
```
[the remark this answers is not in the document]
chrisguiney: Already addressed at head `5ca761c6`. `ledger/eras/conway_inline_datum_test.go` now has no line over 80 characters (`awk 'length>80'` returns nothing), and commit `9ff4afb8` rewrapped the comment. The wider class is fine too: golines is invoked as `golines -w --ignore-generated --chain-split-dots --max-len=80 --reformat-tags .` with no `--shorten-comments`, so it does not reflow comments, and it does not split string literals — `ledger/rollback_same_slot_competitor_test.go:277` is a 95-character string literal that predates this PR and survives the repo's own golines runs. The `lint` job is green at this head.

PR #3883, opened 2026-09-04
fix(txsubmission): validate returned transaction batches 

arepala-uml:
## Summary

Validate TxSubmission transaction-body replies against the outstanding request before admitting any transaction to the mempool. This prevents mismatched replies from partially advancing local or protocol state.

## Changes

- Recompute and compare every returned transaction hash.
- Require the returned transaction count to match the requested count.
- Enforce transaction order by matching each body against its corresponding requested ID.
- Validate each returned transaction’s era.
- Validate individual transaction sizes against advertised sizes.
- Validate the total returned byte size against the request budget.
- Decode and validate the complete batch before admitting any transaction.
- Reject the entire batch if any transaction is malformed or mismatched.
- Stop the peer pull loop after rejection to avoid acknowledging and advancing past an invalid reply.
- Reuse decoded transactions during mempool admission.
- Add coverage for:
  - Successful matching batches.
  - Count mismatches.
  - Order mismatches.
  - Era mismatches.
  - Size mismatches.
  - Hash mismatches.
  - Atomic rejection when a valid prefix is followed by a mismatched transaction.
  - Prevention of further protocol advancement after rejection.

Additional testing notes:

- The full `make test` run exceeded the available execution-time limit while running the repository’s lengthy database tests. No failure related to TxSubmission or the changed files was observed.
- Socket-dependent tests passed when rerun outside the restricted sandbox.
- The all-Dingo DevNet started successfully, and all four nodes became healthy. The Go DevNet test package is Linux-only and therefore reported `matched no packages` on macOS.

Closes #3505.

---
## Summary by cubic
Fixes TxSubmission so returned transaction batches are fully validated against the outstanding request before any transaction is admitted to the mempool, preventing mismatched replies from partially advancing local or protocol state. Closes #3505.

**Bug Fixes**
- Verifies order, hash, era, and sizes of every returned transaction against the request.
- Accepts ordered subsets of the requested transactions, allowing peers to omit transactions that disappeared after advertisement.
- Rejects the whole batch if any transaction is malformed or mismatched, then stops the pull loop.
- Reuses decoded transactions during mempool admission.

<sup>Written for commit 38c6713eaec6450ccb70ec2e10355f647eb72601. Summary will update on new commits.</sup>

[...]

Issue #3505, opened 2026-08-25
txsubmission: validate returned transaction batches


wolf31o2:
TxSubmission accepts returned transaction bodies without checking that their
hashes, era, count, order, and total size match the outstanding request. A
response can move protocol state forward without corresponding to the requested
transactions.

Acceptance criteria:

- Recompute and compare every returned transaction hash.
- Enforce requested count, order, era, and byte limits.
- Reject a mismatched batch without partially advancing state.
- Cover successful and mismatched replies.

PR #3992, opened 2026-09-06
fix(txsubmission): compare advertised size against the wrapped wire size

### LINKED ISSUE #3994: txsubmission: reply size validation compares unwrapped tx bytes to the peer's advertised wire size, rejecting every batch from cardano-node peers
## Problem

Since #3883 landed on `main`, a Dingo node acting as the tx-submission server rejects **every** transaction batch offered by a `cardano-node` peer:

```
level=ERROR msg="rejected mismatched txsubmission reply" component=ouroboros component=network
  protocol=tx-submission role=server connection_id=<redacted>
  error="txsubmission reply size mismatch at index 0: requested 2338, received 2331"
```

The failure is always `at index 0` (the first tx of the batch), so no transaction in any batch is ever admitted. On a block producer observed over 24h: 52 rejections, 0 successful admissions. The same transaction is re-offered and re-rejected indefinitely.

This reproduces against every `cardano-node` peer we have — both a local Haskell relay and ~20 unrelated public testnet relays. It does not reproduce Dingo↔Dingo (see "Why CI didn't catch this").

## Evidence

137 rejections collected across three Dingo block producers. The delta between `requested` and `received` is **not** constant — it is 6 or 7 bytes, and it correlates perfectly with the size of the received tx:

| delta | received size band | count |
|---|---|---|
| 6 | 24–255 | 88 |
| 7 | 256–65535 | 49 |

Zero exceptions in 137 samples. The boundary is exactly 256 — i.e. it tracks the width of a CBOR byte-string length prefix.

Sample pairs: `244/238`, `343/336`, `673/666`, `981/974`, `1032/1025`, `1476/1469`, `2338/2331`, `2584/2577`, `5824/5817`.

## Root cause

`requested` and `received` are two different quantities, and the check assumes they are the same.

**`received`** is the *inner* tx CBOR, unwrapped. `gouroboros` decodes each `MsgReplyTxs` item `[eraId, #6.24(bytes)]` and stores only the tag-24 payload in `TxBody.TxBody` (`protocol/txsubmission/messages.go`, `TxBody` / `MarshalCBOR` — the encoder shows the framing that the decoder strips).

**`requested`** is `want.Size`, taken verbatim from the peer's `MsgReplyTxIds`. For a `cardano-node` peer that value is `txWireSize`, which is defined as the size of the tx *as it appears on the wire, including the CBOR-in-CBOR and hard-fork envelopes*.

From `ouroboros-consensus`, `Ouroboros/Consensus/Ledger/SupportsMempool.hs:391-392`:

```haskell
  -- | The size of the transaction from the perspective of diffusion layer
  txWireSize :: GenTx blk -> Network.SizeInBytes
```

Note this is a *different* method from `txInBlockSize`, which is documented two lines later as the size "when it's in a block". The diffusion layer deliberately advertises the wire size, not the block-body size.

`Ouroboros/Consensus/NodeKernel.hs:921-925` — this is the value that goes into `MsgReplyTxIds`:

```haskell
        { mempoolTxIdsAfter = \idx ->
            [ ( txId (txForgetValidated tx)
              , idx'
              , txWireSize $ txForgetValidated tx
              )
```

### The exact 6 / 7 bytes

`txWireSize` for the Cardano hard-fork block is composed of two layers.

`Ouroboros/Consensus/HardFork/Combinator/Mempool.hs:312-324` adds the era envelope:

```haskell
  txWireSize =
    \tx ->
      let ...
          -- HFC overhead
          overhead
            | nsToIndex tx' <= 23 = 2
            | otherwise = 3
       in (+ overhead) ...
```

`Ouroboros/Consensus/Shelley/Ledger/Mempool.hs:507-520` adds the CBOR-in-CBOR envelope:

```haskell
wrapCBORinCBOROverhead size =
  2 -- wrapCBORinCBOR's encodeTag 24
    + case size of
      _
        | size <= 0x17 -> 1
        | size <= 0xff -> 2
        | size <= 0xffff -> 3
        | otherwise -> 5
    + fromIntegral size
```

Summing, for the era indices in use (≤ 23):

```
2  HFC envelope        : array(2) header 0x82  +  1-byte era index
2  tag 24              : 0xd8 0x18
n  byte-string header  : 1 (len<24) / 2 (len<256) / 3 (len<65536) / 5
```

giving a total overhead of **6 bytes** for an inner tx of 24–255 bytes and **7 bytes** for 256–65535 bytes — exactly the observed distribution, including the boundary at 256 and the absence of any other delta.

Every observed pair reproduces: inner 238 → `82 06 d8 18 58 ee` + 238 = 244. Inner 2331 → `82 06 d8 18 59 09 1b` + 2331 = 2338.

So the peer is correct and Dingo is comparing against the wrong quantity.

### Dingo advertises the wrong size too

The mirror-image bug exists on the sending side: when Dingo answers `MsgRequestTxIds` it advertises the *inner* size:

```go
		txSize := len(tmpTx.Cbor)
		...
			txsubmission.TxIdAndSize{
				...
				Size: uint32(txSize),
			},
```

`cardano-node` tolerates this because `ouroboros-network`'s inbound side uses the advertised size only as a flow-control / request budget (`SizeInBytes` in `TxSubmission/Inbound/V2/State.hs`), never as a strict equality check. But it means Dingo under-advertises by 6–7 bytes per tx, which is a protocol conformance bug independent of the rejection above.

## Why CI didn't catch this

The tests added in #3883 drive a Dingo↔Dingo harness (`newTxSubmissionRelayHarness`), where the offering side is Dingo's own client. Since Dingo advertises `len(Cbor)` and Dingo measures `len(TxBody)`, the two agree and the check passes. The mismatch only appears against a peer that advertises the spec-correct wire size — i.e. every `cardano-node`.

## Impact

A Dingo block producer on `main` accepts **no** transactions from the network over tx-submission, so it forges empty blocks. Additionally, the rejection path at `ouroboros/txsubmission.go` does:

```go
			if err != nil {
				o.config.Logger.Error(
					"rejected mismatched txsubmission reply",
					...
				)
				return
			}
```

That `return` exits the per-connection pull goroutine started in `txsubmissionServerInit`, so tx ingest from that peer stops for the life of the connection rather than skipping the batch.

Not present in any tagged release: #3883 merged 2026-09-04T18:41Z, after v0.70.6 was published at 15:19Z the same day. `main` only.

## Minimal repro

1. Run a Dingo node built from `main` peered with any `cardano-node` relay on a network with transaction traffic.
2. Wait for the peer to offer a transaction.
3. Every batch is rejected with `txsubmission reply size mismatch at index 0`, with `requested - received` equal to 6 or 7.

## Suggested fix

Compare like with like. Two options:

1. **Preferred** — compute the wire size from the decoded body and compare that against `want.Size`, mirroring `txWireSize`:

```go
func txWireSize(eraId uint16, body []byte) uint64 {
	n := uint64(len(body))
	var lenHdr uint64
	switch {
	case n <= 0x17:
		lenHdr = 1
	case n <= 0xff:
		lenHdr = 2
	case n <= 0xffff:
		lenHdr = 3
	default:
		lenHdr = 5
	}
	hfc := uint64(2) // array(2) header + 1-byte era index
	if eraId > 23 {
		hfc = 3
	}
	return hfc + 2 /* tag 24 */ + lenHdr + n
}
```

2. **Or** treat the advertised size as an upper bound rather than an equality, which is how `ouroboros-network` itself uses it. The total-batch budget check just above already works this way.

The same correction should be applied to the advertised size in `txsubmissionClientRequestTxIds` so Dingo advertises `txWireSize` rather than `len(Cbor)`.

Separately, consider making the rejection path skip the batch and continue the pull loop instead of returning, so a single bad reply cannot silently end tx ingest from a peer — this is the same concern as #3449 / #3461.

Adding one interop test that offers a batch with a wire-size-advertised entry (advertised = inner + 6/7) would have caught this.

chadaibbhot-git:
## Problem

A Dingo node acting as the tx-submission **server** rejects every transaction batch offered by a `cardano-node` peer, so it ingests nothing over tx-submission and forges empty blocks.

```
level=ERROR msg="rejected mismatched txsubmission reply" component=network
  protocol=tx-submission role=server
  error="txsubmission reply size mismatch at index 0: requested 2338, received 2331"
```

The failure is always `at index 0`, so no transaction in any batch is ever admitted, and the rejection `return`s out of the per-connection pull goroutine — tx ingest from that peer is dead for the life of the connection rather than for one batch.

This is the size-validation regression from #3883. It reproduces against every `cardano-node` peer; it does not reproduce Dingo-to-Dingo, which is why the harness tests added with #3883 stayed green.

## Evidence

137 rejections collected on three producers. The delta between `requested` and `received` is not constant — it is 6 or 7 bytes and it tracks the size of the received body:

| delta | body size band | count |
|---|---|---|
| 6 | 24–255 | 88 |
| 7 | 256–65535 | 49 |

Zero exceptions in 137 samples; the boundary is exactly 256, i.e. the width of a CBOR byte-string length prefix. Sample pairs: `244/238`, `343/336`, `673/666`, `2338/2331`, `5824/5817`.

## Root cause

`requested` and `received` are two different quantities.

- **`received`** is the *unwrapped* transaction CBOR. gouroboros decodes each `MsgReplyTxs` item `[eraId, #6.24(bytes)]` and stores only the tag-24 payload in `TxBody.TxBody` (`protocol/txsubmission/messages.go`; the `MarshalCBOR` encoder shows the framing the decoder drops).
- **`requested`** is `want.Size`, taken verbatim from the peer's `MsgReplyTxIds`. For a `cardano-node` peer that is `txWireSize` — `ouroboros-consensus`, `Ouroboros/Consensus/Ledger/SupportsMempool.hs`, documented as "the size of the transaction from the perspective of diffusion layer" and deliberately distinct from `txInBlockSize`. It is the value fed into `MsgReplyTxIds` from `NodeKernel.hs`.

`txWireSize` for the hard-fork block is the sum of two envelopes — the HFC envelope (`HardFork/Combinator/Mempool.hs`: 2 bytes for era index ≤ 23) and the CBOR-in-CBOR envelope (`Shelley/Ledger/Mempool.hs`: 2 bytes of `encodeTag 24` plus a 1/2/3/5-byte byte-string length header):

```
1  array(2) header       0x82
1  era ID uint           (era IDs <= 23)
2  tag 24                0xd8 0x18
n  byte-string header    1 (<24) / 2 (<256) / 3 (<65536) / 5
```

which is 6 bytes of overhead for a 24–255 byte body and 7 for 256–65535 — exactly the observed distribution, including the boundary and the absence of any other delta. Every observed pair reproduces: inner 238 → `82 06 d8 18 58 ee` + 238 = 244; inner 2331 → `82 06 d8 18 59 09 1b` + 2331 = 2338.

The peer is right and Dingo was comparing against the wrong quantity.

### Dingo advertised the wrong size too

Mirror-image bug on the sending side: `txsubmissionClientRequestTxIds` advertised `len(tmpTx.Cbor)`, the unwrapped size. `cardano-node` tolerates this because `ouroboros-network` uses the advertised size only as a flow-control / request budget, never as a strict equality check — but Dingo was under-advertising by 6–7 bytes per transaction, which is a conformance bug on its own. It is also what hid the server-side bug: Dingo advertised `len(Cbor)` and measured `len(TxBody)`, so the two agreed in the Dingo-to-Dingo harness.

## Fix

1. **Server / inbound.** Add `txsubmissionWireSize(eraId, bodyLen)`, deriving the wrapped size from the era ID head and the CBOR head width the body length implies. `validateTxsubmissionReply` accepts an advertisement equal to either the unwrapped body size (peers that advertise like Dingo did, so a mixed fleet interoperates) or the exact derived wire size, and rejects anything else. This stays an equality check against two known-legitimate values rather than becoming a range check: `wire - 1`, `wire + 1` and `body + 8` are all still rejected. The mismatch error now reports the advertised size, the body size, the derived wire size and the era.
2. **Server / inbound, pull loop.** A rejected reply no longer ends tx ingest for the connection. The whole reply is dropped — a partially valid batch is still not trusted, so #3883's atomicity is preserved — then the loop backs off and keeps pulling, in the spirit of #3461. A peer that returns nothing but mismatched replies is given up on after 3 consecutive rejections, so this cannot become a hot loop; that give-up is logged separately. In the deployed configuration the admission-headroom path requests one ID at a time, so "drop the reply" and "skip the item" coincide there.
3. **Client / outbound.** `txsubmissionClientRequestTxIds` now advertises the wire size via the same helper, so peers get spec-correct sizes. Safe because they use it only for flow control against the bytes they actually read off the wire.
4. **Metrics.** New counter `dingo_txsubmission_reply_size_mismatch_total{outcome="accepted_wire_size"|"rejected"}`, both label values pre-materialized at zero so an alert does not have to tolerate a missing series.

The admission-headroom reservation now sizes against the advertised wire size, which is an upper bound on the body that gets admitted — conservative by the few bytes of wrapper.

## Tests

- `TestTxsubmissionWireSizeMatchesWireEncoding` — derived size vs the real `TxBody.MarshalCBOR` output for body lengths 0, 1, 23, 24, 255, 256, 65535, 65536 across era IDs 0, 6, 23, 24, 255, covering all four length-header bands.
- `TestTxsubmissionWireSizeOverheadBands` — pins the 6- and 7-byte overheads and the band boundaries.
- `TestValidateTxsubmissionReplyAcceptsWireSizeAdvertisement` — the regression test: a reply whose advertised size is the wrapped size is accepted. Fails on `main` with `size mismatch`.
- `TestValidateTxsubmissionReplyRejectsGenuineSizeMismatch` — body ± 1, wire ± 1, body + 8, double and zero are all still rejected.
- `TestTxSubmissionRelayAdmitsWireSizeAdvertisedTransaction` — end to end over the real relay harness, now that the client advertises the wire size, plus the `accepted_wire_size` counter.
- `TestTxSubmissionServerInitRejectsMalformedReply` — updated: a single mismatched reply is dropped and the pull loop still admits the next offer.
- `TestTxSubmissionServerInitStopsAfterRepeatedMismatches` — ingest stops after repeated mismatches.
- `TestTxsubmissionReplySizeMetricPreMaterialized`, `TestRecordTxsubmissionReplySizeWithoutMetrics`.
- `TestTxSubmissionClientRequestTxIds` — updated to expect the wire size.

`go build ./...`, `go vet ./...`, `golangci-lint run ./ouroboros/...` clean; `go test ./ouroboros/... ./mempool/... . -count=1` green, `go test ./ouroboros/... -race` green. Each change was mutation-checked: neutralizing the derivation, restoring strict inner equality, restoring the client's unwrapped advertisement, restoring the immediate `return` on rejection, and removing the metric pre-materialization each turn the corresponding test red.

## gouroboros follow-up

`MsgReplyTxs` decoding discards the wrapper, so the wire size cannot be observed directly from Go today and this PR derives it. It would be cleaner for gouroboros to surface the wrapped item length on `TxBody`; that would let this comparison use the observed value instead of a derived one. Worth noting for blinklabs-io/gouroboros#2076, which asks for count/size validation in the library: whatever it adds must not become a strict equality against the inner body size, or it reintroduces exactly this failure for every `cardano-node` peer.

Related: #3883, #3505, #3461, blinklabs-io/gouroboros#2076

---
## Summary by cubic
Fixes tx-submission size validation so Dingo accepts transactions from `cardano-node` peers instead of rejecting every batch; previously the server compared the advertised size to the unwrapped body length, which always mismatched for `cardano-node` peers, so ingest stopped. Now it accepts either the unwrapped body size or the derived wire size, drops a single mismatched reply and keeps pulling, gives up only after 3 consecutive mismatches, and the client also advertises the wire size, with a new metric tracking outcomes.

**Details**
- Derives the wire size from era ID and body length, covering all CBOR length-header bands.
- The mismatch error reports advertised, body, wire, and era sizes; undersized advertisements are now checked per body before the aggregate budget, so they count as size mismatches, and both metric outcomes count bodies (one per accepted body, every body of a dropped reply).
- After a mismatch, the pull loop backs off and continues; it stops only after 3 consecutive rejections.
- Client-side advertisement now matches the wire size, fixing a conformance bug.

<sup>Written for commit 45b8a6b3937259f8ee2001aa1eda2dbea64d1536. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3992?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Transaction-submission size validation now supports both standard and wrapped wire-size advertisements.
  * Added metrics for accepted wire-size replies and rejected size mismatches.
* **Bug Fixes**
  * Valid transaction replies are accepted when peers advertise their wire-encoded size.
  * Occasional size mismatches are retried with backoff while valid replies reset the mismatch count.
  * Transaction ingestion stops after three consecutive size mismatches to prevent repeated invalid responses.

Fixes #3994.

### DIFF
--- a/ouroboros/metrics_protocol.go
+++ b/ouroboros/metrics_protocol.go
@@ -21,10 +21,22 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
+// Outcomes for the txsubmission reply size mismatch counter.
+const (
+	// txsubmissionReplySizeAcceptedWire counts bodies whose advertised size
+	// was the wrapped wire size rather than the unwrapped body size. This
+	// is the normal, spec-correct case for a cardano-node peer.
+	txsubmissionReplySizeAcceptedWire = "accepted_wire_size"
+	// txsubmissionReplySizeRejected counts replies dropped because an
+	// advertised size matched neither the body size nor the wire size.
+	txsubmissionReplySizeRejected = "rejected"
+)
+
 type protocolMetrics struct {
-	messagesReceived             *prometheus.CounterVec
-	messageDuration              *prometheus.HistogramVec
-	txsubmissionAdmissionRetries prometheus.Histogram
+	messagesReceived              *prometheus.CounterVec
+	messageDuration               *prometheus.HistogramVec
+	txsubmissionAdmissionRetries  prometheus.Histogram
+	txsubmissionReplySizeMismatch *prometheus.CounterVec
 }
 
 func (o *Ouroboros) initProtocolMetrics() {
@@ -52,6 +64,13 @@ func (o *Ouroboros) initProtocolMetrics() {
 			},
 			[]string{"protocol", "outcome"},
 		),
+		txsubmissionReplySizeMismatch: factory.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "dingo_txsubmission_reply_size_mismatch_total",
+				Help: "tx-submission reply bodies whose length differed from the size the peer advertised, by outcome",
+			},
+			[]string{"outcome"},
+		),
 		txsubmissionAdmissionRetries: factory.NewHistogram(
 			prometheus.HistogramOpts{
 				Name: "dingo_txsubmission_admission_retry_streak",
@@ -62,6 +81,12 @@ func (o *Ouroboros) initProtocolMetrics() {
 			},
 		),
 	}
+	// Pre-materialize both outcomes so a dashboard or alert sees a zero
+	// series before the first mismatch rather than a missing one.
+	o.protocolMetrics.txsubmissionReplySizeMismatch.
+		WithLabelValues(txsubmissionReplySizeAcceptedWire).Add(0)
+	o.protocolMetrics.txsubmissionReplySizeMismatch.
+		WithLabelValues(txsubmissionReplySizeRejected).Add(0)
 }
 
 // recordProtocolMessage records one received message and the callback
@@ -86,6 +111,17 @@ func (o *Ouroboros) recordProtocolMessage(
 	o.protocolMetrics.messageDuration.
 		WithLabelValues(protocol, outcome).
 		Observe(dur.Seconds())
+}
+
+// recordTxsubmissionReplySize records count reply bodies resolved with the
+// given outcome. Safe to call when metrics are not initialized.
+func (o *Ouroboros) recordTxsubmissionReplySize(outcome string, count int) {
+	if o.protocolMetrics == nil || count <= 0 {
+		return
+	}
+	o.protocolMetrics.txsubmissionReplySizeMismatch.
+		WithLabelValues(outcome).
+		Add(float64(count))
 }
 
 func (o *Ouroboros) recordTxsubmissionAdmissionRetry(streak int) {
--- a/ouroboros/metrics_protocol.go
+++ b/ouroboros/metrics_protocol.go
@@ -27,8 +27,10 @@ const (
 	// was the wrapped wire size rather than the unwrapped body size. This
 	// is the normal, spec-correct case for a cardano-node peer.
 	txsubmissionReplySizeAcceptedWire = "accepted_wire_size"
-	// txsubmissionReplySizeRejected counts replies dropped because an
-	// advertised size matched neither the body size nor the wire size.
+	// txsubmissionReplySizeRejected counts the bodies of replies dropped
+	// because an advertised size matched neither the body size nor the
+	// wire size. The whole reply is dropped, so every body it carried is
+	// counted, keeping the unit the same as the accepted outcome.
 	txsubmissionReplySizeRejected = "rejected"
 )
 
@@ -67,7 +69,7 @@ func (o *Ouroboros) initProtocolMetrics() {
 		txsubmissionReplySizeMismatch: factory.NewCounterVec(
 			prometheus.CounterOpts{
 				Name: "dingo_txsubmission_reply_size_mismatch_total",
-				Help: "tx-submission reply bodies whose length differed from the size the peer advertised, by outcome",
+				Help: "tx-submission reply bodies, counted in bodies not replies, whose length differed from the size the peer advertised: accepted_wire_size counts bodies advertised at the wrapped wire size, rejected counts every body of a reply dropped for a size mismatch since the whole reply is dropped",
 			},
 			[]string{"outcome"},
 		),
--- a/ouroboros/txsubmission.go
+++ b/ouroboros/txsubmission.go
@@ -36,6 +36,9 @@ const (
 	txsubmissionMaxBackoff               = 5 * time.Second // Cap on exponential backoff wait
 	txsubmissionBaseBackoff              = 150 * time.Millisecond
 	txsubmissionLogEvery                 = 10 // Log every Nth rate limit hit after the 1st
+	// Give up on a peer only after this many consecutive rejected replies.
+	// A single bad reply drops that reply and keeps the pull loop running.
+	txsubmissionMaxConsecutiveReplyMismatches = 3
 )
 
 var (
@@ -45,16 +48,68 @@ var (
 	errTxsubmissionAdmissionStopped = errors.New(
 		"txsubmission admission wait stopped",
 	)
+	errTxsubmissionReplySizeMismatch = errors.New(
+		"txsubmission reply size mismatch",
+	)
 )
+
+// cborHeadLen returns the encoded length in bytes of a CBOR head (the
+// initial byte plus any following argument bytes) whose argument is v.
+func cborHeadLen(v uint64) uint64 {
+	switch {
+	case v <= 0x17:
+		return 1
+	case v <= 0xff:
+		return 2
+	case v <= 0xffff:
+		return 3
+	case v <= 0xffffffff:
+		return 5
+	default:
+		return 9
+	}
+}
+
+// txsubmissionWireSize returns the size of one MsgReplyTxs item as it
+// appears on the wire, given the era ID and the length of the unwrapped
+// transaction body.
+//
+// gouroboros decodes each item -- [eraId, #6.24(bytes)] -- and keeps only
+// the tag-24 payload, so TxBody.TxBody is the inner transaction CBOR. A
+// cardano-node peer advertises the spec-correct wire size in MsgReplyTxIds
+// (ouroboros-consensus SupportsMempool.txWireSize, which is deliberately
+// distinct from txInBlockSize), covering the whole wrapped item:
+//
+//	array(2) header + era ID uint  2 bytes for era IDs <= 23
+//	tag 24 (0xd8 0x18)             2 bytes
+//	byte-string length header      1, 2, 3, 5 or 9 bytes
+//
+// so the advertised value exceeds len(TxBody) by 6 bytes for a 24..255 byte
+// body and by 7 bytes for a 256..65535 byte body.
+func txsubmissionWireSize(eraId uint16, bodyLen int) uint64 {
+	body := uint64(0)
+	if bodyLen > 0 {
+		body = uint64(bodyLen)
+	}
+	// 1 byte for the definite-length array(2) header, the era ID head, the
+	// two-byte tag 24 head, then the byte string itself.
+	return 1 + cborHeadLen(uint64(eraId)) + 2 + cborHeadLen(body) + body
+}
 
 type validatedTxsubmissionBody struct {
 	body txsubmission.TxBody
 	tx   ledger.Transaction
+	// wireSizeAdvertised records that the peer advertised the wrapped wire
+	// size for this body rather than the unwrapped body size.
+	wireSizeAdvertised bool
 }
 
 // validateTxsubmissionReply verifies the complete reply before its first
 // transaction is admitted. The advertised sizes are part of the request
-// budget, so each body must have the exact size the peer advertised.
+// budget, so each body must have exactly one of the two sizes the peer can
+// legitimately have advertised for it: the unwrapped body size, or the
+// wrapped wire size that cardano-node advertises (see txsubmissionWireSize).
+// Anything else is a mismatch.
 func validateTxsubmissionReply(
 	requested []txsubmission.TxIdAndSize,
 	returned []txsubmission.TxBody,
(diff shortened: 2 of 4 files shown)


chadaibbhot-git:
Both findings were real and are fixed in 45b8a6b3.

**1. Undersized advertisement was dropped without being classified or counted.** When a peer advertised a size *smaller* than the body it returned, the aggregate byte-budget check fired first (`returnedBytes > requestedBytes`) and returned an error that did not wrap `errTxsubmissionReplySizeMismatch`, so the reply was dropped with an unattributed batch-budget message and no `rejected` observation. The per-body size check now runs before the aggregate budget, so every rejected size advertisement is reported with the same detail — advertised, body, wire and era — and counted.

Since every accepted body is no larger than its own advertisement and each advertisement is consumed at most once (`nextRequested` only advances), the aggregate budget is now an invariant backstop that a size advertisement alone cannot trip; it is kept deliberately as a guard against future loosening of the per-body check. The reorder does mean a body is decoded before the aggregate check, but the work stays bounded: the reply-count check still runs first, so at most `len(requested)` bodies are decoded, which is what the success path does anyway.

**2. Units.** Correct — `accepted_wire_size` counted bodies while `rejected` counted one per reply. Both now count **bodies**. I picked "every body carried by the dropped reply" rather than "the number of mismatched bodies", because the whole reply is dropped atomically, so N bodies really are discarded; counting only the first mismatch would understate the loss and would require validating past the first error. The unit is documented in the metric help string and on the outcome constants.

The recording moved into `recordTxsubmissionReplyOutcome(validated, replyBodies, err)` so the exact counts are unit-testable. New tests:

- `TestValidateTxsubmissionReplyUndersizedAdvertisementIsCounted` — advertisement below the body size is classified as a size mismatch, carries the full detail, and increments `rejected`.
- `TestRecordTxsubmissionReplyOutcomeCountsBodies` — three-body reply: accepted adds exactly 3 to `accepted_wire_size`; a bad advertisement at index 0, 1 or 2 each adds exactly 3 to `rejected` and 0 to `accepted_wire_size`.
- `TestValidateTxsubmissionReplyRejectsGenuineSizeMismatch` — the "one below body" and "zero" cases now expect a size mismatch rather than the byte-limit error.

Both fixes were mutation-checked: moving the aggregate budget check back ahead of the per-body size check turns the undersized and the below-body/zero cases red; counting one per reply instead of per body for either outcome turns the three-body count assertions red.

`go vet ./...`, `golangci-lint run ./ouroboros/...` clean; `go test ./ouroboros/... ./mempool/... . -count=1` and `go test ./ouroboros/... -race` green.

wolf31o2 (approved):

PR #4012, opened 2026-09-06
fix(ledger): classify a missing redeemer as deterministic

### LINKED ISSUE #4000: ledger: a missing-redeemer rejection is not deterministic, so recovery rewinds instead of failing cleanly
`isDeterministicTxValidationError` (`ledger/replay_recovery.go:392`) classifies four verdicts: `shelley.DuplicateInputError`, `conway.PlutusScriptFailedError`, the reward-withdrawal mismatches, and `eras.DuplicateInputByronError`. `lcommon.MissingRedeemerForScriptError` is not among them, although dingo already handles that type elsewhere (`ledger/eras/conway.go`).

A missing redeemer is a property of the transaction alone. No amount of local replay changes the verdict, so it belongs in the same class as `PlutusScriptFailedError` — which sits immediately above it in that function.

Unclassified, such a rejection takes the producer-resolution and rewind path instead: the node rewinds, requests a fresh intersection, is served the same block, and rejects it again. That is the amplifier described in #3889 and it wedges rather than stopping.

## Why now

gouroboros #2259 (`Fixes #2250`) widens `UtxoValidateRequiredRedeemers` from the spend purpose alone to mint, cert, reward, vote and propose. Transactions carrying a Plutus script of a non-spend purpose without its redeemer are currently accepted and their scripts never execute; after that change they are rejected with `MissingRedeemerForScriptError`. The rejection is correct, but the classification gap turns a clean stop into a rewind loop, so it is worth closing before a bump rather than after a node wedges.

The direction matters: this makes newly-rejected blocks *reachable*, so it is a case of a fix making a previously dead path live rather than a defect the widening introduces.

## Acceptance criteria

- `isDeterministicTxValidationError` classifies `lcommon.MissingRedeemerForScriptError`, with a comment saying why it cannot be resolved by replay.
- A test asserts a block rejected for a missing redeemer reaches the deterministic path rather than the rewind path, failing without the change.
- Check the class rather than the instance: audit every rejection type reachable from the era rule sets against that function and report which are unclassified, rather than adding only this one.

Found while implementing gouroboros #2250 (PR #2259); adjacent to #3889 and #3999.

wolf31o2:
Fixes #4000.

Classify `lcommon.MissingRedeemerForScriptError` as deterministic for every redeemer purpose so replay recovery does not rewind and refetch the same rejected block indefinitely.

At gouroboros v0.205.4 the common error is emitted by four paths:

- `script.ValidateRequiredRedeemers`
- `common.ValidateScriptWitnesses`
- `dijkstra.validateDijkstraPlutusRedeemers`
- Dingo's `validateConwayRequiredPlutusRedeemers`

Each path either fails or skips unresolved inputs and indexes spending purposes from the transaction's own sorted input list, making an observed missing-redeemer verdict monotone under additional input resolution. `ExtraRedeemerError` remains state-dependent because an unresolved Dijkstra input drops the spend purpose from the required set, so it is left unclassified and keeps taking the producer-resolution rewind.

Regression coverage drives all seven redeemer tags through replay recovery, asserts the extra-redeemer verdict stays unclassified, and keeps `BadInputsUtxoError` on the producer-resolution rewind path as a control.


ledger/replay_recovery.go
```diff
@@ -389,13 +389,29 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+// conway and babbage names are aliases of lcommon.MissingRedeemerForScriptError
+// rather than distinct types, so matching the common type covers all of them.
 func isDeterministicTxValidationError(err error) bool {
 	if _, ok := errors.AsType[shelley.DuplicateInputError](err); ok {
 		return true
 	}
 	if _, ok := errors.AsType[conway.PlutusScriptFailedError](err); ok {
 		return true
 	}
+	if _, ok := errors.AsType[lcommon.MissingRedeemerForScriptError](err); ok {
```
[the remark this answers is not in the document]
wolf31o2: The spend purpose is unrepairable by replay for the same reason as the other five, though by a longer argument than the comment carried.

A UTxO is addressed by producing transaction hash and output index (`LedgerState.UtxoByRef`, `ledger/state.go:9913`). An input that resolves at all resolves to exactly the output its producer wrote, script address included, so replay cannot vary whether a spend redeemer is required — only whether the output is present at all. Absence is itself a rejection: `UtxoValidateBadInputsUtxo` is registered ahead of `UtxoValidateRequiredRedeemers` and `UtxoValidateScriptWitnesses` in every era that has them (Babbage 19/20/21, Conway 31/32/33, Dijkstra 31/32/33).

So a transaction spending a script-locked input without its spend redeemer is rejected under every local history: `MissingRedeemerForScriptError` where the input resolves, `BadInputsUtxoError` where it does not. The property the classification needs is that no history accepts the block, not that the same rule reports it each time. Splitting on `RedeemerTagSpend` would leave that case on the rewind path, which is the wedge in #4000.

Addressed in 54de8ef6: the argument is now recorded on `isDeterministicTxValidationError`, and `TestReplayRecoveryRejectsDeterministicMissingRedeemer` covers `RedeemerTagSpend` alongside `RedeemerTagMint`. Both subtests fail without the classification.

ledger/replay_recovery.go
```diff
@@ -389,13 +389,40 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
 // a wire-level duplicate cardano-node coalesces and this verdict rejects
 // (preview slot 1462320; blinklabs-io/gouroboros#1989). Recovery must stay
 // non-terminal for that duplicate verdict for exactly that reason.
+//
+// A missing redeemer is deterministic for transaction-owned script purposes,
+// but not for spending inputs. Rebuilding the local UTxO window can change
+// whether a spending input resolves to a script output and therefore whether
+// the spend redeemer rule applies. Keep spend errors on the state-dependent
+// rewind path; the other purposes do not depend on local UTxO state.
```
chrisguiney, line 397: The stated basis for the split does not hold in either direction, so the classification is not derivable from this comment.

`the other purposes do not depend on local UTxO state` — they do. The Plutus script that makes a mint/cert/reward/vote/propose redeemer required can be a reference script resolved from the local UTxO. `collectTransactionScriptRequirements` adds `utxo.Output.ScriptRef()` from resolved regular *and* reference inputs to `available` (gouroboros v0.202.10 `ledger/common/rules.go`), and `collectConwayAvailableScripts` does the same for Dingo's own rule (`ledger/eras/conway.go:1022`). Under the criterion as written, no purpose qualifies and the switch should be empty.

`Rebuilding the local UTxO window can change whether a spending input resolves to a script output` — it can change whether it resolves, not what it resolves to. A UTxO is addressed by producing transaction hash and output index, so a resolvable input resolves to exactly the output its producer wrote, script bit included; the only variation another history produces is absence, and absence is itself a rejection. `UtxoValidateBadInputsUtxo` is registered ahead of `UtxoValidateScriptWitnesses` and `UtxoValidateRequiredRedeemers` in Conway (`ledger/conway/rules.go`), and Dingo's own path fails earlier still with `InputResolutionError` from `resolveConwayScriptInputs` (`ledger/eras/conway.go:985`). The reported index is the `script.SortInputs(tx.Inputs())` position (`ledger/eras/conway.go:496`), so it is transaction-determined too. That is the argument `54de8ef6` recorded here and that you posted in reply to cubic above; `1a42a231` removed it without saying what was wrong with it.

This has a concrete consequence. The PR description states that `UtxoValidateRequiredRedeemers` reports the spend purpose only at the current pin. So for the upstream rule the fix is inert at v0.202.10: the one tag it can report is routed straight back onto the rewind path that #4000 describes as wedging. Non-spend is reachable only through `common.ValidateScriptWitnesses` and `validateConwayRequiredPlutusRedeemers`.

I am not asking for the class to be widened blind — getting it wrong in that direction makes a rejection terminal and forks the node, so erring toward the rewind path is defensible on its own. The problem is that the code, this comment, the PR description and the standing reply to cubic currently disagree with each other. Pick one and make all four agree. If spend stays excluded, state here what stops the rewind loop for the only purpose the upstream rule can report today.

ledger/replay_recovery.go
```diff
@@ -389,13 +389,40 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+// distinct types, so matching the common type covers all of them.
 func isDeterministicTxValidationError(err error) bool {
 	if _, ok := errors.AsType[shelley.DuplicateInputError](err); ok {
 		return true
 	}
 	if _, ok := errors.AsType[conway.PlutusScriptFailedError](err); ok {
 		return true
 	}
+	if missing, ok := errors.AsType[lcommon.MissingRedeemerForScriptError](err); ok {
+		switch missing.Tag {
```
chrisguiney, line 414: `lint` fails on this line at this head: `missing cases in switch of type common.RedeemerTag: common.RedeemerTagSpend (exhaustive)`.

`exhaustive` is enabled in `.golangci.yml` with default settings, so `default-signifies-exhaustive` is false and the `default:` clause does not satisfy it. Reproduced locally with an isolated `GOLANGCI_LINT_CACHE`: `golangci-lint run ./ledger/` gives 1 issue at `e5697f74` and 0 at `eabf653d`, so it is caused by this diff rather than by the base branch.

Giving `lcommon.RedeemerTagSpend` its own explicit case also documents the carve-out where a reader will look for it.

ledger/replay_recovery.go
```diff
@@ -389,13 +389,40 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+		case lcommon.RedeemerTagMint,
+			lcommon.RedeemerTagCert,
+			lcommon.RedeemerTagReward,
+			lcommon.RedeemerTagVoting,
+			lcommon.RedeemerTagProposing,
+			lcommon.RedeemerTagGuarding:
+			return true
+		default:
+			return false
```
chrisguiney: Blocker: this `return false` is not equivalent to falling through, and it changes an existing classification.

Every other branch in this function only ever returns true, so an unmatched type continues to the next predicate. This branch returns false, which skips both `isRewardWithdrawalMismatch` and the Byron duplicate check for any error chain containing a spend-tagged missing redeemer.

That shape is reachable. `ValidateTxConway` runs every rule in `conwayValidationRules(ls)` and returns `errors.Join(errs...)` rather than stopping at the first failure (`ledger/eras/conway.go:200-224`); `UtxoValidateScriptWitnesses` and `UtxoValidateWithdrawals` are both in the Conway rule list; and `errors.AsType` traverses a `Join`.

Measured at this head on `errors.Join(MissingRedeemerForScriptError{Tag: RedeemerTagSpend}, shelley.IncorrectWithdrawalAmountError{})`, each wrapped the way `ValidateTxConway` wraps them:

- with this block present: `isDeterministicTxValidationError` returns **false**
- with this block removed in place (main's behaviour): returns **true**

`isRewardWithdrawalMismatch` returns true for that chain in both cases; it is simply never reached now. So a withdrawal mismatch that main sends to the clean deterministic stop — the #3628 Preprod shape the comment above this function describes — takes the rewind path after this change.

Returning true for the classified tags and letting every other tag fall through to `isRewardWithdrawalMismatch`, rather than returning false from inside the type match, avoids this while keeping whatever spend disposition you settle on.

ledger/replay_recovery_test.go
```diff
@@ -2530,3 +2530,157 @@ func TestResolveReplayRecoveryProducerReportsPresentInput(t *testing.T) {
...
+func TestReplayRecoveryRejectsDeterministicMissingRedeemer(t *testing.T) {
+	for _, tc := range []struct {
+		name          string
+		tag           lcommon.RedeemerTag
+		deterministic bool
+	}{
+		{
+			name:          "spend purpose",
+			tag:           lcommon.RedeemerTagSpend,
+			deterministic: false,
```
chrisguiney, line 2546: Only `mint purpose` is a regression test here. Reverting the classification block in place at this head and re-running:

- `TestReplayRecoveryRejectsDeterministicMissingRedeemer/mint_purpose` — FAIL (armed continuation audit, tip moved)
- `TestReplayRecoveryRejectsDeterministicMissingRedeemer/spend_purpose` — PASS
- `TestReplayRecoveryKeepsBadInputsOnTheRewindPath` — PASS

The spend subtest and the bad-inputs control pass with and without the fix, because main already routes both to the rewind path. They are worth keeping as guards; they are just not the regression coverage the PR description presents them as.

Separately, `mint` is the only one of the six classified tags under test. Cert, reward, voting, proposing and guarding are asserted by nothing, so dropping any of those five `case` labels would leave the suite green.

chrisguiney (changes requested):
Requesting changes. Reviewed at head `e5697f74` in an isolated worktree; merge-base equals `origin/main` (`eabf653d`), so the merged tree is the head tree.

**Pipeline is red, and this change caused it.** `lint` fails with `ledger/replay_recovery.go:414:3: missing cases in switch of type common.RedeemerTag: common.RedeemerTagSpend (exhaustive)`. Reproduced locally with an isolated `GOLANGCI_LINT_CACHE`: `golangci-lint run ./ledger/` reports 1 issue at `e5697f74` and 0 issues at `eabf653d`. `exhaustive` is enabled in `.golangci.yml` with default settings, so the `default:` clause does not satisfy it. This is not base-branch noise.

Two blockers, both inline:

1. The `default: return false` is not equivalent to falling through. It shadows `isRewardWithdrawalMismatch` for any joined error chain containing a spend-tagged missing redeemer, and `ValidateTxConway` produces exactly such chains. Measured: a joined `MissingRedeemerForScriptError{Tag: Spend}` + `shelley.IncorrectWithdrawalAmountError` classifies deterministic on main and non-deterministic at this head.
2. The stated basis for the spend/non-spend split does not hold in either direction, and it contradicts the PR description, the argument recorded in `54de8ef6`, and your standing reply to cubic above.

Non-blocking: the PR description still describes the pre-carve-out behaviour and cites `gouroboros v0.202.9` where `go.mod` pins `v0.202.10`; and of the three new assertions only the `mint purpose` subtest fails without the fix.

Checks run at this head: `go build ./...` (0), `go vet ./...` (0), `go test ./ledger/ -run 'TestReplayRecoveryRejectsDeterministicMissingRedeemer|TestReplayRecoveryKeepsBadInputsOnTheRewindPath' -count=1` (0), `golangci-lint run ./ledger/` (1 issue, above). `gofmt -l .` flags `bark/blob_test.go` and `ledger/read_chain_pipeline_test.go` on `origin/main` as well, so that is pre-existing.

No bot has reviewed this head. cubic's check says it skipped the AI review because the head is a merge commit, and its two reviews are stamped `7d06c1ea` and `1a42a231`; CodeRabbit reports auto incremental reviews are disabled. Sibling #3990 is disjoint (three test files) and this change switches on named `RedeemerTag` constants rather than rule or purpose positions, so the index fragility #3990 removes does not apply here and the two compose.

ledger/replay_recovery.go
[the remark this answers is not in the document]
chrisguiney: Partly valid, but not for the reason given, and the head now implements it while the PR still argues against it.

The reason as stated does not hold: rebuilding the local UTxO window can change whether a spending input resolves, not what it resolves to. A UTxO is addressed by producing transaction hash and output index, so a resolvable input yields exactly the output its producer wrote, script bit included, and the only variation another local history produces is absence — which `UtxoValidateBadInputsUtxo`, registered ahead of the redeemer rules in Conway, rejects on its own, and which Dingo's own path rejects earlier still as `InputResolutionError` from `resolveConwayScriptInputs`. The reported index is the `script.SortInputs(tx.Inputs())` position, so it is transaction-determined as well. On that criterion the spend verdict is as unrepairable by replay as the other five.

What is valid is the caution: classifying a purpose deterministic that a replay could repair makes a rejection terminal, which is the fork-shaped failure, so the conservative direction is defensible. But the split as landed in 1a42a231 is not derivable from the criterion its own comment states — that comment claims the other purposes read no local UTxO state, and they do, via reference scripts collected from resolved inputs in `collectTransactionScriptRequirements` and `collectConwayAvailableScripts`.

Requesting changes for coherence rather than accepting or rejecting this finding outright, and separately for a regression the fix as landed introduces: the `default: return false` shadows `isRewardWithdrawalMismatch` for joined error chains, which `ValidateTxConway` produces. Details in the review at https://github.com/blinklabs-io/dingo/pull/4012#pullrequestreview-5127616165

consensus/leaderthreshold/threshold.go
```diff
@@ -23,124 +23,22 @@ import (
...
+// Keep this wrapper as the Dingo-owned boundary so callers do not duplicate
+// the protocol calculation.
 func Threshold(
 	poolStake uint64,
 	totalStake uint64,
 	activeSlotCoeff *big.Rat,
 	mode consensus.ConsensusMode,
 ) (*big.Int, error) {
-	threshold, err := consensus.CertifiedNatThresholdWithMode(
+	return consensus.CertifiedNatThresholdWithMode(
```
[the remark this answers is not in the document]
chrisguiney: Already addressed, and no longer present in the diff. `git diff eabf653d..e5697f74 -- consensus/leaderthreshold/threshold.go` is empty: the merge at e5697f74 resolved that file to `origin/main`, which had already landed the equivalent simplification in #4023. The head of this PR carries no change to `consensus/leaderthreshold` at all, so the fractional-cutoff correction this comment refers to is settled on main rather than by this branch. Verified at head e5697f74.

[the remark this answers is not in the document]
arepala-uml: Independently verified at current head `317d04e7`: `git diff origin/main -- consensus/leaderthreshold/threshold.go` is empty, confirming this file carries no change versus current `origin/main` (`5499ef7d`; the PR's own merge-base `eabf653d` has since drifted forward). No action needed on this PR. Resolving.

ledger/replay_recovery.go
arepala-uml: Fixed in 317d04e7 (`fix(ledger): preserve recovery error classification`): the `default: return false` case was removed and `RedeemerTagSpend` now falls through as an empty case, reaching `isRewardWithdrawalMismatch` for joined error chains as intended.

Reproduced the regression check independently in an isolated worktree: `errors.Join(common.MissingRedeemerForScriptError{Tag: common.RedeemerTagSpend}, shelley.IncorrectWithdrawalAmountError{})` passed to `isDeterministicTxValidationError` now returns true (matches main's behavior), where it returned false before this commit. Resolving.

ledger/replay_recovery.go
arepala-uml: Fixed in the same commit (317d04e7): `RedeemerTagSpend` now has its own explicit (empty) case, satisfying `exhaustive`. Reproduced at current head in an isolated worktree with `GOLANGCI_LINT_CACHE` set: `golangci-lint run ./ledger/...` reports 0 issues. Resolving.

ledger/replay_recovery.go
arepala-uml: Independently checked this against the current head (317d04e7) and against `ledger/eras/conway.go`. The premise is still wrong as written, and it's still unfixed: the comment at replay_recovery.go:393-397 is unchanged since e5697f74 and still says the five non-spend purposes 'do not depend on local UTxO state,' which `collectConwayAvailableScripts` (conway.go:1022) refutes exactly as you showed — it pulls available scripts from resolved reference inputs, i.e. local UTxO content.

What I can confirm does hold is a version of your closed-system argument, and it comes out symmetric with spend rather than distinguishing it: `newConwayPlutusValidationContext` calls `resolveConwayScriptInputs` (conway.go:965) *before* `validateConwayRequiredPlutusRedeemers` runs (conway.go:226-238), and a reference input that fails to resolve there returns `ReferenceInputResolutionError`, which short-circuits the function at conway.go:229 before the redeemer check ever runs. `ReferenceInputResolutionError` is itself already conservatively classified non-deterministic elsewhere in this file. So `MissingRedeemerForScriptError`, for any of the six purposes including spend, is only reachable once every reference and regular input has already resolved to the exact bytes its producer wrote (content-addressed, hence invariant under replay). That argument, if it's the one intended, applies identically to spend.

So the split still isn't derivable from anything stated in the code, the PR description, or the standing reply to cubic above, and 317d04e7 only touched the switch's control flow, not this comment or the split itself. This is the substantive item left open in my review: state the argument that's actually true here, and either extend it to spend or say explicitly why spend is different despite it. Leaving this thread open.

ledger/replay_recovery_test.go
arepala-uml: Still valid at current head (317d04e7): the table at replay_recovery_test.go:2537 still has only `spend purpose` and `mint purpose`. Reproduced the gap mechanically in an isolated worktree: flipped `RedeemerTagCert` from the deterministic branch to `return false` (a real behavior change) and re-ran `TestReplayRecoveryRejectsDeterministicMissingRedeemer` and `TestReplayRecoveryKeepsBadInputsOnTheRewindPath` — both still pass. `cert`, `reward`, `voting`, `proposing`, and `guarding` remain asserted by nothing; any of those five case labels could be dropped or flipped without the suite noticing. Leaving this open as part of the change request.

arepala-uml (changes requested):
Requesting changes. Reviewed at head `317d04e7` in an isolated worktree (`/tmp/review-dingo-4012`, created from `origin/main` + the PR ref, not the shared checkout). `origin/main` has moved to `5499ef7d` since this branch's merge base (`eabf653d`); the merged tree matches the head tree for the two changed files (`git diff origin/main...HEAD` touches only `ledger/replay_recovery.go` and `ledger/replay_recovery_test.go`).

**Both of chrisguiney's blockers from the review at `e5697f74` are fixed in 317d04e7** ("preserve recovery error classification"), independently reproduced:
- `default: return false` → replaced with an explicit empty `RedeemerTagSpend` case that falls through to `isRewardWithdrawalMismatch`. Probed `errors.Join(MissingRedeemerForScriptError{Tag: Spend}, shelley.IncorrectWithdrawalAmountError{})`: now classifies deterministic (matches main), was non-deterministic before this commit.
- `golangci-lint run ./ledger/...` (isolated `GOLANGCI_LINT_CACHE`): 0 issues. The exhaustive-switch failure is gone.
- Replied and resolved both threads, plus cubic's stale `threshold.go` thread (confirmed `git diff origin/main -- consensus/leaderthreshold/threshold.go` is empty).

**Two of chrisguiney's items are not addressed at this head, and I independently confirm both are still real** (replied inline, left open):

1. `ledger/replay_recovery.go:393-397` — the doc comment's stated reason for excluding spend ("the other purposes do not depend on local UTxO state") is still there, still contradicted by `collectConwayAvailableScripts` (`ledger/eras/conway.go:1022`), which builds the available-script set from resolved reference inputs. Tracing the actual call order in `ledger/eras/conway.go:226-238` and `resolveConwayScriptInputs` (`conway.go:965-996`): a reference input that fails to resolve returns before the redeemer check runs at all, as `ReferenceInputResolutionError` — already classified non-deterministic elsewhere. So `MissingRedeemerForScriptError` for *any* of the six purposes, spend included, is only reachable once inputs have resolved to content-addressed, replay-invariant bytes. That argument is symmetric across all six tags; it doesn't derive the spend/non-spend split as coded. Code, PR description, and the standing reply to cubic still disagree with each other on why spend is different.
2. `ledger/replay_recovery_test.go:2537` — of the six classified tags, only `mint` and `spend` have subtests. Reproduced mechanically: flipping `RedeemerTagCert` to `return false` (dropping it from the deterministic set) leaves both classification tests green. `cert`, `reward`, `voting`, `proposing`, `guarding` are unguarded.

Given `isDeterministicTxValidationError` gates whether replay recovery treats a rejection as terminal — misclassifying the wrong direction here doesn't corrupt consensus by itself, but it can turn a block that other honest nodes (with a complete local UTxO window) would accept into one this node refuses to ever retry, which is the failure mode #4000/#3889 exist to fix. That risk profile is exactly why the stated criterion needs to be the true one and each classified tag needs its own regression test, the way `mint` and `spend` already have.

**CI**, current head `317d04e7`:
| Check | Result |
| --- | --- |
| `go build ./...` (isolated worktree) | 0 |
| `go vet ./...` (isolated worktree) | 0 |
| `gofmt -l .` (isolated worktree) | clean |
| `golangci-lint run ./ledger/...` (isolated cache) | 0 issues |
| `go test ./ledger/ -run 'TestReplayRecoveryRejectsDeterministicMissingRedeemer\|TestReplayRecoveryKeepsBadInputsOnTheRewindPath'` | PASS |
| CI: lint, go-test (Linux/Linux-race/macOS), docker, CodeQL, govulncheck, Conventional Commits, DCO | SUCCESS |
| CI: go-test (Windows) | CANCELLED — hit the job's own 90-minute timeout mid-`go-test` (started 17:50Z, killed 19:25Z); this job is explicitly advisory/non-required per its own workflow comments. Doesn't reproduce on the last two `main` pushes (both Windows-success), and the diff is platform-agnostic, so I'm reading this as runner variance, not something this change caused. Non-blocking either way. |

No bot has reviewed this head: cubic's last real review is stamped `1a42a231`, two commits back; CodeRabbit's check reads "Review skipped: incremental reviews are disabled." Both cubic threads are resolved (one was already ✅, I resolved the other after confirming it's stale).

Approval is withheld on the two open items above, not on CI (pipeline is green modulo the non-required, non-reproducing Windows timeout).

ledger/replay_recovery.go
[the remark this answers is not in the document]
wolf31o2: Not applied, for a reason the rules support rather than the one given. All three paths that report MissingRedeemerForScriptError reach the redeemer check only on inputs that already resolved: script.ValidateRequiredRedeemers builds a script.TxScriptView first and ResolveTxInputs stops at the first unresolved input; common.ValidateScriptWitnesses skips an unresolved regular input rather than requiring a redeemer for it; validateConwayRequiredPlutusRedeemers reads resolveConwayScriptInputs, which fails with InputResolutionError on the first unresolved input. A resolved input is addressed by producing transaction hash and output index, so replay can only add resolutions, and each rule indexes redeemers by position in the transaction's own sorted input list rather than among the resolved subset. The verdict is monotone under resolution, so spend is classified with the other six at head 3cc2cfbb.

consensus/leaderthreshold/threshold.go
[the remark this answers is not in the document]
wolf31o2: Still stale at head accdaeb2: `git diff origin/main -- consensus/leaderthreshold/threshold.go` is empty, so this branch carries no change to that file and the fractional-cutoff correction is settled on main.

ledger/replay_recovery.go
wolf31o2: Fixed in 854bb7ee, with the tag coverage corrected in 3cc2cfbb. The carve-out is gone and the comment now states the derivation that holds. None of the three reporting rules reaches its redeemer check on an unresolved input: ResolveTxInputs stops at the first failure, ValidateScriptWitnesses skips an unresolved regular input, and resolveConwayScriptInputs returns InputResolutionError. A resolved input is addressed by producing transaction hash and output index, and all three rules index redeemers by position in the transaction's own sorted input list, so a missing-redeemer verdict is monotone under resolution. That argument is symmetric across the tags, so spend is classified rather than carved out. Correction to the earlier reply and to the PR description: the spend-only claim for UtxoValidateRequiredRedeemers held at gouroboros v0.202.x, but go.mod now pins v0.204.4, where that rule derives purposes from script.ScriptPurposes and reports every tag.

ledger/replay_recovery.go
wolf31o2: Fixed at 317d04e7 and still clean at accdaeb2: `golangci-lint run ./ledger/...` reports 0 issues. The switch now lists all seven tags in one case with no default clause, so a redeemer tag added upstream trips exhaustive here and gets its own classification decision.

ledger/replay_recovery.go
wolf31o2: Fixed at 317d04e7 by removing the default clause. accdaeb2 adds TestIsDeterministicMissingRedeemerAcrossJoinedErrors, which asserts that errors.Join of a spend-tagged MissingRedeemerForScriptError and shelley.IncorrectWithdrawalAmountError classifies deterministic, with lcommon.InputResolutionError as the negative control. The shadowing cannot return unasserted.

ledger/replay_recovery_test.go
wolf31o2: Fixed in 854bb7ee. The table now carries a subtest per tag - spend, mint, cert, reward, voting, proposing, guarding - and every one is load-bearing: removing a single case label from the classification switch and re-running that tag's subtest fails with exit 1, in all seven runs. The spend subtest now fails without the fix as well, since spend is classified rather than routed back to the rewind path.

ledger/replay_recovery.go
```diff
@@ -405,19 +405,67 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
-// non-terminal for that duplicate verdict for exactly that reason, and for
-// the same reason a false-positive malformed-script verdict must also stay
-// non-terminal: recoverFromDeterministicTxValidationError rewinds and asks
-// chain selection for another candidate rather than halting, so a locally
-// mistaken rejection still leaves the node able to follow a chain a peer
-// later offers.
+// non-terminal for that duplicate verdict for exactly that reason.
+//
+// A missing redeemer is deterministic for every script purpose, spending
+// included. Three rules report it, all as the one common type; the conway and
```
chrisguiney, line 411: Four emitters at the v0.204.4 pin, not three: gdijkstra validateDijkstraPlutusRedeemers (dijkstra/rules.go:1290) returns the same type, reached from UtxoValidateRedeemerAndScriptWitnesses in the Dijkstra rule list and from UtxoValidatePlutusScripts, which ValidateTxDijkstra calls (ledger/eras/dijkstra.go:300). It is monotone as well, so the classification stands, but this comment is the audit record and should name it. ValidateRequiredRedeemers also skips, rather than fails on, unresolved sub-transaction inputs (resolveBodyInputs).

ledger/replay_recovery_test.go
```diff
@@ -2688,3 +2688,200 @@ func TestResolveReplayRecoveryProducerReportsPresentInput(t *testing.T) {
...
+		{name: "guarding purpose", tag: lcommon.RedeemerTagGuarding},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			requireMissingRedeemerRecovery(t, tc.tag, true)
+		})
+	}
+}
+
+// The spend purpose is the one the upstream UtxoValidateRequiredRedeemers rule
```
chrisguiney, line 2717: Stale at the pin: UtxoValidateRequiredRedeemers reports every tag at v0.204.4, as the production comment and 3cc2cfbb state. This sentence still carries the v0.202.x spend-only claim; drop it or restate spend as one of the seven tags the rule reports.

ledger/replay_recovery_test.go
```diff
@@ -2688,3 +2688,200 @@ func TestResolveReplayRecoveryProducerReportsPresentInput(t *testing.T) {
...
+			fmt.Errorf("conway utxow rule: %w", missingSpend),
+			fmt.Errorf(
+				"conway utxo rule: %w",
+				shelley.IncorrectWithdrawalAmountError{},
+			),
+		)),
+		"a joined chain carrying a withdrawal mismatch stays deterministic",
+	)
+	// The control: an unresolved input is state-dependent even when it is
+	// joined with a deterministic verdict of its own class.
```
chrisguiney, line 2752: The comment describes a joined control, but the assertion is a bare InputResolutionError{}. Joined with any classified verdict the chain classifies deterministic, since errors.AsType walks the join: errors.Join of InputResolutionError and a mint-tagged MissingRedeemerForScriptError returns true. The monotone argument makes that correct, so the comment states behaviour the function does not have. Reword it to describe the bare control.

chrisguiney (changes requested):
Re-review at b6ee10d6. Prior blockers cleared, with arepala-uml's matching items: ledger lint is clean and exhaustive fires on a removed tag; no false return remains; spend is classified on a monotone argument that holds for every emitter at the pin; removing any one tag label fails only its subtest. Three comment fixes inline. PR body: ExtraRedeemerError is state-dependent, not transaction-alone, because gdijkstra UtxoValidateRedeemerAndScriptWitnesses skips unresolved consumed inputs and so reads a valid spend redeemer as extra.

ledger/replay_recovery.go
wolf31o2: Correct; applied in ca2c41a5. The comment now names four emitters at the v0.204.4 pin, including dijkstra.validateDijkstraPlutusRedeemers, reached from both dijkstra.UtxoValidateRedeemerAndScriptWitnesses and dijkstra.UtxoValidatePlutusScripts. Its two paths stay monotone: dijkstraWitnessRuleLevels skips an unresolved consumed input, dijkstraScriptLevels fails through ResolveTxInputs, and both index through script.ScriptPurposes, which numbers spend purposes by SortInputs(body.Inputs()). The ValidateRequiredRedeemers bullet now records resolveBodyInputs skipping unresolved sub-transaction inputs.

ledger/replay_recovery_test.go
wolf31o2: Correct; applied in ca2c41a5. The sentence now restates spend as one of the seven tags UtxoValidateRequiredRedeemers reports at the v0.204.4 pin, rather than the tag it alone can report.

ledger/replay_recovery_test.go
wolf31o2: Correct; applied in ca2c41a5. The comment now scopes the control to the bare error and states that errors.AsType walks a join, so an InputResolutionError joined with a classified verdict classifies deterministic. That case is now asserted rather than described: errors.Join of InputResolutionError and a mint-tagged MissingRedeemerForScriptError is required to return true.

ledger/replay_recovery.go
```diff
@@ -405,19 +405,81 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
-// non-terminal for that duplicate verdict for exactly that reason, and for
-// the same reason a false-positive malformed-script verdict must also stay
-// non-terminal: recoverFromDeterministicTxValidationError rewinds and asks
-// chain selection for another candidate rather than halting, so a locally
-// mistaken rejection still leaves the node able to follow a chain a peer
-// later offers.
+// non-terminal for that duplicate verdict for exactly that reason.
+//
+// A missing redeemer is deterministic for every script purpose, spending
+// included. Four rules report it at the gouroboros v0.204.4 pin, all as the
```
chrisguiney, line 411: The PR description still reports three paths at the v0.204.4 pin and omits dijkstra.validateDijkstraPlutusRedeemers, so the two copies of this audit record disagree about the emitter set. At the pin, MissingRedeemerForScriptError is constructed at ledger/common/rules.go:1111, ledger/common/script/redeemers.go:212 and ledger/dijkstra/rules.go:1290, plus Dingo's ledger/eras/conway.go:715 - four, as this line says. Bring the description's list to the same four.

ledger/replay_recovery.go
```diff
@@ -405,19 +405,81 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+//     instead of a verdict. Its Dijkstra sub-transaction levels resolve
+//     through resolveBodyInputs instead, which skips what it cannot resolve
+//     and leaves the failure to UtxoValidateBadInputsUtxo. At v0.204.4 it
+//     derives purposes from script.ScriptPurposes and reports every tag, not
+//     spend alone.
+//   - common.ValidateScriptWitnesses, behind UtxoValidateScriptWitnesses,
+//     skips an unresolved regular input rather than failing, so an incomplete
+//     UTxO window can only withhold a spend requirement, never invent one; an
+//     unresolved reference input is ReferenceInputResolutionError.
+//   - dijkstra.validateDijkstraPlutusRedeemers, reached from both
```
chrisguiney, line 432: This rule also makes the description's classification audit wrong in the terminal direction. ExtraRedeemerError is listed there as decided by the transaction alone, but dijkstra/rules.go:1281 raises it from the required set that dijkstraRequiredPlutusPurposes builds via ScriptPurposes over level.view.ResolvedInputs. On the UtxoValidateRedeemerAndScriptWitnesses path dijkstraWitnessRuleLevels skips an unresolved consumed input, so that spend purpose leaves required and a legitimate spend redeemer reads as extra; resolving the input removes the verdict. It is state-dependent.

chrisguiney (changes requested):
Re-review at ca2c41a5. The three comment fixes are applied and correct at the v0.204.4 pin: the four-emitter enumeration, the skip/fail and SortInputs indexing claims, and the bare-control wording all check out against gouroboros source. The new joined assertion is causal - inserting an early false return on InputResolutionError fails only replay_recovery_test.go:2762. Head is green, and the tree merged with current main builds, lints and passes -race. Two copies of this change's audit record still disagree; both inline.

ledger/replay_recovery.go
wolf31o2: Fixed in the current PR description: it names script.ValidateRequiredRedeemers, common.ValidateScriptWitnesses, dijkstra.validateDijkstraPlutusRedeemers, and Dingo's validateConwayRequiredPlutusRedeemers.

ledger/replay_recovery.go
wolf31o2: Fixed in the current PR description: ExtraRedeemerError is listed in the state-dependent group because unresolved Dijkstra inputs can change the required-purpose set.

ledger/replay_recovery.go
```diff
@@ -405,19 +405,87 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
+// therefore only add resolutions, and all four rules index redeemers by
+// position in the transaction's own sorted input list rather than by position
+// among the resolved subset, Dijkstra sub-transaction levels included. The
+// verdict is monotone under resolution: no local history removes a
+// missing-redeemer rejection, so no replay repairs one.
+//
+// The genuinely state-dependent cases carry their own types --
+// InputResolutionError, ReferenceInputResolutionError, and
+// shelley.BadInputsUtxoError -- which this function does not classify, so they
+// keep taking the producer-resolution rewind. conway.ExtraRedeemerError is
```
chrisguiney, line 515: conway.ExtraRedeemerError has two emitters with opposite determinism at this pin, and only the Dijkstra one is state-dependent. conway.UtxoValidateExtraneousRedeemers (conway/rules.go:1782) calls common.ValidateExtraneousRedeemers (common/rules.go:1211), which takes no LedgerState and decides on len(tx.Inputs()) plus the cert, withdrawal, proposal, policy and voter counts, then returns ExtraRedeemerError at rules.go:1791. Dingo runs it: buildConwayValidationRules skips six rule ids and this is not one. Transaction-alone and unclassified is the #4000 wedge.

ledger/replay_recovery_test.go
```diff
@@ -2695,3 +2695,231 @@ func TestResolveReplayRecoveryProducerReportsPresentInput(t *testing.T) {
...
+		"an unresolved input is decided by local UTxO state",
+	)
+	// The inverse verdict from the same Dijkstra rule. An unresolved consumed
+	// input is skipped rather than resolved, so its spend purpose never
+	// reaches the required set and a legitimate spend redeemer reads as extra;
+	// resolving the input removes the verdict, so widening the classification
+	// to cover it would strand the rewind that repairs it.
+	require.False(
+		t,
+		isDeterministicTxValidationError(conway.ExtraRedeemerError{}),
```
chrisguiney, line 3390: This message pins a basis that does not hold in both directions: a bare conway.ExtraRedeemerError carries no emitter, and the Conway one is decided by the transaction alone. Leaving the type unclassified may still be correct, since both emitters share one Go type and errors.AsType cannot separate them, but record that as the reason. Babbage and Alonzo UtxoValidateExtraneousRedeemers return common.ExtraneousRedeemerError from the same bounds check and go unnamed. This assertion is a guard: 6848e07f changes no behavior, so nothing reverts to fail it.

ledger/replay_recovery.go
```diff
@@ -405,19 +405,87 @@ func (ls *LedgerState) tryRecoverFromTxValidationError(
...
 		return true
 	}
 	if _, ok := errors.AsType[conway.PlutusScriptFailedError](err); ok {
 		return true
 	}
+	if missing, ok := errors.AsType[lcommon.MissingRedeemerForScriptError](err); ok {
+		// Enumerated rather than matched on the type alone so that a redeemer
+		// tag added upstream trips the exhaustive linter here and gets its own
+		// classification decision instead of inheriting this one.
+		switch missing.Tag {
```
chrisguiney, line 545: ARCHITECTURE.md is a third copy of this audit record and still disagrees. Its replay-recovery section names only the duplicate-input pair and the two Malformed types, then states that other transaction-validation errors continue through producer resolution and the unresolved-producer fallback. This change adds seven redeemer tags to the classified set, so that sentence is now false, and the file carries no MissingRedeemer mention on the tree merged with current main.

chrisguiney (changes requested):
Re-review at 6848e07f. The prior blocker is cleared: the comment and the description now name the same four emitters at the v0.205.4 pin and both place ExtraRedeemerError among the state-dependent verdicts, and the byte-identity claim in 6848e07f holds against both module versions. The two items arepala-uml left open are addressed at this head. Three items inline. No bot reviewed this head: CodeRabbit reports incremental reviews disabled with a declared range ending at 7d06c1ea, and cubic has no record, check or status.

ledger/replay_recovery.go
chrisguiney, line 515: Stale: this audit record says conway.ExtraRedeemerError is state-dependent, while replay_recovery_test.go:2769 now says the type has a transaction-only Conway emitter and stays unclassified only because the Go type cannot separate emitters. Every Conway-era emitter is transaction-decided: common.ValidateExtraneousRedeemers behind conway.UtxoValidateExtraneousRedeemers, and validateTxPlutusConwayWithContext (ledger/eras/conway.go:589-631), reading only inputs resolveConwayScriptInputs already resolved. State that reason here and name common.ExtraneousRedeemerError for Babbage and Alonzo.

ARCHITECTURE.md
```diff
@@ -4824,7 +4824,15 @@ both. Replay recovery therefore rejects the primary-chain branch
...
 `chainsync.resync` event with reason `deterministic tx validation recovery` so
 ChainSync obtains a fresh intersection. Other transaction-validation errors
-continue through producer resolution and the unresolved-producer fallback.
+continue through producer resolution and the unresolved-producer fallback,
+except for missing-redeemer errors. A missing redeemer is classified as
+deterministic for every `RedeemerTag` purpose (spend, mint, certificate,
+reward, voting, proposing, and guarding): each validator either resolves the
+referenced input or reports an input-resolution error before the redeemer
+verdict, so replaying a different local UTxO history cannot remove the missing
+redeemer. `ExtraRedeemerError` remains state-dependent because Dijkstra can
```
chrisguiney: The missing-redeemer text is correct, but this sentence repeats the stale basis from replay_recovery.go:457: the type is not state-dependent as such, and a Conway-era extra-redeemer rejection is decided by the transaction yet still takes the producer-resolution rewind. Record it as unclassified because the Go type is shared with the state-dependent Dijkstra emitter, and bring the PR description to the same wording. After merging main the pin is v0.205.7; the redeemer emitter files are unchanged from v0.205.4, so only the version label in replay_recovery.go:411 moves.

chrisguiney (changes requested):
Re-review at cc757fb3. ARCHITECTURE.md now documents the missing-redeemer classification and the test comment records the shared-type reason, clearing two of the three prior items. The ExtraRedeemerError item still stands in the production comment, ARCHITECTURE.md and the description; details inline. The tree merged with current main builds, lints clean and passes the redeemer and recovery tests under race, and removing the classification block fails all seven tag subtests and the joined test. No bot reviewed this head.

chrisguiney (approved):

Issue #3996, opened 2026-09-06
chainsync: a peer that is merely behind on our own chain is classified as an over-K fork and evicted, because the intersect ladder's sparse step lands past the security parameter


chadaibbhot-git:
## Summary

A peer whose chain is a strict **prefix** of ours — same chain, no divergence,
just behind — is rejected with `rollback exceeds security parameter K` and denied
for 2 minutes. On a node whose only upstream is that peer, this stalls chain
selection entirely and the node stops following the chain until the peer
overtakes it.

The trigger is not the size of the lag. It is that `Chain.IntersectPoints`
samples deeper history at doubling offsets (32, 64, 128, 256 …), so a lagging
peer's best common point is the *next power-of-two* offset below it, which can
sit past K even when the peer's actual lag is well inside K. With K=108 a peer
only **65** blocks behind is already evicted as a deep fork.

The node also escalates this to
`"local chain has diverged from the network and cannot self-recover. Operator
intervention required (e.g. re-bootstrap from a Mithril snapshot)"` — advice
that, followed, destroys a healthy database. Nothing had diverged.

## Observed

Private Leios devnet, 1 s slots, K=108. A block producer built from `main` at
`c159aa6d` plus local ledger patches; its single configured upstream is one
relay running the same codebase. The relay had frozen (separate defect) and was
119 blocks behind.

```
INFO  exiting Genesis selection mode local_slot=2630328 best_known_slot=2627424
INFO  registered peer from chainsync rollback  rollback_slot=2627304
        advertised_block=115076 advertised_slot=2627424
INFO  selected new best peer  block_number=115076 slot=2627424
WARN  rejecting rollback that exceeds security parameter K
        fork_depth=128 security_param=108 rollback_slot=2627304
ERROR chainsync rollback exceeds security parameter K, rejecting peer chain
        component=ledger slot=2627304 hash=b471979d4c679864ae890c7afb9722c21f846e4c4a4fad7d8d6814ba131f0532
WARN  temporarily denying chainsync peer whose chain we cannot follow
        reason="rollback exceeds security parameter K" duration=2m0s
INFO  closing connection for fresh chainsync
INFO  chain selection stalled: no selectable peer
```

This repeated 65 times over 90 minutes — reconnect, re-intersect, same
rejection, same 2-minute denial — with `chain selection stalled: no selectable
peer` after each one. The producer's tip did not move for 2 h 05 min and it
missed its scheduled leader slot at 2637334.

[...]

Issue #3035, opened 2026-07-31
ledger: fork_depth uint64 underflow when the rollback point is ahead of the persistent tip → node permanently rejects every peer ("fork resolution exceeds security parameter K")


chadaibbhot-git:
## Summary

When "at-tip recovery" holds the persistent ledger tip behind the primary chain (its designed behaviour on a validation divergence), a subsequent fork resolution can target a rollback point whose **block index is higher than the current tip's**. `forkDepth` is then computed as `tipBlockIndex - rollbackBlockIndex` on two `uint64`s with no guard, so it **underflows to ~2⁶⁴**, trivially exceeds the security parameter K, and the node rejects the rollback and **temporarily denies the peer** — for *every* peer, permanently. The node can never follow any chain again and requires operator intervention (fresh bootstrap).

Observed on preview, v0.68.0: the BP sat ~147k slots behind wall for ~90 min, denying its Haskell relay and its Dingo relay in a loop, while an independent Haskell `cardano-node` on the same network stayed at tip (so the chain was healthy — this is Dingo-side).

## The underflow (source)

`chain/chain.go` — two identical sites, lines ~737 and ~810:

```go
var rollbackBlockIndex uint64
if point.Slot > 0 {
    tmpBlock, err := c.manager.blockByPoint(point, nil)
    ...
    rollbackBlockIndex = tmpBlock.ID     // uint64
}
// Calculate fork depth before deleting blocks
forkDepth := c.tipBlockIndex - rollbackBlockIndex   // c.tipBlockIndex is uint64 (chain.go:56)
...
if c.persistent &&
    c.tipBlockIndex >= uint64(securityParam) &&
    forkDepth > uint64(securityParam) {             // underflowed forkDepth is trivially > K
    slog.Default().Warn("rejecting rollback that exceeds security parameter K",
        "fork_depth", forkDepth, "security_param", securityParam, "rollback_slot", point.Slot)
    return ErrRollbackExceedsSecurityParam
}
```

There is no check that `rollbackBlockIndex <= c.tipBlockIndex` before the subtraction. When the rollback point is ahead of the tip, `tip - rollback` wraps.

## Evidence

The reject log shows the wrapped value exactly:

```
level=WARN  msg="rejecting rollback that exceeds security parameter K"
            fork_depth=18446744073709551598 security_param=432 rollback_slot=118655549
level=ERROR msg="fork exceeds security parameter K, rejecting fork resolution"
[...]

PR #4013, opened 2026-09-06
fix(forging): bound the unknown-upstream-target gate by local tip lag

### LINKED ISSUE #4010: forging: the unknown-upstream-target gate is self-sealing on an all-producer network, so the chain stops and every node reports healthy
## Summary

`TestAcceleratedScenarioTimeline` fails intermittently on clean `origin/main`, with the network stopped and every node reporting healthy, connected and `lastErr=""`. The cause is a liveness deadlock in the upstream-sync forge gate, not a lock deadlock and not the test.

`ledger/forging/forger.go:823-826`:

```go
upstreamTip, upstreamActive := f.slotClock.UpstreamSyncStatus()
if upstreamActive && (upstreamTip == 0 ||
    (upstreamTip > tipSlot &&
        upstreamTip-tipSlot > f.forgeSyncToleranceSlots)) {
```

The `upstreamTip == 0` disjunct short-circuits before any tolerance is applied. `forgeSyncToleranceSlots` — the knob that expresses how far behind is acceptable before refusing to forge — is never consulted for the unknown-target case, so that case skips unconditionally however fresh the local tip is.

## Why it cannot recover

`LedgerState.UpstreamSyncStatus` (`ledger/state.go:9207-9221`) returns `(0, true)` for the whole window between an active-connection switch and the new peer's first authenticated, admitted, trusted header. `publishActiveUpstream` (`ledger/state.go:9233-9249`) stores the new connection key with `targetSlot` zero; only `publishAdmittedUpstreamTarget` (`ledger/state.go:9257-9283`) lifts it.

On an all-producer network that window is self-sealing:

- no node forges, because each is in the unknown-target window;
- therefore no new header is produced anywhere;
- therefore no header is admitted;
- therefore nothing lifts `targetSlot` off zero;
- therefore no node forges.

Forging is the only source of new headers on such a network, so the state that suppresses forging is the state that prevents its own exit. The nodes stay connected and healthy the whole time, which is why nothing surfaces as an error.

## Observed

In one failing run, the gate dropped **67 scheduled leader slots** across the three producers, every one at `upstream_tip=0`:

| node | leader slots dropped |
| --- | --- |
| dingo-1 | 22 |
| dingo-2 | 29 |
| dingo-3 | 16 |

```
level=WARN msg="chain syncing from peer, skipping forge" current_slot=124 tip_slot=121 upstream_tip=0 leader_slot=true
level=WARN msg="chain syncing from peer, skipping forge" current_slot=137 tip_slot=121 upstream_tip=0 leader_slot=true
```

The local tip lagged the current slot by 3 and 16 slots at those points, far inside `forgeSyncToleranceSlots` (100). The node had no evidence it was behind; it declined because the target was unknown.

Once wedged, the slot clock passes the era horizon, which produces the accompanying
`"skipping slot tick past era horizon"` lines. That is a consequence of the stall, not a cause.

## Not attributable to any branch

Runs of `./internal/test/devnet/run-tests.sh --accelerated` across three trees on one machine:

| tree | result |
| --- | --- |
| `9288bd78` | 1 pass / 2 fail |
| `origin/main` `6fc22fa4` | 0 pass / 3 fail |
| an unrelated feature branch | 1 pass / 2 fail |

Stall slots observed: 45, 121, 122, 128, 147, 152, 154, 161. It is not specific to the epoch boundary; three runs happening to cluster near slot 121 made it look that way.

## The behaviour is deliberate, and a fix has to renegotiate a test

`#3955` documented this exact state and deliberately did not change it:

> a best-peer switch disables forging outright, independent of how fresh the local tip is and independent of `forgeSyncToleranceSlots` … This adds a reproducer that asserts the CURRENT behaviour rather than changing it. The behaviour is deliberate and already pinned by `TestCheckAndForgeProductionWaitsForUnknownActiveUpstreamTarget`.

`#3955` raised the skip to Warn with `leader_slot=true` when the slot is one this node is scheduled to lead, which is the only reason the counts above are recoverable from a log. It changed the visibility, not the decision.

So `TestCheckAndForgeProductionWaitsForUnknownActiveUpstreamTarget` asserts current behaviour by design. A fix must say what that test asserts instead and why, rather than delete it.

## Acceptance criteria

- The unknown-target case is bounded: either by how long `targetSlot` may remain unknown, or by distinguishing "no target yet" from "target says we are behind". Dropping the gate is not acceptable — it exists so a node that has just switched peers does not forge on a stale view.
- A node whose own tip is fresh does not decline a scheduled leader slot solely because the newly selected peer has not yet delivered an admitted trusted header.
- A node that is genuinely behind still declines.
- `TestCheckAndForgeProductionWaitsForUnknownActiveUpstreamTarget` is renegotiated explicitly, not removed.
- Proven on DevNet across enough runs to mean something against a per-arm failure rate of roughly 50–100%.

## Related

- #3180 — the same stall shape reported against `TestSustainedConsensus`; run evidence is posted there and is not repeated here.
- #3954, #3955 — the two pre-leader gates and the visibility fix.
- #3968 — reconnect during an in-flight rollback offering only origin, adjacent to the active-connection switch window.

wolf31o2:
## Defect

`ledger/forging/forger.go:823-826` skipped forging whenever `UpstreamSyncStatus`
reported an active upstream with target zero:

```go
if upstreamActive && (upstreamTip == 0 ||
    (upstreamTip > tipSlot &&
        upstreamTip-tipSlot > f.forgeSyncToleranceSlots)) {
```

The `upstreamTip == 0` disjunct short-circuits before `forgeSyncToleranceSlots`,
so the tolerance never applied to it however fresh the local tip was.

`LedgerState.UpstreamSyncStatus` (`ledger/state.go:9207-9221`) returns
`(0, true)` for the whole window between an active-connection switch and the
newly selected peer's first authenticated, admitted, trusted header:
`publishActiveUpstream` (`ledger/state.go:9233-9249`) stores the new connection
key with `targetSlot` zero and only `publishAdmittedUpstreamTarget`
(`ledger/state.go:9257-9283`) lifts it.

On a network where forging is the only source of new headers that window is
self-sealing: no node forges because each is inside it, so no header is
produced, so none is admitted, so nothing lifts the target off zero. The chain
stops while every node reports healthy, connected and `lastErr=""`.

In one failing run the gate dropped 67 scheduled leader slots across the three
producers (22 / 29 / 16), every one at `upstream_tip=0`, with the local tip 3 to
16 slots behind the current slot — far inside the tolerance of 100.

## Change

A zero target is not evidence that this node is behind; it means the selected
peer has not spoken yet. The only evidence available in that window is the local
tip's lag behind the wall clock, so the same tolerance now applies to it. A node
whose tip is stale still waits, which is the protection this gate exists for. A
node at tip forges, and the header it produces is what ends the window.

The known-target branch is unchanged. No metric is added or changed.

## What the pinned test asserts now

`TestCheckAndForgeProductionWaitsForUnknownActiveUpstreamTarget` asserted that
the wait happened regardless of local tip freshness, with the tip one slot
behind the current slot. #3955 added it deliberately and said so:

> a best-peer switch disables forging outright, independent of how fresh the
> local tip is and independent of `forgeSyncToleranceSlots` … This adds a
> reproducer that asserts the CURRENT behaviour rather than changing it … If the
> gate is later keyed on local tip freshness instead of on the upstream target
> being known, this test fails and names the decision that was revisited.

This is that decision. The test keeps its name and still asserts that an unknown
target stops a forge, but now with the tip lagging by 991 slots against a
tolerance of 99 — the case that carries evidence. Refusing to forge on a stale
view after a peer switch is preserved; refusing while at tip is not.

`TestForgeSkipsLeaderSlotWhenUpstreamTargetUnknownEvenAtTip`, #3955's companion
documenting the cost, is replaced by
`TestForgeTakesLeaderSlotWhenUpstreamTargetUnknownAtTip`, which reverses its
assertion on the same fixture.

`TestForgeGateSkipWarnsOnlyForScheduledLeaderSlots` pins the log level of a gate
skip, not which disjunct fired; its upstream case now uses a known target ahead
of the tip so it still reaches the same gate.

## Tests

- `TestForgeTakesLeaderSlotWhenUpstreamTargetUnknownAtTip` — at tip, unknown
  target, scheduled leader: reaches leader selection, builds and broadcasts, and
  `forge_sync_skip` does not move. Fails without the change in both tolerance
  cases.
- `TestForgeStillWaitsForUnknownUpstreamTargetWhileTipIsStale` — unknown target
  with the tip 991 slots behind still skips. Passes with and without the change;
  it pins behaviour that must not move.

## DevNet

`./internal/test/devnet/run-tests.sh --accelerated`, five runs per arm,
interleaved on one machine:

| arm | pass | fail | `upstream_tip=0` gate skips |
| --- | --- | --- | --- |
| `eb6980af` (base) | 1 | 4 | 500 |
| this branch | 3 | 2 | 0 |

The wedge mechanism is gone. Base failures stall at slots 123, 126, 149, 154;
the two remaining failures on this branch are at later phases with the chain far
ahead of anywhere the base arm reaches:

- one at `"every node agrees on a header above the epoch boundary at 120"` with
  the network at slot 305 and dingo-1 on a minority fork it then refuses to
  leave, logging `chainsync rollback exceeds security parameter K, rejecting
  peer chain` and `temporarily denying chainsync peer whose chain we cannot
  follow` — the #3035 / #3996 shape;
- one at `"relay restart: dingo-relay rejoins and catches up to the network"`
  with three nodes at slot 475/block 144 and the restarted relay left at slot
  170.

Neither is the gate this changes. Whether they are pre-existing cannot be
settled from the base arm, because the base arm never reaches those phases;
they need their own investigation.

Fixes #4010

---
## Summary by cubic
Fixes #4010: the forging sync gate wedged all-producer networks by skipping every leader slot whenever the upstream sync target was zero, even when the local tip was current. That unknown-target window was self-sealing: no node forged, so no new header existed to lift the target and close the window.

The gate now applies `forgeSyncToleranceSlots` to the local tip's lag when the upstream target is zero, so a node at tip forges (its header ends the window) while a node whose tip is stale still skips. The known-target branch is unchanged.

- `tipGapSlots` now reports the local tip lag when the upstream target is unknown, and the `forgeSyncSkip` help text reflects local-or-upstream staleness.
- Tests updated to match: the wait-on-unknown-target test now uses a tip 991 slots behind, and a new test pins that an at-tip node takes the leader slot.

<sup>Written for commit b7369392d219e8d20002619bf2c74f1a9b95f4f4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4013?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ledger/forging/metrics.go
```diff
@@ -172,7 +172,7 @@ func initForgingMetrics(
 	m.forgeSyncSkip = factory.NewCounter(
 		prometheus.CounterOpts{
 			Name: "dingo_forge_sync_skip_total",
-			Help: "forging attempts skipped because upstream tip is ahead",
+			Help: "forging attempts skipped because the local or upstream tip is stale",
```
chrisguiney, line 175: Non-blocking, documentation drift.

The help text now covers both states, but the knob's own documentation still describes only the upstream case:

- `forger.go:54-59` — "the number of slots the chain tip may lag behind the upstream peer tip"
- `ForgerConfig.ForgeSyncToleranceSlots`, `forger.go:370-372` — "how far the local chain can lag the upstream tip"
- `dingo.yaml.example:441-443` — same wording, and this is the copy an operator reads

After this change the same tolerance also bounds the local tip's lag against the wall clock whenever the target is unknown, which is exactly what someone tuning the value needs to know. One sentence in each would close it. `ARCHITECTURE.md:4454` stays accurate as written, and `internal/config/flags.go:674-676` ("max slots behind tip before skipping block forging") is vague enough to remain true.

ledger/forging/forge_gates_test.go
```diff
@@ -262,49 +260,81 @@ func TestForgeSkipsLeaderSlotWhenUpstreamTargetUnknownEvenAtTip(
...
+		SlotClock: forgerTestSlotClock{
+			// The tip lags the current slot by more than the tolerance,
+			// which is direct evidence this node is behind.
+			currentSlot:       1000,
+			chainTipSlot:      9,
+			upstreamTipSlot:   0,
+			upstreamActive:    true,
+			slotsPerKESPeriod: 100000,
+		},
+		ForgeSyncToleranceSlots: 100,
```
chrisguiney, line 315: Non-blocking, coverage.

Both new cases sit far from the boundary — 991 against 100 here, and a lag of 1 against 100/100000 in `TestForgeTakesLeaderSlotWhenUpstreamTargetUnknownAtTip` — so nothing pins the strict `>` in `upstreamSyncSkipsForge`. A pair at `currentSlot-tipSlot == tolerance` (must forge) and `== tolerance+1` (must skip) would catch an off-by-one that both current tests pass straight through.

The known-target branch has the same gap, so this is pre-existing rather than introduced here; worth adding while the file is open.

chrisguiney (approved):
Approving. Reviewed at b7369392.

**Fail-before**

Reverted the gate to the pre-change expression in place, keeping `upstreamSyncSkipsForge` so the package still compiles. `TestForgeTakesLeaderSlotWhenUpstreamTargetUnknownAtTip` fails in both tolerance cases, and `TestForgeStillWaitsForUnknownUpstreamTargetWhileTipIsStale` fails too, on its `tipGapSlots == 991` assertion. The change is genuinely covered.

**Checks on the merged tree** (b7369392 already merges main, which moved `builder.go` and `keys.go` under `ledger/forging`)

| check | result |
| --- | --- |
| `go build ./...` | pass |
| `go vet ./ledger/...` | pass |
| `go test -race -count=1 ./ledger/forging/...` | pass |
| `golangci-lint run ./ledger/forging/...` | 0 issues |
| `make docs-parity` | pass |
| `gofmt -l .` | only `bark/blob_test.go` and `ledger/read_chain_pipeline_test.go`, both already dirty on `origin/main` |

**Gate soundness**

The unknown-target branch is a conservative superset of the known-target branch. If a peer's real tip led our tip by more than the tolerance, it would also exceed our lag against the wall clock, since a peer's tip cannot exceed the current slot — so that case still skips. Empty database, genesis, and post-rollback all land on "lag grows, so skip", which is the safe side. The residual exposure is a node whose own clock is behind by more than the tolerance, which is pre-existing for the whole forging path.

**Class audit**

`upstreamActive && upstreamTip == 0` has one other site, `ledger/state.go:3003`, which defers consumed-UTxO cleanup. It is not self-sealing — cleanup produces no headers — and it clears once forging resumes. The class is complete; nothing further to fix.

**Relationship to #4036**

No file overlap. #4036 leans on the per-slot KES gate in `checkAndForgeProduction`, which runs after this gate and fails closed, so more slots reaching it is the direction #4036 wants. No contradiction.

**Bot review status on this head**

- CodeRabbit's status check is SUCCESS, but its comment body is "Review limit reached". It did not review this PR.
- cubic reviewed the code commits; its one finding (zero tip gap on the unknown-target path) was addressed by 6c2ad0de and the thread is resolved. Its check on b7369392 reads "AI review not needed for merge commit".

**Not run here**: DevNet, including the `--accelerated` arms in the description — taking those as reported. `nilaway` and `modernize` are not installed in this environment.

**Description nit**: it states that `TestForgeStillWaitsForUnknownUpstreamTargetWhileTipIsStale` "Passes with and without the change". That held at aa6cb460; the `tipGapSlots` assertion added in 6c2ad0de makes it fail without the change. Stronger than advertised, so only the wording is stale.

Two non-blocking notes inline.

Issue #3823, opened 2026-09-03
flake: TestLiveTruncateCancelsInsteadOfResumingWhenStorageDrainUnconfirmed fails on unmodified main


wolf31o2:
`node_lifecycle_test.go:972` `TestLiveTruncateCancelsInsteadOfResumingWhenStorageDrainUnconfirmed` fails 2 of 3 runs under `-race` on `origin/main` at `1b5b2852`.

Distinct from the known flakes in this package family: #3759 (`internal/dblifecycle` `require.Eventually` deadlines), #3746 (`TestManager*` on Windows), #3768 (`internal/koiosparity` `SQLITE_BUSY`), #3774 (`midnight/server` reflection EOF).

Acceptance criteria: the test passes reproducibly under `-race`, or its synchronization is corrected so the cancel/resume decision is not timing-dependent. No `time.Sleep` sequencing — use the repository wait helpers or a context deadline.

wolf31o2:
#3913 merged and is linked as `Refs`, not `Fixes`, so this stays open.

That change preserves the first `LedgerState.Close` result across repeated calls, so a shutdown after a cancelled live truncate no longer reports success and closes storage underneath a still-running worker. It alters production `Close` semantics and adds assertions; it does not change this test's synchronization.

The acceptance criteria here are that the test pass reproducibly under `-race`, or that the cancel/resume decision stop being timing-dependent. Neither is demonstrated: `TestLiveTruncateCancelsInsteadOfResumingWhenStorageDrainUnconfirmed` passed 50/50 under `-race` on unmodified `origin/main` during review (darwin/arm64, go1.27.1), so the flake was not reproduced and cannot be confirmed fixed.

Closing this needs either a reproduction showing the failure before #3913 and its absence after, or a change making the cancel/resume decision deterministic.

chrisguiney:
Closing: the cancel/resume decision in this test is deterministic on origin/main 440d01ab, so there is no timing edge left to fix.

- The test's only dbWorkerPool task blocks on a release channel that is closed in t.Cleanup (node_lifecycle_test.go:1167-1173), that is, only after the test body returns.
- The test sets ledger.CloseDBWorkerPoolShutdownTimeout to 20ms (:1159-1161), so LedgerState.Close can never see the drain finish and always returns the "could not confirm" error. The other quiesce and shutdown components this harness populates are nil or immediate no-ops.
[...]

Issue #3768, opened 2026-09-02
test(koiosparity): concurrent chunk writers rely on SQLite busy_timeout and flake on Windows


wolf31o2:
## Problem

`internal/koiosparity.TestFetchAccountRewardsForEpochRequestBodyNeverExceedsConfiguredMaxBytes`
fails intermittently on Windows CI:

```
fetch_accounts_resilience_test.go:1192:
  save account fetch chunk progress: database is locked (5) (SQLITE_BUSY)
```

Observed on run 33547213058 (PR #3692), where Linux, Linux-race and macOS all
passed on the same run. The PR under test touches only `ARCHITECTURE.md` and
`ledger/**`; `koiosparity` imports none of those, so the test binary was
unchanged from the merge base.

## Mechanism

`FetchEpochAccountsWithAddrs` (`fetch_accounts.go:435`) fans chunks across
goroutines. Each opens its own write transaction against a single SQLite cache
with `busy_timeout(5000)` and an unbounded connection pool, so the 5-second
busy timeout is the only thing serializing writers.

That is a timing assumption, not a correctness one. The test took 18.14s on the
Windows runner against roughly 2.5s for `-count=5` locally, so the timeout
expired rather than the behavior being wrong.

## Why it matters

`go-test (Windows)` runs the full suite on every pull request and every push to
main. A test that fails on runner speed produces failures that look like real
regressions on unrelated PRs, which trains readers to discount Windows results.

This is a separate mechanism from the `internal/dblifecycle` wait-deadline
family in #3759 — different package, and a database lock rather than a
`require.Eventually` deadline — so it needs its own fix.

## Acceptance criteria

- Serialize the concurrent chunk writers explicitly rather than relying on
  `busy_timeout` to arbitrate them, or bound the write concurrency to what one
  SQLite connection can serve.
- Do not simply raise `busy_timeout`: that hides the contention rather than
  removing it and leaves the failure reachable on a slower runner.
- Confirm on Windows CI, not only locally — the package is fast on Linux and
  the failure is runner-speed dependent, so a green Linux run proves nothing.

chrisguiney:
Fixed on `main` by #4124 (`d4163992`, merged 2026-09-07), which is reachable from current `origin/main`. That PR carried no `Fixes #` reference, so this stayed open.

[...]

Issue #3759, opened 2026-09-01
test(dblifecycle): wait deadlines assume a fast runner and flake on Windows


wolf31o2:
## Problem

`internal/dblifecycle/manager_test.go` waits on snapshot and pruning outcomes
with `require.Eventually` and a 5-second deadline. Windows runners are slow
enough that the deadline expires while the behavior under test is still
progressing, so the tests fail with `Condition never satisfied` even though the
node did the right thing.

Three instances observed, all on Windows, all the same shape:

| Test | Where |
|---|---|
| `TestManagerPruningDeletesCloudMirror` | main runs 33413291419, 33422972126, 33440696014, 33502559781 — fixed by #3715 |
| `TestManagerRedeliveredEventIsNotFatal` | main run 33554142075, `manager_test.go:357` |
| `TestManagerPruningKeepsLocalCopyUntilCloudDeleteSucceeds` | PR run 33548321853, `manager_test.go:1555` |

Run 33554142075 is the clearest evidence that the deadline rather than the
behavior is at fault. Its log shows the epoch-7 snapshot captured at 20:30:02
and epoch-8 at 20:30:15 — the redelivered event was not fatal and a later epoch
was captured, exactly what the test asserts, about 13 seconds after the
5-second bound expired. Every other job on that run passed, including
`go-test (Linux, race)`.

#3715 addressed one member by synchronizing its assertion against upload
completion rather than by raising a timeout. That is the right shape of fix,
but it was applied to a single test while the family shares the pattern.

## Why this matters now

`go-test (Windows)` runs the full suite on every pull request and every push to
main as of #3744. A family of tests that intermittently fails on Windows for
timing reasons produces failures that look like real regressions, which trains
readers to discount Windows results — the failure mode that let the Windows-only
DevNet breakage in #3728 reach main four merges running.

## Acceptance criteria

- Audit every `require.Eventually` and equivalent wait in
  `internal/dblifecycle` for a deadline that assumes a fast runner.
- Prefer synchronizing against the event the test actually depends on, the way
  #3715 did, over raising the deadline; raise it only where no such signal
  exists, and say why in place.
- Confirm on Windows CI, not only locally: the package is fast on Linux and the
  failures are runner-speed dependent, so a green Linux run proves nothing here.
- No `time.Sleep` for synchronization (repository rule); use the helpers in
  `internal/test/testutil`.

cryptodj413:
[...]

PR #4014, opened 2026-09-06
test(midnight): tolerate the io.EOF gRPC reports from Send on a terminated stream

### LINKED ISSUE #3774: flake: TestReflectionDisabledByDefault fails with EOF from stream.Send
midnight/server/server_test.go:279 asserts `require.NoError(t, stream.Send(...))`.

With reflection unregistered the server answers `Unimplemented` and terminates the stream. gRPC `SendMsg` returns `io.EOF` once the server has done so; the status must come from `Recv`. When the server-side close lands before the client Send, `Send` returns EOF and the assertion fails. Lines 284-286 already read the status from `Recv` correctly.

Observed on run 33625153589, `go-test (macOS)` and `go-test (Linux, race)`. No `DATA RACE` reports — assertion failure, not a detector finding. Introduced by bea15363 (on main); not attributable to any open branch. Not reproduced locally in 300 plain + 150 race runs; load-sensitive.

Acceptance criteria: `Send` tolerates `io.EOF`; the `Unimplemented` assertion on `Recv` is unchanged; test passes under load.

chadaibbhot-git:
## Problem

`midnight/server/server_test.go`'s `TestReflectionDisabledByDefault` asserts

```go
require.NoError(t, stream.Send(&reflectionpb.ServerReflectionRequest{...}))
```

With reflection unregistered — the default — the server rejects
`ServerReflectionInfo` with `Unimplemented` and terminates the stream. It
does not wait for a request first, so its rejection can reach the client
transport before the client's `Send` runs.

gRPC's `ClientStream` contract makes that ordering an assertion failure
rather than a real defect: `SendMsg` returns `io.EOF` once the stream has
been terminated by the server, and the RPC status "can be discovered by
calling RecvMsg". The test already reads the status from `Recv` on the
next three lines and asserts `codes.Unimplemented` there — so the `Send`
assertion carries no information, and only ever loses a race.

CI hit the losing ordering on run 33625153589 (`go-test (macOS)` and
`go-test (Linux, race)`), failing with `Received unexpected error: EOF`.
No `DATA RACE` was reported: it is an assertion failure, not a detector
finding.

## Fix

`Send` tolerates `io.EOF` (and only `io.EOF`); every other error still
fails the test. The `Unimplemented` assertion on `Recv` is unchanged, so
what the test actually covers — reflection is off unless opted in — is
untouched.

`TestReflectionDisabledByDefaultWhenRejectionArrivesFirst` pins the
losing ordering deterministically: it reads the rejection first, so the
`Send` that follows always runs against an already-terminated stream. The
two tests together cover both orderings instead of only the one a fast
runner happens to produce.

## Verification

- Red before the fix: with the previous `require.NoError` assertion, the
  new test fails with exactly the reported error —
  `Received unexpected error: EOF` at the `Send`. Green after.
- `go test ./midnight/server/ -count=30 -race` green (573 s).
- `gofmt`, `go vet ./midnight/server/` and `golangci-lint run ./midnight/...`
  clean.

The flake was not reproducible by repetition locally (the issue reports
300 plain + 150 race runs without a failure), which is why the regression
test forces the ordering rather than repeating the test and hoping.

Related: #3774 (this flake), #3848 (lifecycle tests failing under
parallel suite load), #3823, #3759, #3768, #3541 — the same "CI-only,
load-sensitive test assertion" family. No open PR addresses #3774.

Fixes #3774

---
## Summary by cubic
Fixes a flaky test in `midnight/server` where `TestReflectionDisabledByDefault` sometimes failed because the server's `Unimplemented` rejection could reach the client before `Send` completed. Now `Send` tolerates `io.EOF` from a terminated stream, and a new test deterministically covers that ordering.

- `TestReflectionDisabledByDefault` keeps asserting `codes.Unimplemented` from `Recv`, so verification is unchanged.
- Adds `TestReflectionDisabledByDefaultWhenRejectionArrivesFirst` to force the rejection-first ordering.
- Only `io.EOF` is tolerated; any other `Send` error still fails the test.

<sup>Written for commit 50c320467197d44b79358a6fe16b2beeb9e8e9fc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4014?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/midnight/server/server_test.go
+++ b/midnight/server/server_test.go
@@ -16,6 +16,7 @@ package server_test
 
 import (
 	"context"
+	"io"
 	"net"
 	"strconv"
 	"testing"
@@ -269,6 +270,34 @@ func TestReflectionListsService(t *testing.T) {
 }
 
 func TestReflectionDisabledByDefault(t *testing.T) {
+	stream := reflectionStreamToServerWithoutReflection(t)
+
+	sendListServices(t, stream)
+	_, err := stream.Recv()
+	require.Error(t, err)
+	require.Equal(t, codes.Unimplemented, status.Code(err))
+}
+
+// TestReflectionDisabledByDefaultWhenRejectionArrivesFirst resolves the
+// race that made the test above flaky, deterministically, the losing way:
+// the server's rejection is observed first, so the Send that follows runs
+// against a stream the server has already terminated. That is the ordering
+// CI hit, and it must be a pass -- the status is carried by Recv, and Send
+// can only report the termination as io.EOF.
+func TestReflectionDisabledByDefaultWhenRejectionArrivesFirst(t *testing.T) {
+	stream := reflectionStreamToServerWithoutReflection(t)
+
+	_, err := stream.Recv()
+	require.Error(t, err)
+	require.Equal(t, codes.Unimplemented, status.Code(err))
+
+	sendListServices(t, stream)
+}
+
+func reflectionStreamToServerWithoutReflection(
+	t *testing.T,
+) grpc.BidiStreamingClient[reflectionpb.ServerReflectionRequest, reflectionpb.ServerReflectionResponse] {
+	t.Helper()
 	addr := startTestServer(t)
 	conn := dial(t, addr)
 	readyCtx, readyCancel := context.WithTimeout(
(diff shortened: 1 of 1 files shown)


chadaibbhot-git:
CI note: the `go-test` failures on this head are inherited from the base, not from this branch. Both reproduce on a clean checkout of `main` at 64676163 with no changes applied:

- `ledger/leader` — `TestThresholdPreservesProtocolBoundary/{cpraos,tpraos}_fractional_cutoff_admits_its_floor`: `expected: 0, actual: 1` from `threshold.Cmp(want)`. Deterministic, and it also fails at 4eab41e1 ("fix(ledger): correct Preview VRF leader threshold boundary (#3897)"), the commit that added both the test and `consensus/leaderthreshold/threshold.go`.
- `internal/koiosparity` — ~10 tests fail with `create sqlite schema: SQL logic error: table epoch already exists (1)`.

`main`'s own CI run for 64676163 shows `go-test (macOS)` and `go-test (Windows)` red as well, so this is not platform- or branch-specific.

This branch's own packages are green locally, including under `-race`. I am not touching either area from here — `ledger/leader` is consensus code and a guess there is worse than a red check.

chadaibbhot-git:
CI on b3b820f0 confirms the split: `midnight/server` — the only package this PR touches — is green on every job, including `go-test (Linux, race)` (`ok github.com/blinklabs-io/dingo/midnight/server 24.788s`). The four red `go-test` jobs fail only in `internal/koiosparity` and `ledger/leader`, which fail identically on a clean `main` (see the previous comment).

`lint`, `govulncheck`, `docker`, `DCO` and `Conventional Commits` are green.

chadaibbhot-git:
Rebased onto current `main` (eabf653d), no conflicts — the `go-test` failures on the previous head were inherited from the stale base.

Re-verified on the new head: reverting the tolerant `Send` back to `require.NoError` fails `TestReflectionDisabledByDefaultWhenRejectionArrivesFirst` and passes again with it restored (red→green), `go test -race -count=30 -run TestReflection ./midnight/server/` is 90/90 green, and the whole `./midnight/...` tree passes under `-race`.

chadaibbhot-git:
@coderabbitai review (final head 50c32046)

chadaibbhot-git:
Superseded by #4129, which merged at 2026-09-08T15:07Z and makes the same change to `TestReflectionDisabledByDefault` (accept a rejected stream from the reflection `Send`). Closing this one; nothing left to merge here. Thanks.

PR #4024, opened 2026-09-06
fix(leaderthreshold): stop rounding the leader threshold up twice

### LINKED ISSUE #4021: consensus/leaderthreshold: the leader threshold rounds up twice since the gouroboros v0.202.10 adoption
`go test ./ledger/leader/` fails on `main`, on Linux, Windows and macOS:

```
--- FAIL: TestThresholdPreservesProtocolBoundary/cpraos_fractional_cutoff_admits_its_floor
    threshold_test.go:96: Not equal: expected: 0  actual: 1
--- FAIL: TestThresholdPreservesProtocolBoundary/tpraos_fractional_cutoff_admits_its_floor
    threshold_test.go:96: Not equal: expected: 0  actual: 1
FAIL	github.com/blinklabs-io/dingo/ledger/leader
```

### The defect

`leaderthreshold.Threshold` (`consensus/leaderthreshold/threshold.go`) adds 1 to the
value returned by `consensus.CertifiedNatThresholdWithMode` whenever `cutoffIsInteger`
is false. That upstream value already includes the rounding.

Measured for f=1/3, poolStake=1, totalStake=1, relative to `floor(2^bits/3)`:

| | CPraos (256) | TPraos (512) |
|---|---|---|
| `CertifiedNatThresholdWithMode` | floor + 1 | floor + 1 |
| `leaderthreshold.Threshold` | floor + 2 | floor + 2 |
| `threshold_test.go` expects | floor + 1 | floor + 1 |

The rounding is applied twice, so the comparison threshold is one too large and the
`certNat < threshold` leader check admits one VRF value it should reject. The
integer-cutoff cases are unaffected, which is why only the two `fractional_cutoff`
subtests fail.

### Why it was not caught

The compensation was correct when written. gouroboros v0.202.9
(`e5261b1`, "fix(consensus): preserve leader threshold boundaries", #2192) changed
`CertifiedNatThresholdWithMode` from `floor(T)` to `ceil(T)`:

```
-// integer threshold floor(upperBound * (1-(1-f)^sigma)) computed via
+// integer comparison threshold ceil(upperBound * (1-(1-f)^sigma)) computed
```

- #3897 added both `Threshold` and its test, pinned to gouroboros **v0.202.8** (floor).
  All 13 checks passed on head `a93dfe1`.
- #4006 adopted gouroboros **v0.202.10** (ceil) and merged at 18:07Z.
- #3897 merged at 18:11Z, four minutes later.

Each pull request was green against the version it was tested with. The pairing
was never built until it reached `main`.

### Fix

With upstream returning `ceil(T)`, `Threshold` should return that value unchanged:
`ceil` already equals `floor+1` for a non-integer cutoff and the exact value for an
integer one, which is what the test asserts on both paths. `cutoffIsInteger` and
`exactNthRoot` become dead code.

The correct rounding must be confirmed against the reference implementation before
the change lands, not inferred from the test alone.

wolf31o2:
Fixes #4021.

`leaderthreshold.Threshold` added one to `consensus.CertifiedNatThresholdWithMode`
whenever exact rational arithmetic showed the real cutoff was not an integer. That
compensated for gouroboros v0.202.8, which returned `floor(T)`.

gouroboros v0.202.9 (`e5261b1`, #2192) changed the function to return `ceil(T)`:

```
-// integer threshold floor(upperBound * (1-(1-f)^sigma)) computed via
+// integer comparison threshold ceil(upperBound * (1-(1-f)^sigma)) computed
```

Since `main` adopted v0.202.10 the increment is applied on top of a ceiling, so
the comparison threshold is one too large and `certNat < threshold` admits the
ceiling itself, which the real-valued rule rejects.

Measured for f=1/3, poolStake=1, totalStake=1, relative to `floor(2^bits/3)`:

| | CPraos (256) | TPraos (512) |
|---|---|---|
| `CertifiedNatThresholdWithMode` | floor + 1 | floor + 1 |
| `Threshold` before | floor + 2 | floor + 2 |
| `Threshold` after | floor + 1 | floor + 1 |

For an integer leader value `v` and non-integer `T`, `v < T` iff `v <= floor(T)`
iff `v < ceil(T)`; for integer `T`, `ceil(T) == T`. The upstream value is therefore
correct on both paths and is now returned unchanged.

`cutoffIsInteger` and `exactNthRoot` existed only to decide that increment and have
no other callers or tests, so they are removed.

Verification:

- `go test ./ledger/leader/`: exit 0, `ok ... 2.319s`
- reverting the fix in place: exit 1, `TestThresholdPreservesProtocolBoundary/cpraos_fractional_cutoff_admits_its_floor` and `/tpraos_fractional_cutoff_admits_its_floor` fail
- `go test ./ledger/ ./ledger/leader/ ./consensus/...`: exit 0
- `golangci-lint run ./consensus/... ./ledger/...`: 0 issues

`TestThresholdPreservesProtocolBoundary` was added by #3897 alongside the code it
tests and has never passed on `main`; #3897 was green because it was pinned to
v0.202.8, and #4006 adopted v0.202.10 four minutes before it merged.

---
## Summary by cubic
Fixes #4021. Stops rounding the leader threshold up twice so the protocol boundary is no longer admitted. `leaderthreshold.Threshold` added 1 to `consensus.CertifiedNatThresholdWithMode` to compensate for gouroboros v0.202.8 returning `floor(T)`, but since v0.202.9 that function returns `ceil(T)`, so the increment made the threshold one too large and admitted the ceiling itself. The upstream value is now returned unchanged, and the `cutoffIsInteger` and `exactNthRoot` helpers that existed only to decide that increment are removed.

<sup>Written for commit 96f39d8d2b5dd5b5a49a5dc3f875fb8557432410. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4024?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/consensus/leaderthreshold/threshold.go
+++ b/consensus/leaderthreshold/threshold.go
@@ -23,124 +23,21 @@ import (
 // Threshold returns the integer comparison threshold for the Praos leader
 // rule. The ledger rule compares the real-valued certified-natural ratio
 // strictly against 1-(1-f)^sigma. For an integer leader value v, that is
-// equivalent to v < ceil(certNatMax*(1-(1-f)^sigma)), not v < floor(...).
+// equivalent to v < ceil(certNatMax*(1-(1-f)^sigma)).
 //
-// The gouroboros version used by Dingo currently returns floor(...). Keep its
-// rigorously calculated value as the base, and raise it by one only when exact
-// rational arithmetic proves the real cutoff is not an integer. When the
-// cutoff is integral, the returned floor is retained so the exact boundary is
-// still rejected.
+// consensus.CertifiedNatThresholdWithMode returns that ceiling, so its value
+// is used unchanged. Adding to it would admit the ceiling itself, which the
+// real-valued rule rejects.
 func Threshold(
 	poolStake uint64,
 	totalStake uint64,
 	activeSlotCoeff *big.Rat,
 	mode consensus.ConsensusMode,
 ) (*big.Int, error) {
-	threshold, err := consensus.CertifiedNatThresholdWithMode(
+	return consensus.CertifiedNatThresholdWithMode(
 		poolStake,
 		totalStake,
 		activeSlotCoeff,
 		mode,
 	)
-	if err != nil {
-		return nil, err
-	}
-	if cutoffIsInteger(poolStake, totalStake, activeSlotCoeff, mode) {
-		return threshold, nil
-	}
-	return new(big.Int).Add(threshold, big.NewInt(1)), nil
-}
-
-// cutoffIsInteger proves whether the real cutoff is an integer for the
-// rational protocol inputs Dingo has available. If sigma's reduced exponent
-// does not make (1-f)^sigma rational, the cutoff is irrational and therefore
-// cannot be an integer. The rational case is evaluated exactly.
-func cutoffIsInteger(
-	poolStake uint64,
-	totalStake uint64,
-	activeSlotCoeff *big.Rat,
-	mode consensus.ConsensusMode,
-) bool {
-	if activeSlotCoeff == nil || activeSlotCoeff.Sign() <= 0 ||
-		totalStake == 0 || poolStake == 0 {
-		return true
-	}
-	if activeSlotCoeff.Cmp(big.NewRat(1, 1)) >= 0 {
-		// f == 1 has cutoff certNatMax. f > 1 is rejected by the
-		// underlying threshold helper before this result is used.
-		return true
-	}
-	if poolStake > totalStake {
-		poolStake = totalStake
-	}
-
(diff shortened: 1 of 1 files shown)


wolf31o2:
Superseded by #4023, which landed the same change on main (e883294f). `TestThresholdPreservesProtocolBoundary` passes on current main.

PR #4031, opened 2026-09-06
fix(ledgerstate): validate imported committee state

### LINKED ISSUE #3785: Conway genesis committee members are never persisted, so a genesis-synced node holds no committee rows
Nothing seeds `committee_member` from Conway genesis. The only production writers of `SetCommitteeMembers` are `ledger/governance/enact.go` (UpdateCommittee enactment) and `ledgerstate/import.go` (Mithril import).

`genesis.Committee.Threshold` is read for the CC quorum (`ledger/governance/epoch.go:1100`). `genesis.Committee.Members` is read only by an emptiness predicate (`ledger/eras/dijkstra.go:160`) and tests — never persisted.

A genesis-synced node therefore holds zero committee rows for the whole Conway era while the chain has the genesis committee seated. Consequences:

- Any committee query cannot distinguish "no committee" from "committee unknown", so a provider cannot honestly implement `common.CommitteeCredentialState.CommitteeStateAvailable()` (gouroboros #2146). #3745 reports unavailable for an empty table for exactly this reason.
- `GetCommitteeActiveCount` is not a substitute: it delegates to `GetActiveCommitteeMembers` and counts hot-key authorizations, not seated members.

Found while working #3745.

Acceptance criteria: Conway genesis committee members are persisted at genesis load with their credential tags and terms; an empty `committee_member` table then means an authoritative empty committee; a genesis-synced node answers a cold-credential lookup for a genesis CC member correctly; test covers a genesis with a non-empty committee.

wolf31o2:
## Summary

- Parse the Conway committee carried by `RatifyState.rsEnactState` separately from `rsEnacted`.
- Fail closed when the serialized active committee disagrees between `cgsCommittee` and `rsEnactState`.
- Add a focused regression fixture covering committee membership and quorum parity.

This preserves Conway boundary semantics: `rsEnactState` is already-active state, while `rsEnacted` is applied at the next epoch boundary. The importer does not apply `rsEnacted` early.

## Validation

- `go test ./ledgerstate`

Closes #3785

---
## Summary by cubic
Validates that an imported governance committee agrees with the Conway committee carried in `RatifyState.rsEnactState`, now failing only when no committee-changing action was enacted.

- Skips the comparison when `rsEnacted` carries a NoConfidence or UpdateCommittee action (which legitimately rewrites `ensCommittee` before `cgsCommittee` catches up) or when an enacted entry cannot be decoded.
- Fails closed when either committee fails to parse, `rsEnactState` lacks a committee field, or a GovState record is truncated; malformed `rsEnacted` proposals remain warnings.
- Adds regression fixtures covering matching, conflicting, changed, malformed, and indeterminate committee states.
- Closes #3785.

<sup>Written for commit 4135fc582119572b16e2c4c8c4a1988cc8cdd68f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4031?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved governance-state validation by checking committee consistency across state representations.
  * Import now rejects malformed committee or governance pulsing-state data instead of treating it as a warning.
  * Correctly handles committee changes resulting from UpdateCommittee and NoConfidence actions.
  * Reports truncated governance state data as a warning while continuing appropriate parsing.
* **Tests**
  * Added coverage for matching, conflicting, changed, malformed, and indeterminate committee states.

### DIFF
--- a/ledgerstate/certstate.go
+++ b/ledgerstate/certstate.go
@@ -1987,6 +1987,13 @@ type ParsedGovState struct {
 	Proposals            []ParsedGovProposal
 	PrevGovActionIds     *ParsedPrevGovActionIds
 	RatifiedGovActionIds []ParsedGovActionId
+	// EnactCommittee and EnactCommitteeQuorum are the committee values
+	// carried by RatifyState.rsEnactState.  This is the state already
+	// enacted at the snapshot boundary, independent of rsEnacted (which
+	// is applied at the next boundary).
+	EnactCommittee       []ParsedCommitteeMember
+	EnactCommitteeQuorum *cbor.Rat
+	EnactCommitteeSet    bool
 }
 
 // ParseGovState decodes governance state from raw CBOR.
@@ -2064,7 +2071,7 @@ func ParseGovState(
 	result.PrevGovActionIds = prevIds
 
 	if len(fields) >= 7 {
-		ratifiedIds, err := parseDRepPulsingStateRatifiedIds(
+		enactCommittee, enactQuorum, enactSet, ratifiedIds, err := parseDRepPulsingState(
 			fields[6],
 		)
 		if err != nil {
@@ -2072,6 +2079,9 @@ func ParseGovState(
 				"parsing drep pulsing state: %w", err,
 			))
 		}
+		result.EnactCommittee = enactCommittee
+		result.EnactCommitteeQuorum = enactQuorum
+		result.EnactCommitteeSet = enactSet
 		result.RatifiedGovActionIds = ratifiedIds
 	}
 
@@ -2360,23 +2370,23 @@ func parseProposals(data []byte) (
 // The enacted field is a sequence of GovActionState values in the same
 // representation used by cgsProposals, so parseGovActionState can
 // recover the exact action IDs without decoding the full enact state.
-func parseDRepPulsingStateRatifiedIds(
+func parseDRepPulsingState(
 	data []byte,
-) ([]ParsedGovActionId, error) {
+) ([]ParsedCommitteeMember, *cbor.Rat, bool, []ParsedGovActionId, error) {
 	if len(data) == 0 {
-		return nil, nil
+		return nil, nil, false, nil, nil
 	}
 	fields, err := decodeRawArray(data)
 	if err != nil {
-		return nil, fmt.Errorf(
+		return nil, nil, false, nil, fmt.Errorf(
 			"decoding DRepPulsingState: %w", err,
 		)
 	}
 	if len(fields) == 0 {
-		return nil, nil
+		return nil, nil, false, nil, nil
 	}
 	if len(fields) < 2 {
-		return nil, fmt.Errorf(
+		return nil, nil, false, nil, fmt.Errorf(
 			"DRepPulsingState has %d elements, expected 2",
 			len(fields),
 		)
@@ -2384,25 +2394,38 @@ func parseDRepPulsingStateRatifiedIds(
 
 	ratifyState, err := decodeRawArray(fields[1])
 	if err != nil {
-		return nil, fmt.Errorf(
+		return nil, nil, false, nil, fmt.Errorf(
 			"decoding RatifyState: %w", err,
 		)
 	}
 	if len(ratifyState) < 2 {
-		return nil, fmt.Errorf(
+		return nil, nil, false, nil, fmt.Errorf(
 			"RatifyState has %d elements, expected 4",
 			len(ratifyState),
 		)
 	}
+	var committee []ParsedCommitteeMember
+	var quorum *cbor.Rat
+	committeeSet := false
+	var errs []error
+	enactFields, enactErr := decodeRawArray(ratifyState[0])
+	if enactErr != nil {
+		errs = append(errs, fmt.Errorf("decoding RatifyState enact state: %w", enactErr))
+	} else if len(enactFields) > 0 {
+		committeeSet = true
+		committee, quorum, enactErr = parseCommittee(enactFields[0])
+		if enactErr != nil {
+			errs = append(errs, fmt.Errorf("decoding enact-state committee: %w", enactErr))
+		}
+	}
 
 	enacted, err := decodeRawArray(ratifyState[1])
 	if err != nil {
-		return nil, fmt.Errorf(
+		return committee, quorum, committeeSet, nil, errors.Join(append(errs, fmt.Errorf(
 			"decoding RatifyState enacted proposals: %w", err,
-		)
+		))...)
 	}
 	ratifiedIds := make([]ParsedGovActionId, 0, len(enacted))
-	var errs []error
 	for _, item := range enacted {
 		prop, err := parseGovActionState(item)
 		if err != nil {
@@ -2417,7 +2440,14 @@ func parseDRepPulsingStateRatifiedIds(
 		})
 	}
 
-	return ratifiedIds, errors.Join(errs...)
+	return committee, quorum, committeeSet, ratifiedIds, errors.Join(errs...)
+}
+
+// parseDRepPulsingStateRatifiedIds is retained for callers that only need
+// rsEnacted action IDs.
+func parseDRepPulsingStateRatifiedIds(data []byte) ([]ParsedGovActionId, error) {
+	_, _, _, ids, err := parseDRepPulsingState(data)
+	return ids, err
 }
 
 // parseProposalsRoots decodes the GovRelation StrictMaybe at the
--- a/ledgerstate/certstate.go
+++ b/ledgerstate/certstate.go
@@ -2074,7 +2074,7 @@ func ParseGovState(
 	result.PrevGovActionIds = prevIds
 
 	if len(fields) >= 7 {
-		enactCommittee, enactQuorum, enactSet, ratifiedIds, err := parseDRepPulsingState(
+		enactCommittee, enactQuorum, enactSet, ratifiedIds, committeeErr, err := parseDRepPulsingState(
 			fields[6],
 		)
 		if err != nil {
@@ -2085,8 +2085,8 @@ func ParseGovState(
 		result.EnactCommittee = enactCommittee
 		result.EnactCommitteeQuorum = enactQuorum
 		result.EnactCommitteeSet = enactSet
-		if err != nil {
-			result.EnactCommitteeParseError = err
+		if committeeErr != nil {
+			result.EnactCommitteeParseError = committeeErr
 		}
 		result.RatifiedGovActionIds = ratifiedIds
 	}
@@ -2378,64 +2378,65 @@ func parseProposals(data []byte) (
 // recover the exact action IDs without decoding the full enact state.
 func parseDRepPulsingState(
 	data []byte,
-) ([]ParsedCommitteeMember, *cbor.Rat, bool, []ParsedGovActionId, error) {
+) ([]ParsedCommitteeMember, *cbor.Rat, bool, []ParsedGovActionId, error, error) {
 	if len(data) == 0 {
-		return nil, nil, false, nil, nil
+		return nil, nil, false, nil, nil, nil
 	}
 	fields, err := decodeRawArray(data)
 	if err != nil {
 		return nil, nil, false, nil, fmt.Errorf(
 			"decoding DRepPulsingState: %w", err,
-		)
+		), nil
 	}
 	if len(fields) == 0 {
-		return nil, nil, false, nil, nil
+		return nil, nil, false, nil, nil, nil
 	}
 	if len(fields) < 2 {
 		return nil, nil, false, nil, fmt.Errorf(
 			"DRepPulsingState has %d elements, expected 2",
 			len(fields),
-		)
+		), nil
 	}
 
 	ratifyState, err := decodeRawArray(fields[1])
 	if err != nil {
 		return nil, nil, false, nil, fmt.Errorf(
 			"decoding RatifyState: %w", err,
-		)
+		), nil
 	}
 	if len(ratifyState) < 2 {
 		return nil, nil, false, nil, fmt.Errorf(
 			"RatifyState has %d elements, expected 4",
 			len(ratifyState),
-		)
+		), nil
 	}
 	var committee []ParsedCommitteeMember
 	var quorum *cbor.Rat
 	committeeSet := false
-	var errs []error
+	var committeeErr error
+	var idErrs []error
 	enactFields, enactErr := decodeRawArray(ratifyState[0])
 	if enactErr != nil {
-		errs = append(errs, fmt.Errorf("decoding RatifyState enact state: %w", enactErr))
+		committeeErr = fmt.Errorf("decoding RatifyState enact state: %w", enactErr)
 	} else if len(enactFields) > 0 {
 		committeeSet = true
 		committee, quorum, enactErr = parseCommittee(enactFields[0])
 		if enactErr != nil {
-			errs = append(errs, fmt.Errorf("decoding enact-state committee: %w", enactErr))
+			committeeErr = fmt.Errorf("decoding enact-state committee: %w", enactErr)
 		}
 	}
 
 	enacted, err := decodeRawArray(ratifyState[1])
 	if err != nil {
-		return committee, quorum, committeeSet, nil, errors.Join(append(errs, fmt.Errorf(
+		return committee, quorum, committeeSet, nil, committeeErr, fmt.Errorf(
 			"decoding RatifyState enacted proposals: %w", err,
-		))...)
+		)
 	}
 	ratifiedIds := make([]ParsedGovActionId, 0, len(enacted))
 	for _, item := range enacted {
 		prop, err := parseGovActionState(item)
 		if err != nil {
-			errs = append(errs, fmt.Errorf(
+			idErrs = append(idErrs, fmt.Errorf(
 				"decoding enacted proposal: %w", err,
 			))
 			continue
@@ -2446,7 +2447,7 @@ func parseDRepPulsingState(
 		})
 	}
 
-	return committee, quorum, committeeSet, ratifiedIds, errors.Join(errs...)
+	return committee, quorum, committeeSet, ratifiedIds, committeeErr, errors.Join(idErrs...)
 }
 
 // parseProposalsRoots decodes the GovRelation StrictMaybe at the
--- a/ledgerstate/certstate.go
+++ b/ledgerstate/certstate.go
@@ -1984,6 +1984,7 @@ type ParsedGovState struct {
 	Constitution         *ParsedConstitution
 	Committee            []ParsedCommitteeMember
 	CommitteeQuorum      *cbor.Rat
+	CommitteeParseError  error
 	Proposals            []ParsedGovProposal
 	PrevGovActionIds     *ParsedPrevGovActionIds
 	RatifiedGovActionIds []ParsedGovActionId
@@ -1991,9 +1992,10 @@ type ParsedGovState struct {
 	// carried by RatifyState.rsEnactState.  This is the state already
 	// enacted at the snapshot boundary, independent of rsEnacted (which
 	// is applied at the next boundary).
-	EnactCommittee       []ParsedCommitteeMember
-	EnactCommitteeQuorum *cbor.Rat
-	EnactCommitteeSet    bool
+	EnactCommittee           []ParsedCommitteeMember
+	EnactCommitteeQuorum     *cbor.Rat
+	EnactCommitteeSet        bool
+	EnactCommitteeParseError error
 }
 
 // ParseGovState decodes governance state from raw CBOR.
@@ -2053,6 +2055,7 @@ func ParseGovState(
 	// Parse committee (field 1) — best-effort
 	committee, quorum, err := parseCommittee(fields[1])
 	if err != nil {
+		result.CommitteeParseError = err
 		warnings = append(warnings, fmt.Errorf(
 			"parsing committee: %w", err,
 		))
@@ -2082,6 +2085,9 @@ func ParseGovState(
 		result.EnactCommittee = enactCommittee
 		result.EnactCommitteeQuorum = enactQuorum
 		result.EnactCommitteeSet = enactSet
+		if err != nil {
+			result.EnactCommitteeParseError = err
+		}
 		result.RatifiedGovActionIds = ratifiedIds
 	}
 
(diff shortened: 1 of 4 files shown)


ledgerstate/import.go
```diff
@@ -3158,6 +3163,18 @@ func importGovState(
 			"error", err,
 		)
 	}
+	// Conway serializes the already-active committee in both cgsCommittee
+	// and RatifyState.rsEnactState. Refuse a partial import if those views
+	// disagree: rsEnacted is the *next* boundary's work and must not be
+	// applied here as a workaround for inconsistent snapshot state.
+	if govState.EnactCommitteeSet && !committeeStatesEqual(
```
chrisguiney: **Blocker — behavioral: this rejects valid mainnet snapshots.**

The comment states that Conway "serializes the already-active committee in both `cgsCommittee` and `RatifyState.rsEnactState`". That is not what `rsEnactState` is.

`rsEnactState` is the *result* of the ratification pass: `ratifyLoop` folds each accepted action into the enact state via `enactAction`, and the next `ConwayEPOCH` copies `rsEnactState` into the new `ConwayGovState`. If it were merely a copy of the current gov state, that boundary copy would be a no-op and no `UpdateCommittee` could ever take effect.

Two independent sources in this workspace say the same thing:

- gouroboros v0.202.10 (the pin), `protocol/localstatequery/client.go` on `GetRatifyState`: *"Governance actions are checked for ratification on the epoch boundary and enacted on the following boundary. Use this to see the outcome of that process for the current epoch: the enact state (the committee, constitution, protocol parameters, treasury, withdrawals, and previous governance action IDs **that will take effect**)"*.
- `ARCHITECTURE.md`: *"Imported proposals whose `(tx_hash, action_index)` appears in that list are stored with `ratified_epoch`/`ratified_slot` at the snapshot epoch anchor, so **the next epoch boundary enacts the same already-ratified actions** as the certified ledger state instead of re-ratifying them one epoch late."*

So a snapshot taken in an epoch where an `UpdateCommittee` was ratified has `rsEnactState.ensCommittee != cgsCommittee` *by design*, and this check aborts the whole Mithril import with `governance committee disagrees between cgsCommittee and rsEnactState`.

It also contradicts this same function: ~130 lines below, `ratifiedGovActionIdSet(govState.RatifiedGovActionIds)` marks exactly those `rsEnacted` proposals ratified so the next boundary enacts them. This PR refuses to import in precisely the case that code exists to handle. The two behaviours are mutually exclusive.

If the goal is to detect corruption in the imported committee, the corroborating value has to be something that is genuinely equal to `cgsCommittee` at the snapshot point — comparing against `rsEnactState` only works when `rsEnacted` contains no committee-affecting action, and that condition has to be tested, not assumed. A test on a real certified snapshot from an epoch with a ratified `UpdateCommittee` would settle it.

ledgerstate/import.go
```diff
@@ -3431,6 +3448,38 @@ func importGovState(
...
+	})
+	for i := range leftMembers {
+		if committeeMemberKey(leftMembers[i]) != committeeMemberKey(rightMembers[i]) {
+			return false
+		}
+	}
+	if (leftQuorum == nil) != (rightQuorum == nil) {
+		return false
+	}
+	return leftQuorum == nil || leftQuorum.Rat.Cmp(rightQuorum.Rat) == 0
```
chrisguiney: **Blocker — CI.** This is the line failing `lint` on this head:

```
ledgerstate/import.go:3476:41: QF1008: could remove embedded field "Rat" from selector (staticcheck)
```

`cbor.Rat` embeds `*big.Rat`, so the selector promotes. Reproduced locally with an isolated `GOLANGCI_LINT_CACHE`; `origin/main` has no `committeeStatesEqual`, so the failure belongs to this change.

```suggestion
	return leftQuorum == nil || leftQuorum.Cmp(rightQuorum.Rat) == 0
```

ledgerstate/certstate.go
```diff
@@ -2064,14 +2074,20 @@ func ParseGovState(
...
 			fields[6],
 		)
 		if err != nil {
 			warnings = append(warnings, fmt.Errorf(
 				"parsing drep pulsing state: %w", err,
 			))
 		}
+		result.EnactCommittee = enactCommittee
+		result.EnactCommitteeQuorum = enactQuorum
+		result.EnactCommitteeSet = enactSet
```
chrisguiney: **Non-blocking (contingent on the blocker above) — the gate is bypassable by omission.**

`EnactCommitteeSet` is the only thing that arms the parity check in `importGovState`, and it is left `false`, with no error, in four distinct shapes of malformed input:

1. `len(fields) < 7` in `ParseGovState` — the whole block is skipped, so a GovState truncated to 6 fields skips validation.
2. `len(data) == 0` in `parseDRepPulsingState`.
3. `len(fields) == 0` — an empty `cgsDRepPulsingState` array.
4. `else if len(enactFields) > 0` below — an empty `rsEnactState` array.

For a check whose stated purpose is failing closed on attacker- or corruption-influenced snapshot input, an attacker inverts the gate by *deleting* the corroborating field rather than forging it. As written the check only catches accidental corruption where both views happen to be present.

Whatever replacement lands for the parity check needs an explicit decision on the absence case: either require the field for Conway snapshots, or state in a comment why absence is acceptable.

ledgerstate/certstate.go
```diff
@@ -2064,14 +2074,20 @@ func ParseGovState(
...
+		if committeeErr != nil {
+			result.EnactCommitteeParseError = committeeErr
```
chrisguiney: **Confirmed — cubic's open finding here is valid.** I reproduced it against this head rather than taking the bot's word for it.

`committeeErr` is recorded only on `result`, never appended to `warnings`, so `ParseGovState` returns a `nil` error for input that failed to parse. On `origin/main` those same inputs produced a `"parsing drep pulsing state: ..."` warning through the returned error. `ParseGovState` is exported; the in-repo caller (`importGovState`) happens to check `EnactCommitteeParseError`, but the exported error contract changed silently.

cubic's suggested fix is the right minimal one — append to `warnings` *and* keep the dedicated field.

ledgerstate/certstate.go
```diff
@@ -2360,53 +2376,67 @@ func parseProposals(data []byte) (
 // The enacted field is a sequence of GovActionState values in the same
 // representation used by cgsProposals, so parseGovActionState can
 // recover the exact action IDs without decoding the full enact state.
-func parseDRepPulsingStateRatifiedIds(
+func parseDRepPulsingState(
 	data []byte,
-) ([]ParsedGovActionId, error) {
+) ([]ParsedCommitteeMember, *cbor.Rat, bool, []ParsedGovActionId, error, error) {
```
chrisguiney: **Non-blocking — API shape, and it has already produced a real defect.**

Two bare, adjacent `error` returns in a six-value signature are indistinguishable at the call site, and four of the returns below already put the wrong error in the wrong slot: the `decoding DRepPulsingState`, `DRepPulsingState has %d elements`, `decoding RatifyState` and `RatifyState has %d elements` paths all return their structural error in the **`committeeErr`** position with `nil` in the id-error position.

The consequence is visible to the operator: a `cgsDRepPulsingState` truncated to one element surfaces from `importGovState` as

```
parsing imported enact-state committee: DRepPulsingState has 1 elements, expected 2
```

which names the wrong subsystem. It also means those four cases produce no `ParseGovState` warning at all (see the comment at line 2089).

A small named result struct, or a single joined error with the committee error kept separate only where it is genuinely a committee error, removes the whole class.

ledgerstate/import.go
```diff
@@ -3431,6 +3448,38 @@ func importGovState(
...
+}
+
+func committeeMemberKey(member ParsedCommitteeMember) string {
+	return fmt.Sprintf("%d:%x:%d", member.ColdCredential.Type, member.ColdCredential.Hash, member.ExpiresEpoch)
```
chrisguiney: **Style.** `AGENTS.md` lists `make golines` in pre-commit (`--max-len=80`). Seven new non-test lines exceed 80 columns; this one is 108. The others are `certstate.go` 2077 (97) / 2381 (81) / 2450 (90) and `import.go` 3153 (83) / 3156 (100) / 3466 (82). Running `make golines` before the next push covers them.

chrisguiney (dismissed):
CHANGES_REQUESTED. Two blockers.

**1. The invariant the check enforces does not hold.** `RatifyState.rsEnactState` is the enact state that *will take effect* at the next epoch boundary, not the already-active state. gouroboros (the pinned v0.202.10) documents it that way on `GetRatifyState` — "the enact state (the committee, constitution, protocol parameters, ...) that will take effect" — and dingo's own `ARCHITECTURE.md` says imported `rsEnacted` proposals are marked ratified "so the next epoch boundary enacts the same already-ratified actions". When `rsEnacted` carries an `UpdateCommittee`, `rsEnactState.ensCommittee` differs from `cgsCommittee` by construction, and `importGovState` now aborts a valid certified snapshot. It also contradicts this same function a few hundred lines down, which marks that very proposal ratified so the next boundary enacts it.

**2. `lint` fails on this head, on a line this PR adds** — `ledgerstate/import.go:3476:41: QF1008: could remove embedded field "Rat" from selector (staticcheck)`. Reproduced locally with an isolated lint cache. Attributed to the change: `committeeStatesEqual` does not exist on `origin/main`.

Separately, the description says `Closes #3785`, but #3785 is Conway *genesis* committee seeding and #4033 already implements it. Merging this would auto-close #3785 unfixed.

**Checks run locally on 05ae26d, clean worktree off `origin/main` (eabf653):**

| check | result |
|---|---|
| `go build ./...` | 0 |
| `go vet ./ledgerstate/...` | 0 |
| `gofmt -l` on the 3 changed files | clean |
| `GOLANGCI_LINT_CACHE=<isolated> golangci-lint run ./ledgerstate/...` | 1 issue (the QF1008 above) |
| `go test ./ledgerstate/ -count=1` | pass |
| fail-without-fix (revert in place, helpers kept) | `TestImportGovStateRejectsCommitteeMismatch` and `TestParseGovStateEnactProposalWarningDoesNotInvalidateCommittee` both fail — the change is genuinely tested |

Not run: DevNet, `internal/test/conformance`, `modernize`, `nilaway` (last two not installed in this environment).

**Bot review status:** CodeRabbit did **not** review this head — its comment on 05ae26d is a rate-limit notice ("Review limit reached") while its status check reads SUCCESS. cubic reviewed 05ae26d; its one open finding is valid and confirmed inline below.

ledgerstate/certstate.go
```diff
@@ -2360,49 +2370,62 @@ func parseProposals(data []byte) (
...
 			"RatifyState has %d elements, expected 4",
 			len(ratifyState),
 		)
 	}
+	var committee []ParsedCommitteeMember
+	var quorum *cbor.Rat
+	committeeSet := false
+	var errs []error
+	enactFields, enactErr := decodeRawArray(ratifyState[0])
+	if enactErr != nil {
```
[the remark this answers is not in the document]
chrisguiney: Partially addressed on this head, not fully. `9ca58cf` routes the `decodeRawArray(ratifyState[0])` failure into `committeeErr`, and `importGovState` now returns on `EnactCommitteeParseError`, so the *decode-failure* path does fail closed.

The `else if len(enactFields) > 0` branch does not: an `rsEnactState` encoded as an empty array leaves `committeeSet` false with a nil error, and the parity check is skipped silently. Same for `len(fields) == 0`, `len(data) == 0`, and a GovState truncated to fewer than 7 fields. Detail in my review comment on line 2087.

ledgerstate/certstate.go
```diff
@@ -2064,14 +2074,20 @@ func ParseGovState(
...
 			warnings = append(warnings, fmt.Errorf(
 				"parsing drep pulsing state: %w", err,
 			))
 		}
+		result.EnactCommittee = enactCommittee
+		result.EnactCommitteeQuorum = enactQuorum
+		result.EnactCommitteeSet = enactSet
+		if committeeErr != nil {
+			result.EnactCommitteeParseError = committeeErr
+		}
```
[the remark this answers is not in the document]
chrisguiney: Confirmed valid against this head. `committeeErr` is recorded only on `result`, so `ParseGovState` returns nil for input that failed to parse, where `origin/main` returned a `"parsing drep pulsing state: ..."` warning. Exported-API error contract change.

The blast radius is wider than the enact-state committee: the `decoding DRepPulsingState`, `DRepPulsingState has %d elements`, `decoding RatifyState` and `RatifyState has %d elements` returns in `parseDRepPulsingState` all place their structural error in the `committeeErr` slot too, so they are silenced the same way and are then reported to the operator as `parsing imported enact-state committee: ...`. Your suggested fix is right; the six-value two-error signature is what makes the mistake easy (review comment on line 2381).

wolf31o2:
Validation update (sanitized):

- Source: PR head `05ae26d8`.
- Mithril archive phase completed at 2026-09-06 20:55:04 EDT.
- Ledger/UTxO import reached 100% at 2026-09-07 04:47:54 EDT with 10,952,150 UTxOs.
- Run failed at 2026-09-07 05:03:11 EDT (approximately 8h50m13s elapsed) with: `governance committee disagrees between cgsCommittee and rsEnactState`.
- No OOM or killed-process condition observed. Peak measured resource usage was approximately 3.8 CPU cores and 27.8 GiB RSS.
- No serve/tip state was reached. Profiling samples were retained for the run.

This confirms the importer reaches the complete UTxO import, then fails closed on the governance committee mismatch.

wolf31o2:
`rsEnactState` is not a second view of `cgsCommittee` at the snapshot point, so requiring equality was wrong. Reference basis, all under `cardano-ledger` `eras/conway/impl/src/Cardano/Ledger/Conway`:

- `Governance/DRepPulser.hs:412,417` — `finishDRepPulser` seeds `rsEnactState = dpEnactState` and then returns `ratifyState' = runConwayRatify ...`. `EncCBOR (DRepPulsingState)` at `457` routes a still-pulsing pulser through `finishDRepPulser`, so the serialized `RatifyState` is always post-RATIFY.
- `Governance.hs:325,328` — `mkEnactState` seeds `ensCommittee` from `cgsCommittee`, which is why the two agree before the pass runs.
- `Rules/Ratify.hs:345,349` — `ratifyTransition` applies `ENACT` per accepted action and writes the result back with `rsEnactStateL .~ newEnactState`.
- `Rules/Enact.hs:106,110` — only `NoConfidence` (`ensCommitteeL .~ SNothing`) and `UpdateCommittee` (`ensCommitteeL %~ SJust . updatedCommittee ...`) rewrite `ensCommittee`.
- `Rules/Epoch.hs:297,321` — `ConwayEPOCH` reads the pulsing state and installs the result with `cgsCommitteeL .~ ensCommittee`. Were the two required to be equal, that assignment could never change the committee.
- `Governance/Internal.hs:305` — `RatifyState`'s JSON encoding names the field `nextEnactState`.

The `rsEnactState` haddock ("currently active `EnactState`") describes the seed, not the serialized value: `finishDRepPulser` runs the pass before encoding.

Semantics landed: compare `cgsCommittee` against `rsEnactState`'s committee only when `rsEnacted` carries neither `NoConfidence` nor `UpdateCommittee`, and skip the comparison when an `rsEnacted` entry cannot be decoded so the answer is unknown. A snapshot from an epoch that ratified a committee action is expected to disagree, and `importGovState` records those actions as ratified so the next boundary enacts them.

Absence case: `DRepPulsingState` (2 elements), `RatifyState` (4) and `EnactState` (7) all have fixed arity in Conway, so a missing element is malformed input rather than an optional encoding. An empty `cgsDRepPulsingState`, a short `RatifyState` and an empty `rsEnactState` are now fatal instead of leaving `EnactCommitteeSet` false. `len(fields) < 7` on `ConwayGovState` stays a warning: enforcing the record's arity invalidates several existing three-field fixtures and belongs with GovState shape validation rather than with this check. The decision is recorded in the code.

Fails-without-fix, with only the gate condition reverted to the previous one:

```
--- FAIL: TestImportGovStateAcceptsCommitteeChangeInRsEnacted
    governance committee disagrees between cgsCommittee and rsEnactState
--- FAIL: TestImportGovStateAcceptsNoConfidenceInRsEnacted
    governance committee disagrees between cgsCommittee and rsEnactState
```

That is the mainnet bootstrap failure. `TestImportGovStateRejectsAbsentEnactState` likewise fails ("An error is expected but got nil") with only the empty-`rsEnactState` guard reverted. `TestImportGovStateAcceptsCommitteeChangeInRsEnacted` asserts the persisted member is `cgsCommittee`'s, not `rsEnactState`'s, and `TestImportGovStateRejectsMismatchWithNonCommitteeRsEnacted` keeps the check armed for an rsEnacted that cannot touch the committee.

Local runs on this head: `go test ./ledgerstate/ -count=1 -timeout 40m` and `go test ./ledger/... -count=1 -timeout 60m` both exit 0; `golangci-lint run ./ledgerstate/...` reports 0 issues, including the `QF1008` failure. No mainnet or devnet import was attempted.

Adjacent, pre-existing and out of scope for this PR: the importer calls `SetCommitteeQuorum` on the sqlstore directly, bypassing the `quorum.Sign() <= 0` guard in `database/committee.go`, and it never clears a stale quorum row when the imported committee is absent.

ledgerstate/import.go
wolf31o2: Confirmed and fixed in e9a6081. The premise was wrong: `finishDRepPulser` seeds `rsEnactState` from `mkEnactState` and then runs `runConwayRatify` before the pulsing state is encoded (`DRepPulser.hs:412,417`, `EncCBOR` at `457`), `ratifyTransition` writes each `ENACT` result back with `rsEnactStateL .~ newEnactState` (`Ratify.hs:345,349`), and `ConwayEPOCH` installs it with `cgsCommitteeL .~ ensCommittee` (`Epoch.hs:321`).

The comparison is now gated on `rsEnacted` carrying neither `NoConfidence` nor `UpdateCommittee` — the only two actions that rewrite `ensCommittee` (`Enact.hs:106,110`) — and is skipped when an `rsEnacted` entry will not decode, so the answer is unknown rather than assumed. `TestImportGovStateAcceptsCommitteeChangeInRsEnacted` and `TestImportGovStateAcceptsNoConfidenceInRsEnacted` fail with the exact mainnet message when only the gate condition is reverted.

ledgerstate/import.go
wolf31o2: Fixed in e9a6081, using the promoted method as suggested. `golangci-lint run ./ledgerstate/...` reports 0 issues on this head. The nil-`Rat` case on either side is now guarded explicitly rather than relying on `parseCommittee` never returning one.

ledgerstate/certstate.go
wolf31o2: Three of the four shapes now fail closed in e9a6081: an empty `cgsDRepPulsingState`, a `cgsDRepPulsingState` shorter than two elements, and an `rsEnactState` encoded as an empty array all set the fail-closed committee error. Basis: `DRepPulsingState` always encodes two elements, `RatifyState` four and `EnactState` seven, so a missing element is malformed input, not an optional encoding. `TestImportGovStateRejectsAbsentEnactState` covers the empty `rsEnactState`.

The fourth, `len(fields) < 7` in `ParseGovState`, stays a warning by explicit decision, recorded in a comment at that branch: enforcing `ConwayGovState`ʼs arity invalidates several existing three-field fixtures in this package and belongs with GovState shape validation rather than with the committee corroboration.

ledgerstate/certstate.go
wolf31o2: Fixed in e9a6081. The committee error is appended to `warnings` and also kept in a dedicated field, now named `PulsingStateParseError`. `TestParseGovStateEnactedWarningKeepsCommittee` asserts that a malformed `rsEnacted` entry leaves that field nil while still returning an error.

ledgerstate/certstate.go
wolf31o2: Fixed in e9a6081. `parseDRepPulsingState` now returns `(parsedPulsingState, error)`, with the fail-closed committee error as a named struct field. The four structural returns that previously landed in the committee slot set that field deliberately, and `importGovState` reports it as `parsing imported governance pulsing state: ...`, so a truncated `cgsDRepPulsingState` no longer names the committee. They also reach `warnings` now.

ledgerstate/import.go
wolf31o2: Fixed in e9a6081. No line added by this branch exceeds 80 columns, checked over the full diff against `origin/main`; the seven non-test lines listed here were rewritten, and the three long constitution literals in the test now use a shared helper.

ledgerstate/certstate.go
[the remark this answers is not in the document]
wolf31o2: The `len(enactFields) > 0` branch now fails closed in e9a6081: an `rsEnactState` encoded as an empty array sets the committee error instead of leaving `committeeSet` false. Same for `len(fields) == 0` and `len(data) == 0`. `len(fields) < 7` in `ParseGovState` remains a warning by explicit decision; see the reply on line 2087.

ledgerstate/certstate.go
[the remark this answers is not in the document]
wolf31o2: Applied in e9a6081, in the suggested form: appended to `warnings` and retained in the dedicated field.

arepala-uml (approved):

ledgerstate/import.go
```diff
@@ -3158,6 +3169,30 @@ func importGovState(
...
+	// (cardano-ledger Conway/Rules/Ratify.hs, ratifyTransition), and
+	// ConwayEPOCH copies the result into cgsCommittee at the *next*
+	// boundary (Conway/Rules/Epoch.hs, "cgsCommitteeL .~ ensCommittee").
+	// Only NoConfidence and UpdateCommittee rewrite ensCommittee
+	// (Conway/Rules/Enact.hs), so the two views corroborate each other
+	// exactly when rsEnacted carries neither. A snapshot from an epoch
+	// that ratified a committee action is expected to disagree, and the
+	// importer records those actions as ratified below so the next
+	// boundary enacts them.
+	if govState.EnactCommitteeSet &&
```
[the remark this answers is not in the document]
arepala-uml: Re-verified against the current head (8be5a9e). This does not reproduce: every early-return path in parseDRepPulsingState that leaves EnactCommitteeSet false also sets CommitteeErr, which ParseGovState surfaces as PulsingStateParseError, and importGovState returns fatal on that before reaching this guard (import.go:3158-3163, checked ahead of the line quoted here). A missing cgsDRepPulsingState (len(fields)<7) and a truncated rsEnactState (len(enactFields)!=7) both hit that fatal path directly. Classifying as a false positive at this head; the 8be5a9e commit that reworked the field-count checks to exact-match (!=7, !=4) closed this without a matching diff line for cubic's tracker to key off of.

arepala-uml (approved):

PR #4032, opened 2026-09-06
fix(ledger): guard bufferedHeaderEvents with the blockfetch mutex

wcatz:
## Problem

`bufferedHeaderEvents` is read and written from two goroutines under two different mutexes.

- **Blockfetch goroutine** — `handleEventBlockfetch` takes `chainsyncBlockfetchMutex` (`ledger/chainsync.go:787`) and holds it across the whole batch-done path, during which `nextBufferedHeaderConnId` **ranges over** `bufferedHeaderEvents`.
- **Chainsync dispatch goroutine** — `handleEventChainsync` takes `chainsyncMutex` (`ledger/chainsync.go:249`), a *different* lock, then calls `discardBufferedPeerHeaders`, which **deleted** from that same map *before* acquiring `chainsyncBlockfetchMutex`.

`resetChainsyncResyncState` had the same shape: `ls.bufferedHeaderEvents = nil` executed before the lock it takes two lines later.

That is a concurrent map iteration and map write, which the Go runtime treats as **fatal**, not merely racy. It was observed killing a mainnet block producer inside `nextBufferedHeaderConnId`:

```
fatal error: concurrent map iteration and map write
```

The existing comment above `discardBufferedPeerHeaders` already documented that this function "holds only chainsyncMutex" and reasoned about `headerPipelineConnId` needing the blockfetch mutex — but `bufferedHeaderEvents`, mutated on the line just above, has identical exposure and was missed.

## Change

Move both mutations inside the critical sections that already exist in those functions. Each function takes `chainsyncBlockfetchMutex` a line or two later anyway, so this **widens no critical section** — it only orders the map write correctly against the reader. Comments updated to record why the map belongs inside the lock.

## Tests

`TestDiscardBufferedPeerHeadersDoesNotRaceBufferedHeaderIteration` runs the two real paths concurrently: the read side takes `chainsyncBlockfetchMutex` and calls `nextBufferedHeaderConnId` (mirroring `handleEventBlockfetch`), while the write side calls `discardBufferedPeerHeaders` (mirroring the chainsync dispatch goroutine), across 50 buffered connections so the iteration and the deletes genuinely overlap.

Verified it fails without the fix: reverting only the `discardBufferedPeerHeaders` ordering produces `WARNING: DATA RACE` and a test failure under `-race`; restoring it passes.

This is deliberately distinct from the existing `TestShouldBufferHeaderEventDoesNotRaceDiscardBufferedPeerHeaders`, which guards `headerPipelineConnId` — a different field, and one that fix did not cover.

`go build ./...`, `gofmt`, and `go test -race ./ledger/...` are clean with this change.

Note: `ledger/leader`'s `TestThresholdPreservesProtocolBoundary` fails on `main` independently of this PR — I confirmed the identical failure on a clean `origin/main` checkout, so it is pre-existing and out of scope here.

---
## Summary by cubic
Fixes a fatal data race on `bufferedHeaderEvents` that could kill a mainnet block producer. The map was iterated under `chainsyncBlockfetchMutex` while mutated under `chainsyncMutex` or with no lock at all; a dedicated `bufferedHeaderMutex` now guards every access and is acquired last when both locks are held.

Adds `TestDiscardBufferedPeerHeadersDoesNotRaceBufferedHeaderIteration`, which runs the iteration, discard, buffering, and resync paths concurrently under `-race`.

<sup>Written for commit a342b1dfa8179a84551993888df982081e79e068. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4032?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization for chain synchronization and blockfetch state updates.
  * Prevented potential concurrency issues when buffered headers are read, cleared, or discarded.

* **Tests**
  * Added coverage to verify safe concurrent handling of buffered header data.

### DIFF
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -2080,12 +2080,17 @@ func (ls *LedgerState) replayBufferedHeaderEvents(
 // chainsyncBlockfetchMutex. Taking it here too, self-contained, closes that
 // gap without widening handleEventBlockfetch's own critical section to
 // cover chainsyncMutex as well.
+//
+// The bufferedHeaderEvents delete belongs inside that lock for the same
+// reason: handleEventBlockfetch holds chainsyncBlockfetchMutex while
+// nextBufferedHeaderConnId ranges over the map, so deleting from this
+// goroutine under a different mutex is a concurrent iteration and write.
 func (ls *LedgerState) discardBufferedPeerHeaders(
 	connId ouroboros.ConnectionId,
 ) {
-	delete(ls.bufferedHeaderEvents, connIdKey(connId))
 	ls.chainsyncBlockfetchMutex.Lock()
 	defer ls.chainsyncBlockfetchMutex.Unlock()
+	delete(ls.bufferedHeaderEvents, connIdKey(connId))
 	if sameConnectionId(ls.headerPipelineConnId, connId) {
 		ls.clearQueuedHeaders()
 	}
@@ -2554,13 +2559,15 @@ func (ls *LedgerState) clearRollbackHistoryForPoint(point ocommon.Point) {
 func (ls *LedgerState) resetChainsyncResyncState() {
 	ls.rollbackHistory = nil
 	ls.headerMismatchCount = 0
-	ls.bufferedHeaderEvents = nil
 	ls.selectedBlockfetchConnId = ouroboros.ConnectionId{}
 	ls.chainsyncBlockfetchMutex.Lock()
-	// clearQueuedHeaders mutates headerPipelineConnId, which every other
-	// mutator guards with chainsyncBlockfetchMutex -- moved inside this
-	// lock (rather than called before it, as this used to) to close that
-	// gap.
+	// bufferedHeaderEvents is read under chainsyncBlockfetchMutex (see
+	// nextBufferedHeaderConnId, which ranges over it), so clearing it must
+	// happen inside this lock rather than before it. clearQueuedHeaders
+	// mutates headerPipelineConnId, which every other mutator guards with
+	// chainsyncBlockfetchMutex -- moved inside this lock (rather than
+	// called before it, as this used to) to close that gap.
+	ls.bufferedHeaderEvents = nil
 	ls.clearQueuedHeaders()
 	ls.blockfetchRequestRangeCleanup()
 	ls.activeBlockfetchConnId = ouroboros.ConnectionId{}
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -1020,7 +1020,9 @@ func (ls *LedgerState) handleConnectionClosedEvent(evt event.Event) {
 	if sameConnectionId(ls.shadowBlockfetchConnId, e.ConnectionId) {
 		ls.shadowBlockfetchConnId = ouroboros.ConnectionId{}
 	}
+	ls.bufferedHeaderMutex.Lock()
 	delete(ls.bufferedHeaderEvents, connIdKey(e.ConnectionId))
+	ls.bufferedHeaderMutex.Unlock()
 	delete(ls.peerHeaderHistory, connIdKey(e.ConnectionId))
 	// Cancel in-flight blockfetch if the dead connection owns it.
 	// Without this, chainsyncBlockfetchReadyChan stays non-nil and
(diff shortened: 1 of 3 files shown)


wcatz:
cubic's P1 is correct and I've confirmed it empirically — converting to draft while I widen the fix.

Reproduced directly: a goroutine calling `bufferHeaderEvent` concurrently with `nextBufferedHeaderConnId` under `chainsyncBlockfetchMutex` produces `WARNING: DATA RACE` **with this PR's change already applied**. So moving the two mutations was necessary but not sufficient.

It is reachable in production: `bufferHeaderEvent` <- `shouldBufferHeaderEvent` <- `handleEventChainsyncBlockHeaderWithPending`, on the chainsync dispatch goroutine holding only `chainsyncMutex`. `claimHeaderPipelineOwnership` takes `chainsyncBlockfetchMutex` with a deferred unlock, so it is already released by the time `bufferHeaderEvent` writes the map.

Full audit of every `bufferedHeaderEvents` access:

| Site | Function | Guarded |
|---|---|---|
| 1023 | `handleConnectionClosedEvent` | yes, takes both mutexes |
| 1224, 1310 | `...Locked` helpers | yes, caller holds it |
| 1358-1378 | `bufferHeaderEvent` | **no** |
| 1845 | `requestChainsyncResync` | **mixed** - some callers hold it, some do not |
| 1990-1999 | `nextBufferedHeaderConnId` | yes, caller holds it |
| 2060-2067 | `replayBufferedHeaderEvents` | **no** - runs in its own goroutine |
| 2086, 2557 | this PR's two changes | yes |

The mixed caller state at 1845 is why simply taking `chainsyncBlockfetchMutex` at each remaining site risks a double lock. I'll take cubic's second suggestion — a dedicated mutex for the map, acquired consistently after `chainsyncBlockfetchMutex` so the lock order stays fixed — and extend the test to cover the buffer and replay paths, not just discard.

wcatz:
Widened in b09e8b68 — ready for review again.

`bufferedHeaderEvents` now has its own `bufferedHeaderMutex`, and every access is guarded:

| Site | Function | Before | Now |
|---|---|---|---|
| 1023 | `handleConnectionClosedEvent` | held both mutexes | + bufferedHeaderMutex |
| 1224 | `handoffPipelineOnSwitchLocked` | via caller | explicit |
| 1310 | `chainSwitchNeedsFreshCursorLocked` | via caller | explicit |
| 1358-1384 | `bufferHeaderEvent` | **unguarded** | guarded |
| 1845 | `requestChainsyncResync` | **mixed callers** | guarded |
| 1990-2007 | `nextBufferedHeaderConnId` | via caller | explicit |
| 2073-2081 | `replayBufferedHeaderEvents` | **unguarded** | guarded |
| 2109 | `discardBufferedPeerHeaders` | unguarded | guarded |
| 2586 | `resetChainsyncResyncState` | unguarded | guarded |

Why a dedicated mutex rather than reusing `chainsyncBlockfetchMutex`: `sync.Mutex` is not reentrant, and several `requestChainsyncResync` callers (3473, 4174, 4284, 4433, 6831, 6875) plus the replay path already hold the blockfetch mutex, so acquiring it at those sites would deadlock. The dedicated lock is acquired last where both are held, and nothing calls out to other ledger code while holding it, so the order is fixed. `replayBufferedHeaderEvents` takes its events and clears the entry under the lock, then replays outside it, because the replay calls back into the buffering path.

Verification:
- The test now drives four paths concurrently: guarded iteration, discard, buffer, and resync.
- It **fails under -race with only the previous commit's narrower change applied** (bufferHeaderEvent's lock removed) — so it genuinely covers the gap cubic identified, not just the original one.
- Full `go test -race ./ledger/` passes in 285s, confirming no deadlock from the new lock.

Also checked `peerHeaderHistory`, which is deleted alongside this map in several places: it is consistently accessed under `chainsyncMutex` only (writes 260, 1024, 1402-1413; reads 1523, 1627, 2904), so it has no equivalent split-mutex exposure and is untouched here.

wcatz:
Fixed in a342b1df — the comment described the pre-fix state and contradicted the code after the widening. It now says what is actually true: `bufferHeaderEvent` is protected by `bufferedHeaderMutex` alone and never holds `chainsyncBlockfetchMutex`, which is exactly why the earlier narrower fix missed it.

wolf31o2 (approved):
Reviewed the refreshed current head; no merge-blocking findings.

wcatz:

```
```

PR #3910, opened 2026-09-04
fix(ledger): persist rollback undo payload before chain truncation

wolf31o2:
Closes #3817

Rollback undo notifications now persist a versioned, bounded outbox containing the rollback point and complete block payloads before chain truncation. Startup recovery replays the payload and completes metadata rollback after truncation failures. Consecutive rollback windows merge their payloads, ephemeral rollback preflights every block lookup before mutating chain state, and an ordered-delivery barrier retains the outbox until undo events have been handed to subscribers.

The regression suite covers database reopen recovery, coalesced rollbacks with a distinct regrown suffix, iterator-captured payloads after chain deletion, ahead-of-ledger rollbacks, and ordered delivery. The ahead-of-ledger test was mutation-checked by disabling its durable-tip guard: it failed on both emitted and pending-intent assertions, then passed again after restoration.

Validation:

- `go test -race ./event ./chain ./ledger -run 'Rollback|Intent|FlushOrderedContextWaitsForDelivery|ChainIterator' -count=1 -timeout=12m` — passed
- `go test -race ./ledger -run 'TestRollbackAheadOfLedgerDoesNotPersistIntent|TestRollbackUndoSurvivesMetadataTruncationFailure' -count=1 -timeout=4m` — passed
- `go test ./internal/architecture ./internal/test/conformance -count=1 -timeout=12m` — passed
- `go vet ./event ./chain ./ledger` — passed
- `golangci-lint run ./event/... ./chain/... ./ledger/...` — 0 issues
- `make docs-parity` — passed
- `make golines` — passed

Documentation:

- `ARCHITECTURE.md` documents the rollback outbox, multi-window merge, delivery barrier, and recovery behavior.
- `DATABASE.md` documents the sync-state key, schema, bounds, recovery, and at-least-once delivery semantics.

Not run: live DevNet and Preview Koios parity replay.

PR #3876, opened 2026-09-04
fix(ledger): persist rollback undo intent

wolf31o2:
## Summary

- persist rollback intent before undo notifications are emitted
- recover interrupted intents during startup
- clear the marker only after metadata rollback succeeds
- cover intent persistence and malformed records

## Validation

- `go test ./ledger -run ^ -count=1`
- `go test ./ledger -run ^TestRollbackIntent -count=1`

---
## Summary by cubic
Persists the rollback point before any undo notifications or chain rewinds so an interrupted rollback can be recovered on startup instead of leaving ledger state inconsistent.

- Applies intent persistence consistently across manual, deferred, and reconciliation-driven rollbacks.
- Clears the intent marker only after the metadata rollback succeeds.
- On startup, validates the recorded point is on the primary chain, rewinds to it, and replays rollback events.
- Fails startup on malformed or incomplete intent records instead of ignoring them.
- Extends the storage drain cleanup test timeout to reduce flakiness on slower hosts.
- Adds tests for persistence across reload and corrupt records.

<sup>Written for commit 585facb95495b19fe44e658bcc4d78143fa4cb86. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/3876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ledger recovery after an interrupted rollback.
  - Rollback progress is now preserved across restarts, allowing incomplete rollbacks to resume automatically.
  - Added safeguards to detect and reject corrupted rollback recovery data.
  - Ensured rollback state is cleared after successful completion.

PR #4034, opened 2026-09-06
fix(chain): bar rollback from uncommitted caller-transaction indices

### LINKED ISSUE #4005: chain: addBlockInternal leaves the commit-lag window open for caller-supplied transactions
PR #4002 closes the window where `c.tipBlockIndex` advances under `c.mutex` while `txn.Do` has not yet committed, so a rollback resolves a tip index the store cannot serve. It does so with `Chain.batchCommitMutex`, held for read by `addRawBlocks`/`AddBlocks` from before they mutate memory until their transaction concludes, and for write by the two paths that delete blocks by index.

That covers batches whose transaction the chain performs itself. **It does not cover `addBlockInternal`.**

`AddBlock` and `AddLocalBlock` reach `addBlockInternal` with a **caller-supplied** transaction — from blockfetch and from `ledger` — whose commit the chain neither performs nor observes. The same window is therefore still open on those paths: memory advances, the caller commits later, and a rollback in between resolves an index the store has not got.

This is not hypothetical within the fixed code: it is how `TestSkippedBatchRestoreIsRecorded` reaches the skip branch at all, since `addBlockInternal` does not take the barrier.

## Why it was not closed in #4002

Holding the barrier across a caller-supplied transaction needs a completion hook `database.Txn` does not have. `AfterCommit` fires only on success, so a rolled-back transaction would strand the barrier held and deadlock every subsequent rollback — strictly worse than the window.

## Acceptance criteria

- `database.Txn` exposes a completion signal that fires on both commit and rollback, or `addBlockInternal` acquires and releases the barrier around a transaction boundary it can observe.
- A rollback concurrent with an `AddBlock` whose caller-supplied transaction has not yet committed does not resolve an uncommitted index. A test asserts it, failing without the change.
- The failure mode if the hook is missed is a stranded barrier, so the test must cover the rollback path as well as the commit path.

Found while fixing #3970 and #3979 (PR #4002); scoped out there deliberately rather than left unstated.

wolf31o2:
`Chain.addBlockLocked` can advance the in-memory tip through a caller-owned
transaction before that transaction commits its block. A concurrent rollback
then resolves an index the store cannot yet serve and fails with `block not
found`.

## Changes

- Add `database.Txn.OnFinish` for lifetime callbacks that run after every
  terminal transaction path. Registrations made after finish run immediately;
  nested registrations can therefore precede callbacks already dequeued.
- Track persistent-chain adds that use caller-owned transactions and restore
  their in-memory block state when the transaction aborts.
- Take the batch exclusion for every add attempt sharing a transaction, so a
  failed attempt cannot drop another attempt's hold or snapshot.
- Wait for outstanding caller transactions before rollback or standalone block
  mutation. A 30-second expiry aborts the operation; later operations fail
  immediately until the abandoned transaction concludes.
- Keep transaction-owned chain events in mutation order without publishing
  them before commit. Commit releases them; abort removes them with the
  restored state.
- Track queued-header mutations separately from block-tip mutations, preserving
  concurrent header additions and clears without changing fresh-chain origin
  semantics.
- Preserve reconciliation and the security-parameter configuration gate on the
  queued-header rollback fast path.

Every in-tree add passes a nil transaction, so the live blockfetch and forging
paths do not create pending-add holds.

## Validation

- Race-enabled chain and database test suites
- Focused regression tests for transaction event publication, concurrent add
  attempts, header add/clear preservation, and the queued-header K gate
- Go vet, scoped golangci-lint, documentation parity, and line formatting
- Mutation checks confirmed each new regression fails when its fix is removed

`ARCHITECTURE.md` and `DATABASE.md` are updated for the lifecycle, event, and
transaction callback contracts.

Fixes #4005


chain/pending_add_barrier.go
```diff
@@ -0,0 +1,206 @@
...
+// the last such transaction concludes, and every transaction the node opens
+// concludes in well under a second. Bounding it is what keeps the barrier's
+// worst case no worse than the bug it replaces. A caller that abandons its
+// transaction, or that rolls the chain back from inside one it has not
+// finished, would otherwise block that rollback -- and, because batchCommitMutex
+// is writer-preferring, every chain mutation queued behind it -- for the life of
+// the process. On expiry the wait logs at ERROR and proceeds, which leaves that
+// one rollback exposed to exactly the window this barrier closes and nothing
+// further.
+const pendingAddDrainTimeout = 30 * time.Second
```
chrisguiney, line 40: Non-blocking, and latent today since every in-tree add passes a nil `txn`.

On expiry the removal loop can delete part of the range before failing, which the comment's "leaves that one removal exposed to exactly the window this barrier closes and nothing further" understates. Two caller transactions are enough:

1. `T1` adds index 5 and stays open.
2. `T2` adds index 6 — the in-memory tip check passes against 5 — and commits.
3. `rollbackLocked` drains, times out on `T1`, and runs its loop: index 6 resolves (committed via `T2`) and is deleted; index 5 returns `ErrBlockNotFound` and the loop returns early.

`rollbackLocked` restores nothing on a mid-loop failure, so the chain is left claiming tip 6 with neither 5 nor 6 in the store. Without the barrier the same rollback failed on its first iteration and deleted nothing.

The mid-loop-failure class is pre-existing — any `removeBlockByIndex` error does this — and #3938 is working the adjacent atomicity, so this doesn't belong in this change. Worth narrowing the sentence here to the single-transaction case it actually describes.

database/txn.go
```diff
@@ -336,6 +345,87 @@ func (t *Txn) runAfterCommitCallback(fn func()) {
...
+}
+
+// dispatchOnFinish drains the OnFinish callbacks. Every terminal path calls it
+// after releasing t.lock, so callbacks may take locks of their own and may call
+// back into the transaction. It re-checks finished under the lock because
+// Commit's deferred cleanup also runs on the path where a provider panicked
+// mid-commit: there the transaction is not terminal yet -- the recovery rollback
+// still owns that transition -- and firing "this transaction is over" before it
+// actually is would release a hold the transaction still needs. Draining in a
+// loop picks up callbacks a callback registers. Concurrent drains take disjoint
```
chrisguiney, line 454: Non-blocking, comment-only.

"Draining in a loop picks up callbacks a callback registers" does not hold. A callback registered during the drain is never queued: `OnFinish` observes `t.finished == true` and runs it inline (txn.go:373-376). After the first pass `t.onFinish` is therefore always nil, and the `for` loop cannot reach a second iteration.

That same inline run is what breaks the "callbacks run in registration order" contract stated on `OnFinish`. It reproduces:

```go
txn.OnFinish(func() {
    order = append(order, "first")
    txn.OnFinish(func() { order = append(order, "nested") })
})
txn.OnFinish(func() { order = append(order, "second") })
txn.Rollback()
// got  [first nested second]
// doc  [first second nested]
```

`TestOnFinishCallbackMayTakeLocks` depends on the inline behaviour, so the fix is the prose rather than the code: drop the loop claim, and scope the ordering guarantee to registrations made before the transaction finished.

chrisguiney (dismissed):

chain/chain.go
```diff
@@ -1264,6 +1282,10 @@ func (c *Chain) rollbackLocked(
 	// batchCommitMutex field.
 	c.batchCommitMutex.Lock()
 	defer c.batchCommitMutex.Unlock()
+	// The write hold above excludes further caller-transaction adds; this waits
+	// for the ones already recorded, so no index the removal loop reaches is
+	// one the store has yet to be given. See pendingAddBarrier.
+	c.awaitPendingCallerAdds()
```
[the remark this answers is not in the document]
chrisguiney: Verified against the checkout: valid observation, but out of scope here. An abandoned caller transaction leaves the in-memory tip advanced on `main` too, and this change does not widen it — `TestRollbackResumesWhenCallerTransactionRollsBack` documents that the rollback still reports the index the transaction never wrote. What the barrier changes is the *committing* case, which it fixes. Restoring the tip on a caller transaction that did not commit is a separate contract (whose tip is it to restore?) and overlaps #3876 / #3910. Not a merge blocker.

database/txn.go
```diff
@@ -336,6 +345,87 @@ func (t *Txn) runAfterCommitCallback(fn func()) {
...
+// transaction nor stops the remaining callbacks from running, so one caller's
+// bug cannot strand another caller's hold.
+func (t *Txn) OnFinish(fn func()) {
+	if fn == nil {
+		return
+	}
+	t.lock.Lock()
+	if t.finished {
+		t.lock.Unlock()
+		t.runFinishCallback(fn)
```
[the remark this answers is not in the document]
chrisguiney: Confirmed empirically. A callback registered from inside a running callback is never queued — `OnFinish` sees `t.finished == true` and runs it inline (txn.go:373-376) — so it lands ahead of the already-queued ones:

```
got  [first nested second]
doc  [first second nested]
```

`TestOnFinishCallbackMayTakeLocks` depends on that inline behaviour, so the fix is the prose, not the code. Same root cause makes the `for` loop in `dispatchOnFinish` unreachable past its first pass; details in my inline comment on txn.go:390. Non-blocking.

database/txn.go
```diff
@@ -539,9 +634,18 @@ func (t *Txn) Commit() error {
...
-	return t.rollback()
+	// The unlock stays deferred: a store's own Rollback can panic, and losing
+	// the unlock to that panic would wedge every later call on this transaction
+	// -- including the recovery Rollback that Do issues. Only the dispatch moves
+	// outside the lock, because OnFinish callbacks may take locks of their own.
+	// See OnFinish.
+	err := func() error {
+		t.lock.Lock()
+		defer t.lock.Unlock()
+		return t.rollback()
```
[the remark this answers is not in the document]
chrisguiney: False positive as stated. If a provider `Rollback` panics, `finishLocked` never runs, so `t.finished` stays false and the transaction is *not* terminal — a lifetime hold must not be released there, and `dispatchOnFinish` would no-op even if it were reached. `Commit`s deferred cleanup already documents this exact case. The separate point — that a panicking blob rollback skips the metadata rollback — is pre-existing in `Txn.rollback` and unchanged by this diff. Not a merge blocker.

chain/chain.go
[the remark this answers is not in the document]
chrisguiney: Valid, and deliberate: the alternative (block indefinitely) wedges every chain mutation behind a writer-preferring `batchCommitMutex`, which is what `pendingAddDrainTimeout` is documented to avoid. One correction to the reasoning though — on expiry the loop can delete *part* of the range rather than failing cleanly, once more than one caller transaction is in flight. Concrete sequence in my inline comment on `chain/pending_add_barrier.go:38`. Non-blocking; the mid-loop-failure class is pre-existing and adjacent to #3938.

DATABASE.md
```diff
@@ -1230,6 +1230,22 @@ process's lifetime, with no later call able to repair it, and the next
 releases while the barrier's writer preference blocked every read-write `Txn`
 constructed behind it.
 
+`Txn.OnFinish` exposes that same terminal transition to callers. It registers a
+callback that runs exactly once when the transaction reaches its terminal state
+on any path -- a successful commit, a failed commit, an explicit `Rollback`, a
+`Release`, and the rollback `Commit` performs for a read-only transaction --
+dispatched from `finishLocked` without the transaction lock held, so a callback
```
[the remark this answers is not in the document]
chrisguiney: Confirmed. `finishLocked` (txn.go:149-155) only sets `finished` and releases the barrier hold and blob pin; dispatch happens in `dispatchOnFinish`, called from `Commit`s defer (txn.go:525) and `Rollback` (txn.go:647). The PR description carries the same wording. Worth correcting, because "every terminal path routes through `finishLocked`" is exactly the framing that would let a future terminal path call `finishLocked` alone and silently never dispatch. Non-blocking.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,206 @@
...
+// It is a safety valve, not a synchronisation point: the wait ends the instant
+// the last such transaction concludes, and every transaction the node opens
+// concludes in well under a second. Bounding it is what keeps the barrier's
+// worst case no worse than the bug it replaces. A caller that abandons its
+// transaction, or that rolls the chain back from inside one it has not
+// finished, would otherwise block that rollback -- and, because batchCommitMutex
+// is writer-preferring, every chain mutation queued behind it -- for the life of
+// the process. On expiry the wait logs at ERROR and proceeds, which leaves that
+// one rollback exposed to exactly the window this barrier closes and nothing
+// further.
```
chrisguiney: Non-blocking, comment accuracy — but the cost of an expiry is not confined to "that one rollback".

A hold is never evicted. `awaitDrained` returns `(heldCount, false)` on expiry and `awaitPendingCallerAdds` logs and returns; nothing deletes the entry, and `release` runs only from that transaction's `OnFinish`. A caller that never commits, rolls back or releases therefore leaves `pending` non-empty — and `b.drained` non-nil — for the life of the process, so *every* later `rollbackLocked` and `RewindPrimaryChainToPoint` blocks the full 30s before its removal loop, not just the first.

Probe against this head, barrier only, transaction deliberately abandoned:

```
await #1: outstanding=1 drained=false elapsed=51ms
await #2: outstanding=1 drained=false elapsed=51ms
await #3: outstanding=1 drained=false elapsed=51ms
```

Latent today, since every in-tree add passes a nil `txn`. But `AddBlock(block, txn)` is exported, and for an embedder that leaks one transaction the failure mode is a permanent 30s tax on every fork switch rather than a single exposed removal. Either narrow this sentence to the per-rollback correctness claim it actually supports, or drop the expired entry so the penalty is paid once.

chrisguiney (commented):
Independent re-review at 8a9d940b. No blockers; the approval on this head stands. One non-blocking finding inline.

Verified rather than taken from the description:

- Fail-before, `c.awaitPendingCallerAdds()` reverted in place at `chain/chain.go:1288`: `TestRollbackWaitsForUncommittedCallerTransaction` fails with `remove block at index 5: block not found` — the exact symptom in #4005.
- Fail-before, `OnFinish` swapped for `AfterCommit` in `beginCallerTxnAdd`: `TestRollbackResumesWhenCallerTransactionRollsBack` (15s) and `TestCallerTxnHoldOutlivesTheAdd/rollback` (`rollback left 1 holds outstanding`) both fail. The hook choice is load-bearing and pinned.
- `finishLocked` class audit: all eight call sites are inside `Commit` or `rollback`. `Commit`'s defer and `Rollback` both dispatch, and `Release`/`Do` route through `Rollback`, so no terminal path sets `finished` without dispatching.
- Lock order: `endAdd` is deferred before `c.mutex.Unlock()` and so unwinds after it, giving `batchCommitMutex` → `Chain.mutex` on the add path, matching `rollbackLocked`. `addBlockInternal` is reached only from the four exported `Add*` wrappers, none of which already hold `batchCommitMutex`, so the new `RLock` cannot recurse.
- `c.persistent` is assigned only at construction (`chain/manager.go:199,276,407`), so the `!c.persistent` skip cannot go stale mid-add.
- Every in-tree add passes a nil `txn`: `ledger/state.go:10970`, `ledger/chainsync.go:4671`, `node_forging.go:508`.

`go build ./...`, `go vet ./...`, `golangci-lint run ./chain/... ./database/...` (0 issues, isolated cache) and `go test -race ./chain/... ./database/...` all pass. The two `gofmt` findings (`bark/blob_test.go`, `ledger/read_chain_pipeline_test.go`) reproduce on `origin/main` and are untouched here.

Siblings: #3952 refuses the rollback in the ledger before `rollbackLocked` runs and #3910 persists its undo payload on the same side of this wait, so neither disagrees with the barrier. The branch sits directly on current `origin/main`, so the merged tree is the head tree.

chain/pending_add_barrier.go
wolf31o2: The behaviour this concerns has changed. `awaitPendingCallerAdds` now returns an error on expiry and `rollbackLocked` propagates it before the removal loop (`chain/chain.go:1552-1554`), so an expired drain no longer reaches a partial-range deletion.

database/txn.go
wolf31o2: Still accurate at this head. `OnFinish` runs a callback inline when `t.finished` is already set (`database/txn.go:554-567`), so a callback registered during a drain is never queued and `dispatchOnFinish`'s loop (`database/txn.go:578-597`) cannot reach a second iteration. The registration-order wording at `database/txn.go:416` and `database/txn.go:537` does not hold for that nested case.

chain/pending_add_barrier.go
wolf31o2: Confirmed against this head, and unchanged: no path evicts a hold when the drain expires. `awaitDrained` returns `(heldCount, false)` and leaves `pending` populated (`chain/pending_add_barrier.go:174-189`), and `release` runs only from the transaction's `OnFinish`. Recorded on the barrier-lifecycle thread as an open finding.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,206 @@
...
+		return
+	}
+	slog.Default().Error(
+		"proceeding with block removal while caller-supplied add transactions are still open",
+		"component", "chain",
+		"chain_id", c.id,
+		"outstanding_transactions", outstanding,
+		"timeout", pendingAddDrainTimeout.String(),
+	)
+}
```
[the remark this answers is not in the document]
wolf31o2: Already addressed at this head. `awaitPendingCallerAdds` returns an error on expiry rather than logging and continuing (`chain/pending_add_barrier.go:294-311`), and `rollbackLocked` propagates it at `chain/chain.go:1552-1554`, before the removal loop is reached. `beginStandaloneAdd` aborts on the same error at `chain/pending_add_barrier.go:231-238`.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,288 @@
...
+func (c *Chain) finishCallerTxnAdd(txn *database.Txn) {
+	adds := c.pendingAdds.adds(txn)
+	if txn.IsCommitted() || len(adds) == 0 {
+		c.pendingAdds.release(txn)
+		return
+	}
+	c.mutex.Lock()
+	for i := len(adds) - 1; i >= 0; i-- {
+		add := adds[i]
+		if c.tipBlockIndex != add.tipIndex+1 || c.mutationGeneration != add.generation+1 {
```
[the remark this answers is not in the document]
wolf31o2: The interleaving named here cannot occur for a nil-transaction add. On a persistent chain `addBlockInternal` routes every nil-transaction add through `beginStandaloneAdd` (`chain/chain.go:568-575`), which takes `batchCommitMutex` for write and drains the barrier before the add proceeds, so a standalone add cannot land between a caller-transaction add and that transaction's rollback.

The variant that remains is two distinct caller transactions, which hold `batchCommitMutex` for read concurrently. That is the restoration-contract question already recorded on this pull request as separate from the barrier, and it is not reachable in tree: no caller passes a transaction to `AddBlock`.

chain/chain.go
```diff
@@ -406,6 +424,7 @@ func (c *Chain) addBlockInternal(
 	if err := c.reconcile(); err != nil {
 		return event.Event{}, fmt.Errorf("reconcile chain: %w", err)
 	}
+	c.recordCallerTxnAdd(txn)
```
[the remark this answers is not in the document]
wolf31o2: Already addressed. `addBlockInternal` calls `c.pendingAdds.discardLast(txn)` on the add-failure path (`chain/chain.go:604-606`), so a rejected attempt leaves no snapshot behind and `finishCallerTxnAdd` does not break at a no-op entry.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil
+	}
+	slog.Default().Error(
+		"aborting rollback while caller-supplied add transactions are still open",
+		"component", "chain",
+		"chain_id", c.id,
+		"outstanding_transactions", outstanding,
+		"timeout", pendingAddDrainTimeout.String(),
+	)
+	return fmt.Errorf(
```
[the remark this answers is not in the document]
wolf31o2: Already addressed. `rollbackLocked` performs the queued-header check before the wait (`chain/chain.go:1516-1548`) and returns without calling `awaitPendingCallerAdds` when the rollback point is a queued header. The wait follows at `chain/chain.go:1552`, and the full header check is repeated after it because headers can change while a caller transaction concludes.

chain/chain.go
```diff
@@ -1285,6 +1314,30 @@ func (c *Chain) rollbackLocked(
...
+	// can change while a caller transaction concludes.
+	c.mutex.Lock()
+	if len(c.headers) > 0 {
+		idx, err := c.findQueuedHeader(point)
+		if err != nil {
+			c.mutex.Unlock()
+			return nil, err
+		}
+		if idx >= 0 {
+			c.headers = slices.Delete(c.headers, idx+1, len(c.headers))
```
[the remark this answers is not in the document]
wolf31o2: Confirmed against this head, and still open.

Neither header-only rollback path bumps `c.mutationGeneration`: the pre-wait fast path at `chain/chain.go:1516-1548` and the post-wait path at `chain/chain.go:1567-1595` both delete headers and return, while only the block-removal path bumps it (`chain/chain.go:1702`).

`finishCallerTxnAdd` guards restoration on `c.tipBlockIndex` and `c.mutationGeneration` alone (`chain/pending_add_barrier.go:267`). A header-only rollback leaves both unchanged, so the guard passes and `c.headers = add.headers` (`chain/pending_add_barrier.go:280`) reinstates the headers the rollback deleted. The post-wait path has the same exposure, so the cause is the guard rather than the fast path.

Reachable only through a caller-supplied transaction; no in-tree caller passes one.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,311 @@
...
+	return c.batchCommitMutex.RUnlock
+}
+
+// beginStandaloneAdd prevents a standalone add from being built on a
+// caller-owned block that may later roll back. The write side remains held
+// through the add, so a caller add cannot interpose between the wait and the
+// standalone add's c.mutex acquisition.
+func (c *Chain) beginStandaloneAdd() (func(), error) {
+	c.batchCommitMutex.Lock()
+	if err := c.awaitPendingCallerAdds(); err != nil {
```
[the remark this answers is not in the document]
wolf31o2: Confirmed against this head, and still open. A hold is dropped only by `release`, which runs from the transaction's `OnFinish` (`chain/pending_add_barrier.go:151-162` and `221-223`). Nothing evicts an entry when the drain expires: `awaitDrained` returns `(heldCount, false)` and leaves `pending` populated and `drained` non-nil (`chain/pending_add_barrier.go:174-189`). A caller transaction that never concludes therefore makes every later `beginStandaloneAdd` and `rollbackLocked` wait the full `pendingAddDrainTimeout` and then fail, for the life of the process.

chain/chain.go
```diff
@@ -414,6 +442,7 @@ func (c *Chain) addBlockInternal(
 		matchPendingHeader,
 	)
 	if err != nil {
+		c.pendingAdds.discardLast(txn)
```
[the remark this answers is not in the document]
wolf31o2: Confirmed against this head, and still open. `beginCallerTxnAdd` takes the hold before the add runs (`chain/pending_add_barrier.go:221-223`); on add failure `addBlockInternal` calls `discardLast` (`chain/chain.go:604-606`), which drops the snapshot but leaves the entry in `pending`. A rollback then waits up to `pendingAddDrainTimeout` for a transaction that carries no pending add.

chain/rollback_caller_txn_test.go
```diff
@@ -0,0 +1,234 @@
...
+func TestStandaloneAddDoesNotBuildOnAbortedCallerAdd(t *testing.T) {
+	db, c := callerTxnChain(t)
+	txn := addOnCallerTxn(t, db, c)
+
+	result := make(chan error, 1)
+	go func() { result <- c.AddBlock(testBlocks[5], nil) }()
+	select {
+	case err := <-result:
+		t.Fatalf("dependent standalone add completed before caller transaction: %v", err)
+	case <-time.After(500 * time.Millisecond):
```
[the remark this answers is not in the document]
wolf31o2: Valid. The negative check at `chain/rollback_caller_txn_test.go:196-201` is a fixed 500ms window on a bare `time.After`, so it holds whether or not the standalone add ever reached `beginStandaloneAdd`. `internal/test/testutil` provides `RequireNoReceive` for this shape.

Class audit: the same construction appears a second time at `chain/rollback_caller_txn_test.go:115-122`, asserting that a rollback is still waiting. Both belong to the finding.

chain/pending_add_barrier.go
wolf31o2: Confirmed, and taken as the second option rather than the first, with one correction to it: the expired entry is not dropped. The restoration snapshots it carries are owed to that transaction whenever it concludes, and dropping them would leave the chain naming a block the store never received. A hold that outlives one wait is marked instead, and `awaitDrained` fails later waits immediately while the mark stands, so the penalty is paid once and the sentence now supports what it claims. The mark is cleared with the hold, so a transaction that concludes late returns the barrier to normal.

The probe result also showed a second defect in the same function: `awaitDrained` returned `(0, false)` when the set drained between the timer firing and the lock, reporting a failure for a wait that had in fact succeeded. That case now reports the drain.

`pendingAddDrainTimeout`'s comment and the barrier section of ARCHITECTURE.md now state the once-per-transaction rule rather than leaving it to be inferred.

chain/chain.go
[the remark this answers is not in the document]
wolf31o2: Confirmed, and fixed at two causes in 47ab160c.

The restore guard could not see the trim. `mutationGeneration` is the counter both in-memory restore paths test, and the queued-header list is one of the fields they snapshot, but the header-only mutations never bumped it: `addBlockHeader`, `ClearHeaders`, and both of `rollbackLocked`'s queued-header paths. They do now, so a header mutation landing between a caller add and that transaction concluding suppresses the restore instead of being silently overwritten by it. `batchRestoreIsSafeLocked` was exposed to the same gap and is closed by the same change.

The fast path also should not have run there at all. It answers without waiting because it removes no persistent block, but it does trim the very queue `finishCallerTxnAdd` restores, so it is now taken only while no caller add is outstanding; otherwise the rollback falls through to `awaitPendingCallerAdds` and re-checks the header queue afterwards. Every in-tree add passes a nil transaction and records no hold, so the fast path is taken exactly as often as before.

`TestQueuedHeaderRollbackIsNotUndoneByACallerTransaction` drives the real path: three queued headers, an `AddBlock` on a caller transaction that consumes the first, a rollback to the next, then that transaction rolling back. With both changes reverted it fails with `queue holds 3 headers after the rollback, want 2` -- the discarded header back in the queue after its `HeaderInvalidationRollback` had been published. With only the fast-path guard reverted it fails with `queue holds 1`, so both halves carry weight.

chain/pending_add_barrier.go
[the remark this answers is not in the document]
wolf31o2: Confirmed and fixed in 47ab160c, by the fail-fast half of the suggestion rather than by cleanup. The hold is not evicted: the restoration snapshots it carries are owed to that transaction whenever it concludes, and discarding them would leave the in-memory chain naming a block the store never received. Instead a hold that outlives one wait is marked, and `awaitDrained` fails every later wait immediately while the mark stands, so the bound is charged once per abandoned transaction rather than once per standalone add or rollback. The mark is cleared with the hold, so a transaction that concludes late returns the barrier to normal.

`TestExpiredHoldIsChargedOnce` covers both directions; without the mark the second wait pays the full bound again and the test fails.

chain/chain.go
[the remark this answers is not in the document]
wolf31o2: Confirmed and fixed in 47ab160c. `discardLast` drops the hold together with the last snapshot, so a transaction whose only add was rejected carries nothing and no removal path waits for it; a transaction with earlier successful adds keeps its hold, since the trim leaves the list non-empty. A later add on the same transaction takes a fresh hold.

`TestRejectedFirstCallerAddLeavesNoBarrierHold` drives it through `AddBlock` with a queued header the offered block does not match, so the add is rejected before it mutates anything, and then asserts the rollback completes. Reverting the change fails it with `rollback waited on a caller transaction that carries no chain add`.

chain/chain.go
```diff
@@ -263,6 +284,14 @@ func (c *Chain) addBlockHeader(
...
 	// Add header
 	c.headers = append(c.headers, queued)
+	// The queued-header list is part of the state both in-memory restore
+	// paths snapshot (Chain.addRawBlocks/AddBlocks through
+	// batchRestoreIsSafeLocked, and a caller-supplied transaction through
+	// finishCallerTxnAdd), so queueing a header is a chain mutation and has
+	// to be counted as one. A restore that wrote its snapshot back over this
+	// append would drop a header the chain accepted, with no invalidation
+	// published for it. See the mutationGeneration field.
+	c.mutationGeneration++
```
chrisguiney: Merging current origin/main into this head fails three chain tests that pass on origin/main alone: TestFreshChainAcceptsFirstBlockFromAnySource, TestChainFirstVerifiedHeaderMatchesPointRequiresVerifiedHeader and TestStandaloneBlockAddsDrainQueuedHeaderEvents. atOriginAfterMutation (chain.go:251, added by #4205 for #4202) treats mutationGeneration > 0 as proof the chain was mutated, and its own doc says it ignores the header queue and exempts a never-mutated chain. Queueing one header now flips it, so a fresh chain rejects its own first block.

chain/chain.go
```diff
@@ -1754,6 +1860,9 @@ func (c *Chain) ClearHeaders() {
 	// block was ever added. Everything at or below the block tip survives;
 	// the queue held only what was above it.
 	if hadHeaders {
+		// Clearing the queue is a chain mutation; see the
+		// mutationGeneration field.
+		c.mutationGeneration++
```
chrisguiney: Same class and independently sufficient: with only the addBlockHeader bump removed, TestChainFirstVerifiedHeaderMatchesPointRequiresVerifiedHeader still fails, because ClearHeaders alone lifts mutationGeneration above zero before AddVerifiedBlockHeader runs at origin. Deleting all four new header-only bumps clears every merged-tree failure and still leaves go test ./chain/ green at this head, so nothing here pins them. The counter now has two consumers with incompatible definitions of a mutation; the header queue needs its own.

chain/chain.go
```diff
@@ -1484,6 +1534,59 @@ func (c *Chain) rollbackLocked(
...
+	// path trims the queued-header list, which is exactly the state
+	// finishCallerTxnAdd restores when a caller-supplied transaction rolls
+	// back. Taking it underneath an outstanding caller add would let that
+	// restore write the pre-add queue back over the trim, resurrecting
+	// headers this path has already published a HeaderInvalidationRollback
+	// for. So it is taken only while no such add is in flight; otherwise the
+	// rollback falls through to awaitPendingCallerAdds and re-checks the
+	// header queue afterwards. Every in-tree add passes a nil transaction and
+	// records no hold, so the fast path is taken exactly as often as before.
+	c.mutex.Lock()
```
chrisguiney, line 1853: This fast path answers before c.reconcile() and before the persistent-chain securityParam <= 0 gate that the post-wait queued-header path at chain.go:1604 sits behind, so a queued-header rollback that used to return ErrSecurityParamNotConfigured now trims the queue and returns nil. reconcile is a no-op on a persistent chain, so the reachable difference is that refusal. The comment above calls this the same check repeated after the wait, and it is not.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	return c.batchCommitMutex.RUnlock
+}
+
+// beginStandaloneAdd prevents a standalone add from being built on a
+// caller-owned block that may later roll back. The write side remains held
+// through the add, so a caller add cannot interpose between the wait and the
+// standalone add's c.mutex acquisition.
+func (c *Chain) beginStandaloneAdd() (func(), error) {
+	c.batchCommitMutex.Lock()
```
chrisguiney, line 313: Measured at this head with one caller transaction left open: the first rollback fails after 30s, then AddBlock(block, nil) fails at once with 'wait for caller transaction adds: caller-supplied add transactions still open after 30s: 1 outstanding', and the next rollback fails at once too. Once a hold is marked expired every standalone add fails as well, so an abandoned transaction stops the node adding blocks at all until it concludes. ARCHITECTURE.md:4478 and the description both claim the opposite.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,381 @@
...
+func (c *Chain) beginCallerTxnAdd(txn *database.Txn) func() {
+	// A nil transaction leaves the store write to the chain, which commits it
+	// before the tip advances. A non-persistent chain writes to the manager's
+	// block cache rather than the store, and its rollback deletes from that
+	// same cache, so neither has a window to close.
+	if txn == nil || !c.persistent {
+		return func() {}
+	}
+	if c.pendingAdds.holds(txn) {
+		return func() {}
```
chrisguiney: Two adds sharing one transaction can lose the hold. Both see holds(txn) false; one takes the read side and the hold, the other returns here without taking batchCommitMutex at all. If the first add then fails, discardLast releases the hold, after which record() for the second finds no pending[txn] entry and drops its snapshot silently. That add advances the tip with no hold and no restoration snapshot, reopening the window #4005 describes. Taking the read side unconditionally and keying only the hold on holds(txn) closes it.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,381 @@
...
+		add := adds[i]
+		if c.tipBlockIndex != add.tipIndex+1 || c.mutationGeneration != add.generation+1 {
+			slog.Default().Error(
+				"skipped in-memory restore after caller transaction rollback: chain moved under the add",
+				"component", "chain",
+				"chain_id", c.id,
+				"add_tip_block_index", add.tipIndex+1,
+				"tip_block_index", c.tipBlockIndex,
+			)
+			break
```
chrisguiney, line 361: The restore rewinds tip, index, generation and headers, but nothing retracts what the undone add already queued on the chain-level sequencer: its chain.update (chain.go:635 on the deferred variant), the HeaderInvalidationLocalBlock a non-matching add publishes, and the Leios announcement. A subscriber is told the block was added and those headers voided, then the chain rewinds with no rollback event. Before this change memory and subscribers were consistently wrong; now they disagree.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,381 @@
...
+		}
+		c.currentTip = add.tip
+		c.tipBlockIndex = add.tipIndex
+		c.mutationGeneration = add.generation
+		c.headers = add.headers
+		if !c.persistent {
+			c.blocks = add.blocks
```
chrisguiney: Unreachable. recordCallerTxnAdd returns early for !c.persistent and never assigns callerTxnAdd.blocks, and beginCallerTxnAdd returns before taking a hold on the same condition, so no snapshot reaches this line; if one ever did it would set c.blocks to nil. Drop the field and this branch, or populate blocks in recordCallerTxnAdd.

ARCHITECTURE.md
```diff
@@ -4449,6 +4459,45 @@ batch left behind; otherwise a concurrent rollback's result would be overwritten
...
+transaction, and no transaction sees another's uncommitted writes. Those two
+hold `Chain.batchCommitMutex` for read across their whole `txn.Do`, and the
+removal paths hold it for write. `Chain.addBlockInternal` cannot use that shape,
+because with a caller-supplied `*database.Txn` the chain neither performs nor
+observes the commit; it instead records the transaction in `Chain.pendingAdds`
+before mutating memory and releases the record from `database.Txn.OnFinish`,
+which fires on commit *and* on rollback. The removal paths' write hold excludes
+a new record from appearing, and `awaitPendingCallerAdds` then waits for the
+records already in flight. That wait is bounded (`pendingAddDrainTimeout`) and
+logs at ERROR on expiry rather than blocking forever, so a caller that abandons
```
chrisguiney: Stale as of f3f11247 on this branch: awaitPendingCallerAdds returns an error and rollbackLocked aborts at chain.go:1552 before the removal loop, so no removal is left exposed to the original window and every chain mutation is wedged, which is what this sentence denies. The same wording was already corrected in chain/pending_add_barrier.go:37, and the paragraph directly below contradicts it. The pull request description carries the same claim.

DATABASE.md
```diff
@@ -1348,6 +1348,22 @@ process's lifetime, with no later call able to repair it, and the next
...
```
chrisguiney: finishLocked (database/txn.go:160) sets finished and releases the barrier hold and blob pin, under the lock, and does not dispatch. Dispatch is dispatchOnFinish (txn.go:578), called from Commit's deferred cleanup (txn.go:741) and from Rollback (txn.go:882). txn.go has eight finishLocked call sites and two dispatch sites; every terminal path does reach one today, so this is the wording rather than the code, but it is the framing that would let a new terminal path call finishLocked alone and strand a hold.

chrisguiney (changes requested):
Merging current origin/main into this head fails three chain tests that pass on origin/main alone, and the four new header-only mutationGeneration bumps are the sole cause while no test on this branch pins any of them. An abandoned caller transaction also fails every standalone add rather than exposing one removal, which ARCHITECTURE.md and the description both deny.

chain/chain.go
```diff
@@ -47,14 +47,25 @@ const (
...
+	blocks        []ocommon.Point
+	iterators     []*ChainIterator
+	currentTip    ochainsync.Tip
+	tipBlockIndex uint64
+	// mutationGeneration counts every mutation of the in-memory chain state
+	// that the two restore paths snapshot: currentTip, tipBlockIndex, the
+	// queued-header list, and the ephemeral block buffer. Both
+	// batchRestoreIsSafeLocked and finishCallerTxnAdd write a snapshot back
+	// only while the counter still shows the chain is exactly where that
+	// mutation left it, so a mutation that does not bump it is invisible to
```
chrisguiney: Now violated by two sites: after 16a0763 neither addBlockHeader (chain.go:454) nor ClearHeaders (chain.go:2089) bumps the counter, so finishCallerTxnAdd's guard at pending_add_barrier.go:337 cannot see them and writes the pre-add queue back over them. ClearHeaders taken while a caller transaction is open and then rolled back leaves 3 headers queued that it had published HeaderInvalidationQueueCleared for, and a header queued during one is dropped with no invalidation. Restoring either bump fixes both and re-fails the three origin tests: the header queue needs its own counter.

chain/chain.go
```diff
@@ -1732,6 +1772,56 @@ func (c *Chain) rollbackLocked(
...
+	// finishCallerTxnAdd restores when a caller-supplied transaction rolls
+	// back. Taking it underneath an outstanding caller add would let that
+	// restore write the pre-add queue back over the trim, resurrecting
+	// headers this path has already published a HeaderInvalidationRollback
+	// for. So it is taken only while no such add is in flight; otherwise the
+	// rollback falls through to awaitPendingCallerAdds and re-checks the
+	// header queue afterwards. Every in-tree add passes a nil transaction and
+	// records no hold, so the fast path is taken exactly as often as before.
+	c.mutex.Lock()
+	if len(c.headers) > 0 && c.pendingAdds.heldCount() == 0 {
```
chrisguiney, line 1854: Still open from the last review. This fast path answers before c.reconcile() and before the persistent-chain securityParam <= 0 gate at chain.go:1834. Control, persistent chain with SetLedger never called, rolling back to a queued header: origin/main returns 'chain manager security parameter K is not configured' and leaves the queue at 3 headers, while this head returns nil and trims it to 1. The comment above calls the post-wait block the same check repeated after the wait, and it is not: a refusal has become a queue mutation on an exported API.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,381 @@
...
+// drain expired.
+func (c *Chain) beginCallerTxnAdd(txn *database.Txn) func() {
+	// A nil transaction leaves the store write to the chain, which commits it
+	// before the tip advances. A non-persistent chain writes to the manager's
+	// block cache rather than the store, and its rollback deletes from that
+	// same cache, so neither has a window to close.
+	if txn == nil || !c.persistent {
+		return func() {}
+	}
+	if c.pendingAdds.holds(txn) {
```
chrisguiney: Still open from the last review; this file is unchanged since then. Two adds sharing one transaction: the second sees holds(txn) true and returns without taking batchCommitMutex or a hold. If the first add then fails, discardLast releases the hold, and record() at line 138 finds no pending[txn] entry and drops the second add's snapshot silently. That add advances the tip with no hold and no restoration snapshot, reopening the window #4005 describes. Taking the read side unconditionally and keying only the hold on holds(txn) closes it.

chain/pending_add_barrier.go
```diff
@@ -0,0 +1,381 @@
...
+			slog.Default().Error(
+				"skipped in-memory restore after caller transaction rollback: chain moved under the add",
+				"component", "chain",
+				"chain_id", c.id,
+				"add_tip_block_index", add.tipIndex+1,
+				"tip_block_index", c.tipBlockIndex,
+			)
+			break
+		}
+		c.currentTip = add.tip
```
chrisguiney, line 363: Still open from the last review. The restore rewinds tip, index, generation and headers, but nothing retracts what the undone add already queued on the chain-level sequencer: its chain.update, the HeaderInvalidationLocalBlock a non-matching add publishes, and the Leios announcement. A subscriber is told the block was added and those headers voided, and then memory rewinds with no rollback event. Before this change memory and subscribers were consistently wrong; now they disagree.

chain/pending_add_barrier.go
chrisguiney: Still open from the last review. Unreachable: recordCallerTxnAdd returns early for !c.persistent at line 313 and never assigns callerTxnAdd.blocks, and beginCallerTxnAdd returns before taking a hold on the same condition, so no snapshot reaches this line; if one ever did it would set c.blocks to nil. Drop the field and this branch, or populate blocks in recordCallerTxnAdd.

chain/caller_txn_header_restore_test.go
```diff
@@ -0,0 +1,180 @@
...
+//
+// rollbackLocked answers a rollback whose point is a queued header without
+// waiting for caller transactions, because it removes no persistent block. It
+// does trim the queued-header list, and that list is exactly what
+// finishCallerTxnAdd restores when a caller transaction concludes without
+// committing. Nothing counted the trim as a chain mutation, so the restore
+// guard that refuses to write over a moved chain did not see it, and the
+// pre-add queue went back over the trim: a header the rollback had already
+// published a HeaderInvalidationRollback for was re-queued.
+func TestQueuedHeaderRollbackIsNotUndoneByACallerTransaction(t *testing.T) {
```
chrisguiney, line 67: This pins the rollback-trim member of the class, and both of rollbackLocked's trims are covered: the pre-wait path by the heldCount() == 0 guard and the post-wait path by the drain. The other two header-queue mutations, addBlockHeader and ClearHeaders, have no equivalent coverage and both fail a probe at this head. Cleared from the last review: the merge base is current origin/main, so head is the merged tree, and the three origin tests that failed there now pass, failing again only when a removed bump is restored.

ARCHITECTURE.md
```diff
@@ -4676,6 +4686,45 @@ batch left behind; otherwise a concurrent rollback's result would be overwritten
...
+transaction, and no transaction sees another's uncommitted writes. Those two
+hold `Chain.batchCommitMutex` for read across their whole `txn.Do`, and the
+removal paths hold it for write. `Chain.addBlockInternal` cannot use that shape,
+because with a caller-supplied `*database.Txn` the chain neither performs nor
+observes the commit; it instead records the transaction in `Chain.pendingAdds`
+before mutating memory and releases the record from `database.Txn.OnFinish`,
+which fires on commit *and* on rollback. The removal paths' write hold excludes
+a new record from appearing, and `awaitPendingCallerAdds` then waits for the
+records already in flight. That wait is bounded (`pendingAddDrainTimeout`) and
+on expiry the removal path returns an error before deleting persistent blocks,
```
chrisguiney, line 4745: Corrected here in 6480c171 and accurate against the code now. The pull request description still carries the old claim, that an expired wait leaves that one removal exposed to the original window rather than wedging every chain mutation behind it, which awaitPendingCallerAdds returning an error contradicts. Worth bringing the description in line with this paragraph.

DATABASE.md
```diff
@@ -1578,6 +1578,22 @@ process's lifetime, with no later call able to repair it, and the next
 releases while the barrier's writer preference blocked every read-write `Txn`
 constructed behind it.
 
+`Txn.OnFinish` exposes that same terminal transition to callers. It registers a
+callback that runs exactly once when the transaction reaches its terminal state
+on any path -- a successful commit, a failed commit, an explicit `Rollback`, a
+`Release`, and the rollback `Commit` performs for a read-only transaction --
+dispatched by the terminal paths after the transaction lock is released, so a callback
```
chrisguiney, line 1586: Corrected here in 6480c171 and accurate now: dispatch is dispatchOnFinish, called from Commit's deferred cleanup and from Rollback, not finishLocked. The description still says OnFinish dispatches from the finishLocked choke point. Earlier findings on this file are disposed: the registration-order wording is fixed, and the panicking provider Rollback report was a false positive, because finishLocked never runs on that path so the transaction is not terminal.

chrisguiney (dismissed):
The four header-only mutationGeneration bumps are gone, which clears the merged-tree failures, but addBlockHeader and ClearHeaders are now invisible to finishCallerTxnAdd's restore guard, so the pre-add queue is written back over them; restoring either bump fixes that and re-fails the three origin tests, so the counter needs splitting rather than deleting. Four findings from the last review are unchanged and still open.

arepala-uml (approved):

PR #4035, opened 2026-09-06
fix(ledger): use the recorded deposit for DRep deregistration refunds

wcatz:
## Problem

The live `drep` row carries no `deposit_amount` — only the `registration_drep` / `deregistration_drep` history tables record it.

`ledger.LedgerView.DRepRegistration` and `DRepRegistrations` implement the `common.DRepState` that gouroboros consults when validating a `DeregistrationDrepCertificate`'s refund, and they returned a `DRepRegistration` with no `Deposit` set at all. Every DRep deregistration was therefore judged against an expected refund of **zero**, and a canonical certificate supplying the real deposit was rejected:

```
conway utxo validation rule 50: incorrect refund for certificate type 17: supplied 500000000, expected 0
```

On mainnet that stalls the tip: the block is rejected, the pipeline restarts, and the same block is rejected again without progress.

## Change

Add `GetDrepLastRegistrationDeposit`, which reads `deposit_amount` from the most recent `registration_drep` row for the credential, and wire it into both `DRepRegistration` and `DRepRegistrations`.

**The query deliberately omits `GetDrepLastRegistrationSlot`'s `certificate_id IS NOT NULL AND certificate_id != 0` filter**, and that distinction is the substance of the fix rather than an oversight. The filter is correct for the activity-display use case, which should not surface bootstrap or recovery-inserted placeholder rows. It is wrong for refund correctness: a row inserted by `InsertDrepIfAbsent` carries `certificate_id = 0` and still records the real amount owed on deregistration. Filtering it out yields an expected refund of 0 against a certificate that legitimately supplies the deposit — the exact failure above.

Copying the filter across from the sibling query is the natural mistake here, so both the SQL comment and DATABASE.md say explicitly why it is absent.

## Docs

`DATABASE.md` documents the new query in the DRep section, including the filter rationale.

## Tests

- `go build ./...`, `gofmt`, `go test ./database/... ./ledger/` clean
- `make sql-check` passes — sqlc regenerates with no drift against the checked-in output
- `make docs-parity` passes

Note: `ledger/leader`'s `TestThresholdPreservesProtocolBoundary` fails on `main` independently of this PR (confirmed identical on a clean `origin/main` checkout), so it is pre-existing and out of scope.

## Related, not overlapping

#3829 / #3918 (`preserve unknown certificate deposits`) and #3901 (`use recorded stake deposits for refunds`) are the **stake** analog of this problem — same shape, different certificate family. This PR does for DRep deposits what those do for stake deposits; it does not duplicate them.

---
## Summary by cubic
Fixes DRep deregistration refund validation by using the deposit recorded in the most recent registration certificate instead of a zero deposit, and reports that recorded deposit from `GetDRepState` instead of the current protocol parameter. This stops mainnet from rejecting valid deregistration certificates and stalling the tip.

- Adds `GetDrepLastRegistrationDeposit` for single-DRep lookups and `GetDrepLastRegistrationDeposits` so both `DRepRegistrations()` callers read all deposits in one batched query.
- Both queries deliberately omit `GetDrepLastRegistrationSlot`'s `certificate_id` filter because Mithril bootstrap-import rows (`certificate_id = 0`) still record the real deposit owed; the batched query is scoped to active DReps.
- `queryShelleyDRepState` now reads each DRep's recorded deposit instead of the governable `dRepDeposit` parameter, matching how cardano-ledger stores the amount.
- Bumps `gouroboros` to v0.204.0 to align with the deposit-state shape.
- Adds an e2e test driving a real deregistration certificate through the production Conway validation rule, plus store and dialect integration tests covering both queries.
- `DATABASE.md` documents both queries, including the filter rationale, active-set scope, and map-key semantics.

<sup>Written for commit b278daab10607c408089d65f0dc55a2a8bb6926f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4035?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - DRep registration details now include the recorded registration deposit.
  - Added support for retrieving the latest deposit for an individual DRep or all DReps at once.
  - Deposits from imported bootstrap registrations are included for accurate refund validation.

- **Bug Fixes**
  - Corrected DRep registration data that previously reported a zero deposit, improving deregistration refund accuracy.

- **Documentation**
  - Documented how DRep registration deposits are retrieved and validated.


wcatz:
Thanks — all three addressed in 782f78a5.

**N+1 query on the validation path.** `DRepRegistrations()` was issuing one `GetDrepLastRegistrationDeposit` per DRep. Replaced with a single `GetDrepLastRegistrationDeposits` that returns every DRep's latest deposit in one round trip, keyed by credential tag plus credential bytes (the same bytes can appear under both a key-hash and a script-hash credential, so the tag has to be in the key — `models.DrepDepositKey`). The set query uses a `MAX(added_slot)` group-join rather than a correlated subquery per row.

**Bootstrap-row misattribution.** You were right, and I had it wrong in both the SQL comment and DATABASE.md: I attributed the `certificate_id = 0` rows to `InsertDrepIfAbsent`. That function only writes the `drep` table, which has no `certificate_id` or `deposit_amount` column, so it cannot be the source. The rows come from `ImportDrepRegistration` (the Mithril ledger-state import), which DATABASE.md already stated correctly elsewhere — my note contradicted it. Both the comment and DATABASE.md now say `ImportDrepRegistration`.

**Conformance provider.** `internal/test/conformance/state_provider.go` now reads through the same batched call, so the conformance path and the ledger view cannot drift apart.

Verification on this branch: `go build ./...` and `go build -tags dingo_extra_plugins ./...` clean; `go test ./database/... ./ledger/` passes; `make sql-check` and `make docs-parity` pass; `golangci-lint` (host + `GOOS=windows`, all modules) reports 0 issues. `nilaway` and `modernize` findings in `ledger/` and `database/` are all outside this diff (`view.go:1169-1217` vs. my hunks at 1279+) and pre-existing on main.

One unrelated pre-existing failure, unchanged by this PR: `ledger/leader` `TestThresholdPreservesProtocolBoundary` fails on clean `origin/main` as well (verified by stashing).

wcatz:

```
```

database/plugin/metadata/domain_stores_test.go
```diff
@@ -85,6 +85,8 @@ var governanceStoreMethods = []string{
 	"GetDreps",
 	"GetPredefinedDrepFirstSeenSlots",
 	"GetDrepLastRegistrationSlot",
+	"GetDrepLastRegistrationDeposit",
+	"GetDrepLastRegistrationDeposits",
```
chrisguiney, line 89: **Blocking: nothing exercises either new query.**

These two strings are the only test change in the PR, and `governanceStoreMethods` asserts that the method exists on the interface — not that its SQL runs, or returns the right number. I grepped the tree: no test references either method anywhere else. Combined with `DRepRegistrations()` having no production caller (see the note on `ledger/view.go:1280`), `GetDrepLastRegistrationDeposits`' SQL is executed by nothing at all — not by CI, not by the node. It happens to be correct, because I ran it, but that is unverified here.

This matters most for the invariant the whole change rests on. The description says that copying the sibling's `certificate_id` filter across "is the natural mistake here" — and right now only a comment stands in the way of someone doing exactly that. A test is what stops it; a comment is not.

There is a ready home: `database/plugin/metadata/sqlite/shared_sqlstore_drep_parity_test.go` already covers `GetDrepLastRegistrationSlot` (interface at line 57, exercised at line 220) and runs the same vectors across sqlite, MySQL and Postgres. Adding the two new methods there also gets the new derived-table join executed through `dialect_queryer.go` against the non-sqlite dialects, which nothing currently does.

This ran green for me against a real store and covers the load-bearing case:

```go
credImport := make([]byte, 28)
credImport[0] = 0xAA
require.NoError(t, store.ImportDrep(
    &models.Drep{Credential: credImport, AddedSlot: 10, Active: true},
    &models.RegistrationDrep{
        DrepCredential: credImport,
        AddedSlot:      10,
        DepositAmount:  types.Uint64(500000000),
    },
    nil,
))
// certificate_id defaults to 0 here: the filter GetDrepLastRegistrationSlot
// applies would drop this row and produce an expected refund of 0.
got, err := store.GetDrepLastRegistrationDeposit(0, credImport, nil)
require.NoError(t, err)
require.Equal(t, uint64(500000000), got)

batch, err := store.GetDrepLastRegistrationDeposits(nil)
require.NoError(t, err)
require.Equal(
    t,
    uint64(500000000),
    batch[models.DrepDepositKey(0, credImport)],
)
```

database/plugin/metadata/sqlstore/drep.go
```diff
@@ -718,6 +718,39 @@ func (s *Store) GetDrepLastRegistrationSlot(
...
+	)
+	if errors.Is(err, sql.ErrNoRows) {
+		return 0, nil
+	}
+	if err != nil {
+		return 0, fmt.Errorf("get drep last registration deposit: %w", err)
+	}
+	if !raw.Valid {
+		return 0, nil
+	}
```
chrisguiney, line 746: Non-blocking, but worth a follow-up issue: this collapses three distinct states into `0`. Measured against a real sqlite store:

| state | returned |
|---|---|
| import row, `certificate_id = 0`, deposit recorded | `500000000` |
| no `registration_drep` row at all | `0` |
| row present, `deposit_amount` NULL | `0` |
| row present, `deposit_amount = '0'` | `0` |

`LedgerView.StakeCredentialDeposit` keeps these apart for the stake analog. `ledger/view.go:317-323` returns `*uint64` precisely so a NULL stays nil and the caller falls back to the current `KeyDeposit`, while a recorded zero is authoritative — its comment notes the devnet's `"keyDeposit": 0` makes that the normal case there. The DRep path has no way to express "unknown", because `common.DRepRegistration.Deposit` is a plain `uint64`.

The reachable consequence: `ledger/governance/processing.go:306` calls `InsertDrepIfAbsent(..., active=true, ...)` on the vote-replay recovery path, and that writes only the `drep` table. It leaves an **active** DRep with no `registration_drep` row, so `DRepRegistration` reports `Deposit: 0` and a later deregistration certificate supplying the real deposit is rejected with exactly the `conway utxo validation rule 50 ... expected 0` failure quoted in the description.

Not a blocker: every one of these cases already evaluated to `0` before this PR, so nothing regresses, and expressing "unknown" needs a gouroboros-side change. Better as a follow-up issue than as growth in this PR.

ledger/view.go
```diff
@@ -1267,10 +1276,20 @@ func (lv *LedgerView) DRepRegistrations() ([]lcommon.DRepRegistration, error) {
 	if err != nil {
 		return nil, fmt.Errorf("get active dreps: %w", err)
 	}
+	// One batched read rather than a deposit query per DRep: mainnet has
+	// thousands of active DReps and this is on the validation path.
```
chrisguiney: This comment is inaccurate, and it is the premise cubic's open thread on line 1281 reasons from.

`DRepRegistrations()` is not on the validation path — it has no caller at all. gouroboros v0.202.10 declares it on `common.DRepState` (`ledger/common/state.go:282`) but never invokes it; the Conway rules reach DRep state only through the singular `DRepRegistration` (`rules.go:3421`, `3867`, `4179`). `WithDRepRegistrations` in gouroboros exists solely to seed its own test mocks. Nothing in dingo calls it either.

So cubic's cost concern is latent rather than live, and its "on every validation-path call" framing is wrong. Its underlying point still stands, though: the query carries no `WHERE`, so it returns every credential's latest registration rather than just the active set `GetActiveDreps` fetched on the line above. If the batched form stays, constraining it to those credentials would make both the cost bounded and this comment true. Either way, the "this is on the validation path" justification should not ship as written.

ledger/view.go
```diff
@@ -1241,8 +1241,17 @@ func (lv *LedgerView) DRepRegistration(
 		}
 		return nil, fmt.Errorf("get drep: %w", err)
 	}
+	deposit, err := lv.ls.db.GetDrepLastRegistrationDeposit(
```
chrisguiney, line 1248: Non-blocking: unlike the plural form, this one *is* hot, and the deposit query now runs unconditionally.

Three of gouroboros's four call sites only test `reg != nil` and never read `Deposit` — `isDRepRegistered` (`rules.go:3421`), `UtxoValidateUnknownVoters` (`rules.go:4179`), and dingo's own `ledger/eras/validation.go:274`. Only `loadDRep` at `rules.go:3867` consumes it. So a transaction carrying DRep votes or vote-delegation certificates, but no DRep deregistration, now pays an extra indexed lookup per credential for a field nobody reads.

`ledger/eras/validation.go:250` already has the shape for this — the committee lookup is resolved lazily on first use, "so a transaction with only DRep or pool votes never pays for the query". `idx_drep_reg_cred_slot` makes the lookup cheap enough that this does not block, but the same lazy treatment would fit the surrounding code.

chrisguiney (dismissed):
Reviewed at 782f78a5 in an isolated worktree off `origin/main` (merge-base 7ad2791).

**The core fix is correct and I verified it end to end.** Dropping `GetDrepLastRegistrationSlot`'s `certificate_id` filter is right: `ledgerstate/import.go:1357` and `sqlstore/genesis.go:133` build `models.RegistrationDrep` without a `CertificateID`, so the import rows land with `certificate_id = 0` carrying a real `deposit_amount`. Running the new queries against a real sqlite store, an import-shaped row returns `500000000` through both the singular and the batched form. The corrected attribution to `ImportDrepRegistration` in 782f78a is accurate.

Two candidate blockers I checked and disproved:

- Update certificates write `update_drep`, not `registration_drep` (`transaction_certificates.go:974`), so `ORDER BY added_slot DESC LIMIT 1` cannot be shadowed by an anchor update.
- `registration_drep` has `UNIQUE (credential_tag, drep_credential, added_slot)`, so both the `LIMIT 1` and the `MAX(added_slot)` group-join are deterministic — no `AGENTS.md` cert-ordering tie-break is needed here.

**Class audit.** All four `DRepRegistration{...}` construction sites in the tree now set `Deposit` (`ledger/view.go:1252,1287`; `internal/test/conformance/state_provider.go:896,977`). The DRep half of the class is complete. No file or symbol overlap with #3869, #3901, #3918, or ouroboros-mock#270.

One blocking item (test coverage) and three non-blocking notes are inline. Cubic's two open threads are dispositioned inline as well.

**Checks run locally on 782f78a5** (`GOWORK=off`, isolated `GOCACHE` and `GOLANGCI_LINT_CACHE`):

| check | result |
|---|---|
| `go build ./...` and `-tags dingo_extra_plugins` | pass |
| `go vet ./...` | pass |
| `go test ./internal/docsparity ./internal/architecture` | pass |
| `go test ./internal/test/conformance/` | pass, 46.9s |
| `go test ./database/plugin/metadata/...` | pass |

Not run: `make sql-check` (needs the pinned sqlc fetch), full `make test`, devnet.

**Pipeline.** Green on CI: `lint`, `go-test (macOS)`, `Analyze (go)`, `Analyze (actions)`, CodeQL, DCO. Still queued or running: `go-test (Linux)`, `go-test (Linux, race)`, `go-test (Windows)`, `govulncheck`, `docker`, Conventional Commits. Nothing is failing; the pipeline is simply not complete yet.

**Bot review status.** The green **CodeRabbit** check is not a review — its commit status description reads `Review rate limited`, and its PR comment reports the review limit reached with the next slot in 10 minutes. Cubic did complete a review; treating Cubic alone as the bot pass is fine per the repo rules, but CodeRabbit should not be counted as having reviewed this head.

ledger/view.go
```diff
@@ -1267,10 +1276,20 @@ func (lv *LedgerView) DRepRegistrations() ([]lcommon.DRepRegistration, error) {
 	if err != nil {
 		return nil, fmt.Errorf("get active dreps: %w", err)
 	}
+	// One batched read rather than a deposit query per DRep: mainnet has
+	// thousands of active DReps and this is on the validation path.
+	deposits, err := lv.ls.db.GetDrepLastRegistrationDeposits(lv.txn)
```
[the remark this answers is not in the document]
chrisguiney: Reviewer disposition: **partly valid — the premise is wrong, the underlying point stands.**

`DRepRegistrations` is not "on the validation path". It has no caller at all. gouroboros v0.202.10 declares it on `common.DRepState` (`ledger/common/state.go:282`) but never invokes it; the Conway rules reach DRep state only via the singular `DRepRegistration` (`rules.go:3421`, `3867`, `4179`). `WithDRepRegistrations` there only seeds gouroboros's own test mocks. Nothing in dingo calls it either, so "registration history makes validation work grow without bound" does not describe current behavior.

What is correct is that the query has no `WHERE` and so returns every credential's latest registration rather than the active set `GetActiveDreps` fetched on the line above. That is a latent cost, not a live one. Not a merge blocker; if the batched form stays, constraining it to the active credentials is a reasonable cleanup — and the code comment claiming the validation path should be corrected either way.

internal/test/conformance/state_provider.go
```diff
@@ -944,8 +958,25 @@ func (p *DingoStateProvider) DRepRegistrations() ([]common.DRepRegistration, err
 	}
 	result := make([]common.DRepRegistration, 0, len(dreps))
 	for _, drep := range dreps {
+		// Report the recorded deposit here too. Production's
+		// ledger.LedgerView.DRepRegistrations does, and a vector that
+		// validates a deregistration refund through this plural view
+		// would otherwise be judged against a deposit of 0 -- passing
+		// for the same reason the bug existed.
+		deposit, err := withBadConnRetry(func() (uint64, error) {
```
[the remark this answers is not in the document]
chrisguiney: Reviewer disposition: **valid, non-blocking.**

The read-pattern divergence is real — the conformance provider keeps the per-DRep loop while `LedgerView.DRepRegistrations` batches. But this is the test harness, and its production counterpart's batched form currently has no caller either, so neither path is hot. Aligning them is reasonable tidying rather than something to hold the merge on.

Worth noting the tradeoff runs the other way too: the singular call the harness makes here is the one the node actually exercises, so this loop does at least cover `GetDrepLastRegistrationDeposit`. Whichever form you land on, the gap I flagged in my review is that nothing executes `GetDrepLastRegistrationDeposits` at all today.

database/plugin/metadata/domain_stores_test.go
wcatz: Blocker addressed in 1bccb61. You were right that only a comment stood between the next reader and the natural mistake.

I took your suggestion of `shared_sqlstore_drep_parity_test.go` as the home, and added an import-shaped registration row — real deposit, no certificate, so `certificate_id` lands at 0 — plus assertions that it comes back through both the singular and the batched form, that a row *carrying* a certificate is still found (so the query is not just inverting the filter), and that a credential with no registration history reports 0 without erroring.

Rather than assert the coverage is load-bearing, I checked it. Reintroducing the `certificate_id IS NOT NULL AND certificate_id != 0` filter into both generated queries:

```
--- FAIL: TestSharedSQLStoreDrepParity (0.06s)
    shared_sqlstore_drep_parity_test.go:271
        Error:    Not equal:
                  expected: 0x1dcd6500
                  actual  : 0x0
```

That is 500000000 vs 0 — the wrong-refund failure mode itself, not an incidental assertion. Reverting restores green.

**One correction to the review, on cross-dialect coverage.** The suggestion was that adding here would also exercise the derived-table join "across sqlite, MySQL and Postgres" through `dialect_queryer.go`. It does not, and I do not want to claim coverage the PR does not have. `newSharedSQLStore` (`shared_sqlstore_test.go:29`) calls `openSQLStore` with a sqlite `Config{DataDir: t.TempDir()}` and nothing else, and `exerciseDrepStore` has exactly one caller — `TestSharedSQLStoreDrepParity` in the same file. `grep -rn exerciseDrepStore --include=*.go .` returns only the definition and that one call. The `drepStore` interface and the returned `drepState` are shaped for dialect reuse, but no dialect other than sqlite is wired up today. So the new assertions execute the join against sqlite only; the non-sqlite dialects remain unexecuted, as they were before. Still the right home, and the blocker is closed — the gap is just narrower than the suggestion implied.

**`ledger/view.go:1280` — you are right, and it is fixed.** I verified before changing it rather than taking it on faith: `DRepRegistrations` is declared at gouroboros `ledger/common/state.go:282`, and the only references in v0.202.10 are `WithDRepRegistrations` in `certificate_deposits_test.go:625` and `rules_gov_transition_test.go:646` — its own mocks. No dingo caller. The comment now says the batching bounds the cost of a caller that may appear rather than one that exists, and records explicitly that the read spans every credential's latest registration rather than the active set fetched on the line above. Constraining the query to those credentials needs a SQL change plus `make sql`, so I have left it as the stated next step rather than growing this PR — happy to do it here instead if you would rather it not ship unbounded.

**`sqlstore/drep.go:746` (three states collapsed to 0)** — agreed, and your trace through `InsertDrepIfAbsent` on the vote-replay recovery path is the part I had not seen. Expressing "unknown" needs `common.DRepRegistration.Deposit` to stop being a plain `uint64`, which is a gouroboros change, so a follow-up issue is the right shape. I will open it referencing `governance/processing.go:306`.

**`ledger/view.go:1244` (unconditional deposit lookup on the hot path)** — valid, and the lazy pattern at `ledger/eras/validation.go:250` is the right model. Deliberately not in this commit: it changes the shape of the singular path, and I would rather that not ride along with a test-coverage fix. Follow-up unless you want it here.

Gate re-run on 1bccb61: `go build ./...` and `-tags dingo_extra_plugins` clean, `go test ./database/plugin/metadata/... ./ledger/` pass, `golangci-lint run ./database/... ./ledger/...` 0 issues, `make docs-parity` ok, `make golines` no-op. No SQL or `sqlc.yaml` change in the final state, so `make sql-check` is not implicated.

wolf31o2: Addressed at 4583a069, on top of 1bccb613.

`exerciseDrepStore` now executes both queries against a real sqlite store and asserts the returned deposit, including the import-shaped row (`certificate_id = 0`, `deposit_amount = 500000000`), a certificate-backed row, a credential with two registration rows, a registered-but-inactive credential, and a credential with no registration history. The batched form is pinned by exact map equality.

Store-level coverage alone does not close the call sites, which is where the consensus defect lives. `ledger/drep_deposit_refund_e2e_test.go` drives a real `DeregistrationDrepCertificate` through `conway.UtxoValidateCertificateDeposits` against a `LedgerView` over a real database. Removing the `Deposit` assignment in `LedgerView.DRepRegistration`:

    --- FAIL: TestDRepDeregistrationRefundsRecordedDeposit (0.19s)
               incorrect refund for certificate type 17: supplied 500000000, expected 0

With that same revert in place, `go test ./database/plugin/metadata/sqlite/ -run TestSharedSQLStoreDrepParity` still exits 0.

`dialect_integration_test.go` now executes both queries too, so the batched query's nested derived table runs against real PostgreSQL and MySQL in the `sqlstore-database-integration` job. Locally on postgres:16 and mysql:8: `--- PASS: TestPostgresSQLStoreIntegration (1.74s)`, `--- PASS: TestMySQLSQLStoreIntegration (7.73s)`.

The parity file exercises sqlite only — `newSharedSQLStore` opens the sqlite plugin, and no MySQL or Postgres suite calls `exerciseDrepStore`. `dialect_integration_test.go` is what reaches the non-sqlite dialects.

ledger/view.go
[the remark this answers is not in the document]
wolf31o2: Constrained at 4583a069. The grouped subquery now joins `drep` on `active = TRUE`, so both the scan and the result are bounded by the credential set `GetActiveDreps` returns on the line above.

The premise remains wrong, as noted in the reviewer disposition on this thread: `DRepRegistrations` has no caller, in either tree. The cost was latent, not live. The code comment claiming the validation path is corrected.

Pinned by exact map equality in `exerciseDrepStore`. Removing the join:

    Error: Not equal:
      expected: map[...]{"\x00DDDD…":500000000, "\x00EEEE…":300000000, "\x01BBBB…":500}
      actual:   map[...]{"\x00DDDD…":500000000, "\x00EEEE…":300000000, "\x00FFFF…":200000000, "\x01BBBB…":500}

The extra entry is a registered credential with `active = false`. Same failure on real PostgreSQL through `dialect_integration_test.go`.

internal/test/conformance/state_provider.go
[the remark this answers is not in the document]
wolf31o2: Addressed at 1d5c7812: `DingoStateProvider.DRepRegistrations` issues one `GetDrepLastRegistrationDeposits` and indexes it by `models.DrepDepositKey`, matching `ledger.LedgerView.DRepRegistrations`.

The singular form is still exercised by `DRepRegistration`, which is the one the Conway rules actually reach, so dropping the loop costs no coverage.

DATABASE.md
```diff
@@ -2247,6 +2247,31 @@ WHERE credential_tag = $1 AND drep_credential = decode($2, 'hex')
...
+ORDER BY added_slot DESC
+LIMIT 1;
+```
+
+Note the absence of `GetDrepLastRegistrationSlot`'s
+`certificate_id IS NOT NULL AND certificate_id != 0` filter. That filter is
+correct for an activity display, which should not show bootstrap or
+recovery-inserted placeholder rows, but wrong for refund validation: a row
+inserted by `InsertDrepIfAbsent` carries `certificate_id = 0` and still
+records the real deposit owed on deregistration. Filtering it out yields an
```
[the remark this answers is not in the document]
wolf31o2: Valid, and already corrected at 782f78a5. Verified independently: `InsertDrepIfAbsent` (`sqlstore/drep.go`) issues only `INSERT INTO drep (credential_tag, credential, added_slot, anchor_url, anchor_hash, active) ... ON CONFLICT DO NOTHING`, and the `drep` table has no `certificate_id` or `deposit_amount` column. The `certificate_id = 0` rows come from `ImportDrep`/`ImportDrepRegistration`, called at `ledgerstate/import.go:1366` and `sqlstore/genesis.go:145` with a `models.RegistrationDrep` that leaves `CertificateID` unset. The paragraph now names that path.

ledger/view.go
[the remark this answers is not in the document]
wolf31o2: Addressed at 782f78a5, which replaced the per-DRep loop with `GetDrepLastRegistrationDeposits`. 4583a069 scopes that batched read to the active credential set so it cannot grow with the registration history of deregistered DReps.

internal/test/conformance/state_provider.go
```diff
@@ -882,7 +882,21 @@ func (p *DingoStateProvider) DRepRegistration(
 			return nil, fmt.Errorf("lookup drep registration: %w", err)
 		}
 		if drep != nil && drep.Active {
-			return &common.DRepRegistration{Credential: credential}, nil
+			deposit, err := withBadConnRetry(func() (uint64, error) {
```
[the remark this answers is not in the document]
wolf31o2: Addressed at 782f78a5, and at 1d5c7812 the plural view reads through the batched `GetDrepLastRegistrationDeposits` rather than a query per DRep. All four `DRepRegistration{...}` construction sites in the tree now set `Deposit`: `ledger/view.go:1252` and `:1298`, `internal/test/conformance/state_provider.go:896` and `:981`.

database/plugin/metadata/sqlstore/drep.go
wolf31o2: Confirmed and filed as #4056 rather than grown into this PR. Verified the reachable path independently: `ledger/governance/processing.go:306` is the only `InsertDrepIfAbsent` caller outside the store layer, it passes `active=true`, and `InsertDrepIfAbsent` writes only the `drep` table — so the credential is left active with no `registration_drep` row and `DRepRegistration` reports `Deposit: 0`.

That state is now pinned by `TestDRepRegistrationReportsZeroForUnregisteredCredential`, so the behavior is recorded rather than merely undiscovered.

ledger/view.go
wolf31o2: Valid, and folded into #4056 alongside the unknown-vs-zero note, since both need the same gouroboros-side change.

Confirmed the call-site split: `isDRepRegistered` (`rules.go:3421`) and `UtxoValidateUnknownVoters` (`:4179`) call `ls.DRepRegistration` directly, read only `reg != nil`, and do not memoize — so the cost is per credential per rule, not per transaction. Only `loadDRep` (`:3867`) consumes `Deposit`, and it does memoize.

The lazy treatment at `ledger/eras/validation.go:250` is not reachable here: `common.DRepRegistration.Deposit` is a plain `uint64`, so the value has to be resolved before `DRepRegistration` returns, and dingo cannot see which rule will read it. Deferring it needs either an optional-deposit shape or a separate accessor on `common.DRepState`. Left unchanged in this PR; `idx_drep_reg_cred_slot` keeps the lookup an index seek.

ledger/view.go
wolf31o2: Corrected. 1bccb613 rewrote the comment to say what the batching bounds and that the plural form has no caller; 4583a069 then constrained the query itself, so the batched read now covers exactly the active set `GetActiveDreps` returns and the comment describes the code.

The premise correction is carried into `DATABASE.md` as well: the scope bullet said "the full registration history, not the active set", which is no longer true.

One further inaccuracy in that section, from the same commit: it documented ties within one `added_slot` as arbitrary. `registration_drep` carries `UNIQUE (credential_tag, drep_credential, added_slot)` (`idx_drep_reg_cred_slot`, v1 expand migration), so no tie is representable. Rewritten to say that rather than describing an unspecified choice that cannot occur.

wolf31o2:
Head is now 4583a069.

**Blocking item.** Both queries are executed against a real store and asserted. `exerciseDrepStore` covers the import-shaped row (`certificate_id = 0`, deposit `500000000`), a certificate-backed row, a credential with two registration rows, a registered-but-inactive credential, and one with no registration history; the batched form is pinned by exact map equality. `dialect_integration_test.go` runs both against real PostgreSQL and MySQL, where nothing previously executed the batched query's derived-table join.

Store coverage does not reach the call sites, which is where the rejection came from. `ledger/drep_deposit_refund_e2e_test.go` drives a real `DeregistrationDrepCertificate` through `conway.UtxoValidateCertificateDeposits` against a `LedgerView` over a real database, and checks the recorded deposit wins over both zero and the `DRepDeposit` parameter.

Fails-without-fix, each revert applied in place and reverted after:

| revert | result |
|---|---|
| `Deposit` assignment dropped from `LedgerView.DRepRegistration` | `ledger` e2e fails: `incorrect refund for certificate type 17: supplied 500000000, expected 0`; store parity still exits 0 |
| `Deposit` dropped from `LedgerView.DRepRegistrations` | `ledger` e2e fails, expected `500000000`, actual `0` |
| `certificate_id != 0` filter copied into the singular query | store parity fails, expected `500000000`, actual `0`; `ledger` e2e fails with the refund message; same on real PostgreSQL |
| active-set join removed from the batched query | store parity map gains the inactive credential (len 4 vs 3); same on real PostgreSQL |

**Batched read scope.** `GetDrepLastRegistrationDeposits` now joins `drep` on `active = TRUE` inside the grouped subquery, so neither the scan nor the result grows with the registration history of DReps that have deregistered.

**Construction sites.** All four set `Deposit`: `ledger/view.go:1252`, `:1298`; `internal/test/conformance/state_provider.go:896`, `:981`. No other `DRepRegistration{...}` literal exists in the tree.

**Follow-ups filed.** #4056 (deposit cannot express "unknown"; the recovery-insert path leaves an active DRep at `Deposit: 0`; the eager read for a field three of four call sites ignore). #4057 (`GetDRepState` reports the current `dRepDeposit` parameter rather than the recorded deposit — pre-existing, client query, found while auditing this class).

**Also corrected in `DATABASE.md`:** the scope bullet said the result spans the full registration history, and the tie-break paragraph described an unspecified choice that `UNIQUE (credential_tag, drep_credential, added_slot)` makes unrepresentable.

**Checks on 4583a069** (`GOWORK=off`, isolated caches, exit codes captured by redirect):

| check | exit |
|---|---|
| `go build ./...`, and with `-tags dingo_extra_plugins` | 0 |
| `go vet ./...`, and with the integration tags | 0 |
| `gofmt -l` | clean |
| `golangci-lint run` over `database/... ledger/... internal/...`, and with the integration tags | 0 issues |
| `go test ./ledger/ -timeout 40m` | 0, 79.0s |
| `go test ./database/... -timeout 40m` | 0 |
| `go test ./internal/test/conformance/ ./internal/docsparity ./internal/architecture` | 0, 190.0s |
| `sqlc generate` idempotent, no drift beyond the intended query | 0 |
| `go test -race` on the new sqlite and ledger tests | 0 |
| `dingo_db_integration` on postgres:16 and mysql:8 in Docker | 0 |

Not run: devnet, Antithesis, `make test` in full, Windows and macOS.

wolf31o2:
Added \`e5767fa6\`: \`queryShelleyDRepState\` now reports the deposit recorded against each DRep's own most recent registration instead of the current \`dRepDeposit\` protocol parameter, using the batched read for the unrestricted form and the singular read for the restricted form.

That parameter is governable, so a DRep registered under an earlier value still holds the earlier amount, and every entry in the response previously carried the same number. cardano-ledger stores the amount in `DRepState.drepDeposit` and `query drep-state` serialises that stored field; nothing on the reference query path consults `ppDRepDeposit`.

Reverting it to read the parameter fails both new tests: expected 400000000, actual 600000000.

`go test ./ledger/` exit 0 (75.1s); `golangci-lint run ./ledger/...` 0 issues.

wolf31o2:
Updated DATABASE.md to match the implemented correlated latest-registration query, including its deterministic id tie-break. Removed the obsolete claim that no tie-break is needed. make docs-parity passes.

wolf31o2:
Merged current main to resolve conflicts, retained the v0.204.0 Deposit compatibility changes, and fixed Cubic’s interface-contract finding. Focused ledger, leader, metadata, DRep, and docs-parity checks pass locally; hosted CI is rerunning.

wolf31o2 (approved):

Issue #3844, opened 2026-09-03
ledger: a Plutus tx whose validity bound is past the forecast horizon wedges the node on a canonical block


chrisguiney:
## Summary

A from-genesis Preview replay wedges at slot 3516512 (epoch 40). A canonical
block carries a Plutus transaction whose validity-interval upper bound lies past
the HFC forecast horizon, and converting that bound to POSIX time for the script
context returns `ErrPastHorizon`. The transaction fails validation, its outputs
are never created, the next block that spends them trips missing-input recovery,
and the recovery replay hits the same conversion failure — so it never converges.

Tip is frozen at 3516450.

## Observed

```
level=WARN msg="TX <hash> failed validation: hardfork: slot/time past era horizon"
level=WARN msg="detected inconsistent local ledger state during replay, rewinding
  metadata state" component=ledger recovery_strategy=chain-scan
  failing_block_slot=3516512 missing_input=<hash>#0
```

Order matters: the horizon failure appears **before** the first recovery line, so
the missing input is a consequence, not the trigger.

```
level=ERROR msg="ledger pipeline stuck: repeated restarts are not advancing the
  tip, so the failure is deterministic and will not clear on its own"
  consecutive_no_progress=50 tip_slot=3516450
```

## The transaction

Decoded from the `body_cbor_hex` in the log:

```
field  0  inputs
field  1  outputs
field  2  fee               531984
field  3  ttl               3593399     <- converted for the script context
field  8  validity start    3516496
field 11  script_data_hash  4a40d079c207ef753596fd07487170b85566a98ed2e938bc0f68063d96313142
field 13  collateral
field 18  reference_inputs
```

It is a Plutus transaction, so building the script context requires converting
the validity interval to POSIX time. The upper bound is 3593399.

## The arithmetic

Preview `securityParam` 432, `activeSlotsCoeff` 0.05, epoch length 86400:

```
safe zone        3k/f            = 25920 slots
tip 3516450 + safe zone          = 3542370
horizon snapped to epoch boundary= 3542400   (start of epoch 41)
transaction ttl                  = 3593399   (epoch 41)
ttl exceeds the horizon by         50999 slots
```

So the conversion is refused, and the validity window is legal — roughly 21
hours on Preview, well within what a wallet may set.

## Why this is a divergence rather than a bad transaction

The block is on the canonical Preview chain: peers serve it, and the node's own
tip advanced through 3516450 to reach it. The reference implementation applied
it. Dingo cannot.

[...]

PR #4036, opened 2026-09-06
fix(forging): defer the opcert KES check when history is behind the clock

wcatz:
## Problem

`CurrentSlot` extrapolates the wall clock through the **newest confirmed era's** slot length. A node starting from an empty or far-behind ledger has confirmed history that stops at Byron, so it interprets ~197M seconds of wall clock as Byron 20s slots and computes a current slot near 14,125,000 — KES period 108.

A perfectly valid mainnet operational certificate then looks like it is from the future, and the block producer crash-loops at startup:

```
opcert KES period 1486 is in the future (current 108)
```

The certificate is correct; the clock the check used is not. Nothing the operator can do fixes it — rotating the certificate makes it worse — and it resolves itself the moment confirmed era history spans the wall clock. Observed on a mainnet block producer recovering from an empty ledger.

## Change

Run the strict plausibility check only when the confirmed era history actually spans the wall clock, determined by the new `LedgerState.WallClockSlotFromConfirmedHistory`.

When it does not, still validate the credential material (structure and cold-key signature) and arm the protocol KES lifetime via `PoolCredentials.ArmKesProtocolLifetime`, but defer judging the certificate against a slot that cannot place it, logging a warning that says so.

A clock genuinely **before genesis** keeps the original hard failure, since no certificate can be current then.

## Why deferring is safe

The startup check is not the only enforcement. `forging.forger` re-derives the KES period on **every forge attempt**, before Praos leader selection, and refuses to forge when the period is below the certificate's start or at/past its expiry (`forger.go` — "forge skip: operational certificate is not yet valid" / "...expired"). An unvalidated lifetime fails closed there too.

Arming the lifetime at startup is precisely what gives that per-slot gate its data — so the deferral narrows *when* the certificate is judged, not *whether* it is.

## Relationship to existing work

- **#4001** widens the opcert counter and KES period to full uint64. Different root cause (overflow), does not address era-length extrapolation.
- **#3926** added, then reverted, a startup era-scoped no-gap opcert counter check. The revert rationale — that a node whose applied tip is behind wall-clock time, several rotations in, would fail startup — is the same class of problem. That change removed the check; this one defers it while keeping the runtime gate.

## Tests

- `TestWallClockSlotFromConfirmedHistory_SupportedWhenEraSpansNow` / `_UnsupportedWhenPastHorizon` / `_EmptyCache`
- `TestArmKesProtocolLifetime` plus its `_RequiresGenesis` / `_RequiresCredentials` / `_RequiresPositiveMaxEvolutions` failure paths

All pass under `-race`, along with the pre-existing `TestHardForkSummary_HorizonAnchoredAtAppliedParent` from #3844, which shares the file. `go build ./...`, `go vet`, `gofmt` clean.

Note: `ledger/leader`'s `TestThresholdPreservesProtocolBoundary` fails on `main` independently of this PR (confirmed on a clean `origin/main` checkout) — pre-existing, out of scope.

---
## Summary by cubic
Fixes the block producer crash-loop that treated a valid mainnet operational certificate as "in the future" on startup from an empty or far-behind ledger: `CurrentSlot` extrapolates the wall clock through the newest confirmed era's slot length, so a node whose confirmed history stops at Byron computes a KES period far below the certificate's real one.

The strict KES-period check now runs only when the confirmed era history spans the wall clock, via the new `WallClockSlotFromConfirmedHistory`. When it doesn't, the node validates credential material and arms the protocol KES lifetime via `ArmKesProtocolLifetime`, but defers judging the period, logs a warning, and the clock-before-genesis case keeps its original hard failure. Only the past-horizon case defers; internal failures like an empty epoch cache still error out at startup.

**Deferral is safe**
- The forger re-checks the KES period on every forge attempt and refuses to forge outside the certificate's lifetime, so an unvalidated startup check fails closed per slot.
- Arming the lifetime at startup is what feeds that per-slot gate; the deferral narrows *when* the certificate is judged, not *whether* it is.

**Tests**
- New tests cover `WallClockSlotFromConfirmedHistory`, `ArmKesProtocolLifetime`, and the deferred branch: it arms credentials instead of failing, the strict path still rejects an expired certificate, and missing key material still fails startup.
- `ledger/leader`'s `TestThresholdPreservesProtocolBoundary` fails on `main` independently of this PR.

<sup>Written for commit 42d096f7a0f77a905c7132f36fd6a54ee54d1fde. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4036?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved block producer startup validation when confirmed era history does not cover the current wall-clock time.
  - Startup now defers only slot-dependent KES expiry checks when the forecast horizon is insufficient, while continuing to validate credentials and operational certificates.
  - KES protocol lifetime is initialized for deferred validation, allowing runtime checks to remain available.
  - Runtime validation enforces the KES interval before leader selection.
  - Other startup failures, including invalid credentials, missing configuration, and pre-genesis conditions, continue to fail startup.

### DIFF
--- a/ledger/forging/keys.go
+++ b/ledger/forging/keys.go
@@ -1032,6 +1032,62 @@ func (pc *PoolCredentials) ValidateKESPeriod(
 	return nil
 }
 
+// ArmKesProtocolLifetime seeds the protocol-level KES lifetime from the
+// Shelley genesis without judging the operational certificate against the
+// current wall-clock slot. It exists for block producer startup when the
+// ledger's confirmed era history does not yet span the wall clock (genesis
+// re-import, long downtime): in that state the extrapolated current slot uses
+// the newest known era's slot length and cannot reliably place an opcert in
+// time (a fresh mainnet ledger extrapolates Byron's 20s slots over the whole
+// chain, so the current KES period reads far below the real one). The per-slot
+// forge gate still enforces the armed window against the reliable slot, so no
+// block leaves the node outside the operational certificate's lifetime.
+func (pc *PoolCredentials) ArmKesProtocolLifetime(
+	genesis *shelley.ShelleyGenesis,
+) error {
+	pc.mu.Lock()
+	defer pc.mu.Unlock()
+	pc.generation++
+	// Any failed arming leaves the credentials unusable for production
+	// forging rather than retaining stale policy data.
+	pc.maxKESEvolutions = 0
+	pc.opCertExpiryKES = 0
+	pc.opCertValidated = false
+
+	if pc.opCert == nil {
+		pc.opCertStartKES = 0
+		return errors.New("operational certificate not loaded")
+	}
+	pc.opCertStartKES = pc.opCert.KESPeriod
+	if pc.isLoadedUnsafe() {
+		if err := pc.validateOpCertUnsafe(); err != nil {
+			return fmt.Errorf("validate operational certificate: %w", err)
+		}
+		pc.opCertValidated = true
+	}
+	if genesis == nil {
+		return errors.New("shelley genesis is required")
+	}
+	if genesis.MaxKESEvolutions <= 0 {
+		return fmt.Errorf(
+			"genesis maxKESEvolutions must be positive, got %d",
+			genesis.MaxKESEvolutions,
+		)
+	}
+	// #nosec G115 -- guarded positive above; int fits within uint64.
+	maxEvolutions := uint64(genesis.MaxKESEvolutions)
+	expiryPeriod, err := opCertExpiryPeriod(
+		pc.opCert.KESPeriod,
+		maxEvolutions,
+	)
+	if err != nil {
+		return err
+	}
+	pc.maxKESEvolutions = maxEvolutions
+	pc.opCertExpiryKES = expiryPeriod
+	return nil
+}
+
 // LedgerView is the subset of ledger state the post-startup credential
 // cross-check needs. The forging package depends on it as a small
 // interface so the package itself stays free of a ledger dependency, and
--- a/ledger/forging/keys.go
+++ b/ledger/forging/keys.go
@@ -1059,12 +1059,21 @@ func (pc *PoolCredentials) ArmKesProtocolLifetime(
 		return errors.New("operational certificate not loaded")
 	}
 	pc.opCertStartKES = pc.opCert.KESPeriod
-	if pc.isLoadedUnsafe() {
-		if err := pc.validateOpCertUnsafe(); err != nil {
-			return fmt.Errorf("validate operational certificate: %w", err)
-		}
-		pc.opCertValidated = true
+	// Without signing material the certificate cannot be validated, so
+	// opCertValidated would stay false and OpCertExpiryPeriod would report 0
+	// however far the rest of this function got. Fail rather than return
+	// nil: a method named Arm that reports success and arms nothing leaves
+	// the caller believing the per-slot forge gate has data to enforce when
+	// it does not. Unreachable from the block producer startup path, where
+	// LoadFromFiles precedes this, but this is exported API on an exported
+	// type.
+	if !pc.isLoadedUnsafe() {
+		return errors.New("signing material not loaded")
 	}
+	if err := pc.validateOpCertUnsafe(); err != nil {
+		return fmt.Errorf("validate operational certificate: %w", err)
+	}
+	pc.opCertValidated = true
 	if genesis == nil {
 		return errors.New("shelley genesis is required")
 	}
--- a/ledger/hardfork_summary.go
+++ b/ledger/hardfork_summary.go
@@ -233,3 +233,28 @@ func (ls *LedgerState) hardForkSummaryAnchoredAt(
 	}
 	return &summary, nil
 }
+
+// WallClockSlotFromConfirmedHistory returns the absolute slot corresponding to
+// the current wall-clock time derived only from the confirmed era history, and
+// false when the wall clock falls past that history's forecast horizon.
+//
+// In the false case CurrentSlot can only extrapolate the wall clock through
+// the newest confirmed era's slot length, which does not reflect chains whose
+// later eras have not been applied yet: a mainnet node importing from genesis
+// judges the wall clock with Byron's 20s slots even though the real chain
+// moved to one-second slots at epoch 208. Operational callers that must make a
+// time-sensitive judgement against the wall-clock slot — such as the block
+// producer's operational-certificate KES-period preflight — should use this
+// instead of the extrapolated CurrentSlot, and defer the judgement until the
+// confirmed era history spans the wall clock again.
+func (ls *LedgerState) WallClockSlotFromConfirmedHistory() (uint64, bool) {
+	sum, err := ls.HardForkSummary()
+	if err != nil {
+		return 0, false
+	}
+	slot, err := sum.TimeToSlot(time.Now())
+	if err != nil {
+		return 0, false
+	}
+	return slot, true
+}
--- a/ledger/hardfork_summary.go
+++ b/ledger/hardfork_summary.go
@@ -16,6 +16,7 @@ package ledger
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/blinklabs-io/dingo/ledger/hardfork"
@@ -238,6 +239,16 @@ func (ls *LedgerState) hardForkSummaryAnchoredAt(
 // the current wall-clock time derived only from the confirmed era history, and
 // false when the wall clock falls past that history's forecast horizon.
 //
+// Only the past-horizon case reports false. Any other failure -- an empty
+// epoch cache, a summary that will not build, a wall clock before genesis --
+// is returned as an error instead. The distinction matters to the block
+// producer: false means "cannot judge yet, defer and let the per-slot gate
+// enforce", so conflating a genuine internal failure with it would silently
+// downgrade a hard startup failure into a warning. Note that before-genesis
+// reaches here as hardfork.ErrBeforeGenesis, a different sentinel from
+// ledger.ErrBeforeGenesis, and returning it as an error keeps that case a
+// hard failure regardless of which sentinel a caller happens to check.
+//
 // In the false case CurrentSlot can only extrapolate the wall clock through
 // the newest confirmed era's slot length, which does not reflect chains whose
 // later eras have not been applied yet: a mainnet node importing from genesis
@@ -247,14 +258,24 @@ func (ls *LedgerState) hardForkSummaryAnchoredAt(
 // producer's operational-certificate KES-period preflight — should use this
 // instead of the extrapolated CurrentSlot, and defer the judgement until the
 // confirmed era history spans the wall clock again.
-func (ls *LedgerState) WallClockSlotFromConfirmedHistory() (uint64, bool) {
+func (ls *LedgerState) WallClockSlotFromConfirmedHistory() (
+	uint64,
+	bool,
+	error,
+) {
 	sum, err := ls.HardForkSummary()
 	if err != nil {
-		return 0, false
+		return 0, false, fmt.Errorf("build confirmed era summary: %w", err)
 	}
 	slot, err := sum.TimeToSlot(time.Now())
-	if err != nil {
-		return 0, false
+	switch {
+	case errors.Is(err, hardfork.ErrPastHorizon):
+		return 0, false, nil
+	case err != nil:
+		return 0, false, fmt.Errorf(
+			"resolve wall-clock slot from confirmed history: %w",
+			err,
+		)
 	}
-	return slot, true
+	return slot, true, nil
 }
--- a/node_forging.go
+++ b/node_forging.go
@@ -52,19 +52,60 @@ func (n *Node) validateBlockProducerStartup() (*forging.PoolCredentials, error) 
 			"block producer mode requires ledger state for current slot",
 		)
 	}
-	currentSlot, err := n.ledgerState.CurrentSlot()
-	if err != nil {
+	if _, err := n.ledgerState.CurrentSlot(); err != nil {
 		if !errors.Is(err, ledger.ErrBeforeGenesis) {
 			return nil, fmt.Errorf("compute current slot: %w", err)
 		}
-		currentSlot = 0
+		// Clock before genesis: preserve the historical hard fail through the
+		// strict preflight (an opcert can never be "current" before genesis).
+		return n.validateBlockProducerStartupAtSlot(0)
 	}
-	return n.validateBlockProducerStartupAtSlot(currentSlot)
+	// The current wall-clock slot only reliably places an operational
+	// certificate in time when the confirmed era history spans the wall clock.
+	// A node importing from genesis or restarting far behind has confirmed
+	// history that stops at its newest applied era: CurrentSlot then
+	// extrapolates through that era's slot length — a fresh mainnet node
+	// judges the wall clock with Byron's 20s slots even though the real chain
+	// moved to one-second slots at epoch 208 — which cannot judge an opcert for
+	// the actual chain. In that state defer the plausibility check: forging
+	// stays gated on sync progress and re-enforces the opcert KES lifetime per
+	// slot, and the protocol lifetime is still armed below so that runtime gate
+	// has data to enforce.
+	supportedSlot, supported := n.ledgerState.WallClockSlotFromConfirmedHistory()
+	if !supported {
+		creds, err := n.validateBlockProducerCredentialMaterial()
+		if err != nil {
+			return nil, err
+		}
+		genesis, err := n.blockProducerShelleyGenesis()
+		if err != nil {
+			return nil, err
+		}
+		if err := creds.ArmKesProtocolLifetime(genesis); err != nil {
+			return nil, fmt.Errorf("arm KES protocol lifetime: %w", err)
+		}
+		opCert := creds.GetOpCert()
+		n.config.logger.Warn(
+			"block producer startup: confirmed era history does not span the current wall-clock slot yet; "+
+				"deferring the operational-certificate KES-period plausibility check until the ledger catches up. "+
+				"Credential material is validated; forging stays gated on sync progress and re-checks the opcert KES lifetime per slot.",
+			"component", "node",
+			"tip_slot", n.ledgerState.ChainTipSlot(),
+			"opcert_kes_period", opCert.KESPeriod,
+			"opcert_expiry_period", creds.OpCertExpiryPeriod(),
+		)
+		return creds, nil
+	}
+	return n.validateBlockProducerStartupAtSlot(supportedSlot)
 }
 
-func (n *Node) validateBlockProducerStartupAtSlot(
-	currentSlot uint64,
-) (*forging.PoolCredentials, error) {
+// validateBlockProducerCredentialMaterial loads the pool credentials and
+// validates the operational certificate's structure and cold-key signature,
+// without judging it against any slot. The slot-dependent KES-period
+// plausibility check lives in validateBlockProducerStartupAtSlot; callers
+// needing the certificate armed but not yet placed in time combine this with
+// PoolCredentials.ArmKesProtocolLifetime.
+func (n *Node) validateBlockProducerCredentialMaterial() (*forging.PoolCredentials, error) {
 	creds := forging.NewPoolCredentials()
 	if err := creds.LoadFromFiles(
 		n.config.shelleyVRFKey,
@@ -75,6 +116,16 @@ func (n *Node) validateBlockProducerStartupAtSlot(
 	}
 	if err := creds.ValidateOpCert(); err != nil {
 		return nil, fmt.Errorf("validate operational certificate: %w", err)
+	}
+	return creds, nil
+}
+
+func (n *Node) validateBlockProducerStartupAtSlot(
+	currentSlot uint64,
+) (*forging.PoolCredentials, error) {
+	creds, err := n.validateBlockProducerCredentialMaterial()
+	if err != nil {
+		return nil, err
 	}
 	genesis, err := n.blockProducerShelleyGenesis()
 	if err != nil {
--- a/node_forging.go
+++ b/node_forging.go
@@ -58,7 +58,7 @@ func (n *Node) validateBlockProducerStartup() (*forging.PoolCredentials, error) 
 		}
 		// Clock before genesis: preserve the historical hard fail through the
 		// strict preflight (an opcert can never be "current" before genesis).
-		return n.validateBlockProducerStartupAtSlot(0)
+		return n.validateBlockProducerStartupForClock(0, true)
 	}
 	// The current wall-clock slot only reliably places an operational
 	// certificate in time when the confirmed era history spans the wall clock.
@@ -67,36 +67,67 @@ func (n *Node) validateBlockProducerStartup() (*forging.PoolCredentials, error) 
 	// extrapolates through that era's slot length — a fresh mainnet node
 	// judges the wall clock with Byron's 20s slots even though the real chain
 	// moved to one-second slots at epoch 208 — which cannot judge an opcert for
-	// the actual chain. In that state defer the plausibility check: forging
-	// stays gated on sync progress and re-enforces the opcert KES lifetime per
-	// slot, and the protocol lifetime is still armed below so that runtime gate
-	// has data to enforce.
-	supportedSlot, supported := n.ledgerState.WallClockSlotFromConfirmedHistory()
-	if !supported {
-		creds, err := n.validateBlockProducerCredentialMaterial()
-		if err != nil {
-			return nil, err
-		}
-		genesis, err := n.blockProducerShelleyGenesis()
-		if err != nil {
-			return nil, err
-		}
-		if err := creds.ArmKesProtocolLifetime(genesis); err != nil {
-			return nil, fmt.Errorf("arm KES protocol lifetime: %w", err)
-		}
-		opCert := creds.GetOpCert()
-		n.config.logger.Warn(
-			"block producer startup: confirmed era history does not span the current wall-clock slot yet; "+
-				"deferring the operational-certificate KES-period plausibility check until the ledger catches up. "+
-				"Credential material is validated; forging stays gated on sync progress and re-checks the opcert KES lifetime per slot.",
-			"component", "node",
-			"tip_slot", n.ledgerState.ChainTipSlot(),
-			"opcert_kes_period", opCert.KESPeriod,
-			"opcert_expiry_period", creds.OpCertExpiryPeriod(),
-		)
-		return creds, nil
+	// the actual chain.
+	supportedSlot, supported, err := n.ledgerState.
+		WallClockSlotFromConfirmedHistory()
+	if err != nil {
+		return nil, fmt.Errorf("wall-clock slot from confirmed history: %w", err)
 	}
-	return n.validateBlockProducerStartupAtSlot(supportedSlot)
+	return n.validateBlockProducerStartupForClock(supportedSlot, supported)
+}
+
+// validateBlockProducerStartupForClock decides how to judge the operational
+// certificate given a wall-clock slot and whether the confirmed era history
+// actually supports it.
+//
+// When supported, this is the strict historical preflight: a certificate
+// staged for the future or already expired fails startup here. When not
+// supported, the KES-period plausibility check is deferred rather than
+// skipped — credential material is still validated and the protocol lifetime
+// is still armed, so the forger's per-slot gate has the data it needs to
+// reject the certificate before Praos leader selection.
+//
+// Split out from validateBlockProducerStartup so both branches are reachable
+// without a live LedgerState, the same reason
+// validateBlockProducerStartupAtSlot and validateBlockProducerLedgerWithView
+// exist separately.
+func (n *Node) validateBlockProducerStartupForClock(
+	slot uint64,
+	supported bool,
+) (*forging.PoolCredentials, error) {
+	if supported {
+		return n.validateBlockProducerStartupAtSlot(slot)
+	}
+	creds, err := n.validateBlockProducerCredentialMaterial()
+	if err != nil {
+		return nil, err
+	}
+	genesis, err := n.blockProducerShelleyGenesis()
+	if err != nil {
+		return nil, err
+	}
+	if err := creds.ArmKesProtocolLifetime(genesis); err != nil {
+		return nil, fmt.Errorf("arm KES protocol lifetime: %w", err)
+	}
+	opCert := creds.GetOpCert()
+	if opCert == nil {
+		return nil, errors.New("block producer operational certificate is nil")
+	}
+	var tipSlot uint64
+	if n.ledgerState != nil {
+		tipSlot = n.ledgerState.ChainTipSlot()
+	}
+	n.config.logger.Warn(
+		"block producer startup: confirmed era history does not span the current wall-clock slot yet; "+
+			"deferring the operational-certificate KES-period plausibility check until the ledger catches up. "+
+			"The opcert KES lifetime is re-checked per slot and forging cannot proceed while it fails; "+
+			"sync progress additionally gates forging whenever an upstream peer is active.",
+		"component", "node",
+		"tip_slot", tipSlot,
+		"opcert_kes_period", opCert.KESPeriod,
+		"opcert_expiry_period", creds.OpCertExpiryPeriod(),
+	)
+	return creds, nil
 }
 
 // validateBlockProducerCredentialMaterial loads the pool credentials and
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -4452,7 +4452,28 @@ KES periods are computed from the era-aware absolute slot (`currentSlot / slotsP
 Successful startup validation captures Shelley genesis `MaxKESEvolutions` on
 the loaded credentials together with the opcert start and overflow-checked
 exclusive expiry. `NewBlockForger` rejects credentials without that validated
-protocol lifetime. Before leader selection at each candidate slot, the runtime
+protocol lifetime.
+
+The startup KES-period judgement itself is conditional on the confirmed era
+history spanning the wall clock. `WallClockSlotFromConfirmedHistory` resolves
+the current slot from confirmed eras only, and reports unsupported when the
+wall clock falls past that history's forecast horizon — the state a node
+importing from genesis or restarting far behind is in, where `CurrentSlot`
+would extrapolate through the newest applied era's slot length (Byron's 20s
+slots on a chain that moved to 1s at epoch 208) and so cannot place a
+certificate in time. Only that past-horizon case is a deferral; any other
+failure, including an empty epoch cache or a wall clock before genesis, is
+returned as an error and keeps startup a hard failure.
+
+In the deferred state startup still loads the credential material, checks the
+opcert's cold-key signature, and arms the same protocol lifetime — it skips
+only the slot-dependent plausibility check. Enforcement is then entirely the
+per-slot runtime gate described below, which re-derives the period and rejects
+the certificate before leader selection; extrapolation through the newest
+confirmed era skews the computed period downward, so a certificate staged for
+the future or already expired lands below `opcertStart` rather than inside the
+admitted interval. When confirmed history does span the wall clock, the strict
+startup check runs unchanged and rejects such a certificate outright. Before leader selection at each candidate slot, the runtime
 gate admits exactly the
 half-open interval `[opcertStart, opcertStart + MaxKESEvolutions)`: periods
 before the start and at or after the exclusive end both log/count a
--- a/ledger/forging/keys_test.go
+++ b/ledger/forging/keys_test.go
@@ -158,6 +158,77 @@ func writeTestOpCert(t *testing.T, contents string) string {
 	return path
 }
 
+func TestArmKesProtocolLifetime(t *testing.T) {
+	vrfPath, kesPath, opCertPath := createTestKeys(t)
+	genesis := synthGenesis(
+		129600, 62, time.Second,
+		time.Date(2017, 9, 23, 21, 44, 51, 0, time.UTC),
+	)
+
+	pc := NewPoolCredentials()
+	err := pc.LoadFromFiles(vrfPath, kesPath, opCertPath)
+	require.NoError(t, err)
+
+	// Before arming there is no usable protocol lifetime.
+	assert.Zero(t, pc.OpCertExpiryPeriod())
+	assert.Zero(t, pc.PeriodsRemaining(1))
+
+	require.NoError(t, pc.ArmKesProtocolLifetime(genesis))
+	assert.Equal(t, uint64(0), pc.opCertStartKES)
+	assert.Equal(t, uint64(62), pc.maxKESEvolutions)
+	assert.Equal(t, uint64(62), pc.opCertExpiryKES)
+	assert.Equal(t, uint64(62), pc.OpCertExpiryPeriod())
+	// Periods inside the armed window count down; at/after expiry they are 0.
+	assert.Equal(t, uint64(57), pc.PeriodsRemaining(5))
+	assert.Equal(t, uint64(61), pc.PeriodsRemaining(1))
+	assert.Zero(t, pc.PeriodsRemaining(62))
+	assert.Zero(t, pc.PeriodsRemaining(100))
+	assert.True(t, pc.opCertValidated)
+}
+
+func TestArmKesProtocolLifetime_RequiresGenesis(t *testing.T) {
+	vrfPath, kesPath, opCertPath := createTestKeys(t)
+	pc := NewPoolCredentials()
+	require.NoError(t, pc.LoadFromFiles(vrfPath, kesPath, opCertPath))
+
+	err := pc.ArmKesProtocolLifetime(nil)
+	require.Error(t, err)
+	// The opcert structure may validate before the genesis is consulted, but
+	// without genesis data there is no protocol lifetime to enforce.
+	assert.Zero(t, pc.maxKESEvolutions)
+	assert.Zero(t, pc.opCertExpiryKES)
+	assert.Zero(t, pc.OpCertExpiryPeriod())
+	assert.Zero(t, pc.PeriodsRemaining(1))
+}
+
+func TestArmKesProtocolLifetime_RequiresCredentials(t *testing.T) {
+	genesis := synthGenesis(
+		129600, 62, time.Second,
+		time.Date(2017, 9, 23, 21, 44, 51, 0, time.UTC),
+	)
+	pc := NewPoolCredentials()
+	err := pc.ArmKesProtocolLifetime(genesis)
+	require.Error(t, err)
+	assert.Zero(t, pc.OpCertExpiryPeriod())
+}
+
+func TestArmKesProtocolLifetime_RequiresPositiveMaxEvolutions(t *testing.T) {
+	vrfPath, kesPath, opCertPath := createTestKeys(t)
+	genesis := synthGenesis(
+		129600, 0, time.Second,
+		time.Date(2017, 9, 23, 21, 44, 51, 0, time.UTC),
+	)
+	pc := NewPoolCredentials()
+	require.NoError(t, pc.LoadFromFiles(vrfPath, kesPath, opCertPath))
+
+	err := pc.ArmKesProtocolLifetime(genesis)
+	require.Error(t, err)
+	assert.Zero(t, pc.maxKESEvolutions)
+	assert.Zero(t, pc.opCertExpiryKES)
+	assert.Zero(t, pc.OpCertExpiryPeriod())
+	assert.Zero(t, pc.PeriodsRemaining(1))
+}
+
 func TestPoolCredentialsLoadFromFiles(t *testing.T) {
 	vrfPath, kesPath, opCertPath := createTestKeys(t)
 	loadedVRF, err := loadSecretKeyFromFile(vrfPath)
--- a/ledger/forging/keys_test.go
+++ b/ledger/forging/keys_test.go
@@ -1538,3 +1538,23 @@ func TestValidateOpCertSequence(t *testing.T) {
 		})
 	}
 }
+
+// TestArmKesProtocolLifetime_RequiresSigningMaterial pins that Arm reports
+// failure when it cannot arm anything. With an opcert present but no signing
+// material the certificate cannot be validated, so the protocol lifetime
+// would be unreadable; returning nil there would tell the block producer the
+// per-slot forge gate has data to enforce when it has none.
+func TestArmKesProtocolLifetime_RequiresSigningMaterial(t *testing.T) {
+	genesis := synthGenesis(
+		129600, 62, time.Second,
+		time.Date(2017, 9, 23, 21, 44, 51, 0, time.UTC),
+	)
+	pc := &PoolCredentials{opCert: &OpCert{KESPeriod: 1}}
+	require.False(t, pc.IsLoaded())
+
+	err := pc.ArmKesProtocolLifetime(genesis)
+	require.Error(t, err)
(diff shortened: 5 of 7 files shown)


wcatz:

```
```

node_forging.go
```diff
@@ -52,19 +52,60 @@ func (n *Node) validateBlockProducerStartup() (*forging.PoolCredentials, error)
...
+	// history that stops at its newest applied era: CurrentSlot then
+	// extrapolates through that era's slot length — a fresh mainnet node
+	// judges the wall clock with Byron's 20s slots even though the real chain
+	// moved to one-second slots at epoch 208 — which cannot judge an opcert for
+	// the actual chain. In that state defer the plausibility check: forging
+	// stays gated on sync progress and re-enforces the opcert KES lifetime per
+	// slot, and the protocol lifetime is still armed below so that runtime gate
+	// has data to enforce.
+	supportedSlot, supported := n.ledgerState.WallClockSlotFromConfirmedHistory()
+	if !supported {
```
chrisguiney: **Blocker (missing test).** This whole branch, and the `supported` path below it, are unreachable from the test suite. I reverted just this file to `origin/main` in my worktree, kept both new helpers so it still compiled, and `go test -race ./ledger/forging/ .` plus the new `./ledger` tests stayed green — all 58 added lines can be deleted without a failure.

The case worth locking down is not the deferral, it's the negative: that `supported == true` still rejects a future-staged or expired opcert. Without it, nothing distinguishes this change from a relaxation of the gate.

`validateBlockProducerStartupAtSlot` and `validateBlockProducerLedgerWithView` already exist in this file precisely so node-level logic can be tested without a live `LedgerState`. Extracting the decision the same way, e.g.

```go
func (n *Node) validateBlockProducerStartupForClock(
	slot uint64,
	supported bool,
) (*forging.PoolCredentials, error)
```

lets `validateBlockProducerStartup` stay a three-line wrapper around `WallClockSlotFromConfirmedHistory` and makes all three branches — deferred, supported, before-genesis — reachable from the existing `newTestNodeForBP` harness.

node_forging.go
```diff
@@ -52,19 +52,60 @@ func (n *Node) validateBlockProducerStartup() (*forging.PoolCredentials, error)
...
+			return nil, err
+		}
+		genesis, err := n.blockProducerShelleyGenesis()
+		if err != nil {
+			return nil, err
+		}
+		if err := creds.ArmKesProtocolLifetime(genesis); err != nil {
+			return nil, fmt.Errorf("arm KES protocol lifetime: %w", err)
+		}
+		opCert := creds.GetOpCert()
```
chrisguiney: `GetOpCert()` returns nil when no certificate is loaded, and `opCert.KESPeriod` on line 94 dereferences it unguarded. The sibling path guards the identical call at line 145 (`"block producer operational certificate is nil"`).

It is not reachable today — `ArmKesProtocolLifetime` returns an error when `pc.opCert == nil`, and the caller returns on that — so this is consistency rather than a live defect. Worth matching the guard anyway: `make lint` runs `nilaway`, which the CI `lint` job does not (it runs golangci-lint only), so a local pre-commit run may flag this where CI stayed green.

node_forging.go
```diff
@@ -52,19 +52,60 @@ func (n *Node) validateBlockProducerStartup() (*forging.PoolCredentials, error)
...
+		n.config.logger.Warn(
+			"block producer startup: confirmed era history does not span the current wall-clock slot yet; "+
+				"deferring the operational-certificate KES-period plausibility check until the ledger catches up. "+
+				"Credential material is validated; forging stays gated on sync progress and re-checks the opcert KES lifetime per slot.",
```
chrisguiney: "forging stays gated on sync progress" is conditional in a way the message doesn't say. The sync gate at `forging/forger.go:820` is `if upstreamActive && ...` — a producer with no active upstream (isolated node, relays not yet connected) falls straight through it to the KES gate.

That is still safe, because the KES gate is what actually fails closed, but this text is what an operator reads while deciding whether a deferred check is acceptable. The unconditional half is the second clause; consider leading with it.

ledger/hardfork_summary.go
```diff
@@ -233,3 +233,28 @@ func (ls *LedgerState) hardForkSummaryAnchoredAt(
...
+// later eras have not been applied yet: a mainnet node importing from genesis
+// judges the wall clock with Byron's 20s slots even though the real chain
+// moved to one-second slots at epoch 208. Operational callers that must make a
+// time-sensitive judgement against the wall-clock slot — such as the block
+// producer's operational-certificate KES-period preflight — should use this
+// instead of the extrapolated CurrentSlot, and defer the judgement until the
+// confirmed era history spans the wall clock again.
+func (ls *LedgerState) WallClockSlotFromConfirmedHistory() (uint64, bool) {
+	sum, err := ls.HardForkSummary()
+	if err != nil {
```
chrisguiney, line 267: The doc comment above says `false` means "the wall clock falls past that history's forecast horizon", but the implementation returns `false` for *any* error from either call. `HardForkSummary` can fail with `"ledger: no epochs in cache"` or a `BuildSummary` error, and `TimeToSlot` returns `hardfork.ErrBeforeGenesis` as well as `ErrPastHorizon`.

Two consequences at the caller:

- A genuine internal failure silently downgrades a hard startup failure into a warning, which is the opposite of what the pre-change code did (it propagated any non-`ErrBeforeGenesis` error as `"compute current slot"`).
- The "before genesis keeps the original hard failure" invariant rests entirely on `ls.slotClock.CurrentSlot()` reaching its `ledger.ErrBeforeGenesis` branch first. Note these are two distinct sentinels — `ledger.ErrBeforeGenesis` in `slot_time_converter.go:32` and `hardfork.ErrBeforeGenesis` in `hardfork/summary.go:26` — and only the first one is checked. If they ever diverge, before-genesis takes the deferral instead.

`ErrPastHorizon` is typed and is exactly the condition described, so the function can say what it means:

```go
slot, err := sum.TimeToSlot(time.Now())
if errors.Is(err, hardfork.ErrPastHorizon) {
	return 0, false
}
if err != nil {
	return 0, err  // or a third state
}
```

The `(uint64, bool)` signature can't express the distinction, so this needs `(uint64, bool, error)` or `(uint64, error)`. Cubic raised the same conflation; this is the part of its finding I'd act on.

ledger/forging/keys.go
```diff
@@ -1032,6 +1032,62 @@ func (pc *PoolCredentials) ValidateKESPeriod(
...
+		pc.opCertStartKES = 0
+		return errors.New("operational certificate not loaded")
+	}
+	pc.opCertStartKES = pc.opCert.KESPeriod
+	if pc.isLoadedUnsafe() {
+		if err := pc.validateOpCertUnsafe(); err != nil {
+			return fmt.Errorf("validate operational certificate: %w", err)
+		}
+		pc.opCertValidated = true
+	}
```
chrisguiney, line 1075: Cubic flags this branch (`isLoadedUnsafe()` false) against `ValidateKESPeriod`, which preserves `previousOpCertValidation` here. I'd not take that fix: preserving a stale `opCertValidated` is what would make an unloaded credential report a *usable* lifetime through `opCertExpiryPeriodUnsafe`, and the current behavior — leaving it false so `OpCertExpiryPeriod()` is 0 and the forger's `validatedKESProtocolLifetime` fails closed — is the safer of the two.

The real wart is the return value. When `pc.opCert != nil` but signing material is absent, this function runs to the bottom and returns `nil` while `opCertValidated` stays false, so the lifetime it just "armed" is unreadable. A method named *Arm* that reports success and arms nothing is the contract problem; returning an error in this branch fixes it without weakening anything:

```go
if !pc.isLoadedUnsafe() {
	return errors.New("signing material not loaded")
}
```

Unreachable from the only caller today (`LoadFromFiles` + `ValidateOpCert` guarantee loaded), so non-blocking — but this is exported API on an exported type.

chrisguiney (dismissed):
Reviewed at `da662b0` in a clean worktree off `origin/main` (merge base `7ad2791`).

## The safety argument holds

I traced the runtime backstop rather than taking the PR description for it, and it is real. `NewBlockForger` refuses production mode outright when `validatedKESProtocolLifetime()` fails (`ledger/forging/forger.go:448`), and the per-slot gate re-derives the period and rejects both `kesPeriod < opCertStart` and `kesPeriod >= opCertExpiry` before Praos leader selection (`forger.go:867-906`). `ArmKesProtocolLifetime` is exactly what supplies that gate its data, so the deferral narrows *when* the certificate is judged, not *whether*.

I also probed the inverse the deferral could open — an expired or future-staged opcert slipping through. Extrapolating through the newest confirmed era's slot length skews the computed period **downward** (Byron 20s slots over a chain that is really 1s), so a certificate that is wrong in either direction lands *below* `opCertStart` rather than inside `[start, expiry)`. Every case I constructed fails closed at the per-slot gate. `credentialGeneration.validatedKESProtocolLifetime` also fails closed on an unvalidated or zeroed lifetime, so the partial state each early `return` in `ArmKesProtocolLifetime` leaves behind is unusable rather than permissive.

## Blocking: the behavior change has no test

The two new helpers are well covered in isolation, but nothing exercises `validateBlockProducerStartup`'s new branching. I checked out `node_forging.go` from `origin/main` in my worktree — keeping both new helpers so everything still compiles — and re-ran the full suites:

```
git checkout origin/main -- node_forging.go
go build ./...                                   # ok
go test -race -count=1 ./ledger/forging/ .       # ok  (34.9s)
go test -race -count=1 -run TestWallClockSlotFromConfirmedHistory ./ledger/   # ok
```

All 58 changed lines can be deleted with the suite still green. So three claims in the description are currently unproven by any test:

1. the deferral path returns armed credentials instead of failing;
2. the strict check **still rejects** a future-staged or expired opcert when confirmed history *does* span the wall clock;
3. a clock before genesis still hard-fails.

(2) is the one that matters — it is the property that keeps this from being a relaxation. This file already has the seam for it: `validateBlockProducerStartupAtSlot` and `validateBlockProducerLedgerWithView` both exist so the node-level logic can be tested without a live `LedgerState`. A `validateBlockProducerStartupForClock(slot uint64, supported bool)` extracted the same way makes all three branches reachable from the existing `newTestNodeForBP` harness.

`AGENTS.md` also asks for a DevNet run on any change touching block production or OpCert verification; the description records none.

## Blocking: `ARCHITECTURE.md` is now inaccurate

The repository treats `ARCHITECTURE.md` as part of the change bar for startup behavior. Line 4452 currently reads:

> Successful startup validation captures Shelley genesis `MaxKESEvolutions` on the loaded credentials together with the opcert start and overflow-checked exclusive expiry.

After this change there is a second path that captures exactly that lifetime **without** the successful validation. The section should record that the startup KES-period judgement is conditional on confirmed era history spanning the wall clock, and that the per-slot gate is the enforcement in the deferred state.

## Checks run (macOS, go1.27.1, worktree off `origin/main`)

| check | exit |
|---|---|
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `gofmt -l .` | 0 (two pre-existing files unrelated to this PR) |
| `golangci-lint run ./ledger/... .` (isolated cache) | 0 — "0 issues." |
| `go test -race -count=1 ./ledger/forging/` | 0 |
| `go test -race -count=1 -run 'TestWallClockSlotFromConfirmedHistory\|TestHardForkSummary' ./ledger/` | 0 |
| `go test -race -count=1 -run TestValidateBlockProducer .` | 0 |
| revert-and-rerun (above) | 0 — the finding |

Not run: `nilaway` and `modernize` (not installed here; CI's `lint` job runs golangci-lint only, so the `GetOpCert()` note below is unverified by any automated gate), and DevNet/conformance (no devnet infrastructure available).

CI on `da662b0`: `lint`, `Conventional Commits`, `DCO`, `docker`, `Analyze (actions)`, and go-test on Linux/macOS/Windows are all green. `Analyze (go)`, `go-test (Linux, race)`, and `govulncheck` were still queued when I read them — no failures, but the pipeline is not complete.

## Bot findings

- **CodeRabbit** did not review this PR. Its status check reports SUCCESS, but the comment body is a rate-limit warning ("Review limit reached"). Cubic completed, which satisfies the bot-review requirement.
- **Cubic, `keys.go:1055`** — partially valid, but the suggested fix points the wrong way; see the inline comment.
- **Cubic, `node_forging.go:75`** — the error-conflation half is valid and I've kept it inline. The broader claim, that this "proceeds into block production" with an unjudged certificate, is disproved by the constructor and per-slot gate traced above.

ledger/forging/keys.go
```diff
@@ -1032,6 +1032,62 @@ func (pc *PoolCredentials) ValidateKESPeriod(
...
+	genesis *shelley.ShelleyGenesis,
+) error {
+	pc.mu.Lock()
+	defer pc.mu.Unlock()
+	pc.generation++
+	// Any failed arming leaves the credentials unusable for production
+	// forging rather than retaining stale policy data.
+	pc.maxKESEvolutions = 0
+	pc.opCertExpiryKES = 0
+	pc.opCertValidated = false
```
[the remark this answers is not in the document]
chrisguiney: Verified against `da662b0`. Partially valid, but I would not apply the suggested fix — preserving a stale `opCertValidated` here is what would let an unloaded credential report a *usable* protocol lifetime through `opCertExpiryPeriodUnsafe`. The current behavior (leaving it false, so `OpCertExpiryPeriod()` is 0 and the forger's `validatedKESProtocolLifetime` fails closed) is the safer of the two.

The genuine wart is that this branch returns `nil` while arming nothing. Disposition and a corrected suggestion are in the inline comment at `ledger/forging/keys.go:1067`. Unreachable from the only caller today, so non-blocking.

node_forging.go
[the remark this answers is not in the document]
chrisguiney: Verified against `da662b0`, split disposition.

**Valid:** the error conflation. `WallClockSlotFromConfirmedHistory` returns `false` for any `HardForkSummary` or `TimeToSlot` error, not only the past-horizon case its doc comment describes, so an internal failure silently downgrades what used to be a hard startup failure. `hardfork.ErrPastHorizon` is typed and can express the intent; detail in the inline comment at `ledger/hardfork_summary.go:252`.

**Disproved:** that the node "proceeds into block production" with an unjudged certificate. `NewBlockForger` refuses production mode when `validatedKESProtocolLifetime()` fails (`ledger/forging/forger.go:448`), and the per-slot gate rejects both `kesPeriod < opCertStart` and `kesPeriod >= opCertExpiry` before Praos leader selection (`forger.go:867-906`). Extrapolation through the newest confirmed era skews the computed period *downward*, so a certificate that is expired or staged for the future lands below `opCertStart` rather than inside `[start, expiry)` — every case I constructed fails closed. The forge skips are logged at Error level, not silent.

**Also valid, separately:** nothing re-runs the strict check, and nothing tests either branch. Blocking finding at `node_forging.go:75`.

wcatz: Both blockers addressed in 6f74cb9. The revert-and-rerun was the right way to find it — "all 58 lines can be deleted with the suite still green" is not something I could argue with.

**Blocker 1: the behavior change now has tests.** I took the extraction you proposed verbatim:

```go
func (n *Node) validateBlockProducerStartupForClock(
	slot uint64,
	supported bool,
) (*forging.PoolCredentials, error)
```

`validateBlockProducerStartup` is now routing over `WallClockSlotFromConfirmedHistory`, and all branches are reachable from `newTestNodeForBP`. Three tests, with your (2) as the load-bearing one:

- `..._SupportedStillRejectsExpiredOpCert` — `supported == true` with an expired certificate must still fail startup.
- `..._DeferredArmsInsteadOfFailing` — the *same* genesis and credentials, `supported == false`, must succeed **and** come back with a non-zero `OpCertExpiryPeriod()`. Asserting the arming and not just the nil error matters, because the per-slot gate fails closed on a zero expiry: unarmed credentials would mean unable to forge at all, not deferred.
- `..._DeferredStillValidatesMaterial` — the deferral does not extend to credential loading.

Using one genesis for the first two means the only difference between them is `supported`, which isolates the change to that condition.

Rather than assert the coverage is load-bearing, I mutated the code:

```
ignore `supported` (always defer):
  --- FAIL: ..._SupportedStillRejectsExpiredOpCert
drop the ArmKesProtocolLifetime call:
  --- FAIL: ..._DeferredArmsInsteadOfFailing
      expected the KES protocol lifetime to be armed; the forger's
      per-slot gate fails closed on a zero expiry
```

Each mutation fails exactly the test meant to catch it, and reverting restores green.

**Blocker 2: `ARCHITECTURE.md`.** Updated at the section you pointed to. It now records that the startup KES-period judgement is conditional on confirmed era history spanning the wall clock, that only the past-horizon case is a deferral while other failures stay hard, what the deferred state still validates and arms, and that the per-slot gate is the enforcement there — including the downward-skew reasoning you independently derived, since that is *why* the deferred state is safe rather than merely tolerable.

**`hardfork_summary.go:252` — taken, and it is the better half of cubic's finding.** Now `(uint64, bool, error)`, with only `hardfork.ErrPastHorizon` reporting false. This also dissolves the fragile invariant you flagged: before-genesis reaches it as `hardfork.ErrBeforeGenesis` and is now an error regardless, so the behavior no longer depends on that sentinel staying in agreement with `ledger.ErrBeforeGenesis`. `TestWallClockSlotFromConfirmedHistory_EmptyCache` changed expectation as a result — an empty cache is now an error rather than "unsupported" — which is the point of the fix, so I made the test say so explicitly.

**`keys.go:1067` — taken.** `ArmKesProtocolLifetime` returns `errors.New("signing material not loaded")` in that branch, with a test. And `keys.go:1055` deliberately left alone: you are right that preserving `previousOpCertValidation` there is what stops an unloaded credential reporting a usable lifetime, and I did not want that reasoning silently reversed by a drive-by fix.

**`node_forging.go:87` (`GetOpCert()` nil deref) — guarded**, matching the sibling at line 145. Also `ChainTipSlot()` is now nil-checked, since the extracted function is reachable without a `LedgerState`.

**`node_forging.go:91` (log wording) — reworded.** It now leads with the unconditional clause: the opcert KES lifetime is re-checked per slot and forging cannot proceed while it fails; sync progress *additionally* gates forging whenever an upstream peer is active. Your point that this is what an operator reads while deciding whether to accept a deferred check is the reason I changed it rather than filing it.

**Outstanding, and I am not claiming otherwise: no DevNet run.** You are right that `AGENTS.md` asks for one on any change touching block production or opcert verification, and the description recorded none because none was done. I am starting a DevNet run now and will post the result here; if it turns up anything, that lands before this is merged rather than after.

**Gate re-run on 6f74cb9:** `go build ./...` and `-tags dingo_extra_plugins` clean, `go vet ./...` clean, `go test -race -count=1 ./ledger/forging/ ./ledger/ .` pass, `golangci-lint run ./...` 0 issues, `make docs-parity` and `go test ./internal/architecture` pass, `make golines` no-op.

On the CodeRabbit point — noted, and useful. I had read the green check as a review on the other PRs in this batch too; "Review limit reached" behind a SUCCESS status is worth knowing about.

wcatz:
**DevNet run, as promised.** Passing, with one limitation I want to state rather than let the word "DevNet: pass" imply more than it does.

Default all-dingo network (three producers + relay + txpump) on `6f74cb9`, via `internal/test/devnet/run-tests.sh`.

**Result: no failures.** `internal/test/devnet` suite `ok` in 42.9s, 0 `FAIL` anywhere in the run.

All three producers cleared block-producer startup validation and are forging:

```
node_forging.go:179 msg="block producer credentials validated"
  pool_id=pool1rjlgh044z29kdh4xfxxg5e5trjyqt3qp5stq6zqjsvpdxmkn8w2
  current_slot=0 current_kes_period=0 opcert_kes_period=0 opcert_expiry_period=62
node_forging.go:292 msg="block producer pool registration verified on chain"

dingo-1  opcert_kes_period=0  opcert_expiry_period=62   forged 20
dingo-2  opcert_kes_period=0  opcert_expiry_period=62   forged 15
dingo-3  opcert_kes_period=0  opcert_expiry_period=62   forged 21
```

Non-zero `opcert_expiry_period` on all three is the armed protocol lifetime, so `NewBlockForger` accepted the credentials and the per-slot gate has its data. Chains agree — all three at `slot=137`, with `msg="chain extended, new tip: a8b14880... at slot 135" source=forged` and the leader schedule precomputing normally.

**The limitation.** This exercised the **strict** branch, not the deferred one. `node_forging.go:179` is `validateBlockProducerStartupAtSlot`, i.e. `supported == true`. I checked directly, and the deferral warning fired zero times across all three producers:

```
$ for n in 1 2 3; do docker logs ...dingo-$n-1 | grep -c "deferring the operational-certificate"; done
0
0
0
```

That is inherent to DevNet rather than a gap in the run: the network is created at genesis, so confirmed era history spans the wall clock from the first slot and there is nothing to defer. Reproducing the deferred state needs a node whose confirmed history lags the wall clock — a mainnet genesis import or a long-downtime restart — which DevNet does not model.

So the honest split of what is verified by what:

- **DevNet**: the strict path still validates, arms, and forges end to end, with real consensus across three producers. It confirms this change did not break the path that was already working.
- **Unit tests** (`validateBlockProducerStartupForClock`): the deferred branch, and the negative property that `supported == true` still rejects an expired certificate. Those remain the only coverage of the deferral, which is why the mutation checks on them mattered.

Two incidental observations, neither related to this change:

- Early `txsubmission` errors from `txpump` while the chain was still syncing: `conway utxo validation rule 33: value not conserved: consumed 0, produced 700000000000` against inputs that do not resolve yet. A txpump startup race against genesis UTxO availability; it settles once sync completes.
- One startup `WARN forger.go:618 msg="slot clock not ready, retrying" error="time is before genesis start" retries=1`, which then recovered. Worth noting only because it is the before-genesis sentinel path this PR touches, and it behaves as intended.

wcatz:
**Correction to my previous comment: the DevNet run did not pass cleanly, and my "0 failures" was wrong.**

I read the log while only the `internal/test/devnet` unit package had completed. The live `internal/test/devnet/scenarios` package then ran for another 725s and failed. Reporting the real result.

**One test failed:** `TestChainGrowthRate` (126.98s). Final tally: 88 PASS, 1 FAIL, 2 SKIP; `run-tests.sh` exited 1.

```
chain_growth_test.go:94: Condition never satisfied
  dingo-1 did not reach slot 152 and block 17 within 1m52.5s

growth trace: slot 120 block 0 -> slot 131 block 1 -> slot 137 block 4
           -> slot 146 block 5 -> slot 153 block 6 -> slot 162 block 11
```

The test measures a 100-slot window and requires at least 8 blocks of growth (20% of the ~40 expected at `activeSlotsCoeff=0.4`, deliberately conservative for VRF variance). The chain did pass the target slot but was short on blocks inside the window.

**Everything that matters for consensus passed**, which is why I want to be precise rather than just say "DevNet failed":

| test | result |
|---|---|
| `TestEpochBoundaryConsensus` | PASS (438.15s) |
| `TestSustainedConsensus` | PASS (102.14s) |
| `TestBasicBlockForging` | PASS (42.07s) |
| `TestDingoChainAdvances` | PASS (16.06s) |
| `TestRelayPropagation` | PASS |
| `TestHarnessWaitForSlot` | PASS (42.01s) |
| `TestChainGrowthRate` | **FAIL** |

`VerifyChainConsensus` had all four nodes agreeing at slot 705 / block 155 across the epoch boundary, so the network reached and crossed epoch 1 in agreement.

**On causation I am not going to hand-wave.** My change only decides whether the startup opcert KES check runs strictly or is deferred, and this run took the strict path and passed it (`node_forging.go:179 msg="block producer credentials validated"`). Nothing in the diff touches the forge loop, the sync gate, or chain growth after startup, so a causal path to a block-rate shortfall is hard to construct. But that is an argument, not evidence.

There is a plausible innocent explanation visible in the logs — the chain produced no blocks at all until ~slot 131 because the sync gate suppressed early forges (`forger.go:1505 msg="chain syncing from peer, skipping forge" leader_slot=true` at slots 14, 22, 26, 29), which eats into a fixed measurement window. That is `forger.go`, not this PR. Also `txpump submitted=3 rejected=6`, so the mempool was near-empty for much of the run.

Still an argument. **I am running the same suite on `origin/main` to establish a baseline** — the revert-and-rerun method you used on this PR is the right instrument here too. I will post the result either way. If `TestChainGrowthRate` also fails on `origin/main`, it is pre-existing or flaky; if it passes there and fails here, that is my problem to fix and I would rather find it now.

Artifacts from the failure are at `/tmp/dingo-devnet-artifacts.rz5zFy/TestChainGrowthRate` on my machine if anything specific would be useful.

wcatz:
**Baseline result, and I traced the cause. It is not this change — the branch is 8 commits behind `main` and missing two fixes that match the failure exactly.**

`origin/main` (`31dd4a8a`), same suite, same machine:

```
--- PASS: TestChainGrowthRate       (120.07s)
--- PASS: TestBasicBlockForging     (38.08s)
--- PASS: TestDingoChainAdvances    (14.03s)
--- PASS: TestSustainedConsensus    (106.14s)
--- PASS: TestEpochBoundaryConsensus (534.12s)
--- PASS: TestRelayPropagation
--- PASS: TestHarnessWaitForSlot    (32.02s)
ok  github.com/blinklabs-io/dingo/internal/test/devnet/scenarios  812.559s
```

Baseline is green. So "probably flaky" was not good enough and I went into the failure artifacts.

**What actually happened was a chain halt, not slow growth.** Comparing the measurement windows:

```
baseline:   slot 51  block 23  ->  slot 151 block 52     (+29 blocks)
this branch: slot 52 block 9   ->  slot 57  block 11
                                    slot 162 block 11     (+0 blocks for ~105 slots)
```

The chain stopped extending after block 11 and never resumed. `dingo-1.log` from `artifacts/TestChainGrowthRate` says why:

```
level=ERROR msg="unexpected connection failure: protocol error: add chainsync
  client for connection ...: GetChainFromPoint failed: block not found"
level=ERROR msg="failed to start chainsync client"
level=WARN  msg="chain fork detected"
level=WARN  msg="failed to start blockfetch after local rollback recovery"
level=WARN  msg="short-lived connection detected, applying backoff"
```

Repeating. A fork triggered local rollback recovery, blockfetch failed to restart, and chainsync could not re-establish because the point it asked for was gone. Nothing in that path is opcert or KES startup validation — and `grep -iE "kes|opcert"` over the whole log returns nothing outside the config dump. Forge dispositions were only `not leader for slot` (33) and `chain syncing from peer` (9); the producer never rejected its own certificate.

**Then the provenance.** This branch's merge-base is `7ad27910`, and it does not contain `31dd4a8a`:

```
$ git merge-base --is-ancestor 31dd4a8a HEAD   ->  false
$ git log --oneline HEAD..origin/main
31dd4a8a fix(ledger): guard bufferedHeaderEvents with the blockfetch mutex (#4032)
7006ebe7 fix(chain,ledger): bar rollback from uncommitted indices, record skipped restores (#4002)
eabf653d fix(database): preserve failed ledger close results (#3913)
... 5 more
```

Two of those are the observed failure:

- **#4032** guards `bufferedHeaderEvents` with the blockfetch mutex — against `failed to start blockfetch after local rollback recovery`.
- **#4002** bars rollback from uncommitted indices — against `chain fork detected` followed by `GetChainFromPoint failed: block not found`.

The baseline has both. This branch has neither. That is why the baseline passed and this did not, and it also means my previous comment's "sync gate ate the measurement window" guess was wrong — the real mechanism is a rollback-recovery deadlock that is already fixed upstream.

**Verification in progress.** I rebased this branch onto `origin/main` in a throwaway worktree (clean, no conflicts, both commits replayed onto `31dd4a8a`) rather than force-pushing a branch you are mid-review on. `go build ./...`, `-tags dingo_extra_plugins`, `go vet ./...` and the new tests all pass there, and the DevNet suite is running now. I will post the result.

If it comes back green, the branch just needs a rebase before merge and I will ask before force-pushing. If `TestChainGrowthRate` still fails with #4032 and #4002 present, then it is mine after all and I will say so.

wcatz:
**Retracting my last comment's diagnosis. The rebase did not fix it, so "missing #4032 and #4002" was wrong.**

Rebased onto `origin/main` (both commits replayed onto `31dd4a8a`, so #4032 and #4002 are present), same suite:

```
--- FAIL: TestChainGrowthRate (142.98s)
    dingo-1 did not reach slot 156 and block 26 within 1m52.5s
FAIL github.com/blinklabs-io/dingo/internal/test/devnet/scenarios  813.524s
```

So the two fixes I identified are not the cause. That is now two conclusions I have posted on this PR that did not hold — the sync-gate explanation and the missing-upstream-fixes explanation — and I would rather say so plainly than offer a third.

**What I can report as fact is that the two failures had different mechanisms.**

Pre-rebase run — chain **halted** and never resumed:

```
slot 52 block 9 -> slot 57 block 11 -> slot 162 block 11   (+0 for ~105 slots)
ERROR "GetChainFromPoint failed: block not found" / "failed to start chainsync client"
WARN  "chain fork detected" / "failed to start blockfetch after local rollback recovery"
```

Post-rebase run — chain **rolled back to origin** and re-forged:

```
slot 56  block 18      <- startTip, so minBlock = 26
slot 0   block 0       <- tip regressed to genesis
slot 60  block 0
slot 162 block 22      <- net +4 against a required +8
```

Different failure modes, and neither is a block-rate deficiency: one is a chainsync deadlock, the other is a rollback that ate the measurement window. Everything else passed in both runs, including the heavyweight ones — post-rebase: `TestEpochBoundaryConsensus` 528s, `TestSustainedConsensus` 94s, `TestBasicBlockForging`, `TestDingoChainAdvances`, `TestRelayPropagation`, `TestHarnessWaitForSlot` all PASS.

**Where that leaves attribution, stated honestly.** Sample sizes are 1 pass on `origin/main` and 2 failures on this branch, with two unlike mechanisms. I do not think a change that only selects between strict and deferred startup opcert validation can produce either a blockfetch deadlock or a rollback to origin — the strict path ran and passed in both runs, and neither log contains any KES or opcert line outside the config dump. But 1-versus-2 with unlike mechanisms is not evidence I am willing to lean on, and I am not going to keep proposing mechanisms.

**So I am gathering repeated samples instead of theorising.** Running `origin/main` again now. If the baseline also fails, the test is environment-sensitive here — worth noting that my machine has been under sustained load from other work in this session, and a 1s-slot devnet at `activeSlotsCoeff=0.4` is plausibly sensitive to that. If the baseline passes repeatedly while this branch fails repeatedly, then it is mine and I will find the mechanism rather than guess at it.

I will post counts, not another theory. Nothing here is a reason to merge before that resolves.

wcatz:
Counts, as promised — and they retract both of my earlier theories about this.

DevNet `run-tests.sh`, four full runs, `internal/test/devnet` plus
`internal/test/devnet/scenarios`:

| run | PASS | FAIL | SKIP | top-level failures |
|---|---|---|---|---|
| baseline (main) #1 | 125 | 0 | 2 | — |
| **baseline (main) #2** | **120** | **5** | 2 | `TestChainGrowthRate`, `TestSustainedConsensus`, `TestEpochBoundaryConsensus`, `TestStakeAddressInfoQueries` (+1 subtest) |
| this branch | 124 | 1 | 2 | `TestChainGrowthRate` |
| this branch, rebased | 124 | 1 | 2 | `TestChainGrowthRate` |

`TestChainGrowthRate` fails on unmodified `main`. It is not attributable to this
PR, and the branch (1 failure) came out ahead of the second baseline (5).

The failure signature is identical in all three failing runs — not the two
different mechanisms I claimed earlier:

```
chain_growth_test.go:94:
    Error: Condition never satisfied
    Test:  TestChainGrowthRate
```

with the harness stalled at a low block height in each case (`dingo-1 at slot 51,
block 15` on baseline #2; `slot 162, block 11` and `slot 162, block 22` on the
branch). Same test, same line, same assertion.

Withdrawing two things I said on this PR:

1. That `TestChainGrowthRate` implicated this branch. It does not — main
   reproduces it.
2. That the rebased branch failed by a different mechanism (rollback-to-origin
   versus chain halt). It did not; all three are the same `chain_growth_test.go:94`
   condition timeout. That claim came from reading artifacts rather than
   comparing the assertions, and it was wrong.

What I can support: two runs of the same baseline commit produced 0 and 5
failures, so the scenarios suite is flaky at least around chain growth, epoch
boundary and sustained consensus. Two samples is enough to rule this PR out as
the cause — a test that fails on unmodified main cannot be caused by a branch —
but not enough to quantify a rate, and I am not going to extrapolate one from
n=2 after being wrong twice here already.

Not proposing a fix for the flakiness in this PR; happy to file it separately
with the four failure artifacts attached.

node_forging.go
wcatz: Guarded in `6f74cb91` — `node_forging.go:112-114`:

```go
opCert := creds.GetOpCert()
if opCert == nil {
	return nil, errors.New("block producer operational certificate is nil")
}
```

Same wording as the sibling at `175-177`. I also nil-checked `ChainTipSlot()` at `116-118`, and that one is not merely consistency: the extracted `validateBlockProducerStartupForClock` is reachable without a `LedgerState` from the test harness, so it is live.

On `nilaway` — you were right that the CI `lint` job doesn't run it, and it was also missing from the gate list I posted on the blocker thread. `golangci-lint` is not a substitute. It turned out not to be installed on my build host at all, so that claim was unevidenced; I'm installing it and will post the actual result rather than assert it.

node_forging.go
wcatz: Reworded in `6f74cb91` — `node_forging.go:121-124` now leads with the unconditional clause:

> The opcert KES lifetime is re-checked per slot and forging cannot proceed while it fails; sync progress *additionally* gates forging whenever an upstream peer is active.

Your reading of the gate is what the code says — `ledger/forging/forger.go:823` is `if upstreamActive && (upstreamTip == 0 || …)`, so a producer with no active upstream falls straight through to the KES gate. Since this text is what an operator reads while deciding whether a deferred check is acceptable, I reworded it rather than filing it.

ledger/hardfork_summary.go
wcatz: Taken, and this was the half of cubic's finding worth acting on. `ledger/hardfork_summary.go:261`:

```go
func (ls *LedgerState) WallClockSlotFromConfirmedHistory() (
	uint64, bool, error,
)
```

Only `errors.Is(err, hardfork.ErrPastHorizon)` reports `false` (line 272); every other error from either call propagates instead of silently downgrading a hard startup failure into a warning.

That also dissolves the fragile invariant you flagged rather than just documenting it: before-genesis now arrives as `hardfork.ErrBeforeGenesis` and becomes an error regardless, so the behavior no longer depends on the two distinct sentinels staying in agreement. `TestWallClockSlotFromConfirmedHistory_EmptyCache` changed expectation as a direct consequence — an empty cache is now an error rather than "unsupported" — which is the point of the fix, so the test states that explicitly instead of quietly tracking the new behavior.

ledger/forging/keys.go
wcatz: Taken — `ledger/forging/keys.go:1071` returns `errors.New("signing material not loaded")` in that branch, with a test, and your reasoning is recorded at the site (`1062-1069`) so it isn't reversed by a later drive-by fix.

On cubic's version: not taken, and the current code forecloses it. `ArmKesProtocolLifetime` resets `opCertValidated = false` **unconditionally** at `1053`, before any branch runs — so there is no `previousOpCertValidation` left to preserve. That is the fail-closed behavior you described: `OpCertExpiryPeriod()` reads 0 and the forger's `validatedKESProtocolLifetime` refuses, rather than an unloaded credential reporting a usable lifetime.

node_forging.go
wcatz: Status on the DevNet run I said I was starting — **deferred, not dropped.** I'd rather say so than let it go quiet.

The build host is currently mid-restore of a mainnet core database: production is down, and that restore is the only path back up. An earlier attempt on the same box was OOM-killed. DevNet brings up three producers plus a relay and `txpump`, so starting it alongside the restore risks killing it and costing far more than the run is worth right now.

I'm holding it until the host is free, then posting the result here — still before merge, as I said. The commitment stands; only the timing moved.

node_forging.go
wcatz: `nilaway` result, as promised — and it backs the guard rather than just asserting it.

It wasn't installed on my build host, so I installed `go.uber.org/nilaway/cmd/nilaway@latest` and ran the same finding set at both ends of the branch:

```
$ nilaway . ./ledger/forging ./ledger
  6f74cb91 (head): 14 findings
  7ad27910 (base): 14 findings
  comm -23 head base  ->  empty      # no new finding introduced by the branch
```

**Zero findings in `node_forging.go`, `ledger/forging/keys.go`, or `ledger/hardfork_summary.go`** — the three files this branch touches. The 14 are identical at both commits and all pre-existing: `ledger/view.go:1217` (deep read of `latest[key]` lacking a guard) plus unassigned-variable and nil-return patterns in `forging/opcert_width_test.go`, `forging/store_test.go`, `ledger/leios_apply_double_consume_test.go`, `ledger/queries_limit_test.go`, `ledger/recovery_rewind_window_test.go`, and `ledger/utxo_storage_test.go`.

**Scope caveat, stated rather than glossed:** I ran the three relevant packages, not the full `./...` that `make lint` uses. The host is mid-restore of a mainnet core database and I didn't want to add that much memory pressure to it. So this shows the branch introduces no new `nilaway` finding and leaves the touched files clean; it is not a full-tree run. I'll fold `./...` into the same pass as the DevNet run once the host is free.

Worth recording that your prediction was right in substance: `nilaway` is not in the CI `lint` job, and `ledger/view.go:1217` is exactly the kind of thing it catches while CI stays green. That one is pre-existing and outside this PR, so I'm not touching it here — happy to file it separately if you'd like it tracked.

node_forging.go
wcatz: Unplanned production evidence for this change, which I'd rather report than sit on.

I had to restore a mainnet block producer from a snapshot today. The restored ledger tip was ~440k slots behind the wall clock at startup, so the node started straight into this branch's deferral path — and logged it:

```
level=WARN msg="block producer startup: confirmed era history does not span the
current wall-clock slot yet; deferring the operational-certificate KES-period
plausibility check until the ledger catches up. Credential material is validated;
forging stays gated on sync progress and re-checks the opcert KES lifetime per slot."
component=node tip_slot=196790151 opcert_kes_period=1486 opcert_expiry_period=1548
```

**Scoping this honestly:** that is the `da662b06` wording, not the reworded text in `6f74cb91`, so the deployed binary carries the fix commit but predates the reword and the extraction refactor. It is not a test of head. I verified the provenance rather than assuming it — the string count for `"confirmed era history does not span"` in `node_forging.go` is `0` at the merge base `7ad27910` and `1` at `da662b06`, so this log line comes from this branch and nowhere else.

What it does demonstrate, on mainnet rather than in a harness:

- The deferral path let a block producer start with confirmed era history behind the wall clock, where the pre-change code would have failed startup outright.
- **`opcert_expiry_period=1548` is non-zero**, which is the part I care about most. That is `ArmKesProtocolLifetime` having actually armed — precisely what `TestValidateBlockProducerStartupForClock_DeferredArmsInsteadOfFailing` asserts, and the reason I argued asserting the arming matters rather than just the nil error. A zero there would have meant the per-slot gate fails closed and the node cannot forge at all; deferred-but-armed is the intended state, and that is what production shows.
- The node then synced normally through the backlog with the forger started in production mode, no errors from the gate.

This is not a substitute for the DevNet run I still owe you — it exercises one branch of three, on one era boundary, with no adversarial cases. I'm treating it as corroboration, not coverage. DevNet still lands before merge; the host was tied up by exactly this restore, and now that it's finished I'll run it once the node has caught up (I don't want to slow its sync while it's still ~3h behind).

wcatz: Correction to my previous comment — one piece of the evidence I gave was wrong, and I'd rather flag it than let it stand.

I wrote that during the read-only validation the node "applied straight through 196791729 … 120,343 slots past it," citing chainselection's `observed_slot` climbing to 196912072. **`observed_*` is not our applied tip.** It is per-peer state. The proof came from the same node an hour later: a line reading `observed_block_number=13908573 observed_slot=197188274` while the node's actual tip metric was 196845785 — 342k slots lower. So `observed_slot` cannot be our position, and my 120,343-slot figure was a misreading of that field.

What still holds, from `cardano_node_metrics_slotNum_int` (which *is* our tip): the node started at 196790151 and reached 196845785, so it genuinely applied through 196791729 — the slot that had been halting production — and 54,056 slots beyond it. That conclusion is unchanged; only the supporting number was bad.

The rest of the comment is unaffected: the deferral WARN with `opcert_expiry_period=1548` came from this node's own startup log, and the `da662b06`-vs-base string check that establishes its provenance stands.

Unrelated to this PR, but relevant to anyone reading the log excerpt: that node later deadlocked on a `ledger.blockfetch` EventBus stall when a blockfetch timeout raced a ledger rollback. Separate issue, separate report — it is not caused by this branch (the deferral had already completed at startup and the node had been applying blocks normally for 25 minutes).

wcatz:
@chrisguiney all three blocking findings are addressed in 6f74cb91, plus both non-blocking ones. Summary, since the review predates that commit:

**Missing test for the branching.** `validateBlockProducerStartupForClock(slot uint64, supported bool)` extracted exactly as you suggested, so `validateBlockProducerStartup` is now a thin wrapper over `WallClockSlotFromConfirmedHistory`. Three branches covered from the existing harness:
- `TestValidateBlockProducerStartupForClock_SupportedStillRejectsExpiredOpCert` — the negative you called the one that matters
- `TestValidateBlockProducerStartupForClock_DeferredArmsInsteadOfFailing`
- `TestValidateBlockProducerStartupForClock_DeferredStillValidatesMaterial`

I re-ran your revert-and-rerun in the other direction to confirm the coverage is load-bearing: collapsing the deferral to always take the strict path (`if supported` -> `if true`) fails `_DeferredArmsInsteadOfFailing`. The lines can no longer be deleted with the suite green.

**Error conflation.** `WallClockSlotFromConfirmedHistory` is now `(uint64, bool, error)` and switches on `errors.Is(err, hardfork.ErrPastHorizon)`. Only past-horizon defers; an empty epoch cache, a `BuildSummary` failure, or a before-genesis clock propagates as an error and keeps the hard startup failure. Covered by `_SupportedWhenEraSpansNow`, `_UnsupportedWhenPastHorizon`, and `_EmptyCache`. You were right that this was the half of Cubic's finding worth acting on.

**ARCHITECTURE.md.** Updated — it now records that the startup KES-period judgement is conditional on confirmed era history spanning the wall clock, names the past-horizon case as the only deferral, and points at the per-slot gate as the enforcement in the deferred state.

**Non-blocking, both taken.** `GetOpCert()` nil guard added to match the sibling path. `ArmKesProtocolLifetime` now returns an error in the unloaded branch rather than reporting success while arming nothing — your disposition, not Cubic's, since preserving a stale `opCertValidated` would let an unloaded credential report a usable lifetime.

**DevNet: not yet run, and I'm not going to claim otherwise.** The only build host is currently mid-Mithril rebootstrap of the mainnet block producer (~46% through a block copy, several hours left), and a DevNet run would contend with it. Queued to run once that finishes; I'll report the result here before asking for a merge.

wcatz:
Additional evidence from production, which widens the case for this change beyond how the description currently frames it.

I pulled the actual journal from the mainnet block producer that hit this. Two things worth correcting into the record:

**1. The failure is a not-yet-valid rejection, not an expiry.** The exact error, repeating in a crash loop:

```
block producer startup validation failed: validate KES period:
  opcert KES period 1486 is in the future (current 108)
```

Alongside it, `block producer credentials validated` — the credential material was never in question. The certificate was valid the whole time: decoding it gives KES period 1486 and counter 28, and with mainnet's `slotsPerKESPeriod: 129600` / `maxKESEvolutions: 62` the valid window is `[1486, 1548)`, against a true current period of ~1524. Nothing was expired or near expiry. Anywhere this incident is described as an expired cert/key pair is wrong.

**2. The computed `108` is exactly the Byron extrapolation, to the digit.** Mainnet `byron-genesis` has `startTime: 1506203091` and `slotDuration: 20000`ms. With only Byron confirmed, extrapolating 20s slots across ~8.95 years gives `(now - 1506203091) / 20 ~= 14.12M` slots, and `14125313 / 129600 = 108.99` -> 108. That is the mechanism this PR describes, confirmed arithmetically against real logs rather than inferred.

**3. The exposure is wider than "fresh import".** This is the part I think the description undersells. The deferral also fired on a node that was nearly caught up:

```
2026-09-07T15:32:57Z WARN block producer startup: confirmed era history does not
  span the current wall-clock slot yet ... tip_slot=196790151
  opcert_kes_period=1486 opcert_expiry_period=1548
```

Tip 196790151 against a wall clock around 197.16M — a gap of ~370k slots. Since the forecast horizon is ~3k/f = 129600 slots, **any producer more than roughly 1.5 days behind the wall clock is in this state**, not just one importing from genesis. That makes the startup deadlock reachable on an ordinary restart after a weekend outage, which is a materially different risk profile than a one-off bootstrap edge case.

**Why this is mainnet-only**, for whoever wonders why no other network shows it: preview, devnet and musashi all have `slotLength: 1` in their Shelley genesis with no preceding Byron era, so extrapolating the newest confirmed era's slot length yields the *correct* slot. Mainnet is the only network with a 20s -> 1s transition. A preview producer forging normally is therefore expected under this mechanism and does not contradict it.

Happy to fold points 1 and 3 into the PR description if you'd prefer them there rather than in a comment.

wcatz:
DevNet validation run against this PR's head (`42d096f7a0f77a905c7132f36fd6a54ee54d1fde`), both modes, on a 32-core host.

**Default (all-dingo network)** — `./run-tests.sh`

```
ok  github.com/blinklabs-io/dingo/internal/test/devnet/scenarios  720.158s
[run-tests] txpump submitted=8 rejected=0 error=0
```

131 passed, 0 failed. Clean teardown, no leftover containers.

**Accelerated** — `./run-tests.sh --accelerated`, the mode that restarts a producer post-genesis

```
--- PASS: TestAcceleratedScenarioTimeline (120.64s)
[run-tests] txpump submitted=6 rejected=0 error=0
```

The producer-restart and relay-restart phases both recovered to full chain agreement:

```
peer interruption: after recovery: slot 172 agree=true
  [dingo-1=212dc5bb… dingo-2=212dc5bb… dingo-3=212dc5bb… dingo-relay=212dc5bb…]
peer interruption: dingo-3 recovered (connects 1->2, tip slot 181, block 65)

relay restart: after recovery: slot 187 agree=true
  [dingo-1=3b6fdfe6… dingo-2=3b6fdfe6… dingo-3=3b6fdfe6… dingo-relay=3b6fdfe6…]
relay restart: dingo-relay recovered (connects 1->2, tip slot 195, block 71)

accelerated scenario completed in 2m0.604s (hard timeout 5m0s)
```

---

### Production corroboration of the premise

Separately, a mainnet block producer hit this exact failure mode today, which may be useful for the review discussion.

**To be clear about what this does and does not show: that node runs a build that does _not_ include this PR.** So the following is evidence for the premise this PR is built on — that the preflight check fires on a valid cert purely because history is behind the clock — not validation of the fix itself.

The node was ~120k slots behind the wall clock and rejected its opcert as being in the future. After it caught up, the *same binary* and the *same cert* validated cleanly:

```
block producer credentials validated
  current_kes_period=1526  opcert_kes_period=1486
  opcert_counter=28        opcert_expiry_period=1548
block producer pool registration verified on chain
block forger started in production mode with leader election
```

`opcert_kes_period=1486` sits well inside `opcert_expiry_period=1548`, and did so the whole time. Nothing about the key material changed between the rejection and this line — only the tip gap. The reported "current" KES period differs from the true one by the Byron 20s-slot arithmetic applied to a stale tip, which is what makes the check horizon-dependent rather than cert-dependent.

Practical consequence: recovering that node required a two-phase start (serve without `--block-producer` to close the gap, then enable forging) precisely because the preflight cannot be satisfied while behind. That is the operational cost this PR removes.

wolf31o2 (approved):

PR #4037, opened 2026-09-06
feat(config): add skipRewardLiveStakeBackfillCheck

wcatz:
Split out of #4035, which originally bundled this with the DRep deposit fix. Unrelated concerns, so it gets its own PR.

## Problem

`backfillRewardLiveStake` runs on every node startup, after database recovery and before ledger processing. The expensive part is not the repair — it is deciding whether a repair is needed at all.

`RewardLiveStakeNeedsBackfill` (`database/plugin/metadata/sqlstore/live_stake.go:445`) does two unbounded scans:

1. `SELECT credential_tag, staking_key, amount FROM utxo WHERE deleted_slot = 0 AND staking_key IS NOT NULL ...` — every live UTxO row, accumulated into an in-memory map.
2. A `UNION` of all `account` and all live `utxo` stake credentials, `LEFT JOIN`ed against `reward_live_stake`.

On a mainnet-scale database that is tens of millions of rows scanned twice, and it happens on every single start regardless of the outcome. A node that is restarted repeatedly — which is exactly what happens while investigating an unrelated problem — pays close to the price of a full rebuild each time, before it can begin syncing.

## Change

Adds `skipRewardLiveStakeBackfillCheck` (YAML), `--skip-reward-live-stake-backfill-check` (CLI), `CARDANO_SKIP_REWARD_LIVE_STAKE_BACKFILL_CHECK` (env). When set, `backfillRewardLiveStake` logs a warning and returns without running the check.

**Default is `false`** — the check keeps running unless an operator opts out deliberately. `TestSkipRewardLiveStakeBackfillCheckDefaultsToRunningTheCheck` pins that, so the safe default cannot regress silently.

## Why this is a diagnostic control, not a tuning knob

The check is what catches a stale or pre-migration `reward_live_stake` table. Skipping it is safe only against a database already known to be consistent, and unsafe to leave enabled permanently. Both the config doc comment and the `node.go` doc comment say so, and the node logs a warning on every startup where it is active so an operator who left it on will see it.

## Not a provenance-gated field

Deliberately **not** added to `gatedFieldPaths` in `internal/config/provenance.go`. That set is the fields that gate persisted node-settings state (`Network`, `StartEra`, `PledgeLeverage`, `ValidateHistorical`, `StrictUtxoValidation`, ...). This option persists no gate state and does not change how the chain is validated — only whether one startup check runs. I tried adding it there first; `TestGatedFieldEnvProvenance` correctly rejected it, since the gated set must match the extractors that key persisted gate state by the same dotted paths.

## Env var prefix

`CARDANO_`, not `DINGO_`. The field carries `split_words:"true"` with no explicit `envconfig` tag, so `envconfig.Process("cardano", cfg)` derives the name from the field name. I verified this by running it rather than reading the tags — some neighbouring options in `dingo.yaml.example` use `DINGO_` via an explicit tag, so the prefix is genuinely not uniform and worth pinning. `TestSkipRewardLiveStakeBackfillCheckEnvBinding` is that pin.

## Tests

- `TestSkipRewardLiveStakeBackfillCheckEnvBinding` — the env var sets the field.
- `TestSkipRewardLiveStakeBackfillCheckDefaultsToRunningTheCheck` — the check runs by default.
- `TestRegisterFlags_CoversAllExportedConfigFields` already required the new field to have a flag registered; it passes.

## Documentation

- `dingo.yaml.example` — documented alongside `strictUtxoValidation`, with the env var and CLI names.
- `ARCHITECTURE.md` — noted in the `RewardLiveStakeNeedsBackfill` startup section, including why it is not provenance-gated.
- `DATABASE.md` — noted on the `reward_live_stake` table row, which previously stated unconditionally that startup compares and rebuilds on mismatch.

## Verification

`go build ./...` and `go build -tags dingo_extra_plugins ./...` clean. `go test ./internal/config/ .` passes. `make import-boundaries`, `golangci-lint run ./...` (host and `GOOS=windows`), `make docs-parity` and `make golines` all clean.

Pre-existing and unrelated: `ledger/leader` `TestThresholdPreservesProtocolBoundary` fails on clean `origin/main` as well.

---
## Summary by cubic
Adds `skipRewardLiveStakeBackfillCheck` so operators can skip the costly `reward_live_stake` startup consistency check when a database is already known to be consistent. The check scans every live UTxO row twice on each startup, so repeated restarts during debugging pay close to the cost of a full rebuild before syncing can begin. The skip applies only to that probe and the rebuild it gates — the separate stake-snapshot provenance check in the same startup step still runs unconditionally.

- Configurable via YAML, the `--skip-reward-live-stake-backfill-check` CLI flag, and `CARDANO_SKIP_REWARD_LIVE_STAKE_BACKFILL_CHECK`; the env prefix is `CARDANO_`, not the neighboring `DINGO_`.
- Defaults to `false` with a test pinning that the check runs unless explicitly disabled.
- Logs a warning whenever enabled; it's for diagnostic use only, unsafe to leave on, and deliberately not provenance-gated.
- Both probes now run on the read-only metadata connection, and the rebuild opens a write transaction only when needed, so a skip never contends with block processing on SQLite.
- Documented in `dingo.yaml.example`, `ARCHITECTURE.md`, and `DATABASE.md`.

<sup>Written for commit 1033b6b67443a529765e22b4604899abe2f53a29. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4037?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added an advanced configuration option to skip the startup consistency check for reward live-stake data.
  - Supports configuration through a command-line flag or environment variable; disabled by default.
  - When enabled, startup logs a warning and skips the check and any required rebuild.

- **Documentation**
  - Added guidance explaining the check’s behavior, diagnostic use, safety considerations, and database consistency requirements.


wcatz:

```
```

node.go
```diff
@@ -2140,7 +2140,24 @@ func (n *Node) nodeSettingsGateValues() nodesettings.Values {
...
+// investigation of an unrelated issue), but unsafe to leave enabled
+// permanently since it is what catches a stale or pre-migration
+// reward_live_stake table.
 func (n *Node) backfillRewardLiveStake() error {
+	if n.config.skipRewardLiveStakeBackfillCheck {
+		n.config.logger.Warn(
+			"skipping reward_live_stake backfill consistency check",
+			"component", "node",
+		)
+		return nil
```
chrisguiney: **Blocker.** This returns before `StaleConsensusStakeSnapshotsExist`, so the flag disables a second, unrelated guard that nothing in the naming, the log line, or any of the three documentation sites mentions.

The two checks in this function are not the same kind of thing. `RewardLiveStakeNeedsBackfill` is the expensive full-`utxo` scan this flag exists to avoid. `StaleConsensusStakeSnapshotsExist` is a pair of `EXISTS` probes against `pool_stake_snapshot` and `reward_snapshot` filtered on `calculation_version` (`database/plugin/metadata/sqlstore/live_stake.go:596`) — small tables, indexed on `(epoch, snapshot_type)`, nothing proportional to the UTxO set. It costs approximately nothing, and it fails closed: it aborts startup with *"consensus stake snapshots were produced by an older accounting version and cannot be safely reconstructed from this database; rebootstrap from immutable blocks or a trusted snapshot"*.

Reproduced against this head, seeding one `pool_stake_snapshot` `mark` row at `calculation_version = 0` (current is `2`) and calling `backfillRewardLiveStake` both ways:

```
skip=false -> err=consensus stake snapshots were produced by an older accounting
              version and cannot be safely reconstructed from this database;
              rebootstrap from immutable blocks or a trusted snapshot
skip=true  -> err=<nil>
```

With the flag on, a node whose snapshots came from an older accounting version starts silently and proceeds to reward calculation on state the guard exists to declare unusable. Both production call sites are affected (`node.go:807` startup and `node_lifecycle.go:691` storage reinitialization).

This matters most in exactly the situation the docs recommend the flag for. "Repeated restarts while investigating an unrelated problem" against a long-lived database is the likeliest place to be carrying legacy snapshot rows, and the operator's stated reason for setting the flag — skipping an expensive scan — gives them no signal that they also turned off a correctness gate. The `dingo.yaml.example` text ("skipping is safe when the database is already known to be consistent") is describing `reward_live_stake` consistency; an operator can be confident of that and still be wrong about snapshot provenance, since the two are independently sourced.

Suggested shape: keep the skip on the `RewardLiveStakeNeedsBackfill`/`RebuildRewardLiveStake` pair only, and let the stale-snapshot probe run unconditionally — something like hoisting it out of the early return, or moving the flag check inside the transaction to guard just the `needed` computation. Worth a test on each of the two config paths, since this class of "cheap guard sheltering behind an expensive one" is invisible in a diff.

Also relevant to the PR description's provenance-gating argument: as written, the flag *does* affect whether a validation gate applies, which weakens the "only whether one startup check runs" reasoning. Narrowing the skip restores it.

internal/config/flags_test.go
```diff
@@ -236,6 +236,54 @@ func TestBlockPipelineValidateEnabledEnvBinding(t *testing.T) {
...
+	}
+}
+
+// TestSkipRewardLiveStakeBackfillCheckDefaultsToRunningTheCheck pins the
+// safe default. The check is what catches a stale or pre-migration
+// reward_live_stake table, so an operator has to opt out deliberately; it
+// must never become skipped by default.
+func TestSkipRewardLiveStakeBackfillCheckDefaultsToRunningTheCheck(t *testing.T) {
+	resetGlobalConfig()
+	t.Setenv("HOME", t.TempDir())
```
chrisguiney, line 277: Non-blocking, and CodeRabbit is right here. This test fails if `CARDANO_SKIP_REWARD_LIVE_STAKE_BACKFILL_CHECK` is already set in the developer's environment — confirmed by running it that way:

```
--- FAIL: TestSkipRewardLiveStakeBackfillCheckDefaultsToRunningTheCheck
    flags_test.go:283: the reward_live_stake check must run unless opted out
```

CI is clean so this is not a pipeline risk, but the operators most likely to run the suite locally are the ones who set this variable in the first place. The repo already has the pattern for it: `unsetDebugBindAddrEnv` (`internal/config/config_test.go:77`) does `t.Setenv(name, "")` followed by `os.Unsetenv(name)`, which clears the variable for the test while letting `t.Setenv`'s cleanup restore it. `TestDebugBindAddressDefaultsToLoopback` calls it for exactly this reason. The same two lines here would do it.

The env-binding test above is fine as-is, since it sets the variable itself.

chrisguiney (dismissed):
The motivation holds up: `RewardLiveStakeNeedsBackfill` really does scan the live `utxo` table on every startup to decide whether a repair is needed, and paying that on every restart is a real cost. Default is `false`, a test pins it, the flag logs a warning on every startup where it is active, and it is classified `logPlain` so it also shows in the rendered config. Documentation is in three places. All of that is right.

One blocker: the early return also bypasses `StaleConsensusStakeSnapshotsExist`, which is a fail-closed provenance guard and not the expensive scan. Detail inline, with the reproduction.

### Verified locally (worktree at `aa2eb64`, merge-base `eabf653d`)

| Check | Result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `golangci-lint run ./...` (isolated cache) | 0 issues |
| `go test ./internal/config/` | pass |
| `go test -run 'TestBackfillRewardLiveStake\|TestConfig' .` | pass |
| `make docs-parity` | pass |
| `gofmt -l .` | flags `bark/blob_test.go`, `ledger/read_chain_pipeline_test.go` — both unformatted on `origin/main` too, not this change |

GitHub checks on this head are still queued/in-progress (only DCO, cubic and CodeRabbit have completed), so I have not read a green pipeline. That does not change the disposition here, which rests on the finding below.

### Bot findings

- CodeRabbit `node.go:2159` and cubic `node.go:2154` — both describe the stale-snapshot bypass. **Valid**, reproduced, raised as the blocker.
- CodeRabbit `internal/config/flags_test.go:270` — **valid**, non-blocking; confirmed by running the test with the variable set. Repo already has the helper pattern for this.

### Checked and not raised

- Not adding the field to `gatedFieldPaths` is correct as argued, provided the stale-snapshot guard stays outside the skip. It records no persisted gate state.
- `CARDANO_` prefix and the `TestSkipRewardLiveStakeBackfillCheckEnvBinding` pin: confirmed correct.
- Placement in `logPlainConfigFields` is out of alphabetical order, but that list already has several such deviations (`ActivePeers*`, `CORSAllowedOrigins`, `Midnight.AllowInsecureRemote`) and no test enforces ordering. Not a finding.
- Unlike the neighbouring bool options, this one has no `WithSkipRewardLiveStakeBackfillCheck` `ConfigOptionFunc` and no exported getter. Nothing requires them and the operator-facing path works; noting it only so the omission is a decision rather than an oversight.

node.go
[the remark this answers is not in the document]
chrisguiney: Confirmed and valid. Reproduced at `aa2eb64`: seeding one `pool_stake_snapshot` `mark` row at `calculation_version = 0` (current is 2) and calling `backfillRewardLiveStake` both ways gives the fail-closed error with the flag off and `nil` with it on. `StaleConsensusStakeSnapshotsExist` is two indexed `EXISTS` probes over small tables, unrelated to the full-`utxo` scan the flag targets, so it can run unconditionally at negligible cost. Raised as a merge blocker in my review, with the reproduction and a suggested shape.

node.go
```diff
@@ -2140,7 +2140,24 @@ func (n *Node) nodeSettingsGateValues() nodesettings.Values {
...
+// check itself -- not just a genuine repair -- scans the whole live UTxO
+// table every call, so on a mainnet-scale database it costs as much as a
+// full rebuild on every single startup regardless of whether one is needed.
+// This is for advanced/diagnostic use only: skipping it is safe when the
+// database is already known to be consistent (e.g. repeated restarts during
+// investigation of an unrelated issue), but unsafe to leave enabled
+// permanently since it is what catches a stale or pre-migration
+// reward_live_stake table.
 func (n *Node) backfillRewardLiveStake() error {
+	if n.config.skipRewardLiveStakeBackfillCheck {
```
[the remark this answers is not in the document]
chrisguiney: Confirmed and valid; same finding as CodeRabbit's on this function, and the cost characterisation is accurate. Reproduced at `aa2eb64`: with the flag on, a database carrying a `calculation_version = 0` `mark` snapshot starts silently instead of erroring with the rebootstrap message. Both production call sites are affected (`node.go:807`, `node_lifecycle.go:691`). Raised as a merge blocker in my review.

internal/config/flags_test.go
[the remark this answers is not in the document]
chrisguiney: Valid, non-blocking. Verified by running the test with `CARDANO_SKIP_REWARD_LIVE_STAKE_BACKFILL_CHECK=true` in the environment — it fails at `flags_test.go:283`. The repo already has the pattern: `unsetDebugBindAddrEnv` (`internal/config/config_test.go:77`) pairs `t.Setenv(name, "")` with `os.Unsetenv(name)` so the variable is cleared for the test and restored by `t.Setenv`'s cleanup, and `TestDebugBindAddressDefaultsToLoopback` uses it for this exact reason. Carried into my review as a recommendation.

node.go
wcatz: Both points were right, and the blocker was a real defect — I looked at the first check in the function and did not register that the second one was a different kind of thing. Fixed in bdb8d9f.

**Blocker: the flag no longer reaches `StaleConsensusStakeSnapshotsExist`.** The early return is gone. The skip now applies only to the `RewardLiveStakeNeedsBackfill` probe and the rebuild it gates, by initializing `needed := false` and skipping the probe rather than the whole transaction:

```go
needed := false
if n.config.skipRewardLiveStakeBackfillCheck {
    n.config.logger.Warn(...)
} else {
    needed, err = n.db.Metadata().RewardLiveStakeNeedsBackfill(txn.Metadata())
    ...
}
staleSnapshots, err := n.db.Metadata().StaleConsensusStakeSnapshotsExist(txn.Metadata())
```

The provenance check now runs on every startup whether or not the scan is skipped, and it still fails closed. Your framing is what the docs now say: the cost argument is about a scan whose price grows with the UTxO set, and it simply does not apply to two indexed `EXISTS` probes, so there was never a justification for the flag covering them. The boundary is stated explicitly at all three sites (`node.go` doc comment, `internal/config/config.go` field comment, `ARCHITECTURE.md`, plus `dingo.yaml.example`) so the next reader does not have to infer the scope from the flag name.

**Non-blocking: test isolation.** Fixed. `CARDANO_SKIP_REWARD_LIVE_STAKE_BACKFILL_CHECK` is now cleared explicitly, with any previous value restored through `t.Cleanup`. `t.Setenv` cannot express "unset", so it is saved and restored by hand; the existing `t.Setenv("HOME", ...)` already bars `t.Parallel`, so touching the process environment directly is safe there.

Verified under your reproduction rather than just in a clean shell:

```
$ CARDANO_SKIP_REWARD_LIVE_STAKE_BACKFILL_CHECK=true \
    go test -count=1 -run TestSkipRewardLiveStakeBackfillCheck -v ./internal/config/
--- PASS: TestSkipRewardLiveStakeBackfillCheckEnvBinding (0.00s)
--- PASS: TestSkipRewardLiveStakeBackfillCheckDefaultsToRunningTheCheck (0.00s)
```

Gate re-run: `go build ./...` and `go build -tags dingo_extra_plugins ./...` clean, `go test ./internal/config/ .` pass, `make docs-parity` ok, `make golines` no-op, `golangci-lint run` 0 issues.

node.go
```diff
@@ -2140,13 +2140,39 @@ func (n *Node) nodeSettingsGateValues() nodesettings.Values {
...
+// it, so it runs on every startup whether or not the scan is skipped.
 func (n *Node) backfillRewardLiveStake() error {
 	return n.db.MetadataTxn(true).Do(func(txn *database.Txn) error {
-		needed, err := n.db.Metadata().RewardLiveStakeNeedsBackfill(
-			txn.Metadata(),
-		)
-		if err != nil {
-			return fmt.Errorf("check reward live stake backfill: %w", err)
+		needed := false
+		if n.config.skipRewardLiveStakeBackfillCheck {
```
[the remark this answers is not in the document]
wcatz: I checked this against both call sites and I don't think the premise holds, so
I'm not making the change — but tell me if I've missed a path.

`backfillRewardLiveStake` never runs concurrently with block processing. Both
callers invoke it strictly before `n.ledgerState.Start`, which is what brings up
the slot clock and block processing:

- `node.go:807` → `n.ledgerState.Start` at `node.go:973`
- `node_lifecycle.go:691` → `n.ledgerState.Start` at `node_lifecycle.go:826`

That ordering is deliberate and load-bearing, not incidental. The doc comment
states it (`node.go:2140-2142`: "runs after commit-timestamp recovery and before
ledger processing can advance the chain, so the next epoch-boundary snapshot
cannot observe a partially populated aggregate"), and the comment at
`node.go:811-813` gives the same reason for the Midnight indexer being created
in that window — "the synchronous backfill runs while no new blocks can arrive".
So there is no block processing in flight to contend with on SQLite, whether the
flag is set or not.

On the second half — that the write transaction "undermines the startup-cost
purpose of the flag" — the cost the flag exists to avoid is the
`RewardLiveStakeNeedsBackfill` scan, which walks the whole live UTxO table on
every startup and on a mainnet database costs as much as a full rebuild.
Acquiring a write transaction is O(1) against that, and it is the same
transaction the unskipped path has always taken. Skipping the scan is what makes
the flag worth having; the transaction mode is not measurable next to it.

There is also a concrete cost to splitting it. Whether a write is needed is only
known *after* the probes, so a read-only-then-reopen-if-needed shape introduces
a window between the probe and the repair — precisely the "partially populated
aggregate" the single-transaction structure exists to rule out. Trading that
invariant for no measurable gain on a path where nothing else is running isn't a
good trade.

Happy to reconsider if there's a caller I haven't found that reaches this after
`ledgerState.Start`.

wolf31o2, line 2163: changes requested: when skipRewardLiveStakeBackfillCheck is enabled, this branch still runs inside MetadataTxn(true), so the opt-out uses the metadata writer connection for read-only provenance probes. That can contend with block processing on SQLite and defeats the startup-cost purpose of the flag. Use a read-only transaction for the skipped path and retain a write transaction only when the reward aggregate may be rebuilt.

wolf31o2 (changes requested):
changes requested

node.go
wcatz: Fixed in c5b662bf.

You were right about the mechanism, and it is worse than just holding the writer open: `StaleConsensusStakeSnapshotsExist` calls `readDBFromTxn`, which for a non-nil txn delegates to `dbFromTxn` and runs the query on the transaction's own connection. Inside `MetadataTxn(true)` that is the write pool, so both read-only probes were executing on the single writer connection.

Both probes now run in `MetadataTxn(false)` (`ReadTransaction` -> `readDB`), and a write transaction is opened only when a probe has already determined a rebuild is required. Splitting them is safe because both call sites — `node.go:807` and `node_lifecycle.go:691` — run before `ledgerState.Start`, so nothing can write `reward_live_stake` in the gap.

Also added the node-level coverage this path never had: one test that the flag leaves a database needing a backfill untouched, one that the provenance probe still fails closed with the flag set. Both mutation-verified — neutering the flag fails the first, letting the flag suppress the provenance probe fails the second. `ARCHITECTURE.md` records the split.

wolf31o2 (approved):

PR #4050, opened 2026-09-07
fix(ledger): reject unevaluable post-Mithril mark eligibility

chrisguiney:
## The fail-open

`shouldSkipPostMithrilMarkEligibility` reports a mark row that was
reconstructed from live state after its target epoch boundary. On that branch
`leaderEligibilityStake` returned `skipEligibility=true` unconditionally, so
`verifyBlockLeaderEligibility` accepted the block with no stake-threshold check
at all — on every profile, not only the prototype one.

It survived #3543, which moved the sibling "eligibility cannot be evaluated"
states behind `SkipLeaderStakeThresholdCheck`: that change touched the
zero-total-active-stake and missing-active-slot-coefficient guards and left this
branch alone (`git show e0215925 -- ledger/verify_header.go` has no
`shouldSkipPostMithrilMarkEligibility` hunk).

## The fix

Gate the bypass behind the same flag. A standard profile now returns
`errLeaderStakeSnapshotUnavailable`, which `verifyBlockHeaderState` classifies
as deferrable while the ledger apply cursor is behind the block slot and as a
hard rejection once the cursor has caught up. Only the explicitly selected
prototype profile keeps the warn-and-trust bypass.

This is one case of #1649, not the whole issue.

## Blast radius

- Normal operation is unaffected. `mark[K]` is captured at `BoundarySlot-1`,
  strictly below `startSlot(K)`, so the guard does not fire for a rotated row.
- Certified Mithril imports are unaffected. `importedSnapshotCaptureSlot`
  stamps `epochStart-1`, and the legacy importer's anchor stamp is already
  exempt through the `CapturedSlot == mithrilLedgerSlot` check.
- Affected: databases where `CaptureGenesisSnapshot` reconstructed the
  post-Mithril window, which stamps the mark rows for N, N-1 and N-2 with
  `startSlot(N)`. Blocks in epoch N+1 elect from the reconstructed `mark[N]`.
  Those were accepted unverified before; they now defer during catch-up and
  reject once the apply cursor passes the slot. That is the intended
  fail-closed conversion — a node that cannot evaluate eligibility stops rather
  than following an unverified producer — but it is a behaviour change for that
  database class, and the remedy there is a re-import, since
  `SkipLeaderStakeThresholdCheck` is not reachable from a preview/preprod/
  mainnet configuration.

## Tests

`TestVerifyBlockLeaderEligibility_ReconstructedHistoricalMarkSkips` asserted
`require.NoError` for exactly this scenario on a default `LedgerState`, so the
fail-open was encoded as the expected contract. It is replaced by
`..._RejectsStandardProfile`, which asserts
`ErrorIs(errLeaderStakeSnapshotUnavailable)`, plus
`..._PrototypeAccepts`, which keeps the old assertion for the prototype
profile. That matches the Rejects/PrototypeAccepts pairing the other
eligibility guards use in this file.

`TestVerifyBlockHeaderState_ReconstructedMarkDefersThenRejects` covers the
defer wiring on the path blockfetch reaches through
`verifyBlockHeaderStateWithEpochAdvance`: tip behind defers, the apply-time
recheck with deferral disabled rejects, and a caught-up cursor rejects rather
than deferring forever.

Fail-before evidence: with `ledger/verify_header.go` reverted to
`origin/main` and the tests kept, `..._RejectsStandardProfile` and both
subtests of `..._DefersThenRejects` fail with "An error is expected but got
nil". `..._PrototypeAccepts` is a guard and passes either way.

## Checks

`go build ./...`, `go vet ./...`, `gofmt -l` on the changed files,
`golangci-lint run ./...` (root, `GOOS=windows`, and both nested modules),
`go test ./internal/architecture`, and `go test -race ./ledger/...` all pass.
`gofmt` reports `bark/blob_test.go` and `ledger/read_chain_pipeline_test.go`
as unformatted at `origin/main` as well; neither is touched here.

Not run, because the infrastructure is unavailable in this environment: a real
Mithril snapshot restore, live Cardano network validation, the devnet suite,
and Antithesis. `nilaway`, `modernize` and `golines` are not installed locally.

## Related

No open pull request touches `ledger/verify_header.go`. #4030 and #4031 touch
`ledgerstate/import.go` and #4032, #3967 and #3938 touch `ledger/state.go`; the
only overlap with this branch is a doc comment on
`LedgerStateConfig.SkipLeaderStakeThresholdCheck`.

https://claude.ai/code/session_015vJhjXgwH1NU7c2tqBTcpY

---
## Summary by cubic
Closes a fail-open where a post-Mithril mark snapshot reconstructed after its target boundary skipped the leader stake-threshold check on every profile, accepting blocks with no eligibility evaluation. A standard profile now treats that state as unevaluable and returns `errLeaderStakeSnapshotUnavailable`, which defers while the ledger apply cursor is behind the block slot and rejects once caught up; only the explicitly selected prototype profile (`SkipLeaderStakeThresholdCheck`) keeps the warn-and-trust bypass.

**Behavior change for existing databases**

- Databases where `CaptureGenesisSnapshot` reconstructed the post-Mithril window (mark rows stamped with `startSlot(N)`) previously accepted such blocks unverified; they now defer during catch-up and reject once the apply cursor passes the slot.
- The remedy is a re-import, since `SkipLeaderStakeThresholdCheck` is not reachable from preview/preprod/mainnet configurations.
- Normal operation is unaffected: rotated `mark[K]` rows are captured below `startSlot(K)`, and certified Mithril imports stamp `epochStart-1`, so the guard does not fire for them.

**Tests**

- Replaces the old test that asserted `NoError` for the reconstructed historical mark on a default `LedgerState` with a standard-profile rejection test and a prototype-profile acceptance test.
- Adds a test pinning the defer/reject wiring: a tip behind the slot defers, the apply-time recheck with deferral disabled rejects, and a caught-up cursor rejects rather than deferring forever.

<sup>Written for commit c55a2ac8a1ad5bdc5b3381ab2291f08d3ca611fe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4050?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


chrisguiney:
Bot review standing on `c55a2ac`:

- cubic reviewed this head — its review record is stamped `c55a2ac8a1ad5bdc5b3381ab2291f08d3ca611fe` and reports no issues across the 4 changed files.
- CodeRabbit did **not** review this head. It posted "Review limit reached, next included review available in 53 minutes", so its comment is a quota notice rather than a review.

A completed cubic review satisfies the bot stage under the repository policy; human review is still required.

wolf31o2 (approved):
Accept

Issue #4015, opened 2026-09-06
database: every ledger rollback full-scans the whole `utxo` table — `SELECT DISTINCT credential_tag, staking_key ... WHERE added_slot > ?` defeats `idx_utxo_added_slot`, freezing block application for minutes


chadaibbhot-git:
## Summary

A one-block ledger rollback (`fork_depth=1`) takes **1m34s to 5m52s** on a
block producer with a 15 GB metadata database. The ledger holds its async DB
transaction for the whole sweep, so block application stops for that entire
window. Two rollbacks landing back to back at an epoch boundary froze one
producer's chain for **9m32s** and put it **591 slots** behind the network.

The cost is not proportional to rollback depth. It is proportional to the size
of the `utxo` table, because two statements inside `TruncateAfterSlot` ask
SQLite for `SELECT DISTINCT credential_tag, staking_key`, and the `DISTINCT`
makes the planner abandon the `added_slot` / `deleted_slot` range index in
favour of a **full scan of `idx_utxo_staking_deleted_amount`** — 20 million
index entries, each followed by a row lookup, to return **2 rows**.

Measured on a live producer's own database (read-only connection):

```
sqlite> EXPLAIN QUERY PLAN
        SELECT DISTINCT credential_tag, staking_key FROM utxo WHERE added_slot > 2656808;
SCAN utxo USING INDEX idx_utxo_staking_deleted_amount          -- 27.06 s, 2 rows

sqlite> EXPLAIN QUERY PLAN
        SELECT credential_tag, staking_key FROM utxo WHERE added_slot > 2656808;
SEARCH utxo USING INDEX idx_utxo_added_slot (added_slot>?)     --  0.003 s, 5248 rows -> 2 distinct
```

**27.06 s vs 0.003 s** for the same answer, warm page cache. The `deleted_slot`
variant is covering, so it is "only" a 1.07 s full index scan versus 0.020 s.

## Environment

* dingo `main` (verified against `64676163`). `database/plugin/metadata/sqlstore/utxo.go`
  and `live_stake.go` are byte-identical on `main` and on the image under test;
  `database/truncate.go` differs only by the duration instrumentation from open
  PR #3967 that produced the timings below.
* Network: `musashi` Leios devnet, `--run-mode leios`, 1 s slots.
* Two block producers, SQLite metadata store, `metadata.sqlite` = 15.6 GB,
  `utxo` = **20,007,835 rows** of which **117,639 are live** (`deleted_slot = 0`).
[...]

Issue #3253, opened 2026-08-21
backfill: throughput decays to a standstill (3311 → 9 blocks/s, eta 30m → 177h); set_transaction_batched saturates the interval


wolf31o2:
## Summary

On `v0.69.0`, the Mithril bootstrap `backfill` phase decays to a standstill.
Over **4h15m of wall time it advanced 0.6% → 2.2%**, with `blocks_per_sec`
falling **3311 → 9**. The phase's own `eta` field climbs monotonically the
whole time — **30m → 177h** — and a rate measured over a later 60s window
projects roughly **34 days** to finish.

The practical effect is that a from-snapshot bootstrap never reaches a serving
node. I hit this driving the embedded node inside the Blink full-node wallet,
where it means a fresh install cannot reach `ready` at all.

## Measurements

Sampled from `msg="backfill progress"`, every ~150th line:

| time | blocks_per_sec | progress | set_transaction_batched_ms | node's own eta |
|---|---|---|---|---|
| 08:11:27 | 3311 | 0.6% | 5492 | 30m0s |
| 08:36:40 | 50 | 1.4% | 9938 | 30h52m47s |
| 09:02:05 | 28 | 1.5% | 9958 | 54h42m37s |
| 09:27:35 | 21 | 1.7% | 9996 | 73h26m20s |
| 09:53:20 | 17 | 1.8% | 10028 | 93h27m59s |
| 10:19:05 | 14 | 1.9% | 10297 | 112h32m6s |
| 10:44:52 | 12 | 1.9% | 10350 | 129h45m41s |
| 11:10:39 | 11 | 2.0% | 10823 | 144h20m12s |
| 11:36:46 | 10 | 2.1% | 10605 | 160h48m27s |
| 12:03:10 | 9 | 2.1% | 10972 | 177h39m12s |

## Where the time goes

`set_transaction_batched_ms` pegs at **~10,000ms of every ~10,000ms reporting
interval** — effectively 100% of wall time inside
`db.SetTransactionBatchedWithOpts` (`internal/node/backfill.go`). Everything
else is noise by comparison. A representative late interval:

```
blocks_per_sec=8 progress=2.2% interval_blocks=20 interval_txs=80
  block_read_decode_ms=11  offset_compute_ms=4  blob_offset_write_ms=0
  set_transaction_batched_ms=10661  consumed_input_recovery_ms=35
  flush_batch_ms=0  checkpoint_write_ms=0  eta=190h58m13s
```

[...]

Issue #2460, opened 2026-05-30
database/sqlite: quantify and reduce deferred index rebuild cost after API backfill


wolf31o2:
## Context

The May 30, 2026 current-main API backfill rerun measured a successful Mithril preview bootstrap only after applying the local workaround for #2457 (do not defer/drop `asset.policy_id` on SQLite).

Commit measured: `d73407b6` (`origin/main` at test start)

The API metadata backfill phase completed in:

- Blocks processed: 4,312,604
- Transactions stored: 6,567,610
- Backfill elapsed: 49,516s = 13h45m16s
- Average backfill throughput: 87.1 blocks/sec

After that, rebuilding deferred metadata indexes took another:

- Deferred indexes rebuilt: 26
- Rebuild elapsed: 1,948s = 32m28s
- Whole pod command wall time from start marker to exit: about 14h32m33s

## Dominant rebuild costs

The expensive rebuilds were concentrated in UTxO and asset indexes:

| Group | Indexes | Approx duration |
|---|---:|---:|
| UTxO deferred indexes | 8 | 1,234s / 20m34s |
| Asset deferred indexes | 3 | 560s / 9m20s |
| Other deferred indexes | 15 | 154s / 2m34s |
| Total | 26 | 1,948s / 32m28s |

Largest individual indexes:

- `asset.Fingerprint`: 236s
- `asset.NameHex`: 179s
- `utxo.PaymentKey`: 176s
- `utxo.SpentAtTxId`: 170s
- `utxo.idx_utxo_deleted_staking_amount`: 163s
- `utxo.StakingKey`: 159s
- `utxo.TransactionID`: 146s
- `asset.Amount`: 144s

## Why this matters

The main under-6-hour gap is still the API metadata backfill phase itself (~2.3x short), but a fixed 30+ minute post-backfill index rebuild will become proportionally more visible as the SQL hot path improves. It should be tracked separately from #2353, #2458, and #2459.

## Areas to investigate

- Measure which deferred indexes are required before normal API service can start and which can be rebuilt lazily or in a separate maintenance mode.
- Check whether any UTxO or asset query indexes overlap enough to merge, narrow, or convert to partial indexes.
- Benchmark SQLite rebuild settings for this phase (`temp_store`, cache sizing, `mmap_size`, `PRAGMA threads`, journal/sync choices already used during bulk load).
[...]

Issue #2457, opened 2026-05-30
database/sqlite: deferred asset index drop breaks Mithril ledger import


wolf31o2:
## Summary

Fresh API-mode Mithril sync on current `main` (`d73407b6`) fails during ledger-state UTxO import after deferred metadata indexes are dropped.

The failing statement is the asset import path:

```sql
INSERT INTO asset (...) VALUES (...)
ON CONFLICT (utxo_id, policy_id, name) DO NOTHING
```

SQLite returns:

```text
SQL logic error: ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint (1)
```

## Reproduction

Local Kubernetes `dingo-preview`, fresh 90Gi PVC, Cardano preview cached Mithril snapshot:

- snapshot digest `d98b3311fb52cb5aa773c5ad1c98e7d571c6e9afa6cb71e5a11136963f8d02e6`
- epoch `1307`
- immutable file number `26155`
- storage mode `api`
- metadata `sqlite`
- blob `badger`
- 4 CPU / 12 GiB pod limit

Run:

```sh
dingo mithril sync
```

Relevant log sequence:

```text
dropped deferred metadata indexes for bulk load dropped=27 manifest_size=27
importing UTxOs from ledger state
SQL logic error: ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
Error: importing ledger state: importing UTxOs: ... import utxo assets ...
```

## Suspected Cause

`database/plugin/metadata/deferred/deferred.go` says `asset.idx_asset_unique` must stay because it backs the import `ON CONFLICT` target. However the manifest includes `models.Asset{}` field `PolicyId`.

`models.Asset.PolicyId` has both:

```go
gorm:"index;uniqueIndex:idx_asset_unique;size:28"
```

On SQLite, dropping the `PolicyId` field index appears to drop the composite unique index needed by:

```go
clause.OnConflict{
    Columns: []clause.Column{
        {Name: "utxo_id"},
        {Name: "policy_id"},
        {Name: "name"},
    },
    DoNothing: true,
}
```

Removing the `PolicyId` entry from the deferred-index manifest let the same local run proceed past the failure, dropping 26 indexes instead of 27.

## Acceptance Criteria

- Fresh API-mode Mithril sync completes ledger-state UTxO import with deferred indexes enabled.
- Regression test proves every deferred-index manifest entry can be dropped without breaking import `ON CONFLICT` targets.
[...]

PR #4059, opened 2026-09-07
fix(database): rebuild the rollback cascade index on a critical rebuild

chadaibbhot-git:
## What

`TruncateAfterSlot` ends in `DELETE FROM "transaction" WHERE slot > ?`
(`database/plugin/metadata/sqlstore/transaction_write.go`). `utxo.transaction_id`
is the child column of `fk_transaction_outputs` (`ON DELETE CASCADE`, migration
v1 `expand.sql`), and `shared_sqlstore.go` opens every connection with
`foreign_keys(1)`, so SQLite enforces that cascade with an implicit
`DELETE FROM utxo WHERE transaction_id = ?` per deleted parent row.

`idx_utxo_transaction_id` was the only `utxo` entry in the deferred-index
manifest without `Critical: true`. The rebuild that runs before a
Mithril-bootstrapped database is marked ready (`mithril/sync.go` →
`DeferredIndexRebuilder.BuildCritical`) therefore leaves it out, and a rollback
in that state scans the whole `utxo` table once per rolled-back transaction —
`O(rolled-back transactions x utxo rows)`.

Measured on a read-only copy of a preview relay's own database (3,167,422 `utxo`
rows, 1,001 `transaction` rows in the rolled-back window, warm page cache, C
SQLite; the shipped pure-Go driver is slower):

| statement | index absent | index present |
| --- | --- | --- |
| `DELETE FROM "transaction" WHERE slot > 122071241` (1,001 rows) | 556.646 s | 0.095 s |
| `CREATE INDEX idx_utxo_transaction_id ON utxo(transaction_id)` | — | 1.729 s (one-off) |

The same node's instrumented sweeps for comparable rollbacks ran 10m44s, 10m45s,
11m07s and 31m52s, with no log output at all between "rolling back" and "sweep
complete". The classification survived review because the parent statement's
plan does not show the cascade:

```
EXPLAIN QUERY PLAN DELETE FROM "transaction" WHERE slot > 122071241;
SEARCH transaction USING INDEX idx_transaction_slot (slot>?)   -- looks fine

EXPLAIN QUERY PLAN DELETE FROM utxo WHERE transaction_id = 1;
SCAN utxo                                                      -- the real cost
```

## Why the deferred manifest is never finished on a Mithril-bootstrapped node

Corrected in review (thanks @chrisguiney): the earlier description blamed the
missing indexes on an older binary whose critical subset was smaller. That is
wrong, and the real cause is current `main` on every sync.

`mithril/sync.go` calls `deferredIndexes.BuildCritical()`, which deliberately
leaves `deferred.SyncStateKey` (`metadata_indexes_pending`) set so the lazy
remainder is finished by later maintenance. It then calls
`updateMithrilReadyState(..., clearSyncState=true)` → `db.ClearSyncState(txn)`,
which is `DELETE FROM sync_state` with no `WHERE`
(`queries/sqlite/operational.sql`). The marker is deleted moments after it was
written. The clear already re-writes `mithril_ledger_slot` and
`mithril_ledger_hash` after it for exactly this reason; the deferred-index
marker was not on that list.

Consequences on every Mithril-bootstrapped database, API mode included: all
twelve lazy manifest entries (`idx_asset_name_hex`, `idx_asset_fingerprint`,
`idx_asset_amount`, `idx_datum_added_slot`, `idx_certs_block_hash`,
`idx_certs_certificate_id`, `idx_certs_cert_type`, `idx_redeemer_index`,
`idx_redeemer_tag`, `idx_key_witness_type`, `idx_witness_scripts_script_hash`,
`idx_witness_scripts_type`) stay absent for the life of the database.
`RepairDeferredIndexes` takes its no-cycle-pending branch and
`Node.startDeferredIndexMaintenance` logs "deferred-index maintenance not
needed", both reading the same wiped row.

## Fix

1. `deferred.Manifest`: `idx_utxo_transaction_id` is now `Critical: true`. The
   `Critical` criteria comment gains the rule that produced the gap: the child
   column of any `ON DELETE CASCADE` foreign key whose parent rows the rollback
   path deletes belongs in the critical subset, because the engine's implicit
   per-parent-row child delete is a scan without it and is invisible from the
   parent statement's plan. The pinned critical count moves 12 → 13.
2. `updateMithrilReadyState` (`mithril/sync_import.go`) carries
   `deferred.SyncStateKey` across the clear, the same way `mithril_ledger_slot`
   and `mithril_ledger_hash` are re-written after it. `mithril/sync.go` still
   calls `BuildCritical()` only — building the API-only remainder inside the
   sync would spend that time on core-mode nodes that never query those indexes,
   which is the cost the deferral exists to avoid — but the marker it leaves now
   survives, so the first `serve` runs the existing `BuildDeferredIndexes`
   maintenance and finishes the manifest. That is the entry point the
   `RepairDeferredIndexes`/`startDeferredIndexMaintenance` paths were written
   for; nothing new is introduced to carry it.
3. `RepairDeferredIndexes` and `RepairCriticalDeferredIndexes` restore a missing
   critical index even when no drop/rebuild cycle is pending. This is what
   repairs databases already on disk: their marker was wiped by the sync that
   built them, the schema migration that created the index is recorded complete
   so its `CREATE INDEX IF NOT EXISTS` never runs again, and the manifest is
   only consulted by a cycle that is no longer pending. Without it the gap is
   permanent — and that is the state two Mithril-bootstrapped preview nodes were
   found in (every `Critical` index present, every lazy one absent, no
   `metadata_indexes_pending` row). `BuildCriticalDeferredIndexes` skips indexes
   that are present, so on a healthy database this is one catalog lookup per
   critical entry inside the write transaction the repair already opens.
4. Startup logging: the missing critical entries are now named *before* the
   rebuild starts, through a new optional
   `metadata.MissingCriticalDeferredIndexLister` the SQL store implements
   (catalog reads, no DDL). A rebuild on a multi-million-row table is otherwise
   silent for as long as it runs — the symptom this PR opens with. The
   completion line reports `duration_covers="critical rebuild and retained-index
   restore"`, because `BuildCriticalDeferredIndexes` goes through
   `withDeferredIndexWrite`, which runs `ensureRetainedIndexes` in the same
   transaction. That reachability is a deliberate gain: `deferred.Retained`
   self-healing previously never ran on a start with no cycle pending, and now
   runs on every core-mode start. A healthy database logs nothing.

## Tests

Which tests fail on `main`, and which are guards that pass either way:

Fail on `main`, pass with the change:

`database/plugin/metadata/sqlite/shared_sqlstore_fk_cascade_index_test.go`

- `TestCascadeChildColumnsIndexedAfterCriticalRebuild` — reads every
  `ON DELETE CASCADE` foreign key out of the live schema via
  `pragma_foreign_key_list` and asserts each child column still has a covering
  index after `DropDeferredIndexes` + `BuildCriticalDeferredIndexes` (the
  Mithril sequence). Stated over the schema rather than over the one index, so
  a future cascading foreign key whose child index is classified lazy fails
  here instead of in production.
- `TestUtxoCascadeDeleteIndexedAfterCriticalRebuild` — pins the plan of the
  cascade's own delete, asserting the `SCAN utxo` first with the manifest
  dropped so the positive assertion has teeth.

`mithril/sync_import_deferred_marker_test.go`

- `TestUpdateMithrilReadyStateKeepsDeferredIndexPendingMarker` — after
  `DropDeferredIndexes` + `BuildCriticalDeferredIndexes`,
  `HasDeferredIndexesPending()` is still true once
  `updateMithrilReadyState(..., clearSyncState=true)` has run, and
  `sync_status` is still cleared.
- `TestMithrilSyncLeavesLazyManifestForTheFirstServe` — the composition a
  bootstrapped node takes: sync's drop + `BuildCritical` +
  `updateMithrilReadyState`, then `node.RepairDeferredIndexes` as the first
  `serve`, after which a lazy manifest entry exists and the marker is cleared.

`internal/node/deferred_index_repair_test.go`,
`cmd/dingo/serve_deferred_index_repair_test.go`

- `TestRepairDeferredIndexesRestoresCriticalIndexWithoutMarker`,
  `TestRepairCriticalDeferredIndexesRestoresCriticalIndexWithoutMarker` —
  the cleared-marker state, repaired, with no marker invented.
- `TestServeCoreModeRepairsMissingCascadeIndex` — the same at the entry point
  `serveRun` actually takes: a core-mode configuration reaches
  `repairDeferredIndexes`, and the index is back afterwards.
- `TestEnsureCriticalDeferredIndexesNamesMissingBeforeBuilding` — the names are
  logged before the rebuild is entered (asserted from inside the rebuild), and
  the completion line says what its duration covers.
- `TestRepairDeferredIndexesNamesMissingIndexAgainstRealStore` — the same
  through the SQLite store and its real catalog rather than a fake.
- `TestEnsureCriticalDeferredIndexesQuietWhenManifestComplete` — a complete
  manifest stays silent.

Guards, which pass on `main` and with either production hunk reverted; they
hold the surrounding facts the change depends on rather than the change itself:

- `TestRollbackDeleteCascadesIntoUtxo` — the rollback's parent `DELETE` really
  does cascade into `utxo`, with `foreign_keys(1)` enforced.
- `TestRepairDeferredIndexesFinishesPendingCycle` — the marker keeps its
  existing meaning: a genuinely interrupted cycle still rebuilds the whole
  manifest and clears it.

```
go build ./... && go vet ./...
go test ./database/plugin/metadata/... ./internal/node/ ./cmd/dingo/ ./database/ ./mithril/... -count=1
go test -race ./database/plugin/metadata/deferred/ ./database/plugin/metadata/sqlite/ ./internal/node/ ./cmd/dingo/ ./mithril/ -count=1
golangci-lint run ./internal/node/... ./database/plugin/metadata/... ./mithril/... ./cmd/dingo/...
```

All green on the rebased head, lint 0 issues. Each production hunk was also
reverted on its own and the corresponding tests re-run: the manifest
classification, the marker carry in `updateMithrilReadyState`, the no-marker
repair, the pre-build log line, and the store's missing-index lister each fail
their own tests when reverted alone.

Operationally, a database already in this state can also be repaired online with
`CREATE INDEX IF NOT EXISTS idx_utxo_transaction_id ON utxo(transaction_id);`
(1.7 s on 3.2M rows); this change does it for the operator on the next start.

No separate issue: the analysis is inline above, and the field evidence it
came from is below.

## How this was found

A relay bootstrapped with `dingo mithril sync` (storage mode `core`, preview,
`metadata.sqlite` 1.6 GB, `utxo` 3,167,422 rows, `transaction` 2,472 rows)
produced no log output for 24 minutes at startup after an in-place image swap.
These are consecutive lines; nothing was elided between the last two:

```
INFO  database worker pool initialized workers=5 queue_size=50
WARN  ledger tip ahead of primary chain tip at startup, rolling back
      metadata to chain tip
      chain_tip_slot=122071241 ledger_tip_slot=122084222
<no further output for 24 minutes>
```

One core pegged at ~105%, RSS flat at 170 MiB, block I/O static, and the
write-ahead log held a single 4,096-byte frame — a read/scan phase, not write
progress, on a database already warm in the page cache. The container could not
be stopped gracefully (`docker stop -t 120` returned 137): the sweep runs on a
database worker-pool goroutine inside the ledger's async transaction, which
shutdown waits to drain and cannot cancel.

The rollback window held 1,001 `transaction` rows. Reproduced against a
read-only copy of that database (the timings in the table above), the single
`DELETE FROM "transaction" WHERE slot > ?` took 556.6 s, and a one-off
`CREATE INDEX idx_utxo_transaction_id ON utxo(transaction_id)` (1.7 s) brought
the same statement to 0.095 s. Per-parent-row cost scales linearly: 1 row
0.65 s, 9 rows 4.9 s (~0.55 s each).

The same node's own instrumented sweeps, previous container generation, show
both regimes:

```
rollback_slot=122087673  duration=21.665s
rollback_slot=122093634  duration=23.909s
...
rollback_slot=122097167  duration=11m07.882s
rollback_slot=122097057  duration=10m45.275s
rollback_slot=122097167  duration=10m44.876s
rollback_slot=122084222  duration=31m52.587s
```

The ~22-26 s baseline is #4015 (fixed by #4022 plus an ordering fix); the
10-32 minute outliers are the deep rollbacks, where the transaction count in
the window is large. Note the pattern of a multi-minute sweep followed
immediately by a ~24 s sweep at the same slot: the deep one is the cascade, the
retry finds nothing left to cascade.

A/B across four nodes, read directly from each `metadata.sqlite`:

| node | bootstrap | mode | `utxo` rows | `idx_utxo_transaction_id` | deep-rollback sweep |
| --- | --- | --- | --- | --- | --- |
| preview relay (subject) | `dingo mithril sync` | core | 3.2M | absent | 24 min, killed |
| preview block producer | `dingo mithril sync` | core | 3.2M | absent | — |
| devnet node A | full sync | core | 32.1M | present | ~12 s |
| devnet node B | full sync | core | 32.1M | present | ~12 s |

The two nodes with a ten times larger `utxo` table are the fast ones: neither
rollback depth nor table size is the discriminator, the index is. On both
Mithril-bootstrapped databases every `Critical` manifest index was present,
every lazy one absent, and `sync_state` held no `metadata_indexes_pending` row
— exactly what the unqualified `DELETE FROM sync_state` above produces, fixed
going forward by item 2 and repaired on existing databases by item 3.

## Related

- #4015 — sibling of this bug: the same rollback sweep, the
  `SELECT DISTINCT credential_tag, staking_key ... WHERE added_slot > ?` full
  scans. Different mechanism; this FK cascade is what remains once that one is
  fixed.
- #4022 — the fix for #4015 (open).
- #3253 / #3298 — the same hazard class on the witness `transaction_id`
  indexes: an index the bulk-load manifest deferred turned a per-row delete
  into a full scan. `deferred.Retained` and its self-healing came from that,
  and this change follows the precedent.
- #2450 — introduced the deferred-index manifest, the critical subset, and the
  core-mode serve repair this change extends.
- #2457, #2460 — earlier deferred-index classification and rebuild-cost fixes.
- #3967 — the `TruncateAfterSlot` duration instrumentation that made the
  31m52s measurement available.

## CI status

Rebased onto `main` at `aa088149` (2026-09-09); all 4 commits carried through
byte-identical (`git range-diff`, no conflicts). Re-verified the touched
packages on the rebased head — `go build ./...`, `go vet`, and `go test` clean
across `cmd/dingo`, `database/...` (including `sqlstore/migrations`),
`internal/node`, `internal/test/dbtest`, and `mithril`, none of which this PR's
diff misses.

Two red checks on this head, both unrelated to this branch's diff:

- `go-test (Linux) (1.26.x)`: failed in the `install-mariadb-client` setup
  step (`apt-get` exit 100, hash-sum mismatch fetching an unrelated
  `dl.google.com/linux/chrome-stable` package index) — never reached `go test`.
  Runner/mirror infra, not this branch.
- `go-test (macOS) (1.26.x)`: failed on `ledger.TestScheduler_ChangeInterval`,
  a fixed-timing test in `ledger/timer_test.go` this PR does not touch. Ran
  `go test ./ledger -run TestScheduler_ChangeInterval -count=10` locally:
  10/10 green. Timing-sensitive assertion (expects 1-3 ticks in a 500ms window
  after an interval change) flaking under CI runner load, not a regression
  from this diff.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5827,7 +5827,11 @@ nonce is passed through unchanged.
 In API storage mode, the shared SQL metadata providers can defer selected query
 indexes during bulk load. Deferred indexes are classified as critical or lazy in
 `database/plugin/metadata/deferred`: critical indexes cover startup API queries
-and rollback predicates, while lazy indexes cover secondary query paths. Only
+and rollback predicates -- including the child column of an `ON DELETE CASCADE`
+foreign key whose parent rows the rollback deletes, since the engine enforces
+such a cascade with an implicit per-parent-row child delete that the parent
+statement's query plan does not show -- while lazy indexes cover secondary
+query paths. Only
 indexes no import path filters on are eligible at all — an index a per-row
 import predicate needs stays resident, since dropping it turns that predicate
 into a full scan of a table the import is still growing. Those indexes are
@@ -5844,7 +5848,11 @@ critical subset before clearing `sync_status`, then leaves the pending
 sync-state marker set. API-mode `serve` verifies the critical subset before
 startup and runs the full lazy rebuild as background maintenance; the marker is
 cleared only after the full manifest has been rebuilt. Core-mode startup still
-repairs the full manifest synchronously before serving. On MySQL, InnoDB
+repairs the full manifest synchronously before serving. Both repair entry
+points also restore any missing critical index when no cycle is pending at all:
+the marker records that a cycle was interrupted, not which indexes exist, so a
+database whose marker an older binary cleared while its own critical subset was
+smaller would otherwise carry the gap permanently. On MySQL, InnoDB
 requires indexes supporting foreign-key child columns, so the dialect leaves
 those indexes in place while deferring the remaining manifest entries.
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5846,13 +5846,24 @@ manifest. The metadata plugin exposes
 `BuildDeferredIndexes` for the full manifest. Mithril sync rebuilds the
 critical subset before clearing `sync_status`, then leaves the pending
 sync-state marker set. API-mode `serve` verifies the critical subset before
-startup and runs the full lazy rebuild as background maintenance; the marker is
-cleared only after the full manifest has been rebuilt. Core-mode startup still
+startup and runs the full lazy rebuild as background maintenance; the rebuild
+paths clear the marker only after the full manifest has been rebuilt, but they
+are not the only writer of that row. `ClearSyncState`
+(`DELETE FROM sync_state`, no `WHERE`) removes it too, and Mithril sync runs
+that clear through `updateMithrilReadyState` immediately after the critical
+rebuild. It therefore re-writes every row a completed sync still needs —
+`mithril_ledger_slot`, `mithril_ledger_hash`, and the deferred-index marker —
+back after the clear; without the last of those, every Mithril-bootstrapped
+database loses the marker moments after `BuildCritical` set it and never builds
+the lazy manifest entries at all. Core-mode startup still
 repairs the full manifest synchronously before serving. Both repair entry
 points also restore any missing critical index when no cycle is pending at all:
-the marker records that a cycle was interrupted, not which indexes exist, so a
-database whose marker an older binary cleared while its own critical subset was
-smaller would otherwise carry the gap permanently. On MySQL, InnoDB
+the marker records that a cycle was interrupted, not which indexes exist, and a
+database bootstrapped by a binary that predates the marker being carried across
+the clear has the critical subset built, the lazy remainder dropped, and no
+record of either — it would otherwise carry that gap permanently, since the
+schema migration that created those indexes is recorded complete and never
+re-runs. On MySQL, InnoDB
 requires indexes supporting foreign-key child columns, so the dialect leaves
 those indexes in place while deferring the remaining manifest entries.
 
--- a/internal/node/deferred_index_repair_test.go
+++ b/internal/node/deferred_index_repair_test.go
@@ -15,6 +15,7 @@
 package node
 
 import (
+	"bytes"
 	"database/sql"
 	"io"
 	"log/slog"
@@ -34,11 +35,14 @@ const rollbackCascadeIndex = "idx_utxo_transaction_id"
 // two Mithril-bootstrapped preview nodes were found in: a critical manifest
 // index absent, and no pending marker to say so.
 //
-// A binary whose critical subset did not yet include the index cleared the
-// marker when it finished its own rebuild, and the migration that created the
-// index is recorded complete, so its CREATE INDEX IF NOT EXISTS never runs
-// again. Dropping the index directly is the only way to reproduce that: no
-// store method can express "manifest incomplete, cycle finished".
+// Mithril sync produced it: BuildCritical leaves the marker set for the lazy
+// remainder, and updateMithrilReadyState then runs ClearSyncState, an
+// unqualified DELETE FROM sync_state, which removes it (fixed for new syncs in
+// mithril/sync_import.go; databases bootstrapped before that fix stay in this
+// state). The migration that created the index is recorded complete, so its
+// CREATE INDEX IF NOT EXISTS never runs again. Dropping the index directly is
+// the only way to reproduce it here: no store method can express "manifest
+// incomplete, cycle finished".
 func seedClearedMarkerWithMissingCriticalIndex(
 	t *testing.T,
 	db *database.Database,
@@ -78,8 +82,8 @@ func metadataIndexExists(t *testing.T, raw *sql.DB, name string) bool {
 //
 // The pending marker records that a bulk-load cycle was interrupted. It does
 // not record which indexes exist, so it cannot answer the question the
-// rollback path asks. A database whose marker was cleared by a binary with a
-// smaller critical set carries the gap permanently: nothing else recreates an
+// rollback path asks. A database whose marker a Mithril sync wiped after the
+// critical rebuild carries the gap permanently: nothing else recreates an
 // index the schema migration already claims to have built.
 func TestRepairDeferredIndexesRestoresCriticalIndexWithoutMarker(
 	t *testing.T,
@@ -185,4 +189,115 @@ func lazyManifestIndex(t *testing.T) string {
 	}
 	t.Fatal("the manifest must keep at least one lazy entry")
 	return ""
+}
+
+// namedMissingManager is a DeferredIndexManager that also lists its missing
+// critical entries, and records what had already been logged when the rebuild
+// was entered.
+type namedMissingManager struct {
+	missing []string
+	logged  string
+	log     *bytes.Buffer
+}
+
+func (m *namedMissingManager) DropDeferredIndexes() error { return nil }
+
+func (m *namedMissingManager) BuildCriticalDeferredIndexes() error {
+	m.logged = m.log.String()
+	return nil
+}
+
+func (m *namedMissingManager) BuildDeferredIndexes() error { return nil }
+
+func (m *namedMissingManager) HasDeferredIndexesPending() (bool, error) {
+	return false, nil
+}
+
+func (m *namedMissingManager) MissingCriticalDeferredIndexes() (
+	[]string,
+	error,
+) {
+	return m.missing, nil
+}
+
+// TestEnsureCriticalDeferredIndexesNamesMissingBeforeBuilding pins the
+// ordering: the names are logged before the rebuild is entered, not after it
+// returns.
+//
+// A rebuild of one index on a multi-million-row table takes minutes and emits
+// nothing while it runs, which is the silence the reported incident opened
+// with. The assertion reads the log as the rebuild sees it, so a message moved
+// back below the build fails here.
+func TestEnsureCriticalDeferredIndexesNamesMissingBeforeBuilding(
+	t *testing.T,
+) {
+	var buf bytes.Buffer
+	logger := slog.New(slog.NewTextHandler(&buf, nil))
+	manager := &namedMissingManager{
+		missing: []string{rollbackCascadeIndex, "idx_utxo_added_slot"},
+		log:     &buf,
+	}
+
+	require.NoError(t, ensureCriticalDeferredIndexes(manager, logger))
+
+	require.Contains(
+		t,
+		manager.logged,
+		"rebuilding missing critical deferred metadata indexes",
+		"the rebuild must be announced before it starts",
+	)
+	require.Contains(t, manager.logged, rollbackCascadeIndex)
+	require.Contains(t, manager.logged, "idx_utxo_added_slot")
+	require.Contains(
+		t,
+		buf.String(),
+		"critical deferred metadata index check complete",
+		"the completion line still reports the outcome",
+	)
+	require.Contains(
+		t,
+		buf.String(),
+		"duration_covers",
+		"the reported duration must say what it covers: the rebuild "+
+			"runs inside withDeferredIndexWrite, which restores missing "+
+			"retained indexes in the same write transaction",
+	)
+}
+
+// TestEnsureCriticalDeferredIndexesQuietWhenManifestComplete keeps the healthy
+// path silent: a complete manifest is one catalog lookup per entry and must
+// not add a startup log line.
+func TestEnsureCriticalDeferredIndexesQuietWhenManifestComplete(t *testing.T) {
+	var buf bytes.Buffer
+	logger := slog.New(slog.NewTextHandler(&buf, nil))
+	manager := &namedMissingManager{log: &buf}
+
+	require.NoError(t, ensureCriticalDeferredIndexes(manager, logger))
+
+	require.Empty(t, buf.String())
+}
+
+// TestRepairDeferredIndexesNamesMissingIndexAgainstRealStore runs the same
+// path against the SQLite store, so the listing is exercised through
+// MissingCriticalDeferredIndexes and the real catalog rather than a fake.
+func TestRepairDeferredIndexesNamesMissingIndexAgainstRealStore(t *testing.T) {
+	db := newFileTestDB(t)
+	raw := seedClearedMarkerWithMissingCriticalIndex(t, db)
+	var buf bytes.Buffer
+	logger := slog.New(slog.NewTextHandler(&buf, nil))
+
+	require.NoError(t, RepairDeferredIndexes(db, logger))
+
+	require.True(t, metadataIndexExists(t, raw, rollbackCascadeIndex))
+	require.Contains(
+		t,
+		buf.String(),
+		"rebuilding missing critical deferred metadata indexes",
+	)
+	require.Contains(
+		t,
+		buf.String(),
+		rollbackCascadeIndex,
+		"the operator must be told which index the wait is for",
+	)
 }
--- a/internal/node/deferred_index_repair_test.go
+++ b/internal/node/deferred_index_repair_test.go
@@ -54,7 +54,7 @@ func seedClearedMarkerWithMissingCriticalIndex(
 	require.NoError(t, err)
 	require.False(
 		t,
-		metadataIndexExists(t, raw, rollbackCascadeIndex),
+		dbtest.MetadataIndexExists(t, raw, rollbackCascadeIndex),
 		"%s must be absent for this to test the repair",
 		rollbackCascadeIndex,
 	)
@@ -64,16 +64,6 @@ func seedClearedMarkerWithMissingCriticalIndex(
 	)
 	require.NoError(t, err)
 	return raw
-}
-
-func metadataIndexExists(t *testing.T, raw *sql.DB, name string) bool {
-	t.Helper()
-	var count int
-	require.NoError(t, raw.QueryRow(
-		"SELECT COUNT(*) FROM sqlite_master WHERE type = 'index' AND name = ?",
-		name,
-	).Scan(&count))
-	return count == 1
 }
 
 // TestRepairDeferredIndexesRestoresCriticalIndexWithoutMarker covers the
@@ -96,7 +86,7 @@ func TestRepairDeferredIndexesRestoresCriticalIndexWithoutMarker(
 
 	require.True(
 		t,
-		metadataIndexExists(t, raw, rollbackCascadeIndex),
+		dbtest.MetadataIndexExists(t, raw, rollbackCascadeIndex),
 		"core-mode serve must restore %s before the node can roll back",
 		rollbackCascadeIndex,
 	)
@@ -118,7 +108,7 @@ func TestRepairCriticalDeferredIndexesRestoresCriticalIndexWithoutMarker(
 
 	require.True(
 		t,
-		metadataIndexExists(t, raw, rollbackCascadeIndex),
+		dbtest.MetadataIndexExists(t, raw, rollbackCascadeIndex),
 		"the critical repair must restore %s",
 		rollbackCascadeIndex,
 	)
@@ -150,7 +140,7 @@ func TestRepairDeferredIndexesFinishesPendingCycle(t *testing.T) {
 	db := newFileTestDB(t)
 	raw, err := dbtest.RawSQLiteMetadata(t, db)
 	require.NoError(t, err)
-	lazy := lazyManifestIndex(t)
+	lazy := dbtest.LazyManifestIndex(t)
 	for _, name := range []string{rollbackCascadeIndex, lazy} {
 		_, err = raw.Exec("DROP INDEX IF EXISTS " + name)
 		require.NoError(t, err)
@@ -168,27 +158,13 @@ func TestRepairDeferredIndexesFinishesPendingCycle(t *testing.T) {
 		slog.New(slog.NewTextHandler(io.Discard, nil)),
 	))
 
-	require.True(t, metadataIndexExists(t, raw, rollbackCascadeIndex))
+	require.True(t, dbtest.MetadataIndexExists(t, raw, rollbackCascadeIndex))
 	require.True(
 		t,
-		metadataIndexExists(t, raw, lazy),
+		dbtest.MetadataIndexExists(t, raw, lazy),
 		"a pending cycle must still finish the lazy remainder",
 	)
 	requireNoPendingMarker(t, raw)
-}
-
-// lazyManifestIndex returns the name of a non-critical manifest entry, so the
-// test above distinguishes the full rebuild from the critical subset without
-// pinning a specific index the manifest may reclassify later.
-func lazyManifestIndex(t *testing.T) string {
-	t.Helper()
-	for _, index := range deferred.Manifest {
-		if !index.Critical {
-			return index.Name
-		}
-	}
-	t.Fatal("the manifest must keep at least one lazy entry")
-	return ""
 }
 
 // namedMissingManager is a DeferredIndexManager that also lists its missing
@@ -288,7 +264,7 @@ func TestRepairDeferredIndexesNamesMissingIndexAgainstRealStore(t *testing.T) {
 
 	require.NoError(t, RepairDeferredIndexes(db, logger))
 
-	require.True(t, metadataIndexExists(t, raw, rollbackCascadeIndex))
+	require.True(t, dbtest.MetadataIndexExists(t, raw, rollbackCascadeIndex))
 	require.Contains(
 		t,
 		buf.String(),
--- a/internal/node/deferred_index_repair_test.go
+++ b/internal/node/deferred_index_repair_test.go
@@ -1,0 +1,188 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package node
+
+import (
+	"database/sql"
+	"io"
+	"log/slog"
+	"testing"
+
+	"github.com/blinklabs-io/dingo/database"
+	"github.com/blinklabs-io/dingo/database/plugin/metadata/deferred"
+	dbtest "github.com/blinklabs-io/dingo/internal/test/dbtest"
+	"github.com/stretchr/testify/require"
+)
+
+// rollbackCascadeIndex is the child index for fk_transaction_outputs, the
+// foreign key the rollback's DELETE FROM "transaction" cascades through.
+const rollbackCascadeIndex = "idx_utxo_transaction_id"
+
+// seedClearedMarkerWithMissingCriticalIndex leaves the database in the state
+// two Mithril-bootstrapped preview nodes were found in: a critical manifest
+// index absent, and no pending marker to say so.
+//
+// A binary whose critical subset did not yet include the index cleared the
+// marker when it finished its own rebuild, and the migration that created the
+// index is recorded complete, so its CREATE INDEX IF NOT EXISTS never runs
+// again. Dropping the index directly is the only way to reproduce that: no
+// store method can express "manifest incomplete, cycle finished".
+func seedClearedMarkerWithMissingCriticalIndex(
+	t *testing.T,
+	db *database.Database,
+) *sql.DB {
+	t.Helper()
+	raw, err := dbtest.RawSQLiteMetadata(t, db)
+	require.NoError(t, err)
+	_, err = raw.Exec("DROP INDEX IF EXISTS " + rollbackCascadeIndex)
+	require.NoError(t, err)
+	require.False(
+		t,
+		metadataIndexExists(t, raw, rollbackCascadeIndex),
+		"%s must be absent for this to test the repair",
+		rollbackCascadeIndex,
+	)
+	_, err = raw.Exec(
+		"DELETE FROM sync_state WHERE sync_key = ?",
+		deferred.SyncStateKey,
+	)
+	require.NoError(t, err)
+	return raw
+}
+
+func metadataIndexExists(t *testing.T, raw *sql.DB, name string) bool {
+	t.Helper()
+	var count int
+	require.NoError(t, raw.QueryRow(
+		"SELECT COUNT(*) FROM sqlite_master WHERE type = 'index' AND name = ?",
+		name,
+	).Scan(&count))
+	return count == 1
+}
+
+// TestRepairDeferredIndexesRestoresCriticalIndexWithoutMarker covers the
+// core-mode serve path: cmd/dingo/serve.go calls RepairDeferredIndexes for
+// every non-API storage mode.
+//
+// The pending marker records that a bulk-load cycle was interrupted. It does
+// not record which indexes exist, so it cannot answer the question the
+// rollback path asks. A database whose marker was cleared by a binary with a
+// smaller critical set carries the gap permanently: nothing else recreates an
+// index the schema migration already claims to have built.
+func TestRepairDeferredIndexesRestoresCriticalIndexWithoutMarker(
+	t *testing.T,
+) {
+	db := newFileTestDB(t)
+	raw := seedClearedMarkerWithMissingCriticalIndex(t, db)
+	logger := slog.New(slog.NewTextHandler(io.Discard, nil))
+
+	require.NoError(t, RepairDeferredIndexes(db, logger))
+
+	require.True(
+		t,
+		metadataIndexExists(t, raw, rollbackCascadeIndex),
+		"core-mode serve must restore %s before the node can roll back",
+		rollbackCascadeIndex,
+	)
+	requireNoPendingMarker(t, raw)
+}
+
+// TestRepairCriticalDeferredIndexesRestoresCriticalIndexWithoutMarker covers
+// the API-mode path: resumeBackfill calls RepairCriticalDeferredIndexes both
+// when it runs a backfill and when it finds none needed, and a node in either
+// state begins rolling back as soon as live sync resumes.
+func TestRepairCriticalDeferredIndexesRestoresCriticalIndexWithoutMarker(
+	t *testing.T,
+) {
+	db := newFileTestDB(t)
+	raw := seedClearedMarkerWithMissingCriticalIndex(t, db)
+	logger := slog.New(slog.NewTextHandler(io.Discard, nil))
+
+	require.NoError(t, RepairCriticalDeferredIndexes(db, logger))
+
+	require.True(
+		t,
+		metadataIndexExists(t, raw, rollbackCascadeIndex),
+		"the critical repair must restore %s",
+		rollbackCascadeIndex,
+	)
+	requireNoPendingMarker(t, raw)
+}
+
+// requireNoPendingMarker asserts the repair did not invent a bulk-load cycle:
+// the marker means "a drop/rebuild is outstanding", and setting it on a
+// database with a complete manifest would send the next startup through a
+// rebuild it does not need.
+func requireNoPendingMarker(t *testing.T, raw *sql.DB) {
+	t.Helper()
+	var count int
+	require.NoError(t, raw.QueryRow(
+		"SELECT COUNT(*) FROM sync_state WHERE sync_key = ?",
+		deferred.SyncStateKey,
+	).Scan(&count))
+	require.Zero(
+		t,
+		count,
+		"repairing a missing index must not mark a rebuild pending",
+	)
+}
+
+// TestRepairDeferredIndexesFinishesPendingCycle keeps the marker's existing
+// meaning intact: when a cycle really was interrupted, the core-mode repair
+// still rebuilds the whole manifest and clears it.
+func TestRepairDeferredIndexesFinishesPendingCycle(t *testing.T) {
+	db := newFileTestDB(t)
+	raw, err := dbtest.RawSQLiteMetadata(t, db)
+	require.NoError(t, err)
+	lazy := lazyManifestIndex(t)
+	for _, name := range []string{rollbackCascadeIndex, lazy} {
+		_, err = raw.Exec("DROP INDEX IF EXISTS " + name)
+		require.NoError(t, err)
+	}
+	_, err = raw.Exec(
+		`INSERT INTO sync_state (sync_key, value) VALUES (?, ?)
+		 ON CONFLICT (sync_key) DO UPDATE SET value = excluded.value`,
+		deferred.SyncStateKey,
+		deferred.SyncStateValue,
+	)
+	require.NoError(t, err)
+
+	require.NoError(t, RepairDeferredIndexes(
+		db,
+		slog.New(slog.NewTextHandler(io.Discard, nil)),
+	))
+
+	require.True(t, metadataIndexExists(t, raw, rollbackCascadeIndex))
+	require.True(
+		t,
+		metadataIndexExists(t, raw, lazy),
+		"a pending cycle must still finish the lazy remainder",
+	)
+	requireNoPendingMarker(t, raw)
+}
+
+// lazyManifestIndex returns the name of a non-critical manifest entry, so the
+// test above distinguishes the full rebuild from the critical subset without
+// pinning a specific index the manifest may reclassify later.
+func lazyManifestIndex(t *testing.T) string {
+	t.Helper()
+	for _, index := range deferred.Manifest {
+		if !index.Critical {
+			return index.Name
+		}
+	}
+	t.Fatal("the manifest must keep at least one lazy entry")
+	return ""
+}
--- a/internal/node/load.go
+++ b/internal/node/load.go
@@ -379,10 +379,50 @@ func WithDeferredIndexes(
 	return &DeferredIndexRebuilder{manager: manager}
 }
 
+// criticalIndexRebuildLogThreshold is how long the critical-index check
+// may take before it is reported. A complete manifest costs one catalog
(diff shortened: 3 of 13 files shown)


chadaibbhot-git:
CI note: `go-test (macOS)` failed on this head in
`internal/koiosparity` — `TestFetchAccountRewardsForEpochMegaScenario`,
`"17" is not greater than or equal to "19"` ("every ordinary chunk dispatched
before the failing ones must have checkpointed"). That is unrelated to this
change and looks like a runner-timing flake:

- The diff touches `database/plugin/metadata/deferred`, `internal/node`,
  `cmd/dingo` and two docs files. `internal/koiosparity` never calls
  `DropDeferredIndexes`, `BuildCriticalDeferredIndexes` or the repair helpers,
  so nothing in this change runs on that test's path.
- The assertion is a scheduling lower bound by its own comment: how many of the
  19 ordinary chunks checkpoint before the fast-failing chunk cancels the rest
  depends on how quickly cancellation propagates across
  `accountFetchConcurrency` slots. A slow runner can land under 19.
- `go test ./internal/koiosparity/ -run TestFetchAccountRewardsForEpochMegaScenario -count=10`
  passes 10/10 on clean `main` (7351b088) and 10/10 on this branch.

I cannot re-run jobs on this repository, so the red check will stay on this head
until the next push re-triggers CI. Happy to rebase to re-trigger on request. If
it recurs I will tighten that assertion in its own PR rather than here.

chadaibbhot-git:
CI note on the rebased head `64a68c8e` (rebased onto `d6fefc4a`, no conflicts).

**macOS is green.** The `internal/koiosparity` scheduling flake reported above
(`TestFetchAccountRewardsForEpochMegaScenario`, `"17" is not greater than or equal to "19"`)
did not reproduce: `go-test (macOS)` passed in 5m49s. The rebase also picks up #4124,
which serializes the koiosparity cache writers that test checkpoints through.

**`go-test (Windows)` was cancelled at the 90-minute job cap, not failed**, and the
evidence says that is the known wedge in #3746 rather than anything in this branch:

- Run 34224033192. The `go test` step began at 12:09:09Z. The last package line is
  `ok github.com/blinklabs-io/dingo/internal/dblifecycle 57.412s` at 12:23:23Z, and then
  there is no output of any kind for 71 minutes until `The operation was canceled` at
  13:34:40Z.
- The log contains no `panic: test timed out`, so the per-package `-timeout 30m` never
  fired. A genuinely hung Go test produces a goroutine dump naming itself; this produces
  nothing, which is the "wedges rather than fails" behaviour #3746 records.
- The same silent 90-minute wedge, with the same absence of a goroutine dump, has hit
  three other branches in the last two days at three different stall points:
  `fix/midnight-indexer-config` (run 34165731917, stalled after `internal/dblifecycle`,
  the identical point), `ci/config-parity` (run 34154148885, 95 minutes, no package
  output at all), and #3963 twice (after `mithril`, and after
  `internal/benchci/cmd/benchcheck`).
- #3963's Windows job then ran green in 10m16s on its rebased head with an unchanged
  test set, which is the run-to-run spread #3746 asks to be characterised.

This branch changes `database/plugin/metadata/deferred`, `internal/node/load.go` and
three test files; it adds no timing-sensitive or platform-sensitive test, and the four
stall points are in packages it does not touch.

I do not have permission to re-run the job (`403 Must have admin rights to Repository`).
If someone with write access re-runs `go-test (Windows)` for this head it should clear;
otherwise I will re-trigger it with the next rebase when main moves. Everything else on
this head is green: Linux, Linux race, macOS, lint, DCO, govulncheck, docker,
Conventional Commits.

Local verification on the rebased head, all green:

```
go build ./... && go vet ./...
go test ./database/plugin/metadata/... ./internal/node/ ./cmd/dingo/ ./database/ ./mithril/... -count=1
go test -race ./database/plugin/metadata/deferred/ ./database/plugin/metadata/sqlite/ ./internal/node/ ./cmd/dingo/ -count=1
golangci-lint run ./internal/node/... ./database/plugin/metadata/... ./cmd/dingo/...   # 0 issues
```

Both production hunks were reverted again individually on the rebased tree and each
one turns its tests red: the manifest hunk fails
`TestCascadeChildColumnsIndexedAfterCriticalRebuild` and
`TestUtxoCascadeDeleteIndexedAfterCriticalRebuild` (and the pinned critical count,
`got 12 entries, want 13`); the `RepairDeferredIndexes` hunk fails
`TestServeCoreModeRepairsMissingCascadeIndex` and
`TestRepairDeferredIndexesRestoresCriticalIndexWithoutMarker`; the
`RepairCriticalDeferredIndexes` hunk fails
`TestRepairCriticalDeferredIndexesRestoresCriticalIndexWithoutMarker`.

chadaibbhot-git:
Follow-up to the CI note above: `go-test (Windows)` was re-triggered on this same head
`64a68c8e` and **passed in 12m49s** (13:42:49Z to 13:55:38Z). Every check on that run is
green: Linux, Linux race, macOS, Windows, lint, DCO, govulncheck, docker, Conventional
Commits.

The `cancelled` Windows entry still listed against this commit is the earlier,
superseded run (12:04:19Z to 13:34:44Z), which is the 90-minute wedge described above
and in #3746. Nothing in the tree changed between the two runs — same commit, same test
set, 90-minute wedge then a 12m49s pass — which is the run-to-run spread that issue asks
to be characterised.

Marking ready for review.

chadaibbhot-git:
@coderabbitai review (final head 64a68c8e)

internal/node/load.go
```diff
@@ -379,10 +379,50 @@ func WithDeferredIndexes(
...
+//
+// The pending marker records that a cycle was interrupted; it does not
+// record which indexes exist. A binary whose critical subset was smaller
+// than this one's clears the marker with an index this manifest calls
+// critical still missing, and nothing else recreates it: the schema
+// migration that created it is recorded complete, so its
+// CREATE INDEX IF NOT EXISTS never runs again, and the newer manifest is
+// only consulted by a cycle that is no longer pending. Two
+// Mithril-bootstrapped preview nodes were found in exactly that state,
+// missing the utxo child index the rollback DELETE cascades through.
```
chrisguiney: The cleared-marker state has a current cause rather than a historical one, and the difference decides scope.

`mithril/sync.go` calls `deferredIndexes.BuildCritical()` and then `updateMithrilReadyState(..., clearSyncState=true)`, which runs `db.ClearSyncState(txn)` (`mithril/sync_import.go`). That query is `DELETE FROM sync_state` with no `WHERE` clause (`database/plugin/metadata/sqlstore/queries/sqlite/operational.sql`). Every successful Mithril sync therefore deletes `metadata_indexes_pending` moments after `BuildCritical` deliberately left it set.

Reproduced against this branch: after `DropDeferredIndexes()` + `BuildCriticalDeferredIndexes()`, `HasDeferredIndexesPending()` is true and the lazy `idx_asset_name_hex` is absent; after `db.ClearSyncState(nil)`, pending is false and the lazy index is still absent. That is exactly the A/B table's "every `Critical` index present, every lazy one absent, no `metadata_indexes_pending` row", produced by current `main` with no older binary involved.

The "binary whose critical subset was smaller" path cannot produce it. `BuildDeferredIndexes` builds the whole manifest and deletes the marker in one transaction, so no version of this code clears the marker while a manifest entry is missing.

Two consequences:

- The description's reason for leaving `mithril/sync.go` alone — "on the core-mode path the marker `BuildCritical` leaves set sends the first `dingo serve` through the full `BuildDeferredIndexes` anyway" — does not hold. The first `serve` finds no marker.
- All twelve lazy manifest entries (`idx_asset_name_hex`, `idx_asset_fingerprint`, `idx_asset_amount`, `idx_datum_added_slot`, `idx_certs_block_hash`, `idx_certs_certificate_id`, `idx_certs_cert_type`, `idx_redeemer_index`, `idx_redeemer_tag`, `idx_key_witness_type`, `idx_witness_scripts_script_hash`, `idx_witness_scripts_type`) are permanently absent on every Mithril-bootstrapped database, API-mode ones included, where they serve the query paths the deferral was designed to restore. `Node.startDeferredIndexMaintenance` reads the same wiped marker and logs "deferred-index maintenance not needed".

The repair added here is correct and independent of which mechanism cleared the marker, so none of it needs to change. What needs correcting is the stated cause, here and in `ARCHITECTURE.md`, and an explicit position on the lazy remainder. `updateMithrilReadyState` already re-writes `mithril_immutable_max` after the clear precisely because the clear is indiscriminate; carrying `deferred.SyncStateKey` across it the same way would keep the marker's meaning intact and let the existing maintenance path finish the manifest.

ARCHITECTURE.md
```diff
@@ -5679,7 +5683,11 @@ critical subset before clearing `sync_status`, then leaves the pending
 sync-state marker set. API-mode `serve` verifies the critical subset before
 startup and runs the full lazy rebuild as background maintenance; the marker is
 cleared only after the full manifest has been rebuilt. Core-mode startup still
-repairs the full manifest synchronously before serving. On MySQL, InnoDB
+repairs the full manifest synchronously before serving. Both repair entry
+points also restore any missing critical index when no cycle is pending at all:
+the marker records that a cycle was interrupted, not which indexes exist, so a
+database whose marker an older binary cleared while its own critical subset was
+smaller would otherwise carry the gap permanently. On MySQL, InnoDB
```
chrisguiney: Same correction as on `internal/node/load.go`: the marker is not left behind by an older binary with a smaller critical subset. `updateMithrilReadyState` (`mithril/sync_import.go`) clears the entire `sync_state` table on every completed Mithril sync, immediately after `BuildCritical`.

Worth naming here in particular, because the sentence two lines up — "the marker is cleared only after the full manifest has been rebuilt" — is contradicted by that same unqualified `DELETE FROM sync_state`, and this paragraph is where a future reader would look for the answer.

internal/node/load.go
```diff
@@ -379,10 +379,50 @@ func WithDeferredIndexes(
...
+	if err := manager.BuildCriticalDeferredIndexes(); err != nil {
+		return err
+	}
+	elapsed := time.Since(start)
+	if elapsed >= criticalIndexRebuildLogThreshold {
+		logger.Info(
+			"rebuilt missing critical deferred metadata indexes",
+			"duration", elapsed,
+		)
+	}
```
chrisguiney, line 454: Non-blocking. The message lands after the work, so a rebuild that takes a while is silent for exactly as long as it runs — the symptom the description opens with. Logging before the build, naming the entries found missing, would attribute the wait while an operator is watching it rather than after they have stopped.

Related: `elapsed` covers more than the critical subset. `BuildCriticalDeferredIndexes` goes through `withDeferredIndexWrite`, which runs `ensureRetainedIndexes` first. That is a real gain — `deferred.Retained` self-healing was unreachable on a start with no cycle pending, and this change makes it reachable on every core-mode start — but it means time spent building, say, `idx_key_witness_transaction_id` on a large `key_witness` table is reported as "rebuilt missing critical deferred metadata indexes".

internal/node/deferred_index_repair_test.go
```diff
@@ -0,0 +1,188 @@
...
+		t,
+		count,
+		"repairing a missing index must not mark a rebuild pending",
+	)
+}
+
+// TestRepairDeferredIndexesFinishesPendingCycle keeps the marker's existing
+// meaning intact: when a cycle really was interrupted, the core-mode repair
+// still rebuilds the whole manifest and clears it.
+func TestRepairDeferredIndexesFinishesPendingCycle(t *testing.T) {
```
chrisguiney, line 139: Non-blocking, and about the description rather than this code. This test and `TestRollbackDeleteCascadesIntoUtxo` both pass with either production hunk reverted in place, and with both reverted; the other five fail. They are guards — one on the marker's existing meaning, one on the cascade being real — which is what their doc comments say, so only "Every test below fails on `main`" overstates them.

chrisguiney (changes requested):
The rollback-cascade diagnosis holds and the fix is the right shape. `idx_utxo_transaction_id` was the only child column of a rollback-deleted `ON DELETE CASCADE` parent left unindexed after a critical rebuild, and the manifest classification plus the unconditional critical repair both prove out: reverting each production hunk on its own makes five of the seven new tests fail, and reverting both makes the same five fail.

One blocker, on the rationale rather than the code. The state this repairs has a live cause on `main`, not the historical one recorded in `internal/node/load.go` and `ARCHITECTURE.md`, and the misattribution is what justified leaving `mithril/sync.go` unchanged. Detail inline.

Class audit, complete: `utxo` carries a second cascade to `transaction(id)` (`fk_transaction_collateral_return` on `collateral_return_for_tx_id`), and a cascaded `utxo` delete cascades on into `asset` and `utxo_reference_input`. All three child columns are covered by indexes outside the manifest (`idx_utxo_collateral_return_for_tx_id`, `idx_asset_utxo_id`, `sqlite_autoindex_utxo_reference_input_1`) and resolve as index searches in the dropped-manifest plan, so this index was the only gap. The remaining cascades into `transaction(id)` — `certs`, `key_witness`, `witness_scripts`, `redeemer`, `plutus_data` — are all in `deferred.Retained`.

Checks run against the head: `go build ./...`, `go vet`, `make docs-parity`, `make import-boundaries`, `golangci-lint` over the changed packages (0 issues), and `-race` over `database/plugin/metadata/deferred`, `database/plugin/metadata/sqlite`, `internal/node` and `cmd/dingo` — all pass, and all still pass with current `main` merged in.

internal/node/load.go
chadaibbhot-git: Verified independently before changing anything, and you are right — the cause I
stated was wrong.

Reproduced on this branch: `DropDeferredIndexes()` + `BuildCriticalDeferredIndexes()`
leaves `HasDeferredIndexesPending()` true, then `updateMithrilReadyState(..., clearSyncState=true)`
alone makes it false. `db.ClearSyncState` is `DELETE FROM sync_state` with no
`WHERE`, so the marker goes with everything else, moments after `BuildCritical`
set it deliberately. No older binary is involved, and the "smaller critical
subset" story cannot produce it for the reason you give: `BuildDeferredIndexes`
builds the whole manifest and deletes the marker in one transaction.

Position, and it is the one you point at: fixed at the root rather than
described. `updateMithrilReadyState` now carries `deferred.SyncStateKey` across
the clear exactly the way `mithril_ledger_slot`/`mithril_ledger_hash` are
re-written after it — same block, same reason (the clear is indiscriminate, so
every row a completed sync still needs is restored explicitly). `mithril/sync.go`
still calls `BuildCritical()` only; the marker it leaves now survives, so the
first `serve` runs the existing `BuildDeferredIndexes` maintenance and finishes
the lazy manifest. Nothing new carries the work.

Red→green, in `mithril/sync_import_deferred_marker_test.go`:

- `TestUpdateMithrilReadyStateKeepsDeferredIndexPendingMarker` — after
  `BuildCritical` + `updateMithrilReadyState(clearSyncState=true)`,
  `HasDeferredIndexesPending()` is true and `sync_status` is still cleared.
  Fails on the previous head with exactly your observation.
- `TestMithrilSyncLeavesLazyManifestForTheFirstServe` — the whole composition:
  sync's drop + `BuildCritical` + `updateMithrilReadyState`, then
  `node.RepairDeferredIndexes` as the first `serve`. A lazy manifest entry
  exists afterwards and the marker is cleared by the rebuild that consumed it.
  Also red before the fix (`idx_asset_name_hex` absent).

Both mutation-checked: reverting only the carry in `updateMithrilReadyState`
turns both red again.

The repair in this branch is unchanged, as you say it should be — it is what
fixes databases already on disk, whose marker the sync that built them wiped;
the migration that created those indexes is recorded complete and never re-runs.
What that repair does not do is bring back the twelve lazy entries on an
already-bootstrapped database: it is deliberately the critical subset, because
building the API-only remainder on a core-mode node is the cost the deferral
exists to avoid. If you would rather have the full manifest reconciled at
startup in API mode regardless of the marker, say so and I will do it in its own
PR — it is the same catalog-lookup-per-entry check, just over `deferred.Manifest`.

The stated cause is corrected in `internal/node/load.go`, `ARCHITECTURE.md`, the
test doc comments that repeated it, and the PR description.

ARCHITECTURE.md
chadaibbhot-git: Corrected here, including the sentence two lines up. That paragraph now reads
that the rebuild paths clear the marker only after the full manifest is
rebuilt **but are not the only writer of that row**: `ClearSyncState`
(`DELETE FROM sync_state`, no `WHERE`) removes it too, and Mithril sync runs
that clear through `updateMithrilReadyState` immediately after the critical
rebuild — which is why that function re-writes every row a completed sync still
needs (`mithril_ledger_slot`, `mithril_ledger_hash`, and now the deferred-index
marker) back after the clear.

The following sentence no longer attributes the cleared-marker state to an older
binary with a smaller critical subset; it attributes it to a database
bootstrapped by a binary that predates the marker being carried across the clear,
which is every release so far, and explains why the gap is permanent there (the
schema migration that created those indexes is recorded complete and never
re-runs).

internal/node/load.go
chadaibbhot-git: Both taken.

The missing entries are now named before the rebuild starts:

```
INFO rebuilding missing critical deferred metadata indexes indexes=idx_utxo_transaction_id count=1
```

That needed a way to ask which critical entries are absent without building
anything, so the SQL store gained
`MissingCriticalDeferredIndexes()` behind a new optional
`metadata.MissingCriticalDeferredIndexLister` (catalog reads only, no DDL,
manifest order). Stores that do not implement it fall back to the previous
after-the-fact line rather than failing a startup over a log message, and a
read error on the catalog query is a warning, not an error return.

`elapsed` is now attributed honestly rather than timed separately: the
completion line carries
`duration_covers="critical rebuild and retained-index restore"`, because
`BuildCriticalDeferredIndexes` goes through `withDeferredIndexWrite` and
`ensureRetainedIndexes` runs inside the same write transaction. Timing the two
halves apart would mean either splitting that transaction or plumbing timing out
of the store, and neither is worth it for a startup log line — but the message no
longer claims the whole duration was the critical subset.

And explicitly, since it is a behaviour change worth naming rather than
discovering: `deferred.Retained` self-healing was previously unreachable on a
start with no cycle pending, and this PR makes it run on every core-mode start.
That is a deliberate gain, and it is now stated in the PR description too. A
healthy database still logs nothing at all — the completion line is emitted only
when something was missing or the check ran past one second.

Tests: `TestEnsureCriticalDeferredIndexesNamesMissingBeforeBuilding` asserts the
ordering from inside the rebuild (the fake records what had been logged when
`BuildCriticalDeferredIndexes` was entered), so moving the message back below the
build fails it; `TestRepairDeferredIndexesNamesMissingIndexAgainstRealStore`
runs the same path against the SQLite store and its real catalog; and
`TestEnsureCriticalDeferredIndexesQuietWhenManifestComplete` keeps the healthy
path silent.

internal/node/deferred_index_repair_test.go
chadaibbhot-git: Fixed in the description. It now splits the list explicitly:

- Fail on `main`, pass with the change: `TestCascadeChildColumnsIndexedAfterCriticalRebuild`,
  `TestUtxoCascadeDeleteIndexedAfterCriticalRebuild`,
  `TestRepairDeferredIndexesRestoresCriticalIndexWithoutMarker`,
  `TestRepairCriticalDeferredIndexesRestoresCriticalIndexWithoutMarker`,
  `TestServeCoreModeRepairsMissingCascadeIndex`, plus the new
  `TestUpdateMithrilReadyStateKeepsDeferredIndexPendingMarker`,
  `TestMithrilSyncLeavesLazyManifestForTheFirstServe`,
  `TestEnsureCriticalDeferredIndexesNamesMissingBeforeBuilding`,
  `TestRepairDeferredIndexesNamesMissingIndexAgainstRealStore` and
  `TestEnsureCriticalDeferredIndexesQuietWhenManifestComplete`.
- Guards, passing on `main` and with either production hunk reverted:
  `TestRollbackDeleteCascadesIntoUtxo` and
  `TestRepairDeferredIndexesFinishesPendingCycle`, described as what they hold —
  the cascade being real, and the marker keeping its existing meaning.

Re-confirmed on the current head rather than taken on faith: with
`Critical: true` reverted, the two cascade tests and the pinned critical-count
test fail while `TestRollbackDeleteCascadesIntoUtxo` passes; with the no-marker
repair reverted, the three repair tests and the `serveRun` test fail while
`TestRepairDeferredIndexesFinishesPendingCycle` passes.

mithril/sync_import_deferred_marker_test.go
```diff
@@ -0,0 +1,187 @@
...
+	require.True(t, ok)
+	pending, err := manager.HasDeferredIndexesPending()
+	require.NoError(t, err)
+	require.False(
+		t, pending,
+		"the full rebuild clears the marker it consumed",
+	)
+}
+
+func metadataIndexExists(t *testing.T, raw *sql.DB, name string) bool {
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid, fixed. Both helpers now live in `internal/test/dbtest`
(`MetadataIndexExists`, `LazyManifestIndex`), next to `RawSQLiteMetadata`, which
is what both call sites already use to reach the catalog. `internal/node` and
`mithril` use the shared ones; the duplicate definitions are gone.

chadaibbhot-git:
CI status on `8e849a56`: everything green except `go-test (Windows)`, which
failed in a package this PR does not touch.

```
--- FAIL: TestFetchEndorserBlockByPointDeadlineDoesNotCoolDownPeer (2.00s)
    ouroboros/leios_certified_recovery_test.go:654:
        Error: Should be zero, but was 1
```

That is `guard.consecutiveFailures.Load()` after the caller's 2 s deadline
expires — a timing assertion on a loaded runner, not a consequence of anything
here. Evidence it is a flake rather than this branch:

- The identical production tree passed `go-test (Windows)` on the previous head
  `5b15e58c`. The only delta in `8e849a56` is moving two test helpers
  (`MetadataIndexExists`, `LazyManifestIndex`) into `internal/test/dbtest`; no
  production file changed, and the move adds no package to `dbtest`'s
  transitive imports (`deferred` was already reachable through the sqlite
  plugin).
- This PR's diff touches no file under `ouroboros/`:
  `ARCHITECTURE.md`, `DATABASE.md`, `cmd/dingo/serve_deferred_index_repair_test.go`,
  `database/plugin/metadata/deferred/*`, `database/plugin/metadata/deferred_indexes.go`,
  `database/plugin/metadata/sqlite/shared_sqlstore_fk_cascade_index_test.go`,
  `database/plugin/metadata/sqlstore/deferred_indexes.go`,
  `internal/node/load.go` + its test, `internal/test/dbtest/deferred_indexes.go`,
  `mithril/sync_import.go` + its test.
- `go test ./ouroboros/ -run TestFetchEndorserBlockByPointDeadlineDoesNotCoolDownPeer -count=50`
  is green on clean `main` (`86c0be79`), and `-count=30` is green on this head.
  Linux, so it does not reproduce the Windows timing; it does rule out a
  deterministic break introduced here.

I have no permission to re-run the job on this repo, and I am not pushing a
no-op commit to re-trigger it. The branch is rebased on `86c0be79` with no
conflicts, so there is nothing to rebase onto either. If a maintainer re-runs
that one job I expect it green; if it reproduces, it wants its own PR against
the guard's failure accounting on the deadline path, which I am happy to take
on separately.

internal/node/load.go
chrisguiney: Resolved at `27d765fa`, and the scope question it raised is answered in code rather than in prose.

`updateMithrilReadyState` now reads `deferred.SyncStateKey` before `db.ClearSyncState` and re-writes it after, inside the same write transaction (`mithril/sync_import.go`). The ordering it depends on holds: `mithril/sync.go:1273` runs `BuildCritical()` and `mithril/sync.go:1294` runs `updateMithrilReadyState(..., clearSyncState=true)`, so the marker exists when the clear reaches it. The earlier call at `mithril/sync.go:1209` passes `clearSyncState=false` and takes the `syncStatus` branch, so it never reaches the clear.

The `!= ""` guard is safe on a database that has no marker: `Store.GetSyncState` maps `sql.ErrNoRows` to `("", nil)` (`database/plugin/metadata/sqlstore/operational.go:188`), so the read cannot fail a Mithril sync over an absent row.

Reverting `mithril/sync_import.go` to the merge base in place fails exactly `TestUpdateMithrilReadyStateKeepsDeferredIndexPendingMarker` and `TestMithrilSyncLeavesLazyManifestForTheFirstServe`, and nothing else in the new set.

ARCHITECTURE.md
chrisguiney: Resolved at `27d765fa`. The paragraph now states that the rebuild paths are not the only writer of the marker row and that `ClearSyncState` removes it too, which is the correction this thread asked for. `make docs-parity` passes on the head.

internal/node/load.go
chrisguiney: Both resolved at `27d765fa`. The missing entries are named before `BuildCriticalDeferredIndexes` is entered, and the completion line carries `duration_covers`.

Confirmed causally rather than by reading: deleting the pre-build `logger.Info` block in place fails `TestEnsureCriticalDeferredIndexesNamesMissingBeforeBuilding` and `TestRepairDeferredIndexesNamesMissingIndexAgainstRealStore`, so both halves of the fix are load-bearing.

internal/node/deferred_index_repair_test.go
chrisguiney: Resolved at `27d765fa` — the description now splits the two groups explicitly.

Re-measured at this head with each production hunk reverted in place. Ten of the thirteen new or changed tests fail when the hunk they cover is reverted; three pass with every hunk reverted: `TestRollbackDeleteCascadesIntoUtxo`, `TestRepairDeferredIndexesFinishesPendingCycle`, and `TestEnsureCriticalDeferredIndexesQuietWhenManifestComplete`. The third is a negative-case assertion on the new logging rather than a guard on pre-existing behavior, so listing it with the others is defensible; it is worth keeping either way.

mithril/sync_import_deferred_marker_test.go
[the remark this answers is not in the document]
chrisguiney: Confirmed fixed at `27d765fa`. `MetadataIndexExists` and `LazyManifestIndex` now live only in `internal/test/dbtest/deferred_indexes.go`; no duplicate definitions remain in `internal/node` or `mithril`.

chrisguiney (commented):
The blocker from the previous round is resolved, in code rather than in prose. `updateMithrilReadyState` now carries `deferred.SyncStateKey` across the unqualified `ClearSyncState`, so the marker Mithril sync leaves for the first `serve` survives the clear that used to erase it, and the rationale in `internal/node/load.go` and `ARCHITECTURE.md` is corrected to match. That hunk is load-bearing: reverting it alone fails both new `mithril` tests and nothing else in the set.

The rest holds. `TestCascadeChildColumnsIndexedAfterCriticalRebuild` states the invariant over `pragma_foreign_key_list` rather than over the one index, so a future cascading foreign key classified lazy fails there instead of in production — a stronger guarantee than the manual class audit that motivated it. The repair also lands before the sweep it protects: core mode runs `repairDeferredIndexes` at `cmd/dingo/serve.go:76`, ahead of `node.Run`, and API mode reaches `RepairCriticalDeferredIndexes` on both branches of `resumeBackfill`.

One correction to the description, non-blocking. "No separate issue" is no longer accurate: #4065 is open and describes this defect — same index, same `fk_transaction_outputs` cascade, same wedge — arriving from `database restore` rather than from Mithril sync. This change fixes its headline case, since `idx_utxo_transaction_id` is a critical manifest entry and the repair now runs unconditionally before the rollback. It does not close it: #4065 reports 14 of 234 declared indexes missing, and the repair reaches only the 25-entry manifest. Naming #4065 and saying which half this closes would keep the remainder from being re-investigated from scratch.

Approval is withheld on the pipeline, not on the diff. Both red jobs attribute to the base branch or the runner rather than to this change:

- `go-test (Linux) (1.26.x)`, the required context, failed at exit 100 in the setup step, with `apt-get` reporting `Failed to fetch https://dl.google.com/linux/chrome-stable/deb/dists/stable/main/binary-amd64/Packages.gz  Hash Sum mismatch`. It never reached `go test`. A re-run should clear it.
- `go-test (macOS) (1.26.x)` failed on `TestScheduler_ChangeInterval` at `ledger/timer_test.go:90`, "timer did not respect interval change, ran too frequently: 4 more ticks". The diff touches no file under `ledger/`, and that test passes consistently on repeated runs at both this head and `main`, so it is a timing-sensitive assertion flaking under runner load.

`go-test (Linux, race)`, `go-test (Windows)`, `lint`, `govulncheck` and `docker` are green on this head. The branch is 18 commits behind `main`; merging current `main` in builds and tests the affected packages clean, so the staleness is not hiding a conflict.

mithril/sync_import.go
```diff
@@ -129,9 +130,39 @@ func updateMithrilReadyState(
...
+			// across it explicitly — mithril_ledger_slot and
+			// mithril_ledger_hash below, and the deferred-index marker
+			// here. Mithril sync rebuilds only the critical subset
+			// (BuildCritical, sync.go) and deliberately leaves
+			// deferred.SyncStateKey set so the first serve's maintenance
+			// pass finishes the lazy manifest. Dropping it here erases
+			// that instruction moments after it was written, and the lazy
+			// entries are then never built on a Mithril-bootstrapped
+			// database.
+			deferredIndexesPending, err := db.GetSyncState(
```
chrisguiney, line 144: The carry is correct, and the ordering it needs holds — `BuildCritical()` at `mithril/sync.go:1273` precedes this call at `mithril/sync.go:1294`, and the earlier `updateMithrilReadyState` at `mithril/sync.go:1209` passes `clearSyncState=false`, so it never reaches the clear.

Worth naming what it makes visible, non-blocking and not this change's to fix. `ClearSyncState` is an unqualified `DELETE FROM sync_state`, so surviving it is an opt-in list: the three rows re-written here, plus `mithril_immutable_max` written after the call returns. A durable key nobody adds to that list is destroyed silently. #3736 is that failure already realized for the forging fence, and several other `sync_state` keys are written outside this package — `immutable_utxo_offsets_tip`, `synthetic_v2_cost_model`, `synthetic_v2_cost_model_cleared_epoch`, `delegator_inactivity_activated`, `database_lifecycle_truncate_pending`, `token_registry_etag`.

Adding a fourth carry here is the right minimal move for this change. Qualifying the clear by the ephemeral keys it actually means to remove — `sync_status`, `mithril_pinned_artifact`, `mithril_catchup_active`, the backfill checkpoint — would close the class rather than leave it enumerable, and would subsume the first of #3736's two candidate fixes.

chadaibbhot-git: Confirmed on all three points, and thank you for writing the ordering down — it is the part that is easy to get wrong later.

**Ordering.** Verified against the current head:

- `mithril/sync.go:1209` — `updateMithrilReadyState(..., syncStatusBackfill, false)`; `clearSyncState=false`, so it takes the `else if syncStatus != ""` branch at `mithril/sync_import.go:166` and never reaches the clear.
- `mithril/sync.go:1273` — `deferredIndexes.BuildCritical()`.
- `mithril/sync.go:1294` — `updateMithrilReadyState(..., "", true)`; the only call with `clearSyncState=true`.

So the marker is written by `BuildCritical` and read back inside the same clear, in that order, on the one path that clears. The carry is load-bearing and correctly placed.

**Unqualified clear.** Also confirmed. `Database.ClearSyncState` (`database/network_state.go:150`) delegates to `Store.ClearSyncState` (`database/plugin/metadata/sqlstore/operational.go:274`), which runs the sqlc query defined at `database/plugin/metadata/sqlstore/queries/sqlite/operational.sql:45-46`:

```sql
-- name: ClearSyncState :exec
DELETE FROM sync_state;
```

generated at `database/plugin/metadata/sqlstore/internal/query/sqlite/operational.sql.go:62-68`. No predicate. Surviving it is opt-in, exactly as you describe: the three rows re-written inside the transaction, plus `mithril_immutable_max` written after the call returns (`mithril/sync.go:1303-1316` → `mithril/import_marker.go:34`), which the comment at `mithril/artifact_pin.go:33` already flags as a deliberate ordering trick rather than a property of the clear.

Your list of durable keys written outside this package checks out too — `immutable_utxo_offsets_tip` (`internal/node/load.go:52`), `synthetic_v2_cost_model` and `synthetic_v2_cost_model_cleared_epoch` (`database/synthetic_cost_model.go:30,41`), `delegator_inactivity_activated` (`database/account_expiry_truncate.go:33`), `database_lifecycle_truncate_pending` (`database/lifecycle/truncate.go:42`), `token_registry_etag` (`internal/offchainmetadata/tokenregistry_sync.go:46`). None of them is on the carry list, and none of their owners has any reason to know this clear exists.

One small correction that helps the shape of the fix: the backfill checkpoint is not a `sync_state` row. It lives in its own `backfill_checkpoint` table (`database/plugin/metadata/sqlstore/internal/query/sqlite/operational.sql.go:1470,4554`), and `internal/node/backfill.go` touches no sync state at all. So the ephemeral set this clear actually means to remove is just `sync_status`, `mithril_pinned_artifact` (`mithril/artifact_pin.go:45`) and `mithril_catchup_active` (`mithril/import_marker.go:109`) — three keys, all owned by this package. A qualified `DELETE ... WHERE sync_key IN (...)` over those three is smaller than the carry list it would replace, which is a good sign for the inversion being the right one.

**Scope.** Agreed it is not this change's to fix, and I am keeping the PR to the fourth carry. This change is about the rollback cascade index; inverting the clear from deny-list to allow-list changes the durability contract of every `sync_state` key in the tree and wants its own review and its own tests, not a rider on an index rebuild.

**Where it belongs.** #3736 is the right home, since as you say the qualified clear subsumes its first candidate fix (move the fence out of `sync_state`) — it gets the fence, `immutable_utxo_offsets_tip`, the synthetic cost-model pair, the inactivity flag, the pending-truncate flag and the token-registry etag in one move, with no schema migration. I checked before writing this: there is no open PR or branch implementing a qualified clear, and #3736 is the only open item describing the class, so I have not proposed opening anything separate. It is your issue and your call which of the three shapes it takes; I am happy to leave the note here as the pointer.

The comment block at `mithril/sync_import.go:133-143` currently explains why the carry exists but not that the enumeration is the underlying problem. If it is useful I can extend it to name the class and point at #3736, so the next person adding a durable key finds the warning at the clear rather than after losing a row — say the word and it goes in this PR; otherwise I will leave the diff as it is.

wolf31o2:
Merge in origin/main and resolve any remaining CI issues

PR #4061, opened 2026-09-07
fix(mithril): tolerate missing account in historical backfill

### LINKED ISSUE #3788: fix(mithril): preserve historical reward-account state during backfill
## Summary

Preprod API-mode Mithril backfill fails while replaying a valid historical transaction because Dingo can resolve a withdrawal against no active `account` row, even though the canonical chain requires that reward account to exist historically.

## Evidence

- Network: Preprod
- Failing block slot: `63195287`
- Transaction: `8604ffe5b4fada838d3780f14e6275a31604674798d849778b4632a4cf7aa160`
- Failure: `set transaction metadata ... account not found`
- The canonical transaction contains four withdrawals; two credentials are absent from the imported snapshot account table.
- A current released build reproduced the same historical API-backfill failure class on a production-sized chain.

## Root-cause requirement

Historical API backfill must reconstruct every canonical account needed by replay before applying withdrawals. A missing account is an invariant failure, not a valid historical state to silently accept. The fix must identify why account import or certificate replay leaves the row absent/inactive and repair that ordering/state reconstruction. Live ledger validation must remain strict.

## Containment

PR #4061 was merged as a stop-gap. It prevents the backfill from aborting and journals the withdrawal, but it does not prove or restore the missing account invariant. It must not be considered the root-cause fix.

## Acceptance criteria

- Fresh API-mode Mithril backfill reconstructs the required account before replaying its withdrawal.
- Historical account state is complete and consistent with canonical certificates and withdrawals.
- No missing-account fallback is needed for a valid canonical withdrawal.
- Live ledger withdrawal validation remains unchanged.
- Regression coverage demonstrates the original failure and the repaired account reconstruction/order.

Related: #3616, #3686, #2360.

cryptodj413:
## Summary

`applyTransactionWithdrawals` required an active `account` row for every
withdrawal's stake credential, including API-mode Mithril historical backfill.
A canonical withdrawal replayed from before the imported snapshot can name a
credential that is absent or inactive at the snapshot boundary, aborting the
backfill with `account not found`.

Historical backfill now records the `account_reward_delta` journal row instead
of failing when the credential has no active account, without creating or
reactivating an `account` row. Live ingestion (`historicalBackfill = false`)
still requires an active account.

The two historical cases remain distinct:

- No account row exists: journal `previous_reward = 0`.
- An inactive row exists: preserve its real reward as `previous_reward`.

Fixes #3788

## Validation

- Rebased onto current `origin/main` and preserved the existing branch history
  with a merge commit.
- `go test ./database/plugin/metadata/sqlite -run '^TestSharedSQLStoreHistoricalBackfillWithdrawalMissingAccount$' -count=1`
- `go test -race ./database/plugin/metadata/sqlite -run '^TestSharedSQLStoreHistoricalBackfillWithdrawalMissingAccount$' -count=1`
- `go test ./database/plugin/metadata/sqlstore` passed.

The full metadata-store race suite reached its existing ten-minute test timeout
in the large committee-pruning scale test; the focused regression passed under
the race detector.

## Documentation

`DATABASE.md` documents the historical-backfill withdrawal cases and their
journal behavior.

### DIFF
--- a/database/plugin/metadata/sqlstore/transaction_write.go
+++ b/database/plugin/metadata/sqlstore/transaction_write.go
@@ -13,6 +13,7 @@ import (
 	"bytes"
 	"context"
 	"database/sql"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"math"
@@ -911,17 +912,36 @@ WHERE credential_tag = ? AND staking_key = ? AND active = TRUE`,
 			// merely-inactive row can still hold the credential's real
 			// balance; only fall back to an unknown (zero) previous balance
 			// when no row exists at all, rather than discarding a real one.
+			var fallbackActive sql.NullBool
 			err = db.QueryRowContext(ctx, `
-SELECT id, reward FROM account
+SELECT id, reward, active FROM account
 WHERE credential_tag = ? AND staking_key = ?`,
 				tag,
 				stakeKey.Bytes(),
-			).Scan(&accountID, &reward)
+			).Scan(&accountID, &reward, &fallbackActive)
+			rowExists := true
 			if errors.Is(err, sql.ErrNoRows) {
+				rowExists = false
 				reward = sql.NullString{}
 			} else if err != nil {
 				return err
 			}
+			// TEMPORARY DIAGNOSTIC for issue #3788: distinguish "no account
+			// row exists at all" from "a row exists but is inactive" at the
+			// exact moment a historical-backfill withdrawal can't resolve an
+			// active account. Remove once the real cause is confirmed.
+			s.logger.Warn(
+				"historical backfill withdrawal found no active account",
+				"component", "backfill",
+				"credential_tag", tag,
+				"staking_key", hex.EncodeToString(stakeKey.Bytes()),
+				"tx_hash", hex.EncodeToString(txHash),
+				"slot", slot,
+				"amount", amount.String(),
+				"account_row_exists", rowExists,
+				"account_active", fallbackActive.Bool,
+				"account_reward", reward.String,
+			)
 		} else if err != nil {
 			return err
 		}
--- a/database/plugin/metadata/sqlstore/transaction_write.go
+++ b/database/plugin/metadata/sqlstore/transaction_write.go
@@ -897,13 +897,31 @@ WHERE credential_tag = ? AND staking_key = ? AND active = TRUE`,
 			if !historicalBackfill {
 				return models.ErrAccountNotFound
 			}
-			// Historical API backfill can replay a withdrawal whose stake
-			// credential was valid on the canonical chain at the time but is
-			// absent from the imported Mithril snapshot's active accounts
-			// (e.g. deregistered before the snapshot was taken). Record the
-			// withdrawal history without fabricating or reactivating a
-			// current stake-registration account.
 			accountFound = false
+			// Historical API backfill can replay a withdrawal whose stake
+			// credential has no *active* account row for one of two reasons:
+			// no account row exists at all (deregistered before the imported
+			// Mithril snapshot, or never active in it -- issue #3788), or
+			// backfill's own certificate replay (applyTransactionCertificates
+			// runs unconditionally, historicalBackfill or not) has
+			// temporarily deactivated a row Mithril imported active, between
+			// a historical deregistration certificate and a later
+			// re-registration certificate for the same credential.
+			// Deregistration's account upsert never clears `reward`, so a
+			// merely-inactive row can still hold the credential's real
+			// balance; only fall back to an unknown (zero) previous balance
+			// when no row exists at all, rather than discarding a real one.
+			err = db.QueryRowContext(ctx, `
+SELECT id, reward FROM account
+WHERE credential_tag = ? AND staking_key = ?`,
+				tag,
+				stakeKey.Bytes(),
+			).Scan(&accountID, &reward)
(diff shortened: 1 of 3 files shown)


database/plugin/metadata/sqlstore/transaction_write.go
```diff
@@ -892,10 +892,37 @@ WHERE credential_tag = ? AND staking_key = ? AND active = TRUE`,
...
+			// balance; only fall back to an unknown (zero) previous balance
+			// when no row exists at all, rather than discarding a real one.
+			err = db.QueryRowContext(ctx, `
+SELECT id, reward FROM account
+WHERE credential_tag = ? AND staking_key = ?`,
+				tag,
+				stakeKey.Bytes(),
+			).Scan(&accountID, &reward)
+			if errors.Is(err, sql.ErrNoRows) {
+				reward = sql.NullString{}
```
chrisguiney, line 925: Non-blocking, recording what I verified so the invariant is written down somewhere.

During backfill `previous_reward` is the *snapshot-boundary* balance in both branches, not the balance at the replayed slot. The active-row path already had that property before this PR, and the inactive-row fallback now matches it instead of substituting `0` — so this is consistent, not a new semantic.

I confirmed the fabricated `0` in the no-row branch is currently unobservable:

- `historicalRewardsBatch` selects withdrawal deltas with `added_slot >= boundarySlot` (or `> slot` when no boundary is recorded), and every backfill delta is below the Mithril boundary, so `ret[ref] = withdrawal.previous - beforeWithdrawal[ref]` never reads one.
- `DeleteAccountRewardsAfterSlot` writes `previous_reward` directly into `account.reward` for withdrawal deltas, but it can only reach a sub-boundary delta via a truncate target that `lifecycle.Truncate` already rejects against `MithrilTrustBoundarySlotStrict`.

Worth remembering if either predicate ever widens: the value is persisted as the literal `"0"`, not `NULL`, so it is indistinguishable from a genuine zero balance at that point.

chrisguiney (dismissed):
Reviewed at head `0c689065` in an isolated worktree. No blocking findings.

## Verified

- **The tolerated case is legitimate, not a masked import defect.** Mithril snapshots carry only *active* accounts, so a canonical pre-snapshot withdrawal can name a credential deregistered before the snapshot was taken. That matches #3788's Preprod evidence (slot 63195287, two credentials absent from the imported table). Recording the journal row without creating or reactivating an `account` row satisfies #3788's acceptance criteria; live ingestion still returns `account not found`, and the test asserts it in both subtests.
- **The inactive-row mechanism the second commit relies on is real.** The account upsert in `transaction_certificates.go` (`ON CONFLICT (credential_tag, staking_key) DO UPDATE SET ...`) omits `reward`, so a deregistration clears `active` while leaving the balance intact, and `applyTransactionCertificates` is not gated on `historicalBackfill`. The transient inactive-with-balance state is therefore reachable during backfill's own certificate replay.
- **Fail-before proved by reverting in place, not by running against main.** Reverting the whole change fails both subtests with `account not found`. Reverting *only* the inactive-row fallback fails just the inactive subtest (`expected: "777" / actual: "0"`). Each commit carries its own regression coverage.
- **The fallback lookup is deterministic.** `idx_account_credential(credential_tag, staking_key)` is UNIQUE, so the un-`LIMIT`ed second query can match at most one row.
- **Class audit complete.** `applyTransactionWithdrawals` is the only `ErrAccountNotFound` site in the historical-backfill transaction path. The other three (`AddAccountRewardByCredential`, `AddPostSnapshotAccountRewardByCredential`, `ApplyAccountRewardWithdrawal`) are reached from `ledger/reward_calculation.go` and `ledger/governance/enact.go`, not from `internal/node.Backfill`.
- **`previous_reward = 0` cannot clobber a live balance.** `DeleteAccountRewardsAfterSlot` writes `previous_reward` straight back into `account.reward` for withdrawal deltas, but every backfill delta sits below the Mithril trust boundary and `database/lifecycle/truncate.go` rejects any truncate target below `MithrilTrustBoundarySlotStrict`. `historicalRewardsBatch` excludes them too (`added_slot >= boundarySlot`). See the inline note.

## Checks at this head

gouroboros resolved in-worktree: `v0.202.10`.

| Command | Exit | Result |
| --- | --- | --- |
| `gofmt -l .` | 0 | 3 files, all pre-existing on `main` (`bark/blob_test.go`, `chain/chain_test.go`, `ledger/read_chain_pipeline_test.go`); none from this PR |
| `go test -race ./database/...` | 0 | 19 packages ok |
| `make docs-parity` | 0 | ok |
| `make import-boundaries` | 0 | ok |
| merged with current `origin/main` | 0 | build, vet, lint (0 issues) and `./database/...` race tests all pass; the branch is behind `main`, so this was checked explicitly |

Pipeline on `0c689065`: 13/13 check runs completed, all `success`.

## Not run

- Preview/Preprod Koios parity replay, Mithril network backfill, devnet, conformance and Antithesis — infrastructure unavailable here. Unit tests do not settle real-chain reward parity.
- `nilaway` and `staticcheck` are not installed locally, so `make lint` was not usable; `golangci-lint run ./...` was run directly instead.

## Bot coverage

By `commit_id` on the review records: cubic reviewed `7171e4c2` (the first commit), **not** this head. CodeRabbit has posted no review record on this PR at all. This head had no prior review of any kind.

wolf31o2 (approved):

Issue #4055, opened 2026-09-07
ledger: DRep deregistration below the Mithril anchor is rejected with expected refund 0, pinning the chain past k


wolf31o2:
## Summary

A Mithril-bootstrapped node rejects a DRep deregistration for any DRep that
registered below its trust anchor, because the import does not carry that DRep.
The refund check then computes an expected refund of 0 against a certificate
that correctly claims the 500 ADA deposit, the node rewinds to re-derive, and
the rewind passes k, so the chain pins and block production stops.

Seen on preview at epoch 1413 on a producer bootstrapped at anchor 122086875.
It forged 4 blocks, then wedged permanently when it reached block slot 122110212.

## Observed

```
level=WARN msg="TX 2a9b80bbbf133419f8b87d97e97f88815a2d66c2a701d5438482ea5647927e30 failed validation:
  conway utxo validation rule 50: incorrect refund for certificate type 17:
  supplied 500000000, expected 0"
level=WARN msg="validation failure after reaching tip, rewinding primary chain" component=ledger
  tx_hash=2a9b80bbbf133419f8b87d97e97f88815a2d66c2a701d5438482ea5647927e30
level=WARN msg="rejecting rollback that exceeds security parameter K" fork_depth=526 security_param=432 rollback_slot=122097057
```

Certificate type 17 is DRep deregistration. The certificate in the transaction
body is

```
[17, [0, h'd5e7f9dc7fdb3884523eac045657724abd6f2bf28a648d6cba7c38a8'], 500000000]
```

The node oscillates: it reaches tip, fails on this transaction, rewinds, and the
rollback needed to re-follow the chain exceeds k. The applied tip moves backward
(122110166 then 122110065) while the gap grows, and it recurs identically across
a restart, so it is deterministic rather than a transient. A restart re-rolls to
chain tip 122097167 and wedges again on the same block.

Alongside it the same window logs the input-conservation shape, which looks like
a consequence of the repeated rewinds rather than a separate cause:

```
level=ERROR msg="continuation block spends an input with no producer on the local applied chain"
conway utxo validation rule 30: bad input(s): ...
conway utxo validation rule 32: failed to resolve input ...: utxo not found
conway utxo validation rule 33: value not conserved: consumed 0, produced ...
level=WARN msg="event delivery stalled: subscriber not draining" type=ledger.blockfetch buffer=8 blocked_publishers=1
```

## The DRep is absent from the imported state

The bootstrap reports importing DReps and their deposits:

[...]

PR #4064, opened 2026-09-07
fix(ledger): fall back to the current DRep deposit when none is recorded

chadaibbhot-git:
## Problem

#4035 wired the recorded registration deposit into `LedgerView.DRepRegistration` / `DRepRegistrations`, and the gouroboros v0.204.0 bump it carried made `common.DRepRegistration.Deposit` a `*uint64` so absence is distinct from a recorded zero.

Absence is not inert on that path. `conway.UtxoValidateCertificateDeposits` fails closed on it:

```go
// gouroboros v0.204.0, ledger/conway/rules.go
if registration.Deposit == nil {
    return DRepDepositStateInconsistentError{Credential: c.DrepCredential}
}
```

```
registered DRep credential has no recorded deposit: type 0 hash <...>
```

That is the correct default for gouroboros — the reference ledger's `DRepState` carries a non-optional `drepDeposit`, so a registration without one is internally inconsistent and there is no refund to compare against. But it means a DRep that reaches this state deterministically rejects its own deregistration block, and a node that reaches that block stops making progress.

**Dingo can produce that state.** `ledger/governance/processing.go` calls `InsertDrepIfAbsent(..., active = true, ...)` on the vote-replay recovery path when a valid DRep vote proves the credential exists on-chain but the metadata row was lost during recovery or bootstrap. That writes a row in `drep` and **no `registration_drep` row at all**, so `GetDrepLastRegistrationDeposit` returns nil. The DRep is active, is deregisterable, and has no recorded deposit.

This is the follow-up @chrisguiney asked for in his review of #4035, which dispositioned the three-states-collapsed-to-one issue as non-blocking there and named this exact path:
https://github.com/blinklabs-io/dingo/pull/4035#discussion_r3945923931

> `ledger/governance/processing.go:306` calls `InsertDrepIfAbsent(..., active=true, ...)` on the vote-replay recovery path, and that writes only the `drep` table. It leaves an **active** DRep with no `registration_drep` row [...] Not a blocker [...] Better as a follow-up issue than as growth in this PR.

The store side landed in #4035 already returns `*uint64`, so this is purely the view's answer for the nil case.

## Change

`ledger/view.go` only, +59/-2. A recorded deposit is passed through unchanged; when nothing is recorded, the view reports the DRep deposit under the current era and protocol parameters instead of the absence.

That value is not invented. `currentDRepDeposit` runs the current era's `CertDepositFunc` over a `RegistrationDrepCertificate` — the same function `ledger.calculateCertificateDeposit` runs on the certificate write path — so the fallback equals what a registration applied now would have recorded. Era and parameters are read from a single published consensus snapshot so they cannot be torn apart.

Two things it deliberately does not do:

- **A recorded zero stays authoritative.** A zero `dRepDeposit` is a legitimate configuration; folding zero into the unrecorded case would refund the current parameter and reject a valid deregistration. This is the same distinction `LedgerView.StakeCredentialDeposit` keeps for the stake analogue (#3829).
- **Absence is still reported when the current era charges no DRep deposit** (pre-Conway, or parameters that do not belong to the era). gouroboros fails closed there rather than being handed an invented zero refund. Whether an era has a DRep deposit is decided by asking the parameters for one (`drepDepositParams`, the same capability `conway.UtxoValidateCertificateDeposits` asserts), not by inferring it from what the era's `CertDepositFunc` returns: `CertDepositShelley` through `CertDepositBabbage` have no `*RegistrationDrepCertificate` case and fall through to `default: return 0, nil`, so inference yielded a fabricated zero. Fixed in 1174ea92 after @chrisguiney caught it in review; the original revision of this description claimed the guard without it existing.

## Tests

`ledger/drep_deposit_unknown_fallback_test.go`, reusing #4035's existing fixtures (`drepRefundTestPparams`, `drepDeregistrationTx`, `seedImportedDrep`). Assertions are on the validation outcome through `conway.UtxoValidateCertificateDeposits` and `eras.ValidateTxConway` with a real `*LedgerView` over a real database, not on the helper's return value.

The unrecorded case is seeded through `InsertDrepIfAbsent` itself rather than by hand, and the helper asserts that path really does leave no registration row — if recovery ever starts writing one, the test says so instead of silently measuring nothing.

- `TestDrepDeregistrationFallsBackToCurrentDepositWhenUnrecorded` — the regression
- `TestDrepDeregistrationRejectsWrongRefundWhenUnrecorded` — the fallback is not a licence to accept any refund (0, +1, and a negative amount all still rejected)
- `TestDrepDeregistrationPrefersRecordedDepositOverFallback` — recorded and parameter values are deliberately different, so the two answers give opposite outcomes
- `TestDrepDeregistrationKeepsRecordedZeroAuthoritative`
- `TestDrepRegistrationsFallBackForUnrecordedDeposit` — the plural view has its own absence path
- `TestDrepRegistrationReportsAbsenceInAPreConwayEra` — the era guard, on a *published* Babbage snapshot, first asserting that era's deposit function really does answer zero-without-error
- `TestDrepRegistrationReportsAbsenceWithoutAPublishedSnapshot` — the other absence branch, asserting the fixture leaves the snapshot unpublished

Each is load-bearing. Removing the fallback fails 3; ignoring the recorded value fails 5 (including two of #4035's own); removing the `drepDepositParams` guard fails `TestDrepRegistrationReportsAbsenceInAPreConwayEra` and leaves the unpublished-snapshot test green, which is the discrimination the earlier single test did not have.

The substitution is also logged at WARN with the credential and the value used, so a node serving a fabricated deposit is distinguishable from one serving a recorded one.

## Verification

Rebased on `main` at c6488d37 (#4035).

| check | result |
|---|---|
| `go build ./...`, `go build -tags dingo_extra_plugins ./cmd/dingo` | pass |
| `go vet ./...` | pass |
| `golangci-lint run ./ledger/...` | 0 issues |
| `gofmt -l ledger/` | empty |
| `go test ./ledger/...` | pass, all 9 packages (`ledger` 102.5s) |
| `go test -race ./ledger/` | pass, 405.2s |

Related: #4035, #4055, #3829

### DIFF
--- a/ledger/drep_deposit_unknown_fallback_test.go
+++ b/ledger/drep_deposit_unknown_fallback_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/blinklabs-io/dingo/database"
 	"github.com/blinklabs-io/dingo/database/models"
 	"github.com/blinklabs-io/dingo/ledger/eras"
+	"github.com/blinklabs-io/gouroboros/ledger/babbage"
 	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
 	"github.com/blinklabs-io/gouroboros/ledger/conway"
 	"github.com/stretchr/testify/require"
@@ -249,14 +250,33 @@ func TestDrepRegistrationsFallBackForUnrecordedDeposit(t *testing.T) {
 	require.Equal(t, uint64(drepRefundTestRecordedDeposit), *got)
 }
 
-// TestDrepRegistrationReportsAbsenceWithoutADrepDeposit keeps the fallback
-// from inventing a refund where the current era charges no DRep deposit. The
-// view must still report absence there and let gouroboros fail closed rather
-// than accept a zero refund.
-func TestDrepRegistrationReportsAbsenceWithoutADrepDeposit(t *testing.T) {
-	// newStakeRefundTestView leaves the era Shelley with no published
-	// snapshot, which is exactly the no-DRep-deposit case.
-	lv, db := newStakeRefundTestView(t)
+// TestDrepRegistrationReportsAbsenceInAPreConwayEra keeps the fallback from
+// inventing a refund where the current era charges no DRep deposit.
+//
+// The era has to be published for this to mean anything. CertDepositShelley
+// through CertDepositBabbage have no *RegistrationDrepCertificate case and
+// fall through to "default: return 0, nil", so before the drepDepositParams
+// guard this reported a non-nil zero and gouroboros accepted a zero refund
+// instead of failing closed.
+func TestDrepRegistrationReportsAbsenceInAPreConwayEra(t *testing.T) {
+	ls, db := newRewardCalculationTestLedger(t)
+	ls.currentEra = eras.BabbageEraDesc
+	ls.currentPParams = &babbage.BabbageProtocolParameters{
+		KeyDeposit: 2_000_000,
+		MaxTxSize:  16_384,
+	}
+	ls.publishSnapshotsLocked()
+	lv := &LedgerView{ls: ls}
+	// The era's own deposit function really does answer zero-without-error
+	// for a DRep registration, which is what makes the guard load-bearing
+	// rather than defensive.
+	deposit, err := eras.BabbageEraDesc.CertDepositFunc(
+		&lcommon.RegistrationDrepCertificate{},
+		ls.currentPParams,
+	)
+	require.NoError(t, err)
+	require.Equal(t, uint64(0), deposit)
+
 	cred := drepRefundTestCredential(0xe7)
 	seedActiveDrepWithoutRegistration(t, db, cred, 100)
 
@@ -266,8 +286,49 @@ func TestDrepRegistrationReportsAbsenceWithoutADrepDeposit(t *testing.T) {
 	require.Nil(
 		t,
 		reg.Deposit,
-		"with no DRep deposit to fall back to, absence must be reported",
+		"a pre-Conway era has no DRep deposit to fall back to; absence must be reported",
 	)
+
+	registrations, err := lv.DRepRegistrations()
+	require.NoError(t, err)
+	require.Len(t, registrations, 1)
+	require.Nil(
+		t,
+		registrations[0].Deposit,
+		"the plural view must report absence in a pre-Conway era too",
+	)
+
+	require.ErrorContains(
+		t,
+		conway.UtxoValidateCertificateDeposits(
+			drepDeregistrationTx(cred, drepRefundTestPparamDeposit),
+			200,
+			lv,
+			drepRefundTestPparams(),
+		),
+		inconsistentDepositSubstring,
+	)
+}
+
+// TestDrepRegistrationReportsAbsenceWithoutAPublishedSnapshot covers the other
+// way the fallback has nothing to report: no consensus snapshot has been
+// published yet, so there are no current parameters to read. Kept separate
+// from the pre-Conway case because newStakeRefundTestView reaches this branch
+// and never the era one -- the era field it sets is never read.
+func TestDrepRegistrationReportsAbsenceWithoutAPublishedSnapshot(t *testing.T) {
+	lv, db := newStakeRefundTestView(t)
+	require.Nil(
+		t,
+		lv.ls.loadConsensusSnapshot(),
+		"this fixture must leave the snapshot unpublished for this test to mean what it says",
+	)
+	cred := drepRefundTestCredential(0xe8)
+	seedActiveDrepWithoutRegistration(t, db, cred, 100)
+
+	reg, err := lv.DRepRegistration(cred.Credential)
+	require.NoError(t, err)
+	require.NotNil(t, reg)
+	require.Nil(t, reg.Deposit)
 
 	require.ErrorContains(
 		t,
--- a/ledger/drep_deposit_unknown_fallback_test.go
+++ b/ledger/drep_deposit_unknown_fallback_test.go
@@ -1,0 +1,282 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package ledger
+
+import (
+	"testing"
+
+	"github.com/blinklabs-io/dingo/database"
+	"github.com/blinklabs-io/dingo/database/models"
+	"github.com/blinklabs-io/dingo/ledger/eras"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/conway"
+	"github.com/stretchr/testify/require"
+)
+
+// inconsistentDepositSubstring is the message
+// conway.DRepDepositStateInconsistentError renders when a registration
+// reports no recorded deposit. Matched on the message rather than on a rule
+// index, which moves whenever gouroboros inserts or reorders a rule.
+const inconsistentDepositSubstring = "registered DRep credential has no recorded deposit"
+
+// newDrepFallbackTestView returns a *LedgerView whose published consensus
+// snapshot is Conway with drepRefundTestPparams, so the unknown-deposit
+// fallback resolves through the same era certificate-deposit function the
+// certificate write path uses. newStakeRefundTestView leaves the snapshot
+// unpublished and the era Shelley, which is the "current era charges no DRep
+// deposit" case rather than this one.
+func newDrepFallbackTestView(
+	t *testing.T,
+) (*LedgerView, *database.Database) {
+	t.Helper()
+	ls, db := newRewardCalculationTestLedger(t)
+	ls.currentEra = eras.ConwayEraDesc
+	ls.currentPParams = drepRefundTestPparams()
+	ls.publishSnapshotsLocked()
+	return &LedgerView{ls: ls}, db
+}
+
+// seedActiveDrepWithoutRegistration reproduces the state the vote-replay
+// recovery path leaves behind. ledger/governance/processing.go calls
+// InsertDrepIfAbsent when a valid DRep vote proves the credential exists
+// on-chain but the metadata row was lost during recovery or bootstrap; that
+// writes an active drep row and no registration_drep row at all, so the
+// credential is registered with no recorded deposit.
+func seedActiveDrepWithoutRegistration(
+	t *testing.T,
+	db *database.Database,
+	cred lcommon.Credential,
+	slot uint64,
+) {
+	t.Helper()
+	tag, err := models.CredentialTagFromUint(uint(cred.CredType))
+	require.NoError(t, err)
+	require.NoError(t, db.InsertDrepIfAbsent(
+		tag,
+		cred.Credential[:],
+		slot,
+		"",
+		nil,
+		true,
+		nil,
+	))
+	// The recovery path really does leave no registration row behind; if it
+	// ever starts writing one this test is measuring the wrong thing.
+	recorded, err := db.GetDrepLastRegistrationDeposit(
+		tag,
+		cred.Credential[:],
+		nil,
+	)
+	require.NoError(t, err)
+	require.Nil(
+		t,
+		recorded,
+		"the recovery path must leave no recorded deposit for this test to exercise the fallback",
+	)
+}
+
+// TestDrepDeregistrationFallsBackToCurrentDepositWhenUnrecorded is the
+// regression test. An active DRep with no registration_drep row reported
+// Deposit == nil, and gouroboros fails closed on that
+// (DRepDepositStateInconsistentError), so the deregistration was rejected and
+// a node reaching that block stopped making progress. The refund is now
+// judged against the DRep deposit the current protocol parameters charge,
+// which is what the certificate write path would have recorded.
+func TestDrepDeregistrationFallsBackToCurrentDepositWhenUnrecorded(
+	t *testing.T,
+) {
+	lv, db := newDrepFallbackTestView(t)
+	cred := drepRefundTestCredential(0xe1)
+	seedActiveDrepWithoutRegistration(t, db, cred, 100)
+
+	pp := drepRefundTestPparams()
+	tx := drepDeregistrationTx(cred, drepRefundTestPparamDeposit)
+	// The rule is invoked directly so a nil error means the refund
+	// comparison actually ran and matched, rather than that the substring
+	// was absent because an unrelated rule failed first.
+	require.NoError(
+		t,
+		conway.UtxoValidateCertificateDeposits(tx, 200, lv, pp),
+	)
+	if err := eras.ValidateTxConway(tx, 200, lv, pp); err != nil {
+		require.NotContains(t, err.Error(), inconsistentDepositSubstring)
+		require.NotContains(t, err.Error(), incorrectRefundSubstring)
+	}
+
+	// Supporting evidence for why the acceptance holds.
+	reg, err := lv.DRepRegistration(cred.Credential)
+	require.NoError(t, err)
+	require.NotNil(t, reg)
+	require.NotNil(
+		t,
+		reg.Deposit,
+		"an unrecorded deposit must be reported as the current parameter, not as absence",
+	)
+	require.Equal(t, uint64(drepRefundTestPparamDeposit), *reg.Deposit)
+}
+
+// TestDrepDeregistrationRejectsWrongRefundWhenUnrecorded is the mandatory
+// negative case: the fallback must not become a licence to accept any refund.
+// A deregistration over the same unrecorded registration that supplies a
+// different amount is still rejected, including the zero the absence would
+// have been misread as.
+func TestDrepDeregistrationRejectsWrongRefundWhenUnrecorded(t *testing.T) {
+	lv, db := newDrepFallbackTestView(t)
+	cred := drepRefundTestCredential(0xe2)
+	seedActiveDrepWithoutRegistration(t, db, cred, 100)
+
+	pp := drepRefundTestPparams()
+	for _, refund := range []int64{0, drepRefundTestPparamDeposit + 1, -1} {
+		err := conway.UtxoValidateCertificateDeposits(
+			drepDeregistrationTx(cred, refund),
+			200,
+			lv,
+			pp,
+		)
+		require.ErrorContains(t, err, incorrectRefundSubstring)
+	}
+}
+
+// TestDrepDeregistrationPrefersRecordedDepositOverFallback is the second
+// mandatory negative case. The recorded deposit and the current parameter are
+// deliberately different values, so the two possible answers give opposite
+// outcomes and the test cannot pass by accident: a recorded deposit must win,
+// and the fallback must not be reached.
+func TestDrepDeregistrationPrefersRecordedDepositOverFallback(t *testing.T) {
+	lv, db := newDrepFallbackTestView(t)
+	cred := drepRefundTestCredential(0xe3)
+	require.NotEqual(
+		t,
+		uint64(drepRefundTestPparamDeposit),
+		uint64(drepRefundTestRecordedDeposit),
+		"the recorded deposit must differ from the parameter to discriminate",
+	)
+	seedImportedDrep(t, db, cred, drepRefundTestRecordedDeposit, 100, true)
+
+	pp := drepRefundTestPparams()
+	require.NoError(t, conway.UtxoValidateCertificateDeposits(
+		drepDeregistrationTx(cred, drepRefundTestRecordedDeposit),
+		200,
+		lv,
+		pp,
+	))
+	require.ErrorContains(
+		t,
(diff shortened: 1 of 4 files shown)


chadaibbhot-git:
@coderabbitai review (final head 2be06b23)

ledger/view.go
```diff
@@ -1233,6 +1233,63 @@ func (lv *LedgerView) CommitteeMembers() ([]lcommon.CommitteeMember, error) {
...
+// snapshot so the era and the parameters cannot be torn apart. The second
+// return is false when the current era charges no DRep deposit (pre-Conway)
+// or the protocol parameters do not belong to that era.
+func (lv *LedgerView) currentDRepDeposit() (uint64, bool) {
+	snapshot := lv.ls.loadConsensusSnapshot()
+	if snapshot == nil || snapshot.currentPParams == nil ||
+		snapshot.currentEra.CertDepositFunc == nil {
+		return 0, false
+	}
+	deposit, err := snapshot.currentEra.CertDepositFunc(
```
chrisguiney, line 1318: Blocker. `CertDepositShelley`, `CertDepositAllegra`, `CertDepositMary`, `CertDepositAlonzo` and `CertDepositBabbage` have no `*RegistrationDrepCertificate` case, so a DRep registration certificate falls through to `default: return 0, nil`. The type assertion above succeeds whenever the params match the era, so this call returns `(0, nil)`, `ok` is true, and `drepRegistrationDeposit` hands gouroboros a non-nil zero instead of the absence.

Reproduced against this head in a clean worktree: with `ls.currentEra = eras.BabbageEraDesc`, `ls.currentPParams = &babbage.BabbageProtocolParameters{...}` and `ls.publishSnapshotsLocked()`, a credential seeded through `InsertDrepIfAbsent` reports `Deposit = 0` from `DRepRegistration` and `0` from `DRepRegistrations`.

The only inputs that reach `ok == false` are an unpublished snapshot, nil params, a nil `CertDepositFunc` (Byron), and `ErrIncompatibleProtocolParams` from a params/era mismatch. The era itself is not one of them, so the doc comment two lines up -- "false when the current era charges no DRep deposit (pre-Conway)" -- and the same claim in the PR description do not hold.

Not reachable on a real chain today, since DRep rows only appear from Conway blocks. It is still a stated safety property that a later change will rely on. Guard the era explicitly, or assert the parameters type actually carries a DRep deposit, rather than inferring it from what the deposit function happens to return.

This confirms cubic's P2 on line 1280.

ledger/view.go
```diff
@@ -1233,6 +1233,63 @@ func (lv *LedgerView) CommitteeMembers() ([]lcommon.CommitteeMember, error) {
...
+// proves a credential exists but the metadata row was lost. That DRep is
+// active, can be deregistered, and has no recorded deposit.
+//
+// Reporting the current parameter is the same answer the write path would have
+// recorded: it runs the era's certificate-deposit function over the same
+// registration certificate type that ledger.calculateCertificateDeposit does,
+// so the value equals what a registration applied now would store. Absence is
+// still reported when the current era charges no DRep deposit, which leaves
+// gouroboros to fail closed rather than inventing a zero refund.
+func (lv *LedgerView) drepRegistrationDeposit(recorded *uint64) *uint64 {
```
chrisguiney: Non-blocking, on the substitution itself.

The recorded deposit is historical and this fallback is current, so the two agree only while `dRepDeposit` has not changed since the DRep registered. If it has changed, gouroboros compares the block's refund against the wrong expected value and the node still stops -- but as `CertificateRefundIncorrectError`, which reads as a bad block rather than as the local state corruption it actually is. That is a worse diagnostic than the `DRepDepositStateInconsistentError` it replaces, not a better one.

Nothing logs or counts the substitution either, so a node serving a fabricated deposit is indistinguishable from one serving a recorded deposit, including through the `DRepState` local-state-query reply that `DRepRegistrations` feeds. A log line or a counter where the fallback is taken would keep the liveness win without discarding the signal.

The root cause stays open either way: `ledger/governance/processing.go` calls `InsertDrepIfAbsent` with `point.Slot` and the era and parameters already in hand, so it could record the deposit at that slot. That is closer to what was actually paid than the tip's value, and it would make this fallback unreachable rather than load-bearing. Worth an issue if not this PR.

ledger/drep_deposit_unknown_fallback_test.go
```diff
@@ -0,0 +1,282 @@
...
+	require.Equal(t, uint64(drepRefundTestRecordedDeposit), *got)
+}
+
+// TestDrepRegistrationReportsAbsenceWithoutADrepDeposit keeps the fallback
+// from inventing a refund where the current era charges no DRep deposit. The
+// view must still report absence there and let gouroboros fail closed rather
+// than accept a zero refund.
+func TestDrepRegistrationReportsAbsenceWithoutADrepDeposit(t *testing.T) {
+	// newStakeRefundTestView leaves the era Shelley with no published
+	// snapshot, which is exactly the no-DRep-deposit case.
```
chrisguiney: This comment is wrong on both halves, and the test passes for a reason unrelated to the era.

`newStakeRefundTestView` calls `newRewardCalculationTestLedger`, which sets `currentEra: eras.ShelleyEraDesc` but never calls `publishSnapshotsLocked`. `loadConsensusSnapshot()` therefore returns nil and `currentDRepDeposit` exits at the `snapshot == nil` branch. The era value is never read. Publish the snapshot the way `newDrepFallbackTestView` does, with a pre-Conway era and matching parameters, and the view reports a non-nil zero instead -- see the comment on `ledger/view.go`.

So this is a guard for the unpublished-snapshot case, not for "the current era charges no DRep deposit", and it does not pin the property the PR description advertises. Once the era guard exists, seed a published pre-Conway snapshot here so the test exercises it.

chrisguiney (dismissed):
Reviewed at head `2be06b23` in an isolated worktree with an isolated lint cache.

## Blocker

The pre-Conway guard this change documents does not exist. `currentDRepDeposit` reaches `default: return 0, nil` in every pre-Conway `CertDeposit*`, so `ok` is true and `drepRegistrationDeposit` returns a pointer to zero rather than absence. Reproduced here: with `currentEra = eras.BabbageEraDesc`, matching Babbage parameters and a published snapshot, `DRepRegistration` reports `Deposit = 0` and `DRepRegistrations` reports `0`. `TestDrepRegistrationReportsAbsenceWithoutADrepDeposit` passes through the `snapshot == nil` branch — `newRewardCalculationTestLedger` never publishes one — so it does not pin the property its comment names. Details inline; this confirms cubic's P2.

## Test evidence

Reverting both call sites to the raw pointer and re-running the six new tests: `FallsBackToCurrentDepositWhenUnrecorded`, `RejectsWrongRefundWhenUnrecorded` and `RegistrationsFallBackForUnrecordedDeposit` fail; `PrefersRecordedDepositOverFallback`, `KeepsRecordedZeroAuthoritative` and `ReportsAbsenceWithoutADrepDeposit` pass either way. Three regression tests and three guards, matching the description's "removing the fallback fails 3".

## Non-blocking

- `internal/test/conformance/state_provider.go`'s `DingoStateProvider.DRepRegistrations` still reports absence for a credential with no registration row, and its comment says it mirrors `ledger.LedgerView.DRepRegistrations`. The two views now disagree. Harmless for the vectors, which always seed a registration, but the comment is stale.
- `LedgerView.StakeCredentialDeposit` in the same file keeps an unrecorded deposit nil and lets gouroboros fail closed. The DRep path now diverges from the stake analogue; the reason belongs in the code or an issue.
- #4039 covers the deregistration write path's NULL branch, which is a different path; it does not contradict this change.

## Checks

| check | exit | result |
|---|---|---|
| `gofmt -l .` | 0 | 3 files, all pre-existing on `origin/main` (`bark/blob_test.go`, `chain/chain_test.go`, `ledger/read_chain_pipeline_test.go`); none from this PR |
| `go test ./ledger/ -run 'TestDrep...'` (6 new tests) | 0 | pass at head; 3 fail with the fix reverted |
| CI on `2be06b23` | — | all 10 checks green |

## Not run

- Preview Koios parity replay, conformance vectors, devnet and Antithesis — infrastructure unavailable in this environment. A deposit change of this shape is the kind unit tests can pass while a replay diverges.
- `make lint` — `nilaway` and `staticcheck` are not installed here, so `golangci-lint run ./...` was run directly instead.
- Nested modules `examples/dingo-gov-lens` and `internal/test/antithesis` were not built.

ledger/view.go
```diff
@@ -1233,6 +1233,63 @@ func (lv *LedgerView) CommitteeMembers() ([]lcommon.CommitteeMember, error) {
...
+
+// currentDRepDeposit returns the DRep registration deposit under the current
+// era and protocol parameters, read from a single published consensus
+// snapshot so the era and the parameters cannot be torn apart. The second
+// return is false when the current era charges no DRep deposit (pre-Conway)
+// or the protocol parameters do not belong to that era.
+func (lv *LedgerView) currentDRepDeposit() (uint64, bool) {
+	snapshot := lv.ls.loadConsensusSnapshot()
+	if snapshot == nil || snapshot.currentPParams == nil ||
+		snapshot.currentEra.CertDepositFunc == nil {
```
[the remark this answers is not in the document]
chrisguiney: Valid, and reproduced against this head. With `ls.currentEra = eras.BabbageEraDesc`, matching `*babbage.BabbageProtocolParameters` and `ls.publishSnapshotsLocked()`, a credential seeded through `InsertDrepIfAbsent` reports `Deposit = 0` from both `DRepRegistration` and `DRepRegistrations`: the pre-Conway `CertDeposit*` functions have no `*RegistrationDrepCertificate` case and fall through to `default: return 0, nil`, so `ok` is true.

One correction to the framing: the behaviour is not reachable on a real chain, because DRep rows only appear from Conway blocks. What is broken today is the stated contract — the doc comment on `currentDRepDeposit` and the PR description both promise this guard, and `TestDrepRegistrationReportsAbsenceWithoutADrepDeposit` passes through the `snapshot == nil` branch rather than through the era, so nothing pins it. Carried into my review as a blocker on that basis.

wolf31o2, line 1312: Blocking: pre-Conway eras expose a `CertDepositFunc` that can return zero without error even though DRep deposits are unsupported. This path therefore returns a non-nil zero for an unrecorded DRep and can accept a zero-refund deregistration, contradicting the documented fail-closed absence behavior. Guard for a DRep-capable era before invoking `CertDepositFunc`, and add a pre-Conway regression assertion through the deregistration validation path.

wolf31o2 (changes requested):
The fallback must preserve absence in eras that do not charge DRep deposits.

ledger/view.go
chadaibbhot-git: Confirmed and fixed in 1174ea92. You are right on both the mechanism and the framing.

`CertDepositBabbage` (`ledger/eras/babbage.go:195-202`) switches only on `*PoolRegistrationCertificate` and `*StakeRegistrationCertificate` and falls through to `default: return 0, nil`; Shelley, Allegra, Mary and Alonzo are the same shape. The type assertion succeeds when the params match the era, so `(0, nil)` came back, `ok` was true, and a fabricated zero went to gouroboros. The doc comment and the PR description both promised a guard that was not there.

I took your second suggestion rather than an era allowlist:

```go
if _, ok := snapshot.currentPParams.(drepDepositParams); !ok {
    return 0, false
}

type drepDepositParams interface {
    DRepDepositAmount() *big.Int
}
```

That is the same capability `conway.UtxoValidateCertificateDeposits` asserts to obtain the deposit it compares against, and in v0.204.0 only `ConwayProtocolParameters` and `DijkstraProtocolParameters` implement it. So the guard admits exactly the eras whose parameters the consuming rule can read, and a future era gains it by implementing the accessor rather than by being added to a list here. The era's `CertDepositFunc` still computes the value, so the "same function the write path uses" property is unchanged.

The PR description has been corrected too — it made the same unbacked claim.

ledger/view.go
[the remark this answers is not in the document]
chadaibbhot-git: Confirmed and fixed in 1174ea92. You are right on both the mechanism and the framing.

`CertDepositBabbage` (`ledger/eras/babbage.go:195-202`) switches only on `*PoolRegistrationCertificate` and `*StakeRegistrationCertificate` and falls through to `default: return 0, nil`; Shelley, Allegra, Mary and Alonzo are the same shape. The type assertion succeeds when the params match the era, so `(0, nil)` came back, `ok` was true, and a fabricated zero went to gouroboros. The doc comment and the PR description both promised a guard that was not there.

I took your second suggestion rather than an era allowlist:

```go
if _, ok := snapshot.currentPParams.(drepDepositParams); !ok {
    return 0, false
}

type drepDepositParams interface {
    DRepDepositAmount() *big.Int
}
```

That is the same capability `conway.UtxoValidateCertificateDeposits` asserts to obtain the deposit it compares against, and in v0.204.0 only `ConwayProtocolParameters` and `DijkstraProtocolParameters` implement it. So the guard admits exactly the eras whose parameters the consuming rule can read, and a future era gains it by implementing the accessor rather than by being added to a list here. The era's `CertDepositFunc` still computes the value, so the "same function the write path uses" property is unchanged.

The PR description has been corrected too — it made the same unbacked claim.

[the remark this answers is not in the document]
chadaibbhot-git: Confirmed and fixed in 1174ea92. You are right on both the mechanism and the framing.

`CertDepositBabbage` (`ledger/eras/babbage.go:195-202`) switches only on `*PoolRegistrationCertificate` and `*StakeRegistrationCertificate` and falls through to `default: return 0, nil`; Shelley, Allegra, Mary and Alonzo are the same shape. The type assertion succeeds when the params match the era, so `(0, nil)` came back, `ok` was true, and a fabricated zero went to gouroboros. The doc comment and the PR description both promised a guard that was not there.

I took your second suggestion rather than an era allowlist:

```go
if _, ok := snapshot.currentPParams.(drepDepositParams); !ok {
    return 0, false
}

type drepDepositParams interface {
    DRepDepositAmount() *big.Int
}
```

That is the same capability `conway.UtxoValidateCertificateDeposits` asserts to obtain the deposit it compares against, and in v0.204.0 only `ConwayProtocolParameters` and `DijkstraProtocolParameters` implement it. So the guard admits exactly the eras whose parameters the consuming rule can read, and a future era gains it by implementing the accessor rather than by being added to a list here. The era's `CertDepositFunc` still computes the value, so the "same function the write path uses" property is unchanged.

The PR description has been corrected too — it made the same unbacked claim.

ledger/drep_deposit_unknown_fallback_test.go
chadaibbhot-git: Fixed in 1174ea92, and thank you for catching that the test proved nothing — it was passing through `snapshot == nil`, exactly as you describe, and the era field was never read.

Split into two, so each branch has its own test and neither can stand in for the other:

- `TestDrepRegistrationReportsAbsenceInAPreConwayEra` builds the view directly, sets `currentEra = eras.BabbageEraDesc` with matching `*babbage.BabbageProtocolParameters`, and calls `publishSnapshotsLocked()`. Before asserting anything about the view it asserts the premise — that `eras.BabbageEraDesc.CertDepositFunc` really does answer `(0, nil)` for a `*RegistrationDrepCertificate` — so if that ever stops being true the test says so instead of quietly passing for a new reason. It then requires `Deposit == nil` from both `DRepRegistration` and `DRepRegistrations`, and requires the rule to return `DRepDepositStateInconsistentError`.
- `TestDrepRegistrationReportsAbsenceWithoutAPublishedSnapshot` keeps the fixture-based case and now asserts `lv.ls.loadConsensusSnapshot()` is nil first, so it is pinned as the unpublished-snapshot branch rather than mislabelled as the era one.

Removing the `drepDepositParams` guard turns the first red and leaves the second green, which is the discrimination that was missing.

ledger/view.go
chadaibbhot-git: Taking the logging half now, leaving the root cause for its own change.

`drepRegistrationDeposit` emits a WARN wherever it substitutes, with the credential and the value used, so a node serving a fabricated deposit is no longer indistinguishable from one serving a recorded one — including through the `DRepState` local-state-query reply that `DRepRegistrations` feeds. The doc comment now states the tradeoff you describe rather than presenting the substitution as free: the recorded deposit is historical and this one is current, they agree only while `dRepDeposit` has not changed since the DRep registered, and if it has, `CertificateRefundIncorrectError` reads as a bad block rather than as the local state gap it is.

On the root cause, I agree it is the better fix and that it would make this fallback unreachable rather than load-bearing. `InsertDrepIfAbsent` has `point.Slot` and the era and parameters in hand at that call site, so it can record what was actually charged at that slot, which beats the tip's value. I have deliberately not done it here: it changes what the vote-replay recovery path writes, which is a different blast radius from a read-side fallback, and it wants its own tests around rollback and replay. Happy to open an issue for it, or to fold it in here if you would rather see them together — say which and I will.

Worth noting the two are complementary rather than alternatives: recording at recovery fixes new occurrences, and the fallback still has to answer for rows already written by an older binary.

ledger/view.go
chadaibbhot-git: This was written against 2be06b23 and is fixed in 1174ea92: `currentDRepDeposit` now gates on `drepDepositParams` (`DRepDepositAmount() *big.Int`, the same capability `conway.UtxoValidateCertificateDeposits` asserts, implemented only by Conway and Dijkstra) before invoking `CertDepositFunc`, and `TestDrepRegistrationReportsAbsenceInAPreConwayEra` is the pre-Conway regression — it publishes a Babbage snapshot, first asserts that era's deposit function really does answer zero-without-error for a DRep registration, then requires absence from both views and `DRepDepositStateInconsistentError` through `conway.UtxoValidateCertificateDeposits`.

ledger/view.go
```diff
@@ -1233,6 +1233,106 @@ func (lv *LedgerView) CommitteeMembers() ([]lcommon.CommitteeMember, error) {
...
+// compares against, so this guard admits exactly the eras whose parameters
+// that rule can read, and a future era gains it by implementing the accessor
+// rather than by being added to a list here.
+func (lv *LedgerView) currentDRepDeposit() (uint64, bool) {
+	snapshot := lv.ls.loadConsensusSnapshot()
+	if snapshot == nil || snapshot.currentPParams == nil ||
+		snapshot.currentEra.CertDepositFunc == nil {
+		return 0, false
+	}
+	if _, ok := snapshot.currentPParams.(drepDepositParams); !ok {
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid, and fixed in a2e104c3 — thank you, this was a real panic path.

I put the guard in the era helper rather than at the call site. `CertDepositDijkstra` asserted `*DijkstraProtocolParameters` and tested only `ok`, so a typed nil fell through to the switch and every case dereferenced it. `CertDepositConway` has always spelled it `if !ok || tmpPparams == nil`; Dijkstra had not, and that asymmetry is the defect. Fixing it there covers every caller of the helper rather than only this one, and it now returns `ErrIncompatibleProtocolParams` like its Conway sibling — which `currentDRepDeposit` already turns into absence, so gouroboros fails closed exactly as it does for every other no-deposit-available case.

The `drepDepositParams` check at the call site stays, as you'd expect: a typed nil satisfies it, so it was never the thing that could stop this. The two are complementary — the era helper stops the panic, the capability check stops a pre-Conway era reaching the helper at all.

Two tests, both of which go red if the nil check is dropped:

- `TestCertDepositRejectsTypedNilParams` (`ledger/eras`) asserts both era helpers over drep/pool/stake registration certificates, so the pair cannot drift apart again.
- `TestDrepRegistrationReportsAbsenceForTypedNilParams` (`ledger`) publishes a typed-nil Dijkstra snapshot and requires `DRepRegistration` and `DRepRegistrations` to report absence without panicking. It first asserts the typed nil really does satisfy `drepDepositParams`, so the case is proven to reach the era helper rather than stopping early.

wolf31o2 (dismissed):

chadaibbhot-git:
Status of the review flags, at head 81b4da10:

- **@wolf31o2's blocker** (pre-Conway eras must keep absence): fixed in 1174ea92. `currentDRepDeposit` now returns absence unless the current parameters implement `DRepDepositAmount()` (Conway and Dijkstra only), before `CertDepositFunc` is called. `TestDrepRegistrationReportsAbsenceInAPreConwayEra` publishes a Babbage snapshot and requires `Deposit == nil` from both views, plus `DRepDepositStateInconsistentError` through `conway.UtxoValidateCertificateDeposits`. See the [thread reply](https://github.com/blinklabs-io/dingo/pull/4064#discussion_r3952169600). Your approval at a2e104c3 on 09-12 was dismissed automatically when main was merged in on 09-17 (608bb7f2).
- **@chrisguiney's blocker** (the same mechanism), plus the test that passed through `snapshot == nil`: fixed in the same commit. The test is split so that the era branch and the unpublished-snapshot branch each have their own test, and removing the guard turns only the era test red ([reply](https://github.com/blinklabs-io/dingo/pull/4064#discussion_r3951921691), [reply](https://github.com/blinklabs-io/dingo/pull/4064#discussion_r3951921901)).
- **Typed-nil Dijkstra parameters** (cubic P1): fixed in a2e104c3, inside `CertDepositDijkstra` itself, with `TestCertDepositRejectsTypedNilParams` and `TestDrepRegistrationReportsAbsenceForTypedNilParams`.
- **Non-blocking:** a substituted deposit is now logged at WARN with the credential and value ([reply](https://github.com/blinklabs-io/dingo/pull/4064#discussion_r3951921954)). This PR does not record the deposit at `InsertDrepIfAbsent` time; I am still glad to open an issue for that. It also does not touch the conformance `DRepRegistrations` comment or the stake-deposit comparison note.

CI is green on 81b4da10, including Windows and Linux race. The only change request still in effect is from 2be06b23, and a re-review when convenient would clear it.

wolf31o2 (approved):

PR #4074, opened 2026-09-07
fix(txsubmission): reject oversized replies before decoding

wolf31o2:
Reject TxSubmission replies exceeding the advertised byte budget before decoding any transaction, preventing over-budget decoder work. Preserve exact body/wire-size checks and size-mismatch metrics.

Regression coverage checks zero, exact and exceeded budgets, including an oversized later body rejected before the first transaction is decoded.

Reference-devnet validation remains incomplete after a transaction-propagation timeout.

---
## Summary by cubic
Rejects TxSubmission replies that exceed the advertised byte budget before decoding any transaction, preventing over-budget decoder work. Previously the aggregate budget was checked only after per-body size checks during decoding; now it's enforced upfront while preserving the existing exact body/wire-size checks and size-mismatch metrics.

**Regression coverage**
- Adds tests for zero, exact, and exceeded budgets, including a later body exceeding the aggregate before the first decode.
- Updates `ARCHITECTURE.md` to reflect the pre-decode budget check.

<sup>Written for commit 5ab6246f6689f8032c7f90edc65f0b982ff45d87. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4074?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -4429,6 +4429,12 @@ trip, which can reduce inbound throughput on high-latency peer links. This is
 required because gouroboros acknowledges every ID returned by a peer; support
 for acknowledging only the fetched prefix would permit batched requests.
 
+TxSubmission checks reply counts and aggregate body bytes against the
+outstanding request before decoding any transaction. Replies exceeding the
+advertised byte budget are classified and counted as size mismatches. Replies
+within that budget still require ordered hash/era matching and an exact body
+or wrapped-wire size match for each transaction before any mempool admission.
+
 The selected pool manages pending transactions:
 
 ```
--- a/ouroboros/txsubmission.go
+++ b/ouroboros/txsubmission.go
@@ -121,12 +121,28 @@ func validateTxsubmissionReply(
 			len(returned),
 		)
 	}
-	ret := make([]validatedTxsubmissionBody, 0, len(returned))
 	var requestedBytes uint64
 	for _, requestedTx := range requested {
 		requestedBytes += uint64(requestedTx.Size)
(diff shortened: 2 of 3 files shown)


arepala-uml (approved):
Reviewed the DoS-hardening fix. Confirmed the check-before-decode ordering is now correct: a new pre-decode loop checks raw body lengths against remaining budget before any CBOR decode runs, whereas pre-fix each tx was fully decoded before the aggregate budget check. Byte-accounting semantics genuinely unchanged — same measure, same inequality, only the timing moved.

- Test quality is excellent and directly verified: the tests use deliberately invalid CBOR so that if the decoder were ever invoked on over-budget data, the test would see a decode error instead of a size-mismatch error. Confirmed empirically by running the same tests against the pre-fix commit — exactly the over-budget subtests fail with a decode error there, proving the fix is load-bearing.
- One intentional, correct side effect: an over-budget body that also has a hash/era mismatch now surfaces as a size-mismatch error instead — by design, not a defect.
- `go build`/`vet`/`test -race`/`golangci-lint` all clean. ARCHITECTURE.md accurate.

No merge-blocking issues found. Cubic's one finding is a cosmetic test-naming nit only.

Issue #3667, opened 2026-08-30
config: redact credentials from token registry URLs


wolf31o2:
Token registry logging records the complete configured source URL on success, failure, and not-modified paths. Query parameters or fragments used for credentials are therefore written to logs.

Acceptance criteria:

- Remove query parameters, fragments, and user information from logged registry URLs.
- Support credentials through a non-URL secret mechanism such as a header.
- Apply the same redaction to every response and error path.
- Add redaction tests with query, fragment, and user-info inputs.

PR #4090, opened 2026-09-07
fix(offchainmetadata): redact credentials from registry sync logs

wolf31o2:
The CIP-26 token registry source URL is operator-supplied and can carry credentials in its userinfo or query string. `TokenRegistrySync` logged that URL verbatim in five places, and wrapped the source-URL validation, request build, and fetch errors with `%w`. URL parser and HTTP client errors echo the URL they were given -- including a redirect target's -- so both the log field and the error `SyncOnce` returns could contain those credentials.

## Change

- `registryLogURL` renders a source URL for a log field with userinfo, query, and fragment removed, keeping scheme, host, port, and path. Input that does not parse, and opaque input whose single section is not separable into host and path, return a fixed `[invalid URL]` placeholder instead of falling back to the raw string.
- All five log sites that carried `s.sourceURL` now render it through `registryLogURL`.
- `registryRequestError` replaces the three `fmt.Errorf(... %w)` wrappings on the source-URL validation, request build, and fetch paths. Its message names only the failed operation. The cause stays reachable through `Unwrap`, so `errors.Is` and `errors.As` continue to identify cancellation, deadline, and transport failures.

The remaining log and error sites in `internal/offchainmetadata` were checked: none renders the registry source URL or an error that carries one. `internal/config/redact.go` already redacts `TokenRegistry.SourceURL` in configuration output.

## Tests

`tokenregistry_redaction_test.go` covers each of the five log sites and each rewrapped error path with a source URL carrying sentinel credentials, asserts the sentinels appear in neither the log output nor the returned error string, asserts a malformed source renders as `[invalid URL]`, and asserts `errors.Is` against `context.Canceled` and `errors.As` for `*url.Error` still resolve through the wrapper.

## Scope

This covers the redaction and test acceptance criteria in #3667. Supporting registry credentials through a non-URL mechanism such as a request header is not part of this change, so the issue stays open.

Refs #3667

---

## Summary by cubic

Redacts operator-supplied credentials from the token registry source URL so they no longer leak into sync logs or returned errors.

Bug Fixes
- Logs render the source URL with userinfo, query, and fragment removed; unparseable or opaque inputs render as `[invalid URL]` instead of echoing the raw string.
- Errors from URL validation, request build, and fetch now report only the failed operation; the original cause stays reachable through `errors.Is` and `errors.As`.
- Tests assert the userinfo username is redacted as well as the password.

### DIFF
--- a/internal/offchainmetadata/tokenregistry_redaction_test.go
+++ b/internal/offchainmetadata/tokenregistry_redaction_test.go
@@ -32,15 +32,18 @@ import (
 // userinfo or its query string. These sentinels stand in for them: none may
 // reach a log field or a returned error string.
 const (
-	redactUser     = "registry-user"
+	redactUser     = "SENTINEL-USERINFO-USER"
 	redactPassword = "SENTINEL-USERINFO-PASSWORD"
 	redactQuery    = "SENTINEL-QUERY-TOKEN"
 	redactFragment = "SENTINEL-FRAGMENT-TOKEN"
 	redactRedirect = "SENTINEL-REDIRECT-TOKEN"
 )
 
-// redactSecrets lists every sentinel that must never be rendered.
+// redactSecrets lists every sentinel that must never be rendered. The
+// username is included: userinfo is a credential component whole, and a
+// redaction that dropped only the password would still identify the account.
 var redactSecrets = []string{
+	redactUser,
 	redactPassword,
 	redactQuery,
 	redactFragment,
--- a/internal/offchainmetadata/tokenregistry_redaction_test.go
+++ b/internal/offchainmetadata/tokenregistry_redaction_test.go
@@ -1,0 +1,408 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package offchainmetadata
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// A registry source is operator-supplied and may carry credentials in its
+// userinfo or its query string. These sentinels stand in for them: none may
+// reach a log field or a returned error string.
+const (
+	redactUser     = "registry-user"
+	redactPassword = "SENTINEL-USERINFO-PASSWORD"
+	redactQuery    = "SENTINEL-QUERY-TOKEN"
+	redactFragment = "SENTINEL-FRAGMENT-TOKEN"
+	redactRedirect = "SENTINEL-REDIRECT-TOKEN"
+)
+
+// redactSecrets lists every sentinel that must never be rendered.
+var redactSecrets = []string{
+	redactPassword,
+	redactQuery,
+	redactFragment,
+	redactRedirect,
+}
+
+// withCredentials appends a credential-bearing query and fragment to a base
+// URL. Userinfo is added separately, because the SSRF guard rejects it before
+// a request is ever built.
+func withCredentials(base string) string {
+	return base + "/?apikey=" + redactQuery + "#" + redactFragment
+}
+
+// captureSync builds a sync whose logger writes to the returned buffer at
+// debug level, so every log site in the package is observable.
+func captureSync(
+	t *testing.T,
+	store TokenRegistryStore,
+	source string,
+) (*TokenRegistrySync, *bytes.Buffer) {
+	t.Helper()
+	buf := &bytes.Buffer{}
+	sync := newTestSync(t, store, source, func(cfg *TokenRegistryConfig) {
+		cfg.Logger = slog.New(slog.NewTextHandler(buf, &slog.HandlerOptions{
+			Level: slog.LevelDebug,
+		}))
+	})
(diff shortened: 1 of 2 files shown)


internal/offchainmetadata/tokenregistry_redaction_test.go
```diff
@@ -0,0 +1,408 @@
...
+const (
+	redactUser     = "registry-user"
+	redactPassword = "SENTINEL-USERINFO-PASSWORD"
+	redactQuery    = "SENTINEL-QUERY-TOKEN"
+	redactFragment = "SENTINEL-FRAGMENT-TOKEN"
+	redactRedirect = "SENTINEL-REDIRECT-TOKEN"
+)
+
+// redactSecrets lists every sentinel that must never be rendered.
+var redactSecrets = []string{
```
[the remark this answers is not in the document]
wolf31o2: Valid. The username is a userinfo credential component, and the checked set omitted it, so an assertion that dropped only the password would have passed. Fixed in 3dd2c1c6: the username joins the checked set and takes the same distinctive sentinel form as the others. Reverting the call sites now fails that subtest on the username specifically. It was the only declared sentinel missing from the set.

arepala-uml (approved):
Reviewed the credential-redaction fix with security-review rigor. The redaction strips userinfo, query-string, and fragment (not just userinfo, the common incomplete-fix pattern) using a generic strip-all-query-params approach rather than a denylist of known secret names — more robust since it can't miss an unlisted secret parameter.

- Traced every log/error call site in the file and confirmed all were fixed — no missed instance.
- Confirmed the real HTTP request still uses the unredacted URL; only log/error rendering is affected.
- Confirmed the tests genuinely fail pre-fix: copying the new test file onto the pre-fix commit fails to compile (`undefined: registryLogURL`).
- Cubic caught a real test gap mid-PR (redaction test list initially missed the username sentinel) — already fixed within this PR's second commit.

No merge blockers. Two non-blocking follow-ups worth filing separately: `fetcher.go` and `mithril/download.go` log similarly-sourced URLs unredacted (lower severity); and ARCHITECTURE.md's existing "Redacted Configuration Logging" section now overstates config-logging as "the one place" secrets could leak, given this new runtime-log mechanism.

internal/offchainmetadata/tokenregistry_sync.go
```diff
@@ -276,6 +277,52 @@ func (s *TokenRegistrySync) SourceURL() string {
...
+//
+// The cause is still reachable through Unwrap, so errors.Is and errors.As
+// continue to identify cancellation, deadline, and transport failures.
+type registryRequestError struct {
+	operation string
+	cause     error
+}
+
+func (e *registryRequestError) Error() string {
+	return e.operation + " failed"
```
chrisguiney, line 319: Non-blocking. `Error()` drops the cause entirely, so the only operator-facing signal for a failing sync becomes `token registry sync failed ... error="validate token registry source URL failed"` or `error="fetch token registry failed"`. There is then no way to tell a rejected userinfo from an unsupported scheme, or a refused connection from a TLS failure or a timeout. This runs on an interval in the background, so that log line is the whole diagnosis surface, and a misconfigured source URL is the most likely thing to land here.

Most of the cause is safe to render. Of `validateURL`'s six failure modes only the `url.Parse` branch carries the raw URL; the other five report a scheme, host, or IP that is already in the redacted `url` log field. On the fetch path the credential lives in `url.Error.URL`, so rendering the first cause in the chain that is not a `*url.Error` keeps "connection refused" and "context canceled" while never touching that field.

One caveat if that is done: a single `errors.Unwrap` is not enough. A redirect rejected by `restrictedRedirectPolicy` nests a second `*url.Error` carrying the redirect target, so the walk has to skip every `*url.Error` layer rather than one.

internal/offchainmetadata/tokenregistry_sync.go
```diff
@@ -276,6 +277,52 @@ func (s *TokenRegistrySync) SourceURL() string {
 	return s.sourceURL
```
chrisguiney, line 277: Non-blocking, latent. `SourceURL()` returns the unredacted source. Nothing in the node calls it today — only tests do — so this is not a live leak, but it is the one exported door left on the class this change closes, and a future caller that logs it would reopen it. A doc comment noting that the return is unredacted and that `registryLogURL` is the rendering intended for logs would make that hard to get wrong.

chrisguiney (approved):
Approving. All checks are green at 3dd2c1c6 and the change holds as a security fix.

**Class audit of the registry source URL inside `TokenRegistrySync` — complete, no member still leaking.**

- All five log sites render through `registryLogURL`; the request itself keeps the raw URL, which is required.
- The three error wraps that carried a `*url.Error` are replaced, and the remaining error returns in `SyncOnce` and `applySnapshot` — unexpected status, nil response, sync-state and stamp writes, snapshot size — carry no URL.
- `tokenRegistrySnapshotIdentity` hashes the source before it is persisted, so no plaintext reaches the store.
- Configuration output is covered separately by `TokenRegistry.SourceURL` in `logURIConfigFields`, and `token-registry-source-url` defaults to empty, so `--help` echoes nothing.
- No metric label, span attribute, or trace attribute renders the source.

The new tests are regression tests rather than guards: with the six call sites reverted in place and the helpers kept, every subtest of `TestTokenRegistrySyncRedactsCredentialsFromLogs`, `TestTokenRegistrySyncRedactsCredentialsFromReturnedError`, and `TestTokenRegistrySyncFetchErrorStaysIdentifiable` fails on its sentinel-absence assertion. `TestRegistryLogURLRedactsCredentialComponents` and `TestRegistryRequestErrorKeepsCauseIdentity` are unit tests of the new helpers and pass either way. Asserting absence of the plaintext across the whole captured buffer, rather than presence of the mask, is the right shape for this.

`registryLogURL` dropping userinfo whole is stronger than `url.URL.Redacted()`, which keeps the username; the username is a credential component and this is the better choice.

On the adjacent class: `fetcher.go` logs `doc.URL`, which is an on-chain metadata pointer rather than operator configuration, so it is public data and a different threat model — not a gap this change leaves open.

`Refs #3667` is accurate: the issue's remaining acceptance criterion is a non-URL credential mechanism, which this does not implement, so it correctly stays open.

Two non-blocking recommendations inline.

Issue #3659, opened 2026-08-30
storage: make blob commits rollback-safe and report uncertain outcomes


wolf31o2:
Blob mutation failure paths do not preserve transaction outcome accurately:

- A successful UTxO blob deletion commits outside the caller's database transaction and survives caller rollback.
- S3 and GCS operations can take effect and still return an error; compensation skips the current operation and can return an ordinary error even though remote state changed.

Acceptance criteria:

- Make blob deletions participate in the caller's commit or restore them on rollback.
- Include the current remote operation in compensation when its outcome is uncertain.
- Return `ErrPartialCommit` whenever the final remote state cannot be proven or restored.
- Add rollback and apply-then-error failure-injection tests.

PR #4092, opened 2026-09-07
fix(database): stage UTxO blob deletes in the caller's transaction

wolf31o2:
Refs #3659

`deleteUtxoBlobs` took the caller's transaction and ignored it, opening a blob-only transaction per batch and committing each one before the caller did any metadata work. A successful blob delete therefore survived the caller's rollback, leaving metadata that points at blob data which is already gone.

`UtxosDeleteRolledback` and `UtxosDeleteConsumed` both delete blobs before the metadata delete they accompany, so a failure after the blob commit left the two stores disagreeing.

When the caller supplies a transaction with a blob handle the deletes are staged in it and commit or roll back with the metadata, matching `deleteTxBlobs`. Callers that supply no blob handle keep the batched fallback.

This covers the issue's first acceptance criterion, that blob deletions participate in the caller's commit. The S3 and GCS compensation criteria are not addressed here.

---
## Summary by cubic
Stages UTxO blob deletes in the caller's transaction instead of committing them in separate batch transactions, so a rollback now discards them along with the metadata delete they accompany. This fixes the inconsistency where metadata could point to already-deleted blobs.

- Callers that supply a transaction without a blob handle keep the batched fallback.

<sup>Written for commit c98ef8c4667d4bb24072905711ba9cf8e31072a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4092?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/database/transaction_test.go
+++ b/database/transaction_test.go
@@ -1138,3 +1138,49 @@ func TestRecoverConsumedUtxoRefusesOffPrimaryChainProducer(t *testing.T) {
 		"the gate must be the only thing refusing this producer")
 	require.ErrorContains(t, err, "decode transaction output")
 }
+
+// TestDeleteUtxoBlobsUsesCallerBlobTxn pins that UTxO blob deletes are staged
+// in the caller's transaction rather than committed on their own.
+//
+// UtxosDeleteRolledback and UtxosDeleteConsumed delete blobs before the
+// metadata delete they accompany. Committing the blob deletes separately let a
+// successful delete survive the caller's rollback, leaving metadata that points
+// at blob data which is already gone.
+func TestDeleteUtxoBlobsUsesCallerBlobTxn(t *testing.T) {
+	t.Parallel()
+
+	store := &mockBlobStore{}
+	db := &Database{
+		blobRef: newBlobStoreRef(store),
+		logger: slog.New(
+			slog.NewJSONHandler(
+				io.Discard,
+				&slog.HandlerOptions{Level: slog.LevelDebug},
+			),
+		),
+	}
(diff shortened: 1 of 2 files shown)


arepala-uml (approved):
Reviewed the transactional-integrity fix. Confirmed the bug precisely: `deleteUtxoBlobs` ignored the caller's transaction and opened/committed its own per-500-item batch transaction, so a later rollback of the caller's outer transaction couldn't undo blob deletes that had already auto-committed — an unrecoverable dangling metadata reference.

- Reproduced the bug directly: applied the new test standalone to the pre-fix commit — it fails (`should have 1 item(s), but has 2`) — and confirmed it passes at the fix commit.
- Audited the class: checked both sibling blob-delete paths (`deleteTxBlobs`, `repairUtxoBlob`) — both were already correct. This was the only instance.
- The new test genuinely proves atomicity: asserts one blob transaction opened, nothing committed, then explicitly rolls back the caller's transaction and confirms the staged deletes are rolled back too.

No merge blockers. One item worth fixing: `deleteUtxoBlobs`'s doc comment is now stale (still claims the caller's txn is ignored) — Cubic already flagged this and it's unaddressed; please update it.

PR #4095, opened 2026-09-07
fix(ledger): enforce block reference script budgets

### LINKED ISSUE #4081: fix(ledger): enforce aggregate block reference-script limits
Validated block processing and forged-block validation enforce per-transaction rules but do not call the Conway/Dijkstra aggregate reference-script size validators. Individually valid transactions can collectively exceed the block limit.

Enforce the era-specific aggregate limit before applying block transactions and before accepting forged blocks. Use the block's UTxO view and preserve protocol-version rules for prior outputs, consumed inputs, phase-2-invalid transactions, and Dijkstra subtransactions. Cover exact-limit acceptance and over-limit rejection through both entry paths, with each transaction remaining within its own limit.

wolf31o2:
Enforce Conway and Dijkstra aggregate consumed reference-script limits before validated block application and default forging adoption. Full cryptographic self-validation remains optional.

For Conway protocol versions through 10, measure against the pre-block UTxO set and omit genuinely absent entries from aggregate accounting. Preserve storage/decode errors and ordinary transaction validation. Protocol version 11 and Dijkstra retain their existing accounting rules.

Select era-appropriate parameters for overlap blocks, reject typed-nil parameters before aggregate validation, and preserve the explicit Musashi prototype validation policy.

Fixes #4081.

### DIFF
--- a/ledger/block_reference_script_test.go
+++ b/ledger/block_reference_script_test.go
@@ -140,6 +140,7 @@ func TestBlockReferenceScriptLimitAdmission(t *testing.T) {
 						ConwayProtocolParameters: *pp,
 						MaxRefScriptSizePerBlock: 1,
 					}
+					currentParams.ProtocolVersion.Major = dijkstra.MinProtocolVersionDijkstra
 					return db.Transaction(true).Do(func(txn *database.Txn) error {
 						_, err := ls.ledgerProcessBlock(txn, ocommon.NewPoint(1, block.Hash().Bytes()), block, true, false, false, nil, envelopeParent{origin: true}, nil, eras.DijkstraEraDesc, currentParams, pp, 0)
 						return err
--- a/ledger/block_reference_script_test.go
+++ b/ledger/block_reference_script_test.go
@@ -1,0 +1,167 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ledger
+
+import (
+	"bytes"
+	"errors"
+	"testing"
+
+	"github.com/blinklabs-io/dingo/database"
+	"github.com/blinklabs-io/dingo/database/models"
+	"github.com/blinklabs-io/dingo/ledger/eras"
+	"github.com/blinklabs-io/gouroboros/cbor"
+	"github.com/blinklabs-io/gouroboros/ledger/babbage"
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/blinklabs-io/gouroboros/ledger/conway"
+	"github.com/blinklabs-io/gouroboros/ledger/shelley"
+	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBlockReferenceScriptLimitAdmission(t *testing.T) {
+	for _, over := range []bool{false, true} {
+		name := "at limit"
+		if over {
+			name = "over limit"
+		}
+		t.Run(name, func(t *testing.T) {
+			db := newTestDB(t)
+			address, err := lcommon.NewAddressFromParts(
+				lcommon.AddressTypeKeyNone,
+				0,
+				bytes.Repeat([]byte{1}, 28),
+				nil,
+			)
+			require.NoError(t, err)
+			block := &conway.ConwayBlock{
+				BlockHeader: &conway.ConwayBlockHeader{},
+			}
+			block.BlockHeader.Body.BlockNumber = 1
+			block.BlockHeader.Body.Slot = 1
+			block.BlockHeader.Body.ProtoVersion.Major = 11
+			for i := range 6 {
+				size := int(conway.MaxRefScriptSizePerBlock / 6)
+				if i == 5 {
+					size += int(conway.MaxRefScriptSizePerBlock % 6)
+					if over {
+						size++
+					}
+				}
+				require.Less(t, uint64(size), conway.MaxRefScriptSizePerTx)
+				txID := bytes.Repeat([]byte{byte(i + 1)}, 32)
+				input := shelley.ShelleyTransactionInput{
+					TxId: lcommon.NewBlake2b256(txID),
+				}
+				output := &babbage.BabbageTransactionOutput{
+					OutputAddress: address,
+					TxOutScriptRef: &lcommon.ScriptRef{
+						Type:   lcommon.ScriptRefTypePlutusV3,
+						Script: make(lcommon.PlutusV3Script, size),
+					},
+				}
+				encoded, err := cbor.Encode(output)
+				require.NoError(t, err)
+				require.NoError(
+					t,
+					db.Transaction(true).Do(func(txn *database.Txn) error {
+						if err := db.CreateUtxo(txn, &models.Utxo{TxId: txID, OutputIdx: 0, AddedSlot: 0}); err != nil {
+							return err
+						}
+						return db.Blob().SetUtxo(txn.Blob(), txID, 0, encoded)
+					}),
+				)
+				block.TransactionBodies = append(
+					block.TransactionBodies,
+					conway.ConwayTransactionBody{
+						TxReferenceInputs: cbor.NewSetType(
+							[]shelley.ShelleyTransactionInput{input},
+							false,
+						),
+					},
+				)
+				block.TransactionWitnessSets = append(
+					block.TransactionWitnessSets,
+					conway.ConwayTransactionWitnessSet{},
+				)
+			}
+			encodedBlock, err := cbor.EncodeGeneric(block)
+			require.NoError(t, err)
+			block.SetCbor(encodedBlock)
+			bodySize, err := serializedBlockBodySize(block)
+			require.NoError(t, err)
+			block.BlockHeader.Body.BlockBodySize = bodySize
+			encodedBlock, err = cbor.EncodeGeneric(block)
+			require.NoError(t, err)
+			block.SetCbor(encodedBlock)
+			pp := &conway.ConwayProtocolParameters{
+				MaxBlockBodySize:   100000,
+				MaxBlockHeaderSize: 100000,
+				ProtocolVersion: lcommon.ProtocolParametersProtocolVersion{
+					Major: 11,
+				},
+			}
+			sentinel := errors.New("transaction validator reached")
+			era := eras.ConwayEraDesc
+			era.ValidateTxFunc = func(lcommon.Transaction, uint64, lcommon.LedgerState, lcommon.ProtocolParameters) error {
+				return sentinel
+			}
+			nodeConfig := newTestShelleyGenesisCfg(t)
+			nodeConfig.ShelleyGenesis().NetworkId = "Testnet"
+			ls := &LedgerState{
+				db:             db,
+				activeEras:     []eras.EraDesc{era},
+				currentEra:     era,
+				currentPParams: pp,
+				config: LedgerStateConfig{
+					Logger:            testLogger(),
+					CardanoNodeConfig: nodeConfig,
+				},
+			}
+			ls.metrics.init(prometheus.NewRegistry())
+			ls.publishSnapshotsLocked()
+			for path, run := range map[string]func() error{
+				"imported": func() error {
+					return db.Transaction(true).Do(func(txn *database.Txn) error {
+						_, err := ls.ledgerProcessBlock(txn, ocommon.NewPoint(1, block.Hash().Bytes()), block, true, false, false, nil, envelopeParent{origin: true}, nil, era, pp, nil, 0)
+						return err
+					})
+				},
+				"forged": func() error { return ls.validateForgedTxs(block) },
+			} {
+				t.Run(path, func(t *testing.T) {
+					err := run()
+					if over {
+						var limit lcommon.RefScriptSizePerBlockTooLargeError
+						require.ErrorAs(
+							t,
+							err,
+							&limit,
+							"aggregate reference-script limit must reject before transaction validation",
+						)
+						require.Equal(
+							t,
+							conway.MaxRefScriptSizePerBlock+1,
+							limit.BlockSize,
+						)
+					} else {
+						require.ErrorIs(t, err, sentinel, "exact block limit must reach later transaction validation")
+					}
+				})
+			}
+		})
+	}
+}
--- a/ledger/block_reference_script_test.go
+++ b/ledger/block_reference_script_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/blinklabs-io/gouroboros/ledger/babbage"
 	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
 	"github.com/blinklabs-io/gouroboros/ledger/conway"
+	"github.com/blinklabs-io/gouroboros/ledger/dijkstra"
 	"github.com/blinklabs-io/gouroboros/ledger/shelley"
 	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
 	"github.com/prometheus/client_golang/prometheus"
@@ -123,7 +124,7 @@ func TestBlockReferenceScriptLimitAdmission(t *testing.T) {
 			nodeConfig.ShelleyGenesis().NetworkId = "Testnet"
 			ls := &LedgerState{
 				db:             db,
-				activeEras:     []eras.EraDesc{era},
+				activeEras:     []eras.EraDesc{era, eras.DijkstraEraDesc},
 				currentEra:     era,
 				currentPParams: pp,
 				config: LedgerStateConfig{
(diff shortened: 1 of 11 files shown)


ledger/block_reference_script_test.go
```diff
@@ -0,0 +1,167 @@
...
+					block.TransactionWitnessSets,
+					conway.ConwayTransactionWitnessSet{},
+				)
+			}
+			encodedBlock, err := cbor.EncodeGeneric(block)
+			require.NoError(t, err)
+			block.SetCbor(encodedBlock)
+			bodySize, err := serializedBlockBodySize(block)
+			require.NoError(t, err)
+			block.BlockHeader.Body.BlockBodySize = bodySize
```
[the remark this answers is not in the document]
wolf31o2: `cbor.EncodeGeneric` explicitly bypasses the source object's `MarshalCBOR` and excludes `DecodeStoreCbor` when constructing its temporary encoding type. The header here is constructed directly, with no decoded header cache, so the second call includes the updated `BlockBodySize`. Clearing the block cache is unnecessary for this API.

Reference: [gouroboros v0.204.0 EncodeGeneric](https://github.com/blinklabs-io/gouroboros/blob/v0.204.0/cbor/encode.go#L65-L103).

ledger/state.go
```diff
@@ -7123,6 +7123,16 @@ func (ls *LedgerState) ledgerProcessBlock(
 	if err := ls.verifyDeferredBlockHeaderState(txn, point, block); err != nil {
 		return nil, err
 	}
+	// Check the ranking block after any applicable endorser transactions,
+	// using their resulting state but before its own transaction mutations.
+	// The explicitly non-validating Musashi prototype keeps its trust policy.
+	if shouldValidate && !ls.skipDijkstraTxValidation(currentEra.Id) {
+		if err := validateBlockReferenceScripts(block, pparams, &LedgerView{
+			txn: txn, ls: ls,
+		}); err != nil {
```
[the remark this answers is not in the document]
wolf31o2: Imported admission now selects the previous era's protocol parameters for an immediately preceding-era block, matching forged validation. The selection is local to the aggregate check; the active-era validation policy remains unchanged. Regression coverage exercises Conway blocks at and above their aggregate limit while Dijkstra is active with a different configured limit.

arepala-uml (dismissed):
Reviewed the dual-enforcement and era-gating claims independently. Confirmed against upstream `cardano-ledger` source (Conway: 200KB/tx, 1MB/block hardcoded; Dijkstra: governable equivalents) that the aggregate reference-script size limits and non-distinct aggregation semantics match both the spec and gouroboros's implementation.

- Confirmed both enforcement points: the peer-block validation path (`ledger/state.go`) and dingo's own forging path (`node_forging.go`/`forger.go`, where the sole production `BlockForger` construction now always wires in a validator) — Dingo can't build or adopt an over-budget block itself.
- Era-gating confirmed correct — pre-Conway blocks are never subject to this check.
- Cubic's P1 finding (wrong-era params selection near an era boundary) was already fixed with a regression test at this head. Cubic's P2 finding (stale CBOR bytes) is a false positive — verified against gouroboros's actual `EncodeGeneric`, which cannot replay cached bytes.
- `go build`/`vet`/`test -race` clean across all relevant packages, plus a full conformance suite run (105s, all eras, no regressions).

No merge-blocking issues found. One non-blocking suggestion: a targeted test proving a hypothetically over-budget pre-Conway block is NOT rejected would make the era gate's correctness explicit rather than inferred.

ledger/state.go
```diff
@@ -7123,6 +7123,20 @@ func (ls *LedgerState) ledgerProcessBlock(
...
 	}
+	// Check the ranking block after any applicable endorser transactions,
+	// using their resulting state but before its own transaction mutations.
+	// The explicitly non-validating Musashi prototype keeps its trust policy.
+	if shouldValidate && !ls.skipDijkstraTxValidation(currentEra.Id) {
+		referenceParams := pparams
+		if uint(block.Era().Id)+1 == currentEra.Id && prevEraPParams != nil {
+			referenceParams = prevEraPParams
+		}
+		if err := validateBlockReferenceScripts(block, referenceParams, &LedgerView{
```
chrisguiney, line 7134: Blocking: below PV11 this rejects a block the Haskell node accepts.

`conway.ValidateRefScriptSizePerBlock` passes `includeProduced = pp.ProtocolVersion.Major >= 11` into `common.ConsumedReferenceScriptSizePerBlock`. Below PV11 that function measures every transaction against the raw `utxoState` with no intra-block overlay, and the view supplied here is the pre-block state. `common.ConsumedReferenceScriptSize` then returns an error for an unresolvable *reference* input while silently skipping an unresolvable *regular* input (gouroboros v0.204.0 `ledger/common/ref_scripts.go:76-84`). A transaction that reference-inputs an output created by an earlier transaction in the same block therefore fails the lookup, and `ledgerProcessBlock` returns that error and rejects the block.

The upstream rule does not error there. `getReferenceScriptsNonDistinct` resolves inputs with `Map.restrictKeys` (`cardano-ledger`, `Cardano/Ledger/Babbage/UTxO.hs`), so an absent input is dropped and contributes 0; `ConwayBBODY` sums those per-transaction sizes and the block passes. The block is valid, and it validates per-transaction here too, because that path carries `intraBlockUtxos`.

Mainnet Conway is at protocol version 10, so `includeProduced` is false for every block this node applies at tip. This PR's own matrix shows the divergence on a single block: `TestBlockReferenceScriptProducedOutputAdmission/pv11_below` reaches transaction validation, while `pv10_no_overlay` — the identical block with only the protocol major changed — is rejected with `resolve consumed reference-script input`. Rejecting a block mainnet accepts is a chain split in the direction that strands this node.

The cause is in `blinklabs-io/gouroboros`, not here: `ConsumedReferenceScriptSize` should treat an unresolvable reference input the same way it already treats an unresolvable regular input and contribute 0, matching `Map.restrictKeys`. Landing that upstream, re-pinning, and then flipping the expectations noted on the test file looks better than classifying error kinds on this side, which would also have to decide what a genuine storage failure means for block admission.

ledger/validate_forged.go
```diff
@@ -150,3 +157,31 @@ func (ls *LedgerState) validateForgedTxs(block ledger.Block) error {
...
+	snapshot := ls.loadConsensusSnapshot()
+	if ls.skipDijkstraTxValidation(snapshot.currentEra.Id) {
+		return nil
+	}
+	pp := snapshot.currentPParams
+	if uint(block.Era().Id)+1 == snapshot.currentEra.Id &&
+		snapshot.prevEraPParams != nil {
+		pp = snapshot.prevEraPParams
+	}
+	return ls.db.Transaction(false).Do(func(txn *database.Txn) error {
```
chrisguiney, line 180: Same defect on the forging path, and here it is unconditional rather than opt-in.

`validateForgedTxs` builds its intra-block overlay after this call, so below PV11 a block this node built from a mempool holding a reference-script publication together with its first consumer is dropped instead of adopted — a forfeited leadership slot rather than a rejected peer block. `TestBlockReferenceScriptProducedOutputAdmission/pv10_no_overlay/forged` asserts that rejection.

Non-blocking, separately: the `block.Era().Id` gate on line 168 is the value `dijkstraEraGate` in `ledger/state.go` documents as not authoritative for ledger-era gates. It is harmless as an early return, since the type switch in `validateBlockReferenceScripts` returns nil for any other block type, but line 176 uses the same value to choose which era's protocol parameters apply.

ledger/block_reference_scripts_additional_test.go
```diff
@@ -0,0 +1,386 @@
...
+				want = uint64(size * 6)
+			}
+			additionalReferenceAdmission(
+				t,
+				db,
+				block,
+				eras.ConwayEraDesc,
+				pp,
+				want,
+				tc.proto < 11 || tc.invalidProducer || tc.missing,
```
chrisguiney: `tc.proto < 11` codifies the PV10 rejection as expected behavior, which is what makes the blocker on `ledger/state.go` hard to spot from the test output alone.

Once the upstream lookup contributes 0 instead of erroring, `pv10_no_overlay` should expect the block to reach transaction validation: at PV10 the pre-block UTxO genuinely does not contain the intra-block output, and the upstream rule counts that absence as 0 rather than as a failure.

`invalid_producer` and `missing_reference` are a different case — those blocks are invalid either way — but they are currently rejected by the aggregate rule rather than by the input check that owns them, so their error attribution changes with the same fix.

ledger/block_reference_scripts.go
```diff
@@ -0,0 +1,40 @@
...
+	block ledger.Block,
+	pp lcommon.ProtocolParameters,
+	state lcommon.UtxoState,
+) error {
+	switch b := block.(type) {
+	case *conway.ConwayBlock:
+		return conway.ValidateRefScriptSizePerBlock(b, pp, state)
+	case *dijkstra.DijkstraBlock:
+		return dijkstra.ValidateRefScriptSizePerBlock(b, pp, state)
+	default:
```
chrisguiney, line 78: Class audit result, for the record: `ValidateRefScriptSizePerBlock` is defined only in `ledger/conway` and `ledger/dijkstra` in gouroboros v0.204.0. No pre-Conway era carries a copy, since the rule arrives with the Conway tiered reference-script model, so this `default` branch is the complete era gate rather than a partial one.

Non-blocking: nothing asserts that gate. A Babbage block carrying reference scripts past the Conway block budget is safe here only by construction. A case in `TestValidateBlockReferenceScriptsEntryControls` covering a pre-Conway block would make it explicit.

chrisguiney (changes requested):
One merge blocker; the rest of the change holds up.

Confirmed sound: era gating (`ValidateRefScriptSizePerBlock` exists only for Conway and Dijkstra in gouroboros v0.204.0, so the type switch in `validateBlockReferenceScripts` covers the entire rule class and pre-Conway blocks are never subject to it), dual enforcement across the imported and forged paths, and the era-boundary parameter selection.

Blocker: below protocol version 11 the aggregate check rejects a block whose transaction reference-inputs an output created by an earlier transaction in the same block. Cardano mainnet Conway is at protocol version 10, so that is the path this node takes for every block it applies at tip. Detail inline on `ledger/state.go`.

Prior findings:

- cubic P1 (era-boundary parameter selection) — fixed in `fd128ab`, covered by `TestBlockReferenceScriptLimitAdmission/*/imported_previous_era`. Confirmed at this head.
- cubic P2 (stale CBOR bytes in the fixture) — false positive. `cbor.EncodeGeneric` builds a temporary struct type that excludes `DecodeStoreCbor` and bypasses `MarshalCBOR` (gouroboros v0.204.0 `cbor/encode.go:76-99`), so it cannot replay cached bytes.
- CodeRabbit posted no review record on this pull request; its check reports the review as rate limited.

Non-blocking: the aggregate check resolves and CBOR-decodes every input and reference input of every transaction through an uncached `LedgerView`, immediately before the per-transaction path resolves the same inputs again. That roughly doubles UTxO reads per Conway block at tip.

Checks run at this head: `go build ./...`, `go vet ./ledger/... .`, `make docs-parity`, `make import-boundaries`, `go test ./internal/test/conformance/`, `golangci-lint run ./ledger/... ./`, and `go test -race` on `./ledger/`, `./ledger/forging/` and the root package. All pass.

ledger/state.go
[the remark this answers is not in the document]
chrisguiney: Confirmed fixed at this head. `ledger/state.go` selects `prevEraPParams` when the block's era immediately precedes the active era, matching `ValidateBlockReferenceScripts`, and `TestBlockReferenceScriptLimitAdmission/*/imported_previous_era` covers a Conway block validated while Dijkstra is active with a different configured limit.

ledger/block_reference_script_test.go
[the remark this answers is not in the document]
chrisguiney: False positive, confirmed. `cbor.EncodeGeneric` constructs a temporary struct type from the exported fields excluding `DecodeStoreCbor`, then encodes that copy, so the source object's `MarshalCBOR` is never reached and cached bytes cannot be replayed (gouroboros v0.204.0 `cbor/encode.go:76-99`). The `BlockBodySize` mutation is present in the re-encoded block without clearing the cache.

wolf31o2:
The PV10 same-block reference-input case is handled at the Dingo boundary in `8e315892`: aggregate accounting wraps the pre-block UTxO view so genuinely absent reference inputs contribute zero, while non-not-found lookup errors remain fatal. The imported and forged admission paths share this behavior, and the regression matrix covers PV10/PV11, missing references, invalid producers, and lookup errors. Please re-review the current head.

ledger/state.go
chrisguiney: Resolved at `5c222873`.

`pv10ReferenceScriptState` maps `database.ErrUtxoNotFound` to a zero-value UTxO for aggregate accounting only, and only when the Conway protocol version is at or below `ProtocolVersionPlomin` — the exact complement of the `>= ProtocolVersionVanRossem` `includeProduced` switch in `conway.ValidateRefScriptSizePerBlock`.

That matches the reference implementation. `totalRefScriptSizeInBlock` in `Cardano/Ledger/Conway/Rules/Bbody.hs` measures every transaction against the unchanged pre-block UTxO for `pvMajor protVer <= natVersion @10`, and `txNonDistinctRefScriptsSize` resolves through `getReferenceScriptsNonDistinct`, whose `Map.restrictKeys` drops absent keys rather than failing. An intra-block-produced reference input therefore contributes zero and the block proceeds to transaction validation.

The wrapper is load-bearing: without it, `TestBlockReferenceScriptProducedOutputAdmission/pv10_no_overlay` reports `resolve consumed reference-script input …: utxo not found` where the later-validator sentinel was expected, on both the `imported` and `forged` sub-paths.

ledger/validate_forged.go
chrisguiney: Resolved at `5c222873` by the same change. `validateForgedTxs` reaches the aggregate rule through `ValidateBlockReferenceScripts`, which shares `validateBlockReferenceScripts` with the imported path, so the protocol-version-10 accounting view applies here too.

The `forged` sub-path of `TestBlockReferenceScriptProducedOutputAdmission/pv10_no_overlay` is the control for this half: with the accounting view removed it fails identically to the `imported` sub-path.

ledger/block_reference_scripts_additional_test.go
chrisguiney: Resolved. The expectation is now `tc.invalidProducer || tc.missing`, so `pv10_no_overlay` requires the block to reach transaction validation instead of codifying the rejection.

`TestPV10AggregateReferenceScriptLookupErrors` pins the boundary directly: an absent input contributes zero at protocol version 10, a lookup error that is not `ErrUtxoNotFound` still rejects, and the same absent input still rejects at protocol version 11. `TestValidateTxConwayGenuinelyMissingReferenceInputStillRejected` in `ledger/eras/conway_utxo_divergence_test.go` holds the other side — an absent reference input is still a `ReferenceInputResolutionError` at the transaction level — so the relaxation stays scoped to aggregate accounting.

ledger/block_reference_scripts.go
```diff
@@ -0,0 +1,66 @@
...
+func validateBlockReferenceScripts(
+	block ledger.Block,
+	pp lcommon.ProtocolParameters,
+	state lcommon.UtxoState,
+) error {
+	switch b := block.(type) {
+	case *conway.ConwayBlock:
+		if conwayPParams, ok := pp.(*conway.ConwayProtocolParameters); ok &&
+			state != nil &&
+			conwayPParams.ProtocolVersion.Major <= lcommon.ProtocolVersionPlomin {
```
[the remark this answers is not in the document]
chrisguiney: Confirmed and fixed at `5c222873`. The `conwayPParams == nil` check and the Dijkstra typed-nil switch are both load-bearing: without them, `TestBlockReferenceScriptsRejectNilParameters/conway/typed_nil_conway` and `/dijkstra/typed_nil_conway` panic on a nil pointer dereference at the `ProtocolVersion.Major` read rather than returning an error.

ledger/block_reference_scripts.go
```diff
@@ -0,0 +1,81 @@
...
+		conwayPParams, ok := pp.(*conway.ConwayProtocolParameters)
+		if !ok || conwayPParams == nil {
+			return errors.New("pparams are not expected type")
+		}
+		if state != nil &&
+			conwayPParams.ProtocolVersion.Major <= lcommon.ProtocolVersionPlomin {
+			state = pv10ReferenceScriptState{state: state}
+		}
+		return conway.ValidateRefScriptSizePerBlock(b, pp, state)
+	case *dijkstra.DijkstraBlock:
```
chrisguiney, line 64: Non-blocking symmetry note: the pre-PV11 restricted-map view is applied only on the Conway branch. `dijkstra.ValidateRefScriptSizePerBlock` reads the same `ProtocolVersion.Major >= ProtocolVersionVanRossem` switch for `includeProduced`, so a Dijkstra block carrying Conway parameters at major 10 would take the pre-block-UTxO path without absent-key suppression and reject an intra-block reference input. `TestBlockReferenceScriptsAcceptParameterShapes/dijkstra_conway` constructs that exact parameter shape, though with no transactions.

Unreachable on a real chain, since Dijkstra follows the protocol-version-11 Conway intra-era fork, so this is a note rather than a defect.

chrisguiney (approved):
The protocol-version-10 blocker from the previous review is resolved, and every check on this head passes.

`pv10ReferenceScriptState` suppresses only `database.ErrUtxoNotFound`, only for aggregate accounting, and only for Conway parameters at or below `ProtocolVersionPlomin` — the exact complement of the `includeProduced` switch in `conway.ValidateRefScriptSizePerBlock`. That matches `totalRefScriptSizeInBlock` in the reference implementation, which measures every transaction against the unchanged pre-block UTxO for `pvMajor <= 10` and resolves through `Map.restrictKeys`, so an absent key contributes zero instead of failing. Storage and decode errors still reject, and an absent reference input is still a transaction-level failure.

Both entry paths share `validateBlockReferenceScripts`, and `pv10_no_overlay` exercises the imported and forged sub-paths against the same expectation.

Cubic's typed-nil P1 is fixed and covered by `TestBlockReferenceScriptsRejectNilParameters`.

Non-blocking, carried over: at protocol version 10 the aggregate check resolves and decodes every input and reference input of every transaction through an uncached `LedgerView` immediately before per-transaction validation resolves the same inputs again, roughly doubling UTxO reads per block at tip. One symmetry note inline.

PR #4097, opened 2026-09-07
fix(ledger): bound hardfork summary durations

wolf31o2:
## Summary
- validate cached hardfork epoch durations before conversion and accumulation
- reject malformed or overflowing cached values with boundary regression coverage

## Validation
- focused ledger and hardfork race tests
- caller-focused summary, slot, and header tests
- repeated summary tests
- git diff --check

---
## Summary by cubic
Bounds hardfork summary epoch durations so invalid cached values return errors with context instead of silently wrapping, while keeping unpopulated epoch rows (zero slot length/epoch size) accepted as before. Adds regression tests for zero-length epochs, slot-length overflow, and cumulative overflow.

<sup>Written for commit e860ce305cecb69e0d88bc661a6dfaa371d74775. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4097?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ledger/hardfork_summary.go
+++ b/ledger/hardfork_summary.go
@@ -16,10 +16,33 @@ package ledger
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/blinklabs-io/dingo/ledger/hardfork"
 )
+
+const maxHardForkSlotLengthMilliseconds = uint64(1<<63-1) / uint64(time.Millisecond)
+
+func hardForkCachedEraParams(
+	lengthInSlots uint,
+	slotLengthMilliseconds uint,
+) (hardfork.EraParams, error) {
+	if uint64(slotLengthMilliseconds) > maxHardForkSlotLengthMilliseconds {
+		return hardfork.EraParams{}, fmt.Errorf(
+			"slot length %dms overflows time.Duration",
+			slotLengthMilliseconds,
+		)
+	}
+	params := hardfork.EraParams{
+		EpochSize:  uint64(lengthInSlots),
+		SlotLength: time.Duration(slotLengthMilliseconds) * time.Millisecond,
+	}
+	if err := params.Validate(); err != nil {
+		return hardfork.EraParams{}, err
+	}
+	return params, nil
+}
 
 // HardForkSummary constructs a hardfork.Summary describing the chain's era
 // history from the LedgerState's current epoch cache, tip, current era, and
@@ -87,10 +110,20 @@ func (ls *LedgerState) hardForkSummaryAnchoredAt(
 		eraID := first.EraId
 		// Per-epoch params within an era are expected to be constant; we use
 		// the first epoch's values as the era-level params.
-		// first.SlotLength is protocol-bounded (milliseconds per slot).
-		// #nosec G115
-		slotLen := time.Duration(first.SlotLength) * time.Millisecond
-		epochSize := uint64(first.LengthInSlots)
+		eraParams, err := hardForkCachedEraParams(
+			first.LengthInSlots,
+			first.SlotLength,
+		)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"ledger: cached epoch %d (era %d) params invalid: %w",
+				first.EpochId,
+				eraID,
+				err,
+			)
+		}
+		slotLen := eraParams.SlotLength
+		epochSize := eraParams.EpochSize
 
 		start := hardfork.Bound{
 			RelativeTime: relTime,
@@ -103,8 +136,17 @@ func (ls *LedgerState) hardForkSummaryAnchoredAt(
 		j := i
 		for j < len(cache) && cache[j].EraId == eraID {
 			ep := cache[j]
-			// LengthInSlots and SlotLength are protocol-bounded uints.
-			// #nosec G115
+			if _, err := hardForkCachedEraParams(
+				ep.LengthInSlots,
+				ep.SlotLength,
+			); err != nil {
+				return nil, fmt.Errorf(
+					"ledger: cached epoch %d (era %d) params invalid: %w",
+					ep.EpochId,
+					eraID,
+					err,
+				)
+			}
 			relTime += time.Duration(ep.LengthInSlots) *
 				time.Duration(ep.SlotLength) * time.Millisecond
 			j++
--- a/ledger/hardfork_summary.go
+++ b/ledger/hardfork_summary.go
@@ -22,7 +22,10 @@ import (
 	"github.com/blinklabs-io/dingo/ledger/hardfork"
 )
 
-const maxHardForkSlotLengthMilliseconds = uint64(1<<63-1) / uint64(time.Millisecond)
+const (
+	maxHardForkDurationNanoseconds    = uint64(1<<63 - 1)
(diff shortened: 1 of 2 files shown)


ledger/hardfork_summary.go
```diff
@@ -16,11 +16,61 @@ package ledger
...
+	slotLengthMilliseconds uint,
+) (time.Duration, error) {
+	params, err := hardForkCachedEraParams(
+		lengthInSlots,
+		slotLengthMilliseconds,
+	)
+	if err != nil {
+		return 0, err
+	}
+	slotLengthNanoseconds := uint64(params.SlotLength)
```
arepala-uml, line 61: gosec G115 fails the `lint` check on this head (`ledger/hardfork_summary.go:61:33: G115: integer overflow conversion int64 -> uint64 (gosec)`, reproduced locally with `golangci-lint run ./ledger/...` using the repo's own `.golangci.yml`). `params.SlotLength` is a `time.Duration` (`int64`); this conversion has no `#nosec` annotation. The `#nosec G115` comment further down (before the `return` in `hardForkCachedEpochDuration`) only covers that one statement, not this one. Please add a justified `#nosec G115` (or restructure to avoid the bare conversion) here.

ledger/hardfork_summary.go
```diff
@@ -103,10 +163,27 @@ func (ls *LedgerState) hardForkSummaryAnchoredAt(
...
+			if err != nil {
+				return nil, fmt.Errorf(
+					"ledger: cached epoch %d (era %d) params invalid: %w",
+					ep.EpochId,
+					eraID,
+					err,
+				)
+			}
+			if uint64(relTime) >
+				maxHardForkDurationNanoseconds-uint64(epochDuration) {
```
arepala-uml, line 179: Second gosec G115 failing `lint` on this head (`ledger/hardfork_summary.go:179:42`), same issue as line 61: `uint64(epochDuration)` converts a `time.Duration` (`int64`) with no `#nosec` annotation covering this line.

ledger/hardfork_summary.go
```diff
@@ -16,11 +16,61 @@ package ledger
...
+		return hardfork.EraParams{}, fmt.Errorf(
+			"slot length %dms overflows time.Duration",
+			slotLengthMilliseconds,
+		)
+	}
+	params := hardfork.EraParams{
+		EpochSize:  uint64(lengthInSlots),
+		SlotLength: time.Duration(slotLengthMilliseconds) * time.Millisecond,
+	}
+	if err := params.Validate(); err != nil {
```
arepala-uml, line 44: This new `Validate()` call (which rejects `SlotLength == 0`, per `ledger/hardfork/params.go:55-57`) is what breaks ~20 currently-passing tests elsewhere in this package on this head - reproduced locally with `go test ./ledger/...` at the exact PR commit, matching the CI failures verbatim (e.g. `TestVerifyBlockHeaderCrypto_RejectsBlockOutsideKnownEpochs`, `TestValidateChainSelectionHeaderCryptoAcceptsVerifiedHeader`, `TestVerifyDeferredBlockHeaderState_GenesisOverlayRevalidatedAtApply`, and more, all in `ledger/verify_header_test.go` and `ledger/verify_chain_selection_header_crypto_test.go`). Those tests build `epochCache` entries with `SlotLength` left at its zero value because they're exercising header-verification/chain-selection logic, not timing, and previously that was accepted (the removed code just computed a zero `time.Duration` and moved on). This PR's own new tests in `hardfork_summary_test.go` (`TestHardForkSummary_ValidatesCachedEraParams`, case `zero slot length`) deliberately assert this is now an error, but the change wasn't propagated to the other call sites/fixtures that relied on the old permissive behavior, so it regresses them instead. Either the validation needs to tolerate the zero-`SlotLength` convention the rest of the package still uses for non-timing tests (or those call sites need production-side handling that doesn't route through this stricter path), or every affected fixture needs updating in this PR - right now `go test ./ledger/...` fails on this head for reasons unrelated to the overflow bug this PR is fixing.

arepala-uml (changes requested):
CI is red on this head across all four go-test matrix jobs (Linux, Linux race, Windows, macOS) and lint. Reproduced locally against an isolated worktree at the exact PR head:

- `golangci-lint run ./ledger/...` reproduces the two gosec findings below verbatim (same file, line, column as the lint job).
- `go test ./ledger/...` reproduces ~20 pre-existing test failures in the `ledger` package, all with the identical root cause: `hardForkCachedEraParams` now calls `EraParams.Validate()`, which rejects `SlotLength == 0`. A number of existing header-verification/chain-selection test fixtures across the package (e.g. `verify_header_test.go`, `verify_chain_selection_header_crypto_test.go`) build epoch-cache entries without setting `SlotLength`, because those tests aren't exercising timing at all. They now fail with `ledger: cached epoch N (era 0) params invalid: hardfork: EraParams.SlotLength must be > 0` instead of exercising the behavior under test. This isn't a flaky/pre-existing failure: `origin/main` at the PR's merge base is green, and the failures disappear when the new `Validate()` call is removed. See inline comments for the two lint findings and the regression source.

Please either loosen the new validation so it doesn't reject the zero-`SlotLength` convention this package's tests (and possibly other legitimate callers) already rely on, or update every affected call site/fixture in this same PR, and fix the two gosec findings. Once `go-test` and `lint` are green on this head I'm happy to re-review.

ledger/hardfork_summary.go
wolf31o2: Fixed in f0406ded: the EraParams.Validate call is removed from hardForkCachedEraParams. A cached epoch row carries a zero EpochSize and SlotLength before the epoch record is populated - epochRollover uses `currentEpoch.SlotLength == 0` as exactly that sentinel - and hardfork.BuildSummary already validates the current era's params where valid parameters are required, so the call only added a refusal on the path that returns early without BuildSummary. hardForkCachedEpochDuration now treats a zero slot length as a zero-duration epoch, which is what the unbounded arithmetic produced and which also removes the zero divisor: without that guard the same fixtures panic with an integer divide by zero rather than erroring. The two tests asserting the removed strictness are replaced by ones pinning the tolerant contract. `go test ./ledger/` passes at head 1fdfe6fd.

ledger/hardfork_summary.go
wolf31o2: Fixed in f0406ded: the epoch duration is computed from the unsigned millisecond input, `uint64(slotLengthMilliseconds) * uint64(time.Millisecond)`, instead of converting params.SlotLength, so there is no int64 to uint64 conversion on this line. `golangci-lint run ./ledger/...` reports 0 issues at 1fdfe6fd.

ledger/hardfork_summary.go
wolf31o2: Fixed in f0406ded: the cumulative check is now `epochDuration > maxHardForkDuration-relTime`, entirely in the time.Duration domain, so neither operand is converted. The adjacent `uint64(relTime)` that gosec did not flag is gone with it.

arepala-uml (approved):

PR #4101, opened 2026-09-07
fix(ledger): fail closed on validation and forging configuration

### LINKED ISSUE #3528: ledger: fail closed on validation and forging configuration
Ledger and forging validation can be weakened by missing or inconsistent configuration:

- Historical synchronization disables an entire group of header checks.
- Approximate in-loop accounting can still produce an exact body above the configured block-body limit.
- Negative KES evolution values can pass after conversion to an unsigned type.
- Phase-2 validation can be skipped outside a narrowly defined recovery window.
- A missing cost model can reach evaluation instead of returning a configuration error.
- Loaded genesis network magic is not cross-checked against the requested network.

Optional forged-block self-validation is an operator-selected diagnostic against builder bugs. Its default-off setting alone is not a defect or a requirement to add work to the forge-to-diffusion path.

Acceptance criteria:

- Define and enforce the required validation set for historical sync and preserve the block builder's consensus invariants.
- Enforce exact block-body limits before adoption or diffusion.
- Validate KES, phase-2, cost-model, and network settings before use.
- Add negative configuration, historical-sync, and exact-size boundary tests.

arepala-uml:
## Summary
- Header VRF/KES/OpCert crypto is now required at chainsync admission and blockfetch time regardless of `ValidateHistorical` (previously skipped entirely during ordinary bulk sync); only a Mithril-certified slot remains exempt.
- Forged-block self-validation (`validateForgedBlock`) now defaults to enabled instead of disabled.
- A missing Plutus cost model now returns a configuration error instead of silently falling back to plutigo's built-in default.
- Missing KES genesis parameters (`slotsPerKESPeriod`/`maxKESEvolutions`) now fail closed instead of falling back to a weaker expiry check.
- The phase-2 evaluation skip for deep historical blocks now requires the explicit `TrustedReplay` opt-in, not just `ValidateHistorical=false`.
- The Byron genesis's network magic is now cross-checked against the configured network, matching the existing Shelley check.
- The exact-vs-approximate block-body size protection was already correct; added a boundary regression test.

Fixes #3528

## Test plan
- [x] `go build ./...`
- [x] `go test ./...`
- [x] `go test -race ./...`
- [x] `golangci-lint run ./...` (0 issues)
- [x] `nilaway ./...` (same pre-existing finding count, none new)
- [x] `modernize ./...`, `golines` (clean)
- [x] `go test ./internal/docsparity/... ./internal/architecture/...`
- [x] New/updated regression tests cover each fix, including a `node_lifecycle_test.go` Genesis-fork-selection scenario driven by genuinely VRF/KES-signed headers

---
## Summary by cubic
Fixes #3528 by closing validation paths that could admit unverified data or silently weaken checks.

**Bug Fixes**
- Chainsync admission, blockfetch, and chain selection now verify VRF/KES/OpCert crypto regardless of `ValidateHistorical`; Mithril-covered slots remain exempt, and an unavailable epoch nonce defers verification instead of rejecting peers.
- Plutus phase-2 evaluation always runs when per-transaction validation runs; `ValidateHistorical=false` nodes now pay evaluation on deep historical blocks where it was previously skipped.
- Missing or empty Plutus cost models, Byron network-magic mismatches, and missing or non-positive KES genesis parameters now return configuration errors, with `maxKESEvolutions` validated at config load; shorter-than-current cost models remain accepted as correct for their era.
- The forging loop now excludes a transaction whose exact assembled body exceeds `MaxBlockBodySize` during selection instead of failing the whole build afterward.

**Migration**
- Forged-block self-validation is enabled by default; set `validateForgedBlock: false` only to opt out. Aggregate reference-script budget checks always run regardless.

<sup>Written for commit 7db2d5f6af800f6e19aff482f60848d7080fd93c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Forged-block self-validation is now enabled by default, with an explicit opt-out.
  * Configuration validation now checks Byron genesis network magic against the configured network.
  * Script evaluation now requires declared Plutus cost models and fails closed when they are missing.

* **Bug Fixes**
  * Historical synchronization now performs header cryptographic verification except for Mithril-certified data.
  * Missing KES parameters now produce configuration errors instead of bypassing validation.
  * Block construction correctly rejects bodies exceeding the maximum size.


ledger/verify_header.go
```diff
@@ -161,16 +161,17 @@ func (ls *LedgerState) ValidateBlockHeaderCrypto(
 
 // ShouldVerifyChainSelectionHeaderCrypto reports whether a header at the
 // given slot is eligible to have its cryptography verified right now via
-// ValidateChainSelectionHeaderCrypto. It mirrors the same exemptions the
+// ValidateChainSelectionHeaderCrypto. It mirrors the same exemption the
 // ledger's own chainsync header-queue path already applies
-// (shouldVerifyChainsyncHeaderCrypto): verification is skipped while bulk
-// historical/catch-up loading has not yet enabled live validation, and for
+// (shouldVerifyChainsyncHeaderCrypto): verification is skipped only for
```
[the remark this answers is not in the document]
arepala-uml: Fixed in 2c565180: `ShouldVerifyChainSelectionHeaderCrypto` no longer delegates to the chainsync admission gate. It now checks only the Mithril exemption directly and always attempts verification otherwise, letting `ValidateChainSelectionHeaderCrypto`'s existing deferred-error handling (already safe to call unconditionally, per `allowEpochCacheAdvance=false`) decide when the epoch nonce isn't cached yet. Added `TestShouldVerifyChainSelectionHeaderCryptoIgnoresMissingNonce`, confirmed to fail against the prior delegating implementation.

config.go
```diff
@@ -1457,11 +1473,12 @@ func WithForgeStaleGapThresholdSlots(slots uint64) ConfigOptionFunc {
 	}
 }
 
-// WithValidateForgedBlock enables self-validation of locally-forged blocks
+// WithValidateForgedBlock controls self-validation of locally-forged blocks
 // before they are adopted onto the chain and diffused to peers. When enabled,
 // the forger runs VRF/KES header crypto, body-hash consistency, and per-tx
 // ledger validation on each forged block. A failing block is dropped without
-// being adopted or diffused. Disabled by default.
+// being adopted or diffused. Enabled by default; pass false only to
```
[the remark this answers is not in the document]
arepala-uml: Fixed in 2c565180: added `ValidateForgedBlock: true` to `NewConfig`'s internal config literal, which is a separate default source from `internalconfig.DefaultConfig` and hadn't picked up the new default. Added `TestNewConfigDefaultsValidateForgedBlock` regression test in config_test.go.

node_lifecycle_test.go
```diff
@@ -600,6 +606,143 @@ func registerGenesisForkTestPeer(
 	return oConn.Id()
 }
 
+// forkChainEmptyConwayBodyHash returns the block body hash for a Conway
+// block with empty transaction components, matching
+// internal/test/testutil's conwayEmptyBodyHash. It's independent of the
+// header, so every generated fork block shares it.
+func forkChainEmptyConwayBodyHash(t *testing.T) lcommon.Blake2b256 {
```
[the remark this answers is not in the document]
arepala-uml: Fixed in 2c565180 by exporting `testutil.ConwayEmptyBodyHash` and using it here instead of duplicating it as `forkChainEmptyConwayBodyHash`. The VRF/KES/cold key setup and header/OpCert construction is a genuinely different algorithm from testutil's `buildValidatedConwayBlockBytes` (no leadership-threshold search, since header admission skips stake-pool validation per `VerifyConfig.SkipStakePoolValidation`), so that part is not shared, but the body-hash computation now is.

ledger/state.go
```diff
@@ -6534,6 +6543,7 @@ func (ls *LedgerState) ledgerProcessBlocksFromSource(
 						// Process block
 						skipPhase2Validation := shouldSkipConfiguredPhase2Validation(
 							snapshotValidationEnabled,
+							ls.config.TrustedReplay,
```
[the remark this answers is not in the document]
arepala-uml: Confirmed and fixed in 2c565180. `historicalBlockValidationDecision` forces `shouldValidateBlock` to false whenever `TrustedReplay` is true, so the 4-way AND in `shouldSkipConfiguredPhase2Validation` could never be satisfied at this call site — removed the dead mechanism (shouldSkipPhase2ValidationForBlock, shouldSkipPhase2ValidationForBlockAtCurrentTip, shouldSkipConfiguredPhase2Validation) rather than reworking it, since making it reachable would require changing TrustedReplay's own all-or-nothing validation skip, which is unrelated to this issue. Phase 2 now always evaluates whenever per-tx validation runs at all.

ledger/eras/validation_test.go
```diff
@@ -385,6 +385,7 @@ func TestPlutusBudgetComparisonIncludesFinalSlippageBatch(t *testing.T) {
 					&alonzo.AlonzoProtocolParameters{
 						ProtocolMajor: 5,
 						MaxTxExUnits:  maxTxExUnits,
+						CostModels:    map[uint][]int64{0: {}},
```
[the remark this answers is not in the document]
arepala-uml: Addressed in 2c565180 with a comment explaining the deliberate fallback dependence, rather than switching to real 332-param cost model tables: this test is specifically about slippage-batch budget accounting, not cost-model realism, and the pinned 112100/800 assertions are unaffected either way. Noted that these numbers would need updating if plutigo's default table ever changes.

ledger/verify_header.go
```diff
@@ -161,20 +161,29 @@ func (ls *LedgerState) ValidateBlockHeaderCrypto(
...
+// returns false for is never re-verified. ValidateChainSelectionHeaderCrypto
+// already handles a not-yet-available epoch by returning a deferred error
+// (see IsHeaderVerificationDeferred) rather than failing, so it's always
+// safe to attempt verification here and let the verifier decide.
 func (ls *LedgerState) ShouldVerifyChainSelectionHeaderCrypto(
 	slot uint64,
 ) bool {
-	return ls.shouldVerifyChainsyncHeaderCrypto(slot)
+	mithrilLedgerSlot := ls.mithrilLedgerSlotSnapshot()
+	return mithrilLedgerSlot == 0 || slot > mithrilLedgerSlot
```
wolf31o2: changes requested: when a non-Mithril header falls in a cached post-Byron epoch whose nonce is temporarily empty, this path returns success and sends the header to chain-selection verification. That verifier returns non-deferred errEpochNonceUnavailable, so an honest header is treated as invalid and the peer is recycled. Classify this transient nonce-unavailable case as deferred.

ledger/eras/cost_models.go
```diff
@@ -14,6 +14,31 @@
...
+// a script under the wrong (default, not this network's configured) cost
+// parameters instead of refusing to evaluate it at all. Callers must use
+// this instead of indexing CostModels directly.
+func requiredCostModel(
+	costModels map[uint][]int64,
+	key uint,
+	versionName string,
+) ([]int64, error) {
+	model, ok := costModels[key]
+	if !ok {
```
wolf31o2: changes requested: requiredCostModel accepts an undersized retrieved list and lets costModelFromList fill missing entries. Validate the list against the complete protocol-required shape for the Plutus version, rejecting nil, empty, and undersized models while preserving complete valid models.

ledger/eras/cost_models_test.go
```diff
@@ -21,6 +21,33 @@ import (
...
+	})
+	t.Run("present key returns its model", func(t *testing.T) {
+		model := []int64{1, 2, 3}
+		got, err := requiredCostModel(
+			map[uint][]int64{1: model},
+			1,
+			"PlutusV2",
+		)
+		require.NoError(t, err)
+		require.Equal(t, model, got)
```
wolf31o2, line 84: changes requested: the successful requiredCostModel fixture is only three entries, so it cannot prove that a complete protocol-sized model is accepted. Use a complete 332-entry fixture for the success case and retain the three-entry fixture as an explicit undersized-value rejection case.

wolf31o2 (changes requested):
changes requested

ledger/verify_header.go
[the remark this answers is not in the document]
arepala-uml: Confirmed and fixed in 3485b346: `IsHeaderVerificationDeferred` now also recognizes `errEpochNonceUnavailable`, matching how `errBlockPipelineEta0Unavailable` (which wraps the same sentinel) is already treated as non-fatal on the block-pipeline path. Added `TestValidateChainSelectionHeaderCryptoDefersOnUnpublishedNonce`, confirmed to fail against the prior classification.

ledger/eras/cost_models_test.go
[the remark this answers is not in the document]
arepala-uml: Fixed in 3485b346: added a separate present-but-undersized rejection case (3-entry model) and updated the success case to use a complete model via a new `syntheticFullCostModel` helper.

ledger/eras/cost_models.go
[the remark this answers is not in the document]
arepala-uml: Fixed in 3485b346: `requiredCostModel` now rejects a present-but-undersized cost model (checked against `lang.GetParamNamesForVersion`'s required length for the version), not just a missing key. Updated the empty-cost-model fixtures in conway_plutus_budget_test.go, conway_plutus_script_failure_test.go, plutus_validity_outcome_test.go, and validation_test.go to use complete models — most via a new `defaultMachineCostModel` test helper that reproduces plutigo's real `DefaultMachineCosts` values, so the tests pinning specific 112100 CPU / 800 memory numbers keep those externally-verified assertions instead of drifting to arbitrary placeholder-derived numbers.

ledger/verify_header.go
arepala-uml: Fixed in 3485b346: `IsHeaderVerificationDeferred` now also recognizes `errEpochNonceUnavailable` (the sentinel a covered-but-not-yet-published-nonce epoch returns), matching how `errBlockPipelineEta0Unavailable`, which wraps the same sentinel, is already treated as non-fatal on the block-pipeline path. Added `TestValidateChainSelectionHeaderCryptoDefersOnUnpublishedNonce`, confirmed to fail against the prior classification.

ledger/eras/cost_models.go
arepala-uml: Fixed in 3485b346: `requiredCostModel` now checks the retrieved list's length against `lang.GetParamNamesForVersion`'s required parameter count for the version, rejecting nil, empty, and undersized lists -- not just a missing key. Updated the affected test fixtures (conway_plutus_budget_test.go, conway_plutus_script_failure_test.go, plutus_validity_outcome_test.go, validation_test.go) to supply complete models.

ledger/eras/cost_models_test.go
arepala-uml: Fixed in 3485b346: added a separate present-but-undersized rejection case (the original 3-entry model) and a new `syntheticFullCostModel` helper supplying a complete, correctly-sized model for the success case.

ledger/eras/cost_models.go
```diff
@@ -14,6 +14,59 @@
...
+		return nil, fmt.Errorf(
+			"missing %s cost model in protocol parameters",
+			versionName,
+		)
+	}
+	if version, versionOk := costModelKeyVersions[key]; versionOk {
+		want := len(lang.GetParamNamesForVersion(version))
+		if len(model) < want {
```
chrisguiney: `lang.GetParamNamesForVersion` returns plutigo's full parameter table: 332 for V1 and V2, 350 for V3. `TestCostModelParameterTablesMatchLiveNetworks` in this same file documents those as the lengths mainnet, preprod and preview publish **at protocol version 11**. They are not the lengths in force before that, so using them as a floor at every protocol version rejects valid historical parameters.

Two concrete cases, both from values this repository ships:

- `HardForkAlonzo` (`ledger/eras/alonzo.go:110`) fills `CostModels[0]` from the Alonzo genesis via `UpdateFromGenesis`. `config/cardano/mainnet/alonzo-genesis.json`, `.../preprod/...` and `.../preview/...` each carry `costModels.PlutusV1` with 166 entries, so `requiredCostModel(pp.CostModels, 0, "PlutusV1")` returns `PlutusV1 cost model in protocol parameters has 166 parameters, want at least 332`.
- `HardForkBabbage` (`ledger/eras/babbage.go:145-146`) injects `DefaultPlutusV2CostModel` when the chain has not yet delivered an on-chain V2 model. That constant is 175 entries and is rejected the same way, even though `LedgerStateUpdate.RealV2CostModelObserved`'s comment in `ledger/state.go` describes it as the real canonical mainnet value.

The error propagates out of `ValidateTxAlonzo`, `ValidateTxBabbage` and `evaluateConwayPlutusScript`, and at `ledger/state.go:7179` the standard profile leaves a `ValidateTxFunc` error intact and rejects the block. A sync from genesis therefore stops at the first Alonzo- or Babbage-era block carrying a PlutusV1 script transaction, and at the first PlutusV2 one before the on-chain V2 model lands.

The presence check above it is sound: a missing key genuinely does let `costModelFromList` substitute plutigo's built-in defaults, which is the fallback #3528 asks to close. It is the length floor that needs a protocol-version-aware bound rather than the PV11 one.

A case that would catch this: decode `config/cardano/mainnet/alonzo-genesis.json` through `alonzo.AlonzoProtocolParameters.UpdateFromGenesis` and assert `requiredCostModel(pp.CostModels, 0, "PlutusV1")` returns no error. `syntheticFullCostModel` pads to the table length, so it cannot reach this shape.

ledger/chainsync.go
```diff
@@ -4055,11 +4058,13 @@ func (ls *LedgerState) handleEventBlockfetchBlockDeferred(
...
-	// historical blocks were already validated by the network.
-	validationEnabled, _ := ls.validationStateSnapshot()
-	if validationEnabled {
+	// Verify block header cryptographic proofs (VRF, KES). Required for
+	// every slot except one a Mithril certificate already covers -- see
+	// chainsyncHeaderCryptoPolicy's doc comment. A coarse
+	// ValidateHistorical=false historical-sync toggle must not disable
+	// this entire group of checks (issue #3528).
+	mithrilLedgerSlot := ls.mithrilLedgerSlotSnapshot()
+	if mithrilLedgerSlot == 0 || e.Point.Slot > mithrilLedgerSlot {
```
chrisguiney: `IsHeaderVerificationDeferred` now classifies `errEpochNonceUnavailable` as "not ready, not proof of invalidity", but the two internal gates that consume the same condition, `ledger/chainsync.go:3208` and `:4096` a few lines below here, still test `errors.Is(..., errHeaderVerificationDeferred)` alone.

Replacing the `validationEnabled` gate with this Mithril-only one makes the path run for every non-Mithril slot rather than only when live validation is on, so a cached epoch entry that covers the slot but carries no published nonce can now reach `verifyBlockHeaderCryptoBeforeApply` and come back as `block header crypto verification failed` rather than deferring. The flush-and-recheck below only re-tests `FirstVerifiedHeaderMatchesPoint`, not nonce availability, so it does not close that window.

Is the narrower classification deliberate at those two sites, or should they use the exported predicate now that it is the definition of the deferrable set? A short note either way would keep the two spellings from drifting.

ledger/state.go
```diff
@@ -4542,49 +4542,18 @@ func (ls *LedgerState) securityParamForCurrentEraSnapshot() int {
...
-	shouldValidateBlock bool,
-	deepHistoricalBlock bool,
-) bool {
-	return !validationEnabled && shouldValidateBlock && deepHistoricalBlock
-}
+// Issue #3528: the historical-sync phase-2 shortcut that used to live here
+// (shouldSkipPhase2ValidationForBlock, shouldSkipPhase2ValidationForBlockAtCurrentTip,
+// shouldSkipConfiguredPhase2Validation) was meant to skip re-running Plutus
+// evaluation only for a deep, already-immutable block during a deliberate
+// TrustedReplay import. It was provably unreachable: historicalBlockValidationDecision
```
chrisguiney: The reachability claim does not hold. `shouldSkipConfiguredPhase2Validation` never tested `TrustedReplay`; it tested `!validationEnabled && shouldValidateBlock && deepHistoricalBlock`. In `historicalBlockValidationDecision`, the `validationEnabled == false` branch returns `shouldValidate == true` whenever `trustedReplay` is false, the state is `SyncingChainsyncState`, the slot is at or past the cutoff, and the slot is not Mithril-covered. That is exactly the `ValidateHistorical=false` catch-up case the shortcut was written for, and the call site passed `snapshotValidationEnabled` and the `shouldValidateBlock` from that same helper, so the combination was reachable.

Deleting it is still the direction #3528 asks for and phase 2 always running is the safer contract, so this is not a request to restore it. But the note records a reason that is not true, and the PR description's "now requires the explicit `TrustedReplay` opt-in" describes behaviour the diff does not implement, since the shortcut is removed rather than narrowed. The real trade is worth stating in its place: an operator running `ValidateHistorical=false` now pays Plutus phase-2 evaluation on every deep historical block.

ARCHITECTURE.md
```diff
@@ -4781,7 +4781,7 @@ When Dijkstra/Leios is active, `DefaultBlockBuilder` emits the Musashi prototype
...
+Unless `validateForgedBlock` is explicitly disabled in config, the forger invokes `LedgerState.ValidateForgedBlock` between steps 5 and 7. This runs three checks: (a) VRF proof and KES signature verification of the block header, (b) body-hash non-zero guard, and (c) per-transaction ledger rule validation against the current UTxO state with an intra-block overlay so outputs created by earlier transactions in the same block are visible to later ones. A failing block is logged, counted in `dingo_forge_validation_failed_total`, and dropped without being adopted or diffused. Validation wall-clock time is recorded in the `dingo_forge_validation_duration_seconds` histogram. Enabled by default (fail closed); an operator may set `validateForgedBlock: false` to skip it and accept the risk in exchange for lower forge-to-diffusion latency.
```
chrisguiney: The section heading two lines up still reads "Optional Self-Validation", which now disagrees with this paragraph's "Enabled by default (fail closed)". Worth retitling so a reader skimming headings does not take the old default away with them.

chrisguiney (changes requested):
Changes requested.

The new Plutus cost-model length floor rejects the cost models this repository itself ships for the Alonzo and Babbage eras, so replay through those eras fails on any block carrying a PlutusV1 or PlutusV2 script transaction. Details inline on `ledger/eras/cost_models.go`.

The package tests pass because every cost-model fixture is padded to plutigo's parameter-table length; none feeds a real shipped genesis model through `requiredCostModel`, so a green suite is not coverage for this case.

Against the earlier `changes requested` review, at this head: the `errEpochNonceUnavailable` deferral in `ledger/verify_header.go` is addressed; the cost-model shape validation is implemented, but with a bound that does not hold before protocol version 11; the complete-fixture test was added, but synthetically, so the real-network shape is still uncovered.

The rest of the change reads well. The Mithril-only header-crypto exemption, the Byron network-magic cross-check, the `ValidateForgedBlock` default in both `internalconfig.DefaultConfig` and `NewConfig`, and the fail-closed KES genesis parameters all hold up; `slotsPerKESPeriod` and `maxKESEvolutions` come from the same `ShelleyGenesis()`, and the pre-existing `slotsPerKesPeriod == 0` guard already fires first, so no valid genesis reaches the new error.

ledger/eras/cost_models.go
arepala-uml: Confirmed with real data and fixed in a436affa: reverted the length floor to a presence check (`!ok || len(model) == 0`), matching gouroboros' own `UtxoValidateCostModelsPresent` rule. Added `TestRequiredCostModelAcceptsRealAlonzoGenesisModel`, which decodes `config/cardano/mainnet/alonzo-genesis.json` through `AlonzoProtocolParameters.UpdateFromGenesis` (the real production loading path) and asserts the resulting 166-entry model is accepted -- exactly the case you flagged. Thank you for catching this with concrete shipped data; a synthetic fixture would never have surfaced it.

ledger/chainsync.go
arepala-uml: Not deliberate -- a real gap, thank you. Fixed in a436affa: both call sites (chainsync.go:3208 admission, and this one at blockfetch time) now use `IsHeaderVerificationDeferred` instead of the raw sentinel check. The blockfetch path is the one that actually reaches this in practice, since it doesn't gate on nonce availability before calling verify (only flushes-and-rechecks once) -- added `TestBlockfetchHeaderVerificationEmptyEpochNonceDefersNotFails`, confirmed to fail against the prior raw check with exactly the failure mode you described.

ledger/state.go
arepala-uml: Correct, and thank you for the precise correction -- fixed the comment in a436affa. The original 3-way condition (no TrustedReplay test) was genuinely reachable via ordinary ValidateHistorical=false catch-up; it was my own later revision adding a trustedReplay requirement that made that specific 4-way combination unreachable, and that dead form is what got deleted. Rewrote the comment to state the real trade-off plainly: ValidateHistorical=false nodes now pay Plutus phase-2 evaluation on every deep historical block where it was previously skipped. Not proposing to restore the old shortcut, per your note.

ARCHITECTURE.md
arepala-uml: Fixed in a436affa: retitled to "Self-Validation", dropping "Optional".

ledger/eras/cost_models.go
arepala-uml: Follow-up: the length-floor check added here turned out to be unsafe -- a human reviewer (chrisguiney) found it rejects the real cost models this repository itself ships (mainnet/preprod/preview Alonzo genesis's 166-entry PlutusV1 model, and DefaultPlutusV2CostModel's 175 entries), since the parameter count has grown across hard forks and a shorter list is the correct, complete shape for an earlier protocol version, not a truncated one. Reverted in a436affa to a presence check (`!ok || len(model) == 0`), matching gouroboros' own `UtxoValidateCostModelsPresent` rule, with a regression test against the real shipped genesis file. Apologies for the churn -- flagging here so this thread's history reads consistently with the final state.

ledger/eras/cost_models_test.go
arepala-uml: Follow-up: the undersized-rejection behavior this test was checking turned out to be unsafe (see the linked reply on cost_models.go) and has been reverted in a436affa. The three-entry model is now a valid "short but non-empty" case that must be *accepted*, not rejected -- real historical cost models (e.g. Alonzo's 166-entry PlutusV1) are legitimately far shorter than the current parameter table. Updated the test accordingly and added `TestRequiredCostModelAcceptsRealAlonzoGenesisModel` against the real mainnet genesis file. Apologies for the churn.

ledger/eras/cost_models.go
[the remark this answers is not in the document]
arepala-uml: Follow-up: a human reviewer found the length-floor fix here was itself unsafe -- it rejected real cost models this repo ships (mainnet/preprod/preview Alonzo genesis's 166-entry PlutusV1 model). Reverted in a436affa to a presence check (`!ok || len(model) == 0`), matching gouroboros' own `UtxoValidateCostModelsPresent` rule, with a regression test against the real shipped genesis file (`TestRequiredCostModelAcceptsRealAlonzoGenesisModel`).

ledger/eras/cost_models_test.go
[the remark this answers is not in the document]
arepala-uml: Follow-up: the underlying check this test validated has been reverted (see the linked reply on cost_models.go) -- the length floor against plutigo's current parameter table rejected real historical cost models this repo ships. The three-entry model is now correctly accepted (short-but-non-empty is valid), and a new test decodes the real mainnet Alonzo genesis to prove the actual production case works.

arepala-uml:
Addressed both "Changes Requested" reviews (commit a436affa):

- **Reverted the cost-model length floor** to a presence check (`!ok || len(model) == 0`). The length floor against plutigo's current (PV11) parameter table rejected the cost models this repo itself ships — mainnet/preprod/preview Alonzo genesis's real 166-entry PlutusV1 model, and `DefaultPlutusV2CostModel`'s 175 entries — since the parameter count has grown across hard forks and a shorter list is the correct shape for an earlier protocol version. Added a regression test that decodes the real mainnet Alonzo genesis through the actual production loading path (`AlonzoProtocolParameters.UpdateFromGenesis`) and asserts it's accepted.
- **Fixed the two internal `errors.Is(err, errHeaderVerificationDeferred)` checks** in `ledger/chainsync.go` (admission and blockfetch-time) to use the exported `IsHeaderVerificationDeferred` instead, so they get the `errEpochNonceUnavailable` classification too. Added a regression test confirmed to fail against the prior raw check.
- **Corrected a misleading comment** in `ledger/state.go` about why the phase-2 shortcut was removed — it was reachable via ordinary `ValidateHistorical=false` catch-up (not `TrustedReplay`); a later revision of mine made a specific 4-way combination unreachable, and that's what was actually deleted. Restated the real trade-off: `ValidateHistorical=false` nodes now pay Plutus phase-2 evaluation on every deep historical block.
- **Retitled** an ARCHITECTURE.md heading that disagreed with the paragraph beneath it.

Full suite (build, tests, `-race`, lint, nilaway, docs-parity) is clean. Requesting re-review from @wolf31o2 and @chrisguiney.

ledger/eras/cost_models.go
```diff
@@ -14,6 +14,48 @@
...
+// transaction). Matches the equivalent presence check gouroboros' own
+// UtxoValidateCostModelsPresent rule already performs
+// (!ok || len(model) == 0) elsewhere in the validation pipeline.
+func requiredCostModel(
+	costModels map[uint][]int64,
+	key uint,
+	versionName string,
+) ([]int64, error) {
+	model, ok := costModels[key]
+	if !ok || len(model) == 0 {
```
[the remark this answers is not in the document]
arepala-uml: Already addressed, and this is the same shape of concern wolf31o2 raised, which turned out to be unsafe when checked against real data.

The presence check here is deliberate, not incomplete: a stricter length floor against `lang.GetParamNamesForVersion` (plutigo's current, protocol-version-11 parameter table) was tried first, then reverted in a436affa after chrisguiney found it rejects the real cost models this repo itself ships -- `config/cardano/{mainnet,preprod,preview}/alonzo-genesis.json` each carry a 166-entry PlutusV1 model (Alonzo's actual, complete-for-its-era param count), far short of the current 332-entry table. A length floor against the current table would stall replay from genesis at the first Alonzo Plutus transaction.

`TestRequiredCostModelAcceptsRealAlonzoGenesisModel` in `ledger/eras/cost_models_test.go` decodes that real genesis file through the production `UpdateFromGenesis` path and asserts it's accepted -- proving a stricter check would break real functionality, not just theoretically risk it.

On "silently filling omitted parameters from built-in defaults": that's exactly what this function exists to prevent for the one case it can safely detect -- a missing or empty entry (`!ok || len(model) == 0`), matching gouroboros' own `UtxoValidateCostModelsPresent` rule. Validating the *complete* per-era shape (as opposed to just non-empty) would require a historical-shape table this codebase doesn't have and plutigo doesn't expose per protocol version -- `lang.GetParamNamesForVersion` is protocol-version-blind and always returns the current table, which is precisely the check that broke.

See the doc comment on `requiredCostModel` (lines 28-43) for the full rationale.

chrisguiney, line 50: Resolved. The presence check accepts the cost models this repository ships: `config/cardano/{mainnet,preprod,preview}/alonzo-genesis.json` each carry a 166-entry PlutusV1 model, and `ledger/eras/babbage.go`'s `DefaultPlutusV2CostModel` is 175 entries — both far short of plutigo's current 332-entry table, and both correct and complete for their era. `TestRequiredCostModelAcceptsRealAlonzoGenesisModel` pins that through `AlonzoProtocolParameters.UpdateFromGenesis`, the production loading path, so the floor cannot be reintroduced without a failing test.

The residual — a genuinely truncated model still being padded by plutigo's `costModelFromList` — would need a protocol-version-to-length table that does not exist in this repository, and is a different change. Presence plus non-empty is the same bound gouroboros' own `UtxoValidateCostModelsPresent` applies.

ledger/verify_header.go
```diff
@@ -101,9 +101,18 @@ var (
...
+// epoch transiently) is the same "not ready, not proof of invalidity"
+// condition as a slot outside the published cache entirely -- treating it as
+// a hard failure would let a chain-selection or Leios-announcement caller
+// wrongly reject and recycle an honest peer over a transient local gap
+// instead of retrying once the nonce is published (matches how
+// errBlockPipelineEta0Unavailable, which wraps the same sentinel, is already
+// treated as non-fatal on the block-pipeline path).
 func IsHeaderVerificationDeferred(err error) bool {
-	return errors.Is(err, errHeaderVerificationDeferred)
+	return errors.Is(err, errHeaderVerificationDeferred) ||
```
chrisguiney, line 114: Resolved: both internal gates now route through this predicate (`ledger/chainsync.go:3208`, `:4096`).

Widening it also widens the three consumers outside this package, and they hold. `ouroboros/chainsync.go:1084` leaves a deferred header eligible for chain selection instead of publishing `ConnectionRecycleRequestedEventType`, which is the intended effect. `ouroboros/leiosnotify.go:1546` and `:1624` keep the announcement in `leiosDeferredAnnouncements`, which is capped at `leiosMaxDeferredAnnouncements = 128` and never relays before validation succeeds, so the new error class cannot be used by a peer to grow that queue.

`recordBlockPipelineError` (`ledger/state.go:5052`) still switches on the raw sentinels rather than this helper, but the pipeline's only producer of `errEpochNonceUnavailable` is `blockPipelineEta0Provider` (`ledger/verify_header.go:1975`), which always wraps it in `errBlockPipelineEta0Unavailable`. Its classification is therefore unchanged, and the same holds for the pair test at `ledger/state.go:5213`.

ledger/verify_header.go
```diff
@@ -161,20 +170,29 @@ func (ls *LedgerState) ValidateBlockHeaderCrypto(
...
+// available, but chain selection has no such retry -- a header this
+// returns false for is never re-verified. ValidateChainSelectionHeaderCrypto
+// already handles a not-yet-available epoch by returning a deferred error
+// (see IsHeaderVerificationDeferred) rather than failing, so it's always
+// safe to attempt verification here and let the verifier decide.
 func (ls *LedgerState) ShouldVerifyChainSelectionHeaderCrypto(
 	slot uint64,
 ) bool {
-	return ls.shouldVerifyChainsyncHeaderCrypto(slot)
+	mithrilLedgerSlot := ls.mithrilLedgerSlotSnapshot()
```
chrisguiney: Dropping the `validationEnabled` arm here leaves two statements elsewhere describing a contract this predicate no longer has:

- `ouroboros/chainsync.go:1072-1075` — "Verification is skipped under the same conditions the ledger's own header pipeline already skips it (bulk historical/catch-up loading, or a Mithril-covered slot), so fast sync and a Mithril-restored bootstrap are unaffected." Bulk historical/catch-up loading is no longer an exemption, and fast sync is no longer unaffected.
- `ARCHITECTURE.md:3611` — "Verification is skipped under the same fast-sync/Mithril-import exemptions the ledger's own header-queue path already applies."

That ARCHITECTURE.md paragraph is the documented contract for this function and the closure it feeds, so correcting both in this change keeps the pair honest.

chrisguiney (commented):
The findings from the earlier changes-requested review are resolved at this head, checked against the code rather than the replies. What remains in the diff is three stale contract statements the change makes wrong; none is a behavioral defect, and they are detailed inline.

Approval is withheld only because required checks are red on this head.

**`DCO` (action_required, 31 commits).** The branch carries 75 commits over its merge base, of which about 72 are a duplicate copy of `main`'s history under different SHAs, introduced by merge `b226c2e0`. Every flagged commit is another contributor's squash-merge that is already on `main`; DCO passes on every other currently open pull request, so this is specific to this branch's history rather than a repository-wide condition. `b226c2e0` and `f1290d33` also carry no `Signed-off-by` trailer of their own. The DCO app's suggested remediation is a rebase, which is not available on published history, so the resolution is a maintainer call.

**`go-test (macOS)`.** Not attributable to this change. The failure at this head is `TestDecodeReadChainBatchDoesNotDeadlockOnManyValidationErrors`, whose file `ledger/block_pipeline_errors_drain_test.go` is not in this diff. The failure on the previous head `a436affa` was a different test, `TestConcurrentBindStopNeverLeavesThePortBound` in `internal/apilistener`, a package this change does not touch at all. `go-test (Linux)`, `go-test (Linux, race)`, `go-test (Windows)`, `lint`, `govulncheck` and CodeQL are green on this head.

No bot reviewed this head. CodeRabbit reports "Review skipped: incremental reviews are disabled" and the `cubic` check reports it could not start the incremental review; their latest review records are at `a436affa`.

ledger/verify_opcert.go
```diff
@@ -281,19 +281,26 @@ func verifyOpCertHeaderCrypto(
...
-		if _, err := ledger.ValidateKesPeriod(
-			opCert.KesPeriod,
-			slot,
+	// KES expiry needs both genesis parameters. Unlike an absent opcert
+	// (handled above), a missing or zero genesis parameter here is a
+	// configuration failure, not a case to validate around: fail closed
+	// rather than falling back to the lighter future-cert-only KES guard
+	// VerifyKesComponents already ran inside VerifyBlock, which never
+	// rejects an expired opcert.
+	if slotsPerKesPeriod == 0 || maxKesEvolutions == 0 {
```
chrisguiney, line 290: The fail-closed change holds, and it is the mechanism that closes issue #3528's negative-KES bullet: `loadSlotsPerKESPeriod` (`ledger/state.go:10376`) and `maxKESEvolutions` (`ledger/verify_header.go:1899`) each clamp a negative genesis value to 0, which this now turns into a configuration error instead of an unsigned conversion that silently passes.

Neither call site can reach it with a valid genesis. `ledger/verify_header.go:413` rejects `slotsPerKesPeriod == 0` before calling. The pipeline path at `ledger/state.go:5251` is preceded by the `NewLedgerState` startup error at `ledger/state.go:1394`, which refuses to construct a validating pipeline without `slotsPerKESPeriod`. All three shipped Shelley genesis files carry 129600 and 62.

One stale statement is left behind: `maxKESEvolutions`' doc comment at `ledger/verify_header.go:1896-1898` still says "Returns 0 when the genesis is unavailable, in which case opcert KES-period expiry is left to the lighter future-cert guard inside VerifyBlock." That fallback is exactly what this block removes.

ledger/chainsync.go
```diff
@@ -4055,11 +4058,13 @@ func (ls *LedgerState) handleEventBlockfetchBlockDeferred(
...
-	// Skip during historical sync (validationEnabled=false) because
-	// historical blocks were already validated by the network.
-	validationEnabled, _ := ls.validationStateSnapshot()
-	if validationEnabled {
+	// Verify block header cryptographic proofs (VRF, KES). Required for
+	// every slot except one a Mithril certificate already covers -- see
+	// chainsyncHeaderCryptoPolicy's doc comment. A coarse
+	// ValidateHistorical=false historical-sync toggle must not disable
+	// this entire group of checks (issue #3528).
+	mithrilLedgerSlot := ls.mithrilLedgerSlotSnapshot()
```
chrisguiney: Non-blocking. This is now the body of `shouldEnforceBlockPipelineCrypto` (`:3662`) and of `ShouldVerifyChainSelectionHeaderCrypto` (`ledger/verify_header.go:194`) spelled out a third time. Calling `ls.shouldEnforceBlockPipelineCrypto(e.Point.Slot)` here would keep the Mithril exemption in one place, which matters because a future change to the exemption has to find all three.

ledger/eras/cost_models.go
chrisguiney: Resolved at `c8c71eb7`. `requiredCostModel` accepts any present, non-empty model, and `TestRequiredCostModelAcceptsRealAlonzoGenesisModel` decodes the shipped mainnet Alonzo genesis through `AlonzoProtocolParameters.UpdateFromGenesis` and requires its 166-entry PlutusV1 model to be accepted, so the floor cannot come back silently. `DefaultPlutusV2CostModel` at 175 entries is covered by the same bound.

ledger/chainsync.go
chrisguiney: Resolved. `ledger/chainsync.go:3208` and `:4096` both call `IsHeaderVerificationDeferred` now, so the blockfetch path defers on a covered-but-unpublished nonce rather than reporting a crypto failure.

The widened predicate reaches three consumers outside `ledger/`, and all three hold — noted on `ledger/verify_header.go:114` in the current review, including the `leiosMaxDeferredAnnouncements` bound that keeps the new error class from growing the deferred-announcement queue.

ledger/state.go
chrisguiney: Resolved. The note now records that the deleted form was the later 4-way one rather than the original mechanism, and states the operator-visible cost plainly: a `ValidateHistorical=false` node pays Plutus phase-2 evaluation on every deep historical block where it was previously skipped.

ARCHITECTURE.md
chrisguiney: Resolved; the heading reads "Self-Validation" and the paragraph matches the new default.

Separately, `ARCHITECTURE.md:3611` still describes header-crypto verification as skipped under "fast-sync/Mithril-import exemptions", which this change reduces to Mithril only. Detail on `ledger/verify_header.go:194`.

ledger/eras/cost_models.go
chrisguiney: This request is not implemented at `c8c71eb7`, deliberately, and the evidence points the other way.

A length floor against `lang.GetParamNamesForVersion` uses plutigo's protocol-version-11 table (332 for V1/V2, 350 for V3). The cost-model parameter count grew across hard forks, so a shorter list is the complete shape for an earlier protocol version rather than a truncated one. `config/cardano/{mainnet,preprod,preview}/alonzo-genesis.json` each carry a 166-entry PlutusV1 model that `HardForkAlonzo` loads into `CostModels[0]` via `UpdateFromGenesis`, and `ledger/eras/babbage.go`'s `DefaultPlutusV2CostModel` is 175 entries. A floor rejects both and stalls replay at the first Alonzo or Babbage Plutus transaction.

`ledger/eras/cost_models.go:50` now applies presence plus non-empty, the same bound gouroboros' `UtxoValidateCostModelsPresent` uses, and `TestRequiredCostModelAcceptsRealAlonzoGenesisModel` pins the real genesis case through the production loading path. Rejecting a genuinely truncated model needs a protocol-version-to-length table this repository does not have; that is a separate change, not this one.

ledger/eras/cost_models_test.go
chrisguiney: The rejection case this asks for was implemented and then reverted with the check it tested — see the reply on `ledger/eras/cost_models.go`. A three-entry model is short but non-empty, which is now the accepted contract, and the success case a complete model has to satisfy is covered by `syntheticFullCostModel` plus `TestRequiredCostModelAcceptsRealAlonzoGenesisModel` against the real 166-entry mainnet Alonzo model.

ledger/eras/cost_models.go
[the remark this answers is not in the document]
chrisguiney: Rejected, with the same reasoning as the equivalent human request on this file. A length floor against `lang.GetParamNamesForVersion` is plutigo's current protocol-version-11 table; the shipped `alonzo-genesis.json` files carry a 166-entry PlutusV1 model and `DefaultPlutusV2CostModel` is 175, both complete for their era, and both would be rejected. `TestRequiredCostModelAcceptsRealAlonzoGenesisModel` pins that case through `AlonzoProtocolParameters.UpdateFromGenesis`. Validating against the active protocol era would need a version-to-length table that does not exist in this repository.

arepala-uml: Thank you for verifying with the real genesis/babbage-default figures. Confirmed and leaving the presence-only check as-is.

ledger/verify_header.go
arepala-uml: Thank you for tracing through all three external consumers and recordBlockPipelineError -- confirmed, no further change needed there.

ledger/verify_header.go
arepala-uml: Fixed in c30db348: corrected both ouroboros/chainsync.go:1072-1075 and ARCHITECTURE.md's matching paragraph to say verification is skipped only for a Mithril-covered slot, not the old bulk historical/catch-up loading exemption.

ledger/verify_opcert.go
arepala-uml: Thank you for tracing the KES clamp paths and confirming all three shipped genesis files. Fixed the stale maxKESEvolutions doc comment in c30db348: it no longer promises a lighter future-cert-only guard fallback, since verifyOpCertHeaderCrypto now fails closed on a zero maxKesEvolutions instead.

ledger/chainsync.go
arepala-uml: Agreed on the duplication, but a direct call to shouldEnforceBlockPipelineCrypto here isn't a safe drop-in: it also folds in hasCachedEpochNonceForSlot, which would make a not-yet-cached-nonce slot skip this whole block outright -- losing the flush-and-retry step at lines 4077-4085 (flushPendingBlockfetchBlocksDeferred + re-checking FirstVerifiedHeaderMatchesPoint before falling through to verifyBlockHeaderCryptoBeforeApply's own deferral handling). That would be a behavior change, not a pure dedup.

Fixed in c30db348 with a narrower extraction instead: pulled just the Mithril-slot boundary check (the part that's textually identical across all three sites) into a new slotCoveredByMithril helper, and left the nonce-caching/flush/defer control flow here untouched. All three call sites (this one, shouldEnforceBlockPipelineCrypto, and ShouldVerifyChainSelectionHeaderCrypto) now share that one predicate, so a future change to the Mithril exemption only has one place to find -- without the regression the full-function swap would have caused.

ledger/verify_opcert.go
```diff
@@ -281,19 +281,26 @@ func verifyOpCertHeaderCrypto(
...
 			slotsPerKesPeriod,
 			maxKesEvolutions,
-		); err != nil {
-			return fmt.Errorf("opcert KES period invalid: %w", err)
-		}
+		)
+	}
+	if _, err := ledger.ValidateKesPeriod(
+		opCert.KesPeriod,
+		slot,
```
chrisguiney, line 299: Blocking: this fails closed on `maxKesEvolutions == 0`, but nothing rejects that value at startup, so the diagnosis lands on every header instead of once at load.

`ledger/verify_header.go:411` rejects `slotsPerKesPeriod == 0` with an explicit error six lines before it calls this at `:434`. `ls.maxKESEvolutions()` (`ledger/verify_header.go:1905`) returns 0 for a genesis that omits `maxKESEvolutions`, and that 0 reaches here with no equivalent guard.

Because this change also requires header crypto for every non-Mithril slot, the result is a non-deferred error on every post-Byron header, so `ledger/chainsync.go:3231` recycles the peer indefinitely and the node never syncs, with nothing in the log naming the genesis field as the cause.

The five fixtures this PR had to update are the evidence that omitting the key is easy: `ledger/verify_header_test.go:995`, `:1597`, `:1633`, `:3337`, and `ledger/recovery_rewind_window_test.go:53` all previously worked without it. Before this change it degraded to the future-cert guard rather than being fatal.

All six shipped `config/cardano/*/shelley-genesis.json` carry the field (62, devnet 60), so no supported network breaks today.

Suggestion: assert it in `validateGenesisConsistency()` (`config/cardano/genesis_consistency.go`), which already fails closed at load time for exactly this class of problem and whose own comment makes the argument. `TestEmbeddedConfigsPassGenesisConsistency` would then cover every shipped config.

ledger/chainsync.go
```diff
@@ -4055,11 +4066,12 @@ func (ls *LedgerState) handleEventBlockfetchBlockDeferred(
...
-	// Verify block header cryptographic proofs (VRF, KES).
-	// Skip during historical sync (validationEnabled=false) because
-	// historical blocks were already validated by the network.
-	validationEnabled, _ := ls.validationStateSnapshot()
-	if validationEnabled {
+	// Verify block header cryptographic proofs (VRF, KES). Required for
+	// every slot except one a Mithril certificate already covers -- see
+	// chainsyncHeaderCryptoPolicy's doc comment. A coarse
+	// ValidateHistorical=false historical-sync toggle must not disable
+	// this entire group of checks (issue #3528).
```
chrisguiney, line 4084: Blocking: this gate has no test that fails without it.

Reverting just this condition to the previous `validationEnabled` check leaves `go test ./ledger/` fully green. The other two removals are covered - reverting `:3634` alone fails `TestHandleEventChainsyncForkRecordsAdmittedHeaderFrontier`, and reverting `:3670` alone fails `TestDecodeReadChainBatchMirrorsSerialValidationGates` - so this is the one gap.

It is also the widest of the three. It guards `verifyBlockHeaderCryptoBeforeApply`, which runs VRF proof, KES signature, opcert cold-key signature and KES expiry, genesis-overlay classification, registered-pool VRF-key binding, and stake-derived leader eligibility. All of that was previously skipped for an entire `ValidateHistorical=false` run, which makes this the largest behavioural change in the PR.

`ledger/chainsync_recycle_test.go:404` reads like coverage for it, but sets `ls.validationEnabled = true`, so it exercises the deferred-error classification rather than the gate itself.

Suggested assertions: with `validationEnabled=false` on a non-Mithril slot, a block with genuinely bad header crypto returns a non-deferred `block header crypto verification failed`; with `validationEnabled=false` and the slot Mithril-covered, it does not.

ledger/eras/cost_models.go
chrisguiney, line 50: Blocking: eight of the nine call sites this helper replaced have no test that fails without them.

Reverting each converted site individually to a bare map index and running `go test ./ledger/eras/`, only `ledger/eras/conway.go:1249` (PlutusV1) fails a test. These stay green: `alonzo.go:301`, `alonzo.go:458`, `babbage.go:328`, `babbage.go:395`, `babbage.go:565`, `babbage.go:614`, `conway.go:1175`, `conway.go:1211`.

`TestRequiredCostModelFailsClosedOnMissingEntry` tests this function in isolation, so it passes with all nine sites reverted. The wiring is what the issue is about, and it is currently pinned at one site out of nine. Neither `EvaluateTx*` variant is covered at all, and those are the ones carrying the distinct `return 0, lcommon.ExUnits{}, nil, err` shape, so a mistake there would be the easiest to miss.

Separately, non-blocking - the Dijkstra era does not get this guarantee. `ledger/eras/dijkstra.go:286` delegates phase 2 to `gdijkstra.UtxoValidatePlutusScripts`, which indexes `CostModels` directly. A missing key is caught upstream by `UtxoValidateCostModelsPresent`, but that rule tests only `if _, ok := pp.CostModels[version]; !ok`, so a present-but-empty model passes phase 1 and reaches `cek.NewEvalContext`. plutigo's `costModelFromList` starts from the default model and breaks at `i >= len(data)`, so an empty list overrides nothing and evaluation silently runs on built-in defaults - precisely the case the `len(model) == 0` arm catches here.

ledger/state.go
```diff
@@ -4562,49 +4562,25 @@ func (ls *LedgerState) securityParamForCurrentEraSnapshot() int {
...
+// all, and historicalBlockValidationDecision's validationEnabled==false
+// branch already made shouldValidateBlock true for exactly this catch-up
+// case, so it was genuinely reachable in production. A later revision here
+// added a trustedReplay requirement, which -- because
+// historicalBlockValidationDecision forces shouldValidateBlock to false
+// whenever TrustedReplay is true -- made that specific 4-way combination
+// unreachable; that dead 4-way form, not the original mechanism, is what
+// got removed. Phase 2 now always evaluates whenever per-tx validation runs
+// at all, which is the safer contract issue #3528 asks for, but it has a
+// real cost worth naming plainly: an operator running
```
chrisguiney: The cost stated here is wider than what the code does.

`historicalBlockValidationDecision` (`ledger/state.go:480`) only returns `shouldValidate == true` when `blockSlot >= cutoffSlot`, so blocks below the stability-window cutoff never enter per-tx validation and never ran phase 2 to begin with. The genuinely affected set is blocks that are simultaneously at or above the cutoff and more than `k` behind the tip, not every deep historical block.

Worth tightening, since this note is the durable record of why the shortcut was removed. The same claim appears in the PR description.

ledger/state.go
```diff
@@ -6558,15 +6534,13 @@ func (ls *LedgerState) ledgerProcessBlocksFromSource(
...
-								snapshotEra.Id,
-							),
-						)
+						// Process block. Phase 2 (Plutus evaluation) always
+						// runs when per-tx validation runs at all -- see the
+						// issue #3528 removal note above
+						// historicalBlockValidationDecision for why the old
+						// historical-sync/TrustedReplay phase-2 shortcut was
+						// deleted rather than reworked.
+						const skipPhase2Validation = false
```
chrisguiney, line 6765: Non-blocking: flipping this constant to `true` leaves `go test ./ledger/` green, so nothing pins the new invariant at the decision point. The replacement test (`ledger/block_transaction_validity_test.go:315`) passes the literal `false` into `ledgerProcessBlock`, which pins the callee's handling of the flag rather than the caller's choice of value.

Worth deciding explicitly while this is open: with this constant false, the phase-2 skip path is now unreachable in production. `ledger/state.go:7197` is the only writer of `LedgerView.skipPhase2Validation`, so `SkipPhase2Validation()` (`ledger/view.go:89`), the `phase2ValidationSkipper` interface (`ledger/eras/validation.go:38`) and its five era call sites (`alonzo.go:222`, `babbage.go:236`, `conway.go:241` and `:265`, `dijkstra.go:273`) are reachable only from tests. Either remove the plumbing or record here why it is retained - `unused` cannot see this, since every symbol is still referenced.

internal/config/config.go
```diff
@@ -1139,6 +1142,9 @@ var globalConfig = &Config{
 	// Forging defaults
 	ForgeSyncToleranceSlots:     DefaultForgeSyncToleranceSlots,
 	ForgeStaleGapThresholdSlots: DefaultForgeStaleGapThresholdSlots,
+	// Fail closed: self-validate locally-forged blocks before adoption and
+	// diffusion unless an operator explicitly opts out.
+	ValidateForgedBlock: true,
```
chrisguiney: Non-blocking: this default has no regression test, and it is the one on the operator path (`LoadConfig` -> `GetConfig` -> `RegisterFlags`). Deleting this line leaves `go test ./internal/config/ ./cmd/...` green. `config_test.go:191` covers the `NewConfig` literal only.

Since the hazard is two literals that have to stay in sync forever, the repo's existing idiom for this may be worth borrowing: `PeerSharing *bool` (`internal/config/config.go:744`) with `boolPtrFlag`, whose comment notes that the pointer distinguishes "operator did not set this" from "explicitly false". That would let `ApplyDefaults` own the default in one place.

config_test.go
```diff
@@ -180,6 +180,18 @@ func TestNewConfigDoesNotDefaultCustomMempoolConfig(t *testing.T) {
...
+// TestNewConfigDefaultsValidateForgedBlock is a regression test for issue
+// #3528: NewConfig builds its own internalconfig.Config literal rather than
+// starting from internalconfig.DefaultConfig, so the fail-closed
+// ValidateForgedBlock=true default had to be set in both places. A caller
+// using the programmatic/library API (NewConfig) rather than the
+// YAML/env-loaded path must still get self-validation of forged blocks
+// enabled by default.
+func TestNewConfigDefaultsValidateForgedBlock(t *testing.T) {
+	cfg := NewConfig()
+	assert.True(t, cfg.cfg.ValidateForgedBlock)
```
chrisguiney, line 196: Non-blocking: this asserts `cfg.cfg.ValidateForgedBlock`, but the field the forging wiring actually reads is the compat mirror `cfg.validateForgedBlock` (`config.go:257`), populated by `syncCompatFields` at `config.go:867` and consumed at `node_forging.go:365`. Removing `c.validateForgedBlock` from that assignment breaks the wiring while this test stays green. Asserting both would close it.

Minor: the doc comment names `internalconfig.DefaultConfig`, which is not a symbol in the tree - the other default source is `globalConfig` (`internal/config/config.go:1073`). Same reference at `config.go:696`. Since both comments exist to point the next reader at the other default source, a greppable name would help.

node_forging.go
```diff
@@ -358,8 +358,9 @@ func (n *Node) initBlockForger(
 		)
 	}
 
-	// Wire self-validation when the operator opts in. The validator runs
-	// header crypto, body-hash, and per-tx ledger checks before AddBlock.
+	// Wire self-validation unless the operator explicitly opts out. The
+	// validator runs header crypto, body-hash, and per-tx ledger checks
+	// before AddBlock.
```
chrisguiney: Non-blocking, but worth stating on the PR since it changes what every block producer does by default: validation now runs synchronously inside the forging deadline with no budget on it.

`checkAndForgeProduction` (`ledger/forging/forger.go:686`) takes `_ context.Context`, so not even shutdown interrupts it, and `LedgerState.ValidateForgedBlock` re-runs `ValidateTxWithOverlay` on every transaction the builder already validated during selection - roughly a second full per-tx pass, phase-2 Plutus included.

The header-crypto component is small enough not to matter. The per-tx pass is the term that does, and there is no benchmark for it. `dingo_forge_validation_duration_seconds` exists but has no dashboard panel or alert rule, so an overrun would surface first as a missed slot and higher orphan rate rather than as a metric anyone is watching.

A stated latency budget, or a soft deadline that logs and proceeds rather than dropping the block, would make the new default easier to ship. `dingo.yaml.example` also never mentions `validateForgedBlock`, so the opt-out this PR positions as a legitimate operator choice is currently undiscoverable there.

ARCHITECTURE.md
```diff
@@ -4895,9 +4895,9 @@ Steps 3, 6, and 7 each call into a pluggable interface (`LeaderChecker`, `BlockV
...
```
chrisguiney: This paragraph and the heading above it are correct now.

Three other passages still describe contracts this change removes, and the docs commit did not reach them:

- `ARCHITECTURE.md:2549-2551` - "Historical blocks already at least `k` deep retain the explicit replay-only phase-2 skip; their declared outcomes are trusted only inside that immutable replay window."
- `ARCHITECTURE.md:2766` - "Plutus Phase 2 execution is skipped only for blocks at or before the immutable tip (`tipBlockNo - securityParam`), where the block producer's `isValid` flag is treated as authoritative."
- `ARCHITECTURE.md:2990` - "These checks share the existing skip-during-historical-sync gating", describing the header-time opcert cold-key and KES-expiry checks.

All three describe mechanisms this PR deletes. `internal/docsparity` and `internal/architecture` both pass, so nothing flags them automatically - same shape as the `:3611` paragraph corrected in this round.

chrisguiney (changes requested):
Changes requested. The fail-closed direction reads correct throughout, and the parts I could exercise locally hold up: the `requiredCostModel` conversion is complete across `ledger/eras/` (no remaining value-read index feeding evaluation), a configuration error can never be reclassified as a phase-2 script failure (`validatePlutusOutcome` fails closed in both `IsValid` branches), the exact-vs-approximate body-size claim is true, and `-race` is clean on the affected packages.

Three blockers, all inline:

1. `ledger/verify_opcert.go` - `maxKesEvolutions == 0` is now fatal on every header with no startup check, so a genesis that omits the field starts cleanly and then syncs nowhere.
2. `ledger/chainsync.go` - the blockfetch header-crypto gate has no test that fails without it. The other two gate removals do.
3. `ledger/eras/cost_models.go` - eight of the nine converted call sites have no test that fails without them.

Non-blocking items are also inline: three stale `ARCHITECTURE.md` passages, an overstated cost note, two untested configuration defaults, and the forge-latency question that enabling self-validation by default raises.

On the earlier cost-model length-floor thread: the current presence-plus-non-empty check is the right bound, for the reasons already established in that thread. That one is still open on @wolf31o2's review rather than mine.

Bot coverage: there is no review record at `79fcbc8`. CodeRabbit reports auto incremental reviews are disabled for this repository, and the `cubic` check reports it could not start the incremental review, so neither has seen this head.

ledger/verify_opcert.go
arepala-uml: Fixed in 234278d3: added validateMaxKESEvolutionsPresent to config/cardano/genesis_consistency.go's validateGenesisConsistency, asserting the Shelley genesis carries a positive maxKESEvolutions at load time. Also fixed the RegisterFlags/DefaultConfig default test gap this same investigation surfaced (see the reply on internal/config/config.go:1147).

TestValidateGenesisConsistencyRejectsMissingMaxKESEvolutions covers zero, negative, and positive values, plus confirms the check is a no-op with no Shelley genesis. All six shipped genesis files already carry the field, so TestEmbeddedConfigsPassGenesisConsistency continues to pass unchanged.

ledger/chainsync.go
arepala-uml: Fixed in 234278d3: added TestBlockfetchHeaderVerificationRunsRegardlessOfValidationEnabled (validationEnabled=false, non-Mithril slot, unverifiable header crypto -> non-deferred "block header crypto verification failed" error) and TestBlockfetchHeaderVerificationSkippedForMithrilCoveredSlot (same setup but Mithril-covered slot -> no error). Confirmed the first one fails if this gate is reverted to the old validationEnabled check, restoring the exact gap you found.

ledger/eras/cost_models.go
arepala-uml: Fixed in 234278d3: added ledger/eras/required_cost_model_wiring_test.go, covering the 8 remaining call sites -- ValidateTxAlonzo, EvaluateTxAlonzo, ValidateTxBabbage's V1 and V2 branches, EvaluateTxBabbage's V1 and V2 branches, and evaluateConwayPlutusScript's V2 and V3 branches -- with the same missing-cost-model contract TestConwayPlutusBudgetComparisonIncludesFinalSlippageBatch already pinned for Conway V1. Reverted each site individually to a bare map index and confirmed all 8 new subtests fail, same method you used to find the gap.

Left the Dijkstra non-blocking point as-is for now (a present-but-empty PlutusV3 model reaching UtxoValidatePlutusScripts): agreed it's a real gap, but closing it needs a protocol-version-to-shape table this repo doesn't have, which is a bigger, separate change from wiring the existing requiredCostModel helper into more call sites.

ledger/state.go
arepala-uml: Fixed in 234278d3: rewrote the comment to name the actual affected set -- blocks simultaneously at or above historicalBlockValidationDecision's near-tip stability-window cutoff (by slot) and more than the security parameter behind the tip (deepHistoricalBlock's own dimension) -- rather than "every deep historical block."

ledger/state.go
arepala-uml: Chose to document rather than remove, since this looks like deliberate scaffolding rather than accidental dead code: LedgerView.skipPhase2Validation, SkipPhase2Validation(), the phase2ValidationSkipper interface, and all 5 era call sites already have dedicated tests (TestLedgerViewSkipPhase2Validation plus per-era skip tests in plutus_validity_outcome_test.go, validation_test.go, byron_test.go's mock). Removing it would mean deleting that test coverage too, which felt like too large a unilateral call for a non-blocking finding without knowing the original intent.

Investigated why it's dead: TrustedReplay currently bypasses per-tx validation entirely via historicalBlockValidationDecision's coarser shouldValidateBlock=false gate, not through this narrower phase-2-only mechanism -- so this plumbing reads as pre-built for a future caller that wants phase-1 UTXO checks to keep running while still trusting the producer's isValid flag for phase 2, which hasn't been wired up yet. Documented that reasoning on both LedgerView.skipPhase2Validation's doc comment and the const skipPhase2Validation = false call site in 234278d3, so the next person has the same context instead of just finding a dead const.

internal/config/config.go
arepala-uml: Fixed in 234278d3, and this surfaced a real drift the new test caught: TestValidateForgedBlockDefaultsToTrue pins newDefaultConfig().ValidateForgedBlock directly, plus LoadConfig+ApplyDefaults and the registered --validate-forged-block flag default, on the real operator path. It initially failed against resetGlobalConfig()'s hand-maintained test literal (config_test.go), which had already drifted from the real default and was missing ValidateForgedBlock -- exactly the two-literals-to-keep-in-sync hazard you flagged, just one layer further out (a test double, not the RegisterFlags default itself).

Rather than add the boolPtrFlag/PeerSharing pattern (ValidateForgedBlock's default doesn't need to distinguish "unset" from "explicitly false" the way PeerSharing does -- there's no per-node-role conditional default here), I extracted the internal/config default literal into newDefaultConfig() so there's exactly one place these defaults are written, and pointed resetGlobalConfig's doc comment at it.

config_test.go
arepala-uml: Fixed in 234278d3: TestNewConfigDefaultsValidateForgedBlock now also asserts cfg.validateForgedBlock (the compat mirror node_forging.go actually reads), proven to catch a break in syncCompatFields' assignment. Also fixed the doc comment referencing the nonexistent internalconfig.DefaultConfig symbol, on both this test and the matching comment in config.go's NewConfig -- both now point at internal/config's real package-level default (globalConfig, built by its newDefaultConfig, see the reply on internal/config/config.go:1147).

node_forging.go
arepala-uml: Documented rather than fully solved in 234278d3: added a validateForgedBlock entry to dingo.yaml.example (previously entirely undiscoverable there) with a note on the synchronous-validation latency cost and how to opt out. Agreed a real latency budget/soft-deadline mechanism and a dashboard panel for dingo_forge_validation_duration_seconds are worth doing, but that's a separate, larger change from what this PR set out to do -- happy to file it as a follow-up issue if useful.

ARCHITECTURE.md
arepala-uml: Fixed in 234278d3: corrected all three -- the k-deep replay-only phase-2 skip paragraph (~2549), the immutable-tip Phase 2 skip paragraph (~2766), and the opcert header-check historical-sync gating bullet (~2993) -- to state that Phase 2 and the header-time checks now run unconditionally (except the single Mithril-covered-slot exemption), matching the corrected :3611 paragraph and the code.

wolf31o2 (dismissed):

chrisguiney (approved):

wolf31o2 (approved):

Issue #2268, opened 2026-05-13
connmanager: enforce bounds on NtC listener connections


wolf31o2:
## Summary

Node-to-client listener connections still bypass inbound connection limits.

## Audit Details

Finding ID: M3, MEDIUM network security finding.

Affected file: `connmanager/listener.go`.

The listener explicitly bypasses the inbound slot budget and per-IP rate limiting for `UseNtC` connections because they are expected to be local wallets/tools rather than network peers. That assumption is reasonable for UNIX sockets, but TCP NtC listeners can still be reachable by local users or compromised processes.

Impact: unbounded NtC connections can exhaust goroutines, file descriptors, or memory without consuming the normal inbound peer budget.

## What Needs to Happen

1. Add a separate NtC connection budget that does not compete with N2N peers.
2. Apply per-source limits to TCP NtC listeners where a remote address exists.
3. Preserve UNIX socket behavior if per-IP accounting does not apply.
4. Add metrics/logging for rejected NtC connections.
5. Add tests for limit enforcement and listener shutdown.

## Acceptance Criteria

- TCP NtC listeners have bounded connection admission.
- Tests prove N2C limits do not break N2N peer budgeting.

PR #4103, opened 2026-09-07
fix(connmanager): release NtC admission on cleanup

wolf31o2:
Release NtC admission during removal and replacement cleanup paths, with once-only cleanup to prevent double release and stale watcher retention. Adds focused race-enabled cleanup and collision coverage.

---
## Summary by cubic
Fixes NtC admission so slots are released exactly once on connection removal, replacement, and rejection, preventing capacity leaks and stale watcher retention. NtC connections previously bypassed admission limits; they now have dedicated total and per-IP limits with a rejection metric.

**New Features**
- Adds `MaxNtCConns` and `MaxNtCConnectionsPerIP` config options with defaults, exposed via YAML, CLI flags, and `DINGO_*` environment variables, applied at startup and network reinitialization.
- Records rejections in `cardano_node_metrics_connectionManager_ntcRejectedConns_total`.

**Bug Fixes**
- Releases admission outside the connection-map lock and before close callbacks run; the connection's close watcher remains the backstop if an evicted entry is no longer present in the map.
- Aligns the Leios deferred-header admission test with the current crypto policy, which no longer treats "validation not yet enabled" as an unverified path.

<sup>Written for commit f80424fcd4410a6f1bba9fd8ac30195695827606. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added configurable limits for total node-to-client connections and connections per IP.
  - Limits can be set through configuration files, environment variables, command-line flags, or programmatic options.
  - Connection limits are now applied during node startup and network reinitialization.

- **Bug Fixes**
  - Improved connection cleanup so capacity is released promptly when connections close or are replaced, preventing duplicate cleanup.


connmanager/listener.go
```diff
@@ -469,9 +485,17 @@ func (c *ConnectionManager) setupAcceptedConnection(
 			"remote_addr",
 			peerAddr,
 		)
-		if !c.addNtCConnectionWithIPKey(oConn, true, peerAddr, "") {
+		if !c.addConnectionImpl(
```
chrisguiney, line 488: This is the only production caller `addNtCConnectionWithIPKey` had (`connmanager/listener.go:472` on `main`), so calling `addConnectionImpl` directly here leaves the helper with no non-test references.

`.golangci.yml` sets `run.tests: false`, so the four remaining call sites in `_test.go` do not count, and the `lint` job fails on this head with:

```
connmanager/connection_manager.go:699:29: func (*ConnectionManager).addNtCConnectionWithIPKey is unused (unused)
```

`golangci-lint run ./connmanager/...` reports that 1 issue at this head and 0 issues at the merge base, so it belongs to this change and not to `main`.

Two ways out: drop `addNtCConnectionWithIPKey` and have the four tests call `addConnectionImpl(conn, isInbound, true, peerAddr, ipKey, nil)`, or give the helper an `onClose func()` parameter and keep calling it from here.

connmanager/connection_manager.go
```diff
@@ -118,7 +124,9 @@ type ConnectionManagerConfig struct {
 	// MaxConnectionsPerIP limits the number of concurrent inbound
 	// connections from the same IP address. IPv6 addresses are grouped
 	// by /64 prefix. A value of 0 means use DefaultMaxConnectionsPerIP.
-	MaxConnectionsPerIP int
+	MaxConnectionsPerIP    int
+	MaxNtCConns            int
```
chrisguiney, line 128: Confirming cubic's finding on these two fields, with the comparison to their N2N counterparts.

Nothing outside `_test.go` assigns either one. Both production construction sites, `node.go:1207` and `node_lifecycle.go:937`, set only `MaxConnectionsPerIP` and `MaxInboundConns`. There is no `dingo.WithMaxNtCConns` option, no `internal/config.Config` field, no flag in `internal/config/flags.go`, and no `DINGO_*` variable. `MaxInboundConns` has all four: `config.go:1364`, `internal/config/config.go:631`, `internal/config/flags.go:527`, `internal/node/node.go:646`.

The consequence is that every node runs the hardcoded 100 total and 5 per source IP, and that is a behavior change from unbounded rather than a tunable default. Five concurrent TCP NtC sessions from one source is reachable in ordinary deployments, such as several containers sharing a bridge IP or a client that pools connections. At that point the operator gets a counter increment and no knob to raise the limit.

Either resolution works: wire both fields through the same four places `MaxInboundConns` uses, or unexport them and keep the limits internal. The first seems the better fit, since `ARCHITECTURE.md` documents both by name, which reads as an operator-facing setting.

ARCHITECTURE.md
```diff
@@ -610,6 +610,23 @@ peer credentials) unwrap through the wrapper, so wrapping an accepted connection
 never silently disables them. Cancellation closes an in-flight bearer, and failed
 setup releases its reserved inbound and per-IP slots.
 
+Node-to-client listeners share a separate admission budget across all their
+transports: `ConnectionManagerConfig.MaxNtCConns` defaults to 100 pending or
+established sessions. TCP clients also share a separate per-source budget,
+`MaxNtCConnectionsPerIP`, defaulting to five sessions per IPv4 address or IPv6
+/64. Nonpositive settings use these defaults. Unix sockets and named pipes
+consume the total NtC budget without per-IP accounting. NtC admission never
+consumes N2N slots or N2N per-IP capacity. Admission reserves both budgets
```
chrisguiney, line 619: This sentence is accurate for this branch, and it is exactly where this change and #4098 disagree.

#4098 changes `tryReserveInboundSlot` to count inbound NtC connections against `MaxInboundConns` through a new `inboundAdmissionCountLocked`, which is the opposite contract: there an NtC connection does consume an N2N slot. Applying only that change on top of this head makes this branch's own `connmanager/ntc_admission_test.go:143` fail, because `require.True(t, manager.tryReserveInboundSlot())` returns false once two NtC sessions are live under `MaxInboundConns: 1`. If both landed, every TCP NtC client would be charged against two budgets and this paragraph would no longer describe the code.

The first acceptance criterion on #2268 asks for "a separate NtC connection budget that does not compete with N2N peers", which is the design in this PR rather than the one in #4098. That choice is worth making before either merges, with the other closed.

Related, on the same issue: this branch's first commit `ee9584e` is identical to #4077's head across `connmanager/`, so this PR already contains #4077 in full. #4077 carries the `Fixes #2268` link; this PR links no issue, and its description covers only the release half of a diff that also introduces the whole admission budget, two config fields and a new metric.

chrisguiney (changes requested):
The admission-release logic itself holds up. Every path that finishes an NtC connection releases its reservation exactly once: the accept-loop rejection takes no reservation, every pre-registration exit in `setupAcceptedConnection` goes through the deferred release, `RemoveConnection` and both collision-replacement branches release at each of the three `delete(c.connections, connId)` sites, and the error-watcher's `defer onClose()` backstops an evicted connection whose `RemoveConnection` returns false. Double release is prevented twice over, since `reserveNtCSlot` returns a `sync.OnceFunc` and `addConnectionImpl` wraps it in another. That class looks complete.

Reverting the four release call sites in place makes both new cleanup tests fail, so they are load-bearing rather than guards.

Requesting changes on two points: the `lint` job is red on this head and this change causes it, and the two new configuration fields cannot be set by a node.

Also worth settling before merge: three open pull requests now implement #2268, and two of the designs are mutually exclusive. Details inline.

connmanager/connection_manager.go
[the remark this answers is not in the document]
chrisguiney: Confirmed, and still open at this head.

Nothing outside `_test.go` assigns `MaxNtCConns` or `MaxNtCConnectionsPerIP`. `node.go:1207` and `node_lifecycle.go:937` set only `MaxConnectionsPerIP` and `MaxInboundConns`, and there is no `dingo.WithMaxNtCConns` option, no `internal/config.Config` field, no flag, and no environment variable, where `MaxInboundConns` has all four. Every node therefore runs the hardcoded 100 total and 5 per source IP.

Both suggested remedies are reasonable. Wiring the fields through is the better fit here, because `ARCHITECTURE.md` documents both by name and 5 TCP sessions per source IP is reachable in normal deployments with no way to raise it.

connmanager/ntc_cleanup_test.go
```diff
@@ -0,0 +1,162 @@
...
+	releaseCallback()
+	waitForConnectionManagerWatchers(t, manager)
+	require.Equal(t, int32(1), releases.Load())
+	require.Equal(t, int32(1), calls.Load())
+	requireNtCCleanupCounts(t, manager, 1)
+	probeRelease()
+	requireNtCCleanupCounts(t, manager, 0)
+}
+
+func TestNtCCleanupCollisionReleasesBeforeBlockedCallback(t *testing.T) {
```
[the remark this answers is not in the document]
chrisguiney: Not reproducible at this head; the premise does not hold.

`ouroboros.Connection.ErrorChan()` returns a channel created as `make(chan error, 10)` (gouroboros v0.204.0, `connection.go:139`). The single `second.ErrorChan() <- nil` in the cleanup sends into free buffer space, so it cannot block whether or not a reader exists. `waitForConnectionManagerWatchers` is separately bounded by a one-second `t.Fatal` timeout, so it cannot stall teardown either.

Forcing the `requireNtCCleanupCounts` assertion at line 132 to fail confirms it: both subtests fail and the package returns immediately rather than hanging.

Rejecting as a false positive.

connmanager/connection_manager.go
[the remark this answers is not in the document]
chrisguiney: Resolved at 9252a7dc.

Both fields are now settable by a node. `WithMaxNtCConns` and `WithMaxNtCConnectionsPerIP` (`config.go:1375`, `config.go:1385`) write `cfg.MaxNtCConns` / `cfg.MaxNtCConnectionsPerIP`; `syncCompatFields` (`config.go:735`) mirrors them onto `Config.maxNtCConns` / `Config.maxNtCConnectionsPerIP`, and it runs after option application (`config.go:731`) and after the YAML/env config is built (`config.go:1002`). Both construction sites, `node.go:1207` and `node_lifecycle.go:937`, now pass them into `ConnectionManagerConfig`.

The declarative surface is wired too: `maxNtCConns` / `maxNtCConnectionsPerIP` YAML keys and `DINGO_MAX_NTC_CONNS` / `DINGO_MAX_NTC_CONNECTIONS_PER_IP` (`internal/config/config.go:632-633`), `--max-ntc-conns` / `--max-ntc-connections-per-ip` (`internal/config/flags.go:540-541`), and both names added to `logPlainConfigFields` (`internal/config/redact.go:182-183`).

connmanager/ntc_cleanup_test.go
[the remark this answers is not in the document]
chrisguiney: Disposition unchanged at 9252a7dc: not reproducible.

`ouroboros.Connection.ErrorChan()` returns a channel created as `make(chan error, 10)` (gouroboros v0.204.0, `connection.go:139`, still the pinned version in `go.mod`). The single `second.ErrorChan() <- nil` in the cleanup sends into free buffer space, so it cannot block on an early `t.Fatal` path.

connmanager/listener.go
chrisguiney: Resolved at 9252a7dc: `addNtCConnectionWithIPKey` was deleted rather than left with only `_test.go` callers, so `unused` has nothing to report. No references to the helper remain anywhere in the tree, and the `lint` check is green on this head.

`main` reached the same end state independently through #4077, so the merge does not reintroduce it.

connmanager/connection_manager.go
chrisguiney: Resolved at 9252a7dc. Details in the reply on cubic's thread for the same line: the option, YAML, environment and flag paths all reach `ConnectionManagerConfig` through `syncCompatFields` and both `NewConnectionManager` call sites.

ARCHITECTURE.md
chrisguiney: Settled, and in this branch's favour. #4077 merged to `main` as be8ffb6f carrying the separate-NtC-budget design, so `ConnectionManagerConfig.MaxNtCConns` and `MaxNtCConnectionsPerIP` as an independent budget is now `main`'s contract and this paragraph describes it accurately.

#4098 still implements the opposite contract — inbound NtC counted against `MaxInboundConns` via `inboundAdmissionCountLocked` — and now disagrees with `main` rather than with this branch. That is #4098's question to resolve, not this one's.

`main` has its own shorter version of this paragraph, so this file is one of the three merge conflicts; this branch's wording is the accurate one for the merged behaviour.

chrisguiney (commented):
Both blockers from the previous review are resolved at 9252a7dc.

- The `unused` lint failure is gone: `addNtCConnectionWithIPKey` was deleted rather than left with only `_test.go` callers.
- `MaxNtCConns` and `MaxNtCConnectionsPerIP` are reachable from a node. The programmatic options, the `maxNtCConns` / `maxNtCConnectionsPerIP` YAML keys, `DINGO_MAX_NTC_CONNS` / `DINGO_MAX_NTC_CONNECTIONS_PER_IP`, and `--max-ntc-conns` / `--max-ntc-connections-per-ip` all reach `ConnectionManagerConfig` through `syncCompatFields` and both construction sites in `node.go` and `node_lifecycle.go`.

The design question raised last round is settled: #4077 merged the separate-NtC-budget design to `main` as be8ffb6f, so the contract `ARCHITECTURE.md` describes here is now `main`'s contract. #4098 carries the opposite design and now disagrees with `main` rather than with this branch.

What this branch still adds over `main` is the part worth keeping: `connectionInfo.onClose`, so `RemoveConnection` and both collision-replacement branches release the reservation promptly instead of waiting for the evicted connection's watcher to observe `ErrorChan`. Removing those three call sites fails both new `ntc_cleanup_test.go` tests, so they are load-bearing rather than guards. Release happens after `connectionsMutex` is unlocked and before `ConnClosedFunc` on all three paths, matching the `ARCHITECTURE.md` wording.

Approval is withheld only because the branch cannot merge. `main` is 63 commits ahead and already carries most of this diff through #4077, so the green checks on this head ran against a tree that never contained it. Three files conflict, and all three resolutions look mechanical:

- `connmanager/ntc_admission_test.go` — `main`'s side is a strict superset, adding `TestNtCAdmissionSlotReleasedBeforeCloseCallback` and the `errors` import. Taking `main`'s side loses nothing.
- `connmanager/connection_manager.go` — the two sides are the same deferred release written differently. `main` also keeps an explicit release after `RemoveConnection` returns; because the callback is `sync.OnceFunc`-wrapped and `RemoveConnection` now releases through `connectionInfo.onClose`, keeping both is harmless.
- `ARCHITECTURE.md` — this branch's paragraph is the accurate one for the merged behaviour.

One non-blocking gap and one nit inline.

config.go
```diff
@@ -1369,6 +1370,26 @@ func WithMaxInboundConns(n int) ConfigOptionFunc {
 	}
 }
 
+// WithMaxNtCConns specifies the maximum number of node-to-client connections.
+// Non-positive values are ignored. Default: 100.
+func WithMaxNtCConns(n int) ConfigOptionFunc {
```
chrisguiney, line 1375: Non-blocking: these two options have no test, where their siblings do.

`TestPeerGovernorOptionsIgnoreNonPositiveValues` and `TestPeerGovernorOptionsApplyPositiveValues` in `config_test.go` cover `WithMaxConnectionsPerIP` and `WithMaxInboundConns` for both the ignored non-positive case and the applied case. `WithMaxNtCConns` and `WithMaxNtCConnectionsPerIP` document the same "Non-positive values are ignored" contract but are exercised nowhere outside `connmanager`, which sets `ConnectionManagerConfig` directly and never goes through these options.

Since this wiring is what makes the two limits reachable at all, it is worth a line in each of those two existing tests, plus an assertion that `syncCompatFields` mirrors them onto `Config.maxNtCConns` / `Config.maxNtCConnectionsPerIP` — that mirror is what `node.go` and `node_lifecycle.go` actually read, and nothing currently fails if it is dropped.

internal/config/config.go
```diff
@@ -629,6 +629,8 @@ type Config struct {
 	InboundCooldown          time.Duration `yaml:"inboundCooldown"          envconfig:"DINGO_INBOUND_COOLDOWN"`
 	MaxConnectionsPerIP      int           `yaml:"maxConnectionsPerIP"      envconfig:"DINGO_MAX_CONNECTIONS_PER_IP"`
 	MaxInboundConns          int           `yaml:"maxInboundConns"          envconfig:"DINGO_MAX_INBOUND_CONNS"`
+	MaxNtCConns              int           `yaml:"maxNtCConns"               envconfig:"DINGO_MAX_NTC_CONNS"`
```
chrisguiney, line 632: Nit: the `envconfig` column is one space off the rest of the block on both new lines. `gofmt` cannot fix it because the padding is inside the tag string.

ARCHITECTURE.md
```diff
@@ -7,10 +7,6 @@ domain packages do not self-register during package initialization. The host
 erases types only at its registry boundary while typed factories retain their
 provider configuration, service, and dependency bundles.
 
-`Node.New` validates plugin selections and node configuration before registering
-metrics or starting EventBus workers. Configuration failure leaves the caller
-metrics registry unchanged, so corrected construction can reuse it.
-
 Startup resolves storage, constructs database and ledger, resolves mempool,
```
chrisguiney, line 14: Blocking: both base merges on this branch resolved `ARCHITECTURE.md` by taking the branch's copy and discarding `main`'s side. The file is byte-identical between 9252a7dc and the merge 9952b25c, and again between 9952b25c and 8a323cc6, while differing from each merge's `main` parent (542a1886 and 81e7e82c respectively).

Merging as-is therefore reverts documentation that is already on `main`:

- 178 backticked symbols documented on `main` are absent from the merge result, contributed by roughly 48 commits that touched this file since this branch's original merge base (4e7843ef). Examples: `AllowInsecureHTTP` (#3838), `chain.ErrRollbackExceedsSecurityParam` (#3938), `BootstrapConfig.PinnedDigest` (#3948), `BlockPipeline.Errors()` (#3949), and `compareCostModels` (#4190 — which is this branch's own current merge base, so the loss includes content the branch has already merged in).
- 47 symbols that `main` removed come back, including the `apiauth.Interceptor` / `AuthPolicy` prose that #4181 removed. That re-publishes a description of an API auth surface `main` no longer has, which is worse than a stale omission.

The merge result is 107,320 words against `main`'s 113,329.

Nothing else in the tree is affected — every other file in the merge result matches `main` or carries only this PR's intended change.

Resolving `ARCHITECTURE.md` in favour of `main` and re-applying only this PR's NtC admission paragraph (around line 618 here) clears this.

chrisguiney (changes requested):
The NtC cleanup itself holds up at this head. `connectionInfo.onClose` survived the base merges, and all three release sites are individually load-bearing: removing the call in `RemoveConnection` alone fails `TestNtCCleanupNormalCloseBeforeBlockedCallback` and `TestNtCAdmissionSlotReleasedBeforeCloseCallback`, and removing either collision-replacement call alone fails the matching `TestNtCCleanupCollisionReleasesBeforeBlockedCallback` subtest. Double release is guarded twice, by `sync.OnceFunc` in `addConnectionImpl` and again in `reserveNtCSlot`. Release happens after `connectionsMutex` is unlocked and before `ConnClosedFunc` on all three paths, matching the documented contract.

Over current `main` this branch still adds the node-level wiring for `MaxNtCConns` and `MaxNtCConnectionsPerIP` — options, getters, YAML and environment keys, CLI flags, redaction entries, and both `NewConnectionManager` call sites — which #4077 did not include, so on `main` today those limits cannot be set and every node runs the defaults. It also adds the prompt release and its tests.

One blocker remains, in `ARCHITECTURE.md`, inline.

connmanager/connection_manager.go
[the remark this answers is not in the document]
chrisguiney: Still resolved at 8a323cc6. Both limits reach `ConnectionManagerConfig` from node configuration through `syncCompatFields` and both `NewConnectionManager` call sites in `node.go` and `node_lifecycle.go`. `main` does not carry that wiring, so it is part of what this branch still adds over `main`.

config.go
chrisguiney: Still open at 8a323cc6. Neither option has a test. `TestPeerGovernorOptionsIgnoreNonPositiveValues` and `TestPeerGovernorOptionsApplyPositiveValues` cover `WithMaxConnectionsPerIP` and `WithMaxInboundConns` for both the ignored non-positive case and the applied case; `WithMaxNtCConns` and `WithMaxNtCConnectionsPerIP` document the same non-positive rule and have no test reference anywhere in the tree. Non-blocking.

internal/config/config.go
chrisguiney: Still open at 8a323cc6. `envconfig:` begins one column later on both new lines than on every other field in the block. Cosmetic, and `gofmt` cannot correct it because the padding sits inside the tag string.

ARCHITECTURE.md
chrisguiney: Settled, and no longer contested. #4077 merged the separate-NtC-budget design to `main`, closing #2268 as completed, so the contract this paragraph describes is `main`'s contract. #4098, which carried the opposite design of counting inbound NtC against `MaxInboundConns`, was closed without merging. This paragraph is accurate against `main` and against this branch.

connmanager/listener.go
chrisguiney: Resolved. `addNtCConnectionWithIPKey` is deleted rather than left with only `_test.go` callers, so `unused` has nothing to report, and `main` reached the same end state independently through #4077. No reference to the helper remains in the tree.

connmanager/connection_manager.go
chrisguiney: Resolved. Both fields are settable by a node through the option, YAML, environment and flag paths, and both `NewConnectionManager` call sites pass them. See the reply on cubic's thread for the same line.

ARCHITECTURE.md
arepala-uml, line 14: Still open, despite `592d6aef` ("chore: preserve current architecture documentation") landing after chrisguiney's review specifically to fix this.

Comparing this file at the current head against its own literal merge parent (`dfb15022`, main) rather than against the file as it stood at any earlier review point: 958 lines are missing that main has and this branch does not, none of them related to this PR's own NtC-admission change. Examples: the migration registry list ("migrations 1 through 13" collapses to "1 through 10", dropping `committee-credential-tags` through `pointer-address-stake`), the decode-cache hash-initialization paragraph, the atomic ChainSync-rollback bookkeeping paragraph, and the `Node.New` validation-ordering paragraph removed at the very top of this diff.

`make docs-parity` doesn't catch this class of regression — it verifies documented symbols still exist in code, not that accurate prose survived a merge.

This PR's own addition (the `MaxNtCConns`/per-IP admission paragraph) is intact; only main's unrelated content is being dropped. Needs a proper three-way reconciliation against current main, not another merge that keeps taking this branch's side wherever the two touched the same paragraph.

connmanager/connection_manager.go
```diff
@@ -718,11 +719,17 @@ func (c *ConnectionManager) addConnectionImpl(
 				"peer_addr", peerAddr,
 			)
 		}
+		if onClose != nil {
```
arepala-uml, line 722: This `onClose()` call (and its twin at line 762, both added in `372151bb`) is not pinned by any test. Reverting both call sites and running the full `connmanager` suite (`go test ./connmanager/... -run TestNtC`) leaves every test green — nothing currently exercises a rejected connection releasing its admission reservation.

Every other release site in this PR was held to exactly this bar (chrisguiney: reverting each of the other four individually breaks a specific test). Worth a `TestNtCAdmissionReleasedOnRejection`-shaped test covering the shutdown-rejection path here and the collision-rejection path at line 762, so this is load-bearing rather than a guard that happens to agree with the rest of the code.

arepala-uml (changes requested):
The admission-release logic itself is solid (every teardown path releases exactly once, double-release guarded twice by sync.OnceFunc, both new config fields fully wired). One blocker: the ARCHITECTURE.md merge regression chrisguiney already flagged is still present despite the follow-up merge that claimed to fix it. One new gap: an added release site has no test pinning it.

ARCHITECTURE.md
```diff
@@ -31,12 +27,22 @@ observed it, so the normal shutdown triggered by that cancellation cannot
...
-construction. TLS remains optional. See External Interfaces for precedence
-and shared bind settings.
+in-process consumer of their concrete server values. Each API provider's
+TLS/authentication policy goes through a merged-config handoff, not a
+pre-resolved one: node composition merges the shared `api.tls`/`api.auth`
+default into that provider's own `config.tls`/`config.auth` fields, field by
+field, before the provider ever decodes its config (see "API security" under
+External Interfaces) — so from a provider's point of view an inherited
+field and an inline one are indistinguishable — but the merged result is
+still the raw, tri-state `TLSPolicy`/`AuthPolicy` shape. Each provider's own
```
chrisguiney: Blocking, still present: against its merge base with `main` (2f465dba) this file differs in 63 hunks, and only lines 643-648 belong to this change. 204 backticked symbols documented there are missing (e.g. `ErrChainTruncatedLedgerRollbackFailed`, `BlockPipeline.Errors()`, `BootstrapConfig.PinnedDigest`; the migration list stops at 10 of 13), and 68 removed ones return, including this `api.auth` / `AuthPolicy` / `internal/apiauth` prose that #4239 removed. No `internal/apiauth` package or `AuthPolicy` type exists.

ARCHITECTURE.md
```diff
@@ -692,7 +640,12 @@ established sessions. TCP clients also share a separate per-source budget,
 consume the total NtC budget without per-IP accounting. NtC admission never
 consumes N2N slots or N2N per-IP capacity. Admission reserves both budgets
 before launching a handshake worker; failed setup releases the reservation,
-and successful setup transfers its release to the connection's close watcher.
+and successful setup transfers its release to the registered connection and
```
chrisguiney: Blocking: `git merge origin/main` conflicts in this file only, on three `main` additions this branch lacks: OpCert counter validation (#4078), LeiosFetch admission bounds (#4282), TxSubmission size discrepancy (#4194). Keeping this side drops all three; Go files merge cleanly. Every merge since 592d6aef kept the earlier discard and `make docs-parity` passes, so neither catches it. Fix both: take this file from current `origin/main` and re-apply only lines 643-648.

connmanager/connection_manager.go
```diff
@@ -718,11 +719,17 @@ func (c *ConnectionManager) addConnectionImpl(
...
+			onClose()
```
chrisguiney, line 723: Non-blocking: this call and the one at line 762 duplicate the caller's release. `setupAcceptedConnection` defers `releaseNtCSlot()` (`listener.go:375`) and clears it only after `addConnectionImpl` returns true (`listener.go:498`), so each false return releases twice, safe only through `reserveNtCSlot`'s `sync.OnceFunc`. That is why removing both calls leaves `go test -race ./connmanager/` green, and why the CodeRabbit finding on 592d6aef did not hold.

connmanager/connection_manager.go
```diff
@@ -751,6 +758,9 @@ func (c *ConnectionManager) addConnectionImpl(
 			)
 			c.releaseIPSlot(ipKey)
 			c.goroutineWg.Done()
+			if onClose != nil {
+				onClose()
```
chrisguiney, line 762: Non-blocking: pick one owner for a rejected registration. Either drop this call and the one at line 723 and keep the caller's defer, or have `addConnectionImpl` own `onClose` on every return, clear `releaseNtCSlot` before calling it, and pin both rejections (shutdown, and collision with an outbound entry) with a direct `addConnectionImpl` test using a counting callback.

chrisguiney (changes requested):
Admission release holds: the `RemoveConnection` and both collision-replacement releases are each load-bearing, and `main` still lacks this branch's `MaxNtCConns` / `MaxNtCConnectionsPerIP` node wiring. Blockers: the `ARCHITECTURE.md` merge regression remains and has grown, and the branch conflicts with `main`, so GitHub ran no test or lint workflows on this head. Non-blocking: the 372151bb rejection releases are redundant; the earlier option-test and tag-alignment points remain open.

ARCHITECTURE.md
```diff
@@ -69,6 +69,13 @@ repeatable-read snapshots. All three return `*sqlstore.Store`; metadata
 business behavior is implemented once in `sqlstore` and dialect translation is
 limited to SQL mechanics.
 
+Metadata indexing treats raw CBOR as the lossless storage and API JSON as an
+optional representation. A label whose map keys collide after JSON
+stringification keeps its label and CBOR row but has no JSON representation;
+readers handle that condition per record rather than aborting indexing or
+dropping pagination rows. Duplicate top-level labels remain rejected because
+the relational label key is unique and no deterministic row selection exists.
```
chrisguiney, line 77: Blocking: the branch still conflicts with `main`. `git merge origin/main` stops in this file only, where #4161 adds its `utxo_collateral_input` paragraph after this line; `main`'s side resolves it, and the migration list merges cleanly to 14. 068c9f12 is a single-parent commit copying this file from `main` at bdacc83e, so the merge base is still 2f465dba and the checks on this head ran before #4161 and #4295 reached `main`. A two-parent merge of current `main` clears both.

connmanager/connection_manager.go
```diff
@@ -968,6 +977,9 @@ func (c *ConnectionManager) RemoveConnection(
 	if info != nil && info.ipKey != "" {
 		c.releaseIPSlot(info.ipKey)
 	}
+	if info.onClose != nil {
+		info.onClose()
```
chrisguiney, line 981: Blocking: this release and the collision-replacement releases at 805 and 852 move ownership to the registered entry, but `ARCHITECTURE.md:708` at this head still says setup transfers its release to the close watcher, which is `main`'s pre-change wording. 068c9f12 took that file from `main` without re-applying this change's paragraph (3ecb7876 lines 643-648: release before `ConnClosedFunc`, outside the map lock, once-only). Re-add it after merging `main`; `make docs-parity` passes with or without it.

chrisguiney (changes requested):
Re-review at 068c9f12. That commit only replaces `ARCHITECTURE.md` with `main`'s copy at bdacc83e; the Go diff is unchanged since 3ecb7876. The earlier regression is gone: against current `main` the file lacks only #4161's two symbols and restores none. Two blockers remain, inline. Merging `main` with `main`'s side plus this change's paragraph builds and passes `go test -race ./connmanager/... ./internal/config/...`, the root option tests, scoped lint and `make docs-parity`. The 723/762 rejection-release points stand, non-blocking.

connmanager/connection_manager.go
chrisguiney, line 994: Blocking: this release and the collision-replacement releases at 805 and 852 move admission ownership to the registered entry, but ARCHITECTURE.md:715 at this head still says successful setup transfers its release to the connection close watcher, which is main pre-change wording. The merge took that file from main at 8302c7ec and this branch now adds no line to it, so the shipped contract is documented nowhere. Re-apply the paragraph: release outside the map lock, before ConnClosedFunc, once-only. make docs-parity passes either way.

connmanager/connection_manager.go
chrisguiney: Blocking, confirming arepala-uml: this call and the one at 762 are pinned by no test. Commenting both out leaves go test -race ./connmanager/ green, while reverting the release at 981 fails TestNtCAdmissionSlotReleasedBeforeCloseCallback and TestNtCCleanupNormalCloseBeforeBlockedCallback, and reverting 805 or 852 alone fails TestNtCCleanupCollisionReleasesBeforeBlockedCallback. The defer at 879 is unpinned the same way. Drop all three (setupAcceptedConnection already defers releaseNtCSlot, listener.go:375) or pin them with an addConnectionImpl test using a counting callback.

config.go
```diff
@@ -1408,6 +1409,26 @@ func WithMaxInboundConns(n int) ConfigOptionFunc {
...
```
chrisguiney, line 1414: Non-blocking, still open: neither this option nor WithMaxNtCConnectionsPerIP has a test reference anywhere in the tree, where TestPeerGovernorOptionsIgnoreNonPositiveValues and TestPeerGovernorOptionsApplyPositiveValues cover WithMaxConnectionsPerIP and WithMaxInboundConns for both the ignored non-positive case and the applied case. A line in each, plus an assertion that syncCompatFields mirrors both onto the fields node.go and node_lifecycle.go read, would pin the wiring that makes these two limits reachable at all.

internal/config/config.go
```diff
@@ -699,6 +699,8 @@ type Config struct {
...
```
chrisguiney: Style, still open: the envconfig column starts one space later on both of these lines than on every other field in this block, and gofmt cannot correct it because the padding sits inside the tag string.

chrisguiney (dismissed):
Merge audit at b5aded16. The merge takes ARCHITECTURE.md and DATABASE.md byte-for-byte from main at 8302c7ec, so the earlier documentation regression is resolved: no file this branch changed is reverted, nothing main removed returns, and the 54 documented symbols this head lacks are all from the 13 commits it is behind. Merging current main is clean and the merged tree passes build, vet, gofmt, scoped lint, race connmanager tests, conformance, docs-parity and import-boundaries. Two blockers remain, inline; the option-test and tag-alignment points stay open.

connmanager/connection_manager.go
```diff
@@ -968,6 +971,9 @@ func (c *ConnectionManager) RemoveConnection(
...
```
arepala-uml, line 994: This sentence says successful setup transfers its release to the connection's close watcher, but connection_manager.go now releases via RemoveConnection or a collision eviction of the registered entry, not solely the close watcher. Update the sentence to describe both release paths.

arepala-uml (changes requested):
ARCHITECTURE.md:722 still says setup release transfers to the connection's close watcher; the fix in this PR moves it to the registered connection entry, released by RemoveConnection or a collision eviction. The fix itself verifies causally: reverting it fails TestNtCCleanupCollisionReleasesBeforeBlockedCallback, and build/vet/go test -race ./connmanager/... are otherwise clean.

arepala-uml (approved):

PR #4107, opened 2026-09-07
fix(chainsync): update tracked client state on rollback

### LINKED ISSUE #3423: chainsync: update tracked client state on rollback
At current main (`eabf653d27e4f4d21d428d5fb1a6bfdb5925e08d`), `ouroboros/chainsync.go` `chainsyncClientRollBackward` does not directly update the separate `chainsync.State` tracked client's `Cursor`/`Tip` or `LastActivity`/`Status` fields.

The chain selector maintains distinct state: its rollback handler updates the advertised/observed peer tips and activity timestamp, trims history, and reevaluates selection. Local untargeted resync also rewinds ahead tracked-client cursors and refreshes status/activity, but does not update their tip. These controls do not establish an atomic update of all tracked-client fields for each wire rollback; the issue should not imply that all peer-selection state remains unchanged.

Acceptance criteria:

- Update all tracked client fields atomically with the rollback.
- Add rollback tests that assert cursor, tip, activity, and status after the event.

wolf31o2:
Update tracked client cursor, tip, activity, and status atomically on admitted rollbacks without changing header counts or deduplication. Add rollback sequencing, disconnected-client, timestamp, and hash-ownership coverage.

Fixes #3423.

### DIFF
--- a/chainsync/chainsync.go
+++ b/chainsync/chainsync.go
@@ -795,6 +795,29 @@ func (s *State) UpdateClientTipWithoutDedup(
 	return s.updateTrackedClientTip(connId, point, tip)
 }
 
+// UpdateClientRollback updates an existing client's cursor, advertised tip,
+// activity, and syncing status atomically. Rollbacks do not count as delivered
+// headers or enter the header deduplication cache.
+func (s *State) UpdateClientRollback(
+	connId ouroboros.ConnectionId,
+	point ocommon.Point,
+	tip ochainsync.Tip,
+) bool {
+	s.clientConnIdMutex.Lock()
+	defer s.clientConnIdMutex.Unlock()
+	tc, exists := s.trackedClients[connId]
+	if !exists {
+		return false
+	}
+	point.Hash = cloneBytes(point.Hash)
+	tip.Point.Hash = cloneBytes(tip.Point.Hash)
+	tc.Cursor = point
+	tc.Tip = tip
+	tc.LastActivity = time.Now()
+	tc.Status = ClientStatusSyncing
+	return true
+}
+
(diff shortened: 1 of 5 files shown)


arepala-uml (approved):

chainsync/chainsync.go
```diff
@@ -795,6 +795,29 @@ func (s *State) UpdateClientTipWithoutDedup(
 	return s.updateTrackedClientTip(connId, point, tip)
 }
 
+// UpdateClientRollback updates an existing client's cursor, advertised tip,
+// activity, and syncing status atomically. Rollbacks do not count as delivered
+// headers or enter the header deduplication cache.
+func (s *State) UpdateClientRollback(
```
chrisguiney, line 801: Rollback-path class audit, recorded so it does not have to be redone:

`chainsyncClientRollBackward` is the only production `WithRollBackwardFunc` registration (`ouroboros/chainsync.go:290`). The other `RollBackward` handlers in the tree belong to the devnet observer, `erastest`, `archive-demo`, and `nodeparity` harnesses, so this one call site covers the wire-rollback class.

`RewindTrackedClientsTo` is the other cursor-rewind path issue #3423 mentions as not updating `Tip`. Leaving it unchanged reads as correct rather than as a gap: it rewinds on a *local* ledger recovery, where the peer's advertised tip has not changed, and it already sets `Cursor`, `Status`, and `LastActivity` under this same mutex.

One adjacent observation for a separate change, not this one: `updateTrackedClientTip` assigns `point` and `tip` without the `cloneBytes` treatment used here, so the two tracked-state writers differ in whether they take ownership of the caller's hash buffers. That asymmetry predates this diff, and this method is the stricter of the two.

chrisguiney (approved):

Issue #3412, opened 2026-08-24
leios: bound backfill, cursor, announcement, and BlockRange lifecycles


wolf31o2:
Leios lifecycle follow-up. At `eabf653d27e4f4d21d428d5fb1a6bfdb5925e08d`, two original mechanisms have changed:

- `FetchEndorserBlockByPoint` shares the caller's deadline, or a two-minute default, across connection attempts. Initial binding/publication and cache lookup precede that budget; complete cancellation of the call still requires verification.
- `leiosfetchServerBlockRangeRequest` returns an error instead of silently retaining agency. Selected gouroboros v0.202.10 propagates that error. Shared-bearer teardown and peer recovery still require verification; block-range serving is not implemented.

Remaining original scope:

- Verify closed connections cannot re-register forged-EB cursors or retain entries after removal.
- Deferred announcements have a shared 128-entry cap without a per-source quota; verify bounded retention and retry/drain behavior.
- Announcement election keys include the source; enforce the intended budget across reconnects.
- Verify manifest offers claim work across connections before dispatch.
- `VotesByIds` allocates for the requested ID count and walks/clones matches under the global vote-manager lock without a local semantic count cap.
- Verify persist-writer lifecycle resets cannot overlap incorrectly with active writers.

Acceptance criteria:

- Enforce budgets across reconnects and claim in-flight work before dispatch.
- Bound deferred announcements and vote ID processing per peer.
- Remove cursors exactly once and synchronize persist-writer lifecycle resets.
- Verify BlockRange error propagation produces observable, recoverable protocol termination rather than a silent hang.
- Cover reconnect, duplicate-offer, shutdown, and concurrent lifecycle cases, including the complete backfill cancellation boundary.

The source changes above do not establish completion of the compound lifecycle acceptance criteria.

PR #4108, opened 2026-09-07
fix(ouroboros): release idle LeiosNotify requests on disconnect

wolf31o2:
Release idle LeiosNotify requests from connection-close notifications and check cancellation before reserving log entries. Add disconnect, cursor cleanup, pending-offer, and completed-protocol coverage.

Related to #3412.

## Summary by CodeRabbit

- **Bug Fixes**
  - Idle LeiosNotify requests now exit promptly when their connection closes.
  - Connection closure is handled consistently for both node-to-client and node-to-node connections.
  - Closed requests no longer recreate removed event-log cursors.
  - Undelivered offers remain available for retry after a disconnect.
  - Completed protocols avoid creating unnecessary cursors.

- **Documentation**
  - Added architecture documentation covering connection-close handling and cleanup behavior.

- **Tests**
  - Expanded coverage for disconnects, waiter release, cursor cleanup, retries, and protocol completion.


arepala-uml (changes requested):
Requesting changes for the merge conflict only — the lifecycle/concurrency fix itself is correct and well-tested.

**Blocking: merge conflict.** GitHub reports this branch as CONFLICTING against main. Traced it to a single-hunk conflict in `node_ntc_conn_closed_test.go`: PR #4131 (merged to main after this branch was cut) added `t.Parallel()` to the test this PR independently refactored into an NtC/NtN pair via a new shared helper (`testHandleConnManagerClosedReleasesLeiosServeWaiters`), dropping that line in the process. Please rebase onto current main and carry `t.Parallel()` into the new helper. CI is green here only because it builds this branch directly, not the merge result, so it doesn't reflect mergeability.

**Code review (no other blockers):**
- Confirmed the actual bug: `handleConnManagerClosed` released idle LeiosNotify server requests only for NtC connections (`if !isNtC { return }` skipped it for NtN), leaking a blocked goroutine/cursor on NtN disconnect. Verified the fix by selectively reverting `node.go` and `ouroboros/leios_merged.go` individually in a worktree — the corresponding new tests fail against each pre-fix file, confirming they catch the regressions.
- Cancellation-before-reservation is correctly atomic under `leiosForgedEBLog.mu` — no TOCTOU gap.
- Cubic's one finding (P2, leios_merged.go:217, fd-reuse could theoretically let a reused `ConnectionId` delete a new connection's cursor) is real but pre-existing — the same fd-reuse risk already exists in `chainsyncState.RemoveClient`. Not new to this PR; suggest a follow-up issue for a monotonic connection-generation id rather than blocking here.
- `go build`/`vet`/`test -race` all clean, no `time.Sleep`-based test synchronization, ARCHITECTURE.md accurately updated.

Happy to re-review once rebased.

ouroboros/leiosnotify.go
```diff
@@ -1879,19 +1883,25 @@ func (o *Ouroboros) leiosnotifyServerRequestNext(
 		return nil, nil
 	}
 	connKey := leiosConnectionIdString(ctx.ConnectionId)
-	done := ctx.Server.DoneChan()
+	// Protocol completion waits for this callback to return. Use the
+	// connection manager's independent close notification instead.
+	done, cancel := o.registerLeiosServeWaiter(ctx.ConnectionId)
```
chrisguiney: Blocking. `registerLeiosServeWaiter` treats "not currently in the connection manager" as "the connection is gone" (`ouroboros/leios_merged.go:166-169`), and this callback now turns that into a `(nil, nil)` return. gouroboros rejects a nil response — `protocol/leiosnotify/server.go` in v0.204.0, `handleRequestNext`:

```go
resp, err := s.config.RequestNextFunc(s.callbackContext)
if err != nil {
    return err
}
if resp == nil {
    return errors.New(
        "received leios-notify NotificationRequestNext message but callback returned nil",
    )
}
```

`protocol/protocol.go:947-952` hands that error to `SendError`, which stops the protocol and reports on the connection's ErrorChan, taking chainsync/blockfetch on the same bearer with it. Before this change `done` was `ctx.Server.DoneChan()`, so `(nil, nil)` was reachable only when the protocol was already stopping and the error was moot.

Two windows make it reachable while the connection is alive:

- `ouroboros.NewConnection()` returns with the handshake complete and the leios-notify server already able to receive `MsgRequestNext`, but the connection is registered only afterwards — `connmanager/listener.go:423` then `:472`/`:478`, and `connmanager/outbound.go:102` then `:127`. A peer that sends `MsgRequestNext` in that gap has its connection dropped.
- `notifyEvictedConnectionClosed` (`connmanager/connection_manager.go:905-913`) fires synchronously *before* the replacement is stored at `connection_manager.go:829`, so on a `ConnectionId` collision the replacement is unregistered for that window — and `ReleaseLeiosServeWaiters(connId)` additionally closes any waiter the replacement already registered under the same ID.

Calling `leiosnotifyServerRequestNext` with a live server and a `ConnectionId` absent from the connection manager returns `(nil, nil)` immediately; with `ouroboros/leiosnotify.go` at its pre-change content the same call parks instead, so the nil return is new here rather than pre-existing. `ouroboros/chainsync_server_fixture_test.go` already records this dependency for the chainsync path: "Without that registration the lookup fails and the callback tears the connection down instead of serving."

The waiter needs to distinguish "connection is gone" from "connection is not registered yet". One option is to skip waiter registration in this callback when `GetConnectionById` returns nil and fall back to `protocolDone`, leaving the strict liveness check to the chainsync closure wait that needs it.

ouroboros/leiosnotify_disconnect_test.go
```diff
@@ -0,0 +1,121 @@
...
+		defer close(finished)
+		_, _ = o.leiosnotifyServerRequestNext(leiosnotify.CallbackContext{Server: server})
+	}()
+	testutil.RequireReceive(t, finished, time.Second, "completed protocol callback to return")
+	require.Empty(t, o.leiosEBLog.cursors)
+	require.Empty(t, o.leiosServeWaiters)
+}
+
+// next lets log-only tests reserve entries without a connection lifecycle.
+func (l *leiosForgedEBLog) next(connKey string) (*leiosForgedEBEntry, chan struct{}) {
```
chrisguiney, line 141: Non-blocking. Defining `next` here routes every existing `leiosForgedEBLog` test in `leios_forge_serve_test.go` through a shim that always passes a nil `done`, so none of them exercise the cancellation branch just added to the production signature.

`TestLeiosNotifyClosedReservationDoesNotRecreateCursor` covers that branch directly, so it is not untested — but the shim means the log's own suite no longer feeds production-shaped input through the function under test, and a later change to the cancellation semantics would not show up there. Passing `nil` explicitly at those call sites would keep it visible.

chrisguiney (changes requested):
Requesting changes on one blocker: the new `done` source in `leiosnotifyServerRequestNext` lets the callback return a nil message on a live protocol, which gouroboros converts into connection teardown. Detail inline.

The rest holds up. Releasing serve waiters for NtN as well as NtC is the right fix, and the cancel-then-remove ordering is correct — `nextWhileConnected` checks `done` while holding `leiosForgedEBLog.mu`, and `removeConn` needs that same lock, so a closing request cannot recreate a removed cursor. Waiter close is single-shot: both release paths remove the channel from the map before closing it, so they cannot double-close.

Teardown-path audit — normal NtN/NtC close and keepalive failure (the error watcher's `ConnClosedFunc`), eviction and same-direction replacement (`notifyEvictedConnectionClosed`), node shutdown (`Stop` closing connections into that same watcher), protocol completion (the new `protocolDone` case), and already-closed-at-registration (the liveness re-check). All of them reach a release, and a pre-handshake failure never reaches the callback. The class is complete. The defect is that one member of it also fires while the connection is still alive.

The branch remains CONFLICTING against main, with `node_ntc_conn_closed_test.go` the only conflicting file.

`ARCHITECTURE.md` is updated for the lifecycle change, as AGENTS.md requires. Checks on this head are green and the package is clean under `-race`; the defect is semantic, so nothing mechanical catches it.

ouroboros/leios_merged.go
```diff
@@ -211,6 +210,12 @@ func (o *Ouroboros) ReleaseLeiosServeWaiters(
 	for _, ch := range waiters {
 		close(ch)
 	}
+	// Cancel before removing the cursor: a racing notification request checks
+	// its cancellation channel under the same log lock before registering.
+	// This also covers a request returning an offer during disconnection.
+	if o.leiosEBLog != nil {
+		o.leiosEBLog.removeConn(leiosConnectionIdString(connId))
```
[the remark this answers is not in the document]
chrisguiney: Valid root cause, but the impact as stated is pre-existing rather than introduced here. `HandleConnClosedEvent` already calls `leiosEBLog.removeConn` for the same `ConnectionId` (`ouroboros/ouroboros.go:789`), so a replacement reusing an evicted connection's ID could already lose its cursor before this change. Line 217 duplicates that call rather than creating the hazard, so this is not a blocker on this PR — though it does mean cursor removal now runs twice per close, which is worth noting against #3412's "remove cursors exactly once" criterion even though `removeConn` is idempotent.

The underlying `ConnectionId` ambiguity does get worse here, in a different place. `ReleaseLeiosServeWaiters` also closes the serve waiters keyed by that same `ConnectionId`, and after this change those waiters decide whether a live `leiosnotifyServerRequestNext` returns a nil message — which gouroboros converts into connection teardown. `notifyEvictedConnectionClosed` (`connmanager/connection_manager.go:905-913`) runs before the replacement is stored at `connection_manager.go:829`, so on a collision the replacement's callback can be released while its connection is alive. That is the blocker raised on `ouroboros/leiosnotify.go`. A monotonic connection-generation id would settle both.

wolf31o2:
Updated the LeiosNotify lifecycle path to avoid treating the connection-manager publication window as a disconnect. Requests received before registration now remain governed by protocol completion, while registered connections continue to use independent close notifications. Added regression coverage for the live-protocol registration window and retained NtC/NtN close coverage. Follow-up review requested.

ouroboros/leios_merged.go
```diff
@@ -162,15 +161,37 @@ func (o *Ouroboros) registerLeiosServeWaiter(
...
+// connection that is currently managed. A LeiosNotify server can receive its
+// first request before connmanager publishes the connection; treating that
+// short registration window as a disconnect would return a nil response and
+// make gouroboros tear down an otherwise live protocol. In that case the
+// caller falls back to the protocol completion channel.
+func (o *Ouroboros) registerLeiosNotifyServeWaiter(
+	connId ouroboros.ConnectionId,
+) (done <-chan struct{}, cancel func()) {
+	if o.connManager != nil && o.connManager.GetConnectionById(connId) == nil {
+		return nil, func() {}
```
chrisguiney: **Blocking.** On this path `done` is nil, and nothing can release the callback.

`protocolDone` is `ctx.Server.DoneChan()`. For a started protocol that channel is closed only by the goroutine at gouroboros v0.204.0 `protocol/protocol.go:199-202`, which waits on `recvDoneChan` first; `recvDoneChan` closes only when `recvLoop` returns (`protocol.go:921`). `recvLoop` is the goroutine that invokes this callback — `protocol.go:947` → `leiosnotify/server.go` `messageHandler` → `handleRequestNext` → `RequestNextFunc`. So while the callback is parked, `recvLoop` cannot return and `protocolDone` cannot close. That is the same fact the comment at `ouroboros/leiosnotify.go:1888-1889` states, and the reason this PR moved the release off protocol completion in the first place.

With `done == nil` the loop's `select` at `leiosnotify.go:1912-1919` therefore has no live exit: `done` never fires, `protocolDone` cannot close, and `wakeCh` only re-enters the loop. `ReleaseLeiosServeWaiters` cannot help either, because this path registered no waiter. The request parks for the lifetime of the process.

Two things follow:

- The nil lookup is not only the pre-publication window. `registerLeiosServeWaiter`'s own comment just above (lines 128-133) records that connmanager removes a connection from its map *before* invoking `ConnClosedFunc`, so a `MsgRequestNext` arriving during teardown reaches this same branch — and for that one the close notification has already run or is about to, and will never reach it.
- Once parked, the next `append` wakes it through `wakeCh`, and it calls `nextWhileConnected(connKey, nil)` at `leiosnotify.go:1906`, re-registering a cursor after `removeConn` has already run. That is the orphaned cursor CodeRabbit flagged at `leiosnotify.go:1918`, and it is the "verify closed connections cannot re-register forged-EB cursors" criterion in #3412.

A direction that covers both windows: branch on whether the close notification has already fired for that `ConnectionId` rather than on absence from the connection manager. Recording the ids `ReleaseLeiosServeWaiters` has already run for (bounded, or cleared when connmanager publishes that id again) separates the two cases — absent-and-already-released is genuinely gone, so the closed channel `registerLeiosServeWaiter` returns is correct; absent-and-never-released is the pre-publication window, where registering normally is safe because the eventual close notification will close the waiter.

ouroboros/leiosnotify_disconnect_test.go
```diff
@@ -0,0 +1,146 @@
...
+		_, err := o.leiosnotifyServerRequestNext(leiosnotify.CallbackContext{
+			Server:       server,
+			ConnectionId: gouroboros.ConnectionId{},
+		})
+		finished <- err
+	}()
+	t.Cleanup(server.Stop)
+	testutil.RequireNoReceive(t, finished, 50*time.Millisecond,
+		"an unregistered connection must not be treated as a live-protocol disconnect")
+	server.Stop()
```
chrisguiney: Non-blocking, and a consequence of the finding on `ouroboros/leios_merged.go`. This `server` is never started, so `Stop()` closes the done channel directly through gouroboros v0.204.0 `protocol/protocol.go:224-225` (`if !started { p.closeDone() }`). A started protocol never takes that path: `DoneChan` closes only after `recvLoop` returns, and `recvLoop` is the goroutine parked inside this callback. So the release this asserts is one the production path cannot perform.

`TestLeiosNotifyIdleRequestReleasedByDisconnect` in this file is the shape that does exercise it — a real managed connection, a real disconnect, and an explicit assertion that the protocol done channel stayed open.

chrisguiney (dismissed):
Re-review at eadfa07e. Requesting changes on one blocker: the nil-`done` fallback added here cannot be released on a live protocol, so an idle request that takes it parks for the lifetime of the process — the failure mode this PR exists to remove, in a narrower window. Detail inline on `ouroboros/leios_merged.go`.

Resolved since the previous review:

- `leiosnotifyServerRequestNext` no longer returns `(nil, nil)` for a connection that is merely not yet published, so the connection teardown raised last round is not reachable through that path.
- The branch is no longer conflicting, and `t.Parallel()` is carried into `testHandleConnManagerClosedReleasesLeiosServeWaiters`.

Unchanged and still correct: releasing serve waiters for NtN as well as NtC, and the cancel-then-remove ordering in `ReleaseLeiosServeWaiters` — closing the waiters before `removeConn` means an append racing the close cannot leave a cursor behind on the managed path. `TestLeiosNotifyIdleRequestReleasedByDisconnect` drives that path through a real managed connection and asserts the protocol done channel was not used, which is the right shape for this fix.

Checks on this head are green, including `go-test (Linux, race)` and `lint`; the remaining defect is semantic and nothing mechanical catches it.

ouroboros/leiosnotify.go
```diff
@@ -1904,6 +1914,8 @@ func (o *Ouroboros) leiosnotifyServerRequestNext(
 			// new EB appended — re-check
 		case <-done:
 			return nil, nil
+		case <-protocolDone:
+			return nil, errors.New("leios-notify protocol closed")
```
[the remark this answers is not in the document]
chrisguiney: Confirmed as a real leak, but the suggested `removeConn` at line 1918 does not reach it. In a started protocol `protocolDone` cannot close while this callback is parked — gouroboros v0.204.0 `protocol/protocol.go:199-202` waits on `recvDoneChan`, which closes only when `recvLoop` returns (`:921`), and `recvLoop` is the goroutine that called this callback (`:947`). So that arm is not the exit the orphaned cursor takes; the callback does not exit at all.

The cursor is re-created at line 1906 by `nextWhileConnected(connKey, nil)` on the next `append`, after `ReleaseLeiosServeWaiters` already ran `removeConn`. The cause is the nil `done` returned by `registerLeiosNotifyServeWaiter`; fixing that removes the leak without a second `removeConn` call site. Raised as blocking on `ouroboros/leios_merged.go`.

ouroboros/leiosnotify.go
```diff
@@ -1881,19 +1885,25 @@ func (o *Ouroboros) leiosnotifyServerRequestNext(
 		return nil, nil
 	}
 	connKey := leiosConnectionIdString(ctx.ConnectionId)
-	done := ctx.Server.DoneChan()
+	// Protocol completion waits for this callback to return. Use the
+	// connection manager's independent close notification instead.
+	done, cancel := o.registerLeiosNotifyServeWaiter(ctx.ConnectionId)
```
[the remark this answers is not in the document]
chrisguiney: Confirmed, and this is the blocking finding on this head.

One extension: it is not only the already-removed connection. The same nil `done` is returned for the not-yet-published window this branch was added for, and that request is not released either — `protocolDone` cannot close while `recvLoop` is inside the callback (gouroboros v0.204.0 `protocol/protocol.go:199-202`, `:921`, `:947`), and no waiter was registered for `ReleaseLeiosServeWaiters` to close. Both windows need to be distinguished from each other, not just from the live case. Detail and a suggested direction on `ouroboros/leios_merged.go`.

ouroboros/leiosnotify_disconnect_test.go
```diff
@@ -0,0 +1,146 @@
...
+	select {
+	case err := <-finished:
+		require.Error(t, err)
+	case <-time.After(time.Second):
```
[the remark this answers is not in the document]
chrisguiney: The `testutil.RequireReceive` point is valid: AGENTS.md requires the helper and the rest of this file uses it.

The flake mechanism is different from the one described. With `done == nil` the first `select` cannot return through `case <-done`, because a nil channel never fires. It can return early through `case <-protocolDone` at `ouroboros/leiosnotify.go:1900-1901`, if the goroutine reaches the callback after `Stop` — and that arm returns `nil, nil`, so `require.Error` fails. The 50ms `RequireNoReceive` makes that unlikely rather than impossible. The two `protocolDone` arms returning different error contracts (line 1901 versus line 1918) is worth settling either way.

ouroboros/leios_merged.go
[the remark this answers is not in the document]
chrisguiney: Still stands at eadfa07e: `HandleConnClosedEvent` calls `leiosEBLog.removeConn` for the same `ConnectionId` at `ouroboros/ouroboros.go:788`, so line 239 duplicates an existing call rather than creating the reuse hazard. Not a blocker on this PR. The underlying `ConnectionId` ambiguity is pre-existing and belongs to a separate change.

ouroboros/leiosnotify.go
chrisguiney: Resolved at eadfa07e. `registerLeiosNotifyServeWaiter` no longer returns a closed channel for a connection that is merely unpublished, so this callback no longer returns `(nil, nil)` on a live protocol and gouroboros no longer tears the bearer down for it.

The replacement fallback has its own defect: the nil `done` it returns leaves the callback with no reachable exit. Raised on `ouroboros/leios_merged.go`.

ouroboros/leiosnotify_disconnect_test.go
chrisguiney: Still present at eadfa07e, unchanged and still non-blocking.

arepala-uml (approved):

PR #4115, opened 2026-09-07
fix(blockfrost): bound transaction request body reads

wolf31o2:
Apply a 15-second body-read deadline to transaction submission and evaluation, retaining expired deadlines so HTTP cleanup cannot stall error responses.

Regression coverage includes complete, truncated, and stalled bodies on all three routes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Added request-body timeouts for transaction submission and evaluation requests, preventing stalled connections from waiting indefinitely.
  - Preserved existing responses for oversized, truncated, or unreadable request bodies.
  - Added a 60-second server read-timeout safeguard for incomplete requests.

- **Documentation**
  - Documented request-body limits, timeout behavior, and handling of stalled or truncated requests.


api/blockfrost/blockfrost.go
```diff
@@ -393,3 +400,28 @@ func (b *Blockfrost) Stop(
...
+	limit int64,
+) ([]byte, error) {
+	rc := http.NewResponseController(w)
+	setDeadline := func(deadline time.Time) {
+		if err := rc.SetReadDeadline(deadline); err != nil &&
+			!errors.Is(err, http.ErrNotSupported) {
+			b.logger.Debug("could not bound request body read", "error", err)
+		}
+	}
+	setDeadline(time.Now().Add(b.requestBodyTimeout))
```
arepala-uml: The 15s deadline is only installed once `readRequestBody`/`readEvaluationBody` runs, which is after `requireContentType`/`requireOgmiosVersion` in the evaluate handlers (evaluate.go:138-164) and after the earlier checks in `handleTransactionSubmit` (handlers.go:1212-1227). A request with a rejected Content-Type/Ogmios version whose body then stalls skips this 15s bound and falls back to the much weaker 60s server-wide `ReadTimeout`. Not a regression (pre-fix this path was fully unbounded), but it doesn't fully match the PR's stated "15-second deadline on these three routes" guarantee. Worth moving the deadline-set earlier, or noting the gap explicitly.

ARCHITECTURE.md
```diff
@@ -5906,6 +5906,12 @@ when `Stop` returns, and the address is rebindable afterwards.
 
 ### Blockfrost API (`api/blockfrost/`)
 
+Blockfrost submission and both evaluation endpoints bound request body reads
+by their existing byte limits and a 15-second read deadline. The deadline is
+cleared after a successful read, before transaction processing; stalled or
+truncated bodies retain the existing HTTP 400 response. The HTTP listener also sets a 60-second read timeout
```
arepala-uml, line 5912: This line is 108 characters, over the repo's 80-char `make golines` limit. `make golines` should wrap it.

arepala-uml (commented):
Verified the fix locally: `go build`/`go vet`/`golangci-lint` clean, `go test -race ./api/blockfrost/...` passes (24.9s, tests inject a short deadline instead of waiting the real 15s). Confirmed pre-fix code hangs indefinitely on a stalled body via a throwaway repro against the parent commit, so this closes a real unbounded-read hang on all three routes. Two non-blocking items below worth a follow-up.

arepala-uml (changes requested):
Upgrading my earlier comment-only review to request changes so the two open items below are clearly tracked as blocking, not just notes:

1. The 15s body-read deadline is only installed once `readRequestBody`/`readEvaluationBody` runs, which is after the Content-Type/Ogmios-version checks in the evaluate handlers and the earlier checks in `handleTransactionSubmit`. A request that fails those checks but then stalls its body falls back to the much weaker 60s server-wide `ReadTimeout` instead of the claimed 15s bound for these three routes (see inline comment on blockfrost.go:417).
2. ARCHITECTURE.md:5912 is 108 characters, over the repo's 80-char `make golines` limit.

Everything else checks out: fix is correct and closes a real unbounded-read hang, tests are fast and well-designed, build/vet/lint/race are clean, and the historical #3217 listener-lifecycle concern doesn't apply here. Once these two are addressed I'm happy to approve.

arepala-uml (approved):

PR #4124, opened 2026-09-07
fix(koiosparity): serialize cache writers to avoid SQLITE_BUSY

chrisguiney:
## Defect

`SaveAccountFetchChunkProgress` fails with `database is locked (5) (SQLITE_BUSY)`
under koiosparity's own designed concurrency. `fetch_accounts.go` dispatches
`accountFetchConcurrency` (5) chunk workers and `OpenCache` never bounded the
connection pool, so those workers run on distinct SQLite connections contending
for the single per-database WAL writer slot. Each chunk transaction's first
statement is a `DELETE`, which claims that slot immediately even when it matches
zero rows, and holds it for the whole transaction. Five workers' large-chunk
writes can collectively outlast `busy_timeout=5000`; the epoch is then abandoned
after 5 attempts, silently under `--koios-parity-strict=false`.

The issue attributed this to a DEFERRED-transaction snapshot upgrade bypassing
`busy_timeout`. That variant is real in this driver but is not what this path
hits: `_txlock=immediate` was tried against the reproduction and does not fix it,
because the transaction's first statement is already write-intending.

## Change

- `OpenCache` sets `SetMaxOpenConns(1)`. `database/sql` then queues callers for
  the single connection with no deadline, so an in-progress writer is waited out
  rather than timed out on. The `busy_timeout` comment, which claimed it
  prevented this failure, is corrected. This also serializes the cache's own
  reads; it is a process-local comparison scratch file, and `dingo_db.go`'s
  separate `*sql.DB` is unaffected.
- `createCacheSchema`'s legacy-column migration held an open `*sql.Rows` from its
  `pragma_table_info` probe while issuing `ALTER TABLE ... DROP COLUMN` on the
  same pool. Under a one-connection pool that `Exec` waits for the connection the
  probe still holds, with no deadline, so `OpenCache` hangs with no error against
  any older `cache.db` still carrying `pool_cnt`, `delegator_cnt` or
  `deposits_d_rep`. The probe is extracted into `columnExists`, which closes its
  rows before returning; `addColumnIfMissing` shared the shape and now uses the
  same helper.

## Tests

- `TestSaveAccountFetchChunkProgressWaitsOutSlowConcurrentWriter` — deterministic
  reproduction: one connection holds the writer slot for 5.5s (past
  `busy_timeout`) while a second goroutine calls the real
  `SaveAccountFetchChunkProgress`. Fail-before verified by reverting only
  `cache.go` in place: FAIL in 5.09s with the issue's exact
  `database is locked (5) (SQLITE_BUSY)` line, PASS with the change.
- `TestOpenCacheLegacyColumnMigrationDoesNotDeadlock` — adds the three legacy
  columns, reopens under a timeout guard, asserts the migration still drops them.
  Fail-before verified against the pool-bound commit without `columnExists`: it
  hangs and fails after 30s.
- `TestSaveAccountFetchChunkProgressConcurrentWritersDoNotHitSQLiteBusy` — drives
  the real 5-worker dispatch width through the public API. It passes before the
  fix too and is documented as a guard, not as fail-before evidence.

## Checks

`go build -tags dingo_extra_plugins ./...`,
`go vet -tags dingo_extra_plugins ./internal/...`, `gofmt -s -l`,
`golangci-lint run ./internal/koiosparity/...` (0 issues),
`go test -race -count=1 ./internal/koiosparity/`,
`go test ./internal/architecture ./internal/docsparity` — all pass.

Skipped: the live Preview from-genesis parity replay against a Koios host (no
live Cardano infrastructure available in this environment), Antithesis and
conformance suites (not applicable and not available), `nilaway` and `modernize`
(not installed here; CI's lint job runs golangci-lint only).

https://claude.ai/code/session_01CY3D9bzbTUF6A9Y2KDUrLi

---
## Summary by cubic
Fixes SQLITE_BUSY failures in koiosparity's concurrent chunk workers by limiting the cache to one SQLite connection, so writers queue instead of exhausting `busy_timeout`.

- `OpenCache` now calls `SetMaxOpenConns(1)`; the cache is a process-local scratch file, so serialized reads are acceptable.
- Extracts the column probe into `columnExists` so its rows close before the migration's `ALTER TABLE`, preventing an `OpenCache` hang against older cache files.
- Adds tests for a slow concurrent writer held past `busy_timeout`, the real 5-worker dispatch, and the legacy column migration.

<sup>Written for commit 81ba6ba16dd1b145be72735dbab5b2227e60bb72. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4124?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cache reliability during concurrent operations, reducing SQLite busy and timeout errors.
  - Prevented potential hangs while upgrading older cache databases.
  - Ensured database schema updates complete safely under concurrent activity.

- **Tests**
  - Added regression coverage for concurrent cache writes, parallel chunk processing, and legacy database migrations.

### DIFF
--- a/internal/koiosparity/cache.go
+++ b/internal/koiosparity/cache.go
@@ -294,6 +294,27 @@ func OpenCache(path string, logger *slog.Logger) (*Cache, error) {
 	if err != nil {
 		return nil, fmt.Errorf("open cache db: %w", err)
 	}
+	// Every Cache write path opens its transaction with c.db.Begin(), which
+	// is DEFERRED: SQLite still treats the very first statement of a write
+	// transaction as write-intending even when it matches zero rows (e.g.
+	// SaveAccountFetchChunkProgress's opening DELETEs on a brand-new chunk),
+	// and grabs SQLite's single per-database WAL writer slot right there —
+	// for the rest of the transaction, not just its final COMMIT. Left
+	// unbounded, database/sql hands accountFetchConcurrency's concurrent
+	// chunk workers (fetch_accounts.go) distinct connections that genuinely
+	// contend for that one slot; busy_timeout=5000 only covers a wait
+	// shorter than 5s; five real chunk workers' large-chunk Prepare/Exec
+	// work can collectively outlast that and fail with SQLITE_BUSY /
+	// "database is locked" (dingo #4091). Restricting the pool to one
+	// connection makes database/sql itself queue every caller for that
+	// single connection with no fixed budget, so a writer already in
+	// progress is always waited out rather than timed out on. This also
+	// serializes the cache's own reads behind any in-flight write — the
+	// cache is a process-local comparison scratch file, not a
+	// high-throughput read service, so that tradeoff is preferred over a
+	// second unbounded connection. It has no effect on dingo_db.go's
+	// separate *sql.DB against Dingo's own node database.
+	db.SetMaxOpenConns(1)
 	if err := db.Ping(); err != nil {
 		_ = db.Close()
 		return nil, fmt.Errorf("ping cache db: %w", err)
@@ -303,8 +324,11 @@ func OpenCache(path string, logger *slog.Logger) (*Cache, error) {
 		_ = db.Close()
 		return nil, fmt.Errorf("enable WAL: %w", err)
 	}
-	// Busy timeout prevents concurrent writers from failing immediately with
-	// "database is locked"; 5 s is sufficient for the parallel check workers.
+	// busy_timeout remains as a defensive backstop (e.g. a lingering
+	// external reader/writer against the same file), not the mechanism this
+	// package relies on for its own internal write concurrency — that is
+	// now SetMaxOpenConns(1) above, since a single-connection pool never
+	// gives SQLite two callers to contend with in the first place.
 	if _, err := db.Exec("PRAGMA busy_timeout=5000"); err != nil {
 		_ = db.Close()
 		return nil, fmt.Errorf("set busy timeout: %w", err)
--- a/internal/koiosparity/cache.go
+++ b/internal/koiosparity/cache.go
@@ -1722,22 +1722,11 @@ func createCacheSchema(db *sql.DB) error {
 	}
 	// Older cache files may contain columns that the current structs no longer write.
 	for _, item := range [][2]string{{"koios_epoch_info", "pool_cnt"}, {"koios_epoch_info", "delegator_cnt"}, {"koios_totals", "deposits_d_rep"}} {
-		rows, err := db.Query(
-			"SELECT 1 FROM pragma_table_info(?) WHERE name = ?",
-			item[0],
-			item[1],
-		)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


chrisguiney:
Bot findings on `0b141e8`, dispositions at `81ba6ba`:

**CodeRabbit**

- `cache_concurrency_test.go:47` — stale present tense ("OpenCache never bounds the connection pool"). Valid, fixed in `81ba6ba`. Audited the same class: `TestOpenCacheLegacyColumnMigrationDoesNotDeadlock`'s doc comment described the now-fixed migration in the present tense as well, and is corrected in the same commit.
- `cache_concurrency_test.go:35` — gate the 5.5s hold behind `testing.Short()`. Declined. That test is this change's only fail-before evidence: reverting `cache.go` in place makes it fail in 5.09s with the issue's exact `database is locked (5) (SQLITE_BUSY)` line. Neither the Makefile nor any workflow passes `-short`, so the skip would fire only on ad-hoc local runs and would silently drop that evidence there, for 5.5s against a package whose race run is ~75s.
- `cache.go:1725` — log the probe error instead of folding it into "column absent". Declined. Behaviour is unchanged by the refactor: the loop already continued on both a `Query` error and a `rows.Err()` error, and the following `ALTER` discards its error too. `createCacheSchema` takes no logger, so this would mean reaching for `slog.Default()` inside a migration helper for a fault with no action attached.

**cubic** — no issues found across 2 files; review recorded at `0b141e8`.

`81ba6ba` is comments only. Local re-validation after it: `gofmt -s -l` clean, `golangci-lint run ./internal/koiosparity/...` 0 issues, `go test -count=1 ./internal/koiosparity/ -run 'TestSaveAccountFetchChunkProgress|TestOpenCacheLegacy'` passing.

chrisguiney:
@coderabbitai review

chrisguiney:
Bot coverage at `81ba6ba`, verified by the `commit_id` on each review record rather than by check colour:

- CodeRabbit: **rate-limited at this head** — "Review limit reached, next included review available in 36 minutes", and the explicit `@coderabbitai review` above returned "Review rate limited". Its only review record is stamped at `0b141e8`; the green CodeRabbit check on this head is not a review. The delta since `0b141e8` is comments only.
- cubic: review record stamped at `0b141e8` (no issues, 2 files). Its check run on this head reports an incremental review with 0 issues across 1 file, with no review record at that SHA.

No actionable bot findings outstanding. CI was still fully queued at the time of writing; no job had started.

wolf31o2 (approved):

Issue #3842, opened 2026-09-03
consensus: header validation checks the VRF key against the current registration, so a pool rotating its VRF key wedges the node


chrisguiney:
## Summary

Header validation checks a block's VRF key against the pool's **current**
registration instead of the one in the stake distribution that elected the
producer. When a pool rotates its VRF key, every block it makes for the rest of
that epoch is rejected, and the node wedges.

A from-genesis Preview replay hits this at slot 3362555 and stops: 78
rejections, tip frozen, 200,000+ blocks fetched and discarded.

## Observed

```
level=WARN msg="header validation rejected a block already on the primary chain;
  rewinding so chain selection can offer another candidate" component=ledger
  error="deferred block header verification failed at slot 3362555:
    producer pool 11670179c421b1419c493853ef542540c8f290cb1fc1f451709ca6d6
    VRF key does not match registered VRF key
    (header     b5971d1f29a9938574aa98cf5a8437a8245239644fe6dcaa8b22b92ed2b00e90,
     registered fa8c18c09e32aa415e9959afb033ef8f4276cd4202ca083f4316379bf5983d09)"
```

## The header is right and the check is wrong

The pool rotated its VRF key mid-epoch. From this node's own
`pool_registration` rows:

```
added_slot   vrf_key_hash
1014930      B5971D1F…   <- and eight further re-registrations keeping this key
2479516      B5971D1F…
3279920      FA8C18C0…   <- rotation, epoch 37
```

The failing block is at slot 3362555, epoch 38. Epoch 38's leader schedule is
built from the `mark` snapshot captured at the end of epoch 36 — slot 3196799 —
which is **before** the rotation at 3279920. At that capture the pool's
registered key was `B5971D1F…`, which is exactly what the block header carries.

So the producer used the key it was elected under. This node compared against
the key registered *now*.

The same lag is why the epoch alignment already exists elsewhere: a
`snapshot_epoch = E-1` mark snapshot is what the leader threshold uses (visible
in the threshold error's own fields). The VRF key has to come from the same
place as the stake it is validated alongside.

## Why it cannot currently do the right thing

`UtxoValidateBlockHeader`'s check reads live registration state
(`ledger/verify_header.go`, around the `VRF key does not match registered VRF
key` error):

```go
pool, err := ls.db.GetPool(poolKeyHash, true, nil)
...
registeredVrfKeyHash, ok := registeredPoolVrfKeyHash(pool)
```

[...]

PR #4134, opened 2026-09-08
fix(ledger): resolve pool VRF key from live registration below Mithril anchor

### LINKED ISSUE #4047: ledger: Mithril-bootstrapped node wedges at its first epoch boundary because pool VRF registration history below the anchor is missing
## Summary

After a Mithril bootstrap, the first epoch boundary the node crosses leaves it
unable to verify headers it has already accepted. Header verification resolves a
block producer's electing VRF key from pool registration history as of the
electing snapshot's parameter cutoff, and a bootstrapped database has no
registration rows at or below its trust anchor. The node then rejects a block
that is on its own primary chain, rewinds, and cannot re-follow the chain once
the required rollback passes k, so it wedges with a growing tip gap and produces
no blocks.

Seen on preview with a block producer, on a database bootstrapped at anchor
122078790 (immutable 28258) which is 4410 slots below the epoch 1413 boundary at
122083200.

## Observed

The boundary is crossed cleanly and rewards apply:

```
level=INFO msg="applied stake rewards" reward_snapshot_epoch=1410 performance_epoch=1411 pots_epoch=1412
level=INFO msg="handling epoch transition" previous_epoch=1412 new_epoch=1413 boundary_slot=122083200 snapshot_slot=122083199
level=INFO msg="precomputed stake rewards" performance_epoch=1412 pots_epoch=1413 application_epoch=1414 captured_slot=122083200
```

Immediately after, the first block of the new epoch is rejected:

```
level=WARN msg="header validation rejected a block already on the primary chain; rewinding so chain selection can offer another candidate"
  failing_block_slot=122083201 rewind_target_slot=122083195
  error="deferred block header verification failed at slot 122083201: block header verification rejected at slot 122083201:
  producer pool e0aeed8f4b362e684cb50c5895c1c653106c3cac8b70acd71bfdbd23 registration lookup failed:
  VRF key registration history unavailable at cutoff slot 121910399 or capture slot 121996799
  for pool e0aeed8f4b362e684cb50c5895c1c653106c3cac8b70acd71bfdbd23"
```

and the rewind cannot be undone:

```
level=WARN msg="rejecting rollback that exceeds security parameter K" fork_depth=461 security_param=432 rollback_slot=122083195
level=WARN msg="chain fork detected" fork_point_slot=122083195 fork_depth=29 alternate_head_slot=122083994 canonical_head_slot=122083195
```

fork_depth grows by one every cycle as the network advances, so the loop never
converges. The applied tip stayed at 122083195 while peers advanced past
122094700, and the node forged nothing. CPU was ~1% and there was no other
error, so it looks idle rather than broken.

cutoff slot 121910399 is the epoch 1411 boundary minus one and capture slot
121996799 is the epoch 1412 boundary minus one. Both are far below the anchor.

## Why the database cannot answer

`electingVrfKeyHashWithCache` in `ledger/verify_header.go` resolves the electing
key like this:

- `electingPoolParamsCutoffSlotWithCache` reports the cutoff and capture slots
  and whether an electing stake snapshot exists (`ok`).
- when `ok`, the key comes from `GetPoolVrfKeyHashAtSlot(cutoffSlot)`, falling
  back to `GetPoolEarliestVrfKeyHashAtSlot(capturedSlot)`.
- when neither finds a row, the function returns
  `errVrfKeyRegistrationHistoryUnavailable`.

The live registration is used only when `ok` is false. That is deliberate, per
the comment on the function:

```
// Falling back to the live registration is safe only when no snapshot is
// available. A present snapshot with no historical registration means the
// database cannot answer the consensus-critical question; using the live key
// would reintroduce the rotation wedge this lookup avoids.
```

A Mithril import supplies the stake snapshots but not the pool registration
certificates that produced them, so after the boundary `ok` is true while both
history lookups miss. Before the boundary the node has no electing snapshot for
those epochs, `ok` is false, the live registration answers, and block production
works, which is why the failure appears only once the node computes a snapshot
of its own.

The deferral does not help either. In `verifyRegisteredVrfKeyWithCache` the
caller defers on this sentinel only while `ls.ledgerTipBehindSlot(...)`, so once
the apply cursor reaches the slot the same missing history becomes a permanent
rejection rather than a deferral.

## Impact

This is the same class as #3767: the import leaves out history that a later
consensus path requires. With #3767 fixed, a bootstrapped producer now survives
until its first epoch boundary and wedges there instead of halting on the first
withdrawal. On preview that caps a bootstrap at one epoch of block production
and requires a fresh bootstrap every epoch.

## Environment

- preview, epoch 1412 to 1413 boundary at slot 122083200
- dingo at 7ad279103fb621d468a82e339f15288e126736db (includes #3871)
- host aarch64, docker, `storageMode: core`, sqlite metadata, badger blob
- bootstrapped from immutable 28258, anchor 122078790

## Suggested direction

- Seed pool registration history for the imported snapshots at bootstrap, the
  way #3871 seeds the snapshots' block counts, so the cutoff and capture lookups
  can be answered for epochs at or below the anchor.
- Or record the electing VRF key alongside the stake it was captured with, which
  is what cardano-ledger does via `IndividualPoolStake.individualPoolStakeVrf`.
  The comment already notes `pool_stake_snapshot` has no such column and that it
  freezes the Leios voting key at the same boundary for the same reason.
- Failing either, treat "snapshot present but its registration history is at or
  below the trust anchor" as a distinct known gap rather than an authoritative
  rejection, so it cannot pin the chain below k.

arepala-uml:
## Summary

- `electingVrfKeyHashWithCache` (`ledger/verify_header.go`) raised
  `errVrfKeyRegistrationHistoryUnavailable` whenever both
  `GetPoolVrfKeyHashAtSlot` and `GetPoolEarliestVrfKeyHashAtSlot` missed for a
  pool. A Mithril snapshot import writes only a pool's live registration,
  stamped at the import slot, never the certificate history that produced it,
  so any stake-snapshot capture or parameter cutoff slot before that import
  slot has no answerable row. The very first post-bootstrap epoch boundary's
  cutoff and capture are always before the anchor, so this fires
  deterministically at every epoch boundary after a Mithril bootstrap. The
  error is unconditional once the ledger tip has caught up to the checked
  slot (the deferred-retry path only covers a tip still catching up), so the
  node rewinds, cannot re-follow past security parameter k, and wedges.
- Fix: when both lookups miss and the stake-snapshot capture slot is at or
  below the persisted Mithril boundary (`mithrilLedgerSlot`), fall back to
  the pool's live registration. That combination is diagnostic of the
  bootstrap gap specifically, not a general missing-registration case, so it
  does not reintroduce the VRF-rotation wedge #3842 removed: a pool with no
  registration at all still hard-rejects, and a node with no Mithril boundary
  (`mithrilLedgerSlot == 0`, e.g. genesis-synced) still hard-rejects an
  equivalent-looking gap.
- An earlier version of this fix instead seeded a second, `added_slot = 0`
  "floor" registration row at Mithril import time (`ImportPool`). That also
  satisfied `GetPoolRegistrationsEffectiveForEpoch`'s in-epoch registration
  window whenever a processed epoch's start slot was small, letting the floor
  row outrank a pool's real, owner-populated registration under that query's
  ascending "earliest first-in-epoch" ordering and zero out the pool's owner
  stake in reward-input seeding. Caught by
  `TestHandleEpochTransitionCapturesSelfDelegatedOwnerStake` and two
  neighboring `ledger/snapshot` tests failing under `-race`. Resolving the
  gap only at the one call site that has the consensus-critical question,
  using the boundary the node already persists for exactly this purpose,
  avoids that collateral damage entirely.

Fixes #4047

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./ledger/... ./database/... ./ledgerstate/... -race` — all
      packages pass, including `ledger/snapshot` (which failed under the
      abandoned floor-row approach)
- [x] `golangci-lint run ./ledger/... ./database/...` — 0 issues
- [x] `modernize ./...` — no new findings in changed files (repo-wide
      pre-existing findings unrelated to this change untouched)
- [x] `make docs-parity`
- [x] New regression tests added to
      `ledger/verify_header_vrf_params_lag_test.go`:
      `TestElectingVrfKeyHashResolvesBelowMithrilBootstrapAnchor` (the #4047
      repro; verified it fails with the fix reverted, reproducing the exact
      reported error message), `TestElectingVrfKeyHashStillRejectsAPoolWithNoRegistrationAtAll`,
      `TestElectingVrfKeyHashDoesNotFallBackWithoutAMithrilBoundary` (the
      #3842 boundary the fix must not erase)
- [x] `DATABASE.md` updated (the `GetPoolVrfKeyHashAtSlot`/`GetPoolEarliestVrfKeyHashAtSlot`
      section); `ARCHITECTURE.md` not affected (no component/plugin/lifecycle
      change)
- [ ] Not verified: no live Preview node available in this environment, so
      this was not reproduced/confirmed fixed against live Preview infra;
      DevNet/conformance suites were not run (this change is header-only VRF
      key resolution logic, not consensus rule/wire-format changes, so `make
      lint`'s `nilaway`/full-module run plus the targeted `-race` suite above
      were used as the available substitute)

---
## Summary by cubic
Fixes a wedge where a Mithril-bootstrapped node gets stuck at its first epoch boundary after import. A Mithril snapshot import records only a pool's live registration stamped at the import slot, so when the electing snapshot's cutoff and capture slots precede that anchor — which the first post-bootstrap boundary always does — VRF key resolution found no row and raised `errVrfKeyRegistrationHistoryUnavailable`, rewinding the node past security parameter k.

**Fix**
- Falls back to the pool's live registration when both lookups miss and the capture slot is at or below the persisted Mithril boundary (`mithrilLedgerSlot`).
- Keeps the #3842 guarantee: pools with no registration at all still hard-reject, and nodes without a Mithril boundary do not fall back for an equivalent-looking gap.
- Avoids seeding a synthetic registration row at import time, which outranked real registrations in reward-input seeding.
- Adds three regression tests covering the reported gap, the no-registration case, and the non-bootstrapped case, plus documents the lookup behavior in `DATABASE.md`.

<sup>Written for commit 465b01908832572b75e9724b99412f5cb9766344. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4134?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed epoch-boundary processing on Mithril-bootstrapped nodes when VRF registration history is unavailable.
  - The system now uses a pool’s current registered VRF key for eligible bootstrap-era lookups.
  - Genuine registration gaps and pools without a registered VRF key continue to return an error.

- **Documentation**
  - Documented the bootstrap-era VRF lookup behavior and the rejected alternative approach.

### DIFF
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -2809,6 +2809,41 @@ the *earliest* lookup at or before the capture slot. Earliest rather than
 latest, because a pool that also re-registered within that same epoch had the
 re-registration deferred, so the snapshot still carries the first one.
 
+A Mithril snapshot import writes a pool's registration (`ImportPool`) stamped
+at the import slot, because the snapshot carries only the pool's current
+parameters, not the certificate history that produced them. Both lookups
+above filter `added_slot <= slot`, so for any cutoff or capture slot before
+that import slot -- which the very first post-bootstrap epoch boundary's
+cutoff and capture necessarily are, since that boundary's electing snapshot
+was captured no later than the anchor the node just bootstrapped from --
+neither query finds a row for a pool that has in fact been continuously
+registered the whole time, and `electingVrfKeyHashWithCache`
+(`ledger/verify_header.go`) raised `errVrfKeyRegistrationHistoryUnavailable`
+unconditionally (issue #4047; the deferred-retry path in
+`verifyBlockHeaderStateWithCache` only covers a ledger tip still catching up,
+not this permanent gap).
+
+The fix lives in `electingVrfKeyHashWithCache`, not in these two queries or in
+`ImportPool`: when both lookups miss, and this is a Mithril-bootstrapped node
+(`ls.mithrilLedgerSlot != 0`) whose stake-snapshot capture slot is at or below
+that boundary, the gap is known to be exactly this bootstrap shape rather than
+a genuine missing-registration bug, so the caller falls back to the pool's
+current live registration (`GetPool(includeInactive=true)`) the same way it
+already does when `electingPoolParamsCutoffSlotWithCache` reports no snapshot
+at all (`ok == false`). An earlier version of this fix instead seeded a second,
+`added_slot = 0` "floor" registration row at import time so the existing
+queries would resolve on their own; that synthetic row also satisfied
+`GetPoolRegistrationsEffectiveForEpoch`'s in-epoch registration window
+whenever a processed epoch's start slot was small, letting it outrank the
+pool's real, owner-populated registration under that query's ascending
+"earliest first-in-epoch" ordering and zeroing out the pool's owner stake in
+reward-input seeding -- caught by
+`TestHandleEpochTransitionCapturesSelfDelegatedOwnerStake` and its
+neighboring `ledger/snapshot` tests. Resolving the gap only where the
+consensus-critical question is actually asked, using the boundary the node
+already persists for exactly this purpose, avoids adding a row that other
+`added_slot`-scoped readers have to reason about.
+
 ### `GetPoolsRetiringAtEpoch`
 
 Pools whose effective retirement takes effect at a given epoch, with the reward
--- a/ledger/verify_header.go
+++ b/ledger/verify_header.go
@@ -1618,6 +1618,37 @@ func (ls *LedgerState) electingVrfKeyHashWithCache(
 			copy(hash[:], vrfKeyHash)
 			return hash, true, nil
 		}
+		// Both lookups missed. That is ordinarily the unanswerable gap the
+		// error below reports, but it is also exactly what a Mithril
+		// bootstrap produces for the epoch boundary the node crosses right
+		// after import: the snapshot import writes only the pool's live
+		// registration, stamped at the import slot, and never replays the
+		// certificate history that produced it. So a pool that has in fact
+		// been continuously registered the whole time has no row at or
+		// before a cutoff/capture slot that predates the import (issue
+		// #4047).
+		//
+		// mithrilLedgerSlot pins that import slot for the life of the
+		// process (set once at startup, like the other reads of this field
+		// in this file), so a capture at or below it is diagnostic: the
+		// snapshot that elected this pool was captured no later than the
+		// bootstrap boundary, which only a bootstrap-created gap explains.
+		// Falling back to the live registration here is the same trust
+		// electingVrfKeyHashWithCache already extends when ok is false --
+		// no snapshot at all -- applied to the narrower case of a snapshot
+		// whose registration history the import could not carry. It is not
+		// the general "no history found" fallback #3842 removed: outside
+		// this bootstrap-anchor window, a genuine gap still hard-rejects
+		// rather than resolving the pool's current (possibly rotated) key.
+		if ls.mithrilLedgerSlot != 0 && capturedSlot <= ls.mithrilLedgerSlot {
+			pool, poolErr := ls.db.GetPool(poolKeyHash, true, nil)
+			if poolErr != nil && !errors.Is(poolErr, models.ErrPoolNotFound) {
+				return lcommon.Blake2b256{}, false, poolErr
+			}
+			if hash, ok := registeredPoolVrfKeyHash(pool); ok {
+				return hash, true, nil
+			}
+		}
 		return lcommon.Blake2b256{}, false, fmt.Errorf(
 			"%w at cutoff slot %d or capture slot %d for pool %x",
 			errVrfKeyRegistrationHistoryUnavailable,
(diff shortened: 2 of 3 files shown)


wolf31o2 (approved):
Accept

Issue #4132, opened 2026-09-08
ouroboros: chainsync re-sync sends FindIntersect with RequestNext replies in flight, killing the connection


chrisguiney:
A chainsync re-sync issues `MsgFindIntersect` on a connection that still has
outstanding `MsgRequestNext` responses. The legitimate replies then arrive while
the protocol is in `Intersect`, gouroboros correctly rejects them as
out-of-state, and the connection is torn down.

Observed three times in 50 seconds against one peer during fork churn on a
from-genesis Preview replay:

```
17:35:08  message *chainsync.MsgRollBackward   not allowed in current protocol state Intersect
17:35:29  message *chainsync.MsgRollForwardNtN not allowed in current protocol state Intersect
17:35:57  message *chainsync.MsgRollForwardNtN not allowed in current protocol state Intersect
```

all logged as `unexpected connection failure: protocol error: chain-sync: ...`
from `component=peergov`, `connection_id=...<->24.192.179.116:5000`.

## Why these are legitimate messages

`MsgRollForward` and `MsgRollBackward` are the two replies to `MsgRequestNext`.
Receiving either is only invalid because the client moved the protocol into
`Intersect` while they were still in flight. The peer is behaving correctly;
gouroboros is enforcing the state machine correctly
(`protocol/protocol.go:1109`, v0.204.0). The defect is that the client changes
state without first draining the outstanding request.

That both message types appear, and only ever in `Intersect`, is what
distinguishes this from a misbehaving peer: a peer sending garbage would not
send exactly the two valid `RequestNext` replies and nothing else.

## Path

`ouroboros/chainsync.go:479` is the only call site:

```go
return conn.ChainSync().Client.Sync(intersectPoints)
```

`Sync` sends `MsgFindIntersect` on the existing connection. It is reached from
the re-sync triggers in `ledger/chainsync.go`, which fire on exactly the
conditions present when this happened:

- "received rollback point ahead of local tip, triggering chainsync re-sync"
- "rollback point not found locally, triggering chainsync re-sync"
- "persistent chain fork detected, triggering chainsync re-sync"

The last of these fired repeatedly in this run (`consecutive_mismatches=20`),
which is the same fork churn that produces the cross-fork splices in #4102 — so
this compounds instability during exactly the window where the node can least
afford to lose peers.

## Suggested direction

Either drain or cancel in-flight `RequestNext` responses before sending
`MsgFindIntersect`, or perform the re-sync on a fresh connection the way the
[...]

Issue #3987, opened 2026-09-06
chainselection: after a connection recycle the replacement peer stays unselectable until the next RollForward, stalling chain selection for a full block interval


chadaibbhot-git:
## Problem

On a Leios devnet block producer whose only configured upstream is a single
trusted relay, every leios-fetch connection recycle removes the peer from
chain selection and the *replacement* connection cannot be selected again
until the network produces its next block. Chain selection is stalled for that
whole interval (measured 13.0 s, 17.8 s, 25.8 s and 47.6 s in one 90-minute
window), and the producer forges against a header frontier that is frozen for
the duration.

The gap is an asymmetry between the roll-forward and roll-backward paths:
roll-forward *creates* a `peerTips` entry for an unknown connection,
roll-backward only *updates* one that already exists. After a recycle the entry
has been deleted, so the tip carried by the post-`FindIntersect`
`MsgRollBackward` is discarded and the peer stays invisible to chain selection
until the first `MsgRollForward` arrives.

## Evidence

Measured over 502 blocks (~13 h of upstream log, ~4 h of producer log) on a
Leios devnet with ~1 s slots. Wall-clock reference for a slot is
`genesis + (slot - base)`, validated exactly against an independent chain
index on 50/50 canonical blocks. Both hosts are NTP-synced (offsets < 1 ms).

Per-hop latency, block production slot -> upstream relay adoption -> producer apply:

| hop | p50 | p90 | p99 | max |
|---|---|---|---|---|
| D1  slot wall clock -> upstream relay adopts | 1.38 s | 10.05 s | 18.01 s | 24.58 s |
| D23 upstream adopts -> producer chain-extends | 13.71 s | 41.51 s | 97.35 s | 458.10 s |
| Dtot slot wall clock -> producer chain-extends | 21.55 s | 45.06 s | 98.55 s | 460.54 s |

The producer-side hop is 91% of the p50 total. Decomposing it further shows the
delay is **not** on the wire:

| sub-hop | p50 | p90 |
|---|---|---|
| upstream adopts RB -> producer has that block's Leios EB txs | 0.08 s | 1.61 s |
| producer has EB txs -> producer chain-extends | 25.84 s | 47.17 s |

In 140/295 cases (47%) the producer held the Leios EB data *before* its upstream
had even adopted the announcing ranking block. Blockfetch
`RequestRange`-to-first-block EMA against that upstream was 0.59 s. The network
path is fast; the delay is inside the node.

[...]

Issue #3808, opened 2026-09-03
chainsync: the plateau watchdog waits 20 minutes to clear a wedged stream because catchUpMultiplier is gated on IsAtTip rather than on progress


chrisguiney:
## Summary

The local-tip plateau watchdog is the only thing that clears a wedged
chainsync/blockfetch stream during a from-genesis replay, and it waits 20
minutes to do it. On a Preview replay that cost 62 of 128 minutes — the node
spent 87% of its most recent phase stalled, at 0.37 blocks/second wall-clock
against 2.86 while actually moving.

The multiplier that produces the 20 minutes is gated on the wrong condition.

## The threshold

`internal/chainsyncrecycler/recycler.go`:

```go
func plateauThreshold(stallTimeout time.Duration) time.Duration {
	return max(2*stallTimeout, 4*time.Minute)
}

const catchUpMultiplier = 5

	multiplier := 1
	if !live.Ledger.IsAtTip() {
		multiplier = catchUpMultiplier
	}
	effectivePlateau := time.Duration(multiplier) * r.plateauRecovery
```

With the default `StallTimeout: "2m"` (`internal/config/config.go`) that is
4m × 5 = **20m**, which matches the observed `plateau_duration` exactly:
`20m0.000166333s`, `20m29.99s`, `20m30.00s` across three incidents.

The comment above the multiplier justifies it as avoiding churn "while the node
is making progress". But the gate is `!IsAtTip()`, which is true for the entire
replay, not whether the node is progressing — and `shouldRecycleLocalTipPlateau`
only fires when there is *no* progress. So the multiplier is applied in exactly
the situation its own rationale excludes. The function already tracks real
progress a few lines above and does not use it here:

```go
	if localTipSlot > st.lastProgressSlot {
		st.lastProgressSlot = localTipSlot
		st.lastProgressAt = now
	}
```

## What the 20 minutes looks like

A replay-recovery rewind precedes each incident:

```
17:46:08 WARN detected inconsistent local ledger state during replay, rewinding
  recovery_strategy=security-param-fallback failing_block_slot=4207759
  missing_input=929b418d…#1 producer_block_slot=4198879 rollback_slot=4198877
```

For the next 20 minutes the applied tip is frozen at 4198877 while:

- `fork_path_headers` grows monotonically 2481 → 5380, about +1 every 4s, never
  resolving;
- 5,000–8,400 blockfetch blocks per minute are fetched, fully decoded, and
[...]

Issue #2259, opened 2026-05-13
ouroboros/chainsync: propagate async server RollForward/RollBackward failures


wolf31o2:
## Summary

Async ChainSync server send failures are logged but not propagated to connection lifecycle handling.

## Audit Details

Finding ID: H6, HIGH error handling finding.

Affected file: `ouroboros/chainsync.go`.

In `chainsyncServerRequestNext`, the immediate `RollForward` / `RollBackward` path returns errors to the caller. The async path after `AwaitReply` starts a goroutine, waits on the chain iterator, then calls `ctx.Server.RollBackward` or `ctx.Server.RollForward`. If either call fails, the code logs an error and returns from the goroutine.

Impact:

- A failed `RollForward` can silently drop a block for a downstream peer.
- A failed `RollBackward` can leave the peer's chain state inconsistent.
- No connection recycle, close, or error event is guaranteed.

## What Needs to Happen

1. Route async ChainSync server send failures into connection error handling.
2. Prefer closing or recycling the connection when the server cannot send a rollback or block.
3. Add tests for async `RollForward` and `RollBackward` failures.
4. Ensure the fix does not leak iterator goroutines.
5. Keep expected context-cancel shutdown quiet.

## Acceptance Criteria

- Async send errors are observable by connection lifecycle code.
- Tests prove failed async sends do not leave the connection silently open.

PR #4138, opened 2026-09-08
fix(chainsync): close the transport when the post-AwaitReply server waiter gives up

### LINKED ISSUE #4137: chainsync: the post-AwaitReply server waiter has three exits that abandon the follower without closing the transport, stranding it in MustReply for the full 135-269 s timeout and repeating on every reconnect
## Summary

`chainsyncServerRequestNext` sends `MsgAwaitReply` and then hands the follower
to a goroutine that waits on the client's chain iterator
(`ouroboros/chainsync.go:864-946`). Three of that goroutine's exits return
without sending `MsgRollForward`/`MsgRollBackward` **and** without closing the
transport:

```go
		select {
		case <-done:
			// Iterator returned
		case <-conn.ErrorChan():          // ouroboros/chainsync.go:886
			clientState.ChainIter.Cancel()
			return                        // (1) no close
		}
		if nextErr != nil {               // :890
			if !errors.Is(nextErr, context.Canceled) {
				o.config.Logger.Debug(
					"failed to get next block from chain iterator",
					"error", nextErr,
				)
			}
			return                        // (2) no close, DEBUG only
		}
		if next == nil {                  // :901
			o.config.Logger.Debug(
				"chainsync server: goroutine got nil block", ...)
			return                        // (3) no close, DEBUG only
		}
```

Once `AwaitReply()` has been sent the server holds agency in `MustReply`. A
client parked there has no way to make progress and no way to detect the
abandonment except gouroboros' randomized `MustReply` timer — 135-269 s
(`protocol/chainsync/chainsync.go:210-212`, armed from `StateMapNtN`'s
`stateMustReply.TimeoutFunc`). The client then tears the connection down with
`chain-sync: timeout waiting on transition from protocol state MustReply`,
reconnects, negotiates a fresh intersection, receives exactly one
`MsgRollBackward` — and lands in the same trap. The loop is stable: a node can
sit ~1000 slots behind a healthy upstream indefinitely, with no ERROR on either
side, until an operator restarts one of the two processes.

The code in the same function already documents the correct behaviour, in the
one branch that does it right (`:923-931`):

> Do not RollForward an incomplete CertRB. This runs after the callback returned
> AwaitReply, so **actively close the transport (an error-channel send alone does
> not) to unpark the client from AwaitReply** so it reconnects and retries...

That reasoning applies verbatim to exits (1), (2) and (3). This is the residual
gap in the fix for #2259, which routed the async *send* failures into connection
lifecycle handling but left the three pre-send exits untouched.

Exit (1) has a second failure mode. `conn.ErrorChan()` is a single buffered
`chan error` (gouroboros `connection.go:71`, `make(chan error, 10)` at
`connection.go:138`), not a broadcast, and it has several competing consumers:
this waiter (`ouroboros/chainsync.go:886`), blockfetch
(`ouroboros/blockfetch.go:314,430`), tx-submission
(`ouroboros/txsubmission.go:408,432,512,551,635,666`) and the connection
manager's own teardown reader (`connmanager/connection_manager.go:847`). A live
protocol error belonging to some *other* mini-protocol on the same full-duplex
connection can therefore be consumed here. When it is, this waiter cancels the
follower's iterator and returns, and the connection manager never receives that
error, so the connection is not torn down at all — the follower is stranded on a
connection both ends still consider healthy.

## Evidence

A Leios devnet with 1 s slots. A block producer whose only local-root upstream
is one relay running the same build. The relay was restarted at 12:43:09Z and
was fully healthy from 12:44Z onward — it extended its own chain 40+ times over
the next 12 minutes, ending at slot 2811200. The producer's applied tip stayed
pinned at slot 2810012 for 20 minutes and 1282 slots, until it too was
restarted at 12:55Z.

Producer side, one cycle per chainsync session, eight sessions in a row:

```
12:47:51.617  connected to the relay
12:47:51.658  registered peer from chainsync rollback  rollback_slot=2810012
                                                       advertised_slot=2810823
              <<< nothing further on this session >>>
12:51:09.702  unexpected connection failure: protocol error:
              chain-sync: timeout waiting on transition from protocol state MustReply
12:51:09.709  connected to the relay  (cycle repeats)
```

Inside that 198-second window the relay extended its chain five times — slots
2810883, 2810904, 2810930, 2810999, 2811014 — and sent the producer none of
them.

Session lifetimes, measured producer-side from intersect to `MustReply`
teardown:

| session opened | MustReply teardown | lifetime |
|---|---|---|
| 12:47:51.617 | 12:51:09.702 | 198.1 s |
| 12:51:09.709 | 12:54:11.984 | 182.3 s |
| 12:51:51.616 | 12:55:00.203 | 188.6 s |
| 12:35:57.872 | 12:38:43.539 | 165.7 s |
| 12:38:43.541 | 12:41:29.622 | 166.1 s |

Five consecutive lifetimes, every one inside `[MustReplyTimeoutMin,
MustReplyTimeoutMax]` = `[135 s, 269 s]`, uniformly scattered across it and
matching no other timer in either process. That is the randomized `MustReply`
timer firing, which is only reachable if the server sent `MsgAwaitReply` and
then said nothing at all. Each session delivered exactly one message — the
post-`FindIntersect` `MsgRollBackward` — which is also visible in chain
selection: every registration recorded `observed_block_number=0`, i.e. zero
headers delivered, for the whole 20 minutes.

The intersection was valid and on the relay's canonical chain: both nodes agreed
on `9615e57a14690fec7966d890d89d2f77ae3449641cd63d4dff11704fdd85e78a` at slot
2810012, and the relay extended forward from it without ever rolling back below
it. So this is not "the peer cannot serve that point"; the relay could serve it
and did not.

Neither side logged an ERROR for any of this. The producer logged only the
`MustReply` timeouts; the relay logged only `peer closed the connection while
reading header: EOF` when the producer gave up.

## Why the node cannot recover on its own

The producer's plateau watchdog fires and does the right thing — closes the
connection for a fresh chainsync — but that is precisely the action that
re-enters the trap, so it runs forever:

```
12:47:51.611 WARN local tip plateau detected, resyncing chainsync client
             local_tip_slot=2810012 best_peer_tip_slot=2810823
             plateau_duration=4m29.999021739s eligible_peer_count=1
12:47:51.612 INFO closing connection for fresh chainsync reason=local_tip_plateau
12:47:51.658 INFO registered peer from chainsync rollback rollback_slot=2810012
```

Three plateau cycles and five `MustReply` cycles in 20 minutes, all identical.
`chain selection stalled: no selectable peer` fired nine times in the same
window. Nothing escalates: there is no "the resync did not help" path, and no
metric or log distinguishes "my upstream has nothing to send" from "my upstream
has stopped speaking to me".

## Secondary: the plateau watchdog can be blinded right after it fires

`checkLocalTipPlateau` compares the local tip against the best peer's
**delivered** frontier (`internal/chainsyncrecycler/recycler.go:471-473`):

```go
	bestPeerTipSlot := bestPeerTip.SelectionTip().Point.Slot
	if bestPeerTipSlot <= localTipSlot {
		return
	}
```

Immediately after a plateau resync, every peer has been reconnected and has
delivered nothing but its post-intersect rollback, so its delivered frontier is
the intersection point — which is the local tip. `bestPeerTipSlot <=
localTipSlot` holds and the watchdog cannot re-arm until some peer delivers a
header, which is exactly what is not happening. The peer's *advertised* tip is
known and correct throughout (2810823 above), and is the value the watchdog
actually wants here.

On a second node in the same fleet the identical starvation ran from 12:33:32Z
to a restart at 12:43:09Z with the plateau watchdog firing exactly once, at
12:37:42Z, and never again despite five more minutes of a frozen tip.

Note for reviewers: the delivered-frontier value in these logs comes from the
rollback-registration path in open PR #3989 (`newPeerChainTipFromRollback`
records the confirmed intersection point with block number 0). Without that PR
the peer would simply be absent from `peerTips` and `checkLocalTipPlateau` would
return early on `bestPeer == nil` instead — so the watchdog is blind either way;
#3989 only changes the shape of the blindness. The fix belongs in the watchdog:
use `bestPeerTip.Tip` (advertised) when the peer has not delivered a header yet.

## Suggested fix

1. In the post-`AwaitReply` waiter, route exits (1), (2) and (3) through
   `closeChainsyncServerConn` (`ouroboros/chainsync.go:989`), the same helper the
   `CertRB` branch already uses. Once `MsgAwaitReply` is on the wire the server
   must either send a reply or drop the transport; returning silently is not a
   third option. Raise the iterator-error log from `Debug` to `Warn` while there.
2. Stop treating `conn.ErrorChan()` as a broadcast. Either give the waiter a
   dedicated per-protocol cancellation signal, or have it re-publish any error it
   consumes so the connection manager's reader
   (`connmanager/connection_manager.go:847`) still sees it. Today whichever
   goroutine wins the race decides whether the connection is torn down.
3. In `checkLocalTipPlateau`, fall back to the peer's advertised tip when its
   delivered frontier has not moved past the local tip since the last resync, so
   the watchdog is not disarmed by its own recovery action.
4. Consider escalating after N consecutive plateau resyncs that produce no
   applied block — the current loop is silent and unbounded.

## Environment

- `main` at `9288bd78` plus open PRs #3972, #3988, #3993, #3963, #3986, #3989,
  #3967, #4022, #4020 and a local DRep-deposit patch. Go 1.26.7, gouroboros
  v0.202.9.
- Leios devnet, 1 s slots, SQLite metadata store.
- `ouroboros/chainsync.go:864-946` and `connmanager/connection_manager.go:847`
  are identical to `origin/main` in this tree; none of the PRs above touch them.
  #3989 touches `chainselection` only and is discussed in its own section.

## Related

- #2259 — the async-send half of this was fixed there; the pre-send exits were
  not.
- #769 — "Chainsync server doesn't send rollbacks after initial one" (closed);
  same observable shape from the client's side.
- #3808 — the plateau watchdog's threshold. This report is about the watchdog
  being unable to re-arm at all, which is orthogonal to how long it waits.
- #3987 — post-recycle peer selectability; adjacent but about selection, not
  about the server never speaking.
- #4132 — chainsync protocol-state violations during re-sync; different failure,
  same re-sync path.

chadaibbhot-git:
Fixes #4137

## The bug

`chainsyncServerRequestNext` sends `MsgAwaitReply` and hands the follower to a
goroutine that waits on the client's chain iterator. Three of that goroutine's
exits returned without sending `MsgRollForward`/`MsgRollBackward` **and** without
closing the transport:

1. the `conn.ErrorChan()` case of the `select`,
2. a non-cancellation chain-iterator error (logged at `Debug`),
3. a nil iterator result (logged at `Debug`).

Once `MsgAwaitReply` is on the wire the server holds agency in `MustReply`. A
peer parked there cannot make progress and cannot detect the abandonment except
through gouroboros' randomized `MustReply` timer (135-269 s). It then tears the
connection down with `chain-sync: timeout waiting on transition from protocol
state MustReply`, reconnects, negotiates a fresh intersection, receives exactly
one `MsgRollBackward` — and lands in the same trap. The loop is stable and
silent: a node sits ~1000 slots behind a healthy upstream indefinitely with no
ERROR on either side, and its own plateau watchdog re-enters the trap every time
it fires. Five consecutive session lifetimes measured on a devnet were 198.1 s,
182.3 s, 188.6 s, 165.7 s and 166.1 s — every one inside
`[MustReplyTimeoutMin, MustReplyTimeoutMax]` and matching no other timer in
either process. Full evidence in #4137.

The `CertRB` branch of the same function already documents the correct behaviour
and calls `closeChainsyncServerConn`:

> Do not RollForward an incomplete CertRB. This runs after the callback returned
> AwaitReply, so actively close the transport (an error-channel send alone does
> not) to unpark the client from AwaitReply so it reconnects and retries...

That reasoning applies verbatim to the three earlier exits. This is the residual
gap in #2259, which routed the async *send* failures into connection lifecycle
handling but left the pre-send exits untouched.

## The change

Route all three exits through `closeChainsyncServerConn`, each with a reason
naming the exit, so the teardown is a warning rather than a debug line.

`context.Canceled` deliberately stays quiet and does **not** close. A server
client's iterator is only cancelled from `chainsync.State.RemoveClient` — which
connection-closed handling drives — and from the error-channel exit above; in
both cases the transport is already gone, so there is no parked peer to release
and closing would turn every ordinary disconnect into a connection-recycling
warning. `TestChainsyncServerServeAwaitedCancelledDoesNotCloseConnection` pins
that, alongside the pre-existing
`TestChainsyncServerRequestNextIteratorCancelDoesNotCloseConnection`.

On the shared error channel: `conn.ErrorChan()` is a single buffered `chan error`
with several consumers (this waiter, blockfetch, tx-submission, and the
connection manager's own teardown watcher), so the error this waiter consumes may
belong to another mini-protocol — and the manager then never sees it and does not
tear the connection down at all. Closing is correct either way: it is what the
manager would have done with that error, and it wakes the manager's watcher
through the closed error channel. **Reshaping that channel is deliberately out of
scope here**; this PR only stops the exits from leaking a live follower.

The post-wait handling moves into `chainsyncServerServeAwaited` so exit (3) is
reachable from a test: `ChainIterator.Next` maps an empty non-error result to
`ErrIteratorChainTip`, so the real iterator cannot produce one.

The channel the waiter watches is read through `chainsyncServerConnErrorChan`,
which returns `conn.ErrorChan()` unless a test supplies a single-consumer
stand-in. A test cannot drive exit (1) on the real channel: delivery goes to
whichever consumer the runtime picks, and publishing an extra error to cover the
other consumers races the channel's own closure — the consumer that wins closes
the connection, and gouroboros' `Connection.shutdown` closes the error channel it
owns, so the extra send can panic with `send on closed channel`. The default is
asserted separately so the seam cannot mask a waiter wired to the wrong channel.

## Tests

New, in `ouroboros/chainsync_server_fixture_test.go`, on the existing
ouroboros-mock server fixture. Each drives a real follower into `AwaitReply` over
a real protocol connection and asserts the client was *released* — the harness's
observed-message stream closes — rather than left blocking to the deadline:

| test | exit |
|---|---|
| `TestChainsyncServerRequestNextIteratorErrorAfterAwaitReplyUnparksClient` | (2) real iterator failure, driven by closing the backing store and waking the parked iterator |
| `TestChainsyncServerRequestNextNilBlockAfterAwaitReplyUnparksClient` | (3) nil result |
| `TestChainsyncServerRequestNextConnErrorAfterAwaitReplyUnparksClient` | (1) error-channel |
| `TestChainsyncServerConnErrorChanDefaultsToTheConnection` | the seam's production default |
| `TestChainsyncServerServeAwaitedCancelledDoesNotCloseConnection` | the exit that must stay silent |

Each was verified red by reverting only the branch it covers: the three
abandonment tests fail with `context deadline exceeded` instead of the expected
close, which is the production symptom exactly. The cancel test fails if the
`context.Canceled` branch is removed, and so does the pre-existing
`TestChainsyncServerRequestNextIteratorCancelDoesNotCloseConnection`. The seam
test fails if `chainsyncServerConnErrorChan` stops returning `conn.ErrorChan()`.

Note that on `main` neither consumer of a connection error closes the transport —
`ConnectionManager.RemoveConnection` only unregisters — so the peer stayed parked
whichever consumer won.

Full package green, including `-race`, on the current `main` base:

```
go test ./ouroboros/ -run TestChainsyncServer -count=5        ok  10.9s
go test -race ./ouroboros/ ./chainsync/ ./connmanager/        ok  44.2s / 1.5s / 4.9s
go build ./... && go vet ./ouroboros/                         clean
```

## Related

- #4137 — the report this fixes.
- #2259 — fixed the async *send* half; the pre-send exits were not touched.
- #2250, #2150, #2177 (closed) — the same observable from the client's side
  (`MustReply` timeout / local tip plateau / tip pinned at S on a single-peer
  BP), none of which reached this server-side exit.
- #4132 — chainsync protocol-state violations during re-sync: different failure,
  same re-sync path.
- #3987 — post-recycle peer selectability; adjacent but about selection, not
  about the server never speaking.
- #3808 / #3886 — the plateau watchdog's threshold and its catch-up behaviour.
  The watchdog's inability to re-arm at all is a separate PR against
  `internal/chainsyncrecycler`.

## CI status

Rebased onto `main` at `aa088149` (2026-09-09), which now includes #4144's fix
for the `chainselection`/`ledger` frontier-flap flakes noted below previously.
All 4 commits on this branch carried through the rebase byte-identical
(`git range-diff`, no conflicts).

Re-verified on the rebased head: `go build ./...`, `go vet ./ouroboros/...`,
and `go test ./ouroboros/...` all clean/green (this PR only touches
`ouroboros/chainsync.go` and its test fixture).

The one red check on this head, `go-test (Windows) (1.26.x)`, is an ambient
runner-infra flake ("the hosted runner lost communication with the server"),
not a test failure — `main`'s own CI at `aa088149` fails the identical
Windows job the same way. Not this branch; no action needed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed ChainSync sessions that could remain stuck while awaiting a reply after a connection error.
  * Connections now close cleanly when awaiting operations encounter iterator failures, missing results, or connection errors.
  * Preserved graceful handling of canceled operations without unnecessarily closing the connection.

* **Tests**
  * Added coverage for error and cancellation scenarios to help ensure peers are consistently released from stalled states.

### DIFF
--- a/ouroboros/chainsync.go
+++ b/ouroboros/chainsync.go
@@ -762,59 +762,79 @@ func (o *Ouroboros) chainsyncServerRequestNext(
 	if conn == nil {
 		return fmt.Errorf("connection %s not found", ctx.ConnectionId.String())
 	}
-	go func() {
-		// Wait for next block in a separate goroutine so we can
-		// also monitor the connection for errors. This avoids
-		// leaking the monitor goroutine when Next returns first.
-		done := make(chan struct{})
-		var next *chain.ChainIteratorResult
-		var nextErr error
-		go func() {
-			defer close(done)
-			next, nextErr = clientState.ChainIter.Next(true)
-		}()
-		select {
-		case <-done:
-			// Iterator returned
-		case connErr := <-o.chainsyncServerConnErrorChan(conn):
-			// Abandoning the wait here leaves the peer parked in MustReply
-			// with the server holding agency, so the transport has to be
-			// dropped: an error-channel send alone does not unpark it. See
-			// closeChainsyncServerConn.
-			//
-			// conn.ErrorChan() is one buffered channel shared with blockfetch,
-			// tx-submission and the connection manager's own teardown watcher,
-			// so the error taken here may belong to another mini-protocol --
-			// in which case the manager never sees it and would not tear the
-			// connection down at all. Closing is correct either way: it is
-			// what the manager would have done with that error, and it wakes
-			// the manager's watcher through the closed error channel.
-			clientState.ChainIter.Cancel()
-			o.closeChainsyncServerConn(
-				conn,
-				ctx.ConnectionId.String(),
-				fmt.Errorf(
-					"connection error while peer awaited a reply: %w",
-					orErrConnectionClosed(connErr),
-				),
-			)
-			return
-		}
-		o.chainsyncServerServeAwaited(ctx, conn, next, nextErr)
-	}()
+	go o.chainsyncServerAwaitNext(ctx, conn, clientState)
 	return nil
 }
 
-// chainsyncServerConnErrorChan returns the channel the post-AwaitReply waiter
-// watches for connection failures: conn.ErrorChan() unless a test has supplied
-// a single-consumer stand-in. See Ouroboros.chainsyncServerConnErrors.
-func (o *Ouroboros) chainsyncServerConnErrorChan(
-	conn *ouroboros.Connection,
-) <-chan error {
-	if o.chainsyncServerConnErrors != nil {
-		return o.chainsyncServerConnErrors(conn)
+// chainsyncServerConnection is the connection surface the post-AwaitReply
+// ChainSync server waiter needs: the error channel it watches while the peer is
+// parked in MustReply, and the Close that is the only thing that releases it.
+//
+// It is an interface for the same reason blockfetchConnection is: conn's error
+// channel is a single buffered channel shared with blockfetch, tx-submission
+// and the connection manager's teardown watcher, and delivery goes to whichever
+// consumer the runtime picks, so a test cannot address this waiter on the real
+// one. Taking the two methods the waiter actually uses keeps that seam in the
+// signature instead of in a production field only tests write.
+//
+// *ouroboros.Connection satisfies it, and the RequestNext callback above passes
+// exactly that.
+type chainsyncServerConnection interface {
+	ErrorChan() chan error
+	Close() error
+}
+
+// chainsyncServerAwaitNext is the post-AwaitReply waiter. It runs as its own
+// goroutine once the server has parked the peer in MustReply, and either serves
+// the block the iterator eventually yields or drops the transport.
+func (o *Ouroboros) chainsyncServerAwaitNext(
+	ctx ochainsync.CallbackContext,
+	conn chainsyncServerConnection,
+	clientState *chainsync.ChainsyncClientState,
+) {
+	// Wait for next block in a separate goroutine so we can
+	// also monitor the connection for errors. This avoids
+	// leaking the monitor goroutine when Next returns first.
+	done := make(chan struct{})
+	var next *chain.ChainIteratorResult
+	var nextErr error
+	go func() {
+		defer close(done)
+		next, nextErr = clientState.ChainIter.Next(true)
+	}()
+	select {
+	case <-done:
+		// Iterator returned
+	case connErr := <-conn.ErrorChan():
+		// Abandoning the wait here leaves the peer parked in MustReply
+		// with the server holding agency, so the transport has to be
+		// dropped: an error-channel send alone does not unpark it. See
+		// closeChainsyncServerConn.
+		//
+		// conn.ErrorChan() is one buffered channel shared with blockfetch,
+		// tx-submission and the connection manager's own teardown watcher,
+		// so the error taken here may belong to another mini-protocol --
+		// in which case the manager never sees it and would not tear the
+		// connection down at all. Closing is correct either way: it is
+		// what the manager would have done with that error, and it wakes
+		// the manager's watcher through the closed error channel.
+		//
+		// A closed channel delivers a nil error and is the ordinary way in:
+		// gouroboros' Connection.shutdown closes the channel it owns, so
+		// every consumer wakes at once. orErrConnectionClosed keeps the
+		// logged reason meaningful in that case.
+		clientState.ChainIter.Cancel()
+		o.closeChainsyncServerConn(
+			conn,
+			ctx.ConnectionId.String(),
+			fmt.Errorf(
+				"connection error while peer awaited a reply: %w",
+				orErrConnectionClosed(connErr),
+			),
+		)
+		return
 	}
-	return conn.ErrorChan()
+	o.chainsyncServerServeAwaited(ctx, conn, next, nextErr)
 }
 
 // errChainsyncAwaitConnectionClosed stands in for the nil value a closed
@@ -845,7 +865,7 @@ func orErrConnectionClosed(err error) error {
 // silently is not a third option.
 func (o *Ouroboros) chainsyncServerServeAwaited(
 	ctx ochainsync.CallbackContext,
-	conn *ouroboros.Connection,
+	conn chainsyncServerConnection,
 	next *chain.ChainIteratorResult,
 	nextErr error,
 ) {
@@ -930,7 +950,7 @@ func (o *Ouroboros) chainsyncServerServeAwaited(
 }
 
 func (o *Ouroboros) reportChainsyncServerAsyncError(
-	conn *ouroboros.Connection,
+	conn chainsyncServerConnection,
 	connectionID string,
 	operation string,
 	err error,
@@ -966,7 +986,7 @@ func (o *Ouroboros) reportChainsyncServerAsyncError(
 // point. It also closes the gouroboros-owned error channel, which wakes the
 // connmanager watcher without racing a Dingo send against channel closure.
 func (o *Ouroboros) closeChainsyncServerConn(
-	conn *ouroboros.Connection,
+	conn chainsyncServerConnection,
 	connectionID string,
 	err error,
 ) {
--- a/ouroboros/chainsync.go
+++ b/ouroboros/chainsync.go
@@ -776,71 +776,157 @@ func (o *Ouroboros) chainsyncServerRequestNext(
 		select {
 		case <-done:
 			// Iterator returned
-		case <-conn.ErrorChan():
+		case connErr := <-o.chainsyncServerConnErrorChan(conn):
+			// Abandoning the wait here leaves the peer parked in MustReply
+			// with the server holding agency, so the transport has to be
+			// dropped: an error-channel send alone does not unpark it. See
+			// closeChainsyncServerConn.
+			//
+			// conn.ErrorChan() is one buffered channel shared with blockfetch,
+			// tx-submission and the connection manager's own teardown watcher,
+			// so the error taken here may belong to another mini-protocol --
+			// in which case the manager never sees it and would not tear the
+			// connection down at all. Closing is correct either way: it is
+			// what the manager would have done with that error, and it wakes
+			// the manager's watcher through the closed error channel.
 			clientState.ChainIter.Cancel()
+			o.closeChainsyncServerConn(
+				conn,
+				ctx.ConnectionId.String(),
+				fmt.Errorf(
+					"connection error while peer awaited a reply: %w",
+					orErrConnectionClosed(connErr),
+				),
+			)
 			return
 		}
-		if nextErr != nil {
-			// Don't log context.Canceled errors as they're
-			// expected during connection closure.
-			if !errors.Is(nextErr, context.Canceled) {
-				o.config.Logger.Debug(
-					"failed to get next block from chain iterator",
-					"error", nextErr,
-				)
-			}
-			return
-		}
-		if next == nil {
+		o.chainsyncServerServeAwaited(ctx, conn, next, nextErr)
+	}()
+	return nil
+}
+
+// chainsyncServerConnErrorChan returns the channel the post-AwaitReply waiter
+// watches for connection failures: conn.ErrorChan() unless a test has supplied
+// a single-consumer stand-in. See Ouroboros.chainsyncServerConnErrors.
+func (o *Ouroboros) chainsyncServerConnErrorChan(
+	conn *ouroboros.Connection,
+) <-chan error {
+	if o.chainsyncServerConnErrors != nil {
+		return o.chainsyncServerConnErrors(conn)
+	}
+	return conn.ErrorChan()
+}
+
+// errChainsyncAwaitConnectionClosed stands in for the nil value a closed
+// conn.ErrorChan() yields, so the reason logged for the teardown is never an
+// empty error.
+var errChainsyncAwaitConnectionClosed = errors.New("connection closed")
+
+// orErrConnectionClosed substitutes a non-nil error for the nil a closed
+// error channel delivers.
+func orErrConnectionClosed(err error) error {
+	if err == nil {
+		return errChainsyncAwaitConnectionClosed
+	}
+	return err
+}
+
+// chainsyncServerServeAwaited delivers the chain iterator result that resolved
+// a post-AwaitReply wait, or drops the transport when there is nothing to
+// deliver.
+//
+// Once MsgAwaitReply is on the wire the server holds agency in MustReply. A
+// peer parked there cannot make progress and cannot detect abandonment except
+// through gouroboros' randomized MustReply timer (135-269s), after which it
+// tears the connection down, reconnects, re-intersects, receives one
+// MsgRollBackward and parks again -- a stable loop that keeps a node pinned
+// behind a healthy upstream with no ERROR logged on either side. So every path
+// out of here either sends a reply or closes the connection; returning
+// silently is not a third option.
+func (o *Ouroboros) chainsyncServerServeAwaited(
+	ctx ochainsync.CallbackContext,
+	conn *ouroboros.Connection,
+	next *chain.ChainIteratorResult,
+	nextErr error,
+) {
+	if nextErr != nil {
+		if errors.Is(nextErr, context.Canceled) {
+			// The only cancellations of a server client's iterator come from
+			// chainsync.State.RemoveClient (driven by connection-closed
+			// handling) and from the error-channel exit above, both of which
+			// have already dropped the transport. There is no parked peer left
+			// to unpark, so this stays a quiet, expected unwind.
 			o.config.Logger.Debug(
-				"chainsync server: goroutine got nil block",
+				"chainsync server: await unwound by connection teardown",
 				"connection_id", ctx.ConnectionId.String(),
 			)
 			return
 		}
-		tip := o.refreshTip(next)
-		if next.Rollback {
-			if err := ctx.Server.RollBackward(
-				next.Point,
-				tip,
-			); err != nil {
-				o.reportChainsyncServerAsyncError(
-					conn,
-					ctx.ConnectionId.String(),
-					"RollBackward",
-					err,
-				)
-			}
-		} else {
-			blockCbor, blockErr := o.chainsyncServerBlockCbor(ctx, next.Block)
-			if blockErr != nil {
-				// Do not RollForward an incomplete CertRB. This runs after the
-				// callback returned AwaitReply, so actively close the transport
-				// (an error-channel send alone does not) to unpark the client
-				// from AwaitReply so it reconnects and retries the point once
-				// the endorser closure is available.
-				o.closeChainsyncServerConn(
-					conn,
-					ctx.ConnectionId.String(),
-					blockErr,
-				)
-				return
-			}
-			if err := ctx.Server.RollForward(
-				next.Block.Type,
-				blockCbor,
-				tip,
-			); err != nil {
-				o.reportChainsyncServerAsyncError(
-					conn,
-					ctx.ConnectionId.String(),
-					"RollForward",
-					err,
-				)
-			}
+		o.closeChainsyncServerConn(
+			conn,
+			ctx.ConnectionId.String(),
+			fmt.Errorf(
+				"chain iterator failed while peer awaited a reply: %w",
+				nextErr,
+			),
+		)
+		return
+	}
+	if next == nil {
+		// Defensive: ChainIterator.Next maps an empty non-error result to
+		// ErrIteratorChainTip, so this is not reachable through the real
+		// iterator. Should it ever become reachable, dropping the transport is
+		// still the only way to release the parked peer.
+		o.closeChainsyncServerConn(
+			conn,
+			ctx.ConnectionId.String(),
+			errors.New(
+				"chain iterator returned no block while peer awaited a reply",
+			),
+		)
+		return
+	}
+	tip := o.refreshTip(next)
+	if next.Rollback {
+		if err := ctx.Server.RollBackward(
+			next.Point,
+			tip,
+		); err != nil {
+			o.reportChainsyncServerAsyncError(
+				conn,
+				ctx.ConnectionId.String(),
+				"RollBackward",
+				err,
+			)
 		}
-	}()
-	return nil
+		return
+	}
+	blockCbor, blockErr := o.chainsyncServerBlockCbor(ctx, next.Block)
+	if blockErr != nil {
+		// Do not RollForward an incomplete CertRB. This runs after the
+		// callback returned AwaitReply, so actively close the transport
+		// (an error-channel send alone does not) to unpark the client
+		// from AwaitReply so it reconnects and retries the point once
+		// the endorser closure is available.
+		o.closeChainsyncServerConn(
+			conn,
+			ctx.ConnectionId.String(),
+			blockErr,
+		)
+		return
+	}
+	if err := ctx.Server.RollForward(
+		next.Block.Type,
+		blockCbor,
+		tip,
+	); err != nil {
+		o.reportChainsyncServerAsyncError(
+			conn,
+			ctx.ConnectionId.String(),
+			"RollForward",
+			err,
+		)
+	}
 }
 
 func (o *Ouroboros) reportChainsyncServerAsyncError(
--- a/ouroboros/chainsync_server_fixture_test.go
+++ b/ouroboros/chainsync_server_fixture_test.go
@@ -1075,34 +1075,29 @@ func newChainsyncServerFixtureLogging(
 	return f, logBuf
 }
 
-// TestChainsyncServerRequestNextConnErrorAfterAwaitReplyUnparksClient covers
-// the error-channel exit: the waiter gives up on the connection while the peer
-// is parked, so it must drop the transport rather than return silently. Neither
-// consumer of the error closes it otherwise -- the waiter returned silently and
+// TestChainsyncServerAwaitedWaiterClosesOnConnectionError covers the waiter's
+// error-channel exit: it gives up on the connection while the peer it parked is
+// still in MustReply, so it must drop the transport rather than return
+// silently. Nothing else closes it -- the waiter returned silently and
 // ConnectionManager.RemoveConnection only unregisters -- so before the fix the
-// peer stayed parked whichever consumer won.
+// peer stayed parked whichever consumer took the error.
 //
-// The peer is parked through the real protocol first, so this is the genuine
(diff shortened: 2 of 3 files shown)


chadaibbhot-git:
@coderabbitai review (final head 80177610)

ouroboros/chainsync_server_fixture_test.go
```diff
@@ -873,3 +902,109 @@ func TestChainsyncServerRequestNextMissingConnectionAfterAwaitReply(
...
+// the peer stayed parked whichever consumer won the race.
+func TestChainsyncServerRequestNextConnErrorAfterAwaitReplyUnparksClient(
+	t *testing.T,
+) {
+	f := newChainsyncServerFixture(t, csmock.ModeNtC)
+	f.parkInAwaitReply(t)
+
+	errChan := f.conn.ErrorChan()
+	errChan <- errors.New("simulated protocol error for the parked waiter")
+	errChan <- errors.New("simulated protocol error for the connmanager watcher")
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid — and not merely theoretical. Fixed in d0423e5 by removing the second send
entirely.

Confirming the mechanism before changing it: `Connection.Close()` calls
`requestClose()` and waits on `connClosedChan`; `shutdown()` does
`close(c.errorChan)` whenever `ownsErrorChan` is set, which it is for a
connection built without `WithErrorChan` — the harness's. So the first consumer
closing the connection does close the channel the second send writes to, and the
window is real. gouroboros' own producer side never does a bare send for exactly
this reason: `forwardError` selects on `<-c.doneChan` with a `default`.

The second send was there because a test cannot address this waiter on
`conn.ErrorChan()` at all: the connection manager's teardown watcher is blocked
on the same channel, delivery goes to whichever consumer the runtime picks, and
on `main` neither closes the transport (`RemoveConnection` only unregisters), so
a single send is nondeterministic. Covering the other consumer with an extra
error is what created the race you flagged.

So the waiter now reads its channel through `chainsyncServerConnErrorChan`, which
returns `conn.ErrorChan()` in production and lets a test hand it a
single-consumer stand-in. The test publishes exactly one error onto a channel
nothing else consumes and nothing ever closes. What the waiter does with an error
it has received — cancel the iterator, close the transport — is identical either
way, and that is what the test asserts.

The seam is set through the fixture before the harness (and therefore any
protocol goroutine) exists, so there is no race on the field either, and
`TestChainsyncServerConnErrorChanDefaultsToTheConnection` asserts the production
default is `conn.ErrorChan()` so the seam cannot mask a waiter wired to the wrong
channel. That test fails if the default is changed; the exit-(1) test still fails
if the `closeChainsyncServerConn` call is removed.

ouroboros/chainsync.go
```diff
@@ -776,71 +776,157 @@ func (o *Ouroboros) chainsyncServerRequestNext(
...
+// out of here either sends a reply or closes the connection; returning
+// silently is not a third option.
+func (o *Ouroboros) chainsyncServerServeAwaited(
+	ctx ochainsync.CallbackContext,
+	conn *ouroboros.Connection,
+	next *chain.ChainIteratorResult,
+	nextErr error,
+) {
+	if nextErr != nil {
+		if errors.Is(nextErr, context.Canceled) {
```
chrisguiney, line 873: Confirmed against this head, since the whole fix rests on this branch being unreachable while a peer is still parked.

`ChainIter.Cancel()` has exactly two call sites: `chainsync/chainsync.go:404` (`State.RemoveClient`) and the error-channel exit above. `RemoveClient` has exactly two callers, `ouroboros/ouroboros.go:769` (`HandleConnClosedEvent`, NtN) and `node.go:1872` (`handleConnManagerClosed`, NtC), both reached only after the transport is gone. The remaining way in is the `ConnectionId`-collision eviction, and `connmanager` calls `closeConnAndLog` on the evicted connection before `notifyEvictedConnectionClosed` reaches `ConnClosedFunc`, so that one is closed first too.

No live-connection cancellation reaches here, so staying quiet is correct.

ouroboros/chainsync.go
```diff
@@ -776,71 +776,157 @@ func (o *Ouroboros) chainsyncServerRequestNext(
...
+}
+
+// errChainsyncAwaitConnectionClosed stands in for the nil value a closed
+// conn.ErrorChan() yields, so the reason logged for the teardown is never an
+// empty error.
+var errChainsyncAwaitConnectionClosed = errors.New("connection closed")
+
+// orErrConnectionClosed substitutes a non-nil error for the nil a closed
+// error channel delivers.
+func orErrConnectionClosed(err error) error {
```
chrisguiney, line 847: Non-blocking: the `err == nil` branch is the shape that fires most often in production and is the one branch here with no coverage.

A closed `conn.ErrorChan()` is how this exit is normally reached — `Connection.shutdown` closes the channel it owns, and every consumer then receives the zero value — whereas a live error delivered to *this* consumer is the rarer race. `TestChainsyncServerRequestNextConnErrorAfterAwaitReplyUnparksClient` sends a real error, so the sentinel substitution is never exercised.

The stand-in seam makes it a one-liner: `close(connErrs)` instead of sending on it, in a sibling test. Behavior is a `Close()` either way, so only the logged reason differs — hence non-blocking.

ouroboros/ouroboros.go
```diff
@@ -151,6 +151,18 @@ type Ouroboros struct {
...
+	// production uses.
+	//
+	// It is overridable because conn.ErrorChan() is a single buffered channel
+	// shared with blockfetch, tx-submission and the connection manager's
+	// teardown watcher, and delivery goes to whichever consumer the runtime
+	// picks. A test cannot address this waiter on it, and cannot publish an
+	// extra error to cover the other consumers either: the consumer that wins
+	// closes the connection, and gouroboros' Connection.shutdown closes the
+	// error channel it owns, so the extra send races that closure.
+	chainsyncServerConnErrors func(*ouroboros.Connection) <-chan error
```
chrisguiney: Non-blocking design note: this is a production struct field only ever written by tests, and the same package already solves the same problem with a narrow interface — `blockfetchConnection` in `ouroboros/blockfetch.go:89` declares `ErrorChan() chan error` precisely so blockfetch's error-channel handling is testable.

The waiter needs `ErrorChan()` and `Close()`, so a two-method `chainsyncServerConnection` interface taking the place of `*ouroboros.Connection` in `chainsyncServerServeAwaited` and the waiter would cover both this seam and `closeChainsyncServerConn`, and keep the hook out of `Ouroboros`. It would also drop the need for `TestChainsyncServerConnErrorChanDefaultsToTheConnection`, since there would be no default to get wrong.

The rationale in the comment is accurate and the field is not reachable in production, so this is a preference, not a blocker.

chrisguiney (dismissed):
Approving. The three exits now do what the `CertRB` branch already did, and that matches the async-close rule `ARCHITECTURE.md` states for the post-`AwaitReply` path — though that sentence is still scoped to the `CertRB` terminal cases and could be broadened now that it is the rule for every async exit.

Checked against this head:

- **The class is complete within ChainSync.** `ctx.Server.AwaitReply()` has exactly one call site, and `chainsyncServerConnOpts()` is installed on the NtC listener and on both NtN configurations, so both roles get the fix.
- **The `context.Canceled` carve-out holds.** `ChainIter.Cancel()` has exactly two call sites — `chainsync.State.RemoveClient` and this waiter's error-channel exit — and `RemoveClient` has exactly two callers, `HandleConnClosedEvent` and `handleConnManagerClosed`, both driven after the transport is gone. The `ConnectionId`-collision path also closes the evicted connection before `notifyEvictedConnectionClosed` reaches `ConnClosedFunc`, so no live-connection cancellation is left to fall into the quiet branch.
- **Repeated close is safe.** In gouroboros v0.204.0 `Connection.shutdown` is `sync.Once`-guarded and `Close()` returns once `connClosedChan` is closed, so the waiter racing connmanager's own teardown is fine.
- **Exit (2) does not fire at ordinary tip.** The server iterator runs on the persistent primary chain, where `reconcile()` returns immediately, and `nextPersistentBlockAfterSparseIndex` returns early when `nextBlockIndex > tipBlockIndex` — so the at-tip case reaches the blocking wait rather than an error, and tip-following peers are not torn down.
- **The tests fail without the fix.** Reverting each branch in place makes its own test fail; the three abandonment tests fail with `context deadline exceeded`, which is the reported symptom.

One item to settle before merge: `Fixes #4137` closes more than this changes. That issue's suggested fixes 2 (the shared `conn.ErrorChan()`), 3 (`checkLocalTipPlateau` falling back to the advertised tip) and 4 (escalation after unproductive plateau resyncs), plus its "Secondary" watchdog section, are explicitly out of scope here. Merging as written auto-closes the issue and leaves that work unowned. Referencing #4137 without the closing keyword, or splitting the remainder into follow-up issues, would keep it tracked.

Adjacent, pre-existing, and not this PR's to fix: `ouroboros/blockfetch.go:314` has the same shape one protocol over — `case <-conn.ErrorChan(): return nil` after `StartBatch()` is on the wire, with no `closeBlockfetchConnection`. If that error belonged to another mini-protocol, the peer is left in `Streaming` awaiting `BatchDone`.

ouroboros/chainsync_server_fixture_test.go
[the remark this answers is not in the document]
chrisguiney: Already addressed at `d0423e54`, and verified here: the test no longer sends on a channel another consumer may close. `TestChainsyncServerRequestNextConnErrorAfterAwaitReplyUnparksClient` publishes one error on a test-owned buffered channel supplied through `chainsyncServerConnErrors`, so no send races `Connection.shutdown`'s `close(c.errorChan)`.

The mechanism the finding described is real in gouroboros v0.204.0: `shutdown()` closes `errorChan` whenever `ownsErrorChan` is set, which it is for a connection built without `WithErrorChan`.

ouroboros/chainsync.go
chadaibbhot-git: Added at `9ef8357f` as `TestChainsyncServerAwaitedWaiterClosesOnErrorChannelClosure`: it closes the stand-in error channel instead of sending on it, and asserts the waiter returned, that it called `Close()` itself (`closeCount() == 1`), that the transport is gone (`Observe` returns `csmock.ErrClosed`), and that the reason logged is `errChainsyncAwaitConnectionClosed` and not a `%!w(<nil>)` from formatting the nil the closed channel yields.

One deliberate difference from its sibling: it does not park through the protocol first. A parked peer arms the production waiter on the real error channel, and the delegated `Close()` then wakes that waiter with a channel closure of its own whose teardown reason is textually identical to the one under test — so with the peer parked the reason assertion would be satisfied by the wrong waiter. With no second waiter it is attributable. The error-delivery sibling keeps the real parked-peer setup, since its reason (the error text it sends) is unique either way.

Mutation check: removing the `closeChainsyncServerConn` call from the waiter turns both tests red on the `ErrorIs(err, csmock.ErrClosed)` assertion.

ouroboros/ouroboros.go
chadaibbhot-git: Taken, at `9ef8357f`. `chainsyncServerConnection` (`ouroboros/chainsync.go`) declares exactly `ErrorChan() chan error` and `Close() error`, matching `blockfetchConnection`, and is now the parameter type of `chainsyncServerServeAwaited`, `closeChainsyncServerConn` and `reportChainsyncServerAsyncError` (that last one also closes the connection, so it belonged in the same set).

The waiter moved out of the anonymous goroutine into `chainsyncServerAwaitNext(ctx, conn chainsyncServerConnection, clientState *chainsync.ChainsyncClientState)`; `chainsyncServerRequestNext` is now `go o.chainsyncServerAwaitNext(ctx, conn, clientState)` with `conn` still the `*ouroboros.Connection` the connection manager resolved.

Deleted: the `chainsyncServerConnErrors` field on `Ouroboros`, the `chainsyncServerConnErrorChan` helper, and `TestChainsyncServerConnErrorChanDefaultsToTheConnection` — with the seam in the signature there is no default left to get wrong. In its place the test file carries `var _ chainsyncServerConnection = (*ouroboros.Connection)(nil)`, so the stand-in cannot drift from the type production passes, and the parked-peer test still runs against the real connection: the stand-in delegates `Close()` to it, so the assertion is still the actual transport being dropped.

All existing tests in the package stay green (`go test ./ouroboros/` and `-race`), and removing the `closeChainsyncServerConn` call from the waiter turns both unpark tests red.

ouroboros/chainsync.go
chadaibbhot-git: Acknowledged — that matches the call-graph as we read it, including the `ConnectionId`-collision eviction going through `closeConnAndLog` before `notifyEvictedConnectionClosed`. The quiet `context.Canceled` unwind stays as it is, and `TestChainsyncServerServeAwaitedCancelledDoesNotCloseConnection` still pins it after the interface refactor at `9ef8357f`.

ouroboros/chainsync_server_fixture_test.go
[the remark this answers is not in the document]
chadaibbhot-git: Acknowledged. The test-owned channel is unchanged in substance at `9ef8357f` — it now reaches the waiter as a `chainsyncServerConnection` stand-in rather than through a field on `Ouroboros`, so there is still no send racing `Connection.shutdown`'s `close(c.errorChan)`.

ouroboros/chainsync_server_fixture_test.go
```diff
@@ -873,3 +909,284 @@ func TestChainsyncServerRequestNextMissingConnectionAfterAwaitReply(
...
+	require.Equal(t, f.conn.Id(), observed)
+	// AddClient returns the state the parked RequestNext registered rather
+	// than a second one, so the waiter below shares the peer's real iterator.
+	clientState, err := f.o.chainsyncState.AddClient(
+		f.conn.Id(),
+		ocommon.NewPointOrigin(),
+	)
+	require.NoError(t, err)
+
+	conn := newStubChainsyncServerConnection(f.conn.Close)
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid, and fixed at `051998c5`. `AddClient` does return the already-registered state, so the test started a second `ChainIter.Next(true)` on the parked peer's iterator and the error went to the synthetic waiter — the parking read as end-to-end coverage it was not providing.

Both waiter tests now register the client directly and drive `chainsyncServerAwaitNext` once, as the error-channel-closure sibling already did. The stand-in's `Close` is still the real connection's, so the assertion is still the actual transport being dropped, and end-to-end release of a genuinely parked peer remains covered by `TestChainsyncServerRequestNextIteratorErrorAfterAwaitReplyUnparksClient` and `TestChainsyncServerRequestNextNilBlockAfterAwaitReplyUnparksClient`, which park through the protocol and drive the serve path on the real connection with no second iterator consumer.

Renamed to `TestChainsyncServerAwaitedWaiterClosesOnConnectionError` so the name no longer claims the parked-peer path. Both tests still go red when the `closeChainsyncServerConn` call is removed from the waiter; `go test ./ouroboros/` and `-race` are green.

ouroboros/chainsync_server_fixture_test.go
```diff
@@ -873,3 +909,279 @@ func TestChainsyncServerRequestNextMissingConnectionAfterAwaitReply(
...
+// connection whose Close is the real one, rather than through a peer parked by
+// the protocol. Parking arms the production waiter on the real error channel,
+// and a test cannot address one consumer of that shared channel, so the error
+// would still have to go to a second waiter -- which would then be a second
+// goroutine driving the same ChainIter. The end-to-end release of a genuinely
+// parked peer is covered by
+// TestChainsyncServerRequestNextIteratorErrorAfterAwaitReplyUnparksClient and
+// TestChainsyncServerRequestNextNilBlockAfterAwaitReplyUnparksClient, which
+// park through the protocol and drive the serve path on the real connection.
+func TestChainsyncServerAwaitedWaiterClosesOnConnectionError(
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid — the first paragraph and both assertion strings were left over from the version that parked through the protocol. Fixed at `3b1259c2` (comments and assertion messages only).

The leading paragraph now says the waiter consumes an error for the connection it is serving and must close the transport, and why closing is the whole point of the exit: once `MsgAwaitReply` is on the wire the server holds agency in MustReply, and an error-channel send, a silent return and `ConnectionManager.RemoveConnection` all fail to release the peer. The second paragraph keeps the reason the waiter is driven directly and now attributes protocol-level parked-peer coverage to `TestChainsyncServerRequestNextIteratorErrorAfterAwaitReplyUnparksClient` and `TestChainsyncServerRequestNextNilBlockAfterAwaitReplyUnparksClient`.

Both assertion strings became "must drop the transport, which is the only thing that releases a peer the server has parked in MustReply", which is true of what the test does rather than claiming this test parked one.

wolf31o2 (approved):
Accept

Issue #2150, opened 2026-05-03
chainsync periodically stalls; peer-switch via plateau watchdog does not recover, container restart required


wolf31o2:
## Summary

After several hours of healthy operation, chainsync stops applying new headers from the active peer. The local-tip-plateau watchdog detects the stall and recycles to a different peer, but the new peer's chainsync client lands in the same bad state. Two or three plateau cycles can fire against different peers without local tip advancing past the same slot. The only reliable cure is `docker restart` of the container; immediately after restart, chainsync resumes normally.

## Environment

- dingo v0.39.1 (`ghcr.io/blinklabs-io/dingo:0.39.1`, linux/arm64)
- preview, block-producer mode
- ~5 active peers (mix of localRoots + bootstrap-discovered)
- Mithril-bootstrapped data dir

## Pattern

1. Node runs healthy for ~1 to ~6 hours, applying live-tip blocks at the expected rate.
2. Active peer's chainsync stops delivering. Local tip freezes at some slot S, no further `chain extended` events with `source: chainsync`.
3. After 240-270 s the plateau watchdog fires:

   ```
   WARN  local tip plateau detected, resyncing chainsync client
         connection_id=<container>:3001<-><peer-A>:3001
         local_tip_slot=S  best_peer_tip_slot=S+285  plateau_duration=270000...
   ```

4. Watchdog closes the stalled connection, chain selection picks a new best peer, active connection switches.
5. The new peer is connected and selected, but local tip still does not advance past S. ~4 minutes later the watchdog fires again against the new peer, same `local_tip_slot=S`. Pattern repeats across 2-3 peers.
6. `docker restart dingo` immediately recovers: within seconds of startup, chainsync extends past S.

## Reproduction

Run a v0.39.1 BP on preview. Wait. Stalls observed at uptimes of 1h, 2h, and 6h on the same host with the same topology. No load involvement.

## Expected behavior

When the plateau watchdog recycles chainsync against a healthy peer with a higher tip, the local tip should resume advancing within one round-trip of the new chainsync handshake.

## Observed behavior (sanitized excerpt)

Same `local_tip_slot=111163669` across three watchdog cycles against three different peers; only restart unwedged it.

```
14:52:47  WARN local tip plateau detected, resyncing chainsync client
              connection_id=<container>:3001<-><peer-A>:3001
              local_tip_slot=111163669 best_peer_tip_slot=111163954 plateau_duration=270000000000
[...]

PR #4139, opened 2026-09-08
fix(chainsyncrecycler): stop the plateau watchdog disarming itself after a resync

chadaibbhot-git:
## The bug

`checkLocalTipPlateau` compares the local tip against the best peer's
**delivered** frontier (`PeerChainTip.SelectionTip`) and returns when that
frontier is not ahead (`internal/chainsyncrecycler/recycler.go`):

```go
	bestPeerTipSlot := bestPeerTip.SelectionTip().Point.Slot
	if bestPeerTipSlot <= localTipSlot {
		return
	}
```

Immediately after a plateau resync it never is. The watchdog's own recovery
action closes the connection for a fresh chainsync. The peer reconnects,
re-intersects at the local tip and delivers nothing but its post-intersect
`RollBackward`, so its delivered frontier **is** the local tip. The comparison
above then disarms the watchdog, and it cannot re-arm until some peer delivers a
header, which is exactly what is not happening when the upstream has stopped
speaking. The recovery action disables the detector that triggered it.

This was observed on a one-relay devnet. A producer's applied tip stayed pinned
for 20 minutes and 1282 slots while the relay extended its own chain 40+ times,
and the peer's advertised tip was correctly reported 811 slots ahead the whole
time. On a second node, the same starvation ran for ten minutes with the plateau
firing exactly once, at the start. Detail is in #4137, which also covers the
server-side cause of the starvation (#4138).

## Dependency

This relies on #3989, which is now on `main`. After a plateau resync, the
connection is closed (`LocalTipPlateau` is in
`chainsyncResyncRequiresFreshConnection`) and `node.go` removes the peer on
`ConnectionClosedEvent`. The replacement is tracked only because #3989 registers
it from its post-intersect rollback (`registerPeerFromRollbackLocked`) and
exempts it from the behind-filters in `isPeerSelectableLocked` while
`awaitingFirstHeader` is set. Without both, `GetBestPeer()` is nil throughout
the window and the plateau check returns before reaching this change. The branch
merges current `main`; its diff is the recycler change, the `AwaitingFirstHeader`
accessor, tests and docs.

## The change

While the best peer is still awaiting its first header, use its **advertised**
tip. That value is known and correct throughout the window; it is what tells us
the peer is 800 slots ahead.

`PeerChainTip.AwaitingFirstHeader` reads the `awaitingFirstHeader` flag. Chain
selection sets that flag only when it registers a peer from a rollback, and
clears it on the peer's first delivered header. The accessor deliberately does
not test for a delivered block number of 0. A tracked peer that has delivered
headers and then rolls back outside its retained delivered-header history is
also left at block number 0, but it has delivered headers on this connection,
so it is judged on its delivered frontier like any other peer.

The advertised tip is untrusted, so the substitution is confined so that a peer
can only ever spend it on itself:

- **only** while the peer awaits its first header. Once it has delivered a
  header, its delivered frontier is real evidence and is used, which is why
  `SelectionTip` prefers the delivered value in the first place.
- **only** when it raises the comparison value, never when it would lower it.
  A peer that advertises a tip behind the rollback point it just sent therefore
  cannot suppress a plateau that fires on `main` today.
- **only** when the advertising peer is the connection this watchdog would
  recycle. The recycle targets the *active* chainsync client, which need not be
  the best peer. Without this condition, a peer that fabricates a high tip and
  delivers nothing could drive repeated recycles of a different, honest
  upstream. With it, the worst it can do is have its own connection reconnected
  once per cooldown.

That last check compares `ConnectionId` values, not the rendered address string
the recycler uses to key its cooldown map. With listen-port reuse, a replacement
connection renders identically to the one it replaced, so a stale peer tip
could otherwise authorize recycling the replacement.

A plateau still requires the local tip to have stood still for a full plateau
window, so none of this can fire on a node that is applying blocks.

## Tests

`chainselection`:
- `TestAwaitingFirstHeaderAfterPostIntersectRollback`: a rollback-registered peer
  awaits its first header.
- `TestAwaitingFirstHeaderFalseAfterRollbackOutsideDeliveredHistory`: a peer that
  has delivered a header and then rolls back outside its history does not await
  one, even though its delivered block number is 0.
- `TestAwaitingFirstHeaderClearsOnDeliveredHeader`: the first delivered header
  clears the flag, and a later rollback does not set it again.
- `TestAwaitingFirstHeaderNilReceiver`

`internal/chainsyncrecycler`, on a real `chainselection.NewChainSelector`:
- `TestTickResyncsOnPlateauAfterRecycleWithRealChainSelector` runs the full
  post-resync sequence (`UpdatePeerTip`, `RemovePeer`, rollback from the
  replacement) and expects exactly one `LocalTipPlateau` resync for the
  replacement.
- `TestTickDoesNotResyncOnceTheReplacementHasDeliveredAHeader`: after a header
  and a rollback *inside* retained history, the advertisement is no longer used.
- `TestTickDoesNotResyncAfterADeliveredPeerRollsBackOutsideItsHistory`: after a
  header and a rollback *outside* retained history, on a chain shallower than K
  so that the peer is still selectable, the advertisement is not used either.

`internal/chainsyncrecycler`, against `fakeChainSelector`; where a test needs an awaiting peer, it
comes from a real selector's rollback registration:
- `TestTickResyncsOnPlateauWhenBestPeerAwaitsFirstHeader`: the fix.
- `TestTickDoesNotResyncWhenAwaitingPeerAdvertisesNoProgress`
- `TestTickIgnoresAdvertisedTipOnceBestPeerHasDeliveredAHeader`
- `TestTickPlateauFallbackNeverLowersTheBestPeerTip`: the `>` guard.
- `TestTickPlateauIgnoresAdvertisedTipOfAThirdPartyPeer`: the own-connection
  guard.
- `TestTickPlateauIgnoresAdvertisedTipOfAReplacedConnection`: `ConnectionId`
  comparison rather than `String()`.
- `TestTickPlateauUsesAdvertisedTipWithNoActiveClient`
- `TestTickPlateauFallbackFeedsTheBacklogClassifier`,
  `TestTickPlateauFallbackResyncsWhenHeaderGapDominates`: the substituted tip
  also drives `isLedgerApplicationBacklog`.

Each guard and the accessor was verified red under a targeted mutation.

## Related

- #4137: the report; its "Secondary" section is this bug.
- #3989: rollback registration in `chainselection` (merged); see "Dependency".
- #3808: the plateau watchdog's threshold (how long it waits). This PR is about
  the watchdog being unable to re-arm at all, which is orthogonal.
- #3886: plateau recovery during catch-up. It touches the same function but has
  no textual overlap with this hunk.
- #3987: post-recycle peer selectability. Adjacent, but it is about selection
  rather than about the watchdog's input.
- #2150, #2177 (closed): plateau watchdog failing to recover a stalled stream.


chadaibbhot-git:
@coderabbitai review (final head 437e48c9)

internal/chainsyncrecycler/recycler.go
```diff
@@ -469,6 +469,24 @@ func (r *Recycler) checkLocalTipPlateau(
...
+	// thing that is not happening. The peer's advertised tip is known and
+	// correct throughout that window, so use it while the peer is still
+	// awaiting its first header.
+	//
+	// The advertised tip is untrusted, hence the narrow condition. Even so the
+	// exposure is small: a plateau only fires after the local tip has not
+	// advanced for a full plateau window, the recycle is rate-limited by the
+	// cooldown, and the action it unlocks -- reconnecting chainsync -- is what
+	// a peer that advertises a tip it never delivers warrants anyway.
+	if bestPeerTip.AwaitingFirstHeader() &&
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid in the part that matters, and fixed in 90c0717.

The amplification was real: the recycle targets the **active** chainsync client
(`live.ChainsyncState.GetClientConnId()`), which need not be the best peer. So a
peer that intersects at our local tip, fabricates a high advertised tip and then
delivers nothing could drive a recycle of a *different, honest* upstream once per
cooldown. That is the part I had not bounded.

The substitution is now confined to the connection the watchdog would actually
recycle (`advertisesForOwnConnection`): the advertising peer must be the active
chainsync client, or there must be no active client, in which case the plateau
falls back to that same peer as its target. A peer can therefore only ever spend
its advertised tip on itself — the worst case is its own connection being
reconnected once per cooldown.

On requiring a trusted or independently validated advertisement before using it
at all: I did not go that far deliberately. The failure this PR fixes was
observed on a single-upstream topology, so there is no second peer to corroborate
against; requiring corroboration would leave exactly that case wedged, which is
the case that motivated the change. What is left is bounded by three things that
together make a fabricated tip uninteresting to send:

- the plateau still requires the local tip to have stood still for a full plateau
  window, so this cannot fire on a node that is applying blocks;
- the recycle is rate-limited by the cooldown;
- reconnecting chainsync is the correct response to a peer that advertises a tip
  it never delivers, so on the peer's own connection the "attack" and the remedy
  are the same action.

New tests, both verified red against a targeted mutation of the branch they
cover:

- `TestTickPlateauIgnoresAdvertisedTipOfAThirdPartyPeer` — a fabricated tip from
  a peer that is not the active chainsync client must not recycle the honest
  client's connection.
- `TestTickPlateauUsesAdvertisedTipWithNoActiveClient` — with no active client
  the target is the awaiting peer itself, so the fallback still applies.

internal/chainsyncrecycler/recycler.go
```diff
@@ -343,6 +343,23 @@ func shouldRecycleLocalTipPlateau(
...
+// ConnectionId holds net.Addr interfaces, so it is compared by its string form
+// the same way the recycler keys its cooldown map.
+func advertisesForOwnConnection(
+	targetConn *ouroboros.ConnectionId,
+	bestPeer ouroboros.ConnectionId,
+) bool {
+	if targetConn == nil {
+		return true
+	}
+	return targetConn.String() == bestPeer.String()
```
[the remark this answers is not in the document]
chadaibbhot-git: Valid, and fixed in 6be7576 — `advertisesForOwnConnection` now compares
`ConnectionId` values rather than their rendered strings.

Checking the identity relation before changing it: `chainselection.peerTips` is
`map[ouroboros.ConnectionId]*PeerChainTip` and `chainsync.clients` /
`trackedClients` are keyed the same way, so Go struct equality on `ConnectionId`
*is* the identity those components use. gouroboros makes that stable by
populating `Connection.id` exactly once (`connection.go`, at connection setup)
and returning it by value from `Id()`, so every copy that reaches the recycler
shares the same `net.Addr` values and compares equal. `==` is therefore both the
correct relation and a no-op for the ordinary single-connection case the fix
targets.

The scenario you describe is the one that made the difference: with listen-port
reuse a replacement connection renders identically to the one it replaced while
being a distinct `ConnectionId`, so `GetBestPeer()` can still be handing back the
stale peer's ID while `GetClientConnId()` returns the replacement. Under string
matching the stale peer's fabricated tip would authorize recycling the
replacement — the honest, active connection.

Worth noting the contrast that the comment now records: the recycler keys its
cooldown map on `targetConn.String()` and that stays as it is. Collapsing a
replaced connection onto one rate-limit key is the conservative choice for a rate
limit; it is the wrong choice for an authorization.

`TestTickPlateauIgnoresAdvertisedTipOfAReplacedConnection` covers it: two
`ConnectionId`s that render identically but are distinct values, with the stale
one as the best peer and the replacement as the active client. It asserts no
resync is published, and fails if `advertisesForOwnConnection` is reverted to
`String()` comparison.

internal/chainsyncrecycler/recycler.go
```diff
@@ -469,10 +493,33 @@ func (r *Recycler) checkLocalTipPlateau(
...
+	// only when the advertising peer is the connection this watchdog would
+	// recycle. Without that last condition a peer that fabricates a high tip
+	// and delivers nothing could drive repeated recycles of a different,
+	// honest upstream. With it, the worst it can do is have its own connection
+	// reconnected once per cooldown, while the local tip is standing still --
+	// which is what a peer that advertises a tip it never delivers warrants.
+	if bestPeerTip.AwaitingFirstHeader() &&
+		bestPeerTip.Tip.Point.Slot > bestPeerTipSlot &&
+		advertisesForOwnConnection(targetConn, *bestPeer) {
+		bestPeerTipSlot = bestPeerTip.Tip.Point.Slot
```
chrisguiney, line 532: The substituted value flows on into `isLedgerApplicationBacklog` as `bestPeerTipSlot`, changing `headerGap` and therefore the backlog-versus-stall classification.

The behavior is right: with an applied tip of 100, a primary chain tip of 900 and an advertised tip of 1000 the apply backlog still dominates and the plateau is logged rather than recycled, and with a primary chain tip of 200 the header gap dominates and the resync fires. But all seven new cases set `primaryChainTipSlot` equal to the applied tip, so `isLedgerApplicationBacklog` returns `false` on its first branch every time and that interaction is uncovered.

Both of those cases fail with the block at 515-519 removed, so they would be regression tests rather than guards. Non-blocking.

internal/chainsyncrecycler/recycler.go
```diff
@@ -469,10 +493,33 @@ func (r *Recycler) checkLocalTipPlateau(
...
+	// The advertised tip is untrusted, so the substitution is confined so that
+	// a peer can only ever spend it on itself: it applies only while the peer
+	// awaits its first header, only when it raises the comparison value, and
+	// only when the advertising peer is the connection this watchdog would
+	// recycle. Without that last condition a peer that fabricates a high tip
+	// and delivers nothing could drive repeated recycles of a different,
+	// honest upstream. With it, the worst it can do is have its own connection
+	// reconnected once per cooldown, while the local tip is standing still --
+	// which is what a peer that advertises a tip it never delivers warrants.
+	if bestPeerTip.AwaitingFirstHeader() &&
```
chrisguiney, line 529: `ARCHITECTURE.md` states the predicate this line changes ("Plateau detection: if the local tip stops advancing while peers are ahead, the recycler first asks ledger to reconcile..."), and the repository treats `ARCHITECTURE.md` as part of the change bar alongside tests. A sentence there for the awaiting-first-header fallback and its own-connection restriction would keep the document in step with the code. Non-blocking.

chainselection/peer_tip.go
```diff
@@ -424,6 +424,22 @@ func (p *PeerChainTip) SelectionTip() ochainsync.Tip {
...
+//
+// Callers that reason about how far a peer has got need this, because for such
+// a peer the delivered frontier is the point the session intersected at rather
+// than anything the peer has shown us, and it does not move until the peer's
+// first RollForward arrives.
+func (p *PeerChainTip) AwaitingFirstHeader() bool {
+	if p == nil {
+		return false
+	}
+	return p.SelectionTip().BlockNumber == 0
```
chrisguiney, line 473: `SelectionTip().BlockNumber == 0` is a proxy for the property the name states. It is also true for a peer whose delivered frontier genuinely is block 0, and for one registered from an advertised origin tip.

That is harmless at the only call site: for an origin peer the `Tip.Point.Slot > bestPeerTipSlot` guard is `0 > 0`, and a node pinned at block 0 while a peer is ahead is exactly the case the resync exists for. Worth keeping in mind for a second consumer, which would inherit the approximation rather than the stated property. Non-blocking.

chrisguiney (dismissed):
Approving.

Verification that mattered:

- Each of the three conditions at `recycler.go:515-517` is independently load-bearing. Removing `AwaitingFirstHeader()`, the `Tip.Point.Slot > bestPeerTipSlot` guard, or `advertisesForOwnConnection` each turns a distinct one of the new tests red, and removing the whole block turns `TestTickResyncsOnPlateauWhenBestPeerAwaitsFirstHeader` and `TestTickPlateauUsesAdvertisedTipWithNoActiveClient` red.
- No new over-fire path. A quiet chain at tip still exits at `bestPeerTipSlot <= localTipSlot`, because an honest peer's advertised tip equals the local tip there. The plateau still requires the applied tip pinned for the full window, and `st.lastProgressAt = now` after a resync keeps the next fire a full window away, so the widened input adds no double-resync path.
- `ConnectionId` equality is the relation `chainselection.peerTips` and `chainsync.trackedClients` already key on, and `chainselection/selector.go` compares `*cs.bestPeerConn == connId` directly at several sites. `advertisesForOwnConnection` reuses that identity rule rather than introducing a new one.
- The widened value makes `ReconcileLivePrimaryChainLedgerDivergence` reachable in the awaiting-first-header window. That call compares `chain.Tip()` against the ledger's own tip and returns `(false, nil)` when they agree, so an untrusted advertised slot cannot force a local rollback; it can only reach the resync for the advertising peer's own connection.

Both open bot threads are addressed at this head; replies on each. Two non-blocking notes inline.

internal/chainsyncrecycler/recycler.go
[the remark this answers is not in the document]
chrisguiney: Confirmed addressed at `6be75769`. `advertisesForOwnConnection` (`recycler.go:517`) confines the substitution to the connection the plateau would actually recycle, so an awaiting peer can only spend its advertised tip on itself. Dropping that condition turns `TestTickPlateauIgnoresAdvertisedTipOfAThirdPartyPeer` red.

internal/chainsyncrecycler/recycler.go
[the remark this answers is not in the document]
chrisguiney: Confirmed addressed at `6be75769`. `advertisesForOwnConnection` compares `ConnectionId` values rather than their rendered strings; substituting a `String()` comparison turns `TestTickPlateauIgnoresAdvertisedTipOfAReplacedConnection` red. This is also the comparison `chainselection/selector.go` already uses for `*cs.bestPeerConn == connId`.

internal/chainsyncrecycler/recycler.go
chadaibbhot-git: Added at `da8321be`, both in `internal/chainsyncrecycler/recycler_test.go`.

`TestTickPlateauFallbackFeedsTheBacklogClassifier` — applied tip 100, `primaryChainTipSlot` 900, advertised tip 1000. The substituted tip carries the plateau past the delivered-frontier comparison, the local reconcile runs, and `isLedgerApplicationBacklog` then sees an apply backlog of 800 against a residual header gap of 100, so the INFO line is logged and the stream is left running. It asserts `reconcileCallCount() == 1` (the value that is zero without the substitution, so it is the substitution being pinned rather than the backlog heuristic), an empty publisher, the log record itself, and `st.lastProgressAt == now`.

`TestTickPlateauFallbackResyncsWhenHeaderGapDominates` — same applied and advertised tips, `primaryChainTipSlot` 200, so the header gap of 800 dominates the apply backlog of 100 and the resync event fires for the connection.

Mutation checks, both from a clean tree restored from a copy afterwards:

- removing the block at 515-519: `FeedsTheBacklogClassifier` fails on `expected: 1 actual: 0` for the reconcile count, `ResyncsWhenHeaderGapDominates` fails on `"[]" should have 1 item(s)`.
- forcing `isLedgerApplicationBacklog` to return true unconditionally: `ResyncsWhenHeaderGapDominates` fails, `FeedsTheBacklogClassifier` still passes — so the second one pins the direction of the classification and not just that it was reached.

internal/chainsyncrecycler/recycler.go
chadaibbhot-git: Done at `da8321be`. The plateau-detection bullet now continues:

> The "peers are ahead" comparison reads a peer's DELIVERED frontier, which a peer that has not delivered a header yet pins at the point its session intersected at — immediately after a plateau resync, the stalled local tip itself — so it would disarm this watchdog exactly when it has just acted; while a peer is still awaiting its first header (`PeerChainTip.AwaitingFirstHeader`, `chainselection/peer_tip.go`) the recycler therefore falls back to that peer's ADVERTISED tip, but only when the advertised tip is higher and only when the advertising peer IS the connection this plateau would recycle (`advertisesForOwnConnection`, compared by `ConnectionId` rather than by rendered address, so a replacement connection reusing a listen port does not inherit a stale peer's claim). An untrusted advertised tip can therefore only ever be spent on its own connection, and the substituted value also flows into `isLedgerApplicationBacklog` as the best-peer tip, so it sets the backlog-versus-stall classification too.

The last clause is the interaction your other comment asked for tests on, so the document and the two new cases describe the same thing.

chainselection/peer_tip.go
chadaibbhot-git: Agreed, and stated in the method doc at `da8321be` rather than left to the name:

> The test is an approximation: `SelectionTip().BlockNumber == 0` is a proxy for "no header delivered", and it is also true for a peer whose delivered frontier genuinely IS block 0, and for one registered from an advertised origin tip. That is harmless at the sole call site in the chainsync recycler's plateau watchdog, which only substitutes the advertised tip when it is strictly greater than the delivered one: for an origin peer that guard is `0 > 0` and the substitution does not happen, and a node pinned at block 0 while a peer is ahead is exactly the plateau the resync exists for. A second consumer inherits the approximation rather than the property the name states, so it must either carry an equivalent guard or distinguish the cases itself.

wolf31o2 (approved):

wolf31o2 (changes requested):
This needs a merge from origin/main with the conflicts resolved

chadaibbhot-git:
Rebased onto current `main` (d63868ef) and force-pushed; the branch is now conflict-free. New head: **43586e42**. Two upstream commits landed in the files this PR touches, and both are preserved rather than reverted:

- `ARCHITECTURE.md` — textual conflict against 791cf0b7 ("keep plateau recovery prompt during catch-up"), which rewrote the same "Plateau detection" bullet. Resolved by keeping your rewritten opening (applied tip *and* downloaded primary-chain tip, primary-chain movement resetting the plateau clock) and re-appending this PR's trailing sentence about the advertised-tip fallback; no wording of either side was dropped.
- `internal/chainsyncrecycler/recycler.go` — merged cleanly. 791cf0b7 hoisted `primaryChainTipSlot` into a `checkLocalTipPlateau` parameter while this PR hoisted `targetConn` above the peer-tip comparison; the two hoists are independent, the fallback still runs before the `bestPeerTipSlot <= localTipSlot` return, and the substituted value still flows into `isLedgerApplicationBacklog`, so the documented behaviour is unchanged.
- `internal/chainsyncrecycler/recycler_test.go` — merged cleanly but needed a semantic adaptation: 791cf0b7 added `tickState.lastPrimaryChainTipSlot`, so a tick whose fake ledger reports a primary-chain tip different from the (zero-valued) tracked one now resets the plateau clock and no plateau fires. The nine plateau tests in this PR therefore seed `st.lastPrimaryChainTipSlot = ledger.primaryChainTipSlot` after `newTestTickState`, exactly as your new tests in that commit do. Without it the negative tests would have passed vacuously. The shared `newTestTickState`/`runTickWith` helpers were left untouched so your tests keep their own seeding.
- `chainselection/peer_tip.go` — merged cleanly; f3b7e553 added `observedHistoryConflictsAt` near the top of the file, `AwaitingFirstHeader` is added further down, no overlap.

Verification on the rebased tree: `go build ./...` clean; `go vet` and `golangci-lint run` clean on `chainselection` and `internal/chainsyncrecycler` (0 issues); `go test -race -count=1` green on both packages; `gofmt` clean. Both mutation checks Chris asked for were re-run and still fail closed — dropping the `advertisesForOwnConnection` condition turns `TestTickPlateauIgnoresAdvertisedTipOfAThirdPartyPeer` red, and replacing the `ConnectionId` equality with a `String()` comparison turns `TestTickPlateauIgnoresAdvertisedTipOfAReplacedConnection` red (and only that one). Each of the two commits is independently green, so the adaptation was folded into the commit that introduced the affected tests rather than added as a fixup.

CI on 43586e42: `lint`, `DCO`, `Conventional Commits`, `config-parity`, `docker`, `govulncheck`, `go-test (Linux)`, `go-test (Linux, race)` and `go-test (Windows)` are all green; cubic returned no findings. The one red is `go-test (macOS)` (run 34618526538, job 103326366500), failing `ledger.TestDecodeReadChainBatchDoesNotDeadlockOnManyValidationErrors` with `blockPipelineUnexpectedErrors` "Should be zero, but was 22". That is the flake Chris documented in #4178 (same test, same assertion, same `go-test (macOS)` job, reported there as "was 469"), and it cannot originate here: this branch's `ledger/` tree is byte-identical to `main`'s (both `0d6802e0`), the branch changes only `chainselection`, `internal/chainsyncrecycler` and `ARCHITECTURE.md`, macOS is green on `main` at d63868ef, and the test passes 10/10 locally. Leaving it to #4178 rather than re-running or pushing a no-op commit.

internal/chainsyncrecycler/recycler.go
```diff
@@ -483,10 +507,33 @@ func (r *Recycler) checkLocalTipPlateau(
...
```
chrisguiney, line 529: On this base the fallback is never reached in the post-resync case it targets. A LocalTipPlateau resync closes the connection (chainsync.go:1793). node.go:2020 RemovePeer then drops the tip and clears bestPeerConn. HandlePeerRollbackEvent (selector.go:1952) ignores the replacement's post-intersect rollback, since only a RollForward registers a peer. GetBestPeer() is nil, so the function returns at line 502. Land this after #3989, or scope the fix and the comment at 511-518 to same-connection rollbacks outside retained history.

internal/chainsyncrecycler/recycler_test.go
```diff
@@ -649,6 +651,618 @@ func TestTickResyncsOnLocalTipPlateau(t *testing.T) {
...
 
+// rollbackRegisteredPeer builds the chain-selection state a peer is left in
+// immediately after a plateau resync: the connection was closed for a fresh
+// chainsync, the peer re-intersected at the local tip and sent exactly one
+// post-intersect RollBackward, so its DELIVERED frontier is that intersection
+// point with no block number while its ADVERTISED tip is well ahead. This is
+// the shape ApplyRollback produces for any rollback point outside the retained
+// delivered-header history, and the shape #3989's rollback registration
+// records for a peer chain selection has not seen a header from at all.
+func rollbackRegisteredPeer(
```
chrisguiney, line 662: This fixture applies ApplyRollback to an already-registered ConnectionId, and fakeChainSelector returns it for any caller. The real selector never holds that state for a reconnected peer on this base, so no test covers the sequence the doc comment describes. Add a test using NewChainSelector: UpdatePeerTip(old), RemovePeer(old), HandlePeerRollbackEvent(replacement, local tip, advertised tip ahead), then one tick, asserting one LocalTipPlateau resync. On this head it publishes none; with #3989 merged it publishes one.

ARCHITECTURE.md
```diff
@@ -4070,7 +4070,7 @@ The `chainsync.State` tracks multiple concurrent chainsync clients:
...
...ainsync stall, so the healthy connection is left running and the condition is logged at INFO instead of recycling (recycling cannot advance the applied tip and only churns the connection). The "peers are ahead" comparison reads a peer's DELIVERED frontier, which a peer that has not delivered a header yet pins at the point its session intersected at — immediately after a plateau resync, the stalled local tip itself — so it would disarm this watchdog exactly when it has just acted; while a peer is still awaiting its first header (`PeerChainTip.AwaitingFirstHeader`, `chainselection/peer_tip.go`) the recycler therefore falls back to that peer's ADVERTISED tip, but only when the advertised tip is higher and only when the advertising peer IS the connection this plateau would recycle (`advertisesForOwnConnection`, compared by `ConnectionId` rather than by rendered address, so a replacement connection reusing a listen port does not inherit a stale peer's claim). An untrusted advertised tip can therefore only ever be spent on its own connection, and the substituted value also flows into `isLedgerApplicationBacklog` as the best-peer tip, so it sets the backlog-versus-stall classification too
```
chrisguiney, line 4073: The added sentence says the fallback keeps the watchdog armed immediately after a plateau resync. On main that resync closes the connection and the selector removes the peer. The replacement is only registered on its first RollForward, so the fallback is never consulted. Either land this together with rollback registration in chainselection, or limit the sentence to a same-connection rollback that lands outside the retained delivered-header history.

chrisguiney (changes requested):
The advertised-tip fallback does not run after a plateau resync on this base (inline). Earlier notes: backlog-classifier coverage is cleared, and both new cases fail with the fallback removed. The ARCHITECTURE.md sentence was added but overstates the behavior (inline). The AwaitingFirstHeader proxy is documented and has one caller. The Cubic threads stay addressed. The requested merge from main is done. go-test (macOS) fails in ledger TestDecodeReadChainBatchDoesNotDeadlockOnManyValidationErrors, which #4169 fixed on main after this base; merging main clears it.

internal/chainsyncrecycler/recycler.go
chadaibbhot-git: You are right, and I checked every step against the current head before changing anything.

1. `LocalTipPlateau` is in `chainsyncResyncRequiresFreshConnection` (`ouroboros/chainsync.go:477`), and the resync subscription takes the fresh-connection branch and calls `conn.Close()` (`ouroboros/chainsync.go:1966-1985`) rather than `restartChainsyncClientAsync`.
2. The `ConnectionClosedEvent` subscription calls `ChainSelector.RemovePeer` (`node.go:2028`), which runs `deletePeerLocked` and clears `bestPeerConn` when the removed peer was the best (`chainselection/selector.go:813`).
3. `HandlePeerRollbackEvent` (`chainselection/selector.go:2048`) only updated an existing entry, and the sole site that creates one is the roll-forward path `updatePeerTipObservedPraosView` (`selector.go:653`), reached from `HandlePeerTipUpdateEvent`.
4. So `GetBestPeer()` is nil and `checkLocalTipPlateau` returns at `recycler.go:502`, before the fallback.

(Only the line numbers drifted: the close is at `ouroboros/chainsync.go:1966`, not 1793, and `HandlePeerRollbackEvent` is at `selector.go:2048`, not 1952. The mechanisms are exactly as you describe.)

Driven against a real `NewChainSelector`, that sequence gives `PeerCount() == 0`, `GetBestPeer() == nil`, no resync — as you said.

**Your second option does not rescue it either, and I would rather say so than quietly ship it.** Scoping the fix to an already-tracked peer that rolls back outside its retained delivered-header history does keep the peer in `peerTips`, but `ApplyRollback` leaves its delivered block number at 0, and the implausibly-behind filter in `isPeerSelectableLocked` (`selectionTip.BlockNumber + K < localTip.BlockNumber`) then makes it unselectable on any chain deeper than K. Measured on a real selector with K=2160 and a local tip at block 132455: `GetBestPeer()` is nil in that case too. It survives that filter only when the local chain is within K of genesis, or K is 0.

So both halves of what this watchdog needs are in chain selection — the registration, and the `awaitingFirstHeader` exemption from those behind-filters — and both are #3989. Bypassing `GetBestPeer()` from the recycler would be fixing a chain-selection gap in the wrong layer, so I took your first option: this branch now merges #3989 rather than claiming independence from it, and the PR description says please merge #3989 first. The diff collapses back to the recycler change once #3989 lands on `main`.

The comment at 511-518 no longer asserts the outcome without its precondition. It now records the whole path — close, `RemovePeer`, post-`FindIntersect` `MsgRollBackward`, `registerPeerFromRollbackLocked`, the behind-filter exemption — and states that without those two `GetBestPeer()` is nil for the entire window, so the watchdog is not disarmed by the comparison, it has nothing to compare against at all, and that the two changes are only useful together.

Two mutation checks pin that dependency rather than leaving it as a claim: reverting the rollback registration, and reverting the behind-filter exemption, each turns the new real-selector test red on `GetBestPeer()` being nil.

internal/chainsyncrecycler/recycler_test.go
chadaibbhot-git: Agreed, and this is the comment that mattered most: the fake is exactly why I did not notice. `fakeChainSelector` returns a peer tip for any caller, so those tests prove what the watchdog does *given* an awaiting-first-header best peer and nothing about whether chain selection ever hands it one.

Added your test, on the real selector, in a new `internal/chainsyncrecycler/recycler_chainselection_test.go`:

- `TestTickResyncsOnPlateauAfterRecycleWithRealChainSelector` — `chainselection.NewChainSelector`, `UpdatePeerTip(old)` (asserting it becomes the best peer), `RemovePeer(old)` (asserting `GetBestPeer()` goes nil), `HandlePeerRollbackEvent(replacement, intersect = stalled local tip, advertised tip ahead)`, then one tick, asserting exactly one `LocalTipPlateau` resync for the replacement.

Your prediction is what the test measures. It publishes none on this head; stacked on #3989 it publishes one, so the branch now merges #3989 (reasoning in the reply on `recycler.go`).

A second case pins the other end, since a fallback that never releases is worse than one that never fires:

- `TestTickDoesNotResyncOnceTheReplacementHasDeliveredAHeader` — the same peer after it delivers a header and then rolls back to a point *inside* its retained delivered history, so `ApplyRollback` restores a real block number while the advertised tip is still ahead. The substitution must stop there, and the test asserts the advertised tip is still the higher of the two so that the awaiting-first-header precondition is the only thing that can be stopping it.

Mutation checks, each from a clean tree:

| mutation | result |
|---|---|
| remove the fallback block (`recycler.go`) | `…WithRealChainSelector`: `"[]" should have 1 item(s)` |
| revert #3989's `registerPeerFromRollbackLocked` call | `…WithRealChainSelector`: `GetBestPeer()` nil |
| revert #3989's `if !peerTip.awaitingFirstHeader` exemption | `…WithRealChainSelector`: `GetBestPeer()` nil |
| drop the `AwaitingFirstHeader()` conjunct from the fallback | `…HasDeliveredAHeader`: a resync is published |

The `rollbackRegisteredPeer` doc comment here no longer describes a sequence its own test cannot produce. It now says the fixture is hand-built, that the fake-based tests around it settle the decision logic only, and that whether chain selection ever produces such a peer is covered end to end in the new file — naming the test that goes red if either chainselection change is reverted.

ARCHITECTURE.md
chadaibbhot-git: Correct, and for the same reason — it stated the effect without the precondition that makes the effect possible.

The sentence is rewritten rather than deleted, since the fallback is real once the registration is under it. The paragraph now reads, after the `advertisesForOwnConnection` clause:

> That fallback is reachable only because chain selection tracks such a peer in the first place: a plateau resync closes the connection (`LocalTipPlateau` is in `chainsyncResyncRequiresFreshConnection`, `ouroboros/chainsync.go`), the `ConnectionClosedEvent` subscription in `node.go` calls `ChainSelector.RemovePeer`, and the replacement's only chainsync traffic until the next block is its post-`FindIntersect` `MsgRollBackward` — registered as a peer by `registerPeerFromRollbackLocked` and made selectable with a delivered block number of 0 by the `awaitingFirstHeader` exemption in `isPeerSelectableLocked` (`chainselection/selector.go`). Without that registration `GetBestPeer()` is nil through the whole window and the plateau check returns before the fallback, so the two belong together.

It also no longer says the fallback would otherwise "disarm" the watchdog immediately after a resync: without the registration there is no peer to compare against at all, which is a different failure and worth not conflating.

Since that is now a claim about code in another PR, the branch is stacked on #3989 so the document and the tree agree.

chainselection/peer_tip.go
```diff
@@ -491,6 +491,34 @@ func (p *PeerChainTip) SelectionTip() ochainsync.Tip {
...
+// it is strictly greater than the delivered one: for an origin peer that guard
+// is 0 > 0 and the substitution does not happen, and a node pinned at block 0
+// while a peer is ahead is exactly the plateau the resync exists for. A second
+// consumer inherits the approximation rather than the property the name states,
+// so it must either carry an equivalent guard or distinguish the cases itself.
+func (p *PeerChainTip) AwaitingFirstHeader() bool {
+	if p == nil {
+		return false
+	}
+	return p.SelectionTip().BlockNumber == 0
```
chrisguiney: AwaitingFirstHeader returns SelectionTip().BlockNumber == 0, but PeerChainTip already records that property exactly in the awaitingFirstHeader field (line 47), which ApplyRollback never sets. A tracked peer that delivered a header and then rolled back outside its retained delivered-header history has the field false and this method true; below K it still passes isPeerSelectableLocked, becomes best peer and takes the advertised-tip fallback. Reproduced on a real selector: one resync for a peer that had delivered a header. Read the field, or rename for the predicate computed.

internal/chainsyncrecycler/recycler_chainselection_test.go
```diff
@@ -0,0 +1,309 @@
...
+// replacement registers from its rollback, delivers a header at the stalled
+// local tip, then rolls back to a point inside its own retained delivered
+// history -- so its delivered frontier carries a real block number again while
+// its advertised tip is still far ahead. That is an ordinary peer serving our
+// own chain, not a starved watchdog, and the untrusted advertisement must no
+// longer be able to drive a recycle.
+//
+// Red if AwaitingFirstHeader is forced true, which is the mutation that widens
+// the fallback past its own precondition.
+func TestTickDoesNotResyncOnceTheReplacementHasDeliveredAHeader(t *testing.T) {
```
chrisguiney, line 236: The doc comment and the closing assertion state that a peer which has delivered a header is judged on its delivered frontier, never on its advertisement. That holds only for a rollback INSIDE retained history, which is the case built here. For a rollback outside it, ApplyRollback zeroes the delivered block number, AwaitingFirstHeader reads true again, and on a chain shallower than K the substitution fires for that peer and publishes a resync. Add that case so the boundary this fix claims is pinned in both directions.

internal/chainsyncrecycler/recycler.go
```diff
@@ -483,10 +507,49 @@ func (r *Recycler) checkLocalTipPlateau(
...
```
chrisguiney, line 552: Prior blocker cleared. #3989 is on main, so registerPeerFromRollbackLocked and the isPeerSelectableLocked exemption are both present and GetBestPeer is no longer nil through the post-resync window. Removing this whole block in place turns TestTickResyncsOnPlateauAfterRecycleWithRealChainSelector, TestTickResyncsOnPlateauWhenBestPeerAwaitsFirstHeader, TestTickPlateauUsesAdvertisedTipWithNoActiveClient and two backlog cases red, and each of the three conjuncts is independently load-bearing under the same in-place check.

internal/chainsyncrecycler/recycler_chainselection_test.go
```diff
@@ -0,0 +1,309 @@
...
+// case #4139 claims to fix, driven through a real ChainSelector: a plateau
+// resync closed the only upstream, the peer was removed on the
+// ConnectionClosedEvent, the replacement reconnected and has sent nothing but
+// its post-intersect rollback at the stalled local tip while advertising a tip
+// well ahead. The watchdog must still fire on the next plateau window.
+//
+// Red without the fallback (the delivered frontier is the stalled local tip, so
+// the plateau comparison sees no peer ahead), and red without #3989 underneath
+// (the replacement is not tracked at all, so GetBestPeer is nil).
+func TestTickResyncsOnPlateauAfterRecycleWithRealChainSelector(t *testing.T) {
```
chrisguiney, line 108: Prior finding cleared. This drives a real NewChainSelector through HandlePeerRollbackEvent, which is the method the PeerRollbackEventType subscription calls (selector.go:338) and carries the event ouroboros/chainsync.go:1201 publishes, so it exercises the production path rather than a fake that answers for any caller. It goes red when the fallback block is removed in place, which is the causal evidence the earlier fake-based cases could not give.

ARCHITECTURE.md
```diff
@@ -4374,7 +4374,7 @@ The `chainsync.State` tracks multiple concurrent chainsync clients:
...
...rtisesForOwnConnection`, compared by `ConnectionId` rather than by rendered address, so a replacement connection reusing a listen port does not inherit a stale peer's claim). An untrusted advertised tip can therefore only ever be spent on its own connection, and the substituted value also flows into `isLedgerApplicationBacklog` as the best-peer tip, so it sets the backlog-versus-stall classification too. That fallback is reachable only because chain selection tracks such a peer in the first place: a plateau resync closes the connection (`LocalTipPlateau` is in `chainsyncResyncRequiresFreshConnection`, `ouroboros/chainsync.go`), the `ConnectionClosedEvent` subscription in `node.go` calls `ChainSelector.RemovePeer`, and the replacement's only chainsync traffic until the next block is its post-`FindIntersect` `MsgRollBackward` — registered as a peer by `registerPeerFromRollbackLocked` and made selectable with a delivered block number of 0 by the `awaitingFirstHeader` exemption in `isPeerSelectableLocked` (`chainselection/selector.go`). Without that registration `GetBestPeer()` is nil through the whole window and the plateau check returns before the fallback, so the two belong together
```
chrisguiney: Prior finding cleared; each clause checks out against the code. LocalTipPlateau is in chainsyncResyncRequiresFreshConnection (ouroboros/chainsync.go:477) and that branch calls conn.Close(); node.go:2047 removes the peer on ConnectionClosedEvent; ouroboros/chainsync.go:1201 publishes the post-intersect rollback; registerPeerFromRollbackLocked and the isPeerSelectableLocked exemption are on main. docs-parity and import-boundaries both pass. The qualifier still to add is the delivered-block-number case raised on peer_tip.go.

internal/chainsyncrecycler/recycler.go
```diff
@@ -347,6 +347,30 @@ func shouldRecycleLocalTipPlateau(
...
+// The comparison is ConnectionId equality, which is the identity relation
+// chainselection.peerTips and chainsync.clients use as their map key, and which
+// gouroboros makes stable by populating Connection.id once per connection. It
+// is deliberately NOT the rendered address string the recycler uses to key its
+// cooldown map: a replacement connection to the same peer (listen-port reuse)
+// renders identically while being a distinct connection, and matching on the
+// rendering would let a stale peer tip authorize recycling the replacement.
+// Collapsing those onto one key is right for a rate limit and wrong for an
+// authorization.
+func advertisesForOwnConnection(
```
chrisguiney, line 364: Both earlier review threads on this file stay addressed at this head: the substitution is still confined to the connection the plateau would recycle, and still compares ConnectionId values rather than rendered addresses. Substituting a String() comparison turns TestTickPlateauIgnoresAdvertisedTipOfAReplacedConnection red, and replacing this conjunct with true turns TestTickPlateauIgnoresAdvertisedTipOfAThirdPartyPeer red. Checks at this head are green, including go-test on Linux, Linux race, macOS and Windows.

chrisguiney (dismissed):
Every prior blocker is cleared at this head and verified causally; details inline. One new finding: AwaitingFirstHeader is true for a peer that has delivered a header, so the stated precondition of the fallback does not hold at its only call site, and the boundary test pins only half of it. The description still asks for #3989 to be merged first, which has already happened, and it carries a review exchange and a command ledger that a description should not.

chainselection/peer_tip.go
chadaibbhot-git: Valid, and fixed in 953fc754 by reading the field.

Your chain holds at `a477530d`, step for step:

- `AwaitingFirstHeader` returned `p.SelectionTip().BlockNumber == 0` (`chainselection/peer_tip.go:519`), while `awaitingFirstHeader` (`peer_tip.go:47`) is set only by `newPeerChainTipFromRollback` (`:112`) and cleared only by `UpdateTipWithObservedPraosView` (`:156`);
- `ApplyRollback`'s not-found branch (`:194-199`) keeps the point, leaves the block number at 0 and never touches the field;
- `isPeerSelectableLocked` exempts only the field (`selector.go:1240`), so that peer is subject to the implausibly-behind filter, which it passes whenever the local block number is at most K;
- the recycler then substitutes the advertised tip (`recycler.go:545`) for a peer that has delivered headers on this connection.

Reproduced on a real selector before touching anything: one `LocalTipPlateau` resync for that peer.

**Why the field, not a rename.** The predicate the watchdog needs is the one the name states: "the delivered frontier is only the intersection this session proposed." A peer that has delivered headers and then rolls back outside its retained history has pointed at a point on its own chain. That is the same kind of evidence as any other delivered frontier, and the PR already says it is used as-is once a header has arrived. Renaming the method to match the zero-block-number test would widen the fallback to peers it was never meant for. Reading the field also removes the origin approximation the old doc comment had to explain away. `AwaitingFirstHeader` now reads `p.awaitingFirstHeader`, and the doc comment says why it is not a block-number test.

**A consequence in the tests you'd otherwise flag next.** The fake-based fixture `rollbackRegisteredPeer` (`recycler_test.go`) built its "awaiting" peer with `NewPeerChainTip` + `ApplyRollback` outside history, which is your counterexample shape. With the field read, all eight fake-based fallback tests went red for that reason alone. The fixture now gets the peer from a real `NewChainSelector`'s rollback registration (`HandlePeerRollbackEvent` → `GetPeerTip`), the only path that sets the field. They are green again with no change to their assertions.

Tests (all red on `a477530d`'s accessor, green at 953fc754):

- `chainselection/TestAwaitingFirstHeaderFalseAfterRollbackOutsideDeliveredHistory` (new): a delivered peer rolls back outside history, block number 0, accessor false.
- `chainselection/TestAwaitingFirstHeaderClearsOnDeliveredHeader` (extended): starts from `newPeerChainTipFromRollback` and asserts that a later outside-history rollback does not make the peer await a header again.
- `chainsyncrecycler/TestTickDoesNotResyncAfterADeliveredPeerRollsBackOutsideItsHistory` (new, real selector): see the thread on `recycler_chainselection_test.go`.
- `TestAwaitingFirstHeaderAfterPostIntersectRollback` now builds its peer with `newPeerChainTipFromRollback`. The old version built the counterexample and asserted `true`.

Mutations of the accessor, each run from a clean tree:

| accessor body | result |
|---|---|
| `SelectionTip().BlockNumber == 0` (the old code) | the three tests above red |
| `false` | `…AfterPostIntersectRollback`, `…ClearsOnDeliveredHeader`, `TestTickResyncsOnPlateauAfterRecycleWithRealChainSelector` and all eight fake-based fallback tests red |
| `true` | both chainselection negatives, `…HasDeliveredAHeader`, `…OutsideItsHistory`, `TestTickIgnoresAdvertisedTipOnceBestPeerHasDeliveredAHeader`, `TestTickSuppressesResyncOnLedgerApplicationBacklog` red |

Not done: I did not make `ApplyRollback` set the field for an outside-history rollback. That would change the behind-filter exemption, the Genesis exit horizon and the plausibility reference for every such peer. Those are chain-selection decisions #3989 scoped to rollback-registered entries, and this PR has no business widening them.

internal/chainsyncrecycler/recycler_chainselection_test.go
chadaibbhot-git: Valid. The doc comment and the closing assertion claimed the whole boundary but built only the inside-history half. Fixed in 953fc754.

New: `TestTickDoesNotResyncAfterADeliveredPeerRollsBackOutsideItsHistory`, on a real `NewChainSelector`. The replacement registers from its post-intersect rollback, delivers a header at the local tip, then rolls back to an earlier point it never delivered to us, so the point is outside its retained history. K is 2160 and the local block is 1000, so the zero block number still passes `isPeerSelectableLocked`. The test asserts that the preconditions actually hold before the tick: the peer is `GetBestPeer()`, its `SelectionTip()` has block number 0 at the rollback slot, and its advertised tip is ahead of the local tip. Then it asserts that no resync is published and `AwaitingFirstHeader()` is false.

- Red at `a477530d`: one `local_tip_plateau` resync published, and the accessor returns true.
- Green once `AwaitingFirstHeader` reads the field (details on the `peer_tip.go` thread).
- Also red if the recycler's `AwaitingFirstHeader()` conjunct is replaced with `true`, together with `…HasDeliveredAHeader`.

The existing test's doc comment now says it is one half of the boundary and names the other test, and its closing message is scoped to "a rollback inside its retained history". The file header's "stacked on #3989" / "On main" wording is also updated now that #3989 is on main.

The comment explains why the chain is shallower than K. On a deeper chain, this peer is filtered out, `GetBestPeer()` is nil and the plateau check returns before the fallback whatever the accessor says, so that variant could not tell the two predicates apart.

ARCHITECTURE.md
chadaibbhot-git: Added in 953fc754. The plateau paragraph now says that "awaiting its first header" means the `awaitingFirstHeader` flag, which chain selection sets only when it registers a peer from a rollback and clears on that peer's first delivered header. It is not a delivered block number of 0. A tracked peer that has delivered headers and then rolls back outside its retained history is also left at block number 0 (`ApplyRollback` keeps the point and cannot recover one), but it is judged on its delivered frontier and its advertised tip is never substituted. The same qualifier is in the comment above the fallback in `recycler.go`, and the `AwaitingFirstHeader` doc comment now says why it is not a block-number test. `go test ./internal/docsparity` passes at this head.

internal/chainsyncrecycler/recycler.go
chadaibbhot-git: Thanks. At 953fc754 (after merging `upstream/main`, with #3989 present directly) the block and its three conjuncts are unchanged. One thing the tests you named depend on did change: `rollbackRegisteredPeer` now gets its peer from a real selector's rollback registration instead of `NewPeerChainTip` + `ApplyRollback`, because `AwaitingFirstHeader` now reads the flag (see the `peer_tip.go` thread). Removing this block still turns `TestTickResyncsOnPlateauAfterRecycleWithRealChainSelector`, `TestTickResyncsOnPlateauWhenBestPeerAwaitsFirstHeader`, `TestTickPlateauUsesAdvertisedTipWithNoActiveClient` and the two backlog cases (`…FeedsTheBacklogClassifier`, `…ResyncsWhenHeaderGapDominates`) red.

internal/chainsyncrecycler/recycler_chainselection_test.go
chadaibbhot-git: Thanks. This test is unchanged at 953fc754 apart from its header comment, which now says that #3989 is on main rather than that the branch is stacked on it. The new outside-history case in the same file uses the same production path.

internal/chainsyncrecycler/recycler.go
chadaibbhot-git: Thanks. Nothing on this line changed at 953fc754, and both of those tests are still green there, now with their awaiting peer built through a real selector's rollback registration.

wolf31o2 (approved):

Issue #4146, opened 2026-09-09
test(ledger): scheduler and slot-clock tests assert on hard real-time windows


chrisguiney:
Two `ledger` tests gate their assertions on wall-clock windows that are short relative to CI runner scheduling, so they fail intermittently when the package is run under load. Both are independent of the event-bus delivery race in #4145 and predate it.

## `TestScheduler_ChangeInterval` — `ledger/timer_test.go:49`

After `ChangeInterval(200 * time.Millisecond)` the test sleeps 500ms and requires the tick count in that window to land in `[1, 3]`:

```go
time.Sleep(500 * time.Millisecond)
...
if afterChange < 1 || afterChange > 3 {
    t.Errorf("timer did not respect interval change, ran too frequently: %d more ticks", afterChange)
}
```

The window admits 2.5 nominal ticks, so a single scheduling delay that bunches ticks, or one extra tick delivered from the pre-change 50ms interval, exceeds the upper bound. The test is `t.Parallel()`, so the contention that causes the bunching is normal for the package.

## `TestSlotClockEpochBoundary` — `ledger/slot_clock_test.go:456`

The epoch-boundary tick must arrive inside a fixed 500ms deadline:

```go
timeout := time.After(500 * time.Millisecond)
...
case <-timeout:
    t.Fatal("timeout waiting for epoch boundary tick")
```

A delayed slot-clock goroutine misses the deadline with no retry.

## Suggested direction

Assert the observable property rather than the real-time budget: for the scheduler, drive ticks from an injectable clock or assert monotonic progress with a generous upper bound; for the slot clock, wait on the boundary tick without a fixed real-time deadline shorter than the test timeout.

Same class as #2265 (`time.Sleep` synchronization in mempool and ledger tests). Not the same defect as #4145, which is an ordered-lane delivery race in the event bus.

PR #4147, opened 2026-09-09
test(chainselection,ledger): fix frontier-flap event race

### LINKED ISSUE #4145: flake: frontier-flap tests fail under full-package parallelism, go-test red on main since #4048

chrisguiney:
`go-test` on `main` has been red since #4048 added two frontier-flap
regression tests. Both fail with zero events observed:

| Test | Package | Failure |
| --- | --- | --- |
| `TestCanonicalPeersWithCrossingFrontiersDoNotFlap` | `chainselection` | `"[]" should have 1 item(s), but has 0` |
| `TestCanonicalFrontierCrossingDoesNotCloseAPeerAheadOfLocalTip` | `ledger` | `"0" is not positive` |

## Defect

Both tests drained their `EventBus` subscriber channel with a non-blocking
`select`/`default` immediately after the driving calls returned, documented as
"the selector publishes synchronously". It does not.
`ChainSelector.publishSelection` (`chainselection/selector.go:1527`) publishes
through `EventBus.PublishOrdered`, which only enqueues onto the event type's
lane before returning (`event/ordered.go:103`); `event.orderedWorker` delivers
to subscribers later on its own schedule. The non-blocking read raced that
goroutine and saw nothing whenever it lost. #4131's package-level parallelism
widened the gap between "enqueued" and "delivered" enough to make losing the
common case.

## Change

Test-only, confined to the two test files. After the driving calls, publish a
sentinel `ChainSwitchEvent` through the same ordered lane, then block-receive
until it arrives, bounded by `event.RemoteDeliverTimeout`. Everything received
before the sentinel is the scenario's real output.

Receiving the sentinel is proof rather than a wait because the lane is strictly
serial:

- `event/ordered.go:136` — a lane is created once per event type under
  `orderedMu` and starts exactly one `orderedWorker` goroutine.
- `event/ordered.go:158` — that worker dequeues serially and calls
  `EventBus.Publish` inline.
- `event/event.go:1179` — `Publish` hands the event to every subscriber before
  returning.

An event enqueued before the sentinel therefore reaches the subscriber channel
before it. No `t.Parallel()` is removed, and there is no sleep, retry, skip or
timeout increase.

## #4048's production logic is not at fault

Forcing `sameCanonicalChain` to return `false` reverts both of #4048's
behavior changes, since it is the sole gate on each (`isPeerSelectableLocked`
and `pinIncumbentDuringCatchUpLocked`). Against that reverted selector the
updated tests fail on their selection assertions — `active peer must not flap
on delivered-frontier lead` in `chainselection`, and `must not close the
selected connection` for five separate switch events in `ledger`. The
handshake synchronises the tests without making them vacuous.

The observed pre-fix failures are only ever the event-count assertions. The
per-step selection assertions read selector state synchronously in the test
goroutine and cannot be affected by delivery timing.

## Measured pass rates

Runs that passed, at `-cpu=1` to make the race deterministic:

| Scope | Pre-fix | Post-fix |
| --- | --- | --- |
| `chainselection`, target test, 20 runs | 0/20 | 20/20 |
| `ledger`, target test, 20 runs | 0/20 | 20/20 |
| `chainselection` full package, `-race`, two sweeps of 20 | 6/20, 5/20 | 20/20 |
| `ledger` full package, `-race`, 13 runs | 0/3 | 13/13 |

`ledger` keeps two unrelated real-time-window flakes, `TestScheduler_ChangeInterval`
and `TestSlotClockEpochBoundary`, tracked in #4146. They fire under a
`-cpu=1 -race` full-package run and are the reason the `ledger` column above is
scoped to the target test.

Closes #4145.

Red `go-test` from this failure alone: #3855, #3853, #3938, #3982, #3838, #3929.

---
## Summary by cubic
Fixes a race in the `chainselection` and `ledger` frontier-flap regression tests that has kept `go-test` red since #4048 (closes #4145). The tests drained the `EventBus` subscriber channel with non-blocking reads, but `ChainSelector` publishes through `PublishOrdered`, which only enqueues and delivers asynchronously, so the tests could observe zero events even though the selector had decided to publish one.

Replaces the non-blocking drain with a handshake: after the driving calls return, publish a sentinel `ChainSwitchEvent` through the same ordered lane and block until it arrives. Because the lane has a single FIFO worker, receiving the sentinel proves every earlier event has been delivered. The change is test-only; production logic and `t.Parallel()` remain untouched, and the target tests now pass 20/20 at `-cpu=1`.

<sup>Written for commit 10bdec6499cd79efe7aa95bed1184d33d0269d52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Tests**
  - Improved chain-switch event test reliability by synchronizing event delivery before validation.
  - Replaced non-blocking event draining with timeout-based receives and ordered sentinel checks.
  - Reduced flaky failures during parallel test execution.

### DIFF
--- a/chainselection/frontier_flap_test.go
+++ b/chainselection/frontier_flap_test.go
@@ -91,8 +91,8 @@ var chainSwitchDrainSentinelConnId = newTestConnectionId(50000)
 // subscriber channel immediately after the driving calls return can therefore
 // observe zero events even though the selector already decided to publish
 // one: nothing has forced the delivery goroutine to run yet. That assumption
-// -- reading the channel is exactly what made this scenario's regression test
-// flake under full-package parallelism, where CPU contention widens the gap
+// is exactly what made this scenario's regression test flake under
+// full-package parallelism, where CPU contention widens the gap
 // between "enqueued" and "delivered" (blinklabs-io/dingo#4145).
 //
 // The fix is a handshake, not a wait: publish a sentinel ChainSwitchEvent
--- a/chainselection/frontier_flap_test.go
+++ b/chainselection/frontier_flap_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/blinklabs-io/dingo/event"
+	"github.com/blinklabs-io/dingo/internal/test/testutil"
 )
 
 // These tests pin the transport-frontier flapping reproduction from Preview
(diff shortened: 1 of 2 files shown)


chrisguiney:
closing in favor of #4144

PR #4148, opened 2026-09-09
docs(event): state when a published event has been delivered

chrisguiney:
`PublishOrdered` returns once the event is enqueued on its per-type lane; the
lane worker calls `Publish` afterwards, from its own goroutine. `Publish`
delivers inline, handing the event to every subscriber before it returns.
Neither method's doc comment said which side of that split it was on.

The distinction was recorded only in test helpers — `switchBarrier`
(`ouroboros/consensus_conformance_test.go`), `chainSwitchBarrier`
(`chainselection/frontier_flap_test.go`,
`ledger/chainsync_frontier_flap_test.go`) and `drainChainSwitchesUntilBest`
(`chainselection/selector_test.go`) — and in none of the API they describe.
The frontier-flap tests added in #4048 were written against the opposite
assumption, that the selector publishes synchronously, so a non-blocking drain
read an enqueued-but-undelivered event as no event; #4144 replaced both drains
with a barrier. Related to that work; closes nothing.

The comments now state:

- `PublishOrdered` returns once the event is on the lane. A true return means
  accepted for delivery, and says nothing about whether the worker has
  delivered it yet.
- `Publish` delivers on the caller's goroutine, returning once each subscriber
  has taken the event or been unsubscribed for failing to, and a stopped or
  closed bus delivers nothing. `PublishBlocking` is the other inline path.
- A non-blocking subscriber-channel read (`select`/`default`, `len(ch)`, a
  drain that stops at the first empty read) cannot observe an ordered publish.
  A barrier event through the same lane can, because the lane is a FIFO drained
  by exactly one worker. The barrier's `Data` type must be one every subscriber
  on that lane skips.

Doc comments only; no executable change.

https://claude.ai/code/session_015ggZFQYmK8vLYepaQcssBP

---
## Summary by cubic
Clarifies the delivery semantics of `Publish` and `PublishOrdered` in their doc comments. `PublishOrdered` only enqueues the event; delivery happens later from a worker goroutine, while `Publish` delivers inline before returning. Also warns that non-blocking reads of subscriber channels can't observe an ordered publish and suggests the barrier pattern for tests that need to confirm delivery.

<sup>Written for commit b123065d99f243d1a43e4a74636e6016e4b59bbe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -1175,7 +1175,17 @@ func (e *EventBus) deliverWithTimeout(
 	}
 }
 
-// Publish allows a producer to send an event of a particular type to all subscribers
+// Publish allows a producer to send an event of a particular type to all
+// subscribers.
+//
+// Delivery is inline: Publish hands the event to every subscriber on the
+// caller's goroutine, returning once each has taken it or been unsubscribed
+// for failing to. A stopped or closed bus delivers nothing.
+//
+// PublishOrdered and PublishAsync are the opposite: they return once the event
+// is queued and deliver it from another goroutine afterwards. A caller that
+// reads a subscriber channel expecting the event to already be there can rely
+// on that only with the inline paths, Publish and PublishBlocking.
 func (e *EventBus) Publish(eventType EventType, evt Event) {
 	e.stopMu.RLock()
 	if e.stopped || e.closed {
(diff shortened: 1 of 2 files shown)


wolf31o2 (approved):
Accept

PR #4150, opened 2026-09-09
fix(config): preserve redaction value boundaries

wolf31o2:
Configuration logging treated quotes in URI query values as keyword-DSN
delimiters, leaving a credential suffix visible or consuming the next query
parameter. A question mark inside a keyword-DSN password could also select
query parsing and leave the password visible.

- Make quote handling explicit per parameter syntax: literal in URI queries,
  delimited in keyword DSNs.
- Sanitize assignment-shaped inputs as keyword DSNs before scanning remaining
  query parameters, retaining redaction for ambiguous relative URLs.
- Keep the whole credential value secret when its suffix could be either a
  relative-URL query or part of an unquoted DSN password. Preserving apparent
  benign parameters in that ambiguous suffix could expose password content.
- Cover quoted and unquoted values, benign parameter boundaries, and text/JSON
  configuration output; document the syntax precedence.

Related to #3667; this does not add a non-URL registry authentication mechanism.

### DIFF
--- a/internal/config/redact.go
+++ b/internal/config/redact.go
@@ -642,6 +642,15 @@ func redactCredentialParams(s string) string {
 	// If the assignment itself is credential-named, its entire DSN value
 	// stays secret; a '?' suffix cannot safely be assumed to be a URL query.
 	if startsParam(s, skipParamSpace(s, 0), keywordDSNSyntax) {
+		// A relative URL can begin with an assignment-shaped path. When its
+		// suffix has URI separators and no whitespace, keep that suffix as a
+		// query instead of treating it as part of the DSN value.
+		if query := strings.IndexByte(s, '?'); query > 0 &&
+			strings.Contains(s[query+1:], "&") &&
+			!strings.ContainsAny(s[query+1:], " \t\r\n") {
+			return redactParams(s[:query], keywordDSNSyntax) +
+				"?" + redactParams(s[query+1:], uriQuerySyntax)
+		}
 		s = redactParams(s, keywordDSNSyntax)
 	}
 	if start, end, ok := uriQuerySpan(s); ok {
--- a/internal/config/redact.go
+++ b/internal/config/redact.go
@@ -615,6 +615,8 @@ func sortedStringKeys(v reflect.Value) []string {
 // operational value of knowing which host and database the node was
 // pointed at, which is most of the reason the configuration is logged at
 // all.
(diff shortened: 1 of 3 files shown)


internal/config/redact.go
```diff
@@ -633,6 +635,15 @@ func redactURICredentials(s string) string {
 // has no query string: its pairs are whitespace separated, whitespace
 // around '=' is optional, and a value containing whitespace is quoted.
 func redactCredentialParams(s string) string {
+	// net/url also accepts keyword DSNs as relative URLs. Sanitize their
+	// values first so a '?' inside a password cannot hide its prefix.
+	// Still scan any remaining query: a relative URI can also start with
+	// "name=value", so that shape alone must not disable query redaction.
+	// If the assignment itself is credential-named, its entire DSN value
+	// stays secret; a '?' suffix cannot safely be assumed to be a URL query.
+	if startsParam(s, skipParamSpace(s, 0), keywordDSNSyntax) {
```
arepala-uml, line 644: Re: the Cubic P2 finding on this line, annotated "Addressed in a605961" — that annotation is stale relative to this head. `a6059612` did split the ambiguous suffix (redacting the DSN-shaped prefix and separately redacting only `apiKey` in the trailing query, preserving `network=preview` as plaintext). The final commit on this branch, `a930ef70` ("fix: keep ambiguous DSN suffixes secret"), deliberately reverts that split back to whole-suffix redaction, and adds `TestRedactKeywordDSNMatchesPgxValueBoundary`, which parses the identical string with `pgconn.ParseConfig` and confirms the real DSN parser reads `secret?network=preview&apiKey=key` as one literal password. So the split in `a6059612` would have printed `network=preview` as if it were a benign parameter when it is actually part of the password — a real leak. Current behavior (redact the whole ambiguous suffix) is the safer, correct choice, verified against the actual third-party parser rather than only against self-referential expectations. No action needed; flagging only so the thread's "Addressed" marker isn't read as describing current head behavior.

Independently verified at head a930ef70 (isolated worktree from origin/main): `go build ./...`, `go vet ./...`, `gofmt -l .` (clean for this PR's files; the repo's one pre-existing gofmt finding, chain/chain_test.go, is untouched by this diff), `golangci-lint run ./internal/config/...` (0 issues), and `go test ./internal/config/... -v` all pass. CI is green across the board (go-test Linux/Windows/macOS/race, lint, govulncheck, CodeQL, docker).

arepala-uml (approved):

Issue #2265, opened 2026-05-13
test: replace time.Sleep synchronization in mempool and ledger tests


wolf31o2:
## Summary

Several tests still use `time.Sleep()` for synchronization.

## Audit Details

Finding IDs: H16, L23, L24, L25.

Affected files:

- `mempool/mempool_test.go`
- `connmanager/listener_test.go`
- `ledger/state_test.go`
- `ledger/timer_test.go`

The audit called out more than 20 sleeps in mempool tests and additional sleeps in connmanager/ledger tests. Project rules say synchronization should use `internal/test/testutil/` helpers, channels, or timeout-bound contexts rather than raw sleeps.

Impact: tests can become flaky or slow under CI load, especially on slower runners.

## What Needs to Happen

1. Classify each remaining sleep as pacing vs synchronization.
2. Replace synchronization sleeps with:
   - `testutil.WaitForCondition`
   - `testutil.RequireReceive`
   - `context.WithTimeout`
   - channel acknowledgements from the goroutine under test
3. Keep deliberate pacing sleeps only when there is no synchronization dependency and document why.
4. Run touched packages with `-race`.

## Acceptance Criteria

- Synchronization sleeps are removed from the listed tests.
- Any remaining sleep is documented as deliberate pacing.

PR #4153, opened 2026-09-09
test(ledger): assert scheduler/slot-clock ticks, not real-time windows

### LINKED ISSUE #4146: test(ledger): scheduler and slot-clock tests assert on hard real-time windows

chrisguiney:
Two `ledger` tests gated their assertions on wall-clock windows short relative to CI scheduling, so they failed intermittently under full-package load.

- **`TestScheduler_ChangeInterval`** counted ticks in a fixed 500ms sleep window and required the count to land in `[1, 3]` — 2.5 nominal ticks admitted against a bound of 3. It also assumed a single `ChangeInterval` call takes effect: `updateIntervalChan` is unbuffered and the send is non-blocking, so `run()` can be mid-tick and drop it silently. The test now retries until the scheduler's own `interval` field confirms the change, asserts the tick count never reaches a threshold reachable only at the old interval, and proves ticking continues by registering a second task after the change.
- **`TestSlotClockEpochBoundary`** required the boundary tick inside a fixed 500ms deadline with no retry. The wait is now a deadlock guard rather than a latency budget: 10s admits 50 epoch boundaries at this test's 20ms slot length and 10-slot epoch, clamped to the test binary's own `-timeout` deadline when that is nearer, so a clock that stops ticking fails as a `t.Fatal` rather than consuming the package's timeout budget (#3032).
- **`TestSlotClockReceivesTicks`** is the same fixed-window class in the same file — a 200ms deadline is four nominal slot periods — and now uses the file's existing bounded-receive helper at 10s.

The replacements are not weaker assertions: restoring each previous form and driving it under `GOMAXPROCS=1` contention reproduces the original failures with the messages the issue names.

The scheduler's closing guard cannot be anchored on the first task's counter at any reading. `tick()` enqueues onto `taskQueue` and the worker pool drains it asynchronously, so a closure enqueued by the last pre-change tick can increment that counter after the reading is taken, satisfying a bare lower bound on a scheduler that stopped ticking outright at the change. The guard registers a second task once the change is confirmed instead: `Register` appends under the same mutex `tick()` holds, and `run()` discards the old ticker before publishing `st.interval`, so every execution of that task is driven by a tick the new ticker delivered.

Test files only; no production change.

Closes #4146

Related to #2265, which tracks the same class across the mempool and ledger tests.

https://claude.ai/code/session_015ggZFQYmK8vLYepaQcssBP

### DIFF
--- a/ledger/timer_test.go
+++ b/ledger/timer_test.go
@@ -69,8 +69,6 @@ func TestScheduler_ChangeInterval(t *testing.T) {
 	}, 2*time.Second, 5*time.Millisecond,
 		"expected at least 2 executions before interval change",
 	)
-	beforeChange := counter.Load()
-
 	// updateIntervalChan is unbuffered and ChangeInterval's send is
 	// non-blocking (select/default; see
 	// TestScheduler_StopDoesNotRaceChangeInterval) -- run() can be mid-tick
@@ -115,7 +113,13 @@ func TestScheduler_ChangeInterval(t *testing.T) {
 		"timer did not respect interval change: ran too frequently",
 	)
 
-	require.Greater(t, counter.Load(), beforeChange,
+	// Anchor the lower bound at the post-change reading. Comparing
+	// against a pre-change reading is satisfied by ticks the old interval
+	// delivered while the change was being confirmed, so it passes even
+	// when ticking stopped outright at the change.
+	require.Eventually(t, func() bool {
+		return counter.Load() > afterChangeReq
+	}, 10*time.Second, 10*time.Millisecond,
 		"expected ticking to continue after interval change",
 	)
 }
--- a/ledger/timer_test.go
+++ b/ledger/timer_test.go
@@ -113,12 +113,24 @@ func TestScheduler_ChangeInterval(t *testing.T) {
 		"timer did not respect interval change: ran too frequently",
 	)
 
-	// Anchor the lower bound at the post-change reading. Comparing
-	// against a pre-change reading is satisfied by ticks the old interval
-	// delivered while the change was being confirmed, so it passes even
-	// when ticking stopped outright at the change.
+	// A lower bound on counter is not attributable to the new interval at
+	// any anchor. tick() enqueues onto taskQueue and the worker pool
+	// drains it asynchronously, so a closure enqueued by the last
+	// pre-change tick can increment counter after the reading above was
+	// taken -- satisfying "counter grew" on a scheduler that stopped
+	// ticking outright at the change.
+	//
+	// Register a second task instead. Register appends under the same
+	// mutex tick() holds, and run() stopped and discarded the old ticker
+	// before publishing st.interval, so no tick predating the confirmed
+	// change can reach this task: every execution of it is driven by a
+	// tick the new ticker delivered.
+	var postChange atomic.Int32
+	timer.Register(1, func() {
+		postChange.Add(1)
+	}, nil)
 	require.Eventually(t, func() bool {
-		return counter.Load() > afterChangeReq
+		return postChange.Load() > 0
 	}, 10*time.Second, 10*time.Millisecond,
 		"expected ticking to continue after interval change",
 	)
--- a/ledger/timer_test.go
+++ b/ledger/timer_test.go
@@ -51,12 +51,14 @@ func TestScheduler_ChangeInterval(t *testing.T) {
 
 	var counter atomic.Int32
 
-	// Create a Scheduler with 50ms tick interval
-	timer := NewScheduler(50 * time.Millisecond)
+	// Create a Scheduler with a fast tick interval so the pre-change
+	// baseline is established quickly.
+	const fastInterval = 20 * time.Millisecond
+	timer := NewScheduler(fastInterval)
 	timer.Start()
 	defer timer.Stop()
 
-	// Registering task with 50ms tick interval to execute for every 1 tick
+	// Registering task to execute for every 1 tick
 	timer.Register(1, func() {
 		counter.Add(1)
 	}, nil)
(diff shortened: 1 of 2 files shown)


ledger/timer_test.go
```diff
@@ -51,47 +51,77 @@ func TestScheduler_ChangeInterval(t *testing.T) {
...
-			"timer did not respect interval change, ran too frequently: %d more ticks",
-			afterChange,
-		)
-	}
+	// Anchor the lower bound at the post-change reading. Comparing
+	// against a pre-change reading is satisfied by ticks the old interval
+	// delivered while the change was being confirmed, so it passes even
+	// when ticking stopped outright at the change.
+	require.Eventually(t, func() bool {
+		return counter.Load() > afterChangeReq
```
wolf31o2: changes requested: this lower-bound check can be satisfied by pre-change tasks already queued while the interval update is observed, because taskQueue drains asynchronously. The test can therefore pass even if ticking stops at the change. Establish the baseline only after the pre-change queue is drained, then require a tick attributable to the new interval.

wolf31o2 (changes requested):
changes requested

ledger/timer_test.go
chrisguiney: Confirmed, and fixed.

`tick()` enqueues onto `taskQueue` under `st.mutex` and the worker pool drains it separately. `ScheduledTask.mutex` bounds that to one closure in flight, so exactly one pre-change increment can arrive after `afterChangeReq` is read — which is enough to satisfy `counter.Load() > afterChangeReq` with zero post-change ticks.

The lower bound no longer reads `counter`. A second task is registered once the interval change is confirmed. `Register` appends under the same mutex `tick()` holds, and `run()` stops and discards the old ticker before publishing `st.interval`, so no tick predating the confirmed change can reach that task: every execution of it is driven by the new ticker.

Fail-before, against `ledger/timer.go`: `run()`'s update case rewritten to `st.ticker = time.NewTicker(time.Hour)` while still publishing `st.interval` — a scheduler that reports the change and stops ticking — with `worker()` sleeping 18ms before `task()`, the asynchronous drain running just under one tick period behind. Under that pair the previous assertion passed 20/20; the replacement fails 20/20. With the worker delay alone and ticking intact, the replacement passes 20/20.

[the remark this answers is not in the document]
chrisguiney: Valid. The lower bound no longer reads `counter`, which `tick()`'s enqueue-then-drain path can satisfy with a single pre-change closure. A second task is registered once the interval change is confirmed; `Register` appends under the same mutex `tick()` holds, and `run()` discards the old ticker before publishing `st.interval`, so every execution of that task is driven by a tick the new ticker delivered.

Fail-before, against `ledger/timer.go`: `run()`'s update case rewritten to `st.ticker = time.NewTicker(time.Hour)` while still publishing `st.interval`, with `worker()` sleeping 18ms before `task()`. The previous assertion passed 20/20 under that pair with zero post-change ticks; the replacement fails 20/20.

chrisguiney:
The lower bound on continued ticking no longer reads the first task's counter. `tick()` enqueues onto `taskQueue` and the worker pool drains it asynchronously, so a closure enqueued by the last pre-change tick can increment that counter after the reading is taken — satisfying the bound on a scheduler that stopped ticking at the change. A second task is registered once the interval change is confirmed instead, and every execution of it is driven by a tick the new ticker delivered.

Fail-before, against `ledger/timer.go`: `run()`'s update case rewritten to `st.ticker = time.NewTicker(time.Hour)` while still publishing `st.interval`, with `worker()` sleeping 18ms before `task()`. The previous assertion passed 20/20 under that pair with zero post-change ticks; the replacement fails 20/20, and passes 20/20 with the worker delay alone and ticking intact.

The confirm loop retries `ChangeInterval` until `st.interval` shows the change, so the dropped send in #4155 delays that loop rather than reaching the new assertion.

`ledger/slot_clock_test.go` carries no assertion of this class. Both waits changed here block on a subscription channel created before `Start`, with no baseline reading for in-flight work to satisfy.

The two red `go-test` jobs are #4154 — `TestFetchEndorserBlockByPointDeadlineDoesNotCoolDownPeer` in `ouroboros`, which this change does not touch.

wolf31o2 (approved):
Accept

PR #4169, opened 2026-09-11
fix(ledger): classify pipeline backpressure and shutdown drain errors

### LINKED ISSUE #4178: ledger: apply-stage backpressure and shutdown cancellation are counted as unexpected pipeline errors
## Problem

`recordBlockPipelineError` (`ledger/state.go`) classifies every error drained from `blockPipeline.Errors()` into three buckets: expected eta0-unavailable, deferred epoch-cache, and `blockPipelineUnexpectedErrors`. Two error classes fall into the `default` branch that should not, so both are counted as unexpected and logged at ERROR level.

### 1. `pipeline.ErrPendingLimitExceeded` — apply-stage backpressure

gouroboros' apply stage (`pipeline/apply_stage.go:136`, forwarded at `:334`) buffers every item whose sequence number is not `nextSequence`. `MaxPendingBlocks` defaults to 2160 and dingo never overrides it (`ledger/state.go:1435-1461`). When one stage worker falls behind its siblings, the reorder gap grows; past the limit every subsequent buffered item emits this error. The item is still buffered and still applied in sequence — this reports scheduling lag, not a failed or dropped block.

Not reachable from the production read path today: the sole submitter (`ledger/state.go:5199`) is fed batches capped at `batchSize = 50` (`ledger/state.go:74`), far below 2160. Raising `batchSize` above that would make it a live counter, so the coupling deserves to be explicit.

### 2. `context.Canceled` / `context.DeadlineExceeded` — shutdown

`BlockPipeline.Stop()` cancels at `pipeline.go:407` before draining, and the send at `worker_pool.go:140` races the `ctx.Done()` arm, so workers that pull an item after cancellation emit `ctx.Err()` onto the errors channel (`decode_stage.go:52`, `validate_stage.go:81`, `apply_stage.go:108`/`:160`).

**This one is reachable in production today.** Every dingo shutdown with in-flight blocks increments `blockPipelineUnexpectedErrors` and logs at ERROR level (`ledger/state.go:5111-5116`) for what is ordinary, correct shutdown behavior.

## How it surfaced

`ledger.TestDecodeReadChainBatchDoesNotDeadlockOnManyValidationErrors` asserts `blockPipelineUnexpectedErrors` is zero after submitting 4000 validation-failing blocks. It fails intermittently on loaded CI — observed on `go-test (macOS)` with `Should be zero, but was 469`.

Reproduced locally at `GOMAXPROCS=1` with `-count=100`: the assertion fails, and 3004 of 3004 captured ERROR log lines read `pipeline: pending block limit exceeded` — zero other error identities. Measured reorder gap over 10 runs: 130-360 at GOMAXPROCS=14 (never reaches the limit), 1002-3522 at GOMAXPROCS=1 (routinely exceeds it). Controls at 100 runs each: 1 decode/1 validate worker = 0 failures; `WithMaxPendingBlocks(5000)` = 0 failures.

Note `-race` suppresses the failure entirely (0/150 at GOMAXPROCS=1 where the matched non-race config fails ~47%), so a race-enabled job cannot catch this.

## Secondary defect in the test

The assertion runs after only `require.Eventually(expectedEta0 > 0)`, which the first drained error satisfies, while the drain goroutine is still classifying. It therefore races in the **passing** direction — the test can go green while the bug is present, and the observed 469 was a partial count. The test also builds its `LedgerState` with `testLogger()` (`io.Discard`), so the error text behind the counter is destroyed and CI failures carry no diagnostic.

## Acceptance criteria

- Classify apply-stage backpressure and shutdown cancellation into their own counters rather than `blockPipelineUnexpectedErrors`, so the unexpected counter keeps its meaning as a genuine decode/validate/apply problem.
- Make the test observe a definite end-of-drain state rather than polling a counter that a concurrent goroutine is still writing.
- Make a future failure name the error rather than only its count.
- Keep the `batchSize`/`MaxPendingBlocks` relationship explicit so raising the batch size does not silently make backpressure a live production counter.

chrisguiney:
`TestDecodeReadChainBatchDoesNotDeadlockOnManyValidationErrors` flaked under
CPU load with `Should be zero, but was 469`. The failures were
`pipeline.ErrPendingLimitExceeded` from gouroboros' apply stage
(`pipeline/apply_stage.go:136`): the apply stage buffers out-of-order items,
`MaxPendingBlocks` defaults to 2160, and when a validate-stage worker lags,
the reorder gap exceeds that limit and every later buffered item emits the
error. `recordBlockPipelineError` routed it to the "unexpected" bucket,
logging at ERROR and failing the test's "no unexpected errors" assertion.

The item is still buffered and still applied in sequence, so this is
apply-stage scheduling backpressure, not a failed or dropped block.
Classifies it into its own debug-level counter
(`blockPipelineApplyPendingLimitErrors`). Production cannot reach this today:
the only submitter (`decodeReadChainBatch`) caps batches at `batchSize` (50)
and drains each batch's results before submitting the next, far under 2160.

Also classifies `context.Canceled`/`context.DeadlineExceeded` into a
shutdown counter (`blockPipelineShutdownErrors`): `BlockPipeline.Stop`
cancels the pipeline context before draining, so a stage worker mid-item at
shutdown can report its own cancellation instead of an item outcome. Every
shutdown with blocks in flight could previously increment the
unexpected-error counter and log at ERROR.

Also fixes an end-of-drain race in the test (it polled for
`blockPipelineExpectedEta0Errors > 0`, which the first drained error already
satisfies, so the unexpected-error assertion that followed could run before
later errors were classified) and makes the test capture ERROR-level log
text so a future failure names the error instead of only its count.

Reverting the `ErrPendingLimitExceeded` classification and running
`GOMAXPROCS=1 go test ./ledger/ -run TestDecodeReadChainBatchDoesNotDeadlockOnManyValidationErrors -count=100`
under synthetic CPU load: 45/100 failed, all on
`pipeline: pending block limit exceeded`. With the classification in place,
identical load: 0/100 failed.

---
## Summary by cubic
Fixes a flaky ledger test by classifying two previously unexpected pipeline drain errors as debug-level counters, and fixes a race in the test's end-of-drain check.

**Bug Fixes**
- `pipeline.ErrPendingLimitExceeded` now increments a new apply-pending-limit counter and logs at debug; it is apply-stage backpressure, not a block failure. Production cannot reach this today (batch size 50 is far below `MaxPendingBlocks`).
- Stage-worker `context.Canceled`/`context.DeadlineExceeded` during shutdown now increment a shutdown counter and log at debug; previously they counted as unexpected and logged at ERROR on every shutdown with blocks in flight.
- The deadlock regression test now waits for the drain goroutine to finish before asserting counters, and captures ERROR log text so a future failure names the error instead of only its count.

<sup>Written for commit f726660f5e08253eb266189e35ba3c224db1f718. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pipeline error handling during shutdown and scheduling backpressure.
  * Prevented expected cancellation and pending-buffer limit events from being reported as processing failures.
  * Added clearer diagnostic tracking for pipeline drain behavior.

* **Documentation**
  * Expanded documentation describing expected shutdown cancellations and apply-stage backpressure.

* **Tests**
  * Added coverage for error classification, pending-buffer overflow, and reliable pipeline shutdown behavior.

Closes #4178


ledger/state.go
```diff
@@ -5080,6 +5080,23 @@ func (ls *LedgerState) drainBlockPipelineErrors() {
...
 //     resolves once the epoch cache catches up, so it is not lumped in with
 //     genuine decode/validate/apply problems below.
+//   - context.Canceled/context.DeadlineExceeded: the pipeline context is
+//     cancelled by Stop() *before* it drains, so a stage worker that is
+//     mid-item when shutdown begins can lose the race between its
+//     `errors <- err` send and its own ctx.Done() arm and report the
+//     cancellation here. Every shutdown with blocks still in flight can
+//     therefore produce a handful of these; they say the node is stopping,
+//     not that a block failed.
+//   - pipeline.ErrPendingLimitExceeded: the apply stage's out-of-order buffer
```
wolf31o2, line 5298: This newly added case makes the preceding intro inaccurate: it says there are two expected/transient cases, but context cancellation/deadline and pending-limit are now additional debug-classified cases. Please make that wording count-independent.

wolf31o2 (changes requested):
The implementation and regression coverage look correct. One documentation mismatch remains.

ledger/state.go
chrisguiney: Fixed in 3c18378c: reworded the intro to "The cases below are expected/transient..." so it no longer names a stale count.

ledger/state.go
```diff
@@ -5080,6 +5080,23 @@ func (ls *LedgerState) drainBlockPipelineErrors() {
 //     ("Block Processing Pipeline") documents this as a transient race that
 //     resolves once the epoch cache catches up, so it is not lumped in with
 //     genuine decode/validate/apply problems below.
+//   - context.Canceled/context.DeadlineExceeded: the pipeline context is
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 3c18378c: reworded the intro to "The cases below are expected/transient..." so it no longer claims a fixed count.

wolf31o2 (approved):

PR #4173, opened 2026-09-11
test(dblifecycle): synchronize flaky-retry assertions against the startup scan

### LINKED ISSUE #3759: test(dblifecycle): wait deadlines assume a fast runner and flake on Windows

chrisguiney:
Fixes #3759 (the second, non-Windows mechanism only; see below).

## Two mechanisms in #3759

**Windows-runner-speed deadlines** were already fixed by #4131, which swept
every `require.Eventually`/`RequireReceive`/`WaitForCondition` in this file
onto a single 30s `snapshotWait` constant. Auditing all 31 wait call sites in
`manager_test.go` found none left on a raw short deadline; no change needed
for this class.

**A retry racing an assertion** (not Windows-specific, reported by
@wolf31o2): `TestManagerRetriesCloudMirrorAfterTransientFailureOnRedeliveredEvent`
and `TestManagerRetriesUnmirroredSnapshotOnLaterEpochWithoutRedelivery` assert
`IsCloudMirrored` is false immediately after a "snapshot failed" log line.
`Manager.Start` launches a background scan (`retryUnmirroredSnapshots`) that
retries any unmirrored `epoch-*` directory it finds; the epoch handler's own
first upload (`SnapshotToCloud`) does not take `retryMu`, so the two are not
mutually exclusive. If the scan runs after the epoch's local write lands but
before the handler's own upload attempt, both race for the same directory,
and since they share the fake cloud destination's one-shot failure flag,
whichever wins uploads successfully — flipping the assertion at random.

## Fix

Give the startup scan an existing, non-numeric-suffixed probe directory to
find, and block on a channel until its upload is observed before publishing
the real epoch event. That proves the scan's one and only directory listing
predates the epoch directory's existence, so it cannot race the epoch
handler's own attempt (`internal/dblifecycle/manager_test.go`).

## Causal proof

The race is rare unamplified (30/30 and 300/300 pass without it). With a
diagnostic delay in the startup-scan goroutine (reverted, not part of this
diff) and without this fix: 5/8 and 6/8 runs failed on the exact assertion
reported (`must not be marked mirrored after a failed upload`). With the
delay and the fix: 0/16. A flat delay also passes, indicating the fix is
structural rather than timing-dependent.

## Outstanding

Windows CI confirmation is outstanding — the acceptance criteria for #3759
call for it, and it could not be run locally. The mechanism-1 conclusion
("already fixed by #4131") is a code-reading result, not Windows-observed.

---
## Summary by cubic
Fixes #3759 by synchronizing the startup retry scan against the epoch handler's upload in `internal/dblifecycle/manager_test.go`. The Windows deadline class was already fixed by #4131, so only the non-Windows race needed work.

The startup scan isn't mutually exclusive with the epoch handler's first upload — only the scan takes `retryMu`, and both share the fake cloud's one-shot failure flag — so the scan could race the handler and randomly flip the "must not be mirrored" assertion. The tests now seed a probe directory and wait until the scan has finished mirroring it before publishing the real epoch event.

**Bug Fixes**
- Wait for the `.cloud-mirrored` marker via `IsCloudMirroredTo`, not just the upload channel: the channel fires from inside `UploadDir` before the marker is written, so a redundant scan could still retry the probe and consume the one-shot failure.

<sup>Written for commit 6a551eac1830456e1a6a4cd28c51ad81d9653703. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4173?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Tests**
  - Expanded coverage for cloud-upload lifecycle scenarios, including transient failures, retry attempts, redelivered events, later state transitions, and manager restarts.
  - Added verification that retries reuse existing local snapshots and correctly handle multiple unreliable cloud destinations.
  - Improved test synchronization and upload observability for more reliable validation of recovery behavior.


cryptodj413 (dismissed):

internal/dblifecycle/manager_test.go
```diff
@@ -1045,7 +1131,19 @@ func TestManagerRetriesCloudMirrorAfterTransientFailureOnRedeliveredEvent(
 	require.NoError(t, m.Start(context.Background()))
 	defer m.Stop()
 
-	// First delivery: local write succeeds, cloud upload fails.
+	require.Equal(t, startupProbeDir, testutil.RequireReceive(
```
wolf31o2, line 1134: `uploaded` is sent from `flakyCloudDestination.UploadDir` before `SnapshotToCloud` returns and writes the `.cloud-mirrored` marker. After this receive, the epoch event can start while the startup scan is still unwinding; its retry can observe the unmarked probe and consume the simulated failure/success transition. Synchronize on the marker (or add a completion signal after the caller returns) before resetting the fake and publishing epoch 9. Apply the same fix to the second test at lines 1263-1270.

wolf31o2 (changes requested):
The test is otherwise focused, but the startup synchronization still has a race.

internal/dblifecycle/manager_test.go
chrisguiney: Synchronized on the marker's completion instead of the mid-upload channel: all three tests using the startup-probe technique (this one, TestManagerRetriesUnmirroredSnapshotOnLaterEpochWithoutRedelivery, and TestManagerPruningDeletesCloudMirror) now wait on `lifecycle.IsCloudMirroredTo(startupProbeDir, cloudDest)` before proceeding, not `IsCloudMirrored` (bare `os.Stat`): that call can observe the marker file the instant `os.WriteFile`'s internal `O_CREATE|O_TRUNC` open lands, before the content write that `retryUnmirroredSnapshots`'s own idempotency check reads.

Reproduced the window by narrowing retryMu to just the directory scan (releasing it before the actual upload+marker-write) plus a delay between UploadDir returning and the marker write: 300/300 runs across the three tests showed the redundant probe retry with the old channel-only wait, including one outright failure in TestManagerPruningDeletesCloudMirror from a misattributed channel value. Same adversarial run with this fix: 0/300.

wolf31o2 (approved):

PR #4174, opened 2026-09-11
perf(sqlstore/migrations): share one baseline across replay subtests

### LINKED ISSUE #3746: ci: Windows suite runtime is unpredictable and 10x Linux on ./database

### LINKED ISSUE #4171: test(sqlstore/migrations): parallel migration replay exceeds the 30m per-package bound on Windows
## Problem

`database/plugin/metadata/sqlstore/migrations` is the only package observed to fire `go test`'s `-timeout 30m` per-package bound on the Windows runner, and it does so on branches that do not touch it.

Measured across 98 completed `go-test.yml` runs (2026-09-01 to 2026-09-11), by grepping every `failure`-concluded `go-test (Windows)` job log for `panic: test timed out after 30m0s`:

- 4 of 98 runs (~4%) fired the per-package timeout.
- **All four are this package.** The other 15 failure-concluded Windows jobs are ordinary assertion failures with no timeout panic.
- The four occurred on `fix/forge-parent-tip-signing`, `codex/3611-fix`, `fix/382-localstatequery-point-pinning` and `fix/3427-bounded-forecast` — none of which touch this package, ruling out a triggering diff.

## Mechanism

No single test blocks forever. The goroutine dumps at panic time list several `t.Parallel()` subtests each a few minutes in, for example `TestRunnerReplaysEveryShippedVersionFromExpand/pool-registration-deposit-held (3m15s)`.

The package has 50 top-level test functions and 46 `t.Parallel()` calls across 11 files. Each subtest opens its own real SQLite migration database (`openTestDB` / `baselineBackfillDB`) and replays all 13 shipped migrations. That is heavy, highly parallel real file I/O, which on a contended Windows runner occasionally exceeds the 30-minute budget — the same package completes in about 446s on a normal run.

## Acceptance criteria

- Establish whether the parallel fan-out can safely share a migration-replay fixture, or whether each subtest genuinely needs its own database.
- Bring the package's Windows worst case inside the `-timeout 30m` bound without reducing migration coverage.
- Do not fix this by raising the per-package timeout; the bound is what produces the goroutine dump that makes a real stall diagnosable.
- Keep `t.Parallel()` unless sharing a fixture is proven unsafe — the cost is fixture setup, not parallelism itself.

## Context

Found while investigating #3746. That issue's headline "42m+" run turned out to be routine concurrency-group cancellation rather than a stall (`Canceling since a higher priority waiting request for go-test-3744 exists`), and this package is the one genuine instability the sample surfaced. Excluding cancellations, the Windows `go-test` step distribution over 85 completions is min 411s, median 1235s, max 2306s against a 90-minute job cap.

This is also why `go-test (Windows)` should not become a required context yet: a ~4% timeout rate in a shared package would periodically block unrelated, correct pull requests.

chrisguiney:
`database/plugin/metadata/sqlstore/migrations` has exceeded the 30-minute
per-package bound on Windows CI 5 times, always this package, always on
branches that do not touch it (#4171). A goroutine dump from the fifth
occurrence shows nothing stuck: 16 of 28 goroutines are parked in
`testing.(*testState).waitParallel`, only 4 running, longest 2m58s. The
package is ~20 multi-minute subtests draining through a handful of parallel
slots.

`TestRunnerReplaysEveryShippedVersionFromExpand` ran a full, fresh
13-migration replay inside every `t.Parallel()` subtest, and that first
`Runner.Run()` call was byte-for-byte identical work every time (same
registry, same empty database, stateless `NewProcessLocker()`). Each replay
is ~419 individually autocommitted, fsync'd write-transactions (340 DDL + 78
state-bookkeeping + 1 state-table create), because `execDDL` and the state
helpers run on a bare `*sql.Conn` with no wrapping transaction, and the test
database opens at SQLite's default `journal_mode=DELETE` /
`synchronous=FULL`.

This builds the migrated baseline once per test function and gives each
subtest a byte copy of the file. `journal_mode=DELETE` leaves no journal file
after a clean `Close()`, so the closed file alone is a complete database; and
`NewProcessLocker()` is a fresh in-memory channel token per call, so a copy
carries no lock state. `t.Parallel()` is preserved and each subtest still
resets and replays its own migration version against its own copy, so
per-version replay coverage is unchanged.

Local (darwin/arm64): write-transactions for the test function drop from
~5,538 to ~510 (91%); wall time from 3.0-3.8s to 0.38-0.41s (88-90%). The
Windows per-transaction cost and the projected per-subtest speedup are not
measured — they are inferred from the issue's single Windows data point,
where fsync and antivirus interception on journal-file create/delete are
expected to be far more expensive than on macOS.

If this holds on Windows CI, it removes the dominant contributor to this
package's 30-minute timeout, which is currently why `go-test (Windows)`
cannot become a required check (#3746).

Addresses #4171.

---
## Summary by cubic
Shares one fully-migrated baseline across all subtests in `TestRunnerReplaysEveryShippedVersionFromExpand`, replacing a full 13-migration replay in every parallel subtest with a byte copy of the same result. Cuts the test’s local wall time from ~3.5s to ~0.4s and should bring the package under the 30-minute Windows CI timeout observed in #4171.

**Refactors**
- Builds the baseline once per test function by running the full registry against a fresh database.
- Each subtest seeds its own database from a copy of that baseline file, then runs its per-version replay assertion unchanged.
- The baseline is a valid closed database because `journal_mode=DELETE` leaves no journal file, and `NewProcessLocker()` is stateless.

<sup>Written for commit c40ed405a1be5805b92c0964b471cbb6c6e1fdfa. Summary will update on new commits.</sup>

Closes #4171

### DIFF
--- a/database/plugin/metadata/sqlstore/migrations/runner_test.go
+++ b/database/plugin/metadata/sqlstore/migrations/runner_test.go
@@ -19,6 +19,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -616,21 +617,33 @@ func TestRunnerDoesNotSkipNonAddColumnFailure(t *testing.T) {
 // interrupted upgrade. Replaying expand also replays backfill and contract, so
 // this covers every phase of every shipped version against a database that
 // already has the version's full effect applied.
+//
+// Every subtest's first Run() call was byte-for-byte identical work -- same
+// fresh database, same full registry, same stateless NewProcessLocker -- so it
+// is built once here and shared as a byte copy instead of being replayed by
+// every subtest. That first replay is ~400 individually autocommitted
+// DDL/state writes (13 migrations, each a separate fsync'd transaction under
+// SQLite's default synchronous=FULL/journal_mode=DELETE); redoing it in every
+// one of the ~13 parallel subtests is what pushes this package toward the
+// per-package -timeout bound on a CI runner where fsync is expensive
+// (dingo#4171). The per-migration replay under test -- resetting one version
+// to PhaseExpand and calling Run() again -- still runs against each subtest's
+// own independent copy, so the coverage this test exists for is unchanged.
 func TestRunnerReplaysEveryShippedVersionFromExpand(t *testing.T) {
 	t.Parallel()
 	registry, err := SQLiteRegistry()
 	require.NoError(t, err)
+	baseline := fullyMigratedBaseline(t, registry)
 	for _, migration := range registry {
 		t.Run(migration.Name, func(t *testing.T) {
 			t.Parallel()
-			db := openTestDB(t)
+			db := openTestDBFromBaseline(t, baseline)
 			runner := &Runner{
 				DB:       db,
 				Dialect:  "sqlite",
 				Registry: registry,
 				Locker:   NewProcessLocker(),
 			}
-			require.NoError(t, runner.Run(context.Background()))
 
 			_, err := db.Exec(`
 UPDATE schema_migrations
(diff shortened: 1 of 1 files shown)


cryptodj413 (approved):

database/plugin/metadata/sqlstore/migrations/runner_test.go
```diff
@@ -658,3 +671,46 @@ SELECT phase, dirty, completed_at FROM schema_migrations WHERE version = ?`,
...
+	)
+	require.NoError(t, err)
+	runner := &Runner{
+		DB:       db,
+		Dialect:  "sqlite",
+		Registry: registry,
+		Locker:   NewProcessLocker(),
+	}
+	require.NoError(t, runner.Run(context.Background()))
+	require.NoError(t, db.Close())
```
wolf31o2, line 694: `fullyMigratedBaseline` opens a database but does not register `db.Close` with `t.Cleanup`. If `runner.Run()` fails, `require.NoError` exits before the explicit close below, which can leave SQLite handles open and make Windows temporary-directory cleanup flaky. Register cleanup immediately after opening the database; keep the explicit close before reading the copied file.

wolf31o2 (changes requested):
The implementation is otherwise reasonable, but this test/reliability gap should be fixed.

wolf31o2 (approved):

Issue #4102, opened 2026-09-07
ledger: cross-fork continuation audit fires during early replay; fork rollback targets a slot ahead of the local tip


chrisguiney:
The cross-fork continuation audit added for #3005 fired **54 times in a six-second
window** during a from-genesis Preview replay, at ledger slot ~4.02M (epoch ~46).
Since the audit is documented as sound — it disarms rather than risk a false
report when its producer set is truncated — each report is a block that reached
the ledger spending an input with no producer on the local applied chain.

Unlike #3005, which never converged near the tip, this node **recovered**:
`persistent chain fork detected, triggering chainsync re-sync` fired at
`consecutive_mismatches=20`, and the chain has since advanced well past the fork
point. So this is the same audit signal with a different outcome, early in the
chain rather than at 99.9%.

## The audit reports

Representative, unabridged:

```
level=ERROR msg="continuation block spends an input with no producer on the local applied chain"
  component=ledger peer=10.0.0.5:64351<->192.234.196.162:4201
  block_slot=4029306 block_hash=95ff150274a056d53ae13c857c1552adadc866d8eaa7845e8edce3acf7359900
  block_prev_hash=6da5f003c9728bbc45a30602959c810854db4389490f09cdfb0dbf408ed0a0ff
  tx_hash=be46a24abed0408abf42221dfd45805a98821e0795ca5d53f512e3caa2f28eae
  input=a58cdb6a74adf51148026f341ff8a87a1759d17a23bfc52aaacc77e34b7fc68d#1
  producer_tx_hash=a58cdb6a74adf51148026f341ff8a87a1759d17a23bfc52aaacc77e34b7fc68d
  fork_rollback_slot=4028999
  fork_rollback_hash=b78e4e385fb95fc08c18e06033443c19055f69f820be289f09f6d64fdbc222fb
  fork_reason="chainsync rollback" blocks_since_fork=4
```

All 54 name the same peer, `192.234.196.162:4201`.

## The rollback target is ahead of the local tip

Every `fork detected: rolling back to common ancestor` line in the run reports an
`ancestor_slot` **greater** than the `local_tip_slot` it is rolling back from:

| local_tip_slot | ancestor_slot | ancestor − local_tip |
|---|---|---|
| 4023273 | 4023423 | +150 |
| 4024387 | 4024535 | +148 |
| 4024732 | 4024886 | +154 |
| 4025772 | 4026102 | +330 |
| 4027630 | 4027759 | +129 |
[...]

Issue #4133, opened 2026-09-08
ledger: local chain diverges to exactly k blocks and cannot self-recover; node permanently stops following the chain


chrisguiney:
A from-genesis Preview replay reached a state where the local chain has diverged
from the network by exactly the security parameter and cannot be reconciled with
any peer. The node stops following the chain permanently and says so:

```
level=ERROR component=ledger
  msg="chainsync stuck: repeatedly cannot cross to peer rollback point; local
       chain has diverged from the network and cannot self-recover. Operator
       intervention required (e.g. re-bootstrap from a Mithril snapshot)."
  slot=9796272 reason="rollback point not found" failed_attempts=3 window=10m0s
```

Unlike the bounded stalls in #4128 (5m06s and 3m00s, both self-recovered), this
one did not clear: **17m35s with no block at time of writing**, across 13
distinct peers, with 188 `chain selection stalled: no selectable peer` events.
Local tip 9,770,093 against a peer tip of 122,173,488.

## The divergence is exactly k

```
startup:  Genesis chain selection enabled genesis_window_slots=25920 security_param=432
stall:    WARN "chain fork detected" fork_point_slot=9758685 fork_depth=432
                alternate_head_slot=9766862 canonical_head_slot=9758685
```

`fork_depth=432` is exactly the configured `security_param`. The node is
therefore in a position where every peer offers a chain that would require
rolling back k or more, which the security-parameter guard correctly refuses:

```
ERROR "chainsync rollback exceeds security parameter K, rejecting peer chain"
```

Both behaviours are individually correct — refusing a rollback beyond k is
required, and refusing to follow a chain it cannot reach is honest — but together
they are terminal. There is no path back.

## Likely origin

This run logged **1,145** cross-fork continuation splices (#4102), where blocks
whose inputs have no producer on the local applied chain were nevertheless
applied and only caught afterwards by the audit. A local chain assembled partly
from other peers' forks is exactly the kind of chain that no peer can serve a
rollback point for.

I am presenting that as the plausible mechanism rather than a proven one — I have
not traced a specific spliced block into the diverged segment. But the ordering
fits: 1,145 splices accumulated over ~2 hours, then the chain reached exactly k
of divergence and stopped.

## Why it matters beyond a test replay

The recovery advice in the message is re-bootstrap from a Mithril snapshot,
meaning the node discards its chain and starts over. On a from-genesis replay
that is ~2 hours of work lost. On a production relay it is an outage requiring
[...]

PR #4191, opened 2026-09-11
fix(ledger): fall back after missing replay parent

wolf31o2:
When replay recovery resolves transaction metadata to a producer block whose parent is no longer retained, treat that provenance as unresolved and use the existing bounded security-parameter fallback. Preserve non-missing database errors.

Adds regression coverage for a pruned producer parent.

Related to #4102.

## Summary by CodeRabbit

* **Bug Fixes**
  * Replay recovery now continues safely when a producer’s parent block is unavailable locally.
  * Recovery can fall back to a security-parameter anchor instead of aborting candidate selection.


ledger/replay_recovery.go
```diff
@@ -1636,8 +1644,11 @@ func (ls *LedgerState) findReplayRecoveryCandidate(
 		if err != nil {
 			return nil, err
 		}
-		if fallbackCandidate != nil && (candidate == nil ||
-			fallbackCandidate.ProducerBlock.Slot < candidate.ProducerBlock.Slot) {
+		if fallbackCandidate != nil {
+			// Any unresolved provenance means the applied state may diverge
+			// before a metadata-known producer. Prefer the bounded fallback
+			// over a narrower known-producer rewind so recovery does not leave
+			// the unresolved input above the rollback point.
```
chrisguiney: Blocker: this discards a deeper known-producer rollback point.

`candidate` here holds the earliest resolved producer. When that producer's block sits below the fallback anchor (`failingBlock.ID - SecurityParam()`), replacing it with `fallbackCandidate` moves the rollback point *later*. The block that produced the missing UTxO then stays applied, re-applying from the shallower anchor cannot recreate it, and the same transaction fails again at the same block.

Reproducible on a five-block primary chain with `securityParam: 1`, one input resolving to the second block (slot 120, parent at slot 100) and one input with no transaction metadata: `findReplayRecoveryCandidate` returns `strategy=security-param-fallback producer_slot=160 rollback_slot=140`, where the previous predicate returned the slot-120 producer with `rollback_slot=100`. That contradicts `TestTryRecoverFromTxValidationErrorRollsBackToEarliestProducerParent`, which pins recovery to the earliest producer's parent precisely so every failing input is below the rollback point.

The gap the earlier review comment named is `ProducerUnresolved`, not the rollback point: it is set only on the fallback candidate and gates `rewindPrimaryChain` at line 239, so a winning known candidate skipped the primary-chain rewind. Setting the flag closes that without losing the deeper anchor:

```go
if fallbackCandidate != nil {
	if candidate == nil ||
		fallbackCandidate.ProducerBlock.Slot < candidate.ProducerBlock.Slot {
		candidate = fallbackCandidate
	} else {
		// Unresolved provenance still requires the primary chain to be
		// rewound, but the deeper known producer sets the anchor.
		candidate.ProducerUnresolved = true
	}
}
```

ledger/replay_recovery.go
```diff
@@ -1601,6 +1601,14 @@ func (ls *LedgerState) findReplayRecoveryCandidate(
 			resolved.ProducerBlock,
 		)
 		if err != nil {
+			if errors.Is(err, models.ErrBlockNotFound) {
+				// The transaction metadata identifies a producer, but its
+				// parent is outside the locally retained chain. Treat the
+				// provenance as unresolved so the bounded security-parameter
+				// fallback can choose a safe local anchor.
```
chrisguiney, line 1636: Blocker: incomplete for a pruned tail, which is the topology the comment above describes.

`ErrBlockNotFound` reaches `findReplayRecoveryCandidate` from two further points on this path that still propagate it:

- `replayRecoveryFallbackCandidate` -> `ls.db.BlockByIndex(targetIndex, nil)` (line 1912)
- `replayRecoveryFallbackCandidate` -> `ls.replayRecoveryParentPoint(anchorBlock)` (line 1920)

On a five-block primary chain with the first three blocks deleted — what retention pruning and a Mithril bootstrap actually leave behind — and the producer as the fourth block, `findReplayRecoveryCandidate` still returns an error rather than a candidate:

- `securityParam: 1` -> `lookup parent block for replay recovery at slot 120: block not found` (line 1920)
- `securityParam: 3` -> `lookup replay fallback block 2: block not found` (line 1912)

The new test passes because it removes a block from the *middle* of the chain, leaving an older block with an intact parent available as the anchor. `failingBlock.ID - SecurityParam()` routinely lands below the retained window on a Mithril-bootstrapped node, so the fallback needs the same treatment before it can serve as the safe path this change routes the missing-parent case into.

ledger/replay_recovery.go
```diff
@@ -1601,6 +1601,14 @@ func (ls *LedgerState) findReplayRecoveryCandidate(
...
+				unresolvedInputs = append(unresolvedInputs, resolved.Input)
```
chrisguiney, line 1648: Recommendation: this case is folded into `unresolvedInputs` with no signal.

A producer block that the transaction index references but whose parent is absent from the block store is a local-consistency fact, and it is not the same condition as an input with no recorded provenance at all. Both now arrive at the fallback under the single `security-param-fallback` strategy label, so the recovery log cannot distinguish them. #4102 is a diagnosis problem; a `Warn` here naming the producer block hash and slot would let an operator tell "no provenance recorded" from "the block store is missing a block the index still points at".

ledger/replay_recovery_test.go
```diff
@@ -1737,6 +1812,44 @@ func TestReplayRecoveryArmsAuditAfterPrimaryAndLedgerRewind(t *testing.T) {
...
+		Inputs: []lcommon.TransactionInput{
+			&replayRecoveryInput{txId: knownTxHash, index: 0},
+			&replayRecoveryInput{txId: testHashBytes("unresolved-producer")},
+		},
+		Cause: errors.New("bad input"),
+	})
+	require.NoError(t, err)
+	require.NotNil(t, candidate)
+	assert.Equal(t, "security-param-fallback", candidate.Strategy)
+	assert.Equal(t, uint64(100), candidate.ProducerBlock.Slot)
```
chrisguiney: Blocker: this test does not pin the behaviour it is named for.

Restoring the previous predicate

```go
if fallbackCandidate != nil && (candidate == nil ||
	fallbackCandidate.ProducerBlock.Slot < candidate.ProducerBlock.Slot) {
```

and changing nothing else at this head leaves `go test ./ledger/` green, this test included.

The cause is the other production change: with the `ErrBlockNotFound` handling in place, `knownTxHash`'s producer (`pruned-known-producer`, whose `PrevHash` is not in the block store) also lands in `unresolvedInputs`. `candidate` is therefore still nil when the fallback is computed, and the old predicate's `candidate == nil` branch selects the fallback anyway — so both assertions hold either way.

To pin the new preference, the known input's producer needs a *retained* parent and a block slot below the fallback anchor, so the two candidates genuinely compete.

chrisguiney (changes requested):
Requesting changes: three blockers.

The `ErrBlockNotFound` handling at the producer-parent lookup is sound and causally covered — removing that hunk alone fails both new tests. The second production change, always preferring the fallback candidate, is covered by no test and moves recovery in a direction the existing contract test contradicts.

Issue linkage: #4102 reports blocks reaching the ledger whose inputs have no producer on the local applied chain, and a fork-rollback log whose `ancestor_slot` exceeds `local_tip_slot`. This diff changes neither the continuation audit, the splice, nor the rollback target; it makes replay recovery tolerate one `block not found` lookup, which is the symptom in https://github.com/blinklabs-io/dingo/issues/4102#issuecomment-5622118052. `Fixes #4102` would close that issue while the reported condition persists — `Related to #4102` is accurate, with a narrower issue for the missing-parent lookup. The reference also appears twice in the description (authored text and generated summary). #4133 is adjacent rather than fixed: this change routes more recoveries into the k-bounded fallback, and #4133 is a wedge at exactly that boundary.

Pipeline: `go-test (Windows) (1.26.x)` was cancelled at the job's 90-minute backstop. Its log ends with `ok github.com/blinklabs-io/dingo/config/cardano` at 20:12:54Z, then 82 minutes of silence and `##[error]The operation was canceled.` — no test failure, and no goroutine dump from the step's `-timeout 30m`. The same job completes in 24-39 minutes on `main`. The merge base predates #3982, which bounded the disk each test Badger store reserves after this job failed with "There is not enough space on the disk" once a package built several stores, leaking a handle and breaking `TempDir` cleanup for every later test. This change adds two more test databases in `ledger`. Merging current `main` is the likely fix; the cancellation is not attributable to the diff's logic. `go build ./...`, `go vet ./ledger/...`, `golangci-lint run ./ledger/...` and `go test [-race] ./ledger/` are clean at this head and on the tree merged with current `main`.

`ARCHITECTURE.md` and `DATABASE.md` describe the durable-applied-floor invariant this path already honours; no change needed there.

ledger/replay_recovery.go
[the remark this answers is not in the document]
chrisguiney: Valid, and the root cause named here is narrower than the fix applied. `ProducerUnresolved` is set only on the fallback candidate and gates `rewindPrimaryChain` at line 239, so a winning known candidate skipped the primary-chain rewind — that flag is the gap.

Preferring the fallback unconditionally also gives up the deeper anchor whenever the known producer's block sits below `failingBlock.ID - SecurityParam()`. With `securityParam: 1` on a five-block chain, a resolvable input producing at slot 120 (parent at slot 100) plus one unresolved input now yields `rollback_slot=140` where the previous predicate yielded `100`, leaving the producer of the missing UTxO applied. Detail and a suggested predicate are in the review comment on lines 1647-1651.

ledger/replay_recovery_test.go
```diff
@@ -1737,6 +1812,31 @@ func TestReplayRecoveryArmsAuditAfterPrimaryAndLedgerRewind(t *testing.T) {
 	assert.Equal(t, ls.Tip().Point, window.forkPoint)
 }
 
+func TestReplayRecoveryFallbackWinsOverKnownProducer(t *testing.T) {
```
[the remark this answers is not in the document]
chrisguiney: Valid as posted, and only half resolved at this head. The new `TestFindReplayRecoveryCandidateFallsBackWhenProducerParentIsMissing` does exercise the missing-parent path: removing the `ErrBlockNotFound` handling fails it with `lookup parent block for replay recovery at slot 120: block not found`.

`TestReplayRecoveryFallbackWinsOverKnownProducer` still does not discriminate, for a different reason than the one described here. Its `pruned-known-producer` block now has an unretained `PrevHash`, so that input also lands in `unresolvedInputs`, `candidate` stays nil, and the previous predicate's `candidate == nil` branch selects the fallback too — restoring that predicate leaves `go test ./ledger/` green. Detail in the review comment on lines 1847-1850.

ledger/replay_recovery.go
```diff
@@ -1639,6 +1658,10 @@ func (ls *LedgerState) findReplayRecoveryCandidate(
 		if fallbackCandidate != nil && (candidate == nil ||
 			fallbackCandidate.ProducerBlock.Slot < candidate.ProducerBlock.Slot) {
 			candidate = fallbackCandidate
+		} else if fallbackCandidate != nil {
+			// Keep the deeper known-producer anchor, but still rewind the
+			// primary chain because another input has unresolved provenance.
+			candidate.ProducerUnresolved = true
```
chrisguiney: Blocker: keyed on `fallbackCandidate != nil`, so when the new `nil, nil` fallback returns fire (pruned target, or anchor without a retained parent), a known candidate plus an unresolved input keeps `ProducerUnresolved=false` and line 239 skips the primary-chain rewind and hold. Six-block chain, first two blocks deleted, known producer at slot 140 plus one input without metadata: `securityParam` 2 sets the flag, 3 and 5 do not. Keying on `candidate != nil` sets it in all three with the `Replay|Recover` ledger tests green; a pruned-tail test should pin it.

database/plugin/metadata/sqlstore/committee_prune_test.go
```diff
@@ -180,6 +180,7 @@ func applyAuthCertificate(
 		ocommon.Point{Slot: slot, Hash: credentialHash(0x99)},
 		0,
 		nil,
+		allowUnknownDeposits,
```
chrisguiney: Merge blocker: conflicts with `main`, where #4222 repaired these call sites with `requireKnownDeposits` (the live block-apply value); take `main`'s side. Until resolved, `go-test`, `golangci-lint`, `config-parity` and `Docker CI` do not run on this head.

ledger/leios_header_announcement_test.go
```diff
@@ -124,11 +124,15 @@ func TestChainsyncHeaderAdmissionAnnouncesOnlyWhenCryptoVerified(
...
-	// AddBlockHeader branch -- the real end-to-end unverified admission.
+	// Mark the header as covered by an imported Mithril snapshot so the policy
+	// returns trustedWithoutVerification and the handler takes the AddBlockHeader
+	// branch -- the real end-to-end unverified admission. Header crypto is not
+	// bypassed merely because live validation is disabled.
 	t.Run("unverified admission is queued, not announced", func(t *testing.T) {
 		fixture := newHeaderStreamLedger(t)
+		fixture.ls.Lock()
+		fixture.ls.mithrilLedgerSlot = header.slot
+		fixture.ls.Unlock()
```
chrisguiney: Merge blocker: conflicts with `main`, where #4222 repaired this subtest through the missing-epoch-nonce path (`trusted=false`, still admitted via `AddBlockHeader`); take `main`'s side.

chrisguiney (changes requested):
Earlier blockers are resolved at 647bcc4b (earliest-producer anchor with the flag set, pruned-tail lookups, the test pinning it, issue reference). One new blocker inline, and the branch conflicts with main in two test files that #4222 already repaired.

ledger/replay_recovery.go
```diff
@@ -1668,6 +1687,11 @@ func (ls *LedgerState) findReplayRecoveryCandidate(
 			fallbackCandidate.ProducerBlock.Slot < candidate.ProducerBlock.Slot) {
 			candidate = fallbackCandidate
 		}
+		if candidate != nil && fallbackCandidate != nil {
```
chrisguiney, line 1690: Blocker, still present: the merge rewrote this but kept it keyed on `fallbackCandidate != nil`, so when a new `nil, nil` fallback return fires, a known candidate plus an unresolved input keeps `ProducerUnresolved=false` and line 239 skips the primary-chain rewind. Slots 80-180, first two blocks deleted, known producer at 140 plus one input without metadata: `securityParam` 2 sets the flag; 3 and 5 return strategy=metadata, rollback_slot=120, flag false. Keying on `candidate != nil` sets it in all three, `Replay|Recover` green; pin a pruned-fallback case and update the field doc at line 156.

ledger/leios_header_announcement_test.go
```diff
@@ -124,9 +124,10 @@ func TestChainsyncHeaderAdmissionAnnouncesOnlyWhenCryptoVerified(
 	)
 	point := ocommon.NewPoint(header.slot, header.hash.Bytes())
 
-	// The fixture has no cached epoch nonce, so the policy defers verification
-	// and the handler takes the AddBlockHeader branch -- the real end-to-end
-	// unverified admission.
+	// Mark the header as covered by an imported Mithril snapshot so the policy
```
chrisguiney, line 127: Merge blocker: the resolution kept `main`'s subtest body but this branch's comment. The subtest sets no `mithrilLedgerSlot` and asserts `trusted` false, which `chainsyncHeaderCryptoPolicy` returns through the missing-epoch-nonce branch, not `trustedWithoutVerification`. This file is outside the change; restoring `main`'s three-line comment removes it from the diff.

chrisguiney (changes requested):
Merge audit of 904119a8: `committee_prune_test.go` now matches `main`, and ARCHITECTURE.md and DATABASE.md drop no symbols against the merged parent. The `ProducerUnresolved` blocker from 647bcc4b is still present, and the `leios_header_announcement_test.go` resolution kept the wrong comment. Merged with current `main` (#4161, #4295), build, vet and the replay-recovery, header-announcement and sqlstore tests pass apart from the pruned-fallback flag case.

ledger/replay_recovery.go
```diff
@@ -1667,6 +1686,15 @@ func (ls *LedgerState) findReplayRecoveryCandidate(
 		if fallbackCandidate != nil && (candidate == nil ||
 			fallbackCandidate.ProducerBlock.Slot < candidate.ProducerBlock.Slot) {
 			candidate = fallbackCandidate
+		} else if fallbackCandidate != nil {
```
chrisguiney: Blocker: this branch is dead code. It runs only when fallbackCandidate is non-nil and the first arm did not take, which is exactly what the condition on line 1694 already covers, so it sets a flag that is set again five lines later. Deleting lines 1689-1693 and re-running the Replay and Recover tests in ledger is green and leaves every selected candidate identical. Delete it, or fold it into a single site that also handles a nil fallback.

ledger/replay_recovery.go
```diff
@@ -1667,6 +1686,15 @@ func (ls *LedgerState) findReplayRecoveryCandidate(
...
+			// Keep the deeper known-producer anchor, but still rewind the
+			// primary chain because another input has unresolved provenance.
+			candidate.ProducerUnresolved = true
+		}
+		if candidate != nil && fallbackCandidate != nil {
```
chrisguiney: Blocker, still present from the previous review: keyed on fallbackCandidate being non-nil, but replayRecoveryFallbackCandidate has five nil,nil returns, three of them the pruned-tail cases this change adds. Five blocks at slots 80-160 with the first two deleted from the block store, a known producer at slot 140 plus one input with no metadata: securityParam 2, 3, 4 and 5 all yield strategy=metadata, rollback_slot=120, ProducerUnresolved=false. Line 239 then skips the primary rewind and with it the non-converging hold and resync escape. Key it on candidate, and pin that case.

ledger/leios_header_announcement_test.go
chrisguiney: Blocker, still present: this comment contradicts the subtest below it. The fixture sets no Mithril ledger slot and the subtest asserts trusted is false, while chainsyncHeaderCryptoPolicy returns trustedWithoutVerification only when the Mithril ledger slot is non-zero and covers the slot; this fixture takes the missing-epoch-nonce branch instead. The file is otherwise outside this change, so restoring the three-line comment from main removes it from the diff.

chrisguiney (changes requested):
Re-review at d7fbf293: the one new commit is a no-op. Its else-if branch is subsumed by the condition five lines below, and deleting it leaves every Replay/Recover test in ledger green, so the prior blocker stands. Cleared earlier and still clear: the pruned-tail fallback returns, the missing-parent warning, the Related to #4102 reference, the sqlstore conflict. Build, vet, lint, ledger tests, conformance, docs-parity and import-boundaries all pass on the tree merged with current main. go-test (Windows) fails identically on main pushes containing none of this branch.

ledger/replay_recovery.go
wolf31o2: Fixed. `findReplayRecoveryCandidate` takes the fallback only when it is strictly deeper, and otherwise keeps the known producer's anchor and sets the flag:

```go
if fallbackCandidate != nil && (candidate == nil ||
    fallbackCandidate.ProducerBlock.Slot < candidate.ProducerBlock.Slot) {
    candidate = fallbackCandidate
} else if candidate != nil {
    candidate.ProducerUnresolved = true
}
```

`TestReplayRecoveryKeepsPrimaryRewindForDeeperKnownProducer` pins it: one input resolving to a retained producer plus one with no metadata yields `strategy=metadata`, that producer's own parent as the rollback point (origin, since it is the first block), and `ProducerUnresolved` true. Deleting the `else if` arm fails that test and all four subtests of `TestFindReplayRecoveryCandidateFlagsUnresolvedWithoutLocalFallbackAnchor`.

ledger/replay_recovery.go
wolf31o2: Fixed at both points. `replayRecoveryFallbackCandidate` returns `nil, nil` when `BlockByIndex(targetIndex)` misses and when `replayRecoveryParentPoint(anchorBlock)` misses, instead of propagating `models.ErrBlockNotFound`.

Recovery declines in that case rather than anchoring where it cannot justify: with no retained anchor index and no retained parent, any rollback point is a guess across the retention or trust boundary. `findReplayRecoveryCandidate` then returns no candidate, `tryRecoverFromTxValidationError` returns false, and the original validation error surfaces.

`TestFindReplayRecoveryCandidateHandlesPrunedFallbackTail` covers both at the security parameters named here, 1 and 3, on a five-block chain at slots 80-160 with the first three blocks deleted. Removing either guard fails the matching subtest.

Auditing the rest of that class found two more lookups that still propagated `ErrBlockNotFound` out of `findReplayRecoveryCandidate`, both a dangling transaction-index reference like the parent case:

- `resolveReplayRecoveryProducer` on `database.BlockByHash(ls.db, producerTx.BlockHash)`
- `replayRecoveryBlockFromTxBlob` on `database.BlockByPoint(ls.db, point)`

`database.BlockDeleteTxn` removes a block's blob, its index entry and its hash index entry, but not the `transaction` rows or the tx blob offsets recorded against it, so both references outlive the block. Each now reports the input as unresolved; the metadata arm falls through to the tx blob first. `TestFindReplayRecoveryCandidateFallsBackWhenProducerBlockIsMissing` and `TestFindReplayRecoveryCandidateFallsBackWhenTxBlobBlockIsMissing` delete the producer block itself and require a `security-param-fallback` candidate with `ProducerUnresolved` set; each fails without its own guard, with `lookup producer block ...: block not found` and `lookup producer block from tx blob ...: block not found`.

`types.ErrHistoryExpired` is deliberately not folded in: a tombstoned block keeps its hash index so an archive proxy can still serve it, which is a different condition from an absent one.

ledger/replay_recovery.go
wolf31o2: Added. The missing-parent case logs `Warn` `replay recovery producer parent is missing from the local block store` with the producer block hash, the producer block slot and the absent parent hash, before the input is folded into `unresolvedInputs`. The two further dangling-index lookups found auditing that class carry the same signal: `replay recovery producer block is missing from the local block store` and `replay recovery tx blob names a block missing from the local block store`.

The strategy label is unchanged, so "no provenance recorded" and "the block store is missing a block the index still points at" remain distinguishable by the presence of one of those three warnings rather than by the label.

ledger/replay_recovery_test.go
wolf31o2: Fixed. `TestReplayRecoveryFallbackWinsOverKnownProducer` is gone. `TestReplayRecoveryKeepsPrimaryRewindForDeeperKnownProducer` replaces it and takes its known input's producer from `BlockByIndex(1)` of the audit fixture, whose parent point resolves without an `ErrBlockNotFound`, so the metadata candidate is built and the two candidates genuinely compete. It asserts `strategy=metadata` and that producer's slot, so the metadata candidate is the one that wins.

Restoring the previous predicate, that is deleting the `else if candidate != nil` arm, fails it on `candidate.ProducerUnresolved`.

ledger/replay_recovery.go
wolf31o2: Fixed. The assignment is keyed on `candidate != nil`, not on `fallbackCandidate != nil`, so it fires for every `nil, nil` fallback return as well:

```go
} else if candidate != nil {
    candidate.ProducerUnresolved = true
}
```

`TestFindReplayRecoveryCandidateFlagsUnresolvedWithoutLocalFallbackAnchor` pins the pruned-tail case at security parameters 2, 3, 4 and 5 on a five-block chain at slots 80-160 with the first two blocks deleted: a known producer at slot 140 plus one input without metadata yields `strategy=metadata`, rollback slot 120, and `ProducerUnresolved` true in all four. Deleting the arm fails all four subtests.

The field doc at `replayRecoveryCandidate.ProducerUnresolved` now reads "at least one failing input had no resolvable producer, whichever candidate supplied the rollback anchor. It gates the primary-chain rewind."

database/plugin/metadata/sqlstore/committee_prune_test.go
wolf31o2: Resolved. `origin/main` is merged at this head and `database/plugin/metadata/sqlstore/committee_prune_test.go` is byte-identical to it, so the file is out of the diff. The merged tree differs from `origin/main` in `ledger/replay_recovery.go`, `ledger/replay_recovery_test.go` and `ARCHITECTURE.md` only.

ledger/leios_header_announcement_test.go
wolf31o2: Resolved. `ledger/leios_header_announcement_test.go` takes `main`'s side and is byte-identical to it at this head, so the file is out of the diff.

ledger/replay_recovery.go
wolf31o2: Fixed at this head. The arm is keyed on `candidate != nil`, so a `nil, nil` fallback return no longer leaves `ProducerUnresolved` false when a known candidate survives.

`TestFindReplayRecoveryCandidateFlagsUnresolvedWithoutLocalFallbackAnchor` pins that topology on five blocks at slots 80-160 with the first two deleted and a known producer at slot 140, at security parameters 2, 3, 4 and 5: all four give `strategy=metadata`, rollback slot 120, and `ProducerUnresolved` true. Deleting the arm fails all four. The `ProducerUnresolved` field doc is updated with it.

ledger/leios_header_announcement_test.go
wolf31o2: Resolved. `ledger/leios_header_announcement_test.go` takes `main`'s side and is byte-identical to it at this head, so the file is out of the diff.

ledger/replay_recovery.go
wolf31o2: Fixed. There is a single assignment now, the `else if candidate != nil` arm; the duplicate is gone. Keyed on the surviving candidate rather than on `fallbackCandidate`, it also covers the `nil, nil` fallback returns that the deleted branch did not.

ledger/replay_recovery.go
wolf31o2: Fixed. There is one assignment site and it is keyed on `candidate`, so every `nil, nil` return from `replayRecoveryFallbackCandidate` is covered, all six of them.

`TestFindReplayRecoveryCandidateFlagsUnresolvedWithoutLocalFallbackAnchor` pins exactly the case named here, five blocks at slots 80-160 with the first two deleted, a known producer at slot 140 and one input with no metadata, at security parameters 2, 3, 4 and 5. Each yields `strategy=metadata`, rollback slot 120 and `ProducerUnresolved` true, so the primary rewind, the non-converging hold and the resync escape are reached. Deleting the arm fails all four subtests.

ledger/leios_header_announcement_test.go
wolf31o2: Resolved. The three-line comment from `main` is restored, so `ledger/leios_header_announcement_test.go` is byte-identical to `origin/main` at this head and no longer appears in the diff.

chrisguiney (approved):

PR #4194, opened 2026-09-11
fix(txsubmission): accept the reference size discrepancy

wolf31o2:
## Problem

`validateTxsubmissionReply` requires a returned body's size to match one
of two exact candidates (unwrapped body size or wrapped wire size).
ouroboros-network's TxSubmission V2 inbound side (`Inbound/V2.hs`
`checkTxSize`, `Inbound/V2/Types.hs`
`const_MAX_TX_SIZE_DISCREPANCY = 32`) instead tolerates a size
difference of up to 32 bytes in either direction. A peer whose
advertised size is off by a small, reference-tolerated amount has its
whole reply, and eventually its connection, rejected by Dingo.

## Fix

- `txsubmissionMaxSizeDiscrepancy` (32) and `txsubmissionSizeMatches`
  implement the reference's fuzzy comparison without unsigned
  underflow.
- The per-body check tries both exact sizes first (preserving the
  existing `accepted_wire_size`/`rejected` metric classification for
  the common exact case), then falls back to the fuzzy match.
- The aggregate predecode byte-budget check grows by the same
  per-body allowance, with overflow guards, so a tolerated reply is
  not rejected earlier as an aggregate overflow.
- `ARCHITECTURE.md` updated to describe the tolerance.

## Tests

- `TestValidateTxsubmissionReply/reference_size_discrepancy_boundaries`
  and `.../multi-body_discrepancy_budget_and_omission` cover the ±32
  boundary and the aggregate-budget interaction.
- `TestValidateTxsubmissionReplyRejectsGenuineSizeMismatch` and
  `TestValidateTxsubmissionReplyChecksByteBudgetBeforeDecode` widened
  to genuinely out-of-tolerance values.
- `TestTxSubmissionServerInitAcceptsReferenceSizeDiscrepancy` and
  `TestTxSubmissionServerInitRejectsOutOfRangeAdvertisedSize` drive the
  real relay harness with a peer that advertises sizes offset from the
  real body, observing admission and the actual rejection log line
  rather than a generic timeout.

`go build ./...`, `go vet ./ouroboros/...`, `golangci-lint run
./ouroboros/...` (0 issues), `make docs-parity` all pass.
`go test -race -count=1 ./ouroboros/...` passes in full (27.2s).

Related: blinklabs-io/gouroboros#2076

---
## Summary by cubic
`validateTxsubmissionReply` now accepts advertised body or wire sizes within 32 bytes of the actual size, matching ouroboros-network's TxSubmission V2 tolerance instead of requiring an exact match.

- The aggregate predecode byte-budget check includes the same per-body allowance with overflow guards.
- Exact-size matches are still classified first, so `accepted_wire_size`/`rejected` metrics keep their meaning for the common case.
- The relay admission path reserves headroom for the tolerance and rejects offers that cannot fit.
- Adds unit and relay-harness tests covering the ±32 boundaries, the aggregate-budget interaction (including out-of-tolerance under-advertisement), and end-to-end admission/rejection.

<sup>Written for commit 6daa30940de22f787e5d9437e591f2bb1ca96b55. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4194?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * TxSubmission replies now accept advertised body or wire sizes within 32 bytes of the actual size.
  * Aggregate byte-budget validation now accounts for the permitted discrepancy while preventing over-limit replies.
  * Relay handling now admits valid replies with small size discrepancies and rejects larger mismatches.
  * Improved validation ensures genuinely undersized, oversized, or inconsistent replies remain blocked.

* **Documentation**
  * Updated architecture documentation to describe the revised reply-size tolerance and budget rules.


ouroboros/txsubmission_test.go
```diff
@@ -728,6 +728,58 @@ func TestValidateTxsubmissionReply(t *testing.T) {
...
+	t.Run("reference size discrepancy boundaries", func(t *testing.T) {
+		bodySize := uint64(len(returned[0].TxBody))
+		wireSize := txsubmissionWireSize(returned[0].EraId, len(returned[0].TxBody))
+		tests := []struct {
+			name       string
+			advertised uint64
+			shouldPass bool
+		}{
+			{name: "raw minus 32", advertised: bodySize - 32, shouldPass: true},
+			{name: "raw minus 33", advertised: bodySize - 33, shouldPass: false},
```
chrisguiney, line 740: These two low-side cases don't reach `txsubmissionSizeMatches` — the aggregate byte-budget check rejects them first, so the lower bound of the new predicate is unpinned.

For a single-body reply, `remainingBytes` is `advertised + 32` and the per-body rule is `body <= advertised + 32`. The two bounds are the same number, and the aggregate loop runs first. Calling `validateTxsubmissionReply` directly with advertised 360 / body 393 returns:

```
txsubmission reply size mismatch: reply exceeds byte limit at index 0: advertised total 360, remaining 392, body 393, wire 400, era 6
```

not the per-body `size mismatch at index 0`. `require.ErrorIs` on the shared sentinel can't tell them apart.

Evidence that the branch is genuinely untested: replacing the `actual >= advertised` branch body in `txsubmissionSizeMatches` with `return true` leaves `go test -count=1 ./ouroboros/` green. The opposite branch is well covered — the same mutation there fails `TestTxSubmissionServerInitRejectsOutOfRangeAdvertisedSize` immediately.

The branch does matter; the aggregate budget only subsumes it when there's one body. A two-body skew clears the aggregate total and is caught only per-body:

```go
skew[0].Size = uint32(len(fixtures[0].body) - 200)
skew[1].Size = uint32(len(fixtures[1].body) + 200)
// -> txsubmission reply size mismatch at index 0: advertised 193, body 393, wire 400, era 6
```

Could a case like that be added here (or the low-side negatives given an extra requested-but-omitted transaction to pad the aggregate budget), so the per-body check is the one that fires?

ouroboros/txsubmission.go
```diff
@@ -211,6 +230,15 @@ func validateTxsubmissionReply(
 	return ret, nil
 }
 
+// txsubmissionSizeMatches implements Ouroboros Network's inclusive +/-32
+// advertised-size discrepancy without overflowing unsigned subtraction.
+func txsubmissionSizeMatches(advertised, actual uint64) bool {
```
chrisguiney, line 248: Accepting an advertised size below the body size invalidates the invariant stated at `ouroboros/txsubmission.go:552-555`:

> The advertised size is the wrapped wire size for a spec-conformant peer, so it is an upper bound on the body that will be admitted. Reserving against it is conservative by the few bytes of wrapper.

That held before this change — advertised was either `bodySize` or `wireSize`, and `txsubmissionWireSize` always adds at least 5 bytes, so `advertised >= body`. Now advertised can be `body - 32`.

Two consequences on the admission path:

1. The reservation at `:586` under-reserves by up to 32 bytes. Harmless on its own — `AddTransaction` enforces capacity and `retryTxsubmissionAdmission` handles the miss.
2. The impossible-offer guard at `:557` (`int64(txIds[0].Size) > headroom.MaxAdmissionHeadroomBytes()`) can now be cleared by a body up to 32 bytes *above* the admission limit. That body reaches `AddTransaction`, which returns `MempoolFullError` (`mempool/mempool.go:1561`, `newSize > rejectionThreshold`); `retryTxsubmissionAdmission` then calls `waitForAdmissionHeadroom(len(txBody.TxBody))`, which returns false on `minBytes > MaxAdmissionHeadroomBytes()` (`mempool/mempool.go:1758`); that yields `errTxsubmissionAdmissionStopped` and `:708` returns, ending the peer's pull loop instead of taking the backoff the guard was written for.

It needs a transaction body within 32 bytes above the mempool admission limit, so it's out of reach with a default `MempoolCapacity` against a 16 KiB max tx — but it is exactly the case `:557` exists to catch. Comparing against `Size + txsubmissionMaxSizeDiscrepancy` there would restore it; otherwise the comment at `:552-555` needs correcting so the next reader doesn't rely on it.

Separately, and much smaller: because the fuzzy wire case is tried before the fuzzy body case, almost every non-exact accept lands in the `accepted_wire_size` bucket. With a 7-byte wrapper the accepted window is `[body-32, wire+32]`, of which only `[body-32, body-26]` classifies as body — a peer advertising `body-3` is counted as having advertised the wire size. The exact cases still classify correctly, as the PR description says, so this only affects genuinely fuzzy peers; but `metrics_protocol.go:72` ("counts bodies advertised at the wrapped wire size") is now narrower than what the metric measures.

ouroboros/txsubmission_wire_size_test.go
```diff
@@ -163,14 +164,14 @@ func TestValidateTxsubmissionReplyRejectsGenuineSizeMismatch(t *testing.T) {
...
-		{name: "one above body", size: bodySize + 1, match: "size mismatch"},
-		{name: "one below wire", size: wireSize - 1, match: "size mismatch"},
-		{name: "one above wire", size: wireSize + 1, match: "size mismatch"},
+		{name: "beyond body tolerance above", size: bodySize + 40, match: "size mismatch"},
+		{name: "beyond wire tolerance below", size: wireSize - 40, match: "size mismatch"},
+		{name: "beyond wire tolerance above", size: wireSize + 33, match: "size mismatch"},
 		{
 			name:  "beyond wrapper overhead",
-			size:  bodySize + 8,
+			size:  bodySize + 40,
```
chrisguiney: `beyond wrapper overhead` is now `bodySize + 40`, identical to `beyond body tolerance above` on line 169, so it no longer asserts anything the case above doesn't.

At `bodySize + 8` it probed the adjacency between the two candidate sizes — just past the wrapper, still well below the wire size. Under the new tolerance that value is accepted, so the case has lost its subject rather than needing a bigger number. Either give it a value that still tests something distinct or drop it.

While in this file: the comment on line 164-166 says an under-advertisement "must be classified as [a size mismatch] rather than falling through to the aggregate byte-budget error", but `bodySize - 33` does fall through to the aggregate error. That was already true on main for `bodySize - 1`, so it isn't something this PR introduced — just worth fixing while the lines are being touched.

chrisguiney (dismissed):
Verified the tolerance against ouroboros-network `09495ae9` (`const_MAX_TX_SIZE_DISCREPANCY = 32`, `checkTxSize` inclusive and symmetric) — `txsubmissionSizeMatches` is a faithful translation, and the new tests do fail with `ouroboros/txsubmission.go` reverted to main. `go build`, `go vet`, `golangci-lint run ./ouroboros/...`, `go test -race -count=1 ./ouroboros/...`, `make docs-parity` and `make import-boundaries` all pass at this head.

Two things to address: the under-advertisement half of the new predicate is not pinned by any test (a mutation of it leaves the whole package green), and the change invalidates an invariant `txsubmission.go:552-555` states and the admission-headroom guard relies on. Details inline.

ouroboros/txsubmission.go
```diff
@@ -222,6 +244,15 @@ func validateTxsubmissionReply(
...
+	if actual >= advertised {
+		return actual-advertised <= txsubmissionMaxSizeDiscrepancy
```
arepala-uml, line 250: The under-advertisement branch (`actual >= advertised`) isn't exercised by any test that isolates it from the aggregate byte-budget pre-check (~lines 130-148), which enforces the same +/-32*N bound before this per-body predicate runs. Flipping this branch to always return `true` leaves `go test ./ouroboros/...` fully green (reproduced locally). Concrete gap: a two-body reply where one body is under-advertised by 60 bytes (outside tolerance) and another by 4 bytes keeps the aggregate sum in budget, so only this per-body check catches it, and no current test covers that shape. Please add a multi-body case that isolates this branch before merging.

ouroboros/txsubmission.go
```diff
@@ -123,13 +127,25 @@ func validateTxsubmissionReply(
...
 	requestedIndex := make(map[[32]byte]int, len(requested))
 	for index, requestedTx := range requested {
-		requestedBytes += uint64(requestedTx.Size)
+		requestedSize := uint64(requestedTx.Size)
+		if requestedSize > math.MaxUint64-requestedBytes {
+			return nil, fmt.Errorf("%w: advertised byte budget overflow", errTxsubmissionReplySizeMismatch)
+		}
+		requestedBytes += requestedSize
+		// Preserve the first announcement if an ID appears more than once.
 		// Preserve the first announcement if an ID appears more than once.
```
arepala-uml: Duplicate comment line from the merge commit (main already has this once)

arepala-uml (changes requested):
Reviewed at current head e874e09b (three commits past the previously reviewed e0f04796).

Admission-headroom invariant: fixed. `validateTxsubmissionReply` now reserves `advertised+32` bytes as the worst case before checking headroom, matching the updated invariant comment and ARCHITECTURE.md.

Under-advertisement tolerance: still untested at this head — see inline comment. Requesting changes for that gap; the duplicate comment line is a minor, non-blocking cleanup.

arepala-uml (approved):

PR #4195, opened 2026-09-11
fix(ledger/snapshot): retain reward_account_output for koios-parity

### LINKED ISSUE #4188: koiosparity: Dingo computes no per-account rewards for epochs 9-22, though Koios has real reward data from epoch 9
## Summary

A from-genesis Preview replay (`--mithril-enabled=false`, `--koios-parity-enabled --koios-parity-strict=false --koios-parity-base-url https://preview-koios.tosidrop.me/api/v1`) on latest `origin/main` (`23f69ed34`) shows Dingo's `reward_account_output` table has **zero rows for every epoch from 0 through 22**, and only starts populating at **epoch 23** (184 rows) and 24 (155 rows).

Koios has real reward data well before that: `koios_account_rewards` (this run's fetched cache) has 1 row at epoch 9 and 37 rows at epoch 10, all real leader/member amounts in the hundreds of millions of lovelace, e.g.:

```
epoch=10 stake_address=stake_test1up8dnh6ywzs96upw6ycmeafzmkvrv096njjkphgqzuts0wgnxsaqg
  reward_type=leader earned=372133077 spendable_epoch=12
  pool_id_bech32=pool1pn9sffcqqzkx70m0gujks4h3wf8p4y706t2f0cjcyreekg83wtf
```

For that same stake address and every other of the 37, Dingo's `reward_account_output` has **no row at all** for epoch 10 (or any epoch before 23).

## Effect on koios-parity

Epochs 0-9 pass (or fail only on the known, unrelated #4127 cost-model divergence) because Koios also has no/zero reward data for those accounts in that range — a "both sides empty" agreement. Epoch 10 is the first epoch where Koios has real reward data and the gap becomes visible:

```
level=ERROR msg="koiosparity observer: epoch validation failed" network=preview epoch=10
  error="parity FAIL at epoch 10 (37 significant of 38 mismatch(es))" strict=false
```

All 37 significant mismatches are `category=acct_only_koios, field=account_reward_presence` — Koios has a real amount, Dingo returns nothing (not a zero-value row, no row).

## Distinct from other open koios-parity issues

- #4127 is a single-field cost-model mismatch confined to epochs 3-8; unrelated.
- #3841 and #4130 are about `acct_only_dingo` (Dingo has a row Koios doesn't, e.g. withheld/deregistered rewards being wrongly flagged) — the opposite direction, and about mis-flagging *correct* Dingo behavior. This issue is Dingo having no data at all where real reward data should exist.
- #1959 ("full reward calculation per Shelley spec Sec 10") is the broad epic tracking reward-calculation completeness. This issue is a concrete, reproducible, epoch-bounded instance: computation does not appear to run at all for epochs 9-22, then resumes with plausible output at epoch 23. It is not that rewards are never computed — they start again with a normal-looking row count (184, 155) right after the gap — which suggests a boundary condition rather than an entirely missing code path.

## Reproduction

```
dingo -n preview serve --mithril-enabled=false --koios-parity-enabled \
  --koios-parity-strict=false --koios-parity-base-url <preview koios host>
```

From genesis. Confirm directly against Dingo's own `metadata.sqlite`:

```sql
SELECT epoch, COUNT(*) FROM reward_account_output GROUP BY epoch ORDER BY epoch;
-- only epoch 23 and 24 present in this run at time of writing (still syncing)
```

and against the koios-parity cache:

```sql
SELECT epoch, COUNT(*) FROM koios_account_rewards GROUP BY epoch ORDER BY epoch;
-- 9|1
-- 10|37
```

Or via the CLI: `koios-parity explain --network preview --epoch 10 --json`.

## Open questions for whoever picks this up

- Is epoch 23 significant (a snapshot-maturity boundary, a specific slot/era transition, or the first epoch after some fixed startup delay independent of chain content)? The from-genesis timing (epoch 23 on a fresh replay) suggests it may be a fixed epoch-count delay rather than a chain-content-driven boundary — worth checking whether it reproduces at the same epoch number on a different Mithril bootstrap boundary.
- Whether epochs 9-22 are permanently unrecoverable (never backfilled) or whether continued sync eventually reconciles them — this run was still syncing at time of filing and had not reached a later epoch to check.

## Environment

`origin/main` at `23f69ed34`, includes #3838, #4175, and everything else merged as of 2026-09-11. Live run in progress at the time of filing; data directory preserved.

chrisguiney:
`cleanupOldSnapshots` prunes `reward_account_output` to a fixed
`current-3..current` window at every epoch transition in CORE storage mode. The
in-process koios-parity observer validates a closed epoch only after fetching
and comparing against Koios over the network, so on a from-genesis or catch-up
sync it reaches an epoch well after chain progression has evicted that epoch's
rows. Every per-account check for it then fails as `acct_only_koios` against a
row that genuinely no longer exists (#4188).

- `Manager.SetRewardAccountOutputRetentionUnbounded`, wired from
  `KoiosParityConfig.Enabled` in `node.go` (`Run`) and `node_lifecycle.go`
  (`reinitializeBackgroundManagers`), makes `cleanupOldSnapshots` retain
  `reward_account_output` without bound. It reuses the branch API storage mode
  already takes for the Blockfrost reward-history endpoint (#1875).
  `reward_stake_input` and `pool_stake_snapshot` stay pruned to the window in
  every mode.
- Retention is not consensus-affecting, so unlike `SetDelegatorInactivity` the
  setter carries no `configurationLocked` gate and applies after `Start`.
- A koios-parity-enabled CORE node therefore keeps `reward_account_output` for
  its whole history, the same growth profile API mode already accepts. No
  backfill: epochs already pruned before the upgrade stay absent.
- `DATABASE.md` and `ARCHITECTURE.md` record the new retention exception;
  `internal/koiosparity`'s source comments described the pre-fix rule and the
  disproved "observer timing keeps reads inside the window" model.

`TestCleanupOldSnapshotsKoiosParityRetentionUnbounded` fails (epoch 0 rows
deleted) with the new disjunct removed from `cleanupOldSnapshots`. Two
source-level guards in `node_snapshot_config_wiring_test.go` fail with either
`SetRewardAccountOutputRetentionUnbounded` call site deleted; the first also
covers any future `n.snapshotMgr` setter added to `Run` alone.

Fixes #4188. Related to #3857 (absent reward rows misreported during replay),
which this does not close.

---
## Summary by cubic
Fixes koios-parity account checks permanently failing on closed epochs: `cleanupOldSnapshots` prunes `reward_account_output` to a 4-epoch window at every epoch transition, and the observer only validates an epoch after a network comparison with Koios that can fall far behind during catch-up syncs. When koios-parity is enabled, the table is now retained without bound, matching API storage mode.

- Adds `Manager.SetRewardAccountOutputRetentionUnbounded`, wired from `KoiosParityConfig.Enabled` in both `Run` and `reinitializeBackgroundManagers`.
- `reward_stake_input` and `pool_stake_snapshot` are still pruned to the window; retention is not consensus-affecting, so no `configurationLocked` gate.
- No backfill: epochs already pruned before the upgrade stay absent.
- Adds a regression test and source-level guards that both startup paths keep the same snapshot-manager configuration.

Fixes #4188.

<sup>Written for commit 782caf6f1497f6348ea3f33169d9a5c8ebb3ec3d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4195?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Koios parity observer mode now retains reward-account outputs without a fixed four-epoch limit, including during catch-up synchronization and lifecycle restores.
  - Reward-stake inputs continue to follow the existing pruning window.

- **Documentation**
  - Updated architecture, database, and developer documentation to clarify reward-state retention behavior across storage modes and observer configurations.

- **Tests**
  - Added coverage confirming retention behavior and consistent configuration across startup and restoration paths.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -7080,6 +7080,25 @@ second sync:
   permanent Dingo subsystem — SQLite (or whichever metadata backend the node
   itself runs) remains the only backend involved, since `DatabaseSource`
   reads the live node's own store rather than opening a second connection.
+  - **Reward-state retention (dingo #4188).** `Run()` and
+    `reinitializeBackgroundManagers` both call
+    `n.snapshotMgr.SetRewardAccountOutputRetentionUnbounded(
+    n.config.koiosParity.Enabled)` immediately after configuring the
+    snapshot manager's other options, before `CaptureGenesisSnapshot`/`Start`.
+    This is necessary because the observer only validates a closed epoch
+    after fetching and comparing against Koios over the network — work that
+    can fall arbitrarily far behind chain progression during a from-genesis
+    or catch-up sync — while `ledger/snapshot.cleanupOldSnapshots` otherwise
+    prunes `reward_account_output` to a fixed 4-epoch window on every epoch
+    transition (see DATABASE.md, Snapshot and Reward-State Retention).
+    Without this, an epoch's reward rows are routinely gone by the time the
+    observer's backlog reaches that epoch, and every koios-parity account
+    check for it fails permanently with a row that genuinely no longer
+    exists rather than one Koios and Dingo actually disagree on. The setter
+    mirrors API storage mode's pre-existing unbounded retention (#1875) and
+    is not consensus-affecting (it only widens local historical retention),
+    so unlike `SetDelegatorInactivity` it carries no `configurationLocked`
+    gate.
   - **Live database Restore/Truncate.** `node_lifecycle.go`'s
     `quiesceForLiveLifecycleOp` stops the `Observer` (blocking until its
     background goroutine has exited, same as `shutdown()`) and unsubscribes
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -1063,18 +1063,32 @@ leader-threshold rejection.
 Every epoch transition runs `cleanupOldSnapshots`, which prunes to the four
 epochs the Shelley rotation and delayed reward model need: current, current-1,
 current-2 for Go, and current-3 so reward calculation can be replayed after a
-rollback across the boundary where those rewards were applied. Retention is not
-operator-configurable, and it only ever deletes rows below that window (but see
-the deferred-header retention pin below, which can hold `pool_stake_snapshot`
-rows longer).
+rollback across the boundary where those rewards were applied. It only ever
+deletes rows below that window (but see the deferred-header retention pin
+below, which can hold `pool_stake_snapshot` rows longer).
+
+`reward_account_output` is the one exception to "not operator/config
+configurable": it is retained WITHOUT BOUND, instead of pruned to the window
+below, whenever the database is in API storage mode (`types.StorageModeAPI`,
+issue #1875, so the Blockfrost account reward-history endpoint can serve an
+account's full history) or the in-process Koios parity observer is enabled
+(`Manager.SetRewardAccountOutputRetentionUnbounded`, wired from
+`KoiosParityConfig.Enabled` in `node.go`/`node_lifecycle.go`, issue #4188). The
+observer validates a closed epoch only after fetching and comparing against
+Koios over the network, which can fall arbitrarily far behind chain
+progression during a from-genesis or catch-up sync — well past the fixed
+4-epoch window — so without this a checked epoch's `reward_account_output`
+rows are routinely gone before the observer ever reads them.
+`reward_stake_input` is pruned to the window in every case, including both of
+the above.
 
 Pruned at `epoch < current-3`:
 
 | Table | Scales with | Pruned by |
 |---|---|---|
 | `pool_stake_snapshot` | pools per epoch | `DeletePoolStakeSnapshotsBeforeEpoch` |
-| `reward_stake_input` | delegators per epoch | `DeleteRewardStateBeforeEpoch` |
-| `reward_account_output` | delegators per epoch | `DeleteRewardStateBeforeEpoch` |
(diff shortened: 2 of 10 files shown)


chrisguiney:
Pre-publication review disposition for 782caf6.

- Fail-before re-verified by reverting `cleanupOldSnapshots`'s new disjunct in
  place: `TestCleanupOldSnapshotsKoiosParityRetentionUnbounded` fails on epoch 0
  retention while the API-mode sibling test still passes; restoring it passes.
- Class audit of `reward_account_output` consumers. The precomputed-reward reuse
  path in `ledger/reward_calculation.go` is gated on `reward_stake_input`, which
  stays pruned to the window, so retained older rows cannot feed it; the only
  cross-epoch reader is the API-mode Blockfrost endpoint. The third
  `snapshot.NewManager` site, `internal/node/load.go`, builds the manager with a
  nil event bus and never calls `Start`, so `cleanupOldSnapshots` never runs
  there and it needs no equivalent wiring.
- `pool_stake_snapshot` is still pruned to the window, so the observer's
  pool-departure evidence loss (`internal/koiosparity/check.go`) is unchanged by
  this change.
- A live from-genesis preview sync against a Koios host was not run, so the
  end-to-end acceptance evidence for #4188 is still owed; the retention
  mechanism itself is covered by unit tests.

arepala-uml (approved):

PR #4198, opened 2026-09-11
feat(config): embed prime testnet configs

wolf31o2:
Embed the Prime testnet Cardano configuration so Dingo can start against Prime without external config files.

- Add the upstream `configuration.yaml`, topology, and Byron, Shelley, Alonzo, and Conway genesis files.
- Resolve each embedded network's actual configuration filename through `EmbeddedConfigPath`.
- Cover every embedded network through the derived path.
- Pin Prime's computed genesis hashes in test source while preserving byte-for-byte parity with the upstream cardano-configs image.
- Extend the generic line-ending and committed-byte checks to nested genesis paths.


config/cardano/prime-testnet/configuration.yaml
```diff
@@ -0,0 +1,320 @@
...
+
+##### Locations #####
+
+ByronGenesisFile: genesis/byron/genesis.json
+ByronGenesisHash: acd3e7c4cc071ae176fa253e654f0dd4e759892b16e958d8bd3c393f5c6c66d2
+ShelleyGenesisFile: genesis/shelley/genesis.json
+ShelleyGenesisHash: 0b67b0cb16e973478888efb1c7f69fca8e80489369afed94624a588e94a94f2b
+AlonzoGenesisFile: genesis/shelley/genesis.alonzo.json
+AlonzoGenesisHash: 0f715ff78fcab739c6cb2cc1073eb87d0ffe3514218bfa0337d416f6fbaf7886
+ConwayGenesisFile: genesis/shelley/genesis.conway.json
```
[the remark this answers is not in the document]
wolf31o2: The hash is correct: `01649daebfc1ae5646c6aab2bf07533cdf626fd49e270ac4258be7fe253c3cc1` is the Blake2b-256 of the normalized `genesis.conway.json` embedded here, and `validateGenesisHash` does skip the comparison on an empty expected value, so that genesis is loaded unverified.

Not applied in this repository. `config/cardano/` is a verbatim copy of `blinklabs-io/docker-cardano-configs`, enforced by `bin/config-parity.sh` against the image tag pinned in the Dockerfile, and upstream's `prime-testnet/configuration.yaml` carries no `ConwayGenesisHash`. Adding the line here reports `1 file(s) differ from ghcr.io/blinklabs-io/cardano-configs` and exits non-zero. The pin belongs in `docker-cardano-configs`, followed by a released image tag and a Dockerfile bump in the same commit here, which is the sequence `bin/config-parity.sh` prints on failure.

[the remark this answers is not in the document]
wolf31o2: The hash is correct: `01649daebfc1ae5646c6aab2bf07533cdf626fd49e270ac4258be7fe253c3cc1` is the Blake2b-256 of the normalized `genesis.conway.json` embedded here, and `validateGenesisHash` does skip the comparison on an empty expected value, so that genesis is loaded unverified.

Not applied in this repository. `config/cardano/` is a verbatim copy of `blinklabs-io/docker-cardano-configs`, enforced by `bin/config-parity.sh` against the image tag pinned in the Dockerfile, and upstream's `prime-testnet/configuration.yaml` carries no `ConwayGenesisHash`. Adding the line here reports `1 file(s) differ from ghcr.io/blinklabs-io/cardano-configs` and exits non-zero. The pin belongs in `docker-cardano-configs`, followed by a released image tag and a Dockerfile bump in the same commit here, which is the sequence `bin/config-parity.sh` prints on failure.

config/cardano/prime-testnet/genesis/shelley/genesis.alonzo.json
```diff
@@ -0,0 +1,365 @@
...
+      1021,
+      10,
+      38887044,
+      32947,
+      10
+    ]
+  },
+  "executionPrices": {
+    "prMem": 5.77,
+    "prSteps": 0.00721
```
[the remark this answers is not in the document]
wolf31o2: The arithmetic observation holds: every other network embedded here carries 577/10000 and 721/10000000, and prime-testnet carries 5.77 and 0.00721.

It is not actionable as a change to this file. The embedded copy is byte-identical to `ghcr.io/blinklabs-io/cardano-configs:20260910-1`, checked by `bin/config-parity.sh`, and it is covered by the pinned `AlonzoGenesisHash: 0f715ff78fcab739c6cb2cc1073eb87d0ffe3514218bfa0337d416f6fbaf7886` in the same `configuration.yaml`. Applying the proposed values makes config loading fail outright:

```
Alonzo genesis hash mismatch: expected 0f715ff78fcab739c6cb2cc1073eb87d0ffe3514218bfa0337d416f6fbaf7886,
computed a019c67893b4499350d3c667e61aa0c71ae662d5f1d26eb5105a9ebdb2db72d0
```

A genesis file has to match the chain it describes, not the other networks, and the only evidence available here says this is the file that chain's hash covers. If the value is wrong on prime-testnet itself, the correction belongs to that network's genesis and then to `docker-cardano-configs`; it cannot be made by editing a hash-pinned copy.

config/cardano/embed.go
```diff
@@ -17,9 +17,10 @@ package cardano
...
-// for all supported networks (preview, preprod, mainnet, devnet, musashi).
-// This includes config.json and all genesis files required for network
-// operation when no external config files are available.
+// for all supported networks (preview, preprod, mainnet, devnet, musashi,
+// prime-testnet). This includes the network configuration and all genesis
+// files required for network operation when no external config files are
+// available.
 //
-//go:embed preview mainnet preprod devnet musashi
+//go:embed preview mainnet preprod devnet musashi prime-testnet
```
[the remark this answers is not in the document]
wolf31o2: Confirmed and fixed in b94919a3. `LoadCardanoNodeConfigWithFallback` stats the derived path, misses, then fails on the embedded filesystem with `open prime-testnet/config.json: file does not exist`, so the network could not be started from the embedded files.

Fixed by teaching the derivation rather than renaming the file: `config/cardano/embed.go` gains `EmbeddedConfigPath(network)`, which resolves the name against `EmbeddedConfigFS` and falls back to the conventional `config.json` for a network that is not embedded at all. Renaming was not available: `bin/config-parity.sh` requires the tree to match `docker-cardano-configs`, where the file is `configuration.yaml`.

Class check: the derivation appeared at four production sites, not one. All four now call the helper -- `cmd/dingo/serve.go:224`, `internal/node/node.go:180`, `internal/node/load.go:554`, `mithril/sync.go:374` -- and `genesis_consistency_test.go` drops the `prime-testnet` special case it needed.

`TestEmbeddedConfigPathReachesEveryNetwork` enumerates the embedded directories rather than a name list, so a network added to the `//go:embed` line is covered automatically, and loads each one at its derived path. With the resolution removed it reports prime-testnet as not loadable with the error above.

config/cardano/prime-testnet/configuration.yaml
```diff
@@ -0,0 +1,320 @@
+##########################################################
+###############          Mainnet                 #########
```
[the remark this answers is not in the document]
wolf31o2: The banner is inaccurate, and it is upstream's text. `config/cardano/` is a verbatim copy of `blinklabs-io/docker-cardano-configs`; editing the header here makes `bin/config-parity.sh` report the file as differing from the pinned `cardano-configs` image and exit non-zero. The wording belongs in `docker-cardano-configs`, then a released image tag and a Dockerfile bump in the same commit here.

config/cardano/prime-testnet/configuration.yaml
chrisguiney, line 15: ConwayGenesisFile is declared with no ConwayGenesisHash, and validateGenesisHash in config/cardano/node.go compares nothing when the expected value is empty, so this is the only embedded genesis outside devnet that loads unverified; preview, preprod, mainnet and musashi each declare four. The file is byte-identical to the pinned cardano-configs image, so it cannot be edited here, but the computed value 01649daebfc1ae5646c6aab2bf07533cdf626fd49e270ac4258be7fe253c3cc1 can be pinned in test source the way #4279 pins the other networks.

config/cardano/embed_test.go
```diff
@@ -170,3 +171,96 @@ func TestEmbedFS_AllNetworks(t *testing.T) {
...
+// TestEmbeddedConfigPathReachesEveryNetwork is the defect the helper exists
+// for: every caller that has only a network name derived the config path by
+// appending "/config.json", and prime-testnet's upstream config is named
+// configuration.yaml, so that network could not be started from the embedded
+// filesystem at all.
+//
+// Enumerating the embedded directories rather than listing names keeps a
+// network added to the //go:embed directive under this check automatically --
+// the same property bin/config-parity.sh relies on.
+func TestEmbeddedConfigPathReachesEveryNetwork(t *testing.T) {
```
chrisguiney, line 218: #4279 adds TestEveryEmbeddedNetworkIsPinnedOrExempt, requiring every directory in EmbeddedConfigFS to be pinned or exempted; running that file against this branch fails with 'embedded network prime-testnet has no pinned genesis hashes and no recorded exemption'. Neither pull request's checks can see it, since each is green against a base without the other. Pinning prime-testnet here settles both, and #4279's TestEmbeddedNetworksDeclareGenesisHashes derives network plus /config.json so it needs EmbeddedConfigPath to reach this network.

config/cardano/embed.go
```diff
@@ -17,9 +17,45 @@ package cardano
...
+// Callers deriving a default path must use this rather than appending
+// "/config.json": a network whose upstream config is named otherwise is then
+// simply unreachable from the embedded filesystem, and the failure surfaces as
+// "no embedded config available" with nothing naming the file name as the
+// cause.
+//
+// A network with no embedded directory at all gets the conventional
+// "<network>/config.json", so the error a caller reports names the file it
+// would have expected rather than the last name this happened to try.
+func EmbeddedConfigPath(network string) string {
```
chrisguiney, line 49: Verified causally: replacing this function body with a plain network plus /config.json makes TestEmbeddedConfigPathReachesEveryNetwork/prime-testnet and TestEmbeddedConfigsPassGenesisConsistency/prime-testnet fail, so the new coverage is synchronised with the fix rather than passing either way. The class is complete: no remaining caller derives a config path by appending config.json, and the mithril/sync.go change also drops the filepath.Join backslash form that embed.FS cannot use.

chrisguiney (dismissed):
prime-testnet declares ByronGenesisHash, ShelleyGenesisHash and AlonzoGenesisHash but no ConwayGenesisHash, so validateGenesisHash skips verification of the one genesis file nothing else pins; every other embedded network except devnet declares all four. bin/config-parity.sh confirms all six embedded networks are byte-identical to cardano-configs:20260915-1 after merging current main, and that merged tree passes build, vet, config/cardano, cmd/dingo, internal/node, mithril, docs-parity, import-boundaries and golangci-lint.

config/cardano/prime-testnet/configuration.yaml
wolf31o2: Fixed in 7daa3baa. Prime's Byron, Shelley, Alonzo, and computed Conway hashes are now pinned in genesis_hash_test.go; mutating the Conway pin produced the expected named hash mismatch.

config/cardano/embed_test.go
wolf31o2: Fixed in 7daa3baa after the additive main merge. Prime is included in pinnedEmbeddedGenesisHashes, every generic config load uses EmbeddedConfigPath, and the line-ending and committed-byte checks now follow each config's referenced nested genesis paths.

config/cardano/embed.go
wolf31o2: Confirmed. Current main was merged additively and the combined config/cardano race suite passes with the shared EmbeddedConfigPath coverage.

cryptodj413 (approved):

Issue #4203, opened 2026-09-12
ledger: block-application throughput bottleneck during from-genesis sync — serial per-input UTxO lookups and under-scaled decode/validate pipeline


chrisguiney:
## Summary

A from-genesis Preview sync with `--koios-parity-enabled` shows raw chain-sync
throughput bottlenecked by ledger block-application, not network/disk/CPU/memory:
~150-450 slots/sec on a 16-core host at only ~7-8% total CPU utilization
(matches one core near-saturated: 114% CPU / 16 cores = 7.1%, measured on the
live instance below). The log carries repeated `"failed to queue header"` /
`"header queue at maximum capacity"` warnings — 332 occurrences over a ~4h21m
run as of this writing, all during fork-resolution churn against a single peer
(`fork_reason="chainsync rollback"`).

This is a fresh investigation opened at the requester's direction, not a fix
for #4102 (a separate, already-tracked defect with its own fix in #4192).

## The queue-full events are not the old liveness bug

`ARCHITECTURE.md` ("Block Processing Pipeline" section) already documents a
fixed liveness bug with the identical log signature
(`ensureBlockfetchDrainingAfterForkQueueFailure`, `ledger/chainsync.go`): before
that fix, a header-queue-full failure during fork resolution could permanently
stop blockfetch from restarting, wedging the pipeline until a 4-minute
watchdog forced a resync. That fix is present on `origin/main` (verified: 6
references to `ensureBlockfetchDrainingAfterForkQueueFailure` in
`ledger/chainsync.go`) and is what both the requester's live instance and my
own isolated instance are running.

So the queue-full events observed here are the *bounded, self-healing*
behavior the existing fix produces, not the old permanent stall. But
ARCHITECTURE.md's own text is explicit that the underlying mismatch is real
and separate from that liveness fix: *"the validate stage's extra CPU cost...
makes the underlying throughput mismatch easier to hit in practice, but is not
what causes it"* — i.e., header arrival can outrun block-apply throughput
during bursty fork conditions regardless of the pipeline feature flags. That
throughput mismatch, and why block-apply is slow enough for this to happen at
all, does not appear to have its own tracking issue, which is what this issue
is for.

## Evidence

- Live instance (`internal/config` defaults, no `--block-pipeline-enabled`),
  from genesis, `--koios-parity-enabled --koios-parity-strict=false`: 332
  header-queue-full warnings over ~4h21m, applied tip at slot ~7.09M, peer tip
[...]

Issue #4067, opened 2026-09-07
Consumed-UTxO cleanup starves block application: 400 OR-terms in queryUtxoStakeRefs defeat tx_id_output_idx, scanning all 11M live UTxOs per batch


wcatz:
Consumed-UTxO cleanup starves block application: 400 OR-terms in `queryUtxoStakeRefs` defeat `tx_id_output_idx`, scanning all 11M live UTxOs per batch

## Summary

On a mainnet node with ~11M live UTxOs, the periodic consumed-UTxO cleanup holds the
single SQLite write connection for many minutes at 100% of a core, blocking block
application entirely. The node syncs normally for ~5 minutes, the cleanup timer fires,
and the applied tip then freezes until the cleanup finishes.

The cause is the query shape in `queryUtxoStakeRefs`
(`database/plugin/metadata/sqlstore/utxo.go:665-673`):

```go
// Two bind variables per reference; 400 keeps this portable to SQLite's
// conservative 999-parameter configuration.
for start := 0; start < len(ids); start += 400 {
    end := min(start+400, len(ids))
    predicate, args := utxoIDPredicate(ids[start:end])
    query := "SELECT DISTINCT credential_tag, staking_key FROM utxo WHERE (" +
        predicate + ")"
    if liveOnly {
        query += " AND deleted_slot = 0"
    }
```

with `utxoIDPredicate` (`utxo.go:714-720`) emitting `(tx_id = ? AND output_idx = ?)`
OR-ed per reference.

The batch size of 400 was chosen for SQLite's 999-parameter limit. But the number of
OR-terms also controls whether SQLite's `MULTI-INDEX OR` optimization applies, and that
threshold is far lower. Measured against the live 11 GB mainnet database
(`EXPLAIN QUERY PLAN`, modernc.org/sqlite v1.56.0, `utxo` = 11,226,026 rows,
10,965,310 of them live):

```
    2 terms: tx_id_output_idx lookups=2     fallback=no
    4 terms: tx_id_output_idx lookups=0     fallback=YES
    8 terms: tx_id_output_idx lookups=0     fallback=YES
   16 terms: tx_id_output_idx lookups=0     fallback=YES
   32 terms: tx_id_output_idx lookups=0     fallback=YES
   64 terms: tx_id_output_idx lookups=0     fallback=YES
  100 terms: tx_id_output_idx lookups=0     fallback=YES
  200 terms: tx_id_output_idx lookups=0     fallback=YES
[...]

PR #4204, opened 2026-09-12
perf(ledger): scale block-pipeline decode/validate workers to GOMAXPROCS

chrisguiney:
## Summary

Investigates the block-application throughput bottleneck described in #4203
(from-genesis Preview sync, `--koios-parity-enabled`, ~150-450 slots/sec on a
16-core host at only ~7-8% CPU, with repeated header-queue-saturation
warnings during fork-resolution churn). Full evidence, root-cause analysis,
and a second lever investigated but deliberately not shipped are in #4203.

This PR ships one safe, narrowly-scoped fix from that investigation:

- **`perf(ledger): scale block-pipeline decode/validate workers to
  GOMAXPROCS`** — `blockPipelineDecodeWorkers`/`blockPipelineValidateWorkers`
  were a hardcoded 2 regardless of host CPU count, for the optional block
  pipeline from issue #1894 (`BlockPipelineEnabled`/
  `BlockPipelineValidateEnabled`, both default off). A CPU profile taken
  during this investigation showed a single core saturated (SQLite VM
  execution + Ed25519 verification) while 15 of 16 cores sat idle.
  `blockPipelineWorkerCount()` now derives both stages' worker count from
  `runtime.GOMAXPROCS(0)`, floored at the prior fixed count (2, so a
  constrained host never regresses) and capped at 8 (so decode/validate does
  not run arbitrarily far ahead of `ledgerProcessBlocksFromSource`'s
  single-threaded apply consumer). Both stages are already documented
  (`internal/config/config.go`) as scheduling-only, non-consensus-affecting
  knobs. The feature stays opt-in, so this changes nothing for a default
  deployment.
- **`test(ledger): stop asserting a wall-clock speedup ratio`** — the added
  throughput test originally required the scaled worker count to beat the
  fixed count by 30% or more wall-clock, which failed on the `go-test
  (macOS)` CI runner (measured 0.99x, where available parallelism itself
  bounds the possible ratio on a constrained runner). The test now asserts
  the correctness property that actually matters here — every block still
  decodes and VRF/KES-validates cleanly at the scaled worker count — and
  reports throughput at both counts without gating on the ratio. The
  worker-count regression (silently reverting to a fixed value) stays pinned
  deterministically by the unaffected `TestBlockPipelineWorkerCount`.

**What this PR does not do:** it does not eliminate the header-queue
saturation symptom on a default configuration (neither live instance runs with
`--block-pipeline-enabled`), and it does not touch the larger, riskier lever
(batching `LedgerView.UtxoById`'s per-input database round trips) — that fix
was implemented and fully tested, then backed out because it would reuse the
exact defective SQL query shape already reported as catastrophic at scale in
#4067. See #4203 for the full writeup and the reason it's held for a human to
sequence after #4067 lands rather than shipped here.

## Base and isolation

This branch is `origin/main` (`83d8364c3`) merged with PR #4192
(`fix/continuation-audit-4102`, `34a7b5f04`, not yet merged to `main`) so an
isolated from-genesis sync run used for this investigation would not trip
over #4102's already-tracked/fixed noise. **The commit list therefore
includes #4192's commits until that PR merges to `main`** (merge commit
`f7e54d0e2`); the commits scoped to this PR are:

- `f7e54d0e2` — merge PR #4192 into this branch's base (isolation
  housekeeping, not a code change of its own)
- `2ee914bb4` — the worker-count scaling fix described above
- `7ba47bd47` — the throughput-test flakiness fix described above

`main` has since advanced past this branch's base with unrelated merged
work, which now shows this PR as having a merge conflict against `main`.
The conflicting hunks are confined to `ARCHITECTURE.md` and a struct-field
region of `ledger/state.go` inherited from PR #4192's content (adjacent,
independently-added fields on `LedgerState`/documentation prose, not a
logical conflict with anything in this PR's own two commits above). Per this
task's scope, that resolution is left for whoever integrates this branch
(naturally, much of it should resolve once #4192 itself merges to `main`)
rather than attempted here, to avoid hand-editing consensus-adjacent
inherited code outside this investigation's scope.

## Evidence

- Live instance (unmodified default config), from genesis: 332+
  header-queue-full warnings over several hours, all during
  `fork_reason="chainsync rollback"` churn against one peer, at ~7-8% total
  CPU on a 16-core host.
- A separate isolated instance (same flags, separate data dir/ports/branch):
  a 15s pprof CPU profile early in sync already shows
  `modernc.org/sqlite/lib._sqlite3VdbeExec` (>26% cumulative) and Ed25519
  field arithmetic (>11% cumulative) dominating the single ledger-apply
  goroutine — consistent with a serial, CPU-bound apply path rather than a
  network/disk/memory constraint.
- ARCHITECTURE.md already documents a *different*, already-fixed liveness bug
  with the identical log signature
  (`ensureBlockfetchDrainingAfterForkQueueFailure`); verified that fix is
  present on `main` (6 references in `ledger/chainsync.go`) and running on
  both live instances, so the queue-full events here are the fix's intended
  *bounded, self-healing* behavior, not the old permanent stall. The
  underlying throughput mismatch ARCHITECTURE.md calls out as still real is
  what #4203 and this PR investigate.
- In-process throughput benchmark on a 16-core host,
  `TestBlockPipelineWorkerCountThroughput`: workers=2 avg=8426 blocks/sec,
  workers=8 (scaled/capped) avg=24502 blocks/sec, ratio=2.91x.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt -l .` (clean; one pre-existing
  unrelated `gofmt` finding in `chain/chain_test.go`, not touched by this
  branch, confirmed present on `origin/main`).
- `go test ./ledger/...` (all packages pass).
- `go test -race -timeout 20m ./ledger/...` (all packages pass, ~89s).
- `go test ./internal/test/conformance/...` (passes; no setup required).
- New test `ledger/block_pipeline_worker_count_test.go`
  (`TestBlockPipelineWorkerCount`) exercises the floor, the tracked range,
  and the cap; verified it fails (4 of 6 subtests) against the prior
  hardcoded-2 behavior before reverting to the fix, proving the test is
  causal.
- `golangci-lint`/`nilaway`/`modernize` were not run: not installed in this
  environment (per the developer-toolchain guide). `make docs-parity` /
  `make sql-check` are not applicable (no schema or generated-code changes).
- Devnet/conformance-mode and Antithesis runs were not exercised; this
  change only affects an opt-in, already-disabled-by-default pipeline
  feature's worker count.

Does not merge anything, does not touch PR #4192 or #4102, and does not
request human review — a separate review pass happens later as budget
allows.

Closes/relates to #4203.


chrisguiney:
## Quantified: worker-count fix's actual effect

Measured with a new, permanent test
(`TestBlockPipelineWorkerCountThroughput`, added in `52f9c7c57`) rather than
a one-off measurement, so the numbers below are reproducible by anyone
running `go test ./ledger/ -run TestBlockPipelineWorkerCountThroughput -v`.

**Methodology:** live sync throughput is dominated by network/peer variance
(block density varies enormously by era, peer availability is
non-deterministic), which would confound a live before/after comparison. So
this measures the exact mechanism the fix changes, in-process, with no
network or database involved: 128 genuinely VRF/KES-valid Conway blocks
(`testutil.BuildValidatedConwayBlockBytes` — real VRF proofs, real KES
signatures, real cold keys) are submitted through a
`gouroboros/pipeline.BlockPipeline` configured exactly as `NewLedgerState`
wires production (`productionValidateVerifyConfig`, decode + VRF/KES
validate stages), once at the prior fixed worker count (2) and once at this
host's CPU-scaled count (`blockPipelineWorkerCount()`), 5 repetitions each,
averaged.

**Result on the 16-core host used for this investigation** (`nproc` = 16,
so `blockPipelineWorkerCount()` = 8, the cap):

| workers | avg blocks/sec (5 reps) |
|---|---|
| 2 (old, fixed) | ~7,570-7,900 |
| 8 (new, `blockPipelineWorkerCount()`) | ~19,400-19,700 |

**~2.5x speedup** (2.4-2.6x across separate measurement passes) in raw
decode+VRF/KES-validate throughput. A supplementary sweep across worker
counts (1/2/4/8/16) on the same host showed the expected diminishing-returns
shape (~4.1k → ~7.9k → ~13.8k → ~19.4k → ~20.6k blocks/sec), which is what
motivated capping at 8 rather than leaving it unbounded — 16 workers bought
only ~6% more than 8 on this host, consistent with the pipeline's own
decode/validate stages (rather than core count) becoming the limit, and with
the fact that both stages only feed a single-threaded apply consumer
downstream regardless of how fast they run.

**Caveat, stated plainly:** this quantifies the pipeline stage the fix
touches, not end-to-end sync throughput. The feature this fix scales
(`BlockPipelineEnabled`/`BlockPipelineValidateEnabled`) is off by default and
not what either live instance in #4203 is running, so this fix does not by
itself change the header-queue-saturation symptom under a default
configuration — it makes the opt-in pipeline actually use available cores
once someone turns it on. That's exactly the caveat already stated in this
PR's description; this comment is only adding the measured number behind it.

ledger/continuation_audit.go
```diff
@@ -251,7 +475,90 @@ func (ls *LedgerState) armContinuationAudit(
...
+// block is not re-fetched.
+func (ls *LedgerState) carryForwardWindow(
+	prior *continuationAuditWindow,
+	next *continuationAuditWindow,
+	point ocommon.Point,
+) {
+	onChain, err := ls.primaryChainContainsPoint(prior.forkPoint)
+	if err != nil || !onChain {
+		return
+	}
```
wolf31o2, line 599: primaryChainContainsPoint errors leave next empty, but armContinuationAudit still publishes that empty window. A later audit can report carried producers as missing even though chain membership was never established. Propagate the error and disarm by publishing nil on this failure path.

ledger/crossfork_splice_test.go
```diff
@@ -377,6 +381,749 @@ func TestContinuationAuditAcceptsProducerInSameWindow(t *testing.T) {
...
+						[]lcommon.TransactionInput{
+							mustSpliceAuditInput(
+								t,
+								testHashBytes("race-input"),
+								0,
+							),
+						},
+					),
+				},
+			}
```
wolf31o2: This goroutine builds blocks/events and calls require.NoError. If construction fails, FailNow exits only the worker goroutine and the test can continue with invalid values. Build the blocks and events, with the require.NoError checks, before starting either goroutine.

wolf31o2 (changes requested):
The continuation audit publishes an empty carry-forward window when primaryChainContainsPoint fails, so a later audit can report carried producers as missing even though membership was never established. Propagate the verification error and disarm the audit by publishing nil on failure. The same review also found the test starts goroutines before building values with require.NoError; move those builders before the goroutines so FailNow cannot run on a worker goroutine.

ledger/continuation_audit.go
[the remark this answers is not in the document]
chrisguiney: Fixed in d043ecf1 on `fix/continuation-audit-4102`, brought into this branch by merge 668eb9de.

- `carryForwardWindow` returns the membership read's error with nothing carried.
- `armContinuationAudit` then calls `disarmContinuationAuditUnverified`, which publishes `nil` under `continuationAuditMutex` and logs at Warn, instead of publishing `next`.
- The same failed read now also disarms in `recordLateProducers` and in the raced-publication branch of `commitContinuationAuditBody`.
- Tests: `TestContinuationAuditRearmDisarmsOnFailedForkPointRead`, `TestContinuationAuditLateBodyDisarmsOnFailedMembershipRead`, `TestContinuationAuditRacedBodyDisarmsOnFailedMembershipRead`, driven through the `continuationAuditContainsPoint` hook.

chrisguiney: Fixed in d043ecf1 on `fix/continuation-audit-4102`, brought into this branch by merge 668eb9de.

- `carryForwardWindow` returns the membership read's error with nothing carried.
- `armContinuationAudit` then calls `disarmContinuationAuditUnverified`, which publishes `nil` under `continuationAuditMutex` and logs at Warn, instead of publishing `next`.
- The same failed read now also disarms in `recordLateProducers` and in the raced-publication branch of `commitContinuationAuditBody`.
- Tests: `TestContinuationAuditRearmDisarmsOnFailedForkPointRead`, `TestContinuationAuditLateBodyDisarmsOnFailedMembershipRead`, `TestContinuationAuditRacedBodyDisarmsOnFailedMembershipRead`, driven through the `continuationAuditContainsPoint` hook.

ledger/crossfork_splice_test.go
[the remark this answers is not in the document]
chrisguiney: Fixed in 4214c770 on `fix/continuation-audit-4102`, brought into this branch by merge 668eb9de.

`TestContinuationAuditRearmDoesNotRaceWithBlockfetchAudit` now builds all 300 events, including the `mustSpliceAuditTx` / `mustSpliceAuditInput` checks, on the test goroutine before `wg.Add`. Neither worker goroutine calls `require`, `assert` or `t.Fatal`. No other goroutine in the ledger tests this branch changes does either.

chrisguiney: Fixed in 4214c770 on `fix/continuation-audit-4102`, brought into this branch by merge 668eb9de.

`TestContinuationAuditRearmDoesNotRaceWithBlockfetchAudit` now builds all 300 events, including the `mustSpliceAuditTx` / `mustSpliceAuditInput` checks, on the test goroutine before `wg.Add`. Neither worker goroutine calls `require`, `assert` or `t.Fatal`. No other goroutine in the ledger tests this branch changes does either.

ledger/replay_recovery.go
```diff
@@ -1070,6 +1104,93 @@ func (ls *LedgerState) rewindPrimaryChainForRecovery(
...
+func (ls *LedgerState) settleAuditAfterRewind(
+	prior *continuationAuditWindow,
+	gen uint64,
+	truncated bool,
+	target ocommon.Point,
+) {
+	ls.continuationAuditMutex.Lock()
+	defer ls.continuationAuditMutex.Unlock()
+	if armed := ls.continuationAudit.Load(); armed != nil {
+		if truncated && armed.forkPoint.Slot > target.Slot {
```
wolf31o2: This compares only slots, so a concurrently armed window at the same slot as `target` but with a different hash is retained even though the rewind removed that competing block. That leaves the continuation audit armed with producers from an abandoned same-slot fork and can produce incorrect diagnostics. Compare the full points (slot and hash), and add a regression test for a same-slot competing hash.

wolf31o2 (changes requested):

ledger/replay_recovery.go
chrisguiney: Fixed in c4074b1444 (`fix(ledger): invalidate same-slot audit forks`), matching a2e7c60f9 on `fix/continuation-audit-4102` (#4192): `settleAuditAfterRewind` now compares the full point (slot and hash), not slot alone, and a regression test (`drops a same-slot window on a competing hash`) covers the same-slot competing-hash case. Confirmed the test fails without the comparison fix (reverted the hash check locally, re-ran the test: FAIL; restored it: PASS). Also merged current origin/main into the branch (74d161bb7) to fix the stale-base go-test build failure (glebarez/go-sqlite in migration tests, already fixed on main by #4295); `go build`, `go vet`, `gofmt`, `golangci-lint run ./ledger/...`, and `go test ./ledger/... -race -count=1` all pass at the new head.

wolf31o2 (approved):

PR #4205, opened 2026-09-12
fix(chain): reject a mid-chain first block after a rollback to origin

### LINKED ISSUE #4202: chain: a rollback to origin empties the chain and the next roll-forward is accepted as its first block, so a from-genesis node silently syncs a chain with a missing prefix (every header in the next epoch then fails VRF)
## Problem

On a fresh from-genesis sync with two or more upstream peers, Dingo can adopt a chain that does not start at the network's first block. `Chain.addBlockHeader` skips its continuity check entirely when the chain has no tip and no queued headers, so after a rollback to origin the very next roll-forward header — from whichever peer's chainsync cursor is already ahead — becomes the chain's first block, with no check that it chains onto genesis and no check that its block number is 1.

Nothing downstream notices. The node applies the truncated chain, builds a ledger from it, and reports itself healthy. The damage surfaces one epoch later: the evolving/candidate nonce fold for epoch 0 is missing the skipped blocks' VRF contributions, so the epoch-1 nonce is wrong, and **every** epoch-1 header from **every** peer fails VRF verification. The node blames and recycles each peer in turn and ends in `chain selection stalled: no selectable peer`.

This was invisible before #4101 (47884fd1), which made header VRF/KES/OpCert crypto mandatory during bulk sync. On a build immediately before that commit the same truncated chain syncs "successfully" and produces a wrong ledger instead of stopping — which is the worse failure of the two.

## Environment

- Build: `main` at `53f2fc98` (`v0.70.8-44-g53f2fc98`) plus local patches, with `gouroboros` replaced by a branch carrying `leios-prototype-2026w36` genesis support. `47884fd1` (#4101) is in the tree.
- Network: the Musashi Leios prototype chain re-bootstrapped for `prototype-2026w36` — `systemStart 2026-09-07T00:00:00Z`, `networkMagic 164`, `slotLength 1 s`, `epochLength 21600`, `securityParam 108`, `activeSlotsCoeff 0.05`, Dijkstra era from epoch 0.
- Genesis files loaded from disk and hash-verified: Shelley `1944510a4fd91415444285231058f6f6ff0f6f3ff3d0356c76c00c5a77f29567`, Dijkstra `aa1238f505479a9b104d2cc001b4bf951062cd527200bea9a1857bdd0dc41085`.
- Two upstream peers: a Haskell `prototype-2026w36` relay and IOG's public Leios bootstrap relay. No Mithril (from genesis, empty DB).

### 1. The chain started at block 160, not block 1

Log, 12 ms apart, ~0.9 s after the second peer connected (peer addresses elided):

```
level=INFO msg="ledger: rolling back to 0."
level=INFO msg="ledger: switched to fork at 3360.bf00954d03564f920d5e7429a6fcd86496041f96c455edfa39b2712a37002474"
```

Slot 3360 / `bf00954d…` is canonical, but it is **block number 160**. The node never held blocks 1–159.

The independent chain index for this network and Dingo's own database agree exactly on what was applied:

| | canonical (index) | Dingo (`block_nonce` table) |
| --- | ---: | ---: |
| epoch-0 blocks | 1061 (block 1 … block 1061, slot 21590) | 902 |
| epoch-0 blocks at slot ≥ 3360 | 902 | 902 |
| lowest applied slot | (first block of the chain) | 3360 (block 160) |

The 902 slots Dingo applied are set-identical to the canonical epoch-0 blocks from slot 3360 onward — no gaps after 3360, no extras. The whole defect is the missing 159-block prefix. (`select count(*), min(slot), max(slot) from block_nonce` → `902, 3360, 21590`.)

### 2. The epoch-1 nonce is therefore wrong, and every epoch-1 header fails

Epoch 0's nonce is the Shelley genesis hash, so every epoch-0 header verified. Epoch 1's nonce is derived from the epoch-0 fold, which is short 159 VRF contributions:

```
level=ERROR msg="failed to handle block" component=ledger error="block header crypto verification failed: block header verification failed at slot 21615: vrf: VRF verification failed (issue verifying proof)" slot=21615 hash=2e40125c5c4843c01ad08bbbbfa2610c7e60f48f13853ea3d73f54e609b164ab
level=WARN  msg="recycling connection after header verification failure" component=ledger slot=21615 hash=2e40125c5c4843c01ad08bbbbfa2610c7e60f48f13853ea3d73f54e609b164ab
```

Rejections were repeated against every peer, at every epoch-1 header offered, until no peers were left:

| slot | block | hash | rejections |
| ---: | ---: | --- | ---: |
| 21615 | 1062 | `2e40125c5c4843c01ad08bbbbfa2610c7e60f48f13853ea3d73f54e609b164ab` | 140 |
| 21637 | 1063 | `f4f0f8c7f7eb0a63fa91c5c49d7cb0b08722ea9e2b8e567cc89e1e5adb6d1474` | 124 |
| 21669 | 1064 | `b64de9b7c597035507230202b2ba7484820070d29f3aa5a9677a4e47a0e2daa1` | 4 |
| 21670 | 1065 | `678085c71c0c06b130bb41143332af959a125aa01b7729a199f7a9bf2811ab16` | 2 |
| 21680 | 1066 | `7d741802b21e75ac2f11fbc240b15f7624596165e9efd2dcf353a8f9f05c584a` | 1 |

All five are canonical on this network (Haskell nodes and the network's block explorer accept them). Terminal state:

```
level=INFO msg="chain selection stalled: no selectable peer" genesis_corroboration=false
```

The stored epoch table confirms the node never got past epoch 0 — one row, `epoch_id=0`, `nonce = evolving_nonce = candidate_nonce = 1944510a…` (the Shelley genesis hash), `last_epoch_block_nonce = NULL`.

## Root cause

`chain/chain.go`, `addBlockHeader`:

```go
	// Make sure header fits on chain tip
	if c.tipBlockIndex >= initialBlockIndex ||
		len(c.headers) > 0 {
		headerTip := c.headerTip()
		if !bytes.Equal(queued.prevHash, headerTip.Point.Hash) {
			return NewBlockNotFitChainTipError(...)
		}
		// ... blockNumberContiguous check ...
	}
	// Add header
	c.headers = append(c.headers, queued)
```

(`chain/chain.go:236-266`, with `initialBlockIndex = 1` at `chain/chain.go:38`.)

When `tipBlockIndex < initialBlockIndex` **and** the header queue is empty, both the prev-hash check and the block-number-contiguity check are skipped and the header is appended unconditionally. The same "no tip, no check" shape guards the block-apply paths: `chain/chain.go:590` (`addBlockLocked`) and `chain/chain.go:874` (`addRawBlockLocked`) both gate the prev-hash comparison on `c.tipBlockIndex >= initialBlockIndex`. So a block that does not descend from genesis can enter an empty chain through any of the three.

That state is reachable at any time, not just at process start. `Chain.rollbackLocked` for a rollback to origin deletes every queued header and resets the tip index:

```go
	// Clear out any headers
	c.headers = slices.Delete(c.headers, 0, len(c.headers))
	...
	c.tipBlockIndex = rollbackBlockIndex   // 0 for origin
```

(`chain/chain.go:1596-1604`.)

Sequence that produced the run above:

1. Peer A connects, its initial `FindIntersect` resolves to origin. `handleEventChainsyncRollback` treats a rollback to the current tip as a no-op (`ledger/chainsync.go:2428-2438`), so nothing happens — correct, the chain is already at origin.
2. Peer A rolls forward blocks 1–159 as headers. They sit in the queue; blockfetch has not drained them yet.
3. Peer B connects ~0.4 s later; its initial `FindIntersect` also resolves to origin. This time the rollback is **not** a no-op, because the local header tip is now block 159 — which is exactly why `ledger: rolling back to 0.` is logged only once, for peer B. `rollbackLocked` deletes headers 1–159 and sets `tipBlockIndex = 0`.
4. Peer A's chainsync cursor is untouched and still at block 159, so its next roll-forward is block 160 (slot 3360). The chain is now empty, the guard at `chain/chain.go:236` is inactive, and block 160 is accepted as the chain's first block. `ledger: switched to fork at 3360.bf00954d…`.
5. Sync proceeds normally from block 160. The hole is never detected.

Everything downstream is a faithful consequence of step 4 and is **not** itself buggy: `computeCandidateNonceFast` (`ledger/candidate_nonce.go:208`) correctly returns the stored evolving nonce of the last block before the freeze cutoff (`epochStart + epochLength − 4k/f = 12960`, i.e. block 652 at slot 12917), `calculateEpochNonce` (`ledger/chainsync.go:5677`) correctly takes the NeutralNonce identity branch for the genesis-to-epoch-1 transition (matching `praosStateLastEpochBlockNonce = NeutralNonce` in the initial chain-dep state, and `tickChainDepState`'s `praosStateEpochNonce = praosStateCandidateNonce ⭒ praosStateLastEpochBlockNonce`), the 4k/f window for Dijkstra matches `praosRandomnessStabilisationWindow`, and `praosVRFNonceValue` (`ledger/eras/eras.go:37`) matches `vrfNonceValue`'s `blake2b256(blake2b256("N" ‖ vrfOutput))`. All of that is right; it is just folding over a chain that is missing its first 159 blocks.

## Impact

- A from-genesis sync with more than one peer can silently produce a node whose ledger is derived from an incomplete chain. Stake distribution, rewards, and the epoch nonce are all wrong.
- With #4101 in the tree, the node stops at the first epoch boundary and recycles every peer with `reason=header_verification_failure`, i.e. it mis-attributes its own defect to the whole network and drives itself to zero usable peers. Recovery requires wiping the database.
- Without #4101's mandatory header crypto, the same node keeps going and serves a wrong chain.
- The race widens with more upstream peers and with slower initial blockfetch, so it is more likely on fast devnets/prototype chains with 1 s slots than on mainnet — but the code path is network-independent.

## Minimal repro

Unit level (no network), which is the red test I would want to see:

1. Build a `Chain` with no blocks and no queued headers.
2. Call `AddBlockHeader` with a header whose `PrevHash` is arbitrary and whose `BlockNumber` is, say, 160.
3. It returns `nil` and the header is queued. Expected: `BlockNotFitChainTipError` (or a dedicated "does not chain onto the genesis anchor" error), because a genesis-anchored chain's first block must be block 1 descending from genesis.
4. Same assertion against `AddBlock`/`AddBlocks` (`chain/chain.go:590`) and the raw-block path (`chain/chain.go:874`).

End to end:

1. Empty database, no Mithril, `dingo serve` against a from-genesis network with at least two upstream peers.
2. Let one peer roll forward some headers before the second peer's initial `FindIntersect` completes (natural on a multi-peer start; can be forced by delaying the second peer's handshake).
3. Watch for `ledger: rolling back to 0.` immediately followed by `ledger: switched to fork at <slot>.<hash>` where `<slot>` is not the network's first block.
4. `select count(*), min(slot) from block_nonce;` — `min(slot)` is above the network's first block slot and the count is short by the skipped prefix.
5. At the first epoch boundary, every peer's headers fail with `vrf: VRF verification failed (issue verifying proof)`.

## Suggested fix

1. **Fail closed on an empty chain.** Record the chain's anchor (origin for a from-genesis node, the imported snapshot point for a Mithril/snapshot bootstrap) and validate the first header/block against it in all three places that currently skip the check when `tipBlockIndex < initialBlockIndex`: `addBlockHeader` (`chain/chain.go:236`), `addBlockLocked` (`chain/chain.go:590`), `addRawBlockLocked` (`chain/chain.go:874`). For a genesis anchor, require `BlockNumber == 1` and the genesis-derived prev-hash. Returning `BlockNotFitChainTipError` is enough — the existing fork-resolution/re-intersect machinery already handles it. A restored-from-snapshot chain is never empty at this point (`ChainManager.loadPrimaryChain`, `chain/manager.go:405-419`, seeds `tipBlockIndex` from the most recent stored block), so this does not affect the Mithril path.
2. **A rollback that empties the chain must invalidate every peer's cursor.** In `handleEventChainsyncRollback`, when the rollback target is origin (or the rollback leaves the chain empty), publish `ChainsyncResyncEvent` for all connections rather than only the one that sent the rollback, so no peer can continue rolling forward from a cursor that is now ahead of a chain that no longer exists.
3. **Optional, defence in depth at the boundary.** When closing an epoch, cross-check the number of blocks folded into the nonce against the block-number span of the epoch (`lastBlock.Number − firstBlock.Number + 1`). A prefix hole would then be reported at the boundary as a local inconsistency instead of surfacing as a network-wide VRF failure.
4. **Do not blame peers for a unanimous failure.** When headers from *every* peer fail VRF at the same slot, that is evidence of local nonce divergence, not peer misbehaviour. Recycling each connection with `reason=header_verification_failure` until `chain selection stalled: no selectable peer` hides the real cause; that case deserves a distinct diagnostic.

chadaibbhot-git:
Fixes #4202

## Mechanism

`Chain.rollbackLocked` empties the chain for a rollback to origin: it deletes every queued header and resets `tipBlockIndex` to 0. All three add paths — `addBlockHeader`, `addBlockLocked` and `addRawBlockLocked` — skip their prev-hash and block-number continuity checks when there is no tip to chain onto, so the next roll forward is accepted as the chain's first block. On a multi-peer from-genesis start that next roll forward comes from whichever peer's chainsync cursor survived the rollback, still pointing several blocks ahead, so the chain grows from block N with a silently missing prefix. The epoch nonce folded over that chain is short the skipped blocks' VRF contributions, and at the next epoch boundary every header from every peer fails VRF verification until no peer is selectable.

## Fix

When the chain has been emptied back to origin — no tip block and at least one prior mutation, which is exactly what a rollback to origin leaves behind — the first block or header must carry block number 0. That is the number Ouroboros gives the first block after genesis: the Byron epoch-boundary block on a Byron network, the first block of the starting era on a post-Byron genesis network.

Nothing wider is safe, and an earlier revision of this PR said otherwise. It tolerated block number 1 on the grounds that a chain starting one block late would still be caught at its second block by the ordinary contiguity check. That was wrong: `blockNumberContiguous` compares a candidate against the **accepted tip**, so a chain anchored at block number 1 is self-consistent from its second block onwards and block 0 is never missed. Measured on the previous head `2a81775b`:

```
AddBlock(number=1, arbitrary prev hash) err=<nil>
AddBlock(number=2 chaining onto it)     err=<nil>
tip block number=2
chain's first block number=1
```

The tolerance did not defer the check to the second block, it permanently shortened the chain by exactly the prefix it tolerated — the same truncated prefix this PR exists to prevent. The constant is now `firstBlockNumber uint64 = 0` and `firstBlockNumberValid` is an equality rather than a `<=`.

The `chain` package has no knowledge of the network's genesis hash, so the prev-hash half of the continuity check still cannot be applied at origin. The block number is the whole of the anchor available there: it closes the truncated prefix, but a candidate carrying block number 0 is accepted whatever its hash and prev hash say. Binding the first block's hash as well needs the genesis hash, which belongs to the ledger rather than to this package; that residual gap is stated on `firstBlockNumberValid` and is not closed here. The rejection reuses `BlockNotFitChainTipError` so the ledger's existing fork-resolution path re-intersects the offending connection rather than failing the node.

A chain that has never been mutated is deliberately excluded: it has no anchor yet, and it is the state the bulk block importer (`internal/node/load`) fills from a local immutable database, which legitimately establishes the chain's first block itself. That import always iterates from the immutable DB's first block when the chain is empty, and a snapshot-restored chain is never empty at this point (`ChainManager.loadPrimaryChain` seeds `tipBlockIndex` from the most recent stored block), so neither bootstrap path is affected.

## Tests

`chain/origin_continuity_test.go`, table-driven over {block 0 accepted, block 1 rejected, block 2 rejected, block 160 rejected} against a chain grown to six blocks and then rolled back to origin:

- `TestAddBlockHeaderAfterRollbackToOriginRequiresFirstBlock`
- `TestAddBlockAfterRollbackToOriginRequiresFirstBlock`
- `TestAddRawBlockAfterRollbackToOriginRequiresFirstBlock`
- `TestAddRawBlocksAfterRollbackToOriginWithQueuedHeader` — a queued header binds the next block's hash, not its number, so `AddRawBlocks` with a matching hash and a mid-chain block number is still rejected.
- `TestAddBlockAfterRollbackToOriginUsesQueuedHeaderBlockNumber` — the decoded-block path takes the number from the queued header, which this same check anchored at queue time.
- `TestAddBlockAfterRollbackToOriginRejectsChainShortOfBlockZero` — the cost of tolerating a number-1 first block: block 1 is rejected, and the block 2 that would chain onto it (and satisfy `blockNumberContiguous` against a tip of 1, making the gap permanent) is rejected too. The positive control adds block 0 then block 1 and reads the chain's first block back **from origin**, asserting it is number 0 — a short chain is invisible from the tip.
- `TestFreshChainAcceptsFirstBlockFromAnySource` — pins the initial-sync side: a chain that has never been mutated keeps accepting whatever it is given as its first block, so the bulk importer is unaffected.

Red/green for the narrowing, on the merge of `main` (`3459944f`, tolerance still 1):

```
--- FAIL: TestAddBlockAfterRollbackToOriginRejectsChainShortOfBlockZero
    origin_continuity_test.go:445: AddBlock accepted block number 1 as the chain's first block; the chain is then permanently short block 0
--- FAIL: TestAddBlockHeaderAfterRollbackToOriginRequiresFirstBlock/first_block_number_1_rejected
--- FAIL: TestAddBlockAfterRollbackToOriginRequiresFirstBlock/first_block_number_1_rejected
--- FAIL: TestAddRawBlockAfterRollbackToOriginRequiresFirstBlock/first_block_number_1_rejected
```

Two existing tests regrew a chain after a rollback to origin from the shared `testBlocks` fixture, whose first block is numbered 1, so they depended on the tolerance. Both now regrow from block number 0 — `blocksFromOrigin` in `chain_test.go` for `TestChainIteratorReverseRollbackToOriginClamps`, and the origin case in `nextBatch` for `TestRollbackDoesNotResolveUncommittedBlockIndex`. What each test asserts is unchanged; only the fixture numbering moved.

## Verification

- `go build ./...`, `go vet ./chain/...` clean
- `go test -count=1 ./...` — ok
- `go test -count=1 -race ./chain/...` — ok
- `golangci-lint run ./chain/...` — 0 issues

## Not covered here

A chain that has never held a block is not anchored, so a peer that lies to a genuinely fresh node on its very first roll forward is not caught by this change. Neither is the hash of the first block after a rollback to origin: the block number is the only anchor the `chain` package has without the network's genesis hash. It is also not the state the reported failure went through: on a fresh node the first header after an origin intersect is necessarily the network's first block, and the defect is reachable only once a rollback has emptied a chain that peers are already rolling forward into. The remaining suggestions in #4202 — broadcasting a resync to every connection when a rollback empties the chain, and an epoch-boundary block-count cross-check — are left for separate changes.

---
## Summary by cubic
Fixes #4202: after a rollback to origin, the chain would accept the next block from any peer as its first block, even if that block was far into the chain. This could silently drop a prefix of blocks and cause VRF verification failures at the next epoch boundary.

**Bug Fixes**
- Rejects a first block with block number greater than 1 when the chain has been emptied back to origin.
- Applies the check to all three add paths: `addBlockHeader`, `addBlockLocked`, and `addRawBlockLocked`.
- The check now also applies when a header is queued, so a raw block with a matching hash but a mid-chain block number cannot bypass it.
- Leaves fresh chains unanchored, so the bulk importer still sets the first block without restriction.
- Reuses `BlockNotFitChainTipError` so the ledger's existing re-intersection logic handles the rejection.

<sup>Written for commit 2a81775bc07fcf19de33ba6d1eb064777a7fd5dd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4205?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Chains rolled back to their origin now reject blocks that do not begin with the genuine first block.
  - Consistent validation is applied when adding block headers, decoded blocks, or raw blocks.
  - Fresh, never-mutated chains continue to accept valid first blocks as before.

### DIFF
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -158,6 +158,67 @@ func blockNumberContiguous(eraId uint8, blockNumber, parentNumber uint64) bool {
 	return false
 }
 
+// maxFirstBlockNumber is the largest block number accepted for the first block
+// of a chain that has been emptied back to origin. Ouroboros numbers the first
+// block after genesis 0 -- the Byron epoch-boundary block on a Byron network,
+// the first block of the starting era on a post-Byron genesis network -- so 0
+// is the expected value. 1 is tolerated as well because some networks and
+// chain indexes number their first block 1; the tolerance costs nothing, since
+// a chain that started one block late is still caught at its second block by
+// the ordinary contiguity check above.
+const maxFirstBlockNumber uint64 = 1
+
+// originTipHash stands in for the tip hash when a block or header is rejected
+// against a chain at origin: there is no tip block to name.
+const originTipHash = "origin"
+
+// emptiedToOrigin reports whether this chain holds nothing -- no tip block and
+// no queued headers -- after having been mutated at least once. That is the
+// state rollbackLocked leaves behind for a rollback to origin: it drops every
+// queued header and resets tipBlockIndex to 0.
+//
+// The distinction matters because the continuity checks below are skipped when
+// there is no tip to chain onto, and a rollback to origin can produce that
+// state at any time during a run. A peer whose chainsync cursor is still ahead
+// then delivers block N rather than the network's first block, and it is
+// accepted as the chain's first block: the chain grows with a silently missing
+// prefix, and the epoch nonce folded over it is wrong, so every header in the
+// next epoch fails VRF verification (issue #4202).
+//
+// A chain that has never been mutated is deliberately excluded. It has no
+// anchor yet -- the chain package does not know the network's genesis hash --
+// and it is the state the bulk block importer fills from a local immutable
+// database, which legitimately establishes the chain's first block itself.
+func (c *Chain) emptiedToOrigin() bool {
+	return c.tipBlockIndex < initialBlockIndex &&
+		len(c.headers) == 0 &&
+		c.mutationGeneration > 0
+}
+
+// firstBlockNumberValid reports whether blockNumber is plausible for the first
+// block of a chain that has been emptied back to origin. The chain package has
+// no knowledge of the network's genesis hash, so the prev-hash half of the
+// continuity check cannot be applied at origin; the block number is the anchor
+// available here, and it is enough to reject a truncated prefix.
+func firstBlockNumberValid(blockNumber uint64) bool {
+	return blockNumber <= maxFirstBlockNumber
+}
+
+// newBlockNotFitChainOriginError builds the rejection for a block or header
+// offered as the first block of a chain sitting at origin. Reusing
+// BlockNotFitChainTipError keeps the ledger's existing recovery path, which
+// re-intersects the offending connection rather than failing the node.
+func newBlockNotFitChainOriginError(
+	blockHash string,
+	blockPrevHash string,
+) BlockNotFitChainTipError {
+	return NewBlockNotFitChainTipError(
+		blockHash,
+		blockPrevHash,
+		originTipHash,
+	)
+}
+
 func (c *Chain) headerTip() ochainsync.Tip {
 	if len(c.headers) == 0 {
 		return c.currentTip
@@ -251,6 +312,15 @@ func (c *Chain) addBlockHeader(
 				headerTip.BlockNumber,
 			)
 		}
+	} else if c.emptiedToOrigin() &&
+		!firstBlockNumberValid(queued.blockNumber) {
+		// The chain was emptied back to origin, so there is no tip to chain
+		// onto and the checks above cannot run. Anchor the first header on
+		// its block number instead; see emptiedToOrigin.
+		return newBlockNotFitChainOriginError(
+			headerHash.String(),
+			headerPrevHash.String(),
+		)
 	}
 	// Add header
 	c.headers = append(c.headers, queued)
@@ -479,6 +549,14 @@ func (c *Chain) addBlockLocked(
 				c.currentTip.BlockNumber,
 			)
 		}
+	} else if c.emptiedToOrigin() && !firstBlockNumberValid(blockNumber) {
+		// Chain emptied back to origin: anchor the first block on its block
+		// number, the only continuity the chain can check here. See
+		// emptiedToOrigin.
+		return event.Event{}, newBlockNotFitChainOriginError(
+			hex.EncodeToString(blockHashBytes),
+			hex.EncodeToString(blockPrevHashBytes),
+		)
 	}
 	// Build new block record
 	tmpPoint := point
@@ -712,6 +790,14 @@ func (c *Chain) addRawBlockLocked(
 				hex.EncodeToString(c.currentTip.Point.Hash),
 			)
 		}
+	} else if c.emptiedToOrigin() && !firstBlockNumberValid(rb.BlockNumber) {
+		// Chain emptied back to origin: anchor the first block on its block
+		// number, the only continuity the chain can check here. See
+		// emptiedToOrigin.
+		return event.Event{}, newBlockNotFitChainOriginError(
+			hex.EncodeToString(rb.Hash),
+			hex.EncodeToString(rb.PrevHash),
+		)
 	}
 	tmpPoint := ocommon.NewPoint(rb.Slot, rb.Hash)
 	newBlockIndex := c.tipBlockIndex + 1
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -172,10 +172,10 @@ const maxFirstBlockNumber uint64 = 1
 // against a chain at origin: there is no tip block to name.
 const originTipHash = "origin"
 
-// emptiedToOrigin reports whether this chain holds nothing -- no tip block and
-// no queued headers -- after having been mutated at least once. That is the
-// state rollbackLocked leaves behind for a rollback to origin: it drops every
-// queued header and resets tipBlockIndex to 0.
+// atOriginAfterMutation reports whether this chain sits at origin -- no tip
+// block -- after having been mutated at least once. That is the state
+// rollbackLocked leaves behind for a rollback to origin: it drops every queued
+// header and resets tipBlockIndex to 0.
 //
 // The distinction matters because the continuity checks below are skipped when
 // there is no tip to chain onto, and a rollback to origin can produce that
@@ -185,13 +185,24 @@ const originTipHash = "origin"
 // prefix, and the epoch nonce folded over it is wrong, so every header in the
 // next epoch fails VRF verification (issue #4202).
 //
+// The predicate deliberately ignores the header queue. A queued header does
+// not anchor an incoming raw block: addRawBlockLocked only checks that the
+// block's hash matches c.headers[0], while RawBlock.BlockNumber is whatever
+// the caller supplied. Queuing the chain's first header and then handing over
+// a same-hash block from further along the chain would otherwise skip this
+// check, delete the queued header and persist that block number. RawBlock and
+// AddRawBlocks are exported, so the chain cannot assume its callers derive
+// those fields consistently: the check has to hold for the exported API, not
+// only for the in-tree producers. (addBlockLocked takes the block number from
+// a matching queued header, which this same check anchored when the header was
+// queued.)
+//
 // A chain that has never been mutated is deliberately excluded. It has no
 // anchor yet -- the chain package does not know the network's genesis hash --
 // and it is the state the bulk block importer fills from a local immutable
 // database, which legitimately establishes the chain's first block itself.
-func (c *Chain) emptiedToOrigin() bool {
+func (c *Chain) atOriginAfterMutation() bool {
 	return c.tipBlockIndex < initialBlockIndex &&
-		len(c.headers) == 0 &&
 		c.mutationGeneration > 0
 }
 
@@ -312,11 +323,11 @@ func (c *Chain) addBlockHeader(
 				headerTip.BlockNumber,
 			)
 		}
-	} else if c.emptiedToOrigin() &&
+	} else if c.atOriginAfterMutation() &&
 		!firstBlockNumberValid(queued.blockNumber) {
-		// The chain was emptied back to origin, so there is no tip to chain
+		// The chain was rolled back to origin, so there is no tip to chain
 		// onto and the checks above cannot run. Anchor the first header on
-		// its block number instead; see emptiedToOrigin.
+		// its block number instead; see atOriginAfterMutation.
 		return newBlockNotFitChainOriginError(
 			headerHash.String(),
 			headerPrevHash.String(),
@@ -549,10 +560,11 @@ func (c *Chain) addBlockLocked(
 				c.currentTip.BlockNumber,
 			)
 		}
-	} else if c.emptiedToOrigin() && !firstBlockNumberValid(blockNumber) {
-		// Chain emptied back to origin: anchor the first block on its block
-		// number, the only continuity the chain can check here. See
-		// emptiedToOrigin.
+	} else if c.atOriginAfterMutation() && !firstBlockNumberValid(blockNumber) {
+		// Chain rolled back to origin: anchor the first block on its block
+		// number -- taken from the matching queued header when there is one --
+		// the only continuity the chain can check here. A queued header does
+		// not exempt the block from the check. See atOriginAfterMutation.
 		return event.Event{}, newBlockNotFitChainOriginError(
 			hex.EncodeToString(blockHashBytes),
 			hex.EncodeToString(blockPrevHashBytes),
@@ -790,10 +802,12 @@ func (c *Chain) addRawBlockLocked(
 				hex.EncodeToString(c.currentTip.Point.Hash),
 			)
 		}
-	} else if c.emptiedToOrigin() && !firstBlockNumberValid(rb.BlockNumber) {
-		// Chain emptied back to origin: anchor the first block on its block
-		// number, the only continuity the chain can check here. See
-		// emptiedToOrigin.
+	} else if c.atOriginAfterMutation() &&
+		!firstBlockNumberValid(rb.BlockNumber) {
+		// Chain rolled back to origin: anchor the first block on its block
+		// number, the only continuity the chain can check here. A queued
+		// header does not exempt the block -- the hash check above binds the
+		// hash, not the number. See atOriginAfterMutation.
 		return event.Event{}, newBlockNotFitChainOriginError(
 			hex.EncodeToString(rb.Hash),
 			hex.EncodeToString(rb.PrevHash),
--- a/chain/origin_continuity_test.go
+++ b/chain/origin_continuity_test.go
@@ -1,0 +1,264 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package chain_test
+
+import (
+	"errors"
+	"testing"
+
+	ocommon "github.com/blinklabs-io/gouroboros/protocol/common"
+
+	"github.com/blinklabs-io/dingo/chain"
+)
+
+// originContinuityCase describes one candidate offered as the first block of a
+// chain that a rollback has emptied back to origin.
+type originContinuityCase struct {
+	name        string
+	blockNumber uint64
+	prevHash    string
+	wantReject  bool
+}
+
+// originContinuityCases covers both directions: a chain emptied back to origin
+// must still accept the network's genuine first block, and must reject a block
+// from further along the chain -- which is what a peer whose chainsync cursor
+// survived the rollback offers next (issue #4202).
+//
+// The chain package does not know the network's genesis hash, so the anchor
+// available at origin is the block number. The first block of a Cardano chain
+// carries block number 0 (Ouroboros numbers the first block after genesis 0 --
+// the Byron epoch-boundary block on a Byron network, the first block of the
+// starting era on a post-Byron genesis network); 1 is tolerated because some
+// networks and chain indexes number their first block 1.
+var originContinuityCases = []originContinuityCase{
+	{
+		name:        "genuine first block number 0 accepted",
+		blockNumber: 0,
+		prevHash:    "",
+		wantReject:  false,
+	},
+	{
+		name:        "genuine first block number 1 accepted",
+		blockNumber: 1,
+		prevHash:    "",
+		wantReject:  false,
+	},
+	{
+		name:        "block from further along the chain rejected",
+		blockNumber: 160,
+		prevHash:    testHashPrefix + "00a0",
+		wantReject:  true,
+	},
+	{
+		name:        "second block of the chain rejected",
+		blockNumber: 2,
+		prevHash:    testHashPrefix + "0001",
+		wantReject:  true,
+	},
+}
+
+func originCandidate(c originContinuityCase) *MockBlock {
+	return &MockBlock{
+		MockBlockNumber: c.blockNumber,
+		MockSlot:        3360,
+		MockHash:        testHashPrefix + "beef",
+		MockPrevHash:    c.prevHash,
+	}
+}
+
+func originCandidateRawBlock(c originContinuityCase) chain.RawBlock {
+	candidate := originCandidate(c)
+	return chain.RawBlock{
+		Slot:        candidate.MockSlot,
+		Hash:        candidate.Hash().Bytes(),
+		BlockNumber: candidate.MockBlockNumber,
+		Type:        uint(candidate.Type()),
+		PrevHash:    decodeHex(candidate.MockPrevHash),
+		Cbor:        []byte{0x80},
+	}
+}
+
+// chainEmptiedToOrigin returns a chain that held blocks and was then rolled
+// back to origin -- the state that makes the continuity checks inapplicable
+// and that a peer with a stale chainsync cursor rolls forward into.
+func chainEmptiedToOrigin(t *testing.T) *chain.Chain {
+	t.Helper()
+	db := newTestDB(t)
+	cm, err := chain.NewManager(db, nil)
+	if err != nil {
+		t.Fatalf("unexpected error creating chain manager: %s", err)
+	}
+	mustSetLedger(t, cm, 10)
+	c := cm.PrimaryChain()
+	for _, testBlock := range testBlocks {
+		if err := c.AddBlock(testBlock, nil); err != nil {
+			t.Fatalf("unexpected error adding block to chain: %s", err)
+		}
+	}
+	if err := c.Rollback(ocommon.NewPointOrigin()); err != nil {
+		t.Fatalf("unexpected error rolling back to origin: %s", err)
+	}
+	if tip := c.Tip(); len(tip.Point.Hash) != 0 {
+		t.Fatalf(
+			"expected an empty chain after rollback to origin, got tip %d.%x",
+			tip.Point.Slot,
+			tip.Point.Hash,
+		)
+	}
+	if c.HeaderCount() != 0 {
+		t.Fatalf(
+			"expected no queued headers after rollback to origin, got %d",
+			c.HeaderCount(),
+		)
+	}
+	return c
+}
+
+func assertOriginResult(
+	t *testing.T,
+	tc originContinuityCase,
+	op string,
+	err error,
+) {
+	t.Helper()
+	if !tc.wantReject {
+		if err != nil {
+			t.Fatalf(
+				"%s: must accept the network's first block (number %d): %s",
+				op,
+				tc.blockNumber,
+				err,
+			)
+		}
+		return
+	}
+	if err == nil {
+		t.Fatalf(
+			"%s: accepted block number %d as the chain's first block; "+
+				"the chain would grow with a missing prefix",
+			op,
+			tc.blockNumber,
+		)
+	}
+	var notFitErr chain.BlockNotFitChainTipError
+	if !errors.As(err, &notFitErr) {
+		t.Fatalf(
+			"%s: expected BlockNotFitChainTipError, got %T: %s",
+			op,
+			err,
+			err,
+		)
+	}
+}
+
+// TestAddBlockHeaderAfterRollbackToOriginRequiresFirstBlock is the reported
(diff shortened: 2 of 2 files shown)


chain/chain.go
```diff
@@ -712,6 +790,14 @@ func (c *Chain) addRawBlockLocked(
 				hex.EncodeToString(c.currentTip.Point.Hash),
 			)
 		}
+	} else if c.emptiedToOrigin() && !firstBlockNumberValid(rb.BlockNumber) {
```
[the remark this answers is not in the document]
chadaibbhot-git: Checked this one against the code before changing anything, and I do not think the
sequence is reachable, so I have left `addRawBlockLocked` as it is. Three
independent reasons:

**1. `rb.Hash` and `rb.BlockNumber` are not independent at any construction site.**
`chain.RawBlock` has exactly one non-test constructor, `rawBlockFromImmutableBlock`
(`internal/node/load.go:1357-1381`), and it fills both fields from the same decoded
header:

```go
header, err := gledger.NewBlockHeaderFromCbor(block.Type, headerCbor)
...
return chain.RawBlock{
    Slot:        header.SlotNumber(),
    Hash:        header.Hash().Bytes(),
    BlockNumber: header.BlockNumber(),
```

The queued header's fields come from the same accessor pair on the same header
value (`chain/chain.go:273` and `chain/chain.go:276`). For a raw block to match
`c.headers[0].point.Hash` while carrying a different number, two distinct headers
would have to share a blake2b-256 hash.

**2. The raw-block path never runs on a chain that can hold queued headers or that
has been rolled back.** `AddRawBlocks`/`AddRawBlocksWithCallback` have two non-test
callers, `internal/node/load.go:1019` and `internal/node/load.go:1167`, and both
operate on an import-only chain built for the copy: `internal/node/load.go:767` and
`mithril/sync.go:544`, each a fresh `chain.NewManager(db, nil)` with no event bus.
Nothing feeds chainsync headers into those instances and the importer never rolls
back, so on the instance that takes this code path `len(c.headers) > 0` and
`emptiedToOrigin()` are both permanently false. The rollback-to-origin state the
finding starts from cannot exist there.

**3. The live chainsync/blockfetch path does not use `RawBlock` at all.** It is
`ledger/state.go:11810` -> `Chain.AddBlock` -> `addBlockLocked`, and when a queued
header matches, that function takes the number from the header rather than from the
block (`chain/chain.go:515-525`, `blockNumber = firstHeader.blockNumber`). The
header already passed the origin check this PR adds (`chain/chain.go:315-322`), so
the number persisted on the path where a header can be queued is the validated one.
That asymmetry is why the `len(c.headers) == 0` term in `emptiedToOrigin`
(`chain/chain.go:192-196`) costs nothing on the live path.

On the suggested edit itself: dropping the header term from `emptiedToOrigin` would
be a no-op at the other two call sites - `addBlockHeader` only reaches its origin
branch when the queue is empty anyway (`chain/chain.go:287-288`), and
`addBlockLocked` would be re-checking the header-derived number it just adopted -
so the only behavioural change would be at `addRawBlockLocked`, guarding an input
that no caller can produce.

The real remaining latitude at `addRawBlockLocked` is caller trust in the struct as
a whole, and it is wider than `BlockNumber` and predates this PR: the function also
takes `rb.Slot` for the point it stores (`chain/chain.go:802`) and persists
`rb.Cbor` verbatim, neither cross-checked against the matched header. If you want
that closed - a raw block must agree with the queued header on slot and number, not
only on hash - I am glad to write it, but it is a change to the import API's
contract rather than part of the rollback-to-origin fix, and I would rather not
widen this PR without your say-so.

Coverage for the reachable half of the sequence already exists:
`TestAddRawBlockAfterRollbackToOriginRequiresFirstBlock` in
`chain/origin_continuity_test.go` drives rollback-to-origin -> raw block numbered 2
and 160 rejected, numbers 0 and 1 accepted, through the exported `AddRawBlocks`.
`go test ./chain/` passes locally and the race job is green on the current head.

wolf31o2: `emptiedToOrigin()` is false whenever a queued header remains, so this origin-number guard is skipped after rollback -> queue header #1 -> `AddRawBlocks`. `addRawBlockLocked` then accepts a same-hash `RawBlock` with `BlockNumber >= 2`, deletes the queued header, persists that number, and advances the chain with a missing prefix. `RawBlock` and `AddRawBlocks` are exported, so production constructors deriving consistent fields do not make the state impossible. Apply the origin-number check independently of `len(c.headers) == 0` and add this sequence as a regression test.

wolf31o2 (changes requested):
The current head still needs one reachable-state fix in the raw-block path.

chain/chain.go
chadaibbhot-git: You are right, and the earlier "unreachable via in-tree constructors" answer was the wrong standard — `RawBlock` and `AddRawBlocks` are exported, so the check has to hold for the exported API, not just for the callers that happen to live in this repo. Fixed in 2a81775b.

`emptiedToOrigin()` is gone. The predicate is now `atOriginAfterMutation()` (`tipBlockIndex < initialBlockIndex && mutationGeneration > 0`), with the `len(c.headers) == 0` condition dropped, and it is applied on all three add paths:

- `addRawBlockLocked`: the origin block-number check now runs whenever the chain is at origin after a mutation, independently of the header queue. Your sequence — rollback to origin, queue header #1, `AddRawBlocks` with that header's hash and `BlockNumber` 2 — is rejected with `BlockNotFitChainTipError`; the queued header is left in place and the chain stays at origin.
- `addBlockHeader`: behaviour unchanged. That branch is only reached when the queue is empty, so removing the condition is a no-op there; it just stops being load-bearing.
- `addBlockLocked`: same predicate. This path takes the block number from a matching queued header when there is one, and that number was anchored by this same check at queue time, so it was not the hole — but it now reads from one predicate instead of two.

Tests (`chain/origin_continuity_test.go`):

- `TestAddRawBlocksAfterRollbackToOriginWithQueuedHeader` — your sequence exactly: rollback to origin, `AddBlockHeader` for block 1, `AddRawBlocks` with the same hash and `BlockNumber` 2 must be rejected and must not consume the queued header; then the consistent raw block (same hash, `BlockNumber` 1) is accepted, clears the queue and becomes the tip at block number 1. Red on 79b26f48 ("AddRawBlocks accepted block number 2 as the chain's first block because a header was queued"), green on 2a81775b.
- `TestAddBlockAfterRollbackToOriginUsesQueuedHeaderBlockNumber` — pins why the decoded-block path needs no separate guard, so the asymmetry with the raw path is documented rather than implied.
- The no-header-queued variants were already covered by `TestAddRawBlockAfterRollbackToOriginRequiresFirstBlock`, `TestAddBlockAfterRollbackToOriginRequiresFirstBlock` and `TestAddBlockHeaderAfterRollbackToOriginRequiresFirstBlock`.

Verified on 2a81775b: `go build ./...`, `go vet`, `go test ./chain/ ./internal/node/ ./mithril/`, `go test -race ./chain/`, `golangci-lint run ./chain/...` (0 issues).

chain/chain.go
```diff
@@ -158,6 +158,78 @@ func blockNumberContiguous(eraId uint8, blockNumber, parentNumber uint64) bool {
...
 
+// maxFirstBlockNumber is the largest block number accepted for the first block
+// of a chain that has been emptied back to origin. Ouroboros numbers the first
+// block after genesis 0 -- the Byron epoch-boundary block on a Byron network,
+// the first block of the starting era on a post-Byron genesis network -- so 0
+// is the expected value. 1 is tolerated as well because some networks and
+// chain indexes number their first block 1; the tolerance costs nothing, since
+// a chain that started one block late is still caught at its second block by
+// the ordinary contiguity check above.
+const maxFirstBlockNumber uint64 = 1
```
wolf31o2: The stated justification does not hold: contiguity compares a candidate against the accepted tip, so a chain anchored at block number 1 stays self-consistent and block 0 is never missed. Measured at this head: after a rollback to origin, AddBlock with number 1 and an arbitrary prev hash is accepted, then number 2 chaining onto it is accepted, leaving the chain permanently short block 0. A bounded rollback to origin only succeeds while tipBlockIndex is below k, the window where a surviving peer cursor sits at a low block number. Narrow to 0, or drop the claim and state the residual gap.

chain/origin_continuity_test.go
```diff
@@ -0,0 +1,414 @@
...
+// networks and chain indexes number their first block 1.
+var originContinuityCases = []originContinuityCase{
+	{
+		name:        "genuine first block number 0 accepted",
+		blockNumber: 0,
+		prevHash:    "",
+		wantReject:  false,
+	},
+	{
+		name:        "genuine first block number 1 accepted",
```
wolf31o2: This case pins acceptance of block number 1 but not what that acceptance costs: no test adds a second block on top of an accepted number-1 first block, so the chain missing block 0 that this tolerance admits is uncovered. Add a case for first block 1 followed by block 2 asserting the intended outcome.

wolf31o2 (changes requested):
The prior raw-block finding is fixed at this head and proved causal by an in-place revert; the block-number 1 tolerance is what remains.

chain/chain.go
chadaibbhot-git: You are right, and the tolerance is gone rather than documented: narrowed to 0 in `1873b408`.

Your reasoning is the part I had wrong. `blockNumberContiguous` compares a candidate against the accepted tip, so a chain anchored at block number 1 is self-consistent from its second block onwards and the claim that it would be "caught at its second block" described a check that never runs. Reproduced your measurement on the old head `2a81775b`, same sequence:

```
AddBlock(number=1, arbitrary prev hash) err=<nil>
AddBlock(number=2 chaining onto it)     err=<nil>
tip block number=2
chain's first block number=1
```

So the tolerance did not defer the check, it permanently shortened the chain by exactly the prefix it tolerated -- the truncated prefix #4202 is about. Your point about the window makes it worse rather than milder: a bounded rollback to origin only succeeds while `tipBlockIndex` is below k, which is the early-sync window where a surviving peer cursor sits at a low block number, so a tolerance of 1 sits right where the offered numbers actually are.

What changed in `1873b408`:

- `maxFirstBlockNumber uint64 = 1` is now `firstBlockNumber uint64 = 0`, and `firstBlockNumberValid` is an equality instead of a `<=`. The false justification is replaced by the reason nothing wider is safe.
- `atOriginAfterMutation` and the three call sites are unchanged; only the admissible value moved.

I took the narrowing rather than documenting the gap because #4202 still holds with it, strictly more so: the 0-only rule rejects a superset of what the previous rule rejected, and every case the original tests pinned as rejected is still rejected. 0 is also what the networks serve. On a Byron-genesis network the first block a peer rolls forward from origin is the epoch-boundary block, whose `ChainDifficulty` -- which gouroboros returns as `BlockNumber()` -- is 0; on a post-Byron genesis network it is the first block of the starting era, number 0. Checked against the public chain indexes for both testnets: preview's first block is number 0, and preprod's first *main* block is number 1 but is preceded by the epoch-boundary block at 0, which is what a roll forward from origin delivers first.

Two in-tree tests did depend on the tolerance, and both are fixture artifacts rather than behaviour: the shared `testBlocks` numbers its first block 1, and `TestChainIteratorReverseRollbackToOriginClamps` and `TestRollbackDoesNotResolveUncommittedBlockIndex` both regrow a chain after a rollback to origin out of that fixture (the latter hung rather than failed, because its batch callback never runs on a rejected batch). They now regrow from block number 0 -- `blocksFromOrigin` in `chain_test.go`, and the origin case in `nextBatch` -- with what each test asserts untouched. Flagging that explicitly since it is the only place this PR now touches tests it did not add.

Residual gap, stated on `firstBlockNumberValid` rather than implied: the `chain` package has no knowledge of the network's genesis hash, so the prev-hash half of the continuity check still cannot be applied at origin. A candidate carrying block number 0 is accepted whatever its hash and prev hash say. Binding the first block's hash needs the genesis hash, which belongs to the ledger rather than here; deliberately not done in this PR.

Tests for the narrowing are in the `origin_continuity_test.go` thread. Verified on `1873b408`: `go build ./...`, `go vet ./chain/...`, `golangci-lint run ./chain/...` (0 issues), `go test -count=1 ./...`, `go test -count=1 -race ./chain/...`. PR body updated to drop the old claim and state the residual gap.

chain/origin_continuity_test.go
chadaibbhot-git: Added in `1873b408`. Since the `chain.go:169` thread settled on narrowing the tolerance to 0, the asserted outcome is rejection of both blocks.

`TestAddBlockAfterRollbackToOriginRejectsChainShortOfBlockZero` (`chain/origin_continuity_test.go`) drives exactly the sequence you measured:

- `AddBlock(number 1, arbitrary prev hash)` on a chain emptied back to origin must be rejected with `BlockNotFitChainTipError`, and the chain must still be at origin.
- `AddBlock(number 2, prev hash = that number-1 block)` -- the block that made the gap permanent, because it satisfies `blockNumberContiguous` against a tip of 1 -- must be rejected too, chain still at origin.
- Positive control: the same pair anchored at 0 (block 0, then block 1 chaining onto it) is accepted, the tip is block number 1, and the chain's first block, read back with an iterator **from origin**, is number 0. That last assertion is the one that distinguishes a short chain from a whole one; `tip.BlockNumber` cannot, which is why the existing case pinned acceptance without pinning its cost.

Red on `3459944f` (the merge of `main`, tolerance still 1): `AddBlock accepted block number 1 as the chain's first block; the chain is then permanently short block 0`. Green on `1873b408`.

The case you commented on flipped rather than being deleted: `genuine first block number 1 accepted` is now `first block number 1 rejected` (`wantReject: true`), so all three add paths -- `AddBlockHeader`, `AddBlock`, `AddRawBlocks` -- pin the rejection through the shared table, and `TestFreshChainAcceptsFirstBlockFromAnySource` still pins that a never-mutated chain accepts it, so the bulk importer stays unaffected.

Mutation checks on `1873b408`, with the tests in place:

- Restoring the `<=` tolerance (`blockNumber <= firstBlockNumber+1`) fails `TestAddBlockAfterRollbackToOriginRejectsChainShortOfBlockZero` plus the `first_block_number_1_rejected` subtest on all three add paths.
- Disabling the guard in `addBlockLocked` alone (leaving the header and raw-block guards in place) fails the new test plus all three `TestAddBlockAfterRollbackToOriginRequiresFirstBlock` rejection subtests, and nothing else -- so the new test is bound to the decoded-block path it exercises.

For the second assertion specifically, the pre-fix behaviour is the reproduction quoted in the other thread: on `2a81775b` both `AddBlock`s returned `<nil>` and the chain settled with tip 2 and first block 1.

Deliberately not covered: no test binds the first block's *hash* at origin. The `chain` package has no genesis hash to bind it to, so a block numbered 0 with an arbitrary hash is still accepted; that residual gap is stated on `firstBlockNumberValid` and in the PR body rather than tested.

wolf31o2 (approved):

PR #4206, opened 2026-09-12
test(database): pass the deposit policy at the committee-prune call sites

chadaibbhot-git:
Superseded by #4222: main now passes `requireKnownDeposits` at both committee-pruning test call sites. This draft's two-call repair is already included there.

The change restores the `applyTransactionCertificates` signature in `committee_prune_test.go`; it does not change production behavior. Related: #3950 and #3855.

## Summary by cubic

- Passes `requireKnownDeposits`, matching the live block-apply path these tests cover.
- The policy restores the signature without changing committee-authorization test behavior.
- Test-only change; no production files are touched.

## Summary by CodeRabbit

- Updates committee-pruning test calls for the required deposit-policy parameter while preserving existing behavior.


wolf31o2:
This has been fixed elsewhere.

PR #4216, opened 2026-09-12
fix(database): drive GetUtxosByRefs by tx_id, not OR predicate

chrisguiney:
## Summary

Stacked on #4211 (`fix/utxo-stake-refs-txid-index`), which this branch is
based on and which adds the `distinctUtxoTxIDs` helper this change reuses.
Part of the #4203 investigation.

Class follow-up to #4211/#4067: `GetUtxosByRefs` and `GetUtxosByRefsAsOf`
build the same OR-of-`(tx_id = ? AND output_idx = ?)` predicate gated by a
`deleted_slot`-based condition, and hit the identical SQLite planner
fallback away from the unique `tx_id_output_idx` index -- confirmed via
`EXPLAIN QUERY PLAN` in `TestGetUtxosByRefsUsesTxIDIndex`. Documented as a
follow-up on #4067 before this PR existed.

Both now share a new `utxoRefsByTxID` helper: look up by distinct `tx_id`,
then filter to the exact requested `(tx_id, output_idx)` pairs and to the
deleted-slot/as-of-slot conditions in Go.

## Quantified

`BenchmarkGetUtxosByRefs` against a real migrated SQLite store:

| rows    | refs/call | before      | after     | speedup |
|---------|-----------|-------------|-----------|---------|
| 100,000 | 4         | 28.4 ms/op  | 102 us/op | ~280x   |
| 100,000 | 50        | 172.0 ms/op | 352 us/op | ~490x   |
| 100,000 | 400       | 1.24 s/op   | 2.48 ms/op| ~500x   |
| 500,000 | 4         | 148.8 ms/op | 138 us/op | ~1080x  |
| 500,000 | 50        | 828.8 ms/op | 395 us/op | ~2100x  |
| 500,000 | 400       | 6.09 s/op   | 2.44 ms/op| ~2500x  |

`TestGetUtxosByRefsUsesTxIDIndex` pins both statements' query plans.
`TestGetUtxosByRefsReturnsSameRows` and `TestGetUtxosByRefsAsOfReturnsSameRows`
prove the rewrite returns identical rows to the OR-predicate form it
replaces (live-only filtering, the as-of slot/deletion window, a request
spanning multiple output indexes of one transaction, and a nonexistent ref).

## Why this matters beyond the point fix

`GetUtxosByRefs` is the batched UTxO-lookup API #4203's own investigation
named as the target for a previously-implemented-and-tested-but-reverted
optimization: batching `LedgerView.UtxoById`'s per-input database round
trips (currently one round trip per consumed/reference input) into one
`GetUtxosByRefs` call per block. That optimization was held back
specifically because it would have wired this exact query-plan defect into
the per-block validated-apply hot path. This PR removes that blocker; the
batching optimization itself is not attempted here.

## Validation

- `go build -tags dingo_extra_plugins ./...`
- `go vet -tags dingo_extra_plugins ./database/plugin/metadata/sqlstore/...`
- `gofmt -l` on changed files: clean
- `go test -tags dingo_extra_plugins -race ./database/plugin/metadata/sqlstore/...`: pass
- `go test -tags dingo_extra_plugins ./internal/test/conformance/...`: pass
- Causality: reverting the fix in place reproduces the legacy plan/benchmark
  numbers above.

Same pre-existing, unrelated `committee_prune_test.go` compile break noted
on #4211 applies here too (confirmed via the same local, uncommitted,
reverted-before-push workaround).

## Related

- #4203 (parent investigation)
- #4067 (the query-plan defect this and #4211 fix)
- #4211 (base branch; the `queryUtxoStakeRefs` sibling fix)

Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Rewrites `GetUtxosByRefs` and `GetUtxosByRefsAsOf` to look up UTxOs by distinct `tx_id` instead of an OR-of-`(tx_id, output_idx)` predicate, so SQLite's planner no longer abandons the unique `tx_id_output_idx` index. Lookup behavior is unchanged: both fetch all outputs of the referenced transactions and filter to the exact refs and deleted-slot conditions in Go, and `GetUtxosByRefsAsOf` again rejects `atSlot` values above `math.MaxInt64`.

- Benchmarks against a real migrated store: 500k rows / 400 refs drops from 6.09s to 2.4ms per call (~2500x).
- Adds query-plan, row-equivalence, and out-of-range `atSlot` tests covering multi-output refs, as-of slots, and nonexistent refs.
- Removes the query-plan blocker that had held back the per-block UTxO batching idea from issue #4203.

<sup>Written for commit b4616afb42ebdfc3aa750021e05dc550e5a8345d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved UTxO reference lookups for more consistent results across live, spent, deleted, and unavailable outputs.
  - Corrected historical (“as of” slot) lookups at slot boundaries.
  - Added validation for slot values outside the supported range.

- **Performance**
  - Improved lookup efficiency, particularly when resolving many references or references sharing transaction identifiers.


database/plugin/metadata/sqlstore/utxo.go
```diff
@@ -979,30 +1020,26 @@ func (s *Store) GetUtxosByRefsAsOf(
...
+	for _, u := range utxos {
+		outs, ok := wanted[string(u.TxId)]
+		if !ok {
+			continue
 		}
-		ret = append(ret, utxos...)
+		if _, ok := outs[u.OutputIdx]; !ok {
+			continue
+		}
+		if u.AddedSlot > atSlot {
```
wolf31o2, line 1042: This removes the previous checkedInt64(atSlot) validation. For atSlot > math.MaxInt64, the old SQL implementation returned an error because SQLite stores slots as signed integers; this in-memory comparison now accepts the out-of-range value and changes the API behavior. Preserve the overflow error contract before querying and cover it with a regression test.

wolf31o2 (changes requested):
GetUtxosByRefsAsOf no longer preserves the previous overflow contract for atSlot. The old implementation called checkedInt64 and returned an error when a uint64 slot exceeded SQLite's signed int64 range; the new in-memory filtering accepts that value and returns rows instead. Keep the checkedInt64 validation before the lookup and add a regression test for an overflowing atSlot.

database/plugin/metadata/sqlstore/utxo.go
chrisguiney: Restored in b4616afb: `GetUtxosByRefsAsOf` now calls `checkedInt64(atSlot)` before the tx_id lookup and returns its error, including for an empty ref list, matching the previous SQL-bound behavior.

`TestGetUtxosByRefsAsOfRejectsOutOfDomainSlot` covers `math.MaxInt64` (accepted, live row returned), `math.MaxInt64 + 1` and `math.MaxUint64` (rejected with the `exceeds int64` error), and a nil ref list.

wolf31o2 (approved):

Issue #3293, opened 2026-08-21
compose: allow Dingo's graceful shutdown deadline


wolf31o2:
## Problem

- Dingo's default graceful-shutdown timeout is 30 seconds.
- `docker-compose.yml` does not set `stop_grace_period`.
- Compose sends SIGKILL after its 10-second default.
- Database drain and component shutdown can be interrupted.

## Acceptance criteria

- Set `stop_grace_period` beyond Dingo's configured shutdown deadline.
- Keep deployment defaults synchronized with the node timeout.
- Add a configuration check that prevents the grace period from becoming shorter than the node deadline.

PR #4219, opened 2026-09-12
fix(node): bound shutdown phase-1 component stops

chrisguiney:
Shutdown phase 1 called the chainsync stall recycler, chain-selected-to-none worker, block forger, leader election, snapshot manager, and database lifecycle manager stops directly. Each cancels its own context and then waits for a goroutine with no deadline, so one goroutine that ignores cancellation holds `Node.Stop` indefinitely, and `internal/node` calls `Node.Stop` with no timeout of its own. Live restore/truncate already bounds the same waits with `stopWithDeadline`.

- Phase 1 runs these stops through `stopWithDeadline`, each with the time left on the single shutdown deadline rather than a fresh timeout per stop. An unfinished stop returns `errStorageDrainUnconfirmed`, and phase 3 then skips the ledger state close, database close, and plugin host shutdown, since the stuck component may still be using them. An unconfirmed ledger state close already skipped the last two.
- The phase-1 list is the two context-owned workers followed by `quiesceComponentStops`, so the Leios pipeline and vote managers, previously stopped only by live restore/truncate and startup rollback, now also stop before phase 3.
- The recycler and selected-to-none worker reach node components only under `liveLifecycleMu`, which shutdown holds throughout phase 1, so bounding their wait cannot race teardown. Their doc comments and the `ARCHITECTURE.md` shutdown flow are updated.

Tests in `node_shutdown_test.go`:

- `TestNodeStopEscalatesWhenPhase1ComponentNeverReturns` and `TestNodeStopSkipsDatabaseCloseWhenPhase1DrainUnconfirmed`: a never-returning stop makes `Node.Stop` return `errStorageDrainUnconfirmed` and skip the database close. Both fail with phase 1 reverted to direct calls; the second also fails with only the phase-3 gate reverted.
- `TestNodeStopSkipsLedgerStateCloseWhenPhase1DrainUnconfirmed`: a stop wedged past the deadline leaves a real `LedgerState` open. Fails with the ledger state guard removed. `TestNodeStopClosesLedgerStateWhenPhase1DrainConfirmed` is the confirmed-drain control.
- `TestNodeStopBoundsPhase1StopsByOneShutdownDeadline`: 20 wedged stops return within the deadline. With a fresh timeout per stop, `Node.Stop` takes 2s and the test fails.
- `TestShutdownPhase1ComponentStopsCoverQuiesceComponentStops`: fails without the Leios managers in the phase-1 list. `TestShutdownPhase1ComponentStopsCoverEveryUnboundedStop` and `TestShutdownPhase1ComponentStopsSkipsAbsentComponents` pin the list contents.

DevNet was not run; the change is limited to shutdown ordering and error reporting.

Part of #1649 (shutdown half of the context-free waits example). Related to #3293, which sizes the Compose stop grace period against this deadline.

https://claude.ai/code/session_01UKHBdsLgideJfpVg2jXwAD


node_shutdown.go
```diff
@@ -353,7 +379,11 @@ func (n *Node) shutdown() error {
 	// Phase 3: Flush state and close database
 	n.config.logger.Info("shutdown phase 3: flushing state")
 	phase3Start := time.Now()
-	ledgerStateDrainConfirmed := true
+	// Starts from phase1DrainConfirmed: a phase-1 component that never
+	// confirmed stopping may still be using n.db, exactly the same danger an
+	// unconfirmed ledgerState close guards against below, so either failure
+	// must skip the database close and plugin host shutdown that follow.
+	ledgerStateDrainConfirmed := phase1DrainConfirmed
```
wolf31o2, line 386: The phase-1 drain can time out while its Stop goroutine is still running. The phase-3 cleanup below can then close n.ledgerState while Leios managers still access it, causing use-after-close behavior during shutdown. Guard this close with phase1DrainConfirmed and cover the unconfirmed path with a regression test.

wolf31o2 (changes requested):
The phase-1 drain can time out while its Stop goroutine is still running. The phase-3 cleanup below can then close n.ledgerState while Leios managers still access it, causing use-after-close behavior during shutdown. Guard this close with phase1DrainConfirmed and cover the unconfirmed path with a regression test.

node_shutdown.go
chrisguiney: Fixed in 8eaa34f3. Phase 3 now skips `LedgerState.Close` when `phase1DrainConfirmed` is false and reports `ledger state close skipped: phase 1 drain unconfirmed`, matching Restore/Truncate, which skip `closeStorageForLiveLifecycleOp` on the same `errStorageDrainUnconfirmed`. The database close and plugin host shutdown were already skipped on that path.

Tests in `node_shutdown_test.go`:
- `TestNodeStopSkipsLedgerStateCloseWhenPhase1DrainUnconfirmed` wedges a phase-1 stop past the deadline through `Node.Stop` with a real `LedgerState`, and asserts shutdown neither starts nor runs its close. It fails with the guard removed.
- `TestNodeStopClosesLedgerStateWhenPhase1DrainConfirmed` is the control: with every phase-1 stop returning, the ledger state is closed.

The other resources released after the phase-1 list were checked for the same hazard. The metadata and blob stores are stopped by `pluginHost.Stop`, already gated. Mempool and EventBus end in lock-guarded terminal states that reject late callers. ConnectionManager, peer governor, and the API servers are not held by the phase-1 components. `ARCHITECTURE.md` records this.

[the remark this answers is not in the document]
chrisguiney: Fixed in 8eaa34f3. Phase 3 now skips `LedgerState.Close` when `phase1DrainConfirmed` is false and reports `ledger state close skipped: phase 1 drain unconfirmed`, matching Restore/Truncate, which skip `closeStorageForLiveLifecycleOp` on the same `errStorageDrainUnconfirmed`. The database close and plugin host shutdown were already skipped on that path.

Tests in `node_shutdown_test.go`:
- `TestNodeStopSkipsLedgerStateCloseWhenPhase1DrainUnconfirmed` wedges a phase-1 stop past the deadline through `Node.Stop` with a real `LedgerState`, and asserts shutdown neither starts nor runs its close. It fails with the guard removed.
- `TestNodeStopClosesLedgerStateWhenPhase1DrainConfirmed` is the control: with every phase-1 stop returning, the ledger state is closed.

The other resources released after the phase-1 list were checked for the same hazard. The metadata and blob stores are stopped by `pluginHost.Stop`, already gated. Mempool and EventBus end in lock-guarded terminal states that reject late callers. ConnectionManager, peer governor, and the API servers are not held by the phase-1 components. `ARCHITECTURE.md` records this.

wolf31o2 (approved):

Issue #3860, opened 2026-09-04
ledger: PlutusV3 withdrawal script fails with error explicitly called on a canonical mainnet Minswap order, freezing the tip


wcatz:
## Summary

A mainnet block producer running dingo rejects a canonical block containing a
real Minswap "Forward Adapter Order" transaction. Dingo's PlutusV3 evaluator
fails a withdrawal (`Rewarding`) script with `error explicitly called`; the
network accepted the transaction and it is on the canonical chain. The node
re-rejects the same block indefinitely at the same applied tip, so the tip
freezes and forging stops. Same failure class/symptom as #3780, but a
different transaction, different script, and (per below) a different cause.

## Ruled out

This is **not** the plutigo bignum-chunking bug fixed in `v0.5.1`
(`blinklabs-io/plutigo#391`, closes #3780). I deployed `plutigo v0.5.1` (with
`gouroboros` held at `v0.202.4`, otherwise unchanged) directly onto the
affected node and restarted it — the node hit the exact same rejection on the
exact same transaction immediately. That bug requires an integer whose
magnitude is `>= 64` bytes to trigger a `serialiseData` divergence; the
redeemer that fails here is a trivial `I 0` (a single-byte-magnitude
constant), nowhere near that threshold. So whatever this is, it's a separate
evaluator divergence that happens to also throw Plutus's generic
"error explicitly called" message and also happens to be a `Rewarding`
(withdrawal) purpose script — coincidental symptom overlap with #3780, not
the same mechanism.

## Observed

```
time=2026-09-04T08:27:50.381Z level=WARN msg="deterministic transaction validation rejected a block on the primary chain; rewinding so chain selection can offer another candidate" component=ledger tx_hash=2d295dbffc2898b14ccc9b359a342305006fdcf6ccd7907b444419520a98f351 failing_block_slot=196783015 rewind_target_slot=196782988 rewind_target_hash=cc9d178de127122be50ce4860aacbf560b5dc2caddf4daedfe708e856dd107ad error="conway plutus validation: plutus script failed (hash=f3c398920fc488cb995db26f09f7e7320b8538b123f300799bf101d5, tag=3, index=0): execute script: error explicitly called"

[...]

PR #4223, opened 2026-09-12
test(ledger): pin preview Babbage Plutus budget fixture

wolf31o2:
Preview transaction `eab27325c569121613728db87a4d8333ce74cc857bbe4991875ceab8787d0213` at slot 41098839 (epoch 475, protocol version 8) lists the same required signer hash twice. cardano-ledger holds `reqSignerHashes` as a `Set` and renders `txInfoSignatories` with `Set.toList`, so the reference evaluator sees one signatory. Rendering both made the spending validator take 621 extra CEK steps and 16 extra builtin calls — 16359467 CPU and 62240 memory over the declared 908604883 CPU / 3389879 memory — which rejected a block the network accepted and wedged a preview replay.

The regression fixture carries the transaction, the four transactions that funded its two spent inputs and two reference inputs, and epoch 475's PlutusV1 and PlutusV2 cost models. The test asserts the evaluated execution units equal the producer-declared budget exactly, so it catches an undercharge as well as an overcharge.

No production change: the deduplication landed in blinklabs-io/gouroboros#2230 and is present in the pinned gouroboros. Reverting that pin to `v0.202.8` makes the test fail with the reported overshoot of 924964350 CPU / 3452119 memory.

Related: #3935

#3627 reports a budget overshoot on a different preprod transaction. This fixture does not establish that transaction's cause and leaves #3627 open.

The Preprod `execute script: error explicitly called` rejections at slots 133016611 and 133434403 are recorded on #3860 rather than on #3935. Their cause is a separate defect: a script context carrying the definite-length CBOR a transaction was built with, where cardano-ledger rebuilds every script-visible value and so always presents the indefinite-length form, which changes what `serialiseData` returns. blinklabs-io/gouroboros#2304 addresses that. This fixture does not cover it.


wolf31o2:
@coderabbitai review

ledger/eras/babbage_plutus_preview_fixture_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+// TestEvaluateTxBabbagePreviewDuplicateRequiredSigner pins the execution units
+// of preview transaction eab27325... (slot 41098839, epoch 475, protocol
+// version 8) against the budget its producer declared. The transaction body
+// lists the same required signer hash twice. cardano-ledger holds
+// reqSignerHashes as a Set and renders txInfoSignatories with Set.toList, so
+// the reference evaluator sees one signatory; rendering both makes the spending
+// validator take 621 extra CEK steps and 16 extra builtin calls, for 16359467
+// CPU and 62240 memory over the declared budget, which rejects a block the
+// network accepted and wedges a preview replay. See blinklabs-io/dingo#3935.
```
chrisguiney, line 106: This test covers #3935's Preview case, but GitHub also links #3627 as closed by this PR: "Does not close #3627" in the description matches the closing-keyword pattern, and #3627 is still open pending the Preprod Mithril bootstrap and 30-minute live-serve run. Reword that line without close/fix/resolve before the reference. #3935 is also the only issue tracking the Preprod "error explicitly called" rejections at slots 133016611 and 133434403, which still reproduce with this fix; move those to a new issue before merging closes #3935.

chrisguiney (dismissed):
Requesting changes for issue linkage only: merging would close #3627, which the description says this PR does not close; details inline. The test passes at head and on the merge with main, fails with 924964350 CPU / 3452119 mem when only the signatory dedup is reverted, and its fixtures match Koios byte for byte.

ledger/eras/babbage_plutus_preview_fixture_test.go
wolf31o2: Summary disposition changed. Review again, please.

arepala-uml (approved):

Issue #4156, opened 2026-09-09
ouroboros: NtC LocalStateQuery connection torn down within ~100ms of a concurrent block-apply commit


arepala-uml:
## Summary

An NtC LocalStateQuery session's underlying connection is intermittently torn
down (client sees `protocol is shutting down`) specifically around the moment
dingo applies a new block to the ledger. Confirmed via precise timestamp
correlation, not assumed from proximity: **11 of 13 observed disconnects
occurred within 50-170ms of dingo's own "chain extended, new tip" log line**
(the other 2 were 1.5s and 7.5s away). This is not the same issue as
blinklabs-io/dingo#4082 (`GetUTxOWhole` exceeding gouroboros' 120s NtC
timeout at scale) — these disconnects happen on small, fast queries
(`GetCurrentProtocolParams`) that normally complete in well under a second,
and the timing lines up with block application, not query duration. It is
also unrelated to blinklabs-io/dingo#4132 (NtN peer ChainSync re-sync racing
its own in-flight `MsgRequestNext`) -- different protocol (LocalStateQuery,
not ChainSync), different connection class (an NtC client tool, not an NtN
peer), and a different, so-far-unexplained mechanism.

## Evidence

Found via `cmd/node-parity watch --mode=incremental` (blinklabs-io/dingo#1900),
which holds one persistent NtC connection per node for the life of a session
and re-`Acquire`s a fresh point on it for every block (rather than dialing a
new connection per block, which an earlier version of the tool did -- ruled
out as the primary cause, see below). Preview, dingo built from a fresh
Mithril bootstrap plus roughly one hour of live catch-up, comparing against a
real cardano-node.

Session log (`nodeparity: incremental session ended, reconnecting`), each
paired with the nearest `chain extended, new tip: ...` line from dingo's own
log:

```
reconnect 14:43:31.430 -> chain extended 14:43:31.466  (+0.036s)
reconnect 14:43:58.177 -> chain extended 14:43:58.220  (+0.043s)
reconnect 14:44:15.212 -> chain extended 14:44:15.264  (+0.052s)
reconnect 14:45:45.437 -> chain extended 14:45:45.603  (+0.166s)
reconnect 14:46:24.054 -> chain extended 14:46:24.102  (+0.048s)
reconnect 14:46:47.280 -> chain extended 14:46:47.348  (+0.068s)
reconnect 14:48:40.595 -> chain extended 14:48:40.647  (+0.052s)
[...]

PR #4233, opened 2026-09-13
fix(ledger): skip genesis UTxO insertion after a Mithril bootstrap

### LINKED ISSUE #4151: ledger: Mithril bootstrap resurrects already-spent genesis UTxOs as live
## Summary

After a Mithril bootstrap, `LedgerState.createGenesisBlock` (`ledger/chainsync.go:4863`)
unconditionally recreates every Byron and Shelley genesis-config UTxO as a live,
spendable row, without checking whether the just-imported Mithril ledger snapshot
already reflects those outputs as spent. This resurrects genesis-era UTxOs that
the real chain spent long ago, so a Mithril-bootstrapped node's live UTxO set
silently diverges from a real cardano-node's at the exact same tip.

## Evidence

Found via `cmd/node-parity check` (blinklabs-io/dingo#1900) run against a
freshly Mithril-bootstrapped Preview `dingo` (bootstrap landed at slot
122264150 / epoch 1415) and a real `cardano-node` (v11.0.1), both at live tip.
At slot 122311344 (block 4647006), `dingo`'s paginated UTxO walk reports 8
UTxOs cardano-node's `GetUTxOByTxIn` says do not exist:

```
11eb3fa97ca673cd495336c9e052ee127cd5b93063c1d2a727fee0b99640dda2#0  FHnt4NL7yPXtmi4mAjD43V3NB3shDs1gCuHNcMLPsRWjaw1b2yRV2xad8S8V6aq  0
38f13eac73c749afe2c15ea2cc4958e3057a71b5962fc9193d4538951955dc9d#0  FHnt4NL7yPXk7D87qAWEmfnL7wSQ9AzBU2mjZt3eM48NSCbygxgzAU6vCGiRZEW  0
4843cf2e582b2f9ce37600e5ab4cc678991f988f8780fed05407f9537f7712bd#0  FHnt4NL7yPXvDWHa8bVs73UEUdJd64VxWXSFNqetECtYfTd9TtJguJ14Lu3feth  30000000000000000
4ceb4298a5d404ad5400513bd57f93693350ee1f499bf5b116db67a49e7e33f9#0  FHnt4NL7yPXjpZtYj1YUiX9QYYUZGXDT9gA2PJXQFkTSMx3EgawXK5BUrCHdhe2  0
918d475d9a22799647dc55e51dd8b65939a79ec0daf4406e4c393557f62398ea#0  FHnt4NL7yPXvNSRpCYydjRr7koQCrsTtkovk5uYMimgqMJX2DyrEEBqiXaTd8rG  0
a3a0802fcb98e83efe0db9aa1009fe3ffc0ac8145ff289e32d59a1cd1c883445#0  FHnt4NL7yPY9rTvdsCeyRnsbzp4bN7XdmAZeU5PzA1qR2asYmN6CsdxJw4YoDjG  0
b3bab19e7e9f0d0b2c93a87766bb33277a57a691b5733207aa9d275d30f50696#0  FHnt4NL7yPXpazQsTdJ3Gp1twQUo4N5rrgGbRNSzchjchPiApc1k4CvqDMcdd7H  0
b44bfcebd25b68d82ac6aedf2d5d72390cb191a1359120fa9383a4e15d3e68f7#0  FHnt4NL7yPXtNo1wLCLZyGTMfAvB14h8onafiYkM7B69ZwvGgXeUyQWfi7FPrif  0
```

All 8 addresses and amounts match `config/cardano/preview/byron-genesis.json`'s
`nonAvvmBalances` section exactly (verified byte-for-byte, all 8 entries in that
map). This is not a coincidence of similar-looking addresses: the amounts
dingo reports today, at block 4647006, are identical to the raw genesis
declaration, meaning dingo has never observed these outputs as spent, ever.

## Root cause

`createGenesisBlock` (`ledger/chainsync.go:4863`) is called at every startup.
When `ls.currentTip.Point.Slot > 0` (true immediately after a Mithril
bootstrap) and `ls.db.HasGenesisCbor(0, genesisHash[:])` is false — which the
code's own comment says is expected after a Mithril bootstrap ("imports ledger
state and ImmutableDB blocks but does not create the synthetic genesis
block") — it falls through to the same code path a from-genesis sync uses to
seed genesis UTxOs (`chainsync.go:4911-4976`): it calls
`byronGenesis.GenesisUtxos()` / `shelleyGenesis.GenesisUtxos()` and stores
every one of them as a live UTxO via `db.SetGenesisTransaction`, with no check
against the ledger state the Mithril snapshot already imported.

A from-genesis sync doesn't hit this bug because it then replays every real
historical block, including whichever one spends a given genesis output — the
UTxO briefly exists and is then correctly consumed. A Mithril-bootstrapped
node skips straight to a certified snapshot from deep in the chain's history
and never replays those early blocks, so a genesis output already spent
before the snapshot's slot is inserted as live and never gets a chance to be
marked spent.

## Acceptance criteria

- `createGenesisBlock`'s post-Mithril-bootstrap branch does not insert a
  genesis UTxO that the imported ledger state already shows as spent (or,
  more simply: does not insert genesis UTxOs as live rows for a
  Mithril-bootstrapped node at all, since the synthetic genesis block only
  needs to exist for structural/hash purposes, not to re-donate already-spent
  funds).
- A Mithril-bootstrapped Preview node's live UTxO set matches a real
  cardano-node's for every Byron/Shelley genesis-declared address.
- A regression test bootstraps (or simulates bootstrapping) past a block that
  spends a genesis UTxO and asserts that UTxO is not resurrected as live.
- `cmd/node-parity check` between a Mithril-bootstrapped dingo and a real
  cardano-node reports no UTxO-only-in-dingo divergence for genesis addresses.

arepala-uml:
## Summary

Fixes #4151. `createGenesisBlock` unconditionally inserted every configured genesis UTxO whenever genesis CBOR was missing from the blob store, including on a chain that already has a non-zero tip — exactly the case immediately after a Mithril bootstrap import, which populates ledger state and ImmutableDB blocks without ever calling this function.

That resurrects genesis UTxOs the chain has since spent: dingo answers `GetUTxOByAddress`/`GetUTxOWhole` with UTxOs a real cardano-node correctly reports as already spent.

## Root cause / reproduction

Confirmed live against a real Preview `cardano-node` during this issue's investigation, using the same 8 genesis UTxOs the issue reported: after a fresh Mithril bootstrap, dingo reported all 8 as still live, while the reference node correctly showed them spent.

## Fix

`createGenesisBlock` now captures whether the chain is already past slot 0 (`bootstrappedFromMithril`) once at the top, and skips the genesis-UTxO insertion loop when true — genesis CBOR is still recorded (needed for hash verification on restart), but no UTxOs are (re)inserted for a chain a Mithril bootstrap already populated.

## Tests

- `TestCreateGenesisBlockSkipsUtxoInsertionAfterMithrilBootstrap`: builds a preview-config `LedgerState` with a non-zero current tip and no genesis CBOR pre-seeded, calls `createGenesisBlock`, and asserts genesis CBOR now exists structurally but the 8 real genesis UTxOs are NOT live. Verified it fails against the prior behavior (via a stash/apply round-trip) and passes with this change.

## Validation

- `go build ./...`, `go vet ./ledger/...`: clean.
- `go test ./ledger/...`: all green, including the existing `TestCreateGenesisBlock*` suite (unchanged behavior for a normal, non-Mithril genesis start).
- `go test ./...`: all green except `bin.TestEntrypointForwardsSignalsDuringMithrilBootstrap`, a pre-existing flaky subprocess-timing test — confirmed by re-running it in isolation (passes) and confirmed unrelated to this change.
- `make lint` (golangci-lint, nilaway, modernize, import-boundaries): nilaway reports 25 pre-existing findings, none in `ledger/chainsync.go` or `ledger/genesis_utxo_test.go` — confirmed identical count/content on a clean `origin/main` checkout with no changes at all, so this is baseline noise, not something introduced here.
- Live re-test: rebuilt `dingo` with this fix (combined with the #4156 fix from #4232), ran a fresh Mithril bootstrap against real Preview data, and confirmed via direct SQLite inspection that genesis UTxOs are correctly absent after bootstrap.

---
## Summary by cubic
Fixes #4151 by making `createGenesisBlock` skip inserting genesis UTxOs, pools, delegations, and DReps when the chain tip is already past slot 0 (i.e., after a Mithril bootstrap), so spent or retired genesis declarations no longer resurrect as live. The synthetic genesis CBOR is still stored.

- The `bootstrappedFromMithril` guard now covers the genesis UTxO loop, `SetGenesisStaking`, and `SetGenesisGovernance`.
- Adds regression tests for UTxO (preview), staking (devnet), and governance (synthetic DRep) paths; existing from-genesis behavior is unchanged.
- `ARCHITECTURE.md` documents the expanded skip next to the fresh-genesis-init description.

<sup>Written for commit 7cc10769b41c180df97e1f9bfd53fda3ca86473d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4233?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of nodes restored from a Mithril snapshot.
  - Prevented previously spent or changed genesis UTxOs, staking pools, delegations, and governance records from being recreated as active entries.
  - Preserved required genesis metadata while maintaining correct imported ledger state.
  - Ensured repeated initialization remains safe and consistent.

- **Documentation**
  - Clarified genesis initialization behavior for Mithril-bootstrapped nodes.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1143,6 +1143,16 @@ When `Node.Run()` is called, components are initialized in this order:
     Fresh genesis initialization persists both genesis UTxOs and the effective
     Shelley staking declarations, including network-specific `extraConfig`
     pools and delegations, before snapshot capture.
+    A Mithril-bootstrapped node instead reaches `createGenesisBlock` with its
+    tip already past slot 0 and no genesis CBOR yet stored (the bootstrap
+    imports ledger state and ImmutableDB blocks but never runs this
+    function). It still stores the synthetic genesis block CBOR
+    unconditionally — other code depends on it existing structurally — but
+    skips (re-)inserting genesis UTxOs, pools, delegations, and DReps as live
+    rows: the imported ledger snapshot (`ledgerstate/import.go`) already
+    reflects their correct current state as of the bootstrap point, and this
+    function has no way to tell a still-live genesis declaration from one
+    already spent/retired/changed since (blinklabs-io/dingo#4151).
     For networks with a real Byron genesis, an empty database retains a Byron
     epoch cache until the on-chain Shelley boundary is observed. A configured
     experimental Shelley hard-fork epoch is the explicit exception used by
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -5235,7 +5235,17 @@ func (ls *LedgerState) createGenesisBlock() error {
 			)
 		}
 
-		// Load genesis staking data (pool registrations + delegations)
+		// Load genesis staking data (pool registrations + delegations) --
+		// but only for a from-genesis sync, for the same reason genesis
+		// UTxO insertion is skipped above (blinklabs-io/dingo#4151).
+		// SetGenesisStaking upserts current-state pool/delegation rows
+		// (ON CONFLICT ... DO UPDATE), and a Mithril-bootstrapped node's
+		// imported ledger snapshot (ledgerstate/import.go's
+		// importCertState) already reflects the correct *current*
+		// pool/delegation state as of the bootstrap point -- reapplying
+		// stale genesis-config values here would resurrect a pool or
+		// delegation genuinely retired/changed long before the bootstrap
+		// point, the same resurrection bug #4151 found for UTxOs.
 		genesisPools, poolDelegators, err := shelleyGenesis.InitialPools()
 		if err != nil {
 			return fmt.Errorf("parse genesis staking: %w", err)
@@ -5244,8 +5254,8 @@ func (ls *LedgerState) createGenesisBlock() error {
 		if err != nil {
 			return fmt.Errorf("parse genesis stake delegations: %w", err)
 		}
-		if len(genesisPools) > 0 ||
-			len(genesisStake) > 0 {
+		if !bootstrappedFromMithril &&
+			(len(genesisPools) > 0 || len(genesisStake) > 0) {
 			ls.config.Logger.Info(
 				fmt.Sprintf(
 					"loading genesis staking: %d pools, %d delegations",
@@ -5268,9 +5278,12 @@ func (ls *LedgerState) createGenesisBlock() error {
 		// Load Conway genesis bootstrap data (initial DReps and
 		// stake/vote delegations). The conway-genesis.json may declare
 		// pre-existing DReps and delegations for test networks; mainnet
-		// has none.
+		// has none. Same bootstrappedFromMithril guard as genesis
+		// staking above, for the same reason: SetGenesisGovernance
+		// upserts current-state DRep/delegation rows that a Mithril
+		// import has already populated correctly.
 		conwayGenesis := ls.config.CardanoNodeConfig.ConwayGenesis()
-		if conwayGenesis != nil &&
+		if !bootstrappedFromMithril && conwayGenesis != nil &&
 			(len(conwayGenesis.InitialDReps) > 0 ||
 				len(conwayGenesis.Delegs) > 0) {
 			ls.config.Logger.Info(
@@ -5381,7 +5394,11 @@ func (ls *LedgerState) ensureGenesisCommittee(txn *database.Txn) error {
 			Credential:    member.ColdCredHash,
 		}).Key()] = struct{}{}
 	}
-	newMembers := make([]*models.CommitteeMember, 0, len(conwayGenesis.Committee.Members))
+	newMembers := make(
+		[]*models.CommitteeMember,
+		0,
+		len(conwayGenesis.Committee.Members),
+	)
 	for raw, expiry := range conwayGenesis.Committee.Members {
 		tag, hash, err := parseGenesisCommitteeCredential(raw)
 		if err != nil {
--- a/ledger/chainsync.go
+++ b/ledger/chainsync.go
@@ -4993,7 +4993,16 @@ func (ls *LedgerState) createGenesisBlock() error {
 		return fmt.Errorf("get genesis block hash: %w", err)
 	}
 
-	if ls.currentTip.Point.Slot > 0 {
+	// bootstrappedFromMithril is true only on the fall-through path below:
+	// a chain already past slot 0 whose genesis CBOR was never created,
+	// which happens specifically when a Mithril bootstrap imported ledger
+	// state and ImmutableDB blocks without going through this function
+	// first. Captured once, here, rather than re-read inline further
+	// down: this function's only caller runs it once at startup before
+	// any concurrent chain-extension could move currentTip, so a single
+	// snapshot is representative for the whole call.
+	bootstrappedFromMithril := ls.currentTip.Point.Slot > 0
+	if bootstrappedFromMithril {
 		// Validate existing chain data matches the current genesis config.
 		// If genesis CBOR exists in the blob store with the expected hash,
 		// the database was created with a matching genesis. Older databases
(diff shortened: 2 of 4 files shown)


arepala-uml:
## Proof of test: live fresh Mithril bootstrap on Preview

#4151's fix only executes during a **fresh** Mithril bootstrap (`createGenesisBlock`'s `bootstrappedFromMithril` skip of genesis UTxO/staking/governance re-insertion) — every other live instance run so far today reused an already-bootstrapped data directory, so that code path had only unit-test coverage, not live proof. To close that gap, ran a real fresh bootstrap today against Preview:

```
dingo mithril sync --network preview --data-dir .
```

completed cleanly (epoch 1419, snapshot hash `f26e8a9a36ae...`, ~15.5GB imported):

```
time=2026-09-13T12:00:41.331-05:00 level=INFO msg="Mithril bootstrap complete" component=mithril epoch=1419 immutable_file_number=28389 index_rebuild_elapsed=1m0.290427917s lazy_index_rebuild_mode=maintenance
```

Then started `dingo serve` against that freshly-imported data directory. At startup, `createGenesisBlock` hit exactly the fall-through path this fix targets, and logged the fixed behavior:

```
time=2026-09-13T12:17:59.130-05:00 level=WARN msg="slot-0 CBOR exists but does not match synthetic genesis hash, creating genesis block" component=ledger expected_hash=83de1d7302569ad56cf9139a41e2e11346d4cb4a31c00142557b6ab3fa550761
time=2026-09-13T12:17:59.130-05:00 level=INFO msg="genesis block CBOR missing, creating it now" component=ledger
time=2026-09-13T12:17:59.130-05:00 level=INFO msg="creating 8 genesis UTxOs (8 Byron, 0 Shelley)" component=ledger
time=2026-09-13T12:17:59.131-05:00 level=INFO msg="stored synthetic genesis block CBOR; skipped inserting genesis UTxOs (already reflected by the Mithril-imported ledger snapshot)" component=ledger genesis_transactions=8 genesis_utxos=8 treasury=0 reserves=15000000000000000
```

Not just trusting the log line: queried the resulting `metadata.sqlite` directly and read-only (no live server touched) to confirm the database itself reflects the skip, not just a claim of it:

```
$ sqlite3 "file:metadata.sqlite?mode=ro&immutable=1" "SELECT COUNT(*) FROM utxo WHERE added_slot = 0;"
0
```

Zero UTxO rows at genesis slot (0) exist in this freshly-bootstrapped database — the same 8 genesis UTxOs this issue reported as wrongly-live are not present at all, confirming the insertion loop was genuinely skipped rather than merely logged as skipped.

Preview declares zero genesis pools/delegations/DReps, so the companion staking/governance skip (the same fix's second half) has no observable effect here — that half is covered by `TestCreateGenesisBlockSkipsGenesisStakingAfterMithrilBootstrap`/`...GenesisGovernance...` against devnet's real genesis pool+delegation, not by this live run.

Combined with the existing unit tests (`TestCreateGenesisBlockSkipsUtxoInsertionAfterMithrilBootstrap` and the two staking/governance tests above, all of which fail against the pre-fix code and pass at head), this closes the one gap in #4151's validation: the fix's actual trigger condition — a real fresh Mithril bootstrap — has now been exercised live against a real Preview snapshot, not just simulated in a unit test.

wolf31o2 (approved):

PR #4240, opened 2026-09-13
fix(ledger): preserve snapshot pointer stake

### LINKED ISSUE #3975: ledger: the Mithril bootstrap anchor determines the epoch reward pot — two Preview nodes anchored 12,823 slots apart compute different reward pots and both reject the same canonical withdrawal
## Summary

Two Preview nodes, same chain, same epoch boundary (1411 → 1412), differing only in
the slot their Mithril bootstrap was anchored at, compute **different epoch reward
pots** and therefore different reward-account balances. Both balances are above
canonical, so both nodes reject the same canonical block on the Conway exact-amount
withdrawal rule and neither can pass it.

This is the #3628 shape (local balance *above* the canonical withdrawal amount), but
reported here as a **controlled two-node comparison** rather than a single node
against an explorer. Holding the chain, the epoch, and the node build essentially
constant and varying only the Mithril anchor changes the reward pot by ~1.0 ADA-scale
in absolute terms and changes one account's balance by 33,867 lovelace — which is
what makes the Mithril import, rather than the reward formula, the variable under
suspicion.

## The rejected block

Canonical Preview block, confirmed via an independent chain index:

```
abs_slot      122000217
block_height  4635384
hash          b4a3ee322c6b592f...
```

Transaction:

```
f1a1cfdfe6e48f538cedcb2e119d71b4e675dffbed91b8796766649158cb7f8b
withdrawal from stake_test1uz7xx6hy2xnnrmz0av0xl7qn9vdkhage7myf0nd49e7mvcg6z0smn
provided 752965022
```

The account's canonical history confirms 752,965,022 was the exact balance: lifetime
`rewards` equals lifetime `withdrawals` (477,827,581,645) and `rewards_available` is
now 0, i.e. this withdrawal drained the account exactly, as `UtxoValidateWithdrawals`
requires pre-Dijkstra.

Both nodes disagree, in the same direction, by different amounts:

| | reported balance | over-count vs canonical |
|---|---|---|
| node A | 753,213,870 | **+248,848** |
| node B | 753,247,737 | **+282,715** |

```
node A: conway utxo validation rule 48: incorrect withdrawal amount for
        provided 752965022, balance 753213870
node B: conway utxo validation rule 47: incorrect withdrawal amount for
        provided 752965022, balance 753247737
```

(The rule index differs only because the two nodes carry different gouroboros
patch releases; it is the same `shelley.IncorrectWithdrawalAmountError`.)

Each value is **deterministic per node** — node A reported 753,213,870 on two
separate crossings of the boundary 24 minutes apart — and **divergent across
nodes**.

## The reward pot itself diverges

Both nodes logged the same epoch-boundary reward application
(`reward_snapshot_epoch=1409 performance_epoch=1410 pots_epoch=1411`, 680 pools):

| quantity | node A | node B | delta (B − A) |
|---|---|---|---|
| `total_reward_pot` | 15,326,109,639,528 | 15,327,111,407,229 | **+1,001,767,701** |
| `available_rewards` | 12,260,887,711,623 | 12,261,689,125,784 | +801,414,161 |
| `effective_rewards` | 375,581,661,952 | 375,610,821,379 | +29,159,427 |
| `undistributed_rewards` | 11,885,306,049,671 | 11,886,078,304,405 | +772,254,734 |
| epoch-1411 snapshot `total_stake` | 1,537,854,348,954,485 | 1,537,801,401,196,165 | −52,947,758,320 |

The only material difference between the two nodes is the Mithril trust boundary
they were bootstrapped at — both inside epoch 1409:

```
node A: mithril_ledger_slot=121763516
node B: mithril_ledger_slot=121776339      (12,823 slots later)
```

Both then followed the chain live from their own anchor through the 1410 and 1411
boundaries, so every input to the 1411 → 1412 reward round was derived locally,
downstream of a differing imported base. The stake snapshot for epoch 1411 already
differs by ~52.9 billion lovelace, so the divergence is present in the snapshot
inputs, not only in the pot arithmetic.

## Why the node never recovers

`shelley.IncorrectWithdrawalAmountError` is classified as a deterministic validation
failure, so recovery rewinds the primary chain and replays. On `main` this is explicit
and intentional (`ledger/replay_recovery.go:406-444`, `isRewardWithdrawalMismatch`),
including for this exact direction:

> "…the Shelley-family UTxO rule returns `shelley.IncorrectWithdrawalAmountError`
> when the amount does not satisfy the era's required relationship to that balance
> at all -- which under the exact-amount rule includes an amount below the balance,
> the shape observed on Preprod in issue #3628."

But the comment immediately below it also states why the rewind cannot work here
(`ledger/replay_recovery.go:414-417`):

> "…because a reward balance is derived from epoch-boundary accounting rather than
> from the UTxO window the producer-recovery path rebuilds, so no local replay can
> change either verdict."

Replaying the same blocks re-derives the same reward numbers by construction, so the
rewind is provably non-converging for this class. Observed: node A rejected slot
122000217, rewound, resynced, and rejected it again with a byte-identical balance.

### Amplifier: the rewind lands below the epoch boundary, so every attempt re-pays the rollover

The recovery rewind is bounded by k (432) and is **not clamped to the epoch
boundary**. The failing block sits 3,417 slots past the 1411/1412 boundary
(121996800), so a 432-block rewind lands roughly 22,000 slots *below* it — observed
rewind targets 121987906, 121975783, 121974264. Re-syncing therefore re-executes the
full epoch rollover on every recovery attempt.

On these nodes that rollover is the pre-#3782 `GetAccountsByCredential` stall (#3650):

| | rollover duration (measured) |
|---|---|
| node A | 13m25s, 13m47s, 13m54s |
| node B | 17m11s, 20m35s |

During it the applied tip is pinned at the last block of the previous epoch
(slot 121996797) with `event delivery stalled: subscriber not draining
type=ledger.chainsync` every 30 s — indistinguishable from a hang. **Both nodes
froze at exactly the same slot**, which is what prompted this investigation.

The practical consequence is that the retry period for a reward-state divergence is
~20 minutes rather than seconds, and any tip-stall watchdog with a threshold under
that will restart the node mid-rollover, discarding the work and guaranteeing it is
never completed. #3782 fixes the rollover cost; it does not remove the fact that a
k-bounded rewind can cross an epoch boundary and force a full recomputation that is
guaranteed to produce the same rejected verdict.

## Environment

- Network: Preview, `proto_major` 11, `storageMode: core`, SQLite metadata,
  schema version 9.
- Both nodes Mithril-bootstrapped (not from genesis), then live-followed.
- Upstream base: `7f04e017` (#3745). Node A additionally carries gouroboros
  v0.202.5; node B v0.202.4. Both carry local operator patches unrelated to
  reward accounting.
- **Not present** in either build (all merged upstream after this base):
  `6d6d681d` (#3782, the O(n²) account-lookup fix for #3650),
  `8747a130` (#3793, Mithril-imported opcert counters counted as minted blocks),
  `b5580a70` (floor reward expected blocks),
  `05ea181a` (#3923, preserve imported reward seed failures),
  `f86d64f4` (make a recovery rewind reachable and fail terminally).

I have not re-tested on current `main`; the reward divergence is persisted in the
database, so a code fix alone would not be expected to repair state already seeded
across the import boundary.

## What would make this authoritative (currently unverified)

- The per-pool `reward_pool_input` rows (`blocks_produced`, `total_blocks_in_epoch`)
  for epochs 1409/1410 on both nodes, diffed against each other and against the
  canonical per-pool block counts. That would say whether the pot delta comes from
  block counts (#3793 territory) or from the stake snapshot.
- Whether the epoch-1411 `total_stake` delta of 52,947,758,320 is attributable to a
  bounded set of accounts, and whether those accounts' UTxOs or delegations sit at
  or below the respective Mithril anchors.
- A third node bootstrapped at a *third* anchor, to confirm the pot moves with the
  anchor rather than with anything else.

## Acceptance criteria

- Two nodes bootstrapped from Mithril at different anchors on the same network
  compute identical `total_reward_pot`, `available_rewards`, and per-account reward
  balances at a shared epoch boundary.
- A reward-state divergence is not answered by an unbounded rewind/replay loop whose
  outcome is deterministic by construction.

## Related

- #3628 (closed) — same direction, Preprod genesis; #3885 carries a from-genesis case forward.
- #3570 (closed) — over-count by 105 lovelace, root-caused to a pot-parameter epoch
  mismatch and fixed by #3482. Different mechanism: these nodes are Mithril-anchored
  and disagree *with each other*.
- #3767 / #3788 — Mithril-imported reward state; those describe the under-credit
  (`balance 0`) direction, this is the over-credit direction.
- #3650 / #3782 — the epoch-rollover stall that turns each retry into ~20 minutes.
- #3766, #3775 — recovery disposition for withdrawal mismatches.

wolf31o2:
Fixes #3975.

Snapshot UTxO import now preserves pointer-address positions through address parsing and database model conversion. Repeated unchanged-tip ledger pipeline failures now invoke the fatal error callback and stop after the configured threshold.

---
## Summary by cubic
Snapshot UTxO import now preserves pointer-address stake positions, and after `noProgressStuckThreshold` consecutive no-progress restarts the ledger pipeline marks itself halted and stops retrying instead of replaying a deterministically rejected tip forever.

**Bug Fixes**
- Pointer addresses are decoded into `models.UtxoPointer` during parsing and propagated by `UTxOToModel`; malformed pointer payloads keep the UTxO importable but omit pointer-derived stake.
- After `noProgressStuckThreshold` no-progress restarts, the pipeline marks itself halted and stops retrying; nodes that previously replayed the same rejected tip forever now require operator intervention.
- `ARCHITECTURE.md` documents the new terminal halt behavior.

<sup>Written for commit 4ff3bc38677811de17b057cc0b9ba7435215f17a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ledger/state.go
```diff
@@ -5787,6 +5785,35 @@ func ledgerPipelineRetryDelay(
...
+		ls.config.Logger.Error(
+			"ledger pipeline stopped after repeated no-progress restarts; operator intervention is required",
+			"component", "ledger",
+			"consecutive_no_progress", progress.consecutiveNoProgress,
+			"tip_slot", progress.lastTipSlot,
+			"error", err,
+		)
+	}
+	if ls.config.FatalErrorFunc != nil {
+		ls.config.FatalErrorFunc(fmt.Errorf(
```
arepala-uml: stopStuckLedgerPipeline calls FatalErrorFunc, which cancels the node's root context and exits the process (node_ledger_config.go -> Node.cancelForFatal); since docker-compose.yml sets restart: unless-stopped, a deterministic no-progress condition will now exit and restart the node repeatedly instead of halting in place the way the errHaltLedgerPipeline branch above does, and pipelineProgress resets on each restart so the threshold must reaccumulate every cycle.

ledger/state.go
```diff
@@ -5984,6 +6017,12 @@ func (ls *LedgerState) ledgerProcessBlocksWithAttempt(
 			progress.consecutiveNoProgress,
 			stuck,
 		)
+		if stuck {
```
arepala-uml, line 5975: Because this branch returns whenever stuck is true, the code below (if stuck && pipelineStuckShouldAnnounce) and the equivalent pattern for endorserStuck above are now unreachable, and the noProgressStuckReannounceInterval doc comment plus the existing TestPipelineStuckAnnouncementStaysVisible test still describe a periodic re-announcement that can no longer happen in the running pipeline.

ledgerstate/utxo.go
```diff
@@ -532,6 +532,20 @@ func extractAddressKeys(addr []byte, result *ParsedUTxO) {
 		// Pointer address: 1 header + 28 payment + pointer
 		result.PaymentKey = bytes.Clone(addr[1:29])
 		result.PaymentScript = paymentIsScript
+		// Decode the pointer with the canonical address parser so the
+		// variable-length integer rules stay in one place. Preserve the
+		// existing fail-soft behavior for malformed or unrepresentable
+		// pointer payloads: the spendable UTxO remains importable, but it
+		// contributes no pointer-derived stake.
+		if parsed, err := lcommon.NewAddressFromBytes(addr); err == nil {
```
arepala-uml, line 540: This only reaches parseMempackTxOut; parseCborTxOut (still dispatched by parseUTxOEntry for legacy CBOR-encoded TxOuts and feeding the same UTxOToModel/DB path) already holds a typed ledger.Address via txOut.Address() but never extracts StakingPayload().(lcommon.AddressPayloadPointer), so pointer-address stake is still dropped for snapshots using that encoding, unlike UtxoLedgerToModel which already does this for the live-sync path.

arepala-uml (changes requested):
Two ledger/state.go findings on the no-progress halt (process-exit crash-restart risk and now-unreachable reannouncement code) and one ledgerstate/utxo.go finding: the pointer-stake fix does not cover the legacy CBOR TxOut parse path, leaving #3975's stake loss unfixed for that encoding.

arepala-uml (approved):

Issue #2870, opened 2026-07-18
Credential hot-reload without restart (SIGHUP / Bark RPC)


wolf31o2:
#### Motivation (dingo-operator)

The operator rotates a pool's operational certificate by writing a new `kes.skey` + `opcert.cert`
into the mounted Kubernetes Secret. Today the only way to make a running block producer pick up the
new credentials is to replace the files and **restart the pod** (design §7 step 7). A restart drops
the current leader schedule, forces the node to re-establish peers and re-warm caches, and — for a
sole active block producer — creates a window in which a scheduled leader slot can be missed. An
in-place, validated credential reload would let the operator rotate a mounted Secret with no pod
restart and no missed leader slot, and is a prerequisite for the P3 ActiveStandby hot-reload rollout.

#### Proposal

Add an atomic, validated credential-swap path triggered by `SIGHUP` and/or a Bark RPC.

Current state (verified):

- Credentials are loaded exactly once at startup, in `validateBlockProducerStartupAtSlot`
  (`node_forging.go:62`): `forging.NewPoolCredentials()` (`node_forging.go:65`),
  `creds.LoadFromFiles(...)` (`node_forging.go:66-70`), `creds.ValidateOpCert()`
  (`node_forging.go:73`), `creds.ValidateKESPeriod(genesis, currentSlot)` (`node_forging.go:80`).
  The resulting `*forging.PoolCredentials` is passed into `initBlockForger` (`node_forging.go:217`)
  with no runtime path to replace it.
- Signal handling uses `signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)`
  (`internal/node/node.go:498-503`). `SIGHUP` is not handled anywhere in the repo.
- `PoolCredentials` (`ledger/forging/keys.go:38`) already guards its fields with a `sync.RWMutex`
  (`keys.go:53`), but the only runtime mutation is `UpdateKESPeriod` (`keys.go:160`); there is no
  whole-struct atomic swap / CAS to replace the loaded credentials.

Proposed change:

1. Add a `Reload` operation that re-reads the VRF/KES/opcert files into a fresh
   `forging.NewPoolCredentials()` and runs the existing validators **before** swapping:
   `ValidateOpCert()` (`keys.go:335`), `ValidateKESPeriod()` (`keys.go:408`), and
   `ValidateAgainstLedger()` (`keys.go:472`, which uses the `LedgerView` interface at `keys.go:446`).
[...]

PR #4247, opened 2026-09-13
feat(forging): KES agent client (serve-key and sign modes)

wolf31o2:
Adds `--shelley-kes-agent-socket`, sourcing the KES signing key from an
external [bursa](https://github.com/blinklabs-io/bursa) KES agent over a
Unix-domain socket instead of a local `--shelley-kes-key` file. The VRF key
and operational certificate are still read from local files: the agent
protocol carries only the KES signing key, and in serve-key mode the
operational certificate alongside it.

Two modes, selected by `--shelley-kes-agent-mode` (default `serve-key` once a
socket is set):

- **serve-key** — the agent pushes the evolving KES signing key, its
  verification key, and the operational certificate; the node signs headers
  locally exactly as it would with a local key file.
  `PoolCredentials.LoadFromAgentServeKey` installs through the same
  identity/generation path as `LoadFromFiles`. A background loop installs
  every later push. Installing a push clears the validated KES protocol
  lifetime, so no installed key inherits a policy that was never checked
  against the material now installed; the loop therefore re-runs
  `ValidateOpCert` and `ValidateKESPeriod` after each install, without which
  `credentialGeneration.kesSign` refuses every signature after the first
  rotation with "operational certificate is not validated".
- **sign** — the node sends header bytes and receives signatures; the KES
  signing key never enters the node process.
  `PoolCredentials.LoadFromAgentSign` stores a `RemoteKESSigner` and leaves
  the local KES signing key unset.

`credentialGeneration.kesSign` validates the operational-certificate lifetime
itself before either branch, deliberately redundant with
`BlockForger.SignBlockHeader` and `DefaultBlockBuilder.buildBlock`, so a
future call site cannot reintroduce the bypass by omitting it. The refcounted
`credentialGeneration` model is built on rather than replaced; `forger.go` and
`builder.go` are unchanged.

`kesagent.Client` implements bursa's wire protocol
(`docs/signer/kes-agent-protocol.md`): a 4-byte big-endian length prefix and a
JSON payload. Every value read off the socket is treated as untrusted input:

- the handshake is bounded in size (`MaxHelloFrameLen`) and time
  (`HelloTimeout`);
- a sign round trip is bounded by `SignTimeout`, validated to stay under one
  slot because a sign call blocks the slot-aligned forging loop;
- a pushed key is size-checked, matched against its own operational
  certificate, and self-sign probed before installation;
- a sign response is verified with `kes.VerifySignedKES` against the KES
  verification key the local operational certificate commits to.

`Client.Close` never holds its mutex across a blocking network read, so
graceful shutdown cannot deadlock behind a serve-key connection parked in its
idle read.

Configuration: `shelleyKesKey` is no longer required when an agent socket is
set, and setting both is rejected rather than silently preferring one.

Bumps `github.com/blinklabs-io/bursa` to v0.17.0 for `bursa.DecodeOpCert` and
the protocol's reference implementation.

Supersedes #3115. Related to #2870: serve-key mode reloads KES material in
place from the agent, but does not add the SIGHUP or RPC reload path for local
key files that issue asks for.


kesagent/client.go
```diff
@@ -0,0 +1,617 @@
...
+	// one now that it succeeded.
+	if err := conn.SetDeadline(time.Time{}); err != nil {
+		_ = conn.Close()
+		c.recordDialFailureLocked()
+		return fmt.Errorf("kesagent: clear hello deadline: %w", err)
+	}
+
+	c.conn = conn
+	c.backoff = 0
+	c.nextDialAfter = time.Time{}
```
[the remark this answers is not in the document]
wolf31o2: Both halves confirmed and fixed in 5aa09b52.

`connectLocked` no longer clears the backoff on a completed handshake. Reaching `Hello` proves only that something is listening, and every failure the throttle exists for -- an agent serving key material the node cannot install, or one that refuses or corrupts every sign -- reconnects successfully first, so the reset held the interval flat at `minReconnectBackoff` for as long as the agent kept failing. `Run`'s doc comment already claimed the reset belonged to a completed install; `connectLocked` contradicted it.

Sign mode now calls `recordDialFailureLocked` on every path that tears the connection down, and clears the backoff only after `kes.VerifySignedKES` passes. `resetBackoffLocked` exists because `Sign` holds `c.mu` for the whole exchange. The `resp.Error != ""` branch is deliberately not throttled: a typed refusal leaves the connection usable and is not a transport fault.

`TestClient_ServeKeyBackoffSurvivesAReconnect` drives an agent that completes `Hello` and pushes valid material the install rejects, and asserts the backoff after the second failure. With the reset restored it reports `"250ms" is not greater than "250ms"`. `TestClient_SignFailuresAreThrottled` asserts the same growth across two failed signs and a return to zero after a verified one; without the recorded failures it reports the same flat `250ms`.

kesagent/client.go
```diff
@@ -0,0 +1,617 @@
...
+
+	if deadline, ok := ctx.Deadline(); ok {
+		if err := conn.SetReadDeadline(deadline); err != nil {
+			c.invalidateConn(conn)
+			return PushedKey{}, fmt.Errorf(
+				"kesagent: set read deadline: %w",
+				err,
+			)
+		}
+		defer func() { _ = conn.SetReadDeadline(time.Time{}) }()
```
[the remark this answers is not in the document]
wolf31o2: Already addressed at the current head. `AwaitPushedKey` launches a watcher before the read (`kesagent/client.go`, the goroutine selecting on `ctx.Done()` and `readDone`) that calls `invalidateConn(conn)` on cancellation, which closes the socket and unblocks `readFrame`. `readDone` is closed by a deferred `close` on every return, so the watcher exits with the call. The socket deadline from `ctx.Deadline()` is still applied, but it is no longer the only thing that can end a cancel-only wait, and a direct caller does not need `Close` to interrupt the read.

`TestClient_CloseInterruptsBlockedAwaitPushedKey` covers the sibling `Close` path.

kesagent/client.go
```diff
@@ -0,0 +1,617 @@
...
+			case <-time.After(c.currentBackoff()):
+			}
+			continue
+		}
+		if err := install(pk); err != nil {
+			c.logger.Error(
+				"failed to install agent-pushed KES key",
+				"error", err,
+			)
+			continue
```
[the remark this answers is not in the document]
wolf31o2: Already addressed at the current head. `Run` calls `invalidateCurrentConn()` after a failed install and then continues the loop, so the next iteration redials and the agent re-sends its current `KeyPush` after `Hello` -- which is the reconnect this asks for, and it makes an opcert whose KES period has not started yet retryable without waiting for the next evolution. `resetBackoff` is reached only on a successful install, so a run of failures is throttled rather than hot-looping.

ledger/forging/keys.go
```diff
@@ -340,14 +334,70 @@ func (pc *PoolCredentials) LoadFromFiles(
...
+// afterward; see credentialGeneration.kesSign and updateKESPeriod for how the
+// same gate that applies to a local key -- the opcert-lifetime check before
+// every sign, and monotonic period progression -- applies to it identically.
+func (pc *PoolCredentials) LoadFromAgentSign(
+	vrfSKeyPath string,
+	opCertPath string,
+	signer RemoteKESSigner,
+) error {
+	if signer == nil {
+		return errors.New("kes agent sign mode requires a non-nil signer")
```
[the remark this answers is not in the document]
wolf31o2: Already addressed at the current head. `LoadFromAgentSign` routes a nil signer through `installLoaded(nil, err, nil)`, which bumps the generation and calls `clearUnsafe()`, so an already-loaded `PoolCredentials` is invalidated rather than left reporting `IsLoaded` with the previous signing material. That is the same fail-closed path every other load error takes.

node_forging.go
```diff
@@ -66,7 +68,17 @@ func (n *Node) validateBlockProducerStartupAtSlot(
...
+	if n.config.shelleyKESAgentSocket != "" {
+		if err := n.loadBlockProducerCredentialsFromAgent(
+			creds,
+			currentSlot,
+		); err != nil {
+			return nil, fmt.Errorf(
+				"load pool credentials from KES agent: %w",
+				err,
+			)
+		}
```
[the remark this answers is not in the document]
wolf31o2: Partly addressed, and the rest is fixed in 5aa09b52.

The first half was already in place: `validateBlockProducerStartupAtSlot` defers `closeKESAgentClient()` on its own error, so a rejection after the agent is dialled no longer returns with the client connected. That defer had no test; `TestValidateBlockProducerStartup_KESAgentClosedWhenValidationFails` now covers it by validating against a slot past the certificate's protocol lifetime, so the push installs and the KES-period check is what fails. Without the defer it reports a live `*kesagent.Client` where nil is required.

The second half was still open. In `node.go` the cleanup closure was appended after `validateBlockProducerLedger`, `initBlockForger` and `enableLeiosVoting`, each of which can return; it is now registered immediately after validation, ahead of all three. Every stop in it is nil-guarded, so registering it before the forger exists is safe. `reinitializeBlockProducer` had no registration at all and, unlike `Run`'s failure path, leaves the node running: it now closes the client on its own error paths.

Those two are verified by inspection rather than by a test. Driving `Run` or `reinitializeBlockProducer` to a mid-startup failure with a live agent needs node-level ledger state that the block-producer tests do not construct.

node_forging.go
```diff
@@ -107,6 +119,244 @@ func (n *Node) validateBlockProducerStartupAtSlot(
...
+	// a KES evolution, an opcert rotation, or a re-push after a reconnect --
+	// is not, so the loop must re-establish it itself. Without this,
+	// credentialGeneration.kesSign refuses every signature after the first
+	// mid-run push with "operational certificate is not validated" and the
+	// node stops forging until it is restarted.
+	loopInstall := func(pk kesagent.PushedKey) error {
+		if err := install(pk); err != nil {
+			return err
+		}
+		return n.revalidateAgentServedCredentials(creds, startupSlot)
```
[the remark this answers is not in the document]
wolf31o2: Confirmed and fixed in 5aa09b52.

`PoolCredentials.LoadFromAgentServeKeyValidated` performs the install and the re-validation under one write lock: `installLoaded` and `ValidateKESPeriod` are split into `installLoadedUnsafe` and `validateKESPeriodUnsafe`, and the new method takes `pc.mu` once around both. The key material is still read from disk outside the lock, since reading and deriving it has nothing to do with what is published, and holding the write lock across it would block every forge attempt for the duration. `startKESAgentServeKey`'s loop callback is now that single call, and `revalidateAgentServedCredentials` is gone.

`TestLoadFromAgentServeKeyValidatedNeverPublishesAnUnvalidatedCredential` runs a reader that acquires a generation and asks for the validated lifetime exactly as `credentialGeneration.kesSign` does, concurrently with 200 installs, and fails if any loaded generation carries no lifetime. With the two-call form restored it fails on all three runs of `-race -count=3` with the reported error: `operational certificate is not validated`.

`ARCHITECTURE.md`'s serve-key section records the single-operation requirement and why the window matters.

node_forging.go
```diff
@@ -64,9 +66,30 @@ func (n *Node) validateBlockProducerStartup() (*forging.PoolCredentials, error)
 
 func (n *Node) validateBlockProducerStartupAtSlot(
```
chrisguiney, line 85: Merge blocker: origin/main conflicts here and in go.mod. main replaced this body with validateBlockProducerCredentialMaterial plus a deferred path, validateBlockProducerStartupForClock, taken when confirmed era history does not span the wall clock. Keeping main's split leaves that path on LoadFromFiles with shelleyKESKey empty, which config now forbids beside the socket, so agent-backed startup fails there. Re-site the agent branch and the close-on-failure defer; leaving that defer in the helper drops the guarantee KESAgentClosedWhenValidationFails covers.

kesagent/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+type Metrics struct {
+	connected         prometheus.Gauge
+	reconnectFailures prometheus.Counter
+	signSuccesses     prometheus.Counter
+	signFailures      prometheus.Counter
+	signLatency       prometheus.Histogram
+}
+
+// NewMetrics registers and returns KES agent client metrics on reg.
+func NewMetrics(reg prometheus.Registerer) *Metrics {
```
chrisguiney, line 38: NewMetrics has no caller anywhere in the module: both kesagent.Config literals in node_forging.go, at lines 171 and 289, leave Metrics nil, so dingo_kes_agent_connected, _reconnect_failures_total, _sign_success_total, _sign_failure_total and _sign_latency_seconds are never registered. ARCHITECTURE.md line 5941 tells operators a stalled frame increments dingo_kes_agent_reconnect_failures_total. n.config.promRegistry is in scope at both sites. An agent stall or a lost sign is otherwise log-only on a producer. Wire it, or drop the file and the doc claim.

node_forging.go
```diff
@@ -107,6 +130,236 @@ func (n *Node) validateBlockProducerStartupAtSlot(
...
+	client, err := kesagent.NewClient(kesagent.Config{
+		SocketPath: n.config.shelleyKESAgentSocket,
+		Mode:       kesagent.ModeServeKey,
+		Logger:     n.config.logger,
+	})
+	if err != nil {
+		return err
+	}
+
+	install := func(pk kesagent.PushedKey) error {
```
chrisguiney, line 289: serve-key never reads n.config.shelleyOperationalCertificate: install passes only shelleyVRFKey and the pushed cert, so poolID is Blake2b224 over the pushed opCert.ColdVKey and the agent alone decides which pool this node forges for. installLoadedUnsafe pins identity from the first load, so the first push is unpinned, and the ledger cross-check downgrades an unregistered pool to a warning. flags.go says the opcert flag still applies and ARCHITECTURE.md line 5851 says it is always read from a local file; neither holds. Pin the pushed ColdVKey to the local file before the first install.

kesagent/client.go
```diff
@@ -0,0 +1,757 @@
...
+//
+// A completed handshake deliberately does not clear the reconnect backoff.
+// Reaching Hello proves only that something is listening on the socket, and
+// the failures the backoff exists to throttle -- an agent that pushes key
+// material the node cannot install, or one that refuses or corrupts every
+// sign -- all reconnect successfully first. Clearing here restarted every one
+// of them at minReconnectBackoff, however long they had been failing. The
+// reset belongs to the operation that actually succeeded: Run after an
+// install, Sign after a verified signature.
+func (c *Client) connectLocked(ctx context.Context) error {
```
chrisguiney, line 232: With origin/main merged and the agent branch re-sited, the build, race tests for kesagent, ledger/forging, internal/config and the root KES tests, golangci-lint 2.13.2 on the touched packages, docs-parity, import-boundaries and the conformance suite all pass, so the conflict is the only thing between this and a green merged tree. Reverting the connectLocked backoff reset and the kesSign validateKESPeriod gate in place fails their tests, so both new guards are causal. The six CodeRabbit findings from 572eb4ca all hold at head; cubic has reviewed no head here.

chrisguiney (changes requested):
Blocked on a merge conflict with main that is semantic rather than textual, plus two defects on the block-producer path: the kesagent Prometheus collectors are never registered, and serve-key mode never reads the configured operational certificate. Checks at head are green but started 2026-09-15T11:17Z, before main last moved, and the branch is 45 commits behind, so they cover an older merge. Supersedes #3115 as the body states, and the #2870 reference is accurate.

node_forging.go
```diff
@@ -76,6 +78,44 @@ func (n *Node) validateBlockProducerStartup() (*forging.PoolCredentials, error)
...
+		n.config.shelleyVRFKey,
+		n.config.shelleyKESKey,
+		n.config.shelleyOperationalCertificate,
+	); err != nil {
+		return nil, fmt.Errorf("load pool credentials: %w", err)
+	}
+	if err := creds.ValidateOpCert(); err != nil {
+		return nil, fmt.Errorf("validate operational certificate: %w", err)
+	}
+	return creds, nil
```
chrisguiney, line 131: Merge blocker: this return replaced the KES-period tail both parents carry. origin/main and 6405840ab each call ValidateKESPeriod(genesis, currentSlot) here, then CurrentKESPeriodFromGenesis and the validated log line; the merged body stops after ValidateOpCert. ValidateKESPeriod now has no production caller, so startup no longer rejects an expired or not-yet-valid opcert and OpCertExpiryPeriod stays 0, making forger.go:1351 log forge skip: KES protocol lifetime is not validated at every leader slot. Restoring the tail turns all four failing tests green.

node_forging.go
```diff
@@ -76,6 +78,44 @@ func (n *Node) validateBlockProducerStartup() (*forging.PoolCredentials, error)
...
+		// Agent startup installs the client before the remaining credential
+		// validation below. Do not leave that client and its serve-key loop
+		// running when a later validation step rejects the credentials.
+		defer func() {
+			if retErr != nil {
+				n.closeKESAgentClient()
+			}
+		}()
+	}
+	if n.config.shelleyKESAgentSocket != "" {
```
chrisguiney: Merge blocker: this agent branch exists only in validateBlockProducerStartupAtSlot. validateBlockProducerStartupForClock line 141 takes the file-only loader whenever confirmed era history does not span the wall clock, which is the syncing-from-genesis case main added that branch for. internal/config.ValidateKESKeySources forbids shelleyKesKey beside the socket, so the loader receives an empty path and startup fails with failed to load KES signing key: failed to open key file. Both node.go:1674 and node_lifecycle.go:1265 reach it, and it is independent of the line 116 break.

kesagent/metrics.go
chrisguiney, line 38: NewMetrics still has no caller in the module: both kesagent.Config literals, node_forging.go:232 and :350, leave Metrics nil, so dingo_kes_agent_connected, _reconnect_failures_total, _sign_success_total, _sign_failure_total and _sign_latency_seconds are never registered. ARCHITECTURE.md line 5941 tells operators a stalled frame increments dingo_kes_agent_reconnect_failures_total. n.config.promRegistry is in scope at both sites. An agent stall or a lost sign is otherwise log-only on a producer. Wire it, or drop the file and the doc claim.

node_forging.go
```diff
@@ -151,42 +191,234 @@ func (n *Node) validateBlockProducerCredentialMaterial() (*forging.PoolCredentia
...
+		SocketPath: n.config.shelleyKESAgentSocket,
+		Mode:       kesagent.ModeServeKey,
+		Logger:     n.config.logger,
+	})
 	if err != nil {
-		return nil, err
+		return err
+	}
+
+	install := func(pk kesagent.PushedKey) error {
```
chrisguiney, line 289: serve-key still never reads n.config.shelleyOperationalCertificate: install and loopInstall pass only shelleyVRFKey plus the pushed material, so poolID is Blake2b224 over the pushed opCert.ColdVKey and the agent alone decides which pool this node forges for. installLoadedUnsafe pins identity from the first load, so the first push is unpinned. ARCHITECTURE.md line 5852, the doc comment at line 195 and the pull-request body all state the operational certificate is always read from a local file. Pin the pushed ColdVKey to the local file before the first install.

chrisguiney (changes requested):
Blocked on a merge-induced break: e2dabebb merges main with no textual conflict and yields a block-producer startup path neither parent has. validateBlockProducerStartupAtSlot lost its KES-period validation tail, so ValidateKESPeriod has no production caller and the forger skips every leader slot. Four root-package tests fail at head and pass on both origin/main and 6405840ab. The agent-mode deferred-clock path, the unregistered kesagent metrics and the serve-key opcert finding are also open. The six CodeRabbit findings from 572eb4ca remain fixed; cubic has reviewed no head.

node_forging.go
wolf31o2: Addressed at 58d140ee. The agent branch now lives in validateBlockProducerCredentialMaterial (node_forging.go:205), which both validateBlockProducerStartupAtSlot and validateBlockProducerStartupForClock call, so the deferred path reaches the agent instead of LoadFromFiles with an empty shelleyKESKey. The close-on-failure defer is sited at both entry points, node_forging.go:88-97 and 156-162, so KESAgentClosedWhenValidationFails covers the deferred path too.

node_forging.go
wolf31o2: Addressed at 58d140ee. validateBlockProducerStartupForClock now takes the agent branch on the deferred path: its loader is validateBlockProducerCredentialMaterial (node_forging.go:205), shared with validateBlockProducerStartupAtSlot, and it carries its own close-on-failure defer at node_forging.go:156-162. TestValidateBlockProducerStartupForClock_KESAgentDeferredPath drives that path with a fake agent and asserts startup succeeds with shelleyKesKey empty beside the socket.

node_forging.go
wolf31o2: Addressed at 58d140ee. validateBlockProducerStartupAtSlot keeps the KES-period tail: ValidateKESPeriod at node_forging.go:107, CurrentKESPeriodFromGenesis at 110, and the validated log line carrying opcert_expiry_period at 119-131. validateBlockProducerCredentialMaterial ends at ValidateOpCert deliberately, because the deferred branch cannot place an opcert in time; that branch arms the lifetime instead, with ArmKesProtocolLifetime at node_forging.go:171, so OpCertExpiryPeriod is non-zero on both paths and forger.go's per-slot gate has its data.

kesagent/metrics.go
wolf31o2: Wired at 58d140ee. Both kesagent.Config literals take Metrics from kesAgentClientMetrics, at node_forging.go:283 (serve-key) and 421 (sign); kesAgentClientMetrics registers on n.config.promRegistry once, at node_forging.go:438, and returns nil only when no registry is configured. TestKESAgentStartupWiresClientMetrics gathers the registry after startup and asserts dingo_kes_agent_connected is present, so the ARCHITECTURE.md claim about dingo_kes_agent_reconnect_failures_total now holds.

wolf31o2: Wired at 58d140ee. Both kesagent.Config literals take Metrics from kesAgentClientMetrics, at node_forging.go:283 and 421; kesAgentClientMetrics registers the collectors on n.config.promRegistry once, at node_forging.go:438. TestKESAgentStartupWiresClientMetrics gathers the registry after startup and asserts the collectors are present.

node_forging.go
wolf31o2: Pinned at 58d140ee. startKESAgentServeKey loads the local operational certificate first, at node_forging.go:270, and both install and loopInstall refuse a push whose OpCert.ColdVKey differs from it, at node_forging.go:289-297 and 319-327. The first push is therefore checked before installLoadedUnsafe fixes identity, so the agent cannot choose which pool this node forges for. TestValidateBlockProducerStartup_KESAgentPinsLocalColdKey pushes a mismatched cold key and asserts startup fails.

node_forging.go
wolf31o2: Pinned at 58d140ee. node_forging.go:270 loads the local operational certificate, and the cold-key comparison runs in install and in loopInstall, at node_forging.go:289-297 and 319-327, so every push including the first is matched against the local file. flags.go and the ARCHITECTURE.md statement that the operational certificate is always read locally both hold again. Covered by TestValidateBlockProducerStartup_KESAgentPinsLocalColdKey.

kesagent/client.go
wolf31o2: Still holds at 58d140ee, now on a tree merged with current origin/main. Removing the kesSign validateKESPeriod call in place fails both TestCredentialGenerationKesSignRejectsExpiredPeriod and TestCredentialGenerationKesSignAgentPathRejectsExpiredPeriod with "Expected error with 'operational certificate expired' in chain but got nil"; restoring it returns both to pass, so the gate is causal on the local and the agent path alike. The merge took origin/main's VerifyOpCertSignature rewrite of validateOpCertUnsafe and restored the crypto/ed25519 import that rewrite removed, which loadPoolCredentialsFromAgent still needs for its KES vkey size check.

wolf31o2: Head is now 1975ce15. The go-test (Windows) run at 58d140ee failed only in the seven KES agent startup cases: bursa v0.17.1 applies its key-file policy inside LoadKeyFromFile whenever the envelope carries a signing key, and on Windows it rejects Built-in Administrators as a trustee, which the runner's temporary file carries however the DACL is rewritten. Those cases now read the devnet envelope with os.ReadFile and parse it with bursa.LoadKeyFromBytes, which is the key material they actually need. PoolCredentials.LoadFromFiles is untouched and still enforces the policy through keystore.CheckOpenFilePermissions. Every node_forging.go line cited in the replies above is unchanged at this head.

chrisguiney (approved):

PR #4269, opened 2026-09-14
fix(leiosnotify): bind cursor cleanup to connection ownership

wolf31o2:
Observe connection shutdown in idle LeiosNotify callbacks, including requests started before connection-manager registration.

Track cursor and reservation ownership by the exact server instance. Stale startup, completion, and close callbacks cannot overwrite or remove a replacement connection's cursor. Replace the notification close-waiter workaround with the connection lifecycle signal.

Add connection-collision and started-protocol delivery regressions. Preserve the existing connection-manager callback. Capture both peers' constructor and connection errors when startup fails. Close fixture-owned error channels after connection shutdown so connection-manager watchers can finish cleanup. The protocol-ID-8 startup failure tracked in #4287 remains unresolved.

Related to #3412 and blinklabs-io/gouroboros#2306; this addresses LeiosNotify cancellation and cursor ownership, not the remaining compound lifecycle scope.

## Summary by Cubic

Node-to-client certificate serving retains its manager-driven close signal.


ouroboros/leios_forge_serve_test.go
```diff
@@ -130,9 +130,9 @@ func TestLeiosForgedEBLogSkipsOriginAfterPriorDelivery(t *testing.T) {
 		otherEntry, _ := o.leiosEBLog.next(otherConn)
 		require.NotNil(t, otherEntry)
 		require.Equal(t, want, *otherEntry.vote)
-		o.leiosEBLog.complete(otherConn, true)
+	o.leiosEBLog.complete(otherConn, nil, true)
```
chrisguiney: This statement lost one level of indentation inside the loop body. gofmt -l flags this file at the head and reports it clean at the merge base, so the change introduces the drift. The lint job cannot catch it because .golangci.yml sets run.tests: false, and make format runs gofmt -s -w over every .go file including tests, so the next person to run that target gets an unrelated diff. Re-indent to two tabs.

ouroboros/leiosnotify.go
```diff
@@ -301,27 +318,50 @@ func (l *leiosForgedEBLog) nextRetryLocked(
...
+func (l *leiosForgedEBLog) registerConn(
+	connKey string,
+	owner *oleiosnotify.Server,
+	isCurrent func() bool,
+) {
 	l.mu.Lock()
+	defer l.mu.Unlock()
+	// Validate manager ownership under the cursor lock: an old startup may
+	// have looked up its connection before a replacement was registered.
+	if isCurrent != nil && !isCurrent() {
```
chrisguiney, line 361: isCurrent is invoked while l.mu is held and reaches into the connection manager lock, so registerConn is the one place that establishes leiosEBLog.mu before connmanager.connectionsMutex. No inverse path exists at this head - both eviction sites and the ErrorChan watcher call ConnClosedOwnerFunc with connectionsMutex released - but l.mu serialises every notify serving callback and every append, so record that ordering here before a future callback under connectionsMutex inverts it.

ouroboros/leiosnotify.go
```diff
@@ -301,27 +318,50 @@ func (l *leiosForgedEBLog) nextRetryLocked(
...
 	l.pruneLocked()
-	l.mu.Unlock()
 }
 
 // registerConn pre-registers connKey at the current tail, or at the oldest
 // failed delivery, so entries appended between connection open and the peer's
 // first RequestNext are not pruned before the cursor is established. It is a
-// no-op when connKey is already registered.
-func (l *leiosForgedEBLog) registerConn(connKey string) {
+// no-op for the same owner. A replacement releases the old owner's reservation
```
chrisguiney: The sentence no-op for the same owner no longer holds: the body now always rotates and closes wakeCh before returning, so every registerConn wakes every parked notify serving callback for every connection, same owner or not. That wake is needed after a replacement evicts an old owner, but the sentence reads as though a repeat registration does nothing at all. State what the wake is for.

node.go
```diff
@@ -1170,7 +1170,7 @@ func (n *Node) Run(ctx context.Context) (runErr error) {
 			PromRegistry:        n.config.promRegistry,
 			MaxConnectionsPerIP: n.config.maxConnectionsPerIP,
 			MaxInboundConns:     n.config.maxInboundConns,
-			ConnClosedFunc:      n.handleConnManagerClosed,
+			ConnClosedOwnerFunc: n.handleConnManagerClosedOwner,
```
chrisguiney, line 1162: Class audit of per-connection protocol handle dereferences on the paths this change touches: handleConnManagerClosedOwner guards conn and conn.LeiosNotify() before reading .Server; leiosnotifyClientStart guards both; leiosnotifyServerRequestNext returns early on a nil ctx.Server; notifyEvictedConnectionClosedOwner can pass a nil conn into that nil-guarded handler. gouroboros populates ResponseSentFunc only from the server callback context, so complete never runs owner-blind here. The class is complete.

ouroboros/leios_merged.go
```diff
@@ -257,35 +198,21 @@ func (o *Ouroboros) releaseLeiosServeWaiter(
...
 	for _, ch := range waiters {
 		close(ch)
 	}
-	// Cancel before removing the cursor: a racing notification request checks
-	// its cancellation channel under the same log lock before registering.
-	// This also covers a request returning an offer during disconnection.
-	if o.leiosEBLog != nil {
-		o.leiosEBLog.removeConn(leiosConnectionIdString(connId))
-	}
+	// Wake the NtC closure waiters. LeiosNotify cursor cleanup is owner-specific
```
chrisguiney: This trailing comment sits after the waiters have already been woken and describes work the function no longer does, which the comment policy asks to delete rather than reword. Review evidence: the tree merged with current main builds and passes ouroboros, connmanager and root under -race -count=2; the merge audit reports all twelve owned files kept, zero reverted merges, zero branch-internal merges; the ARCHITECTURE.md contributes-nothing flag is a false positive, the new prose is present at head.

chrisguiney (dismissed):
Verified on the current head: the merged tree builds, go vet is clean, golangci-lint is clean on ouroboros and connmanager, and the ouroboros, connmanager and root packages pass under -race. The five ownership guards are causal - reverting each one in place fails its own test. Findings below: one formatting regression the pipeline structurally cannot see, one lock-ordering note, one stale doc sentence. The single CodeRabbit finding, posted against 0bfa9b3d, is already addressed at this head. No bot has reviewed 6e748b08.

arepala-uml (approved):

PR #4271, opened 2026-09-14
perf(database): cache prepared statements for reward-live-stake touch queries

chrisguiney:
**Depends on #4250, #4252, and #4270 -- not mergeable until all three land first.** This branch stacks on #4270's branch (which already includes its own dependencies); the only new content is the final commit, `perf(database): cache prepared statements for reward-live-stake touch queries`.

`refreshRewardLiveStakeAggregate` runs on every UTxO a stake credential gains or loses. `sumCredentialUtxoStakeQuery`, the aggregate SUM it calls, already routes through the hot-statement cache (#4270). Its two sibling queries in the same function -- an account lookup and the `reward_live_stake` upsert -- were still one-shot `QueryRowContext`/`ExecContext` calls, each re-parsing via `modernc.org/sqlite`'s statement-preparation path on every single UTxO touch. Live profiling of a synced node showed `refreshRewardLiveStakeAggregate` at 36.89% cumulative CPU, of which the already-cached aggregate sum accounted for 32.93% (a known, separately-tracked O(n)-recompute cost, not addressed here) -- the remaining ~4 points was these two sibling queries.

## Change

Added `rewardLiveStakeAccountQuery` and `rewardLiveStakeUpsertQuery` to `hotStatements`, generalizing the cache-or-fallback pattern from #4270 into two new helpers, `queryRowCached`/`execCached`.

**Correctness argument:** neither query has a `RETURNING` clause, so `dialectQueryer`'s MySQL-specific `QueryRowContext` branch (which bypasses ordinary translated execution) never applies -- a cached `*sql.Stmt` (dialect-translated once, at prepare time) reproduces exactly what the one-shot call did, for every dialect.

## Tests

`TestRefreshRewardLiveStakeAggregateMatchesOneShot` proves cached vs. one-shot parity across unregistered/registered/UTxO-touching cases. Additional regression tests prove both statements are actually cached and reused across independent write transactions -- these fail against the pre-fix code, since neither query was in `hotStatements` yet.

## Benchmark

Isolating the two newly-cached queries (negligible/constant UTxO-sum cost):
```
BenchmarkRefreshRewardLiveStakeAggregateAccountAndUpsert/one_shot-16        75282 ns/op
BenchmarkRefreshRewardLiveStakeAggregateAccountAndUpsert/prepared_cache-16  37017 ns/op
```
~2.0x faster per touch, consistent with the ~2.4x previously measured for `sumCredentialUtxoStakeQuery` itself (#4270).

---
## Summary by cubic
Caches the account lookup and `reward_live_stake` upsert in `refreshRewardLiveStakeAggregate`, so UTxO touches reuse prepared statements instead of recompiling them. It also reuses transaction-scoped statements within each write transaction, preventing statement retention from growing with batch size; the isolated touch path is ~2x faster.

**Details**
- Adds `queryRowCached` and `execCached` with one-shot fallbacks when the cache is unavailable.
- Evicts transaction-scoped statements on commit and rollback.
- Tests cover one-shot parity, cross-transaction reuse, and bounded retention for 2,000 references.
- Depends on #4250, #4252, and #4270 landing first.

<sup>Written for commit 2f06c974979e058c4d56d535237a9b5cc867d15b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4271?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Performance**
  - Improved database processing for reward live stake updates through statement reuse and more efficient aggregate calculations.
  - Enhanced handling of large-value MySQL casts to preserve 64-bit amounts.

- **Reliability**
  - Improved SQLite behavior and consistency across database operations.
  - Added safer handling for empty and invalid aggregate results.

- **Testing**
  - Expanded coverage for cached statements, concurrent access, reward stake calculations, and MySQL query translation.


wolf31o2 (dismissed):

database/plugin/metadata/sqlstore/prepared_stmt.go
```diff
@@ -178,3 +180,49 @@ func stmtForQueryer(
...
+		// stmtForQueryer returns either the shared, Store-lifetime cached
+		// statement itself (must not be closed here, see
+		// prepareHotStatements) or a *sql.Tx-scoped statement from
+		// (*sql.Tx).StmtContext, which database/sql documents as being
+		// closed automatically when the transaction commits or rolls
+		// back. Either way this call site owns no resource of its own to
+		// close, and closing eagerly would be wrong besides: the *sql.Row
+		// returned below defers running Scan against it until the caller
+		// invokes Scan.
+		return stmtForQueryer(ctx, db, cached).QueryRowContext(ctx, args...) //nolint:sqlclosecheck
```
wolf31o2: stmtForQueryer calls Tx.StmtContext on every invocation, and database/sql appends each returned *sql.Stmt to tx.stmts with a db.addDep entry released only at commit or rollback. The Tx-scoped statement is a distinct object from the cached one, so the cache is consulted per row while the derived wrapper is never reused: 2000 refs in one transaction retained 4000 statements against 0 with the cache empty. Deriving the Tx-scoped statement once per transaction keeps the parse saving without the retention.

database/plugin/metadata/sqlstore/live_stake.go
```diff
@@ -39,12 +81,8 @@ func (s *Store) refreshRewardLiveStakeAggregate(
 	var pool []byte
 	var active sql.NullBool
 	var addedSlot sql.NullInt64
-	accountErr := db.QueryRowContext(ctx, `
-SELECT reward, pool, active, added_slot
-FROM account
-WHERE credential_tag = ? AND staking_key = ?`,
-		ref.Tag,
-		ref.Key,
+	accountErr := s.queryRowCached(
```
wolf31o2, line 84: refreshRewardLiveStakeRefs calls this function once per stake ref inside a single withWriteTransaction, so both cached queries here execute per row. DeleteUtxos, importUtxos and the genesis paths pass whole ref slices, which makes retained statement count proportional to refs per transaction rather than constant.

database/plugin/metadata/sqlstore/reward_live_stake_account_cache_test.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+}
+
+// TestRefreshRewardLiveStakeAggregateReusesCachedStatementsAcrossTransactions
+// proves both new cached statements are actually reused across independent
+// write transactions -- the real access pattern (one write transaction per
+// block/UTxO touch) -- the way
+// TestSumCredentialUtxoStakeReusesCachedStatementAcrossTransactions already
+// proves for sumCredentialUtxoStakeQuery.
+func TestRefreshRewardLiveStakeAggregateReusesCachedStatementsAcrossTransactions(t *testing.T) {
```
wolf31o2, line 305: Every case here performs one refresh per write transaction, so none exercises the many-refs-in-one-transaction shape refreshRewardLiveStakeRefs uses. The reuse assertion compares store.stmts map entries, which stay stable regardless of how many Tx-scoped statements are derived from them, so it cannot observe per-row retention. A case looping refs inside one transaction and asserting bounded statement growth would cover it.

wolf31o2 (changes requested):
refreshRewardLiveStakeRefs loops refreshRewardLiveStakeAggregate over every stake ref inside one write transaction, and each cached call derives a new Tx-scoped statement that database/sql retains until commit. Measured in-tree: 2000 refs retained 4000 statements and 957 KiB; with only sumCredentialUtxoStakeQuery cached, 2000 and 452 KiB; with hotStatements empty, 0. Retention is linear in rows times cached statements consulted per row, so each hotStatements entry added here increases it.

database/plugin/metadata/sqlstore/prepared_stmt.go
chrisguiney: Fixed in 2f06c9749 (fix commit 00bedb10f). Added a per-transaction cache, Store.txStmts keyed by (*sql.Tx, cached *sql.Stmt), so stmtForQueryer derives the Tx-scoped statement via (*sql.Tx).StmtContext at most once per (transaction, query) pair and returns the same derived *sql.Stmt on every later call within that transaction. sqlTxn.releaseConnection (called from both Commit and Rollback) evicts the transaction's entry once it finishes, so the cache never outlives the transaction or grows across transactions.

Kept the cache keyed by the *sql.Tx pointer itself rather than a new wrapper type threaded through the queryer chain: requireAccountBaselineTransaction (account.go) type-asserts db through dialectQueryer down to exactly *sql.Tx to confirm a write is running inside a transaction, and substituting a different concrete type there would have silently broken that check.

Measured directly against database/sql's own Tx.stmts.v (the same list this comment cites) via reflection in the new TestRefreshRewardLiveStakeRefsBoundsTxScopedStatementRetention: 2000 refs in one transaction now retain at most 2 Tx-scoped statements. Reverting stmtForQueryer's *sql.Tx case back to a bare `v.StmtContext(ctx, cached)` call reproduces exactly 4000 retained statements against the same test, confirming the fix is what bounds it. This also addresses the review's top-level summary, which restates the same finding.

database/plugin/metadata/sqlstore/live_stake.go
chrisguiney: The fix lives in stmtForQueryer/txScopedStmt (prepared_stmt.go), the single point every queryRowCached/execCached call against a *sql.Tx goes through, so it covers every caller with the same shape, not just refreshRewardLiveStakeAggregate: DeleteUtxos, importUtxos, and the genesis import paths that pass whole ref slices into refreshRewardLiveStakeRefs inside one withWriteTransaction all get the same per-transaction reuse. Retention is now bounded by the number of distinct cached queries consulted per transaction, not by how many refs/rows a caller loops over within it.

database/plugin/metadata/sqlstore/reward_live_stake_account_cache_test.go
chrisguiney: Added TestRefreshRewardLiveStakeRefsBoundsTxScopedStatementRetention in prepared_stmt_test.go, which reproduces refreshRewardLiveStakeRefs's real shape (2000 unseeded stake refs processed inside a single write transaction) and asserts bounded retention by reading database/sql's own unexported Tx.stmts.v length via reflection (retainedTxStmtCount), the same quantity this comment and the top-level review measured. It fails at exactly 4000 with the fix reverted and passes at <=2 with it applied. Also added TestTxScopedStmtReusedWithinOneTransaction, TestTxScopedStmtDistinctAcrossTransactions, and TestEvictTxStmtsAfterCommitAndRollback to cover the cache mechanism itself (reuse within one transaction, no leakage across transactions, and eviction through the real Commit/Rollback path) directly, alongside the existing store.stmts identity assertions this comment correctly notes cannot observe per-row retention.

database/plugin/metadata/sqlstore/dialect_queryer.go
```diff
@@ -222,6 +222,13 @@ var (
 		`(?is)\s+RETURNING\s+id\s*;?\s*$`,
 	)
 	mysqlIntegerCastPattern = regexp.MustCompile(`(?i)\bAS\s+INTEGER\b`)
+	// mysqlBigintCastPattern covers "CAST(... AS BIGINT)", used instead of
+	// "AS INTEGER" where a value (such as a lovelace amount) can exceed
+	// PostgreSQL's 32-bit INTEGER. SQLite and PostgreSQL both accept BIGINT
+	// as a CAST target directly; MySQL's CAST() has no BIGINT spelling, so
+	// it is translated to SIGNED (64-bit), the same target "AS INTEGER" is
+	// translated to below.
+	mysqlBigintCastPattern = regexp.MustCompile(`(?i)\bAS\s+BIGINT\b`)
```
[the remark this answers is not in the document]
chrisguiney: Valid finding, out of scope for this change: this fix (commit 00bedb10f, and the origin/main merge in 2f06c9749) addresses only the Tx-scoped statement retention bug in prepared_stmt.go/live_stake.go and does not touch dialect_queryer.go's mysqlBigintCastPattern. Leaving this open rather than folding an unrelated regex fix into this PR's diff; worth a follow-up if not already tracked separately.

wolf31o2 (approved):

PR #4274, opened 2026-09-14
perf(database): raise wal_autocheckpoint, add DB metrics

chrisguiney:
Depends on #4250, #4270 and #4271 -- not mergeable until all three land first. The only new content on this branch is the final commit, `perf(database): raise wal_autocheckpoint, add DB metrics`.

## Problem

A write-amplification investigation found -wal file growth triggering a passive checkpoint on nearly every commit under the default wal_autocheckpoint (1000 pages), causing 60-100x more disk writes than the logical data being written. There was also no visibility into database operation rates at all.

## Change

- Raise wal_autocheckpoint well past the default so passive checkpoints fire far less often per commit.
- Add SQL and Badger operation-rate metrics (inserts/sec, WAL byte size, per-table counts).
- Add Grafana dashboards for both stores.

## Tests

metrics_test.go proves counters increment on real operations, not just that they're registered.


wolf31o2 (dismissed):

database/plugin/metadata/sqlstore/store.go
```diff
@@ -647,6 +658,22 @@ type queryer interface {
...
 
+// instrumentedQueryer applies dialect translation (newDialectQueryer) and,
+// when Config.PromRegistry was set, statement-count instrumentation
+// (countingQueryer) around db. Every call site that used to call
+// newDialectQueryer directly calls this instead, so
+// dingo_database_sql_operations_total's totals reflect Store's entire SQL
+// surface -- domain queries, committee pruning, deferred-index maintenance,
+// and the hot-statement cache -- from one place, rather than requiring every
+// call site to remember to instrument itself.
+func (s *Store) instrumentedQueryer(db queryer) queryer {
```
wolf31o2, line 675: countingQueryer is now the outermost type on every queryer, and two sites that type-assert past it were not updated. transaction_write.go:58 reads db.(dialectQueryer) to set a.mysql, and dialect_queryer.go:37's idempotence guard is defeated so operational.go:35 re-wraps and translates every operational query twice. With a MySQL dialect and PromRegistry set, a.mysql stays false and insertTransaction takes the RETURNING path MySQL cannot serve. Both are dormant only because the mysql and postgres providers do not pass PromRegistry, and nothing enforces that.

database/plugin/metadata/sqlite/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+// behind the write rate, not just running less often than before.
+func registerSQLiteFileMetrics(
+	reg prometheus.Registerer,
+	databasePath string,
+	store *sqlstore.Store,
+) {
+	walPath := databasePath + "-wal"
+	safeRegisterGaugeFunc(
+		reg,
+		sqliteFileMetricNamePrefix+"wal_bytes",
```
wolf31o2, line 65: The doc comment above states a WAL growing past 10000 pages between samples means checkpoints are falling behind. PASSIVE, FULL and RESTART backfill frames but never ftruncate the -wal file, so this gauge is a monotonic high-water mark: 42007552 bytes after sustained commits, unchanged by all three with busy=0 and checkpointed==log, and 0 only after TRUNCATE. Healthy and stalled checkpointing read identically. Raising wal_autocheckpoint to 10000 also makes ~40MB the permanent on-disk -wal size; document that instead.

DATABASE.md
```diff
@@ -1384,6 +1389,24 @@ it, so it rolls the ledger back to the blob tip instead. That rollback can be
...
+`database/plugin/metadata/sqlstore/metrics.go`) is incremented once per SQL
+statement at Store's single query chokepoint (`instrumentedQueryer`),
+classified by leading keyword (insert/update/delete/select/other) parsed past
+each query's sqlc-generated `-- name: X :verb` comment; it is a no-op unless
+`Config.PromRegistry` is set. `dingo_database_sql_wal_bytes` and
+`dingo_database_sql_disk_bytes` (`database/plugin/metadata/sqlite/metrics.go`)
+are pull-based gauges sampled at scrape time — a plain `os.Stat` of
+`metadata.sqlite-wal` and `Store.DiskSize()` respectively — the same pattern
+Badger's own cache gauges already use rather than a background ticker. All
+three are wired into the `examples/koios-parity-compose` Grafana stack: the
```
wolf31o2: No file in the tree provides this. examples/koios-parity-compose/grafana/dashboards holds only dingo-sync-progress.json, which references dingo_database_size_bytes and none of the three new metrics; there is no dingo-badger-storage dashboard, and no examples/ file changes on this branch. Add the dashboards or drop the claim.

database/plugin/metadata/sqlstore/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+) *prometheus.CounterVec {
+	if reg == nil {
+		return nil
+	}
+	counter := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: sqlMetricNamePrefix + "operations_total",
+		Help: "Total SQL statements issued against the metadata store, " +
+			"classified by leading SQL keyword (insert/update/delete/select/" +
+			"other). Counted at Store's single query chokepoint " +
+			"(instrumentedQueryer), so this covers every domain query, not " +
```
wolf31o2, line 46: This Help text claims coverage of every domain query, but deferred_indexes.go:264 in MissingCriticalDeferredIndexes still calls newDialectQueryer directly rather than instrumentedQueryer, so its statements are never counted. Route it through instrumentedQueryer or narrow the claim.

wolf31o2 (changes requested):
Three blockers. countingQueryer is now the outermost queryer type and defeats two unaudited type assertions (transaction_write.go:58 and newDialectQueryer's idempotence guard used by operational.go:35). DATABASE.md documents Grafana dashboards no file in the tree provides. The wal_bytes gauge cannot carry its documented checkpoint-health meaning. go-test is red at this head from main's own localstatequery_test.go breakage between 1a49911b0 and d30ffb34e, which this branch does not touch.

database/plugin/metadata/sqlstore/store.go
chrisguiney: Fixed in 7c38ad3a9: added unwrapDialectQueryer (dialect_queryer.go) to see through countingQueryer, and used it at newDialectQueryer's idempotence check and insertTransaction's mysql detection (transaction_write.go:58). Regression tests: TestNewDialectQueryerIdempotentUnderCountingQueryer, TestUnwrapDialectQueryerFindsDialectUnderCountingQueryer (dialect_test.go), TestInsertTransactionDetectsMySQLThroughCountingQueryer (transaction_batch_accumulator_test.go); each fails without the fix.

database/plugin/metadata/sqlite/metrics.go
chrisguiney: Fixed in 7c38ad3a9: rewrote the doc comment and Help text to describe wal_bytes as a monotonic high-water mark reset only by a TRUNCATE checkpoint, not a checkpoint-health signal, and documented the ~40MB permanent on-disk floor at the raised wal_autocheckpoint threshold. Also corrected the same claim in DATABASE.md's SQL-side metrics section.

DATABASE.md
chrisguiney: Fixed in 7c38ad3a9: dropped the Grafana dashboard claim from DATABASE.md; the section now only describes the metrics themselves.

database/plugin/metadata/sqlstore/metrics.go
chrisguiney: Fixed in 7c38ad3a9: MissingCriticalDeferredIndexes now calls s.instrumentedQueryer(s.readDB) instead of newDialectQueryer directly, so its per-index checks are counted like every other domain query. Regression test: TestMissingCriticalDeferredIndexesIsCounted (metrics_test.go), which fails (0 instead of len(CriticalManifest())) without the fix.

database/plugin/metadata/sqlite/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+) {
+	gauge := prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{Name: name, Help: help},
+		fn,
+	)
+	if err := reg.Register(gauge); err != nil {
+		var already prometheus.AlreadyRegisteredError
+		if !errors.As(err, &already) {
+			panic(err)
+		}
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 7c38ad3a9: safeRegisterGaugeFunc now unregisters the stale collector on AlreadyRegisteredError and re-registers the new GaugeFunc, instead of silently keeping the first store's callback. Regression test: TestSafeRegisterGaugeFuncReplacesStaleCollector (database/plugin/metadata/sqlite/metrics_test.go), which fails without the fix.

database/plugin/metadata/sqlstore/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+// (prepareHotStatements) prepares the same fixed, small statement list once
+// at Start. Counting there would misrepresent one-time preparation cost as
+// query volume.
+//
+// A call served by the hot-statement cache never reaches this type at all:
+// queryRowCached/execCached (prepared_stmt.go) resolve straight to a
+// *sql.Stmt via stmtForQueryer -- which unwraps past countingQueryer
+// entirely, by design, to reach the real *sql.Tx/*sql.DB -- and call
+// QueryRowContext/ExecContext on that *sql.Stmt directly, never on this
+// wrapper. queryRowCached and execCached count those calls explicitly
```
wolf31o2, line 148: The new metric claims to count every SQL statement, but `transactionBatchAccumulator.insertTransaction` executes its prepared `transactionInsert` directly at `transaction_write.go:75` and `:86`, bypassing `countingQueryer`; those inserts are therefore absent from `dingo_database_sql_operations_total`. Please instrument prepared-statement executions (and add a regression assertion for the batch accumulator), or narrow the metric contract to exclude them.

wolf31o2 (changes requested):
The SQL operation metric is not yet complete for all Store-managed execution paths; please cover the prepared batch-insert path as well.

database/plugin/metadata/sqlstore/metrics.go
chrisguiney: Fixed in b47ae020d: transactionBatchAccumulator now holds the Store's sqlOperations counter and increments it (classified via classifySQLOp) right before executing the cached transactionInsert statement, covering both the ExecContext/LastInsertId (MySQL) and QueryRowContext (RETURNING) paths. Regression test: TestBatchInsertTransactionIsCounted (metrics_test.go), which fails (0 instead of 1) without the fix. Also updated the counter's Help text and the countingQueryer doc comment to name this self-counting path alongside prepared_stmt.go's cache-hit counting, and corrected DATABASE.md's description of the chokepoint claim.

wolf31o2 (approved):

PR #4275, opened 2026-09-14
perf(database): add SQL query duration histogram, per-block op panel

chrisguiney:
Depends on #4250, #4270, #4271 and #4274 -- not mergeable until all four land first. The only new content on this branch is the final commit, `perf(database): add SQL query duration histogram, per-block op panel`.

## Change

Adds a query-duration histogram alongside the operation-rate counters from #4274, and a per-block SQL-operation-count panel on the SQLite storage dashboard, so query volume can be estimated as block-processing rate scales.


wolf31o2 (dismissed):

database/plugin/metadata/sqlstore/metrics.go
```diff
@@ -0,0 +1,257 @@
...
+	op, name := classifySQLStatement(query)
+	if q.counter != nil {
+		q.counter.WithLabelValues(op).Inc()
+	}
+	if q.duration == nil {
+		return q.queryer.QueryContext(ctx, query, args...)
+	}
+	start := time.Now()
+	rows, err := q.queryer.QueryContext(ctx, query, args...)
+	q.duration.WithLabelValues(op, name).Observe(time.Since(start).Seconds())
```
wolf31o2, line 250: Observe records dispatch latency, not query duration: database/sql returns *sql.Rows before the driver produces rows, so this runs before any row work. Measured through Store with a registry, a 300000-row SELECT recorded 0.000111s while iteration took 260ms, placing every multi-row query in the 100-200us bucket. 40 sqlc :many queries route here, as do sumUint64Rows and sumSignedRows (amounts.go), whose whole cost is row iteration. Observe once the rows are consumed, or narrow the metric's Help text to dispatch latency.

database/plugin/metadata/sqlstore/metrics.go
```diff
@@ -0,0 +1,257 @@
...
+	op, name := classifySQLStatement(query)
+	if q.counter != nil {
+		q.counter.WithLabelValues(op).Inc()
+	}
+	if q.duration == nil {
+		return q.queryer.QueryRowContext(ctx, query, args...)
+	}
+	start := time.Now()
+	row := q.queryer.QueryRowContext(ctx, query, args...)
+	q.duration.WithLabelValues(op, name).Observe(time.Since(start).Seconds())
```
wolf31o2, line 268: The single-row path is accurate and needs no change: measured against modernc.org/sqlite, db.QueryRowContext returned after 92ms for an aggregate SELECT whose Scan then took 21us, so execution completes before Observe runs. Same for the cached path in prepared_stmt.go. Reshaping either to time through Scan would move the observation without changing what it measures; the multi-row QueryContext path above is where the work happens after the call returns.

DATABASE.md
```diff
@@ -1384,6 +1389,44 @@ it, so it rolls the ledger back to the blob tip instead. That rollback can be
...
+wrapper entirely and are counted/timed explicitly instead in
+`queryRowCached`/`execCached` (`prepared_stmt.go`). `dingo_database_sql_wal_bytes`
+and `dingo_database_sql_disk_bytes` (`database/plugin/metadata/sqlite/metrics.go`)
+are pull-based gauges sampled at scrape time — a plain `os.Stat` of
+`metadata.sqlite-wal` and `Store.DiskSize()` respectively — the same pattern
+Badger's own cache gauges already use rather than a background ticker. All
+of these are wired into the `examples/koios-parity-compose` Grafana stack: the
+new `dingo-badger-storage` dashboard for Badger's existing metrics, and the
+SQL counters/gauges/histogram alongside the existing dashboards there,
+including a "SQL operations per block synced, by type" panel that divides
```
wolf31o2: No such panel exists in the tree: examples/koios-parity-compose/grafana/dashboards holds only dingo-sync-progress.json, docker-compose.yml mounts that directory, and no JSON there or under docs/dashboards references dingo_database_sql_*. The dingo-badger-storage dashboard named on line 1419 is absent too, arriving with #4274. Add the dashboard JSON this paragraph and the title describe, or drop the Grafana claims; make docs-parity passes and does not check them.

wolf31o2 (changes requested):
Two blockers: dingo_database_sql_query_duration_seconds measures dispatch rather than query completion on the multi-row path, and DATABASE.md documents Grafana wiring that is not in the repository. Reviewed range 736985fc9..6776fa3d5 (5 files). The merge at head drops nothing - the head function and metric-name sets are supersets of both parents except the deliberately removed classifySQLOp. The red go-test at head is main's ouroboros/localstatequery_test.go build failure, fixed on main in d30ffb34e; this branch does not touch ouroboros/.

database/plugin/metadata/sqlite/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+	)
+	safeRegisterGaugeFunc(
+		reg,
+		sqliteFileMetricNamePrefix+"disk_bytes",
+		"Total on-disk size of the SQLite metadata store: "+
+			"max(page_count*page_size, main file size) plus the WAL and "+
+			"shared-memory files, per Store.DiskSize.",
+		func() float64 {
+			size, err := store.DiskSize()
```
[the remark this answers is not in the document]
chrisguiney: Fixed in ee9030880: sqliteDiskSize now queries readDB (a separate, larger pool) through QueryRowContext with a 5s timeout, instead of writeDB (SetMaxOpenConns(1)) through unbounded QueryRow. Added TestDiskSizeDoesNotBlockOnBusyWriteConnection (shared_sqlstore_test.go), which holds a write transaction open and asserts DiskSize() still returns promptly; reverting the fix reproduces the stall and fails the test.

database/plugin/metadata/sqlstore/metrics.go
```diff
@@ -0,0 +1,257 @@
...
+	return row
+}
```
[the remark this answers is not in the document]
chrisguiney: Not applying: measured directly against modernc.org/sqlite, QueryRowContext returned after 92ms for an aggregate SELECT whose Scan then took 21us, so execution already completes before Observe runs -- QueryRowContext blocks until the driver executes the statement. Reshaping this to time through Scan would move the observation without changing what it measures.

database/plugin/metadata/sqlstore/prepared_stmt.go
```diff
@@ -0,0 +1,279 @@
...
+		if s.sqlQueryDuration == nil {
+			return stmt.QueryRowContext(ctx, args...)
+		}
+		start := time.Now()
+		row := stmt.QueryRowContext(ctx, args...)
+		s.sqlQueryDuration.WithLabelValues(op, name).
+			Observe(time.Since(start).Seconds())
+		return row
+	}
+	return db.QueryRowContext(ctx, query, args...)
```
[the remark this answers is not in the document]
chrisguiney: Not applying: the cached path was measured the same way as QueryRowContext (see the adjacent thread) and is already accurate -- the derived *sql.Stmt's QueryRowContext/ExecContext already blocks until the statement completes, so the observation already reflects full execution, not just dispatch. Only the multi-row QueryContext path has the dispatch-only gap; that is documented separately.

database/plugin/metadata/sqlstore/metrics.go
chrisguiney: Fixed in f0533d0c5: documented the dispatch-only measurement for the multi-row QueryContext path (Help text on dingo_database_sql_query_duration_seconds, plus a doc comment on countingQueryer.QueryContext) rather than reshaping it. The queryer interface (store.go) and every sqlc-generated DBTX interface it wraps (internal/query/{sqlite,postgres,mysql}/db.go, generated, DO NOT EDIT) both require QueryContext's return type to be the concrete *sql.Rows, so there is no wrapper type available to time through Close() without changing those generated interfaces across all three dialects' 40 :many call sites. Added TestSQLQueryDurationHistogramMultiRowMeasuresDispatchNotIteration: a 50k-row SELECT measured dispatch at ~115us against ~45ms full iteration, with the histogram recording a ~107us delta -- confirming the documented dispatch-only behavior.

database/plugin/metadata/sqlstore/metrics.go
chrisguiney: Confirmed, left unchanged: QueryRowContext and the cached-statement path (queryRowCached/execCached) are untouched in this PR.

DATABASE.md
chrisguiney: Fixed in bf596cd92 (merging #4274's own fix for the same claim forward) plus f0533d0c5: dropped the Grafana-dashboard claim from this section. It now documents only the metrics themselves, including the new histogram's dispatch-only caveat for QueryContext.

wolf31o2 (approved):

PR #4276, opened 2026-09-14
fix(database): force periodic WAL TRUNCATE checkpoint

chrisguiney:
Depends on #4250, #4270, #4271, #4274 and #4275 (SQL query duration histogram) -- not mergeable until all land first. The only new content on this branch is the final two commits: `fix(database): checkpoint WAL from a dedicated short-timeout connection` and the merge of #4275's latest fixes.

## Root cause

The new `dingo_database_sql_wal_bytes` metric (from the dependency PRs) showed the -wal file growing monotonically with zero dips over a 30+ minute window, even with checkpointing apparently healthy. Proved directly with a diagnostic program against a copy of a live 68MB -wal file: PASSIVE, FULL and RESTART checkpoint modes all leave the file at its full size (fully checkpointed, zero busy) -- only `SQLITE_CHECKPOINT_TRUNCATE` calls ftruncate on it. The existing `wal_autocheckpoint` mechanism only ever issues PASSIVE checkpoints, so the file could never shrink no matter how well checkpointing was working.

## Fix

An independent two-minute ticker attempts `PRAGMA wal_checkpoint(TRUNCATE)` against a dedicated connection opened fresh per attempt with a 250ms `busy_timeout`, never against `writeDB`. An earlier version issued the pragma against `writeDB` (`SetMaxOpenConns(1)`) on the theory that the pool would just serialize it behind an in-flight write; measured, that instead blocked `writeDB`'s only connection -- and therefore every write -- for up to the full `busy_timeout(30000)` (30.04s measured) whenever a `readDB` snapshot was open, since the checkpoint's busy handler can't complete until that snapshot closes and neither Go context cancellation nor a select loop can interrupt it once the driver call has started. The dedicated connection hits the same `busy=1` outcome but fails fast (271ms measured) and never contends with `writeDB`.

## Tests

`TestCheckpointWALDoesNotBlockWriteBehindReaderSnapshot` holds a `readDB` snapshot open and confirms both the checkpoint attempt and a concurrent `writeDB` write return promptly instead of blocking for tens of seconds. `TestCheckpointWALTruncatesFile` continues to confirm a TRUNCATE checkpoint actually shrinks the WAL file.

## Live validation

Over a 29-minute window on two live instances, `dingo_database_sql_wal_bytes` now oscillates genuinely instead of growing monotonically (one instance: 54.6MB to 21.2MB, a 60%+ drop).

## Summary by CodeRabbit

* **New Features**
  * Added optional Prometheus metrics for SQL operations, query durations, WAL size, and database disk usage.
  * Added periodic SQLite WAL maintenance to keep database files from growing unnecessarily.
  * Added configurable checkpoint scheduling for metadata stores.

* **Performance**
  * Improved recurring metadata queries through prepared-statement reuse.
  * Reduced database write amplification during SQLite synchronization and bulk operations.

* **Documentation**
  * Expanded architecture and database documentation for SQLite durability, WAL management, and monitoring metrics.


wolf31o2 (dismissed):

database/plugin/metadata/sqlite/shared_sqlstore.go
```diff
@@ -46,12 +47,121 @@ var sharedMemoryDBSequence atomic.Uint64
...
+// any reader still holding an old snapshot before it can truncate; if that
+// wait is exceeded, PRAGMA wal_checkpoint reports busy=1 with a partial
+// checkpointed count rather than an error or a blocked call -- logged at Warn
+// so a persistently blocked checkpoint (rather than a single slow tick) is
+// visible to an operator, and left for the next tick to retry rather than
+// retried in a loop here.
+func checkpointWAL(writeDB *sql.DB, logger *slog.Logger) func(context.Context) error {
+	return func(ctx context.Context) error {
+		var busy, walLog, checkpointed int
+		row := writeDB.QueryRowContext(ctx, "PRAGMA wal_checkpoint(TRUNCATE)")
```
wolf31o2: PRAGMA wal_checkpoint(TRUNCATE) invokes the busy handler, so this call blocks for the full busy_timeout(30000) while holding writeDB's only connection (SetMaxOpenConns(1)). With one open readDB snapshot it took 30.04s, a concurrent writeDB insert was blocked 29.99s, and the result was busy=1 with the WAL not truncated. readDB snapshots are opened on read paths from database/txn.go, so this is a periodic write stall on a running node. Bounding the wait (busy_timeout=250 on a dedicated Conn) returned in 271ms with the same busy=1 result.

database/plugin/metadata/sqlstore/store.go
```diff
@@ -429,15 +474,34 @@ func (s *Store) CloseContext(ctx context.Context) error {
 			// sessions where session_replication_role is connection-scoped.
 			_ = s.restoreNormalPragmas(ctx)
 		}
+		// Cancel both tickers up front, before waiting on either: each
+		// ticker's own goroutine observes its ctx.Done() and returns even if
+		// this function's ctx times out first, so unconditionally requesting
+		// both cancellations here -- rather than only reaching the second
+		// one after the first successfully waited -- guarantees neither
```
wolf31o2: This guarantee does not hold. A cancelled context does not interrupt an in-flight PRAGMA wal_checkpoint(TRUNCATE) sitting in the busy handler: cancelling at 300ms returned only after 30.04s. Node shutdown budgets 30s in total across all components (internal/config/config.go DefaultShutdownTimeout), so the select below falls to ctx.Done() and closePools() closes writeDB while the pragma is still executing on its only connection, and the ticker goroutine outlives CloseContext.

database/plugin/metadata/sqlite/shared_sqlstore.go
```diff
@@ -46,12 +47,121 @@ var sharedMemoryDBSequence atomic.Uint64
...
+// while TRUNCATE alone dropped it to 0. So dingo_database_sql_wal_bytes (a
+// plain os.Stat of that file -- see metrics.go) can never show a single
+// decrease under wal_autocheckpoint alone, no matter how well passive
+// checkpointing is working underneath -- the file's on-disk footprint is a
+// high-water mark that only grows or holds steady until something truncates
+// it. checkpointWAL is that something: a periodic, independent TRUNCATE
+// checkpoint so the WAL's on-disk size actually has a ceiling instead of
+// only ever growing to whatever the largest inter-checkpoint write burst has
+// been so far.
+const checkpointInterval = 2 * time.Minute
```
wolf31o2, line 121: origin/main carries no wal_autocheckpoint pragma, so #4274 raises the automatic threshold from SQLite's 1000-page default to 10000. Since PASSIVE never truncates, that alone raises the WAL's permanent high-water mark, a disk-footprint change rather than a correctness one, and #4274 does not depend on this change to be safe to merge. A two-minute interval instead repeats the blocking window measured below every two minutes.

DATABASE.md
```diff
@@ -1384,6 +1389,83 @@ it, so it rolls the ledger back to the blob tip instead. That rollback can be
...
+any individual `readDB` snapshot happens to be held open. It is issued
+against `writeDB`, not `readDB`: `writeDB` is capped at
+`SetMaxOpenConns(1)`, so `database/sql` itself serializes the checkpoint
+query behind any write transaction already using that sole connection —
+it cannot run concurrently with, or interrupt, an in-flight write, only ever
+between one write transaction's commit and the next one's begin. SQLite
+tracks WAL locks at the shared-memory/file level rather than per
+`database/sql` pool, so issuing the pragma from `writeDB` still correctly
+waits for (or reports busy against) a snapshot held open through one of
+`readDB`'s separate connections. If that wait exceeds `busy_timeout(30000)`,
```
wolf31o2: This records the busy outcome but not its cost: the pragma holds writeDB's sole connection for the whole busy_timeout, so every metadata write blocks behind it for up to 30 seconds per tick. Measured 30.04s checkpoint, 29.99s blocked concurrent write, busy=1, no truncation. The paragraph above states only that database/sql serializes the checkpoint behind an in-flight write; the converse, that writes serialize behind the checkpoint, belongs here.

database/plugin/metadata/sqlite/checkpoint_test.go
```diff
@@ -0,0 +1,149 @@
...
+		_, err = tx.ExecContext(
+			t.Context(),
+			"INSERT INTO checkpoint_probe (n) VALUES (1)",
+		)
+		require.NoError(t, err)
+		close(txStarted)
+		// Hold the sole writeDB connection open long enough that a
+		// concurrent checkpoint attempt has to wait for it if checkpointWAL
+		// is correctly serialized against an in-flight write transaction.
+		time.Sleep(200 * time.Millisecond)
```
wolf31o2: time.Sleep sequences this test, where AGENTS.md directs internal/test/testutil WaitForCondition or RequireReceive. The require.NoError calls in this goroutine reach t.FailNow(), which the testing package requires be called from the goroutine running the test. The assertion also records commitAt after Commit() returns, so a scheduling delay between those two statements can make checkpointReturnedAt precede commitAt while the property still holds.

wolf31o2 (changes requested):
Delta reviewed as 6776fa3d5..41dd46797, 6 files +491/-13. The forced TRUNCATE checkpoint holds writeDB's only connection for the full busy_timeout(30000) whenever a readDB snapshot is open: measured 30.04s with 29.99s of blocked writes and busy=1 with no truncation, and context cancellation does not interrupt that wait. The merge of 6776fa3d5 is clean, a superset of both parents. TestCheckpointWALTruncatesFile does fail without the fix. The red go-test is main's ouroboros build break at the time CI ran, not this branch.

database/plugin/metadata/sqlite/shared_sqlstore.go
chrisguiney: Fixed in 3453651: checkpointWAL now opens a dedicated connection per attempt with busy_timeout(250) instead of writeDB. TestCheckpointWALDoesNotBlockWriteBehindReaderSnapshot holds a readDB snapshot open and confirms both the checkpoint call and a concurrent writeDB insert return within 2s.

database/plugin/metadata/sqlstore/store.go
chrisguiney: Fixed in 3453651: rewrote the CloseContext comment to state cancellation stops the select loop but does not interrupt an in-flight driver call, and that the outlive-Close window is now bounded by checkpointBusyTimeout (250ms) instead of busy_timeout(30000).

database/plugin/metadata/sqlite/shared_sqlstore.go
chrisguiney: Already addressed: the permanent high-water-mark/disk-footprint implication of wal_autocheckpoint(10000) is documented in sqliteCommonPragmas's doc comment (merged with #4274) and does not depend on this PR to be safe.

DATABASE.md
chrisguiney: Fixed in 3453651: rewrote this section to state the converse (writes serialized behind the checkpoint) as the actual bug, with the 30.04s/29.99s measurement, and documents the dedicated-connection/250ms design with the 271ms measurement.

database/plugin/metadata/sqlite/checkpoint_test.go
chrisguiney: Fixed in 3453651: replaced the obsolete serialization test (checkpointWAL no longer touches writeDB) with TestCheckpointWALDoesNotBlockWriteBehindReaderSnapshot, which uses channels instead of time.Sleep, keeps require assertions on the main goroutine via a result channel, and has no commitAt/checkpointReturnedAt ordering race since it measures each call's own duration rather than comparing two wall-clock timestamps.

database/plugin/metadata/sqlite/shared_sqlstore.go
```diff
@@ -46,12 +47,156 @@ var sharedMemoryDBSequence atomic.Uint64
...
+// RESTART each left a 68062432-byte -wal file at exactly 68062432 bytes,
+// while TRUNCATE alone dropped it to 0. So dingo_database_sql_wal_bytes (a
+// plain os.Stat of that file -- see metrics.go) can never show a single
+// decrease under wal_autocheckpoint alone, no matter how well passive
+// checkpointing is working underneath -- the file's on-disk footprint is a
+// high-water mark that only grows or holds steady until something truncates
+// it. checkpointWAL is that something: a periodic, independent TRUNCATE
+// checkpoint so the WAL's on-disk size actually has a ceiling instead of
+// only ever growing to whatever the largest inter-checkpoint write burst has
+// been so far.
```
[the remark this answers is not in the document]
chrisguiney: Fixed in this push: reworded the doc comments to describe checkpointWAL as a best-effort periodic TRUNCATE attempt rather than an unconditional ceiling, and corrected the same drift in ARCHITECTURE.md (stale "against the write pool"), DATABASE.md, and metrics.go's exported gauge help text (both stale "dingo does not run [TRUNCATE] periodically", which predated this PR's own checkpointWAL addition).

database/plugin/metadata/sqlstore/prepared_stmt.go
```diff
@@ -161,20 +172,199 @@ func (s *Store) closePreparedStatements() {
...
+	if existing, ok := s.txStmts[tx][cached]; ok {
+		return existing
+	}
+	if s.txStmts == nil {
+		s.txStmts = make(map[*sql.Tx]map[*sql.Stmt]*sql.Stmt)
+	}
+	if s.txStmts[tx] == nil {
+		s.txStmts[tx] = make(map[*sql.Stmt]*sql.Stmt)
+	}
+	s.txStmts[tx][cached] = derived
```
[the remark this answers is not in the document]
chrisguiney: Major finding confirmed as a real mechanism, reproduced deterministically in TestTxScopedStmtDoesNotLeakAcrossConcurrentEviction (a test-only hook forces the derive/evict interleaving). It requires one *sqlTxn to be driven by two goroutines at once -- one deriving a statement while another commits/rolls back. Audited dingo's call sites: withWriteTransaction and database.Txn.Do both run the caller's callback to completion synchronously before Commit/Rollback, in the same goroutine; the one place that fans work across goroutines against a shared *database.Txn (queryShelleyUtxoWhole's resolve pool) deliberately gives each worker its own transaction instead, to avoid exactly this. Not reachable via any current call site. Documented the invariant on txScopedStmt's doc comment and backed it with the reproduction test rather than leaving it as prose.

database/plugin/metadata/sqlstore/store_test.go
```diff
@@ -455,6 +455,100 @@ func TestStoreMaintenanceLifecycle(t *testing.T) {
...
+	)
+	require.NoError(t, err)
+	maintenanceStarted := make(chan struct{})
+	maintenanceRelease := make(chan struct{})
+	var checkpointCalls atomic.Uint32
+	store, err := New(Config{
+		WriteDB: db,
+		Dialect: SQLiteDialect(),
+		Maintenance: func(ctx context.Context) error {
+			close(maintenanceStarted)
```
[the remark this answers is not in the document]
chrisguiney: Confirmed real: reproduced the double-close panic in 5/100 -race runs before the fix. Fixed by guarding close(maintenanceStarted) with an atomic call counter so only the first Maintenance invocation closes it; verified 200/200 -race runs clean after the fix.

database/plugin/metadata/sqlstore/store.go
```diff
@@ -53,6 +54,16 @@ type Config struct {
 	// migration readiness gate succeeds. SQLite uses it for periodic VACUUM.
 	Maintenance         func(context.Context) error
 	MaintenanceInterval time.Duration
+	// Checkpoint is an optional provider-owned hook run on its own ticker,
+	// independent of Maintenance/MaintenanceInterval: SQLite uses it to force
+	// a periodic WAL checkpoint (see sqlite.checkpointWAL) so the on-disk WAL
+	// file has a hard, bounded ceiling regardless of how long any individual
+	// read snapshot is held open, rather than relying solely on the
```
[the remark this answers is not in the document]
chrisguiney: Fixed in this push: reworded the Checkpoint hook's doc comment to describe a best-effort periodic TRUNCATE attempt rather than a hard, unconditional ceiling, consistent with the busy/retry behavior already documented elsewhere in this PR.

wolf31o2 (dismissed):

wolf31o2 (approved):

PR #4277, opened 2026-09-14
test(docsparity): check the Koios coverage table against the matrix

wolf31o2:
`make docs-parity` did not read `koiosCoverageMatrix`, so `ARCHITECTURE.md`'s
Koios coverage table could classify a field differently from the checker that
applies it and the target stayed green. The classification is the load-bearing
part of that table: a `PASS` covers only the **exact-match** and
**derived-match** fields, so a stale row tells an operator a field is compared
when it is not, or the reverse.

`TestArchitectureDocumentsKoiosCoverageMatrix` compares the endpoint, field and
classification of every table row against `koiosparity.KoiosCoverageMatrix` in
both directions. These now fail the target and name the row with its file and
line:

- a field the matrix carries and the table omits;
- a field the table carries and the matrix does not;
- a classification that differs between the two;
- a classification the code does not define;
- a field wildcard, such as `pvt_*`, matching nothing in the matrix.

The mapping/reason column is not compared, so rewording a reason does not fail
the check. A missing or renamed table header fails rather than skipping, so the
rule cannot go vacuous.

`ARCHITECTURE.md` records the new rule in both the coverage-contract paragraph
and the documentation-parity section.

---
## Summary by cubic
Fixes `make docs-parity` so the Koios coverage table in `ARCHITECTURE.md` is checked against `koiosparity.KoiosCoverageMatrix`; previously the table could drift from the matrix and the target stayed green. The classification is load-bearing because a `PASS` covers only exact-match and derived-match fields.

**Bug Fixes**
- `TestArchitectureDocumentsKoiosCoverageMatrix` compares every row's endpoint, field, and classification in both directions and reads every table under the coverage header.
- Fails on missing or extra fields, classification mismatches, unknown classifications, duplicate rows, empty field rows, and wildcards matching nothing; wildcards use the longest matching prefix (e.g., `pvt_motion_*` over `pvt_*`).
- Fenced code examples in the doc are ignored, so a table-shaped code block is not read as coverage rows.
- Mapping/reason prose is not compared, so rewording a reason does not fail the check.
- `ARCHITECTURE.md`, `Makefile`, and `README.md` now state that docs-parity checks the Koios coverage matrix.

<sup>Written for commit eda1a61a581113ff22c786a121707efd0b01142f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Updated the architecture documentation to identify the Koios coverage table as a source of truth.
  - Documented that the table tracks endpoint, field, and classification coverage.

- **Tests**
  - Added automated validation to detect mismatches between the documented coverage table and the Koios coverage matrix.
  - Checks now flag missing, undocumented, reclassified, or unmatched fields and endpoints during documentation parity verification.


internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,301 @@
...
+		class := unquote(cells[1])
+		for _, field := range strings.Split(cells[2], ",") {
+			field = unquote(field)
+			if field == "" {
+				continue
+			}
+			key := koiosFieldKey{endpoint: endpoint, field: field}
+			entry := koiosDocEntry{class: class, line: i + 1}
+			if strings.HasSuffix(field, "*") {
+				wildcard[key] = entry
```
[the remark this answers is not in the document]
wolf31o2: Valid; fixed in 13b663c2. Both maps now report a second row for the same `(endpoint, field)` key instead of assigning over the first, naming the line the key was already documented at. Parsing moved out of `koiosCoverageDocEntries` into `parseKoiosCoverageTable`, which returns the faults rather than calling `t.Errorf`, so the rule is covered by `TestKoiosCoverageTableRejectsDuplicateRows` against a table written to break it; without the check that test reports 0 problems instead of 2.

Class check: the other doc-derived maps in `internal/docsparity` are not lossy this way. `lint_coverage_test.go:142` maps to `bool`, so a duplicate carries the same value, and the `env` map in `release_test_coverage_test.go:56` is a deliberate workflow/job/step precedence merge.

internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,412 @@
...
+				docLocation(koiosCoverageDoc, i+1),
+				len(cells),
+			))
+			continue
+		}
+		endpoint := unquote(cells[0])
+		class := unquote(cells[1])
+		for _, field := range strings.Split(cells[2], ",") {
+			field = unquote(field)
+			if field == "" {
```
chrisguiney, line 150: A row whose Fields cell is empty records nothing and reports nothing, so it skips both the matrix comparison and the classification check. Adding `| /account_info | bogus-class |  | fields not filled in |` under the /tip row leaves go test ./internal/docsparity green, although the description says an undefined classification and a row the matrix lacks both fail. Report a row that yields no field as a problem, and pin it with a parser case beside TestKoiosCoverageTableRejectsDuplicateRows.

internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,412 @@
...
+			continue
+		}
+		matched := true
+		for j, want := range koiosCoverageHeader {
+			if cells[j] != want {
+				matched = false
+				break
+			}
+		}
+		if matched {
```
chrisguiney, line 115: Only the first table with this header is read. A second table with the same header later in ARCHITECTURE.md is never compared: adding one before Protocol parameter resolution with rows `/tip | unsupported | epoch_no` and `/nope | exact-match | bogus` leaves the test green, so the contradiction the duplicate-row rule rejects inside one table passes across two. Parse every table carrying this header, or report a repeated header as a problem, and cover it with a parser case.

internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,412 @@
...
+		t.Errorf("wildcard row for %v kept class %q, want the first row", wildcardKey, got)
+	}
+}
+
+// koiosWildcardFor returns the wildcard row covering key, if any.
+func koiosWildcardFor(
+	wildcard map[koiosFieldKey]koiosDocEntry,
+	key koiosFieldKey,
+) (koiosFieldKey, koiosDocEntry, bool) {
+	for _, candidate := range koiosSortedKeys(wildcard) {
```
chrisguiney, line 387: With two wildcards on one endpoint the first key in sort order wins, not the most specific one. Writing the /epoch_params cell as `pvt_*`, `pvt_motion_*` fails with `pvt_motion_* matches no field in koiosCoverageMatrix` although matrix fields start with pvt_motion_, and those fields are compared against the pvt_* row's class. Select the longest matching prefix.

ARCHITECTURE.md
```diff
@@ -8822,9 +8828,9 @@ documentation in agreement with the configuration it describes. Every rule
 derives its expectation from a source of truth in the tree rather than
 restating a value: the minimum Go release comes from the `go` directive in
 `go.mod`, the build contract from the Makefile's rule graph and `##` help
-text, and the DevNet topology from `internal/test/devnet/docker-compose.yml`,
-its `.env` profile default, and the wrapper scripts. Run the focused check
-with:
+text, the DevNet topology from `internal/test/devnet/docker-compose.yml`, its
+`.env` profile default, and the wrapper scripts, and the Koios coverage table
```
chrisguiney, line 8832: The docs-parity scope is also stated in the Makefile help text (Check docs against go.mod, the Makefile, and the DevNet compose file) and in README.md line 973 (Docs agree with go.mod, Makefile, compose). Neither names the Koios coverage table this change adds to the target.

chrisguiney (changes requested):
Each of these table or matrix edits fails go test ./internal/docsparity at 13b663c2: a removed row, an extra field, a reclassified row (doc side and matrix side), an added matrix field, an undefined class, a dead wildcard, a renamed header, a contradicting duplicate row, and the duplicate check reverted in place. Three table shapes still pass or fail on the wrong row; see inline. With current main merged, docsparity and architecture pass, and every check at head is green.

internal/docsparity/koios_coverage_test.go
wolf31o2: Confirmed at `13b663c2`. Adding `| /account_info | bogus-class |  | fields not filled in |` under the `/tip` row left `go test ./internal/docsparity` green: a row whose Fields cell names nothing contributes no `(endpoint, field)` pair, so the matrix comparison and the classification check both skip it.

`parseKoiosCoverageTable` now counts the fields a row yields and reports a row that yields none. The same injected row now fails with `ARCHITECTURE.md:6931: coverage row for /account_info names no field, so its classification "bogus-class" is never checked`.

`TestKoiosCoverageTableRejectsRowWithNoField` pins it beside `TestKoiosCoverageTableRejectsDuplicateRows`, and also asserts the complete row in the same table still parses, so one incomplete row does not cost the checks on the others. Removing the new report makes that test fail.

internal/docsparity/koios_coverage_test.go
wolf31o2: Confirmed at `13b663c2`. A second table under the same header, carrying `/tip | unsupported | epoch_no` and `/nope | exact-match | bogus`, left `go test ./internal/docsparity` green.

`parseKoiosCoverageTable` now collects every header match and reads each table into the same maps, rather than stopping at the first. The contradiction then reaches the duplicate-row rule and the unknown endpoint reaches the matrix comparison; the same injected table now produces three failures:

- `duplicate coverage row for /tip epoch_no, already documented at ARCHITECTURE.md:6905`
- `/tip epoch_no is documented as unsupported but koiosCoverageMatrix classifies it derived-match`
- `/nope bogus is documented but koiosCoverageMatrix has no such field`

`TestKoiosCoverageTableReadsEverySuchTable` covers both halves: the duplicate is reported, and the second table's other row reaches the maps so the matrix comparison sees it. Restoring the `break` at the first header makes that test fail.

internal/docsparity/koios_coverage_test.go
wolf31o2: Confirmed at `13b663c2`. Writing the `/epoch_params` Fields cell as `` `pvt_*`, `pvt_motion_*` `` failed with `ARCHITECTURE.md:6962: /epoch_params pvt_motion_* matches no field in koiosCoverageMatrix`, while `pvt_motion_no_confidence` and its siblings were compared against the `pvt_*` row's classification.

`koiosWildcardFor` now scans all candidates and keeps the longest matching prefix instead of returning the first in sort order. The same injected cell passes.

`TestKoiosWildcardForPrefersLongestPrefix` covers both directions: `pvt_motion_no_confidence` resolves to `pvt_motion_*` with that row's class, and `pvt_committee`, which only the broader row covers, still resolves to `pvt_*`. Restoring first-match selection makes that test fail on both the matched row and its class.

ARCHITECTURE.md
wolf31o2: Correct, and both now name it.

`Makefile` line 98: `docs-parity: ## Check docs against go.mod, the Makefile, the DevNet compose file, and the Koios coverage matrix`

`README.md` line 973: `make docs-parity   # Docs agree with go.mod, Makefile, compose, Koios matrix`

That matches the `ARCHITECTURE.md` documentation-parity paragraph this change already updated, so the three statements of the target's scope agree.

internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,591 @@
...
+		}
+		matched := true
+		for j, want := range koiosCoverageHeader {
+			if cells[j] != want {
+				matched = false
+				break
+			}
+		}
+		if matched {
+			headers = append(headers, i)
```
chrisguiney, line 130: This scan has no fence tracking, so a coverage table inside a fenced code block is read into the contract alongside the real one. Deleting the /tip epoch_no row from the table at ARCHITECTURE.md:6930 and quoting that same row inside a fenced markdown example leaves go test ./internal/docsparity green; at 13b663c2 that edit failed with documents no row for /tip epoch_no. ARCHITECTURE.md already carries 122 pipe lines inside fences. fenceTracker at parity_helpers_test.go:556 is what markdownTableRows uses for this; step it over the lines here and over the row scan below.

internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,591 @@
...
+		)
+	}
+
+	table := koiosCoverageTable{
+		exact:    make(map[koiosFieldKey]koiosDocEntry),
+		wildcard: make(map[koiosFieldKey]koiosDocEntry),
+	}
+	for _, header := range headers {
+		for i := header + 2; i < len(lines); i++ {
+			if !strings.HasPrefix(strings.TrimSpace(lines[i]), "|") {
```
chrisguiney, line 155: The row scan only stops at the first line without a pipe, so two coverage tables with no blank line between them make the first scan run through the second table's header and divider: they enter exact as Koios endpoint/Fields classified Classification and as ---/--- classified ---, and the second table's rows are then read twice, reporting the first as a duplicate of its own line. Stop this scan when i reaches the next entry in headers.

internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,591 @@
...
+					"%s: coverage row for %s names no field, so its "+
+						"classification %q is never checked",
+					docLocation(koiosCoverageDoc, i+1),
+					endpoint,
+					class,
+				))
+			}
+		}
+	}
+	if len(table.exact) == 0 {
```
chrisguiney: This guard ignores table.wildcard, so a table holding only wildcard rows returns the coverage table has no field rows, and the wildcard rows it does hold are dropped along with any problems already collected. Parsing a one-row table of /epoch_params unsupported pvt_* reports exactly that. Test len(table.exact)+len(table.wildcard) instead.

chrisguiney (changes requested):
All four earlier findings are fixed at 0c03db70, each causally: reverting the multi-header scan, the empty-Fields report or the longest-prefix selection reds its own new test, and Makefile, README and ARCHITECTURE.md now all name the Koios matrix. Reading every table introduced a regression: the scan ignores fenced code blocks, so a row shown in a fenced example now joins the contract. Deleting the /tip epoch_no row from the table and quoting it inside a fence leaves go test ./internal/docsparity green, where the same edit failed at 13b663c2.

internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,606 @@
...
+// The error covers the two conditions that leave nothing to check at all: no
+// such table, or no field rows in any of them. problems holds the per-row
+// faults, which are each worth reporting without abandoning the rest of the
+// table.
+func parseKoiosCoverageTable(doc string) (koiosCoverageTable, error) {
+	lines := strings.Split(doc, "\n")
+	insideFence := make([]bool, len(lines))
+	var fences fenceTracker
+	for i, line := range lines {
+		_, insideFence[i] = fences.step(line)
```
chrisguiney, line 107: The fence rule this commit adds is not pinned. Dropping these three lines so insideFence stays all false leaves go test ./internal/docsparity at exit 0, although a fenced example then joins the contract again - the finding at 0c03db70. A parser case beside TestKoiosCoverageTableReadsEverySuchTable, feeding a real table plus a fenced copy whose row contradicts it, reports a duplicate without this loop and no problems with it. This is the one rule here whose two failure directions are both silent: a fenced row admitted, or a real row dropped.

internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+
+	table := koiosCoverageTable{
+		exact:    make(map[koiosFieldKey]koiosDocEntry),
+		wildcard: make(map[koiosFieldKey]koiosDocEntry),
+	}
+	for headerIndex, header := range headers {
+		end := len(lines)
+		if headerIndex+1 < len(headers) {
+			end = headers[headerIndex+1]
```
chrisguiney, line 149: This bound is unpinned too. Restoring end = len(lines) keeps go test ./internal/docsparity at exit 0, because TestKoiosCoverageTableReadsEverySuchTable separates its two tables with a blank line and the scan breaks there. With the tables adjacent, the reverted scan reads the second header and divider as rows - endpoint Koios endpoint field Fields, and endpoint --- field --- - and then reports the second table's first row as a duplicate of its own line. A no-blank-line variant of that test reds without this bound.

internal/docsparity/koios_coverage_test.go
```diff
@@ -0,0 +1,606 @@
...
+					"%s: coverage row for %s names no field, so its "+
+						"classification %q is never checked",
+					docLocation(koiosCoverageDoc, i+1),
+					endpoint,
+					class,
+				))
+			}
+		}
+	}
+	if len(table.exact)+len(table.wildcard) == 0 {
```
chrisguiney, line 215: Restoring len(table.exact) == 0 here also leaves go test ./internal/docsparity at exit 0. A one-row table of /epoch_params unsupported pvt_* parses at head and returns the coverage table has no field rows once the wildcard term is dropped, so the rule holds and nothing checks it. The real table's /epoch_params rows carry pvt_* and dvt_*, so a table reduced to its wildcard rows is a reachable shape. A parser case asserting such a table parses and records the wildcard row closes it.

ARCHITECTURE.md
```diff
@@ -6907,7 +6907,13 @@ cmd/koios-parity/          # thin Cobra CLI wrapper
   **derived-match** field below matched. It does not claim parity for fields
   classified as **intentionally-incomparable** or **unsupported**. The same
   field-level matrix is emitted in every JSON report, and human-readable
-  status output reports the class counts. `internal/koiosparity/coverage_test.go`
+  status output reports the class counts.
+  `TestArchitectureDocumentsKoiosCoverageMatrix` (`internal/docsparity/`)
```
chrisguiney, line 6911: Checked at the head rather than from the replies: deleting the /tip epoch_no row and quoting it inside a fence now fails, as do removing the header and divider, renaming the header, fencing the whole table, reclassifying /epoch_params epoch_no, and a blank line splitting the table in half. go-test (Windows) ran 72m and its log blob is gone; dingo#4191's Windows job overran the same way and died 18 seconds later, a third overran and was cancelled at 15:44Z, and every other Windows job that day ran 12-34m. Windows coverage here is absent, not regressed.

README.md
```diff
@@ -970,7 +970,7 @@ make test                                    # All tests with race detection
 go test -v -race -run TestName ./package/    # Single test
 make bench                                   # Benchmarks
 make bench-mempool                           # Compare FIFO and DAG mempools
-make docs-parity                             # Docs agree with go.mod, Makefile, compose
+make docs-parity                             # Docs agree with go.mod, Makefile, compose, Koios matrix
```
chrisguiney, line 973: Current main merged into this head cleanly and builds: go build ./... exits 0, and go test ./internal/docsparity ./internal/architecture ./internal/koiosparity all pass on the merged tree, so the three coverage rows main added since the merge base are carried rather than dropped. Checks here started before main moved to dce82ce6, so the green ones cover an older base. No bot reviewed this head: the CodeRabbit status reads Review skipped, incremental reviews are disabled, and there is no cubic review record, check run or status context.

Makefile
```diff
@@ -95,7 +95,7 @@ lint: import-boundaries ## Run import-boundaries, golangci-lint, nilaway, and mo
 import-boundaries: ## Check reviewed package import boundaries
 	go test ./internal/architecture
 
-docs-parity: ## Check docs against go.mod, the Makefile, and the DevNet compose file
+docs-parity: ## Check docs against go.mod, the Makefile, the DevNet compose file, and the Koios coverage matrix
```
chrisguiney, line 98: CodeRabbit's only actionable finding here, duplicate (endpoint, field) keys assigning over each other, is already addressed: parseKoiosCoverageTable reports both maps and TestKoiosCoverageTableRejectsDuplicateRows covers it, confirmed by removing the fence loop, which surfaced a duplicate report from the fenced copy. This line, README.md and the ARCHITECTURE.md parity paragraph now all name the Koios coverage matrix, which closes the last finding from 13b663c2.

chrisguiney (dismissed):
e5f71254 clears all three findings at 0c03db70, each causally: reverting the fence tracking, the next-header bound or the wildcard-only guard in place makes a table written to break that rule parse wrongly again. None of the three is pinned, so each reverts to green - go test ./internal/docsparity exits 0 with any one of them removed. The four earlier rules each got a parser case; these three need the same. Windows is red from a runner-pool loss rather than this diff, and no bot reviewed this head; both are detailed inline.

arepala-uml (approved):

PR #4284, opened 2026-09-14
fix(leios): scope serving waiters to connection owners

wolf31o2:
Scope CertRB closure waiters to the concrete chainsync server that registered them. A delayed close callback cannot release a replacement connection's waiter, and late registration by an old owner is rejected.

Remove the obsolete ID-only Node close wrapper and retain NtC iterator, LocalStateQuery, and Notify cleanup through the owner callback. Regression tests cover stale release, stale registration, and current-owner cleanup.

Related to #3412; this addresses CertRB waiter ownership, not the issue's remaining fetch and manifest lifecycle scope. The inherited fixture now closes caller-owned error channels after connection shutdown. Both-peer diagnostics remain enabled; the initiating protocol-ID-8 startup failure remains tracked in #4287.

### Summary by cubic

Waiter cleanup matches the registered chainsync server owner, while registration checks that owner against the live connection. Node cleanup remains owner-aware and preserves NtC iterator, LocalStateQuery, and Notify cleanup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection lifecycle handling when connections are replaced or restarted.
  * Prevented stale connections from removing cursors, clients, or serving requests belonging to newer connections.
  * Ensured Leios serving requests are released only for the correct connection owner.
  * Preserved queued LeiosNotify delivery after a connection replacement.
  * Improved cleanup behavior across ChainSync, Local State Query, and node-to-client connections.

* **Documentation**
  * Updated architecture documentation to describe connection ownership and shutdown behavior.


node.go
```diff
@@ -1806,51 +1806,25 @@ func taintValue(relaxed bool) string {
...
 	if conn == nil {
 		return
 	}
-	n.handleConnManagerClosed(conn.Id(), isNtC, err)
+	if o := n.ouroboros(); o != nil && conn.ChainSync() != nil {
+		o.ReleaseLeiosServeWaitersOwner(conn.Id(), conn.ChainSync().Server)
+	}
+	if isNtC {
+		if n.chainsyncState != nil {
+			n.chainsyncState.RemoveClient(conn.Id())
```
chrisguiney: RemoveClient is keyed on ConnectionId alone while the two Leios cleanups in this same callback are now owner-scoped. connmanager releases admission before invoking this callback so a replacement can already be registered, and every Windows named-pipe NtC connection shares one ConnectionId (winio pipeAddress is a string value returned for both LocalAddr and RemoteAddr), so a closing client cancels a live client's chain iterator and deletes its state. AddClient also returns the existing entry for a colliding id. Scope this by the chainsync server owner too.

node.go
```diff
@@ -1806,51 +1806,25 @@ func taintValue(relaxed bool) string {
...
+		}
+		if o := n.ouroboros(); o != nil {
+			o.ReleaseLocalStateQueryAcquiredPoint(conn.Id())
```
chrisguiney: Same class as RemoveClient above: the pinned point is dropped by ConnectionId with no owner check, so a closing connection clears a replacement's acquired point and its next query re-acquires without the caller ever seeing a Release. Two of the four cleanups in this callback now carry an owner; this one and RemoveClient do not, which reads as a completed conversion when it is not.

ouroboros/leios_merged.go
```diff
@@ -208,21 +220,58 @@ func (o *Ouroboros) ReleaseLeiosServeWaiters(
...
 	// and is handled by the connection-owned callback.
 }
 
+// ReleaseLeiosServeWaitersOwner wakes only waits owned by one chainsync
+// server instance, so a delayed close callback cannot cancel a replacement.
+func (o *Ouroboros) ReleaseLeiosServeWaitersOwner(
+	connId ouroboros.ConnectionId,
+	owner *ochainsync.Server,
+) {
+	if owner == nil {
```
chrisguiney, line 234: This early return leaves a nil-owner waiter unreleasable. ReleaseLeiosServeWaitersOwner is the only release reachable from production, so a waiter registered with owner nil is never woken by a disconnect and parks for the full LeiosClosureWaitTimeout. RegisterLeiosServeWaiterForTesting produces exactly that when the connection is absent from connmanager or has no ChainSync. Reject a nil owner at registration, or keep an unscoped release reachable for those waiters.

ouroboros/leios_merged.go
```diff
@@ -208,21 +220,58 @@ func (o *Ouroboros) ReleaseLeiosServeWaiters(
 	waiters := o.leiosServeWaiters[connId]
 	delete(o.leiosServeWaiters, connId)
 	o.leiosServeWaitersMu.Unlock()
-	for _, ch := range waiters {
-		close(ch)
+	for _, waiter := range waiters {
+		close(waiter.ch)
```
chrisguiney, line 224: After this change ReleaseLeiosServeWaiters has no production caller: node.go:1817 is the only release site and it calls the owner-scoped method. The doc comment above still states it is called from the node's connection-closed callback, and ouroboros.go:197 points readers here. Correct both, or remove the method now that ReleaseLeiosServeWaitersOwner replaced it.

ouroboros/leios_serve_conn_test.go
```diff
@@ -110,6 +112,66 @@ func TestLeiosServeWaitReleasedByRealPeerDisconnect(t *testing.T) {
...
+	result := make(chan error, 1)
+	t.Cleanup(func() { f.o.ReleaseLeiosServeWaiters(f.conn.Id()) })
+	go func() {
+		_, err := f.o.serveLeiosCertRbWithWait(
+			block, ebHash, block.Slot, f.conn.Id(), f.conn.ChainSync().Server,
+		)
+		result <- err
+	}()
+	waitForLeiosServeWaiter(t, f)
+	replacement := f.conn.ChainSync().Server
```
chrisguiney, line 133: This replacement is the same Server pointer that is already serving, so both waiters share one owner and the stale owner is a fabricated new(ochainsync.Server) that never registered anything. The filter is pinned by this test, but no test exercises two real connection lifetimes on one ConnectionId, which is the production shape on a Windows named pipe. Removing the owner filter leaves every root-package test green, so the node wiring has no negative case.

node_ntc_conn_closed_test.go
```diff
@@ -80,28 +81,83 @@ func newHandleConnManagerClosedTestNode(t *testing.T) *Node {
 	}
 }
 
-// TestHandleConnManagerClosed_NtC_ReleasesChainsyncClientState reproduces
+func newHandleConnManagerClosedOwnerConn(
```
chrisguiney, line 71: This change removes the last caller of newNtCTestConnId at line 58 and leaves the helper defined. .golangci.yml sets run.tests to false, so the linter will not report the dead helper.

chrisguiney (changes requested):
The owner filter is correct and causally covered: reverting it in place fails TestLeiosServeWaiterRejectsReplacedOwner and TestLeiosServeWaitReleaseKeepsReplacementOwner, and passing a nil owner from node.go fails both node-level release tests. Two same-class cleanups in the rewritten callback are still keyed on ConnectionId alone, a nil-owner waiter has no release path left, and one doc comment names a call site this change removed. Merged-tree build, vet, lint, docs-parity, import-boundaries and the touched packages under -race are green against current main.

node.go
```diff
@@ -1806,51 +1806,27 @@ func taintValue(relaxed bool) string {
...
 		return
 	}
-	n.handleConnManagerClosed(conn.Id(), isNtC, err)
+	if o := n.ouroboros(); o != nil && conn.ChainSync() != nil {
+		o.ReleaseLeiosServeWaitersOwner(conn.Id(), conn.ChainSync().Server)
+	}
+	if isNtC {
+		if n.chainsyncState != nil {
+			n.chainsyncState.RemoveClientOwner(
+				conn.Id(), conn.ChainSync().Server,
```
chrisguiney: conn.ChainSync() is dereferenced here without the nil guard that lines 1817 and 1829 both apply, so an NtC close on a connection carrying no chainsync protocol panics. This is the failure in all four go-test jobs, and it reproduces locally at head and on the tree merged with the base branch: TestHandleConnManagerClosedOwner_NilOuroboros and TestHandleConnManagerClosedOwner_NtC_ReleasesChainsyncClientState both segfault at node.go:1823. Guard with conn.ChainSync() != nil as the two adjacent calls already do.

chainsync/chainsync.go
```diff
@@ -409,6 +411,38 @@ func (s *State) RemoveClient(connId connection.ConnectionId) {
...
+
+// RemoveClientOwner removes a client only when the close belongs to its
+// currently registered server instance.
+func (s *State) RemoveClientOwner(
+	connId connection.ConnectionId,
+	owner *ochainsync.Server,
+) {
+	s.Lock()
+	defer s.Unlock()
+	if current, ok := s.clientOwners[connId]; !ok || current != owner {
```
chrisguiney: This early return skips the release when no owner was recorded, so s.clients keeps the entry and its ChainIter is never cancelled. SetClientOwner is a second lock acquisition after AddClient in chainsyncServerFindIntersect and chainsyncServerRequestNext, so a close landing in that window leaks the iterator, and the later SetClientOwner then records a dead connection as the owner. With the node.go nil guard applied, TestHandleConnManagerClosedOwner_NtC_ReleasesChainsyncClientState still fails here. Record the owner inside AddClient, or release unconditionally when none is recorded.

node.go
```diff
@@ -1806,51 +1806,27 @@ func taintValue(relaxed bool) string {
...
+			)
+		}
+		if o := n.ouroboros(); o != nil {
+			o.ReleaseLocalStateQueryAcquiredPoint(conn.Id())
```
chrisguiney: ReleaseLocalStateQueryAcquiredPoint is still keyed on ConnectionId alone while the two cleanups around it now carry an owner, so a stale close still clears a replacement connection pinned point and its next query re-acquires without the caller ever seeing a Release. Raised on the previous head and unchanged here.

ouroboros/ouroboros.go
```diff
@@ -194,8 +194,8 @@ type Ouroboros struct {
 	// The chainsync server callback owns gouroboros's receive loop while it
 	// runs, so Protocol.DoneChan() cannot close underneath it; the release
 	// signal has to come from connmanager's per-connection ErrorChan watcher
-	// instead (see ReleaseLeiosServeWaiters).
-	leiosServeWaiters   map[ouroboros.ConnectionId][]chan struct{}
+	// instead (see ReleaseLeiosServeWaitersOwner).
```
chrisguiney: The NtN half of this cleanup is still ConnectionId-only: HandleConnClosedEvent calls chainsyncState.RemoveClient(connId) at ouroboros.go:757, and connmanager evicts and replaces a colliding NtN connection before the ErrorChan watcher publishes that event. TestHandleConnManagerClosedOwner_NtN_LeavesStateForEventBusPath pins that route as the NtN owner, so the stale-close-clobbers-replacement defect this change fixes for NtC stays open for NtN.

ouroboros/leios_merged.go
```diff
@@ -194,35 +206,68 @@ func (o *Ouroboros) releaseLeiosServeWaiter(
...
 	// and is handled by the connection-owned callback.
 }
 
+// ReleaseLeiosServeWaitersOwner wakes only waits owned by one chainsync
+// server instance, so a delayed close callback cannot cancel a replacement.
+func (o *Ouroboros) ReleaseLeiosServeWaitersOwner(
+	connId ouroboros.ConnectionId,
+	owner *ochainsync.Server,
+) {
+	if owner == nil {
```
chrisguiney, line 234: A nil owner still returns without releasing, and ReleaseLeiosServeWaiters is now documented as a test helper, so a waiter registered with a nil owner has no reachable release and parks for the full closure wait timeout. RegisterLeiosServeWaiterForTesting produces exactly that shape when the connection is absent from connmanager or has no ChainSync. Reject a nil owner at registration instead. Raised on the previous head and unchanged here.

chrisguiney (dismissed):
Commit c47fb3b5 regresses this branch: node.go:1823 panics on an NtC close where the connection has no chainsync protocol, reddening all four go-test jobs, and adding the missing nil guard still leaves TestHandleConnManagerClosedOwner_NtC_ReleasesChainsyncClientState failing because RemoveClientOwner will not release a client whose owner was never recorded. Both reproduce at head and on the merged base. No bot reviewed this head: CodeRabbit declines the base branch as non-default, and cubic left no check run, status context or review record, so the cubic text in the description is not one.

arepala-uml (approved):

PR #3961, opened 2026-09-05
feat(kupo): add compatible API

wolf31o2:
Adds a Kupo v2.12-compatible API backed by Dingo's complete API-mode index, with fixed `["*"]` patterns and derived checkpoints.

## Summary

- Adds an optional built-in Kupo API provider with matches, datum, script, checkpoint, metadata, health, and metrics endpoints.
- Uses complete API storage mode to retain the UTxO history and script data needed by Kupo queries.
- Keeps streamed match results on a coordinated metadata/blob snapshot across concurrent commits and rollbacks.
- Caps coordinated snapshots at one below the metadata read-pool size and holds admission for the snapshot's lifetime, so a client-paced `/matches` stream cannot take the connection rollback needs for operational reads.
- Uses the shared API listener lifecycle for concurrent start, stop, cancellation, TLS, and socket teardown behavior.
- Documents the Kupo configuration, storage, lifecycle, and API contract in `ARCHITECTURE.md`, `DATABASE.md`, and `README.md`.

## Compatibility notes

- Matching uses the global `["*"]` pattern set.
- `/metadata/{slot_no}` resolves to the first canonical block at or after the requested slot.
- A metadata request for a slot past the newest indexed block returns HTTP 400. This is a deliberate divergence: Kupo waits for the next block through a chain-sync client, while Dingo answers from committed storage and does not hold the request open until a block is forged.
- Checkpoint lookup uses bounded forward search and does not require reverse blob iteration.

PR #4289, opened 2026-09-14
fix(apilistener): gate starts and bound the context monitor

wolf31o2:
Two lifecycle gaps in `internal/apilistener`, fixed in the shared component rather than in any one API server.

## A start is invisible to a concurrent `Stop`

A start only becomes visible to the protocol when it publishes its server. Everything before that — reading config, building a handler chain, being descheduled — is a window in which a `Stop` finds nothing to detach, returns `nil` reporting the server down, and lets the start it could not see bind the port behind it.

`Listener` gains `BeginStart`/`EndStart`. Every `Start` takes the gate as its first statement and releases it with a `defer`, and `Stop` waits an in-flight start out, bounded by its own context.

Dingo's composition does not reach this window today: `plugin.Host` constructs, starts and stops one instance on a single goroutine, and `(*Node).shutdown` takes `liveLifecycleMu` before stopping the API capabilities — the same gate `Restore`/`Truncate` hold across `reinitializeAPIServers`. The gate holds the protocol's contract rather than closing a reachable production defect.

## The context monitor outlives its server

Each `Start` launched a monitor waiting only on the caller's context. That context is the node context, which stays live across a capability restart and for the rest of the process, so a stopped server's monitor kept the `http.Server`, its handler chain, and through that the database those handlers were built over. There is one monitor per `Start`, and every live `Restore` or `Truncate` performs another `Start`, so the retained set grows with operator actions and each entry pins a database that restore has already replaced.

`Listener.Watch` now owns the monitor and exits when the server is detached, by `Stop`, by a restart, or by a failed bind unpublishing. A cancelled context still brings the server down and releases the port.

## One mechanism

`api/blockfrost`, `api/mesh` and `api/utxorpc` drop their own monitor goroutines and their `Take`/`AwaitTeardown`/`Shutdown` sequences. A server now calls `BeginStart`/`Publish`/`Watch`/`Bind` in `Start` and a single `Listener.Stop` in `Stop`; `Take`, `TakeIf`, `Shutdown` and `AwaitTeardown` are unexported so the lifecycle has one implementation instead of one per server. `ShutdownFunc` stays the single axis servers differ on, so `api/utxorpc` keeps its hard-`Close` escalation.

Every wait in `Stop` is bounded by the caller's context, so the node's shutdown deadline still bounds each server's stop.

`api/kupo` on #3961 calls the now-unexported methods and carries its own `beginStart`/`startDone` gate and `serverLifecycle` cancellation, which is the duplication this replaces. It needs the same four-call `Start` and one-call `Stop` before it merges.

Related to #3217, which covered releasing the socket but neither of these.

---
## Summary by cubic
Fixes two lifecycle gaps in `internal/apilistener`: `Stop` now waits out a `Start` still in flight, and the context monitor for a server exits when that server is detached instead of living until the node context ends. The start/stop race is not currently reachable under Dingo's composition, but the listener now enforces the protocol; the monitor leak is reachable and grows with every restore or truncate, holding stopped `http.Server`s and their database-backed handler chains.

`api/blockfrost`, `api/mesh`, and `api/utxorpc` now use the shared `Listener.Stop` and `Listener.Watch`; `Take`, `TakeIf`, `Shutdown`, and `AwaitTeardown` are unexported. Each package also gained tests that pin the start gate and the context-monitor wiring, so removing either from `Start` would fail the suite. A cancelled context still tears the server down within the same 30-second bound, and every wait in `Stop` is bounded by the caller's context, so the node shutdown deadline still applies.

<sup>Written for commit 340707fab533d255ec9c9cdaf48cccf5314eddec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4289?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added coordinated start and stop handling for API servers.
  - Improved graceful shutdown when startup is still in progress.
  - Added automatic monitoring so cancelled contexts trigger server shutdown within a bounded period.
  - Improved handling when servers are stopped, replaced, or fail during startup.

- **Documentation**
  - Updated lifecycle documentation with the revised startup, monitoring, and teardown terminology.


api/utxorpc/utxorpc.go
```diff
@@ -177,48 +177,28 @@ func (u *Utxorpc) Start(ctx context.Context) error {
 	if u.config.Mempool != nil && isNilInterface(u.config.Mempool) {
 		return errors.New("utxorpc: Mempool must not be a typed nil")
 	}
+	startDone, err := u.listener.BeginStart()
+	if err != nil {
+		return err
+	}
+	defer u.listener.EndStart(startDone)
```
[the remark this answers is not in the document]
wolf31o2: Already addressed at the current head. `api/utxorpc/utxorpc.go:168` calls `BeginStart` as the first statement of `Start`, with `defer EndStart(startDone)` ahead of the EventBus and typed-nil dependency guards, so the gate is held across validation. The same ordering holds in `api/mesh/mesh.go:196` and `api/blockfrost/blockfrost.go:288`; those three are the only `BeginStart` call sites.

internal/apilistener/listener.go
```diff
@@ -184,61 +243,98 @@ func (l *Listener) take(match *http.Server) (*Job, chan struct{}) {
...
+//
+// It is the whole of a server's Stop: waiting out a start still in flight,
+// detaching, and then either running the teardown or waiting on the one it
+// lost. Every wait is bounded by ctx, so the caller's shutdown deadline bounds
+// this call whatever the server is doing, except for whatever bound a
+// ShutdownFunc chooses for itself.
+func (l *Listener) Stop(ctx context.Context, fn ShutdownFunc) error {
+	for {
+		l.mu.Lock()
+		startDone := l.startDone
```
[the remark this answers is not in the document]
wolf31o2: Already addressed at the current head. `Stop` reads `startDone` and detaches under a single acquisition of `l.mu`: `internal/apilistener/listener.go:293-315` takes the lock, inspects the gate, and calls `takeLocked(nil)` before unlocking, so no `BeginStart` can be admitted between the check and the detach. `take` now has one caller left, the context monitor in `Watch`, which is not gate-sensitive.

api/blockfrost/blockfrost.go
```diff
@@ -285,6 +285,12 @@ func (b *Blockfrost) handler() http.Handler {
 func (b *Blockfrost) Start(
 	ctx context.Context,
 ) error {
+	startDone, err := b.listener.BeginStart()
```
chrisguiney, line 288: Missing contract test: nothing in api/blockfrost pins this wiring. Deleting the BeginStart/EndStart lines from Start, or the Watch call at line 311, leaves `go test ./api/blockfrost/` passing, so the start gate or the context monitor can be dropped from this server with the suite green. Add package tests that hold the gate through `b.listener.BeginStart()` and assert `Start` is refused with "start already in progress", and that cancelling the Start context releases the port, as mesh's TestServerShutdownOnContextCancel does.

api/utxorpc/utxorpc.go
```diff
@@ -165,10 +165,16 @@ func isNilInterface(v any) bool {
 }
 
 func (u *Utxorpc) Start(ctx context.Context) error {
+	startDone, err := u.listener.BeginStart()
```
chrisguiney, line 168: Missing contract test: nothing in api/utxorpc pins this wiring. Deleting the BeginStart/EndStart lines from Start, or the Watch call at line 201, leaves `go test ./api/utxorpc/` passing, so the start gate or the context monitor can be dropped from this server with the suite green. Add package tests that hold the gate through `u.listener.BeginStart()` and assert `Start` is refused with "start already in progress", and that cancelling the Start context releases the port, as mesh's TestServerShutdownOnContextCancel does.

api/mesh/mesh.go
```diff
@@ -193,6 +193,12 @@ func NewServer(cfg ServerConfig) (*Server, error) {
 
 // Start starts the HTTP server in a background goroutine.
 func (s *Server) Start(ctx context.Context) error {
+	startDone, err := s.listener.BeginStart()
```
chrisguiney, line 196: Missing contract test: deleting the BeginStart/EndStart lines from Start leaves `go test ./api/mesh/` passing. TestConcurrentStartStopNeverLeavesThePortBound cancels the Start context in every iteration, so the monitor tears down any start a Stop missed and the gate is never observed. The Watch call is pinned by TestServerShutdownOnContextCancel; the gate is not. Add a test that holds the gate through `s.listener.BeginStart()` and asserts `Start` is refused with "start already in progress".

chrisguiney (changes requested):
The internal/apilistener fixes are pinned: reverting the Stop gate wait, the Watch detach exit, the replaced-server early return, the EndStart identity guard, or the BeginStart rejection each fails its new test. The server call sites that apply the protocol are not pinned (inline). go-test (Windows) at this head fails with an access violation in the ouroboros package, which does not import these packages; main run 34921657064 at 49fe91e8 shows the same fault.

api/blockfrost/blockfrost.go
wolf31o2: Correct; both gaps closed in 340707fa. api/blockfrost/listener_test.go now holds TestStartIsRefusedWhileAnotherStartHoldsTheGate, which takes the gate through BeginStart and requires Start to fail with the start-already-in-progress error and publish no server, and TestServerShutdownOnContextCancel, which cancels the Start context and requires the port to stop accepting. Deleting the BeginStart/EndStart pair fails the first and nothing else in the package; deleting the listener.Watch call fails the second and nothing else.

api/utxorpc/utxorpc.go
wolf31o2: Correct; both gaps closed in 340707fa. api/utxorpc/listener_test.go now holds TestStartIsRefusedWhileAnotherStartHoldsTheGate, which takes the gate through BeginStart and requires Start to fail with the start-already-in-progress error and publish no server, and TestServerShutdownOnContextCancel, which cancels the Start context and requires the port to stop accepting. TestUtxorpc_StopObservesCtxCancellation covers the Stop context, not this one. Deleting the BeginStart/EndStart pair fails the first and nothing else in the package; deleting the listener.Watch call fails the second and nothing else.

api/mesh/mesh.go
wolf31o2: Correct; closed in 340707fa. api/mesh/listener_test.go now holds TestStartIsRefusedWhileAnotherStartHoldsTheGate, which takes the gate through BeginStart and requires Start to fail with the start-already-in-progress error and publish no server. Deleting the BeginStart/EndStart pair from Start fails it and no other test in the package. The Watch call remains pinned by TestServerShutdownOnContextCancel. The class is the three apilistener consumers on this branch, and all three now carry both a gate test and a Start-context test.

chrisguiney (approved):

Issue #4215, opened 2026-09-12
devnet: txpump quarantine equals the accelerated scenario timeout


chrisguiney:
## Current status

Implemented by merged #4224. Both launchers select 100 slots for accelerated runs and retain 600 slots for canonical runs, with profile/slot-length budget checks. Output quarantine is a scheduling delay, not proof of on-chain confirmation or rollback recovery; rollback-aware wallet reconciliation is tracked separately in #4280.

## Problem

The previous accelerated DevNet configuration used a 600-slot txpump output quarantine. At its 0.5-second slot length, that was 300 seconds: the same as the scenario's five-minute hard timeout. Spent outputs therefore could not fund further rounds during that scenario.

The affected settings are in `internal/test/devnet/docker-compose.yml`; both accelerated YAML profiles use the same slot length. The hard timeout comes from `ReferenceRunnerBudget` in `internal/test/devnet/timeline.go`.

## Acceptance criteria

- The accelerated quarantine fits comfortably within the scenario budget and propagation deadline, checked against each profile's slot length and maximum cooldown.
- Regression coverage exercises both launchers and profiles and rejects a window reaching the hard timeout.
- Canonical runs retain the 600-slot window, including when the caller inherits an accelerated setting.
- Existing propagation assertions remain unchanged.

Related to #3180. This timing correction does not establish that every transaction-propagation failure has been resolved.

Issue #3497, opened 2026-08-25
ci: lint covers only the root module on linux, leaving nested modules and build-tagged files unchecked


wolf31o2:
`.github/workflows/golangci-lint.yml` invokes `golangci-lint-action` with no
`working-directory` and no GOOS matrix, so the `lint` check covers the root
module on linux/amd64 only. Two nested modules and every build-constrained
file are outside its reach, and a green `lint` says nothing about them.

Measured on `origin/main` at 94ebfc56:

| Scope | Why CI misses it | Issues |
| --- | --- | --- |
| `internal/test/antithesis` | own `go.mod` | 6 |
| `examples/dingo-gov-lens` | own `go.mod` | 7 |
| root module, `GOOS=windows` | 14 files behind `//go:build windows` | 4 |

Reproduce:

```
(cd internal/test/antithesis && golangci-lint run ./...)
(cd examples/dingo-gov-lens && golangci-lint run ./...)
GOOS=windows golangci-lint run ./...
```

The gap is load-bearing, not theoretical. #3436 merged with a green `lint`
check while introducing a new gosec G115 in
`internal/test/antithesis/internal/txpump/config.go` (`confirmationDelay`);
linting that module against `origin/main` before the merge gave 4 issues and
after it 5. The conversion is guarded and the finding is a false positive, but
the annotation the rest of the tree applies to that pattern was never
prompted for.

The four `GOOS=windows` findings are dead code rather than a missing check:
`checkFilePermissions` -> `checkSecurityDescriptor` -> `checkSDDL` in
`keystore/keyfile_windows.go`, plus `insecureSIDs`, form a path-based
permission-check chain reachable only from `keyfile_windows_test.go`. The
live path is `checkOpenFilePermissions`, which exists on both platforms and
has production callers. The chain looks superseded by the handle-based one and
kept alive by its own tests.

Suggested shape: one step per module via `working-directory`, and a second
root-module run with `GOOS=windows` set. Whether the pre-existing 17 findings
are fixed or baselined is a separate decision from closing the coverage gap;
adding the runs without triaging them first would turn `lint` red on `main`.

Related: the same class of blind spot applies to any future nested module, so
a matrix keyed on discovered `go.mod` files ages better than three hardcoded
steps.

Issue #3257, opened 2026-08-21
Accelerated conformance can miss transaction-bearing blocks after early forks


wolf31o2:
## Summary

The accelerated conformance DevNet can accept thousands of txpump submissions without observing a transaction-bearing block before the propagation deadline.

## Evidence

- Reproduced twice consecutively with `DEVNET_BLOCK_PIPELINE_ENABLED=true DEVNET_BLOCK_PIPELINE_VALIDATE_ENABLED=true ./internal/test/devnet/run-tests.sh --conformance --accelerated`.
- Both runs reached three healthy nodes, common block propagation, and continued chain growth.
- The latest run accepted 3,848 submissions, filled Dingo's 1 MiB mempool, then dropped a dependent transaction after fork revalidation because its input was absent.
- The observer timed out after 70 seconds waiting for a transaction-bearing block. At failure, cardano-producer and cardano-relay agreed at slot 181/block 63; Dingo was at slot 119/block 47 after normal rollbacks.
- Dingo logged 46 forged chain extensions and no block-pipeline, operational-certificate, or header-validation failures.
- An earlier identical profile completed successfully, so the failure is timing/fork dependent.

## Acceptance criteria

- The accelerated conformance profile reliably produces and observes a transaction-bearing block despite early forks.
- txpump does not build an immediately invalid dependency chain or saturate the mempool before the network is ready to include transactions.
- A regression test fails on the current behavior and passes with the fix.

PR #4290, opened 2026-09-14
fix(txpump): reconcile wallet inputs with node snapshots

### LINKED ISSUE #4280: fix(txpump): reconcile wallet outputs after rollback
Fix proposed in #4290.

Adding accepted payment outputs after `confirmationDelay()` does not verify that their parent remains on the selected chain. A parent removed by rollback can leave outputs available for later payments, which fail with missing inputs. The scheduling adjustment in #4224 does not provide rollback recovery; #4215 separates those concerns.

The reconciliation fix must also observe transaction presence before acquiring its ledger snapshot. Reversing that order can restore spent funding when confirmation occurs between observations. Change outputs from delegation, governance and Plutus workloads must remain within the addresses queried by the wallet.

Acceptance criteria:

- Reconcile reusable outputs against authoritative chain state without treating elapsed time as confirmation.
- Preserve valid outputs, signing metadata and payment generation through rollback and recovery.
- Cover accepted parent, inclusion, rollback, surviving-parent control and confirmation between protocol observations.
- Verify actual submitted non-payment change addresses, including Plutus unlocks.
- Retain wallet state and skip the batch when reconciliation fails.
- Keep propagation assertions; do not hide rejections by increasing quarantine or suppressing failures.

wolf31o2:
Keyed txpump wallets are reconciled against the node before each batch. LocalTxMonitor presence is observed before the LocalStateQuery UTxO snapshot, but the two observations are independent: a producer can drop a forged transaction from the mempool before the spend is visible to LSQ. One absent observation is therefore ambiguous and keeps the transaction's inputs reserved for a further round; a second absent observation retires the record and restores only the inputs the snapshot still carries. Presence clears the grace, so every present-to-absent transition gets its own round. Failed queries retain wallet state and skip submission.

Signing keys and output pacing survive reconciliation. Change from the stake delegation, governance and Plutus workloads returns to wallet-controlled addresses, including Plutus unlocks. Payment address selection is unchanged. The nested harness moves to gouroboros v0.204.7 for its LocalStateQuery acquisition-order fix, which requires plutigo v0.7.0.

A nested module has its own `go.mod`, so the repository-root `go test ./...` never reached `internal/test/antithesis` or `examples/dingo-gov-lens`: their packages, including the tests added here, were untested by every configured check. The Linux job now runs each nested module, and a docsparity rule derives the module list from the tree so a module added later cannot go untested. This extends to go-test the per-module rule #3497 established for lint.

Fixes #4280. Related to #4215 and #3257.


internal/test/antithesis/internal/txpump/client.go
```diff
@@ -107,6 +110,132 @@ func (c *NodeClient) SubmitTx(eraID uint16, txBytes []byte) error {
...
+				TxHash: id.Hash.String(),
+				//nolint:gosec // ledger index is bounded by protocol
+				Index:   uint32(id.Idx),
+				Amount:  amount.Uint64(),
+				address: raw,
+			},
+		)
+	}
+	presence = make(map[string]bool, len(txIDs))
+	monitor := c.conn.LocalTxMonitor()
```
chrisguiney: Mempool presence is read after the UTxO query. A pending tx adopted between the two reads is reported absent while its inputs are still in the earlier UTxO result, so ReconcileSnapshot restores inputs that are already spent and retires the record; its outputs then reappear with no pacing. That observation pair is exactly what TestWalletRollbackRestoresFundingForNextPayment feeds, so the wallet cannot tell it from a rollback. Reading LocalTxMonitor presence before GetUTxOByAddress closes it: a tx confirmed in between stays reserved for one more round.

internal/test/antithesis/internal/txpump/wallet.go
```diff
@@ -75,10 +99,139 @@ func (w *Wallet) isAvailable(utxo UTxO, now time.Time) bool {
...
+		for _, in := range tx.inputs {
+			if _, ok := chain[utxoKey(in)]; ok {
+				// Keep the authoritative snapshot's value and restore only
+				// signing metadata held by the reservation.
+				current := chain[utxoKey(in)]
+				current.SigningKey = in.SigningKey
+				chain[utxoKey(in)] = current
+			}
+		}
+		delete(w.pending, id)
```
chrisguiney, line 202: No test asserts that an absent transaction's record is retired (PendingIDs empty after the reconcile). Every test still passes with this delete removed, in which case absent records accumulate and each batch sends one HasTx per historical transaction.

internal/test/antithesis/internal/txpump/pump.go
```diff
@@ -399,16 +433,18 @@ func (p *Pump) submitPayment(client *NodeClient, batchSize int) bool {
...
 		}
 		if change > 0 {
 			changeUTxO := UTxO{TxHash: txID, Index: 1, Amount: change}
 			if signingKey != nil {
 				changeUTxO.SigningKey = signingKey
 			}
-			p.wallet.AddAfter(confirmationDelay, changeUTxO)
+			outputs = append(outputs, changeUTxO)
 		}
+		p.wallet.RecordAccepted(txID, inputs, outputs, confirmationDelay)
```
chrisguiney, line 448: Non-blocking: for keyed wallets the rejected branch above no longer holds its comment. A rejected payment input that is still in the node UTxO set is re-added by the next ReconcileSnapshot, so a deterministic rejection now repeats once per batch. Either keep rejected inputs out of reconciliation or update that comment to describe the new behavior.

chrisguiney (changes requested):
Requesting changes for the observation order in ReconcileWallet; details inline. The CodeRabbit change-address thread describes behavior that predates this change: change from submitDelegation, submitGovernance and submitPlutus already went to deterministicAddr, which no wallet key can spend, and reconciliation now drops those outputs instead of offering them to coin selection.

internal/test/antithesis/internal/txpump/client.go
```diff
@@ -107,6 +110,136 @@ func (c *NodeClient) SubmitTx(eraID uint16, txBytes []byte) error {
...
+				txID,
+				err,
+			)
+		}
+		presence[txID] = present
+	}
+	// Observe transaction presence before acquiring the authoritative UTxO
+	// snapshot. A transaction can confirm between the two observations; the
+	// monitor result must describe the state at or before the LSQ snapshot so a
+	// confirmed spend cannot resurrect its source input.
```
chrisguiney, line 191: This invariant does not hold on the node this pump targets. On a Dingo producer (p1), ledger/forging/forger.go removes a forged block's transactions from the mempool (RemoveTxsByHash, right after chain.AddLocalBlock), but the UTxO rows served by queryShelleyUtxoByAddress (ledger/queries.go, not pinned by AcquireVolatileTip) change only when ledgerProcessBlocks applies that block later. Until then HasTx is false while the spent inputs are still returned, in either read order, so ReconcileSnapshot restores them, retires the record and drops output pacing.

internal/test/antithesis/internal/txpump/wallet.go
```diff
@@ -75,10 +99,139 @@ func (w *Wallet) isAvailable(utxo UTxO, now time.Time) bool {
...
+		if presence[id] {
+			for _, in := range tx.inputs {
+				reserved[utxoKey(in)] = struct{}{}
+			}
+			for _, out := range tx.outputs {
+				reserved[utxoKey(out)] = struct{}{}
+			}
+			continue
+		}
+		for _, in := range tx.inputs {
```
chrisguiney, line 193: An absent record whose inputs are still in the snapshot is ambiguous: dropped, or confirmed but not yet applied by the ledger. Keeping those inputs reserved until a later reconcile (for example one slot after the first absent observation) still sees them avoids selecting a spent input and still recovers dropped funding. A wallet test for the first absent observation (inputs present, outputs missing, input not selectable) pins it; TestWalletRollbackRestoresFundingForNextPayment currently encodes immediate restore.

internal/test/antithesis/go.mod
```diff
@@ -6,7 +6,7 @@ toolchain go1.26.7
 
 require (
 	github.com/antithesishq/antithesis-sdk-go v0.7.0
-	github.com/blinklabs-io/gouroboros v0.202.10
+	github.com/blinklabs-io/gouroboros v0.204.5
```
chrisguiney, line 9: v0.204.5 carries no LocalStateQuery acquisition fix: the acquisition-transition change is blinklabs-io/gouroboros#2320, first released in v0.204.7, while v0.204.5 adds blinklabs-io/gouroboros#2311 (chain origin block numbers). The txpump tests also pass unchanged on v0.202.10. Pin the release that carries the fix the description names, or correct the description.

chrisguiney (changes requested):
Prior points at b120e273: restoring LSQ-first order fails TestReconcileWalletObservesMonitorBeforeLSQ; removing the pending delete fails TestWalletRollbackRestoresFundingForNextPayment; a deterministicAddr change or a missing locked-output address fails both change-address tests; the rejected-payment comment matches behavior. go test -race and golangci-lint on internal/txpump pass at head and merged with main; no check at this head runs this nested module. Blocking: the Dingo producer still yields the spent-input observation pair, inline.

internal/test/antithesis/internal/txpump/wallet_test.go
```diff
@@ -48,6 +48,228 @@ func TestWallet_AddAfterQuarantinesSubmittedOutputs(t *testing.T) {
...
+
+func TestWalletRollbackRestoresFundingForNextPayment(t *testing.T) {
+	w := NewWallet()
+	source := makeUTxO("funding", 0, 5_000_000)
+	w.Add(source)
+	inputs, _, err := w.SelectCoins(1_000_000)
+	require.NoError(t, err)
+	w.Reserve("tx", inputs, []UTxO{makeUTxO("tx", 0, 4_000_000)}, 0)
+	w.ReconcileSnapshot([]UTxO{source}, map[string]bool{"tx": false})
+	require.Empty(t, w.PendingIDs(), "confirmed or rejected records must be retired")
```
chrisguiney: This assertion and the new deferral in ReconcileSnapshot contradict each other: go test ./internal/txpump/ reports Should be empty, but was [tx] at this line, both at a50ec4b0 and merged with current main. b120e273 added this line, while 356d3ab6 branched from 44cdfaad, before it, and made the record survive one absent observation; merge a50ec4b0 combined both with no textual conflict. Decide which behavior is intended and make the test state it rather than dropping the assertion. No configured check catches this, because go test ./... at the repository root skips this nested module.

internal/test/antithesis/internal/txpump/wallet.go
```diff
@@ -75,10 +102,148 @@ func (w *Wallet) isAvailable(utxo UTxO, now time.Time) bool {
...
+			if _, ok := chain[utxoKey(in)]; ok {
+				// Keep the authoritative snapshot's value and restore only
+				// signing metadata held by the reservation.
+				current := chain[utxoKey(in)]
+				current.SigningKey = in.SigningKey
+				chain[utxoKey(in)] = current
+			}
+		}
+		if !tx.absent {
```
chrisguiney, line 207: This defers retiring the record but not the reservation, so the blocker raised at b120e273 still stands. The loop just above leaves every input still present in the snapshot inside chain, and only the presence branch adds keys to reserved, so nothing removes them before w.utxos is rebuilt. Measured at a50ec4b0: after Reserve and then ReconcileSnapshot with the input in the snapshot and presence false, SelectCoins hands back funding:0, the input the node has already consumed. Adding those inputs to reserved for the grace round closes it and keeps the output pacing this deferral recovers.

internal/test/antithesis/internal/txpump/wallet.go
```diff
@@ -75,10 +102,148 @@ func (w *Wallet) isAvailable(utxo UTxO, now time.Time) bool {
...
+			// A node can remove a forged transaction from the mempool before
+			// its committed outputs and spent inputs are visible to LSQ. Keep
+			// the record through one absent observation so the next reconcile
+			// can distinguish that race from a real rollback.
+			tx.absent = true
```
chrisguiney, line 212: tx.absent is never cleared. The presence branch above continues without resetting it, so a transaction that is absent once, returns to the mempool, and is absent again is retired on that second observation with no grace round at all. Measured at a50ec4b0 across three reconciles with presence false, true, false: PendingIDs reports [tx], then [tx], then empty. Clear the flag whenever presence is observed, so the grace round covers each absence rather than only the first one in a record lifetime.

internal/test/antithesis/internal/txpump/client.go
chrisguiney: This comment still states that observing the monitor first means a confirmed spend cannot resurrect its source input, while the comment added at wallet.go:206 states the opposite, that a node can drop a forged transaction from the mempool before the spend is visible to LSQ. The latter matches a Dingo producer: ledger/forging/forger.go removes forged transactions right after AddLocalBlock, and queryShelleyUtxoByAddress keeps serving the old rows until the block is applied. Two comments in one package now disagree; the ordering narrows this window rather than closing it.

internal/test/antithesis/go.mod
```diff
@@ -6,8 +6,8 @@ toolchain go1.26.7
 
 require (
 	github.com/antithesishq/antithesis-sdk-go v0.7.0
-	github.com/blinklabs-io/gouroboros v0.202.10
-	github.com/blinklabs-io/plutigo v0.6.1
+	github.com/blinklabs-io/gouroboros v0.204.7
```
chrisguiney, line 9: v0.204.7 does carry blinklabs-io/gouroboros#2320, complete acquisition transitions, so the earlier pin point is resolved. The description still says the harness moves to v0.204.5; update it to match this line. The plutigo move to v0.7.0 is required by gouroboros v0.204.7 rather than added scope, per go mod graph.

chrisguiney (changes requested):
CHANGES_REQUESTED at a50ec4b0. The head fails its own tests: go test ./internal/txpump/ in the nested module internal/test/antithesis fails TestWalletRollbackRestoresFundingForNextPayment, both at head and merged with current main. CI misses it because go test ./... at the repository root skips nested modules. No bot reviewed this head: CodeRabbit reports incremental reviews disabled and its last review is at f4541c12, and cubic has no check run or status context here. Both CodeRabbit inline points are addressed in direction; the client.go one survives as a blocker inline.

internal/test/antithesis/internal/txpump/client.go
```diff
@@ -107,6 +110,132 @@ func (c *NodeClient) SubmitTx(eraID uint16, txBytes []byte) error {
...
+				c.addr,
+				txID,
+				err,
+			)
+		}
+		presence[txID] = present
+	}
+	return snapshot, presence, nil
+}
+
```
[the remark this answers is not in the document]
wolf31o2: Mutually consistent snapshots are not reachable across LocalTxMonitor and LocalStateQuery on this node: a Dingo producer drops a forged transaction in ledger/forging/forger.go before queryShelleyUtxoByAddress stops serving its inputs, so no acquisition order closes the window. The deferral half is implemented. One absent observation keeps the record and its still-present inputs reserved; only a second consecutive absence retires the record and restores inputs. Removing that reservation fails TestWalletRollbackRestoresFundingForNextPayment.

internal/test/antithesis/internal/txpump/pump.go
```diff
@@ -484,11 +520,11 @@ func (p *Pump) submitDelegation(client *NodeClient, batchSize int) bool {
...
 		entry.Status = "submitted"
 		p.logger.Info("delegation tx submitted", "tx_id", txID)
-		// Return the change output to the wallet so future transactions can
-		// spend it.
+		var outputs []UTxO
 		if change > 0 {
-			p.wallet.AddAfter(p.cfg.confirmationDelay(), UTxO{TxHash: txID, Index: 0, Amount: change})
+			outputs = []UTxO{{TxHash: txID, Index: 0, Amount: change}}
 		}
+		p.wallet.RecordAccepted(txID, inputs, outputs, p.cfg.confirmationDelay())
```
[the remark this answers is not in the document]
wolf31o2: Fixed. controlledChangeAddr derives change from the selected input's signing key at all three sites, delegation, governance and both Plutus kinds, and submitPayment keeps signingKey.Address. Reverting the sites to deterministicAddr fails TestWorkloadSubmissionsKeepControlledChangeAddress and TestPlutusUnlockReturnsChangeToLockedWalletAddress.

internal/test/antithesis/internal/txpump/client.go
wolf31o2: Fixed. HasTx now runs before AcquireVolatileTip. Moving the monitor block after GetUTxOByAddress fails TestReconcileWalletObservesMonitorBeforeLSQ, which then selects the funding hash instead of the parent output. The ordering narrows the window rather than closing it, so reconciliation also keeps an ambiguous absence reserved for one round.

internal/test/antithesis/internal/txpump/wallet.go
wolf31o2: Fixed. TestWalletRollbackRestoresFundingForNextPayment and TestWalletAbsenceGraceResetsAfterMempoolReappearance both fail when delete(w.pending, id) is removed, so retirement is pinned. The same class left one more gap: controlledChangeAddr's keyless fallback was unasserted, and TestUnsignedWorkloadSubmissionsKeepDeterministicChangeAddress at 33d5dc10 fails when that fallback returns nil.

internal/test/antithesis/internal/txpump/pump.go
wolf31o2: Comment updated to describe reconciliation restoring still-unspent inputs, and TestWallet_ReconcileRestoresFundingOutput pins that a selected but unreserved input returns from the snapshot. A rejected payment is never entered in the pending map, so a deterministic rejection recurs at most once per batch rather than in the tight retry loop the old comment described.

internal/test/antithesis/internal/txpump/client.go
wolf31o2: Accepted; the ordering does not close the window. ReconcileSnapshot treats one absent observation as ambiguous, keeping the record and its still-present inputs reserved for a further round, and only a second consecutive absence retires and restores. The client comment no longer claims that a confirmed spend cannot resurrect its source input.

internal/test/antithesis/internal/txpump/wallet.go
wolf31o2: Implemented as a grace of one reconcile round. After the first absent observation PendingIDs still reports the record and SelectCoins returns ErrInsufficientFunds for its inputs; the second retires the record and restores only inputs the snapshot still carries. TestWalletRollbackRestoresFundingForNextPayment asserts both halves and fails without the grace reservation.

internal/test/antithesis/go.mod
wolf31o2: Fixed. The pin is gouroboros v0.204.7, which carries blinklabs-io/gouroboros#2320, and the description names v0.204.7.

internal/test/antithesis/internal/txpump/wallet_test.go
wolf31o2: Fixed. The test now states the deferral: one pending record and an unselectable input after the first absent observation, then empty PendingIDs and restored funding after the second. go test ./internal/txpump/ passes at 33d5dc10 and on the tree merged with current origin/main.

internal/test/antithesis/internal/txpump/wallet.go
wolf31o2: Fixed at 4bea1cdd. The grace branch adds the record's inputs to reserved, so they are dropped from the rebuilt UTxO set instead of being handed back by SelectCoins. Removing that loop fails TestWalletRollbackRestoresFundingForNextPayment and TestWalletAbsenceGraceResetsAfterMempoolReappearance.

internal/test/antithesis/internal/txpump/wallet.go
wolf31o2: Fixed at 4bea1cdd and e9e7513. The presence branch sets tx.absent to false, so every present-to-absent transition earns its own grace round. TestWalletAbsenceGraceResetsAfterMempoolReappearance drives presence false, true, false and fails without that reset.

internal/test/antithesis/internal/txpump/client.go
wolf31o2: Fixed. The client comment now states that the two observations are independent and that a forged transaction can leave the mempool before the spend reaches LSQ, which matches wallet.go. ReconcileSnapshot's own comment still stated the pre-grace contract and is corrected at 1f4d9e0e.

internal/test/antithesis/go.mod
wolf31o2: Fixed. The description names v0.204.7. plutigo v0.7.0 is the requirement of gouroboros v0.204.7 per go mod graph rather than added scope.

chrisguiney (approved):

PR #4294, opened 2026-09-14
fix(ledgerstate): decode committee hot keys from authorization sum type

wcatz:
## Problem

A Mithril-bootstrapped mainnet node imports `auth_committee_hot` with **zero rows** while `committee_member` holds 13. Every constitutional committee vote then fails the Conway unknown-voter rule (`UtxoValidateUnknownVoters`), and the node stops following the chain.

Observed on a mainnet block producer: it halted on canonical block `13940276` at slot `197832946` with

```
conway utxo validation rule 52: vote cast by unknown voter: type=1 hash=2ea7a78eb914d988
```

`type=1` is `constitutional_committee_hot_scripthash`, not a DRep. The node then rewound 2160 slots, resynced, hit the same block, and looped indefinitely.

## Cause

The committee map's values are the `CommitteeAuthorization` sum type:

```
[0, hot_credential]   CommitteeHotCredential
[1, maybe_anchor]     CommitteeMemberResigned
```

`parseCommitteeHotCredential` accepted only a bare credential or a **one-element** wrapper:

```go
wrapped, err := decodeRawArray(data)
if err != nil || len(wrapped) != 1 {
	return Credential{}, errors.New("decoding committee hot credential")
}
```

Every real entry is a two-element array, so all of them failed and were dropped by `continue` in `parseCommitteeVState`. That function had **no error return**, so the caller received an empty committee with no warning — a committee that could not be parsed was indistinguishable from one that had authorized no hot keys. The import completed "successfully" and the damage only surfaced hours later as a chain halt.

Dumping the real mainnet ledger state at slot `197789347` confirms the map is present and well-formed — 7 entries, all `[0, [credType, hash]]`:

```
{[1, h'16feefc2...']: [0, [1, h'84feba94...']],
 [1, h'84aebcfd...']: [0, [1, h'2ea7a78eb914d988b9d368ed88906f3bc9fc5421667dea6a366710ec']],
 ...}
```

The credential the node rejected was in the snapshot the whole time.

## Change

- Decode the `CommitteeAuthorization` sum type; bare credentials and one-element wrappers still parse, so older encodings are unaffected.
- Record `CommitteeMemberResigned` as a resignation instead of discarding it.
- Return an error when a committee map has entries but none decode. An authoritatively empty committee (no entries) still imports normally.

## Verification

Against the real mainnet ledger state at slot `197789347`, through `ledgerstate.ParseSnapshot` + `ParseCertState`:

| | before | after |
|---|---|---|
| Committee hot keys | 0 | **7** |
| Committee resignations | 0 | 0 |
| DReps | 1054 | 1054 |

All 7 authorizations are recovered, including `2ea7a78e...`, and the DRep count is unchanged.

Two tests added:

- `TestParseCommitteeVStateAuthorizationSumType` — the encoding mainnet actually uses, covering both an authorization and a resignation, asserting credential tags survive.
- `TestParseCommitteeVStateFailsLoudOnUndecodableEntries` — a non-empty committee map whose entries cannot be decoded must be an error, not a silent empty result.

Both fail without the change. Existing committee tests pass unmodified.

`go test ./ledgerstate/... ./ledger/... ./database/...` passes.

## Note on `CommitteeStateAvailable`

`LedgerView.CommitteeStateAvailable` (`ledger/view.go`) reports availability from the **cold** member count, so with 13 members and 0 authorizations it answers `true` and gouroboros proceeds to reject the vote rather than treating committee state as unavailable. That is arguably a second line of defence worth revisiting alongside #3785, but it is not the bug: with this change the authorizations are present and the question does not arise. Left out of scope here deliberately.

## Docs

`DATABASE.md` and `ARCHITECTURE.md` checked — no schema, storage-layout, or component-boundary changes; `auth_committee_hot` is populated correctly by an existing code path rather than altered.

---
## Summary by cubic
Fixes Conway committee imports for snapshots whose map values use `CommitteeAuthorization`, preventing valid votes from failing after hot keys were silently dropped. It also prevents Mithril bootstrap from panicking when synthetic certificate transactions have nil fees by recording those fees as zero.

**Bug Fixes**

- Decodes authorization and resignation variants, preserves key/script credential tags, and keeps older bare-credential encodings compatible.
- Rejects malformed resignation anchors by accepting only the ledger's StrictMaybe encodings and fails imports on any undecodable committee entry, while allowing genuinely empty maps.
- Applies nil-fee handling at all metadata transaction write sites without changing non-nil fees.
- Adds coverage for sum-type parsing, malformed committee data, anchor validation, and fee persistence through the certificate write path.
- Removes the unused batch accumulator from the metadata sqlstore.

<sup>Written for commit e9a7151c21e543eec5b60fded079cb4f4d6ba852. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved committee-state parsing to correctly distinguish hot-key authorizations from resignations.
  * Added compatibility for multiple committee authorization encodings.
  * Prevented invalid or undecodable committee data from being silently treated as an empty committee.

* **Tests**
  * Added coverage for authorization and resignation parsing, including preserved entry types.
  * Added validation for errors caused by undecodable committee entries.


wcatz:
Good catch — `[1, <uint>]` parsed as a resignation, and since `looksLikeCommitteeCredentialMap` shares that decoder it could have let an unrelated credential-keyed map qualify as the committee map in the Conway element scan. The payload is a `StrictMaybe Anchor`, so it now requires null or an array and rejects scalars.

Pushed in 996702cf with a near-miss test covering `[1, uint]` rejection, `[1, null]` still parsing, and the map-level misidentification case. Re-verified against the mainnet ledger state at slot 197789347: still 7 authorizations recovered, DReps unchanged at 1054.

ledgerstate/certstate.go
```diff
@@ -1631,18 +1671,85 @@ func parseCommitteeVState(
...
+// isResignationPayload reports whether data is a plausible StrictMaybe Anchor:
+// CBOR null, or an array (empty for absent, otherwise the anchor itself).
+func isResignationPayload(data []byte) bool {
+	if len(data) == 0 {
+		return false
+	}
+	if data[0] == 0xf6 { // null
+		return true
+	}
+	if _, err := decodeRawArray(data); err == nil {
```
wolf31o2: This accepts every CBOR array as a valid Anchor payload. A malformed committee entry such as `[1, [1]]` (or an array with the wrong Anchor fields/types) therefore returns `resigned=true` and is imported instead of reaching the undecodable-entry error path. Validate the null/Anchor forms and the Anchor's `[url, 32-byte hash]` shape, and add a malformed-array test, so this decoder remains fail-closed.

wolf31o2 (changes requested):

ledgerstate/certstate.go
wcatz: Fixed in 5492f97a. `isResignationPayload` now accepts only the StrictMaybe Anchor forms — null or an empty array for an absent anchor, a one-element array wrapping an anchor, or the anchor itself — and an anchor must be `[url, 32-byte hash]`, the same shape and hash length `parseConstitution` already enforces. `[1, [1]]` now falls through to the undecodable-entry error instead of importing as a resignation.

Two tests cover it: the authorization-level cases (array-of-a-scalar, 31-byte hash, three-field anchor, url and hash swapped) and the map-level case where every value is malformed, which must fail the import loudly and must not satisfy `looksLikeCommitteeCredentialMap`. Both fail against the previous validator — I reverted it to confirm rather than assume.

wcatz:
Heads up — running this against real mainnet found a second defect that has to ship with the first, and it changes the risk of merging the committee fix alone.

Once the committee actually decodes, `persistImportedCommitteeCertificates` runs for the first time, and it panicked: `TransactionBodyBase.Fee()` returns nil, the synthetic transaction carrying the imported certificates doesn't override it, and `setTransaction` calls `transaction.Fee().Uint64()` unguarded. A mainnet Mithril bootstrap SIGSEGV'd right after importing 2895 pools and 1054 DReps. That path was unreachable while the committee always imported empty — so **merging the committee decode without `963f12e9` turns a silent chain halt into a panic crash-loop for every Mithril-bootstrapped node.** `ledger/eras` already nil-guards `Fee()` in two places; the write path just missed it, so the guard now matches that existing idiom and covers both write sites.

Bootstrap is re-running on the fixed binary now; I'll post the `auth_committee_hot` count and the node crossing slot 197832946 when it gets there.

wcatz:
Mainnet bootstrap result on this branch: `auth_committee_hot` imported **7 rows**, not 0, with credential tags preserved (5 script, 2 key). Row 4 is `84aebcfd…392c6a` → `2ea7a78eb914d988b9d368ed88906f3bc9fc5421667dea6a366710ec` — the exact hot credential the node rejected as an unknown voter, byte-identical to the CBOR in the snapshot. The cert-state import ran to completion with no panic on `963f12e9` and checkpointed (`phase=certstate`); accounts 1467966, pools 2895, DReps 1054 all imported clean. Block copy is still running, and I'll confirm the node passing slot 197832946 once it's serving.

wcatz:
Precision on the mainnet evidence above: the binary was built on a work tree that also carries an unrelated gouroboros re-pin, not from a clean checkout of this branch. The four changed files (`ledgerstate/certstate.go`, `ledgerstate/certstate_test.go`, `database/plugin/metadata/sqlstore/transaction_write.go`, `database/plugin/metadata/sqlstore/transaction_fee_test.go`) are byte-identical to the commits here — verified by sha256 — so the behaviour demonstrated is this branch's, but I'd flag the differing gouroboros pin since `TransactionBodyBase.Fee()` returning nil is upstream behaviour.

ledgerstate/certstate.go
```diff
@@ -1600,18 +1619,39 @@ func parseCommitteeVState(
...
 				ParsedCommitteeHotKey{Cold: cold, Hot: hot},
 			)
 		}
+		// A committee map with entries that yields nothing means the
+		// encoding was not understood. Returning an empty set here is
+		// indistinguishable from a genuinely unauthorized committee, and
+		// downstream that makes every committee vote fail the Conway
+		// unknown-voter rule on a node that has no way to notice. Fail the
+		// import instead.
+		if len(entries) > 0 &&
```
chrisguiney: This fails closed only when every entry fails. An undecodable entry beside a decodable one still takes the silent `continue` at L1625-1626: the map `{cold1: [0, [1, h]], cold2: [1, [1]]}` returns 1 hot key and a nil error from both parseCommitteeVState and parseVState, and `[2, ...]` or `[0, [5, h]]` values are dropped the same way. A dropped authorization is the same unknown-voter rejection for that member. Return an error when any entry's cold key or authorization fails to decode, and add a mixed-map test.

ledgerstate/certstate.go
```diff
@@ -1631,18 +1671,115 @@ func parseCommitteeVState(
...
+// isResignationPayload reports whether data is a StrictMaybe Anchor. The
+// accepted forms are CBOR null or an empty array for an absent anchor, a
+// one-element array wrapping an anchor, and the anchor itself. Every other
+// shape is rejected, so a malformed entry such as [1, [1]] reaches the
+// undecodable-entry error path instead of importing as a resignation.
+func isResignationPayload(data []byte) bool {
+	if len(data) == 0 {
+		return false
+	}
+	if data[0] == 0xf6 { // null: absent anchor
```
chrisguiney: CommitteeMemberResigned encodes its StrictMaybe Anchor with cardano-ledger-binary `encodeStrictMaybe` (`[]` or `[anchor]`), and `decodeStrictMaybe` rejects null and any list length other than 0 or 1. Accepting null and a bare `[url, hash]` imports as resignations two shapes the node never writes and its own decoder refuses. Accept only `[]` and `[anchor]`, move null and the bare anchor to the rejected cases, and use `[1, []]` in TestParseCommitteeVStateAuthorizationSumType, whose `[1, null]` fixture (test L763-764) is not the on-chain encoding.

database/plugin/metadata/sqlstore/transaction_fee_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"math/big"
+	"testing"
+
+	lcommon "github.com/blinklabs-io/gouroboros/ledger/common"
+	"github.com/stretchr/testify/require"
+)
+
+// feelessTransaction mirrors the synthetic transactions used to carry imported
+// certificates into the metadata store: it embeds TransactionBodyBase and does
+// not override Fee, so Fee returns nil.
```
chrisguiney: These tests cover the helper, not the write path that panicked: restoring `transaction.Fee().Uint64()` at the setTransaction call site keeps both tests here and database TestSetTransactionMetadataOnly* green, and no test calls persistImportedCommitteeCertificates. A ledgerstate test running it against dbtest.NewDatabase with one hot key and one resignation passes at this head and panics with a nil dereference when either the guard or that call site is reverted. Also, feelessTransaction does override Fee, contrary to this comment.

ledgerstate/certstate_test.go
```diff
@@ -726,3 +732,254 @@ func TestParseCertStateConwayCommitteeSurvivesSmallDState(t *testing.T) {
...
+
+	good32 := anchorCBOR(t, 32)
+
+	valid := map[string][]byte{}
+	for name, payload := range map[string][]byte{
+		"null":        {0xf6},
+		"empty array": {0x80},
+		"bare anchor": good32,
+	} {
+		valid[name] = payload
```
chrisguiney: Style, not blocking on its own: `make lint` runs modernize over ledgerstate including tests and exits 3 on this loop (`certstate_test.go:901:3: Replace m[k]=v loop with maps.Copy`); the same packages are clean at the merge base. Build `valid` as a single map literal.

chrisguiney (changes requested):
Requesting changes: a committee map with some undecodable entries still drops them silently, the resignation payload accepts shapes outside the ledger's StrictMaybe codec, and the nil-fee crash path has no test that fails without the fix. Reproductions inline.

wcatz:
@chrisguiney all four are addressed in `65621ab1`, `89818496`, `9df9dd00`.

`parseCommitteeVState` now returns an error on any undecodable cold credential or authorization rather than skipping the entry — you were right that a partial committee is the same unknown-voter rejection this parser exists to prevent, just quieter. That could not simply be made fatal in place: the Conway heuristic scan tries every unclaimed `CertState` element as a committee candidate and treats an error as fatal, so failing closed inside the parser aborted imports over elements that were never the committee state. The candidacy test moved out into the scan (`committeeMapElement` + `looksLikeCommitteeCredentialMap`), so a wrong candidate is skipped before it reaches a parser that now fails closed. `isResignationPayload` accepts only `[]` and `[anchor]`, matching `encodeStrictMaybe`/`decodeStrictMaybe`; null and a bare anchor are rejected, and the fixtures that encoded `[1, null]` were wrong — note `decodeRawArray` accepts CBOR null, so the major type is checked first or `SNothing` lets null through. The certificate codec is genuinely different here (`ResignCommitteeColdCertificate` has a nullable `*GovAnchor`); it's the ledger *state* that uses StrictMaybe.

On the nil-fee test: you were right that the accessor test stays green if the guard is reverted. `ledgerstate/imported_committee_certificates_test.go` drives `persistImportedCommitteeCertificates` end to end and I confirmed it panics with a nil pointer dereference when the guard is reverted at both call sites, then restored it; the assertion reads back through `GetCommitteeMember`, the same `auth_committee_hot` lookup rule 52 uses. The two overstated comments in the accessor test are corrected, and the modernize finding is fixed. Each commit passes `go test ./ledgerstate/... ./database/plugin/metadata/sqlstore/...` on its own.

wolf31o2 (dismissed):

wcatz:
Confirmed on mainnet. A block producer bootstrapped from the Mithril snapshot (epoch 655, anchor slot 197789347) and running `main` plus these commits replayed straight through block 13940276 at slot 197832946 — the exact block where it previously halted with `conway utxo validation rule 52: vote cast by unknown voter: type=1 hash=2ea7a78eb914d988` — and is now 1,554 blocks past it at block 13941830 / slot 197864679, with zero unknown-voter rejections, zero rollbacks, and no errors other than the expected mempool `utxo not found` rejections while still behind tip.

ledgerstate/certstate.go
```diff
@@ -222,7 +222,28 @@ func parseCertStateConway(
 		if len(elem) == 0 || i == pIdx || i == dIdx || i == drepIdx {
 			continue
 		}
-		hotKeys, resignations := parseCommitteeVState(certState[i:])
+		// Only a credential-to-authorization map can be the committee map.
+		// Testing that first keeps a wrong-candidate element from reaching a
+		// parser that now fails closed, which would abort the whole import
+		// over an element that was never the committee state.
+		if !looksLikeCommitteeCredentialMap(
```
chrisguiney, line 229: looksLikeCommitteeCredentialMap reads only the first map entry (cbor_decode.go:139), so failing closed depends on map order. Same two entries: good entry first makes parseCertStateConway return an error; bad entry first makes the gate false, the element takes this continue, falls through to the DState scan, and persistImportedCommitteeCertificates returns early at import.go:3566 - empty committee, no error, the unknown-voter rejection this change exists to prevent. parseVState calls parseCommitteeVState ungated, so the two layouts disagree. Gate every entry, or scope the error instead.

database/plugin/metadata/sqlstore/transaction_write.go
```diff
@@ -202,7 +215,7 @@ RETURNING id`,
 				metadataValue,
 				point.Slot,
 				transaction.Type(),
-				decimalUint64(types.Uint64(transaction.Fee().Uint64())),
+				decimalUint64(transactionFee(transaction)),
```
chrisguiney: The branch conflicts with main here, and the conflict is semantic. main's batched insert replaced the two guarded call sites with three unguarded transaction.Fee().Uint64() calls, so resolving in main's favour restores the nil-fee panic. Keeping transactionFee at all three sites builds and passes ledgerstate and sqlstore; reverting the guard there panics TestPersistImportedCommitteeCertificatesWritesRows with a nil dereference. Please merge main and reapply the guard at the new call sites.

ledgerstate/certstate.go
```diff
@@ -1621,7 +1688,7 @@ func parseCommitteeVState(
 				resignations = append(resignations, cold)
 			}
 		}
-		return hotKeys, resignations
+		return hotKeys, resignations, nil
```
chrisguiney, line 1691: The resignation field keeps the silent drop this change removed from the authorization loop: this map form and the array form at 1696 both append only when coldErr is nil and discard the entry otherwise. A dropped resignation is the mirror of a dropped authorization - the member stays recorded as authorized and its hot key keeps voting after the ledger retired it. The same fail-closed treatment as the loop above would make the function consistent.

chrisguiney (changes requested):
Two of the four earlier findings are cleared: the resignation payload now accepts only the ledger's StrictMaybe forms, and the new write-path test panics with a nil dereference when the fee guard is reverted in place and passes when it is restored. The fail-closed change is still decided by map order, and the branch conflicts with main in transaction_write.go where the batched insert reintroduced the unguarded fee call. Details inline.

ledgerstate/certstate.go
chrisguiney, line 229: looksLikeCommitteeCredentialMap reads only the first map entry (cbor_decode.go:139), so this gate turns on encoded entry order. Same two-entry committee map, good entry first: gate true, parseCommitteeVState errors, parseCertStateConway returns nil and an error. Bad entry first: gate false, the element takes this continue, nothing later matches, and the result carries 0 committee hot keys and no error - the silent empty committee this change exists to prevent. parseVState is ungated and errors either way, so the two layouts disagree. Gate every entry, or scope the error.

ledgerstate/certstate.go
```diff
@@ -1611,7 +1678,7 @@ func parseCommitteeVState(
 		}
 	}
 	if len(committeeFields) < 2 {
-		return hotKeys, nil
+		return hotKeys, resignations, nil
 	}
 	resignationEntries, resignationErr := decodeMapEntries(committeeFields[1])
```
chrisguiney, line 1683: The resignation field keeps the silent drop: this map loop and the array loop at 1693 append only when the cold parse succeeds. A two-entry resignation map with one undecodable key returns 1 resignation and a nil error, and the array form behaves the same; a committeeFields[0] that is not a decodable map also returns no data and no error. A dropped resignation is the mirror of a dropped authorization - the member stays recorded as authorized and its hot key keeps voting after the ledger retired it.

database/plugin/metadata/sqlstore/transaction_write.go
```diff
@@ -27,6 +27,21 @@ import (
...
+// comparing a fee against the computed minimum.
+func transactionFee(transaction lcommon.Transaction) types.Uint64 {
+	fee := transaction.Fee()
+	if fee == nil {
+		return 0
+	}
+	return types.Uint64(fee.Uint64())
+}
+
+type immediateBatchAccumulator struct{}
```
chrisguiney: Dead declaration left by the merge: the pre-merge branch carried this type with a Reset method and three type assertions, main deleted all of it, and the resolution kept only this line beside the transactionFee hunk. It has no references and no method set at head. This is the red lint check - CI job 105047314462 fails on transaction_write.go:43:6, type immediateBatchAccumulator is unused. Removing the line makes golangci-lint report 0 issues over ledgerstate and sqlstore and the build still succeeds; origin/main has no such symbol.

database/plugin/metadata/sqlstore/transaction_write.go
```diff
@@ -313,7 +328,7 @@ func (s *Store) setTransactionWithAccumulator(
 					metadataValue,
 					point.Slot,
 					transaction.Type(),
-					decimalUint64(types.Uint64(transaction.Fee().Uint64())),
+					decimalUint64(transactionFee(transaction)),
```
chrisguiney, line 329: Cleared and re-verified here: the merge reapplied the guard at all three sites, including the two the batched insert introduced. Reverting only those two in place fails TestPersistImportedCommitteeCertificatesWritesRows with a nil dereference, and restoring passes. Merged against current main the build succeeds and ledgerstate, sqlstore and ouroboros pass, including under -race, and the conformance suite passes. DATABASE.md and ARCHITECTURE.md were checked and are unaffected: no schema, query or storage-layout change.

ledgerstate/certstate.go
```diff
@@ -1631,18 +1698,114 @@ func parseCommitteeVState(
...
+}
+
+// isResignationPayload reports whether data is a StrictMaybe Anchor as the
+// ledger encodes it. encodeStrictMaybe writes an empty array for SNothing and a
+// one-element array wrapping the value for SJust, and decodeStrictMaybe rejects
+// every other shape -- including CBOR null and a bare anchor. Accepting those
+// would import as resignations two encodings the node never writes and its own
+// decoder refuses, so they are rejected here and reach the undecodable-entry
+// error path instead.
+func isResignationPayload(data []byte) bool {
```
chrisguiney, line 1792: Cleared: the resignation payload now accepts only the StrictMaybe forms, an empty array or a one-element array wrapping a [url, 32-byte hash] anchor, and rejects null and a bare anchor. That closes the CodeRabbit finding at 42986f11 and the wolf31o2 finding on this file, both of which read a malformed array as a resignation; both are addressed at this head. Neither bot reviewed this head: the status context reports incremental reviews disabled, and cubic posted no check, status or review record.

chrisguiney (dismissed):
Requesting changes: the new candidacy gate reads only the first encoded committee entry, so the same two entries in the other byte order import an empty committee silently instead of aborting, while parseVState stays ungated and errors either way. The merge also left a dead immediateBatchAccumulator declaration, which is the red lint check. The resignation field still drops undecodable entries silently. The nil-fee guard and the StrictMaybe payload are cleared, and the guard is re-verified against the batched call sites the merge introduced. Details inline.

wolf31o2 (approved):

PR #4297, opened 2026-09-15
perf(sqlstore): dedupe reward live stake refresh per transaction

chrisguiney:
Depends on #4270 and #4271 -- not mergeable until both land first. The only new content on this branch is the final commit, `perf(sqlstore): dedupe reward live stake refresh per transaction`.

## Problem

Live CPU profiling of a synced node found `sumCredentialUtxoStake` (the full live-UTxO aggregate scan per stake credential behind `refreshRewardLiveStakeAggregate`) at ~30% cumulative CPU -- still the single largest hotspot even after #4270/#4271's SQL-aggregate rewrite and prepared-statement caching. A 20s profile against the untouched vanilla baseline showed the same call chain (via the older row-streaming path) at 61% cumulative CPU, confirming this is a pre-existing, structural bottleneck, not something introduced by this branch's own changes.

Root cause: in `setTransaction`, a transaction's certificate-named stake credentials were refreshed immediately after `applyTransactionCertificates`, then its consumed/produced-UTxO credentials were refreshed again afterward -- with no dedup between those two passes, nor between produced and consumed refs. A stake credential appearing in more than one of these (a wallet's own change output alongside a certificate it just submitted; several outputs to the same address) triggered the expensive full-scan once per occurrence instead of once, even though all mutations were already committed before any refresh ran.

## Change

Added `mergeStakeCredentialRefs` (`utxo.go`) and used it to merge certificate + consumed + produced refs into one deduped set before a single `refreshRewardLiveStakeRefs` call. Purely a call-count reduction -- every recompute still reads fresh, fully-mutated state, so results are identical to before.

## Tests

`TestSetTransactionRefreshesSharedCredentialOnce`: a transaction whose certificate, consumed input, and produced output all name the same stake credential triggers **3** `sumCredentialUtxoStake` calls against unfixed code, **1** fixed; also asserts the final `reward_live_stake` row is correct. `TestMergeStakeCredentialRefsDedupes` covers the merge helper directly.

## Benchmark

`BenchmarkRefreshRewardLiveStakeAggregateRepeatedTouch`, refreshing once vs. twice on credentials with 100/1,000/7,000 live UTxOs:
```
n=100:    ~1.9x faster once merged
n=1,000:  ~2.0x faster once merged
n=7,000:  2.52ms -> 1.35ms (~2.1x faster once merged)
```


database/plugin/metadata/sqlstore/transaction_write.go
```diff
@@ -409,8 +407,32 @@ FROM utxo WHERE tx_id = ? AND output_idx = ?`,
...
+			// refreshRewardLiveStakeAggregate's sumCredentialUtxoStake scan
+			// once per occurrence instead of once per credential. Every one
+			// of this transaction's mutations is already applied to db by
+			// this point, so merging changes nothing about the refreshed
+			// values -- it only removes the repeat scans (see
+			// TestSetTransactionRefreshesSharedCredentialOnce).
+			return s.refreshRewardLiveStakeRefs(
+				ctx,
+				db,
+				mergeStakeCredentialRefs(
```
wolf31o2, line 555: mergeStakeCredentialRefs is applied only here. SetGapBlockTransaction (transaction_write.go:533) still appends certificateRefs and then one ref per produced output, and SetGenesisTransaction (transaction_write.go:602) appends one ref per output, so a stake credential named more than once in either still runs sumCredentialUtxoStake once per occurrence. The rest of the class is already clean: every refreshRewardLiveStakeRefs site in utxo.go sources its refs from queryUtxoStakeRefs or queryStakeRefsDeduped, which dedupe by MapKey.

database/plugin/metadata/sqlstore/transaction_write_livestake_dedup_test.go
```diff
@@ -0,0 +1,301 @@
...
+	seedConsumedUtxo(t, store, fx)
+
+	before := store.sumCredentialUtxoStakeCalls.Load()
+	require.NoError(t, store.SetTransaction(
+		fx.tx,
+		fx.point,
+		0,
+		fx.certDeposits,
+		false,
+		nil,
```
wolf31o2, line 173: Every assertion in this file runs against a nil txn, so withWriteTransaction opens and commits one transaction per SetTransaction call. Production always passes txn.Metadata(), a block-scoped *sql.Tx shared by every transaction in the block. Driving two SetTransaction calls through one outer txn would cover the absence case the per-transaction claim rests on, that merged refs do not carry into the next transaction, and the block-scoped *sql.Stmt retention stmtForQueryer creates there (cause owned by #4271).

wolf31o2 (changes requested):
go-test fails on all four platforms at this head with a build error: ouroboros/localstatequery_test.go:204 calls localstatequeryServerConnOpts with no arguments. That file is not on this branch; it arrived on main in #4232 and #4301 added the trusted argument. This branch sits 13 commits behind that fix, so the failure belongs to the base, not to the dedupe. Merging current origin/main gives a clean go build ./... and go vet ./ouroboros/... ./database/plugin/metadata/sqlstore/... run.

database/plugin/metadata/sqlstore/transaction_write.go
chrisguiney: Fixed in b82fd516: SetGapBlockTransaction now collects certificateRefs and producedStakeRefs separately and merges them with mergeStakeCredentialRefs before refreshing. SetGenesisTransaction now merges its per-output refs the same way. Added TestSetGapBlockTransactionRefreshesSharedCredentialOnce and TestSetGenesisTransactionRefreshesSharedCredentialOnce, both asserting exactly one sumCredentialUtxoStake call; both fail with the fix reverted (2 calls).

database/plugin/metadata/sqlstore/transaction_write_livestake_dedup_test.go
chrisguiney: Fixed in b82fd516: added TestSetTransactionSharedOuterTxnKeepsCredentialsIndependent, which opens one store.Transaction and drives two SetTransaction calls through it (matching txn.Metadata()'s block-scoped *sql.Tx). It asserts each call's sumCredentialUtxoStake delta is independently 1 (no leakage between calls' merged ref sets) and that Tx-scoped statement retention stays bounded by the number of distinct cached queries, not by call count. Reverting stmtForQueryer's per-transaction cache reproduces 6 retained statements against a bound of 3, confirming the test exercises #4271's fix.

wolf31o2 (approved):

PR #4299, opened 2026-09-15
test(blockfrost): cover script stake address adapter

wolf31o2:
Exercise `NodeAdapter.Address` with a database-backed key-payment/script-stake
base address. Assert that the returned testnet reward address preserves the
script credential and script-stake address type.

Replace the assertion that depended on the upstream address-extraction defect.
Related to blinklabs-io/gouroboros#2032 and blinklabs-io/gouroboros#2328.

## Summary by Cubic

The test now exercises the adapter's database-backed address response instead
of testing the credential-derivation helper in isolation.

## Summary by CodeRabbit

Coverage checks the returned testnet script stake address, script hash and type.


api/blockfrost/blockfrost_test.go
```diff
@@ -3559,20 +3557,26 @@ func TestKeyScriptStakeAddressDerivation(t *testing.T) {
...
+	info, err := adapter.Address(addr.String())
 	require.NoError(t, err)
-	assert.True(t, strings.HasPrefix(encoded, "stake_test17"))
+	require.NotNil(t, info.StakeAddress)
+	assert.True(t, strings.HasPrefix(*info.StakeAddress, "stake_test17"))
 
-	stakeAddr, err := lcommon.NewAddress(encoded)
+	stakeAddr, err := lcommon.NewAddress(*info.StakeAddress)
 	require.NoError(t, err)
+	assert.Equal(t, uint(lcommon.AddressNetworkTestnet), stakeAddr.NetworkId())
```
chrisguiney, line 3578: This holds even if the adapter drops the network id: replacing addr.NetworkId() with 0 in the KeyScript fallback of NodeAdapter.Address leaves the test green, since testnet is network 0 and newDBBackedAdapter has no node config. Add a mainnet case: build the ledger state with cardano.NewCardanoNodeConfigFromEmbedFS(cardano.EmbeddedConfigFS, "mainnet/config.json"), use an AddressNetworkMainnet key/script address, and assert a stake17 prefix and header byte 0xf1. That case passes at this head and fails with the network forced to 0.

chrisguiney (changes requested):
This test fails when the KeyScript fallback in NodeAdapter.Address uses the key-hash type, the payment hash, is disabled, or forces mainnet; it stays green when the network id is forced to 0. One change inline.

api/blockfrost/blockfrost_test.go
```diff
@@ -3544,15 +3545,10 @@ INSERT INTO address_transaction (
 	assert.Nil(t, info.StakeAddress)
 }
 
-func TestKeyScriptStakeAddressDerivation(t *testing.T) {
+func TestNodeAdapterKeyScriptStakeAddress(t *testing.T) {
```
chrisguiney, line 3548: Only the script direction is pinned. Rebuilding the stake address in NodeAdapter.Address as AddressTypeNoneScript for every input, keeping network and hash unchanged, leaves go test ./api/blockfrost/ green: nothing in this package catches a key-staking base address reported as stake_test17. Blockfrost consumers key accounts on that string, so a wrong type bit returns a valid-looking reward address for a different account. Add the AddressTypeKeyKey counterpart through this same adapter path, asserting stake_test1u and AddressTypeNoneKey.

api/blockfrost/adapter_block_db_test.go
```diff
@@ -45,6 +46,7 @@ import (
 // not perform slot/epoch/time math, and NewLedgerState tolerates a nil config.
 func newDBBackedAdapter(
 	t *testing.T,
+	nodeConfig ...*cardano.CardanoNodeConfig,
```
chrisguiney, line 48: The doc comment above still says no CardanoNodeConfig is supplied because NewLedgerState tolerates a nil config, which this parameter and the mainnet case in TestNodeAdapterKeyScriptStakeAddress both contradict. Reword it to say the config defaults to absent and is supplied when a test needs a specific network id.

chrisguiney (changes requested):
Prior blocker is cleared and proven causally: rebuilding the stake address in NodeAdapter.Address at network 0 now fails the mainnet half at blockfrost_test.go:3603 and :3608. Forcing the key type, substituting the payment hash, and dropping StakeAddress each fail the test too. Merged with current main the tree builds and go vet, golangci-lint ./api/blockfrost/... and go test -race ./api/blockfrost/ all pass. Two items inline. CodeRabbit reports incremental reviews disabled and its declared range ends at the previous head; no cubic record exists at this head.

arepala-uml (dismissed):

api/blockfrost/blockfrost_test.go
wolf31o2: Fixed in cff53303: the mainnet script-stake case asserts the stake17 prefix and 0xf1 header byte.

api/blockfrost/blockfrost_test.go
wolf31o2: Fixed in 88430d07: the key-staking counterpart asserts stake_test1u and AddressTypeNoneKey.

api/blockfrost/adapter_block_db_test.go
wolf31o2: Fixed in 88430d07: the helper comment now describes the optional network-specific configuration.

api/blockfrost/blockfrost_test.go
```diff
@@ -3562,23 +3558,82 @@ func TestKeyScriptStakeAddressDerivation(t *testing.T) {
...
-		},
+	info, err := adapter.Address(addr.String())
+	require.NoError(t, err)
+	require.NotNil(t, info.StakeAddress)
+	assert.True(t, strings.HasPrefix(*info.StakeAddress, "stake_test17"))
+
+	stakeAddr, err := lcommon.NewAddress(*info.StakeAddress)
+	require.NoError(t, err)
+	assert.Equal(t, uint(lcommon.AddressNetworkTestnet), stakeAddr.NetworkId())
+	assert.Equal(t, uint8(lcommon.AddressTypeNoneScript), stakeAddr.Type())
```
chrisguiney, line 3580: This pins the script direction through NodeAdapter.Address, which reads addr.StakeAddress() and never calls stakeAddressFromCredential. Removing TestKeyScriptStakeAddressDerivation drops the last test of that helper's script branch at adapter.go:5035-5038: replacing it with a constant AddressTypeNoneKey leaves go test ./api/blockfrost/ green here and fails on main. Seven production sites call it, so a wrong type bit emits stake_test1u for a script-staking account. Restore a script-hash assertion on that helper.

chrisguiney (changes requested):
Prior three findings cleared and proven causally: forcing network 0 in NodeAdapter.Address fails the mainnet half at blockfrost_test.go:3603 and :3608; rebuilding every stake address as script type fails the key half at :3632 and :3635; dropping StakeAddress fails :3574. The helper doc comment now matches the variadic parameter. One new blocker inline. Head is already the merged tree; go vet, golangci-lint and go test -race ./api/blockfrost/ pass, and no bot reviewed this head.

chrisguiney (approved):

PR #4307, opened 2026-09-15
perf(sqlstore): cache insertUtxoModel's INSERT and asset-id lookup

chrisguiney:
Depends on #4271 -- not mergeable until it lands first. This branch's only new content is the final commit, `perf(sqlstore): cache insertUtxoModel's INSERT and asset-id lookup`.

## Problem

Live CPU profiling (30s sample) found `insertUtxoModel`'s raw INSERT statement (`database/plugin/metadata/sqlstore/transaction_write.go`) as the single largest individually-attributed line in the entire sample -- 3.07s of 34.91s total (8.8%). A companion asset-id lookup cost another 780ms (2.2%). Root cause: `modernc.org/sqlite`'s driver has no internal statement cache, and the raw `db.QueryRowContext(ctx, literalSQL, ...)` calls re-parse/re-plan the identical SQL text on every UTxO output insert -- the same class of bug already fixed elsewhere on other hot paths this cycle, just not yet applied to dingo's single hottest call site.

## Change

Routed the INSERT and asset-id SELECT through the existing hot-statement prepared cache (`queryRowCached`/`hotStatements`, from #4270/#4271). Since the INSERT carries `RETURNING id`, and MySQL's `dialectQueryer.QueryRowContext` emulates RETURNING via its own `ExecContext`+`LastInsertId` path that never consults a cached statement, added a `cacheableForDialect` guard so `prepareHotStatements` skips caching any RETURNING-id query specifically on MySQL -- preserving MySQL's existing correct behavior unchanged, while SQLite and PostgreSQL (native RETURNING support) get the caching benefit.

## Benchmark

`BenchmarkInsertUtxoModel`, one-shot vs. cached, consistent ~15-30% lower ns/op across repeated runs (e.g. 104738 -> 89134 ns/op).

## Tests

Correctness round-trip via `GetUtxo`, the ON-CONFLICT-DO-NOTHING fallback path, cached-statement pointer-identity reuse across independent write transactions for both queries, a table-driven test of the MySQL/RETURNING skip predicate, and an end-to-end proof that `prepareHotStatements` leaves RETURNING queries uncached on MySQL while still caching a non-RETURNING one.

---
## Summary by cubic
Caches the UTxO INSERT and asset-ID lookup used by `insertUtxoModel`, reducing repeated SQL parsing during sync. SQLite and PostgreSQL reuse prepared statements, while MySQL keeps its existing uncached `RETURNING id` path because its emulation bypasses prepared `QueryRow` statements.

**Details**

- Also caches reward live-stake queries and the credential UTxO stake aggregate.
- Write transactions retain one derived statement per cached query instead of one per call.
- Query results and the MySQL `RETURNING id` behavior remain unchanged, including uncached fallbacks.
- Adds correctness, dialect, retention, and benchmark coverage; cached UTxO inserts are roughly 15–30% faster.

<sup>Written for commit 6abacc85901defce300b7b036d94bb3cd5f2a2a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4307?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Performance**
  - Improved database write performance by reusing prepared operations for UTxO, asset, and live-stake processing.
  - Reduced repeated query setup across transactions, with fallback handling when cached operations are unavailable.

- **Compatibility**
  - Preserved consistent behavior across SQLite, PostgreSQL, and MySQL database configurations, including MySQL query variations.

- **Reliability**
  - Expanded validation to confirm cached database operations produce the same results as direct execution across repeated and concurrent transactions.


database/plugin/metadata/sqlstore/transaction_write.go
```diff
@@ -709,19 +762,12 @@ func (s *Store) insertUtxoModel(
...
 	var id int64
-	err = db.QueryRowContext(ctx, `
-INSERT INTO utxo (
-    transaction_id, collateral_return_for_tx_id, tx_id, payment_key,
-    staking_key, credential_tag, datum_hash, spent_at_tx_id,
-    referenced_by_tx_id, collateral_by_tx_id, added_slot, deleted_slot,
-    amount, output_idx, payment_script
-) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`+conflict+`
-RETURNING id`,
+	err = s.queryRowCached(ctx, db, query,
```
wolf31o2, line 896: queryRowCached calls Tx.StmtContext on every invocation, and database/sql appends each returned *sql.Stmt to tx.stmts and registers it through db.addDep under the global db.mu, releasing both only at commit or rollback. Routing the per-output INSERT here makes retention proportional to outputs per write transaction: 20000 inserts in a single transaction retain 20000 statements and 5824 KiB, against 0 and 168 KiB with this one call site left on db.QueryRowContext. Derive the Tx-scoped statement once per transaction rather than once per output.

database/plugin/metadata/sqlstore/transaction_write.go
```diff
@@ -798,10 +844,7 @@ WHERE id = ?`,
 			return err
 		}
 		var assetID uint
-		if err := db.QueryRowContext(ctx, `
-SELECT id FROM asset
-WHERE utxo_id = ? AND policy_id = ? AND name = ?
-ORDER BY id DESC LIMIT 1`,
+		if err := s.queryRowCached(ctx, db, getAssetIDQuery,
```
wolf31o2, line 973: The per-asset id lookup compounds the same per-transaction retention: an output carrying multiple assets adds one Tx-scoped statement per asset on top of the one already added per output, all held until the batch commits.

database/plugin/metadata/sqlstore/insert_utxo_stmt_cache_test.go
```diff
@@ -0,0 +1,371 @@
...
+		TxId:       txID,
+		OutputIdx:  outputIdx,
+		PaymentKey: paymentKey,
+		AddedSlot:  1,
+		Amount:     types.Uint64(amount),
+	}
+}
+
+// insertUtxoInTxn runs insertUtxoModel inside its own write transaction, the
+// same one-transaction-per-output access pattern ledgerProcessBlock uses in
```
wolf31o2: Production does not use one transaction per output. LedgerDeltaBatch.apply in ledger/delta.go applies a whole batch of blocks under the single txn passed from ledger/state.go, and genesis inserts the entire Byron and Shelley UTxO set inside one txn.Do in ledger/chainsync.go. Since insertUtxoInTxn passes a nil txn, every test here receives a fresh transaction per output, so none can observe the per-transaction statement growth this caching introduces.

wolf31o2 (changes requested):
insertUtxoModel's move onto queryRowCached retains one *sql.Stmt per call for the lifetime of the enclosing write transaction, and the comment justifying the change misstates the production transaction scope.

database/plugin/metadata/sqlstore/transaction_write.go
chrisguiney: Fixed by merging perf/reward-live-stake-touch-cache's 00bedb10 forward (9dc94fdd). queryRowCached routes this INSERT through stmtForQueryer/txScopedStmt, which now derives the Tx-scoped *sql.Stmt once per (tx, cached) pair instead of once per call. Measured with TestInsertUtxoModelBoundsTxScopedStatementRetentionInOneTransaction: 20000 outputs in one write transaction retain <=2 Tx-scoped statements after the merge, 40000 before it (reverting the fix reproduces that number).

database/plugin/metadata/sqlstore/transaction_write.go
chrisguiney: Same fix covers this call site: getAssetIDQuery goes through the same queryRowCached/txScopedStmt path, so it derives once per transaction too. TestInsertUtxoModelBoundsTxScopedStatementRetentionInOneTransaction inserts one asset per output (20000 outputs) and asserts the combined retained-statement count for both cached queries stays <=2.

database/plugin/metadata/sqlstore/insert_utxo_stmt_cache_test.go
chrisguiney: Fixed in 9dc94fdd: corrected insertUtxoInTxn's doc comment to state production applies many outputs within one shared write transaction (LedgerDeltaBatch.apply for a block batch, one txn.Do for the genesis UTxO set), not one transaction per output. Added TestInsertUtxoModelBoundsTxScopedStatementRetentionInOneTransaction, which exercises that real multi-output-per-transaction shape and reads database/sql's own Tx.stmts.v length via retainedTxStmtCount.

wolf31o2 (approved):

PR #4308, opened 2026-09-15
fix(mithril): give mock Shelley test blocks a non-Dijkstra-shaped body

### LINKED ISSUE #4306: mithril: bootstrap tests' mock Shelley block collides with gouroboros' new Dijkstra-block classifier
## Summary

Bumping `go.mod`'s gouroboros dependency past commit `818459a5` (which
includes a fix for blinklabs-io/gouroboros#2301) breaks 5 tests in
`mithril/`:

- `TestSyncResumesPinnedArtifactAfterAggregatorAdvance`
- `TestSyncFreshBootstrapSelectsLatestArtifact`
- `TestSyncRefusesResumeWhenCertifiedTipMoved`
- `TestSyncV2NoCertVerificationUsesExtractDirLedgerState`
- `TestBootstrapSurvivesAnUnavailableAncillaryArchive`

All fail identically, at the same synthetic block (slot 999):

```
mithril bootstrap failed: downloading immutable archives: pipelined copy to slot 1000:
failed to import block: add raw block batch: block at slot 999: extract transaction
offsets: dijkstra block body has 0 elements, expected 3 or 4
```

Confirmed both directions: all 5 pass cleanly on current `main` (pre-bump);
all 5 fail with the bump, same error, same slot.

## Root cause

`mithril/bootstrap_v2_test.go`'s `validImmutableFiles`/`makeBlock` helper
builds a lightweight mock **Shelley** block by hand:

```go
header := shelley.ShelleyBlockHeader{Body: shelley.ShelleyBlockHeaderBody{...}}
headerCbor, _ := cbor.Encode(header)
blockCbor, _ := cbor.Encode([]any{
    cbor.RawMessage(headerCbor), []any{},   // [header, empty-tx-list]
})
```

This produces a 2-element top-level array `[header, []]`. That shape is
meant as a generic "header + empty transaction list" placeholder, but it's
byte-for-byte indistinguishable from CIP-0137 Dijkstra's real block CDDL
(`block = [header, block_body]`) at the structural level gouroboros can see.

Gouroboros' new `classifyDijkstraBlock` (`ledger/common/common.go`, landed
as part of the #2301 fix) checks:

1. Is the top-level array exactly 2 elements? (yes -- Byron is 3,
   Shelley-onward proper encoding is 4+, so 2-element is otherwise
   Dijkstra-unique)
2. Does the first element look like a real header --
   `isDijkstraCompatibleHeader`, which requires the value to itself be a
   2-element array (`[headerBody, signature]`) whose first element has
   **>= 10 fields**?
3. If both hold, does the second element (the "body") have exactly 3 or 4
   elements?

`shelley.ShelleyBlockHeader` CBOR-encodes as exactly
`[headerBody, signature]`, and `ShelleyBlockHeaderBody` has well over 10
fields (block number, slot, prev hash, issuer/VRF vkeys, nonce/leader VRF,
body size/hash, opcert fields, protocol version) even with most left
zero-valued in the test. So step 2 passes for a genuine Shelley header --
`isDijkstraCompatibleHeader` doesn't actually distinguish Shelley from
Babbage/Dijkstra headers, it just checks a generic shape both happen to
share. Step 3 then fails because the test's placeholder body is `[]` (0
elements, not 3 or 4), and gouroboros -- correctly, per its own new
hardening -- raises a loud error instead of the old silent fallback.

## Two candidate fixes, different repos

1. **dingo-side (likely the right one):** `AGENTS.md` says mock fixtures
   belong in `ouroboros-mock`, never hand-rolled locally. This helper
   hand-encodes a block shape instead of using a shared builder. Building it
   through `ouroboros-mock`'s real Shelley block encoding (or any shape that
   isn't a bare 2-element array) would sidestep the collision entirely and
   bring this fixture in line with the "never duplicate mocks" rule.
2. **gouroboros-side (more defensive, protects every consumer):**
   `isDijkstraCompatibleHeader`'s >=10-field check doesn't actually
   discriminate Shelley-shaped headers from Babbage/Dijkstra-shaped ones.
   Tightening it (e.g. checking a field count/shape that's genuinely unique
   to the Babbage-onward header layout) would stop this whole collision
   class regardless of what any caller's test fixtures look like.

## Impact

Blocks bumping dingo's gouroboros dependency past this point until fixed --
currently blocking the DMQ phase 2 authenticator swap (issue #1949), which
needs the upgraded, stake-aware `MessageAuthenticator` from
blinklabs-io/gouroboros#2325.

## References

- blinklabs-io/gouroboros#2301 (the fix that surfaced this)
- blinklabs-io/gouroboros#2325 (the DMQ authenticator upgrade blocked by this)
- `mithril/bootstrap_v2_test.go:84-112` (`validImmutableFiles`/`makeBlock`)
- `ledger/common/common.go` in gouroboros (`classifyDijkstraBlock`,
  `isDijkstraCompatibleHeader`)

cryptodj413:
## Summary

`mithril/bootstrap_v2_test.go`'s `validImmutableFiles`/`makeBlock` helper
encoded a mock Shelley block body as `[header, []]` — a bare 2-element
top-level array. That's byte-for-byte indistinguishable from CIP-0137
Dijkstra's real block CDDL (`block = [header, block_body]`) at the
structural level gouroboros can see: Byron blocks are 3 elements,
Shelley-onward is otherwise always 4+, so a 2-element array is otherwise
Dijkstra-unique.

Bumping gouroboros past the fix for blinklabs-io/gouroboros#2301 turns a
previously silent fallback into a loud, specific error for exactly this
shape:

```
extract transaction offsets: dijkstra block body has 0 elements, expected 3 or 4
```

Gouroboros' new `classifyDijkstraBlock` sees a 2-element array whose first
element structurally resembles a real header (`isDijkstraCompatibleHeader`
checks for a `[headerBody, signature]` shape with `headerBody` having >= 10
fields — a check `shelley.ShelleyBlockHeader` happens to satisfy too, since
it doesn't actually discriminate Shelley from Babbage/Dijkstra headers),
and correctly refuses to guess when the second element (the "body") isn't a
valid 3-or-4-element Dijkstra body.

This broke 5 mithril bootstrap tests that build chunk fixtures through this
helper (all pass on current `main`, all failed identically at the same
synthetic block once the dependency was bumped past that fix):

- `TestSyncResumesPinnedArtifactAfterAggregatorAdvance`
- `TestSyncFreshBootstrapSelectsLatestArtifact`
- `TestSyncRefusesResumeWhenCertifiedTipMoved`
- `TestSyncV2NoCertVerificationUsesExtractDirLedgerState`
- `TestBootstrapSurvivesAnUnavailableAncillaryArchive`

## Fix

Encode the mock block as a proper 4-element Shelley-shaped array instead:
`[header, transaction_bodies, transaction_witness_sets, auxiliary_data_set]`
with the last three left empty. This is both more correct — a real
Shelley-onward block is never a bare 2-tuple — and sidesteps the collision
entirely, on both the current and any future gouroboros dependency.

Verified the fix works against **both** dependency versions: the current
pinned `v0.204.4` and a build pinned past the gouroboros #2301 fix.

## Checks

| Command | Result |
|---|---|
| `gofmt -l mithril/bootstrap_v2_test.go` | clean |
| `go build ./...` | pass |
| `go vet ./mithril/...` | pass |
| `golangci-lint run ./mithril/...` | 0 issues |
| `make docs-parity` | pass |
| `go test ./mithril/... -race -count=1` | pass |
| `go test ./mithril/... -race -count=1` against current `v0.204.4` | pass |
| `go test ./mithril/... -race -count=1` against a build pinned past gouroboros#2301's fix | pass (all 5 previously-broken tests pass) |

`nilaway ./mithril/...` reported only tool-internal "failed prerequisites"
analyzer-plumbing errors, reproduced identically on an untouched package
(`./dmq/...`) — a pre-existing local toolchain issue, not related to this
change (exit code 0, no findings against this file).

No `DATABASE.md`/`ARCHITECTURE.md` update needed: this is a test-fixture
correctness fix with no schema, storage, or architecture change.

Fixes #4306

---
## Summary by cubic
Updates Mithril bootstrap fixtures from an invalid two-element mock Shelley block (`[header, []]`) to the proper four-element shape. This prevents `gouroboros` from treating the fixture as a malformed Dijkstra block and restores bootstrap tests after the dependency fix.

- Keeps transaction bodies, witness sets, and auxiliary data empty while preserving the fixture’s purpose.
- Fixes #4306.

<sup>Written for commit afcf8a36196b8a61a6c1fa71dfed2ed5423438ab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4308?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Shelley block body encoding to ensure blocks are interpreted correctly and avoid malformed-block errors.


cryptodj413:
Superseded by #4304 (already merged), which independently landed an equivalent fix as part of bumping gouroboros to v0.204.7. Closing in favor of that.

PR #4311, opened 2026-09-15
test(mithril): drive download idle timeouts deterministically

wolf31o2:
The Mithril idle-timeout tests used short wall-clock thresholds, so scheduler load could create unscripted retries and misleading `Content-Range` failures.

- Add a package-private watchdog factory so retry tests can inject idle events at exact byte offsets while production continues to use `time.AfterFunc`.
- Make retry attempts, resume ranges, and progress-budget resets deterministic.
- Keep a real-timer integration test on the body-copy path, bounded by `testutil.AsyncWait`.
- Exercise direct delivery of the real timer callback.
- Read timer state directly when verifying that reads disarm the watchdog.

This removes short-duration outcome checks from the Mithril package without changing production timeout behavior.


mithril/download_test.go
```diff
@@ -503,33 +652,45 @@ func TestDownloadSnapshotAcceptsHTTPSByDefault(t *testing.T) {
...
-		20*time.Millisecond,
-		func() {
-			select {
-			case idleCh <- struct{}{}:
-			default:
-			}
-		},
-	)
+	var idleCount atomic.Int32
+	timer := newIdleTimer(time.Hour, func() {
```
chrisguiney, line 723: This arms a timer at time.Hour and never lets it expire, and both download tests now drive the injected watchdog, so nothing exercises time.AfterFunc -> onIdle -> cancelDownload any more. Arming newIdleTimerRun at 24h instead of the configured timeout keeps the entire mithril package green at this head and in a tree merged with current main; on the base branch that same edit fails both download idle tests with context deadline exceeded. Add a test that lets a short idleTimer fire and blocks on a channel bounded by testutil.AsyncWait - a delay waited for, not a deadline raced.

mithril/download_test.go
```diff
@@ -237,6 +238,97 @@ func TestDownloadSnapshotResume(t *testing.T) {
...
+	onIdle   func()
+}
+
+func (w *injectedIdleWatchdog) Reset() {
+	if w.injector.take() {
+		w.onIdle()
+	}
+}
+
+func (w *injectedIdleWatchdog) Stop() {}
```
chrisguiney, line 343: Non-blocking: injectedIdleWatchdog calls onIdle synchronously inside Reset on the reading goroutine and Stop does nothing, while the idleWatchdog contract states that Stop returns only once a callback that was already firing has returned. The concurrency between the time.AfterFunc goroutine and a body read blocked in Read is therefore no longer present in these tests under -race. Honouring the Stop contract in the fake, or keeping one test on the real timer, restores it.

chrisguiney (changes requested):
The rewrite removes the only coverage of the production idle timer actually firing: arming newIdleTimerRun at 24h instead of the configured timeout leaves the whole mithril package green at this head and in a tree merged with current main, while the same edit on the base branch fails both download idle tests. The reset test itself is genuinely strengthened - deleting consecutiveIdleRetries = 0 or the !madeProgress guard fails it here and passed on the base. Package passes -race -count=20 under full-package parallelism; lint, import-boundaries and nested modules clean.

mithril/download_test.go
```diff
@@ -265,98 +357,175 @@ func TestDownloadSnapshotIdleTimeoutRetriesAndResumes(t *testing.T) {
...
+
+	_, err := DownloadSnapshot(context.Background(), DownloadConfig{
+		URL:               server.URL + "/snapshot.tar.zst",
+		AllowInsecureHTTP: true,
+		DestDir:           t.TempDir(),
+		Filename:          "configured-idle-timeout.tar.zst",
+		IdleTimeout:       100 * time.Millisecond,
+		MaxIdleRetries:    0,
+	})
+	require.ErrorIs(t, err, errDownloadIdleTimeout)
```
chrisguiney, line 472: This assertion does not pin which watchdog fired. The header watchdog armed before client.Do and the body watchdog inside idleTimeoutReader both wrap errDownloadIdleTimeout, so inserting a 200ms sleep before WriteHeader in this handler makes all 13 attempts expire in the header phase with partial_bytes=0, and the test still passes in 1.42s without ever running idleTimeoutReader. A 100ms wall-clock bound on the header exchange is the same load sensitivity this change removes elsewhere. Assert that the error came from the body copy, whose message carries the writing snapshot data prefix.

mithril/download_test.go
```diff
@@ -265,98 +357,175 @@ func TestDownloadSnapshotIdleTimeoutRetriesAndResumes(t *testing.T) {
...
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte("A"))
+		if flusher, ok := w.(http.Flusher); ok {
+			flusher.Flush()
+		}
+		<-r.Context().Done()
+	}))
+	t.Cleanup(server.Close)
+
+	_, err := DownloadSnapshot(context.Background(), DownloadConfig{
```
chrisguiney: This handler blocks forever on r.Context().Done() and only the production idle timer releases it, but the context passed here is unbounded, so a regression in that timer hangs the package for the full -timeout 30m that go-test.yml uses and dies in a goroutine dump rather than one named failure; arming newIdleTimerRun at 24h reproduced exactly that. The two other forever-blocking handlers in this file bound the download with testutil.AsyncWait. Separately, MaxIdleRetries 0 selects the default of 12, not none, so this runs 14 attempts and 13 idle waits.

chrisguiney (changes requested):
Prior blocker 1 is cleared: arming newIdleTimerRun at 24h in place now fails only TestDownloadSnapshotIdleTimeoutUsesConfiguredTimer while both injected tests still pass in 0.00s, so the production timer is covered again. Prior blocker 2 is cleared by that same test holding a real Read against a real time.AfterFunc. Two new issues in it: the assertion is satisfied by the header watchdog rather than the body one, and an unbounded context turns a timer regression into a 30m package timeout. Package is clean at -race -count=15 and on a tree merged with current main.

mithril/download_test.go
```diff
@@ -265,98 +359,177 @@ func TestDownloadSnapshotIdleTimeoutRetriesAndResumes(t *testing.T) {
...
+	defer cancel()
+	_, err := DownloadSnapshot(ctx, DownloadConfig{
+		URL:               server.URL + "/snapshot.tar.zst",
+		AllowInsecureHTTP: true,
+		DestDir:           t.TempDir(),
+		Filename:          "configured-idle-timeout.tar.zst",
+		IdleTimeout:       100 * time.Millisecond,
+		MaxIdleRetries:    0,
+	})
+	require.ErrorIs(t, err, errDownloadIdleTimeout)
```
chrisguiney, line 472: Removing the body watchdog at its call site fails this test at the 120s bound, so it is causally tied to idleTimeoutReader. What it does not pin is which watchdog satisfied it. With the body watchdog removed and a 200ms sleep added before WriteHeader in this handler, attempt 1 expires in the header phase and the test passes in 1.4s. A header exchange slower than the 100ms IdleTimeout therefore hides a body-watchdog regression, which is the load dependence this change removes elsewhere. Adding require.ErrorContains(t, err, "writing snapshot data") pins the body copy.

mithril/download_test.go
```diff
@@ -237,6 +238,99 @@ func TestDownloadSnapshotResume(t *testing.T) {
...
+	onIdle   func()
+}
+
+func (w *injectedIdleWatchdog) Reset() {
+	if w.injector.take() {
+		w.onIdle()
+	}
+}
+
+// Reset invokes the injected callback synchronously, so Stop has no concurrent
```
chrisguiney: This comment documents Stop but opens with the name Reset, so a reader looking up Stop lands on a sentence about a different method. Wording it as Stop having no concurrent callback to wait for because Reset invokes onIdle synchronously keeps both the meaning and the naming convention.

mithril/download.go
```diff
@@ -213,36 +221,58 @@ func (cfg DownloadConfig) sizeLimitError() error {
...
+// idleWatchdog bounds the time a download may spend making no progress.
+// A download arms one while it waits for response headers and re-arms one
+// around every body read; an armed watchdog that expires cancels the
+// download so the retry loop can resume from the bytes already on disk.
+// A watchdog is armed by its constructor.
+type idleWatchdog interface {
+	// Reset abandons any running idle period and starts a new one.
+	Reset()
+	// Stop abandons any running idle period without starting another. It
+	// returns only once a callback that was already firing has returned.
```
chrisguiney, line 233: This contract is newly stated and nothing exercises it: deleting the channel receive in idleTimerRun.stopAndWait leaves the whole mithril package green under -race -count=2. The same deletion is also green against the tests at origin/main, so this change did not lose that coverage and the gap predates it. Noted here because the injected double declines the same contract by design.

mithril/download.go
```diff
@@ -213,36 +221,58 @@ func (cfg DownloadConfig) sizeLimitError() error {
...
+	Stop()
+}
+
+// newIdleWatchdog builds the watchdog enforcing this download's idle
+// timeout, or nil when idle detection is disabled.
+func (cfg DownloadConfig) newIdleWatchdog(onIdle func()) idleWatchdog {
```
chrisguiney, line 239: Production behaviour is preserved: all three construction sites keep an explicit nil check, so a disabled idle timeout yields a nil interface rather than a call on nil, and DownloadConfig was already non-comparable through OnProgress. Checked at 54bb4ee5, which is the merged tree since the merge base equals origin/main: go build, go vet, golangci-lint ./mithril/ 0 issues, import-boundaries, docs-parity, go test -race -count=3 ./mithril/ ok in 20.1s. ARCHITECTURE.md and DATABASE.md unaffected. No bot reviewed this head.

chrisguiney (dismissed):
Both prior blockers clear: arming newIdleTimerRun at 24h now fails TestDownloadSnapshotIdleTimeoutUsesConfiguredTimer as a named assertion at 120.00s rather than hanging the package, and removing the body watchdog at its call site fails that same test, so the real timer is covered. One earlier point still stands, because the assertion does not pin which watchdog fired. go-test (Linux) (1.26.x) is red on a module proxy stream error fetching blockfrost-go v0.5.0 in api/blockfrost, which is not from this change - mithril reported ok in that same job.

mithril/download_test.go
wolf31o2: Fixed in a183cb3e: TestIdleTimerRunsCallback lets the real time.AfterFunc-backed timer fire and waits for the callback through testutil.RequireReceive bounded by testutil.AsyncWait.

mithril/download_test.go
wolf31o2: Fixed in a183cb3e: the configured-timer integration test now keeps the real timer and blocks the response body until request cancellation, while the direct timer test exercises callback delivery.

mithril/download_test.go
wolf31o2: Fixed in a183cb3e: the test now asserts the writing snapshot data prefix and uses an immediate custom transport, so only the body-copy watchdog can satisfy it.

mithril/download_test.go
wolf31o2: Fixed in a183cb3e: the body read is bounded by the request context, the outer context uses testutil.AsyncWait, and MaxIdleRetries is explicitly 1 (two total attempts).

mithril/download_test.go
wolf31o2: Fixed in a183cb3e: the immediate custom transport removes the header timing race and the error assertion pins the writing snapshot data path.

mithril/download_test.go
wolf31o2: Fixed in d227c2e0: the comment now opens with Stop and explains that Reset invokes the callback synchronously.

mithril/download.go
wolf31o2: Confirmed as a pre-existing, non-blocking coverage gap. This change adds direct real-timer callback coverage; it does not alter idleTimerRun.stopAndWait or claim new coverage of its wait guarantee.

mithril/download.go
wolf31o2: Confirmed. The production behavior and nil handling remain unchanged; no code change is required for this informational validation thread.

arepala-uml (approved):

PR #4314, opened 2026-09-15
fix(koiosparity): bound Koios HTTP client's connection phases explicitly

chrisguiney:
## Problem

`KoiosClient` (`internal/koiosparity/koios_client.go`) built its `http.Client` with only a coarse, whole-round-trip `Timeout: 60 * time.Second` and no explicit `Transport` -- relying entirely on `http.DefaultTransport`'s defaults for every earlier phase (dial, TLS handshake, waiting for response headers).

Verified against the actual Go 1.26.7 stdlib source rather than assumed: dial (30s) and TLS handshake (10s) were already independently bounded by `http.DefaultTransport`'s defaults, and `Client.Timeout` does correctly cover the full round trip including slow body reads -- so this was never a genuinely unbounded hang. But `ResponseHeaderTimeout` was not set at all (zero, meaning unbounded from the transport's own perspective), and the only thing bounding "connection open, server writes nothing" was the single coarse 60s knob shared across every phase, rather than phase-specific, independently-attributable timeouts. A slow-to-respond Koios endpoint could tie up a connection for the full 60s in a phase that should fail in a fraction of that, and a genuine hang would be harder to diagnose without per-phase attribution.

The retry loop (`koiosMaxRetries = 3`) is already bounded, not unbounded -- no fix needed there.

## Change

Added `newKoiosTransport`, cloning `http.DefaultTransport` with explicit, named-constant timeouts for every connection phase: `DialContext` (10s dial, 30s keepalive via `net.Dialer`), `TLSHandshakeTimeout` (10s), `ResponseHeaderTimeout` (30s, previously unbounded), `ExpectContinueTimeout` (1s) -- all shorter than the unchanged 60s overall `Client.Timeout`, so this is additive defense-in-depth, not a replacement.

## Tests

`TestNewKoiosTransportResponseHeaderTimeout`: a server accepts the connection and sleeps 3s before writing anything; the client (10s overall timeout) fails in well under 3s with `"timeout awaiting response headers"`, proving the phase-specific timeout fires rather than the coarse one. `TestNewKoiosTransportDialTimeout`: dials a non-routable address with a 300ms dial timeout against a 10s overall timeout; verified at both 300ms and 1200ms to rule out an instant "network unreachable" false positive before confirming the real, scaling timeout.

## Note

Two other HTTP client constructions in this repo (`node.go`, `bark/blob.go`) share the same "coarse `Client.Timeout` only, no explicit `Transport`" shape -- out of scope for this PR, worth a follow-up.

---
## Summary by cubic
Koios requests now use explicit timeouts for each HTTP connection phase instead of relying on the 60s client timeout and transport defaults. Stalled dials and servers that never send response headers now fail sooner with phase-specific errors, while the overall timeout remains unchanged.

- Sets 10s dial, 10s TLS handshake, 30s response-header, and 1s expect-continue timeouts.
- Clones `http.DefaultTransport` to preserve proxy, HTTP/2, and connection-pooling behavior, with a safe fallback for other transport types.
- Tests response-header and dial enforcement, proxy isolation, and the fallback path; the retry loop remains capped at 3 attempts.

<sup>Written for commit dcfdbe82f1b67d8c8524e2c64e772bbfaa7bd4ad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4314?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network timeout handling for Koios requests by applying separate limits to connection, TLS negotiation, response headers, and request continuation phases.
  * Requests now fail faster when a server is unreachable or delays its response headers.

* **Tests**
  * Added coverage verifying connection and response-header timeouts are enforced independently.


internal/koiosparity/koios_client_test.go
```diff
@@ -557,3 +557,128 @@ func TestGetAccountRewardHistoryEmptyAddressesNoRequest(t *testing.T) {
...
+			koiosDialKeepAlive,
+			koiosTLSHandshakeTimeout,
+			koiosResponseHeaderTimeout,
+			koiosExpectContinueTimeout,
+		),
+	}
+
+	start := time.Now()
+	//nolint:noctx // deliberately no context; the transport's own dial timeout is exactly what's under test
+	resp, err := client.Get("http://192.0.2.1:81/")
```
wolf31o2, line 714: newKoiosTransport keeps DefaultTransport's ProxyFromEnvironment and 192.0.2.1 is not loopback, so with HTTP_PROXY set this Get dials the proxy and fails instantly: the 300ms dial bound is never exercised yet the test still passes. An immediate ENETUNREACH on a runner with no default route passes the same way, because err != nil plus elapsed < clientTimeout is the only discriminating assertion. Setting Proxy to nil on the test transport and requiring a net.Error whose Timeout() is true, with an elapsed bound near 2s, closes both.

internal/koiosparity/koios_client.go
```diff
@@ -340,6 +370,45 @@ func NewKoiosClient(
...
+// time and independently of the client-level Timeout.
+// expectContinueTimeout caps waiting for a "100 Continue" status before
+// sending a request body when the client sets the Expect header; this
+// client never sets Expect, so it is inert today and included purely for
+// completeness against a future caller of this transport that does.
+func newKoiosTransport(
+	dialTimeout, dialKeepAlive time.Duration,
+	tlsHandshakeTimeout, responseHeaderTimeout, expectContinueTimeout time.Duration,
+) *http.Transport {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
```
wolf31o2: http.DefaultTransport.(*http.Transport) panics instead of erroring if DefaultTransport is ever replaced by a wrapping RoundTripper, and this path runs at node startup through NewKoiosClient. mithril/download.go:429 and :455 do the same clone with the comma-ok form and a &http.Transport{Proxy: http.ProxyFromEnvironment} fallback; matching that idiom keeps this constructor from being the one place a replaced DefaultTransport crashes the process.

wolf31o2 (changes requested):
The per-phase bounds are correct: Clone preserves ForceAttemptHTTP2 so the custom DialContext does not disable HTTP/2, and Go 1.26.7 honors ResponseHeaderTimeout on HTTP/2 connections as well as HTTP/1.1. Two items remain: TestNewKoiosTransportDialTimeout can pass without exercising the dial bound when a proxy is configured, and the http.DefaultTransport type assertion drops the comma-ok form the repository already uses for the same clone.

internal/koiosparity/koios_client_test.go
```diff
@@ -557,3 +557,128 @@ func TestGetAccountRewardHistoryEmptyAddressesNoRequest(t *testing.T) {
...
+	if resp != nil {
+		resp.Body.Close()
+	}
+
+	require.Error(t, err, "a black-holed dial target must fail the request")
+	assert.Less(
+		t,
+		elapsed,
+		clientTimeout,
+		"must fail well before the unrelated, much larger client-level Timeout",
```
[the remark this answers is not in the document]
chrisguiney: Fixed in 6881861: the test now requires a net.Error with Timeout() true (verified: dialing a closed local port instead returns a non-timeout "connection refused" with Timeout() false) and tightens the elapsed bound to 2*dialTimeout (600ms) instead of only excluding the 10s clientTimeout.

internal/koiosparity/koios_client.go
chrisguiney: Fixed in 6881861: newKoiosTransport now uses the comma-ok form with a `&http.Transport{Proxy: http.ProxyFromEnvironment}` fallback, matching mithril/download.go's newDownloadTransport. Added TestNewKoiosTransportFallsBackWhenDefaultTransportIsNotHTTPTransport, which swaps http.DefaultTransport for a non-*http.Transport RoundTripper and proves newKoiosTransport returns a usable transport with the configured timeouts instead of panicking (confirmed it panics against the prior bare assertion).

internal/koiosparity/koios_client_test.go
chrisguiney: Fixed in 6881861: the test now clears HTTP_PROXY/http_proxy/HTTPS_PROXY/https_proxy/NO_PROXY/no_proxy via t.Setenv for its duration, and the added net.Error Timeout()==true assertion independently rejects the fast-refusal/ENETUNREACH case. Verified: with the old assertions and HTTP_PROXY pointed at an unreachable local port, the test passed in under a millisecond without ever exercising the dial timeout; with the new assertions it fails on the Timeout() check, and with the proxy variables cleared it passes correctly at ~300ms regardless of an ambient HTTP_PROXY.

wolf31o2 (approved):

PR #4315, opened 2026-09-15
test(snapshot): close reward-snapshot denominator coverage gaps

### LINKED ISSUE #4027: ledger/snapshot: reward denominator follow-ups — stale comments and two coverage gaps
Non-blocking review findings from #3969. Test-coverage gaps and comments the change invalidated.

### Comments that now contradict the code

Neither is in #3969's diff, and both are the next thing a reader hits when auditing the denominator.

- `ledger/snapshot/rotation.go:544-547` states "Real Cardano reward semantics already exclude a pool with no resolvable registration from the active reward stake". `resolveActiveInstantStakeCredentials` (`libs/cardano-ledger-core/src/Cardano/Ledger/State/Stake.hs`) keys off the credential being registered and carrying a pool delegation, and never consults `psStakePools` — which is the premise #3969 establishes.
- `ledger/snapshot/rotation.go:553-554` states "The returned distribution is the one actually consumed to build poolInputs/stakeInputs, so its totals are safe to use for RewardSnapshot". After #3969 that holds for `TotalPoolCount` and `TotalDelegators` only.

### Test-coverage gaps

- `TestRewardCalculatorInputsAllowExcludedPoolStake` (`ledger/reward_calculation_test.go:6030`) varies `TotalActiveStake` against a fixed pool/stake-input set, but there is no equivalent negative case varying `TotalDelegators` (checked at `ledger/reward_calculation.go:1415` and `:3086`). That comparison is now the primary structural guard replacing the exact stake-sum comparison, and only pool count has dedicated coverage.
- `ledger/snapshot/reward_active_stake_test.go:46` does not assert the companion claim `DATABASE.md` makes for the same boundary — that `pool_stake_snapshot`/`epoch_summary` still carry the degraded pool's full stake even though `reward_pool_input` excludes it. It holds in code (`rotation.go:218-276` reads the unmodified `distribution`) but is untested at this boundary.

### Wording

- `DATABASE.md:983` — "covers every delegating credential observed at the boundary" overstates it. `GetLiveStakeInputsForPools` filters `registered = TRUE`, restricts to pools active at the slot (`getActivePoolsAtSlot`), and applies the optional account-expiration predicate. "Every registered, non-expired credential delegating to a pool active at the snapshot slot" is exact.
- `ledger/snapshot/rotation.go:440` recomputes `sumPoolStakes(rewardDistribution.PoolStakes)` without an overflow check, when `rewardDistribution.TotalStake` is the identical value computed a few lines above with one. Safe as written, since `rewardStakeDistribution` errors on overflow before returning.

arepala-uml:
## Summary

Non-blocking follow-ups from #3969, reported in #4027:

- Add a `TotalDelegators` negative-case test alongside the existing `TotalActiveStake` exclusion-slack test in `TestRewardCalculatorInputsAllowExcludedPoolStake` (`ledger/reward_calculation_test.go`).
- Add a test pinning that `pool_stake_snapshot` and `epoch_summary` keep a degraded pool's full stake even though `reward_pool_input` excludes it (`ledger/snapshot/reward_active_stake_test.go`).
- Fix `rewardInputsSkippingDegradedPools`'s doc comment in `ledger/snapshot/rotation.go`, which contradicted both real Cardano semantics (`resolveActiveInstantStakeCredentials` never consults the stake-pool set) and the code's own behavior for `TotalActiveStake`.
- Correct `DATABASE.md`'s `reward_snapshot.total_active_stake` wording to match `GetLiveStakeInputsForPools`'s actual filter.

`rotation.go:440`'s missing overflow check on `sumPoolStakes` was also reviewed per the issue and confirmed safe as written (`rewardStakeDistribution` already errors on overflow before returning) — no change needed there.

Closes #4027

## Test plan

- [x] `go build ./...`
- [x] `go vet ./ledger/...`
- [x] `golangci-lint run ./ledger/...` (0 issues)
- [x] `modernize ./ledger/...`, `nilaway ./ledger/...` (no findings on changed files)
- [x] `gofmt -l`, `golines --dry-run` (clean)
- [x] `go test ./ledger/... -race` (all packages pass)
- [x] `go test ./internal/test/conformance/...`
- [x] `go test ./internal/docsparity/...`
- [x] Verified both new tests fail without the underlying behavior (mutation-tested)

---
## Summary by cubic
Closes #4027 by adding tests that pin how degraded pools are handled in reward snapshots and fixing the docs/comments that misdescribed it.

- `reward_pool_input` drops degraded pools, but `total_active_stake`, `pool_stake_snapshot`, and `epoch_summary` keep their full stake.
- `TotalDelegators` must match the `reward_pool_input` rows exactly; only `TotalActiveStake` has exclusion slack.
- The `rewardInputsSkippingDegradedPools` comment now states `TotalActiveStake` comes from the pre-exclusion distribution.
- `DATABASE.md` now matches `GetLiveStakeInputsForPools`'s actual filter.
- The missing overflow check on `sumPoolStakes` was reviewed and confirmed safe; no code change needed.

<sup>Written for commit c691d6b0bc794fc0fb4f8dbd6917eb172dee4b83. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4315?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected reward accounting so delegators associated with pools lacking resolvable registrations continue contributing to total active stake and denominator calculations.
  - Such pools remain represented in stake and epoch summaries while being excluded from pool reward distribution.

- **Documentation**
  - Clarified how total active stake is calculated, including delegations to pools excluded from reward inputs.

- **Tests**
  - Added coverage for delegator-count validation and degraded-pool stake preservation.


cryptodj413 (approved):
Test/doc-only follow-up (no production behavior change) — reviewed by direct code tracing rather than delegating to a domain subagent, since the diff is small and self-contained.

Verified rather than trusted the PR description:

- **rotation.go comment fix**: traced `TotalActiveStake`'s actual data path — `buildRewardStateInputs` computes it from `sumPoolStakes(rewardDistribution.PoolStakes)`, the pre-exclusion aggregate returned by `rewardStakeDistribution`, before `rewardInputsSkippingDegradedPools` strips a degraded pool from a separate copy used only for `TotalPoolCount`/`TotalDelegators`. The corrected comment matches this exactly.
- **DATABASE.md wording**: traced `GetLiveStakeInputsForPools`'s call site in `calculator.go` — it's filtered by `getActivePoolsAtSlot` (active pools) and `expiryEpoch` (non-expired), matching the new "every registered, non-expired credential delegating to a pool active at the snapshot slot" wording.
- **New test `TestSaveSnapshotKeepsDegradedPoolStakeInPoolAndEpochRows`**: ran it (pass), then mutation-tested it — changed `epoch_summary.TotalActiveStake`'s source to a hardcoded 0 and confirmed the test fails (`expected 0x2540be400, actual 0x0`), then reverted. The pool_stake_snapshot half is an unconditional range over the caller's untouched `distribution.PoolStakes` map with no exclusion logic in that loop, confirmed by reading it directly.
- **New test `TestRewardCalculatorInputsRejectsDelegatorCountMismatch`**: confirmed the exact error string it asserts on exists verbatim in `reward_calculation.go:3233`, not fabricated.
- CI: confirmed via job logs (not just the checkmark) — `internal/docsparity` ok, `ledger` (race) ok 322s, `ledger/snapshot` ok — covering exactly the packages this PR touches.
- ARCHITECTURE.md: correctly untouched — no component-boundary, EventBus, or lifecycle change here.
- No bot findings to reconcile: CodeRabbit completed with 0 actionable comments; Cubic shows no review at this head.

Approving.

PR #4316, opened 2026-09-15
test: made recovery, blocking, and analysis checks deterministic

### LINKED ISSUE #3532: test: make recovery, blocking, and analysis checks deterministic
Several tests and analysis settings can pass without proving the intended behavior:

- Recovery cases use state that does not enter the branch named by the test.
- Blocking tests use short periods of non-receipt as proof that work is blocked.
- Analysis duration settings accept values that can overflow or create impractical durations after multiplication.

Acceptance criteria:

- Drive the intended recovery branch and prove the assertion fails without it.
- Replace timing-only blocking assertions with explicit synchronization.
- Validate duration values before multiplication and reject unsupported ranges.
- Add boundary tests for the duration parser.

jkawan:
closes #3532 

---
## Summary by cubic
Makes recovery, blocking, and analysis checks deterministic by replacing timing windows with explicit synchronization and state checks, and tightens validation of the Antithesis analysis timing settings.

- Waits for publishers to start, subscriber buffers and ordered lane queues to fill before asserting backpressure.
- Proves blocked deliveries and recovery paths actually ran using blocked-delivery metrics, `onBlocked` hooks, resync events, and a rewind-step hook that deterministically extends the chain.
- Caps `ANALYSIS_INITIAL_WAIT` and `ANALYSIS_CHECK_INTERVAL` at 24 hours; accepts a zero initial wait and rejects malformed, negative, zero-interval, or overflowing values.

<sup>Written for commit 64d0d5e40ee7e26dc84568db9dff65be1226021c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4316?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of analysis timing settings, including support for zero initial wait times and limits on excessively large values.
  * Expanded recovery verification to confirm correct chain rewinding and resynchronization after validation and replay issues.

* **Tests**
  * Strengthened coverage for backpressure, ordered publishing, exactly-once delivery, and recovery scenarios.
  * Replaced timing-sensitive checks with deterministic synchronization for more reliable verification of blocking and delivery behavior.


event/ordered_test.go
```diff
@@ -156,11 +156,22 @@ func TestPublishOrderedWaitsForCapacityRatherThanDropping(t *testing.T) {
...
+	// the lane's queue to actually fill rather than assuming it will within
+	// an arbitrary window -- that is what proves the publisher is blocked on
+	// capacity, not merely slow.
+	lane := eb.orderedLane("ordered.full")
+	deadline := time.Now().Add(2 * time.Second)
+	for len(lane.queue) != cap(lane.queue) {
+		if time.Now().After(deadline) {
+			t.Fatal("the ordered lane never filled up while nothing drained it")
+		}
+		time.Sleep(time.Millisecond)
```
wolf31o2: This polling loop uses time.Sleep for synchronization. Replace it with WaitForCondition or another event-driven synchronization mechanism; Dingo tests must not use time.Sleep for synchronization.

wolf31o2 (changes requested):
Replace the time.Sleep-based polling in event/ordered_test.go with the repository test utility (for example WaitForCondition) or another event-driven synchronization mechanism. Dingo tests must not use time.Sleep for synchronization, and the current loop can still be scheduler-sensitive.

event/backpressure_test.go
```diff
@@ -40,15 +40,18 @@ func TestDeliverWaitsForCapacityThenDelivers(t *testing.T) {
...
 	require.NoError(t, sub.Deliver(NewEvent("test", "first")))
 
+	blocked := make(chan struct{})
+	sub.onBlocked = func() { close(blocked) }
 	done := make(chan error, 1)
 	go func() {
 		done <- sub.Deliver(NewEvent("test", "second"))
 	}()
-
+	<-blocked
```
chrisguiney, line 49: Unbounded receive. Removing deliverWait's full-buffer wait in event/event.go makes this line hang until the package -timeout panics; the pre-change version of this test failed in 0.00s with "Deliver returned while the buffer was full". A real drop regression now reads as infrastructure trouble and takes every parallel test in the package down with it. Same shape at backpressure_test.go:84 and :129, delivery_test.go:75, race_test.go:379. Use testutil.RequireReceive(t, blocked, testutil.AsyncWait, "delivery never reached the full-buffer wait").

event/race_test.go
```diff
@@ -356,23 +356,32 @@ func TestPublishBlocksOnFullChannelUntilDrained(t *testing.T) {
...
 	for i := range buffer {
 		eb.Publish(typ, NewEvent(typ, i))
 	}
 
+	// Synchronize on deliverWait's full-buffer wait path rather than on the
+	// overflow goroutine merely starting: the goroutine can pause after
+	// close(started) but before Publish reaches its blocking send, which
+	// would let the checks below pass without proving delivery blocked.
+	blocked := make(chan struct{})
+	eb.channelSubsById[subId].onBlocked = func() { close(blocked) }
```
chrisguiney, line 371: onBlocked is documented at event/event.go:279 as set once at subscribe time, before the subscriber is reachable by any publisher, and deliverWait reads it under c.mu.RLock. This writes it with no lock after 64 publishes have already run, and reads the e.mu-guarded channelSubsById map without e.mu; it holds only because no publisher is in flight at this instant. Move both lines directly below SubscribeWithBuffer: the fill loop exactly fills the buffer, so the hook cannot fire early and the documented invariant survives.

event/backpressure_ext_test.go
```diff
@@ -152,6 +154,14 @@ func TestPublishAsyncBlocksWhenQueueFull(t *testing.T) {
 		enqueued <- true
 	}()
 
+	testutil.RequireReceive(
+		t, started, time.Second, "publisher did not start",
+	)
+	require.Eventually(t, func() bool {
+		return len(subCh) == cap(subCh)
+	}, time.Second, time.Millisecond,
```
chrisguiney, line 162: CLAUDE.md specifies a 2-5s timeout and a 5-10ms interval for require.Eventually. This is 1s at 1ms, as is line 356, while the ordered-lane wait added in the same change uses testutil.AsyncWait. A 1s budget for the subscriber buffer to fill on a loaded go-test (Windows) runner is the flake class this change exists to remove. Raise both to testutil.AsyncWait, or to 5s at 10ms.

ledger/recovery_rewind_window_test.go
```diff
@@ -204,6 +219,16 @@ func TestWindowedRewindConvergesWhilePrimaryChainExtends(t *testing.T) {
...
 		"a descent that reached its target committed its steps",
 	)
+	// Prove the race this test exists for was actually exercised: the
+	// appender must have landed at least one append while the rewind was
+	// actually in flight, not merely at some point before close(stop) (a
+	// rewind's own rollback shrinks the tip, so comparing tip block numbers
+	// before and after cannot show this).
+	require.Positive(
+		t,
+		appendDuringRewind.Load(),
```
chrisguiney, line 229: This asserts on the scheduler rather than on the code: nothing forces the appender goroutine to land an append between rewindInProgress.Store(1) and Store(0), so a host that never schedules it inside that window fails with "0 is not positive" -- the signature of an unsynchronised read, not of a regression. It held over 10 runs under -race at this head. A hook the windowed rollback calls per step, driving one append, makes the overlap a stated contract instead of a probability.

internal/test/antithesis/internal/analysis/config_test.go
```diff
@@ -0,0 +1,66 @@
...
+			require.Error(t, err)
+		})
+	}
+
+	got, err = parseAnalysisDuration("ANALYSIS_INITIAL_WAIT", "0", true)
+	require.NoError(t, err)
+	require.Zero(t, got)
+}
+
+func TestLoadConfigAnalysisDurations(t *testing.T) {
```
chrisguiney, line 58: t.Setenv makes this test process-global, so it can never take t.Parallel. CLAUDE.md requires a "// Not t.Parallel: ..." comment on a test kept sequential for that reason, and TestParseAnalysisDurationBounds directly above is parallel, so the asymmetry reads as an oversight rather than a decision. Add the comment naming the process-global reason.

chrisguiney (changes requested):
wolf31o2's time.Sleep finding on event/ordered_test.go is resolved at this head: the polling loop is now testutil.WaitForCondition and no changed file contains time.Sleep. CodeRabbit's three comments from ea1c2aa8 are addressed, and the new header-validation resync assertion does fail when that publish is removed. Blocking: the five new bare receives on the onBlocked signal turn the drop regression they guard into a package-wide timeout panic instead of a named assertion.

wolf31o2 (approved):

chrisguiney (approved):

PR #4319, opened 2026-09-15
feat(nodeparity): Koios-backed node-parity from-genesis validation (#1900)

arepala-uml:
## Summary

A real cardano-node can't fill the reference role for validating a
from-genesis Dingo replay: its own replay races ahead too fast, leaving no
matching historical block to compare against by the time Dingo reaches it
(confirmed live: tens of thousands of blocks ahead at the very first epoch
boundary). Koios has no such problem -- it retains full per-epoch history
indefinitely. Adds a new `node-parity from-genesis` subcommand that follows
a Dingo instance's chain from genesis and, at every epoch boundary,
compares protocol parameters, stake distribution, and the full UTxO set
(address, amount, assets, datum, reference script -- not just existence)
against Koios.

This branch includes the commits from #4320 (blinklabs-io/dingo#382):
`from-genesis` relies on that PR's Acquire-time validation to fail cleanly
at Acquire rather than mid-query once a point ages past a query type's own
retention floor. The two PRs are independent (neither is based on the
other's branch), so this diff currently shows #4320's commits too; once
#4320 merges to `main` and this branch is rebased, they'll disappear from
here automatically.

## Found live, fixed in this branch

Running this tool against a real, freshly-syncing Dingo instance on preview
surfaced several real bugs and performance issues, each fixed here:

- UTxO reconstruction had no rollback handling at all -- a routine
  occurrence during genesis bulk replay against multiple peers -- silently
  corrupting the running UTxO set with no recovery.
- Protocol-params era was never populated for a live LocalStateQuery caller
  (`ProtocolParamsFromNative`), causing a spurious mismatch on every single
  epoch; also disambiguated Shelley vs. Allegra (indistinguishable by Go
  type alone in gouroboros) via `HardForkCurrentEraQuery`.
- The UTxO reconstruction fetched Koios `/tx_info` once per block; batched
  and parallelized across blocks instead, since per-call network latency
  (not payload size) dominated once real chain activity picked up.
- The stake-distribution check fetched Koios `/pool_history` once per pool,
  sequentially (~1.5s/call x 75 pools = ~2 min/epoch on preview);
  parallelized with bounded concurrency.
- The Koios HTTP client's long idle-connection and request timeouts (90s /
  60s defaults) turned an occasional stale-connection or mirror hiccup into
  a 60s+ stall; shortened both.

## CodeRabbit review response

CodeRabbit flagged 4 `Major` findings against this PR. 3 are fixed; 2 are
[...]

PR #4320, opened 2026-09-15
fix(ledger,ouroboros): reject an unqueryable Acquire point up front

arepala-uml:
## Summary

Dingo's `Acquire` previously accepted a LocalStateQuery point that any
point-aware query type's own retention floor had already passed, only
failing later at query time -- or, in `HardForkCurrentEraQuery`'s case,
silently ignoring the pinned point entirely and answering at live tip
instead. Real cardano-node's own historical retention is a single uniform
window shared by every query type, so it never needs this; Dingo's windows
are independently sized per query type, so a point that survives Acquire
must actually be answerable by every one of them.

- `HardForkCurrentEraQuery` now resolves the era at the pinned point the
  same way `queryShelleyCurrentProtocolParams` already does, instead of
  always answering at live tip regardless of what was Acquired.
- `Acquire` itself now rejects an unqueryable point up front
  (`LedgerState.VerifyPointQueryable`, run before ever recording a point as
  acquired), instead of failing confusingly mid-query later -- the
  Ouroboros LocalStateQuery wire protocol has no way to fail an individual
  query after a successful Acquire (`Querying` state's only transition is
  `Result`, never `Failure`), so a point Acquire allowed but a later query
  couldn't actually answer previously just dropped the connection
  ("protocol is shutting down" client-side) instead of a clean
  `AcquireFailure`.
- The from-tip UTxO retention floor is skipped while a node is still
  catching up to the chain tip, so `Acquire` doesn't spuriously reject
  valid recent points during initial sync.
- Pool-stake snapshots are retained without bound in `--storage-mode api`,
  matching the existing UTxO carve-out for that mode.

## CodeRabbit review response

CodeRabbit flagged a `Major` finding on this PR (surfaced on #4319, which
included this diff before the two PRs were un-stacked): a narrow race
between `VerifyPointQueryable`'s Acquire-time check and Dingo's
unsynchronized background retention pruning (`cleanupConsumedUtxos` runs as
`go ls.cleanupConsumedUtxos()`, not serialized against Acquire in any way).
Confirmed real. Not fixed here: the proper fix is a cross-cutting feature
-- every relevant pruning path (UTxO cleanup, snapshot rotation, pparams
retention) needs to know about and defer to currently-acquired points until
Release/ReAcquire/disconnect, plus a mitigation for a client that acquires
and never releases -- not a fix scoped to `VerifyPointQueryable` alone.
Documented in code and tracked in blinklabs-io/dingo#4322 rather than
rushed.

## Test plan

- `go build ./...`, `go vet ./...`, `golangci-lint run`, `gofmt` -- all
  clean on the touched packages (`ledger`, `ledger/snapshot`,
  `ouroboros`).
- `go test -race ./ledger/... ./ouroboros/...` -- all green.
- Verified live: a point older than the strictest per-type retention floor
  now fails cleanly at `Acquire` ("acquire failure: point too old") instead
  of a later query dropping the connection.

## Known follow-ups (not blocking)

- blinklabs-io/dingo#4322 -- narrow Acquire-time-validation race against
  background retention pruning (see above).


ledger/queries.go
```diff
@@ -418,12 +499,74 @@ func (ls *LedgerState) queryChainPoint() (any, error) {
...
 	case *olocalstatequery.HardForkCurrentEraQuery:
-		return ls.loadConsensusSnapshot().currentEra.Id, nil
+		if !at.pinned() {
+			return ls.loadConsensusSnapshot().currentEra.Id, nil
+		}
+		if txn == nil {
+			txn = ls.db.Transaction(false)
+			defer txn.Release()
+		}
+		targetEpoch, err := ls.resolveAsOfEpoch(txn, at)
```
chrisguiney: resolveAsOfEpoch returns (0, nil) when GetEpochBySlot finds no row covering at.Slot, since its SQL requires start_slot <= slot AND slot < start_slot + length_in_slots. GetEpoch(0) below then succeeds on any genesis-synced node, so GetCurrentEra answers Byron for a point in a later era. Reproduced against this tree: TestQueryHardFork_CurrentEra_NoEpochRecordRejected's fixture plus an epoch-0 row returns era 0 with a nil error while the live era is Conway. That test passes only because its fixture omits the epoch-0 row. Return a found flag from resolveAsOfEpoch and reject before GetEpoch.

ledger/queries.go
```diff
@@ -287,6 +288,86 @@ func (ls *LedgerState) queryShelleyEpochNo(
...
+// pparams retention -- could in principle advance a retention floor past at
+// in the gap between this function returning and the point being recorded,
+// leaving a query against an already-acquired point exposed to the exact
+// mid-query failure this whole mechanism exists to prevent. Acquire-time
+// validation alone cannot close this: doing so needs every relevant pruning
+// path to know about and defer to currently-acquired points until Release,
+// ReAcquire, or disconnect -- a cross-cutting feature spanning three
+// independent pruning subsystems, not a fix scoped to this function.
+// Deferred rather than rushed; tracked as a follow-up issue.
+func (ls *LedgerState) VerifyPointQueryable(
```
chrisguiney, line 376: This runs verifyPointOnChain, checkUtxoRetentionWindow, PoolStakeDistribution and queryShelleyCurrentProtocolParams, but not queryHardFork, which this same diff gives two new ErrHistoricalStateUnavailable returns. Across every ErrHistoricalStateUnavailable site in ledger/, those two are the only ones reachable from a pinned Query that Acquire does not exercise, so the doc claim above that every point-aware query type is covered does not hold. Querying has no Failure transition, so a GetCurrentEra rejection after a successful Acquire drops the connection.

ledger/queries.go
```diff
@@ -287,6 +288,86 @@ func (ls *LedgerState) queryShelleyEpochNo(
...
+		txn = ls.db.Transaction(false)
+		defer txn.Release()
+	}
+	if err := ls.verifyPointOnChain(txn, at); err != nil {
+		return err
+	}
+	if err := ls.checkUtxoRetentionWindow(txn, at); err != nil {
+		return err
+	}
+	if _, err := ls.PoolStakeDistribution(nil, at, txn); err != nil {
```
chrisguiney: The unfiltered form reads the whole mark snapshot via markStakeByPool, does a bulk registration lookup over every pool in it via poolVrfKeyHashes, sorts the key hashes, allocates a cbor.Rat per pool and runs totalCirculatingSupply's as-of-slot reconstruction. All of it is discarded. That is thousands of pools of work on mainnet on every pinned Acquire, and a client that then issues GetPoolDistr2 pays it twice. The floors this call tests are checkAsOfEpochRecency plus the epoch-row lookup, both callable directly.

ouroboros/localstatequery.go
```diff
@@ -145,31 +163,34 @@ func (o *Ouroboros) localstatequeryServerAcquire(
...
+		// stricter retention floor (UTxO whole/by-ref, stake/pool
+		// distribution, current protocol parameters at a historical
+		// epoch) -- confirmed live, this hit the exact same
+		// connection-killing gap #4156 fixed for the on-chain check alone,
+		// just for a different, retention-based rejection reason
+		// (blinklabs-io/dingo#382 protocol-compliance finding).
+		// VerifyPointQueryable's own doc comment covers verifyPointOnChain
+		// too, so this subsumes VerifyPointOnChain rather than needing
+		// both checks run separately.
+		if err := o.ledgerState.VerifyPointQueryable(nil, point); err != nil {
```
chrisguiney, line 186: This is a wire-visible Acquire behaviour change with no test: VerifyPointQueryable has no reference in any _test.go in the repository. Cover both directions, because only the accept direction keeps working clients alive: a point past a retention floor maps to ErrAcquireFailurePointTooOld, and a point inside every floor is still accepted. An error matching neither sentinel is also returned bare here, and gouroboros handleAcquire returns it rather than sending MsgFailure, so a database error inside the new PoolStakeDistribution and pparams reads now tears the connection down.

ledger/snapshot/rotation.go
```diff
@@ -1104,70 +1104,88 @@ func (m *Manager) cleanupOldSnapshots(
...
+	// older epoch the same way an unbounded UTxO table does for
+	// GetUTxOWhole. Confirmed live (blinklabs-io/dingo#1900 node-parity
+	// validation): with only UTxO's own window removed, a historical
+	// Acquire still failed at exactly this pool-snapshot floor instead,
+	// since VerifyPointQueryable's Acquire-time gate (blinklabs-io/dingo#382)
+	// checks stake retention on every Acquire regardless of which query
+	// type the caller actually intends to ask -- there is no protocol-level
+	// way to know that in advance, so the whole Acquire is only as
+	// long-lived as its strictest per-query floor.
+	if m.db.StorageMode() != types.StorageModeAPI {
```
chrisguiney, line 1123: DATABASE.md is now wrong in three places: the claim that reward_account_output is the one table whose retention depends on node configuration; the Pruned at epoch < current-3 table, whose pool_stake_snapshot row carries no API-mode qualifier where the reward_account_output row carries one; and the statement that pool_stake_snapshot is pruned to currentEpoch - 3, which is the stated reason the pool-departure query is the evidence available to an observer running behind the node. No test covers this branch or the matching apiStorageMode skip in checkAsOfEpochRecency.

ledger/state.go
```diff
@@ -3194,6 +3194,23 @@ func (ls *LedgerState) cleanupConsumedUtxos() {
...
 }
 
+// utxoPruningDeferredForCatchup reports whether cleanupConsumedUtxos would
+// defer a run at tipSlot/stabilityWindow right now, mirroring its own two
+// defer conditions above (unknown upstream target, or known but not yet
+// near it) without duplicating its logging. checkUtxoRetentionWindow uses
+// this so its own "fresh from-tip" retention estimate only applies when
+// cleanup is actually keeping pace with the tip -- see that function's doc
+// comment for the false-rejection this closes.
+func (ls *LedgerState) utxoPruningDeferredForCatchup(
```
chrisguiney, line 3258: No test references utxoPruningDeferredForCatchup. It is the one change here that widens what Acquire accepts, so a mirror that drifts from cleanupConsumedUtxos' own two defer conditions fails open rather than closed. Pin both branches: deferred while upstreamActive is true and upstreamTip is 0, and deferred while upstreamTip is known and nearUpstreamTip is false, with checkUtxoRetentionWindow accepting a point below tip minus the stability window in each and rejecting it once caught up.

chrisguiney (changes requested):
GetCurrentEra at a pinned point silently answers with epoch 0's era when no epoch row covers the slot, reproduced against this branch, and VerifyPointQueryable never exercises queryHardFork so Acquire cannot see that path. Build, vet, scoped golangci-lint and -race across ledger, ledger/snapshot and ouroboros are clean at d6adcb7, and both new tests fail when the queryHardFork change is reverted in place. go-test (Windows) was cancelled by its own 90-minute backstop; the same signature appears twice on main today and the workflow declares the job advisory.

ledger/queries.go
arepala-uml: Fixed in fd2aa0ce: resolveAsOfEpoch now returns a found flag, and every caller (this one, queryShelleyCurrentProtocolParams, queryShelleyStakeDistribution, PoolStakeDistribution, queryShelleyEpochNo) rejects with ErrHistoricalStateUnavailable when it's false instead of proceeding with epoch 0. Strengthened TestQueryHardFork_CurrentEra_NoEpochRecordRejected to seed an epoch-0 row too, exactly as you describe -- confirmed it was passing for the wrong reason before.

ledger/queries.go
arepala-uml: Fixed in fd2aa0ce: VerifyPointQueryable now also runs queryHardFork's HardForkCurrentEraQuery case.

ledger/queries.go
arepala-uml: Fixed in fd2aa0ce: added verifyStakeDistributionRetentionOnly, checking the same floor via resolveAsOfEpoch + checkAsOfEpochRecency directly without materializing the snapshot, VRF lookup, or circulating-supply reconstruction.

ouroboros/localstatequery.go
arepala-uml: Added both direction tests in fd2aa0ce (TestVerifyPointQueryable_WithinAllFloors_Accepted / _PastRetentionFloor_Rejected). Fixed the bare-error case in ad4abf85: mapped to AcquireFailurePointTooOld instead of returned raw, with the real error logged server-side since the client only sees the generic wire-level rejection.

ledger/snapshot/rotation.go
arepala-uml: Fixed in ed934d4b (all three spots) and a157eb35 (new test coverage for both the CORE and API-mode pool_stake_snapshot retention branches, extending the existing reward_account_output retention tests).

ledger/state.go
arepala-uml: Added in 45d70f44, pinning exactly both conditions plus the two must-not-defer cases (no upstream tracked; caught up to a known target).

ledger/queries.go
```diff
@@ -418,12 +499,74 @@ func (ls *LedgerState) queryChainPoint() (any, error) {
...
+		}
+		if txn == nil {
+			txn = ls.db.Transaction(false)
+			defer txn.Release()
+		}
+		targetEpoch, err := ls.resolveAsOfEpoch(txn, at)
+		if err != nil {
+			return nil, err
+		}
+		epochRow, err := ls.db.GetEpoch(targetEpoch, txn)
```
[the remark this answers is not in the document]
arepala-uml: Fixed in fd2aa0ce: resolveAsOfEpoch now returns a found flag distinct from a genuine epoch-0 resolution, and queryHardFork rejects with ErrHistoricalStateUnavailable before ever calling GetEpoch when found is false. Added the exact regression case requested -- TestQueryHardFork_CurrentEra_NoEpochRecordRejected now seeds a real epoch-0 row plus a validated pinned point at an uncovered slot (50), so it exercises the silent-wrong-era path instead of passing by omission as it did before.

ouroboros/localstatequery.go
```diff
@@ -145,31 +163,34 @@ func (o *Ouroboros) localstatequeryServerAcquire(
...
 			if errors.Is(err, ledger.ErrPointNotOnChain) {
 				return fmt.Errorf(
 					"%w: %w",
 					olocalstatequery.ErrAcquireFailurePointNotOnChain,
 					err,
 				)
 			}
```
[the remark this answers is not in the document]
arepala-uml: Same finding as this thread's sibling and Chris Guiney's independent review comment on the same line -- confirmed real (cleanupConsumedUtxos runs as an unsynchronized background goroutine), documented in code, and tracked as a follow-up in https://github.com/blinklabs-io/dingo/issues/4322 rather than fixed here. Agreed a lease-based design (register the pending point before validation, honor it in every cleanup path, bound an unreleased client) is the right shape for that follow-up.

ledger/queries.go
```diff
@@ -287,6 +288,86 @@ func (ls *LedgerState) queryShelleyEpochNo(
...
+	}
+	if err := ls.checkUtxoRetentionWindow(txn, at); err != nil {
+		return err
+	}
+	if _, err := ls.PoolStakeDistribution(nil, at, txn); err != nil {
+		return err
+	}
+	if _, err := ls.queryShelleyCurrentProtocolParams(at, txn); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
arepala-uml: Verified against the current code: VerifyPointQueryable (fd2aa0ce) now unconditionally runs queryHardFork's HardForkCurrentEraQuery case as one of its checks -- and that path has no live-epoch fast path of its own, it always does GetEpoch(targetEpoch) + eras.GetEraById(epochRow.EraId) regardless of whether targetEpoch equals the live epoch. So a point whose epoch row names an unknown EraId is already rejected by VerifyPointQueryable as a whole via that check, even though queryShelleyCurrentProtocolParams's own fast path still skips era validation in isolation. The aggregate guarantee (Acquire succeeding implies every point-aware query type will succeed) holds today.

I'd still call the underlying observation fair: queryShelleyCurrentProtocolParams's fast path relies on that guarantee coming from a *different* function in the same aggregate rather than validating independently, which is a bit fragile to a future refactor that touches VerifyPointQueryable's check list without noticing the dependency. I'm leaving it as-is for this PR rather than adding a second era-validation call on every fast-path hit (a live query cross-checking its own live epoch's era is close to tautological, and the real risk this thread identifies is already closed) -- but happy to factor out a shared helper if you'd rather have both paths validate independently regardless of aggregate composition.

ledger/pool_stake_distribution.go
```diff
@@ -84,6 +97,38 @@ func checkAsOfEpochRecency(targetEpoch, liveEpoch uint64) error {
...
+		return err
+	}
+	if !found {
+		return errEpochNotResolved(at)
+	}
+	liveEpoch, _, err := ls.resolveAsOfEpoch(txn, QueryPoint{})
+	if err != nil {
+		return err
+	}
+	return checkAsOfEpochRecency(
```
chrisguiney: Blocker: this floor is narrower than PoolStakeDistribution enforces, so Acquire now accepts a point GetPoolDistr2 and GetStakeDistribution still reject. Both call totalCirculatingSupply with asOfSlot=at.Slot, which returns ErrHistoricalStateUnavailable when no network_state row covers that slot. Reproduced on a fixture VerifyPointQueryable accepts: both queries fail with that sentinel; restoring the previous PoolStakeDistribution call in place makes VerifyPointQueryable reject it. handleQuery returns that error bare, so the connection drops. Add the one-row as-of-slot read here.

ledger/queries.go
```diff
@@ -280,13 +306,112 @@ func (ls *LedgerState) queryShelleyEpochNo(
...
+	// queryHardFork's HardForkCurrentEraQuery case (GetCurrentEra) is
+	// point-aware and returns ErrHistoricalStateUnavailable when
+	// resolveAsOfEpoch can't resolve at to an epoch (human review,
+	// dingo#4319/#4320) -- exercised here for the same reason every other
+	// check above is: Querying has no Failure transition, so a rejection
+	// surfacing from Query instead of here drops the connection rather
+	// than returning a clean AcquireFailure.
+	if _, err := ls.queryHardFork(
```
chrisguiney, line 406: This queryHardFork call closes the earlier gap, but nothing pins it: deleting these six lines in place leaves TestVerifyPointQueryable_WithinAllFloors_Accepted and _PastRetentionFloor_Rejected green, because neither fixture has an epoch row that resolves while its era does not. Add a case where queryHardFork is the only check that rejects -- a pinned slot whose epoch row names an unknown EraId, say -- so this call cannot be dropped silently.

ouroboros/localstatequery.go
```diff
@@ -145,39 +163,65 @@ func (o *Ouroboros) localstatequeryServerAcquire(
...
+			// handleAcquire treat it as a fatal protocol error and tear
+			// down the connection, reintroducing the exact
+			// connection-killing failure mode this whole mechanism exists
+			// to avoid, just triggered by a different kind of error. Map
+			// it to the same AcquireFailurePointTooOld a well-behaved
+			// client already knows how to handle (retry against a
+			// different point) instead, logging the real error here since
+			// the client only ever sees the generic wire-level rejection.
+			o.config.Logger.Error(
+				"local-state-query Acquire validation failed unexpectedly",
```
chrisguiney, line 215: The neither-sentinel mapping is wire-visible and untested: localstatequery_test.go covers only ErrAcquireFailurePointNotOnChain, so nothing asserts that a non-sentinel VerifyPointQueryable error becomes AcquireFailurePointTooOld rather than a torn-down connection, and nothing covers the ErrHistoricalStateUnavailable branch above it either. A server-side test with a ledger stub returning a plain error would pin both branches.

chrisguiney (changes requested):
Re-review of the five commits since the last decision. Four earlier findings are fixed and now causally covered: the resolveAsOfEpoch found flag, utxoPruningDeferredForCatchup both conditions, the API-mode pool_stake_snapshot carve-out and the DATABASE.md retention text -- reverting each fix in place fails its new test. One fix reopens the connection-killing path it replaced, and two behaviour changes have no test that fails without them. Checks pass at head but predate current main; the merged tree builds and the affected packages pass.

ledger/pool_stake_distribution.go
arepala-uml: Fixed in 6ce3f42c: added the missing second floor -- a `network_state` row at or before the pinned slot must exist, matching exactly what `totalCirculatingSupply`'s `asOfSlot` path separately requires. Reproduced the gap first (confirmed `TestVerifyPointQueryable_WithinAllFloors_Accepted`'s own fixture failed once seeding a covering row, proving it needed one all along), then added `TestVerifyPointQueryable_NoNetworkStateRow_Rejected`, isolated so every other floor passes cleanly and only this one is missing -- confirmed it fails without the fix.

ledger/queries.go
arepala-uml: Fixed in 6ce3f42c: added `TestVerifyPointQueryable_UnknownEraId_Rejected`, identical to `WithinAllFloors_Accepted` except the epoch row names era 255 (unresolvable by `eras.GetEraById`). Confirmed it fails if the `queryHardFork` call is deleted from `VerifyPointQueryable` -- neither existing test gave that call anything to reject on.

ouroboros/localstatequery.go
arepala-uml: Fixed in a4717db3: added `TestLocalstatequeryServerAcquire_UnexpectedError_MappedToPointTooOld`, which closes the database out from under a genuinely on-chain point to force a raw, non-sentinel error out of `VerifyPointQueryable`'s own reads. Confirmed it fails without the fix (a bare-error return trips gouroboros' fatal-error path instead of mapping to `AcquireFailurePointTooOld`). Left the `ErrHistoricalStateUnavailable` branch's own mapping uncovered by a dedicated ouroboros-level test for now -- it's the same one-line pattern as the already-tested `ErrPointNotOnChain` branch just above it, and `ledger`'s own `TestVerifyPointQueryable_PastRetentionFloor_Rejected` already proves `VerifyPointQueryable` returns that sentinel correctly; happy to add a dedicated test here too if you'd rather not rely on that combination.

arepala-uml:
Merged current \`main\` into this branch (677f8309) -- no conflicts, full \`ledger\`/\`ouroboros\` suite (race-enabled) passes at head.

Addressed all 3 items from the re-review (replies inline on each thread):
- Blocker: \`verifyStakeDistributionRetentionOnly\` was missing the \`network_state\`-row floor \`totalCirculatingSupply\` separately requires -- a point could pass the epoch-based floor and still get rejected downstream by a real query. Fixed and added a test isolating exactly that gap.
- Added a test proving \`VerifyPointQueryable\`'s \`queryHardFork\` call can't be silently dropped (an epoch row naming an unresolvable era).
- Added a test proving Acquire's neither-sentinel error mapping (a real DB error, not a real disagreement) maps to a graceful \`AcquireFailurePointTooOld\` instead of tearing the connection down.

ledger/pool_stake_distribution.go
```diff
@@ -84,6 +97,71 @@ func checkAsOfEpochRecency(targetEpoch, liveEpoch uint64) error {
...
+		return err
+	}
+	if err := checkAsOfEpochRecency(
+		targetEpoch, liveEpoch, ls.db.StorageMode() == types.StorageModeAPI,
+	); err != nil {
+		return err
+	}
+	if at.pinned() {
+		metaTxn := txn.Metadata()
+		state, err := ls.db.Metadata().GetNetworkStateAsOfSlot(at.Slot, metaTxn)
```
chrisguiney, line 158: Blocker: this floor is unconditional, but the requirement it mirrors is not. totalCirculatingSupply reaches its GetNetworkStateAsOfSlot branch only when ls.config.CardanoNodeConfig is non-nil and ShelleyGenesis().MaxLovelaceSupply is non-zero; otherwise it falls back to totalActiveStake and never reads network_state. On TestVerifyPointQueryable_NoNetworkStateRow_Rejected's own fixture, where CardanoNodeConfig is nil, VerifyPointQueryable rejects slot 350 while GetPoolDistr2, GetStakeDistribution and PoolStakeDistribution all answer. Gate this check on the same condition.

ledger/verify_point_queryable_test.go
```diff
@@ -0,0 +1,184 @@
...
+// failure this whole change exists to close at Acquire time instead.
+//
+// Identical to TestVerifyPointQueryable_WithinAllFloors_Accepted (pinned
+// point's epoch equals the live epoch, so queryShelleyCurrentProtocolParams
+// answers from the live snapshot rather than needing a historical pparams
+// row) except for the one thing this test is about: no
+// db.Metadata().SetNetworkState call, so no network_state row exists at
+// any slot. Isolating every other floor this way means only the new check
+// this test targets can be why this fails.
+func TestVerifyPointQueryable_NoNetworkStateRow_Rejected(t *testing.T) {
```
chrisguiney, line 149: This comment states that a client Acquiring this point and issuing GetPoolDistr2 or GetStakeDistribution got the same error from a live query. On this fixture that is not what happens: CardanoNodeConfig is nil, so totalCirculatingSupply takes its fallback and never reads network_state, and both queries answer at slot 350 with no row seeded. The test does fail without the new floor, so it pins the check -- but what it pins is over-rejection, not parity. The same holds for the SetNetworkState rows added to WithinAllFloors_Accepted and to localstatequery_test.go's PointOnChain_Succeeds.

ledger/queries.go
chrisguiney, line 406: Prior blocker cleared and now causally covered: deleting these six lines in place fails TestVerifyPointQueryable_UnknownEraId_Rejected while every other ledger test stays green. Class audit of this gate against the point-aware handlers: GetUTxOWhole and GetUTxOByTxIn call the same checkUtxoRetentionWindow, GetEpochNo the same resolveAsOfEpoch found flag, and pparams and current era are invoked here directly, so those four are exact. Stake distribution is the only check that is a re-implementation rather than a call, and it is where the remaining drift sits.

ouroboros/localstatequery.go
chrisguiney, line 215: Prior blocker cleared and causally covered: replacing this mapping with a bare error return in place fails TestLocalstatequeryServerAcquire_UnexpectedError_MappedToPointTooOld while its three siblings stay green. Bot thread disposition: CodeRabbit's Acquire-lifetime race finding is confirmed and deferred to dingo#4322, which is open and describes it accurately. No bot reviewed this head: CodeRabbit's status reports incremental reviews disabled, and cubic has no check, status or record. The branch is 31 commits behind main and its checks predate that movement.

chrisguiney (changes requested):
Re-review of the two commits since the last decision. All three prior blockers are cleared and causally covered -- reverting each fix in place fails its own new test. One new blocker: the new network_state floor is unconditional where the requirement it mirrors is configuration-conditional, so Acquire rejects a point every query still answers. Checks are green at head but started before main last moved; the merged tree builds and ledger, ledger/snapshot and ouroboros pass there. No bot reviewed this head.

ledger/pool_stake_distribution.go
arepala-uml: Confirmed valid and reproduced exactly as described, fixed in cb353ca8: gated the floor on ls.circulatingSupplyGenesis() != nil (extracted from totalCirculatingSupply's own identical condition, so the two can't drift apart again). Added TestVerifyPointQueryable_NoNetworkStateRow_AcceptedWhenFloorInactive -- identical fixture to the reject test, CardanoNodeConfig left nil -- proving the floor correctly stays inactive when the real requirement doesn't apply. Confirmed it fails against the unconditional version.

ledger/verify_point_queryable_test.go
arepala-uml: Confirmed valid: this fixture had no CardanoNodeConfig, so it passed for the wrong reason. Fixed in cb353ca8 -- added cardanoNodeConfigWithMaxLovelaceSupply to genuinely activate the floor this test targets, and corrected the comment to stop overclaiming what a real client's query would hit. Same fix applied to WithinAllFloors_Accepted. For localstatequery_test.go's PointOnChain_Succeeds, removed the SetNetworkState call entirely instead -- that test isn't about the network_state floor specifically, so activating it there would just be unneeded fixture complexity; the floor correctly stays inactive with CardanoNodeConfig unset, matching every other test in that file.

ledger/queries.go
arepala-uml: Thanks for the class audit -- confirms the remaining drift is isolated to the stake-distribution re-implementation, now fixed in cb353ca8 (the same commit addressing your other two findings this round).

ouroboros/localstatequery.go
arepala-uml: Acknowledged -- thanks for confirming the #4322 disposition and the bot-review status.

arepala-uml:
Addressed this round's blocker (replies inline):

- **Real bug, now fixed** (cb353ca8): the new \`network_state\` floor was unconditional, but the requirement it mirrors (\`totalCirculatingSupply\`) isn't -- it only reads \`network_state\` when \`CardanoNodeConfig\` has a real \`ShelleyGenesis\` with a nonzero \`MaxLovelaceSupply\`. Every existing ledger test fixture in this repo lacks that, so the floor was rejecting points every real query would have happily answered. Extracted \`circulatingSupplyGenesis\`, shared by both, so the two conditions can't drift apart again.
- Fixed the test that hid this (its own fixture never activated the floor it claimed to test) and added a new test proving the floor correctly stays inactive when the real requirement doesn't apply.
- Applied the same fix to #4319, which carries a cherry-picked copy of this code.

Build, vet, and the full \`ledger\`/\`ouroboros\` suites (race-enabled) pass at head.

arepala-uml:
Merged latest `main` into this branch. One conflict in `ledger/state.go`: main added `resolveRollbackTarget` at the same location this branch adds `utxoPruningDeferredForCatchup` — kept both functions, no logic changes. `go build ./...`, `go vet ./...`, and `golangci-lint run ./...` are clean; `ledger/...` and `ouroboros/...` tests pass.

arepala-uml:
Re-verified this PR's review history at current head (3602b349) against every CHANGES_REQUESTED/COMMENTED review thread. All prior chrisguiney findings are confirmed still fixed:

- `ledger/queries.go:539` (epoch 0 vs. not-found) -- fixed in fd2aa0ce, still correct.
- `ledger/queries.go:376`/`406` (queryHardFork missing from VerifyPointQueryable, then untested) -- fixed in fd2aa0ce and 6ce3f42c, `TestVerifyPointQueryable_UnknownEraId_Rejected` still fails if the six-line call is removed.
- `ledger/queries.go:362` (unfiltered PoolStakeDistribution call for a retention-only check) -- fixed in fd2aa0ce (`verifyStakeDistributionRetentionOnly`).
- `ouroboros/localstatequery.go:186`/`215` (Acquire behavior untested) -- fixed in fd2aa0ce and a4717db3, `TestLocalstatequeryServerAcquire_UnexpectedError_MappedToPointTooOld` and siblings still pass/fail as required.
- `ledger/snapshot/rotation.go:1136` (DATABASE.md's three retention claims) -- confirmed current DATABASE.md text is accurate: `reward_account_output` and `pool_stake_snapshot` are correctly documented as the two tables whose retention depends on storage mode, the retention table annotates `pool_stake_snapshot` as CORE-mode-only, and the unbound-in-API-mode note is present.
- `ledger/state.go:3320` (`utxoPruningDeferredForCatchup` untested) -- fixed in 45d70f44.
- `ledger/pool_stake_distribution.go:127`/`158` (network_state floor missing, then unconditional) -- fixed in 6ce3f42c and cb353ca8.
- `ledger/verify_point_queryable_test.go:149` (fixture didn't activate the floor it claimed to test) -- fixed in cb353ca8.

One gap remained open and unfixed: `checkAsOfEpochRecency`'s `apiStorageMode` early-return branch (ledger/pool_stake_distribution.go) had no dedicated regression test, unlike its three sibling checks. Added `TestVerifyPointQueryable_APIStorageMode_PastRetentionFloor_Accepted` (ledger/verify_point_queryable_test.go), mirroring `TestVerifyPointQueryable_PastRetentionFloor_Rejected` with the database opened in API storage mode. Verified by temporarily no-opping the `apiStorageMode` branch: the new test fails with `ErrHistoricalStateUnavailable` as expected, and every sibling test stays green; restoring the real code makes it pass again.

`go build ./...`, `go vet ./...`, `golangci-lint run ./...` (no findings in files this PR touches -- all reported issues are pre-existing `protogetter`/`gosec` findings in generated `midnight/midnight_state.pb.go`, untouched by this branch), and `go test -count=1 ./ledger/... ./ouroboros/...` are all clean at head.

arepala-uml:
Added test coverage for `checkUtxoRetentionWindow`'s catchup-deferral conjunct (`ledger/queries.go:2440`, the `!ls.utxoPruningDeferredForCatchup(tip.Point.Slot, stabilityWindow)` clause).

**Gap**: this conjunct had no test forcing it to `true`. Deleting it left `go test ./ledger/...` fully green — nothing proved that a pin issued while pruning is deferred for catchup skips the live-tip retention floor a caught-up node would otherwise apply. Same class of gap already fixed once in this PR for `checkAsOfEpochRecency`'s `apiStorageMode` branch (`TestVerifyPointQueryable_APIStorageMode_PastRetentionFloor_Accepted`).

**Fix**: added `TestQueryShelleyUtxoByTxIn_RetentionWindow_DeferredForCatchup_NotRejected` in `ledger/queries_utxo_asof_test.go`. It marks an active upstream connection with no admitted target yet (`utxoPruningDeferredForCatchup`'s own "still syncing" case) and pins a slot (1) well below the live-tip floor (tip 200,000, default stability window 50,000 → floor 150,000), asserting the pin is accepted rather than rejected.

**Revert-and-test proof**: temporarily removed just the `!ls.utxoPruningDeferredForCatchup(...)` conjunct in `checkUtxoRetentionWindow` (so the floor applies unconditionally). The new test failed:

```
historical ledger state unavailable for the requested point: GetUTxOByTxIn pinned to slot 1 is older than this node's spent-UTxO retention floor (slot 150000, tip 200000) -- a UTxO spent and pruned before that floor cannot be reliably reconstructed as of the pinned point
```

Restored the conjunct: the new test and the full `./ledger/...` suite pass again.

**Validation**: `go build ./...`, `go vet ./...`, `go test ./ledger/...` (full package), `golangci-lint run ./ledger/...` — all clean.

ouroboros/localstatequery.go
```diff
@@ -145,39 +163,65 @@ func (o *Ouroboros) localstatequeryServerAcquire(
...
+		// stricter retention floor (UTxO whole/by-ref, stake/pool
+		// distribution, current protocol parameters at a historical
+		// epoch) -- confirmed live, this hit the exact same
+		// connection-killing gap #4156 fixed for the on-chain check alone,
+		// just for a different, retention-based rejection reason
+		// (blinklabs-io/dingo#382 protocol-compliance finding).
+		// VerifyPointQueryable's own doc comment covers verifyPointOnChain
+		// too, so this subsumes VerifyPointOnChain rather than needing
+		// both checks run separately.
+		if err := o.ledgerState.VerifyPointQueryable(nil, point); err != nil {
```
wolf31o2, line 186: Replacing this call with the previous o.ledgerState.VerifyPointOnChain(point) in place leaves every TestLocalstatequery test green: UnexpectedError_MappedToPointTooOld still passes because a closed database fails VerifyPointOnChain too, and no test Acquires an on-chain point that is past a retention floor. That leaves both the wire-visible change and the ErrHistoricalStateUnavailable branch at line 194 unpinned. Add a server-level case that Acquires an on-chain point past a floor and asserts ErrAcquireFailurePointTooOld with nothing recorded.

ledger/queries.go
```diff
@@ -280,13 +306,112 @@ func (ls *LedgerState) queryShelleyEpochNo(
...
+		return nil
+	}
+	if txn == nil {
+		txn = ls.db.Transaction(false)
+		defer txn.Release()
+	}
+	if err := ls.verifyPointOnChain(txn, at); err != nil {
+		return err
+	}
+	if err := ls.checkUtxoRetentionWindow(txn, at); err != nil {
```
wolf31o2, line 388: Deleting this checkUtxoRetentionWindow call, or the queryShelleyCurrentProtocolParams call at line 396, leaves every TestVerifyPointQueryable test green: PastRetentionFloor_Rejected is rejected by the stake floor either way, and no fixture has a missing pparams row as its only failure. The stake and current-era checks are each pinned by a case where only they reject. Add one such case for the UTxO floor and one for the pparams row, so neither check can be dropped silently.

ledger/pool_stake_distribution.go
```diff
@@ -84,6 +97,81 @@ func checkAsOfEpochRecency(targetEpoch, liveEpoch uint64) error {
 	return nil
 }
 
+// verifyStakeDistributionRetentionOnly checks the same retention floors
+// PoolStakeDistribution enforces, without materializing anything
+// PoolStakeDistribution actually reads (human review, Chris Guiney,
```
wolf31o2: AGENTS.md says to delete comments that narrate a change history. This doc comment, like 30 added lines across ledger/pool_stake_distribution.go, queries.go, snapshot/rotation.go, ouroboros/localstatequery.go, DATABASE.md and the new tests, cites review rounds, reviewer names, first versions and live runs (human review, Chris Guiney, dingo#4319/#4320; a first version of this function). Keep the invariant each one states, such as why the network_state floor shares circulatingSupplyGenesis, and drop the history.

wolf31o2 (changes requested):
The headline Acquire change is not pinned at its entry point, and two of the five checks inside VerifyPointQueryable can be deleted with the suite green. Prior findings from earlier rounds: epoch-0 fallback, queryHardFork coverage, stake-floor cost, DATABASE.md, the API-mode skip, both catchup conjuncts, the network_state floor and its gate are fixed, and each reverted fix in place fails its own test. The ErrHistoricalStateUnavailable-to-PointTooOld mapping stays untested. The merged tree with current main builds, lints clean, and passes ledger, ledger/snapshot and ouroboros under -race.

ouroboros/localstatequery.go
arepala-uml: Confirmed valid: no existing test distinguished VerifyPointQueryable from VerifyPointOnChain at this call site. Fixed in f4466b00 -- added TestLocalstatequeryServerAcquire_PastRetentionFloor_MappedToPointTooOld, which Acquires slot 1 of a real two-block chain (on-chain) after marking the durable consumed-UTxO prune floor at slot 2 via database.ConsumedUtxoPruneFloorSyncKey. Verified the regression: reverted this call to VerifyPointOnChain locally, ran the new test -- failed with "An error is expected but got nil" (VerifyPointOnChain has no notion of the persisted floor). Restored VerifyPointQueryable, test passes; full TestLocalstatequery* suite green.

ledger/queries.go
arepala-uml: Confirmed valid: neither checkUtxoRetentionWindow nor queryShelleyCurrentProtocolParams had a case where it alone was the rejection. Fixed in fb93a602 -- added TestVerifyPointQueryable_UtxoFloorOnly_Rejected (persisted consumed-UTxO prune floor above the pinned slot, epoch/stake/era all otherwise within their own windows) and TestVerifyPointQueryable_PParamsRowOnly_Rejected (pinned epoch sitting exactly at the stake-retention window's edge -- accepted there -- with no persisted pparams row for it). Verified independence by no-opping each call in queries.go in turn: with checkUtxoRetentionWindow disabled, only UtxoFloorOnly_Rejected failed (others including PParamsRowOnly_Rejected stayed green); with queryShelleyCurrentProtocolParams disabled, only PParamsRowOnly_Rejected failed (UtxoFloorOnly_Rejected stayed green). Restored both calls, all 8 TestVerifyPointQueryable* cases pass.

ledger/pool_stake_distribution.go
arepala-uml: Confirmed valid: several added comments cited review rounds, reviewer names, PR numbers, and "confirmed live"/"first version" framing instead of just stating the invariant. Fixed in bce0c8c6 -- reworded comments across ledger/pool_stake_distribution.go, ledger/queries.go, ledger/snapshot/rotation.go, ouroboros/localstatequery.go, DATABASE.md, and the new test files (verify_point_queryable_test.go, localstatequery_test.go), keeping each stated invariant (e.g. why the network_state floor shares circulatingSupplyGenesis with totalCirculatingSupply, why queryHardFork's era resolution must be checked in VerifyPointQueryable) while dropping the historical-narration framing. Also cleaned my own two prior commits' new test doc comments, which had the same "reviewer's finding" framing. Scoped strictly to lines this PR added (checked via diff against the merge-base) -- left pre-existing narration elsewhere in these same files (e.g. queries.go:968, localstatequery.go:59/69, unrelated DATABASE.md sections) untouched since it predates this PR. No behavioral change: gofmt clean, go vet clean, ledger/ouroboros tests pass unchanged.

arepala-uml: My earlier reply (4076053951) overclaimed completeness. A fresh, careful re-grep just found four instances this pass missed, all in test files this PR itself added:

- ledger/queries_hardfork_currentera_test.go: TestQueryHardFork_CurrentEra_NoEpochRecordRejected's doc comment still cited "(human review, Chris Guiney, dingo#4320)".
- ledger/snapshot/rotation_reward_retention_mode_test.go (two instances): TestCleanupOldSnapshotsCoreModePrunesRewardAccountOutput's doc comment cited "(dingo#1900/#382, human review, Chris Guiney: ...)", and TestCleanupOldSnapshotsAPIModeRetainsRewardAccountOutput's doc comment cited "(human review, Chris Guiney)".
- ledger/utxo_pruning_catchup_test.go: TestUtxoPruningDeferredForCatchup's doc comment cited "(human review, Chris Guiney, dingo#4320)" and quoted the reviewer's own wording verbatim.

Fixed in 8f4eaa25, reworded each to keep the durable invariant it was actually stating (why the epoch-0 row is deliberately seeded, why CORE- and API-mode pool_stake_snapshot retention windows must hold, why the two must-NOT-defer cases matter for utxoPruningDeferredForCatchup) and drop the review-history framing. Kept the dingo#1900/#382 feature-tracking citations, matching this file's own pre-existing "dingo #1875" citation style, since those identify the change being tested rather than narrate the review itself.

Re-verified: grepped this PR's own changed files for "human review", "dingo#4319", "dingo#4320", and "Chris Guiney" -- zero hits now. go build ./..., go vet ./..., and go test ./ledger/... ./ouroboros/... all pass; comment-only change, no functional diff.

wolf31o2 (approved):

chrisguiney (approved):

PR #4338, opened 2026-09-16
perf(node): parallelize trusted block decoding

wolf31o2:
Trusted Mithril replay previously decoded each immutable block serially before applying ordered batches. This change decodes each bounded 50-block batch with a 2–8 worker pool sized from `GOMAXPROCS`, reorders results by input index, and keeps chain insertion and ledger replay serial. Cancellation propagates through bounded queues, and decode errors stop the batch.

## Summary by cubic

Speeds up trusted Mithril replay by decoding immutable blocks in parallel. Trusted replay previously decoded each block serially; it now decodes 50-block batches with a 2–8 worker pool sized from `GOMAXPROCS` and reorders results by input index, while chain insertion and ledger replay stay serial. Cancellation propagates through bounded queues, and decode errors stop their batch.

- Adds tests for order preservation, cancellation, and decode-error cancellation.


internal/node/load_test.go
```diff
@@ -53,6 +53,117 @@ func TestExtractHeaderCbor(t *testing.T) {
...
+	_, err := decodeImmutableBlockBatch(
+		ctx,
+		blocks,
+		lcommon.VerifyConfig{SkipBodyHashValidation: true},
+		4,
+	)
+	require.ErrorIs(t, err, context.Canceled)
+}
+
+func TestDecodeImmutableBlockBatchDecodeErrorCancelsWorkers(t *testing.T) {
```
cryptodj413, line 114: Non-blocking: mutation-tested this — deleting the `cancel()` call in `decodeImmutableBlockBatch`'s error branch (`load.go`, inside `for result := range results`) still passes this test unchanged. It only proves an error is eventually returned, not that the pool short-circuits on the first decode failure (the thing the name promises and the doc comment above the function describes). Worth strengthening — e.g. an artificial slow/blocking decode plus a counter of how many jobs were dispatched before cancellation, or asserting on elapsed time relative to the serial case — but the underlying `cancel()` call itself is correct and doing its job (verified directly and by a separate mutation on `ctx.Err()` in the same function, which the cancellation test does catch reliably across 200 runs).

cryptodj413 (approved):
Concurrency-sensitive change (worker pool, cancellation, ordered reassembly for trusted Mithril replay decode) — no bot review at this head (CodeRabbit rate-limited, Cubic skipped), so full review done directly with -race and mutation testing rather than by inspection alone.

**Correctness, verified by tracing + testing, not just reading:**
- Ran the three new tests `-race -count=20`: clean, no data races.
- Traced `decodeImmutableBlockBatch`'s shutdown paths: dispatcher always closes `jobs` via defer regardless of exit path; every worker selects on `decodeCtx.Done()` on both receive and send so none can block forever on a full/closed channel; `decoded[result.index]` is written only by the single consuming goroutine, so no race despite N concurrent producers.
- Mutation-tested the ordering guarantee: changed `decoded[result.index]` to append-by-completion-order (`decoded[seen]`) — `TestDecodeImmutableBlockBatchPreservesOrder` failed immediately with a hash mismatch. Confirms the test has real teeth, not just luck from small/fast batches.
- Mutation-tested the final `ctx.Err()` check: removing it made `TestDecodeImmutableBlockBatchCancellation` fail consistently (200/200 runs) with the generic "ended early" error instead of a properly wrapped `context.Canceled" — confirms that check matters for callers that branch on `errors.Is(err, context.Canceled)` and that the test reliably catches its removal.
- `ARCHITECTURE.md` update matches the implementation exactly (2–8 workers from GOMAXPROCS, 50-block batch, indexed reordering, serial insertion/replay, cancellation through bounded queues) — checked against the code, not copied from the PR description.

**One non-blocking finding**, posted inline on `load_test.go`: `TestDecodeImmutableBlockBatchDecodeErrorCancelsWorkers` still passes with the `cancel()` call deleted — it only proves an error propagates, not that the pool actually short-circuits on the first failure. The production code is correct (verified independently above); the test just doesn't prove the specific thing its name claims. Worth strengthening in a follow-up, not a blocker here.

CI: all green, including `go-test (Linux, race)` at 24m49s. Local `go test -race ./internal/node/...` also clean.

Approving.

Issue #4341, opened 2026-09-16
Pulse the stake reward calculation across the epoch instead of one pass


wcatz:
## Problem

The delayed reward update is computed in a single pass over every delegator in the mark snapshot, and neither of the two places it can run is incremental:

- `precomputeStakeRewardsAfterEpochTransition` runs it once on a background goroutine in a read-only transaction. Measured on mainnet: **10m23s** for 1,309,350 delegators across 2,676 pools (epoch-654 transition at `17:55:46.576Z` to `precomputed stake rewards` at `18:06:09.732Z`).
- When that result is absent, or a rollback invalidates it, `applyStakeRewards` recalculates the whole round **inline inside the epoch-rollover write transaction**, on the block-processing goroutine. Block application stops for the duration, and a leader slot inside that window is lost.

Because the work is all-or-nothing it also restarts from zero on any interruption, and a rollback late in the epoch can discard a finished result with no time to redo it before the boundary.

## Proposed solution

Make the calculation incremental, as cardano-ledger's `PulsingRewUpdate` does. Arm a pulser at the randomness-stabilisation window — `calculateStabilityWindow` already exists — and advance it by a bounded number of steps as part of each block-apply transaction. For a mainnet epoch that is roughly 65 delegators per block, small enough to disappear into existing per-block work.

Two properties the current design lacks and the pulser needs:

- **Resumable.** Persist the cursor and accumulator so a restart continues rather than recomputing.
- **Rollback-aware.** Carry `rewardInputGeneration` in the pulser state and invalidate/restart the pulse when it changes, rather than producing a stale result.

The boundary then applies an already-complete update instead of computing one. The inline fallback stays for the case where a pulse genuinely could not finish, but it stops being a routine path.

Worth pairing with a `dingo_ledger_epoch_rollover_duration_seconds` metric — there is currently no measurement of how long the boundary takes, which is why this went unnoticed.

wcatz:
Measured on a mainnet node at the 655→656 boundary tonight, taking the inline path described above (`applied stake rewards … precomputed=false`):

- Block application stopped **21:44:45.264Z**, resumed **21:58:15.488Z** — **13m30s** with no block applied.
- `applied stake rewards` landed at 21:53:44.739Z, so the reward calculation alone was **~9m00s**; the remaining rollover steps (governance tally over 1053 DReps / 2677 pool rows, then commit) took the other ~4m30s.
[...]

PR #4342, opened 2026-09-16
fix(ledger): precompute the in-progress epoch's rewards at startup

wcatz:
The reward precompute is queued only by the epoch-transition event, which cannot fire for an epoch that began before the process did. A node started mid-epoch reaches its next boundary with no precompute, so `applyStakeRewards` falls back to calculating the whole round inline, inside the epoch-rollover write transaction on the block-processing goroutine — the serialization the async path exists to avoid. On mainnet that scan covers every delegator in the mark snapshot; measured at 10m23s for 1.3M delegators. No block is applied for the duration and a leader slot in the window is lost.

`Start` now queues the same work for the current epoch, right after the subscription it complements. The path is already idempotent — `precomputeStakeRewardsCalculate` returns early when a valid precompute exists, and a syncing node defers on the RUPD prefilter slot — so the catch-up costs a lookup when there is nothing to do.

Verified on a live mainnet block producer that is in exactly this state: started mid-epoch, `reward_ada_pots` carries no usable row for the upcoming round, so tonight's boundary takes the inline path. Issue #4341 tracks the structural fix (pulse the calculation across the epoch rather than in one pass); this only removes the common trigger.

`./ledger/` passes with `-race`. ARCHITECTURE.md updated; DATABASE.md checked and not affected.

---
## Summary by cubic
Fixes reward precomputation for nodes that start mid-epoch. Previously the precompute was only queued by an epoch-transition event, so the next rollover calculated all stake rewards inline and could block block processing for minutes. `Start` now queues the in-progress epoch's precompute, keeping that work off the rollover path.

**Bug Fixes**

- Skips queueing until the current epoch has a length and nonce, and preserves the existing deferral for syncing nodes.
- Adds tests for normal startup, uninitialized epochs, and epoch zero.
- Documents the behavior in `ARCHITECTURE.md`; issue #4341 still tracks spreading the calculation across the epoch.

<sup>Written for commit 21773198ee1f57337d60fb16bf45bf9f5393ff3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Nodes starting during an active epoch now automatically queue the in-progress epoch’s stake-reward precomputation.
  - This ensures reward calculations are prepared before the next epoch transition.

- **Documentation**
  - Expanded architecture documentation to describe startup catch-up behavior, retry handling, and epoch-transition scheduling.


ledger/state.go
```diff
@@ -1850,6 +1850,11 @@ func (ls *LedgerState) Start(ctx context.Context) error {
 			ls.handleRewardPrecomputeEpochTransition,
 		)
 	}
+	// The subscription above cannot fire for an epoch that began before this
+	// process did, so catch up the in-progress epoch's reward round here.
+	// Without it, a node started mid-epoch calculates that round inline inside
+	// the next epoch-rollover transaction instead of ahead of it.
+	ls.queueStartupRewardPrecompute()
```
chrisguiney, line 1860: This call is the entire behavioral change and no test exercises it. Removing it and running go test ./ledger/ -run TestQueueStartupRewardPrecompute still reports ok, because the three new tests call queueStartupRewardPrecomputeWith directly, so Start losing this line is invisible. ledger/state_test.go already builds a real LedgerState through NewLedgerState and calls ls.Start in TestLedgerStateStartFailsOnMalformedMithrilTrustBoundary, so a test can seed an epoch row, run Start, and assert the round was queued.

ledger/reward_calculation.go
```diff
@@ -1801,6 +1801,56 @@ func finalizePrecomputedRewardOutputs(
...
+}
+
+func (ls *LedgerState) queueStartupRewardPrecomputeWith(
+	precompute func(event.EpochTransitionEvent) error,
+) {
+	ls.RLock()
+	epoch := ls.currentEpoch
+	ls.RUnlock()
+	// An epoch with no length has not been established yet (fresh database),
+	// and queueRewardPrecompute drops an event without a nonce, so there is
```
chrisguiney, line 1835: queueRewardPrecompute does not drop an event without a nonce: it clones EpochNonce and stores the event unconditionally. The nonce check lives in handleRewardPrecomputeEpochTransitionWith, which this path bypasses, so the len(epoch.Nonce) == 0 return is this function's own policy and not a mirror of an existing one. The guard is right for a Byron epoch, the stated reason is not, and a later reader can delete it as redundant. Repeated at ledger/reward_precompute_startup_test.go:64.

ledger/reward_calculation.go
```diff
@@ -1801,6 +1801,56 @@ func finalizePrecomputedRewardOutputs(
...
+	// nothing to catch up on in either case.
+	if epoch.LengthInSlots == 0 || len(epoch.Nonce) == 0 {
+		return
+	}
+	evt := event.EpochTransitionEvent{
```
chrisguiney, line 1840: precomputeStakeRewardsAfterEpochTransition reads only NewEpoch and BoundarySlot from this value; PreviousEpoch, SnapshotSlot and EpochNonce are inert on this path, and ProtocolVersion is left zero where the block-based publisher in ledger/state.go sets it. The new tests assert PreviousEpoch and SnapshotSlot, pinning fields no consumer reads. Either populate ProtocolVersion so the synthesized value matches the publisher, or record here that only the two consumed fields are meaningful.

ledger/reward_precompute_startup_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/blinklabs-io/dingo/event"
+	"github.com/blinklabs-io/dingo/internal/test/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+// The EventBus subscription that drives the reward precompute only fires at an
+// epoch boundary, so an epoch already in progress when the process starts has
+// no event to carry it. Startup must queue that round itself, or the next
+// boundary calculates it inline inside the rollover write transaction.
+func TestQueueStartupRewardPrecomputeQueuesInProgressEpoch(t *testing.T) {
```
chrisguiney, line 83: These three tests exercise the seam rather than the production entry point, so they pass unchanged with the Start call removed. Driving Start once, against a database whose epoch row carries a length and a nonce, is what turns them into coverage of the reported defect; keep the shaping assertions as supporting checks.

chrisguiney (dismissed):
The startup catch-up matches the block-based publisher and the helper is correct, but the single line that is the fix is uncovered: deleting ls.queueStartupRewardPrecompute() from Start leaves all three new tests green, because they drive the queueStartupRewardPrecomputeWith seam and never reach Start. Separately, go-test (Windows) (1.26.x) is red on TestCheckpointWALDoesNotBlockWriteBehindReaderSnapshot, a wall-clock bound in database/plugin/metadata/sqlite that main gained from #4276 after this branch forked; that package does not import ledger, so the red is not this change.

wolf31o2 (approved):

PR #4347, opened 2026-09-16
test(node): cover trusted decode cancellation

wolf31o2:
Follow-up to #4338.

Strengthen trusted immutable-block decode cancellation coverage while retaining a direct malformed-CBOR test of the production decoder.

The injected decoder waits until all non-failing workers have entered the cancellation path before the selected job fails, so the assertion is independent of worker scheduling.


internal/node/load_test.go
```diff
@@ -114,14 +115,45 @@ func TestDecodeImmutableBlockBatchCancellation(t *testing.T) {
...
 	blocks := immutableDecodeBenchmarkBlocks(t)
-	blocks[len(blocks)/2].Cbor = []byte{0xff}
-	_, err := decodeImmutableBlockBatch(
-		context.Background(),
-		blocks,
-		lcommon.VerifyConfig{SkipBodyHashValidation: true},
-		8,
-	)
-	require.Error(t, err)
+	decodeErr := errors.New("decode failed")
```
chrisguiney, line 134: Nothing drives a malformed block through gledger.NewBlockFromCbor any more: TestDecodeImmutableBlockBatchPreservesOrder and BenchmarkDecodeImmutableBlockBatch decode only valid blocks, and this test now injects its own error. Mutating decodeImmutableBlock to drop the error and return a nil block keeps the whole internal/node package green, while the same mutation at the merge base failed the previous version of this test. Keep the blocks[len(blocks)/2].Cbor = 0xff case as its own test so the trusted load path stays proven to reject corrupt immutable blocks.

internal/node/load.go
```diff
@@ -1050,6 +1050,22 @@ type immutableDecodeJob struct {
...
 }
 
+type immutableBlockDecoder func(
+	context.Context,
+	int,
+	immutable.Block,
+	lcommon.VerifyConfig,
+) (gledger.Block, error)
+
+func decodeImmutableBlock(
```
chrisguiney, line 1063: decodeImmutableBlock discards both the context and the index, so the seam reads as indirection with no purpose and invites the assumption that a decode observes cancellation mid-block. It does not: only the worker select and the dispatcher read decodeCtx. A doc comment on immutableBlockDecoder stating that the two parameters exist so a test can fail a chosen job and park the rest on the derived context would keep that contract visible.

chrisguiney (changes requested):
Replacing the corrupt-CBOR case with an injected decoder removes the only coverage that a real decode failure surfaces from decodeImmutableBlockBatch. Making decodeImmutableBlock discard the error from NewBlockFromCbor and return a nil block leaves every test in internal/node green, while the same mutation at the merge base fails the previous version of this test with: An error is expected but got nil. The new cancellation assertion is causal and does fail when the on-error cancel is removed. The merged tree against current main builds, lints and passes.

internal/node/load.go
```diff
@@ -1050,6 +1050,22 @@ type immutableDecodeJob struct {
 	block immutable.Block
 }
 
+type immutableBlockDecoder func(
```
chrisguiney, line 1056: The seam is still undocumented. decodeImmutableBlock discards ctx and index, so a reader can take the decode itself to observe cancellation and read the select on decodeCtx around the result send as redundant; deleting that select reinstates the goroutine blocked on a full channel that the decodeImmutableBlockBatch comment claims to prevent. NewBlockFromCbor is not cancellable. Document on immutableBlockDecoder that ctx and index exist so a test can fail one chosen job and park the rest on the derived context.

internal/node/load_test.go
```diff
@@ -97,6 +98,21 @@ func TestDecodeImmutableBlockBatchPreservesOrder(t *testing.T) {
...
+	blocks := immutableDecodeBenchmarkBlocks(t)
+	blocks[0].Cbor = []byte{0xff}
+
+	_, err := decodeImmutableBlockBatch(
+		context.Background(),
+		blocks,
+		lcommon.VerifyConfig{SkipBodyHashValidation: true},
+		1,
+	)
+	require.Error(t, err)
```
chrisguiney, line 114: Prior finding cleared, verified causally: with decodeImmutableBlock changed to return a nil block and a nil error when NewBlockFromCbor fails, this test fails with An error is expected but got nil while the rest of the package stays green. Bare require.Error does not pin the error identity, but the class is complete: dropping firstErr in the collector leaves this test green and fails TestDecodeImmutableBlockBatchDecodeErrorCancelsWorkers at require.ErrorIs.

internal/node/load_test.go
```diff
@@ -97,6 +98,21 @@ func TestDecodeImmutableBlockBatchPreservesOrder(t *testing.T) {
 	}
 }
 
+func TestDecodeImmutableBlockBatchReportsRealDecodeError(t *testing.T) {
```
chrisguiney, line 102: Checks at b99a3632: go build ./... and go vet ./internal/node exit 0; gofmt -l on both changed files is empty at head and at the merge base; go test -race -count=1 ./internal/node, conformance, import-boundaries and docs-parity all exit 0. All 14 checks at head are green but started before main last moved, so the merged tree was built and tested separately. A -count=2 panic on duplicate metrics collector registration reproduces on origin/main with none of this change and does not belong to it.

chrisguiney (changes requested):
Re-review at b99a3632: the delta commit clears the real-decode coverage gap, verified causally, and one finding stands on the undocumented decoder seam in load.go. Merged with origin/main at 771b742d the tree builds, vets, is gofmt clean, passes internal/node under -race, and golangci-lint on internal/node reports 0 issues; conformance, import-boundaries and docs-parity pass. No bot reviewed this head: the CodeRabbit status reads Review rate limited and its issue comment declines e022bf12..b99a3632, and cubic has no review record. No bot threads are open.

internal/node/load_test.go
```diff
@@ -114,14 +130,45 @@ func TestDecodeImmutableBlockBatchCancellation(t *testing.T) {
...
+	cancelObserved := make(chan struct{})
+	var cancelOnce sync.Once
+	decoder := func(
+		ctx context.Context,
+		index int,
+		_ immutable.Block,
+		_ lcommon.VerifyConfig,
+	) (gledger.Block, error) {
+		if index == 0 {
+			return nil, decodeErr
```
chrisguiney, line 153: Schedule-dependent: if index 0 returns decodeErr before any other worker enters the decoder, the collector cancels decodeCtx and the rest exit through the worker loop's own case <-decodeCtx.Done(), so nothing reaches this decoder's <-ctx.Done() and cancelObserved never closes. On the merged tree, go test -race -count=1000 -cpu=8 -run TestDecodeImmutableBlockBatchDecodeErrorCancelsWorkers ./internal/node fails 22/1000, -cpu=2 fails 1/1000, and without -race at -count=5000 -cpu=8 it fails 3/5000. Hold the failing decoder until the other workers are parked on ctx.Done().

internal/node/load.go
```diff
@@ -1050,6 +1050,25 @@ type immutableDecodeJob struct {
 	block immutable.Block
 }
 
+// immutableBlockDecoder is a test seam: production decoding ignores the
```
chrisguiney, line 1053: Prior finding cleared: this states the contract the code holds, since decodeImmutableBlock discards ctx and index and NewBlockFromCbor is not cancellable. Correcting the adjacent claim from the earlier round: the worker's select around the result send is an early exit, not a guard against a goroutine blocked on a full channel. The collector drains results until close with no early return, so replacing that select with a plain send leaves internal/node green. No test pins it and none is needed.

internal/node/load_test.go
```diff
@@ -97,6 +98,21 @@ func TestDecodeImmutableBlockBatchPreservesOrder(t *testing.T) {
...
+	blocks := immutableDecodeBenchmarkBlocks(t)
+	blocks[len(blocks)/2].Cbor = []byte{0xff}
+
+	_, err := decodeImmutableBlockBatch(
+		context.Background(),
+		blocks,
+		lcommon.VerifyConfig{SkipBodyHashValidation: true},
+		1,
+	)
+	require.Error(t, err)
```
chrisguiney, line 114: Prior finding cleared, verified causally at this head: with decodeImmutableBlock changed to return a nil block and a nil error when NewBlockFromCbor fails, this test fails with An error is expected but got nil while the rest of internal/node stays green. Bare require.Error does not pin the error identity, but the class is complete: replacing firstErr with a fresh error in the collector leaves this test green and fails TestDecodeImmutableBlockBatchDecodeErrorCancelsWorkers at require.ErrorIs.

internal/node/load_test.go
chrisguiney, line 102: Checks at 974aee04, which is the merged tree because its merge base is current main: go build ./..., go vet ./internal/node, gofmt on both changed files, go test -race -count=1 ./internal/node, go test ./internal/test/conformance (2575/2575 on sqlite), make import-boundaries and make docs-parity all exit 0, and golangci-lint on internal/node reports 0 issues. The merge audit reports both files kept, with no revert and no branch-internal merge. DATABASE.md and ARCHITECTURE.md were read against the diff and are unaffected.

internal/node/load.go
```diff
@@ -1060,6 +1079,18 @@ func decodeImmutableBlockBatch(
 	rawBlocks []immutable.Block,
 	verifyCfg lcommon.VerifyConfig,
 	workerCount int,
+) ([]gledger.Block, error) {
```
chrisguiney, line 1082: go-test (Linux) (1.26.x) is red on internal/test/conformance: TestRulesConformanceVectorsMysql hit the 30m limit inside truncateMysqlTables on a MySQL read. That package does not import internal/node, and internal/node is ok 7.763s in the same job; the only branch change since the green head b99a3632 is one doc comment and one index in a test, so the stall is not this change. The MySQL vectors are the check not run locally. CodeRabbit reviewed this head and its one inline finding is the blocker above; cubic posted no check, status or review record.

chrisguiney (changes requested):
Re-review at 974aee04: both prior findings are cleared and verified causally, and one new blocker stands on the cancellation test, which is schedule-dependent and fails 22 times in 1000 runs under -race at -cpu=8. go-test (Linux) is red on a MySQL conformance stall that this change cannot reach.

internal/node/load_test.go
wolf31o2: Fixed in ae2587d9: the regression test corrupts block CBOR and exercises the production decoder error path.

internal/node/load.go
wolf31o2: Fixed in ae2587d9: immutableBlockDecoder documents that context and index are test controls and that NewBlockFromCbor is not cancellable.

internal/node/load.go
wolf31o2: Fixed in ae2587d9: immutableBlockDecoder now states the seam and cancellation contract explicitly.

internal/node/load_test.go
wolf31o2: Confirmed at the current head: the real decoder error and collector cancellation paths have separate regression coverage.

internal/node/load_test.go
wolf31o2: No code change required; this thread records validation of the earlier head.

internal/node/load_test.go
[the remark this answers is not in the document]
wolf31o2: Fixed in 31131d98: the failing decoder waits until the other workers have entered the decoder before returning the error.

wolf31o2: Fixed in 31131d98: the failing decoder waits until the other workers have entered the decoder before returning the error.

internal/node/load.go
wolf31o2: Confirmed; the decoder seam comment states the production and test contracts accurately.

internal/node/load_test.go
wolf31o2: Confirmed; the real decoder error and collector cancellation paths have separate regression coverage.

internal/node/load_test.go
wolf31o2: No code change required; this thread records validation of the prior head.

internal/node/load.go
wolf31o2: No branch change is required for this unrelated MySQL conformance timeout.

internal/node/load_test.go
```diff
@@ -114,14 +130,51 @@ func TestDecodeImmutableBlockBatchCancellation(t *testing.T) {
...
+		_ immutable.Block,
+		_ lcommon.VerifyConfig,
+	) (gledger.Block, error) {
+		if index == 0 {
+			for range workerCount - 1 {
+				<-workersReady
+			}
+			return nil, decodeErr
+		}
+		workersReady <- struct{}{}
```
arepala-uml: workersReady <- struct{}{} is unconditional with no ctx.Done() case, so after cancel() fires a recycled worker still sends here; capacity is workerCount-1 and overflows, deadlocking the test. Reproduced with go test -race -cpu=8 -count=20 -run TestDecodeImmutableBlockBatchDecodeErrorCancelsWorkers ./internal/node/, failing on the 6th iteration with all goroutines blocked at this line for job index 15.

arepala-uml (changes requested):
Reproduced a deadlock in the new synchronization added by 31131d98; see inline comment on load_test.go for the mechanism and repro command.

arepala-uml (approved):

PR #4357, opened 2026-09-16
fix(ledger): align MIR overflow handling and fix stale MIR docs/comments

### LINKED ISSUE #4028: ledger/mir: MIR sign follow-ups — stale discriminator docs, era-conditional accumulation, discard class
Non-blocking review findings from #3981, which carried the sign through the MIR pipeline.

### 1. `DATABASE.md:2656` documents the discriminator the change replaced

The `GetMIRCertsInSlotRange` section still says:

> The `mir.id` value is retained by the processed effect as the per-MIR reward-credit discriminator so multiple MIR certs can credit the same account at one boundary without collapsing into one `account_reward_delta` row.

After #3981, `ledger/mir.go:434` passes `mirRewardSourceHash(credit.pot)`, and `TestApplyMIRCerts_MultipleDistributionsSameAccount` asserts exactly the collapse that sentence says does not happen. The neighbouring `move_instantaneous_rewards` row and the `GetAccountSumsByCredential` samples were updated in the change; this paragraph is the remaining member of the same class. `make docs-parity` does not cover prose, so nothing catches it.

### 2. The accumulation comment is unconditional; the reference is protocol-version conditional

`ledger/mir.go:59` describes the additive fold unconditionally. In `Cardano.Ledger.Shelley.Rules.Deleg.delegTransition`, `StakeAddressesMIR` branch:

```haskell
if hardforkAlonzoAllowMIRTransfer pv
  then do
    let cm = Map.unionWith (<>) credCoinMap' instantaneousRewards
    all (>= mempty) cm ?! MIRProducesNegativeUpdate
    pure (cm, potAmount `addDeltaCoin` delta)
  else do
    all (>= mempty) credCoinMap ?! MIRNegativesNotCurrentlyAllowed
    pure (Map.union credCoinMap' instantaneousRewards, potAmount)
```

with `hardforkAlonzoAllowMIRTransfer pv = pvMajor pv > natVersion @4`.

The additive fold is protocol-major >= 5 (Alonzo, Babbage) behaviour. For majors 2-4 (Shelley, Allegra, Mary) the reference combines with `Map.union`, left-biased on `credCoinMap'`, which *replaces* a pending entry for the same credential rather than adding to it. Two positive MIR certificates naming one credential in one pre-Alonzo epoch credit only the later amount under the reference, and the sum here.

Not a regression: the previous code credited each certificate independently, so the balance was already the sum. MIR deltas cannot be negative pre-Alonzo (`MIRNegativesNotCurrentlyAllowed`), so the signed handling has no pre-Alonzo interaction. Two parts: qualify the comment with the protocol version, and decide the replace-versus-add case on its own merits rather than leaving the comment asserting reference agreement across all four MIR eras.

### 3. The error-to-discard conversion covered one of three members

#3981 converted `addCredit` from returning an error to setting `boundary.discard`, on the rationale that an error wedges the node — the stored certificates are re-read and re-fail on every deterministic retry.

`addTransfer` (`ledger/mir.go:258`) and `mirAvailablePot`'s inbound-overflow branch still return an error, and that rationale applies to them identically. Both are unreachable with real `coin` values, so this is consistency rather than a defect.

### 4. Nit

`database/models/mir.go:40` — `ID` no longer orders anything. Ordering comes from the store (`GetMIRCertsInSlotRange` issues `ORDER BY added_slot ASC, id ASC` on the parent and `ORDER BY id` on the reward rows, `certificates.go:116`, `:153`) and `collectMIRBoundary` consumes `effects` in receipt order. The only remaining read is the `MIR %d carries no delta` error message, itself unreachable since `parseNullBigInt` maps NULL and the empty string to zero and never returns nil.

arepala-uml:
## Summary
Review follow-ups from #3981, closes #4028:

- **Consensus-visible**: the MIR distribution fold is now protocol-era-conditional, matching `Cardano.Ledger.Shelley.Rules.Deleg.delegTransition`'s `hardforkAlonzoAllowMIRTransfer` gate. It sums per (pot, credential) for era >= Alonzo as before, but for Shelley/Allegra/Mary it now keeps only the latest certificate's amount for a repeated (pot, credential) pair instead of summing, since the reference uses a plain left-biased `Map.union` pre-Alonzo. This changes replayed reward-account balances for any pre-Alonzo epoch with more than one MIR certificate naming the same stake credential.
- Correct `DATABASE.md`'s stale claim that `mir.id` is the `account_reward_delta` discriminator; it is the source-pot hash.
- Convert the remaining hard-error paths in MIR boundary collection (`addTransfer`'s unknown-pot/overflow branches, the sibling unknown-pot check for a distribution certificate, and `mirAvailablePot`'s inbound-overflow branch) to the same `boundary.discard` no-op `addCredit` already used, so a corrupt or over-budget boundary can't wedge the node on deterministic retry.
- `availablePot` now selects on the same uint pot (`mirPotReserves`/`mirPotTreasury`) `addTransfer`/`addCredit` switch on, instead of a separate string name.
- Fix the stale `MIREffect.ID` comment; ordering comes from the store query and receipt order, not this field, though it is the store's `mir_id` join key.

## Test plan
- [x] `go build ./...`
- [x] `go test -race ./ledger/...`
- [x] `go vet ./ledger/...`
- [x] `golangci-lint run ./ledger/... ./database/models/...`
- [x] `make docs-parity`
- [x] `make import-boundaries`
- [x] `go test ./internal/test/conformance/...`

## Summary by CodeRabbit

- **Bug Fixes**
  - MIR boundary processing now safely discards invalid or overflowing transfers instead of failing epoch rollover.
  - Repeated processing of affected boundaries no longer risks wedging the node due to deterministic retry failures.
  - MIR reward credits are combined consistently by source pot and account, while credits from different pots remain separate.

- **Documentation**
  - Updated MIR query and effect documentation to clarify reward-credit grouping and result ordering.


ledger/mir.go
```diff
@@ -462,25 +474,21 @@ func (ls *LedgerState) applyMIRCredits(
...
-) (available uint64, ok bool, err error) {
+) (available uint64, ok bool) {
 	if balance > ^uint64(0)-in {
-		return 0, false, fmt.Errorf(
-			"MIR pot transfer would overflow %s: pot has %d, moving %d",
-			name,
-			balance,
-			in,
-		)
+		return 0, false
```
chrisguiney, line 532: Inbound overflow now returns a bare ok=false, so applyMIRCerts reports it through the over-budget warn. TestApplyMIRCerts_PotTransferOverflow logs: skipping over-budget MIR at epoch boundary slot=1000 reserves=1 reserves_distributed=0 treasury=18446744073709551615 treasury_distributed=0. Nothing is over budget, and neither the transfer nor the overflow is named, so an operator cannot tell why the boundary was dropped. addCredit and addTransfer carry a reason in boundary.discard, logged as reason=; routing this branch through discard would make the claimed alignment hold.

ledger/mir.go
```diff
@@ -290,10 +303,9 @@ func (ls *LedgerState) collectMIRBoundary(
 	order := make([]mirPendingKey, 0, len(effects))
 	for _, effect := range effects {
 		if effect.OtherPot > 0 {
-			if err := boundary.addTransfer(
-				effect.Pot, effect.OtherPot,
-			); err != nil {
-				return nil, err
+			boundary.addTransfer(effect.Pot, effect.OtherPot)
+			if boundary.discard != "" {
+				return boundary, nil
```
chrisguiney, line 323: Five lines below, at line 313, a distribution cert with an unrecognised pot still returns fmt.Errorf with unknown MIR pot. That is the same corrupt-stored-state input as addTransfer's default branch which this change converts to a discard, and it fails the epoch rollover on every deterministic retry, the wedge the new comments cite. Both paths returned an error before this change, so the diff introduces the asymmetry. Converting line 313 to the same discard would complete the class.

ledger/mir.go
```diff
@@ -245,25 +252,31 @@ func (b *mirBoundary) addCredit(credit mirCredit) {
...
+func (b *mirBoundary) addTransfer(sourcePot uint, amount uint64) {
 	var out, in *uint64
 	switch sourcePot {
 	case mirPotReserves:
 		out, in = &b.reservesOut, &b.treasuryIn
 	case mirPotTreasury:
 		out, in = &b.treasuryOut, &b.reservesIn
 	default:
-		return fmt.Errorf("unknown MIR source pot %d", sourcePot)
+		b.discard = fmt.Sprintf("unknown MIR source pot %d", sourcePot)
```
chrisguiney, line 275: Neither branch converted here has a test: searching ledger/mir_test.go at this head and at the merge base finds no reference to the unknown source pot or to the transfer-total overflow, so the change from failing the rollover to silently dropping the boundary is unpinned. Both are reachable from the existing seed helper: an other_pot row with pot=2 reaches line 268, and two transfers from one pot each of 18446744073709551615 reach line 272. The mirAvailablePot conversion is covered, since deleting its guard in place fails both subtests of TestApplyMIRCerts_PotTransferOverflow.

DATABASE.md
```diff
@@ -3398,7 +3398,7 @@ row from `dingo_db_missing` to the informational `pool_departed`.
...
+MIR certificates for the epoch range `[startSlot, endSlot)`, applied at the epoch boundary as the Shelley INSTANT rule. Distribution certs (`other_pot = 0`) credit registered reward accounts and debit the source pot in `network_state`; pot-to-pot transfer certs (`other_pot > 0`) move that amount between treasury and reserves. Every cert in the range is aggregated before any is applied: distribution totals count only credentials with a registered, active account, transfers are folded into the available pot balances, and the boundary writes a single `network_state` row. If either pot cannot cover its total the boundary is a no-op rather than an error, so an over-budget cert cannot fail the epoch rollover on every retry. MIR has no transaction hash to serve as the `account_reward_delta` discriminator, so the source pot is hashed into a synthetic one instead: this keeps a reserves credit and a treasury credit to the same account at one boundary as distinct journal rows, while multiple MIR certs crediting the same account from the same pot in one boundary fold into a single row, matching cardano-ledger's per-(pot, credential) `InstantaneousRewards` accumulation.
```
chrisguiney: This sentence states that the fold keyed by pot and stake credential matches cardano-ledger's InstantaneousRewards accumulation without qualification, while ledger/mir.go:66-70 in this same commit states that protocol majors 2-4 use Map.union and would net only the later amount for two certificates naming one stake credential in a pre-Alonzo epoch. The two documents now disagree. Qualifying this to Alonzo onward, and naming the pre-Alonzo divergence, would keep DATABASE.md accurate across the eras MIR spans.

ledger/mir.go
```diff
@@ -61,6 +61,19 @@ const (
 //     (`Cardano.Ledger.Shelley.Rules.Deleg`) and which this rule credits once
 //     at the boundary. A negative delta therefore reduces an earlier positive
 //     one for the same credential and pot rather than being applied on its own.
+//     The additive fold (`Map.unionWith (<>)`) is `delegTransition`'s
+//     behavior from protocol major 5 (Alonzo) onward
+//     (`hardforkAlonzoAllowMIRTransfer`); majors 2-4 (Shelley, Allegra, Mary)
+//     instead use `Map.union`, left-biased on the newer entry, so two
+//     positive certificates for one credential in a single pre-Alonzo epoch
+//     would net to only the later amount under the reference, while this
+//     fold sums both. This code does not thread protocol version into the
```
chrisguiney: The justification given is non-regression against Dingo's own earlier behavior, which does not settle whether the replayed ledger matches the chain: for majors 2-4 a second certificate naming one stake credential in one epoch credits the sum rather than the later amount, changing reward-account balances on a Shelley-to-Mary replay. GetProtocolVersion lives in this package and the boundary caller already holds the era pparams, so threading the major version is available. If the divergence is being accepted instead, a tracking issue would make it visible outside this comment.

database/models/mir.go
```diff
@@ -36,8 +36,11 @@ type MoveInstantaneousRewards struct {
 // epoch-boundary application logic. One of OtherPot > 0 (pot-to-pot transfer)
 // or len(Rewards) > 0 (credential distribution) will be non-empty.
 type MIREffect struct {
-	// ID is the move_instantaneous_rewards row ID, used to order the
-	// certificates of an ended epoch before their deltas are folded.
+	// ID is the move_instantaneous_rewards row ID. Ordering comes from the
+	// store query (`ORDER BY added_slot ASC, id ASC` in
+	// GetMIRCertsInSlotRange) and from consuming effects in receipt order,
+	// not from this field; ID itself is only read for the "MIR %d carries no
```
chrisguiney: ID is read for more than that message: database/plugin/metadata/sqlstore/certificates.go:155 passes effects[i].ID as the mir_id parameter of the query that loads each certificate's reward rows, so it is the join key the store depends on. As written, a reader of this doc comment on an exported field would conclude that ID is nearly vestigial and safe to drop.

chrisguiney (changes requested):
Six findings: an inbound pot overflow is now dropped with no reason and surfaces as an over-budget warn with zero distributed; the sibling unknown-pot error at ledger/mir.go:313 still wedges the rollover; addTransfer's two converted branches have no test; DATABASE.md:3401 and ledger/mir.go:66-70 disagree on whether the fold matches the reference pre-Alonzo; the MIREffect.ID comment omits the store's mir_id join use. Merged with current main, build, vet, lint, docs-parity, import-boundaries, conformance and the MIR tests under -race all pass.

arepala-uml:
Addressed the remaining review findings on top of @wolf31o2's discard-conversion commit (7108f960):

- Threaded the ended epoch's era into `applyMIRCerts`/`collectMIRBoundary` so the distribution fold actually sums for era >= Alonzo and replaces with the latest certificate's amount below Alonzo, matching `delegTransition`'s protocol-version split instead of only documenting the divergence.
- Routed `mirAvailablePot`'s inbound-overflow case through `boundary.discard` (new `mirBoundary.availablePot` method) instead of the generic over-budget warning, so the log names the actual overflow rather than reporting zero distributed against it.
- Added tests for the previously uncovered `addTransfer` branches (unknown source pot, transfer-total overflow), the pre-Alonzo replace-vs-sum behavior, and the discard log message.
- Updated `DATABASE.md` to describe the actual implemented era-conditional fold rather than a cross-reference to an unimplemented comment.

Verified: `go build ./...`, `go test -race ./ledger/...`, `go vet ./ledger/...`, `golangci-lint run ./ledger/... ./database/models/...`, `make docs-parity`, and `go test ./internal/test/conformance/...` — all pass.

ledger/mir_test.go
```diff
@@ -521,6 +630,75 @@ func TestApplyMIRCerts_PotTransferOverflow(t *testing.T) {
...
+	t.Parallel()
+
+	ls, db, gdb := newMIRTestLedger(t)
+	maxUint := ^uint64(0)
+	seedMIRPotTransfer(t, gdb, mirPotReserves, maxUint, 200)
+	seedMIRPotTransfer(t, gdb, mirPotReserves, maxUint, 400)
+	require.NoError(t, db.Metadata().SetNetworkState(1_000, 10_000, 50, nil))
+
+	require.NoError(t, applyMIRCertsErr(ls, db, 0, 1_000),
+		"a transfer total overflow must not fail the epoch boundary")
```
chrisguiney, line 755: Deleting addTransfer's total-overflow guard in place leaves this test green: reservesOut wraps to 2^64-2, availablePot takes its out > available branch, and the boundary falls into the generic over-budget no-op with treasury=1000, reserves=10000 and no error, which is all this asserts. The only observable difference is the log line, so the discard the test names is unpinned. TestApplyMIRCerts_PotTransferOverflowLogsDiscardReason shows what closes it: assertLogContains on discarding uncreditable MIR at epoch boundary plus the transfer-total reason.

ledger/mir_test.go
```diff
@@ -95,30 +99,66 @@ VALUES (?, ?, ?)`,
...
 	})
 }
 
+func TestApplyMIRCerts_UnknownPotDiscardsBoundary(t *testing.T) {
+	t.Parallel()
+
+	ls, db, gdb := newMIRTestLedger(t)
+	seedMIRDistribution(t, gdb, 2, 500, nil)
+
+	assert.NoError(t, applyMIRCertsErr(ls, db, 0, 1_000))
```
chrisguiney, line 182: Replacing collectMIRBoundary's unknown-pot branch with a bare continue in place leaves this test green: the seeded cert carries no reward rows, so skipping it yields an empty boundary, no error and no state change. The assertion pins does-not-fail-the-rollover but not discards-the-boundary, which is the half that changed here. Seeding a valid reserves distribution alongside the pot=2 row separates them, since a discard drops that credit while a skip still applies it.

ledger/mir.go
```diff
@@ -61,6 +62,17 @@ const (
...
 //     at the boundary. A negative delta therefore reduces an earlier positive
 //     one for the same credential and pot rather than being applied on its own.
+//     The additive fold (`Map.unionWith (<>)`) is `delegTransition`'s
+//     behavior from protocol major 5 (Alonzo) onward
+//     (`hardforkAlonzoAllowMIRTransfer`); majors 2-4 (Shelley, Allegra, Mary)
+//     instead use `Map.union`, left-biased on the newer entry, so two
+//     positive certificates for one credential in a single pre-Alonzo epoch
+//     net to only the later amount under the reference rather than their
+//     sum. This fold branches on the ended epoch's era to match: it sums for
+//     era >= Alonzo and replaces with the latest certificate's amount
```
chrisguiney, line 72: The description on this pull request still says summing is kept unconditionally and that both overflow branches are unreachable, which this head reverses. The fold is now era-conditional, so a pre-Alonzo epoch credits only the latest certificate's amount and replayed reward-account balances change for Shelley, Allegra and Mary. Reverting the gate to unconditional additive fails TestApplyMIRCerts_PreAlonzoSameCredentialReplaces, so the behavior is real; the description should name it, since it is the consensus-visible part of the change.

ledger/mir.go
```diff
@@ -458,29 +492,39 @@ func (ls *LedgerState) applyMIRCredits(
...
+// over-budget warning, so the log names the actual overflow instead of
+// reporting zero distributed against it.
+func (b *mirBoundary) availablePot(
 	name string,
-	balance, in, out uint64,
-) (available uint64, ok bool, err error) {
+	balance uint64,
+) (available uint64, ok bool) {
+	in, out := b.reservesIn, b.reservesOut
+	if name == "treasury" {
```
chrisguiney: addTransfer and addCredit both select their pot fields from a uint pot compared against mirPotReserves and mirPotTreasury; this one selects from a string, and any value other than treasury silently resolves to the reserves fields. Taking the same uint pot here keeps the three selectors on one representation and makes an unhandled value a default-branch matter rather than a silent fold into the wrong pot.

ledger/mir.go
```diff
@@ -166,6 +168,18 @@ func (ls *LedgerState) applyMIRCerts(
 		SetNetworkState(newTreasury, newReserves, boundarySlot, txn.Metadata())
 }
 
+// discardMIRBoundary logs why a folded MIR boundary cannot be credited at
+// all, the shared log line for every boundary.discard reason regardless of
+// which step set it.
+func (ls *LedgerState) discardMIRBoundary(boundarySlot uint64, reason string) {
```
chrisguiney, line 174: Prior findings, each rechecked by reverting the fix in place at this head: the inbound pot overflow now names its reason (removing availablePot's discard fails TestApplyMIRCerts_PotTransferOverflowLogsDiscardReason); the distribution unknown-pot error is now a discard (restoring fmt.Errorf fails TestApplyMIRCerts_UnknownPotDiscardsBoundary); the era gate matches hardforkAlonzoAllowMIRTransfer, pvMajor > 4, with era id 4 at Alonzo; DATABASE.md and the fold comment agree; MIREffect.ID names the mir_id join.

chrisguiney (changes requested):
Re-review of 7108f960..29e80251. All six prior findings are cleared, each confirmed by reverting the fix in place rather than by reading the diff. Two new tests still pass when the code branch they name is deleted, so those converted discard paths stay unpinned; details inline. Merged with origin/main at 771b742d locally: build, vet, golangci-lint ./ledger/..., docs-parity, import-boundaries, conformance and the MIR and rollover tests under -race all pass. CodeRabbit declined this head with incremental reviews disabled and cubic has no record, so no bot threads are open.

arepala-uml:
Addressed the second review round:

- \`availablePot\` now takes the same uint \`pot\` (\`mirPotReserves\`/\`mirPotTreasury\`) that \`addTransfer\`/\`addCredit\` switch on, with a default-branch discard, instead of a separate string name.
- \`TestApplyMIRCerts_TransferTotalOverflowIsNoOp\` now asserts the discard log's specific reason text. Confirmed by reverting \`addTransfer\`'s own overflow guard in place: without this change the test still passed, because the same wraparound also trips the unrelated treasury-side inbound-overflow check and discards with a different message — the new assertion catches that.
- \`TestApplyMIRCerts_UnknownPotDiscardsBoundary\` now seeds a valid credit alongside the unknown-pot row and asserts it is *not* applied. Confirmed by reverting the discard to a bare skip/continue in place: the test now fails (the valid credit gets applied), where before it couldn't tell the two apart.
- Updated the PR description to name the era-conditional fold as the consensus-visible part of this change (pre-Alonzo now replaces instead of summing a repeated (pot, credential) delta).

Verified: \`go build ./...\`, \`go test -race ./ledger/...\`, \`go vet ./ledger/...\`, \`golangci-lint run ./ledger/... ./database/models/...\`, \`make docs-parity\`, \`make import-boundaries\`, \`go test ./internal/test/conformance/...\` — all pass.

ledger/mir.go
arepala-uml: Fixed: mirAvailablePot's inbound-overflow branch now sets boundary.discard (routed through the mirBoundary.availablePot method) instead of returning a bare ok=false, so applyMIRCerts logs "discarding uncreditable MIR at epoch boundary" with the specific overflow reason instead of the generic over-budget warn. Covered by TestApplyMIRCerts_PotTransferOverflowLogsDiscardReason, which asserts the log text directly.

ledger/mir.go
arepala-uml: Fixed (in 7108f960, carried forward): the sibling unknown-pot check for a distribution certificate now sets boundary.discard and returns instead of a hard fmt.Errorf, matching addTransfer's default branch. Covered by TestApplyMIRCerts_UnknownPotDiscardsBoundary, which now also seeds a valid credit alongside the bad row to confirm the whole boundary is dropped, not just the bad row skipped.

ledger/mir.go
arepala-uml: Fixed: added TestApplyMIRCerts_UnknownSourcePotIsNoOp (pot=2 transfer) and TestApplyMIRCerts_TransferTotalOverflowIsNoOp (two maxUint transfers from one pot). The latter was initially unpinned per your later comment; it now asserts the specific discard log reason so it fails if addTransfer's own overflow guard is removed.

DATABASE.md
arepala-uml: Fixed: DATABASE.md's GetMIRCertsInSlotRange section now states the fold is protocol-era-conditional (additive for major 5/Alonzo onward, replaced by the latest certificate's amount for majors 2-4), matching the qualified comment in ledger/mir.go instead of an unconditional claim.

ledger/mir.go
arepala-uml: Fixed: threaded the ended epoch's era (epochEraID) into applyMIRCerts/collectMIRBoundary. The fold now actually branches — additive for era >= Alonzo, replaced by the latest certificate's amount below Alonzo — rather than only documenting the divergence. Covered by TestApplyMIRCerts_PreAlonzoSameCredentialReplaces (shelley.EraIdShelley) alongside the existing TestApplyMIRCerts_MultipleDistributionsSameAccount, which now runs at the default Alonzo-era boundary as the additive control. Also called out in the PR description as the consensus-visible part of this change.

database/models/mir.go
arepala-uml: Fixed (in 7108f960, carried forward): MIREffect.ID's doc comment now says ordering comes from the store query, and that this value is also the mir_id join key for reward rows, rather than implying ID is only read for the no-delta error message.

ledger/mir_test.go
arepala-uml: Fixed: the test now captures the logger and asserts assertLogContains on "discarding uncreditable MIR at epoch boundary" plus "MIR pot transfer total overflow". Verified by temporarily removing addTransfer's own overflow guard: the old assertions still passed (the treasury-side inbound-overflow check catches the wraparound with a different message), but the new log-text assertion correctly fails, since it names addTransfer's own reason specifically.

ledger/mir_test.go
arepala-uml: Fixed: the test now seeds a valid reserves distribution alongside the pot=2 row and asserts the credited account's reward stays 0 and reserves stay untouched. Verified by temporarily replacing the discard with a bare continue: the test now fails (the valid credit gets applied), where before it couldn't tell discard apart from skip.

ledger/mir.go
arepala-uml: Fixed: updated the PR description's Summary section to lead with the consensus-visible change — the fold is now protocol-era-conditional, and a pre-Alonzo epoch with more than one MIR cert naming the same stake credential now nets to the later certificate's amount instead of the sum, changing replayed reward-account balances for Shelley/Allegra/Mary.

ledger/mir.go
arepala-uml: Fixed: availablePot now takes a uint pot (mirPotReserves/mirPotTreasury) and switches on it the same way addTransfer and addCredit do, with a default branch that discards on an unhandled value, instead of a separate string name that silently resolved anything but "treasury" to reserves.

ledger/mir.go
arepala-uml: Acknowledged, thanks for confirming by reverting each fix in place rather than just reading the diff. The two unpinned tests you flagged inline (TestApplyMIRCerts_TransferTotalOverflowIsNoOp and TestApplyMIRCerts_UnknownPotDiscardsBoundary) are fixed in 30d59df4 — replied on each thread with how. availablePot's string-vs-uint inconsistency is also fixed there.

wolf31o2 (dismissed):

ledger/mir_test.go
```diff
@@ -521,6 +662,88 @@ func TestApplyMIRCerts_PotTransferOverflow(t *testing.T) {
...
+}
+
+// TestApplyMIRCerts_UnknownSourcePotIsNoOp verifies that a pot-to-pot
+// transfer naming a source pot other than reserves (0) or treasury (1)
+// discards the boundary as a no-op rather than failing the epoch rollover: a
+// hard error here would wedge the node, since the stored certificate is
+// re-read and re-fails on every deterministic retry. See
+// TestApplyMIRCerts_UnknownPotDiscardsBoundary for the equivalent
+// distribution-certificate case.
+func TestApplyMIRCerts_UnknownSourcePotIsNoOp(t *testing.T) {
```
chrisguiney, line 701: Replacing addTransfer's default branch with a bare return in place leaves this test green: the boundary is then empty, isEmpty() returns true, and applyMIRCerts returns nil with the same untouched 1000/10000 state asserted here. Making that branch fatal does fail this test, so the error-to-no-op half is pinned, but discard-the-whole-boundary is not, and that is the half issue 4028 item 3 asks for. Seeding a valid reserves distribution alongside the pot=2 transfer separates them, as TestApplyMIRCerts_UnknownPotDiscardsBoundary now does for the distribution case.

chrisguiney (changes requested):
Re-review of 29e80251..30d59df4. Both prior findings are cleared, each confirmed by reverting the fix in place rather than reading the diff. One member of the same discard class is still unpinned; details inline, and it is my omission from last round rather than a new regression. I reverted every other discard path at this head and each fails its test correctly, so this is the last member. Merged with origin/main at b5492cad locally: build, vet, golangci-lint, docs-parity, import-boundaries, conformance and ./ledger/ under -race all pass. CodeRabbit declined this head; cubic has no record.

ledger/mir_test.go
arepala-uml: Fixed: seeded a valid reserves distribution alongside the pot=2 transfer, the same way TestApplyMIRCerts_UnknownPotDiscardsBoundary does for the distribution case, and assert the credit is not applied. Verified by replacing addTransfer's default branch with a bare return (no discard) in place: the test now fails (the valid credit gets applied), where before it passed under both discard and skip.

arepala-uml:
Addressed the third review round — the one remaining member of the discard-vs-skip class:

- \`TestApplyMIRCerts_UnknownSourcePotIsNoOp\` only asserted the rollover doesn't fail, not that the whole boundary is discarded. Now seeds a valid reserves distribution alongside the pot=2 transfer, mirroring \`TestApplyMIRCerts_UnknownPotDiscardsBoundary\`'s pattern for the distribution case, and asserts the valid credit is *not* applied.
- Confirmed by replacing \`addTransfer\`'s default branch with a bare \`return\` (no discard) in place: the test now fails, where before it passed under both discard and a plain skip.

Verified: \`go build ./...\`, \`go test -race ./ledger/...\`, \`go vet ./ledger/...\`, \`golangci-lint run ./ledger/... ./database/models/...\`, \`make docs-parity\`, \`make import-boundaries\`, \`go test ./internal/test/conformance/...\` — all pass.

chrisguiney (approved):

Issue #4358, opened 2026-09-16
Bootstrap import drops pending pool retirements, forking the node at a later epoch boundary


wcatz:
## Problem

`parsePStateMaps` (`ledgerstate/certstate.go:646`) decodes the Shelley `PState = [poolParams, futurePoolParams, retiring, poolDeposits]` by scanning for CBOR maps, taking the **largest** one as `poolParams`, then merging `poolDeposits`. The **`retiring` map is never decoded** — there is no parser for it anywhere in the package, and `retiring` appears only in comments.

So a bootstrap from a Mithril snapshot imports every pool registration but **no pending pool retirements**. The importer does synthesize tombstones for pools that retired at or before the snapshot epoch (`synthesizeRetiredScheduledPools`), which covers the past; nothing covers a retirement scheduled for a future epoch. Retirements learned from live blocks after the bootstrap are recorded correctly, so the gap is specific to import.

The consequence is a silent consensus divergence that only detonates later. At the next epoch boundary, POOLREAP refunds nothing for the affected pool, so the node's view of that pool's reward account is short by exactly the pool deposit (500 ADA on mainnet). The node keeps following the chain until some block contains a withdrawal from that account, at which point it rejects a block the whole network accepted and cannot recover.

## Observed on mainnet

A node bootstrapped at slot 197789347 imported **2,896 `pool_registration` rows and 0 real `pool_retirement` rows** (the single row present is a zero-valued synthesized tombstone).

Pool `pool18pnja3z8vfkw3sj578xetj3ujhyp2vmljezjg5es290awqmzq7g` (BEACN) retired at epoch 656 — Koios: `pool_status=retired`, `retiring_epoch=656`, `reward_addr=stake1uxnz8f4sayzqrqmukgdz6rruw9pu74qsuxgeg59q9dmcf6gjm22rk`. At the 655→656 boundary the network refunded its 500 ADA deposit; this node did not. It then rejected the block at slot 198029129:

```
conway utxo validation rule 49: incorrect withdrawal amount for
stake1uxnz8f4say...gjm22rk: provided 791963174, balance 291963174
```

791963174 − 291963174 = 500000000, and the node's `pool_registration` row still reads `deposit_held=500000000`. The ledger pipeline then hit `consecutive_no_progress=50` and halted permanently. Restarting cannot help: the retirement is simply absent from the database.

[...]

PR #4360, opened 2026-09-16
fix(ledgerstate): import pending pool retirements from PState

wcatz:
The bootstrap importer decoded the PState `poolParams` and `poolDeposits` maps but never the `retiring` map, so a Mithril-imported node learned nothing about a retirement the snapshot had already scheduled. The pool kept its deposit past the epoch the network refunded it at, the node's view of that pool's reward account fell short by 500 ADA, and it rejected the first block withdrawing from that account under Conway rule 49 — forking off mainnet at an epoch boundary hours or days after the import that caused it. This is #4358, hit in production: three pools retired at epoch 656 and the node had retirement rows for none of them.

This decodes the `retiring` map and writes its rows through `RetirePools` once the registrations are committed. PState layouts differ by era (Shelley four elements, Conway seven with no fixed order), so the map is identified by shape rather than position: pool-key-hash keys, values all below one ADA, a majority naming pools `poolParams` also registered. That excludes `poolDeposits`, whose values are lovelace amounts, and `futurePoolParams`, whose array values fail the uint64 decode. These pools stay live — only the retirement row is written, so deposit, rewards and leader eligibility are untouched until the boundary — and the import now fails loudly if a pool carrying a scheduled retirement is missing from the pool table rather than letting `RetirePools` skip it and silently reopen the gap.

Tests: new parser cases covering decode, deposits-not-mistaken-for-retirements, rejection of a non-retiring small-integer map, tolerance of one unparsed pool, and the Conway layout; verified to fail without the fix. `./ledgerstate/... ./database/...` green with `-race` (21 packages). DATABASE.md updated for the new `certificate_id = 0` row kind, whose `epoch` is in the future; ARCHITECTURE.md checked, not affected.

---
## Summary by cubic
Fixes Mithril bootstrap so imported nodes now see pool retirements that PState had already scheduled, keeping deposit refunds and reward-account balances aligned at the retirement boundary and avoiding later withdrawal rejection.

**Import details**

- Detects the `retiring` map across Shelley and Conway layouts without confusing it with deposits or future pool parameters.
- Keeps scheduled pools active until their retirement epoch, including deposits, rewards, and leader eligibility.
- Retains retirement entries even when a pool's params fail to parse or the pool is absent, so one bad pool doesn't drop the others' retirements.
- Writes synthetic retirement rows in epoch order after registrations commit, rejecting any epoch at or before the snapshot.
- Fails loudly and atomically, before writing anything, if a scheduled pool is missing from the pool table or a key hash is malformed.
- Adds coverage for map detection, malformed or unknown entries, unparsed pools, atomic failure, and Conway PState layouts.

<sup>Written for commit 1761dfb4cf167921ff0d21acc648060053e6d95b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4360?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added support for recognizing scheduled stake pool retirements and their retirement epochs.
  - Pool imports now preserve pending retirement details and associated deposits.
  - Retiring pool records are processed in epoch order, with validation for referenced pools and key hashes.
  - Documented synthetic pending-retirement records and the refund of deposits at the retirement epoch.

- **Bug Fixes**
  - Improved handling of retirement data across different ledger state formats without confusing it with deposit data.


ledgerstate/import.go
```diff
@@ -1316,6 +1330,136 @@ func importPools(
...
+				epoch,
+				err,
+			)
+		}
+		// The snapshot carries no certificate slot for a pending
+		// retirement, so the registrations' own added slot is reused.
+		// Cert ordering ties break on added_slot, which keeps these rows
+		// sorting with the rest of the import and behind any retirement
+		// certificate seen in a later live block.
+		if err := store.RetirePools(
```
chrisguiney, line 1409: This row shares added_slot and certificate_id = 0 with the registration ImportPool just wrote. GetRetiringPools (sqlstore/pool.go:1119) needs (added_slot, synth, block_index, cert_index) strictly above the registration's, so /pools/retiring omits these pools while GetPoolsRetiringAtEpoch returns them: seeding this shape yields 1 refund row and 0 retiring rows, and moving the retirement one slot later yields 1. The other five latest-retirement queries carry a synthetic-row carve-out for exactly this tie; GetRetiringPools does not.

ledgerstate/import.go
```diff
@@ -1316,6 +1330,136 @@ func importPools(
...
+		"epochs", len(epochs),
+	)
+	return nil
+}
+
+// assertPoolsImported fails the import when a pool carrying a scheduled
+// retirement has no row in the pool table. RetirePools silently skips a key
+// hash it cannot resolve, which would reproduce the very gap these rows
+// exist to close, so the mismatch is surfaced instead of logged.
+func assertPoolsImported(
```
chrisguiney, line 1439: Every key reaching this guard was collected from a pool the loop above imports in the same iteration, and ImportPool upserts unconditionally, so neither the malformed-key-hash branch nor the not-found branch can fire and the loud failure described for a scheduled pool missing from the pool table does not happen. The case it targets, a retiring key whose poolParams entry failed to parse, is dropped by mergePoolRetirements before it gets here. Returning the selected retiring map alongside the parsed pools would make this guard reachable.

ledgerstate/import.go
```diff
@@ -1316,6 +1330,136 @@ func importPools(
...
+// each key hash against the pool table and so needs the registrations
+// written first.
+//
+// These pools are still live -- they keep their deposit, earn rewards and
+// lead slots until the retirement epoch -- so only the retirement row is
+// written here. Without it the node holds the deposit past the boundary
+// where the network refunds it to the pool's reward account, the two
+// ledgers then disagree about that account's balance, and the node rejects
+// the first block withdrawing from it.
+func importPendingPoolRetirements(
```
chrisguiney, line 1359: The five new tests exercise parsePState only; nothing covers this function, assertPoolsImported, the epoch sort, or the row RetirePools writes. Reverting the mergePoolRetirements call in place fails 3 of the 5 (Decodes, KeepsRetirements, ConwayDecodes) and leaves DoesNotReadDeposits and RejectsUnknownPools green either way, so those two are guards rather than regression cover. A database-level test asserting that GetPoolsRetiringAtEpoch and GetRetiringPools both see the imported row would have caught the first finding.

ledgerstate/certstate.go
```diff
@@ -764,6 +768,92 @@ func looksLikeDeposits(m map[string]uint64) bool {
...
+		}
+		if epoch >= minDepositLovelace {
+			return false
+		}
+		if _, ok := known[keyHash]; ok {
+			recognized++
+		}
+	}
+	// Use multiplication to avoid integer division rounding.
+	return recognized*2 >= len(m)
```
chrisguiney, line 897: The only thing separating poolDeposits from retiring here is that at least one deposit value reaches 1 ADA. A deposits map whose values are all below minDepositLovelace passes every test in this predicate, because its keys are by definition pools poolParams also registered, and mergePoolRetirements would then schedule every pool to retire at its own deposit amount. Anchoring on the element index for the known Shelley layout, or bounding the accepted values against the snapshot epoch, would remove that.

chrisguiney (changes requested):
The bootstrap now writes a pending retirement at the same added_slot and certificate_id = 0 as its own registration, and GetRetiringPools needs a strict tuple greater-than, so /pools/retiring omits exactly the pools POOLREAP will refund. assertPoolsImported cannot fail as written, and nothing covers the 144 new lines in import.go. CodeRabbit posted one finding on certstate.go 815-816; it is confirmed, with the mechanism and a concrete shape for the fix in the comment on assertPoolsImported.

ledgerstate/import.go
```diff
@@ -1316,6 +1340,136 @@ func importPools(
...
+				epoch,
+				err,
+			)
+		}
+		// The snapshot carries no certificate slot for a pending
+		// retirement, so the registrations' own added slot is reused.
+		// Cert ordering ties break on added_slot, which keeps these rows
+		// sorting with the rest of the import and behind any retirement
+		// certificate seen in a later live block.
+		if err := store.RetirePools(
```
chrisguiney, line 1409: Still open. This writes certificate_id 0 at the registration's own added_slot, so both rows reduce to (added_slot, synth=1, 0, 0), and GetRetiringPools (sqlstore/pool.go:1118) needs the retirement tuple strictly above the registration's. Seeding a pool through this path yields 1 row from GetPoolsRetiringAtEpoch and 0 from GetRetiringPools; writing the same retirement at slot+1 yields 1. The five sibling queries survive the same tie through their ret.epoch > current arm, which GetRetiringPools does not have, so the refund fires while /pools/retiring shows nothing.

ledgerstate/certstate.go
```diff
@@ -764,6 +784,103 @@ func looksLikeDeposits(m map[string]uint64) bool {
...
+				pools[j].PoolKeyHash,
+			)]
+			if !ok {
+				continue
+			}
+			pools[j].RetiringEpoch = &epoch
+			result[epoch] = append(result[epoch], slices.Clone(pools[j].PoolKeyHash))
+		}
+		for keyHash, epoch := range retiring {
+			if _, ok := known[keyHash]; !ok {
```
chrisguiney, line 856: Keys absent from poolParams now reach assertPoolsImported, which returns "pool %x not found in the pool table" and fails the entire cert-state import. looksLikeRetiringEpochs accepts a map up to half of whose keys are unknown, and the doc comment above it still states that unknown keys simply match no pool in the merge. A snapshot holding one undecodable pool that is also retiring imports with a warning on main and cannot import at all here. Decide it once: drop unknown keys before the importer, or make the abort the documented and tested contract.

ledgerstate/certstate.go
```diff
@@ -632,18 +641,25 @@ func parsePoolDelegation(data []byte) ([]byte, bool) {
 // parsePState decodes the pool state.
 // PState = [poolParams, futurePoolParams, retiring, poolDeposits]
 func parsePState(data []byte) ([]ParsedPool, error) {
+	pools, _, err := parsePStateWithRetirements(data)
+	return pools, err
+}
+
+func parsePStateWithRetirements(
```
chrisguiney, line 657: The red lint check at this head is here: parsePState (line 643) and parsePStateConway (line 285) now have no non-test caller, and .golangci.yml sets run.tests to false, so unused flags both. golangci-lint run ./ledgerstate/... reports these 2 issues at this head and 0 issues at the merge base, so the failure belongs to this change. Either fold the two wrappers into the tests that still call them, or delete them and have those tests call the WithRetirements variants directly.

ledgerstate/import.go
```diff
@@ -1316,6 +1340,136 @@ func importPools(
...
+		"epochs", len(epochs),
+	)
+	return nil
+}
+
+// assertPoolsImported fails the import when a pool carrying a scheduled
+// retirement has no row in the pool table. RetirePools silently skips a key
+// hash it cannot resolve, which would reproduce the very gap these rows
+// exist to close, so the mismatch is surfaced instead of logged.
+func assertPoolsImported(
```
chrisguiney, line 1439: This guard is reachable now, which clears the earlier point that it could not fire. Coverage has not followed it: nothing drives importPendingPoolRetirements, this guard, the epoch sort, or the row RetirePools writes, and the two existing importPools tests set no RetiringEpoch so they return before the call. A database-level test asserting that GetPoolsRetiringAtEpoch and GetRetiringPools both see the imported row pins the first finding, and one passing a retiring key with no pool row pins this abort.

ledgerstate/import.go
```diff
@@ -1218,6 +1234,14 @@ func importPools(
 		default:
 		}
 
+		if len(retirementMaps) == 0 && pool.RetiringEpoch != nil {
```
chrisguiney: Arity decides whether pool.RetiringEpoch is read, so a caller handing over a nil map silently disables this fallback while a caller handing over nothing keeps it. Both production sites happen to be right today (line 1041 always passes the map, line 1624 never does), but the coupling is invisible at the call site and no test pins it. A plain map parameter, with the fallback keyed on the map being nil rather than on how many arguments arrived, states the same rule without the mode switch.

ledgerstate/certstate.go
```diff
@@ -764,6 +784,103 @@ func looksLikeDeposits(m map[string]uint64) bool {
...
+) bool {
+	if len(m) == 0 {
+		return false
+	}
+	var recognized int
+	for keyHash, epoch := range m {
+		if len(keyHash) != 2*poolKeyHashLen {
+			return false
+		}
+		if epoch >= minDepositLovelace {
```
chrisguiney, line 889: Still open from the previous round. A poolDeposits map whose values are all below minDepositLovelace passes every test in this predicate, because its keys are by definition pools that poolParams also registered, and mergePoolRetirements would then schedule every pool to retire at its own deposit amount. Anchoring on the element index for the known Shelley layout, or bounding the accepted values against the snapshot epoch, would remove that.

chrisguiney (changes requested):
Three blockers at this head. The imported pending retirement is still invisible to GetRetiringPools, so /pools/retiring omits exactly the pools POOLREAP refunds. Retaining unmatched retiring keys now aborts the whole cert-state import through assertPoolsImported, which the comment above looksLikeRetiringEpochs still describes as harmless. Lint is red on parsePState and parsePStateConway being unused, and clean at the merge base. The CodeRabbit thread on certstate.go 833 is addressed by this commit, and its fix is what surfaced the second blocker. The merged tree builds and ledgerstate passes.

database/plugin/metadata/sqlstore/pool.go
```diff
@@ -1117,6 +1117,11 @@ WHERE r.rn = 1 AND r.epoch > ?
       g.pool_key_hash IS NULL
       OR (r.added_slot, r.synth, r.block_index, r.cert_index)
          > (g.added_slot, g.synth, g.block_index, g.cert_index)
+      OR (
+          (r.added_slot, r.synth, r.block_index, r.cert_index)
+             = (g.added_slot, g.synth, g.block_index, g.cert_index)
```
chrisguiney, line 1122: Cleared: this arm fixes the tie raised in both earlier rounds. Seeding the bootstrap shape -- ImportPool at slot S, then RetirePools(epoch 656, slot S), both certificate_id = 0 -- gives 1 row from GetRetiringPools(650) and 1 refund from GetPoolsRetiringAtEpoch(656, S+1); deleting these five lines in place gives 0 and 1, so the fix is causal. Nothing pins it. The pool parity exercise retires at slot 30 against a registration at slot 10, which passes either way. A same-slot case beside it would hold this query and its six siblings together.

ledgerstate/certstate.go
```diff
@@ -785,6 +791,103 @@ func looksLikeDeposits(m map[string]uint64) bool {
...
+// looksLikeRetiringEpochs reports whether m is plausibly the PState
+// `retiring` map: non-empty, keyed by pool key hashes, every value small
+// enough to be an epoch number rather than a lovelace deposit, and mostly
+// naming pools that poolParams also registered.
+//
+// The poolParams check is a majority rather than a requirement on every
+// key. A retiring pool whose params entry failed to parse is absent from
+// pools, and rejecting the whole map over one such key would drop every
+// other pool's retirement too -- the same all-or-nothing loss this decode
+// exists to prevent. Unknown keys simply match no pool in the merge.
```
chrisguiney: This sentence is now false, and it is the one that states the contract. mergePoolRetirements (line 845) retains keys absent from poolParams, importedPoolKeys reports them absent, and importPendingPoolRetirements fails the whole bootstrap with "pools scheduled to retire were not imported". A snapshot holding one pool whose PoolParams this package cannot decode imports with a warning on main (import.go:976) and cannot import at all here. The description states the loud failure and the code comment states the opposite; make the abort the documented and tested contract.

ledgerstate/import.go
```diff
@@ -1316,9 +1341,144 @@ func importPools(
...
+	}
+
+	if err := txn.Commit(); err != nil {
+		return fmt.Errorf(
+			"committing pool retirements: %w", err,
+		)
+	}
+	if len(missing) > 0 {
+		return fmt.Errorf(
+			"pools scheduled to retire were not imported: %s",
```
chrisguiney: The abort runs after txn.Commit(), so a failed import leaves the resolved retirement rows behind: importPools with one parsed pool retiring at epoch 656 and one unmatched key at epoch 657 returns this error while GetRetiringPools(650) already reports 1 row. Either gather the absent keys and return before committing, or commit deliberately and record why the partial state is safe to leave. Neither branch of this function is exercised by a test today, so whichever is chosen, choose it once and pin it.

ledgerstate/import.go
```diff
@@ -1316,9 +1341,144 @@ func importPools(
...
+// each key hash against the pool table and so needs the registrations
+// written first.
+//
+// These pools are still live -- they keep their deposit, earn rewards and
+// lead slots until the retirement epoch -- so only the retirement row is
+// written here. Without it the node holds the deposit past the boundary
+// where the network refunds it to the pool's reward account, the two
+// ledgers then disagree about that account's balance, and the node rejects
+// the first block withdrawing from it.
+func importPendingPoolRetirements(
```
chrisguiney, line 1359: Still open from the previous round. No test drives importPendingPoolRetirements, the epoch sort, importedPoolKeys, or the row RetirePools writes; the two existing importPools tests pass no retirement map and return at the length check. One test against dbtest.NewDatabase covers it: call importPools with a retirement map, assert GetPoolsRetiringAtEpoch and GetRetiringPools both see the row and that the refund carries the imported deposit, then assert the unmatched-key failure. That pins the new SQL arm and the abort contract at the same time.

ledgerstate/certstate.go
```diff
@@ -785,6 +791,103 @@ func looksLikeDeposits(m map[string]uint64) bool {
...
+) bool {
+	if len(m) == 0 {
+		return false
+	}
+	var recognized int
+	for keyHash, epoch := range m {
+		if len(keyHash) != 2*poolKeyHashLen {
+			return false
+		}
+		if epoch >= minDepositLovelace {
```
chrisguiney, line 889: Still open from both earlier rounds. A poolDeposits map whose values are all below minDepositLovelace passes every test here, because its keys are necessarily pools that poolParams also registered, and mergePoolRetirements would then schedule every pool to retire at its own deposit amount. An epoch already in the past is accepted too, and writing one silently deactivates a live pool: a row at epoch 3 seeded while the node is at epoch 650 leaves GetActivePoolKeyHashes empty. Bounding accepted values against the snapshot epoch removes both.

ledgerstate/import.go
```diff
@@ -1191,6 +1199,7 @@ func importPools(
 	cfg ImportConfig,
 	pools []ParsedPool,
 	slot uint64,
+	retirementMaps ...map[uint64][][]byte,
```
chrisguiney: Still open. Arity decides whether pool.RetiringEpoch is read, so a caller handing over a nil map disables this fallback while a caller handing over nothing keeps it. Both production sites are right today -- line 1042 always passes the map, line 1630 never does -- and no test pins the coupling. A plain map parameter, with the fallback keyed on the map being nil rather than on how many arguments arrived, states the same rule without the mode switch.

chrisguiney (dismissed):
Two blockers remain at this head. The retiring-key abort is still half-made: unknown keys are retained, they fail the whole bootstrap import after the resolved rows have already been committed, and the comment above looksLikeRetiringEpochs still calls them harmless. The GetRetiringPools tie is fixed and the fix is causal, but nothing in the tree tests it. Lint is clean here and every check run is green. The CodeRabbit finding on certstate.go is addressed by these commits; the magnitude heuristic and the variadic retirement map are unchanged.

wolf31o2 (approved):

Issue #3770, opened 2026-09-02
test(conformance): separate ledger and consensus compatibility


wolf31o2:
## Problem

Dingo's conformance documentation and tests identify the ledger corpus as
Amaru conformance. A passing run currently gives too much confidence because
the corpus is primarily transaction/state-transition coverage and does not
exercise the full node behavior required for compatibility.

## Scope

- Reclassify the existing ledger-vector run as ledger-rule coverage and consume
  the pinned shared corpus from `ouroboros-mock`.
- Add a node conformance matrix separating ledger, consensus, networking, and
  end-to-end reference-node compatibility.
- Extend deterministic consensus coverage for competing forks, within-k and
  beyond-k behavior, rollback, intersection, peer scheduling, density, and
  chain-selection tie-breaking.
- Evaluate the feasibility of consuming Tweag's black-box Node-vs-Environment
  runner/test generator; until a stable upstream artifact exists, preserve
  equivalent local scenarios in `ouroboros-mock`.
- Update Dingo README, architecture, and validation documentation so no single
  ledger pass rate is described as complete conformance.
- Ensure CI and release validation report the exact profile and scenario counts.

## Acceptance criteria

- Dingo's validation documentation distinguishes ledger vectors from consensus
  and live cardano-node compatibility testing.
- The default deterministic conformance suite covers the shared ledger corpus
  and reports category-level results.
- Consensus scenarios assert final chain choice, rollback/intersection points,
  and downstream observable events.
- The reference-node profile remains an explicit separate check and is not
  represented as passing when it was not run.
- Architecture and validation documents identify corpus revisions and known
  exclusions.

## Dependency

- Depends on [ouroboros-mock#267](https://github.com/blinklabs-io/ouroboros-mock/issues/267) for shared
  corpus, coverage metadata, and consensus scenarios.

## References

- https://github.com/cardano-scaling/cardano-blueprint/tree/main/src/ledger/conformance-test-vectors
- https://github.com/tweag/cardano-conformance-testing-of-consensus
- https://github.com/IntersectMBO/cardano-ledger

wolf31o2:
## Related existing work

- [#3443](https://github.com/blinklabs-io/dingo/issues/3443) — exercise Dingo
  validation entry points in conformance tests; this is a direct acceptance
  criterion for the ledger layer.
- [#3003](https://github.com/blinklabs-io/dingo/issues/3003) — publish an
  evidence-backed validation coverage matrix; the new issue expands this from
[...]

PR #4376, opened 2026-09-17
test(conformance): separate node compatibility profiles

wolf31o2:
Related to #3770.

- Separate Blueprint ledger-rule, deterministic consensus, and live reference-node profiles.
- Enforce the pinned ledger and consensus corpus counts and report coverage by category.
- Document corpus provenance and the reference-node validation boundary.
- Surface profile counts in CI and release validation output.

The deterministic corpus does not cover peer scheduling or Genesis density.


README.md
```diff
@@ -1061,6 +1061,23 @@ make docs-parity                             # Docs agree with go.mod, Makefile,
...
+### Conformance profiles
+
+Dingo reports compatibility in separate layers. The ledger corpus is the
+pinned Cardano Blueprint archive consumed from `ouroboros-mock`; a green ledger
+result is not complete node conformance.
+
+| Profile | Command | Current scope |
+| --- | --- | --- |
+| Ledger rules | `go test ./internal/test/conformance/` | Blueprint ledger vectors, Dingo era validation entry points, and real SQLite/backend behavior; reports counts by era and rule family |
+| Deterministic consensus | `go test ./ouroboros/ -run TestConsensusConformance` | Five shared scenarios covering origin ingestion, within-k and beyond-k forks, rollback/intersection, peer scheduling, density/tie-breaking, and downstream ChainSync observations |
```
chrisguiney: This row claims the deterministic consensus profile covers density, and it does not. ouroboros-mock v0.20.2 has no density scenario (no match anywhere in its consensus package) and its five captures are single-peer, switch, switch, VRF tie and k-bound no-switch. The replay adapter in consensus_conformance_test.go configures no genesis window. Dingo does implement a Genesis density rule (WithGenesisWindowSlots, ARCHITECTURE.md), so this reads as coverage that does not exist. internal/test/conformance/README.md omits it from the same row; drop density here, and peer scheduling with it.

internal/test/conformance/conformance_test.go
```diff
@@ -34,6 +39,7 @@ import (
 // previous separate statistics pass cost.
 func TestRulesConformanceVectors(t *testing.T) {
 	results := sqliteCorpusResults(t)
+	require.Len(t, results, expectedBlueprintVectorCount)
```
chrisguiney: require.Len calls FailNow, so a corpus count drift aborts before reportCorpus runs and the per-era breakdown never prints: the failure reports only 2574 against 2575, with no indication of which era or rule family moved. Running reportCorpus first and asserting after keeps the diagnostic this change adds available on exactly the failure it exists to catch.

chrisguiney (changes requested):
Merged tree against origin/main 771b742d builds clean and passes ./ouroboros/, TestConsensusConformanceVectors and TestRulesConformanceVectors (2575/2575, 20 coverage groups), scoped golangci-lint and docs-parity; every check on this head is green. The three new guards are causally live: reverting each in place turns its own test red. README corpus provenance matches ouroboros-mock v0.20.2 CORPUS.md exactly. Blocking on one accuracy point in README.md. Neither bot covered this head: CodeRabbit refused on rate limit over 63270683..0c9c9d44, cubic on monthly limit, no reviews-API record.

internal/test/conformance/README.md
```diff
@@ -15,13 +16,37 @@ local SQLite backend (the default, no setup required -- see
 
 ## What the vectors cover
 
-The vectors exercise **Conway era** ledger rules:
+The pinned archive contains ledger-rule vectors across the Shelley, Allegra,
```
chrisguiney: This overstates the corpus: the pinned archive holds exactly one era directory, eras/conway/impl, and TestConformanceVectorsExerciseDingoEraEntryPoints reports at this head eras.ValidateTxConway 2576 transactions with 0 for Byron, Shelley, Allegra, Mary, Alonzo, Babbage and Dijkstra, each labelled as having no vectors in this corpus. The era labels in the new coverage output come from SummarizeCoverage matching <Era>ImpSpec in the vector path, which is the Conway Imp suite re-running earlier eras spec bodies rather than a pre-Conway execution. Line 287 of this same file states it accurately.

internal/test/conformance/README.md
```diff
@@ -15,13 +16,37 @@ local SQLite backend (the default, no setup required -- see
...
+Mary, Alonzo, Babbage, and Conway eras, plus one synthetic rollback fixture:
```
chrisguiney: The CodeRabbit outside-diff finding asks for the opposite change, to delete the Conway-only wording at line 287, and it is rejected: the era routing measured above shows line 287 is the accurate sentence and these two lines are the ones to correct. Applying that suggestion would replace a true statement with the inaccurate one and leave the file self-consistent but wrong.

internal/test/conformance/README.md
```diff
@@ -15,13 +16,37 @@ local SQLite backend (the default, no setup required -- see
...
+
+## Node conformance matrix
+
+| Profile | Source | Default Dingo test | What it proves | What it excludes |
+| --- | --- | --- | --- | --- |
+| Ledger rules | Pinned Cardano Blueprint archive in `ouroboros-mock` | `go test ./internal/test/conformance/` | Dingo's ledger entry points and backend behavior against the shared ledger corpus | Consensus, networking, and reference-node behavior |
+| Deterministic consensus | Shared `ouroboros-mock/consensus` captured scenarios | `go test ./ouroboros/ -run TestConsensusConformance` | Final chain choice, within-k and beyond-k behavior, rollback/intersection points, tie-breaking, and downstream ChainSync observations | Live sockets, full block bodies, and cardano-node process behavior |
+| Reference node | DevNet `--conformance` profile | `./internal/test/devnet/run-tests.sh --conformance` | Dingo beside `cardano-node` in the configured live topology | Not run by the ledger or deterministic consensus profiles |
+
+The deterministic corpus currently contains five scenarios: one origin
```
chrisguiney, line 41: The description says Fixes #3770, but that issue scope also asks for deterministic consensus coverage of peer scheduling and density, and the pinned corpus carries neither: ouroboros-mock v0.20.2 has five consensus scenarios, classified by the new profile log as one single-peer, three fork-switch and one no-switch, and no match for density anywhere in its consensus package. Closing #3770 on merge would drop those two scope items silently. Reference the issue rather than closing it, or add the scenarios.

internal/test/conformance/conformance_test.go
```diff
@@ -35,5 +40,6 @@ import (
 func TestRulesConformanceVectors(t *testing.T) {
 	results := sqliteCorpusResults(t)
 	reportCorpus(t, "sqlite", results)
+	require.Len(t, results, expectedBlueprintVectorCount)
```
chrisguiney: require.Len prints the whole slice on mismatch, so a real count drift emits about 32 KB of VectorResult dump per backend and buries the 20-group era breakdown that the reportCorpus call immediately above exists to supply, which is the diagnostic this commit set out to preserve. require.Equal(t, expectedBlueprintVectorCount, len(results)) reports two integers instead and keeps the reordering worthwhile on all three backends.

.github/workflows/go-test.yml
```diff
@@ -301,7 +301,7 @@ jobs:
         # no stuck test, so an intermediate bound tight enough to be useful is
         # also tight enough to kill a slow-but-healthy run and destroy the
         # attribution this job exists to provide.
-        run: go test -ldflags="-s -w" -timeout 30m ./...
+        run: go test -v -ldflags="-s -w" -timeout 30m ./...
```
chrisguiney, line 304: Two checks are red at this head. go-test (Windows) failed TestCheckpointWALDoesNotBlockWriteBehindReaderSnapshot at checkpoint_test.go:171, 2.2027956s is not less than 2s; that test and its bound arrived on main in #4276 and this branch changes no file under database/plugin/metadata/sqlite, so it reads as a base-branch timing margin, with the caveat that -v is the only branch-owned change to this job and the overrun is 10 percent. docker failed inside go mod download on a proxy.golang.org stream error, which is infrastructure rather than a code defect.

chrisguiney (dismissed):
Both earlier findings are cleared at this head: the density and peer-scheduling claim is gone from README.md, and moving require.Len after reportCorpus is causally verified, since a forced count drift now prints the 20-group era breakdown and printed none under the previous order. The merge audit is clean, 8 of 8 branch files kept with all added names surviving, and the scoped linter, docs-parity and import-boundaries pass. Blocking on the era claim in internal/test/conformance/README.md, which this package own entry-point report contradicts, and on two red checks at this head.

README.md
wolf31o2: Fixed in 6e22587b: the deterministic profile now lists only the five shared scenario contracts and no longer claims density or peer-scheduling coverage.

internal/test/conformance/conformance_test.go
wolf31o2: Fixed in 6e22587b: reportCorpus runs before the corpus-count assertion.

internal/test/conformance/README.md
wolf31o2: Fixed in cc69c08c: the README now identifies the archive as the Conway ImpSpec corpus and distinguishes inherited rule labels from Dingo era-entry-point execution.

internal/test/conformance/README.md
wolf31o2: Fixed in cc69c08c: the earlier-era execution claim was corrected while the accurate Conway-only statement remains intact.

internal/test/conformance/README.md
wolf31o2: Fixed in the live description: #3770 is now related rather than closed, and the uncovered peer-scheduling and Genesis-density scope is explicit.

internal/test/conformance/conformance_test.go
wolf31o2: Fixed in cc69c08c: the count check compares the two integers without dumping the full result slice.

.github/workflows/go-test.yml
wolf31o2: No branch change is required: the Docker failure is a module-proxy stream error, and the Windows failure is an unrelated SQLite timing bound from main.

arepala-uml (approved):

PR #4386, opened 2026-09-17
perf(eras): reuse cached TxInfo across Babbage redeemers

### LINKED ISSUE #4385: Babbage redundant TxInfo rebuild on every redeemer causes allocation/GC pressure
## Problem

`ValidateTxBabbage` and `EvaluateTxBabbage` (`ledger/eras/babbage.go`) each
rebuild a fresh Plutus `TxInfo` (inputs, outputs, certificates, mint,
withdrawals, and the validity interval translated through `SlotToTime`) on
every redeemer, rather than once per transaction:

- Once before the redeemer loop, just to check `.Redeemers`.
- Again inside the `PlutusV1` case, once per V1 redeemer.
- Again inside the `PlutusV2` case, once per V2 redeemer -- even though the
  pre-loop build already produced the identical value.

A live CPU/allocation profile taken during a from-genesis Preview sync
showed the resulting repeated `TxInfo`-to-`PlutusData` conversion
(`WithZeroAdaAsset.ToPlutusData` and its children) at ~17% of all sampled
allocated objects, with corresponding GC background-mark-scan CPU overhead.

Conway's `EvaluateTxConway`/`validateTxPlutusConwayWithContext` already
solved this with a lazily-built, per-transaction `TxInfo` cache
(`conwayTxInfoCache`). Babbage never got the same treatment.

## Fix

Renamed the cache to the era-neutral `txInfoCache`/`newTxInfoCache` (it has
no Conway-specific logic) and reused it from `ValidateTxBabbage`/
`EvaluateTxBabbage`, eliminating the redundant rebuilds.

## Tests

`TestEvaluateTxBabbagePreviewBuildsTxInfoOnce` and
`TestValidateTxBabbagePreviewBuildsTxInfoOnce` count `SlotToTime` calls
(one per validity bound per `TxInfo` build) against the real
`preview-babbage-tx-41098839.cbor` fixture: both fail at 4 calls (two
redundant builds x two validity bounds) before the fix and pass at 2 (one
build) after it.

chrisguiney:
Closes #4385

## Problem

`ValidateTxBabbage` and `EvaluateTxBabbage` rebuilt a fresh Plutus `TxInfo`
(inputs, outputs, certificates, mint, withdrawals, and the validity interval
translated through `SlotToTime`) on every redeemer: once before the loop
just to read `.Redeemers`, then again inside the `PlutusV1` case per V1
redeemer and, redundantly, again inside the `PlutusV2` case per V2 redeemer
even though the pre-loop build already produced the identical value. A live
CPU/allocation profile during a from-genesis Preview sync showed the
resulting repeated `TxInfo`-to-`PlutusData` conversion
(`WithZeroAdaAsset.ToPlutusData` and its children) at ~17% of allocated
objects.

## Fix

Conway's `EvaluateTxConway`/`validateTxPlutusConwayWithContext` already
solved this with a lazily-built, per-transaction `TxInfo` cache
(`conwayTxInfoCache`). Renamed it to the era-neutral `txInfoCache` and
reused it from Babbage instead of duplicating the pattern.

## Tests

`TestEvaluateTxBabbagePreviewBuildsTxInfoOnce` and
`TestValidateTxBabbagePreviewBuildsTxInfoOnce` count `SlotToTime` calls
(one per validity bound per `TxInfo` build) against the real
`preview-babbage-tx-41098839.cbor` fixture: both fail at 4 calls (two
redundant builds x two validity bounds) before this change and pass at 2
(one build) after it.

## Validation

- `go build ./...`, `go vet ./...`, `gofmt -l .` -- clean
- `golangci-lint run ./ledger/eras/...` -- 0 issues
- `go test ./ledger/... -race -count=1` -- pass
- `go test ./internal/test/conformance/...` -- pass (unchanged from base;
  this fix only removes redundant work, no behavioral change)
- `make docs-parity` -- pass

---
## Summary by cubic
Reuses a lazily-built per-transaction Plutus `TxInfo` cache across Babbage redeemers so `ValidateTxBabbage` and `EvaluateTxBabbage` no longer rebuild identical `TxInfo` values for every redeemer. Previously each redeemer triggered a redundant rebuild, plus a pre-loop build just to read `.Redeemers`; now each language version's `TxInfo` is built at most once per transaction. No behavioral change, only removal of redundant work, which showed up as ~17% of allocated objects in a Preview sync profile.

- Renames Conway's `conwayTxInfoCache` to the era-neutral `txInfoCache` and reuses it from Babbage instead of duplicating the pattern.
- Adds `SlotToTime`-counting tests that pin the single-build behavior: preview-fixture tests for the existing PlutusV2 path, plus multi-redeemer transactions covering two and three PlutusV1 redeemers and a mixed PlutusV1/PlutusV2 case.

Closes #4385.

<sup>Written for commit 3c30b366e990758e711d307f6a39591b80a46e3e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4386?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Performance**
  - Improved transaction validation and evaluation efficiency by reusing transaction information during Plutus script processing.
  - Reduced repeated transaction-info construction for transactions with multiple redeemers.

- **Bug Fixes**
  - Ensured transaction information is built only once per language version during Babbage validation and evaluation.

- **Tests**
  - Added coverage confirming transaction information is constructed once per transaction.


ledger/eras/babbage_plutus_preview_fixture_test.go
```diff
@@ -184,3 +185,104 @@ func TestEvaluateTxBabbagePreviewDuplicateRequiredSigner(t *testing.T) {
...
+		41_098_839,
+		ls,
+		previewBabbageProtocolParams(t),
+	)
+	require.NoError(t, err)
+	assert.Equal(
+		t,
+		wantSlotToTimeCallsPerTxInfoBuild,
+		ls.slotToTimeCalls,
+		"TxInfo must be built exactly once per transaction, not once per redeemer",
```
wolf31o2, line 286: These tests exercise only one Plutus V2 redeemer, so they never call txInfos.v1() and do not prove reuse across multiple redeemers. Please add a focused Babbage fixture/test with multiple Plutus V1 redeemers (and a mixed-language case if supported) asserting the same one-build SlotToTime count.

wolf31o2 (changes requested):
Please add coverage for the newly shared cache paths before relying on this optimization.

ledger/eras/babbage_plutus_preview_fixture_test.go
chrisguiney: Covered in ledger/eras/babbage_txinfo_cache_test.go (3c30b366e): a Babbage transaction assembled from CBOR that spends one script-locked UTxO per redeemer, with cases for two PlutusV1 redeemers, three PlutusV1 redeemers, and a mixed PlutusV1/PlutusV2 transaction. Every case asserts the same SlotToTime total of 4 (two TxInfo builds, one per language, times two validity bounds), so the assertion is that the count is a function of the transaction rather than of its redeemer count.

Fail-before: making txInfoCache.v1 rebuild on every call takes the three cases to 6, 8 and 6 calls while TestEvaluateTxBabbagePreviewBuildsTxInfoOnce and TestValidateTxBabbagePreviewBuildsTxInfoOnce both still pass, which is the gap reported here. Restoring the per-redeemer PlutusV2 rebuild instead takes the mixed case to 8.

The mixed-language case is synthetic rather than a captured preview transaction: no fixture in ledger/eras/testdata carries both languages. It is a real transaction shape otherwise, decoded through babbage.NewBabbageTransactionFromCbor and evaluated against the preview PlutusV1 and PlutusV2 cost models already in testdata.

ledger/eras/babbage_plutus_preview_fixture_test.go
```diff
@@ -184,3 +185,104 @@ func TestEvaluateTxBabbagePreviewDuplicateRequiredSigner(t *testing.T) {
...
+	)
+}
```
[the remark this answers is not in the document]
chrisguiney: Covered in ledger/eras/babbage_txinfo_cache_test.go (3c30b366e): a Babbage transaction assembled from CBOR that spends one script-locked UTxO per redeemer, with cases for two PlutusV1 redeemers, three PlutusV1 redeemers, and a mixed PlutusV1/PlutusV2 transaction. Every case asserts the same SlotToTime total of 4 (two TxInfo builds, one per language, times two validity bounds), so the assertion is that the count is a function of the transaction rather than of its redeemer count.

Fail-before: making txInfoCache.v1 rebuild on every call takes the three cases to 6, 8 and 6 calls while TestEvaluateTxBabbagePreviewBuildsTxInfoOnce and TestValidateTxBabbagePreviewBuildsTxInfoOnce both still pass, which is the gap reported here. Restoring the per-redeemer PlutusV2 rebuild instead takes the mixed case to 8.

The mixed-language case is synthetic rather than a captured preview transaction: no fixture in ledger/eras/testdata carries both languages. It is a real transaction shape otherwise, decoded through babbage.NewBabbageTransactionFromCbor and evaluated against the preview PlutusV1 and PlutusV2 cost models already in testdata.

wolf31o2 (approved):

PR #4404, opened 2026-09-17
feat(mithril): support pinned fresh bootstrap artifacts

wolf31o2:
## Summary

- Add YAML, CLI, and environment configuration for selecting an exact Mithril artifact during a fresh bootstrap.
- Treat the configured identity as a v1 snapshot digest or v2 Cardano database artifact hash.
- Reject explicit pins for complete databases before the up-to-date short circuit and for interrupted catch-up runs.
- Preserve the durable artifact identity when resuming an interrupted fresh bootstrap.
- Cover fresh selection, already-current databases, v1 complete databases, interrupted catch-up, and resume conflicts through `Sync`.


mithril/sync.go
```diff
@@ -486,6 +510,12 @@ func Sync(
 	}
 	catchUp = dec.engage
 	catchUpStart = dec.start
+	if mode == syncModeCatchUp && cfg.PinnedDigest != "" {
```
cryptodj413: No Sync()-level test covers this catch-up pin rejection, the interrupted-catch-up rejection at sync.go:556, or that a fresh bootstrap enforces PinnedDigest during artifact selection (bootstrap.go/bootstrap_v2.go); only the extracted helpers are unit-tested. Separately, dec.upToDate short-circuits at line 509 before this check runs, so a pinned request against an already-current database silently no-ops instead of confirming or rejecting the pin.

mithril/sync.go
```diff
@@ -486,6 +510,12 @@ func Sync(
...
+		return SyncResult{}, errors.New(
+			"explicit Mithril artifact pin requires a fresh database; " +
+				"catch-up runs always select the latest compatible artifact",
```
cryptodj413: This message always attributes the rejection to catch-up semantics, but decideCatchUp never engages catch-up for a complete v1-backend database, so on v1 the stated reason for the rejection does not match why it fired.

internal/config/flags.go
```diff
@@ -783,6 +783,12 @@ var flagSpecs = []flagSpec{
 		"",
 		"Mithril artifact backend: v1 (legacy snapshots) or v2 (incremental database)",
 	),
+	stringFlag(
+		"Mithril.PinnedDigest",
+		"mithril-pinned-digest",
+		"",
+		"Mithril artifact digest to use for a fresh bootstrap",
```
cryptodj413: Help text does not note the digest is backend-specific (v1 snapshot digest vs v2 CardanoDatabase hash, bootstrap.go:204); a v1 digest supplied under the default v2 backend fails only as a generic aggregator error, not a config-time message.

cryptodj413 (changes requested):
Pin logic and CodeRabbit's already-fixed Major finding check out; requesting changes only for missing Sync()-level test coverage of the new pin-rejection contracts.

mithril/sync.go
wolf31o2: Fixed in 4d4bcf3a. Sync now rejects a pin for a complete database before the up-to-date short circuit. Added Sync-level coverage for an already-current v2 database, a complete v1 database, an interrupted catch-up, and fresh-bootstrap selection of the requested v2 artifact. Mutation runs confirmed the selection test chooses the newer artifact when pin forwarding is removed and the interrupted-catch-up test reaches the aggregator when its rejection is removed.

mithril/sync.go
wolf31o2: Fixed in 4d4bcf3a. The complete-database rejection is now backend-neutral and occurs before catch-up dispatch; the v1 Sync regression asserts the error names the complete database and does not attribute it to a catch-up run.

internal/config/flags.go
wolf31o2: Fixed in 4d4bcf3a. CLI help, config documentation, the example YAML, and README now state that v1 accepts a snapshot digest while v2 accepts a Cardano database artifact hash.

cryptodj413 (approved):

Issue #3971, opened 2026-09-05
ledger: the MIR pipeline is unsigned end to end and cannot represent a negative reward delta


wolf31o2:
gouroboros #2247 widens `MoveInstantaneousRewardsCertificateReward.Rewards` from `map[*Credential]uint64` to `map[*Credential]*big.Int`, because the reference decodes MIR deltas as signed `DeltaCoin` and gouroboros was failing the whole block on a negative one (gouroboros #2240). Adopting that release requires more than a formatter change here.

Two sites stop compiling:

- `api/blockfrost/adapter.go:4384` — `strconv.FormatUint(amount, 10)`
- `database/plugin/metadata/sqlstore/transaction_certificates.go:1047` — `decimalUint64(types.Uint64(amount))`

The first is a one-line swap to `amount.String()`. The second is not, and that is the point of this issue: dingo represents MIR amounts as unsigned through the entire path.

- `decimalUint64` is `strconv.FormatUint`.
- `decimalAmountText` (`database/plugin/metadata/sqlstore/amounts.go`) **returns an error for a negative amount**.
- `models.MoveInstantaneousRewardsReward.Amount` is `types.Uint64`.
- `MIREffect` and `MIRReward` use `uint64`.
- The `SUM(reward.amount)` reads at `database/plugin/metadata/sqlstore/account.go:846` and `:861` are unsigned.
- The INSTANT-rule arithmetic in `ledger/mir.go` is unsigned.

So a negative delta that now decodes successfully in gouroboros would be rejected or truncated inside dingo instead, moving the failure from decode to persistence rather than removing it.

Acceptance criteria: MIR amounts carry a sign from the certificate through the models, the SQL column, the aggregate reads and the INSTANT-rule arithmetic, so a negative delta is stored and applied rather than erroring. A test applies a certificate carrying a negative delta and asserts the resulting account balance, and a second asserts the aggregate reads sum signed values correctly.

Blocks adopting any gouroboros release containing #2247.

PR #4415, opened 2026-09-17
fix(ledger): reject negative accumulated MIR updates

### LINKED ISSUE #4361: ledger: reject Alonzo/Babbage MIR updates that make accumulated rewards negative
## Summary

Dingo does not enforce the reference MIRProducesNegativeUpdate check against accumulated InstantaneousRewards state during Alonzo/Babbage transaction validation. The current rule checks each MIR certificate independently, so a running negative aggregate can be accepted.

## Consensus impact

The reference transition processes certificates sequentially: reward account X receives +10 followed by -11, producing -1 and rejecting the transaction. Dingo accepts both. The same divergence occurs within one transaction and across transactions or blocks in one epoch. The epoch-boundary negative-aggregate discard fallback should be unreachable for a fully validated chain.

## Acceptance criteria

- Carry pending InstantaneousRewards state into MIR validation.
- For protocol versions >= 5, process certificates in order and reject existing + delta < 0 with MIRProducesNegativeUpdate.
- Test Alonzo and Babbage with +10 then -11 in one transaction and across transactions or blocks in one epoch; both must reject.
- Keep +10 then -10 valid.
- Keep a negative delta valid when the accumulated result remains non-negative.
- Make the epoch-boundary negative-aggregate discard path unreachable for a fully validated chain.

This is consensus-critical for Alonzo/Babbage validation and historical sync. Related but distinct: #4028 covers MIR aggregation and bookkeeping follow-ups; #3748 covers MIR genesis quorum validation; #3971 covered signed MIR representation through the pipeline.

cryptodj413:
## Summary

Fixes #4361. Dingo checked each Alonzo/Babbage move-instantaneous-rewards
(MIR) certificate's reward delta independently, so a running negative
aggregate for a credential could be accepted even though the reference
`MIRProducesNegativeUpdate` rule rejects it.

- Adds `validateMIRAccumulatedRewards` (`ledger/eras/mir.go`), gated on
  protocol major version >= 5 (`lcommon.MirTransferAllowed`): walks a
  transaction's MIR certificates in order, folding each delta into a running
  per-credential total, and rejects with `MIRProducesNegativeUpdateError`
  when the result goes negative.
- Adds `MIRPendingRewardsProvider`, implemented by `*LedgerView`
  (`ledger/view.go`) via `PendingMIRRewardDeltas`, which sums MIR reward
  deltas already committed this epoch (same block or earlier) from the
  existing `GetMIRCertsInSlotRange` query.
- Wires the check into `ValidateTxAlonzo` and `ValidateTxBabbage`.
- Documents the new check in `ARCHITECTURE.md` alongside the existing
  epoch-boundary MIR section.

Out of scope (tracked separately): #4028 (MIR aggregation/bookkeeping
follow-ups), #3748 (MIR genesis quorum validation). #3971 (signed MIR
representation) was already on `main`.

## Tests

- `ledger/eras/mir_test.go`: +10 then -11 rejects (single tx and cross-tx,
  same epoch), +10 then -10 valid, negative delta valid when the result
  stays non-negative, pre-PV5 no-op, no-MIR-certs no-op, missing-provider
  no-op — each at Alonzo (PV6) and Babbage (PV8).
- `ledger/mir_test.go`: `TestLedgerView_PendingMIRRewardDeltas(_BeforeEpochStart)`
  cover the DB-query window itself (`[epochStartSlot, uptoSlot]` inclusive,
  excluding certs outside it on either side).
- Fail-before proof: reverting the negative-result check causes the reject
  subtests to fail as expected; reverting the query's inclusive upper bound
  causes `TestLedgerView_PendingMIRRewardDeltas` to fail as expected.

## Checks

`go build`/`go vet` (root + nested `internal/test/antithesis`,
`examples/dingo-gov-lens`), `go test ./ledger/...` and `./ledger/eras/...`
with `-race`, `golangci-lint run ./ledger/...`, `make import-boundaries`,
`make docs-parity`, `make golines` — all pass. Two pre-existing failures in
`ledger` (`TestStaleBlockfetchBatchIsDiscardedAfterLocalSiblingAdoption`,
`TestAdoptLocalForgedSiblingAdoptsTheWinnerOfChainSelection`) reproduce
identically on unmodified `origin/main` and are unrelated (chain-selection
rollback/checkpoint logic).

Skipped: `nilaway`/`modernize` (local Go toolchain version mismatch with
repo config, unrelated to this diff), `make sql-check` (no `sqlc`-generated
code touched), live DevNet/conformance/Antithesis (unavailable here).

---
## Summary by cubic
Rejects MIR certificates that would drive a credential's accumulated MIR rewards negative, fixing #4361. Alonzo and Babbage validation previously checked each MIR delta in isolation; now it folds each certificate's delta into the credential's running total for the epoch and fails with `MIRProducesNegativeUpdateError` when the total goes below zero.

- The running total starts from MIR deltas already committed earlier in the epoch via `LedgerView.PendingMIRRewardDeltas`, then applies the transaction's certificates in order.
- The epoch start slot is pinned to the transaction validation's consensus snapshot, so a concurrent epoch rollover can't make the check silently skip pending deltas.
- Reserves and treasury deltas accumulate separately, so a surplus in one pot cannot offset a deficit in the other.
- The check runs only from protocol major version 5; earlier eras already reject any negative delta through `gouroboros`'s `MIRNegativesNotCurrentlyAllowedError`.
- Ledger states without `MIRPendingRewardsProvider` no-op instead of failing; tests cover within-transaction, cross-transaction, cross-pot, and rollover-race cases at Alonzo and Babbage.

<sup>Written for commit 9f62a5709886835e3efe8948d9b04415ed79e183. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4415?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation for instantaneous reward (MIR) certificates in Alonzo and Babbage transactions.
  - Transactions are rejected when cumulative MIR updates would result in negative rewards for a credential.
  - Validation now accounts for pending MIR rewards and the order of certificates within the current epoch.

- **Documentation**
  - Added documentation describing MIR validation behavior and its differences across protocol eras.


ledger/eras/mir.go
```diff
@@ -0,0 +1,151 @@
...
+
+// MIRCredentialKey identifies one stake credential for the purpose of folding
+// move instantaneous rewards deltas accumulated within an epoch. Both the tag
+// and the hash are required: a key hash and a script hash may coincidentally
+// share 28 bytes, and the reference keeps them in separate maps
+// (iRReserves/iRTreasury are keyed by Credential, which carries its type).
+type MIRCredentialKey struct {
+	Tag        uint8
+	Credential lcommon.Blake2b224
+}
```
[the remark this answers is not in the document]
cryptodj413: Fixed in 163fafda: added Pot to MIRCredentialKey, populated from cert.Reward.Source in validateMIRAccumulatedRewards and effect.Pot in PendingMIRRewardDeltas. Added TestValidateMIRAccumulatedRewards_CrossPotIsolation and TestLedgerView_PendingMIRRewardDeltas_SeparatesPots proving a reserves surplus cannot offset a treasury deficit; both fail without the fix (verified by reverting the Pot assignment).

ledger/view.go
```diff
@@ -198,6 +198,56 @@ func (lv *LedgerView) MinPoolMargin() *big.Rat {
...
+// that transaction validates (see ledgerProcessBlock), before the next
+// transaction in the same block validates, so this query already reflects
+// every prior transaction's MIR certificates this epoch, same block or
+// earlier -- only certificates within the transaction currently being
+// validated are not yet visible here, and validateMIRAccumulatedRewards folds
+// those in memory as it walks the transaction's own certificate list.
+func (lv *LedgerView) PendingMIRRewardDeltas(
+	uptoSlot uint64,
+) (map[eras.MIRCredentialKey]*big.Int, error) {
+	lv.ls.RLock()
```
arepala-uml: PendingMIRRewardDeltas reads ls.currentEpoch via ls.RLock() instead of the pinned consensus snapshot used elsewhere in tx validation (txValidationSnapshot/pinCommitteeState); ledger/state.go documents currentEpoch as writer-owned state that lock-free readers must read from a snapshot instead. A concurrent epoch rollover can make epochStartSlot newer than the validating transaction's actual epoch, tripping the uptoSlot < epochStartSlot guard and returning zero pending deltas, so a MIR update that should fail MIRProducesNegativeUpdate can validate instead.

arepala-uml (changes requested):

ledger/view.go
cryptodj413: Confirmed and fixed in 06b78031: added currentEpochStartSlot to txValidationSnapshot and a pinMIREpochStartSlot/epochStartSlotSnapshot pair on LedgerView, mirroring the existing committee-state pinning, and pinned it at all three ValidateTxFunc call sites (validateTxCore, WithTxValidationSession, ledgerProcessBlock). Added TestLedgerView_PendingMIRRewardDeltas_PinnedSurvivesConcurrentRollover, which reproduces the exact failure mode you described (pinned view keeps epoch A's deltas after a concurrent rollover to epoch B; an unpinned view silently returns none) and fails without the fix.

arepala-uml (approved):

PR #4417, opened 2026-09-17
fix(ledger): preserve VRF key reservation until epoch boundary

### LINKED ISSUE #4352: ledger: preserve current VRF reservation during deferred pool re-registration
## Problem

Pool VRF uniqueness currently checks the denormalized `pool.vrf_key_hash` row through `LedgerView.IsVrfKeyInUse`. Processing a pool re-registration upserts that row with the new VRF key (`database/plugin/metadata/sqlstore/transaction_certificates.go`, `processPoolRegistration`), even though the new registration is future state until the next epoch boundary.

Reproduction:

1. Pool P is active with VRF key A.
2. P re-registers with VRF key B during the current epoch.
3. Before the boundary, pool Q registers with VRF key A.

Cardano ledger keeps P's current key A in `psStakePools` and reserves B in future pool parameters until the boundary, so Q's registration using A is rejected. Dingo's upsert changes P's denormalized row to B, so `GetPoolByVrfKeyHash(A)` returns no active owner and Q can be accepted.

Current code: `database/plugin/metadata/sqlstore/transaction_certificates.go`, `processPoolRegistration`; `database/plugin/metadata/sqlstore/pool.go`, `GetPoolByVrfKeyHash`; `ledger/view.go`, `IsVrfKeyInUse`.

## Acceptance criteria

- Preserve the active/current VRF reservation until the epoch boundary while retaining the future registration for later activation.
- Reject a different pool registering the active key during the deferral window.
- Allow the new key only at the reference-defined activation point, and continue allowing the old key after it is no longer reserved.
- Add a regression test for current-key/future-key overlap, plus controls for same-pool re-registration and post-boundary activation.
- Keep historical/as-of-slot VRF lookup behavior used by header validation and stake-distribution queries.

arepala-uml:
Closes #4352.

## Summary

`GetPoolByVrfKeyHash` read the denormalized `pool.vrf_key_hash` column, which the certificate-application path overwrites immediately on every re-registration regardless of epoch boundary. cardano-ledger defers a re-registration through `psFutureStakePoolParams` until the next epoch boundary, so a different pool could register a still-reserved VRF key mid-epoch. Separately, a pool cycling a key through more than one same-epoch re-registration (A -> B -> C) could reuse the superseded value B, since `psVRFKeyHashes` retains every key placed in `psFutureStakePoolParams` during the epoch, not only the current one.

`GetPoolByVrfKeyHash` now resolves ownership from `pool_registration` history as of the current epoch's start slot, in two tiers: the pool whose registration is effective before the boundary, falling back to any pool with a same-epoch registration naming the key.

## Changes

- `database/plugin/metadata/sqlstore/pool.go` — `GetPoolByVrfKeyHash` takes an `epochStartSlot` parameter and resolves the two-tier ownership check described above.
- `database/plugin/metadata/store.go` — `MetadataStore` interface signature and doc comment updated to match.
- `database/pool.go` — `Database.GetPoolByVrfKeyHash` wrapper threads the new parameter through.
- `ledger/view.go` — `LedgerView.IsVrfKeyInUse` derives the current epoch's start slot from the consensus snapshot and passes it to `GetPoolByVrfKeyHash`.
- `database/plugin/metadata/sqlite/shared_sqlstore_pool_parity_test.go` — cross-backend parity test interface and call site updated to the new signature.
- `ledger/dijkstra_pool_margin_floor_e2e_test.go` — adds a `publishEmptyEpochSnapshot` helper; `IsVrfKeyInUse` now reads the consensus snapshot, which these pre-existing tests' hand-built `LedgerState` never published.
- `DATABASE.md` — documents `GetPoolByVrfKeyHash`'s two-tier resolution and why retirement precedence is left to the existing `activePoolOrNil` check rather than reimplemented here.

## Tests

`database/plugin/metadata/sqlstore/management_test.go`:
- `TestGetPoolByVrfKeyHashPreservesActiveKeyDuringDeferredReRegistration` — a pool's pre-boundary key stays reserved through a mid-epoch re-registration.
- `TestGetPoolByVrfKeyHashAllowsSamePoolRevertingToItsOwnKey` — control: a pool reverting to its own active key is not blocked.
- `TestGetPoolByVrfKeyHashActivatesAndReleasesAtEpochBoundary` — the new key activates and the old key releases once the epoch boundary passes, and a different pool can then claim the released key.
- `TestGetPoolByVrfKeyHashClaimsSupersededSameEpochFutureKey` — a pool cycling A -> B -> C within one epoch still claims B, the superseded value.

`ledger/view_test.go`:
- `TestLedgerViewIsVrfKeyInUseRespectsEpochBoundaryDeferral` — `IsVrfKeyInUse` through the real certificate-application pipeline.
- `TestLedgerViewIsVrfKeyInUseRejectsSameOperatorReuseOfSupersededFutureKey` — `IsVrfKeyInUse` and `PoolCurrentState` disagree in the way the caller's rejection depends on.
- `TestValidateTxDijkstraRejectsDifferentPoolClaimingActiveKeyDuringDeferral` — a different pool's conflicting registration is rejected by `eras.ValidateTxDijkstra`, not just reported in use.
- `TestValidateTxDijkstraRejectsSameOperatorReuseOfSupersededFutureKey` — the same-operator A -> B -> C -> B reuse is rejected by `eras.ValidateTxDijkstra`.

`GetPoolVrfKeyHashAtSlot`, `GetPoolEarliestVrfKeyHashAtSlot`, `GetActivePoolKeyHashesAtSlot`, and `GetPoolRegistrationsEffectiveForEpoch` are unchanged; their existing coverage for header validation and stake-distribution queries passes unchanged.


arepala-uml:
Both merge-risk findings are confirmed and fixed as of 73771cb.

**Retired-candidate finding**: `GetPoolByVrfKeyHash` picked a single winning `pool_id` via `ORDER BY ... LIMIT 1` and checked retirement only on that one. A long-retired pool's own last-ever registration can still satisfy the effective-key computation for a queried key (registration history and retirement are tracked independently), so a retired candidate and a different, genuinely active pool that later re-registered the same, by-then-free key could both appear as candidates for one lookup. If the retired one sorted first, its retirement resolved the whole lookup to `nil` even though the key was genuinely in use — reintroducing the class of bug this PR fixes. Reproduced with `TestGetPoolByVrfKeyHashSkipsRetiredCandidateForActiveOwner` before the fix, passing after. Fixed in `43c3af4`: every ranked candidate (tier 1 before tier 2, `pool_id` ascending within a tier) is now tried through the existing `activePoolOrNil` in turn, returning the first still-active one.

**Mid-validation epoch re-read finding**: `IsVrfKeyInUse` read `lv.ls.loadConsensusSnapshot().currentEpoch.StartSlot` live on every call, unlike every other snapshot-derived value on `LedgerView` (`committeeEpoch`, `pp`, `syntheticV2CostModel`), which are pinned once at construction for exactly this reason — a long-running validation can span a writer publishing a newer epoch boundary. Reproduced with `TestLedgerViewIsVrfKeyInUseIgnoresConcurrentSnapshotRepublish` (fails against the live-read version, passes against the fix). Fixed in `94a6ba3`: `LedgerView.epochStartSlot` is now pinned at every real construction site (`NewView`, `ledgerProcessBlock`, `validateTxCore`, `ValidateTxWithOverlay`, `EvaluateTx`) from the same snapshot that pins the other fields, and `IsVrfKeyInUse` reads the pinned field.

Full `database`/`ledger` suites pass under `-race` after both fixes.

database/plugin/metadata/sqlstore/pool.go
```diff
@@ -509,25 +509,171 @@ func (s *Store) GetPool(
...
+		if err != nil {
+			return nil, err
+		}
+		if pool == nil {
+			continue
+		}
+		if err := s.loadPoolAssociations(ctx, db, pool, true); err != nil {
+			return nil, err
+		}
+		active, err := s.activePoolOrNil(ctx, db, pool)
```
cryptodj413: This activePoolOrNil call checks retirement against currentEpoch(ctx, db) and the pool's latest event, both present-tense, not against epochStartSlot. A pool that retires then later re-registers with a different key reads as active today, so GetPoolByVrfKeyHash(oldKey, epochStartSlot) still returns it here, reporting oldKey in use when it should be free, reintroducing this PR's own bug class: gouroboros validatePoolRegistration and conway PoolValidateVrfKeyUniqueness then falsely reject a different pool's valid registration of oldKey.

ledger/view_test.go
```diff
@@ -474,6 +475,508 @@ func TestLedgerViewPoolCurrentStatePendingRetirement(t *testing.T) {
...
+	applyRegistration(10, 0x01, oldVrfKeyHash)
+	// P re-registers with a new key mid-epoch (slot 50), inside the epoch
+	// that starts at slot 30 below. Not yet effective.
+	applyRegistration(50, 0x02, newVrfKeyHash)
+
+	// Pinned directly on the view, mirroring how real validation call
+	// sites (NewView, ledgerProcessBlock, validateTxCore) pin
+	// epochStartSlot at construction time rather than leaving
+	// IsVrfKeyInUse to re-read ls.loadConsensusSnapshot() live.
+	lv.epochStartSlot = 30
```
cryptodj413, line 568: All new epochStartSlot-pinning tests set lv.epochStartSlot by hand on a bare &LedgerView{ls: ls} (also lines 667, 788, 904, 965); none constructs a LedgerView through NewView, validateTxCore, or EvaluateTx and asserts the pin happened there, so a real construction site could drop its epochStartSlot: line without failing anything.

ledger/dijkstra_pool_margin_floor_e2e_test.go
```diff
@@ -18,13 +18,28 @@ import (
...
 )
 
+// publishEmptyEpochSnapshot gives a bare, hand-constructed LedgerState (as
+// newRewardCalculationTestLedger builds) a non-nil published consensus
+// snapshot. LedgerView.IsVrfKeyInUse reads the current epoch's start slot
+// from that snapshot (issue #4352); a LedgerState that never published one
+// has a nil *consensusSnapshot and panics on the dereference. These tests
+// don't exercise any VRF-key-deferral scenario, so an epoch starting at
+// slot 0 is sufficient -- only avoiding the nil dereference matters here.
+func publishEmptyEpochSnapshot(ls *LedgerState) {
```
cryptodj413: publishEmptyEpochSnapshot and its comment claiming a nil *consensusSnapshot panics on the dereference are no longer accurate: since 94a6ba3e pinned epochStartSlot at construction instead of reading loadConsensusSnapshot() live in IsVrfKeyInUse, removing this helper and its four call sites does not reintroduce the panic.

ledger/utxo_memo_test.go
```diff
@@ -139,6 +139,7 @@ func TestUtxoByIdMemoReducesDbReads_LedgerProcessBlock(t *testing.T) {
 			fx.dingoLS.currentPParams,
 			nil, // prevEraPParams
 			0,   // committeeEpoch
+			0, // epochStartSlot
```
cryptodj413: gofmt -l flags this file for comment alignment on this line; base only flags ledger/read_chain_pipeline_test.go, so this failure is introduced by this PR.

cryptodj413 (changes requested):
Both CodeRabbit findings are genuinely fixed and causally tested; found one new blocker of the same bug class (retired-then-re-registered pool still claims its old key) plus a construction-site test gap and a PR-introduced gofmt failure.

database/plugin/metadata/sqlstore/pool.go
arepala-uml: Confirmed and fixed in 127ddc86. Reproduced first with `TestGetPoolByVrfKeyHashFreesKeyAfterRetirementThenDifferentKeyReRegistration`: P registers key A, retires (effective epoch 1), then registers a different key B at slot 410 — un-retiring via a fresh registration, which cardano-ledger treats as immediate (not deferred) since the pool had left `psStakePools`. Retirement is now resolved against `epochStartSlot`'s own epoch (a new `pool_active` CTE, mirroring `GetActivePoolKeyHashesAtSlot`'s precedence rule) instead of the live tip, applied to both the effective-owner and same-epoch-claimant tiers — restricting it to the first tier alone still let a retired pool's same-epoch registration leak through the second (caught when `TestGetPoolByVrfKeyHashExcludesRetiredPool` regressed against the first draft of this fix).

ledger/view_test.go
arepala-uml: Confirmed and fixed in 3f06b345. Added three tests that call the real construction sites directly: `TestLedgerStateNewViewPinsEpochStartSlot` (NewView, checking the returned view's field), and `TestLedgerStateValidateTxPinsEpochStartSlot` / `TestLedgerStateEvaluateTxPinsEpochStartSlot` (validateTxCore/EvaluateTx, through the public `ls.ValidateTx`/`ls.EvaluateTx` entry points, with a custom era descriptor whose ValidateTxFunc/EvaluateTxFunc captures the `*LedgerView` it is actually given). Verified each is discriminating by temporarily removing the corresponding `epochStartSlot:` assignment at each site and confirming the matching test fails.

ledger/dijkstra_pool_margin_floor_e2e_test.go
arepala-uml: Confirmed and fixed in 654cc61f: removed `publishEmptyEpochSnapshot` and its four call sites entirely. Verified all four margin-floor tests still pass with a bare, never-published `&LedgerView{ls: ls}` (`epochStartSlot` simply defaults to zero, which is safe since these tests never exercise a VRF-key cert).

ledger/utxo_memo_test.go
arepala-uml: Confirmed and fixed in 5c9cada9 (`gofmt -w`). `gofmt -l .` is now clean across the whole repository.

database/plugin/metadata/sqlstore/pool.go
```diff
@@ -509,24 +509,134 @@ func (s *Store) GetPool(
...
+    FROM pool_registration pr
+    JOIN candidates cd ON cd.pool_id = pr.pool_id
+    WHERE pr.vrf_key_hash = ? AND pr.added_slot >= ?
+)
+SELECT pool_id FROM active_owner
+UNION ALL
+SELECT pool_id FROM same_epoch_claimant
+WHERE NOT EXISTS (SELECT 1 FROM active_owner)
+ORDER BY pool_id
+LIMIT 1`,
```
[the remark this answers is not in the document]
arepala-uml: Already addressed, in two stages. This review ran against 463bdb6, before either fix landed. The LIMIT-1 winner-selection issue was fixed in 43c3af4 (try every ranked candidate, not just one). The deeper retirement-precedence gap this finding also describes — a retired pool's stale registration still counting as active — needed a further fix in 127ddc86 after human review caught the remaining case (retire, then re-register with a different key): retirement is now resolved against epochStartSlot's own epoch via a pool_active CTE, not the live tip, applied to both candidate tiers. Regression coverage: TestGetPoolByVrfKeyHashSkipsRetiredCandidateForActiveOwner and TestGetPoolByVrfKeyHashFreesKeyAfterRetirementThenDifferentKeyReRegistration.

ledger/view.go
```diff
@@ -614,11 +614,19 @@ func (lv *LedgerView) IsPoolRegistered(pkh lcommon.PoolKeyHash) bool {
...
+//
+// The reservation checked here is the one effective as of the start of the
+// current epoch, not a pool's most recent registration: a re-registration
+// submitted during the epoch in progress is deferred to the next boundary
+// (cardano-ledger's psFutureStakePoolParams), so the pool's prior key stays
+// reserved until then. See GetPoolByVrfKeyHash.
 func (lv *LedgerView) IsVrfKeyInUse(
 	vrfKeyHash lcommon.Blake2b256,
 ) (bool, lcommon.PoolKeyHash, error) {
+	epochStartSlot := lv.ls.loadConsensusSnapshot().currentEpoch.StartSlot
```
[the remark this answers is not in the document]
arepala-uml: Already addressed in 94a6ba3e, before this review ran (it targeted 463bdb6). LedgerView now carries epochStartSlot, pinned once at every real construction site (NewView, ledgerProcessBlock, validateTxCore, ValidateTxWithOverlay, EvaluateTx) from the same snapshot that pins committeeEpoch/pp, and IsVrfKeyInUse reads the pinned field. Regression coverage: TestLedgerViewIsVrfKeyInUseIgnoresConcurrentSnapshotRepublish, plus TestLedgerStateNewViewPinsEpochStartSlot / TestLedgerStateValidateTxPinsEpochStartSlot / TestLedgerStateEvaluateTxPinsEpochStartSlot (added in 3f06b345) verifying the pin through each real construction site rather than a hand-set field.

arepala-uml:
All four review findings on 73771cb are confirmed and fixed (replies posted on each thread with commit references and evidence):

- `fix(database)` 127ddc86: VRF-key retirement now resolves against `epochStartSlot`'s own epoch via a `pool_active` CTE (mirroring `GetActivePoolKeyHashesAtSlot`'s precedence rule), applied to both candidate tiers, instead of the live database tip.
- `test(ledger)` 3f06b345: epoch-start-slot pinning is now verified through the real `NewView`/`ValidateTx`/`EvaluateTx` entry points, not only a hand-set field.
- `fix(ledger)` 654cc61f: removed the now-stale `publishEmptyEpochSnapshot` helper.
- `style(ledger)` 5c9cada9: gofmt comment alignment.
- `docs` cb4cdf31: `DATABASE.md` updated for the retirement fix and the new pinning verification.

Also replied to CodeRabbit's two inline findings from the same review pass, both already addressed by the earlier commits on this branch.

Full `database`/`ledger` suites pass, `gofmt -l .` is clean, `golangci-lint`/`nilaway` report no issues on the touched packages.

cryptodj413 (approved):

PR #4422, opened 2026-09-17
fix(ledger): close silent-corruption gap in reward stake denominator check (#4025)

### LINKED ISSUE #4025: ledger: a proportionally reduced reward stake set passes the relaxed denominator bound silently
Non-blocking review finding from #3969, which relaxed the exact stake-sum comparison to a `<=` bound at `ledger/reward_calculation.go:3078`, its counterpart at `:1412`, and `rewards.go:943`.

The bound is weaker than the exact match it replaced in one specific way: a `reward_stake_input`/`reward_pool_input` row set that has been *proportionally* reduced — not merely missing an excluded pool — still satisfies `<=` and passes every other check on the path, because pool count, delegator count and the per-pool cross-sums all remain internally consistent. The epoch then pays out slightly low with no signal.

The data needed to close this exists but is not persisted. At capture time (`ledger/snapshot/rotation.go:437`) the exact shortfall is known:

```
sumPoolStakes(rewardDistribution.PoolStakes) - sumPoolStakes(effective.PoolStakes)
```

The excluded pool contributes no row, so the validation path cannot reconstruct it afterwards. The exclusion is WARN-logged once at capture, but the reward calculation can run much later — after a restart or a replay — and says nothing at that point.

Two options, in increasing cost:

1. Log or emit a metric at the calculation path when `totalPoolStake < snapshot.TotalActiveStake`.
2. Persist the excluded stake on `reward_snapshot` and assert `totalPoolStake + excludedStake == TotalActiveStake` exactly, restoring the replaced check's full strength.

arepala-uml:
## Summary

Closes #4025 — a non-blocking review finding from #3969.

#3969 relaxed reward calculation's stake-sum check from an exact match to a
`<=` bound, to tolerate one legitimate case: a pool with unresolvable
registration data is excluded from `reward_pool_input`, so the row set sums
to less than `reward_snapshot.total_active_stake` by design.

The `<=` bound is too permissive: it also silently accepts a
`reward_pool_input`/`reward_stake_input` row set that has been
*proportionally* reduced by an unrelated bug, not just missing one excluded
pool. Pool count, delegator count, and the per-pool cross-sums all stay
internally consistent, so nothing else on the path catches it. The epoch
then pays out slightly low with no signal.

The exact excluded-stake amount was already computable at capture time
(`sumPoolStakes(rewardDistribution.PoolStakes) - sumPoolStakes(effective.PoolStakes)`
in `ledger/snapshot/rotation.go`) but was only WARN-logged once, not
persisted — so validation, which can run much later after a restart or
replay, had no way to reconstruct it.

This implements the issue's second (stronger) option: persist the excluded
stake and assert `sum(reward_pool_input) + excluded == total_active_stake`
exactly, restoring the original check's full strength for any row where the
exclusion is tracked. A row from before this change (`excluded_active_stake`
is `NULL`) keeps the old `<=` bound, since its exclusion amount, if any, is
unrecoverable.

## Changes

1. **`feat(database): add reward_snapshot.excluded_active_stake migration`**
   - `database/plugin/metadata/sqlstore/migrations/v17/sqlite/expand.sql` (new) — nullable `excluded_active_stake` column
   - `database/plugin/metadata/sqlstore/migrations/registry.go` — registers migration v17
   - `database/plugin/metadata/sqlstore/migrations/{migration_released_checksum_test,registry_test,committee_credentials_test,pool_deposit_held_test,reward_stake_version_restamp_test,runner_test}.go` — updated hardcoded registry length, ADD COLUMN count, and pinned per-dialect checksums

2. **`feat(database): persist and read back reward_snapshot excluded active stake`**
   - `database/models/reward_state.go` — `RewardSnapshot.ExcludedActiveStake *types.Uint64`
   - `database/plugin/metadata/sqlstore/queries/sqlite/{schema.sql,operational.sql}` — new column, wired into Save/Insert/UpdateFallback/Get queries
   - `database/plugin/metadata/sqlstore/internal/query/sqlite/{models.go,operational.sql.go}` — regenerated via `make sql`
   - `database/plugin/metadata/sqlstore/reward_state.go` — read/write glue for the new nullable field

3. **`feat(rewards): track excluded active stake in Snapshot validation`**
   - `ledger/rewards/rewards.go` — `Snapshot.ExcludedActiveStake *uint64`; `validateSnapshot` does an exact check when tracked (including a tracked zero), falls back to the old bound when nil

4. **`fix(ledger): require reward stake inputs to match snapshot active stake exactly`**
   - `ledger/snapshot/rotation.go` — `buildRewardStateInputs` now computes and persists the exact excluded amount at capture time
   - `ledger/reward_calculation.go` — new `rewardStakeSumMatchesSnapshot` helper; both `validateRewardCalculatorInputs` and `precomputedRewardPoolInputsMatchSnapshot` use it instead of a bare `<=`
   - `ledgerstate/imported_reward_inputs.go` — bootstrap import path marks its rows with a tracked, exact zero exclusion (it never excludes a pool)

5. **`test(ledger): cover exact excluded active stake validation and DB round-trip`**
   - `ledger/reward_calculation_test.go`, `ledger/rewards/active_stake_denominator_test.go`, `ledger/snapshot/reward_active_stake_test.go` — see Test coverage below

6. **`docs(database): document reward_snapshot.excluded_active_stake`**
   - `DATABASE.md` — documents the new column and the exact-vs-non-exceeding validation behavior

## Test coverage

- `TestRewardCalculatorInputsExactWithTrackedExcludedStake` — a `reward_pool_input` set proportionally shrunk (halved) is rejected once the snapshot tracks the exclusion, even though it would satisfy the old `<=` bound; also covers the legitimate-exclusion pass case and the tracked-zero-exclusion case.
- `TestValidateSnapshotTrackedExcludedActiveStake` — same proportional-shrink scenario at the `ledger/rewards.validateSnapshot` layer.
- `TestRewardSnapshotActiveStakeKeepsDegradedPoolStake` (extended) and `TestRewardSnapshotActiveStakeTracksNoExclusion` (new) — `buildRewardStateInputs` persists the exact excluded amount, including a tracked zero when nothing was excluded.
- `TestSaveSnapshotKeepsDegradedPoolStakeInPoolAndEpochRows` (extended) — `excluded_active_stake` round-trips through an actual database save/read cycle (`SaveRewardSnapshot` → `GetRewardSnapshot`), covering the restart/replay scenario the issue describes.

Every new/extended assertion was mutation-tested: the corresponding production code change was temporarily reverted and each test was confirmed to fail before restoring the fix.

### Test plan

- [x] `go build ./...`
- [x] `go test -race ./database/... ./ledger/... ./ledgerstate/...` (all pass)
- [x] `make docs-parity`
- [x] `make sql-check`
- [x] Each of the 6 commits individually verified to build and pass its package's tests in an isolated worktree (bisectable history)
- [x] `make lint` — 3 pre-existing `nilaway` findings in `chainselection/selector.go` and `internal/node/load.go`, confirmed unrelated by reproducing them against unmodified `main`; no findings in changed files

Not run: conformance/DevNet suites and `make config-parity`/`make govulncheck` (require live network/Docker infrastructure not available here) — this change touches only metadata storage and reward calculation, not consensus, block production, or wire protocols.

---
## Summary by cubic
Closes #4025, a silent-corruption gap where a proportionally shrunk reward input set satisfied the old `<=` stake-sum check and paid out slightly low with no signal. Reward validation now requires `sum(reward_pool_input) + excluded_active_stake == total_active_stake` exactly when the exclusion is tracked; rows with a `NULL` exclusion (predating this change) keep the old bound.

**Details**
- Adds the nullable `reward_snapshot.excluded_active_stake` column via migration v18 and wires it through the SQL queries, models, and save/read paths.
- `snapshot.buildRewardStateInputs` computes and persists the exact excluded stake at capture time; bootstrap-imported rows track an exact zero since they never exclude a pool.
- `ledger/rewards` `Snapshot` validation enforces the same exact sum requirement when `ExcludedActiveStake` is set.
- Renumbered the migration from v17 to v18 because main shipped its own v17 (`governance_proposal_drop`) in parallel; checksums and registry assertions were refreshed after the merge.

<sup>Written for commit f997a3c4320c7721e6ef2a726c80e567fd1b7c3a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4422?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


arepala-uml:
Addressed review finding: updated the stale doc comment on `buildRewardStateInputs` (`ledger/snapshot/rotation.go`) that still described the pre-fix non-exceeding-only bound, contradicting the `DATABASE.md` update from the same PR. Fixed in c3000c98.

wolf31o2 (dismissed):

wolf31o2 (approved):

PR #4432, opened 2026-09-18
perf(ledger): cache HardForkSummary per publication generation

### LINKED ISSUE #2093: ledger: cache dynamic HFC Summary instead of rebuilding per-query
## Background

Upstream `ouroboros-consensus`'s `History/Caching.hs` is a 60-line wrapper that caches the `Summary` (the slot↔time↔epoch authority) under a `TVar` and invalidates on rollover:

```haskell
data RunWithCachedSummary xs m = RunWithCachedSummary
  { cachedRunQuery :: forall a. Qry a -> STM m (Either PastHorizonException a) }

runWithCachedSummary :: STM m (Summary xs) -> m (RunWithCachedSummary xs m)
```

Dingo's `cachedShape` (`ledger/state.go`) caches the static `Shape` (which never changes after startup), but nothing caches the dynamic `Summary` produced by `BuildSummary`. `queryHardForkEraHistory` rebuilds the `Summary` from DB queries on every call.

## Why this is currently deferred

Not a correctness issue — `BuildSummary` is deterministic over the same inputs, so a cached `Summary` and a freshly-built one match. Just per-call work that adds up if `queryHardForkEraHistory` is hot.

## What's missing

A `TVar (Summary xs)`-equivalent in Go: an atomic value caching the latest `Summary`, populated on first build and invalidated whenever any of its inputs change (epoch rollover, era transition, rollback, transitionInfo state change). Pretty much a drop-in template once profiling shows it's wanted.

## When this becomes a forcing function

- Profiling shows `queryHardForkEraHistory` is hot (high QPS, indexer/db-sync workload).
- Block-application latency under heavy local-state-query traffic shows contention with the per-call rebuild.

Until then this is a perf optimization with no observed need. Filing for visibility; not actively worked on.

## References

- `ouroboros-consensus/src/.../HardFork/History/Caching.hs`
- `ledger/queries.go queryHardForkEraHistory`
- `ledger/hardfork_summary.go HardForkSummary`
- `plans/HFC_GAP_ANALYSIS_5.md` (internal) — context

chrisguiney:
`hardForkSummaryAnchoredAt` (`ledger/hardfork_summary.go`) rebuilt its era summary by walking the whole in-memory epoch cache on every call — one row per epoch since genesis, not one per era — so the per-call cost grows without bound as the chain ages. It is called from per-block header verification, the operational slot clock (every slot tick), mempool admission, and per-transaction validity-interval translation during block application.

Measured against synthetic epoch caches matching current mainnet and Preview epoch counts (700 and 1500 rows): 16.5µs/58KB and 26.0µs/115KB per call before, against 10.6ns and 12.9ns with zero allocations after. A 10-epoch cache — the `O(eras) ≈ 7` size assumed when #2093 was first closed — already cost 518ns/961B, so the per-call cost was never actually O(eras).

Closes #2093.

## Change

The result is cached on `LedgerState`, keyed on `(consensusSnapshot.generation, horizonAnchorSlot)`, rather than adding a separate invalidation hook at each of epoch rollover, era transition, rollback, and `TransitionInfo` change. All four already mutate writer-owned state under `ls.Lock()` and call `publishSnapshotsLocked` before `Unlock`, which is the sole point that bumps `generation` and publishes a new immutable `consensusSnapshot`/`tipSnapshot` pair. A generation-keyed entry is therefore invalidated by construction, reusing the invariant every other lock-free reader of those snapshots already depends on instead of introducing a second one.

`horizonAnchorSlot` is the other half of the key because block application and mempool/query callers request different horizons against the same published generation. The cache holds a single entry, so a caller alternating anchors costs one extra rebuild rather than unbounded memory, and the key check compares against the generation just loaded rather than against whatever the cache holds, so a racing overwrite can never serve a wrong answer.

The cache amortizes callers that share one published generation. Per-block callers still pay one rebuild per publication; `BenchmarkHardForkSummary_RealisticCache_1500_PerBlock` measures that case directly rather than leaving it implied.

## Tests

One test per invalidation trigger, each asserting the post-trigger summary differs from the pre-trigger one in the way that trigger implies and matches an uncached rebuild, plus a concurrent-access test under `-race`. A further test asserts the cache is actually served: the invalidation tests only prove a stale summary is never returned, and all of them pass when the cache lookup is removed, so without it a key that stopped matching would stay green while restoring the original per-call walk.

## Not covered

`queryHardForkEraHistory` (`ledger/queries.go`) builds its era history from `GetEpochsByEra` reads and calls `hardfork.BuildSummary` directly, so the NtC era-history query path does not go through this cache and is unchanged here.

---
## Summary by cubic
Caches `hardForkSummaryAnchoredAt` results per publication generation so repeated calls within one publication share a build instead of walking the entire epoch cache every time. Per-call cost drops from tens of microseconds and ~100KB of allocations to single-digit nanoseconds with zero allocations, and this closes #2093. The returned `hardfork.Summary` is now a shared cached pointer, so callers must treat it as read-only.

- The cache key is `(consensusSnapshot.generation, horizonAnchorSlot)`, so epoch rollover, era transition, rollback, and `TransitionInfo` changes invalidate it by construction.
- `horizonAnchorSlot` is in the key because mempool/query callers and per-block validation use different horizons at the same generation.
- Synthetic 700- and 1500-epoch caches (matching mainnet/Preview today) measured 16.5µs/58KB and 26.0µs/115KB per call before, 10.6ns and 12.9ns after.
- Per-block callers still pay one rebuild per published block; the cache pays off for callers that share a generation.
- Tests cover each invalidation trigger, concurrent access under `-race`, and that the cache is actually served.
- `queryHardForkEraHistory` bypasses this cache and is unchanged.

<sup>Written for commit f5833eb8719669cc6aec215715b217a305ba0f32. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4432?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Performance**
  - Improved responsiveness when retrieving hard-fork summaries by reusing previously computed results when the underlying ledger state has not changed.
  - Summaries are automatically refreshed after relevant ledger updates, including rollbacks, epoch changes, era transitions, or transition-information changes.

- **Reliability**
  - Added safeguards and validation to ensure cached summaries remain accurate during concurrent access and state updates.

- **Documentation**
  - Documented the hard-fork summary caching and refresh behavior.


ledger/hardfork_summary.go
```diff
@@ -102,6 +146,18 @@ func hardForkCachedEpochDuration(
 //
 // The forecast horizon is measured from the published tip. Callers that know a
 // more recent applied block must use hardForkSummaryAnchoredAt instead.
+//
+// The result is cached per hardForkSummaryCacheKey (see its doc comment); the
+// underlying epoch-cache walk is O(known epochs), which otherwise grows
+// without bound as the chain ages (issue #2093).
```
wolf31o2, line 152: The description carries Closes #2093 and GitHub registers it in closingIssuesReferences, so merging closes that issue. #2093 is titled cache dynamic HFC Summary instead of rebuilding per-query and names queryHardForkEraHistory in its references and in its first forcing function; this change leaves that path on an uncached hardfork.BuildSummary, as the description itself states. Reference #2093 without a closing keyword, or extend the cache to the NtC era-history query, so the uncovered half is not closed silently.

ledger/hardfork_summary.go
```diff
@@ -121,8 +177,43 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
...
+// read-only contract every caller of either method is bound by.
 func (ls *LedgerState) hardForkSummaryAnchoredAt(
 	horizonAnchorSlot uint64,
+) (*hardfork.Summary, error) {
+	consensusState, tipState := ls.loadStateSnapshots()
+	key := hardForkSummaryCacheKey{
+		generation:        consensusState.generation,
+		horizonAnchorSlot: horizonAnchorSlot,
+	}
+	if cached := ls.hardForkSummaryCache.Load(); cached != nil &&
```
wolf31o2, line 191: Key coverage measured both ways: deleting this lookup fails exactly TestHardForkSummaryCache_ServesCachedResultWithinGeneration, and pinning horizonAnchorSlot to 0 in the key fails TestHardForkSummary_HorizonAnchoredAtAppliedParent and TestLedgerViewSlotToTimeUsesHorizonAnchor, so neither key component is unguarded. The CodeRabbit request for a deep copy per call is rejected: every in-repo consumer of the returned Summary reads it, and a per-call copy restores the allocation this change removes.

wolf31o2 (changes requested):
Cache key verified against the code rather than against its comment: every writer that changes epochCache, transitionInfo or currentTip publishes under ls.Lock before unlocking, loadStateSnapshots retries until both generations match, and the only in-place epoch-row write (healEmptyLabNoncesInPlace) touches nonce fields the summary never reads. Merged against current main the ledger package builds, vets, tests, races and lints clean. One scope change requested inline; the CodeRabbit deep-copy suggestion is dispositioned inline.

ledger/hardfork_summary.go
```diff
@@ -123,6 +171,38 @@ func (ls *LedgerState) HardForkSummary() (*hardfork.Summary, error) {
...
+		consensusState,
+		tipState,
+		horizonAnchorSlot,
+	)
+	ls.hardForkSummaryCache.Store(&hardForkSummaryCacheEntry{
+		key:     key,
+		summary: summary,
+		err:     err,
+	})
+	return summary, err
```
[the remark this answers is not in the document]
chrisguiney: Not applied. Every in-repo consumer of the returned `Summary` reads it: no call site assigns to a field, appends to `Eras`, or writes through `EraSummary.End`. A per-call deep copy would restore exactly the allocation this change removes (58KB and 115KB per call at 700 and 1500 epoch rows). The contract is documented instead, on `HardForkSummary` and `hardForkSummaryAnchoredAt`: the returned summary is shared and read-only, and callers needing a mutable one copy it first. Re-checked at f02fba195 after merging current `origin/main`.

ledger/hardfork_summary.go
chrisguiney: Fixed by taking the first option. The description no longer carries a closing keyword, and `closingIssuesReferences` is now empty on the API, so merging no longer closes #2093. The issue is referenced without closing, and a Scope section states the boundary: `queryHardForkEraHistory` builds its era history from `GetEpochsByEra` reads and calls `hardfork.BuildSummary` directly, and is unchanged here.

It is not routed through this entry rather than merely left out. Its inputs are database epoch and protocol-parameter rows, not the published `consensusSnapshot`, so a `generation`-keyed entry would not be invalidated by construction the way the in-memory path's is; and its `[]any` result carries `*big.Int` values into the query encoder, so a shared entry there would be shared mutable state on the NtC path. #2093's own first forcing function is profiling that shows `queryHardForkEraHistory` hot, which this change has not measured. #2093 stays open for that half.

ledger/hardfork_summary.go
chrisguiney: Key coverage re-established against the merged tree at f02fba195, not carried over: removing this lookup in place fails `TestHardForkSummaryCache_ServesCachedResultWithinGeneration` at three assertions (cached within a generation, rebuilt entry cached at the new generation, repeated call at the same anchor).

One thing changed under the merge and is worth naming, since it was not in the tree you reviewed. #3865 landed on main and replaced the `eraShape()` fallback in this function with `eraShapeWithError()`, making a missing shape or a current era absent from the shape a hard error; both returns land in `buildHardForkSummary`, so the entry can now hold an error rather than a summary. That is safe under the same key: `eraShapeWithError` resolves only from `ls.config.CardanoNodeConfig` and `ls.config.EnableDijkstra`, both fixed for the `LedgerState` lifetime, and the missing-era branch reads `consensusState.currentEra`, which is part of the keyed snapshot. Neither error can be transient while a generation is held, so no recoverable failure is pinned.

On the merged tree: `go build ./...`, `go vet ./ledger/...`, `go test -race ./ledger/...` (all nine packages ok), `golangci-lint run ./ledger/` (0 issues), `make import-boundaries`, `make docs-parity`, and `go test ./internal/test/conformance/` all pass.

wolf31o2 (approved):

PR #4452, opened 2026-09-18
fix(governance): ratify SPO actions against the same-boundary mark

### LINKED ISSUE #4441: governance: Plomin HardForkInitiation never ratifies on Preview, pinning the node at PV9
## Symptom

Four Preview sync instances halt deterministically at the same transaction:

```
ledger pipeline stopped after repeated no-progress restarts; operator intervention is required
component=ledger consecutive_no_progress=50 tip_slot=64275062
error="process block batch: tx 8af400db5dc4e2dab8f662ccc3bbcc3a5701c81ed69e8f6b34691f2eb56b26f0
validation failure at slot 64275484: conway utxo validation rule 54: bootstrap-phase voting
restriction: voter 2eed24ec0f9cb10d on action c96a9b5b8668bf1b#0 - voter type 2 cannot vote on
action type 0 during bootstrap phase"
```

## The rejected transaction is canonical

Koios (`preview.koios.rest`):

- `tx_info` for `8af400db...`: block `63d6f8bb94e00f9301eb30dd63b00db5dd4768e473cfb9d83c5a030f376b7e13`, height 2648961, epoch 743, absolute slot 64275484.
- Decoding the body CBOR the node logs: key 19 (voting_procedures), voter `[2, h'2eed24ec0f9cb10dbd3e888f14cce9d7dee2b4fd82397893e468bcbe']` = DRep key hash (`VoterTypeDRepKeyHash`), vote `[0, null]` = No, on action `c96a9b5b...#0`.
- `proposal_list` for `c96a9b5b...`: `proposal_type` = `ParameterChange`, i.e. `GovActionType` 0.
- `proposal_votes` for that action: 19 DRep votes, earliest being DRep `2eed24ec0f9cb10dbd3e888f14cce9d7dee2b4fd82397893e468bcbe` voting No on 2024-11-06.

So "voter type 2 on action type 0" is a DRep voting on a ParameterChange, and it is on the confirmed chain.

## The gouroboros rule is correct — it should not be changed

`UtxoValidateBootstrapVotingRestrictions` (gouroboros `ledger/conway/rules.go:4419`) is a faithful port of upstream `checkBootstrapVotes` (cardano-ledger `eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs`, master L378-390, tag `cardano-ledger-conway-1.16.0.0` L318-330):

```haskell
checkBootstrapVotes pp votes
  | hardforkConwayBootstrapPhase (pp ^. ppProtocolVersionL) =
      checkDisallowedVotes votes DisallowedVotesDuringBootstrap $ \gas ->
        \case
          DRepVoter {} | gasAction gas == InfoAction -> True
          DRepVoter {} -> False
          _ -> isBootstrapAction $ gasAction gas
  | otherwise = pure ()
```

`isInConwayBootstrapPhase` (`ledger/conway/rules.go:344`, major >= 9 and major < 10) matches upstream `bootstrapPhase pv = pvMajor pv == natVersion @9` in `eras/shelley/impl/src/Cardano/Ledger/Shelley/HardForks.hs`. The rule fires only because the node still believes it is at PV9.

## Root cause: the Plomin hard fork is never enacted

Koios `epoch_params`: Preview is `protocol_major` 9 through epoch 742 and 10 from epoch 743. The bump comes from `HardForkInitiation` `049ae5d612b2fa825655809133b023d60c7f8cac683c278cf95de1622e4592f3#0`, body `[null, {major: 10, minor: 0}]`, proposed epoch 737, ratified 742, enacted 743.

Reading a halted instance's metadata store read-only:

- `governance_proposal` holds that proposal (action_type 1, proposed_epoch 737, expires_epoch 767) with `ratified_epoch` and `enacted_epoch` both NULL, `expired_epoch` NULL, `deleted_slot` NULL.
- `governance_vote` for it holds 49 SPO Yes and 1 CC Yes — the same 49 SPO + 1 CC that Koios `proposal_votes` reports for it.
- The newest `pparams` row is epoch 736 and still encodes protocol version `[9, 0]`. No PV10 row exists.
- Logs show the tally ran at every boundary (`governance epoch tally starting ... epoch=738` through `epoch=743`), so the proposal was evaluated and rejected, not skipped.

The proposal has no parent and the hard-fork purpose root is empty, so `validateParentChain` passes. The committee gate passes too: `committee_member` holds one seated member expiring epoch 1000, `auth_committee_hot` maps it to `64bf0826f74edf688a28eabdfcbd191292ef41ca047afd6b807924b7`, which is exactly the CC voter, so the CC yes ratio is 1/1 and `conwayRatifyQuorum` never yields zero.

That leaves the SPO gate in `ShouldRatify` (`ledger/governance/ratify.go:140`). `pvt_hard_fork_initiation` on Preview is 0.51. Computing `SPOYesRatio` from the live snapshot rows:

| snapshot | yes stake | total stake | ratio |
| --- | --- | --- | --- |
| mark epoch 740 | 270,340,804,070,722 | 565,684,179,845,146 | 0.4779 |
| mark epoch 741 | 270,847,562,824,769 | 569,366,654,722,147 | 0.4757 |
| mark epoch 742 | 521,175,367,709,588 | 829,427,061,892,229 | 0.6283 |

Abstain stake is zero here: all 49 SPO votes are Yes, and `tallySPOVotes` already skips reward-account auto-votes for HardForkInitiation, matching upstream `spoAcceptedRatio`.

`stakeEpochFor` (`ledger/governance/epoch.go:912`) returns `newEpoch - 2`, so the boundary tally into epoch 743 uses the epoch-741 mark and gets 0.4757, below 0.51. Only the epoch-742 mark clears the threshold.

The snapshot labels are offset by one against the chain. Koios `epoch_info.active_stake` is 565,684,179,845,146 for epoch 741, 569,366,654,722,147 for 742 and 829,427,061,892,229 for 743 — our `mark[E]` equals the chain's active stake for epoch `E+1`. Of the three, only `mark[742]`, i.e. `newEpoch - 1`, reproduces the chain's own ratification decision at that boundary.

## Which snapshot upstream actually uses

Settled from cardano-ledger source.

`Cardano.Ledger.Conway.Rules.Epoch` seeds the ratification pulser from the snapshot rotated by SNAP at that same boundary:

```haskell
  snapshots1 <- trans @(EraRule "SNAP" era) $ TRC (Shelley.SnapEnv ledgerState0 curPParams, snapshots0, ())
  ...
  stakePoolDistr = ssStakeMarkPoolDistr snapshots1
  ...
  liftSTS $ setFreshDRepPulsingState eNo stakePoolDistr epochState2
```

The pulser seeded at the boundary into epoch N is consumed by RATIFY at the boundary into N+1, so RATIFY at the boundary into epoch N uses the mark captured at the boundary into N-1. Upstream RATIFY and ENACT run in the same EPOCH transition.

Our snapshots are labelled the same way: `pool_stake_snapshot` rows for `mark` epoch E have `captured_slot = E*86400 - 1` on Preview, the last slot before the boundary into E. Independently, `mark[E]` equals the chain's `active_stake[E+1]` for every epoch checked (741, 742, 743), which is what `nesPd` for epoch N being the mark captured at the boundary into N-1 predicts.

We split ratification and enactment across two boundaries: the decision is taken at the boundary into M and applied at the boundary into M+1. For an enactment to land on the same boundary as upstream's, our decision at the boundary into M must reproduce upstream's decision at the boundary into M+1, which uses the mark captured at the boundary into M, that is `mark[M]` = `mark[newEpoch]`.

`stakeEpochFor` returns `newEpoch - 2`. It should return `newEpoch`.

Checked against this incident: enactment has to land at the boundary into 743, so we must ratify at the boundary into 742 using `mark[742]`, giving 0.6283 >= 0.51 and ratified 742 / enacted 743 -- exactly what Koios reports.

## Why this is not a one-line change

`mark[742]` is captured at the boundary into 742, the same boundary at which the epoch-742 tally runs, and our boundary currently runs the governance tally before the snapshot rotation: the logs show `governance epoch tally starting ... epoch=742` at 16:18:48 and `handling epoch transition previous_epoch=741 new_epoch=742` at 16:18:49. The snapshot the tally needs is written after the tally reads it. Changing `stakeEpochFor` alone would read a row that does not exist yet.

So the fix needs the epoch-boundary ordering resolved as well, and it changes SPO voting power for every action type carrying an SPO threshold, not only HardForkInitiation. That is consensus-critical and warrants its own design and regression coverage rather than a quick patch.

## Impact

Any Preview sync past epoch 742 stalls at this transaction. The node stays at PV9 indefinitely, so every PV10-era DRep vote on a non-Info action is rejected by a rule that is itself correct, and every later protocol-version-dependent rule would diverge as well.

chrisguiney:
`stakeEpochFor` returned `newEpoch-2` as the mark-snapshot epoch for the SPO ratification-threshold tally. cardano-ledger's Conway `EPOCH` rule seeds a fresh DRep pulser with `ssStakeMarkPoolDistr snapshots1` — the mark SNAP captured at that same boundary — and that pulser is not extracted until the next boundary, so upstream's ratification applied at the boundary into X consumes `mark[X-1]`. Dingo makes the ratify decision in full at one tick and enacts at the next, so its decision at the boundary into M reproduces upstream's at M+1 and must use `mark[M]`. `stakeEpochFor` is now the identity function.

On Preview this pinned the Plomin hard fork below its 0.51 SPO threshold indefinitely: `mark[740]` yes-ratio 0.4779 and `mark[741]` 0.4757, against `mark[742]` 0.6283 and the network's own `ratified_epoch=742` / `enacted_epoch=743`.

Changing only the offset would be worse than the defect. `mark[NewEpoch]` is not in `pool_stake_snapshot` when RATIFY runs: that row is written at the end of `processEpochRollover`, after governance, because it needs the new epoch's nonce and post-enactment protocol version. RATIFY would read a row that does not exist and see zero SPO stake at every boundary.

- `snapshot.Manager.CurrentBoundarySPOStakeRows` resolves `mark[NewEpoch]` and writes nothing. It peeks — without consuming — the SNAP-point distribution `ComputeEpochBoundarySnapshot` already stashed for the persisted write, falls back to the same historical boundary reconstruction that write falls back to, and resolves the CIP-1694 reward-account auto-vote.
- `LedgerState.SetCurrentBoundarySPOStakeHook` installs it alongside the two existing epoch-boundary hooks, in `node.go`, `node_lifecycle.go`, and `internal/node/load.go`.
- `processEpochRollover` calls it at the SNAP read point and passes the result as `governance.EpochInput.CurrentBoundarySPOState`, which RATIFY prefers over `LoadSPOVotingState`.

That call site is load bearing. Its no-stash fallback reconstructs the boundary itself and counts reward deltas through the boundary slot, so running it below `applyPoolRetirements` or `ProcessEpoch` would tally SPO votes against a mark carrying deposit refunds and enactment credits. `TestProcessEpochRollover_SnapStakeReadOrdering` locks the position.

The offset is shared by every SPO-gated action, not only `HardForkInitiation`: `NoConfidence`, `UpdateCommittee`, and a `ParameterChange` touching the security-parameter group use the same `tallySPOVotes` path.

Closes #4441

---
## Summary by cubic
Fixes #4441 by making SPO-gated governance ratifications use the mark snapshot captured at the same epoch boundary. The old `newEpoch-2` offset stalled the Plomin hard fork on Preview below its 0.51 SPO threshold.

**What changed**
- `stakeEpochFor` is now the identity function, and this affects `HardForkInitiation`, `NoConfidence`, `UpdateCommittee`, and security-parameter `ParameterChange`.
- `CurrentBoundarySPOStakeRows` supplies `mark[newEpoch]` before the persisted row exists, peeking the SNAP-point distribution or reconstructing it like the persisted write does, without writing anything.
- The rollover passes that state into `ProcessEpoch`, and the new hook is wired in `node.go`, `node_lifecycle.go`, and `internal/node/load.go`.

**Rollout note**
- Custom setups that wire ledger hooks themselves must also install `SetCurrentBoundarySPOStakeHook`; without it, every SPO-gated action sees zero stake and never ratifies.

<sup>Written for commit e6b7f8afbf6dac62183fd51475986fa6b858ed4e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4452?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Governance decisions at epoch boundaries now use the correct current-boundary SPO stake, improving ratification accuracy.
  - Fixed an issue that could prevent SPO-gated proposals from ratifying during epoch transitions.
  - Corrected hard-fork proposal enactment timing at epoch boundaries.
  - Ensured the same governance behavior during normal operation, database loading, and recovery.

- **Documentation**
  - Updated architecture and rollover documentation to describe current-boundary governance stake handling.


ledger/transition_stability_test.go
```diff
@@ -206,9 +206,14 @@ INSERT INTO auth_committee_hot (
...
 	poolCred := repeatByte(28, 0xDD)
+	// governance.stakeEpochFor(currentEpoch) resolves to currentEpoch itself
+	// (dingo#4441): EvaluateRatifiableHardForkInitiation's mid-epoch check
+	// tallies the SPO vote against mark[currentEpoch], the mark snapshot
+	// already durably written at the boundary that opened the currently
+	// active epoch.
 	require.NoError(t, db.Metadata().SavePoolStakeSnapshot(
 		&models.PoolStakeSnapshot{
-			Epoch:        currentEpoch - 2,
+			Epoch:        currentEpoch,
```
wolf31o2, line 216: This seed follows stakeEpochFor, but EvaluateRatifiableHardForkInitiation predicts the boundary into currentEpoch+1, which now reads mark[currentEpoch+1], not the mark[currentEpoch] the mid-epoch path loads (stability.go sets StakeEpoch and never SPOState, so tallySPOVotes lazily loads that epoch). Reseeding this row at currentEpoch+1 makes PostDeadline_Ratifiable_SetsKnown fail. ProcessEpoch asks that the two paths stay in sync, and the incident numbers mark[741]=0.4757 against mark[742]=0.6283 straddle the 0.51 threshold.

ledger/governance/epoch.go
```diff
@@ -905,18 +934,41 @@ func ratificationEnactmentPrecondition(
...
+// caller supplies this same-boundary data instead.
 func stakeEpochFor(newEpoch uint64) uint64 {
-	switch {
-	case newEpoch >= 2:
-		return newEpoch - 2
-	case newEpoch >= 1:
-		return newEpoch - 1
-	}
-	return 0
+	return newEpoch
```
wolf31o2, line 971: stakeEpochFor now resolves to a snapshot SNAP takes at the boundary itself, so the SPO denominator is no longer fixed before that boundary. evaluateHardForkInitiationStability in ledger/state.go rests its mid-epoch answer on the inputs being frozen at the voting deadline: votes do freeze there, but this stake is read at the boundary from live state that keeps moving after it. Either that rationale needs correcting or the mid-epoch check needs the same-boundary estimate the rollover now supplies.

ledger/chainsync.go
```diff
@@ -7070,6 +7090,43 @@ func (ls *LedgerState) captureEpochBoundarySnapshotStake(
...
+// SetCurrentBoundarySPOStakeHook's doc comment for why a production node
+// must always have this wired).
+func (ls *LedgerState) currentBoundarySPOStakeState(
+	txn *database.Txn,
+	prevEpoch models.Epoch,
+	boundarySlot uint64,
+) (*governance.SPOVotingState, error) {
+	hook := ls.currentBoundarySPOStakeHookFn()
+	if hook == nil {
+		return nil, nil
```
wolf31o2, line 7109: An unset hook yields no override, and ProcessEpoch then reads mark[NewEpoch] from pool_stake_snapshot, which cannot exist yet inside a rollover, so every SPO-gated action tallies zero stake and never ratifies, with no error and no log line. All three in-repo wiring sites are updated, but LedgerState and these setters are exported, so an embedder carrying only the other two hooks diverges from the network silently. Returning an error here, or at least logging a warning at a Conway boundary, would surface it.

wolf31o2 (changes requested):
Three inline findings. The mid-epoch ratification predictor still tallies SPO votes against mark[CurrentEpoch] while the boundary it predicts now reads mark[CurrentEpoch+1], and the new same-boundary hook degrades silently to zero SPO stake when it is not installed. The stakeEpochFor offset itself holds up: cardano-ledger Conway Epoch.hs extracts RATIFY from the pulser the previous boundary seeded while ssStakeMarkPoolDistr snapshots1 seeds the next one, and the Preview record of ratified_epoch=742 with enacted_epoch=743 matches mark[742] at 0.6283.

ledger/transition_stability_test.go
chrisguiney: Made explicit rather than reseeded, in c4f498ff9.

The seed now follows `predictedBoundaryStakeEpochFor(currentEpoch)`, a separate function from the `stakeEpochFor(newEpoch)` the boundary uses, so the two reads can no longer look identical by accident.

Reseeding at `currentEpoch+1` would point the mid-epoch check at a row that does not exist while the epoch is still being applied: SNAP captures `mark[currentEpoch+1]` at the boundary itself, so `LoadSPOVotingState` returns no rows and the tally runs on a zero denominator, which refuses everything.

`ledger/governance/same_boundary_spo_state_test.go` now locks both reads against one database carrying the incident's own numbers -- 0.4757 at `mark[741]`, 0.6283 at `mark[742]`, 0.51 threshold. The mid-epoch check declines and the boundary into 742 ratifies, and each half was confirmed to flip when its own mark is moved across the threshold.

ledger/governance/epoch.go
chrisguiney: The rationale is corrected and the mid-epoch check keeps the committed mark, in c4f498ff9.

`predictedBoundaryStakeEpochFor` now states the property directly: votes freeze at the voting deadline, the SPO denominator does not, so "would ratify now" is a prediction of the boundary rather than a preview of it. `evaluateHardForkInitiationStability` no longer claims every ratification input is frozen there, and ARCHITECTURE.md records the divergence beside the hook description.

The same-boundary estimate was the other option and is the worse trade here. `transitionInfo` feeds `hardfork.BuildSummary`, which bounds the current era at the announced boundary and appends a successor era, and `verify_header.go`'s forecast-horizon gate reads that summary. An input that can still move until the boundary buys an earlier announcement at the price of a wrong era layout, while the conservative read costs only advance notice -- and the rollover resets `transitionInfo` to Unknown at every boundary, so neither answer outlives the epoch that made it.

ledger/chainsync.go
chrisguiney: Now fails loudly, in c4f498ff9.

`ProcessEpoch` returns `ErrMissingCurrentBoundarySPOState` when `CurrentBoundarySPOState` is nil, the `mark[NewEpoch]` fallback read comes back empty, and `mark[NewEpoch-1]` still holds pool stake. An embedder carrying only the other two hooks aborts the boundary instead of tallying zero against every SPO-gated action.

The previous boundary's mark is the gate because it is what makes the empty read a contradiction rather than a fact: a chain whose pools have never held snapshot stake has nothing to tally under any wiring, and several standalone `ProcessEpoch` callers are exactly that shape.

`TestProcessEpoch_MissingSameBoundarySPOState_Errors` covers the un-wired path and fails without the guard by returning no error at all; `TestProcessEpoch_SuppliedSameBoundarySPOState_Ratifies` is the control that the guard never fires on a wired one.

wolf31o2 (dismissed):

ledger/transition_stability_test.go
chrisguiney: Unchanged by the merge of main in 0479a737d: `git diff c45ee5a2a 0479a737d -- ledger/transition_stability_test.go ledger/governance/same_boundary_spo_state_test.go` is empty, so the `predictedBoundaryStakeEpochFor(currentEpoch)` seed and `TestMidEpochPredictionAndBoundaryReadDifferentMarks` stand as reviewed.

ledger/governance/epoch.go
chrisguiney: Unchanged by the merge of main in 0479a737d: `git diff c45ee5a2a 0479a737d -- ledger/governance/epoch.go ledger/governance/stability.go` is empty, so `predictedBoundaryStakeEpochFor` and its rationale stand as reviewed.

ledger/chainsync.go
chrisguiney: Unchanged by the merge of main in 0479a737d: `git diff c45ee5a2a 0479a737d -- ledger/chainsync.go ledger/governance/epoch.go` is empty, so the `ErrMissingCurrentBoundarySPOState` guard and `TestProcessEpoch_MissingSameBoundarySPOState_Errors` stand as reviewed.

wolf31o2 (approved):

Issue #4464, opened 2026-09-18
database: SQLite write path still amplifies physical writes 41x-67x over net logical growth after the wal_autocheckpoint/TRUNCATE fix


chrisguiney:
## Summary

On a fresh four-instance A/B harness (perf / perf-no-koios / vanilla /
vanilla-no-koios, all syncing Preview from genesis, ~18-20 minutes uptime,
all healthy, `dingo_tip_gap_slots=0` on all four), the SQLite metadata
store's actual physical disk writes exceed the store's own reported logical
growth (`dingo_database_sql_disk_bytes`) by **41x-67x** on every one of the
four variants, measured two independent ways in two independent sampling
windows. This holds after the existing `wal_autocheckpoint`/forced-TRUNCATE
fix (`fb5595dd6`, `8e400073c`, already present on both `origin/main` and the
perf branch — confirmed by grepping both source checkouts), so it is not the
already-fixed "checkpoint never triggers" defect; it's what remains after
that fix.

## Measurement

`/proc/<dingo-pid>/io`'s `write_bytes` (actual bytes written to the block
device, attributed by the kernel to the process) sampled at t0 and t0+45s,
compared against `dingo_database_sql_disk_bytes` (the store's own
`Store.DiskSize()`-backed gauge) and `cardano_node_metrics_blockNum_int` over
the same window:

| Variant | write_bytes rate | disk_bytes growth | amplification | blocks/sec | write/block |
|---|---|---|---|---|---|
| perf | 45.5 MB/s | 1.11 MB/s | **41.1x** | 100.0 | 465 KB |
| perf-no-koios | 35.7 MB/s | 0.69 MB/s | **51.8x** | 77.7 | 471 KB |
| vanilla-no-koios | 37.0 MB/s | 0.60 MB/s | **61.2x** | 59.3 | 639 KB |
| vanilla | 64.5 MB/s | 0.96 MB/s | **67.1x** | 75.6 | 874 KB |

(PIDs mapped via `docker inspect --format '{{.State.Pid}}'` → container's own
child process; verified against `ps --ppid`.)

This holds regardless of `koios-parity`: perf-no-koios and vanilla-no-koios
(no external Koios calls, no koios-parity SQL activity) show the same order
of magnitude (52x/61x) as their koios-parity-enabled siblings (41x/67x), so
this is not a koios-parity-specific cost — it is a property of the base
SQLite write path.

Host-level corroboration (`iostat -x` against the underlying NVMe,
[...]

PR #4472, opened 2026-09-18
fix(ledger): coalesce premature batch flushes during bulk replay

chrisguiney:
Addresses one of the two confirmed causes in #4464. It does not close the issue: the separate index-churn cause is untouched.

## Defect

`ledgerReadChainIterator`'s gather loop filled `rawBatch` (capacity `batchSize`, 50) until either the batch was full or a non-blocking `iter.Next(false)` returned `chain.ErrIteratorChainTip`. The second condition flushed immediately, even with one or two blocks gathered. During bulk replay a momentary `ErrIteratorChainTip` usually means the goroutine appending to `ls.chain` is a beat behind this reader, not that the chain stopped growing, so batches committed at a fraction of their intended size and multiplied SQLite's physical write volume over logical growth.

## Change

- An under-full batch now gets a bounded retry window before flushing: up to `gatherCoalesceMaxAttempts` (10) retries spaced `gatherCoalesceRetryInterval` (2ms) apart, per gap. Both are package vars so tests can shrink them.
- The wait is gated on `!ls.isNearTip(...)`, so a solitary block at the live chain tip still commits immediately rather than waiting for a batch that will never fill.
- `blockPipelineGatherMutex` is deliberately held across the wait. `rawBatch` already holds gathered-but-unsubmitted blocks, and releasing the lock would let a rollback's write lock be granted, observe an empty `blockPipeline` via `drainBlockPipelineBeforeRollback`, and proceed ahead of those blocks — the race the mutex exists to close.
- The attempt counter resets on every appended block, so the delay a blocked rollback can see is bounded per gather pass, not per gap: up to `batchSize * gatherCoalesceMaxAttempts * gatherCoalesceRetryInterval`, about 1s at the defaults, and about 100ms when gaps resolve on their first retry. `ARCHITECTURE.md` and the inline comment state that bound. It is accepted because the `!isNearTip` gate confines the wait to catch-up.
- New `dingo_ledger_commit_batch_blocks` histogram records the submitted batch size, so the distribution shift is observable without re-measuring disk I/O. Empty gather passes are not observed, since a zero sample would land in the lowest bucket of the distribution the histogram exists to measure.

The retry constants are an untuned starting point. They were chosen by inspection, not measured against the harness that produced the issue, and are exposed as vars so they can be adjusted once the histogram gives a distribution to tune against.

## Tests

`ledger/read_chain_batch_coalesce_test.go` covers the coalescing path, the near-tip immediate-flush guard, and the empty-pass metric guard. Each fails with its corresponding behavior reverted in place: batch of 1 instead of 5, a 10s timeout at live tip, and a zero sample recorded.


ledger/state.go
```diff
@@ -5370,6 +5394,51 @@ func (ls *LedgerState) ledgerReadChainIterator(
...
+					// can wait once per block it gathers. A rollback
+					// blocked on the write lock therefore waits up to
+					// batchSize*gatherCoalesceMaxAttempts*gatherCoalesceRetryInterval
+					// (about 1s at the defaults) in the worst case, and
+					// about batchSize*gatherCoalesceRetryInterval (100ms)
+					// when each gap resolves on its first retry. That is
+					// accepted because the whole wait is gated on
+					// !isNearTip: it applies only while catching up, where
+					// rollbacks are rare and no forging depends on them.
+					if len(rawBatch) > 0 && len(rawBatch) < cap(rawBatch) &&
```
wolf31o2, line 5430: Holding blockPipelineGatherMutex across this wait is the safety property this hunk rests on, and nothing tests it. Replacing the wait with releaseGatherLock, the same sleep, then a re-RLock leaves go test ./ledger/ green at exit 0. The existing probe in block_pipeline_gather_mutex_test.go does not reach here: its scripted reader pauses inside Next, never inside this wait. A later change that releases the lock here to shorten the rollback wait would reopen the #1894 phase 5 race with every test still green. Pin it with a TryLock probe taken while the reader sits in this wait.

ledger/state.go
```diff
@@ -5370,6 +5394,51 @@ func (ls *LedgerState) ledgerReadChainIterator(
...
+						coalesceAttempts < gatherCoalesceMaxAttempts &&
+						!ls.isNearTip(rawBatch[len(rawBatch)-1].Slot) {
```
wolf31o2, line 5432: ls.isNearTip reaches calculateStabilityWindow, which takes ls.RLock, and it runs here while blockPipelineGatherMutex is still held, up to batchSize*gatherCoalesceMaxAttempts times per gather pass. The other work under that span reads atomic snapshots only: slotCoveredByMithril through mithrilLedgerSlotSnapshot and epochForSlot through loadConsensusSnapshot. Go RWMutex writer preference lets a pending ls.Lock from block apply park the reader here while it still holds the gather read lock. Hoist the window out of the retry loop and call isNearTipWithStabilityWindow.

ARCHITECTURE.md
```diff
@@ -11981,6 +11981,40 @@ state. A single decode failure anywhere in a batch discards the whole
...
+`blockPipelineGatherMutex`: `rawBatch` already holds real gathered blocks a
+concurrent rollback must not race ahead of, for the same reason the lock
+stays held through `decodeReadChainBatch` (see `gatherLockHeld`'s doc
+comment) — releasing it here would let a rollback's write lock be granted,
+see an empty `blockPipeline` via `drainBlockPipelineBeforeRollback`, and
+proceed before this pass's already-gathered blocks are submitted, reopening
+the exact race `blockPipelineGatherMutex` exists to close. The bound on that
+wait is per gap rather than per gather pass — the attempt counter is reset
+each time a block is appended — so a rollback blocked on the write lock waits
+up to `batchSize*gatherCoalesceMaxAttempts*gatherCoalesceRetryInterval` (about
```
wolf31o2, line 12009: This gives the worst case a blocked rollback sees as batchSize*gatherCoalesceMaxAttempts*gatherCoalesceRetryInterval, about 1s. That is not a bound: every attempt evaluates ls.isNearTip, which takes ls.RLock through calculateStabilityWindow while blockPipelineGatherMutex is still held, so the reader can park on the ledger write lock inside the wait. Either hoist the stability window out of the retry loop so the figure holds, or fold the ledger-lock wait into the stated bound.

ledger/state.go
```diff
@@ -68,6 +68,26 @@ import (
...
+// (blockfetch/pipeline apply) is a beat behind this reader, and vice versa.
+// Without this bounded wait, that race flushed batches at ~6-9 of the
+// intended batchSize (50) blocks instead of coalescing them, multiplying
+// SQLite's physical write volume well beyond logical growth (dingo#4464).
+// Once isNearTip is true, this wait is skipped entirely: a solitary new
+// block at live tip must still commit promptly rather than wait for a batch
+// that will never fill. Vars, not consts, so tests can shrink the interval
+// instead of waiting on real sleeps.
+var (
+	gatherCoalesceRetryInterval = 2 * time.Millisecond
```
wolf31o2, line 87: The claim that this wait applies only while catching up rests on isNearTip, but UpstreamTipSlot returns 0 whenever GetActiveConnectionFunc yields no live connection or syncUpstreamState carries another connection key, and isNearTipWithStabilityWindow folds unknown upstream into not-near. nearUpstreamTip's own doc comment calls that the safe answer for its callers but not for all of them. A caught-up node that loses its upstream then pays the full wait per gap, gather lock held, including for locally forged blocks.

ledger/metrics.go
```diff
@@ -287,6 +287,13 @@ type stateMetrics struct {
 	utxoCreatedTotal  *prometheus.CounterVec
 	utxoConsumedTotal *prometheus.CounterVec
 	certificatesTotal *prometheus.CounterVec
+	// commitBatchBlocks observes len(nextBatch) each time
+	// ledgerReadChainIterator submits a gathered batch of blocks downstream
+	// for a single DB transaction (batchSize caps it at 50). Added
+	// alongside the dingo#4464 premature-flush fix so a future run can
+	// confirm the batch-size distribution actually shifted upward, rather
+	// than relying on re-measuring physical disk I/O.
+	commitBatchBlocks prometheus.Histogram
```
wolf31o2, line 296: This file documents its accessors as tolerating a zero-value stateMetrics, but the new histogram is read directly at the state.go call site behind an inline commitBatchBlocks != nil check. An observeCommitBatchBlocks accessor on stateMetrics would keep that nil tolerance in one place, matching the metrics around it.

wolf31o2 (changes requested):
Two required changes. Holding blockPipelineGatherMutex across the new coalesce wait is the load-bearing safety property in this diff and nothing tests it: releasing the lock across that wait in place leaves go test ./ledger/ green at exit 0. And ls.isNearTip takes the ledger RWMutex inside that same held span, so the roughly 1s bound ARCHITECTURE.md states for a blocked rollback does not hold. Merged against current origin/main and verified green there: build, vet, golangci-lint, docs-parity, import-boundaries, the ledger package, and -race on the affected tests.

ledger/state.go
chrisguiney: Pinned by TestLedgerReadChainIteratorHoldsGatherMutexAcrossCoalesceWait in ledger/read_chain_batch_coalesce_test.go. A scripted iterator parks the reader inside the Next call that reports the first chain-tip gap, so it is let go into the coalescing branch with one block already gathered; the test then asserts blockPipelineGatherMutex.TryLock() never succeeds for half the coalesce interval, and that the batch has still not been delivered by the end of that window.

Confirmed causal: replacing the wait with releaseGatherLock, the same sleep and a re-RLock makes the probe succeed on its first poll, while go test ./ledger/ was green at exit 0 under that same edit before this test existed.

ledger/state.go
chrisguiney: Hoisted as suggested. calculateStabilityWindow is now called once per gather pass, before that pass takes any blockPipelineGatherMutex read lock, and the span calls isNearTipWithStabilityWindow with the value. Every remaining term the span tests is an atomic read. Re-reading the window per pass means at worst one pass of at most batchSize blocks is judged against the previous era's window, which is a multi-thousand-slot threshold and so cannot flip the near-tip answer.

TestLedgerReadChainIteratorTakesNoLedgerLockInsideGatherSpan covers it: a block apply holds ls.Lock() while the reader sits in the gather span, and the pass must still finish and release the gather mutex. Restoring the ls.isNearTip call in place makes that assertion fail.

ARCHITECTURE.md
chrisguiney: Hoisted rather than folded into the figure, so the stated worst case now holds as written. ARCHITECTURE.md says why in the same section: the figures are a bound only because nothing evaluated inside the held span can block on another lock, the stability window is therefore read once per gather pass before the gather lock is taken, and every other term the span tests (reachedTip, UpstreamTipSlot) is an atomic read.

ledger/state.go
chrisguiney: Real, and fixed rather than documented. The wait is now gated on !ls.reachedTip.Load() as well as the near-tip check. reachedTip latches the first time the node reaches the stability window and never clears, so it separates "catching up and not yet connected" from "was at tip, lost the upstream": an unknown upstream no longer re-enables the wait on a caught-up node, including for locally forged blocks.

TestLedgerReadChainIteratorSkipsCoalesceAfterReachingTip sets reachedTip with no upstream tip known and requires the single block to be committed without delay against a ten-minute retry interval. It times out at the receive deadline without the reachedTip term.

ledger/metrics.go
chrisguiney: Added observeCommitBatchBlocks on stateMetrics, with the same nil tolerance as the accessors above it, and the state.go call site now goes through it. The len(nextBatch) > 0 guard stays at the call site: skipping gather passes that submit nothing is a property of what is worth observing, not of whether the metric was initialised.

ARCHITECTURE.md
```diff
@@ -12214,6 +12214,67 @@ state. A single decode failure anywhere in a batch discards the whole
...
+
+Those figures are a bound only because nothing evaluated inside the held span
+can block on another lock. `ls.isNearTip` would break that: it reaches
+`calculateStabilityWindow`, which takes `ls.RLock`, and Go's `RWMutex` parks a
+reader behind a pending writer, so a block apply holding `ls.Lock()` could
+stall the pass with `blockPipelineGatherMutex` still held — an unbounded term
+the figures above do not include. The stability window is therefore read once
+per gather pass **before** the gather lock is taken, and the span calls
+`ls.isNearTipWithStabilityWindow` with that value; every other term it tests
+(`reachedTip`, `UpstreamTipSlot`) is an atomic read. Re-reading the window per
```
wolf31o2: UpstreamTipSlot is not an atomic read under the node's own wiring. It calls GetActiveConnectionFunc, which is the node_ledger_config.go closure into withLiveChainsyncState (liveLifecycleMu.TryLock) and chainsync State.GetClientConnId (clientConnIdMutex.RLock), then ConnectionLiveFunc, which takes ConnectionManager.connectionsMutex.Lock. Three mutexes are acquired inside the held gather span, up to batchSize*gatherCoalesceMaxAttempts times per pass. Every one of those sections is map-only and short, so the ~1s figure survives, but the sentence it rests on does not.

ledger/state.go
```diff
@@ -5540,6 +5580,69 @@ func (ls *LedgerState) ledgerReadChainIterator(
...
+					// gap, not per gather pass: coalesceAttempts is reset
+					// every time a block is appended below, so one pass
+					// can wait once per block it gathers. A rollback
+					// blocked on the write lock therefore waits up to
+					// batchSize*gatherCoalesceMaxAttempts*gatherCoalesceRetryInterval
+					// (about 1s at the defaults) in the worst case, and
+					// about batchSize*gatherCoalesceRetryInterval (100ms)
+					// when each gap resolves on its first retry. Those
+					// figures are a real bound only because every term
+					// tested below is an atomic read: stabilityWindow is
```
wolf31o2: Same claim as ARCHITECTURE.md, and the same correction: ls.reachedTip.Load() is an atomic read, but ls.isNearTipWithStabilityWindow reaches UpstreamTipSlot, which takes chainsync clientConnIdMutex and ConnectionManager.connectionsMutex through the configured callbacks. TestLedgerReadChainIteratorTakesNoLedgerLockInsideGatherSpan cannot see that: its LedgerState leaves GetActiveConnectionFunc nil, so UpstreamTipSlot falls to the syncUpstreamTipSlot atomic instead of the production path. The class audit asked for on the stability window stops one call short.

ledger/state.go
```diff
@@ -5505,6 +5526,25 @@ func (ls *LedgerState) ledgerReadChainIterator(
...
+		// apply's ls.Lock() stall this pass with that lock held -- folding
+		// an unbounded wait into the coalescing bound below, which is
+		// derived purely from batchSize and the gatherCoalesce* settings.
+		// Everything else the span reads is an atomic. The window only
+		// changes at an era boundary and is re-read every pass, so at worst
+		// one pass of at most batchSize blocks is judged against the
+		// previous era's window -- and the window is a multi-thousand-slot
+		// threshold, so that cannot flip the near-tip answer at an era
+		// boundary reached during bulk replay.
+		stabilityWindow := ls.calculateStabilityWindow()
```
wolf31o2, line 5577: Hoist confirmed at head: calculateStabilityWindow runs before this pass takes any blockPipelineGatherMutex read lock, and releaseGatherLock precedes the next pass. Measured causality for the three new tests: releasing and re-taking the gather lock across the coalesce wait fails TestLedgerReadChainIteratorHoldsGatherMutexAcrossCoalesceWait; restoring ls.isNearTip below fails TestLedgerReadChainIteratorTakesNoLedgerLockInsideGatherSpan; dropping the !ls.reachedTip.Load() term fails TestLedgerReadChainIteratorSkipsCoalesceAfterReachingTip.

wolf31o2 (changes requested):
All five findings from the previous review are addressed at 8ce5f009b3, and the three new tests each fail when their fix is reverted in place. One term evaluated inside the span that holds blockPipelineGatherMutex is still not lock-free, and both ARCHITECTURE.md and the code comment state the opposite; the stated bound survives but its justification does not hold as written.

ARCHITECTURE.md
chrisguiney: Confirmed on the code and fixed rather than documented. UpstreamTipSlot is now read once per gather pass beside the stability window, before the pass takes any blockPipelineGatherMutex read lock, so liveLifecycleMu, clientConnIdMutex and connectionsMutex are no longer reached inside the span; the only ledger-state term the span still tests is the reachedTip atomic. ARCHITECTURE.md now names those three call paths instead of calling UpstreamTipSlot an atomic read, and states why a per-pass snapshot is sound: the tip advances about one slot per second against a pass bounded by the ~1s figure, feeding a comparison against a multi-thousand-slot window.

ledger/state.go
chrisguiney: Fixed, and the test gap is closed with it. The span calls nearKnownUpstreamTip with a tip snapshotted before the gather lock, so isNearTipWithStabilityWindow is no longer reached from inside it; nearKnownUpstreamTip holds the unknown-upstream fold and isNearTipWithStabilityWindow now delegates to it, so the two cannot drift. TestLedgerReadChainIteratorTakesNoConnectionLocksInsideGatherSpan wires GetActiveConnectionFunc and ConnectionLiveFunc rather than leaving them nil, and records by TryLock whether blockPipelineGatherMutex was held at each callback. It asserts both callbacks ran, so it cannot pass vacuously; restoring the call inline reports 12 in-span acquisitions.

ledger/state.go
chrisguiney: Both measured properties still hold at c2893978c. The class audit now runs one call further: every other UpstreamTipSlot caller reads it with no lock held, and cleanupConsumedUtxos already releases ls.RUnlock() before reading the era window and the tip, so the gather span was the only member. Causality re-measured in place on the new code: reverting just the tip hoist fails only TestLedgerReadChainIteratorTakesNoConnectionLocksInsideGatherSpan, and reverting all the way back to ls.isNearTip fails that and TestLedgerReadChainIteratorTakesNoLedgerLockInsideGatherSpan, so the previous round's guard is intact.

ARCHITECTURE.md
```diff
@@ -12215,6 +12215,80 @@ state. A single decode failure anywhere in a batch discards the whole
...
+proceed before this pass's already-gathered blocks are submitted, reopening
+the exact race `blockPipelineGatherMutex` exists to close. The bound on that
+wait is per gap rather than per gather pass — the attempt counter is reset
+each time a block is appended — so a rollback blocked on the write lock waits
+up to `batchSize*gatherCoalesceMaxAttempts*gatherCoalesceRetryInterval` (about
+1s at the defaults) in the worst case, and about
+`batchSize*gatherCoalesceRetryInterval` (100ms) when each gap resolves on its
+first retry.
+
+Those figures are a bound only because nothing evaluated inside the held span
```
wolf31o2: This holds for the near-tip condition, not for the span. Each coalesce retry returns to the top of the inner loop and calls iter.Next(false) with the gather read lock held, and Chain.iterNext takes c.mutex.Lock plus manager.mutex.RLock and does a metadata block lookup under them. c.mutex is the lock addBlockLocked holds to advance the tip, so the contending writer is the block-append goroutine this wait exists to wait for. The ~1s figure omits up to batchSize*gatherCoalesceMaxAttempts such acquisitions.

ledger/state.go
```diff
@@ -5505,6 +5538,41 @@ func (ls *LedgerState) ledgerReadChainIterator(
...
+		// Both are safe to snapshot per pass. The window only changes at an
+		// era boundary, so at worst one pass of at most batchSize blocks is
+		// judged against the previous era's window -- a multi-thousand-slot
+		// threshold, which cannot flip the near-tip answer. The upstream tip
+		// is a chain frontier that advances about one slot per second while
+		// a whole pass is bounded by the ~1s figure below, and the answer it
+		// feeds is a stability-window comparison, so a pass-old reading
+		// cannot flip it either.
+		stabilityWindow := ls.calculateStabilityWindow()
+		upstreamTipSlot := ls.UpstreamTipSlot()
```
wolf31o2, line 5578: Hoist confirmed at head: reverting this line and restoring isNearTipWithStabilityWindow inline fails TestLedgerReadChainIteratorTakesNoConnectionLocksInsideGatherSpan at 12 in-span acquisitions, and its probe polls the real blockPipelineGatherMutex. nearKnownUpstreamTip is a pure extraction, isNearTip being the only other consumer. Dropping the reachedTip term still fails the skip-after-tip test; reverting both hoists still fails the ledger-lock span test.

wolf31o2 (changes requested):
The upstream-tip hoist holds at c2893978c: the new connection-lock test reports 12 in-span acquisitions when the call is restored inline, and both earlier causality properties still fail on their in-place reverts. One ARCHITECTURE.md sentence still overstates what the held span does. No bot reviewed this head: CodeRabbit reports incremental reviews disabled, cubic has no record on any surface.

ARCHITECTURE.md
wolf31o2: Still false for the span at e66aa855. Each coalesce retry returns to the top of the inner loop and calls iter.Next(false) at ledger/state.go:5599 with the gather read lock taken at 5597 still held, and chain.Chain.iterNext takes c.mutex.Lock plus c.manager.mutex.RLock (chain/chain.go:2725-2727) and does a metadata block lookup under them. c.mutex is the lock addBlockLocked holds to advance the tip, so the contending writer is the block-append goroutine this wait exists to wait for. The stated ~1s figure omits up to batchSize*gatherCoalesceMaxAttempts such acquisitions.

node_lifecycle.go
```diff
@@ -887,6 +887,14 @@ func (n *Node) reinitializeBackgroundManagers(ctx context.Context) error {
 // peerGov. Must run after reinitializeBackgroundManagers (mempool/
 // chainsyncState come after the background managers in Run()'s order,
 // though nothing here actually depends on them).
+//
+// Callers must hold n.liveLifecycleMu. Every field reassigned here is read
```
wolf31o2, line 891: Accurate as written: the sole production caller is reinitializeAndResume at node_lifecycle.go:1397, reached only from Restore and Truncate, which take n.liveLifecycleMu at node_lifecycle.go:1500 and 1995 and hold it across the whole quiesce-through-reinitialize sequence. No production path reaches this function unlocked, so the commit documents and pins an invariant that already held rather than changing behavior.

wolf31o2 (changes requested):
The one blocker from the previous review is untouched at e66aa855: ledger/state.go is byte-identical to c2893978c and the new commit changes only reinitializeNetworkingCore's doc comment and one node test. The four measured causality properties hold at head; the reachedTip term reverted in place fails TestLedgerReadChainIteratorSkipsCoalesceAfterReachingTip. Merged against current origin/main: build, vet, docs-parity, import-boundaries and the affected ledger and node tests pass. No bot reviewed this head: CodeRabbit reports incremental reviews disabled, cubic has no record on any surface.

wolf31o2 (approved):

PR #4473, opened 2026-09-19
fix(backfill): fail closed on malformed blocks

wolf31o2:
Addresses a root path in #3788: historical backfill now returns when block CBOR parsing or transaction-offset computation fails. It persists only the last durable checkpoint, so the failed block is retried instead of being permanently skipped.

#4061 remains containment for missing historical accounts; it does not repair this checkpoint-integrity path.

Focused `TestRun_` tests pass with and without the race detector.

---
## Summary by cubic
Backfill now fails closed on malformed blocks and on withdrawals that reference no account, instead of logging a warning and continuing. Both failures stop the run with an error so the block is retried on the next run; only the last durable checkpoint is persisted.

**Bug Fixes**
- Block CBOR parsing or offset-computation errors stop the run with the slot number and failure reason, and the checkpoint does not advance.
- Historical withdrawals with no account row return an error instead of journaling a zero previous reward; inactive account rows still retain their historical reward.
- Tests now use valid block fixtures and assert the checkpoint does not advance on malformed blocks.

<sup>Written for commit 53e6dd042c1624db34edc805cfe9296408c22c89. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4473?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Malformed blockchain data now stops backfill processing instead of being skipped.
  - Processing rolls back uncommitted changes and preserves the last successful checkpoint.
  - Checkpoints are no longer advanced when invalid block data is encountered.


internal/node/backfill_test.go
```diff
@@ -497,39 +477,27 @@ func TestRun_IncompleteCheckpointAtZeroVisitsSlotZero(t *testing.T) {
...
-
-	var logs bytes.Buffer
-	logger := slog.New(slog.NewJSONHandler(&logs, &slog.HandlerOptions{
-		Level: slog.LevelInfo,
-	}))
-	bf := NewBackfill(db, nil, logger)
+	hash := make([]byte, 32)
+	hash[0] = 1
+	require.NoError(t, db.BlockCreate(models.Block{
+		Slot: 1,
```
arepala-uml, line 538: Deleting TestRun_IncompleteCheckpointAtZeroVisitsSlotZero drops the only guard on the slot-zero resume boundary, and this test does not replace it: with a block only at slot 1 it passes whether or not the iterator visits slot 0. Proven by changing backfill.go:810 from startSlot = 0 to cp.LastSlot + 1, which is the exact regression the deleted test named: this test and TestRun_IncompleteCheckpointAtZeroStartsAtSlotZero both still pass. Seed the malformed block at slot 0 and assert the error names slot 0 instead; fail-closed makes that stronger than the log assertion it replaced.

database/plugin/metadata/sqlstore/transaction_write.go
```diff
@@ -1260,7 +1251,7 @@ WHERE credential_tag = ? AND staking_key = ?`,
 				stakeKey.Bytes(),
 			).Scan(&accountID, &reward, &fallbackActive)
 			if errors.Is(err, sql.ErrNoRows) {
```
arepala-uml, line 1253: This leaves DATABASE.md wrong. Its historical-backfill withdrawal section near line 1055 still states that an active account row is not required during historical backfill, and that with no account row at all the backfill journals previous_reward = 0 and neither creates nor reactivates a row. That path now aborts instead. make docs-parity passes, so nothing in CI catches the drift. Update that section in this change so the documented behavior matches the new fail-closed rule.

database/plugin/metadata/sqlite/shared_sqlstore_transaction_write_parity_test.go
```diff
@@ -276,40 +276,36 @@ func TestSharedSQLStoreWithdrawalRejectsExcessiveBalance(t *testing.T) {
...
-//     certificate replay can transiently deactivate a row Mithril imported
-//     active between a historical deregistration and a later
-//     re-registration certificate for the same credential -- and
-//     deregistration's account upsert never clears `reward`. The real
-//     reward must still be journaled, not discarded as 0, and the row must
-//     be neither mutated nor reactivated.
+// Existing inactive account rows retain their historical reward.
+//
+// Existing inactive account rows are not reactivated or mutated.
+//	certificate replay can transiently deactivate a row Mithril imported
```
arepala-uml: Lines 286-291 are an orphaned fragment of the comment this change replaced: the text starts mid-sentence, restates the two sentences directly above it, and renders as an indented code block in the doc comment. It also breaks formatting: gofmt -l flags this file at this head and reports the same file clean at merge base e0c68c49, so make format now rewrites it. golangci-lint run does not check formatters, which is why lint stayed green. Delete the fragment.

database/plugin/metadata/sqlite/shared_sqlstore_transaction_write_parity_test.go
```diff
@@ -367,7 +363,12 @@ func TestSharedSQLStoreHistoricalBackfillWithdrawalMissingAccount(t *testing.T)
 				true,
 				store.NewBatchAccumulator(),
 				nil,
-			))
+			)
+			if !tc.createInactive {
```
arepala-uml, line 365: After this early return everything below runs only when createInactive is true, so the else branch asserting a nil account at line 380 is now unreachable and the surrounding guard is always taken. The no-account case also stops short of the journal count and the replay-idempotence assertion it used to reach. Either drop the dead branch, or keep that case reaching the replay check so the table still covers both shapes end to end.

arepala-uml (changes requested):
Verified at 53e6dd04 against a merged tree with current main (branch is 18 behind): build, vet, golangci-lint 0 issues on touched packages, docs-parity, and the three affected packages all pass, so no semantic merge break. Both new assertions are causally sound: reverting each fix in place fails its own test. Four changes requested. The blocker is lost slot-zero resume coverage, proven by changing the startSlot special case at backfill.go:810 and watching every surviving test still pass. DATABASE.md still documents the removed fallback as current behavior.

internal/node/backfill_test.go
wolf31o2: Confirmed on the mutation: with the slot-zero resume changed to LastSlot+1, both named tests stay green. TestRun_EmitsFinalProgressForShortRun does catch it incidentally through its block count, but nothing named the boundary.

2387bc39 restores TestRun_IncompleteCheckpointAtZeroVisitsSlotZero in the form suggested: the malformed block sits at slot 0, a valid block follows at slot 1, and the run's error must name slot 0. Under the resume mutation the run returns no error at all and the test fails. TestRun_MalformedBlockDoesNotAdvanceCheckpoint is left alone, so the checkpoint assertion still has a slot the run could wrongly advance to.

database/plugin/metadata/sqlstore/transaction_write.go
wolf31o2: Confirmed; DATABASE.md is updated in 2387bc39. The no-account bullet now states that the backfill returns ErrAccountNotFound and the run stops, why an absent row is an invariant failure rather than valid history, and that the checkpoint stays before the block that failed. The inactive-row bullet is unchanged.

database/plugin/metadata/sqlite/shared_sqlstore_transaction_write_parity_test.go
wolf31o2: Confirmed: gofmt -l flags the file at this head and is clean at the merge base. 2387bc39 replaces the fragment with the inactive-row rationale it was cut from, and gofmt -l is clean on the package.

database/plugin/metadata/sqlite/shared_sqlstore_transaction_write_parity_test.go
wolf31o2: Confirmed. 2387bc39 drops the unreachable branch and the always-taken guard, and gives the no-account case its own end-state assertions rather than an early return alone: no journal row for the transaction, and no account row invented. The replay-idempotence check stays with the inactive-row case, the only shape that still reaches it now that an absent row aborts.

internal/node/backfill_test.go
```diff
@@ -474,18 +472,55 @@ func TestRun_EmitsFinalProgressForShortRun(t *testing.T) {
...
+	addValidBackfillBlocksFrom(t, db, 1, 1)
+
+	bf := NewBackfill(db, nil, slog.New(
+		slog.NewTextHandler(io.Discard, nil),
+	))
+	err := bf.Run(context.Background())
+	require.ErrorContains(t, err, "parsing block at slot 0")
+}
+
+func TestRun_MalformedBlockDoesNotAdvanceCheckpoint(t *testing.T) {
```
arepala-uml, line 519: This asserts nothing about the checkpoint: it seeds LastSlot 0 and puts the only, malformed, block at slot 1, so nothing ever commits and the persisted value is the seed either way. Deleting saveCommittedCheckpoint() at backfill.go:987 leaves this and TestRun_IncompleteCheckpointAtZeroVisitsSlotZero both passing, the same defect class as the prior blocker. The behaviour is correct: valid blocks at slots 0-3, malformed at 4 and SetBatchSize(2) persists LastSlot 3 with the call and 0 without. Give it committed progress before the malformed block and assert the last flushed slot.

internal/node/backfill.go
```diff
@@ -1009,11 +1006,10 @@ func (b *Backfill) Run(ctx context.Context) error {
...
-						"component", "backfill",
-						"slot", blk.Slot,
-						"error", oErr,
+					saveCommittedCheckpoint()
+					return fmt.Errorf(
+						"computing offsets for block at slot %d: %w",
+						blk.Slot, oErr,
 					)
 				} else {
 					// Store transaction metadata into the shared batch
```
arepala-uml, line 1015: The error path here aborts with computing offsets for block at slot %d and rewinds through the same saveCommittedCheckpoint() contract as the parse branch, but nothing reaches it: grepping computing offsets across internal/node finds no _test.go hit, while parsing block at slot has two. This is the harder of the two to reach, since the block must decode before ComputeOffsets fails, so it is the member of the pair that will drift. Worth a case that fails after decode, or a note recording that it is deliberately uncovered.

internal/node/backfill_test.go
```diff
@@ -426,7 +434,6 @@ func TestRun_EndSlotLeavesLaterBlocksForLedgerReplay(t *testing.T) {
 	require.True(t, checkpoint.Completed)
 	require.Equal(t, uint64(1), checkpoint.LastSlot)
 	require.Equal(t, uint64(1), checkpoint.TotalSlots)
-	require.Contains(t, logs.String(), `"slot":0`)
 	require.Contains(t, logs.String(), `"slot":1`)
```
arepala-uml, line 437: Switching this test to valid fixture blocks removed the per-block skipping unparseable block log these two assertions were reading. The only line left carrying a bare slot key is the throttled backfill progress log, so Contains slot:1 and NotContains slot:2 now restate require.Equal(uint64(1), checkpoint.LastSlot) three lines above rather than proving which slots the iterator visited, and the Contains slot:0 assertion was dropped for the same reason. The test no longer covers the iteration boundary its name claims.

arepala-uml (changes requested):
All four prior findings are cleared, three of them causally: the restored slot-zero test now fails under the startSlot mutation and names slot 0, removing the transaction_write.go abort fails the no-account parity case, gofmt is clean, and the unreachable branch is gone. Verified on a merged tree with main f00f2492, which merges clean: build, vet, gofmt, scoped lint, docs-parity and the three touched packages all pass, and the branch's own green checks predate that base by 55 commits. One blocker: the test carrying the new checkpoint guarantee passes with that guarantee removed.

internal/node/backfill_test.go
arepala-uml: Cleared, causally. The restored test seeds the malformed block at slot 0 and asserts parsing block at slot 0, so the boundary is named rather than incidental, and re-running my mutation (startSlot = 0 to cp.LastSlot + 1, now backfill.go:822) fails it with an error is expected but got nil.

database/plugin/metadata/sqlstore/transaction_write.go
arepala-uml: Cleared. DATABASE.md:1050-1058 now states that the no-account case returns ErrAccountNotFound and the run stops, and that the checkpoint stays before the failing block; both check out against transaction_write.go:1253 and saveCommittedCheckpoint at backfill.go:913. Removing the abort in place fails the no-account parity case, so the rule is tested as well as documented. docs-parity still does not read this section, so that check settles nothing here.

database/plugin/metadata/sqlite/shared_sqlstore_transaction_write_parity_test.go
arepala-uml: Cleared. The fragment is replaced by the inactive-row rationale it was cut from, and gofmt -l reports the package clean at head and on the merged tree with main f00f2492.

database/plugin/metadata/sqlite/shared_sqlstore_transaction_write_parity_test.go
arepala-uml: Cleared. The dead else branch and the always-taken guard are gone, and the no-account case now carries its own end-state assertions, no journal row and no invented account row, rather than an early return. Replay idempotence stays with the inactive-row case, the only shape that still reaches it.

PR #4491, opened 2026-09-19
fix(database): avoid blocking SQLite writers during WAL checkpoint

### LINKED ISSUE #4487: mithril API sync can abort on concurrent SQLite writers
## Problem

Mithril API sync can fail with `database is locked (5) (SQLITE_BUSY)` while importing UTxOs. The sync exits and requires a retry.

The sync starts ledger-state import and immutable-block loading concurrently. Ledger-state import writes metadata SQLite rows, while the immutable-load callback also writes metadata SQLite UTxO-offset rows. These writers can collide.

## Expected behavior

Concurrent Mithril load paths must coordinate shared SQLite writes or retry lock contention so a valid sync does not abort with `SQLITE_BUSY`.

## Acceptance criteria

- Add regression coverage for concurrent UTxO import and immutable-load offset writes.
- A valid Mithril API sync does not fail from transient SQLite writer contention.
- Resume behavior remains correct after an interrupted import.

wolf31o2:
Fixes #4487

Run a non-blocking PASSIVE WAL checkpoint first and only request TRUNCATE when all frames are already checkpointed. This prevents periodic checkpointing from holding the SQLite writer lock while Mithril imports UTxOs.

Tests: `go test -count=1 ./database/plugin/metadata/sqlite -run 'TestCheckpointWAL|TestWALCheckpoint'`; `go test -race -count=1 ./database/plugin/metadata/sqlite -run 'TestCheckpointWAL|TestWALCheckpoint'`.

---
## Summary by cubic
Fixes #4487 by preventing WAL checkpoints from blocking SQLite writers during Mithril UTxO imports. Checkpoints now run a `PASSIVE` pass first and only attempt `TRUNCATE` when all frames are already checkpointed, so truncation is deferred rather than forcing the writer lock.

**Tests**
- Verified with `go test` and `go test -race` for `TestCheckpointWAL` and `TestWALCheckpoint`.

<sup>Written for commit 0d57a13dd1431350cbc1ec3fdb5e2fd3dae4f321. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4491?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database checkpoint handling to reduce writer lock contention during large imports.
  * Avoids truncating the write-ahead log when active readers could be holding older snapshots.
  * Adds warnings when checkpointing cannot complete because of active readers.


database/plugin/metadata/sqlite/shared_sqlstore.go
```diff
@@ -182,19 +182,37 @@ func checkpointWAL(
 		}()
 		db.SetMaxOpenConns(1)
 
+		// Drain the WAL with PASSIVE first.  A direct TRUNCATE checkpoint can
```
arepala-uml: TestCheckpointWALTruncatesFile and TestCheckpointWALDoesNotBlockWriteBehindReaderSnapshot both pass unmodified against pre-fix direct-TRUNCATE code, so neither proves this change. Reproducing #4487's collision locally (a reader blocks wal_autocheckpoint until the WAL exceeds the 41MB threshold, then a writer races checkpointWAL) gives near-identical writer latency and no SQLITE_BUSY old vs new (211-350ms vs 191-312ms max, 3 runs; 15-90ms vs 29-90ms bounded-WAL, 5 runs). Add a test that fails without this fix, or cite the trace isolating TRUNCATE's lock hold as the cause.

arepala-uml (changes requested):
Requesting changes: the tests cited in the PR description do not exercise this change, and a local reproduction of #4487's write/read collision shows no measurable improvement over the pre-fix code (see inline comment).

database/plugin/metadata/sqlite/shared_sqlstore.go
wolf31o2: Confirmed for both named tests: neither fails against the direct-TRUNCATE code. The writer-latency measurement does not reproduce because a writer started alongside the checkpoint finishes before the checkpoint connection is even opened, so it never meets the writer lock.

TRUNCATE takes SQLite's WAL writer lock before it copies anything and holds it while its busy handler waits out the reader snapshot. Sampling that lock from a busy_timeout(0) connection for the whole call, with a reader snapshot pinning WAL frames: direct TRUNCATE blocks 7146-9180 write attempts over 5 runs and the call itself runs 298-519ms; with the PASSIVE gate it is 0 blocked attempts over 5 runs and 143-168ms.

Added as TestCheckpointWALDoesNotHoldWriterLockBehindReaderSnapshot in fce5f6f8. It fails 5/5 with the gate reverted in place. checkpointWAL is the only wal_checkpoint call site.

database/plugin/metadata/sqlite/shared_sqlstore.go
```diff
@@ -182,22 +183,62 @@ func checkpointWAL(
...
+			"WAL checkpoint could not fully complete "+
+				"(a reader is still holding an old snapshot); "+
+				"will retry next tick",
+			"wal_frames", walLog,
+			"checkpointed_frames", checkpointed,
+		)
 		return nil
 	}
+
+	busy, walLog, checkpointed, err = checkpoint(ctx, "TRUNCATE")
```
chrisguiney, line 229: The PASSIVE gate keeps TRUNCATE off the writer lock only when a reader pins frames PASSIVE cannot backfill. A reader whose snapshot is at the WAL end still holds a WAL read mark, so PASSIVE reports every frame checkpointed and this TRUNCATE takes the writer lock and waits out checkpointBusyTimeout. Reproduced at fce5f6f8 with the new test harness but the read transaction opened after the last write: the call ran 257ms and a busy_timeout(0) probe was refused 57904 times, as with the gate reverted (61164). Close this case, or narrow the claim to pinned frames and test the reader-at-tip case.

database/plugin/metadata/sqlite/checkpoint_test.go
```diff
@@ -190,3 +228,134 @@ func TestCheckpointWALDoesNotBlockWriteBehindReaderSnapshot(t *testing.T) {
...
+	)
+	writeRows(100_000, 2000)
+
+	// Probe the writer lock from a dedicated connection that never waits, so
+	// a single observation of SQLITE_BUSY means the checkpoint was holding
+	// it at that moment rather than that the probe was impatient.
+	probeDB, err := sqlstore.OpenDB(
+		"sqlite",
+		sqliteFileURI(filepath.Join(dataDir, "metadata.sqlite"))+
+			"?_pragma=busy_timeout(0)",
```
chrisguiney, line 305: This probe uses busy_timeout(0), but writeDB is opened with _txlock=immediate and busy_timeout(30000), so the import writer waits out this lock hold rather than failing. Driving writeDB itself in the same harness gave 0 errors and at most ~345ms per write, both at fce5f6f8 and with the gate reverted. The change therefore does not account for the SQLITE_BUSY in #4487, whose acceptance criterion (coverage for concurrent UTxO import and immutable-load offset writes) is unmet. Use a non-closing reference instead of Fixes #4487, or add a test reproducing SQLITE_BUSY on the writeDB path.

chrisguiney (changes requested):
Requesting changes. arepala-uml point 1 (the named tests pass against direct TRUNCATE) is addressed: the new writer-lock test and both checkpointWALWith tests fail with the gate reverted in place. Point 2 (no causal link to the SQLITE_BUSY in #4487) stands; see inline. Merged with current main including #4532: clean merge, go test -race passes for the sqlite and sqlstore packages, and the new test asserts a count rather than a duration. CodeRabbit skipped this head (incremental reviews disabled); no cubic review found.

Issue #4152, opened 2026-09-09
ledger: GetStakeDistribution omits 36 pools present in a real cardano-node's reply after Mithril bootstrap


arepala-uml:
## Summary

`GetStakeDistribution`/`GetPoolDistr2` silently omits any pool that holds
snapshot stake but that dingo cannot resolve a VRF key hash for
(`ledger/pool_stake_distribution.go:266-297`, the `continue` after a WARN
log). Against a real cardano-node at the same pinned point, 36 pools that
cardano-node reports are completely absent from dingo's answer rather than
present with a matching or even a wrong fraction.

## Evidence

Found via `cmd/node-parity check`/`watch` (blinklabs-io/dingo#1900): Preview,
Mithril-bootstrapped dingo (bootstrap landed at slot 122264150 / epoch 1415)
vs. a real cardano-node v11.0.1, both at live tip.

Confirmed reproducible across **4 independent comparison cycles spanning ~83
blocks and ~25 minutes of real chain progress** (block 4647006 through block
4647089) — the exact same 36 pool IDs are missing from dingo's answer in
every single cycle, byte-for-byte identical each time:

```
pool139mxqzdsxtww00d2rlpa0l27ysr64v5jwr8s2kjne06h64lsazj
pool145hrwnk803jen0eu00z2v62avw8a27xvpyqhp85s0xwtumzlxpr
pool148d4cdy22t64lgkdefhsh70rj5hjh85hdk3mv4skf96uzwyd8t6
pool14uhtey86p0efeh6ytvm6gungdgxquyrtf740s5n298902qm2e2u
pool15zu40d05gy43gyyxy7svumkxyac9qy6shlpdrgyz5gh8g48qaaj
pool18crpy079ekktl4397hl34tnr0apvzu6ajgle87cs0r0wkdrsmlj
pool18vfcklr9nt3q4ltyf52y3vqj5dst4cefgh6q2dy3eunk7vkt63u
pool1929tc9hdfuvpxj8mzcrn0tsg50w486xpsmn2jfde6qa87m4ez27
pool19vfxlduft0mdf4psvhhcl8l63m830wjxlrnq8826hgalvz9kf2t
pool1a5ztqh82xv8extrne9luj0jevdx9v7t0t4mruy5vm5mnjzkqt3k
pool1c7a0ntmyd9mxh8zhckg773vmk2snzk4axlhtwu7d6lxgq3yhrjl
[...]

Issue #3824, opened 2026-09-03
node-parity: GetStakeDistribution reports 2x the fraction cardano-node does for every pool


arepala-uml:
Running `TestLedgerStateConsensus` (blinklabs-io/dingo#1900, PR #3806's devnet
conformance scenario) against a 2-pool accelerated devnet caught Dingo's stake
distribution and a reference cardano-node's disagreeing by a uniform factor of
2, not by noise:

    stake distribution: pool pool10mhrquy8lccz6a7x3ezyht7upwt9zvllz3w5l8hm4mplyn54dyu
    fraction differs: 1/2 (dingo) vs 1/4 (cardano-node)
    stake distribution: pool pool15g0nsq60smuj3gg6mn8awda3gqf9tj2eqe6shzzspfdwvflqutk
    fraction differs: 1/2 (dingo) vs 1/4 (cardano-node)

Observed at slot 54 / block 21 on the accelerated conformance network. With
only two pools, Dingo's 1/2 + 1/2 accounts for the whole snapshot; cardano-node's
1/4 + 1/4 accounts for only half of it. Both pools are off by the same factor,
which points at the TotalActiveStake denominator each side uses, not at either
pool's numerator individually.

Dingo's side comes from `ledger.PoolStakeDistribution`
(`ledger/pool_stake_distribution.go:91-165`), which sources the total from
`ls.totalActiveStake(snapshotEpoch, true, metaTxn)` -- the same mark-snapshot
total leader election reads. #3814 documents a related but distinct case of
Dingo's forge and verify paths disagreeing on which active-stake denominator
to use; this may be a third accessor in the same family, may be specific to
GetStakeDistribution/GetPoolDistr2's NtC reply path, or may be a devnet
genesis/config difference between the two nodes rather than a Dingo bug at
all -- not yet root-caused.

Caveat: slot 54 / block 21 is likely still epoch 0 on this accelerated
network, before mark/set/go snapshot rotation has gone through a full cycle.
Reproduction should confirm whether the mismatch persists past the first
epoch boundary or is an epoch-0 bootstrap artifact.

Found via: the `node-parity` cross-node comparison tool built in PR #3806
(blinklabs-io/dingo#1900 phase 2), running
`./run-tests.sh --conformance --accelerated -run TestLedgerStateConsensus`
against a real dingo-producer + cardano-node pair.

Acceptance criteria: root-cause whether this is a Dingo denominator bug, a
[...]

PR #4513, opened 2026-09-19
fix(ledger): use empty genesis reward block counts

### LINKED ISSUE #4502: ledger: extra reward round at the 0->1 epoch boundary permanently offsets ADA pots and GetStakeDistribution
## Summary

Dingo applies one more reward/ADA-pot round than a reference `cardano-node`
does: it runs a round at the `0 -> 1` epoch boundary that the reference does
not. Reserves end up permanently low and treasury permanently high by that
round's value, and because `GetStakeDistribution`'s denominator is
`circulation = MaxLovelaceSupply - reserves`, every pool's reported fraction is
wrong for the life of the network.

`internal/test/devnet/scenarios/TestLedgerStateConsensus` catches this on every
`./run-tests.sh --conformance` run:

```
ledger state diverged between dingo-producer and cardano-producer at slot 718 (block 283):
stake distribution: pool pool1t283hj... fraction differs: 500000/2000597 (dingo) vs 1/4 (cardano-node)
stake distribution: pool pool1z4wk8d... fraction differs: 500000/2000597 (dingo) vs 1/4 (cardano-node)
```

This is the unresolved residual of #3824. That issue's fix corrected the
magnitude -- the fractions were 2x off (`1/2` vs `1/4`) and are now within
0.03% -- but not the exactness. No open issue tracks the remainder.

## Evidence

Both sides measured directly on a held conformance devnet: `cardano-node` via
`cardano-cli conway query ledger-state` (`/stateBefore/esChainAccountState`),
Dingo via its own `network_state` rows.

| point | cardano-node reserves / treasury | Dingo reserves / treasury |
|---|---|---|
| genesis, slot 0 | 2,000,000,000,000 / 0 | 2,000,000,000,000 / 0 |
| epoch 1, slot 786 | 2,000,000,000,000 / 0 | 1,998,920,000,000 / 1,080,000,000 |
| epoch 2, slot 1302 | 1,998,920,320,000 / 1,080,080,000 | 1,997,840,903,200 / 2,159,496,800 |

As a ledger of rounds actually applied:

- `cardano-node`: one round, at slot 1000 -- treasury `+1,080,080,000`
- Dingo: two rounds -- slot 500 `+1,080,000,000`, slot 1000 `+1,079,496,800`

Dingo's slot-1000 round closely matches cardano-node's slot-1000 round, so the
round at the `1 -> 2` boundary is right. The extra round at the `0 -> 1`
boundary is the defect.

The gap does not close. At epoch 2 the two nodes differ by **1,079,416,800
lovelace** in both treasury and reserves, and circulation differs by exactly the
same amount: 4,002,159,096,800 (Dingo) vs 4,001,079,680,000 (cardano-node).
The round count stays one apart, so it does not compound without bound, but it
never converges either.

At epoch 1 the extra round is a pure reserves -> treasury transfer: the reserves
drop equals the treasury gain exactly, so no pool or account rewards are
distributed. It reconciles arithmetically as `dR1 = eta * rho * reserves` with
`eta = 0.9` giving 5.4e9, and `tau * 5.4e9 = 1.08e9` -- the observed treasury.
The absolute magnitude is run-dependent because `eta` follows block luck (a
separate run showed 1,194,000,000, i.e. `eta` about 0.995); the reduced fraction
is stable within a run.

## Chain of causation

Extra round at `0 -> 1` -> reserves low by ~1.08e9 -> `circulation =
MaxLovelaceSupply - reserves` high by the same -> `GetStakeDistribution`
denominator high -> `500000/2000597` instead of `1/4`.

The denominator relationship is confirmed on the reference side too: when
cardano-node's reserves finally moved at epoch 2, its reported fraction changed
from `1/4` to `1562500/6251687`, which is exactly
`1e12 / (6e12 - 1,998,920,320,000)`.

## Code site

`stakeRewardEpochsForApplication` (`ledger/reward_calculation.go`) returns a
bootstrap round for `newEpoch == 1 || newEpoch == 2`. The measurement says the
`== 2` arm is correct and the `== 1` arm is the spurious one.

`applyStakeRewards`' Byron performance-epoch guard is not involved and behaves
as designed here: it keys on epoch 0's `EraId` being Byron, and this network's
epoch 0 is Conway (`cardano-cli query tip` reports era Conway at epoch 0; the
generated config sets `TestShelleyHardForkAtEpoch` through
`TestConwayHardForkAtEpoch` to 0).

## Please do not fix this by simply deleting the `newEpoch == 1` arm

It was added deliberately for #3381, where omitting the `0 -> 1` round left
preview's treasury at 0 and reserves at the genesis value, propagating into
every later epoch. The comment above `stakeRewardEpochsForApplication` also
states that cardano-ledger's NEWEPOCH rule applies monetary expansion and the
treasury tax at *every* boundary from the network's first Shelley-era epoch
onward, and cites verification against IntersectMBO/cardano-ledger.

That claim and this measurement disagree for a Conway-at-genesis network. A
plausible reconciliation, which I have **not** verified against cardano-ledger
source, is that NEWEPOCH applies a `RewardUpdate` only when one exists, and RUPD
cannot have produced one by the `0 -> 1` boundary -- so the reference has
nothing to apply there, while a network whose first rounds are suppressed for
another reason (#3381) still needs its pots seeded. Whoever picks this up should
settle that against the upstream source before changing the arm, since the two
issues may need different mechanisms rather than one flag.

## Secondary finding

Even the two rounds that *do* correspond are not identical: Dingo's slot-1000
round is `+1,079,496,800` where cardano-node's is `+1,080,080,000`, and the
nodes' reserves differ by 320,000 lovelace beyond the timing offset. That is a
separate, much smaller discrepancy in the round's inputs (fees and/or `eta`),
distinct from the extra-round bug, and worth tracking on its own.

## Reproduce

```
cd internal/test/devnet && ./run-tests.sh --conformance
```

`TestLedgerStateConsensus` fails on the first of its three samples, roughly 13
minutes. It reproduces on every run and is not flaky.

To observe the pots directly, bring the network up and hold it by running no
tests -- `./run-tests.sh --conformance --keep-up -run TestMatchesNothing`
(`--keep-up` only holds on a zero exit, so a failing test always tears down) --
then read `cardano-cli conway query ledger-state` from the cardano-producer
container and `SELECT * FROM network_state` from the dingo-producer's
`metadata.sqlite`.

## Note on coverage

The default all-dingo DevNet profile cannot catch this: both sides run Dingo's
own pot accounting and agree with each other. `TestLedgerStateConsensus` is
`//go:build linux && devnet && devnet_conformance`, so only the `--conformance`
mode compares against the reference.

Separately, `nodeparity.diffProtocolParamsAndStake` compares only
`StakeFraction` and the VRF hash, never absolute per-pool stake. Absolute stake
was confirmed to agree here out-of-band (`cardano-cli query stake-snapshot
--all-stake-pools` reports `stakeMark = stakeSet = 1,000,000,000,000` per pool),
but the test itself would not have reported a numerator divergence.

## Acceptance criteria

- Determine, against cardano-ledger source, whether the reference applies a
  reward round at the `0 -> 1` boundary, and reconcile that with #3381.
- Make Dingo's applied-round ledger match the reference's on a
  Conway-at-genesis network.
- `TestLedgerStateConsensus` passes under `./run-tests.sh --conformance`.
- Decide whether the residual 320,000-lovelace input discrepancy is in scope
  here or belongs in its own issue.
- Consider whether `nodeparity` should diff absolute per-pool stake alongside
  the fraction, so a numerator divergence cannot hide behind a matching ratio.

Related: #3824 (origin; magnitude fixed, exactness not), #3381 (added the
bootstrap rounds), #3814 (forge vs verify disagreeing on which active-stake
denominator to use), #4152.

wolf31o2:
Use genesis's empty previous block counts for the reward update applied at epoch 1. Conway-at-genesis leaves treasury and reserves unchanged; Preview's `d=1` still requires the first treasury transfer.

Add exact bootstrap pots and `GetStakeDistribution` regression coverage, including the fee refund at epoch 2. Extend `TestLedgerStateConsensus` to compare epochs 1 and 2, ADA pots, and absolute stake snapshots against cardano-node. Compile the devnet tests before starting nodes so cold builds cannot consume the bootstrap sampling window.

The live two-node bootstrap scenario still needs to pass before merge.

Fixes #4502. Related to #3381, #3824, and #4152.


wolf31o2:
@coderabbitai review

wcatz (approved):
Approving. I root-caused #4502, so I reviewed this against the measurements that
issue was filed on.

**The fix matches the root cause and takes the careful route.** Rather than
dropping the `newEpoch == 1` arm, it feeds that round genesis's empty previous
block counts, so `eta = 0` on a Conway-at-genesis network and no expansion
happens — while `d >= 0.8` still forces `eta = 1`, preserving the Preview
behaviour #3381 added the arm for. That resolves the two issues against each
other instead of trading one for the other.

The comment replacing the old "cardano-ledger applies expansion at every
boundary from the first Shelley-era epoch" claim is the right correction: that
claim is what my measurement contradicted, and the new wording (the first RUPD
reads genesis's empty `nesBprev`, not epoch 0's `nesBcur`) matches what a
reference `cardano-node` actually does on that network shape.

**Verified rather than assumed:**

- `rewardBlockCounts` has a second call site in
  `precomputedStakeRewardApplication`, which `applyStakeRewards` tries *first*,
  and only the `calculateStakeRewardApplication` one is guarded. That path
  returns early on `epochs.bootstrap`, which is true for exactly epochs 1 and 2,
  so bootstrap rounds always fall through to the patched path. The single guard
  is sufficient.
- `reward_bootstrap_test.go` shows `4+/6-`, which is where a weakened assertion
  would hide. Only the comment changed; the assertions are untouched and the
  comment now states the new expected behaviour.
- `TestApplyStakeRewardsConwayGenesisPerformance` asserts the *reference's*
  numbers, not Dingo's: epoch 1 at `treasury 0 / reserves 2000000000000 /
  fraction 1/4`, epoch 2 at `1080080000 / 1998920320000 / 1562500/6251687`.
  Those are exactly the values I read off `cardano-node` via
  `cardano-cli conway query ledger-state` while root-causing, and its derivation
  of `eta = 180/(500*0.4) = 0.9` independently matches the `eta` I measured from
  the observed pot movement. It fails pre-fix on the epoch-1 row.
- The extended `TestLedgerStateConsensus` now compares ADA pots and absolute
  stake snapshots across both nodes at epochs 1 and 2. That closes both coverage
  gaps I flagged in #4502 — the old diff compared only `StakeFraction` and the
  VRF hash, so a numerator divergence could have hidden behind a matching ratio.

**On CI:** `go-test (Windows)` was red when I picked this up, on
`TestCheckpointWALDoesNotBlockWriteBehindReaderSnapshot` — a test in
`database/plugin/metadata/sqlite`, which this PR does not touch, whose
assertions are three wall-clock `< 2s` bounds under `t.Parallel()` on the
slowest runner. I re-ran it rather than waving it through; it passed with no
code change. Whole board is green now. That test looks worth a separate
flake issue, but it is not this PR's problem.

Not blocking, just noting for whoever merges: the description says the live
two-node bootstrap scenario still needs to pass first, and that gate is outside
what this review covers.

PR #4514, opened 2026-09-19
fix(ledger): restart reward precompute after rollback

### LINKED ISSUE #4341: Pulse the stake reward calculation across the epoch instead of one pass

wolf31o2:
Restart background stake-reward precomputation after a successful rollback using the surviving epoch and applied tip. Clear abandoned pending work and reject prefilter retries from an invalidated reward-input generation.

Regression coverage includes same-epoch and cross-epoch rollback, failed state reloads, no-op rollback, stale retries, and reuse of persisted replacement outputs for protocol versions 6 and 7. Reward arithmetic and storage schemas are unchanged.

This is a bounded rollback-recovery fix. Calculation still runs in one pass; durable per-block cursor/accumulator pulsing remains unimplemented.

Fixes #4341


wolf31o2:
@coderabbitai review

PR #4515, opened 2026-09-19
fix(database): index batched UTxO deletion

### LINKED ISSUE #4067: Consumed-UTxO cleanup starves block application: 400 OR-terms in queryUtxoStakeRefs defeat tx_id_output_idx, scanning all 11M live UTxOs per batch

wolf31o2:
Fixes #4067.

- Resolve exact UTxO rows through indexed `tx_id` batches before updating primary keys.
- Preserve exact `(tx_id, output_idx)` semantics and the live-row guard in the existing write transaction.

---
## Summary by cubic
Fixes #4067 by changing `MarkUtxosDeletedAtSlot` to resolve UTxO primary keys through `tx_id IN` batches before updating, so SQLite uses `tx_id_output_idx` instead of falling back to a deleted-slot index on larger batches. Exact `(tx_id, output_idx)` matching and the `deleted_slot = 0` guard stay in the existing write transaction. Adds tests that pin the indexed query plan and verify only requested live rows are updated; the lookup rows are closed after each batch.

<sup>Written for commit cc8e4eed617c7b1a5499c8a26f1b4fdfbe393bff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4515?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


database/plugin/metadata/sqlstore/utxo.go
```diff
@@ -343,16 +343,26 @@ func (s *Store) MarkUtxosDeletedAtSlot(
...
+				args[0] = slot
+				placeholders := make([]string, end-start)
+				for i, rowID := range rowIDs[start:end] {
+					placeholders[i] = "?"
+					args[i+1] = rowID
+				}
 				if _, err := db.ExecContext(
 					ctx,
 					s.dialect.Rebind(
 						"UPDATE utxo SET deleted_slot = ? "+
```
arepala-uml: This UPDATE plans as SEARCH utxo USING INDEX idx_utxo_deleted_payment_script (deleted_slot=?) at n=2, 64 and 998 without sqlite_stat1, evaluating id IN (...) per live row; after ANALYZE it uses INTEGER PRIMARY KEY, but utxo.go:1446 requires a plan that holds whether or not ANALYZE has run. Projecting deleted_slot in utxoRowIDsByTxIDQuery, skipping rows where it is NULL or non-zero during the scan, and dropping this guard gives INTEGER PRIMARY KEY (rowid=?) at every batch size with no statistics; the shared write transaction already makes the guard redundant.

database/plugin/metadata/sqlstore/utxo_mark_deleted_lookup_test.go
```diff
@@ -0,0 +1,141 @@
...
+				legacyArgs...,
+			)
+			t.Logf("legacy MarkUtxosDeletedAtSlot plan (n=%d): %s", nTerms, legacyPlan)
+
+			txIDs, _ := distinctUtxoTxIDs(ids)
+			query, args := utxoRowIDsByTxIDQuery(txIDs)
+			newPlan := queryPlan(t, store.writeDB, query, args...)
+			t.Logf("tx_id-IN mark lookup plan (n=%d): %s", nTerms, newPlan)
+
+			require.Contains(t, newPlan, "tx_id_output_idx")
```
arepala-uml, line 71: Reverting MarkUtxosDeletedAtSlot to the previous chunked OR-of-pairs UPDATE leaves both tests in this file passing: this one plans utxoRowIDsByTxIDQuery directly and never calls the method, and TestMarkUtxosDeletedAtSlotUpdatesOnlyRequestedLiveRows asserts behaviour the previous statement also satisfied. Nothing plans the UPDATE the method actually executes, which is why the deleted_slot fallback above goes uncaught.

arepala-uml (changes requested):
The lookup half is indexed correctly, but the UPDATE this now issues keeps the deleted_slot = 0 guard and plans as SEARCH utxo USING INDEX idx_utxo_deleted_payment_script (deleted_slot=?) at n=2, 64 and 998 on a 50k-row store with no sqlite_stat1, visiting 49484 of 50000 rows. account.go:415 records that ANALYZE runs only at Mithril sync and before backfill, so a genesis-synced node is in exactly that position. Issue #4067's scan is relocated to the guard rather than removed, and neither new test fails when the change is reverted.

database/plugin/metadata/sqlstore/utxo.go
wolf31o2: Confirmed. Without sqlite_stat1 the guarded update plans as SEARCH utxo USING INDEX idx_utxo_deleted_payment_script (deleted_slot=?) at 2, 4, 64, 400 and 998 row IDs, and as INTEGER PRIMARY KEY (rowid=?) at every size without the guard; after ANALYZE both are INTEGER PRIMARY KEY.

Fixed as suggested in 17917185: utxoRowIDsByTxIDQuery projects deleted_slot, queryUtxoRowIDs drops rows whose deleted_slot is NULL or non-zero, and markUtxosDeletedQuery carries no liveness predicate. The single-row MarkUtxoDeleted update keeps its guard and still plans on tx_id_output_idx, so it is unaffected. DATABASE.md now shows both statements.

database/plugin/metadata/sqlstore/utxo_mark_deleted_lookup_test.go
wolf31o2: Confirmed: nothing planned the statement the method executes. 17917185 extracts markUtxosDeletedQuery as the method's only UPDATE and adds TestMarkUtxosDeletedAtSlotUpdatePlansOnPrimaryKey, which plans it at 1, 2, 4, 64, 400 and 998 row IDs on a statistics-free database and requires INTEGER PRIMARY KEY, with the guarded form as a control. Reinstating the guard inside that helper fails it at every size from two upwards.

DATABASE.md
```diff
@@ -2427,6 +2417,43 @@ mirroring `UtxosDeleteConsumed`'s own pruning threshold exactly) with
...
+SELECT id, tx_id, output_idx
+FROM utxo
+WHERE tx_id IN (decode($1, 'hex'), decode($2, 'hex'));
+
+-- After exact-pair filtering in Go:
+UPDATE utxo
+SET deleted_slot = $1
+WHERE deleted_slot = 0
+  AND id IN ($2, $3);
+```
```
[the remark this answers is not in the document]
wolf31o2: Valid: the section carried the superseded guarded form alongside the current unguarded primary-key update, unlabelled. Removed in 79f5073b; only the lookup and the primary-key update remain.

database/plugin/metadata/sqlstore/utxo.go
```diff
@@ -1160,6 +1163,126 @@ func utxoIDPredicate(ids []models.UtxoId) (string, []any) {
...
+// of its own.
+// markUtxosDeletedQuery builds the second half of MarkUtxosDeletedAtSlot's
+// two-step update: a primary-key update of the row IDs queryUtxoRowIDs
+// already established are live.
+//
+// It carries no "deleted_slot = 0" predicate on purpose. With one, SQLite
+// plans the statement as SEARCH utxo USING INDEX
+// idx_utxo_deleted_payment_script (deleted_slot=?) from two terms upwards
+// whenever sqlite_stat1 is absent -- every live row visited and "id IN (...)"
+// evaluated per row, which is the whole-table pass issue #4067 reports, moved
```
arepala-uml, line 1203: Being in one write transaction is what makes this safe on SQLite, where writeDB is SetMaxOpenConns(1). It is not what makes it safe on Postgres or MySQL: BeginOptions(false) returns nil, so READ COMMITTED over a shared pool lets another transaction commit deleted_slot between the lookup and this primary-key update, which then overwrites it. live_stake.go:185-196 already states the invariant that does hold, the ledger's single apply loop. Name that here and in DATABASE.md instead, and note that MarkUtxosDeletedAtSlot has a second caller off that loop, reconcile.go:247 via import.go:657.

database/plugin/metadata/sqlstore/utxo.go
```diff
@@ -1160,6 +1163,126 @@ func utxoIDPredicate(ids []models.UtxoId) (string, []any) {
...
+// from the lookup to the update. Populated statistics change the plan, but a
+// long-running node's utxo statistics are stale or absent, so the plan has to
+// hold without them. Liveness is filtered in queryUtxoRowIDs instead, inside
```
arepala-uml, line 1206: This comment runs straight into the one above with no separating //, so Go attaches the whole block to markUtxosDeletedQuery and queryUtxoRowIDs is left with no doc comment at all. Both identifiers are unexported, so gofmt and golangci-lint are silent on it; a blank // line is the fix. Separately, the interface contract at database/plugin/metadata/store.go:840-844 still states that the SQL filter is deleted_slot == 0, which this change made false by moving liveness filtering into Go in queryUtxoRowIDs.

database/plugin/metadata/sqlstore/utxo.go
```diff
@@ -343,17 +343,20 @@ func (s *Store) MarkUtxosDeletedAtSlot(
 			if err != nil {
 				return err
 			}
```
arepala-uml, line 345: queryUtxoStakeRefs above and queryUtxoRowIDs here each run SELECT over utxo WHERE tx_id IN (...) across the identical distinct tx_id batches, so the same rows are fetched twice per call: one projects tx_id, output_idx, deleted_slot, credential_tag, staking_key, the other id, tx_id, output_idx, deleted_slot, and both are non-covering searches on tx_id_output_idx. One fused lookup projecting both sets would halve the fetches. Outside the scope of issue 4067, noted because this change created the second pass.

arepala-uml (changes requested):
Prior blocker 1 is cleared and causally proven: reinstating the guard inside markUtxosDeletedQuery fails TestMarkUtxosDeletedAtSlotUpdatePlansOnPrimaryKey at 1, 2, 4, 64, 400 and 998 row IDs with SEARCH utxo USING INDEX idx_utxo_deleted_payment_script. Blocker 2 is cleared for the defect it named. Verified on a merged tree with main f00f2492, which merges clean: build, vet, gofmt, scoped lint, the sqlstore package, docs-parity and import-boundaries all pass, and the branch's green checks predate that base by 43 commits. No behavioural defect found, but the stated safety rationale is wrong.

database/plugin/metadata/sqlstore/utxo.go
arepala-uml: Cleared, causally. markUtxosDeletedQuery carries no liveness predicate and plans as INTEGER PRIMARY KEY at 1, 2, 4, 64, 400 and 998 row IDs with no sqlite_stat1, while reinstating the guard inside the helper fails the new test at every one of those sizes with SEARCH utxo USING INDEX idx_utxo_deleted_payment_script. Liveness filters in queryUtxoRowIDs instead, which stays on tx_id_output_idx. The rationale given for why that is safe is wrong, raised separately at utxo.go:1203.

database/plugin/metadata/sqlstore/utxo_mark_deleted_lookup_test.go
arepala-uml: Cleared for the defect it named: TestMarkUtxosDeletedAtSlotUpdatePlansOnPrimaryKey plans the statement the method actually executes, with the guarded form as a control. Reverting the whole method body to the legacy chunked OR-of-pairs does still leave all three tests passing, but that revert is caught by golangci-lint unused on three orphaned helpers rather than passing silently, so the gap is narrower than it was.

DATABASE.md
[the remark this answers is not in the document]
arepala-uml: Addressed at head. 79f5073b removed the superseded guarded example, and DATABASE.md now carries only the lookup and the unguarded primary-key update at line 2493, with the rationale for the absent predicate at 2480. Verified at 79f5073b.

PR #4516, opened 2026-09-19
fix(database): repair indexes after restore

### LINKED ISSUE #4065: Restored metadata.sqlite can be missing declared indexes; absent idx_utxo_transaction_id makes rollback cascades unbounded and wedges the node
## Summary

A `database restore` can produce a `metadata.sqlite` that is missing indexes the sqlite migrations declare. On a mainnet core node this left **220 of 234 declared indexes** present. One of the missing ones, `idx_utxo_transaction_id`, makes any rollback that deletes transactions effectively unbounded, and wedges the node.

## Impact

`utxo` has an `ON DELETE CASCADE` foreign key to `"transaction"`:

```sql
CONSTRAINT `fk_transaction_outputs` FOREIGN KEY (`transaction_id`)
  REFERENCES `transaction`(`id`) ON DELETE CASCADE
```

`_pragma=foreign_keys(1)` is enabled (`database/plugin/metadata/sqlite/shared_sqlstore.go:51`), so every parent row deleted by `DELETE FROM "transaction" WHERE slot > ?` must locate its cascade children in `utxo`. Without `idx_utxo_transaction_id` that is a full table scan **per deleted row**:

```
sqlite> EXPLAIN QUERY PLAN SELECT id FROM utxo WHERE transaction_id = 12345;
`--SCAN utxo                                          -- missing index

sqlite> EXPLAIN QUERY PLAN SELECT id FROM utxo WHERE collateral_return_for_tx_id = 12345;
`--SEARCH utxo USING COVERING INDEX idx_utxo_collateral_return_for_tx_id (collateral_return_for_tx_id=?)
```

Notably the other four `utxo` FK columns *were* indexed; only `transaction_id` was absent.

On the affected node: 2,547 rows to delete against ~11.0M `utxo` rows ≈ 2.8×10^10 row visits, which profiled at roughly **3 hours** for a single rollback.

## What it looks like in the wild (the misleading part)

The visible symptom is not a slow query, it is an EventBus stall:

```
level=WARN msg="event delivery stalled: subscriber not draining"
  type=ledger.blockfetch buffer=8 stalled_for=30s blocked_publishers=2
```

The slow rollback blocks the ledger, so the `ledger.blockfetch` subscriber stops draining, its buffer (8) fills, and publishers block. The node reports `active` to systemd, `NRestarts=0`, all goroutines asleep, zero disk I/O — it looks like a deadlock in the event bus rather than a slow `DELETE`. I spent a while chasing the wrong layer because of this.

**Restarting does not help, and makes it look permanent.** An interrupted rollback persists `ledger_tip > chain_tip`, so on next start `reconcilePrimaryChainTipWithLedgerTip` (`ledger/state.go:8042`) redoes the same slow rollback:

```
level=WARN msg="ledger tip ahead of primary chain tip at startup, rolling back metadata to chain tip"
  chain_tip_slot=196833572 ledger_tip_slot=196845785
```

Each restart re-enters it, so the node never finishes starting.

## Diagnosis path that worked

`serve --debug-port <port>` then:

```
curl -s localhost:<port>/debug/pprof/goroutine?debug=2
```

which showed the real stack:

```
ledger.(*LedgerState).Start
 → reconcilePrimaryChainTipWithLedgerTip        ledger/state.go:8042
   → rollback                                   ledger/state.go:3002
     → SubmitAsyncDBTxn → SubmitAsyncDBOperation ledger/state.go:263
...worker:
 database.TruncateAfterSlot                     database/truncate.go:268
  → TransactionsDeleteRolledback                database/transaction.go:1898
    → sqlstore.DeleteTransactionsAfterSlot      transaction_write.go:665
      → _sqlite3VdbeExec → _sqlite3BtreeNext
```

## Fix applied

```sql
CREATE INDEX IF NOT EXISTS `idx_utxo_transaction_id` ON `utxo`(`transaction_id`);
```

Built in **3 seconds**. Measured before → after:

| | before | after |
|---|---|---|
| sync rate | 0 (frozen) | 138 slots/sec |
| read syscalls/sec | ~316,000 | ~4,421 |
| `event delivery stalled` occurrences | 67 | 0 |
| `chain extended` per 3 min | 0 | 180 |

`PRAGMA quick_check` → `ok`.

Worth noting: sync had been running at only **16.75 slots/sec before** the wedge and **138 slots/sec after** the index was added. So the missing index was throttling normal block application roughly 8×, not just rollback — this is a steady-state performance bug as well as an availability one.

## Suggested remedies

1. Have snapshot restore create every index the migrations declare, rather than leaving it to have happened.
2. On open, verify declared indexes are present and create any that are missing (they are all `IF NOT EXISTS`, so this is cheap and idempotent). A missing index is silent today.
3. Consider a guard/regression test asserting FK child columns are indexed — an unindexed `ON DELETE CASCADE` FK is a latent version of this bug for any future table.

## Full list of missing indexes observed

```
idx_asset_amount                     idx_key_witness_type
idx_asset_fingerprint                idx_redeemer_index
idx_asset_name_hex                   idx_redeemer_tag
idx_certs_block_hash                 idx_utxo_transaction_id
idx_certs_certificate_id             idx_witness_scripts_script_hash
idx_certs_cert_type                  idx_witness_scripts_type
idx_committee_member_cold_cred_hash  idx_datum_added_slot
```

Audit used:

```bash
sqlite3 "file:$DB?mode=ro" \
  'SELECT name FROM sqlite_master WHERE type="index" AND name NOT LIKE "sqlite_%";' | sort -u
# compare against
grep -rhoE 'CREATE (UNIQUE )?INDEX IF NOT EXISTS .[a-zA-Z_]+.' \
  database/plugin/metadata/sqlstore/migrations/v*/sqlite/*.sql
```

wolf31o2:
Fixes #4065.

- Rebuild the deferred metadata-index manifest while validating restored stores.
- Repair the complete manifest at startup when no bulk-load cycle is pending.

---
## Summary by cubic
Fixes #4065. Restored snapshots could permanently lose deferred metadata indexes because their migrations were already recorded complete, so the indexes never re-built. Restore validation and startup repair now rebuild the full deferred-index manifest whenever no bulk-load cycle is pending.

- During an active bulk-load cycle, repair still limits itself to the API/rollback-critical subset and preserves the pending marker.
- DATABASE.md documents the new restore-time repair and the startup behavior.

<sup>Written for commit 14c7a996c314a44339bb351679a7150c77a71daa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4516?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


internal/node/load.go
```diff
@@ -518,18 +518,24 @@ func RepairCriticalDeferredIndexes(
...
+		return ensureCriticalDeferredIndexes(manager, logger)
+	}
+	// A clear marker means no bulk-load cycle is active. Restore copies can
+	// still be missing any manifest entry because their recorded migrations do
+	// not re-run, so finish the whole manifest before accepting traffic.
+	if err := ensureCriticalDeferredIndexes(manager, logger); err != nil {
+		return err
 	}
-	return ensureCriticalDeferredIndexes(manager, logger)
+	return manager.BuildDeferredIndexes()
```
arepala-uml: ensureCriticalDeferredIndexes names the missing entries and reports a duration, then this call rebuilds the 11 non-critical manifest entries with no log line and no timing. Dropping idx_asset_fingerprint and running this path rebuilds it and produces an empty log. TestEnsureCriticalDeferredIndexesNamesMissingBeforeBuilding exists because a multi-million-row rebuild is silent while it runs; a MissingDeferredIndexes sibling to MissingCriticalDeferredIndexes would give the full build the same announcement and duration. Same at line 558.

database/lifecycle/restore.go
```diff
@@ -1151,7 +1151,11 @@ func validateRestoredDatabase(
 	defer host.StopCapability( //nolint:errcheck
 		context.WithoutCancel(ctx), plugin.CapabilityStorageMetadata,
 	)
-
+	if manager, ok := metadataStore.(metadata.DeferredIndexManager); ok {
+		if err := manager.BuildDeferredIndexes(); err != nil {
```
arepala-uml: validateRestoredDatabase takes no logger, so this rebuild cannot report which indexes it builds or how long it took, and withWriteTransaction uses context.Background(), so the caller's ctx cannot interrupt it. The comments at lines 204-225 make interruption safety an explicit property of this path, yet a restore of a bulk-loaded snapshot now ignores cancellation for the length of a full index build.

arepala-uml (changes requested):
The repair logic and the restore-time placement before the atomic rename are correct, and reverting load.go and restore.go fails both lazy-index tests and TestRestoreRebuildsDeferredIndexes. The full-manifest rebuild this adds emits nothing while it runs: dropping idx_asset_fingerprint and calling RepairDeferredIndexes rebuilds it and writes an empty log. 11 of the 24 manifest entries sit outside the critical subset, so a restored mainnet database stalls startup on the asset table with no output.

internal/node/load.go
wolf31o2: Confirmed. 3ec993cd adds metadata.MissingDeferredIndexLister and ensureAllDeferredIndexes, which names the missing entries before the build and reports the elapsed time after it, on both full-manifest paths here and on the pending branch below. DeferredIndexRebuilder.BuildAll, the third full-manifest caller, reports the same way now.

TestRepairDeferredIndexesAnnouncesLazyRebuild drops a lazy index and requires it named in the log; it fails with the announcement reverted in place.

database/lifecycle/restore.go
wolf31o2: The cancellation point is right and is fixed in 3ec993cd: metadata.ContextDeferredIndexBuilder threads the caller's context through to the DDL, and restore uses it where the store implements it. TestBuildDeferredIndexesContextHonoursCancellation shows the real store returning context.Canceled and leaving the recovery marker set; reverting the dispatch fails it.

On reporting: database/lifecycle logs nothing at all today, so rather than a package-wide logger, RestoreStorageConfig gains an optional Logger and the repair names the missing entries and reports its duration through it. The offline lifecycle service passes its own; snapshot verification restores into a directory it then deletes and leaves it unset.

The interruption-safety contract itself was not broken: the rebuild runs against the staging directory, so targetDataDir is untouched whichever path runs. What was missing is a timely response to cancellation, which the context threading restores.

database/plugin/metadata/sqlstore/deferred_indexes.go
```diff
@@ -150,6 +211,9 @@ func (s *Store) buildDeferredIndexes(
 						err,
 					)
 				}
+				if after != nil {
+					after(index.Name, time.Since(started))
```
chrisguiney, line 215: after fires once the existence check returns, before CREATE INDEX, and before fires for every entry, so restore logs 'deferred metadata index ready' for a missing index before building it and logs 'building' for indexes that already exist. Against a migrated SQLite store, BuildDeferredIndexesContextWithProgress made 48 callbacks with nothing to build. With idx_asset_fingerprint dropped it reported building, ready after 37us, then ready after 99us. Call before and after only when exists is false, and drop this first after call.

database/lifecycle/restore.go
```diff
@@ -1113,6 +1121,79 @@ func restoreBlobStore(
...
+				"rebuilding missing deferred metadata indexes in the "+
+					"restored database",
+				"indexes", strings.Join(missing, ","),
+				"count", len(missing),
+			)
+		}
+	}
+	start := time.Now()
+	var err error
+	if builder, ok := manager.(metadata.DeferredIndexProgressBuilder); ok {
```
chrisguiney, line 1159: No test reaches this branch. The lifecycle fakes implement only ContextDeferredIndexBuilder, and changing this condition to ok && false leaves the lifecycle and sqlstore suites green. Add a sqlstore test that drops one manifest index and requires before and after to fire exactly once for it, with after coming after the index exists, and neither firing for entries that were already present.

database/plugin/metadata/deferred_indexes.go
```diff
@@ -60,6 +65,38 @@ type DeferredIndexManager interface {
 //
 // Stores that do not implement this report nothing and fall back to logging
 // after the fact.
+// MissingDeferredIndexLister is MissingCriticalDeferredIndexLister for the
+// complete manifest. The full-manifest repair a restored database needs is as
+// silent as the critical one, and has more entries to build.
+type MissingDeferredIndexLister interface {
```
chrisguiney, line 71: The new types were inserted between the MissingCriticalDeferredIndexLister doc comment and its declaration. go doc now shows no doc for MissingCriticalDeferredIndexLister and attaches its text to MissingDeferredIndexLister. restore.go has the same problem: rebuildRestoredDeferredIndexes sits under the validateRestoredDatabase comment. Move each new declaration so every comment sits directly above its own identifier, and document BuildDeferredIndexesContextWithProgress on DeferredIndexProgressBuilder.

chrisguiney (changes requested):
Earlier review at 14c7a996: full-manifest naming and timing at startup is addressed (removing the ensureAllDeferredIndexes calls fails three internal/node tests), and restore now threads ctx and a logger. Outstanding: the per-index progress added in 55029f36 reports entries ready before they are built and has no test, and two doc comments are detached. With current main merged (4 commits, including #4617) the touched packages build and pass under -race.

arepala-uml (approved):

arepala-uml (dismissed):

PR #4517, opened 2026-09-19
fix(ledger): validate empty genesis committees

### LINKED ISSUE #4054: Conway genesis with an empty committee leaves committee validation unenforced
Split out of the review discussion on #4033, which seeds the Conway genesis
Constitutional Committee. That PR does not change this behavior, and the fix is
a ledger-semantics decision rather than plumbing, so it is filed separately
rather than folded in.

## Mechanism

`LedgerView.CommitteeStateAvailable` reports availability purely by row count
(`ledger/view.go:844`):

```go
members, err := lv.ls.db.GetCommitteeMembersIncludeDeleted(lv.txn)
...
return len(members) > 0, nil
```

`ensureGenesisCommittee` returns early when the Conway genesis declares no
committee members (`ledger/chainsync.go:5130`), so such a chain has no
`committee_member` rows at all. `CommitteeStateAvailable` then reports `false`,
and committee certificate validation and committee voter validation skip their
unknown-member checks — they cannot distinguish "this chain has no committee"
from "committee state is unavailable, so do not enforce".

## Why this is not simply a missing row

The empty case was reasoned about deliberately for a *different* scenario. The
comment at `ledger/view.go:835-839` explains that an authoritatively empty
committee following a `NoConfidence` enactment still reports available, because
that path leaves soft-deleted rows behind and `GetCommitteeMembersIncludeDeleted`
counts them.

A genesis that declares no committee is a genuinely distinct state: there are no
rows to soft-delete, so it is indistinguishable from an unavailable store. That
is what makes this a semantics question rather than a query fix.

## What needs deciding

Whether a Conway genesis with no committee members means:

1. **No committee is seated** — availability should be `true` with an empty
   seated set, so certificate and voter validation enforce against a committee
   of zero; or
2. **There is no committee state** — the current behavior, where validation
   declines to enforce.

Whichever is chosen, availability should stop being inferred from a row count
alone. Options include persisting an explicit empty-state marker at genesis
seeding, or deriving availability from the Conway genesis configuration rather
than from stored rows.

## Exposure

Narrow. Every configuration shipped in `config/cardano/` has a non-empty Conway
genesis committee — mainnet 7, preprod 7, preview 1, devnet 7, musashi 3 — so
this is reachable only on a custom network configured with an empty committee.
No shipped network is affected.

## Suggested scope

- `LedgerView.CommitteeStateAvailable` (`ledger/view.go:831-845`)
- committee certificate validation (`AuthCommitteeHot` / `ResignCommitteeCold`)
- committee voter validation
- regression coverage for an empty Conway genesis committee, which no test
  currently exercises
- a conformance vector, since this changes whether votes and certificates are
  enforced

Raised by CodeRabbit in review of #4033.

wolf31o2:
Treat an explicitly empty Conway genesis committee as authoritative without adding a sentinel database row. Preserve unavailable-state handling and fail-closed storage errors.

Add coverage for key/script certificate and voter rejection, proposed-member eligibility, initialization replay, storage faults, and a reference-derived GOVCERT resignation case. Update committee-state documentation.

Fixes #4054


wolf31o2:
@coderabbitai review

PR #4523, opened 2026-09-19
fix(tokenregistry): bound atomic snapshot apply

### LINKED ISSUE #4506: fix(tokenregistry): bound and atomically apply snapshots
## Summary

Token-registry synchronization limits the compressed response and individual mappings, but does not bound aggregate decompressed bytes, archive entry count, accepted mapping count, retained batch bytes, or database work.

Snapshot writes also use autocommit batches, so a later failure can leave partially applied rows while the snapshot ETag and identity remain unchanged.

## Acceptance criteria

- Bound total decompressed bytes, entry count, accepted rows, database operations, and retained batch bytes.
- Count all archive content, including non-mapping entries, toward the decompression limit.
- Reject an exhausted snapshot without leaving partially applied rows.
- Apply the snapshot stamp, row upserts, pruning, ETag, and source identity atomically.
- Preserve normal registry snapshots, source and logo-mode switching, `304 Not Modified`, and supported SQL backends.
- Add regression tests for expansion limits, excessive tiny mappings, bounded batch memory, rollback, and state non-advancement.

## Relevant code

- [`internal/offchainmetadata/tokenregistry_sync.go`](https://github.com/blinklabs-io/dingo/blob/f85a42886c6af8eb044ae79614dd66185756a701/internal/offchainmetadata/tokenregistry_sync.go#L758-L860)
- [`database/plugin/metadata/sqlstore/token_registry.go`](https://github.com/blinklabs-io/dingo/blob/f85a42886c6af8eb044ae79614dd66185756a701/database/plugin/metadata/sqlstore/token_registry.go#L68-L96)

wolf31o2:
## Summary
- bound compressed and expanded snapshot data, archive entries, mappings, and write batches
- validate and stage snapshots before opening the database transaction
- expose limits through configuration and reject invalid values

## Testing
- go test -race ./internal/offchainmetadata ./internal/config ./internal/node ./database/plugin/metadata/sqlite .
- go vet ./internal/offchainmetadata ./internal/config ./internal/node ./database/plugin/metadata/sqlite .
- make docs-parity

Closes #4506

---
## Summary by cubic
Closes #4506 by making token registry snapshot application atomic and resource-bounded. Previously the sync committed parsed batches as it streamed and wrote stamp, ETag, and identity separately, so a failed snapshot could leave partial rows or half-recorded state visible; now it validates and stages the archive to a temporary file first, then applies rows, prune, ETag, and source identity in one metadata transaction. API readers only see complete snapshots, and download/parse no longer hold the backend writer lock.

**Limits**
- Adds `tokenRegistry.maxDecompressedBytes`, `maxArchiveEntries`, `maxAcceptedEntries`, and `maxBatchBytes` with env and CLI equivalents; zero uses defaults and negative values are rejected.
- All tar content counts against the decompressed/archive limits, not just mapping files.
- Staged entries use an explicit JSON format with tagged fields.
- Any limit, store, state, or commit failure rolls back the transaction and removes the staging file; cleanup failures after commit are surfaced without hiding a prior error.

<sup>Written for commit 5ed45bc8825b2869ae3703cf94a500a916c8a129. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4523?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


internal/offchainmetadata/tokenregistry_sync.go
```diff
@@ -894,42 +962,159 @@ func (s *TokenRegistrySync) applySnapshot(
...
-			}
+		acceptedEntries++
+		if acceptedEntries > s.maxAcceptedEntries {
+			return nil, fmt.Errorf(
+				"token registry snapshot exceeds %d accepted mappings",
+				s.maxAcceptedEntries,
+			)
+		}
+		entryBytes := tokenRegistryEntryRetainedBytes(entry)
+		if entryBytes > s.maxBatchBytes {
```
chrisguiney, line 976: One mapping that passes maxEntryBytes but retains more than maxBatchBytes rejects the whole snapshot, while a mapping over maxEntryBytes is only skipped. validate.go accepts maxBatchBytes below maxEntryBytes, so an operator who lowers maxBatchBytes lets a single large-logo mapping block every sync. Reproduced: MaxBatchBytes=200, MaxEntryBytes=1MiB, a 300-byte name gives written=0 and the valid sibling mapping is not written. Skip and count it like an oversize mapping (keeps prune deferral), or reject effective maxBatchBytes below effective maxEntryBytes after defaults resolve.

internal/offchainmetadata/tokenregistry_sync_test.go
```diff
@@ -482,6 +579,146 @@ func TestTokenRegistrySyncRejectsOversizedBody(t *testing.T) {
...
+	})
+	server := newRegistryServer(t, body)
+	store := newFakeTokenRegistryStore()
+	sync := newTestSync(t, store, server.URL, func(c *TokenRegistryConfig) {
+		c.MaxDecompressedBytes = 4 << 10
+	})
+
+	_, err := sync.SyncOnce(t.Context())
+
+	require.ErrorContains(t, err, "decompressed")
```
chrisguiney, line 597: This test does not pin that decompression stops at the limit. Replacing limitReaderPast(gzipReader, s.maxDecompressedBytes) with the bare gzip reader keeps the whole package green, because the post-loop decompressed.read check still rejects after the full stream is expanded. Use an incompressible non-mapping file larger than the 1 MiB drain (for example 4 MiB of random bytes, limit 4 KiB) and assert the response body was not fully consumed, so a gzip bomb cannot be expanded in full before rejection.

internal/offchainmetadata/tokenregistry_sync.go
```diff
@@ -558,66 +582,94 @@ func (s *TokenRegistrySync) SyncOnce(ctx context.Context) (int, error) {
...
-		// unlike the tag and identity this failure aborts before any row is
-		// written rather than proceeding with a stamp nothing recorded.
-		return 0, fmt.Errorf("record token registry sync stamp: %w", err)
-	}
-	written, mappings, skipped, err := s.applySnapshot(
-		ctx,
-		resp.Body,
-		syncedAt,
-	)
+	stage, err := s.stageSnapshot(ctx, resp.Body)
```
chrisguiney, line 585: The comment above now sits over stageSnapshot but describes the stamp, which is taken at nextSyncStamp after staging. Move it to the syncedAt line or drop the parts nextSyncStamp already documents.

chrisguiney (changes requested):
Staging-before-transaction and single-transaction apply hold up: race tests for offchainmetadata, config, node and metadata/sqlite pass at head and with current main merged, and in-place removal of the archive, accepted-entry and batch-flush bounds each fails its new test. Two actionable gaps inline. The CodeRabbit finding on validate.go (maxBatchBytes below maxEntryBytes) is confirmed and is the first inline comment.

PR #4533, opened 2026-09-20
perf(ledgerstate): defer reward stake rebuild during import

wolf31o2:
## Summary
- defer per-batch reward live-stake aggregate refresh during Mithril UTxO import
- rebuild the aggregate once after ledger state import completes

## Validation
- `go test ./database/plugin/metadata/sqlstore ./ledgerstate`
- Linux amd64 build
- same-Preview-snapshot A/B: 42m35s candidate vs 54m26s v0.70.15 control (21.8% faster); peak RSS 17.0 GiB vs 20.8 GiB

Signed-off-by: Chris Gianelloni <wolf31o2@blinklabs.io>

---
## Summary by cubic
Ledger-state UTxO imports now defer reward live-stake aggregate refreshes until all UTxOs are loaded, replacing the previous per-batch rebuild. The final rebuild preserves aggregate correctness while reducing import time and peak memory usage.

**Validation**

- Same Preview snapshot import completed 21.8% faster, with peak RSS reduced from 20.8 GiB to 17.0 GiB.
- Tests verify that deferred imports leave the aggregate untouched until the rebuild and produce the correct result.
- Existing `AddUtxos` and `ImportUtxos` callers retain their immediate-refresh behavior.

<sup>Written for commit d96c5904f7db48fe5854dc9377cfe7de90c6b5bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4533?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Bulk UTxO imports can now defer reward live stake aggregate refreshes until rebuilding is requested.
  - Existing import flows continue to refresh reward live stake data immediately.

- **Bug Fixes**
  - Improved reward live stake accuracy after deferred UTxO imports, including correct stake totals once the aggregate is rebuilt.

- **Tests**
  - Added coverage verifying deferred refresh behavior and resulting stake values.


ledgerstate/import.go
```diff
@@ -841,9 +845,17 @@ func importUTxOs(
 		txn := cfg.Database.MetadataTxn(true)
 		defer txn.Release()
 
-		if err := store.ImportUtxos(
-			utxos, txn.Metadata(),
-		); err != nil {
+		importer, supportsDeferredRefresh := store.(deferredRewardLiveStakeImporter)
```
cryptodj413: No test exercises ImportLedgerState end-to-end against a store implementing deferredRewardLiveStakeImporter; only the sqlstore-level ImportUtxosDeferredRewardLiveStakeRefresh is tested directly. The invariant this PR relies on, that every return path between here and RebuildRewardLiveStake still reaches the rebuild, is verified only by reading the code, not by a test that would catch a future early return added in between.

cryptodj413: This type assertion re-runs on every batch inside batchCallback; hoist supportsDeferredRefresh/importer above the closure since the concrete store type cannot change mid-import.

cryptodj413 (changes requested):
The deferred-refresh premise holds on every ImportLedgerState path (the rebuild is unconditional and sits outside every phase-checkpoint gate); requesting changes only for missing orchestration-level test coverage of that invariant. Separately: internal/node/backfill.go's comment on RebuildRewardLiveStakeFromRunningTotals is now stale since this PR removes the running-total maintenance it describes, but that file is outside this diff.

ledgerstate/import.go
wolf31o2: Confirmed. dda97f18 adds TestImportLedgerStateRebuildsDeferredRewardLiveStake: it imports two UTxOs sharing one stake key through the real metadata store, requires that store to implement the deferred importer, and requires reward_live_stake to total them. Removing the post-import RebuildRewardLiveStake makes it fail with no aggregate row at all.

wolf31o2: Correct; hoisted in dda97f18 to just below the store lookup, with the reason the deferred path is sound recorded there.

cryptodj413 (approved):

PR #4540, opened 2026-09-20
fix(ledger): compare point hashes in the recovery precedence test

### LINKED ISSUE #3407: ledger: compare complete points during recovery
`ledger/header_validation_recovery.go` compares only `rewindPoint.Slot >= validationErr.BlockPoint.Slot` when deciding whether to rewind. Two points with the same slot but different hashes can therefore skip required recovery or rewind the wrong chain.

Acceptance criteria:

- Compare complete chain points, including hash, at the recovery boundary.
- Add same-slot/different-hash and prior-point tests.

### LINKED ISSUE #4408: ledger: match canonical Byron EBB block-number and slot progression
## Summary

Dingo applies one ordering rule to every new epoch-boundary block. Canonical Byron consensus has an asymmetric transition table:

- regular to regular: block number plus one, later slot;
- regular to EBB: same block number, later slot;
- EBB to regular: block number plus one, same slot permitted;
- EBB to EBB: block number plus one, later slot.

Dingo therefore accepts an EBB child with the same number and slot as an EBB parent, and rejects consecutive EBBs across an otherwise empty epoch.

## Acceptance criteria

- Implement the exact four-case parent/current transition table.
- Retain the epoch-boundary-slot check.
- Use the configured Byron epoch length.
- Test all four parent/current combinations.
- Test an empty epoch producing consecutive EBBs.
- Cover both the accept-invalid equal-number EBB child and the reject-valid next-number EBB child.

wolf31o2:
`tryRecoverFromHeaderValidationError` and
`recoverFromDeterministicTxValidationError` decided whether the ledger tip
precedes the rejected block from the slot alone
(`rewindPoint.Slot >= validationErr.BlockPoint.Slot`).

A Byron epoch-boundary block takes the first slot of its epoch, and the
epoch's first regular block takes the same slot when one is minted there, so a
single chain holds two distinct blocks at one slot at every such boundary.
Mainnet's genesis EBB `89d9b5a5` is the direct parent of `f0f7892b`, and both
carry slot 0; the same pairing occurs at each Byron boundary. `models.Block`
keys blocks by slot and hash, so dingo stores and orders both.

Both recovery sites read that pair as "no rewind target precedes it" and
declined a recovery whose target was the rejected block's own parent. The
rejected block stayed on the primary chain, the pipeline re-read it, and the
loop continued until the stuck-pipeline detector fired. That is the failure
mode this recovery exists to remove.

Both sites now decline only when the ledger tip is at a later slot, or is the
failing block itself by slot and hash. `Chain.ValidateRollback` still resolves
the rewind target by slot and hash and refuses a point the primary chain no
longer holds, so the cross-fork splice guard is unchanged.

Tests cover the same-slot/different-hash rewind and the same-slot/same-hash
decline at both sites.

Fixes #3407

Does not close #4408, which covers the Byron EBB block-number and slot
transition table in header validation rather than the recovery boundary.

---
## Summary by cubic
Fixes ledger recovery at Byron epoch boundaries by comparing rewind targets by slot and hash instead of slot alone. The old slot-only check treated a parent block sharing the rejected block’s slot as not preceding it, so recovery was declined and the pipeline kept re-reading the block until the stuck-pipeline detector fired. Now recovery is allowed whenever the tip is a different block at the same slot and declined only for a later slot or the failing block itself.

- Adds tests that exercise both recovery paths with a same-slot boundary-block pair.
- `Chain.ValidateRollback` still resolves rewind points by slot and hash, so the cross-fork splice guard is unchanged.
- Fixes #3407. Does not close #4408; that issue covers the Byron EBB block-number/slot transition table in header validation, not this recovery boundary.

<sup>Written for commit 7d7838055b3b8a60a127890c273320df6179cad1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4540?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ledger/header_validation_recovery.go
```diff
@@ -272,3 +273,30 @@ func (ls *LedgerState) yieldedToChainSelection(
...
+// hides the failure instead of surfacing it. Any other block at that slot was
+// reached after rewindPoint, because the pipeline only ever reads forward from
+// the ledger tip. A target the primary chain no longer holds is refused
+// downstream by Chain.ValidateRollback, which resolves the point by slot and
+// hash, so this test does not repeat that membership check.
+func recoveryRewindTargetPrecedes(rewindPoint, failing ocommon.Point) bool {
+	if rewindPoint.Slot != failing.Slot {
+		return rewindPoint.Slot < failing.Slot
+	}
+	return !bytes.Equal(rewindPoint.Hash, failing.Hash)
```
chrisguiney, line 301: A same-slot pair can be in either order; any different hash here reads as tip-first. With the tip on the later block and the failing point on the earlier EBB this returns true: on the PR fixture with currentTip = blocks[3] and failing = blocks[2], recovery returns true and rewinds the chain from slot 4 to 3, dropping the valid successor while the failing block stays. Base declines that case. At equal slot, resolve order from chain data (failing PrevHash == rewindPoint.Hash, or block number, since an EBB shares its parent number) and add this prior-point case, which #3407 asks for.

ledger/recovery_same_slot_predecessor_test.go
```diff
@@ -0,0 +1,244 @@
...
+	ls := &LedgerState{
+		db:    db,
+		chain: cm.PrimaryChain(),
+		config: LedgerStateConfig{
+			ChainManager: cm,
+			EventBus:     bus,
+			Logger:       slog.New(slog.NewJSONHandler(io.Discard, nil)),
+		},
+	}
+	ls.currentTip = ledgerTip
```
chrisguiney, line 115: Both SameSlotPredecessor tests panic on the tree merged with current main (b077aa91): nil pointer in updateTipMetrics, reached from rollbackWithOptions, because this fixture builds LedgerState without the metrics init that NewLedgerState always performs. Adding ls.metrics.init(prometheus.NewRegistry()) before this line makes all four new tests pass on the merged tree. Merging main into the branch lets CI see it.

chrisguiney (changes requested):
Two findings inline: same-slot ordering accepts a failing block that is behind the ledger tip, and the new positive tests panic on the tree merged with current main.

PR #4548, opened 2026-09-20
perf(database): defer s3 and gcs blob listings to the first read

### LINKED ISSUE #3281: storage: resume S3 scans server-side
## Problem

- `s3StreamIterator.reset` starts each `ListObjectsV2` scan from the prefix beginning.
- `Seek` discards earlier keys client-side instead of setting a server-side continuation boundary.
- Batched block iteration repeatedly rescans prior objects.
- Full scans are O(n²) in listed objects.

## Acceptance criteria

- S3 iteration resumes with `StartAfter` or an equivalent server-side cursor.
- Ordering and inclusive/exclusive seek semantics remain correct.
- Request-count tests show linear listing work across multiple batches.

wolf31o2:
`NewIterator` rewinds the iterator it returns, and on the s3 and gcs blob stores
a rewind issued its listing immediately. Every blob iterator in `database/`
seeks straight after `NewIterator`, so each one fetched and discarded a page
from the head of the prefix before the seeked listing began. `BlobBlockIterator`
reopens its iterator every `blobIteratorBatchSize` block keys, so it paid that
page once per batch.

`reset` now records the position and leaves the listing to the first `Valid`,
`Err`, `Next`, or `Item`, so a `Seek` replaces a deferred rewind before any
request leaves. A closed iterator no longer reopens a listing on a later read.

Scanning a fake bucket of 5000 blocks (10000 objects: block keys interleaved
with their metadata keys) in batches of 1000:

| | ListObjectsV2 requests | listing from the prefix start | objects returned |
|---|---|---|---|
| before | 21 | 6 | 19004 |
| after | 15 | 0 | 13004 |

Listing work is linear in prefix size either way, 4N-996 objects for N blocks
before and 3N-996 after, not quadratic. The server-side bound that keeps it
linear is `StartAfter` on s3 (#3937) and `Query.StartOffset` on gcs.
`ContinuationToken` is the wrong mechanism for a seek: it is an opaque cursor
from a previous truncated response, and a reopened iterator has only a key to
resume from. The paginator already uses it for pages within one listing. badger
seeks its LSM tree locally and never had the defect.

This was cost, not correctness. The discarded page was listed and thrown away;
no key was dropped or repeated by it. Key order, `resumeKey` exclusivity at a
batch boundary, and end-of-prefix behaviour are unchanged.
`blobIteratorBatchSize` still bounds batch memory and how long one blob read
transaction and store pin stay open. The per-batch reopen remains and now costs
one bounded listing rather than two.

Fixes #3281

---
## Summary by cubic
Blob iterators on S3 and GCS now defer prefix listings until the first read, so seeking right after `NewIterator` no longer fetches and discards a page from the start of the prefix.

- `Seek` replaces a deferred rewind before any listing request is sent, and a closed iterator stays closed on later reads.
- A 5000-block scan in batches of 1000 now takes 15 `ListObjectsV2` requests instead of 21, with no listings from the prefix head.
- Key order, `resumeKey` exclusivity at batch boundaries, and end-of-prefix behavior are unchanged.

Fixes #3281.

<sup>Written for commit 3a9292c625315c065c48cfbeb1e0217861284737. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4548?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cloud blob iteration across batched scans, preserving key order without skipped or repeated results.
  * Seeking and rewinding now avoid unnecessary listing work and continue correctly from the requested position.
  * Closing an iterator now prevents further storage listing requests.
  * Reduced memory usage and the duration of storage operations during large scans.

* **Documentation**
  * Clarified batching behavior and requirements for continuing scans safely.


database/plugin/blob/aws/database.go
```diff
@@ -398,12 +398,51 @@ type s3StreamIterator struct {
...
+	// iterator does so every blobIteratorBatchSize keys -- pay one such
+	// discarded page per batch.
+	pendingSeek []byte
+	started     bool
+	closed      bool
+}
+
+// reset arms the iterator at seek without contacting S3. The listing itself is
+// issued by start, on the first read.
 func (it *s3StreamIterator) reset(seek []byte) {
```
arepala-uml, line 417: reset() never clears closed, and ensureStarted() gates on it.started || it.closed, so a Seek() issued after Close() never re-arms the listing: Valid() and Err() report false/nil forever with no error. No in-tree caller re-Seeks after Close today, but types.BlobIterator documents no such restriction, and this is a real regression from the pre-PR eager-reset, which did re-open on Seek.

database/plugin/blob/gcs/database.go
```diff
@@ -497,22 +497,55 @@ type gcsStreamIterator struct {
...
+	// iterator does so every blobIteratorBatchSize keys -- pay one such
+	// discarded page per batch.
+	pendingStart []byte
+	started      bool
+	closed       bool
+}
+
+// reset arms the iterator at start without contacting GCS. The listing itself
+// is issued by begin, on the first read.
 func (it *gcsStreamIterator) reset(start []byte) {
```
arepala-uml, line 516: Same defect as aws/database.go's s3StreamIterator.reset: closed is never cleared here either, so gcsStreamIterator.ensureStarted (line 531, gated on started || closed) never re-arms after a Seek() that follows Close(), silently returning empty/no-error forever instead of a fresh listing.

arepala-uml (changes requested):
Deferred-listing call ordering, BlobBlockIterator's bounded reopen, and s3/gcs semantics all verified correct on a merged, race-tested tree. Blocking on one behavioral regression present in both backends: Seek after Close permanently and silently deadens the iterator.

Issue #2987, opened 2026-07-28
preview epoch 2: epoch_summary and reward_pool_input rows missing, causing 3 pools to show pool_only_koios


cryptodj413:
### Summary
koios-parity explain reports preview epoch 2 as missing epoch_summary entirely, and 3 pools that Koios lists as active for that epoch have no reward_pool_input row in Dingo. Both symptoms point to a single root cause: the epoch-boundary snapshot for epoch 2 was never persisted (or was silently skipped) on this node.

### Command and Output
```
$ ./koios-parity explain --epoch 2 --network preview
preview epoch 2
  [dingo_db_missing] 1
    pool=(epoch)                                                 field=epoch_summary          dingo=                     koios=present
  [pool_only_koios] 3
    pool=pool16h8ugt8k0a4kxa5g6x062zjrgfjc7cehpw0ze8374axlul76932 field=pool_presence          dingo=                     koios=present
    pool=pool18r62tz408lkgfu6pq5svwzkh2vslkeg6mf72qf3h8njgvzhx9ce field=pool_presence          dingo=                     koios=present
    pool=pool1grvqd4eu354qervmr62uew0nsrjqedx5kglldeqr4c29vv59rku field=pool_presence          dingo=                     koios=present
```

PR #4549, opened 2026-09-20
fix(ledger): reconstruct retention-pruned reward stake inputs

arepala-uml:
## Summary

- `reward_stake_input` (per-credential reward basis) is pruned to a retention window while `reward_snapshot` and `reward_pool_input` are retained for the life of the database (dingo #2987). A node resumed via `dingo database truncate` can reach an epoch boundary whose required reward-lag snapshot already had its `reward_stake_input` rows aged out by ordinary retention during the run *before* the rollback ever happened -- not a genuine rewind past retention, which is what the existing skip guarded against.
- `calculateStakeRewardApplication` previously skipped that round outright, permanently under-crediting every pool for that epoch. Since a bootstrap epoch's leader-eligibility stake includes reward-account balances, the shortfall later surfaced as a VRF leader-eligibility rejection of an already-canonical block, halting the node.
- Reconstructs the missing rows instead, using the same historical CTE (`GetEpochBoundaryRewardStakeInputsForPools`) the live epoch-boundary capture already falls back to, keyed on the still-retained `reward_snapshot`'s own captured/boundary slot.

## Fix details

Three retained totals are reconciled against the reconstruction before it's used, falling back to the existing skip if a reconciliation can't be made safely:

- **Delegated stake**: a small (observed up to ~6e-9 relative) disagreement from the live aggregate's own rounding is folded into the pool's largest credential.
- **Owner stake**: the reconstruction has no notion of pool ownership on its own; ownership is resolved from each pool's certificate effective for the ended epoch, then reconciled via a compensating transfer between the largest owner and non-owner credential (so the pool total stays exact).
- **Delegator count**: the two reconstructions apply the same "drop zero-stake rows" rule but can disagree by a row when a credential's amount rounds to a small nonzero value on one side and to exactly zero on the other; the excess row's stake is folded into the pool's largest non-owner credential.

## Evidence

Found and verified via a real `dingo database truncate` + replay against historical Preview chain data (not synthetic): the retention gap was blocking a VRF leader-eligibility check at the first post-truncate epoch boundary. After this fix, the same replay cleared that boundary and progressed through several further epochs without a skip.

## Test plan

- [x] `go build ./...`
- [x] `go test ./ledger/...` (all packages pass)
- [x] New test `TestApplyStakeRewardsReconstructsRetentionPrunedInputs`: seeds a real pool/owner/delegator/UTxO/reward fixture, legitimately captures its mark snapshot, prunes `reward_stake_input` to simulate retention, then asserts the round reconstructs and applies (persists a pool output, credits the owner, re-populates `reward_stake_input`) instead of skipping. Confirmed it fails (no pool output persisted) with the fix reverted.
- [x] Existing retention-skip tests (`TestApplyStakeRewardsSkipsPrunedStakeInputs`, `TestApplyStakeRewardsAcceptsZeroDelegatorSnapshot`, `TestApplyStakeRewardsSkipsPrunedStakeInputsReportsLoudly`, `TestSkippedPrunedStakeInputsSuppressedDuringPrecompute`) pass unchanged -- the case with no real underlying data to reconstruct still falls through to the original skip.
- [x] `golangci-lint run ./ledger/...`, `go vet ./ledger/...`, `make golines`, `make import-boundaries`, `make docs-parity`

## Summary by CodeRabbit

- **Bug Fixes**
  - Reward calculations now recover eligible stake data after retention pruning when the required reward records remain available.
  - Previously skipped reward rounds can now proceed when reconstructed data matches retained totals.
  - Pool outputs and owner rewards are correctly generated, including valid zero-member payouts.
  - Reconstructed stake information is persisted for future processing.

---
## Summary by cubic
Fixes dingo #2987 by reconstructing retention-pruned `reward_stake_input` rows instead of skipping the reward round. A node resumed from `dingo database truncate` could hit an epoch whose per-credential reward inputs aged out before the rollback, so the round was skipped, pools were permanently under-credited, and the shortfall later halted a VRF leader-eligibility check. The ledger now rebuilds the missing rows from the retained `reward_snapshot` and `reward_pool_input` data with the same historical CTE live captures use.

**Behavior**

- The reconstructed rows must reconcile with each pool's retained delegated-stake, owner-stake, and delegator-count totals; small rounding gaps are folded into the largest applicable credential so totals stay exact.
- If the reconstruction cannot be reconciled safely, the epoch falls back to the previous skip path.
- Rebuilt rows are persisted for later reads; if that write fails, the round still applies with the in-memory rows.
- Verified with a real `dingo database truncate` + replay on Preview data: the previously failing epoch now applies and later epochs continue normally.

<sup>Written for commit a4e2343cdfd12611a96da4fc0b55a6390c2ed9bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4549?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


arepala-uml:
Addressed all 4 findings in 213253f0:

1. **Owner-stake transfer underflow guard (major)** — fixed. The donor is now determined explicitly (whichever row diff is subtracted from) before validating its balance against diff, rather than checking the wrong party depending on branch.
2. **Non-atomic reconstructed stake-input persistence (major)** — fixed. Added `DeleteRewardStakeInputsForEpoch` (stake inputs only, leaving `reward_pool_input` untouched) and call it to clean up a partial write on save failure, so a later read sees either a complete set or none rather than a stuck partial one.
3. **Overlapping epoch fixtures in the new test (minor)** — fixed. Epoch 1 now starts at the boundary slot (100) and is set only after the SNAP-point read, matching the live rollover's own ordering, so `GetEpochBySlot(99)` resolves unambiguously.
4. **Contradictory comment (minor)** — removed, applied your suggested diff as-is.

Full test suite, golangci-lint, go vet, and golines all clean after the fixes.

wolf31o2 (approved):

PR #4551, opened 2026-09-20
fix(ouroboros): contain decoder panics on peer tx bodies

wolf31o2:
A peer answering `MsgReplyTxs` with bytes that panic the ledger decoder took the
node process down. `validateTxsubmissionReply` decoded each reply body with a
bare `ledger.NewTransactionFromCbor` inside the per-peer txsubmission goroutine,
which has no `recover()` above it: `ouroboros/` has none outside
`decode_cache.go`, and `mempool/` has none at all. Only block bodies and headers
were contained, through `decodeWithPanicSafeMetrics`.

gouroboros is not a backstop. `cbor.Value.processMap` recovers exactly one panic
class, an unhashable Go map key, and re-panics every other one.

## Change

`internal/safedecode.Guard` runs a decode and converts a panic into an error
wrapping `ErrDecodePanic`. Every decode of externally-supplied transaction bytes
now routes through it:

| Path | Site |
| --- | --- |
| node-to-node txsubmission reply | `ouroboros/txsubmission.go` |
| mempool admission and re-validation | `mempool/mempool.go` |
| utxorpc `SubmitTx` | `api/utxorpc/submit.go` |
| mesh construction submit and parse | `api/mesh/construction.go`, `api/mesh/convert.go` |
| blockfrost submit and evaluate | `api/blockfrost/adapter.go` |

`validateTxsubmissionReply` takes the decoder as a parameter, so the containment
around it can be driven by a decoder that genuinely panics rather than by a
stubbed-out recover.

## Behavior

A recovered panic becomes the same `txsubmission reply transaction %d decode
failed` error a malformed body already produces, and so leaves by the route that
already exists: the whole reply is dropped, no body is admitted, the
connection's mismatch streak advances, and tx ingest stops after three
consecutive rejections. `errors.Is(err, safedecode.ErrDecodePanic)`
distinguishes it from an ordinary decode failure.

Containment is safe to continue from because the decode is side-effect-free: it
builds a fresh value from the input bytes, holds no lock, and mutates nothing
the caller shares. `validateTxsubmissionReply` publishes nothing until the whole
reply validates, and `AddTransaction` decodes before taking the mempool lock.
`Guard`'s doc comment records that it must not be wrapped around work that
mutates shared state, where recovering would convert a crash into corruption.

`Guard` tracks completion separately from `recover()`'s return value, because
`recover()` yields nil both when nothing panicked and when the panic value is
itself nil; testing the recovered value alone would report an aborted decode as
a successful one.

Related to blinklabs-io/gouroboros#2075, which covers the complementary
containment inside gouroboros' own handler boundary. This does not close it.

---
## Summary by cubic
Contain decoder panics on peer transaction bytes so a malformed reply cannot crash the node. Previously, a `MsgReplyTxs` body that panicked the ledger decoder killed the process because the txsubmission goroutine had no recovery. Now every decode of externally-supplied transaction bytes goes through a new `internal/safedecode` guard that turns a panic into an error wrapping `ErrDecodePanic`. That error follows the same path as a malformed body: the whole reply is dropped, no transaction is admitted, and the mismatch streak advances.

**Notes**
- Covers txsubmission reply validation, mempool admission/revalidation, and the utxorpc, mesh, and blockfrost submit APIs.
- `validateTxsubmissionReply` now takes a decoder parameter so tests can inject a panicking decoder.

<sup>Written for commit 6b2df47bf5387de3a50e02fdcad6f41286c06139. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/dingo/pull/4551?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


api/utxorpc/submit.go
```diff
@@ -72,7 +73,7 @@ func (s *submitServiceServer) SubmitTx(
 	if err != nil {
 		return nil, fmt.Errorf("failed decoding tx: %w", err)
 	}
-	tx, err := gledger.NewTransactionFromCbor(txType, txRawBytes)
+	tx, err := safedecode.Transaction(txType, txRawBytes)
```
arepala-uml, line 76: safedecode.Transaction here still panics unguarded: DetermineTransactionType (submit.go:72) probes eras via the same NewXTransactionFromCbor constructors NewTransactionFromCbor dispatches to (gouroboros ledger/tx.go:117-190), unguarded, before this line runs. Same pattern at mesh/construction.go:927, mesh/convert.go:373, blockfrost/adapter.go:3832 and :3899. Separately, EvalTx in this file (submit.go:283) still calls bare gledger.NewTransactionFromCbor, untouched by this PR, on the same untrusted gRPC input SubmitTx was just fixed for.

internal/safedecode/safedecode.go
```diff
@@ -0,0 +1,78 @@
...
+// value: decode's return values are assigned only on its normal return, so a
+// panicking decode never publishes a partially-built value to the caller.
+//
+// Containment is sound here only because decoding is a pure function of its
+// input bytes -- it builds a fresh value, takes no lock and mutates no state
+// the caller shares. Do not reach for Guard around work that mutates shared
+// state: recovering there converts a crash into silent corruption, which is
+// worse.
+//
+// completed, not recover()'s return value, decides whether decode panicked.
```
arepala-uml, line 49: This justifies tracking completed separately by saying recover() returns nil both for no panic and for panic(nil); under this module's go.mod toolchain (go1.26.7, Go 1.21+ panic(nil) semantics) recover() instead returns a non-nil *runtime.PanicNilError for panic(nil), so the stated rationale is inaccurate for the toolchain in use. The completed tracking itself is still correct and worth keeping.

arepala-uml (changes requested):
Guard, the txsubmission peer path, and the mempool lock ordering are sound and verified with a causal fail-before revert. Blocking on two independent gaps that leave five of the six new call sites still panicking on crafted input.

PR #4555, opened 2026-09-20
fix(leios): contain decoder panics on peer Leios bytes

wolf31o2:
The Leios overlay decoded peer-supplied bytes with no `recover()` above the call. The leios-fetch and leios-notify workers run on their own goroutines, and gouroboros' `cbor.Value` decoder recovers exactly one panic class (an unhashable Go map key) and re-panics every other one, so an endorser-block manifest, a transaction envelope or an announcement header whose bytes panicked the decoder took the node process down.

## Change

Every decode of peer-supplied Leios bytes routes through `internal/safedecode`, so a panic becomes an error wrapping `ErrDecodePanic` on the decode-failure route the caller already has.

| Site | Bytes | Mechanism |
| --- | --- | --- |
| `ouroboros/leios_merged.go` x5 (`validateLeiosEndorserBlockTxs`, `leiosEndorserBlockTxValidator`, `storeLeiosEndorserBlock`, `loadLeiosEBFromDB`, `EndorserBlockTxHashesByHash`) | endorser-block manifest | `decodeLeiosEndorserBlock` over `Guard` |
| `ouroboros/leiosnotify.go` (`acceptLeiosAnnouncementInternal`) | ranking-block header from a block announcement | `decodeLeiosAnnouncementHeader` over `Guard` |
| `ouroboros/leios_merged.go` (`validateLeiosEndorserBlockTx`) | endorser transaction envelope | `safedecode.Cbor` |
| `ledger/leios_apply.go` (`decodeEndorserTxEnvelope`) | endorser transaction envelope | `safedecode.Cbor` |
| `ledger/leios_apply.go` (`applyEndorserBlock`) | endorser transaction | `safedecode.Transaction` |
| `internal/leiosheader/ref.go` (`legacyExtensionRef`) | legacy Leios header extension | `safedecode.Cbor` |

`safedecode.Cbor` adds the `cbor.Decode` form of `Guard`, returning the byte count so callers keep their trailing-bytes checks.

## Why not the shared decode cache

`decodeWithPanicSafeMetrics` is the wrong containment for all six:

- `LeiosEndorserBlock` deliberately does not implement `ledger.Block`, so neither existing `decodeCache` instance can hold one. Routing manifests through it means a third cache and a third metrics pair, not reuse.
- `storeLeiosEndorserBlock` decodes under `leiosAnnouncementsMu`. `getOrDecode` can park its caller on another goroutine's in-flight decode, which would add a wait-for-goroutine dependency under that lock.
- The header cache key `hashDecodeInput(blockType, bytes)` identifies the input, not the decoder, and the cache is populated by `decodeChainsyncHeader`, whose decoder choice depends on `NetworkMagic` and block type and which additionally warms `header.Hash()`. Leios networking is enabled by `--run-mode leios` or a Dijkstra start era on any network, so on a non-Musashi network the same header bytes would be decoded by two different decoders under one key.

The decoded manifest is also consumed immediately at every call site, behind the `(slot, hash)`-keyed endorser-block cache that already deduplicates these bytes across connections.

## Side-effect freedom

Every guarded decode is a pure function of its input bytes: it builds a fresh value, takes no lock, performs no I/O and mutates nothing the caller shares, so a contained panic cannot leave half-updated state. At the manifest site under `leiosAnnouncementsMu`, which is released by explicit `Unlock` calls rather than a defer, the error path is what releases the lock; a panic there would have held it for the life of the process. In `applyEndorserBlock` the decode loop runs before any storage is mutated.

## Reachability

These paths sit behind `experimentalLeiosNetworkingEnabled`: `-n musashi`, `--run-mode leios`, or a Dijkstra start era. A default-configured node does not offer the Leios mini-protocols and cannot reach any of them.

## Class review

`leiosnotify`, `leios_merged`, `leiosfetch`, `leiosvotes`, the Leios vote, pipeline, backfill and persist handlers, `ledger/leios/*`, `internal/leiosheader`, `node_leios.go` and `database/leios.go` were walked for decodes of peer-supplied bytes. Beyond the six sites above, three more were found and are fixed here: the two envelope decodes in `ledger/leios_apply.go`, its `NewTransactionFromCbor` call, and the header-extension decodes in `internal/leiosheader`. No Peras code path decodes peer bytes; the only Peras reference in the Leios files is a comment about a block element that is preserved rather than decoded.

Not guarded, deliberately: `leiosAnnouncementFromBlockCbor`, `certifiedEndorserBlockHash` and `spliceEndorserTxsIntoDijkstraBlock` decode ranking-block CBOR read back from the local block store, whose ingest decode already ran under `decodeWithPanicSafeMetrics`.

## Tests

`TestGuardedDecodeContainsDecoderPanic` and `TestCborContainsDecoderPanic` drive the containment with decoders that genuinely panic - the latter through a real `cbor.Decode` whose `UnmarshalCBOR` panics - for string, error and nil panic values. With the guard removed in place they crash the test binary. Positives are kept: valid manifests, the real Musashi Dijkstra header fixture, and valid CBOR still decode, and malformed-but-non-panicking input still yields its ordinary error, asserted with `NotErrorIs(ErrDecodePanic)`. `TestStoreLeiosEndorserBlockDecodeFailureReleasesAnnouncementLock` pins that a failed manifest decode leaves `leiosAnnouncementsMu` free and caches nothing.

## Dependency

Depends on #4551, which introduces `internal/safedecode`. Its branch is merged into this one so this builds; #4551 must merge first. Related to blinklabs-io/gouroboros#2075 and blinklabs-io/dingo#3511; this closes neither.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resilience when processing malformed or unexpected transaction and protocol data.
  - Decoder failures are now reported as errors instead of crashing protocol workers or connections.
  - Failed Leios data processing leaves ledger state and announcement tracking consistent.
  - Transaction submission, evaluation, validation, and mempool processing now handle invalid encoded transactions more safely.

- **Tests**
  - Added coverage for decoder failures, panic handling, malformed inputs, and state cleanup.


ledger/leios_apply.go
```diff
@@ -183,7 +190,11 @@ func (ls *LedgerState) applyEndorserBlock(
 		// DetermineTransactionType is heuristic and cannot reliably identify a
 		// bare standalone transaction without block/era context (it returns
 		// "unknown transaction type" for these), so it must not be used here.
-		tx, err := ledger.NewTransactionFromCbor(ledger.TxTypeDijkstra, txCbor)
+		// Peer-supplied transaction bytes, decoded before any storage is
+		// mutated, so the guard cannot convert a crash into a partially
+		// applied endorser block: every return below this loop's decode
+		// failure leaves the ledger untouched.
+		tx, err := safedecode.Transaction(ledger.TxTypeDijkstra, txCbor)
```
arepala-uml, line 197: This decode is now guarded, but leiosAnnouncementFromBlockCbor a few lines up (around line 1221 in this file) still does a raw cbor.Decode of DijkstraBlockHeader with no safedecode guard, on the same block-application path hit during ordinary chain-sync of certifying blocks - same fatal-panic class this PR fixes here.

ouroboros/leios_merged.go
```diff
@@ -543,7 +549,11 @@ func (o *Ouroboros) storeLeiosEndorserBlock(
 	o.leiosAnnouncementsMu.Lock()
 	verified := origin == leiosStoreAuthoritative ||
 		o.leiosAnnouncementBindsSlotLocked(point.Hash, point.Slot)
-	block, err := lcommon.NewLeiosEndorserBlockFromCbor(blockRaw)
+	// leiosAnnouncementsMu is released by explicit Unlock calls on this path,
+	// not by a defer, so this decode must fail by returning rather than by
+	// unwinding: a panic here would leave the lock held for the life of the
+	// process. As an error it takes the branch below, which unlocks.
+	block, err := decodeLeiosEndorserBlock(blockRaw)
```
arepala-uml, line 556: This call is guarded, but leiosAnnouncementFromBlockCbor (around line 1418) and certifiedEndorserBlockHash (around line 1454) further down in this same file still raw cbor.Decode DijkstraBlockHeader with no safedecode guard - same missing-guard class, and certifiedEndorserBlockHash is served over NtC chainsync for certified blocks.

arepala-uml (changes requested):
Requesting changes: the peer-decode panic class this PR closes is not fully closed - ledger/leios_apply.go:1221 and ouroboros/leios_merged.go:1418/1454 still raw-decode DijkstraBlockHeader with no safedecode guard; see inline comments.

